PR #60484, opened 2022-08-27
power_draw in energy

Hirmuolio:
#### Summary
Infrastructure "power_draw in energy"

#### Purpose of change

Work towards units::energy based batteries.

#### Describe the solution

Currently power_draw is integer number of mW.
Use units::energy (per second) for it instead.

Create units::power and use that. But that would always be converted to units::energy when used so it wouldn't help much.
If someone made system so that in code units::energy/time_duration=units::power then units::power could be used without making code unnecessarily complicated.

#### Testing

Tests pass.
Heavy duty flashlight consumes about same amount of energy before and after when left on for 6 hours.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -13265,7 +13265,8 @@ bool item::process_tool( Character *carrier, const tripoint &pos )
         // kJ (battery unit) per second
         energy = units::to_kilojoule( type->tool->power_draw );
         // energy_bat remainder results in chance at additional charge/discharge
-        energy += x_in_y( units::to_millijoule( type->tool->power_draw ) % 1000000, 1000000 ) ? 1 : 0;
+        const int kj_in_mj = units::to_millijoule( 1_kJ );
+        energy += x_in_y( units::to_millijoule( type->tool->power_draw ) % kj_in_mj, kj_in_mj ) ? 1 : 0;
     }
 
     if( energy > 0 ) {


dseguin (approved):

src/item.cpp
```diff
@@ -13261,11 +13261,11 @@ bool item::process_tool( Character *carrier, const tripoint &pos )
...
         energy = std::max( ammo_required(), 1 );
-    } else if( type->tool->power_draw > 0 ) {
-        // power_draw in mW / 1000000 to give kJ (battery unit) per second
-        energy = type->tool->power_draw / 1000000;
+    } else if( type->tool->power_draw > 0_J ) {
+        // kJ (battery unit) per second
+        energy = units::to_kilojoule( type->tool->power_draw );
         // energy_bat remainder results in chance at additional charge/discharge
-        energy += x_in_y( type->tool->power_draw % 1000000, 1000000 ) ? 1 : 0;
+        energy += x_in_y( units::to_millijoule( type->tool->power_draw ) % 1000000, 1000000 ) ? 1 : 0;
```
anothersimulacrum, line 13268:
```suggestion
        const int mj_in_kj = units::to_millijoule( 1_kJ );
        energy += x_in_y( units::to_millijoule( type->tool->power_draw ) % mj_in_kj, mj_in_kj ) ? 1 : 0;
```

anothersimulacrum (approved):

PR #60490, opened 2022-08-27
[Xedra Evolved] Fix eater spells deal damage to player

GuardianDll:
#### Summary
None

#### Purpose of change

`"energy_source": "HP"` consume little amount of HP even when `"base_energy_cost": 0`

#### Describe the solution

delete `"energy_source": "HP"` at all
also delete `"final_energy_cost": 0`, because it equal `base_energy_cost` by default

#### Testing

Now i can eat dreamross and not bite my ~~tongue~~ torso each time (also no need to pick a body part each cast)


Maleclypse (approved):
Thanks for catching that!

PR #60496, opened 2022-08-27
Place explosive slug in the obsolete.json for lack of purpose.

### LINKED ISSUE #60437: Explosive slugs do shit all
### Describe the bug

Fired explosive slugs seem to have little to no effect on terrain, target, etc. The damage appears to calculate as ~10 damage rolled against every limb individually, but interacts with armor values and falls off dramatically with distance. The result is a projectile that underperforms severely compared to even 00 shot, which is unusual given the rarity of the explosive slug.

### Steps to reproduce

Load an explosive slug into a Mossberg 500.
Shoot an enemy.
Cry because the explosive slug does hardly anything.

### Expected behavior

Whatever gets hit with the slug is vaporized into a fine mist, along with everything else in a 10-foot radius.

### Screenshots

image
This is a fresh character wearing a set of plate armor with no helmet and shooting the wall directly in front of himself with explosive slugs. You can clearly see that his unprotected head is only taking 5-10 damage per refreshing blast of explosive breeze, while his armored limbs are completely unharmed.

Game version: 0.F-9d29e93-Tiles, Build 2022-08-25-0113
Default mods only

### Additional context

_No response_

Marcos6226:
#### Summary
Balance "Place explosive slug in the obsolete.json, as currently it has no purpose."

#### Purpose of change
Closes #60437 

Check #60437 for full details, but basically the explosive slugs are not realistic, and for lack of information and technical problems, it was suggested to simply remove them until we have the possibility to implement them again.

#### Describe the solution
I simply moved it to `/items/obsolete.json` and `/json/obsolete.json` until we can implement it in a better way than it's currently.

Leaving them alone and not bothering about it being anything other than a shotgun-fired firework.

#### Testing
Didn't throw any errors so should be good.

#### Additional context
If you disagree, please express your thoughts in the comments section.

### DIFF
--- a/data/json/obsolete.json
+++ b/data/json/obsolete.json
@@ -735,5 +735,16 @@
     "volume": "10000 ml",
     "bashing": 10,
     "to_hit": -5
+  },
+  {
+    "type": "uncraft",
+    "activity_level": "MODERATE_EXERCISE",
+    "result": "shot_he",
+    "skill_used": "gun",
+    "difficulty": 5,
+    "time": "50 s",
+    "//": "A little more carefully...",
+    "qualities": [ { "id": "PULL", "level": 1 } ],
+    "components": [ [ [ "shot_hull", 1 ] ], [ [ "shotgun_primer", 1 ] ], [ [ "gunpowder_shotgun", 8 ] ], [ [ "chem_rdx", 1 ] ] ]
   }
 ]


kevingranade:
"Lack of purpose" is not a reason for obsoletion.

Marcos6226:
> "Lack of purpose" is not a reason for obsoletion.

For future reference, could you please tell me what are some valid reasons for obsoletion? This way, myself and others don't repeat the same mistake. Thanks.

PR #60500, opened 2022-08-27
Add looks_like to Freezer, Glass Fridge, Combo Oven

ANickelN:
#### Summary
Interface "Add looks_like to Freezer, Glass Fridge, Combo Oven"

#### Purpose of change

Noticed that the recent addition of the combo oven/microwave did not have a proper sprite yet, decided they were similar enough to the regular oven to warrant a looks_like.  Did the same with the glass fridge and the freezer looking like a fridge for any tilesets that were missing those.

#### Describe the solution

Add necessary lines of JSON.

None

#### Testing

cataclysm-tiles_gdYFTiZyvJ


Maleclypse:
I had decided not to add the looks_like since the combo_oven doesn't contain fire.  But I'll keep an eye out for issues of people setting fires on it and burning down houses on accident if so remove the looks_like.

ANickelN:
> I had decided not to add the looks_like since the combo_oven doesn't contain fire. But I'll keep an eye out for issues of people setting fires on it and burning down houses on accident if so remove the looks_like.

This is a separate thing, but why isn't the gas oven able to "fire_container" if the standard electric oven is?  This seems like a strange inconsistency, even if the justification is supposed to be "it could be dangerous".  Just have them both work the same way, as I don't really see a real reason to justify having them separated.

Maleclypse:
> > I had decided not to add the looks_like since the combo_oven doesn't contain fire. But I'll keep an eye out for issues of people setting fires on it and burning down houses on accident if so remove the looks_like.
> 
> This is a separate thing, but why isn't the gas oven able to "fire_container" if the standard electric oven is? This seems like a strange inconsistency, even if the justification is supposed to be "it could be dangerous". Just have them both work the same way, as I don't really see a real reason to justify separating them in such a way.

I was living in New England when a gas line blew up a whole neighborhood a few towns over.  So that's the inspiration.  This was only a couple years back.

ANickelN:
> > > I had decided not to add the looks_like since the combo_oven doesn't contain fire. But I'll keep an eye out for issues of people setting fires on it and burning down houses on accident if so remove the looks_like.
> > 
> > 
> > This is a separate thing, but why isn't the gas oven able to "fire_container" if the standard electric oven is? This seems like a strange inconsistency, even if the justification is supposed to be "it could be dangerous". Just have them both work the same way, as I don't really see a real reason to justify separating them in such a way.
> 
> I was living in New England when a gas line issue blew up a whole neighborhood a few towns over. So that's the inspiration. This was only a couple years back.
This isn't really the proper place to discuss this point anyways, I can open an issue about it and can be discussed there.  The fridges and freezers being look_like seems good?

Maleclypse (approved):

PR #60505, opened 2022-08-27
Integrated armor for scale mutations

anoobindisguise:
#### Summary
Content "Integrated Armor for Scales Mutations"

#### Purpose of change
Scales didn't have integrated armor!

#### Describe the solution
There are three scales mutations - "basic" scales, fish scales and reptile scales. I haven't changed these into multiple tiers, but instead just gave the ones that were already there integrated armors.
Scales are made of a combo of bone and chitin with an underlying flesh layer. I chose those materials since they are similar to real life scales - keratin and chitin are common features of real scales.
The basic scales mutation is just scales on parts of your body, so the coverage is low. This is the "oh shit I started growing fingernails in places they don't belong" stage. Does not interfere with the skintight layer.
Fish scales are a bit thicker and have insanely high coverage, in addition to their usual water-friendly properties.  Moderate encumbrance, occupies the skintight layer. 
Reptile scales have high coverage, though not as high as that of fish scales, and are the thickest by far. Underlying flesh is partially leather and much tougher. Occupies normal layer, and its encumbrance is high (similar to wearing full chainmail)

More granular scale stages.
Make PERSONAL not conflict with other PERSONAL items so that it and Epicuticle won't interfere with Subdermal Carbon Filament etc. Out of scope imo.

#### Additional context
https://www.youtube.com/watch?v=0TSmMqZPCY4


Terrorforge:
The mutation is called "sparse scales", but the integrated armor is called "patchy scales". It would probably be better if they had the same name, just to avoid confusion.

Also it strikes me as a bit weird that reptile scales are pretty much just worse than [gray fur](https://nornagon.github.io/cdda-guide/#/item/integrated_lupine_fur) in every way that matters. I feel like the protective power of reptilian scales ought to be a selling point of the Lizard line. Granted, this is probably more a consequence of Gray Fur being way better than it has any right to be.

It also has a very similar profile to the [steel-plated Kevlar jumpsuit](https://nornagon.github.io/cdda-guide/#/item/hsurvivor_jumpsuit)*, while also taking up the Skintight layer. I don't think we necessarily _have_ to balance around that, but it seems like a shame because almost every other mutation armor brings something unique to the table. One thing that's nearly universal is that mutated armors tend to have better encumbrance than comparable manufactured armor, which makes sense to me. Of course your own skin gets in the way less than if a human was wearing something of equivalent thickness and rigidity. So honestly, I think you could drop the encumbrance on the reptile scales all the way down to like 4 (fish scales down to 2, why not) _and_ add another mm to both bone and chitin and it wouldn't be out of line. Like, you know what other integrated armor is normal-layer with 10 encumbrance? [Distributed bark](https://nornagon.github.io/cdda-guide/#/item/integrated_bark2_a), with has an average bash resist of 34 and an absolute minimum of 18.

*Note that this thing is Varsize, so in practice its 17 encumbrance is actually 8.

anoobindisguise:
> The mutation is called "sparse scales"...
> 
> Also it strikes me as a bit weird that reptile scales...
> 
> The reptile scales also...

Good points all around. Applied some suggestions. Reptile scales is still pretty encumbering though since it no longer provides Dex -2. But they should be mildly more protective now.

Terrorforge:
I don't think that -2 Dex was ever justified, particularly since it was just canceled out by Very Dextrous. And when I think "lizard", I don't really think "slow and clumsy". Granted, in many ways the Lizard line is more of a Crocodile line, but those guys aren't lacking for speed and precision when they need it, either.

Buuuut mostly I just really want Lizard to be remotely worth it, and 6 encumbrance on that is when I start to feel like I'd be getting a good deal, because taking up two layers is a pretty big cost.

anoobindisguise:
> I don't think that -2 Dex was ever justified, particularly since it was just canceled out by Very Dextrous. And when I think "lizard", I don't really think "slow and clumsy". Granted, in many ways the Lizard line is more of a Crocodile line, but those guys aren't lacking for speed and precision when they need it, either.
> 
> Buuuut mostly I just really want Lizard to be remotely worth it, and 6 encumbrance on that is when I start to feel like I'd be getting a good deal, because taking up two layers is a pretty big cost.

I think you might be underestimating the new defensive capabilities. That's 3/8/5 100% of the time, with a 95% chance of 10/16/5.6 and a 95% chance of 8/12/4. It's pretty pushed and 8 encumbrance feels pretty fair to me for that level of thickness and protection. There's a 90% chance you're seeing 21/36/14.6, and that's enough to be practically indestructible, fitting for a post threshold mutation. The encumbrance is about as low as I think it could reasonably be.

Terrorforge:
I suppose that's the thing. It's fair, whereas most integrated armor is kinda bullshit. But you do have a point. I'm mostly comparing to steel-plated jumpsuit + nomad bodymesh, and since we're talking about armor values that will usually absorb the entirety of an attack, 90% chance of full protection is actually twice as good as the jumpsuit's ~80%. And of course it goes over your entire body and doesn't interfere with your other mutations. I still think it could be pushed a teensy bit further, but it's fine and I'm not gonna argue any more. Thank you for taking my suggestions into consideration.

Just as a point of interest, though: the super high cut resistance, which you'd think would be a nice perk, turns out to be kind of redundant. I had a look at this the other day and, with the possible exception of some special attacks I probably missed, the highest cut damage I could find in the entire game was 15. Granted, not a lot of things that deal more bashing than that either, but at least those things do exist.

PR #60506, opened 2022-08-27
Floodlight/Lamp barter value tweaking

cherdaq:
#### Summary
None

#### Purpose of change
It's pretty strange that floodlights (pre-Cata 900$ each) cost 5$ per unit in barter, while the regular lamps (pre-Cata 90$ each) cost 50$ per unit, basically ten times more for the weaker version.

#### Describe the solution
Swap the barter values of floodlight and lamp appliances, also carry over that change to the vehicle part versions of floodlights.

Keep the current barter value of floodlights but turn the value of lamps down even further than 5$

#### Testing
There isn't much to test, it's a value swap.

#### Additional context
None

### DIFF
--- a/data/json/items/vehicle/lights.json
+++ b/data/json/items/vehicle/lights.json
@@ -93,7 +93,7 @@
     "volume": "2 L",
     "bashing": 5,
     "price": 90000,
-    "price_postapoc": 5000,
+    "price_postapoc": 1500,
     "copy-from": "car_headlight"
   },
   {


data/json/items/appliances.json
```diff
@@ -270,7 +270,7 @@
     "longest_side": "170 cm",
     "bashing": 5,
     "price": 90000,
-    "price_postapoc": 500,
+    "price_postapoc": 5000,
```
Maleclypse, line 273:
```suggestion
    "price_postapoc": 1500,
```
three times the value rather than ten

data/json/items/vehicle/lights.json
```diff
@@ -93,7 +93,7 @@
     "volume": "2 L",
     "bashing": 5,
     "price": 90000,
-    "price_postapoc": 500,
+    "price_postapoc": 5000,
```
Maleclypse, line 96:
```suggestion
    "price_postapoc": 1500,
```

Maleclypse (approved):

PR #60513, opened 2022-08-28
Fixing plushie shark weight and length

LordBarkBread:
#### Summary
Content "Fixes plushie shark"

#### Purpose of change
I'm a goof and made a goof with my previous PR

#### Describe the solution
Correctly input the plushies weight and gave it proper length as it should be

~~I should stop being a goofball~~


ashGlaw:
Blåhaj?

PR #60514, opened 2022-08-28
directed_push documentation and MAGIC.md changes

RedMisao:
#### Summary
Infrastructure "Documentation: MAGIC.md changes"

#### Purpose of change

A bunch of editions to MAGIC.md: First added documentation for the already existing `directed_push` spell `effect`, alongside a short description. Then added an index, reordered several lists, reordered some sections, added tables (for further ordering), added descriptions to existing subsections, added new subsections, checked grammar.

#### Testing

For `"effect": "directed_push"` testing: 
Made a chained, the main spell targets `hostile` creatures and deals 5 bash damage, the subspell has a `directed_push` component with damage value of 6 (i.e. should push back enemies a distance of 6 tiles), tested it on a random zombie I found.
a

b

Then, changed the damage value of the spell to -6 (i.e. should pull enemies a distance of 6 tiles), used it on the same testing subject:

c

d

Do note that, when the actual distance to the target is less than the pull distance, the zombie ended up on top of the PC.

e

A similar effect happens when you use several `area_pull` with `item` as targets: the items are pulled until they move to the same tile as the caster. I don't know if this is intended behavior or a bug, at least when applied to creatures.

### DIFF
--- a/doc/MAGIC.md
+++ b/doc/MAGIC.md
@@ -466,7 +466,7 @@ Note: `valid_targets` has both `ground` and `hostile` so it can be targeted in a
   }
 ```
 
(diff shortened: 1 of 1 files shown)


GuardianDll:
Good job!

PR #60516, opened 2022-08-28
fix cables

### LINKED ISSUE #60512: Extension cord item not removed after connecting two appliances
### Describe the bug

After connecting two appliances with extension cord, said extension cord item stays in your inventory but the appliances are still connected with extension cord.

### Steps to reproduce

1. Make a new world
2. Spawn yourself two appliances and an extension cord
3. Place said appliances and connect them with extension cord
4. Notice that you still have the extension cord item and the appliances still connected. Pulling one of the appliances pops up a new extension cord

### Expected behavior

The extension cord item be removed after successfully connecting the appliances

### Screenshots

Screenshot_20220828-083216

- OS: Android
- Game Version: 1631976-dirty [32-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

The same thing happened with outdoor extension cord. Jumper cable might have similar issue but I'm not sure.

Hirmuolio:
#### Summary
Bugfixes "Fix power cables"

#### Purpose of change

Fix #60512

#### Describe the solution

Remove the item in the item use action.

Use `"SINGLE_USE"` flag.

#### Testing

Jumper cable is removed after use (vehicle) and works.
Extension cord is removed after use (appliance) and works.

### DIFF
--- a/src/iuse.cpp
+++ b/src/iuse.cpp
@@ -8894,7 +8894,8 @@ cata::optional<int> iuse::cable_attach( Character *p, item *it, bool, const trip
                                       source_veh->name, target_veh->name );
             }
 
-            return 1; // Let the cable be destroyed.
+            p->i_rem( it );
+            return 0; // Let the cable be destroyed.
         }
     }
 
@@ -9025,8 +9026,8 @@ cata::optional<int> iuse::cord_attach( Character *p, item *it, bool, const tripo
                 p->add_msg_if_player( m_good, _( "You link up the electric systems of the %1$s and the %2$s." ),
                                       source_veh->name, target_veh->name );
             }
-
-            return 1; // Let the cable be destroyed.
+            p->i_rem( it );
+            return 0; // Let the cable be destroyed.
         }
     }
 


irwiss:
Between #60489 and this cables issue, I think something is broken with how iuse::* return value is used

Returning 1 should already remove the item from inventory, no?

If not then the comment here needs to be updated too (well, it needs to be updated to int regardless but...), and the functions changed to void
https://github.com/CleverRaven/Cataclysm-DDA/blob/70905b58ae97ee626c70d6d9810db734af48526c/src/iuse.h#L24

Hirmuolio:
There is a more up to date comment in iuse.cpp https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/iuse.cpp#L493

Seems like that comment in iuse.h is several years out of date.

irwiss:
Ah I see, might be worth moving the up to date comment to the header then - IDEs can pick it up for hints as well.

Not sure how anyone will find a comment buried 500 lines down in the middle of unrelated methods

Issue #60525, opened 2022-08-28
Wire draw machine can't draw wire


SeamairGB:
### Is your feature request related to a problem? Please describe.

Wire draw machine is incapable of drawing steel wires, while perfectly capable of creating welding rods out of steel and copper wires. It is quite weird, considering it's what it is meant for.

### Solution you would like.

Adding wire recipe with that machine which allows for a fast way of making wires/making them in bulk without getting exhausted due to physical labour of standing near forge making single wire in 3 minutes at best.

### Additional context

https://nornagon.github.io/cdda-guide/#/item/wire_draw_machine

PR #60535, opened 2022-08-28
Adds several wire crafting recipes using wire draw machine

Marcos6226:
#### Summary
Content "Adds several wire recipes, now using the wire draw machine"

#### Purpose of change
Closes https://github.com/CleverRaven/Cataclysm-DDA/issues/60525

It's weird that we have a wire drawing machine but cannot use it to draw steel wires. 
This PR adds several steel wire recipes that use this machine to facilitate the amount of labor necessary to draw wires.

#### Describe the solution
I have added the recipes based on the copper wire one, but with the addition of necessary proficiencies and crafting materials, as well as tools, in the case of the tempered steel wire.

I'm not entirely sure if the time and charges required are perfectly balanced, but if you have any suggestions, I'm all ears.

#### Testing
I have created a new character and opened the crafting menu to check if the crafting recipes are correct, and they are working as intended, as shown in the screenshots below. 

#### Additional context
Captura de Tela (23)
Captura de Tela (24)
Captura de Tela (25)
Captura de Tela (26)
Captura de Tela (27)

### DIFF
--- a/data/json/recipes/recipe_others.json
+++ b/data/json/recipes/recipe_others.json
@@ -632,6 +632,27 @@
   },
   {
     "type": "recipe",
+    "activity_level": "LIGHT_EXERCISE",
+    "result": "wire",
+    "id_suffix": "wire_draw_machine",
+    "category": "CC_OTHER",
+    "subcategory": "CSC_OTHER_MATERIALS",
+    "skill_used": "fabrication",
+    "difficulty": 3,
+    "time": "20 m",
+    "autolearn": true,
+    "qualities": [ { "id": "ANVIL", "level": 3 }, { "id": "HAMMER", "level": 3 } ],
+    "proficiencies": [ { "proficiency": "prof_metalworking" }, { "proficiency": "prof_blacksmithing" } ],
+    "tools": [
+      [ [ "tongs", -1 ] ],
+      [ [ "hotcut", -1 ] ],
+      [ [ "forge", 40 ], [ "oxy_torch", 40 ] ],
+      [ [ "wire_draw_machine", 100 ] ]
+    ],
+    "components": [ [ [ "scrap", 1 ] ] ]
+  },
+  {
+    "type": "recipe",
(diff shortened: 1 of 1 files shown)


ANickelN:
image
Those items are nested properly as of the latest version, not sure why they are throwing an error for you.  Would recommend updating, starting a new world and seeing if it still throws those errors.

Marcos6226:
> Those items are nested properly as of the latest version, not sure why they are throwing an error for you. Would recommend updating, starting a new world and seeing if it still throws those errors.

Thanks, I'll update to latest and test if it will throw any erros.

Marcos6226:
It worked! Thanks ANickelN. ❤️

<sub> As a side note I feel a bit ashamed for not updating my game and causing this trouble...</sub>

ANickelN:
It is no trouble, this is a good addition.  I'll make a nest for it when it gets merged.

Fris0uman:
Should those recipes really require a forge and an anvil since they use the drawing machine already? Very quick look at wikipedia suggest that wire drawing is done at room temperature. https://en.wikipedia.org/wiki/Wire_drawing

Marcos6226:
> Should those recipes really require a forge and an anvil since they use the drawing machine already? Very quick look at wikipedia suggest that wire drawing is done at room temperature. https://en.wikipedia.org/wiki/Wire_drawing

The forge, hammer, and anvil are present because, as in the normal recipe, raw materials are used to make crude wires, but the wire draw machine is used to make refined wires. Normally, you'd have to manually file them to the desired thickness. By using the wire draw machine to do this refining work, we reduce the amount of labor required, as it is now a light activity, as well as the time required to complete it.

ANickelN:
Just got around to nesting the new wire recipes in with the older ones.  They will be properly nested when #60708 gets merged.

Marcos6226:
> Just got around to nesting the new wire recipes in with the older ones. They will be properly nested when #60708 gets merged.

Thank you very much. =)

PR #60553, opened 2022-08-29
Some more npc quips and fixes

NetSysFire:
#### Summary
None

#### Purpose of change

Some typo and terminal punctuation fixes, the `<fuck_you>` category of snippets is only used in contexts where terminal punctuation is already added (e.g `blah, <fuck_you>!`). It caught my eye when I was looking through the file due to its inconsistent use of terminal punctuation.

#### Describe the solution

Added some very shallow lore into the snippet category. Also replaced some usages of zombies and apocalypse to their snippet categories for variety.

1. Replacing "Mr. Cuddles" with a list of possible pet names, but this would mean that if you get lucky with the RNG or are incredibly persistent, a NPC would claim they had multiple and possibly a lot of cats, ranging from "Mr. Cuddles" to "Fluffy". This is relatively minor yet it still bugs me.
2. Adding more puns and potentially an own category for it. I am very good at coming up with bad puns but I feel like they will make humanity suffer more than the debilitating mental effects of the blob.
3. I renamed `npc_socialize` to `<npc_socialize>` and made `npc_socialize` contain `<npc_socialize>`. This is because I did not want to touch C++ and is a hack, more or less.

https://github.com/CleverRaven/Cataclysm-DDA/blob/6791af54f0fd56c03be899f4b632d498a8c79034/src/npctalk_funcs.cpp#L767-L769

But I realized this is a *very* ugly hack. Even though it works and adds even more flavor to the neutral chitchat, I do not think we need it and reverted it.

#### Testing

Applied changes locally, gave myself the debug mind control trait, forced NPCs to have chitchat with me. This is also how I found the need to use a hack and made up my mind that I do not want to use it.

#### Additional context

When I used the hack, but this still demonstrates that the other snippet category I added works properly:

image

### DIFF
--- a/data/json/snippets/npc_quips.json
+++ b/data/json/snippets/npc_quips.json
@@ -26,7 +26,7 @@
       "I am SO glad I don't need to deal with taxes or debt anymore.",
       "Geez, this is nothing like those Hollywood zombie films!",
       "Do you ever think about the ethics of killing <zombies>?",
-      "Do you think living on a farm is dangerous because because the roosters might attract <zombies>?",
+      "Do you think living on a farm is dangerous because the roosters might attract <zombies>?",
       "I have dreamt of owning a house by the sea, but now I have nightmares of what lurks beneath the waves.",
       "I liked walking through the forest, listening to the chirping of the birds but now I am afraid that the birds might devour me.",
       "So you wet your toothbrush AFTER you apply toothpaste?  That's so weird.",


data/json/snippets/npc_quips.json
```diff
@@ -1,21 +1,47 @@
...
       "So do you like cats or dogs?",
-      "It's nice to know that zeds don't feel pain.",
+      "It's nice to know that the <zombies> don't feel pain.",
       "Wherever we end up, at least we're still alive, right?",
       "Somehow, I don't think I'll be going back to my old job anytime soon…",
+      "I am SO glad I don't need to deal with taxes or debt anymore.",
       "Geez, this is nothing like those Hollywood zombie films!",
-      "Do you ever think about the ethics of killing zombies?",
+      "Do you ever think about the ethics of killing <zombies>?",
+      "Do you think living on a farm is dangerous because because the roosters might attract <zombies>?",
```
Night-Pryanik, line 29:
```suggestion
      "Do you think living on a farm is dangerous because the roosters might attract <zombies>?",
```

PR #60554, opened 2022-08-29
Cane toad no more

Karol1223:
#### Summary
Bugfixes "The TropiCata ported mutant toads were cane toads, this fixes that"

#### Purpose of change
Cane toads are not naturally found in New England, having them here made zero sense. Having said that, since they were already mutants that copied from the mutant frogs, all I did was change the descriptions to no longer reference the Cane Toad.

#### Describe the solution
- Changes the description of the Huge Toad and the Giant Toad to no longer reference Cane Toads
- Changes the time it takes for Huge Toads to grow into Giant Toads to be the same as it takes for the same process to happen in mutant frogs
- The Giant Toad no longer upgrades into the **_zombie_** Gastro Bufo

I did initially want to backport these two back to TropiCata but after a bit, I realised it's far easier to just make them no longer Cane Toads

#### Testing
It is just a bunch of one-line changes, I don't expect the game to implode.

### DIFF
--- a/data/json/monsters/reptile_amphibian.json
+++ b/data/json/monsters/reptile_amphibian.json
@@ -374,7 +374,7 @@
     "id": "mon_toad_giant",
     "type": "MONSTER",
     "name": { "str": "giant toad" },
-    "description": "This cane toad has grown bigger than a human.  It looks uninterested in eating you but you better not give it a chance.",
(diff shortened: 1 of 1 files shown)


a-chancey:
I’ll wait for this to be merged before fixing #60565 to avoid conflicts

PR #60555, opened 2022-08-29
Adding freezer spawns

a-chancey:
Adds freezer spawns to a handful of house basements, the hunting lodge, furniture store, the home improvement superstore and a house garage.

Also will partially implement #56401 by adding to butcher's shops and LMOE Shelters. Restaurants usually use Walk In Freezers, not chest freezers, so it doesn't make sense there.

#### Summary
Content "Adding additional freezer spawn locations"

#### Purpose of change
This simply adds a few locations for the chest freezer furniture to spawn.

#### Describe the solution
I added spawns for the chest freezer in the Hunting Lodge, furniture store, home improvement superstore, a few basements and a house garage variant.

#### Testing
Loaded up my test world and found/spawned each of the locations to make sure the freezer shows up properly and doesn't throw errors.


Golfavel:
This should probably also close https://github.com/CleverRaven/Cataclysm-DDA/pull/56401 once finished, it hasn't had any activity since March.

a-chancey:
Somewhat ironically, it doesn’t add to any of those locations. When I figure out this mod issue, I may add to a few of them. LMOE absolutely makes sense for a chest freezer

a-chancey:
Whenever I figure out which mod is causing this test to fail (as it seems something is overwriting palettes), I'll get freezers added to the other locations. 

I didn't add any item groups to spawn in the freezers yet.

Maleclypse:
> Whenever I figure out which mod is causing this test to fail (as it seems something is overwriting palettes), I'll get freezers added to the other locations.
> 
> I didn't add any item groups to spawn in the freezers yet.

Did you see where I responded on discord this morning.  It's not a mod issue.  it's a palette issue.

data/json/mapgen/homeimprovement_superstore_new.json
```diff
@@ -314,6 +314,7 @@
       "s": "f_sink",
       "b": "f_bathtub",
       "(": "f_shower",
+      "é": "f_freezer",
```
Maleclypse, line 317: You have to give it a terrain entry as well.   Otherwise the game thinks you are loading furniture over a void.

a-chancey:
Kid woke up early and I couldn't sleep, so I got to work heh.
Found the issue. I did miss the terrain, but it was getting filled in with fill_ter. There was a reason it only failed on the all-mods world test runs:
No Hope was overwriting the home_improvement_store palette, but has a separate version of the store - so the vanilla store with the freezer was being checked but the palette from no hope was used.
So i added it to the mod, and the house garage that i added it to in vanilla as well.

Issue #60452, opened 2022-08-26
Problems with proficiency learning during very slow crafts


actually-a-cat:
### Describe the bug

There are several problems with proficiency training during very slow crafts, e.g. plate armor:

* Proficiency training is accrued only every 5% of craft progress. For slow crafts, this can be many hours of crafting over many sessions. This is confusing (why am I not getting any proficiency progress?) and can cost a lot of time (you should have enough progress in the proficiency to be benefitting from it, but you're still at 0% and crafting at initial speed). See `craft_activity_actor::do_turn`.
* For every 5% increment, proficiency training is capped at 1000 seconds. If the craft is taking more than 16.6 minutes per 5%, you're losing out on progress. If the craft is taking many times more than that, you're getting effectively no progress at all. See `Chracter::adjust_for_focus`
* Focus is checked only at the 5% point. If it takes you days to get that 5%, your focus over that time doesn't matter at all, only at the instant you cross the 5% (or 10%, 15%, etc.) mark.

### Steps to reproduce

This was determined through looking at code and debugging but the effects are visible whenever you're crafting something very slow and checking your proficiency progress during.

### Expected behavior

Proficiency training should be continuous and give the same effects per time spent regardless of how long the craft is.

### Versions and configuration

By the way, the game report doesn't work on curses. It says written to debug.log, but doesn't. d2a84e3 I guess.

PR #60564, opened 2022-08-29
Fix and de-RNG skill gain from reading, uncap proficiency gain

actually-a-cat:
#### Summary
Bugfixes "Fix slow crafts not giving enough proficiency XP"

#### Purpose of change
These are two changes that I'm bundling because they overlap enough that separating them would be an issue.

First is #60452, specifically point 2. The cap now doesn't apply to proficiency training.

Second, learning from books is weird. The calculations are performed on low-value integers, repeatedly losing precision to rounding errors, resulting in various factors having reduced or no effect on XP gain. For example, when reading "chemical reference-CLASSIFIED" at 6 applied science, there is no difference in XP gain between 1 and 50 effective focus, then between 51 and 100, etc. The exact values depend on book and level. Effect of intelligence is also very "chunky" when it works at all. It also has an unnecessary RNG component to it.

#### Describe the solution

`Character::adjust_for_focus` takes and returns floats, normalizes its result (divides by 100 after multiplying by focus), and the cap has been moved to the one call site that needs it. 
Since the function is used to adjust various things - seconds, practice ticks (also seconds, kinda), minutes, arbitrary experience points - it can't assume it's okay to round anything, and the cap has unknown meaning when applied there.

XP calculations for reading are now done in floats, and the RNG roll has been removed. The new value is set at the midpoint between the old minimum and maximum roll so the expected value should be the same (rounding errors aside)

TODO: examine if the clamping there is necessary and what the values should be

Going straight for a more complete fix to #60452, but I think that's beyond me right now.

#### Testing
TODO: write test to back up above claim that book reading balance is roughly the same, and a regression test for proficiency

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -1351,26 +1351,16 @@ void read_activity_actor::read_book( Character &learner,
     // Calculate experience gained
     /** @EFFECT_INT increases reading comprehension */
     // Enhanced Memory Banks modestly boosts experience
-    int min_ex = std::max( 1, to_minutes<int>( islotbook->time ) / 10 + learner.get_int() / 4 );
-    int max_ex = to_minutes<int>( islotbook->time ) / 5 + learner.get_int() / 2 - originalSkillLevel;
+    float exp = to_minutes<int>( islotbook->time ) / 7.5f + learner.get_int() / 3.0f;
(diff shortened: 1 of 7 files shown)


actually-a-cat:
I'm dropping the book stuff, it ended up being more confusing than I thought, the proficiency bit I'm continuing in #60711.

Issue #47630, opened 2021-02-20
The state of volume of bundles (rag, leather)


Lamandus:
To be honest, getting rid of the bundles all together would be best. It doesn't make sense that those bundles have less volume then the original rags, especially that high reduction. We are making those rags more dense, can we use it as a heat source, by the energy we are creating?

Just to put it into perspective we are reducing 10 rags á 250 ml (2500 ml) to 750 ml by what force exactly? I mean those bundles still can have a purpose (easier handling, faster transfer from A to B) but such a reduction of volume? I mean we are already saying, we are using the room in our pockets as efficient as possible. Wouldn't make it much "easier" to leave the rags as is? Since it is easier to store 100 single tissues into my backpack than a "block" of it.

actual-nh:
I have been reluctantly forced to learn over the years that putting clothing neatly into a bag fits more of it than if one simply stuffs it (although stuffing is frequently advisable to fill in the cracks). To produce that much compression of rags, I suggest a recipe involving a sealable plastic bag and a straw (to withdraw the rest of the air).

wapcaplet:
See #24175 (the PR that added rag bundles) for rationale.

aerrant:
Rag bundles aren't very necessary, but plank bundles are life savers - it'd be a chore to drag those by the floor otherwise.

If anything, I'd have liked bundles to be automatically disassembled whenever a recipe requires their contents.

The IRL analogy is solid, too: you really can make clothing more compact by organizing it or by using vacuum bags.

Lamandus:
but then we need to see, what the volume is actually about.

is it the actual volume of the object
or is it the volume the item takes away from storing somewhere, making the shown volume bigger than the object itself.

Another take can be, to make those stacks of rags with bigger packages, for example 100 per bundle, so the ground hauling is much much faster. But I don't buy the extreme benefit on volume to be honest. From 2500 to 2000 maaaaybe, but that is less than half the volume.

Fosheze:
As far as the hauling goes bundles are really no longer needed for that for small items now that nested containers are a thing. If you want to haul 500 rags just stick them all in a duffel bag and then you only have 1 item to move. You can also use them directly out of the bag that way. I can still see the point of having bundles for large items like planks because there is no better way to haul them but for small items no one is going to realistically make multiple bundles of exactly 10 rags rather than just stuffing them all in a bag. I've personally been using body bags for this for a while now. 

[...]

PR #60567, opened 2022-08-29
fix fabric bundle weight and volume

nornagon:
#### Summary
Bugfixes "Fixed textile bundle sizes to be 100 times the individual items"

#### Purpose of change

Weight and volume for fabric bundles were wildly wrong.

Ref https://github.com/CleverRaven/Cataclysm-DDA/issues/47630

#### Describe the solution

This fixes the volume and weight for fabric bundles, and also ups the size of the bundles to be a larger number of items instead of 10. The idea is that 10 items is not that hard to deal with.

- bundle_rag renamed to "bundle of cotton patches", size is now 100 instead of 10
- bundle_leather size is now 50 instead of 10
- bundle_wool size is now 100 instead of 10
- introduced bundle_sheet "bundle of cotton sheets" at a size of 25

these numbers are chosen to roughly keep the bundle size around 0.5 L - 1 L in volume.

Could leave the bundles at size 10. Maybe this would be better for savegame compatibility?

### DIFF
--- a/data/json/recipes/recipe_others.json
+++ b/data/json/recipes/recipe_others.json
@@ -2338,7 +2338,7 @@
     "autolearn": true,
     "using": [ [ "filament", 16 ] ],
     "qualities": [ { "id": "SEW", "level": 1 } ],
-    "components": [ [ [ "cotton_sheet", 25 ] ] ],
+    "components": [ [ [ "sheet_cotton", 25 ] ] ],
     "flags": [ "BLIND_HARD" ]
   },
   {


ANickelN:
Would you add the new "bundle of cotton sheets" to the bundle_of_items group located in nested.json?

nornagon:
brought leather bundle size down to 50, to keep bundle volume around 500 g - 1 L for all fabrics

nornagon:
The density check here is failing, but i think it would also fail for leather?

https://github.com/CleverRaven/Cataclysm-DDA/blob/b5e0cb58afcfea8ea3f86fbba4853a12b12aad59/data/json/items/resources/tailoring.json#L161-L162

this would be a density of 1.6 g/ml for leather patch, which is above the 1.1 density for the leather material.

why is this test failing for the bundle but not for the individual patch?

nornagon:
ah: https://github.com/CleverRaven/Cataclysm-DDA/blob/b5e0cb58afcfea8ea3f86fbba4853a12b12aad59/data/mods/TEST_DATA/known_bad_density.json#L1077

i'll add the bundle to that list then, since it's just a flat multiplier on the underlying item.

Maleclypse:
> The density check here is failing, but i think it would also fail for leather?
> 
> https://github.com/CleverRaven/Cataclysm-DDA/blob/b5e0cb58afcfea8ea3f86fbba4853a12b12aad59/data/json/items/resources/tailoring.json#L161-L162
> 
> this would be a density of 1.6 g/ml for leather patch, which is above the 1.1 density for the leather material.
> 
> why is this test failing for the bundle but not for the individual patch?

Probably means we know that the patch is wrong but no one has fixed it yet because there is a long list of wrong densities so any item that copies from the original patch won't benefit from that wrongness.

data/mods/TEST_DATA/known_bad_density.json
```diff
@@ -1075,6 +1075,7 @@
       "test_waterproof_bag",
       "fake_goggles",
       "leather",
+      "bundle_leather",
```
anothersimulacrum, line 1078: Please don't add to this list. We'll never get through it if we allow that.

nornagon:
How should I resolve this, then? I could edit either the weight or the volume of the leather patch to match the density of the leather material. I could edit the density of leather. There are sources which claim the density of leather as being anything from [0.2–0.3](https://www.researchgate.net/figure/density-of-each-leather-type-at-ambient-conditions_tbl1_328549546) to [0.86](https://www.engineeringtoolbox.com/density-solids-d_1265.html) to [1.5+](https://nvlpubs.nist.gov/nistpubs/jres/31/jresv31n3p169_A1b.pdf) g/ml in some cases.

Which source are we choosing here?

I'd prefer to avoid this PR ballooning into a reevaluation of the density of all leather items.

Maleclypse:
I would suggest this "adjust leather patch to match the density of the leather material." path

nornagon:
Okay! Sounds reasonable. There are two options then, for a leather density of 1.1 g/ml:

1. keep weight at 37 g, adjust volume to 34 ml
2. keep volume at 23 ml, adjust weight to 25 g

does one of these options strike you as better than the others?

Maleclypse:
I think option 2 makes the most sense as most of our tailoring recipes are balanced towards volume rather than weight

nornagon:
alrighty, done. I also upped the volume of leather_bundle a bit to fit within the max density. The other bundles allow for a bit of compression (i.e. you're cinching the thread tight to increase the density of the fabric) but for leather I figure it's not very compressible, so I stuck to flat 50x volume, 50x weight for the leather bundle.

PR #59682, opened 2022-07-27
Steel crafting improvements

Drew4484:
#### Summary
Content "Improvements to carbon steel crafting and repairs"

#### Purpose of change
Allows basecamps to craft various grades of carbon steel lumps, chunks, wire, and link sheets.
Adds welding blankets, used to craft a welding kit. Welding kit allows for repairs to tempered steels. 
Adds a small increase to the cut/bash/ballistic values of the carbon steels (1/1/1.5). 

#### Describe the solution
- [x] Add recipes to basecamp craft lists.
- [x] Add asbestos
- [x] Add welding blanket
- [x] Add welding kit
- [x] Remove tempered repairs from arc welder and makeshift arc welder.
- [x] Add tempered repairs to new welding kit.

#### Describe alternatives you've considered
None

PR #60582, opened 2022-08-30
Modular Bsaecamp Workshop recipe update

Sathra225:
quality steel and steel products for modular workshop

#### Summary
Content "Update modular basecamp workshop 2 recipes"

#### Purpose of change
#59682 added the different steel quality lumps and wire and such to the basecamps, but the modular basecamp workshop 2 uses its own recipe set that was missed. Adding the recipes to that list as well (with roughly the same progression but the modular isn't 1 to 1).

#### Describe the solution
Copy-pasting the new recipe fields into the modular basecamp workshop 2 recipe groups.

None.

#### Testing
Made changes, loaded game with a modular basecamp workshop 2 and verified that the recipes were available.

### DIFF
--- a/data/json/recipes/basecamps/expansion/recipe_modular_workshop/version_2/modular_workshop_recipe_groups.json
+++ b/data/json/recipes/basecamps/expansion/recipe_modular_workshop/version_2/modular_workshop_recipe_groups.json
@@ -25,9 +25,11 @@
       { "id": "scrap_copper", "description": " Craft: Copper, Scrap" },
       { "id": "charcoal_npc", "description": " Craft: Charcoal" },
       { "id": "spike", "description": " Craft: Spike" },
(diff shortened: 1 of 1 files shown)


Sathra225:
Huh, the modular version 2 didn't have glass caltrops so those those got added too.

PatrikLundell:
The reason the workshop 2 recipe lists was kept short is that the UI works very poorly when the lists are bloated. The glass caltrops were not introduced since I didn't think they'd be sufficiently useful for companion construction to be worth a place on the list. However, I've been overruled on the bloat issue in the past, and so expects to be so again (The argument was approximately that nobody will fix the bloat issue until it becomes painful. My counter argument is that it's been painful for a fair while, but nobody has overhauled the UI yet despite that).

Sathra225:
True, though this is largely so the recipe lists between the two 'versions' of basecamp workshops are the same.
Or close to the same.

PR #60584, opened 2022-08-30
virtually remove encumbrance from dustmask

anoobindisguise:
#### Summary
Balance "dust mask doesn't encumber"

#### Purpose of change
Dust masks were pretty pointless as they fairly seriously restricted your breathing (as much as half of a filtering-type mask like Firefighter PBA)
Per source: https://www.ncbi.nlm.nih.gov/pmc/articles/PMC8535201/ found that surgical masks do not meaningfully reduce oxygen intake (among young people). Dust mask isn't exactly a surgical mask as those are often a kind of plastic material, but this is the closest equivalent in game atm.

#### Describe the solution
Encumbrance unfitted 10->1

Differentiate "dust mask" (basic cotton mask), "n95 mask", "plastic mask" (the kind you see littering the parking lots and sidewalks on occasion) and/or introduce more like "nomex mask" etc.

#### Additional context
Is COVID canon in cdda?

### DIFF
--- a/data/json/items/armor/masks.json
+++ b/data/json/items/armor/masks.json
@@ -167,7 +167,7 @@
     "material_thickness": 0.1,
     "environmental_protection": 3,
     "flags": [ "VARSIZE", "SKINTIGHT" ],
-    "armor": [ { "encumbrance": 10, "coverage": 100, "covers": [ "mouth" ] } ],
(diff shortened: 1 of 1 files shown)


Marcos6226:
> Is COVID canon in cdda?

I'm more worried about the blob possibly being canon in real life.

Issue #59166, opened 2022-07-11
Allow unplugging appliance


Qrox:
### Is your feature request related to a problem? Please describe.

Sometimes you may want to disconnect an appliance and connect it to a different grid, or may want to recover an extension cord and use it elsewhere, but currently you can only unplug an appliance by dragging it around which sometimes you cannot do because the appliance is blocked by other appliances/vehicles.

### Solution you would like.

An option to remove a particular connection from an appliance, like how jumper cable is removed from a vehicle

### Describe alternatives you have considered.

None

esotericist:
note that you currently also cannot disconnect jumper cables from appliances that you attached from a vehicle. of course directly connecting jumper cables to arbitrary appliances might not be appropriate, see #59169

PR #60586, opened 2022-08-30
Appliance: keep cable connected when dragging appliances around (within max cable length)

### LINKED ISSUE #59167: Extension cord dropped when dragging appliance around
### Describe the bug

When you connect two appliances using an extension cord, the extension cord is dropped when you drag one of the appliances around, even if the distance between the two appliances is less than the maximum length of the extension cord.

### Steps to reproduce

1. Put up twp appliances
2. Connect the appliances using an extension cord
3. Drag one of the appliances
4. See that the extension cord is dropped to the ground

### Expected behavior

Extension cord not dropped to the ground unless the distance between appliances exceeds the maximum length of the extension cord. And a message indicating the extension cord is dropped in the latter case.

### Screenshots

_No response_

### Additional context

_No response_

dseguin:
#### Summary
None

#### Purpose of change
- Fixes #59167

#### Describe the solution
Keep the power cable connected when moving appliances (and vehicles) around within the maximum cable length. If a connected appliance/vehicle moves beyond the cable length, the cable is detached and falls to the ground.

Note: cable length is determined by the item's `"max_charges"`. So the length of a "power cord" is 3 squares, an "extension cable" is 10 squares, and a "long extension cable" is 30 squares.

A few implementation details:
- When moving an appliance/vehicle, check for power connections and update the target position at both ends of the cable
- The cable connection is preserved/updated when calling `vehicle::shed_loose_parts` with source and destination positions, otherwise it defaults to the previous behaviour
- Updated some uses of `tripoint` to `tripoint_bub_ms` in `map::displace_vehicle`

#### Testing

Added a unit test for checking cable detachment when moving an appliance:

```
./tests/cata_test --rng-seed time "power_cable_stretch_disconnect"
```

##### In-game testing

https://user-images.githubusercontent.com/12537966/187387691-7b939260-8480-40fd-a80a-ab28825f621b.mp4


Fris0uman:
Won't this make https://github.com/CleverRaven/Cataclysm-DDA/issues/59166 worse?

dseguin:
> Won't this make https://github.com/CleverRaven/Cataclysm-DDA/issues/59166 worse?

You're right, I'll leave this in draft for now until that gets done.

tests/vehicle_test.cpp
```diff
@@ -266,3 +273,152 @@ TEST_CASE( "Unfolding vehicle parts and testing degradation", "[item][degradatio
...
+    target_part.target.second = source_veh->global_square_location().raw();
+    target_veh->install_part( vcoords, target_part );
+}
+
+TEST_CASE( "power_cable_stretch_disconnect" )
+{
+    clear_map();
+    clear_avatar();
+    map &m = get_map();
+    m.load( { 0, 0, 0 }, false );
```
andrei8l, line 325: The test should be generalized to work at any map coordinates (if it doesn't already). Or you can use `game::place_player_overmap()` if you absolutely need this anchored at zero.

The load on its own breaks some game-wide assumptions, but they may not have been valid in the first place (as seen in #60674).

dseguin: I see, cheers! I didn't think too much of it, but after looking at it further it did expose a bug. Should be fixed in #60714

PR #60587, opened 2022-08-30
Fix quickloading

### LINKED ISSUE #38557: Unnecessary autosave after quickloading an autosave
# Describe the bug
The game performs an autosave 1 turn after quickloading an autosave

# Steps To Reproduce
1. Set 'Real minutes between autosaves' to 0 and 'Game turns between autosaves' to 10
2. Skip turns until autosave occurs
3. Skip 1 turn to enable quickloading
4. Quickload
5. Skip 1 turn. The game makes another autosave

# Expected behavior
Autosave after skipping 10 turns, not 1

# Versions and configuration
- OS: Windows
    - OS Version: 10.0 1809
- Game Version: 0.D-12808-g5d392c0d6b-dirty [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

# Additional context
When loading an autosave after 'crashing' the game and skipping 1 turn, the game does not autosave - as intended.
This bug makes it really annoying to investigate another autosave-related bug...

angleKH:
#### Summary
Bugfixes "Restored quickload to working order"

#### Purpose of change
This PR fixes the quickload function (currently crashes) and enables it in release builds under the debug menu, and adds the same warning as the "Quit without saving" option. This would also be especially helpful for development.

This also happens to fix #38557.

#### Describe the solution
Rather than attempting to call the functions to load the game directly, as the current implementation does, this simply loads the world as if the user exited to the menu and immediately loaded the world again. This means the cleanup code from exiting will also run, unlike the current implementation which is missing some.

For this PR, on quickloading, the current world and player save is queued to load the next time the main menu is entered. Then the game exits to the main menu but loads the world immediately. Because of this the data files still need to go through the loading/validating process, but this is also the case for the current implementation.

The cli option `--world` has also been changed to use this queue rather than being part of the main loop as an if statement.

Loading the world again without simulating exiting and loading the world from nothing would be best, but is infeasible. I could not get it to work. This would also likely add a lot of duplicated code from the necessary cleanup.

Before using the main menu, I also tried using the current implementation of the cli `--world` option's method to load the world (`g->load` in the main loop before menu) after running cleanup code. This worked, but segfaulted after a couple of quickloads. This also implies that there may be some problem with the current implementation of the cli `--world` option.

#### Testing
Tested only on Linux.
- Tested quickloading several times.
- Tested with the same world.
- Tested after switching between different worlds.
- Tested with worlds containing multiple characters to ensure the correct character was quickloaded.
- Tested quickloading after loading a world with the cli `--world` option.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -12066,10 +12066,10 @@ void game::quicksave()
 void game::quickload()
 {
     const WORLD *active_world = world_generator->active_world;
-    std::string const world_name = active_world->world_name;
     if( active_world == nullptr ) {
         return;
     }
+    std::string const world_name = active_world->world_name;
     std::string const &save_id = u.get_save_id();
 
     if( active_world->save_exists( save_t::from_save_id( save_id ) ) ) {


src/debug_menu.cpp
```diff
@@ -345,6 +346,7 @@ static int game_uilist()
         { uilist_entry( debug_menu_index::CRASH_GAME, true, 'C', _( "Crash game (test crash handling)" ) ) },
         { uilist_entry( debug_menu_index::ACTIVATE_EOC, true, 'E', _( "Activate EOC" ) ) },
         { uilist_entry( debug_menu_index::QUIT_NOSAVE, true, 'Q', _( "Quit to main menu" ) )  },
+        { uilist_entry( debug_menu_index::QUICKLOAD, true, 'Q', _( "Quickload" ) )  },
```
Night-Pryanik, line 349: The same keybinding as the `Quit to main menu` entry above.

dseguin (approved):

src/game.cpp
```diff
@@ -12067,21 +12066,25 @@ void game::quicksave()
 void game::quickload()
 {
     const WORLD *active_world = world_generator->active_world;
+    std::string const world_name = active_world->world_name;
```
BevapDin, line 12069: This will trigger UB, as apparently `active_world` can be null. Note the check for null is *after* it gets dereferenced here. You need to move this line after the if clause.

angleKH: Great catch! I can't believe I missed that with the explicit nullptr check.

PR #60592, opened 2022-08-30
Add Recipes of Nomad Plate Armor

nameless-feng:
Summary
Content "nomad plate armor blueprint"

Purpose of change
Provides players with recipe for nomad plate armors,players can get it at Rubik

Describe the solution
Added them to the JSON .

Describe alternatives you've considered
Testing
I tested this in mod form

Additional context


anoobindisguise:
Personally I was fine with making this Cody-only since it lets us abstract some of the elements of its creation. If the player can make it themselves, you have to work a lot harder to justify its process.

nameless-feng:
> Personally I was fine with making this Cody-only since it lets us abstract some of the elements of its creation. If the player can make it themselves, you have to work a lot harder to justify its process.


Can you give me some advice on this issue?

bombasticSlacks:
Aside from the fact that your files are a mess, this is generally a no. I don't think this should be merged even if you fix your messy code. The EXODII wouldn't have a recipe lying around for this specific item, it's supposed to be custom made for the player. 

The best a player can do is just wear the pieces separately, the actual combined suit fully optimized for the player should stay NPC craft only.

Fris0uman:
Closing according to feedback from bombasticSlacks

PR #60202, opened 2022-08-16
Cardio overhaul; divorce BMR from cardio and fix health, athletics, and traits effects on cardio

ArgonMercury:
#### Summary
Balance "Divorces BMR from cardio. Fixes health, athletics, and traits effects on cardio"

#### Purpose of change

- Cardiofit trended towards kcal spent per day, inevitably ignoring all other variables.
- BMR had a massive effect on cardio, making things like fast metabolism, being overweight, being tall give you absurd amounts of stamina.
- Morbidly obese characters gained a butt ton of stamina, and the health modifier in cardio fit was functionally irrelevant.
- Drastic changes in BMR had ridiculous effects, a default character going tiny increased their stamina regen from 20 to 50.
- Health, athletics, and traits (indefatigable and hyperactive) had no effect on cardio in the long run, due how the cardio accumulator was updated.

#### Describe the solution

Fixes all problems listed above by;

Refactored how cardio fitness and the cardio accumulator works. Cardio accumulator is now a number representing the ratio of your average activity (kcal burned each day/BMR). Cardio accumulator starts at 1000 (1/1 ratio) and goes up to 3000 max (3/1 ratio of kcal burned a day to bmr). Health, athletics, and traits are multipliers of cardio a that stack additively, increasing (or decreasing, if you have negative health) your cardio fitness. BMR, and things that affect it (height, age, obesity, etc) have no effect on stamina any more.

- Health -200 to 200 -> -40% cardio to +20% cardio
- Athletics 0 to 10 -> +0% cardio to +10% cardio
- Indefatigable and hyperactive work as intended (+30% and +60% cardio)

I changed these from additive bonuses to multipliers because it seemed to make more sense to me. Technical cardio skill via athletics or a proficiency would be proportionally beneficial to both someone with high physical endurance and someone with low physical endurance. Traits have always been multipliers anyways. Health being a multiplier helps it maintain relevance (especially as a penalty for messing up your health stat), but I could see it as an additive effect, though it would be the odd one out at that point.

#### Describe alternatives you've considered

Changing around the numbers a bit. Indefatigable and hyperactive seem rather high powered now that they don't get eaten by the cardio accumulator over time.

#### Testing

Compiled and ran game, checked to make sure everything (cardio accumulator, health, athletics, and traits) had the expected effect on both stamina regen and max stamina. Re-configured the cardio test and made sure it produced sensible results.

#### Additional context

- I tried to keep cardio accumulator to about the same value as before to prevent messing up current characters.
[...]

PR #58380, opened 2022-06-12
Max stamina is now reduced based on player's overweight level

Night-Pryanik:
#### Summary
Balance "Max stamina is now reduced based on player's overweight level"

#### Purpose of change
Contribute to #53922.
People with overweight should really have problems with stamina.

#### Describe the solution
- Effects of max stamina reduction now begins when character is halfway to being obese. On this level stamina reduction is 12%, upon reaching morbidly obese level the reduction is almost 40%.

#### Describe alternatives you've considered
None.

#### Testing
Checked max stamina of character with normal weight, obese and morbidly obese.

#### Additional context
изображение

PR #60597, opened 2022-08-30
Fix BMI system and add penalties for obesity

anoobindisguise:
#### Summary
Features "Dynamic BMI penalties"

#### Purpose of change
There were no real consequences for being fat and it was often viewed as beneficial. Originally, I wanted to add a system of penalties for this and I based those penalties off the player's BMI as that is the only meaningful measure of the player's overweightedness but as it turns out, BMI itself is completely broken and so I comprehensively reworked that system also.

#### Describe the solution
BMI used to not work. Originally, all characters were made with 55,000 kcal which is their "healthy" amount, irrespective of their height. BMI was determined by 13 + ( current stored kcal / healthy stored kcal ), and in turn your actual weight was determined by BMI * (height in meters)^2. 

This was ridiculously wrong. Take a 200cm tall character who eats 3500 kcal at their healthy weight, who now has 58500 stored kcal and is 1.064x their healthy kcal now, so their BMI is now 25.76. This means their bodyweight has increased from (25 * 4 = 100kg) to (25.76 * 4 = 103kg). 

There are countless other ways this backwards system just produced entirely nonsensical numbers, and also ignored your strength entirely for factoring your BMI. Even for a 175cm standard character being 3x your stored healthy kcal (which should be only be an increase of 110,000 kcal, 31 lbs) means you were now BMI 50 and weighed twice as much.

So I have entirely re-done BMI. Your BMI is now broken up into three parts: "base", "muscle" and "fat".
- Base BMI is always 12. For an average person (BMI 25) roughly 40% of your bodymass is neither fat nor muscle, such as bone, brain, blood, etc, then I add 2 more for the remaining vital muscle/fat tissue you cannot lose without dying. Your base weight before fat is determined by your "Base BMI" and "Muscle BMI" added together and multiplied by your height^2 as per normal BMI calculations. So for an avg 8 str character they start off at a "true" bmi of 25 (12+8+5), while a 12 str char starts at "29" and a 4 str character starts at "21" (note that they all have 5 fat BMIs which is on the border between normal and overweight; this is what actually matters for BMI-related health, their "true" bmi is used to determine weight and BMR)

- Muscle BMI is equal to your strength. This means that a 12 strength character now actually weighs more than a 4 strength character of the same height. 
- Also, if you are underweight or starving (stored kcal is below 60% of your "healthy" amount) your strength will trend downwards to 0  (debuffing the stat by a percentage rather than a flat amount like dex/int). I'm not simulating permanent muscle damage this way, though it's something that could happen in the future. This also affects the BMI formula by the same amount, causing your "true" bmi at point of death to be 12 instead of just 18-20 if only your "fat bmi" went down. You are at BMI 13 a moment before death which is roughly the lowest someone can go IRL without dying (at least for men; women can sometimes survive going a little lower)
- BMI from fat is now determined by the number of kilograms of fat ( stored kcal / (3500*2.204) ) so 20 kilograms of fat factors differently to a 200cm individual than a 145cm individual and is an actual real number - eating net 3500 kcal means you are going to weigh exactly 1 pound more no matter how tall you are. This uses the BMI formula, where (stored kcal converted to kg) = x * (height in meters) squared, where x is your "fat BMI" level.
- Your "healthy' stored kcal is equal to 20% of your hypothetical bodymass at BMI 25, and in addition to being set to this value as the game starts (175cm character has about 118,000 stored kcal, 200cm character has about 150,000 stored kcal) it can shift dynamically during the game by for example becoming Huge causing you to be severely underweight due to your new height causing the static stored kcal to become dwarfed by comparison.
- This gives "Fat BMIs" a standard number of 5, and almost every effect in the game that cared about BMI has been redefined to care about "Fat BMI" instead of "True BMI", with for example 10 "Fat BMI" marking the start of obesity. Underweight starts at <3.0 BMI and starvation begins at <2BMI.  Even though having 16+ strength might make you clinically obese, the game only cares about your "Fat BMIs" and completely ignores the 14+ BMIs coming from bone/muscle for every calculation except your character's weight.
- While this may seem like it gives you less buffer for starvation resistance it actually overall results in a bigger window, as the previous number of 55,000 represented less than 10% bodyfat of an average 175cm character and the more realistic number leads to an actual longer time to starve to death.
- Kcal healthiness has been reworked. Now instead you lose 25 points of "max healthy" for each fat BMI above 5 you are, so you start becoming forcibly unhealthy partially through obesity and hit minimum health -200 just after morbid obesity. You lose 200 points of "max healthy" for each point of fat BMI below 3.0 you are, so you become actively unhealthy just below starvation BMI and bottom out at -200 healthiness at point of Skeletal (<bmi 1). This was necessary to fix the system being centered around a BMI of 5.

These changes broke a lot of tests so I had to fix those too, as they were still expecting BMI to work like it used to.

There are now a bevvy of penalties for obese characters.
- Each limb has a JSON definition in body_parts.json for the following. This lets the penalties be easily adjustable and be dynamic per limb (fat builds up the most on the torso, much less so on the hands, and not ever on the eyes - hopefully) and can be set for mutant limbs in JSON too.
    - bmi_encumbrance_threshold. The BMI at which point encumbrance penalties start applying,
    - bmi_encumbrance_scalar. The multiplier of the difference between your "fat BMI" and the threshold, applies that much encumbrance. So if your threshold is 10 and your "fat BMI" is 15 and the scalar is 0.6 you'd get 3 encumbrance.
    - ~~bmi_rigid_problems_threshold. If BMI exceeds this much wearing rigid armor on the limb is uncomfortable, applying a separate -3 speed penalty and occasional pain per limb.~~
- ~~Character's bodyfat weight (excess weight above fat BMI 5) is applied as carryweight to the player at a 2:1 ratio. So if you are 100lbs overweight, you are permanently carrying 50lbs extra in your inventory. This shows directly as weight carried in your tracker which does allow you to track your weight in real time. This isn't totally ideal but generally speaking you should have an estimate of how burdensome your bodyfat is and it's the only way to provide feedback for the player on how their weight is affecting their carry capacity. This means if you exceed 5 BMI and become Overweight you will be carrying a few extra pounds, but it should be inconsequential for Overweight levels.~~
- The above now only silently applies to movement, it does not result in numbers in your inventory. Your walking will become much more strenuous, but it will not apply Overburdened speed penalties and it also won't tell you exactly how many penalty pounds are being applied.
- Being fat applies extra warmth to your character. This occurs a little over halfway through Overweight. At Obese (~10 fat BMI) you have roughly 3-4 extra warmth on each part, it is not specifically per limb because homeostasis should be roughly evenly distributing your total excess body heat. At Morbidly Obese (~20 fat BMI) you have about 15 extra warmth. this is a sq. root function so the added warmth levels off. It isn't an excessive amount of heat in most cases which should hopefully be true to life - you're not going to walrus your way through the winter, you just might need one less layer and are mildly more uncomfortable in the summer.
- ~~As described earlier wearing rigid armor is uncomfortable if you are too fat. No more fitting into plate armor until you diet! Some rigid items with low per-limb coverage like knee pads don't apply this penalty and soft items also won't become uncomfortable (currently)~~
- ~~If BMI exceeds 50, you cannot traverse windows or pilot mechs.~~

- Make NPCs comment on player's weight.
- Different penalties per limb - these are JSON so they can be easily adjusted in the future.
- The huge/small mutations should probably be adjusted at some point in the future. Growing taller should probably involve you getting REALLY hungry and growing smaller should probably involve your body wastefully burning off a lot of excess stored calories.

#### Testing
- Spawned 3 different characters, a 145 cm char, a 175 cm char and a 200cm char. Confirmed they had healthy stored kcal to start off with appropriate to their height (~80k, ~115k, ~150k). Then repeated the process for these characters with the XS trait and the XXL trait and confirmed they had 20% and 500% of the expected stored kcal, respectively, placing them at barely skeletal and barely morbidly obese. 
- For each of the three characters I tried increasing stored kcal by 3500 and saw them each gain 1 pound on a scale, and reduced it by 3500 and saw them each lose 1 pound on a scale. 
- Increased strength by 5 on each character and saw bodyweight increase by 5 BMIs' worth of weight in each instance.
- Added the LARGE trait and saw myself become Underweight from a previously healthy weight.
- Adjusted stored kcals to various "starvation" values and observed strength trend down to 0 no matter how much base strength I had, and also saw "true" bmi reflect those changes. 
- Gave myself the strength enhancer CBM and confirmed it didn't affect my BMI.
- Confirmed widgets are accurately displaying my "Fat BMIs" and not my "true BMI". Increased strength to 99 and didn't become morbidly obese.
- weighed myself on a livestock scale at various heights and various amounts of strength and stored kcal and saw that my weight on the scale was always equal to (True BMI) * (height in meters)^2. 
-----------------------------------------------------------------------------------------------------------------------
- Adjusted stored kcal values to various levels of obesity and observed expected encumbrance penalties. Set stored kcal to underweight or emaciated levels and didn't see me gaining "negative" encumbrance.  
- Adjust kcal values to various obesity levels and observed temperature changes acting as expected.
- Dug holes for several hours while very obese and watched penalties slowly creep down as weight decreased.
- Made sure that stamina burn from moving accurately reflected my penalty BMI - picked up a 200kg object and moved around, and set my stored kcal such that I was 400kg over the expected healthy value, and saw the same stamina drain in each case. 

#### Additional context
I don't know C++. Hopefully my code is OK. Probably needs to be squash-and-merged because that's a lotta commits.


Night-Pryanik:
With #58380 and #60202 this will finally make XXXL characters difficult to play with.

anoobindisguise:
Actually yours may not be necessary anymore since mouth encumbrance is heavily determining available stamina and is in JSON instead of hardcode, but there wouldn't be anything wrong with stacking the penalty further.

PatrikLundell:
I don't think extra weight should apply 1:1 as a carry penalty:
- Obese people get extra training from lugging all that weight around, so their muscles are stronger than a comparatively equally unfit slimmer person, allowing them to actually move in a relatively normal fashion.
- Applying the penalty 1:1 penalizes tall people unfairly (and probably favors short ones), since they weigh more for the same overweight ratio. If the system scaled the carrying capacity with size it probably wouldn't matter, but currently it doesn't scale.

It can be noted that there are several kings (and probably others, but the kings are naturally more in focus) from history who've lugged around a significant amount of extra weight but still managed to wear full plate armor as the armor was made for them, and thus had the required extra space where it was needed.

anoobindisguise:
> I don't think extra weight should apply 1:1 as a carry penalty:
> 
> * Obese people get extra training from lugging all that weight around, so their muscles are stronger than a comparatively equally unfit slimmer person, allowing them to actually move in a relatively normal fashion.
> * Applying the penalty 1:1 penalizes tall people unfairly (and probably favors short ones), since they weigh more for the same overweight ratio. If the system scaled the carrying capacity with size it probably wouldn't matter, but currently it doesn't scale.
> 
> It can be noted that there are several kings (and probably others, but the kings are naturally more in focus) from history who've lugged around a significant amount of extra weight but still managed to wear full plate armor as the armor was made for them, and thus had the required extra space where it was needed.

I was thinking about that myself, whether it feels realistic or not. There are 300lb people who can still jog, but they just tire really fast. That's about equivalent to carrying 150lbs of stuff. For average characters (100lbs carryweight @ 8 str) that's enough to slowly exhaust you as you walk, or sustainable for a stronger individual. It could be pretty easily cut in half in character.cpp but I was not sure whether or not to go ahead with it - would be open to opinions from senior devs who have a more definitive say on these things.

anoobindisguise:
> It can be noted that there are several kings (and probably others, but the kings are naturally more in focus) from history who've lugged around a significant amount of extra weight but still managed to wear full plate armor as the armor was made for them, and thus had the required extra space where it was needed.

as far as this goes, we can adjust the point at which penalties kick in, although I'm assuming that the overwhelming majority of rigid armors in game will be intended for people with healthy weights. Plus medieval kings weren't nearly as fat as a lot of people today.

PatrikLundell:
Yes. Ideally we'd have a point at which you'd need a custom made armor, and beyond that a point at which you wouldn't be able to carry both your (excess) body weight and an extra large suit of armor, having to shed one to be able to move.

RelMayers:
Try holding something close to body vs on extended arm forward, see how heavy is 2L drink in feel when having in bag vs after you drank it..

RelMayers:
Also BMI is quite simple and probably isn't good base to use.

src/bodypart.cpp
```diff
@@ -1013,6 +1016,21 @@ int bodypart::get_temp_conv() const
     return temp_conv;
 }
 
+int bodypart::get_bmi_encumbrance_threshold() const
+{
+    return id->bmi_encumbrance_threshold;
```
bombasticSlacks, line 1021: Don't need "id->" here youre already in the object

data/json/effects.json
```diff
@@ -2172,6 +2172,15 @@
     "part_descs": true,
     "base_mods": { "speed_mod": [ -2 ], "pain_min": [ 1 ], "pain_chance": [ 2 ], "pain_max_val": [ -1 ] }
   },
+  {
+    "type": "effect_type",
+    "id": "squeezed",
+    "name": [ "Uncomfortable" ],
+    "desc": [ "You're too fat to wear hard armor on your %s.  It's pretty uncomfortable." ],
```
bombasticSlacks, line 2179: I would avoid using fat here as a descriptor. Something like your extra fat makes it hard to wear armor on your %s. Or being overweight makes it hard to wear hard armor on your %s.

src/character.cpp
```diff
@@ -4065,6 +4094,9 @@ void Character::set_stored_calories( int cal )
 
         //some mutant change their max_hp according to their bmi
         recalc_hp();
+        //need to check obesity penalties when this happens
+        calc_encumbrance();
```
bombasticSlacks, line 4098: Duplicated and redundant with the calls above. Should probably just be in this function

bombasticSlacks (commented):
Here's a few things I noticed. Id have to load it up in game for more. One other thing is that there should be UI letting the player know their carry weight is effected. IMO it should be removed from their max, not added to current.

anoobindisguise:
> Here's a few things I noticed. Id have to load it up in game for more. One other thing is that there should be UI letting the player know their carry weight is effected. IMO it should be removed from their max, not added to current.

Thanks for the feedback! I don't want to remove the capacity from max, because weight penalties are based on a fraction of overburdenedness - if the player became heavy enough to reduce weight to 0, there would be divide-by-zero errors, and at extremely low weight capacities you could no longer safely lift any object and become massively overburdened by picking up a gallon jug of water or something. Adding weight to base weight sidesteps a lot of those issues IMO. I can look into differentiating overburdened messages for body mass and carried items though.

Maleclypse:
How does this interact with traits in the `LARGE/HUGE` family?  Do they currently change the PC's weight at in game or is it just described as they get bigger but no weight change occurs?

anoobindisguise:
> How does this interact with traits in the `LARGE/HUGE` family? Do they currently change the PC's weight at in game or is it just described as they get bigger but no weight change occurs?

Yes, these do affect your base weight. you can observe this by spawning a survivor, seeing weight 175 lbs, then mutating LARGE (289 lbs) and SMALL (88 lbs) however this has no bearing on BMI, so that will definitely need to change. BMI stays at 25 regardless of your size (55000 kcal)

Maleclypse:
> > How does this interact with traits in the `LARGE/HUGE` family? Do they currently change the PC's weight at in game or is it just described as they get bigger but no weight change occurs?
> 
> Yes, these do affect your base weight. you can observe this by spawning a survivor, seeing weight 175 lbs, then mutating LARGE (289 lbs) and SMALL (88 lbs) however this has no bearing on BMI, so that will definitely need to change. BMI stays at 25 regardless of your size (55000 kcal)

Ok good, so this wouldn't result in characters with those traits suddenly breaking and that BMI impact for those traits could be either adjusted in this PR or in a followup PR?

anoobindisguise:
> > > How does this interact with traits in the `LARGE/HUGE` family? Do they currently change the PC's weight at in game or is it just described as they get bigger but no weight change occurs?
> > 
> > 
> > Yes, these do affect your base weight. you can observe this by spawning a survivor, seeing weight 175 lbs, then mutating LARGE (289 lbs) and SMALL (88 lbs) however this has no bearing on BMI, so that will definitely need to change. BMI stays at 25 regardless of your size (55000 kcal)
> 
> Ok good, so this wouldn't result in characters with those traits suddenly breaking and that BMI impact for those traits could be either adjusted in this PR or in a followup PR?

No, if you are barely obese and become LARGE you don't become "less obese", you just remain the same amount of obese, because obesity doesn't care about how tall you are currently.

src/character.cpp
```diff
@@ -3468,6 +3470,14 @@ void Character::calc_discomfort()
             add_effect( effect_chafing, 1_turns, bp, true, 1 );
         }
     }
+    // do the same for obesity conflicting with rigid armors
+    remove_effect( effect_squeezed );
```
Maleclypse, line 3474: I'd suggest adding a second flag for armor made by the character or maybe a hidden trait that tracks BMI for when the character received it and whether it fit then?  I'm trying not to overcomplicate it but it would be bizarre for the PC to received the `squeezed` status from armor they just finished making or that Cody made specially fitted to them.

anoobindisguise: My thinking initially was that Cody was specifically making armor around your "healthy" weight and might refuse to even measure you if you were currently too overweight, and it does cause some questions about whether the armor should fit you when you're also thin, but I can go ahead and add a simple flag that makes it ignore BMI "BMI_BESPOKE" or something

anoobindisguise: For now we're going to get rid of this check, and later on we will have clothing sized per different body sizes as a granular system. So no need to have this basic incomplete system in the meantime. McDonalds and platemail will remain compatible for the forseeable future.

kevingranade:
BMI is the wrong thing to attach pretty much all these effects to.
I'm not clear on how a player ever becomes overweight enough that they're encumbered simoly from existing. DDA characters should not reach that level of post-morbid obesity, if theyvdo it's a bug.
Excess weight isn't any different from any other weight, all carried weight calculations should just use total weight.
For insulation, we need sources that say what exactly is driving it.
Again, I'm sceptical that dda characters ever reach "can't fit through a window" due to obesity in particular.  That's a LOT of excess weight.
For mechs, they *should* have sizing constraints, but BMI wouldn't be the driver for that.

What we want here is determining a body size, which we'd then compare to sizing of clothes.
Probably the simplest version that would be representative is height and weight.
For "street clothes", that could get evaluated to a broad category like, XS, S, M, L, XL etc, possibly multiplied by short, medium, tall parameters fio height.
Then a more restrictive article like plate would have finer tolerances, within just inches of height and a few pounds of weight.
Mechs would presumably be designed for varying levels of sizing, civilian mechs would probably go from S to XL, military mechs would probably only handle M to L.

anoobindisguise:
> BMI is the wrong thing to attach pretty much all these effects to. I'm not clear on how a player ever becomes overweight enough that they're encumbered simoly from existing. DDA characters should not reach that level of post-morbid obesity, if theyvdo it's a bug. Excess weight isn't any different from any other weight, all carried weight calculations should just use total weight. For insulation, we need sources that say what exactly is driving it. Again, I'm sceptical that dda characters ever reach "can't fit through a window" due to obesity in particular. That's a LOT of excess weight. For mechs, they _should_ have sizing constraints, but BMI wouldn't be the driver for that.
> 
> What we want here is determining a body size, which we'd then compare to sizing of clothes. Probably the simplest version that would be representative is height and weight. For "street clothes", that could get evaluated to a broad category like, XS, S, M, L, XL etc, possibly multiplied by short, medium, tall parameters fio height. Then a more restrictive article like plate would have finer tolerances, within just inches of height and a few pounds of weight. Mechs would presumably be designed for varying levels of sizing, civilian mechs would probably go from S to XL, military mechs would probably only handle M to L.

Currently our BMI models ONLY amount of fat/stored kcal, so it's basically an idealized (if wrong) version of BMI that doesn't account for muscle mass.
It's also quite easy to become obese, At least currently, this may be due to it being too easy to build body mass due to our simplistic stored kcal system, but by abusing high calorie foods you can easily intake over 20k calories per day. This may be wrong.
Obesity incurring encumbrance represents the increased mass on your body, just like wearing heavy clothing would encumber you. If you arms have an extra 5-10lbs of fatty mass on them it's going to be a bit harder to use them - more effort to swing a weapon. Torso encumbrance, you're carrying a heavy load on your torso so it would throw off your balance and whatnot. As far as difficulty crawling through a window or whatever goes, I can definitely remove that and wait for a proper rework that takes into account stuff like carrying a giant backpack etc.

anoobindisguise:
Made some edits, and compiled + loaded up in-game. Now instead of adding real carryweight, stamina burn from movement now just pretends you're carrying more than you really are based on your obesity. I think this is a lot more of an elegant solution.

I-am-Erk:
I think handling this as 'extra carried weight' is probably the right way to do it. Echoing Kevin's concerns though, I think using BMI as the measure of it will potentially cause confusion as we model more and more physiological symptoms.

I would suggest that you can probably keep the equations of this PR about the same, the numbers look ok... but add a simple secondary filter, instead of using BMI directly calculate "body fat percentage". At the moment that can be simply derived from stored calories (ie. the same calculation as bmi). However then later, for example after we add the ability to gain muscle mass, we can increase the player's weight without increasing their encumbrance or having other unexpected downstream effects.

Another option, and I don't mind this, would be to tie these effects to the XXL starting trait, and encode that trait to go away if you lose sufficient weight.

Night-Pryanik:
You suggest removing the trait if character loses some weight. How can you remove the trait if it's already removed after game start?

anoobindisguise:
OK so currently, assuming my code works (need to test it) things should work like this:
Healthy stored kcal is no longer always 55000, but is instead 20% of your weight at BMI 25. This ranges from 80k at 145cm to 150k at 200cm (10-20kg of fat total for the body with the remaining 80% being muscle, bone, etc). This means if you suddenly get the Huge mutation you're gonna be severely underweight.
Your BMI is 8 by default (no muscles or fat, just bones, the brain and stuff) and is increased by 1 for each point of natural strength you have. For an average, sedentary person that puts you at BMI 16 + your fat. Fat BMI is the "weight" of your stored kcal relative to your total body mass. So 10kg of fat on a 145cm character is exactly the same of a BMI increase as 20kg of fat on a 200cm character. Previously, BMI was not tied to height, which led to silly shit like a 200cm character gaining 6 pounds on a bathroom scale by eating 3500 kcal's worth of food.
Limb encumbrance as well as healthiness is gonna be moved from true BMI to the "fat only" BMI. If you had 20 strength, you are likely clinically obese, but you shouldn't get unhealthy from that so we should probably tell the player they are "obese" or "underweight" based solely on their fat mass (stored kcal) and ignore musculature for that, even if their true BMI is now reporting correctly.

anoobindisguise:
OK, so currently if you are under 2.0f points of "fat bmi" (bmi points that come from stored calories) your base strength starts getting linearly reduced to 4 as you reach zero stored calories. This means you die at BMI 12 no matter how much strength you have. I also intend to make calories decrement or increment slower the more strength you have assuming you are malnourished, likely half as much fat burn and gain while at str 8, 1/3 at str 12, 1/4 at str 16, so on, and no change at str 4 (since you don't really have any more muscle to burn). Your strength will harmlessly recover to its starting value if you get well fed rather than being permanently damaged, but you need to eat that many more calories to get it back.

src/character.cpp
```diff
@@ -4069,16 +4100,25 @@ void Character::set_stored_kcal( int kcal )
 void Character::set_stored_calories( int cal )
 {
     if( stored_calories != cal ) {
+        int cached_bmi = std::floor( get_bmi_fat() );
```
KorGgenT, line 4103: you don't need to cache this if you're only using it once

anoobindisguise: I believe it's necessary because the check calc_encumbrance(); is heavy enough that I don't want it to run every time stored calories updates. It records the whole-number BMI before the change, then compares the whole number BMI after the change to see if it's different because the limb encumbrance from obesity only cares about the whole number differences. I don't think there's a way to get BMI at an "artificial" kcal value, it only checks for your current kcal and you can't manipulate that in the function call. If that make sense

src/character.cpp
```diff
@@ -4057,16 +4067,22 @@ void Character::set_stored_kcal( int kcal )
...
+            calc_encumbrance();
+        }
     }
 }
 
 int Character::get_healthy_kcal() const
 {
-    return healthy_calories / 1000;
+    float healthy_weight = 5.0f * std::pow( height() / 100.0f, 2 );
+    return std::floor( 7716.17 * healthy_weight );
```
Skrassh, line 4085: What is the 7716.17 here? It seems like it should be a const to avoid having a magic number.

src/character.cpp
```diff
@@ -6326,7 +6379,10 @@ void Character::mod_stamina( int mod )
 void Character::burn_move_stamina( int moves )
 {
     int overburden_percentage = 0;
-    units::mass current_weight = weight_carried();
+    //add half the difference between current stored kcal weight and healthy stored kcal weight to weight of carried gear
+    units::mass fat_penalty = units::from_kilogram( 0.5f * std::max( 0.0f,
+                              ( get_healthy_kcal() - get_stored_kcal() ) / 7716.17f ) );
```
Skrassh, line 6384: Same as my other comment, make this a const since it's used in more than one place and to make the code more readable.

src/character.cpp
```diff
@@ -5883,12 +5902,46 @@ float Character::healing_rate_medicine( float at_rest_quality, const bodypart_id
...
 
 units::mass Character::bodyweight() const
 {
-    return units::from_kilogram( get_bmi() * std::pow( height() / 100.0f, 2 ) );
+    return bodyweight_fat() + bodyweight_lean();
+}
+
+units::mass Character::bodyweight_fat() const
+{
+    return units::from_kilogram( get_stored_kcal() / 7716.17 );
```
Skrassh, line 5933: Same as the other, make 7716.17 a const.

src/character.cpp
anoobindisguise: 7716.17 is 3500 * 2.20462 which is the number of kilocalories in a kilogram of fat. It could potentially be defined somewhere else, just not sure how to do that exactly.

Skrassh: Easiest way would be to make it a const in character.h make sure to leave a comment above it explaining what it is.

src/character.h
```diff
@@ -116,6 +116,8 @@ using drop_locations = std::list<drop_location>;
 using bionic_uid = unsigned int;
 
 constexpr int MAX_CLAIRVOYANCE = 40;
+// used for some calculations to convert stored kcal into body weight. this is 7716.17
+constexpr float KCAL_PER_KG = 3500 * 2.20462;
```
anothersimulacrum, line 120:
```suggestion
// kcal in a kilogram of fat, used to convert stored kcal into body weight. 3500kcal/lb * 2.20462lb/kg = 7716.17
constexpr float KCAL_PER_KG = 3500 * 2.20462;
```
```suggestion
// kcal in a kg of fat, used to converted stored kcal to body weight
constexpr float KCAL_PER_KG = 7716.17;
```

anoobindisguise: Thanks!!

alp7292:
this doesnt update docs

kebkaldanil:
Cool addition
Actually, there was bad consequences: almost infinite cold

adamkad1:
I think height should be edited to say something like 'Plays a role in your weight calculation.'

adamkad1:
> Cool addition Actually, being fat had bad consequences: almost infinite cold


That doesnt make sense

PR #60601, opened 2022-08-30
Turning off spell explosions where they not intended

### LINKED ISSUE #60594: Guilt is stopped by a fence
### Describe the bug

Ordinarily dispatching a zombie child gives you a negative mood, something about feeling guilty for doing that. The effect plays a very brief visual effect when it happens. When there is a fence between you and the zombie child, the effect plays briefly on the fence instead of on the player character, and no negative moods are applied to you.
This is especially notable when playing with the 'Pacifist' trait, where this negative mood is strongly amplified.

### Steps to reproduce

1. Find spot where a fence encircles you completely, TCL for example (I don't know if a gap in the fence affects this)
2. Dispatch zombie child which is on the other side of the fence.
3. Observe effects.

### Expected behavior

Guilt can get past fences.

### Screenshots

34545656r7

- OS: Windows
    - OS Version: 10.0.19043.1889 (21H1)
- Game Version: 9d29e93-dirty [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

There is a great joke to be made with this bug about how the human mind handles past trauma by building a big fence around it in order to avoid it at all cost, but I'm afraid that would be terribly off-topic.

GuardianDll:
#### Summary
None

#### Purpose of change

There is a lot of explosion spell effects, when there is no reason for them to be used

#### Describe the solution

Prevent a lof of unnecessary explosions by using `NO_EXPLOSION_SFX` flag (yes, it named **S**FX, and yes, it prevent **visual** effect)

Not doing it
Ask an artist to make a custom textures for each of spell (i dont think its a good thing, but i know in some tilesets it exist)

#### Testing

Works good imo

#### Additional context

Necro-boomer do not explode like nuclear warhead anymore
GIF 30 08 2022 20-04-10

Zombie child do not ignite you when you kill it
GIF 30 08 2022 20-28-23

Also i fix #60594 when i was here

### DIFF
--- a/data/json/monster_special_attacks/spells.json
+++ b/data/json/monster_special_attacks/spells.json
@@ -755,4 +755,3 @@
     "extra_effects": [ { "id": "hallu_touch_suicide", "hit_self": true } ]
   }
 ]
-    


Ramza13:
Excellent, I kept thinking I should do something like this after adding the flag, thanks for taking care of it.

GuardianDll:
Btw someone need to change it from SFX to VFX in the future

Ramza13:
> Btw someone need to change it from SFX to VFX in the future

Ha yea I didn't even think of what the name meant just copied it form some of the code it was around.

PR #60602, opened 2022-08-30
Remove gamma globulin shot

Venera3:
#### Summary
None

#### Purpose of change

IV immunoglobulin therapy is a real thing that does nothing to "strengthen your immune system", much less help your wounds to heal.

#### Describe the solution

- Remove the item and its spawn entries.
- Remove the `VACCINE` iuse (caused pain, added a syringe and set your health mod to 200)
- Migrate the ivig shot to an empty syringe

Migrate to homeopathic pills, but that's a bit too cheeky.
Add a 1/1,000,000 chance of the flu or food poisoning to progress into Guillan-Barré and have repeat gamma globuline shots maybe help a bit if you squint.

#### Testing

Loads without errors, a save made before the changes with shots in the inventory replaces them with syringes.

#### Additional context

If any mod depended on them effects and EoCs (`u_mod_healthy`) can modify the health mod without needing hardcode and keeping all functionality.


anoobindisguise:
They could be made to very slightly boost recovery from very low health or help in case of an infection. I hear they are used to prevent hep A after contact, but it's a bit dubious to extrapolate them to other infections.

Venera3:
It's used for severe immunosuppression when "whatever antibodies you can scrounge up from the general populace" is a step forward, but if you're severly immunocompromised you're not surviving the end of the world.

Aside from that, "health" does a lot of magic bullshit that has nothing to do with having an immune system, and even the stuff that *has* would never be impacted by circulating IG.

Amoebka:
While the current item is clearly nonsensical and has to go, the strong health boosting effect is an interesting and useful thing to have in the game. Perhaps it could be reflavoured as some super-science serum and moved to labs, similar to royal jelly / panaceus pills? Resistance to diseases and faster wound recovery sounds eactly what the military would want, after all.

Venera3:
The usual rule of "if you want stuff that does magic, add magic stuff, not act like a real thing does magic" still applies, of course.

Amoebka:
Well, that's what I'm saying. Why not rename it to something magical and move it to labs instead of outright axing it?
It performs a somewhat important function of a failsafe for the health system as a whole. If an unsuspecting player nukes their health to -200 somehow, they at least can recover it with a shot instead of ctrl+Q. The item can be rare, but, imo, it should still exist in some form.

Venera3:
Neither of those are an argument to keep it around. There's no reason to keep a hardcoded iuse for a thing that can be done in JSON, and if there's an *unintended* way for players to fall into a health spiral we need to know about it insteas of slapping a bandaid on it - and on the flipside having a freebie "get out of bad health free" card makes the system irrelevant.

Also, "my health is -200 so I'll just kill myself" is quite overstating the effects.

Terrorforge:
I was a bit concerned about the -200 scenario as well, but it turned out to be because I had some misconceptions about how Health works. I thought healing stopped entirely at -200, and I assume this used to be the case since it's still in some test cases, but looking at the current code it only cuts your healing by 50%. Given how quickly you heal by default, this is really not much of a problem unless you're also an Imperceptive Healer. Wear a gas mask to deal with airborne illnesses and 99% of players won't have an issue.

It does completely turn off your ability to mend broken limbs, though, which is where a lot of the complaints that you "can't heal" after e.g. starting with the XXXL trait come from.

Issue #60565, opened 2022-08-29
Odd upgrade path for toads


a-chancey:
### Is your feature request related to a problem? Please describe.
Oof i don't know how to fix the formatting into a table easily, but at least it makes sense.

#60234 mainlined a handful of animals from tropicataclysm, and the toads seem a bit *odd*, which I'll wind up writing a PR to address shortly.
Current evolution path prior for bullfrogs was:
`T1           |T2               |T3               |T4`
`Bullfrog     |Mutant Bullfrog  |Giant Bullfrog   |Colossal Croaker`
And zombification was:
`T1           |T2               |T3               |T4`
`N\A          |Zombullfrog      |Zombullfrog      |Frog Mother`

Toads are handled a bit more oddly:
`T1              |T2               |T3               |T4`
`toad tadpole    | toad            |Mutant Bullfrog  |as above'
and zombification is:
`T1           |T2               |T3               |T4`
`N\A          | N\A             | Zombullfrog     |as above`

Outside of that there's:
`T1           |T2               |T3                   `
`Huge Toad    |Giant Toad       |Gastro Bufo (zombie) `
and zombification is:
`T1           |T2               |T3                   `
`Zombullfrog  |n/a              |Gastro Bufo (revives)`

The IDs are a mess as well, but I'm not going to touch that.

### Solution you would like.

I'll write up the PR in a day or two but I'll implement it as follows:

Evolution:
`T1           |T2               |T3               |T4`
`Toad Tadpole |Toad             |Huge Toad        |Giant Toad`
Zombification:
`T1           |T2               |T3               |T4`
`N\A          |N\A              |Zombie Toad      |Gastro Bufo`

I'll add the zombie toad based on the zombullfrog i added, and i'll get all the evolution paths and whatnot adjusted properly. This will stop the toad from evolving into a frog (it even says it's a sister to frogs in the description), and will provide a clear (and not over powered) upgrade and zombification path for the toads.

### Describe alternatives you have considered.

Not doing it.

### Additional context

I am working on a few other things, i just happened to notice this today while testing my other changes and don't have time to get it knocked out at the moment.

Marcos6226:
Related #60554

PR #60605, opened 2022-08-30
Add a common zombification route to mutant fauna

Venera3:
#### Summary
Content "Add meat cocoons and zombie amalgamations"

#### Purpose of change

Zombification is a bit weird currently in that it only happens to normal critters / humans, mutants just fall to the ground and stay down. That's not very apocalyptic, but on the other hand adding 6000 zombie variants would be a huge chore and not very interesting.
This PR tries for a solution that fixes both, and serves as a little teaser for the flesh factories that the (far, far) future has in store.

#### Describe the solution

- Add meat cocoons of different sizes to serve as an intermediate step of reconstituting the weird mutant biomass on the spot
- Add 6 new purpose-made abominations to emerge from the different sizes of cocoons (see Additional Context for the critters)

TODO: 
- [x] Harvest snippets, those are always fun
- [x] Combat testing
- [x] Actually add the cocoons as a `zombify_into` field to all mutant animals, arhtropods included
- [x] Since zombos now have a reason to hunt bugs dissolve the zombie / creepy crawly truce and watch swamps neighboring towns erupt into a multigenerational free-for-all. That should be fun.

Lots.

#### Testing

Monsters zombify into the cocoons, the cocoons spawn the correct number of critters respecting spawn conditions (huge thanks to dseguin for tweaking the nested mongroups / growup spawns).
Weakpoints trigger, the forbidden / required effects logic up the monster specials as intended.

#### Additional context

~~I'm aware `abomination` isn't very inspired as far as names go, so I'm open to suggestions.~~ Decided on `amalgamation`, that's a bit less generic.
 The general flavor for these is an alien bodyplan with 7 limbs warped to resemble earthly animals in a slightly mix-and-match fashion, meant as single-tasked throwaway drones/workers. 

Currently three small and three medium-sized critters, the composition of the monstergroups is meant to allow for different spawns adding up to roughly the cocoon's initial weight.

Small amalgamations:
- `swarming` - small, antish things running around and taking a bite out of you if you let them, basically the Blob's interpretation of army ants
- `vigilant` - great vision range and a warning screech, meant to guide the other abominations to your position (their vision isn't that hot), pretty much nonthreatening alone
- `caustic` - what's better than bites? Acid! Lots of it, for that matter, but it needs a powerup attack to unlock its best special and weakpoint hits will disable it consistently. Starts spawning 120 days in, because acid.

Medium amalgamations:
- `leaping` - Cat-bat-bird things running slowly but leaping fast to take a bite out of you.
- `brutish` - Bear-Dinosaur-Centipede beatstick to run up to you and kick your teeth in with a clubbed tail and a lot of chomp.
- `charged` - What if shockers, but worse(and crabby)? Heaps a glowing organ filled with Blob Bullshit-flavored lightning around to lug at you, starts spawning 120 days in for understandable reasons.

Zombification rules: 
- Monsters below 10 kg don't cocoonify
- 10 - 35 kg monsters zombify into the tiny cocoon, spawning a single swarmer
- 36 - 100 kg monsters turn into the small cocoon, spawning (surprise) a mix of small monsters
- 101 - 300 kg monsters turn into the medium cocoon, spawning a mix of small and medium monsters
- 301+ kg monsters turn into a large cocoon, that spawns a larger number of small/mediums


sadenar:
Small question, would this new system be easy to opt out of if somebody does get an interesting idea for a specific directly revivified mutant zombie?
My other issue with it, and I get this is probably personal, but how do we explain the fact that the blob doesn't do this to our existing mutated animal zombies, or even normal animal zombies? Are their physical capabilities and morphologies just considered too useful to throw into the meat furnace to radically transform?
I do agree this is an interesting fix that would go around the fact most of our current mutants don't revivify and that getting a zombified version of each would be a lot of work and this goes around it without some sort pseudolore of "yeah the blob just doesn't like bugs".
As for the name, chimera probably also be a good idea, after all, it IS by your own explanation a jumbled up mix of various animal features, just like a chimera is.

Maleclypse:
Anathema, perversions, torments, or shames would be less common.  Dybbuk if you want to go biblical.

NetSysFire:
>with Blob Bullshit-flavored lightning

Something-something electric eels perhaps? They have an organ to naturally generate electricity, so it could be explained that an organ vaguely similar to that has grown in them.

a-chancey:
I was gonna work on toads tomorrow or Thursday with #60565 if you don't mind me leaving toads with an upgrade and zombie path similar to bullfrogs

Venera3:
I'm not principally opposed to direct zombification, I'm just not a fan of boilerplate "slower but amgy" zutants. It would be neater if it was a random thing (ie a toad either stands up or gets cocoon'd), but that's a tangent until more specific zombies happen.

>Something-something electric eels perhaps? They have an organ to naturally generate electricity, so it could be explained that an organ vaguely similar to that has grown in them.

Don't get me wrong, I'm completely happy with Blob Bullshit™. I was *this* close to doing cold/fire damage variants, but I decided to stay within the established zombo abilities.

>My other issue with it, and I get this is probably personal, but how do we explain the fact that the blob doesn't do this to our existing mutated animal zombies, or even normal animal zombies, from there, why doesn't it do this to everything? Are their physical capabilities and morphologies just considered too useful to throw into the meat furnace to radically transform?

There are broad strokes long-term plans for that, actually -
blobbified areas recycling their biomass in zombie factories after a while. That needs a lot of infrastructure that doesn't exist yet, but it's the direction things are going to go.

NetSysFire:
>There are broad strokes long-term plans for that, actually

Is there a link to that? I'd be really interested in that and general planning how the future ingame (6 months after the Cataclysm and beyond) is going to look like.

Venera3:
> > There are broad strokes long-term plans for that, actually
> 
> Is there a link to that? I'd be really interested in that and general planning how the future ingame (6 months after the Cataclysm and beyond) is going to look like.

~~It has been revealed to me in a dream~~ mostly devcord chatter, if you search for "flesh/meat/zombie factory" you'll find most of the relevant stuff.

NetSysFire:
>mostly devcord chatter

it'd be great if such roadmaps can be piped into the wiki or something because some people, like me, do not have discord and miss out on all that, plus it would also be great for organization if there are some plans/roadmaps you can actually work on because chat is rather fragmented

bombasticSlacks:
I think for next experimental we should create a "world progression" road map with some goals/ projects for factions (monster and NPC) to get that progression all together. So what does 6 months look like for X, how about 1 year? 2? Where X is migo, zombies, Hub, EXODII, Free Merchants, etc.

bombasticSlacks:
https://github.com/orgs/CleverRaven/projects/5/views/1 

We can populate as stuff get solidified as issues.

NetSysFire:
I have thought about something in the past few days: The Cataclysm has given birth to myriads of new predators. Why only predators? Where are the herbivores and related critters? E.g the blob making something to feed on the mycus only. Or something designed to conquer plants (both vegetation and triffids).

The remaining humans and exodii are a very tiny force compared to the now vast legions of zombies, so why does the blob continue producing such predators when the human race is already basically vanquished, when the mycus and triffids etc are spreading rapidly?

Venera3:
Something that kills a human (or a souped-up cyborg) won't have much problem tearing a triffid apart, and `HATEHATEHATE` is kind of the blob's flavor. I suppose the shock/acid would work less in theory, but currently the other factions are pretty much pushovers.

I don't want to communicate any relatably logical thought behind these in the "hmmm, there is a human wearing a gas mask and wielding a gun, release the zappers" sense, it's more "something's making a noise that's not us, go check it out and kill them". Direct application of blunt force is very much the zombies' calling card, with the minor players having to specialize to keep up like triffid fungal fighters (that don't actually do anything extra to fungals, but that's an issue for another day).

kevingranade:

https://discourse.cataclysmdda.org/t/what-turns-into-zombie/28167/5

Venera3:
Well that took bloody ages but it should be about good to go. Added a one-week grace period for ever cocoon to fizz out, and a pretty generous padding on the rest of the groups to keep them from being too jarring(zappers don't start spawning until Winter, for example). I need to do some live-fire testing on e.g. an ant hill getting flipped, but that's going to be the sort of thing that'll shake out in play as well.

data/json/monster_weakpoints/abomination_weakpoints.json
```diff
@@ -0,0 +1,339 @@
...
+            "chance": 50,
+            "duration": [ 5, 25 ],
+            "message": "The %s's proboscis is maimed!",
+            "damage_required": [ 16, 30 ]
+          },
+          {
+            "effect": "maimed_mandible",
+            "chance": 50,
+            "duration": 86400,
+            "message": "The %s's proboscis is maimed!",
```
TheShadowFerret, line 151: Shouldn't this have a different description if it's the 1 day version, like the other disabling effects have

TheShadowFerret, line 151:
```suggestion
            "message": "The %s's proboscis is destroyed!",
```
To match the other 1d mandible disabling message

data/json/monsters/zed_amalgamation.json
```diff
@@ -0,0 +1,331 @@
...
+    "death_function": { "effect": { "id": "death_shock_blast", "hit_self": true } },
+    "harvest": "zombie_amalgamation_electric",
+    "extend": { "flags": [ "ELECTRIC" ], "weakpoint_sets": [ "wps_amalgamation_armor", "wps_amalgamation_electric" ] }
+  },
+  {
+    "type": "MONSTER",
+    "id": "mon_amalgamation_soldier",
+    "copy-from": "mon_amalgamation_abstract_med",
+    "name": { "str": "brutish amalgamation" },
+    "description": "A large, thickly armored creature built not unlike a bear, if a bear had a long, clubbed tail and two short limbs ending in black talons framing a vertical mouth on an otherwise smooth, featureless head.  On second though, not that much like a bear at all.",
```
TheShadowFerret, line 290:
```suggestion
    "description": "A large, thickly armored creature built not unlike a bear, if a bear had a long, clubbed tail and two short limbs ending in black talons framing a vertical mouth on an otherwise smooth, featureless head.  On second thought, not that much like a bear at all.",
```

Fris0uman:
> Zombification rules:
    Monsters below 10 kg don't cocoonify
    10 - 35 kg monsters zombify into the tiny cocoon, spawning a single swarmer
    36 - 100 kg monsters turn into the small cocoon, spawning (surprise) a mix of small monsters
    101 - 300 kg monsters turn into the medium cocoon, spawning a mix of small and medium monsters
    301+ kg monsters turn into a large cocoon, that spawns a larger number of small/mediums

Can you document that somewhere?

PR #60607, opened 2022-08-30
Handle fuel energy and vehicle energy_consumption as units::energy

Hirmuolio:
#### Summary
Infrastructure "Handle fuel and some vehicle things as units::energy"

#### Purpose of change

* `"energy_consumption"` was in integer watts. It is now energy (per second).
* `"energy"` was in float MJ/L. It is now energy/L

Work towards units::energy based batteries.

#### Describe the solution

Refactor sections of vehicle and bionic code that use these.

Fixed bugs: 
* Pedaling would not produce bionic power at low levels due to rounding error.
* Vehicle batteries got significantly more than 1 kJ energy from single 1 kJ battery charge. Fixed that blatant violation of laws of physics.

The latter one means that electric vehicles now consume much more electricity. Your driving range is reduced by about 60%.

What is fuel `"energy"`? Doc and vehicle code treat it as MJ/L but bionic code treats it as kJ/charge. I have used the MJ/L as the correct definition.
Bionic code has some hacky workarounds to make its wrong energy use to give correct results. It seems to mostly work. Magicalysm blood bionic will show wrong values in some situations but that should be visual problem only.
It is a dirty hack where item sizes and fuel energies are so that MJ/L /1000 = KJ/unit.
Proper fix is full rewrite on how bionic fuel works (from materials to items). Left some TODO comments behind in there.

#### Testing

Tests pass.
Battery, gasoline, blood bionics generate correct amount of energy.
Loaded old save with some vehicles around.

#### Additional context

Please do not tell any physics teachers that we are using "Joule" for power here.

Problems with bionic fuel will be fixed by #60662


src/bionics.cpp
```diff
@@ -1493,21 +1494,22 @@ void Character::burn_fuel( bionic &bio, const auto_toggle_bionic_result &result
 
     map &here = get_map();
     weather_manager &weather = get_weather();
+    // TODO: Rewrite fuels to use item volume.
+    // The energy is in energy/L but we just divide by 1000 and treat it as energy/unit
```
Fris0uman, line 1498: isn't a unit 250 mL?

Hirmuolio: Unit here means one unit of item (one charge of battery, one charge of gasoline, etc.).

Fris0uman: Yes but to convert volume into unit you need to know the ratio between them, no?

Hirmuolio: Materials do not have volumes. That is the problem.

Fris0uman: Ah so it's just an arbitrary convertion untill there's a better fix?

Hirmuolio: Yes.

Fris0uman:
Is the failing vehicle test unrelated?

Hirmuolio:
Probably unrelated.
Locally ran tests pass. And I can't think of a way for changing energy to cause vision cache to fail like that.

PR #60628, opened 2022-08-31
Update documentation for recipes

RenechCDDA:
#### Summary
Infrastructure "JSON_INFO.MD documents more recipe fields"

#### Purpose of change
I'm looking into adding new canning recipes. The existing recipes use some fields which JSON_INFO.MD did not document.

#### Describe the solution
Research those fields and see how they're used in recipe.cpp. Document an example of each field and write an explanatory comment.


Maleclypse (approved):
Thanks for adding this documentation!

PR #60631, opened 2022-08-31
Active Defense System tweak

### LINKED ISSUE #41267: Active Defense CBM doesn't provide any feedback on its work
# Describe the bug

Subj. It outright doesn't work. Had a bandit throw at me full clip of his SIG P226, and it didn't proc once.

# Steps To Reproduce

1. Make a player with AD CBM and some power.
2. Spawn bandit, turret or w/e can shoot.
3. Activate AD, count how many times it procs.

Alternatively, load [this save](https://github.com/CleverRaven/Cataclysm-DDA/files/4773742/Grand.Meadow.10751.zip), wait till the fire burns out, activate AD and see if it works (hint: it does not)

# Expected behavior

Its description clearly says the bullets have a chance to be deflected. Was it nerfed somewhere in late 0.D or 0.E so it now just reduced their damage by 5% instead? If so, maybe its description should be updated. If not, it should actually deflect them, like it says.

# Versions and configuration

    OS: Windows 7
    Game Version: 0.E-3128-gb880329 [64-bit]
    Graphics Version: Tiles
    Game Language: English [en]
    Mods loaded: [
    Dark Days Ahead [dda],
    Aftershock [aftershock],
    Magiclysm [magiclysm],
    sees-player icon, retrodays [sees_player_retro],
    ]

Night-Pryanik:
#### Summary
Balance "Active Defense System CBM tweak"

#### Purpose of change
ADS's decrease of incoming damage is minuscule. In best case it will lower amount of incoming damage by 8 units (for bullet type of damage). It's simply not worth activating, considering its fixed cost of 25 kJ per attack and 10 J every turn in idle mode.
Also CBM doesn't notifies player when it's triggered, making player wonder whether it works at all.
Closes #41267.

#### Describe the solution
- Replaced fixed and minuscule values of amount of incoming damage by which this CBM lowers them with percentages, with a limit of blocking 50 units of incoming damage:
- - Bash damage: `rng( 1,2 )` -> 25%
- - Cut damage: `rng( 1,4 )` -> 33%
- - Stab and bullet damage: `rng( 1,8 )` -> 50%
- Added message which notifies player of triggering the CBM.
- Rewrote description of bionic as it mentioned "a chance of deflection" while in fact bionic works on any attack.

Leave fixed values but increase them significantly, like 2, 3, or 5 times.

#### Testing
Installed ADS and activated it. Spawned MP5 turret. Waited for it to fire at me. Checked the message in the log.

#### Additional context
None.

### DIFF
--- a/src/character_armor.cpp
+++ b/src/character_armor.cpp
@@ -318,12 +318,20 @@ const weakpoint *Character::absorb_hit( const weakpoint_attack &, const bodypart
         // The bio_ads CBM absorbs damage before hitting armor
         if( has_active_bionic( bio_ads ) ) {
             if( elem.amount > 0 && get_power_level() > 24_kJ ) {
+                // ADS requires 25 kJ to trigger
+                // Assuming it absorbs damage at 10% efficiency, it can absorb at most 2500 J of energy
+                // Taking muzzle energy of 5.56mm ammo as a reference, 1936 J is equal to 44 damage
+                // Having this in mind, let's make 2500 J equal to 50 damage, which will be the maximum damage ADS can absorb
+                const int max_absorption = 50;
+
+                // If damage is higher than maximum absorption capability, lower the damage by a flat amount of this capability
+                // Otherwise, divide the damage by X times, depending on damage type
                 if( elem.type == damage_type::BASH ) {
-                    elem.amount /= 2;
+                    elem.amount = elem.amount > max_absorption ? elem.amount - max_absorption : elem.amount / 2;
                 } else if( elem.type == damage_type::CUT ) {
-                    elem.amount /= 3;
+                    elem.amount = elem.amount > max_absorption ? elem.amount - max_absorption : elem.amount / 3;
                 } else if( elem.type == damage_type::STAB || elem.type == damage_type::BULLET ) {
-                    elem.amount /= 4;
+                    elem.amount = elem.amount > max_absorption ? elem.amount - max_absorption : elem.amount / 4;
                 }
                 mod_power_level( -bio_ads->power_trigger );
                 add_msg_if_player( m_good,


anoobindisguise:
Does it not use 10kJ every turn? The 10 in its power cost has not been given an energy unit, and generally speaking that always means kJ - see Sonic Resonator taking 100kJ and specifying a cost of 100. Generally speaking that leads to really annoying micro (it's optimal to turn it off whenever there are no zombies nearby and then turn it back on when you get into melee range, repeat 1000x for fighting in a city) and I'd even accept the power cost being raised on react just to get rid of the upkeep price because of how annoying it is in terms of gameplay.

Night-Pryanik:
Completely removing upkeep cost feels wrong and unnatural - forcefield requires power to operate in any case - but we could make it really low so it won't be noticeable in short term, like 10 J instead of of 10 kJ.

SirPendrak:
Looks nice, but shouldnt there be some limit in max ammount of dmg reduced? Cutting half damage off a tank round for human-portable bionic seems wrong.

Night-Pryanik:
Well, I'm not against it, I just need some number to set as a limit.

Night-Pryanik:
So, according to your suggestion, let's calculate.
ADS spend 25000 J to trigger.
Assuming 10% efficiency, we get 2500 J.
Further dividing this by 10, we get 250 J.

For example, .223 has 44 points of damage.  How should I compare 250 J to 44 points of damage?

anoobindisguise:
> So, according to your suggestion, let's calculate. ADS spend 25000 J to trigger. Assuming 10% efficiency, we get 2500 J. Further dividing this by 10, we get 250 J.
> 
> For example, .223 has 44 points of damage. How should I compare 250 J to 44 points of damage?

Easy, just reverse muzzle energy calculation, that's 1936J -> 44 damage, so 44 damage -> 1936J.

Night-Pryanik:
So, with some rounding we get 45 damage = 2000 J. 2500 J of ADS's trigger cost (assuming 10% efficiency) ~= 50 damage.
 I think setting a maximum of 50 units of blocked damage will be fine.

Night-Pryanik:
Ok, I added max amount of 50 damage ADS can absorb.

anoobindisguise:
> Ok, I added max amount of 50 damage ADS can absorb.

OK, so it requires 25 kJ of energy to trigger, if you have less it will not trigger. Then it consumes the square of the damage it absorbs? So a hypothetical 100 damage attack, costs 25kJ at 10% efficiency to reduce to 50 damage. Then quadratically less for less damage (20 bashing damage = 5 absorbed = 25 J / 0.1 efficiency = 250J)

Maleclypse (approved):
At this point any further testing needs to happen in a larger group of testers. Thanks for bringing it this far.

PR #60633, opened 2022-08-31
[Magiclysm] allow adept weapons as recipe tools

MrDraMaster:
#### Summary
Mods "Allow adept weapons as tools in recipes already using normal class weapons"

#### Purpose of change

Magiclysm recipes that allow a class weapon instead of a class rune as a tool do not allow the class adept weapon.

#### Describe the solution

Add the technomancer, animist, and magus adept weapon in all recipes with the "normal" version was used as a tool option.

#### Testing

* create new world
* debug => all recipes known
* check that recipes with `t:animist`, `t:technomancer`, and `t:magus` also mention the adept version of the weapon
* debug => grant the three adept weapons
* check recipes again to see that the adept weapons are highlighted green for present


Maleclypse (approved):
This makes sense.  Thanks for contributing!

PR #47374, opened 2021-02-09
fixed vehicle racking menu

Jamuro-g:
#### Summary
Bugfixes "Fixes vehicle racking menu. No more duplicates&phantom names"

#### Purpose of change

Overlapping bike rack lines (for example in corners) caused the racking ui prompt to count the vehicle for both rack lines.
It couldn't cause any issues (fixed that in an earlier pr) but it showed basically useless entries.

Fixes #47179

#### Describe the solution

Added a method that checks the found vehicle parts and bike racks for duplicates accross different vehicle entries and removes 
 the "shorter" one.
Basically bike racks accross a cornerpiece would find the same vehicle twice (once the full vehicle and once just the 1 tile at the corner).

#### Describe alternatives you've considered
Rewriting ( and probably splitting) the use_bike_rack method
This issue happens mostly due to how carried vehicles are fetched at the same time and the same way as potential racking candidates.

#### Testing

- Spawn in vehicle with bike racks
- Add bike racks to form a corner (see linked issue for an example)
- Spawn in a bike/motorcycle/... and rack it
- Go to the cornerpiece of the bike racks and activate it to open the racking menu

It no longer should show the racked vehicle twice.

PR #60634, opened 2022-08-31
[CR] Racking refactor (and bugfixes)

### LINKED ISSUE #40547: Nested Bike Racks causes mounted vehicles to merge permenantly
# Steps To Reproduce

1. Mount a bike rack on the back of a motorcycle or bike.
2. Mount another vehicle to the back of the bike (like a shopping cart)
3. Put 4 bike racks on the back of a car.  
4. Mount the bike with the currently mounted shopping cart to the back of the car
5. Unmount the bike and shopping cart from the car (this may take several tries)
6. Now the shopping cart is no longer detachable from the bike, they have merged into one vehicle permanently.

# Expected behavior
Either the shopping cart should still be dis-mountable from the bike after the bike is dis-mounted from the car, OR, nested mounts should be blocked and the bike should not be allowed to be mounted while it has a shopping cart mounted to it.

# Versions and configuration
 - OS: Xubuntu 18.04
 - Game Version: 0.E
 - Graphics version: Tiles
 - Ingame language: English
 - Mods loaded: Nothing extra

### LINKED ISSUE #52079: "Unracking task failed" error after vehicle(s) vanish post unload/reload on bike rack.
### Describe the bug

This is kind of two part and may be related to #52005.
First part, detaching one of several 'identical' vehicles causes the other 'identical' vehicles to vanish.
Second, if re-attaching the recently detached vehicle then detaching it again this error gets thrown:

`ERROR : src/activity_actor.cpp:1227 [virtual void bikerack_unracking_activity_actor::finish(player_activity&, Character&)] Unracking task failed.  Parent-Vehicle:Electric SUV with Bike Rack; Found parts size:12`

It seems the other vehicles didn't vanish but got merged into one invalid vehicle (3 frames in one slot).

### Steps To Reproduce

[Test_BikeRack.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/7274269/Test_BikeRack.zip)
From save:
1. Attach all three vehicles to bike rack. (technically 2 of them side by side is sufficient)
2. Detach one of the vehicles.  Observe there is only one choice with 3 attached.  Observe the others attached vanish.
3. Attach the vehicle.
4. Detach the vehicle (it's name is missing first letter in dialog?).
5. Observe error.

### Expected behavior

That the invalid state by step 4 is prevented... somehow...

There seems to be at least a couple different options to prevent that particular chain of events from resulting in an error.
- A new mechanism which makes identical vehicles perfectly valid. So step 2 would list 3 items and only one of the three would detach.
- Disallow identically named vehicles to be attached... So in step 1 when trying to attach the second Laundry Cart instead of it attaching it would not.  Bonus some feedback to the player letting them know there was already an identically named vehicle attached.  Having not yet poked around the relevant code this one seems easier to pull off.

### Screenshots

_No response_

- OS: Linux
    - OS Version: Distributor ID: Ubuntu; Description: Ubuntu 20.04.3 LTS; Release: 20.04; Codename: focal; 
- Game Version: 0.F-3146-g81ae26757a [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

I seem to recall something like this being known back in 0.E stable... I didn't notice an issue on it in my searching though so likely called something else.

A snippet of the log around the error from a debug build:
[debug-snip.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/7274306/debug-snip.zip)

### LINKED ISSUE #59152: Cannot unload 1 tile vehicle from bike rack
### Describe the bug

When I try to unload my 1 tile vehicle from my bike rack I select the option "Load or unload vehicle" and nothing happens, the vehicle is still attached to the main one.

There are no message on the log, no errors, no crashes.

### Steps to reproduce

1. Make a vehicle with a bike rack
2. Load a one tile vehicle to the bike rack
3. Try to unload the vehicle

### Expected behavior

The vehicle to be unloaded

### Screenshots

[Screencast from 10-07-2022 19:59:53.webm](https://user-images.githubusercontent.com/145339/178165344-2f2ef873-0023-49da-8b21-dc5917843e54.webm)

image

image

- OS: Linux
    - OS Version: LSB Version: n/a; Distributor ID: EndeavourOS; Description: EndeavourOS Linux; Release: rolling; Codename: rolling; 
- Game Version: de006e8 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Generic Guns [generic_guns],
]

### Additional context

_No response_

### LINKED ISSUE #60453: Bike racks causing vehicle tiles to merge
### Describe the bug

On a two tile vehicle, with bike racks on both tiles and identically-named sub-vehicles loaded on both racks, attempting to unload one of the sub-vehicles will cause it to merge with the other

### Steps to reproduce

1.) Have a two tile vehicle with bike racks on each tile
2.) Create two sub-vehicles and name them identically
3.) Load both sub-vehicles onto a rack
4.) Attempt to unload a vehicle from the rack

### Expected behavior

That the sub-vehicles not re-enact Jeff Goldblum's role in The Fly.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19044.1889 (21H2)
- Game Version: 3d2829c-dirty [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

Here is a video demonstrating the bug in action:
https://www.youtube.com/watch?v=Toim8dtdvWU

Attached save has a vehicle ready to demonstrate the bug; simply attempt to unload a vehicle from either rack:

[two_tile_vehicle_bike_rack_bug_demonstrator.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/9429904/two_tile_vehicle_bike_rack_bug_demonstrator.zip)

irwiss:
#### Summary
None

#### Purpose of change

Fixes #40547
Fixes #52079
Fixes #60453

Fixes #59152 (likely) - Out on a limb for this issue, the cause is likely either the merging issue fixed here or equivalent of #60507 but triggered on unracking or both. There's no save game attached so there isn't much that can be done beyond what's done in 60507 and this PR.

#### Describe the solution

Vehicle names are used to discriminate between parts belonging to racked vehicles, racking vehicle with same name, same axis(or 1-tile) and no tiles spacing between them makes racked vehicles merge together as there's nothing to discriminate between them. Positions are saved relative to their own pivot which when unfolded cause the invalid state where bunch of random parts are merged on the same tile.

The issues themselves are easy to fix - disallow racking if same name already racked somewhere on the vehicle; however, writing tests is close to impossible as racking code had no clear boundaries or a nice api to plug into tests.

These commits refactor the code so it's more easily testable and change the racking menu to allow racking specific vehicle rather than first found.

The final refactor isn't perfect:
* Bike racks with cardinal 1-tile away "direction" vector similar to headlights, could simplify racking significantly as the tile bike rack points at is always the target, and [make workarounds like this obsolete](https://github.com/CleverRaven/Cataclysm-DDA/blob/e6abe25a00f2d37aaed11987ff7bf49a0056033b/src/vehicle.cpp#L2025-L2041).
* There's still the [string soup even if it was consolidated to one place](https://github.com/CleverRaven/Cataclysm-DDA/blob/44345c6671b3ef56a66290ae77774b9453b41cfd/src/savegame_json.cpp#L3221-L3234), if this doesn't need a revert in a short while ( and I'll remember it ) - I'll revisit to rewrite proper object serialization.

#### Refactors

8687e854a044bf1f8cf69328817b13b656b51a91, dda445fa3f65c085d9c766cb676df928a4ce75bc - Splits up methods to allow looking up vehicles to rack/unrack, allows selection of vehicle in racking menu.
fce5953b13b02ae3bebc2dfcea0b2d9ae48fe89e, 2b72c50dc756ca67834b468589d612764b0203de - Moves the loose logic that was left outside merge_rackable_vehicle and remove_carried_vehicle into the methods.
44345c6671b3ef56a66290ae77774b9453b41cfd - Moves the string parsing to savegame_json.cpp - it's still not ideal but better than string manipulation in the middle of logic, and I don't want to change save format or write bidirectional migrations (yet)
90a8423d541fd42cbffd469fdfd7ab721d631502 - Tags some of the loose vehicle tests with [vehicle], as running tests with this filter didn't run these tests
ea8d478e23f10cadca8e25906937c8bc880e99d4 - Adds a couple tests to check the basic functionality and one specifically for testing #52079 and #60453

#### Function changes

`try_to_rack_nearby_vehicle` - got split into `find_vehicles_to_rack` and the remaining functionality is in `merge_rackable_vehicle`.

The code to find vehicles to unrack was moved from `vehicle::use_bike_rack` to `vehicle::find_vehicles_to_unrack`.

Const-ified `find_lines_of_parts` - because why not, and I needed to call it from const function.

Removed functions from vehicle.h: `remove_carried_flag`, `remove_tracked_flag`, `clear_bike_racks` - all their functionality was merged into merge_rackable_vehicle / remove_carried_vehicle and it's the only place they're used from.

Removed functions from vehicle.h: `validate_carried_vehicles` - hack from #47374 no longer necessary, [this should cover it](https://github.com/CleverRaven/Cataclysm-DDA/blob/dda445fa3f65c085d9c766cb676df928a4ce75bc/src/vehicle.cpp#L1578-L1591).

Adding something to discriminate vehicles with same name, for example the turn racking activity finished, to the weird string pasta that's used to save vehicle name, but that'd just pile on to already brittle code, I think requiring players to have distinct vehicle names isn't too much of a controversy.

#### Testing

Some tests for rack / unrack activities and functions are added.

Manual test of scenario described in #60453 should produce the image in attached screenshot.

#### Additional context

image

### DIFF
--- a/tests/vehicle_test.cpp
+++ b/tests/vehicle_test.cpp
@@ -30,6 +30,8 @@ static const itype_id itype_test_power_cord( "test_power_cord" );
 static const vpart_id vpart_ap_test_standing_lamp( "ap_test_standing_lamp" );
 
 static const vproto_id vehicle_prototype_bicycle( "bicycle" );
+static const vproto_id vehicle_prototype_car_rack( "car_rack" );
+static const vproto_id vehicle_prototype_wheelchair( "wheelchair" );
 
 TEST_CASE( "detaching_vehicle_unboards_passengers", "[vehicle]" )
 {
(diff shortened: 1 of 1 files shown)


PatrikLundell:
It's one thing to require players to name their vehicles with distinct names, and another to expect players to somehow know they'll have to rename existing vehicles when bringing them home if they want more than one. Welding and shopping carts can both be found in sufficient numbers that it's not unreasonable to try to gather them up and try to haul them back to base.
As a minimum, there's a need for the game to explicitly tell the player to rename additional vehicles when there's a name collision.

Would it be possible to add a suffix to the name of loaded vehicles, where the suffix is the location of the center tile (0, 0) of the loaded vehicle in the coordinate system of the loading vehicle? I.e. "Welding Cart (0,-4)" would be attached to the vehicle straight behind the driver's seat (assuming that's the center of the vehicle, which is usually the case), while "Welding Cart (-1, -4)" would be attached to the frame to the left of the previous cart.

irwiss:
The message sort of hints that you need to rename, I wanted to avoid immersion breaking metagame messages "because that's how serialization works right now", changed message to more obvious one.

As for solutions there are plenty even without considering string concat/split hacks, but all of them will break backward compatibility on saves which is why I postponed it until this is tested a bit.

Issue #54507, opened 2022-01-17
ERROR Failed to read from "./save/SSD krashen/#U3lzdGVtZmVs.sav": map::at


hexagonrecursion:
### Describe the bug

While trying to reproduce #43067 I have ran into an issue: I can not load the save file https://github.com/CleverRaven/Cataclysm-DDA/issues/43067#issuecomment-677671889 https://github.com/CleverRaven/Cataclysm-DDA/files/5103435/save.zip

```c++
09:20:47.645 ERROR : src/cata_utility.cpp:427 [bool read_from_file(const string&, const std::function<void(std::basic_istream<char>&)>&)] Failed to read from "./save/SSD krashen/#U3lzdGVtZmVs.sav": map::at
```

After this the game quits to the main menu

### Steps to reproduce

1. Load the save
2. Ignore the errors
3. The game quits to the main menu

### Expected behavior

I expected the save to load

### Versions and configuration

e49949d

### Additional context
```c++
-----------------------------------------
09:19:31.717 : Starting log.
09:19:31.718 INFO : Cataclysm DDA version e49949d
09:19:31.718 INFO : [main] C locale set to C.UTF-8
09:19:31.718 INFO : [main] C++ locale set to C.UTF-8
09:19:31.718 INFO : SDL version used during compile is 2.0.8
09:19:31.720 INFO : SDL version used during linking and in runtime is 2.0.20
09:19:32.309 INFO : Number of render drivers on your system: 4
09:19:32.309 INFO : Render driver: 0/opengl
09:19:32.309 INFO : Render driver: 1/opengles2
09:19:32.309 INFO : Render driver: 2/opengles
09:19:32.309 INFO : Render driver: 3/software
09:19:32.326 INFO : [options] C locale set to en_US.UTF-8
09:19:32.326 INFO : [options] C++ locale set to en_US.UTF-8
09:19:32.326 INFO : Setting language to: 'en'
09:19:32.326 INFO : [i18n] Scanning core translations from lang/mo/
09:19:32.326 INFO : [i18n] Scanning mod translations from ./mods/
09:19:32.327 WARNING : opendir [./mods/] failed with "No such file or directory".
09:19:32.446 INFO : Active renderer: 3/software
09:19:32.466 INFO : USE_COLOR_MODULATED_TEXTURES is set to 0
09:19:32.574 INFO : Setting language to: 'en'
09:19:32.794 WARNING : opendir [./mods/] failed with "No such file or directory".
[...]

Issue #43067, opened 2020-08-20
Foldable vehicles cant be dragged/moved


systemfelet:
Can't move my foldable shopping cart vehicle, only get "The foldable cart collides with errors in the movement code".
Tried unfolding/folding, loading/emptying even remade one from scratch, nothing works.

# Steps To Reproduce
Make a foldable vehicle out of  2x foldable light frame, folding shopping cart basket and casters , try to drag it.

# Expected behavior
It should be dragable as it was before. Stopped working last night, didnt patch in between, have updated since that to the latest experimental and still not working.

# Screenshots
![CDDA bug](https://user-images.githubusercontent.com/68592466/90760257-e5d9da00-e2e1-11ea-8883-73b7e87cbf6e.png)

# Versions and configuration

- OS: Windows
    - OS Version: 10.0 2004
- Game Version: 0.E-5228-g8731f05 [64-bit]
- Graphics Version: Tiles
- Game Language: Deutsch [de]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Rail Stations [No_Rail_Stations],
    Alternative Map Key [alt_map_key],
    SpeedyDex [speedydex],
    Stats Through Skills [StatsThroughSkills]

Well, that's a bit wierd that the game language says deutch through the generated game report becuase it's in English o.O

Otherwise, thanks for the wonderfull game, keep the good work up!

anothersimulacrum:
Can you upload the save?

systemfelet:
> 
> 
> Can you upload the save?

yes ofc, forgot that. 
[save.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/5103435/save.zip)

wapcaplet:
I could not reproduce this with a folding bicycle or folding shopping cart.

- OS: Linux
    - OS Version: Distributor ID: Pop; Description: Pop!_OS 20.04 LTS; Release: 20.04; Codename: focal; 
- Game Version: 0.F-5147-ge23fd18712 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

LeahLuong:
[...]

PR #60639, opened 2022-09-01
Support loading legacy save from #43067

irwiss:
#### Summary
None

#### Purpose of change

Legacy save from 0.E can't be loaded

1) Reading legacy json for daily calories is trying to read a missing key

2) Solving the previous it can now load, but once loaded it triggers debugmsg on every turn because gas mask calls get_var trying to read a double, but the variable has ',' for decimal separator

3) Error on loading missing SELFAWARE trait, seen this in a bunch of legacy saves

#### Describe the solution

1) Took the fix from https://github.com/CleverRaven/Cataclysm-DDA/issues/54507#issuecomment-1014366769

2) If ',' is the character the double parser fails on - replace it with a '.' and retry the read

3) Make a migration for SELFAWARE removal

There's more stuff missing - terrains, bionic, few item defs and legacy recipes missing but i'm not sure what's the correct fixes for those as it seems some overmap migrations are hardcoded, others are json and some are a mix of the 2. All of it can be `i`d through

#### Testing

Load the save from #54507, it should load without crashes ( still a bunch of migrations missing, so debugmsgs trigger but can be skipped )

### DIFF
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -1609,7 +1609,7 @@ void avatar::daily_calories::read_activity( const JsonObject &data )
         int times;
         if( !jo.read( member.first, times ) ) {
             continue;
-        };
+        }
         activity_levels.at( member.second ) = times;
     }
 }


src/avatar.cpp
```diff
@@ -1607,7 +1607,9 @@ void avatar::daily_calories::read_activity( const JsonObject &data )
     JsonObject jo = data.get_object( "activity" );
     for( const std::pair<const std::string, float> &member : activity_levels_map ) {
         int times;
-        jo.read( member.first, times );
+        if( !jo.read( member.first, times ) ) {
+            continue;
+        };
```
BevapDin, line 1612:
```suggestion
        }
```

PR #60642, opened 2022-09-01
Added golf ball to dodge practice

rj8216:
Golf balls for dodging practice

#### Summary
Balance "Added Golf Ball to required item for Dodge (Beginner) practice recipe"

#### Purpose of change
The description of the dodge practice "Practice your dodging by throwing a ball at a wall and dodging it when it bounces off." would imply that one only needs a ball capable of bouncing off a hard surface when thrown. In reality a golf ball certainly merits being added to the various balls that are currently available. They have hard rubber cores, and reliably bounce when thrown with any force. They are usually brightly colored to help a person to aid with eye / body coordination while dodging.  

#### Describe the solution
I added golf ball to the dodge practice recipe.

In the dev discord, someone suggested adding "rubber golf ball," but that seems needless since golf balls are already 80% rubber. This seemed the simplest fix.

#### Testing
I did edit the recipe on my local game on my PC, and successfully was able to practice with a golf ball, and gain dodge proficiency. Nothing bad seemed to happen.

None
#### Additional context
I can give anyone a hilarious gifycat link to an emu throwing a golf ball on a side walk, to watch it fall and practice dodging it. I just figured on here wasn't the right place. I also have links to articles about the makeup and composition of a standard golf ball. Just saying I did my research!

### DIFF
--- a/data/json/recipes/practice/melee.json
+++ b/data/json/recipes/practice/melee.json
@@ -224,6 +224,6 @@
     "practice_data": { "min_difficulty": 0, "max_difficulty": 2, "skill_limit": 3 },
     "autolearn": [ [ "dodge", 1 ] ],
     "book_learn": [ [ "manual_dodge_kid", 0 ], [ "manual_dodge", 0 ] ],
-    "tools": [ [ "basketball", "indoor_volleyball", "beach_volleyball", "baseball" ] ]
+    "tools": [ [ "basketball", "indoor_volleyball", "beach_volleyball", "baseball", "golf_ball" ] ]
   }
 ]


rj8216:
It felt weird that golf balls weren't eligible for this dodge training. They have rubber cores, and the description says the PC throws a ball against a hard surface, so golf balls should count for sure. This is my first time doing this, LMK if everything is alright.

PatrikLundell:
The PR validator fails because the summary doesn't match the format the comment describes. That format is parsed by a script, and so is rigid. The summary is used to automatically add revision history entries (I believe manual processing then removes most of them, but that's after the script has been satisfied). I typically just use None here, as I don't consider my PRs to be revision history entry worthy.

Purpose of change ought to have a line. That can basically be a repeat of the title. Similarly the solution section should have something. I would assume it could be "Added golf balls as valid balls for dodge practice recipe." (without the quotes).

Testing DEFINITELY should have something. Untested things are suspect in general, and doubly so when they're from new people. This doesn't have to be anything extreme, but I'd expect you to say that you've loaded the game and verified that you've been able to use a golf ball for dodge training (and that you've actually done that, not just lied about it, of course). There are various reasons for why "trivial" changes may not work as intended, such as e.g. modifying a similar recipe instead of the one intended (which might result in a check whether both should be modified when verification fails).

rj8216:
Thanks for the help. Sorry had a long day at work, and this is my first time doing this, so it took a hot minute for me to get to this. I appreciate the patience, love this game and just want to help make it better. :-D hopefully I followed directions a little better this time.

PR #60646, opened 2022-09-01
Add electric spinning wheel, recipes, to itemgroups

DPavonis:
#### Summary
Content "Added an electric spinwheel item, alongside relevant recipes, nested recipes and itemgroups."

#### Purpose of change

Allows for nomadic players to enjoy the benefits of a spinning wheel without having to deal with setup and teardown of a stationary spinning wheel.

#### Describe the solution

It's a portable, electricity-powered spinning wheel, 

#### Testing

Make sure the game properly loads, check to see if the new recipes are there with appropriate costs by searching "rope, t:electric" as well as search for, "electric spin"


bombasticSlacks (approved):
Looks good to me,

But I'm no expert on electronics crafting

DPavonis:
> Looks good to me,
> 
> But I'm no expert on electronics crafting

I went with "what seemed reasonable", though I'll fully admit I didn't really math everything out re: weight/density and size. I did do my best to get realistic energy usage for the recipes, but I'm also willing to admit I could very well be wrong, but all the numbers I ran suggested that energy usage for one of these machines would be minimal to the point of some recipes going below 1 battery charge (which as I understand represents 1 kJ)

bombasticSlacks:
The not an electrical guy was more for someone else to also puts some eyes on this before merging.

PR #60471, opened 2022-08-26
Retractable "3D" sprites for walls etc.

mlange-42:
#### Summary
Interface "Retractable '3D' sprites for walls etc."

#### Purpose of change

Upcoming isometric tilesets offer the opportunity for a more 3-dimensional look. For a consistent perspective across terrain, furniture and monsters, walls need to have a certain height. Orthogonal tilesets could also profit from higher walls for a more consistent perspective.

3D walls, however, will obstruct other terrain, etc.:

grafik
*Courtesy tileset: UlticaISO by @vetall812*

This PR tries to resolve this conflict and adds the ability to change/toggle wall height using a mapped key (see GIFs below).

#### Describe the solution

* In addition to the normal offset per spritesheet, add an optional offset for the retracted state
* Add an option (cached) to toggle retracted walls, under debug
* Add an action to make a key bind for the toggle possible
* Adapt gfx tools Python scripts for the new properties

Things to note:

* Retract/toggle hight seems to be a good solution, as it works well for most wall-like tiles and requires very few work by artists
* The retracted offset is optional, and defaults to the normal offset. Existing tilesets should simply do nothing when toggled.
* Artists can specify retracted offset on a per-sheet basis. This allows for great control over how much sprites retract.
* Added the toggle as an option to the debug tab. Move it? Remove it from the options?
* Did not add a default key binding

#### Describe alternatives you've considered

* Use a true/false property for tilesheets and switch between normal and zero offset (less flexible)
* Add the possibility to switch between two tilesets (normal/retracted, a lot of redundency)
* Add the possibility for individual alternative, lowered sprites (increases tileset size)
* Use and alpha threshold to mark sprite regions that should disappear (retracted often looks better, as it shows the top)

#### Testing

* Tested the toggle ingame, with and without key binding
* Tested `compose.py` with UlticaISO (thanks @vetall812)

#### Additional context

GIFs showing the toggle using a key binding.

retracting_walls

retracting_walls_clip
*Courtesy tileset: UlticaISO by @vetall812*

PR #60648, opened 2022-09-01
Fix setting sub-tile retracted offset in multi-tiles

mlange-42:
#### Summary
Bugfixes "Correct multitile retracted height"

#### Purpose of change

A bug in #60471 causes multitile sub-tiles to retract completely and ignore the retracted offset. Fix it!

#### Describe the solution

Correct a copy/paste error

None

#### Testing

Tries in-game with multi-tile ISO walls from UlticaISO (thanks @vetall812 for providing!)

#### Additional context

Screenshot of partially retracted multitile walls:

grafik


mlange-42:
CI tests failing unrelated due to item density tests. See [run 8134215844 step:16:230](https://github.com/CleverRaven/Cataclysm-DDA/runs/8134215844?check_suite_focus=true#step:16:230).

PR #60655, opened 2022-09-01
Aquariums & Glass Cutting

Karol1223:
#### Summary
Content "Adds two new aquariums and glass cutting quality"

#### Purpose of change
In light of recent fish breeding PR, I was asked to take a look at aquariums and potentially make them craftable/able to house fish. Now, honestly neither of these two I actually managed to do but I guess I set some foundations.

#### Describe the solution
- Adds two new aquarium sizes (30 gal and 100 gal)
- Renames current aquarium from 'glass aquarium' to '50 gallon aquarium', makes the weight more accurate, gives it actual price as well as a disassembly recipe (really just smashing it to shards but shhh)
- Adds a new Glass Cutting quality
- Adds a new handheld glass cutter tool
- Adds a new resource: silicone caulk, as well as two squeeze tube variants for it to spawn in
- Makes the aquariums actually craftable!

None right now without using c++.

#### Testing
Doesn't crash, items and recipes seem to be working fine.

#### Additional context
The ideal approach would be to have aquariums _with pumps_ as appliances that while powered can house and keep fish and other small aquatic monsters alive and able to reproduce. However, I believe the suggestion issues about this are already afloat somewhere up there. When/If those actually become implemented I will revisit the aquariums.

ALSO, things to note about the crafting recipes:
- The amount of silicone caulk used is based on the assumption that 300ml of caulk can cover 16 linear meters of surface for this purpose.
- The amount of duct tape used is based on the assumption that a single unit of duct tape is 4cm long (duct tape needs generalising and revisiting anyway)
- The 100 gal recipe involves, with the longest sides, connecting two glass sheets next to each other with caulk, as we do not have glass sheets large enough to fulfil the length for this craft without this.

### DIFF
--- a/data/json/recipes/tools/containers.json
+++ b/data/json/recipes/tools/containers.json
@@ -1152,7 +1152,7 @@
   },
   {
     "type": "recipe",
-    "activity_level": "BRISK_EXERCISE",
+    "activity_level": "LIGHT_EXERCISE",
     "result": "aquarium_medium_drying",
     "category": "CC_OTHER",
     "subcategory": "CSC_OTHER_CONTAINERS",
@@ -1165,7 +1165,7 @@
   },
   {
     "type": "recipe",
-    "activity_level": "BRISK_EXERCISE",
+    "activity_level": "LIGHT_EXERCISE",
     "result": "aquarium_small_drying",
     "category": "CC_OTHER",
     "subcategory": "CSC_OTHER_CONTAINERS",
@@ -1178,7 +1178,7 @@
   },
   {
     "type": "recipe",
-    "activity_level": "BRISK_EXERCISE",
+    "activity_level": "LIGHT_EXERCISE",
     "result": "aquarium_large_drying",
     "category": "CC_OTHER",
     "subcategory": "CSC_OTHER_CONTAINERS",


Karol1223:
Okay so in light of new events I didn't know we had adobe bricks which already do dry out like the silicone here would. Crafting is actually viable. Tomorrow I'll try adding silicone to the game and making all necessary changes to make these crafts work.

anoobindisguise:
Nice. Should probably have a powered version as well and/or a way to craft. Now that we have a dedicated glass cutter and not just repurposing the angle grinder we can probably turn tempered glass and whatnot into some cool new goggles.

Karol1223:
> Nice. Should probably have a powered version as well and/or a way to craft. Now that we have a dedicated glass cutter and not just repurposing the angle grinder we can probably turn tempered glass and whatnot into some cool new goggles.

I considered adding a powered version but to be honest I know barely anything about glass cutting. I don't want to go around reworking the entire mechanic when my knowledge is so limited. I will, however, note this down and take a look at how hard a powered version would be to implement so that at least if someone else wants to work on other glass-cutting recipes they have more room to play with.

Karol1223:
Alright looking through this again, I actually cannot find any electric glass cutters for sale anywhere. The closest there are are electric 'bottle cutters' and I am not sure that they'd be that useful in the exact same regard (i.e. anywhere but glass bottles)

data/json/items/containers.json
```diff
@@ -2696,12 +2696,13 @@
     "id": "aquarium_medium",
     "type": "GENERIC",
     "category": "container",
-    "name": { "str": "glass aquarium" },
-    "description": "A 50 gallon glass aquarium, used for safely storing small animals for use as pets and food.",
-    "weight": "83 kg",
+    "name": { "str": "50 gallon aquarium" },
```
phoeo, line 2699: Is this consistent with container naming? AFAIK other containers list their volume in liters (3L jar, steel drum (100L), vehicle tank (60L))

Zireael07: IIRC at least one drum is listed as <amount> gallon drum

Karol1223: No, no, no. The drums reference gallons in the IDs whereas in their names they are litres. I went with American sizes for Aquariums as the game is set in New England and, sure, I could instead call it a 189-litre aquarium but it doesn't roll off the tongue the same way.

Karol1223: Okay after asking around in the discord it appears that this shouldn't be an issue as that is exactly what the aquariums are named in New England. American aquariums reference gallons in their names so these, which are also American, referencing gallons should not be a problem.

data/json/items/containers.json
```diff
@@ -2718,6 +2719,172 @@
...
+      }
+    ],
+    "qualities": [ [ "CONTAIN", 1 ] ],
+    "use_action": {
+      "target": "aquarium_medium",
+      "msg": "You tear off the tape from the silicone, looks like the aquarium is ready.",
+      "moves": 100,
+      "type": "delayed_transform",
+      "transform_age": 604800,
+      "not_ready_msg": "The silicone hasn't dried quite yet.",
```
phoeo, line 2754: I think it's better to give amount of time until seams dry. It could get annoying if player forgets when they made the aquarium, and the number is arbitrary anyway.

data/json/items/containers.json
```diff
@@ -2718,6 +2719,172 @@
...
+        "watertight": false,
+        "open_container": true,
+        "max_contains_volume": "189 L",
+        "max_contains_weight": "300 kg"
+      }
+    ],
+    "qualities": [ [ "CONTAIN", 1 ] ],
+    "use_action": {
+      "target": "aquarium_medium",
+      "msg": "You tear off the tape from the silicone, looks like the aquarium is ready.",
```
phoeo, line 2750: Aquariums are checked by filling them with water, not just by examining the seams. Also, what tape?

data/json/items/containers.json
Karol1223: I am not aware of any way to do this. The only other item that I know has a similar transformation code is the soft adobe brick which does not reference this either way. I also wouldn't call the number arbitrary. In the conditions of 25 celsius and 50% air humidity, it should take 24 hours to dry and a week to achieve maximum adhesion. This number is a week.

data/json/items/containers.json
Karol1223: You are right, that is a way that they should be checked. However, I am not able to code it in such a way that the player has to fill the aquarium with water every time they check whether it is ready or not. The tape is duct tape used in the crafting recipes, I was not going to add a new type of tape solely for aquarium crafting.

data/json/recipes/tools/containers.json
```diff
@@ -1149,5 +1149,44 @@
     "proficiencies": [ { "proficiency": "prof_metalworking" } ],
     "qualities": [ { "id": "CHISEL", "level": 3 } ],
     "components": [ [ [ "scrap_aluminum", 120 ], [ "material_aluminium_ingot", 10 ] ] ]
+  },
+  {
+    "type": "recipe",
+    "activity_level": "BRISK_EXERCISE",
```
Fris0uman, line 1155: Shouldn't it be light execise to glue the glass sheets together?

Karol1223: This recipe also includes cutting the glass sheets into appropriate sizes, though, hence it being Brisk. Having said that, that may be overkill for glass cutting. I have not done it myself so I am more than open to editing that accordingly.

irwiss: https://youtu.be/gfmRLTxVFmg?t=103

Unless fish tanks get some bulletproof glass instead of the sheets you buy in home improvement stores that doesn't look similar to blacksmithing levels of exertion

Karol1223: You are right, I'll change this to light exercise.

PR #60660, opened 2022-09-01
Attempt to audit power costs of welding

anoobindisguise:
#### Summary
Balance "welding power costs increase"

#### Purpose of change
Welding was also using too little power.

#### Describe the solution
Assumption 1: MIG welder uses 2-3 kW while welding.

Assumption 2: A skilled welder can weld about 140 inches per hour, or 360 cm. This number is pretty important and if there are any welders here I'd appreciate a second look at this. I've seen "140 per hour" in multiple sources but I've also seen wildly different numbers in many places. Not only will this determine the time it takes to craft welded items, but it will also determine the power usage. If the number is really only 70, power usage doubles and craft time doubles.

Assumption 3: Welders don't spend 100% of the time welding, only roughly 33%-50%, the rest is spent arranging metal and letting welds cool.

Assumption 4: Acetylene gas is theoretically worth roughly 50 charges of battery, at an energy density of about 50kJ per gram of gas. Thermal efficiency is lower than this in practice at about 40%. compared to about 85% of arc welding, so in practice each unit of acetylene is worth ~23 units of battery.

The in-game welding gas tanks are wrong. Firstly they use a nonsensical acetylene/oxygen mixture which is insanity, this would instantly explode. It is a workaround for Multimag but the tanks should be done right first. As such I've rebalanced them to represent real welding tanks (MC tank and B tank) which contain, respectively, 10 and 40 cubic feet of acetylene (measured at STP). A cubic foot of acetylene weighs about 33 grams, which gives the tanks 330/1320 charges. Increased length, volume, and longest_size to match product descriptions. I also added oxygen cylinders to garages and made them actually contain oxygen (they were missing a magazine). They currently do not have any use, but once we have multimag they can be drawn from just like the acetylene canisters. The acetylene-from-calc-carbide recipe is also fucked but I have no clue how to fix it.

This does not touch any of the existing weld recipes in game. I'm assuming roughly 10 seconds for 1cm of weld per unit of the requirement, at 20kJ of arc welder battery, or 1 charge of acetylene. 
Therefore a recipe that requires `[ "welding_standard", 20 ]` would represent a 60cm weld, add 10 minutes to the craft time and consume 60 welding consumable and 1200kJ of battery or 60 charges of acetylene. 
If you were going to weld a 1mx1m square metal armor plate onto a car and joined all 4 sides along their length, this weld would cost 400 welding consumable (each charge represents 1cm), take about 67 minutes, and take 8000 charges of battery or 400 charges of acetylene, although these would be lower if you aren't joining literally every part of all 4 edges of the sheet of metal. 
The real costs will come from the uses of welding_standard being audited. As of right now this is only a modest increase to power costs and in the case of acetylene is actually a buff, but that's because the acetylene welds will be generating magic oxygen until we get multimag. At which point you will need to pressurize oxygen cylinders to keep your welder supplied.

#### Additional context

https://weldgears.com/how-much-electricity-does-a-mig-welder-use/
http://www.planningplanet.com/wiki/507732/pipeline-construction
https://www.engineeringtoolbox.com/acetylene-d_1412.html
https://www.aqua-calc.com/calculate/volume-to-weight


anoobindisguise:
This is "ready to go", but it will require us to collectively decide for each welding recipe and vehicle install how many cm of weld is needed.

AcidAntOnAMinefield:
Seeing this was "just" merged, I have to point out that after updating the game and reloading my save file, my tanks were altered:

The previously 60/60 tanks are now 60/350 and the 240/240 are now 240/1400. Meaning that the maximum capacity was, indeed, increased, but unfortunately my tanks did not remain full once they were updated. This might be worth looking into.

Sumlord-047:
Maybe the cost of building recipes and cutting metal should be adjusted.

PR #60662, opened 2022-09-01
Rewrite bionic fuel system.

Hirmuolio:
#### Summary
Content "Rewrite bionic fuel system."

#### Purpose of change

Bionic fuel system currently works by saving {"material_id", "quantity"} pairs as variables into player data when player eats fuel for the bionics. (Yes the fuel is still handled as being eaten!)

This is not a good way to do it. The biggest problem is that it uses material energy attribute wrong. The material energy is in energy per volume. But there is no volume info available.

To use material energy attribute correctly the bionic system must work with items.

#### Describe the solution

Bionic fuel exists as items. 
It was supposed to be simple to switch where bionics look for their fuel. But it turns out that the bionic code was very "side effect oriented". One function would set the fuels in player variables then another function would read the fuels from player variables. Instead of trying to refactor that mess I ended up rewriting everything...

All items with `"BIONIC_FUEL_SOURCE"` will have their contents available as fuel for fuel powered bionics.

Fuel powered bionics now install pseudo item container for the fuel. This container is visible in worn equipment window.

Player refuels the bionic by placing fuel into the appropriate fuel tank pseudo item.

Added ability to limit pocket contents by item material. So you can only put items made out of gasoline into gasoline tank.

Battery system bionic can use medium and light batteries. This does lower its max stored charges a bit (used to be 2000 which is about same as heavy batteries). But the more convenient reload action should compensate.

Also simplified the autostart and fuel saving systems.
Autostart system has been removed. (it used to turn fueled bionic on if power level was below treshold).
Reworked fuel saving (it used to turn fueled bionic off if power was above treshold).

Now the bionic will simply not consume any fuel unless power is below the treshold.
Fuel powered bionics will behave as if they were "off" if power is above treshold. (muscle bionics slow you down less when off).
Bionic that is "ON" but not active due to fuel saving will show "(inactive)" in the bionic menu.

Bionic tanks visible in inventory. You can insert/remove items to/from them just like any other container:
inventory

Bionic window is basically the same as before:
bionicnew

Known issue: NPCs do not refill their bionic tanks. NPCs do not reload their magazines either and I do not really know enough about the NPC code to quickly fix.
Probably best behavior would be to every now and then when out of combat see if any container with `"BIONIC_FUEL_SOURCE"` flag can be refilled and refill them.

#### Testing

Gasoline fuel cell works (8 kJ/s).
Oil generator works (9 kJ/5s).
Cable charger works when connected to UPS, solar backpack and vehicle.

Fuel can be reloaded and unloaded.
Running out of fuel turns bionic off.
Saving and loading works.
Loading old saves migrates properly. Even the charges are migrated!

### DIFF
--- a/src/bionics.cpp
+++ b/src/bionics.cpp
@@ -1646,10 +1646,14 @@ static bool attempt_recharge( Character &p, bionic &bio, units::energy &amount )
 
 void Character::process_bionic( bionic &bio )
 {
-    bionic_fuels result = bionic_fuel_check( bio, false );
+    bionic_fuels result;
 
     // Only powered bionics should be processed
     if( !bio.powered ) {
+        if( get_effective_efficiency( bio, bio.info().passive_fuel_efficiency ) == 0 ) {
+            return;
+        }
+        result = bionic_fuel_check( bio, false );
         burn_fuel( bio, result );
         return;
     }
@@ -1678,6 +1682,7 @@ void Character::process_bionic( bionic &bio )
 
     bio.charge_timer = std::max( 0_turns, bio.charge_timer - discharge_rate );
     if( bio.charge_timer <= 0_turns ) {
+        result = bionic_fuel_check( bio, false );
         if( bio.info().charge_time > 0_turns ) {
             if( bio.info().has_flag( STATIC( json_character_flag( "BIONIC_POWER_SOURCE" ) ) ) ) {
                 // Convert fuel to bionic power


anoobindisguise:
Hurrah, another use for the integrated bionic armor system! This is a pretty elegant implementation of something I did wonder about when I was working on that.

chrispikula:
So, just to be clear, there's no way to set priority for fuel usage anymore, based on the starting thresholds?  Some of the fuel sources are much worse than others, such as the battery vs ethanol burner making you very hot vs metabolic interchange making your weight plummet.  Staggering there turn ons based on character need was a nice feature.

Likewise, for those of us using magiclysm, staying between 30-50% of full bionic charge was occasionally useful*, and could be done with the auto start/auto turn off of charging systems.  Is this still doable?

*you get max mana penalized based on how much charge you have stored.

Hirmuolio:
You can still set the fuel saving treshold % for each bionic.
So if you want the bionic to stop generating power at 50% just set the fuel saving treshold to 50%.

chrispikula:
> You can still set the fuel saving treshold % for each bionic. So if you want the bionic to stop generating power at 50% just set the fuel saving treshold to 50%.

Yeah, but you need to remember to manually turn them back on in the bionics menu each time, if I'm understanding this right?  No more 'set and forget'.  If it helps clarify things, I was using the autostart/power saving modes as a sort of hysteresis band, so that bionic power could stay within the hysteresis band without wide swings due to forgetfulness of micromanagement.

Hirmuolio:
> remember to manually turn them back on 

Not really since auto turn off does not exist anymore. With multiple bionics it would be more of a tiered system.

For example you could have something like this: Battery bionic >80%, gasoline >50%.

Start at 85% power. All are on. None of them generates power.  
Drop down to 60%. Battery starts generating power.  
Drop down to 40%. Gasoline and battery generate power.   
Jump up tp 70%. Gasoline stops generating power but stays on.
Drop back to 40%. Gasoline starts generating power again.

At no point does any of the bionics actually turn off. They are on but if the power level is above the treshold they just do nothing.

data/mods/Magiclysm/bionics.json
```diff
@@ -6,10 +6,35 @@
     "description": "Embedded into your back and connected to your bionic power supply is a powerful spell focus able to consume blood from dead creatures to produce bionic power.  It can store up to 100 mL of blood.",
     "occupied_bodyparts": [ [ "torso", 8 ] ],
     "fuel_options": [ "blood", "dragon_blood", "tainted_blood", "hblood" ],
-    "fuel_capacity": 100,
     "fuel_efficiency": 0.6,
     "exothermic_power_gen": true,
     "time": "6 s",
-    "flags": [ "BIONIC_TOGGLED", "BIONIC_POWER_SOURCE" ]
+    "flags": [ "BIONIC_TOGGLED", "BIONIC_POWER_SOURCE" ],
+    "passive_pseudo_items": [ "integrated_blod_bank" ]
```
Fris0uman, line 13:
```suggestion
    "passive_pseudo_items": [ "integrated_blood_bank" ]
```

data/mods/Magiclysm/bionics.json
```diff
@@ -6,10 +6,35 @@
...
+  },
+  {
+    "id": "integrated_blod_bank",
```
Fris0uman, line 16:
```suggestion
    "id": "integrated_blood_bank",
```
Spelling is not critical but this might spare some confusion

data/json/bionics.json
```diff
@@ -465,9 +465,9 @@
     "occupied_bodyparts": [ [ "torso", 20 ] ],
     "fuel_options": [ "battery" ],
     "fuel_efficiency": 1,
-    "fuel_capacity": 2000,
     "time": "1 s",
-    "flags": [ "BIONIC_POWER_SOURCE", "BIONIC_TOGGLED", "BIONIC_NPC_USABLE" ]
+    "flags": [ "BIONIC_POWER_SOURCE", "BIONIC_TOGGLED", "BIONIC_NPC_USABLE" ],
```
Fris0uman, line 469:
```suggestion
    "flags": [ "BIONIC_POWER_SOURCE", "BIONIC_TOGGLED" ],
```
Maybe those CBMs should lose the flag since NPCs can't reload them now?

data/json/bionics.json
```diff
@@ -780,11 +780,11 @@
     "description": "You burn alcohol as fuel in an extremely efficient reaction.",
     "occupied_bodyparts": [ [ "torso", 20 ] ],
     "fuel_options": [ "alcohol" ],
-    "fuel_capacity": 400,
     "fuel_efficiency": 0.5,
     "exothermic_power_gen": true,
     "time": "1 s",
-    "flags": [ "BIONIC_POWER_SOURCE", "BIONIC_NPC_USABLE", "BIONIC_SHOCKPROOF", "BIONIC_TOGGLED" ]
+    "flags": [ "BIONIC_POWER_SOURCE", "BIONIC_NPC_USABLE", "BIONIC_SHOCKPROOF", "BIONIC_TOGGLED" ],
```
Fris0uman, line 786:
```suggestion
    "flags": [ "BIONIC_POWER_SOURCE", "BIONIC_SHOCKPROOF", "BIONIC_TOGGLED" ],
```

data/json/bionics.json
```diff
@@ -793,11 +793,11 @@
     "description": "Slowly generates bionic power out of kerosene or motor oil.",
     "occupied_bodyparts": [ [ "torso", 20 ] ],
     "fuel_options": [ "lamp_oil", "motor_oil" ],
-    "fuel_capacity": 750,
     "fuel_efficiency": 0.35,
     "exothermic_power_gen": true,
     "time": "5 s",
-    "flags": [ "BIONIC_POWER_SOURCE", "BIONIC_NPC_USABLE", "BIONIC_SHOCKPROOF", "BIONIC_TOGGLED" ]
+    "flags": [ "BIONIC_POWER_SOURCE", "BIONIC_NPC_USABLE", "BIONIC_SHOCKPROOF", "BIONIC_TOGGLED" ],
```
Fris0uman, line 799:
```suggestion
    "flags": [ "BIONIC_POWER_SOURCE", "BIONIC_SHOCKPROOF", "BIONIC_TOGGLED" ],
```

data/json/bionics.json
```diff
@@ -1674,11 +1674,11 @@
     "occupied_bodyparts": [ [ "torso", 8 ] ],
     "encumbrance": [ [ "torso", 5 ] ],
     "fuel_options": [ "gasoline" ],
-    "fuel_capacity": 500,
     "fuel_efficiency": 0.25,
     "exothermic_power_gen": true,
     "time": "1 s",
-    "flags": [ "BIONIC_TOGGLED", "BIONIC_POWER_SOURCE", "BIONIC_NPC_USABLE" ]
+    "flags": [ "BIONIC_TOGGLED", "BIONIC_POWER_SOURCE", "BIONIC_NPC_USABLE" ],
```
Fris0uman, line 1680:
```suggestion
    "flags": [ "BIONIC_TOGGLED", "BIONIC_POWER_SOURCE" ],
```

Fris0uman (commented):
If I understand the way fuel management works now, there's no reason to ever turn off the powergen bionics so maybe they should just be "passive" and don't need to be toggleable?
Somewhat related to that can we get something in the bionic menu to tell if a powergen CBM is currently producing power or not? Like a chnage in color or some symbol next to the name

Hirmuolio:
Yes there is not much reason to turning off these types of bionics anymore. Setting the treshold to 0% is effectively same as turning them completely off. But I'd still keep the turn off mechanic for them to be consistent with other bionics. (also the current UI does not allow setting the limit to 0%)

Adding a power indicator for the bionic should be possible. Maybe a green "+pwr" at the end?

Hirmuolio:
Uh oh. I completely messed up the branch. I'll make a new pull request from copy of the non-messed branch.

PR #60666, opened 2022-09-01
Curtains only indoors and street light over the road! -- Auto terrain and furniture rotation.

### LINKED ISSUE #51619: add "rotates_toward" attribute for furniture/terrain
# Is your feature request related to a problem? Please describe.
We're having a tough time drawing street lights with the normal L shape, as they can only face one direction in the code.

# Describe the solution you'd like
Furniture currently has a nice dynamic rotation script that makes chairs face away from walls.

I would like the option to specify `rotates_toward` in a furniture or terrain object's JSON, with the accepted argument being a current `connects_to` group. This would override any other default rotation behaviour. For example:

```
"rotates_toward": [ "PAVEMENT" ]
```
--> causes a street light to face the nearest pavement.

```
"rotates_toward": [ "WATER" ]
```
--> could be used for dockside furniture and terrain

## Complications
There are a few things I would like to be able to call here that are not connects_to groups and couldn't easily be. The main one is indoor floors. It would be nice if windows and doors could rotate to face indoor floors, because then we could put the blinds/shutters/boards on the correct side of the window. I am unaware of any other times we might want to do that but couldn't use a connects_to group.

# Describe alternatives you've considered
Other ways of specifying rotation and connection, such as explicitly listing terrain a furniture should rotate towards

# Additional context
If we could combine rotation with multitiles, we'd be able to do some other cool effects.

mlange-42:
#### Summary
Interface "Curtains only indoors and street light over the road! -- Auto terrain and furniture rotation."

#### Purpose of change

Allow terain and furniture to rotate towards certain terrain. E.g. on which side is the curtain?

Make it compatible with existing `connects_to`.

Fixes #51619

#### Describe the solution

The solution is similar to the `connects_to` mechanic. However, unlike there, it should not be bi-directional/symmetric.
In analogy to `map_data_common_t::connect_group` (JSON: `connects_to`).

For usage docs, see [TILESET.md#rotates_to](https://github.com/mlange-42/Cataclysm-DDA/blob/terrain-rotate-towards/doc/TILESET.md#auto-rotating-terrain-and-furniture---rotates_to).

* [x] Add `map_data_common_t::rotate_to_group` and `map_data_common_t::rotate_to_group_member`
   * [x] JSON: `rotates_to` and `rotates_to_member`
   * [x] 1st is for the active part (center), 2nd for the passive part (neighbors)
* [x] Auto-set `rotate_to_group` for flags `WINDOW` and `DOOR`
* [x] Auto-set `rotate_to_group_member` for flag `INDOORS`
* [x] Hook into [`cata_tiles::get_rotation_and_subtile()`](https://github.com/CleverRaven/Cataclysm-DDA/blob/d8ef623a842b84491983eae8a3570960f106b923/src/cata_tiles.cpp#L4270)
   * [x] Add argument for bitmask of `rotates_to` neighbors
   * [x] Calculate a rotation in range 0...7 for `edge` and `end_piece` multitile type. rotation index represents edge direction AND inside/outside combination
   * [x] Calculate a rotation in range 0...15 for `unconnected` multitile type.
* [x] Add connect group `INDOOR_FLOOR`
   * Set as `rotates_to` for flags `WINDOW` and `DOOR`
   * Set as `rotates_to_member` for flag `INDOORS`
   * Makes all windows and doors use the feature without further changes
* [x] Make `unconnected` work with the example of `f_street_light` rotate towards `t_pavement`
* [x] Add tests for `connects_to` 
* [x] Add tests for `rotates_to` 
* [x] Document the new feature (rendered: [TILESET.md#rotates_to](https://github.com/mlange-42/Cataclysm-DDA/blob/terrain-rotate-towards/doc/TILESET.md#auto-rotating-terrain-and-furniture---rotates_to))
* [x] Document existing but rarely documented `connects_to` (rendered: [TILESET.md#connects_to](https://github.com/mlange-42/Cataclysm-DDA/blob/terrain-rotate-towards/doc/TILESET.md#connecting-terrain-and-furniture---connects_to))

None

#### Testing

Tried ingame (see screenshouts)

Added tests for the id and rotation calculations using `connects_to` as well as `rotates_to`. See [tests/connect_rotate_test.cpp](https://github.com/CleverRaven/Cataclysm-DDA/blob/4e679b488a0badc4443bfb924fafdf8a6a4eedb0/tests/connect_rotate_test.cpp).

#### Additional context

Things to note:

* Implemented for terrain, and for furniture rotating towards terrain (currently not towards furniture!)
* Implemented only for `edge`, `end_piece` and `unconnected` elements
   * Works for doors and windows out of the box
   * Works for street lights (furniture)
   * `end_piece` primarily to show windows with one obscured wall correctly
* Terrain analyzes surrounding terrain only
* Furniture analyzes surrounding terrain and furniture
* Uses all neighbors, not only known. Reasoning is that the orientation of the particular terrain/furniture is visible for the player avatar
* Due to auto-setting the new properties based on `WINDOW`, `DOOR` and `INDOORS`, it should work out of the box for all doors and windows in combination with indoor floor. Simply define the multitile
* Does not break existing tilesets that use only 2 rotation values for doors and windows
* For `unconnected`, works well with specifying only 4 instead of 16 rotation sprites

Test sprites / required sprites example (*UlticaISO (WIP), thanks @vetall812*):

sprites

All possible combinations of window direction and inside/outside:

grafik

All possible combinations of street light / pavement relations:

grafik

### DIFF
--- a/doc/TILESET.md
+++ b/doc/TILESET.md
@@ -137,6 +137,8 @@ For terrain or furniture that is intended to auto-connect using multitiles, set 
 }
 ```
 
+For details, see [JSON_INFO.md](./JSON_INFO.md#connects_to).
+
 Connections are only set up between types that have the same `connects_to` group, so it is symmetric or bi-directional.
 Available groups are:
 
@@ -210,6 +212,7 @@ In JSON, the multitile would be defined like this:
 #### Auto-rotating terrain and furniture - `rotates_to`
 
 Terrain and furniture can auto-rotate depending on other surrounding terrain or furniture using `rotates_to`.
+For details, see [JSON_INFO.md](./JSON_INFO.md#rotates_to-and-rotates_to_member).
 Usage examples for terrain are doors and windows that look differently, seen from inside and outside (e.g. curtain).
 An example for furniture are street lights that orient towards the pavement.
 
--- a/doc/JSON_INFO.md
+++ b/doc/JSON_INFO.md
@@ -185,6 +185,7 @@ Use the `Home` key to return to the top.
       - [`name`](#name-1)
       - [`flags`](#flags-1)
       - [`connects_to`](#connects_to)
+      - [`rotates_to` and `rotates_to_member`](#rotates_to-and-rotates_to_member)
       - [`symbol`](#symbol)
       - [`comfort`](#comfort)
       - [`floor_bedding_warmth`](#floor_bedding_warmth)
@@ -4560,6 +4561,8 @@ Strength required to move the furniture around. Negative values indicate an unmo
     "max_volume": "1000 L",
     "flags": ["TRANSPARENT", "DIGGABLE"],
     "connects_to" : "WALL",
+    "rotates_to" : "INDOORFLOOR",
+    "rotates_to_member" : "INDOORFLOOR",
     "close": "t_foo_closed",
     "open": "t_foo_open",
     "lockpick_result": "t_door_unlocked",
@@ -4768,17 +4771,24 @@ Displayed name of the object. This will be translated.
 
 (Optional) The group of terrains to which this terrain connects. This affects tile rotation and connections, and the ASCII symbol drawn by terrain with the flag "AUTO_WALL_SYMBOL".
 
-Current values:
-- `WALL`
-- `CHAINFENCE`
-- `WOODFENCE`
-- `RAILING`
-- `WATER`
-- `POOLWATER`
-- `PAVEMENT`
-- `RAIL`
+Available groups are:
 
+##### Connect groups
 
+```
+NONE                 PIT_DEEP
+WALL                 LINOLEUM
+CHAINFENCE           CARPET
+WOODFENCE            CONCRETE
+RAILING              CLAY
+POOLWATER            DIRT
+WATER                ROCKFLOOR
+PAVEMENT             MULCHFLOOR
+RAIL                 METALFLOOR
+COUNTER              WOODFLOOR
+CANVAS_WALL          INDOORFLOOR
+SAND
+```
 
 Example: `-` , `|` , `X` and `Y` are terrain which share the same `connects_to` value. `O` does not have it. `X` and `Y` also have the `AUTO_WALL_SYMBOL` flag. `X` will be drawn as a T-intersection (connected to west, south and east), `Y` will be drawn as a horizontal line (going from west to east, no connection to south).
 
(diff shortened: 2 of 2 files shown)


src/cata_tiles.h
```diff
@@ -725,4 +736,13 @@ class cata_tiles
         std::string memory_map_mode = "color_pixel_sepia";
 };
 
+struct cata_tiles_test_helper {
```
anothersimulacrum, line 739: The definition for this should just be able to go in the tests, if you want.

mlange-42: And it will be found to make it a friend class of `cata_tiles`? Would have thought not.

anothersimulacrum: Yep. I would have thought you need to forward declare it, but not even that. I can't find any source to back it up, but it works for me locally.
All I can find is 
> The name of the class that is used in this friend declaration does not need to be previously declared.

From https://en.cppreference.com/w/cpp/language/friend

mlange-42: Ah, cool! Thanks! Will do...

mlange-42: Done.

dseguin (approved):
LGTM, but someone more familiar with the tiles code might need to weigh in.

doc/TILESET.md
```diff
@@ -114,6 +123,221 @@ You can add `"rotates": true` to allow sprites to be rotated by the game automat
 
 `"multitile": true` signifies that there is an `additional_tiles` object (redundant? [probably](https://github.com/CleverRaven/Cataclysm-DDA/issues/46253)) with one or more objects that define sprites for game entities associated with this tile, such as broken versions of an item, or wall connections.  Each object in the array has an `id` field, as above, and an `fg` field, which can be a single [root name](#root-name), an array of root names, or an array of objects as above. `"rotates": true` is implied with it and can be omitted.
 
+#### Connecting terrain and furniture - `connects_to`
+
+For terrain or furniture that is intended to auto-connect using multitiles, set the property `connects_to` in the object definition (not in the tileset!) to an appropriate group:
+
+```json
+{
+    "type": "terrain",
```
Fris0uman, line 132: This should probably go into <https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md#terrain> since it's not a tileset property

mlange-42: Thanks! Will add it there, but would prefer to leave a a bit of information also in the tileset docs.

mlange-42: Done.

PR #60668, opened 2022-09-01
B&T APC9 PRO K bug fixes 

DPavonis:
#### Summary
Bugfixes "Various bugfixes for the APC9"

#### Purpose of change

To bring the B&T APC9 Pro K in line with its real-world counterpart. At current, its ingame counterpart lacks a location for the user to install sights on (its default red dot sight notwithstanding), and uses "underbarrel mount" and "rail mount" which from my understanding are mod locations intended for older firearms that lack any sort of modern accessory mounting system for the given location.

#### Describe the solution

Added "sights" to "valid_mod_locations"
Changed "rail mount" to "rail"
Changed "underbarrel mount" to "underbarrel"

#### Testing

Boot up game and make sure it has the appropriate mod locations.

#### Additional context

Attached is an image of the APC9 in what I understand as its most likely configuration as a military weapon.

BT-36090-2_big-660x426

### DIFF
--- a/data/json/items/gun/9mm.json
+++ b/data/json/items/gun/9mm.json
@@ -1320,7 +1320,7 @@
       [ "stock", 1 ],
       [ "stock accessory", 1 ]
     ],
-    "default_mods": [ "folding_stock", "red_dot_sight" ],
+    "default_mods": [ "wire_stock", "red_dot_sight" ],
     "pocket_data": [
       {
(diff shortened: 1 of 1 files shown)


DPavonis:
Unless I'm reading the failure message wrong, it looks like the failure is in an unrelated test regarding power cable stretch disconnect

Issue #60157, opened 2022-08-14
Can't seem to level unarmed beyond 6


anoobindisguise:
### Describe the bug

I've punched a *lot* of stuff, from regular zombies to wasps to experimental mutants to bio-ops, and I've done the punching with a variety of setups - armored gauntlets, brass knuckles, bare fists for example. Yet I can't seem to level unarmed beyond 6 no matter what I do, everything is deemed too trivial to train the skill beyond 6.

### Steps to reproduce

1. Set skill to 6.
2. Punch various baddies.
3. Observe skill not increasing.

### Expected behavior

Skill should increase beyond 6.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.1889 (21H2)
- Game Version: 0471811 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Bionic Slots [cbm_slots]
]

buffog:
I have the same problem.

PR #60670, opened 2022-09-01
Update get_melee_hit_base to account for unarmed attacks having melee…

### LINKED ISSUE #60369: Wearing unarmed weapons is not counted as a weapon in calculations and does not get melee skill bonus.
### Describe the bug

Now that we can wear knuckles directly, fighting unarmed while wearing the unarmed weapons will mean that your character will suffer in terms of hit chance and crit chance compared to previous calculations when you wielded unarmed weapons.

### Steps to reproduce

1. Debug or use a character with 10 melee skill and some unarmed skill.
2. Wield pair of scrap knuckles
3. Notice that you hit pretty frequently.
4. Wear the scrap knuckles instead
5. Notice that you're very unlikely to hit compared to before.

### Expected behavior

There should not be a difference in the hit chance depending on if you wield unarmed weapons or if you wear them.

### Screenshots

I think the culprit for hit chance is this:
image

And I think the culprit for hit/crit chance is this:
image

- OS: Windows
    - OS Version: 10.0.22000.856 (21H2)
- Game Version: 0.F-10794-ga2991c7cef-dirty [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Slots [cbm_slots]
]

### Additional context

_No response_

jim-huynh:
… skill bonus.

#### Summary
Bugfixes "Wearing unarmed weapons is not counted as a weapon in calculations and does not get melee skill bonus."

#### Purpose of change
Fixes #60369

#### Describe the solution
I updated get_melee_hit_base to use a similar method for checking weapon usage that was added to account for worn weapons or body slots.  If that is null, then the unarmed skill is used.

I've also updated melee_skill in item.cpp so that the unarmed weapon check goes before the melee check.

N/A - the recent changes made it so that worn unarmed weapons never use the melee skill in the attack calculations.

Do item checking, but that solution would explicitly exclude plain fists with no armor which seems odd.
#### Testing
I've been playing with this fix for a week or so and no crashes so far.  When stepping through the hit calculations, I get the +10 melee skill that my character has.

#### Additional context
Unarmed is still rather weak due to weak raw damage.  It is slightly balanced by the fact that unarmed attacks act roughly equivalent to a +5 weapon.  Previously without the melee skill addition your attack rolls would be the same whether you were 0 melee skill or 10 melee skill.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -10497,12 +10497,12 @@ gun_type_type item::gun_type() const
 
 skill_id item::melee_skill() const
 {
-    if( !is_melee() ) {
-        return skill_id::NULL_ID();
+    if (is_unarmed_weapon()) {
+        return skill_unarmed;
     }
 
-    if( has_flag( flag_UNARMED_WEAPON ) ) {
-        return skill_unarmed;
+    if( !is_melee() ) {
+        return skill_id::NULL_ID();
     }
 
     int hi = 0;
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -10497,7 +10497,7 @@ gun_type_type item::gun_type() const
 
 skill_id item::melee_skill() const
 {
-    if (is_unarmed_weapon()) {
+    if( is_unarmed_weapon() ) {
         return skill_unarmed;
     }
 
--- a/src/melee.cpp
+++ b/src/melee.cpp
@@ -345,9 +345,11 @@ float Character::get_hit_weapon( const item &weap ) const
 float Character::get_melee_hit_base() const
 {
     float hit_weapon = 0.0f;
-    if( used_weapon() ) {
-        hit_weapon = get_hit_weapon( *used_weapon() );
-    }
+
+    item_location cur_weapon = used_weapon();
+    item cur_weap = cur_weapon ? *cur_weapon : null_item_reference();
+
+    hit_weapon = get_hit_weapon(cur_weap);
 
     // Character::get_hit_base includes stat calculations already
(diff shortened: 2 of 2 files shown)


jim-huynh:
This does not fix the issue of knuckle weapons showing an incorrect low crit rate due to the calculations not knowing that the weapon is unarmed.
image

Scrap knuckles which still has the UNARMED_WEAPON flag and shows a much higher crit rate.
image

Both will result in the same crit rate after this merge, but the display difference remains.

AJimmyU:
Does this also fix https://github.com/CleverRaven/Cataclysm-DDA/issues/60157 or is it a different issue causing it?

jim-huynh:
> Does this also fix #60157 or is it a different issue causing it?

It should not.  The block that trains melee skills is in melee.cpp but separate from the bits I changed.

image

You gain exp 2x-2.5x faster when you hit so this indirectly helps that, but it doesn't resolve the issue if there is a straight up training cap.  I think the problem might be that unarmed trains at (1 * exp range) whereas other weapons train at (weapon damage * exp range).  So a katana would gain (36 total damage * exp range) total exp, with 31/36 of that being cut, whereas hitting someone with your fist would gain you (1 * exp range).  Seems like you'd have to do magnitudes more swings to get any exp at higher levels.

If you're also having that issue, could you lock every melee skill except for unarmed and see if you lose focus when you attempt to hit something?  If you do then it means that you're training the skill even if it's a very small amount.

PR #60675, opened 2022-09-02
Make wooden wheels attachable without a wrench

reed501:
#### Summary
Balance "Make wooden wheels attachable without a wrench"

#### Purpose of change

Currently if you're playing innawoods there is no way to attach any wheels to vehicles to make a cart without a tool with bolt turning 1. Obviously there were wooden carts before there were wrenches, however.

#### Describe the solution

For now I'm just going to make it possible using a hammer and nails instead of a wrench. Neither really makes sense, just use a wrench on a wheel and a box and they're attached! Voila! In reality it needs to be mounted to an axle that needs attachment to the cart. I also required a light wheel mount (implied that it'll be a wooden mount, which currently doesn't require a wrench) to be a little more realistic I guess. At least now you can make a cart without iron.

I think wheels and wheel mounts need an audit. From what I can see, none of them use an axle. Where are the axles? How are these wheels spinning? Probably too much during feature freeze, so for now here's a stopgap.

In the future: Is axle a new part? Built into frames? Built into wheel mounts? How would that work unless you had to install them in pairs? Shouldn't this cart tip over with just a box and a pair of wheels? I would like to think about these things, but later.

#### Testing

None yet, I want to get opinions first.

### DIFF
--- a/data/json/vehicleparts/wheel.json
+++ b/data/json/vehicleparts/wheel.json
@@ -672,14 +672,14 @@
     "damage_modifier": 50,
     "breaks_into": [ { "item": "2x4", "count": [ 0, 4 ] }, { "item": "nail", "charges": [ 5, 15 ] } ],
     "requirements": {
-      "install": { "skills": [ [ "mechanics", 1 ] ], "time": "5 m", "using": [ [ "vehicle_wrench_1", 1 ] ] },
-      "removal": { "skills": [ [ "mechanics", 1 ] ], "time": "5 m", "using": [ [ "vehicle_wrench_1", 1 ] ] },
+      "install": { "skills": [ [ "mechanics", 1 ] ], "time": "5 m", "using": [ [ "vehicle_nail_install", 1 ] ] },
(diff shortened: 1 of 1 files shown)


PatrikLundell:
I'd say the axle is implied by the wheel mounts in a hand wavey way. An axle doesn't really fit the tile based vehicle construction model, as it's something that spans multiple tiles.

You don't actually need to install wheels in pairs. Firstly, there are three wheeled vehicles, and secondly you don't really need wheels on opposite sides of a vehicle to be connected to each other via an axle (shopping cart castor wheels aren't, for instance, and I think some modern electric cars with motors attached to the individual wheels do away with axles as well). There are presumably good reasons for why vehicles tend to use axles, though.

reed501:
I think casters and tricycles are in the minority though and could use a separate system. Currently you install them all at once anyway which is fine.

In terms of axle abstraction I think that's a good play for simplicity without too much unnecessary burden. Maybe someone else can chime in but abstracting it away is probably the best course of action at least for now. However, it'd be nice to at least show that kind of abstraction somewhere. Maybe requiring a long piece of metal like a pipe in wheel mount recipes to show that it has to exist somewhere. Maybe a long stick for the wooden one.

PR #60676, opened 2022-09-02
colorschemes: add gruvbox light

sonphantrung:
#### Summary
Interface "colorschemes: add gruvbox light"

#### Purpose of change
We have gruvbox dark, but no gruvbox light (which is also quite a popular colorscheme)

#### Describe the solution
Add the colorscheme

None

#### Testing
See below

#### Additional context

image

image


Aerin-of-the-Toast:
Now who's that beautiful devil in the first pic?

Issue #60677, opened 2022-09-02
Copper does not spawn in gun related spawns


Sevaaas1:
### Describe the bug

Currently, copper does not spawn alongside guns, ammo or whatever is required to make a gun function, this is due to the item group that governs ammo parts not having the id, even though it spawns literally everything else needed to make bullets, aside from tools

### Steps to reproduce

1. Create new world
2. Go into any gunshops
3. No copper

OR

Check the relevant mapgen files related to ammo (gunsmith.json in "mapgen" folder, and gunstore.json in "mapgen_palettes" folder), copper wont be contained in any item group, nor will it be listed separately, thus no copper spawns

### Expected behavior

It's expected that copper spawns alongisde everything else needed to make bullets, as of right now the only way to get copper ("id": "copper") is processing other forms of copper like wires, currency and jewelry

### Screenshots

copper should be in this group for it to spawn alongside other bullet materials (itemgroups\Weapons_Mods_Ammo\ammo.json)

```
{
    "type": "item_group",
    "id": "ammo_parts","//": "Raw materials for assembling ammo excluding casings.",
    "subtype": "distribution",
    "entries": [
       {"item": "gunpowder_shotgun", "prob": 20 },
       { "item": "gunpowder_pistol", "prob": 105 },
       { "item": "gunpowder_magnum_pistol", "prob": 60 },
       { "item": "gunpowder_rifle", "prob": 85 },
       { "item": "gunpowder_large_rifle", "prob": 40 },
       { "item": "gunpowder_artillery", "prob": 5 },
       { "item": "incendiary", "prob": 20 },
       { "item": "lead", "prob": 160 },
       { "item": "shotgun_primer", "prob": 80 },
       { "item": "smpistol_primer", "prob": 80 },
       { "item": "smrifle_primer", "prob": 80 },
       { "item": "lgpistol_primer", "prob": 40 },
       { "item": "lgrifle_primer", "prob": 40 }
    ]
}
```

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.1889 (21H2)
- Game Version: 9bd27c3 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    <color_light_red>Secronom</color> [secronom],
[...]

PR #60679, opened 2022-09-02
Fixes copper not spawning for bullet crafting

Sevaaas1:
#### Summary
Bugfixes "Currently copper does not spawn alongside crafting materials for bullets, this fixes it"

#### Purpose of change

- Fixes: #60677 

#### Describe the solution

All it was lacking was an item id in the group that contains the raw materials, i added the item id with the same prob as lead, this lowers the spawn probability of the other items, however

None
#### Testing

Not much testing, i set the copper spawn probability as an extremely high value, then teleported to gunshops to check if it spawned, it spawned in all expected spawns
#### Additional context

None


RelMayers:
Should be half of lead according to crafting requirements

Sevaaas1:
> Should be half of lead according to crafting requirements

I actually considered half the spawnrate of lead, but some crafts use equal parts of lead and copper, maybe 3/4 of the probability of lead will be better?

Maleclypse (approved):

PR #60680, opened 2022-09-02
Scrap metal to scrap steel (2)

Karol1223:
#### Summary
Bugfixes "Changes scrap metal to scrap steel"

#### Purpose of change
We've got scrap copper and scrap aluminum now, scrap metal is a bit ambiguous nowadays.

#### Describe the solution
Simply changes one word in the string line of the scrap metal json. Also apparently multiplies the volume of scrap  steel 10x, as otherwise it actually ended up being just ungodly dense at 20g/cm3.

Leaving it as is but that's boring.

#### Testing
It's a _single word string change_, I refuse to accept this could cause crashes

#### Additional context
This is the second iteration of this PR as #60650 kept getting weird test failures I couldn't quite understand, this is an attempt at checking if the failures are somehow my fault or not. EDIT: The failures were actually because I couldn't grasp the concept of weight and volume for charge-based items. Now it's all clear.

### DIFF
--- a/tests/iteminfo_test.cpp
+++ b/tests/iteminfo_test.cpp
@@ -2494,7 +2494,7 @@ TEST_CASE( "disassembly time and yield", "[iteminfo][disassembly]" )
            "<color_c_white>Disassembly</color> takes about 20 minutes, requires 1 tool"
            " with <color_c_cyan>cutting of 2</color> or more and 1 tool with"
            " <color_c_cyan>screw driving of 1</color> or more and <color_c_white>might"
-           " yield</color>: 2 electronic scraps, 1 copper, 1 scrap metal, and 5 copper"
+           " yield</color>: 2 electronic scraps, 1 copper, 1 scrap steel, and 5 copper"
            " wire.\n" );
 
     CHECK( item_info_str( metal, disassemble ) ==


snipercup:
Thank you for contributing. The change seems simple on the surface but so many (deconstruction) recipes use scrap metal and they need to be adjusted when you change the volume of scrap metal. For example, chunk of steel has a volume of 63ml but disassembles into 5 scrap metal

image

To be fair, the disassembly recipe defies physics right now too, but after your change it will be even further from reality. If you intend on adjusting the disassembly recipe for chunk of steel, you'd have to go trough all of the recipes to adjust to the new volume of scrap of steel

Karol1223:
> Thank you for contributing. The change seems simple on the surface but so many (deconstruction) recipes use scrap metal and they need to be adjusted when you change the volume of scrap metal. For example, chunk of steel has a volume of 63ml but disassembles into 5 scrap metal
> 
> image
> 
> To be fair, the disassembly recipe defies physics right now too, but after your change it will be even further from reality. If you intend on adjusting the disassembly recipe for chunk of steel, you'd have to go trough all of the recipes to adjust to the new volume of scrap of steel
> 

To be fair, there's a deeper issue here since the mass, which arguably is the most important part, actually persists when the volume over doubles. For any and all disassembly recipes I added I followed solely the mass, as that has to stay the same, no new mass can EVER be generated whereas volume can go slightly higher. The issue here is the volume would over double. One fix would be to increase the volume of the chunk but then recipes involving that would flip upside down as well. I'll be honest, I don't know how to tackle that. I fixed the density for scrap metal but now that's an issue as well and, to be fair, if that's the case I'll just revert this change tomorrow morning and go back to it being a simple string change. This PR already caused too many issues with constant test failures I don't understand, I'll leave it for someone else to tackle.

PR #60681, opened 2022-09-02
Add ability to merge appliance into grid

Fris0uman:
#### Summary
Infrastructure "Add ability to merge appliance into grid"

#### Purpose of change

For performance purposes it would make sense to merge powergen and battery appliances into the grid to reduce the number of vehicles the code has to walk through

#### Describe the solution

- [x] Encapsulate code to merge appliances
- [x] Add menu to merge an appliance into a grid
- [x] Limit the option to powergen and batteries
- [x] Only merge adjacent appliances, no gap in the grid 
- [x] Prevent rotation of the grid upon merging
- [x] Make sure `Take down appliance` still works properly
- [x] Forbig taking down appliances that would split the grid
- [x] Reset `turn_dir` before merge to prevent rotation

- is_powergen() aplliance that make power, are reactores, batteries or wire
- if an appliance returns true on is_powergen() upon being installed it's made undraggable, that way it's rotation can't change and it won't mess things up when merging it
- only powergrid elements have the merging option

#### Testing

- spawn some wall wirings
- intall storage batteries, solar panels and generators around 
- can't drag them
- can merge them
- no rotation issues
- Place a standing lamp > can drag it around > can't merge it

To test rotation
- Allow dragging solar panel
- spawn a bunch 
- drag them around
- merge into wall grid > no issue
- merge wall grid into panel > no issue
- do it on a different wall > no issue
- make a different grid of dragged around panels and merge that into existing grid > no problem

#### Additional context

The merge option could probably be removed after next stable since the powergen appliances can't be dragged, can only be merged into adjacent appliences and will merge atuomatically upon being put down.

### DIFF
--- a/src/veh_appliance.cpp
+++ b/src/veh_appliance.cpp
@@ -515,27 +515,36 @@ void veh_app_interact::plug()
 
 void veh_app_interact::merge()
 {
-    const std::optional<tripoint> dir = choose_direction(
-                                            _( "Merge the appliance into which grid?" ) );
-    if( !dir ) {
+    map &here = get_map();
+
+    const int part = veh->part_at( a_point );
+    const tripoint app_pos = veh->global_part_pos3( part );
+
+    const std::function<bool( const tripoint & )> f = [&here, app_pos]( const tripoint & pnt ) {
+        if( pnt == app_pos ) {
+            return false;
+        }
+        const optional_vpart_position target_vp = here.veh_at( pnt );
+        if( !target_vp ) {
+            return false;
+        }
+        vehicle &target_veh = target_vp->vehicle();
+        if( !target_veh.has_tag( flag_APPLIANCE ) || !target_veh.is_powergrid() ) {
+            return false;
+        }
+        return true;
+    };
+
+    const std::optional<tripoint> target_pos = choose_adjacent_highlight( app_pos,
+            _( "Merge the appliance into which grid?" ), _( "Target must be adjacent." ), f, false, false );
+    if( !target_pos ) {
         return;
     }
-
-    const tripoint target_pos = get_player_character().pos() + *dir;
-    map &here = get_map();
-    const optional_vpart_position target_vp = here.veh_at( target_pos );
+    const optional_vpart_position target_vp = here.veh_at( *target_pos );
     if( !target_vp ) {
         return;
     }
     vehicle &target_veh = target_vp->vehicle();
-    if( !target_veh.has_tag( flag_APPLIANCE ) ) {
-        popup( _( "Target must be an appliance." ) );
-        return;
-    }
-    if( !target_veh.is_powergrid() ) {
-        popup( _( "A power grid must be wires, power generation or batteries." ) );
-        return;
-    }
     veh->merge_appliance_into_grid( target_veh );
 }
 
--- a/src/veh_appliance.cpp
+++ b/src/veh_appliance.cpp
@@ -116,6 +116,10 @@ void place_appliance( const tripoint &p, const vpart_id &vpart, const std::optio
 
     veh->last_update = calendar::turn;
 
+    if( veh->is_powergrid() ) {
+        veh->add_tag( flag_CANT_DRAG );
+    }
+
     // Update the vehicle cache immediately,
     // or the appliance will be invisible for the first couple of turns.
     here.add_vehicle_to_cache( veh );
@@ -353,6 +357,19 @@ bool veh_app_interact::can_siphon()
     return false;
 }
 
+bool veh_app_interact::can_unplug()
+{
+    vehicle_part_range vpr = veh->get_all_parts();
+    return std::any_of( vpr.begin(), vpr.end(), []( const vpart_reference & ref ) {
+        return ref.info().has_flag( "POWER_TRANSFER" );
+    } );
+}
+
+bool veh_app_interact::can_merge()
+{
+    return veh->is_powergrid();
+}
+
 // Helper function for selecting a part in the parts list.
 // If only one part is available, don't prompt the player.
 static vehicle_part *pick_part( const std::vector<vehicle_part *> &parts,
@@ -571,6 +588,20 @@ void veh_app_interact::populate_app_actions()
     imenu.addentry( -1, true, ctxt.keys_bound_to( "PLUG" ).front(),
                     ctxt.get_action_name( "PLUG" ) );
 
+    // Unplug
+    app_actions.emplace_back( [this]() {
+        unplug();
+    } );
+    imenu.addentry( -1, can_unplug(), ctxt.keys_bound_to( "UNPLUG" ).front(),
+                    ctxt.get_action_name( "UNPLUG" ) );
+
+    // Merge
+    app_actions.emplace_back( [this]() {
+        merge();
+    } );
+    imenu.addentry( -1, can_merge(), ctxt.keys_bound_to( "MERGE" ).front(),
+                    ctxt.get_action_name( "MERGE" ) );
+
     /*************** Get part-specific actions ***************/
     veh_menu menu( veh, "IF YOU SEE THIS IT IS A BUG" );
     veh->build_interact_menu( menu, veh->mount_to_tripoint( a_point ), false );
@@ -582,6 +613,34 @@ void veh_app_interact::populate_app_actions()
         app_actions.emplace_back( it._on_submit );
     }
     imenu.setup();
+}
+
+void veh_app_interact::merge()
+{
+    const cata::optional<tripoint> dir = choose_direction(
+            _( "Merge the appliance into which grid?" ) );
+    if( !dir ) {
+        return;
+    }
+
+    const tripoint target_pos = get_player_character().pos() + *dir;
+    map &here = get_map();
+    const optional_vpart_position target_vp = here.veh_at( target_pos );
+    if( !target_vp ) {
+        return;
+    }
+    vehicle &target_veh = target_vp->vehicle();
+    if( !target_veh.has_tag( flag_APPLIANCE ) ) {
+        popup( _( "Target must be an appliance." ) );
+        return;
+    }
+    if( !target_veh.is_powergrid() ) {
+        popup( _( "A power grid must be wires, power generation or batteries." ) );
+        return;
+    }
+    veh->merge_appliance_into_grid( target_veh );
+
+    return;
 }
 
 shared_ptr_fast<ui_adaptor> veh_app_interact::create_or_get_ui_adaptor()
--- a/src/veh_appliance.cpp
+++ b/src/veh_appliance.cpp
@@ -132,7 +132,7 @@ void place_appliance( const tripoint &p, const vpart_id &vpart, const std::optio
             continue;
         }
         vehicle &veh_target = vp->vehicle();
-        if( veh_target.has_tag( flag_APPLIANCE ) || veh_target.has_tag( flag_WIRING ) ) {
+        if( veh_target.has_tag( flag_APPLIANCE ) ) {
             if( veh->is_powergrid() && veh_target.is_powergrid() ) {
                 veh->merge_appliance_into_grid( veh_target );
                 continue;
--- a/src/veh_appliance.cpp
+++ b/src/veh_appliance.cpp
@@ -541,6 +541,10 @@ void veh_app_interact::merge()
 
 void veh_app_interact::populate_app_actions()
 {
+
+    int const part = veh->part_at( a_point );
+    const vehicle_part &vp = veh->part( part >= 0 ? part : 0 );
+
     const std::string ctxt_letters = ctxt.get_available_single_char_hotkeys();
     imenu.entries.clear();
     app_actions.clear();
@@ -568,7 +572,7 @@ void veh_app_interact::populate_app_actions()
     app_actions.emplace_back( [this]() {
         remove();
     } );
-    imenu.addentry( -1, true, ctxt.keys_bound_to( "REMOVE" ).front(),
+    imenu.addentry( -1, veh->can_unmount( vp ).success(), ctxt.keys_bound_to( "REMOVE" ).front(),
                     ctxt.get_action_name( "REMOVE" ) );
     // Plug
     app_actions.emplace_back( [this]() {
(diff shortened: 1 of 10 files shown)


robob27:
Hey @Fris0uman I was just wondering if this is considered working minus the outstanding checklist items? I was testing it with my save and had some odd results after merging an appliance. Obviously this is still a draft so good chance it's not done yet, but I am watching this with great anticipation to test it asap 😄.

If the merge part is supposed to be working now though I can share my save for debugging purposes. I don't really understand the vehicle mount stuff 😅

Fris0uman:
It works but I need to add restrictions to it, becasue for exemple if you put down a fridge aplliance, drag it a bit to make it turn and then merge it into the grid the grid will probably rotate and mess up everything

irwiss:
Is there a need to expose these choices to the player?

As in, what are the factors players need to consider whether they want to merge a particular appliance or not?

And what are the conditions where you wouldn't want to merge? If it's transparent to the player, may be appliances should automatically merge by default

robob27:
>the grid will probably rotate and mess up everything

Ahh okay, yeah that's what I was seeing. I saw your note about restricting it to powergen and batteries so I was trying with only solar panel and batteries, but I see there has been some more updates so I will continue testing.

I have a save I haven't really been able to play due to appliance slowdown so this is very exciting to me. Huge personal thanks from me for your efforts on this.

robob27:
Still having some rotation issues after the latest changes. After merging the battery directly north of my character into the wall wiring to the east of the battery, the wall wiring ends up south of my character.

image
image

I also noticed that solar panels don't count as `vehicle::is_powergrid()` at night so they can only be merged during the day.

Fris0uman:
So maybe we would need some migration code to reset all facing on previousy installed appliances?

robob27:
I noticed that when I check voltage on wall wiring/battery that has been merged with solar panels it says "The solar panel has voltage" instead of the name of the specific appliance I'm checking.

robob27:
My experiences with taking down appliances with these changes in case it's helpful:

- Taking down an appliance in a merged vehicle removes all power connections from the entire vehicle. So if you remove a solar panel off the end of your array, you'll have to plug your fridge back in.
- Using appliance removal, you can make large gaps in the vehicle, larger than what you'd be able to pull off with deploying/merging appliances only (without the use of extension cords). In the first screenshot below, all of the wires are a single merged vehicle and there is a kiln connected with "plug in vehicle" to the northeast wire. In the second screenshot, I removed all of those wires. The remaining wires are still the same vehicle - powering one of them powers both of them despite the lack of any real connection.

image

image

This actually seems kinda cool with wall wiring. It's like I'm installing the wires and then covering them up. I guess it's a bit weirder in the case of solar panels or batteries though.

dseguin:
Here's the fix for the "Take down appliance" bug:

```diff
diff --git a/src/veh_interact.cpp b/src/veh_interact.cpp
index 09a577a161..63c1487c46 100644
--- a/src/veh_interact.cpp
+++ b/src/veh_interact.cpp
@@ -3495,14 +3495,13 @@ void veh_interact::complete_vehicle( Character &you )
             }
 
             // Remove any leftover power cords from the appliance
-            if( appliance_removal && veh->part_count() >= 2 ) {
+            if( appliance_removal ) {
                 veh->shed_loose_parts();
-                veh->part_removal_cleanup();
                 //always stop after removing an appliance
                 you.activity.set_to_null();
             }
 
-            if( veh->part_count( true ) < 2 ) {
+            if( veh->part_count( true ) - veh->removed_part_count < 2 ) {
                 you.add_msg_if_player( _( "You completely dismantle the %s." ), veh->name );
                 you.activity.set_to_null();
                 // destroy vehicle clears the cache
```

The issue is was that `vehicle::part_removal_cleanup` can shuffle the part indexes so that the selected part to remove now points to a different part (or nothing!). So instead of calling `part_removal_cleanup` early, subtract the removed part count from the check, which achieves the same behaviour without touching the part indexes. Also, we don't have to worry about checking the part count when calling `vehicle::shed_loose_parts`, just need to check if it's an appliance.

Fris0uman:
It's almost working, there's still a bug where if you remove an appliance and then reinstall it it messes up the whole grid. For exemple place 3 wires, remove the middle one, reinstall it, examine the grid, now one of the tiles has two wire on the same spot. I don't know why

robob27:
Hey @Fris0uman, in case it helps at all here's a branch with these changes re-applied on the latest master without conflicts: https://github.com/CleverRaven/Cataclysm-DDA/compare/master...robob27:merge_app_rebase?expand=1

I've testing merging an appliance and it worked. I haven't tried to resolve the last bug you mentioned, or tested anything else yet. I set this branch up because for a while I had been building my game with this branch on top of the latest master. A couple weeks ago there were some pretty significant changes so I was no longer able to easily do this, so I did this.

The rebase was complex for me, so I didn't do a true rebase. I found it easier to start fresh from master and use this PR as a reference. Since I couldn't use your actual commits, I put you as a co-author in my commit just in case. Let me know if you'd like me to remove this. Didn't feel right copy/pasting your code without doing it, but also feels weird putting your name on it without consulting you 😅

dseguin:
It seems like there are two issues when placing wall wiring:
- Fake parts interfere with the bounding box and vehicle part merging
- All mounting points get shifted if the "origin" part is removed, but `vehicle::coord_translate` doesn't reflect the updated position

I tested briefly with this patch. This will only merge "real" parts, and the merge target is shifted back to zero if the origin part was previously removed (I'm really not sure why `vehicle::coord_translate` doesn't account for the pivot point). There might be other issues but this seems to solve those two above.
```diff
diff --git a/src/vehicle.cpp b/src/vehicle.cpp
index e296cde447..5a89dff4a7 100644
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -1827,6 +1827,9 @@ bool vehicle::merge_rackable_vehicle( vehicle *carry_veh, const std::vector<int>
 bool vehicle::merge_vehicle_parts( vehicle *veh )
 {
     for( const vehicle_part &part : veh->parts ) {
+        if( part.is_fake || part.removed ) {
+            continue;
+        }
         point part_loc = veh->mount_to_tripoint( part.mount ).xy();
 
         parts.push_back( part );
@@ -1847,13 +1850,14 @@ void vehicle::merge_appliance_into_grid( vehicle &veh_target )
     if( &veh_target == this ) {
         return;
     }
+    veh_target.shift_parts( get_map(), veh_target.pivot_displacement() );
 
-    bounding_box vehicle_box = get_bounding_box( false );
+    bounding_box vehicle_box = get_bounding_box( true, true );
     point size;
     size.x = std::abs( ( vehicle_box.p2 - vehicle_box.p1 ).x ) + 1;
     size.y = std::abs( ( vehicle_box.p2 - vehicle_box.p1 ).y ) + 1;
 
-    bounding_box target_vehicle_box = veh_target.get_bounding_box( false );
+    bounding_box target_vehicle_box = veh_target.get_bounding_box( true, true );
 
     point target_size;
     target_size.x = std::abs( ( target_vehicle_box.p2 - target_vehicle_box.p1 ).x ) + 1;
@@ -7310,7 +7314,7 @@ bool vehicle::restore_folded_parts( const item &it )
     return true;
 }
 
-const std::set<tripoint> &vehicle::get_points( const bool force_refresh ) const
+const std::set<tripoint> &vehicle::get_points( const bool force_refresh, const bool no_fake ) const
 {
     if( force_refresh || occupied_cache_pos != global_pos3() ||
         occupied_cache_direction != face.dir() ) {
@@ -7318,6 +7322,9 @@ const std::set<tripoint> &vehicle::get_points( const bool force_refresh ) const
         occupied_cache_direction = face.dir();
         occupied_points.clear();
         for( const std::pair<const point, std::vector<int>> &part_location : relative_parts ) {
+            if( no_fake && part( part_location.second.front() ).is_fake ) {
+                continue;
+            }
             occupied_points.insert( global_part_pos3( part_location.second.front() ) );
         }
     }
@@ -7656,7 +7663,7 @@ void vehicle::calc_mass_center( bool use_precalc ) const
     }
 }
 
-bounding_box vehicle::get_bounding_box( bool use_precalc )
+bounding_box vehicle::get_bounding_box( bool use_precalc, bool no_fake )
 {
     int min_x = INT_MAX;
     int max_x = INT_MIN;
@@ -7667,7 +7674,7 @@ bounding_box vehicle::get_bounding_box( bool use_precalc )
 
     precalc_mounts( 0, turn_dir, point() );
 
-    for( const tripoint &p : get_points( true ) ) {
+    for( const tripoint &p : get_points( true, no_fake ) ) {
         point pt;
         if( use_precalc ) {
             const int i_use = 0;
diff --git a/src/vehicle.h b/src/vehicle.h
index 5632670155..4d55e69dcd 100644
--- a/src/vehicle.h
+++ b/src/vehicle.h
@@ -1760,7 +1760,7 @@ class vehicle
         bool assign_seat( vehicle_part &pt, const npc &who );
 
         // Update the set of occupied points and return a reference to it
-        const std::set<tripoint> &get_points( bool force_refresh = false ) const;
+        const std::set<tripoint> &get_points( bool force_refresh = false, bool no_fake = false ) const;
 
         /**
         * Consumes specified charges (or fewer) from the vehicle part
@@ -1909,7 +1909,7 @@ class vehicle
         // Called by map.cpp to make sure the real position of each zone_data is accurate
         bool refresh_zones();
 
-        bounding_box get_bounding_box( bool use_precalc = true );
+        bounding_box get_bounding_box( bool use_precalc = true, bool no_fake = false );
         // Retroactively pass time spent outside bubble
         // Funnels, solar panels
         void update_time( const time_point &update_to );
```

robob27:
I tried testing with this patch and noticed a couple issues. If I reveal wall wiring on a tile with no other appliances adjacent, this works normally. If I then reveal another wall wiring in an adjacent tile, it causes the first one to look like the fake part sprite, Then if I remove the wall wiring that looks like a fake part, it seems to shift the second wall wiring up

After revealing first wire:
image

After revealing the second:
image

After removing the first:
image

This gets us closer though.. fake parts definitely explain the weirdness.

actually-a-cat:
Some observations from trying to figure out the "missing part after merge" issue:

* When you examine a vehicle tile, parts to list are looked up from `vehicle::relative_parts`. The broken "empty" tile actually contains the part it's supposed to, but `relative_parts` doesn't.
* `relative_parts` is populated on `vehicle::refresh` in this loop:
```cpp
    // Main loop over all vehicle parts.
    for( const vpart_reference &vp : get_all_parts() ) {
```
but this only runs once even after the merge, when the vehicle definitely contains two real parts.

and, possibly unrelated but:
* Freshly placed appliances seem to come with an inactive fake part. When merging two of those, the real part is copied over without its fake, but it still has `has_fake == true`.

dseguin:
Sorry, I forgot to include this line originally:

```diff
diff --git a/src/vehicle.cpp b/src/vehicle.cpp
index d6d85e6143..ea61ea7e7f 100644
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -1833,6 +1833,7 @@ bool vehicle::merge_vehicle_parts( vehicle *veh )
         }
         point part_loc = veh->mount_to_tripoint( part.mount ).xy();
 
+        remove_fake_parts();
         parts.push_back( part );
         vehicle_part &copied_part = parts.back();
         copied_part.mount = part_loc - global_pos3().xy();
```

Otherwise the relative part list doesn't match the part list order. Seems to fix the issue.

Fris0uman:
I think it works now

andrei8l:
You'll have to check for vehicle splitting whenever you take down an appliance:
<details>
<summary>I built this power grid</summary>

Screenshot from 2022-11-16 11-37-46
</details>

<details>
<summary>Removed one wiring</summary>

Screenshot from 2022-11-16 11-37-57
</details>

<details>
<summary>But the two halves are still connected</summary>

Screenshot from 2022-11-16 11-38-08
</details>


<details>
<summary>Physically too</summary>

Screenshot from 2022-11-16 11-38-24
</details>

but that doesn't have to block this PR.

Fris0uman:
so the last commit fixes the issue of grids not splitting but brings back the weird rotation and wrong index issues. Basically make a grid, remove a piece, install a new piece > the whole grid might rotate
OR remove a pievce, remove a second piece > it uninstall something different than what you selected or two parts at the same time. I don't know how to fix this

src/vehicle.h
```diff
@@ -1763,7 +1767,8 @@ class vehicle
         bool assign_seat( vehicle_part &pt, const npc &who );
 
         // Update the set of occupied points and return a reference to it
-        const std::set<tripoint> &get_points( bool force_refresh = false ) const;
+        const std::set<tripoint> &get_points( const bool force_refresh = false,
+                                              const bool no_fake = false ) const;
```
Brambor, line 1771:
```suggestion
        const std::set<tripoint> &get_points( bool force_refresh = false, bool no_fake = false ) const;
```

src/veh_appliance.cpp
```diff
@@ -563,6 +585,34 @@ void veh_app_interact::populate_app_actions()
...
+        popup( _( "Target must be an appliance." ) );
+        return;
+    }
+    if( !target_veh.is_powergrid() ) {
+        popup( _( "A power grid must be wires, power generation or batteries." ) );
+        return;
+    }
+    veh->merge_appliance_into_grid( target_veh );
+
+    return;
```
Brambor, line 613:
```suggestion
```

Brambor (changes requested):
Tidy code according to Clang-tidy

Fris0uman:
Can't reproduce the bug that make the grid rotate around, so as far as I can tell it works

Fris0uman:
image
I did get one situation where removing a solar panel and placing it back teleported some part of it away, not sure how to reproduce

Fris0uman:
Initial setup with everything connected:
image

Remove the panel and re place it:
image
 The wires are moved

Fris0uman:
I'm going to list the bugs I find here to keep track of them, there are at least two things going on:
- First is if you have a line of wires and you merge into it one other aplliance when you remove the extra appliance the wire grid loses it's appliance and wire flags
- Second is remove something and add something back and the grid shifts around

Issue #60588, opened 2022-08-30
Nested recipes make it hard to quickly scan your recipes


Fris0uman:
### Is your feature request related to a problem? Please describe.

- 1 Nested recipes appear in your list in a nice interactable blue that always make me think I can craft a thing when in fact I can't
- 2 There's no way to know what is in the nest whithout opening it
- 3 Once you get inside the nested recipe you can't go back to where you were without tabbing back to the category and scrolling down again

### Solution you would like.

1 and 2 could be solved by listing the recipes in the category in the right panel, right now that panel is pretty much empty it could instead list all the recipes listed in the category in white if craftable and grey if not

3 was partially adressed in #60378 but when you back out to NESTED you're still not where you were when you opened the list of nested recipes

### Additional context

Nested recipes are a great idea but personally right now I'm more annoyed at navigating through them than I was at scrolling through thousands of recipes.

bombasticSlacks:
These are all on my list fwiw I've just been on vacation. 

If someone else wanted to tackle 1/2 with your suggestion it is quite easy to do.

irwiss:
Perhaps consider tree-like nesting? it keeps visually consistent context without jumping around in the UI while providing the means to drill down nested items, for example:

```
flyswatter
maces
  great pipe mace
  pipe mace
  simple mace
nail bat
sling-ready explosives
  sling-ready grenade
  ...
```

Could also help mitigate the fact crafting menu is turning into 4 dimensional monstrosity:
* categories on top ( switch with tab / backtab )
* categories below them ( switch left/right arrow ), which default to "all" anyway, i doubt most players even notice them
* nested categories which are kinda-items ( switch with enter and pretty sure you can't go back )
* finally the actual items which you wanted to actually see

Tree view would make navigation super consistent; up/down to select, right arrow to expand and left to collapse, filters/search could also show you which categor you're in

bombasticSlacks:
If you wanna scrap what I've got and redo it as a tree while also doing the entire crafting menu rework required to make your control scheme work feel free 🤷

Let me know if you are and I won't bother with these last two bits of cleanup.

ANickelN:
To your point #2 @Fris0uman , which nests are unclear as to what is inside?  I've tried to label them all as obviously as possible, but if there is any that seem too ambiguous let me know and I can go back and change them.

Fris0uman:
No but the names are fine they still don't give me the list of all recipes inside

[...]

PR #60683, opened 2022-09-02
UI Upgrades for Nests

bombasticSlacks:
#### Summary
None

#### Purpose of change
parts 1 and 2 of #60588 

#### Describe the solution
Nested recipes now list all their internal crafts in the right panel.
They check if you can craft anything inside them and are colored based on that.
The recipes on the right side are also colored based on their craft-ability.

#### Additional context
image

image

image

### DIFF
--- a/src/crafting_gui.cpp
+++ b/src/crafting_gui.cpp
@@ -941,7 +941,7 @@ static void perform_nested( const recipe *rec, std::string filterstring, tab_lis
 }
 
 static std::string list_nested( const recipe *rec, const inventory &crafting_inv,
-                         const std::vector<npc *> &helpers, int indent = 0 )
+                                const std::vector<npc *> &helpers, int indent = 0 )
 {
     std::string description;
     availability avail( rec );


ANickelN:
I will still never be able to understand your color scheme, but this is an awesome change nonetheless!  😁

src/crafting_gui.cpp
```diff
@@ -915,6 +940,25 @@ static void perform_nested( const recipe *rec, std::string filterstring, tab_lis
     recalc = true;
 }
 
+std::string list_nested( const recipe *rec, const inventory &crafting_inv,
```
bombasticSlacks, line 943:
```suggestion
static std::string list_nested( const recipe *rec, const inventory &crafting_inv,
```

src/crafting_gui.cpp
```diff
@@ -915,6 +940,25 @@ static void perform_nested( const recipe *rec, std::string filterstring, tab_lis
...
+                         const std::vector<npc *> helpers, int indent = 0 )
```
bombasticSlacks, line 944:
```suggestion
                         const std::vector<npc *> &helpers, int indent = 0 )
```

NetSysFire:
While playtesting I noticed that the darker tone of blue is detrimental since it makes it a bit harder to intuitively see what recipes you can craft. It is almost as dark as the unavailable recipes. Unfortunately I can not see many benefits of coloring a nested group bright when you have the components for *all* recipes in that group since that is rarely relevant.

bombasticSlacks:
I use an alternate pallet so if it is hard to see in vanilla I think it would be reasonable to swap!

NetSysFire:
I'll make a separate issue for this with screenshots

PR #60686, opened 2022-09-02
Nerf combat knife damage

### LINKED ISSUE #60684: Nerf the Combat Knife

We all know the [combat knife](https://nornagon.github.io/cdda-guide/#/item/knife_combat) is really good, but there seems to be less consensus on whether it's too good. Here's why I think it is, in fact, too good:

To start with, the US army [doesn't actually seem to issue soldiers with combat knives](https://www.quora.com/What-is-the-standard-issue-US-Army-knife) these days, so they're probably too common. In as far as they do get knives, it's mostly the [M9 bayonet](https://en.wikipedia.org/wiki/M9_bayonet). The "combat knife" item seems to be loosely based on this knife, as it's the same length and can be attached as a bayonet.

The thing is that the M9 isn't a dedicated fighting knife. It's a multipurpose utility knife, complete with wirecutting functionality and a sawblade. It's a compromise design, which is why it seems weird to me that it deals noticeably more damage than dedicated fighting knives like the [kukri](https://nornagon.github.io/cdda-guide/#/item/kukri), [tanto](https://nornagon.github.io/cdda-guide/#/item/tanto) and [trench knife](https://nornagon.github.io/cdda-guide/#/item/knife_trench). In fact, its damage is closer to certain swords, like the [rapier](https://nornagon.github.io/cdda-guide/#/item/rapier), than it is to most knives.

Those weapons have their own pros and cons, of course, but I just cannot get over the fact that this tertiary utility sidearm is up there competing with actual weapons.

### Solution you would like.

Just drop its pierce damage a few points. I think you could go down all the way to 20, tbh. Maybe do like 22 and make it a bit slower to represent that it's not designed for maximum agility.

_No response_

### Additional context

Part of the problem here is that the damage range for melee weapons is heavily compressed into the 20-40 range. There's just not a lot of room to make a particularly good knife, because just one or two points of damage too many and it's competing with swords. But that's a problem that would require a much more radical intervention to solve.

Terrorforge:
#### Summary
Balance "Nerf combat knife damage"

#### Purpose of change

Fixes #60684

The combat knife does an unreasonable amount of damage. I've laid out detailed reasoning in the aforementioned issue, but the TL;DR is that the weapon it represents is more of a utility tool, so it doesn't make sense for it to be better than the dedicated fighting knives.

#### Describe the solution

Decreased damage to 20.

### DIFF
--- a/data/json/items/melee/swords_and_blades.json
+++ b/data/json/items/melee/swords_and_blades.json
@@ -147,7 +147,7 @@
     "price_postapoc": 1250,
     "to_hit": { "grip": "weapon", "length": "hand", "surface": "point", "balance": "good" },
     "bashing": 4,
-    "cutting": 26,
+    "cutting": 20,
     "material": [ "steel", "plastic" ],
     "symbol": ";",
     "color": "dark_gray",


GuardianDll:
you need to edit a `CHECK( calc_expected_dps( "knife_combat" ) == 19.0 );` value in tests/effective_dps_test.cpp
i have no clue how it calculated, but i expect percent decrease would work (in my case the final value is 14.6)

anoobindisguise:
Combat knife is not that OP. It cannot block. That is a major limitation on its abilities, blocking is really important. It should instead be based on a Ka-Bar military knife and have its material changed to High Steel and its damage under that condition becomes about expected for a high steel weapon. @Drew4484 currently plans on setting the cut/stab damage of Tempered Steel weapons to about 20% higher than that of current "steel" weapons which are in-between low and medium steel in terms of quality - eventually "steel" will be replaced for a carbon steel type, but this is the current balance basis. In addition it can be fixed to not be repairable with Plastic since that's just its grip. Military knives aren't necessarily meant as a dedicated stabbing tool, but lowering to 20 is probably too low especially if the quality of its steel increases.

Terrorforge:
> 20 is barely better than a kitchen carving knife (18)

Yes, but that's where every other knife sits. Trench knife at 19, tanto at 21, kris at 20. The kukri is a bit of an outlier at 23, and that's a honking great cleaver bordering on a shortsword. Compared to that, 20 is if anything too generous. That's about where the ka-bar would go, the M9 could easily be bumped down to 18.

And yes, I understand that all those have other properties that make up for slightly lower damage in terms of gameplay balance, but I'm more concerned with the realism. I don't see any compelling real-world reason the combat knife should do 30%, 20%, or even 10% more damage then every other fighting knife.

That said, part of the problem is that weapon damage is a little wonky in general. Many of the weapons are still at least partly balanced for gameplay rather than realism, and the margins are razor thin.

anoobindisguise:
I do agree that it would be too good if it was made of the same material as other knives, which IS currently the case, although the baseline steel quality is assumed to be very poor, and 22 is much more in line with other knives if we assume that's the damage it would have as a standard steel weapon. Other steel weapons might become medium steel and see a 1 ish damage increase from that according to the current plan.

Terrorforge:
That's where the narrow margins come in. 22 is not in fact "in line" with other knives, it's really high. Higher than any knife of the same size. Base damage assuming standard steel should be 20, and if we make it high carbon steel that damage bonus gets you to like 22-23.

By the way, why did you want to base it on the ka-bar? My (admittedly limited) research suggests the M9 is more common in US army service, and the ka-bar doesn't have the bayonet functionality we see in the game.

Drew4484:
> That's where the narrow margins come in. 22 is not in fact "in line" with other knives, it's really high. Higher than any knife of the same size. Base damage assuming standard steel should be 20, and if we make it high carbon steel that damage bonus gets you to like 22-23.
> 
> By the way, why did you want to base it on the ka-bar? My (admittedly limited) research suggests the M9 is more common in US army service, and the ka-bar doesn't have the bayonet functionality we see in the game.

The M9/Ka-bar discussion happens often enough that it might be worth making two distinct items. JM2C

catdach:
> > That's where the narrow margins come in. 22 is not in fact "in line" with other knives, it's really high. Higher than any knife of the same size. Base damage assuming standard steel should be 20, and if we make it high carbon steel that damage bonus gets you to like 22-23.
> > 
> > By the way, why did you want to base it on the ka-bar? My (admittedly limited) research suggests the M9 is more common in US army service, and the ka-bar doesn't have the bayonet functionality we see in the game.
> 
> The M9/Ka-bar discussion happens often enough that it might be worth making two distinct items. JM2C

Could probably throw in the OKC-3S too if people really want a Ka-bar that's also a bayonet.

anothersimulacrum (changes requested):
I think the issue lays out a decent rationale for change, but I think the effective change is too much, I would expect DPS (as per the test) to be around 17 or so. It's not _the best_ knife (which it currently is, according to the test, by a decent margin of 1 point), but it should still be a good knife.
I think name and description changes should accompany this to make it clear that this is a bayonet/multitool. I would advise renaming it the the m9 combat knife and updating it's description to match.

Terrorforge:
Looking at the test values, I think the problem there is that the kirpan specifically also has an absurd damage value. I'm guessing it was keyed off the combat knife, but that just means that it too deals significantly more damage than any other knife for even less real reason. I contend that both it and the combat knife should be down at ~15 dps with the trench knife.

Sevaaas1:
If the combat knife is recieving a damage buff because its a utility combat knife, which it is due to being serrated (if we assume its based on the M9 Bayonet), it should recieve ``` wood sawing 1 ```, and maybe ``` tree cutting 1 ``` since it is a sturdy knife with a lot of mass for a knife (some sellers claim between 500 grams to a kilo of weight. One of the current manufacturers claims 600 grams) that, plus with the saw part you should be able to take care of small trees, i think it should also have ```dig 1 ``` with a severe speed penalty, since it is a sturdy knife, you can use it to loosen the soil and move it with your hands

anothersimulacrum:
I'd read that more as the trench knife could go up, the status quo is dedicated combat knives are in this range, I think it's a much bigger change to get into that.

Terrorforge:
> the status quo is dedicated combat knives are in this range

No, it's not. It's literally just the combat knife. The kirpan is not a dedicated fighting knife. Sharp or not, it is a ceremonial weapon, and its current state is obviously a mistake. Every other dedicated fighting knife is much closer to the trench knife, and many are even worse. The kukri is an exception, but that's a different class of weapon altogether. Doesn't fit in an ankle sheath, for starters.

I'm not necessarily opposed to the idea that all of those could go up a bit instead of dumpstering the combat knife, but then we're looking at a much bigger audit.

Drew4484:
> > the status quo is dedicated combat knives are in this range
> 
> No, it's not. It's literally just the combat knife. The kirpan is not a dedicated fighting knife. Sharp or not, it is a ceremonial weapon, and its current state is obviously a mistake. Every other dedicated fighting knife is much closer to the trench knife, and many are even worse. The kukri is an exception, but that's a different class of weapon altogether. Doesn't fit in an ankle sheath, for starters.
> 
> I'm not necessarily opposed to the idea that all of those could go up a bit instead of dumpstering the combat knife, but then we're looking at a much bigger audit.

Seeing as that's the appropriate solution, I would either redirect this PR towards updating other blades damages or just closing this.

Terrorforge:
Could you elaborate on why you think that's the appropriate solution? Like I said, I'm not necessarily opposed, buy it's also not obvious to me that it would be the better solution.

In fact I just had a quick look, and the DPS of knives is already uncomfortably close to swords, so I'm feeling a bit skeptical

anothersimulacrum:
You're right, my bad. This was the behaviour prior to e0b9547961a. I don't have too much of a game in the weapons race, but these tests date to a pretty extensive balance pass of all weapons (https://github.com/CleverRaven/Cataclysm-DDA/pull/41079), ensuring certain classes to be around the same DPS, and I'd like to preserve in broad strokes that work. If it be decided that 15-16 is the correct range, make it so (but also nerf monomolecular blade), and update the comment.

Drew4484:
> Could you elaborate on why you think that's the appropriate solution? Like I said, I'm not necessarily opposed, buy it's also not obvious to me that it would be the better solution.
> 
> In fact I just had a quick look, and the DPS of knives is already uncomfortably close to swords, so I'm feeling a bit skeptical

I surrender my argument to anothersimulacrum, basically I just want the majority of knives to deal similar damage with few outliers.

Amoebka:
The combat knife is literally used as a balance benchmark for all knives in tests. So yes, it's everything else that might need to change.

In general, if your test requires tests to be adjusted, you should stop and ask yourself if you are doing the right thing. That's kinda the reason tests exist.

x-qq:
Combat knife already has -1 to hit and no blocking ability, this is enough to represent that it is not a dedicated combat weapon. 26 piercing damage makes it one of the few early game options (and in some starts, the only option) that can make a dent in zombie soldiers in a reasonable amount of time. As such I do not think any reduction is necessary.

However it would be nice to have its utility abilities (wood cutting and wire cutting) added.

Altogolik:
I use this knife a lot, if you want to weaken it, add military variations of cooler knives. I think what's missing is a variety of rarer and more dangerous knives.

need cool knives at the end of the game (A few examples)

Ontario MK 3 Navy Knife (not a big but very dangerous knife is great for killing)
https://i.gyazo.com/a0b10135e307da1fcfb1812410a49011.jpg

Fairbairn-Sykes (Legendary British Knife)
https://i.gyazo.com/0785e0042118f49fc7c8682fd7b4e9af.jpg

Gerber Mark II (Commando’s Knife)
https://i.gyazo.com/799b9efd923da2189ccd6a3383eeb495.jpg

Cold Steel Recon Tanto (variation of Japanese knife)
https://i.gyazo.com/4bd7abb0cb147cdf756539af5a46b4b6.jpg

Boker Applegate-Fairbairn (Excellent stealth knife)
https://i.gyazo.com/b6053f381d19007ae1f6722d7eef2873.jpg

Glock 78 \ 81
https://i.gyazo.com/63ef0a8f9f79da5bf486672b4af42d58.jpg

Translated with www.DeepL.com/Translator (free version)

Terrorforge:
> You're right, my bad. This was the behaviour prior to [e0b9547](https://github.com/CleverRaven/Cataclysm-DDA/commit/e0b9547961a3f0e849e06d9c1ca1677cad9abd21). I don't have too much of a game in the weapons race, but these tests date to a pretty extensive balance pass of all weapons (#41079), ensuring certain classes to be around the same DPS, and I'd like to preserve in broad strokes that work. If it be decided that 15-16 is the correct range, make it so (but also nerf monomolecular blade), and update the comment.

Oooh, I see what happened here. When they did the audit, many knives lost their +hit bonuses, which made the tests upset. So they just... changed the tests to match the new damage. What they should have done in order to comply with game balance as laid out by GAME_BALANCE.md (thank you @Amoebka) was increase the damage to match the old test values at the new to-hit.

I maintain that the combat knife still has undeservedly high damage, but this is the real reason it's _so_ out of line. I'll write up an issue detailing the problem properly.

Terrorforge:
Wrote it up as #60766.

I'm closing this PR because it turned out the main problem was something else. I still don't particularly see why the combat knife deserves to be the very bestest knife, but the difference being as big as it is is the result of an error, and I'm not sufficiently annoyed to quibble over this much smaller difference.

If anyone wants to implement the other changes discussed, such as adding utility functions or differentiating between the M9 and ka-bar, be my guest. I'm dropping this for now.

PR #60699, opened 2022-09-02
Luty 9mm SMG fixes

DPavonis:
#### Summary
Bugfixes "Brings the 9mm Luty SMG in line with its real-world counterpart."

#### Purpose of change

To bring the ingame Luty 9mm SMG in line with its real-world counterpart. The real-world Luty and the book in which its construction methods were detailed predates the widespread adoption of rail interface systems, in addition to having no provisions for a stock, though with its many, many flat surfaces, a survivor subsequently refitting the weapon with rails and a stock is well within the realm of possibility.

#### Describe the solution

Changed rail, sights, stock and underbarrel to their respective mounts.

#### Testing

Fired up the game, made sure the Luty 9mm had the appropriate mod locations.

#### Additional context

Image of a Luty SMG taken from: https://youtu.be/sIhGCRIQnCA?t=477

Weapons as Political Protest_ P A  Luty's Submachine Gun 7-53 screenshot

### DIFF
--- a/data/json/items/gun/9mm.json
+++ b/data/json/items/gun/9mm.json
@@ -713,7 +713,7 @@
     "looks_like": "hk_mp5",
     "type": "GUN",
     "name": { "str": "Luty SMG: 9x19mm", "str_pl": "Luty SMGs: 9x19mm" },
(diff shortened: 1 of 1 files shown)


data/json/items/gun/9mm.json
```diff
@@ -740,11 +740,11 @@
       [ "grip", 1 ],
```
TheShadowFerret, line 740: Shouldn't this also be turned into a mount 🤔

DPavonis: Well the Luty does already have a grip, though perhaps not provisions to make it easily replaceable I suppose, so eh, sure, why not?

PR #60707, opened 2022-09-03
Leaping legs mutation will now delete the spell when lost

### LINKED ISSUE #53860: Losing "leaping legs" mutation does not also lose leap spell
### Describe the bug

Losing the "leaping legs" mutation, either through purification/mutagen or through the debug menu, does not prevent the player from knowing and using the short leap spell, even if that spell was only added by the mutation.

### Steps to reproduce

1. Obtain the leaping legs mutation
2. Lose the leaping legs mutation
3. Go into the spell menu and cast leap

### Expected behavior

I expected to forget the short leap spell when I lost the legs, especially since I am not playing with the Magicalysm mod.

### Screenshots

_No response_

OS: Windows 10
Game Version: d0422db
Graphics: Tiles
Language: English
Mods: Default (no magicalysm)

### Additional context

_No response_

MNG-cataclysm:
#### Summary
Bugfixes "The leaping legs mutation will now delete the spell when lost"

#### Purpose of change
Make the leaping legs mutation delete the spell when lost

#### Describe the solution
Adds a "REQUIRED_MUTATIONS" field to the leaping legs and crushing leap spells. 

None

#### Testing
None. The solution was described in issue #53860

#### Additional context
Closes #53860


Maleclypse:
I can't find any evidence that `"required_mutations"` will work with the spell code but it probably could be added to it pretty easily assuming my search is correct.

Maleclypse:
Reset the test because it coredumped rather than printing a real error message.

Stadler76:
> I can't find any evidence that `"required_mutations"` will work with the spell code but it probably could be added to it pretty easily assuming my search is correct.

I've suggested to add that field to the codebase as it doesn't exist yet. Meaning: It's not a simply json-only fix. Or else I would have filed a PR that fixes it myself.

GuardianDll:
Making PR without checking is it works is bad

PR #60710, opened 2022-09-03
Allow roundtripping damage in folded vehicle items

### LINKED ISSUE #60700: Damaging folded vehicles (the item) doesn't damage the vehicle
### Describe the bug

Feral grannies drop damaged folded wheelchairs, but when you unfold them, none of the parts are damaged, and when you fold it back up, it becomes a `folded wheelchair` instead of, say, a `|. folded wheelchair`.

Similarly, if you run over a folded bicycle (the item, not the vehicle) so that it becomes a `|. folded bicycle`, you can unfold it and none of the parts will be damaged, and folding it back up will change back to an undamaged `folded bicycle`.

On the plus side, if you damage an unfolded vehicle, then fold it and unfold it, the game will remember the vehicle's damage, so this isn't an exploit where you can repair vehicles without tools and skill just by folding and unfolding them.

### Steps to reproduce

1. Spawn a feral granny.
2. Kill it and pick up the `|. folded wheelchair`.
3. Unfold the wheelchair, examine the resulting vehicle, and see that all of the parts are `||`.
4. Fold it back up and now you have a `folded wheelchair` (undamaged).

### Expected behavior

First of all, it would be cool if feral grannies drove wheelchairs instead of carrying them.

I expect that damaging the item should also damage the vehicle parts underneath and damaging the vehicle should damage the item.

### Screenshots

_No response_

Linux, last commit is cb944bfe7b5bfbf08296133b7014f412d43d3bbf.

### Additional context

_No response_

irwiss:
#### Summary
None

#### Purpose of change

Fixes #60700

#### Describe the solution

On folding calculate average part damage and serialize into item_var and item itself.

On unfolding get the difference between current item and old item_var damage;
Start applying it to random parts in chunks

be40dc0de62ec757d8066d824066a756485196bc also fixes a couple more MSVC warnings I missed in #60568

Note that I had need for a random-looking sequence with fixed seed - I don't need it to actually be random, just so parts "look" randomly damaged in a way that tests can snapshot (and so don't rely on global rng state) and there seems to be none existing - so I [added one into rng.h/cpp](https://github.com/CleverRaven/Cataclysm-DDA/blob/be40dc0de62ec757d8066d824066a756485196bc/src/rng.h#L58-L65) 

Using one of the simple hashes available for the rng sequence, but it'd do the same job but with extra steps and possible bugs.

#### Testing

Spawn and kill zombie granny, unfold wheelchair, parts should be damaged.

Unit test is added, could also run the unit test scenario manually

### DIFF
--- a/src/rng.cpp
+++ b/src/rng.cpp
@@ -115,7 +115,6 @@ int djb2_hash( const unsigned char *input )
 
 std::vector<int> rng_sequence( size_t count, int lo, int hi, int seed )
 {
-    static std::uniform_int_distribution<int> rng_int_dist;
     if( lo > hi ) {
         std::swap( lo, hi );
     }
@@ -124,8 +123,10 @@ std::vector<int> rng_sequence( size_t count, int lo, int hi, int seed )
 
     // NOLINTNEXTLINE(cata-determinism)
     cata_default_random_engine eng( seed );
+    std::uniform_int_distribution<int> rng_int_dist;
+    const std::uniform_int_distribution<int>::param_type param( lo, hi );
     for( size_t i = 0; i < count; i++ ) {
-        result.push_back( rng_int_dist( eng, std::uniform_int_distribution<>::param_type( lo, hi ) ) );
+        result.push_back( rng_int_dist( eng, param ) );
     }
     return result;
 }


dseguin (approved):
LGTM

src/rng.cpp
```diff
@@ -113,6 +113,23 @@ int djb2_hash( const unsigned char *input )
     return hash;
 }
 
+std::vector<int> rng_sequence( size_t count, int lo, int hi, int seed )
+{
+    static std::uniform_int_distribution<int> rng_int_dist;
```
BevapDin, line 118: This should probably not be static as that could leak data from one invocation to the next one. (`uniform_int_distribution::operator()` is not `const`, it may change the state of the distribution.)

Or call `reset` before using the distribution again.

irwiss: Was like 90% sure it's const, but turns out it was [MSVC quirk](https://developercommunity.visualstudio.com/t/operator-of-uniform-int-distribution-is-a-const-me/740558), thanks

PR #35734, opened 2019-11-26
Bring EMP grenades in line with reality

anothersimulacrum:
#### Summary
SUMMARY: Balance "Revamp EMP grenade recipe"

#### Purpose of change
EMP grenades as they currently exist do not exist in real life.
This PR aims to make work as a plausible EMP device would, based off of [this paper](https://apps.dtic.mil/dtic/tr/fulltext/u2/a332511.pdf), but slightly miniaturized from the size of a full bomb.

#### Describe the solution
Add 4 new items:
The magnetron/microwave generator, a component of EMP grenades, that can be found in microwaves.
The flux compression generator, a component of EMP grenades, that can be rarely found in labs, or crafted.
The large copper tube, a component of the flux compression generator, that can be forged.
The EMP bomb, an explosive device that puts out an EMP pulse. This replaces the emp grenade.

I also moved an equivalent of the EMP grenade to be find/buy from Hub 01 only.
I also obsoleted the recipe for the scrambler grenade, and it can only be found or bought from Hub 01.

See 2fb335d7f202e57b9d364916a3f092f88abccefb for more detailed commentary.

#### Describe alternatives you've considered
There may be some other equivalents for the capacitor bank, but they need some special handling.

#### Testing
Tried crafting everything.
empblast
empuse

#### Additional context
This EMP device works by draining a high-current capacitor (represented
by a plutonium fuel cell) into a 'explosively propelled flux compression
generator'. When this generator is activated, with the aid of an
explosion, it creates a strong magnetic field. This magnetic field is
then passed through a microwave generator (magnetron), into a single
loop antenna, which creates the EMP.

The items added could spawn in more places.

TODO:
- [x] insulators
- [x] Fudge explosion less

PR #60723, opened 2022-09-03
EMP blast renders all electronic items in the radius of the blast non-functional

### LINKED ISSUE #46697: RC Car should probably not survive point blank EMP grenade
# Describe the bug

When you use an EMP grenade attached to an RC Car it is highly implausible that the RC Car survives intact.

# Steps To Reproduce

1. Make radio activation mod
2. Attach to EMP grenade
3. Put EMP grenade on RC Car
4. Use RC Controller to blow up EMP Grenade
5. Observer that the RC Car is unharmed and still can be controlled.

# Expected behavior

I would expect the radio control part of the RC car (at least) to be wiped out.

# Versions and configuration

- OS: Window x64 8.1
- Game Version: 0.E-8914-g8c0034b-dirty [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
]

Night-Pryanik:
#### Summary
Features "EMP blast renders all electronic items in the radius of the blast non-functional"

#### Purpose of change
* Closes #46697.

#### Describe the solution
~Assumed that any item that breaks in water (i.e. has a `WATER_BREAK` flag) should be destroyed by EMP blast as well.~
- Added newly `ELECTRONIC` flag to most electronic items.
- Added check for this flag for any item in EMP blast radius. If true, fry this item for it to become a paperweight.
- Also made it so EMP blast drain batteries and battery-operated items from their charge not only on the ground, but also in player inventory.
- Also fixed error which prevented draining batteries at all.

- [x] Add separate flag for electronic items instead of relying on `WATER_BREAK` flag.

#### Testing
Got two fully charged electric lanterns. Dropped one to the ground, the other one was left in my inventory. Activated EMP bomb and dropped it to the ground next to the lantern. Waited for bomb to explode. Checked battery charge in both lanterns, checked `broken` state of both lanterns.

#### Additional context
- Should immunity to electricity granted by worn items, such as faraday suit, also stop the effect of EMP blast?
- The WHOLE new code in this PR was written by Github Copilot, after I write the comments. All hail robots!


Karol1223:
While I don't mean to discard your work, I feel like a specific flag for this is sooner or later going to be useful (or necessary). Because as far as I can tell with this now EMPs also destroy matches, and to be fair WATER_BREAK feels like it should be given to a bunch of existing paper items as well. Though it's just a suggestion, really, and personally I can live with matches being destroyed in that one-in-a-hundred chance that they somehow get caught in an EMP blast.

PatrikLundell:
Unfortunately, it's not a 1:100 chance to get blasted by EMP, unless you count per monster (in which case it's too high). At least one of the electric monsters (Incandescent Husk?) uses EMP in its attacks, as do the alien electric "flora" (not quite sure which one of them, but given their spawning and clustering it probably doesn't matter).
I've also gotten blasted by something kept in lab cells by Autodocs (may or may not be the same electric monster: don't remember). I think the electric "traps" in labs (and the irradiation facility) also zaps you with EMP, but I'm not sure.

In addition to matches, I'd expect stuff such as aspirin and sugar to be dissolved in water, but those should be totally unaffected by EMP blasts.

In principle a Faraday cage should keep EMP blasts on the outside, but that's only if it's completely sealed. Otherwise the cage would let in (and out) emissions of wavelengths shorter than the holes/mesh. This means that anything worn would have to be equipped with its own oxygen supply and update of the air exhaled if it's to be completely immune (and you'd have to use some kind of transparent but still conductive material to be able to see anything (and seeing means letting in wavelengths in the visible part of the spectrum)).
I have no idea which wavelengths EMP blasts are supposed to emit, though.

Karol1223:
> Unfortunately, it's not a 1:100 chance to get blasted by EMP, unless you count per monster (in which case it's too high). At least one of the electric monsters (Incandescent Husk?) uses EMP in its attacks, as do the alien electric "flora" (not quite sure which one of them, but given their spawning and clustering it probably doesn't matter). I've also gotten blasted by something kept in lab cells by Autodocs (may or may not be the same electric monster: don't remember). I think the electric "traps" in labs (and the irradiation facility) also zaps you with EMP, but I'm not sure.
> 
> In addition to matches, I'd expect stuff such as aspirin and sugar to be dissolved in water, but those should be totally unaffected by EMP blasts.
> 
> In principle a Faraday cage should keep EMP blasts on the outside, but that's only if it's completely sealed. Otherwise the cage would let in (and out) emissions of wavelengths shorter than the holes/mesh. This means that anything worn would have to be equipped with its own oxygen supply and update of the air exhaled if it's to be completely immune (and you'd have to use some kind of transparent but still conductive material to be able to see anything (and seeing means letting in wavelengths in the visible part of the spectrum)). I have no idea which wavelengths EMP blasts are supposed to emit, though.

Honestly, that is very fair. I was under the assumption that EMP blasts came solely from EMP grenades and that shockers and other electrical friends used something different but in hindsight, it should have been a given they were one and the same. In this case, things get a bit more complicated, because every time you fight a shocker your matches will spontaneously disappear, it seems.  This certainly calls for a custom flag in this case.

PatrikLundell:
Well, it's not ALL shockers, only the higher (or highest) tier ones, but I still think it calls for a separate flag, unfortunately.

The lower tier shockers only provide electric damage. I guess you don't really care unless you go the CBM enhancement route, though, so many players aren't really affected to any significant extent currently.

chrispikula:
Does anyone have any sources on what level of emp actually destroys vs disrupts electronics?

I'm not asking about blob-magic zombies, I'm asking more about emp grenades.  'cause from what I understand via my electronics education, is that emp resistance is easy.  Heat sinks, EMF shielding, etc, can protect quite a number of electronics quite well.  

The classic example is wireless 2.4 GHz : Old microwaves that were poorly shielded would often emit high amounts of extra EM power that routers and other things that the 2.4 Amateur band were specifically designed to pick up.  And the equipment didn't often die from it, or if they did, were generationally upgraded to survive it.  Sure, your wifi would get knocked out, but it wouldn't kill the wifi card on your laptop.

I'd like to propose that the items get damaged, instead, with amount based also on the damage from the EMP.  How much, I dunno.  This correlates a bit to the damage electronics take from high frequency power line-riding EMF, and the rogue-wave effect.  Transistor leakage increases based on a rogue-wave collision of multiple high frequencies, if you don't use enough ferrite beads on your power supplies.

Night-Pryanik:
>in EMP grenade form factor

We don't have EMP _grenades_ [any more](#35734), instead we have EMP _bombs_.

anothersimulacrum:
We still hawe grenades, you can buy them from the hub or find them, and they operate on ???. I have little reason to doubt EMPs strong enough to destroy robotic enemies would damage electronics.

chrispikula:
> We still hawe grenades, you can buy them from the hub or find them, and they operate on ???. I have little reason to doubt EMPs strong enough to destroy robotic enemies would damage electronics.

Why should the grenades be able to destroy robotic enemies then, as well?  There's no reason for it, given how little effort it takes to block EMF.  Using your logic, I have little reason to think they'd be vulnerable in the first place, other than gamification mechanics.

src/explosion.cpp
```diff
@@ -754,11 +755,37 @@ void emp_blast( const tripoint &p )
...
+        for( item_location &it : electronic_stuff ) {
+            // Drain any items in player's possession of their battery charge
+            if( it->is_tool() && it->ammo_current() == itype_battery ) {
+                it->ammo_unset();
+            }
+            // Render any electronic stuff in player's possession non-functional
+            add_msg( m_bad, _( "The EMP blast fries your %s!" ), it->tname() );
+            it->deactivate();
+            it->set_flag( flag_ITEM_BROKEN );
+        }
```
dseguin, line 774:
```suggestion
        for( const item_location &it : player_character.all_items_loc() ) {
            if( it->has_flag( flag_ELECTRONIC ) && !it->is_broken() ) {
                // Drain any items in player's possession of their battery charge
                if( it->is_tool() && it->ammo_current() == itype_battery ) {
                    it->ammo_unset();
                }
                // Render any electronic stuff in player's possession non-functional
                add_msg( m_bad, _( "The EMP blast fries your %s!" ), it->tname() );
                it->deactivate();
                it->set_flag( flag_ITEM_BROKEN );
            }
        }
```

Night-Pryanik: Applied your suggestion (with little change), thanks.

Night-Pryanik:
Now that I think about it, do we really need separate flag for electronic items? We can consider any tool that uses batteries as electronic.

anoobindisguise:
There needs to be a way to protect against this. Ideally if you're storing your electronic goodies inside an electric proof armor, they are immune, but as that's probably super difficult, it may be better to just check for if you have that flag.
Also, it should probably be decided whether leechplants/incandescent husk deserve to have a 100% always on EMP. There's actually a bug where they'll drain your bionic power even with full electric resist which is silly.

PatrikLundell:
In the real world there is a difference in strength between different kinds of EMP sources. Something that's shielded to resist industrial environments is probably not shielded against military EMP sources, which I would assume EMP bombs and grenades to be, and those are probably weaker than those from an atomic blast (at some distance).
We could use the game assumption that all sources that are represented in the game emit military level EMP, of course, but then some sources should probably be removed (like arcing electrics).

Also note that protection against electricity isn't the same as protection against EMP. The metal shell of a car (one that isn't made out of composites, that is) protects against lightning strikes, but that would do little to protect against EMP, as the windows would provide access to the pulse.

anothersimulacrum:
> > We still hawe grenades, you can buy them from the hub or find them, and they operate on ???. I have little reason to doubt EMPs strong enough to destroy robotic enemies would damage electronics.
> 
> Why should the grenades be able to destroy robotic enemies then, as well? There's no reason for it, given how little effort it takes to block EMF. Using your argument, I have little reason to think they'd be vulnerable in the first place, other than gamification mechanics.

I think it's very reasonable to remove that, I'm not attempting to argue for it. I'm just saying that anything that generates an EMP which would have this effect on monsters, would probably have the same effect on other electronics. If the first one doesn't make sense in any form with what we have, we should fix that.

PatrikLundell:
The amount of effort required to block the effects of an EMP depends heavily on how strong it is. Military won't make EMP weapons that are useless: they'd have to be strong enough to at least knock out civilian equipment, but preferably also military equipment that hasn't got exceptional shielding.
However, a number of comparatively recent wars have shown a much larger dependence on civilian equipment than was probably anticipated (usage of civilian GPS equipment during the Iraq war, for instance, causing the US not to disable the civilian frequencies of the service over the area, and ordinary vehicles are used to transport things both in the various "small" conflicts in Africa and in the "special operation" war in Ukraine).

The Soviet Union kept using tubes for their aircraft electronics long after they were able to make much smaller and lighter circuits, because the tubes were much better at surviving EMPs (presumably from nuclear weapons).

chrispikula:
> In the real world there is a difference in strength between different kinds of EMP sources. Something that's shielded to resist industrial environments is probably not shielded against military EMP sources, which I would assume EMP bombs and grenades to be, and those are probably weaker than those from an atomic blast (at some distance). We could use the game assumption that all sources that are represented in the game emit military level EMP, of course, but then some sources should probably be removed (like arcing electrics).
> 
> Also note that protection against electricity isn't the same as protection against EMP. The metal shell of a car (one that isn't made out of composites, that is) protects against lightning strikes, but that would do little to protect against EMP, as the windows would provide access to the pulse.

The problem is that this is just gut feelings, though, and I'm pretty sure this is more TV and movie reality than actual reality.  High Energy Radio Frequency (HERF) Weapons do exist IRL, and are known to be able to disrupt computers via a targeted microwave pulse.  Not damage, not destroy, _disrupt_.  I know that there are appliance sized devices that can disrupt electronics around them such as old pacemakers, but the ones that I know of do so via rotating electromagnets, and take a lot of power.  MRI's, older mass spectrometers, inductor fusors, etc.

Remember, since any EMP would be an impulse of a radiating field, not a space-filling fluid, something like the metal components of a car _would_ do significant reduction of an EMP in their shadow.  You need to match conductive path lengths of your target to absorb the energy of an EMP, and generate the destructive voltage.   Power lines, street poles and such have many such paths.  Thankfully almost every electrical conductor we use inside our homes for that self-same purpose uses grounded shielded wire, other than if you still have copper pipes.  If your absorbing device is smaller than the EMP wavelength frequency, the energy will mostly bypass it.

Zireael07:
Here's something I found on the topic: https://www.thebugoutbagguide.com/faraday-cages-emp/

TLDR: Faraday cages can be easily made, but they will do zilch against high-energy EMP which is what I assume we're talking about (although interesting fact, a lightning blast is also an EMP). Also EMP can not only disrupt electronics, but also cause fires (c.f. most famous such natural event, Carrington Event, set fire to telegraph stations)

Also Wiki seems to agree with the previous poster that "NNEMP [non nuclear emp] generators can be carried as a payload of bombs, [cruise missiles [...] and drones, with diminished mechanical, thermal and ionizing radiation effects" The only handheld device that exists seems to be a disruption simulator, and since CDDA is currently emulating current tech levels and not 2040s anymore....

chrispikula:
> [](https://www.thebugoutbagguide.com/faraday-cages-emp/)
> 
> TLDR: Faraday cages can be easily made, but they will do zilch against high-energy EMP which is what I assume we're talking about (although interesting fact, a lightning blast is also an EMP). Also EMP can not only disrupt electronics, but also cause fires (c.f. most famous such natural event, Carrington Event, set fire to telegraph stations)

It depends on the wavelength of your generator.  Higher the frequency, shorter the wavelength, smaller the vulnerable target.  2.4GHz, microwave magnetron, is about 12cm, so as long as you've got a cage of 1.2 cm paths around it, you've got shielding.  (1/10th wavelength dipole issues, don't want to get to far into it, it's a crude rule of thumb for simple antennas).  30 GHz magnetrons to make a 1cm wavelength which is mostly stopped* by 1 mm mesh is going to be hard to find, given that we, today, aren't great at making those small.  [Here](https://www.cpii.com/product.cfm/8/30/160)'s a link to a company that makes them, providing 19 inch rack-mount versions.

*by mostly stopped I mean cuttoff attenuation.  That gets into frequency filters, for which the math is annoying and hokey.

High energy stuff is often conflated with nuclear, which is a whole different ballpark.  Faraday cages are themselves going to get hot and arc in those cases.  No idea what you'd need for protection of that stuff, and no one really wants to do any active testing of those anymore.  :-p

Night-Pryanik:
This could be done in subsequent PRs.

PatrikLundell:
@AJimmyU's suggestion would essentially kill CBM usage completely, as they're rare and take a long time to install. Getting completely fried out of several season's worth of scavenging and installation work once every few weeks (with intervals shortening as zombies evolve) is no fun at all, and it doesn't make sense to use such a system for people who are presumably well aware of these monster capabilities. The alternative to having CBMs fried by EMPs is to assume they're shielded to an above military grade level.

Night-Pryanik:
Instead of permanent frying, we could instead apply temporary disruption effect on CBMs, which could make them non-functional for some amount of time.

anoobindisguise:
So am I understanding this?
- EMPs don't necessarily permanently ruin an electronic item, but instead disrupt them (similar to how at the moment bionics get temporarily disabled instead of broken)
- faraday cages don't necessarily protect against EMPs but if you are covered head to toe in a full coverage faraday cage (most electric proofed armor) it will at least do more than nothing
- many electronics are emp shielded or emp resistant. We don't have vacuum tubes anymore but off the top of my head I can think of: 5 pt anchor, phase immersion suit, power armor, rm13 armor, military flight helmet, flashlights, many power tools, electric firestarter, NRE recorder, minigun, coilgun. So basically any simple circuit and any ultratech item should likely be exempt, leaving to be fried smartphones, laptops, EMF detectors and other "complex" circuits.

chrispikula:
Yup.  I'd add radios to that as well, mind you.  Transmitting/receiving devices almost always have internal shielding, ala [tiny metal boxes](https://3.imimg.com/data3/CL/WX/MY-2829346/rf-shielding-500x500.jpg).  The rest of the faraday cage is a layer of copper layered inside the pcb.

Maleclypse (approved):

PR #60725, opened 2022-09-03
Boiled and bleached bandages use one makeshift bandage instead of 3

Fris0uman:
#### Summary
Balance "Boiled and bleached bandages use one makeshift bandage instead of 3"

#### Purpose of change

boiled/bleached makeshift bandage is a makeshift bandage that has been boiled or bleached, it doesn't make sense to use 3 makeshift bandages to get one boiled/bleached bandage

#### Describe the solution

Change three to one

#### Testing
image

image

### DIFF
--- a/data/json/recipes/other/medical.json
+++ b/data/json/recipes/other/medical.json
@@ -45,7 +45,7 @@
     "time": "5 m",
     "batch_time_factors": [ 95, 4 ],
     "autolearn": true,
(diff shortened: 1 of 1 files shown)


casswedson:
the 3 water units used could probably go too; I guess they were there because we were boiling 3 bandages so... 3 water are needed

actually-a-cat:
The recipes make 3 though? Now you're multiplying 1 bandage into 3. (`charges` strikes once again)

Fris0uman:
You're right it does produce 3 bandages, nevermind then X)

PR #60729, opened 2022-09-03
Rename survivor telescope, make them less common in favor of binoculars.

Aerin-of-the-Toast:
#### Summary
None

#### Purpose of change

"Survivor" as a descriptor is overused to the point of being meaningless in some cases, and retractable telescopes _are_ a real thing. This change is purely out of pettiness. They're also rare in a world with binoculars, but they were surprisingly common to find atop radio towers.

#### Describe the solution

Rename to "retractable telescope" and replace one of the radio tower spawns with binoculars instead (the one that spawns with a map, the sniper one still has a telescope). Also added a thing about pirates in the description, because the only time I think you'd see one of these antiques is in a museum or someone's pirate cosplay. 

"Collapsible" or "monocular" telescopes were also ideas, but retractable telescope seemed the simplest. Just calling it "telescope" would've technically been accurate as well but that could be confused for the kind you use to look at the stars.

#### Testing

Tested to make sure the name changed properly and binos spawned at a radio tower instead of the telescope.

#### Additional context

None

### DIFF
--- a/data/json/items/tool/misc.json
+++ b/data/json/items/tool/misc.json
@@ -649,8 +649,8 @@
   {
     "id": "survivor_scope",
     "type": "TOOL",
-    "name": { "str": "survivor telescope" },
-    "description": "A homemade collapsible telescope.  Too large and inaccurate to use as a weapon scope, but carrying this item in your inventory will double the distance that is mapped around you during your travels.",
+    "name": { "str": "retractable telescope" },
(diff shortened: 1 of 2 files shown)


GuardianDll:
Tbh i always thought about it as about this old telescope-spyglass thingy, but then i google, and found there is a real thing named "spotting scope", which looks and works exactly as survivor need to. Yes, its also way more rare than binocular

PR #60730, opened 2022-09-03
[Magiclysm] Add ELECTRIC_IMMUNE to Stormshaper class

GuardianDll:
#### Summary
Mods "Magiclysm's stormshaper now immune to electricity"

#### Purpose of change

I mean really? The guy that throws a 1000 volt charges to enemies must suffer from his own attacks?

#### Describe the solution

Add ELECTRIC_IMMUNE to Stormshaper class

Not adding it
Add some armor against electric damage

#### Testing

Works great, now i can throw lightnings and be not afraid of my own attacks

### DIFF
--- a/data/mods/Magiclysm/traits/classes.json
+++ b/data/mods/Magiclysm/traits/classes.json
@@ -45,6 +45,7 @@
     "purifiable": false,
     "valid": false,
     "cancels": [ "KELVINIST" ],
+    "flags": [ "ELECTRIC_IMMUNE" ],
     "spells_learned": [ [ "create_rune_stormshaper", 1 ] ]
   },
   {


casswedson:
it would be a nice idea to add this to the class, but not an addition you get right after you learn your first stormshaper spell, this is like becoming fire immune after learning finger firelighter

fearing the power of your own spell is something that I think fits the game's themes and probably magiclysm's too

there's definitely a place and need for resistances to your own powers or just a damage reduction would be cool

GuardianDll:
That make sense, but its hard to me to design how such "permanent passive effects" may work in the game. If you have some good ideas feel free to describe them here, as separate pr, or in discord (just ping me so i wont miss it)

casswedson:
this in https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md#traitsmutations would do the trick
```json
"armor" : [ // Protects selected body parts this much. Resistances use syntax like `PART RESISTANCE` below.
  {
    "part_types": [ "arm", "leg" ],
    "bash": 2,
    "electric": 5                  // The resistance provided to the body part type(s) selected above
  },
  {
    "parts": [ "arm_l", "arm_r" ], // Overrides the above settings for those specific body parts
    "bash": 1,                     // ...and gives them those resistances instead
    "cut": 2
  }
],
```

GuardianDll:
oh. i mean not the particular stormshaper effect, i can easily make the similar thing using enchantment system - it has `ARMOR_ELEC` which can do this
what i mean is "how it wokrs" part - it must be some spell, that you train, and in the end you give a permanent armor (kinda cool, but i dont like it, because it means you will be left with spell, that do nothing, and you cant delete/forget it (or maybe EoC can do the trick, i dont know)), or just a rare item, that you can use and take an electric armor forever (also dont like it, because magiclysm already has a ton of rare, but almost unobtainable items)

Maleclypse:
Ok só i’d make an eoc tied to the stormshsper trait that checks for x class spells at level 15 and when condition is met it grants a trait that provides zap immunity and some electricity armor. Immune seems a bit much since we could theoretically be talking hundreds of damage points.

GuardianDll:
thats a great idea
ill process it in the future then

PR #60734, opened 2022-09-03
Sheet metal armor, round 3: Revengeance of the Case Hardening

Aerin-of-the-Toast:
#### Summary
Content "Case hardened sheet metal armor"

#### Purpose of change

Case-hardening mild steel into something much stronger is something that'd be pretty viable to do to sheet metal armor, as a quick-and-dirty way of taking crude armor and making it stronger.

#### Describe the solution

A helluva lot of new "case hardened" sheet metal armors, one version for every armor piece. The case hardened stuff is just a strict upgrade, having better defensive stats and being sturdy, but requires more effort (and a book)to make and repair, as well as needing an actual forge set up. I've also added a "heavy" variant of the chest guard, which is just the same as the regular version but with doubled-up layers, and also added a case hardened version of that too.

Lastly, I added full suit options that combine all the chest, arm, and leg pieces together for full 95% coverage; this is mostly a prophylactic move for when I add variants with chainmail connectors later. It also lets you case harden the whole set faster than doing the pieces individually.

I've also made a few adjustments, notably upping the chestguard's encumbrance by quite a bit. Instead of 10, it's now 18 (and the heavy's chestguard is 30).

Doing the chainmail'd versions in this PR, but it was already getting pretty unwieldy large.

Saving ANickle some work and trying to nest the recipes myself, but I probably would've screwed up their organization system and my brain was starting to melt so I didn't trust myself not to break something.

#### Testing

A lot of checking the stats, if the game loads, if the recipes exist and are in the right place. Pretty sure I got everything.

#### Additional context

One more PR after this for the suits with chainmail underneath and I'll be done with junkyard knight, I promise.

### DIFF
--- a/tools/spell_checker/dictionary.txt
+++ b/tools/spell_checker/dictionary.txt
@@ -1308,6 +1308,8 @@ parthenogenetic
 partyin
 pasteurize
 pathfinding
+pauldron
+pauldrons
 peanutbutter
 peccari
 pecial
@@ -1550,6 +1552,8 @@ rugerlcp
 ruleset
 rummagin
 s
+sabaton
+sabatons
 saboted
 sadisticly
 safemode


ANickelN:
Don't worry about the nesting bit, your stuff is next on the list, especially if you're doing the light, normal, heavy variant thing like Drew did.

PR #59165, opened 2022-07-11
NPC follow-up quests

onura46:
#### Summary
Content "Add follow-up quests in NPC backgrounds"

#### Purpose of change
Adds follow-up quests and dialog to visit locations from three generic NPC backgrounds

#### Describe the solution
Added three quests and conditional dialogs that were commented into the code as a to-do by @I-am-Erk (I believe) about three years ago. The NPCs with hospital_2, rural_1, and prepper_1 backgrounds now give you options to persuade the NPC to give up the location of their cabin, infested farm, and LMOE, respectively.

The conditional variable checks look kind of insane on paper but they give the dialog a nicely natural feel. If you ask the NPC about the location and fail, they get offended and you won't get that dialog option for 8 hours. Once you complete the quest, the dialog changes and no other NPC of the same type will bring it up to you to prevent exploiting the NPC for locations or having unnatural dialog doubling, assuming you live that long. Idk, I'm just sort of proud of it.

#### Describe alternatives you've considered
None

#### Testing
Tested on my own compiled version.

PR #60737, opened 2022-09-04
Prepper_1 NPC background expansion and locked LMOE computer

onura46:
#### Summary
Content "Add a computer and connected NPC dialog to locked LMOE"

#### Purpose of change
Follow-up to #59515; adds a computer inside the locked LMOE for storytelling context and conversation with the related NPC.

#### Describe the solution
Follow-up to #59515. This PR is in response to a couple people who have been confused by the locked LMOE from the prepper_1 quest. Since it's pretty unlikely for players to meet this NPC (especially those who don't interact with them), I thought it would be pertinent to imply a little bit of their story in a computer located inside their LMOE. There are also natural dialog options for confronting them about it, depending on which order you discovered the prepper and the LMOE.

The computer in the LMOE shows a conversation between the prepper and the previous owner of the land. It also contains personal documents and conversations showing the prepper is a bit of a conspiracy theorist. At a certain point, it became a little hard to wrap up the conversations without exploring this character's mind a little, so (if you're nice to them) you can chat with them a little bit about the apocalypse, including telling them about Hub 01 and Brigitte, the Bone Seer. I wanted this dialog to be natural and respond to the player's previous adventures.

Another related minor tweak: the Locked LMOE is now a globally unique spawn instead of just unique. This was an oversight of my original PR.

I thought about just leaving the shelter unlocked. However, not only does this kind of ruin the original intent of the mission - reward players who are interacting with and doing quests for an NPC - but I don't think _anyone_ would leave their valuable shelter unlocked if they planned to return to it. Locked shelters should probably realistically be quite common. Anyway, I would bet that the likelihood that players interact with this computer versus the NPC is very high, and it would (at least, in my opinion) make it pretty cool to realize you stumbled upon their secret LMOE and read their secret diary before you even met them.

#### Testing
Tested on my own compiled version. I went through the dialog trees every way that I could think of. 

#### Additional context
**SPOILERS BELOW**

Snooping through the Prepper's documents:

ChattingWithComputer

Chatting about the apocalypse with the Prepper after confronting them:

ChattingWithPrepper


RelMayers:
Lore dump addition sounds bad, also LMOE aren't bunkers, Computer is drain on power which would be bad preparation. I don't see people making their home bunkers for survival showing off their Computers there.

onura46:
> Lore dump addition sounds bad, also LMOE aren't bunkers, Computer is drain on power which would be bad preparation. I don't see people making their home bunkers for survival showing off their Computers there.

I understand what you're saying. Fwiw, this NPC is just as clueless - probably moreso - than the average citizen pre-cataclysm, so you're getting less a lore dump (as in: sound, factual information about what really happened) and instead just another biased view of the confusion before the start date of the game. Read a bit of the dialog that I added and it should become pretty obvious the prepper is _not_ on the right track lmao.

Zireael07:
PR title claims it adds a computer but looking at the checklist and commits reveals there are many other changes... so which is it?

onura46:
> PR title claims it adds a computer but looking at the checklist and commits reveals there are many other changes... so which is it?

It's a work in progress.

Zireael07:
Yes, but the title claims it adds a computer, while the rest of the OP and your comments seem to indicate the scope is much much larger. The title should be changed to reflect the contents better

onura46:
> Yes, but the title claims it adds a computer, while the rest of the OP and your comments seem to indicate the scope is much much larger. The title should be changed to reflect the contents better

It's really not _much_ larger. It's just not a half-assed implementation of an idea.

onura46:
And actually, I spent a lot of time today trying to wrap it up and I think it's at a good enough place for open critique. It's going to stay in draft form for a couple days while I do more testing, and run the story/lore-related stuff by longer-standing community members. The title should properly reflect the scope of the PR; it was the last thing I was thinking about until the code was finished tbh.

onura46:
This keeps inheriting the "Lore" tag because I mentioned the word in the very first PR related to this NPC. It wasn't super important but now it might be a problem. These PRs are not related to lore (this, #59515, and #59165). If someone with privileges could remove that tag, that'd be appreciated! <3

I-am-Erk (approved):
I think the only trouble we have here now is that this is a really epic story that none of our other backgrounds meet in quality.

I haven't checked the loot in the shelter but even though it's been somewhat looted we should make sure it's a bit better than your average LMOE because of the rarity adn difficulty of getting it.

PR #60745, opened 2022-09-04
Item snails mutate into Big Snails

Karol1223:
#### Summary
Balance "Makes the item version of snails mutate via the hatch code into the smallest mutant snail variant"

#### Purpose of change
Since all our mutant bugs are just mutated versions of regular bugs it felt weird to have small item snails available this way indefinitely. They didn't spoil, they didn't mutate. They just stayed in this peculiar limbo.

#### Describe the solution
Uses the code responsible for egg hatching to make the item snails 'hatch' into Big Snails after the time of 30 days (takes 15 days to rot).

Initially, I had the rot timer set to 30 days as that's the time it takes for most insect nymphs to grow up but the hatching code utilises twice the rot timer so I took it down to 15.

#### Testing
Booted up the game and fast forwarder time 60 (this was before the rot timer got lowered) days with a spawned snail item on the ground. The snail item rotted and in its place came a Big Snail.

### DIFF
--- a/data/json/monstergroups/misc.json
+++ b/data/json/monstergroups/misc.json
@@ -472,9 +472,6 @@
   {
     "name": "GROUP_MUTATION_SNAIL",
     "type": "monstergroup",
-    "monsters": [ 
-      { "monster": "mon_null", "weight": 85 },
-      { "monster": "mon_snail_forest", "weight": 15 }
-    ]
+    "monsters": [ { "monster": "mon_null", "weight": 85 }, { "monster": "mon_snail_forest", "weight": 15 } ]
   }
 ]


Xaritscin:
the indefinite shelf life was to treat them as living items so players would have these gastropods sitting around vibing until they got purged for cooking. the mutant variants were added to follow with the giant slugs that were already in mainline before this was expanded as swamp spawns.

dont have much to say about this PR except slugs may also be missing the hatching tag.

Karol1223:
> the indefinite shelf life was to treat them as living items so players would have these gastropods sitting around vibing until they got purged for cooking. the mutant variants were added to follow with the giant slugs that were already in mainline before this was expanded as swamp spawns.
> 
> dont have much to say about this PR except slugs may also be missing the hatching tag.

Still should mutate as theoretically anything does. The idea is that insects and other critters mutate as Cataclysm continues, so if everything around mutates small snails should too, especially if we already have something to mutate them into. Slugs don't have an item form.

Xaritscin:
ok had to doublecheck but yeah it appears there was only garden snails for bushes. makes sense i guess, slugs arent really a snack.

anothersimulacrum:
Not everything inevitably has to mutate. Some have, but most won't. I'd advise a very very low chance for the rot spawn if at all (should be able to make the monstergroup mostly null).

Karol1223:
> Not everything inevitably has to mutate. Some have, but most won't. I'd advise a very very low chance for the rot spawn if at all (should be able to make the monstergroup mostly null).

Actually I guess that is fair. If this was the case mutant bug density would be far higher. I'll change it to be a chance tomorrow.

data/json/monstergroups/misc.json
```diff
@@ -468,5 +468,13 @@
       { "monster": "mon_spawn_raptor" },
       { "monster": "mon_spawn_raptor_shady" }
     ]
+  },
+  {
+    "name": "GROUP_MUTATION_SNAIL",
+    "type": "monstergroup",
+    "monsters": [ 
+      { "monster": "mon_null", "weight": 85 },
+      { "monster": "mon_snail_forest" "weight": 15 }
```
GuardianDll, line 477:
```suggestion
      { "monster": "mon_snail_forest", "weight": 15 }
```

PR #60752, opened 2022-09-04
[Magiclysm] Make Summoning Great Again!

GuardianDll:
#### Summary
Mods "[Magiclysm] Make Summoning Great Again!"

#### Purpose of change

Right now all summoning (both creature and item summons) spells suffer from one similar problem - they summon stuff for too little time. For example, all animist summons lasts only for **15 mitutes at max level** (and the cast itself consume half of minute). And technomancer's animated blade summons only for **45 seconds at max level**. How you can prepare anything with such tiny amount of time? they just despawn themselves when you'll be ready. Initially i thought the reason is DnD inspiration, where effects can last only for few turns, but then i checked and found their summoning lasts from hour till the whole day

#### Describe the solution

Buffed all summoning spells in all schools - the most beneficial from this pr is an animist class
All creature summons has now 30 levels, effect starts from 6 minutes, and lasts for 3 hours at max lvl (+6 minutes for each level)
Each item summon spell has 15 levels, effect starts from 4 minutes, and lasts for 1 hour at max (+4 minutes for each level).
At max level of item summon spell player learn improved version of this spell - it has 25 levels, **can be trained only by using the spell itself (no more reading books)**, has higher difficulty, but has less penalties (like lack of CONCENTRATE flag), decrease the mana consumption and cast time with each level, and can lasts from 1 hour at first level to 6 hours at max level (+12 minutes for each level)

also i buffed a flame sword item to deal more fire damage, to make it more useful

Left summoning useless

#### Testing

Finally
i can make a small army, as any another normal summoner

Theyre still pretty dumb tho, but its better than before

#### Additional context

Draft, because i need some huge sanity check from the side

### DIFF
--- a/data/mods/Magiclysm/Spells/kelvinist.json
+++ b/data/mods/Magiclysm/Spells/kelvinist.json
@@ -450,7 +450,7 @@
     "id": "kelvinist_summon_flamesword_plus",
     "type": "SPELL",
     "name": "Enhanced Flamesword",
-    "description": "This spell ignites a wooden sword, creating a flaming blade able to burn, bruise, and chop. Now you know this spell like the back of your hand, and start to design your own version.",
+    "description": "This spell ignites a wooden sword, creating a flaming blade able to burn, bruise, and chop.  Now you know this spell like the back of your hand, and start to design your own version.",
     "valid_targets": [ "self" ],
     "flags": [ "NO_LEGS", "NO_HANDS" ],
     "effect": "spawn_item",


data/mods/Magiclysm/Spells/animist.json
```diff
@@ -89,9 +89,9 @@
     "min_aoe": 2,
     "max_aoe": 2,
     "flags": [ "SOMATIC", "VERBAL", "CONCENTRATE" ],
-    "min_duration": 30000,
-    "max_duration": 150000,
-    "duration_increment": 4000,
+    "min_duration": 36000,
```
Maleclypse, line 92: I'm guilty of not doing this but I think we should start adding comments to a spell that states the min and max duration in hrs minutes etc instead of only in seconds.   That would make it more human readable, I'm saying.

GuardianDll: id prefer to have some readable values in the code instead, so not `36000`, but `6 m`

Maleclypse: I'll make a feature request.

chrispikula:
If you want to make a radical change, you could make it so that summons, until dismissed, keep part of the maximum mana pool locked, which doesn't get a chance to refill until the summon is dismissed, either through a long time, or a secondary spell.  Probably also leaving a certain radius of the PC, so reality-bubble shenanigans don't occur.

I'm thinking more than a 1000 reduction in maximum mana for a base level summon here, btw, not something trivial.  And a duration lasting over a day.  

Not sure how you would level the spell.  But it could use less max mana as it does so.

GuardianDll:
> keep part of the maximum mana pool locked
thats a good idea, and it can work with items, but not with monsters you summon

PR #60754, opened 2022-09-04
Better version.h generator script for MSVC builds

akrieger:
#### Summary
None

#### Purpose of change
The prebuild script always runs and always prints some sort of output. Plus, it always updates the mtime on the version.h file which upsets Visual Studio's incremental build. Fixing both helps my OCD with not having output for no net changes and makes the IDE happier on sequential builds.

#### Describe the solution
Grep for the version string in version.h and only overwrite version.h if the version is different than what's encoded in it.

- Write the new version.h to a temp file and diff the existing one (which is the first thing I did but grep is strictly superior).

#### Testing
- Run it sequentually and only get output the first time.
- Delete the existing version.h and rebuild and it recreates version.h
- Change branches, cherry pick the commit over, build, it recreates the right version.h


akrieger:
Wait I'm dumb, there's a grep equivalent I can use.

PR #60760, opened 2022-09-04
removed useless BELT_CLIP flag from war hammer

Terrorforge:
#### Summary
Bugfixes "removed useless BELT_CLIP flag from war hammer"

#### Purpose of change

The war hammer has the BELT_CLIP tag, which causes it to display the "this weapon can be clipped onto a belt loop" message in its description. However, at 120cm long, it's too long to actually fit into any belt in the game. Thus, the only thing the BELT_CLIP flag does is confuse and annoy players.

#### Describe the solution
Removed the flag.

It's unclear to me exactly what kind of weapon the war hammer is supposed to be. Its stats seem to be inbetween those of a one-handed "horseman's hammer" type weapon and a two-handed polearm hammer like a bec de corbin. I considered altering it to more closely reflect the one-handed type, in which case I would've reduced the length and let it keep the BELT_CLIP flag, but that would basically mean rewdoing the whole thing from scratch. I might still do that at some point, but for now I'm just getting rid of this weird UI thing.

### DIFF
--- a/data/json/items/melee/bludgeons.json
+++ b/data/json/items/melee/bludgeons.json
@@ -1476,7 +1476,7 @@
     "//": "Craftable from steel, shouldn't be silver.  A warhammer is essentially one end of a pickaxe with balancing weight on the other and crushes/pierces the armor.",
     "material": [ "steel", "wood" ],
     "techniques": [ "WBLOCK_1", "BRUTAL", "SWEEP" ],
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):

catdach:
Yeah, the Lucerne hammer is the polearm version. Not sure why the Warhammer is at 120 cm.

PR #60764, opened 2022-09-04
Add urban explorer profession

garnetgw:
#### Summary
Content "Adds an urban explorer profession with an urban explorer's map item that unveils nearby abandoned buildings"

#### Purpose of change
I wanted to make a "burglar lite" profession, and expanded the concept into this B&E photographer.

#### Describe the solution
The urban explorer costs 1 less point than the burglar, and starts with only a crowbar and the lockpicking proficiency, but has a map that reveals local roads and abandoned buildings and has low-encumbrance storage capacity.

1 point cost? the points don't matter and everything is made up

#### Testing
Inserted changes into a fresh copy of latest experimental and loaded urban explorer into several worlds to check that intended gear spawns and the map functions correctly and spawns in zombie small bags.

#### Additional context

urbexer

### DIFF
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -5467,7 +5467,6 @@
           "sneakers",
           "hoodie",
           "tshirt_text",
-          "pants",
           "crowbar",
           "camera_bag",
           "slingpack",


TheShadowFerret:
You need a summary or GitHub will be angry

```Summary```
```Content "Urban Explorer profession"```

Amoebka:
The map could be added as an uncommon drop to general zombie loot tables, doesn't seem to be a reason to make it start-only.

PR #60767, opened 2022-09-04
Disable cut/remove grass under vehicles

### LINKED ISSUE #54317: Vehicles block resuming partial deconstructions
### Describe the bug

If you begin a deconstruction that is under a vehicle tile (a shopping cart that got stuck because it could be pulled *onto* but not *off of* tall grass) and interrupt it, `e`xamining the tile only offers vehicle prompts and not 'continue construction' prompts. 

### Steps to reproduce

Pull a shopping cart over onto a patch of grass of any length
`*`, remove thata tile of grass
Interrupt the process with `.`
`e`xamine the tile, observe that there is no prompt to continue removing the grass.

Remove the shopping cart. 
`e`xamine the tile, observe that there is now a prompt to continue the construction and if refused there is a followup prompt to cancel. 

### Expected behavior

I expected to be able to resume tasks that can normally be resumed. I *can* just move all my heavy things out of the shopping cart long enough to get it off the tile and put them back, so it's not game-breaking, but this seems like an interface error, rather than just a hassle. If it's not suppose to be possible to deconstruct grass under vehicles, then it being impossible to continue would make sense, but it should then be impossible to begin.

### Screenshots

_No response_

ongoing issue, still present in 0.F-5229-g55b805eb93.

irwiss:
#### Summary
None

#### Purpose of change
Fixes #54317

#### Describe the solution

Disable cut/remove grass under vehicles

EDIT: reverted the vehicle check from can_construct and used empty_check_lite in json instead, I kept some of the tidying up;

can_construct had bitwise &= which [seems to block short circuiting even in trivial cases](https://godbolt.org/z/P8d8bosqj) and queried map::furn/map::ter twice because it shadowed the names of the previous calls - changed it use normal boolean logic instead of bitwise &

Also make explain_fail_map in load_construction static, and select the correct key using the string instead of constructing new map per load_construction call

#### Testing

Replicate scenario in #54317 - grass tile under vehicle (or items/monsters/any of the empty_check_lite conditions) shouldn't be selectable for construction.

Walk to a window with a wall near it, select deconstruct, select the wall, should add message "You cannot deconstruct this!"
Make crafting spot, select tile with the wall should add message "You cannot build there"

### DIFF
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -3983,6 +3983,7 @@
     "required_skills": [ [ "survival", 0 ] ],
     "time": "4 m",
     "byproducts": [ { "item": "straw_pile", "count": [ 1, 3 ] } ],
+    "pre_special": "check_empty_lite",
     "pre_terrain": "t_grass_long",
     "post_terrain": "t_grass"
   },
@@ -3995,6 +3996,7 @@
     "required_skills": [ [ "survival", 0 ] ],
     "time": "5 m",
     "byproducts": [ { "item": "straw_pile", "count": [ 2, 4 ] } ],
+    "pre_special": "check_empty_lite",
     "pre_terrain": "t_grass_tall",
     "post_terrain": "t_grass"
   },
@@ -4007,6 +4009,7 @@
     "required_skills": [ [ "survival", 0 ] ],
     "time": "4 m",
     "byproducts": [ { "item": "straw_pile", "count": [ 1, 2 ] } ],
+    "pre_special": "check_empty_lite",
     "pre_terrain": "t_grass_dead",
     "post_terrain": "t_grass"
   },
@@ -4019,6 +4022,7 @@
     "required_skills": [ [ "survival", 0 ] ],
     "time": "40 s",
     "byproducts": [ { "item": "straw_pile", "count": [ 3, 6 ] } ],
+    "pre_special": "check_empty_lite",
     "pre_terrain": "t_grass_long",
     "post_terrain": "t_grass"
   },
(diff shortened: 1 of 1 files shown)


andrei8l:
There are 2 slightly misnamed `pre_special`s that are supposed to handle this, but some definitions don't use them - `check_empty` (should be `check_empty_flat`) and `check_empty_lite` (should be `check_empty`).

IMO it'd be better to spread those special checks to construction definitions or even make them default, rather than hard-coding a check.

irwiss:
Not sure what construction work can be viable under a vehicle but I'll take a fresh look tomorrow

PatrikLundell:
Building a roof over an appliance (or vehicle) seems reasonable to me, as long as the floor remains unchanged (that's what I could come up with for now, anyway).

Qrox:
I remember there has been a hack to place a vehicle across a door and board up the door to make a walk-in freezer. It would be nice if we can make this an actual feature by allowing constructing special walls on a vehicle tile (think of it as cementing the whole tile into place).

PatrikLundell:
You can do it in the reverse order currently by placing an appliance on a wall tile, as with wall wiring. However, placing the appliance first and constructing the wall around it makes a lot of sense for specific applications.

irwiss:
I suppose it makes sense to do the json check instead, I reverted the veh_at check and kept/cherry picked some of the c++ changes i had; the bitwise to boolean logic change and made explain_fail_map static instead of constructed for every construction.

Added the pre_special entry for all cut and remove grass constructions.

I also checked if check_empty_lite can be defaulted to as andrei8l suggested but it doesn't look like it, at least not easily since some constructions run the validation checks via flags, others require furniture/terrain.
Making it the new default and overriding with the current would add 318 lines to the json, when there's around 600ish constructions total so probably not worth it either.

PR #60770, opened 2022-09-04
CR Steel crafting guide

Drew4484:
#### Summary
Content "Documentation for steel crafting"

#### Purpose of change
Adds a document with established guidelines for the various steel grades and crafting/working with them.

#### Describe the solution
Lots of text for both items and recipe guidelines.

Continuing to keep things vague and having to re-do my research each time I want to add more steel items.

#### Testing
WIP but it's a .md anyway.

#### Additional context
I am **NOT** a professional and none of this should be taken as real world advice. Furthermore I will gladly accept any corrections or feedback from people who know better.

### DIFF
--- a/doc/STEEL_CRAFTING.md
+++ b/doc/STEEL_CRAFTING.md
@@ -4,11 +4,11 @@
 
 # Defining Steel Grades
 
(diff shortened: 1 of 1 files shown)


Aerin-of-the-Toast:
I love the part standardizing the fabrication requirements, that'll make it easier to keep things consistent. I wonder if it'd be worth denoting which processes should be book learn only vs autolearn too?

Drew4484:
> I love the part standardizing the fabrication requirements, that'll make it easier to keep things consistent. I wonder if it'd be worth denoting which processes should be book learn only vs autolearn too?

I'll add more on the books required and check if there are any others I should include.

Issue #57152, opened 2022-04-25
Game preformance issue


Jninja1:
### Describe the bug

The game begins to notably slow down once four or more solar panels are connect to each other with an extension cord of sorts is connected to one leading to a battery.

### Steps to reproduce

1) Install solar panels as an appliance
2) Install any sort of battery as an appliance
3) Connect panels to battery or vice versa 

### Expected behavior

I expected the game to run with no noticeable slow down and provide power to the batteries.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.22000.613 (21H2)
- Game Version: 44c51a4 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Blaze Industries [blazeindustries],
    Cataclysm++ [Cata++],
    Locked and Loaded [Locked_and_Loaded],
    MADA Mod [MADA_bows_rebalance],
    SpeedyDex [speedydex],
    Stats Through Kills [stats_through_kills]
]

Inglonias:
can you provide the save file where this is happening, or a screenshot explaining your power setup?

robob27:
I can provide a save if needed but it's very easy to repro. I tested how long it took my PC to wait 6 hours before and after making this setup (linear, one connection between each device):

![image](https://user-images.githubusercontent.com/2104555/186037211-887894f2-fd89-4119-9b4c-cbbe5325a2ce.png)

Before setup: 11.32s
After setup: 17.83s

I made a slightly bigger grid than the creator of the issue described. This is because when I tested the grid setup described by the creator, there was only about a 1 second difference on my hardware (AMD Ryzen 7 3700X @ 3.60 GHz) but I imagine even their setup is noticeably slower on older hardware. It scales very quickly, adding only a few more solar panels and an oven and I'm consistently waiting about 6 seconds longer for sleeping.. and this is still a very simple setup.

I first noticed this issue myself the other day. After setting up ~15 solar panels, a battery and several appliances, it all of a sudden took _forever_ to sleep/craft/read at my base (this was on slower hardware, a 2.30GHz cpu). I checked github and saw there had previously been some issues with setups with parallel connections, or many redundant connections, so I redid the setup to make sure there was none of that, but it was still just as slow.

robob27:
[...]

PR #60776, opened 2022-09-05
Vehicle/Appliance performance improvements

robob27:
#### Summary
Performance "Misc vehicle/appliance performance improvements"

#### Purpose of change
Improve performance of vehicle code that gets called every turn and causes some noticeable slowdown with even a minimal appliance setup. This doesn't completely resolve the issue, but I used the debug turn timer to take measurements of the irl time before and after applying the changes, and there does appear to be an improvement.

Related to #57152 

#### Describe the solution
- Switch some containers in `vehicle::traverse_vehicle_graph` to `std::vector`:
    - `connected_vehs` only gets modified before we start iterating, and after we're finished with `connected_veh`, so we don't really need a guarantee that iterators are not invalidated, which I think was the purpose of using a queue here?
    - I initially changed `visited_vehs` and `visited_targets` to `std::unordered_set` (which also showed a performance improvement) since the order of the elements didn't appear to matter. After that, I experimented with a vector thinking that, although we would need to iterate to check if they contained a certain element now, maybe the performance allocation/deallocation/pushing elements would outweigh that. Switching these to vectors showed additional improvement over `std::unordered_set` in the save I was testing. I could see `std::unordered_set` performing better here in situations with really big vehicle graphs, but I wonder how big it would need to be (and how realistic that would be) to outweigh the vector speed. I want to test this more.
- Remove an unnecessary variable assignment from `precalc_mounts` 
- Change `precalc_mounts` to `void`. The return value isn't used anywhere after removing the assignment mentioned in the last point
- Remove the redundant check `visited_vehs.count( target_veh ) < 1` which happened right after we `continue` if `target_veh == nullptr || visited_vehs.count( target_veh ) > 0`. It has to be less than 1 if we didn't continue above.
- Only call `vehicle::traverse_vehicle_graph` with the `nil_visitor` once every 5 turns instead of every turn.

After the performance related changes above, I am seeing about a 1.7 irl second decrease per in game hour passed.

#### Testing
- Enable hour debug timer
- Pass time waiting with and without changes to compare irl time passed
- Confirm batteries are still being charged/discharged, appliances still work etc

#### Additional context

**Disclaimer:** this is the first time I've really tried to dive into both the performance of different STL containers and appliance code. Some of my understandings/takeaways from the research I did may be incorrect. This should not be taken at face value and information in this PR should not be used as a reference by anyone. I'm trying to learn 😅 

This is the setup used in these tests (from my save where I first experienced the slowdown):

- 16x assorted solar panels
- 1x large storage battery
- 7x wall wiring
- 1x drill press
- 1x hydraulic press
- 1x air compressor
- 1x mounted electric kiln
- 1x fridge
- 1x oven
- 1x actual vehicle plugged in
- 2x extension cords to connect wall wirings separated by more than 1 space

image

Without changes:
image

With changes (excluding `nil_visitor` changes - mentioned my hesitation about this in a comment below):
image

All changes including `nil_visitor` change:
image

With all changes + changes from #60681 😍:
image

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -3105,7 +3105,7 @@ tripoint vehicle::mount_to_tripoint( const point &mount, const point &offset ) c
 }
 
 void vehicle::precalc_mounts( int idir, const units::angle &dir,
-        const point &pivot )
+                              const point &pivot )
 {
     if( idir < 0 || idir > 1 ) {
         idir = 0;
@@ -5106,7 +5106,7 @@ int vehicle::traverse_vehicle_graph( Vehicle *start_veh, int amount, Func action
         return amount;
     }
     // Breadth-first search! Initialize the queue with a pointer to ourselves and go!
-    std::vector< std::pair<Vehicle *, int> > connected_vehs = std::vector< std::pair<Vehicle*, int> >{ std::make_pair( start_veh, 0 ) };
+    std::vector< std::pair<Vehicle *, int> > connected_vehs = std::vector< std::pair<Vehicle *, int> > { std::make_pair( start_veh, 0 ) };
     std::vector<Vehicle *> visited_vehs;
     std::vector<tripoint> visited_targets;
 
@@ -5646,12 +5646,13 @@ void vehicle::gain_moves()
     }
 
     // Force off-map vehicles to load by visiting them every time we gain moves.
-    // Shouldn't be too expensive if there aren't fifty trillion vehicles in the graph...
-    // ...and if there are, it's the player's fault for putting them there.
-    auto nil_visitor = []( vehicle *, int amount, int ) {
-        return amount;
-    };
-    traverse_vehicle_graph( this, 1, nil_visitor );
+    // This is expensive so we allow a slightly stale result
+    if( calendar::once_every( 5_turns ) ) {
+        auto nil_visitor = []( vehicle *, int amount, int ) {
+            return amount;
+        };
+        traverse_vehicle_graph( this, 1, nil_visitor );
+    }
 
     if( check_environmental_effects ) {
         check_environmental_effects = do_environmental_effects();


src/vehicle.cpp
```diff
@@ -5109,18 +5110,18 @@ int vehicle::traverse_vehicle_graph( Vehicle *start_veh, int amount, Func action
...
     while( amount > 0 && !connected_vehs.empty() ) {
-        auto current_node = connected_vehs.front();
+        auto current_node = connected_vehs.back();
         Vehicle *current_veh = current_node.first;
         int current_loss = current_node.second;
 
-        visited_vehs.insert( current_veh );
-        connected_vehs.pop();
+        visited_vehs.push_back( current_veh );
+        connected_vehs.pop_back();
```
robob27, line 5124: I want to confirm my understanding here that doing this is okay:

```
auto someElement = vector.back();
vector.pop_back();
```

Because although documentation on `std::vector` says that `pop_back()` invalidates iterators and references to the last element/end, we are not using a reference or iterator here. We are copying the value to `current_node` and so deleting from the vector has no impact on `current_node`?

src/vehicle.cpp
```diff
@@ -5129,38 +5130,39 @@ int vehicle::traverse_vehicle_graph( Vehicle *start_veh, int amount, Func action
...
-                add_msg_debug( debugmode::DF_VEHICLE, "After remote %p, %d power",
-                               static_cast<void *>( target_veh ), amount );
-
-                if( amount < 1 ) {
-                    break; // No more charge to donate away.
-                }
+            // but only if we haven't seen this one before (checked above)
+            int target_loss = current_loss + current_veh->part_info( p ).epower;
+            connected_vehs.push_back( std::make_pair( target_veh, target_loss ) );
+            // current_veh could be invalid after this point
```
robob27, line 5153: Based on my other comment I think I was incorrect when I wrote this, since `current_veh` is not a reference or iterator?

src/vehicle.cpp
anothersimulacrum: Correct.

robob27: Thank you!

robob27:
I did some more testing locally with lowering the frequency at which we load off-map vehicles that are connected to on-map vehicles. I was able to save another ~700ms per in game hour with my setup by doing this every 5 turns instead of every turn.

image

I modified [this code](https://github.com/CleverRaven/Cataclysm-DDA/blob/ed056e940da0d87fe35dd5ae7306973fae7fc3a1/src/vehicle.cpp#L5651-L5657) as shown below:

```
// Force off-map vehicles to load by visiting them every time we gain moves.
// Shouldn't be too expensive if there aren't fifty trillion vehicles in the graph...
// ...and if there are, it's the player's fault for putting them there.
if ( calendar::once_every( 5_turns ) ) {
    auto nil_visitor = [](vehicle*, int amount, int) {
        return amount;
    };
    traverse_vehicle_graph(this, 1, nil_visitor);
}
```

I'm a bit fuzzy on _exactly_ how it works, but every turn does seem maybe a bit excessive with all of the vehicles we can reasonably be expected to have nearby. If doing it every turn is just to ensure we get the freshest result, perhaps we could back off a little here? I don't yet understand if this is something that _needs_ to happen every turn for offmap vehicles to work at all.

robob27:
Testing these changes in tandem with #60681 - after merging all of the vehicles in my setup together.. 😍 (this is only 200ms slower than a save I tested with no appliances nearby at all)

image

Maleclypse:
Did you intend for this to still be in draft?

Maleclypse:
> > Did you intend for this to still be in draft?
> 
> Yeah I haven't yet tested how it performs in builds other than MSVC (not sure exactly how necessary this is but seemed like a good idea with my limited knowledge of differences between compilers), or with really large vehicle graphs (only tested with the 32 appliances mentioned so far, but maybe it's worse than before with like 60 or 100 vehicles because we now iterate over `visited_vehs` and `visited_targets`?)
> 
> Other than that I guess I'm a little unsure about changing the `nil_visitor` stuff to once every 5 turns but I guess I could just see how that goes in review.

You're not expected to test in multiple compilers, and like you said the second part needs outside input and this might not get outside input until it's out of draft and ready for review.

PR #60777, opened 2022-09-05
Cardboard folding

harperers:
#### Summary
Content "Adds folded versions of cardboard boxes to item groups and recipes."

#### Purpose of change
Cardboard being fold-able is one of its best features in real life, so it only makes sense to add it in the game. 

#### Describe the solution
I basically did four things
1. Created folded versions of cardboard boxes changing their dimensions and recalculating the item stats accordingly. 
2. Added the folded versions of cardboard boxes to all the loot tables that already had boxes in them, with the exception of small boxes in places, and sometimes entirely replacing the assembled version with the folded version where it made sense (IE the paper recycling group).
3. Added recipes for assembling, which required lowering the difficulty for the cardboard and box nest group to 0.
4. Cardboard boxes can be used to be folded through an item action, and I added a need_empty boolean to use_action's transform type so it would check that the box was empty before folding it.

I was considering having the folded version of the cardboard box be in an item category other than container since it can't contain anything while folded, but I could not think of a better one.
#### Testing

Tried out all the recipes, and checked to make sure the items spawned correctly. And made sure that the fold action worked as intended.

#### Additional context
This is my first time PRing, so I did accidentally make a ton of commits instead of just doing it all at once. Not sure how to fix that now.

### DIFF
--- a/data/json/recipes/nested.json
+++ b/data/json/recipes/nested.json
@@ -3534,13 +3534,7 @@
     "name": "cardboard boxes",
     "description": "Recipes related to constructing cardboard boxes.",
     "skill_used": "fabrication",
-    "nested_category_data": [
-      "box_small_crafted",
-      "box_medium_crafted",
-      "box_large_crafted",
-      "box_large_reinforced",
-      "box_oversize_crafted"
-    ],
+    "nested_category_data": [ "box_small_crafted", "box_medium_crafted", "box_large_crafted", "box_large_reinforced", "box_oversize_crafted" ],
     "difficulty": 1
   },
   {
@@ -3552,12 +3546,7 @@
     "name": "assembled cardboard boxes",
     "description": "Recipes related to assembling cardboard boxes.",
     "skill_used": "fabrication",
-    "nested_category_data": [
-      "box_small_assembly",
-      "box_medium_assembly",
-      "box_large_assembly",
-      "box_oversize_assembly"
-    ]
+    "nested_category_data": [ "box_small_assembly", "box_medium_assembly", "box_large_assembly", "box_oversize_assembly" ]
   },
   {
     "id": "nested_wooden_boxes",


Fris0uman:
Folding cardboard boxes should probably not be a recipe and instead be something you get by activating the box

harperers:
> 

Like i said, I did try this implementation and got it to work, but it would delete all the items inside the box when you did it. If someone could point me towards the right place to try and add functionality to the use_action transform type, I could see what i can do.

src/iuse_actor.cpp
```diff
@@ -235,6 +237,10 @@ cata::optional<int> iuse_transform::use( Character &p, item &it, bool t, const t
         p.add_msg_if_player( m_info, _( "You need to wield the %1$s before activating it." ), it.tname() );
         return cata::nullopt;
     }
+    if( possess && need_empty && !it.empty() ) {
```
irwiss, line 240: Is the possess check necessary here? It looks like "need_empty" will block ALLOWS_REMOTE_USE flag and you don't have this flag on the boxes if I read it correctly

harperers: You're right, it looks like that check was unnecessary

irwiss (approved):
The c++ part looks alright now

data/json/recipes/nested.json
```diff
@@ -3534,8 +3534,17 @@
     "name": "cardboard boxes",
     "description": "Recipes related to constructing cardboard boxes.",
     "skill_used": "fabrication",
-    "nested_category_data": [ "box_small", "box_medium", "box_large", "box_large_reinforced", "box_oversize" ],
-    "difficulty": 1
+    "nested_category_data": [
```
Fris0uman, line 3537: shouldn't there also be a nested recipe for _assembly recipes?

Fris0uman: My bad, everything is here already. But maybe make two different nests?

harperers: that good?

PR #60782, opened 2022-09-05
Salt water should be mixible and not auto-drinkable.

chrispikula:
#### Summary
Bugfixes "Salt water should be mixible and not auto-drinkable."

#### Purpose of change

To be able to mix found, and crafted salt water together.
Also makes it so salt-water isn't auto consumed from a drinking pile, w/ NO_AUTO_CONSUME flag.

#### Describe the solution

Added `"flags": [ "NUTRIENT_OVERRIDE", "NO_AUTO_CONSUME" ],`

#### Testing
Before:
image
After:
image

### DIFF
--- a/data/json/items/chemicals_and_resources.json
+++ b/data/json/items/chemicals_and_resources.json
@@ -27,7 +27,7 @@
     "name": { "str_sp": "sulfur" },
     "symbol": "=",
     "color": "yellow",
-    "flags": [ "EDIBLE_FROZEN", "NO_INGEST" ], 
+    "flags": [ "EDIBLE_FROZEN", "NO_INGEST" ],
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):

Maleclypse:
> Failed again due to `11:37:08.348 ERROR : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\construction.cpp:2114 [check_constructions] construction app_wall_wiring_homemade is constructing appliance but has over 2 components, expected 1`

You have to put in a blank commit or rebase because your test is stuck on that commit even though it's been fixed since then.

chrispikula:
If this fails, a 4th time, I am cursed.

PR #60785, opened 2022-09-05
Remove redundant check

irwiss:
#### Summary
None

#### Purpose of change

Remove redundant components check

#### Describe the solution

Thought I had already removed it, and it didn't trigger since there were no appliances with more than 1 component before #60671.

This is getting checked in done_appliance / done_vehicle, as they expect a single component for the vpart_from_item / vpart_appliance_from_item.

#### Testing

Tests shouldn't fail due to the debugmsg


misterprimus:
Should it fix this?
image

Issue #58887, opened 2022-07-02
Portal Storm Dungeons Feedback and Ideas


Ramza13:
### Is your feature request related to a problem? Please describe.

#58500 has been merged adding the first version of portal storm dungeons.  Right now they are very basic. 

### Solution you would like.

 I am already working on improvements to the dungeon design, rewards available and system which gives out rewards.  This issue is intended to be a place to discuss and gather feedback on this new feature.

Currently planned:

1. More types of dungeon levels, with better randomization
2. Improve the enemies and enhance the endless hunger to be more of a constant threat, possibly give it dialog of a sort.
3. A lot more rewards.
4. A new way to give out rewards.

I'd love to hear your feedback on what already exists, and any ideas you have for improvements.  I'd love to get some really out there or strange stuff in here like a text adventure level or minigames so far out there ideas are fine.

a-chancey:
I really like them so far, it's a neat concept, except it's randomly instakilling my character.
It seems that the floor has a chance to shift into a ledge, and if this happens below the player, they instantly fall to z-0 (the dungeon says it's level 6), crushing and killing them instantly. It also has a chance to do it when exiting via the portal.

NetSysFire:
This was fixed in #59251

a-chancey:
Hah! of course it was merged a few hours after i pulled latest. disregard, i'll give more feedback after i've updated and experienced them more.

I will say, on the first one, I got to dead tired from the monsters inside and had to spend almost and entire 24 hours sleeping. I passed out drinking water in my car after leaving the dungeon

Ramza13:
> Hah! of course it was merged a few hours after i pulled latest. disregard, i'll give more feedback after i've updated and experienced them more.
> 
> I will say, on the first one, I got to dead tired from the monsters inside and had to spend almost and entire 24 hours sleeping. I passed out drinking water in my car after leaving the dungeon

Yea the reliance on exhaustion and shifting masses as a challenge is partially a placeholder.  I'm looking to make the challenges more specific to the dungeon by improving the endless hunger and maybe adding some other new enemies/traps.  Haven't worked out all the details yet.

nullinsky:
Just adding feedback that the forced Strange Location mission is really annoying. I didn't even realize that it started until the portal storm ended and I failed it. It blemished my otherwise perfect mission record.

An option to start the quest or ignore it would be great. Otherwise, this would continue cluttering up the mission tab.

descan:
[...]

PR #60788, opened 2022-09-05
Improve portal dungeon monsters. Add ability to use variables in enchantment strengths

Ramza13:
#### Summary
None

#### Purpose of change
Part of #58887
Add ability to use variables into enchantments.
Improve portal storm dungeon monsters.

#### Describe the solution
Replace current portal storm dungeon monsters with new ones based on the player character.  Add stages so it gets stronger over time.  Add summons for it that only make you slower and carry weights heavier.

Add ability to use variables to enchantments.  This was a pain and required refactoring enchantments into two different classes, enchantments and enchant_cache.  Enchantments were being used both to store a value and to store the sum of a characters/relics values.  Now each of those is a separate class joined by inheritance.

Smaller cleaner changes would have been nice but I could see no cleaner way to do this.

### DIFF
--- a/src/dialogue_helpers.h
+++ b/src/dialogue_helpers.h
@@ -59,7 +59,8 @@ struct talk_effect_fun_t {
         void set_add_var( const JsonObject &jo, const std::string &member, bool is_npc = false );
         void set_remove_var( const JsonObject &jo, const std::string &member, bool is_npc = false );
         void set_adjust_var( const JsonObject &jo, const std::string &member, bool is_npc = false );
-        void set_u_spawn_item( const itype_id &item_name, int count, const std::string &container_name );
+        void set_u_spawn_item( const JsonObject &jo, const std::string &member, int count,
+                               const std::string &container_name );
         void set_u_buy_item( const itype_id &item_name, int cost, int count,
                              const std::string &container_name, const JsonObject &jo );
         void set_u_spend_cash( int amount, const JsonObject &jo );


GuardianDll:
Thats great!

src/condition.cpp
```diff
@@ -1491,14 +1491,13 @@ std::function<int( const T & )> conditional_t<T>::get_get_int( const JsonObject
...
                 if( !var.empty() ) {
-                    return std::stoi( var );
-                } else {
-                    try {
-                        return std::stoi( info.default_val );
-                    } catch( const std::exception & ) {
-                        return 0;
-                    }
+                    // NOLINTNEXTLINE(cert-err34-c)
+                    return std::atoi( var.c_str() );
```
BevapDin, line 1495: According to https://en.cppreference.com/w/cpp/string/byte/atoi this triggers UB if the result would be out of range.

Maybe std::strtol is suitable if you need something non-throwing?

src/magic_enchantment.cpp
```diff
@@ -631,10 +720,25 @@ void enchantment::cast_enchantment_spell( Character &caster, const Creature *tar
...
 
-bool enchantment::operator==( const enchantment &rhs ) const
+bool enchant_cache::operator==( const enchant_cache &rhs ) const
 {
+    if( this->values_add.size() != rhs.values_add.size() ) {
+        return false;
+    }
+    auto iter_add = this->values_add.cbegin();
+    auto iter_add2 = rhs.values_add.cbegin();
+    while( iter_add != this->values_add.cend() && iter_add2 != rhs.values_add.cend() ) {
```
BevapDin, line 730: The loop does not change iter_add nor iter_add2 nor anything at all. The loop is either skipped (because the condition is false) or infinite (because the condition is true and never changes).

Same for below.

And you don't check whether values_multiply and rhs.values_multiply have the same size, but that is probably required.

src/magic_enchantment.cpp
```diff
@@ -462,32 +490,93 @@ void enchantment::force_add( const enchantment &rhs )
...
+}
+
+double enchantment::get_value_multiply( const enchant_vals::mod value, const Character &guy ) const
+{
+    const auto found = values_multiply.find( value );
+    if( found == values_multiply.cend() ) {
+        return 0;
+    }
+    dialogue d( get_talker_for( guy ), nullptr );
+    return static_cast<double>( found->second.evaluate( d ) * 0.01 );
```
BevapDin, line 576: The cast seems redundant. The expression should be of type double anyway.

src/magic_enchantment.cpp
```diff
@@ -462,32 +490,93 @@ void enchantment::force_add( const enchantment &rhs )
...
+    dialogue d( get_talker_for( guy ), nullptr );
+    for( const std::pair<const enchant_vals::mod, int_or_var<dialogue>> &pair_values :
+         rhs.values_add ) {
+        values_add[pair_values.first] += pair_values.second.evaluate( d );
+    }
+    for( const std::pair<const enchant_vals::mod, int_or_var<dialogue>> &pair_values :
+         rhs.values_multiply ) {
+        // values do not multiply against each other, they add.
+        // so +10% and -10% will add to 0%
+        values_multiply[pair_values.first] += 0.01 * static_cast<double>( pair_values.second.evaluate(
```
BevapDin, line 504: The cast seems unnecessary. It will be done by the compiler anyway in order to multiply with the double literal `0.1`.

src/condition.cpp
Ramza13: This part is actually me updating it to the latest version in master, I didn't make this change someone else did.

src/magic_enchantment.cpp
```diff
@@ -631,10 +719,30 @@ void enchantment::cast_enchantment_spell( Character &caster, const Creature *tar
...
+    while( iter_add != this->values_add.cend() && iter_add2 != rhs.values_add.cend() ) {
+        if( iter_add->second != iter_add2->second ) {
+            return false;
+        }
+        iter_add++;
+        iter_add2++;
+    }
+    auto iter_mult = this->values_multiply.cbegin();
+    auto iter_mult2 = rhs.values_multiply.cbegin();
+    while( iter_mult != this->values_multiply.cend() && iter_mult2 != rhs.values_multiply.cend() ) {
```
BevapDin, line 739: Another thing I noticed here: is it correct to compare only the *values* of the map, but not the keys? Currently a map [a => 1, b => 2] is equal to a map [x => 1, y => 2] because the values (1, 2) are equal and in same order, but the maps have completely different keys.

GuardianDll:
Can you make an operations inside the enchantment? i need to multiply my variable to use in enchantment, but i see no examples in this pr. Am i need to make a separate EoC and separace variable for calculation?

Ramza13:
> Can you make an `arithmetic` operations inside the enchantment? i need to multiply my variable to use in enchantment, but i see no examples in this pr. Am i need to make a separate EoC and separate variable for calculation?

So this allows the usage of int_or_var instead of just a float.  int_or_vars can be arithmetic expressions, take a look at https://github.com/CleverRaven/Cataclysm-DDA/pull/57983 for examples.

Zirok-BYOND:
Note that the quadratic fatigue gain used by reflections here can cause very high numbers, ~500 seems to be about dead tired, I've observed 4000 which caused my character to sleep for 5+ days continuosly until they eventually died of dehydration.
(I don't think killing the player in nonobvious ways like this is a good idea, I'd suggest capping fatigue manipulation from portal s torms)

GuardianDll:
Maybe thats a reason people are dead tired after a dungeon?

Zirok-BYOND:
Also portal_dungeon_reflect_strength is not reset when you leave the dungeon

PR #60793, opened 2022-09-05
increase volume of jackhammers

nornagon:
#### Summary
Balance "Increased the volume of jackhammers to 10 L"

#### Purpose of change

It seems a bit surprising that a jackhammer could fit in 5 L!

#### Describe the solution

10 L is from a rough eyeball guess comparing images of people holding jackhammers to how many 1/2 gal jugs of milk I think would take up the same volume. Highly scientific.


snipercup:
Could you please change the crafting/deconstruction recipes as well? The sum of the components should end up with the same volume.

anothersimulacrum:
Same mass (or the result has less) is more important than same volume.

PatrikLundell:
As indicated, volume can change in both directions:
- Building a box out of flat sheets results in an object with a much larger volume than the parts, as the air isn't considered an ingredient, and deconstructing it again releases the air.
- If you mount things inside a container you can end up with a smaller volume item than the volume of the parts (as air is replaced), and deconstruction can thus result in a larger volume as the inside components are placed outside of the container (again, replacing the components with air).

PR #60795, opened 2022-09-05
[XEDRA Evolved] Investigation -> deduction

GuardianDll:
#### Summary
None

#### Purpose of change

All skills in the game starts from small letter, except the investigation

#### Describe the solution

Replace upper case wilt lower case
Also replace the whole name 🤷 

#### Additional context

One small step for man..

### DIFF
--- a/data/mods/Xedra_Evolved/skills.json
+++ b/data/mods/Xedra_Evolved/skills.json
@@ -2,7 +2,7 @@
   {
     "type": "skill",
     "ident": "deduction",
-    "name": { "str": "investigation" },
+    "name": { "str": "deduction" },
     "description": "Your ability to see the pattern behind a series of data points.  Allows you to ask questions that you otherwise couldn't ask and find things you might have missed otherwise.",
     "companion_industry_rank_factor": 1,
     "tags": [ "combat_skill" ],


TheShadowFerret:
Why is this skill called investigation when its id is deduction 🤔

GuardianDll:
> Why is this skill called investigation when its id is deduction 🤔

@Maleclypse?

Maleclypse:
> > Why is this skill called investigation when its id is deduction 🤔
> 
> @Maleclypse?

I couldn’t decide which I liked better and it probably makes more sense as deduction. Especially since it’s used for spells in this mod currently. I’d originally planned to have two skills but the mod isn’t big enough for that yet.

data/mods/Xedra_Evolved/skills.json
```diff
@@ -2,7 +2,7 @@
   {
     "type": "skill",
     "ident": "deduction",
-    "name": { "str": "Investigation" },
+    "name": { "str": "investigation" },
```
Maleclypse, line 5:
```suggestion
    "name": { "str": "deduction" },
```
Let's just go ahead and make the change based on the discussion below.

Maleclypse (approved):

PR #60799, opened 2022-09-06
New Nether monster that uses the flaming eye's old attack.

Daved27hundred:
#### Summary
Content "Reintroduces the flaming eye's old destructor ray attack in a new Nether monster."

#### Purpose of change

The flaming eye's old attack, a massive laser that destroyed terrain, was removed from the flaming eye back in 2015 for presumably being too unfair. This PR tries to bring it back (along with the sheer terror that comes with it) By giving it to a different monster.
The Jeweled Tortoise is much slower than the flaming eye, and only has half the sight radius of it in the day along with none of the night vision. This was done so that it's much less of an annoyance to encounter, and that the player can actually attempt to run away from it.

#### Describe the solution

I discovered that the mod No Hope included the flaming eye's destruction ray, so I repurposed the code from there.

giving the monster a different attack, or not adding it entirely. The monster could also be made to spawn on quest locations instead of in the wild, if it's too powerful.

#### Testing

apart from loading a world with the No Hope mod and debug spawning a flaming eye to see how the ray worked, I have not tested it yet.


data/json/monsters/nether.json
```diff
@@ -394,6 +394,48 @@
...
+    "armor_acid": 6,
+    "armor_fire": 6,
+    "armor_elec": 8,
+    "bleed_rate": 10,
+    "vision_day": 25,
+    "vision_night": 8,
+    "harvest": "zombie_meatslug",
+    "special_attacks": [
+      [ "destruction_ray_warmup", 10 ],
+      {
```
Venera3, line 430: This won't really gate the ray behind the warmup beyond coming first in line on a shorter cooldown - you can add a self-effect and hide the ray behind a `required_effects_any` condition for that, and change the warmup to a melee type to utilize `attack_chance`.

data/json/monsters/nether.json
```diff
@@ -394,6 +394,48 @@
...
+    "armor_bash": 17,
+    "armor_cut": 18,
+    "armor_bullet": 20,
+    "armor_acid": 6,
+    "armor_fire": 6,
+    "armor_elec": 8,
+    "bleed_rate": 10,
+    "vision_day": 25,
+    "vision_night": 8,
+    "harvest": "zombie_meatslug",
```
Venera3, line 427: I wouldn't use tainted meat here. Maybe disappearing in a brilliant explosion or something like that if you don't want to get bogged down in the "what exactly are these made of" question.

Daved27hundred: yeah I could do that. The design of the monster is “inspired” by the angels in Evangelion and they have some explosive deaths there. I’ll only be available some time after 2PM. though so It will be a while.

data/json/monsters/nether.json
Daved27hundred: Will do this after I get home tomorrow. well technically today because it’s 1 AM but you get what I mean.

data/json/monsters/nether.json
```diff
@@ -394,6 +394,48 @@
...
+    "special_attacks": [
+      [ "destruction_ray_warmup", 10 ],
+      {
+        "type": "spell",
+        "spell_data": { "id": "destruction_ray" },
+        "cooldown": 30,
+        "monster_message": "A piercing beam of light bursts forth!"
+      }
+    ],
+    "flags": [ "SEES", "NOHEAD", "NO_BREATHE", "BASHES", "WARM" ]
```
Night-Pryanik, line 437: Please add a RANGED_ATTACKER flag.

Amoebka:
What's the actual design purpose behind having free-roaming instakill monsters? In what way are they supposed to make the game more immersive or fun?

Daved27hundred:
> What's the actual design purpose behind having free-roaming instakill monsters? In what way are they supposed to make the game more immersive or fun?

This monster uses the deathray from No Hope, which after I tested it, doesn’t seem to instakill. the 999 damage appears to effect terrain only - It did 18-20 damage per shot to my player and other monsters.

It’s not supposed to be “free-roaming” at all, I made the creature slower and cut it’s sight radius to half of the flaming eye. It’s meant to stay in place and pose as an obstacle, not relentlessly chase the player until it fries them.

If you needed to kill it you would need to approach it smartly. Outrange it with a ranged weapon, attack it at night where it’s sight radius is far lower, or cripple it with a flashbang since it doesn’t have ears.

data/json/monsters/nether.json
Daved27hundred: well I took a crack at it, but I'm not sure it will work or not. It's better I attempt it and not get it right than not do it at all, so tell me if It needs any improvements.

taatu:
At least it should not be in spawn groups whose normally most dangerous entities are the flaming eye and kreck. It fits more alongside stuff like the flying polyp and yugg

Daved27hundred:
I'm going to leave this PR sitting here for a while, I feel like adding batch time savings to some recipes.

PR #60806, opened 2022-09-06
[FEATURE] add cow armors

leoCottret:
#### Summary
Content "Adds cow armors"

#### Purpose of change
Cows can be killed in a few turns by a weak monster because they don't have armors. Especially when transporting cows in a truck, monsters can phase diagonally through the car body and leave little times for the player to react.

#### Describe the solution
The solution I choose is to add cow armors. I used the horse armors as a template, compared the difference in weight between a horse and a cow corpse (~20%), and modified the components required, weight, price, material thickness and crafting time accordingly.

Adding the ability for cows to wear horse armor could have been a solution, although they don't really have the same dimensions.

#### Testing
I spawned all armors and was able to equip the cow with them.
I crafted one of the armors too, they all appear in the crafting menu.
I could still ride/milk/butcher/lead/tie the cow.
I got a cow killed by 1 Zombie with and without armor and saw that the armor was working as intended.

The one change that could add regression is the modification of the bodytype from dog to cow. I don't know what it is used for. I couldn't equip dog armor on the cow but changing the bodytype to cow allowed me to create working armors for this bodytype.
Also this is my first PR so I guess you should be extra careful.

#### Additional context

Equipping an armor
image

2 tied cows with a zombie brute, cow without armor on the right is dead while the cow with steel armor on the left has 80% health
image

### DIFF
--- a/data/json/recipes/recipes.json
+++ b/data/json/recipes/recipes.json
@@ -107,12 +107,7 @@
   {
     "type": "recipe_category",
     "id": "CC_ANIMALS",
-    "recipe_subcategories": [
-      "CSC_ALL",
-      "CSC_ANIMALS_CANINE ARMOR",
-      "CSC_ANIMALS_EQUINE ARMOR",
-      "CSC_ANIMALS_EQUINE STORAGE"
-    ]
+    "recipe_subcategories": [ "CSC_ALL", "CSC_ANIMALS_CANINE ARMOR", "CSC_ANIMALS_EQUINE ARMOR", "CSC_ANIMALS_EQUINE STORAGE" ]
   },
   {
     "type": "recipe_category",


LyleSY (commented):
I'm not sure it makes sense to define a new bodytype "cow" redundant with dog. If it is it should be documented. Instead I'd suggest keeping it pretty similar but using the description to make it clear that it's for cows and not dogs, just make it big dog armor on the backend.

Zireael07:
Can't we just have a 'quadruped' body type, it'd fit cows/dogs/horses/Exodii quadrupeds/whatevers else

leoCottret:
I can indeed set the bodytype as dog, I didn't realised that the important fields are min_pet_vol and max_pet_vol. Should I include the calf in the dog armors, cow armors, or neither?

Sevaaas1:
would it be possible instead, to make a vehicle part that prevents the cow from moving, and prevents the cow from taking damage until the part breaks?

leoCottret:
Although thinking about it, it wouldn't make much sense to have an armor that would fit a cow/horse/elephant... so I think we can stay like that, a cow being of "dog" bodytype -> "a four legged animal with a short neck elevating the head above the line of the body" (that fits I think). If one day we create a **huge**, tameable dog, it would fit in a cow's armor, which seems possible I guess. About the craft menu clutter you just have to write "cow", "dog" etc. to have the correct results anyway.

leoCottret:
> would it be possible instead, to make a vehicle part that prevents the cow from moving, and prevents the cow from taking damage until the part breaks?

That wouldn't fix the issue of cow needing an armor in general, I personnaly only use mobile bases but you could use a cow to transport your bags, or just have a cow as a pet (some people do that IRL) and want it protected.

LyleSY:
I think this is mostly there now. I'd recommend using the existing CSC_ANIMALS_CANINE ARMOR instead of defining a new subcategory. I like the idea of creating a simpler "quadraped" category that is more clearly inclusive but I'm not sure this PR needs to take that on

leoCottret:
> I think this is mostly there now. I'd recommend using the existing CSC_ANIMALS_CANINE ARMOR instead of defining a new subcategory. I like the idea of creating a simpler "quadraped" category that is more clearly inclusive but I'm not sure this PR needs to take that on

Yes I find it a bit out of scope, also it seems to me that the only check for whether or not a "monster" can wear an armor is the min_pet_vol/max_pet_vol and the bodytype of this armor. If several tammable monsters used the quadruped body type, we would have to be very precise with the min_pet_vol/max_pet_vol values of the armor, otherwise 2 different monsters with roughly the same size could wear each other armor.

For example, currently if the elephant, horse and cow were all marked as quadruped, and their armor too, they could wear each other armors.

LyleSY (approved):

PR #60807, opened 2022-09-06
Spend cordless drills charges when installing vehicle tanks

### LINKED ISSUE #47122: Cordless drill does not use charges when installing 60L tank in vehicle
# Describe the bug

Installing a tank in a vehicle requires a drill, however if you use an electric drill it does not use any charges.

# Steps To Reproduce

1. Load save ( [Narnia.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/5897804/Narnia.zip) ) (or spawn 60L tank, vehicle, and other required tools)
2. Check charge level in cordless drill (it's 1200 for save)
2. 'e'xamine vehicle
3. Install 60L tank
4. Confirm charge level is unchanged.

# Expected behavior

If you're required to have drill you can presume that it will be used, and if you use an electric drill it should lose some charge.

# Screenshots

# Versions and configuration

- OS: Windows x64 8.1
- Game Version: 0.E-9155-g1319073-dirty [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
]

# Additional context

Note that if you unload the battery from the cordless drill you can still install the 60L tank with it.

Night-Pryanik:
#### Summary
Bugfixes "Spend cordless drills charges when installing vehicle tanks"

#### Purpose of change
* Closes #47122.

#### Describe the solution
- Added hand drill, which has `DRILL` quality of 2, to the `drilling_standard` requirement
- Replaced `WRENCH` and `DRILL` qualities with `vehicle_wrench_x` and `drilling_standard` requirements, respectively, when installing vehicle tanks
- Also, while I'm here, I increased drill charges required to install boat hulls, as they were minuscule.

None.

#### Testing
Got 2l tank, got fully charged cordless drill. Installed tank on the vehicle, checked that drill charges are consumed.

#### Additional context
None.

### DIFF
--- a/data/json/vehicleparts/vehicle_parts.json
+++ b/data/json/vehicleparts/vehicle_parts.json
@@ -1101,7 +1101,7 @@
     "item": "plastic_boat_hull",
     "location": "under",
     "requirements": {
-      "install": { "skills": [ [ "mechanics", 3 ] ], "time": "60 m", "using": [ [ "vehicle_screw", 1 ], [ "drilling_standard", 1 ] ] },
+      "install": { "skills": [ [ "mechanics", 3 ] ], "time": "60 m", "using": [ [ "vehicle_screw", 1 ], [ "drilling_standard", 40 ] ] },
       "removal": { "skills": [ [ "mechanics", 2 ] ], "time": "30 m", "using": [ [ "vehicle_screw", 1 ] ] },
       "repair": { "skills": [ [ "mechanics", 4 ] ], "time": "60 m", "using": [ [ "soldering_standard", 15 ] ] }
     },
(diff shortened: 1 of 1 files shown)


Fris0uman:
Doesn't that mean that an hand drill is always better than a powered drill for any recipe using drilling_standard?

Night-Pryanik:
Yes, as there is no way to set different installation times based on tool used.

Fris0uman:
the current way would be to set two recipe (using id_suffix) one fast using powered tool and one slow with handtool and maybe having a higher level of exercise

Night-Pryanik:
It'll work for recipes, sure, but won't work for installing vehicle parts.

Night-Pryanik:
Oh, I just found out that no recipes currently uses `drilling_standard` requirement at all.

PR #60813, opened 2022-09-06
Stock Pot Great Helm

anoobindisguise:
#### Summary
Content "Stock Pot helmet"

#### Purpose of change
I wanted to add another stupid early game armor option made out of cookwear, and the stock pot seemed like a reasonable choice to make a helmet. The existing pot helmet is more of a Crazy Dave sort of thing where it just covers your cranium, this one should cover quite a bit more.

#### Describe the solution
Based the helmet off of 20-gauge stainless steel (0.95mm walls) with a base coated in 4mm of aluminum. It has a blanket stuffed inside to cushion it and make it slightly less awful to wear. Takes up both Normal and Outer layers, occupies outer layer for mouth and eyes. Needs metal sawing to cut a T-shape into the front of it to see out of. Stupidly high encumbrance. Has strings tied to the handles, then tied to your shoulders, in order to fasten it down and prevent it from coming off as easily. Beginner level craft (0 fab). High warmth too due to the blanket. Not sturdy.

Lower coverage of aluminum, it's just on the roof of the helmet. 
Lower coverage of the cotton/lower thickness as it's a wadded-up blanket.
Require 1 Fab since it involves sawing metal.
Make it sturdy, although a stock pot is kinda flimsy. 
Different grade of steel. Medium steel felt like the most appropriate representation of stainless cookware steel.
Don't add it at all as I cannot prove whether 20-gauge steel is structurally sound against zombies.


chrispikula:
You've got a warmth of 40, and an insulated near-zero airflow box around the head.   I'm partially wondering if this is low, given you lose 10% of your body heat through your head, and an additional 20-35% through breathing.  This is going to get glasses foggy uncomfortable in _seconds_, irl.

anoobindisguise:
> You've got a warmth of 40, and an insulated near-zero airflow box around the head. I'm partially wondering if this is low, given you lose 10% of your body heat through your head, and an additional 20-35% through breathing. This is going to get glasses foggy uncomfortable in _seconds_, irl.

Good points, jumped up the encumbrance for eyes/mouth, and increased warmth to 65. Although, such a stock pot is designed to trap heat inside it already, so it could even be higher (75-80)

Aerin-of-the-Toast:
So I don't think this should have 100% coverage. I did some intensive research on the size of stock pots compared to human heads (i.e. googling) and only the largest stock pots would cover the whole head down past the chin. In this case it still would leave substantial gaps underneath the helmet and the neck would be uncovered, compared to something like the actual great helm that comes with a bevor for the neck. 

This is all before taking into account the helmet will be raised a few inches off the head due to the blanket, so it's probable even the larger 12-inch-tall stock pots would leave part of the chin exposed.

PatrikLundell:
Pots are not designed to trap heat. They're made out of metal so they can conduct heat from the heat source faster than it escapes through the walls and the top, with little regard for efficiency (good enough is good enough). Kettles (the water boiler kind, not the stove top ones) are made out of materials that actually provide some level of insulation.
A blanket provides insulation, though.

chrispikula:
> Pots are not designed to trap heat.

The lack of airflow due to the pot's walls will intensify heat gain when combined with the blanket.  IRL insulation is almost always about making more air pockets and stopping air movement, as air is the great insulator, not whatever makes those pockets.

data/json/items/armor/helmets.json
```diff
@@ -20,6 +20,57 @@
...
+    "price": 4500,
+    "price_postapoc": 50,
+    "bashing": 10,
+    "material": [ "mc_steel", "aluminum" ],
+    "symbol": "[",
+    "looks_like": "helmet_plate",
+    "color": "light_gray",
+    "warmth": 70,
+    "material_thickness": 1,
+    "qualities": [ [ "COOK", 2 ], [ "BOIL", 1 ] ],
```
Fris0uman, line 41: There are holes in it now, I don't think it's going to be as good for cooking anymore

data/json/recipes/armor/head.json
```diff
@@ -2750,5 +2750,21 @@
...
+    "category": "CC_ARMOR",
+    "subcategory": "CSC_ARMOR_HEAD",
+    "skill_used": "fabrication",
+    "time": "15 m",
+    "autolearn": true,
+    "qualities": [ { "id": "SAW_M", "level": 1 } ],
+    "components": [
+      [ [ "stock_pot", 1 ] ],
+      [ [ "string_36", 2 ] ],
+      [ [ "blanket", 1 ], [ "down_blanket", 1 ], [ "quilt", 1 ], [ "quilt_patchwork", 1 ] ]
```
Fris0uman, line 2767: Aren't blankets a bit too big for this?

data/json/items/armor/helmets.json
anoobindisguise: yeah, it went to boiling level 1 instead since saucepans have boiling 1 and the cuts don't go all the way up the pot. So it's no longer super suitable for boiling (stock pot has 2 boil) but it can still do it a little bit

data/json/recipes/armor/head.json
anoobindisguise: I don't know personally I'd probably need to try and shove a quilt into a stock pot and see how it goes. I was thinking of just using a pillow but that felt like it could be too small.

Night-Pryanik:
@anoobindisguise could you please resolve errors?

PR #60761, opened 2022-09-04
In Tab Nesting

bombasticSlacks:
#### Summary
None

#### Purpose of change
fixes #60588
fixes #60603
fixes #60359

#### Describe the solution
Removed code for the separate nested tabs, they were getting unweildly and hard coded. 

- Instead track a list of "expanded recipes" similar to favorites or hidden.
- Track a vector of every items "indent" for nested items.
- When redrawing check if any nests are in that list, if so locally sort there contents and add it to the list of recipes. As well populate their "availability" and their "indent".
- Inherently (because they were added after the currently checked item) do this for those nested items too.
- Hitting confirm on an item nest either adds it to the expand list or removes it.

#### Testing
https://user-images.githubusercontent.com/4514073/188329926-437c6e58-b18a-485c-825d-10454e52eab6.mp4

#### Additional context
Sorry for everyone who originally suggested this and I claimed it would be "too much work". I didn't expect the feature to get used on the JSON side as heavily as it did so may as well do it right.

PR #60819, opened 2022-09-07
Fix crash on opening batch crafting screen

### LINKED ISSUE #60804: Batch Crafting, Error Message and Crash
### Describe the bug

Whenever I try to batch craft any item, the game crashes.  Only occurs when crafting in base area.

### Steps to reproduce

Batch Craft any Item, receive error message, crash

Error Message:
The program has crashed.
See the log file for a stack trace.
CRASH LOG FILE: ./config/crash.log
VERSION: e2e7914
TYPE: St12length_error
MESSAGE: basic_string::_M_create

(For background, my base is kind of set up as a pack rat, Lots of Loot Zones, and 4 NPC Followers.  Crashes, with all loot zones disabled/enabled, with followers present/gone. There are a lot of items in the base, which may or may not be reason for the crash.)

crash.log & debug.log attached

### Expected behavior

Go into crafting menu, press b, batch craft items as usual.

### Screenshots

_No response_

Os: Windows 10
Game Version: Experimental: 2022-09-06-0724
Graphics: Tiles
Ingame Language: English
Mods Loaded: DDA, Disable NPC Needs, No Fungal Growth, Bionic Professions, Blaze Industries, Magiclysm, Alternative Map Key, Extra Mutated Scenarios, Mythical Martial Arts, Translate Complex Dialogue, Sees-player icon, sees-player (run,crouch bleed, etc..)

### Additional context

[config.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/9495244/config.zip)

ZeroInternalReflection:
#### Summary
Bugfixes "Fix crash on opening batch crafting screen"

#### Purpose of change
Indentation for nested crafting recipes (added by #60761) was not being updated properly when the user went to the batch crafting screen.  For most tabs, this resulted in strange indentation on the batch crafting screen:

Improper_indentation_from_Other_crafting_tab

However, for the "favorite", "recent", and "search" tabs/subtabs, in resulted in an immediate and hard crash.

Fixes #60804

There were also several other actions in the crafting GUI for nested recipes that result in unexpected behaviour:
- Hitting `b` with a recipe group selected:
Attempting_to_batch_craft_a_recipe_group

- Hitting `L` (related recipes) with a recipe group selected fails silently
- Hitting `e` (recipe details) with a recipe group selected:
Recipe_group_details

#### Describe the solution
 - Create a new zero-ed vector of indentations when going to the batch crafting screen
 - Add a debug message to check if future changes to indentation hit the same issue
 - Check that the selected recipe isn't a nested group prior to various actions, failing with a popup if it is

I believe the actual crash was a combination of a) a mismatch in vector sizes and b) potentially attempting to create a string of size <0 (due to the mismatch in vector sizes).
 - Would it be a good practice to add a specific check for 'indent' size vs. 'current' size just prior to using indent?
```
if( current.size() != indent.size ) { debugmsg... }
```
 - Or to have a clamp on the indentation:
```
std::string tmp_name = std::string( std::max( 0, indent[i] ), ' ' )...
```

Both of these would have prevented the crash (though not the odd indentation)

#### Testing

Pulling up the batch crafting screen on the 'favorite' subtab:

https://user-images.githubusercontent.com/89038572/188783653-7d57f0dd-50a0-47bb-85c4-e373fab1faa8.mp4

The 'recent' and the (deprecated?) 'nested' subtab:

https://user-images.githubusercontent.com/89038572/188783773-6d8e2a13-d176-42bd-ae42-620d4a3a75bf.mp4

The search tab:

https://user-images.githubusercontent.com/89038572/188783894-772bed48-7b97-4df9-b7ba-87fe8418ba68.mp4

Attempting various other crafting menu actions on the nested recipes:
 - Favouriting/unfavouriting seems to work roughly as expected, though the indentation is thrown off slightly by the "* " prefix
 - Related recipes/recipe details should probably list the recipes, but I think a failure popup is reasonable for now
 - Hiding/unhiding nested recipes seems to work, though it doesn't seem to prevent searching for recipes within a hidden nest
 - Mark as read/unread should probably work recursively, but I'd call that a low priority
 - Being able to open a nested recipe group from within the search screen when the recipes within that group are already listed isn't ideal, but all the required information _does_ show up when searching

### DIFF
--- a/src/crafting_gui.cpp
+++ b/src/crafting_gui.cpp
@@ -979,6 +979,25 @@ static void nested_toggle( recipe_id rec, bool &recalc, bool &keepline )
     keepline = true;
 }
 
+static bool selection_ok( const std::vector<const recipe *> &list, const int current_line,
+                          const bool nested_acceptable )
+{
+    std::string error_message;
+    if( list.empty() ) {
+        error_message = _( "Nothing selected!" );
+    } else if( list[current_line]->is_nested() && !nested_acceptable ) {
+        error_message = _( "Select a recipe within this group" );
+    } else {
+        return true;
+    }
+
+    query_popup()
+    .message( "%s", error_message )
+    .option( "QUIT" )
+    .query();
+    return false;
+}
+
 const recipe *select_crafting_recipe( int &batch_size_out, const recipe_id goto_recipe )
 {
     recipe_result_info_cache result_info;
@@ -1513,8 +1532,6 @@ const recipe *select_crafting_recipe( int &batch_size_out, const recipe_id goto_
         const bool mouse_in_list = coord.has_value() && mouseover_area_list.contains( coord.value() );
         const bool mouse_in_recipe = coord.has_value() && mouseover_area_recipe.contains( coord.value() );
 
-        const std::string nested_selected = _( "Select a recipe within this group" );
-
         // Check mouse selection of recipes separately so that selecting an already-selected recipe
         // can go straight to "CONFIRM"
         if( action == "SELECT" ) {
@@ -1662,14 +1679,7 @@ const recipe *select_crafting_recipe( int &batch_size_out, const recipe_id goto_
                 done = true;
                 uistate.read_recipes.insert( chosen->ident() );
             }
-        } else if( action == "HELP_RECIPE" ) {
-            if( current.empty() || current[line]->is_nested() ) {
-                query_popup()
-                .message( "%s", current.empty() ? _( "Nothing selected!" ) : nested_selected )
-                .option( "QUIT" )
-                .query();
-                continue;
-            }
+        } else if( action == "HELP_RECIPE" && selection_ok( current, line, false ) ) {
             uistate.read_recipes.insert( current[line]->ident() );
             recalc_unread = highlight_unread_recipes;
             ui.invalidate_ui();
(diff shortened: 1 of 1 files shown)


PatrikLundell:
I consider it poor practice to hide errors by clamping the results of incorrect calculations without generating an error message as an alert that something is wrong. Thus, either assume things are done correctly (i.e. no check), or both an error message and error handling.
Note that this is my view, and has no relation to any potential official stance of the core team.

ZeroInternalReflection:
> I consider it poor practice to hide errors by clamping the results of incorrect calculations without generating an error message as an alert that something is wrong. 

The stance of "If the programmer has done something wrong, warn them rather than quietly ignoring it" absolutely makes sense to me.

Given how hard this error crashed the game (in my case, complete terminal/process lockup that didn't even generate a crash log.  I also saw a report of someone needing to reboot their computer), I've added a check and debug message if indentation is expected to cause an error.  This should make it a little less likely to have this problem reoccur in further tweaks to the indentation/nesting system.

Clang-tidy won't like this update either.  I _can_ do some work simplifying select_crafting_recipe, but I think that's probably best as a separate PR.

Also, thanks to Fris0uman for once again fixing a typo in my summary line.

OnTheDouble:
Can't wait for this fix to go through. CDDA just doesn't feel the same without the ability to craft from the Favorites menu.

robob27:
Fixes my crash, thank you!

Daved27hundred:
I'd like to suggest, if anyone here is tired of waiting for the fix, try going back to an earlier build.

OnTheDouble:
Right but what's exactly the hold up right now?

bombasticSlacks:
It doesn't pass clang, the additional logic needs to be moved to a separate function.

If this is merged every subsequent PR will fail clang as well. Until the simplification happens we can't merge it.

dseguin:
I think that `select_crafting_recipe` would need to be broken out into a class of its own, something like this: https://github.com/dseguin/Cataclysm-DDA/commit/6dba3cf8d10461d669169eafe41a25fd9e9c8fee

ZeroInternalReflection:
> I think that most of select_crafting_recipe would need to be broken out into a class of its own

While I'm very much in favour of a major rework of select_crafting_recipe as dseguin is proposing, I don't think a bugfix like this is the spot for it.   So, in an effort to get this particular bug resolved, I've made a tangentially-related change that should pull enough out of select_crafting_recipe to avoid a clang-tidy error.

If there are any other issues preventing this from getting merged, please let me know.

EternalKitten:
Count me excited.

PR #60828, opened 2022-09-07
Nomex zombie

EtherealArcturus:
#### Summary
Content "Add an evolution line for the firefighter zombie"

#### Purpose of change
Since I started playing Cataclysm, I always found it strange that during the later stages of the game, certain zombies didn't evolve. This PR is a step towards rectifying that problem.

#### Describe the solution
This PR adds an evolution line for the firefighter zombie, one of the basic zombie varieties.

Not adding this, making it more generic or taking a different approach for this evolution, maybe also having an alternative evolution with a perpetually burning zombie (thanks to blob magic)?

#### Testing
I compiled and loaded the game without a crash. I also made sure that the firefighter zombie actually evolved and I did a bit of combat testing, making sure the abilities and weakpoints worked as intended. It would be great if anyone who knows more about the balancing would take a look at it, making sure everything is fine.

#### Additional context
Instead of making a generic evolution I wanted to make something with a bit more flavor. At first I wanted to make the nomex zombies have no sight at all, but it didn't feel right, so I just heavily impaired it. I also tried to make it get stunned every time it performed a head bash, but couldn't figure out how.


data/json/monsters/zed_misc.json
```diff
@@ -527,6 +527,132 @@
       "FILTHY"
     ]
   },
+  {
+    "id": "mon_zombie_nomex_1",
+    "type": "MONSTER",
+    "name": { "str": "nomex zombie" },
+    "description": "This was at some point a zombie wearing some kind of fire resistant protective gear.  That same outfit is now shredded and has become part of its outer tissue, making it more resiliant against physical attacks and heat induced damage.  It also seems to be rather muscle packed. At least its overgrown and hardened faceplate still holds up, preventing it from biting you and also rendering its eyes impaired.  However, it looks like it can still hear and smell you just fine.",
```
Night-Pryanik, line 534:
```suggestion
    "description": "This was at some point a zombie wearing some kind of fire-resistant protective gear.  That same outfit is now shredded and has become part of its outer tissue, making it more resilient against physical attacks and heat-induced damage.  It also seems to be rather muscle packed.  At least its overgrown and hardened faceplate still holds up, preventing it from biting you and also rendering its eyes impaired.  However, it looks like it can still hear and smell you just fine.",
```

data/json/monsters/zed_misc.json
```diff
@@ -527,6 +527,132 @@
...
+      "PUSH_MON",
+      "NO_FUNG_DMG",
+      "FILTHY"
+    ]
+  },
+  {
+    "id": "mon_zombie_nomex_2",
+    "type": "MONSTER",
+    "name": { "str": "nomex hulk" },
+    "description": "This was at some point a zombie wearing some kind of fire resistant protective gear.  It has grown to an enormous size with bulging muscles and features a hide in the likeness of its former uniform.  The faceplate that once shielded its owners face against fire has now hardened to more closely resemble a slightly curved, bone crushing hammerhead.  As a result its ability to see is heavily impaired, but its hearing and sense of smell appear to be working just fine.",
```
Night-Pryanik, line 604:
```suggestion
    "description": "This was at some point a zombie wearing some kind of fire-resistant protective gear.  It has grown to an enormous size with bulging muscles and features a hide in the likeness of its former uniform.  The faceplate that once shielded its owner's face against fire has now hardened to more closely resemble a slightly curved, bone-crushing hammerhead.  As a result its ability to see is heavily impaired, but its hearing and sense of smell appear to be working just fine.",
```

Night-Pryanik:
Ever-burning zombie sounds like an interesting idea, but not as a part of firefighter's evolution line.

EtherealArcturus:
What would you suggest? Maybe a weird offshoot of the gasoline zombie line?

Night-Pryanik:
Well, personally I'd prefer a completely new type of zombie. Let's say some unfortunate that died in a fire and turned to zombie while still burning, and blob thought that maintaining an ever-burning fire on this zombie would be a good idea. Something like that.

EtherealArcturus:
Maybe we can make this work with the firefighter zombie line after all. We can utilize the "burn_into" function for this. Since the firefighter zombie and its evolutions are supposed to be more resistant to fire than a lot of other zombie, there may be like a 20% chance for it to turn into a "burning zombie" rather than a "scorched zombie" because it didn't just turn into a extra crispy chicken nugget.

Amoebka:
Is it really OK to use weakpoints for parts of armor that are stronger? From what I understand, high skills and proficiencies increase chances of striking weakpoints, with an assumption that it's always beneficial to do so. Breaking this rule will make high skill players worse at fighting these than low skill players.

EtherealArcturus:
Actually, the "wps_humanoid_full_helmet" weakpoint I used as reference did this before I did.

Amoebka:
Copying a mistake doesn't make it any less of a mistake. Both should be fixed.

Amoebka:
This is faulty logic. It's not necessarily your job to fix the initial mistake, but you shouldn't make it worse at least. Do you WANT skilled players to hit strong parts of armor more often than unskilled ones? If no, you shouldn't copy bad data that makes it so.
What's the point of intentionally introducing faulty behaviour that you know will have to be fixed later?

Venera3:
Weakpoints that have higher armor than the rest and have low (or zero) difficulty mean that at higher skill/prof levels your chances of hitting them go down since "proper" weakpoints tend to have higher difficulties.

Amoebka:
Yes, but don't your overall chances of hitting any weakpoints go up? This means that at some intermediate skill level the expected DPS might have a minimum, because you are most likely to hit the "trap" weakpoints. It would depend on the exact set of weakpoints and their difficulties.
Admittedly way out of scope here, but the overall logic seems error-prone. There are no tests to ensure these minimums don't happen either.

GuardianDll:
I think if there was an ability to hit a weakpoint without weakpoint knowledge (related to its difficulty) it would make sense. But i dont know how the current weakpoints works

About PR - that is a nice addition, good job. I guess there is no way to make a "stun after melee" thingy using special_attack, but its doable via spell system. If you're interested, i can make one

EtherealArcturus:
Yes, that would be great, since it would mean more mechanical depth for those zombies. I'm sure that there are also other monsters that would benefit from an effect like this as well.

data/json/monsters/zed_misc.json
```diff
@@ -527,6 +527,132 @@
       "FILTHY"
     ]
   },
+  {
+    "id": "mon_zombie_nomex_1",
+    "type": "MONSTER",
+    "name": { "str": "nomex zombie" },
+    "description": "This was at some point a zombie wearing some kind of fire-resistant protective gear.  That same outfit is now shredded and has become part of its outer tissue, making it more resilient against physical attacks and heat-induced damage.  It also seems to be rather muscle packed.  At least its overgrown and hardened faceplate still holds up, preventing it from biting you and also rendering its eyes impaired.  However, it looks like it can still hear and smell you just fine.",
```
TheShadowFerret, line 534:
```suggestion
    "description": "This was at some point a zombie wearing some kind of fire-resistant protective gear.  That same outfit is now shredded and has become part of its outer tissue, making it more resilient against physical attacks and heat-induced damage.  It also seems to be rather muscle packed.  At least its overgrown and hardened faceplate still holds up, preventing it from biting you and also rendering its eyes impaired.  However, it looks like it can still hear and smell you just fine.",
```

How would you know it can hear & smell you if the faceplate is overgrown 🤔

EtherealArcturus: When I designed that, I was inspired by the headless zombie evolution line and imagined it a little something like this: When the faceplate of the zombie gets hardened and overgrown, its ability to see gets impaired. As a result, the blob does as usual, making the best of what it got. Since the helmet with a faceplate is not a perfect isolation barrier, with little gaps and cracks opening up the inside to the outside world, it would just greatly improve the senses of hearing and smell of the zombie to overcome the barrier and/or rout some of those sensors directly to those cracks and gaps to make them more effective.
Sure, it could also make some new outside eyes grow, but since those are fairly complex and expensive organs, it just can't be bothered, y'know?

From a balance perspective, it makes up for the zombies shortcomings in other areas.

If it is too much handwavium, I could change it to be different.

PatrikLundell:
Regarding the ever burning zombie tangent, it would be absolutely horrible with the current state of the game. Fire is severely overpowered and lethal, and the damned thing would burn down the neighborhood in no time at all (well, the burning will take forever, but the setting fire to everything is very quick). In addition to that, a player would probably die of the heat damage sustained just from trying to close to melee range.

EtherealArcturus:
I'm not really sure how to solve those failing checks.

Night-Pryanik:
> Regarding the ever burning zombie tangent, it would be absolutely horrible with the current state of the game. Fire is severely overpowered and lethal, and the damned thing would burn down the neighborhood in no time at all (well, the burning will take forever, but the setting fire to everything is very quick).

No one said this zombie should set everything on fire where it steps. It might burn by itself, but that's it, no generating of fire fields.

> In addition to that, a player would probably die of the heat damage sustained just from trying to close to melee range.

No need to make heat damage from it _that_ damaging. Heat damage might pose a serious threat, but nothing an experienced player couldn't cope with. This looks like an interesting challenge and might encourage melee players to find other ways to kill it, or to evade it completely.

GuardianDll:
Okay i just did some stuff, and found the spell realisation would be way worse than you have now. Sorry bout that

TheShadowFerret (approved):

data/json/monster_special_attacks/monster_attacks.json
```diff
@@ -314,5 +314,24 @@
...
+    "cooldown": 12,
+    "move_cost": 400,
+    "range": 2,
+    "damage_max_instance": [ { "damage_type": "bash", "amount": 30 } ],
+    "hitsize_min": 12,
+    "throw_strength": 20,
+    "hit_dmg_u": "%1$s's head rams into your %2$s!",
+    "hit_dmg_npc": "%1$s's head rams into <npcname>!",
+    "miss_msg_u": "%1$s's head comes at you like a guillotine, but misses its mark!",
+    "miss_msg_npc": "%1$s's head comes at <npcname> like a guillotine, but misses its mark!",
```
Fris0uman, line 331: I'm very confused at how a headbut can behave like a guillotine, do they have blade shape head and fall on you from above?

data/json/monsters/zed_misc.json
```diff
@@ -527,6 +527,130 @@
...
```
Fris0uman, line 534:
```suggestion
    "description": "This was at some point a zombie wearing some kind of fire-resistant protective gear.  That same outfit is now shredded and has become part of its outer tissue, making it more resilient against physical attacks and heat-induced damage.  It also seems to be rather muscle packed.  At least its overgrown and hardened faceplate still holds up, preventing it from biting and impairing its vision.",
```
The part about hearing and smelling is a bit akward and make the decription very long. Most zombies can hear and smell, we usually describe something when they can't.

data/json/monsters/zed_misc.json
```diff
@@ -527,6 +527,130 @@
...
+      },
+      {
+        "id": "headbash",
+        "cooldown": 20,
+        "move_cost": 300,
+        "damage_max_instance": [ { "damage_type": "bash", "amount": 16 } ]
+      }
+    ],
+    "upgrades": { "half_life": 42, "into": "mon_zombie_nomex_2" },
+    "death_drops": "mon_zombie_fireman_death_drops",
```
Fris0uman, line 581: This probably includes some turnout gear that this zombie doesn't have anymore since it's merged into its body

data/json/monsters/zed_misc.json
```diff
@@ -527,6 +527,130 @@
...
+    "weakpoint_sets": [ "wps_humanoid_body", "wps_humanoid_turnout_gear", "wps_humanoid_full_helmet_armored" ],
+    "families": [ "prof_intro_biology", "prof_wp_zombie", "prof_wp_syn_armored", "prof_wp_nat_armored", "prof_wp_hulk" ],
+    "vision_day": 6,
+    "vision_night": 3,
+    "harvest": "zombie_nomex",
+    "special_attacks": [
+      { "id": "slam", "cooldown": 12, "move_cost": 220, "damage_max_instance": [ { "damage_type": "bash", "amount": 30 } ] },
+      { "id": "headbash", "move_cost": 600, "damage_max_instance": [ { "damage_type": "bash", "amount": 40 } ] }
+    ],
+    "death_drops": "mon_zombie_fireman_death_drops",
```
Fris0uman, line 636: same issue

data/json/monsters/zed_misc.json
```diff
@@ -527,6 +527,130 @@
...
+      "PUSH_MON",
+      "NO_FUNG_DMG",
+      "FILTHY"
+    ]
+  },
+  {
+    "id": "mon_zombie_nomex_2",
+    "type": "MONSTER",
+    "name": { "str": "nomex hulk" },
+    "description": "This was at some point a zombie wearing some kind of fire-resistant protective gear.  It has grown to an enormous size with bulging muscles and features a hide in the likeness of its former uniform.  The faceplate that once shielded its owner's face against fire has now hardened to more closely resemble a slightly curved, bone-crushing hammerhead.  As a result its ability to see is heavily impaired, but its hearing and sense of smell appear to be working just fine.",
```
Fris0uman, line 603:
```suggestion
    "description": "This was at some point a zombie wearing fire-resistant protective gear.  It has grown to an enormous size with bulging muscles and a hide in the likeness of a firefighter uniform.  The faceplate that once shielded its face has now hardened to into slightly curved, bone-crushing hammerhead.  As a result its ability to see is heavily impaired, but its hearing and sense of smell appear to be working just fine.",
```
I tried to streamline the description a bit

Fris0uman (commented):
This is not a hard blocker but having yet another flavor of brute and hulk is a bit boring in my opinion

Daved27hundred:
is this PR still being worked on? I want to suggest making the evolved form of the Nomex zombie something other than a hulk. Perhaps something fast and capable of leaping like a predator? It would make it stand out from the Kevlar evolutions: Quick yet lightly armored as opposed to Kevlar's bullet resistance and bulk.

EtherealArcturus:
It is still being worked on, I just did not have a lot of time to look at it recently.

data/json/monster_special_attacks/monster_attacks.json
EtherealArcturus: It is just a figure of speech really. From the perspective of the victim it sure looks and feels like a guillotine.

EtherealArcturus:
As far as changing the flavor of the evolution line goes, I can do that. However, that would mean almost all of the work I did so far would go down the drain, so it will take more time to think of something new and implementing.

EtherealArcturus:
If the goal is to make something with more flavor, then I think a predator-like evolution line would still be too generic, since we already have a few of those.

So, I just had an interesting idea. What if, instead of a giant creature that hits like a truck, we just let it become a fire-resistant slab of meat? It would be rooted to the ground and act as a obstacle of sorts, rather than a hunting enemy. That would solve the flavor problem and would be wildly different from pretty much any other zombie evolution in the game.

PatrikLundell:
I believe there are two problems with that:
- An immobile enemy with no ranged attack is a sitting duck; and
- An immobile enemy is pointless to attack unless you can device a mechanism by which it completely blocks off access to something worth getting access to. That could be done by spawns on specific maps, but then it probably wouldn't be much of a progression.

A different approach would be to make it gobble up all lose items in the vicinity (thus destroying them), but without the ability to "breed" (as a certain top tier monster with an area clearing ability has). I can't figure out any reason for why it would be tied to fire resistance, though, or, in fact, any rationale for that behavior.

EtherealArcturus:
Back to the drawing board it is.

Daved27hundred:
> As far as changing the flavor of the evolution line goes, I can do that. However, that would mean almost all of the work I did so far would go down the drain, so it will take more time to think of something new and implementing it.

I don't think it would be that drastic. You don't need to delete what you have, just edit the name and stats of the nomex hulk, and you should be fine.

> If the goal is to make something with more flavor, then I think a predator-like evolution line would still be too generic,

The thing you did with the faceplate mutating into a battering ram is already a really unique idea, probably the coolest I've seen. It'll definitely turn heads if you saw it ingame, and it being similar to a predator wouldn't detract from that at all. 

It's just a suggestion of course, I only said predator because it was the quickest zombie I could think of.

EtherealArcturus:
Heya

A few hours ago I saw the Release of 0.G and the ensuing flood of new PRs that were committed as a result. Not to be outdone, I completed the reflavor of the Nomex Zombie evolution line.
Why did I decide to go with a Predator-like Zombie? Because:

A:
I find the idea of a quick hunter with a chungus hammerhead leaping after you positively horrifying, so that's good.

B:
Despite dedicating almost half a year to the single question of what the ideal Firefighter Zombie evolution line would be, I could not arrive at a better answer. So, there ya go.

I do hope that this flavor of Nomex Zombie is more appropriate than another hulky boy.

PR #60832, opened 2022-09-07
Cody charges more and can make chainmail

anoobindisguise:
#### Summary
Balance "cody can make chainmail armor and charges more"

#### Purpose of change
I felt that Cody's prices for making me plate armor were too low and that I was taking advantage of her. I'm basically asking her to devote 3-5 weeks of her remaining life to smithing me armor, and while she's apparently happy to make the armor at cost, her prices are too low to even break even.
In addition to that, I figured it would be reasonable for her to also make you chainmail armor.

#### Describe the solution
1 Merch is apparently worth 1 piece of jerky, or ~350 calories. Cody has to work for about a month on your armor regardless of how thick the plate armor is, the crafting times for each thickness are very close to each other. 6mm plate armor isn't 3x as hard to make as 2mm plate armor. The amount she charges is based on the amount of metal she's buying, with 100 merch for each 2mm of armor you're requesting. The issue is that this doesn't account for food, and presumably Cody needs to have the food to support her blacksmithing over such a long period of time. Since she's extremely active blacksmithing, I'm operating under the assumption that she needs about 3500 kCal per day, or 10 pieces of jerky. This gives a nice, round price of food for the labor of 300 Merch which is added to the price of whatever armor she's smithing. Thus her prices are 400, 500, 600 which feel more fair to me.
She can also make chainmail armor, which now shows up as an additional thickness of armor you can order (2mm, 4mm, 6mm, 1.2mm). Nomad chain is possible too! Price of this is (3)50 since it's the thinnest, and requires by far the least metal.

Shorten time needed for her to make chainmail. Currently the player can put together a suit in a couple days, but this may be an indication the player's crafting is wrong, rather than Cody being too slow.

#### Testing
I was able to ask Cody to make me the armor, but I didn't try waiting 5 weeks to see if it was successfully delivered. New values look OK in dialogue.


data/json/items/armor/suits_protection.json
```diff
@@ -2925,5 +2925,73 @@
...
+  },
+  {
+    "id": "armor_nomad_chainmail_on",
+    "copy-from": "armor_nomad_chainmail",
+    "repairs_like": "armor_nomad_chainmail",
+    "type": "TOOL_ARMOR",
+    "name": { "str": "nomad chainmail (on)", "str_pl": "suits of nomad chainmail (on)" },
+    "revert_to": "armor_nomad_chainmail",
+    "extend": { "flags": [ "CLIMATE_CONTROL" ] },
+    "power_draw": 100000,
```
Night-Pryanik, line 2988: Could you please represent power draw in Joules?

I-am-Erk:
Please fix conflicts, then this should be OK to merge.

PR #60838, opened 2022-09-07
Add new house - 2 Story Modern 02

LilShiningMan:
#### Summary
CONTENT "Add New House - 2 Story Modern 02"

#### Purpose of change

increasing house spawn variety.

#### Describe the solution

Created new house layout.

#### Testing

Linted, tested in-game. Spawns just fine.

#### Additional context
basement
1st
2nd
roof

### DIFF
--- a/data/json/mapgen/house/2storymodern02.json
+++ b/data/json/mapgen/house/2storymodern02.json
@@ -106,6 +106,7 @@
         "|": "t_strconc_wall",
         "G": "t_floor_noroof",
         "K": "t_floor_noroof",
+        "¶": "t_door_glass_c",
         "$": "t_floor_noroof"
       },
       "furniture": { "$": "f_table" },


data/json/mapgen/house/2storymodern02.json
```diff
@@ -0,0 +1,197 @@
...
+        ".[-----------%o  h  V||.",
+        "..|::::|-----%o     VV|.",
+        ".|| h  ||----%| |||   o.",
+        ".|y ff y|::*::| |>||+||.",
+        ".| hffh      Ly + |S08o.",
+        ".|  ff          |||008|.",
+        ".o  h           <||t9||.",
+        ".o       A A  |o||||||,.",
+        ".|y    6 JJJ  o```````,.",
+        ".|| aa 6      ¶``G````-.",
```
Fris0uman, line 96:
```suggestion
        ".|| aa 6       ``G````-.",
```
Need to replace this character with the correct one, or give it a definition

LilShiningMan: That symbol is what I found in the palette for closed glass door. It spawns correctly, as per the ground floor screenshot.

LilShiningMan: Ok, I'm home now and able to look into this a bit further but still not sure why that error is popping up, as that symbol is listed in standard_domestic_palette

PALETTE - symbol


And also is used in house_08, as called by the same palette

house08


It spawns perfectly in my game, so I'm baffled.

Maleclypse: Can confirm that  the symbol is in `standard_domestic_palette`

LilShiningMan:
Can I request another review of this? I'm not sure how to resolve this error that's being thrown if I'm just calling a symbol defined in the palette and the check doesn't like it for some reason.

data/json/mapgen/house/2storymodern02.json
```diff
@@ -0,0 +1,197 @@
...
+        "....[..%%%%%%%,,,,,,,,,.",
+        "........................"
+      ],
+      "terrain": {
+        ",": "t_railing_v",
+        "0": "t_linoleum_gray",
+        "`": "t_floor_noroof",
+        "|": "t_strconc_wall",
+        "G": "t_floor_noroof",
+        "K": "t_floor_noroof",
```
Maleclypse, line 108:
```suggestion
        "K": "t_floor_noroof",
        "¶": "t_door_glass_c",
```

LilShiningMan:
thanks for that. I guess I was hoping someone might be able to tell me why it was throwing the error in the first place, as in theory it should have been absolutely fine. I have a few more houses that also use the same symbol, so looking to head off similar problems in the future.

Maleclypse:
> thanks for that. I guess I was hoping someone might be able to tell me why it was throwing the error in the first place, as in theory it should have been absolutely fine. I have a few more houses that also use the same symbol, so looking to head off similar problems in the future.

I’m guessing that something is going wrong with selecting multiple palettes but I don’t have time or ideas on how best to test it currently. You might want to make an issue and hopefully someone who’s familiar with how the test suite works can see it.

PR #60839, opened 2022-09-07
Ingot separation

Drew4484:
#### Summary
Content "Recipes to dismantle ingots"

#### Purpose of change
Allows disassembly of 5L, 10L, and the Exodii's gigantic ingots.

#### Describe the solution
This was being worked on by someone else (who is no longer part of the project) so I didn't add them when I created the steel ingots. This picks up where they left off.

None

#### Testing
WIP

### DIFF
--- a/data/json/uncraft/generic.json
+++ b/data/json/uncraft/generic.json
@@ -237,7 +237,7 @@
     "components": [ [ [ "lc_steel_lump", 39 ] ] ]
   },
   {
-    "result": "10l_mcsteel",
+    "result": "10l_mc_steel",
     "type": "uncraft",
     "activity_level": "MODERATE_EXERCISE",
     "time": "10 m",
--- a/data/json/items/resources/plastic.json
+++ b/data/json/items/resources/plastic.json
@@ -30,7 +30,7 @@
     "flags": [ "NO_SALVAGE" ]
   },
   {
-    "id": "10lplastic",
+    "id": "10l_plastic",
     "type": "AMMO",
     "category": "spare_parts",
     "name": { "str": "10L plastic block" },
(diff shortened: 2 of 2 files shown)


Drew4484:
> I was considering giving these "charges" and then using a recipe to craft a chunk of their respective materials with metal sawing and using 1 charge of their "ammo".  It would require making the individual ingots weigh nothing and giving them a fake ammo type with no unload, and making them no reload, then weighting the ammo appropriately, but it would probably allow for more gradual disassembly of the ingots. Only issue I can think of is that it will give you "this X is too large and bulky to disassemble properly" so it could instead be activated with a metal sawing activity.

Not a bad idea but I think charges are on their way out, right?

data/json/uncraft/generic.json
```diff
@@ -52,6 +52,222 @@
     "qualities": [ { "id": "SAW_M", "level": 2 } ],
     "components": [ [ [ "scrap", 5 ] ] ]
   },
+  {
+    "result": "exodii_ingot_tin",
+    "type": "uncraft",
+    "activity_level": "MODERATE_EXERCISE",
+    "time": "16 h",
+    "tools": [ [ [ "angle_grinder", 500 ] ] ],
```
phoeo, line 60: Why use a specific tool, if there are multiple tools with grinding quality?

Drew4484: The largest ingots are nearly a meter across. I'm all for adding other tools but doing it with hand tools isn't practical.

phoeo: Isn't angle grinder a hand tool, too?

What if stronger tools are added in the future? For example, in light industry you can find arc furnaces, and mines have unspecified heavy machinery, which may eventually include industrial grinders.

Drew4484: "Hand tools" in the sense of an unpowered tool, using only the user's strength and any mechanical advantage. We may need to revisit qualities like metal sawing but I'd like to hold off on that until we work out a way to improve craft times based on that tool quality.

PR #60841, opened 2022-09-07
Enviro suit/activity suit allow tail

actually-a-cat:
#### Summary
Bugfixes "Environmental suit and activity suit can be worn by characters who have tails"

#### Purpose of change
Current convention is that garments that have only soft fabric in the tail area can be worn by characters with tails (through some unspecified off-screen adjustment). Enviro/activity suit has rigid materials only on hands and feet so it should allow tails.

#### Describe the solution
Added flag

At some point we'll need a solution for adjusting clothes for tails that makes more sense but it currently is what it is

#### Testing
Created character with long tail, wore both variants, made sure both can transform to the casual version while worn.

### DIFF
--- a/data/json/items/armor/suits_protection.json
+++ b/data/json/items/armor/suits_protection.json
@@ -2529,7 +2529,7 @@
       "msg": "You slip out of the top of the suit and tie the sleeves around your waist.",
       "target": "robofac_enviro_suit_casual"
     },
-    "flags": [ "VARSIZE", "WATERPROOF", "RAINPROOF", "GAS_PROOF", "RAD_PROOF", "ELECTRIC_IMMUNE", "STURDY" ]
+    "flags": [ "VARSIZE", "WATERPROOF", "RAINPROOF", "GAS_PROOF", "RAD_PROOF", "ELECTRIC_IMMUNE", "STURDY", "ALLOWS_TAIL" ]
   },
   {
     "id": "robofac_enviro_suit_casual",


Venera3:
Environmental protection of 20 doesn't really make sense for "yeah I cut a hole for my tail".

actually-a-cat:
Yes, but currently the only reason why these suits don't allow tail is the presence of steel and plastic on the feet and hands, which is incorrect. 

There is no rule about allowing tail depending on level of protection, and other suits with comparable protection and general construction (e.g. hazmat suit, firesuit, entry suit) allow tail because they happen to have no rigid parts, and even if we wanted to go the other way and make them all incompatible with tails, there is no way to do so. The only rule is "is it rigid (on the butt)?" and we add `ALLOWS_TAIL` to correct for the (on the butt) part.

So, while I agree with you, this is a valid fix to bring the items within current rules, unless you're asking me to instead design new mechanics, implement them, and apply them to all similar items.

Venera3:
That's slightly annoying but reasonable - I haven't checked the rest of the suits. I suppose it will get sorted out with the limby armor changes when those start.

actually-a-cat:
In general I would love if having a tail came with an appropriate level of drawback and inconvenience, but for now while we don't have the mechanics for that, and our choices are "pretend it's not a problem" and "make life hell for tailed characters" I think we should stick with the former whenever sensible.

actually-a-cat:
#61521 obsoletes this

PR #60843, opened 2022-09-08
audit chain link crafting times and quantities

anoobindisguise:
#### Summary
Balance "make fewer chain links per craft"

#### Purpose of change
Carbon-steel wires were making 750 chain links per craft, which is too many. The regular wire only makes 100. A wire weighs 151 grams, and each chain link is 1 gram, so a theoretical 100% efficiency could make 151 links assuming 1 link is 1 gram.

#### Describe the solution
Make the recipes only make 100 links, instead of 750. This accounts for some waste in the process and mirrors the regular chain links' recipe.

Even fewer links per wire.


chrispikula:
The Ring Lord website changed, so I can't easily use them as a reference, but a single link is going to way much less than 1 gram.  
[wikipedia](https://en.wikipedia.org/wiki/Chain_mail#Historical_re-enactment) suggests 15-45k for 10kg, which is like .67-.25 grams.  And that's historical.  Modern, I'd expect to weigh less.  Gimme a second, and I'll weigh out some of my costume aluminum stuff.

chrispikula:
72 links (6-in-1 style, unwelded, hand clipped, galvanized steel, just tested as magnetic, made from electric fencing wire) coming in at 23 grams
image

So that's .31g/link +-probably .02 grams.  Call it 3 links per gram, just to make the math somewhat easy.

anoobindisguise:
New paradigm: 1kg steel lump makes 6 0.15 kg steel wires. Takes longer. 0.15 kg steel wire makes 425 330mg chainlinks. 1 0.15 kg  steel wire and 2550 330 mg chainlinks makes 1 1kg chainmail sheet. That takes a lot longer (10 hours instead of 1 hour). Hopefully the math should be slightly more sensible now.

This means that it takes about 11 hours of labor to make 1 sheet, when previously it took about 100 minutes. That's probably a more fair representation of the total effort to make a suit of chainmail (a fuckton)

chrispikula:
> This means that it takes about 11 hours of labor to make 1 sheet, when previously it took about 100 minutes. That's probably a more fair representation of the total effort to make a suit of chainmail (a fuckton)

Agreed.  There's a reason I never finished my 6-in-1 vest.  Wrists damn near died.  lol

PR #60844, opened 2022-09-08
Add a tab for player visible vitamins/counters

Maleclypse:
#### Summary
Interface "Adds tab showing appropriate for player visible vitamins/counters"

#### Purpose of change
The vitamin system has tremendous potential for use in in cause and effect eocs.   Part of the limiting factors on them currently is that all vitamin/counter charges can only be viewed in debug.  The goal of this PR is to give devs a method in json to list a vitamin as `"display": true` and that will cause it to display on the @ screen.

#### Describe the solution
Adds a new tab for vitamins.  Needs a lot more work though

I'd like for the tab to not be visible unless the PC has acquired either their first charge of a visible vitamin or a means of naturally regening the charge.

### DIFF
--- a/src/player_display.cpp
+++ b/src/player_display.cpp
@@ -39,6 +39,7 @@
 #include "ui_manager.h"
 #include "units.h"
 #include "units_utility.h"
+#include "vitamin.h"
 #include "weather.h"
 #include "weather_type.h"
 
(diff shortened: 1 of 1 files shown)


Maleclypse:
Outside of my skill range

PR #60850, opened 2022-09-08
Fix helicopter total rotor area calculation

SurFlurer:
#### Summary
Bugfixes "Fixed helicopter total rotor area calculation"

#### Purpose of change
The current calculation of rotor area first adds up all diameter, then uses the result as a new diameter to get the area of a round. In fact, this is not the case. Two diameter-2 rounds don't have the same area as one diameter-4 round. So the rotor area calculated on a multi-rotor helicopter is different from in reality.

#### Describe the solution
I changed the calculation method to a real one.

Leave it as is, since multi-rotor helicopters are rare.

#### Testing
I apologize for that I'm currently unable to compile to test.

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -4121,7 +4121,7 @@ double vehicle::lift_thrust_of_rotorcraft( const bool fuelled, const bool safe )
 {
     int rotor_area_in_feet = 0;
     for( const int rotor : rotors ) {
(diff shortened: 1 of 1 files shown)


SurFlurer:
I've test this particular part of code. This PR doesn't introduce differences in value on the result of single-rotor helicopter's thrust calculation.

PR #60851, opened 2022-09-08
Connecting power cables between appliances and vehicles

### LINKED ISSUE #59169: Extension cord cannot be plugged into vehicle
### Describe the bug

You can connect an appliance with a nearby vehicle, but cannot connect an appliance with a vehicle using extension cord. You can do so using a jumper cable or heavy-duty cable though.

### Steps to reproduce

1. Put up appliance
2. Construct vehicle
3. Plug extension cord into the appliance
4. Cannot plug the extension cord into the vehicle

### Expected behavior

Being able to plug extension cord into a vehicle. Alternatively, disallow connecting an appliance to a vehicle without using a jumper cable or heavy-duty cable, and make it clear in the item description and show a warning message when the player tries to do so.

### Screenshots

_No response_

### Additional context

_No response_

dseguin:
#### Summary
None

#### Purpose of change
- Fixes #59169

The other concern pointed out in the comments in the linked issue is that you shouldn't be able to transfer power directly between vehicles and appliances, since vehicles run on DC power while most appliances connect to AC sources. In reality that would require some kind of rectifier (like the ones found in power supplies) or a commercial power inverter:

image

#### Describe the solution
- [x] Fix the linked issue - *(extension cables can connect to vehicles)*
- [x] Add power inverters - *(using an abstract term here, the nitty-gritty details are not important)*
  - [x] Item action for converting cables with a power inverter - *(can be activated from the cable or the inverter)*
  - [x] Restrict appliance-vehicle connections to cables with an attached inverter
  - [x] Add recipe and item group spawns for power inverters
- [x] Unit test

So the order of operations for connecting an appliance to a vehicle would be:
1. Get an extension cable and power inverter
2. `a`ctivate either the cable or the inverter to attach the inverter to the cable
3. `a`ctivate the converted cable and connect both the appliance and the vehicle

https://user-images.githubusercontent.com/12537966/189047926-4f479bcb-1f7c-467e-9254-417fefeeb9d5.mp4

Forgetting the AC-DC power conversion and just fixing the linked issue. In that case, the fix would only be 2 lines:

```diff
diff --git a/src/iuse.cpp b/src/iuse.cpp
index 713caf1d23..18950a50f1 100644
--- a/src/iuse.cpp
+++ b/src/iuse.cpp
@@ -8850,7 +8850,7 @@ cata::optional<int> iuse::cord_attach( Character *p, item *it, bool, const tripo
         }
         const tripoint posp = *posp_;
         const optional_vpart_position vp = here.veh_at( posp );
-        if( !vp || !vp->vehicle().has_tag( "APPLIANCE" ) ) {
+        if( !vp ) {
             p->add_msg_if_player( _( "There's no appliance here." ) );
             return cata::nullopt;
         } else {
@@ -8911,7 +8911,7 @@ cata::optional<int> iuse::cord_attach( Character *p, item *it, bool, const tripo
         const tripoint vpos = *vpos_;
 
         const optional_vpart_position target_vp = here.veh_at( vpos );
-        if( !target_vp || !target_vp->vehicle().has_tag( "APPLIANCE" ) ) {
+        if( !target_vp ) {
             p->add_msg_if_player( _( "There's no appliance there." ) );
             return cata::nullopt;
         } else {
```

#### Testing
Added a unit test to check that the power inverter iuse action works from both the inverter and the cable:

```
./tests/cata_test --rng-seed time "power inverter"
```

Also tested in-game:
- Converting a cable by activating the cable :heavy_check_mark:
- Converting a cable by activating the inverter :heavy_check_mark:
- Detaching the inverter by activating the converted cable :heavy_check_mark:
- Attaching a normal cable from appliance to appliance :heavy_check_mark:
  - Cannot use a power cable with an inverter attached
- Attaching a normal cable from vehicle to vehicle :heavy_check_mark:
  - Cannot use a power cable with an inverter attached
- Attaching a converted cable from appliance to vehicle :heavy_check_mark:
  - Cannot use a power cable without an inverter

### DIFF
--- a/data/mods/TEST_DATA/appliance.json
+++ b/data/mods/TEST_DATA/appliance.json
@@ -57,7 +57,7 @@
     "price_postapoc": 100,
     "max_charges": 3,
     "initial_charges": 3,
-    "use_action": [ "CABLE_ATTACH" ],
+    "use_action": [ "CABLE_ATTACH", "INVERT_CURRENT" ],
     "flags": [ "CABLE_SPOOL", "POWER_CORD" ]
   },
   {
@@ -85,7 +85,7 @@
     "copy-from": "jumper_cable",
     "max_charges": 10,
     "initial_charges": 10,
-    "use_action": [ "CORD_ATTACH" ]
+    "use_action": [ "INVERT_CURRENT" ]
   },
   {
     "type": "vehicle_part",
@@ -105,5 +105,21 @@
     "requirements": { "removal": { "time": "5 s" } },
     "flags": [ "NOINSTALL", "UNMOUNT_ON_DAMAGE", "UNMOUNT_ON_MOVE", "POWER_TRANSFER" ],
     "breaks_into": [ { "item": "cable", "charges": [ 1, 10 ] }, { "item": "plastic_chunk", "count": [ 1, 2 ] } ]
+  },
+  {
+    "id": "test_power_inverter",
+    "type": "GENERIC",
+    "category": "spare_parts",
+    "name": { "str": "power inverter" },
+    "description": "This is an implement that rectifies the current between AC and DC sources.  Can be fitted to a power cable for connecting appliances and vehicles.",
+    "weight": "50 g",
+    "volume": "50 ml",
+    "price": 1000,
+    "price_postapoc": 10,
+    "material": [ { "type": "plastic", "portion": 4 }, { "type": "copper", "portion": 1 } ],
+    "symbol": ":",
+    "color": "green",
+    "use_action": [ "INVERT_CURRENT" ],
+    "flags": [ "INVERTER_SOURCE", "SINGLE_USE" ]
   }
 ]
--- a/tests/iuse_test.cpp
+++ b/tests/iuse_test.cpp
@@ -616,3 +616,51 @@ TEST_CASE( "xanax", "[iuse][xanax]" )
     }
 }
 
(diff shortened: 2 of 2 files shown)


actually-a-cat:
If we are bringing in the AC/DC distinction, does it make any sense to limit it to just vehicles? Solar panels and batteries are also DC and would require inverters to power AC appliances, and charging batteries from AC sources would require a charger (rectifier and regulator). Plus it's not even necessarily true that vehicle = DC. The "Portable Generator" vehicle should output mostly AC.

Is this a can of worms worth opening?

dseguin:
I'm not really attached to the power conversion stuff, I'm fine with the simple fix as well. But I see your point about current types in vehicles/appliances. I'll add a way for vehicle/appliance definitions to declare how they should be treated in the grid.

Ciac32:
I don't think the AC/DC distinction is as big as a deal for anything other than the big motor appliances. Lamps and heating elements don't care what the current is doing, and DC will make it through a rectifier fine for any of the control circuits (albeit not at peak efficiency/reliability). In addition, universal motors are surprisingly common in smaller appliances and will cope well with either type of current.

The bigger issue is voltage & amperage, but that is already pretty abstracted from what the survivor is doing. If its really that important we could just make a middling electronics skill and the generic "power converter" item required for some of the more sensitive types of appliances; presumably that would cover any glaring AC/DC problems too

PR #60859, opened 2022-09-08
Added batch time savings to several cooking recipes that needed it.

Daved27hundred:
#### Summary

None

#### Purpose of change

Certain foodstuffs didn't have any batch savings when they theoretically should. This was originally going to just be for starch, but I decided to do several other recipes.

#### Describe the solution

looked at some analogues to recipes, took their batch savings and gave them to the recipe. For instance several baking recipes have a saving of `[ 50, 4 ]` so this was applied to the pancake recipes.

#### Additional context

Venera3 mentioned it doesn't seem right for the soup recipes to not need a cutting tool, someone should look at that.


Maleclypse (approved):
Seems reasonable

PR #60862, opened 2022-09-08
Introduce "draped over legs" sublocation for robes and dusters

anoobindisguise:
#### Summary
Content "Robes and dusters can be worn with leg armor"

#### Purpose of change
Dusters and robes don't really occupy the same space on the legs as leg armor would. It has been previously discussed on the dev discord the possibility of adding a layer to represent that.

#### Describe the solution
There are now the "leg_draped_l" and "leg_draped_r" sublocations. These are SECONDARY like front and back torso hanging items, and as a result they are BELTED but can be worn with other leg-occupying BELTED items such as scabbards, holsters, drop pouches, etc.

MASSIVE thanks to @bombasticSlacks for helping me fix some errors with this. Couldn't have done it without you.

Adjust coverage - for example a duster covers 86% of the legs - I kept these values the same for their new layers but they could change.
Probably needs to change over maid clothes too. Should probably be saved for the inevitable Kevlar maid clothes PR though.

#### Testing
I initially loaded this in game and found that while the coverage worked and layering worked, it caused all leg armor to say "The Hips. The Thighs. The Knees. The Lower Legs" instead of just the legs. This broke a test. It also complained that these SECONDARY locations were being used on non-belted armor, even though they weren't - the armor itself was OUTER causing the problem. However, Bombastic ended up helping me fix these, really grateful for that!

### DIFF
--- a/data/json/items/armor/coats.json
+++ b/data/json/items/armor/coats.json
@@ -751,7 +751,7 @@
     "color": "blue",
     "armor": [
       { "covers": [ "torso" ], "specifically_covers": [ "torso_lower" ], "coverage": 100, "encumbrance": [ 11, 20 ] },
-      { "covers": [ "torso" ], "specifically_covers": [ "torso_upper" ], "coverage": 95 },
+      { "covers": [ "torso" ], "specifically_covers": [ "torso_upper" ], "coverage": 95, "encumbrance": [ 0, 0 ] },
       {
         "covers": [ "leg_l", "leg_r" ],
         "coverage": 79,
@@ -865,12 +865,12 @@
     "looks_like": "coat_lab",
     "color": "blue",
     "armor": [
-      { "covers": [ "torso" ], "specifically_covers": [ "torso_lower" ], "coverage": 100, "encumbrance": 10 },
+      { "covers": [ "torso" ], "specifically_covers": [ "torso_lower" ], "coverage": 100, "encumbrance": [ 0, 0 ] },
       { "covers": [ "torso" ], "specifically_covers": [ "torso_upper" ], "coverage": 95, "encumbrance": [ 5, 10 ] },
       {
         "covers": [ "leg_l", "leg_r" ],
         "coverage": 75,
-        "encumbrance": 5,
+        "encumbrance": [ 5, 5 ],
         "specifically_covers": [ "leg_draped_l", "leg_draped_r" ],
         "layers": [ "BELTED" ]
       },
@@ -2330,7 +2330,8 @@
       {
         "covers": [ "arm_l", "arm_r" ],
         "specifically_covers": [ "arm_shoulder_l", "arm_shoulder_r" ],
-        "coverage": 40
+        "coverage": 40,
+        "encumbrance": [ 0, 0 ]
       }
     ],
     "pocket_data": [
@@ -2409,7 +2410,8 @@
       {
         "covers": [ "arm_l", "arm_r" ],
         "specifically_covers": [ "arm_shoulder_l", "arm_shoulder_r" ],
-        "coverage": 40
+        "coverage": 40,
+        "encumbrance": [ 0, 0 ]
       }
     ],
     "pocket_data": [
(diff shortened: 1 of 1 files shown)


Terrorforge:
So this functions as a layer, then? That is, it hangs "over" other leg armor, and will tend to absorb damage before they do?

Also, and I'm not saying this needs to be done in this PR or by you, but this reasonably applies to skirts and dresses as well.

It also applies to the various gambesons and chainmail hauberks. They're both supposed to be coat-like garments, but are nothing like that in the game currently, as up until now it hasn't simulated that very well. The current gambeson can probably stay, maybe with a rename, since it's essentially an arming doublet already. I might see about adding more traditional coat-style gambeson as an "arming jack" or something once this is in.

anoobindisguise:
> So this functions as a layer [...] arrived at this solution instead

Not super concerned about blankets, although they could be moved to strapped - it's more or less thrown over you, but the encumbrance is already so ridiculously high that you already don't really care about layering penalties.
Skirts etc can definitely be moved to the layer, I just touched coats.
As far as gambesons go, I definitely plan to make them "doublets" and "coats", with a similar principle - coat is what you wear on OUTER, doublet is what you wear on INNER, and they have different thicknesses (3mm vs 5mm is current plan, based on 28-30 layer gambesons of ye olden times). But that'll require a lot of out of scope work.

PatrikLundell:
While you may be able to wear a scabbard under a dress (or robe), actually accessing its contents ought to be fairly clumsy and time consuming, as you'd have to hike up the fabric first in order to actually be able to reach the scabbard.
I'm not sure it would work under a skirt, since whatever you'd have in a scabbard may well poke up over the waist line the skirt is hanging from. A robe hanging from the shoulders wouldn't have that problem, but a dress with a waist line probably would.

Terrorforge:
People do love their cloaks, though. I don't think it's something we need to lose sleep over, but it is... unsightly that, as far as the game is concerned, you wear an armband pouch _over_ your cloak.

> While you may be able to wear a scabbard under a dress (or robe), actually accessing its contents ought to be fairly clumsy and time consuming, as you'd have to hike up the fabric first in order to actually be able to reach the scabbard. I'm not sure it would work under a skirt, since whatever you'd have in a scabbard may well poke up over the waist line the skirt is hanging from. A robe hanging from the shoulders wouldn't have that problem, but a dress with a waist line probably would.

This would be true of coats as well. Although you technically could, you almost certainly wouldn't wear a sword inside your gambeson. If you did, it would be mighty hard to draw.

That's mainly an artifact of how holsters and scabbards are implemented, though. They sit in the leg's "hip" sublocation completely independently, when realistically speaking they ought to be accessories you attach to your belt. Given that the MOLLE framework now exists, it probably wouldn't be too hard to implement that if we wanted, but I recall there was a desire to get rid of the "belt" location entirely. Plus the MOLLE stuff is intentionally very optional, and I'm not sure it's a great idea to make an accessory system like that mandatory for basic gear like holsters.

It's something to look at, but shouldn't block this PR. It will create a somewhat weird situation, but mechanically it's not a big deal and in terms of verisimilitude it's much less noticeably weird than the current situation where a trenchcoat will conflict with kneepads but let you wear a thigh holster over it without issue.

Termineitor244:
> Should probably be saved for the inevitable Kevlar maid clothes PR though

HAHAHA you have summoned me from the abyss! Behold how a new wave of maid clothes PRs take over the repository! (Said by the one that added the Lycra bodysuits and apron and bikini variants!).

In all seriousness, I had the plan to implement some modified/reinforced civilian clothing to the game since ages ago, and I have some maid clothing variants pending to implement, so this PR can be a good opportunity/excuse to start working on those doable (With my limited time) projects.

anoobindisguise:
> > Should probably be saved for the inevitable Kevlar maid clothes PR though
> 
> HAHAHA you have summoned me from the abyss! Behold how a new wave of maid clothes PRs take over the repository! (Said by the one that added the Lycra bodysuits and apron and bikini variants!).
> 
> In all seriousness, I had the plan to implement some modified/reinforced civilian clothing to the game since ages ago, and I have some maid clothing variants pending to implement, so this PR can be a good opportunity/excuse to start working on those doable (With my limited time) projects.

Hell yeah. Battle maid time. Gotta get those stylish bonuses.

PR #60873, opened 2022-09-09
Add lifter mech to TCL mech roulette

anoobindisguise:
#### Summary
Balance "Lifter mech can spawn in TCL"

#### Purpose of change
Lifter mech is cool and it only spawns in a couple lab finales.

#### Describe the solution
Lifter mech can spawn in TCL in the same place that the recon and combat mechs spawn.

give it a bespoke, unique location instead of behind a wall in the armory.


Fris0uman:
Does it make sense to have an utility mech spawn in a military section of the tcl?

PR #60881, opened 2022-09-09
Auto-retract "3D" sprites for walls etc.

mlange-42:
#### Summary
Interface "Auto-retract '3D' sprites for walls etc."

#### Purpose of change

Mitigate occlusion problems with high sprites, particularly with ISO tilesets.

Adds auto-retracting near player/screen center as option, in addition to retration toggle as added in #60471.

#### Describe the solution

Add a third retract-mode, where walls retract automatically near the player or looked-at position. See screenshot/GIF for illustration.

* Add a tile retract option value "Auto", in addition to "Normal" and "Retracted" (change from `bool` to `int`)
* In auto mode, retract linearly depending on distance to screen center, between minimum and maximum distance
* Add optional properties `retract_dist_min` and `retract_dist_max` to tileset info JSON
* Add options for min and max distance to debug options; overwrite tileset values if set to > 0

* Add the distance settings only to options, or only to tilesets

#### Testing

Tested in-game with different tileset and options values.

#### Additional context

GIF showing the feature with tileset Ultica-ISO:

auto-retract

Things to note:

* Tilesets can specify minimum and maximum distance, as best values are tileset-dependent
* With no distance values given in a tileset, defaults to values that disable automatic retract
* Users can set distance values in theoptions to overwrite tileset values

### DIFF
--- a/src/point.cpp
+++ b/src/point.cpp
@@ -1,5 +1,6 @@
 #include "point.h"
 
+#include <cmath>
 #include <algorithm>
 #include <sstream>
 #include <string>
@@ -39,7 +40,7 @@ point point::rotate( int turns, const point &dim ) const
 
 float point::distance( const point &rhs ) const
 {
-    return std::sqrt( static_cast<float>( pow( x - rhs.x, 2 ) + pow( y - rhs.y, 2 ) ) );
+    return std::sqrt( static_cast<float>( std::pow( x - rhs.x, 2 ) + std::pow( y - rhs.y, 2 ) ) );
 }
 
 int point::distance_manhattan( const point &rhs ) const
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -2549,12 +2549,12 @@ void options_manager::add_options_debug()
        );
 
     add( "RETRACT_DIST_MIN", "debug", to_translation( "Minimum distance for auto-retracting walls" ),
-         to_translation( "Minimum distance for auto-retracting walls. Values above zero overwrite tileset settings." ),
+         to_translation( "Minimum distance for auto-retracting walls.  Values above zero overwrite tileset settings." ),
          0.0, 60.0, 0.0, 0.1
        );
 
     add( "RETRACT_DIST_MAX", "debug", to_translation( "Maximum distance for auto-retracting walls" ),
-         to_translation( "Maximum distance for auto-retracting walls. Values above zero overwrite tileset settings." ),
+         to_translation( "Maximum distance for auto-retracting walls.  Values above zero overwrite tileset settings." ),
          0.0, 60.0, 0.0, 0.1
        );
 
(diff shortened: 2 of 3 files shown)


Golfavel:
For clarification, this and https://github.com/CleverRaven/Cataclysm-DDA/pull/60471 are also available for orthographic tilesets, correct? These additions could potentially close https://github.com/I-am-Erk/CDDA-Tilesets/issues/327 if so (once tilesets implement it).

mlange-42:
> For clarification, this and #60471 are also available for orthographic tilesets, correct? These additions could potentially close [I-am-Erk/CDDA-Tilesets#327](https://github.com/I-am-Erk/CDDA-Tilesets/issues/327) if so (once tilesets implement it).

Yes, this works with ortho tiles too. To fix the named issue, some work on tilesets would still be necessary, as retract behaviour must be defined there per sheet. This is intentionally implemented so that existing tilesets don't change their behaviour unless adding the required properties.

Technically works for trees, but it will probably not look as good as for walls and other cuboid shapes.

mlange-42:
> This pull request **introduces 2 alerts** when merging [3d4e71c](https://github.com/CleverRaven/Cataclysm-DDA/commit/3d4e71cf07495d3c99b130406e833607a5800a41) into [5986364](https://github.com/CleverRaven/Cataclysm-DDA/commit/5986364ce2f2b3150a376ebb9b6998e2e1e5c978) - [view on LGTM.com](https://lgtm.com/projects/g/CleverRaven/Cataclysm-DDA/rev/pr-7099283ba897d5a0de07f28df4e1ecae40502e2a)
> 
> **new alerts:**
> 
>     * 2 for Use of c-style math functions

Fixed in 5ffe686

src/point.cpp
```diff
@@ -37,6 +37,16 @@ point point::rotate( int turns, const point &dim ) const
     return *this;
 }
 
+float point::distance( const point &rhs ) const
+{
+    return std::sqrt( static_cast<float>( std::pow( x - rhs.x, 2 ) + std::pow( y - rhs.y, 2 ) ) );
```
BrettDong, line 42: Looks like you need to `#include <cmath>` in `point.cpp`.

mlange-42: Thanks! Wanted to fix the above LGTM complains here. Compiled locally, but does seemingly not work on every system.

src/point.cpp
```diff
@@ -37,6 +38,16 @@ point point::rotate( int turns, const point &dim ) const
     return *this;
 }
 
+float point::distance( const point &rhs ) const
+{
+    return sqrt( static_cast<float>( pow( x - rhs.x, 2 ) + pow( y - rhs.y, 2 ) ) );
```
BrettDong, line 43: This line also needs to be updated.

mlange-42: Ah, ok... thought I have to remove the `std::`, but this gives the above LGTM alerts. So I guess including `cmath` and keeping `std::` is the way to go....

mlange-42: Done. No respective CI fails anymore.

PR #60888, opened 2022-09-10
tweaked soup ingredient amount/calories

### LINKED ISSUE #57848: Crafting Woods Meat Soup Output to low (probably the case with most soups)
### Describe the bug

Most Soup Recipes are weird that the Volumes dont really match up correctly.

For Example the Woods Meat Soup 3L Version (the Version i cooked)
Requires
3L of Bone Broth/Broth/Pine Needle Tea
1,5l Of Meat (Salted Meat Slices in my Point)
3l Of Veg (canned or cooked Beans)
2,5l of Water (which will be discarded as a byproduct)

And it Makes 3L of Soup

### Steps to reproduce

Craft the Woods Meat Soup 3l Version

### Expected behavior

The Recipe Should Require 2 x 3l Jars and Create Double the Amount
This is probably a case with most Soups kinda.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19042.1706 (20H2)
- Game Version: 01f7b57 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex]
]

### Additional context

_No response_

yanpisani:
#### Summary

Balance "Tweaked values for 3L soups"

#### Purpose of change

Fixes #57848 

#### Describe the solution

- For every 3L Jar soup, the amount of ingredients necessary for its creation were changed (most were halved, some needed a bit more than half). 
- Reason for this is the amount of ingredients necessary to create a 3L soup would output a much larger amount than the actual final output (single 3L jar of soup).

I've considered following @estebandellasilva suggestion and doubling the output of the soup and jars necessary to make the soup. But if that were the case, the player would be unable to make a singular 3L Jar of Soup. In that case, I've decided to instead change the ingredients necessary to make the soup. So now the soup maker still has to use a relatively large amount of ingredients, but not as much as before, but now have a realistic output.

#### Testing

Using the debug feature, I've crafted -a lot- of soups, and it worked just fine. The change was relatively simple, so everything works as it should. 
Utilizing the tests, I've managed to tweak the value of some soup calories and ingredients (specifically, the fish soup and the mushroom soup) so that it'd pass the test, and still have reduced ingredients. 

#### Additional context

Utilizing select ingredients for Woods Meat Soup as an example for context:

Previously it would be necessary:
- 12 Broth (3L)
- 6 Salted Meat Slices (3L)
- 12 Canned Beans (6L)
- 10 Clear Water (2.5L, discarded as byproduct)

Even with the discarded Clear Water, we're still left with 12L of soup in a 3L jar!

Now, with the changes made:
- 6 broth (1.5L)
- 3 Salted Meat Slices (1.5L)
- 6 Canned Beans (3L)
- 10 Clear Water (2.5L, discarded as byproduct)

Now we're left with 6L of soup in a 3L jar. Still a bit more than the volume of the jar, but better than before. As well as simplifying the soup recipe.

(reopened PR due to making an incorrect pull request earlier!)


RenechCDDA:
I'm not sure I understand the intentions behind these changes... you are specifically changing *only* the jarred forms of soups? Not the uncontained soup which you could make outside of the jar and then put in the jar? Not the same soup in the same size tin can container?

I do agree there is an existing volume issue - soups require ingredients which collectively take up more volume than the resulting soup. But that is mostly because many items gravitate towards having a 'uniform' size of 250mL per unit, which is much too large for many things.

yanpisani:
@RenechCDDA, I completely understand where you're coming from. I focused on changing the 3L sizes specifically to solve the requested issue, but the general volume problem remains. We could attempt to alter the sizes away from the 250ml standard, or just accept the existing volume issue.

Fris0uman:
So this recipe is basically a canning recipe right? It gives soup in a jar that will not rot in a long time? Because if not you could just remvoe the jar from the recipe and just have it give you 6L of liquid that the player has to put in whatever they can

PR #60890, opened 2022-09-10
Correct encumbrance for ballistic vests

anoobindisguise:
#### Summary
Bugfixes "ballistic vests now correctly define encumbrance for their pockets"

#### Purpose of change
Several ballistic vests (most notably the soft armor vest) didn't properly define encumbrance-per-pouch, causing them to have 8 times as much encumbrance as liters of volume. For example "press ballistic vest" had 1 liter of space but using that space took up 8 encumbrance.

#### Describe the solution
Pockets now define encumbrance correctly. I went with 1:1 volume:encumbrance on the 0.5L pockets, and 0 encumbrance for the pen holder pocket since it's 0.01L and basically insignificant.

none

### DIFF
--- a/data/json/items/armor/ballistic_armor.json
+++ b/data/json/items/armor/ballistic_armor.json
@@ -688,7 +688,7 @@
         "max_item_length": "25 cm",
         "moves": 50,
         "description": "Low-profile zipper storage pouch.",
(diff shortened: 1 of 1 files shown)


bombasticSlacks:
.4 to .3 is the normal torso encumbrance per volume

PR #43288, opened 2020-08-28
some fliers, notes, news

Jerimee:
#### Summary

SUMMARY: Content "adds a handful of new fliers and news reports"

#### Purpose of change

I'm trying my hand at writing textual ephemera for the game.  These snippets build upon existing lore as found in newspapers, lab reports, and fliers.

#### Describe the solution

One new survivor note:

_...bioluminescent, 5' diameter⋯  appears to be an immense disembodied eyeball, capable of flight and [illegible] thought_

Six fliers:

1. _This ancient missing persons poster is faded and torn.  It features a photo of a stern looking woman in her early fifties.  It reads "Missing Adult - Prof. Amy Takatoshi - Last Seen Thursday, July 12 - Please contact the Boston Police Missing Persons Unit."_
2. _Join the Concerned Consumers of Foodpeople.  Join the fight to⋯  -Stop the neutralization of cultural icons.  -Endow Foodperson with vitals that befitting heroics superb.  -Protect RU-238 and other FRUIT flavors from the threat of prohibition.  -Inspire future consumers of the superhero genre._
3. _This is an ad for a job fair.  \"Summer Internship & Job Fair!  Open to students, alumni, and the public.  Both free and unpaid internships available.  Live and virtual representatives from exciting organizations like: Actem Labs, Cuppatech, MindStone, Edison, NourishMart, and more!\"_
4. _This scrap of paper features a faded ad for Cascade Cola.  \"All the sweet you can stand.  Don't just drink it; let it cascade into you.\"_
5. _This promotional palm card is emblazoned "DigiMart has E-Ink" in a garish purple typeface.  "E-Ink Tablets come packed with octo-core cpu, veripixel lenses camera, neoTouch screen, DRM locket, InnerVision ready near field, a free datavault storage plan, wi-fi, and more!"_
6. _What you've got here is an old coupon insert for Marigold Market:\n\n - BOGO!  Cascade Cola six-packs!\n - ATOMIC POWER THIRST!  Now with 20% more FRUIT RU-238!\n - $2 MEAT! Aunt Janice's Pickled Meat now just $2.49!\n - 75% OFF! DaiZoom bars by SoyPelusa._
7. _From Your Devoted Sister_  - see comments or files changed for content

Four news reports:

[...]

PR #60899, opened 2022-09-10
Hint about other files on the lore page

NetSysFire:
#### Summary
None

#### Purpose of change

While digging for lore in the newspapers I thought it might be helpful to mention the newspapers in the lore document. I also found a mention about bees getting bigger (from #43288), which may be a light-hearted addition since bees getting bigger is basically an external running gag but I decided to repurpose it into serious lore.

#### Describe the solution

I decided to make the month of appearance march since this is when bees have their first spike of activity or something.

Adding giant fuzzy pacifist bumblebees just because.

#### Testing

n/a

#### Additional context

I am annoyed about missing lore and decided I will partially take matters into my own hands (also see the living world project: https://github.com/orgs/CleverRaven/projects/5).

### DIFF
--- a/lore-background.md
+++ b/lore-background.md
@@ -67,7 +67,6 @@ Additional information can also be found in the [newspaper snippets](https://git
 - **December 18**: Sample XE-037 is brought back to Earth from the netherum. It is about thirty grams of blob matter. Standard decontamination protocols are observed. However, standard protocols do not take into account the possibility of malicious, intelligent molecular-scale contaminants. Microscopic particles of XE-037 migrate from the contamination filters onto the skin of scientists and escape containment. Soon after, the Blob enters the groundwater and begins to spread at extremely low concentrations.
 
 #### Year -1
-- **March**: The first effects of the blob contamination become visible as bees two or even three times their size are sighted. They are non-aggressive and appear to cause more good than harm due to increased honey production.
 - **July**: Several smaller powered exoskeletons, including both the private defence contractor Rivtech's version and a US government one, are released for field use.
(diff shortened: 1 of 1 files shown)


Maleclypse (dismissed):
Ties together details in a more concrete fashion.

lore-background.md
```diff
@@ -66,6 +67,7 @@ Note: Canon date of game start is in "YEAR", which equals the IRL year the game
 - **December 18**: Sample XE-037 is brought back to Earth from the netherum. It is about thirty grams of blob matter. Standard decontamination protocols are observed. However, standard protocols do not take into account the possibility of malicious, intelligent molecular-scale contaminants. Microscopic particles of XE-037 migrate from the contamination filters onto the skin of scientists and escape containment. Soon after, the Blob enters the groundwater and begins to spread at extremely low concentrations.
 
 #### Year -1
+- **March**: The first effects of the blob contamination become visible as bees two or even three times their size are sighted. They are non-aggressive and appear to cause more good than harm due to increased honey production.
```
Fris0uman, line 70: I don't think bigger bees would make more honeys, actually bigger bees would probably need more pollen which would either be impossible to aquire enough to sustain themselves or could lead to a significant decrease in biodiversity as the giant bees overcompete every other pollinating insects.

I-am-Erk: That is actually meant as an unreliable narration. Bigger bees would be terrible.

Fris0uman: So not a thing to inscribe in the design doc as being actually true

kevingranade, line 70:
```suggestion
```

PR #60929, opened 2022-09-11
Jsonize electroreceptors

irwiss:
#### Summary
None

#### Purpose of change

From discord:
`[18:48]GuardianDll: oof, electroreceptors are hardcoded 😦`

#### Describe the solution

Move electroreceptors vision bonus to SIGHT_RANGE_ELECTRIC enchantment, make the value represent the range in tiles.

Move extra electro damage pain to EXTRA_ELEC_PAIN enchantment and handle it in deal_damage_handle_type. It has side effect; any electric damage on characters (e.g. melee hits from shockers) now also does extra pain, not just electric fields on map.

#### Testing

Apply patch, find a robot behind wall, debug electroreceptors mutation off/on - it should still work when on, and not work when off

Add SIGHT_RANGE_ELECTRIC to something like aftershock heat retention cream effect - activating it should give the electroreceptors effect

With electroreceptors mutation active the extra pain from electrocution should still produce "You're painfully electrocuted" message in addition to zapped effect even from melee hits.

#### Additional context

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -5510,8 +5510,9 @@ float Character::active_light() const
 
 bool Character::sees_with_specials( const Creature &critter ) const
 {
-    double elec_sight_range = calculate_by_enchantment( 0.0, enchant_vals::mod::SIGHT_RANGE_ELECTRIC );
-    if( critter.is_electrical() && rl_dist_exact( pos(), critter.pos() ) <= elec_sight_range ) {
+    const double sight_range_electric = calculate_by_enchantment( 0.0,
+                                        enchant_vals::mod::SIGHT_RANGE_ELECTRIC );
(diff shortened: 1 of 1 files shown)


src/character.cpp
```diff
@@ -5511,8 +5510,8 @@ float Character::active_light() const
 
 bool Character::sees_with_specials( const Creature &critter ) const
 {
-    // electroreceptors grants vision of robots and electric monsters through walls
-    if( has_trait( trait_ELECTRORECEPTORS ) && critter.is_electrical() ) {
+    double elec_sight_range = calculate_by_enchantment( 0.0, enchant_vals::mod::SIGHT_RANGE_ELECTRIC );
```
Night-Pryanik, line 5513:
```suggestion
    const double elec_sight_range = calculate_by_enchantment( 0.0, enchant_vals::mod::SIGHT_RANGE_ELECTRIC );
```

PR #60936, opened 2022-09-12
Warped trees

NetSysFire:
#### Summary
Content "Add portal-twisted dead trees (cosmetic-only)"

#### Purpose of change

Currently there are only generic dead trees and I felt like adding some cosmetic-only varieties which signify portal damage. This should basically imply that everything is affected, shaping the environment to be more "immersive".

#### Describe the solution

Adds some pseudo terrain for this so the dead tree generation can remain dynamic. Also directly modifies the terrain groups so the changes will apply anywhere where those groups are used. This will also "replace" the dead tree entries so the probabilities of a dead tree being generated stays the same in forests.

I specifically targeted labs as you can see, meaning it will be more likely to encounter those variants there. Just a bit of flavor y'know.

Adding more varieties. Someone on the IRC suggested a fossilized variant.
And sure, I could also add twisted shrubs and whatever but those may be more than just cosmetic.

Also less silly symbols for the new terrain, but it has to stand out in some way when there are no tiles for it yet.

#### Testing

Applied changes locally, checked for natural spawns.

#### Additional context

n/a

### DIFF
--- a/data/json/furniture_and_terrain/terrain-flora.json
+++ b/data/json/furniture_and_terrain/terrain-flora.json
@@ -488,7 +488,7 @@
     "symbol": "G",
     "color": "brown",
     "name": "warped dead tree",
-    "description": "This deciduous tree's branches are warped into a spiral.  It is pretty clear that it died due to whatever caused this."
+    "description": "This deciduous tree's branches are warped into a spiral.  This appears to have killed it."
   },
   {
     "type": "terrain",
@@ -1139,7 +1139,7 @@
     "looks_like": "t_tree_deadpine",
     "symbol": "Y",
     "color": "brown",
-    "description": "The branches of this dead pine have been twisted into a shape that remind you of Sierpinski triangles.  It is pretty clear that it died due to whatever caused this."
+    "description": "The branches of this dead pine have been twisted into a shape that remind you of Sierpinski triangles.  It did not survive getting warped like this."
   },
   {
     "type": "terrain",


data/json/furniture_and_terrain/terrain-flora.json
```diff
@@ -480,6 +480,16 @@
...
   },
+  {
+    "type": "terrain",
+    "id": "t_tree_dead_warped",
+    "copy-from": "t_tree_dead",
+    "looks_like": "t_tree_dead",
+    "symbol": "G",
+    "color": "brown",
+    "name": "warped dead tree",
+    "description": "This deciduous tree's branches are warped into a spiral.  It is pretty clear that it died due to whatever caused this."
```
Fris0uman, line 491:
```suggestion
    "description": "This deciduous tree's branches are warped into a spiral.  This appears to have killed it."
```
Or something like this, the other formulation feels a bit too convoluted

data/json/furniture_and_terrain/terrain-flora.json
```diff
@@ -1121,6 +1131,16 @@
...
   },
+  {
+    "type": "terrain",
+    "id": "t_tree_deadpine_warped",
+    "name": "warped dead pine tree",
+    "copy-from": "t_tree_deadpine",
+    "looks_like": "t_tree_deadpine",
+    "symbol": "Y",
+    "color": "brown",
+    "description": "The branches of this dead pine have been twisted into a shape that remind you of Sierpinski triangles.  It is pretty clear that it died due to whatever caused this."
```
Fris0uman, line 1142:
```suggestion
    "description": "The branches of this dead pine have been twisted into a shape that remind you of Sierpinski triangles.  It did not survive getting warped like this."
```
Same, I think the formulation could be a bit more straight forward

PR #60952, opened 2022-09-12
Rebuild active_item cache for vehicles after refresh

### LINKED ISSUE #58824: (cool) Food stacks [E]aten from a fridge gets placed in player inventory and immediately spoils
### Describe the bug

When eating certain foods (it doesn't always happen) that are being cooled in a fridge, if the item is in a stack the item is taken out of the fridge and placed in the players inventory after the first item is eaten. Thankfully this first item doesn't get immediately Rotten, but (especially after leaving the Eat menu and going back in) the foodstuff in the players inventory is suddenly rotten.

### Steps to reproduce

1. Have a fridge with cool food stacks (past their "expiration" date?)
2. Attempt to eat them
3. See them placed in player inventory
4. (Possibly): Leave Eat menu, go back in to Eat menu
5. Whoops it's all rotten suddenly, guess you shouldn't have taken it all out of the fridge, dumbass

### Expected behavior

I expect my character to be smart enough not to dump this big stack of tomatoes or meat in their inventory if they're gonna just eat one.

Failing that, I expect a stack of food that's maintained chill-freshness to not immediately, 1 second later, spoil after being taken out of the fridge - a couple of minutes at least before all the bugs descend on it, no?

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.22000.739 (21H2)
- Game Version: cc52f60 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
    Generic Guns [generic_guns],
]

### Additional context

Fridges as a whole seem to behave weirdly, and I can understand if food being good for weeks past their spoil time, almost being in stasis-lock, is not exactly desired. That being said, it's a little silly if you can ruin entire stacks of food by trying to eat one of them, when a sensible person (given the information of "hey if I take ALL this stuff out of this weird refrigerator, it'll immediately go bad") would just... not take the entire stack out of their Magically Preserving Fridge.

I'm also not entirely sure what the cirumstances are where the food is taken out. Food in a container will remain in the fridge. A stack of crops I had also remained in the fridge. But I know it happened with an item of curry. In my current save, I can't actually replicate this behaviour, so I'm not sure what's going on or in what situations the character will grab the whole stack.

robob27:
#### Summary
Bugfixes "Some items taken out of fridges rot immediately"

#### Purpose of change
Fixes #58824

#### Describe the solution
After we modified `vehicle::refresh()` to add/remove fake parts, it seems to have started causing issues with the `active_item` cache for vehicles. More information in [my comment here](https://github.com/CleverRaven/Cataclysm-DDA/issues/58824#issuecomment-1244238144). TL;DR, `vehicle::refresh` caused `active_items` to become empty, which caused items inside the vehicle not to be processed for rot etc.

It seems we need to refresh the `active_item` cache as part of `vehicle::refresh` now based on this. So I made the following changes:

- Add a new method `vehicle::refresh_active_item_cache` to encapsulate the logic for setting up `active_items`
- Call the new `vehicle::refresh_active_item_cache` method at the end of the `vehicle::refresh` method
- Stop setting up `active_item` cache directly in `vehicle::deserialize` - this method calls `vehicle::refresh` which handles this now by calling `vehicle::refresh_active_item_cache`.

- Maybe adding fake parts to vehicles isn't supposed to impact `active_item` cache and this isn't the right solution. At the very least I hope my naive attempt at fixing this might serve to push someone with a bit more context towards the right solution.

#### Testing
Using [the debug stuff described here](https://github.com/CleverRaven/Cataclysm-DDA/issues/58824#issuecomment-1243086802) I verified that after applying the changes, items in fridges were no longer frozen in time after a save/load, and had rot values updated as expected while waiting nearby.

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -5778,6 +5778,20 @@ void vehicle::enable_refresh()
     refresh();
 }
 
+void vehicle::refresh_active_item_cache()
+{
+    // Need to manually backfill the active item cache since the part loader can't call its vehicle.
+    for( const vpart_reference &vp : get_any_parts( VPFLAG_CARGO ) ) {
+        auto it = vp.part().items.begin();
+        auto end = vp.part().items.end();
+        for( ; it != end; ++it ) {
+            if( it->needs_processing() ) {
+                active_items.add( *it, vp.mount() );
+            }
+        }
+    }
+}
+
 /**
  * Refreshes all caches and refinds all parts. Used after the vehicle has had a part added or removed.
  * Makes indices of different part types so they're easy to find. Also calculates power drain.
@@ -6064,6 +6078,7 @@ void vehicle::refresh( const bool remove_fakes )
     zones_dirty = true;
     invalidate_mass();
     occupied_cache_pos = { -1, -1, -1 };
+    refresh_active_item_cache();
 }
 
 vpart_edge_info vehicle::get_edge_info( const point &mount ) const
--- a/src/savegame_json.cpp
+++ b/src/savegame_json.cpp
@@ -3473,9 +3473,6 @@ void vehicle::deserialize( const JsonObject &data )
         auto it = vp.part().items.begin();
         auto end = vp.part().items.end();
         for( ; it != end; ++it ) {
-            if( it->needs_processing() ) {
-                active_items.add( *it, vp.mount() );
-            }
             // remove after 0.F
             if( savegame_loading_version < 33 ) {
                 migrate_item_charges( *it );
--- a/src/vehicle.h
+++ b/src/vehicle.h
@@ -837,6 +837,9 @@ class vehicle
         //Refresh all caches and re-locate all parts
         void refresh( bool remove_fakes = true );
 
+        // Refresh active_item cache for vehicle parts
+        void refresh_active_item_cache();
+
         /**
          * Set stat for part constrained by range [0,durability]
          * @note does not invoke base @ref item::on_damage callback


src/vehicle.cpp
```diff
@@ -5778,6 +5778,20 @@ void vehicle::enable_refresh()
     refresh();
 }
 
+void vehicle::refresh_active_item_cache()
+{
+    // Need to manually backfill the active item cache since the part loader can't call its vehicle.
+    for( const vpart_reference &vp : get_any_parts( VPFLAG_CARGO ) ) {
```
dseguin, line 5784: It looks like this can added as a part of the loop in `vehicle::refresh` instead of a separate loop, something like this:

```diff
diff --git a/src/vehicle.cpp b/src/vehicle.cpp
index ac704ba924..14be259663 100644
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -5778,20 +5778,6 @@ void vehicle::enable_refresh()
     refresh();
 }
 
-void vehicle::refresh_active_item_cache()
-{
-    // Need to manually backfill the active item cache since the part loader can't call its vehicle.
-    for( const vpart_reference &vp : get_any_parts( VPFLAG_CARGO ) ) {
-        auto it = vp.part().items.begin();
-        auto end = vp.part().items.end();
-        for( ; it != end; ++it ) {
-            if( it->needs_processing() ) {
-                active_items.add( *it, vp.mount() );
-            }
-        }
-    }
-}
-
 /**
  * Refreshes all caches and refinds all parts. Used after the vehicle has had a part added or removed.
  * Makes indices of different part types so they're easy to find. Also calculates power drain.
@@ -5980,6 +5966,14 @@ void vehicle::refresh( const bool remove_fakes )
         if( vpi.has_flag( VPFLAG_ENABLED_DRAINS_EPOWER ) ) {
             accessories.push_back( p );
         }
+        if( vpi.has_flag( VPFLAG_CARGO ) ) {
+            // Need to manually backfill the active item cache since the part loader can't call its vehicle.
+            for( auto it = vp.part().items.begin(); it != vp.part().items.end(); ++it ) {
+                if( it->needs_processing() ) {
+                    active_items.add( *it, vp.mount() );
+                }
+            }
+        }
     }
 
     rail_wheel_bounding_box.p1 = point( railwheel_xmin, railwheel_ymin );
@@ -6078,7 +6072,6 @@ void vehicle::refresh( const bool remove_fakes )
     zones_dirty = true;
     invalidate_mass();
     occupied_cache_pos = { -1, -1, -1 };
-    refresh_active_item_cache();
 }
 
 vpart_edge_info vehicle::get_edge_info( const point &mount ) const
diff --git a/src/vehicle.h b/src/vehicle.h
index 60a169a8b8..a05ed1f3ef 100644
--- a/src/vehicle.h
+++ b/src/vehicle.h
@@ -837,9 +837,6 @@ class vehicle
         //Refresh all caches and re-locate all parts
         void refresh( bool remove_fakes = true );
 
-        // Refresh active_item cache for vehicle parts
-        void refresh_active_item_cache();
-
         /**
          * Set stat for part constrained by range [0,durability]
          * @note does not invoke base @ref item::on_damage callback
```

robob27: @dseguin I can try to make that change asap and see if it still works. I think it might not though since the part that was causing `active_items` to become empty was further down in the method in [this part](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/vehicle.cpp#L6032-L6035):

```
// add all the obstacles first
for( const std::pair <const point, std::vector<int>> &rp : relative_parts ) {
    add_fake_part( rp.first, "OBSTACLE" );
}
```

If you check `active_items.get_for_processing()` before that loop, it will have the expected items. Afterwards it's empty. Still not sure why though.

dseguin: That's strange, I don't see anything that would clear the cache there :thinking:
If it causes issues then feel free to ignore the suggestion.

PR #60953, opened 2022-09-12
Scrollbar dragging batch two

ZeroInternalReflection:
#### Summary
Interface "Scrollbar dragging batch two"

#### Purpose of change
Expand the scrollbar click-and-drag support implemented in a few UIs in #60006.

#### Describe the solution
Move handling of scrollbar dragging to the scrollbar() class.  Scrollbar dragging can then be implemented anywhere that uses scrollbar()/draw_scrollbar() with the following changes:
1. Have a scrollbar object that sticks around, rather than just creating it and discarding it when it's time to draw one
2. Call scrollbar.set_draggable( input_context ) to ensure the appropriate actions are defined
3. When handling input, call scrollbar.handle_dragging() and see if the user dragged the scrollbar,
4. If dragging the scrollbar would change the selected item in a list then parse the scrollbar position back to the new selection (this gets a bit fiddly, and I don't think it's the best long-term solution, but it works for now.  See discussion of alternatives)

Implement these changes for:
- uilist menus (should work in a lot of different places)
- draw_item_info() (such as examining items in the inventory/crafting screen)
- Character creation tabs (Scenario list, trait list, etc.)
- Martial arts information screen

 - For discrete entry lists, such as the list of scenarios in character creation, I think I'd prefer it if the selected entry and the list position were disconnected (e.g. you've got "Evacuee" selected at the top, but you've dragged the scrollbar down to the bottom to look at the expensive scenarios.  Evacuee stays selected until you do something else).  However, to do that properly, I'd need to add mouse-click selection to the various character creation lists, which I'm not quite ready to do.
 - Converting the martial arts information screen to scrolling_text_view
Eventually
- Handling scrollbar dragging on a per-UI bases
Would result in a staggering amount of code duplication

#### Testing
Tested in English at a variety of screen sizes (generally smaller ones so there were more scrollbars) in Linux.

Character creation UIs:

https://user-images.githubusercontent.com/89038572/189771382-7825e90b-94d5-44df-bc26-dd5f4344a568.mp4

Draw_item_info UI:

https://user-images.githubusercontent.com/89038572/189771413-f94d2f1c-1768-43a9-b24b-ff8e92051caf.mp4

uilist and martial arts information screen:

https://user-images.githubusercontent.com/89038572/189771443-9aa3b506-8d68-477b-be3e-7ca7bb8bb66f.mp4

Testing in curses on gnome-terminal (it mostly works, but isn't as nice):

https://user-images.githubusercontent.com/89038572/189771557-c6ca6656-4bb8-41ab-967b-821832c8fae7.mp4

In a subsequent commit (2022-09-14), I updated the dragging code to actually interpolate properly (rather than the tried-and-tested method of "Eh, looks about right").  Having the mouse cursor over the `^` should now have the scrolling window at the very top, and having the mouse cursor over the `V` should now have the scrolling window at the very bottom. (Compare to some of the videos above, where it was occasionally necessary to move the mouse _beyond_ the window in order to scroll all of the way)

https://user-images.githubusercontent.com/89038572/190187363-6cd4eba9-751b-4854-8273-591953a8c284.mp4

#### Additional context
If I didn't screw up in this implementation, there are a whole lot more places to add it to.

### DIFF
--- a/src/output.cpp
+++ b/src/output.cpp
@@ -1780,24 +1780,21 @@ bool scrollbar::handle_dragging( const std::string &action, const cata::optional
     if( ( action != "MOUSE_MOVE" && action != "CLICK_AND_DRAG" ) && dragging ) {
         // Stopped dragging the scrollbar
         dragging = false;
-        if( action == "SELECT" ) {
-            // We don't want to accidentally select something on mouse-up after dragging the scrollbar, so tell
-            // the UI that we've handled it
-            return true;
-        } else {
-            // The user is doing something else while dragging the scrollbar, don't get in their way
-            return false;
-        }
+
+        // We don't want to accidentally select something on mouse-up after dragging the scrollbar, so if
+        // there's a mouse-up event, tell the UI that we've handled it
+        return action == "SELECT";
     } else  if( action == "CLICK_AND_DRAG" && coord.has_value() &&
                 scrollbar_area.contains( coord.value() ) ) {
         // Started dragging the scrollbar
         dragging = true;
         return true;
     } else if( action == "MOUSE_MOVE" && coord.has_value() && dragging ) {
-        // Currently dragging the scrollbar
-        int y_position = ( coord->y - scrollbar_area.p_min.y ) * ( content_size_v ) /
-                         scrollbar_area.p_max.y;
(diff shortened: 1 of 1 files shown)


leemuar:
awesome work!

PR #60969, opened 2022-09-13
Made all cat ears fancy clothing

Nikole-Scheutz:
#### Summary
Bugfixes "Changed cat ears to be fancy clothing"

#### Purpose of change

Because cat ears have the description "It does nothing, but there's no reason not to look good even if no one's looking." they should count as fancy clothing

#### Describe the solution

Added FANCY flag to "Fur cat ears" and "Leather cat ears", which also through "copy-from" carries over to the "Faux fur cat ears"

#### Testing

Copied and pasted the modified json file into my game folder and checked if the "Faux fur/Leather/Fur cat ears" had the "This item is fancy" tag, which they did

#### Additional context

This is my first ever PR, so please do tell me if I did something wrong, or if there's something I should consider for the future, as I would love to help out more in the future!

### DIFF
--- a/data/json/items/armor/jewelry.json
+++ b/data/json/items/armor/jewelry.json
@@ -1553,7 +1553,7 @@
     "material": [ "fur", "plastic" ],
     "symbol": "^",
     "color": "brown",
-    "flags": [ "NO_WEAR_EFFECT" ]
+    "flags": [ "FANCY", "NO_WEAR_EFFECT" ]
   },
   {
     "id": "faux_fur_cat_ears",
@@ -1623,7 +1623,7 @@
     "material": [ "leather", "plastic" ],
     "symbol": "^",
     "color": "dark_gray",
-    "flags": [ "NO_WEAR_EFFECT" ]
+    "flags": [ "FANCY", "NO_WEAR_EFFECT" ]
   },
   {
     "id": "leather_cat_tail",


Maleclypse:
I hate to shootdown someone's first PR but FANCY is so subjective that it's continually on the verge of being removed from the game.  As I understand it Kevin only allows fancy for black tie appropriate clothing.

Termineitor244:
Given what you want to do I would suggest just adding a similar flag and trait, say "COSPLAY" that gives morale bonuses to a character with the "COSPLAY_FAN" trait, and just putting that flag with cat ears and other cosplay items, you can even add it as a mod.

Nikole-Scheutz:
@Maleclypse you don't have to feel bad, I totally understand Kevin's reasoning! I just looked at (both open and closed) issues and PR, saw nothing for cat ears, and took a shot in the dark ^^

@Termineitor244 that's a really good suggestion! I have no idea how modding this in would work, and I have never used C++, but I'm assuming I could basically look at how the morale boost for fancy clothing is implemented, copy paste, change the phrasings/words from "Fancy" to "Cosplay", and then do some playtesting? (For which I would have to compile the game myself, cuz I'd have to change the game code? Sorry for question dump, heh)

Termineitor244:
@Nikole-Scheutz you can look at #3059 for guidance of where to look at for adding a clothing based morale bonus, the linked PR is from when fancy was added to the game (Which, fun fact, was added as a replacement of a trait with similar functionality, the cross dressing trait!).

But in essence, yes, you would only need to do a copy and paste of the fancy code (and compile, sadly).

Fris0uman:
As explain above this is a no. Also the idea of a Cosplayer trai is very likely to get shut down. The stylish trait has been a tarpit of never ending about what is fancy or not I don't think we're interested in having more trait like that.

PR #60974, opened 2022-09-13
fix: nerf ambulance drug spawns

casswedson:
#### Summary
None

#### Purpose of change
do note this only prevents some item duplication bug related to spawning items with stack size or charges or something in vehicles

this is a super old bug I encountered back in the day and I guess I skipped over the other items that could be duplicated, this is fairly rare so...

I made an itemgroup with the drugs that previously spawned in ambulances as to prevent duplication, also add inhalers cause apparently ambulances also carry those around

I also gave the itemgroup a big ol' `"count": [ 1, 4 ]` this should be OK for an ambulance since I saw some comments about there being very little in ambulances

`prob` were taken from the old entries tho I presume this is a hefty buff, I really don't get more than one level of nested `prob` too many numbers

#### Additional context
Kevin did say 200 tramadol in an ambulance isn't that much but _this_ is not intended so I'm cutting it
Screenshot from 2022-09-13 16-46-32
Screenshot from 2022-09-13 16-46-51
Screenshot from 2022-09-13 16-47-14
Screenshot from 2022-09-13 17-00-35
Screenshot from 2022-09-13 17-00-50
Screenshot from 2022-09-13 17-01-44

### DIFF
--- a/data/json/itemgroups/Locations_MapExtras/locations.json
+++ b/data/json/itemgroups/Locations_MapExtras/locations.json
@@ -1753,6 +1753,26 @@
   },
   {
     "type": "item_group",
+    "id": "ambulance_drugs",
+    "subtype": "distribution",
+    "entries": [
+      { "item": "morphine", "prob": 5, "count": [ 1, 4 ] },
+      { "item": "thorazine", "prob": 5, "count": [ 1, 4 ] },
+      { "item": "tramadol", "prob": 5, "count": [ 1, 4 ] },
+      { "item": "adrenaline_injector", "prob": 5, "count": [ 1, 4 ] },
+      { "item": "oxycodone", "prob": 5, "count": [ 1, 4 ] },
+      { "item": "quikclot", "prob": 10, "count": [ 1, 4 ] },
+      { "item": "antibiotics", "prob": 12, "count": [ 1, 4 ] },
+      { "item": "disinfectant", "prob": 15, "count": [ 1, 4 ] },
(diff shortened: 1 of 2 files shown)


anothersimulacrum (commented):
Does `count` (as opposed to `charges`) work for items with charges - as in, does it spawn that many chargen or N full stacks? That's the root of the issue with it spawning 200 or so of pain pills.

casswedson:
honestly my explanation may be very misleading; I didn't dig too much into this
my conclusion was based on this:
```json
     { "x": -3, "y": 0, "chance": 5, "items": [ "tramadol" ] },
```
this results in 200 tramadol spawned, makes some sense the vehicle's item array would be interpreting this as "for every item mention spawn the default `stack_size`" in this case 200

in #44273
```json
      { "x": -4, "y": 0, "chance": 8, "items": [ "quikclot" ] },
```
results in
image
with
```json
    "stack_size": 60,
```
I would predict finding 60 "charges" of hemostatic powder, but it's maxing out the pocket's volume, hence I think there's some item duplication going on

by the way, I don't think either charges or count work within the vehicle's item array, there were quite a few repeated ids

anothersimulacrum:
I meant more your itemgroup uses count for items that have charges, which I believe would spawn that many stacks, not that many items.

casswedson:
very much intentional, in this case I use count mostly because I wanted to buff spawns a little - like a couple full painkiller bottles at every ambulance or something similar to that

PR #60975, opened 2022-09-13
Make the syringe a weapon

NetSysFire:
#### Summary
None

#### Purpose of change

A syringe is mentioned in `GAME_BALANCE.md` yet it is missing the properties. It is a teeny little detail.

#### Describe the solution

Adding some properties to the syringe.

Also looking at the scalpel. It has an effective to-hit of -3, while the syringe has -4. It looks fine though, honestly.
The pre-apocalypse price is also way off, but I think many items need an audit.

#### Testing

Applied changes locally.

image

### DIFF
--- a/data/json/items/tool/med.json
+++ b/data/json/items/tool/med.json
@@ -108,11 +108,14 @@
     "weight": "4240 mg",
     "volume": "10ml",
     "longest_side": "10 cm",
+    "cutting": 1,
+    "to_hit": { "grip": "none", "length": "hand", "surface": "point", "balance": "neutral" },
     "price": 2500,
     "price_postapoc": 10,
     "material": [ "plastic" ],
     "symbol": ",",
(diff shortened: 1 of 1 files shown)


NetSysFire:
For some reason I have to imagine a hobo syringe spear now. Just a syringe duct-taped to a stick. I could see ferals wielding it and dishing out tetanus.

Aerin-of-the-Toast:
Part of me wants to debate if a syringe is truly "neutral" balanced or if it should be "good" balance, but that's probably bikeshedding a bit too much, even for me. 😄

NetSysFire:
>if a syringe is truly "neutral" balanced or if it should be "good" balance

I have thought about it, too but a syringe is simply not intended to be used for aggressive stabbing. Seeing how many times trained nurses missed my veins, I would argue it might even be a little off balance.

Issue #59430, opened 2022-07-20
Inappropriate Quests from Evac Shelter NPCS


ANickelN:
### Describe the bug

Evac Shelter starting NPCs seem to give out the incorrect quests, often talking as if they were a member of the Refugee Center.

### Steps to reproduce

Start a new EVAC SHELTER start and talk to the starting NPC, often they will give you a quest that will in some way help the refugee center.

### Expected behavior

Evac shelter survivors should give out the appropriate quests, and not quests that have to do with the refugee center they are not a part of and aren't aware of.

### Screenshots

![image](https://user-images.githubusercontent.com/52714184/179876315-8cada53c-fcc3-4583-bcb5-c455ad00ae2d.png)
![image](https://user-images.githubusercontent.com/52714184/179877203-0dcf4466-9bc7-4cbf-bcf4-9369896eb61c.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19043.1826 (21H1)
- Game Version: b9db130 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

ReynTheLord:
![image](https://user-images.githubusercontent.com/37010048/179887892-c4c2dc22-2447-4784-ba41-d9091fc4a224.png)
potentially relevant

ANickelN:
>potentially relevant

Seems like the same issue described above, what version was that taken on?

DamienRoyan:
![image](https://user-images.githubusercontent.com/88375697/180185729-77a636e7-73a2-48dd-90bc-a78b5f8b9395.png)
Can confirm. Have never seen this in Refugee Center, but this is definitely not an Evac Shelter NPC mission.

![image](https://user-images.githubusercontent.com/88375697/180186337-6cf80b86-817b-401d-99be-4fd0e8c00c05.png)
I remember this is a Homeless camp mission


![image](https://user-images.githubusercontent.com/88375697/180186741-be70a9c4-8ebd-4104-bd73-2e475e68bc9b.png)
However it seems the normal missions aren't completely gone, because I have this. So it seems the game is pulling missions from the whole roster


- OS: Windows
    - OS Version: MINGW/CYGWIN/MSYS2 on unknown Windows version
[...]

PR #60981, opened 2022-09-14
More mission origin fixes

Termineitor244:
#### Summary
Bugfixes "Fixes the Norse Evangelicals missions so they no longer trigger from generic npcs."

#### Purpose of change
Follow-up on #60407
Contributes towards fixing #59430 with more faction specific missions becoming, well, faction specific, this PR changes the new missions of the Norse Evangelicals so they no longer trigger from any random NPC.

#### Describe the solution
Changing the origin entry of these missions to a restricted origin, "secondary".

To search for more missions like these.

#### Testing
None.

#### Additional context
I found this bug while trying to return to CDDA after work occupied all my time, I first thought the norse mission was some kind of new random mission for NPCs, but after searching around I discovered their missions are supposed to come from a specific NPC at a specific faction...

**The mission origins need proper documentation, any help with this would be appreciated.**

### DIFF
--- a/data/json/npcs/valhalla_cult/missions.json
+++ b/data/json/npcs/valhalla_cult/missions.json
@@ -18,7 +18,7 @@
         { "u_add_var": "u_joined_valhallists", "type": "general", "context": "meeting", "value": "yes" }
       ]
     },
-    "origins": [ "ORIGIN_OPENER_NPC" ],
+    "origins": [ "ORIGIN_SECONDARY" ],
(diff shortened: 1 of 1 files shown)


Inglonias:
Does this close #59430, and if not, what still needs to be done?

Maleclypse:
> Does this close #59430, and if not, what still needs to be done?

Good catch. These two plus the documentation fix should close it.

PR #60984, opened 2022-09-14
Expanded descriptions for rail mounted mandible guard

### LINKED ISSUE #60086: Method to know that HELMET_MANDIBLE_GUARD_STRAPPED and HELMET_MANDIBLE_GUARD are different in game

For the helmet attachments, specifically, mandible guards, there are two types.

As it stands, other than the item itself saying it fits in hard hats and helmets, there is no indication that it doesn't fit in helmets (military).
image

### Solution you would like.

They weren't merged into one in the original PRs, but I couldn't find a reason why?
#58749
#58020

If there's a reason not to, could there be a way to say that they aren't compatible, rather than having to search *.json?

_No response_

### Additional context

Example of two incompatible mandible items:

Plastic Chin Guard:
image
Tactical Helmet:
image

Termineitor244:
#### Summary
Bugfixes "Expands some descriptions to make more clear that some mandible guard are rail mounted."

#### Purpose of change
Fixes #60086 (I hope).

#### Describe the solution
Some simple text additions.

#### Testing
None.

### DIFF
--- a/data/json/items/armor/helmets.json
+++ b/data/json/items/armor/helmets.json
@@ -150,7 +150,7 @@
     "type": "ARMOR",
     "category": "armor",
     "name": { "str": "tactical helmet" },
-    "description": "A lightweight black helmet that provides excellent protection from all sorts of damage.",
+    "description": "A lightweight black helmet that provides excellent protection from all sorts of damage.  It has a military rail system for attachments.",
     "weight": "1180 g",
     "volume": "1850 ml",
     "price": 20000,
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):

Termineitor244:
> I saw this and wanted to to add a rail mounted flashlight like https://www.amazon.com/Tactical-Flashlight-Mounted-Compatible-MICH2000/dp/B09CNMMYDQ/ref=asc_df_B09CNMMYDQ/?tag=hyprod-20&linkCode=df0&hvadid=532845793812&hvpos=&hvnetw=g&hvrand=13742622480565808309&hvpone=&hvptwo=&hvqmt=&hvdev=c&hvdvcmdl=&hvlocint=&hvlocphy=9013478&hvtargid=pla-1460923253252&psc=1 but I don't think I have it in me right now.

I think I saw one of those when I was making the original PR, but decided against because its function is already totally occupied by the strapped flashlight and just having a flashlight in your inventory... But I don't know, I could end up adding it when I revisit some values of the attachments.

Issue #58961, opened 2022-07-04
Body temperature calculations are weird and probably a bit broken


Hirmuolio:
### Is your feature request related to a problem? Please describe.

Body temperature calculations uses three different temperature units:
* C (usually 0.01 C).
* F (usually 0.01 F).
* u (weird custom temperature unit. May be fahrenheit/50 or arbitrary values).

The math for body temperature mixes these three units seemingly without unit conversions.

For example on this line: https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/character_body.cpp#L500

`set_part_temp_conv( bp, BODYTEMP_NORM + adjusted_temp + windchill * 100 + clothing_warmth_adjustment + bp_temp_min );`

where:
* `BODYTEMP_NORM` is 5000 (u)
* `adjusted_temp` is 100 * ( environment temperature (C) - 31 (C) )
* windchill is in fahrenheit.
* clothing_warmth_adjustment is in u.
* bp_temp_min is always zero (only used by "debug tail" mutation).

Summing temperatures of different units together like this is nonsense. It can't be right.

Also variables are all done with integers for some reason. Extra effort is made to keep everything in integers. For example divisions will do cast to float and then cast back to integers. Occasionally doubles are used because we really need full 64 bit precision for this number that will be rounded to integer.

For example this beuty is in the code:
```
double total_windpower = get_local_windpower(...
int bp_windpower = total_windpower;
bp_windpower = static_cast<int>( static_cast<float>( bp_windpower ) * ... );
```
The for loop also has math in it that does not depend on looped variable (ie. same for all body parts). They could be just moved out of the for loop.

### Solution you would like.

All the temperature should be in same units (preferably units::temperature).
Refactoring is very hard since it is not clear what units are wrong and how wrong they are.
Refactor to more consistently use float when appropriate.
Refactor some more.

### Describe alternatives you have considered.

Nuke it and rewrite from scratch.

Venera3:
For context, the `bp_min` bit is in preparation for Limb Stuff with different limbs having different levels of base warmth and warmth compensation, but that Project won't happen this cycle.

bombasticSlacks:
u is defined in weather.h 

```c++
/**
 * @name BODYTEMP
 * Body temperature.
 * Body temperature is measured on a scale of 0u to 10000u, where 10u = 0.02C and 5000u is 37C
[...]

Issue #53411, opened 2021-12-13
Excessive "Your shivering prevents you from sleeping" messages.


NetSysFire:
### Describe the bug

When trying to sleep when you are too cold, there are waaaaaay too many messages. I am talking about about ~1000 per try.

### Steps to reproduce

1. Have a naked character on spring, day 61 at night.
2. Try to sleep in a shallow pit.
3. Witness the excessive messages.

![image](https://user-images.githubusercontent.com/59517351/145762784-ac2ffeeb-71f1-4bb1-af33-a21b0f9bc5f1.png)

### Expected behavior

This should be ratelimited.

### Versions and configuration

- experimental

Fris0uman:
#61708 should make it so you only get a message once per hour at most

PR #60991, opened 2022-09-14
Fix shivering mechanism

### LINKED ISSUE #47671: Shivering Mechanics
# Is your feature request related to a problem? Please describe.
Currently, if you're cold and attempt to go to bed, you will not fall asleep due to shivering even though you may have a blanket on the same place you're attempting to sleep in.

# Describe the solution you'd like
It would be divine if the game could assume you are "wearing" the blanket for the warmth bonuses in the case of sleeping. Otherwise, you have to actually wear the blanket, just to go to sleep. (Note: The blanket DOES apply it's warmth while your asleep, "You use your blanket to keep warm" so I wonder why can't it apply that while attempting to sleep.

# Describe alternatives you've considered
The blanket could simply add warmth to your body for being on the same space as it *and* attempting to sleep. I suppose it could add warmth to you even when spending time there and not sleeping. Or maybe apply the "you use your blanket to keep warm" effect could just apply while in the same place as a blanket.

# Additional context
It's not like wearing the blanket and then taking it off is the worst thing in the world, but attempting to sleep then shivering awake, then having to wear the blanket your sleeping with, then sleep, then wake up and take off the blanket, then drop it, is just repetitive.

### LINKED ISSUE #53443: Shivering mechanics are inconsistent
### Describe the bug

My character, which is almost naked except for a straw hat, can not sleep due to shivering, yet the temperature in the panel says comfortable. The torso temp is -21, which is still in the green, quite literally. The area temperature is 14°C.

### Steps to reproduce

This appears to be the part of the code causing wake-ups and many messages (see also #53411)

https://github.com/CleverRaven/Cataclysm-DDA/blob/611e3ea7898cd0eba8b1ff7d49e286de870b4524/src/character_body.cpp#L615-L617

Yet the code for the widget displays the most extreme part, which is still alright temperature-wise in my case:

https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/panels.cpp#L723-L725

### Expected behavior

No or only mild shivering. It is 14°C and it causes sleep deprivation because my character can not sleep due to the shivering.

The shivering mechanics only care about the torso and my character is not taking damage, so sleeping should not be prevented, at least not immediately and consistently. An occasional message about you feeling cold and shivering mildly would be appropriate.

As far as I know this was implemented so folks did not die in their sleep because of hypothermia which would have normally woken up the character but did not, this seems excessive in this particular case though.

Potentially introduced by #39827

### Screenshots

_No response_

- experimental

### Additional context

_No response_

peter4370:
#### Summary
Bugfixes "Fix shivering mechanism"

#### Purpose of change
Fix #47671
Fix #53443
While being cold could prevent you from sleep in real life, It is unclear if the warmth in @ window is enough to prevent shivering and leads to confusion.
Also you can fall asleep while being too cold if you are tired enough in real life.

#### Describe the solution
Lowering the threshold to let player could sleep in spring.
Also make player able to sleep if they are deadly tire to prevent extreme sleep deprivation.

Make body temperature more accurate. (see #58961)
Remove this mechanism as we have another being too cold wakes you up mechanism at [player_hardcoded_effects.cpp](https://github.com/CleverRaven/Cataclysm-DDA/blob/22325d508dc3a55128efe3f12225265085adad91/src/player_hardcoded_effects.cpp#L1121)

#### Testing
Test 1:
Create a character spawn in spring
Being naked
Set fatigue to tired
Being able to sleep without waking up
Test 2:
The same as above but in winter
Wake up due to cold after falling asleep
Test 3:
The same as test 2 but being exhausted
Ignore warnings and keep trying to sleep
Wake up from cold because the code in player_hardcoded_effects.cpp
Try sleep again and die because of too cold

#### Additional context
Some comments in update_bodytemp.cpp are incorrect. I updated them in this PR too.
I don't see #53411 happening so if someone checked it doesn't happen anymore that issue could be closed.


src/character_body.cpp
```diff
@@ -484,7 +485,8 @@ void Character::update_bodytemp()
             continue;
         }
 
-        // This adjusts the temperature scale to match the bodytemp scale,
+        // 0.01c is 5 body temp unit so it needs to times 5 to scale to match the bodytemp scale,
+        // but this will change current behavior,
```
Fris0uman, line 489: this is a very weird comment, what current behaviour? If this get merged the comment would stay and the "current behaviour" would be whatever this PR implements

src/character_body.cpp
```diff
@@ -484,7 +485,8 @@ void Character::update_bodytemp()
             continue;
         }
 
-        // This adjusts the temperature scale to match the bodytemp scale,
+        // 0.01c is 5 body temp unit so it needs to times 5 to scale to match the bodytemp scale,
```
Fris0uman, line 488: This is super confusing,  I have no idea what this comments mean X)

Zireael07: "need to be multiplied times 5" I guess. this is about the units not matching, and the problem was raised on the temperature class PR too

src/character_body.cpp
peter4370: I think I'll just revert these comments since the unit may not be right is already mentioned #58961 and the comment may be confuse.

Fris0uman:
Maybe make it so you still get a message about shivering with the same condition as before but it only wakes you up if it's really cold (and you're not dead tired)?

peter4370:
I'm busy recently and don't want to work on this PR. If anyone want to fix the issue I would happy to see that

PR #60999, opened 2022-09-14
Add an I4 diesel engine

Jarewill:
#### Summary
Content "Adds an I4 diesel engine"

#### Purpose of change
Diesel engines only come in large sizes, making them way rarer than they should be.
In real life even common cars come with I4 diesel engines, so why wouldn't they in Cataclysm?
(At least that's the case here in Europe, if it's not the case in the US then I can close this PR)

#### Describe the solution
Adds the I4 diesel engine and adds it to item groups that have the I4 combustion engine.
Also added two variants of existing vehicles: Car and Hatchback in diesel flavor.

Not adding the cars, but if I recall correctly PRs that only add items without uses for them aren't accepted.

#### Testing
Spawned the vehicles and drove a little.

#### Additional context
I am not 100% certain on values chosen, so any feedback about them will be appreciated.
However I do hope this gets merged because I need those engines for a little forklift tweaks I'm planning.
I love forklifts.

### DIFF
--- a/data/json/itemgroups/Locations_MapExtras/locations.json
+++ b/data/json/itemgroups/Locations_MapExtras/locations.json
@@ -1479,6 +1479,7 @@
       [ "v2_combustion", 10 ],
       [ "i4_combustion", 10 ],
       [ "v6_combustion", 10 ],
+      [ "i4_diesel", 10 ],
       [ "v6_diesel", 10 ],
       [ "v8_combustion", 10 ],
       [ "1cyl_combustion_small", 10 ],


Maleclypse:
GM has at least one line of vehicles that uses this engine size in America so that should work.

data/json/vehicles/cars.json
```diff
@@ -610,6 +664,56 @@
...
+      { "x": -1, "y": 0, "parts": [ "frame_cross", "seat_back_left", "seatbelt", "roof" ] },
+      { "x": -1, "y": 1, "parts": [ "frame_cross", "seat_back_right", "seatbelt", "roof" ] },
+      { "x": -1, "y": 2, "parts": [ "frame_vertical_right", "door_se" ] },
+      { "x": -2, "y": -1, "parts": [ "frame_sw", "wheel_mount_medium", "wheel" ] },
+      { "x": -2, "y": -1, "part": "tank_medium", "fuel": "diesel" },
+      { "x": -2, "y": -1, "part": "halfboard_hatch_wheel_left" },
+      { "x": -2, "y": 0, "parts": [ "frame_horizontal_rear", "hatch", "muffler" ] },
+      { "x": -2, "y": 1, "parts": [ "frame_horizontal_rear", "hatch" ] },
+      { "x": -2, "y": 2, "parts": [ "frame_se", "wheel_mount_medium", "wheel" ] },
+      { "x": -2, "y": 2, "part": "tank_medium", "fuel": "gasoline" },
```
irwiss, line 705: Seems this needs a correction

Jarewill: Oh I didn't even notice that car had two tanks, my bad.
I'll PR some fixes in a bit.

PR #61012, opened 2022-09-15
Force electronics controls to appear in appliance menu

irwiss:
#### Summary
None

#### Purpose of change

#60868 did some changes to menus, this fixes one of the side effects

#### Describe the solution

Electronics menu needs to be force enabled for appliances

#### Testing

Place ASRG, see no option to enable reactor
Apply patch, repeat, should be able to enable reactor

### DIFF
--- a/src/vehicle_use.cpp
+++ b/src/vehicle_use.cpp
@@ -1841,7 +1841,7 @@ void vehicle::build_interact_menu( veh_menu &menu, const tripoint &p, bool with_
         .on_submit( [this] { toggle_autopilot(); } );
     }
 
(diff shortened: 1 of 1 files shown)


anoobindisguise:
Somewhere along this and the previous PR something broke, I can no longer examine the car with e while driving it. This is messing with my muscle memory pretty bad, if I want to see my fuel usage, I have to let go of the car otherwise muscle memory will shut off the car as e->enter now shuts it off instead of examining its menu (and fuel usage, damage, etc), and if I want to turn the car off, I end up putting it in idle by accident since e->downarrow->enter now puts it into idle instead of turning it off.

PR #61013, opened 2022-09-15
Add Standard Combat Test mod for better monster balance testing

Venera3:
#### Summary
Mods "Add standard combat test mod"

#### Purpose of change

Eyeballing monster balance can be hard, and playtesting tends to be very variable between contributors when it comes to test character loadouts and such. This adds some baseline characters to help bring everybody onto the same page.

#### Describe the solution

Add the `Standard combat test suite` misc mod, consisting of (for now) three scenarios with one profession each:
- Evacuee, spawning a set of improvised weapons (makeshift machete, pipe mace, knife spear, and a Glock 17) and allowing the `Day1` profession wearing the standard Evacuee loadout and no skills
- Scavenger, spawning a set of lootable/non-forged weapons ( Fire axe, ~~Shigell~~ ~~Shilleg~~ bonky stick, combat knife, and an AR-15) and allowing the `Midgame` profession with 4 in all skills and wearing a gambeson, turnout pants, combat boots, a ballsitic vest with plates and tactical helmet/gloves
- Survivor, spawning high-end weapons (nodachi, broadsword, mace, lucerne hammer, rapier, pike, Barrett), and allowing the `Lategame` profession with 8 in all combat skills and wearing nomad plate, Hub greaves/braces, a nomad bodyglove, combat boots and high steel helm/gauntlets.

#### Testing

Loads, gear and armor spawns as intended.

#### Additional context

I haven't really kept up with the armor meta lately, so if anything seems off on the loadouts let me know.
If this gets established we can start asserting rules like "a midgame enemy beats a day1 toon 10/10 times" or whatever, and the melee portion of the test would probably be doable as a unit test if somebody codey has the inclination for that.

### DIFF
--- a/data/mods/Standard_Combat_Tests/modinfo.json
+++ b/data/mods/Standard_Combat_Tests/modinfo.json
@@ -31,24 +31,26 @@
     "id": "scavenger",
     "name": "Scavenger",
     "points": 0,
-    "description": "You are getting by in the Apocalypse.  Your gear is still scavenged from the ruins of the old world, but you have gathered the skills to use it to good effect.",
+    "description": "You are getting by in the Apocalypse.  Your gear is still scavenged from the ruins of the old world, but you have gathered the skills to use it to good effect.  Midsummer start.",
     "allowed_locs": [ "sloc_shelter_a" ],
     "start_name": "Evac Shelter",
     "professions": [ "sct_mid" ],
     "map_extra": "mx_sct_mid",
-    "flags": [ "CITY_START", "LONE_START" ]
+    "flags": [ "CITY_START", "LONE_START" ],
+    "custom_initial_date": { "hour": 8, "day": 45, "season": "summer" }
   },
   {
     "type": "scenario",
     "id": "survivor",
     "name": "Survivor",
     "points": 0,
-    "description": "You are as close to thriving as anybody can be these days.  Your gear is handcrafted with the threats of the new world in mind, and your skills are honed to perfection.",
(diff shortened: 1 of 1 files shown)


GuardianDll:
Is it possible to use it in autotests?

Venera3:
I could see something built on it as an on-demand test like the monfaction matrix, but I'm not even going to attempt hooking it up. I don't see these as an automated tests unless we're talking very abstracted cases ("a zombie kills an unarmed survivor") to check if any melee or other changes break them.

Venera3:
To fix the basic build failure allowed scenario blacklists to overwrite each other, leaving a debug.log message. Tested with Innawoods + Backrooms locally, worked without a hitch.

Fris0uman:
Will the error still show for players? Because if it makes the error silent I think it's a problem

Venera3:
The problem with using a red debugmsg for it is *any* of those triggering will fail the test automatically. The debuglog felt like an acceptable compromise in that regard.

Fris0uman:
But no one looks at the log, so without the errors we're missing the thing that tells players they messed up.

Venera3:
Basically the options I see are:

- Disable every mod using a scenario blacklist from the `all-mods` test run: pretty sure unacceptable since that will mean no total conversion mod gets tested
- Don't use a scenario blacklist in this, but the entire point of this exercise is to go "use these loadouts", not "aight, first choose Scenario X, Y or Z"
- Make some tests ignore debug messages: even if I dared touch that code "treating result as failure because of errors logged on load" exists for a reason, like catching JSON validation errors that don't get explicitly tested
- Adding an easily-accessible log to it, so that *if* somebody loads up Innawoods before Backrooms and wonders why they can't start in the wilderness and opens a bug report it's easily solvable what the problem is

PR #61021, opened 2022-09-15
flu shot chang

sonphantrung:
#### Summary
Content "flu shot change"

#### Purpose of change
Currently, flu shots in CC ONLY FROGS YOU, which is unfun

#### Describe the solution
Minor edits to flu shot in CC:
- I split the flu shot into two types: the frog and the normal

NOOOOOOOOOOOOOOOOOOOOOOOOOOOOOONE

#### Testing
Ehe, should be good


data/mods/CrazyCataclysm/crazy_items.json
```diff
@@ -304,7 +304,7 @@
     "flags": [ "FIRE", "LIGHT_310", "CHARGEDIM", "FLAMING", "DURABLE_MELEE", "TRADER_AVOID", "WATER_EXTINGUISH" ]
   },
   {
-    "id": "flu_shot",
+    "id": "flu_shot_fun",
```
Fris0uman, line 307: changing an id will break saves

Fris0uman:
You made new item groups but didn't change the effect of the shot

data/mods/CrazyCataclysm/crazy_items.json
sonphantrung: Hmm, If I do the migration, will all of the `flu_shot` be migrated to `flu_shot_fun`?

Fris0uman:
What do you mean? Please don't make pointless Pull Requests

sonphantrung:
Oh, I mean, the naming is a joke, but my intention is legit: I made the flu shot in CC split into two types: The normal and the frog one. Rewriting the desc (I guess I'm bad at jokes).

data/mods/CrazyCataclysm/crazy_items.json
kevingranade: In this case it won't because the id still exists, all of the existing "fun" shots will just "swap gender" (frog joke) to the old version.

PR #61027, opened 2022-09-15
Added bundled javelin recipe.

mamiral:
#### Summary
Content "Added bundled javelin crafting recipe"

#### Purpose of change

Added a bundled javelin crafting recipe. It is quite problematic to move "large" quantities of wooden javelins around without some kind of a javelin friendly container aka. javelin bag, spear strap or golf bag. Since the javelin is crafted from a long stick and long sticks can be bundled together - javelins should also be bundled together easily. Disassemble to use...

#### Describe the solution

Added a new recipe to JSON files. Bundle volume calculated as: 1 L (javelin vol.) * 5 + 2 L (for good measure) = 7 L. Bundle weight calculated as: 600 g (javelin weight) * 5 = 3 kg. Longest side: 140 cm is taken from javelin recipe. Bundle price/barer value: 5 * javelin price/barter value = $200/$12.50

#### Testing

Tested by playing a semi-recent build with the changes. Doesn't make the javelin OP. ;)

### DIFF
--- a/data/json/items/generic.json
+++ b/data/json/items/generic.json
@@ -559,8 +559,8 @@
     "volume": "7 L",
     "weight": "3 kg",
     "longest_side": "140 cm",
-    "price": 0,
-    "price_postapoc": 50,
+    "price": 20000,
+    "price_postapoc": 1250,
     "material": [ "wood" ],
     "symbol": "H",
     "color": "brown"


MNG-cataclysm:
What if we could use the javelin bundle as a thrown weapon, like a mega-javelin?

PR #61030, opened 2022-09-16
Redoes Ammo stats in Game_Balance

tenmillimaster:
#### Summary
None
#### Purpose of change
Previously, a linear model for changes in barrel length was used to determine damage bonuses/modifiers. This was a more pronounced problem in cartridges that were not designed to be used  in very short barrels, or the opposite.

Additionally, it was not clear what the convention was for a  cartridge's default damage - should it be 0 on the guns, and the rest in the caliber? Or a little bit on each gun and less so on the cartridges.

It was also unclear as to what calibers would not be considered as "ballistic damage" from things like baseballs to transonic, hypervelocity projectiles.

#### Describe the solution
I propose here that logarithmic regressions be used for each caliber, with deviations from the default barrel lengths to be calculated with each gun, and compared against the default damage in this document. I have set most rifles to be evaluated at 16"- the legal minimum for rifles in the United States, 18" for shotguns, 6" for anything pistol (or deemed pistol-like), 26" inches for any large caliber rifles, and 46" for any anti-tank/anti-material/smol cannons.

These regression analyses were performed on the bullet travel vs bullet velocity graphs I produced using interior ballistics modelling software.

I've also thrown in a line about what isn't "ballistic damage". How might this not-ballistic damage be evaluated? ¯\_(ツ)_/¯ Hey I just do guns man.

Several calibers that were not real were rationalized:

I did my best to emulate 4.73 caseless using calculated case dimensions. From  there, I scaled it to fit the 8x40 case dimensions and tried to select a sensible bullet  to use. Thus - 8x40 is just a battle rifle version of HK G11, as we've often posited in development circles.

I took the same approach with 20x66 - it's now just 10 gauge in a 4.73 type package.

1.23 Ln? I checked the stats, which were identical to 30-06. As I feel the Exodii would value heavier bullet weights against zombies/later evolutions, I went with 8mm-06 using an early round nose bullet, a wildcat cartridge unlikely to ever be used elsewhere in game, which would yield similar statistics to 30-06, as well as being something believable to come out of the conflict they  sourced the caliber from.

To futureproof things I also added a considerable number of cartridges that are not yet in the game, that one could conceivably find in the US.

Not spending days punching values into a glorified calculator.

#### Testing
I tried to check against common calibers but I might have made a mistake or two. Also, trying to cram this data into a predictive regression is not a flawless process, so it doesn't always neatly line up with reality.

#### Additional context

I had difficulty rationalizing 5x50, so I just didn't.  The SPIW cartridges were not shot from Hulls and weren't really 'side-arm' sized, nor were the single flechette+sabot cartridges very sensical, and the CAWS was really a lot closer to 12 ga.

Dam specifies that it is rounded to the nearest integer, but I have provided the first decimal place to reduce rounding errors in calculations of modifiers.

There are more calibers I want to do but I am nearing the limit of my attention span.


RenechCDDA:
The text says damage is represented by `Dmg` but the chart uses `Dam`.

I might be missing something, but this does not seem to provide an actual example of a non-standard barrel length modifier? If we use your formula for 12 gauge 2.75in, a 3-inch barrel shotgun pistol (e.g. the existing "12 gauge pistol" with a mere -5 modifier) should do ~26.28 damage, or a -25 modifier. If this is correct (and it immediately raises some eyebrows on my end), it would be best demonstrated by use of an in-line example.

tenmillimaster:
> The text says damage is represented by `Dmg` but the chart uses `Dam`.
> 
> I might be missing something, but this does not seem to provide an actual example of a non-standard barrel length modifier? If we use your formula for 12 gauge 2.75in, a 3-inch barrel shotgun pistol (e.g. the existing "12 gauge pistol" with a mere -5 modifier) should do ~26.28 damage, or a -25 modifier. If this is correct (and it immediately raises some eyebrows on my end), it would be best demonstrated by use of an in-line example.

Approximate 2 years ago I started this work, using real world statistics. I hit a wall due to lack of people wanting to chop off their barrels and shoot bullets through the, 1 inch at at time.
One of the calibers I had modelled was 12 ga, 2.75" slugs. For the purposes of modelling, I considered these masses in flight to be slugs.
I believe, but am not certain, my source was <https://smallarmsreview.com/energy-by-the-inch-the-great-shotgun-chop-of-2007/>. The data I think I found through the same article, but a different location. It is lost to me.
image
image

Punching in the difference between a 3" barrel and an 18", there's quite a lot here too: -19.96 damage modifier.

You may compare to the results yielded from here: https://www.kommandoblog.com/2017/05/16/shotgun-barrel-length-velocity/, in the super shorty.  Shotshells, not slugs here, and a 6" barrel vs the 3 inch.

The graph I produced the regression in this draft from is this one: 
image


2 additional points:
1) You must remember that I'm doing this to try and fill in the gaps for where ballistics by the inch data doesn't exist, or pull into reality where the data cannot exist (caseless ammo). Squeezing this information into a simple-to-reproduce regression isn't going to be perfect and will result in a point or two of inaccuracy that I feel is acceptable.

2) A 3 inch barrel for a 2.75" shell leaves, what, 0.25" of barrel for the projectile(s) to pick up speed?

I can, and intend to, try to improve the performance where IRL data exists. If need be I can substitute the real world data in, but for now I have gone in the direction that the other devs have instructed me in: mostly uniform modelling.

RenechCDDA:
I do not intend to dispute that a very short shotgun pistol might do so little damage, but I'm still confused as to how you're arrived at the given values. I used the shotgun pistol a sanity check.

>Punching in the difference between a 3" barrel and an 18", there's quite a lot here too: -19.96 damage modifier.

This is what I mean, I'm not sure if it's an issue in the revised docs or just a misunderstanding on my end. (And I'm sure some contributors will be even more confused than I am if they aren't familiar with logarithms!)

If I put the given formula `( 13.794 * Ln(3) ) + 11.13` into wolfram alpha I get 26.2843, which is.... different. If that's supposed to be total damage of the 12 gauge 2.75in round in a 3-inch barrel, the damage modifier for the shotgun pistol should be -24.7157, not -19.96. How have you arrived at -19.96?

tenmillimaster:
> This is what I mean, I'm not sure if it's an issue in the revised docs or just a misunderstanding on my end. (And I'm sure some contributors will be even more confused than I am if they aren't familiar with logarithms!)
> 
> If I put the given formula `( 13.794 * Ln(3) ) + 11.13` into wolfram alpha I get 26.2843, which is.... different. If that's supposed to be total damage of the 12 gauge 2.75in round in a 3-inch barrel, the damage modifier for the shotgun pistol should be -24.7157, not -19.96. How have you arrived at -19.96?


1) By using the empirical data regressions. y = 11.008ln(x) + 19.145.

dam modifier (shotgun pistol) = (11.008ln(18) + 19.145) - (11.008ln(3) + 19.145)


2) A rounding error. I accidentally gave you (11.008ln(18) + 19.145)  -31.

tenmillimaster:
I feel comfortable with where this is now.

PR #61034, opened 2022-09-16
Pressurized tanks of ammonia should be pressurized

### LINKED ISSUE #59539: (Occasional) Batch Crafting of Rubber Cement destroys Liquid Ammonia despite adequate skill and no crafting-mistake message
### Describe the bug

Sometimes, usually when doing a batch, Rubber Cement will destroy the Liquid Ammonia in the same way as with messing up a recipe. However, my characters skills are not so inadequate that this would actually be from a mess-up, and there is no log indication of messing up.

As said, this usually happens with a batch. I noticed this the first time with 20 Rubber Cement, the second time with 2, and right after I did the 2, trying single batches also destroyed the ammonia. Other times it did not in single batches. I am not sure if I have successfully done a multiple batch of Rubber Cement without destroying the liquid ammonia.

### Steps to reproduce

I'm not entirely sure, as it doesn't reproduce all the time in my own game world, and debugged-in liquid ammonia tanks worked. 

But, uh...

1. Get a UPS Food Processor and UPS Chemistry Set, put outside of inventory.
2. Put UPS in inventory.
3. Put liquid ammonia (and other ingredients) outside of inventory.
4. Try to do a batch craft of at least 2 Rubber Cement.
5. Hopefully fuck up?

### Expected behavior

I expect that if there is a skill issue, that it would display. Otherwise, I expect my character to be able to create rubber cement without destroying the item, as with every other crafting of similar skill requirements met.

### Screenshots

First image is the crafting display, showing no skill notices. 
Second image is the log, showing no "You made a mistake!" message.
Third image is the location of the liquid ammonia tanks.
image
image
image

- OS: Windows
    - OS Version: 10.0.22000.795 (21H2)
- Game Version: b86bfea [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
]

### Additional context

Other potential caveats:
The crafting takes place on a placed furniture workbench in a Light Industry structure that is just 1 OMT north of a player-faction Basecamp.
The liquid ammonia are in tanks found from a hidden lab, and they are not on my character - they're on a display rack 4-5 tiles to my east.
The chemistry set and food processor are to my north, are UPS modded, and are running off a UPS in my inventory.
I debugged in some liquid ammonia to replace this, and both the batch and single-craft In Progress items then would work. I am not sure why this is.

RenechCDDA:
#### Summary
Balance "Lab ammonia now comes in full containers"

#### Purpose of change
See https://discord.com/channels/598523535169945603/598523535169945607/1020179885995602022

Per @Drew4484 (authored the changes adding liquid ammonia spawns), the tanks SHOULD be spawning full. Currently, they are spawning with exactly 1 charge.

Fixes #59539

#### Describe the solution
~~Why does changing the allowed weight of a tank cause this? Good question. Good question...~~ See #61035.

Pressure tank adjusted to have a defined max_contains_weight. Item spawns adjusted so that the pressure tanks come filled with liquid ammonia. 

~~I originally tried for an array of 3-6 ammonia (e.g. some containers partially used).  But this is more straightforward and doesn't require me to fight the item groups. It also avoids questions of how a 1.6L pressure tank contains less than 1.6L of ammonia.~~
If desired, it is fully possible to set ammonia to spawn in any quantity from 1 to the maximum the container may carry using `charges-min` and `charges-max` as usual, **after** this fix is applied. I could not get it to work in testing because this behavior fails when the pocket's weight is undefined.

#### Testing
Made the changes locally, generated a new world, and debugged the starting character into a microlab where I confirmed that all pressure tanks in the area spawned completely full.

### DIFF
--- a/data/json/items/tool/science.json
+++ b/data/json/items/tool/science.json
@@ -613,7 +613,7 @@
   },
   {
     "id": "pressure_tank",
-    "type": "TOOL",
+    "type": "GENERIC",
     "name": { "str": "pressurized tank" },
     "description": "This is a tank designed to hold high pressure gasses.",
     "weight": "2268 g",


RenechCDDA:
I know this PR looks very strange on the face of it, but after double checking the behavior again today I've edited the summary to remove some misunderstandings. I also pushed another commit which will correct the spawns post-fix - my summary initially claimed they would always spawn full. This wasn't true with the item groups having a `charges-min` value of 1. They now do, explicitly, spawn full (`charges-min` set to 6).

RenechCDDA:
Sorry (again, if you remember me!) to those inappropriately notified for code review. I was rebasing onto master locally and visual studio's reset option *immediately pushes without confirmation*. Very cool, thank you VS.

PR #57912, opened 2022-05-25
[CR] Draft: Mutation system refinements

Ilysen:
#### Summary
Features "Further tweaks, adjustments, and improvements to the new mutation system"

#### Purpose of change

The new mutation system is cool, but the poor thing needs some love. I wanted to try my hand.

#### Describe the solution

This PR does a bunch of things:
1. Vitamins now support a new field called `decays_into`, which is a list of pairs. The field allows vitamins to affect the values of other vitamins when they themselves are removed; for instance, vitamin X can reduce the vitamin Y by 2 units when it is reduced.
2. Standard liquid mutagen and mutagenic chelators now use unique vitamins that utilize the decay framework, each of which emulate their respective item's current effects over a short period of time.
4. Liquid mutagens are now ingested as drugs and not liquids - i.e. the code considers them more like pills than soda. This is done because, with the way stomach mechanics work, vitamin intake from liquid mutagens was very slow, and almost nonexistent; giving them `consume_drug` makes them give the vitamins immediately and thus function as intended.
5. All mutations should now properly be purifiable. Quite a few currently are not due to lacking `types` entries or specific exceptions in the dummy human traits, which means that traits like Albino or Mood Swings can never be purified.
6. Minor writing improvements - descriptions are now more clear about how the items actually work. For the sake of clarity, "typed mutagens" are now just simply primers - i.e. Alpha Mutagen is now called Alpha Primer. This is an attempt to further distinguish the two, and to give the two components of mutation (mutagen and primer) 100% distinct names.
7. Created the file `MUTATIONS.md` in the docs folder, which contains detailed information about how the new mutation system works.
8. Instability is now added through all mutations, instead of just through mutagen-induced ones. This prevents things like Genetic Chaos from being an effectively infinite source of positive traits.
9. Implements a unit test that ensures every mutation obtained through mutagen is purifiable, either by having an entry in its `types` that corresponds with a dummy mutation's types or by existing in the `cancels` list of one of those types.
10. Lots of bugfixes and improvements across the code.

#### Describe alternatives you've considered

* I'm uncertain if the vitamin decay framework is clean or effective for its intended purpose. The goal was to shrink the gigantic lists of vitamins in the item descriptions for default mutagen and mutagenic chelator, as well as to allow them and other future items to work over time. I'm happily open to other ideas regarding this.
[...]

PR #61036, opened 2022-09-16
Mutation system refinements, take two

Ilysen:
#### Summary
Features "The new mutation system has received a bunch of changes and consistency updates, with the goals of stability and bugfixing. Ask your local XEDRA scientist today!"

#### Purpose of change

So back in May of this year, I started working on a refinement PR for the mutation system (#57912). Things were going well, up until June, when I was evicted from my apartment with roughly eight hours of warning. This obviously made it kind of hard to work on the project; I had no PC access, and I was flown to stay with a relative for a while.

I'm still not back home yet, but I really want to get this finished. I figured that someone else would make some adjustments to the system after I had to drop the project, but it still seems unloved. I have a laptop that takes 30 minutes to compile the game and a dream, so let's try again. Worst case scenario, I have to drop it again and it's nobody's loss but my own. :p

#### Describe the solution

For brevity, I've copied the old PR's body here with a few changes.

This PR does a bunch of things:
1. Vitamins now support a new field called `decays_into`, which is a list of pairs. The field allows vitamins to affect the values of other vitamins when they themselves are removed; for instance, vitamin X can reduce the vitamin Y by 2 units when it is reduced.
2. Standard liquid mutagen and mutagenic chelators now use unique vitamins that utilize the decay framework, each of which emulate their respective item's current effects over a short period of time.
4. Liquid mutagens are now ingested as drugs and not liquids - i.e. the code considers them more like pills than soda. This is done because, with the way stomach mechanics work, vitamin intake from liquid mutagens was very slow, and almost nonexistent; giving them `consume_drug` makes them give the vitamins immediately and thus function as intended.
5. **All** mutations, where applicable, should now properly be purifiable. Many mutations currently cannot be purified due to lacking `types` entries or specific exceptions in the dummy human traits, which means that traits like Albino or Mood Swings can never be purified, as well as some more egregious ones like Gastropod Foot. This PR causes all mutations (with exceptions of specifically non-purifiable ones) to be valid targets for purification by adding new dummy mutations and expanding existing ones. Magiclysm mutation lines are included as well.
6. The vitamins used for type mutagens are now called primers to represent what they actually are - i.e. Alpha Mutagen (the vitamin) is called Alpha Primer. This is an attempt to further distinguish the two, and to give the two components of mutation (mutagen and primer) 100% distinct names.
7. Created the file `MUTATIONS.md` in the docs folder, which contains detailed information about how the new mutation system works.
8. Instability is now added through all mutations, instead of just through mutagen-induced ones. This prevents things like Genetic Chaos from being an effectively infinite source of positive traits.
9. Implements a unit test that ensures every mutation obtained through mutagen is purifiable, either by having an entry in its `types` that corresponds with a dummy mutation's types or by existing in the `cancels` list of one of those types.
10. Lots of bugfixes and improvements across the code.

There are some changes I've made to the timing logic for testing purposes to make things easier on myself - more specifically, mutations happen in rapid-fire bursts, and it's much faster to start mutating after you take mutagen. This is something I'd mulled over including as a feature in the original PR, but in this go-around I think it's out of scope.

#### Testing

WIP. This PR will have a lot of different effects, and I need to spend a while testing it (or maybe even doing a playthrough with it) to get a sense for how the effects are felt.

#### Additional context

To-do list in no particular order:
- [x] Hash out MUTATIONS.md
- [x] Make sure that the unit tests pass after such a long time
- [x] Testing, testing, testing

### DIFF
--- a/data/mods/Magiclysm/traits/mutation.json
+++ b/data/mods/Magiclysm/traits/mutation.json
@@ -36,7 +36,7 @@
       "MANA_SEEKER_BOLTS",
       "MANA_SUBTLE_SPELL",
       "MANA_SILENT_SPELL",
-	  "MANA_LUM",
+      "MANA_LUM",
       "OVERLY_ACID_GUT",
       "DRAGON_BREATH_BLACK",
       "DRAGON_SPELLS",


Maleclypse:
Would it make sense for Genetic Chaos/Downward spiral to multiply the amount of instability or set a minimum floor value for instability?

Ilysen:
Downward Spiral causes your instability to tank extremely fast, but Genetic Chaos doesn't currently do anything with instability. It could go either way, so I'm open to going wherever with it.

Ragecrabbo:
> Downward Spiral causes your instability to tank extremely fast, but Genetic Chaos doesn't currently do anything with instability. It could go either way, so I'm open to going wherever with it.

I feel like genetic chaos should remain somewhat ''safe'' in the sense that it shouldnt tank your instability and become a nightmare to manage for people who enjoy randomly mutating aswell as to set it apart from downward spiral. 
My suggestion would be to either have it give the player much less instability than normal (or none) for each mutation but have the mutations be always random similar to the old system or have it mutate you less frequently so that the players have a chance to keep their instability levels normal by scavenging for instability reducing items.

Ilysen:
I'm debating how to handle the genetic chaos stuff; I might just not touch it in this PR and leave it open to adjustment in the future, so as to avoid making too many sweeping balance-affecting changes at once, but I'm not sure.

In any case, I think the only thing on my to-do list for this PR is testing right now, so I'll undraft it and flag it as ready for review.

Ilysen:
Purifier doesn't currently let you remove starting traits; they have (and have always had) special behavior that prevents them from being purified.

I reasoned that such traits being purifiable is valid because they're induced by targeted mutation that's decidedly non-human; being naturally Sleepy is different from being Sleepy because you injected rabbit mutagen into yourself. the trait may represent the same effects, but it's still a sudden and forceful change that was brought on by rabbit mutagen - it isn't a normal human thing. same with things like Ugly, Albino, Night Vision, you name it; I believe that traits induced by mutation can feasibly be explained as nonhuman variants, and the game even backs me up on this because a variety of such traits (such as Weak Stomach) that can feasibly be explained as "humanlike" can still be purified if you get them through mutation, but are permanent if you take them in chargen, since they're part of your normal body instead of being added after the fact through mutating.

I think a lot of traits are nonpurifiable right now in the current state of the game as a result of oversights in the unfinished system, of which there are many. some examples of traits that are very much nonhuman but that cannot be purified include Gastropod Foot, Webbed Hands, Leaves, Thorns, Solar Sensitivity, Sores, Heat Dependent and other ectothermic mutations, etc. coupled with the gameplay stance that some "baseline" mutations can never be removed compared to others - especially since purifier currently doesn't work properly to begin with - isn't consistent, I'm inclined to stick to my guns on this one unless a dev, like, asks me to change it.

Terrorforge:
Oh, I think I have misunderstood how mutating interacts with starting traits now. It used to be possible to override, say, Slow Healer as a starting trait by mutating Fast Healer. But this seems to no longer be the case? Fast Healer can mutate into Very Fast Healer, but Slow Healer can't turn into Fast Healer, nor can Fast Healer turn into Deterioration. That seems _very_ wrong to me, but I guess it's not your problem.

And given that the game does apparently draw this distinction, I think your argument holds up. Please carry on, and thanks for putting in the work.

Ilysen:
I hope I didn't come off as confrontational ^^;. since you mentioned, I got curious and did some local testing, since I remember that behavior as well. it appears that with this code, starting traits can't be changed anymore through mutation, and definitely not through purifier, which will never remove them. when I used debug to keep mutating myself using medical primer and Slow Healer as a starting trait, it was never removed at all - until I flagged it as a normal trait and not a starting trait, at which point it was removed successfully.

I'm not certain if this is something that the existing (i.e. current state of the game) system does as well, but unfortunately, my laptop compiles so slowly (>1 hour) that it might take me a bit to check. for a very long time, starting traits could be mutated out of but not purified on their own; I'm not sure if this is desired behavior with the new system, but if it is, I'll make sure the functionality is preserved

Terrorforge:
I just did a bunch of similar testing on a very recent build, and Slow Healer wouldn't budge no matter how much Lizard or Troglobite mutating I did, nor would Fast Healer turn into Weakening when I did a bunch of Alpha. So the existing system definitely does that.

When I brought it up in the Discord just now, everyone seemed as confused by it as me, so I don't _think_ it's desired behavior. But I didn't get 100% confirmation from a senior dev, so if you're thinking about changing it back to the old behavior, maybe run it past the Devcord real quick

Ilysen:
I guess that leaves it in an awkward spot. so this PR doesn't *technically* change the existing functionality, but it might be a good idea to do so due to that inconsistency (i.e. to change it so that starting traits can be "lost" by mutating other traits, but that purifier will bring them back.) I wonder what the best call is on that. :p

Terrorforge:
Mylie mentioned in the last PR that phenotype isn't supposed to be preserved, because of the whole "human mutagen" conceit. The idea is that if you're nearsighted, and you mutate stalked snail eyes, and then you get rid of them with human mutagen, you're going to get human eyes again, but there's no guarantee they'll be the _same_ human eyes. So I think the intended implementation is that purifier cannot get rid of starting traits directly, but if you mutate them away with other mutagens, it can just get rid of those mutations without restoring the old traits.

But I feel like that runs into the "default human" problem I mentioned at the start. Like, why would it _always_ get rid of your myopia? If phenotype isn't preserved, shouldn't it be equally possible for someone with 20/20 vision to become shortsighted?

I suppose one possible solution would be to really lean into the "human mutagen" thing and just slap every "normal" trait into the "Human" category. It would be a bit of a mess, to put it mildly, since many of those traits would conflict with each other. But I feel like this is the logical endpoint of "human mutagen".

Your suggestion is also an option, and I think it would work great mechanically, but imo it would mean abandoning the "human mutagen" concept. Restoring you to exactly how you were before you started taking mutagen is actually far _more_ like nonspecifically "purifying" than Purifier ever used to be.

Ilysen:
it's a tough call for me without any easy answers. Mylie had a lot of ideas that she kind of just dropped, since it seems like she's abandoned her DDA projects for the time being, and given how unfinished the system feels right now (what with how many mutations are unpurifiable due to lack of being listed in the human traits and so on), I'm not sure she had a solid plan on some of this stuff.

I figure that the idea of having purifier try to keep you at your starting trait works because it brings you back to *your* state of existing as a pure human being. it's kind of really hard to work strictly within that fluff; what defines a baseline human? should it affect things like your skin color, hair and eye color, etc - because these things are genetically determined too? why doesn't it affect your starting traits in the first place, given that some things like Albino are strictly because of genetic abnormalities that purifier would correct if it's aiming for a "baseline" human? what would happen if you gave cattle mutagen to a cow, or feline mutagen to a cat? is a mutagen's effects determined in part by the genetics of the beings whose samples are used to make it, or does it function in some other way?

I think it presents a kind of difficult place where there's always going to be some kind of disconnect between the originally-envisioned fluff and the actual mechanics, so it's really hard for me to determine what the right path forward about it is. :n

Terrorforge:
Yeah, I don't know either. I guess seek input from senior devs, and if they're as unsure as we are, just make an executive decision and do what sounds best to you.

Ilysen:
Ended up chatting with kevin and some other people about it in the community 'cord. Apparently the idyll is to have it aim for zero traits at all - with starting mutations either not being treated specially at all, or to make them resistant to purification but not immune - that ties into the broader design goal of separating points from character gen and turning it into a "this is who you are at the start of the game" kind of thing instead of "this is who you are largely for the whole save."

I think to keep the gameplay scope roughly unaffected for this PR that I won't change it right now unless a dev wants that. as a result, the current behavior is that starting traits can directly turn into other mutations, but they cannot be purified on their own and that purifier will aim to revert the player back to those traits when possible. for future work, though, it seems like we have a solid path forward at least!

Ragecrabbo:
One small thing to note is that purifier requires a human sample and is largely found in labs which often also have cloning tubes so having the "default human" be based on some sort of born-in-a-test-tube ideal blank slate human produced by scientists could be an option. Ultimately though the decision is yours on how you want to go about it and im sure you have more insight on the subject given how much you've worked on it 😄

Ilysen:
the tough part of that for me is that you can also get human samples by dissecting NPCs, human corpses, ferals, etc., so having purifier aim for a specific template presumed to originate in labs might not make sense. in any case, I'll hold the purifier stuff (with regards to its interaction with starting traits) for a future MR, since I wanna avoid doing too much on this one.

other than that, I ttthink this is ready for review? I haven't encountered any problems during testing, so it's just up to the dev team if anything needs to change.

src/mutation_data.cpp
```diff
@@ -104,6 +104,7 @@ void mutation_category_trait::load( const JsonObject &jsobj )
 
     jsobj.get_member( "mutagen_message" ).read( new_category.raw_mutagen_message );
     new_category.wip = jsobj.get_bool( "wip", false );
+    new_category.skip_test = jsobj.get_bool( "skip_test", false );
```
dseguin, line 107: Looks like `skip_test` currently serves the same purpose as `wip`. Any reason not to just use `wip` instead?

dseguin (approved):
Seems good to me aside from just a minor observation

src/mutation_data.cpp
Ilysen: Nomenclature, mainly - `wip` is misleading for categories that aren't still in the process of being developed, so I wanted to avoid having a misnomer. I can merge the two into a unified `skip_test` variable if that'd work, although I'm not sure if I should put in obsoletion warnings and etc to account for mods that might use it.

anoobindisguise:
What exactly does "mutagenic slurry" do?

Ilysen:
that's correct. it was made to replace the ludicrous vitamin list in the generic mutagen previously, which had all of those things manually defined, hochi mama; now you can just include that gross slurry in anything that's meant to be slightly mutagenic but generally pretty terrible for you, ala mutated legs and whatnot.

Dignut:
So, I don't quite know how to feel about the changes to chaos and unstable with them giving instability. it feels like they've become another "why would I ever in my right mind take this" trait since they don't give enough points to justify dooming your character's future, I feel like it should give less or maybe have damage gained from these cap off at a certain point.

Ilysen:
I'd point out that "Genetic Chaos" is the name of the mutation and it's exclusive to the mutation tree based around your body being super fucked up. predictability and benefit are not priorities there, and the previous functionality meant that it was the opposite, of literally free positive mutations with no downside. they're not supposed to be positive traits. :p

that said, if you have some coding ability and want to make a PR to alter the functionality, I'd encourage you to do so. it's not something I'd pursue directly, but if people agree with you, I could see it changing again.

Dignut:
I just think it becomes kind of pointless to pick unstable or chaos now since you might as well just go "well If I'm going to get fucked I might as well just take downward spiral" I also don't know how you'll get untyped mutations like pointed horns or gourmand anymore outside of radiation or yuggs since regular mutagen no longer mutates you generically with the new system with chaos and unstable the only source of non-typed positive mutations being rendered a write-off basically.

Ilysen:
actually, after thinking about it for a little while longer, it does kind of bother me that the changes end up making the traits less interesting because it now inevitably becomes a strict downside rather than a strict upside, when it does seem like it'd be more fun for it to always hover somewhere in between. I'm not really sure the best way to handle that - maybe true random mutations could be *true* random, and not care about normal positive/negative chances, or maybe a different approach would work better. I'll give it some thought and see what I come up with.

Dignut:
Oh yeah I wouldn't mind it just giving negatives and positives at random with no regard but the instability really hurts especially if you ever plan on eventually popping regular mutagen.

PR #60234, opened 2022-08-17
Animal ports from TropiCata to Mainline + egg density fixes for insects and birds

Xaritscin:
#### Summary
Mods "ports animals from mod"

#### Purpose of change

Longstanding plan to move several mobs from the mod to mainline based on accuracy and plausibility seeing that they arent exclusive to the tropics.

#### Describe the solution

The following animals were ported from mod:

-birds: hummingbirds, woodpeckers, rails, cormorants, grebes, geese (domestic variants)
-amphibians: toads (including a zombie version)
-arthropods: butterflies and moths, cicadas, woodlice, diving beetles, water scorpions
-mammals: llamas

EDIT: to evade fails on item test, i updated the weights and volumes of the generic bird(55g/50ml), insect(275g/250ml) and canadian goose(143g/130ml) eggs to get most of the animal eggs out of the "bad density" list. this also fixed it for owlbear eggs in Magiclysm.

#### Describe alternatives you've considered

I was considering on moving guinea pigs, parrots, cockatoos and budgerigars too but due to New England temperatures i dont think those creatures would survive without indoor heating.

#### Testing

Grabbed the files and generated a new map, started strolling around, found several of the animals in their respective places. only crashes at load were due to build syntax errors that are being fixed or already solved.

#### Additional context

most if not all the animals ported have a worldwide distribution, and some like llamas can probably survive just fine in the far north due to their wooly covering.

PR #61037, opened 2022-09-16
Allow newly added eggs to be cooked

RenechCDDA:
#### Summary
None

#### Purpose of change
#60234 brought over some very nice birds (and insects*) from a mainline mod, and their eggs. Unfortunately, they can't be cooked.

#### Describe the solution
Add them to the list of bird (and insect*) eggs that can be used in existing recipes.

Balut? Please don't add insect egg balut.

#### Testing
Modified locally, spawned new character and debugged all recipes. Looked at the egg recipes and made sure that they included the new eggs.


data/json/requirements/cooking_components.json
```diff
@@ -77,7 +77,15 @@
         [ "egg_goose_canadian", 1 ],
         [ "egg_turkey", 1 ],
         [ "egg_pheasant", 1 ],
-        [ "egg_cockatrice", 1 ]
+        [ "egg_cockatrice", 1 ],
+        [ "egg_goose", 1 ],
+        [ "egg_goose_golden", 1 ],
+        [ "egg_hummingbird", 1 ],
```
Fris0uman, line 83: Apparently hummingbird eggs are ridiculously tiny irl so the ratio is not 1:1 with the other eggs. 


https://pbs.twimg.com/media/Cir_3u0WYAArUgj.jpg

RenechCDDA: That is true, but I had originally checked their copy-froms and they end up copying from crows (which copy from chickens). However your comment made me double check and the hummingbirds in-game are not some enlarged mutant variety, but just normal-sized hummingbirds - so copying from chickens is definitely wrong. I will put in another commit to resolve this when I can.

RenechCDDA: I decided to simply change their requirement to the `baby_count` for the two biggest offenders, so one portion of eggs for cooking is one litter. There are some bigger issues with the existing (not ported from tropiclysm) bird eggs, but I think that's a little out of scope.

PR #61025, opened 2022-09-15
More misc vehicle menu fixes

irwiss:
#### Summary
None

#### Purpose of change

Make controlled vehicle menu offer examine menu item to match muscle memory before #60868

#### Describe the solution

Enable the menu item when controlling vehicle

Also slight cleanups;
Merge `vehicle::start_folding_activity` to call site - trivial, used twice before, only one call left and not much reuse potential
Remove `vehicle::is_open` - unused in game code, only in tests, adjusts tests to use `vehicle_part::open`

#### Testing

^ twice on controls tile - should see examine/remember vehicle menu items
Test suite should pass using vehicle_part::open

#### Additional context
image

PR #61042, opened 2022-09-16
Replace veh->is_open with vehicle_part::open (android)

irwiss:
#### Summary
None

#### Purpose of change

Fix android build, missed 2 calls https://github.com/CleverRaven/Cataclysm-DDA/pull/61025#issuecomment-1249217412

#### Describe the solution

Replace with vehicle_part::open

#### Testing

Android build should work


akrieger:
Confirmed this fixes the build.

PR #61047, opened 2022-09-16
Forklifts!

Jarewill:
#### Summary
Content "Updates forklifts"

#### Purpose of change
I love forklifts! However I felt their representation in the game wasn't correct.
For one, in game forklifts are made out of the usual car frames/boards instead of proper heavy duty ones.
Forklifts are meant to be tools that lift and carry really heavy things, so having a simple sheet metal for a roof instead of a heavy duty one that wouldn't collapse on the driver was really off.

#### Describe the solution
I based all the changes on my own forklift.
For one, it's now completely heavy-duty with frames, quarterpanels and roofs.

The back quarterpanels were also replaced with a new counterweight part, which is essentially a heavier quarterpanel.
A single counterweight is 750kg, which is doubled since there are two installed, again this was based on my own forklift's counterweight.

Wheels I was very unsure about, my forklift has front wheels that are bigger than the back ones and only the back ones are steerable.
I measured them with tape and the front ones ended up being roughly the same diameter as my car's, so I went with a car wheel for the front and a small wheel for the back.

I also gave them headlights, but I'm not sure about that.
I don't know if there's any item to represent a caged spotlight in the game however.

There's also a diesel variant to the electric one using the newly added I4 diesel engine!

Not adding a counterweight item.

#### Testing
Spawned the forklifts and drove a little.
Spawned a military helipad and confirmed both variants spawn there.

#### Additional context
I need help with this.
The only place where I know a forklift can spawn is the military helipad, so in terms of spawn groups this PR is still incomplete.
There are frankly way too many individual mapgen files to go through individually to check for forklift spawns, so any pointers to where they can spawn would be very helpful.


Golfavel:
Did a quick search, these should be all of the forklifts spawns:
 - Recycle centers
 - Apple orchards (this I can confirm having seen)
 - ```aircraft_carrier_z1```
 - Abandoned warehouses
 - Helipads
 - Lumberyards
 - ```lab_cargo_surface```
 - ```lab_1x1_security```
 - ```mil_base_z0```
 - Malls (roof and second floor)

Jarewill:
So if I understand this correctly, making the forklifts heavier messed up the vehicle drag tests?
How would I go about fixing that?
And also fixing the spelling errors because (at least according to google) they are spelled correctly.

mqrause:
> So if I understand this correctly, making the forklifts heavier messed up the vehicle drag tests? How would I go about fixing that?

The test has some precalculated expected values here: https://github.com/CleverRaven/Cataclysm-DDA/blob/fd463d46976756135990435f9b00ff295afcdd32/tests/vehicle_drag_test.cpp#L282
The test failure prints a replacement of the expected values: `test_vehicle_drag( "forklift", 0.565988, 6.504838, 3683.790625, 7182, 7507 );`
You just need to switch them out (assuming all your changes make sense).

Jarewill:
I am pretty sure they make sense as forklifts are pretty heavy-duty machines.
They need to weight a lot in order to balance out the things they are supposed to lift, otherwise they would turn over from the unbalanced weight of the object.

Sadly I still have no clue how to compile the game and all my attempts have been in vain, so I'm hoping that the code you provided will be a fix.

mqrause:
That's honestly nothing you should have to compile the game for. Json changes are meant to be just json changes.
This should just work, though.

data/json/items/vehicle/plating.json
```diff
@@ -177,5 +177,22 @@
...
+    "type": "GENERIC",
+    "id": "counterweight",
+    "name": { "str": "counterweight" },
+    "description": "A heavy block of steel used in balancing other heavy things.",
+    "weight": "750000 g",
+    "to_hit": -10,
+    "color": "dark_gray",
+    "symbol": "]",
+    "material": [ "steel" ],
+    "volume": "20 L",
```
TheShadowFerret, line 191: Did you also get the dimensions for the counterweight, as this is much denser than our steel ingots 🤔

Jarewill: I have to admit that I didn't because I didn't think that would matter, so I just copied the volume of a heavy duty frame.
However since it does, I went ahead and took measurements of the counterweight in my forklift and came with a number of 660 liters, which I divided by 2 because there are no multi-tile parts as far as I know.

I also searched around on what material they are made out of which came up with either cast-iron or concrete, iron being more commonly used.

I also did more research on how to install and uninstall those counterweights and it turns out you only need a wrench rather than a whole welding setup.

propurty:
Not sure if the lab cargo location listed above includes a TCL. But forklifts do spawn there, just in case the TCL wasn't covered.

Jarewill:
> Not sure if the lab cargo location listed above includes a TCL. But forklifts do spawn there, just in case the TCL wasn't covered.

Hah, fellow forklift enthusiast!
Yes, the lab locations do include TCLs, which causes the new forklift variant to spawn there as well.

PR #61049, opened 2022-09-16
Rewrite bionic fuel system

Hirmuolio:
#### Summary
Content "Rewrite bionic fuel system."

This is same as #60662 but I somehow messed up that branch horribly.

#### Purpose of change

Bionic fuel system currently works by saving {"material_id", "quantity"} pairs as variables into player data when player eats fuel for the bionics. (Yes the fuel is still handled as being eaten!)

This is not a good way to do it. The biggest problem is that it uses material energy attribute wrong. The material energy is in energy per volume. But there is no volume info available.

To use material energy attribute correctly the bionic system must work with items.

#### Describe the solution

Bionic fuel exists as items. 
It was supposed to be simple to switch where bionics look for their fuel. But it turns out that the bionic code was very "side effect oriented". One function would set the fuels in player variables then another function would read the fuels from player variables. Instead of trying to refactor that mess I ended up rewriting everything...

All items with `"BIONIC_FUEL_SOURCE"` will have their contents available as fuel for fuel powered bionics.

Fuel powered bionics now install pseudo item container for the fuel. This container is visible in worn equipment window.

Player refuels the bionic by placing fuel into the appropriate fuel tank pseudo item.

Added ability to limit pocket contents by item material. So you can only put items made out of gasoline into gasoline tank.

Battery system bionic can use medium and light batteries. This does lower its max stored charges a bit (used to be 2000 which is about same as heavy batteries). But the more convenient reload action should compensate.

Also simplified the autostart and fuel saving systems.
Autostart system has been removed. (it used to turn fueled bionic on if power level was below treshold).
Reworked fuel saving (it used to turn fueled bionic off if power was above treshold).

Now the bionic will simply not consume any fuel unless power is below the treshold.
Fuel powered bionics will behave as if they were "off" if power is above treshold. (muscle bionics slow you down less when off).
Bionic that is "ON" but not active due to fuel saving will show "(inactive)" in the bionic menu.

Bionic tanks visible in inventory. You can insert/remove items to/from them just like any other container:
inventory

Bionic window is basically the same as before:
bionicnew

Known issue: NPCs do not refill their bionic tanks. NPCs do not reload their magazines either and I do not really know enough about the NPC code to quickly fix.
Probably best behavior would be to every now and then when out of combat see if any container with `"BIONIC_FUEL_SOURCE"` flag can be refilled and refill them.

#### Testing

Gasoline fuel cell works (8 kJ/s).
Oil generator works (9 kJ/5s).
Cable charger works when connected to UPS, solar backpack and vehicle.

Fuel can be reloaded and unloaded.
Running out of fuel turns bionic off.
Saving and loading works.
Loading old saves migrates properly. Even the charges are migrated!


Maleclypse:
I want to make sure I understand "Known issue: NPCs do not refill their bionic tanks. NPCs do not reload their magazines either and I do not really know enough about the NPC code to quickly fix."  You are saying that these two are pre-existing issues that are not fixed by this refactor or are you saying that these two are new issues that are side effects of the refactor and you don't know how to fix them?

Hirmuolio:
In old system NPCs knew how to eat fuel for bionics (at least there was code for it. Didn't test if it worked).
NPCs do not reload the new bionic fuel tanks. This is new issue.

The inability to reload magazines is old issue. 
But if someone made NPCs reload magazines then it should be trivial to also make them reload bionics via the same code.

anoobindisguise:
> In old system NPCs knew how to eat fuel for bionics (at least there was code for it. Didn't test if it worked). NPCs do not reload the new bionic fuel tanks. This is new issue.
> 
> The inability to reload magazines is old issue. But if someone made NPCs reload magazines then it should be trivial to also make them reload bionics via the same code.
> 
> I left the code for activating the bionics when fuel is available. So if they start reloading the tanks the bionics should start working.

I've had trouble with my NPCs not eating the fuel I gave them (not eating motor oil for their oil generators) so I had eventually just decided to give them trickle chargers, which worked a lot better.

PatrikLundell:
While it's presumably faster to replace a battery in a battery slot than it is to suck the charges out of a battery, I don't think that compensates for having a comparatively low backup power storage, as that means using up valuable inventory space to lug around additional batteries (that can easily be drained by the same ("flora") EMP attacks that forced you to make use of the backup in the first place).

Terrorforge:
> While it's presumably faster to replace a battery in a battery slot than it is to suck the charges out of a battery, I don't think that compensates for having a comparatively low backup power storage, as that means using up valuable inventory space to lug around additional batteries (that can easily be drained by the same ("flora") EMP attacks that forced you to make use of the backup in the first place).

It doesn't need to be compensated, because it used to be hilariously overpowered. Electricity is easily and infinitely renewable and batteries are easy to acquire and carry, so the fact that it also gave you 2000 storage units for free was just absurd. Even after being nerfed into the ground like this, it's still probably the second best power source after Metabolic Interchange. The fact that it leaves you vulnerable to EMPs is basically the only downside it has, and that's not exactly an everyday problem.

KHeket:
> Would it be hard to create a separate category for the internal bionic fuel storage, rather than putting it in "worn items"? Like you could have worn items, but then also "bionic fuel storage" below it so they are visually separated rather than being mixed in.
> 
> Also, I just updated and am playing with the changes. So far so good, but I do miss the "auto off" capability in the old version to where when fuel hit max the bionic would shut itself off.

Me too missed this option, Like and autoturn on option. Now you must turn on cabel chatger every time by yourself, but previously you just set outoturn on persent of charge. This update make me pressing more buttons

Terrorforge:
> So far so good, but I do miss the "auto off" capability

That option still exists, it just works a little differently. You can set a max threshold in the bionics menu, and just keep the power generating bionic on. It won't explicitly turn off, but it'll stop burning fuel when you go above the threshold.

> Now you must turn on cabel chatger every time by yourself, but previously you just set outoturn on persent of charge.

This, however, is pretty annoying. The cable charger won't let you turn it on unless you're hooked up to a power source, and automatically turns off when it becomes unhooked. But this PR has already been merged, so make a new issue for it.

PenceGL:
@Hirmuolio I just noticed this as well:
cataclysm-tiles_dVSIpc0yf1

PR #61057, opened 2022-09-16
Update missions.json

pantherpusher:
#### Summary
Bugfixes "Fixes crash caused by typo"

#### Purpose of change

To fix a typo that was causing a crash

#### Describe the solution
 
 corrects the spelling so the game doesn't freak out and crash when a wrong name is used

#### Testing

Tried the change on my own copy of the game, it fixed the crash.

### DIFF
--- a/data/json/npcs/valhalla_cult/missions.json
+++ b/data/json/npcs/valhalla_cult/missions.json
@@ -285,7 +285,7 @@
     "end": {
       "effect": [
         { "u_spawn_item": "signed_chit", "count": 100 },
-        { "mapgen_update": "valahllist_remove_raiding_party", "om_terrain": "pagan_cult_henge" }
+        { "mapgen_update": "valhallist_remove_raiding_party", "om_terrain": "pagan_cult_henge" }
       ]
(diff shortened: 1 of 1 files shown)


MNG-cataclysm (approved):

Maleclypse (changes requested):
So your change would break the game. Because you aren't changing it in both places.

Maleclypse:
The repo currently has `{ "mapgen_update": "valahllist_remove_raiding_party", "om_terrain": "pagan_cult_henge" }` and `"update_mapgen_id": "valahllist_remove_raiding_party",`  so they match. If they don't that will cause it to fail.  It may be failing silently but it's still failing. Can you post the error you are receiving on crash?

pantherpusher:
well then what that means is the crash was caused by something I don't know how to fix and only fixed by making it not execute properly

Maleclypse:
That’s fair and can happen. Could you share the crash log so I can try and fix it?

Maleclypse:
Or make an issue for it with the error

PR #61058, opened 2022-09-16
Partially decouple disease resist from enviro protection

anoobindisguise:
#### Summary
Balance "gas masks no longer make you immune to getting sick"

#### Purpose of change
Total immunity to disease from wearing gas masks was silly and made having bad health a non issue, and made disease basically a complete non issue. It also didn't make sense - the mouth isn't the only vector for disease, and there's no way to wear a gas mask 24/7 because you have to take it off to eat, drink, etc, which exposes you to disease in intervals.

#### Describe the solution
Mouth environment protection matters, but much less so. You are ( (mouth environment protection)^0.3 / 2 )+ 1 times less likely to get sick. If you have 1 mouth protection, this is 33% fewer diseases. If you have 15 protection (24/7 gas mask or the filter bionic) you are about less than half as likely to get sick. If you are dead-set on being immune to diseases, and live in a plastic bubble (powered phase immersion suit) disease rate is cut down to almost 1/3. 

Different rates.

#### Testing
This will be a little tough to test due to the time required.

#### Additional context
Better hunt down those panaecii!

### DIFF
--- a/data/core/tips.json
+++ b/data/core/tips.json
@@ -637,7 +637,7 @@
   {
     "type": "snippet",
     "category": "tip",
-    "text": [ "Face masks protect you from smoke and even reduce your odds of getting sick, but make it harder to breathe, affecting stamina." ]
+    "text": [ "Face masks can help protect from smoke and disease, but make it harder to breathe, affecting stamina." ]
   },
   {
     "type": "snippet",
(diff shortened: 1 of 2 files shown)


Inglonias:
There's a main menu tip that suggests that gas masks will keep you from getting sick. Suggest you edit that tip if you plan to merge this.

Fris0uman:
I think it'd be better to fix the randomly catching cold out of nowhere mechanic first.

PR #61059, opened 2022-09-16
Add looks like to mutagens and primers

secretstamos:
#### Summary
Infrastructure "None"

#### Purpose of change
Adds looks like to all of the mutagens for tilesets

#### Describe the solution
Added looks like fields to a variety of entries

Could be handled for each tileset locally but that ends up being a lot less efficient than just doing it here

#### Testing
N/A

### DIFF
--- a/data/json/items/comestibles/mutagen.json
+++ b/data/json/items/comestibles/mutagen.json
@@ -835,4 +835,4 @@
     "fun": -20,
     "vitamins": [ [ "mutagen", 60 ] ]
   }
-] 
+]


data/json/items/comestibles/mutagen.json
```diff
@@ -786,4 +835,4 @@
     "fun": -20,
     "vitamins": [ [ "mutagen", 60 ] ]
   }
-]
+]
```
Fris0uman, line 838: Not surr what got changed there but it's better to not introduce unecessary changes

secretstamos: It's a space that was removed when I linted the file.
I'll add it back

secretstamos: Done

PR #61060, opened 2022-09-17
min 2 zeds to suffocate while grabbed

anoobindisguise:
#### Summary
Balance "minimum of two zombies to receive suffocation"

#### Purpose of change
Suffocation from zombie grabs is intended to represent multiple zombies putting piling on top of you and crushing you, however even a single zombie's grab could reduce your oxygen and suffocate you. It's my understanding that this mechanic is meant to be a way for getting surrounded to be more dangerous, rather than to enhance the grabs of a single zombie fought one at a time.

#### Describe the solution
Set the minimum grab intensity to 2.

A single zombie could try to choke you to death, while grabbing you, but to my knowledge that would be best represented with a different mechanic for the individual zombie grab, as being strangled is not something that would scale with multiple zombies all grabbing you.

#### Additional context
This was brought on personally by a particular scenario; I was dissecting a corpse when a zombie wandered into my position and started attacking me. Having spent about 20 minutes working on the corpse I didn't want to start over and, being basically invincible against a single zombie (it being completely unable to damage me whatsoever) just let the zombie sit there and attack me. I ended up losing half my torso health before I could stop out of the activity and address the problem. Stupid, maybe, but it felt kind of... dumb. If it literally cannot scratch me it shouldn't be able to crush my windpipe. Although you probably shouldn't be able to perform a dissection or whatever while being attacked.

There was a second time where a single zombie grabbed me instantly on the turn I spawned a new character and I immediately began to suffocate. This is an unrelated bug and this isn't necessary to fix this extreme edge case but it would tangentially fix that issue.

### DIFF
--- a/src/suffer.cpp
+++ b/src/suffer.cpp
@@ -311,7 +311,7 @@ void suffer::while_grabbed( Character &you )
     int grab_intensity = you.get_effect_int( effect_grabbed, body_part_torso );
 
     // you should have trouble breathing as you get swarmed by zombies grabbing you
-    if ( grab_intensity < 2 ) {
+    if( grab_intensity < 2 ) {
         return;
     } else if( grab_intensity == 2 ) {
         // only a chance to lose breath at low grab chance, none with only a single zombie


PatrikLundell:
I agree your hands shouldn't be steady enough to perform a dissection while being under attack by something that can't hurt you physically (but might cause you to hurt yourself with the scalpel....).
Also agree that piling suffocation should require a pile.

src/suffer.cpp
```diff
@@ -311,8 +311,8 @@ void suffer::while_grabbed( Character &you )
     int grab_intensity = you.get_effect_int( effect_grabbed, body_part_torso );
 
     // you should have trouble breathing as you get swarmed by zombies grabbing you
-    if( grab_intensity <= 2 ) {
-        // only a chance to lose breath at low grab chance
+    if( grab_intensity == 2 ) {
```
fungamer2-2, line 314: I think you may also want to add the following:
```
if ( grab_intensity < 2 ) {
    return;
}
```
Because, as you have it now, it won't exactly work when `grab_intensity` is 1, because if `grab_intensity` is 1, the `grab_intensity == 2` will be false, but then the `grab_intensity <= 4` will be true and it will do the suffocation part, which isn't desired behavior in this case.

anoobindisguise: Thank you, I can't believe I missed that.

PR #61061, opened 2022-09-17
Pixelscale per spritesheet

mlange-42:
#### Summary
Infrastructure "Pixelscale per spritesheet"

#### Purpose of change

Allow to mix spritesheets with different pixel scaling.

Particularly useful for include via symlinks.

Required to make the Ultica-ISO tileset. Proportions are competely off when constructing ISO size from 32x32 Ultica tiles (44x22 in ISO). This can be solved by making the tileset 48px or 52px wide, and including Ultica or MSX characters and monsters with pixelscale 2.

#### Describe the solution

* Add an optional property `pixelscale` to sheets
* Use it to calculate the render size (only, not offset etc.)

None

#### Testing

Tried is-game. See screenshot

#### Additional context

Screenshot of 52px Ultica-ISO, with Ultica character included via symlink, and scaled up by factor 2:

grafik

### DIFF
--- a/src/cata_tiles.h
+++ b/src/cata_tiles.h
@@ -259,7 +259,7 @@ class tileset_cache::loader
 
         point sprite_offset;
         point sprite_offset_retracted;
-        float sprite_pixelscale;
+        float sprite_pixelscale = 1.0;
 
         int sprite_width = 0;
         int sprite_height = 0;


Inglonias:
This appears to be the root cause of issue #61103

mlange-42:
Thanks for reporting @Inglonias. Is this based on a before/after merge test, or any other reasoning there?

Already any idea if on the C++ or Python/tileset compose side?

Inglonias:
I did a "bisect" (using the Catapult launcher, not Git) and found that 2022-09-18-0726 is the last build before the above-mentioned issue appears. This is the only graphics-related PR that was merged between that build and the next.

mlange-42:
Indeed. When reverting [this](https://github.com/CleverRaven/Cataclysm-DDA/pull/61061/files#diff-c8561880131eb6314d85f58032f9063284bcfc226b75b3b5df585dcaa219605cL2510), roads are back:

grafik

Inglonias:
Is it possible to turn off pixel scaling for fallback tiles? Or is there a better solution?

mlange-42:
There are tiles with scale 0.0. Default should be 1.0 if not specified. Not sure how/why OM tiles are not initialized with the default.

mlange-42:
Ok, solved. PR coming...

mlange-42:
See #61133.

PR #61063, opened 2022-09-17
Canvas cloth

anoobindisguise:
#### Summary
Content "add canvas cloth, convert appropriate clothing to it"

#### Purpose of change
Canvas is a notable absent cloth type from our lineup. It is waterproof and noticeably sturdier per thickness than cotton - per a grandparent who sews as a hobby, canvas should be significantly stronger than an equivalent thickness of regular cotton due to thicker threads and its strong plain weave.

#### Describe the solution
Adds canvas cloth and heavy duty threads. Canvas specifically represents a GSM 400-450 type of fabric, compared to cotton's 203. Its defensive stats are 2/2.5/2, and it is of POOR breathability. It can be found in tailor spawngroups, or made yourself. Heavy duty thread is made like regular thread with a triple material count, either meticulously twisting individual threads together or (FAR MORE EFFICIENTLY) spinning them at a wheel. It can be used as part of "durable filaments" for sewing survivor gear or stitching regular products, and it's needed for canvas assembly. Individual canvas sheets can be made patchwork out of patches or woven out of threads. These sheets require wax to construct as real canvas is either waterproofed with wax or a nylon blend - since there are no nylon threads, any made by the survivor will be waxed (also Innawoods-friendly!) 

Numerous items were converted to use canvas, and were made waterproof where they were previously not. Most cotton bags such as rifle cases, utility vest, duffel bags, etc are now canvas, as are cotton dusters/trenchcoats, work pants and shirts, canvas bags and sacks (whodathunk) and swag bags, among a few others.

Different defense values

#### Testing
Loaded up good in game and I took a few screens!

#### Additional context
There are some things to note:
I had to adjust the test in requirements.cpp because the addition of heavy duty threads caused some recipes to shoot past 100 alternatives. The test error did not tell me how far beyond 100 I was, so I added that functionality to the test. It then informed me that my recipes had 104 alternatives. Since that's barely an overshoot I decided to raise the test limit to 105. Hopefully this is OK.

What materials can be cut up are hardcoded. I had to add canvas as a cut-uppable material. While I was there I decided to add Epoxy and Thermoplastic since it'll save some time. This is out of scope but seemed like an appropriate place to do that.

Also, when I was adding the recipes for weaving canvas sheets, I noticed that some of the previous cotton sheet recipes weren't adjusted for new tailoring paradigms and were still treating them like they were 22 sheets: 1 patch, as a result you could spin a cotton sheet out of only 5 thread. I dramatically increased the thread cost to 50 per (turning the sheet into scraps makes 32 threads total) and also raised the time required to weave the sheet hugely, because weaving takes a shitton of time IRL. Make patchwork sheets if you value your time.

canvas1
canvas2
canvas3
canvas4
canvas5
canvas6


data/json/items/resources/tailoring.json
```diff
@@ -541,5 +541,84 @@
...
     "count": 50,
     "stack_size": 200
+  },
+  {
+    "id": "sheet_canvas",
+    "type": "GENERIC",
+    "category": "spare_parts",
+    "name": { "str": "canvas sheet" },
+    "description": "A 12 inch by 24 inch sheet of durable and waterproof waxed canvas fabric, suitable for making tough bags and clothing.",
+    "weight": "80 g",
```
TheShadowFerret, line 551: Doesn't this mean all the altered clothing weights need to be changed 👀

LyleSY (approved):
Amazing, what I've spot checked looks thorough and well thought out. Excited to see this in game

PR #61067, opened 2022-09-17
[Magiclysm] Attunement messages

### LINKED ISSUE #59989: hit_me_effect leaves no message in the log
### Describe the bug

Hit_me_effects should leave a message in the message log explaining what happens.  Currently these spells are basically invisible to the message log unless they cause an effect like blindness.

### Steps to reproduce

Debug trait SUN_MAGE
Fight zombies in hand to hand.  
See that they are blinded in the log and by examining the monster.
Ponder why there is no message in the log explaining what happened to blind the monster.

### Expected behavior

I'd expect a log to say what the spell is and who it hits.

### Screenshots

_No response_

08/05/22 download.  I'll put this in tomorrow. Shutting down for the night.

### Additional context

_No response_

Maleclypse:
#### Summary
None

#### Purpose of change
Adds messages for the attunement on hit and hit you enchantments, so that the player can know when they are working and active.  If they aren't messaging, they probably aren't working now.  Fixes #59989

#### Describe the solution
Adds message and npc_message where appropriate.  If it's totally invisible no npc_message only pc message.

None

#### Testing
TBD

### DIFF
--- a/data/mods/Magiclysm/enchantments/Shaman.json
+++ b/data/mods/Magiclysm/enchantments/Shaman.json
@@ -1,30 +1,26 @@
-[{
-		"id": "mind_read",
-		"type": "SPELL",
-		"name": "Mind Reading",
-		"description": "You can predict some movements of your enemies (just slowdown in fact).  If you are seeing this spell, you have debugged it in.",
-		"valid_targets": ["hostile"],
-		"flags": ["SILENT", "NO_PROJECTILE"],
-		"effect": "attack",
-		"shape": "blast",
-		"effect_str": "mind_reading",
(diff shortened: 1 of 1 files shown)


Maleclypse:
@GuardianDll if you have a second let me know what you think of these messages?

GuardianDll:
I don't really know is this a good idea, because it will make a mess out of player's log, spamming a message each time someome attack the player

Maleclypse:
Honestly I’ve hated the attunements I’ve tried because it was hard to know if they worked. Every other trait that has these kinds of effects has a message I thought. This would put them in line with other traits.

PR #61071, opened 2022-09-17
Combat Adaptation -> Berserk (rework)

KittyTac:
#### Summary
Content "Replace Beast's Combat Adaptation with Berserk"

#### Purpose of change
Combat Adaptation was pretty weak and not very flavorful, and something like this change was desired by another person working on mutations.

#### Describe the solution
Beast, instead of Combat Adaptation, now has a Berserk mutation. When activated, for a random duration of 30 seconds to 1 minute, it gives +14 Str, -20 Int, -20 dodge, -20 blocking, -30% move cost, and a strong martial art (that needs to be attached to the mutation somehow). After the time is up, for 10 minutes, the following debuff is applied: -6 Per, -6 Dex, -8 Str, -30% speed, and constant mild pain.

Leaving as is or tweaking the numbers a bit.

#### Testing
Everything works properly, except I can't figure out how to make the mutation give the martial art. Also, the berserk ability should stop working during the crash.

#### Additional context
I had fun writing the flavor text for this and unleashing my inner edgelord.

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -888,7 +888,7 @@
     "id": "ADRENALINE2",
     "name": { "str": "Berserk" },
     "points": 3,
-    "description": "When you are in great danger, you may flip a chemical switch in your brain and for a cost in hunger, fatigue, and thirst your strength will be greatly increased for a random time between 2 and 5 minutes.  However, your mind will become so clouded that dodging will become impossible, and the process will take a toll on your body.  The crash will last 10 minutes.",
+    "description": "When you are in great danger, you may flip a chemical switch in your brain and for a cost in hunger, fatigue, and thirst your strength will be greatly increased for a random time between 2 and 5 minutes.  However, your mind will become so clouded that dodging and blocking will become impossible, and the process will take a toll on your body.  The crash will last 10 minutes.",
     "prereqs": [ "ADRENALINE" ],
     "category": [ "BEAST" ],
     "active": true,
@@ -906,10 +906,11 @@
     "description": "A red film covers your vision, and your mind is empty except for one thought: Kill.  Kill.  Kill.  Kill.  Kill.  Kill.  Kill.",
     "prereqs": [ "ADRENALINE" ],
     "category": [ "BEAST" ],
-    "dodge_modifier": -20,
+    "enchantments": [ { "condition": "ALWAYS", "values": [ { "value": "BONUS_BLOCK", "add": -20 }, { "value": "BONUS_DODGE", "add": -20 } ] } ],
     "movecost_modifier": 0.7,
     "passive_mods": { "str_mod": 14, "int_mod": -20 },
-	"initial_ma_styles": [ "style_berserk" ]
+	"initial_ma_styles": [ "style_berserk" ],
+	"valid": false
   },
   {
     "type": "mutation",


Venera3:
+14 Str and 30% speed for no downsides is ridiculously overpowered (no, dodge is not nearly relevant enough if you kill things in one hit) for a duration that will definitely cover the entirety of a fight, and that's before you add the double-damage melee techs. Consciously activating something called "berserk" is pretty weird on its own, for that matter - a random activation on damage would make more sense.

If you want to have huge boosts (and half the bonuses would already do the trick) the duration should be short enough to have a chance at running out in the middle of a scrap - something like 20 s as the low point. The crash could use higher debuffs, and a hefty stamina/tiredness cost.

KittyTac:
> +14 Str and 30% speed for no downsides is ridiculously overpowered (no, dodge is not nearly relevant enough if you kill things in one hit) for a duration that will definitely cover the entirety of a fight, and that's before you add the double-damage melee techs. Consciously activating something called "berserk" is pretty weird on its own, for that matter - a random activation on damage would make more sense.
> 
> If you want to have huge boosts (and half the bonuses would already do the trick) the duration should be short enough to have a chance at running out in the middle of a scrap - something like 20 s as the low point. The crash could use higher debuffs, and a hefty stamina/tiredness cost.

Yeah I'll nerf the duration, I had my suspicions about it being OP.

KittyTac:
I also wanted to make the bonuses scale back the more armor you wear but I have no idea how to do that.

Maleclypse:
In case you are asleep for the night I answered your question in the discord here https://discord.com/channels/699895400253095956/699896174156709930/1020756130491793468

Amoebka:
14 STR doesn't even remotely allow oneshotting anything though. It's a very modest DPS increase under most realistic situations.
Call me crazy, but I think trading dodge for strength is not even good. The most practical use I see for this would be activating it to break out of a particularly nasty grab and run away.

data/json/mutations/mutations.json
```diff
@@ -886,19 +886,30 @@
...
+  {
+    "type": "mutation",
+    "id": "ADRENALINE2_active",
+    "name": { "str": "Berserk (active)" },
+    "points": 3,
+    "description": "A red film covers your vision, and your mind is empty except for one thought: Kill.  Kill.  Kill.  Kill.  Kill.  Kill.  Kill.",
+    "prereqs": [ "ADRENALINE" ],
+    "category": [ "BEAST" ],
+    "dodge_modifier": -20,
+    "movecost_modifier": 0.7,
```
Terrorforge, line 910: This isn't +30% speed, this is -30% walking cost. Check "Quick" for the format to add Speed.

Terrorforge:
If you're looking to punish defense, what you probably want to do is disable blocking. With decent melee skill, it's a very consistent and very powerful damage negation tool. Especially because Insanely Strong + 14str berserk buff makes it really easy to hit the 40 threshold of Strength + Melee + wblock bonus where you just plain block 100% of all damage. I think it's fine for this to come with some defensive advantages, but greater skill at active blocking seems weird. Immunity to pain, perhaps?

Personally I think activating it on purpose is fine. It's in line with both the sagas and most game implementations. More importantly, making it a random buff would require making it a lot less punishing, gutting the entire high highs/low lows concept and turning it into something less interesting.

KittyTac:
> If you're looking to punish defense, what you probably want to do is disable blocking. With decent melee skill, it's a very consistent and very powerful damage negation tool. Especially because Insanely Strong + 14str berserk buff makes it really easy to hit the 40 threshold of Strength + Melee + wblock bonus where you just plain block 100% of all damage. I think it's fine for this to come with some defensive advantages, but greater skill at active blocking seems weird. Immunity to pain, perhaps?
> 
> Personally I think activating it on purpose is fine. It's in line with both the sagas and most game implementations. More importantly, making it a random buff would require making it a lot less punishing, gutting the entire high highs/low lows concept and turning it into something less interesting.

Ooooh yeah I'll do this, thanks!

data/json/mutations/mutations.json
KittyTac: ...actually I like this more, it fits the flavor to be able to get past obstacles much faster, more _Speed_ speed would just make the buffs even stronger.

KittyTac:
> In case you are asleep for the night I answered your question in the discord here https://discord.com/channels/699895400253095956/699896174156709930/1020756130491793468

It doesn't work.

Maleclypse:
> > In case you are asleep for the night I answered your question in the discord here https://discord.com/channels/699895400253095956/699896174156709930/1020756130491793468
> 
> It doesn't work.

Yep, just realized my error on that all of that is tied to character creation not midgame and the other thing I was thinking of was giving a trait not a martial art.  Sorry about that. I'll try and see if it's something I can link into easily.

PR #61075, opened 2022-09-17
Edit coat rack components to include long stick

AirRice:
#### Summary
Balance "Edit coat rack components to include long stick"

#### Purpose of change
The ingame description for the furniture item "coat rack" states, and I quote: 

> "A tall wooden pole with a set of hooks used to store outdoor jackets and hats to allow easy access."

Before this proposed change deconstructing a coat rack provided two 2x4 planks. However, it makes sense to me that such a furniture item, when deconstructed, would provide a long and sturdy enough pole or stick that can be used for crafting purposes. 

#### Describe the solution

Edit the deconstruction recipe of the coat rack furniture item (ID: f_rack_coat) to provide:

- BEFORE: 2~6 nails, 2 planks
- AFTER: 2~6 nails, 1 plank, 1 long stick
image

Additionally, edit the construction recipe of the coat rack furniture item for consistency:

- BEFORE: 8 nails, 3 planks
- AFTER: 8 nails, 2 planks, 1 long stick
image

The products from smashing the furniture has not been changed, as it would make sense for forcefully smashing apart a coat rack to break the long pole part of it to be unusable.

Do not make this change. Make the coat rack provide heavy sticks, long poles (would be insanely OP), or some other long wooden component instead.

#### Testing

1. Spawn in the world.
2. Use the debug menu to spawn in coat rack furniture (ID: f_rack_coat) and the necessary tools for deconstructing it (1 prying, 1 screw driving qualities)
3. Deconstruct the coat rack and observe that it provides the expected components as seen above.
4. Look in the construction menu (default hotkey *) and check that the coat rack recipe is changed as seen above.

#### Additional context
This is my first PR and while it is quite rudimentary I hope to contribute more in the future :)

### DIFF
--- a/data/json/furniture_and_terrain/furniture-storage.json
+++ b/data/json/furniture_and_terrain/furniture-storage.json
@@ -649,9 +649,9 @@
     "coverage": 30,
     "required_str": 4,
     "flags": [ "TRANSPARENT", "FLAMMABLE", "PLACE_ITEM", "BLOCKSDOOR", "MOUNTABLE" ],
-    "deconstruct": { 
-	  "items": [ { "item": "nail", "charges": [ 2, 6 ] }, { "item": "2x4", "count": 1 }, { "item": "stick_long", "count": 1 } ] 
-	},
+    "deconstruct": {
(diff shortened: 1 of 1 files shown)


jeffplata:
Nice change :)

PR #61078, opened 2022-09-17
Coloured concrete walls

gettingusedto:
#### Summary
None

#### Purpose of change

Add painted concrete walls and the ability to paint concrete walls.

#### Describe the solution

Adds painted variant for concrete wall, simple concrete wall, and reinforced concrete wall.

Figuring out why the painted walls in a basement were built out of wood.

#### Testing

I made a mod of it, it works as expected.

#### Additional context

why do there's pink wall but no pink paint?


LyleSY (approved):
Looks like the right approach, I don't see any problems here

PR #61080, opened 2022-09-17
Fix Windows crossbuild

akrieger:
#### Summary
None

#### Purpose of change
PR workflows don't exercise exactly the same build as the release workflows, which error out after the flexbuffer merge with
```
 src/third-party/flatbuffers/util.cpp:34: error: macro "NOMINMAX" is not used [-Werror=unused-macros]
   34 | #    define NOMINMAX
```

#### Describe the solution
Only define `NOMINMAX` when not compiling under mingw.

#### Testing
https://github.com/akrieger/Cataclysm-DDA/actions/runs/3074394060/jobs/4967151736#step:14:450 build succeeded. PR workflows should also succeed.

### DIFF
--- a/src/third-party/flatbuffers/util.cpp
+++ b/src/third-party/flatbuffers/util.cpp
@@ -31,7 +31,9 @@
 #    define WIN32_LEAN_AND_MEAN
 #  endif
 #  ifndef NOMINMAX
+#  if !defined(__MINGW32__)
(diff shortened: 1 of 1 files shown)


akrieger:
The linked build from my repo *built* but the upload to the release failed because my repo isn't setup to host releases yet.

PR #61085, opened 2022-09-18
Add new mutant bird harvest ids and edit monsters that used old ids

AirRice:
#### Summary
Balance "Add new mutant bird harvest ids and edit monsters that used old ids"

#### Purpose of change

In the process of preparing to work on some additional mutated bird monsters I noticed that there was no specific harvest ID set for mutant birds. While the two existing mutant bird enemies in the game (the oversized crow and bloated corvid, specifically) do in fact give mutant butchering products, this was based off of "bird_large", what I can only assume is a very outdated harvest id used for the albino penguin, yet another enemy I assume is quite old in its implementation.

With this change I wanted to make it clear that **mutant birds would drop mutant products**, and reduce any uncertainty that could be introduced from this vague harvest id name.

#### Describe the solution
- Add new "mutant_bird" and "mutant_bird_flightless" harvest ids to cover mutant bird type enemies. The former is for mutant birds that can still fly and provide a similar ratio of butchery products as "bird_small" does, just mutant; The latter is for mutant birds that _cannot_ fly such as the albino penguin or the bloated corvid, and provide mutant fat and more bone than the former due to them being flightless, which would mean that they do not need to be light enough to fly anymore. 
- Edit the existing "bird_large" harvest id to provide non-mutant products as it would be confusing otherwise. This likely fixes a weird quirk that occurs in several mods for non-mutant bird butchering; An example is Flamingoes from TropiCataclysm, which would have yielded mutant meat otherwise. Mutant bird-type enemies that were originally using this have been changed to use one of the prior two specifically mutant bird harvest ids.
- ~~Move the existing mutant bird JSON entries into a different file. So far this is only the corvids: oversized crow, and bloated corvid. This is mostly for record keeping purposes and future development on my end.~~ Reverted as moving json entries is not usually done.

Not making this change. Simply changing the id "bird_large" to be "mutant_bird_large" and not making other changes. 

#### Testing

1. Spawn in a bloated corvid, kill it and observe that the butchery products are as expected.
2. Spawn in an albino penguin, kill it and observe that the butchery products are as expected.
3. Spawn in afeaver, kill it and observe that the butchery products are as expected.

#### Additional context
I will likely make a contribution in the near future regarding mutant waterfowl such as geese. I know this has been talked about quite a bit on the devcord and I am currently in the process of brainstorming; This change is simply for QoL of development prior to these.


LyleSY:
I think I've fixed the merge conflicts. This will conflict with my PR #61187 but otherwise seems good to go

LyleSY (changes requested):
Please migrate the families and weakpoint_sets changes to the new mutant_bird JSON from my PR. Otherwise good to go

LyleSY (approved):
Good stuff, thanks!

Maleclypse:
`Move the existing mutant bird JSON entries into a different file. So far this is only the corvids: oversized crow, and bloated corvid. This is mostly for record keeping purposes and future development on my end.` 
So this messes with history and gitblame so I would say that there is some reticence on merging this while they are being moved to a new file.  Generally we try to avoid moving existing things from their current file even if it isn't' a great fit until the file becomes so large it needs to be moved for organization sake.  That occurs around 10k lines of JSON and the current bird file is less than 1k lines of json.

AirRice:
Ahh I see, I wasn't aware of when exactly people move json entries, I've only ever worked in semi small repos where this is fine to do, haha.

I'll put in a commit to put the entries back into bird.json

LyleSY:
Mapgen error doesn't look related

Maleclypse (approved):

PR #61101, opened 2022-09-18
Craft seeds from spurge flowers

Light-Wave:
#### Summary
Content "Craft seeds from spurge flowers"

#### Purpose of change
Spurge flowers are useful and valuable. If you find some seeds you can grow them, which yield flowers and more seeds. But you couldn't craft seeds directly from the flowers. This PR seeks to remedy that.

#### Describe the solution
Make the seeds craftable from the flowers

Make the foraged flowers drop seeds instead

### DIFF
--- a/data/json/recipes/food/seed.json
+++ b/data/json/recipes/food/seed.json
@@ -215,6 +215,12 @@
     "components": [ [ [ "raw_burdock", 4 ] ] ]
   },
   {
+    "result": "seed_spurge",
+    "type": "recipe",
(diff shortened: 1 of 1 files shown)


Light-Wave:
Upon closer inspection, foraging spurge flowers provides seeds as well, thus making the flowers farmable without a need for this recipe.
Previously you were not able to turn flowers to seeds, but instead got seeds along with the flowers. This kind of simulates that the part of the flower that is useful, and the part that holds seeds, are two different parts and not interchangeable. I think I like that better.

PR #61102, opened 2022-09-18
Force locale independent parsing for flatbuffer parser.

### LINKED ISSUE #61095: Error: Json error: data/json/character_modifiers.json:119:13: denominator cannot be set to 0
### Describe the bug
```

 DEBUG    : Error: Json error: data/json/character_modifiers.json:119:13: denominator cannot be set to 0

    "description": "Slip prevention modifier",
    "mod_type": "x",
    "value":
            ^
             {
      "limb_score": [ [ "grip", 3.0 ], [ "lift", 2.0 ], "footing" ],
      "override_encumb": true,


 FUNCTION : bool main_menu::new_character_tab()
 FILE     : src/main_menu.cpp
 LINE     : 947
 VERSION  : 29cdbc1
```

### Steps to reproduce

1. Create new world.
2. Try to create new character in that world.

Or try to load any save.

### Expected behavior

Game works.

### Screenshots

_No response_

Build: 2022-09-18-0726 
Version: 29cdbc1

### Additional context

Game works on build: 2022-09-17-0626, version cdcb071

akrieger:
#### Summary
None

#### Purpose of change
Fixes #61095. Non-msvc builds were using locale dependent float parsing in the flatbuffer parser. This broke float parsing if the user's locale was set to something where the decimal mark is a ',' instead of a '.'. Various asserts would trip because the parser would silently zero out the value instead of erroring.

#### Describe the solution
Force locale independent parsing in the flatbuffer parser with a #define.

- Rewriting the parser to throw on parse failures. Probably worthwhile to do in the future.

#### Testing
Discussion on #61095 showed that the -msvc builds worked fine, and visual inspection of the code showed flatbuffers defaults to locale-independent parsing in msvc builds. A test build in https://github.com/akrieger/Cataclysm-DDA/actions/runs/3077977683 showed that the osx and windows cross builds both used locale dependent parsers.

### DIFF
--- a/src/third-party/flatbuffers/base.h
+++ b/src/third-party/flatbuffers/base.h
@@ -253,6 +253,11 @@ namespace flatbuffers {
 #endif // !FLATBUFFERS_HAS_NEW_STRTOD
 
 #ifndef FLATBUFFERS_LOCALE_INDEPENDENT
+#define FLATBUFFERS_LOCALE_INDEPENDENT 1
+#elif FLATBUFFERS_LOCALE_INDEPENDENT != 1
+#error CDDA requires locale independent number parsing.
+#endif
+#ifndef FLATBUFFERS_LOCALE_INDEPENDENT
   // Enable locale independent functions {strtof_l, strtod_l,strtoll_l, strtoull_l}.
   #if ((defined(_MSC_VER) && _MSC_VER >= 1800)            || \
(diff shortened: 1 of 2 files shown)


Wishbringer:
This breaks MinGW builds as seen under Checks.

dseguin:
> This breaks MinGW builds as seen under Checks.

Oops sorry, really should have let the last couple of tests finish before merging

akrieger:
More likely root cause is the json loader change. I did test changing auto forage settings after realizing I had broken it, but I'll check auto loot again.

PR #61104, opened 2022-09-18
Fixed tourniquet bleed ratios

JoshuaCF:
#### Summary
Bugfixes "Tourniquets now reduce bleeding properly"

#### Purpose of change
Bleeding logic previously had a bug where having Wound Care and Wound Care Expert proficiencies would actually make tourniquets less effective, rather than more effective. As part of that bug, tourniquets would default to only stopping 1/3rd of bleeds, rather than 2/3rds. This has now been fixed.

#### Describe the solution
Assuming first aid is not being applied, a tourniquet is on, and there are no relevant proficiencies, the if condition can be simplified to this:
`if( !( true && one_in( 3 ) ) && true )`
The conditional would be satisfied if one_in returns false, which it did 2/3rds of the time. The conditional is intended to be satisfied only 1/3rd of the time with a tourniquet. Increasing the prof_bonus (3 -> 4 -> 5) through proficiencies would actually make bleeding happen more often with a tourniquet. To fix this and hopefully make the if statement easier to read, I rewrote it to follow this logic: bleeding occurs when there is no tourniquet OR the tourniquet is ignored (1/3rd of the time, decrease with prof bonus) AND first aid is not being applied.

The change could have been as simple as just adding another inversion (`!`) before the one_in, but I felt that would have made the if statement even more difficult to parse so I decided to change it a bit more so it will be easier to maintain in the future (hopefully).

#### Testing
I inflicted bleed on a wound and applied a tourniquet. I then waited until the first if condition succeeded 30 times (using a removed temp debug msg), and used how long I had to wait to compute the ratio of turns to successful bleeds (only caring about the first if statement). It was 32%, close enough to 1/3rd of the time. I then did the same test with Wound Care and Wound Care Expert to get a value of 19%, which was close enough to the intended 20% that I considered it fixed.

### DIFF
--- a/src/player_hardcoded_effects.cpp
+++ b/src/player_hardcoded_effects.cpp
@@ -282,7 +282,7 @@ static void eff_fun_bleed( Character &u, effect &it )
     prof_bonus = u.has_proficiency( proficiency_prof_wound_care ) ? prof_bonus + 1 : prof_bonus;
     prof_bonus = u.has_proficiency( proficiency_prof_wound_care_expert ) ? prof_bonus + 1 : prof_bonus;
 
-    if ( ( !tourniquet || one_in( prof_bonus ) ) && u.activity.id() != ACT_FIRSTAID ) {
+    if( ( !tourniquet || one_in( prof_bonus ) ) && u.activity.id() != ACT_FIRSTAID ) {
         // Prolonged hemorrhage is a significant risk for developing anemia
         u.vitamin_mod( vitamin_redcells, -intense );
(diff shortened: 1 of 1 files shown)


Fris0uman:
Does this have a related bug report that it fixes?

JoshuaCF:
> Does this have a related bug report that it fixes?

As far as I'm aware, no. Searching "tourniquet" in the bug reports only yields an issue regarding balance around applying pressure.

PR #61109, opened 2022-09-18
Correct repair costs in vehicle UI

### LINKED ISSUE #55200: Vehicle repair costs do not account for degradation

When repairing a vehicle, there are resource costs, based on how badly damaged the part is. This repair will return the part to it's "maximum" HP, which is reduced by how badly degraded the part is. 

However, these two interact in an illogical fashion: If you have an object with very low current and max HP, the total cost for the repair is high, even though the amount of HP repaired is small. For example, repairing a steel plating at pink \ with pink degradation costs 54 welding resources plus 45 scrap metal. Repairing a different steel plating at yellow |\ and yellow degradation costs only 18 welding resources and 15 scrap metal, one-third the cost. But both repairs will restore roughly amount of HP. 

### Solution you would like.

To mitigate this, I think the game should scale the cost on degraded max part HP minus current part HP, so patching  nearly-broken items can be economically viable if replacements haven't yet been found.

The time investment could be left the same, since it stands to reason that trying to keep a nearly-broken thing together still takes more time and effort than small patches on a nearly pristine thing. 

### Additional context

_No response_

irwiss:
#### Summary
None

#### Purpose of change

Fixes #55200

#### Describe the solution

Vehicle interact UI has a few spots where repair time/requirements are evaluated and most of them are slightly different from each other.

Patch makes a few functions that account for part degradation and uses them so repair time/component requirements are uniform both in UI and what gets eaten by repair activity.

#### Testing

Spawn 4 frames, place 2 of them as vehicle frames
Edit the other 2 to have:
3000 degradation, 3000 damage
1000 degradation, 3000 damage
Install them as 2 doors;
Degraded part should require just 1 chunk of steel to repair, 20 welder charges
Normal part should require 3 chunks of steel to repair, 60 welder charges

Before the patch the UI would show the same, but what gets eaten by repair activity is different

### DIFF
--- a/src/veh_interact.cpp
+++ b/src/veh_interact.cpp
@@ -146,9 +146,8 @@ player_activity veh_interact::serialize_activity()
             if( pt != nullptr ) {
                 if( pt->is_broken() ) {
                     time = vp->install_time( player_character );
-                } else if( pt->base.max_damage() > 0 ) {
-                    time = vp->repair_time( player_character ) * ( pt->base.damage() - pt->base.damage_floor(
-                                false ) ) / pt->base.max_damage();
+                } else if( pt->is_repairable() ) {
+                    time = vp->repair_time( player_character ) * pt->repairable_levels();
                 }
             }
             break;
@@ -1199,7 +1198,7 @@ void veh_interact::do_repair()
 
     if( reason == task_reason::INVALID_TARGET ) {
         vehicle_part *most_repairable = get_most_repairable_part();
-        if( most_repairable && most_repairable->damage_percent() ) {
+        if( most_repairable && most_repairable->is_repairable() ) {
             move_cursor( ( most_repairable->mount + dd ).rotate( 3 ) );
             return;
         }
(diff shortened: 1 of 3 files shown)


src/veh_interact.cpp
```diff
@@ -2382,18 +2379,14 @@ void veh_interact::move_cursor( const point &d, int dstart_at )
...
         for( size_t i = 0; i < parts_here.size(); i++ ) {
             vehicle_part &pt = veh->part( parts_here[i] );
 
-            if( pt.base.damage() > pt.base.damage_floor( false ) && pt.info().is_repairable() ) {
+            if( pt.is_repairable() ) {
                 need_repair.push_back( i );
             }
-            if( pt.info().has_flag( "WHEEL" ) ) {
-                wheel = &pt;
-            }
```
Fris0uman, line 2396: Why remove the wheel check?

src/veh_interact.h
```diff
@@ -281,9 +281,6 @@ class veh_interact
          * Updated whenever the cursor moves. */
         std::vector<int> parts_here;
 
-        /* Refers to the wheel (if any) in the currently selected square. */
-        struct vehicle_part *wheel;
-
```
Fris0uman, line 286: same question, this seems unrelated to the other changes

PR #61115, opened 2022-09-19
Update astyle docs to exclude src/third-party

akrieger:
#### Summary
None

#### Purpose of change
It was pointed out in discord that the astyle docs can lead one to accidentally styling the third-party folder. The Makefile recipes do not do this, but manually invoking astyle might. Correct the docs.

#### Describe the solution
Add a --exclude flag to exclude src/third-party.

#### Testing
Ran it myself, it didn't style the third-party folder.

### DIFF
--- a/doc/DEVELOPER_TOOLING.md
+++ b/doc/DEVELOPER_TOOLING.md
@@ -26,7 +26,7 @@ make astyle
 If you have only `astyle` then use:
 
 ```BASH
-astyle --options=.astylerc --recursive src/*.cpp,*.h tests/*.cpp,*.h`
+astyle --options=.astylerc --recursive src/*.cpp,*.h tests/*.cpp,*.h
 ```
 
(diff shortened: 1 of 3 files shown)


doc/DEVELOPER_TOOLING.md
```diff
@@ -26,7 +26,7 @@ make astyle
 If you have only `astyle` then use:
 
 ```BASH
-astyle --options=.astylerc --recursive src/*.cpp,*.h tests/*.cpp,*.h`
+astyle --options=.astylerc --recursive src/*.cpp,*.h tests/*.cpp,*.h --exclude=src/third-party
```
Qrox, line 29: I suggest putting the new option into `.astylerc` instead.

Also, it seems `--ignore-exclude-errors` is also needed because otherwise astyle freaks out over `tests/*.cpp` not containing `src/third-party`.

akrieger: Thanks for pointing that out! Incorporated this feedback.

PR #61122, opened 2022-09-19
Make reload_activity_actor take item::reload_option

irwiss:
#### Summary
None

#### Purpose of change

Follow up on #61046

#### Describe the solution

Simplify call sites for reload_activity_actor by having it take item::reload_option&& directly instead of each call site manually juggling variables for targets vectors and moves.

To make tests work this also changes activity interruptions test to provide a setup function per activity, removing book binder hack.

Also changes reload_activity_actor to work with 2 item_location variables instead of vector of them.

Writing migration for the old activity json, but it's generally a short activity so at the cost of a rare user facing debugmsg when loading - a few lines of code can be saved from being made in actor's deserialize()

#### Testing

Reload/unload guns via "Reload"
Reload/unload guns via "Reload weapon" and "reload wielded item" hotkeys
Reload turret using "reload weapons" hotkey on a turret tile
Reload tests should pass

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -4014,23 +4014,23 @@ reload_activity_actor::reload_activity_actor( item::reload_option &&opt, int ext
 {
     moves_total = opt.moves() + extra_moves;
     quantity = opt.qty();
-    reload_targets.emplace_back( std::move( opt.target ) );
-    reload_targets.emplace_back( std::move( opt.ammo ) );
+    target_loc = std::move( opt.target );
+    ammo_loc = std::move( opt.ammo );
 }
 
 bool reload_activity_actor::can_reload() const
 {
-    if( reload_targets.size() != 2 || quantity <= 0 ) {
+    if( quantity <= 0 ) {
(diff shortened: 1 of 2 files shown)


src/activity_actor.cpp
```diff
@@ -4010,6 +4010,14 @@ std::unique_ptr<activity_actor> insert_item_activity_actor::deserialize( JsonVal
     return actor.clone();
 }
 
+reload_activity_actor::reload_activity_actor( item::reload_option &&opt, int extra_moves )
+{
+    moves_total = opt.moves() + extra_moves;
+    quantity = opt.qty();
+    reload_targets.emplace_back( std::move( opt.target ) );
+    reload_targets.emplace_back( std::move( opt.ammo ) );
```
mqrause, line 4018: Would be great if you also got rid of the `reload_targets` vector, that's probably just a holdover from the old activity system.

irwiss: Done in https://github.com/CleverRaven/Cataclysm-DDA/pull/61122/commits/5190a74c6d3833f4e3c8e245c99e97af3ff92bb8 if you'd like to review

PR #61116, opened 2022-09-19
Fix MinGW build

dseguin:
#### Summary
None

#### Purpose of change
MinGW build currently fails to compile (see https://github.com/CleverRaven/Cataclysm-DDA/pull/61102#issuecomment-1250448532)

#### Describe the solution
Check for `_WIN32` when defining the `locale_type` wrapper

#### Testing
Let the MinGW CI test run

PR #61124, opened 2022-09-19
Fix mingw locale independent build for real.

akrieger:
#### Summary
None

#### Purpose of change
@dseguin had the right idea in #61116 but it needed just a bit more. Flatbuffers already had the right infra to choose the right locale independent functions for windows/not windows, we just needed to make the same detection logic change there.

#### Describe the solution
Check for _WIN32 both in the independent locale block and for choosing the parsing functions. mingw apparently doesn't have the 'float' version of the parsers, but there is already the pattern of casting the double parser to float below, so replicate that too.

#### Testing
https://github.com/akrieger/Cataclysm-DDA/actions/runs/3084417759/jobs/4986587088 actually builds. This PR CI run should get through the mingw build step.


akrieger:
Cross build passed \o/

Issue #59657, opened 2022-07-27
Exosuit features no longer work


Drew4484:
### Describe the bug

Something in the last few weeks broke most of the exosuit functionality. Flags and relic passive effects no longer work when activated. Transforms still work. 

### Steps to reproduce

1. Spawn an exosuit frame and power supply.
2. Activate exosuit.
3. Note that exosuit is active as vision is no longer obstructed.
4. Check player stats, +10 strength bonus is not visible. 
5. Notice you are freezing to death as the climate control is not functioning.

Also impacts modules like the small and large load support modules, which can be reproduced easily.

### Expected behavior

Activating a module or the frame itself should provide passive bonus effects.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.1826 (21H2)
- Game Version: 2fbf47c [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Aftershock [aftershock],
    Aftershock: Exoplanet [aftershock_exoplanet]
]

### Additional context

Also happens on non-exoplanet worlds, see below
- OS: Windows
    - OS Version: 10.0.19044.1826 (21H2)
- Game Version: 2fbf47c [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Aftershock [aftershock]
]

Drew4484:
Confirmed by @irwiss and @Merikona

Drew4484:
First noticed the issue while working on #58670, so approximately 06/24/2022 for a start date.

Maleclypse:
I can't make it work on [Cataclysm-DDA experimental build 2022-05-02-2346](https://github.com/CleverRaven/Cataclysm-DDA/releases/tag/cdda-experimental-2022-05-02-2346) and I tested with saving and reloading to see if it required a reload like some errors do occasionally.

Maleclypse:
irwiss — Today at 6:19 AM
[...]

PR #61139, opened 2022-09-20
Convert CLIMATE_CONTROL to enchantments and simplify effect

actually-a-cat:
#### Summary
Balance "Climate Control CBM (and equivalent effects) now works more consistently, doesn't completely shut down in extreme temperatures, and can stack"

#### Purpose of change
CLIMATE_CONTROL has some problems:

* It snaps body temperature to fixed values, so the actual effect provided can vary dramatically depending on exact temperature.
* It suddenly stops working when the temperature gets too high or too low, so you can go from uncomfortable but fine to at risk of injury with a 1 degree temperature change
* It always has the same strength, regardless of source
* It doesn't stack, even in situations where it would make sense 
* The cooling ability and heating ability are always equal even though the mechanisms for achieving those are very different

#### Describe the solution
Climate control is now a pair of enchantments, controlling the heating and chilling ability respectively. They simply move body temperature towards normal by up to a fixed value. For now I set it to 50 everywhere, which is close to the previous performance of climate control, and most sources received equal heating and cooling. Exceptions are some Aftershock items that are explicitly for protection against cold and those only have heating.

#### Testing
Checked:
- [x] Basic functionality
- [x] Stacking of multiple sources
- [x] Climate control bionic heat/chill
- [x] RM13 armor heat/chill
- [x] Advanced nomad jumpsuit heat/chill
- [x] Combat exoskeleton heat/chill
- [x] Heavy combat exoskeleton heat/chill
- [x] Field combat exoskeleton heat/chill
- [x] Nomad plate heat/chill
- [x] Light nomad plate heat/chill
- [x] Heavy nomad plate heat/chill
- [x] Phase immersion suit heat/chill
- [x] Exosuit frame heat/chill
- [x] Magellan exosuit heat
- [x] Frontier cryo suit heat
- [x] Heat retention cream heat
- [x] Heat retention xtreme cream heat
- [x] Cryopod bodyglove heat/chill
- [x] Vehicle CC heat/chill and stacking with enchantment CC

Climate control on exosuit underlayers doesn't work due to #59657, but they have the definitions and should work once that's fixed.

#### Additional context
Remaining problems that will require more substantial changes to address:
* Worn items apply climate control to entire body, regardless of coverage
* No ability to vary power consumption depending on how much work climate control is actually doing

----
cc_chart

### DIFF
--- a/data/json/bionics.json
+++ b/data/json/bionics.json
@@ -602,7 +602,12 @@
     "react_cost": "100 J",
     "time": "1 s",
     "flags": [ "BIONIC_TOGGLED" ],
-    "enchantments": [ { "condition": "ACTIVE", "values": [ { "value": "CLIMATE_CONTROL", "add": 50 } ] } ]
+    "enchantments": [
+      {
+        "condition": "ACTIVE",
+        "values": [ { "value": "CLIMATE_CONTROL_HEAT", "add": 50 }, { "value": "CLIMATE_CONTROL_CHILL", "add": 50 } ]
+      }
+    ]
   },
   {
     "id": "bio_cloak",
--- a/data/json/enchantments.json
+++ b/data/json/enchantments.json
@@ -25,7 +25,11 @@
     "type": "enchantment",
     "id": "RM13",
     "condition": "ACTIVE",
-    "values": [ { "value": "FOOTSTEP_NOISE", "multiply": -0.67 }, { "value": "CLIMATE_CONTROL", "add": 50 } ]
+    "values": [
+      { "value": "FOOTSTEP_NOISE", "multiply": -0.67 },
+      { "value": "CLIMATE_CONTROL_HEAT", "add": 50 },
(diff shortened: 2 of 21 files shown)


Maleclypse:
I'd much prefer if climate control's power costs went up with the temperature difference.  Though I guess that might be out of scope for this PR.

dseguin:
Looks like this would fix the climate control problem in #59657 (might even resolve the issue)

actually-a-cat:
@dseguin 

This will make climate control work in extreme cold but the effect won't be strong enough for an exosuit - it's equivalent to ambient temperature changing by 25C/45F.

Ideally this would be adjustable for every source of the effect but I'm not sure how to do that. Make it an enchantment?

actually-a-cat:
I'm going to go ahead and JSON-ify climate control while I'm at it.

dseguin (approved):

PR #61145, opened 2022-09-20
Magiclysm Purified and Impure Meat Inheritance Error

Morat2255:
#### Summary
Mods "Magiclysm: Purified and Impure Meat should extend flags instead of override"

#### Purpose of change
Magiclysm adds recipes for Purified and Impure Meat. By the in-game description and they copy-from line in json, these appear to be intended to be identical to Meat and Mutant Meat respectively (the description of Purified Meat even says "Indistinguishable from pre-Cataclysm lab grown beef." However, in the json, it sets the flags for both items to NUTRIENT_OVERRIDE. I believe that this is intended to add that flag, rather than replace the list of flags in the item being copied. As a result, Purified Meat is unable to be put in a smoking rack (despite having a smoking_result line inherited from Meat), cannot gain calories from cooking despite having the same calories as raw meat, and doesn't give additional fun to Predator mutants.

#### Describe the solution

Changes the "flags": [ "NUTRIENT_OVERRIDE" ] line to "extend": { "flags": [ "NUTRIENT_OVERRIDE" ] }, for both purified and impure meat, allowing it to inherit the flags from meat and mutant meat.

Writing this as an issue and having someone else change two lines. Continuing to avoid practicing using github.

#### Testing

Made a new world, spawned in a smoking rack and purified and impure meat, smoking rack accepts both, because they correctly have the SMOKEABLE flag. Put them in and smoke them, wait 6 hours, they correctly turn into their smoked versions, calories have increased due to the original item having the RAW flag.

#### Additional context

It still isn't possible to simply cook Purified or Impure Meat, because the Cooked Meat recipe asks for the specific item "Meat" and has different recipes for regular meat, human meat, humanoid meat, and mutant meat, rather than using meat_red_raw list in the recipe and conditional_names to differentiate the types, like all (at least most?) other meat recipes. I could extend the recipe maybe? Or just change it to what's standard for all the other related recipes? But I don't want to touch that without knowing why it's different to the others, and that's beyond the scope of this PR anyway.

### DIFF
--- a/data/mods/Magiclysm/items/comestibles.json
+++ b/data/mods/Magiclysm/items/comestibles.json
@@ -79,8 +79,8 @@
   {
     "id": "purified_meat",
     "type": "COMESTIBLE",
-    "flags": [ "NUTRIENT_OVERRIDE" ],
     "copy-from": "meat",
+    "extend": { "flags": [ "NUTRIENT_OVERRIDE" ] },
     "name": "purified meat",
     "description": "Indistinguishable from pre-Cataclysm lab grown beef.  It should be excellent but somehow it's just edible.",
     "price": 400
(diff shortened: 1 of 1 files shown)


Maleclypse:
Conditional names and component lists didn’t exist when this was added iirc so making the more extensive change to make them more useful is fine.

Maleclypse (approved):

PR #61151, opened 2022-09-20
Unarmed gauntlets and boots

Dan1ss1mo:
#### Summary
Bugfixes "abjust some armor for new unarmed combat system"

#### Purpose of change

A lot of armor pieces that are saposed to be sturdy break very fast then used in unarmed combas as they lack durable_melee
tag. On top of that some armore bashing value seems to be unfitting for items. Getting hit by someone with chain gloves should definitely hurt a lot more than by bare hand. Hits by someone wearing power armore shuld be atleast as deadly as hit with pladed gloves and probably even more powerful than its active

#### Describe the solution
the easyest solution (for someone how doesnt understand much in source code) would be to add durable melee flag to all sturdy clothing, becouse if somthing can take a lot of hits it shouldnt fall apart then you smacking someone with it

Much more rasonable solution would be to add properties of durable_melee tag to sturdy flag and vice versa(for things like knuckles) but it would requare changes in sorce code that im not capable of.

#### Testing
summon a debug monster and smash him(her?) with differnt gloves/gauntlets and see how fast they will break

#### Additional context
It would also be nice do add a doc which would contain general guideline for how much bash damage differnt clothing should provide. Even better solution would be to automatically assign this values based on item material


Zireael07:
The best solution would be to add something that is like durable melee but only for unarmed gauntlets. Your changes mean that if someone wields the gloves instead of using them in unarmed style, they'll be treated as durable too (And they're definitely not meant to be used to bash people or things willy-nilly, it's not a mace)

Dan1ss1mo:
> The best solution would be to add something that is like durable melee but only for unarmed gauntlets. Your changes mean that if someone wields the gloves instead of using them in unarmed style, they'll be treated as durable too (And they're definitely not meant to be used to bash people or things willy-nilly, it's not a mace)

that would also work, i just would like to avoid manualy editing all existing items. Also, not only gloves should be changed as with new attack vectors system you can use differnt body parts in unarmed combat

Dan1ss1mo:
also, i dont think it would be that bog of a deal even if it would work as you describe becouse with out ability to use matial art they would just be really shity bashing weapons

Dan1ss1mo:
@Zireael07 I don't think  wielding any king of gauntlets over somthing like a quarterstaff or club would be better just becuase they are durable, but i agree that adding a specific tag for such items would be overall better solution. I just don't think it worth a hassle of overwriting dozens or even hundreds of existing items.

data/json/items/armor/boots.json
```diff
@@ -2286,9 +2285,10 @@
     "price": 5012,
     "price_postapoc": 2000,
     "to_hit": -1,
+    "bashing": 4,
     "symbol": "[",
     "color": "light_red",
-    "flags": [ "VARSIZE", "STURDY", "OUTER" ],
+    "flags": [ "VARSIZE", "STURDY", "OUTER", "DURABLE_MELEE" ],
```
nopenoperson, line 2291: this shouldnt have durable melee since the metal is poor quality

nopenoperson (changes requested):

data/json/items/armor/boots.json
Dan1ss1mo: fair enough

kevingranade:
> Getting hit by someone with chain gloves should definitely hurt a lot more than by bare hand.

Why? It's not significantly heavier than an open hand,  and it's flexible so it doesn't distribute force across the hand.

Dan1ss1mo:
> > Getting hit by someone with chain gloves should definitely hurt a lot more than by bare hand.
> 
> Why? It's not significantly heavier than an open hand, and it's flexible so it doesn't distribute force across the hand.

I assume it would  be somewhat like being hit with the cestus, but i may be wrong on that. I was mainly using leather gloves\boots as a guideline, tho maybe the damage bonus on them it too high?

PR #61153, opened 2022-09-20
add DURABLE_MELEE flag for powered quarterstaff and ironshod quarterstaff

Dan1ss1mo:
#### Summary
Balance "add DURABLE_MELEE flag for powered quarterstaff and ironshod quarterstaff"

#### Purpose of change
Currently there is only a single staff in the game with durable flag, which is a bo. I think that staff that had been reinforced with metal bands and caps shuld be at least as durable as a long stick treated with oil.

#### Describe the solution
Add DURABLE_MELEE flag to mentiond weapons

Also change ironshod quarterstaff to require bo, instead of regular quarterstaff(already did it)

### DIFF
--- a/data/json/recipes/weapon/bashing.json
+++ b/data/json/recipes/weapon/bashing.json
@@ -827,7 +827,7 @@
       { "proficiency": "prof_blacksmithing" },
       { "proficiency": "prof_toolsmithing" }
     ],
-    "components": [ [ [ "q_staff", 1 ] ] ]
+    "components": [ [ [ "bo", 1 ] ] ]
   },
   {
     "result": "knuckle_impact",


nopenoperson:
the bo recipe  takes more than 4 hours to make while quarterstaff into ironshod takes less than 1 hour

Dan1ss1mo:
@nopenoperson but ironshod require higher fab level and metalworking equipment while bo require some easy accessable tools and some oil\tallow. Plus, as i mention, i may also change ironshod to require bo instead of regular quarterstaff

nopenoperson:
maybe the recipe for quarterstaves should reflect why they are called quarterstaves, when you quarter wood length wise after felling a tree you get better structural integrity because of the uniform direction of the wood grain.

tl;dr make quarterstaves durable and have them require a log?

Dan1ss1mo:
Apparently it doesn't, but it also a LOT thicker, so i guess keeping roughly the same durability for much thinner stave steel counts as been more durable than a regular stick

nopenoperson:
this pr is probably a good idea because the ironshod quarterstaff can break really quickly even for survivors with very high skills

Dan1ss1mo:
This is exactly the reason why i made it. On of the first martial art that i found was karate and naturaly a decided to try ironshod quarterstaff. I was very disappointed then i found out that despite having such good stats(although i dont understand why it has lower +hit bonus, but +2 is fine) it breaks extremely fast. I eventually just discarded it in favour of my trusty fire axe, as i dont have to wrap in in duct tape after every encounter.

PR #61155, opened 2022-09-20
draped over legs aprons

anoobindisguise:
#### Summary
Balance "shift aprons to the draped-over-legs layer"

#### Purpose of change
Previously it was difficult to wear an apron with say, a hip flask, or holster, due to their layering. now that there is a draped layer this is the ideal place to put it. It will also mean you cannot wear it with a wizard robe or whatever without penalties, which I think more or less mirrors what makes sense IRL.

#### Describe the solution
move from hips/thighs to draped over legs

### DIFF
--- a/data/json/items/armor/torso_clothes.json
+++ b/data/json/items/armor/torso_clothes.json
@@ -14,7 +14,13 @@
     "delete": { "flags": [ "WATER_FRIENDLY" ] },
     "armor": [
       { "covers": [ "torso" ], "specifically_covers": [ "torso_waist" ], "coverage": 100, "encumbrance": [ 1, 1 ] },
-      { "covers": [ "leg_l", "leg_r" ], "specifically_covers": [ "leg_draped_l", "leg_draped_r" ], "coverage": 70, "encumbrance": 2, "volume_encumber_modifier": 0.3 }
+      {
+        "covers": [ "leg_l", "leg_r" ],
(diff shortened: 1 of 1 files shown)


data/json/items/armor/torso_clothes.json
```diff
@@ -115,16 +121,10 @@
       { "covers": [ "torso" ], "specifically_covers": [ "torso_waist" ], "coverage": 80, "encumbrance": [ 0, 0 ] },
       {
         "covers": [ "leg_l", "leg_r" ],
-        "coverage": 70,
-        "specifically_covers": [ "leg_hip_l", "leg_hip_r" ],
+        "coverage": 22,
+        "specifically_covers": [ "leg_draped_l", "leg_draped_r" ],
```
Maleclypse, line 125: I'd remove leg coverage entirely.  This is like a servers apron in length but missing the pockets.

LyleSY: Googling around on this the items I see do appear to cover the top of the leg, even the shortest ones
like this https://www.webstaurantstore.com/choice-black-standard-waist-apron-with-3-pockets-12l-x-26w/160WAISTFHBK.html?utm_source=google&utm_medium=cpc&utm_campaign=GoogleShopping&gclid=CjwKCAjw7eSZBhB8EiwA60kCW5in744IGX9fbKHaZmrNeTDs1S9AwnDjZoYfDs3_zqTXpYFwvkLl1hoCyE4QAvD_BwE

PR #61156, opened 2022-09-20
Kevlar wetsuit tweaks

anoobindisguise:
#### Summary
Balance "updates to Kevlar wetsuit"

#### Purpose of change
Kevlar wetsuits weren't SKINTIGHT despite being made of a skintight wetsuit. Also, there was no variant for the thick wetsuit.

#### Describe the solution
Splits kevlar wetsuit into two options: regular and thick, made with the respective wetsuits. They occupy SKINTIGHT, NORMAL layers both instead of just the NORMAL layer. Environmental protection lowered a bit. Warmth and encumbrance increased (regular kevlar wetsuit had less encumbrance than its constituent wetsuit). Crafting made slightly more complicated, as you now need silicone caulk to act as a sealant.


nopenoperson:
silicone caulk wouldnt stick to neoprene foam. super glue would work a lot better. blindstiching with a curved needle with superglue and tape to fix any horrible accidents is way more realistic

nopenoperson:
rubber cement would also work

anoobindisguise:
> You posted this in ready to merge but it's still in draft.

Sorry! I forgot about that when editing the material.

data/json/recipes/armor/bespoke_armor/survivor.json
```diff
@@ -49,35 +49,69 @@
     "difficulty": 7,
     "time": "12 h",
     "autolearn": true,
-    "using": [ [ "sewing_kevlar", 200 ], [ "fastener_large", 1 ], [ "clasps", 8 ] ],
+    "using": [ [ "sewing_kevlar", 200 ], [ "tailoring_neoprene_patchwork", 30 ], [ "fastener_large", 1 ], [ "clasps", 8 ] ],
```
ZhilkinSerg, line 52: You have removed kevlar from the list of the components, so resulting item should not be called "Kevlar" anymore or components list should be fixed (same for recipes below).

PR #61164, opened 2022-09-20
Vehicle water faucet fixes

### LINKED ISSUE #45710: Refrigerated tank (60L) isn't cooling liquids
# Describe the bug
Using a refrigerated tank doesn't actually cool liquids and instead functions as a standard tank only with a power draw, maybe the insulation feature still works with it. In my current playthrough I'm just getting to the end of summer, water stayed hot inside of the refrigerated tank, not once have I been prompted to fill up with cold water.

# Steps To Reproduce
1. Set season to summer, spawn a vehicle, give hammerspace mutation.
2. Install refrigerated tank, fill with water.
3. Wait a few days with the tank turned on.
4. Fill a container with liquid and see that it's still water (hot)

# Expected behavior
Refrigerated tank should keep liquids inside cold.

# Versions and configuration
- OS: Windows
    - OS Version: 10.0 2009
- Game Version: 0.E-7886-g76fd581 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda]
]

irwiss:
#### Summary
None

#### Purpose of change

Fixes #45710

#### Describe the solution

Use the vpart and the liquid item instead of just item type when siphoning or drinking, this removes the need to do temperature overrides and makes UI correctly display (cold).

There are also various assumptions made about dealing with liquids using just the itype_ids that aren't correct; vehicle::fuel_specific_energy assumes temperatures of all liquids of same itype_id in all tanks can be averaged.

~~Tried using item_location + vehicle_cursor for consume_activity_actor, but seems it either deletes the item twice or has issues finding the item in the tank if it's the last charge drained and triggers a debugmsg.~~ apparently derped the item_location setup, testing solution from mqrause.

#### Testing

Before applying patch:
Load attached save for convenience or repeat steps; spawn RV, install refrigerated tank, spawn 1 empty jerrycan, 10 bottles of clean water, fill tank with one bottle (refrigerated tank should have 20 water), turn on refrigerator, pass a few days until water becomes cold, siphon jerrycan, reload a bottle, use bottle to fill RV, you should have: 2 clean water (cold) in RV, 18 clean water (cold) in jerrycan.

1) e the kitchen unit, "fill container", "pour into container", into jerrycan - see you have 40 clean water(cold) - extra 20 water spawned out of nowhere. Siphon doesn't limit itself as it was given "free" water item, rather than the water in the tank.
2) e the kitchen unit, "drink water", v to see morale - see you got no bonus from drinking cold water. "drinking" from faucet was giving consume_activity_actor item generated from itype_id without temperature.
3) fill the RV with water from jerrycan, start drinking until you can't drink anymore, continue drinking and cancelling - see water gets drained even if you cancel force drinking query.

Apply patch;

Repeat 1: jerrycan should only fill with water in the tank
Repeat 2: should see +1 "enjoyed clean water" from cold bonus
Repeat 3: once you're too full to drink the option to drink should be disabled

#### Additional context
[Test AFS Mgc XED.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/9610826/Test.AFS.Mgc.XED.zip)

### DIFF
--- a/tests/vehicle_part_test.cpp
+++ b/tests/vehicle_part_test.cpp
@@ -26,6 +26,7 @@
 #include "requirements.h"
 #include "type_id.h"
 #include "units.h"
+#include "veh_utils.h"
 #include "veh_type.h"
 #include "vehicle.h"
 #include "vpart_position.h"
@@ -40,6 +41,7 @@ static const itype_id itype_fridge_test( "fridge_test" );
 static const itype_id itype_metal_tank_test( "metal_tank_test" );
 static const itype_id itype_oatmeal( "oatmeal" );
 static const itype_id itype_water_clean( "water_clean" );
+static const itype_id itype_water_faucet( "water_faucet" );
 
 static const recipe_id recipe_oatmeal_cooked( "oatmeal_cooked" );
 
@@ -143,9 +145,9 @@ static void test_craft_via_rig( const std::vector<item> &items, int give_battery
     character.learn_recipe( &recipe );
 
     get_map().add_vehicle( vehicle_prototype_test_rv, test_origin, -90_degrees, 0, 0 );
-    const optional_vpart_position &ovp = get_map().veh_at( test_origin );
-    vehicle &veh = ovp->vehicle();
+    const optional_vpart_position ovp = get_map().veh_at( test_origin );
     REQUIRE( ovp.has_value() );
+    vehicle &veh = ovp->vehicle();
 
     REQUIRE( veh.fuel_left( itype_water_clean, true ) == 0 );
     for( const vpart_reference &tank : veh.get_avail_parts( vpart_bitflags::VPFLAG_FLUIDTANK ) ) {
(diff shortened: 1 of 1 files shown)


mqrause:
> Tried using item_location + vehicle_cursor for consume_activity_actor, but seems it either deletes the item twice or has issues finding the item in the tank if it's the last charge drained and triggers a debugmsg.

Kinda curious about this. Consume activity should handle reducing charges/removing items on it's own. So if you remove the last `if`, I would think it should just work?

irwiss:
No, since I couldn't make item_location work and explicitly gave it the water item's copy and then manually subtracted the charges.

I tried using item_location, here's the [patch with the test](https://github.com/irwiss/Cataclysm-DDA/compare/master...748b70b8e6caddce00f849772565b282fb55bcbb.patch) if you wanna try it out.

Load the attached save, it'll work right (as in, consumes water subtracting the charges) until you drink the last water charge, it then reaches the [debugmsg here](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/visitable.cpp#L550) with this call stack:
```
>	visitable::remove_item(item & it) Line 550	C++
 	item_location::impl::item_on_vehicle::remove_item() Line 537	C++
 	item_location::remove_item() Line 923	C++
 	Character::consume(item_location loc, bool force, bool refuel) Line 1875	C++
 	consume_activity_actor::finish(player_activity & act, Character & __formal) Line 2594	C++
 	player_activity::do_turn(Character & you) Line 377	C++
 	do_turn() Line 664	C++
 	WinMain(HINSTANCE__ * __formal, HINSTANCE__ * __formal, char * __formal, int __formal) Line 793	C++
 	[External Code]	
```

mqrause:
> So i guessed it either deletes the water item twice or fails to look it up (or i derped something when building the cursor/item_location)

Ah, yeah, the issue is that the water isn't in the vehicle part but in the base item of the vehicle part. So the `item_location` would need to have the base item as the parent, which might be a bit sketchy. But at least I think something like this should work:
```cpp
item_location base( vehicle_cursor( *this, vp_tank_idx ), &vp_tank.base ) );
item_location water( base, &vp_tank.base.only_item() );
```
No idea about potential issues arising from exposing the base item like this, though.

irwiss:
Ah so I did derp the setup, not sure about the exposing but I'll playtest it a bit and see

mqrause:
Actually thinking about it, I'm not sure the `item_location base` is really correct like. Vehicle cursor locations probably only really work for items in `vehicle_part::items`. It should be good enough for removing the last charge at least, but it's definitely something to keep in mind for the future.

irwiss:
I added a comment near the code and added a test to check it in case item_location behavior changes

PR #61165, opened 2022-09-20
Reasonable launchers practice (beginner)

RenechCDDA:
#### Summary
Content "Basic launchers practice recipe"

#### Purpose of change
Launchers is in a weird place. Without equipment, you can't train it, and without training, you can't make any equipment. Even using a military grenade launcher in a simple point-towards-enemy manner takes launchers 2!

#### Describe the solution
This PR aims to bridge the gap, by introducing a practice routine that the survivor could reasonably carry out and actually *learn* from.  Judging distance, getting familiar with the feel of a standalone launcher, and eyeballing angles for launch can all be practiced with no risk of blowing oneself up.

I considered making the launcher an actual weapon, but considering the forces involved it would not be useful. Keeping it as a simple tool makes it much easier.

#### Testing
Modified locally, carried out the practice and didn't notice anything unexpected.

#### Additional context
~~The novelty pneumatic launcher should probably be seeded into some default itemgroups as fluff/clutter, instead of only being available as a craft.~~ It has been added to some item groups!


Fris0uman:
Is the launcher like those shirt launcher?
image

Could spawn in close shops of malls, dollar stores and maybe  stadiums

RenechCDDA:
Yes, and that's a very good image! I'm updating the PR to include it in some item groups, and should have a commit in shortly after this comment.

PR #61172, opened 2022-09-21
Chargen only proficiency lets you dissect humans without penalty

anoobindisguise:
#### Summary
Balance "chargen-only proficiency allows dissection of humans without morale loss"

#### Purpose of change
~~The physiology proficiency represents extensive familiarity with human biology, and is either learned by slaughtering hundreds of feral humans, preforming dozens of autopsies, or before the cataclysm in formal higher education, likely also working with human cadavers in that context.~~ Given that people who are trained for autopsies IRL don't become highly depressive for days on end after performing an autopsy it seems reasonable that someone with training could "dissect" a human corpse, but there's a difference to be made between doing that and outright butchery (chopping up the corpse for meat)

#### Describe the solution
There are five cases when attempting to butcher a human now.
Case 1:  The player is a psychopath, cannibal, or flying purple people eater (sapiovore). If so none of the new logic applies, it is discarded.
Case 2: The player is attempting to butcher the corpse, and they don't have prof_dissect_humans. This proceeds exactly as it did before. They get very sad.
Case 3: The player is attempting to butcher a corpse and they do have prof_dissect_humans. In this case they get some special dialogue, acknowledging the fact that they are outright butchering the corpse instead of simply dissecting it. They get very sad.
Case 4: The player is attempting to dissect the corpse, and they don't have prof_dissect_humans. They get special text to reflect this and get a somewhat lower penalty than outright butchery. 
Case 5: The player is attempting to dissect the corpse and they do have prof_dissect_humans. In this case there is no morale penalty and they get special text acknowledging that it's unpleasant but bearable.

For some reason dissections seem to be providing meat, I don't know why this is or if it's a bug, nor do I know whether or not it should be fixed. Previously they just gave a pile of butchery refuse.
We could also use a different proficiency than physiology, but right now it appears to be the "living humans" weakpoint proficiency. EDIT: Did this!

#### Testing
Compiled and tested to make sure all 5 scenarios play out correctly.


Venera3:
Acclimating yourself to autopsies is a process taking multiple months and *extremely* contingent in the social context of it - ~8 hours of cutting people up alone in the woods without reason, alone is a very far cry from that.

PatrikLundell:
Comment on the side comment: It is reasonable that dissecting a corpse (of whatever creature) would still produce recoverable products, as I doubt dissecting involves cutting every muscle into a paste, cut all tendons to shreds, and pulverize every bone. but organs are probably cut up. It ought to result in a lot less chunks of meat (possibly zero), with much of it turned into scraps.

The easy way would be to just consider the corpse pulped when determining the butchery results, but otherwise the results could be either a dissected corpse (that could then be butchered), or a special dissection yield (that may or may not be reduced further by the original damage level of the corpse).

Dan1ss1mo:
@Venera3 considering that by the time you get this proficiency you will probably kill a few hundreds zombies and a few dozens ferals you're character should be very accustomed to the sight of human(or decayed human) insides, so it shouldn't be that big of a deal. Especially if mentioned "human" try to beat you to death with crowbar or chop with axe

Dan1ss1mo:
@Fris0uman the moral penalty actually encourage you to NOT dissect every human\feral corpse that you find. The point is that you eventually get used do dissecting humans, like with killing zombie childs, etc

Venera3:
"Zombie dissection is a freebie" is the incongruence here, not "dissecting people is too hard".

Dan1ss1mo:
@Venera3 in that case it just wrong, most people would probably throw up even from looking at freshly butchered animal corpse. Now imagine performing autopsy of partially decomposed rotten human body. If anything, i think it should give even higher penalty the dissecting a feral

anoobindisguise:
In any caae the existing code could be repurposed to use a special autopsy proficiency that only medical professions get.

Dan1ss1mo:
I mean, you probably wouldnt dissect ferals for the sake of getting immunity to morale debuff for dissecting ferals. You probably doing it for human sample and you will continu to do it until you get desirable amount even if that give you penalty. The point is that youre character eventually getting used to this as you continu to do it. I don't understand why is it ok to get used to kill zombie children but not to dissect humans. Also the moral penlty is long enough that you cant sleep it of in a single night

Dan1ss1mo:
i think that the approach that you mention is already the optimized play. in order to avoid morale penalty getting out of control, i curntly try to dissect one or two corpses and wait until i get rid of penalty

Venera3:
> In any caae the existing code could be repurposed to use a special autopsy proficiency that only medical professions get.

A two-level dissection prof (second unattainable outside of starts) to govern the drop rates instead of healthcare is something I want to hook up before stable, I'm not principally opposed to the lower level dampening the morale effects slightly. If you want to hook the morale part up that would be nice, otherwise a single prof for morale immunity works until then.

anoobindisguise:
I still think the base logic here is worth inclusion, even if we don't use a trainable Physiology proficiency and instead include a specific autopsy proficiency that a few professions get (medical residents for example)

Venera3:
A chargen-only immunity prof would be fine to have, yes.

PR #61173, opened 2022-09-21
Add a new zombie: the Malicious Mane

MNG-cataclysm:
#### Summary
Content "Add a new zombie: the Malicious Mane"

#### Purpose of change
Add a new zombie: the Malicious Mane.

#### Describe the solution
The malicious mane is just a zombie in a fursuit. It will rarely spawn as part of normal zombie groups, and drops one of the five fursuits on death in addition to normal clothing.

Not adding this, or moving it to the Crazy Cataclysm mod.

#### Testing
I spawned the zombie, killed it, looted it, and searched through a city for it. Everything works as I intended, with no errors.

#### Additional context
None.


MNG-cataclysm:
I added anthropomorphized to the dictionary.

Amoebka:
Is there any particular reason a torn fursuit offers 3 times more armor than the pristine version players can wear?

PatrikLundell:
I'd rather see it moved to a mod, as there's an extremely small chance of such people moving about in the open, in particular donning a suit when the world ends.

MNG-cataclysm:
> Is there any particular reason a torn fursuit offers 3 times more armor than the pristine version players can wear?
> For reference, it has higher armor than zombie cops, while also being faster and doing more damage. This feels like a common sense violation.

That was an oversight on my part.

Amoebka:
Flavor-wise I see these mostly as corporate mascots, rather than furries. I think they are fine for mainline if sufficiently rare.

MNG-cataclysm:
I dropped the zombies armor from ten to seven.

Jarewill:
Personally I don't mind seeing this zombie in game, but I feel some changes should still be made.

7 bash and cut armor still makes them tougher than zombie cops, which I feel shouldn't be the case.
I think 4 bash, cut armor should be fine, while ballistic armor should maybe go to 2?

Them being as common as SWAT zombies also I feel is too much.
I would lower the weight (assuming that means chance) to 1 as I don't think people walking around in fursuits is that common.

Also they are missing the wolf suit in the drop table.

LovamKicsiGazsii:
I'm personally not against this addition either. I've been meaning to make a convention center location that might make good use of this new enemy.

MNG-cataclysm:
@Jarewill I applied the changes you suggested.

Jarewill:
When I said 3/4 I meant "3 or 4", but that's my fault for the confusion.

Anyways yeah, I think the stats are fine now.
They have as much bash armor as fat zombies while having more cut armor but less health and speed.

kevingranade:
This needs to be at least an order of magnitude (10x) rarer.

MNG-cataclysm:
@kevingranade Will reducing the cost multiplier by one do that?

Maleclypse:
> @kevingranade Will reducing the cost multiplier do that?

No, at this point you'd need to multiply every other weight in the monster group by 10.  So other weight 1's would become weight 10 etc.

MNG-cataclysm:
@Maleclypse I've fixed the order of magnitude. The malicious mane is now 10x rarer than before.

I-am-Erk:
It seems to me this is still mostly a base zombie with a bit of light armour, it should probably use copy-from?

MNG-cataclysm:
> It seems to me this is still mostly a base zombie with a bit of light armour, it should probably use copy-from?

What would it copy from? I can’t think of a good candidate, or anything like it.

I-am-Erk:
Probably just a regular zombie. What are its key differences besides a bit of armour and its dropped items?

MNG-cataclysm:
> Probably just a regular zombie. What are its key differences besides a bit of armour and its dropped items?

Well, the drops are mainly false fur, plastic, and felt, in place of more protective Kevlar or steel.

MNG-cataclysm:
@I-am-Erk I’ve added a copy from entry to the malicious mane.

LyleSY:
I'm good with the broad concept but I think this should be frozen for next experimental

PR #61174, opened 2022-09-21
2StoryModern03

LilShiningMan:
Add 2StoryModern03 house

#### Summary
Content "Add House - 2StoryModern03"

#### Purpose of change

Add to house spawn variety

#### Describe the solution

created new house and added entries to spawn files

Adding nothing.

#### Testing

Checked house in-game, spawns just fine.  Linted.

#### Additional context
base

gr

roof

### DIFF
--- a/data/mods/No_Hope/palettes.json
+++ b/data/mods/No_Hope/palettes.json
@@ -93,7 +93,7 @@
       "|": "t_wall_w",
       "#": "t_brick_wall",
       ":": "t_wall_glass",
-	  "¶": "t_door_glass_c",
+      "¶": "t_door_glass_c",
       "-": "t_sidewalk",
       "o": [
         [ "t_window_frame", 5 ],


Maleclypse:
@anothersimulacrum Am I missing something here?  The errors in this can be fixed by adding `"¶": "t_door_glass_c",` in the terrain section of mapgen, but `"¶": "t_door_glass_c",` is already defined in the palette and works fine defined that way in multiple already extant houses in game.

LilShiningMan:
oops. Tried to fix this, but added the symbol to basement instead of ground and 2nd floor, which will fix the issue if it comes to that. Before I do that though, it would be interesting to know if the reason this throws an error can be deduced by someone.

LilShiningMan:
And, to add context, I tried adding this without defining that symbol to see if the error repeated as it did when I submitted 2storymodern02, and of course it did. However, it only seems to be an error in git. The screenshots show the screen door spawning successfully by calling the palette and they work as expected in-game.

dseguin:
The test fails when it tries to load this with mods (most likely No Hope, since it redefines the palette and it's missing the `¶` terrain symbol).

Try adding the `¶` terrain symbol to `standard_domestic_palette` in `data/mods/No_Hope/palettes.json` and see if that fixes the tests

LilShiningMan:
Funnily enough, I had this exact scenario over in a PR in Bright Nights and someone else just mentioned the same thing. Thanks, I'll add the symbol to the No Hope palette as well.

PR #61176, opened 2022-09-21
Replace Railgun CBM

Drew4484:
#### Summary
Content "Replaces railgun CBM"

#### Purpose of change
The railgun CBM is one of the most nonsensical bionics. It somehow uses magnets in combination with the player's throwing action to accelerate objects and leaves a trail of sparks behind it. If the magnets are propelling the projectile, why does it need to be thrown? If they are making the arm throw faster, how does that work with the structure of the arm, as railguns are famous for having, well, rails. The sparks are completely out of place.

This adds a new CBM, flavored as a bionic from a world that cyberized sports, adding a throwing assist bionic that adds a passive bonus to throwing skills.

#### Describe the solution
Obsoletes the Railgun CBM and removes it from spawn lists. Add new Throwing Assist CBM to spawn lists.

Mainlining the Aftershock grenade launcher CBM as a partial replacement. 

#### Testing
Spawned Throwing Assist CBM, installed it, noted Throwing skill went from 0 to 5.

#### Additional context
Using the changes in #61539, the replacement bionic will add a skill modifier to throwing.

### DIFF
--- a/data/json/bionics.json
+++ b/data/json/bionics.json
@@ -1786,6 +1786,6 @@
     "description": "Taken from a world that embraced cybernetics in their sports, this CBM links the eyes to a series of nerve stimulators along the arms and hands.  The result is a significant increase to throwing accuracy, and it turns out it works great with fragmentation grenades.",
     "occupied_bodyparts": [ [ "eyes", 1 ], [ "arm_l", 2 ], [ "arm_r", 2 ], [ "hand_l", 2 ], [ "hand_r", 2 ] ],
     "flags": [ "BIONIC_NPC_USABLE", "BIONIC_SHOCKPROOF" ],
-    "enchantments": [ { "condition": "ALWAYS", "skills": [ { "value": "throwing", "add": 5 } ] } ]
+    "enchantments": [ { "condition": "ALWAYS", "skills": [ { "value": "throw", "add": 5 } ] } ]
   }
 ]


anoobindisguise:
Still more realistic than the electromagnetic unit, the quantum tunnel unit, the uncanny dodge, the active defense force field, all the weapon bionics, the tools bionics, and quite a few others. That's IMO an argument to remove all those too, but it's worth considering the implications of the removal. The sparks should definitely go at the very least though

GuardianDll:
I thought the whole idea of cbm is "it can do things that make no sense", deleting is because it "makes no sense" means we need to delete 99.9% of all our bionic becuase "having metal bones/hydraulic inside your body/metal sheet instead of your skin makes no sense"

I thought about it previously, and i think it can work as "you have a coil in your hand, it charges the metal item inside your hand, and right in moment you throw the item, cbm emit a huge magnetic impulse of the same charge (ie positive charged metal + positive charged impulse), which gives a good boost for item you throw". Of course sparks are still not a thing (and i don't know electric good enough to even think is it possible)

Drew4484:
> I thought the whole idea of cbm is "it can do things that make no sense", deleting is because it "makes no sense" means we need to delete 99.9% of all our bionic becuase "having metal bones/hydraulic inside your body/metal sheet instead of your skin makes no sense"
> 
> I thought about it previously, and i think it can work as "you have a coil in your hand, it charges the metal item inside your hand, and right in moment you throw the item, cbm emit a huge magnetic impulse of the same charge (ie positive charged metal + positive charged impulse), which gives a good boost for item you throw". Of course sparks are still not a thing (and i don't know electric good enough to even think is it possible)

The repulsion coil works but how does it tie into our hardwired brains releasing our fingers just right for an accurate throw? And if it provides the kind of force evident by the CBM in action, surely it's not relying on the throwing motion for any significant amount of energy. 

In general I feel that CBMs should be at least feasible, either as an installed bionic or as a complete cybernetic replacement. This will likely require culling or at least heavily rewriting some of the CBMs, but I think we'll wind up with some cool stuff as a result.

tenmillimaster:
In which alternate dimension did some implant-reviewing-agency approve this, anyways? Good riddance to this thing.

GuardianDll:
Is this "this thing is bad" or "this thing is bad as CBM" discussion? because its like a different things, and not related to each other (and also require a lot more effort to fix it than "ill just obsolete it")

AirRice:
The "Railgun" in this context is clearly a reference to the well-known _Toaru Majutsu no Index_ series and one of its main characters' abilities to launch a coin at high speeds utilising the character's electrical manipulation abilities. In that series the "railgun" projectile is launched by electrical currents emitted alongside the user's arms and no throwing motion is needed. 

The original reasoning for the CBM makes no sense however and in reality the electrical discharge needed to even make this happen at a reasonable speed is immense. I propose changing the flavor text on how it works and increasing the energy cost of this CBM significantly but a full removal might be overdoing it. Or move it to aftershock, like @John-Candlebury mentioned.

Ragecrabbo:
If this does end up getting removed i hope we'll get some other form of throwing related CBM, the skill could use some love.

Fris0uman:
Can you add the wrist CBM we discussed in discord as part of this PR? The removal and replacement can happen in the same PR I don't think we need to do it in two steps.

Drew4484:
> Can you add the wrist CBM we discussed in discord as part of this PR? The removal and replacement can happen in the same PR I don't think we need to do it in two steps.

Yeah, I'll draft that up today.

Maleclypse:
I like the replacement cbm.  Thanks for doing this!

data/json/items/bionics.json
```diff
@@ -1276,5 +1264,17 @@
...
     "difficulty": 3,
     "installation_data": "AID_bio_trickle"
+  },
+  {
+    "id": "bio_pitch_perfect",
+    "copy-from": "bionic_general",
+    "type": "BIONIC_ITEM",
+    "name": { "str": "Throwing Assist CBM" },
+    "looks_like": "bio_int_enhancer",
+    "description": "Taken from a world that embraced cybernetics in their sports, this CBM links the eyes to a series of nerve stimulators along the arms and hands.  The result is a significant increase to throwing accuracy, and it turns out it works great with fragmentation grenades.",
```
anothersimulacrum, line 1274: I think this should be a bit more vague.
Here's a suggestion
```suggestion
    "description": "This CBM links the eyes to a series of nerve stimulators along the arms and hands, resulting in a significant increase to throwing accuracy.  The user enjoys a significant edge in baseball, and it turns out it works great with fragmentation grenades too.",
```

data/json/items/software.json
```diff
@@ -243,5 +236,12 @@
     "description": "Huge archives of numerous IC circuit datasheets from several major manufacturers.  Probably valuable to the right person, as it would be hard to salvage and reuse these components without them.",
     "price": 40000,
     "price_postapoc": 0
+  },
+  {
+    "id": "AID_bio_pitch_perfect",
+    "type": "GENERIC",
+    "copy-from": "AID_abstract",
+    "name": { "str_sp": "Throwing Asssist CBM installation data" },
+    "description": "Precise step-by-step instructions and other required data for installation of the Throwing ASsist CBM, stored on a data card of proprietary form-factor, designed to be used extensively by an Autodoc, and allowing bionic installation with minimal failure chance.  Integrated security subroutine renders it unusable after installation."
```
TheShadowFerret, line 245:
```suggestion
    "name": { "str_sp": "Throwing Assist CBM installation data" },
    "description": "Precise step-by-step instructions and other required data for installation of the Throwing Assist CBM, stored on a data card of proprietary form-factor, designed to be used extensively by an Autodoc, and allowing bionic installation with minimal failure chance.  Integrated security subroutine renders it unusable after installation."
```

PR #61180, opened 2022-09-21
Caterpillar cleanup & Cocoons

Karol1223:
#### Summary
Content "Splits caterpillars into two species like they should be and adds cocoons"

#### Purpose of change
I was annoyed that moths and butterflies laid eggs that hatched into the same baby form. A moth could be a parent to a butterfly and vice versa. No more.

#### Describe the solution
- Smallest mutant butterflies and moths can no longer breed
- The remaining mutant moths now lay unique moth eggs
- Smallest mutant caterpillars are obsoleted, there is no need for two different size variants in larva
- Larger mutant caterpillars have been split into two separate monsters. Additionally they now use the snake bodytype instead of insect, are insect faction instead of ant faction (same as their parent monsters, I am not sure why it was ant in the first place) and have fear triggers
- Added cocoons, two separate for the two caterpillars. Not much to say, they're essentially identical to wasp pupae except slightly larger and don't have the wasp faction
- Removed the obsoleted caterpillars from wilderness spawns, replaced them with new ones

I considered not doing the cocoons but they seem like a nice flavour. I also really thought migrating monsters was possible and intended to have old caterpillars migrate into the new ones but this doesn't seem possible. The obsoletion code still has them upgrading, though, so any old caterpillars in current saves will eventually evolve out of this state. I also considered adding butterfly eggs to recipes alongside the moth ones but that's already handled by someone else and just needs merging.

#### Testing
Appears to work ingame now. Doesn't throw errors anymore (initially I messed up upgrading for cocoons but I fixed that right after I submitted the draft).

### DIFF
--- a/data/json/monsters/insect_spider.json
+++ b/data/json/monsters/insect_spider.json
@@ -3212,7 +3212,7 @@
   {
     "id": "mon_caterpillar_butterfly",
     "type": "MONSTER",
-    "name": { "str": "caterpillar" },
+    "name": { "str": "giant caterpillar" },
     "description": "The larval stage of a butterfly or a moth.  This one appears to be about your size so you can only imagine the size of the insect that it will grow into.",
     "default_faction": "insect",
     "bodytype": "snake",
@@ -3235,7 +3235,7 @@
   {
     "id": "mon_butterfly_cocoon",
     "type": "MONSTER",
-    "name": { "str": "cocoon" },
+    "name": { "str": "giant cocoon" },
     "description": "A caterpillar undergoing a metamorphosis.  Considering the size of the cocoon you have a feeling the emerging adult will be massive as well.",
     "symbol": "0",
     "color": "light_gray",
@@ -3254,7 +3254,7 @@
   {
     "id": "mon_caterpillar_moth",
     "type": "MONSTER",
-    "name": { "str": "caterpillar" },
+    "name": { "str": "giant caterpillar" },
     "description": "The larval stage of a butterfly or a moth.  This one appears to be about your size so you can only imagine the size of the insect that it will grow into.",
     "default_faction": "insect",
     "bodytype": "snake",
@@ -3277,7 +3277,7 @@
   {
     "id": "mon_moth_cocoon",
     "type": "MONSTER",
-    "name": { "str": "cocoon" },
+    "name": { "str": "giant cocoon" },
     "description": "A caterpillar undergoing a metamorphosis.  Considering the size of the cocoon you have a feeling the emerging adult will be massive as well.",
     "symbol": "0",
     "color": "light_gray",


data/json/monsters/insect_spider.json
```diff
@@ -3101,9 +3101,7 @@
     "harvest": "arachnid_flying",
     "weakpoint_sets": [ "wps_arthropod", "wps_arthropod_flying" ],
     "families": [ "prof_wp_basic_bug" ],
-    "reproduction": { "baby_egg": "egg_butterfly", "baby_count": 1, "baby_timer": 15 },
```
LyleSY, line 3104: Why aren't these able to lay eggs anymore?

data/json/monsters/insect_spider.json
```diff
@@ -3167,9 +3165,7 @@
...
```
LyleSY, line 3170: Same question here

data/json/monstergroups/eggs.json
```diff
@@ -245,11 +245,12 @@
   {
     "name": "GROUP_EGG_BUTTERFLY",
     "type": "monstergroup",
-    "default": "mon_caterpillar",
-    "monsters": [
-      { "monster": "mon_caterpillar_giant", "freq": 25, "cost_multiplier": 1, "starts": "28 days" },
-      { "monster": "mon_caterpillar", "freq": 100, "cost_multiplier": 1 }
-    ]
+    "monsters": [ { "monster": "mon_caterpillar_butterfly" } ]
+  },
```
LyleSY, line 249: I'm not sure eliminating the giant caterpillar entirely is the best move here. Consider having both giant and regular for both butterfly and moth and growing into their related butterfly and moth size. It may also help to define some larger eggs so this has a  more logical flow

LyleSY (changes requested):
Overall I agree that separating out moth and butterfly life cycles is the right design choice and I like your cocoons. I'd like to better understand your thinking on why some sizes of monster no longer lay eggs and why you think there should be only one size caterpillar. Ideally I'd like to see us keep the same feature richness and add more, not trade unless there's a good reason

Karol1223:
I personally believe that having larva sizes is just really confusing. Sure, it could be done, but I prefer consistency all along the insects we have instead of one singular insect being different and needing to edit all others to match that. I'm of the opinion that adding larva sizes is unnecessary and same applies for egg sizes (ladybugs do that, not a fan of that one either). Same for smallest bug sizes, I've been going with the mindset of smallest mutants are kind of chaos and when they grow larger they sort of balance out and eventually stop mutating into insane sizes (above mega tier) as that's what I've overheard during a discussion about this topic on the development discord. Also, reading through the issues lead me to believe this is the approach Venera prefers and I'm not going against their will in the field of insects. What my bug PRs aim to do is unify the progression across the board, keeping two different larva sizes kind of contradicts that. It could be argued that so do cocoons and I don't disagree, I was on the fence adding those, but after discussing where those would be applicable with Venera I came to the conclusion that for butterflies and moths they do indeed fit.

LyleSY (approved):
Some big design changes here but they appear to be well thought out and consistent with what @Venera3 is doing. Forking reproduction and adding cocoons is a win

Venera3 (approved):
Yeah, the generic progression I arrived at for bugses and discussed with Karol is something like
```
smallish dog/cat-sized "in progress" bugs
man-sized "stable" species capable of reproduction where applicable
mega variants that are random "dead ends" - either no or throttled reproduction, warped way off normal anatomy/behavior
```
Colony builders will not really fit in there, of course, and there's always room for weird one-off offshoots but the smallish version reproducing on-screen makes that a bit too messy for my tastes. Cocoons are a bit weird for pack-spawned monsters (the whole pack flips at the same time), but that's not a factor for solo mobs and it's pretty nice for flavor.

PR #61183, opened 2022-09-21
fries can spoil

LyleSY:
#### Summary
Balance "fries can spoil"

#### Purpose of change

Realism

#### Describe the solution

Fries not made by the player no longer are fresh in perpetuity. Flavor text talking about how that makes no sense is removed. Player made fries spoil in one day instead of three.

Develop a deep french fry mythos with horrifying implications

#### Testing

Simple JSON edit

#### Additional context

Inspired by staring at some french fries found in a mansion and thinking "never? French fries never spoil?" and then googling https://foodiosity.com/how-long-do-french-fries-last/ which agrees with others that two hours outside the fridge and it's growing bacteria and needs to be thrown out

### DIFF
--- a/data/json/items/comestibles/junkfood.json
+++ b/data/json/items/comestibles/junkfood.json
@@ -688,7 +688,6 @@
     "name": { "str_sp": "French fries" },
     "conditional_names": [ { "type": "COMPONENT_ID", "condition": "groundnut_prepared", "name": { "str_sp": "ground nut fries" } } ],
     "copy-from": "fries",
-    "spoils_in": "1 days",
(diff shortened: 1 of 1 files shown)


TheShadowFerret:
Does this mean Foodplace food/snack will no longer be magic too 👀?

LyleSY:
Over my dead body. FoodPlace Food is made of pixie dust and the dreams of murdered orphans, realism be damned

Sniper27IT:
I've always though that unexpirable fries was a meme about " a famous fast food chain whose mascot is a clown and a couple of yellow arches" and junk food coming from those places.

PR #61189, opened 2022-09-22
Adds looks like to seeds

secretstamos:
#### Summary
Infrastructure "None"

#### Purpose of change
Adds looks like to the seeds for tilesets

#### Describe the solution
Added looks like fields to a variety of entries

Could be handled for each tileset locally

#### Testing
N/A


Fris0uman:
If they copy from `seed` they don't need a looks_like, tilesets can just set up a sprite for `seed` I think

secretstamos:
> If they copy from `seed` they don't need a looks_like, tilesets can just set up a sprite for `seed` I think

Should I close this request then?

PR #61198, opened 2022-09-22
Throwing practice is now difficulty zero

Inglonias:
#### Summary
Balance "Throwing practice is now autolearned at throw 0"

#### Purpose of change
While requiring a throwing skill of 1 to use the practice recipe was a balance choice when I first added it, all it really does is add a bunch of unnecessary keypresses to get from a throwing skill of 0 to a throwing skill of 1 to unlock the recipe, and that assumes that a player knows that practice recipe is an option to begin with. If you can do essentially the same thing with a few dozen keypresses at any skill level, then why bother, right?

#### Describe the solution
Throwing practice is now auto-learned by default. No books required. Literally anyone can chuck a rock at something. Even feral humans throw rocks, and they're... well, feral. Assuming your survivor is smarter than that, they can throw rocks at the wall, or a tree, or whatever, to practice their aim.

Not doing this? It's a spur-of-the-moment proposal on my part.

#### Testing
I'm leaving it in the hands of the automated testing suite. If they say it's ok, then it's ok.

### DIFF
--- a/data/json/recipes/practice/ranged.json
+++ b/data/json/recipes/practice/ranged.json
@@ -11,8 +11,7 @@
     "skill_used": "throw",
     "time": "1 h",
     "practice_data": { "min_difficulty": 0, "max_difficulty": 2, "skill_limit": 3 },
-    "autolearn": [ [ "throw", 1 ] ],
-    "book_learn": [ [ "mag_throwing", 0 ], [ "manual_throw", 0 ] ],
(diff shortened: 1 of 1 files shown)


Inglonias:
I would love to know how a two-line JSON change causes some vision and map tests to fail...

NetSysFire:
My guess is that it might have been a race condition or some other unlucky RNG. Someone already kicked the tests, so it should work for now.

PR #61204, opened 2022-09-22
Draw vehicle parts in two stages: first below roof, then roof

mlange-42:
#### Summary

Interface "Draw vehicle parts in two stages: first below roof, then roof"

#### Purpose of change

Don't let things on vehicle roofs like solar panels or security cameras "erase" all other parts.
Draw things on the roof over the top-most part below the roof. See screenshot for an illustration.

#### Describe the solution

Draw vehicle parts in two passes. First, select from below roof following the normal logic. Draw it. Then, if there is something on the roof and the player is not in the vehicle, draw that.

Leave it as is.

#### Testing

Tried in-game with Ultica-ISO tileset (see screenshot).

#### Additional context

Screenshot using the feature. Without it, the vehicle parts below the solar panels and cams would disappear.

grafik

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -3102,7 +3102,7 @@ int vehicle::part_displayed_at( const point &dp, bool include_fake, bool below_r
     int hide_z_at_or_above = in_vehicle ? ON_ROOF_Z : INT_MAX;
 
     int top_part = -1;
-    int top_z_order = below_roof ? -1 : 8;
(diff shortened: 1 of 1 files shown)


mlange-42:
> Does this work for orthographic tilesets as well?

Sure! So no oversized full-tile security cameras needed anymore. ;)

dseguin (approved):

PR #61206, opened 2022-09-23
[Innawood] bulletin board

### LINKED ISSUE #61201: [Innawood] Override the bulletin board recipe with something easy

The bare bones base camp requires the manual construction of a bulletin board to be useful. However, the standard bulletin board requires resources that are a bit hard to acquire in Innawood (nails and planks). Therefore, I suggest the recipe for the bulletin board should be overridden by something comparatively simple to construct in that mod.

### Solution you would like.

Construct a "bulletin board" from e.g. 3 sticks and a short string/cord (which would make a tripod) in Innawood, and also make sure the deconstruction/bashing doesn't magic up the standard resources, but rather things based on the local recipe.

A short string/cord requires a little work to get, but not that much.

Leave it as is.

### Additional context

_No response_

PatrikLundell:
#### Summary

None

#### Purpose of change

Provide an easy to make "bulletin board" for Innawood usage with the bare bones base camp.
Fix #61201 

#### Describe the solution

- Added an overriding mod description/deconstruct/smash description for the bulletin board.
- Added an Innawood specific bulletin board construction recipe from easy to get materials (with the deconstruct/smash results as compatible as possible, given that variant construction materials isn't recorded).
- Added an Innawood specific construction order for the new recipe.

Tried to override the "real" bulletin board construction recipe, but was met with an error message claiming the recipe to be duplicated. It seems mod overrides might not be supported for constructions.

#### Testing

Loaded an existing Innawood save, verified that the new construction alternative appeared and contained the components expected, constructed it, created a bare bones base camp, and verified the bulletin board worked as base camp interface.
Failed to figure out how to interact with the bulletin board to deconstruct it, smashed it in a fit of rage ;), and verified the resulting debris met the expectations.

#### Additional context

A side effect of the failure to replace the construction recipe is that the original recipe still exists, using the new description, but using the original components. I expect it to not return any nails when smashed, as the furniture has been overridden.

All 3 of the new entries were placed in the least ill fitting existing mod override file. It might be desirable to structure it differently, such as e.g. creating one mod file (and associated directory structure, when needed) for each file the originals resided in.

### DIFF
--- a/data/mods/innawood/furniture_and_terrain/furniture-tools.json
+++ b/data/mods/innawood/furniture_and_terrain/furniture-tools.json
@@ -32,5 +32,53 @@
         { "item": "splinter", "count": 1 }
       ]
     }
+  },
+  {
+    "type": "furniture",
+    "id": "f_bulletin",
+    "name": "bulletin board",
+    "description": "A crude approximation of a 'real' bulletin board.  Good for pinning various notices for other survivors to read.",
+    "symbol": "6",
+    "color": "blue",
+    "move_cost_mod": -1,
+    "coverage": 75,
+    "required_str": -1,
+    "flags": [ "FLAMMABLE", "ORGANIC", "TRANSPARENT" ],
+    "max_volume": "120 ml",
+    "examine_action": "bulletin_board",
+    "deconstruct": {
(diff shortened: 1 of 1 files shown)


Light-Wave (approved):

ashGlaw (approved):
Trying to get some more code reviewing experience. Looks good to me, but take that for what it's worth

PR #61210, opened 2022-09-23
Zombie animals don't fear fire

Karol1223:
#### Summary
Bugfixes "Zombie animals don't fear fire"

#### Purpose of change
It was brought up in the development discord that antlered horrors actually fear fire. Looking into this, it appeared to be true but a huge chunk of zanimals did as well, so there's the fix.

#### Describe the solution
Removed the fear triggers from all zombie animals that had them. This should affect all zombies from the families of moose, cougar, beaver, wolf, bear, pig, bull and tiger.

#### Testing
It's just fear trigger removal, shouldn't cause errors.

### DIFF
--- a/data/json/monsters/zed-animal.json
+++ b/data/json/monsters/zed-animal.json
@@ -169,7 +169,6 @@
     "harvest": "zombie_fur",
     "special_attacks": [ [ "HOWL", 10 ], { "type": "bite", "attack_upper": false, "cooldown": 2, "accuracy": 4, "infection_chance": 8 } ],
     "anger_triggers": [ "PLAYER_CLOSE", "HURT" ],
-    "fear_triggers": [ "FIRE" ],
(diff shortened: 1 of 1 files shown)


NetSysFire:
Because zombies literally do not have the brains to process fire as a danger anymore. This is also why blackberry bushes are effective fortifications now.

GuardianDll:
Okay. Why zanimals still have brains, when zeds don't?

ashGlaw (approved):

Nebnis:
I think you misunderstood Netsysfire. zanimals are like zeds, they don't have the cognitive ability to avoid danger anymore. Unless they are some very specific evolved one.

GuardianDll:
Oh I'm dumb, my bad

LyleSY (approved):

PR #61223, opened 2022-09-24
[Magiclysm] Buff spells rebalance

Dan1ss1mo:
#### Summary
Balance "Some spell adjustments to make them more equal"

#### Purpose of change

All this changes are based on my personal experience with this spell. It is my first rebalnce balanse change though, so i would highly appreciate feedback from other people about this particular spells. I'll also try to explain my reasoning behind all the changes. 

Let's start with Synaptic Stimulation and Magus +4 spells. I really love Synaptic Stimulation and it was my go to spell for pretty much any sitation, but it was so good compared to other simular spell that i just can't bring my self not to nerf it to be more reasonable compared to other buffs. It has absolutly ridiculous base duration, making it arguable the best 1 level spell, as well as not giving much incentive to actually level, as duration increment was not that noticeable. This is become especially noticeable then you compare it to Cat's Grace and Fox's Cunning. At max level, they give you LESS bonuses fot DOUBLE the manacost. It is even worse at lower levels because of how high base duration of  Synaptic Stimulation is. I tried to change they mana to duration ratio(MDR for short) to be closer at all levels, but leave Synaptic Stimulation to be slightly better, because its a higher level spell. I also change magus buffs to be level 20, like almost all other buffs and remove the concentration requirement as the effect isn't as strong compared to other simular spells to have this limitations.

Bless;
I dont increase its duration as mush as some other spells  because of its AOE, which make it much more impactful if you running around with npc companions,
Haste: i really love how powerful this spell is. As probably all peaple how experienced this spell in DND. But i think for how good it is it was too easy to use, so i increase it's difficulty, manacost and base cast time, while leaving the effect the same.
Grotesque Enhancement: I abolutly adore the flavor of this spell, but it low duration, combined with very long castiong time, manacost, and marginaly stronger effect then what of the bless made it almost useless in any practical situation. I decided to change it so that it still last relatively short time copared to other spells, but also pretty fast to cast once you level it high enough, so you can transform in your super for in emergency, kind of like a Hulk)

"Coagulant Weave":
While the bleed and bite immunity is neat, both of those effects are rearly a concern past really early game. The stamina regeneration effect is nice, but not enough to jastify such poor MDR.

"windrun":
Add manacost and cast time reduction based on level and slightly boost duration, as effect are pretty minor copared to other buff sells.

"Glide on Ice"
while not technically a buff spell, i still want to adjust it a bit, specificaly the ice coating is summoned. Curantly it's not much better then just using regular rollerblades. I changed it to act more like heelys, with faster transition between on and off state and higher movecost reduction.

#### Describe the solution
Implemant changes from this pr, maybe with some adjustments of specific values.

Leave everything as it is

#### Testing
Use debug menu to learn mentioned spell and check they stats/effects

#### Additional context

"Glide on Ice" was rolled back for now to avoid string freeze


data/mods/Magiclysm/Spells/magus.json
```diff
@@ -183,9 +183,11 @@
     "shape": "blast",
     "spell_class": "MAGUS",
     "energy_source": "MANA",
-    "difficulty": 4,
-    "base_casting_time": 200,
-    "base_energy_cost": 300,
+    "difficulty": 9,
```
GuardianDll, line 186: kk i thought it still overpovered, but it has only 2 minute length, so its fine

data/mods/Magiclysm/Spells/technomancer.json
```diff
@@ -24,8 +24,8 @@
     "range_increment": 1.0,
     "//": "duration is in moves",
     "min_duration": 6000,
-    "max_duration": 10000,
-    "duration_increment": 400
+    "max_duration": 18000,
+    "duration_increment": 600
```
GuardianDll, line 28:
```suggestion
    "max_duration": 120000,
    "duration_increment": 5700
```
3 minutes is better than before, but still nothing in my eyes. Lets make it 20 minutes at least

data/mods/Magiclysm/Spells/technomancer.json
```diff
@@ -14,7 +14,7 @@
     "base_energy_cost": 100,
     "energy_source": "MANA",
     "spell_class": "TECHNOMANCER",
-    "difficulty": 1,
+    "difficulty": 2,
```
GuardianDll, line 17:
```suggestion
    "difficulty": 5,
```
...and a bit harder to cast

GuardianDll (commented):
Overall its good imo

data/mods/Magiclysm/Spells/biomancer.json
```diff
@@ -332,13 +336,14 @@
...
+    "difficulty": 4,
     "base_casting_time": 400,
-    "base_energy_cost": 150,
+    "base_casting_time": 200,
+    "base_energy_cost": 100,
     "max_level": 20,
     "min_duration": 6000,
-    "max_duration": 12000,
-    "duration_increment": 300
+    "max_duration": 90000,
```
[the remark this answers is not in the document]
Dan1ss1mo: seem a forget to chang this one, it was in my check list

data/mods/Magiclysm/Spells/magus.json
GuardianDll: press "resolved" on this review, i forgot it doesnt work as a comment field

Dan1ss1mo: i think it still way to good, even if it's just 2 minutes, you become an absolute killing machine, even just been able to outwalk everything in the game is very strong

data/mods/Magiclysm/Spells/technomancer.json
```diff
@@ -24,8 +24,8 @@
     "range_increment": 1.0,
     "//": "duration is in moves",
     "min_duration": 6000,
-    "max_duration": 10000,
-    "duration_increment": 400
+    "max_duration": 30000,
+    "duration_increment": 1200
```
Dan1ss1mo, line 28: on the second the second thought, 20 minutes seems to be too god, as it gives you more then 1.5 MDR of magus +4 spells. Espasialy cosidering it fast cast time and AOE

data/mods/Magiclysm/items/ethereal_items.json
```diff
@@ -116,9 +116,9 @@
     "armor": [ { "encumbrance": 15, "coverage": 70, "covers": [ "arm_l", "arm_r", "hand_l", "hand_r" ] } ]
   },
   {
-    "id": "ice_gliders",
+    "id": "ice_gliders_on",
```
Fris0uman, line 119: Don't change ids it will break saves where the itme is present

data/mods/Magiclysm/Spells/kelvinist.json
```diff
@@ -260,7 +260,7 @@
     "min_damage": 1,
     "max_damage": 1,
     "effect": "spawn_item",
-    "effect_str": "ice_gliders",
+    "effect_str": "ice_gliders_on",
```
Fris0uman, line 263:
```suggestion
    "effect_str": "ice_gliders",
```
don't change ids

data/mods/Magiclysm/Spells/kelvinist.json
```diff
@@ -237,7 +237,7 @@
...
```
Fris0uman, line 240:
```suggestion
    "effect_str": "ice_gliders",
```
don't change ids

data/mods/Magiclysm/items/ethereal_items.json
Dan1ss1mo: sure, didn't think about it

GuardianDll (approved):

GuardianDll (approved):
ok, overall its good, except i still think it must be longer than 5 minute at max lvl, but we can discuss it later

LyleSY:
Seems fine but I think it should be in string freeze

LyleSY (approved):
I see no strings, I have no issues at this point

PR #61224, opened 2022-09-24
Harmonize connects_to and rotates_to - better end pieces

mlange-42:
#### Summary

Interface "Harmonize connects_to and rotates_to - better end pieces"

#### Purpose of change

So far, it is not possible to have different North vs. South or East vs. West end pieces with multitile and `rotates_to`.
This is ok for windows and doors, but not for furniture that connects to itself and is intended to rotate into the room/towards indoor floor.

#### Describe the solution

* [x] Add support for the different end piece variants
* [x] Optimize `rotates_to` sprite order for best fallbacks (modulo #sprites)
* [x] Add support for 5x5 template to `slice_autotile.py` and `unslice_autotile.py` in [I-am-Erk/CDDA-Tilesets](https://github.com/I-am-Erk/CDDA-Tilesets). See https://github.com/I-am-Erk/CDDA-Tilesets/pull/1606
* [x] Add `rotates_to` to some furniture
* [x] Adapt tests for changed rotation order
* [x] Document it (see [TILESET.md ->](https://github.com/mlange-42/Cataclysm-DDA/blob/fix-rotate-end-pieces/doc/TILESET.md#full-rotates_to-template))

Note that the order of rotated sprites was chosen to have unchanged behaviour on tilesets not using the feature.

Builds up on #60666. Contributes to #51619.

None

#### Testing

Tried in-game. See screenshot.

#### Additional context

Ultica-ISO sinks orienting away from glass walls, using the feature. Below is the used 5x5 multitile.

grafik

f_sink_multitile

Similar for desks, which self-connect:

grafik

### DIFF
--- a/src/cata_tiles.cpp
+++ b/src/cata_tiles.cpp
@@ -4580,7 +4580,7 @@ void cata_tiles::get_tile_values_with_ter(
     }
     // calculate rotation for unconnected tiles based on surrounding walls
     // if not any rotates_to neighbours
-    if( subtile == unconnected && rotation_targets == 0 ) {
+    if( subtile == unconnected && ( rotation_targets == 0 || rotation_targets == CHAR_MAX ) ) {
         int val = 0;
         bool use_furniture = false;
 
--- a/tests/connect_rotate_test.cpp
+++ b/tests/connect_rotate_test.cpp
@@ -435,7 +435,7 @@ TEST_CASE( "unconnected windows rotate to indoor floor", "[multitile][rotates]" 
             cata_tiles_test_helper::get_connect_values( pos, subtile, rotation, TERCONN_NONE,
                     TERCONN_INDOORFLOOR );
             CHECK( subtile == unconnected );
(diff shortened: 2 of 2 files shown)


dseguin (approved):
LGTM, but looks like this will temporarily break some tileset rotations. Would it be worth it to coordinate merging with the next tileset update?

mlange-42:
> LGTM, but looks like this will temporarily break some tileset rotations. Would it be worth it to coordinate merging with the next tileset update?

Thanks for review!

The changes here (I think you mean changed rotation order) only break things related to the relatively new `rotates_to` feature. It looks like this feature is so far only used in Ultica-ISO, which I fix in https://github.com/I-am-Erk/CDDA-Tilesets/pull/1614.

Other than that, rotation order is arranged in a way that the `rotates_to` feature has no effect on tilesets not making use of it.

dseguin:
Good enough for me!

PR #60159, opened 2022-08-14
Build Windows Release with CMake

alef:
#### Summary
None

#### Purpose of change
**WARNING**: CMake build is **NOT** official and should be used for *dev purposes ONLY*.
Use CMake and VCPKG to build a MinGW release, including GitHub Actions.

#### Describe the solution
- [x] Raise CMake requirement to v3.21
- [x] User documentation under `doc/COMPILING`
- [x] Branch interactivelty rebased to explain reasons for changes in commits
- [x] Extends and improve the work done in #56875
- [x] The MinGW build runs under Ubuntu 22.04 and replaces MXE with VCPKG
- [x] SDL version now dictated by VCPKG
- [x] The names of the CMake presets are now the same as in `release.yml`
- [x] The GHA caches needs optimization. For now, trash them with gh-actions-cache

#### Describe alternatives you've considered
Same as in #56875

#### Testing
- [x] Run command as by `doc/COMPILING/COMPILING-CMAKE-VCPKG.md` on Windows and Ubuntu.
- [x] Run `cmake.yml` under GHA at my account.
You may prefer disabling it by clicking _Actions -> All Workflows: CMake Release -> [...] -> Disable workflow_.

PR #61225, opened 2022-09-24
Build Windows Release with CMake (part1)

alef:
#### Summary
None

#### Purpose of change
Try to align CMake build to match the release.

#### Describe the solution
- [x] Deal with some package name differing between VCPKG and other distributions
- [x] Use the proper way for supporting ccache
- [x] Add missing files to the relase

#### Testing
1. Drafting this PR
2. Watch CI build
3. Undraft if no side-effects

#### Additional context
Part of discarded #60159


BrettDong:
I tried your branch and I can successfully build `cataclysm-tiles.exe` on Windows with MSVC and CMake. However I found I need to delete `findSDL2_ttf.cmake` in `CMakeModules` directory, otherwise there will be a CMake configuration error saying unable to find either target `SDL2_ttf::SDL2_ttf` or `SDL2_ttf::SDL2_ttf-static`. I guess the `findSDL2_ttf.cmake` in `CMakeModules` is too old that it does not provide a modern CMake style target for the project to consume?

PR #61226, opened 2022-09-24
western holster + leather belt  tweaks

nopenoperson:
#### Summary
None

#### Purpose of change
fix western holster so it layers properly, doesnt work with MOLLE and encumbrance is split up between the torso and one of the legs, add some material definitions and change the name to reflect the fact that its a belt with a metal belt buckle ( this part of the item is slightly thicker because of the leather overlapping + belt buckle being there) with a holster attached 

#### Describe the solution
change some stuff in the json

rewriting holster.json, slightly tweaking recipe, not doing anything

#### Testing
i tried to layer it in a lot of different ways, put stuff in the holster.


data/json/items/armor/holster.json
```diff
@@ -233,20 +233,32 @@
...
     "armor": [
-      { "covers": [ "torso" ], "coverage": 5, "encumbrance": [ 1, 1 ], "specifically_covers": [ "torso_lower" ] },
+      {
+        "covers": [ "torso" ],
+        "coverage": 50,
+        "specifically_covers": [ "torso_waist", "torso_lower" ],
+        "encumbrance": [ 1, 1 ],
+        "material": [
+          { "type": "leather", "covered_by_mat": 100, "thickness": 0.75 },
+          { "type": "steel", "covered_by_mat": 100, "thickness": 0.75 }
```
Maleclypse, line 256: There's no steel mentioned in either the description or the item material list.  The coverage provides 50% more thickness than the item itself.  I would have thought that would trigger a test failure, but maybe I'm misunderstanding.

nopenoperson: its the belt buckle from the leather belt

Maleclypse (approved):

PR #61231, opened 2022-09-24
Bugfix for Mafia Boss profession

Raiksler:
#### Summary
Bugfixes "Starting item 'brass knuckles' from mafia boss profession moved to tux pocket, to avoid conflict with golden ring worn"

#### Purpose of change
The mafia boss profession has a starting item "brass knuckles". Brass knuckles are categorized as "unarmed_weapons" and also occupy the "fingers" layer. However, brass knuckles are not in the character's inventory at the start of the game and are not equipped on him. This is due to the fact that brass knuckles conflict with golden ring, which also occupies the "fingers" layer.

#### Describe the solution
I removed the brass knuckles from the character and moved them to the pockets of the tux.

Remove golden ring or brass knuckles from the character's inventory. Not cool. Any mafia boss should have a pair of trusty brass knuckles and some good jewelry =)

#### Testing
I created the character and made sure everything works as it should.

#### Additional context
In addition, I made sure that all weapons from the category "unarmed_weapons" do not cause similar conflicts.

### DIFF
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -5046,7 +5046,7 @@
           { "item": "cigar", "container-item": "case_cigar", "charges": 2 },
           { "item": "deagle_44", "ammo-item": "44fmj", "container-item": "holster", "charges": 8 },
           { "item": "deaglemag", "ammo-item": "44fmj", "charges": 8, "count": 2 },
-          { "item": "knuckle_brass", "container-item": "tux"}
+          { "item": "knuckle_brass", "container-item": "tux" }
         ]
       },
       "male": [ "boxer_shorts", "diamond_gold_cufflinks" ],


PatrikLundell:
A mafia boss doesn't need any knuckles. They'd just remark that it would be unfortunate if someone would be beaten up, and the mafia goons that DO need knuckles would take care of it. Thus, you'd either have a boss overladen with valuable jewelry or a goon with knuckles.

Raiksler:
> A mafia boss doesn't need any knuckles. They'd just remark that it would be unfortunate if someone would be beaten up, and the mafia goons that DO need knuckles would take care of it. Thus, you'd either have a boss overladen with valuable jewelry or a goon with knuckles.

But nevertheless, at the moment we have a mafia boss with declared brass knuckles, which, however, we cannot use

PR #61236, opened 2022-09-24
Bugfix for lost submissive profession

Raiksler:
#### Summary
Bugfixes "Fixed matchbox disappearing bug in lost submissive profession"

#### Purpose of change
When you start the game with the profession "lost submissive", the box of matches, which is the starting item of this profession, disappears. This is due to the fact that none of the clothing items of this profession has pockets. (It's literally a simple latex suit haha)

#### Describe the solution
I gave the box of matches the "auto-wield" flag, now the character correctly appears in the world with a box of matches in his hands.

Remove the matchbox from the list of starting items. But it would make starting in the "escape from the manor" scenario less variable when playing as this character.

#### Testing
I created a character and made sure everything works as expected

#### Additional context
The original idea was to add a candle to a box of matches as starting items when starting this profession. The bug caused by the lack of pockets was discovered by accident. I think it would be great to add a candle as a starting item (because it fits perfectly into the context of the profession and will make the "escape from the manor" scenario more variable), but I can't think of where to put it.

### DIFF
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -2906,7 +2906,7 @@
     "items": {
       "both": {
         "items": [ "bondage_suit", "bondage_mask", "boots", "gloves_leather", "leather_belt" ],
-        "entries": [ { "group": "charged_matches" } ]
+        "entries": [ { "group": "charged_matches", "custom-flags": [ "auto_wield" ] } ]
       }
     },
     "missions": [ "MISSION_LOST_SUB" ]


ashGlaw (approved):

PatrikLundell:
Have you made sure this change doesn't screw up regular usage of matchboxes by e.g. placing those in the hands rather than the inventory of every character who picks those up (including the one targeted for the change of encountering matchboxes later)?

If the reason for the change is to help with a particular start location you might consider modifying that start location(s) to always have those items in the starting room(s). I don't think it matters much for anyone encountering these items in those rooms when exploring a mansion they've found, as neither item is something that would have to be made sure it's rare.

Raiksler:
> Have you made sure this change doesn't screw up regular usage of matchboxes by e.g. placing those in the hands rather than the inventory of every character who picks those up (including the one targeted for the change of encountering matchboxes later)?
> 
> If the reason for the change is to help with a particular start location you might consider modifying that start location(s) to always have those items in the starting room(s). I don't think it matters much for anyone encountering these items in those rooms when exploring a mansion they've found, as neither item is something that would have to be made sure it's rare.

Yes, this change cannot affect the use of a matchbox, as this is a simple professions.json change that only affects the location of the matchbox when starting for this specific profession. 
The reason for the change is that at the moment, we are facing an error screen when trying to start a game with this profession. I think I didn't point it out clearly enough in the "purpose of change" section.

Maleclypse (approved):

PR #61237, opened 2022-09-24
wolf skulls

LyleSY:
#### Summary

Content "wolf skulls"

#### Purpose of change

Realism, crafting options

#### Describe the solution

Defines wolf skull items

TODO:

- [x] Add them to dog and wolf harvest lists
- [x] Create a recipe for wolf masks using them
- [x] Uncraft recipe to pull bones out

Reach goals:

- [ ] Give the bonesinger something to do with them (holding off for now but could be a good followup PR)
- [x] Add them to some item spawn groups where they might make sense. (also added human skulls)

Cow skulls, deer skulls, cat skulls

#### Testing

Game loads no errors, everything appears and works as expected
Screen Shot 2022-09-27 at 12 24 42 AM

#### Additional context

Based on the work creating human skulls and the wolf skull mask the bonesinger has


CocoBatPone:
If it could be a skull of a dog wouldn't it make more sense for it to be named a canid skull? something like that.
I guess it make sense thematically though.

Jarewill:
I wasn't aware there was a canine skull helmet in the game, that's something that fits my aesthetic well.

Looking closely into it, I found out that the Bone Seer NPC had such an item on them, along other unique bone-themed items.
The description of the helmet seems to describe it as very intricately carved which I presume was made to fit with the NPC's theme.

Perhaps to not take away this unique version from the NPC, you could make a generic carving-less variant that's craftable by the player?
I don't imagine a survivor would bother to carve various symbols into a mask if they only want protection and that way the NPC would still have their unique version of the mask.

Regardless of that, this PR makes me want to also make a human skull mask and maybe bring back the bone armor in some more-realistic capacity, like a lamellar bone armor.

MNG-cataclysm (approved):

PR #61255, opened 2022-09-25
Add some (monster) leap functionality

Venera3:
#### Summary
Features "Enhance monster leap functionality"

#### Purpose of change

Leaps were a bit barebones, and more JSON special attack stuff is always nice.

#### Describe the solution

- Add a load of debugmode messages to tell you what's happening under the hood
- Allow defining the leap message
- Allow defining `self_effects` following the same syntax as the melee special attacks
- Add two new bool modifiers- `prefer_leap` to have monsters still leap when adjacent to the player, and `random_leap` to have them derp out and jump all over the place
- Moved monster spell docs to MONSTER_SPECIAL_ATTACKS.md and clarified some weirdness in there.

Add a chance element to randomness - something like a definable percent chance to ignore their target distance when jumping.
Less debugmode messages, but more readily-available info is better than less.

#### Testing

Leaps behave as previously. Messages show up with the correct syntax. Preferentially leaping monsters keep leaping when adjacent. Random leapers just jump without a care in the world.

#### Additional context

The range calc is terribly weird, but with a lot of testing (helped by the debugmode messages) it should be doable. Tested @AirRice 's proposed mobs, worked as they should with min/max range of 2 and `prefer_leap`.

### DIFF
--- a/doc/MAGIC.md
+++ b/doc/MAGIC.md
@@ -573,28 +573,7 @@ Explanation: Here we have one main spell with two subspells: one on the caster a
 
 ### Monster spells
 
-`MONSTER`s can also cast spells.  To do this, you need to declare the spells in `special_attacks`.  Spells with `target_self: true` will only target the casting monster, and will be casted only if the monster has a hostile target.
-
-```json 
-  { 
-    "type": "spell", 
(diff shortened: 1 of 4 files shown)


Fris0uman:
Would be nice to use the new features in at least one monster

PR #61256, opened 2022-09-25
Use logical instead of bitwise or for conditionals

sparr:
#### Summary
None

#### Purpose of change
Eliminate occurrences of the -Wbitwise-instead-of-logical warning added in Clang 14
https://releases.llvm.org/14.0.0/tools/clang/docs/ReleaseNotes.html#improvements-to-clang-s-diagnostics

#### Describe the solution
Replace `|` with `||` in `if( boolfunc() | boolfunc() )` constructs.

I tried annotating rotates() and connects() with `__attribute__ ((pure))` which should tell the compiler they have no side effects and invalidate part of the description of the warning. However, that had no obvious effect.

#### Testing
I compiled with this change and the warning was gone. I then played for a bit and looked at furniture tiles and did not see anything obviously out of place.

### DIFF
--- a/src/cata_tiles.cpp
+++ b/src/cata_tiles.cpp
@@ -2727,7 +2727,7 @@ bool cata_tiles::draw_terrain( const tripoint &p, const lit_level ll, int &heigh
         int rotation = 0;
         int connect_group = 0;
         int rotate_group = 0;
-        if( t.obj().connects( connect_group ) | t.obj().rotates( rotate_group ) ) {
+        if( t.obj().connects( connect_group ) || t.obj().rotates( rotate_group ) ) {
             get_connect_values( p, subtile, rotation, connect_group, rotate_group, {} );
             // re-memorize previously seen terrain in case new connections have been seen
             here.set_memory_seen_cache_dirty( p );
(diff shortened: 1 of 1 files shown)


mlange-42:
These bitwise operators were placed intentional, as both expressions should be checked (i.e. both functions called). This would not be the case with the `||` operator.

Further, #61224 removes them entirely, as the second condition causes bugs with the rotation code.

mlange-42:
> I compiled with this change and the warning was gone. I then played for a bit and looked at furniture tiles and did not see anything obviously out of place.

This rotation feature is very new, so not used anywhere outside the Ultica-ISO tileset yet.

sparr:
I'll close this since #61224 eliminates them. I'd still like to know why you'd ever need to run the second check if the first check is true.

mlange-42:
> I'll close this since https://github.com/CleverRaven/Cataclysm-DDA/pull/61224 eliminates them. I'd still like to know why you'd ever need to run the second check if the first check is true.

Take a look at the functions called there:

```cpp
bool map_data_common_t::connects( int &ret ) const
{
    if( connect_group != TERCONN_NONE ) {
        ret = connect_group;
        return true;
    }
    return false;
}
```

It does not just return a bool, but also assigns to the reference argument.
So, `rotate_group` would never be set with `||`.

sparr:
Thank you. That seems obvious in hindsight.

Do you know when #61224 is expected to land? I can build my branch locally but having compile failures in the master branch might be worth addressing sooner.

mlange-42:
> Do you know when #61224 is expected to land? I can build my branch locally but having compile failures in the master branch might be worth addressing sooner.

Could happen the next hours, or take a few more days given the amount of changes. It is ready and just waiting for someone to merge it.

PR #61259, opened 2022-09-25
scenario: create breaking bad scenario

belens:
#### Summary
Content "New Scenario: Breaking Bad (play as a chemistry teacher in the middle of nowhere with his inconspicuous RV.)"

### DIFF
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -4024,11 +4024,11 @@
   },
   {
     "type": "profession",
-    "id": "chemistry_teacher",
+    "id": "breaking_bad_heisenberg",
     "name": "Chemistry Teacher",
     "description": "After turning down a top job at a pharma company, you have chosen to use your talents to teach high schoolers.",
     "points": 0,
-    "skills": [ { "level": 3, "name": "speech" }, { "level": 2, "name": "chemistry" } ],
+    "skills": [ { "level": 3, "name": "speech" }, { "level": 6, "name": "chemistry" } ],
     "items": {
       "both": {
         "items": [ "briefs", "apron_cut_resistant", { "group": "full_gasmask" } ],
--- a/data/json/scenarios.json
+++ b/data/json/scenarios.json
@@ -587,7 +587,7 @@
     "points": 0,
     "description": "While cooking your latest batch in your trusty RV, the world around you has descended in chaos.",
     "vehicle": "food_truck_delivery_mission",
-    "professions": [ "chemistry_teacher"],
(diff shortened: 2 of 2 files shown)


data/json/scenarios.json
```diff
@@ -580,6 +580,19 @@
     "missions": [ "MISSION_LAST_DELIVERY" ],
     "flags": [ "LONE_START" ]
   },
+  {
+    "type": "scenario",
+    "id": "breaking_bad",
+    "name": "Breaking Bad: Dark Days Ahead",
```
NetSysFire, line 586: why not Breaking Zed? would potentially avoid issues with a copyrighted name.

data/json/scenarios.json
```diff
@@ -580,6 +580,19 @@
...
+    "points": 0,
+    "description": "While cooking your latest batch in your trusty RV, the world around you has descended in chaos.",
```
NetSysFire, line 588:
```suggestion
    "description": "While cooking your latest batch in your trusty RV, the world around you has descended into chaos.",
```

data/json/professions.json
```diff
@@ -4022,6 +4022,21 @@
...
     }
   },
+  {
+    "type": "profession",
+    "id": "breaking_bad_heisenberg",
+    "name": "Chemistry Teacher",
+    "description": "After turning down a top job at a pharma company, you have chosen to use your talents to teach high schoolers.",
+    "points": 0,
+    "skills": [ { "level": 3, "name": "speech" }, { "level": 6, "name": "chemistry" } ],
+    "items": {
```
TheShadowFerret, line 4032: I haven't watched BB, but I'm pretty sure he wasn't nude minus an apron & gasmask 😛

sadenar: It was underwear, gas mask and apron, plus some mittens to handle hot glassware I'm pretty sure?

belens: > I haven't watched BB, but I'm pretty sure he wasn't nude minus an apron & gasmask 😛

He was actually, or close to it! It's a big throwback to the series, there are periods where the main character goes to cook meth in the desert and he insists on not wearing his clothes because of how they would smell. So technically his clothes would be in or around the van, but I'm taking some artistic liberty here to make the scenario more interesting

Maleclypse:
You'll want to lint your changes here https://dev.narc.ro/cataclysm/format.html.  Copy the entire file into the window lint it and if it doesn't respond with syntax error copy it from the window back over your original file.  If it does have a syntax error you can correct it at jsonlint.com

PR #61262, opened 2022-09-25
Clarify world option city sizes

sparr:
#### Summary
Bugfixes "Rephrase city size options more accurately"

#### Purpose of change
Existing verbiage describes two sizes effectively equivalently ("twice as small" and "50% smaller") when they have different effects. Rephrasing makes them less ambiguous and more accurate.

#### Describe the solution
Replace "twice as small" with "half as large" for consistency, and "50% smaller" with "25% smaller" for accuracy.

I wrote a few other versions, and tried rephrasing other options to be more consistent with these, but none read as cleanly to me.

#### Testing
Loaded the game and viewed the options sliders during world creation.

### DIFF
--- a/data/core/world_option_sliders.json
+++ b/data/core/world_option_sliders.json
@@ -15,13 +15,13 @@
       {
         "level": 1,
         "name": "Rural",
-        "description": "Cities are twice as small and twice as far apart.",
+        "description": "Cities are half as large and twice as far apart.",
         "options": [ { "option": "CITY_SIZE", "type": "int", "val": 4 }, { "option": "CITY_SPACING", "type": "int", "val": 8 } ]
       },
       {
         "level": 2,
         "name": "Semirural",
-        "description": "Cities are 50% smaller and 50% farther apart.",
+        "description": "Cities are 25% smaller and 50% farther apart.",
         "options": [ { "option": "CITY_SIZE", "type": "int", "val": 6 }, { "option": "CITY_SPACING", "type": "int", "val": 6 } ]
       },
       {


sparr:
While updating the translations for 50%->25%, I think I found a bunch with inaccurate info for the other half of that description (some say half as far apart, instead of 50% farther apart, or something equivalent). I also think some of the translations for the other string, "twice as small" vs "half as large" should be clarified like the original change here but I'm not confident enough in those languages, and a few of them also seem to contain mistakes.

I'm unsure of the best way to proceed on the localization front.

GuardianDll:
I don't think you need to edit a translation files, they would be changed automatically
And translation.md point they must left untouched

dseguin:
That's strange, this PR should have triggered a warning comment from #55922. In any case, you shouldn't need to modify the `.po` files since they're automatically updated from Transifex every week.

sparr:
Do changes here also get pushed to Transifex? I'm not just changing the English strings; I'm also changing some of the translated strings.

dseguin:
Yeah, sometime ago the translation infrastructure was overhauled (more than once). For every master branch build, [this workflow](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/.github/workflows/push-translation-template.yml) updates Transifex with the new json strings, then translators will update the string for their locale, and on Fridays [this workflow](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/.github/workflows/pull-translations.yml) creates a PR to overwrite the `.po` and `.pot` files with the updated translations.

Even though your changes here may be fairly obvious, they would be overwritten anyway the next Friday. Some of this is described in the comment in the failed test: https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/3123878613/jobs/5066817951#step:2:99

BrettDong:
Please don’t edit any translations in the Git repository. Translation updates flow from Transifex to GitHub one-way.

sparr:
I've removed the translation commits.

PR #61276, opened 2022-09-26
Fix butane cans being inappropriately empty

onura46:
#### Summary
Bugfixes "Fix butane cans being inappropriately empty"

#### Purpose of change
Make sure butane cans in head shops are full; guarantee charges in item groups that would have spawned empty before.

#### Describe the solution
This is a small PR to fix empty butane cans spawning where full ones should have been. I updated the hardcoded spawns in the head shop - which was the issue that brought this to my attention - and made sure charges are explicitly defined in two item groups that previously didn't have them.

#### Testing
Tested on my own compiled version.

#### Additional context
I never opened a proper issue; just fixed it quickly after I noticed a comment from a user on Discord: "Did anyone ever saw a filled butane can/tank? Must be poor luck for me cause all cans I find, ~10 already, are empty. Even in Head Shops."

### DIFF
--- a/data/json/itemgroups/collections_domestic.json
+++ b/data/json/itemgroups/collections_domestic.json
@@ -233,7 +233,7 @@
       { "item": "denat_alcohol", "prob": 6 },
       { "item": "methed_alcohol", "prob": 4 },
       { "item": "textbook_extraction", "prob": 5 },
-      { "item": "butane_can", "prob": 10, "charges-min": 1 },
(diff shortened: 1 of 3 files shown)


PatrikLundell:
As far as I understand there is no requirement to create an issue prior to making a PR to fix a problem.

My understanding is that the issue has the advantage of allowing a discussion when there is doubt, potentially shutting an idea down before wasting time to implement it, or resulting in a better solution than the one originally envisioned without actually having to implementing the inferior one first.

PR #61278, opened 2022-09-26
Make alpha a bit smarter and nimbler

Dan1ss1mo:
#### Summary
Balance "Make alpha a bit smarter and nimbler"

#### Purpose of change
I was always wondering why alphas recive a bit smaller bonus to int and dex. Considering it's a "perfect human mutagen" it would make sense for it to make you equally good in all tasks. It makes even more sense now, considering alpha threshold problems.

#### Describe the solution
change Prime Dexterity and Prime Intelligence to require +2 versions of respective traits and abjust their bonus accordingly.

#### Testing
Chug some alpha primer and mutagen, aqire both Very Smart and Very Dextrous, both turn in to prime versions after crossing  threshold

#### Additional context
this uneven bonuses physically hurt my inner perfectionist, please merge this one

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -5051,7 +5051,7 @@
     "description": "You are smarter.  +2 Intelligence.",
     "types": [ "INT" ],
     "prereqs": [ "INT_UP" ],
-    "changes_to": [ "INT_UP_3", "INT_ALPHA", ],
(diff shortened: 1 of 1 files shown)


Perset:
gives a little more wiggle room to reach the threshold, too, IIRC there were a couple of PRs about that

PR #61282, opened 2022-09-26
audit and document monster speeds

LyleSY:
#### Summary
Balance "audit and document monster speeds"

#### Purpose of change

Balance

#### Describe the solution

Audit zombie animals and zombie animal upgrades for speed balance, document approach. Rebuilt to handle conflicts after new zombie horse upgrades. Also looked over insects and that looks about right.

N/A

#### Testing

Game loads no errors. Spawned a zombie horse and attempted to run away, was slowly murdered as intended

#### Additional context

Based on discussion in Devcord, especially from @Venera3 and Nebii and Kevin


Nebnis:
"Further, under no circumstances should a day one zed be more than 100% faster than base character speed."

I feel like zeds should be replaced with monster instead, because there has been quite a controversy and discussion about several day 1 monsters that appear randomly or frequently to be so much speedier than the character only to get killed as soon they peek out of the windows. Ie. Old wasp, old ladybug. 

I would also add something of the likes of. "Judgment of monster strength should be balanced upon their frequency, a faster or stronger creature could be okay in a special place(migo tower guards) but not so much in a forest or any house"

LyleSY:
I think what you're talking about is an entirely new section on monster difficulty scaling, which I agree would be useful and is actually possible now since we have quantified difficulty ratings. I'll give that a think but not for this PR

LyleSY:
That bit has now been put into a separate PR and merged there

PR #61302, opened 2022-09-27
Lets player construct earth ramp (low and high).

dan-netizen:
#### Summary
Content "Construct earth ramps"

#### Purpose of change
If one can construct concrete ramps with 5 units of concrete and 5 units of water (x2), one should be able to construct an earthen one with dirt, rocks and planks/sticks for reinforcement.

#### Describe the solution
Cloned all the aspects of concrete ramp building and made changes where appropriate.

Not adding this.

#### Testing
Spawned components, built earth ramp next to building, climbed on said building using ramp.

#### Additional context
Maybe fine tune the components.

### DIFF
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -3563,7 +3563,7 @@
     "pre_note": "Build an earth ramp leading to the next z-level above, and the corresponding ramp down leading from the z-level above to this level.  It must be built next to a low end of a ramp to allow moving between z-levels in both directions.",
     "category": "DIG",
     "required_skills": [ [ "fabrication", 3 ] ],
-    "time": "150 m",
+    "time": "300 m",
     "qualities": [ [ { "id": "HAMMER", "level": 2 } ], [ { "id": "DIG", "level": 2 } ] ],
     "components": [ [ [ "material_soil", 50 ] ], [ [ "stick", 10 ], [ "2x4", 8 ] ], [ [ "rock", 20 ] ] ],
     "pre_special": "check_ramp_high",


MNG-cataclysm:
@dan-netizen I've made a pull request to your branch which lints construction.json, merging that should fix the current errors.

dan-netizen:
> @dan-netizen I've made a pull request to your branch which lints construction.json, merging that should fix the current errors.

Thanks, I went to bed after the pull request.
I bookmarked the formatter for future use.
Cheers!

ashGlaw:
This is a great idea. Big improvement to the construction menu

dan-netizen:
The Windows build check fail, from what I saw in the details, was because of this error:
`01:51:21.638 ERROR : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\item.cpp:1501 [put_in] tried to put an item (talking_doll) count (1) in a container (purse) that cannot contain it: pocket unacceptable because not enough space`
My json code in the commit has nothing to do with any of these items.
What should I do?

MNG-cataclysm:
> The Windows build check fail, from what I saw in the details, was because of this error: `01:51:21.638 ERROR : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\item.cpp:1501 [put_in] tried to put an item (talking_doll) count (1) in a container (purse) that cannot contain it: pocket unacceptable because not enough space` My json code in the commit has nothing to do with any of these items. What should I do?

I'd try making an extremely small change in the code to restart the tests. Alternatively, you could ask one of the developers to restart that test.

MNG-cataclysm:
@Maleclypse Could you restart the Windows build test, please?

MNG-cataclysm (approved):

LyleSY:
Idea makes sense and I think you're going about it in a good way. Not sure I buy that a ramp can be shoveled together as quickly as the concrete can be poured and smoothed. I would think it would take twice as long and work half as well. Otherwise I think this looks good. I think you're building wooding frames, pounding them in place with sticks and your nice hammer, and then filling that in with earth which I think you're also pounding flat with your shovel?

LyleSY (commented):
I don't think this is a deal breaker, but I think that doubling the time to get this done is more realistic since you're not using all the wonderful modern equipment that would be used today. We don't have a quality system but if we did I'd suggest that this should be twice as breakable as the concrete version

dan-netizen:
> I don't think this is a deal breaker, but I think that doubling the time to get this done is more realistic since you're not using all the wonderful modern equipment that would be used today. We don't have a quality system but if we did I'd suggest that this should be twice as breakable as the concrete version

Thanks for the review, yeah it makes sense for the work to take longer.
I was thinking about the office complex that is being built 2 blocks from my place, and I saw the excavators and whatnot carry stuff from -2 basement level to +1 or more, on literally a mound of dug out earth that was growing daily.
Ofc, a lone man with a shovel and a hammer won't be able to compare to modern tools and equipment.
Regading the quality of the construction, I also agree, however it's meant for occasional use by a lonely survivor, not a metropolis highway,  so maybe when it will be possible, make a deathmobile-pass-until-break-counter, but a cart or bike shouldn't be a problem.

That being said, the time is now double, and more suggestions are welcomed.

LyleSY (approved):
Solid

PR #61314, opened 2022-09-27
Fix issue with dialog tags in EOC

### LINKED ISSUE #61188: Crash when using a variable name string in EOC dialogue
### Describe the bug

When attempting to use the variable name strings ((<full_name>, <city>, etc)) inside of an EOC, the game loads fine but then crashes as soon as the player triggers the EOC.

### Steps to reproduce

1. Add a name string to an EOC
2. Trigger the EOC in-game

### Expected behavior

The game doesn't crash and the string works as expected.

### Screenshots

Example JSON:
image

The resulting crash, which occurs as soon as the EOC triggers:
image

d9f3bb7

### Additional context

[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/9621030/crash.log)
[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/9621031/debug.log)

Ramza13:
#### Summary
None

#### Purpose of change
Fixes #61188

### DIFF
--- a/src/npctalk.cpp
+++ b/src/npctalk.cpp
@@ -1552,11 +1552,7 @@ void parse_tags( std::string &phrase, const Character &u, const Character &me,
         }
 
         const item_location u_weapon = u.get_wielded_item();
-#pragma clang diagnostic push
-#pragma clang diagnostic ignored "-Wtautological-compare"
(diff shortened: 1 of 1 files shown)


pshepjr:
Thanks for this!

PR #61319, opened 2022-09-28
Xl punching armor

Nebnis:
#### Summary
Content "Added a variety of XL and XS versions of all the punching armor/knuckles"

#### Purpose of change
Mutants like ursine who acquired mutations like broad paw were unable to wear most of the punching equipment because they lacked their respective size variants since they have been changed to be worn instead of wielded. 

#### Describe the solution
I have added XL and XS variants for most of the unarmed equipment. And also added a crafting recipe for brass knuckle since it could only be acquired from prisons.

#### Testing
Logged and checked most of the gear, everything seems to be in order unless i missed something.

#### Additional context
If anyone got any good idea i would love to add more ways of making unarmed  a lil bit better, since they feel kinda underwhelming right now. Maybe unique equipment for mutants who can use their physiology better than humans?

### DIFF
--- a/tools/spell_checker/dictionary.txt
+++ b/tools/spell_checker/dictionary.txt
@@ -305,6 +305,7 @@ ceratopsian
 ceratosaurus
 cere
 cerium
+cesti
 cestus
 chainmail
 chamberings


MNG-cataclysm:
You need to add "cesti" to the tools/spell_checker/dictionary.txt file, so it won't produce an error with future work.

MNG-cataclysm (approved):

MNG-cataclysm:
Good job with this, I like what you've done.

Issue #51689, opened 2021-09-18
Bring drug brand names back


Entity-y:
# Is your feature request related to a problem? Please describe.
 I recently noticed a setting in the Interface menu to toggle the use of brand names for guns. This made me think about the removal of real world drug names in favor for tiny descriptions.  See #45308
# Describe the solution you'd like
Add an option to toggle between real world drug names, active ingredient, or as it is now with a tiny description.
# Describe alternatives you've considered
Use snippets for the various drug brands in the description.
# Additional context
![image](https://user-images.githubusercontent.com/50155659/133884776-1714eaae-73be-4b4b-b01a-a8a2d4c07393.png)
I created a [mod](https://github.com/Entity-y/OG_Drugs/blob/main/OG_Drugs/data/json/items/comestibles/med.json)  awhile back that restores the drug names. Not sure how to add a new setting to toggle this.

anothersimulacrum:
It would look something like this
```diff
diff --git a/data/json/items/comestibles/med.json b/data/json/items/comestibles/med.json
index b7a3a9d5ff..633082bfc3 100644
--- a/data/json/items/comestibles/med.json
+++ b/data/json/items/comestibles/med.json
@@ -25,6 +25,7 @@
     "comestible_type": "MED",
     "name": { "str_sp": "prescription stimulant" },
     "description": "Medical grade amphetamine salts mixed with Dextroamphetamine salts, commonly prescribed to treat hyperactive attention deficits.  It suppresses the appetite, and is quite addictive.",
+    "variants": [ { "id": "adderall", "weight": "1", "name": { "str_sp": "Adderall" }, "description": "You fill this out" } ],
     "weight": "1 g",
     "volume": "250 ml",
     "price": 14500,
diff --git a/src/item.cpp b/src/item.cpp
index 149a93b490..ed2c38deae 100644
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -7265,6 +7265,8 @@ bool item::has_itype_variant( bool check_option ) const
     switch( type->variant_kind ) {
         case itype_variant_kind::gun:
             return get_option<bool>( "SHOW_GUN_VARIANTS" );
+       case itype_variant_kind::drug:
+           return get_option<bool>( "SHOW_DRUG_VARIANTS" );
         default:
             return true;
     }
diff --git a/src/itype.h b/src/itype.h
index 54120442b9..ff540f37e3 100644
--- a/src/itype.h
+++ b/src/itype.h
[...]

PR #61321, opened 2022-09-28
added option to include drug names again (#51689)

RobertFontan:
#### Summary
Features "Added an option to toggle between drug brand names and generic functional names"

#### Purpose of change

#51689 Added this option for players who prefered the older version of the game. which had brand names. 

#### Describe the solution

Added a variant to the med.json adding brand names to item desciptions and use actions.  item.cpp, options.cpp, and item.h adds functionality for a toggle like the "show gun brand" option

#### Testing

Ran it until I found prescription stimulants and toggled the button to ensure it worked. (Should display adderall brand name in description)

### DIFF
--- a/src/itype.cpp
+++ b/src/itype.cpp
@@ -58,6 +58,8 @@ std::string enum_to_string<itype_variant_kind>( itype_variant_kind data )
             return "gun";
         case itype_variant_kind::generic:
             return "generic";
+        case itype_variant_kind::drug:
+            return "drug";
         case itype_variant_kind::last:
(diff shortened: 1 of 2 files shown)


anothersimulacrum (changes requested):
This doesn't look like it'll work. You're adding the variants, but not telling the game they are drug variants, or even providing a way for the game to be that they are drug variants. As is, it looks like it renames all the drugs, and adds an option that does nothing.

RobertFontan:
still a baby when it comes to contributing, let me know what can be improved!

PR #61323, opened 2022-09-28
rebalance deathrattle serpent

sonphantrung:
#### Summary
Balance "rebalance deathrattle serpent"

#### Purpose of change
From my old Discord memories, there came up a conversation about how STUPIDLY POWERFUL the deathrattle serpent is: 90kg, 70L (much lower than hulks), but the damage is `8d8 bash`. HOW OP IT IS!

#### Describe the solution
Rebalanced the serpent: `6d8 bash + 12 cut` now.

Not doing this

#### Testing
Should be good ;)

### DIFF
--- a/data/json/monsters/reptile_amphibian.json
+++ b/data/json/monsters/reptile_amphibian.json
@@ -286,7 +286,7 @@
     "melee_skill": 9,
     "melee_dice": 6,
     "melee_dice_sides": 6,
-    "melee_damage": [ { "damage_type": "cut", "amount": 12 } ],
+    "melee_damage": [ { "damage_type": "stab", "amount": 8 } ],
     "dodge": 5,
     "armor_bash": 3,
     "armor_cut": 4,


Nebnis:
I think it should be lowered even further. A hulk got 4d8 for reference, so it should be that or lower maybe increase cut again. The snake already have a deadly poison and rattlesnakes are not known for being constrictors.

MNG-cataclysm (approved):

data/json/monsters/reptile_amphibian.json
```diff
@@ -284,9 +284,9 @@
     "morale": 100,
     "aggro_character": false,
     "melee_skill": 9,
-    "melee_dice": 8,
+    "melee_dice": 6,
     "melee_dice_sides": 8,
-    "melee_damage": [ { "damage_type": "cut", "amount": 8 } ],
+    "melee_damage": [ { "damage_type": "cut", "amount": 12 } ],
```
Maleclypse, line 289:
```suggestion
    "melee_damage": [ { "damage_type": "stab", "amount": 8 } ],
```

data/json/monsters/reptile_amphibian.json
```diff
@@ -284,9 +284,9 @@
     "morale": 100,
     "aggro_character": false,
     "melee_skill": 9,
-    "melee_dice": 8,
+    "melee_dice": 6,
     "melee_dice_sides": 8,
```
Maleclypse, line 288:
```suggestion
    "melee_dice_sides": 6,
```

data/json/monsters/reptile_amphibian.json
Maleclypse: These bites should probably be stab instead of cut so I've shrunk their damage output.

Maleclypse (approved):

LyleSY (approved):
Love it, puts this monster's damage within the range in the balance doc, fits under "Very High Damage" https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/GAME_BALANCE.md

PR #61328, opened 2022-09-28
tweaks to basic kevlar vest

anoobindisguise:
#### Summary
Balance "kevlar vest rebalance"

#### Purpose of change
Kevlar vest was still using the old style of armor, and had some issues because of this. it was made of layered kevlar despite being an abnormal thickness compared to the standard 4.4mm. It also occupied the skintight layer despite that layer no longer being intended for such thick gear.

#### Describe the solution
Kevlar vest -> Kevlar torso armor, as it is no longer a "vest" (compared to the other ballistic vests which are OUTER and have storage) and now occupies the normal layer. made of 0.5mm nylon and 3.3mm kevlar. Weight reduced. 
I searched for numerous "concealable ballistic vests" and found most were level II or IIA, and that the thinnest one I could find was 3.5mm, and that level II typically caps out at about 4.2mm (1/6th inch). Also, these were roughly 1.5-2kg depending on body fit. Coverage is 94% on upper torso and 85% on lower torso, this should hopefully about match up with real life. Although some of these vests were depicted like "tank tops" worn directly on the skin, many more were worn on top of t-shirts, so NORMAL layer feels right to me. Encumbrance lowered to 3 since it is lighter than a III or IIIa vest.

Different coverage, lower torso coverage may be slightly too high. 
move it from torso_armor to ballistic_armor

### DIFF
--- a/data/json/items/armor/torso_armor.json
+++ b/data/json/items/armor/torso_armor.json
@@ -974,8 +974,8 @@
     "repairs_like": "survivor_suit",
     "type": "ARMOR",
     "category": "armor",
-    "name": { "str": "Kevlar torso armor" },
-    "description": "A lightweight level II ballistic vest, suitable for wearing under a coat or other outerwear.",
+    "name": { "str": "Kevlar vest" },
+    "description": "A lightweight level II ballistic vest, suitable for wearing under a coat or other outer wear.",
     "weight": "1960 g",
     "volume": "4 L",
     "price": 35000,
@@ -993,7 +993,7 @@
       {
         "material": [
           { "type": "nylon", "covered_by_mat": 100, "thickness": 0.5 },
-          { "type": "kevlar", "covered_by_mat": 100, "thickness": 3.3 }
+          { "type": "kevlar", "covered_by_mat": 95, "thickness": 3.3 }
         ],
         "encumbrance": 3,
         "coverage": 94,
@@ -1004,7 +1004,7 @@
       {
         "material": [
           { "type": "nylon", "covered_by_mat": 100, "thickness": 0.5 },
-          { "type": "kevlar", "covered_by_mat": 100, "thickness": 3.3 }
+          { "type": "kevlar", "covered_by_mat": 95, "thickness": 3.3 }
         ],
         "//": "encumbrance handled above",
         "encumbrance": 0,


data/json/items/armor/torso_armor.json
```diff
@@ -974,22 +974,46 @@
     "repairs_like": "survivor_suit",
     "type": "ARMOR",
     "category": "armor",
-    "name": { "str": "Kevlar vest" },
-    "description": "A lightweight, bullet resistant vest.  Suitable for wearing under clothing.",
-    "weight": "2860 g",
-    "volume": "6 L",
-    "price": 80000,
+    "name": { "str": "Kevlar torso armor" },
+    "description": "A lightweight level II ballistic vest, suitable for wearing under a coat or other outerwear.",
```
MNG-cataclysm, line 978:
```suggestion
    "description": "A lightweight level II ballistic vest, suitable for wearing under a coat or other outer wear.",
```

data/json/items/armor/torso_armor.json
```diff
@@ -974,22 +974,46 @@
     "repairs_like": "survivor_suit",
     "type": "ARMOR",
     "category": "armor",
-    "name": { "str": "Kevlar vest" },
-    "description": "A lightweight, bullet resistant vest.  Suitable for wearing under clothing.",
-    "weight": "2860 g",
-    "volume": "6 L",
-    "price": 80000,
+    "name": { "str": "Kevlar torso armor" },
```
MNG-cataclysm, line 977:
```suggestion
    "name": { "str": "Kevlar vest" },
```
I think vest works better for this situation. "Torso armor" sounds a bit unnatural.

MNG-cataclysm (changes requested):

MNG-cataclysm (approved):
You've done a good job with this.

MNG-cataclysm:
The Windows build test is failing because of this:
 20:51:10.318 ERROR : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\item.cpp:1501 [put_in] tried to put an item (creepy_doll) count (1) in a container (purse) that cannot contain it: pocket unacceptable because not enough space

@Maleclypse Could you restart the Windows build test, please?

Maleclypse:
Can you make an issue regarding the failure of the test on `20:51:10.318 ERROR : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\item.cpp:1501 [put_in] tried to put an item (creepy_doll) count (1) in a container (purse) that cannot contain it: pocket unacceptable because not enough space`

Issue #61308, opened 2022-09-27
Artifacts powers erased


Blorkman:
### Describe the bug

Since today updating, then saving your game with artifacts will delete their effects.

### Steps to reproduce

1.Update
2.Load a save with artifacts
3.Save
4.Reload
5.Your artifacts are now useless shiny rocks !

### Expected behavior

Artifacts effects being saved obviously.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.2006 (21H2)
- Game Version: 5542e8a [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

Ramza13:
Crap this is probably caused by #61258.  Do you happen to have a save with artifacts that can reproduce this issue?

anoobindisguise:
I've found the same thing. All my magical napkins, gone.

KymasTuran:
> Crap this is probably caused by #61258. Do you happen to have a save with artifacts that can reproduce this issue? Or actually can you tell me how you get artifacts in game? I've never encountered them and am not sure how to get them which makes debugging hard.

The "easy" way is in "Amigara Mines".

Blorkman:
> Crap this is probably caused by #61258. Do you happen to have a save with artifacts that can reproduce this issue? Or actually can you tell me how you get artifacts in game? I've never encountered them and am not sure how to get them which makes debugging hard.

I do have a save with two artifacts on my character, i found them in a distorted lab. I'd be happy to share it with you if you need it, but mind its 500mb large

KymasTuran:
I can confirm:
- Artifacts made in a new game work as intended, I spawn one using the debug menu.
- Importing saves to a new version and artifacts lost all is data, discover it after browsing and compare save games.
In less words:  the problem is importing old enchant system. New games are ok. 
I guess a warning somewhere is mandatory.
Sorry for my poor english.

anoobindisguise:
If I create a new save file and debug spawn an artifact (max abilities 10 power 10000 max negative power -1000) I can clearly see it has a number of abilities but as soon as I save and quit it loses all of them except for its actives.

KymasTuran:
[...]

PR #61341, opened 2022-09-29
Artifact "resonance" penalties from carrying too many

### LINKED ISSUE #60966: Consequences for carrying (too many) artifacts

Now that artifacts provide buffs while in your back pocket like in S.T.A.L.K.E.R. you can get pretty insanely powerful by carrying lots of them around. Although they are typically pretty hard to get (locations include portal storms, strange temples, mine finales, distorted labs, and a few others) and they have an opportunity cost of taking up space in your inventory depending on how big the artifact is, there should still probably be some increasingly dire limitations to having too many, especially with the spooky XEDRA anomalous artifact reports seen near them.

### Solution you would like.

Some examples:
- Teleglow
- Micro-teleports (safe teleport, move 1 space) - potentially fatal due to car driving, should probably be prohibited while driving a car
- Radiation (same as in STALKER)
- Attention
- Artifacts straight up vanishing from reality (only in extreme cases, from carrying way too many)
- EMPs?
- Other malicious effects?

effects should probably only kick in after carrying more than 1 artifact, and be minor unless you really get greedy. I don't know how specifically it should be balanced. 

_No response_

### Additional context

_No response_

anoobindisguise:
#### Summary
Balance "increasing levels of discouragement against carrying too many artifacts"

#### Purpose of change
Fixes #60966
Artifacts are busted OP right now. There needs to be a hard limit on the number of artifacts you can hoard.

#### Describe the solution
Each artifact is now created with a value called ARTIFACT_RESONANCE (a new enchantment) equal to its power, so stronger artifacts are more resonant. Then, if you have more than 2000 resonance, roughly once per hour, divided by how many 2000s of resonance you have (4000 resonance = 1/halfhour, 6000 = 1/20min, 8000 = 1/15min, 10000 = 1/12min) you get a negative effect.
There are 4 "tiers" of negative effects, of varying severity. Each negative effect has a 50% chance to "downgrade" to a lower tier, recursively. So if you are with enough resonance to hit the "tier 3" resonance, you have a 50% chance to get a "tier 2" effect instead and a further 50% chance to get a "tier 1" effect instead of that.

Tier 1 effects:
- Minor pain
- Brief hallucinations (doesn't spawn "fake enemies", just gives you the stat debuff)
- Nausea

Tier 2 effects:
- Safe microteleport (1 space away, does not occur while driving to prevent instant deaths)
- Minor radiation (blocked by rad resistant gear)
- extremely loud noise (louder than .50 bmg)

Tier 3 effects:
- Unsafe teleport (up to 12 tiles)
- Agonizing pain (makes you yell too)
- Deadly radiation (partially pierces rad resistant gear)

Tier 4 effects:
- Incorporeal
- Tindalos
- Dimensional fatigue

You should be able to carry 2-3 artifacts safely depending on their power, but after that things start getting out of hand. Tier 4 penalties in particular are meant to be so bad that they are basically not survivable for any period of time, but tier 1 penalties should be minor nuisances and tier 2 penalties can be avoided with the right gear (dimensional anchor, radproof gear, hearing protection). I've never seen an artifact spawn with more than 1500 total power.

Different values at which point resonance occurs. Different consequences for resonance.

#### Testing
Confirmed effects are working at each level of artifact resonance, observed this by generating fixed-power artifacts to artificially adjust that enchantment value. Confirmed looking in game files that the artifacts generate with the ARTIFACT_RESONANCE field if they have any power and they apply those effects properly to the player when hoarded, then stop causing those effects when dropped.
It is currently very difficult to test this due to #61308 causing breakages, because the artifact abilities only exist for the current game session. I saw numerous bugs relating to those passive abilities disappearing that I hope are not in any way associated with this PR.


John-Candlebury:
Dont get me wrong I think this is a good idea for vanilla artifacts, but I have serious doubts about this being mandatory. I know I wouldn't want Mi-go and golden age humanity "artifacts" to have this in Aftershock. 

Presumambly KorgGent would like to disable this too for random Magiclysm artifacts too.

Venera3:
In addition to that, this would be more expandable as an item property->EoC interaction instead of hardcoding it - either counting flags or a new `artifact_resonance` field. EoC can also do more fun stuff by now.

PatrikLundell:
The description is somewhat contradictory in that it talks about "hoard", but then goes on to describe carrying. Hoarding, to me, means collecting, which would mean a hoarding issue would make usage of artifact storage zones dangerous.

You might be able to hoard a lot of artifacts in a zone and then explain it away by saying interaction only has an effect when carried, though, because the interactions happen within the carrier.

anoobindisguise:
This will only apply to procgen artifacts, anything fixed will not have these. Magiclysm will be exempt, your bracers of protection will not resonate. You can also give the player an enchantment to multiply artifact resonance by 0 in AFS to prevent this system from working. I don't know if I have the chops to do a more sophisticated workaround, but a hacky solution would be to give the altered objects and art objects a flag to count towards resonance penalties leaving custom migo stuff separate. I definitely also don't have any experience with EoCs to work it in JSON unfortunately.

As far as hoarding goes current character is carrying like 50 and has 400+ speed, stats in the high 60s, and can carry over 4k lbs so some limitation is definitely in order, it's easy enough to flavor "hoarding" them as your human body can only handle so many artifacts affecting you at once, or artifacts only have effects when held close to a living being.

pantherpusher:
Would it be possible to make this feature toggleable with the default set to true, for people who either are carrying enough to be in big trouble if/when this merges, or just for people who don't want the mechanic.

Maleclypse:
> Dont get me wrong I think this is a good idea for vanilla artifacts, but I have serious doubts about this being mandatory. I know I wouldn't want Mi-go and golden age humanity "artifacts" to have this in Aftershock.
> 
> Presumably KorgGent would like to disable this for random Magiclysm artifacts too.

I do want this for Xedra Evolved random artifacts but I'd like to be able to change the effects that occur.

Maleclypse:
> Would it be possible to make this feature toggleable with the default set to true, for people who either are carrying enough to be in big trouble if/when this merges, or just for people who don't want the mechanic

Toggles are frowned upon.  Jsonized is accepted.

anoobindisguise:
OK, so my current plan for this is to include a bool called 'resonant' which determines whether the artifact will be given resonance or not, but I'm not sure whether to define it in the procgen JSON (so all artifacts of a specific type like 'cult' or 'nethereum_tunnels' will get the resonance) or to define in the artifact generation rules and default to false. Obviously as it isn't compiling rn I need to work on it.

anoobindisguise:
Success! Any artifact can be defined in JSON to have ARTIFACT_RESONANCE or not, and by default it's false. You can also decide this when spawning the artifact through debug. I spawned a ridiculously powerful artifact with no resonance and had no adverse effects, and then spawned a moderately powerful artifact with resonance and then got tier 2 penalties. I think I got all the artifact spawns in game; let me know if I missed one.

Maleclypse:
You've missed the procgen artifacts in xedra evolved.

Maleclypse (changes requested):
Thanks for adding the XE procgen group!  Please add documentation of resonance to ARTIFACTS.md

doc/ARTIFACTS.md
```diff
@@ -101,3 +101,12 @@ This provides a list of possible items that this artifact can spawn as, if it ap
...
+
+Resonance is assigned to artifacts in the JSON responsible for adding them to mapgen, where artifacts are given a power level and a set of procgen data to draw from. If flagged as true, this artifact will have resonance. If it does not include this optional flag, it will default to false.
+
+Resonance represents the potential dangers of artifacts on the human body to counterbalance their positive effects.  For the purposes of lore, artifacts should primarily cause such resonance effects when they are affecting a human body - when held close at hand, similar to artifacts in S.T.A.L.K.E.R.  Resonance effects are currently hardcoded, and scale with the amount of resonance you have, beginning at 2,000 resonance (2,000 total power of artifacts carried).  Resonance of 2,000 causes an effect roughly every hour, but effect also occur more frequently the more resonance you have, becoming 100% more common for every whole 2,000 resonance. There are 4 "tiers" of effects in increasing severity kicking in at 2,000, 4,500, 7,500, 12,500 resonance. Each "tier" of resonance has a 50% chance to downgrade to a lower tier of severity, recursively, so lower-severity effects never go away.
```
Maleclypse, line 110: Don't use references to other games to explain effects.   I've never played STALKER and might not ever.
```suggestion
Resonance represents the potential dangers of artifacts on the human body to counterbalance their positive effects.  For the purposes of lore, artifacts should primarily cause such resonance effects when they are affecting a human body - such as when carried or wielded by the PC.  Resonance effects are currently hardcoded, and scale with the amount of resonance you have, beginning at 2,000 resonance (2,000 total power of artifacts carried).  Resonance of 2,000 causes an effect roughly every hour, but effect also occur more frequently the more resonance you have, becoming 100% more common for every whole 2,000 resonance. There are 4 "tiers" of effects in increasing severity kicking in at 2,000, 4,500, 7,500, 12,500 resonance. Each "tier" of resonance has a 50% chance to downgrade to a lower tier of severity, recursively, so lower-severity effects never go away.
```
Give an example please.   I'm not sure if you are saying that higher tier effects have a chance to add additional lower tier effects or that there is a chance for higher tier effects to transform into a lower tier effect instead of applying the higher tier effect.

doc/ARTIFACTS.md
```diff
@@ -101,3 +101,12 @@ This provides a list of possible items that this artifact can spawn as, if it ap
...
+
+These effects should be relatively vague - it should not be explicitly clear that carrying multiple artifacts is what's causing a player's misfortune, but it should be implied due to the effects not happening until the player picks up a bunch of cool-looking napkins and rubick's cubes.  Tier 1 effects are nausea, low-strength hallucinations, and minor pain. Tier 2 effects are minor radiation (blocked by rad resistant gear), safe micro-teleports and a very loud noise.  Tier 3 effects include severe pain, severe radiation (not fully blockable by gear), and unsafe teleportation.  Tier 4 effects include hounds of Tindalos, incorporeal, and creating a tear in reality. The resonance factor of an artifact should be measurable, even if the other properties of the artifact are not, although at time of writing there is no way to do so.
```
Maleclypse, line 112: Do these resonance effects also occur to NPCs? I'm not even sure Artifacts affect NPCs in general, they probably do but I've never checked.  What happens if you give Smokes 40k worth of artifact power?

Maleclypse (commented):
This is definitely improving.  Thank you!

doc/ARTIFACTS.md
anoobindisguise: Smokes will asplode, yes. Suffer.cpp occurs on NPCs and they are also affected by enchants.

Maleclypse (approved):

PR #61351, opened 2022-09-29
Fix Chickens can hatch from fried and scrambled eggs

### LINKED ISSUE #61300: Chickens can hatch from fried and scrambled eggs
### Describe the bug

When the aforementioned comestibles rot, chickens may hatch.

### Steps to reproduce

1. Craft scrambled or fried eggs.
2. Let them rot.
3. Have some chickens.

### Expected behavior

Chickens do hatch from cooked eggs.

The cause for this is actually quite simple:

https://github.com/CleverRaven/Cataclysm-DDA/blob/c20efab8c7a7b6684afa12a3abc508c4b33181a7/data/json/items/comestibles/egg.json#L741-L743

which inherits:

https://github.com/CleverRaven/Cataclysm-DDA/blob/c20efab8c7a7b6684afa12a3abc508c4b33181a7/data/json/items/comestibles/egg.json#L27-L31

The solution would be to set the chance to 0 and/or the resulting monster to `mon_null`. Not sure what the way to go here is exactly, so feedback welcome.

### Screenshots

_No response_

n/a

### Additional context

There may be other affected items, too.

Originally reported by `az-ssh` on IRC. I asked them to report this but they apparently forgot.

```irc
00:47 <az-ssh> "Something has crawled out of the fried eggs (rotten)!"
00:47 <az-ssh> What a resilient chicken!
00:48 <az-ssh> and what a disturbing sentence!
01:09 <az-ssh> When I read it I was extremely worried, especially since I ate the other ones before they went rotten
01:10 <az-ssh> I know it's technically a bug (not sure where to report.. github site?) but tbh that is some quality content.
01:10 <az-ssh> feature level.. and now I have chickenfriend
```

FooterReal:
#### Summary
Bugfixes "Fixes some egg foods spawning chickens"

#### Purpose of change

Fixes #61300

#### Describe the solution

Overwritten spawn group to 'GROUP_NULL' and spawn chance to '0' in egg.json for all food items that copy from egg_bird or egg_chicken. 

#### Testing

I've built cdda (tiles), created a new character, added 10 of each food item using debug and then waited for them to rot. No chickens spawned. It's not perfect but hopefully enough.

### DIFF
--- a/data/json/items/comestibles/egg.json
+++ b/data/json/items/comestibles/egg.json
@@ -734,7 +734,9 @@
     "material": [ "egg" ],
     "volume": "50 ml",
     "flags": [  ],
-    "fun": -2
+    "fun": -2,
+    "rot_spawn": "GROUP_NULL",
+    "rot_spawn_chance": 0
   },
   {
     "type": "COMESTIBLE",
@@ -751,7 +753,9 @@
     "volume": "125 ml",
     "charges": 2,
     "flags": [ "EATEN_HOT", "FREEZERBURN" ],
-    "fun": 3
+    "fun": 3,
+    "rot_spawn": "GROUP_NULL",
+    "rot_spawn_chance": 0
   },
   {
     "type": "COMESTIBLE",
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):

NetSysFire (approved):
Cool! Thanks for looking into other foods which are also affected.

PR #61353, opened 2022-09-29
Imperialchanges

myuksel123:
#### Summary

Features "Height matches imperial setting when making a new character"

#### Purpose of change

When the user selected imperial for height units in the settings interface, it would show the height of the character in an imperial manner during the game, but it would not when the user was creating a new character. The user would have to create the new character using metric centimeters for height regardless if they chose imperial in settings.

#### Describe the solution

I found this problem in an issue someone else had posted, and someone had found a solution to display the height in imperial when creating a new character, but pressing to the right and left would still only add/subtract a centimeter, so the user would have to press to the right/left 2-3 times to move the height up and down an inch. I changed the mod_base_height function in character.cpp to add/subtract 2.54 centimeters (which corresponds to one inch) when the height settings were set to imperial. To do this, I had to change the variable init_height in character.h to a float so that I could add the decimal. This made it so when the user pressed to the right/left it would modify by an inch in imperial settings. In the newcharacter.cpp file I just put the solution another GitHub user had found in the issue to display the height as imperial when making a new character.

#### Testing
I compiled and ran Cataclysm-DDA with the new changes on my computer, and it works (at least in the console version it does, I don't have Gentoo).

### DIFF
--- a/src/newcharacter.cpp
+++ b/src/newcharacter.cpp
@@ -3346,7 +3346,6 @@ static void draw_height( const catacurses::window &w_height, const avatar &you,
     werase( w_height );
     mvwprintz( w_height, point_zero, highlight ? COL_SELECT : c_light_gray, _( "Height:" ) );
     unsigned height_pos = 1 + utf8_width( _( "Height:" ) );
-/*prints out height according to imperial or metric setting */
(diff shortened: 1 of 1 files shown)


Maleclypse:
Obsoleted by #61055

PR #61361, opened 2022-09-30
Aftershock: Smartguns

John-Candlebury:
#### Summary
Mods "Aftershock: Add a pistol and sniper smartguns"

#### Purpose of change

Add more high tech ballistic weapons. In this case very easy to aim Wraitheon smart weapons.

#### Describe the solution

Smart guns fire 0 dispersion ammo and have an enchantment that greatly increases the players precision. So long as you don't fire completely unaimed shots, you have an almost guaranteed chance of scoring a good hit on your opponent.

Because you arent actually aiming anything, these wont train marksmanship, and use a new "smartgun" skill.  Your skills still affects how quickly you can acquire a target with a smart gun.

Theres two guns added here:

The Wraitheon W1011 Silvergram: An elegant and very expensive 10mm  PDW made for company executives. The related profession starts with that one.

The Wraitheon Armory 2519. 25mm smart sniper that is slightly slow to aim but guarantees good hits at even max range.

The executive also starts with a custon nanofab template that allows them to craft the ammo.

#### Testing

Fired both guns.

#### Additional context

Also part of the very large building PR I'm in the process of splitting.

### DIFF
--- a/data/mods/Aftershock/items/gun/voltaic.json
+++ b/data/mods/Aftershock/items/gun/voltaic.json
@@ -24,34 +24,5 @@
     ],
     "ammo_effects": [ "ELECTRO_ANOMALY" ],
     "flags": [ "NEVER_JAMS", "NON_FOULING", "NEEDS_NO_LUBE" ]
-  },
-  {
-    "id": "afs_wm14_volt",
-    "type": "GUN",
-    "copy-from": "laser_rifle",
-    "name": { "str": "WM-14 Volt" },
-    "description": "pulsed electic discharges common with both public and private security outfits.",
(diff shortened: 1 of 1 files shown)


TheShadowFerret:
Only criticism of this is that the ammo's damage should scale up with range, if it's based on the Gyrojet guns

https://www.youtube.com/watch?v=W3IiWjdbUZg
https://www.youtube.com/watch?v=98c2t_uK5Uo

John-Candlebury:
I know yeah but it seems like unneeded complexity atm, especially because I think the changes would have to be slightly invasive.

Issue #60724, opened 2022-09-03
Add voluntary field Reload_Time for individual rounds of ammo


Constellati:
### Is your feature request related to a problem? Please describe.

For weapons that have multiple types of ammunition, particularly those with ammunition with different qualities or weapons without external magazines, the ammo used can greatly increase or decrease the load time of the gun. I would like this to be able to be easily modeled in game.

The most pertinent example I can think of would be the use-case of muzzleloading firearms, where the ammo used can make loading the gun significantly faster. 

For instance, loading a rifled musket takes one to two minutes even for a skilled user, in real life; however when loading with a technologically superior minie ball cartridge or a modern, saboted projectile, rate of fire increases dramatically to around three aimed shots a minute in skilled hands, or somewhere in the range of 15-20 seconds to load the gun.

Another example would be muzzleloading cap-and-ball revolvers, for which necessarily careful loading of loose shot is a very time consuming ordeal on the order of about ten minutes, whereas loading with properly greased and nitrated paper cartridges can be done in about a third of the time.

For modern examples, revolvers and other common breech loaders like tube-fed shotguns and lever action rifles, and internal-magazine bolt actions, may benefit from having a per-shell reload time, although I believe they already function in such a fashion.
Another potential and deficit is if someone wished to model the time it takes to insert loose rounds into a magazine in a different fashion.

### Solution you would like.

I would like if there was an optional field for ammunition that describes the time to reload that specific type of projectile, to differentiate it from the armament that is being reloaded, much in the way magazines can augment the base reload time of certain firearms.

This would permit me to go forward with modeling appropriate, individual-ammo-based reload times for muzzleloaders, and provide a means for alternative, more realistic reload times in the future for other types of weapons as well, should someone wish to model those.

### Describe alternatives you have considered.

I have thought about using loose, "uncased" powder and shot--which is still a functional albeit slow means of loading a muzzleloader in real life--as the actual projectile itself, and treating the paper cartridge as a 'magazine' so as to make use of the existing reload time functionality.

However, this is an awkward and unsatisfactory method of solving the problem due to many unnecessary crafting steps and a potential inability to destroy used 'cartridge-magazine' things on use.

I have also considered simply dropping the patched round ball and hand-waving all muskets, rifled or no, to use minie balls, as for fighting purposes they are in every regard superior;

[...]

PR #61373, opened 2022-09-30
Adding new reload type for muzzle loading firearms

Constellati:
Attempted reload changes with flintlocks being used and a noticeable test time, 10000 turns.

#### Summary

Features "I have tried to enumerate an optional variable to lengthen reload times based on loose ammo being loaded."

#### Purpose of change

I am attempting to program the feature I proposed in #60724.

#### Describe the solution

I have added an optional integer variable, loose_reload_time, to ammo, and have tried to define it fully and then call it as an addition to the reload time of a firearm when said firearm is being loaded from a single bullet in the character's inventory, not contained in a magazine.

I outline those in #60724.

#### Testing

At first, I tried programming this and it threw several errors, mainly with a difficulty to compile and also a undefined memory problem.
I replaced the variable, which I originally had named reload_time for simplicity's sake, with a new variable loose_reload_time, and the errors stopped--but on the other hand, nothing seems to have changed.

### DIFF
--- a/data/json/items/ammo/flintlock.json
+++ b/data/json/items/ammo/flintlock.json
@@ -12,6 +12,7 @@
     "symbol": "=",
     "color": "white",
     "count": 30,
+	"loose_reload_time": 10000,
     "stack_size": 10,
     "ammo_type": "flintlock",
     "range": 6,
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -9611,7 +9611,7 @@ int item::get_reload_time() const
     }
     int reload_time = 0;
     if( is_gun() ) {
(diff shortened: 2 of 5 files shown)


Maleclypse (changes requested):
You'd need to document this in the gun section of JSON_INFO.md. And you'll need to lint your changes.

edit: and astyle as well.

Constellati:
I've decided to close this PR.
It's proven to be REALLY damn hard to program, and I don't think it's a wise or particularly elegant solution now that I've gotten into the code intimately and (mostly) understand how it works.
Furthermore, my only real reason for starting it in the first place isn't appropriate, which I now understand as well.

PR #61375, opened 2022-09-30
yarn is unsalvageable

### LINKED ISSUE #61331: Cutting up 1 yarn results in multiple felt patches.
### Describe the bug

Cutting up yarn results in several felt patches which can be disassembly for more yarn, creating an infinite loop of extra materials.

### Steps to reproduce

1. Use any knife to cut up yarn which results in several felts patches

### Expected behavior

Dunno, i think yarn is supposed to turn into something else

### Screenshots

image

- OS: Windows
- Game Version: 5542e8a [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Slots [cbm_slots],
]

### Additional context

_No response_

anoobindisguise:
#### Summary
none

#### Purpose of change
fixes #61331

### DIFF
--- a/data/json/items/ammo.json
+++ b/data/json/items/ammo.json
@@ -233,6 +233,7 @@
     "looks_like": "thread",
     "description": "Wool yarn, could be used to knit wool clothing.",
     "material": [ "wool" ],
+    "flags": [ "NO_SALVAGE" ],
     "volume": "250 ml",
     "weight": "1 g",
     "ammo_type": "thread",


Maleclypse (changes requested):
Yarn literally turns into felt.

https://support.lionbrand.com/support/solutions/articles/17000035404-felting-techniques#:~:text=Yarns%20made%20of%20wool%20(non,together%20and%20%E2%80%93%20voil%C3%A0%20%E2%80%93%20felt!

edit: if there is a problem it's that we may be turning it into too much felt? but if you can turn yarn into an equivalent weight of felt that seems like it would be correct.

anoobindisguise:
> Yarn literally turns into felt.
> 
> https://support.lionbrand.com/support/solutions/articles/17000035404-felting-techniques#:~:text=Yarns%20made%20of%20wool%20(non,together%20and%20%E2%80%93%20voil%C3%A0%20%E2%80%93%20felt!
> 
> edit: if there is a problem it's that we may be turning it into too much felt? but if you can turn yarn into an equivalent weight of felt that seems like it would be correct.

The problem is is that in Cata Yarn is to Felt as Thread is to Cotton Patches. One felt disassembles into 50 yarns.  So it would be like cutting up threads and getting cotton patches. Those ratios may be entirely wrong though, because they seem to be missed by the tailoring rework.

Nebnis:
> Yarn literally turns into felt.
> 
> https://support.lionbrand.com/support/solutions/articles/17000035404-felting-techniques#:~:text=Yarns%20made%20of%20wool%20(non,together%20and%20%E2%80%93%20voil%C3%A0%20%E2%80%93%20felt!
> 
> edit: if there is a problem it's that we may be turning it into too much felt? but if you can turn yarn into an equivalent weight of felt that seems like it would be correct.

I think you misunderstood the issue maleclypse, yarn turn into felt patch if you craft it not if you hacking up with your knife with butcher choice.  Anoob is right about giving yarn an unsalvageable tag much like thread can't be cut up with a knife and turn into cotton scraps.

> Yarn literally turns into felt.
> 
> https://support.lionbrand.com/support/solutions/articles/17000035404-felting-techniques#:~:text=Yarns%20made%20of%20wool%20(non,together%20and%20%E2%80%93%20voil%C3%A0%20%E2%80%93%20felt!
> 
> edit: if there is a problem it's that we may be turning it into too much felt? but if you can turn yarn into an equivalent weight of felt that seems like it would be correct.

Well i think i know what the problem of that issue was, according to your link, agitating yarn, (i assume with cutting?) turn into felt patches but felt patch should never go back to being yarn again. So what we actually can't do add a unsalvageable tag to felt patch. I think it already has but it isn't working?

Maleclypse:
I think felt shouldn’t disassemble into yarn is the correct solution. Felt shouldn’t disassemble into anything.

anoobindisguise:
Nebris has a separate PR they submitted to my fork; I asked them to just submit it here.

PR #61377, opened 2022-10-01
Insect mutation tweaks

anoobindisguise:
#### Summary
Balance "tweaks to insect mutations"

#### Purpose of change
Insects often have shells (beetles for example) so it felt fitting to give them shells, then that led to a bunch of other changes.

#### Describe the solution
- Shell was added to Insect.
- There is a new Shell variant that is post thresh Insect, Sleek Shell. It offers 9mm of sclerotin (compared to 10mm of light chitin for the standard shell) and is only OUTER rather than BELTED. 
- 
Doing this exposed an underlying issue with Insect that it had way more than 90 thresh breach power. So I had to lower it a bit.

- Proboscis was moved back to being post-threshold.
- Bulging Reservoir was moved to post-threshold.
- At this point Insect still had 91 breach power so I got rid of Deformed and left them with just Ugly. This seemed like the least consequential way to get them within acceptable bounds.

Different shifting of pre threshold stuff.

#### Testing
Confirmed no errors and could successfully stack shell on top of a chitin carapace through mutation.

#### Additional context
image

### DIFF
--- a/data/json/items/armor/integrated.json
+++ b/data/json/items/armor/integrated.json
@@ -788,6 +788,31 @@
     ]
   },
   {
+    "id": "integrated_shell4",
+    "type": "ARMOR",
+    "category": "armor",
+    "name": { "str_sp": "sleek shell" },
+    "description": "You have a sturdy shell that fits your exoskeleton snugly.",
+    "weight": "8650 g",
+    "volume": "9 L",
+    "price": 0,
+    "price_postapoc": 0,
+    "material": [ "sclerotin" ],
+    "symbol": "x",
+    "color": "magenta",
+    "warmth": 5,
+    "environmental_protection": 10,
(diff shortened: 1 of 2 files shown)


Maleclypse:
This is one of those moments where I will repeat my requests for someone anyone to normalize breach tests against whatever the total breach power of any given tree is.  Or break insect up into multiple trees.

Venera3:
100 coverage is too high for it if it wants to be a stand-in for elytra. I'd also make it incompatible with Insect/Butterfly wings on principle, unless you want to do a whole "open shell->unfold wings->jump down cliff" sequence to it.

anoobindisguise:
All these seem like good reasons to hold off doing this for now. There are numerous features that should probably be in place first, such as branching thresholds or the aforementioned breach test stuff.

PR #61379, opened 2022-10-01
[Aftershock] Flickering Lights EOC AFS

Maleclypse:
#### Summary
Mods "Flickering Lights EOC Aftershock"

#### Purpose of change
Adds a flickering lights eoc to go along with the crashing ship.

#### Describe the solution
Adds the terrains and eoc needed to create a flickering lights effect.

None

#### Testing
TBD

### DIFF
--- a/data/mods/Aftershock/effects_on_condition.json
+++ b/data/mods/Aftershock/effects_on_condition.json
@@ -65,5 +65,103 @@
       },
       { "u_teleport": { "global_val": "new_map" }, "force": true }
     ]
+  },
+  {
+    "type": "effect_on_condition",
+    "id": "afs_flickering_lights",
+    "recurrence": [ "15 seconds", "72 seconds" ],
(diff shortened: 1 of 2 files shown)


Maleclypse:
I'm half asleep right now and I'm missing several things I want this to do and I'm not positive what's the best way to make it happen.   

I would like for it to randomize which light terrains switch instead of flipping all of them in radius each time.

Maleclypse:
@Ramza13 I can't decide if this is the best way to create this type of EOC or if I should do a bunch of nested eoc randomizers?  Do you have any suggestions?

John-Candlebury:
I'm uncertain is this integrated with the crashing ship already?

Maleclypse:
> I'm uncertain is this integrated with the crashing ship already?

I'll double check but I don't think so.  I've kind of wondered if the crashing ship is already too much going on to add flashing on-off lights.

John-Candlebury:
Thematically I think it'd help convey whats going on. And the general effect can probably be reused in several parts

Maleclypse:
Need to test this in the crashing spaceship

PR #61393, opened 2022-10-01
No case hardening for non-CH'd sheet metal arm guards

TheShadowFerret:
#### Summary
None

#### Purpose of change

Non-case hardened sheet metal arm guards shouldn't utilize the case hardening proficiency.

#### Describe the solution

Remove it from the recipe.

#### Testing

Doesn't show usage of prof ingame.


Maleclypse (approved):

LyleSY:
Mapgen build error isn't related

dseguin:
Temporarily closing-&-reopening to point the tests at the current master branch

PR #61397, opened 2022-10-01
Garlic Powder

Daved27hundred:
#### Summary
Content "Adds Garlic Powder"

#### Purpose of change

Garlic Powder is a common seasoning and I'm pretty certain every household has some. In fact I'm pretty sure that it's used more often then whole garlic at least where I live. More options for seasoning is nice, can't put it any other way.

#### Describe the solution

created garlic powder item, added the recipe, added it as a possible ingredient to several recipes, some that already used garlic cloves.

making garlic appear in pantries more often.

### DIFF
--- a/data/json/recipes/recipe_food.json
+++ b/data/json/recipes/recipe_food.json
@@ -4128,9 +4128,9 @@
     "autolearn": true,
     "book_learn": [ [ "cookbook_italian", 1 ] ],
     "batch_time_factors": [ 83, 3 ],
-    "charges": 200,
+    "charges": 50,
     "tools": [ [ [ "surface_heat", 3, "LIST" ] ], [ [ "rock_quern", -1 ], [ "clay_quern", -1 ] ] ],
-    "components": [ [ [ "garlic", 1 ], [ "garlic_clove", 6 ] ] ]
+    "components": [ [ [ "garlic_clove", 1 ] ] ]
   },
   {
     "type": "recipe",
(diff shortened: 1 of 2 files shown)


onura46:
I recommend adding this to the `SUS_spice_collection` item group so that it dynamically populates in kitchens. There are item groups for restaurant kitchens which might be of interest, as well: 

https://github.com/CleverRaven/Cataclysm-DDA/blob/edb21b9806a9bbe53d50c4d459273b0a93daa548/data/json/itemgroups/SUS/domestic.json#L633

https://github.com/CleverRaven/Cataclysm-DDA/blob/edb21b9806a9bbe53d50c4d459273b0a93daa548/data/json/itemgroups/food_service.json#L1047

data/json/recipes/recipe_food.json
```diff
@@ -4115,6 +4116,38 @@
...
   },
+  {
+    "type": "recipe",
+    "activity_level": "LIGHT_EXERCISE",
+    "result": "garlic_powder",
+    "category": "CC_FOOD",
+    "subcategory": "CSC_FOOD_OTHER",
+    "skill_used": "cooking",
+    "difficulty": 2,
+    "time": "45 m",
```
phoeo, line 4127: Doesn't it take 3-8 hours to dry?

Daved27hundred: does it? I could alter the recipe then. I'm not sure people will be happy that it takes so long so I'll make it between 1-2 hours.

Daved27hundred: after looking it up, it seems that is for a massive amount of garlic. the 45 minutes is for a single clove. I'm sure if you're dehydrating several bulbs worth of garlic it'd take 3-8 hours ingame.

phoeo: It just seems that in every recipe I saw there is a long time for passive dehydration. For example, here it says "4-6 hours in oven, or 2-3 days in open air"

https://www.alphafoodie.com/how-to-make-garlic-powder/

Daved27hundred: huh. I assumed the player was just speed-dehydrating the garlic if that even exists. Similar to how you dehydrate fish or meat within 20 minutes ingame.

Daved27hundred: I'm, going to resolve this conversation if that's okay with you, it seems like this isn't just the powder but dehydrated foods as a whole.

Issue #61407, opened 2022-10-02
can't load the save file


terminaldweller:
### Describe the bug

I cannot load my save anymore. I get this error message:
```
 An error has occurred!  Written below is the error report:
 -----------------------------------------------------------
 DEBUG    : Failed to read from "./save/somewhere2/#VHJleSAnUm9ndWUnIFNjaG5laWRlcg==.sav": Json index out of range of uint16_t

 FUNCTION : bool read_from_file(const ghc::filesystem::path&, const std::function<void(std::basic_istream<char>&)>&)
 FILE     : src/cata_utility.cpp
 LINE     : 433
 VERSION  : 5a82316
 -----------------------------------------------------------
 Press space bar to continue the game.
 Press I (or i) to also ignore this particular message in the future.
```
I either get this message or the game just terminates to the terminal without any messages.

### Steps to reproduce

I realize this is unhelpful but the steps involved are:
1. try to load this specific save
2. it wont load either with a crash to the terminal or the error message provided above

### Expected behavior

The game should load normally

### Versions and configuration

- OS: linux
- Game Version: 5a82316, experimental x64-2022-09-23-0555
- Graphics version: terminal
- Ingame langauge: english
- Mods loaded: 
```json
[
  "dda",
  "no_npc_food",
  "package_bionic_professions",
  "aftershock",
  "blazeindustries",
  "DinoMod",
  "megafauna",
  "deadly_bites",
  "extra_mut_scens",
  "cbm_slots",
  "speedydex",
  "stats_through_kills",
  "StatsThroughSkills",
  "Tamable_Wildlife"
]
```

akrieger:
Sonuvagun.

akrieger:
I would still like to have the save game in hand so I can understand what array in it is getting so big but that PR should fix this.

terminaldweller:
I can't upload the save in its entirety since my internet connection can't really handle it but if you want specific files I could put them up somewhere.

Zireael07:
@terminaldweller: Can't you zip or tar the save?

akrieger:
Dropbox or Google drive possibly have better upload performance for slower connections.

terminaldweller:
The directory is  5.1GB.  Even when compressed this is still beyond anything my current internet connection can handle.

akrieger:
The saves/<your world name>/ directory is over 5GB large? Damn. Even when compressed to zip? Damn!

terminaldweller:
[...]

PR #61416, opened 2022-10-02
Fix savegame loading when savegame arrays can have more than uint16_t…

akrieger:
… elements

#### Summary
None

#### Purpose of change
Presumptive fix for #61407. Based on the exception message, savegame json evidently can create arrays with more than 64k elements in them.

#### Describe the solution
Make the json index type 32 bits giving us space for more than 4 billion elements in json arrays (or 4 billion keys in objects).

- Preventing saves making such big arrays. Without a savegame to debug I can't determine if this is a bug or not.

#### Testing
CI. Loaded existing saves fine.

### DIFF
--- a/src/flexbuffer_json.h
+++ b/src/flexbuffer_json.h
@@ -27,7 +27,7 @@ struct JsonPath {
         // the json value at that location. Arrays as native indices, objects as an index
         // into the object's corresponding key vector.
 
-        using index_type = uint16_t;
+        using index_type = uint32_t;
 
         JsonPath() = default;
         JsonPath( const JsonPath & ) = default;
@@ -40,12 +40,12 @@ struct JsonPath {
             if( idx < std::numeric_limits<index_type>::max() ) {
                 path_.push_back( static_cast<index_type>( idx ) );
             } else {
-                throw std::runtime_error( "Json index out of range of uint16_t" );
+                throw std::runtime_error( "Json index out of range of uint32_t" );
             }
             return *this;
         }
 
-        uint16_t size() const {
+        uint8_t size() const {
             return path_.size();
         }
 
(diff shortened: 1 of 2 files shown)


src/flexbuffer_json.h
```diff
@@ -40,27 +40,27 @@ struct JsonPath {
...
                 path_.push_back( static_cast<index_type>( idx ) );
             } else {
-                throw std::runtime_error( "Json index out of range of uint16_t" );
+                throw std::runtime_error( "Json index out of range of uint32_t" );
             }
             return *this;
         }
 
-        uint16_t size() const {
+        uint8_t size() const {
```
anothersimulacrum, line 48: This doesn't seem quite right, or am I missing something?

akrieger: The size this is referring to is the "depth" of the Json object from the root. Eg. the deepest thing in `{"x": [1, {"y": 2, "z": [3]}]}` is at depth 4 with a JsonPath that would look like  `[0, 1, 1, 0]`. I tried to make it `decltype(path_.size())` but trailing return type syntax has weird restrictions on what members are in scope or not and I didn't feel like typing out the whole type. uint8_t is the default size and I'm hoping compiler flags will flag if that ever changes because the underlying size type in `path_` changes.

akrieger:
The reason this restriction existed in the first place was to limit how much space JsonPath used which in turn affected how much space every derived Json object used except for JsonValue. With this change these objects grow from 48 bytes to 64 bytes, which is still a reasonable and round number. Previously I attempted to use uint8_t but core cdda json data has some arrays over 1k large so that wouldn't have worked, but using uint8_t could've kept sizes down to 32 bytes for those types.

akrieger:
Confirmed I can load the savegame from the issue.

Issue #59513, opened 2022-07-22
Crafting generates prompt spam when identical tools are nearby


bouchacha:
### Is your feature request related to a problem? Please describe.

(This issue is similar in spirit to [#27144](https://github.com/CleverRaven/Cataclysm-DDA/pull/27144), back when shooting a bow and arrow prompted you to specify which arrow EACH time)

This issue comes up when you have identical tools nearby and you craft a recipe that requires either of the tools. There are many different scenarios but I'll focus on two examples to illustrate.

The [onboard chemistry lab](https://nornagon.github.io/cdda-guide/#/vehicle_part/chemlab) provides both a 'chemistry set' and a 'hotplate' in one place. As far as I know, any recipe that requires one allows the other, with the exact same number of charges used. Nevertheless, if you craft a recipe near an onboard chemistry set that uses _either_ a hotplate or a chemistry set, the game will still prompt you to choose which one to use even though the choice literally makes no difference. If a choice results in exactly the same outcome, it shouldn't be presented as a choice.

The other scenario does not have an obvious solution. If you have a welding rig as well as an arc welder, crafting recipes prompt you to ask which of the tools you want to use. This makes sense in a lot of scenarios, but still can get annoying. Other examples include 'smoking rack' vs 'hotplate', 'kerosene cooker' vs 'hotplate', etc. For me, it encourages me to intentionally avoid using some items (e.g. kerosene cooker) or intentionally keeping batteries on items unloaded just to avoid the prompt spam.

### Solution you would like.

The solution to the first scenario should be obvious: if a recipe that can use multiple tools consume the same number of charges, and if the tools draw charges from the same source, then don't ask the player which one they should use (because it doesn't matter).

A solution to the second scenario is more complicated, and might ultimately not be desirable. You would need to contemplate what the most common scenarios are and then build up some assumptions. For example, if you have a portable soldering iron but you're in your vehicle with a welding rig, you _probably_ don't want to use up the battery charge on your soldering iron. The rule therefore might be to default to whichever tool has the _highest_ number of charges. Another solution would be to prioritize less scarce energy sources. If you're crafting a recipe that can use either a 'hotplate' or a 'charcoal smoker', you _probably_ don't want to use the smoker since coal is far more precious a resource than electricity.

I admit my proposed solutions to the second scenario are not perfect, but perhaps that can be addressed by a toggleable options menu.

PatrikLundell:
[...]

PR #61424, opened 2022-10-02
Remove pseudo hotplates on parts with chemistry set

irwiss:
#### Summary
None

#### Purpose of change

Doesn't completely res*olve #59513 but reduces the annoyance with vehicle parts that provide hotplate and chemistry set pseudo tools triggering a useless menu choice.

#### Describe the solution

Remove hotplate from such parts, make pseudo chemistry set bound to `h` instead to provide 'heat item'

This removes the pointless dialog in screenshot

#### Testing

1. Start world with blaze industries mod, spawn an RV, 'uncooked toaster pastry', 'cooking rig', wrench
2. Replace kitchen part with "onboard chemistry set", test cooking the toaster pastry - no menu should show up
3. Repeat step 3 with a cooking rig and KitchenMaster part

#### Additional context

image

### DIFF
--- a/data/json/vehicleparts/utilities.json
+++ b/data/json/vehicleparts/utilities.json
@@ -137,9 +137,8 @@
     "durability": 80,
     "flags": [ "CARGO", "OBSTACLE", "COVERED" ],
     "pseudo_tools": [
-      { "id": "pseudo_veh_chemlab_chemset" },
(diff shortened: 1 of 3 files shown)


Rjtheperson:
image
This makes it so I can't use the lab/oven hotplate option in some recipes now with the electric oven and chemistry set appliances. @esotericist  said they would take a look at it already but I figured I should bring it up here.

PR #61427, opened 2022-10-03
Better graffitis: wall vs. floor & variants

mlange-42:
#### Summary
Content "Better graffitis: wall vs. floor & variants"

#### Purpose of change

Allow for better rendering of graffitis, particularly with ISO tilesets.

#### Describe the solution

* Allow to distinguish wall vs. floor graffiti (uses rotation)
* Allow for random variants like for terrain etc.

None

#### Testing

Tried in-game. See screenshot and JSON example.

#### Additional context

Usage:

```json
{
  "id": "graffiti",
  "fg": [
    { "weight": 1, "sprite": [ "graffiti_wall_01", "graffiti_floor_01" ] },
    { "weight": 1, "sprite": [ "graffiti_wall_02", "graffiti_floor_02" ] }
  ],
  "bg": ""
}
```

grafik

### DIFF
--- a/doc/TILESET.md
+++ b/doc/TILESET.md
@@ -360,6 +360,29 @@ Each JSON file can have either a single object or an array of one or more object
 ]
 ```
 
+#### Graffitis
+
+For graffitis, rotation can be used to distinguish between wall and floor drawings:
+
+```json
+{
+  "id": "graffiti",
+  "fg": [ "graffiti_wall", "graffiti_floor" ],
+}
+```
+
+Weighted variations are also possible, and can be combined with rotation:
+
+```json
+{
+  "id": "graffiti",
+  "fg": [ 
+    { "weight": 1, "sprite": [ "graffiti_01_wall", "graffiti_01_floor" ] },
+    { "weight": 1, "sprite": [ "graffiti_02_wall", "graffiti_02_floor" ] },
+  ],
+}
+```
+
 ### `tile_info.json`
 ```c++
 [


src/cata_tiles.cpp
```diff
@@ -2345,9 +2345,17 @@ bool cata_tiles::draw_from_id_string( const std::string &id, TILE_CATEGORY categ
         case TILE_CATEGORY::OVERMAP_TERRAIN:
         case TILE_CATEGORY::MAP_EXTRA:
             seed = simple_point_hash( pos );
+            break;
         case TILE_CATEGORY::ITEM:
         case TILE_CATEGORY::TRAP:
         case TILE_CATEGORY::NONE:
+            // graffiti
```
dseguin, line 2352: Should `ITEM` and `TRAP` be moved below `NONE` to avoid the added check? It looks like graffiti only applies to the `NONE` category

Fris0uman:
Can you document this somewhere? Because from the outside there's no way to know about this feature otherwise

mlange-42:
> Can you document this somewhere? Because from the outside there's no way to know about this feature otherwise

Didn't find a place where it fits, but will try to integrate it in `TILESET.md`.

src/cata_tiles.cpp
mlange-42: Done.

mlange-42:
> > Can you document this somewhere? Because from the outside there's no way to know about this feature otherwise
> 
> Didn't find a place where it fits, but will try to integrate it in `TILESET.md`.

Done.

mlange-42:
CI fail seems unrelated:

```
src/map.cpp:1616 [bool map::furn_set(const tripoint &, const furn_id &, const bool, bool)] Setting furniture f_sandbag_half at (10,16,0) where terrain is t_open_air (which is_open_air)
```

PR #59827, opened 2022-08-02
Rework repair menu

Night-Pryanik:
#### Summary
Interface "Rework repair menu"

#### Purpose of change
* Resurrect abandoned #50294 by @WelpThatWorked.
* Closes #50048.

#### Describe the solution
The work was mostly done by @WelpThatWorked. I just replaced all references to obsoleted `player` class with `Character` class and added missing translation stuff.

#### Describe alternatives you've considered
None.

#### Testing
Got sewing kit. Got some damaged clothing and repair components. Activated sewing kit, repaired clothing.

#### Additional context
изображение

PR #61438, opened 2022-10-03
Fix repair menu shows slow

### LINKED ISSUE #61380: Repairing articles of clothing is laggy asf
### Describe the bug

When I try to repair clothes, the game lags/freezes for a good few seconds.

### Steps to reproduce

1. get clothes to repair and a sewing kit
2. Loading.gif
3. List of items loaded after a while.
4. Change page from 1/3 to 2/3.
5. Loading.gif
6. Repair an item.
7. Loading.gif
8. Try to fit/repair/reinforce.
9. Loading.gif

### Expected behavior

No freezes or lags

### Screenshots

_No response_

- OS: Android
    - OS Version: Manufacturer: Lenovo; Model: Lenovo L58041; Release: 8.1.0; Incremental: 3.9.166_200309; 
- Game Version: 5e89406 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
    Teleporter [Teleporter],
    Artifact Jewelries [Artifact Jewelries]
]

### Additional context

_No response_

EIIKaO:
#### Summary
Performance "Fix Repair menu shows slow"

#### Purpose of change
- Fix #61380
  - Stopped comparing against all recipes every time one item's repair probability is calculated, and instead used the recipes held by the item type.
- Now use the difficulty in uncraft recipes.
  -  And if the recipe did not define the difficulty, 5( on items' recipe and obsolete recipes ) or 0 ( on uncraft recipe ) are used for the default. As a result, some items show changes in repair success chances( see details in Testing and Additional context ).

#### Describe the solution
`repair_recipe_difficulty` function is that each time called compares the item to be repaired with all recipes in the recipe dictionary( recipes over 5,500 currently ).

If about 50 items appear in the repair menu, then at least 50*5,500( 275,000 ) comparisons will be done. I have concluded that the large number of times this comparison occurs is the bottleneck of the problem.

So I stopped comparing with all recipes and used the recipes held by the item type. If a recipe is not held or matched, it is matched either with the dictionary of uncraft or obsoletes. If none is found, -1 is returned.

Use a cache for comparison with all recipes ( converted to vector, not map, and kept in recipe_dictionary, but did not improve performance enough as far as I tried ).

#### Testing
In-game testing done. Performance was improved dramatically( in one case, what used to take about 30 seconds became 0.35 seconds ), and the returned difficulty level was as designed.

As far as I could confirm in-game, there were changes in the repair success chance. <del>This is probably all due to the difference in the handling of difficulty 0,</del> which is explained later.
- Decrease cases
`turnout coat` ( current code returned 0 because the obsolete recipes were not defined difficulty )
`US ballistic vest` ( cur code returned 0 because it has obsolete recipe and the other recipe in uncraft recipes with difficulty 2 was ignored )
- Increase cases
`acetylene torch`, `processor board`, `oxygen tank`, `flare`, `e-ink tablet PC` ( current code does not search for uncraft recipes, determining no recipe and returning difficulty -1 )
`some of materials` ( i.e. sheet of kevlar, lycra, synthetic fabric and etc... )

#### Additional context
Difficulty 0 has two main types.

The first one is explicitly defined as 0.
1. This meaning is very clear. That is, the difficulty level is very low( however, there does not seem to be a recipe with 0 defined in vanilla ).

The second one is complicated.

1. The most common cases would be that there is no need to define it, in other words, it is not expected to be created( but we can repair it ).
2. In other cases, by not defining it, they may expect a default value to be assigned.
3. Moreover, there may be cases where they forgot to define it.

In this PR, I decided to use 5 if the difficulty level was not defined. It is expected that case (2.) will be largely affected.

Advice and criticism **welcome**.

### Notes:
*Updated
It seems that there are cases where delays do not occur, evidenced by the video presented by  @Night-Pryanik https://github.com/CleverRaven/Cataclysm-DDA/pull/59827#issuecomment-1203081482.
In my environment there is about 6.5 seconds delay, almost the same as @pjf who provided saved data for testing.
If someone has an environment without delays, I would appreciate it if you could provide me with information ( PC environment, OS, etc. ) or your opinion.

*Updated 2
While reading the latest release code, I found a strange point.
https://github.com/CleverRaven/Cataclysm-DDA/blob/3f63c448f6089a1c4264c24ead234d19d612adf1/src/iuse_actor.cpp#L2822-L2842
In this case, the difficulty would return a maximum of 5 ( 7 after adjustment +2). As a result, a `handheld game system` with a difficulty of 8 in the recipe, for example, will return a minimum of 5 and a maximum of 7, making it easier to repair than expected. I wonder if this is as designed.

### DIFF
--- a/src/iuse_actor.cpp
+++ b/src/iuse_actor.cpp
@@ -2811,38 +2811,36 @@ bool repair_item_actor::handle_components( Character &pl, const item &fix,
     return true;
 }
 
-// Find the difficulty of the recipes that result in id
+// Find the difficulty of the recipe for the item type.
 // If the recipe is not known by the player, +1 to difficulty
 // If player doesn't meet the requirements of the recipe, +1 to difficulty
 // Returns -1 if no recipe is found
-static int find_repair_difficulty( const Character &pl, const itype &it, bool training )
+static std::pair<int, bool> find_repair_difficulty( const Character &pl, const itype &it,
+        bool training )
 {
-    // If the recipe is not found, this will remain unchanged
     int min = -1;
     const int def_diff = 5;
     const itype_id iid = it.get_id();
-    recipe uncraft_recipe = recipe_dictionary::get_uncraft( iid );
+    bool found = false;
+    bool difficulty_defined = false;
 
-    if( it.recipes.empty() && uncraft_recipe ) {
-        int uncraft_difficulty = uncraft_recipe.difficulty;
-        if( uncraft_difficulty == 0 ) {
-            uncraft_difficulty = -1;
-        } else if( !training && !pl.has_recipe_requirements( uncraft_recipe ) ) {
-            uncraft_difficulty++;
-        }
-        min = std::min( uncraft_difficulty, def_diff );
-    } else {
+    if( !it.recipes.empty() ) {
         for( const recipe_id &rid : it.recipes ) {
             const recipe &r = recipe_id( rid ).obj();
             if( !r ) {
                 continue;
             }
             // If this is the first time we found a recipe
-            if( min == -1 ) {
+            if( !found ) {
                 min = def_diff;
+                found = true;
             }
 
             int cur_difficulty = r.difficulty;
+
+            // Recipes with difficulty 0 are not present. Difficulty is considered undefined.
+            difficulty_defined = cur_difficulty != 0 || difficulty_defined;
+
             if( !training && !pl.knows_recipe( &r ) ) {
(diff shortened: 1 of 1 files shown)


KymasTuran:
Hope this is merged soon. I m repairing naked and with just the broken item. (old pc)
Are there plans to change the entire repair system?

Issue #61214, opened 2022-09-23
Trail guides and maps change their name if you travel to a different city


sethsimon:
### Describe the bug

The name (ie what you see in your inventory) of trail guides and other maps is dynamically generated based on your position and the nearest city. Thus, if you drive (or long-range teleport) to another city, your `Corinna trail guide` would now be called a `Lemington trail guide`.

### Steps to reproduce

1. Go to the middle of some city (say, Corinna).
2. Find/debug yourself a trail guide.
3. Enter your inventory with `i` to see that you have a `Corinna trail guide`.
4. Travel to the middle of another city (say, Lemington).
5. Enter your inventory with `i` to see that you have a `Lemington trail guide`.

### Expected behavior

Dynamically generating the map name is fine, but the city/area that's revealed should be stored in the item instead of being based on the player's position. This would also allow behavior such as spawning maps of various nearby cities at bus stations and airports.

### Versions and configuration

Linux, last commit is 25fd53ce63259b4c232782b1bcd15e86e7d7585c

PatrikLundell:
I haven't seen this exact issue personally, as I always look at them immediately. However, I've seen what is most likely the same thing with notes hidden away in wallets, where the note claims to show info about secret facilities, and that gets somewhat silly when I finally get to examine the wallet contents and find notes about facilities under cities that don't have anything underground.

Terrorforge:
They also reveal the map around wherever you happen to be, rather than where you initially picked them up.

This gets particularly silly with the Visitor's Pass for the TCL. I have them on autopickup so I don't have to manually dig through wallets, and this results in me always having a big stack of maps I can use to reveal all the nearby roads and TCLs whenever I want.

ZhilkinSerg:
> They also reveal the map around wherever you happen to be, rather than where you initially picked them up.
> 
> This gets particularly silly with the Visitor's Pass for the TCL. I have them on autopickup so I don't have to manually dig through wallets, and this results in me always having a big stack of maps I can use to reveal all the nearby roads and TCLs whenever I want.

No, it does not work that way. Maps store center of reveal location when generated and map details are revealed in a radius from that center point - https://github.com/CleverRaven/Cataclysm-DDA/pull/27012

ZhilkinSerg:
> debug yourself a trail guide.

Do not do that - debug-spawned items are not originating from a specific city, so they do not have center point added to them.

ZhilkinSerg:
[...]

PR #61439, opened 2022-10-03
Maps no longer change name from travel #61214 [CR]

Harrison-Salzverg:
#### Summary
Bugfixes "Maps no longer change name from travel"

#### Purpose of change
Issue https://github.com/CleverRaven/Cataclysm-DDA/issues/61214 describes a bug in which found maps and trail guides will reveal map elements for the closest city to the player at the time of opening, rather then the location the map was found in. This pull request is an attempt to fix the bug by implementing the fix suggested by ZhilkinSerg https://github.com/CleverRaven/Cataclysm-DDA/issues/61214#issuecomment-1257054422. Bug was recreated in a virtual machine running Ubuntu 20.04 following Steps to Reproduce and instructions from https://github.com/CleverRaven/Cataclysm-DDA/issues/61214#issuecomment-1256946108

#### Describe the solution
The is_map function now returns true if either the item is in the Maps category, OR if the item has a "reveal_map" use action id, rather then purely the former.

Considered using the action name rather then id, but there are possibly cases that can circumvent these checks without being maps.

#### Testing
Game compiled and testing with new build did not seem to recreate the bug. This is admittedly more complex then the bug testing I usually do, so further testing is welcome.

#### Additional context
My first commit, please be gentle : )


.github/workflows/build.yml
```diff
@@ -0,0 +1,47 @@
+name: Build
```
anothersimulacrum, line 1: This looks to be added accidentally.

Harrison-Salzverg: My apologies.  I'm still new to GitHub, and in trying to add support for my static analyzer of choice to my local fork, I pushed it as well.  It can be disregarded.

anothersimulacrum: You need to remove it, or your PR cannot be merged.

Harrison-Salzverg: It has been removed.

src/item.cpp
```diff
@@ -9862,7 +9862,11 @@ std::string item::get_book_skill() const
 
 bool item::is_map() const
 {
-    return get_category_shallow().get_id() == item_category_maps;
+    if((get_category_shallow().get_id() == item_category_maps) || (get_action() == "reveal_map")) {
```
ZhilkinSerg, line 9865:
```suggestion
    if( ( get_category_shallow().get_id() == item_category_maps ) || ( get_action() == "reveal_map" ) ) {
```

Maleclypse:
Hi wanted to check in on why this was closed. It looks like it was waiting on an available merger.

PR #61440, opened 2022-10-03
Regional settings don't silently overwrite overlay

### LINKED ISSUE #60394: region_settings overwrites all previous region_overlays
### Describe the bug

If you load a mod with a region_overlay, and then load a mod with a region_settings, the region_overlay gets completely discarded.

### Steps to reproduce

Edit the "field" field in mods/DinoMod/regional_overlay.json to read `"field": { "chance": 1, "extras": { "mx_mass_grave_dino": 300000 } }`
Create a world where you add DinoMod first, and TropiCataclysm second
Boot up the world and notice there are no dino mass graves.
Create a world with TropiCataclysm first and DimoMod second
Boot up the world and notice there are dino mass graves on most every field tile.

### Expected behavior

The player should not have to manually manage load order in order to make sure the mods they're playing with works. Or if they do need to do so, they should be given some clues as to what order the mods have to be in.

region_overlays should probably be applied after all region_settings have finished being applied. (IMHO the preferred solution)
Alternatively, mods with a region_settings could be forced to load before mods with a region_overlay

### Screenshots

mass_grave

- OS: Windows
    - OS Version: 10.0.19042.1110 (20H2)
- Game Version: 630e438 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    DinoMod [DinoMod],
    TropiCataclysm [tropicata]
]

### Additional context

_No response_

anothersimulacrum:
#### Summary
None

#### Purpose of change
Fixes https://github.com/CleverRaven/Cataclysm-DDA/issues/60394

#### Describe the solution
Track whether regional overlays have been loaded for a region, and when loading a region with that id (or any region, for "all" overlays), if any overlays have been loaded and we have not told the game to overwrite them, produce a debugmsg detailing the overlay id and what it has been loaded from.

This solution seems pretty clear and straightforward.

#### Testing
`tests/cata_test --mods=dda,DinoMod,tropicata '~*' ` produces an error,
`tests/cata_test --mods=dda,tropicata,DinoMod '~*' ` does not.
Load a game with dinomod and then tropicata loaded, receive an error. Exit that game without quitting the binary, and load a game with them loaded in the correct order, receive no error.

#### Additional context
This will likely cause lots of existing saves that have silently had mods loading in the wrong order to produce a message complaining about it.


PatrikLundell:
If this change results in a lot of reports for existing errors, is there a way to "repair" those saves? Getting spammed with "your save is damaged" reports with no way to recover is probably not going to be well received.

Ideally, the error report ought to indicate either how to repair a save or provide a reference to documentation that describes how to do it (in a way that most average players would be able to follow).

anothersimulacrum:
The way to do so it fix the loading order. I think you can do this from the main menu.

Light-Wave:
If all else fails, you can go to save/<world name>/mods.json and edit it to change the load order around.

anothersimulacrum:
TODO: Also add a feature suggesting mod ordering so that I can actually pass the all mods test :)

src/regional_settings.cpp
```diff
@@ -448,12 +451,19 @@ static void load_region_terrain_and_furniture_settings( const JsonObject &jo,
...
     regional_settings new_region;
     if( !jo.read( "id", new_region.id ) ) {
         jo.throw_error( "No 'id' field." );
     }
+    bool overwrite_overlays = false;
+    optional( jo, false, "overwrite_overlays", overwrite_overlays, false );
+    if( ( regions_applied_overlay.count( new_region.id ) != 0 || loaded_all_region_overlay ) &&
+        !overwrite_overlays ) {
+        debugmsg( "Region settings for region %s (from %s) is loaded after an overlay for this region has been applied!"
+                  " Adjust world mod load order to fix.", new_region.id, src );
```
ZhilkinSerg, line 465:
```suggestion
                  "  Adjust world mod load order to fix.", new_region.id, src );
```

LyleSY (approved):
I don't understand the code here, but the logic makes sense to me

anothersimulacrum:
Working on the UI. Not all the way there, and I ~~want~~need tests, but here's where we are.
Mods automatically update when selected to follow loading hints

Maleclypse:
Hi anothersimulacrum, it looks like some of the tests around region overlays fail on this Pr.

PR #61452, opened 2022-10-04
#61451 - rebalance methanol from planks

### LINKED ISSUE #61451: Methanol from wood planks recipe needs rebalancing
### Describe the bug

Currently it takes 12 planks (52.8l / 31.2kg), or 40 splinters (10l / 6kg) to generate 0.25L of methanol.  This is 0.8% and 4.2% conversion rate by weight.  This seems wildly off compared to some youtube videos I've seen and https://www.researchgate.net/publication/334738889_PRODUCTION_OF_LIQUID_FUEL_FROM_WOOD which state that 1kg of wood could make 0.268kg of methanol.  This page https://www.fs.usda.gov/treesearch/pubs/9818 says that "The yield of methanol from wood is about 38 percent." 100 gallons from 1000 ODT.

### Steps to reproduce

Craft methanol from wood planks.

### Expected behavior

The methanol from wood planks conversion ratio should be much higher than 4% - maybe 20%?

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.22000.1042 (21H2)
- Game Version: 0.F-11751-g23fc96560c-dirty [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Slots [cbm_slots]
]

### Additional context

_No response_

jim-huynh:
#### Summary
Bugfixes "Methanol from wood planks recipe needs rebalancing"

#### Purpose of change
Fixes #61451 - Rebalance the methanol recipe so that it is more approximately correct.
Also fixes trying to create a batch of 50 from disappearing due to excessive volume.

#### Describe the solution
The current recipe uses 12 planks which is 52.8L or 31.2kg to make 0.25L/0.25kg of methanol.  I've changed it to 2 planks which is 8.8L or 5.2kg to make 1L/1kg of methanol.  This is roughly 20% efficiency, which is lower than some of the other listed efficiencies reported by industrial processes.
https://www.researchgate.net/publication/334738889_PRODUCTION_OF_LIQUID_FUEL_FROM_WOOD which state that 1kg of wood could make 0.268kg of methanol. This page https://www.fs.usda.gov/treesearch/pubs/9818 says that "The yield of methanol from wood is about 38 percent." 100 gallons from 1000 ODT.

Reduce materials required and leave recipe multiplier alone.

#### Testing
Updated json and loaded crafted methanol from planks in game.

### DIFF
--- a/data/json/recipes/recipe_medsandchemicals.json
+++ b/data/json/recipes/recipe_medsandchemicals.json
@@ -1212,13 +1212,12 @@
     "subcategory": "CSC_CHEM_CHEMICALS",
     "skill_used": "chemistry",
     "difficulty": 4,
-    "result_mult": 4,
     "time": "1 h",
     "batch_time_factors": [ 80, 4 ],
     "autolearn": true,
     "qualities": [ { "id": "BOIL", "level": 2 }, { "id": "DISTILL", "level": 1 } ],
     "tools": [ [ [ "surface_heat", 25, "LIST" ] ] ],
-    "components": [ [ [ "splinter", 40 ], [ "2x4", 2 ] ], [ [ "water", 3 ], [ "water_clean", 3 ] ] ]
+    "components": [ [ [ "splinter", 55 ] ], [ [ "water", 50 ], [ "water_clean", 50 ] ] ]
   },
   {
     "type": "recipe",
--- a/data/json/recipes/recipe_medsandchemicals.json
+++ b/data/json/recipes/recipe_medsandchemicals.json
@@ -1212,12 +1212,13 @@
     "subcategory": "CSC_CHEM_CHEMICALS",
     "skill_used": "chemistry",
     "difficulty": 4,
+    "result_mult": 4,
     "time": "1 h",
     "batch_time_factors": [ 80, 4 ],
     "autolearn": true,
     "qualities": [ { "id": "BOIL", "level": 2 }, { "id": "DISTILL", "level": 1 } ],
     "tools": [ [ [ "surface_heat", 25, "LIST" ] ] ],
-    "components": [ [ [ "splinter", 40 ], [ "2x4", 12 ] ], [ [ "water", 3 ], [ "water_clean", 3 ] ] ]
+    "components": [ [ [ "splinter", 40 ], [ "2x4", 2 ] ], [ [ "water", 3 ], [ "water_clean", 3 ] ] ]
   },
   {
     "type": "recipe",


Maleclypse:
The method that article is referencing uses a catalytic reactor process similar to #58234.  The simple distillation method https://sciencing.com/make-wood-alcohol-through-distillation-7762993.html will have a much lower rate, probably closer to the industrial rates of the 1910's which seems to be just in the low single digits of methanol production.  http://journeytoforever.org/biofuel_library/wood_alcohol.html

data/json/recipes/recipe_medsandchemicals.json
```diff
@@ -1212,12 +1212,13 @@
     "subcategory": "CSC_CHEM_CHEMICALS",
     "skill_used": "chemistry",
     "difficulty": 4,
+    "result_mult": 4,
```
Maleclypse, line 1215:
```suggestion
```

data/json/recipes/recipe_medsandchemicals.json
```diff
@@ -1212,12 +1212,13 @@
...
     "time": "1 h",
     "batch_time_factors": [ 80, 4 ],
     "autolearn": true,
     "qualities": [ { "id": "BOIL", "level": 2 }, { "id": "DISTILL", "level": 1 } ],
     "tools": [ [ [ "surface_heat", 25, "LIST" ] ] ],
-    "components": [ [ [ "splinter", 40 ], [ "2x4", 12 ] ], [ [ "water", 3 ], [ "water_clean", 3 ] ] ]
+    "components": [ [ [ "splinter", 40 ], [ "2x4", 2 ] ], [ [ "water", 3 ], [ "water_clean", 3 ] ] ]
```
Maleclypse, line 1221:
```suggestion
    "components": [ [ [ "splinter", 55 ] ], [ [ "water", 50 ], [ "water_clean", 50 ] ] ]
```

Maleclypse (changes requested):
So I went ahead and tried to match the math to the resources on the more low tech versions that this recipe represents.   The time and heat are probably low but I haven't really found any references on the time it takes to do this in any of my searches.   If you want to do an industrial scale one I would suggest looking at the recipe for kerosene to get an idea of the types of equipment and the like needed.

data/json/recipes/recipe_medsandchemicals.json
Maleclypse: I also removed all the planks as you can't use this method with wood that hasn't been broken down into pellet sizes.  And one hour doesn't get the wood broken down and processed into methanol from planks.

Maleclypse (approved):

jim-huynh:
> So I went ahead and tried to match the math to the resources on the more low tech versions that this recipe represents. The time and heat are probably low but I haven't really found any references on the time it takes to do this in any of my searches. If you want to do an industrial scale one I would suggest looking at the recipe for kerosene to get an idea of the types of equipment and the like needed.

Oh, there is something to point out regarding this change you did.  When you create a batch of 50 of this, would the resulting in progress item be >1000L causing it to disappear on placement on the workbench/ground?

Maleclypse:
> > So I went ahead and tried to match the math to the resources on the more low tech versions that this recipe represents. The time and heat are probably low but I haven't really found any references on the time it takes to do this in any of my searches. If you want to do an industrial scale one I would suggest looking at the recipe for kerosene to get an idea of the types of equipment and the like needed.
> 
> Oh, there is something to point out regarding this change you did. When you create a batch of 50 of this, would the resulting in progress item be >1000L causing it to disappear on placement on the workbench/ground?



> > So I went ahead and tried to match the math to the resources on the more low tech versions that this recipe represents. The time and heat are probably low but I haven't really found any references on the time it takes to do this in any of my searches. If you want to do an industrial scale one I would suggest looking at the recipe for kerosene to get an idea of the types of equipment and the like needed.
> 
> Oh, there is something to point out regarding this change you did. When you create a batch of 50 of this, would the resulting in progress item be >1000L causing it to disappear on placement on the workbench/ground?

It might go over but i don't actually think this recipe should get any batch savings.  You'd really have to design specialty equipment to achieve any time savings from reading over those articles.

jim-huynh:
> > > So I went ahead and tried to match the math to the resources on the more low tech versions that this recipe represents. The time and heat are probably low but I haven't really found any references on the time it takes to do this in any of my searches. If you want to do an industrial scale one I would suggest looking at the recipe for kerosene to get an idea of the types of equipment and the like needed.
> > 
> > 
> > Oh, there is something to point out regarding this change you did. When you create a batch of 50 of this, would the resulting in progress item be >1000L causing it to disappear on placement on the workbench/ground?
> 
> > > So I went ahead and tried to match the math to the resources on the more low tech versions that this recipe represents. The time and heat are probably low but I haven't really found any references on the time it takes to do this in any of my searches. If you want to do an industrial scale one I would suggest looking at the recipe for kerosene to get an idea of the types of equipment and the like needed.
> > 
> > 
> > Oh, there is something to point out regarding this change you did. When you create a batch of 50 of this, would the resulting in progress item be >1000L causing it to disappear on placement on the workbench/ground?
> 
> It might go over but i don't actually think this recipe should get any batch savings. You'd really have to design specialty equipment to achieve any time savings from reading over those articles.

This merge should probably just get cancelled then since one of the reasons for trying to merge it in was because the batch of 50 kept on disappearing on me when I attempted to craft it.

PR #61458, opened 2022-10-04
Graded weapons descriptions

Drew4484:
#### Summary
Content "Improve the descriptions of graded weapons."

#### Purpose of change
Unlike the graded armors, the difference in quality between the versions was not simply tougher steel, but also an improvement over the lower grades, using better grips and dealing more damage to represent better balance and a sharper edge. The load is to represent the player's increasing skills, hopefully the new descriptions aid in this.

#### Describe the solution
Update the descriptive text for all graded steel weapons.

None.

#### Testing
Nothing changing but descriptive text.

#### Additional context
None.


nopenoperson:
the text about  razor sharpness makes sense for a knife that you would want to use for shaving. for a typical sword task like cutting through a zombie you would want a thicker edge that would be much less sharp so it would actually be able to do the task it was intended for without easily breaking.

Drew4484:
> the text about  razor sharpness makes sense for a knife that you would want to use for shaving. for a typical sword task like cutting through a zombie you would want a thicker edge that would be much less sharp so it would actually be able to do the task it was intended for without easily breaking. 

It's an figure of speech but I'll gladly take any alternative descriptions or changes you come up with.

nopenoperson:
> It's an figure of speech but I'll gladly take any alternative descriptions or changes you come up with.
im probably being too picky but you could try "The balance feels perfect and the cutting edge seems exceptionally well honed"

PR #61473, opened 2022-10-04
MilComp repairs

Drew4484:
#### Summary
Content "Makes MilComp item unrepairable"

#### Purpose of change
An oversight with making military composite plating unrepairable meant that it could be repaired by uninstalling it from the vehicle, repairing the item, and then reinstalling it on the vehicle. This A) should not work and B) is extremely tedious for the player taking advantage of it.

#### Describe the solution
Adds "NO_REPAIR" to the military composite item.

None.

#### Testing
Spawned item, confirmed it shows as not repairable.

#### Additional context
I understand the desire to repair military composite armor and the frustration it creates to not be able to do so. Realistically, the repair of such plating is to completely replace the piece of armor if it is damaged beyond a certain point, the piece is then recycled. Simple welds cannot restore the ceramic (which is what makes this **composite**) armor. It may be possible to create new ceramic sheets or at least a ceramic coating on metal armor, but those are extremely advanced processes and generally do not scale down to the creation of one single plate.

### DIFF
--- a/data/json/items/vehicle/plating.json
+++ b/data/json/items/vehicle/plating.json
@@ -135,6 +135,7 @@
     "category": "veh_parts",
     "price": 16000,
     "price_postapoc": 750,
+    "flags": [ "NO_REPAIR" ],
     "qualities": [ [ "COOK", 1 ] ]
   },
   {


Maleclypse (approved):

anothersimulacrum (approved):
Any reason this is still in draft?

Drew4484:
> Any reason this is still in draft?

Wrote it up on my phone, I'll test it today.

PR #61474, opened 2022-10-04
Fix ObjectCreator build on Windows

akrieger:
#### Summary
None

#### Purpose of change
#59906 broke the ObjectCreator build on Windows, but it's not entirely it's fault because the check didn't run on that PR. Fix both issues.

#### Describe the solution
- Remove the object_creator exclusions from the github workflow so it should run on this PR.
- Fix the include to use the same folder relative includes the other Qt includes use.

#### Testing
Build ObjectCreator on Windows.

### DIFF
--- a/object_creator/flowlayout.cpp
+++ b/object_creator/flowlayout.cpp
@@ -49,7 +49,7 @@
 ****************************************************************************/
 
 #include "flowlayout.h"
-#include <QtWidgets>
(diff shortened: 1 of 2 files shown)


snipercup:
> @snipercup can you test that object_creator still works for you with these changes?

I can build it on windows and linux and they both work. Thank you for fixing it! I will put the include correctly next time.

PR #61476, opened 2022-10-04
No sawing off (the most demanding) recoil operated guns' barrels

tenmillimaster:
#### Summary
Bugfixes "Removes Sawable Barrel Volume from M2HB, M107"

#### Purpose of change

The M107 and M2HB are both recoil operated .50BMG guns, in which the barrel reciprocates as part of the action to delay the opening of the breech until pressures drop to safe levels. Having less mass as part of the action means it opens earlier, and is harder on the system. Without knowing how much can be safely taken off, removal is the best option.

#### Describe the solution
Deletes the defined barrel volumes for the M2 and M107.

Implementing a NO_SAW flag, but the game already has a "you can't saw that!" condition.

#### Testing

image

#### Additional context

None

### DIFF
--- a/data/json/items/gun/50.json
+++ b/data/json/items/gun/50.json
@@ -91,10 +91,10 @@
       "volume": "-1500 ml",
       "price": -1210000,
       "ranged_damage": { "damage_type": "bullet", "amount": -4 },
-      "dispersion": 60,
-      "barrel_volume": "-250 ml"
+      "dispersion": 60
     },
     "blackpowder_tolerance": 24,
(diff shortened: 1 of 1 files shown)


TheShadowFerret:
>**Describe alternatives you've considered**
Implementing a NO_SAW flag, but the game already has a "you can't saw that!" condition.

I was hoping for "eye and gun damage + blown-out casing jams it for being dumb" :3

tenmillimaster:
I think Kevin frowns on that kinda gotcha

PR #61482, opened 2022-10-05
Spears, spears, and more spears! A spear overhaul.

Aerin-of-the-Toast:
#### Summary
Balance "Overhaul spears"

#### Purpose of change

Spears had a lot of issues. A lack of DPS in the lategame, wildly varying volumes causing spears to have wildly varying attack speeds, lengths that were too short, and no way to differentiate spears with proper shafts vs just a long stick. This aims to address that.

#### Describe the solution

This got LONG, so for a full document explaining my thoughts on each step, please refer to this document: https://docs.google.com/document/d/18Pql4CyWintBQQOjhxRN-aVtaI4VSXVcFYBVRXsSS-w/edit?usp=sharing

Just leaving spears as they were.

#### Testing

Checked all the spears and crafting recipes in-game. It's always possible I missed a spot, but everything seems to be in order and the recipes all accurate.

#### Additional context

Overall a buff for lategame spears. Early spears got a slight nerf, but now there's a better progression, you don't have to jump straight from a crude knife spear to a steel spear anymore and its easier to get a "durable" spear that's on par with the old steel spear.

### DIFF
--- a/data/json/items/melee/spears_and_polearms.json
+++ b/data/json/items/melee/spears_and_polearms.json
@@ -498,7 +498,7 @@
     "bashing": 8,
     "cutting": 43,
     "to_hit": { "grip": "weapon", "length": "long", "surface": "point", "balance": "neutral" },
-    "flags": [ "DURABLE_MELEE", "SPEAR", "REACH_ATTACK", "NONCONDUCTIVE", "SHEATH_SPEAR" ],
+    "flags": [ "DURABLE_MELEE", "SPEAR", "REACH_ATTACK", "NONCONDUCTIVE", "SHEATH_SPEAR", "POLEARM" ],
     "weapon_category": [ "POLEARMS", "SPEARS" ],
     "price": 80000,
     "qualities": [ [ "COOK", 1 ] ]
@@ -810,7 +810,7 @@
     "color": "yellow",
     "techniques": [ "WBLOCK_1", "IMPALE" ],
     "qualities": [ [ "COOK", 1 ] ],
-    "flags": [ "SPEAR", "REACH_ATTACK", "DURABLE_MELEE", "SHEATH_SPEAR", "NONCONDUCTIVE" ],
+    "flags": [ "SPEAR", "REACH_ATTACK", "DURABLE_MELEE", "SHEATH_SPEAR", "NONCONDUCTIVE", "POLEARM" ],
     "weapon_category": [ "POLEARMS", "SPEARS" ]
   },
   {


SirPendrak:
Copper spear being durable seems strange. Bronze or brass, sure, but not copper

AirRice:
> Copper spear being durable seems strange. Bronze or brass, sure, but not copper

Agreed, although using a copper spear would likely more be hampered by the tip becoming blunted extremely quickly due to copper's high malleability and softness, not necessarily the weapon breaking completely as would be expected of a flimsy spear  built with tape. As there is no current implementation of blunting or sharpness of melee weapons, the best way to represent this would probably be to make it not durable, although it's probably not what players would expect

Dan1ss1mo:
Finally, some love for poking sticks!)

Jarewill:
Since the length of Dory and Qiang were buffed to 250cm, maybe they can also get the "REACH3" flag the pike has?
That would increase the reach of their attacks to 3 tiles instead of 2.
They could also get the "POLEARM" flag from the pike as well, which would make them weaker up close.

Anyways I like what I'm seeing.
Making the bronze spear a shorter version of Dory is a good idea.

Aerin-of-the-Toast:
> Since the length of Dory and Qiang were buffed to 250cm, maybe they can also get the "REACH3" flag the pike has? That would increase the reach of their attacks to 3 tiles instead of 2. They could also get the "POLEARM" flag from the pike as well, which would make them weaker up close.
> 
> Anyways I like what I'm seeing. Making the bronze spear a shorter version of Dory is a good idea.

You're not holding the larger spears by the butt, so even though they're 250cm long, you're not getting nearly that much reach. I did want to give them the POLEARM flag though, forgot about that.

Jerald-of-Libya:
Cheers. Cool changes. Love the stats, but some of the crafting times seem a bit derped.

The simplest t1 knife spear for example, which is essentially just tying/taping a knife to a stick has been changed from 15m to 2h.. making it equal crafting time to the t2 which has bolting/nailing/drilling involved. Meanwhile your t3 knife spear is at a chill 30m while seemingly using the same bolting/nailing/drilling.

Sure, the shaft for the t3 takes several hours by itself, but nailing/bolting/drilling process should ideally be around equal like with the previous version

Something like:
t1 -> 15min
t2 -> 45m
upgrade t1 to t2 -> 30m
t3 -> 45m

KHeket:
And there is problem now. Steel spear was always autolearned at level 4 fabrication. But now to craft it, you need to craft spear shaft, which are learned at level 5 fab. There is recipe of spear shaft at level 2 fab, but it need long pole, which is also need level 5 fabrication, so there in no way to obtain steel spear until level 5 fabrication

Jerald-of-Libya:
> And there is problem now. Steel spear was always autolearned at level 4 fabrication. But now to craft it, you need to craft spear shaft, which are learned at level 5 fab. There is recipe of spear shaft at level 2 fab, but it need long pole, which is also need level 5 fabrication, so there in no way to obtain steel spear until level 5 fabrication

True that. Also, the bo staff is craftable at fab 2, and that involves arguably more rigorous oiling and sanding than a spear shaft, so no reason for it to be fab 5.

Always a lot of stuff to take into account when adding craftables sadly. 😭

KHeket:
> > And there is problem now. Steel spear was always autolearned at level 4 fabrication. But now to craft it, you need to craft spear shaft, which are learned at level 5 fab. There is recipe of spear shaft at level 2 fab, but it need long pole, which is also need level 5 fabrication, so there in no way to obtain steel spear until level 5 fabrication
> 
> True that. Also, the bo staff is craftable at fab 2, and that involves arguably more rigorous oiling and sanding than a spear shaft, so no reason for it to be fab 5.
> 
> Always a lot of stuff to take into account when adding craftables sadly. 😭

Can somebody make pr for that? Because english is not my native language and its a bit painful to type on it

Nabo92:
> > And there is problem now. Steel spear was always autolearned at level 4 fabrication. But now to craft it, you need to craft spear shaft, which are learned at level 5 fab. There is recipe of spear shaft at level 2 fab, but it need long pole, which is also need level 5 fabrication, so there in no way to obtain steel spear until level 5 fabrication
> 
> True that. Also, the bo staff is craftable at fab 2, and that involves arguably more rigorous oiling and sanding than a spear shaft, so no reason for it to be fab 5.
> 
> Always a lot of stuff to take into account when adding craftables sadly. sob

Yeah, I'd say use the Bo as the base item, since it already includes good treatment in its recipe, and do a fast recipe to just sand and polish it a bit on the top to fit the spear head and call this item "spear shaft", could be a fab 3 auto-learn recipe, maybe? It would fix all these inconsistencies and you'd be using an item that it's already included in game.

Aerin-of-the-Toast:
> The simplest t1 knife spear for example, which is essentially just tying/taping a knife to a stick has been changed from 15m to 2h.. making it equal crafting time to the t2 which has bolting/nailing/drilling involved. Meanwhile your t3 knife spear is at a chill 30m while seemingly using the same bolting/nailing/drilling.

Most of the crafting time comes from carving down the long stick into something resembling a usable spear shaft. There probably is space for an even cheaper version without any carving that just involves breaking the branch off, but it would be much thicker, which would increase the volume and weight of the item dramatically and thus the attack cost. I figured 2 hours of relatively simple carving seemed like a reasonable sacrifice most people would be willing to make to not have a 3.5L spear with an attack cost closer to halberds.

> And there is problem now. Steel spear was always autolearned at level 4 fabrication. But now to craft it, you need to craft spear shaft, which are learned at level 5 fab. There is recipe of spear shaft at level 2 fab, but it need long pole, which is also need level 5 fabrication, so there in no way to obtain steel spear until level 5 fabrication

The difficulty of the steel spear comes from blacksmithing the spearhead. I'm neither a blacksmith nor a carpenter, but from what little I know, carving out a proper spear shaft requires a fair bit of skill while a spearhead is one of the more simpler things you can blacksmith. Anyone with relevant skill can correct me on this, but this isn't necessarily a problem, since it also leaves the possibility of *scavenging* a long pole (or some form of large wooden dowel) to turn into a spear shaft, once those are made lootable. 

> It's throwing an error on load related to my character already having an oversized spear(qiang) holstered in the spear strap. It can't figure out how to handle that and deletes the spear.

Yeah, that's just because this made the qiang longer now, as IRL qiangs were 2.5 meters at the minimum and the spear strap only holds 2m. I recommend debugging in a new qiang to replace the missing one.

> True that. Also, the bo staff is craftable at fab 2, and that involves arguably more rigorous oiling and sanding than a spear shaft, so no reason for it to be fab 5.

I just copied the recipe for the long pole and slightly adjusted it. It's possible the bo's fabrication requirement is far too low, or the long pole's is too high, or maybe both are right and the carpentry really is just a lot more complex for the long pole, etc...

To clarify overall, I'm not saying the spear shaft *has* to be 5 fabrication; it's very possible I'm overstating the difficulty and it should be closer to 2 or 3 or some other number. I am not a carpenter and as such, do not have experience in this regard, so I just copied another seemingly similar recipe (long pole), and it's possible I should have copied a *different* similar recipe (spear shaft). It would probably be best to open a PR to change it alongside your reasonings though.

I don't think a spear shaft will necessarily be *easier* than a bo though. It does require a fairly precise tapering to reduce weight properly without making it more fragile, whereas a bo can be a uniform diameter throughout.

anothersimulacrum:
Please don't have discussion on closed PRs, it gets lost. If there are issues after a PR has been merged, an issue (or new PR) is the appropriate place to discuss them.

PR #61490, opened 2022-10-06
Added ability to think to yourself in the message log

ashGlaw:
#### Summary
Interface "Added ability to think to yourself in the message log"

#### Purpose of change

The only way to write arbitrary strings to the message log is to yell. This is fine, if you don't mind making a lot of noise. Sometimes it would be better to leave yourself a note in the message log in a less potentially dangerous manner. Plus, it adds more roleplay potential for people who like that.

#### Describe the solution

I added a new enum to npctalk.cpp, put a new option in the npc chat menu, created a think message string, a new case to let you type in your character's thoughts, and an if statement that puts your string in the log.

I could have created a new action with its own keybind that wasn't involved with the chat menu. That would probably be more elegant, but this works, and I copied the syntax the shouting functions used in the same file.

#### Testing

I compiled and started a new game with Play Now. I brought up the npc chat menu and Think something was an option. I selected it and it prompted me for a string. I entered one and hit enter, making the string I typed following the words "You think"  appear in the message log.  It didn't cause any sound.

#### Additional context

I've considered making another npc chat option to say a string at a normal speaking volume instead of shouting, but the get_shout_volume stuff really confused me, so I stopped here.


ashGlaw:
There seems to be something wrong with speedydex tests, but that shouldn't have anything to do with this, right?

NotSeawhite:
I wonder if you don't type anything. Will the log say 'You think.'?

ashGlaw:
> I wonder if you don't type anything. Will the log say 'You think.'?

I tried it and it didn't print anything to the log at all. 👍

ashGlaw:
This has been ready to merge for a while now. Is there something I need to do before it's eligible?

PR #61501, opened 2022-10-06
2 new mutations for post thresh bear.

Nebnis:
#### Summary
Content "Added two new post thresh bear mutations"

#### Purpose of change
I figured that bear could use some more variety and added some mutations for unarmed styles on top of making them actually feel like a bear when they complete the transformation.

#### Describe the solution
Added Ursine paws: 15 bash unarmed bonus at the cost of even more hands encumbrance and debuffs to crafting. Grizzly bears can have a paw swipe force up to 25,200 lb. -ft./s (about 783 pounds of force), which is between two to five times stronger than the punch force of the average, untrained human. Average punching damage ingame is between 4~6 bash with 10 strength. However lacks of thumbs suck so that will impact them harder.

Added Crushing fangs: 10 bash 10 stab bite attacks to be in line with similar damaging bite attacks. Polar Bears have the strongest bite force of all bears, with a bite force of 1200 PSI. The Grizzly Bear comes in a close second with a bite force of about 1160 PSI. 

Add more requirements or downsides. Maybe dexterity debuff to crushing paws? But i guess the hand encumbrance already takes care of that. or a Strong requirement for crushing paws.

#### Testing
Logged and everything seems to be working with the mutations.

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -5683,7 +5683,7 @@
     "attacks": {
       "attack_text_u": "You bite %s",
       "attack_text_npc": "%1$s bites %2$s",
-      "blocker_mutations": [ "FANGS" ],
+      "blocker_mutations": [ "FANGS", "CRUSHING_FANGS"   ],
       "body_part": "mouth",
       "chance": 20,
       "base_damage": { "damage_type": "cut", "amount": 5 }
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -6110,7 +6110,7 @@
   {
     "type": "mutation",
     "id": "PAWS_HUGE",
-    "name": { "str": "Crushing Paws" },
+    "name": { "str": "Ursine Paws" },
     "points": 2,
     "visibility": 6,
     "ugliness": 3,
@@ -6135,7 +6135,7 @@
     "threshreq": [ "THRESH_URSINE" ],
     "prereqs": [ "PAWS_LARGE" ],
     "cancels": [ "TALONS" ],
-    "category": [ "BEAST", "URSINE" ]
+    "category": [ "URSINE" ]
   },
   {
     "type": "mutation",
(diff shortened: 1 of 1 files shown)


data/json/mutations/mutations.json
```diff
@@ -6104,8 +6104,60 @@
...
+    "prereqs": [ "MUZZLE_BEAR" ],
+    "threshreq": [ "THRESH_URSINE" ],
+    "category": [ "URSINE" ],
+    "attacks": {
+      "attack_text_u": "You bite a chunk out of %s",
+      "attack_text_npc": "%1$s bites a chunk out of %2$s",
+      "blocker_mutations": [ "FANGS" ],
+      "body_part": "mouth",
+      "chance": 20,
+      "base_damage": [ { "damage_type": "stab", "amount": 10 }, { "damage_type": "bash", "amount": 10 } ]
```
Maleclypse, line 6158: I'd be interested in how much damage this does as compared to FANGS on the same character.

Nebnis: Same as fangs, 20. I thought about just adding fangs to it, maybe i should? i figured a bear would have some bash value along with their bite rather than puncture alone.

Maddremor:
A completely subjective note, but I don't think they should both be named Crushing X. It seems a bit repetitive to me.

Nebnis:
> A completely subjective note, but I don't think they should both be named Crushing X. It seems a bit repetitive to me.

Maybe ursine paws is better then? If you got suggestions im all ears

Maleclypse (approved):
I'm sorry I'm just now getting back to this.

Maleclypse:
Did you mean to close this? as it’s waiting on 0.G release for any further review.

Nebnis:
> Did you mean to close this? as it’s waiting on 0.G release for any further review.

Ye i figured im gonna close this because the limbification of mutations will probably make the system much better, I might reopen again later when the system is in place

PR #61504, opened 2022-10-06
Balance and tweaks to relic procgen

anoobindisguise:
#### Summary
Balance "edits to relic progen for better cohesion"

#### Purpose of change
Relic progen was a little wonky, something I noticed after they were assigned resonance.

#### Describe the solution
Fixes include:
- ATTACK_SPEED was inverted - negative attack speed was being seen as "bad" and positive attack speed was being treated as "good" when in reality the opposite was the case. This has been corrected.
- some active effects have been changed. A few "bad" actives are worth less negative points, and a few "good" actives are now bad (FIRESTORM was worth 1000 points when it's really the most negative effect by far). Some really strong actives like healing and adrenaline are worth more points.
- Carryweight and stamina were worth way too many points, an artifact with 20kg carryweight had a ridiculous 6,000 resonance. Cut those down substantially (the same artifact has only 600 now) 
- ARMOR_BIO and ARMOR_COLD were removed from the pools because these are not used at all in the base game, so they were wholly superfluous effects. HEAT, ACID and ELEC were made to be worth less than CUT, STAB, BULLET and BASH. An artifact that makes you twice as weak to shocks is not as bad as an artifact that makes you twice as weak to bullets.
- Added relic effects for FOOTSTEP_NOISE, WEAPON_DISPERSION, RECOIL_MODIFIER, REGEN_STAMINA and EFFECTIVE_HEALTH_MOD. 
- A few passive relic effects like shout noise are worth less power., A few passive effects such as speed are worth more value.

Balance is hard to nail down and not necessarily even good in this context, but I wanted to fix the obviously wacko stuff at the very least.

#### Testing
There are basic build test failures but I don't *think* they're mine. Will be trying this in game.
Edit: have tested this for a bit and it seems to work fine. The only thing I might change is spawning more passive properties on spawned artifacts since they self-balance.

### DIFF
--- a/data/json/artifact/relic_procgen_data.json
+++ b/data/json/artifact/relic_procgen_data.json
@@ -13,36 +13,36 @@
       }
     ],
     "active_procgen_values": [
-      { "weight": 100, "spell_id": "AEA_ADRENALINE", "base_power": 150 },
-      { "weight": 100, "spell_id": "AEA_BLOOD", "base_power": 50 },
-      { "weight": 100, "spell_id": "AEA_HEAL", "base_power": 250 },
+      { "weight": 100, "spell_id": "AEA_ADRENALINE", "base_power": 250 },
+      { "weight": 100, "spell_id": "AEA_BLOOD", "base_power": 0 },
+      { "weight": 100, "spell_id": "AEA_HEAL", "base_power": 500 },
       { "weight": 100, "spell_id": "AEA_CONFUSED", "base_power": 250 },
       { "weight": 100, "spell_id": "AEA_PAIN", "base_power": -150 },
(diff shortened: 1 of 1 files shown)


Maleclypse:
I use those in the Xedra Procgens

KymasTuran:
I did something similar months ago with the intention of balance artifacts. But don't dare to upload it.
Made them extremely powerful but also very dangerous. The idea was that the artifact claim a "price" for is power.  Like a bad mutation, something as "the artifact claim your eyesight" and you end "near sight", but in the other side can clean the entire screen of Z (hurt all).
In others I heal but also made lots of noise. Stuff like that.
Another option was that the artifact only gain a charge in certain circumstances like after several kill count or player bleed.
Just my two cent.

PR #61516, opened 2022-10-07
Acidic blood refinement

flybye3o6:
new blood types for acid enemies that can be refined into useable chemicals

#### Summary
Content "different blood types for acidic monsters"

#### Purpose of change
different acidic monsters should have acid blood types more inline with what they have access to or produce themselves, you should be able to extract those acidic blood types when bleeding them from butchering, then refine it into the acid types they are based on

#### Describe the solution
acid blood for each acid type based on realistically what acid types monsters could use
recipes for refining these blood types into the acids they are based on
`blood_acid` is **muriatic** from stomach acid
`blood_acid_insect` is **formic** from venom glands
`blood_acid_plant` is **acetic** for anti-fungal, candidate for a triffid acid spitter 
`blood_acid_underground` is **sulphuric** for underground digger type creatures like worms and moles that can end up eating the stuff

also fixed up the PPE requirements
`gloves_medical` being disposable is a good idea it just didnt work as is, they would need to be changed into charges/ammo
`hazmat_suit` lacks a mask so would only give protection for gloves and skin but not eyes or gas
`ppe_gas` did not use gas mask filters in its configuration, gas masks do not provide respiratory protection without filters

#### Additional context
i am not a chemist, and only have a basic level of understanding of chemistry
im solid on the IDs but not on the names or descriptions, if someone wants to change them go ahead, descriptions are not my strong suit, i picked the names on unique adjectives for the acids

image
image


flybye3o6:
image
image
image
image
image
image
image
image
image

PatrikLundell:
Sulfuric acid doesn't sound very biological, but I don't know what type of acid would make some sort of sense. Nitric acid?

dseguin:
Sorry, accidentally undrafted this. I just meant to kick the tests :P

flybye3o6:
> Sulfuric acid doesn't sound very biological, but I don't know what type of acid would make some sort of sense. Nitric acid? Formic acid could make sense for acid ants.

i agree i think it should be something else but when you use the blood draw kit on these enemies you get Sulfuric acid so i thought id start there
image
image

i can imagine a system where there are multiple types of acid blood for the different types of enemies that have it, all under the same name "acidic blood" and are only reviled after you have the proficiency for it "acidic blood[formic]" say "Principles of Chemistry" or "Organic Chemistry" and will also unlock the recipes to do something with them
you get "Biochemistry" for butchering them already so something before that makes sense to me

that is beyond my ability atm though and i should prob just pick something that makes sense and is balanced for now

data/json/recipes/recipe_medsandchemicals.json
```diff
@@ -894,6 +894,21 @@
     "tools": [ [ [ "surface_heat", 10, "LIST" ] ], [ [ "platinum_small", -1 ] ] ],
     "components": [ [ [ "water_clean", 1 ] ], [ [ "chem_sulphur", 375 ] ] ]
   },
+  {
+    "type": "recipe",
+    "activity_level": "LIGHT_EXERCISE",
+    "result": "chem_sulphuric_acid",
+    "id_suffix": "with blood",
```
TheShadowFerret, line 901:
```suggestion
    "id_suffix": "with_blood",
```
Spaces in suffixes are valid, but also bad form.

flybye3o6:
i made the zombie acid muriatic from stomach acid, bug acid formic from venom glands, plant acid acetic for anti-fungal, and a acid blood type for underground digger type creatures sulphuric for like worms and moles that can end up eating the stuff

other things of note: you can fill a glass bottle with any of these acidic bloods and throw it at enemies or on the ground and it will create short lived acid puddle, pretty effective weapon if you throw a few on the ground at a doorway and pull enemies through it
the snail types spit acid but they dont explode into acid puddles when they die so idk if they would have acid blood or if they have a gland or something
idk how to change the blood draw kit to use these blood types, it still pulls sulphuric acid out of all these creatures
dark wyrm uses the "arachnid_acid" harvest type, unsure it that needs to be changed
formic acid explodes in fire, should the blood do the same?
the recipes still need work but i dont know much about chemistry

MNG-cataclysm (approved):

flybye3o6:
i changed the recipes, they use alcohol to kick the acid out of solution, and a 2nd method that uses distillation to recover and separate most of the components
i wanted to use "byproduct_group": [ { "item": "chem_ethanol", "charges": [ 200, 250 ] } ],
to give variation to the returned on alcohol and other byproducts because its only meant to be 80%+ efficient, but could not get it to stop spawning containers for the byproducts to come in

PR #61517, opened 2022-10-07
swords looks_like fix

nopenoperson:
#### Summary
None

#### Purpose of change
many swords in swords_and_blades.json have no tiles, this will let most of them have tiles

#### Describe the solution
added looks_like to a bunch of swords that every other quality of the same sword copy from. 

adding "looks_like": "kukri" to the falx, not adding "looks_like": "scimitar" to the kreigsmesser (im 99% sure msx would use the scimitar tile), letting someone else do it 

#### Testing
some


Fris0uman:
More looks_like is always a good idea

Issue #54872, opened 2022-01-28
Update "restricts_gear" in mutations to use the new bodyparts


Jarewill:
### Is your feature request related to a problem? Please describe.

Currently mutations restrict gear if worn on a specific limb, even if it doesn't make any sense, such as the Hub 01 greaves being blocked out due to having a tail, even if they only cover the shins and knees.
There is a flag that gets around this issue called ALLOWS_TAIL, but I believe I have a better idea.

### Solution you would like.

The idea is to be able to specify what part of the limb a mutation restricts, such as tails restricting only hips. (and maybe thighs)
That way the tail mutations will no longer block out any items they shouldn't, just because they lack the relevant flag.

This will also allow for more specific mutations, for example: Elbow spikes that give a strong attack, but prevent wearing long sleeved clothing/armor.

### Describe alternatives you have considered.

Keep using the ALLOWS_TAIL instead.
Also give that flag to the Hub 01 greaves in this case.

### Additional context

In #54810 I was told to open a new issue about it, so I did.
I apologize for necro'ing a closed issue from 2 days ago, I didn't realise when it was closed until after I posted there.

bombasticSlacks:
To paraphrase @Venera3 who is working on the overhaul to something similar to this suggestion they said something along the lines of: I haven't really looked into the sublimb stuff yet.

So this (or something similar) is in the pipeline. The delay is Venera is already plugging away at stuff in an order they want, and no one else really seems to have the drive. If someone wanted to implement this suggestion now it probably wouldn't be too hard. Mutations can have a "blocks_sublimb" attribute, or something similar and then where the code normally checks it also compares the bodies blocked sublimbs vs the armors covered sublimbs.

Venera3:
The armor stuff will take somebody codier than me to hook up properly(wink, wink, nudge nudge). The way I could imagine it is filtering based on the sublimbs the bodypart is defined with, this would of course need armor to accept alternative coverage arrays -e.g. elbow pads either covering left/right elbow or Insect Leg legs' left/right 1st joint, stuff like that. When *that* happens we'd need a way for a bodypart to define which sublimb to block from the connecting upstream/downstream part - tails blocking waist on your torso, if you have a torso with a waist. 

I'll need to give the system a think and write up an issue on how it could work, but I haven't found the time yet.

Jarewill:
Alright, I see.
Thank you both for your time.

[...]

PR #61521, opened 2022-10-07
Mutation gear restrictions can target sub-bodyparts, and check for rigidity only there

actually-a-cat:
#### Summary
None

#### Purpose of change
We have a wonderfully specific (sub)body part system that allows us to reason about which bits of gear go exactly where and what they are like. However mutation conflicts with gear do not use it. Instead they conflict when *any* part of the item is rigid, even if that's not the part that covers the mutation site. Additionally mutations can only target top-level body parts, not sub body parts. In situations where that doesn't make sense we then go on to special-case it with override flags like ALLOWS_TAIL. This is unnecessarily manual.

#### Describe the solution
* Allow mutation conflicts to specify sub body parts in addition to main body parts
* Only conflict with rigid items when they are rigid on the mutation site specifically (doesn't affect conflicts with soft items)
* Narrow mutation conflicts where appropriate, e.g. tails conflict with hips, talons with toes...
* Remove override flags where this made them redundant

This doesn't completely remove the need for override flags. For example, fingerless gloves still need ALLOWS_TALONS because they have 50% coverage on fingers, and fingertips aren't a separate part. But it does eliminate them in many cases.

#### Testing
Made sure that items that lost ALLOW_TAIL but don't cover hips can be worn with tail.
Made sure rigid items that cover hips cannot be worn with tail.
Made sure soft conflicts work by trying to wear normal clothes while Huge
Checked that talons conflict with normal gloves but not fingerless
Checked that toe talons conflict with shoes
Checked that hooves conflict with shoes but not with sabatons

### DIFF
--- a/data/json/items/armor/legs_armor.json
+++ b/data/json/items/armor/legs_armor.json
@@ -17,7 +17,7 @@
     "color": "brown",
     "warmth": 5,
     "material_thickness": 3,
-    "flags": [ "OUTER", "WATER_FRIENDLY", "ALLOWS_TAIL" ],
+    "flags": [ "OUTER", "WATER_FRIENDLY" ],
     "armor": [
       {
         "covers": [ "leg_l", "leg_r" ],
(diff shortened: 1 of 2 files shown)


Jarewill:
This should close #54872.

ashGlaw:
Yeah, this is an ambitious and useful contribution, thanks so much. Great idea!

PR #61523, opened 2022-10-07
New Zombie type: Tearjerker

Daved27hundred:
#### Summary
Content "Adds the Tearjerker, a zombie that explodes into a cloud of teargas upon death."

#### Purpose of change

I got this idea after messing with teargas in #61484 and noting how dangerous it was, Not only causing a stat decrease but blinding the player as well. A zombie filled with this stuff would be an interesting variant of the regular poisonous zombie.

#### Describe the solution

discovered there was a death effect in monster_deaths.json called "death_tearburst" and used that as it's death effect. I assume it's from a teargas hack.

Not adding it.

#### Testing

Didn't test yet.

### DIFF
--- a/data/json/monsters/zed_explosive.json
+++ b/data/json/monsters/zed_explosive.json
@@ -211,8 +211,7 @@
   {
     "id": "mon_zombie_tear_gasbag",
     "type": "MONSTER",
-    "name": { "str": "tearjerker zombie" },
-    "//": "(I cannot think of a better name.)",
+    "name": { "str": "swollen tearjerker" },
     "description": "A bloated corpse, its ballooning skin blistered and rashy and its face caked in mucus.  The off-colour orange gas that occasionally leaks from it makes your eyes water and smells faintly of pepper.",
     "default_faction": "zombie",
     "bodytype": "human",


ashGlaw:
Where do you think this zombie would spawn?

Daved27hundred:
> Where do you think this zombie would spawn?

The description is supposed to imply that this was someone who was peppersprayed before they turned, so it'd be something you'd occasionally see in the streets, a leftover from the riots.

Theawesomeboophis:
I have a name suggestion, Crying Bomber, it references the tear gas and it's relation to boomers and other exploding enemies, and I think it sounds cool.

Daved27hundred:
> I have a name suggestion, Crying Bomber, it references the tear gas and it's relation to boomers and other exploding enemies, and I think it sounds cool.

I’ve started to take a liking to “tearjerker”, but changing the name to better indicate it’s a bloated zombie is a good idea. How’s “Swollen Tearjerker” sound?

I’m currently in a car heading to another province for thanksgiving so I can’t make the change right now, but someone else is welcome to.

MNG-cataclysm (approved):

Issue #61051, opened 2022-09-16
Cannot build most bridge types on deep water tiles


Beatdotroot:
### Describe the bug

A large majority of the bridge types require flowing deep water in particular for some reason, making building docks off of things like prison islands in lakes essentially impossible. The only construction that can be made in deep water is the pontoon bridge, which doesn't make much sense.

### Steps to reproduce

1: Make a player with suitable resources to build a bridge.
2: Attempt to build most any bridge in deep water.
3: Be unable to construct due to location restrictions.

### Expected behavior

I expected that you would be able to build a deep water dock in a similar, if not easier way as you would a flowing deep water dock.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.1889 (21H2)
- Game Version: fd463d4 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Tamable Wildlife [Tamable_Wildlife],
    Stats Through Kills [stats_through_kills],
    SpeedyDex [speedydex],
    Mythical Martial Arts [MMA],
    Mining Mod [Mining_Mod],
    My Sweet Cataclysm [my_sweet_cataclysm],
    Alternative Map Key [alt_map_key],
    Gun Cotton [guncotton],
    Generic Guns [generic_guns],
    Blaze Industries [blazeindustries]
]

PR #61526, opened 2022-10-07
Dock Construction Updates

Beatdotroot:
#### Summary
Bugfixes "Allows certain docks to be placed on non-flowing shallow and deep water."

#### Description
#61051

#### Describe the solution
Adds 2 new dock construction options in "build_deep_river_dock" and "build_river_dock_shallow_bridge" groups

Unsure.

#### Testing
Creating a testing world, spawned resources with debug menu, and built multiple bridges.

#### Additional context
N/A

### DIFF
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -4820,7 +4820,7 @@
   },
   {
     "type": "construction",
-    "id": "constr_dock",
+    "id": "constr_dock_moving",
     "group": "build_river_dock_shallow_bridge",
     "//": "Set up dock",
     "category": "CONSTRUCT",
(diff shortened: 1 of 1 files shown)


ashGlaw:
The construction menu needs so much attention. Thank you for giving it some love <3

MNG-cataclysm (approved):

Beatdotroot:
Of course, thank you!!

Fris0uman:
Did you check that changing the ids of the recipe doesn't affect save made before that change? Also does this lead to the magical situation where I'm digging fundation for support deep underwater with just a shovel and a hammer from the confort of any floating vehicle?

Beatdotroot:
I actually had made it with the express purpose of using it on a prior save, so it definitely works with saves made before the change. I also believe that second issue was technically already a thing you could do in a vehicle floating in a river, just not a lake.

data/json/construction.json
```diff
@@ -4820,7 +4820,7 @@
   },
   {
     "type": "construction",
-    "id": "constr_dock",
+    "id": "constr_dock_moving",
```
Maleclypse, line 4823:
```suggestion
    "id": "constr_dock_moving",/
```

data/json/construction.json
```diff
@@ -4820,7 +4820,7 @@
   },
   {
     "type": "construction",
-    "id": "constr_dock",
+    "id": "constr_dock_moving",/
```
Maleclypse, line 4823:
```suggestion
    "id": "constr_dock_moving",
```

PR #61531, opened 2022-10-08
Check covered body parts before setting sub body part uncomfortable

sparr:
#### Summary
Bugfixes "Covered body parts prevent uncomfortable sub body parts"

#### Purpose of change
A non-comfortable item worn on top of a comfortable item for the same body part should be padded and not cause uncomfortable status. This is currently not working when the non-comfortable outer item covers a sub body part and some inner layer comfortable item covers the whole body part.

#### Describe the solution
In addition to assembling a list of comfortable covered sub body parts to check when encountering a non-comfortable, also make and check a list of comfortable covered body parts.

#### Testing
I am playing and confirmed this solves my immediate problem of binoculars over a shirt+hoodie+jacket being uncomfortable. I have not tested enough other combinations to see if this PR is overkill and makes things comfortable that it shouldn't.

### DIFF
--- a/src/character_attire.cpp
+++ b/src/character_attire.cpp
@@ -1877,10 +1877,16 @@ std::unordered_set<bodypart_id> outfit::where_discomfort() const
 {
     // get all rigid body parts to begin with
     std::unordered_set<sub_bodypart_id> covered_sbps;
+    std::unordered_set<bodypart_id> covered_bps;
     std::unordered_set<bodypart_id> uncomfortable_bps;
 
     for( const item &i : worn ) {
         // check each sublimb individually
+        for( const bodypart_str_id &bp : i.get_covered_body_parts() ) {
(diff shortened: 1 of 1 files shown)


bombasticSlacks:
I'm pretty sure this is too wide of a solution. 

Since for example a t-shirt would make elbow armor now comfortable.

sparr:
I think the appropriate solution is to only count armor that covers a body part without specifically_covers'ing a [different] sub body part. I'm trying to figure out how to implement that.

bombasticSlacks:
That's impossible, something that doesn't specifically cover any subparts, covers all subparts internally. I already explained the correct solution to this on the Dev Discord. There isn't a more convenient solution that works as far as I'm concerned.

sparr:
Closing in favor of #61545. It turns out that the exclusion of secondary parts from implicit coverage is closer to the root cause here.

PR #61539, opened 2022-10-08
enchantments now can alter skills

KorGgenT:
#### Summary
Features "enchantments now can alter skills"

#### Purpose of change
more enchantment versatility
also drew sort of half mentioned it on discord and i thought it sounded like a pretty quick code, plus low impact in regards to feature freeze.

#### Describe the solution
added a member to enchantment and associated member functions that do something similar to `values`.

none

#### Testing
i'll be making at least one test json item

### DIFF
--- a/src/magic_enchantment.h
+++ b/src/magic_enchantment.h
@@ -269,8 +269,8 @@ class enchant_cache : public enchantment
         std::map<enchant_vals::mod, double> values_multiply; // NOLINT(cata-serialize)
 
         // the exact same as above, though specifically for skills
-        std::map<skill_id, int> skill_values_add;
-        std::map<skill_id, int> skill_values_multiply;
+        std::map<skill_id, int> skill_values_add; // NOLINT(cata-serialize)
+        std::map<skill_id, int> skill_values_multiply; // NOLINT(cata-serialize)
 };
 
 template <typename E> struct enum_traits;


src/character.cpp
```diff
@@ -3048,12 +3048,13 @@ SkillLevel &Character::get_skill_level_object( const skill_id &ident )
 
 int Character::get_skill_level( const skill_id &ident ) const
 {
-    return _skills->get_skill_level( ident );
+    return std::round( enchantment_cache->modify_value( ident, _skills->get_skill_level( ident ) ) );
```
anothersimulacrum, line 3051: `_skills` is private to modify only, its data is exposed publicly through `get_all_skills()`, so you need to expand this check to more places.

KorGgenT: oh drat, that's a little awkward... it returns a ref. the easiest thing i can think of is to walk that back and make it a copy

KorGgenT:
image
here we go.

John-Candlebury:
Oh the eel rings of Magiclysm were supposed to do this instead of giving extra dodges per turn. Those can serve as the test items

KorGgenT:
looks like clang is mad that i wrote the serialize function in such a way as that it only indirectly deals with skills_value_add  and skills_value_multiply

src/character.h
```diff
@@ -2127,7 +2127,7 @@ class Character : public Creature, public visitable
         int get_knowledge_level( const skill_id &ident ) const;
         int get_knowledge_level( const skill_id &ident, const item &context ) const;
 
-        const SkillLevelMap &get_all_skills() const;
+        const SkillLevelMap get_all_skills() const;
```
anothersimulacrum, line 2130:
```suggestion
        SkillLevelMap get_all_skills() const;
```

PR #61542, opened 2022-10-08
Makes headscarves skintight and autolearned

Light-Wave:
#### Summary
Content "Makes headscarves skintight and autolearned"

#### Purpose of change
In Innawood, there is no good autolearned clothing you can wear under helmets during the summer. Sure, there is the balaclava, but it's overly warm.

#### Describe the solution
Makes headscarves skintight and gives them an autolearned recipe, using the same requirements as the bandana.

Undercaps could also work, but they seem to traditionally only be worn by muslim women, so maybe not everybody's go-to choice when they want to wear a helmet.
A thinner version of a helmet liner
Make all helmets be on the outer layer
Make the scarf have a bit less warmth

#### Testing

Tested crafting it and wearing it under a helmet. It worked fine.

### DIFF
--- a/data/json/items/armor/scarfs.json
+++ b/data/json/items/armor/scarfs.json
@@ -37,7 +37,7 @@
     "warmth": 5,
     "material_thickness": 0.1,
     "environmental_protection": 1,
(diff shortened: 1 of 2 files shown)


Maleclypse (approved):
I think this makes sense

PR #61544, opened 2022-10-08
Bites can't target eyes and mouth by default

Venera3:
#### Summary
Balance "Bites don't target tiny bodyparts by default"

#### Purpose of change

While combat testing the (very special-attack-heavy) amalgamations I noticed them biting eyes/mouths leading to some extra damage, which felt a bit on the weird side.

#### Describe the solution

If no `hitsize_min` is defined for a bite-type attack it will default to 1 - meaning eyes and the mouth are discarded but hands are fair game. If one wants a bite that can hit eyes you can add `hitsize_min: 0`.

None

#### Testing

Using debugmode message logs confirmed no targeting of eyes/mouths unless hitsize is set to 0.

### DIFF
--- a/doc/MONSTER_SPECIAL_ATTACKS.md
+++ b/doc/MONSTER_SPECIAL_ATTACKS.md
@@ -203,7 +203,8 @@ The common type for JSON-defined attacks. Note, you don't have to declare it in 
 
 ### "bite"
 
-Makes monster use teeth to bite opponent, uses the same fields as "monster_attack" attacks. Monster bites can give infections if the target is grabbed at the same time.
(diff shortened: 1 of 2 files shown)


Venera3:
That would mean every *other* nonrestricted attack would need `hitsize_min: 0` added. I have long-term plans re unhardcoding bites' functionality and doing away with the explicit bite-type, but that's ETA whenever.

Venera3:
bite_actor uses melee_actor::load_internal to keep the arguments consistent and appends the hitsize default and infection chance - generic `melee`-typed attacks are not affected by the change.

anoobindisguise:
The fix I didn't know I needed. Zombies biting my eyes was horrifying but also really silly.

PR #61545, opened 2022-10-08
Count full-coverage armor as having coverage of secondary sub parts

sparr:
#### Summary
Bugfixes "Full coverage armor covers secondary sub parts"

#### Purpose of change
Secondary sub body parts are excluded when treating a covered body part as implicitly covering all sub parts. This leads to uncomfortable items in secondary locations not being padded/softened by comfortable items that cover the whole body part under them.
Specifically: At present, a non-comfortable item in torso_hanging_front, like binoculars, or torso_hanging_back, like a dive_bag, causes the chafing/uncomfortable penalty, because clothes that cover the whole torso don't [under-]cover those (or any) secondary sub body parts.

#### Describe the solution
Eliminate the exclusion, include secondary sub body parts when adding implicit coverage of sub body parts when a body part is covered.

@bombasticSlacks suggests adding a field to secondary body part definitions that indicates which non-secondary part is under/behind them. This could then be selectively used during coverage checks.
I have also considered an approach where the implicitness of sub_coverage is tracked and can be checked. This would be usable in places like the debugmsg check that this PR disables.

#### Testing
I am playing with this PR and confirm that it resolves my original problem with binoculars causing discomfort. I am not familiar enough with the other use cases of sub part coverage to thoroughly test situations such as item damage, wetness, etc.

#### Additional context
I don't expect this PR to get merged. The current state of secondary sub part coverage logic is complex and will require much more effort than this in the future. I am publishing this because I would prefer to give up a JSON rule enforcement rather than play with a game mechanic bug, and I think others would use this patch as well.

### DIFF
--- a/src/item_factory.cpp
+++ b/src/item_factory.cpp
@@ -1886,18 +1886,18 @@ void Item_factory::check_definitions() const
                 }
             }
 
-            // check the hanging location aren't being used on the non strapped layer
-            if( std::find( type->armor->all_layers.begin(), type->armor->all_layers.end(),
-                           layer_level::BELTED ) == type->armor->all_layers.end() ) {
-                for( const armor_portion_data &portion : type->armor->sub_data ) {
-                    for( const sub_bodypart_str_id &sbp : portion.sub_coverage ) {
-                        if( sbp->secondary ) {
-                            msg += string_format( "Secondary hanging locations should only be used on the BELTED layer: %s\n",
-                                                  sbp.str() );
-                        }
-                    }
-                }
-            }
+            // // check the hanging location aren't being used on the non strapped layer
+            // if( std::find( type->armor->all_layers.begin(), type->armor->all_layers.end(),
+            //                layer_level::BELTED ) == type->armor->all_layers.end() ) {
+            //     for( const armor_portion_data &portion : type->armor->sub_data ) {
+            //         for( const sub_bodypart_str_id &sbp : portion.sub_coverage ) {
+            //             if( sbp->secondary ) {
+            //                 msg += string_format( "Secondary hanging locations should only be used on the BELTED layer: %s\n",
+            //                                       sbp.str() );
+            //             }
+            //         }
+            //     }
+            // }
 
             // waist is deprecated
             if( type->has_flag( flag_WAIST ) ) {


bombasticSlacks:
This means every item with full coverage on the torso and legs will be hit twice and absorb damage twice. 

You deleted error testing to accomplish this. 

And it still doesn't fix the original bug since something like a crop top or the light ballistic vest that's just upper torso STILL won't count as under a rifle or backpack.

sparr:
> This means every item with full coverage on the torso and legs will be hit twice and absorb damage twice.

I'd like to understand how/why this will happen. Is there specific logic to hit secondary coverage areas in addition to non-secondary sub parts?

> You deleted error testing to accomplish this.

Yes, and I called that out in the original comment. Given the choice between a bug that is always present in the game and a bug that will only appear through potential future inappropriate JSON entries, I think the latter is preferable.

> And it still doesn't fix the original bug since something like a crop top or the light ballistic vest that's just upper torso STILL won't count as under a rifle or backpack.

Should it? A rifle will also hit the lower torso and its strap will also wrap around the lower torso. Some, possibly most, backpacks hang with pressure and friction on the lower torso. Something that covers only the upper torso shouldn't protect from discomfort from these items, unless I'm misunderstanding what chafing/discomfort is meant to represent.

Did you intend for the torso hanging sub parts to only be on top of the upper torso? That doesn't seem to match where the implementation is headed at this time.

> As a corollary something like a stomach wrap would "cover" a pair of binoculars with this change. So this barely fixes anything and breaks a pile of stuff.

I would expect a stomach wrap to explicitly specifically cover the lower torso, and so not trigger the logic that applies coverage to other sub parts.

kevingranade (changes requested):
This makes sweeping changes, but the only specific issue called out is a trivial corner case, and that case isn't fully described, it's just something about binoculars and encumbrance?

bombasticSlacks:
>  Sub-part-specific chafing would also be a viable direction to go, but we aren't there yet.

We are already there, you are the one trying to break it in weird ways and simplify it to the point of not working. 

> I think the chafing/discomfort system should be a lot more nuanced in general. I'd probably start with weight; a 300 gram item shouldn't chafe as much as a 30kg item

Nothing chafes that much. It's a small penalty to encourage proper roleplaying (wearing clothes). If you extend the system the only possible thing to do is make it worse for players which seems pointless considering the system is alleviated by wearing pants and a shirt. 

I'm not going to argue with you on specific nuances of a proper solution since the point of a proper solution is to handle all the edge cases. You just keep deleting chunks of code and making small changes that break as much as you fix and that *won't be good enough*.

sparr:
@kevingranade I updated OP with more details on the specific situations that I and other reporting players are encountering.

I can't figure out why this exclusion was included in the first place. Attempting to discuss it on Discord was not proving fruitful. I'd be happy to implement a better solution with a bit more information about why this is implemented the way it is now, and what effects the exclusion (or removing it) does/should have.

dseguin:
Superseded by #61576

PR #50143, opened 2021-07-23
Json loading from efficient binary representation for performance wins

akrieger:
#### Summary
Performance "Binary json input for various performance wins"

#### Purpose of change

Performance profiling shows that a lot of time is spent inside the json input stack simply manipulating json. A hypothesis I had for a long time is that there is potential for wins here by using an efficient 'pre-parsed' binary representation of the json. Simply making loading faster is a worthwhile goal in and of itself, but we can also unify the json input stack off the 'streaming' model and have everything consistently using the object oriented deserialize approach. Furthermore, there is potential for runtime wins by 'externing' strings sourced from json. By changing data types to hold pointers into the binary json for each string, instead of a std::string member with a possible heap allocation, we can shrink objects and save on runtime with faster string comparisons. This latter approach has not yet been explored.

#### Describe the solution

First, a description of the binary format being used. Google FlatBuffers are a binary hierarchical data representation that can basically be described as 'typed json', except they require a schema defined ahead of time. For slightly less efficient representation, however, there are FlexBuffers, which do not require a schema. A full description can be found at https://google.github.io/flatbuffers/flexbuffers.html. The shorthand description is that the representation is a flat binary file consisting of vectors and possibly cached/reused values. Objects are represented as two vectors: a key vector, and a value vector. Arrays are simply a single value vector. The key and value vectors store offsets into the binary to where the keys/values are actually stored. The FlexBuffer parser is configurable to share keys, string values, and key vectors across the input. So, a series of objects with the same keys will share the same key vector in the final binary. Repeated usage of the same string value will be reduced to a single string in the output.

The solution I have implemented starts by deprecating almost all usage of `JsonIn` in the entire project. I did first attempt to reimplement the semantics of `JsonIn`, except backed by the binary, but ran into a long series of issues when crossing between code using `JsonIn`, converting to eg. `JsonObject`, then back to `JsonIn` using get_raw(). Simply eliminating get_raw() was insufficient for resolving these issues. Ultimately I decided to remove `JsonIn` because it is essentially lying to the user that the performance is different over `JsonObject/JsonArray`. All parsing would be effectively O(1) post conversion.

[...]

PR #61555, opened 2022-10-08
Initialize datadir_path_value earlier in path_info

perryprog:
#### Summary
None

#### Purpose of change
This seems to correct an edge case that I was running into that would cause `prefix_path` to get initialized to nothing, causing Cataclysm to crash with a segfault whenever I started it outside of the root of the repo. (macOS, CMake, release build, no odd build flags besides that.)

Cataclysm ran fine at the root of the git directory as there's the data/ sub-directory there, which is what CDDA would search for as `gfxdir_path_value` would end up being simply "data/" instead of the expected release value of "/usr/local/share/cataclysm-dda/" or similar. The crash itself probably occurs (instead of being caught within main.cpp) because the file exists checks worked fine since it was checking `datadir_value`, so the game assumed `gfxdir_path_value` was safe to use—I'm not to worried about this aspect of the issue, though, since this resolves the crash anyway.

#### Describe the solution
I moved the initialization of `datadir_path_value` to be earlier in `PATH_INFO::set_standard_filenames`.

None.

#### Testing
Building and running under CMake works without manually specifying `--datadir` (which uses a different code path to initialize `datadir_path_value`, and will end up working as-is)

#### Additional context
This looks like it was introduced as a part of d329307a51126a3cb93d32fc6a99f20ae87a2367 in #50143, so hopefully @akrieger can take a look at this to make sure this fix is reasonable. I'm fairly certain there aren't any edge cases that I'm unable to check that this change breaks.

### DIFF
--- a/src/path_info.cpp
+++ b/src/path_info.cpp
@@ -104,6 +104,10 @@ void PATH_INFO::set_standard_filenames()
     // Special: data_dir and gfx_dir
     std::string prefix;
     cata_path prefix_path;
+
+    // Data is always relative to itself. Also, the base path might not be writeable.
+    datadir_path_value = cata_path{ cata_path::root_path::data, fs::path{} };
+
     if( !base_path_value.empty() ) {
 #if defined(DATA_DIR_PREFIX)
         datadir_value = base_path_value + "share/cataclysm-dda/";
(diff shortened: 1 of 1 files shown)


akrieger (approved):
Good find, thanks!

src/path_info.cpp
```diff
@@ -104,6 +104,10 @@ void PATH_INFO::set_standard_filenames()
     // Special: data_dir and gfx_dir
     std::string prefix;
     cata_path prefix_path;
+
+    // Data is always relative to itself. Also, the base path might not be writeable.
+    datadir_path_value = cata_path{ cata_path::root_path::data, fs::path{} };
+
     if( !base_path_value.empty() ) {
 #if defined(DATA_DIR_PREFIX)
```
akrieger, line 112: Yeah the problem happens inside this block at line 111 (before code). I didn't notice that prefix_path was reading from `datadir_path_value` which clearly isn't initialized yet. Then `prefix_path_value` is forever bad and is what causes issues down at line 127.

PR #61557, opened 2022-10-09
Update suits_protection.json

Veskaida:
#### Summary
None

#### Purpose of change

Fixes typos in the flavor text of the item descriptions for 'Tempered Steel Armor' and its variants (Light, Normal and Heavy). The item descriptions are all similar to one another, but have small deviations in the hyphenation. They currently read as: "... top-of-the-line protection.", "... top of the line protection." and "... top-of-the-line-protection.". This fix standardizes all three of these texts as "... top-of-the-line protection.".

#### Describe the solution

Fixed the item description text for "Tempered Plate Armor" and "Tempered Heavy Plate Armor" to read as the same as the item description text for "Tempered Light Plate Armor", which reads as: "... top-of-the-line protection.".

Each type of plate armor could have its own unique description, but given there are 3 types (Light, Normal and Heavy) and three sizes (XS, Normal and XL) and that their descriptions and functions are more similar than different, I felt it was more appropriate to bring all the items descriptions in-line with one another than to deviate further. I think that having wildly different descriptions, for so many similar items may be more confusing to new players than it would be beneficial, so I decided against this.

#### Testing

Loaded up a test world and checked the new descriptions with debug menu. It worked!

#### Additional context

new

### DIFF
--- a/data/json/items/armor/suits_protection.json
+++ b/data/json/items/armor/suits_protection.json
@@ -991,7 +991,7 @@
     "copy-from": "armor_lc_plate",
     "material": [ "qt_steel", "qt_steel_chain" ],
     "name": { "str": "tempered plate armor" },
-    "description": "A suit of plate armor with a 4mm thick chestpiece.  The medium steel has been quenched and tempered, offering top-of-the-line-protection.",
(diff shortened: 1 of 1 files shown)


Veskaida:
Minor typo fixes/standardized the text in the item description for the Tempered Steel Plates, to more accurately reflect one another.

PR #61564, opened 2022-10-09
mr 73 high end revolver

nopenoperson:
#### Summary
Content "rare high end .357 revolver"

#### Purpose of change
the game has 0 six shot revolvers in 357 magnum so i wanted to add one

#### Describe the solution
did a bunch of research online and some calculations then put the stuff in 38.json. used .38 as a template for the nested ammo stuff. they are tested to shoot atleast 2.8 moa before leaving the factory so i gave it 280 dispersion. these things are supposed to hold up to constant horrible abuse day after day for years so 10 reliability .

making it more or less accurate. spending more time figuring out volume. asking for advice on item groups. not making this pr

#### Testing

yes its a revolver that has good dispersion and the correct speedloader works.  its very potent with the weaker 38 ammo because of the synergy that the great dispersion has with lower recoil ammo but you only have 6 shots so it doesnt really get out of hand 
the complicated nested ammo thingy works pretty well and if you make the json entry simpler,  it doesnt produce very good results in my testing

### DIFF
--- a/data/json/items/gun/38.json
+++ b/data/json/items/gun/38.json
@@ -133,10 +133,10 @@
       [ "bore", 1 ],
       [ "grip", 1 ],
       [ "mechanism", 4 ],
-      [ "rail", 1 ],
-      [ "sights", 1 ],
-      [ "stock", 1 ],
-      [ "underbarrel", 1 ]
+      [ "rail mount", 1 ],
+      [ "sights mount", 1 ],
+      [ "stock mount", 1 ],
+      [ "underbarrel mount", 1 ]
     ],
     "pocket_data": [
       {


data/json/items/gun/38.json
```diff
@@ -103,6 +103,50 @@
       }
     ]
   },
+  {
+    "id": "mr73",
+    "copy-from": "pistol_revolver",
+    "looks_like": "sw_619",
+    "type": "GUN",
+    "name": { "str_sp": " Manurhin MR 73 " },
```
TheShadowFerret, line 111:
```suggestion
    "name": { "str_sp": "Manurhin MR 73" },
```

tenmillimaster:
Looks like the rarity is about right . I compared it to the raging bull in 454 and BFR in 45-70.

I suggest adding .38 special to the on hand, 357 mag ammo  item group.

I question the mod locations this has available, but that's a problem with most guns.

nopenoperson:
> I suggest adding .38 special to the on hand, 357 mag ammo item group.
you can type out out what you think it should be as suggestion if you feel like it

data/json/items/gun/38.json
```diff
@@ -103,6 +103,50 @@
...
+    "blackpowder_tolerance": 56,
+    "clip_size": 6,
+    "valid_mod_locations": [
+      [ "accessories", 2 ],
+      [ "barrel", 1 ],
+      [ "bore", 1 ],
+      [ "grip", 1 ],
+      [ "mechanism", 4 ],
+      [ "rail", 1 ],
+      [ "sights", 1 ],
```
tenmillimaster, line 137: This should be a sight mount if there's no rail.

data/json/items/gun/38.json
```diff
@@ -103,6 +103,50 @@
...
+    "durability": 10,
+    "blackpowder_tolerance": 56,
+    "clip_size": 6,
+    "valid_mod_locations": [
+      [ "accessories", 2 ],
+      [ "barrel", 1 ],
+      [ "bore", 1 ],
+      [ "grip", 1 ],
+      [ "mechanism", 4 ],
+      [ "rail", 1 ],
```
tenmillimaster, line 136: Same here, there's no side rail.

data/json/items/gun/38.json
```diff
@@ -103,6 +103,50 @@
...
+      [ "sights", 1 ],
+      [ "stock", 1 ],
+      [ "underbarrel", 1 ]
```
tenmillimaster, line 139: And definitely no underbarrel rail.

tenmillimaster (changes requested):
Please review the mod location comments.

tenmillimaster (approved):
Looks good to me.

PR #61569, opened 2022-10-09
Chainmail fix

Dan1ss1mo:
#### Summary
Bugfixes "Chainmail weight inconsistency fix"

#### Purpose of change
Curantly there is 2 problems with chainmail
1) chainmails suit and hauberk magicaly weights more than its component
2) Hauberk do not have chainmail leggings attached to them. Turning your chainmail into jumpsuit isn't realy a good idea because chain is not particulary elastic material 
3) weight of some chaim pieces is just wrong, there is now way chain gloves , chausses and coif would have the same weight

#### Describe the solution
Adjust chainmail armor parts weight and detach leggings from hauberk

#### Additional context
For calculating weight i used 1,9m2 as total body area. Around 34% of it is torso, 15% arms, 23% legs, 5 hand, 7 feets and 7 for head with neck but with out face(total is below 100 % becouse i do not include face coverage, plus its probably unnecessary to count total area of buttocks and each individual finger of the leg fot most clothing). Would be nice to add this walues to ARMOR_BALANCE_AND_DESIGN doc

### DIFF
--- a/data/json/recipes/armor/suit.json
+++ b/data/json/recipes/armor/suit.json
@@ -968,7 +968,7 @@
       { "proficiency": "prof_closures", "fail_multiplier": 1 },
       { "proficiency": "prof_chain_armour", "fail_multiplier": 1 }
     ],
-    "components": [ [ [ "xs_lc_chainmail_hood", 1 ] ], [ [ "xs_lc_chainmail_hauberk", 1 ] ], [ [ "xs_lc_chainmail_legs", 1 ] ] ]
+    "components": [ [ [ "xs_lc_chainmail_hood", 1 ] ], [ [ "lc_chainmail_hauberk_xs", 1 ] ], [ [ "xs_lc_chainmail_legs", 1 ] ] ]
   },
   {
     "result": "mc_chainmail_suit",
@@ -1069,7 +1069,7 @@
       { "proficiency": "prof_closures", "fail_multiplier": 1 },
       { "proficiency": "prof_chain_armour", "fail_multiplier": 1 }
     ],
-    "components": [ [ [ "xs_mc_chainmail_hood", 1 ] ], [ [ "xs_mc_chainmail_hauberk", 1 ] ], [ [ "xs_mc_chainmail_legs", 1 ] ] ]
+    "components": [ [ [ "xs_mc_chainmail_hood", 1 ] ], [ [ "mc_chainmail_hauberk_xs", 1 ] ], [ [ "xs_mc_chainmail_legs", 1 ] ] ]
   },
   {
     "result": "hc_chainmail_suit",
@@ -1151,7 +1151,7 @@
       { "proficiency": "prof_closures", "fail_multiplier": 1 },
       { "proficiency": "prof_chain_armour", "fail_multiplier": 1 }
     ],
-    "components": [ [ [ "xl_hc_chainmail_hood", 1 ] ], [ [ "xl_hc_chainmail_hauberk", 1 ] ], [ [ "xl_hc_chainmail_legs", 1 ] ] ]
+    "components": [ [ [ "xl_hc_chainmail_hood", 1 ] ], [ [ "xl_lc_chainmail_hauberk", 1 ] ], [ [ "xl_hc_chainmail_legs", 1 ] ] ]
   },
   {
     "result": "hc_chainmail_suit_xs",
@@ -1170,7 +1170,7 @@
       { "proficiency": "prof_closures", "fail_multiplier": 1 },
       { "proficiency": "prof_chain_armour", "fail_multiplier": 1 }
     ],
-    "components": [ [ [ "xs_hc_chainmail_hood", 1 ] ], [ [ "xs_hc_chainmail_hauberk", 1 ] ], [ [ "xs_hc_chainmail_legs", 1 ] ] ]
+    "components": [ [ [ "xs_hc_chainmail_hood", 1 ] ], [ [ "hc_chainmail_hauberk_xs", 1 ] ], [ [ "xs_hc_chainmail_legs", 1 ] ] ]
   },
   {
     "result": "ch_chainmail_suit",
@@ -1271,7 +1271,7 @@
       { "proficiency": "prof_closures", "fail_multiplier": 1 },
       { "proficiency": "prof_chain_armour", "fail_multiplier": 1 }
     ],
-    "components": [ [ [ "xs_ch_chainmail_hood", 1 ] ], [ [ "xs_ch_chainmail_hauberk", 1 ] ], [ [ "xs_ch_chainmail_legs", 1 ] ] ]
+    "components": [ [ [ "xs_ch_chainmail_hood", 1 ] ], [ [ "ch_chainmail_hauberk_xs", 1 ] ], [ [ "xs_ch_chainmail_legs", 1 ] ] ]
   },
   {
     "result": "qt_chainmail_suit",
@@ -1372,7 +1372,7 @@
       { "proficiency": "prof_closures", "fail_multiplier": 1 },
       { "proficiency": "prof_chain_armour", "fail_multiplier": 1 }
     ],
-    "components": [ [ [ "xs_qt_chainmail_hood", 1 ] ], [ [ "xs_qt_chainmail_hauberk", 1 ] ], [ [ "xs_qt_chainmail_legs", 1 ] ] ]
+    "components": [ [ [ "xs_qt_chainmail_hood", 1 ] ], [ [ "qt_chainmail_hauberk_xs", 1 ] ], [ [ "xs_qt_chainmail_legs", 1 ] ] ]
   },
   {
     "result": "chainmail_suit_faraday",


data/json/items/armor/suits_protection.json
```diff
@@ -1557,7 +1557,7 @@
     "armor": [
       {
         "material": [ { "type": "budget_steel_chain", "covered_by_mat": 100, "thickness": 1.2 } ],
-        "covers": [ "torso", "leg_l", "leg_r", "arm_l", "arm_r" ],
+        "covers": [ "torso", "arm_l", "arm_r" ],
```
Drew4484, line 1560: Replace existing leg coverage with sublocation coverage for upper thighs and hips.

data/json/items/armor/suits_protection.json
```diff
@@ -1766,7 +1766,7 @@
     "category": "armor",
     "name": { "str": "low steel chainmail armor" },
```
Drew4484, line 1767:
```suggestion
    "name": { "str": "mild steel chainmail armor" },
```

data/json/items/armor/suits_protection.json
```diff
@@ -1766,7 +1766,7 @@
...
     "description": "A fully customized low steel chainmail suit.  Combine with a gambeson or other padded layer for optimal protection.",
```
Drew4484, line 1768:
```suggestion
    "description": "A fully customized mild steel chainmail suit.  Combine with a gambeson or other padded layer for optimal protection.",
```

data/json/items/armor/suits_protection.json
```diff
@@ -1569,7 +1569,7 @@
     "category": "armor",
     "name": { "str": "low steel chainmail hauberk" },
```
Drew4484, line 1570:
```suggestion
    "name": { "str": "mild steel chainmail hauberk" },
```

data/json/items/armor/suits_protection.json
```diff
@@ -1569,7 +1569,7 @@
...
     "description": "A fully customized low steel chainmail outfit, leaving the head uncovered.  Combine with a gambeson or other padded layer for optimal protection.",
```
Drew4484, line 1571:
```suggestion
    "description": "A fully customized mild steel chainmail outfit, leaving the head uncovered.  Combine with a gambeson or other padded layer for optimal protection.",
```

Drew4484:
The hauberk should cover the subparts "leg_upper" and "leg_hip", as they typically extended to mid-thigh (see <https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/ARMOR_BALANCE_AND_DESIGN.md#explanation-3> for more information about covering sublocations). Please let me know if you need assistance with converting to detailed coverage.
The legs should remain on the chainmail suit, there are many examples of full-body chainmail and using the term **Suit** in game refers to something that generally covers the torso and limbs at a minimum. If you want to add variants of the hauberk that includes a hood but no sleeves, that should be a separate item with a distinct name ("Hooded hauberk"?) to avoid confusion.

Drew4484 (changes requested):
The hauberk should cover the subparts "leg_upper" and "leg_hip", as they typically extended to mid-thigh (see <https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/ARMOR_BALANCE_AND_DESIGN.md#explanation-3> for more information about covering sublocations). Please let me know if you need assistance with converting to detailed coverage.

The legs should remain on the chainmail suit, there are many examples of full-body chainmail and using the term **Suit** in game refers to something that generally covers the torso and limbs at a minimum. If you want to add variants of the hauberk that includes a hood but no sleeves, that should be a separate item with a distinct name ("Hooded hauberk"?) to avoid confusion.

Dan1ss1mo:
> The hauberk should cover the subparts "leg_upper" and "leg_hip", as they typically extended to mid-thigh (see https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/ARMOR_BALANCE_AND_DESIGN.md#explanation-3 for more information about covering sublocations). Please let me know if you need assistance with converting to detailed coverage.
> 
> The legs should remain on the chainmail suit, there are many examples of full-body chainmail and using the term **Suit** in game refers to something that generally covers the torso and limbs at a minimum. If you want to add variants of the hauberk that includes a hood but no sleeves, that should be a separate item with a distinct name ("Hooded hauberk"?) to avoid confusion.

Yeah, i know that hauberk should also cover upper legs, but in that case it's recipe should be adjusted to reqire some extra chainmail sheets, also it would conflict with chainmail leggings. It wold also be nessesary to change protection of suit, to have higher thickness on upper legs and hips. In order to avoid all of that i think it would make sense to keep it a bit shorter than historical one, especially if it supposed to be part of full body suit(it seems that hauberks mostly worn on they own). Maybe there is also proper name for such variant but i am not sure. I will revert chainmail suit change though. Also, shouldn't weigh for chain gloves and chausses should be reduced? Currently they weight almost as much as sleeves. Recipe should also probably give some linceges back as byproduct, as you probably won't need 2 full sheets

Dan1ss1mo:
Maybe call it long sleeve Haubergeon

Dan1ss1mo:
It also seems that chain gloves wasn't fully made from chainmail. They commonly looks like leather gloves with chain coverege on outer side

Drew4484:
> > The hauberk should cover the subparts "leg_upper" and "leg_hip", as they typically extended to mid-thigh (see https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/ARMOR_BALANCE_AND_DESIGN.md#explanation-3 for more information about covering sublocations). Please let me know if you need assistance with converting to detailed coverage.
> > 
> > The legs should remain on the chainmail suit, there are many examples of full-body chainmail and using the term **Suit** in game refers to something that generally covers the torso and limbs at a minimum. If you want to add variants of the hauberk that includes a hood but no sleeves, that should be a separate item with a distinct name ("Hooded hauberk"?) to avoid confusion.
> 
> Yeah, i know that hauberk should also cover upper legs, but in that case it's recipe should be adjusted to reqire some extra chainmail sheets, also it would conflict with chainmail leggings. It wold also be nessesary to change protection of suit, to have higher thickness on upper legs and hips. In order to avoid all of that i think it would make sense to keep it a bit shorter than historical one, especially if it supposed to be part of full body suit(it seems that hauberks mostly worn on they own). Maybe there is also proper name for such variant but i am not sure. I will revert chainmail suit change though. Also, shouldn't weigh for chain gloves and chausses should be reduced? Currently they weight almost as much as sleeves. Recipe should also probably give some links back as byproduct, as you probably won't need 2 full sheets.
> Edit: maybe also reduce weight for hauberk a bit as well? According to my research it weight 10 kg, but consider that in-game variant lack leg coverege it should be slightly lighter

We can actually use the layers added in #60862 to avoid conflict for the legs. Alternatively if the hauberk gets trimmed, let's add to the description indicating that it stops at the waist. I'm perfectly okay with updating designs as long as they remain practical. Chainmail is probably too heavy, I only updated the existing suit materials, so they could use an overall review.

Drew4484:
> It also seems that chain gloves wasn't fully made from chainmail. They commonly looks like leather gloves with chain coverege on outer side

There's both but the full chain is usually PPE for working around sharp knives. Adding a version with leather and adjusting the encumbrance would be good. I don't think there's enough of a gameplay difference to worry about fully articulated gloves vs. mittens or half-mitten versions. While these are often integrated into a hauberk I suggest keeping them separate if you're adding a new version of the gloves.

Dan1ss1mo:
> > > The hauberk should cover the subparts "leg_upper" and "leg_hip", as they typically extended to mid-thigh (see https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/ARMOR_BALANCE_AND_DESIGN.md#explanation-3 for more information about covering sublocations). Please let me know if you need assistance with converting to detailed coverage.
> > > The legs should remain on the chainmail suit, there are many examples of full-body chainmail and using the term **Suit** in game refers to something that generally covers the torso and limbs at a minimum. If you want to add variants of the hauberk that includes a hood but no sleeves, that should be a separate item with a distinct name ("Hooded hauberk"?) to avoid confusion.
> > 
> > 
> > Yeah, i know that hauberk should also cover upper legs, but in that case it's recipe should be adjusted to reqire some extra chainmail sheets, also it would conflict with chainmail leggings. It wold also be nessesary to change protection of suit, to have higher thickness on upper legs and hips. In order to avoid all of that i think it would make sense to keep it a bit shorter than historical one, especially if it supposed to be part of full body suit(it seems that hauberks mostly worn on they own). Maybe there is also proper name for such variant but i am not sure. I will revert chainmail suit change though. Also, shouldn't weigh for chain gloves and chausses should be reduced? Currently they weight almost as much as sleeves. Recipe should also probably give some links back as byproduct, as you probably won't need 2 full sheets.
> > Edit: maybe also reduce weight for hauberk a bit as well? According to my research it weight 10 kg, but consider that in-game variant lack leg coverege it should be slightly lighter
> 
> We can actually use the layers added in #60862 to avoid conflict for the legs. Alternatively if the hauberk gets trimmed, let's add to the description indicating that it stops at the waist. I'm perfectly okay with updating designs as long as they remain practical. Chainmail is probably too heavy, I only updated the existing suit materials, so they could use an overall review.

Didn't know about this feature, that's cool! I will update protection for chain suits as well then. That's nice, considering that most leg guards don't cover hips.

Dan1ss1mo:
@Drew4484 what about only keeping suit recipe that reqire chain shirt\Haubergeon(still undesided about ulternative mame)

Drew4484:
> @Drew4484 what about only keeping suit recipe that reqire chain shirt\Haubergeon(still undesided about alternative name)
> 

You can add as many combinations as you want, honestly. As long as they're clearly defined for players so they know what they're crafting. We can also have as many alternate recipes as are needed, as long as the result is the same you don't need to worry about menu bloat.

Dan1ss1mo:
> > @Drew4484 what about only keeping suit recipe that reqire chain shirt\Haubergeon(still undesided about alternative name)
> 
> You can add as many combinations as you want, honestly. As long as they're clearly defined for players so they know what they're crafting. We can also have as many alternate recipes as are needed, as long as the result is the same you don't need to worry about menu bloat.

well, i think that menu bloat is a realy big problem for new players. It is realy hard to navigate through it if you dont already know what you need

Dan1ss1mo:
@Drew4484 i also recalculate armor weight using approximated values from diagram that i show you, could you check if new values seems reasonable to you, please?

Drew4484:
> @Drew4484 i also recalculate armor weight using approximated values from diagram that i show you, could you check if new values seems reasonable to you, please?
> Edit: would be glad if you also help me to figure out what cousing the error

Here's the error, will need to double check that the xs ch chainmail hauberk is listed correctly in the recipe below. I'll have to check the weight later on Sunday.

2022-10-16T10:19:54.0843077Z (~[slow] ~[.],starting_items)=> 10:19:23.036 ERROR : src/requirements.cpp:597 [void component::check_consistency(const std::string &) const] xs_ch_chainmail_hauberk in inline_recipe_ch_chainmail_suit_xs_hauberk is not a valid item template

Dan1ss1mo:
> > @Drew4484 i also recalculate armor weight using approximated values from diagram that i show you, could you check if new values seems reasonable to you, please?
> > Edit: would be glad if you also help me to figure out what cousing the error
> 
> Here's the error, will need to double check that the xs ch chainmail hauberk is listed correctly in the recipe below. I'll have to check the weight later on Sunday.
> 
> 2022-10-16T10:19:54.0843077Z (~[slow] ~[.],starting_items)=> 10:19:23.036 ERROR : src/requirements.cpp:597 [void component::check_consistency(const std::string &) const] xs_ch_chainmail_hauberk in inline_recipe_ch_chainmail_suit_xs_hauberk is not a valid item template

i don't know who the fuck has put xs at the end of hauberk and suit id, instead of at the beggining like with all other such items, but that was REALLY confusing. That was what one of the cause of the problem

Drew4484 (approved):

PR #61572, opened 2022-10-09
Increase shelf life of fruit jam

merrygin:
#### Summary
Balance "Increases shelf life of fruit jam"

#### Purpose of change
Fruit jam is nominally spoiling after 10 days. In reality, shelf life is several months even with opened jams. Jam is a way to preserve fruits through the added sugar, so canning should not be required for it to have use (while canning is of course better).

#### Describe the solution
Increase shelf life of fruit jam to 180 days (roughly half a year/6 months - which is the lower end of numbers i found).

None.

#### Testing
None.

#### Additional context
Some sources
https://www.chefsresource.com/does-jam-go-bad/
http://shelflifeadvice.com/condiments-herbs-spices-spreads/spreads/jelly-jam-and-preserves
http://shelflifeadvice.com/content/how-long-should-i-keep-open-jar-jelly-jam-or-preserves

Jummy.

### DIFF
--- a/data/json/items/comestibles/fruit_dishes.json
+++ b/data/json/items/comestibles/fruit_dishes.json
@@ -252,7 +252,7 @@
     ],
     "weight": "15 g",
     "color": "red",
-    "spoils_in": "10 days",
+    "spoils_in": "180 days",
     "container": "jar_3l_glass_sealed",
     "comestible_type": "DRINK",
     "symbol": "~",


Zireael07:
Quote from your source: 
> Once opened, the jar should be stored in the fridge.

This is the deal-breaker, as Cataclysm doesn't have power anymore so even storing in a fridge will not help.

My mum makes lots of jam during the summer, and she pasteurizes the jars. If a jar doesn't seal, we have to eat it fast or it spoils. 10 days sounds roughly right.

merrygin:
> Quote from your source:
> 
> > Once opened, the jar should be stored in the fridge.
> 
> This is the deal-breaker, as Cataclysm doesn't have power anymore so even storing in a fridge will not help.
> 
> My mum makes lots of jam during the summer, and she pasteurizes the jars. If a jar doesn't seal, we have to eat it fast or it spoils. 10 days sounds roughly right.

Hmm I see. So the "spoils_in" is always assumed to be unrefridgerated?
With the value liek it is, it will also spoil in roughly 10 days even when refrigerated though, doesn't it? So perhaps at least a shelf life of 3-4 weeks or something instead of 10 days?

Maleclypse:
> > Quote from your source:
> > > Once opened, the jar should be stored in the fridge.
> > 
> > 
> > This is the deal-breaker, as Cataclysm doesn't have power anymore so even storing in a fridge will not help.
> > My mum makes lots of jam during the summer, and she pasteurizes the jars. If a jar doesn't seal, we have to eat it fast or it spoils. 10 days sounds roughly right.
> 
> Hmm I see. So the "spoils_in" is always assumed to be unrefridgerated? With the value liek it is, it will also spoil in roughly 10 days even when refrigerated though, doesn't it? So perhaps at least a shelf life of 3-4 weeks or something instead of 10 days?

Refrigerated it will spoil in 30 days after opening.

merrygin:
Alright! Then I'll close the PR.

PR #61574, opened 2022-10-09
Add Hub-01 modular headgear

NoddingCrow:
#### Summary
Content "New Hub-01 modular headgear and ablative inserts"

#### Purpose of change
Hub-01 has long been without a head version of its signature upgradable armor. This adds a new high-tech headrig with both NV and IR cycle goggles, and the full progression of inserts. Additionally a new half-face gas mask providing full protection, currently limited only to the HUB-01 vendor.
This also fills the fun niche of an OUTER layer helmet suitable for wearing over the HUB Envirosuit.

#### Describe the solution
After consulting @John-Candlebury on basic design premise, I settled on simply expanding what was already present in the HUB design aesthetic. The new headgear carries the same armor stats as the poncho, though a little heavier than it proportionally should be on account of the NV/IR scanner. The inserts follow exactly the same progression as the existing poncho ones, with coverage 85% (same as the army helmet). This leads to the Soldier-tier insert being quite a bit stronger than the army helmet, but also considerably heavier so I think it works out well enough. The Kevlar-tier insert is more or less on par with existing army helmet, though probably harder to come across in comparison.

I then added the various new parts to the appropriate vendor and item groups - for example Cranberry the NPC now spawns with a headrig with a kevlar insert to protect their noggin.

In an ideal world I'd love to see a multi-option item transform menu to allow the user to select which mode they wanted, rather than cycle through. This would allow e.g. wearing the goggles on a 'passive' mode with no scanner, and adding basic HUD elements like time and temperature. Unfortunately as far as I can tell there's no way to achieve this, so I settled on the current cycle of Stowed -> NV -> IR -> Stowed

#### Testing
basic build test -- json error resolved
headgear w/ prototype insert spawns on dead scientist 
tier 2 inserts visible in HUB store
new gasmask given correctly when purchasing environment kit before Light Retrieval 
cranberry foster spawns with headgear > kevlar insert 
tier 3 insert given after Light Retrieval Prototype

#### Additional context
image
Mockup MSX+ sprites showing general layout - Stowed goggles on forehead -> NV mode -> IR mode, plus half-face mask


data/json/items/armor/robofac_armor.json
```diff
@@ -243,6 +243,132 @@
...
     "flags": [ "DIMENSIONAL_ANCHOR", "PORTAL_PROOF", "STURDY", "OUTER", "WATERPROOF", "VARSIZE", "RAINPROOF", "HOOD", "SOFT" ]
   },
+  {
+    "id": "robofac_head_rig",
+    "repairs_like": "coat_rain",
+    "type": "TOOL_ARMOR",
+    "category": "armor",
+    "name": { "str": "Hub 01 modular recon headgear" },
+    "looks_like": "tac_helmet",
+    "description": "Heavy duty hooded headwear for the true frontier explorer, made with robust nylon sheeting reinforced with kevlar fibers. The top consists of a series of plastic plates with hooks and velcro strips, perhaps for attaching additional parts. On the forehead of the assembly is a bulky box visor on a mechanism to quickly pull it down over your eyes, with a cover on the side to insert standard batteries. Activate to lower the visor and use the light amplification functions.",
```
TheShadowFerret, line 253:
```suggestion
    "description": "Heavy duty hooded headwear for the true frontier explorer, made with robust Nylon sheeting reinforced with Kevlar fibers. The top consists of a series of plastic plates with hooks and velcro strips, perhaps for attaching additional parts. On the forehead of the assembly is a bulky box visor on a mechanism to quickly pull it down over your eyes, with a cover on the side to insert standard batteries. Activate to lower the visor and use the light amplification functions.",
```
*Angry Dupont noises*

John-Candlebury:
Quite like whats here. 
The only note Is that I probably wouldnt spawn it on the first mission corpse. Might be better to just have it offered for sale after the second mission or so.

NoddingCrow:
> Quite like whats here. The only note Is that I probably wouldnt spawn it on the first mission corpse. Might be better to just have it offered for sale after the second mission or so.

You're the boss! Headgear with proto insert item group deleted; removed from spawning on early corpse. 
Re-booted up a world and tested, headgear is not spawning on corpse but still pops up in the trader inventory eventually. 

Briefly looked at making it a prototype that the intercom will offer you, but good lord I don't get the NPC trees. If anyone else wants to give that a whirl they're more than welcome. 

Still, it's now gated behind getting the HUB to trade with you, making some credit to trade with, then finally the trade RNG. 
Or killing Cranberry if you're a heartless monster, I guess.

John-Candlebury:
Yeah I think that works well for now.

Maddremor:
Would it be possible for the goggles to be an insert themselves? That way you could both hand them out as loot progression at a chosen point, and have the player choose what they actually care to carry with them.

NoddingCrow:
> Would it be possible for the goggles to be an insert themselves? That way you could both hand them out as loot progression at a chosen point, and have the player choose what they actually care to carry with them.

Hiya. Technically possible, absolutely yes. I've just run up a version of it locally and it's working fine. 
However, the helmet with integrated visor was part of the original design inspiration for this work and so was what I based my design around.
I've run both options by Candlebury just now, and he's said to leave it as-is.

It's a fairly easy change to make, and I'll certainly keep the separate headgear/goggles version local for a while in case he changes his mind, but for now I'm going to say it being a little bulky / heavier than maybe people would like is by design.

PR #61578, opened 2022-10-09
Implement scrolling for world menu

### LINKED ISSUE #61369: Load/World menu stops displaying
### Describe the bug

It seems that when the number of worlds to be displayed is greater that the length of one page (max seems to be 18 entries), the world/load selection UI stops displaying. Note that the "New World" option also counts as an item, so when exactly 18 worlds are present, the "load" option will work, but "world" won't.
Do note that this just prevents the UI from displaying, you can still move around and select an option in the menu just fine.

### Steps to reproduce

1.) Create 17 new worlds(likely includes tutorial world), note that the UI for [LOAD] and [WORLD] still works.
2.) Create one more world, and observe that the [LOAD] menu works while the [WORLD] menu has stopped working.
3.) Create another world (you can still interact with the world menu, move and select options until world creation comes up), and observe that the [LOAD] menu stops working too.

### Expected behavior

It should be displaying the various pages etc, not failing to render the list of worlds entirely.

### Screenshots

_No response_

- OS: Android
    - OS Version: Manufacturer: Google; Model: Pixel 4a (5G); Release: 13; Incremental: 8927612; 
- Game Version: 07a9641 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

- I first noticed the bug after accidentally starting the tutorial and quitting while it was loading, there's a chance it broke something. But if it didn't, that would also mean tutorial saves count towards the above world count, which shouldn't be the case anyway.
- This might also mean that creating a tutorial world when the page is full could have caused this instead, but I have my doubts. Test that if the above steps don't work though.
- I have yet to test this in a more isolated case (I'm nervous about messing up my local version, and can't use an Android emulator right now), let me know if you don't experience this.
- The bug first occurred in an earlier version of CDDA Experimental, but after updating to the latest version the UI still wasn't displaying. It only partially resolved after I deleted empty worlds from my save folder (including the tutorial world, but I still had to delete another empty save after it).

Ezrashaw:
#### Summary
Bugfixes "Implement scrolling for world menu (and other main menu sub-menus)"

#### Purpose of change
Fixes #61369. Small terminal sizes, coupled with lots of worlds causes the whole world menu to disappear. This occurs at the smallest terminal size (24) with 17 worlds; this is a _very_ rare edge case (#61369 was playing on android). 

#### Describe the solution
Implemented scrolling for all the sub-menus including for `[World]` and `[Load]`. Note that existing behavior (and visuals) for people on standard screens/ terminals is unchanged.  We use a extra variable to keep track of where the top of the menu is, relative to the scrolling: `sub_opt_off`. I'm not sure if this could have been better done with existing scrolling architectures and overall `main_menu.cpp` needs a refactor to unify all the different implementations of scrolling, menu drawing etc.

- Try and understand a way to do this with other scrolling logic in `main_menu.cpp`: Too difficult, other code solves a slightly different problem and this is best addressed by a proper refactor.

#### Testing
Opened game, created excessive number of worlds, all scrollbars and logic is working as intended. Also successfully tested resizing the terminal with the menu open.

#### Additional context
I'm reasonably new to this codebase and might not have done this in the optimal way.

### DIFF
--- a/src/main_menu.cpp
+++ b/src/main_menu.cpp
@@ -236,18 +236,38 @@ void main_menu::display_sub_menu( int sel, const point &bottom_left, int sel_lin
         return;
     }
 
-    const point top_left( bottom_left + point( 0, -( sub_opts.size() + 1 ) ) );
-    catacurses::window w_sub = catacurses::newwin( sub_opts.size() + 2, xlen + 4, top_left );
+    point top_left( bottom_left + point( 0, -( sub_opts.size() + 1 ) ) );
+    int height = sub_opts.size();
+    if( top_left.y < 0 ) {
+        height -= abs( top_left.y ); // abs unnecessary but more clear
+        top_left.y = 0;
+    } else {
+        sub_opt_off = 0;
+    }
+
+    if( sel2 - 1 < sub_opt_off ) {
+        sub_opt_off = sel2;
+    } else if( sel2 + 1 > sub_opt_off + height ) {
+        sub_opt_off = sel2 - height + 1;
+    }
+
+    catacurses::window w_sub = catacurses::newwin( height + 2, xlen + 4, top_left );
     werase( w_sub );
     draw_border( w_sub, c_white );
-    for( int y = 0; static_cast<size_t>( y ) < sub_opts.size(); y++ ) {
-        std::string opt = ( sel2 == y ? "» " : "  " ) + sub_opts[y];
+
+    for( int y = 0; y < height; y++ ) {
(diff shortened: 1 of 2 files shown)


Ezrashaw:
Note that the lgtm alert was fixed by [a1b7986](https://github.com/CleverRaven/Cataclysm-DDA/pull/61578/commits/a1b7986a3b1cbb18ecdc42603846dace10719221)

PR #61582, opened 2022-10-10
Cinema polishing

NetSysFire:
#### Summary
None

#### Purpose of change

I noticed that restrooms in the cinema simply had no door. Especially the mens restroom was inappropriately exposed this way.
When I looked into this, I noticed that the stalls also all had metal doors and the chairs you sit on while watching a movie are uncomfortable wooden ones.

#### Describe the solution

Gave the restrooms a wall. Replaced `+` with normal wooden doors, made `D` pickable metal doors and `d` locked wooden doors. I know that some of them are supposed to be fire exits, but we do not have one way doors yet.
I also replaced the normal wooden chairs with armchairs. All cinemas have these characteristic seats and some have a premium row where you would have the best view. I implemented that by replacing the armchairs with sofas in that row.

image

Make some doors alarmed. Especially the one-way doors would be alarmed in real life, to deter people using them as shortcuts.

#### Testing

Applied changes locally.


NetSysFire:
I think the entire building can also use more linoleum.

Maleclypse (approved):
Great adjustments!

PR #61595, opened 2022-10-10
Smart controller one engine support

### LINKED ISSUE #52009: Smart Engine Controller shuts down with single engine
### Describe the bug

I'm unsure if this is more appropriate as a bug or feature request since the issue is borderline a bug via it's text and borderline a feature request since it seems like it might be asking to alter the item to add a feature.

The Smart Engine Controller lists as one of it's uses cases a Stationary Generator. However, in the case of a single engine vehicle it will shut itself down and refuse to function at all.

This feels like a bug, since when the vehicle isn't moving and you aren't in control of it, it functions as a stationary generator. If I'm running a high power operation like an autoclave and I don't have a large amount of battery capacity, then I would expect that if I activated it while not driving and left it to it's devices, it'd work to keep the battery level topped off.

Instead, like I mentioned it shuts itself down.

### Steps To Reproduce

Add a Smart Engine Controller to a single engine vehicle.
Activate it while stationary and without taking control of the vehicle.
Wait
SEC will shut itself off complaining that it is attached to a single engine vehicle and that it's intended for multiple engines.

### Expected behavior

I would expect that as long as you don't take control of the vehicle, that it would work to maintain the battery charge level.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: 0.F-2 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

kebkaldanil:
#### Summary
Bugfixes "Smart controller one engine support"

#### Purpose of change
fixes #52009

#### Describe the solution

Now when there is only combustion engines in vehicle smart controller will turn on/off engine when necessary

Previous version added same logic, but it was optional and could be enabled in the smart controller configuration

#### Testing

Manually tested by adding/removing engines

#### Additional context

Added smart_controller_ui_selection enum
Added new option in general section "Force smart engine controller off". If enabled, turn off the smart engine controller when you turn off the engine of the car without an electric motor. Has option "ask"


kebkaldanil:
> "new logic" is very undescriptive, how are players expected to understand what the toggle does without diving into code?
> 
> I'm not sure why you need the toggle, eyeballing it majority of the changes are to make the toggle and a smaller part is the actual fix; IMO if the fix works then it should be the "only" correct logic

Thanks for your comment. Updated

src/vehicle.cpp
```diff
@@ -1463,6 +1463,12 @@ int vehicle::install_part( const point &dp, const vehicle_part &new_part )
     bool enable = false;
     if( new_part.is_engine() ) {
         enable = true;
+        // if smart conntroller is enabled there is charge in battery, no need to test it
```
Fris0uman, line 1466:
```suggestion
        // if smart controller is enabled there is charge in battery, no need to test it
```

kebkaldanil: thanks

PR #61601, opened 2022-10-10
Fix MA mod / loading gzipped json with json_loader

akrieger:
#### Summary
None

#### Purpose of change
`flexbuffer_cache` was naively reading json to memory but CDDA supports also loading gzipped json. gzipped json has to get decompressed to be parsed. `flexbuffer_cache` didn't do that so didn't support gzipped json so broke the MA mod.

#### Describe the solution
Use apis that handle loading and automatically decompressing gzipped json in `flexbuffer_cache`. The runtime is the same for non-gzipped json but incurs an extra in memory copy of the file because of the istream interface.

Copypasting the gzip handling into `flexbuffer_cache` or equivalent to avoid the extra copy. Would take longer than I had time to do and I'd probably introduce a bug in the process.

#### Testing
Created a world with MA mod enabled, made a random start, game loaded.
image

### DIFF
--- a/src/cata_utility.cpp
+++ b/src/cata_utility.cpp
@@ -353,12 +353,90 @@ std::istream &safe_getline( std::istream &ins, std::string &str )
     }
 }
 
+namespace
+{
+
+std::string read_compressed_file_to_string( std::istream &fin )
+{
+    std::string outstring;
+
+    std::ostringstream deflated_contents_stream;
+    std::string str;
+
+    deflated_contents_stream << fin.rdbuf();
+    str = deflated_contents_stream.str();
+
+    z_stream zs;
+    memset( &zs, 0, sizeof( zs ) );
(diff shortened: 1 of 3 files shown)


PatrikLundell:
It it's not ready, why not make it a draft in the mean time to ensure it's not merged by mistake?

akrieger:
I didn't realize it wasn't ready until after putting it up, at which point I was already out the door and busy with the rest of my day and couldn't take the time to figure out how to mark as draft from the mobile UI.

akrieger:
I wasn't able to induce an error in the gzipped omap json (not sure how to force any particular omap file to get parsed), but I did verify this doesn't break the loading & error reporting for non-gzipped-json and does successfully load gzipped omaps because the game at least successfully starts in MA mode.

PR #61602, opened 2022-10-10
remove roadside steam engine spawns

taatu:
#### Summary
None

#### Purpose of change
Though fairly rare, the fully functional steam engines found on roadsides are still far more common than I think can be reasonably justified.

All the steam engines can still be found in some other places after this change.

#### Describe the solution
remove them from the roadside spawn group. 

make them about 3-5 orders of magnitude rarer. But that would be more work and have pretty much the same effect as removal


anothersimulacrum (approved):

NetSysFire:
Your PR is fine, but for the next one consider not using the master branch in your fork. It should be kept clean or you may have trouble updating your fork.

PR #61603, opened 2022-10-10
Heavy punching bag construction

ashGlaw:
#### Summary
Content "Heavy punching bag construction"

#### Purpose of change

There had been no way to build your own heavy punching bags. You could drag one home, but that makes no sense either because they're mounted to the ceiling with chains.

#### Describe the solution

I made a new construction recipe, but there were a few more steps involved along the way

I didn't want you to have to sew up a big leather bag and then unsew it whenever you want to deconstruct it. That's silly. So I made a new heavy punching bag sack container. Of course that required that I make a crafting recipe for the sack, too. And I may as well set up a deconstruction for the sack while I'm at it.

So I finished the construction recipe, modified the furniture and what it deconstructed and bashed into. I made sure you could only construct the bags indoors, as it's suspended from the ceiling, and as such you can't grab and drag the bags now either.

I could go farther. There's no makeshift equivalent to the punching bag, and there probably should be. There should also probably be a whole lot of variation in the materials usable for the recipe, but that's beyond the scope of this fairly simple PR.
#### Testing

I spawned in materials and sewed up a heavy punching bag sack. I constructed it into a heavy punching bag. I constructed a second and smashed it to see what was left behind. I deconstructed the first bag. I deconstructed the heavy punching bag sack. Everything worked as it should.

### DIFF
--- a/data/json/items/containers.json
+++ b/data/json/items/containers.json
@@ -291,7 +291,7 @@
     "name": { "str": "heavy punching bag sack" },
     "looks_like": "bag_canvas",
     "description": "A nigh five foot long cylindrical leather sack, a foot in diameter, with attachment points at the top.  With some bags of sand, some chains and a bunch of blankets you could hang it from the ceiling as a heavy punching bag.",
-    "weight": "2500g",
+    "weight": "2400g",
     "volume": "2260 ml",
     "longest_side": "150 cm",
     "price": 0,


Maleclypse (approved):

MNG-cataclysm (approved):

Heitomos:
Suggestion for the makeshift variant - Wooden Beam + Some Cordage + 4 Pillows. Maybe a shovel so you can bury the end in the ground.

PR #61613, opened 2022-10-11
Added batch time savings to tallow/lard.

Daved27hundred:
#### Summary
 
None

#### Purpose of change

Tallow is just fat rendered in a pot and cut up. all you're doing when crafting it in bulk is adding more fat to the pot, that should absolutely save a lot of time. I don't know why we've gone so long without some BTS but here it is.

#### Describe the solution

added batch time factors of 80 and 1 to both tallow and lard, which should apply to the rest of the variants seeing as they copy from it. I've never cooked tallow in real life but I assume that it's fair for something that takes ten minutes to do.


anoobindisguise:
https://heygrillhey.com/beef-tallow/
I used this source for my cooking costs audit - that gives roughly 2 hours for the amount of calories produced by our recipe, but obviously you aren't standing there watching the pot boil; the 10 minutes represents the prep time of cutting it up and whatnot. If we were going to make these batchable the batching would only save time on that 2h of boiling, rather than the 10 minutes of chopping (which scales linearly with the amount of fat you're rendering)

Daved27hundred:
> If we were going to make these batchable the batching would only save time on that 2h of boiling, rather than the 10 minutes of chopping

yeah, I thought that too. it should be lower than eighty, then?

anoobindisguise:
> > If we were going to make these batchable the batching would only save time on that 2h of boiling, rather than the 10 minutes of chopping
> 
> yeah, I thought that too. it should be lower than eighty, then?

Unless we want to model the boiling time in the recipe too, it's honestly probably best to just not make them batch IMO.

Zireael07:
@anoobindisguise: I seem to recall an issue about having unattended crafting modeled at some point, this recipe and boiling would be a great use case

Daved27hundred:
> Unless we want to model the boiling time in the recipe too

I was under the impression that the entire recipe _was_ both boiling the fat and chopping up the tallow.

> I seem to recall an issue about having unattended crafting modeled at some point, this recipe and boiling would be a great use case

should I close this, then? leave it for when unattended crafting becomes a thing?

Daved27hundred:
I'm going to close this PR, I feel like unattended crafts should be touched first.

PR #61619, opened 2022-10-11
Rm42 Knife adjusts

anoobindisguise:
#### Summary
Balance "RM42 Fighting Knife tweaks"

#### Purpose of change
The RM42 fighting knife was a silly weapon made out of ceramic and plastic that didn't make a whole lot of sense.  It could be repaired by soldering the (presumably plastic) grip and it's unclear how a ceramic blade would be especially effective compared to a metal one.

#### Describe the solution
The knife is now made of tempered steel and diamond, and it is effectively a "pre-CVDed" weapon with extremely high damage but cannot be repaired.  It is longer and heavier than before to justify its ability to block.  Thus it should hopefully to a better job following the Rivtech "design philosophy" of "really good, but...".

Remove entirely.
I rewrote its lore snippet, hopefully that looks OK.


anoobindisguise:
hmm, this passed basic build on my fork, not sure what failed it here.

SirPendrak:
What about making it repairable at nanofabricator?

anoobindisguise:
> What about making it repairable at nanofabricator?

So my thinking is that there were two "generations" of knives, the ones that were made the old fashioned way and thus are found in the world, and the ones made by fabricator (doable by survivor) and thus would be repairable just by virtue of having been fabricated.

Rivet-the-Zombie:
Would it even still count as a knife if it's two feet long?

[It was originally intended to be a super-modern version of this.](https://en.m.wikipedia.org/wiki/Fairbairn%E2%80%93Sykes_fighting_knife)

anoobindisguise:
> Would it even still count as a knife if it's two feet long?
> 
> [It was originally intended to be a super-modern version of this.](https://en.m.wikipedia.org/wiki/Fairbairn%E2%80%93Sykes_fighting_knife)

I based it off the baselard which is also 60cm, but that's a good point, I'll lower it a little.

anothersimulacrum (commented):
Please add this to the DPS tests if you are able.

anoobindisguise:
DPS might be too high as is. However since it is "pre-cvd" and tempered steel (1.3x and 1.2x damage multipliers respectively) it has a "base" DPS of 25.2. Willing to lower it if that seems too extreme but that feels about right overall.

nopenoperson:
im a huge fan of making this item actually better than the combat knife.

anothersimulacrum:
I would suggest a "base DPS" of at most 20.

nopenoperson:
if we look at  GAME_BALANCE..md and try to find where this fits in i would assume that this is a lower end scifi weapon. so im thinking high 20s or even low 30 is an appropriate dps

tests/effective_dps_test.cpp
```diff
@@ -527,6 +527,7 @@ static void check_knives( const std::function<Approx( const std::string & )> &ca
         CHECK( calc_expected_dps( "tanto" ) == 15.78 );
         CHECK( calc_expected_dps( "kris" ) == 14.08 );
         CHECK( calc_expected_dps( "knife_rambo" ) == 14.36 );
+        CHECK( calc_expected_dps( "knife_rm42" ) == 39.13 );
```
anothersimulacrum, line 530: Keep the comment around, please.
Doing the math, `39.13 / ( 1.3 * 1.2) ~= 25`, it's still too high.
```suggestion
        CHECK( calc_expected_dps( "knife_rm42" ) == 39.13 );
```

anoobindisguise: I don't know why the DPS test is overreporting it so much, or rather why the test is underreporting the combat knife by comparison. A CVD'd combat knife would have 34 damage, so this shouldn't be weaker than that. I will increase its weight more and see what that does.

nopenoperson: formula for moves per attack = 65 + (4 per each 250ml of volume) + (1 per 60 grams of weapon weight)

anoobindisguise: OK, as is it is slower than the combat knife being longer and heavier, it has +1 to hit due to being SHORT and not HAND length, and its cut damage is exactly 30% higher than a combat knife (CVD property) so why exactly is its DPS over twice as high?

tests/effective_dps_test.cpp
```diff
@@ -527,6 +527,8 @@ static void check_knives( const std::function<Approx( const std::string & )> &ca
         CHECK( calc_expected_dps( "tanto" ) == 15.78 );
         CHECK( calc_expected_dps( "kris" ) == 14.08 );
         CHECK( calc_expected_dps( "knife_rambo" ) == 14.36 );
+        CHECK( calc_expected_dps( "knife_rm42" ) ==
+               40.51 ); //DPS is unusually high but this is a bigger, slower combat knife with 30% more cut damage.
```
anothersimulacrum, line 531: Please include the calculation and the "base" dps too. Also, 39/40 is still too high, it's over 25 before you apply the modifiers you mentioned. Please keep it at a cap of 20.

anoobindisguise: I don't know what to say, its damage is literally a combat knife + CVD coating, *and* it is slower and heavier than a combat knife. I don't understand why its DPS is reporting as over 2x as high as a generic combat knife when it has total only 30% more stab damage and the same bash damage, and is slower.

anothersimulacrum: 26 * 1.2 is 31, not 34, for a start. To-hit increases also have large effects on this metric.

anoobindisguise: well, CVD is 1.3x damage, tempered steel is ~1.2x damage. I'm assuming that the combat knife is made of either tempered or high carbon steel so giving the RM42 knife identical damage to it, then x1.3 = 33.8 cut. I do think the DPS test probably overrates to hit a bit here, because there's no way +1 accuracy is accounting for over 50% more DPS compared to a CVD'd combat knife, or at least that shouldn't ever be the case in gameplay.

anothersimulacrum: > tempered steel is ~1.2x damage. I'm assuming that the combat knife is made of either tempered or high carbon steel so giving the RM42 knife identical damage to it, then x1.3 = 33.8 cut. 

This doesn't make sense to me.

anoobindisguise: why doesn't it make sense to you?

tests/effective_dps_test.cpp
nopenoperson: i think it has to do with piercing damage on low damage rapid strike weapons having trouble with the soldier zombie and the combat knife sits fairly close to the sweetspot to basically "cheese"  that part of the test. low accuracy probably makes it perform poorly vs the smoker as well

tests/effective_dps_test.cpp
anothersimulacrum: You're switching the multiplier.

anoobindisguise: How so? For a given item, I'm assuming tempered steel is 20% stronger than "steel" (generic) as that's how Drew's weapons worked. CVD applies a 30% increase to cut damage. For the combat knife I'm assuming it's already made of good quality steel, so I'm skipping the 20% increase. Then, I'm adding the 30% more damage from the CVD to go from 26 to 34. As it is a bigger, heavier and longer blade, it should probably have more damage on top of that even.

anothersimulacrum: Oh, I see. Very sorry for the delayed reply.
The DIAMOND flags automatically applies the 30% increase, I thought you knew that and were just applying a 20% increase, as though it were made of better material than the other combat knife, but then it has a 30% increase.
By manually applying this 30% increase, you're double-dipping.

anoobindisguise: Oh jeez, I'm so sorry about that. I could have sworn it added the 30% cut damage during the CVD process and the DIAMOND flag was just for the description. I will lower damage to 27, slightly above that of a combat knife, which should hopefully produce a more sane number. thanks for clearing that up!

kevingranade:

https://discourse.cataclysmdda.org/t/why-does-repairing-the-rm42-knife-require-such-insanely-high-skill/28614/2

PR #61621, opened 2022-10-11
Taping up more windows

ashGlaw:
#### Summary
Content "Taping up more windows"

#### Purpose of change

The construction menu artificially limits us from taping up a number of windows craftable in the game. 
#### Describe the solution

I created a new taped furniture and two constructions per type of paned window, single, double, triple, and quadruple, in nomal and reinforced variants.

Finding a way to hardcode windows to let them be taped without having to create so many json entries for each type, but that seems opposite the direction the project has been taking.
#### Testing

I spawned in one of each of the types of windows, a bunch of duct tape, and a pair of scissors for their cutting property. I then used the construction menu to tape up each window. I examined the windows to ensure they seemed correct. I used the construction menu to untape each window. It all seemed to work just fine.

### DIFF
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -6403,7 +6403,7 @@
     "pre_terrain": "t_window_empty",
     "post_terrain": "t_triple_pane_glass"
   },
-   {
+  {
     "type": "construction",
     "id": "constr_triple_pane_glass_curtain",
     "group": "build_triple_glazed_glass_window",


Maleclypse:
"Finding a way to hardcode windows to let them be taped without having to create so many json entries for each type, but that seems opposite the direction the project has been taking."

I would say that if you have a solid way to do that like making a terrain flag that represents `WINDOW` and making it so any terrain or furniture with the flag `WINDOW` can be taped up into an obfuscated version and untaping it reverts it back to it's original window type would be desired.  That would still be json editable but it would also simplify the construction menu, both are results the project wants.

ZhilkinSerg:
You might want to look into using `pre_flags`:

`"pre_flags": [ "WALL", { "flag": "DIGGABLE", "force_terrain": true } ]`

ZhilkinSerg:
It is very likely that this might do the thing, so you won't need multiple taped window terrains at all:

```
  {
    "type": "construction",
    "id": "constr_tape_up_window",
    "group": "tape_up_window",
    "//": "Tape up window",
    "category": "REINFORCE",
    "required_skills": [ [ "fabrication", 0 ] ],
    "time": "2 m",
    "qualities": [ [ { "id": "CUT", "level": 1 } ] ],
    "byproducts": [ { "item": "duct_tape", "charges": [ 15, 25 ] } ],
    "pre_flags": [ "TAPEABLE_WINDOW", { "flag": "TAPED", "force_terrain": true } ]
  },
```

You will also need to add `TAPEABLE_WINDOW` flag to all terrains that can be taped.

PR #61625, opened 2022-10-11
Dark Days of the Dead map updating

NetSysFire:
#### Summary
Mods "[Dark Days of the Dead] Updated maps"

#### Purpose of change

There have been many changes that would also work over there.

#### Describe the solution

I vimdiffed the mods regional settings and the default ones. I applied most of the changes but took special care to avoid applying military and gun stuff as is.

I made gun stores rare on purpose but if they do not fit in, they can be left out, that is no problem at all.

@I-am-Erk and @Hirmuolio, I most definitely need your eyes on this one. While I am still going over this and adjusting stuff, I am unsure how to proceed with stuff like the ~~blobs, mutant animals and more.
I would plan to blacklist some more buildings, mostly the spider and wasp things, but what is your opinion on this? Giant ants, triffids, other mutant animals do not feel like they would fit in.~~
Edit: Shame there is no `conflicts` for modinfo.json. The default mods messed the mapgen up.

I declare that adding new native alberta flora is out of scope for this PR. I already adjusted some things and went over some of the trees but looking over all of them would send me down a different rabbit hole. One thing at a time.

#### Testing

TBD

### DIFF
--- a/data/mods/classic_zombies/exclusions.json
+++ b/data/mods/classic_zombies/exclusions.json
@@ -49,7 +49,7 @@
           "mx_spider": 0,
           "mx_jabberwock": 0,
           "mx_exocrash_1": 0,
-          "mx_exocrash_1": 0
+          "mx_exocrash_2": 0
         }
       },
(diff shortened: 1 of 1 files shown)


I-am-Erk:
I so far don't have any problems with this, posting to unstale it. If you want to solve conflict and remove draft, it could potentially even get in for stable.

PR #61627, opened 2022-10-11
Adjust Obrez

Drew4484:
#### Summary
Content "Adjusts the Obrez"

#### Purpose of change
The Obrez is currently as damaging as a purpose-built mace and retains a grip mount point. 

#### Describe the solution
Reduce the melee damage to 30, putting it bove our makeshift maces (simple and trench mace at 22 damage). A steel mace deals 40, other maces of similar dimensions range from 22-35. 
Reduce the balance. As a firearm, the weapon suffers from balance issues as the amount of material sawn off the front and rear of the weapon are not equal in weight. As a melee weapon, this is exacerbated by most of the weight being below the striking surface.
Reduce the grip quality. If being used as a mace, the user cannot wrap their hand completely around the weapon. The grooves of the existing furniture cannot easily be used in a swinging motion.

Removing the weapon entirely.

#### Testing
Spawned Obrez, observed new stats.

#### Additional context
While weapons featured in other games can have unique appeal, keep in mind that most games do not attempt to simulate things like melee weapon handling by any mechanism other than swing speed. We should apply rigorous comparisons to existing items with similar traits to ensure we are not adding an unrealistic item that outperforms a significant portion of the equipment already added.


MNG-cataclysm (approved):

Rivet-the-Zombie:
Dang! I wonder what is the rationale for that ridiculous melee damage?

data/json/items/gun/artisan_guns.json
```diff
@@ -7,20 +7,21 @@
     "description": "A cut-down, modified Mosin-Nagant.  It has a short barrel, and the stock has been cut down and fitted with a mace head.  The sides of the firearm have been reinforced with metal to withstand a beating.",
```
tenmillimaster, line 7: Please mention new front and rear sights?

data/json/items/gun/artisan_guns.json
```diff
@@ -7,20 +7,21 @@
...
+    "//1": "20cm barrel length, 15cm remaining stock.",
     "price": 23000,
     "price_postapoc": 3000,
-    "//": "should be very inaccurate and hard to aim by comparison.  Setting handling to that of pistols",
+    "//2": "should be very inaccurate and hard to aim by comparison.  Setting handling to that of pistols",
     "handling": 10,
     "relative": {
       "range": -30,
-      "ranged_damage": { "damage_type": "bullet", "amount": -6 },
+      "ranged_damage": { "damage_type": "bullet", "amount": -17.3 },
```
tenmillimaster, line 18: I don't think we do fractional damage.
```suggestion
      "ranged_damage": { "damage_type": "bullet", "amount": -17 },
```

PR #61637, opened 2022-10-12
reduce barter value of fuels

taatu:
#### Summary
Balance "Reduce the barter value of vehicle fuels"

#### Purpose of change
The barter values of several fuels are absurdly high, with a single 10L jerrycan filled with jp8 being worth $500 - the value of several weeks worth of preserved food, or an assault rifle with 800 rounds of some of the most advanced ammo available.

Also a few other adjustments are needed:
- JP8 is twice as valuable as diesel even though they are considered interchangeable in game, and reality is not that different
- avgas is more valuable than gasoline despite there being less potential uses for it, what with planes being out of the reach of most factions

#### Describe the solution
Reduce the prices of gasoline, diesel, biodiesel, jp8 and avgas by roughly 5-10x, landing them all at similar values.

different values

#### Testing
spawned fuels with new values, they seem reasonable to me


anoobindisguise:
Fuel being pretty valuable tracks, though it seems totally reasonable to make JP8 less valuable than diesel (lower energy per mL) and Avgas much less valuable (nearly impossible to use)

taatu:
It's not really about making it not valuable, but currently the fuel you get from 1 gas station (about 100 litres of diesel and gasoline) is enough to buy out the entire stock of the refugee center (except for the currency they printed). - twice over if you ignore the outlier items of welding supplies. It's disproportionately effective compared to pretty much anything else you could sell.

Maleclypse (approved):

PR #61550, opened 2022-10-08
[CR] iOS build

smileynet:
#### Summary
Build "Support for iOS builds using Xcode"

#### Purpose of change

To create a reasonably automated way for people to build any branches, including experimental, to play on iOS devices. Additionally this lays the groundwork for future improvements of the native iOS experience (and possibly integrate some existing android features). So far I have been able to keep all additions within the scope existing libraries and code languages.

#### Describe the solution

The script `ios-xcode.sh` uses Homebrew (which includes Xcode CLI Tools) and Xcodegen to create a new Xcode project that allows users to build an image for use on their own devices using an apple developer account. The biggest hack currently is the manually included versions of static SDL libraries, as adding linked source directories would have added bloat to the repo I didn't feel comfortable introducing on my first PR here. Also, the alternative of writing up directions on how to do it manually would have dramatically increased the complexity of successfully building, and would have been a lot to navigate for people who are not familiar with Xcode.

#### Describe alternatives you've considered
I would eventually like to rework the build system to fit within the existing make/cmake patterns, but at this point it's low on my list relative to working on continuing to improve the in-game iOS experience.

#### Testing

Testing has been done on an iPad Pro 11 (3rd gen). My particular use case is with a hardware keyboard, so while a soft keyboard works, at the moment is it not a great user experience. I did not include a build chain for simulators, but may add that at a later date as well.

#### Additional context

Given that it's my first time contributing here, I'm new to the codebase, and that this change hits some fundamental sections, I am happy to take whatever extra steps/ provide additional explanations are needed to help make this fit cleanly and non-disruptively in to the existing codebase.

PR #61638, opened 2022-10-12
Correctly use TARGET_OS_IPHONE macro

BrettDong:
#### Summary
None

#### Purpose of change
As pointed out in https://github.com/CleverRaven/Cataclysm-DDA/pull/61550#issuecomment-1272750695, current code checks if `TARGET_OS_IPHONE` macro is defined to see if we're compiling for iOS. However the correct way to use the macro is to check its value. `TARGET_OS_IPHONE` is defined to `0` on macOS and `1` on iOS.

#### Describe the solution
`if defined(TARGET_OS_IPHONE)` => `if (defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE  == 1)`

#### Testing
Compiles on macOS. Haven't tested compiling for iOS but it should.

### DIFF
--- a/src/input.cpp
+++ b/src/input.cpp
@@ -88,7 +88,7 @@ bool is_mouse_enabled()
 
 bool is_keycode_mode_supported()
 {
-#if defined(TILES) && !defined(__ANDROID__) && !defined(TARGET_OS_IPHONE)
+#if defined(TILES) && !defined(__ANDROID__) && !(defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE == 1)
     return keycode_mode;
 #else
     return false;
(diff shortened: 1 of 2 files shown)


smileynet:
I would be curious to get your feedback on how you're using this as I look at address related issues and expand functionality. I'm currently looking in to an issue where hardware keyboards don't correctly repeat keypresses.

smileynet:
I saw this PR so it made me think that you had been working on iOS in one form or another already: https://github.com/CleverRaven/Cataclysm-DDA/pull/48075

This is the issue I am currently trying to figure out: https://discourse.cataclysmdda.org/t/ios-functionality-key-repeat-issue/28250

PR #61641, opened 2022-10-12
Bypass move_effect check if force true when moving vertically

### LINKED ISSUE #60896: When flung from a roof fall damage iterates indefitinely
### Describe the bug

When the survivor is flung while downed the impact will repeat indefinitely, leading to an instant kill (or a softlock with Debug Invulnerability) without them moving down a Z-level.

### Steps to reproduce

1. Modify a monster with a smash attack to have a cooldown of 1 turn
2. Move to a roof and spawn the monster 1 tile away
3. Wait for the monster to reach the adjacent tile
4. (Optional, to verify looping) Modify all limb health to an arbitrary large number or add Debug Invulnerability
5. Add `downed` to yourself and wait a turn for the special to fling you off the roof
6. Behold dying (and falling multiple times, if you have a high health) or softlocking the game when invulnerable

### Expected behavior

Falling one story to happen once.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19044.1826 (21H2)
- Game Version: 0.F-11164-g42abc28f21-dirty [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

Not *terribly* common, but still turns a technically-survivable fall into an instakill.

Fris0uman:
#### Summary
None

#### Purpose of change

Fixes #60896 

#### Describe the solution

Falling while downed would get you in an infinite loop because the downed effect never clears and thus you never get past the move_effect check so you never actually move

Bypass the check when `force == true` which is the case when falling and make sense, if we're forcing the movement we don't care about move effects

#### Testing

Teleport in the sky one > fall and get `downed`
Teleport a second time
    > before patch get into infinite loop
    > after patch get to the ground no issue

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -11234,7 +11234,7 @@ void game::vertical_move( int movez, bool force, bool peeking )
         return;
     }
 
-    if( !u.move_effects( false ) ) {
+    if( !u.move_effects( false ) && !force ) {
         u.moves -= 100;
         return;
     }


kevingranade:
I think this makes the situation better, but looking over what happens in move_effect, we need a better solution for canceling effects when they no longer make sense.

Maybe pass an enum saying what kind of movement it is into move_effect() so it can decide whether it should cancel it.

PR #61656, opened 2022-10-13
Update the owner faction of count_by_charges items picked up

### LINKED ISSUE #61642: Inconsistent trading and inventory handling
### Describe the bug

I'm foraging and harvesting stuff, and try to offload the items collected to my companion. The weird think is that not all of my items are shown for trade, although I've so far not detected any case of my companion not showing any items.
I've not set up any kinds of rules for hiding, pocket preferences, etc. so that shouldn't come into play, especially since the hidden object can show up a short while later. It's also not a case of mismatching carrying capacity, since both my character and my companion are restricted to what fits in a straw basket (starting Innawood equipment, with a character made one for the companion).

The other weird thing that may or may not be related is that the carrying capacity varies for no apparent reason. Usually I can carry 3 standard rocks in the basket, but sometimes a fourth one, and most of the time I can carry 4 lumps of clay in it, but sometimes 5 or even 6.

### Steps to reproduce

1. Start with a character and a companion that both are restricted to a straw basket for carrying inventory items. However, my character is usually also wearing a full small water skin that's never been shown to contain any of the carried items.
2. Forage stuff and fill up the inventory.
3. Bring up the trade screen to hand the collected items over to the companion and check to see whether all items collected are available for trade. If they are, retry (e.g. after handing what you can over to the companion, continue to forage, and then return to base to drop off everything before trying again).

### Expected behavior

I should be able to hand everything collected over to a companion provided the companion has exactly the same inventory capability (and isn't incredibly weak, unable to carry the items).

### Screenshots

My inventory:
Screenshot (263)

Trying to offload the items. Note that only the food items are available this time, although that can vary so only some food items are visible, some non food items are available as well, all the way to everything being available:
Screenshot (264)

- OS: Windows
    - OS Version: 10.0.19043.2006 (21H1)
- Game Version: 64e17c6 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Innawood [innawood]
]
+ "mod" to disable portal storms
+ Bare bones basecamp addition of recipes for steel from bog iron and plant thread from plants.

### Additional context

Another weird behavior is that the ability to carry "twohanded" items is inconsistent. The basket usually blocks the ability to carry such items, but occasionally does not.

EIIKaO:
#### Summary
None

#### Purpose of change
- Fix #61642
Update the `count_by_charges` items owners faction when its picked up

#### Describe the solution
Insert `item::handle_pickup_ownership` function into appropriate place.

None

#### Testing
In-game testing done.
- Confirmed that the owner is my faction after picking up supplies found in the bushes.
- Confirmed that when a person picks up from another faction( In short, steals ), a caution for theft is raised, then when the person is caught and persuaded to give it away, the owner faction is restored to the old ones.

#### Additional context
Thanks to PatrikLundell who reported the bug.
PatrikLundell's specific report and attached images, which clearly describe the situation, were very helpful in locating the cause of the bug.

### DIFF
--- a/src/item_pocket.cpp
+++ b/src/item_pocket.cpp
@@ -1884,7 +1884,7 @@ bool item_pocket::can_unload_liquid() const
     return will_spill() || !cts_is_frozen_liquid;
 }
 
-int item_pocket::fill_with( const item &contained, const Character &guy, int amount,
+int item_pocket::fill_with( const item &contained, Character &guy, int amount,
                             bool allow_unseal, bool ignore_settings )
 {
     int num_contained = 0;
@@ -1915,7 +1915,7 @@ int item_pocket::fill_with( const item &contained, const Character &guy, int amo
         return 0;
     }
 
-    contained_item.handle_pickup_ownership( const_cast<Character &>( guy ) );
+    contained_item.handle_pickup_ownership( guy );
     if( !insert_item( contained_item ).success() ) {
         debugmsg( "charges per remaining pocket volume does not fit in that very volume" );
         return 0;
--- a/src/item_pocket.h
+++ b/src/item_pocket.h
@@ -325,7 +325,7 @@ class item_pocket
         void add( const item &it, item **ret = nullptr );
         bool can_unload_liquid() const;
 
-        int fill_with( const item &contained, const Character &guy, int amount = 0,
+        int fill_with( const item &contained, Character &guy, int amount = 0,
                        bool allow_unseal = false, bool ignore_settings = false );
 
         /**


src/item_pocket.cpp
```diff
@@ -1914,10 +1914,13 @@ int item_pocket::fill_with( const item &contained, int amount, bool allow_unseal
     if( contained_item.charges == 0 ) {
         return 0;
     }
+
+    contained_item.handle_pickup_ownership( const_cast<Character &>( guy ) );
```
dseguin, line 1918: Why not just pass `Character &guy` into the function instead of const casting here?

EIIKaO: Thanks for the review.

I may have added const without much thought.
```diff
diff --git a/src/item_pocket.cpp b/src/item_pocket.cpp
index 6c2590b25e..67f878ca6d 100644
--- a/src/item_pocket.cpp
+++ b/src/item_pocket.cpp
@@ -1884,7 +1884,7 @@ bool item_pocket::can_unload_liquid() const
     return will_spill() || !cts_is_frozen_liquid;
 }
 
-int item_pocket::fill_with( const item &contained, const Character &guy, int amount,
+int item_pocket::fill_with( const item &contained, Character &guy, int amount,
                             bool allow_unseal, bool ignore_settings )
 {
     int num_contained = 0;
@@ -1915,7 +1915,7 @@ int item_pocket::fill_with( const item &contained, const Character &guy, int amo
         return 0;
     }
 
-    contained_item.handle_pickup_ownership( const_cast<Character &>( guy ) );
+    contained_item.handle_pickup_ownership( guy );
     if( !insert_item( contained_item ).success() ) {
         debugmsg( "charges per remaining pocket volume does not fit in that very volume" );
         return 0;
diff --git a/src/item_pocket.h b/src/item_pocket.h
index 22e14831f6..a2ece5dc00 100644
--- a/src/item_pocket.h
+++ b/src/item_pocket.h
@@ -325,7 +325,7 @@ class item_pocket
         void add( const item &it, item **ret = nullptr );
         bool can_unload_liquid() const;
 
-        int fill_with( const item &contained, const Character &guy, int amount = 0,
+        int fill_with( const item &contained, Character &guy, int amount = 0,
                        bool allow_unseal = false, bool ignore_settings = false );
 
         /**
```

Is this code better than the old one?

dseguin (approved):

PR #48669, opened 2021-04-26
medical zombies

LyleSY:
#### Summary
Content "Medical zombie loot and medical zombie variants"

#### Purpose of change

Realism, monster diversity

#### Describe the solution

Creates new monsterdrop groups "zombie_medical" and "zombie_medical_clothes" and gives it to some new copy-from medical zombies and spawns those, especially on medical lists. New medical upgrade group passing to zombies with medical or no drops.

#### Describe alternatives you've considered

Bizarre new medical zombie powers and evolutions. Hopefully a future PR.

#### Testing

Game loads no errors, monsters spawn and look right

Screen Shot 2021-06-10 at 10 39 57 PM

Monsters drop appropriate loot

Screen Shot 2021-06-10 at 10 50 50 PM

#### Additional context

Inspired by https://github.com/cataclysmbnteam/Cataclysm-BN/pull/493

PR #47089, opened 2021-01-28
Added Regenerating zombie

Daved27hundred:
#### Summary

Content "Added a new zed that regenerates a tenth of it's HP every turn"

#### Purpose of change

the regenerating zombie's constant health regen would make for an interesting foe, difficult for early game characters and characters unable to deal damage quickly enough, such as with a firearm or late-game melee weapon

#### Describe the solution

zed uses the same line of code that the shoggoth uses to regenerate health, only the value is set to regenerate 12 health instead of the shoggoth's 50 health per turn

#### Testing

currently untested, looking to iron out any problems it may create.

PR #61669, opened 2022-10-13
Fixed regenerating zombie mutating into a crawler + rewrote medical zombie descriptions

Daved27hundred:
#### Summary
Content "Removed regenerating zombie mutating into crawlers + rewrote some medical zombie's descriptions"

#### Purpose of change

The descriptions of the medical zombies from #48669 were very barebones, simply reading variations of "A zombie dressed like a medical professional" followed by a snippet from the description of whatever it's based on. Underwhelming considering they all have unique and creative names, except the regenerating variant (though I've fixed that as well.)

Speaking of, I noticed the regenerating zombie I made in #47089 mutates into a crawler or a pupating crawler, which I don't think is intentional. It could explain why it appears so little ingame.

#### Describe the solution

Rewrote the descriptions to help them stand out, as well as removed the line that evolves the regenerating zombie

In the future I'll add evolved forms of the regenerating zombie.

#### Testing

none done

### DIFF
--- a/data/json/monsters/zed-medical.json
+++ b/data/json/monsters/zed-medical.json
@@ -13,7 +13,7 @@
     "id": "mon_zombie_medical_brute",
     "type": "MONSTER",
     "name": { "str": "medical horror" },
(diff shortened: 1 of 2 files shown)


Maleclypse (approved):
Ok I think this makes sense

PR #61675, opened 2022-10-14
Added mission wrap up for Find Monster goal

### LINKED ISSUE #19220: The "find my dog" mission is being finished incorrectly, the dog won't be given to its owner
In the mission where you must find and retrieve a dog for an NPC the dog will stay with you (friendly) even when the mission is completed and appears in Completed missions tab. Also, if this dog gets killed, the mission will be failed too, making it appear in _both_ Completed and Failed tabs.

### LINKED ISSUE #58556: Bug on death of the mission dog
### Describe the bug

In the first survivor mission, the NPC that appears next to me sends me to look for a dog, and gives me a whistle to call it when it is near. I then return with the dog and talk to the NPC who gave me the mission, it is marked as completed but the dog continues to follow me. 
After some other monster kills it I get the crash. 
Even if I give the sibate back to the NPC the dog continues to follow me.

### Steps to reproduce

1. Obtain the mission 'Find Lost Dog'.
2. Search for the dog and blow the whistle when it is near.
3. Complete the mission by talking to the NPC
4. Go to fight some monster and get the dog killed.

### Expected behavior

When the mission is completed the ownership of the dog should pass to the NPC.

### Screenshots

Screenshot_51

- OS: Windows
    - OS Version: 10.0.19044.1766 (21H2)
- Game Version: 060b19e [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Deadly Zombie Virus [deadly_bites],
    sees-player, run, crouch, bleed icons, retrodays [sees_player_retro]
]

### Additional context

[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/8936209/debug.log)
Save: https://drive.google.com/file/d/1wbFHx34s3mTNB5bj9b-ioXOgwt0w0DaR/view?usp=sharing

captainego:
#### Summary
None

#### Purpose of change
Fixes #19220
Fixes #58556
Previously when completing the "Find my dog" starting NPC mission or the  "Bring me a cat" Isherwood mission the animal would still belong to the user.  

#### Describe the solution
With this fix the animal is no longer owned by or friendly to the character, and the mission id is removed from the animal so that if the animal dies after delivery the mission is no longer marked as both completed and failed.

I considered adding new "npctalk" options.  One option would have been similar to the "u_buy_monster" and another option that would have removed the mission id tag from the monster.

#### Testing
Ran automated tests.

Modify the JSON in data/json/npcs/refugee_center/beggars/BEGGAR_2_Dino_Dave.json so that the mission_offered entity has a value of **MISSION_RESCUE_DOG**

Start the game
Create a new character.
Reveal the map
Find Isherwood farm and teleport to the main house.
Talk to Claire, accept the first mission.
Spawn the necessary items (dandelions)
Talk to Claire, complete the first mission and accept the second mission
Spawn the necessary items (bee balm seeds)
Talk to Claire, complete the second mission and accept the third mission (the cat mission we are interested in)
Spawn a unit of cat food.
Teleport to the mission location
Teleport directly next to a cat
Activate the cat food toward the cat
Pickup the cat in the carrrier
Teleport back to the Isherwood main house
Walk next to Claire, place the cat nearby
Talk to Claire, complete the mission
OBSERVE:  Cat is no longer friendly, mission is marked as completed successfully

Teleport to a Refugee Center
Talk to Dino Dave, accept the first mission (the dialog is awkward, but you can verify that the correct mission is assigned in the mission log)
Teleport to the mission location
Kill any nearby zombies
Spawn a vehicle
Get in the vehicle and drive back to the refugee center
Talk to Dino Dave and complete the mission. 
OBSERVE:  Dog is no longer friendly, mission is marked as completed successfully
Spawn dog food
Activate the dog food toward the dog
Spawn a small rope
Go outside the refugee center
Tie the dog in place
Move away
Spawn a grenade
Increase throwing to 10
Activate the grenade
Throw the grenade at the dog
OBSERVE: Dog dies, mission log does not show that the "Find my dog" mission is failed.

### DIFF
--- a/src/mission.cpp
+++ b/src/mission.cpp
@@ -41,6 +41,7 @@
 #define dbg(x) DebugLog((x),D_GAME) << __FILE__ << ":" << __LINE__ << ": "
 
 static const efftype_id effect_pet( "pet" );
+static const efftype_id effect_run( "run" );
 static const itype_id itype_null( "null" );
 
 static const mission_type_id mission_NULL( "NULL" );
@@ -427,7 +428,9 @@ void mission::wrap_up()
                 if( found_monster != nullptr ) {
                     found_monster->mission_ids.erase( uid );
                     found_monster->remove_effect( effect_pet );
+                    found_monster->remove_effect( effect_run );
                     found_monster->friendly = 0;
+                    found_monster->morale = 100;
                 }
             }
         }
--- a/data/json/monsters/mammal.json
+++ b/data/json/monsters/mammal.json
@@ -394,6 +394,14 @@
     ]
   },
   {
+    "id": "mon_cat_isherwood",
+    "type": "MONSTER",
+    "copy-from": "mon_cat",
+    "description": "A small domesticated predator gone feral in the absence of human stewardship.  Located near the old Isherwood cabin.",
+    "morale": 50,
+    "fear_triggers": [ "SOUND", "FRIEND_ATTACKED" ]
+  },
+  {
     "id": "mon_cat_longhair",
     "type": "MONSTER",
     "copy-from": "mon_cat",
--- a/data/json/npcs/isherwood_farm/NPC_Claire_Isherwood.json
+++ b/data/json/npcs/isherwood_farm/NPC_Claire_Isherwood.json
@@ -352,7 +352,7 @@
         "search_range": 20,
         "min_distance": 8
       },
-      "update_mapgen": { "place_monster": [ { "monster": "mon_cat", "pack_size": [ 3, 6 ], "x": 20, "y": 22, "target": true } ] },
+      "update_mapgen": { "place_monster": [ { "monster": "mon_cat_isherwood", "pack_size": [ 3, 6 ], "x": 20, "y": 22, "target": true } ] },
       "effect": [ { "u_spawn_item": "chicken_cage", "count": 1 } ]
     },
     "end": {


PatrikLundell:
It can be noted that the Isherwood cat mission is almost impossible anyway (unless you teleport) as cats run away from you at a high speed whenever getting remotely close to them. The only way to catch a cat is if you corner it, and the chance of doing that to the specific cat the game wants (without marking it in any visible way) is extremely low. If the cat was spawned inside the building it starts in the vicinity of the mission would be possible.

Note that this is a separate issue from the one solved, though.

Fris0uman:
Can the animal be made friendly to the quest giver instead? Because right now it looks like after finishing the mission the animal will just run off which will look a bit ridiculous.

captainego:
@Fris0uman Based on my reading we (currently) cannot make the animal friendly to the new owner.  Friendly seems to be a weighted boolean (0 is unfriendly, != 0 is friendly) that only applies to the character.  

I briefly considered just making the monster disappear, does that feel any better as an outcome?

@PatrikLundell The dog mission requires a specific dog, but the cat mission will (I believe) work with any cat, not just the ones near the mission marker.  That being said, I agree that trying to catch one of the cats "near the cabin" is nearly impossible without teleporting.

src/mission.cpp
```diff
@@ -416,6 +417,20 @@ void mission::wrap_up()
...
+            Creature *found_creature = g->get_creature_if( [&]( const Creature & critter ) {
+                const monster *const mon_ptr = dynamic_cast<const monster *>( &critter );
+                return mon_ptr && mon_ptr->mission_ids.count( uid );
+            } );
+            if( found_creature != nullptr ) {
+                monster *found_monster = found_creature->as_monster();
+                if( found_monster != nullptr ) {
+                    found_monster->mission_ids.erase( uid );
+                    found_monster->remove_effect( effect_pet );
+                    found_monster->friendly = 0;
```
dseguin, line 430: You can probably make the pet friendly to the quest giver by assigning their faction to the pet. Try something like this:
```suggestion
                    found_monster->friendly = 0;
                    const npc *n = g->find_npc( npc_id );
                    if( n ) {
                        found_monster->faction = n->get_faction()->mon_faction;
                    }
```

PatrikLundell:
I wasn't able to finish the Isherwood cat mission using a cat I managed to corner and capture elsewhere...

captainego:
@PatrikLundell  is correct, you do have to use a cat in the "mission area".  I also think that @PatrikLundell is correct and that changing the starting positions of the cats is a separate PR.

src/mission.cpp
captainego: I added this code in my local branch,  I was able to observe the change in the save file (the faction changes to "human"), but I did not observe any change in the behavior of the animals; the dog still stood around in the lobby of the Refugee Center, and the cat still ran away almost immediately.  

After further investigating the issue seems to be that the cat base monster "cat_mon", has the "fear_triggers" of "PLAYER_CLOSE", which results in the cat losing morale and running away.  I tried increasing the cat's morale to 100 which causes the cat to hang nearby for a few ticks, but it still runs away very quickly.  I then tried removing the "PLAYER_CLOSE" "fear_triggers" and the cat no longer ran away.

I then modified the code provided by @dseguin so that the monster is added to the character's faction "player".  This results in the cat showing as friendly (it has a green dot on the monster icon, and is listed as friendly in the V panel), but the cat still flees from the character.

I think this probably could be resolved by adding a new set of "mon_cat" monster types that do not have the "PLAYER_CLOSE" "fear_triggers", but that seems a bit error prone if someone adds a new "mon_cat" subtype and forgets to add the corresponding type that does have the "PLAYER_CLOSE" "fear_triggers".

I will also spend some time debugging through the monmove.cpp monster::plan code to see if I can see any other options that could be used to change this outcome.

captainego: After reading and debugging and logging the monmove.cpp monster::plan code I am unwilling to change it due to the risk of unintended consequences.

I then pursued the idea of creating a new type of cat.  This actually worked out better than I had hoped.  I was able to add one new cat entry (mon_cat_isherwood), that does not have the "PLAYER_CLOSE" "fear_triggers" and I gave it a higher morale.  The combination of these results in the cat hanging around the Isherwood house after it has been returned to Claire, and it has the side effect of addressing @PatrikLundell's concern about the "catchability" of the cats.  The player can now get closer to the cats before the cats start fleeing, and (if the player runs) a cat can now be caught without needing to teleport.

PR #61684, opened 2022-10-14
Follow-up mutation tweaks

Ilysen:
#### Summary
Balance "Even  M O R E  mutation system adjustments"

#### Purpose of change

#61036 established some positive momentum that I wanna keep rolling. That one focused on consistency, and next on my list of things I wanted to do was transparency and responsiveness - adjusting messages to be more clear, communicating instability in a more straightforward way, and such, as well as fixing some issues that the last PR either missed or introduced.

#### Describe the solution

here's a changelog:
* introduced a `Mutation` distraction type that interrupts long activities when you gain or lose a visible mutation. like other distractions, it can be freely toggled on and off in the distraction manager.
* true random mutations (from radiation, Genetic Chaos, etc) are now actually true random; they don't use the instability mechanic to determine whether they're good or bad, and can instead be literally anything. as a result, they also no longer cause instability, since they're separate from the balance and gameplay of voluntary mutation.
  * since this means that Genetic Downwards Spiral no longer respects instability chance determination, it's gotten its old behavior back (the one that makes all mutations negative or neutral) to replicate the same thing, while keeping its rapid instability reduction as a just-in-case.
* entries in `effects.json` that mentioned things like "frog mutagen" now refer to primers instead, ala "frog primers", to go along with what they are now.
* rewrote the descriptions and names of the effect used to telegraph instability. hopefully they should be clearer!
* rewrote the message displayed when you fail to mutate due to a lack of primer - it now explicitly mentions the cause, since right now it's a bit misleading `(...) You crave a more exotic mutagen.`
* messages displayed before, at the start of, and after a mutation cycle can now be one of several varieties, to spice things up.

the main thing I've thought of is keeping the instability in involuntary mutations, even though the initial inception of the new mutation system did not do so. I implemented it in my initial PR because it seemed more sensible, but I heard some legitimate concerns that swayed me towards reverting that change, since they're effectively different experiences.

#### Testing

I made a debug character and did a few different things -
* used radiation and then Genetic Chaos to test that true random mutations were actually true random and caused no instability, along with some hideous debug code to verify;
* gave myself alpha primer + mutagen to check the new messages and make sure that instability was incrementing correctly;
* purified those mutations to make sure that they'd be removed correctly, and that instability was still incrementing.
* repeated the alpha step but with 2000 instability to make sure that negative mutations were being applied.
* started fresh, and gave myself Genetic Downward Spiral. waited for a while with it on its own, then gave myself beast primer and a buncha mutagen to make sure all mutations from both instances were negative, or if they were positive, because they were prequisites for a negative.

everything seemed to work as intended, at least!

#### Additional context

N/A

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -1905,7 +1905,7 @@
     "name": { "str": "Genetic Downward Spiral" },
     "points": -8,
     "purifiable": false,
-    "description": "The events of the Cataclysm have damaged your DNA beyond repair.  You mutate frequently, and your genetic instability is rapidly climbing.",
+    "description": "The events of the Cataclysm have damaged your DNA beyond repair.  You mutate frequently, all mutations you receive (from any source) are negative, and your instability is rapidly growing out of control.",
     "starting_trait": true,
     "cancels": [ "ROBUST" ],
     "valid": false,
--- a/src/mutation.cpp
+++ b/src/mutation.cpp
@@ -61,6 +61,7 @@ static const mutation_category_id mutation_category_ANY( "ANY" );
 
 static const trait_id trait_BURROW( "BURROW" );
 static const trait_id trait_BURROWLARGE( "BURROWLARGE" );
+static const trait_id trait_CHAOTIC_BAD( "CHAOTIC_BAD" );
 static const trait_id trait_DEBUG_BIONIC_POWER( "DEBUG_BIONIC_POWER" );
 static const trait_id trait_DEBUG_BIONIC_POWERGEN( "DEBUG_BIONIC_POWERGEN" );
 static const trait_id trait_DEX_ALPHA( "DEX_ALPHA" );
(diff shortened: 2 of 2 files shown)


Nebnis:
Really nice to see you back Ava, mutations really needed some care for this game.

Ilysen:
i appreciate it! I'm really nothing special; if I hadn't made changes, someone else would have. I'm just glad to help out if I can.

Dignut:
Oooh! I like this, good work! I love this better than the old system where mutations with robust would basically be all good if you used unstable or chaos, them being totally unbiased is awesome and them not giving instability makes it more than diet downward spiral, makes it something incredibly interesting to play with.

bombasticSlacks (approved):

LiteralMushroom:
**BRING BACK ROBUST GENETICS-GENETIC CHAOS**

PR #61708, opened 2022-10-15
Shivering lower fatigue recovery but only wakes you up if very cold

Fris0uman:
#### Summary
Balance "Shivering lower fatigue recovery but only wakes you up if very cold"

#### Purpose of change

Resurect #60991

Fix https://github.com/CleverRaven/Cataclysm-DDA/issues/53443

#### Describe the solution

- Cold wakes you up if Very Cold and not Dead Tired
- Being Cold on your torso just gives a message about shivering once per hour
- Every main part (Head, Torso, Arm L, Arm R, Leg L and Leg R) being cold lowers fatigue recovery by 5% this also affect `effective_sleep` for the purpose of the health bonus from good night sleep

#### Testing
Get naked in the evac shelter basement
Try to sleep
Manage to sleep despite the shivers
#### Additional context

Does *not* actually adresse #47671 because it doesn't solve the fact that sleeping on a blanket doesn't give any warmth despite telling you it does

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -4714,6 +4714,16 @@ void Character::calc_sleep_recovery_rate( needs_rates &rates ) const
     const effect &sleep = get_effect( effect_sleep );
     static const std::string fatigue_regen_modifier( "fatigue_regen_modifier" );
     rates.recovery = 1.0f + mutation_value( fatigue_regen_modifier );
+
+    // -5% sleep recovery rate for every main part below cold
+    float temp_mod = 0.0f;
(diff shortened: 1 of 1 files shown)


PatrikLundell:
This will be a major improvement. It's incredibly annoying that you can't sleep in a basement because your mouth can't handle the freezing cold down there, but you can't sleep upstairs because it's so incredibly hot you're dying of thirst...

NetSysFire:
I wonder if this also addresses the message spam reported in #53411

PR #61718, opened 2022-10-15
Remove unused gasfilter_l and recipe

ehughsbaird:
#### Summary
Bugfixes "Remove unused chemical mask filter"

#### Purpose of change
This item was added in 651ca19e3c50cd0c19c1575e11265ed0216a4fac, and all uses of it were removed 3 months later in 3a511b3526af0091bd4ab48f90caea6b678cf4a5, yet it has stuck around for 4 years more.

#### Describe the solution
Migrate the item to a gas mask filter, remove and obsolete the item and recipe, as well as the ammotype. 

Adding a use for this item.

### DIFF
--- a/data/json/items/ammo.json
+++ b/data/json/items/ammo.json
@@ -1119,21 +1119,6 @@
   },
   {
     "type": "AMMO",
-    "id": "gasfilter_l",
-    "category": "spare_parts",
-    "price": 2500,
-    "price_postapoc": 1000,
-    "name": { "str": "chemical mask cartridge" },
-    "symbol": "=",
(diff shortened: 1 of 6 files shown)


tenmillimaster:
IMO the large filter should be turned into a CBRN filter, what folks think of when they think of a gas mask. Mostly thinking of NATO standard 40mm gas masks.

The medium one into an commerical/industrial respirator mask, with particulate+chemical filtration, like that of 3M, MSA, or Scott's full/half face respirator+cartridge systems.

And the light ones, into the above, but particulate-only filtration. N95 or P100 filters of the above system.

PR #61726, opened 2022-10-15
Rework and limbify staunching bleeds

Venera3:
#### Summary
Balance "Rework staunching bleeding wounds barehanded"

#### Purpose of change

Make putting pressure onto wounds make a bit more sense, remove reading enough books making your hands emenate hemostatic energies.

#### Describe the solution

- Remove healthcare skill scaling from staunching bleeds
- Include the average of grip and manipulation scores in the calculation
- Add a penalty to staunching on your arms and scale it by the number of arms you have
- Scale the broken arm penalty based on the number of arms you have
- Limit the maximum intensity of bleeding you can staunch with your hands
- Add tourniquets help staunching
- Sprinkled some debugmode messages around

Don't overengineer the arm scaling parts, but the more things that are limb-ready in advance the better.

#### Testing

Got bled a lot with and without the wound care profs, mittens, broken arms and tourniquets, checked the values both in debugmode and through direct debugging.

#### Additional context

Staunch limit calculation:
- Base limit 20 (out of bleeding's 40 max intensity)
- multiplied by the average of grip and manipulation scores (the distinction is a bit academic for now since base hands have the same amount of both and both are enc scaled)
- +5 if you have either level of the wound care prof
- Multiplied by `number of broken arms / number of arms` (meaning no staunching with two broken arms)

Bodypart selection:
- Arms' effective bleed intensity is diveded by `1 / number of arms` to scale it for Limb Stuff
- Tourniquets halve the effective intensity
- The bodypart with the highest *base* intensity bleed will be targeted (meaning an unencumbered survivor can staunch every bleed on their tourniqueted legs)

After the selection the bleeding effect's duration is reduced by `5 + staunch limit`, leading to an effective range of 5 - 30 seconds of bleed duration reduction (reduced from the theoretical limit of 5 - 135 seconds with both profs and 10 HC).

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -11652,7 +11652,7 @@ void Character::pause()
         // Don't warn about encumbrance if your arms are broken
         if( num_broken_arms ) {
             // Handle multiple arms
-            max *= ( 1.0f - num_broken_arms / num_arms );
+            max *= ( 1.0f - num_broken_arms / static_cast<float>( num_arms ) );
             add_msg_debug( debugmode::DF_CHARACTER, "%d out of %d arms broken, staunch limit %d",
                            num_broken_arms, num_arms, max );
             add_msg_player_or_npc( m_warning,
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -2467,9 +2467,7 @@ body_part_set Creature::get_drenching_body_parts( bool upper, bool mid, bool low
 
 int Creature::get_num_body_parts_of_type( body_part_type::type part_type ) const
 {
-    int ret = 0;
-    ret = get_all_body_parts_of_type( part_type ).size();
-    return ret;
+    return static_cast<int>( get_all_body_parts_of_type( part_type ).size() );
 }
 
 int Creature::get_num_broken_body_parts_of_type( body_part_type::type part_type ) const


src/character.cpp
```diff
@@ -11631,53 +11632,76 @@ void Character::pause()
...
     if( !is_armed() && has_effect( effect_bleed ) ) {
+        // Calculate max staunchable bleed level
+        // Top out at 20 intensity for base, unencumbered survivors
+        int max = 20;
+        max *= get_modifier( character_modifier_bleed_staunch_mod );
+        add_msg_debug( debugmode::DF_CHARACTER, "Staunch limit after limb score modifier %d", max );
+
+        // +5 bonus if you know your first aid
+        if( has_proficiency( proficiency_prof_wound_care ) ||
+            has_proficiency( proficiency_prof_wound_care_expert ) ) {
```
TheShadowFerret, line 11645: why is this an OR argument if you inherently must have Wound Care mastered to even get Wound Care Expert 🤔

Venera3: Just a failsafe in case anything (debug commands or EoCs) teach you one skipping the pereqs.

src/creature.cpp
```diff
@@ -2464,6 +2464,25 @@ body_part_set Creature::get_drenching_body_parts( bool upper, bool mid, bool low
     }
     return ret;
 }
+
+int Creature::get_num_body_parts_of_type( body_part_type::type part_type ) const
+{
+    int ret = 0;
+    ret = get_all_body_parts_of_type( part_type ).size();
+    return ret;
```
ehughsbaird, line 2472:
```suggestion
    return static_cast<int>( get_all_body_parts_of_type( part_type ).size() );
```

src/character.cpp
```diff
@@ -11631,53 +11632,76 @@ void Character::pause()
...
+            add_msg_debug( debugmode::DF_CHARACTER, "Wound care proficiency found, new limit %d", max );
+        }
+
+        int num_broken_arms = get_num_broken_body_parts_of_type( body_part_type::type::arm );
+        int num_arms = get_num_body_parts_of_type( body_part_type::type::arm );
+
+        // Don't warn about encumbrance if your arms are broken
+        if( num_broken_arms ) {
+            // Handle multiple arms
+            max *= ( 1.0f - num_broken_arms / num_arms );
```
ehughsbaird, line 11656: This will do integer division, you'll get either 1 or 0.
```suggestion
            max *= ( 1.0f - num_broken_arms / static_cast<float>( num_arms ) );
```

nopenoperson:
completely removing skill from the formula doesnt feel right

Venera3:
Compressing a wound is a straightforward, mechanical thing, knowing biochemistry or the dosage of aspirin doesn't really make you better at it.
To skip a few rounds of discussion this is also part of My Actual Day Job, so what "feels" right is not really a priority.

nopenoperson:
> Compressing a wound is a straightforward, mechanical thing, knowing biochemistry or the dosage of aspirin doesn't really make you better at it. To skip a few rounds of discussion this is also part of My Actual Day Job, so what "feels" right is not really a priority.

you can get to very high levels of health care skill in game without doing any book learning so what is the  health care skill supposed to represent when the description text says "Your skill in effecting medical treatment" ? also having it completely based on proficiencies lets the player only be really good at it or really bad at it with no inbetween which seems wierd for a simple task

nopenoperson:
so im thinking about the way proficiencies and the health care skill work together. wouldnt it make sense to have proficiences that represent specialized medical knowledge and have the skill handle very hands on practical stuff? right now it seems to be the other way around?

Venera3:
Proficiencies are for singular self-contained skill sets, skills are a generic abstraction for "knowledge/ability" that really doesn't fit healthcare - dissection has nothing to do with surgery that has nothing to do with bandaging a wound that has nothing to do with crafting medicine. 

As far as I'm concerned the skill is completely superflous and should move over to partly locked profs/traits.

npostavs:
> * Multiplied by `number of broken arms / number of arms` (meaning no staunching with two broken arms)

This means a 3-armed person with 1 broken arm (i.e., 2 working arms) has a penalty vs a 2-armed person? Shouldn't it be just `number of arms - number of broken arms` (maybe with a cap/diminishing returns for more than 2)?

Venera3:
`intensity /= ( 1.0f - 1.0f / num_arms );` -> for 2 arms intensities on the arm are doubled, 3 arms x1,5 and so on.

npostavs:
> `intensity /= ( 1.0f - 1.0f / num_arms );`

I think that only applies if an arm is bleeding, right? I'm rather talking about
 `max *= ( 1.0f - num_broken_arms / static_cast<float>( num_arms ) );`

So trying to stop a torso bleed with your 2 remaining functioning arms if you happen to have a broken arm growing out of your forehead is penalized (`max * 2/3`) vs the case of trying to stop the bleeding with just a normal set of two arms (`max * 2/2`). Even though both cases have 2 arms available for staunching.

(side note, I'm having trouble following what significance the variable `intensity` is supposed to have (it looks like it's **only** for limb selection?), so maybe I've missed something)

Venera3:
Yes, intensity is for the selection only. 
The extra arm will presumably contribute some grip/manipulation scores that factor into the `max` limit calc, so you're starting at your theoretical maximum and working your way down if your arms are broken - if it's just a normal arm+hand you'll have started at 30 max and get mitigated down to 20. For now it's a bit on the hypothetical side, but Limb Stuff is coming in the next cycle so might as well prepare.

Issue #19976, opened 2017-01-08
Phonebook suggestion


Cynthius:
The phonebook in-game mentions that it is a directory of phone numbers and addresses of the local area. Now, I know it also says it's "pretty useless" now, but the fact that it says it has addresses of local businesses suggests that it could logically be used to reveal locations on the map. I'm not sure how one would work this, whether it would be a single random tile revealed each read, but with limited uses, or what. It could cause problems with realism if the phonebook had too few uses, or didn't reveal very much.

I'm just putting this out there because it seems to be an inconsistency in logic. Whether this is fixed by making the phonebook a useable item or just changing the description to explain why you can't is up to you guys. What do you think?

Totentag:
When you brought up the idea, I was mostly indifferent, but your "random tile" suggestion sounds like a good one. A reasonable sort of short-order balancing to keep it from being quite as great as survivor maps and other map-items. Perhaps to add onto the idea, a relatively brutal short-term morale hit thanks to how incredibly boring a "chapter" would be to read.

sniperm13:
I actually 'read' the phone book in one game i had where i was having trouble finding maps (on .2 item spawn). I had the same logic where it 'might' reveal some tiles but had no effect instead. I think 'logically' it should as a real phone book even has a local streetmap in it.

natsirt721:
I followed a similar line of thinking after reading a phone book in a game of mine, however consider this. Phonebooks don't have maps - it's a list of names, numbers, and addresses. Without prior knowledge of the area the addresses aren't going to mean jack.  There may be a handful of small maps in the businesses section but I doubt the ability of one to glean a significant amount of information from a phonebook.

sniperm13:
True that. But as you say, the business section has a redimentary map, usually the yellow pages. Perhaps, not as efficient as an actual 'map' in game but having either random tile revelation or a small radii reveal would be appropriate, unless it's too much of a hassle to code in. As it stands, its only good for rolling papers for pot as is. XD

Aivean:
Sometimes you have roads revealed on the map (like the way to refugee center). Could this mean that you know addresses along such roads? Maybe it will be sensible to reveal buildings near such roads when reading phonebook?

natsirt721:
[...]

PR #61734, opened 2022-10-16
Changing "phone book" description to clarify (even more) that it does not reveal map information

Luca-spopo:
#### Summary
Interface "Changing phone book description to clarify (even more) that it does not reveal map information"

#### Purpose of change

Relevant conversation: in https://github.com/CleverRaven/Cataclysm-DDA/issues/19976
TL;DR:
1) Phone book had a description that, to same players, implied that it can reveal map area.
2) The solution implemented to fix this problem was to change the description to reflect that the book doesn't do that.
3) The fix didn't work for me - the first time I discovered the item I still thought it reveals map area.

#### Describe the solution

Changed the description a little more, to remove mention of "in the nearby area"

Leaving it as it is

#### Testing

No testing done

### DIFF
--- a/data/json/items/book/misc.json
+++ b/data/json/items/book/misc.json
@@ -685,7 +685,7 @@
     "id": "phonebook",
     "type": "BOOK",
     "name": { "str": "phone book" },
-    "description": "This hefty volume contains the telephone numbers of individuals, businesses, and utilities. Pretty useless during the apocalypse and all.",
+    "description": "This hefty volume contains the telephone numbers of individuals, businesses, and utilities.  Pretty useless during the apocalypse and all.",
     "weight": "854 g",
     "volume": "2500 ml",
     "price": 800,


Leland:
Welcome, first time contributor :) You got a check failure: https://github.com/CleverRaven/Cataclysm-DDA/pull/61734/files

(Two spaces after periods)

Luca-spopo:
Thank you, fixed it now. This is also the first time I've learned that monospace fonts take two spaces after a period by convention, I never knew something like that existed.

Luca-spopo:
A unit test appears to have failed
```
(~[slow] ~[.],starting_items)=> ../tests/health_test.cpp:78: FAILED:
(~[slow] ~[.],starting_items)=>   CHECK( dude.get_daily_health() < init_daily_health )
Error: (~[slow] ~[.],starting_items)=>   0 < 0
```

but I can't imagine how my change could have caused that

PR #61738, opened 2022-10-16
Add Big Sur-style macOS icons

Leland:
#### Summary
Build "Add Big Sur-style macOS icons"

#### Purpose of change
I updated our macOS icons to higher quality resolutions in #23181, back in 2018. Since then, macOS has landed on "[squircle](https://liamrosenfeld.com/posts/apple_icon_quest/)" icons, and leaves vertical whitespace. CDDA's older icon caused it to really sit out of place in the dock as a result.

This PR adds such a backing to the macOS CDDA app icons. The sizes are all generated with our [SVG](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/xdg/org.cataclysmdda.CataclysmDDA.svg), so they're crisp.

#### Describe the solution

N/A

N/A

#### Testing

N/A

#### Additional context

Before:
Screenshot 2022-10-16 at 10 34 12 AM

After:
Screenshot 2022-10-16 at 10 30 40 AM


Leland:
Failing test looks fine...? I guess this is just flaky:


> ([slow] ~starting_items)=>     ??:?
> ([slow] ~starting_items)=>     Catch::RunContext::runTest(Catch::TestCase const&)
> ([slow] ~starting_items)=>     ??:?
> ([slow] ~starting_items)=> Backtrace emission took 2 seconds.
> ([slow] ~starting_items)=> (continued from above) ERROR : src/item.cpp:1500 [ret_val<void> item::put_in(const item&, item_pocket::pocket_type, bool)] tried to put an item (thermos) count (1) in a container (daypack) that cannot contain it: pockets unacceptable because item is too short, item too big, or not enough space27.572 s: shot_features

PR #61743, opened 2022-10-16
[Xedra Evolved] fix Forge Dreamdross

GuardianDll:
#### Summary
None

#### Purpose of change

Forge Dreamdross spell was broken for a while, so even if the school is underdeveloped, it would be better to make it work

#### Describe the solution

fixed the spell component id
added `"max_damage": 1,` to spell, so it will spawn 1 ignot, and not 0
added `final_casting_time` and `final_energy_cost` thingy so it will make sense to level it up (remember - having spell, that has more than 0 levels, but has no progression effect is bad)

#### Testing

Works fine

#### Before
image
#### After
image


Maleclypse (approved):
Thank you for catching these things!

PR #61747, opened 2022-10-16
Remove 'EDIBLE_FROZEN' flag from 'bandages_makeshift' and 'bandages_makeshift_boiled'.

Chewy-Pollywog:
#### Summary
Balance "I removed the flag that makes it so a comestible can freeze from the boiled makeshift bandage and the makeshift bandage."

#### Purpose of change
Makeshift bandages and boiled makeshift bandages are able to be frozen despite containing no liquid and being made of cotton. Boiled makeshift bandages use a liquid in their crafting recipe, but they do not contain liquid. The water is more of a tool than it is a component.

#### Describe the solution
The solution was simply to remove 2 flags.

The only other alternative I could think of would be to leave boiled makeshift bandages as is, or make it so that they're created wet, can be frozen when wet, but can not be frozen once they are dried out. I decided against this because water is notoriously slow to change temperature. Wringing out a wet piece of cloth would be more than enough to ensure that it remains unfrozen by the time it is dry.

#### Testing
I launched the game and ensured the flag was functionally removed without causing a crash.

### DIFF
--- a/data/json/items/comestibles/med.json
+++ b/data/json/items/comestibles/med.json
@@ -187,7 +187,8 @@
     "copy-from": "bandages",
     "description": "A reasonably clean piece of cloth wrapped for use as a simple bandage.  Better than nothing.",
     "price_postapoc": 100,
-    "flags": [ "NO_INGEST", "EDIBLE_FROZEN" ],
+    "flags": [ "NO_INGEST" ],
+    "//": "'Once upon a winter freezing, while I pressed 'E', hypovalemic...Quoth the Makeshift Bandage, (frozen).' -Edgar Allen Pollywog",
     "use_action": { "type": "heal", "bandages_power": 2, "bleed": 10, "move_cost": 6000 }
   },
   {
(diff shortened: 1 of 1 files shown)


GuardianDll:
Does it really fix the issue? Doc says - `EDIBLE_FROZEN - Being frozen doesn't prevent eating it. No morale bonus`, which looks like opposite to what you want to do

PatrikLundell:
My understanding of the flags is the same as that of GuardianDll: Removal of the flags would make the items unusable when it's cold, as the flags make the items usable when "frozen". To solve the underlying issue you'd have to device a way to mark consumables as being unable to be frozen in the first place, and mark each of the items that shouldn't be possible to freeze as such. That's probably more trouble than it's worth.

In addition to this, the testing is flawed: It checks that the flag isn't displayed and the game doesn't crash, but doesn't test that the item can actually still be used when "frozen" (which I believe it would no longer be able to be), i.e. when used in a sub zero environment where it's been left long enough to reach the ambient temperature.

Issue #61257, opened 2022-09-25
Make Isherwoods and other large map specials global unique


Fosheze:
### Is your feature request related to a problem? Please describe.

In my game I've recently been on the hunt for labs and I've noticed that certain map specials take up absurd amounts of the game world. Namely this is the Isherwood farm and the refugee center. 

### Solution you would like.

Make the Isherwood farm special and the refugee center global unique like the aircraft carrier is. Also while it doesn't take up much space this could probably be done with Hub01 as well considering it is currently impossible to do missions for more than one anyways and it doesn't really make sense for more than one to exist.

### Describe alternatives you have considered.

Make all unique NPC/mission related specials global unique.

Ramza13:
So this has come up in the past and the code to do it exists.  The decision from the devs is that we have to wait till next version to do this.  The reason being if these are made unique they are harder to find and other changes would need to be made.

PatrikLundell:
Expanding on @Ramza13's comment (this has been said before, with little to nothing new for those who don't want to read a wall of text):
There are a number of problems with using the current global unique system, with the aircraft carrier being sufficiently niche and optional to warrant that treatment despite the issues.
- It can be hard to find these sites. There are currently bread crumbs to find them, but not yet a sufficient number, and those that do exist currently mostly depend on having stumbled upon other such sites. The refugee center is the hub of this web of references, but only default start starts in the shelter where you can get a reference to the refugee center. More breadcrumbs and a robust way of making it likely you can encounter them is needed before ensuring uniqueness would be a good thing.
- The maps currently have poor connectivity, i.e. large parts of the map can be cut off from vehicle travel because the road network is too sparse and insufficiently interconnected, with large regions of forest, rivers, and lakes causing plains travel to fail as a work around.
- As a result of the above, the often single route to such sites can be blocked by things like wasp nests, triffid groves, fungii, and mi-go. While some or all of these might be possible to overcome by an advanced character, it's not desirable to have most of the content being blocked until you're able to push through these obstacles.
- There is currently no real protection of unique locations from local threats. A fungal or ant area with a center right next to the refugee center will have it overwhelmed and likely get wiped out as soon as you get there (and that's ignoring that these threats will likely also prevent you from getting there anyway, at least before progressing to an advanced character).

Zireael07:
[...]

PR #61749, opened 2022-10-17
Change Isherwood to GLOBALLY_UNIQUE

Perset:
#### Summary
Content "Make Isherwood unique"

#### Purpose of change
It's a really big special, not important enough to appear multiple times. I'd do the center, castle, and HUB too, but this one's bigger than all three and then some so pick your battles.

#### Describe the solution
Slaps the new tag on the special.

Continue scowling at Isherwood after Isherwood.

#### Testing
Talk first, then test maybe.


kevingranade:
See https://github.com/CleverRaven/Cataclysm-DDA/issues/61257#issuecomment-1257266529

PR #61751, opened 2022-10-17
Hash updates to latest for vcpkg

smileynet:
#### Summary
Infrastructure "Update vcpkg hash"

#### Purpose of change
Workflow update needed due to third-party vulnerability patches

#### Describe the solution
This updated the vcpkg hash to the latest correct version.

None

#### Testing
CI Runs


akrieger:
To be specific - this hash doesn't include the patch to zlib. But it includes changes to the zlib portfile in vcpkg which make it download from GitHub, which is still hosting the old versions, instead of zlib.net, which took down or moved the old release which vcpkg was providing.

PR #61774, opened 2022-10-18
Fix elevators in hazardous waste sarcophagi

MNG-cataclysm:
#### Summary
Bugfixes "Align elevators in hazardous waste sarcophagi"

#### Purpose of change
Fix issue #60068, align the elevators in the sarcophagi to allow for their use in the Old Guard mission.

#### Describe the solution
Shift the mapgen placement of the top elevator to match the bottom, and move some crates around to open up the space.

#### Describe alternatives you've considered
I considered moving the bottom to align with the top, but went with the choice listed here due to the extra space on the ground floor.

#### Testing
Set computers to 10, spawned the access code, and re-activated the elevator. I could go to z -2 as intended using it, allowing me to access the bottom floor.

#### Additional context
None

PR #61777, opened 2022-10-18
iexamine: don't allow elevator use from outside

andrei8l:
#### Summary
None

#### Purpose of change
#61774 exposed an oversight in the elevator code: you can use the elevator controls while standing outside the elevator and get teleported to solid walls

#### Describe the solution
Only run the elevator code if the player is standing inside the elevator

N/A

#### Testing
Spawn a `Hazardous Waste Sarcophagus`, reactivate the elevator, and try to use the controls from all adjacent tiles.

#### Additional context
N/A

### DIFF
--- a/src/iexamine.cpp
+++ b/src/iexamine.cpp
@@ -1246,7 +1246,7 @@ void iexamine::elevator( Character &you, const tripoint &examp )
 
     auto const uit = std::find( this_elevator.cbegin(), this_elevator.cend(), you.pos() );
     if( uit == this_elevator.cend() ) {
-        popup( _( "You must stand inside the elevator to use it" ) );
+        popup( _( "You must stand inside the elevator to use it." ) );
         return;
     }
 


src/iexamine.cpp
```diff
@@ -1238,12 +1238,18 @@ void iexamine::elevator( Character &you, const tripoint &examp )
...
-    for( tripoint const &pos : closest_points_first( you.pos(), SEEX - 1 ) ) {
+    for( tripoint const &pos : closest_points_first( examp, SEEX - 1 ) ) {
         if( here.has_flag( ter_furn_flag::TFLAG_ELEVATOR, pos ) ) {
             this_elevator.emplace_back( pos );
         }
     }
 
+    auto const uit = std::find( this_elevator.cbegin(), this_elevator.cend(), you.pos() );
+    if( uit == this_elevator.cend() ) {
+        popup( _( "You must stand inside the elevator to use it" ) );
```
Night-Pryanik, line 1249:
```suggestion
        popup( _( "You must stand inside the elevator to use it." ) );
```

Issue #57521, opened 2022-05-07
TCL drones can scan otherworldly enemies, causing massive lag


EtherealMoon0:
### Describe the bug

A tear in reality spawned flaming eyes and other critters, which wandered into a room with multiple TCL security drones, who scanned them every turn and caused unbearable lag. This is probably only a problem for passive creatures that don't actually attack, like flaming eyes and gracken.

### Steps to reproduce

1. Spawn a few TCL security drones. 
2. Spawn a flaming eye or gracken nearby.
3. Wait for 5 minutes and listen to your CPU struggle.

### Expected behavior

From a performance perspective, drones don't scan zombies, so they probably shouldn't try to scan otherworldly beings. To be fair, seeing them properly call in security on the aliens would be pretty neat.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.1645 (21H2)
- Game Version: 46d0df6 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Crow's Hairstles 2: Bright [hairmod2_bright],
    Crow's Hairstles 2: Ears and Tails [hairmod2_earsAndTails],
    Crow's Hairstles 2: Natural [hairmod2_natural],
    Whaleys Hair & Tattoo - 0.F [WHALEYHAIR1],
    Extra Mutated Scenarios [extra_mut_scens],
    Bionic Slots [cbm_slots],
    Tamable Wildlife [Tamable_Wildlife]
]

Jan-Blasiak:
/confirmed YES!

Fs! Those drones are unbelievably stupid! They cause SO MUCH LAG I lost 2h of gameplay because I had to shut down CDDA because it took 30 min to sleep through 1h of those stupid drones scanning everything few floors below.

Wyrdean:
Drones in general should scan less often as well, they can cause non-insignificant lag even when they're just scanning players

andrei8l:
The lag should be fixed now. Is it really a bug that drones can scan monsters though?

Maleclypse:
[...]

PR #61780, opened 2022-10-18
Prevent security drones from scanning monsters

MNG-cataclysm:
#### Summary
Bugfixes "Lab security drones can no longer scan monsters"

#### Purpose of change
Lab security drones could scan monsters, causing massive lag. This PR fixes that, drones can only scan the player.

#### Describe the solution
I set the "require_targeting_player" value to true in the special attack security drones use to scan.

Not doing this.

#### Testing
Entered a lab and observed the security drones. They didn't scan any monsters, only the character, and alerted security.

#### Additional context
Fixes issue #57521

### DIFF
--- a/data/json/monsters/lab_security_drones.json
+++ b/data/json/monsters/lab_security_drones.json
@@ -59,6 +59,7 @@
         "type": "spell",
         "spell_data": { "id": "green_maintenance_check" },
         "cooldown": 20,
+        "require_targeting_player": true,
         "monster_message": "Remain still.  Scanning for authorization."
       }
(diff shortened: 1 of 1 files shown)


anothersimulacrum:
I am pretty sure that issue ends up with a conclusion that this does not cause performance issues, and I agree it is not a bug that drones can scan monsters.

Venera3:
`mattack_actor::spellcasting_actor` has no argument `require_targeting_player`, so this does literally nothing ( it not complaining about the unknown value is a bug, though).

PR #61782, opened 2022-10-18
greatly reduce glass caltrop crafting time

Jerald-of-Libya:
#### Summary
Balance "Greatly reduce glass caltrop crafting time"

#### Purpose of change
Gluing together 4 shards of glass isn't rocket science. Real superglue (cyanoacrylate) takes like 30seconds for a full cure. Even collagen based glue from animal parts has a typical setting time of 1-2 minutes. Curing time for these glues takes 12+ hours for full strength, but that's irrelevant here as we're working with small pieces and it should be good to go once the glue has set. The darn thing's single use anyway, so there's no need for a full strength cure even from a gameplay standpoint.

I just really want some more sensible traps, and to get rid of those pesky glass shards.

#### Describe the solution
Crafting time down from 25m to 2m 30s.

Separate recipe for bone glue that takes longer, but definitely still much lower than 25m.

#### Testing
N/A

#### Additional context
N/A

### DIFF
--- a/data/json/recipes/recipe_traps.json
+++ b/data/json/recipes/recipe_traps.json
@@ -23,7 +23,7 @@
     "skill_used": "fabrication",
     "skills_required": [ "traps", 1 ],
     "difficulty": 1,
-    "time": "25 m",
+    "time": "2 m 30 s",
     "autolearn": true,
     "qualities": [ { "id": "HAMMER", "level": 1 } ],
     "components": [ [ [ "glass_shard", 4 ] ], [ [ "superglue", 1 ], [ "bone_glue", 1 ] ] ]


Zireael07:
we use a lot of superglue IRL in my household, and I agree with the reasoning

Jerald-of-Libya:
Minor oversight, but since the item is plural (loose glass **caltrops**) and there's a hammer involved, this means the item is actually made up of several caltrops instead of one chunky caltrop made up of 4 whole shards that I had in my head. 🤕 

Instead of thinking of these as actual caltrops, maybe "chunks of glue with tiny shards of glass sticking out" might be a better mental image.

PR #61790, opened 2022-10-18
Yet another attempt to "sane-ify" alloy plating

anoobindisguise:
#### Summary
Features "alloy plating uses integrated armor"

#### Purpose of change
Alloy plating bionics have a shitton of problems and I have unsuccessfully attempted to fix them multiple times now. I'm hoping that after a lot of trying and feedback I've finally found an implementation that doesn't suck.

#### Describe the solution
Alloy is now subdermal instead of dermal. It is plating within your body, so it helps absorb impacts and provides bashing and ballistic resist but nearly no cut resist (since it won't stop you from being clawed/bitten). It no longer conflicts with most mutations, so it can no longer be used as a hedge against becoming a furry while using mutagen. Coverage is not total; there are significant gaps that mean it is not 100% reliable.

This IS the alternative.

#### Additional context
alloy4head
alloy4torso
alloy4arms
alloy4legs1
alloy4legs2


anothersimulacrum:
> so it can no longer be used as a hedge against becoming a furry while using mutagen

This is a separate bug. I can look into fixing it, but I don't think this is an appropriate justification for the change.

anoobindisguise:
> This is a separate bug. I can look into fixing it, but I don't think this is an appropriate justification for the change.

Bionic/mutation CONFLICTS do prevent you from getting the specific mutations, as far as I can tell this is intended but undesirable because we eventually want to do "your mutating body broke your bionic". There are a litany of reasons aside from that to do these reworks though, mutation cheese with them being no longer possible is just one of those.

anothersimulacrum:
I implemented that feature, I am aware. If there are other reasons, please expand on them.

anoobindisguise:
> I implemented that feature, I am aware. If there are other reasons, please expand on them.

1. Alloy plating as is, is mostly incoherent and has no real logic behind it. It's not actually clear what it's even doing as an implant, whether it's supposed to be "robot arms", some sort of hybrid, or something else.

2. Alloy plating cannot be an exoskeleton (covers your human parts) because the lack of ability to sweat or breathe whatsoever being sealed inside metal would kill your skin. There's also no reason for it to be surgically attached as opposed to just worn.

3. Alloy plating cannot replace the skin because again, that means no sweating, and no bleeding, or rather if you started bleeding you would be unable to stop. In addition, full-on robotic arms (apparently) won't be possible without a full conversion because the attachment points would present a massive structural weakness if your torso wasn't also mechanized.

anoobindisguise:
> Shouldn't Bionic Combatives also allow you to block without this? Considering that alloy plating would become much more subtle, i can't see how the lack of it should prevent you from blocking. Honestly i always found this restriction pretty weird

BC is also basically not obtainable in normal gameplay tbf and is dubiously canon at this point.

Dan1ss1mo:
> > Shouldn't Bionic Combatives also allow you to block without this? Considering that alloy plating would become much more subtle, i can't see how the lack of it should prevent you from blocking. Honestly i always found this restriction pretty weird
> 
> BC is also basically not obtainable in normal gameplay tbf and is dubiously canon at this point.

Aren't you get it from cqb cbm?

Dan1ss1mo:
Also, what makes them non canon?

anoobindisguise:
> Also, what makes it non canon?

Lots of stuff, first of all Exodii can't really mess with your brain like that, they don't have the technology to install new memories in you. This is why you can't have a helicopter pilot cbm. It contains tons of martial arts from our world specifically that the Exodii shouldn't have (for example the works of a specific Japanese swordsman). Also, the martial art itself operates on purely fictional principles, causing you to gain defense by blocking like Toad Kung Fu, and the descriptions of the techniques break the fourth wall.

Dan1ss1mo:
> > Also, what makes it non canon?
> 
> Lots of stuff, first of all Exodii can't really mess with your brain like that, they don't have the technology to install new memories in you. This is why you can't have a helicopter pilot cbm. It contains tons of martial arts from our world specifically that the Exodii shouldn't have (for example the works of a specific Japanese swordsman). Also, the martial art itself operates on purely fictional principles, causing you to gain defense by blocking like Toad Kung Fu, and the descriptions of the techniques break the fourth wall.

In that case i really hope it stays this way because finding medieval ma is a pain unless you take them at chargen, and i prefer to level my charecters from the scratch. Considering the defence from blocking, i personally considering it some sort of special evasive move\technique that allow you to move you're weak spot out the harms way without requiring as much effort as actually dodging, somthing like that. I honestly think its pretty hard to represent so mane unique fighting styles with with just setting different numbers, so this "unique" effects make sense for me

Zireael07:
> Alloy plating as is, is mostly incoherent and has no real logic behind it

Dermal or subdermal armor is a sci-fi/cyberpunk staple. It never had any logic behind it other than "looks cool and protects from bullets"

anoobindisguise:
> Dermal or subdermal armor is a sci-fi/cyberpunk staple. It never had any logic behind it other than "looks cool and protects from bullets"
Yeah. the problem is that it isn't clear what it's actually trying to do, and making it like, Adam Jensen arms or whatever won't fly, per my many discussions in the dev discord about this.

Zireael07:
Tbh I don't care whether it's dermal or subdermal, and I don't think it can ever be "sane-ified" to the level the core game seems to expect nowadays, so... to Aftershock maybe?

PR #61791, opened 2022-10-18
Audit CBM prices

### LINKED ISSUE #49956: Postapoc prices need audit
### Describe the bug

Im trading with a free merchant in HUB01, suddenly he sells all CBMs for just 10$. This is extremely low. I've visited this merchant on 0.E-2, CBMs were VERY expensive

Also this merchant buys weapons for decreased value. M4A1 was $2000+, now he only buys it for like 60$

### Steps To Reproduce

I really don't know how to reproduce it

Here is the save file for 0.F stable
[Хатто.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/6821527/default.zip)

Here is save file for 0.E-2 I used to migrate to 0.F:
[0.E-2.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/6821523/0.E-2.zip)

### Expected behavior

Prices should be higher

### Screenshots

image

image

image

- OS: Windows 10 x64
    - OS Version: 10.0 1909
- Game Version: bc28921 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda]
]

### Additional context

Im playing save exported from 0.E-2 stable. I've visited this trader in 0.E-2. Maybe it is important.

RenechCDDA:
#### Summary
Balance "CBMs are now priced according to their relative utility"

#### Purpose of change
Closes #49956

All non-faulty CBMs were priced at $10. No longer.

#### Describe the solution
I followed the suggestions in POSTAPOC_PRICE_GUIDE.md extensively, despite the fact this may be sometimes counterintuitive to the player. Many of the prices are based on their *perceived utility* to someone trying to survive in the cataclysm, not necessarily their perceived value to the average survivor (who's been skilling up, raiding labs, and hoarding like a maniac). I also strictly limited the maximum price to 15000 cents ($150) even in the case of CBMs which essentially "gamebreakers" on their own.

Well, time dilation should probably be *priceless*...

#### Testing
Linted

#### Additional context
Please direct hatemail to my github inbox  👉 👉 

I don't think $150 is an appropriate maximum price point when $20/day is defined as subsistence. You won't spend more than a week's worth of food on the ability to stop time?


Maleclypse:
"I don't think $150 is an appropriate maximum price point when $20/day is defined as subsistence. You won't spend more than a week's worth of food on the ability to stop time?"

I agree with you that in at least some cases the $150 max should be a suggestion not an iron law.

Dan1ss1mo:
the main problem is that bartering system as a whole incentivise you to collect a lot of useless crap, which is not really fun gameplay wise. It would be nice to have more dedicated trade goods/currency that you could actualy find, like denarius in magiclysm

PR #61799, opened 2022-10-19
Aftershock: Remove permanent disabling from EMPs

John-Candlebury:
#### Summary

Mods "Aftershock: Remove permanent disabling from EMPs"

#### Purpose of change

With both electric and EMP attacks being more common and the greater reliance and focus on electric gadgets within the mod, the recently added permanent breakage from EMP effects becomes an extremely punitive measure.  Trashing a sizeable amount of the player inventory in a single attack doesn't mesh well with the stated goals of the mod.

EMP effects will still drain batteries, which poses more than sufficient short term danger as long as the mod is concerned.

#### Describe the solution

Add an external option that governs whether EMP effects will permanently disable items. This is true for vanilla but disabled in Aftershock.

Selectively disabling this effect with item overrides would be a maintenance burden that we don't have the manpower to address. As such I have opted for a simpler and decidedly more permanent solution.

#### Testing

Enable aftershock spawn an electronic item and EMP it. It still works after a battery change.

### DIFF
--- a/src/explosion.cpp
+++ b/src/explosion.cpp
@@ -764,7 +764,8 @@ void emp_blast( const tripoint &p )
                 it->ammo_unset();
             }
             // Render any electronic stuff in player's possession non-functional
-            if( ( it->has_flag( flag_ELECTRONIC ) || it->is_battery() ) && !it->is_broken() && get_option<bool>( "EMP_DISABLE_ELECTRONICS" ) ) {
+            if( ( it->has_flag( flag_ELECTRONIC ) || it->is_battery() ) && !it->is_broken() &&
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):
This makes sense.

Night-Pryanik:
After #61810 EMP blast doesn't drain battery charge from items, so EMP blast becomes almost harmless in the mod.

John-Candlebury:
Yes I know.
Proper fix would be having EMPs provide a self removing fault to electronic items, so they work in under the usual game conventions. 

Still better than the alternative however.

PR #61800, opened 2022-10-19
Declayration

Perset:
#### Summary
Content "Bring clay demanded by Isherwood to reasonable amount"

#### Purpose of change
I may be completely mistaken, or blind to some solution here, but has any contributor on this repo done the 1000 clay quest? Do you think it's acceptable? Plz explain why. Anyway, since 1000 clay appeared to be an utter ass-pull for what the mission yields, I just did a little pulling on my own and came out with 50. Nice, eh?

#### Describe the solution
I highlighted "1000" and typed "50". Without quotes. Three times, iirc.

Just letting this sit even though it's dumb. Or I'm dumb and I don't know how to gather bejeezusing loads of clay. 

Actually, we could change the rewards to three complete exoskeletons, one of each variety, and maybe a smart shot or two...the shots are negotiable, though.

#### Testing
The tests have only been metaphorical, so far.

#### Additional context
Seriously I may have no clue about the clay gathering, I'll close this if MurderUnicorn's experiences aren't representative.

### DIFF
--- a/data/json/npcs/isherwood_farm/NPC_Eddie_Isherwood.json
+++ b/data/json/npcs/isherwood_farm/NPC_Eddie_Isherwood.json
@@ -261,12 +261,12 @@
   {
     "id": "MISSION_ISHERWOOD_EDDIE_3",
     "type": "mission_definition",
-    "name": { "str": "Gather 50 units of clay" },
+    "name": { "str": "Gather 100 units of clay" },
     "goal": "MGOAL_FIND_ITEM",
     "difficulty": 3,
     "value": 20000,
     "item": "clay_lump",
-    "count": 50,
+    "count": 100,
     "end": {
       "opinion": { "trust": 2, "value": 2 },
       "effect": [
@@ -280,7 +280,7 @@
     "has_generic_rewards": false,
     "dialogue": {
       "describe": "I do have some resource gathering I could use help if you have time.",
-      "offer": "Now that the clay kiln is set up, we could use 50 units of clay to get production going.",
+      "offer": "Now that the clay kiln is set up, we could use 100 units of clay to get production going.",
       "accepted": "Thank you.",
       "rejected": "Oh well.  I'll try to find time to get it myself, thanks.",
       "advice": "You can find it along river banks, or maybe look for deposits in the forest.",


PatrikLundell:
I've done the quest some time in the past. 1000 clay is a huge weight/volume, and you can't drive up to the quest giver with a car either, so you have to haul it manually. Digging the clay requires something like 50-100 ½ hour sessions.
I consider exoskeletons trophies to have a higher value than that, but I think the effort should be in the vicinity of the reward, even if you add a generous helping of "good guy" into it.

The value of a cow varies depending on the situation. In a normal game with lots of dairy farms a cow in a location with no road access for a transport vehicle is of a limited value (if I remember it correctly, I just tied the cow down there and left it, already having enough cows in the livestock expansion). If, on the other hand, there would essentially be no other way to get a cow, I'd consider 1000 clay to be well worth it.

A clay canning pot uses 15 clay, while I think a regular clay pot uses 5, to provide a perspective of how much clay you need to do stuff. Bricks, on the other hand, use 1. I'd probably set the demand to 100 rather than 50.

I wonder if this quest has been changed indirectly by some change on what the unit of clay was at some time in the past?

data/json/npcs/isherwood_farm/NPC_Eddie_Isherwood.json
```diff
@@ -261,12 +261,12 @@
   {
     "id": "MISSION_ISHERWOOD_EDDIE_3",
     "type": "mission_definition",
-    "name": { "str": "Gather 1000 units of clay" },
+    "name": { "str": "Gather 50 units of clay" },
```
MNG-cataclysm, line 264:
```suggestion
    "name": { "str": "Gather 100 units of clay" },
```
Applying changes PatrickLundell suggested.

data/json/npcs/isherwood_farm/NPC_Eddie_Isherwood.json
```diff
@@ -261,12 +261,12 @@
...
     "goal": "MGOAL_FIND_ITEM",
     "difficulty": 3,
     "value": 20000,
     "item": "clay_lump",
-    "count": 1000,
+    "count": 50,
```
MNG-cataclysm, line 269:
```suggestion
    "count": 100,
```

data/json/npcs/isherwood_farm/NPC_Eddie_Isherwood.json
```diff
@@ -280,7 +280,7 @@
     "has_generic_rewards": false,
     "dialogue": {
       "describe": "I do have some resource gathering I could use help if you have time.",
-      "offer": "Now that the clay kiln is set up, we could use 1000 units of clay to get production going.",
+      "offer": "Now that the clay kiln is set up, we could use 50 units of clay to get production going.",
```
MNG-cataclysm, line 283:
```suggestion
      "offer": "Now that the clay kiln is set up, we could use 100 units of clay to get production going.",
```

MNG-cataclysm (changes requested):

Zireael07:
> I wonder if this quest has been changed indirectly by some change on what the unit of clay was at some time in the past?

Most likely.

PR #61802, opened 2022-10-19
Add sand patch map extra

NetSysFire:
#### Summary
Content "Add sand patch map extra spawning in fields"

#### Purpose of change

We have clay patches in forests, but no sand patches in open fields. Sand is pretty hard to obtain sometimes and this remedies this. Also realism, sandy patches are real.

image

#### Describe the solution

Standard map extra, uses nested chunks for more randomness in shape.
I adjusted the pond spawn rate in fields a tiny bit so ponds are not as common as these sandy patches.

Doing this in C++ like the clay patch map extra. Adding some dead grass around the edges (grass has trouble growing in sand), but theres no way to replace terrain yet. Would be a nice detail but there is no reasonable way to implement this.

Also potentially a different color for the autonote entry because of accessibility reasons. Yellow and light-gray may be hard to distinguish in certain cases. My eyes are fine but I found it hard to distinguish the sand patch from burned ground on the first glance.

#### Testing

Applied changes locally. Spawned it manually and looked for natural spawns.


NetSysFire:
CC @Light-Wave, I think you might want to have this in Innawood. I am currently unsure if your mod would get this automagically due to region overlay or if the map extra is missing a flag or something

Light-Wave:
I'm fairly certain this will spawn in Innawood as well :+1:
Good content. I once considered adding this myself, but found that clay patches were indeed hardcoded, and didn't manage to make it work with sand. So I scattered the occasional tile of sand among the plains instead.
I will have to consider if I want to remove these single sand tiles once this PR has been merged.

NetSysFire:
I think it might even make sense to eventually unhardcode the clay deposit. The only reason it is in C++ is because of

https://github.com/CleverRaven/Cataclysm-DDA/blob/c837fffea7c3474ecb7908f78c8be034bd615ac4/src/map_extras.cpp#L1709-L1710

I think what I am doing here could be a good enough substitute for this. Of course, we lose the blobs (hail the blob) of clay and instead have a small uniform-ish area of clay, but this is pretty minor. And one can do something similar to `mx_science`, but the complexity needed to have a more faithful cellular automata depiction is something which may not be worth doing.

Fris0uman:
Where do the sand patches come from? Why is there suddenly sand in the middle of a dirt field?

Zireael07:
https://link.springer.com/referenceworkentry/10.1007/978-1-4614-9213-9_324-1 is the best I found - TLDR sand patches start as sand sediment deposits so I assume ancient rivers that have long since dried, or other water bodies (I found references to sand patches in the Celtic Sea, too)

NetSysFire:
The above and I just looked at satellite imagery of New England. It turns out most of the fields in New England are man made, some of them even have suspiciously rectangle-like remnants of structures abandoned long ago. Where there is construction, there will also be sand. Pretty minor source of sand patches. All of this does make way for more erosion though.

In addition, soil naturally contains small amounts of sand (more than clay!). Due to myriads of natural processes, sand (and other types of soil) may dominate small areas. We already have clay patches, clay is also naturally in soil, but due to erosion, sand is much more likely to be found in fields, where there is relatively little plant coverage.

NetSysFire:
@I-am-Erk Any chance this can be merged in time for stable? This should balance a minor aspect of the game.

I-am-Erk:
Fiiiine.

PR #61811, opened 2022-10-19
Makes the NRE recorder EMP-proof and replaceable

Ilysen:
#### Summary
Content "The NRE recorder is now EMP-proof; if it breaks, you can get a replacement from the intercom"

#### Purpose of change

image

The non-chatlog version: the NRE recorder is currently irreplaceable, so if it gets broken, you are softlocked with the mission. You'll have no way to complete or fail the mission, so it's stuck in your log forever.

#### Describe the solution

1. Removes the `ELECTRONIC` flag from the NRE recorder, making it EMP-proof. @John-Candlebury stated that it should be immune, so there we are.
2. Adds several new topics to the intercom's dialogue that allow you to requisition a new recorder. In exchange, you'll need an NRE microphone from a microlab (the intercom will point you in the right direction). This means that, even if the recorder does still manage to get broken, you have a way to get another one.

I'd originally thought to make it purchaseable using HGC as well, but candlebury preferred to just have the microphone instead. It  does make more sense this way; they aren't exactly in production anymore, after all.

#### Testing

I used a debug character and opened the contract dialogue at various stages:
* If I hadn't accepted Hazard Meteorology in the past, the dialogue didn't appear.
* When accepting Hazard Meteorology, the new topic explaining that the recorder can be replaced appeared and worked correctly.
* If I already had a recorder on me (broken ones excluded), the dialogue didn't appear.
* Entering the contracts menu without a recorder showed the new topic. I went through each of the options, then spawned in an NRE microphone to exchange. It was consumed when I handed it over, and a new recorder was spawned (battery not included.)

#### Additional context

cataclysm-tiles_EsYRIvBOrv
cataclysm-tiles_KOoTR8TH1N
cataclysm-tiles_CKJkDagXKv
cataclysm-tiles_2hTARHXxvF

### DIFF
--- a/data/json/npcs/robofac/robofac_intercom/robofac_intercom_missions.json
+++ b/data/json/npcs/robofac/robofac_intercom/robofac_intercom_missions.json
@@ -874,7 +874,7 @@
   {
     "id": "MISSION_ROBOFAC_INTERCOM_PORTAL_SM_1_OFFER_ADVICE_Replacement",
     "type": "talk_topic",
-    "dynamic_line": "The recorder is not irreplaceable, but it is constructed using delicate components that are now in short supply.  In the event that the device is rendered nonfunctional, we are willing to provide a replacement at an equivalent cost.  Specifically, if you can locate an NRE microphone, which can be located in certain laboraties, then we are willing to provide a replacement recorder in exchange for it."
+    "dynamic_line": "The recorder is not irreplaceable, but it is constructed using delicate components that are now in short supply.  In the event that the device is rendered nonfunctional, we are willing to provide a replacement at an equivalent cost.  Specifically, if you can locate an NRE microphone, which can be located in certain laboratories, then we are willing to provide a replacement recorder in exchange for it."
   },
   {
     "id": "MISSION_ROBOFAC_INTERCOM_PORTAL_SM_1_ACCEPTED",
--- a/tools/spell_checker/dictionary.txt
+++ b/tools/spell_checker/dictionary.txt
@@ -1294,6 +1294,7 @@ nodachi
 nodosaurus
 noisiness
 nomex
+nonfunctional
 nonpareils
 nonrevealing
 nonsedated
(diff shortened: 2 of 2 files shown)


data/json/npcs/robofac/robofac_intercom/robofac_intercom_missions.json
```diff
@@ -145,6 +166,34 @@
...
+  },
+  {
+    "id": "MISSION_ROBOFAC_INTERCOM_NRE_RECORDER_REPLACEMENT_Microphone",
+    "type": "talk_topic",
+    "dynamic_line": "The microphone is a static version of the same device used in the recorder, mounted on a tripod.  The glass ampoule is larger, but should look roughly the same in design.  They may be somewhat scarce given the current state of the world, but I would advise you to look inside of research labs.  Subway lines may be a good place to start."
+  },
+  {
+    "id": "MISSION_ROBOFAC_INTERCOM_NRE_RECORDER_REPLACEMENT_Complete",
+    "type": "talk_topic",
+    "dynamic_line": "&You fold up the microphone's legs, stuff it awkwardly into the tray, and slide it shut.  A few moments later, it slides open again, revealing a replacement NRE recorder.\n\n\"Try not to lose this one.\""
```
Fris0uman, line 195:
```suggestion
    "dynamic_line": "You fold up the microphone's legs, stuff it awkwardly into the tray, and slide it shut.  A few moments later, it slides open again, revealing a replacement NRE recorder.\n\n\"Try not to lose this one.\""
```

Ilysen: This is intentional - the & changes the syntax of the message so that it doesn't show the default quotations. You can see a screenshot of this in the OP. If I removed it, it'd display something like:

```
the intercom
"You fold up the microphone's legs..."
```

which is improper, since it's something you're doing, not something the intercom is saying.

PR #61817, opened 2022-10-19
Caviar sandwich + bread material!

nopenoperson:
#### Summary
None

#### Purpose of change
add a recipe that works with caviar but i didnt want to cheat on density test so i added a food related material that isnt 1.1 density

#### Describe the solution
cataclysm bread has 14 charges at 60g per charge. this works out to 300ml per slice of bread if we use 0.2 density for white US style bread.

#### Testing
tried crafting it

### DIFF
--- a/data/json/materials.json
+++ b/data/json/materials.json
@@ -1206,7 +1206,7 @@
   {
     "type": "material",
     "id": "bread",
-    "//": "copied from junk food sans resists,  then i googled density of bread and didnt do any other research, its supposed to be US style white bread",
+    "//": "copied from junk food sans resists,  then I googled density of bread and didn't do any other research, it's supposed to be US style white bread",
     "name": "Bread",
     "density": 0.2,
     "specific_heat_liquid": 3.3,


data/json/materials.json
```diff
@@ -1203,6 +1203,29 @@
     "cut_dmg_verb": "cut",
     "burn_data": [ { "fuel": 1, "smoke": 1, "burn": 1 }, { "fuel": 1, "smoke": 1, "burn": 2 }, { "fuel": 1, "smoke": 1, "burn": 3 } ]
   },
+  {
+    "type": "material",
+    "id": "bread",
+    "//": "copied from junk food sans resists,  then i googled density of bread and didnt do any other research, its supposed to be US style white bread",
```
TheShadowFerret, line 1209:
```suggestion
    "//": "copied from junk food sans resists,  then I googled density of bread and didn't do any other research, it's supposed to be US style white bread",
```

Maleclypse:
Hmmm now I'm wondering if scrambled eggs with roe is already in the game.  I will try to remember to add it if not.

Maleclypse (approved):
Thanks for doing this!

PR #61823, opened 2022-10-20
Don't discharge fake vehicle batteries

### LINKED ISSUE #61787: Appliance power usage when crafting depends on battery size
### Describe the bug

The number of battery charges used by an appliance during crafting seems to vary depending on what battery is attached.
This behaviour is probably masked in many cases by the use of multiple batteries and attached power sources such as solar panels.

### Steps to reproduce

1] The attached save has two rows of test appliances with different batteries and all required materials for crafting
2] Go to each FOODCO kitchen buddy and craft oatmeal (10 charges per crafting screen).  Observe the following:

| Attached battery | Charges used  (1 oatmeal) | 
| ------------- |-------------:|
| Small storage battery | 10 |
| Car battery | 0 |
| Medium storage battery | 10 |
| Large storage battery | 0 |
| Very large storage battery | 0 |

| Attached battery | Charges used  (10 oatmeal) | 
| ------------- |-------------:|
| Small storage battery | 100 |
| Car battery | 35 |
| Medium storage battery | 100 |
| Large storage battery | 5 |
| Very large storage battery | 5 |

3] Go to each welding kit and craft a fuel bunker (200 charges per crafting screen).  Observe the following:

| Attached battery        | Charges used          | 
| ------------- |-------------:|
| Small storage battery | 200 |
| Car battery | 100 |
| Medium storage battery | 200 |
| Large storage battery | 10 |
| Very large storage battery | 10 |

4] Start experimenting with other batch sizes
5] Start experimenting with combinations of batteries
6] Observe that different battery combinations seem to behave differently
7] Get a headache

### Expected behavior

The number of charges used when crafting should be the same no matter what battery combination is provided.
As I understand it, the charges used when crafting proceed in steps of ~5% at a time, which suggests that the first step proceeds correctly for the large/very large storage batteries, but breaks afterwards.  For the car battery, I have no idea.

### Screenshots

_No response_

- OS: Linux
- Game Version: 27c1f39 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

[2022-10-18_Appliance_power_usage.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/9812130/2022-10-18_Appliance_power_usage.zip)

I've done a similar test with a vehicle-mounted welding rig, and in all of the cases I tested, the expected number of charges were used.

RenechCDDA:
#### Summary
Bugfixes "Don't discharge fake vehicle batteries"

#### Purpose of change
Fixes #61787 ?

#### Describe the solution
vehicle::discharge_battery is politely told it cannot drain any part that is_fake.

#### Testing
Compiled locally and repeated the tests from #61787. The results now matched the expected power draw.  I tried a few other batch sizes and had a few interruptions as well (dang swamp wasps) too, even after resuming crafting all results were as expected.

I also confirmed that the crafting menu correctly reports the amount of available charges and gives the normal warning that you don't have enough charges if you, in fact, do not.

#### Additional context
The vehicle code is quite complex and I'm not so skilled, so if vehicle power problems crop up shortly after merging this PR it's pretty safe to blame me!

@ZeroInternalReflection did excellent work mapping out the issue and even finding it in the first place.

### DIFF
--- a/src/veh_appliance.cpp
+++ b/src/veh_appliance.cpp
@@ -60,6 +60,8 @@ void place_appliance( const tripoint &p, const vpart_id &vpart, const cata::opti
         return;
     }
 
+    veh->add_tag( flag_APPLIANCE );
+
     if( base ) {
         item copied = *base;
         veh->install_part( point_zero, vpart, std::move( copied ) );
@@ -67,8 +69,6 @@ void place_appliance( const tripoint &p, const vpart_id &vpart, const cata::opti
         veh->install_part( point_zero, vpart );
     }
     veh->name = vpart->name();
-
-    veh->add_tag( flag_APPLIANCE );
 
     // Update the vehicle cache immediately,
     // or the appliance will be invisible for the first couple of turns.
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -6009,7 +6009,7 @@ void vehicle::refresh( const bool remove_fakes )
     };
     // re-install fake parts - this could be done in a separate function, but we want to
     // guarantee that the fake parts were removed before being added
-    if( remove_fakes && !has_tag( "wreckage" ) ) {
+    if( remove_fakes && !has_tag( "wreckage" ) && !has_tag( "APPLIANCE" ) ) {
         // add all the obstacles first
         for( const std::pair <const point, std::vector<int>> &rp : relative_parts ) {
             add_fake_part( rp.first, "OBSTACLE" );


RenechCDDA:
> I'd consider turning off fake parts entirely for appliances

I would too if I had figured out how to!

>Adding something like && !has_tag( "APPLIANCE" ) to the condition

I had actually looked right at this particular chunk of code but discounted it because of the comment. 

I implemented your suggestion on a local test branch (without the above PR's code change) to see how it would do. It does appear* to correct the power draw. I also spent a bit of time throwing in debug messages to see exactly how it would work - that change reduces the amount of messages about installing fake parts, (from 3 sets to 1) so it seemed to be working, but...

Here's the kicker. *Newly built testing setups were still experiencing incorrect power draw, that only helped with the pre-existing already setup test*. (The existing changes to vehicle::discharge_battery work for both existing and new setups).

Now I don't know *why* that is, I really don't. I'd like to disable fake parts for appliances but that doesn't seem to do it, or at least it doesn't resolve the power issue (and debugging messages suggest some are still being added).

ZeroInternalReflection:
> Newly built testing setups were still experiencing incorrect power draw, that only helped with the pre-existing already setup test

I think I've figured out where that's going wrong:

https://github.com/CleverRaven/Cataclysm-DDA/blob/26a4725dc07f8feedcc29cec91889d9140ddd5db/src/veh_appliance.cpp#L63-L71

So, when you place a car battery, it:
1. Installs the part
2. Refreshes
3. Checks to see if it's an appliance and doesn't see a tag
4. Adds a second, bonus car battery
5. _Then_ tags it all as an appliance
6. Subsequent refreshes can clear away the fake part, since this is an appliance

Moving the add_tag() call to before installing the part seems to work properly with irwiss's suggestion.

I favour having both checks, since there's nothing stopping someone from adding a big vehicle battery with the "OBSTACLE" tag, which could then get a fake part.

PR #61824, opened 2022-10-20
Fix charge usage for short string from electric spinwheel

Jerald-of-Libya:
#### Summary
None

#### Purpose of change
Go search your recipes with `t:spinwheel,-rope` and note the charge usage for each string/cordage recipe. The one for short string immediately pops out as out of place.

long cordage - 4, long string - 4
short cordage - 1, short string - 100

Time taken for the corresponding cordage vs string recipes are the same, so it makes sense that the charges used should be the same as well. Honest mistake most likely.

#### Describe the solution
Change charges used by the electric spinwheel short string recipe from 100 down to 1

N/A

#### Testing
N/A

#### Additional context
Uhhh.. that's how you use Summary:None, right?

Also, the short cordage recipe is right above the short string one in the json, so it's easy to see the charge discrepancy without going in-game.

### DIFF
--- a/data/json/recipes/other/cords_and_ropes.json
+++ b/data/json/recipes/other/cords_and_ropes.json
@@ -499,7 +499,7 @@
     "time": "15 s",
     "reversible": true,
     "autolearn": true,
-    "tools": [ [ [ "electric_spinwheel", 100 ] ] ],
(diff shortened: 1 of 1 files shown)


PatrikLundell:
Yes, that's how to use the "None" summary variant.

PR #61825, opened 2022-10-20
Improve JSON linting on windows/VS

irwiss:
#### Summary
None

#### Purpose of change

Improve JSON linting on windows / visual studio:
1) Lint just changed json files rather than all json files
2) Optionally run lint when building in VS and present lint errors nicely

#### Describe the solution

Currently the (windows) lint script iterates over all json files, there's currently ~4300 linted files and it takes around 2 minutes to lint on a modern CPU and gen 4 SSD (i'm pretty sure the bottleneck is mostly running the 4300 processes sequentially though).

The patch changes so lint script queries git for changed *.json files in: master, origin/master, upstream/master, fork/master branches. Concats the lists together, filters unique and runs lint just on the changes, this takes a few seconds for reasonable amount of changes.

---

The patch also ( optionally ) allows lint-on-build functionality.

For this to work some manual work is required to define CDDA_POST_BUILD_JSON_LINT in props file, then the script will trigger on build, and also formats output in a way VS understands, so clicking on a lint warning will bring you to the json file and put cursor on the lint warning row/column.

The props file needs to be located at Cataclysm-DDA/Directory.Build.Props or Cataclysm-DDA/msvc-full-features/Directory.Build.Props (similar to how ccache is set up) and have following content ('true' can be anything except empty string):
```
<Project>
  <PropertyGroup>
    <CDDA_POST_BUILD_JSON_LINT>true</CDDA_POST_BUILD_JSON_LINT>
  </PropertyGroup>
</Project>
```

#### Testing

Change a json file to have an error, run the lint script "manually", it should only lint the changed file

Create the optional props file, run build in Visual Studio, as part of build process the lint should trigger and a lint warning should appear in "output" and "error list" windows as depicted on screenshot, double clicking the warning in error list window should open the json file and move caret to where the linter pointed at.

#### Additional context

image

### DIFF
--- a/msvc-full-features/Cataclysm-vcpkg-static.vcxproj
+++ b/msvc-full-features/Cataclysm-vcpkg-static.vcxproj
@@ -141,6 +141,11 @@
     <Link>
       <SubSystem>Windows</SubSystem>
     </Link>
+    <CustomBuildStep Condition="'$(CDDA_POST_BUILD_JSON_LINT)' != ''">
+      <Command>C:\windows\system32\windowspowershell\v1.0\powershell.exe $(SolutionDir)\style-json.ps1 -FromMSBuild</Command>
+      <Message>Linting JSON</Message>
+      <Outputs>non-existent-file-to-always-trigger-custom-step.txt</Outputs>
+    </CustomBuildStep>
   </ItemDefinitionGroup>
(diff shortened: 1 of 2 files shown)


Maleclypse:
So how will I use this differently than my lint all json tool I have currently setup in VS? Or will this auto replace that?

PR #61826, opened 2022-10-20
Give sledgehammers durable_melee

taatu:
#### Summary
Balance "Give sledgehammers durable_melee"

#### Purpose of change
Sledgehammers not having the durable_melee flag seems like an oversight, especially since many easier to craft weapons already have it, and they're not very heavy hammers designed for demolition.

#### Describe the solution
add the flag to the engineer's hammer, short sledge hammer, sledge hammer and heavy sledge hammer


Maleclypse:
Having thrown the head from several sledgehammers I'm going to say I don't believe that's an oversight.

TheShadowFerret:
>Having thrown the head from several sledgehammers

~~Maleclypse is Captain Schwarz Walder confirmed~~

Maleclypse:
Checking in with some of the other collaborators and it seems to be unanimous that a normal sledge won't be "melee_durable".   I looked up unbreakable sledge and got this 
https://www.grozusa.com/products/30-indestructible-sledge-hammer-soft-30-hrc?variant=7738926989358&currency=USD&utm_medium=product_sync&utm_source=google&utm_content=sag_organic&utm_campaign=sag_organic&gclid=CjwKCAjw79iaBhAJEiwAPYwoCPN3UgAdJM2T3vdyEO63Xq2-CNQWmtaLk9o8G2Ahf0HqHDd3N4W4mRoC87kQAvD_BwE

One of their ad images suggests expected lifespans of various sledges when used properly.

omni-Var:
Having used multiple types of sledgehammer to demolish home walls (to include brick), I'd say these are very durable. The Durable flag has never meant "invulnerable". Durable weapons still eventually break, as sledges eventually break. I'd say tearing down multiple hard walls and seeing now thrown heads and no damage to the handle or head ... I'd completely and utterly disagree with Maleclypse. 

I believe this is a bias issue. Either remove durable from many of the weapons out there, or give sledge durable.

PR #61839, opened 2022-10-21
Improve handling of gasses

### LINKED ISSUE #61605: Loose gaseous items will sit on the ground until you bottle them up with an airtight container
### Describe the bug

Gas items sit on the ground, I would expect them to disperse or at least not be recapturable after they spill out.

### Steps to reproduce

Load the game
Spawn nitrox
See nitrox unpickable on the ground 
spawn plastic bottle
catpure nitrox in your handy plastic bottle

### Expected behavior

Gas to vanish once it spills out of it's container.

### Screenshots

pickup gasses into plastic bottles

- OS: Windows
    - OS Version: 10.0.19043.2006 (21H1)
- Game Version: b1fb6f2 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm],
    Mythical Martial Arts [MMA],
    SpeedyDex [speedydex],
]

### Additional context

_No response_

RenechCDDA:
#### Summary
Bugfixes "Player cannot handle gasses like solids"

#### Purpose of change
Closes #61605

#### Describe the solution
This prevents the player from doing the following with uncontained gasses: Picking up, consuming, wielding, moving with AIM (either individually or with move all), hauling, or loot zone sorting.  *Ugh*

All existing gasses that I found with grep had `NO_DROP` added, this will cause them to vanish(be destroyed) if they should ever find themselves outside a container. But *just in case* they don't, or new gasses are added without `NO_DROP`... the player should not be able to interact with them in any untoward ways.

Making an item with `PHASE_ID::GAS` instantly be destroyed when uncontained, as if they all had `NO_DROP` flagged. Decided against it because I already had enough fun looking through functions and testing.

#### Testing
Compiled locally each step of the way and confirmed the changes function as advertised. Spawned in all the items and made some acetlyene (a `NO_DROP` item), then further crafted it into oxyacetlyene, confirmed I was able. 

#### Additional context

### DIFF
--- a/src/advanced_inv.cpp
+++ b/src/advanced_inv.cpp
@@ -915,7 +915,7 @@ bool advanced_inventory::move_all_items()
     size_t liquid_items = 0;
     for( const advanced_inv_listitem &elem : spane.items ) {
         for( const item_location &elemit : elem.items ) {
-            if( elemit->made_of( phase_id::LIQUID ) && !elemit->is_frozen_liquid() ||
+            if( ( elemit->made_of_from_type( phase_id::LIQUID ) && !elemit->is_frozen_liquid() ) ||
                 elemit->made_of_from_type( phase_id::GAS ) ) {
                 liquid_items++;
             }


src/activity_actor.cpp
```diff
@@ -1858,7 +1858,8 @@ void move_items_activity_actor::do_turn( player_activity &act, Character &who )
         }
 
         // Check that we can pick it up.
-        if( !target->made_of_from_type( phase_id::LIQUID ) ) {
+        if( !target->made_of_from_type( phase_id::LIQUID ) &&
+            !target->made_of_from_type( phase_id::GAS ) ) {
```
mqrause, line 1862: I'd probably invert checks like this and just check for solid items, considering plasma would be equally unhandleable. Though I'm not sure if PNULL could ever show up in these cases and what it would even mean. It'd probably be a bug anyway and shouldn't be treated as if it was solid.

src/advanced_inv.cpp
```diff
@@ -913,14 +913,24 @@ bool advanced_inventory::move_all_items()
...
-    if( spane.items.empty() || liquid_items == spane.items.size() ) {
+
+    for( const advanced_inv_listitem &elem : spane.items ) {
+        for( const item_location &elemit : elem.items ) {
+            if( elemit->made_of_from_type( phase_id::GAS ) ) {
+                gas_items++;
+            }
+        }
+    }
+    if( spane.items.empty() || liquid_items == spane.items.size() || gas_items == spane.items.size() ) {
```
mqrause, line 932: Kinda like the other condition, unless you explicitly need the count of liquids and gases, count the solids and return if it's 0. That also lets you exit the loop early. And you can use `item::made_of` instead of `item::made_of_from_type` if you want frozen items to be handleable.

If you do need them explicitly, count them in the same loop. And I guess compare the size to liquids+gases in the condition instead of both individually?

src/activity_actor.cpp
RenechCDDA: I agree that plasma is an equally unsuitable material for handling, although I'm unsure if we actually have anything in the game in that form.

made_of_from_type seems to be doing strange things in regards to phase_id, especially for spilt frozen liquids. I would agree that it should probably just check for solid, but based on what I reviewed+tested I'll hold off on doing so until I get a chance to test it, probably sometime tomorrow.

mqrause: `made_of_from_type` is the default phase of the `itype`. Use `made_of` instead if you want the phase the item is currently in.

Plasma in fact only seems to be used by fields and not items. But who knows what the future brings.

RenechCDDA: Ah yes, that is indeed the difference between those two. As soon as I had closed VS it all went straight out of my head. Opening it fresh today and reviewing your comment cleared it right up.

PR #61845, opened 2022-10-21
Burrowing creatures leave a temporary field instead of a permanent terrain change

### LINKED ISSUE #58317: Tunnelling monsters should not turn gravel into dirt
### Describe the bug

This is annoying and can also be used as an exploit to get dirt where there is only gravel (meaning subway tunnels). This exploit is relatively minor, so no exploit label.

### Steps to reproduce

1. Teleport into a subway tunnel.
2. Spawn a gigantic naked mole-rat.
3. It will eventually turn the railroad rubble on the tracks into dirt.

image

Ignore the mess this random slug has made.

### Expected behavior

Turn it into rocky rubble instead. That seems more appropriate.

### Screenshots

_No response_

- OS: Linux
- Game Version: 0.F-8336-g400bccb542c [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

Ilysen:
#### Summary
Content "Burrowing creatures now leave a temporary pile of churned earth, instead of a permanent mound of dirt"

#### Purpose of change

Because pictures tell a thousand words:
image

But in all seriousness, burrowing creatures currently frequently leave behind a permanent mount of dirt terrain change whenever they move a tile. This is fairly janky, and not only does it make terrain super fugly, but it also encourages players keeping these monsters out of the reality bubble so that they don't screw up the terrain with incredible speed and ease. At the same time, burrowing creatures use dirt mounds as a telegraph, so a solution was needed to make sure that you could still see them coming from a distance.

Fixes #58317

#### Describe the solution

Burrowing creatures no longer have a chance to change the terrain type depending on their size. Instead, when they move, they place cosmetic `churned earth` fields on their tile that fade after a few ~~seconds~~ **minutes**.

Churned earth is also placed adjacent to the monster's occupied tile when it moves, with the amount depending on the creature's size. This is so that much larger worms (like graboids) visibly disturb the ground more than smaller ones (like large worms), maintaining part of the existing behavior in which mounds of dirt are placed more frequently for larger creatures.

I used `looks_like` to make sure that churned earth shared a sprite with mounds of dirt, to avoid confusion with appearances.

One thing I thought about is having the creature *always* leave a churned earth field over itself every turn, even when it was moving. I decided against this, since I figured it might make the creature too obvious/telegraphed. Otherwise, I was pretty sold on the idea of replacing the existing problem with the mound of dirt creation. :P

#### Testing

I spawned a few different types of worms at varying distances and used shouting (and later a whistle) to kite them around, observing the behavior. I tried it in open field tiles, as well as on roads, where they would often lurk in wait on the sides, invisible until disturbed or until they moved. I recorded some videos of these tests, but they're too large for github. :p

#### Additional context

I used to have some videos in here, but unfortunately the records I took of the updated behavior are too big to upload. Sorry :<

<details><summary>The trail left by a freshly-spawned worm</summary>

cataclysm-tiles_Gsvzjsom2E
</details>

<details><summary>The trail gradually disappears over several minutes...</summary>

cataclysm-tiles_kL7nqAuK8e
</details>

<details><summary>Skipping forward a little bit, the trail is fully gone after about ten minutes. Yay!</summary>

cataclysm-tiles_xdSKKEUqhL
</details>

### DIFF
--- a/src/monmove.cpp
+++ b/src/monmove.cpp
@@ -1776,6 +1776,9 @@ bool monster::move_to( const tripoint &p, bool force, bool step_on_critter,
             case creature_size::num_sizes:
                 debugmsg( "ERROR: Invalid Creature size class." );
                 break;
+            default:
+                factor = 4;
+                break;
         }
         here.add_field( pos(), fd_churned_earth, 2 );
         for( const tripoint &dest : here.points_in_radius( pos(), 1, 0 ) ) {


NetSysFire:
This would fix #58317

Ilysen:
Edited the OP, thanks

NetSysFire:
Only feedback I have so far is the lifetime of the churned earth field. It might make sense to have it last longer, like blood, bile, gore etc. This would also be realistic because churned earth will eventually revert to normal earth, after some time (weeks? but not feasible in this case).

Additionally, maybe renaming the field to "churned ground" may be better because burrowing monsters may also try to move into gravel.

PatrikLundell:
I'm not a fan of restricting the indications this much. Currently you tend to have some chance to see there are magical critters about due to the mounds, so you can stay well clear of the area (but still get your vehicles smashed up by worms on a regular basis since just a few mounds can occur naturally, and surely there are none left after the previous one got killed by the crash when there is a disturbed field).

With this you won't see anything until the creatures have detected you (or, in the case of the bugged roaming trapdoor spiders, until they attack you out of nowhere), and companions don't know how to run...

Ilysen:
for what it's worth: having done some testing, burrowing critters that haven't detected anything still roam around, and when they do so, they leave trails. I recorded a video of this locally, but it was too big to upload. in general, the behavior still attempts to closely mirror what exists in vanilla right now.

what do we think a good development would be to stop any perceived issues? should the creature always leave a churned earth field on the tile it's on, regardless of if it's moving, as mentioned in the alternatives section? or should the fields linger for longer?

PatrikLundell:
Wishing, I'd like to see one long term indication that huge monsters are pushing through the earth at high speed, thus providing a warning about the general area, as well as an immediate indication showing the immediate effects of the movement, visible even when moving though previously disturbed tiles, with the "immediate" indication being present at the current location of the monster regardless of whether it's moving or not.

Replacing upturned earth with other kind of effect has the advantage of not impeding movement when passing through disturbed areas, as well as the disadvantage of not preparing the ground for planting for free. Another issue to consider is that the current upturning of the earth is highly visible, while I'm not sure how much you'd see when not removing the ground cover, such as e.g. when messing up high and medium grass (not to mention forests, where trees block a lot of  the player's view of things the character should have free views of).

Conclusion: Probably a new type of "disturbed ground" terrain tile providing a long term effect (ideally timed to disappear after a season or so, or even fade. Could plant growth stages be used here?), plus a field effect for the moving monsters.

Ilysen:
I feel that moving back to a terrain tile just restores the same problem with the current implementation, which is screwing up the terrain in a whole area if you keep the monster in your reality bubble. I have to admit that it doesn't sound super appealing to me. the current execution of the PR is an attempt to compromise on that by making them clearly visible in the moment without causing long-term effects to the terrain; I feel like having those long-term effects just results in the same issue of ruining entire map tiles from roaming movement or them chasing down zombies. I'm inclined to wait for a dev's input on whether or not that's actually desirable before I put in the effort to maintain a behavior that I'm specifically trying to address in this PR. :p

data/json/field_type.json
```diff
@@ -1637,5 +1637,16 @@
     "half_life": "1 seconds",
     "phase": "solid",
     "display_field": false
+  },
+  {
+    "id": "fd_churned_earth",
+    "type": "field_type",
+    "intensity_levels": [ { "name": "churned earth", "sym": "^", "color": "brown", "dangerous": false } ],
+    "half_life": "1 seconds",
```
Fris0uman, line 1645: Would probably make sense to bump this up, maybe a few hours. If the point of the PR is to keep burrowing monster from defacing terrain for ever you could even bump it to a few days.

Ilysen: I'll bump up the duration. is there a way to account for field lifetime seeming to get locked in place when the field is out of the reality bubble to prevent it from happening? I'd like to avoid that if at all possible

Fris0uman: Field not getting processed out of the bubble probably deserved an issue, I think it could/should be solved like fo fires. Waiting for that to get fixed you can still bump the half life a bit, it will disapear if players stay close and that's when it matters.

Ilysen: got it - I'll push a commit once I get some time

Ilysen:
duration bumped up; instead of lasting for a few seconds, they now last several minutes, averaging around up to 5 minutes. hopefully this strikes a nice compromise.

the length of the new durations means I can't record a clip small enough to upload to github, so I don't have any more videos and I have to replicate the same behavior with screenshots. sorry :n

Fris0uman:
Need to fix the Clang errors

PR #42023, opened 2020-07-11
Reactor Room for microlabs

John-Candlebury:
#### Summary

SUMMARY: Content "Reactor map for microlabs"

#### Purpose of change

A simple reactor room to explain why the lab still has its lights on.  In the future, once facility power is implemented, restarting the reactors should be necessary for getting power back to the lab.

Also, more inching forward to a 4th hub mission.

#### Testing

Game loads properly

#### Additional context
Heres how they look

image

PR #61846, opened 2022-10-21
New map: a nuclear power plant

8street:
#### Summary
Content "New map: a nuclear power plant"

#### Purpose of change

Adding a new nuclear power plant map in game.

#### Describe the solution

I added a very large map (16x8x3 overmap tiles + a little more): a nuclear power plant. There are two power units here. Each power unit is a ~~graphite~~ light-water nuclear reactor with an electric power of 1000 MW with a reactor central hall, deaerator, two steam drums, pool bubbler, 8 main circulation pumps, control room, server room, 2 turbogenerators with a capacity of 500MW in engine room, ventilation with exhaust pipe, many corridors and various staff offices. 

I tried to be realistic. In many rooms there is nothing to take. 

Some equipment is not present in the game, so I replaced it with what I found.

There should also be an assembly shop and an administration building here, I'm a bit tired of fiddling with this map, maybe I'll add it later.

#### Testing

Yes. Many times, in game. I noticed a radiation doesn't work properly. I can't adjust the radiation for the rooms respectively, only for the entire territory.

#### Additional context
<details>
  <summary>PHOTOS THERE</summary>

Only first floor there. The rest of the floors let be a surprise.
cata2

Overhead crane in the central hall.
cata1
</details>

### DIFF
--- a/data/json/items/ammo.json
+++ b/data/json/items/ammo.json
@@ -434,7 +434,7 @@
     "symbol": "=",
     "looks_like": "cobalt_60",
     "color": "green",
-    "description": "The uranium dioxide (UO2) powder is fired in a high-temperature, sintering furnace to create hard, ceramic pellete of enriched uranium dioxide.",
+    "description": "The uranium dioxide (UO2) powder is fired in a high-temperature, sintering furnace to create hard, ceramic pellets of enriched uranium dioxide.",
     "material": [ "uran_cell" ],
     "flags": [ "LEAK_DAM", "RADIOACTIVE" ],
     "volume": "2 ml",


RenechCDDA:
> Yes. Many times, in game. I noticed a radiation doesn't work properly. I can't adjust the radiation for the rooms respectively, only for the entire territory.

You can set multiple squares for different radiation values, though it's a bit tedious.

Looks like your JSON needs to be linted. You can use the json_formatter.exe in your cdda directory or the online linter at https://dev.narc.ro/cataclysm/format.html. Personally I recommend the bundled executable, very simple and no chances of copy-paste errors.

andrei8l:
> I noticed a radiation doesn't work properly. I can't adjust the radiation for the rooms respectively, only for the entire territory.

You can add `fd_nuke_gas` fields

> Each power unit is a graphite light-water nuclear reactor

Do you mean graphite moderated, water cooled reactor? Wikipedia says there aren't any in the US. Not sure how much effort it'd take to re-fluff that.

8street:
> You can set multiple squares for different radiation values, though it's a bit tedious. See https://github.com/CleverRaven/Cataclysm-DDA/pull/42023 for a good example.

I tried. It doesn't work properly now.

> You can use the json_formatter.exe
You can add fd_nuke_gas fields

Thank you.

> Wikipedia says there aren't any in the US. 
Yes. I was inspired by the Chernobyl power plant, sorry. Therefore, I think, except for safety reactor confinement, this reactor will not differ in any way from a typical other reactor in the US.

MNG-cataclysm:
What if you designed the facility to resemble a [pressurized water reactor](https://en.m.wikipedia.org/wiki/Pressurized_water_reactor)? It's a more advanced design model than the Chernobyl-esque reactor pictured here, and the United States uses them as a majority of their commercial nuclear reactors.

8street:
@MNG-cataclysm
Now the current reactor in this PR is impersonal. I will explain. There is not any type reactors in game as suitable item, tile or furniture. In game is no nuclear fuel elemets and cells where I can put it. The core of the reactor is not named in any way and is simply surrounded by sheets of metal, and inside is practically empty. The game convention does not even allow me to place the pipes as I want. The game doesn't even have solid graphite tiles, for that matter. 

All of your named types of reactors contain common elements: core, pumps, cooling, generators. Which are in the game and which I already did. You can imagine that the reactor in this PR is a PWR or molten salt reactor and you won't be far from the truth. I only need to add a couple cooling towers. But it's not in this PR.

I also understand in PWR the rooms are located a little differently. It's very hard to find some good drawings.

anoobindisguise:
If the inside of the reactor is just empty and it's a "don't look to hard at it" sort of situation sounds like you should probably create some custom tiles to add there. Nothing pulls you out of the game faster than looking behind the curtain and finding that the nuclear reactor was merely hoping that you wouldn't check its innards to see if it actually had a reactor inside it.

8street:
> you should probably create some custom tiles to add there

Yes I will do it.

8street:
So, the core looks something like this now.
<details>

cata_core

</details>

MNG-cataclysm (approved):

data/json/mapgen/nuclear_plant/nuclear_plant_z0.json
```diff
@@ -0,0 +1,371 @@
...
+        "F................................................................................................................../f</../I                    +   + %  I/...................H.........................................H.................../I    +  %+                    I/.../<f/...............//aaaaaaaAAaaaaaaa//.........................................................................F",
+        "F...1...........................................................................................................////f/////I                    Iu  I27I7I////////////////////H/////////////////////////////////////////H////////////////////I7I72I  uI                    I//////f///.............//aaaaaaaAAaaaaaaa//.....................................................................1...F",
+        "F.............................................................................................................../IIII$IIIIII+++IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII$$$$$IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII$$H+$IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII+++IIIIII$IIII/.............//aaaaaaaAAaaaaaaa//.........................................................................F",
```
onura46, line 201:
```suggestion
        "F.............................................................................................................../IIII$IIIIII+++IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII$$$$$IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII$$H+$IIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIIII+++IIIIII$IIII/.............//aaaaaaaAAaaaaaaa//.........................................................................F",
```
Characters 120 and 273 here (I believe, unless I've gotten lost in this impressive mapgen), the `$` `t_door_metal_locked`, appear to be the only two doors made for people who work at the plant to enter. There's one on either side. There are other openings for train tracks, but it just struck me as I was playing that there were literally _zero_ emergency exits and the ones that are there are _locked_!

I would suggest adding an electric card reader that takes Industrial ID cards next to these doors, or perhaps a chance for them to be open. It might also be worth adding extra doors, but I realize you probably designed routes through this place intentionally.

data/json/scenarios.json
```diff
@@ -824,5 +824,16 @@
     "start_name": "Retirement Community",
     "professions": [ "senior", "old_vet" ],
     "flags": [ "CITY_START", "LONE_START" ]
+  },
+  {
+    "type": "scenario",
+    "id": "nuclear_plant_start",
+    "name": "Nuclear power plant",
+    "points": 2,
```
onura46, line 832:
```suggestion
    "points": 1,
```
I wouldn't take my word for this. Perhaps wait for more feedback from players to balance it. Anyway, everyone will just say "play freeform", "balance is subjective", "skill issue", blah blah blah...

A skilled player can make it worth two points for sure, especially if they try clearing it - I mean, there's some interesting loot here: electronics, cars, axes and armor if you know where to look - but atm with some of the terrifying monsters lurking around, the lack of medication/first aid, and the lack of exits, this is actually a pretty tough map to escape. I died once - quickly - and made it out twice in about 30 minutes, although pretty battered.

onura46 (commented):
I love this mapgen! Insanely detailed and _HUGE_. It's literally the size of a small city. Absolutely amazing work <3

My only wish is more emergency exits and/or for Industrial ID card readers requiring you to make a more tactical escape. There's at least six exits for train tracks that we don't have in-game yet, but only two for people. The OSHA inspector in my head kept going "why aren't there more unlocked emergency exits on the backside of this nuclear power plant per regulation 🤓". There's a little conflict with realism there, but I understand there's a gameplay balance to be struck as well.

data/json/mapgen/nuclear_plant/nuclear_plant_z0.json
8street: This wall, that you mentioned, it's outer wall of the power unit. This wall may be the so-called dirty wall. Behind it are some radioactive rooms, so-called dirty rooms. So it is not intended for personnel entry. But I'm not a power plant worker, so I could be wrong.

Cleaner (no radioactive) rooms are located on the north side of the power plant. This is a deaerator shelf (longest building in the middle) and also this is a engine room. So personnel should enter through the side of the deaerator and go straight up to the second floor, I guess.

Also, in reality, somewhere there should be a pedestrian gallery to the administrative building, which is not presents now.

Adding an industrial card to some doors would be a good idea, but I'm not sure this card exists now. This pull request is already too large. Maybe we'll add it later?

8street:
@onura46 
Thanks for your review. It is very helpful and timely, don't worry.

NetSysFire:
> The OSHA inspector in my head kept going "why aren't there more unlocked emergency exits on the backside of this nuclear power plant per regulation 🤓"

We do not have one-way doors yet, unfortunately. I had the same complaint when I polished the cinema. Windows are one-way already, so maybe someone can hack something together.

My feedback:

There is a single nuclear power plant irl in new england, and given all its equipment and loot, I would make it globally unique. It may cause problems with your starting scenario though.

My ideas:

Since you put a lot of effort into this, what would happen if you take away all of the workers at the plant and *maybe* have some mild vandalism due to the apocalypse? When would it go nuclear (pun intended)? This would make an excellent candidate for "transform over time" in the future, eventually becoming an irradiated wasteland.

Maybe some people had the idea that the evil nuclear power plant must be the cause of the Cataclysm, so I could imagine some smaller riots and protests at the plant. Basically adding a chance of spawning an additional crowd of protesting people (now zombies) with some posters and banners.

8street:
> There is a single nuclear power plant irl in new england, and given all its equipment and loot, I would make it globally unique. It may cause problems with your starting scenario though.

I'll test and tweak this a bit later. Perhaps the overall spawn chance will increase.

> When would it go nuclear (pun intended)? This would make an excellent candidate for "transform over time" in the future, eventually becoming an irradiated wasteland.

I don't think this is the proper game concept. I had an idea to add an activated atomic bomb to the reactor, with some chance. But I refused it. A shutdown nuclear reactor is very stable. Maybe your idea is more applicable to a separate mod?

> Maybe some people had the idea that the evil nuclear power plant must be the cause of the Cataclysm, so I could imagine some smaller riots and protests at the plant
 
Maybe if someone else approves these. Although according to the lore, there is a lot of unrest, but I don't think people would gather around the station. They have plenty of other things to do.

data/json/mapgen/nuclear_plant/nuclear_plant_z0.json
onura46: Sure! You're welcome to ignore it if you like; I just wanted to note it and suggest one possibility.

The card reader you'd be looking for is called `t_card_industrial`. The Irradiation Plant and the Prison already use them; not sure if anything else. But the Industrial ID (`id_industrial`) cards that they use drop fairly consistently in the wallets of zombie technicians, so that was my thought! Kill a few zombies, escape out the back.

8street: Thanks, I'll see what I can do.

8street: So, I'd added some industrial readers to map. In my opinion, it has much easier to get out and get inside in a power plant. So I added some more closed doors. But the balance is somewhat broken anyway if you have a lot of id cards.

8street:
I also mentioned, when I tested it, that `t_door_metal_interior_locked` is impossible to open with a crowbar or a lockpick. So I added these options, which I apparently forgot to copy from `t_door_metal_pickable` which is the same without the flag `OPENCLOSE_INSIDE`.

data/json/items/ammo.json
```diff
@@ -424,6 +424,25 @@
...
+    "type": "AMMO",
+    "id": "uran_cell",
+    "category": "spare_parts",
+    "price": 5000,
+    "price_postapoc": 1000,
+    "name": { "str": "uranium fuel cell" },
+    "symbol": "=",
+    "looks_like": "cobalt_60",
+    "color": "green",
+    "description": "The uranium dioxide (UO2) powder is fired in a high-temperature, sintering furnace to create hard, ceramic pellete of enriched uranium dioxide.",
```
Maleclypse, line 437:
```suggestion
    "description": "The uranium dioxide (UO2) powder is fired in a high-temperature, sintering furnace to create hard, ceramic pellets of enriched uranium dioxide.",
```

Maleclypse (approved):

kevingranade:

https://discourse.cataclysmdda.org/t/how-long-would-the-nuclear-power-plant-last/28448/3

PR #61847, opened 2022-10-21
Apocalypse themed zombie horse mutations.

Daved27hundred:
#### Summary
Content "Adds four possible mutations for the zombie horse, based on the four horsemen of the apocalypse."

#### Purpose of change

The horse does not mutate into anything, zombie or otherwise. The zombie horse is quite fast (speed of 215) but weak as a regular zombie, and I believe something like that could use some more dangerous mutations. I decided to give them a special flair instead of just naming them "acidic horse" or "skeletal horse" because I'm a big fan of zombies with unique names.

#### Describe the solution

Four new monsters in zanimal_upgrade.json, themed after the white, red, black and pale horses respectively:

- The pestilent horse, which has acid blood as well as an acid spitting attack
- The sinewy warhorse, which has a brute slam attack, along with regeneration (1/10 of its hp)
- The smoldering stallion, which emits a cloud of smoke around it, blocking line of sight and choking anyone without a gasmask.
- The pale charger, which has a brute attack like the warhorse, though instead of regeneration it has armor plating.

The zombie horse has an equal chance to mutate into either one of the variants, and starts mutating after 42 days.

#### Testing

I figured out how to run these changes on my copy of CDDA, and they do work.

### DIFF
--- a/data/json/monsters/zanimal_upgrade.json
+++ b/data/json/monsters/zanimal_upgrade.json
@@ -306,7 +306,7 @@
   {
     "id": "mon_zombie_horse_white",
     "type": "MONSTER",
-    "name": { "str": "pestilent horse" },
+    "name": { "str": "pestilent prancer" },
     "copy-from": "mon_zombie_horse",
     "description": "The abdomen of this once majestic horse has grown rotund to the point you can see through the skin, at the reservoir of caustic bile inside.  Its neck is equally enlarged as well as slightly longer than normal.",
     "diff": 3,


LyleSY:
Love the flavor. I'd like to see  "attack_cost" lines added for each matching their speed, but otherwise this makes sense. I don't know a lot about regen balance so I'm not sure if that one is balanced but the rest makes sense to me.

Daved27hundred:
> I'd like to see "attack_cost" lines added for each matching their speed

Sounds good. I’ll do it this afternoon.

LyleSY (approved):
Makes sense, clear improvement, fun choices

AcidAntOnAMinefield:
Out of curiosity, is the "hardtoshoot" flag used just to simulate their nimbleness or something? I thought that tag was exclusively used for smaller/tiny creatures that ... well... hard to shoot at. 
This said, looking forward to seeing these 4 in the wild.

Zireael07:
@AcidAntOnAMinefield in my experience it was mostly used for skeletons

AcidAntOnAMinefield:
Ah yeah those have it too, kinda like the Amigara Horror too. Yeah it does make some sense they'd have it. It's just the description mentioning size threw me off.

PR #61859, opened 2022-10-22
[MSC] Sweet Candy Flesh

Fris0uman:
#### Summary
Mods "Butchering candy monsters gives candy flesh"

#### Purpose of change

Break the cyclce of infinite monster spawning by replacing candies by candy flesh in the butchery drops

#### Describe the solution

- Introduce various form of candy flesh
- Put candy flesh in the drops from butchery
- Don't get candy from butchery anymore to break the cyclce of infinite monster spawning

#### Testing

- Load without error
- Do get hte proper flesh from butchering the candy monsters


Maleclypse (approved):
I feel what is probably an appropriate level of horror at this PR, lol.

PR #61863, opened 2022-10-22
Respect `Force unarmed` martial art style when calculating stamina cost

### LINKED ISSUE #54117: Taekwondo and "force unarmed" attacks cost stamina based on wielded weapon
### Describe the bug

The stamina cost of unarmed attacks while using Taekwondo or Force unarmed is based on the object you are wielding, even though you're not attacking with it.

### Steps to reproduce

1. Select "Force unarmed" and/or "Taekwondo"
2. Wield something heavy
3. Attack

### Expected behavior

Unarmed attacks have unarmed attack stamina cost

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: 3c42e07 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Slots [cbm_slots],
]

### Additional context

I'm not 100% sure this actually qualifies as a bug, since you do still have to hold and deal with whatever you're holding. But just holding your heavy backpack while kicking should probably exhaust you less than swinging it around, and from a mechanical standpoint this further limits the utility of these already quite narrow options.

Night-Pryanik:
#### Summary
Bugfixes "Respect `Force unarmed` martial art style when calculating stamina cost"

#### Purpose of change
* Closes #54117.

#### Describe the solution
In `get_standard_stamina_cost` added a check for `Force Unarmed` martial art style.

None.

#### Testing
Wielded heavy sledgehammer. Smashed the wall, checked stamina cost. Set Force Unarmed style, smashed wall once again, checked stamina cost.

#### Additional context
None.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -2652,7 +2652,10 @@ int Character::get_standard_stamina_cost( const item *thrown_item ) const
     // for each strike, regardless of weight. This is compensated
     // for by the additional move cost as weapon weight increases
     //If the item is thrown, override with the thrown item instead.
-    const int weight_cost = ( thrown_item == nullptr ) ? weapon.weight() /
+
+    item current_weapon = used_weapon() ? *used_weapon() : null_item_reference();
+
+    const int weight_cost = ( thrown_item == nullptr ) ? current_weapon.weight() /
                             16_gram : thrown_item->weight() / 16_gram;
     return ( weight_cost + 50 ) * -1 * get_modifier( character_modifier_melee_stamina_cost_mod );
 }


PatrikLundell:
Is smashing the same as performing an attack? To me it would seem reasonable that wielding a sledge hammer to smash something you can't destroy with your bare hands would actually use the tool wielded, rather than performing a shoulder smash.

I'd test attacking a summoned zombie as well (regardless of the answer to the question above) and verify that all the attack parameters (time, stamina, damage, possibly something else I've missed) are as expected.

Night-Pryanik:
> Is smashing the same as performing an attack?

I doubt it. The terrain/furniture you're smashing is very rarely trying to actively dodge your attacks and/or attack you in their turn, unlike most monsters in the game. I'd say smashing is much more "relaxed" compared to performing an attack.

> To me it would seem reasonable that wielding a sledge hammer to smash something you can't destroy with your bare hands would actually use the tool wielded, rather than performing a shoulder smash.

What about wielding not a sledgehammer, but a long stick instead? A glass bottle? Still use tool wielded no matter what?

If one has purposefully selected Force unarmed style when trying to smash something, then one can expect that game will follow his orders, even if one can't destroy this something with his unarmed attacks, and that game won't be trying to pretend that it's smarter than player. 

> I'd test attacking a summoned zombie as well

Attacking how exactly? With a sledgehammer? With a sledgehammer but with Force unarmed? Truly unarmed?

PatrikLundell:
I would expect a player to typically have selected Force Unarmed and leave it at that to use it for, well, attacks. However, I accept the argument that such a player ought to realize they're using unarmed smashing after failing for a while, switch style, smash, and switch back. It is rather annoying to have your delicate vase (or whatever) damaged when pulping a zombie normally, after all.

Attacking wielding a sledgehammer while having Force Unarmed applied to verify that the sledgehammer's only effect on the attacks is to make attacks using the arms and hands unavailable (or that the sledgehammer is disposed of to enable the use of the hands).

Night-Pryanik:
> the sledgehammer's only effect on the attacks is to make attacks using the arms and hands unavailable

My changes didn't affect anything related to arms and hands availability. The changes only affect stamina cost with Force Unarmed enabled.

PatrikLundell:
If, for some reason, the sledgehammer actually IS used for the attack then the stamina cost for the attack should remain unchanged, as the change would tackle only half of the problem, which wouldn't be suitable. If, on the other hand, the issue was that unarmed attacks were made while still penalizing them with the sledgehammer penalties the change is reasonable, but it should still be verified that the end result is the expected one.

If arm/hand attacks were made while still holding the sledgehammer I agree that's a partially different error that should be registered and then dealt with in a separate PR.

Night-Pryanik:
> If, for some reason, the sledgehammer actually IS used for the attack

I suppose you meant "the sledgehammer actually IS used for the attack even with Force Unarmed enabled"? If so, then no, there wasn't any issue on this matter, at least I haven't heard anything about this.

> If, on the other hand, the issue was that unarmed attacks were made while still penalizing them with the sledgehammer penalties

This is exactly the original issue was about.

> it should still be verified that the end result is the expected one.

Tested right now.

Attacked with wielded sledgehammer - stamina cost for sledgehammer.
Attacked with wielded sledgehammer, but with Force Unarmed - stamina cost for fists.
Attacked with nothing in hands - stamina cost for fists.

PatrikLundell:
Thanks. Sorry for the miscommunication and being a pain in the butt.

PR #57239, opened 2022-04-29
Power Tools do tool things

ducanishah:
#### Summary
Features "Angle grinder can cut metal, and is very good at it. Reciprocating saw nerfed to be in line with reality. Aftershock murdersaw gains immense sawing power, and Magiclysm artificer toolkit gets a better hacksaw."

#### Purpose of change
Power tools weren't getting use_methods assigned to them from json based on charged_qualities, and as such could not be used for actions such as cutting metal. This changes that, and makes it so any current and future power tools whose charged_qualities should give them use_methods from the tool_quality json, will get them. 
This PR also increases the metal cutting power of the angle grinder so it can cut through metal five times faster than a hacksaw, because it makes sense.
Buffed the SAW_M of the artificer's magical multitool (twice as fast as a hacksaw, being a magical tool that is, I would imagine, the perfect possible incarnation of a hand hacksaw, but still a hacksaw) and the aftershock murdersaw (10 times faster than hacksaw), being a bigass diamond saw, and sci fi. The safetyless version has sawing quality equivalent to an angle grinder, half as good as the safety-having version. This is because I reason you have to go slower to avoid injuring yourself, but it still cuts incredibly well when you do manage to get it safely wrangled.
Fixes #55472
Fixes #56407

#### Describe the solution
Added code to the itemfactory finalize_pre method to add use_methods based on tool charged_qualities.
Add code to the hacksaw activity_actor to allow arbitrarily higher SAW_M qualities to increase the speed of sawing through gates/windows/fences/bars.

#### Describe alternatives you've considered
Do not make power tools useful.

#### Testing
- [x] Hacksaw on bars
- [x] Angle grinder on bars
- [x] Attempt to use angle grinder on bars without any charge (should not start)
Tested both with no battery and empty battery.
- [x] Angle grinder on bars without enough charge (should cut off partway through)
- [x] Attempt to use hacksaw to continue cutting (should not resume, but restart from beginning)
- [x] Angle grinder reload and resume (should resume, not sure if using hacksaw will drop the activity_actor or not)
This does resume properly, but using a hacksaw in between *does* stop you from resuming with the angle grinder afterwards, which is fine. To note: Using either a hacksaw or angle grinder on *another* set of bars does not reset it, but partially using a hacksaw on the first set of bars *does* reset it, even if you don't fully saw through with the hacksaw. Still fine functionality, though I am curious why the activity_actor with the angle grinder is getting dropped.
[...]

PR #61865, opened 2022-10-22
Fix Reciprocating Saw

Holli-Git:
#### Summary
Bugfixes "Changes qualities of reciprocating saw"

#### Purpose of change

The reciprocating saw was previously metal sawing and wood sawing 2, but was changed to a quality of "1" for "realism" purposes. ~~Sawing with these is much faster than with handtools,~~ Sawing with these are on par with wood/metal saws, and so it should be such in game. It also allows it to be able to cut metal bars, saw off stocks and barrels. It's not the same level of sawing as a multitool.

#### Describe the solution

Makes the qualities higher.

None

#### Testing

Tested locally, works fine

### DIFF
--- a/data/json/items/tool/workshop.json
+++ b/data/json/items/tool/workshop.json
@@ -969,7 +969,7 @@
     "material": [ "steel", "plastic" ],
     "symbol": "/",
     "color": "yellow",
-    "charged_qualities": [ [ "SAW_M", 1 ], [ "SAW_W", 1 ] ],
+    "charged_qualities": [ [ "SAW_M", 3 ], [ "SAW_W", 2 ] ],
     "ammo": [ "battery" ],
     "charges_per_use": 1,
     "power_draw": "800 J",


Zireael07:
Quality has nothing to do with how fast we craft, though.

(Adding some speed factors has been long discussed but never materialized yet)

Paprika-H:
> Quality has nothing to do with how fast we craft, though.
> 
> (Adding some speed factors has been long discussed but never materialized yet)

Yes, but the saws can still be used to cut metal (window bars, etc) and vehicle work, while before, it could not.

harperers:
I'm almost certain that the low qualities on this tool is intentional.

Paprika-H:
> I'm almost certain that the low qualities on this tool is intentional. And as was said earlier, tool quality shouldn't have anything to do with crafting times.

It was intentional, but it's a pretty crappy excuse, being "realism", despite a reciprocating saw being at least on par with a wood/hacksaw. #57239. And yes, I know, it was bad wording. But with a higher tool quality it's able to cut metal and be used in vehicle work.

Zireael07:
@Paprika-H: Tbh I see no reasoning in that PR. I agree with the other changes (removing use actions) that PR made but a change to tool quality is imho undeeded

Fris0uman:
So what thing the saw can't cut in game that it can cut irl?

Paprika-H:
> So what thing the saw can't cut in game that it can cut irl?

Metal bars, removing vehicle parts, gun stocks, gun barrels, tree trunks into planks. With a level 1 sawing, you're very much limited to a few crafts, while IRL you could 100% use a reciprocating saw to cut metal apart.

PR #61868, opened 2022-10-22
Some more disassembly recipes & Size cleanup

Karol1223:
#### Summary
Balance "A few more disassembly recipes and item size alterations"

#### Purpose of change
More ways to obtain resources. And more uses for some of this trash (I'm looking at you tea box). Also, some of these items had sizes wildly off what they should have been.

#### Describe the solution
The following items have had their sizes adjusted and were given a disassembly recipe
- Jerrycan
- Pliers
- Locking pliers
- Teapot
- Pizza cutter
- Cigar cutter

Also, the box of tea bags was given a disassembly recipe. It annoyed the hell out of me not being to take it apart for cardboard.

#### Additional context
The sizes were taken from Amazon products fitting the item's description. Pliers were a 6'' slip-joint plier model and Locking pliers were a 7'' locking plier model. An exception here is the teapot. Amazon only had one fitting steel teapot and the description didn't add up to the given package dimensions or weight. So I used some British website instead, which expectedly had accurate information about teapots.

### DIFF
--- a/data/json/items/containers.json
+++ b/data/json/items/containers.json
@@ -641,7 +641,15 @@
     "material": [ "wood", "leather" ],
     "symbol": "u",
     "color": "brown",
-    "pocket_data": [ { "pocket_type": "CONTAINER", "rigid": true, "max_contains_volume": "145 ml", "max_contains_weight": "1 kg" } ]
+    "pocket_data": [
+      {
+        "pocket_type": "CONTAINER",
+        "rigid": true,
+        "max_contains_volume": "145 ml",
+        "max_contains_weight": "1 kg",
+        "max_item_length": "16 cm"
+      }
+    ]
   },
   {
     "id": "box_snack",


Zireael07:
Confused - how does one disassemble a jerrycan IRL? I think you can only cut it apart (i.e. butcher)

Karol1223:
> Confused - how does one disassemble a jerrycan IRL? I think you can only cut it apart (i.e. butcher)

Correct, cutting it apart. But well, you can't butcher a jerrycan made of _steel_. So it needed a disassembly that uses a hacksaw. And here it is.

Zireael07:
You can 'B'utcher things like televisions for instance, so imho this should be under 'B' not worked around with disassembly.

Karol1223:
> You can 'B'utcher things like televisions for instance, so imho this should be under 'B' not worked around with disassembly.

The TV's can be butchered because their material is part plastic. I do agree that this should probably be under cutting apart but cutting is so ungodly simple code-wise right now that its simply not plausible. Cutting cannot differentiate between cutting tool quality, it just needs any tool with said quality. Which roughly translates to you can cut apart a 2x4 into splinters with a kitchen knife in 6 seconds. This needs cleanup in itself, the time it takes and tool quality should depend on volume or mass of the item and perhaps different materials should need different quality (would open up the possibility of metal sawing being used to Butcher metal items) but as it stands its simply not feasible.

PR #61872, opened 2022-10-23
Add TRANSLUCENT terrain flag which blocks vision but not light

### LINKED ISSUE #56633: Let light through but not sight

Standing in front of stained glass windows in broad day light and yet I'm still in darkness.

image

It would be nice if the engine could support letting light through a tile without making the tile see through

### Solution you would like.

Being able to have some terrain/furniture obstruct sight without bloking light completly. It could allow stained glass to let light through but alos stuff like having some light pass throguh boarded up windows or under doors.

_No response_

### Additional context

_No response_

ryanaldrich:
#### Summary
Features "Add TRANSLUCENT terrain flag which blocks vision but not light"

#### Purpose of change

TRANSLUCENT can be used in combination with TRANSPARENT to create tiles which admit light but block vision, such as fosted glass, stained glass, or to simulate tall walls with small windows placed high up.

This is a binary implementation, which does not leverage the transluceny and lighting falloff system used by (e.g.) fields. This approach was taken in order to simplify the developer experience: you can simply add a flag to the terrain without having to know precisely what transluceny value would be requried to block all vision through the tile. Future work could leverage the transluceny and lighting falloff system to enable (e.g.) semi-translucent windows that let you see just a few tiles out.

* Closes #56633.

#### Describe the solution

This re-uses the approach taken by the crouching functionality. In map::build_vision_transparency_cache() we check tiles within visible range, and set them opaque for vision purposes if they have the TRANSLUCENT flag.

Leverage the transluceny and lighting falloff system used by (e.g.) fields. I didn't do this for the reasons listed above, and also because there is some bug in the translucency accumulation code that causes you to be able to see right through tiles with translucency as long as you're looking at them from a diagonal axis. I got deep down into the light casting code before I gave up on that.

#### Testing

Added a test case using the stained glass windows. Loaded it up in game and confirmed that it worked as expected in a church, and that monster vision was blocked as expected.

#### Additional context

TRANSLUCENT_church


RAldrich:
Prompted by a discussion on Discord, I have included profiler data for this change, as it does involve a new loop through all tiles in radius 60 whenever we build_vision_transparency_cache(). This data is pretty rough and isn't really a true 1:1 comparison, it was collected by loading the game and walking in a circle around the starting evac shelter. I can attach the raw profiler files, but to summarize:

## Baseline (without the changes in this PR)

|Function Name|Total CPU \[unit, %\]|Self CPU \[unit, %\]|Module|
|-|-|-|-|
|\| - public: void \_\_cdecl game::draw\(class ui\_adaptor &\)|1363 \(13.83%\)|0 \(0.00%\)|cataclysm-tiles|
|\| - public: void \_\_cdecl map::build\_map\_cache\(int,bool\)|223 \(2.26%\)|0 \(0.00%\)|cataclysm-tiles|

map::build\_vision\_transparency\_cache\(int\) time is not present, meaning the profiler did not detect it at all at a sample rate of 1000.

## PR Impact

|Function Name|Total CPU \[unit, %\]|Self CPU \[unit, %\]|Module|
|-|-|-|-|
|\| - public: void \_\_cdecl game::draw\(class ui\_adaptor &\)|1177 \(12.15%\)|0 \(0.00%\)|cataclysm-tiles|
|\| - public: void \_\_cdecl map::build\_map\_cache\(int,bool\)|252 \(2.60%\)|0 \(0.00%\)|cataclysm-tiles|
|\| - protected: bool \_\_cdecl map::build\_vision\_transparency\_cache\(int\)|34 \(0.35%\)|5 \(0.05%\)|cataclysm-tiles|


So, yes, we spend an increased amount of time in building the vision transparency cache. In a full circuit around an evac shelter, this cost us 34 milliseconds. I didn't count exactly how many tiles the route I walked was, but it was definitely more than 34, so this change costs less than a millisecond per step (on my hardware).

Note that this is a fixed cost - the number of tiles in radius 60 is always the same, so the time cost of this function will always be the same.

So, "does the increase matter?" In my personal opinion - no, it doesn't matter. I see a negligible impact on the calling build_map_cache() function, and by the time we get up to the calling game::draw() function, we're into slop trying to compare the profile data (the PR version was actually faster).

KymasTuran:
Polarized windshields soon...

Fris0uman:
Fixes #56633 ?

Fris0uman:
Why have TRANSPARENT and TRANSLUCENT be required together? TRANSLUCENT is a thing that you can't see through, so it's not transparent, it's weird to have a flag that cannot be used on its own and need to be used with another flag that it will cancel

RAldrich:
> Why have TRANSPARENT and TRANSLUCENT be required together? TRANSLUCENT is a thing that you can't see through, so it's not transparent, it's weird to have a flag that cannot be used on its own and need to be used with another flag that it will cancel.

I see how this is a bit weird, and can change it given that dependencies between flags seems like a bad approach.

I originally did it this way because it's a less invasive change: I didn't have to touch as much of the lighting code (the alternative requires checking for both flags in a variety of other places).

I suppose I could also rename TRANSLUCENT to something like BLOCKS_VISION and keep the current implementation, as in-code one is about light and the other about vision. But that seems more clunky for people working in the JSON. Is it better to bias towards slightly increased code complexity if it means more fluent JSON?

> Also please edit the documation to include the new flag https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_FLAGS.md

Yes, will do.

I'm currently on a road trip, and it will be a few weeks before I'm back at home my dev machine. Should I put this PR back into draft mode in the meantime?

Fris0uman:
> I'm currently on a road trip, and it will be a few weeks before I'm back at home my dev machine. Should I put this PR back into draft mode in the meantime?

I don't think this PR risk being accidentally merged since it's a pretty big change, but sure since it's not quite ready it'd make sense to have it as a draft

PR #61878, opened 2022-10-23
Add ability to define immunity character flags to effects, deprecate bleed_resist

Venera3:
#### Summary
Features "Add immunity character flags to effects"

#### Purpose of change

There was previously no way to confer immunity to effects via character flags (only resistance with trait IDs or blocking the application with blocked_effects). In addition, the bleed_resist trait variable effectively conferred immunity at all levels above 1 (where it did nothing).

#### Describe the solution

Add the check for flags in `Character::is_immune_effect`
Remove the check for bleed_resist form `Character::make_bleed`
Add `BLEED_IMMUNE` flags to the Blood Mage attunement and Made of Sugar traits.

A freely-defined intensity/duration modifier per flag would be more flexible, but effects are very clunky under the hood and hooking that up looked more trouble than it was worth.

#### Testing

Compiles, loads, Made of Sugar characters don't bleed. After removing the trait bleed gets applied.

#### Additional context

`Character::make_bleed` is just a detour before `add_effect` - I have a WIP branch unhardcoding bleeding and deprecating it for characters entirely, but that will definitely be Feature Freeze-worthy so this can happen on its own.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -5365,7 +5365,6 @@ bool Character::is_elec_immune() const
 
 bool Character::is_immune_effect( const efftype_id &eff ) const
 {
-    bool ret = false;
     if( eff == effect_downed ) {
         return is_throw_immune() || ( has_trait( trait_LEG_TENT_BRACE ) && footwear_factor() == 0 );
     } else if( eff == effect_onfire ) {
@@ -5383,10 +5382,10 @@ bool Character::is_immune_effect( const efftype_id &eff ) const
     }
     for( const json_character_flag &flag : eff->immune_flags ) {
         if( has_flag( flag ) ) {
-            ret = true;
+            return true;
         }
     }
-    return ret;
+    return false;
 }
 
 bool Character::is_immune_damage( const damage_type dt ) const


Fris0uman (commented):
I'm not sure but you might even be able to replace the whole for loop with a one line check of wether or not `immune_flags` and the character flags overlap.

dseguin:
> I'm not sure but you might even be able to replace the whole for loop with a one line check of wether or not immune_flags and the character flags overlap.

Not possible here since `Character::has_flag` checks against multiple flag sets. The only optimization you could do here is something like:

```diff
diff --git a/src/character.cpp b/src/character.cpp
index c41067d8a8..2d8bf3f732 100644
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -5384,12 +5384,10 @@ bool Character::is_immune_effect( const efftype_id &eff ) const
         return is_immune_damage( damage_type::ACID ) || has_trait( trait_SLIMY ) ||
                has_trait( trait_VISCOUS );
     }
-    for( const json_character_flag &flag : eff->immune_flags ) {
-        if( has_flag( flag ) ) {
-            return true;
-        }
-    }
-    return false;
+    return std::find_if( eff->immune_flags.begin(), eff->immune_flags.end(),
+    [this]( const json_character_flag & flag ) {
+        return this->has_flag( flag );
+    } ) != eff->immune_flags.end();
 }
 
 bool Character::is_immune_damage( const damage_type dt ) const
```

But I don't think it matters that much

Venera3:
> But I don't think it matters that much

Well this will only run once per effect addition so I'm not going to learn the spicy brackets for that 😄

PR #61895, opened 2022-10-24
scar_l and scar_h high end folding stock

nopenoperson:
#### Summary
None

#### Purpose of change
scar rifles have  an adjustable length folding stock irl

#### Describe the solution
unvariant scar_l and add a default mod. add default mod to scar_h

make it built_in

#### Testing
some. dunno how the migration would go down exactly if you had to migrate a modded scar_l


Maleclypse:
```Missing Generic Guns migrations for these types of guns:
scar_l

The above errors can be resolved by either adding suitable migrations to Generic Guns or adding to the whitelists of things not requiring migration in tools/json_tools/generic_guns_validator.py```

Do you need help with this error?

PR #61908, opened 2022-10-25
add bungalow 09 and 10

LilShiningMan:
#### Summary
Content "add bungalow 09 and 10"

#### Purpose of change

more houses

#### Testing

linted and spawned in-game.

#### Additional context

bungalow09 ground

bungalow09 roof

bungalow10 ground

bungalow10 roof


data/json/mapgen/house/bungalow09.json
```diff
@@ -0,0 +1,94 @@
...
+        "``3..................|``",
+        "``3..................|``",
+        "``22222222222222222222``",
+        "````````````````````````",
+        "````````````````````````",
+        "````````````````````````",
+        "````````````````````````"
+      ],
+      "terrain": { "`": "t_open_air", ".": "t_shingle_flat_roof" },
+      "furniture": { "s": "f_solar_unit" }
```
Maleclypse, line 91:
```suggestion
      "furniture": { "s": "f_solar_unit" }
```
I'd prefer if you made a nested chunk for rooftops of a solar setup that's some solar units and whatnot. Then have those chunks have a low chance on lots of rooftops

LilShiningMan: I've thought about something similar, but in my opinion, having a generic rooftop "chunk" that contains yet other chunks including solar, enemy encounters, etc. etc. would be better. That way, you'd only have to add the generic residential rooftop chunk overlay on each house, and whenever you wanted to add/remove/tweak the chunks for, say, all residential rooftop spawns, you don't edit each house json file individually -- rather, you'd tweak the "master" chunk that fills in each of the house rooftops. That would be beyond the scope of this specific PR though.

PR #61912, opened 2022-10-25
Fix #61901 activating a lit Molotov causes it to explode

### LINKED ISSUE #61901: Molotov Lit up Twice is death
### Describe the bug

When having a Molotov wielded and you activate it twice, it burns the 3x3 area around the player and the player. 

### Steps to reproduce

1. get a lighter and a molotov
2. wield the molotov
3. activate the molotov
4. activate the lit up molotov

### Expected behavior

i would expect a molotov be can't activated twice and burning the player.

### Screenshots

_No response_

Linux 5.15.65-1-MANJARO
Game Version 26a4725dc07f8feedcc29cec91889d9140ddd5db
Ultica and CC Sound
English
No Mods

### Additional context

_No response_

ryanaldrich:
#### Summary
Bugfixes "Activating a lit Molotov cocktail no longer causes it to explode in your hand"

#### Purpose of change

Fix #61901

#### Describe the solution

The `iuse::molotov_lit` function looks pretty old, and doesn't seem to use the newer JSON explosion structure (as I'm not sure that supports creating the fire field needed for the Molotov). This just refactors it to check the activation state and charge level in the correct order to avoid blowing up until thrown or dropped.

Allowing hapless survivors to continue blowing themselves up. Didn't make any sense, as Molotovs don't explode until the glass is broken IRL, and in game if you just hold them they burn out on their own after a few turns.

#### Testing

Manual testing, spawned in an unlit Molotov and confirmed that:
- Lighting it still works.
- Activating a lit Molotov gets you a message that it's already lit (no explosion).
- Throwing a lit Molotov causes it to explode 1 turn after landing.
- Dropping a lit Molotov causes it to explode 1 turn after dropping.
- Holding a lit Molotov gives a 20% chance per turn that it burns out and turns back into an unlit Molotov.

#### Additional context

N/A

### DIFF
--- a/src/iuse.cpp
+++ b/src/iuse.cpp
@@ -3816,31 +3816,31 @@ cata::optional<int> iuse::grenade_inc_act( Character *p, item *it, bool t, const
 
 cata::optional<int> iuse::molotov_lit( Character *p, item *it, bool t, const tripoint &pos )
 {
-    if( pos.x == -999 || pos.y == -999 ) {
-        return cata::nullopt;
-    } else if( !t ) {
-        map &here = get_map();
-        for( const tripoint &pt : here.points_in_radius( pos, 1, 0 ) ) {
-            const int intensity = 1 + one_in( 3 ) + one_in( 5 );
-            here.add_field( pt, fd_fire, intensity );
+    if( !t ) { // The Molotov is no longer active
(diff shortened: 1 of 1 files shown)


RAldrich:
I am Just Quit on Discord.

PR #61917, opened 2022-10-25
Magiclysm minor weight and volume  fixes

Knut-Aage-Hofseth:
#### Summary
Mods "Fixes to volume and length for Ichavail and Technomancer toolbar"

#### Purpose of change
Update Ichaival with a sane longest side and Technomancer toolbar with a sane volume.

#### Describe the solution
Ichaival showed up in game with a length of 17cm, and a volume of 5L. I brought it in line with the reflexrecurvebow, the one it looks like, when it comes to length and volume. Now 150cm and 2L. Updated flags to include "SHEATH_BOW" and armour coverage to reflect the new sub-bodyparts for wearing the bow on your back.
The technomancer toolbar was shorter and lighter than a warhammer, but had a volume that was about 60% larger. I did a new weight definition that simplified some things and gave me about the same weight. Using the rules for that I estimated a volume that seems more reasonable.

This is for some weird balancing reason I did not understand. 
Various other ways of trying to calculate physical properties of imaginary objects. (I was bored.)

#### Testing
Will build and test at home. Test planned is simply to spawn in the items and check that the values seem reasonable. Compare the bow to the olympic bow and the Toolbar to the crowbar.
Building and spawning in the items did not present any issues. Show below is screenshots of item comparisons between the Magiclysm items and the control items I used. Changing weight and volume does impact some use characteristics of weapons, but I believe the impact to be low enough to not negatively impact balance.

Ichaival turns out to be slightly better in aim speed, improves by 9 aim speed, and loose 1 range on careful and precise for my test character.

The Toolbar ends up improving "Base moves per attack" with 21 moves. It hits a bit harder than a crowbar but is slower.

Old Bows comparison
Screenshot 2022-10-26 20 13 41

New Bows comparison
Screenshot 2022-10-26 20 58 25

Old Toolbar comparison
Screenshot 2022-10-26 20 14 59

New Toolbar comparison
Screenshot 2022-10-26 21 00 34

#### Additional context
I ended up sketching mockups of the Toolbar to try to understand the parts it needed.

I'm starting to think the Toolbar should be modeled on a 1handed infantry warhammer and perform somewhat worse in combat as the tool vs weapon compromises of the Toolbar would swing more towards the tool side than for the warhammer.


GuardianDll (approved):
Thank you for your work!

PR #61946, opened 2022-10-27
feat: change labeler action for more smarter regex

casswedson:
#### Summary
None

#### Purpose of change
I want to change the labeler action such that it is capable of substring searches, they are the bulk of the labeling rules and very much the *smart* part of this little workflow, if it can't do semi smart moves then someone will step up and do the hard work; I am looking for better smarter automation

#### Describe the solution
use this [action](https://github.com/Decathlon/pull-request-labeler-action); pretty much the same, only it's better 

see this pr for testing https://github.com/casswedson/Cataclysm-DDA/pull/68
- it does file extension detection
if I change json and .h or .cpp files it applies `json` and `c++` labels
- directory searches
if I change some `data/mods/innawood/` file it adds the `innawod` `mods` labels because the have separate rules that happen to meet here
- substring searches
if I change `data/json/deconstruction.json` it adds the `crafting / construction / etc` label cuz it looks for (among others) the substring `construction`
if I change `data/mods/innawood/traps.json` it adds the `furniture / traps` labels cuz it looks for (among others) the substring `traps`

there's also the action's output that list the files changed, labeling rules and labels to be applied; very useful https://github.com/casswedson/Cataclysm-DDA/actions/runs/3334016607/jobs/5516537963#step:4:17 

also number 2. we did lose a little logic, I am not so sure how to replicate it, in time I'm sure I'll figure it out, I just need to get the heavy work done first

also number 3. this runs on the `pull_request` event now (won't work otherwise) just a note 

#### Testing
yes

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,225 +1,170 @@
-"[C++]":
-  - "**/*.cpp"
-  - "**/*.h"
+- regExp: ".*\\.(cpp|h)+$"
+  labels: ["[C++]"]
 
-"[JSON]":
-  - "**/*.json"
+- regExp: ".*\\.json+$"
+  labels: ["[JSON]"]
 
-"[Markdown]":
-  - "**/*.md"
+- regExp: ".*\\.md+$"
+  labels: ["[Markdown]"]
 
-"Game: Achievements / Conducts / Scores":
-  - "**scores**"
-  - "**statistics**"
-  - "**achievements**"
-  - "**conducts**"
+- regExp: ".*(scores|statistics|achievements|conducts).*"
+  labels: ["Game: Achievements / Conducts / Scores"]
 
-"[Python]":
-  - "**/*.py"
+- regExp: ".*\\.py+$"
+  labels: ["[Python]"]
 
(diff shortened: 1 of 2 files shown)


BrettDong:
Workflows running on `pull_request` events have read-only permission. For write permission it needs to run on a trusted environment like `pull_request_target`.
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

casswedson:
I know for certain this does not work if I change the event trigger
I just re-tested that to be extra sure here https://github.com/casswedson/Cataclysm-DDA/pull/69 the output https://github.com/casswedson/Cataclysm-DDA/actions/runs/3351962862/jobs/5553766523
a pic to illustrate the point
image

the action is probably just old

BrettDong:
Found https://github.com/Decathlon/pull-request-labeler-action/issues/44

casswedson:
I'm gonna be digging in its code; maybe I'll figure something out

casswedson:
I'm gonna be digging in its code; maybe I'll figure something out

Issue #61702, opened 2022-10-15
Please expand "intensity" to use ranges in "place_field"


Maleclypse:
### Is your feature request related to a problem? Please describe.

I'd like to be able to randomize field intensity when setting up terrain based storytelling gore.  fd_blood should be able to be randomized :)

### Solution you would like.

`"place_fields": [ { "field": "fd_blood", "x": [ 0, 4 ], "y": [ 0, 4 ], "intensity": [ 2, 4 ]`

ZhilkinSerg:
```PATCH
@@ -0,0 +1,35 @@
diff --git a/src/mapgen.cpp b/src/mapgen.cpp
index 33aa998d7e3..deaddfd63ae 100644
--- a/src/mapgen.cpp
+++ b/src/mapgen.cpp
@@ -1707,14 +1707,19 @@ class jmapgen_field : public jmapgen_piece
 {
     public:
         mapgen_value<field_type_id> ftype;
-        int intensity;
+        std::vector<int> intensities;
         time_duration age;
         bool remove;
         jmapgen_field( const JsonObject &jsi, const std::string &/*context*/ ) :
             ftype( jsi.get_member( "field" ) )
-            , intensity( jsi.get_int( "intensity", 1 ) )
             , age( time_duration::from_turns( jsi.get_int( "age", 0 ) ) )
             , remove( jsi.get_bool( "remove", false ) ) {
+            for( JsonValue jv : jsi.get_array( "intensity" ) ) {
+                intensities.push_back( jv.get_int() );
+            }
+            if( intensities.empty() ) {
+                intensities.push_back( jsi.get_int( "intensity", 1 ) );
+            }
         }
         void apply( const mapgendata &dat, const jmapgen_int &x, const jmapgen_int &y,
                     const std::string &/*context*/ ) const override {
@@ -1726,7 +1731,7 @@ class jmapgen_field : public jmapgen_piece
                 dat.m.remove_field( tripoint( x.get(), y.get(), dat.m.get_abs_sub().z() ), chosen_id );
             } else {
                 dat.m.add_field( tripoint( x.get(), y.get(), dat.m.get_abs_sub().z() ), chosen_id,
-                                 intensity, age );
+                                 random_entry( intensities ), age );
             }
         }
 ```

PR #61947, opened 2022-10-27
Field intensity range instead of integer

Maleclypse:
Co-Authored-By: Zhilkin Serg <ZhilkinSerg@users.noreply.github.com>

#### Summary
None 

#### Purpose of change
Allows contributors to put in randomized fields where appropriate.  Should give more variation to chunks and map extras.

#### Describe the solution
Made the changes ZS recommended in #61702.   Then applies that solution in a couple of locations.

None

#### Testing
TBD

### DIFF
--- a/src/mapgen.cpp
+++ b/src/mapgen.cpp
@@ -1714,9 +1714,11 @@ class jmapgen_field : public jmapgen_piece
             ftype( jsi.get_member( "field" ) )
             , age( time_duration::from_turns( jsi.get_int( "age", 0 ) ) )
             , remove( jsi.get_bool( "remove", false ) ) {
-            for( JsonValue jv : jsi.get_array( "intensity" ) ) {
-                intensities.push_back( jv.get_int() );
-            }
+            if ( jsi.has_array( "intensity" ) ) {
+                for( JsonValue jv : jsi.get_array( "intensity" ) ) {
+                    intensities.push_back( jv.get_int() );
+                }
+            } 
             if( intensities.empty() ) {
                 intensities.push_back( jsi.get_int( "intensity", 1 ) );
             }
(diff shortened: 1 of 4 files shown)


NetSysFire:
Cool! Could you please also update the documentation when you have time?

Maleclypse:
> Cool! Could you please also update the documentation when you have time?

Sure thing, when I looked last night I didn't see any documentation of intensity in mapgen so I left that part for later.

Maleclypse:
@mqrause I think I've reverted it back to the original option ZS made.  If you have a moment could you show me the edit needed to make it accept both arrays and single integers?

src/mapgen.cpp
```diff
@@ -1707,14 +1707,19 @@ class jmapgen_field : public jmapgen_piece
...
         time_duration age;
         bool remove;
         jmapgen_field( const JsonObject &jsi, const std::string &/*context*/ ) :
             ftype( jsi.get_member( "field" ) )
-            , intensity( jsi.get_int( "intensity", 1 ) )
             , age( time_duration::from_turns( jsi.get_int( "age", 0 ) ) )
             , remove( jsi.get_bool( "remove", false ) ) {
+            for( JsonValue jv : jsi.get_array( "intensity" ) ) {
+                intensities.push_back( jv.get_int() );
+            }
```
mqrause, line 1719:
```suggestion
            if ( jsi.has_array( "intensity" ) ) {
                for( JsonValue jv : jsi.get_array( "intensity" ) ) {
                    intensities.push_back( jv.get_int() );
                }
            } 
```

I think doing this should be enough. This checks if the given value is an array before trying to read it. And if it's not the fallback below will be triggered to read an int.

PR #61949, opened 2022-10-27
Remove HURT fear trigger from nether species

### LINKED ISSUE #61948: Persons can be farmed for combat experience
### Describe the bug

Person aggression can be dropped by landing hits on them. Combining this with the portal storm enemy disappearance bug can allow a survivor to farm whatever combat skills they want on them.

### Steps to reproduce

1. Wait indoors for a portal storm.
2. During the storm once a Person has spawned go outside, fight some monsters a bit and go and back inside.
3. Wait for the storm to end. If you are still able to see Persons they will not despawn naturally.
4. Attack a Person until its aggression drops to "tracking". A fast weapon with a sweep or stunning attack works best. This is not necessary if you are able to kite or stunlock the Person but will prevent it from attacking back and disappearing.
5. Equip your weapon of choice or go unarmed and go to town on it.
6. Enjoy level 10 in combat skills.

### Expected behavior

I expect that Portal Storm monsters should disappear when the storm ends. Failing that, I expect that Persons should not be able to be engaged in combat.

### Screenshots

image
image

- OS: Windows
    - OS Version: 10.0.19043.2130 (21H1)
- Game Version: 0.F-12130-g90e9a75e28 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

Venera3:
#### Summary
Bugfixes "Fix pacifying nether creatures with a stick"

#### Purpose of change

Fixes #61948

#### Describe the solution

Removed the trigger from species data. 

None

#### Testing

Whacked a hostile horror from beyond with a stick, didn't get any less hostile

#### Additional context

Species triggers are loaded after the main def (and so after extend/delete) meaning there's no way to remove the trigger on a per-species basis. The only non-100 morale nether critter is my unseen hunter, and they already had the trigger explicitly defined. The whole bunch could of course use a facelift, but in the meantime "angry all the time" works well enough.

### DIFF
--- a/data/json/species.json
+++ b/data/json/species.json
@@ -56,8 +56,7 @@
   {
     "type": "SPECIES",
     "id": "NETHER",
-    "description": "a nether creature",
-    "fear_triggers": [ "HURT" ]
+    "description": "a nether creature"
   },
   {
     "type": "SPECIES",


GuardianDll:
Oh, right
That will also do the trick

MNG-cataclysm (approved):

Maleclypse:
Are there any remaining creatures with the Nether species that should be moved to Demihuman?

Numbuh474:
Blank bodies, gozu, and probably gracken and albino penguins too.

Maleclypse:
> Blank bodies, gozu, and probably gracken and albino penguins too.

I guess it was my mistake for asking someone else to go and look but I wasn't at a  computer at that moment.  All of those are already done except for albino penguins which neither speak human tongues nor look humanoid shaped.

Numbuh474:
Ah. I was ignoring the "to Demihuman" part. I picked the ones that I did because they had the "nether" species tag, are vaguely animal-esque, and should probably have some self-preservation instinct that (I thought) may have been disrupted by removing their fear trigger.

Maleclypse:
You are correct but that’s a different question than what I was asking. They will need a new species to give them back self preservation but again probably not for the albino penguins as they are basically shoggoths.

Venera3:
Nothing's preventing anybody from adding the trigger to *penguins* explicitly, species-level trigger definition is a remnant from a weird time where you had to mine JSON lines from the bedrock or something like that. 

The species ID gets checked for some venom immunity things so just adding a new species would not really work as well.

PR #61950, opened 2022-10-27
Epower and power in units::power

Hirmuolio:
#### Summary

Infrastructure "Epower and power in units::power"

#### Purpose of change

Epower (electric power) of vehicle parts was in integer watts.
Power (engine output) of vehicle parts was in integer watts.
Now they are in units::power.

#### Describe the solution

Change units.

Also it seems like out of reality bubble drain from emitter vehicle parts was not calculated at all. It is now.

#### Testing

Tests pass. Game runs. Nothing obviously weird happens with vehicles.

TODO: Test the emitter change.

#### Additional context

Drag is a bit odd. It is basically calculated as `water wheel power + wind turbine power` with extra `+1500` if towing. Since it directly comes from power its unit is units::power. 
But later it is used directly as acceleration. Which feels very wrong. It is not even scaled to speed in any way.


mqrause:
> For example
> `units::energy produced = power * to_seconds<int64_t>( d );`
> would beciome something like
> `units::energy produced = units::from_millijoule( units::to_milliwatt(power) ) * to_seconds<int64_t>( d );`

To simplify this you would need an overload for `units::power * time_duration` that returns `units::energy`. Might be worth it, if there are a bunch of calculations like this. Some template magician could probably even set it up so we can define unit relations and then it Just Works™.

Maleclypse:
Adding P5 to make this stale proof. I’m assuming this is targeting for next experimental?

Zireael07:
At last...

PR #61951, opened 2022-10-27
[Magiclysm] Mixed stuff

GuardianDll:
#### Summary
None
#### Purpose of change
I thought to replace a magnetism mage spell effect with targeted_monster_species spell, and describe targeted_monster_species itself in the magic.md, but accidentally found there is no place for this. So i made it
#### Describe the solution
Rewrite a magnetism mage spell to use new targeted_monster_species field
Add information about different spell fields and examples (there is some duplication of existing info, but i think its okay)
Add info ARMOR_X can both decrease and increase the damage caused to the target
#### Testing
Autotests, you are my hope
#### Additional context
I still not sure how to make proper examples: the existing way looks a bit ugly
image

But the only alretnative i may do isn't better
image

### DIFF
--- a/data/mods/Magiclysm/Spells/attunements/Magnetism_Mage.json
+++ b/data/mods/Magiclysm/Spells/attunements/Magnetism_Mage.json
@@ -289,7 +289,7 @@
     "id": "robot_disabler_explosion",
     "type": "SPELL",
     "name": "Surge",
-    "description": "Send a wave of high-frequency alternating pulse.  It can't hurt living creatures, but extremely good in destroying and stunning robots.  No, it's not EMP.",
+    "description": "Send a wave of high-frequency alternating pulse.  It can't hurt living creatures, but is extremely good destroying and stunning robots.  No, it's not EMP.",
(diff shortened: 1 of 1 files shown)


data/mods/Magiclysm/Spells/attunements/Magnetism_Mage.json
```diff
@@ -328,9 +289,10 @@
     "id": "robot_disabler_explosion",
     "type": "SPELL",
     "name": "Surge",
-    "description": "Send a wave of high-frequency alternating pulse.  It can't hurt really much, but it works great against circuits, shutting them down for a small time.  Probably it will work against drones or robots.  No, it's not EMP.",
+    "description": "Send a wave of high-frequency alternating pulse.  It can't hurt living creatures, but extremely good in destroying and stunning robots.  No, it's not EMP.",
```
KorGgenT, line 292:
```suggestion
    "description": "Send a wave of high-frequency alternating pulse.  It can't hurt living creatures, but is extremely good destroying and stunning robots.  No, it's not EMP.",
```

PR #61956, opened 2022-10-27
Magiclysm: fixes use action on drinkable mutagen

Knut-Aage-Hofseth:
#### Summary
Mods "Extends the fix from  #61930 to the drinkable mutagens in Magiclysm"

#### Purpose of change
 #61930 fixes and issue with the drinkable mutagens. This copies the fix to the Magiclysm mutagens that had the same code as the pre-fix vanilla mutagens.
Verified that drinking the Magiclysm mutagens pre fix does not introduce mutagen vitamins to the character. Verified with debug menu.

#### Describe the solution
Copied the code part from  #61930. Changed the message and vitamin.

#### Testing
Build game. Debug in the drinkable Mana mutagen and Black Dragon mutagen. Drink them.
Correct vitamins now show in debug needs.

### DIFF
--- a/data/mods/Magiclysm/items/mutagen.json
+++ b/data/mods/Magiclysm/items/mutagen.json
@@ -35,7 +35,11 @@
     "color": "light_blue",
     "description": "A mana potion mixed with mutagen.  It might help you connect with magic.",
     "name": { "str": "mana mutagen" },
-    "vitamins": [ [ "mutagen_manatouched", 175 ], [ "mutagen", 125 ] ]
+    "use_action": {
+      "type": "consume_drug",
+      "activation_message": "You drink the mana mutagen.",
+      "vitamins": [ [ "mutagen_manatouched", 175 ], [ "mutagen", 125 ] ]
+    }
   },
   {
     "id": "mutagen_black_dragon",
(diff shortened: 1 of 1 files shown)


Maleclypse:
Does this also need to be done to Dinomod and Aftershock mutagens?

Knut-Aage-Hofseth:
I think so. One would need to check. 

I looked up foods and drink now. It might enough or even a better solution to set the `"comestible_type": "DRINK"` instead? That might work.

Maleclypse:
> It seems to need either an use action or a comestible type that will be digested. Since calcium from milk hits the need overview after a period of digestion but the mutagen, that is only comestible and does not have a comestible_type does not show up after a digestion period.
> 
> I'm too tired and keep making bad assumtions. The NO_INGEST flag on the `"abstract": "mutagen_flavor"` is probably what demands the use action. But I keep making mistakes so?

@Ilysen I think you probably know what's needed here better than I do.

Ilysen:
this PR is indeed the method I went with for the vanilla mutagens. I'm not actually sure if there's a possible better way to do it - I chose `consume_drug` because it added the defined vitamins to your body instantly instead of over time, which felt better to me; I admit, though, that there might be a better way I don't know of. this method should be fine, however!

GuardianDll (approved):
Thank you for your work!

Issue #61915, opened 2022-10-25
U.I. Lies to the player - the Activity Suit


AcidAntOnAMinefield:
### Describe the bug

Activity Suit gives 100 coverage for the whole body. If you examine it and spend about half an hour scrolling down on the [e]xamine item UI, you will reach to the part where the important details are. If you pay attention, you can see that it states the following:

- "This gear **completely protects** you from electric discharges (which is true, can confirm);
- "This clothing **completely protects** you from radiation (which, so far, seems to be true);
_- "This clothing **completely protects** you from any gas. (which is 100% false, seeing I got prompted if I wanted to step into a toxic gas tile, and I, indeed, got poisoned._ 

### Steps to reproduce

1. Create a new world in a lab or something where toxic gas is guaranteed to happen (or spawn in some bloated zombies, really). Create your average Survivor.
2. Spawn Activity Suit.
3. Spend some precious time scrolling down in order to read Activity Suit's description;
4. Step on a gas tile;
5. Notice how you get poisoned.

### Expected behavior

I would expect the UI to have the important information visible to me, as well as said information to be true and not a red herring or an outright lie.

### Versions and configuration

Windows 7
e532171
no mods worth mentioning.

### Additional context

This seems to be a recurring problem with a lot of items. Not necessarily the "misinformation", but mainly the UI being so overly bloated that it is outright uncomfortable to get any information. It is painful to wade through a menu window that features endless amounts of information that is either not explained to the player or that is poorly conveyed to the point where the player is left even more confused.

For example: 
Unexplained Information: nowhere in the game we are told what Breathability does exactly. Yes, we can discern that it has to do with sweating and stuff, but the point is: Is 30% better than 80%? Number% of what exactly, what unit is this? It's the same thing with the "protection for (body part)". What is the "red%" and the "green%"? Where in the game is this explained? Or rather, why are these values so cryptic and placed in a makeshift-columnesque way under "protection" with no further context to them? 

This then brings up yet another issue: why do we have a new line for each little inch of body? The foot arches, the pinky finger, the bottom lip, the back of the hand? Can't we just have the flat numbers for the flat limbs, if the coverage and protection values are the same? Or rather, why not have just a table with the values for that item, instead of several dozens lines per limb?

[...]

PR #61959, opened 2022-10-27
Updated activity suit description to be more clear

bean-b:
#### Summary
Content "Updated activity suit description"

#### Purpose of change
Other environmental suits mention the need for a gas mask for complete protection, and I added that to the activity suit for futher clarification. 

#### Describe the solution
Added "It requires a separate gas mask for full protection." to the end of the activity suit description.

Moving the "It requires a separate gas mask for full protection." to the beginning. However, other environmental suits have that at the end, so I wanted to include it there for consistencies sake.

#### Testing
Tested in game, description looks right

#### Additional context
Response to this issue
https://github.com/CleverRaven/Cataclysm-DDA/issues/61915

### DIFF
--- a/data/json/items/armor/suits_protection.json
+++ b/data/json/items/armor/suits_protection.json
@@ -2797,7 +2797,7 @@
     "copy-from": "robofac_enviro_suit",
     "type": "ARMOR",
     "name": { "str": "activity suit" },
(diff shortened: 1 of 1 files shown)


bean-b:
Does anyone know why this build failed? I'm a first time contributer, and to me it seems like such a tiny change such of adding one line of description to one item shouldn't cause a build faliure, but of course I dont know much about this. The last error message made sense because I was missing a space, however this one of "
 (~[slow] ~[.],starting_items)=>   CHECK( dummy.pos() != player_initial_pos )
Error: (~[slow] ~[.],starting_items)=>   (60,60,0) != (60,60,0)
"
I'm not sure whats wrong with that

MNG-cataclysm (approved):

PR #61972, opened 2022-10-28
fix `t_hole` not acting like what holes should

gettingusedto:
#### Summary
Bugfixes "Fix unable to peek down/ climb down `t_hole` and drop something onto it so it'll fall through"

#### Purpose of change

Fix not being able to peek down and climbing down through `t_hole`. I noticed it when I was looking at the aftermath of a house I burned down. 

#### Describe the solution

Give `"examine_action": "ledge"` to `t_hole`, just like `t_open_air`.

Not doing so and let me be in constant annoyance.

#### Testing

I made it a mod, try to examine the hole and it works as expected.

### DIFF
--- a/data/json/furniture_and_terrain/terrain-traps.json
+++ b/data/json/furniture_and_terrain/terrain-traps.json
@@ -8,7 +8,7 @@
     "color": "black",
     "move_cost": 2,
     "trap": "tr_ledge",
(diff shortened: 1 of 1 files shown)


gettingusedto:
I noticed I can't drop stuff onto the `t_hole`, so I delete the `NOITEM` flag so it's possible.

For testing I made the changes on my aforementioned mod, made holes on a house's roof, and drop something onto it. It falls through and goes down a z-level.

PR #46153, opened 2020-12-19
Needs testing: Rewrite crafting skill checks to use normal_roll.

I-am-Erk:
#### Summary
SUMMARY: Infrastructure "Rewrite crafting skill checks to use normal_roll"

#### Purpose of change
I have revived this PR after a few years inactive, because we still need it and it turns out it wasn't that stale.

~~1. Proficiency failure penalties are currently not being applied correctly. This is indirectly because the way skill checks are calculated is needlessly complicated.~~ no longer relevant.
1. Per #44738, it is a good idea to start moving major skill checks into normal rolls. Then when we create a standardized roll_check routine, it will be much easier to write and to migrate. The current crafting skill check method is impenetrably complex and the complexity offers no real benefit.
2. Per #66064, there are two crafting check issues that need resolution: first, crafting checks from skill 0 always fail, and second, crafting checks do not take into account marginal skill gain which interacts poorly with rust mechanics and will cause problems when skill gain is slowed.

#### Describe the solution
To solve both of these, I have migrated the crafting skill check to use normal_roll and make it mathematically predictable. This brings in a few major changes:

1. NPCs no longer have to be more skilled than you to increase your chance of success. Instead, each NPC able to help that also has all the prerequisites to be able to do the craft on their own gets a skill check just like you. Then we add those skill checks together using the formula: 
```
final skill value = sqrt( your skill value^2 + sum (all contributing NPC skill values^2) )
```
This root-sum-of-squares is the same system used in summing morale: it creates diminishing returns the more contributors you have, and makes more skilled sources of help substantially outweigh smaller ones. Using this formula, for example, if you and a helper NPC both have an effective skill level of 7 once all effects are applied, your combined skill level is around 10. If you add another NPC of effective skill level 7, you get a total combined skill of 12.

In the end this makes NPC helpers much more useful (previously they just added a small flat bonus, and only if they were more skilled than you).

3. Changes the role of secondary and primary skills in your check, making secondary skills potentially more important. Previously the weight of primary and secondary skill requirements was flat: 3/4 of your skill check was from your primary skill and 1/4 was from the average of the secondary skills. Now it is a weighted average of: `2*primary skill*required level of primary skill, + sum of secondary skills * required level of secondary skills`. That means if a recipe asks for electronics 2 and cooking 1 (what the hell recipe is this??), your overall level in electronics will be twice as important to your final skill value. If that same recipe uses fabrication 2 as its primary skill, then fabrication will weigh in at 2x as important as electronics and 4x as important as cooking.
[...]

PR #61985, opened 2022-10-28
Characters with 0 skill can now craft level 1 recipes

### LINKED ISSUE #50787: Never gain any progress on crafting items
### Describe the bug

With the new practice/knowledge system the only reliable way to gain practice level is to craft things( train first aid by using bandages? Or practice computer by hacking terminals? They are just punishing and provide only tiny exp).

Now the problem comes: there are no recipes using firstaid with 0 difficulty, and the player can not gain any progress crafting item with difficulty > 0 (and therefore can not gain any exp).

in `set_next_failure_point`:
`const int failure_point_delta = crafter.crafting_success_roll( get_making() ) * percent;`

in `crafter.crafting_success_roll`:
`if( secondary_difficulty > 0 ) {`
`        skill_dice = get_skill_level( making.skill_used ) * 3 + secondary_dice;`
`    } else {`
`        skill_dice = get_skill_level( making.skill_used ) * 4;`
`    }`

If I understood the codes correctly, 0 dice means the roll can only be 0 (Yeah I don't have a die!), thus the return value is 0 if the difficulty roll > 0, thus the failure delta will be always 0.

PS: Now it is really painful to practice computer and social... Or there is a way to practice skills which I don't know?

### Steps To Reproduce

1. Make a player with 0 firstaid skill.
2. Read a book boosting knowledge level.
3. The player can never gain any progress crafting items using firstaid skill.
4. The player must spam using bandage to gain firstaid exp.

### Expected behavior

The player with 0 skill should have a chance to craft items with difficulty > 0. 

### Screenshots

_No response_

Current Master e816cb754606b874fbfc11735a7779e36851746c, vanilla game no mods.

### Additional context

_No response_

captainego:
#### Summary
Bugfixes "Characters with 0 skill level can make progress on difficulty level 1 recipes"

#### Purpose of change
Fixes #50787.

Previously a character with skill level 0 would automatically fail any attempt to craft a level 1 or higher recipe.

#### Describe the solution

Change the skill and difficulty dice calculations to be 1 indexed instead of 0 indexed. For example a character with skill level 0 attempting a level 1 recipe will now have 1 skill die and 2 difficulty dice (previously this would have been 0 skill dice and 1 difficulty die).  Another example a character with skill level 5 attempting a level 6 recipe will now have 6 skill dice and 7 difficulty dice (previously this would have been 5 skill dice and 6 difficulty dice).  This change allows the difficulty of crafting checks to be maintained, but they are shifted down 1 level.

I considered (and implemented) a solution that would give a character with skill level 0 an effective 1/4 level of skill when crafting.  This option felt too difficult compared to the other crafting above your level checks so I implemented the solution in this PR.

#### Testing
Create a character with 8 Intelligence and 0 levels in all skills
Use the debug menu to give the character knowledge of all recipes
Spawn 50 medical gauze, 250 medical tape, and 250 antiseptic.  Place all of these items on the ground for your sanity.
Attempt to craft bandage (a level 1 recipe that requires no proficiency).  
Verify that you can craft the bandage (even though you will likely have several failures along the way).
Use the debug menu to give the character level 5 in all skills.
Spawn 10 RX12 jet injectors and 100 duct tape.
Attempt to craft wearable RX12 jet injector ( a level 6 recipe that requires no proficiency).
Verify that you can craft the wearable RX12 jet injector (even though you will likely have several failures along the way).

### DIFF
--- a/src/crafting.cpp
+++ b/src/crafting.cpp
@@ -1013,19 +1013,32 @@ double Character::crafting_success_roll( const recipe &making ) const
     if( has_trait( trait_DEBUG_CNF ) ) {
         return 1.0;
     }
+
+    // Adjust skill and difficulty such that the lowest level is 1 not 0
+    // This allows characters with 0 level skills to attempt higher level recipes
+    int primary_skill_level = get_skill_level( making.skill_used ) + 1;
+    int primary_difficulty = making.difficulty + 1;
+
     int secondary_dice = 0;
     int secondary_difficulty = 0;
     for( const auto &pr : making.required_skills ) {
-        secondary_dice += get_skill_level( pr.first );
-        secondary_difficulty += pr.second;
+        // Adjust skill and difficulty such that the lowest level is 1 not 0
+        // This allows characters with 0 level skills to attempt higher level recipes
+        secondary_dice += get_skill_level( pr.first ) + 1;
+        secondary_difficulty += pr.second + 1;
     }
 
     // # of dice is 75% primary skill, 25% secondary (unless secondary is null)
     int skill_dice;
     if( secondary_difficulty > 0 ) {
-        skill_dice = get_skill_level( making.skill_used ) * 3 + secondary_dice;
+        skill_dice = primary_skill_level * 3 + secondary_dice;
     } else {
-        skill_dice = get_skill_level( making.skill_used ) * 4;
+        skill_dice = primary_skill_level * 4;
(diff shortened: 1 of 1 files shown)


anothersimulacrum:
I would rather we have https://github.com/CleverRaven/Cataclysm-DDA/pull/46153 to accomplish this

PR #61981, opened 2022-10-28
Aquatic land-based specials no longer float

MNG-cataclysm:
#### Summary
Bugfixes "Aquatic specials consisting of land should no longer float"

#### Purpose of change
Several aquatic overmap specials had no land underneath them, leaving them to float on the surface. I aim to fix that.
Closes #61954 

#### Describe the solution
I added soil and rock underneath several specials, including the island prisons, lake cabins, islands, and lighthouses.

#### Describe alternatives you've considered
Leaving them to drift on open water.

#### Testing
I added the code and spawned several locations via the debug menu. Everything spawned as normal.

#### Additional context
None

PR #61988, opened 2022-10-28
ci: give labeler workflow write permissions

casswedson:
#### Summary
None

#### Purpose of change
see https://github.com/CleverRaven/Cataclysm-DDA/pull/61981#issuecomment-1295626349 so the labeler ain't working due to missing permissions, best case scenario we just need to add them manually to the workflow; this pr adds them that would make some sense, but I don't trust github so much, I am gonna give it an honest shot

this should be all we need <https://github.com/actions/labeler#create-workflow>

#### Describe the solution
let it read the contents of prs (it already did that) and write to prs
```yml
    permissions:
      contents: read
      pull-requests: write
```

#### Testing
this pr, maybe - github I hate that this works on my fork, just cause it's my fork

so like I need at least another person to test for real, for real

or maybe another account for testing permission stuff?

#### Additional context
gonna revert everything if this still ins't working, this #61946 and #61904

### DIFF
--- a/.github/workflows/labeler.yml
+++ b/.github/workflows/labeler.yml
@@ -4,6 +4,9 @@ on: [pull_request]
 
 jobs:
   triage:
+    permissions:
+      contents: read
+      pull-requests: write
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3


casswedson:
## oh well
https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/3349130284/jobs/5548856805 the file there has the permissions set up
doesn't work at all, I guess I could look into updating the action's code some other time, for now let's just revert this baby

PR #61996, opened 2022-10-29
Check for character's electric immunity before applying EMP blast effects

Night-Pryanik:
#### Summary
Features "Check for character's electric immunity before applying EMP blast effects"

#### Purpose of change
Make electricity-immune characters immune to EMP-blast-related effects.

#### Describe the solution
Added check for electricity immunity.

None.

#### Testing
Got Dielectric Capacitance System bionic and activated it. Got MP3-player, Spawned incandescent husk and made him attack me. My MP3-player was safe and sound.

#### Additional context
None.

### DIFF
--- a/src/explosion.cpp
+++ b/src/explosion.cpp
@@ -739,7 +739,7 @@ void emp_blast( const tripoint &p )
         }
     }
     if( player_character.posx() == p.x && player_character.posy() == p.y &&
-        player_character.posz() == p.z ) {
+        player_character.posz() == p.z && !player_character.is_elec_immune() ) {
         if( player_character.get_power_level() > 0_kJ ) {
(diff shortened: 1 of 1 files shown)


Night-Pryanik:
I admit: I'm not keen on electrics or electronics, so I'm not totally sure whether _electric_ immunity should be extended to _electromagnetic_ immunity. So this PR has more gameplay reasons rather than realism reasons.

PatrikLundell:
I don't see how a Faraday cage system in your body would protect things you carry (assuming the Dielectric Capacitance System provides a cage, which isn't unreasonable), unless it's placed specifically in a weird inside body "pocket" that at least has existed as a bionic module.

PR #61998, opened 2022-10-29
Player can no longer make zombie Jell-O

Daved27hundred:
#### Summary
Content "Removed tainted hide as a possible ingredient in gelatin"

#### Purpose of change

Tainted hide is a possible ingredient in the recipe for acid soaked hide, which is the first step in making gelatin for use in aspic/gelatin dessert. I'm not entirely sure that eating something from a zombie is a good idea. This seems like an oversight when gelatin was first added so I'm fixing it.

#### Describe the solution

Removed tainted hide from the craft recipe.

If there is an actual reason for it being in the recipe let me know, I can close this PR.

### DIFF
--- a/data/json/recipes/food/other.json
+++ b/data/json/recipes/food/other.json
@@ -122,7 +122,7 @@
     "components": [
       [ [ "formic_acid", 10 ] ],
       [ [ "water", 10 ], [ "water_clean", 10 ] ],
-      [ [ "raw_leather", 1 ], [ "raw_tainted_leather", 1 ], [ "raw_hleather", 1 ], [ "raw_demihumanleather", 1 ] ]
+      [ [ "raw_leather", 1 ], [ "raw_hleather", 1 ], [ "raw_demihumanleather", 1 ] ]
     ]
   },
   {


RenechCDDA:
An alternative approach might be to remove the `NUTRIENT_OVERRIDE` tags from that chain of crafts and give tainted hide toxins so that gelatin made this way would contain toxins. However nutrient inheritance means that *all* nutrients are always inherited, which might not be appropriate for the resulting gelatin. I believe raw hide is also edible (while a very unappetizing choice, which would unfortunately mean a change in calories along the line.

I think removing tainted hide from the acid soaked hide recipe (your solution) is probably the best choice.

Maleclypse (approved):

PatrikLundell:
Does gelatin have any use beyond being a food ingredient? Mutant fat ->lard/tallow can be used e.g. for tanning as an example of a corresponding case (but without any override, so the toxin is carried over).

If there are non-food usages for gelatin then the addition of a tainted gelatin recipe is probably needed (it's probably better than to remove nutrient override), as being able to use otherwise useless resources for non-food usages is rather useful, and real hides is a somewhat scarce resource.

Daved27hundred:
there aren't any non-food uses for gelatin ingame, and searching up "uses for gelatin" only brings up food related things.

PR #62000, opened 2022-10-29
[Test] Update labeler version

RipleyTom:
#### Summary
None


casswedson:
changing the event trigger to pr_target would not fix this https://github.com/CleverRaven/Cataclysm-DDA/pull/61946#issuecomment-1295873433

PR #62010, opened 2022-10-29
Standardize hairstyle names

ryanaldrich:
Change all instances of "Hairstyle:" to "Hair:" in order to fix list ordering in chargen and save on UI space.

#### Summary
None

#### Purpose of change

A small follow-on to #61940 - this simply changes the hairstyle trait names to match the existing hair trait names. So you'll see "Hair: black, buzz cut" instead of "Hairstyle: black, buzz cut"

This makes the hairstyle traits correctly sort together in the character creation menu, and is a bit easier to read, in my opinion.

#### Describe the solution

As above.

N/A

#### Testing

Manually fired it up and checked the character creation menu, and followed #61940's approach of spawning a bunch of NPCs to see if they were still getting hairstyles.

#### Additional context

I'm not sure what tileset they used, because none of the ones I tried actually displayed these hairs (before or after this change). This has resulted in much higher ratio of visually bald NPCs, at least until someone fixes the tilesets.

### DIFF
--- a/data/json/mutations/appearance_hair_styles.json
+++ b/data/json/mutations/appearance_hair_styles.json
@@ -2,7 +2,7 @@
   {
     "type": "mutation",
     "id": "buzzcut",
-    "name": "Hairstyle: buzz cut",
+    "name": "Hair: buzz cut",
     "points": 0,
     "description": "Your hair has been shaved close to the head.",
     "starting_trait": true,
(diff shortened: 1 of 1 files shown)


MNG-cataclysm (approved):

Fris0uman:
MSX has the sprites, but it's not in mainline yet. The tileset update should happen automatically at some point today.

PR #62017, opened 2022-10-30
Hub 01 armor adjustment

Dan1ss1mo:
#### Summary
Balance "Hub 01 armor weight adjustment"

#### Purpose of change
Curant weigh of hub 01 armor parts doesn't make much sense. Mantle is ridiculously light, while most other parts are to heavy

#### Describe the solution
I recalculated armor weight based on armor thickness, materials dencity and body area coverage. I used 1.9m^2(19000cm^2) as total body area. Torso is roughly 33% of that, arms 15, legs 23 nad head 7. I also changed materials and thickness a bit, as resulting weight was too high. More specifically i changed total thickness from 8mm to 7. I reduced prototype armor plastic layer  thickness from 6 to 5 and replace 2mm of steel with 1mm of tempered steel for other variants 
example of calculations for prototype mantle: ((19000x0.33x0.81)+(19000x0.15x0.53))x(1x0.5+2.7x0.2)=6853 g. In terms of protection - it is -1 cut and bash and +1.5 ballistic for non proto and -2 all around for proto

#### Testing
Works as intended 

#### Additional context

[hubArmor.docx](https://github.com/CleverRaven/Cataclysm-DDA/files/10309798/hubArmor.docx)


RAldrich:
Changing the material thickness will have changed the actual armor protection values. What are the old values / new values?

You might want to open an issue to discuss the idea of adding a sling and/or webbing. I don't like it, but don't want to block this PR on that debate. (I would be much more supportive of adding a new "cargo module insert" for the hub 01 system that's less armor and more utility gear)

Dan1ss1mo:
> Changing the material thickness will have changed the actual armor protection values. What are the old values / new values?
> 
> You might want to open an issue to discuss the idea of adding a sling and/or webbing. I don't like it, but don't want to block this PR on that debate. (I would be much more supportive of adding a new "cargo module insert" for the hub 01 system that's less armor and more utility gear)

Sure, i'll open an issue about this. About protection: it is -2 bash, cut and ballistic for prototype, -1 cut and bash +1.5 ballistic for all others

bombasticSlacks (changes requested):
I'm pretty sure the mantle and skirt will no longer fit in their respective pockets

Dan1ss1mo:
> I'm pretty sure the mantle and skirt will no longer fit in their respective pockets

yeah, will upadate pockets on mds

bombasticSlacks (approved):

John-Candlebury (approved):

Sumlord-047:
The steel material of soldier mantle has not been changed to tempered carbon steel.

Dan1ss1mo:
> The steel material of soldier mantle has not been changed to tempered carbon steel.

Damn it, my bad, thanks for pointing that out.

PR #62018, opened 2022-10-30
Mutagen changes for clarity and consistency

actually-a-cat:
#### Summary
Balance "Basic mutagen is now worse than primers. Also, many mutagen and mutagen-related things renamed for clarity."

#### Purpose of change
The word "mutagen" is terribly overloaded and can refer to:

* The general concept of stuff that makes you mutate
* "mutagen" item
* "(type) mutagen" items
* "mutagen" vitamin
* mutation-related vitamins in general

There is also a conflict between primer items and primer vitamins. 

This cases interminable confusion in trying to get across information about this complex system. Every conversation either becomes incomprehensible or stuffed full of hedging and clarifications. I can't even imagine what this nonsense sounds like to new players trying to read a guide.

#### Describe the solution
I am removing the term "mutagen" from specific in-game names entirely to reserve it for reference to the entire system. Drinkable mutagens are now "mutagenic slurry". Mutation vitamins are "mutation factors".

* "mutagen" (item) -> "raw mutagenic slurry"
* "(type) mutagen" (item) -> "(type) mutagenic slurry"
* "mutagen" (vitamin) -> "catalytic mutation factor"
* "(type) primer" (vitamin) -> "(type) mutation factor"
* "mutagenic slurry" (vitamin) -> "impure mutation factor"

plus description changes to make it clearer what is what. 

I'm also changing typed mutagens (typed mutagenic slurry) to be actually inferior to primers, right now they are just as effective and actually more efficient despite being a "lower-tier" item.

Migrating the ids as well. Should I?

#### Testing
Loaded and looked at the items (and their vitamin contents) in item spawn menu.
Looked at vitamin list.


data/json/items/comestibles/mutagen.json
```diff
@@ -3,7 +3,7 @@
     "abstract": "mutagen_flavor",
     "type": "COMESTIBLE",
     "name": { "str": "abstract mutagen flavor" },
-    "description": "A rare substance of uncertain origins.  Causes you to mutate.",
+    "description": "A slurry of organic chemicals and compounds, in half-processed form.  In high enough dose, it will cause you to mutate, but it would be safer and more predictable if further refined into a primer.",
```
Maleclypse, line 6: OK so I've been focused in other places  but the sentence below implies that if possible the player should only use primers to mutate.  Is that currently true?   Last I checked you needed both the primer and the base mutagen vitamins to mutate or you needed just base mutagen vitamins?   It's cool if this has changed but I want to make sure of it before we make statements like the one below.
```suggestion
    "description": "A slurry of organic chemicals and compounds, in half-processed form.  In high enough dose, it will cause you to mutate, but it would be safer and more predictable if further refined into a primer.",
```

data/json/items/comestibles/mutagen.json
```diff
@@ -28,7 +28,7 @@
     "copy-from": "mutagen_flavor",
     "type": "COMESTIBLE",
     "name": { "str": "abstract iv mutagen flavor" },
-    "description": "A processed mutagenic primer.",
+    "description": "A processed mutagenic primer.  Combined with a mutagenic catalyst, this will cause you to mutate in a controlled fashion… relatively speaking.",
     "price": "3000 USD",
```
Maleclypse, line 32: See these lines say "combined with" which conflicts with the line at the top.

data/json/items/comestibles/mutagen.json
```diff
@@ -460,302 +464,295 @@
...
+    "name": { "str_sp": "alpha mutagenic slurry" },
     "price": "5000 USD",
     "looks_like": "mutagen",
     "//": "*May* have gotten a dose smuggled out.",
     "use_action": {
       "type": "consume_drug",
-      "activation_message": "You drink the alpha mutagen.",
-      "vitamins": [ [ "mutagen_alpha", 225 ], [ "mutagen", 125 ] ]
+      "activation_message": "You drink the alpha mutagenic slurry.",
+      "vitamins": [ [ "mutagen_alpha", 225 ], [ "mutagenic_slurry", 15 ] ]
```
Maleclypse, line 474:
```suggestion
      "vitamins": [ [ "mutagen_alpha", 225 ], [ "mutagenic_slurry", 125 ] ] did you intend to go from 125 to 15 mutagenic_slurry vitamins?  It looks like most of them have this error if it is an error. 
```

Maleclypse (commented):
Your changes in vitamin I need to review sometime with a look at how the current system works because I'm not certain what the difference is between mutagen and mutagenic_slurry vitamins are currently.

data/json/items/comestibles/mutagen.json
actually-a-cat: You are correct that you can't use primers only to mutate, but this is the description for slurry, not for primer. Slurry (eventually) provides both catalytic factor and typed factor so it will cause mutations on its own, although it's inefficient and unhealthy to do that (it also increases toxins and instability).

data/json/items/comestibles/mutagen.json
actually-a-cat: Intentional. 1 `mutagenic_slurry` decays into 8 `mutagen` (among other things), so in terms of that it's close (125 vs 120).

actually-a-cat:
@Maleclypse 

To generally explain what the deal with `mutagenic_slurry` is - it uses the new vitamin decay feature introduced in #61036. Over time, 1 unit of `mutagenic_slurry` will turn into a variety of things:

```    
"decays_into": [
      [ "mutagen", 8 ],
      [ "mutagen_bird", 1 ],
      [ "mutagen_cattle", 1 ],
      [ "mutagen_feline", 1 ],
      [ "mutagen_fish", 1 ],
      [ "mutagen_gastropod", 1 ],
      [ "mutagen_insect", 1 ],
      [ "mutagen_lizard", 1 ],
      [ "mutagen_lupine", 1 ],
      [ "mutagen_medical", 1 ],
      [ "mutagen_mouse", 1 ],
      [ "mutagen_rat", 1 ],
      [ "mutagen_troglobite", 1 ],
      [ "instability", 1 ],
      [ "mutant_toxin", 2 ]
    ]
```

Ilysen:
so I spent a little while thinking about this when I saw it, and I have some concerns about it I wanted to bring up.
1. why nerf the typed mutagens? the advantage of the injected variants is that they're much more cost-efficient and allow greater control. you can achieve the same thing with liquid mutagens, but it will take a lot more time and materials to make it happen. replacing the mutagen vitamin with the mutagenic slurry vitamin makes them effectively never worth using due to the instability tanking as a result. I'm not sure why making them basically useless is necessary here?
2. with regards to consistency, I believe this PR is a regression in its current state. "mutation factor" feels a lot less clear than "primer" to me; a primer is, of course, something that primes you, but what would a new player assume a mutation factor does? the only thing I think might increase clarity here is unifying the name of the liquid mutagens, but at that point, why not just make the baseline mutagen just give straight mutagen vitamin instead of its own handling, to unify the functionality? the fact that maleclypse was thrown off by your changes just reinforces to me that it makes the system more unclear, not less so.

I'm not sure if writing all this out will have any effect or do anything, but it's concerns I had and I wanted to voice them. I think that some of these changes are probably a net negative for the stated goal of system clarity, :v

actually-a-cat:
@Ilysen

> why nerf the typed mutagens? the advantage of the injected variants is that they're much more cost-efficient and allow greater control. you can achieve the same thing with liquid mutagens, but it will take a lot more time and materials to make it happen;

This is not currently the case. The primer recipe is 2 typed mutagens + sample. 2 basic mutagens is 450 typed mutagen and 250 generic mutagen. 1 primer is 500 typed mutagen (average) and no generic mutagen. By making a primer you spend time and an additional sample to get less effect overall, with no significant upside. Besides possibly being able to get typed mutagen without any generic, but I don't see when it would be relevant. So, currently, making primers is quite pointless.

This wasn't a glaring problem when drinkable mutagens had to be digested before taking effect, which made them annoying to use and wasteful, but you removed that.

> replacing the mutagen vitamin with the mutagenic slurry vitamin makes them effectively never worth using due to the instability tanking as a result. I'm not sure why making them basically useless is necessary here?

Some players will be happy to eat whatever funny juice they can get just to see what happens, and if it's less predictable that's just a bonus for them. Those that *do* want optimal, predictable results will have to put in the effort to make primers, which I see as a win.
I'm not married to this particular solution, buffing primers is on the table too, or maybe something entirely different, I just want making primers to make sense. Although I do think that it's fitting that the most basic mutagen you can make would be pretty horrible for you, having the opinion that mutagen-making is too easy and simple overall.

> with regards to consistency, I believe this PR is a regression in its current state. "mutation factor" feels a lot less clear than "primer" to me; a primer is, of course, something that primes you, but what would a new player assume a mutation factor does?

"Mutation factor" is vague by necessity. I want a term that can be used to unambiguously refer to mutation vitamins. At the same time, there are three types of those, that do entirely different things, so there is no way it can be particularly specific. If you have a better idea that satisfies these requirements I'll be happy to use it.
"Primer" is an excellent word, but it's already used for the primer items. Having to specify whether you are talking about an item, or a vitamin, when you say "primer", is the kind of trouble I would like to avoid.

Your thoughts are appreciated by the way, even if I disagree.

Ilysen:
why not reduce the primer present in liquid mutagens? I often see the archetype of "player who drinks any mutagen they find" thrown around in discussions, but I've never actually met anyone that actually does that; that's not to say they don't exist, but at the least, they don't frequent the spaces I'm in. instability cannot currently be managed or interacted with in a meaningful way since it was left in an unfinished state, so I think any kind of things that add it to the player need to be looked at with an abundance of care, as it's extremely punishing to gain instability that you don't need to have, yknow?

I still think primer as a name is fine, honestly. you could change the name and descriptions of the injected items to mention that it gives a dose of primer, instead of the item itself *being* the primer, but I'm personally very averse to the idea of changing the terms for a system that's only now having people start to get used to it. I think any user-facing aspect for a major system should very much be the opposite of "deliberately vague" unless there's a good reason for it, and aiming for clarity seems like the opposite of a good reason for being deliberately vague. :p

actually-a-cat:
I would prefer primers to be qualitatively different from slurries (to use my newfangled words). "Safer+more predictable" is a more interesting reward for climbing up the crafting tree than "does the same thing but more". 

I thought a bit more about the primer situation and I'll agree that it's not the worst. The primer is in the syringe, you inject it, you have primer in your blood, that's fine. But I also do think we need a common name for mutation-related vitamins. I don't know which is more important. 

> I'm personally very averse to the idea of changing the terms for a system that's only now having people start to get used to it

I see your point but it will never be any easier to change than it is now, while it's still not in any stable release.

Ilysen:
we may have to agree to disagree on this one, then. if nothing else, all I can encourage you to do is to give the changes here a shot in regular playthroughs to see how it shakes out.

actually-a-cat:
We'll agree to disagree but you've convinced me enough that I'll think about other approaches to the vitamin names (and thank you for that)

fairyarmadillo:
Perhaps the mechanical changes and the nomenclature changes should be split into separate PRs, as the former is a fairly sticky subject.

I would suggest something other than "catalytic mutation factor" for the mutagen vitamin, as it's confusing that it's named identically to both the typed ones and the generic one. Even just like "mutagenic catalyst" would make it clearer when trying to explain to someone how it works.

Also, maybe there should be in-game documentation  explaining what the names mean along with dosing guidelines. Maybe in a post-it or as part of the description of Best Practices for Compound Delivery and one of the Lab Journals, like Dionne?

PR #62024, opened 2022-10-30
Jsonize test data for vehicle drag and effective dps tests

mqrause:
#### Summary
Infrastructure "Jsonize test data for vehicle drag and effective dps tests"

#### Purpose of change

These tests regularly cause confusion in json contributors. Jsonizing the data means they don't have to touch c++ files anymore.

#### Describe the solution

Add a class for test data with static members to be accessed in the tests.
Add members for vehicle drag and effective dps and also move the known bad density data there.
Move the data from the tests to json.

Adding data for other tests is easy.

#### Testing

Explicitly ran the affected tests and all loaded their data and passed.


src/itype.h
```diff
@@ -1411,12 +1411,5 @@ struct itype {
 
 void load_charge_removal_blacklist( const JsonObject &jo, const std::string &src );
 void load_charge_migration_blacklist( const JsonObject &jo, const std::string &src );
-// can be removed once all known bad items got fixed
-class known_bad_density
```
Leland, line 1415: I'll note that we are losing some context from comments in a few places.

mqrause: Yeah, I judged the comment isn't really relevant enough when it's not it's own class anymore. I can readd it if people think it should still be there.

mqrause:
Oh yeah, forgot about that. I was also considering doing a full object instead of the array, but I'm definitely too lazy make the json for that.

ZhilkinSerg:
> Note on the vehicle test data json on what the numbers mean would be nice.

Test data is generated automatically, so it should not matter much.

src/itype.h
mqrause: Added a comment.

PR #62025, opened 2022-10-30
Added millwall brick

bean-b:
#### Summary
Content "Added millwall Brick"

#### Purpose of change
Wanted to add this item as a weak early game improvised weapon.

#### Describe the solution

Added the millwall brick, a improvised weapon only obtained through crafting (autolearned at level 0)

I'm not sure if this is acceptable as newspapers only represent singular sheets in this game, but there doesn't seem to be any item that represents a full newspaper so i figured it would be okay considering a millwall brick only needs a few sheets of newspaper? Alternatives of course include not adding this item, allowing it to be crafted with paper/newspapers etc.. Other versions of the millwall brick have a rock or a coins as added weights, but I figured I would go with the simplest version since this item is crafted out of convenience not power. I wasn't sure if it should have the plastic material tag, but with only paper it would break in about three hits, and while it shouldn't be durable, it was breaking way to quick without it. 

#### Testing

Crafted in game. Spawned in game. Bashed some zeds in with it in game.

#### Additional context
Research:
https://en.wikipedia.org/wiki/Millwall_brick
https://www.youtube.com/watch?v=y7dL2U7Ub2I

### DIFF
--- a/data/json/recipes/weapon/bashing.json
+++ b/data/json/recipes/weapon/bashing.json
@@ -879,7 +879,7 @@
         [ "weeks_old_newspaper", 1 ]
       ],
       [ [ "duct_tape", 10 ] ],
-      [ ["pebble", 5] ]
+      [ [ "pebble", 5 ] ]
     ]
   }
 ]


data/json/items/melee/bludgeons.json
```diff
@@ -1485,5 +1485,24 @@
     "to_hit": { "grip": "weapon", "length": "long", "surface": "line", "balance": "good" },
     "price": 16000,
     "qualities": [ [ "HAMMER", 1 ] ]
+  },
+  {
+    "type": "GENERIC",
+    "id": "millwall_brick",
+    "name": { "str": "millwall brick" },
```
TheShadowFerret, line 1492:
```suggestion
    "name": { "str": "Millwall brick" },
```
Millwall refers to Millwall Football Club, so it could be capitalized.

data/json/items/melee/bludgeons.json
```diff
@@ -1485,5 +1485,24 @@
...
+    "color": "light_gray",
+    "symbol": "/",
+    "looks_like": "cudgel",
+    "material": [ "paper", "plastic" ],
+    "techniques": [ "RAPID" ],
+    "weapon_category": [ "MACES" ],
+    "volume": "200 ml",
+    "longest_side": "30 cm",
+    "bashing": 10,
+    "to_hit": { "grip": "solid", "length": "short", "surface": "any", "balance": "neutral" }
```
TheShadowFerret, line 1506:
```suggestion
    "to_hit": { "grip": "none", "length": "hand", "surface": "any", "balance": "neutral" }
```

data/json/items/melee/bludgeons.json
```diff
@@ -1485,5 +1485,24 @@
     "to_hit": { "grip": "weapon", "length": "long", "surface": "line", "balance": "good" },
     "price": 16000,
     "qualities": [ [ "HAMMER", 1 ] ]
+  },
+  {
+    "type": "GENERIC",
+    "id": "millwall_brick",
+    "name": { "str": "millwall brick" },
+    "description": "An improvised club, made of a tightly rolled newspaper, folded with lengthwise and duct-taped. It packs a surprising punch for what is essentially paper.",
```
TheShadowFerret, line 1493:
```suggestion
    "description": "An improvised club, made of a tightly rolled newspaper, folded lengthwise and duct-taped. It packs a surprising punch for what is essentially paper.",
```

RAldrich:
I watched the video you linked, but 10 bashing for a bit of rolled up newspaper and tape? I just don't buy it - that's the same damage as a wooden club.

data/json/items/melee/bludgeons.json
```diff
@@ -1485,5 +1485,24 @@
...
+    "price_postapoc": 0,
+    "color": "light_gray",
+    "symbol": "/",
+    "looks_like": "cudgel",
+    "material": [ "paper", "plastic" ],
+    "techniques": [ "RAPID" ],
+    "weapon_category": [ "MACES" ],
+    "volume": "200 ml",
+    "longest_side": "30 cm",
+    "bashing": 10,
```
MNG-cataclysm, line 1505:
```suggestion
    "bashing": 3,
```
This should be more reasonable for rolled-up newspaper.

MNG-cataclysm (changes requested):

NetSysFire:
The reason they pack a punch is because they are usually filled with hard objects, like spare change. I think pebbles would work, too since there is no one searching you for weapons and they should be the right size.

bean-b:
I think modifying the recipe to require some weights, i.e. pebbles makes sense. In that case what would be a reasonable damage? I realize 10 was defintly too high for just the paper version, but would like 7 be good for the paper + pebble version?

bean-b:
Ok i changed the stats up to reflect the changes, I did end up changing the length back to short, because although the first google image result of them is very short, many are longer, i.e. 
https://www.reddit.com/r/weapons/comments/avuphz/made_a_millwall_brick_a_week_or_so_ago_its/

But can ofc change it back to hand to if that seems like it would be better

PR #60510, opened 2022-08-28
Make security bots not leave its position

Marcos6226:
#### Summary
Balance "make security bots not leave its position on the bank."

#### Purpose of change
This PR closes #59961 

Steps to test if the changes are working as intended:
1. Create a character in a big city or spawn a bank.
2. Check if the bank have security bots, if not, trigger them to spawn by failing to hack the vault computer.
3. Try getting them to follow you to the outside of the bank.
4. If they not follow you, them the fix is working.

#### Describe the solution
Thinking about pleasing both parts in the discussion, I've chosen to not remove them, but instead make them not leave the bank.
So in a similar manner that I did on the mi-go tower, I have added a `"patrol":` line so they are still present but will not wander around anymore.

#### Describe alternatives you've considered
I'm open to any suggestions

#### Testing
I have spawned a bank and triggered the bots attention. They had shot me, but they didn't get out of the building, as expected.

#### Additional context
none

PR #62039, opened 2022-11-01
spawn mon_copbot instead of mon_secubot in banks

### LINKED ISSUE #59961: Talon UGV killbots spawning on street overmap
### Describe the bug

Early game player spawning near a bank can be gunned down by a Talon UGV with an automatic rifle with no ability to predict or counter the instakill

### Steps to reproduce

Start a new world, spawn near a bank, explore the bank, see a killbot, run and get gunned down

### Expected behavior

I was surprised to see military grade robots with automatic weapons spawning day one in a normal bank. That's not typical in the banks I've seen.

### Screenshots

_No response_

Newest experimental on Mac, running DinoMod but not related

### Additional context

In general IMO, a day one instakill with no counter is not good outside a challenge start

anoobindisguise:
#### Summary
Balance "Banks deploy cop bots instead of TALON killbots when agitated"

#### Purpose of change
Fixes #59961
Fixes #60510
Banks spawning ED-209 was pretty dumb and also one of the easiest ways to die in a city (especially if you are unspoilered) when a zombie breaks the glass on a bank and causes it to deploy murderbots that take potshots at you, a random passerby, for daring to be in the presence of a criminal activity.

#### Describe the solution
Banks instead spawn less-lethal force to subdue criminals, in the form of policebots.

Completely remove any spawns whatsoever, just blare an alarm and do nothing else.
Spawn riot control bots instead of policebots. 

#### Testing
Looks like any newly generated banks will work properly, with no secubots as far as I can see and copbots spawning in their place.

### DIFF
--- a/data/json/mapgen/bank.json
+++ b/data/json/mapgen/bank.json
@@ -78,7 +78,7 @@
           "failures": [ { "action": "shutdown" }, { "action": "alarm" } ]
         }
       },
-      "place_monster": [ { "monster": "mon_secubot", "x": [ 7, 18 ], "y": [ 7, 18 ], "chance": 30, "repeat": 4 } ]
+      "place_monster": [ { "monster": "mon_copbot", "x": [ 7, 18 ], "y": [ 7, 18 ], "chance": 30, "repeat": 4 } ]
     }
   },
   {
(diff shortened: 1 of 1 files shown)


anothersimulacrum:
I don't really think it makes sense to be spawning any bots in banks.

bean-b:
> I don't really think it makes sense to be spawning any bots in banks.

I feel like sense we have wondering cop bots, spawning cop bots in banks makes enough sense. There is a larger discussion about removing all of the cop/eye bots etc.. Which is a seperate thing, but given that we have those bots, i think they make sense in banks

PR #62048, opened 2022-11-01
Updated Riot Armor to modern armor standards

bean-b:
#### Summary
Content "Updated Riot Armor to modern armor standards."

#### Purpose of change
The old riot armor made little sense, 55%/60% coverage all around, 1000g of weight, and as such was not up to modern armor standards and made little sense. Therefore, I updated the armor values to make more sense.

#### Describe the solution

The materials are now  plastic padding and thermoplastic resin for protection. The suit has varying protection values across it, as i tried to base the material thicknesses off of current riot armors. It also has a small amount of molle webbing(on the thighs) as i saw that that was common across many riot suits.

The riot arm/shin guards are very similar to the hard leg/arm guards and thus may be a bit confusing. Also any of the values could be changed of course. Currently it isnt repairable which im not sure is ideal, but that is because thermoplastic resin and plastic padding are not repairable, which seems out of the scope of this pr.

#### Testing

Spawned the item in game, saw defensive values, tested attaching and detaching molle. 

#### Additional context

Designed after this suit vaguely after:
https://www.securityprousa.com/collections/riot-gear/products/damascus-gear-modular-hard-shell-crowd-control-suit

### DIFF
--- a/data/json/items/armor/legs_armor.json
+++ b/data/json/items/armor/legs_armor.json
@@ -1610,8 +1610,8 @@
         "specifically_covers": [ "foot_arch_l", "foot_arch_r", "foot_ankle_r", "foot_ankle_l" ]
       }
     ]
-   },
-   {
+  },
+  {
     "id": "legguard_larmor",
     "type": "ARMOR",
     "category": "armor",
--- a/data/json/items/armor/torso_armor.json
+++ b/data/json/items/armor/torso_armor.json
@@ -1973,8 +1973,8 @@
         "specifically_covers": [ "leg_hip_l", "leg_hip_r" ]
       }
     ]
-   },
-   {
+  },
+  {
     "id": "armor_larmor_chest",
     "type": "ARMOR",
     "category": "armor",


Jarewill:
Since the armor now incorporates a jumpsuit, maybe it should also be on the normal layer alongside the outer layer as well as be padded/comfortable?

bean-b:
I'm pretty sure its already padded/comfortable given the materials within. But I could totatly make it normal + outer if that makes more sense given its components and that it sorta has a normal layer built in. Another option would be to model it after a version without the nylon jumpsuit underneath, i.e. https://www.securityprousa.com/collections/riot-gear/products/secpro-riot-suit-1
I saw about half and half of both styles of designs when I was researching, and so I think this could go either way.

bean-b:
I guess my main question with making it have both layers is what the difference between this vs say the track touring suit is. The track touring suit has an entire 1.3mm leather jumpsuit that holds all of the plastic padding etc, but only occupies the outer layer.

nopenoperson:
break up all the hard plastic into multiple pieces in a way that makes sense.  like shin guards (could have foot guards, knee pads that could be modular), groin protector ( with hip protection?)  chest protector (with protection for shoulders and upper arm? protective foam vest could be separate piece worn on normal layer), forearm protectors (with built in elbow pads or might be separate?) etc. 

having it be a single article of clothing makes no sense.

bean-b:
Yea i can do that

bean-b:
I'm pretty happy with how the armor turned out, but one other option I considered would be to make the unique chest + shoulder + hip guard, and have it combined with the existing hard arm + leg guards in order to form the riot suit.

anoobindisguise:
Only thing I could think to change is to bump up the plastic padding thickness a little. Other uses of plastic padding have 12-14mm.

bean-b:
Is there anything particularly wrong with less thick plastic padding layers? I know many of the motorcycle armors have a lottt of plastic padding, but I don't see any reason against having less on another piece of armor. I chose the plastic padding material because most riot suit advertise shock absorbing foam _tm_ or something along those lines, although the thicknesses didint seem insane as it was all underneath the thermoplastic resin. Would there be a sort of halway point of like 6-8mm? Because 12+mm would more then double armor values and probably increase encumbrance values as well.

anoobindisguise:
> Is there anything particularly wrong with less thick plastic padding layers? I know many of the motorcycle armors have a lottt of plastic padding, but I don't see any reason against having less on another piece of armor. I chose the plastic padding material because most riot suit advertise shock absorbing foam _tm_ or something along those lines, although the thicknesses didint seem insane as it was all underneath the thermoplastic resin. Would there be a sort of halway point of like 6-8mm? Because 12+mm would more then double armor values and probably increase encumbrance values as well.

There's absolutely a right number there that likely isn't 2mm or 12mm. Just depends on the design specs of RL armors.

bean-b:
@TheShadowFerret Could you review this when you get the chance? Thanks so much. Also if there is a better way to request reviews please let me know.

bean-b:
or @MNG-cataclysm if you could review that would be great!

MNG-cataclysm (approved):

PR #62049, opened 2022-11-01
ci: more labeler rules

casswedson:
#### Summary
None

#### Purpose of change
so like this makes the labeler less broad in a sense specially with the ui label, as in don't just mark things as ui if the file happens to have `ui` in it, this one for example:
`data/json/items/armor/suits_protection.json`

after a little bashing of the head against a wall I remembered that it would be beneficial to be specific, here I added a bunch of "verbose" regex rules

the "originals" were something like `**/**field**`: match stuff that happen to have field in the file name, as you know it may not be case always so I added `**/field**` this would cover directories starting with the search as well as files that start with the search, something the first one didn't do that well

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -9,10 +9,13 @@
   - "**/*.md"
 
 "Game: Achievements / Conducts / Scores":
-  - "**/**scores**"
+  - "**/scores**"
   - "**/**statistics**"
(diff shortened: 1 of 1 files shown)


bean-b:
Haha nice, I was wondering why my recent pr that modified armor suits got the ui tag

PR #62052, opened 2022-11-02
acorn fixes

### LINKED ISSUE #63743: Cooked Acorns Too Big
### Describe the bug

4 acorns are .25 liters. 1 cooked acorn is .25 liters.

### Attach save file

N/A

### Steps to reproduce

...cook acorns?

### Expected behavior

Equal volume for cooked acorns and non cooked acorns.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19044.2604 (21H2)
- Game Version: 95521e0 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm],
    SpeedyDex [speedydex],
    My Sweet Cataclysm [my_sweet_cataclysm]
]

### Additional context

_No response_

flybye3o6:
fixes some discrepancies with acorns

#### Summary
Bugfixes "fixes some size weight and value discrepancies with all the acorns"

#### Purpose of change
stat differences between the acorns didn't make sense, cooked acorns gained 0.19L extra volume despite the description saying they are dried, blanching uses 2 units of water but only has the volume of 0.25L and the price value varies wildly with blanching being worth $2 but it turns into blanched is only $0.50
Fixes #63743

#### Describe the solution
changed the values to line up with others

- added copy from acorns to cooked acorns, fixes weight and volume
- added weight and spoil time to roasted to be the same from cooked
- blanching made to look like a liquid, price was reduced, weight and volume increased to reflect the 2 unites of water
- blanching recipe reduced by x10 now only needing (2 unites of acorns, 4 unites of water, 3 minutes and results 2 charges) and given a batch time saving(the acorns are already a handful not individual acorns, if a handful is about 3-5, 6-10 should be enough to start the process also likely you will be doing more at the same time anyway, most of the craft time would be hulling so only a small batch time saving)
- added charges to cooked acorn recipes so they dont produce more then what goes in due to cooked copying its data from acorns and adding 4 charges to it

#### Additional context
i want to change the blanching process in the future so the tannin infused water can also be recovered and used for tanning


flybye3o6:
acorns cooked
acorns blanching

evankimori:
Question: still having an issue with Blanching into Blanched Acorns and recipes that require it in **936b383**. Should I post issue here or create a separate Bug ticket?

anothersimulacrum:
Create a separate bug report, comments on merged PRs generally get lost

PR #62055, opened 2022-11-02
Add 1 tile radius to store ebook to device

tivec:
#### Summary
Features "Allow ebook storage from books in adjacent tiles"

#### Purpose of change
Storing books in your smartphone or similar device is great, but i always got a bit annoyed that you have to pick up the ones you should store. Why can't I take a book directly out of the bookcase and flip through the pages while taking photos?

#### Describe the solution
This adds two variants to titled_filter_menu that allow for a radius to be passed. It is currently used by the iuse method `ebooksave`, but can be adapted to other usages.

Pain and suffering having to pick up my books all the time?

#### Testing
Tested with books in hand, on ground, in furniture, and in vehicles.

### DIFF
--- a/src/game_inventory.h
+++ b/src/game_inventory.h
@@ -118,7 +118,7 @@ item_location gun_to_modify( Character &you, const item &gunmod );
 /** Book reading menu. */
 item_location read( Character &you );
 /** eBook reading menu. */
-item_location ebookread( Character &you, item_location &ereader ); 
+item_location ebookread( Character &you, item_location &ereader );
 /** Menu for stealing stuff. */
 item_location steal( avatar &you, Character &victim );
 /** Item activation menu. */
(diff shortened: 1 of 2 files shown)


mqrause:
It's probably fine to use `PICKUP_RADIUS` (=6) instead of the radius of 1 for stuff like this.
And just personal preference, but you could add the radius as a parameter with a default value of -1 instead of adding overloads.

tivec:
> It's probably fine to use `PICKUP_RADIUS` (=6) instead of the radius of 1 for stuff like this. And just personal preference, but you could add the radius as a parameter with a default value of -1 instead of adding overloads.

I feel maybe PICKUP_RADIUS is a bit much (if it is 6) since you're not actually picking up the book to scan it.

For default parameter, I wanted to keep the position of radius relevant to the avatar, and not the messages, which would block default values. However, I see that the `inv_internal` method has the position just before `none_message`, so I'll make it a default value parameter to the original method.

tivec:
Not entirely happy with having to change other implementations to use -1 for range (which would have been default anyway) when they use a none_message, but there's really no other option.

PR #62073, opened 2022-11-04
Fix Nonrigid nested container can't store picked up items

### LINKED ISSUE #62068: Nonrigid nested containers can't store picked up items
### Describe the bug

I have a high priority gallon zip plastic bag I set to accept soap. I pick up a stack of soap. The soap is instead distributed among a variety of pockets instead of the empty high priority gallon bag. I would understand if it at least partially/totally filled the bag and distributed the remainder, but none ends up in the gallon bag.

### Steps to reproduce

- Spawn a nonrigid container like a plastic zipper bag, as well as a wearable container like a backpack.
- Wear the backpack, put the plastic bag in the backpack.
- Set the priority on the bag to a high value
- (Optionally) set an item whitelist for a specific item that fits in the bag.
- Spawn soap and put it on the ground.
- Pick up the soap.

### Expected behavior

Soap goes in the plastic bag not into the backpack.

### Screenshots

_No response_

- OS: Windows
- Game Version: 0.F-12257-g579e201ecc-dirty [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

EIIKaO:
#### Summary
Bugfixes "Nonrigid nested containers can't store picked up items"

#### Purpose of change
- Fix #62068
Improved to collect all pockets, even nested non-rigid pockets, and then fill them with items in the order of priority.

#### Describe the solution
In current codes, nested non-rigid containers did not contain picked up `count_by_charges` item**s**.
Improved the method to collect all pockets, even nested non-rigid pockets, calculate capacity limits by the parent-child relationship, and then fill the items based on priority.

None

#### Testing
In-game testing done.

Repro method by the issue #62068
- Spawn a nonrigid container like a plastic zipper bag, as well as a wearable container like a backpack.
- Wear the backpack, put the plastic zipper bag in the backpack.
- Set the priority on the bag to a high value
- (Optionally) set an item whitelist for a specific item that fits in the bag.
- Spawn soap and put it on the ground.
- Pick up the soap.
- [**NEW**] Confirm that the soap were put into the plastic zipper bag.

#### Additional context
Thanks to @akrieger, his report, which accurately pointed out this problem, struck a chord with my laziness.

### DIFF
--- a/src/character_attire.cpp
+++ b/src/character_attire.cpp
@@ -2283,11 +2283,7 @@ std::vector<pocket_data_with_parent> Character::get_all_pocket_with_parent(
     }
 
     for( item_location &loc : locs ) {
-        item *i = loc.get_item();
-        if( !i ) {
-            continue;
-        }
-        for( const item_pocket *pocket : i->get_all_contained_pockets() ) {
+        for( const item_pocket *pocket : loc->get_all_contained_pockets() ) {
             std::vector<pocket_data_with_parent> child =
                 get_child_pocket_with_parent( pocket, item_location::nowhere, loc, 0, filter );
             ret.insert( ret.end(), child.begin(), child.end() );


mqrause:
Can't do a full review right now, but it looks like most if not all you're doing there could be done with `item_location` already.

EIIKaO:
@mqrause Thanks for advice.
This is actually code I wrote about two months ago.
Since then I learned a bit more about item_location too. ( I haven't PR'd it yet, but the following content, related to the feature that allows follower to read e-books: https://github.com/EIIKaO/Cataclysm-DDA/commit/8d3f274bb9bfd0db2598d8b9c4d9fbfb76513393 )

I would like to go over the code again myself, but it would be even more helpful if you could indicate more specific points( or direction ) of improvement .

mqrause:
I'll have to look at it a bit closer this weekend, but as a quick and hopefully helpful pointer:
You can build the top level outfit locations with `item_location parent_location( guy, pointer_to_worn_item )` and then go deeper level by level with `item_location content_location( parent_location, pointer_to_content_item )`. With these you can call `item_location::parents_can_contain_recursive` and `item_location::max_charges_by_parent_recursive` to check if/how much of something really fits inside a nested container.

EIIKaO:
Following mqrause's advice, I adopted a style that utilizes item_location.
Now uses item_location's function to allow check if/how much of something really fits inside a nested container.

Unfortunately, my poor knowledge and skills, have resulted in the generation of code that is not very readable.

anothersimulacrum:
Would you be able to add a test for this case, so we can prevent bugs like this?

EIIKaO:
> Would you be able to add a test for this case, so we can prevent bugs like this?

I am interested but it may be difficult for me to make it.
One reason is that I do not have an environment to build the tests.

> You may also be able to improve on readability by using a struct instead of a tuple.

I will try this one quickly.
Thanks you for your advice.

src/character_attire.cpp
```diff
@@ -2230,6 +2230,61 @@ void outfit::pickup_stash( const item &newit, int &remaining_charges, bool ignor
...
+                    ret.emplace_back( pock_d );
+                }
+            }
+        }
+        ret.emplace_back( pocket_data );
+    }
+    return ret;
+}
+
+std::vector<pocket_data_with_parent> outfit::get_all_pocket_with_parent(
```
mqrause, line 2261: Since this is checking worn and wielded items, it should be `Character` scope, not outfit. I suggest moving the sorting and filtering here, so it's kind of like an advanced `best_pocket`.

src/character_attire.cpp
```diff
@@ -2245,52 +2300,71 @@ void outfit::add_stash( Character &guy, const item &newit, int &remaining_charge
...
+            if( lpd->settings.priority() != rpd->settings.priority() ) {
+                return lpd->settings.priority() > rpd->settings.priority();
+            }
+            if( lp_nested_lvl != rp_nested_lvl ) {
+                return lp_nested_lvl < rp_nested_lvl;
+            }
+            if( lpd->rigid() != rpd->rigid() ) {
+                return lpd->rigid();
+            }
+            return rpd->better_pocket( lpp, temp_it, false );
```
mqrause, line 2336: I assume this is all done just for the nesting level preference? While that is certainly nice to have, I don't think it warrants bypassing `better_pocket` like this, because whenever that gets touched this would need to be checked as well. You either need to work it into `better_pocket` somehow or ignore it.

As an aside, `better_pocket` currently has a slight bug with it's fallback `obtain_cost` check. It currently can't check the true `obtain_cost` since it isn't aware of any parent pockets. If that was properly checked, there would be no need to worry about the nesting level at all, because deeper nested pockets are naturally slower to access and would lose in a comparison. But that's definitely not something to fix in this PR.

src/character_attire.cpp
```diff
@@ -2230,6 +2230,61 @@ void outfit::pickup_stash( const item &newit, int &remaining_charges, bool ignor
...
+        items.emplace_back( &i );
+    }
+    for( item *i : items ) {
+        for( const item_pocket *pocket : i->get_all_contained_pockets() ) {
+            item_location loc = item_location::nowhere;
+            std::vector<pocket_data_with_parent> child =
+                get_child_pocket_with_parent( pocket, loc, item_location( guy, i ), 0 );
+            for( const pocket_data_with_parent &pock_d : child ) {
+                ret.emplace_back( pock_d );
+            }
```
mqrause, line 2282:
```suggestion
            ret.insert( ret.end(), child.begin(), child.end() );
```

src/character_attire.cpp
```diff
@@ -2230,6 +2230,61 @@ void outfit::pickup_stash( const item &newit, int &remaining_charges, bool ignor
...
+
+        for( const item *contained : pocket->all_items_top() ) {
+            const item_location poc_loc = item_location( it, const_cast<item *>( contained ) );
+            for( const item_pocket *pocket_nest : contained->get_all_contained_pockets() ) {
+                std::vector<pocket_data_with_parent> child =
+                    get_child_pocket_with_parent( pocket_nest, new_parent,
+                                                  poc_loc, nested_level + 1 );
+                for( const pocket_data_with_parent &pock_d : child ) {
+                    ret.emplace_back( pock_d );
+                }
```
mqrause, line 2253:
```suggestion
                ret.insert( ret.end(), child.begin(), child.end() );
```

src/character_attire.cpp
EIIKaO: Thank you for code review.
I agree with your view in the main. Though, I feel that it is better to make the another function for sort.
Not sure if it will work, but I will try.

src/character_attire.cpp
EIIKaO: > I don't think it warrants bypassing better_pocket like this, because whenever that gets touched this would need to be checked as well. You either need to work it into better_pocket somehow or ignore it.

I understand. My mind was also a bit confused.
I used to think that the nesting level should be incorporated into the sorting priority, but now I think that is more of a game design issue than a programming issue.
After this, I would like to brush up on the overall process and then test it, and if either way ( discard the unique condition and use only better_pocket or not use better_pocket) works as expected, then I would like to do so.

src/character_attire.cpp
EIIKaO: These changes are committed in batches just before committing after changing other parts of the file ( to avoid consuming extra watts in CI ).

src/character_attire.cpp
mqrause: > discard the unique condition and use only better_pocket or not use better_pocket

You absolutely have to use `better_pocket`, because we can't have two different implementations of (automatically) selecting a pocket to insert items into. The question is just if `better_pocket` is missing something that you want to implement in the scope of this PR.

src/character_attire.cpp
mqrause: If you want to keep it flexible, you can implement the sorting (and filtering) as predicate parameters. As long as this isn't used anywhere else, it's not strictly necessary, though.

src/character_attire.cpp
EIIKaO: I can understand what you are saying. I would like to adopt better_pocket this time.

src/character_attire.cpp
EIIKaO: Okay.
I am still learning modern programming, but I would like to give it a try.

src/character_attire.cpp
EIIKaO: Changed sorting conditions to be delegated to better_pocket.

src/character_attire.cpp
EIIKaO: I followed your advice and changed the scope and moved the filtering and sorting method to the function side.
Readability and code flexibility seem to have improved.

src/character_attire.cpp
```diff
@@ -2230,6 +2230,72 @@ void outfit::pickup_stash( const item &newit, int &remaining_charges, bool ignor
...
+        }
+    }
+    return ret;
+}
+
+std::vector<pocket_data_with_parent> Character::get_all_pocket_with_parent(
+    const std::function<bool( const item_pocket * )> &filter,
+    const std::function<bool( const pocket_data_with_parent &a, const pocket_data_with_parent &b )>
+    *sort_func )
+
```
EIIKaO, line 2263: I found a useless line break. Will correct it later.

src/character_attire.cpp
```diff
@@ -2230,6 +2230,72 @@ void outfit::pickup_stash( const item &newit, int &remaining_charges, bool ignor
...
+    for( item *i : items ) {
+        for( const item_pocket *pocket : i->get_all_contained_pockets() ) {
+            item_location loc = item_location::nowhere;
+            std::vector<pocket_data_with_parent> child =
+                get_child_pocket_with_parent( pocket, loc, item_location( *this, i ), 0 );
+            for( const pocket_data_with_parent &pock_d : child ) {
+                if( filter( pock_d.pocket_ptr ) ) {
+                    ret.emplace_back( pock_d );
+                }
+            }
```
mqrause, line 2290: Getting into nitpicky territory here, so sorry about that, but you can also pass the filter along to `get_child_pocket_with_parent` to simplify insertions a bit more.

src/character_attire.cpp
```diff
@@ -2230,6 +2230,72 @@ void outfit::pickup_stash( const item &newit, int &remaining_charges, bool ignor
...
+{
+    std::vector<pocket_data_with_parent> ret;
+    const auto sort_pockets_func = [ sort_func ]
+    ( const pocket_data_with_parent & l, const pocket_data_with_parent & r ) {
+        const auto sort_pockets = *sort_func;
+        return sort_pockets( l, r );
+    };
+
+    // Collect all pockets
+    std::list<item *> items;
```
mqrause, line 2273: Just a minor thing as well, but since you use wielded and worn item_locations, this could be an `std::list<item_location>`. Also saves you remaking the item_location for `get_child_pocket_with_parent`.

src/character_attire.cpp
mqrause: I think it's fine where it is, since it's about character equipment.

src/character_attire.cpp
EIIKaO: Thanks again for the code review.
I'll try it.

src/character_attire.cpp
```diff
@@ -2230,6 +2230,75 @@ void outfit::pickup_stash( const item &newit, int &remaining_charges, bool ignor
...
+            locs.emplace_back( worn_loc );
+        }
+    }
+
+    for( item_location &loc : locs ) {
+        item *i = loc.get_item();
+        if( !i ) {
+            continue;
+        }
+        for( const item_pocket *pocket : i->get_all_contained_pockets() ) {
```
mqrause, line 2290:
```suggestion
        for( const item_pocket *pocket : loc->get_all_contained_pockets() ) {
```

Minor stuff that wouldn't hurt even if left as is. You already checked the locations before inserting them into the vector, so no need to check them again. And `item_location` works pretty much like a pointer to an item. You can even do stuff like `if ( !loc ) { //something }`.

mqrause (approved):

src/character_attire.cpp
EIIKaO: I am gradually getting a better understanding of `item_location`. I will be able to make full use of it when I create future PRs.

PR #62077, opened 2022-11-04
Add national flags

MNG-cataclysm:
#### Summary
Content "Add flags of various countries"

#### Purpose of change
Add more flags, these are of various nations, and places where they will spawn.

#### Describe the solution
Add a single flag item, and add the alterations using the variants system.

Making each flag its own item, and/or adding more flags at a later date.

#### Testing
I spawned all the flags in, and everything works as intended.

#### Additional context
Unfortunately, adding images for tilesets is currently beyond my knowledge.


data/json/items/armor/cloaks.json
```diff
@@ -148,6 +148,298 @@
...
+        "id": "antiguan_flag",
+        "name": { "str": "Antiguan flag" },
+        "description": "It has a black V-shape surrounded by red.  The bottom of the V is blue, with a rising sun immediately above.",
+        "weight": 10,
+        "append": true
+      },
+      {
+        "id": "bahamanian_flag",
+        "name": { "str": "Bahamanain flag" },
+        "description": "It has three stripes of light blue and yellow, with a black triangle on the far left side.",
```
TheShadowFerret, line 241:
```suggestion
        "description": "It has three stripes of light blue and yellow, with a black triangle on the far-left side.",
```

data/json/items/armor/cloaks.json
```diff
@@ -148,6 +148,298 @@
...
+        "name": { "str": "Cameroon flag" },
+        "description": "It's divided into three green, red, and yellow stripes, with a golden star in the center.",
+        "weight": 10,
+        "append": true
+      },
+      {
+        "id": "canadian_flag",
+        "name": { "str": "Canadian flag" },
+        "description": "It has three red and white stripes, with the iconic maple leaf in red.",
+        "weight": 10,
```
TheShadowFerret, line 389: This needs a much higher relative weight(see beginning comment), eh.

TheShadowFerret (commented):
https://data.census.gov/cedsci/table?q=New%20England%20Division&g=0300000US1&hidePreview=false&table=B04006&tid=ACSDT1Y2018.B04006&syear=2020&vintage=2018&cid=DP05_0001E&layer=division&lastDisplayedRow=24&mode=

The US Census does break down demographics by ancestry for NE, if you want more accurate weights.

~~Also where's my Crazy Cataclysm override of Chad flag into the Gigachad flag~~

data/json/items/armor/cloaks.json
```diff
@@ -148,6 +148,298 @@
...
```
MNG-cataclysm, line 389:
```suggestion
        "weight": 50,
```

Qrox:
The variant list seems to be incomplete (it only contains countries starting with A to C).

MNG-cataclysm:
> The variant list seems to be incomplete (it only contains countries starting with A to C).

Yes. It was getting pretty large, so I though it best to create a pull request now and add the rest of the countries later.

Fris0uman:
Wow, I'm so not looking forward to spriting 190 flags X)
You weighted national flag slightly less than pride flag in every group, shouldn't it be the same or even a bit more likely?

KittyTac:
It should be "Bosnian" and not "Boznian" flag.

MNG-cataclysm:
> It should be "Bosnian" and not "Boznian" flag. And "Bhutanese" not "Bhutanian" flag. Probably more stuff like this but I'm a bit lazy.

I fixed the grammar for those two flags.

MNG-cataclysm:
> Wow, I'm so not looking forward to spriting 190 flags X) You weighted national flag slightly less than pride flag in every group, shouldn't it be the same or even a bit more likely?

I wouldn't think so. I doubt many people have flags of foreign countries, at least from what I've seen.

I-am-Erk:
Freezing because of the number of translation strings and sprites this adds

PR #62080, opened 2022-11-04
Provide hint for msvc build tag on downloads page

irwiss:
#### Summary
None

#### Purpose of change

"What does msvc stands for?" is a recurring question.

#### Describe the solution

Provide hint on downloads page


KymasTuran:
Also: suggested version if you still have Windows 7.

PR #62087, opened 2022-11-05
MOLLE Belt Loops

Spaceb0i:
#### Summary
Content "Added three different-sized MOLLE-compatible belt loops, along with recipes"

#### Purpose of change

Currently, the MOLLE webbing belt is the only MOLLE gear that has a belt loop, and it's restricted to 70cm. This is a problem if one wants to carry a larger belt clip item like a fire axe or halligan bar, or wants to carry multiple. MOLLE-compatible belt loops solve this problem.

#### Describe the solution

Three belt loops of different sizes (30cm, 60cm, and 100cm) can be crafted and attached to any MOLLE gear. Their PALS pocket use is relative to their size. The small belt loop is crafted from nylon, with the medium crafted from small and large crafted from medium. The weight, volume, and material are modeled roughly off of real-life PALS KA-BAR sheaths, which can be used for holding axes.

The only alternative would be to introduce variants of the MOLLE webbing belt, which somewhat defeats the purpose of the MOLLE/PALS system.

#### Testing

All belts were crafted and a screwdriver, pair of scissors, and a fire axe were all inserted successfully after attachment to a load-bearing vest (naturally, the fire axe did not fit in the small and medium belt loops).


data/json/recipes/armor/other.json
```diff
@@ -450,5 +450,57 @@
...
+    "result": "makeshift_ear_plugs_off",
+    "category": "CC_ARMOR",
+    "subcategory": "CSC_ARMOR_OTHER",
+    "skill_used": "fabrication",
+    "time": "5 m",
+    "autolearn": true,
+    "qualities": [ { "id": "CONTAIN", "level": 1 } ],
+    "tools": [ [ [ "lighter", 5 ], [ "ref_lighter", 5 ], [ "surface_heat", 2, "LIST" ] ] ],
+    "components": [ [ [ "wax", 1 ], [ "candle", 1 ] ], [ [ "cordage_short", 1, "LIST" ] ] ]
+  },
```
TheShadowFerret, line 466:
```suggestion
```
No double earplugs for you 😛

TheShadowFerret:
Both files need [a linting](https://dev.narc.ro/cataclysm/format.html) for JSON styling.

Spaceb0i:
Linting done. Didn't notice any changes but I guess that's what the program's for.

TheShadowFerret:
I don't think purging ammo_pouch.json was intended with that second commit 😛 revert

Dan1ss1mo:
seems like some linting is needed

TheShadowFerret (approved):

Dan1ss1mo:
Maybe change lengt limit to be 30/60/100 or 33/66/100, considering the amount of space the take?(1,2 and 3 respectively)

Spaceb0i:
Good suggestion, I'll do 30/60/100 a. because it looks nicer and b. because the Halligan Bar is too powerful and heavy to make sense on a medium belt loop

PR #62091, opened 2022-11-05
Prevent player from fatigue death after portal dungeon

GuardianDll:
#### Summary
Bugfixes "You won't die from fatigue after portal dungeon raid"

#### Purpose of change

So yeah, i went to portal storm, turn back, and died on my `steel drum (100L) > clean water (160)` from thirst, because character were not able to to wake up. Checking the save file reveal that i have 6533 fatigue (which is 11 times more than exhausted (575) status).

#### Describe the solution

Added an EoC, that starts same as portal reward, checks the player's fatigue level, and if it more than 1200, it make it 1200. If less, it do nothing

UPD: it seems `portal_dungeon_reflect_strength` just do not reset after the portal storm, which make the fatigue damage of mon_reflection to increase **forever** (so after 2 or 3 portal raid the value will be like 150, which makes you dead tired after 3 attacks (remember, it is a range attack, that ignore walls)). added a reset, but will left an EoC just in case

~~I believe the problem is some mad `EOC_mon_reflection_attack` fatigue damage scale, but i am unable to find the exact problem~~ see UPD

#### Testing

Went to a portal dungeon, give the player 99999 fatigue, escaped the dungeon, checked it again. Got 1200 fatigue
Went to a portal dungeon, give the player 1000 fatigue, escaped the dungeon, checked it again. Still have 1000 fatigue

#### Additional context
image
image
image
 
 
 
image
image
image

### DIFF
--- a/data/json/portal_storm_effect_on_condition.json
+++ b/data/json/portal_storm_effect_on_condition.json
@@ -212,6 +212,9 @@
       { "arithmetic": [ { "global_val": "var", "var_name": "cause_early_portal_storm" }, "=", { "const": 0 } ] },
       { "arithmetic": [ { "global_val": "var", "var_name": "portal_dungeon_slow_strength" }, "=", { "const": 0 } ] },
       { "arithmetic": [ { "global_val": "var", "var_name": "portal_dungeon_carry_strength" }, "=", { "const": 0 } ] },
+      {
+        "arithmetic": [ { "global_val": "var", "var_name": "portal_dungeon_reflect_strength" }, "=", { "const": 0 } ]
+      },
       "next_weather",
       {
         "u_message": "As suddenly as it began, the portal storm fades.  Reality returns to normal.  Hopefully the scarring is minimal…",


GuardianDll:
Ok something really wrong here - it seems mon_reflection can attack you ~150 times till you are in portal dungeon, dealing nearly 12k fatigue damage
not sure how to fix it
luckily the EoC prevent you from dying, decreasing the fatigue to more doable 1200

NetSysFire:
CC @Ramza13

Ramza13:
So I have been meaning to get around to this but I've been slow.  My plan was to add a max value into the arithmetic that adds fatigue, that way it will never go above  the value rather than resetting it.  See EOC_PORTAL_STRENGTHEN for an example of how to use a max value.   Not resetting the reflect strength was an oversight most likely, that whole encounter needs some more work.

I do really need to get around to tuning the portal dungeon more, thanks for the assistance.

GuardianDll:
For me it seems pretty hard to make it scalable. I must investigate it further, but, according to my precious tests, the refleaction can attack you with fatigue nearly 30-50 times per, eeh, lets name it "run" - the distance you need to walk between the exit vortex and vortex with reward. Picking this number, and "575" as "dead tired" fatigue level, we can calculate, that we can see the max value we can use for portal_dungeon_reflect_strength is 12 for the first run. If we allow to player to do more runs, the number of attack is increases, and the value will be 6 for second run, and 4 for third run. Not much of space for scale, as for me.

I-am-Erk:
IMO this is a reasonable-sounding fix and I think needs to be in for G-stable. We can further refine this stuff but this, as it stands, should merge if it is ready.

GuardianDll:
Close it prior to #62431

PR #62092, opened 2022-11-05
Add bodypart on-hit effects, unhardcode bleeding / blinding on damage, allow effects to modify limb scores

Venera3:
#### Summary
Features "Add bodypart on-hit effects, allow effects to directly modify limb scores"

#### Purpose of change

Unhardcode and limbify all the things - specifically different limbs should bleed differently, and different eyes should be able to be more/less resistant to blinding, and player status effects beyond those should move out of monster special attacks and instead be a function of the player's anatomy itself.

#### Describe the solution

- Add on-hit limb effects, checked by `Character::deal_damage_handle_type` (which needed a bit of adjustment to track the damage source)
- Deprecate `Character::make_bleed`, the `Monster`-level function is useful enough until we get freely-defined monster effect resistances. Lead to having to mess with a lot of one-off hardcoded bleed sources
- Unhardcode the dodge effect of adjacent creatures and adjust the grab logic to prevent zombies sponatneously letting you go after two turns
- Slightly limb-proof blinding ranged attacks, but that whole function will need a slight overhaul for Limb Stuff in the future
- Slightly limb-proof the hardcoded monster attacks, though they should mostly be on the way out
- Limb-proofed adding glare
- Added the previously hardcoded effects to the limbs where they make sense (for a full list see Additional Context)
- Added ability to modify limb scores via effects - either as global modifiers per effect instance or locally on the affected limb itself

TODO:
- [x] Audit monster attacks for effects that fit more as on-hit ones
- [ ] More playtesting

Overhaul bleeding in general - `int_dur_factor` means the scaling is pretty much linear, and the whole functionality could comfortably move over to EoCs / effect vitamin adjustments but that's pretty out of scope for this. I might look into that in a separate PR.

Don't bundle effect/limb score stuff into this, but that really opens the design space up for new onhit effects.

#### Testing

Loads, unit tests including the bleed inheritance one work. Debugmode messages tell you what is happening under the hood, intensity/duration limits are respected, scaling works as intended. Differential handling of minor and major parts works.

#### Additional context

The major change is you will usually get bled slightly more, but different bodyparts have different capacity to bleed and slightly different damage scaling based on damage type or limb - there's only so much bleeding a head can do, but a torso will rush up to major bleeding when shot pretty quickly. 

The other important points are the potential to get stunned (by major bash damage to your head), staggered (bash on torso), or downed (high amounts of any damage on your torso or bash damage on your legs), that should even the playing field slightly when it comes to incoming/outgoing status effects. Numbers are WIP, but there's no rush what with it being a definite Feature Freeze PR.
<details>
  <summary>Bodypart hit effects</summary>

 # Torso
 - Bleed on cut (max 1800s duration, 3% max HP threshold), stab, and bullet damage (unlimited duration, 1% damage threshold and faster scaling)
 - Winded on bash (10% base chance at 10% max HP threshold, max 15s duration)
 - Staggered (a new character version using limb score modifiers) scaling from 5% max HP for bash and 15% max HP for other damage types
 - Downed (15% max HP for bash and 25% max HP threshold for other damage types, 5% base chance)

# Head
- Bleed (lower max duration than torso)
- Stun (lower threshold on bash damage)

# Eyes
- Chance to blind on any damage above 3 HP, base 50% scaling with damage

# Arms
- Bleed with a relatively low max duration limit
- Staggered with the same setup as the torso

# Legs
- Bleed (higher max duration than arms, lower than torso)
- Bouldering 
- Downed
</details>

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -4223,11 +4223,11 @@ void Character::on_damage_of_type( const effect_source &source, int adjusted_dam
 {
     // Handle bp onhit effects
     bodypart *body_part = get_part( bp );
-    bool is_minor = bp->main_part != bp->id;
+    const bool is_minor = bp->main_part != bp->id;
     add_msg_debug( debugmode::DF_CHARACTER, "Targeting limb %s with %d %s type damage", bp->name,
                    adjusted_damage, name_by_dt( type ) );
     // Damage as a percent of the limb's max hp or raw dmg for minor parts
-    float dmg_ratio = is_minor ? adjusted_damage : ( 100 * adjusted_damage ) / body_part->get_hp_max();
+    float dmg_ratio = is_minor ? adjusted_damage : 100 * adjusted_damage / body_part->get_hp_max();
     add_msg_debug( debugmode::DF_CHARACTER, "Main BP max hp %d, damage ratio %.1f",
                    body_part->get_hp_max(), dmg_ratio );
     for( const bp_onhit_effect &eff : body_part->get_onhit_effects( type ) ) {
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -3784,9 +3784,9 @@ int Character::get_arm_str() const
 
 int Character::get_eff_per() const
 {
-    return Character::get_per() * get_limb_score( limb_score_vision ) + int( Character::has_proficiency(
+    return ( Character::get_per() + int( Character::has_proficiency(
                 proficiency_prof_spotting ) ) *
-           Character::get_per_base();
+           Character::get_per_base() ) * get_limb_score( limb_score_vision ) ;
 }
 
 int Character::ranged_dex_mod() const
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -4217,7 +4217,7 @@ bool Character::is_mute() const
                    is_wearing( itype_foodperson_mask_on ) ) ) ||
            has_trait( trait_MUTE );
 }
-void Character::on_damage_of_type( effect_source &source, int adjusted_damage, damage_type type,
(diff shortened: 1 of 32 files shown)


src/character.cpp
```diff
@@ -4243,8 +4229,61 @@ bool Character::is_mute() const
                    is_wearing( itype_foodperson_mask_on ) ) ) ||
            has_trait( trait_MUTE );
 }
-void Character::on_damage_of_type( int adjusted_damage, damage_type type, const bodypart_id &bp )
-{
+void Character::on_damage_of_type( effect_source source, int adjusted_damage, damage_type type, const bodypart_id &bp )
+{
+    // Handle bp onhit effects
+    bodypart *body_part = get_part( bp );
+    bool is_minor = bp->main_part != bp->id;
```
Night-Pryanik, line 4236:
```suggestion
    const bool is_minor = bp->main_part != bp->id;
```

src/character.cpp
```diff
@@ -4243,8 +4229,61 @@ bool Character::is_mute() const
...
+    add_msg_debug( debugmode::DF_CHARACTER, "Targeting limb %s with %d %s type damage", bp->name,
+                   adjusted_damage, name_by_dt( type ) );
+    // Damage as a percent of the limb's max hp or raw dmg for minor parts
+    float dmg_ratio = is_minor ? adjusted_damage : ( 100 * adjusted_damage ) / body_part->get_hp_max();
```
Night-Pryanik, line 4240:
```suggestion
    float dmg_ratio = is_minor ? adjusted_damage : 100 * adjusted_damage / body_part->get_hp_max();
```

Venera3:
Includes changes from #63097 for the tests to pass properly, that should probably merge before this.

Issue #24688, opened 2018-08-07
Creature hallucinations too easy to recognize as such


FulcrumA:
Divided #24659 as per @kevingranade recommendation:

It's still too easy to differentiate between actual creatures and hallucinations. Hallucinations are visible through walls and often include pretty regular creatures clearly not fitting the locale. Perhaps the game could concentrate on spawning hallucinations of creatures believably - even if not necessarily from actual technical standpoint - fitting in the place they appear at? Random eldritch abomination isn't bad from time to time, but it's actual skitterbots coming into view when one's exploring a laboratory that really make one question the reality.

Night-Pryanik:
No discussion in over half a year, closing.

PR #62098, opened 2022-11-05
Location-dependent hallucinations

Night-Pryanik:
#### Summary
Content "Location-dependent hallucinations"

#### Purpose of change
Inspired by #24688.

Hallucinations are too easy to differentiate from real creatures as very frequently they spawn clearly out of place, like a searchlight or a zombie shark inside a city house.

#### Describe the solution
If there's `spawns` exist for player's current location, then spawn random hallucination monster from `spawns`-dependent monster group (with 90% chance) or a completely random hallucination monster (10% chance). Thus, ordinary city houses will mostly spawn hallucination from `VANILLA` monster group, wasp towers will spawn hallucinations from `WASP_GUARD` monster group, and so on.

None.

#### Testing
Made my character hallucinating. Teleported to wasp tower, FEMA camp, butcher shop and checked which hallucinations are spawned. Also checked that completely random hallucinations are spawned on roads and other locations without defined `spawns`.

#### Additional context
This change severely lowers overall variety of spawning hallucination monsters, which may reduce fun for some players, so I'm putting [CR] tag.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -4809,8 +4809,18 @@ bool game::spawn_hallucination( const tripoint &p )
         }
     }
 
-    return spawn_hallucination( p, MonsterGenerator::generator().get_valid_hallucination(),
-                                cata::nullopt );
+    mtype_id hallu = MonsterGenerator::generator().get_valid_hallucination();
+
+    // If there's 'spawns' exist for player's current location, then
+    // spawn random hallucination monster from 'spawns'-dependent monster group
+    // instead of a completely random hallucination monster
+    const oter_id &terrain_type = overmap_buffer.ter( get_player_character().global_omt_location() );
+    const overmap_static_spawns &spawns = terrain_type->get_static_spawns();
+    if( !spawns.group.is_null() ) {
+        hallu = MonsterGroupManager::GetRandomMonsterFromGroup( spawns.group );
+    }
+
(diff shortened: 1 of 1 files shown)


RenechCDDA:
What about hallucination-only monsters like 'Your Mother'(it's real, not a jab!)? They don't spawn outside of hallucinations and are don't appear to be defined in any spawn groups. There are a handful of such monsters, this is just one I know exists offhand.

Night-Pryanik:
Well, as you correctly mentioned, hallucination-only monsters will spawn only on `spawns`-less locations, and right now I don't see an easy way to make them spawn on locations with `spawns` too. 

In general though I think it's a rather small issue, especially because there are lots of `spawns`-less locations and there is a small number of such monsters in the game.

Paprika-H:
Man this is exciting!

Rllaman:
Is that necessarily a problem? "I took hallucinogenics and saw a zombie shark swimming around in a 3rd floor apartment" kind of matches the nature of how hallucinations should happen.

If it's too easy to tell what is a hallucination, maybe it should be a random chance that you hallucinate something that makes sense for the location and another chance that you hallucinate something entirely random.

Night-Pryanik:
Ok, I added a 10% chance to spawn a completely random hallucination monster even in location with `spawns`.

PR #62099, opened 2022-11-05
silat unarmed + tkd tweaks

nopenoperson:
#### Summary
None

#### Purpose of change
silat is an extremely varied martial art irl. the game does a good job of representing the fighting stances and complicated movement patterns that are used in silat matches but it ignores the fact that a large subset of  silat practitioners only engage in unarmed sparring due to safety reasons.  we have chinese swords in game that have been used in indonesia for more than a 1000 years that should be usable with silat as well. 

tkd is the most popular martial art in the world and has a codified set of techniques. it has more than 20 named blocking techniques and none of these involve using your legs to block an incoming attack, instead relying on arm blocks, push kicks and quick distance shifts to avoid them.  the sideways fighting stance used in tkd sparring limits upper body movement and makes any circular motion very awkward but makes perfect sense in an art where both grappling and low kicks are outlawed because it enables rapid distance changes and offers less of a target to hit.  the game should represent the tradeoffs inherent in the tkd fighting stance more accurately.  

#### Describe the solution
letting the player use silat unarmed, adding medium and short swords to silat. add a leg sweep takedown to silat (very common technique in modern competitive silat). removing leg blocks from tkd, +1 dodge skill on the static buff, removed the disarm

changin eskrima as well. splitting more silat techs into unarmed and weapon versions 

#### Testing
a bit

#### Additional context
removing leg blocks from tkd makes it more consistent with the shotokan karate we have in game which uses a more square stance and could a lot more easily adapt to leg blocks than tkd (look up kyokushin karate or muay thai and notice the more square stance that allows for easier lateral movement and good balance while grappling or blocking kicks with your shin  ).

any vaguely sword shaped object up to a meter in length has most likely seen extensive use in a silat martial arts context. if the weapon was used by a european or asian navy at some point it is even more likely. 

if you type stuff like kiam silat into a search engine you get wild indonesian wuxia novels from the 30s
Giok+Siauw+Gin+Kiam+9-748673050

### DIFF
--- a/data/json/martialarts.json
+++ b/data/json/martialarts.json
@@ -1105,17 +1105,20 @@
     "type": "martial_art",
     "id": "style_silat",
     "name": { "str": "Silat" },
-    "description": "Pencak Silat, of Indonesian origin, is a fighting style that covers the use of short blades and bludgeons.  Fighters stay low and mobile to avoid attacks, then unleash deadly critical hits.",
+    "description": "Pencak Silat is a style that covers a wide variety of weapons and unarmed fighting techniques.  Fighters stay low and mobile to avoid attacks, then unleash deadly critical hits.",
     "initiate": [ "You give a salute of respect as you prepare to combat.", "%s gives a combat salute." ],
     "learn_difficulty": 7,
     "primary_skill": "cutting",
-    "strictly_melee": true,
+    "arm_block": 3,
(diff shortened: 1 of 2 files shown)


Dan1ss1mo:
there are few small conflicts in the descriptions currently

PR #62107, opened 2022-11-06
[Xedra Evolved] some small fixes

GuardianDll:
#### Summary
Mods "Another few Xedra fixes"

#### Purpose of change

I was pissed off because i was not able to find any of spells i made to test them in game, a random spark insight me, and i recall another bugs i thought to fix

#### Describe the solution

- Replace SUS_junk_drawer loot with loot groups i made before, because now, when we have no locations, its the most easy place to obtain stuff (and use them, and see is there some issues with them)
- Changed the dreamdross scrap and lump weights and densities - now to make 1 lump you need 10 scrap, and for 1 ignot you need 100 scrap, also now scrap has correct weight, because before it was 50g per scrap
- Fixed a rip ticket recipe - now it can use pens and tools (i don't know why i thought -1 means "use 1 charge", it was obvious it means "don't use charges at all")
- Tweaked a forge dreamdross requirement, so it require not the hammer itself, but just his tool quality
- Added NO_HANDS to eat_dreamdross - it has 5 second cast, so its not a battle spell anyway, just QoL

#### Testing

Everything works as intended, i was surprised tbh

#### Additional context

and now its 2 am, i am pissed of a github, and of an old me, that didn't spot the issues i fix now

### DIFF
--- a/data/mods/Xedra_Evolved/spells/eater_spells.json
+++ b/data/mods/Xedra_Evolved/spells/eater_spells.json
@@ -17,7 +17,7 @@
     "base_casting_time": 500,
     "base_energy_cost": 0,
     "components": "spell_components_dreamdross",
-    "flags": [ "SILENT", "NO_LEGS" ],
+    "flags": [ "SILENT", "NO_LEGS", "NO_HANDS" ],
     "learn_spells": { "eat_more_dreamdross": 20 },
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):
Thank you!

PR #62111, opened 2022-11-06
Add astyle-diff target in Makefile to format modified files only

BrettDong:
#### Summary
None

#### Purpose of change
`make astyle-fast` still walks through all source files in the project. That could be slow on systems with slow disk I/O. It would be nice if I can just let `AStyle` read and format the files that I've changed.

#### Describe the solution
Add a `astyle-diff` Makefile targets that only format files modified. The list of edited files is generated by `git diff --name-only`.

#### Testing
On a clean source tree, touch a single `.cpp` file, and run `make astyle-diff`. Only that single file is read and formatted by `AStyle`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1210,6 +1210,9 @@ endif
 astyle-fast: $(ASTYLE_SOURCES)
 	$(ASTYLE_BINARY) --options=.astylerc -n $(ASTYLE_SOURCES)
 
+astyle-diff: $(ASTYLE_SOURCES)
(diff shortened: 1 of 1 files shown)


akrieger:
`make astyle` already only changes files updated since the last `make astyle`, and it relies on make's own logic for that instead of forking out to another program. The only time that this new make command helps is running `make astyle` after a rebase or other git operation which changes a lot of files, after which the timestamps are updated already. Is that sufficiently painful that this is useful, in which case, should we just change `make astyle` to use this logic (which would strictly regress it on Windows because shelling out and git are both substantially slower on Windows than elsewhere)?

Issue #53482, opened 2021-12-15
Items with FIRE flag are not recognized as a source of flame


lapis413:
### Describe the bug

When an everburning torch is lit, it's description has the note "This item counts as fire for crafting purposes" but does not appear to fulfill that for any crafting recipes.

### Steps to reproduce

1. Spawn an everburning torch item
2. Activate the torch
3. Check torch's description to confirm it should function as fire
4. Check crafting recipes for anything that requires fire

### Expected behavior

Either for the item to serve as fire, whether in all recipes or only certain ones, or for the entry in the item's description to be removed if it is not intended.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: 9b60b4c [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Bionic Professions [package_bionic_professions],
    Aftershock [aftershock],
    Blaze Industries [blazeindustries],
    Magiclysm [magiclysm],
    DinoMod [DinoMod],
    My Sweet Cataclysm [my_sweet_cataclysm],
    Mythical Martial Arts [MMA],
    SpeedyDex [speedydex],
    Stats Through Skills [StatsThroughSkills]
]

catdach:
This isn't exclusive to magiclysm, the refillable lighter (lit) has the same problem.

Night-Pryanik:
Fixed by #62129.

PR #44700, opened 2020-10-07
[CR] Jsonize vehicle tools

irwiss:
#### Summary

SUMMARY: None

#### Purpose of change

Jsonize handling of vehicle tools / crafting stations etc - similar to `crafting_pseudo_item` on furniture

Fixes #34851

#### Describe the solution

This jsonizes vehicle tools removing a bunch of flags and instead
uses normal type_ids to provide the tools, this should also allow mods to
provide their own crafting stations if implemented in normal type_ids and "iuse"s
and provide flexibility when choosing what tools to provide in general

Flags removed: CHEMLAB, CRAFTRIG, FAUCET, FORGE, KITCHEN, TOWEL, WELDRIG

**Shady stuff I did:**
* Hotkey from json is read as string and then mutilated `static_cast<int>( hotkey_str[0] )` not sure this is street legal
* water_faucet is used as tool id that allows access to liquids - in another PR i'd like to switch it to rubber hose, also (unchanged) it's still a cheaty component that allows magical access to any liquid on the vehicle for crafting
* I'm storing tools in vpart_info as std::set<std::pair<itype_id, int>> instead of std::map to preserve order declared in JSON

Less shady stuff:
* Assume ALL tools don't work if part providing them is broken - functions pulling tools don't accept ignoring broken-ness
* Parts with KITCHEN receive FLAT_SURF instead for map::has_nearby_table for equivalent functionality
* Parts now print the exact tools they provide, descriptions for parts need to be updated (in another PR), see in additional info
* Tweaked weldrig a bit to also support pulling the tools by name instead of part index, theoretically it could open more stuff with battery input and repair activity, which I don't think there is any of yet

For modders (and myself if I forget):
* Parts with KITCHEN - in addition to other things should also receive `FLAT_SURF` flag if you want to keep parity
* Parts which earlier had flags should receive:
* FAUCET => `[ "water_faucet" ]`
* KITCHEN => `[ "hotplate", "pot", "pan" ]`
* WELDRIG=> `[ "welder", "soldering_iron" ]`
* CRAFTRIG => `[ "vac_sealer", "dehydrator", "food_processor", "press" ]`
* CHEMLAB => `[ "chemistry_set", "electrolysis_kit" ]`
* FORGE => `[ "forge" ]`
* KILN => `[ "kiln" ]`
* TOWEL => `[ "towel" ]`

#### Testing

I've been playtesting this for a while and it's mostly stable.

The tests I've written in #44735 should verify crafting works with
[...]

PR #62129, opened 2022-11-08
allow a couple items to act as fire for crafting

casswedson:
#### Summary
None

#### Purpose of change
so there's this flag "FIRE" that says items with that flag will act as a nearby fire when crafting, that's a lie, rather it doesn't work

I somehow got the "FIRE" flag proposed behavior working after some tinkering inspired by a pr in the bn repo (their solution also didn't seem to work hence the tinkering), I referenced a pseudo item or something the crafting requirements list use "fire" the bn pr also used this json object `"sub"` that works in a similar? manner as copy-from; I copied something from "fire" and now I can use a lit torch as a fire in crafting

dirty hack; there's no code support for the "FIRE" flag, this is just an odd interaction

a couple items now serve as fire when crafting:
- the lit torch
- magiclysm's everburning torch and pocket sun

I realize now "sub" is thing we use quite a bit here, like for vehicles, or some tools, hey I didn't know

#### Testing
Screenshot from 2022-11-07 20-27-23

### DIFF
--- a/data/json/items/tool/lighting.json
+++ b/data/json/items/tool/lighting.json
@@ -762,6 +762,7 @@
     "max_charges": 20,
     "turns_per_charge": 60,
     "revert_to": "torch_done",
+    "sub": "fire",
     "use_action": [
       { "type": "firestarter", "moves": 30 },
       {
--- a/data/mods/Magiclysm/items/enchanted_misc.json
+++ b/data/mods/Magiclysm/items/enchanted_misc.json
@@ -104,6 +104,7 @@
       }
     ],
     "techniques": [ "WBLOCK_1" ],
+    "sub": "fire",
     "extend": { "flags": [ "FIRE", "LIGHT_240", "FLAMING", "TRADER_AVOID", "WATER_EXTINGUISH" ] }
   },
   {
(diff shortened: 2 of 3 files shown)


Night-Pryanik:
I can't check right now if there's documentation for this `sub` stuff? If not, maybe add it?

Also, could you please check if this closes #53482?

Fris0uman:
If the FIRE flag doesn't work as advertised we should change the doc about it

casswedson:
there's a little info about "sub" here https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md#tools

this certainly doesn't fix any issues, problem is: there's no code support for the flag itself

I suspect the functionality of FIRE was removed at some point; why would so many items have if it doesn't do anything, this is just a little hack please don't blame for not wanting to dig too deep into the details

RenechCDDA:
Both the documentation and description were added in 9e4e5d52fbe55726dcdde8ce11304fcd3888302a

There are two checks in character.cpp which look for FLAG_FIRE. They must have broken at some point. *Probably* #44700, looking at the changes to inventory.cpp and this PR's summary.

There is in fact code support for FLAG_FIRE, it just isn't working.

PR #62130, opened 2022-11-08
acetylene recipe batch savings + byproduct

Daved27hundred:
#### Summary
Content "Added batch time savings to the oxyacetylene and acetylene recipe and added slaked lime as a byproduct to the latter."

#### Purpose of change

Acetylene is made ingame via putting calcium carbide in water and extracting the gas it produces, however this recipe doesn't consider the fact that the calcium doesn't just disappear: it forms into calcium hydroxide (AKA slaked lime) as the carbon switches to bond with the hydrogen of the water. I also added batch savings to the recipe since acetylene is absolutely craftable in bulk: just add more water and CaC2.

#### Describe the solution

added a byproduct line to acetylene, 50 slaked lime created per batch. the calcium carbide used for 1 batch amounts to 0.10 L and the acetylene created is 0.04 L, so I added the remaining 0.05 L as the lime.

I also made it so 80% of time was saved per 1 unit crafted of acetylene. I did this with oxyacetylene as well, because I felt like the recipe was similar enough.

shorter craft time, though this seems like the better option

### DIFF
--- a/data/json/recipes/ammo/weldgas.json
+++ b/data/json/recipes/ammo/weldgas.json
@@ -8,6 +8,7 @@
     "skill_used": "chemistry",
     "difficulty": 4,
     "time": "45 m",
+    "batch_time_factors": [ 80, 1 ],
     "charges": 120,
     "book_learn": [ [ "textbook_chemistry", 4 ], [ "textbook_gaswarfare", 4 ], [ "atomic_survival", 3 ] ],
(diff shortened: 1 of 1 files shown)


RenechCDDA:
This produces 12,500 slaked lime. Byproducts are multiplied by count/charges and slaked lime has a default count of 250.

image

image

Daved27hundred:
uh. oops.
I'm fixing that RIGHT away lol.

PR #62135, opened 2022-11-08
Allow HWP to use HWP brand 12.3 magazine

Highmongrel:
#### Summary
Bugfixes "Allow HWP to use HWP brand 12.3 magazine"

#### Purpose of change

Currently the HWP with EXODIC configuration can only take the exodii provided 60 round 12.3 magazines, and not the ones provided by the HUB 01 purpose built for it. The change fixes that.

#### Describe the solution

Add the magazines id to the accepted magazines of the HWP

Not fixing it

#### Testing

Changed the JSON
Booted game
Checked that I could load the HWP with HWP 60-round 12.3 magazine
Loaded gun
Fired gun

#### Additional context

Not sure why the 2 magazines are almost identical outside of length and reload time, what with one being a magazine for an extra-worldly battle rifle/light emplacement and the other designed for a bullpup, but that's outside of scope

### DIFF
--- a/data/json/items/gun/robofac_gun.json
+++ b/data/json/items/gun/robofac_gun.json
@@ -48,7 +48,8 @@
           "robofac60",
           "exodiiBRmag",
           "robofacmultimag",
-          "robofac30"
+          "robofac30",
+          "robofac60_123ln"
         ]
       }
     ]


tenmillimaster:
There are lore reasons as to why I think this should not be accepted. The HWP is to me, an overly optimized "perfect" platform with overly optimized, "perfect" magazines. That's what I would expect the Hub to build, and there's a little bit of that already in the design.

- It is known that STANAG 5.56 magazines are not "perfect". The magazines was designed with 20 rounds in mind, and thus a straight magazine was acceptable. When 30 rounders were desired, a tapered magazine section was stuck onto the straight portion of the magazine. This discontinuity is a minor sticking point, but the magazines work well enough. If the design were redesigned, it could  be *perfect*, but there are a ridiculous number of AR-pattern magazines around. It's not worth it. Later magazines that were designed for such larger capacities are curved throughout. It's of this pattern that I speculate the 5.56 mag of the HWP is. Failing to evaluate the abundance and practicality of using a commonly available magazine is exactly the kind of failure in logic I expect the HUB01 to make. Who cares if it's the *perfect magazine* if there aren't piles of them sitting in zombified soldiers' chest rigs?
-  The Exodii battle rifle shouldn't be using anything like a "perfect" magazine, it should be some mostly decent alternate history WW1-era design. Think Madsen LMG (an awful gravity fed magazine!) It's supposed to be another example of "good enough", in essence the opposite approach of what I speculate the HUB approach has been. Where the HUB would set out to design (without sufficient input from the end users!) the "perfect gun" with "the perfect magazine", perfecting the unattainable modular weapons concept that end users have scoffed at, the Exodii would look at if it meets certain criteria (Payload, Size, Range, Recoil characteristics, acceptable accuracy and reliability, and most importantly, quantity of materiel).
- I'm uncertain as to why it would have occurred to the Hub to even make an exodii compatible gun, or if it would even be feasible. For a given cartridge, the action length that is too long is just a lot of dead weight when you're using the gun, and an action length that is too short means you can't chamber it for a longer cartridge. The longest cartridge I think they'd be designing the gun for is 277 Fury which is 2.825" long. .30-06 (which is the model for 1.23ln is 3.34" long. IMO the exodii model should be obsoleted.

Maleclypse (approved):

Issue #61874, opened 2022-10-23
excessively large byproduct when making baking soda


robotic-gamer:
### Describe the bug

the crafting recipe for baking soda seems to produce far too much quicklime (see screen shot). mass wise for crafting 1 unit the washing soda and limestone combined weigh 0.21kg. by comparison the quicklime weighs 6kg so the recipe clearly seems to be creating a very large amount of calcium from thin air.

### Steps to reproduce

look up the recipe

### Expected behavior

the recipe producing material at a net loss of material relative to the materials inputed.

### Screenshots

![image](https://user-images.githubusercontent.com/78816055/197368837-683dd02b-3631-4d1f-bb64-bd79548edb8f.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.2130 (21H2)
- Game Version: 26a4725 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Xedra Evolved [xedra_evolved],
    No Hope [no_hope],
    Aftershock [aftershock],
    Bionic Slots [cbm_slots],
    Stats Through Skills [StatsThroughSkills],
    SpeedyDex [speedydex],
    Tamable Wildlife [Tamable_Wildlife]
]

RenechCDDA:
/confirmed

Baking soda is turning 8 limestone into 400 quicklime.

Quicklime recipe is turning 50 limestone into 50 quicklime.

This is because quicklime is defined as having a default `count` of 50. 

Baking soda recipe needs to be defined as in JSON_INFO.md

`"byproduct_group": [         // Optional (default: empty). Same as above, but using item group definitions.
  { "item": "item_id_1", "count": [ 1, 4 ] },
  { "item": "item_id_2", "charges": [ 8, 15 ] },`

Set the count/charges to the desired output (presumably 8). 

The following recipes also seem to be affected:

chem_washing_soda, budget_steel_chunk, budget_steel_lump, cotton_ball, cotton_ball_spinningwheel, cotton_ball_by_hand, 

Just took a quick check through the CHEM and OTHER category. Maybe there are other recipes hiding.

RenechCDDA:
[...]

PR #62142, opened 2022-11-08
acetylene recipe no longer makes 250x more slaked lime than it should

Daved27hundred:
#### Summary
Bugfixes "acetylene recipe makes 50 slaked lime instead of 12500 slaked lime as a byproduct"

#### Purpose of change
 
I messed up in #62130, and I'm hastily correcting it (see comments on that PR)

#### Describe the solution

changed byproduct line to read `[ { "item": "chem_slaked_lime", "charges": 50 } ]` so it creates 50 individual charges instead of 50 fully charged instances.

However, the byproduct now inexplicably manifests a plastic bag when created, due to that being the default container of the item. I plan to fix that later by moving the container from the item to the spawngroup but fixing the mistake I made is my priority.

edit: I think I've fixed it now, moved the plastic bag to the itemgroup instead of the item.

#### Testing

MrDraMaster on the discord gave me this solution and even tested it on their own build, so thank you for that.
image
byproduct shows up as just a bag, but the lime is in it.

### DIFF
--- a/data/json/itemgroups/supplies.json
+++ b/data/json/itemgroups/supplies.json
@@ -352,7 +352,7 @@
       { "item": "chem_antimony_trichloride", "prob": 2, "charges-min": 1 },
       { "item": "chem_ferric_chloride", "prob": 15, "charges-min": 1 },
       { "item": "chem_nickel_powder", "prob": 5, "charges-min": 1 },
-      { "item": "chem_slaked_lime", "prob": 5, "charges-min": 1 },
+      { "item": "chem_slaked_lime", "prob": 5, "charges-min": 1, "container-item": "bag_plastic", "sealed": false },
       { "item": "chem_baking_soda", "prob": 10, "charges-min": 1 },
       { "item": "chem_washing_soda", "prob": 10, "charges-min": 1 },
       { "item": "chem_calcium_chloride", "prob": 5 },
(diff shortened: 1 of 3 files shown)


Daved27hundred:
okay, so despite thinking I did twice, this branch isn't lined up properly with the master. Anyone know how to rebase a branch?

RenechCDDA:
Unfortunately `byproducts` does not support this input. `byproduct_group` does, but 1) forces it to spawn in a container if default container is specified (specifying null just returns default container) and 2) *duplicates the container and contents*. I've been looking into these issues as part of #61874 and some C++ changes are needed to properly spawn in-container/default container items.

The straightforward path to resolving this *without* C++ changes is to remove the item's default container and change your current field to `byproduct_group`, then it should output properly.

Maleclypse:
Please check that I fixed your conflicts correctly.

Maleclypse:
```Error: ([slow] ~starting_items)=> Expected an array, got a 9

  {
    "result": "acetylene",
    "byproducts": [
                   ^
                    { "item": "chem_slaked_lime", "charges": 50 } ],
    "type": "recipe",

([slow] ~starting_items)=> 23:04:51.191 INFO : Make sure that you're in the correct working directory and your data isn't corrupted.
test exited with code 1```

RenechCDDA:
I grepped for chem_slaked_lime earlier and confirmed that is the only item group which uses it.

I've also implemented the changes locally and confirmed that they work as advertised. 50 units of slaked lime, not 12,500. The crafting gui is still wrongly reporting 12,500 but that should be resolved by my PR.

Daved27hundred:
thank you for that renech

Maleclypse (approved):
Thanks!

PR #19302, opened 2016-11-15
Sorted/Added WorldNames based on real NE cities.

DifferentLevelDan:
resolves #19292

Names were merged/sorted automatically using a small utility program.

Issue #19292, opened 2016-11-15
Updated Repository of World Names


TheWebfrog:
[THIS](https://en.wikipedia.org/wiki/List_of_New_England_towns) is a list of all the New England Towns, altough this is an ardous task it would be great to update the World name gen with more of the ACTUAL New England Towns, Plantations and Cities.

Rod-Serling:
I took that list, copied it, stripped out all the extra text, and alphabetically sorted it.

Then I took the list from names/en.json, stripped the json formatting, and compared.

Names from names/en.json are on the left.
Names from the wiki are on the right.
https://www.diffchecker.com/FvtLBTsM

Looks like someone already used wiki to get the full list of names, as evidenced by Line 348 of the diff "Former town:" appearing in the list of names already in the game.  While this is an issue, the only other differences are in alphabet order, and three towns missing (Dana, Jeffersonville, and Wentworth's Location) two of which are former towns no longer incorporated (Dana and Wentworth's Location).

All in all the list of names is fine. I'll issue a PR shortly adding the three missing towns and removing the "Former town:" bug.

kevingranade:
The real issue here is that few if any places actually use the town names.

Rod-Serling:
On second looks, I compared with city names which are not really used anywhere. WorldNames, there is a good amount of differences that we could rectify.

PR #62144, opened 2022-11-08
Removed questionable town names

Spaceb0i:
#### Summary
Content "Removal of questionable town names and surnames"

#### Purpose of change
"Coon" is often used as a slur. "Coon Valley" and "Coon Rapids" lie in Wisconsin and Minnesota respectively, rather than in New England, so it's not really a problem to remove them.

#### Describe the solution
"Coon" has been removed from the autogenerated surname list. "Coon Valley" and "Coon Rapids" have been removed from the autogenerated town name list.

None

#### Testing
None necessary


I-am-Erk:
Obviously not a huge issue but a reasonable change. What I find interesting is that these weren't in the master list of NE place names in #19292 but were in the initial addition in #19302, so someone made the conscious choice to add them which I find a bit sus.

PR #62145, opened 2022-11-08
Cold Tolerance and Heat Tolerance traits

anoobindisguise:
#### Summary
Content "Traits for minor temperature tolerance"

#### Purpose of change
Some people are naturally better at dealing with the cold or the heat regardless of how much/little clothing they wear. 
The most extreme example of this might be Wim Hof, who while almost certainly a quack HAS displayed the ability to swim through arctic-cold water with no protective gear whatsoever. There are both genetic and conditioning factors to this, which makes it an ideal candidate for a trait (either born with it or trained it long term before the cataclysm)

#### Describe the solution
Currently I have added two traits - heat tolerant and cold tolerant, which use the new CLIMATE_CONTROL_CHILL/HEAT enchantments to give you 10 climate control towards their respective temperatures. Each are worth 1 pt, which felt reasonable given they are a fairly minor benefit to a circumstance that only occasionally comes up. They were also added to mutation categories - cold tolerance mostly for furry mammals (except rats) and heat tolerance for cold blooded lines.

Make them more powerful, 10 points might be too low but I wanted to be a bit conservative to make sure they don't strain realism too hard. But I could easily see them being 15 or 20 and still being "worth" 1 point (under my perspective of our arbitrary and largely meaningless point balance system)

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -205,6 +205,8 @@
       "POISONOUS",
       "POISONOUS2",
       "RADIOGENIC",
+      "RESISTCHILL",
+      "RESISTWARM",
       "ROBUST",
       "STOCKY_TROGLO",
       "STRONGBACK",
@@ -8251,5 +8253,25 @@
     "points": 0,
     "valid": false,
     "purifiable": false
+  },
+  {
+    "type": "mutation",
+    "id": "RESISTCHILL",
+    "name": { "str": "Cold Tolerance" },
+    "points": 1,
+    "description": "You fare better than most in colder climates and are less bothered by lower temperatures.  Note that while you are slightly less prone to developing frostbite as a result, you aren't any better at dealing with the symptoms.",
+    "starting_trait": true,
+    "category": [ "LUPINE", "URSINE", "CATTLE", "FELINE", "RABBIT", "MOUSE", "BEAST", "PLANT" ],
+    "enchantments": [ { "condition": "ALWAYS", "values": [ { "value": "CLIMATE_CONTROL_CHILL", "add": 10 } ] } ]
+  },
+  {
+    "type": "mutation",
+    "id": "RESISTWARM",
+    "name": { "str": "Heat Tolerance" },
+    "points": 1,
+    "description": "You fare better than most in warmer climates and are less bothered by hotter temperatures.  Note that while you'll sweat a little less in close proximity to a fire as a result, burns will still hurt just as much.",
+    "starting_trait": true,
+    "category": [ "FISH", "CEPHALOPOD", "GASTROPOD", "LIZARD", "BATRACHIAN", "PLANT" ],
+    "enchantments": [ { "condition": "ALWAYS", "values": [ { "value": "CLIMATE_CONTROL_HEAT", "add": 10 } ] } ]
   }
 ]


Maleclypse (approved):

PatrikLundell:
Given that you can overheat in summer while completely naked. I think heat tolerance would be an extremely worthwhile buy.

I'm not sure about the fur part, though. Doesn't fur cause warmth, which in itself provides the cold tolerance added again here, and at the same time it makes the overheating problem worse, so tolerance of heat to compensate for the fur's inherent increased heat would be very helpful to mitigate that problem?

anoobindisguise:
> Given that you can overheat in summer while completely naked. I think heat tolerance would be an extremely worthwhile buy.
> 
> I'm not sure about the fur part, though. Doesn't fur cause warmth, which in itself provides the cold tolerance added again here, and at the same time it makes the overheating problem worse, so tolerance of heat to compensate for the fur's inherent increased heat would be very helpful to mitigate that problem?

so first of all the enchantments are entirely beneficial, one way. Having cold tolerance doesn't make you more weak to the heat. As far as I know the enchantments simply say "am I below 0 warmth? If so add warmth equal to my enchantment value, up to 0" (or vice versa). I gave furry mammal lines cold tolerance since they clearly survive the cold better than the warmth of their fur alone would indicate, but their thermoregulation abilities are not fantastic enough to warrant giving them heat tolerance. I'm thinking of a wolf, field mouse, or rabbit that can be in the snow just fine, but have a tougher time in summer - but not so tough that they're just totally screwed which would be the case now if I just bumped up the warmth of their fur more.

Second heat tolerance is definitely more immediately appealing due to our current game start, which is when heat matters - the end of spring and the duration of summer, while you are outside, and during the day, and not underground - strictly speaking this is a pretty small slice of the year, it's just that currently games go very fast and most never see the end of summer, and experienced players can make a character endgame with maxed skills, bionics, etc, by the end of spring, so heat is more immediately relevant. Also, 10 isn't a HUGE amount, I often look at my warmth and easily have ~60 in the summer and taking it down to 50 isn't going to make it super tolerable by itself - it will be just a compounding factor in your staying cool.

PatrikLundell:
I'm definitely not suggesting increasing the warmth of fur, as that would make things worse, as you say.

10 points may not be a huge difference, but it can definitely help push you down to warm from overheating in many situations.

When it comes to cold tolerance, I had to change my Innawood start time from day 60 to day 70 because my characters simply couldn't sleep due to the cold (which includes the lower level of the starting caves). Thus, there's definitely a use for it (plus it being convenient to not have to change clothing all of the time because the temperature changes over the day in later stages). I believe the shivering/sleep code has been updated since, but still.

NetSysFire:
>my characters simply couldn't sleep due to the cold

This might have been fixed by #53443

The PR itself is fine but to discuss the heat issues with fur in the summer:

There are different kinds of fur and fur can grow denser in the winter, too. Some animals even combat heat with [Aestivation](https://en.wikipedia.org/wiki/Aestivation), the summer counterpart of hibernation, although it tends to be lighter and appears unsuitable for human-sized mammals.

Perspiration looks to be the right tool to use here. Canines pant to cool down while e.g kangaroos can lick their arms to reduce their body temperature. Housecats (and possibly feral felines) are passively cooled down by their big ears and a human gently wetting them with a few droplets of water can help them to endure the heat in the summer.

Whether mutations would rework the players body in a way that the normal sweat glands and pores will not work anymore is an important factor, too. Fur has an insulating quality on its own, so I would give it a little heat resistance but mutations will need to do the rest of the work. Players with canine mutations would automatically start to pant while feline ones could activate a mutation that makes them spend a little water stored in the body to wetten their ears with a bit of saliva if the purely passive cooling is not enough.

The whole matter would probably benefit from an own issue so it can be discussed there.

PR #62148, opened 2022-11-09
Extend byproduct groups, allow overriding default container with null

### LINKED ISSUE #61683: Unintended byproduct of crossbow practice
### Describe the bug

Intermediate crossbow practice needs 20 wooden broadhead bolts and return 30.

### Steps to reproduce

1. Practice advanced crossbow.

### Expected behavior

Return of the 20 bolts needed for doing it and not getting any extra out if thin air.

### Screenshots

imagen

- OS: Windows
    - OS Version: 6.1 7601
- Game Version: feaf1b8 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Aftershock [aftershock],
    Mythical Martial Arts [MMA],
    Magiclysm [magiclysm],
    Magiclysm No Class Limit [magiclysm_no_class_limit],
    Xedra Expanded No Class Limit [xedra_no_class_limit],
    Tamable Wildlife [Tamable_Wildlife]
]

### Additional context

_No response_

RenechCDDA:
#### Summary
Features "Extended byproduct groups functionality"

#### Purpose of change
Closes #61683

Resolves several issues with using byproduct groups, which previously went unnoticed.

#### Describe the solution
There are three commits in this pull.

The first prevents a byproduct group with any defined container (default or otherwise) from duplicating its contents. If we initialize at 0 it always spawns *one* extra group because, well, 0 is less than the batch size of 1. And 1 is less than the batch size of 2... and so on. (e.g. batch of 1 produces 2 byproduct groups, batch of 10 produces 11 byproduct groups, batch of 50 produces 51). Therefore we initialize at 1, and get exactly the correct amount of byproduct groups as we should.

The second allows "null" to be defined in json for a byproduct group **or item group** and have it override the default container. Previously, an item group would always spawn with either the defined `container-item` or its default if one existed. It was not possible to prevent the items from spawning without a container without removing both the defined and default containers.

The third fixes the crafting display when using byproduct groups. Before and after images will be in the additional context section.

*What if we just killed charges?* (I'm letting someone else deal with that)

#### Testing
Verify that the quantities (amount, volume, etc) displayed in the rightmost pane match the ones in the middle pane, as well as the actual results, for 10 of each:

wooden club (free item with no charges)
oxidizer powder (free item with charges and conventional byproducts)
fruit juice, 3L tinned recipe (contained item with charges and byproducts)
(test procedure shamelessly adopted from an earlier PR addressing these same problems)

#### Additional context
Before: 
image

After:
image
(Yes if you're wondering: that is still way too much slag even though the display is correct; the recipe itself will be fixed in a following PR)

### DIFF
--- a/src/item_group.cpp
+++ b/src/item_group.cpp
@@ -415,9 +415,9 @@ void Item_modifier::modify( item &new_item, const std::string &context ) const
 
     // create container here from modifier or from default to get max charges later
     item cont;
-    if( container != nullptr && cont.is_null() ) {
+    if( container != nullptr ) {
         cont = container->create_single( new_item.birthday() );
-    } else if( cont.is_null() && new_item.type->default_container.has_value() ) {
+    } else if( new_item.type->default_container.has_value() ) {
         cont = item( *new_item.type->default_container, new_item.birthday() );
     }
 


src/item_group.cpp
```diff
@@ -418,7 +418,9 @@ void Item_modifier::modify( item &new_item, const std::string &context ) const
     if( container != nullptr ) {
         cont = container->create_single( new_item.birthday() );
     }
-    if( cont.is_null() && new_item.type->default_container.has_value() ) {
+    if( container != nullptr && cont.is_null() ) {
+        cont = container->create_single( new_item.birthday() );
+    } else if( cont.is_null() && new_item.type->default_container.has_value() ) {
```
mqrause, line 423: I think you should simplify this as discussed on discord. Apart from the code duplication and possible duplicated execution, it's also harder to understand what exactly is supposed to happen here.

src/crafting_gui.cpp
```diff
@@ -393,7 +393,7 @@ static std::vector<std::string> recipe_info(
             const itype *t = item::find_type( bp.first );
             int amount = bp.second * batch_size;
             if( t->count_by_charges() ) {
-                amount *= t->charges_default();
+                amount *= t->count_by_charges();
```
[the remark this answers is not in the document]
RenechCDDA: Ah, that is indeed a boolean! I assumed it was passing an integer for (a count of) charges.

src/item_group.cpp
```diff
@@ -415,14 +415,10 @@ void Item_modifier::modify( item &new_item, const std::string &context ) const
 
     // create container here from modifier or from default to get max charges later
     item cont;
-    if( container != nullptr ) {
+    if( container != nullptr && cont.is_null() ) {
```
mqrause, line 418:
```suggestion
    if( container != nullptr ) {
```

`cont` is always null in these checks, so it can also go.

src/item_group.cpp
```diff
@@ -415,14 +415,10 @@ void Item_modifier::modify( item &new_item, const std::string &context ) const
...
         cont = container->create_single( new_item.birthday() );
-    }
-    if( cont.is_null() && new_item.type->default_container.has_value() ) {
-        const itype_id &cont_value = new_item.type->default_container.value_or( "null" );
-        if( !cont_value.is_null() ) {
-            cont = item( cont_value, new_item.birthday() );
-        }
+    } else if( cont.is_null() && new_item.type->default_container.has_value() ) {
```
mqrause, line 420:
```suggestion
    } else if( new_item.type->default_container.has_value() ) {
```

PR #62163, opened 2022-11-11
port: add info for MOUNTED_GUN flag

casswedson:
#### Summary
Interface "MOUNTED_GUN flag info in item details screen"

#### Purpose of change
from: https://github.com/cataclysmbnteam/Cataclysm-BN/pull/728

it's nice to have items with this flag say something, it pretty
much says the same thing if you attempt to fire guns with this flag but
knowing beforehand is good; display information if it is beneficial to
the player

#### Testing
it tells you can't fire anywhere but you didn't know (maybe)
Screenshot from 2022-11-10 19-02-06
info
Screenshot from 2022-11-10 19-02-23

### DIFF
--- a/data/json/flags.json
+++ b/data/json/flags.json
@@ -1454,6 +1454,7 @@
   },
   {
     "id": "MOUNTED_GUN",
+    "info": "This weapon is <bad>too unwieldy</bad> to fire on its own and <info>must be mounted</info> on a vehicle or furniture (window, table, mound of dirt, etc.) before use.",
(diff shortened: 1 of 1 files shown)


Fris0uman:
Maybe specify how to know what furniture qualifies?

casswedson:
that would need terrain, furniture etc. with `MOUNTABLE` to display some sort of information, what that would say I can't imagine, like: the characteristics of _this_ makes it suitable for deploying heavy machine guns.

PR #62167, opened 2022-11-11
fix: run summary labeler on open/edit pr body

casswedson:
#### Summary
None

#### Purpose of change
this makes sense, be specific about the events that trigger this workflow only run it when a pr is open or edited, instead of whatever it was set before, if someone knows how the workflow works they will use it to label their own pr how they like (more or less it's fairly limited)

I want to do that, sometimes

reference:
https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

```yml
on:
  pull_request_target:
    types: [assigned, opened, synchronize, reopened]
```

#### Testing
none I guess? just hope yaml won't pelt me for indentation mistakes and hope github gives us an actual github specific yml workflow syntax checker

### DIFF
--- a/.github/workflows/summary-labeler.yml
+++ b/.github/workflows/summary-labeler.yml
@@ -1,7 +1,8 @@
 name: Summary Labeler
 
 on:
-  - pull_request_target
+  pull_request_target:
+    types: [opened, edited]
 
 jobs:
   uses:


casswedson:
I think I changed my mind on the level of interactivity for this, leaving it on the low side is probably better, as is it's running on open and push, that's probably ok

PR #62175, opened 2022-11-11
baton fix

nopenoperson:
#### Summary
Bugfixes "baton weight and volume tweaks"

#### Purpose of change
a 60cm cylinder with a thick 1.06 inch grip is not anywhere close to 750 ml in volume.

#### Describe the solution
take measurements from https://www.asp-usa.com/collections/batons/products/talon-baton-60cm?variant=31227922087959, calculate thickness for the different sections, change numbers in the json, change the material, change deconstruction recipe.

reduce  bashing damage due to having lower movecost now, pick 11.35 or 12.9 as the new expected dps number.

removing the belt clip, creating a belt clip item mod, removing deconstruction recipe

#### Testing
it scales a bit harder with strength and skill now, for melee experts its worse against lategame enemies but better vs weak stuff.


Karol1223:
If you are changing the material you should take a look at their disassembly recipes and alter those as well. I'd like to avoid another 'material made from X gives resource that's from material Y' PR being needed.

Maleclypse (approved):

PR #62204, opened 2022-11-13
Mutation distractions no longer appear when NPCs mutate

### LINKED ISSUE #62177: Game thinks an NPC mutating is the player mutating
### Describe the bug

When an NPC companion mutates, the "you mutate!" popup appears, interrupting an activity like crafting/reading, as if your character was mutating.

### Steps to reproduce

1. Feed an NPC follower mutagen
2. Do an activity like crafting, reading or repairing
3. Wait for the NPC to mutate, the popup should appear.

### Expected behavior

Either the popup should read "<npcname> mutates! stop crafting?" or it should not appear at all.

### Screenshots

image

- OS: Windows
- Game Version: 6f0e11a [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm],
]
o version 19044.2251

### Additional context

_No response_

Ilysen:
#### Summary
Bugfixes "Activity interrupt prompts no longer appear when NPCs mutate"

#### Purpose of change

Fixes #62177

#### Describe the solution

The interrupt popups lacked an `is_avatar()` check, so I put one into both of the checks. This'll make it so that the messages only appear when the PC mutates, and not any other characters.

N/A

#### Testing

I spawned in an NPC and used Debug Mind Control to make them a follower, then slapped them and myself with a bunch of mutagen and beast primer. Then I waited for mutations to make sure that the distraction wasn't appearing, even though I had them enabled. They still appeared when I gained mutations myself.

#### Additional context

No interruptions!

image

### DIFF
--- a/src/mutation.cpp
+++ b/src/mutation.cpp
@@ -1529,7 +1529,7 @@ bool Character::mutate_towards( const trait_id &mut, const mutation_category_id 
     if( !mdata.dummy ) {
         set_mutation( mut, chosen_var );
     }
-    if( do_interrupt && uistate.distraction_mutation ) {
+    if( do_interrupt && uistate.distraction_mutation && is_avatar() ) {
         g->cancel_activity_or_ignore_query( distraction_type::mutation, _( "You mutate!" ) );
     }
 
@@ -1852,7 +1852,7 @@ void Character::remove_mutation( const trait_id &mut, bool silent )
                                    lost_name );
         }
     }
-    if( !silent && uistate.distraction_mutation ) {
+    if( !silent && uistate.distraction_mutation && is_avatar() ) {
         g->cancel_activity_or_ignore_query( distraction_type::mutation, _( "You mutate!" ) );
     }
 


andrei8l:
NPCs only mutate due to the player's actions so their mutations *should* trigger an interruption, either of a different type or at least with a different message.

Ilysen:
I considered that, but I felt like it clashed with all of the other interruption prompts, which occur when something is relevant to the PC's well-being or happens directly to them. it should be trivial to implement as a different distraction type in a follow-up PR, though, if that's something someone's interested in

Maleclypse:
> I considered that, but I felt like it clashed with all of the other interruption prompts, which occur when something is relevant to the PC's well-being or happens directly to them. it should be trivial to implement as a different distraction type in a follow-up PR, though, if that's something someone's interested in

I think that would be a neat followup, please.

Ilysen:
I can make it so once the time's right. what are the criteria for it; should it only happen when you visibly see a friendly NPC mutate? not care about LoS as long as they're in the bubble? should it care about affiliation (i.e. friendly, neutral...) or not? etc etc

Maleclypse:
> I can make it so once the time's right. what are the criteria for it; should it only happen when you visibly see a friendly NPC mutate? not care about LoS as long as they're in the bubble? should it care about affiliation (i.e. friendly, neutral...) or not? etc etc

I would think limiting it to friendly NPCs would be best and as long as they are in the bubble.  Though that could be an issue if they were ready to mutate and the mutations don't process until you bring the bubble to them.  If you make that option ping me and we can talk it over but I don't want to crowd this PR with a tangentially related request :)

PR #62209, opened 2022-11-13
Sam browne belt

tw1st31337:
#### Summary
Content "Added Sam Browne belt"

#### Purpose of change
I made it mainly so i can finish my military look, and thought about uploading it here so other people can also use it.

#### Describe the solution
You put it on, nothing else

None

#### Testing
I went ingame untill there were no more errors, went onto jsonlint untill it said the code was valid


esotericist:
it looks like you've done something to cause churn on the entirety of `data/json/itemgroups/Clothing_Gear/clothing.json`.

this is most often caused by using a defective text editor (such as windows notepad), or failing to lint the file properly (linting instructions are in the new contributor guide).

PR #62217, opened 2022-11-14
Create the polesaw.

Azurethorn:
#### Summary
Content "Adds Polesaws to the game."

#### Purpose of change

Pure content addition - adding a polesaw to the game.

Done due to popular interest for such a tool on reddit, and as a first attempt to adding content to CDDA as a whole for the coder.

#### Describe the solution

Adds polesaws to the game, a reach-based chainsaw with somewhat less damage but increased range. Spawns in similar spots as chainsaws, but slightly less or equal.

#### Testing

Tested the weapon itself, spawn rates, and crafting - it spawns in the world in the appropiate spots, it can be crafted, and it's damage is equal to it's contemporaries. 

#### Additional context

This is my first attempt at adding content to the game - so if there are any issues or things I've missed, I'll gladly fix.


Daved27hundred:
Are you sure it'd do 60 damage while on? I have never seen one of these tools being used but I imagine the blade is way smaller than the blade of a full on chainsaw, due to being used to cut through branches rather than trunks as well as for ease of use.

Bluarianknight:
> 

I did some testing - I kept it at 60 for two reasons.

1. Even with a decrease to 40, it becomes incredibly weak - it's just too big and clunky. Given my testing, an average character swings it roughly 250-300 move cost, even a pipe or a plank - or even weaker spears, just beat them. I didn't think it'd make sense for a polesaw to be beaten by a pointy stick in damage output. 

2. It's still a chainsaw in design. With the models I looked at, while it's a little smaller, it's still the same mechanisms - it'd still cut through reasonably well through dead. 

I'm just a little worried that decreasing it's damage output would make it completely inept in combat - which, while it surely shouldn't be a combat weapon, it should at least beat many of the weakest used weapon options, given it's fuel cost. 

I'll happily decrease it, but I thought I'd give my reasons to explain why first.

data/json/items/tool/woodworking.json
```diff
@@ -77,6 +77,46 @@
...
+    "id": "polesaw_off",
+    "type": "TOOL",
+    "name": { "str": "polesaw (off)", "str_pl": "polesaws (off)" },
+    "description": "This is a lumbering tool intended for trimming hedges and tree branches. If loaded with gas, using this item will cause it to turn on, making it into an unwieldy polearm.",
+    "weight": "9525 g",
+    "volume": "2500 ml",
+    "longest_side": "325 cm",
+    "price": 9000,
+    "price_postapoc": 450,
+    "to_hit": -4,
```
TheShadowFerret, line 90:
```suggestion
    "to_hit": { "grip": "solid", "length": "long", "surface": "line", "balance": "clumsy" },
```

data/json/items/tool/woodworking.json
```diff
@@ -77,6 +77,46 @@
...
+    "use_action": [ "CHAINSAW_OFF" ],
+    "flags": [ "NONCONDUCTIVE", "FRAGILE_MELEE", "REACH_ATTACK", "REACH3" ]
+  },
+  {
+    "id": "polesaw_on",
+    "copy-from": "polesaw_off",
+    "type": "TOOL",
+    "name": { "str": "polesaw (on)", "str_pl": "polesaws (on)" },
+    "description": "This polesaw is on and making a lot of noise.  Use it to turn it off.",
+    "to_hit": -5,
```
TheShadowFerret, line 109: "to_hit": { "grip": "solid", "length": "long", "surface": "line", "balance": "clumsy" },

TheShadowFerret (changes requested):

data/json/items/tool/woodworking.json
TheShadowFerret, line 109:
```suggestion
    "to_hit": { "grip": "solid", "length": "long", "surface": "line", "balance": "clumsy" },
```

data/json/items/tool/woodworking.json
```diff
@@ -77,6 +77,46 @@
     "use_action": [ "CHAINSAW_ON" ],
     "flags": [ "MESSY", "TRADER_AVOID", "NONCONDUCTIVE", "POWERED", "FRAGILE_MELEE" ]
   },
+  {
+    "id": "polesaw_off",
+    "type": "TOOL",
+    "name": { "str": "polesaw (off)", "str_pl": "polesaws (off)" },
+    "description": "This is a lumbering tool intended for trimming hedges and tree branches. If loaded with gas, using this item will cause it to turn on, making it into an unwieldy polearm.",
```
TheShadowFerret, line 84:
```suggestion
    "description": "This is a lumbering tool intended for trimming hedges and tree branches.  If loaded with gas, using this item will cause it to turn on, making it into an unwieldy polearm.",
```

PatrikLundell:
The argument that a tool used to cut branches is a worse weapon than a primitive but actual weapon is a poor one.

Such a tool would be incredibly unwieldy given how top heavy it is (like trying to use a sledge hammer as a spear). It works fine against stationary trees (the only kind we have IRL), but to swing it around trying to hit moving targets would be rather difficult. The reach would probably not be of much value given that the enemy can close the distance during the time it takes you to move it.
It can also be noted that these tools are intended to use their reach upwards at an angle, rather than horizontally, where you'd probably struggle to keep it up at full reach.
It's probably quite dangerous if it does connect though, but good luck managing to actually score a hit.

fairyarmadillo:
The reason something like this hasn't already been added is that it would be a terrible weapon. 60 damage is way too high and there's no way it should have the impale technique- the blade is not going to slide cleanly through an enemy like a spear would and the pole is too flimsy to stand up to somebody ramming it through a human body even it could.

Bluarianknight:
Is 35 a reasonable level of damage for this weapon, then? Given previous testing, it is quite bad.

Zireael07:
I'd fully expect it to be 'quite bad' in line with previous commenters

Bluarianknight:
I've changed it to a smaller version - No more REACH3, as well as limited the damage to 35 damage - which was a value I used previously in testing. It should be far less powerful as expected.

Bluarianknight:
From what I've looked at - a polesaw is still dangerous while on. Gas powered polesaws - which is what I've based the item on - are built to handle thicker branches. It's unwieldy, but if it can cut through large tree branches, it should deal at least some significant damage while pressing through a person.

Knut-Aage-Hofseth:
I think the idea is cool. The item is something that exist, and we should be able to press it into service as an improvised weapon. The problem is game balance. There is a huge difference between how much damage a tool can inflict to an unlucky user and how good it is as a weapon.
And how good it is as a weapon is connected to things that are difficult to model. There is of course the to-hit property. But other things like how well it bites into a target or glance off. How often the chain jams with flesh or clothes getting caught. Does the engine turning makes it harder to use. This has to be abstracted into to-hit, damage and swing speed. 

My personal thought is slow, low damage, not durable but with high crit damage (I do not know if that is possible) to represent the potential damage when it hits just right. Those horrible accidents you see on the news and not all the ones no-one hears about because it went well.

Bluarianknight:
That's true - I still think 35 is somewhat reasonable, given testing - I would certainly prefer to make to-hit harder if possible (which it already is with the spear property (I think) that makes near-attacks far harder. Given it's large action-cost, it's a risk to swing far, and not great at close.

nopenoperson:
is it based on a particular model of electric pole saw that exists irl? it seems quite heavy when compared to the models ive looked at online.

Bluarianknight:
It's based on gas-powered polesaws - which seem to be the longest and heaviest type of polesaws.

Bluarianknight:
If anyone has any more concerns, do let me know and I'll try to fix it - otherwise I'll mention it in the merge channel in the server if everyone thinks the changes make it work.

data/json/items/tool/woodworking.json
```diff
@@ -77,6 +77,46 @@
...
+    "looks_like": "chainsaw_off",
+    "material": [ "steel", "plastic" ],
+    "symbol": "/",
+    "color": "red",
+    "ammo": [ "gasoline" ],
+    "pocket_data": [ { "pocket_type": "MAGAZINE", "ammo_restriction": { "gasoline": 450 }, "watertight": true, "rigid": true } ],
+    "charges_per_use": 5,
+    "techniques": [ "SWEEP" ],
+    "use_action": [ "CHAINSAW_OFF" ],
+    "flags": [ "NONCONDUCTIVE", "FRAGILE_MELEE", "REACH_ATTACK" ]
```
Paprika-H, line 101:
```suggestion
    "flags": [ "NONCONDUCTIVE", "FRAGILE_MELEE", "REACH_ATTACK", "POLEARM" ]
```
If the powered version should have polearm I think the off version should have it too

data/json/items/tool/woodworking.json
```diff
@@ -77,6 +77,46 @@
     "use_action": [ "CHAINSAW_ON" ],
     "flags": [ "MESSY", "TRADER_AVOID", "NONCONDUCTIVE", "POWERED", "FRAGILE_MELEE" ]
   },
+  {
+    "id": "polesaw_off",
+    "type": "TOOL",
+    "name": { "str": "polesaw (off)", "str_pl": "polesaws (off)" },
+    "description": "This is a lumbering tool intended for trimming hedges and tree branches.  If loaded with gas, using this item will cause it to turn on, making it into an unwieldy polearm.",
```
Paprika-H, line 84:
```suggestion
    "description": "This is a lumbering tool intended for trimming hedges and tree branches at a distance.  If loaded with gas, using this item will cause it to turn on, making it into an unwieldy polearm.",
```

data/json/items/tool/woodworking.json
```diff
@@ -77,6 +77,46 @@
...
+    "weight": "9525 g",
+    "volume": "2500 ml",
+    "longest_side": "275 cm",
+    "price": 9000,
```
Paprika-H, line 88:
```suggestion
    "price": 40000,
```
Price is in decimals, and I couldn't find a gas one for $90, $400 seems a bit more reasonable, but that's just my opinion.

data/json/items/tool/woodworking.json
Paprika-H, line 101:
```suggestion
    "flags": [ "POLEARM", "NONCONDUCTIVE", "FRAGILE_MELEE", "REACH_ATTACK" ]
```
Both versions should have the polearm flag

data/json/items/tool/woodworking.json
```diff
@@ -77,6 +77,46 @@
...
+    "copy-from": "polesaw_off",
+    "type": "TOOL",
+    "name": { "str": "polesaw (on)", "str_pl": "polesaws (on)" },
+    "description": "This polesaw is on and making a lot of noise.  Use it to turn it off.",
+    "to_hit": { "grip": "solid", "length": "long", "surface": "line", "balance": "clumsy" },
+    "bashing": 4,
+    "cutting": 35,
+    "turns_per_charge": 4,
+    "charges_per_use": 0,
+    "revert_to": "chainsaw_off",
```
Paprika-H, line 114:
```suggestion
    "revert_to": "polesaw_off",
```
Otherwise it'd turn into a chainsaw

Holli-Git (changes requested):

Bluarianknight:
Done!

Daved27hundred:
This got me thinking, should chainsaws and similar weapons have a special attack? I don't think you'd slice or stab with them like a normal blade, You'd probably dig through a large portion of the enemy before pulling the saw back.

Paprika-H:
> This got me thinking, should chainsaws and similar weapons have a special attack? I don't think you'd slice or stab with them like a normal blade, You'd probably dig through a large portion of the enemy before pulling the saw back.

I think a realistic attack would probably be the first swing and the continuous damage until you retreat, or more realistically it getting gummed up

PR #62233, opened 2022-11-15
Railgun name and description

wishki:
#### Summary
Content "Update railgun name and description"

#### Purpose of change

Futuristic railgun has no real world equivalent so does not need manufacturer name

#### Describe the solution

Update gun, magazine and ammo definitions


Fris0uman:
Why remove the name though?

wishki:
> Why remove the name though?

Its not accurate?

hexagonrecursion (changes requested):
Real companies can have counterparts in a sci-fi setting making sci-fi tech. hk_g80 can be found in the "military" and "mil_hw" item groups. This appears to indicate that it was made by humans (and not for example exodii). Does this contradict the game lore? if not, this PR should be closed.

https://github.com/CleverRaven/Cataclysm-DDA/blob/bd7a3976e465a68c5d16eb236bdfd6c7e72c18f7/data/json/itemgroups/military.json#L679
https://github.com/CleverRaven/Cataclysm-DDA/blob/bd7a3976e465a68c5d16eb236bdfd6c7e72c18f7/data/json/itemgroups/military.json#L871

PR #61880, opened 2022-10-23
Build Windows Release with CMake (part 2)

alef:
#### Summary
None

#### Purpose of change
Try to align CMake build to match the release.

#### Describe the solution
- [x] Tidy up `CMakeModules/`
- [x] Provides draft documentation
- [x] Uses `release.yml` names for presets and scripts
- [x] Uses `vswhere.exe`

More details in commits comments.

#### Testing
It builds locally on VS 2022
1. Drafting this PR
2. Watch CI build
3. Undraft if no side-effects

#### Additional context
Part of discarded https://github.com/CleverRaven/Cataclysm-DDA/pull/60159

PR #62234, opened 2022-11-15
Correct CMake install locations on *nix

perryprog:
#### Summary
None

#### Purpose of change
PR #61225 changed the location some stuff got installed in, which while correct for the Windows style install location, is incorrect for the default *nix style install locations. (`CMAKE_INSTALL_DIR` is by default `/usr/local` instead of `c:/Program Files/${PROJECT_NAME}/` according to [here](https://cmake.org/cmake/help/latest/variable/CMAKE_INSTALL_PREFIX.html).)

#### Describe the solution
Adjust these install locations to be correct for both OS's, hopefully. I'm not sure what the install locations for `gfx/` and `doc/` are intended to be for Windows, so I've set it to what it was previously, which is under `DATA_PREFIX`. I'd like it if this can be confirmed before the PR is merged so I can make the default install locations for both of those directories also depend on OS.

#### Testing
Builds and runs on macOS. I'd appreciate it if someone on Windows can test.

#### Additional context
CC @alef: I'd like to know what would be the correct directories according to what you were trying to do.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -124,6 +124,9 @@ if ("${CMAKE_SYSTEM_NAME}" MATCHES "(Linux|FreeBSD|Darwin)")
     if (NOT DATA_PREFIX)
         set(DATA_PREFIX ${CMAKE_INSTALL_PREFIX}/share/cataclysm-dda)
     endif ()
+    if (NOT README_DIR)
+        set(README_DIR ${DATA_PREFIX})
+    endif ()
     if (NOT LOCALE_DIR)
         set(LOCALE_DIR ${CMAKE_INSTALL_PREFIX}/share/locale)
     endif ()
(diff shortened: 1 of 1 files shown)


alef:
The CMake builds do not install correctly, it should be fixed in https://github.com/CleverRaven/Cataclysm-DDA/pull/61880.
I suggest you rebase this PR after mine gets merged (@akrieger is back)

akrieger:
I mean I'm still *on* vacation but I found time to hop on discord today :p

perryprog:
Ah, thanks, I didn't realize you had opened the second PR already. Because install is effectively broken on *nix right now, do we want to merge this as a temporary fix? (I'm thinking it's find to wait considering I seem to be the first person to notice, unless there's an issue I missed.)

Night-Pryanik:
@perryprog could you please resolve conflicts?

perryprog:
All done; sorry for the delay. @alef, can you confirm this still installs as expected for Windows? It looks good to me on macOS.

Issue #59873, opened 2022-08-03
`professional_zombie_gear` item group can cause errors on monster death


mqrause:
### Describe the bug

It is possible a monster death can create an error message leading to CI test fails.
https://github.com/CleverRaven/Cataclysm-DDA/runs/7653184359?check_suite_focus=true#step:16:430

### Steps to reproduce

I have not been able to reproduce the issue with the seed from the failed test and in another 60 random test runs locally.

However digging into the test code I can fairly safely conclude the cause is the chance for `professional_zombie_gear` in the `default_zombie_death_drops` item group.
`professional_zombie_gear` contains the group `hardware_plumbing`, which in turn contains `plumbing_clothing` which has the explicit comment `"//2": "Not intended to be a set of clothing for NPCs or monsters"`, most likely because it does not guarantee in any way that proper container items are spawned for any additional loot to be put in, since it seems to only spawn one item from the list and not a set of gear the zombie would be wearing.

### Expected behavior

Items spawn without issues. However in this case it might be unlikely enough to not be a problem.

### Versions and configuration

bc8d87c8203708c2580c89370a253e190f1ef7e4

Stadler76:
I assume, that this one is related: https://github.com/CleverRaven/Cataclysm-DDA/runs/7708088941?check_suite_focus=true#step:16:259

mqrause:
Presumably another case: https://github.com/CleverRaven/Cataclysm-DDA/runs/7868235039?check_suite_focus=true#step:16:568

PR #62236, opened 2022-11-15
Update Aftershock dead survivor loot item group

### LINKED ISSUE #62220: afs_corpse_old fails on item group test
### Describe the bug

Likely related to new corpse behaviour after https://github.com/CleverRaven/Cataclysm-DDA/pull/61575

https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/3464245256/jobs/5785540569

(all_mods)=> (continued from above) ERROR : src/item.cpp:1501 [ret_val<void> item::put_in(const item &, item_pocket::pocket_type, const bool)] tried to put an item (whistle) count (1) in a container (afs_corpse_old) that cannot contain it: Can't store anything in this.

### Steps to reproduce

Run test suite with `--mods=dda,aftershock`, fails ~4 out of 5 times

### Expected behavior

Tests should pass

### Screenshots

_No response_

Latest master 65e5b97df3bee495c495e90ae5fccf630637e40f

### Additional context

_No response_

BrettDong:
#### Summary
None

#### Purpose of change
* Fix #62220 

#61575 forbids corpses containing items. This change breaks an item group defined in the Aftershock mod. This is causing errors in basic build test in almost all recent pull requests.

#### Describe the solution
Update the broken item group to spawn the corpse alongside with the loot items.

#### Testing
`./tests/cata_test --mods=dda,aftershock "~*"` no longer errors.

### DIFF
--- a/data/mods/Aftershock/itemgroups/loot_corpse_groups.json
+++ b/data/mods/Aftershock/itemgroups/loot_corpse_groups.json
@@ -3,9 +3,9 @@
     "id": "afs_dead_survivor",
     "type": "item_group",
     "subtype": "collection",
-    "container-item": "afs_corpse_old",
     "//": "The corpse of someone who managed to survive the Discontinuity for a while. Can spawn common weapons and misc gear.",
     "entries": [
+      { "item": "afs_corpse_old", "prob": 100 },
       {
         "distribution": [
           { "item": "afs_tattered_garment", "prob": 90 },


Fris0uman:
It should spawn correctly in game though, this https://github.com/CleverRaven/Cataclysm-DDA/pull/61575/commits/393f630db7729f07a000b2d8149bf56e045ba436 fixed the error when loading the game at least. Maybe we need to do somethign similar for the test run?

Fris0uman:
So you mean it doesn't actually work in game?

mqrause:
I'm not 100% sure how these item groups with corpses work. But as far as I understand they're supposed to create corpses that "wear" stuff with the worn stuff containing other items. Not sure if this fix will still produce that result or if it will spawn a corpse next to some other items. If it's the latter, then the real issue is more like #59873: some item group doesn't have appropriate container items to contain all the additional things it spawns.

Fris0uman:
Okay so I don't know where I messed up my test. Still I think it would be better to find a c++ fix so that the items are still put into corpses as if they were wearing it rather than droping it next to them. Plus in vanilla the mapgened corpse can revive and they'll keep their loot if they have it inside the pocket

mqrause:
Looking into it a bit more, the issue seems to be even still different:
https://github.com/CleverRaven/Cataclysm-DDA/blob/d1f765d85f6cf7df6d0fe07fa676c3a639b3f70e/src/item_group.cpp#L126-L128
`item::can_contain` will also check all nested pockets of the container, however `item::put_in` can only put things directly in the container. That's why it sometimes tries to insert things into a corpse the normal way instead of forcing it in or putting it into an appropriate container. So that needs to be sorted out one way or the other.

PR #62237, opened 2022-11-15
Post a warning if a pull request is created from the master branch

BrettDong:
#### Summary
None

#### Purpose of change
It's a common mistake for a new contributor without Git experience to directly work on the master branch on their fork and create a pull request from their master branch. This pull request adds a workflow that automatically posts a warning if the head branch of a new pull request is `master`.

#### Describe the solution
The workflow is triggered by a pull request `opened` event and runs if the head branch name is `master`.

### DIFF
--- a/.github/workflows/check-branch-name.yml
+++ b/.github/workflows/check-branch-name.yml
@@ -1,0 +1,24 @@
+name: Check pull request head branch name
+
+on:
+  pull_request_target:
+    types:
+      - opened
+
+jobs:
+  check_branch_name:
+    runs-on: ubuntu-latest
+    if: github.head_ref == 'master' && github.repository == 'CleverRaven/Cataclysm-DDA'
+    steps:
+      - name: Post warning
+        uses: actions/github-script@v6.3.3
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            await github.rest.issues.createComment({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              issue_number: context.issue.number,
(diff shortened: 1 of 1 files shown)


RenechCDDA:
It should check for 'main' as well, as (from what I understand) that is going to be the new default name moving forward, although there is no ETA on when further changes are expected. https://sfconservancy.org/news/2020/jun/23/gitbranchname/

Unsure if this will ever apply to forks of cdda (which, presumably, will continue with 'master'), but just in case?

.github/workflows/check-branch-name.yml
```diff
@@ -0,0 +1,21 @@
...
+    if: github.head_ref == 'master' && github.repository == 'CleverRaven/Cataclysm-DDA'
+    uses: actions/github-script@v6.3.3
+    with:
+      github-token: ${{ secrets.GITHUB_TOKEN }}
+      script: |
+        await github.rest.issues.createComment({
+          owner: context.repo.owner,
+          repo: context.repo.repo,
+          issue_number: context.issue.number,
+          body: "You are creating a merge request with the master branch as the head branch. This is likely a mistake unless you really know what you are doing. You may read https://docs.github.com/en/get-started/quickstart/contributing-to-projects#creating-a-branch-to-work-on for a typical workflow of contributing to a project on GitHub."
```
casswedson, line 19:
```suggestion
          body: "You are creating a pull request with the master branch as the head branch. This is likely a mistake unless you really know what you are doing. You may read https://docs.github.com/en/get-started/quickstart/contributing-to-projects#creating-a-branch-to-work-on for a typical workflow of contributing to a project on GitHub."
```

casswedson: if I were to suggest something to someone new to github, I would try to not refer to one thing with, say, 2 different names; merge request may mean the same thing as pull request but keeping it simple for new contributors is nice

BrettDong:
> Unsure if this will ever apply to forks of cdda (which, presumably, will continue with 'master'), but just in case?

I don't think it'll be the case. We can update the script when it really happens in the future.

Issue #56241, opened 2022-03-20
Unhardened leather should have less cut resistance, add leather hardening and rework leather armor to use hardened leather


avianArsonist:
### Is your feature request related to a problem? Please describe.

All resources I could find point to unhardened leather being less cut resistant, by weight, than cotton ([cw: blood in initial image] https://www.nationalsafetyinc.com/Assets/Downloadables/Understanding%20Cut%20resistance.pdf, https://www.grainger.com/know-how/safety/ppe-in-the-workplace/hand-protection/kh-choosing-the-right-safety-gloves). I can't find any sources on leather's impact resistance, but it could possibly use a buff to compensate.

Leather can, however, be hardened. (https://www.youtube.com/watch?v=hwGW_qwpxYs, https://medium.com/@jasontimmermans/a-comparative-study-of-leather-hardening-techniques-16-methods-tested-and-novel-approaches-8574e571f619). This would likely be stronger than the current unhardened leather, and would probably be a much better material for armor. 

### Solution you would like.

Change leather's cut resistance from 3 to 1, add hardened leather as a material, replace leather with hardened leather in leather armors, and add hardening to leather armor recipes, preferably in a manner that allows you to work on other crafts while waiting for it to harden.

### Describe alternatives you have considered.

Increasing leather's bash resistance to 2 to compensate for lowered cut resistance, assuming leather armor is hardened by default and adding fake thickness to simulate this, doing more research on leather's impact resistance.

bean-b:
If nobody else is working on this, I may take a stab at it

Drew4484:
I would make alternate versions of the leather armors instead of modifying the existing items, that allows for separate recipes to account for the various materials.

bean-b:
> I would make alternate versions of the leather armors instead of modifying the existing items, that allows for separate recipes to account for the various materials.

Personally, I dont think that making leather body arm from cut up leather pants and gloves seems all that realistic. What if the Leather padded pants were exapnded upon, with makeshift level leather armor for arms torso etc. and then the leather body armors were reserved for the treated leathers?

Drew4484:
> > I would make alternate versions of the leather armors instead of modifying the existing items, that allows for separate recipes to account for the various materials.
> 
[...]

PR #62256, opened 2022-11-16
Added hardened leather

bean-b:
#### Summary
Content "Added Hardened Leather"

#### Purpose of change
un-hardedened leather isnt as protective as it is in game, but hardening reperesnents a signifigant increase in its protection values. Therefore, they should be two different materials

#### Describe the solution
Added new material, put it into armors and recipes as nescary.

There is a poteitnal reason to change leather to be graded (i.e. light meduim heavy), but that seemed like a lot to change, and hardened leather was a good simpler solution for now. Also values could of course be changed. I think regular leather could have even less resistances, but going down a full 0.5 seemed like a lot on any given value, going to 0.25 or 0.75 will leave a lot of unsatisfying numbers for common clothing. 

#### Testing
Spawned items in game, looked at recipes

#### Additional context
closes (hopefully)
https://github.com/CleverRaven/Cataclysm-DDA/issues/56241
hardening process:
https://medium.com/@jasontimmermans/a-comparative-study-of-leather-hardening-techniques-16-methods-tested-and-novel-approaches-8574e571f619


bean-b:
I'm not quite sure why it requested a code review, i still need to lint it and a few things, so sorry if you got pinged on that if you shouldnt have

bean-b:
Sorry this pr is such a mess lol, was gonna lint it after pulling but then ended up linting it through all the bots comments lol

RenechCDDA:
Okay... a few things. First, it seems you're also adjusting the values of regular leather. After the given changes, leather will be less protective vs all damage types than... canvas. Or faux fur. Or... lycra?? Maybe that's correct, but I'd like to see at least some reasoning for it.

Similarly, you are *adding* hardened leather as a material and giving it to a wide variety of existing items. It has superior stats to the old leather (representing boiled leather, a type of hardened leather as outlined by your medium page). Why? Let's see some reasoning, comparisons of how it holds up as compared to canvas, that sort of thing. NIJ test data for stab vests or something of the sort would probably be the best possible data. Mostly, I want to see data *comparing it to other materials*.

Your medium article indicates that ammonia-treated "hardened leather" has less puncture resistance ("average force in kg required to push the awl to 5mm depth in the leather") at 4.87kg for ammonia-treated vs 5.53kg for boiled, which we currently have ingame. Thus it doesn't make much sense to me that the cut and bullet resistance of your new material, produced by the ammonia-treated method, should be *increased* over the existing boiled leather.

Your medium article indicates ammonia-treated "hardened leather" loses approximately 63% of its water resistance. If so, existing items should probably lose their `WATERPROOF` and/or `WATER_FRIENDLY` tags. I don't see any changes on this, so maybe they never had them, but please make sure and check if you haven't already.

It seems you've made changes to.... the lorica segmentata? The changes seem to add material definitions to its armor, and change the overall item's composition from iron/leather to steel/leather. The old leather, too. I am not sure that is what you intended to commit.

This is not intended to be an exhaustive review, just some things I saw while looking over the total changes.

fairyarmadillo:
> Okay... a few things. First, it seems you're also adjusting the values of regular leather. After the given changes, leather will be less protective vs all damage types than... canvas. Or faux fur. Or... lycra??

My impression from many hours of play in the past year is that we might need a full material audit. Faux fur and lycra have really overblown values, the game seems to treat lycra as this miracle material and not fragile itchy spandex.

I haven't gone over OP's numbers, but fabrics in general may be due for a reckoning.

RenechCDDA:
It could very well be that Canvas, faux fur, and lycra are all wrong. (I agree that Lycra in particular seems to be a bit of a 'miracle material'). But what I'd like to see is some reasoning, even just like... a text explanation here saying "I took the values from the medium.com link and tried to apply them as in [such and such manner]". For example, look at the comments for the steel grades which reference industrial standards. If we wanted to compare the malleability or shear strength or *any other value* of a given metal we could easily compare to that. That's good data, easy to compare, makes a strong case that it is a valid material and different from other materials in the specific ways it is implemented as being different.

I don't think the PR needs to verify every inch of every related material, it just needs to reasonably explain why it's been placed at the values it has been.

LyleSY (approved):
The DinoMod components here look reasonable and consistent with what you're trying to do. I support the overall approach though I haven't dug into any specifics

bean-b:
> As far as I am aware, the thickness values were explicitly added by yourself in #62196.
No I did choose the thickness of boiled leather armor, but that was because there wasnt any alternative material to have them made out of, and also because that was the way that boiled armor was before #62196, and i wasnt trying to add a new material in that pr.
>...Speaking of which, your recipe for sheet_leather_hardened_patchwork calls for a 50% solution of ammonia-water, far in excess of what is suggested. Or an astounding 2L of animal fats per ~186mL of hardened leather. You probably want to check those numbers.
No that makes sense Ill do that when i get a chance
> The only thing I would find questionable about it is the use of rendered animal fats, but that's easy to fix without completely removing the recipe and its attendant item. The only difference between the treatment provided for in the existing boiled leather armor and your sheet_leather_hardened_patchwork is the addition of ammonia solution as a possible sealing agent.
And yes the hardening process i very similar to what we have before. If it seemed like it would be better for hardened leather to instead be boiled leather and not use anything other then water + heat, I could adjust the recipe accordingly, but if it seems better with the adition of sealing/hardening agents, then i can deftinly go through and adjust the amounts it calls for

bean-b:
Also, if anyone knows what to change to get this test to pass that would be awesome:

 (~[slow] ~[.],starting_items)=> ../tests/item_test.cpp:734: FAILED:
(~[slow] ~[.],starting_items)=>   CHECK( item_density( target ) <= max_density )
Error: (~[slow] ~[.],starting_items)=>   1.74731f <= 1.2f

anoobindisguise:
I remember there being someone who posted sources suggesting soft leather wasn't even better than wet cotton for cutting resistance. Don't remember where it was though. While leather is actually fairly stiff and would be decent at absorbing incoming force. So you could quite honestly leave standard leather at 2 bash, 1 cut, 1-1.5 ballistic, then make hardened leather 2.5/2.5-3 or so. I assume soft leather is for stuff like jackets, dusters, corsets, purses, while hard leather is for boots and vaguely ahistoric armor? 
all our material values are pretty arbitrary so it's understandable that none of it makes much sense atm though.

bean-b:
> I remember there being someone who posted sources suggesting soft leather wasn't even better than wet cotton for cutting resistance. Don't remember where it was though. While leather is actually fairly stiff and would be decent at absorbing incoming force. So you could quite honestly leave standard leather at 2 bash, 1 cut, 1-1.5 ballistic, then make hardened leather 2.5/2.5-3 or so. I assume soft leather is for stuff like jackets, dusters, corsets, purses, while hard leather is for boots and vaguely ahistoric armor? all our material values are pretty arbitrary so it's understandable that none of it makes much sense atm though.

Yea, leather was mentioned here https://github.com/CleverRaven/Cataclysm-DDA/issues/56241, which was part of why I wanted to go through and make these changes. However, judging from this source leather does have more resistance to puncturing then cotton, and while I couldnt find a great source on cutting, I think along with some my general intuition says that leaather should have more then 1 protection value. This was futher complicated looking at materials like lycra, which has 2 protection value. I could see bringing the resitance to like 2, 2, 1.5 for leather though if you think that would be better. And yes the intention is that leather is for modern clothes and general usage, while hard leather is mainly for crafted armors

bean-b:
This should be merged in after the stable realize if it is merged

GuardianDll:
Why closed?

bean-b:
> Why closed?

I saw that #56241 was closed as not planned, so I was'nt sure if that means this isn't planned either?

NetSysFire:
I did the deed. Stalebot is running berserk. I think we need more triage team members instead of that lousy bot. Closing issues doesnt magically fix bugs or balance issues.

PR #62262, opened 2022-11-17
add weight to crystallized_mana

### LINKED ISSUE #61980: Magiclysm crystallized_mana generates Zero weight error despite ZERO_WEIGHT flag
### Describe the bug

 DEBUG    : Item 'crystallized mana' with zero weight

 FUNCTION : int item::charges_per_weight(const mass&) const
 FILE     : src/item.cpp
 LINE     : 1227
 VERSION  : 41acb30

### Steps to reproduce

Spawn in crystallized mana with debug menu.

Drop on floor.

Pick up.

Read error message.

### Expected behavior

I expect to be able to pick the item up without an error message as the item has the ZERO_WEIGHT flag.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.22621.674 (22H2)
- Game Version: 41acb30 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm]
]

### Additional context

 DEBUG    : Item 'crystallized mana' with zero weight

 FUNCTION : int item::charges_per_weight(const mass&) const
 FILE     : src/item.cpp
 LINE     : 1227
 VERSION  : 41acb30

Knut-Aage-Hofseth:
#### Summary
Bugfixes "#61980 crystallized_mana generates Zero weight error "

#### Purpose of change
Fixes #61980

#### Describe the solution
Wanted to fix #61980. Does this by adding weight of 1 mg to crystallized mana and removing ZERO_WEIGHT flag. Followed the suggestion to use mg instead of g that came from RenechDDA to have virtually no weight. I didn't know that mg was valid at first.

Figure out how to make item::charges_per_weight handle ZERO_WEIGHT flag.

#### Testing
Spawn in 100 crystallized mana with debug menu. Pick it up. Error message should not appear.

### DIFF
--- a/data/mods/Magiclysm/items/mana_crystals.json
+++ b/data/mods/Magiclysm/items/mana_crystals.json
@@ -15,7 +15,7 @@
     "color": "yellow",
     "description": "Some crystallized mana.  This can be reloaded into rechargeable mana crystals, but can never be unloaded.",
     "flags": [ "TRADER_AVOID" ],
-    "weight": "1 g",
+    "weight": "1 mg",
     "volume": "100 ml",
     "ammo_type": "crystallized_mana",
     "material": [ "crystallized_mana" ],


RenechCDDA:
Note that you can define weight in mg as well - so 1mg is an option. I think this would be preferable, as otherwise it would weigh 100g per 100mL (due to `count` size) of mana, in other words equivalent in density to water. Water is somewhat heavy! Using 1mg gives us 100mg per 100mL, a thousand times less, and as close to weightless as you can get without *actually* being weightless.

data/mods/Magiclysm/items/mana_crystals.json
```diff
@@ -14,7 +14,8 @@
     "symbol": "=",
     "color": "yellow",
     "description": "Some crystallized mana.  This can be reloaded into rechargeable mana crystals, but can never be unloaded.",
-    "flags": [ "TRADER_AVOID", "ZERO_WEIGHT" ],
+    "flags": [ "TRADER_AVOID" ],
+    "weight": "1 g",
```
Knut-Aage-Hofseth, line 18:
```suggestion
    "weight": "1 mg",
```

Knut-Aage-Hofseth:
> Note that you can define weight in mg as well - so 1mg is an option. I think this would be preferable, as otherwise it would weigh 100g per 100mL (due to `count` size) of mana, in other words equivalent in density to water. Water is somewhat heavy! Using 1mg gives us 100mg per 100mL, a thousand times less, and as close to weightless as you can get without _actually_ being weightless.

You are of course correct. I didn't know about mg. That is a better solution.

MNG-cataclysm (approved):

PR #62268, opened 2022-11-17
No more "right from the factory" loot in the museum

GuardianDll:
#### Summary
Bugfixes "Museum now contain damaged loot instead of brand new"
#### Purpose of change
When i last visit museum irl, the stuff here was in way worse condition, barely useful, comparing to current pinata in the game
#### Describe the solution
Made most museum loot to spawn in `| .` - `XX` quality
Also removed the leather armor, as proposed in comments
Make more precise changes, where some stuff may be less damaged
Edited museum to have more consistent loot, and not just mix of revolvers, flintlock mushkets, stone spears and kabuto
#### Testing
Opened the game, teleported to the museum, checked the loot, found all stuff are damaged
#### Additional context
image


bean-b:
I'm wondering if it should spawn in like damage 2-4? While you would expect most of it to be damaged, not every single piece is 100% damaged

data/json/itemgroups/Clothing_Gear/clothing.json
```diff
@@ -3217,84 +3217,93 @@
...
-      [ "armor_lamellar", 20 ],
-      [ "armor_lorica", 25 ],
-      [ "armor_samurai", 50 ],
-      [ "armor_larmor", 40 ],
-      [ "gambeson", 50 ],
-      [ "armor_cuirass", 25 ],
-      [ "chainmail_junk_suit", 10 ],
-      [ "chainmail_junk_vest", 3 ]
+      { "item": "armor_junk_plate", "prob": 60, "damage": [ 4, 4 ] },
+      { "item": "armor_junk_lightplate", "prob": 45, "damage": [ 4, 4 ] },
```
Maleclypse, line 3253: junk plate already has damage reflected in it's base stats.  

Also variety is probably best.   Cuts down on the pinata aspect but leaves a location still as a location of interest.

bean-b:
Historical examples of leather armor are super rare, so I wonder if they should be removed from the spawning list?

GuardianDll:
> While you would expect most of it to be damaged
Because no one consult me how damaged it must be

> Historical examples of leather armor are super rare
Ok, I'll remove it if it will be agreed

data/json/itemgroups/Clothing_Gear/clothing.json
GuardianDll: Do you think i need to make junk plate to be in perfect condition?

snipercup:
What would realistically cause damage to the contents of museums? Aren't they behind protected glass? Is there damage to the building too? Your PR description doesn't describe your reasoning for the damage.

Maleclypse:
> What would realistically cause damage to the contents of museums? Aren't they behind protected glass? Is there damage to the building too? Your PR description doesn't describe your reasoning for the damage.

He's saying that it's not that these items have received damage in the cataclysm, he's saying that they were found and brought to the museum damaged and then age has continued to deteriorate those items at a slower pace once they are at the museum.  But basically armor that spent thirty years even underground before it was discovered in archeology of a battlefield probably isn't useful as armor anymore.

GuardianDll:
maybe because its an old, rusty armor and weapon, and even with good maintenance it still wear out of the time?
in fact it must be separated to differ: some museums are with rifles and revolvers, some with WW2 guns, like that
but its too big for me to do now

snipercup:
Thank you for adding your justification.

kevingranade:

https://discourse.cataclysmdda.org/t/i-dont-understand-the-nerf-to-musems/28377/12

PR #62269, opened 2022-11-17
increased junk platemail encumbrance

bean-b:
#### Summary
Content "Removed varsize from junk platemails, increased encumbrance"

#### Purpose of change
fitted Junk platemail has a very small 10 encumbrance, much less then any other platemail that isnt falling apart and is made for combat. Similarly junk ornamental platemail has 20 encumbrance when fitted, which is less then even medium platemail, despite being ornamental, "extremely heavy" and impractical.

#### Describe the solution
removed var size, slightly changed encumbrance values

Keeping varsize, increasing enumbrace. Editing the junk suits so they use the new coverage system and dont have 2.5mm of leather everywhere.

#### Testing
spawned in game


Aerin-of-the-Toast:
Ornamental plate armor had 35 encumbrance when fitted before, not 20. If the item's encumbrance is higher than 20, fitting only reduces encumbrance by 10; otherwise it reduces it by half if 20 or lower.

I definitely agree with the encumbrance being much higher, but I feel like they should still have VARSIZE since it can come in sizes that are relatively close to your body, but, obviously, still higher than proper plate armor (even when fitted) because it's not tailored specifically to you. Otherwise I agree with your changes, only thing I'd do is re-add VARSIZE and increase the encumbrance by 10 more to account for that (so that they're the same encumbrance as you have them currently set to when fitted).

bean-b:
> Ornamental plate armor had 35 encumbrance when fitted before, not 20. If the item's encumbrance is higher than 20, fitting only reduces encumbrance by 10; otherwise it reduces it by half if 20 or lower.
> 
> I definitely agree with the encumbrance being much higher, but I feel like they should still have VARSIZE since it can come in sizes that are relatively close to your body, but, obviously, still higher than proper plate armor (even when fitted) because it's not tailored specifically to you. Otherwise I agree with your changes, only thing I'd do is re-add VARSIZE and increase the encumbrance by 10 more to account for that (so that they're the same encumbrance as you have them currently set to when fitted).

Sure thing

PR #62276, opened 2022-11-18
Metal lamellar armor, part 1, the sheetening

Aerin-of-the-Toast:
#### Summary
Content "Sheet metal lamellar armor."

#### Purpose of change

My quest to make everything ever out of sheet metal continues. Just you wait until I perfect my recipe for sheet metal tacos. For a serious answer, lamellar armor was a versatile, *extremely* common armor used the whole world over, for quite a lot of reasons. Also, sheet metal just happens to be about the right thickness for lamellae (give or take half a milimeter), and lets you skip the most laborious part of steel lamellar (forging about a gazillion lamellae). So, with the power of modern power tools, lamellar sheet metal armor is surprisingly easy to make and very effective.

#### Describe the solution

Added two variants of lamellar armor; one made from mild steel and an upgraded case-hardened version. Since the armor doesn't cover the shins or forearms (I could not find a *single* example of lamellar that covered there), and there's a bit of a glut of forearm/shin-only armor, I also added sheet metal bracers and greaves, which are basically the arm/leg guards but missing the upper arm/leg part. Most images I saw of lamellar armor included a solid bracer instead of lamellar anyway.

The craft for lamellar was split into a two-part craft for a few reasons. One was for repairing; the lamellar armors have "NO_REPAIR" because them being repaired with a welder makes no sense, and instead have "repair recipes" somewhat similar to the fire drill, where you just re-craft the armor using the armor as a component alongside some extra lamellae and cordage, to represent tying in new lamellar where's it's been knocked off. The armor also doesn't have "STURDY," so it will require maintenance more often than other armors, but that's helped by it being easy to repair.

Some oddities you might notice; the extremely high cordage requirement is actually pretty realistic; it takes a lot of cord to tie a bajillion small plates around your body! Most things just have thread/cordage requirements that are way too low. Myeh.

The recipe, meanwhile, is actually quite easy. Since we cheat past any forging necessary by using sheet metal, the tools needed are fairly minimal. Though, trying to drill all those holes with a hand drill takes extremely long, so there's two recipes for the lamellae; one with the hand drill and one with a cordless drill/drill press that takes 1/12th the time. The only real downside is requiring at least one of the armor books to make it. Doesn't matter which one, lamellar was used basically everywhere from Rome to China to Japan, but sadly the average person probably doesn't know about lamellar so it's relegated to book-land.

As for layers, proper lamellar's lamellae was layered close enough together that there's a lot of overlap, the amount of area not overlapping is pretty minimal, so it's just treated as one big contiguous sheet. A lamellae is 1mm thick, so the full armor is 2mm thick to account for double-layered lamellae. The leather lamellar armor was also adjusted and buffed accordingly.

Just doing all the forged versions now. Splitting up the arm guards as a separate item (the skirt kinda needs to be attached though). Copper and bronze lamellar, but I don't feel like making a molding system or w/e that'd require. Assume the gaps between lamellae are larger and make the lamellae similar to the previous leather lamellae (100% 1mm thick and 40% 1mm thick).

#### Testing

Checked all the recipes and items in-game, they seem to be working well, though it's always possible I miss something in a PR this size.

#### Additional context

Design based more-or-less off this set. 
image
In addition, here are the sources I consulted for the creation process.
https://detyre.tripod.com/Articles/lamellar1.html
https://www.researchgate.net/publication/286173075_The_Evolution_of_Splint_Armour_in_Georgia_and_Byzantium_Lamellar_and_Scale_Armour_in_the_10th-12th_Centuries/fulltext/5666a53108ae418a786f4e64/The-Evolution-of-Splint-Armour-in-Georgia-and-Byzantium-Lamellar-and-Scale-Armour-in-the-10th-12th-Centuries.pdf
This last one isn't historical, but serves as a good example for how you can easily DIY a set.
https://www.instructables.com/Metal-Armor-on-a-Budget-or-Lowes-Lamellar/
(I was very close to calling the low steel lamellar "Lowe's steel lamellar" believe me)
Also big thanks to Drew for helping advise this PR, again.

Balance-wise, the lamellar is actually super strong for its ease of crafting. It's somewhere between chain and light plate in terms of defense to encumbrance, with its main downside being a lack of sturdy and relatively little coverage (no lower arm/shin coverage, no helmet, feet, or hand coverage either), but is faster and easier to craft than chainmail and WAY faster than plate. If you're using a hand drill it'd probably take about 3-4 days to craft, but with a cordless drill/drill press it's easily makeable in under a day.

Also, the PR changed the name of the original "lamellar armor" to "leather lamellar armor" to make it more obvious it's, well, leather.

### DIFF
--- a/data/json/recipes/nested.json
+++ b/data/json/recipes/nested.json
@@ -1754,7 +1754,14 @@
     "name": "steel lamellar cuirasses",
     "description": "Recipes related to constructing steel lamellar cuirasses in various shapes and sizes.",
     "skill_used": "tailor",
-    "nested_category_data": [ "armor_lamellar_lc", "armor_lamellar_ch", "xs_armor_lamellar_lc", "xs_armor_lamellar_ch", "xl_armor_lamellar_lc", "xl_armor_lamellar_ch" ],
+    "nested_category_data": [
+      "armor_lamellar_lc",
+      "armor_lamellar_ch",
+      "xs_armor_lamellar_lc",
+      "xs_armor_lamellar_ch",
+      "xl_armor_lamellar_lc",
+      "xl_armor_lamellar_ch"
+    ],
     "difficulty": 4
   },
   {
@@ -1766,7 +1773,14 @@
     "name": "steel lamellar repair",
     "description": "Recipes related to repairing steel lamellar armor in various shapes and sizes.",
     "skill_used": "tailor",
-    "nested_category_data": [ "armor_lamellar_lc_repair", "armor_lamellar_ch_repair", "xs_armor_lamellar_lc_repair", "xs_armor_lamellar_ch_repair", "xl_armor_lamellar_lc_repair", "xl_armor_lamellar_ch_repair" ],
+    "nested_category_data": [
+      "armor_lamellar_lc_repair",
+      "armor_lamellar_ch_repair",
+      "xs_armor_lamellar_lc_repair",
+      "xs_armor_lamellar_ch_repair",
+      "xl_armor_lamellar_lc_repair",
+      "xl_armor_lamellar_ch_repair"
+    ],
     "difficulty": 4
   },
   {


Aerin-of-the-Toast:
> So  if you were gonna make case hardened lamellar, it would need to have the holes drilled first, then each individual scale would be case hardened before being tied together? 

More or less yeah, there's a really high batch time save though since you're cooking a larger pile all at once. 

I suppose theoretically you could case harden before drilling holes or even when it's still metal sheets but it seemed redundant to have so many options and would probably just make cutting and drilling harder.

MNG-cataclysm (approved):

PR #62283, opened 2022-11-18
Update JSON_FLAGS doc with Mapgen entry

solerante:
#### Summary
None

#### Purpose of change
Update JSON_FLAGS doc with Mapgen flags so that all flags can be found in same place.

#### Describe the solution
Added flags found in https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/MAPGEN.md. `ERASE_ALL_BEFORE_PLACING_TERRAIN`, `ALLOW_TERRAIN_UNDER_OTHER_DATA`, `NO_UNDERLYING_ROTATE`, `AVOID_CREATURES`, and hotlink to `remove_all` in mapgen doc https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/MAPGEN.md#remove-everything-with-remove_all

#### Testing
Preview


solerante:
What category can I give this pr which modifies a github doc?

Night-Pryanik:
The closest one is Infrastructure, but actually we don't have a fitting category for such things, so you can use None.

PR #62287, opened 2022-11-18
New zombie line: tapeworm host

Karol1223:
#### Summary
Content "Adds a new zombie evolutionary line: tapeworm zombies"

#### Purpose of change
I really wanted mutant tapeworms and this felt like a good way to implement them.

#### Describe the solution
Adds three new zombies: the Ruptured Host, the Spindly Carrier and the Twisted Hive as well as a Mutant Tapeworm and an Anaconda Tapeworm (the Mega evolutionary stage). 

The Ruptured Host is simply a ranged attacker, utilising its Tapeworm friends to hit and weaken targets at range. 

The Spindly Carrier is obviously stronger but also has a possibility of pulling the player to themselves like a grappler.

Then the Twisted Hive is the hulk equivalent. It is stronger, slightly faster. Hits from further away and for more and can pull the player from further away as well using a spell. On top of that, the Twisted Hive will periodically spawn mutant tapeworms the same way a Centipede Mother spawns her babies. 

Additionally, upon the death of any of these zombies a Tapeworm will spawn, or, in the case of the Twisted Hive, the Anaconda Tapeworm. None of these zombies revive as themselves, with the Ruptured Host reviving as a Decayed Zombie, the Spindly Carrier as a Grappler Zombie and the Twisted Hive not reviving at all (with the impression that without the worms the body is automatically too mangled to function and revive).

Plenty, primarily when it comes to balancing, but I think the equilibrium is quite nice as is.

#### Testing
DOESN'T WORK YET. IT IS POSTED AS A DRAFT SO OTHERS CAN BETTER SEE MY CODE CHANGES. However, the balancing seems pretty on point. A midgame test character can melee the Twisted Hive and win on a coin flip, taking massive damage in the process. However, at that point the Anaconda Tapeworm would spawn and most likely finish them off. I'm not going to pretend like that's not intentional - this line is ranged and you should either outrange them or be prepared to fight the worms once you make them homeless.

#### Additional context
Still gotta make them actually spawn. And there's a bunch of more spell/death effect weirdness to iron out. Oh and the Hive's description is a WIP.


data/json/monster_special_attacks/monster_deaths.json
```diff
@@ -690,5 +690,44 @@
...
+    "valid_targets": [ "ground", "self" ],
+    "min_damage": 1,
+    "max_damage": 1,
+    "flags": [ "SILENT", "HOSTILE_SUMMON", "PERMANENT", "SPAWN_WITH_DEATH_DROPS", "NO_EXPLOSION_SFX" ],
+    "shape": "blast",
+    "effect": "summon",
+    "effect_str": "mon_tapeworm_mega",
+    "min_aoe": 1,
+    "max_aoe": 1,
+    "extra_effects": [ { "id": "emit_blood_explosion", "hit_self": false, "max_level": 3 } ]
```
GuardianDll, line 723:
```suggestion
    "extra_effects": [ { "id": "emit_blood_explosion", "hit_self": true, "max_level": 3 } ]
```

data/json/monsters/zed_worm.json
```diff
@@ -0,0 +1,243 @@
...
+      [ "scratch", 20 ],
+      {
+        "id": "tapeworm",
+        "damage_max_instance": [ { "damage_type": "stab", "amount": 14, "armor_penetration": 4 } ],
+        "range": 7
+      }
+    ],
+    "death_function": { "effect": { "id": "death_tapeworm", "hit_self": true } },
+    "death_drops": "mon_zombie_hulk_death_drops",
+    "zombify_into": "mon_zombie_grappler",
```
NetSysFire, line 96: Maybe add a monster group so 90% of the time it revives as a grappler zombie but the remaining 10% the stars align so a tapeworm can grow anew.

Karol1223: The tapeworms that were to grow would use the corpse as the nutrients to grow, though, so the corpse would just be left in too sorry of a state to revive at that point. The idea is kind of cool I'm just not convinced how plausible that'd be.

data/json/monsters/zed_worm.json
```diff
@@ -0,0 +1,243 @@
...
+      "FILTHY",
+      "RANGED_ATTACKER"
+    ]
+  },
+  {
+    "id": "mon_zombie_worm_hulk",
+    "type": "MONSTER",
+    "name": { "str": "twisted hive" },
+    "description": "An unnaturally large emaciated quadrupedal creature that you can only assume has once been human.  The actual host body is almost skeletal, visibly starved of any excess biomass to feed the evergrowing mass or mutant worms inside.  The forelimbs are visibly elongated and end in long bony claws, while the hindlimbs have developed a digitigrade foot structure.  On its torso, both from the top and from the bottom, there are snaking dozens if not hundreds of worms of various sizes and on the central part of the back resides the largest one, coiled several times around the creature.",
+    "default_faction": "zombie",
```
TheShadowFerret, line 119: Why would a "zombie" that's mostly the mutant parasites(which should be hostile to zombies) still be the same faction 🤔

>Since zombos now have a reason to hunt bugs dissolve the zombie / creepy crawly truce and watch swamps neighboring towns erupt into a multigenerational free-for-all. That should be fun. ~ #60605

PatrikLundell:
Tapeworms as biting external monsters able to sense the location of targets doesn't make any sense at all in my view.

data/json/monsters/zed_worm.json
Karol1223: The idea here was something akin to a Plague Vector, which is somewhere in the middle between a zombie and a mutant due to the fact it fed on enough 'tainted' flesh. I'm not actually sure where Plague Vectors fall on the faction chart now after #60605, though, and I have to check that.

Karol1223:
> Tapeworms as biting external monsters able to sense the location of targets doesn't make any sense at all in my view.

You are actually not wrong. This is something I have to look into a bit more before I get it out of draft. Outside the host the tapeworms are basically blind so them making the host able to see further (and attack from a long range) doesn't add up at all at the moment. I'll look into this a bit more when I have some more time (and once I make everything work in the first place).

PatrikLundell:
If you're looking for a really scary real life "monster" I'd rather consider something like the south american blowfly. This insect lays its eggs in open wounds and the larvae eat their way through the flesh of the host. I would guess they're the inspiration for the D&D rot grubs.
The larvae probably aren't able to bite through the skin once outside of a body, and are probably not particularly capable of sensing creatures when out of the body either, but it would be less of a stretch to "upgrade" these than something that's more or less a passive filter feeder.

Karol1223:
> south american blowfly

One of the problems here is that the species has to be present in New England, and I am not convinced the south American blowfly fits that criteria.

> The larvae probably aren't able to bite through the skin once outside of a body, and are probably not particularly capable of sensing creatures when out of the body either, but it would be less of a stretch to "upgrade" these than something that's more or less a passive filter feeder.

I just want to say I don't disagree with you at all on this one. My headcanon here was that the tapeworms have to mutate in such a way that they adapt to life outside the host's body. Because the hosts are simply not large enough for them to live inside anymore. That's why they'd evolve some form of sensory organs and changing the hooks present on some tapeworm species into natural weapons isn't that farfetched considering what we've seen with other mutants. What I believe here is that the giant tapeworms, and by extension the first zombie tiers here, should have a harder time seeing the target because even if the sensory organs are starting to appear they wouldn't be very good at their job yet. By the Anaconda Tapeworm, though, the animal is better suited for out-of-host life, as evidenced by the fact that while the normal Giant Tapeworms are almost blind and incredibly slow the Anaconda will see and pursue its prey.

Yes, it is kind of farfetched. I am not convinced its too farfetched, however.

PatrikLundell:
I still find the tapeworm to be an extremely unlikely evolution source.

The blowfly has been "exported" to Libya with infected cattle, but I don't know if they've managed to eradicate it after that (I think that happened well before the dictator was toppled and the area descended into chaos). Some rancher could well have decided to improve their cattle breeding stock as well as considering quarantine to be government oppression (that's a stretch as well, of course)...

data/json/monsters/zed_worm.json
```diff
@@ -0,0 +1,243 @@
...
+    "special_attacks": [
+      [ "RANGED_PULL", 75 ],
+      { "type": "bite", "cooldown": 5 },
+      {
+        "id": "tapeworm",
+        "damage_max_instance": [ { "damage_type": "stab", "amount": 23, "armor_penetration": 7 } ],
+        "range": 10
+      },
+      {
+        "type": "spell",
```
GuardianDll, line 156:
```suggestion
        "type": "spell",
        "id": "tapeworm_summon",
```

data/json/monsters/zed_worm.json
```diff
@@ -0,0 +1,243 @@
...
+        "spell_data": { "id": "tapeworm_summon" },
+        "monster_message": "A tapeworm falls off the hive's body",
+        "cooldown": 20
+      },
+      { "type": "spell", "spell_data": { "id": "tapeworm_pull" }, "required_effects_any": [ "tapeworm_pull_aimed" ] },
```
GuardianDll, line 161:
```suggestion
      { "type": "spell",  "id": "tapeworm_pull", "spell_data": { "id": "tapeworm_pull" }, "required_effects_any": [ "tapeworm_pull_aimed" ] },
```

Zireael07:
Is this the blowfly subspecies also known as screwworm?

PatrikLundell:
Don't know. Could be. A translation into English of the name I know is "boring blowfly", referring to drilling, not tedious, of course.

Zireael07:
Some other parasite ideas: https://mind-blowingfacts.com/sinister-parasites/

Some of those do consume the host. Also see https://www.youtube.com/watch?v=klP_GyOyjn4

kevingranade:
I agree with the feedback that this needs to be not tapeworm based. 
I'm a lot less concerned with the "it needs to be a native species" concept compared to "this animal doesn't act like that".
The idea is you have some random carrion eating or parasitic larva attack a zombie, but it in turn gets mutated into a symbiote for the zombie.

Karol1223:
> I agree with the feedback that this needs to be not tapeworm based. I'm a lot less concerned with the "it needs to be a native species" concept compared to "this animal doesn't act like that". The idea is you have some random carrion eating or parasitic larva attack a zombie, but it in turn gets mutated into a symbiote for the zombie.

I still gotta think a lot of this through and odds are I'll be asking a lot of questions on discord in the following days/weeks. I rushed the design for this and hit a wall. I want to bring it back to life so that it's in a workable state but what the zombie will end up in the end is still in the air.

Ideally what I want is a mix of the two things you mentioned. I wanted a tapeworm that mutated into becoming a zombie symbiote but that can survive outside the host. The problem in that is I cannot have the zombie host be weaker if the worm is a symbiote, can I? That's where things complicate. I still gotta put a lot of thought into this.

Karol1223:
Yes, this looks stupid but bear with me. I am closing it manually so I can reopen it when I have time to properly finish it (hopefully second half of June), if stalebot does it *again* I'll have to ask for help from people with merge rights. *Again*.

PR #62288, opened 2022-11-18
fix: simpler reqs for "upgrade basecamp stove"

casswedson:
#### Summary
None

#### Purpose of change
my lad Rycon has been a little upset at the requirements for building a stove in his basecamp for some time now, there's a hard requirement for a cast-iron frying pan which is a hard to come across in certain scenarios, innawood for example

change the pot or copper pot or clay pot *and* iron frying pan to be *or* iron frying pan, that seems reasonable to me

first time editing a recipe I followed the docs but I didn't really test

#### Additional context
welp, he fixed it on his own, I was just a little late

### DIFF
--- a/data/json/recipes/basecamps/expansion/recipe_modular_canteen/version_2/recipe_modular_canteen_common.json
+++ b/data/json/recipes/basecamps/expansion/recipe_modular_canteen/version_2/recipe_modular_canteen_common.json
@@ -84,7 +84,7 @@
     ],
     "blueprint_excludes": [ { "id": "fbmk_2_fire" } ],
     "blueprint_resources": [ "fake_fireplace", "fake_fireplace", "pot" ],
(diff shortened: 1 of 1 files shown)


Maleclypse:
I do think a different solution would be to have primitive cooktops be a thing because that's basically what pans replace.  

https://www.appliancesconnection.com/blog/cooktop-history

"The design was exceedingly simple. A moderately flat-topped earthen dome was shaped over a fire cavity and food or cooking containers would be placed on top."

PatrikLundell:
I agree a specific type of frying pan is too restrictive, but it makes sense to have one tool for boiling and one for frying. Thus, in my view, the solution ought to still be to have two tools, but relax the requirements on what the frying tool can be.

It would also make sense to provide one or more alternatives to an actual stove (i.e. parallel upgrade recipes), such as e.g. a clay oven (but the tool requirements would probably be the same ones).

Issue #62001, opened 2022-10-29
Viking Traders don't seem to know their stock


AcidAntOnAMinefield:
### Describe the bug

Found the "vikings", checked them out, talked to their trader(?) / warehouse keeper / whatever it is they do. They only show me the stuff they have on them, despite having full shelves around them. 

Completed a couple quests for them, nothing changed other than the trader being more willing to sell me the ammo to the weapon they refuse to sell me anyway(?). 

Debugged in a different world, completed all quests via that way, and they still do not show anything other than their personal inventory. 

### Steps to reproduce

1. make a world
2. debug reveal map
3. visit the "vikings" 
4. try to trade with the "warehouse manager"
5. notice they only offer you access to whatever it is they're holding.
6. debug complete all the quests for them, check their trade menu again, notice how it doesn't change in size. 
7. wonder why they have so much stuff on display if it isn't available for sale. 

### Expected behavior

I would expect the shopkeep to have access to the stuff around them. 

### Versions and configuration

Windows
Tiles
7082ac9 - 2022-10-29-0658
No particular mods.

### Additional context

I had the scrap yard guy nearby so I decided to pay him a visit and he seemed fine(?). So not sure what the deal is with the dogwhistle people.

andrei8l:
This just needs more zones defined [here](https://github.com/CleverRaven/Cataclysm-DDA/blob/743cc4474a1696665e1bc33a3b7a7d5c1ddee14c/data/json/mapgen/valhalla_cult/mapgen.json#L196). The existing zones aren't correct either as they're placed on top of a wall.

MNG-cataclysm:
/confirmed

I'll start fixing that.

PR #62294, opened 2022-11-19
Added valhalla trophy armor

bean-b:
#### Summary
Content "Added valhalla trophy armor"

#### Purpose of change
To add new armor. I thought it would be cool for the valhallists to get some more specialized armor, and this armor seemed cool and within reason.

#### Describe the solution

The armor is a bit complicated but the base idea is that it is a suit of leather armor, basically the armor_larmor, but a bit thinner, and a small amount of fur has been sewn inbetween the gaps of the leather armor. It has protective chitin inserts, the same as our current armor_plarmor, but with chitin instead of budget steel. Ornamental bones adorn the armor, which add some minimal protection, but arent really designed to protect much. The armor features a heavy fur coat and upper body cloak and hood, and the hood can be raised or lowered.

The armor itself is pretty protective, but has fairly high encumbrance as it isnt the most functional suit of armor. It is also hard to craft, because you need to buy the blueprint from the valhallists, and the recipe itself is quite difficult. By the time it is craftable, there may be better options, but oh well (shrug), cool impracticable monster armor is still cool.

Not adding this item, different values.

#### Testing
Spawned the suit and its variants in game looked at their values. teleported to a valhlla cult, saw that the plans spawned in their store room. Spawned the blueprint, and read it, learning the recipe. Looked at the recipe, it seemed good.

#### Additional context
**Notes about other PR's / issues**
If and when my https://github.com/CleverRaven/Cataclysm-DDA/pull/62256 gets merged, ill adjust the recipe and armor to use hardened leather, and balance it as needed of course

Also, because of https://github.com/CleverRaven/Cataclysm-DDA/issues/62001 the blue prints are impossible to obtain without breaking into their store rooms, but seeing as this issue effects the entire trading stock of the entire viking faction, it is both out of the scope of this pr, and when fixed will allow this blueprint to be bought normally.

**Armor justification**
The fur upper is not unlike this, just with an integrated hood
image
The leather armor is basically just what we have now, just with that thing layer of fur filling up many of the gaps.
The chitin plates are attached the same way as plated leather armor (armor_plarmor) is right now
The fur lower is somehting like this, but a bit shorter, covering only the hips
image
The bone bits are simple bones sewn into the leather in spots where they will not get in the way much, and again arent designed for protection. 

Obviously this armor isnt the most historically realistic, but neither is a viking cult in the end of the world, nor giant ants with chitin carapaces that can be made into armor. While it isnt historically accurate of course, I think it is all something within the means of a skilled survivor to craft, and its fairly high encumbrance represents some of the downsides of this semi impracticable nature.

### DIFF
--- a/data/json/items/armor/suits_protection.json
+++ b/data/json/items/armor/suits_protection.json
@@ -3224,7 +3224,7 @@
     "repairs_like": "armor_larmor",
     "category": "armor",
     "name": { "str": "valhallist trophy armor" },
-    "description": "A fearsome, if somewhat impractical suit of ornamental armor, crafted with an amalgamation of monster bits slain by the wearer. A thin layer of fur covers the gaps in the leather armor, and the whole armor is adorned with protective chitin plating and ornamental bones. A fur skirt is affixed to the hips. The upper has an integrated thick fur coat and hood. Use it to lower the hood.",
+    "description": "A fearsome, if somewhat impractical suit of ornamental armor, crafted with an amalgamation of monster bits slain by the wearer.  A thin layer of fur covers the gaps in the leather armor, and the whole armor is adorned with protective chitin plating and ornamental bones.  A fur skirt is affixed to the hips.  The upper has an integrated thick fur coat and hood.  Use it to lower the hood.",
     "weight": "6750 g",
     "volume": "6500 ml",
     "price": 25000,
@@ -3354,7 +3354,7 @@
     "id": "armor_valhalla_trophy_casual",
     "type": "ARMOR",
     "name": { "str": "valhallist trophy armor (casual)", "str_pl": "valhallist trophy armors (casual)" },
-    "description": "A fearsome, if somewhat impractical suit of ornamental armor, crafted with an amalgamation of monster bits slain by the wearer. A thin layer of fur covers the gaps in the leather armor, and the whole armor is adorned with protective chitin plating and ornamental bones. A fur skirt is affixed to the hips. The upper has an integrated thick fur coat and hood. Use it to raise the hood.",
+    "description": "A fearsome, if somewhat impractical suit of ornamental armor, crafted with an amalgamation of monster bits slain by the wearer.  A thin layer of fur covers the gaps in the leather armor, and the whole armor is adorned with protective chitin plating and ornamental bones.  A fur skirt is affixed to the hips.  The upper has an integrated thick fur coat and hood.  Use it to raise the hood.",
     "copy-from": "armor_valhalla_trophy",
     "armor": [
       {


Venera3:
How exactly did they learn to use chitin from day 1? A whole viking LARP cult in full medieval gear springing up two weeks after the end of the world is shaky enough already, but "oh no they also worked out chitincrafting aren't they cool" is more of a Fallout late postapocalypse thing than what we're going for.

MNG-cataclysm:
I agree with Venera3, this sort of change is too Fallout-y for what the Valhallists are about. Besides, they focus more on practicability than fashion. They'd stick with metal/leather armor; the ingredients of which are more widely available, less expensive, and do not require addition research to effectively produce.

PatrikLundell:
Echoing other critics in that this seems more like something for fantasy barbarians than neo-vikings (unless the "vikings" have drinking horns on their helmets, in which case all attempts at realism are shot anyway).

bean-b:
> How exactly did they learn to use chitin from day 1? A whole viking LARP cult in full medieval gear springing up two weeks after the end of the world is shaky enough already, but "oh no they also worked out chitincrafting aren't they cool" is more of a Fallout late postapocalypse thing than what we're going for.
>
I do agree from day one is a bit fast, but I'm not sure that theres currently a way to have merchants only sell an item after X days. My logic was that the surivvor is able to figure out chitin working in a few days and even a survivor with 0 tailoring can make chitin body armor in a week or less given a supply of chitin(Although maybe this in and of itself is unrealistic, but this is most of how crafting in this game works). And I would imagine that at least one of the vikings would have some prior knowledge of tailoring/armor esp given the LARP background. Plus, the chitin working in this piece isnt meant to be like a "suit" of armor like the chitin body armor, but more leather armor with sewnin chitin plates. But even given this, I can understand how this may not feel believable still in the game to the player. 


> I agree with Venera3, this sort of change does not fit well with the more traditional and historic designs of the Valhallists. They focus more on practicability than fashion, and would stick with metal/leather armor; the ingredients of which are more widely available, less expensive, and do not require addition research to effectively produce and maintain.
>
 I read the valhalla_cult tag in json and summed they would be a more cult-y then i guess they are in actually. I do agree the armor as a whole is not very practical, but the idea in and of itself of having chitin plates in leather armor seems somewhat practicle if you live in a world where you are consistently fighting creatures that can be butchered for protective chitin chunks. 
> However, I can see this working well for a more tribalistic or survivalist faction, possibly as a show of prowess with the exotic materials included in the design.
> 
That would be really cool, although im not sure if i have the time to implement all of that

MNG-cataclysm:
> Echoing other critics in that this seems more like something for fantasy barbarians than neo-vikings (unless the "vikings" have drinking horns on their helmets, in which case all attempts at realism are shot anyway).

Fortunately, the Vikings do not have drinking horns on their helmets. 
That might be very good party attire or practical day-to-day wear in the Cataclysm, so you can hold one or more horns full of beverages without having to use your hands, and always having 1+ cup(s) available.

PR #62297, opened 2022-11-19
content: mainlined Whaley's Hair And Tattoo mod

sonphantrung:
#### Summary
Content "mainlined Whaley's Hair And Tattoo mod"

#### Purpose of change
moar cosmetic stuff

#### Describe the solution
Mainlined [the mod](https://github.com/neonspectra/whaleys-hair-and-tattoo), originally made by `u/whaleywhaley`, updated version maintained by @neonspectra

NONE

#### Testing
Loaded and it works

#### Additional context
image


data/json/mutations/mutation_appearance.json
```diff
@@ -467,6 +467,894 @@
...
+    "starting_trait": true,
+    "valid": false,
+    "purifiable": false,
+    "player_display": false,
+    "types": [ "makeup_face" ]
+  },
+  {
+    "type": "mutation",
+    "id": "FUMANCHU_MOUSTACHE",
+    "name": "Facial Hair: fumanchu Moustache",
```
DSeyka, line 881:
```suggestion
    "name": "Facial Hair: Fu Manchu Moustache",
```

Fris0uman:
Add the people that worked on the mod as co-authors, make a commit and do
```
[describe commit]


Co-authored-by: name <username@users.noreply.github.com>
```
In the commit mesage

neonspectra:
Would love to see this mainlined, but the reason I never did was because the original mod was never licensed, and Whaley does not use GitHub. Quite frankly I've also been just a little lazy about this 😅.

I had a long discussion about this with @Fris0uman in [this reddit thread](https://www.reddit.com/r/cataclysmdda/comments/q5hnfn/unofficial_patch_whaleys_hair_tattoo_for_0f/) for the release of my updated version.

I can try DMing Whaley on Reddit to see if he'd sign off on this. That being said, I see this PR only has json... the actual art itself would need to be added onto the tilesets for msxotto+ derived tiles. I'm sure Ultica could make their own versions but yeah. I think I'd agree that at this point so many people have been using these styles for so long that it probably makes sense to have the style names themselves mainlined.

Anyway, yeah, I'll message Whaley and see what we can do on that front.

Fris0uman:
So the lack of license is a problem. I don't think we can't integrate it like that without at least the original author coming in and agreeing to add this content to dda under cc by sa

neonspectra:
Agreed. I reached out to Whaley on Reddit to see if we can get this to happen-- I told them to potentially reach out to you as well since you're the tileset maintainer and would be a better person to talk to with regard to signing off on this. So keep an eye out in the event of stray Reddit messages.

For now I think this PR stays in limbo for if/until we hear anything from Whaley. That being said, in the meantime, I think this PR is a good catalyst for me to go through and clean up the metadata and IDs so that they match the style of mainline for consistency's sake. Already did a pass on the meta descriptions, but the ID names will be fun to ensure I don't mismatch and break any while changing them.

To be honest, I probably should have done that forever ago when I did the quick json edit to get it working on 0.F, but you get how life is :)

Fris0uman:
So I recieved this via reddit, which puts us in a better spot already
image
 But I'm not sure if a screen shot like this is enough or not

neonspectra:
I would argue that having Whaley post on GH actually is probably *less* good, since there's no way to verify that someone posting as Whaley on GH is the same person as on reddit. 

Now that I think about it more, I think what would probably make the most sense and would put it in the green would be to have Whaley edit the original reddit post to include CC-BY-SA language.

Probably easier than having him make a GH account if he doesn't want to use it, too.

Fris0uman:
Sure, that would probably do

kevingranade:
For reference here and later, "do whatever you like" is not a license and is not sufficient for us to accept content from a third party. 
A clear statement in public that they release the content in question under our license (by name) is what we need.

PR #62299, opened 2022-11-19
Mark overmap location if player drops favorited item

### LINKED ISSUE #59917: Remembering location of an item on the map

I often drop my backpack to fight with less encumbrance, and then struggle to find it later because I forgot where I dropped it.

### Solution you would like.

Maybe something like "L - Remember location of item on map" on this menu, similar to the option for vehicles?
image
When this is selected for an item, an autonote would automatically be created at its location when it gets dropped, and removed when picked up.

Simply having better memory

### Additional context

_No response_

Night-Pryanik:
#### Summary
Interface "Mark overmap location if player drops favorited item"

#### Purpose of change
* Closes #59917.

#### Describe the solution
- If player drops a favorited item, create a new auto-note on the location player is standing on. If there's already a note on that location, instead append the note.
- Added a game option (under Auto-notes section of General options), defaulted to false, for those who don't want this.

Create a more complex system like for tracked vehicles.

#### Testing
Favorited my Barrett, dropped it one the ground. Checked global map. Dropped my favorited blood draw kit, checked global map once more.

#### Additional context
изображение

изображение

### DIFF
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -48,6 +48,8 @@
 #include "mtype.h"
 #include "npc.h"
 #include "optional.h"
+#include "options.h"
+#include "overmapbuffer.h"
 #include "pickup.h"
 #include "player_activity.h"
 #include "point.h"
@@ -378,6 +380,15 @@ void drop_on_map( Character &you, item_drop_reason reason, const std::list<item>
                     it_name, ter_name
                 );
                 break;
+        }
+
+        if( get_option<bool>( "AUTO_NOTES_DROPPED_FAVORITES" ) && it.is_favorite ) {
+            const tripoint_abs_omt your_pos = you.global_omt_location();
+            if( !overmap_buffer.has_note( your_pos ) ) {
+                overmap_buffer.add_note( your_pos, it.display_name() );
+            } else {
+                overmap_buffer.add_note( your_pos, overmap_buffer.note( your_pos ) + "; " + it.display_name() );
+            }
         }
     } else {
         switch( reason ) {


Qrox:
For people who drop their favorited items regularly this will clutter their overmap with outdates notes unless the note is removed after picking up the items. I'm not sure if that's easyly achieveable considering that there may be multiple favorited items with the same name and that people can mess around with the notes before picking the items up. It would probably be better to store the items locations of the dropped items in a separate list and display the items on the overmap on the fly.

Night-Pryanik:
> It would probably be better to store the items locations of the dropped items in a separate list and display the items on the overmap on the fly.

Maybe, but it's a much more complex stuff to implement, and I didn't want to mess with something like that, so I went the easy way.

src/activity_item_handling.cpp
```diff
@@ -379,6 +381,15 @@ void drop_on_map( Character &you, item_drop_reason reason, const std::list<item>
                 );
                 break;
         }
+
+        if( get_option<bool>( "AUTO_NOTES_DROPPED_FAVORITES" ) && it.is_favorite ) {
```
ZhilkinSerg, line 385: Is there a reason it was not added outside of `if-else` block?

Night-Pryanik: I needed `it` from this if-else block.

ZhilkinSerg: I see. Maybe you could gather list of all favorite items in the code block below and create overmap note for all dropped favorites after that?

```
    for( const item &it : items ) {
        here.add_item_or_charges( where, it );
        item( it ).handle_pickup_ownership( you );
    }
```

Night-Pryanik: I'll try that, when I get to my PC (tomorrow).

PR #62300, opened 2022-11-19
fixed the golf club not having any melee techniques

Daved27hundred:
#### Summary
Bugfixes "Golf club now has a medium blocking ability"

#### Purpose of change

The golf club isn't capable of any techniques at all, even a wooden plank has a blocking ability. This change was made to put it in line with other early game melee "weapons."

#### Describe the solution

added the medium blocking ability `WBLOCK_1` to the golf club's json. I was considering adding a sweep or stun attack but my goal was to fix a bug not make it a powerful weapon.

### DIFF
--- a/data/json/items/melee/bludgeons.json
+++ b/data/json/items/melee/bludgeons.json
@@ -620,6 +620,7 @@
     "price": 13000,
     "price_postapoc": 50,
     "material": [ "aluminum" ],
+    "techniques": [ "WBLOCK_1" ],
     "weight": "300 g",
     "volume": "1 L",
     "longest_side": "110 cm",


kevingranade:
Golf clubs are quite unwieldy and fragile,  unless this is paired with a (high) chance of breaking when blocking, it doesn't seem representative of their usability.

Daved27hundred:
can do, how would I do that?

MNG-cataclysm (approved):

oosyrag:
There are sturdier things like guns that don't have blocking either. A golf club can definitely snap under the weight of a body being applied to it the wrong way.

Since the sturdiness of weapons for defense isn't implemented yet (although such a feature probably could utilize the fragile_melee tag)  it seems reasonable to leave the blocking skill off of items not particularly suitable for blocking. Otherwise an argument could be made to add it to literally any item that could be wielded past a certain size.

oosyrag:
Damage on  block would be a c addition, but I imagine it wouldn't be too much code. The code for blocking exists, and the code for applying damage to items exists.

There might be significant additional considerations that could increase the code required though: what are the chances and factors involved? Various tags, weapon durability, enemy damage type and amount?

I think it's a fair step towards verisimilitude so it might not be so bad to get something in place and iterate on it. Begs the question if the blocking tag should even be obsoleted, I always thought it was strange for weapons to bestow "skills". I think it would be a natural instinct for any human to attempt to block an attack with whatever they're holding in their hands, be it a teapot or wooden plank or their arms if nothing is available.

Daved27hundred:
> I think it would be a natural instinct for any human to attempt to block an attack with whatever they're holding in their hands

yep, that was my thought process with this PR - while It'd damage the club, the player would probably still try to block something with it.

I-am-Erk:
I'm ok with this for now, it would indeed be nice to ensure fragile melee interacts with blocking, or have a separate fragile_blocker trait.

PR #62306, opened 2022-11-20
acid blood

flybye3o6:
#### Summary
Content "adds acid blood"

#### Purpose of change
adds acid blood as a blood type for harvest function

#61516 

#### Additional context

Screenshot 2022-11-20 220412
Screenshot 2022-11-20 220418
Screenshot 2022-11-20 220422

Screenshot 2022-11-20 220432
Screenshot 2022-11-20 220438
Screenshot 2022-11-20 220955
Screenshot 2022-11-20 220959
Screenshot 2022-11-20 221001
Screenshot 2022-11-20 220453


Night-Pryanik:
https://github.com/CleverRaven/Cataclysm-DDA/blob/f6e4f2feed5ac31e40378553de800e606debc186/src/iuse.cpp#L4799-L4800

When drawing blood from acidblood monsters (or acidblood player), the result is the sulphuric acid. 

In any case, I think blood drawing function needs to be updated too to reflect your changes.

PatrikLundell:
There's a problem with butchering/dissecting acid containing corpses if they generate acid blood pools (rather than a small acid blood contaminant, like ordinary blood does), namely that you have to endure standing in acid in order to acquire the material required to allow you to stand in acid without taking damage unless you can get hold of the protection elsewhere (which you can't in Innawood).

flybye3o6:
> 
i did consider that for Innawood some other means mite need to be made for that for protection, it is intentional that this is meant to be hazardous.
the acid only drops apron death(because of death function which is unchanged) and at the end of a butcher which the size of the spread seems to be random, bleeding the corpse doesn't drop blood on the ground same with skinning and dismembering, quick butcher and field dress sometimes dose sometimes doesn't, seems the longer the butcher task the more blood drops to the ground but again only once completed, it also only lasts a few minutes sometimes only seconds

rn if you bleed a corpse it doesn't change how much blood it drops after a full butcher or dissect, if a corpse was to truly be drained of blood when you bleed it that would fix the issue

flybye3o6:
test failing is because is this issue #62892

flybye3o6:
> This isn't ready until the linked issue is fixed

i can do the quick n dirty fix for this to pass, the larger issue is above my ability to fix and it seems others mite already be working on it
iv already tested it just need to update it and rerun some tests then all add it to this PR

flybye3o6:
i reverted that "fix" because removing charges deletes items from saves, and adds an item that will just be removed later
setting `"reversible"` to `false` for acidbomb is a much more elegant fix that can be easily be undone later when the issue is fixed proper

PR #62331, opened 2022-11-22
Electric spinwheel recipe options for materials (e.g. thread, sheet)

Reaper-Of-Souls:
#### Summary
Content "Electric Spinwheel recipes for materials (e.g. thread, sheet)"

#### Purpose of change

Allow material recipes (materials.json) that can be crafted with a spinning wheel (spinwheelitem) to also be crafted with an electric spinwheel (electric_spinwheel), much like the recipe options available for cords and ropes (cords_and_ropes.json).

#### Describe the solution

Every recipe in materials.json with a spinwheelitem recipe gets a corresponding electric_spinwheel recipe with the following changes (values derived based on cords_and_ropes.json):

LIGHT_EXERCISE - > NO_EXERCISE
spinwheelitem -> electric_spinwheel
charges used = (2 * minutes) + 1 (if no preexisting value in cords_and_ropes.json for the duration)

I did not add a corresponding recipe for "rags" as that appears to be in the process of obsoletion and so shouldn't need additional recipes.

#### Testing

Copied changes into my local experimental and verified they load the recipes and that thread_canvas could be crafted with electric_spinwheel (the original impetus behind this change).


Zireael07:
Can't this be done with the qualities system, instead of duplicating the recipes, or is the charges part a blocker??

Reaper-Of-Souls:
It looks like there are two options for recipe combination:

- A new quality could be added for spinwheelitem and electric_spinwheell, with electric_spinwheel using `charged_qualities` and `charges_per_use` to handle the electrical use.
  - But since charged_qualities works per use, the duration of the recipe would have no impact on charge cost, which means either long recipes cost very little, or short, small recipes cost a huge amount of charges in bulk.
- `"tools": [ [ [ "spinwheelitem", -1 ], [ "electric_spinwheel", 1 ] ] ],`
  - This would solve the recipe charge cost vs time issue from using charged_qualities, but... 

Both cases end up with the recipe always having the same `activity_level`.  [cords_and_ropes.json](https://github.com/CleverRaven/Cataclysm-DDA/blob/5b9a85a58d5e90b195d3d80cd1d6f5ea4bd5e73b/data/json/recipes/other/cords_and_ropes.json) clearly has using the electric_spinwheel set as `activity_level: "NO_EXERCISE"` and spinwheelitem as `activity_level: "LIGHT_EXERCISE"`

So, either we keep the recipe options separate to respect the preexisting LIGHT/NO difference, or we also merge all the other existing electric_spinwheel recipes into their related spinwheelitem recipe and treat electric_spinwheel as LIGHT_EXERCISE going forward.

Maleclypse (approved):

Issue #61405, opened 2022-10-02
Milkshakes and other milk-related crafts cause allergic reactions for the lactose intolerant when made from lactose-free ingredients


d15ddd:
### Describe the bug

![image](https://user-images.githubusercontent.com/98374042/193438239-6f0a7bf8-a2bf-4256-b86d-3ac15ef2f9dd.png)
![image](https://user-images.githubusercontent.com/98374042/193438275-48f9c921-75a4-414a-af86-a35b4dad0482.png)
Notice the soy milk used

### Steps to reproduce

Self-explanatory

### Expected behavior

Milk products should be lactose-free when made from lactose-free ingredients

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.22000.318 (21H2)
- Game Version: 07a9641 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

talltitanicc:
I can confirm that this bug is indeed happening as well on a fresh copy of 349262c. Provided are screenshots of sweetened fortified milk and coffee substitute with milk with the flags ALLERGEN_MILK. Both of which cause allergic reactions.
The issue seems that the recipes point to a singular item and must be separated to instead point to a lactose-free version to fix the issue.
![wrong2](https://user-images.githubusercontent.com/66237588/194770529-fb405b7a-f510-4b73-96d9-65cf6e6fc137.PNG)
![Wrong](https://user-images.githubusercontent.com/66237588/194770526-4d813f9e-68c6-4600-b8e3-3e6fade326d5.PNG)

aaronstevenson408:
the first item is found in json\items\comestibles\drink.json
![Sweet milk](https://user-images.githubusercontent.com/64069763/202925376-272522f7-10b5-4f1c-86e7-77fd50aad9a1.PNG)
and json\recipes\food\drinks.json
![Sweet milk recipes](https://user-images.githubusercontent.com/64069763/202925370-7b451780-4461-4512-9b1a-5cd438e54de1.PNG)
the second item also is found in json\items\comestibles\drink.json
![coffee Substitute](https://user-images.githubusercontent.com/64069763/202924849-aa511af1-7cbb-4c72-bce3-868309745c56.PNG)
[...]

PR #62342, opened 2022-11-23
allow allergens to be inherited

KorGgenT:
#### Summary
Infrastructure "change allergens to be more json-based, using material ids instead of flags on items."

#### Purpose of change
the main purpose is actually to allow allergens to be inherited, and the ability to use regular recipes to just use materials that you are not allergic to to create foods you can eat, without making double items and double recipes.

#### Describe the solution
the first part is changing the function that looks at your allergies, making it check materials. then, i had to backtrack to materials and add a morale_type id to materials. then, i needed to add enchantments (since i used the enchantment system for allergies) to the traits that matched the tuples that were originally coded.

#### Testing
spawn some things that you're allergic to, and make sure you're allergic.
then, spawn things you're not allergic to, that can be crafted into an item that nominally you would be allergic to if you spawned it in, and see if you're not allergic to it.

#### Additional context
there's probably some extra bits in here for something like wool allergies, but i didn't touch any of that.
see also #61405


KorGgenT:
image
i'm running into this issue again. it seems like every time i come back after a while i get this, and i always forget how i fix it.

anoobindisguise:
My characters sometimes had the antifruit and antiveggy traits but would make cookies out of rice and dehydrated fruit, or junkfood intolerant chars laundering sugar by making sweet water... so many loopholes will be fixed by this, love to see it!

TiberianErin:
image
seems like this PR makes multiple of the item traits related to allergens obsolete, and therefore the errors are generated because of this

TiberianErin:
image
This error is complaining because magic enchantments do not know what allergies are anymore.

PR #62358, opened 2022-11-24
replace BIOLUM0_active with conditional enchantment

### LINKED ISSUE #53324: Repeatedly acquiring Reflex Photophore mutation
### Describe the bug

Using concentrated mana serum can cause you to gain the Relfex Photophore mutation when you already have it. The issue appears to be related to the fact that the Relfex Photophore can be in the (on) state, that is to say the check for existing mutations looks for "Reflex Photophore", but not "Reflex Photophore (on)".

This appears to be a regression, see https://github.com/CleverRaven/Cataclysm-DDA/issues/46217 and https://github.com/CleverRaven/Cataclysm-DDA/pull/46507

### Steps to reproduce

1. Acquire the Reflex Photophore mutation.
2. Take concentrated mana serum while Reflex Photophore is (on).
3. Lose and regain the mutation.

### Expected behavior

The Reflex Photophore mutation should not be available in the possible mutation pool if it's already been acquired.

### Screenshots

photophore

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: 9600194 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm],
]

### Additional context

_No response_

GuardianDll:
#### Summary
Bugfixes "You can't get multiple Reflex Photophore mutations anymore"
#### Purpose of change
Fix #53324
#### Describe the solution
Replaced the BIOLUM0 mutation transforming into conditional enchantment
Obsoleted the BIOLUM0_active
#### Testing
The game works fine, the effect looks nice, except the fact it reads only the first condition (currently only hunger change your lumination)
#### Additional context
*wow, you are really hungry*
image

### DIFF
--- a/data/json/mutations/migration.json
+++ b/data/json/mutations/migration.json
@@ -6,6 +6,11 @@
   },
   {
     "type": "TRAIT_MIGRATION",
+    "id": "BIOLUM0_active",
+    "trait": "BIOLUM0"
+  },
+  {
+    "type": "TRAIT_MIGRATION",
     "id": "eye_blue",
     "trait": "eye_color",
     "variant": "blue"
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -274,49 +274,19 @@
     "name": { "str": "Reflex Photophore" },
     "points": -1,
     "description": "A photophore has grown from your head.  You can't consciously control it, and it might start to shine in response to your emotions or your physiological state.",
-    "cost": 9,
-    "time": "225 h",
-    "kcal": true,
     "encumbrance_covered": [ [ "head", 5 ] ],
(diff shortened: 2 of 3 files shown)


Fris0uman:
The loss of the morale effect is not great though. Also should the trigger thing from mutation be scrapped eventually if enchantments can replace it?

GuardianDll:
> The loss of the morale

Its not really lost, as the effect is still here, someone just need to fix enchantment issue

> should the trigger thing from mutation be scrapped

You mean in obsoleted mutation? i can, but i think it has 0 difference

Fris0uman:
No I meant the cpp support for it, but I haven't looked at what uses it, it's probably a dumb idea.

Issue #58795, opened 2022-06-28
Game lags during some activities when the "force redraw" option is enabled


mqrause:
### Describe the bug

I experienced massive lag when autoforaging my crop field. Taking a step and foraging two crop tiles took roughly one second.

### Steps to reproduce

1. Start a new game in a default world.
2. Advance time by two seasons.
3. Teleport to some corn field.
4. Turn on autoforaging crops.
5. Move in a straight line through the field both with and without "force redraw" turned on.
6. Notice a difference (probably only on weaker systems).

Instead of autoforaging, differences are also noticeable when sorting large amounts of loot with zones, or when picking up, dropping or hauling large quantities of items. Other activities might also be affected. It might be that movement of characters and items simply makes the lag more obvious.

### Expected behavior

The game runs smooth and only does as much redrawing as necessary.

### Screenshots

This is the profiling result:
![force_redraw](https://user-images.githubusercontent.com/38702195/176251169-57bab30a-9782-4afa-ac85-1c932ab2edaa.png)

Most notably this is caused by the "force redraw" option being enabled, which is the default setting. The code in question is in `do_turn`:
![do_turn](https://user-images.githubusercontent.com/38702195/176251163-774b2228-5571-4ed4-b301-ac0eb222db0f.png)

After turning the option off, movement was pretty much like moving without autoforage, for comparison the profiling result doing the same thing:
![no_force](https://user-images.githubusercontent.com/38702195/176251166-87c11827-8fba-4964-8b59-84fdc5df7620.png)

### Versions and configuration

Tested and profiled on ba775d74b44e51235368b892b7a7f99a133ee8dd

PR #62362, opened 2022-11-24
Fix Disconsolate Merc crash

### LINKED ISSUE #60938: Disconsolate Merc crashes game.
### Describe the bug

The game Crashes when purchasing the 5-Point-Anchor from the disconsolate Merc.

### Steps to reproduce

1. Advance the game far enough along that the terminal bar will populate near the hub.
2. Speak to the Disconsolate Merc and accept the offer to purchase the 5-Point-Anchor.
3. Game will crash the next time you pass a turn in game.

### Expected behavior

I expect the game to not crash when speaking to NPC's.

### Screenshots

_No response_

OS: Windows 10
build type: windows-tiles-sounds-x64-msvc
build number: 2022-09-05-2002
Ingame Language: English
Mods Loaded: dda, xedra_evolved, No_Rail_Stations, no_npc_food, no_fungal_growth, StatsThroughSkills, speedydex.

### Additional context

I can't fit the save here but I still have it and am able to send it over discord. I'm Catnip#6227.

mqrause:
#### Summary
Bugfixes "Fix Disconsolate Merc crash"

#### Purpose of change

Partially fixes #60938

#### Describe the solution

The crash is caused by outdated info on visible monsters when the forced redraw option is enabled and the compass widget tries to access the already removed npc. This adds a call to update the monster info before redrawing.

Figure out which order all the stuff happens in and find a way to shuffle things around without adding another call. That's probably not possible, though.
Or remove the force redraw option because it also causes other isses (#58795). But that obviously has more implications.

#### Testing

Load the save and walk north. To cause a crash both the "force redraw" option and the compass widget have to be enabled. With this fix there won't be a crash even with them enabled.

#### Additional context

http://files.catbox.moe/oevx9t.zip
Thanks @RenechCDDA for providing the save.

### DIFF
--- a/src/do_turn.cpp
+++ b/src/do_turn.cpp
@@ -809,6 +809,7 @@ bool do_turn()
     g->mon_info_update();
     u.process_turn();
     if( u.moves < 0 && get_option<bool>( "FORCE_REDRAW" ) ) {
+        g->mon_info_update();
         ui_manager::redraw();
         refresh_display();
     }


I-am-Erk:
Do we still get the crash if you give him a debug invisibility trait, prior to running the EOC?

mqrause:
Assuming that'd happen at the same point during turn processing, I don't think that would change anything. Plus it would mean actually figuring out the involved json and/or EOC code and fiddling around with it.

I-am-Erk:
I would still like it tested but I am okay with merging this to fix the issue. THe underlying code still needs a fix at some point so I think we should call this a partial fix, not complete.

PR #62368, opened 2022-11-24
Fix inconsistent recipes, All Kevlar jumpsuits need kevlar sheets, none need layered kevlar

bean-b:
#### Summary
Bugfixes "changed xl light kevlar suit recipe"

#### Purpose of change
XL light kevlar jumpsuit should require kevlar sheets like all other kevlar jumpsuits, not layered kevlar panels. Layered kevlar makes no sense here and is likely an oversight, as all other kevlar jumpsuits need kevlar sheets, and the kevlar jumpsuit is made out of kevlar sheets. 200 layered kevlar sheets is an insane amount, which futher proves that this isnt supposed to be like this

#### Describe the solution
changed recipe. Found the same problems with xs and xl steel plated jumpsuit so fixed it there too. And also in xl fur kevlar jumpsuit

na

#### Testing
na


LyleSY (approved):
Logic is solid, numbers make sense to me

PR #62370, opened 2022-11-24
Simple patch for suicidal healing NPCs

### LINKED ISSUE #57254: NPCs suicidally try to heal you
### Describe the bug

If you have a friendly NPC, even if they're guarding, can't see you, and there's a dangerous situation going on, any time you get hurt, they'll run over and try to bandage you. This is not only annoying, it doesn't make a ton of sense. Someone on the other side of a cave shouldn't know my leg is bleeding if I didn't say anything.

A temporary fix for anyone running into this issue is to have NPCs not carry medicine, but that's often a bad idea and doesn't make sense in roleplay.

### Steps to reproduce

Give bandages to a friendly NPC
Get hurt
They will come try to heal you

### Expected behavior

NPCs should have a toggle in their dialog for healing, and they also should have a check for if they can actually see their ally before they run over and try to heal. They should also never do this in combat, though it's possible there's a check there already and I'm just seeing corner cases.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19043.1645 (21H1)
- Game Version: 3e81ac1 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

I-am-Erk:
#### Summary
Bugfixes "Stop NPCs from diving into a firefight to give you a bandaid"

#### Purpose of change

Fixes #57254

#### Describe the solution

This is a simple patch, as noted in the comments it should be expanded on with better behaviour later. However, I think this is also quite reasonable. It just forces NPCs to check for danger before deciding if they want to heal you.

See attached issue.

At a minimum it would be nice to tell NPCs to "never try to heal people" and to "offer healing even if it's a bit dangerous". The latter would increase the acceptable danger level, but not too high.

#### Testing
I am at work and cannot compile. I will remove from draft if someone can playtest this, otherwise I will try to get it running tonight.

### DIFF
--- a/src/npcmove.cpp
+++ b/src/npcmove.cpp
@@ -1811,7 +1811,7 @@ npc_action npc::address_needs( float danger )
     // rng because NPCs are not meant to be hypervigilant hawks that notice everything
     // and swing into action with alarming alacrity.
     // no sometimes they are just looking the other way, sometimes they hestitate.
-    // Eventually we could replace this with somewhat more robust calculations, like basing it 
+    // Eventually we could replace this with somewhat more robust calculations, like basing it
     // on perception and healthcare.
     // Also, NPCs shouldn't consider giving you a bandage if there's still a battle ongoing.
     // The specific danger level they tolerate could later be a behaviour variable.


I-am-Erk:
See comments in linked issue

Issue #3951, opened 2013-10-29
Replace add_spawn() calls in mapgen with place_spawn()


kevingranade:
There are several building layouts in mapgen.cpp and building_generation.cpp that have add_spawn() calls placing specific monsters on the map instead of place_spawn, which is more generic and selects monsters from a group.  This causes issues with the options that modify spawn behavior, because add_spawn() doesn't support those options.
##

Leland:
Only a handful of these left if someone wants to take care of them: https://github.com/CleverRaven/Cataclysm-DDA/search?utf8=%E2%9C%93&q=add_spawn&type=

kevingranade:
Afraid there are quite a few of these left in mapgen.cpp and mission_start.cpp

ZhilkinSerg:
> Afraid there are quite a few of these left in mapgen.cpp and mission_start.cpp

Yes, sorry. I've somehow searched for this function only in `mapgen_functions.cpp`. Working on the rest now.

ZhilkinSerg:
I won't replace `add_spawn` in `mission_start.cpp` for now as `place_spawns` functions is missing `mission_id` and `name` parameters, so it won't fit for mission related spawns:

````C++
void mission_start::place_zombie_mom( mission *miss )
{
    const tripoint house = random_house_in_closest_city();

    miss->target = house;
    overmap_buffer.reveal( house, 6 );

    tinymap zomhouse;
    zomhouse.load( house.x * 2, house.y * 2, house.z, false );
    zomhouse.add_spawn( mon_zombie, 1, SEEX, SEEY, false, -1, miss->uid,
                        Name::get( nameIsFemaleName | nameIsGivenName ) );
    zomhouse.save();
}
````

ZhilkinSerg:
Step 2 is in #22852. When that will be merged only 85 `add_spawn` calls will be left in `mapgen.cpp` - these will be changed in step 3.

ZhilkinSerg:
Step 3 is in #22854. When that will be merged only 67 `add_spawn` calls will be left in `mapgen.cpp` - these will be changed in step 4.

xiaohuynh:
Hi, I'm looking to start contributing to Cataclysm. Could I finish this task, if any `add_spawn` calls are still around?

ZhilkinSerg:
Beware, it can end [bad](https://github.com/CleverRaven/Cataclysm-DDA/pull/29258)

[...]

PR #62374, opened 2022-11-25
convert add_spawn to place_spawns in mx_helicopter

nornagon:
#### Summary
None

#### Purpose of change

Refactoring away from add_spawn per #3951.

#### Additional context

This tweaks the spawn rates slightly, but it should be roughly similar.


nornagon:
Pretty sure the CI failure here is not related to my change :/

Issue #62369, opened 2022-11-24
remove_npcs causes ctd without log


RenechCDDA:
### Describe the bug

See #60938 for a specific occurrence of the issue. Game hangs for a moment or two and then crashes without producing a log. According to investigations by @mqrause this is caused by a combination of the compass sidebar widget trying to update when forced redraw is enabled.

Call stack:
![image](https://user-images.githubusercontent.com/84619419/203868218-f50d4b9b-ceff-4843-a352-cd7a6cc560a8.png)

### Steps to reproduce

http://files.catbox.moe/oevx9t.zip Load this save. Walk north in any build before 18761d73c027906c3ad1cbf42a719fe5eeb815f4.
(Hosted 3rd party due to filesize of ~41MB. In case of loss, another save is available in #60938)

### Expected behavior

Crashing is not desired.

### Versions and configuration

http://files.catbox.moe/oevx9t.zip Load this save. Walk north in any build before 18761d73c027906c3ad1cbf42a719fe5eeb815f4.
(Hosted 3rd party due to filesize of ~41MB. In case of loss, another save is available in #60938)

### Additional context

@I-am-Erk suggested applying debug invisibility to the NPC before mapgen removes him. I locally changed the EOC trigger to this:

```    "speaker_effect": {
      "effect": [
        { "u_add_var": "ANCHOR_SELLER_paranormal", "type": "dialogue", "context": "BEM_ANCHOR_SELLER", "value": "yes" },
	{ "npc_add_trait": "DEBUG_CLOAK" },
        { "queue_eocs": ["EOC_BEM_ANCHOR_SELLER_GONE"], "time_in_future": [ "0s", "0s" ] }
```

It prevents the player from crashing in most instances, but it notably does not work against clairvoyance, which may be acquired with artifacts.
A further suggestion was to change the delay trigger to 1s, like this:
```
        { "queue_eocs": "EOC_BEM_ANCHOR_SELLER_GONE", "time_in_future": [ "1s", "1s" ] }
```

This also prevents the crash in all circumstances that were tested. It is unclear to me or anyone involved *why*. Per @mqrause's discord participation: "[...] I've put up #62362 to fix the crash. But I couldn't figure out what causes the state (he's still visible after completing the dialog) that could lead to the crash in the first place."

RenechCDDA:
/confirming my own issue as I was asked to post this so we could keep track of the general issue.

I-am-Erk:
[...]

PR #62376, opened 2022-11-25
mapgen: remove npc from moninfo cache too

andrei8l:
#### Summary
None

#### Purpose of change
`remove_npcs` mapgen function can lead to segfaults due to stale visible monster cache
Fixes: #62369

#### Describe the solution
Remove the npcs from the visible cache too

mqrause's solution was sufficient

#### Testing
Updated test unit that reproduces the crash condition

#### Additional context
@RenechCDDA can you please verify this solution? I still can't reproduce the crash naturally (without the test unit)

The ASan job is the important one here, but it seems to be broken at the moment.

### DIFF
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -2053,3 +2053,13 @@ void avatar::set_location( const tripoint_abs_ms &loc )
 {
     Creature::set_location( loc );
 }
+
+void monster_visible_info::remove_npc( npc *n )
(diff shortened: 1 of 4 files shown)


RenechCDDA:
I can confirm the crash does not occur on this branch, used the save I provided in the linked issue and tried it a few times just to be sure.

PR #62385, opened 2022-11-25
Change default memory map overlay preset option to Blue Dark

oosyrag:
#### Summary
Interface "Change default memory map overlay preset option to Blue Dark"

#### Purpose of change

I recall being very disorientated as a new player when the memory tiles are rendered brighter than your immediate visible surroundings while in the dark. A new player is very likely to find themselves in the dark when starting a game. Screenshots attached for emphasis. I think this would be a nice change for people coming into a new stable version and ease the barrier of entry.

Sepia light:
Screenshot_20221125-091825_a77f7b094295b5a083bf400fa615cbfb

Blue dark:
Screenshot_20221125-091639_a77f7b094295b5a083bf400fa615cbfb

#### Describe the solution

Changed the default memory map overlay preset setting to Blue Dark instead of Sepia Light, which I find much more intuitive, and is also suggested per https://www.reddit.com/r/cataclysmdda/comments/ppwq9f/cataclysm_dda_on_android_an_indepth_tweaking_guide/

Leave as is

#### Testing
None

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -2201,7 +2201,7 @@ void options_manager::add_options_graphics()
         { "color_pixel_sepia_dark", to_translation( "Sepia Dark" ) },
         { "color_pixel_blue_dark", to_translation( "Blue Dark" ) },
         { "color_pixel_custom", to_translation( "Custom" ) },
-    }, "color_pixel_sepia_light", COPT_CURSES_HIDE
+    }, "color_pixel_blue_dark", COPT_CURSES_HIDE
        );
 
     add( "MEMORY_RGB_DARK_RED", "graphics", to_translation( "Custom dark color RGB overlay - RED" ),


esotericist:
obviously as the person who implemented the sepia tint (back before we had all these settings) i'm a little biased, but part of the intent was to make it as distinct from both low-light and full-light vision as reasonably possible; with washed out colors, it helped draw a distinction to "this is something different", especially in the context where it shows up. bluedark might be more aesthetic to many people, but it would be really easy for a player to confuse it for just being a variant on low light.

and that's not hypothetical, the original problem sepia was implemented to solve was "darker coloration gets confused for low light vision".

oosyrag:
Thanks for the explanation, I understand changing defaults is usually a touchy subject since the original default was probably set for a reason. I was hoping it was simply that that was what it was at first before the other options, but I see that's not the case.

I did think that it was weird, wondered if I was in a dark spot and I could see those light areas because they were lit by a window or lamp or something. Wandered around some more, realize the dark spot moved with me and literally everything else is "lit". Found the option, changed it, and never looked back. I'm personally not convinced that dark tint=dark area is a problematic assumption. It also comes with the assumption that I can't see what is in the dark, and that changes in areas I can't see won't be visible until I can see it.

Anyways I'm not particularly against leaving the default as is, considering it doesn't really affect me forever after I set it once, and remain grateful for the option to change it.

kevingranade:
Basically on one side we have the status quo which comes with a couple of arguments for why it's better (matches curses, harder to confuse with generic dimming) and on the other side we have, "I think it looks better".

Unless a stronger rationale for something else develops, we're sticking with the current setting.

oosyrag:
Just wanted to clarify that I didn't suggest this change because just because I think it looks better, but specifically that blue dark may be more intuitive at a glance to a new player as to what the tint represents.

kevingranade:
"Even darker" intuitively means, "even darker", not "you can't see this right now, but you remember seeing it".

PR #62392, opened 2022-11-25
Fixes for temperature regulation

### LINKED ISSUE #57033: Identical clothing made of high (Lycra) breathability material performs worse than poor (Fur) at lowering body heat
### Describe the bug

As I understand it, the intent of the new perspiration mechanics and breathability values is to allow the player to (somewhat) better cool off in hot weather by sweating, just as in real life.  Being _wet_ from sweat isn't the thing that cools the player (or a real person) down but rather the evaporative cooling (https://en.wikipedia.org/wiki/Evaporative_cooler#Physical_principles) effect carries heat away and reduces body temperature.

So if we sweat, but something (like wearing non-breathable clothing) prevents the sweat from _evaporating_ efficiently, this impedes the evaporative cooling effect so body temperature will remain higher.  (The person will also feel uncomfortable and 'muggy' with all the sweat trapped against their skin and unable to evaporate)

Based on this -- clothing with high breathability like Lycra (110) should perform better than those with poor breathability like Fur (30) in terms of allowing the player to stay cool and comfortable in hot weather.

Instead what I have observed through some methodical testing is that the opposite seems true:  A player wearing an otherwise identical (same coverage, same warmth, same flags, only material changed) jumpsuit made of Lycra and simply standing in the sun for several hours ends up with _higher_ body temperatures and incurring more Heat-Slowdown penalties for longer than one wearing a Fur jumpsuit.

Breathability does appear to be 'working' in the sense that the Lycra wearing player stays _dryer_ overall and does not end up with "Wet" morale penalties like the Fur wearing player.  However the enhanced sweat _removal_ the Lycra wearing player gets doesn't appear to translate into enhanced _evaporative cooling_ the way one would expect.

[test_jumpsuits.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/8543915/test_jumpsuits.zip)

[Eskridge.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/8543921/Eskridge.zip)

### Steps to reproduce

It's a bit difficult to accurately test this in-game currently since the different clothing items with different materials/breathability values _also_ have completely different warmth, coverage, flags, and other variables that would confuse the issue.  So in order to perform a 'clean' test I quickly modded some identical jumpsuits with 95% coverage of torso, arms, legs, and feet and 60 warmth.  I started with a newly generated world with starting day in Summer on the latest experimental and a no-skills no-traits all 8-8-8-8 character.

1. Load up the attached save which I used as the starting point for all tests performed, along with the test_jumpsuits.zip mod.
2. Use Debug menu to set Weather to Clear and Wind Speed to 0 (in that order) to remove these variables that might affect the test.
3. Find test jumpsuits of Fur, Cotton, and Lycra on the ground at your feet along with some bottled water.
4. Note that player is already Turgid as I started by drinking two bottles of water to allow for plenty of sweating.
5. Wear the Lycra jumpsuit.
6. Wait 1 hour and observe body temperature results via debug menu, check status/morale penalties & Heat Slowdown.
7. Drink a unit of water to replenish sweat & Wait 1 more hour, observe results again.
8. Drunk another unit of water, wait 1 more hour, observe results.
9. Repeat the above by reloading the save and wearing the Fur jumpsuit instead.
10. See that wearing an identical jumpsuit with 30 breathability (fur) results in lower overall body temperature and less Heat Slowdown than 110 breathability (Lycra) over time.

### Expected behavior

I expected the Lycra (high 110 breathability) to not only more efficiently _remove_ sweat from the player, but for the improved moisture-wicking evaporation to much more effectively cool the player than Fur (poor 30 breathability) garment with otherwise identical properties.

In general, the player should be able to maintain noticeably cooler body part temperatures while wearing high breathability clothing as a result of all that sweat efficiently evaporating away & causing evaporative cooling effect _rather_ than being trapped against their skin making them soaked and miserable.

Right now it seems the "soaked and miserable" part of poor breathability is working, and the "fast sweat _removal_" part of high breathability is working, but the "enhanced evaporative cooling effect" of high breathability is _not working_.  

### Screenshots

Baseline before starting tests, naked: Summer day 35 @ 11 AM, 75 F with "Clear" weather & 0mph wind set by debug menu - player drank 4 units of water for 'Turgid' status:
Naked-Baseline-11AM

After wearing **FUR** test jumpsuit (95 coverage, 60 warmth, 30 breathability) for **1 hour** & drinking 1 unit water:
Fur-1HOUR

After wearing **FUR** test jumpsuit for **2 hours** & drinking 1 unit water:
Fur-2HOUR

After wearing **FUR** test jumpsuit for **3 hours**:
Fur-3HOUR

After wearing **LYCRA** test jumpsuit for **1 HOUR**:
Lycra-1HOUR

After wearing **LYCRA** test jumpsuit for **2 HOURS**:
Lycra-2HOUR

After wearing **LYCRA** test jumpsuit for **3 HOURS**:
Lycra-3HOUR

- OS: Windows
    - OS Version: 10.0.22598.200 (22H2)
- Game Version: 61562af [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Temperature Test Jumpsuits [testing_jumpsuits]
]

### Additional context

_No response_

bombasticSlacks:
#### Summary
Bugfixes "Fixes for perspiring under water and good breath ability actually being bad"

#### Purpose of change
fixes #57033 and an old bug never reported

#### Describe the solution
Added a check if you are under water before perspiration occurs,

Added a limit to how dry you can be while hot, this means that you will never over perspire missing ticks of cooling down. This is a bit of a hacky solution but it does lead to the intended fix of breathable clothing being much better at keeping you cool.

The whole system would need a rewrite, which is way too big of undertaking this close to .G

#### Testing
There are now unit tests for this

### DIFF
--- a/data/mods/TEST_DATA/items.json
+++ b/data/mods/TEST_DATA/items.json
@@ -3554,6 +3554,75 @@
     "armor": [ { "encumbrance": 70, "coverage": 100, "covers": [ "hand_l", "hand_r" ] } ]
   },
   {
+    "id": "test_jumpsuit_lycra",
+    "type": "ARMOR",
+    "name": { "str": "test_jumpsuit_lycra" },
+    "description": "Lycra test jumpsuit",
+    "weight": "610 g",
+    "volume": "3500 ml",
+    "price": 8900,
+    "price_postapoc": 50,
+    "to_hit": -3,
+    "material": [ "lycra" ],
+    "symbol": "[",
+    "looks_like": "touring_suit",
+    "color": "yellow",
+    "armor": [
+      { "covers": [ "torso" ], "coverage": 95, "encumbrance": [ 1, 2 ] },
+      { "covers": [ "arm_l", "arm_r" ], "coverage": 95, "encumbrance": [ 1, 1 ] },
+      { "covers": [ "leg_l", "leg_r" ], "coverage": 95, "encumbrance": [ 1, 2 ] },
+      { "covers": [ "foot_l", "foot_r" ], "coverage": 95, "encumbrance": [ 1, 2 ] }
+    ],
+    "warmth": 60,
+    "material_thickness": 0.2
+  },
+  {
+    "id": "test_jumpsuit_fur",
+    "type": "ARMOR",
+    "name": { "str": "test_jumpsuit_fur" },
+    "description": "Fur test jumpsuit",
+    "weight": "610 g",
+    "volume": "3500 ml",
+    "price": 8900,
+    "price_postapoc": 50,
+    "to_hit": -3,
+    "material": [ "fur" ],
+    "symbol": "[",
+    "looks_like": "touring_suit",
+    "color": "yellow",
+    "armor": [
+      { "covers": [ "torso" ], "coverage": 95, "encumbrance": [ 1, 2 ] },
+      { "covers": [ "arm_l", "arm_r" ], "coverage": 95, "encumbrance": [ 1, 1 ] },
+      { "covers": [ "leg_l", "leg_r" ], "coverage": 95, "encumbrance": [ 1, 2 ] },
+      { "covers": [ "foot_l", "foot_r" ], "coverage": 95, "encumbrance": [ 1, 2 ] }
+    ],
+    "warmth": 60,
+    "material_thickness": 0.2
+  },
+  {
+    "id": "test_jumpsuit_cotton",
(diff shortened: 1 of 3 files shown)


kenneth-ellis:
this is awesome, thank you!

PR #62395, opened 2022-11-26
Riot armor nerf

bean-b:
#### Summary
Balance "Nerfed riot armor"

#### Purpose of change
I updated riot armor in https://github.com/CleverRaven/Cataclysm-DDA/pull/62048, but I_Am_Erik mentioned that it might have been a little to strong, and I totally see that, especially with how relatively easily they are too loot from cops etc.. 

#### Describe the solution
The hip armor now represents just a cod piece, and doesn't contain the plastic padding, as that seemed more consistent with models online. The chest piece is 0.5mm less thick. This still feels like reasonable amount of thickness, but will reduce its protection values a bit.

Increasing encumbrance, but given the hard chest/leg/arm guard encumbrances, which are v similar to riot armor, i didint wanna increase the encumbrance to much,

#### Testing
spawned in game, looked at values

#### Additional context
new values
image

### DIFF
--- a/data/json/items/armor/suits_protection.json
+++ b/data/json/items/armor/suits_protection.json
@@ -1382,7 +1382,7 @@
         "covers": [ "torso" ],
         "material": [
           { "type": "plastic_pad", "covered_by_mat": 100, "thickness": 7 },
-          { "type": "thermo_resin", "covered_by_mat": 90, "thickness": 2 }
+          { "type": "thermo_resin", "covered_by_mat": 90, "thickness": 1.75 }
         ]
       },
       {
@@ -1391,7 +1391,7 @@
         "coverage": 90,
         "material": [
           { "type": "plastic_pad", "covered_by_mat": 100, "thickness": 5 },
-          { "type": "thermo_resin", "covered_by_mat": 90, "thickness": 2 }
+          { "type": "thermo_resin", "covered_by_mat": 90, "thickness": 1.75 }
         ],
         "specifically_covers": [ "arm_lower_l", "arm_lower_r", "arm_elbow_l", "arm_elbow_r", "leg_knee_l", "leg_knee_r", "leg_lower_l", "leg_lower_r" ]
       },
--- a/data/json/items/armor/arms_armor.json
+++ b/data/json/items/armor/arms_armor.json
@@ -1705,7 +1705,7 @@
         "coverage": 90,
         "material": [
           { "type": "plastic_pad", "covered_by_mat": 100, "thickness": 5 },
-          { "type": "thermo_resin", "covered_by_mat": 90, "thickness": 2 }
+          { "type": "thermo_resin", "covered_by_mat": 90, "thickness": 1.75 }
         ],
         "specifically_covers": [ "arm_lower_l", "arm_lower_r", "arm_elbow_l", "arm_elbow_r" ]
       }
--- a/data/json/items/armor/legs_armor.json
+++ b/data/json/items/armor/legs_armor.json
@@ -1640,7 +1640,7 @@
         "coverage": 90,
         "material": [
           { "type": "plastic_pad", "covered_by_mat": 100, "thickness": 5 },
-          { "type": "thermo_resin", "covered_by_mat": 90, "thickness": 2 }
+          { "type": "thermo_resin", "covered_by_mat": 90, "thickness": 1.75 }
         ],
         "specifically_covers": [ "leg_knee_l", "leg_knee_r", "leg_lower_l", "leg_lower_r" ]
       },


bean-b:
If people dont think it is to strong, im happy to close this pr and leave it be too.

oosyrag:
Well riot armor was stronger before the layered armor update early this year. I'm sure mylie had his reasons for the previous values.

On the other hand, I'm constantly reminded that ease of access doesn't necessarily determine power levels just for the sake of gaming. Things manufactured pre cataclysm can be both good and common.

bean-b:
> On the other hand, I'm constantly reminded that ease of access doesn't necessarily determine power levels just for the sake of gaming. Things manufactured pre cataclysm can be both good and common.
No thats fair, I did leave my nerfs fairly small so that the armor is still pretty strong. But yea I could see not doing this too, I was just going off a comment by  I_Am_Erik on reddit.

bean-b:
> Well riot armor was stronger before the layered armor update early this year. I'm sure ~mylie~ Erk had his reasons for the previous values.
> 
> On the other hand, I'm constantly reminded that ease of access doesn't necessarily determine power levels just for the sake of gaming. Things manufactured pre cataclysm can be both good and common.
> 
> Edit: Oops it wasn't Mylie, Mylie did the chitin armor. It was Erk. Ref: #52013

image

I based my armor values off of something like this, which seems to be a largely different suit then erik had as his refernce picture. At least to me, his refernce picture looks more like arm guards + leg guards + a protective vest, and not the whole suit

data/json/items/armor/suits_protection.json
```diff
@@ -1400,7 +1400,7 @@
         "coverage": 85,
         "material": [
           { "type": "plastic_pad", "covered_by_mat": 100, "thickness": 6 },
-          { "type": "thermo_resin", "covered_by_mat": 85, "thickness": 1.5 }
+          { "type": "thermo_resin", "covered_by_mat": 80, "thickness": 1.5 }
```
I-am-Erk, line 1403:
```suggestion
          { "type": "thermo_resin", "covered_by_mat": 70, "thickness": 1.5 }
```
The riot armour I see still leaves a lot of open space on the underarms and places like that, the hard plates only cover the outside of the arm. Those would definitely absorb a majority of attacks but I don't think we should oversell them.

data/json/items/armor/suits_protection.json
```diff
@@ -1413,11 +1413,8 @@
       },
       {
         "covers": [ "leg_l", "leg_r" ],
-        "coverage": 95,
-        "material": [
-          { "type": "plastic_pad", "covered_by_mat": 100, "thickness": 6 },
-          { "type": "thermo_resin", "covered_by_mat": 20, "thickness": 1.5 }
-        ],
+        "coverage": 30,
+        "material": [ { "type": "thermo_resin", "covered_by_mat": 100, "thickness": 1.5 } ],
```
I-am-Erk, line 1417: I think your original stats were probably better here?

data/json/items/armor/torso_armor.json
```diff
@@ -2083,17 +2083,14 @@
         "coverage": 85,
         "material": [
           { "type": "plastic_pad", "covered_by_mat": 100, "thickness": 6 },
-          { "type": "thermo_resin", "covered_by_mat": 85, "thickness": 1.5 }
+          { "type": "thermo_resin", "covered_by_mat": 80, "thickness": 1.5 }
```
I-am-Erk, line 2086:
```suggestion
          { "type": "thermo_resin", "covered_by_mat": 70, "thickness": 1.5 }
```
as above

data/json/items/armor/torso_armor.json
```diff
@@ -2083,17 +2083,14 @@
...
       },
       {
         "covers": [ "leg_l", "leg_r" ],
-        "coverage": 100,
-        "material": [
-          { "type": "plastic_pad", "covered_by_mat": 100, "thickness": 6 },
-          { "type": "thermo_resin", "covered_by_mat": 20, "thickness": 1.5 }
-        ],
+        "coverage": 30,
+        "material": [ { "type": "thermo_resin", "covered_by_mat": 100, "thickness": 1.5 } ],
```
I-am-Erk, line 2093: as above

I-am-Erk (commented):
I'd like to see what your final protection values come to as well.

bean-b:
> I'd like to see what your final protection values come to as well.
> 
> EDIT: You may also want to slightly reduce thickness of the plastic and reduce its coverage to like 95%, and have a thick cloth layer for the 100%. There are still small weak points all over the armour in armpits and elbows and things.

Here are the protection values as of right now.
image

I did change the groin protector to model the second image more then the first.

Nowhere is the total coverage 100%, even if the plastic padding is 100% in some places, the default coverage is always less then 100%. This has a multiplicative effect with the coverage of the thermoplastic resin. For example, in the thighs, shoulders etc. the chance for a no protection at all is 25%, and the chance for full protection inc. the thermoplastic resin layer is only 0.75*0.85 =63.75%.

bean-b:
> Looking at the torso only, I would consider the 90% torso coverage to be too generous, considering there's no armor on the sides, and even the front/back coverage isn't anywhere near complete, such as around the waist. Consider a fully encased torso in plate without sleeves is considered 95% coverage (5% being the armpit/arm opening area). All the gaps are for the benefit of the low encumbrance of such a suit, so I'd put the total coverage at 70-80%. Likewise for the arm and leg parts, those look to be 50-60% to me. Again coming from the perspective that a fully fitted sleeve or pants being 100%, and think how much material would you have to cut out to get coverage as shown in the picture.
> 
Yea i get that, Ill go ahead and bring the coverage values down some more.
> As for thickness, you said you based it "current" riot armor in the previous pr, but the previous value was 1.25mm for both, so how did it get 3-4 times thicker all of a sudden? Any data on the actual thickness of the material from source?
>
The differnce here is that it now uses plastic padding instead of plastic, which has lower protection values. The thermoplastic thickness is not far off from where it was before. I ended on the 6-8mm values based on this convo on the previous pr. 
image

bean-b:
> I'd like to see what your final protection values come to as well.
> 
Ill go ahead and put this at the top of the pr

oosyrag:
Sure sounds good. Does help to see the final values too, the old values were 7.5/10/10. I'm all for a buff but we're up to 14/16/16 now. For reference 3mm of steel plate is 20/20/13 or so. I don't have any objective input on what it should be, but subjectively speaking I'd lean more towards 10/12/12 values as seen on the limbs, which would still be an improvement over the previous.

bean-b:
> Sure sounds good. Does help to see the final values too, the old values were 7.5/10/10. I'm all for a buff but we're up to 14/16/16 now on the torso. For reference 3mm of steel plate is 20/20/13 or so. I don't have any objective input on what it should be, but subjectively speaking I'd lean more towards 10/12/12 values as seen on the limbs, which would still be an improvement over the previous.

>
ok, i lowered the thicknesses, you can see it reflected in the prot values at top. And even while the riot armor has good protecton values, it is severly limited by its coverages, an armor with 65/75% coverage in most places can only ever be so good no matter the prot values

I-am-Erk:
These values look pretty good now I think. THank you.

bean-b:
> These values look pretty good now I think. THank you.

Yep of course :). sorry for getting the numbers wonky on the first go around

Issue #57585, opened 2022-05-09
Avoid possible epilepsy hazard by keeping Adv Inventory open during brief tasks


tenth-speed-writer:
### Is your feature request related to a problem? Please describe.

When performing fast actions with short in-game action times from the Advanced Inventory Screen (such as looting containers), especially when using a bright tileset, the player experiences a severe flashing light effect from the window closing and opening again.

This poses an epileptic seizure hazard.

### Solution you would like.

Don't close or hide the Advanced Inventory Screen if an action the player character takes from the Advanced Inventory Screen requires less than a certain amount of time to pass in-game.

### Describe alternatives you have considered.

The issue is considerably less severe when playing without tiles.

If the above suggestion is difficult or impractical, consider adding an epilepsy hazard notice at game launch and suggest ascii-based graphics or alternative tilesets to users who are sensitive to flashing lights.

I-am-Erk:
This doesn't seem like a tough thing to fix (though that's loaded when AIM and UI are concerned) and I agree it's an accessibility issue for some, though the seizure risk should be low.

Jan-Blasiak:
/confirmed

Though I usually play darker tilesets, I got nauseous few times due to flashing menus and stuff after longer games.

Chewing 5-10 second things causes flashing, moving lots of different stuff via Adv. Menu, shooting with snap-to-target aiming, there are probably few more things I can't remember of the top of my head.

I-am-Erk:
I am inclined to leave this on release blockers because I agree it is a sensory issue for accessibility. However, I am concerned it remains untouched because this is some very hard code. I would really appreciate an educated thought on how difficult this will be to solve, and if it is going to be too hard for stable I will reluctantly defer it.

Zireael07:
Not an epilepsy sufferer, but I agree this is annoying to us healthy folks, and a real problem for sufferers. I think adding a requirement of a minimum action time to AIM close shouldn't be too hard unless there's some weird coupling between the AIM and duration classes preventing the easiest fix (which should amount to just slapping an if)

mqrause:
The menu needs to be closed for activity processing to happen, because the menu itself is an activity. That's what I did with wield/wear from pickup menu and I pretty much copied that from the eat menu. Past me from #53793 says AIM also works like that.
[...]

PR #62400, opened 2022-11-26
AIM: keep ui_adaptor running during move activities

andrei8l:
#### Summary
None

#### Purpose of change
Keep the AIM open when moving items to prevent bright flashes
Fixes: #57585

#### Describe the solution
Keep the `ui_adaptor` running while moving items

Zen

#### Testing
1. Open the AIM and move one/all items. Try again with the source pane set to All, which has a completely separate processing path for reasons. Try again with the destination pane set to Inventory, which has a completely separate processing path for _reasons_.
2. Spawn a zombie or otherwise set up an interruption (including cancelling manually) and try 1 again
3. Close and reopen the AIM after 1 and 2. Everything should work as expected

#### Additional context
I-am-Erk ping back

### DIFF
--- a/src/advanced_inv.cpp
+++ b/src/advanced_inv.cpp
@@ -104,10 +104,27 @@ std::string enum_to_string<aim_entry>( const aim_entry v )
 
 } // namespace io
 
+namespace
+{
+std::unique_ptr<advanced_inventory> advinv;
+} // namespace
+
 void create_advanced_inv()
 {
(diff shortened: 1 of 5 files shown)


mqrause:
Does this still allow manual interruption?

andrei8l:
Yes. But there are a lot of edge cases in the `move_all()` decade-old spaghetti. That function should have been removed wholesale when player activities were introduced.

PR #62405, opened 2022-11-26
Fix Clang 12 on Ubuntu build failures on GitHub Actions

BrettDong:
#### Summary
None

#### Purpose of change
The "Clang 12 on Ubuntu" job on GitHub Actions has been [failing](https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/3554228319/jobs/5970407308#step:13:92) recently because Clang is mixing system headers from GCC 11 and GCC 12 on the system.
```
In file included from src/animation.cpp:43:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/thread:44:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/this_thread_sleep.h:36:
Error: /usr/bin/../lib/gcc/x86_64-linux-gnu/12/../../../../include/c++/12/bits/chrono.h:82:12: error: redefinition of '__duration_common_type<_CT, _Period1, _Period2>'
    struct __duration_common_type<_CT, _Period1, _Period2,
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/chrono:97:12: note: previous definition is here
    struct __duration_common_type<_CT, _Period1, _Period2,
           ^
```

#### Describe the solution
I'm using the solution described in https://stackoverflow.com/a/41982109 which sets up symbolic links to create a mock GCC toolchain root path that contains only one GCC version, so Clang won't get confused.

An alternative is just switch to `libc++` for standard libraries.

#### Testing
The "Clang 12 on Ubuntu" job should pass.

### DIFF
--- a/.github/workflows/matrix.yml
+++ b/.github/workflows/matrix.yml
@@ -121,6 +121,7 @@ jobs:
             localize: 1
             native: linux64
             sanitize: address
+            cxxflags: --gcc-toolchain=/opt/mock-gcc-11
             mods: --mods=magiclysm
             dont_skip_data_only_changes: 1
             title: Clang 12, Ubuntu, Tiles, ASan
@@ -211,6 +212,7 @@ jobs:
         MODS: ${{ matrix.mods }}
         SANITIZE: ${{ matrix.sanitize }}
         TEST_STAGE: ${{ matrix.test-stage }}
+        CXXFLAGS: ${{ matrix.cxxflags }}
         LDFLAGS: ${{ matrix.ldflags }}
         EXTRA_TEST_OPTS: --error-format=github-action
         NUM_TEST_JOBS: ${{ matrix.num_test_jobs }}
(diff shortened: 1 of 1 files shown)


BrettDong:
This fixes the compile error and surfaces a magicalysm item density test error, which is addressed by #62413

akrieger:
I don't know why but this doesn't seem to work in my repo. Seems like it changed one error for another. https://github.com/akrieger/Cataclysm-DDA/actions/runs/3604675557/jobs/6074285815

```
Error: pch/main-pch.hpp:13:10: fatal error: 'algorithm' file not found
#include <algorithm>
         ^~~~~~~~~~~
1 error generated.
```

BrettDong:
Because it's running on Ubuntu 20.04:

https://github.com/akrieger/Cataclysm-DDA/actions/runs/3604675557/jobs/6074285815#step:1:4

So my fix for Ubuntu 22.04 does not apply here.

---

https://github.blog/changelog/2022-11-09-github-actions-ubuntu-latest-workflows-will-use-ubuntu-22-04/

During the transitional period, a job `ubuntu-latest` may run on either Ubuntu 20.04 or Ubuntu 22.04. It's bad. I think we should clearly specify `ubuntu-22.04` in workflow yaml.

PR #62407, opened 2022-11-26
unhardcode mx_roadblock

nornagon:
#### Summary
None

#### Purpose of change

Refactoring away from add_spawn per #3951.


nornagon:
Merged in code from #62415 to unhardcode mx_roadblock.

Night-Pryanik:
Why are replacing working generator with a broken one? This change will also render searchlights non-operable.

nornagon:
tagging @GuardianDll who did the mapgen json, do you have thoughts about what @Night-Pryanik said above?

gettingusedto:
Does adding the "portable generator" vehicle and a couple of small storage batteries suffice or is it just a little too much?

data/json/mapgen/map_extras/roadblock_mil.json
```diff
@@ -0,0 +1,43 @@
...
+        "----------------------  ",
+        "----------------------  ",
+        "----------------------  ",
+        "-----------t----------  ",
+        "----------------------  ",
+        "----bbbbbbb--bbbbbbb--  ",
+        "                        ",
+        "                        "
+      ],
+      "furniture": { "s": "f_sandbag_half", "g": "f_generator_broken", "b": "f_barricade_road" },
```
DSeyka, line 33: If the generator is broken, how do turrets and searchlights still work?
I thought ASRG was added as an explanation to that.

GuardianDll:
The reason is "we have no working generator furniture in the game except asrg, and adding one is out of scope of this pr. If someone know how to make such generator, I'd be glad to add one

Night-Pryanik:
The special "attack" of the searchlight is hardcoded and explicitly requires `Compact Advanced Sterling Radioisotope Generator` to work:
https://github.com/CleverRaven/Cataclysm-DDA/blob/b50f199d9b48572021bcf7fc9bf81ef8a7ce8dca/src/monattack.cpp#L3828-L3829
It won't work with any other type of generators.

GuardianDll:
Well. then it means someone else need to replace this generators in the future
I'll turn the ASRG back then

data/json/mapgen/map_extras/roadblock.json
```diff
@@ -0,0 +1,59 @@
...
+        "----------------------  ",
+        "----------------------  ",
+        "----------------------  ",
+        "------------t---------  ",
+        "----------------------  ",
+        "----bbbbbbb--bbbbbbb--  ",
+        "                        ",
+        "                        "
+      ],
+      "furniture": { "s": "f_sandbag_half", "g": "f_generator_broken", "b": "f_barricade_road" },
```
GuardianDll, line 43:
```suggestion
      "furniture": { "s": "f_sandbag_half", "g": "f_compact_ASRG", "b": "f_barricade_road" },
```

data/json/mapgen/map_extras/roadblock_mil.json
GuardianDll, line 33:
```suggestion
      "furniture": { "s": "f_sandbag_half", "g": "f_compact_ASRG", "b": "f_barricade_road" },
```

data/json/mapgen/map_extras/roadblock.json
Night-Pryanik, line 43:
```suggestion
      "furniture": { "s": "f_sandbag_half", "g": "f_compact_ASRG_containment ", "b": "f_barricade_road" },
```

data/json/mapgen/map_extras/roadblock_mil.json
Night-Pryanik, line 33:
```suggestion
      "furniture": { "s": "f_sandbag_half", "g": "f_compact_ASRG_containment ", "b": "f_barricade_road" },
```

oosyrag:
Wouldn't it make more sense to change the searchlight ability to not require an asrg instead? 

The generator could be a part of the searchlight itself.
https://www.larsonelectronics.com/category/599/Light-Towers-with-Generators

GuardianDll:
After 0.G

I-am-Erk:
We definitely don't want ASRG to be this common, they're a pretty rare and expensive object.

RenechCDDA:
Heya, with #62459 merged you should be able to go ahead with this by replacing the ASRG with `f_active_backup_generator`.

data/json/mapgen/map_extras/roadblock_mil.json
GuardianDll, line 33:
```suggestion
      "furniture": { "s": "f_sandbag_half", "g": "f_active_backup_generator", "b": "f_barricade_road" },
```

data/json/mapgen/map_extras/roadblock.json
GuardianDll, line 43:
```suggestion
      "furniture": { "s": "f_sandbag_half", "g": "f_active_backup_generator", "b": "f_barricade_road" },
```

GuardianDll (commented):
so as RenechCDDA says, we can use better thing that ASRG from now

Night-Pryanik:
@nornagon could you please resolve conflicts?

nornagon:
@Night-Pryanik done

nornagon:
@Night-Pryanik Conflicts resolved again, mind giving this another look before we conflict again 😅

Night-Pryanik:
Another conflict coming on.

nornagon:
aargh 😓 fixed! looks like Dark Skies Above got removed?

Night-Pryanik:
Yep.

Issue #56990, opened 2022-04-21
BUG/Suggestion: Either make trader NPCs NOT investigate sounds or lock them behind reinforced glass.


Jan-Blasiak:
### Is your feature request related to a problem? Please describe.

Last time I did the "Traitor" quest for the Merchants, shooting the traitor caused the Merchant behind the locked counter to smash his own window, and run into the corridor, where he now sits, away from his designated location AND his pile of goods for sale.

### Solution you would like.

Either make trader NPCs, preferably all of them for safety sake, NOT investigate sounds or lock them behind reinforced glass/steel bars or other applicable "cages". It's EXTREMELY irritating having to go to the corridor to trade with Merchant.

Inglonias:
I think the simplest solution here is to lock Smokes behind reinforced glass, yeah.

Ilysen:
Smokes actually \*is\* behind reinforced glass, it's just that his weapon (a Halligan bar) has very high bash and can break it. #56195 has been up to fix that, but has yet to be merged.

I-am-Erk:
So, the solution here is probably not to make Smokes pound ineffectually on the indestructible glass. I suggest we set up an effect series for merchants like Smokes, and the doctor on the other side of the room, and probably a few more: drop their movement speed to 0 so that they cannot leave their spot, and add an Effect on Condition that turns that effect off if they're in danger. This will hold them in place and stop them from doing weird things in other situations

For the EoC I'd turn it off if:
- The door opens
- The window is broken
- The merchant takes damage
- The player becomes hostile to the faction

The EoC that makes the merchant mobile could also unlock the merchant's door temporarily, as long as we have a setup that can get them back to their position and re-lock the door later.

Maleclypse:
@Jan-Blasiak I can’t seem to tag you on https://github.com/CleverRaven/Cataclysm-DDA/pull/57072

but it’s been merged and I’m concerned maybe I added your name incorrectly.

I-am-Erk:
I am making some recommendations to push forward for stable release. For this issue, I suggest that we resolve it *in stable only* by:
1. Create a mutation that makes a character unable to move. 
2. Apply that mutation to smokes, the refugee doctor, the broker, Rubik, and any other merchant or similar NPCs that really shouldn't leave their spots.
3. Although the mutation can be kept in the code (it can be handy for mods), it should only be applied to these NPCs in the G-stable release. It should be removed for H-experimental.

[...]

PR #62408, opened 2022-11-26
Create NOMOVE trait and add to faction NPC's

mythosmod:
#### Summary

Bugfixes "Create NOMOVE trait and add to faction NPC's"

#### Purpose of change

Temporary fix to #56990 as suggested by @I-am-Erk - create a mutation to set NPC speed to 0 and apply to suggested NPC's to stop them smashing their way out of where they are to investigate noise, hopefully removing a blocker for 0.G

#### Describe the solution

- Add NOMOVE mutation in JSON
- Add NOMOVE trait into character.cpp
- Under Character::get_speed(), make speed return 0 if this mutation is present
- Add trait to Smokes, the refugee doctor, the broker, Rubik and the Artisans (Jay and Cody)

This trait should not be applied to player as it locks the game loop up permanently!

Implement full fix as outlined by Erk in #56990 but as this is my first C++ change for CDDA wanted to take baby steps :)

#### Testing

Complied and built to Cataclysm-lib-vcpkg-static
Created standard world and loaded default scenario
Added refugee center location from PC
Teleported to refugee center
Spoke to Smokes and made sure he was able to talk
Spawned feral in lobby, guards killed but no movement from Smokes
Yelled around the corner, Smokes reacted but did not move to investigate like the guards

#### Additional context

Yelling out of sight in the refugee center
Smokes 1

All others come to investigate but Smokes stays put
Smokes 2

Smokes still able to talk
image


Drew4484:
Artisans? Any reason for them to move around given their turrets defending the exterior?

EIIKaO:
This is a good initiative.

However, with this process, it seems that they will continue to stand on their sticks without escape even if they are attacked by enemies( include player character ) or are in direct mortal danger (e.g., fire). In my opinion, it is not a good behavior.

What are your thoughts on this point?

mythosmod:
> This is a good initiative.
> 
> However, with this process, it seems that they will continue to stand on their sticks without escape even if they are attacked by enemies( include player character ) or are in direct mortal danger (e.g., fire). In my opinion, it is not a good behavior.
> 
> What are your thoughts on this point?

That is entirely the case - as they have no speed they cannot move or attack, the aim of this PR was to get past the block in the way suggested in #56990 which would then be removed for the next experimental.

Two alternatives suggest themselves:

Alter the code to add in a trait\mutation\option to turn off investigating unfamiliar noises within the NPC AI and apply it to these NPC's - this would mean that the behavior was unchanged beyond not investigating noises, the code in question appears to be in npcmove.cpp and I am happy to look into this but I don't think it would be as quick a fix (there is also some code in there regarding returning guards to positions which could possibly be leveraged to implement the NPC go home previously suggested)

The other suggestion would be an extension of this trait as suggested by Erk - create an EOC that removes the trait under dangerous conditions, so the NPC is able to react to that danger as expected. Again this would be a longer fix but one that could be implemented in JSON without having to touch the source code.

andrei8l:
Please add the flag to the NPCs in the caravan encounter as well (`FM_caravan_merchant_random`, `FM_caravan_guard_A`, `FM_caravan_guard_B`). Otherwise they can wander off and never vanish as intended.

LyleSY:
I like the tight focus here. It would be nice to add the Tacoma Ranch NPCs here to try and preserve that mission chain

mythosmod:
@LyleSY @I-am-Erk @Drew4484 @EIIKaO @andrei8l - I have created an alternative PR with a slightly different approach (see above) - I think this new one may be the better approach as it stops NPC's investigating sounds but doesn't prevent them from moving/attacking completely, let me know what you think

PR #62414, opened 2022-11-26
Allow the ability to smash a skylight

gettingusedto:
#### Summary
None

#### Purpose of change

Allow the ability to (s)mash up skylights.

#### Describe the solution

When I was playing, I was disappointed by the fact that I can't smash a skylight so I can make a hole in the celling and go down from it. After looking in its json for awhile, it looks like it has `"bash_below": true`, which I assume means that it's supported from below by a beam of sorts. It doesn't make much sense to me because most of the skylight I've seen aren't supported by a beam at all. So i deleted `"bash_below": true` to allow me to bash the skylight down.

Not doing it.

#### Testing

Made the aforementioned change on my game, put a skylight with a map editor, and smash it. It's destroyed and leaves a hole behind, which I can then climb down.

### DIFF
--- a/.github/workflows/matrix.yml
+++ b/.github/workflows/matrix.yml
@@ -121,6 +121,7 @@ jobs:
             localize: 1
             native: linux64
             sanitize: address
+            cxxflags: --gcc-toolchain=/opt/mock-gcc-11
             mods: --mods=magiclysm
             dont_skip_data_only_changes: 1
             title: Clang 12, Ubuntu, Tiles, ASan
@@ -211,6 +212,7 @@ jobs:
         MODS: ${{ matrix.mods }}
         SANITIZE: ${{ matrix.sanitize }}
         TEST_STAGE: ${{ matrix.test-stage }}
+        CXXFLAGS: ${{ matrix.cxxflags }}
         LDFLAGS: ${{ matrix.ldflags }}
         EXTRA_TEST_OPTS: --error-format=github-action
         NUM_TEST_JOBS: ${{ matrix.num_test_jobs }}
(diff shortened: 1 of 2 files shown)


PatrikLundell:
My assumption is that "bash_below" is intended to allow the skylights to be destroyed from below, so I would suggest that you test that they can be before the change and still can be after it (I would guess explosives ought to be able to achieve that). Otherwise there is a risk the change may make them destructible only from above.
Thus, there is a risk there's a need for either a change of the meaning of "bash_below" to add to rather than replace from above destruction, provided it makes sense for all usages of the flag, or the introduction of a "bash_both" flag.

gettingusedto:
I have tested it with explosion and it breaks the skylight.

My procedure:
1. I made my change on my build
2. Find a subway station, which have skylight on its roof
3. Spawn in a c-4
4. Arm the c-4 under the roof
5. Explosion
6. Skylight destroyed

Skylight before the explosion
Screenshot_20221127-164221

Skylight after the explosion
Screenshot_20221127-164405

MNG-cataclysm (approved):

fairyarmadillo:
> My assumption is that "bash_below" is intended to allow the skylights to be destroyed from below, so I would suggest that you test that they can be before the change and still can be after it (I would guess explosives ought to be able to achieve that). Otherwise there is a risk the change may make them destructible only from above. Thus, there is a risk there's a need for either a change of the meaning of "bash_below" to add to rather than replace from above destruction, provided it makes sense for all usages of the flag, or the introduction of a "bash_both" flag.

Skylights are all the way up in the ceiling, how are you bashing them from below, exactly?

PR #61909, opened 2022-10-25
Spawn Point Zones for scenarios and Aftershock bugfixing

John-Candlebury:
#### Summary
Features "Mappers can now define player spawn point zones in mapgen."

#### Purpose of change

Allow maps to define a specific player spawn point through zones, For when scenario maps require such exactitude.  Implemented to fix: #54157 

#### Describe the solution

Implemented through zones, maybe there's a better solution.

#### Testing

Test by spawning within the aftershock escape pod.

#### Additional context

You can define start point zones from the zone manager. I dont know how to fix this.

PR #62420, opened 2022-11-27
Mansion escape update

Termineitor244:
#### Summary
Content "Improves the Mansion Escape scenario with a new mansion variant and better start locations"

#### Purpose of change
Many people complain about the difficulty of the mansion escape scenario and how they can't survive while playing it... I had wanted to make improvements to the scenario since some months ago, so now that I have time I thought to improve the experience of playing with the scenario. 

#### Describe the solution
Adds a new variant mansion, an enclosed mansion, for now it only differs from the normal mansions in having more enemies and being totally enclosed from outside, with boarded up windows and metal locked doors in the entrance, but I would like to add secret rooms and more in the future.

Changes the start location of the scenario from "Somewhere in the basement" to the panic room in the ground floor or the sex lair underground, taking good use of the Spawn Point Zones implemented in #61909, this new starting locations allow for an intelligent use of the darkness and poor visibility of your enemies to attack, agitate the enemy and return to your safe (and hidden) base to heal your wounds or prepare for your next raid in the mansion. The spawn rate of enemies inside the safe rooms was reduced but not eliminated, so you will probably have to fight as soon as you start your game to secure your safe space, if you choose to start in the panic room of the ground floor you can make use of the supplies in there from the start but you will probably have to fight for them with the old inhabitants of the mansion... While in the sex dungeon you can make use of... Kinky clothing? But your enemies will most likely be weak scantly dressed zombies. Make your choice.

I also adjusted some spawns of these locations and changed the points given by the scenario, many professions were added as possibles starts to reflect every kind of horror movie idea you have in mind! Football players, cheerleaders, pizza delivery guys, burglars and even ninjas! We have it all for your crazy imagination.

To only make the change to the starting location inside the normal mansions, but I think a boarded up mansion is way more interesting to escape from.

#### Testing
Started the game, played a little, debug spawned the mansion... Everything seems to be working ok.

#### Additional context
I want to add a story driven EOC to this scenario in the future, something like an armored feral and friends starting to appear in your way of raiding the mansion whenever you leave your safe zones, the more time you spend outside the safe rooms, the more closer and more commonly your encounters until you finally escape the mansion (Or die! Its your choice, not mine).

New start in the safe rooms:
imagen
imagen

You can't escape:
imagen

Will you be able to find a way to leave the mansion before its too late?


data/json/scenarios.json
```diff
@@ -750,30 +750,52 @@
...
     "type": "scenario",
     "id": "Mansion",
-    "name": "Mansion Escape",
-    "points": -1,
-    "description": "You were living the high life when the world ended, surrounded by riches and extravagant food every day, but now you find yourself awakening from an induced paralysis and carried to the basement of the mansion in which you used to live, heavy footsteps resound around you, are you ready to survive in this new world?",
-    "allowed_locs": [ "sloc_mansion_basement" ],
-    "start_name": "Mansion",
+    "name": "Challenge - Mansion Escape",
+    "points": -5,
+    "description": "You were living the high life when the world ended, surrounded by riches and extravagant food every day.  One day you were drugged and brought to a hidden room for who knows what purpose, you lost consciousness and now heavy footsteps nearby awake you, are you ready to escape from what confined you here?",
```
MNG-cataclysm, line 755:
```suggestion
    "description": "You were living the high life when the world ended, surrounded by riches and extravagant food every day.  One day you were drugged and brought to a hidden room for who knows what purpose. You lost consciousness and now heavy footsteps nearby awake you. Are you ready to escape from what confined you here?",
```

MNG-cataclysm (changes requested):
I think you should make a separate scenario for starting inside the boarded mansion, instead of replacing the existing one.

Termineitor244:
The thing with the scenario is that the secret rooms are variants of specific parts of the mansion, so I can't guarantee a spawn in the safe rooms with the normal mansions, and I don't think this type of variation requires a different scenario... At least for now (Until I add EOCs), I could add again the option (start location) to start somewhere random in a basement of the normal mansions like before, would that be alright?

MNG-cataclysm:
> The thing with the scenario is that the secret rooms are variants of specific parts of the mansion, so I can't guarantee a spawn in the safe rooms with the normal mansions, and I don't think this type of variation requires a different scenario... At least for now (Until I add EOCs), I could add again the option (start location) to start somewhere random in a basement of the normal mansions like before, would that be alright?

That's a good plan.

Termineitor244:
This PR should probably receive a string freeze label so the bot goes away? The PR is finished as it is, a second part will probably be added after seeing the reception of the current changes but I do not intend to do anything more by now...

Issue #54463, opened 2022-01-15
Portal Storm Enemies Pushing character to Extreme Exhaustion


coyo7e:
### Describe the bug

I noticed that if you get stuck bandaging or stuff and get hit a bunch of times by portal storm enemies, you quickly end up at "Exhausted" however, it appears that the tiredness keeps going down. I'm not sure how far this goes, but after sleeping for 36 hours straight with no breaks except to eat, I finally got the character back up to "Tired" after some portal enemies caught me bandaging. 

This is extraordinarily unfun and can take quite a long time IRL if your character is in an area that has a lot going on (like near the refugee camp, near active faction camp, etc), and I'm not sure if it's intended to force a player to sleep for 10-20 minutes of IRL playtime to recover from enemies which only deal exhaustion.

### Steps to reproduce

1. Get hit to "Exhausted" by portal storm enemies.
2. Get hit some more.
3. Sleep until recovered.

### Expected behavior

Expected to be able to recover from Exhaustion after 18 hours of sleep.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: f1997bf [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Graphical Overmap [Graphical_Overmap],
    Bionic Slots [cbm_slots],
    SpeedyDex [speedydex],
    Stats Through Kills [stats_through_kills],
    Stats Through Skills [StatsThroughSkills]
]

LeahLuong:
This may be related to #53644. How did you bandage?? 'E' or activation from inv?? It's also possible this type of attack simply needs a warning prompt.

coyo7e:
'a'ctivating bandage. It's unrelated to the bandaging, as far as I can tell. 

I had this happen once before with a new character that got caught walking across a great big field while searching for a vehicle, and I accidentally swung a heavy backpack and got hit a ton of times - after that the character was so exhausted it was a downward spiral which I just bore out and had to sleep it off for well over 24-36 hours (in 4 hour increments because he started with a bad infection, so I lost track of how long it took the first time).

The attacks do give a message about draining you, and with ultica at least it causes the flash on your sprite (same as when a zombie child dies, or other effects hit you which seems a new feature that's pretty helpful.)

[...]

PR #62431, opened 2022-11-27
Cap portal storm dungeon fatigue gain

Ramza13:
#### Summary
None

#### Purpose of change
A slightly different version of #62091

#### Describe the solution
This does the same thing as #62091 but uses imo slightly easier to follow syntax.  Also noticed player was gaining stamina from attacks and fixed that.
Thanks to [GuardianDll](https://github.com/GuardianDll) for the inital PR and the inspiration/numbers.

### DIFF
--- a/data/json/monster_special_attacks/spells.json
+++ b/data/json/monster_special_attacks/spells.json
@@ -503,12 +503,13 @@
           { "u_val": "fatigue" },
           "+=",
           { "arithmetic": [ { "global_val": "var", "var_name": "ps_str" }, "+", { "const": 20 } ] }
-        ]
+        ],
+        "max": 1200
       },
       {
         "arithmetic": [
           { "u_val": "stamina" },
-          "+=",
+          "-=",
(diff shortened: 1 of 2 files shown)


GuardianDll:
It is better syntax, yeah, but i still find it bad fix - it still means the player will turn back from portal dungeon **always** tired, and will **always** spend two days to rest
Is there some better solution we can implement after 0.G?

Maleclypse:
#54463 may fix this

Ramza13:
> It is better syntax, yeah, but i still find it bad fix - it still means the player will turn back from portal dungeon **always** tired, and will **always** spend two days to rest Is there some better solution we can implement after 0.G?

You are right.  Hmmm, I'm gonna do some testing and see about toning this down a little.  I'd love to replace fatigue at least partially with other debuffs instead.  Either way, portal storms in general need more play testing/balance and I aim to do it before stable.  Thanks

PR #62433, opened 2022-11-27
Bugfix - dissection drop rate

### LINKED ISSUE #61423: Prototype cyborgs and power storage cbm scarcity
### Describe the bug

It seems like you cannot succeed in retrieving power storage cbms by dissecting prototype cyborgs. I dissected more than 10 of them with a scalpel and max skills and it failed everytime. Which makes power storage cbms abnormally rare. The only one i found was by dissecting a zomborg, and i raided something like 6 cbm vaults. For the later, maybe im just unlucky, but you should be able to retrieve them from prototype cyborgs.

### Steps to reproduce

1. Dissect prototype cyborgs
2. "Failed to extract power storage cbm" everytime

### Expected behavior

You should be able to retrieve power storage cbms from prototype cyborgs.

### Screenshots

_No response_

- OS: Windows
- Game Version: b87b910 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

cathalpern:
#### Summary

Bugfixes "Dissection drop rate"

#### Purpose of change

Fixes #61423 

#### Describe the solution

Previously, the section to set roll value had a condition where the value was being set, except in cases where the action was dissection. So, the roll value was always 0 for dissection, causing items to never be dropped in some cases. Removing the check that the action is not dissection allows the roll value to get set.

I'm just doing what the smart people said to do

#### Testing

Debug spawned and killed 10 prototype cyborgs. Was able to successfully harvest a Power Storage CBM from all.

### DIFF
--- a/src/activity_handlers.cpp
+++ b/src/activity_handlers.cpp
@@ -828,8 +828,7 @@ static bool butchery_drops_harvest( item *corpse_item, const mtype &mt, Characte
             roll = std::max<int>( corpse_damage_effect( roll, entry.type, corpse_item->damage_level() ),
                                   entry.base_num.first );
         }
-        add_msg_debug(debugmode::DF_ACT_BUTCHER, "Roll  = %s",
-            roll);
+
         itype_id drop_id = itype_id::NULL_ID();
         const itype *drop = nullptr;
         if( entry.type->is_itype() ) {


RenechCDDA:
Well, trying it out now that I got back. 

This isn't going to work, because it breaks dissection groups. Here's a quick test change I made and the results:
image
(Note: Setting the item group to a count of say 100-1000 produces a seemingly completely random range within those values, regardless of your degree of success or, y'know, **roll**, so that's... not good either. )
image
(Reported success rate includes corrected debug output)
Notice we should fail every single one of these dissections but we always get a power storage CBM. Essentially this just breaks it in the opposite direction by guaranteeing we get defined single items regardless of how badly we do on the dissection.

Some other things I'd also like to address to fully fix the issue:

-There's a typo in the debug display on line 362, `19.0f` should change to `10.0f` (like the line below it). The above images already include this fix, but it's important moving forward.

-Debug output doesn't appear on corpses which define a single item for their dissection drop (e.g. broken cyborg with cyborg_harvest removed).

cathalpern:
I did think that it was odd that I was successfully harvesting with things like 95% failure chances - but thought maybe that was intentional, in that certain things would always drop based on config.

I-am-Erk:
So, this does appear to be the correct change after some play testing in the discord. What we need is to go here: https://github.com/CleverRaven/Cataclysm-DDA/blob/695709056140a0aad40689dd52d244627fe8687c/data/json/harvest_dissect.json#L686

and add something like 
```
"base_num": [ 0, 3 ],
"scale_num": [ 0.5, 2.5 ]
```
to each entry in zomborg and broken_cyborg in that file. if someone could add that to this PR and test it, I'd be much obliged... preliminary reports are that this should work.

cathalpern:
Adding scale numbers to dissection JSON and reopening

I-am-Erk:
I will try to review the JSON end tonight or tomorrow

RenechCDDA:
Please include the typo fix mentioned earlier, in case this ends up being the preferred solution.

mqrause:
A nice to have would be a test that makes sure result numbers are within expectations. Doesn't necessarily have to be part of this PR, though.

mqrause:
There's nothing wrong with leaving the debug messages in.

cathalpern:
> There's nothing wrong with leaving the debug messages in.

Eh, they were mostly just for checking the numbers as I was testing - don't want to spam the logs with stuff that wouldn't be super meaningful. If other people think the min/max and roll values would be helpful, I can make the messages a little more explicit and add them back.

mqrause:
There's a filter for debug messages and I wouldn't consider it spam in the first but. But of course there's also nothing wrong with removing them. And making them more compact and more meaningful is obviously also a good thing.

Issue #37980, opened 2020-02-12
Option to remove auto-generated number shortcuts from various item menus


WeaverThree:
# Is your feature request related to a problem? Please describe.
Movement keys are used as item hotkeys regardless of whether you have auto item letters on or off. I sometimes consume things accidentally because of this. Eventually, I'm going to accidentally drink bleach or something because of it.

It's far worse than accidentally messing up my advanced inventory screens because I forgot that numpad doesn't do cursor control there, and that has value in being that way.

# Describe the solution you'd like
An option in the interface settings that disables the generation of numeric hotkeys in the consume, eat, drink, medical, wear, wield, and disassemble menus, and any others I'm forgetting that have this feature: 

![cataclysm-tiles — 2020·02·12 — 14·44·06](https://user-images.githubusercontent.com/1582240/74384487-38bc5d00-4da6-11ea-913f-ce3aa4b44f84.png)

I think the numbers are only generated for stuff not in the inventory already? Since those would get letters under standard game options.

Also when this option is enabled I'd like the numpad to act as navigation keys for these menus, since there's then no reason for it to not.

# Describe alternatives you've considered
Some people probably want this so just getting rid of it is a bad solution.

Keeping dangerous items away from food might be doable but it depends on the space you're working with and how much effort you want to put into inventory sorting. And it doesn't prevent you from accidentally choosing the wrong thing that is edible.

Funguss:
So you've never tried to drink bleach before?! Well done. FYI you get a confirmation dialogue, so it's not too bad. However, I've frequently eaten things I really didn't want to eat, e.g. uncooked microwave dinners (therefore having just one left, which you're unable to then cook). Eating raw eggs I was leaving to hatch has happened, too, amongst other things I can't recall. It would be nice to have the option to reconfigure them to letters as well, as they are incredibly convenient when the selection decides it's going to reset.

scorpion451:
Being able to blacklist certain letters from auto-assigning in general would be nice, really- I'd leave the feature on if I could blacklist other problematic hotkeys like "e", "w", and "a".

Funguss:
I often assign 'a' to something I apply often, e.g. lockpicks, and for 'w' I set my earplugs or blindfold, depending on. 'e' is for bandages. You could set 'w' to something you can't wear but always carry, and 'a' to something you can't apply. Personally, I prefer this set up. Blacklisting would be a good QOL feature though!

PR #62436, opened 2022-11-27
inv_ui: Allow disabling numeric map item shortcuts

Kamayana:
#### Summary
Interface "Allows disabling the numeric hotkeys for map items"

#### Purpose of change

Adds a new interface option for disabling numeric hotkeys in menus that interact with map items. These hotkeys take the numpad as input, and thus can easily lead to mistakes, as mentioned in #37980.

#### Describe the solution

Added a new setting, and then added an additional conditional to inventory_ui.cpp's inventory_selector::reassign_custom_invlets() that will cancel adding the shortcuts if that setting is false.

Adjusting inventory_ui to respect the key code setting, which would allow it to distinguish numpad keys from numeric keys. Alternately, allowing the alphabetical hotkeys to spill over to map items if there aren't enough items in the inventory to use up every letter. Both of these options seemed further beyond my scope and I wanted to keep it simple.

#### Testing

Started a new game as Evacuee, walked next to shelf with items, pressed 'w' to bring up wield menu. Numeric shortcuts are visible, and using the numpad wields an item. Went into options and turned "Use Numeric Item Shortcuts" to FALSE. Back in game, pressed 'w' to bring up wield menu. No numeric shortcuts are visible, and using the numpad navigates the menu. Turned the setting back to TRUE and the behavior went back to the default.

### DIFF
--- a/src/inventory_ui.cpp
+++ b/src/inventory_ui.cpp
@@ -1961,7 +1961,8 @@ void inventory_selector::prepare_layout( size_t client_width, size_t client_heig
 
 void inventory_selector::reassign_custom_invlets()
 {
-    if( ( invlet_type_ == SELECTOR_INVLET_DEFAULT || invlet_type_ == SELECTOR_INVLET_NUMERIC ) && get_option<bool>("ITEM_NUMERIC_SHORTCUTS") ) {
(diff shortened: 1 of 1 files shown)


Kamayana:
Whoops, I'm a doofus. Pardon my git rust.

PR #62443, opened 2022-11-28
[BN Port] Fix automatic sprite rotation for non-linear overmap tiles

Maleclypse:
Port fix from https://github.com/cataclysmbnteam/Cataclysm-BN/pull/1762/files

Co-Authored-By: Olanti <olanti-p@yandex.ru>

#### Summary
None

#### Purpose of change
This fix was requested by tileset makers and from Olanti it would only take this single line from his PR to fix this in DDA.

#### Describe the solution
Replaces the following line `rotation = static_cast<int>( get_dir() );`.

None

### DIFF
--- a/src/overmap.cpp
+++ b/src/overmap.cpp
@@ -979,7 +979,7 @@ void oter_t::get_rotation_and_subtile( int &rotation, int &subtile ) const
         rotation = t.rotation;
         subtile = t.subtile;
     } else if( is_rotatable() ) {
-        rotation = om_direction::get_num_ccw_rotations( get_dir() );
+        rotation = ( 4 - static_cast<int>( get_dir() ) ) % 4;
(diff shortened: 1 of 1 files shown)


src/overmap.cpp
```diff
@@ -979,7 +979,7 @@ void oter_t::get_rotation_and_subtile( int &rotation, int &subtile ) const
         rotation = t.rotation;
         subtile = t.subtile;
     } else if( is_rotatable() ) {
-        rotation = static_cast<int>( get_dir() );
+        rotation = om_direction::get_num_ccw_rotations( get_dir() );
```
Wishbringer, line 982:
```suggestion
        rotation = ( 4 - static_cast<int>( get_dir() ) ) % 4;
```

PR #62444, opened 2022-11-28
json/containers: collapse food and drink containers by default

andrei8l:
#### Summary
None

#### Purpose of change
Most containers aren't collapsed by default so (for ex.) the pickup menu is littered with liquids you can't pick or duplicated information.
<details>
<summary>Example: typical fridge</summary>

Screenshot from 2022-11-28 08-40-16
</details>

I don't think any functionality or information is lost by collapsing these containers.
<details>
<summary>Same fridge, collapsed</summary>

Screenshot from 2022-11-28 08-39-54
(keep in mind that the hidden foods still show in the `E`at menu)
</details>

I think people have gotten used to (un)collapsing containers by now so we can collapse more of them by default

#### Describe the solution
Add the `COLLAPSE_CONTENTS` flag to every container that:
1. typically holds only one item of food, drink, or powder when spawned by mapgen
2. doesn't have a `use_action`, to prevent hiding contents in the `a`ctivate menu
3. isn't explicitly player-made (like sealed stomachs)

That turns out to be most of the containers in `containers.json`

Dealing with it

#### Testing
`COLLAPSE_CONTENTS` is already weathered and well tested from MRE packages and first aid kits.

#### Additional context
N/A

### DIFF
--- a/data/json/items/containers.json
+++ b/data/json/items/containers.json
@@ -36,7 +36,7 @@
         }
       ]
     },
-    "flags": [ "BELTED", "OVERSIZE", "WATER_FRIENDLY" ]
+    "flags": [ "BELTED", "OVERSIZE", "WATER_FRIENDLY", "COLLAPSE_CONTENTS" ]
   },
   {
     "id": "30gal_barrel",
@@ -63,7 +63,8 @@
         "max_contains_weight": "255 kg"
       }
     ],
-    "qualities": [ [ "CONTAIN", 1 ] ]
+    "qualities": [ [ "CONTAIN", 1 ] ],
+    "flags": [ "COLLAPSE_CONTENTS" ]
   },
   {
     "id": "30gal_drum",
@@ -90,7 +91,8 @@
         "max_contains_weight": "250 kg"
       }
     ],
-    "qualities": [ [ "BOIL", 2 ], [ "CONTAIN", 1 ] ]
+    "qualities": [ [ "BOIL", 2 ], [ "CONTAIN", 1 ] ],
+    "flags": [ "COLLAPSE_CONTENTS" ]
   },
   {
     "id": "55gal_drum",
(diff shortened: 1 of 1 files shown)


PatrikLundell:
The problem I see with this is that the pickup logic sometimes stuff unrelated items into containers along the original contents, and I definitely want to see that it has stuffed e.g. some marbles in along with the flour, for instance, since that has to be resolved manually or loot sorting will get screwed up.
I'm definitely in favor of collapsing single item type contents though.

flybye3o6:
ideally id like to have all sealed containers have collapse contents/hidden by default but then have it automatically undone when it is no longer sealed, items also will not be put into sealed containers even though the game will still say that volume is free

andrei8l:
> ideally id like to have all sealed containers have collapse contents/hidden by default 

I can do this, but it doesn't have any benefit over the flag

> but then have it automatically undone when it is no longer sealed, 

Not sure about this, but it's also trivial.

<details>
<summary>At the moment I am not convinced that it would be the better approach, but you can take this patch and argue the case in a separate PR (make sure you test it though).</summary>

```diff
diff --git a/src/item_pocket.cpp b/src/item_pocket.cpp
index caab19c514..abcf157033 100644
--- a/src/item_pocket.cpp
+++ b/src/item_pocket.cpp
@@ -820,12 +820,14 @@ bool item_pocket::seal()
         return false;
     }
     _sealed = true;
+    settings.set_collapse( true );
     return true;
 }
 
 void item_pocket::unseal()
 {
     _sealed = false;
+    settings.set_collapse( false );
 }
 
 bool item_pocket::sealed() const
```
</details>

> items also will not be put into sealed containers even though the game will still say that volume is free

This should be a thing already - there's a test for it but maybe some old code uses a different path. Open a bug report please if it's a problem.

flybye3o6:
> This should be a thing already - there's a test for it but maybe some old code uses a different path. Open a bug report please if it's a problem.

its not a bug, i was referring to the pickup logic, you can still force things into a sealed container but it breaks the seal, but when looting items the game will not put stuff into sealed containers just say "some items were not picked up" and maybe it would be best not to have COLLAPSE_CONTENTS on ridged containers that always gets things put in them then looting ex: the 3L jar, it would make uploading more of a chore
and cool ill try to test that bit of code

PR #62452, opened 2022-11-28
Add units::temperature_delta

### LINKED ISSUE #58820: Making units::temperature safer

The new units::temperature is nice. But it is too easy to do wrong math with it.

For example:
units::temperature temperature_delta = units::from_celcius( 20 )
units::temperature temperature = units::from_celcius( 15 )

One could without thinking do `temperature_delta * 2` and expect it to result in 40 °C but it actually results in 313.15 °C.
One could without thinking do `temperature + temperature_delta ` and expect it to result in 35 °C but it actually results in 307.85 °C.

The main problem is that the three different temperature scales (kelvin, celcius and fahrenheit) have their zero in different places. So 20 C + 15 C is different from 293.15 K + 288.15 K even though the temperatures in both are same. Operations on units::temperature always operate with kelvins.

### Solution you would like.

* Do not allow multiplying units::temperature with scalars.
* Do not allow adding units::temperature with units::temperature.

These two are technically ok operations to do. But it is too easy to make errors when doing so. So it is best to not even allow trying to do them.

Create new units::temperature_delta.

* units::temperature_delta could be summed with itself or multiplied with scalar. 
* It could be summed with units::temperature and that would result in units::temperature.
* Units::temperature - units::temperature would result in units::temperature_delta.

Now:

units::temperature_delta temperature_delta = units::from_celcius( 20 )
units::temperature temperature = units::from_celcius( 15 )

Now `temperature_delta * 2` results in 40 °C.
Now `temperature + temperature_delta ` results in 35 °C.

Be very very careful when writing temperature math.

### Additional context

_No response_

Hirmuolio:
#### Summary
Infrastructure "Add units::temperature_delta"

#### Purpose of change

Closes: #58820
Makes it harder to write wrong math with temperature differences.

#### Describe the solution

Add `units::temperature_delta` and use it for places where temperature differences are handled.

`units::temperature - units::temperature` results in `units::temperature_delta`
`units::temperature + units::temperature_delta` results in `units::temperature`
`units::temperature_delta + units::temperature` results in `units::temperature`
`+=` works for `temperature += temperature_delta` (but not the other way around as that feels weird)

It is no longer possible to add temperatures directly together. It is very easy to do wrong math that way. Use tempereatur_delta for that instead.

Purge all uses of celsius and fahrenheit and only use kelvins in code. The problems of mixing temperature and temperature_delta onlny happen with celsius/fahrenheit scales. Absolute (like kelvin) scale has no problems with deltas and does not need this system.

#### Testing

Tests pass. There should be no changes in behavior.

#### Additional context

* Mostly copied what #62155 did.
* Multiplying `units::temperature` with scalar is also pretty risky operation and maybe should not be allowed.


Zireael07:
I think all the other temperature units related PRs went in but this one was forgotten for some reason?

PR #61486, opened 2022-10-05
Exoskeleton update

Drew4484:
#### Summary
Content "Additional updates to exoskeletons"

#### Purpose of change
Bring exoskeletons in line with detailed armor coverage (weakpoints at joints/eyes, ablative plating) as well as some additional QOL features. 
Replaces existing combat exoskeletons with Mark 1, 2, and 3 frames. Frames will have swappable armor plating so when the issued composite plates break the player can replace them with forged steel or makeshift scraps if they're desperate. Removes separate helmet item, all frames will have integrated climate control and toggleable NVG/IR vision, clock, weather, glare protection, hearing protection, radio.

Weight capacity is designed for combat load (23.18 kg) consisting of:
|Weight|Count|Item
|---: |---: |---: |
|6.9 kg|1|M3E1 Recoilless Rifle
|16.5 kg|5|84x246mm HEDP rocket
|16 kg|5|84x246mm HE rocket
|3.5 kg|1|M27 assault rifle
|2.4 kg|10|STANAG 60-round magazine
|12 kg|600|7.62x51mm M80
|0.4 kg|6|Fragmentation grenade
|2.88 kg|2|IFAK

Plus the weight of the stock composite plating, 36.85 kg.
|Weight|Volume|Count|Item
|---: |---: |---: |---: |
|2.75 kg|4 L|1|Composite Helmet
|17.5 kg|25.5 L|1|Composite Torso
|4.15 kg|4.5 L|2|Composite Arm
|4.15 kg|5.5 L|2|Composite Leg
|3.9 kg|4 L|1|Steel/Makeshift Helmet
|24 kg|25.5 L|1|Steel/Makeshift Torso
|6 kg|4.5 L|2|Steel/Makeshift Arm
|6 kg|4.5 L|2|Steel/Makeshift Leg

| Model | Efficiency | Capacity | Power Supply | Enc/Active | Dry Weight | Notes
|---: |---: |---: |---: |---: |---: |---: 
| Mark 1 |  80% | 150 kg | Fixed Diesel |  80/15 | 90 kg | +5 str, +5 enc to hands
| Mark 2 | 100% |  135 kg | Modular |  60/10 | 75 kg | +5 str
| Mark 3 | 120% |  120 kg | Modular |  40/0 | 60 kg | +5 str, advanced HUD blocks vision while suit is offline
| ICE Utility | 100% | 215 kg | Fixed Diesel | 20/20 | 110 kg | +10 str
[...]

PR #62456, opened 2022-11-29
Aftershock Exosuit Module Fixes

Drew4484:
#### Summary
Bugfixes "Minor corrections to exosuit modules"

#### Purpose of change
Corrects an issue preventing worn equipment from providing intended effect (load support module not providing carry capacity bonus as an example).

#### Describe the solution
Apply c++ changes from @andrei8l. Also some small json corrections.

None.

#### Testing
WIP

#### Additional context
None

### DIFF
--- a/data/mods/Aftershock/items/armor/exosuit/exosuit_mutations.json
+++ b/data/mods/Aftershock/items/armor/exosuit/exosuit_mutations.json
@@ -8,6 +8,6 @@
     "valid": false,
     "purifiable": false,
     "types": [ "Equipment" ],
-    "flags": "wall_cling"
+    "flags": "WALL_CLING"
   }
 ]
--- a/data/mods/Aftershock/items/armor/exosuit/exosuit_ranged.json
+++ b/data/mods/Aftershock/items/armor/exosuit/exosuit_ranged.json
@@ -47,7 +47,7 @@
     "category": "exosuit",
     "name": { "str": "exosuit microwave cannon" },
     "description": "A 'less-lethal' weapon mounted on the shoulder of an exosuit.  Causes a painful burning sensation as water molecules are turned to high temperature vapor.  Infamous for it's use by UICA forces used it to break up protests, the effectiveness can be hard to predict due to variables like the moisture in the air and protective layers on the target.",
-    "weight": "7500 kg",
+    "weight": "7500 g",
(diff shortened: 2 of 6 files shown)


Raikiri:
Is there an issue tracking exosuits roadmap progress and plans? I'm very interested to know what's in the works, what's planned for aftershock and what's going into CDDA.

Drew4484:
> Is there an issue tracking exosuits roadmap progress and plans? I'm very interested to know what's in the works, what's planned for aftershock and what's going into CDDA.

https://github.com/users/Drew4484/projects/1

Big issues remaining are UI/UX and balance testing.

Raikiri:
Very interesting, I hope you will have enough enthusiasm to pull it through. Currently power armors are in a really sad state: they're extremely OP and extremely boring because you can't combine them with anything else. Basically once you find a power armor set, there's no reason to be looking for anything else and they are so strong that doing it would not be interesting anyway.

I really hope that power armors will become a longer-term commitment for a player, ideally similar to building a vehicle. A good comparison would be fallout 4 power armors that you find, build and assemble piece by piece and only use for certain particularly dangerous situations.

What are the plans on including them to the base game vs aftershock? I'm really afraid that if all this work goes into making an aftershock-restricted content, then 90% of players would never get the chance to interact with it.

Drew4484:
> Very interesting, I hope you will have enough enthusiasm to pull it through. Currently power armors are in a really sad state: they're extremely OP and extremely boring because you can't combine them with anything else. Basically once you find a power armor set, there's no reason to be looking for anything else and they are so strong that doing so would not be interesting anyway.
> 
> I really hope that power armors will become a longer-term commitment for a player, ideally similar to building a vehicle. A good comparison would be fallout 4 power armors that you find, build and assemble piece by piece and only use for certain particularly dangerous situations. Having a rack on your vehicle dedicated for power armor maintenance would be amazing.
> 
> It would also be great to have multiple semi-deterministic locations for individual armor pieces rather than spawning the whole set in a TCL. I'd love to raid multiple dangerous POI's hunting for different power armor parts.
> 
> What are the plans on including them to the base game vs aftershock? I'm really afraid that if all this work goes into making an aftershock-restricted content, then 90% of players would never get the chance to interact with it.

Take a look at #61486 for the plans on core game exoskeletons.

PR #62457, opened 2022-11-29
Move place zombie mom to mission JSON

cathalpern:
#### Summary
None

#### Purpose of change
In the interest of achieving #3951, this removes the place_zombie_method (and its add_spawn call) and uses the mission templates to place a zombie as the target of the Kill Zombie Mom quest

#### Describe the solution
Use the now built in JSON-ification of generating mission based objects and monsters instead of the hard coded C++ implementation

Update the hard coded add_spawns in place_zombie_mom to use place_spawns - but that seemed silly

#### Testing
Assigned start NPC the Kill Zombie Mom quest, went to house, killed zombie mom, verified that NPC was indeed grateful


MNG-cataclysm (approved):

Night-Pryanik:
The only difference is that right now through json we can't give this zombie a random female name. Might need to add this functionality some time in the future.

cathalpern:
> The only difference is that right now through json we can't give this zombie a random female name. Might need to add this functionality some time in the future.

I guess I could hard code one for now, to at least make it easier to tell what zombie you're supposed to be targeting?

Night-Pryanik:
Yeah, as a temporary solution this will suffice.

MNG-cataclysm (approved):

Night-Pryanik:
@cathalpern could you please resolve conflicts?

bombasticSlacks:
need to appease clang and remove the unused var

PR #62475, opened 2022-11-29
Fix: connect extension cables between vehicles and appliances

### LINKED ISSUE #59169: Extension cord cannot be plugged into vehicle

dseguin:
#### Summary
None

#### Purpose of change
- Fixes #59169

#### Describe the solution
Incorporate only the fix portion of #60851.

The extra power conversion stuff from that PR can be implemented for 0.H

### DIFF
--- a/src/iuse.cpp
+++ b/src/iuse.cpp
@@ -8899,7 +8899,7 @@ cata::optional<int> iuse::cord_attach( Character *p, item *it, bool, const tripo
         }
         const tripoint posp = *posp_;
         const optional_vpart_position vp = here.veh_at( posp );
-        if( !vp || !vp->vehicle().has_tag( "APPLIANCE" ) ) {
+        if( !vp ) {
             p->add_msg_if_player( _( "There's no appliance here." ) );
             return cata::nullopt;
         } else {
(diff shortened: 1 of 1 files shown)


Raikiri:
excuse me, I have tried connecting two vehicles with an extension cord, it gave me the message "You link up the electric systems of <vehicle0> and <vehicle1>", but I don't see the systems as actually linked and I also can't now remove my cable either? It has just straight up disappeared. I tried the same thing with another cable and the same thing happened.

dseguin:
Are you sure you're running an up-to-date build? I just tested it with two cars and two welding carts and it works as expected. I even tried it across overmap boundaries just to check and it's balancing power properly.

Can you create a new issue and attach your save?

Raikiri:
Maybe i'm just using it wrong? I go "a"ctivate->extension cord, select a vehicle with the engine turned on, then I hit "a"ctivate again, the extension cord is now highlighted yellow and says that one of its ends is attached, then I select the other vehicle (which is one tile apart), it says that their electric systems are now connected and the cable disappears from my inventory.

But I don't see any indication of the battery in one vehicle being charged by the running engine (with an alternator) in the other one. I also don't see a way to remove the cable. How am I expected to detach the cable?

I am running experimental from about 5 days ago.

PR #62486, opened 2022-11-30
Grammar maintenance for some of the newer scenarios

Ilysen:
#### Summary
Content "Grammar tweaks for various scenarios"

#### Purpose of change

Writing consistency is always a thing that needs to happen; this PR's just one bit of progress. I'm attaching a bit of rationale for each of the individual changes in the next section:

#### Describe the solution

* (un)living History - renamed to `Unliving History`, description completely rewritten, now `CHALLENGE` and `LONE_START`

The original description for this scenario was really simple and broke canon. The Cataclysm was obviously a lot more drawn out and high-visibility than `a scream and gunshot noises`, and I felt like bastion forts in particular offered a neat medium for telling a more nuanced story, even a minor one. In light of that, I made it a lone start scenario, and because of the comparative difficulty of starting inside the fort, I made sure that `Play Now!` wouldn't select it as a possible option.

* Mansion Escape - description updated

The existing desc. for Mansion Escape is one giant run-on sentence, and because of that it flows really weirdly. I split it up into several different sentences and adjusted the grammar to make those sentences flow together.

* CVN Sailor - `...more important things to do` changed to `...more important things to worry about`

It wasn't actually obvious what the "important things to do" in this case were other than "escape the carrier". This is a really minor tweak, but I feel like it helps establish the actual intent of the scenario.

* Nuclear Power Plant - renamed to `Emergency Shutdown`, description updated

This one was also kind of barebones, and it seems like a portion of it was copied from the Safe Place scenario. I made the description more urgent, with the intent of hinting at the radiation hazard and the threat inside and around the power plant. The name change is referencing the reactor itself, since the scenario stuck out in that it lacked a thematic name and instead just described the place you started at.

N/A, really, although I'm open to tweaking any of these changes. Writing's always subjective. :P

#### Testing

I opened up the game and made sure that all of the descriptions and names were correct. I also started a new save in Unliving History to make sure that I no longer started with a static NPC.

#### Additional context

N/A

### DIFF
--- a/data/json/scenarios.json
+++ b/data/json/scenarios.json
@@ -827,7 +827,7 @@
   {
     "type": "scenario",
     "id": "nuclear_plant_start",
-    "name": "Scrammed",
+    "name": "Emergency Shutdown",
     "points": 1,
     "description": "When the Cataclysm happened, you were working in a nuclear power plant.  Now you're trapped inside, surrounded by the shambling bodies of what used to be your coworkers.  The reactor's shut down, but you don't know if there's been a radiation leak and you need to get away from here fast before you find out the hard way.",
     "start_name": "Nuclear Power Plant",
--- a/data/json/scenarios.json
+++ b/data/json/scenarios.json
@@ -226,12 +226,13 @@
   },
   {
     "type": "scenario",
-    "name": "(un)living History",
-    "description": "After doing historical reenactment, you and your friend were taking a break inside the fort, and then you heard a scream and gunshot noises.",
+    "name": "Unliving History",
+    "description": "New England is dotted with old forts from the Revolutionary War, and many of them are still standing.  When things started getting really bad, it seemed like a natural place to hole up for a while, and it turns out that many, many other people had the same idea.  The peace didn't last long.  Now everyone else is undead, and you're trapped inside the walls with them.  The single entrance being a tight choke point is looking a lot less useful now…",
     "id": "BastionFortStart",
     "points": -2,
     "start_name": "Bastion Fort",
-    "allowed_locs": [ "sloc_fort" ]
+    "allowed_locs": [ "sloc_fort" ],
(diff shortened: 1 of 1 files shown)


8street:
> Scrammed

I request `Nuclear Scrammed` or something like that. Your version does not give a clear understanding of the scenario. And also this term is difficult to translate into another languages for a non-specialists.

Ilysen:
"Nuclear Scrammed" itself is more unclear, since "scram" is a verb here in the states. I'll probably change it to "Meltdown" or something similar in the morning, as I'm in bed rn

8street:
No no, meltdown is definitely not the right term. How do you like these options: `Emergency Shutdown`, `Reactor Trip` (seems to be more commonly mentioned at pressurized water reactors, which we have there) or `Emergency Protection`?

Or just leave Scrammed unless someone suggest a better term.

Termineitor244:
I changed somewhat the description for mansion escape in #62420... Sorry for my weird syntax when writing the description...

Ilysen:
I'm not blamin ya. writing is a discipline in its own right, and I don't judge folks for not being able to stick the landing perfectly on the first try. it's fiiiine

JonathanLochridge:
This seems cool. I like the text changes overall. I am not sure why one of the tests failed though.

Removing the bastion fort npc makes it a bit harder. Since if you can get them to follow you they make getting out a lot easier. As managing to outmaneuver and outrun them without running out of stamina can be a pretty difficult.

I kind of felt that 3 points would have been too much though. Although, maybe if the amount of zombies was increased by a little more that might justify making it a 3 point challenge?

I like the new lore for it a lot though. Pointing out a bunch of other people having the idea to hide there causing the zombie horde is cool. And also a little inspirational.
Maybe adding some more signs of people having stayed there awhile might fit then?
Like maybe some tents or something?

Issue #51172, opened 2021-08-28
Disassembling a crafted Handcrank Charger should not give you Superglue back


estebandellasilva:
### Describe the bug

Crafted a Handcrank Charger - to learn the recipe and to level up electronics
Afterwards i wanted to disassmble it.
I even got the superglue back that was used in the creation

Gonna Check if this behaviour iss reproduceable with other crafting recipes

### Steps To Reproduce

1. Create a Player and get all the ressources needed to craft a Handcrank Charger
2. Craft a Handcrank Charger
3. Disassemble Charger

### Expected behavior

Once you craft something with superglue - you should not be able to get it back from disassembly

### Versions and configuration

Build 2021-08-27-0643

actual-nh:
I suggest a "disassembly results blacklist" - a list of things that you shouldn't get back from an automatically-derived disassembly recipe.

Night-Pryanik:
Seems like this is only happening with crafted items.

![изображение](https://user-images.githubusercontent.com/11132525/152647315-58108037-4e6c-4c41-a9fe-b71eed7a047c.png)

The first hand-crank charger (weapon held) was crafted, and I indeed can get a superglue from disassembling it.
The second charger (in yellow rectangle) is debug-spawned, and there is no superglue.

LeahLuong:
@esotericist You added this item to the game in #30378. Do you know why there is a discrepancy in materials betw/ the debug & crafted versions of the item??

Zireael07:
It's not a disrepancy in materials - its the fact that generally, uncrafting something returns what you made it out of.

LeahLuong:
@Zireael07 Night-Pryanik's screenshot above shows superglue missing from the disassembly yield in the debug version. How is that not a discrepancy??

kevingranade:
Recipe authors are not generally the right person to ask about this kind of systemic issue, and please don't ping people to help you troubleshoot issues.

Night-Pryanik:
https://github.com/CleverRaven/Cataclysm-DDA/blob/7ba05c95d6819fe4589777b97e1ff8dee3214695/src/requirements.cpp#L1226-L1234

The code to remove unrecoverable components (which have `UNRECOVERABLE` flag) while disassembling stuff, such as superglue, seems to be working right, but only for non-crafted versions of items.

We only need to find out what's the difference between crafted and non-crafted versions that cause that code snippet to fail.

estebandellasilva:
[...]

PR #62498, opened 2022-11-30
Refactor item components to be stored as a map

mqrause:
#### Summary
Infrastructure "Refactor item components to be stored as a map"

#### Purpose of change

Assigning components to crafting results from a list is pretty sketchy, when it should get split up to multiple result items. I need it to be less sketchy for #60885.

#### Describe the solution

Turn components into a map and make a wrapper class around it for some easier handling. Then adjust all the code to work again. Also replace a hacky solution in the invlet test with a proper one.

This adds a fairly prevalent error message when components can't be evenly distributed to resulting items when crafting a recipe. I do believe that needs to be fixed one way or another, but if it's too much of an annoyance, the messages can be removed and some automated way to identify such problematic recipes can be devised.

Go even further with the refactor, but this wasn't planned for the charge removal project so I don't want to spend more time on it than necessary.

#### Testing

Not much yet. Game loads old saves, can disassemble old items, can craft new items, tests pass locally.
Also indirectly tested with #60885 and #63704.

#### Additional context

### DIFF
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -1342,7 +1342,7 @@ static activity_reason_info can_do_activity_there( const activity_id &act, Chara
             if( i.is_disassemblable() ) {
                 // Are the requirements fulfilled?
                 const recipe &r = recipe_dictionary::get_uncraft( ( i.typeId() == itype_disassembly ) ?
-                                  i.components.front().typeId() : i.typeId() );
+                                  i.components.only_item().typeId() : i.typeId() );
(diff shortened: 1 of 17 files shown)


misterprimus:
Does it avoid adding consumables like welding rods and bone glue to the components list after crafting to avoid recovering items that should be unrecoverable? https://github.com/CleverRaven/Cataclysm-DDA/issues/51172

mqrause:
Not this one, but I added something to do that in #60885.

PR #62506, opened 2022-12-01
clean up bio_op_takedown

anoobindisguise:
#### Summary
Balance "Make JSONified bio_op_takedown less insane"

#### Purpose of change
The current bio_op_takedown was basically instant death unless you were in power armor. Regardless of whether or not it did damage to you, it inflicts the "downed" status effect for 3 turns and the "stunned" effect for 100 turns, which is itself a recent addition in the PR that made it JSONified. "Stunned" is a brutal status effect that works like "confusion" in roguelikes such as ADOM and DC:SS, in effect you can only do two things while stunned - move in a random direction or stand still and do nothing. Being totally helpless for an entire minute and forty seconds is very silly especially when the enemy (bio op or otherwise) can re apply the attack every 10 seconds. ~~It also was made completely undodgeable for some reason, even though it historically was dodgeable and even has dodge texts, it hits 100% of the time regardless of your dodge value.~~ This is wrong and I'm just dumb, didn't test it thoroughly enough.

#### Describe the solution
Reduce stun duration to 1 second, down from 100 seconds. Also, make it dodgeable. 

I considered making it only apply its effects apply if you took damage from the attack but decided against it.

### DIFF
--- a/data/json/monster_special_attacks/monster_attacks.json
+++ b/data/json/monster_special_attacks/monster_attacks.json
@@ -418,14 +418,15 @@
     "hitsize_min": 4,
     "attack_upper": false,
     "damage_max_instance": [ { "damage_type": "bash", "amount": 10 } ],
-    "effects_require_dmg": false,
+    "effects_require_dmg": true,
     "hit_dmg_u": "%1$s kicks your %2$s and slams you to the ground!",
     "hit_dmg_npc": "%1$s kicks <npcname> and slams them to the ground!",
     "no_dmg_msg_u": "%1$s tries to kick %2$s and knock you down, but your armor protects you!",
     "no_dmg_msg_npc": "%1$s tries to kick and knock <npcname> down, but their armor protects them!",
     "miss_msg_u": "%s tries to kick your %2$s but you dodge!",
     "miss_msg_npc": "%s tries to kick and knock down <npcname>, but they dodge!",
-    "effects": [ { "id": "downed", "duration": 3 }, { "id": "stunned", "duration": 100 } ]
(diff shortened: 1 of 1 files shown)


data/json/monster_special_attacks/monster_attacks.json
```diff
@@ -425,7 +425,8 @@
     "no_dmg_msg_npc": "%1$s tries to kick and knock <npcname> down, but their armor protects them!",
     "miss_msg_u": "%s tries to kick your %2$s but you dodge!",
     "miss_msg_npc": "%s tries to kick and knock down <npcname>, but they dodge!",
-    "effects": [ { "id": "downed", "duration": 3 }, { "id": "stunned", "duration": 100 } ]
+    "effects": [ { "id": "downed", "duration": 3 }, { "id": "stunned", "duration": 1 } ],
```
Venera3, line 428:
```suggestion
    "effects": [ { "id": "downed", "duration": 3 }, { "id": "stunned", "duration": [ 2, 3 ] } ],
```
A duration of 1 means the effect gets removed in the same turn it is applied.

data/json/monster_special_attacks/monster_attacks.json
```diff
@@ -425,7 +425,8 @@
...
+    "dodgeable": true
```
Venera3, line 429: Dodgeable (and blockable) are the defaults, this is redundant.

Raikiri:
Jesus Christ I encountered it ingame. I was so confused (pun unintended) while trying to figure out why my movements are now random and why this effect does not go away.

PR #59618, opened 2022-07-26
Add toxic waste map extra

NetSysFire:
#### Summary
Content "Add toxic waste map extra"

#### Purpose of change

Take 2 of #55211. This adds an illegal toxic waste dump where some random company dumped their trash, which includes chemicals. Rarely however, this company might be XEDRA...

image

#### Describe the solution

The map extra is basically a pool of sewage with normal or trashy rubble ontop with barrels containing mostly vile stuff scattered. There is an okay chance to get a small amount of some chemicals from the barrels. There is also a big variety of chemical spills, which are basically cosmetic only.

Furthermore, there are three "optional" features which the map extra may feature depending on how the RNG feels:

- Radioactive waste (20% chance), to make it fairer and less surprising to the player it will have a broken fence around it and plutonium slurry.
- Toxic gas vents (90% chance to have at least one of them).
- Mutant corpses (10% chance to encounter at least one), lore-stuff: lab staff dumped them there and may also conveniently forget a note with vague clues where to find labs.

#### Describe alternatives you've considered

I am open to changing the chances of some stuff. I am also not 100% sure about the chemicals because I am a noob when it comes to that.

I also thought of adding empty and/or damaged atomic batteries and stuff like atomic lamps to the radioactive part.

#### Testing

Loads correctly, manually explored the world until I found a natural spawn.

#### Additional context

I might subject the non-illegal toxic waste dump to some love, too. So it is only radioactive when plutonium and stuff is being spawned.
Also, I think `liquids` does not work with item groups which is a shame really or copy-pasting the liquids into the toxic gas vent would not be necessary, if setting fields would not erase spilled liquids which is a bug I still have to report.

PR #62515, opened 2022-12-01
Fix ferals in hazardous waste sarcophagus

### LINKED ISSUE #53591: The hazardous waste sarcophagus should not have zombies inside them
### Describe the bug

There are already only robots in the upper level because of the radiation, so why are there so many ordinary (!) zombies in the lower level?

### Steps to reproduce

1. Teleport into the lower level of a hazardous waste sarcophagus.
2. Have a horde of zombies

### Expected behavior

There are two ways of solving this:

- Adding a few very dangerous creatures down there (maybe something something slime?) because radiation + blob + portal storms do not mix. There should no humans be in there, so maybe add some slime-like blob monster? Or if a foolish technician tried to take shelter in there he died and mutated.
- Add some lore why the zombies are in there. They certainly could not have wandered in. One possibility is to add some secret military snippets saying that there are some heavily infected corpses or they ran out of space in the mass graves or something and they put the corpses in there.

In any case, ordinary zombies do not belong there.

### Screenshots

_No response_

- experimental

### Additional context

On a related note, the hazardous waste sarcophagus currently looks like solid earth, which might need to get fixed.

NetSysFire:
#### Summary
Bugfixes "No more ferals trapped in chernobyl"

#### Purpose of change

Fixes #53591. I know that my complaint of lore is the main point there but I am content with just closing it.

#### Describe the solution

Seperate monster group `GROUP_VANILLA_NO_FERAL`. Can potentially be used in more than just the hazardous waste sarcophagus. Also normalized `GROUP_VANILLA` to 1000.

#### Testing

image

Teleported into the hazardous waste sarcophagus: No more ferals.

Teleported into the streets of a city: Theres still ferals and zombies.

### DIFF
--- a/data/json/mapgen/hazardous_waste_sarcophagus.json
+++ b/data/json/mapgen/hazardous_waste_sarcophagus.json
@@ -263,7 +263,8 @@
         { "square": "radiation", "amount": [ 10, 30 ], "x": 24, "y": 24, "x2": 47, "y2": 47 }
       ],
       "monster": { " ": { "monster": "mon_zombie", "chance": 10 } },
-      "monsters": { "~": { "monster": "GROUP_VANILLA", "chance": 20, "density": 0.1 } },
+      "//": "The government would not put live humans into a highly irradiated cavern with lots of maneating zombies.  Right?  Right?!",
+      "monsters": { "~": { "monster": "GROUP_VANILLA_NO_FERAL", "chance": 20, "density": 0.1 } },
       "computers": {
         "5": {
           "name": "SRCF Security Terminal",
--- a/data/json/monstergroups/zombies.json
+++ b/data/json/monstergroups/zombies.json
@@ -177,7 +177,7 @@
   },
   {
     "type": "monstergroup",
-    "name": "GROUP_VANILLA",
+    "name": "GROUP_VANILLA_NO_FERAL",
     "default": "mon_zombie",
     "monsters": [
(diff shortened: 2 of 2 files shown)


kevingranade:
I agree special radiation zonbies would be better, but this is definitely an improvement.

NetSysFire:
I had a brief discussion with kevin on IRC. Hazardous waste sarcophaguses (or sarcophagi?) are actually very rare in the world. Probably a handful, if not less. The one in chernobyl is the most known one, but there is also one in switzerland.
In fact, most of the stuff that came up when searching about that was from our forums, lol.

As you could imagine, I am bringing up the "not realistic" argument. However, this is salvagable. Because of XEDRA.

We officially have nuclear power plants ingame now and there is some nuclear (house hold) stuff anyways. Plus waste sarcophagi are not solely for radioactive things, also for stuff like pcb and dioxine.

I would bet XEDRA will have waste to hide. Heavily contaminated mutant bodies, alien debris from the netherium, byproducts of experiments etc etc. There are two things that could work:

1. Having a normal hazardous waste depot and hiding the stuff in a cavern beneath or something.
2. Having a supposedly normal hazardous waste depot but it is heavily secured with turrets ("so no vandals can leak the radioactive dioxine and pcb!111") and has no hidden area, so if you manage to have the elevator move down you will get a straight to the point alien mess.

This would both work with the lore I introduced in #59618, since they had to hastily depose of stuff once things got hectic. As always, feedback welcome.

PR #50439, opened 2021-08-02
Regenerating creatures don't regen while onfire

psuedolet:
#### Summary
Balance "Stop regenerating creatures from doing so while on fire"

#### Purpose of change
Introduces a way to disable health regeneration. 

#### Describe the solution
If a creature has the "onfire" effect, it doesn't heal from its regeneration type

#### Describe alternatives you've considered
I considered having the regeneration amount decreased instead of skipped, but this implementation seemed simpler. (This is my first PR for this project)

PR #62519, opened 2022-12-02
[Magiclysm] Harden Vault Protectors

Maleclypse:
#### Summary
None

#### Purpose of change
Prevents some slow but cheesy methods of dealing with the Vault Protectors.

#### Describe the solution
Regen.

None

#### Testing
They regen.


MNG-cataclysm (approved):

GuardianDll:
what is "regeneration_modifiers"? i is not described anywhere in the docs

Maleclypse:
> regeneration_modifiers

When I was looking through it looked like effects that reduce a monsters regeneration and that can be set per monster.  But I'd prefer someone more familiar with c++ to tell me that for sure before I'd feel comfortable adding to the docs.  Lemme see if I can find the PR that added them.

Maleclypse:
https://github.com/CleverRaven/Cataclysm-DDA/pull/50439

PR #62532, opened 2022-12-02
fix ordering for filament_canvas

anoobindisguise:
#### Summary
none

#### Purpose of change
filament_canvas had Kevlar thread first, so stuff found in the world made of canvas would naturally disassemble into kevlar threads. This is undesirable because while Kevlar threading would work for sewing canvas, it would not be cost effective compared to just using canvas-weight threads which should be the default for such projects.

#### Describe the solution
Put thread_canvas first before thread_kevlar.

### DIFF
--- a/data/json/requirements/materials.json
+++ b/data/json/requirements/materials.json
@@ -297,7 +297,7 @@
     "id": "filament_canvas",
     "type": "requirement",
     "//": "Thread for sewing canvas.  Unlike filament_durable it's ok to use a single 10 inch canvas thread for this, but half-length aramid threads are still double.  Kept separate from filament_durable because canvas thread is doubled up for those to alleviate it being a bit weaker than kevlar threading.",
-    "components": [ [ [ "thread_canvas", 2 ], [ "thread_kevlar", 2 ], [ "thread_nomex", 1 ] ] ]
+    "components": [ [ [ "thread_canvas", 2 ], [ "thread_kevlar", 2 ], [ "thread_nomex", 2 ] ] ]
   },
   {
     "id": "glazing",


data/json/requirements/materials.json
```diff
@@ -297,7 +297,7 @@
     "id": "filament_canvas",
     "type": "requirement",
     "//": "Thread for sewing canvas.  Unlike filament_durable it's ok to use a single 10 inch canvas thread for this, but half-length aramid threads are still double.  Kept separate from filament_durable because canvas thread is doubled up for those to alleviate it being a bit weaker than kevlar threading.",
-    "components": [ [ [ "thread_kevlar", 2 ], [ "thread_nomex", 2 ], [ "thread_canvas", 1 ] ] ]
+    "components": [ [ [ "thread_canvas", 2 ], [ "thread_kevlar", 2 ], [ "thread_nomex", 1 ] ] ]
```
nopenoperson, line 300: this line should  read: 
`"components": [ [ [ "thread_canvas", 1 ], [ "thread_kevlar", 2 ], [ "thread_nomex", 2 ] ] ]`

kevlar and nomex are supposed to be the same length if you look at the comments on their entries and the way they are being used elsewhere.

nopenoperson (changes requested):

data/json/requirements/materials.json
anoobindisguise: yup 100% just a mistake from throwing this together in 30 seconds at work

nopenoperson (approved):

Maleclypse (approved):

PR #62536, opened 2022-12-02
Don't persist flatbuffers for save files.

### LINKED ISSUE #62384: Place compiled, cached JSON files in $XDG_CACHE_HOME
### Describe the bug

Currently, `$XDG_CONFIG_HOME/cataclysm-dda` contains a `cache` dir with the compiled config files.

### Steps to reproduce

1. Have a fresh installation of cdda
2. Start it.
3. Have cache in your config dir, not the cache dir.

### Expected behavior

They should be placed in `$XDG_CACHE_HOME` (defaults to ~/.cache), since this is not the actual, effective config.

Minor issue as the config files are small enough that it does not break things. Some people have the cache dir mounted as a tmpfs (in-ram storage) for extra speed and can not benefit from that.

### Screenshots

_No response_

- OS: Linux
- Game Version: 0.F-12326-gfae2b2ac65b [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

akrieger:
#### Summary
None

#### Purpose of change
Flatbuffers bloat savegame on-disk size by a nontrivial amount. A seriously large %age, even, on the order of 50%. The compressed size regresses 100%. But the load time savings from having them are trivial. So don't save them, because we don't have to.

Fixes #62384.

#### Describe the solution
Make the cache directory path argument to `flexbuffer_cache` optional. Without it, a `flexbuffer_disk_cache` instance is not created. There are only two places where we call to it - once to load from disk, once to save to disk. Both can be made optional without regressing functionality because on-the-fly parsing still works.

#### Testing
Loaded a save without my changes to generate all the flatbuffers. Then instrumented the load at 5k samples/second. Then deleted the caches inside the world (character and map caches), and reinstrumented it. The worst case difference was ~300 samples, or 60ms. This includes writing out the flatbuffers to disk.
image

Made my changes. Deleted the caches from the save. Reloaded the world. The world loaded, but the caches were not generated.

### DIFF
--- a/src/flexbuffer_cache.cpp
+++ b/src/flexbuffer_cache.cpp
@@ -372,7 +372,9 @@ class flexbuffer_disk_cache
 flexbuffer_cache::flexbuffer_cache( const fs::path &cache_directory,
                                     const fs::path &root_directory )
 {
-    disk_cache_ = flexbuffer_disk_cache::init_from_folder( cache_directory, root_directory );
+    if( !cache_directory.empty() ) {
+        disk_cache_ = flexbuffer_disk_cache::init_from_folder( cache_directory, root_directory );
+    }
 }
 
 flexbuffer_cache::~flexbuffer_cache() = default;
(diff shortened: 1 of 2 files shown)


a666:
so the cache directory can be deleted now on saves and save backups?

akrieger:
Yes, feel free.

a666:
great, thank you.

Issue #58823, opened 2022-06-29
Cauterization don't heal bitten body parts and don't stop bleeding


zakhad:
### Describe the bug

Currently attempting to cauterize a bite wound or at least used to a few months back, used to give you a 50/50 chance of cleaning the wound, in exchange for bodypart damage and pain, it was a method of saving on some disinfectant, if you were in a position to do so or if you had no disinfectant and wanted to take a chance.

After doing some equieries in discord chat with the help of Harakka, and Venera3, it was concluded that the iuse_actor.cpp doesn't actually give you any chance to clean the wound via cauterization. So this bug report is over whether or not that the UI is giving out misinformation or it's not displaying the correct information as i've attempted multiple times on a new character and it never once cleaned the infection via cauterization. 

### Steps to reproduce

Get bitten by a zombie, attempt to clean it with cauterization, with some metal object and a fire source.

### Expected behavior

Have the UI not inform the player of giving them a 50/50 chance of cleaning bite wounds, when it no longer does.

### Screenshots

![image](https://user-images.githubusercontent.com/5264667/176490334-d0f4c363-78e8-4cb4-bd48-8955beabde74.png)
![image](https://user-images.githubusercontent.com/5264667/176491235-78a9fabe-2e1f-4733-a382-3c91ecfd6e9e.png)
![image](https://user-images.githubusercontent.com/5264667/176491287-465b8a34-73bd-47c4-a2e7-4bd19a5ccb4c.png)

### Versions and configuration

- OS: Windows
    - OS Version: MINGW/CYGWIN/MSYS2 on unknown Windows version
- Game Version: cc52f60 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Craft Ammo Mags & Belt Links [SD_Magcrafting],
    Stats Through Skills [StatsThroughSkills],
    Arcana and Magic Items [Arcana],
    Mythical Martial Arts [MMA],
    Magiclysm [magiclysm],
    SpeedyDex [speedydex],
    Disable NPC Needs [no_npc_food],
    Medieval Mod Reborn [Medieval_Mod_Reborn],
    Blaze Industries [blazeindustries],
    Useful Helicopters Experimental [useful_helicopters],
    Mining Mod [Mining_Mod],
    Dorf Life [Dorf_Life],
    Aftershock [aftershock],
[...]

PR #62539, opened 2022-12-03
Remove all references to cauterizing wounds

ghost:
#### Summary
Bugfixes "Remove all references to cauterizing wounds"

#### Purpose of change

Cauterize having a beneficial effect has been broken for some time - in coordinance with reality. This merely completes removal of the feature as players are currently being misinformed and hurting themselves for no useful purpose.

#### Describe the solution

Remove all cauterize functions from existing items and all references to it in-game. The sole exception is note_180, in survivor_notes. This appears to be flavor and harms nothing by staying.

Allowing players to continue burning themselves. As the capability was only added in order to cauterize wounds, it doesn't seem worth supporting an obsolete feature just to keep self-harm.

#### Testing

Game compiles and launches. Tailoring kits, soldering irons, hotplates and knives were checked to insure they no longer allow cauterization.

### DIFF
--- a/doc/JSON_FLAGS.md
+++ b/doc/JSON_FLAGS.md
@@ -1378,7 +1378,7 @@ These flags apply to the `use_action` field, instead of the `flags` field.
 - ```SMOKEBOMB``` Pull the pin on a smoke bomb.
 - ```SOLARPACK_OFF``` Fold solar backpack array.
 - ```SOLARPACK``` Unfold solar backpack array.
-- ```SOLDER_WELD``` Solder or weld items, or cauterize wounds.
+- ```SOLDER_WELD``` Solder or weld items.
 - ```SPRAY_CAN``` Graffiti the town.
 - ```SURVIVORMAP``` Learn of local points-of-interest that can help you survive, and show roads.
 - ```TAZER``` Shock someone or something.
--- a/doc/JSON_INFO.md
+++ b/doc/JSON_INFO.md
@@ -2492,7 +2492,7 @@ Examples of various usages syntax:
 ```C++
 "usages": [ [ 1, [ "PICK_LOCK" ] ] ]
 "usages": [ [ 2, [ "LUMBER" ] ] ]
-"usages": [ [ 1, [ "salvage", "inscribe", "cauterize" ] ] ]
+"usages": [ [ 1, [ "salvage", "inscribe"] ] ]
 "usages": [ [ 2, [ "HACKSAW", "saw_barrel" ] ] ]
 "usages": [ [ 1, [ "CHOP_TREE", "CHOP_LOGS" ] ], [ 2, [ "LUMBER" ] ] ]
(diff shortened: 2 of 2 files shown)


Night-Pryanik:
Should this close #58823?

RenechCDDA:
>Cauterize having a beneficial effect has been removed for some time. 

It's been broken* for some time. Important difference.

intergalacticdata:
> 

I misunderstood the second+ hand report I received then. I'll adjust.

oosyrag:
While I don't doubt it's (lack of) efficacy, I also wouldn't doubt people would try it in a pinch due to common misconception. Why not let people burn themselves if they want to?

Kamayana:
> While I don't doubt it's (lack of) efficacy, I also wouldn't doubt people would try it in a pinch due to common misconception. Why not let people burn themselves if they want to?
> 
> Note that suicide is also a thing, that happens to be bound by default when there are definitely more useful things that could be graced with a default bind in my opinion, but I digress.

The thing is, the UI implies that it would work - it shows "minor bleeding -> 100% recovery" which is just a lie. If it said "minor bleeding -> Maybe this will work?" or something, with it becoming even more obviously worthless if you have enough survival skill, it'd be more debatable but that feels like going to too much effort just to do a bit of Mythbusters for players.

intergalacticdata:
These functions were intended for cauterization and merely had self harm shimmed in.

I think it would make more sense to create a new, more universal self harm activity that is clearly intended as such if we want to support any of these use cases.

MNG-cataclysm:
What if we left cauterization in, but gated it behind a medical proficiency? Without it, you have a very high chance of injuring yourself with little to no productive results. With it, you can reasonably close and disinfect wounds with the process, maybe with a little less pain by avoiding healthy tissue.

intergalacticdata:
> 

Would probably need a rewrite as cauterization was already broken for several years.

harakka:
> With it, you can reasonably close and disinfect wounds with the process

No, Cauterization _increases_ infection risk. Just let it go into the night, a lot of people from medical backgrounds have already chimed in on this in other discussions.

harakka:
```
#### Summary
Bugfixes "Remove all references to cauterizing wounds"
```
should help you get past the PR validator.

Fris0uman:
Documentation seems to still reference "cauterize" in JSON_INFO line 2495 and line 3888 and in JSON_FLAGS line 1381

dseguin:
Have you checked that existing saves that autosave during the cauterizing activity will continue to work?

intergalacticdata:
> Have you checked that existing saves that autosave during the cauterizing activity will continue to work?

Just checked it. Save loaded.

intergalacticdata:
> Have you checked that existing saves that autosave during the cauterizing activity will continue to work?

Just checked it. Save loaded.

PR #62543, opened 2022-12-03
Preserve ammo linkage on zone unload

### LINKED ISSUE #60310: Using the unload zone on ammo belts destroys the linkages.
### Describe the bug

If you create an unload zone on a tile within a vehicle and place an ammo belt onto it and then unload them using the zone interface, the linkages disappear but the ammunition is correctly removed from the belt and sorted.

### Steps to reproduce

1. Create a loot zone on a cargo space or other vehicle container for unloading.
2. Place an ammo belt onto the space.
3. Unload the ammunition.
4. The linkages dissapear.

### Expected behavior

The linkages should still exist.

### Screenshots

N/A

07d8e66

### Additional context

none

robob27:
#### Summary
Bugfixes "Preserve ammo belt linkages when unloading with zones"

#### Purpose of change

Fixes #60310

#### Describe the solution

As a quick fix for stable, and as suggested by @RenechCDDA, copies the code that is responsible for preserving ammo belt linkages from `unload_activity_actor::unload` (the player manually unloading)  to other code that is responsible for unloading using zones so that the linkages will also be preserved there. An ideal fix would be more DRY.

The copied code is from: https://github.com/CleverRaven/Cataclysm-DDA/blob/57c865ece207986decd6db4ccb403e14a15fad5a/src/activity_actor.cpp#L2929-L2936

I made some small modifications to spawn the linkages on the tile the unloaded item is on and not show the disassembly message so it doesn't spam them if many belts get unloaded back to back.

#### Testing
- Spawn some ammo belts
- Put them in an unload zone
- Unload them using SHIFT+O
- Check that the ammo linkages don't disappear and that the number of linkages created = the number of ammo unloaded
- The linkages should be unloaded on the tile that the ammo belt was on
- The ammo belt should be removed if it's empty

### DIFF
--- a/tests/clzones_test.cpp
+++ b/tests/clzones_test.cpp
@@ -1,27 +1,79 @@
 #include <iosfwd>
 #include <vector>
 
+#include "activity_actor_definitions.h"
 #include "cata_catch.h"
 #include "clzones.h"
 #include "item.h"
 #include "item_category.h"
 #include "item_pocket.h"
 #include "map_helpers.h"
+#include "player_helpers.h"
 #include "point.h"
 #include "ret_val.h"
 #include "type_id.h"
 
 static const faction_id faction_your_followers( "your_followers" );
 
+static const itype_id itype_556( "556" );
+static const itype_id itype_ammolink223( "ammolink223" );
+static const itype_id itype_belt223( "belt223" );
+
 static const zone_type_id zone_type_LOOT_DRINK( "LOOT_DRINK" );
 static const zone_type_id zone_type_LOOT_FOOD( "LOOT_FOOD" );
 static const zone_type_id zone_type_LOOT_PDRINK( "LOOT_PDRINK" );
 static const zone_type_id zone_type_LOOT_PFOOD( "LOOT_PFOOD" );
+static const zone_type_id zone_type_zone_unload_all( "zone_unload_all" );
+
+static int count_items_or_charges( const tripoint src, const itype_id &id )
+{
+    int n = 0;
+    map_stack items = get_map().i_at( src );
+    for( const item &it : items ) {
+        if( it.typeId() == id ) {
+            n += it.count();
+        }
+    }
+    return n;
+}
 
 static void create_tile_zone( const std::string &name, const zone_type_id &zone_type, tripoint pos )
 {
     zone_manager &zm = zone_manager::get_manager();
     zm.add( name, zone_type, faction_your_followers, false, true, pos, pos );
+}
+
+TEST_CASE( "zone unloading ammo belts", "[zones][items][ammo_belt][activities][unload]" )
+{
+    avatar &dummy = get_avatar();
+    map &here = get_map();
+
+    clear_avatar();
+    clear_map();
+
+    tripoint_abs_ms const start = here.getglobal( tripoint_east );
+    dummy.set_location( start );
+    create_tile_zone( "Unload All", zone_type_zone_unload_all, start.raw() );
+
+    item ammo_belt = item( itype_belt223, calendar::turn );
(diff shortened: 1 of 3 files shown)


robob27:
After I got the test setup somewhat properly it didn't fail a single time locally but of course it fails in CI and I can't replicate the failure. It seems like after 

```
complete_activity( dummy, unload_loot_activity_actor() );
```

the items are not always available on the map right away, so it can fail. Not sure what the best fix is for that but it's probably real simple. I gotta sleep but this gets this issue most of the way there, I think!

tests/clzones_test.cpp
```diff
@@ -1,29 +1,97 @@
...
+            return it.typeId() == itype_belt223;
+        } ).front()->ammo_remaining();
+    };
+
+    const int belt_ammo_count_before_unload = get_belt_ammo_count();
+
+    CHECK( belt_ammo_count_before_unload == ammo_belt.ammo_capacity(
+               ammo_belt.loaded_ammo().ammo_type() ) );
+
+    complete_activity( dummy, unload_loot_activity_actor() );
```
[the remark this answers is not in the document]
robob27: Thanks, wasn't aware of this helper.

tests/clzones_test.cpp
```diff
@@ -1,29 +1,97 @@
...
+
+    calendar::turn += 1_turns;
+
+    THEN( "Check that the ammo and linkages were both unloaded" ) {
+        CHECK( count_items_or_charges( tripoint_east, itype_belt223 ) == 1 );
```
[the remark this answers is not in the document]
robob27: Yeah you're right, fixing

robob27:
The test is still passing 100% of the time locally and failing 100% of the time on the "GCC 9, Curses, LTO" check (I assume my latest commit won't change this). I don't understand 🙃. Otherwise, everything seems to be working as described in the testing steps. I'll probably have to leave this as is, I'm out of ideas for now. Please feel free to ping me here or on the discord if anyone has any suggestions to fix the test, or drop it or whatever works best for stable release 🚀

tests/clzones_test.cpp
```diff
@@ -1,29 +1,79 @@
...
+TEST_CASE( "zone unloading ammo belts", "[zones][items][ammo_belt][activities][unload]" )
+{
+    avatar &dummy = get_avatar();
+    map &here = get_map();
+
+    clear_avatar();
+    clear_map();
+
+    dummy.setpos( tripoint_east );
+    create_tile_zone( "Unload All", zone_type_zone_unload_all, tripoint_east );
```
andrei8l, line 55: You need to use absolute coordinates. Something like this:
```suggestion
    tripoint_abs_ms const start = here.getglobal( tripoint_east );
    dummy.set_location( start );
    create_tile_zone( "Unload All", zone_type_zone_unload_all, start.raw() );
```
The other tests work because they don't interact with the reality bubble.

robob27: Thank you! This fixed the test failure. Is this documented somewhere or do you just happen to have this knowledge from personal experience? And do you have any idea why, even without absolute coordinates, the test was working locally? Either way, thanks again. The test not working consistently between my computer and CI was driving me mad.

andrei8l: The zone manager is global so it makes sense for it to use global coordinates. The `tripoint` type is supposed to give you this info, but the zones code (particularly `zone_manager::add()`) hasn't been fully migrated yet.

Using local coordinates can still work if the reality bubble is anchored at zero, so local coord == global coord. That's always the case if you only run one test case in isolation, though I'm not entirely sure why the LTO build with the full test suite exposes this issue more often than the non-LTO one.

robob27:
Current failure is from an unrelated test, I think this is probably good to go.

PR #62547, opened 2022-12-03
Fix AIM's container interactions

### LINKED ISSUE #59981: Moving items from a container using AIM seems to be broken
### Describe the bug

Advanced Inventory Manager system does not seem to allow manipulating items inside a container.

### Steps to reproduce

1. Spawn a new character.
2. Spawn "messenger bag", "pair of combat boots", "pair of sneakers".
3. Insert boots and sneakers into the messenger bag.
4. Drop messenger bag to the adjacent tile.
5. Now attempt to move both items from inside the messenger bag to some other AIM location like your inventory or another tile by selecting the messenger bag and using the C key.
6. Notice that AIM is not showing all of the bag contents.

### Expected behavior

Selecting a container (possibly a nested container) in AIM and pressing C should display all contents of this container.

### Screenshots

get

aim

Game version: cdda-linux-tiles-x64-2022-08-06-2052
Mod list: default

### Additional context

_No response_

Kamayana:
#### Summary
Interface "Fixes AIM's interactions with containers"

#### Purpose of change
Fixes #59981. The AIM is currently quite broken when trying to interact with container contents. Aiming at a container will only show a single item inside of it at a time, and upon moving that item out AIM will say it's now an invalid container until you switch out and back. It's also impossible to move items into a container using AIM, as you only get incongruous popups saying "You can unload only liquids into target container" or "Source must be container", even when liquids aren't involved. Lastly, you can use container aim with any item, including non-containers, which is confusing.

#### Describe the solution
The AIM container code appears to be fairly outdated and mostly intended for interaction with liquid containers. However, those liquid container functions don't even work because there's currently no way to have both AIM panels aim at containers. I first made AIM actually show all contents using the methods used for listing inventory and equipment. Next I added a check to see if the item is actually a container so you can only aim the panel into an actual container and not, say, a spoon. Next I removed some is_container_valid checks that were incorrectly flagging containers as invalid just because you emptied them or took out the arbitrary legacy_front() item. 

Lastly, I changed the moveitem code to utilize insert_item_activity_actor rather than its own liquid-container-specific specialized function. This currently means move_content() isn't used at all, though as I said earlier it's impossible to use anyway because both panels can't be containers. However, in the future I would want it to be possible to make both panels containers, which could make move_content relevant again, so I didn't want to remove it outright. I could also be mistaken and these functions could be used somewhere but I couldn't think of any way to do so.

Making it so you can only interact with containers you're wearing or in your inventory, as I believe being able to insert in and out of a container on the tile next to you is new functionality. Also, leaving the move_content() reliance intact which would make it impossible to put things into a container using AIM but would keep everything else.

#### Testing
Started a normal start, put all of the shelter's items in a 3x3 pile. Debug spawned a hiking backpack onto the floor and put things in and out of it using AIM. Spawned 200 plastic bottles, confirmed I couldn't put them all in the backpack because of the volume limit. Spawned a quarterstaff to confirm it couldn't be moved over since there's no pocket big enough. Wore the backpack and moved things in and out using AIM. Spawned a car and moved things to and from the backpack sitting in the car and out. Spawned a shopping cart and dragged it, moved things to and from it and the bag.

#### Additional context
Move all doesn't currently work, nor does moving an entire stack at once. Moving something into a container will often pop up a "inserting: XX%" progress notice at the top that gets stuck and doesn't go away, but it's possible that could be caused by a new AIM screen refresh bug that popped up recently that I haven't investigated or reported yet, so I can't say for sure if it's related or not.


andrei8l:
>  but it's possible that could be caused by a new AIM screen refresh bug that popped up recently that I haven't investigated or reported yet, so I can't say for sure if it's related or not.

Sorry for going offtopic here, but what's this? Please open a bug report soon as it's likely related to a stable blocker (#62400)

Kamayana:
> > but it's possible that could be caused by a new AIM screen refresh bug that popped up recently that I haven't investigated or reported yet, so I can't say for sure if it's related or not.
> 
> Sorry for going offtopic here, but what's this? Please open a bug report soon as it's likely related to a stable blocker (#62400)

Opened: #62551

Kamayana:
I confirmed that the changes from #62560 removes the stuck insert popup so that's one issue down.

Kamayana:
I think this is ready for review now. I'm able to move things in and out of containers, both worn and not, and all the new UI elements are working properly. It's a big revision so more testing could always be needed. There are still things I'd like to do for the AIM in the future, but my goal for this PR was to get the baseline working.

I've discovered a number of inventory bugs while putting this together. While some of them remain (#62792, #62764), they're bugs for the inventory system as a whole and not specifically for the AIM or these revisions.

Kamayana:
Two bugs I've noticed on testing this out again:
1. Bucket-style containers can be manually placed into containers erroneously (but not with move all, which correctly passes over them).
2. Doing Move All on a tile with a zombie corpse can cause the AIM to lose item targets if the zombie isn't moved last. This is because corpses are treated as special containers that list all the contents in the AIM rather than hiding them.

I'll look into fixing these. A quick fix for the corpse issue would be to make Move All ignore the corpse if it has contents - someone using Move All in the AIM is probably aiming to move the zombie's items rather than the zombie itself. An alternative quickfix would be to treat corpses the same as other containers and require the user to press 'c' while selecting them, but I foresee that causing confusion and annoyance since it'd be a change in behavior. I think the best long term solution would be to rework how containers are displayed in the AIM and make it behave more like the normal inventory menus where contents are displayed indented under the container, but that would be better suited for another PR.

andrei8l:
Since this lets you insert items into containers on the map without picking them up, it needs to update the active item cache too. Easiest way is probably to call `map::make_active( container )`.

It can be tested easily by activating a grenade and then AIMing it into a container on the map (not in inventory).

Kamayana:
> Since this lets you insert items into containers on the map without picking them up, it needs to update the active item cache too. Easiest way is probably to call `map::make_active( container )`.
> 
> It can be tested easily by activating a grenade and then AIMing it into a container on the map (not in inventory).

Good call! Added a make_active call to the insert action and it fixed it.

Kamayana:
There we go, both issues I noticed fixed.

Luca-spopo:
Can this be backported to 0.G?

Kamayana:
> Can this be backported to 0.G?

Unfortunately, while I was able to cleanly merge this PR with 0.G, I was unable to build it - I think the 0.G build scripts are too out of date for my machine. I don't want to propose a PR I cannot test myself. The game is currently pushing towards a 0.H stable so it'd probably be best to wait until that release, which hopefully won't be too far off.

PR #62563, opened 2022-12-04
run stale checker hourly

kevingranade:
#### Summary
None

I assumed the stale checker's default scheduling was hourly, it was daily, which is not going to keep up with our workload.


Zireael07:
Why? What difference does closing a couple hours later make?

I imagine running it hourly is quite a drain on resources - and CDDA already uses a lot of those for all the builds and whatnot

kevingranade:
It has a rate limit for how many operations it will perform, if we don't run it enough it will never catch up.

Issue #61682, opened 2022-10-14
Unarmed Skill Learning Issues


fairyarmadillo:
### Describe the bug

There are currently several issues with training the unarmed skill in combat.

1) If you go around fighting barehanded, unarmed skill will train extremely slowly compared to how quickly weapon skills train if you use those. 30 unarmed attacks at 0 skill against a dummy got me to 3%, while the same number with a weapon in hand got me to 90% bashing. This was with a naked survivor with 8 in all stats, debug life support and HVAC, and starting with neutral morale and 100 focus both times. In both cases, I used No Style.

2) If you are wearing armor (such as gloves) which confers bonus damage, you train a weapon skill instead of unarmed. In this test I used a pair of chitinous gauntlets and saw zero improvement in my unarmed skill while my bash gained an entire rank. This does not happen if you use a mutation, such as Claws. In that case, you gain no cut experience and a very small amount of unarmed experience.

### Steps to reproduce

1) Make a survivor
2) Go punch something
3) Notice your unarmed training very slowly
4) go hit something with a stick
5) notice your bash training relatively quickly
6) put on some chitinous gauntlets and punch something
7) notice that your bash is training quickly and your unarmed is not moving at all

### Expected behavior

Unarmed skill should learn at a similar rate to other weapon skills, and fighting unarmed while wearing gloves should train unarmed, or possibly a mix of unarmed and whatever damage type the gloves are adding.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19043.2130 (21H1)
- Game Version: 07a9641 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

anoobindisguise:
I have also consistently found that it is literally impossible to train unarmed beyond a certain point unless you are wearing specifically scrap knuckles.

Nandorianen:
Can confirm. After training all possible weapons skills to 3 through practice recipes and playing for a season, I've got dodging, bashing, cutting, handguns at 7+. Meanwhile unarmed is at 50% past 3, even though I try to use it the most.
[...]

PR #62564, opened 2022-12-04
Unarmed crafting organization

Termineitor244:
#### Summary
Interface "Creates a new subcategory for unarmed weapons crafting"

#### Purpose of change
One comment of #61682 reminded me that unarmed weapons (armor) do not have their own subcategory for ease of finding, most are in the bashing department, one is in stabbing and another one is all the way until armor/hands...

So, if you want to craft a weapon (armor) for unarmed combat, you have to search among different crafting lists and read the descriptions to try and guess which weapon is an unarmed weapon or not.

#### Describe the solution
A new crafting subcategory for weapons, "UNARMED" where you can find all kinds of knuckles, the cestus, the studded gloves and any future unarmed weapon/armor that could be added.

To make it a new subcategory of armor crafting, but these things will be searched as weapons first.

#### Testing
Started the game, opened the crafting menu, the items are under a new header; "unarmed".

#### Additional context
imagen


Dan1ss1mo:
i think adding sabatons, demi gauntlets and mittenn gauntlets here would also be reasonable, can item be in 2 categories at the same time though?

Termineitor244:
A good idea, but as you say I don't think its possible to add a recipe to 2 categories, I will try when I arrive at home, but if not then this would probably need some c++ changes to work.

bombasticSlacks (changes requested):
Can this PR be rearranged to not move the items to new files? It makes it harder to track the GIT history.

Termineitor244:
Of course! I considered a petition like this when making the PR so I just have to revert the last commit, which only created and moved the items to the new file.

I will do it when I arrive at home after work today.

AcidAntOnAMinefield:
Absolutely legendary PR. Thank you!

PR #62580, opened 2022-12-05
Nerf integrated fur armor to reasonable levels

KittyTac:
#### Summary
Balance "Severely reduce integrated fur armor values for mutants"

#### Purpose of change
Mutant fur armor was quite strong, as shown by playtesting and brought up in #60505. This was because it used the "Fur" material, which assumed that there was leather under the fur, which works fine for clothing, but in the case of mutations it essentially gives you free leather armor in addition to your fur. Pretty sure a bear's fur is not equivalent to plate armor.

#### Describe the solution
Make a new material that has most protection values divided by 3, make integrated fur armor use it. I also reduced encumbrance somewhat to match, and to make the armor not terrible.

Leaving as is, but it's unrealistic and unbalanced.

#### Testing
Before (used 3 different mutations):
1. Light fur
image
2. Gray fur
image
3. Shaggy fur
image

After:
1. Light fur
image
2. Gray fur
image
3. Shaggy fur
image

#### Additional context
I did not change the fur armor that was part of chimera patchwork skin, because there having thicker skin underneath makes sense.

### DIFF
--- a/data/json/items/armor/integrated.json
+++ b/data/json/items/armor/integrated.json
@@ -172,7 +172,7 @@
         "material": [ { "type": "mut_fur", "covered_by_mat": 100, "thickness": 2 } ],
         "covers": [ "hand_l", "hand_r", "foot_l", "foot_r", "leg_l", "leg_r", "arm_l", "arm_r", "torso", "head" ],
         "coverage": 100,
-        "encumbrance": 1
+        "encumbrance": 0
       },
       {
         "material": [ { "type": "mut_fur", "covered_by_mat": 100, "thickness": 1 } ],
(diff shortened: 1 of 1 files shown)


KittyTac:
I will also nerf the encumbrance somewhat.

fairyarmadillo:
I think a better fix might be to look at the protection values on fur and faux fur. Fur clothing does not have thick tanned leather underneath it, it has a very thin layer of treated hide that would not provide much more protection than the hair itself.

bean-b:
> I think a better fix might be to look at the protection values on fur and faux fur. Fur clothing does not have thick tanned leather underneath it, it has a very thin layer of treated hide that would not provide much more protection than the hair itself.

Yea i agree fur's protection values seem a little off. IMO all fabrics/leathers/furs need an audit of their protection values but of course this is out of the scope of this pr.

KittyTac:
> I do agree they need a nerf, but i wonder if this is a bit to much? For example from a gameplay perspective, Shaggy fur now locks you out of both close to skin and normal layers, and is super warm ( which is generally a bad thing) and has 10 encumbrance. all for between 7, 4, 3.5 protection values. By the time you have access to mutagens, even mid-tier armor will be providing a much better protection/encumbrance ratio, in addition to potential additionally effects, i.e. activity suit. From a lore perspective given that cotton is 1,1,1 protection mutant fur having between 1/2 to 1/4 of that seems low especialy sense i might imagine your skin would harden to some sort of leather like material if you have shaggy bear fur

Hmm... Fair. I will increase the protection again a bit and further reduce encumbrance.

PR #62583, opened 2022-12-05
Separate global spawn rate for each item category

Night-Pryanik:
#### Summary
Features "Separate global spawn rate for each item category"

#### Purpose of change
Allow for easier and more precise control over item spawning. One can change spawn rate for, let's say, guns, (while leaving spawn rate for all other item categories unaffected) with simple changing of one json value.

#### Describe the solution
- Added `item_category_spawn_rate` function which calculates global spawn rate based on item category, which checks for new `spawn_rate` parameter in `ITEM_CATEGORY` object.
- If not set, the value of this parameter is 1.0.
- Setting 0.0 forbids spawning items from this item category altogether.
- Values less than than 0 are set to 0.
- Added check for aforementioned new function to `map::place_items`. 
- If `item_category_spawn_rate` is less than or equal to 1.0, roll a dice to determine if item should spawn at all.
- If `item_category_spawn_rate` is more than 1.0, spawn at least one item. Then create a list of items from item group which original item was located in and remove items with different item categories. Thus, if original item was from `guns` category, the list will contain only items from the same `guns` category. Then roll a dice to determine if additional items from this list should spawn.
- `spawn_rate` parameter will only work for items placed in mapgen through use of item groups and won't work for items placed solitary (e.g. through use of `item` or `place_item`).

- Add this feature to game options.
- [x] Make global spawn chance affect only placing items through mapgen, and don't affect placing items after monster death.

#### Testing
Set global spawn chance for `food` and `drugs` categories to 0. Debug-spawned `s_pharm` OMS and checked that there's no food or drugs spawned at all.
~Set global spawn chance for `guns`, `armor`, and `clothing` categories to 0. Debug-spawned 81 zombie soldiers and debug-killed them all. Looked through their drops and checked that there's no guns, armor, or clothing spawned at all.~

#### Additional context
Pharmacy from testing above: note complete absence of food and drugs, while items with `tools`, `manuals`, `other`, and all other categories spawned like they should
изображение


src/map.cpp
```diff
@@ -4936,6 +5022,11 @@ item &map::add_item( const tripoint &p, item new_item )
     if( !inbounds( p ) ) {
         return null_item_reference();
     }
+
+    if( !check_for_spawn_chance( new_item ) ) {
+        return null_item_reference();
+    }
```
mqrause, line 5028: This is most definitely the wrong way to check for spawn chances, since it's (indirectly) called by pretty much everything, like dropping things on the floor etc.

Night-Pryanik: Yeah, you're right. I didn't test this case.

Night-Pryanik: [6268d2c](https://github.com/CleverRaven/Cataclysm-DDA/pull/62583/commits/6268d2c5b88848d98fb0c81de68cf7e88d32811d) should fix this.

data/core/game_balance.json
```diff
@@ -355,5 +355,180 @@
     "info": "Sets eternal weather.  Possible values are 'normal', 'clear', 'cloudy', 'light_drizzle', 'drizzle', 'rain', 'rainstorm', 'thunder', 'lightning', 'flurries', 'snowing', 'snowstorm', 'early_portal_storm', 'portal_storm'.  'normal' clears all eternal weather overrides and sets normal weather pattern.  Requires restart of the game after changing value to take effect.",
     "stype": "string_input",
     "value": "normal"
+  },
+  {
+    "type": "EXTERNAL_OPTION",
+    "name": "SPAWN_CHANCE_GUNS",
+    "info": "Set global chance to spawn items from 'guns' category.  Valid values are 0-100.  Set 0 to forbid spawning items from this category.  Default: 100.",
+    "stype": "int",
+    "value": 100
```
ZhilkinSerg, line 364: I believe `spawn_chance` would be better to be placed in https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/json/item_category.json - that way we won't have to deal with hardcode when we introduce new category.

PatrikLundell:
In addition to full and no spawn, I'd try to verify that partial spawn works as well, by e.g. setting the value to 10% (in theory there could be a bug that resulted in an all or nothing result, rather than a gradual scale), which ought to result in a rather noticeable reduction that can verify the functionality.

Also, is there a technical reason to cap it 100 rather than e.g. 1000? If someone wanted the overall spawn rate to be very low, but then wanted some category to be at or near the standard level (such as e.g. sticks and rocks) you'd have to reduce all categories except the target one, rather than reduce the overall rate and elevate the target.

Night-Pryanik:
> I'd try to verify that partial spawn works as well, by e.g. setting the value to 10% 

I'm not completely sure what did you mean by "partial spawn". I'm afraid there's a misunderstanding.

Percentage in external options don't increase or decrease amount of items spawned - because this is what `Item spawn scaling factor` world option does. Setting to 10% for some category doesn't mean that one will get only 10% of usual amount of items from that category spawned; it means that items from that category have only 10% chance to spawn at all.

> Also, is there a technical reason to cap it 100 rather than e.g. 1000?

The same here. This is not a spawn rate, this is a spawn chance. I think there's no need to put global chance to spawn items from some category higher than 100%.

Night-Pryanik:
> check for e.g. gun drops and verify that you'd get e.g. 10% (ish) of the normal number when the value is set to a reduced drop rate.

I tested this. 10%, 50%, 1%. I just didn't write in OP about all tests I've committed.

src/map.cpp
```diff
@@ -119,6 +120,32 @@ static const flag_id json_flag_AVATAR_ONLY( "AVATAR_ONLY" );
...
+static const item_category_id item_category_maps( "maps" );
+static const item_category_id item_category_mods( "mods" );
+static const item_category_id item_category_mutagen( "mutagen" );
+static const item_category_id item_category_other( "other" );
+static const item_category_id item_category_seeds( "seeds" );
+static const item_category_id item_category_spare_parts( "spare_parts" );
+static const item_category_id item_category_tools( "tools" );
+static const item_category_id item_category_veh_parts( "veh_parts" );
+static const item_category_id item_category_weapons( "weapons" );
+
```
ZhilkinSerg, line 148:
```suggestion
```

Night-Pryanik:
Ok, I moved spawn chance from external options to new parameter for `ITEM_CATEGORY` object. My only concern is that average player will have approximately zero chance to discover this new feature by himself. Even if he reads the documentation (which is by itself rather unlikely), he might not think that docs for item spawn chance is located in `Item Category` section.

We might need to increase availability of this new feature for players. Placing it into world options might be the perfect solution, but I know Kevin's position on this.

Night-Pryanik:
I reworked the whole feature. Now players can set both increased and lowered spawn rates for item categories.

alp7292:
si this planned to be added after the stable

Night-Pryanik:
Yes, I think so.

alp7292:
why it doesnt affect monster drops i think its fine. for example if you wanted to play 0.25 scale you would had less items in buildings but you would always find clothing armor and tools from zombies shifting focus from looting to killing zombies

Night-Pryanik:
The current status quo is that item spawn modifier doesn't affect monster drop rate, and I don't want to change it.

src/mapgen.cpp
```diff
@@ -6311,8 +6311,43 @@ std::vector<item *> map::place_items(
             tries++;
         } while( is_valid_terrain( p ) && tries < 20 );
         if( tries < 20 ) {
-            auto put = put_items_from_loc( group_id, p, turn );
-            res.insert( res.end(), put.begin(), put.end() );
+            for( const item &itm : item_group::items_from( group_id, turn, spawn_flags::use_spawn_rate ) ) {
+                const float item_cat_spawn_rate = std::max( 0.0f, item_category_spawn_rate( itm ) );
+                // for items with category spawn rate less than or equal to 1, roll a dice to see if they should spawn
+                if( item_cat_spawn_rate <= 1 ) {
+                    if( rng_float( 0.1f, 1 ) <= item_cat_spawn_rate ) {
```
Kromgart, line 6318: Looks like if you set item_cat_spawn_rate too low for some category (less than 0.1f) the item will always spawn? Because this rng_float will always roll greater than 0.1f.

Kromgart:
Can you please confirm:

1) To use this feature we need to edit corresponding item categories defined in data/json/item_category.json (or creating a tiny mod with them)

2) If the ITEM_SPAWNRATE game option is set (between 0 and 1), a particular item's creation must succeed 2 rolls: first against the ITEM_SPAWNRATE (old behavior), and then this new one.

src/mapgen.cpp
Kromgart, line 6318: I think rng_float( 0.1, 1.0) does uniform distribution? By starting from 0.1 the actual spawn rate will deviate from the expected category setting (the closer you set it to 0.1) since this is a direct comparison here.

For example: if I set Guns spawn rate to 0.15 I would expect them to be 15% rarer that usually. But rng_float( 0.1, 1.0 ) will roll a value <= 0.15 only in approx 5/90 = 5.6% cases making the guns 3 times rarer than my setting says.

Is there a reason to start the roll from 0.1 and not from zero?

Night-Pryanik:
> To use this feature we need to edit corresponding item categories defined in data/json/item_category.json (or creating a tiny mod with them)

Yes. I do however plan to change it to use options groups (#62989) when/if they will be merged.

> If the ITEM_SPAWNRATE game option is set (between 0 and 1), a particular item's creation must succeed 2 rolls: first against the ITEM_SPAWNRATE (old behavior), and then this new one.

With this new option item spawn must always succeed 2 rolls, assuming this item's category spawn rate is set in json.

src/mapgen.cpp
Night-Pryanik: I reserved setting to 0 to fully forbid spawning items from category in question.

alp7292:
Ekran Görüntüsü (141)
is it intentional that it increases item count but not variety

alp7292:
Ekran Görüntüsü (141)
is it intentional that item count increases but not the item variety

Night-Pryanik:
> is it intentional that item count increases but not the item variety

Well, yes. This option don't increase variety of items - this is a work entirely for definition of item groups. If original item group contains only Finger Lighter and Oil Generator CBMs, this option won't add nothing new to item spawn results, it will only increase the amount of these CBMs spawned.

bombasticSlacks:
It would probably end up being separate but related infrastructure. 

Would an ignore list for categories make more sense/be an easier quick fix? So all guns are at 3%, but this list of unique guns are still 100%.

Procyonae:
> It would probably end up being separate but related infrastructure.
> 
> Would an ignore list for categories make more sense/be an easier quick fix? So all guns are at 3%, but this list of unique guns are still 100%.

I mean that doesn't fix this not affecting item and place_item, it would be nice in addition to that but wouldn't be that useful on it's own

bombasticSlacks:
Oh I now understand your original question. I'm not sure why it doesn't already. I can put it on the list.

PR #62589, opened 2022-12-05
Migrate the "Retrieve Deposit Box" mission to JSON format

MNG-cataclysm:
#### Summary
Infrastructure "Move the Retrieve Deposit Box mission to JSON format"

#### Purpose of change
Move the missions reliant on C++ functions to a more accessible JSON format.

#### Describe the solution
I removed the C++ functions and replaced them with JSON.

Leaving it as is.

#### Testing
I took the mission, went to the bank, stole the deposit box, and returned it to its rightful owner.

#### Additional context
None.


data/json/npcs/missiondef.json
```diff
@@ -290,8 +290,23 @@
...
+    },
+    "end": {
+      "effect": [
+        "stop_following",
+        {
+          "weighted_list_eocs": [
+            [ { "id": "get_katana" }, { "const": 2 } ],
+            [ { "id": "get_deagle_44" }, { "const": 3 } ],
+            [ { "id": "get_m4_carbine" }, { "const": 4 } ]
+          ]
```
cathalpern, line 306: If I correctly understand how weighted list EOC works, I think this isn't quite a direct correlation to frequency in the code. Also, I'm bad at statistics, so I might be saying this wrong.
The code says, 33.3% of the time, end up with M4, 25% of the time, end up with katana, otherwise end up with desert eagle. Based on the NPC README, a higher value here corresponds to a higher likelihood. So I think desert eagle should be higher than M4? Ignoring the fact that the likelihood of katana and M4 are unrelated (there's a fancy stats word I'm forgetting), I think the desert eagle should be the result more frequently than the M4. If I were going to take a shot at roughly converting, I think I'd do - katana 5, M4 6, desert eagle 8

But anyone feel free to let me know if I'm totally misunderstanding how this attribute works :)

MNG-cataclysm: I tried to copy the general order of reward chances from the C++ as best I could, in their highest order, with the M4 being first.

cathalpern: std::string itemName = "deagle_44";
    if( one_in( 4 ) ) {
        itemName = "katana";
    } else if( one_in( 3 ) ) {
        itemName = "m4_carbine";
    }

I believe this translates to: start with item being desert eagle. roll a d4. if you get a 4, item changes to katana. roll a d3, if you get a 3, item changes to M4. give item to player.
So the likelihood of getting a desert eagle in the end is 3/4 x 2/3 = 6/12
so 3/12 chance of katana, 4/12 chance of M4, 6/12 chance of desert eagle
with some rounding... 
25/100, 33/100, 50/100
so I thiiink katana: 5, m4: 6, desert eagle: 10 (slight update after I actually attempted to math properly) would appropriately weight drop chances

cathalpern (approved):

data/json/npcs/missiondef.json
```diff
@@ -300,7 +315,7 @@
       "accepted": "You may make a tidy profit from this.",
       "rejected": "I know it isn't pressing but the world is going to be just as corrupt when we start rebuilding unless we take measure to stop those who seek to rule over us.",
       "advice": "This shouldn't be hard unless we run into a horde.",
-      "inquire": "Any luck?",
+      "inquire": "Do you have the deposit box?",
```
MNG-cataclysm, line 318:
```suggestion
      "inquire": "Any luck?",
```

Maleclypse (approved):

MNG-cataclysm:
@Maleclypse Could you remove the string freeze label, please?

PR #62608, opened 2022-12-06
Jsonify beehive

FarkasKBence:
#### Summary
Content "Jsonify beehive"

#### Purpose of change

Continues regional terrain project and roof project.

#### Describe the solution

Moved the hardcoded beehive to JSON.

I overhauled its mapgen to be more dungeon-style. Now it's a maze full of hexagonal chambers, with a reward *(royal jelly)* at the end if the player manages to find/fight their way there. I also gave it a simple vax roof and increased the number of bees around its perimeter.

I removed the beekeepers and their gear from this location. The beehive should no longer be a battleground between the bees and the constantly reanimating zombies.

A more complex colony with a queen, drones, guards, workers, nurses, and larvae.  I didn't feel compelled enough to add them here, but adding them becomes easier once their hive is moved to a more accessible format.

#### Testing

Loaded up an old save with the new changes.  No errors occurred during migration.

#### Additional context

<details>

  <summary>Screenshots:</summary>

kép
kép

</details>

This is my first time using GitHub Desktop and editing C++ so apologies for any mistakes. :)


Night-Pryanik:
So, it will always be a fixed layout instead of a randomized one?

> The beehive should no longer be a battleground between the bees and the constantly reanimating zombies.

I personally think it's better to extract beekeepers into their own faction and make bees and them neutral to each other.

Venera3:
Real bees aren't even neutral to beekeepers, hence the gear. And bees mgagically being chill when a zombie wears a net hat but not when it's wearing a hazmat suit would make no sense.

>A more complex colony with a queen, drones, guards, workers, nurses, and larvae.

I do have some plans for bees in general and a hive map sketched out, but no ETA.

LovamKicsiGazsii:
> So, it will always be a fixed layout instead of a randomized one?

Yes, unfortunately that feature couldn't translate well to JSON.

> I do have some plans for bees in general and a hive map sketched out, but no ETA.

Is my beehive map compatible with what you have planned or is there anything that needs to be changed?

Venera3:
Heh, the outline is pretty much the same with some extra openings on my end - for the cells I'll do 2x1 alcoves (the cells are oriented horizontally) in an alternating pattern, like
```
X X X
X X X
XXXXX
 X X
 X X
```
and some more Z-levels. As I said with Limb Stuff coming up I'm not sure when I'll find the time, and this is a huge step in the right direction already.

Inglonias:
Royal jelly is no longer the panacea that it was in earlier versions of the game. Not sure the reward justifies the risk here. That said, I also don't know what else you'd put at the end of the hive.

SirPendrak:
> Royal jelly is no longer the panacea that it was in earlier versions of the game. Not sure the reward justifies the risk here. That said, I also don't know what else you'd put at the end of the hive.

A maounatin of honey

cathalpern:
Minor note - might make sense to update to use the MON_BEEHIVE, which spawns a combination of mon_bee and mon_bee_small instead of MON_BEE. The original code I updated to use place_spawn included mon_bee (AKA giant bees), so switching to only space mon_bee_small (only member of MON_BEE) might unintentionally make this terrain easier than intended

Venera3:
Bees evolve (for now) so they will get bigger with time anyway.

Xaritscin:
about the Beekeeper dilemma, maybe instead of zeds it coul be Ferals(and even so i doubt it), or some kind of insect mutant gang that lives alongside the bugs.

Apis being the only actual mutant here is a very old design choice but i bet this will be changed at some point with the Bee rework.

Ilysen:
re: the beekeepers, I feel like what this PR does is fine honestly. I feel like there's basically zero loss in just... not having them there. if giant beehives are a post-Cataclysm development, why would there be any beekeepers there at all, even feral? just pitching my two cents to the convo here, at any rate - I think removing the beekeepers and having the inhabitants only be bees and Apis is probably fine

PR #62613, opened 2022-12-07
stomach_size_multiplier for dragon size mutations

AkkoRishtar:
#### Summary
Balance "Added stomach_size_multiplier in line with mainline size mutations to black dragon size mutations."

#### Purpose of change
Makes the black dragon mutation line viable considering the increased necessary nutrient intake.

#### Describe the solution
Added stomach_size_multiplier in line with mainline size mutations to black dragon size mutations.

Increasing metabolism_modifier and vitamins_absorb_multi, but that would both be illogical and go against the creators' intent with this mutation line.

#### Testing
Drank water till full, confirmed stomach contents, emptied stomach. Repeated with all 4 concerned mutations.

#### Additional context
Exact values are copied from mainline, can be considered placeholder until whoever is working on this decides what they should be.

### DIFF
--- a/data/mods/Magiclysm/mutations/mutations.json
+++ b/data/mods/Magiclysm/mutations/mutations.json
@@ -227,6 +227,7 @@
     "category": [ "DRAGON_BLACK" ],
     "passive_mods": { "str_mod": 2 },
     "weight_capacity_modifier": 1.05,
+    "stomach_size_multiplier": 1.5,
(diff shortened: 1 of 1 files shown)


AkkoRishtar:
I'll be frank, I have no idea how to fix the windows build fail.

Oatmeal crafting fail? What does that even have to do with my changes? 
Or are there simply weird quirks that only longtime contributors would understand?

Maleclypse (approved):

PR #62633, opened 2022-12-08
Nuclear Monster group and a specific comment

Maleclypse:
#### Summary
None

#### Purpose of change
Removes innappropriate mutant spawns from Nuclear Plant. 

#### Describe the solution
Removes many monsters from mutant spawns

None


8street:
It's a pity that we don't have many locations with terrible monsters in game.

PR #62666, opened 2022-12-10
make the one metal door in hunting lodge pickable

gettingusedto:
#### Summary
None
#### Purpose of change

The locked metal door in the hunting lodge being unpickable doesn't really makes sense to me. Unless someone tells me if the metal door is intentional, I'm swapping the locked metal door with the lockpickable locked metal door.

#### Describe the solution

Change `t_door_metal_locked` to ~~`t_door_metal_pickable`~~ `t_door_metal_interior_locked` in the lodge's mapgen palette.

Not doing so.

#### Testing

It's just 1 line change. Assuming I write the id correctly it should work as expected.

### DIFF
--- a/data/json/mapgen_palettes/lodge_palette.json
+++ b/data/json/mapgen_palettes/lodge_palette.json
@@ -6,7 +6,7 @@
       "#": "t_rock_wall",
       "+": [ [ "t_door_c", 3 ], "t_door_locked" ],
       "=": "t_door_c",
-      "z": "t_door_metal_pickable",
+      "z": "t_door_metal_interior_locked",
       ".": [ [ "t_region_groundcover", 4 ], "t_region_groundcover_forest" ],
       "~": [ [ "t_region_groundcover_barren", 3 ], "t_region_groundcover" ],
       "*": [ [ "t_region_groundcover", 15 ], "t_region_shrub" ],


data/json/mapgen_palettes/lodge_palette.json
```diff
@@ -6,7 +6,7 @@
       "#": "t_rock_wall",
       "+": [ [ "t_door_c", 3 ], "t_door_locked" ],
       "=": "t_door_c",
-      "z": "t_door_metal_locked",
+      "z": "t_door_metal_pickable",
```
Maleclypse, line 9:
```suggestion
      "z": "t_door_metal_interior_locked",
```
I specifically didn't use the door you've selected as your change would have made it unlocked on it's inside sides.  Which are both sides.  Now that someone has added a pickable metal door that doesn't open to both interior sides this can be changed.

Maleclypse (changes requested):

Maleclypse (approved):
Thanks for updating this map!

PatrikLundell:
Assuming you write the ID correctly and chose the right ID, understood the door functionality correctly, modified the correct file,... it should work.

I'd spawn the location and verify the door can actually be picked.

gettingusedto:
I made the change in my build. The test as follows:
1. I spawn in two hunting lodge, the one with glooscaps and the one without with debug.
2. Checked that the metal door is the `t_door_metal_interior_locked` one, as stated in the palette change.
3. Giving myself lockpicking proficiencies and then lockpick the door, it opens for both types of the hunting lodge
Screenshot_20221211-165216

PR #47080, opened 2021-01-28
Added Gelatin and Several Gelatin Based Recipes

GrubGrubgit:
#### Summary
Content "Added several gelatin based items and recipes." 

#### Purpose of change

I felt gelatin and gelatin based products were lacking considering the depth of CDDA's cooking and food.

#### Describe the solution
Added

- Powdered gelatin and recipe to dry fresh gelatin
- Fresh gelatin and recipe to extract gelatin from raw hide
- 4 flavours of gelatin dessert powder
- 4 flavours of gelatin dessert
- 2 Homemade gelatin dessert recipes
- 2 vegan dessert recipes using agar
- gummy candies and recipes both from gelatin and agar

Changes

- Modified both aspic recipes to use gelatin
added fresh gelatin as a byproduct from making bone broth
- added powdered gelatin, gelatin dessert powder and agar to dry food item group
added gelatin dessert cups to fridge item group
- added vegan gelatin dessert to vegan SUS fridge item group

#### Describe alternatives you've considered

I'm unsure about the initial recipes for extracting gelatin. It currently uses the distillation vat. I'm not sure this is the best method. it's the only way i found to have things happen passively over time though.

#### Testing

Recipes and items all tested and working correctly.

PR #62669, opened 2022-12-10
Fridges no longer spawn loose, container-less eggs

### LINKED ISSUE #58536: Some eggs spawn outside of egg cartons
### Describe the bug

Refrigerators in houses can have egg cartons with some eggs in them, but they also contain standalone eggs.

### Steps to reproduce

1. Go to a fridge. (You can start in a house by selecting an In Town scenario like Missed, then going to the last screen on character creation and pressing `/`.)
2. Look at what's in the fridge (such as with `g`) and see what the eggs look like.

### Expected behavior

Eggs should be in egg cartons, because that's why egg cartons were added to the game (source: #47020), and it seems odd to have half of them rolling around on their own inside fridges.

### Screenshots

CDDA escaped eggs

- OS: Windows
    - OS Version: 10.0.19043.1766 (21H1)
- Game Version: d422ab7 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

Sometimes the total number of eggs is more than 12, so it's possible there are also twice as many eggs than there should be—one set in a carton, one set out. Of course, there's no reason a fridge couldn't have more than 12 eggs or 1 carton.

Ilysen:
#### Summary
Content "Fridges no longer spawn eggs without cartons"

#### Purpose of change

I've only recently noticed that this is actually happening, and it's very silly indeed. I certainly haven't met anyone in real life that leaves loose eggs lying around in their fridge, at least. The change was implemented in #47080 but it wasn't mentioned in the comments or the OP, so I can only assume it was an oversight.

Resolves #58536

#### Describe the solution

Removed the extraneous egg spawn line that could spawn up to an extra dozen eggs without a container. Now, eggs will always be in cartons. This has the side effect of reducing the overall number of egg spawns, but I have a hunch that fixing loose eggs on the bottom of the fridge won't affect balance too badly :P

Keeping the extra eggs, just putting them in a carton. One thing that comes to mind is having the potential for a carton with one or two eggs to show up on a lower spawn rate, to represent that the fridge owner had gotten another carton before running out of the first one, but I'm not sure how many people in real life do that kind of thing or not so I opted on the safe side of making sure that eggs would always show up in cartons.

#### Testing

Jumped to some houses on my new world that were generated before the change. Eggs generated both in cartons as well as loosely inside of fridges. Then, I generated some new terrain and checked the fridges there. Extra eggs no longer appeared.

### DIFF
--- a/data/json/itemgroups/SUS/fridges.json
+++ b/data/json/itemgroups/SUS/fridges.json
@@ -43,7 +43,6 @@
       { "item": "butter", "prob": 80 },
       { "item": "pudding", "prob": 30 },
       { "item": "gelatin_dessert_processed", "prob": 20 },
-      { "item": "egg_bird_unfert", "prob": 85, "count-min": 1, "count-max": 12 },
       { "item": "egg_bird_unfert", "prob": 85, "charges-min": 1, "charges-max": 12, "container-item": "carton_egg" },
       { "item": "bacon", "prob": 25 },
       {


Night-Pryanik:
#58536?

Ilysen:
Will add mention of it to the OP - thanks

Rabidweasel:
I always assumed that fridges in the US typically had one of those [plastic egg holders](https://www.4fridgemaster.co.uk/fridge-door-shelf-egg-tray/product.pl?pid=3239578) fitted into the door. They're pretty common in the UK even though most people I know don't use them and don't refrigerate eggs.

Zireael07:
My fridge has it too, I still think it counts as a container in CDDA terms

Ilysen:
I've heard that thrown around a few times, but for a few reasons I don't think it's what was trying to be represented here - in addition to spawning in 85% of fridges (and no fridge I've ever used has such a holder, at least here in the US), it also spawned *in addition* to eggs in cartons, which would negate the point of the holder to begin with. I figure that if it was meant to be an egg holder, it would be mutually exclusive with the cartons, instead of potentially spawning a full dozen eggs in a carton and a full dozen loose eggs as well. That, coupled with the fact that if it *was* an egg holder we could easily use a purpose-made pocket, is what makes me think it's not what was intended here, especially since there was no mention of such a thing in the PR that originally implemented the loose eggs

Zireael07:
I wasn't saying that the holder explained loose eggs. I was saying that in CDDA terms, those eggs should still have a container of some sort themselves ;) But I phrased that badly

PR #62676, opened 2022-12-10
fix: magiclysm's thunder pillar spells

casswedson:
#### Summary
None

#### Purpose of change
- fix ranges for special attact activation from just a tile range e.g. [12-12] to a say [5-12]
- also randomize field intensity a little; I like randomness quite a lot you see


GuardianDll:
Wait, min_damage 0 means it doesn't show the info about summoned monster?

casswedson:
testing lead me to believe it 0 as damage in a summon spell means it doesn't summon anything? on 0 it didn't spawn the pillar, so it's a bug right

GuardianDll:
No, 0 means it spawn nothing indeed, but all attunement spells are given at lvl 5, which round the damage to 1, and makes you able to spawn 4 monsters exactly at level 35
changing damage_min to 1 makes player able to summon two of them right after level 5 - so right from the start
image

casswedson:
Ohh dang I didn't realize the attunement business, I can live with reverting the summon spell damage change

GuardianDll (approved):
👍

PR #62678, opened 2022-12-10
fix: Exodii quadruped spells (again)

casswedson:
#### Summary
None

#### Purpose of change
I discovered how to get multiple spells to work as special attacks in a monster something the exodii quadruped wanted but it didn't quite had

it would attempt to fire a wrapper spell, but ranges and cooldown was really limited by that, like it would fire all spells at one, super deadly - kinda meh, not it fires them at different rates and ranges FUN!

#### Testing
don't mind the typo in the screenshot

you can see it here using its saw spell as well as one of its blasters, also the gun too
Screenshot from 2022-12-10 16-02-22

### DIFF
--- a/data/json/monster_special_attacks/weapon_spells.json
+++ b/data/json/monster_special_attacks/weapon_spells.json
@@ -54,6 +54,7 @@
   {
     "type": "SPELL",
     "id": "exodii_weapon_pod",
+    "//": "deprecated, remove later.",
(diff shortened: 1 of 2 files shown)


GuardianDll:
Why not remove the spell here? The pr is holded by string freeze anyway

casswedson:
obsoleted by #63730

Issue #62681, opened 2022-12-10
Track by scent is broken in several ways


Maleclypse:
### Describe the bug

Firstly track by scent is currently set by species despite almost every species in game is too broad to have all the monsters in a given species to track by scent. Secondly, the logic in the c++ seems to be broken so as to prevent any individualized changes by mon_id to whether something tracks by scent or not. Thirdly, basically every species of monster currently tracks by scent including Robots.

### Steps to reproduce

akrieger — Today at 4:55 PM
```
    const scenttype_id &type_scent = scents.get_type();
    bool right_scent = false;
    // is the monster tracking this scent
    if( !tracked_scents.empty() ) {
        right_scent = tracked_scents.find( type_scent ) != tracked_scents.end();
    }
    //is this scent recognised by the monster species
    if( !type_scent.is_empty() ) {
        const std::set<species_id> &receptive_species = type_scent->receptive_species;
        const std::set<species_id> &monster_species = type->species;
        std::vector<species_id> v_intersection;
        std::set_intersection( receptive_species.begin(), receptive_species.end(), monster_species.begin(),
                               monster_species.end(), std::back_inserter( v_intersection ) );
        if( !v_intersection.empty() ) {
            right_scent = true;
        }
    }
```

basically the first right_sent is ignored if the species is receptive to the scent type at all

### Expected behavior

I would expect scent tracking to not be a universal thing performed by every creature in existence that is not the player character.

### Versions and configuration

Current version until fixed

Maleclypse:
```
 {
    "type": "scent_type",
    "id": "sc_human",
    "receptive_species": [
      "MAMMAL",
      "BIRD",
      "AMPHIBIAN",
      "REPTILE",
      "FISH",
      "MUTANT",
      "NETHER",
      "SLIME",
      "FUNGUS",
      "INSECT",
      "CENTIPEDE",
      "SPIDER",
      "PLANT",
      "MOLLUSK",
      "WORM",
      "ZOMBIE",
      "ROBOT",
      "HORROR",
      "ABERRATION",
      "HALLUCINATION",
      "HUMAN",
      "UNKNOWN"
    ]
  },
```

Maleclypse:
Ideally I had hoped that everything was working and I could just create a project list to go through the monster ids and make it so base zombies don't automatically have the noses of bloodhounds, but it appears that all monsters have the noses of bloodhounds currently.

mqrause:
[...]

PR #62685, opened 2022-12-11
Scent Tracking Zeds

Maleclypse:
#### Summary
Balance "What Zombies should track by scent?"

#### Purpose of change
Currently all Zombies are bloodhounds as far as finding people.  So are almost all monsters.  As part of #62681 I'm focusing on one species at a time.  Zombies to start with.

#### Describe the solution
Removed the species wide ability for Zombies to sniff out prey and specifically added it back on zombies that seemed reasonable to have it. 

Attracting attention to the issue about this and making it more clear I'm looking for help on this project.  

#### Testing
Night raiding became much safer until I got overconfident and a zombie dog caught my scent.


Venera3:
I think the shady, screecher and the predator lines definitely need it (+- `KEENNOSE` for some better tracking) either from a functional or a flavor perspective, and I'd add the base level to the evolved brawlers as well mostly for want of a flavor on them. 

Headless/acid/bone/shocky zeds not smelling you is fair enough, and slightly restricting smell should keep single zeds from peeling off and saying Hello 1v1 randomly.

Night-Pryanik:
> specifically added it back on zombies that

That what?

SabrinaMori:
I am playing on stable, but i never had zombies sniff me out, during the night especially i can walk almost past them, do you have strong scent perk or maybe you just encountered `KEENNOSE` special zombies?

Maleclypse:
> I am playing on stable, but i never had zombies sniff me out, during the night especially i can walk almost past them, do you have strong scent perk or maybe you just encountered `KEENNOSE` special zombies?

This is more of a review prompted by comments from Kevin about baseline humans not being able to track by smell.  So my assumption is that non-mutated risen human dead, ferals, and probably most robots shouldn't track by smell.

Kribylet:
Is the accuracy of scent tracking the same across all scent tracking zeds? Could we consider baseline zombies being able to track some extremely high threshold of scent? For example, survivor hiding in a room, next to the door with a baseline zombie on the other side? That could make negatives that affect scent intensity matter some more, maybe.

SabrinaMori:
> This is more of a review prompted by comments from Kevin about baseline humans not being able to track by smell. So my assumption is that non-mutated risen human dead, ferals, and probably most robots shouldn't track by smell.

Oh yeah in that context it does make more sense that zombies just like normal humans shouldn't be able to track their prey using directional smells. I'm not an expert on the game mechanics but i remember smell is based on how long you stayed in the area, so the longer you stay in one spot, the more smell you leave there, and that high amount of smell should be detectable even by normal people.

Maleclypse:
> Is the accuracy of scent tracking the same across all scent tracking zeds? Could we consider baseline zombies being able to track some extremely high threshold of scent? For example, survivor hiding in a room, next to the door with a baseline zombie on the other side? That could make bonuses/maluses that affect scent intensity matter some more, maybe.

There are currently two levels of smell, normal and keen nose.

AcidAntOnAMinefield:
Zombies shouldn't be able to track scent in any way - assuming they made of rotting flesh (as some of them have it in their descriptions), that scent would overpower pretty much everything else. And this is assuming their smelling senses even work to begin with. Plus they also have rotting clothes, or their own "smells" (gas, smoke, murky water, etc) over their "pleasant rotting flesh" aroma.

But if we want to stretch it a bit, I'd probably suggest that the only zombie creatures that should reasonably be allowed to follow scent would be zombie animals and predators. 

Also Headless/Brainless Zombies are still a strange concept given that most of the sensory receptors require a working brain - how can a headless creature even stand to begin with, how can it move with no central coordination to itself (ie "brain") or how can they even track sound? This is a legitimate question, given that the "brainless zombie" (and the "Listener Zombie", which evolves from it) have no brain, *yet* their hearing ability still exists? :thinking:

PR #62646, opened 2022-12-09
Add space for easy translation

GeekDuanLian:
#### Summary
I18N "Add space for easy translation"

#### Purpose of change
Because some languages ​​such as Chinese do not need spaces to separate, so you can delete the spaces when translating.

#### Describe the solution
Add space.

#### Describe alternatives you've considered
Don't Add space.

#### Testing
Will CI automatically compile for me? After compiling, I will test it.

#### Additional context
None.

PR #62690, opened 2022-12-11
Allow self-compilation after user fork

GeekDuanLian:
#### Summary
Build "Allow self-compilation after user fork"

#### Purpose of change
I hope that those who want to commit changes don't have to configure a build environment on their own machines to test the changes. Or use the official standard compilation environment to compile to ensure that it will not cause its own errors. Foreword and inspiration from #62646

#### Describe the solution
I modified the release.yml to run on any repository, since that's the only thing I need to run.
And I added the option to manually trigger the run (`workflow_dispatch`), so that it can be run on a branch other than `master`, and it will not automatically run when the user does not need it, nor will it interfere with the way the official repository operates.

These are how users operate:
- Open a specific file in the official repository
- Click 🖊 in the upper right corner to edit
- Github automatically forks and creates new branch `patch-1` and then changes
- The user submits a PR, or he wants to test the changes before doing so
- The user enters the repository of his fork, selects the `Actions` tab at the top of the page, and clicks to enable actions
- Select `Experimental Release` in the actions list on the left, click the `Run workflow` button on the right, select the branch you have modified, it should be `patch-1`, and click `Run`
- Wait for a while, you should be able to see the compilation results on the `Release` page of the user repository
- Repeat the above steps, make changes, manually run compilation when needed, and finally submit PR

Burn user's own CPU.

#### Testing
I just used this method to let CI compile successfully for me, and downloaded it to play for a while, it is the same as the release in the official repository.
See https://github.com/GeekDuanLian/Cataclysm-DDA/releases

#### Additional context
Maybe a document should be written to tell users that CI can be used to automatically compile, but my English is poor, I don't know how to write it. 😢

By the way, press the `.` button in the repository can open the web version of `Visual Studio Code`, which is convenient for editing and submitting changes purely online.


GeekDuanLian:
In fact, all compilations show failures, because the user repository is not configured with `github pages`, and Android compilation needs to fill in the packaging key. But before that release has been uploaded.
Some judgment needs to be added to selectively execute, so as not to cause errors to harass users' mailbox reminders.

GeekDuanLian:
Shit, I used the wrong command, wait for me to process it again.

PR #62694, opened 2022-12-11
Mapgen bash now avoids terrain

### LINKED ISSUE #54237: Libraries have holes in the floor
### Describe the bug

Some variants of the library overmap special have holes in the floor (ledges).

### Steps to reproduce

1. Use the map editor to spawn a bunch of libraries
2. Walk to them (long range teleport masks this bug)
3. Find a variant with holes in the floor

### Expected behavior

I expected to find the concrete foundation in places where the wooden library floor is damaged.

### Screenshots

DeepinScreenshot_select-area_20220110100832

- OS: Linux
    - OS Version: LSB Version: :core-4.1-amd64:core-4.1-noarch; Distributor ID: Fedora; Description: Fedora release 34 (Thirty Four); Release: 34; Codename: ThirtyFour; 
- Game Version: dd5c1fc [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

Possibly related: #53549

Ilysen:
#### Summary
Bugfixes "Vandalized libraries, collapsed towers, etc. should no longer spawn with holes in the floor and walls"

#### Purpose of change

Mapgen bash logic is kind of ancient, and for a while now it's caused some oddities when it was used. Generally speaking, because it would bash terrain with a max-strength hit, it would leave a bunch of holes in the ground. Then, on a save-load, the holes would be replaced with dirt. For things like libraries, this was an obvious problem, where rioters were apparently strong enough to smash holes straight through the foundation and into the sewers. Crazy happenings, I tell ya.

Fixes #54237

#### Describe the solution

`bash_params` has two new optional arguments that allow it to skip furniture or terrain while it's being run. Both of these arguments are false by default, but the bash call invoked by `"point": "bash"` in JSON is now set to skip terrain. As a result, floors and walls are no longer targeted by the bash, but furniture remains a valid target and will be smashed up by mapgen.

There's room for a flag or other way to identify things like windows and doors, which might be valid bashing targets even though floors and walls should not be. Because vandalized libraries already include separate handling for their windows, though, I opted on the side of keeping it simple for now to make sure it works for stable.

#### Testing

I spawned in a set of libraries outside of generated terrain, and tried approaching them both before and after setting the bash call to skip terrain. Before, the libraries looked as they do now; after, they looked as they do in the screenshot below, with the terrain being intact while the furniture was damaged.

I also spawned a collapsed tower and jumped over there to make sure it looked correct. The furniture was still all heavily damaged, but the floor was intact, leaving the only entrances to the lower level as the elevator shafts and the destroyed staircase nearby.

I'd like to play my normal run using these changes for a while to make sure it didn't incidentally break anything else bash-related. I smacked up some furniture and terrain on my own and it worked fine, and the optional nature of the variable should mean that it only affects mapgen, but DDA's a big game and it's definitely possible I missed something.

#### Additional context

Here's a library generated after the fix. Note how the bookshelves, counters, chairs, etc. are all bashed up, but doors, walls, and floors are not.

cataclysm-tiles_tqbp2gouvI

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -3707,11 +3707,11 @@ void map::bash_ter_furn( const tripoint &p, bash_params &params )
 
     bool success = false;
 
-    if( has_furn( p ) && furnid.bash.str_max != -1 ) {
+    if( has_furn( p ) && furnid.bash.str_max != -1 && !params.skip_furniture ) {
         bash = &furnid.bash;
         smash_furn = true;
-    } else if( ter( p ).obj().bash.str_max != -1 ) {
-        bash = &ter( p ).obj().bash;
+    } else if( terid.bash.str_max != -1 && !params.skip_terrain ) {
+        bash = &terid.bash;
         smash_ter = true;
     }
 
@@ -3967,10 +3967,11 @@ void map::bash_ter_furn( const tripoint &p, bash_params &params )
 
 bash_params map::bash( const tripoint &p, const int str,
                        bool silent, bool destroy, bool bash_floor,
-                       const vehicle *bashing_vehicle )
+                       const vehicle *bashing_vehicle,
+                       bool skip_furniture, bool skip_terrain )
 {
     bash_params bsh{
-        str, silent, destroy, bash_floor, static_cast<float>( rng_float( 0, 1.0f ) ), false, false, false, false
+        str, silent, destroy, bash_floor, skip_furniture, skip_terrain, static_cast<float>( rng_float( 0, 1.0f ) ), false, false, false, false
     };
     if( !inbounds( p ) ) {
         return bsh;
@@ -4201,7 +4202,7 @@ void map::shoot( const tripoint &p, projectile &proj, const bool hit_items )
                 const int max_damage = shoot.destroy_dmg_max - shoot.destroy_dmg_min;
                 if( x_in_y( min_damage, max_damage ) ) {
                     // don't need to duplicate all the destruction logic here
-                    bash_params bsh{ 0, false, true, false, 0.0, false, false, false, false };
+                    bash_params bsh{ 0, false, true, false, false, false, 0.0, false, false, false, false };
                     bash_ter_furn( p, bsh );
                     destroyed = true;
                 }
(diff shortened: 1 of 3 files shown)


NetSysFire:
I wonder if that could also potentially fix the holes made by mx_crater since iirc some complained about them since autodrive will happily crash your vehicle into them

Ilysen:
I actually tried that out locally while I was making this; the holes actually seem to be deliberate *for the crater specifically*, because the way they're actually generated is just by blowing up a debug bomb in the center. This is also why you sometimes see them cut off halfway - I think the "bomb" is exploding on the edge of the reality bubble when that happens, and so some of the terrain that might otherwise be hit isn't loaded.

A save-and-load next to the crater also doesn't replace the dirt, so those things together lead me to believe that it's intentional, as weird as it might seem. The main problem with them is as a result in autodrive being a goober, rather than the crater being weird. I tested to make sure, believe me :P

Venera3:
Bashing terrain needs to be at least an option on the JSON side, and keeping doors intact is already a bit weird in this case. There's at least one microlab setpiece that will get messed by this (as in, a mutant pack separated by now-unbashed doors between them).

Ilysen:
I'll look into it

Ilysen:
so I'm trying to make this work, but the code is very ancient and it may necessitate me having to add straight-up new variables to the base info class used here, which seems extremely jank. the problem here is that, evidently, some uses of `"point": "bash"` assume smashable terrain even though it smashing terrain is unintentional, and I'm not super sure which parts should. is it not possible to use nested palettes to make sure that doors spawn smashed/open? I can happily do some of that on my end if it'd help

John-Candlebury:
Please make sure the mapgen bashing that happens in the collapsed tower actually smashes walls. That is meant to happen and the place doesn't look properly ruined if it doesn't.

Ilysen:
bwah, fuck it. this code is ancient and is giving me a lot of trouble, unfortunately. it's someone else's problem now :P

PR #62696, opened 2022-12-11
Allow self-compilation after user fork

GeekDuanLian:
#### Summary
Build "Allow self-compilation after user fork"

#### Purpose of change
I hope that those who want to commit changes don't have to configure a build environment on their own machines to test the changes. Or use the official standard compilation environment to compile to ensure that it will not cause its own errors. Foreword and inspiration from #62646

#### Describe the solution
I modified the `release.yml` to run on any repository, since that's the only thing I need to run.
And I added the option to manually trigger the run (`workflow_dispatch`), so that it can be run on a branch other than `master`, and it will not automatically run when the user does not need it, nor will it interfere with the way the official repository operates.
Finally I added some judgments:
- When user doesn't set `secrets.KEYSTORE`, android compilation is not performed because it needs this.
- Does not `Trigger GitHub pages rebuild` when it is not an official repository, because user repositories do not have GitHub pages configured.

These are how users operate:
- Open a specific file in the official repository.
- Click 🖊 in the upper right corner to edit.
- Github automatically forks and creates new branch `patch-1` and then changes.
- The user submits a PR, or he wants to test the changes before doing so.
- The user enters the repository of his fork, selects the `Actions` tab at the top of the page, and clicks to enable actions.
- Select `Experimental Release` in the actions list on the left, click the `Run workflow` button on the right, select the branch you have modified, it should be `patch-1`, and click `Run`.
- Wait for a while, you should be able to see the compilation results on the `Release` page of the user repository.
- Repeat the above steps, make changes, manually run compilation when needed, and finally submit PR.

Burn user's own CPU.

#### Testing
I just used this method to let CI compile successfully for me, and downloaded it to play for a while, it is the same as the release in the official repository.
See https://github.com/GeekDuanLian/Cataclysm-DDA/releases

#### Additional context
Maybe a document should be written to tell users that CI can be used to automatically compile, but my English is poor, I don't know how to write it. 😢

By the way, press the `.` button in the repository can open the web version of `Visual Studio Code`, which is convenient for editing and submitting changes purely online.

### DIFF
--- a/.github/workflows/push-translation-template.yml
+++ b/.github/workflows/push-translation-template.yml
@@ -12,11 +12,7 @@ on:
 jobs:
   push-template:
     runs-on: ubuntu-latest
-    if: >
-      ${{ github.event.workflow_run.event == 'push' &&
-      github.event.workflow_run.conclusion == 'success' &&
-      github.event.workflow_run.status == 'success' &&
-      github.repository == 'CleverRaven/Cataclysm-DDA' }}
+    if: ${{ github.event.workflow_run.event == 'push' && github.event.workflow_run.conclusion == 'success' && github.event.workflow_run.status == 'success' && github.repository == 'CleverRaven/Cataclysm-DDA' }}
     steps:
       - name: "Install gettext tools"
         run: sudo apt install gettext
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,7 +64,7 @@ jobs:
           gh api \
             --method POST \
             -H "Accept: application/vnd.github.v3+json" \
-            '/repos/${{ github.repository }}/releases/generate-notes' \
+            /repos/'${{ github.repository }}'/releases/generate-notes \
             -f tag_name='${{ steps.generate_env_vars.outputs.tag_name }}' \
             -f target_commitish='${{ github.ref_name }}' \
             -q .body > CHANGELOG.md
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,6 +1,7 @@
 name: "Experimental Release"
 concurrency: release
 on:
+  workflow_dispatch:
   push:
     branches:
       - master
@@ -29,6 +30,7 @@ jobs:
       upload_url: ${{ steps.create_release.outputs.upload_url }}
       timestamp: ${{ steps.get-timestamp.outputs.time }}
       release_already_exists: ${{ steps.tag_check.outputs.exists }}
+      matrix: ${{ steps.matrix.outputs.matrix }}
     steps:
       - name: Get build timestamp
         id: get-timestamp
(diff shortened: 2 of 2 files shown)


snipercup:
If this gets merged I will be happy to document what you wrote in your PR. Do you think something similar could be done for the object creator?

casswedson:
I made something a little similar to this for personal use: https://github.com/casswedson/Cataclysm-Compiler/blob/main/.github/workflows/compiler.yml
https://github.com/casswedson/Cataclysm-Compiler/actions/workflows/compiler.yml
a few optimizations I made:
- ccache by default, sometimes (not most of the time) it really speeds up compilation
- very specific target + preferred compilation options, I compile a version I would use, to test or play
- "smart" use of uploading assets, my workflow uploads just the bare minimum, just the executable, in my case I just want that; download times matter for me and I have a local copy of the repo I really don't need anything else

I would probably not do this how you present it; the points above are reasons, [1] speed, the releases can get really slow, I don't know much about the capacity of a random fork but the releases generate a considerable amount of processes which probably prolongates compilation for your target - [2] if you use windows why compile linux or osx not once but multiple times with multiple compilation options, tiles, no tiles, a soundpack. selectable target/compilation options would be amazing - [3] if you have a computer to test you probably already have most of the data you need as a copy of your fork, this ships a full game

GeekDuanLian:
@casswedson
Compilation is performed synchronously, all versions of all platforms will be compiled at the same time and the results will be uploaded after completion, so it takes the same time as compiling only the version you want.

The main reason is that I pursue the same result as the official compilation, so that the problem can be reproduced without causing new problems, and the upstream update can be synchronized without having to maintain it myself.

I took a look at my compile times, probably faster than yours:
21m6s https://github.com/GeekDuanLian/Cataclysm-DDA/actions/runs/3669598868/jobs/6203809818
23m18s https://github.com/casswedson/Cataclysm-Compiler/actions/runs/3697104299/jobs/6261679540
In the absence of ccache.

So the only function I might want to add is ccache, but I don't know the technical principle, and I don't know what impact it will cause, otherwise it should be officially used by default.

Maybe you should open another issue or PR to add ccache function to the official to speed up compilation? For example, ccache is automatically used in the case of workflow_dispatch. I don't know how to add it and what effect it might have.

What this PR needs to solve is to add a simple self-compilation function, which is all I know how to do.

GeekDuanLian:
> If this gets merged I will be happy to document what you wrote in your PR. Do you think something similar could be done for the object creator?

Sure! I still have some tutorials to add, but I'll wait for the official merge. Otherwise, it would be a waste of time not to merge in the end.

andrei8l:
> When user doesn't set secrets.KEYSTORE, android compilation is not performed because it needs this.

Better generate a temp keystore and use that

GeekDuanLian:
@andrei8l 
The keystore is for apk signatures. APKs with different signatures cannot be overwritten and installed by default. They can only be uninstalled and reinstalled.

So using a new signature for each compilation means uninstalling and reinstalling for each test, which may be troublesome. What I think is that if you need to compile apk, you should manually generate a signature for your own use and put it in secrets.

andrei8l:
> @andrei8l The keystore is for apk signatures. APKs with different signatures cannot be overwritten and installed by default. They can only be uninstalled and reinstalled.

You won't be able to overwrite the play store version with a self-compiled version either way. Having the wrong signature is also not a problem for android emulators.

>  What I think is that if you need to compile apk, you should manually generate a signature for your own use and put it in secrets.

You also need to replicate the variable names in `keystore.properties` which aren't standardized or documented anywhere.  If someone wants to work on andoid extensively, they can add a key, but for one-off patching there should be a working solution by default.

PR #62702, opened 2022-12-11
fixes #62655 Adjusted powdered eggs requirement from 10 to 2 in leverpostej food recipe, to be consistent with other egg types and recipes

### LINKED ISSUE #62655: Leverpostej Recipe and Powdered Eggs
### Describe the bug

It seems like powdered eggs and normal-sized bird eggs are interchangeable at a rate of 1:1 in basically all recipes where powdered eggs are allowed.  However, in the leverpostej recipe, the ratio is 10:2 instead.

### Steps to reproduce

1. Look at leverpostej recipe:
image

2. Look at other recipes which can use powdered eggs and other bird eggs interchangeably:
image

### Expected behavior

Presuming there isn't anything special about leverpostej, the conversion rate between normal-sized bird eggs and powdered eggs should be equivalent across recipes.

### Screenshots

_No response_

- OS: Windows
- Game Version: 50d81a1 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Slots [cbm_slots]
]

### Additional context

_No response_

Gabriel-Pociecha:
#### Summary
Balance "Changed amount of powdered eggs required in leverpostej recipe from 10 to 2 to be equal to all other egg types"

#### Purpose of change
Fixes #62655 For other recipes that can use multiple egg types, the amount required is 1:1. In the case of Leverpostej recipe, powdered eggs specifically required 10, while all others required 2. This is inconsistent with other recipes allowing different egg types, in which the ratio is 1:1 still, so I lowered the amount of powdered eggs required to 2 like all of the other eggs available to use in the recipe.

#### Describe the solution
Changing the amount of powdered eggs required for the Leverpostej recipe from 10 to 2 (the amount required of all other egg types) balances the game for this recipe and makes it consistent with all other recipes like it.

An alternative solution, if powdered eggs were meant to have a discrepant value compared to other eggs, would be to adjust its values in all other recipes that require it accordingly, rather than lowering it in this single recipe. But considering it is the only case this discrepancy occurs, I believe that it was meant to be the same value as the other eggs, hence the change in this pull request.

#### Testing
I made the simple change locally on my device in its files, and then used the debug commands (before and after the change) to learn and view the leverpostej recipe to confirm that the change went through. I also spawned in the necessary items themselves and confirmed that only 2 powdered eggs are consumed now.

#### Additional context
Images of the leverpostej recipe in game before and after the change in this pull request, marked to emphasize the change:
egg_before
egg_after

Also, images of inventory with 22 powdered eggs before crafting and 20 powdered eggs after crafting the leverpostej recipe.
22eggs
20eggs

### DIFF
--- a/data/json/recipes/food/offal_dishes.json
+++ b/data/json/recipes/food/offal_dishes.json
@@ -252,7 +252,7 @@
     "components": [
       [ [ "meat_liver", 8, "LIST" ] ],
       [ [ "edible_fat", 2, "LIST" ], [ "edible_lard", 2, "LIST" ] ],
-      [ [ "powder_eggs", 10 ], [ "eggs_bird", 2, "LIST" ] ],
+      [ [ "powder_eggs", 2 ], [ "eggs_bird", 2, "LIST" ] ],
       [ [ "onion", 1 ], [ "irradiated_onion", 1 ] ],
       [ [ "flour_any", 2, "LIST" ] ],
       [ [ "salt", 3 ] ],


GuardianDll (approved):
👍

anoobindisguise:
I already fixed this in my food costs PR, I wish I had seen this sooner so I could have saved you some time!

PR #62707, opened 2022-12-12
dissasemble block engines

Lalomb0000:
#### Summary
Content "cutting up the engine blocks into lumps of steel"

#### Purpose of change

there is no use for engine blocks once disassembled, and the progress of survivors requires raw materials to fight the horde

#### Describe the solution

adding disassembly recipes for massive, large, medium, small and tiny engine blocks into lumps of steel

not doing anything
#### Testing

tested

#### Additional context
cataclysm-tiles_GErpxRoFfD

### DIFF
--- a/data/json/uncraft/generic.json
+++ b/data/json/uncraft/generic.json
@@ -465,7 +465,7 @@
     "activity_level": "MODERATE_EXERCISE",
     "time": "5 h",
     "tools": [ [ [ "angle_grinder", 500 ] ] ],
-    "components": [ [ [ "steel_lump", 100 ] ] ]
+    "components": [ [ [ "steel_lump", 95 ] ] ]
   },
   {
     "result": "engine_block_massive",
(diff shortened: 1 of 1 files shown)


Aeiouuu01234:
tru

Xaleth:
Steel engine blocks are extremely rare due to how impractical it is compared to the vastly more common and superior iron and aluminum materials.

Aeiouuu01234:
Aluminum blocks are rather expensive, could be found on sports cars or premium motors, but not in the majority of other engines, and due to the lack of raw materials for iron I went to the closest and most efficient alternative, because how much will be 100k of scrap metal?

Xaleth:
I'm not sure, but just saying that steel being the only material from engine blocks is not realistic, let alone it being common. I'd recommend doing a separate PR adding/incorporating iron, and I'm more than happy to help with that if it means removing steel from blocks in this PR.

Aeiouuu01234:
so i close this request?

Xaleth:
I'd mark it as a draft so that way iron can be worked on first. It'll be a while with the content freeze and all

PR #62714, opened 2022-12-12
Increase price of enchanted sledge hammers

AkkoRishtar:
#### Summary
None

#### Purpose of change
You shouldn't be able to obtain the weapons with the highest bashing in the game in exchange for a small amount of miscellaneous items you have no other uses for. You should at the very least look into acquiring valuables or part with some useful items to make such a purchase.

#### Describe the solution
Boosted price and price_postapoc to result in slightly more reasonable prices

Removing sledges from shop item pools, but I am not familiar with how item pools work.

#### Testing
Spawned forges of wonders until able to check the selling price of each changed item

#### Additional context
The normal heavy sledge hammer usually takes 2-4 in-game days to craft even if you already have the skills and recipe. Being able to buy a better version of it for less than 100$ just seems very unbalanced.

### DIFF
--- a/data/mods/Magiclysm/items/enchanted_melee.json
+++ b/data/mods/Magiclysm/items/enchanted_melee.json
@@ -68,7 +68,7 @@
     "id": "sledge_plus_one",
     "copy-from": "hammer_sledge",
     "name": { "str": "sledge hammer +1", "str_pl": "sledge hammers +1" },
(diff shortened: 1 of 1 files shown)


8street:
I may be wrong, but the price in json file is given in cents. So the prices must be multiplied by 100.

8street:
Your prices seem to be correct.

AkkoRishtar:
> Something like this should probably be done with the other high-tier enchanted weaponry, too.

That would be harder to balance for me because I'm not that familiar with it, but I'll look into it, I guess

PR #62715, opened 2022-12-12
Patch the Limeware Platter exploit

### LINKED ISSUE #62322: Getting caught stealing and dropping the item has no consequences which can be abused
### Describe the bug

You can steal something and drop it on demand over and over again without any consequences. By moving to a tile adjacent to where you dropped the stolen item and picking it up again, you can move the item tile by tile out of line of sight and get away with it even though you were caught many times.

### Steps to reproduce

1. Steal an item in the refugee center
2. Get caught, drop the item
3. Move to a tile towards the door, pick it up again
4. Repeat steps 2 and 3 until you can pick up the item outside of the room and outside of line of sight
5. Go drop the item outside of the room
6. Repeat until you have all the items outside of the room

### Expected behavior

Getting caught stealing, even if you drop the item when confronted, should probably lower the NPC opinion/trust of you/increase anger towards you by some amount.

It would also be cool to apply this to other NPCs in the same faction, who would likely hear about your attempts at theft - maybe on some delay as word gets around. Maybe NPCs who already have a high opinion of you could have a chance to not have their opinions lowered - choosing not to believe what they heard about you.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.22623.891 (22H2)
- Game Version: 0.F-12427-g021d32be79 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    No Hope [no_hope]
]

### Additional context

_No response_

MNG-cataclysm:
#### Summary
Bugfixes "Attempted theft has consequences"

#### Purpose of change
Prevent players from robbing NPCs with no consequences by stealing items and dropping them when prompted. Fix #62322. 

#### Describe the solution
NPCs will now have a slight increase in anger whenever you try and steal something, admit to it, and drop it. You can do this up to four times before they realize what you're up to and attack.

Not allowing players to steal something outright, or having an NPC attack the first time you try and steal something.

#### Testing
I went to the refugee center, attempted to rob the merchant, and was shot dead after four attempts to steal a glass pane.

#### Additional context
None.


RenechCDDA:
A morrowind reference? Nice...

Issue #62741, opened 2022-12-15
A save will be rendered unplayable if it autosaved during AIM moving all items from all surrounding areas


Kamayana:
### Describe the bug

If the game triggers an autosave while the Advanced Inventory Manager is moving all items from all areas, and then the game is force quit or quit via the window X (anything other than save and quit), that save file will become unplayable. The save will load up to a never ending and uncancellable "picking up..."/"moving items..." popup and become unresponsive.

### Steps to reproduce

1. Go into options and ensure that autosave is on. Can also set "Real minutes between autosaves" to 0 to remove waiting time later.
2. Start a new game in the default scenario.
3. Walk up to some item-loaded lockers in the shelter so you have a good number of items near you.
4. Open the advanced inventory manager with '/'
5. The default layout should open - left pane at Aim All and right pane at Inventory. This is fine, though you could also change the right pane into any of the other locations and the bug will still trigger.
6. Wait around long enough for the real minutes autosave timer to count down, which is only a couple seconds if it's set to 0. The game won't actually autosave in the AIM window yet, it waits until the next concrete action is taken.
7. Move all items from the left Aim All pane to the right, by pressing ','. You should see an autosave message during the movement.
8. Quit the game by clicking the window X.
9. Reopen the game and load the save.
10. Once it loads into the game world with a hanging popup. A few moments later Windows will tell me the game became unresponsive and needs to be force quit.

### Expected behavior

The same process doesn't break anything if the source pane is anything other than Aim All. Loading such an autosave will open up to the same frame Aim All does, but instead of becoming unresponsive it will actually complete the Move All action as expected and continue working normally. However, while this is obviously better, it does seem unexpected that it's autosaving _before_ the Move All action, rather than waiting until it's over.

### Screenshots

![image](https://user-images.githubusercontent.com/6759778/207864840-f88763fd-4e13-4950-9ce1-c162c9f87abb.png)
![image](https://user-images.githubusercontent.com/6759778/207863749-5a4bdcce-3a32-4e08-824e-b0cf0ad5d6ff.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.2251 (22H2)
- Game Version: 4c6e8b0 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
[...]

PR #62744, opened 2022-12-15
AIM: prevent infinite loop on move_all deserialization

andrei8l:
#### Summary
None

#### Purpose of change
The AIM can get stuck in an infinite loop if deserialized in the middle of a move_all process.
Fixes: #62741

#### Describe the solution
Add a sanity check

Nuking `move_all_items()`: this function is an ugly monster made up of at least 3 generations of legacy code held together with staples and it really, *really*, doesn't need to be this complicated and certainly not recursive. But that's for after 0.G /rantover

#### Testing
The save from the linked issue loads fine. Moving all items in all possible combinations still works.

#### Additional context
N/A

### DIFF
--- a/src/advanced_inv.cpp
+++ b/src/advanced_inv.cpp
@@ -939,6 +939,10 @@ bool advanced_inventory::move_all_items()
     if( spane.items.empty() || liquid_items == spane.items.size() ) {
         if( !is_processing() ) {
             popup( _( "No eligible items found to be moved." ) );
+        } else if( spane.get_area() != AIM_ALL ) {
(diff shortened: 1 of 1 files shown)


andrei8l:
@Kamayana please test this. You've been particularly good at uncovering AIM bugs.

Kamayana:
> @Kamayana please test this. You've been particularly good at uncovering AIM bugs.

Tested on all the saves I had that were hanging, all load just fine now, and I can't activate the bug again. Also tried doing move_all autosaves from singular locations and they loaded fine, too. Nothing changed into AIM_ALL that wasn't supposed to. Looks good to me!

PR #62745, opened 2022-12-15
Changing mil composite armor removal requirements 

SabrinaMori:
#### Summary
Balance "Change military composite armor removal requirements to power tool level metal sawing"
#### Purpose of change
Military composite armor consists of much stronger materials than a handmade or regular store hacksaw can reliably saw, so changing the requirements to a power tool level sawing would make more sense. I think it also fits better with the fact that you can't craft military composite armor, so it should make sense it would be removable only with a tool you can't craft 

#### Describe the solution
Added a new requirement in requirement/vehicle.json called "vehicle_weld_removal_cut_resistant" which requires level 3 of metal sawing, and updating the military composite armor to require that. 

Hacksaw theoretically would be able to saw trough the armor, but it would require adding a charge mechanic to be able to replace the sawing blade. To actually saw the armor you would have to replace the blade multiple times over and over, which would also require a break in the part deconstruction, which isn't possible with current implementation. Also for this new hacksaw mechanic to work it would have to somehow know the material of the thing it's sawing to appropriately use more of it's charge (to emulate faster blade dulling).

Another solution would be to make the process of sawing take longer or faster depending on the level of the tool you are using, which would still make it fairly unrealistic to be able to cut trough heat treated composite steel armor using a regular hacksaw.

#### Testing
It's a simple json change which doesn't affect any other elements and just changes recipy requirements

### DIFF
--- a/data/json/requirements/vehicle.json
+++ b/data/json/requirements/vehicle.json
@@ -41,5 +41,11 @@
     "type": "requirement",
     "//": "Counterpart to welding_* installation requirements",
     "qualities": [ { "id": "SAW_M", "level": 2 } ]
+  },
+    {
+    "id": "vehicle_weld_removal_cut_resistant",
+    "type": "requirement",
+    "//": "Used for parts that can't be cut with hand tools",
+    "qualities": [ { "id": "SAW_M", "level": 3 } ]
   }
 ]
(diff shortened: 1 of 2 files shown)


SabrinaMori:
I somehow deleted the summary, can someone add balance tag to this PR

Maleclypse (approved):

Fris0uman:
> you can't craft military composite armor, so it should make sense it would be removable only with a tool you can't craft

No that's the wrong logic, it should be removeable with something that can remove it irl. And that thing should be craftable if it is irl.

PR #62752, opened 2022-12-16
Make AIM minimap show an <A> for appliances

Kamayana:
#### Summary
Interface "Make AIM minimap show an A for appliances"

#### Purpose of change
When looking inside an appliance with the AIM the minimap displays a \<V\> for it. While this technically makes sense as grid-attached appliances are actually vehicles programming-wise, to a player it would seem odd.

#### Describe the solution
This change makes the minimap display an \<A\> instead of a \<V\> for any vehicle part with APPLIANCE flag set. The part will also need the CARGO flag for the AIM to be able to look inside in the first place, so this should cover most household appliances with storage. This notably does _not_ affect vehicle appliances, as none of them have the APPLIANCE flag. I didn't want to complicate things so I let that be, especially since the \<V\> still makes sense for appliances that are actually in a vehicle.

Using an \<F\> for \<F\>urniture. Adding a new flag to all of the vehicle appliances that'd make them show an \<A\> as well. Dragging a grid-attached appliance will still display "Grabbed Vehicle" in the AIM location header and in the popup when you try to move an item to itself. Changing that looks like it might be fairly complicated, for a situation that wouldn't happen very often, and the project is in string freeze right now anyway, so I decided not to bother.

#### Testing
Spawned a bunch of appliances and attached them to the grid, checked them in the AIM and saw the \<A\>s. Spawned an RV outside, checked its appliances and storage, saw the \<V\>s.


Night-Pryanik:
Please make both A and V translatable and add comments on how to translate them properly.

Kamayana:
> Please make both A and V translatable and add comments on how to translate them properly.

Done.

Qrox:
Is the hotkey to switch between vehicle/appliance and the ground changed as well? Making "A" and "V" translable will also make the hotkey unclear in other languages, I think.

Kamayana:
> Is the hotkey to switch between vehicle/appliance and the ground changed as well? Making "A" and "V" translable will also make the hotkey unclear in other languages, I think.

I have to admit I didn't actually know that was a hotkey, I've always been activating it by pressing the location multiple times. The 'A' hotkey is already used for All 9 Squares so now I'm wondering if this could cause confusion. At the very least I suppose the keybinding description should be changed to "Toggle vehicle/appliance".

PR #62756, opened 2022-12-16
Added knife qualities to knife spear

whatdice:
Added knife qualities to knife spear

#### Summary
Balance "Added knife qualities to knife spear in line with those on makeshift and simple knife spears."

#### Purpose of change

Seems a bit silly that upgrading the shaft quality of your knife spear removes it's ability to act as a knife, thus making the knife spear upgrade somewhat of a sidegrade. This PR fixes that.

#### Describe the solution

Added butchering and cutting qualities. This consequently adds the cut up, cauterize, and write on item actions.

Increasing butcher value, as the knife spear line as a whole feels like its butcher values are too low, however, that would be better as a separate PR.
Changing knife spear's material to iron, to match the other knife spears, or changing the other knife spears to steel for the same reason. I was concerned that distinction existed for a good reason, so I refrained from changing it for now.

#### Testing

Manually made the same change to release 2022-12-16-1130, loaded in and was able to write on an item with a knife spear without issue.

#### Additional context

Image of results:
image

### DIFF
--- a/data/json/items/melee/spears_and_polearms.json
+++ b/data/json/items/melee/spears_and_polearms.json
@@ -243,7 +243,7 @@
     "flags": [ "DURABLE_MELEE", "SPEAR", "REACH_ATTACK", "NONCONDUCTIVE", "SHEATH_SPEAR" ],
     "weapon_category": [ "POLEARMS", "SPEARS" ],
     "price": 16000,
-    "qualities": [ [ "CUT", 1 ], [ "COOK", 1 ], [ "BUTCHER", -18 ] ]
+    "qualities": [ [ "CUT", 1 ], [ "COOK", 1 ], [ "BUTCHER", -28 ] ]
   },
   {
     "id": "spear_homemade_halfpike",


whatdice:
Not sure how to get past this test fail, or even what is means
https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/3716657867/jobs/6304848291#step:17:340
Any suggestions would be welcome

PatrikLundell:
Why would a better attachment of a knife to a pole make the resulting item better at butchering? Attaching a long pole to the end of a knife would make the knife very unwieldy for any knife activities regardless of the quality of the pole attachment.

whatdice:
> Why would a better attachment of a knife to a pole make the resulting item better at butchering? Attaching a long pole to the end of a knife would make the knife very unwieldy for any knife activities regardless of the quality of the pole attachment.

I lowered the butchering value down to the makeshift knife spear's level, while I'm not sure I like the value where it's at now, it can wait to be increased in a future PR that addresses the knife spear family as a whole.

PR #62760, opened 2022-12-17
Added a way to change color for 'Explored' tiles

Night-Pryanik:
#### Summary
Interface "Added a way to change color for 'Explored' tiles"

#### Purpose of change
I can't find it right now, but I'm pretty sure someone asked for a way to change default color for `Explored` overlay on overmap from dark gray to something else.

#### Describe the solution
- Added `explored` dummy overmap tile with a default dark gray color.
- Added a check for color of this OMT in `draw_ascii` function.

None.

#### Testing
Changed color of `Explored` overlay to yellow, opened overmap, marked some tiles as Explored, checked their color.

#### Additional context
изображение

### DIFF
--- a/data/json/overmap/overmap_terrain/overmap_terrain.json
+++ b/data/json/overmap/overmap_terrain/overmap_terrain.json
@@ -17,6 +17,14 @@
   },
   {
     "type": "overmap_terrain",
+    "id": "explored",
+    "name": "Explored",
+    "sym": "#",
+    "color": "dark_gray",
+    "flags": [ "NO_ROTATE", "SHOULD_NOT_SPAWN" ]
+  },
+  {
+    "type": "overmap_terrain",
     "abstract": "generic_city_building_no_sidewalk",
     "name": "city building",
     "sym": "^",
--- a/data/json/overmap/overmap_terrain/overmap_terrain.json
+++ b/data/json/overmap/overmap_terrain/overmap_terrain.json
@@ -1,30 +1,6 @@
 [
   {
     "type": "overmap_terrain",
-    "id": "",
-    "name": "nothing",
-    "sym": "%",
-    "color": "white",
-    "flags": [ "NO_ROTATE" ]
-  },
-  {
-    "type": "overmap_terrain",
-    "id": "unexplored",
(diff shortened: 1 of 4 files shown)


src/overmap_ui.cpp
```diff
@@ -601,7 +602,8 @@ static void draw_ascii(
         // Ok, we found something
         if( info ) {
             const bool explored = show_explored && overmap_buffer.is_explored( omp );
-            ter_color = explored ? c_dark_gray : info->get_color( uistate.overmap_show_land_use_codes );
+            ter_color = explored ? oter_explored.obj().get_color() :
```
ZhilkinSerg, line 605: I did not profile this, but it is very likely that it would have less overhead and be more performant if explored color would be stored in variable of `nc_color` type that would be assigned only once in finalize step of overmap terrain data load (i.e. `overmap_terrains::finalize()`) instead of accessing `oter_explored.obj().get_color()` for each overmap tile.

data/json/overmap/overmap_terrain/overmap_terrain.json
```diff
@@ -15,6 +15,14 @@
     "color": "dark_gray",
     "flags": [ "NO_ROTATE", "SHOULD_NOT_SPAWN" ]
   },
+  {
+    "type": "overmap_terrain",
+    "id": "explored",
```
anothersimulacrum, line 20: I think this should go in data/core, it's something the game always expects to exist.
We don't enforce anything going there, so it's not too important, though.

Night-Pryanik: Done in [816f430](https://github.com/CleverRaven/Cataclysm-DDA/pull/62760/commits/816f430435240c1545c8c98c2ac66db0c5c22ebb).

Zireael07:
Why close?

Night-Pryanik:
That's a ragequit because I can't work with tyrants anymore.

Raikiri:
interesting mechanic, i think it's useful

Zireael07:
It's an accessibility feature for players with poor vision, I hope someone picks it back up fast.

PR #62763, opened 2022-12-17
Fix crash on loading autosave made during insert action

Kamayana:
#### Summary
Bugfixes "Fix crash on loading autosave made during insert action"

#### Purpose of change
If the game autosaves during an insert action (both AIM and the normal insert action) and you attempt to load it, you will get a debug message about an item_location losing its target item. Continue and the game crashes.

#### Describe the solution
Added a null pointer check to the insert activity. The debug message will still show up, but you can continue past it without crashing.

Fixing what's causing the item_location to lose its target. I'm currently looking into that but it's probably going to be part of a bigger solution. This should work as a stopgap.

#### Testing
Load the attached save in the master branch, crashes after the debug message. Added my code and loaded the save, was able to continue.

#### Additional context
[autosaved during insert.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10252401/autosaved.during.insert.zip)

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -3962,7 +3962,7 @@ void insert_item_activity_actor::finish( player_activity &act, Character &who )
     }
 
     items.pop_front();
-    if( items.empty() || !success ) {
+    if( items.empty() || !success || items.front().first == item_location::nowhere ) {
         handler.handle_by( who );
         act.set_to_null();
         return;


Kamayana:
Not too sure why this failed the Basic Build and Test, it doesn't appear like item inserts are a part of it, and that's all this commit changes.

PR #62778, opened 2022-12-19
Lower recipe spear shaft recipe difficulty and change quality requirements.

dino217:
#### Summary
Balance "Lessen spear shaft recipe difficulty"

#### Purpose of change

[PR 61482](https://github.com/CleverRaven/Cataclysm-DDA/pull/61482) made several needed changes to spears. However this included making the fabrication difficulty for spear shafts unrealistically difficult.

#### Describe the solution

- Sets the spear shaft fabrication difficulty to 2 from 5 which is in line with the bo. 
- Changes the required tool qualities to also be the same as the bo. 
- Removes the book learning as you will have auto-learned the recipe before the books would have been helpful.

I considered leaving the required tool qualities as is. I decided to make the change because the only reason I can think of for needing wood sawing is for shortening the long stick to the desired length. While a saw would be useful, I don't think it would be required when you could instead cut a notch in the stick and break it to the desired length.

#### Testing

I loaded the game and saw that the recipe was learned at level 2 fabrication and was as expected.

#### Additional context

Here's a [YouTube video of someone making a spear shaft](https://youtu.be/K0kMIIKDwhY). As you can see, it's not a terribly difficult process. It's just a carved, polished, and oiled stick.

### DIFF
--- a/data/json/recipes/other/parts_construction.json
+++ b/data/json/recipes/other/parts_construction.json
@@ -336,7 +336,12 @@
     "autolearn": true,
     "proficiencies": [ { "proficiency": "prof_carving", "time_multiplier": 1.25, "max_experience": "100 m" } ],
     "//2": "Max experience set because waiting for oil to soak in isn't really teaching you anything.  Remove this when crafting time reduced to only carving time.",
-    "qualities": [ { "id": "CUT", "level": 2 }, { "id": "SAW_W", "level": 1 }, { "id": "CUT_FINE", "level": 1 }, { "id": "FILE", "level": 1 } ],
+    "qualities": [
+      { "id": "CUT", "level": 2 },
+      { "id": "SAW_W", "level": 1 },
+      { "id": "CUT_FINE", "level": 1 },
+      { "id": "FILE", "level": 1 }
+    ],
     "components": [
       [ [ "stick_long", 1 ] ],
       [


oosyrag:
The saw isn't for cutting the stick to the correct length. You can indeed already break a long stick into 2 sticks with just the use action. The tools for shaping and carving a long stick into a balanced and even surfaced staff suitable for a weapon, the part where the guy in the video uses a circular saw, router, and hand planer.

Basically have you ever seen a 6+ foot long branch that is nice and round and of even diameter the whole length on a tree that just needs it's two ends cut off to be a perfectly shaped staff?

I'd say this would be a case where the Bo could use additional tooling to match the spear shaft, not the other way around.

dino217:
Okay, I suppose that's fair. Do you think the fine cutting and filing qualities should be kept, or should I just revert to how it was?

oosyrag:
Personally, I think filing and sanding are necessary, so I'd say filing at least.

Cutting I imagine would be useful for carving the spear tip (in lieu of a handheld planer), but that's not exactly a required step for the spear shaft or staff by itself.

dino217:
Okay. I put the wood sawing requirement back. I think the fine cutting would still be required for carving the tapered end of the spear shaft so I left it and the filing requirements.

Aerin-of-the-Toast:
Author of the original PR here. I mostly picked 5 because it was just the difficulty for the long pole craft. I agree 5 is probably too high, though I think 2 is too low (for the bo as well) since 2 fabrication is still considered "novice" level. I think 3 or 4 would make more sense. 

I do like the addition of the wood sawing and filing qualities though, those probably should have been in from the start.

PR #62788, opened 2022-12-19
fix: updated f_glass_fridge to be more useful as a fridge

### LINKED ISSUE #62710: Glass door fridges should work like regular fridges.

The glass door fridges (`f_glass_fridge`) that appear in grocery stores, gas stations, what have you, are literally just fridges with an insulated glass door, yet the player can't make use of them like a regular fridge. They serve as obstacles and sources of scrap but not much else.

### Solution you would like.

The glass door fridge should be installable as an appliance like the common fridge is.

If there's a real reason why this isn't a thing, the description should be changed to point it out.

### Additional context

_No response_

clark-pegg:
#### Summary
Bugfixes "Added new glass_fridge item and appliance"

#### Purpose of change
Fixes #62710

#### Describe the solution
Makes glass fridges more useful by allowing them to be installed and used as an appliance just like a regular fridge.

I considered using the pre-existing fridge item, however losing the glass from the glass_fridge when deconstructing would've caused a loss of continuity.

#### Testing
I entered a game in debug mode and went to a grocery store, there, I tested deconstructing the f_glass_fridge. I then installed the glass_fridge as an appliance and hooked it up to power to make sure it worked. I also tested disassembling a glass_fridge item to make sure that worked.

#### Additional context
Added the EASY_DECONSTRUCT to f_glass_fridge to keep in-line with f_fridge. Also if there's no reason why f_fridge has the MINEABLE tag and f_glass_fridge doesn't, then it could be added to this PR.


clark-pegg:
> There will probably be confusion if you have a glass fridge irreversibly transform into a normal one. They can look different for one, besides having different flags and materials.

You bring up a fair point, so I added the glass_fridge item and appliance so they would stay different. I also added a disassemble recipe for the glass_fridge item so that you could recover the glass_shards. Didn't add a recipe to craft it though, doesn't seem like there's precedence for large scale glass crafts like what would be needed here.

data/json/furniture_and_terrain/appliances.json
```diff
@@ -34,6 +34,23 @@
     ],
     "damage_reduction": { "all": 32 }
   },
+  {
+    "type": "vehicle_part",
+    "id": "ap_glass_fridge",
+    "name": { "str": "glass fridge" },
+    "looks_like": "f_glass_fridge",
+    "description": "A modern refrigerator with a thick sheet of glass in the door, often specially treated to be more insulative.  Allows seeing the contents without letting out the cold air, which used to be a minor convenience, and now saves precious minutes until spoilage.",
```
Fris0uman, line 42:
```suggestion
    "description": "A modern refrigerator with a thick sheet of glass in the door, often specially treated to be more insulative.  Allows seeing the contents without letting out the cold air.",
```
I don't think we want to acknowledge in game that things instantly turn spoiled from one second to the next

data/json/items/appliances.json
```diff
@@ -13,6 +13,13 @@
     "weight": "70 kg",
     "pocket_data": [ { "pocket_type": "CONTAINER", "rigid": true, "max_contains_volume": "500 L", "max_contains_weight": "400 kg" } ]
   },
+  {
+    "type": "GENERIC",
+    "id": "glass_fridge",
+    "name": { "str": "glass door fridge" },
+    "description": "A modern refrigerator with a thick sheet of glass in the door, often specially treated to be more insulative.  Allows seeing the contents without letting out the cold air, which used to be a minor convenience, and now saves precious minutes until spoilage.",
```
Fris0uman, line 20:
```suggestion
    "description": "A modern refrigerator with a thick sheet of glass in the door, often specially treated to be more insulative.  Allows seeing the contents without letting out the cold air, which used to be a minor convenience.",
```
Same

data/json/recipes/recipe_deconstruction.json
```diff
@@ -7400,5 +7400,25 @@
...
+      [ [ "condensor_coil", 1 ] ],
+      [ [ "evaporator_coil", 1 ] ],
+      [ [ "cable", 3 ] ],
+      [ [ "hose", 1 ] ],
+      [ [ "pipe_fittings", 6 ] ],
+      [ [ "thermostat", 1 ] ],
+      [ [ "motor_small", 1 ] ],
+      [ [ "refrigerant_tank", 1 ] ],
+      [ [ "plastic_chunk", 16 ] ],
+      [ [ "glass_shard", 100 ] ]
```
Fris0uman, line 7421: maybe you should get an actual glass panel of the appropriate size?

Maleclypse, line 7421:
```suggestion
      [ [ "glass_sheet", 2 ] ]
```

PR #62790, opened 2022-12-20
rigid plastic sheet fix length

flybye3o6:
#### Summary
Bugfixes "set a length for rigid plastic sheet"

#### Purpose of change
rigid plastic sheet has a length of 23cm

#### Describe the solution
set rigid plastic sheet length to 122cm same as wooden panel and sheet of glass


alp7292:
cant you just bend it to fit backpack

flybye3o6:
> cant you just bend it to fit backpack

it is 15kg and 12L, at 23cm it is a dense cube, its only use ingame is to make a window
image
image

PR #62794, opened 2022-12-20
Rooftop fridge/freezer stowing

Drew4484:
#### Summary
Content "Mount fridge/freezer as rooftop cargo."

#### Purpose of change
Without a larger storage option, players are holding refrigerators and freezers in their hands while driving. This is ridiculous for a long list of reasons.

#### Describe the solution
Adds two new vehicle parts, a strapped to the rooftop fridge and freezer. Requires the appliance and some rope to tie it down.

Larger storage capacity tiles, some kind of roof rack storage for large bulky items.

#### Testing
Spawned refrigerator, freezer, ropes. Crafted tied-down versions, attached to roof of vehicle. Conflicts with other roof-mounted parts appropriately. Uninstalled and disassembled into fridge/freezer/rope.

#### Additional context
There are probably more appliances that could use this treatment.


flybye3o6:
i was working on something like this awhile ago but i gave up when i found out you can use bike racks
you need to place them as an appliance on the ground next to the bike rack then you can attach it
image
this is still good though i like this, mite need to do it for washing machine and other appliances

Drew4484:
> i was working on something like this awhile ago but i gave up when i found out you can use bike racks you need to place them as an appliance on the ground next to the bike rack then you can attach it image this is still good though i like this, mite need to do it for washing machine and other appliances

That works, but it still feels very clunky. This isn't a perfect solution but it seems like something that a character would naturally want to try doing IRL.

flybye3o6:
the way i was tying to do it was to make a storage part with very large volume but limit it to 1 item, couldn't do it so tried minimum item size, also couldn't do it and realized you cant have more then 1 cargo part per tile so gave up on it
also really wanted to have it so you could put large cardboard boxes on the roof and store stuff in them, still wanna try and figure out some way of doing that without it being a cargo part

PatrikLundell:
Fake vehicle parts that are really intended to just be transported goods shouldn't integrate with the vehicle. Have you verified that it's not possible to turn these "vehicle parts" on? Do they allow storage of items inside of them (I don't think it's important if they do, but knowing how they behave is useful)?

Drew4484:
> Fake vehicle parts that are really intended to just be transported goods shouldn't integrate with the vehicle. Have you verified that it's not possible to turn these "vehicle parts" on? Do they allow storage of items inside of them (I don't think it's important if they do, but knowing how they behave is useful)?

They don't turn on or act as storage.

flybye3o6:
another thing this method could be used for is bundles, could attach bundles of planks/sticks to the roof or hang off the side of a vehicle

PR #62800, opened 2022-12-20
Napkins and Cardboard are no longer comestible

Karol1223:
#### Summary
Bugfixes "Napkins and Cardboard are no longer comestible"

#### Purpose of change
Napkins and Cardboard were classified as comestible despite not actually being edible. Not only does this add useless lines of code, it also means they'd clutter the 'E' menu on top of that. May or may not have caused additional lag with them freezing, though I'm not staying on top of what actually causes lag and what doesn't.

#### Describe the solution
Changed TYPE from COMESTIBLE to GENERIC and removed all references to them being comestible from the code. Also apparently removed charges from them because code yelled at me very rudely.

### DIFF
--- a/data/json/obsoletion/charge_removal.json
+++ b/data/json/obsoletion/charge_removal.json
@@ -31,7 +31,9 @@
       "sheet_nomex",
       "sheet_nomex_patchwork",
       "sheet_nylon",
-      "sheet_nylon_patchwork"
+      "sheet_nylon_patchwork",
+      "napkin",
+      "cardboard"
     ]
   },
   {


MNG-cataclysm (approved):

Karol1223:
Yeah... definitely NOT ready for approval lol. The spawning somehow got fucked, I am not sure why but it doesn't appear to be ready just yet.

mqrause:
https://github.com/CleverRaven/Cataclysm-DDA/blob/4179a45695f8b53b2b019b39edd2ff90a1361d7b/data/json/obsoletion/charge_removal.json#L3

You should add those to this list as to not destroy peoples stashes.

Karol1223:
> https://github.com/CleverRaven/Cataclysm-DDA/blob/4179a45695f8b53b2b019b39edd2ff90a1361d7b/data/json/obsoletion/charge_removal.json#L3
> 
> You should add those to this list as to not destroy peoples stashes.

Good and done. Apologies for the long time it took me to do that, I wasn't able to get on my computer for the past few days. Hopefully I understood what you asked of me correctly.

Karol1223:
Aaaaaaaaaaaand here we go again with the map test failure that I haven't touched. Well, if anyone's reading this and would be so kind as to kick the test I'd appreciate it. I have no idea why the map tests keep bullying me but I think they're winning.

PR #62804, opened 2022-12-21
Make the Chemist shop use the trade interface

onura46:
#### Summary
Bugfixes "Make the Chemist shop use the trade interface"

#### Purpose of change
Make the Chemist shop use the trade interface

#### Describe the solution
This PR replaces the dialog-based trade system in the Chemist shop. Having worked on the previous system, I discovered a bug whereby attempting to buy something without having items to trade would cause the dialog option to disappear regardless. This change would fix that and also remove the fixed pricing from that menu, instead opting to use `shopkeeper_price_rules` and premiums. Since their inventory refreshes at the default rate (6 days iirc), I attempted to balance availability of shop items against their old refresh rates.

A typical store inventory, once the player has brought the Chemist proper supplies:
Capture

#### Testing
Many, many reloads and checks of this pricing against their old prices. Price premiums should hopefully be realistic for someone able to synthesize a few chemicals, but still scavenging for most resources.

#### Additional context
The same thing needs to be done for the lumbermill employees (`data\json\npcs\lumbermill_employees`) and the homeless encampment (`data\json\npcs\other\homeless\group_camp`).

### DIFF
--- a/data/json/mapgen/chemist_holdout/chemical_lab.json
+++ b/data/json/mapgen/chemist_holdout/chemical_lab.json
@@ -33,7 +33,7 @@
       ],
       "palettes": [ "cabin_palette" ],
       "faction_owner": [ { "id": "wasteland_scavengers", "x": [ 1, 23 ], "y": [ 1, 23 ] } ],
-      "items": { "C": [ { "item": "cabin_chemist_sell", "chance": 100, "repeat": [ 4, 5 ] } ] },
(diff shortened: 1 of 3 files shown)


onura46:
It looks like #63064 is likely to address the same bug that this addressed in a simpler way. This draft was never finished and still lacks a little of the time-based logic of the old system where you place orders with the chemist rather than go through the storefront like this PR.

I don't have time to figure out how to do the EOCs needed, and I don't want to hold up the other PR, so I'm going to close this. But I would still prefer to move towards this storefront system since it's more future-proof.

PR #62807, opened 2022-12-21
Made appliance freezer display as furniture freezer rather than a symbol

AkkoRishtar:
#### Summary
None

#### Purpose of change
Fix: Stops the freezer from remaining a symbol despite tileset

#### Describe the solution
Since the furniture freezer and the appliance fridge look like the furniture fridge, I made the appliance freezer look like it too

n/a since this is a fix

#### Testing
loaded game, looked at freezer

#### Additional context
none

### DIFF
--- a/data/json/furniture_and_terrain/appliances.json
+++ b/data/json/furniture_and_terrain/appliances.json
@@ -45,6 +45,7 @@
     "id": "ap_freezer",
     "name": { "str": "freezer" },
     "symbol": "H",
+    "looks_like": "f_fridge",
     "categories": [ "cargo" ],
     "color": "blue",
     "broken_symbol": "#",


PatrikLundell:
I like the change, but not the title. The title just says "did something unspecified to file X", while a better title/description would be something like "Made appliance freezer display as furniture freezer rather than a symbol".

The automatic tests rejects your Purpose of change contents because it has to rigidly follow the format described in the template comment as it's used by a script to generate change log entries. This format mandates the usage of one of the approved tokens followed by a string in quotes, or the token None with nothing following it.
I generally use None, since my PR tend not to be important enough to clutter change logs.

AkkoRishtar:
> I like the change, but not the title. The title just says "did something unspecified to file X", while a better title/description would be something like "Made appliance freezer display as furniture freezer rather than a symbol".
> 
> The automatic tests rejects your Purpose of change contents because it has to rigidly follow the format described in the template comment as it's used by a script to generate change log entries. This format mandates the usage of one of the approved tokens followed by a string in quotes, or the token None with nothing following it. I generally use None, since my PR tend not to be important enough to clutter change logs.

Copy pasted the title, but I had already put "None" as the summary, if I did it wrong, I don't know how because I followed instructions to the letter

PR #62814, opened 2022-12-22
ASRG adjustments

anoobindisguise:
#### Summary
Balance "Adjustments to ASRGs"

#### Purpose of change
ASRGs were too big and heavy. My original plan for them split them into the bare minimum (1:1 with NASA specs) and a super heavily armored one that made it into the game. However it's just too big and too heavy. My thinking was that it was made super armored by the military to make it resistant to zombies/portals/etc as they were deployed en masse before the cataclysm (plutonium sourced from XEDRA/the nether making them cheap) but the military is swapping over to diesel generators so these are becoming almost entirely a XEDRA thing.

#### Describe the solution
The ASRG by default is 39x46x76cm, though its actual displacement is not exactly this much due to it not being a perfect square, and it is only 32 kg. I'm assuming they're still being shielded a *little* so I've redesigned them to have a milcomp bounding box giving them about 150 liters total (down from 200) and 100 kg (68 kg of milcomp plating) and reduced their durability a bit. 
I also let them work in cars, because I originally thought they were incompatible with cars due to the appliance weirdness but it turns out perpetual reactors work perfectly in cars! So now you can provide 0.1% of your total necessary electric motive power with an ASRG if you want!

Remove the armor plating entirely, go with ONLY the NASA specs and make them rather fragile to car crashes/zombies/etc. However the Ultica/MSX/UDP/etc sprites depict a pretty large box so I don't wanna force tileset authors to make a new sprite for it.

#### Testing
I tested and confirmed no errors + everything loads up and worked perfectly, spawned an electric car and saw +130w per ASRG I installed in it.


anoobindisguise:
Looks like basic build failure isn't actually mine.

irwiss:
May be remove CARGO flag and size field on the vpart definition while you're there, or is there a reason you can place items in it?

anoobindisguise:
> May be remove CARGO flag and size field on the vpart definition while you're there, or is there a reason you can place items in it?

Only the appliance has it, I think the reason was that you could stack stuff on top of it since it's not moving around a lot. Not super necessary though, I'll do away with it.

data/json/items/appliances.json
```diff
@@ -327,11 +327,13 @@
     "id": "compact_ASRG_containment",
     "name": { "str_sp": "Compact Advanced Stirling Radioisotope Generator" },
     "looks_like": "t_plut_generator",
-    "description": "A really heavy brick of metal housing an ASRG, formerly used by NASA and repurposed by the military.  Will provide about 130w, basically forever as far as you're concerned.  Should be safe due to the large amount of shielding.  Although these were designed for spacecraft, the sheer amount of armor plating and lead means it is much more massive, so you probably shouldn't try to power a vehicle with it.",
-    "weight": "400 kg",
+    "description": "A rather heavy brick of metal housing an ASRG, formerly designed by NASA and repurposed for terrestrial use.  Will provide about 130w, basically forever as far as you're concerned.  Although the alpha radiation these run off of isn't normally very dangerous, this unit appears to be armored to take a beating and is a good deal heavier owing to its composite plating.",
```
TheShadowFerret, line 330: I don't think this LEAK_DAM is justified if it's releasing _alpha_ radiation when cracked.
(Assuming nobody tries to eat it)

RenechCDDA:
The added vehicle part `ap_compact_ASRG_containment` has the same id as the appliance, causing it to overwrite the appliance's values. This is causing some rather strange, erroneous behavior. Just need to bump the vehicle part's ID out to a unique one.

PR #62819, opened 2022-12-22
Replace integrated toolset with an integrated armor

anoobindisguise:
#### Summary
Content "Integrated Multitool uses an integrated armor"

#### Purpose of change
Integrated Multitool was clunky and was a good candidate for an integrated armor replacement.

#### Describe the solution
Rather than give you a fake hidden toolset with tool qualities and a toggleable toolset that takes up your hands it now simply installs an integrated armor in your hands. This means you no longer have to activate the bionic in the bionics menu, then additionally activate the item in your hands, you can now just use the item in your inventory. This does mean you can't turn the toolset's functions off but I have no idea why you'd want to do that anyways. I also lowered the arms slot usage a little because most of the stuff is in your hands.

This makes no attempt to get rid of the portable welder which may be desirable at some point although I also have plans to adjust weld costs such that a welder running of your bionic energy will be impractical. Even then though you'd probably want it as a last resort backup.

#### Testing
Confirmed it loaded up OK in game and seemed to be functional. I couldn't find any uses of the toolset that would require me to hold it in my hands, everything seems to work as just a piece of armor and the toolset has USE_BIONIC_POWER to make sure it works off your energy correctly.


anoobindisguise:
why tf is "fabrication" no longer a valid skill? this is literally copypasted from every other welder-type tool. what.

andrei8l:
> why tf is "fabrication" no longer a valid skill? this is literally copypasted from every other welder-type tool. what.

It's invalid only in `obsolete.json` because it's loaded before `skills.json`.

anoobindisguise:
For the longest time it was complaining about the one in integrated.json too. weird.

PR #62831, opened 2022-12-22
fix: Recipe makes: X wasn't displayed in crafting GUI for non-charge items

### LINKED ISSUE #61753: Crafting Neoprene Sheet Yields 9, but Recipe Indicates Yield of 1
### Describe the bug

I was surprised to see that, upon finishing crafting the neoprene sheet recipe, that nine units of neoprene sheet were created.  Given that the recipe doesn't indicate that it makes anything other than a single unit (i.e., no "recipe makes" quantity nor any byproducts shown), I was expecting crafting the recipe to create only a single unit of neoprene sheet.

image

### Steps to reproduce

1. Start crafting neoprene sheet recipe.
2. Upon finishing the crafting, check that nine units of neoprene sheet were created as a result.

A few things to note:
- I tried crafting the recipe twice, and both times resulted in nine units each.
- The recipe uses applied science as its primary skill with a difficulty level of 5.  I only tried this on my character with an applied science level of 4.  When I did the two crafting attempts, one failure was encountered during each.  Not sure if the results could be different if no failures were encountered.
- Here are the specific components I used:
image

### Expected behavior

Either:
1. crafting the neoprene sheet recipe once should result only in one unit of neoprene sheet being created, or
2. the neoprene sheet recipe should be updated (probably via whatever controls the "recipe makes" value) so that it shows that the result of crafting the recipe one time will be nine units.

### Screenshots

_No response_

- OS: Windows
- Game Version: 4bf0543 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Slots [cbm_slots]
]

### Additional context

_No response_

### LINKED ISSUE #62823: Plank recipes "recipe makes X" are not displayed
### Describe the bug

Plank recipes (long plank -> plank -> short plank) suposedly make multiple and it is in the JSON, but player cannot see it as "Recipe makes" is missing in the recipe.

### Steps to reproduce

1. look in crafting menu at recipes for plank and short plank

### Expected behavior

I can always see how many a recipe makes.

### Screenshots

image

- OS: Windows
- Game Version: ab60b1b [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

From discord:
> I was going to say it might only exist for food but that's wrong, I don't see a reason why it doesn't display how many of the thing you get.  Presumably the "result_mult" means that it does only make 1 it just multiplies that and gives it to the player.  So it's possible that using result_mult circumvents whatever mechanism tells the player how many you get

Brambor:
#### Summary
Bugfixes "Crafting GUI: show how much recipe makes for non-charge items"

#### Purpose of change

Fix #61753
Fix #62823

#### Describe the solution

Adjust implementation of a function to consider `recipe_makes` for non-charge items.

None

#### Testing

A recipe shouldn't have both: charges and recipe_makes != 1. I hope that's true, but didn't check.

Still working
<details>
<summary>Seasoned Salt</summary>

image
</details>

Fixed items "Recipe makes" weren't shown here before
<details>
<summary>Neoprene sheet</summary>

image
</details>
<details>
<summary>Planks</summary>

image
image
image
</details>

#### Additional context

The changed function is used when uncrafting recipes. I wonder if it fixed something there.
https://github.com/CleverRaven/Cataclysm-DDA/blob/af92bd266c9a01a62b5e00fee9d85f3fc47fed26/src/iuse_actor.cpp#L1464

Not sure, if this is related:
https://github.com/CleverRaven/Cataclysm-DDA/blob/af92bd266c9a01a62b5e00fee9d85f3fc47fed26/src/recipe.cpp#L587

### DIFF
--- a/src/recipe.cpp
+++ b/src/recipe.cpp
@@ -1216,14 +1216,13 @@ bool recipe::hot_result() const
 
 int recipe::makes_amount() const
 {
-    int makes = 0;
+    int makes_charges = 1;  // stays 1 if item isn't counted in charges
     if( charges.has_value() ) {
-        makes = charges.value();
+        makes_charges = charges.value();
     } else if( item::count_by_charges( result_ ) ) {
-        makes = item::find_type( result_ )->charges_default();
+        makes_charges = item::find_type( result_ )->charges_default();
     }
-    // return either charges * mult or 1
-    return makes ? makes * result_mult : 1 ;
+    return makes_charges * result_mult;
 }
 
 void recipe::incorporate_build_reqs()
--- a/src/recipe.h
+++ b/src/recipe.h
@@ -280,7 +280,7 @@ class recipe
 
         bool removes_raw() const;
 
-        // Returns the amount or charges recipe will produce.
+        // Return the amount the recipe will produce (be it charges, or whole items).
         int makes_amount() const;
 
     private:


mqrause:
> Not sure, if this is related:

Kind of, but I want to mention that I'm reworking pretty much all of this in #60885, just in case you're considering working on this some more.

Brambor:
> Kind of, but I want to mention that I'm reworking pretty much all of this in #60885, just in case you're considering working on this some more.

Ok, I'm leaving it up to you, I just wanted to fix this simple but influential bug :)

src/recipe.cpp
```diff
@@ -1216,14 +1216,15 @@ bool recipe::hot_result() const
 
 int recipe::makes_amount() const
 {
-    int makes = 0;
+    int makes_charges;
```
mqrause, line 1219: You should just initialize this to 1 and remove the else case.

Brambor: I thought this is more clear

Brambor: Ok, did it your way :)

Issue #50248, opened 2021-07-27
Replace flamethrower with thermobaric weapons


eJ0opb:
# Is your feature request related to a problem? Please describe.
Flamethrowers haven't been used by the US military since 1978.  They have been replaced with [thermobaric weapons](https://en.wikipedia.org/wiki/Thermobaric_weapon).

# Describe the solution you'd like
Replace flamethrowers with weapons currently used by the US military, such as the [SMAW-NE](https://en.wikipedia.org/wiki/Mk_153_Shoulder-Launched_Multipurpose_Assault_Weapon).

# Describe alternatives you've considered
* Pretend flamethrowers as depicted in popular culture are still being used by the US military in this timeline
* Make flamethrowers represent commercial flamethrowers and change stats/loot/spawns accordingly

# Additional context
Don't blame me for taking away the toys; the US DOD unilaterally decided on this in 1978.

RenechCDDA:
AFAIK, the M202 FLASH (used by Talon UGVs ingame) is a thermobaric weapon, and likely the only man-portable weapon that would qualify.

I'm not sure that replacing flamethrowers would necessarily be appropriate though, they're already somewhat rare spawns in military outposts/FEMA camps, and the FLASH just doesn't occupy quite the same role in a zombie scenario. 

I think the government could, pretty quickly, dust off some old equipment and concepts to reintroduce proper flamethrowers since they would be more appropriate than incendiary rockets. We already do have mass graves and other examples of the government's ad-hoc response.

DalenTalas:
The military tends to keep their old stuff in reserve, even if it's obsolete. Just in case. Breaking out the flamethrowers from storage in case of a zombie apocalypse seems like an acceptable break from reality to me.

Furthermore, as The Wiki tells us, they are also in use for civilian applications like agriculture and forestry. And it's not unexpected that enterprising survivors could put one together.

My proposal:
- Keep the flamethrowers.
- Consider adding them as rare drops in civilian locations (farms, forestry areas of some sort).
- Add thermobaric weapons for extra !!!fun!!!

eJ0opb:
> I'm not sure that replacing flamethrowers would necessarily be appropriate though, they're already somewhat rare spawns in military outposts/FEMA camps, and the FLASH just doesn't occupy quite the same role in a zombie scenario.

[...]

PR #62833, opened 2022-12-23
Remove spawn of zombie with a flamethrower.

### LINKED ISSUE #62252: Zombie Burners are kind of dumb

Zombie Burners are dead flamethrower soldiers. Their tank is described as leaking, and it in fact has a limited capacity of napalm it can leak before it stops leaking entirely. In addition to this, they have a rare chance to explode in a lethal conflagration when killed. 

This obviously doesn't make a whole lot of sense. Why do they explode? Does "Blob Magic" ignite the napalm when they die? Napalm tanks exploding is a hollywood/videogame fabrication first of all, it doesn't really conform to real life. Secondly, there is a good chance the napalm tank is entirely empty by the time they would die, but they still explode anyways. Third, the napalm tanks only start leaking once the zombies are in your reality bubble, when in reality their tanks should have emptied within 10 minutes of them resurrecting as a zombie.

Them exploding is very much a "gotcha" mechanic, the player has no way to reasonably expect that they will explode into flames when killed because of how improbable and unrealistic it is. And one explosion can very easily kill a player in melee especially if they are new enough to the game that they don't know that these enemies can erupt into flames on death.

### Solution you would like.

1. Zombie burners don't leak napalm when moving.
2. They don't explode when killed.

Remove zombie burners.

### Additional context

_No response_

kevingranade:
#### Summary
None

#### Purpose
Flamethrowers aren't fielded by the modern military or police or pretty much anyone. It's not reasonable to run across a zombie wearing one, much less with all the baggage this implementation brings with it (infinite fuel supply that disappears when they die?)

Partially adresses #50248 
Fixes #62252 

#### Implementation
Removes remaining spawns of flamers, moves them to obsolete holding file to avoid breaking saves and mods.


TheShadowFerret:
The Military Corpse Disposal(which I made to be the profession version of the burner) needs to go too

oosyrag:
Possibly fielded as a specific counter to the opposition?

Night-Pryanik:
Flamethrower units were sort of a desperate action by the government when they realized that zombies ain't killed by firearms, it's noted in their description. These units are not a part of organized military or police units.

Night-Pryanik:
If you're still inclined to get rid of it, please move it to No Hope mod instead. I don't want my hard work be thrown out like this.

PapaOscarWhiskey:
> If you're still inclined to get rid of it, please move it to No Hope mod instead. I don't want my hard work be thrown out like this.

Kevin is an utterly selfish and entitled retard who doesn't give a shit about what others want or like, do you seriously expect him to listen to your pleads?

oosyrag:
I generally don't expect anything actually, but not stating a reasonable opposing opinion that may have been overlooked just because I expect it to not to result anything would be even more idiotic. It doesn't cost anything to have a differing opinion rejected unless the one making it has overinflated sensitive ego. All of our time here is volunteered anyway, and this isn't our house.

kevingranade:
> please move it to No Hope mod instead.

That's what I did? It's in he obsoletion file so mods can stil use it.

Night-Pryanik:
You also removed it from monster group. To make it 
accessible in the mod, it will require re-adding it in mod's monster group, which I'm asking you for. Sure, I could do it myself later, I'm asking for it just because you're working with it now.

data/json/obsoletion/professions.json
```diff
@@ -0,0 +1,49 @@
+]
```
TheShadowFerret, line 1:
```suggestion
[
```

kevingranade:
> 

Aha, sure no problem.

PR #62836, opened 2022-12-23
Update microlabs cbm item groups

Dan1ss1mo:
#### Summary
Bugfixes "Microlabs cbms loot group fix"

#### Purpose of change
Microlabs currently use two cbm groups one of which has been obsoleted. This pr replaced it with another one

#### Describe the solution
Replace bionics_sci with bionics_mil group

Tweaking both groups a bit so they don't overlap as much.

#### Testing
Trust me

#### Additional context
While there is an intention to have more dedicated places for cbms, we won't get those for the upcoming stable version. So, even though temporary fixes like that are not generally appreciated, i think it make sense now as we are very close to next stable release.

### DIFF
--- a/data/json/mapgen/microlab/microlab_special_tiles.json
+++ b/data/json/mapgen/microlab/microlab_special_tiles.json
@@ -313,7 +313,7 @@
       "items": {
         "p": { "item": "bionics", "chance": 20 },
         "T": [
-          { "item": "bionics_sci", "chance": 100, "repeat": [ 1, 2 ] },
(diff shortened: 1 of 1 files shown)


Maleclypse:
The Xedra Evolved mod actually features Xedra using portal hopping to hunt cyborgs to steal their cbms as part of it's storyline, so this would make sense in it.

Dan1ss1mo:
> The Xedra Evolved mod actually features Xedra using portal hopping to hunt cyborgs to steal their cbms as part of it's storyline, so this would make sense in it.

It certainly is, but until we get more ways to get cbms in vanilla, i feel like it should stay there as well

PR #62843, opened 2022-12-23
Wax changes

Karol1223:
#### Summary
Balance "Plenty of changes regarding wax and the addition of paraffin wax"

#### Purpose of change
I was incredibly annoyed that I could eat the wax rings around toilets.

#### Describe the solution
**Beeswax changes**
- the current wax chunks were renamed to chunks of beeswax
- beeswax is no longer edible, as it is entirely indigestible for humans - I did consider having it be consumable but give no calories but... why?
- beeswax density got slightly lowered to be more realistic
- beeswax no longer conducts electricity
- the weight of beeswax chunks was adjusted accordingly to the accurate density
- beeswax was made unrecoverable, meaning you should no longer be able to scrape it off armor you make
- should now appear under 'spare parts' in the inventory menu

**New material: Paraffin Wax**
- a slightly less-dense alternate-reality beeswax
- functionally pretty much identical at the moment, I am not experienced with working with any wax - any recipe that can use beeswax can use paraffin wax as well, I did this as with the changes I'll describe later beeswax should be much harder to obtain now and I wasn't aware which recipes could only use beeswax, which could only use paraffin and which could use either **DISCLAIMER: if you are actually experienced with this and can tell me if any of these recipes should not take one of the waxes for one reason or another, please let me know. I am open to editing them, I am simply not aware of whether I have made mistakes or not.**
- replaces beeswax as the item dropped from bashing toilets - wax rings around toilet base are paraffin in the real world
- candles now have paraffin wax as their material

**Candle changes**
- as mentioned above the material is now paraffin wax
- had the time it takes for one charge to deplete slightly reduced - now the candle burns exactly 35 hours
- had its weight and volume increased and was given a longest side - the numbers are based off 35 hours paraffin wax candles I found for sale on Amazon

I really really wanted to have candles be cuttable the same way plastic or cotton is so that you could salvage paraffin wax chunks from them but the code stumped me. I am almost certain I made the wax materials able to be cut yet during my testing I could not for the love of me get candles to be cuttable. I am not sure what I did wrong and I'd appreciate feedback on this.

#### Testing
The recipes and items appear to work just about fine. The only issue I've come across was when I tried making candles cuttable - still can't get that to work.

#### Additional context
Requires #63931 to be fixed before being able to get green test results


bombasticSlacks:
Fixed issues and am running this for merging rn, 

However in the future it might be worth it to move the waxes into a crafting group so they don't need to be duplicated everywhere they are used interchangably.

Karol1223:
> Fixed issues and am running this for merging rn,
> 
> However in the future it might be worth it to move the waxes into a crafting group so they don't need to be duplicated everywhere they are used interchangably.

That is a very fair point. I did this manually because, like mentioned in the description, I have no idea how often these waxes can be used interchangeably. I know in cooking they wouldn't be but we don't use wax anywhere in cooking yet.

bombasticSlacks:
>That is a very fair point. I did this manually because, like mentioned in the description, I have no idea how often these waxes can be used interchangeably. I know in cooking they wouldn't be but we don't use wax anywhere in cooking yet.

Yeah, I don't think they exclusively should be grouped, but having it as an option in the future might be nice. If you want to do it for this PR that's fine too but no pressure it was a future suggestion :)

data/json/recipes/ammo/40x46mm.json
```diff
@@ -20,7 +20,7 @@
     "qualities": [ { "id": "CUT", "level": 2 } ],
     "components": [
       [ [ "sheet_metal_small", 1 ] ],
-      [ [ "paper", 1 ], [ "wax", 1 ] ],
+      [ [ "paper", 1 ], [ wax_any", 1, "LIST" ],
```
harakka, line 23: Missing " before wax_any

data/json/recipes/tools/lights.json
```diff
@@ -11,7 +11,10 @@
     "qualities": [ { "id": "CONTAIN", "level": 1 } ],
     "flags": [ "FULL_MAGAZINE" ],
     "tools": [ [ [ "lighter", 5 ], [ "ref_lighter", 5 ], [ "surface_heat", 2, "LIST" ] ] ],
-    "components": [ [ [ "wax", 1 ], [ "any_tallow", 4, "LIST" ], [ "edible_lard", 1, "LIST" ] ], [ [ "cordage_short", 1, "LIST" ] ] ]
+    "components": [
+      [ [ "wax_any", 1, "LIST", [ "any_tallow", 4, "LIST" ], [ "edible_lard", 1, "LIST" ] ],
```
harakka, line 15: Closing square bracket missing?

bombasticSlacks:
I think I've fixed it for you :)

Fris0uman:
Damn I messed up the conflict solve and accidentally removed the turn per charge line U_U

Karol1223:
Closing this until the test failure is resolved, since it's not my fault but a fault in the tests. I am doing this manually because if stalebot does it I won't be able to reopen this myself.

Karol1223:
Tests still borked 😿

Maleclypse:
What test are you talking about being borked?

Karol1223:
> What test are you talking about being borked?

``Basic Build and Test (Clang 10, Ubuntu, Curses)``

```(~[slow] ~[.],starting_items)=> (continued from above) ERROR : src/requirements.cpp:606 [static void requirement_data::check_consistency(const std::vector<std::vector<T> > &, const std::string &) [T = item_comp]] Finalization failed to inline wax_any in inline_recipe_leather_armor_cow_from_scratch
(~[slow] ~[.],starting_items)=> 17:22:02.150 ERROR : src/requirements.cpp:595 [void component::check_consistency(const std::string &) const] wax_any in inline_recipe_leather_armor_cow_from_scratch is not a valid item template
(~[slow] ~[.],starting_items)=> 17:22:02.150 ERROR : src/requirements.cpp:606 [static void requirement_data::check_consistency(const std::vector<std::vector<T> > &, const std::string &) [T = item_comp]] Finalization failed to inline wax_any in inline_recipe_leather_armor_horse_from_scratch
(~[slow] ~[.],starting_items)=> 17:22:02.150 ERROR : src/requirements.cpp:595 [void component::check_consistency(const std::string &) const] wax_any in inline_recipe_leather_armor_horse_from_scratch is not a valid item template```

When I asked a few months back on discord, I was told it's because there's a weird interaction with copy-from, not my fault. If it is my fault then uh... well I'd like to know what I broke.

**Comment by Vzor-** [comment]
This commit history is fantastic. Keep at it. I added the wax to toilets and it always bothered be that you could eat it lol.

File: data/json/recipes/armor/pets_cow.json [hunk 3]
```diff
@@ -109,7 +109,7 @@
     "components": [
       [ [ "water", 36 ], [ "water_clean", 36 ] ],
       [
-        [ "wax", 7 ],
+        [ "wax_any", 7, "LIST" ],
```
Maleclypse, line 112: So if I'm understanding you correctly, you need to get rid of the copy-from in this and two other recipes and explicitly define everything and that will make this PR green?

Karol1223: I think the copy-from is the problem caused, yeah. But also I have no idea how copy-from works with recipes, so I could not tell you what exactly would need to be done. I technically could probably do that but I've opted not to in hope that the underlying issue would be fixed... But it's been almost a year so maybe it's time to admit defeat...

GuardianDll:
I think you also need to copy books where this recipes can be found?

Karol1223:
> I think you also need to copy books where this recipes can be found?

You're right. I hate ``copy-from``

PR #62857, opened 2022-12-24
Readjusts gun and ammo spawns in hunting stores.

DoctorBoomstick:
#### Summary
None
#### Purpose of change
Over my last few games, during which I routinely visited hunting supply stores, I felt that the included spawns of weapons and ammo had some room for improvement. My personal issues with the present spawns, as they stand, include:
1.	The incorporated loot groups that the location pulls from when it comes to rifles, namely the common rifles item group, led to weapons spawning that weren’t hunting related guns. This led to M1 Garands rubbing shoulders with Winchester M70 lever-guns, 9mm pistol caliber carbines spawning alongside hunting rifles, and SKS, VZ 58, and M90 rifles constantly showing up, despite their presence not making all that much sense.
2.	All the guns, be them rifles or shotguns, were spawning either loaded or with partly filled magazines sitting on the same rack.
3.	A complete lack of high-caliber pistols spawning, despite the fact that handgun hunting is an actual real-world sport.
#### Describe the solution
To fix the issues sighted above, I created a number of new item groups dedicated to hunting related guns. These new groups include an item group for hunting rifles and pistols respectively, as well as loot groups for related ammo and magazines. I have swapped the hunting store’s map loot spawns to pull from these new groups, whilst also making some very small tweaks to the building’s layout to facilitate these changes (namely adding a locker for the new magazine spawns). In addition, I have also created a hunting store pallet to streamline the amount of code present.
I thought about extending these changes to other hunting related areas, such as the hunting lodge, though I decided against this as I have never seen the map in question, and as such cannot tell if the same issues with the hunting store apply.
#### Testing
I spawned into a fresh world and checked out a number of hunting stores to verify that the changes work as intended.
#### Additional context
None


Maleclypse (approved):
These look like good changes.  The hunting lodge is basically a privately owned retreat for rich people.   So you're more likely to find things that aren't just for hunting there plus there's at least one variant of it where they hunt people.

Issue #62861, opened 2022-12-24
Cutting cloth to sheets before patches (Makeshift bandage balance)


oosyrag:
### Is your feature request related to a problem? Please describe.

Since the change to make makeshift bandages from cotton sheets/patchwork cotton sheets (very reasonable change), there has occasionally been confusion voiced about how to quickly access materials for a makeshift bandages. The previous way of cutting up clothing for bandages (also reasonable) results in patches, which no longer can be quickly turned into makeshift bandages (the issue). I imagine this concern will be echoed well into the future if it remains as is for stable release.

Currently, the cut action calls for `// Decompose the item into irreducible parts`, with the result being a cotton item is shredded into small cotton patches unsuitable for bandaging wounds (until you sew them up again into a patchwork cotton sheet). This is unintuitive for a player who has the materials and tools to cut up clothing quickly, but is unable to do so to make the desired bandages in an emergency situation. The controls/implementation of the game gets in the way of the desired action. 

Given that it takes less time and effort to cut an item into larger pieces as opposed to small pieces, and that the intent or goal of the player may be to acquire those larger pieces those larger pieces in the first place, it's frustrating that there is no option to do so.

### Solution you would like.

Cut items such as clothing are first cut into (patchwork) cotton sheets, and then further cut into patches. 

Instead of the cut action defaulting to result in the smallest possible part based on the mass/volume/material of the original item, it would result in the largest reasonable piece instead (minimum 2 parts, otherwise resulting in the next smaller unit of material instead).

Optional ideas:
Add choice to recursively cut into irreducible parts (effectively preserving current behavior of the cut up action).
Failure chance on individual pieces, resulting in the smaller item instead.

### Describe alternatives you have considered.

The alternative would be to manually add more deconstruction recipes to result in cotton(and other material) sheets from. I intend to do this for common items such as tshirts and pants as a temporary fix that could be applied for stable, as the above would be a more significant c++ change that may not be suitable to develop at this point during feature freeze.

However, I don't think this is an ideal solution for the following reasons:
1. It's going to be a lot of work adding/changing deconstruction recipes to the multitudes of clothing items in game, prone to mistakes, discrepancies, and simply missing items. Could be a maintenance burden.
[...]

PR #62862, opened 2022-12-24
Adjust deconstruction recipes for patchwork cotton sheets

oosyrag:
#### Summary
Balance "Add additional sources of patchwork cotton sheets from deconstruction"

#### Purpose of change

Temporary fix for bandage accessibility from clothing. See #62861

#### Describe the solution

Change some cloth deconstruction recipes to result in sheet_cotton_patchwork instead of cotton_patchwork.

Edit: Added new deconstruction recipes by skimming through professions list, adding recipes for shirts and pants that people might start with that are made of cotton. I probably missed a lot.

Refer to #62861

#### Testing
None yet

### DIFF
--- a/data/json/recipes/recipe_deconstruction.json
+++ b/data/json/recipes/recipe_deconstruction.json
@@ -6547,6 +6547,16 @@
     "flags": [ "BLIND_HARD" ]
   },
   {
+    "result": "jeans",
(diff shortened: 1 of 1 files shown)


oosyrag:
Rerun tests

PR #62864, opened 2022-12-24
inv_ui: add key for overriding collapsed status

### LINKED ISSUE #62854: keybind for collapse and un-collapse all containers

rigid containers will have items placed in them while looting, meaning that you have to 1 by 1 `U`nload or `<`uncollapse the rigid containers to unload things proper from your inventory after a loot run

### Solution you would like.

keybind for collapse and un-collapse all, same as `> hide contents` and `< show contents` keybinds but for all the containers in an inventory panel your looking at

zone Loot: Unload Everything

### Additional context

image

andrei8l:
#### Summary
None

#### Purpose of change
Add an action to inventory UIs for overriding the collapsed status of all containers
Fixes #62854

#### Describe the solution
Add an action (unbound by default) that cycles collapsed status override between true/false/none. Since we're in string freeze, the override status is only indicated by colored chevrons.

Actually (un)collapsing: I playtested with that and I *hated* it. I don't want my carefully set-up inventory to get scrambled by a stray keypress. People who like floor/pocket pasta won't notice a difference anyway.

Only overriding (or uncollapsing) the visible containers or the top-level ones: seems like a half-way measure that won't satisfy anyone (except hardcore floor pasta enthusiasts).

#### Testing
Only the status of containers is overridden.

[output.webm](https://user-images.githubusercontent.com/68240139/209451563-906c265c-c866-40e1-abb8-f7112976e7e2.webm)

#### Additional context
N/A

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -2020,6 +2020,13 @@
   },
   {
     "type": "keybinding",
+    "id": "SHOW_HIDE_CONTENTS_ALL",
+    "category": "INVENTORY",
+    "name": "Show/hide all contents",
+    "bindings": [ { "input_method": "keyboard_char", "key": "E" } ]
+  },
+  {
+    "type": "keybinding",
     "id": "TOGGLE_SKIP_UNSELECTABLE",
     "category": "INVENTORY",
     "name": "Toggle skipping unselectable items"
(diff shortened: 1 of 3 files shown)


andrei8l:
I'll take another look at this after #63270 gets merged. Maybe we can come up with a default keybinding for this in the meantime

PR #62867, opened 2022-12-25
Turn engines on by default only for non-owned vehicles

### LINKED ISSUE #33955: Set default vehicle state to engine off
# Describe the bug
Vehicles around the Refugee center will sometimes have the engine running on world gen. Previously you could just turn it off (which is still possible) but doing so will potentially anger the free merchants if a window is open/you're seen. 

# Steps To Reproduce
1. Start as Sheltered scenario, any profession
2. Check the vehicles surrounding the building (specifically the buses I've noticed) 
3. Check message log for notification you hear engine running from X direction

# Expected behavior
Vehicle should be off as gasoline/diesel is now a valuable commodity. Running engine would make sense if they're still getting refugees but the center states they're too full to take anyone else. My main reasoning for the change though is I figured it attracts zombies in the area (doubt it's enough to draw wandering spawns though) and I need to protect my poor refugees. 

# Versions and configuration
 - OS: Android 9.0
 - Game Version: Build 9577 (version 0.D-7213-ga0159bf)
 - Graphics version: Tiles
 -  Mods loaded: [DDA, CRIT Exp, Garden Pots, Makeshift Items, Medieval and Historic, More Survival Tools, Mythological Replicas, Dinomod, Modular Turrets, Salvaged Robots, Alt Map Key, EZ Medical, Hydroponics, Beta National Guard, More Locations, Fujis More Buildings, Urban Development, Folding Parts, Vehicle Additions, Tanks and Other, Necromancy]

Night-Pryanik:
#### Summary
None

#### Purpose of change
* Closes #33955.

#### Describe the solution
Added a check for owner when initiating state of the vehicle.

Completely remove code snippet for turning engines on by default.

#### Testing
Visited several refugee centers, checked that engines in their vehicles are turned off.

#### Additional context
None.

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -251,9 +251,9 @@ bool vehicle::remote_controlled( const Character &p ) const
 void vehicle::init_state( map &placed_on, int init_veh_fuel, int init_veh_status,
                           bool may_spawn_locked )
 {
-    // vehicle parts excluding engines are by default turned off
+    // vehicle parts excluding engines in non-owned vehicles are by default turned off
     for( vehicle_part &pt : parts ) {
-        pt.enabled = pt.is_engine();
+        pt.enabled = !has_owner() && pt.is_engine();
     }
 
     bool destroySeats = false;


irwiss:
I think the change should be around line 338 instead (or in addition to), `engine_on` flag is used in some places like sfx, car HVAC and may be some other stuff might be confused even if you disable the engine part

irwiss:
on second thought it looks like has_engine_type_not should check enabled flag, scratch that

PR #62877, opened 2022-12-26
fix: drops of profession flavored zombies

### LINKED ISSUE #62670: "Profession" zombies generally spawn nearly completely nude
### Describe the bug

The `professional_zombie_gear` item group, while intended to let zombies spawn with clothing sets specific to certain professions (i.e. a full set of mailman's gear, or a set of work clothes) cites item groups that are distributions, rather than collections, and are specifically written in order to function as such. This means that if the game rolls that group for a zombie, they'll frequently spawn with a single piece of clothing or no clothing at all.

This seems to have been around since #58651.

### Steps to reproduce

1. Skew up the drop rate of `professional_zombie_gear` for reproducing purposes (`zombie_default.json`, line 276).
2. Kill a buncha zombies.
3. See that items spawn nude or close to nude, which is rather silly.

### Expected behavior

"Profession" zombies should spawn with full sets of appropriate clothes and items, instead of just a single piece of relevant clothing.

### Screenshots

cataclysm-tiles_giemfUEXUz
cataclysm-tiles_yuvFJukKoU
cataclysm-tiles_rmAoqxlGpQ

- OS: Windows
- Game Version: 0.F-12755-gf9cc1b1a43-dirty [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Slots [cbm_slots]
]

### Additional context

This could likely be fixed by adding purpose-made collection item groups for the relevant clothes and items, instead of using existing item groups that aren't meant to be used in that way.

casswedson:
#### Summary
None

#### Purpose of change
every once in a while you'd find a zombie with 'postman' and 'plumber' stuff on them, tools, some hardware and mail carrier items on 'em this fixes them so they spawn correctly and stand out more, you'll find zeds with full mail carrier equipment, letters, envelops in their packs, etc

the 'plumbers' also get different outfits than other zeds, less clothing in general, a simpler look, stuff with work in the name, some protective equipment a couple tools; mostly what was defined as their spawns before

closes #62670
acktually probably (maybe (maybe) not quire sure) no, it should fix it as 'plumbers' get 'large' backpacks for their tools

#### Testing
some samples
Screenshot from 2022-12-25 18-38-23
Screenshot from 2022-12-25 18-46-22
Screenshot from 2022-12-25 18-47-29

#### Additional context
writing this I remembered I wanted to also make the mail peps carry packages, oh well, for later I'm sure

### DIFF
--- a/data/json/monsterdrops/zombie_default.json
+++ b/data/json/monsterdrops/zombie_default.json
@@ -266,25 +266,36 @@
   {
     "type": "item_group",
     "subtype": "collection",
-    "//": "contains cloth (always), a wallet (30% chance), and items in pockets or a bag (50% chance)",
+    "//": "contains clothing (always), a wallet (30% chance), and items in pockets or a bag (80~% chance)",
     "id": "default_zombie_death_drops",
     "entries": [
+      { "group": "wallets", "damage": [ 1, 4 ], "prob": 30 },
       {
         "distribution": [
-          { "group": "default_zombie_clothes", "prob": 90 },
-          { "group": "default_zombie_clothes_christmas", "event": "christmas", "prob": 10 },
-          { "group": "professional_zombie_gear", "prob": 1 }
+          {
+            "collection": [
+              {
+                "distribution": [
+                  { "group": "default_zombie_clothes", "prob": 90 },
+                  { "group": "default_zombie_clothes_christmas", "event": "christmas", "prob": 10 }
+                ]
+              },
+              {
+                "distribution": [ { "group": "default_zombie_items", "prob": 80 }, { "group": "default_zombie_items_pockets", "prob": 20 } ]
+              }
+            ],
+            "prob": 96
+          },
+          {
+            "collection": [ { "group": "postman_gear", "damage": [ 1, 4 ] }, { "group": "default_zombie_items_pockets", "prob": 10 } ],
+            "prob": 2
+          },
+          {
+            "collection": [ { "group": "prof-plumbing_gear", "damage": [ 1, 4 ] }, { "group": "default_zombie_items_pockets", "prob": 10 } ],
+            "prob": 2
+          }
         ]
-      },
-      { "group": "default_zombie_items", "prob": 50 },
-      { "group": "wallets", "damage": [ 1, 4 ], "prob": 30 }
+      }
     ]
-  },
-  {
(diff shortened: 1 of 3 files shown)


data/json/monsterdrops/zombie_default.json
```diff
@@ -266,25 +266,30 @@
...
-          { "group": "default_zombie_clothes_christmas", "event": "christmas", "prob": 10 },
-          { "group": "professional_zombie_gear", "prob": 1 }
+          {
+            "collection": [
+              { "group": "default_zombie_clothes", "prob": 90 },
+              { "group": "default_zombie_clothes_christmas", "event": "christmas", "prob": 10 },
+              { "group": "default_zombie_items", "prob": 80 }
+            ],
+            "prob": 96
+          },
```
mqrause, line 282:
```suggestion
          {
            "collection": [
              {
                "distribution": [
                  { "group": "default_zombie_clothes", "prob": 90 },
                  { "group": "default_zombie_clothes_christmas", "event": "christmas", "prob": 10 }
                ]
              },
              { "group": "default_zombie_items", "prob": 80 }
            ],
            "prob": 96
          },
```

I didn't test this beyond linting, but I think it should look something like this for them to definitely spawn with one set of clothing.

casswedson: it does look like there's a bug in the logic of `event` I am getting some sparsely clothed zombies applying this patch
Screenshot from 2022-12-25 22-57-53
Screenshot from 2022-12-25 22-58-42

Screenshot from 2022-12-25 22-57-40
by the way I did look at the christmas itemgroups, just didn't look hard enough I guess, I was like they're just hats

mqrause: > it does look like there's a bug in the logic of `event` I am getting some sparsely clothed zombies applying this patch 

I'm not seeing anything obvious, except christmas goes from the 14th to the 25th in the code, so depending on your timezone it might be over already.

Not sure what the issue with my suggestion really is. But I think your version has a 10% chance to roll naked zombies, since `default_zombie_clothing` has a 90% chance in a collection. I was trying to amend that by wrapping the normal and christmas clothing in a distribution so you always get one of them.

Maleclypse (approved):

PR #62883, opened 2022-12-26
Hwp stock

Dan1ss1mo:
#### Summary
Balance "modular folding stock for HWP"

#### Purpose of change
Swap hwp integrated adjustable stock for modular folding stock

Not do it if there is some specific reason for hub to favour adjustable stock. May also leave name as it is for now to dodge string freeze. (realized it wouldn't work as i have to also add folded stock variant)

#### Testing
WIP

#### Additional context
I feel like hub would prefer to integrate best possible attachment into they're signature weapon

### DIFF
--- a/data/json/items/gun/robofac_gun.json
+++ b/data/json/items/gun/robofac_gun.json
@@ -293,7 +293,7 @@
     "id": "robofac_stock_folded",
     "type": "GUNMOD",
     "copy-from": "high_end_folding_stock_folded",
-    "name": { "str": "bullpup modular folding stock (folded)" },
+    "name": { "str": "bullpup modular folding stock (folded)", "str_pl": "bullpup modular folding stocks (folded)" },
     "use_action": {
       "type": "transform",
       "msg": "You unfold the stock for combat.",


Maleclypse:
Hub01 specifically is stated to suffer from "not designed here" syndrome.  So this particular PR has been mentioned as a no go previously iirc.

Dan1ss1mo:
> Hub01 specifically is stated to suffer from "not designed here" syndrome. So this particular PR has been mentioned as a no go previously iirc.

But why can't they design they're own folding stock instead of adjustable one? As you can see it's still treated as unique gun mod. It's just copy base stats from different mod now

Maleclypse:
> > Hub01 specifically is stated to suffer from "not designed here" syndrome. So this particular PR has been mentioned as a no go previously iirc.
> 
> But why can't they design they're own folding stock instead of adjustable one? As you can see it's still treated as unique gun mod. It's just copy base stats from different mod now

Because of actual game balance. You've already said this is a better stock than what it currently has.  It's supposed to be a gun with a whole lot of tradeoffs and not the best in anything.

Dan1ss1mo:
> > > Hub01 specifically is stated to suffer from "not designed here" syndrome. So this particular PR has been mentioned as a no go previously iirc.
> > 
> > 
> > But why can't they design they're own folding stock instead of adjustable one? As you can see it's still treated as unique gun mod. It's just copy base stats from different mod now
> 
> Because of actual game balance. You've already said this is a better stock than what it currently has. It's supposed to be a gun with a whole lot of tradeoffs and not the best in anything.

I see, make sense. Guess i miss understood the intended balance for this gun

PR #62895, opened 2022-12-27
Added message when pouring a liquid into container when parent is full.

### LINKED ISSUE #62872: Non-rigid containers in inventory don't fill up with liquids when asked to
### Describe the bug

Trying to fill up a non-rigid container in inventory transfers no liquids.

Workaround: Throw the container on the ground, fill it there, and pick it up when filled.

### Steps to reproduce

Have a huge container (`hiking backpack`) with a `gallon-sized zipper bag`
<details><summary>Get to a toilet, examine and decide to pour liquid into a container</summary>

image</details>

<details><summary>Choose yours `gallon-sized zipper bag`</summary>

image</details>

<details><summary>This message appears in the log, but no liquid is transferred!</summary>

image</details>

### Expected behavior

Firstly, the solution:
1. Calculate how much can a container be filled to still fit into ANY non-blacklisted (etc.) pocket. (If only the same pocket were checked, a `gallon-sized zipper bag` in my pans would decide to contain only 0.25 L instead of throwing it into my `hiking backpack` for full 4.5 L).
2. Fill it up.
3. Throw into a suitable pocket.

Secondly, be truthful:
If I don't fill it up, then don't tell me I did!
E. g. check for volume transferred > 0 L
And a debug warning for when this happens.

### Screenshots

_No response_

- OS: Windows
- Game Version: 99451f2 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

oosyrag:
#### Summary
None

#### Purpose of change
Resolves #62872

#### Describe the solution

Added a message "There is not enough room to add any more (liquid)." when amount to be transferred == 0. This will display upon attempting to fill the container when the parent container cannot contain any more.

I'm not familiar with C++ at all, please let me know if I did something wrong.

Do not have the not full, non-rigid container show up on the list of valid containers to pour into if the parent pocket of the container cannot fit any more of the liquid. Seems convoluted compared to the simple fix of just letting the player know what's going on. This alternative could also possibly cause confusion along the lines of "This container has space, why can't I choose it?"

#### Testing

Compiled and tested, looks ok. Although I'm not sure why the condom doesn't get "filled to the brim", while the gallon sized-zipper bag does.
image (1)

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -5612,7 +5612,7 @@ bool Character::pour_into( item_location &container, item &liquid, bool ignore_s
     }
 
     if( amount == 0) {
-        add_msg_if_player( _( "There is not enough room to add any %1$s." ), liquid.tname() );
+        add_msg_if_player( _( "There is not enough room to add any more %1$s." ), liquid.tname() );
         return false;
     }
 
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -501,8 +501,8 @@ std::string enum_to_string<blood_type>( blood_type data )
         case blood_type::blood_B: return "B";
         case blood_type::blood_AB: return "AB";
             // *INDENT-ON*
-    case blood_type::num_bt:
-        break;
+        case blood_type::num_bt:
+            break;
     }
     cata_fatal( "Invalid blood_type" );
 }
@@ -920,7 +920,7 @@ double Character::fastest_aiming_method_speed( const item &gun, double recoil,
         int effective_iron_sight_dispersion = effective_dispersion( gun.type->gun->sight_dispersion );
         double iron_sight_speed = modified_sight_speed( 0, effective_iron_sight_dispersion, recoil );
         if( effective_iron_sight_dispersion < recoil && iron_sight_speed > aim_speed_modifier &&
-                recoil <= iron_sight_FOV ) {
+            recoil <= iron_sight_FOV ) {
             aim_speed_modifier = iron_sight_speed;
         }
     }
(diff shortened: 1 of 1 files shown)


Brambor:
Let's discuss! When condom is in small pocket, could there be msg like:
`There is not enough room to add any more %1$s. Pocket for condom is too small.`
I believe it would solve OP's concern
> This alternative could also possibly cause confusion along the lines of "This container has space, why can't I choose it?"

oosyrag:
That's reasonable, but the current implementation doesn't actually confirm that it's because the parent pocket has no space, it just displays when nothing gets moved. I'm unfortunately a complete c++ novice so that's all I've got for now. I'll look into it though.

There's also the desire to keep the message log short and clean without too much verbosity. I'm thinking something along the lines of "(container) can't expand to add any more (liquid)" would be more precise, but sometimes a more generic statement can cover more cases. Again, I'll have to check to see what other situations might benefit from the message or if the parent container problem is the only one.

Brambor:
Well, every container has a pocket, so if an item is in inventory it is in a pocket.
I don't know if there are cases when you cannot fill a container up because its volume is capped by **something**. And that case the **something** not being in a pocket.

It would be great to have a correct implementation with all edge cases, but if something slips, it is still better than before ;)

Brambor:
Note the astyle error, maybe you are using different astyler?

Brambor:
And, after astyle, you will want to squash all your commits into one. Hit me on Discord for help with that.
Or you can follow https://stackoverflow.com/questions/25356810/git-how-to-squash-all-commits-on-branch

Brambor:
@oosyrag Do you plan on finishing this up? Otherwise, I will try to pretty much copy this and make a new PR.

oosyrag:
Oh yeah sure, I'll look into astyling properly. This was on the back burner because of string freeze right now, it won't get merged until G stable releases anyway.

oosyrag:
Rerun tests

PR #62899, opened 2022-12-27
Reasons for slow or fast crafting are now described in details

### LINKED ISSUE #62868: [User Experience] The crafting UI should show if hands are encumbered

I have just found out that crafing time depends on hand encumbrance. It happened like this:
1. I knew how long crafting a throwing stick is supposed to take
2. I noticed a substantially longer crafting time
3. I sat there puzzled for several minutes trying to figure out whether this is a bug and what is causing this

image
0 hand encumbrance

image
20 hand encumbrance

### Solution you would like.

We should display a warning in the crafing UI under the "Proficiencies used" line
image

_No response_

### Additional context

_No response_

Night-Pryanik:
#### Summary
Interface "Reasons for slow or fast crafting is now described in details"

#### Purpose of change
* Closes #62868.

#### Describe the solution
- Display reasons for slow crafting: morale, lighting, and manipulator score modifiers. Modifiers are displayed only if they actually affect crafting.
- Display manipulator score modifier for fast crafting.
- Reworded message for slow/fast crafting. Previously the message was `crafting is slow 22%`. What does this mean? That crafting is slowed by 22% or to 22%? Unclear.

None.

#### Testing
Wore mittens to get hands encumbrance. Went into a dark tile. Opened crafting menu and checked message for slow crafting.

#### Additional context
Message for slow crafting due only to hands encumbrance
изображение

Message for slow crafting due to bad lighting and hands encumbrance
изображение

Can anyone please explain to me why it's cropping the message while there's obviously enough space for the whole message?

### DIFF
--- a/src/crafting_gui.cpp
+++ b/src/crafting_gui.cpp
@@ -47,9 +47,10 @@
 #include "ui_manager.h"
 #include "uistate.h"
 
+static const limb_score_id limb_score_manip( "manip" );
+
 static const std::string flag_BLIND_EASY( "BLIND_EASY" );
 static const std::string flag_BLIND_HARD( "BLIND_HARD" );
-
 
 class npc;
 
@@ -2050,7 +2051,7 @@ static void draw_can_craft_indicator( const catacurses::window &w, const recipe 
     } else if( player_character.crafting_speed_multiplier( rec ) < 1.0f ) {
         int morale_modifier = get_player_character().morale_crafting_speed_multiplier( rec ) * 100;
         int lighting_modifier = get_player_character().lighting_craft_speed_multiplier( rec ) * 100;
-        int limb_modifier = get_player_character().get_limb_score( limb_score_id( "manip" ) ) * 100;
+        int limb_modifier = get_player_character().get_limb_score( limb_score_manip ) * 100;
 
         std::stringstream modifiers_list;
         if( morale_modifier < 100 ) {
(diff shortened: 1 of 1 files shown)


Qrox:
> Can anyone please explain to me why it's cropping the message while there's obviously enough space for the whole message?

The window used for the message (`w_head_info`) only covers part of the screen so `right_print` trims the message to fit it into the window.

https://github.com/CleverRaven/Cataclysm-DDA/blob/e171fb3e3a5c819ef90c6c0c0b18dce06e71cbca/src/crafting_gui.cpp#L1087-L1088

Night-Pryanik:
The thing is, if I made it print the _fixed_ long string like `crafting is slowed to 45%: lighting 23%, hands encumbrance/wounds 56%`, the whole message happily fits. See this pic for an example.
изображение

mqrause:
https://github.com/CleverRaven/Cataclysm-DDA/blob/e171fb3e3a5c819ef90c6c0c0b18dce06e71cbca/src/crafting_gui.cpp#L2011-L2021

The width is also capped to a quarter of the screen or the length of the single longest reason. Multiple reasons are simply not accounted for.

Night-Pryanik:
Space for crafting speed modifier string is now no less than `Recipe Result` window width.

When there's enough space
изображение

When there's not enough space
изображение

Raikiri:
I'm really struggling without this, especially when I unequip everything from my hands and it still says that I'm 55% slow.

PR #62913, opened 2022-12-28
drop leg pouches fix

Dan1ss1mo:
#### Summary
Bugfixes "fix max weight for drop leg pouches"

#### Purpose of change
pair of drop leg pouches seems to have extremly low max weight relative to volume compared to othetr similar items.

#### Describe the solution
Changing stats to be compared to fanny pack, as it is the closest storage item in terms of properties

#### Testing
Spawned pair of drop leg pouches and checked description

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -251,9 +251,9 @@ bool vehicle::remote_controlled( const Character &p ) const
 void vehicle::init_state( map &placed_on, int init_veh_fuel, int init_veh_status,
                           bool may_spawn_locked )
 {
-    // vehicle parts excluding engines are by default turned off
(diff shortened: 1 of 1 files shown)


data/json/items/armor/storage.json
```diff
@@ -1124,18 +1124,18 @@
...
         "max_item_length": "30 cm",
         "magazine_well": "500 ml",
-        "moves": 200
+        "moves": 80
       },
       {
         "pocket_type": "CONTAINER",
         "max_contains_volume": "2250 ml",
-        "max_contains_weight": "2 kg",
+        "max_contains_weight": "4500 kg",
```
Wishbringer, line 1135:
```suggestion
        "max_contains_weight": "4500 g",
```

PR #62922, opened 2022-12-28
Birb nests and egg foraging tweaks

NetSysFire:
#### Summary
Content "Bird nests can be found on roofs"

#### Purpose of change

I am surprised we did not have this already.

#### Describe the solution

Adds a 1x1 nested mapgen, always placing 5-10 twigs, some feathers with 70% chance and 1-3 eggs with 30% chance. Also replaces the hardcoded generic bird egg in the foraging table with a random bird egg.
I derived the probabilities from http://www.bio.brandeis.edu/fieldbio/Birds_Kamm_Kuss/Pages/PAGE_SEARCH_HABITAT.html

<s>Still work in progress because I need help resolving an issue, see testing below.</s>I plan to add that to basically all roofs where it would make sense. Also structures like radio towers.

Also removes NOITEM from gutters. Seeing how much clutter you can have them irl, it feels fine to remove it.

The changes in eggs is actually beneficial for the player as duck and goose eggs are more nutritious than others, since currently even tiny hummingbird eggs have the same nutrients as a generic bird egg.

1. Adding a bird nest item or furniture. It would be tedious for the player to have to disassemble them every time, but on the other hand avian mutants could use them as a sleeping aid or something. Requires C++ tweaks though.
2. Also spawning a birb. Due to technical limitations the likelyhood of the egg and bird being mismatched (e.g crow -> sparrow egg) is high, though.
3. Additionally spawning bird litter around the nest. I doubt birds defecate in or near their nests though, also this would add yet another instance of items floating in midair.

#### Testing

image

### DIFF
--- a/data/json/mapgen/radio_tower.json
+++ b/data/json/mapgen/radio_tower.json
@@ -118,12 +118,12 @@
         "          Raaa,R        ",
         "                        ",
         "                        ",
-        "         ...........3   ",
-        "         ...........5   ",
-        "         ...........3   ",
-        "         .......&...3   ",
-        "          R   ......3   ",
-        "              ......3   ",
+        "         222222222223   ",
+        "         |..........5   ",
+        "         |..........3   ",
+        "         -----|.&...3   ",
+        "          R   |.....3   ",
+        "              ------3   ",
         "                        ",
         "                        ",
         "                        ",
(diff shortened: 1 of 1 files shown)


LovamKicsiGazsii:
> Uhh. Well. I am not finished yet and would like some help on the spawning of the nested chunk. It appears to always spawn one tile _besides_ the intended location. 

That's because items cannot be added to rain gutters (because of the `NOITEM` flag), so it overflows to a neighboring tile.

NetSysFire:
>so it overflows to a neighboring tile.

Dammit. I thought I could avoid having to manually put in coordinates. Thanks for your help.

LovamKicsiGazsii:
> Dammit. I thought I could avoid having to manually put in coordinates. Thanks for your help.

No problem. I feel like the rain gutter can do without the `NOITEM` flag. That way you'd only have to add nests to the roof palette and avoid having to define them individually.

NetSysFire:
@Night-Pryanik I did not typo "birb". But yeah you are probably right to keep the title a bit more serious.

NetSysFire:
I figured out that I can use palettes for this. It is amazing and saved me so much work. Only thing left to do is to check all mapgen things using the roof palette to ensure there are no improper spawns.

PR #62923, opened 2022-12-29
Aftershock: Alien scanner thingy

John-Candlebury:
#### Summary
Mods "Aftershock: Add a motion scanner"

#### Purpose of change
Can a setting really claim to be alien inspired and not have this thing?

Also theres plenty of pretty scary aftershock monsters inside ruins so this thing seems like a sensible thing to carry if you are attempting to be stealthy.

#### Describe the solution
This item actually has dual functions, and how it works depends on whether you are holding it.

While on and in your inventory, it'll scan a medium sized area around you every 10 seconds and make noise of varying intensity depending on the amount of monsters nearby.  If you also wield it, the noise will be accompanied by  a brief glimpse of where the monsters are located, each indicated as a red `?`

In the future I also want to make furniture pieces that count as motion sources. So this isnt super reliable.

#### Testing

Spawned and used the tracker.


GuardianDll:
What exactly `MOTION_VISION_RANGE` do? Can you add this info into magic.md? Thank you

PR #62934, opened 2022-12-30
Why no PALS comment on the MDS

Maleclypse:
#### Summary
None

#### Purpose of change
Putting discord information in game so it doesn't need to be repeated often.

#### Describe the solution
Adds comment from HWP author to HWP json.

None

#### Testing
Automated testing for comment addition.


Dan1ss1mo:
shouldn't it be added to modular defence instead though?

PR #62944, opened 2022-12-31
Tweak/add fishing tool craft/uncraft recipes

cake-pie:
#### Summary
Content "Added / adjusted craft/uncraft recipes for fishing tools"

---

#### Purpose of change

Adjust recipe for `hoboreel`, because requiring 2 × `2x4` as components is excessive

#### Describe the solution

Since the item isn't much more than a 6" handle, replace the 2 × `2x4` requirement with just one `wood_block`, which appears to be the best fit for the overall size. Crafting would involve whittling down the "2x2x5 inches" block. It's slightly shorter than the defined `"longest_side": "150 mm"` of the hobo reel, but the difference is small enough to be within acceptable variance.

Also give the activity `LIGHT_EXERCISE` as it involves either whittling the `wood_block` or cutting down a `stick`

`wooden_post_short` was also considered, but it's too big. `splinter` would be a good size but is too damaged.

I also considered adjusting the existing `stick` requirement (`"longest_side": "130 cm"`) which is like 8 times as long as actually needed, but there is no good "short stick" item that could be used so I just left it as-is.

---

#### Purpose of change

Add uncraft for `fishing_rod_professional`.

The player used to be able to make this item, but that crafting recipe was outright removed in e57b89354ad1a783ff0881965fb00321fe273f89 in relation to plastic crafting changes (vacuum forming) presumably because it was thought to be beyond any reasonable capability. Still, even if we don't want this to be craftable, there was no replacement uncraft recipe, making it no longer possible to salvage resources from the item.

#### Describe the solution

Uncraft yield determined with reference to crafting recipe that previously existed: 3 ×`plastic_chunk`, 1 × `wire`, 1 × `scrap`

For `weights` requirement replace with 200 × `bismuth` (checking git blame takes us to 78b11923444975260435ab4849504cefa3a8e101 -- there are 5 possible alternative inputs, but when dismantling an item that was map-spawned rather than player-crafted it would use the first in the list by default.)

Add string (amount per `fishing_rod_basic` recipe) and fishing hook, as those were missing from recipe due to an obvious oversight.

Instead of producing string, have actual nylon fishing line which would be stronger and have other potential uses. That'll involve a new item, adding it to applicable item groups and material requirements, etc etc which is a bit beyond the scope of what I hoped to achieve with this PR. Also, best deferred due to the 0.G string freeze. Could file an issue to be picked up again later on, if there is demand for this.

---

#### Purpose of change

Add uncraft for `plastichoboreel` -- why not, since I'm doing all this related stuff anyway.

#### Describe the solution

Like the `hoboreel` crafting recipe, but yields plastic instead of wood.

---

#### Purpose of change

Move recipe for `fishing_rod_basic` from `data/json/recipes/recipe_others.json` to `data/json/recipes/tools/tool.json`. This is purely for json organization reasons.

---

#### Testing

Load up in game and try the changed recipes.

#### Additional context
-nil-

### DIFF
--- a/data/json/recipes/tools/tool.json
+++ b/data/json/recipes/tools/tool.json
@@ -464,7 +464,7 @@
   },
   {
     "type": "recipe",
-    "activity_level": "fake",
+    "activity_level": "NO_EXERCISE",
     "result": "fishing_rod_basic",
     "category": "CC_OTHER",
     "subcategory": "CSC_OTHER_TOOLS",


RenechCDDA:
`fake` should not be used as an activity level (defaults to moderate), if you are adjusting the recipe this is an excellent time to resolve that.

cake-pie:
> `fake` should not be used as an activity level (defaults to moderate), if you are adjusting the recipe this is an excellent time to resolve that.

Replaced with `NO_EXERCISE` as I see very little physical exertion involved here, nothing much beyond winding string around a stick.

PR #62945, opened 2022-12-31
Consolidate fishing implements logic to use only tool quality, eliminate redundant item flags

cake-pie:
#### Summary
Infrastructure "Consolidate fishing implements logic to use only tool quality"

#### Purpose of change

a)
Fishing rods/reels have both `FISHING` tool quality as well as a flag, either `FISH_POOR` or `FISH_GOOD`:
https://github.com/CleverRaven/Cataclysm-DDA/blob/0eea824edfc3d54ee006544b40f47c38a488da45/data/json/items/tool/fishing.json#L100-L101

In the source code, some places checked the tool quality while other places used the flags. It was all rather redundant and inconsistent.

b)
`FISHING` tool quality applies only to fishing rods and are not related to fish traps, which are a different type of fishing implement.

c)
The `use_action` of `FISH_ROD` in each fishing rod's item definition is redundant with having the fishing rod tool quality

#### Describe the solution

a)
Replace all source code that uses flags with corresponding code that checks the tool quality.
Eliminate the flags.

b)
Rename `FISHING` tool quality to `FISHING_ROD` to improve distinction vs `FISH_TRAP`.

c)
Declare `FISH_ROD` as a usage in the `FISHING_ROD` tool quality and remove it from `use_action` of individual items.

Switch everything to use flags only and eliminate the tool quality. That seems a poorer choice -- if one wanted to expand the possibilities with one more tier that will require another flag e.g. `FISH_EXCELLENT`. With the use of qualities it's just a number and will be much simpler to update to code to handle any future expansion.

#### Testing

Compile, run, set up a new character as fisher profession in wilderness desert island location, go fishing.

#### Additional context

nope

### DIFF
--- a/data/json/items/tool/fishing.json
+++ b/data/json/items/tool/fishing.json
@@ -77,7 +77,6 @@
     "material": [ "wood" ],
     "symbol": "/",
     "color": "brown",
-    "use_action": [ "FISH_ROD" ],
     "qualities": [ [ "FISHING_ROD", 1 ] ],
     "flags": [ "SHEATH_SPEAR" ]
   },
@@ -96,7 +95,6 @@
     "material": [ "plastic" ],
     "symbol": "/",
     "color": "brown",
-    "use_action": [ "FISH_ROD" ],
     "qualities": [ [ "FISHING_ROD", 2 ] ],
     "flags": [ "SHEATH_SPEAR" ]
   },
@@ -117,7 +115,6 @@
     "material": [ "plastic" ],
     "symbol": "/",
     "color": "black",
-    "use_action": [ "FISH_ROD" ],
     "qualities": [ [ "FISHING_ROD", 1 ] ]
   },
   {
@@ -136,7 +133,6 @@
     "material": [ "wood" ],
     "symbol": "/",
     "color": "brown",
-    "use_action": [ "FISH_ROD" ],
     "qualities": [ [ "FISHING_ROD", 1 ] ]
   }
 ]
--- a/data/json/items/tool/fishing.json
+++ b/data/json/items/tool/fishing.json
@@ -78,7 +78,7 @@
     "symbol": "/",
     "color": "brown",
     "use_action": [ "FISH_ROD" ],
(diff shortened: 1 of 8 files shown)


data/json/items/tool/fishing.json
```diff
@@ -118,8 +118,7 @@
     "symbol": "/",
     "color": "black",
     "use_action": [ "FISH_ROD" ],
-    "qualities": [ [ "FISHING", 1 ] ],
-    "flags": [ "FISH_GOOD" ]
+    "qualities": [ [ "FISHING", 1 ] ]
```
Karol1223, line 121: It appears that while FISH_GOOD got translated to FISHING 2 in another tool, here FISH_GOOD would translate to FISHING 1. This looks to be an oversight that took place when this tool was initially implemented if I had to take a guess, but I suspect it's not intended to be like this with the current changes?

cake-pie: The discrepancy between the two here was likely to have been an oversight initially, but keeping this at FISHING 1 in this PR is deliberate. Reasoning as follows:
- The descriptions for both `hoboreel` and `plastichoboreel` indicate that they are "not as effective as a normal fishing pole", so they have to be inferior to the commercial product `fishing_rod_professional`
- The mechanical operation of the wooden hobo reel and its plastic commercial counterpart are so simple that the commercial version shouldn't confer much advantage in terms of how well it will catch fish. It's a spool for the line integrated to a fixed handle, simple as.
- By comparison, with the professional rod what comes to mind is metal or graphite or carbon fiber construction, spinning reel with hand crank; even the way the rod bends is by design.

Karol1223: Ah, that is fair. I'll be perfectly honest, I did not check what item this was - I simply noticed the code inconsistency and figured it's better to point it out than not to. If these are both hoboreels then I do agree that they should probably not differ a whole lot if at all. 

Either way, good job cleaning this all up. Thanks for this!

PR #62947, opened 2022-12-31
Fix 4.5$ price per bullet for 762x39 when other ammo costs <0.5$

borzykin:
#### Summary
Balance Fix 7.62 ammo costing 1.5$ per shot when other ammo types between 0.25$ - 0.6$ per shot

#### Purpose of change
Make 7.62 ammo more accessible from traders like Banker (Jey?). With my changes it costs 0.68$ per shot instead of 1.5$. It is a bit more than average, but this caliber is rare for NA setting so seem to be fair

#### Describe the solution
Drops price for 7.62 shots from >1.5$ per shot to be in line with other caliber prices

n/a

#### Testing
Checked price at isolated_road encounter trader

#### Additional context
n/a

### DIFF
--- a/data/json/items/ammo/762.json
+++ b/data/json/items/ammo/762.json
@@ -30,7 +30,7 @@
     "//": "Entirely provisional.  Could someone design a better wound ballistics system?",
     "description": "7.62x39mm M67 ammunition with a 123gr bullet.  The inadequate terminal ballistics of the M43 round led to the development of the M67 round in Yugoslavia in the 1960s.  It destabilizes much faster than M43 after hitting a target, leading to greater damage.",
     "price": 150,
-    "price_postapoc": 980,
+    "price_postapoc": 98,
     "relative": { "damage": { "damage_type": "bullet", "amount": 4, "armor_penetration": 2 } }
   },
   {


data/json/items/ammo/762.json
```diff
@@ -30,7 +30,7 @@
     "//": "Entirely provisional.  Could someone design a better wound ballistics system?",
     "description": "7.62x39mm M67 ammunition with a 123gr bullet.  The inadequate terminal ballistics of the M43 round led to the development of the M67 round in Yugoslavia in the 1960s.  It destabilizes much faster than M43 after hitting a target, leading to greater damage.",
     "price": 150,
-    "price_postapoc": 6800,
+    "price_postapoc": 980,
```
Maleclypse, line 33:
```suggestion
    "price_postapoc": 98,
```
I don't think any round should cost more than a dollar post apoc.

kevingranade: I'm not sure I agree with that, though for intermediate rounds like this yea probably.

kevingranade:
This leaves a ton of rounds out, lots of 7.62 and 5.56 variants are in this range.

borzykin:
Actually I was mistaken with title. Those rounds cost 4.5$, not 1.5$. So fix is needed indeed:
image

borzykin:
@Maleclypse setting price to 98 makes them cost 0.07$ at trader, which is not matching with other rounds. My guess that initial 6800 supposed to be 680. It bring the cost down from 4.5$ to 0.48$

PR #62949, opened 2023-01-01
[Xedra Evolved] Fix the mod has lack of own mana bar

GuardianDll:
#### Summary
None
#### Purpose of change
Did you know xedra mod has no his own mana widget? everyone just uses the magiclysm one? an, what is fun, no one spot it for *checks the dates* 7 month?
#### Describe the solution
Add our own widget, which currently just copypasted mana widget from magiclysm, with changed color - currently its `c_pink`, but i need to discuss it with @maleclypse to pick the right color (the original idea was orange, but we have no orange color)
#### Testing
It works
#### Additional context
image
image

### DIFF
--- a/data/mods/Xedra_Evolved/ui/sidebar.json
+++ b/data/mods/Xedra_Evolved/ui/sidebar.json
@@ -1,0 +1,184 @@
+[
+  {
+    "//": "Maximum mana points, in light blue color.",
+    "id": "xe_max_mana_num",
+    "type": "widget",
+    "label": "Max Mana",
+    "style": "number",
+    "var": "max_mana",
+    "colors": [ "c_pink" ]
+  },
+  {
+    "//": "Base widget for other mana widgets to copy-from.  Sets var and colors.",
+    "id": "xe_mana_base",
+    "type": "widget",
+    "var": "mana",
+    "colors": [ "c_red", "c_light_red", "c_yellow", "c_pink" ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


Maleclypse:
Pink looks fine to me. Thank you for catching this.

Maleclypse (approved):

GuardianDll:
How do you think, can it pass the string freeze as a bugfix, or it will be merged after the 0.G?

TheShadowFerret:
I can pick just one's widget if I have both loaded, right?

~~It will be Magiclysm's because I like blue more, don't hurt me~~

Maleclypse:
> How do you think, can it pass the string freeze as a bugfix, or it will be merged after the 0.G?

I believe that it’s not a new string because it it is letter for letter the same as Magiclysm’s widget but I wouldn’t be surprised if I misunderstood.

GuardianDll:
> I can pick just one's widget... 

Sure thing, you still can pick magiclysm one
I made them different to not make it like 100% copy (also even if they have same id, somehow they do not overwrite each other, so dumb copypaste make you have two same widgets. So i decide to change the color, to at least make some difference, and for future plans ro separate them)

Maleclypse:
> I can pick just one's widget if I have both loaded instead of having to show both, right?
> 
> ~It will be Magiclysm's because I like blue more, don't hurt me~

I hope so but I’m not super sure how widgets work and I will never forgive you for preferring a different color. :)

PR #62957, opened 2023-01-01
Paint changes

### LINKED ISSUE #61479: Paint should be migrated to a bucket holding liquid paint

Paint is currently self-contained, with no paint can containing it.

### Solution you would like.

Make paint a liquid that must be held in a can.

N/A

### Additional context

_No response_

Karol1223:
#### Summary
Content "Vast edits to existing paints and some new colors"

#### Purpose of change
The paints have come as solids with built-in paint cans since literally forever. But there is no reason for this to be the case. It's an artifact from ages past alongside the eye drops (but I ain't touching those) and so it should probably be fixed. Fixes #61479

#### Describe the solution
1. Adds a new paint material (almost everything is copied from alcohol, it is however denser, non-conductive and inedible)
2. Adds two new paint cans which the paints can spawn in. One is steel (and can be disassembled into scrap) and one is plastic
3. Plenty of changes to the existing paints:
- changed type from AMMO to COMESTIBLE - not sure why they were AMMO in the first place, all liquids are either that or COMESTIBLE so I went with the latter so that it can freeze and have proper dirty prefixes, it is still inedible due to the flag I copied from bleach
- gave them 15 charges 250ml each - this is the maximum that can fit in a gallon paint can with these sizes of charges
- made one charge of paint 325g to go in line with the average paint density I found online
- the description no longer references a pre-built paintcan
- they are now liquid
- all colours copy_from from the red paint - this removes excess code that would otherwise be identical across different colours
4. Added several new colours of paint: orange, pink, cyan, gray and brown

None anymore! Now old paint will migrate into a full to the brim can of paint upon the merging of this PR.

#### Testing
The itemgroups spawn just right. The paint is certainly a liquid now.

#### Additional context
This adds new items and alters existing descriptions - absolutely should be given a String Freeze label


Night-Pryanik:
#61479?

Karol1223:
> #61479?

I was on the verge of checking if someone made an issue about this and it looks like I should have followed through with that. I'm hesitant to say that this PR fixes that 100% as I'm struggling with actually migrating the existing paints (still no idea if there's even support for something like this) but I guess it pretty much does as much as there can be done to solve the issue.

mqrause:
> I really wanted to somehow migrate old paints because as it stands all stashes of them on existing saves will turn into puddles when this PR gets merged. If someone has a solution to this issue I am all ears.

Maybe not the solution you want, but you could make the liquid paint a new item and leave old paint as is (but obsoleted). That's pretty much a prerequisite for migrating it to a container with a liquid anyway. The game wouldn't know the phase changed on some already existing item.

Karol1223:
> Maybe not the solution you want, but you could make the liquid paint a new item and leave old paint as is (but obsoleted). That's pretty much a prerequisite for migrating it to a container with a liquid anyway. The game wouldn't know the phase changed on some already existing item.

I considered this but I would have to absolutely butcher the ID for black paint at that point and I am not sure if making the code less readable is better than letting existing paint drop to the floor as puddles. Frankly, I've never seen anyone use paint before, it is a very niche thing to stockpile so I suspect that not a whole lot of people will run into a problem if the PR stays as is. Unfortunately, I have to choose the lesser evil, it seems, and I am kind of torn on which is which...

mqrause:
I'm not sure what the consensus is on basically deleting existing items. If it's fine, it's fine. But the id thing should be easy to solve. Just make the new ones `paint_<color>` instead of `<color>_paint`

Karol1223:
> I'm not sure what the consensus is on basically deleting existing items. If it's fine, it's fine. But the id thing should be easy to solve. Just make the new ones `paint_<color>` instead of `<color>_paint`

I have no words to describe how stupid I feel for not thinking about changing the word order in IDs. Yeah, I mean, you're not wrong... Give me like 5 minutes...

data/json/items/containers.json
```diff
@@ -3328,5 +3328,55 @@
     ],
     "qualities": [ [ "CONTAIN", 1 ] ],
     "flags": [ "COLLAPSE_CONTENTS" ]
+  },
+  {
+    "id": "paint_can_steel",
+    "type": "GENERIC",
+    "category": "container",
+    "name": { "str": "metal paint can" },
+    "description": "A metal paint can holding a gallon of liquid.",
```
TheShadowFerret, line 3337:
```suggestion
    "description": "A metal paint can, able to hold a gallon of liquid.",
```
Can't have descriptions lying to you if the can is empty 😛

data/json/items/containers.json
```diff
@@ -3328,5 +3328,55 @@
...
+    ],
+    "qualities": [ [ "BOIL", 1 ], [ "CONTAIN", 1 ] ],
+    "flags": [ "COLLAPSE_CONTENTS" ]
+  },
+  {
+    "id": "paint_can_plastic",
+    "type": "GENERIC",
+    "category": "container",
+    "name": { "str": "plastic paint can" },
+    "description": "A plastic paint can holding a gallon of liquid.",
```
TheShadowFerret, line 3362:
```suggestion
    "description": "A plastic paint can, able to hold a gallon of liquid.",
```

TheShadowFerret (changes requested):

data/json/items/containers.json
Karol1223: Good catch, I have a bad habit of using thought shortcuts in text. I'll fix it right away.

PatrikLundell:
I suspect adding new colors may require additions elsewhere, such as walls of the new colors (and I think someone expanded wall painting to support additional wall materials fairly recently).

You've updated construction.json with changed components, but would probably need to also add new ones for new colors. I think you'd also need to expand the home_improvement.json file to contain the new colors.
I don't know what tutorial.json does, but a check for whether it needs the additional colors may be warranted.
Added colors in materials.json and/or toolset.json?

Karol1223:
> I suspect adding new colors may require additions elsewhere, such as walls of the new colors (and I think someone expanded wall painting to support additional wall materials fairly recently).

Another contributor has expressed interest in this back when I initially added black paint and upon talking to them they said they'll handle the new painted walls and doors once this PR gets merged.

I'll also say right now that painting walls needs a bit more change - with the changes in the volume of paint charges it should now take 3 paint to paint a wall as opposed to 25 (25 appears to have always been overkill regardless).

> Added colors in materials.json and/or toolset.json?

toolset.json I'll update in a second, you are right in that it warrants new colors being added there. I am not sure what tutorial.json and materials.json actually list them for, though, so I am not convinced I should be adding them there just yet without knowing what they're used for.

> I think you'd also need to expand the home_improvement.json file to contain the new colors.

I am also not sure I understand what home_improvement.json file you are referring to as the only one I see is under items and is the file where I have added the new paint colors already?

PatrikLundell:
home_improvements.json: My bad. Looking at the file's latest modification I saw only modifications, but the additions had been done in the original change...

PR #62959, opened 2023-01-01
Disallow towing by racked parts

### LINKED ISSUE #62953: Crash when (e)xamining the bikerack when towing cable is installed on the racked vehicle
### Describe the bug

When towing a vehicle with the towing cable installed on a bikeracked vehicle, the game crashes when you try to (e)xamine the bikerack.

### Steps to reproduce

1. Spawn a vehicle with bikerack and a bike
2. attach bike to bikerack
3. spawn another vehicle to be towed
4. attach a towing cable to the bike and the other vehicle
5. (e)xamine the bikerack

### Expected behavior

opening the actionmenu to be able to remove bike from rack, either stating that you cant remove bike while towing or removing towing cable in the process

### Screenshots

_No response_

- OS: Windows
- Game Version: 7f881d8 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

[debug.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10328690/debug.zip)

irwiss:
#### Summary
None

#### Purpose of change

Fixes #62953

#### Describe the solution

Disallow attaching tow cable to carried parts.

This adds a string for the failure message, not sure how strict the string freeze is, I don't think silently failing is too elegant though

#### Testing

Use tow cable, select racked vehicle tile as first or second vehicle - should fail, non-racked tiles should work as usual

### DIFF
--- a/src/iuse.cpp
+++ b/src/iuse.cpp
@@ -8478,6 +8478,10 @@ cata::optional<int> iuse::tow_attach( Character *p, item *it, bool, const tripoi
                     p->add_msg_if_player( _( "You can't attach the tow-line to an internal part." ) );
                     return cata::nullopt;
                 }
+                if( !source_veh->part( vp->part_index() ).carried_stack.empty() ) {
+                    p->add_msg_if_player( _( "You can't attach the tow-line to a racked part." ) );
+                    return cata::nullopt;
+                }
             }
             const tripoint abspos = here.getabs( posp );
             it->set_var( "source_x", abspos.x );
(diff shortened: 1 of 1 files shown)


PatrikLundell:
I don't find this to be a good hack. I believe it used to work, so it would be better to try to find out what caused it to fail.

The biking rack is usually at the rear end of vehicles, which is the place where you want to attack a tow cable. Attaching the cable further up the tower is bad because that reduces the distance between the vehicles which will probably increase the rate of collisions when the tower runs into slowing terrain substantially (and collisions now cause damage that can't even be repaired).

It's true that attaching a tow cable to a biking rack in real life will fail, probably because the rack is ripped off the car, but real life also allows to to attach the cable to the tow point below the rack, as real life doesn't work on rather boxy tiles.

irwiss:
the "hack" is modifying parts on carried tiles, which both makes no sense IRL as you mentioned and causes weird interactions with other parts of the code

i don't think 1 tile difference is a game breaker, if you wanna fix rear ending the bike rack there's other options;
1) place rack in front/on side
2) adjust tow cable length in `vehicle::no_towing_slack` and deceleration distance in `vehicle::do_towing_move` to try prevent rear ending

PatrikLundell:
Neither racks in the front nor to the side are acceptable work arounds. The front end will ram things from time to time, and you don't want neither your rack nor a precious carried vehicle to take the blow. Making the vehicle two tiles wider (when the rack is used) makes the vehicle very difficult to get past all the stuff blocking roads.

PR #55928, opened 2022-03-07
Exclude grinding from `cooked_acorns` recipe

DanielVartanov:
#### Summary
Content "Cooked acorns is no longer a meal and does not require grinding"

#### Purpose of change

Unlike common edible nuts, acorns need to be leached to get rid of tanines, otherwise they are too bitter. Basically, current recipe in DDA is:
- Shell the acorns
- Chop them
- Boil 
- Toast until dry
- _Grind into meal (with a quern or mortar&pestle)_

The main point of this PR is that _grinding is unnecessary_. This PR deliberately removes grinding from the recipe and carefully inspects the consequences of it.

##### Motivation

Being kind of trapped recently in a shelter next to the woods I had to dive into acorns topic quite deeply. I have read relevant discussions https://github.com/CleverRaven/Cataclysm-DDA/issues/24255 and https://github.com/CleverRaven/Cataclysm-DDA/pull/24261 with great care.

At first, let me cite a comment https://github.com/CleverRaven/Cataclysm-DDA/pull/24261#issuecomment-406672123 from the aforementioned discussion (emphasis mine):

> **Chopping them up roughly allows you to leach the tannins**, but that's not acorn meal

Secondly, all you can do with `acorns_cooked` is:
- Add them as an ingredient to soups
- Mill them into flour (with a quern or a windmill/watermill)

Which means that grinding already cooked acorns into a meal is really unnecessary. 

##### What this PR changes exactly

* `quern` and `food_processor` were removed from the list of tools needed for the recipe.

* The word "meal" was removed from the display-name of the `acorns_cooked` item.

* recipe `time` was reduced from 24 to 19 mins (some advice is welcome if anyone knows better).
* `batch_time_factor` is adjusted due to now it's mostly boiling which scales greatly.

* A quality of `HAMMER` level 1 was added to the recipe to match up with other hard-shell nuts, indeed one needs to hull the acorns. It is, by the way, related to a recent PR of https://github.com/CleverRaven/Cataclysm-DDA/pull/55888, here is an IRL illustration of the how acorns are shelled with mortar&pestle: https://www.youtube.com/watch?v=eB22HfpBVXY

* A quality of `BOIL` level 1 was added since leaching is done via boiling the acorns.

* Also `roasted_acorns` changed to be made from cooked acorns instead of raw acorns and therefore `prof_knife_skills` is excluded from its proficiencies as the cutting is done at a stage of preparing cooked acorns.

#### Describe alternatives you've considered

[...]

PR #62963, opened 2023-01-02
Changes roasted acorn description

ghost:
Updates the description for roasted acorns.

#### Summary
Bugfixes "Roasted acorns description"

#### Purpose of change
Updates the description to more clearly show the multi-step process involved. 

I may be entirely wrong about this. I've never eaten acorns of any sort.

#### Testing
Checked in game. New description is displayed.

### DIFF
--- a/data/json/items/comestibles/nuts.json
+++ b/data/json/items/comestibles/nuts.json
@@ -500,7 +500,7 @@
     "id": "acorn_roasted",
     "name": { "str_sp": "roasted acorns" },
     "copy-from": "acorns",
-    "description": "A handful of shelled and roasted acorns. Soaked, boiled and roasted. There's hardly any bitterness left.",
+    "description": "A handful of shelled and roasted acorns.  Soaked, boiled and roasted.  There's hardly any bitterness left.",
     "price": 90,
     "price_postapoc": 200,
     "flags": [ "EATEN_HOT", "EDIBLE_FROZEN" ],


Maleclypse (dismissed):

flybye3o6:
most irl recipes iv seen require you to boil or leach acorns to remove the tannins, iv not seen a single recipe that allows you to cook acorns before doing that, if you know of any methods that allow you to cook acorns without the lengthy process of removing the tannins id like to know 
i did some research on this when i was working on #62052

RenechCDDA:
#55928, which *replaced normal acorns with cooked acorns* details the logic behind using cooked acorns. Apparently they are inedible without that preparation. Acorns in particular have come up as a topic several times and I believe #55928 is correct on all counts.

The description was not changed, so if anything the description if erroneous.

intergalacticdata:
Nope, I've done zero research at all. I just couldn't understand why the roasted recipe said 'a handful of nuts from a tree' and the cooked acorns recipe describes turning it into paste etc. As this is intentional, I suggest updating the description for roasted acorns instead so that it is more clear.

flybye3o6:
you could change this PR into doing that, or i can take a look at it myself

intergalacticdata:
> 

If you have a suggested wording, I'll just throw that in this and revert the ingredient change, or you can just handle it if you prefer. No difference to me.

flybye3o6:
well they kinda look like larger darker peanuts when roasted with a wood looking texture
"A handful of roasted nuts from an oak tree. A smoked earthy flavored treat after all the work that went into remove the bitterness."
im not good with descriptions
btw eating acorns without removing the tannins can cause inflammation in your digestive tract

intergalacticdata:
For reference, this is the text from cooked acorns "A serving of acorns that have been hulled, chopped, and boiled in water before being thoroughly toasted until dry. Filling and nutritious."

The "from a tree" part seems confusing to me as it kind of implies directly from the tree, whereas these are shelled and chopped.

How about:
"A handful of shelled and roasted acorns. A smoked, earthy flavored treat. There's hardly any bitterness left."

PR #62966, opened 2023-01-02
debug: add action for generating minimized save archive

andrei8l:
#### Summary
None

#### Purpose of change
I want to make `Save file` a mandatory field in the bug report template, but we need to make it easy for people to provide that.

Save folders can get really large and most of their data is not needed for specific bugs.

#### Describe the solution
Add a debug action that creates a `tar.gz` archive of the current save folder with trimmed map buffer and map memory, as well as the config folder.

Uses a hand-rolled, bare-bones implementation of UStar for archiving and `zlib` for compressing which creates an archive accepted by github.

Not archiving/compressing at all: this still leaves a critical step to the user

Archiving to zip with minizip: this needs third-party code but is otherwise superior in every way (except compression ratio). Adding minizip to the repo would also pave the way for archiving save folders by default.

#### Testing
<details>
<summary>Quick usage demo</summary>

[output.webm](https://user-images.githubusercontent.com/68240139/210305062-6b98e0af-fbf3-4f2a-be6b-9a1b558e5b04.webm)

(I captured this from a debug build so it's intentionally slower than normal)
</details>

Major `UStar` implementations (GNU tar, libarchive, and 7zip) can unpack the resulting archive without error or warnings.
File integrity after unpacking can be verified with `md5sum` or similar.
Game still loads the trimmed folders.
Github accepts the archives as attachments.

Tested on linux and with cross-mingw.

#### Additional context
N/A

### DIFF
--- a/src/debug_menu.cpp
+++ b/src/debug_menu.cpp
@@ -9,11 +9,13 @@
 #include <chrono>
 #include <csignal>
 #include <cstdlib>
+#include <functional>
 #include <iomanip> // IWYU pragma: keep
 #include <iostream>
 #include <iterator>
 #include <limits>
 #include <list>
+#include <locale>
 #include <map>
 #include <memory>
 #include <new>
@@ -97,6 +99,7 @@
 #include "stomach.h"
 #include "string_formatter.h"
 #include "string_input_popup.h"
+#include "tgz_archiver.h"
 #include "trait_group.h"
 #include "translations.h"
 #include "try_parse_integer.h"
@@ -207,6 +210,7 @@ std::string enum_to_string<debug_menu::debug_menu_index>( debug_menu::debug_menu
         case debug_menu::debug_menu_index::WRITE_TIMED_EVENTS: return "WRITE_TIMED_EVENTS";
         case debug_menu::debug_menu_index::SAVE_SCREENSHOT: return "SAVE_SCREENSHOT";
         case debug_menu::debug_menu_index::GAME_REPORT: return "GAME_REPORT";
+        case debug_menu::debug_menu_index::GAME_MIN_ARCHIVE: return "GAME_MIN_ARCHIVE";
         case debug_menu::debug_menu_index::DISPLAY_SCENTS_LOCAL: return "DISPLAY_SCENTS_LOCAL";
         case debug_menu::debug_menu_index::DISPLAY_SCENTS_TYPE_LOCAL: return "DISPLAY_SCENTS_TYPE_LOCAL";
         case debug_menu::debug_menu_index::DISPLAY_TEMP: return "DISPLAY_TEMP";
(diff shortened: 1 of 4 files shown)


GuardianDll:
Can it be also added into player's buffer, same as game report?

andrei8l:
> Can it be also added into player's buffer, same as game report?

Not sure. I can't get that to work - pasting only adds the file path as text without uploading anything.

PatrikLundell:
I can see a few issues with this:
- Resulting files being too large for the anemic site file limit despite the effort, blocking the ability to make reports if it's absolutely mandatory (can be worked around by providing a dummy file, but that's probably not immediately apparent to everyone).
- Requirement to use Unix style tar plus zip, expanding the mandatory set of functionality beyond zip. Do all widely used compression SW for non Unix lookalikes support tar compression? I think there's some crappy zip support built into Windows, for example.
- Confusion when bug investigation requires some of the omitted data.
- Potential user confusion if they try to use the file to move a save to some other machine (although it should be rather apparent that it doesn't work, but not necessarily obvious what the reason is).

andrei8l:
> Resulting files being too large for the anemic site file limit despite the effort, blocking the ability to make reports if it's absolutely mandatory (can be worked around by providing a dummy file, but that's probably not immediately apparent to everyone).

I haven't yet encountered the case of the archive being too large and I'm not sure it's even possible in regular play. If I do see that case, I can add a contingency path that trims the map buffer even more (to just the current segment or even less).

Regarding the dummy file, my current plan is to use fluff text along the lines of 
<details>
<summary>fluff text</summary>

```
Write N/A if not applicable, but note:

- your bug report is unlikely to get resolved without a save file, unless it's trivially 
reproducible or your reproduction steps are complete.
```
</details>

So people can write `N/A` and skip that field. People sometimes write garbage in the other mandatory fields too so it's not a hard blocker.


> Requirement to use Unix style tar plus zip, expanding the mandatory set of functionality beyond zip. Do all widely used compression SW for non Unix lookalikes support tar compression? I think there's some crappy zip support built into Windows, for example.

Doesn't matter. The archive is only for contributors and developers so adding an archiving tool to their kit is not a problem. But AFAIK: 
 - all linux and BSD distros can handle `tar.gz` archives by default
 - MacOS should be able to handle it since they're the driving force behind `libarchive`
 - For Windows, the most popular programs (7zip, [WinZip](https://www.winzip.com/en/learn/file-formats/tar-gz/), and [WinRar](https://www.win-rar.com/open-gz-file.html?&L=0)) can handle `tar.gz` archives

> 
> Confusion when bug investigation requires some of the omitted data.

Not sure what you mean.

> Potential user confusion if they try to use the file to move a save to some other machine (although it should be rather apparent that it doesn't work, but not necessarily obvious what the reason is).

Generating this archive needs an explicit action in the debug menu so if there's any confusion about the purpose of the archive, it's likely beyond my ability to prevent.

PatrikLundell:
The file requirement not being absolute solves the blocking issue.

I don't understand your comment about the archive only being for contributors/developers. If a bug report has an (almost) mandatory archive file requirement every bug reporter will have to be able to generate the archive. Sure, processing the archive is done by a limited set of people, but not the generation.

Confusion about omissions: When an investigator determines omitted data is needed and asks for it, and the bug reporter confusedly says they'd supplied the zip file with all the data (not knowing the process strips parts of it away). Can be managed by explaining that parts are stripped away in these cases, but they'll probably happen. It's unlikely to be much of a problem, though.

I agree it's not your task to prevent people from trying to use the report files for other purposes they aren't suitable for, but I certainly wouldn't be surprised if it happened.

andrei8l:
> If a bug report has an (almost) mandatory archive file requirement every bug reporter will have to be able to generate the archive

That's why this PR is here. This code generates the archive we need for bug reports and doesn't require anything else from the OS or the player. 

If you think this is insufficient, please tell me exactly what else it needs. Trimming is currently fairly conservative but we have room to trim even less (I'm seeing compression ratios of 5+ so we have significant leeway).

> When an investigator determines omitted data is needed and asks for it, and the bug reporter confusedly says they'd supplied the zip file with all the data

We'll deal with that edge case the same way we do now: ask for the (full) save file, get ignored, put the issue on the backburner, and move on. Overall it's still an improvement.

PatrikLundell:
Ah, I didn't realize you included the full compression code here, but thought you were using external resources expected to be present.

andrei8l:
The archive now includes the config folder

andrei8l:
Thanks for merging!!

Raikiri:
is there an easy way for the user to test their compressed save file to see if the issue is still reproducible?

andrei8l:
Rename the original folder then unpack the archive in place with your favorite archiving software. On Windows, you might need 2 steps - decompress then unarchive (strange Windows software smh).

But you're not expected to do that. If the issue reproduces with the full save folder, it should reproduce with the archive as well or else there's a serious flaw in this feature.

Raikiri:
Should there be an in-game way to load archived savegames without any OS-level steps? As in, do the decompression in memory (presumably with the same library that was used to compress) and load. 

I also imagine it can be useful to generate these archives with a hash in the name to be able to run multiple save files for side-by-side comparisons without them conflicting.

andrei8l:
First one needs a lot of effort that would be worthwhile if we also started archiving save folders by default, but `UStar` is not suitable for that and other formats need third-party libraries.

Yes for the second one.

PR #62972, opened 2023-01-03
added lengths for EXODII Mags and the MDS

bombasticSlacks:
#### Summary
None

#### Purpose of change
Brought up on discord that the EXODII mags were fitting in places they probably shouldn't

#### Describe the solution
Added lengths to the EXODII Mags. 

10 round clip is just the diameter of 10 30-06 rounds.

Length of the 60 rounder is the same as the EXODII one, a reasonable approximation.

For the HWP I looked at specs for 1.5L dry bags and used a similar height. 

#### Additional context
This doesn't address the issue brought up on discord but does make things more true to life, which is the goal.


Dan1ss1mo:
Shouldn't hwp also be tweaked to only accept hub 12.3 mags?

PR #62982, opened 2023-01-03
Fix typos in error message

reinerh:
#### Summary
None

#### Purpose of change

Fixes two typos.


LyleSY:
Makes sense but I'm not sure if this falls under string freeze

LyleSY (approved):

Issue #54683, opened 2022-01-23
Game froze for several minutes trying to spawn an npc


moxian:
### Describe the bug

I was playing around in debug mode.

- started new game (all default survivor evacuee)
- revealed overmap
- teleported away from the shelter
- started crafting long cordage pieces
- game unexpectedly froze on 95% completion of 14th craft. After the craft completion the time is 07:00:49 PM which might be important?..

Inspecting the running process got me this stacktrace:

```
0x0000000000000000
cataclysm-tiles.exe!std::_Hash<std::_Umap_traits<pf::`anonymous namespace'::node_address,pf::A0x452dcbd1::navigation_node,std::_Uhash_compare<pf::`anonymous namespace'::node_address,pf::A0x452dcbd1::node_address_hasher,std::equal_to<pf::`anonymous namespa
cataclysm-tiles.exe!pf::find_overmap_path+0x744
cataclysm-tiles.exe!overmapbuffer::get_travel_path+0xef
cataclysm-tiles.exe!npc::set_omt_destination+0x60a
cataclysm-tiles.exe!npc::long_term_goal_action+0xb1
cataclysm-tiles.exe!game::perhaps_add_random_npc+0x3eb
cataclysm-tiles.exe!do_turn+0x414
cataclysm-tiles.exe!WinMain+0x825
cataclysm-tiles.exe!__scrt_common_main_seh+0x106
KERNEL32.DLL!BaseThreadInitThunk+0x14
ntdll.dll!RtlUserThreadStart+0x21
```

trying to get a peek of the trace again interminently has or doesn't have the first hashmap-related frame, but the rest (find_overmap_path and below) are still present.

While filing this report the game did unfreeze (presumably, succeeded in finding the path?). But it took longer than my patience, and in a normal run i would just angrily force-close the game.

### Steps to reproduce

n/a

### Expected behavior

Either:
- no freeze
- a warning that a freeze might happen
- the freeze happening at a more.. expected point in time. Like when trying to get to survivor center at the shelter console.
- a "hold on, it's taking a bit longer than we expected, we are not frozen, promise! Here, get a progressbar: 23342/13123123" kind of popup.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: 0.F-5469-g8ba04d36fe [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
[...]

PR #62985, opened 2023-01-04
Add spinner for overmap pathfinding

irwiss:
#### Summary
None

#### Purpose of change

Mitigate #54683 slightly - game process can appear to be hanging while the pathfinding code is running.

#### Describe the solution

Draw a spinner every 100ms while the overmap A* is running, letting the player know the game isn't frozen.

This adds no strings ( `Hang on a bit…` is stolen from animation.cpp ).

#### Testing

Write some code to trigger `game::perhaps_add_random_npc` with `ignore_spawn_timers_and_rates` set to true.

I can't add debug menu option to trigger this at this moment; due to string freeze and for some reason debug menu is translated.

#### Additional context
cataclysm-tiles_Cgoge9sHIn

### DIFF
--- a/src/simple_pathfinding.cpp
+++ b/src/simple_pathfinding.cpp
@@ -276,8 +276,9 @@ omt_score::omt_score( int node_cost, bool allow_z_change ) : node_cost( node_cos
 
 simple_path<tripoint_abs_omt> find_overmap_path( const tripoint_abs_omt &source,
         const tripoint_abs_omt &dest, const int radius, const omt_scoring_fn &scorer,
-        cata::optional<int> max_cost )
+        const std::function<void( size_t, size_t )> &progress_fn, const cata::optional<int> &max_cost )
 {
+    cata_assert( progress_fn != nullptr );
     simple_path<tripoint_abs_omt> ret;
     const omt_score start_score = scorer( source );
     const omt_score end_score = scorer( dest );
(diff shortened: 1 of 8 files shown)


src/simple_pathfinding.cpp
```diff
@@ -291,9 +291,23 @@ simple_path<tripoint_abs_omt> find_overmap_path( const tripoint_abs_omt &source,
...
+
+    auto last_report = std::chrono::high_resolution_clock::now();
+    int progress = 0;
     while( !open_set.empty() ) {
         const node_address cur_addr = open_set.top().addr;
         open_set.pop();
+        if( progress++ >= 100 ) { // stagger progress checks to 1 in 100 nodes
+            progress = 0;
+            if( progress_fn != nullptr ) {
+                const auto now = std::chrono::high_resolution_clock::now();
```
Qrox, line 303: According to https://en.cppreference.com/w/cpp/chrono/high_resolution_clock, `high_resolution_clock` is not always steady (i.e. time only goes forward and is unaffected by time sync). If we use `high_resolution_clock`, and the user's system happens to sync time backwards during pathfinding, they may again experience a freeze, so `steady_clock` should be used instead.

irwiss: It appears to be [aliased to steady_clock since VS 2015](https://learn.microsoft.com/en-us/cpp/standard-library/chrono?view=msvc-170#:~:text=high_resolution_clock%20is%20now%20a%20typedef%20for%20steady_clock%20In%20the%20Microsoft%20C%2B%2B%20implementation%2C), not sure about other compilers though, I changed it to steady_clock

Other parts of cata codebase uses a bunch of `high_resolution_clock`; specifically sdl sounds has a bunch of those which is where I peeked at trying to measure intervals, those worth fixing? They did seem to work fine for the most part

PR #63008, opened 2023-01-05
rebreather mask padding 

Dan1ss1mo:
#### Summary
Bugfixes "rebreather mask fix"

#### Purpose of change
Wearing rebreather mask considered to be uncomfortable without somthing under it

#### Describe the solution
"PADDED" flag

#### Testing
~~trust me~~ spawned mask, it is no longer rubbing against my teeth and lips

### DIFF
--- a/data/json/items/tool_armor.json
+++ b/data/json/items/tool_armor.json
@@ -1582,7 +1582,7 @@
     "color": "dark_gray",
     "name": { "str": "rebreather mask" },
     "description": "A mask worn over your mouth which, when loaded with the proper filters, recycles your exhaled breath for rebreathing while underwater.  Use it to turn it on.",
-    "flags": [ "WATER_FRIENDLY" ],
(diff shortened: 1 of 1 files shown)


PatrikLundell:
I trust very few people (actually, none) to never make any mistakes anywhere. Thus, I would have to know you extremely well to trust that it works without a basic level of testing.

Dan1ss1mo:
> I trust very few people (actually, none) to never make any mistakes anywhere. Thus, I would have to know you extremely well to trust that it works without a basic level of testing.

Sure, I'll test it if it is the case)

Dan1ss1mo:
> I trust very few people (actually, none) to never make any mistakes anywhere. Thus, I would have to know you extremely well to trust that it works without a basic level of testing.

done, everything working as intended

Maleclypse (approved):

PR #63020, opened 2023-01-07
Swap out the colors for the `|.` and `\.` damage states

moxian:
#### Summary
Interface "Swap out the colors for the `|.` and `\.` damage states"

#### Purpose of change

When viewing the damage colors side by side, i realized that the `|.` color is darker and feels more "damaged" than the `\.` one, although in actuality it's the opposite.
It also turns out that i'm not alone in my thinking and that at least two other people were similarly confused (https://discord.com/channels/598523535169945603/642349261320880128/1061084321093517372)
20230107-014707-Discord

So i thought it would be nice to swap the colors around to the more "natural" progression, on the assumption that not too much thought were put into them originally.

#### Describe the solution

Swap the colors around when determining which one to use for the damage symbol. Two-line change.

Leave things as is?

#### Testing

20230107-013535-cataclysm-tiles

### DIFF
--- a/tests/item_tname_test.cpp
+++ b/tests/item_tname_test.cpp
@@ -344,7 +344,7 @@ TEST_CASE( "item health or damage bar", "[item][tname][health][damage]" )
 
             // magenta `|.`
             THEN( "it appears moderately damaged" ) {
-                CHECK( shirt.tname() == "<color_c_magenta>|.</color>\u00A0long-sleeved shirt (poor fit)" );
+                CHECK( shirt.tname() == "<color_c_light_red>|.</color>\u00A0long-sleeved shirt (poor fit)" );
             }
         }
 
@@ -355,7 +355,7 @@ TEST_CASE( "item health or damage bar", "[item][tname][health][damage]" )
 
             // red `\.`
             THEN( "it appears heavily damaged" ) {
-                CHECK( shirt.tname() == "<color_c_light_red>\\.</color>\u00A0long-sleeved shirt (poor fit)" );
+                CHECK( shirt.tname() == "<color_c_magenta>\\.</color>\u00A0long-sleeved shirt (poor fit)" );
             }
         }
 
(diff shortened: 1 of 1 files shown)


PatrikLundell:
I agree with the logic.

However, how does it match up with other usages of damage indications, such as e.g. with limbs? In widgets? It's preferable to use the same scheme throughout. Or are all these usages just referring back to the parts you just changed?

moxian:
The limb damage has its own color grading - it does not use the magenta for the 50% damage state, and does use a red for "basically broken" (green -> light green -> yellow -> light red -> red ([code](https://github.com/CleverRaven/Cataclysm-DDA/blob/c4b37691185b32b230c8ce8692fab89f0acc2892/src/character_body.cpp#L1168), [other code](https://github.com/CleverRaven/Cataclysm-DDA/blob/c4b37691185b32b230c8ce8692fab89f0acc2892/src/output.h#L615))). I think it's fine to leave it as is, and not unify with the item damage, just because i like the blood red so much (and it makes much more sense in limb context than in item context)

20230107-152710-cataclysm-tiles

Stamina uses the same grading as health (although there is [code](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/output.cpp#L2640) to have it be blueish rather than greenish?? Which is unused?..). I think it's fine to leave stamina consistent with health and inconsistent with items.

The other place i can think of that uses this is vehicle health oveview. It uses slightly different color progression from both of the above (green -> yellow -> magenta -> red ([code](https://github.com/CleverRaven/Cataclysm-DDA/blob/c4b37691185b32b230c8ce8692fab89f0acc2892/src/veh_interact.cpp#L3090))):


There is also ammo/battery charge numbers ([code](https://github.com/CleverRaven/Cataclysm-DDA/blob/c4b37691185b32b230c8ce8692fab89f0acc2892/src/item.cpp#L6705)) which go white -> light green -> yellow -> red -> light red. And perishable food that goes cyan -> yellow -> brown..
[Weariness](https://github.com/CleverRaven/Cataclysm-DDA/blob/c4b37691185b32b230c8ce8692fab89f0acc2892/src/display.cpp#L664) (green -> yellow -> light red -> light red -> red -> red), [Hydration](https://github.com/CleverRaven/Cataclysm-DDA/blob/c4b37691185b32b230c8ce8692fab89f0acc2892/src/display.cpp#L731) (green -> greeen -> nothing -> yellow -> yellow -> light red -> light red), hunger, focus...

I... I'm not sure this is feasible to unify at this point. Or if it is, i'm probably not up to the task..

PatrikLundell:
Thanks for investigating the different mappings.

I'm not a great fan of the magenta color due to its low contrast, but as your investigation shows, there is a fair number of color schemes, so it would be a larger task to try to whittle them down to a smaller number of patterns, and that would be beyond the scope of this PR.

Issue #59679, opened 2022-07-27
(sacrilege warning) remove 'suicide' from root level keybindings


I-am-Erk:
### Is your feature request related to a problem? Please describe.

This is part of a larger effort to identify places where our "top level" keybindings have been mapped to things that players do not need rapid, single-keypress access for. This is a discussion: all these issues are valid places to raise concerns about reasons a player should have immediate single-key access to this function.

Long and short: nobody needs an immediate access key for suicide. Despite q-y-y being a meme, it's not a good use of a top level key.

### Solution you would like.

Move "suicide this character" to the ESC menu to free that binding for another use

### Describe alternatives you have considered.

We could also move "S"ave and quit off the default keybindings, but that is a more contentious one.

It would be nice if there were some semi-humorous ways to get the same effect such as attempting to drink bleach.

NetSysFire:
I approve of this and think this maybe even should get done in time for 0.G, since 0.G may attract new players.
"Alright lets quit for now and call it a day. Suicide? What? Is this how you quit?"
A quick solution would be to move Save and quit to this and make ctrl+s just a regular quicksave or something.

I-am-Erk:
Yeah, that is a pretty nice move actually. I don't remember what lower-case `q` is bound to but another option might be to do `q`uicksave, `Q`uit and save

Actually from loading my game it seems `q` is unbound, probably because of how dangerous `Q` is, so that should be a pretty easy fix. Then that frees up `shift+S` for some as-yet undetermined use (I would like it to become Sleep/Wait, or maybe Fighting Style, but I don't want to remap too much stuff in one pass)

pjf:
If we could have the same functionality, but call it "abandon character", I would personally appreciate it. I've lost too many friends iRL, and somehow the wording of the `Q` keybinding still hits me every time.

liminalRoar:
`Shift+S` could be Spell, hehe, since that is undedicated although i suppose that has no use outside of Magiclysm

I don't really see a need for any dedicated default bindings for quitting and think it's fine on the ESC menu, at least if we have better, frequent use cases for the "Q"s and "S".

NetSysFire:
>Shift+S could be Spell, hehe, since that is undedicated although i suppose that has no use outside of Magiclysm

Some mutations have spells. E.g the leap spell when you have frog legs.

liminalRoar:
[...]

PR #63026, opened 2023-01-07
Unbind Abandon character

ghost:
#### Summary
Interface "Remove abandon character default keybind, and set `q` to look at sky"

remove default keybind for abandon character and give `q` to look at sky, knowing the weather in more detail is useful imo.
Removed default keybind from abandon character due to it being unnecessary, leaving `Q` for perhaps something more useful later on?

#### Purpose of change
i don't think we need a dedicated button for abandonment (sorry old timers, i know, this is blasphemous)
look at sky is a useful feature not many know (from my knowledge at least) setting it to q allows it to be more easily discovered.

#### Describe the solution
simple json edit, assuming i did it correctly. rebind the keybinds in raw/keybinds.json

#### Testing
opened the game with the changes, and by default abandonment has no keybind, while look at sky is set at q.

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -2780,8 +2780,7 @@
     "type": "keybinding",
     "name": "Quicksave",
     "category": "DEFAULTMODE",
-    "id": "quicksave",
-    "bindings": [ { "input_method": "keyboard_char", "key": "Q" }, { "input_method": "keyboard_code", "key": "q", "mod": [ "shift" ] } ]
+    "id": "quicksave"
   },
   {
     "type": "keybinding",
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -2540,8 +2540,7 @@
     "type": "keybinding",
     "name": "Abandon character",
     "category": "DEFAULTMODE",
-    "id": "SUICIDE",
-    "bindings": [ { "input_method": "keyboard_char", "key": "Q" }, { "input_method": "keyboard_code", "key": "q", "mod": [ "shift" ] } ]
+    "id": "SUICIDE"
   },
   {
     "type": "keybinding",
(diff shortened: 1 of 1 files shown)


Night-Pryanik:
I'd leave `Q` as a keybind for abandoning character.

Devengel:
i think it is unnecessary for abandon character to have a default keybind. if someone really wants a hot key to quickly end their characters, they can assign one themselves. but i doubt it is something many people readily want it to be default.

moxian:
On at least a couple of occasions i have found myself fat-fingering and *almost* accidentally abandoning the character. Luckily neither of those were fatal, but they were all scary nevertheless. I think removal of the easy to access suicide button is a good idea. It's not something you do often enough to need a dedicated keybind.
Without the keybind you can still go `Enter` -> actions -> abandon the character to achieve the same goal.

Rimplestilt:
Suicide has no reason for a keybinding and instead should be an option accessible from the ESC menu, with a possible tucked away default key, like a function key.

There are more useful keybinds for "Q" imo, but as various ppl have various views on what is useful, i submit "q" and "Q" be left unbound by default for now. (for example: useful in spellcasting in magiclysm, or setting q and Q to cycle run speed up and down once, or bringing up character sheet, the list goes on)

Especially with the possibility that there is more drastic rework in both menu design and ui, and also keybindings, with possible entirely new keybinding sets (for example once more designed more modern keyboard usage with an emphasis on WASD or arrow key style gameplay)

Golfavel:
This would close #59679.

Devengel:
sky is already W on overmap it seems, Bound save and quit to Q, and quicksave to S. as per #59679 recommendations.

Fris0uman:
Why also change S? just binding Q to `Q`uick save is a way more straight forward solution

Fris0uman:
The PR is about unbiding the suicide key for safety, there is an unbound action that start with a Q which is Quick save, it doesn't make sense to alter another key that is just working fine.

Devengel:
failure is unrelated

kevingranade:
Do one thing at a time please.

Devengel:
alright, i'll leave the quicksave unchanged then.

PR #63037, opened 2023-01-08
Clarify the small body graph

Kamayana:
#### Summary
Interface "Edit the small body graph for clarity"

#### Purpose of change
The small body graph used in the Structured sidebar uses tiny dots for the hands and mouth which makes it quite hard to determine their color. Also, the eyes and mouth are depicted floating outside of the head, which confused me for a while when I first started using it.

#### Describe the solution
image

I edited all the parts to be large enough to make their colors clear. I made generous use of unicode characters for aesthetics, making sure all the characters used are available in all three of the fonts included with CDDA.

I went through plenty of experimentation. I ultimately settled on the one above for its clarity and appeal, but this is definitely pretty subjective. I also could've used less unicode, but the design document specifically says not to avoid using them.

smallBodyGraphAlternates

#### Testing
All these screenshots are from in-game.


moxian:
Opinion: I *really* like the `(°-°)` one (first row, third column). It feels much more recognizable as a face than the looking-sideways one (first row, 4th column and friends).
The mouth can be adjusted to be thicker if `-` is too thin and deemed a problem, but i think having both eyes is very good for readability.

Kamayana:
> Opinion: I _really_ like the `(°-°)` one (first row, third column). It feels much more recognizable as a face than the looking-sideways one (first row, 4th column and friends). The mouth can be adjusted to be thicker if `-` is too thin and deemed a problem, but i think having both eyes is very good for readability.

image

Kamayana:
I think I agree with Moxian that the front face is more intuitive, so I'm gonna go with that as the default, but I like my profile face enough that I want to make it an alternative. However, the current way the Structured layout is set up makes that untenable, so I'm going to expand the scope of this PR and make some changes to Structured as a whole.

Kamayana:
I had a change of heart and am just going to use the front-facing head (⁰⎼⁰) and call this PR done - I'm still working on other changes but because of string freeze I'll do a separate PR for that.

mlange-42:
Would have preferred the topmost standalone side view version, or the one in the top right corner of the table (original sidebar author here).

PR #63044, opened 2023-01-09
fixes dark sight's duration increment

b3brodie:
#### Summary
Bugfixes "Changes the Dark Sight spell's duration increment from 2000 to 45000"

#### Purpose of change

Dark Sights min duration is 100,000 and max duration is 1,000,000, yet its duration increment is 2,000.  With a max level of 20, leveling the spell up 20 times results in a duration of 140,000, which is far shorter than the max duration.  Therefore, the json defines a max value that is impossible to be reached.  Furthermore, such a small difference in duration leaves little reason to level the spell outside of the initial successful cast chance improvements, as only duration is changed by leveling the spell (not cast cost, speed, etc).

#### Describe the solution

Changing the duration increment to 45,000 linearly bridges the difference between the min and max durations over 20 levels of improvement, which for the player is roughly a 50% increase of the initial spell duration per level.

Leave no reason to level dark sight and incorrect json definitions

#### Testing

100,000 + (45,000 * 20) = 1,000,000

#### Additional context

N/A

### DIFF
--- a/data/mods/Magiclysm/Spells/classless.json
+++ b/data/mods/Magiclysm/Spells/classless.json
@@ -103,7 +103,7 @@
     "max_level": 20,
     "min_duration": 100000,
     "max_duration": 1000000,
-    "duration_increment": 2000
+    "duration_increment": 45000
   },
   {
     "id": "megablast",


Termineitor244:
So...  Same min/max/increment duration as a magical light? Sounds reasonable I guess.

Termineitor244 (approved):

Issue #63045, opened 2023-01-09
"Remove Gravel" construction blueprint is sysiphean (and has a bonus "Make gravel floor" bundled in)


moxian:
### Describe the bug

I have set up some construction blueprints to remove gravel from the would-be field, and sent NPC to work on them. They complained a bit about not having tools, but otherwise were doing something. From time to time i was getting notifications that "Remove gravel" and "Make gravel floor" tasks were getting complete. The latter one was weird to see, but i chalked it up to a display error or something. The progress was also rather slow, but i thought that's just how NPCs be (i don't have much experience with them really).

Until just now i was passing by and noticed that there is an in-progress construction on a dirt field. And that is a "make gravel floor" construction. Which would explain all of the above issues (slow progress, interminent lack of tools).

Except i don't have any "make gravel floor" blueprints defined. I have no idea why the NPC decided it was a good idea. This is a bug.

### Steps to reproduce

Load save file, examine the `Y` zones, see that the defined zones are all "remove gravel"
Look at the construction in progress nearby, see that it is "Make gravel floor" not "remove gravel"

Wait for an hour. See the gravel floor construction get finished, and a gravel removal one get started.

Wait another hour. See the gravel removal get finished, and gravel construction get started...

This repeats seemingly infinitely...

### Expected behavior

No gravel on my farmland >:(

### Screenshots

![20230109-041557-cataclysm-tiles](https://user-images.githubusercontent.com/7831163/211241240-1d745211-9c1c-4771-8cc0-bbb4e2dd508e.png)
![20230109-041611-cataclysm-tiles](https://user-images.githubusercontent.com/7831163/211241243-b5437833-c6af-4794-a1f7-fb5c82b3b55a.png)
![20230109-041618-cataclysm-tiles](https://user-images.githubusercontent.com/7831163/211241252-59a5c7e9-391d-4a27-b2e4-a9407aa4a085.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.2364 (21H2)
- Game Version: ba42c0a [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

### Additional context

[...]

PR #63049, opened 2023-01-09
activity/constr: don't loop prereq with only pre_special

andrei8l:
#### Summary
None

#### Purpose of change
Construction activity can loop between starting and resulting terrains if there's a matching construction definition with only `pre_special`

Fixes: #63045

#### Describe the solution
Add a check for this specific type of loop
Add a test for this case
Add test data to preserve test validity (one test was already not checking correctly due to json changes)

N/A

#### Testing
Test units

<details>
<summary>Manual testing</summary>

Spawn 1000 gravel and a shovel, add a construction zone for `Remove Gravel` over a grass tile and start the construction activity. Nothing should happen. Change the terrain to `t_railroad_rubble` and try again. This time the construction should start.

</details>

#### Additional context
N/A


dseguin:
Woops, added the Freeze label by mistake. Didn't realize this was all test data

dseguin (approved):

andrei8l:
This breaks a few of constructions such a `Build Small Railroad Track` and `Dig Grave and Bury Sealed Coffin` where the genuine prerequisite only has a `pre_special`.

andrei8l:
This is not going to work. There is insufficient data in the construction definition to break this kind of loop.

A possible solution is to add a flag that marks a definition as a deconstruction or explicitly single-step construction.

PR #63051, opened 2023-01-09
AIM: Add density sort, rename damage sort

### LINKED ISSUE #63032: AIM: sort by density

I can have enormous bags on my back and I am pushing a cart around. I want to haul as much as I can. For that I need to put the heavy items in the cart and the light in my bags.

### Solution you would like.

Sort by density
image

I don't even need a new column for density, I just need to sort the items, I don't care about density, only which items are denser.

My way of doing this is approximate density on the fly:

1. sort by weight
2. look at the volume column
3. if the density is big, put the item into the cart

It is slow or imprecise. I end up putting every item over 1 g/cm^3 into the cart as the math is easy, but then I run out of such items

### Additional context

Keybinding for density could be `D` since `d` is taken up by damage. Unless it is not a good idea to have uppercase and lowercase keybinding. I would use `h` since "density" is "hustota" in my language.

Kamayana:
#### Summary
Interface "Add AIM density sort"

#### Purpose of change
Closes #63032.

#### Describe the solution
Adds a sorting option that calculates density with weight/volume (with Divide-By-Zero protection). I decided this was useful enough to deserve taking over the 'd' hotkey in the sort menu, so I renamed "damage" to "offensive power". "Damage" was fairly unclear anyway; I thought it meant sorting by how damaged items were and only found it meant sorting by DPS while writing this PR!

If anyone has a better name for damage sort than "offensive power" that still doesn't clash with an existing hotkey (unwvdxcasb) I'd like to hear it. Alternately, keep the damage hotkey and give density 'D' or 't', but as mentioned above I think "damage" should be renamed regardless.

#### Testing
Loaded my test save to take the below images and confirmed they look good. Also loaded my current proper save to density sort my massive shelves of spare parts and it worked lightning fast so no problems there.

### By weight
image
### By volume
image
### By density - very different!
image


Brambor:
Awesome! Works as expected. I currently play with your branch rebased on master.

Only thing is that if you switch between your inventory and a inventory of dragged vehicle, you cannot see the density of items in destination vs density of items in inventory. Either a column with density or rows like "density below 1", "density below 3/2" etc. would solve this problem. But I can calculate density from head as well for this one off.

However, it does work fantastically and it did solve my problem fully. Thanks!

PR #63056, opened 2023-01-09
Effect type modifiers rewrite

Farseer2:
#### Summary
Performance "Effect modifier rewrite"

#### Purpose of change
The current effect.cpp `mod_data` code is rather slow: upon testing it took around 7-8% of runtime during sleep on a random save file.
Most of that time was spent hashing the large tuple of strings.

This change both attempts to speed this up (modifier calculation is now negligible), make the code MUCH more readable, and de-duplicate code.

#### Describe the solution
Instead of a tuple with strings, we use 2 levels of map indirection, and replace the strings with enums for fast hashing and lower memory usage.

#### Testing
Effects testing suite

### DIFF
--- a/src/effect.cpp
+++ b/src/effect.cpp
@@ -42,6 +42,8 @@ static const itype_id itype_money_one( "money_one" );
 static const trait_id trait_LACTOSE( "LACTOSE" );
 static const trait_id trait_VEGETARIAN( "VEGETARIAN" );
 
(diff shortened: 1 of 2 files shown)


Maleclypse:
When you are ready for review it will be helpful to include any performance tracking you did to demonstrate results.

Farseer2:
It IS ready for review, but I currently do not have the setup to provide the performance numbers.
However, I would suggest merging it for the sake of code simplicity alone...

PR #63058, opened 2023-01-10
Remove five pin bow sight crafting recipe

Karol1223:
#### Summary
Balance "Five pin bow sights are uncraftable"

#### Purpose of change
The description of the five pin bow sights refer to an intricate construction with fiber optics. Ain't no way in hell you're making that from scrap metal and a chunk of plastic you picked up after kicking a trash can to death.

#### Describe the solution
Simply removes the recipe to craft it

I was initially pointed here because the materials didn't add up with the recipe ingredients. Then I read the description and it turned out the recipe made no sense at all. Numbers on here as well as material proportions could probably use modifying but I don't think I have the necessary knowledge and patience to do so.

### DIFF
--- a/data/json/recipes/weapon/mods.json
+++ b/data/json/recipes/weapon/mods.json
@@ -1096,22 +1096,6 @@
   {
     "type": "recipe",
     "activity_level": "LIGHT_EXERCISE",
-    "result": "bow_sight",
-    "category": "CC_WEAPON",
-    "subcategory": "CSC_WEAPON_MODS",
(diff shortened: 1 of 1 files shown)


NetSysFire (approved):
Dug a bit in the blame and found out the recipe was added in c69ec503664563516ffdfe4e42531a675db65f50, aka 2015.

Also +1 for "Ain't no way in hell you're making that from scrap metal and a chunk of plastic you picked up after kicking a trash can to death.", made me chuckle.

PR #57483, opened 2022-05-06
More npc locations revived

MNG-cataclysm:
#### Summary
Content "Add more NPC locations"

#### Purpose of change

The purpose of this change is to add more places where NPC's can spawn, either alone or in groups; providing a greater variety within the world of Cataclysm.

This adds more NPC's in various places. Currently, two random lone NPC's will spawn at evac shelters and homeless camps. This also adds a homeless camp with a group of survivors; so far including a chain of missions and an ability to trade. Additionally, this adds several new varieties of evac shelters.

Essentially, this is a revival of pull request #55196, with great thanks to onura46 for resurrecting it.

#### Describe the solution

This branch contains the same content as #55196, adding various NPCs to the world of Cataclysm. onura46 added cooldown timers for certain rare products sold by the chemist, fleshed out dialogue, and had the player spawn in the basement of the evac shelter for the “Infested Shelter” challenge start.

#### Describe alternatives you've considered
 None

#### Testing
According to onura46, all code runs smoothly and works as intended.

#### Additional context

The original content was inspired by two Discourse boards, seen below:

https://discourse.cataclysmdda.org/t/more-evac-shelter-varieties/27336/9

https://discourse.cataclysmdda.org/t/npc-s-what-do-you-want-changed-and-improved/27102/18

PR #63062, opened 2023-01-10
Fix an error relating to the chemist NPC

MNG-cataclysm:
#### Summary
Bugfixes "Fix a transactional error regarding the chemist NPC"

#### Purpose of change
Fix a bug before stable, so everything can work in a satisfactory fashion. When trading with the chemist NPC from #57483, specifically buying chemicals in bulk, the player could be locked out of purchasing something they wanted to buy if they could not afford it, as though they had purchased it, due to the way the dialogue effects were structured.  

#### Describe the solution
As an alternative to #62804, the effects applied when purchasing something have been moved to a true-false EOC, in the 'u_buy_item' effect, which will apply the cooldown timer properly when the player successfully purchases the merchandise in question.

None, besides what's going on in #62804.

#### Testing
Went to a chemist, did their jobs, asked to buy stuff, and repeatedly exited the trade menu for items I can't afford. They still appeared in the dialogue box as a selectable option (working as intended). When I got some valuables and bought the items, they were no longer available for purchase, and the cooldown timers for them had been set (also works as intended). 

#### Additional context
None.

### DIFF
--- a/data/json/npcs/cabin_chemist/chemist_npc.json
+++ b/data/json/npcs/cabin_chemist/chemist_npc.json
@@ -292,7 +292,14 @@
       },
       {
         "text": "[300$]  I'll buy some napalm.",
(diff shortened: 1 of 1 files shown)


MNG-cataclysm:
Whoops. Forgot to lint. Closing this.

PR #63064, opened 2023-01-10
Fix the chemist NPC

MNG-cataclysm:
#### Summary
Bugfixes "Fix a transactional error for the chemist NPC"

#### Purpose of change
The chemist would lock you out of buying something, as though you purchased it, if you exited the trade menu when making up costs. This will fix that.

#### Describe the solution
Move the dialogue effects to a true-false EOC, defined in the `u_buy_item` effect.

This is an alternative to #62804.

#### Testing
I went to a chemist, did some jobs, asked to buy supplies, and repeatedly exited the trade menu for things I couldn't afford. The option to purchase it remained available (as intended). I then bought it with valuables, and the option went away; the cooldown timer having been properly applied.

#### Additional context
I've done this with lumbermill NPC's too (#63084).


Maleclypse (dismissed):

MNG-cataclysm:
Updated to not cause problems like those in #63102.

MNG-cataclysm:
@Maleclypse Could you re-review this, please?

Issue #54009, opened 2022-01-03
UI: make the [?] keybindings shortcut discoverable in the item action menu


hexagonrecursion:
### Is your feature request related to a problem? Please describe.

I like that the [?] keybindings menu is advertised in many places in your UI. I have notices one place, where it is not advertised even though it should: the [%] item action menu.

### Solution you would like.

Add a label saying "[?] keybindings" at the bottom of the item action menu similarly to how it is done in many other places

matshou:
It is not advertised in many more places then just the action menu.
Here are examples of menus that do not have it:

- Inventory menu.
- Compare menu.
- Unload menu.
- Take off menu.
- Multidrop menu.

I think it is fair to say that it is much more common place to **not** find it advertised.

LeahLuong:
Funnily enough, this is 1 place where playing on mobile has the leg up. "?" automatically generates on the hotbar in pretty much every place it's supposed to. I can't think of 1 where it doesn't. The irony is that it's frequently unneeded on mobile _because_ of the hotbar.

Pointabc:
I started to do task #54009, how to indicate the task that it is in progress?

Zireael07:
Github issues can only be open or closed. The only thing that can be in progress is projects, which is a grouping of issues AFAIK and therefore doesn't apply

Pointabc:
> Github issues can only be open or closed. The only thing that can be in progress is projects, which is a grouping of issues AFAIK and therefore doesn't apply

This task is open, according to the messages above.

Pointabc:
@hexagonrecursion, can you check this issue? It was fixed and merged.

anothersimulacrum:
Confirmed as fixed.

PR #63066, opened 2023-01-10
#54009 add the [?] keybindings at the bottom of the item action menu

Pointabc:
#### Summary
Interface "Added [?] keybindings at the bottom of the item action menu"


src/item_action.cpp
```diff
@@ -370,6 +370,11 @@ void game::item_action_menu( item_location loc )
         num++;
     }
 
+    // custom footer
+    std::string footer = string_format(_("< [<color_yellow>%s</color>] keybindings >"),
+        ctxt.get_desc("HELP_KEYBINDINGS"));
+    kmenu.footer_text = footer.substr(1, footer.length() - 3);
```
Qrox, line 376: What's the purpose of substr here? If you don't need part of the string you can simply remove it from the original string.

Pointabc: Hi, if you pass a string without "< " and " >" ([<color_yellow>%s</color>] keybindings), then the string will not be translated. Angle brackets are not very suitable for the footer, so I removed them after the translation.
Clarified translation for 39810 segment "<{%s] Assigned Keys >". You need to change { to [ (For EN-RU - https://www.transifex.com/cataclysm-dda-translators/cataclysm-dda/viewstrings/#ru/0f-cataclysm-dda/316259458?q=text%3Akey).

Qrox: > the string will not be translated

It will, after the change is merged and the string is translated on Transifex.

I don't think it is a good idea to use the substring of an existing string's translation because there's no guarantee that the translated string will follow the same format as the original string. For example, the angle brackets might be replaced by characters with different number of utf-8 bytes or multiple characters in a different language, and taking the substring of the translated text will break it in that case.

Pointabc: Thank you for the feedback, I'll try to find a more universal way.

Qrox, line 376:
```suggestion
    kmenu.footer_text = string_format( _( "[<color_yellow>%s</color>] keybindings" ),
        ctxt.get_desc( "HELP_KEYBINDINGS" ) );
```
Just change the original string. It will be extracted as a new string and translated on Transifex. Also code style.

Pointabc: I tried the option earlier, the translation is not substituted (that's why I decided to use substr). I assume that this happens, because there is no source segment in the translation database without angular quotes. To make your version work, you can add the segment [ <color_yellow>%s</color>] assigned keys to the translation memory.

Pointabc: image

Qrox: The addition of new strings to the translation files is done automatically via scripts after a PR is merged. There's no need to worry about translation here as long as you marked all new strings for translation (which you already does in your code).

Pointabc: Qrox, can you tell me from the workflow what else I need to do? The code has been corrected, everything has been checked by AStyle, but checks are still needed.

Qrox: The remaining checks need to be approved by a maintainer. For now you don't have to do anything.

PR #63089, opened 2023-01-12
Prevent occlusions via (auto)-semi-transparent sprites

mlange-42:
#### Summary
Features "Prevent occlusions via (auto)-semi-transparent sprites (similar to retraction)"

#### Purpose of change

Adds the necessary code to support sprites that toggle between normal and semi-transparent as the player approaches, or when the player toggles the respective option (option already exists, for retracting walls etc.).

Besides being required for ISO tilsets, this also offers a solution for the problem of occlusion in forests.

#### Describe the solution

* The player proximity calculation used for retraction is re-used to determine the use of a potential semi-transparent sprite variant. Settings like on/off/auto and distances apply for both mechanisms now.
* Two switches were added to the options, which lets the player select none, one ot both of the two mechanisms.
* The naming of the related options and keybindings was made consistent.
* Fixes/prevents re-calculation of player proximity due to recursive function calls

Sprites are provided by using the suffix `_transparent` on any usual ID.

None.

#### Testing

Tested in-game with Ultica-ISO, using multitile furniture and seasonal trees, to check for compatibility with other tile variants. See screenshots.

#### Additional context

Screenshot of the resp. section in debug options:

grafik

Screenshot showing semi-transparent bookcases and tables, in cunjunction with retracting walls in Ultica-ISO:

grafik

Screenshot showing semi-transparent trees in Ultica-ISO:

grafik


mlange-42:
CI fails with unrelated error:

```
Could not connect to azure.archive.ubuntu.com:80 (52.252.75.106), connection timed out [IP: 52.252.75.106 80]
```

esotericist:
happy to see this, and i'm betting it'll get merged shortly after stable.

PR #63097, opened 2023-01-13
Limbify and rework dodge encumbrance scaling

Venera3:
#### Summary
Features "Rework dodge encumbrance scaling"

#### Purpose of change

Encumbrance's effect on dodge abilities was pretty inconsistent with the general score scaling, and was influenced by minimal amounts of encumbrance. Additionally, at high levels of dodge skill / dexterity armor encumbrance became a non-issue allowing you to dodge even high-skill enemies while wearing ornamental plate. 

#### Describe the solution

- Rework the encumbrance scaling from a subtractive penalty to a multiplier based on the balance, movespeed and footing scores - for a comparison of the old and new calculation see Additional Context
- Slightly modify the character modifier calculation so that modifiers calculated via multiplying different scores use the power of the scores when they are weighted
- [x]  Do a balance pass on monster melee skill to rein in some outliers. As a rule monster skill was universally outpaced by lategame dodge skill levels, which led to newer monsters getting higher skills to compensate (he said in the passive voice *cough cough*). Since this will rein in armored high-level characters some reduction in skills is probably in order. -- Monsters and special attacks in the 8+ melee skill bracket got a 1-2 point reduction pending more testing.

Use less scores, but logically your whole body needs to be free for effective dodging (reaction score comes into play separately as a flat modifier on dodge chances).
Rework the encumbrance score scaling formula but that's too mathy and I'm not entirely sure what I'd like to have in its stead - the current formula penalizes the first levels of encumbrance after the threshold very heavily but reaches a minimum at around 0,33x penalty before the limit, which feels wrong for half the scores and right for the other. A project for another day, maybe.

#### Testing

Tested an inordinate permutations of modifiers using the standard combat test mod's loadouts, results of the winning formula see Additional context.
Multiplicatively-calculated modifiers use the base score's power as intended.

#### Additional context

The base dodge rating of the character is `DEX/2 + dodge skill`, this remains unchanged.
Old calculation:
```
Limbs with "encumbrance_impacts_dodge" (torso and legs) reduce this by 1/10th of their encumbrance
Sample final dodge ratings( SCT mod loadouts and an extra test in full EOD gear, 0/4/8 dodge skill, 8 DEX):
  Day1 4,0 EOD Day1 0,0
  Mid 4,8 EOD Mid 0,0
  Late 6,8 EOD Late 0,0",
```
New calculation:
```
Multiplied by the limb_dodge_mod character modifier ( balance^1.5 x move_speed x footing )
Sample final ratings (same setup as above):
  Day1 4,0 EOD Day1 0,3
  Mid 5,0 EOD Mid 0,5
  Late 4,7 EOD Late 0,8
```


CMakePresets.json
```diff
@@ -14,41 +14,26 @@
       "condition": {
         "type": "allOf",
         "conditions": [
-          {
-            "type": "equals",
```
bombasticSlacks, line 18: you have some bonus files included

org.cataclysmdda.CataclysmDDA.json
```diff
@@ -16,22 +16,10 @@
       "name": "cataclysm-tiles",
       "buildsystem": "simple",
       "build-options": {
-        "env": {
```
bombasticSlacks, line 19: lotsa bonus files

PR #63100, opened 2023-01-13
Shotgun speedloader dimensions

vyrondrosos:
#### Summary
Balance "Change shotgun speedloader dimensions/weight and add belt clip property"

#### Purpose of change
The current volume and weight of the shotgun speedloaders is unrealistic. The item's description seems to match the "syringe speedloader" type, which is essentially a long tube holding shotgun shells. We can thus approximate the volume of a 8 shot speedloader with 8 x (volume of shot shell) which would give a volume of ~0.56 liters. The current volume of the item is 2 liters. 
Also, there is no length attribute for the items, which considering the long-and-thin shape of the object leads to the game calculating a wildly inaccurate default length.

#### Describe the solution
Changes the volume, weight and length of shot_speedloader6 and shot_speedloader8 to more realistic values. The volume of both items was cut by more than half, the weight was slightly reduced and a length attribute was added. Because the items are now 50cm and 65cm respectively and unable to fit in most quick-to-use pockets and because this is an item that could conceivably have a belt clip, the BELT_CLIP property is also added.

Instead of using the BELT_CLIP property (since not all speedloaders available on the market feature belt clips), alternatively wearables designed to hold the speedloaders could be added. [Something like this](https://www.kiesler.com/product/details/21ccfe55-8101-4e99-a2ae-871ee377c653/shotgun-speedloader-holder/) for example. In my opinion however, belt clips are the most elegant and realistic solution.

#### Testing
I added the aforementioned items using the debug menu. I confirmed that their dimensions and weight are correct. I inserted them in belts to confirm the BELT_CLIP property functions.

#### Additional context
When asking one vendor about the weight of their 6-shot speedloaders, the best quote I got is that it weighs "way less than a pound). Personally I think it should be slashed further considering its just a hollow plastic tube, but I was conservative with my estimate.


GuardianDll (approved):
My brain was too smooth to do it for myself. Thank you for you work

oosyrag:
Now if only shotgun speedloaders actually worked :3

PR #63084, opened 2023-01-12
Fix the lumbermill trader

MNG-cataclysm:
#### Summary
Bugfixes "Fix a transactional error with the lumbermill trader"

#### Purpose of change
Like the chemist trader (#63064), the lumbermill trader had a bug in which purchase options would become unavailable if the player could not pay for them, as though they had bought the merchandise.  

#### Describe the solution
Like the chemist, I moved the cooldown timer application to a true-false EOC, defined in the `u_buy_item` effect.

#### Describe alternatives you've considered
This is the alternative, mainly for #62804.

#### Testing
Went to a lumbermill, tried to buy some logs, and exited the trade menu when I couldn't afford it. The option remained available for purchase, and I successfully bought them later with some valuables. I repeated this with the other options that have cooldown timers.

#### Additional context
None.

PR #63102, opened 2023-01-13
Fix the check error originating from `TALK_lumbermill_fabricate.json`

MNG-cataclysm:
#### Summary
Bugfixes "Fix the check error coming from `TALK_lumbermill_fabricate.json`"

#### Purpose of change
As pointed out [here](https://github.com/CleverRaven/Cataclysm-DDA/pull/63084#issuecomment-1382033650), a check error has originated in this file due to changes made to the trade system. This should fix it.

#### Describe the solution
Added brackets around the `effect` entries in trade dialogue.

None.

#### Testing
This PR should test it. If the error doesn't appear in the tests, it will be fixed.

#### Additional context
None.

### DIFF
--- a/data/json/npcs/lumbermill_employees/TALK_lumbermill_fabricate.json
+++ b/data/json/npcs/lumbermill_employees/TALK_lumbermill_fabricate.json
@@ -166,6 +166,7 @@
   {
     "type": "effect_on_condition",
     "id": "EOC_BUY_LOGS_TRUE_EFFECT",
+    "global": false,
     "effect": [
       { "npc_add_effect": "lumbermill_waiting_for_log", "duration": "2 h" },
       { "npc_add_effect": "currently_busy", "duration": "2 h" }
@@ -174,6 +175,7 @@
   {
     "type": "effect_on_condition",
     "id": "EOC_BUY_PANEL_TRUE_EFFECT",
+    "global": false,
     "effect": [
       { "npc_add_effect": "lumbermill_waiting_for_wood_panel", "duration": "3 h" },
       { "npc_add_effect": "currently_busy", "duration": "3 h" }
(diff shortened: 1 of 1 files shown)


Maleclypse:
```> Extracting strings from JSON
Exception when parsing JSON data type "effect_on_condition"
Error in JSON file: 'data/json/npcs/lumbermill_employees/TALK_lumbermill_fabricate.json'
'effect'
Error in extracting strings from JSON. Aborting.
Error: Process completed with exit code 1.```

File: data/json/npcs/lumbermill_employees/TALK_lumbermill_fabricate.json [hunk 1]
```diff
@@ -177,15 +190,13 @@
...
     "effect": [
       { "npc_add_effect": "lumbermill_waiting_for_wood_beam", "duration": "4 h" },
       { "npc_add_effect": "currently_busy", "duration": "4 h" }
     ]
   },
   {
     "type": "effect_on_condition",
-    "id": "EOC_LUMBER_BUY_FALSE_EFFECT",
-    "global": false
+    "id": "EOC_LUMBER_BUY_FALSE_EFFECT"
```
dseguin, line 200:
```suggestion
    "id": "EOC_LUMBER_BUY_FALSE_EFFECT",
    "effect": []
```
The string extractor is expecting an "effect" field here, but it can just be blank.

dseguin:
You can undo the other changes if you want. The only fix here is to add the blank "effect" field.

Don't worry about the Text Changes Analyzer failure on the base commit, the merge commit is what matters here.

Screenshot from 2023-01-13 19-50-58

MNG-cataclysm:
> You can undo the other changes if you want. The only fix here is to add the blank "effect" field.
> 
> 
> 
> Don't worry about the Text Changes Analyzer failure on the base commit, the merge commit is what matters here.
> 
> 
> 
> Screenshot from 2023-01-13 19-50-58
> 
> 

Personally, I think the change should be left as-is, just to cover all the bases.

PR #62975, opened 2023-01-03
ursine faces

TiberianErin:
8(:0o)
actually it kind of looks like a clown face

*honk*

#### Summary
Interface "Ursine has its own faces now (no horizontal)"

#### Purpose of change

I need something to work on
And the ursine faces were copying the human faces
So I fixed that

#### Describe the solution
Copied the Feline values and made them ursine (im bored)

#### Describe alternatives you've considered
dealing with the lack of custom face

#### Testing
it should work, it just has new values
tested it to make sure (it works)

#### Additional context
8(:0))

PR #63109, opened 2023-01-14
Structured sidebar modifications

Kamayana:
#### Summary
Interface "Structured sidebar modifications"

#### Purpose of change
The structured sidebar is missing some information that'd be nice to have, specifically vehicle fuel/azimuth and nuclear badge status. More subjectively, some of the information could be laid out more intuitively. It's my favorite sidebar format but it has quite a learning curve.

#### Describe the solution
- Moves all of the weariness + stamina information to the right side of the health panel, and adds letter indicators of which sidebar is which, S for Stamina and W for weariness transition (context included for translators). I laid them out in an intuitive arrangement: Stamina bar > Activity intensity > weariness transition bar > weariness > weary malus. Every replaced element was moved elsewhere.
- Made a new row containing martial arts style, wield, movespeed, and carried weight, one with weight percentage and one with weight graph. When driving a vehicle, wield changes to fuel % and movespeed changes to vehicle speed.
- Moved Environment Temperature and Temp Change to the middle panel next to body temperature, where the old speed indicator was. A small nuclear badge indicator was added as well if one is worn.
- Moved mood next to focus, abbreviating Focus to 'F' to make room.

Enough has changed here that we could potentially list both this and the current Structured as alternatives of one another. I would not be opposed to that, but I went with a more straightforward replacement for the initial PR.

#### Testing
I made sure all elements work as expected. Any issues found are also found on the master branch - one is that the Self-Aware interface screws up element spacing, but that occurs among all sidebar layouts with SA health and fixing it is outside the scope of this PR.

#### Additional context
image
image
image


mlange-42:
The original author of the sidebar here...
Good to see the sidebar is used, and thanks for investing time in it's improvement!

Would like to give this an in-depth review the next days.

mlange-42:
> Added a row that shows the wielded weapon and one that contains both the weapon and fighting style. This is to make up for the right health panel not showing them anymore. Removed the wielded printout at the top of all the compasses, since the new row achieves the same purpose with more flexibility.

What I do not like 100% about this is the additional space used when using a larger overmap.
The reason for placing this row right above the compass (next to the overmap) was to make use of that space that would be empty with a an overmap larger than 5x5 (esp. 9x9).
Also, the horizonal space was definitely sufficient there.

Also not 100% convinced of the new places for speed, mood and safe mode. But will have to check it in-game.

mlange-42:
Regarding vehicle azimuth, is there any real use for it?

Kamayana:
> > Added a row that shows the wielded weapon and one that contains both the weapon and fighting style. This is to make up for the right health panel not showing them anymore. Removed the wielded printout at the top of all the compasses, since the new row achieves the same purpose with more flexibility.
> 
> What I do not like 100% about this is the additional space used when using a larger overmap. The reason for placing this row right above the compass (next to the overmap) was to make use of that space that would be empty with an overmap larger than 5x5 (esp. 9x9). Also, the horizonal space was definitely sufficient there.
> 
> Also not 100% convinced of the new places for speed, mood and safe mode. But will have to check it in-game.
> 
> EDIT: And BTW, who added those football heads? 🤔

I was reluctant to have multiple versions of the compass because there are multiple versions of each one so it would grow the element list quite a bit, but you are correct that that's empty space left on the table. Might be worth having both options.

I wanted to give speed enough room to grow to show both fuel percentage and triple-digit vehicle speed, and the old place for the weariness info seemed like the best choice there. Most of the other elements have a logical group they should be placed next to. Mood could go anywhere so I put that to the left of speed to make the text match up nicely with the body parts above. SAFE mode can also be placed anywhere, but it felt out of place if it was moved under the bodygraphs with mood and I wanted to keep it roughly in the same location it was before. Focus could also go anywhere so it's an option for swapping things around.

I was the one who added the football heads (#63037). It was hard to make out the mouth temperature when the mouth was just a period, I wanted to add more clarity. They admittedly look pretty goofy but it seemed like the option everyone found acceptable. We could always change it again if you dislike it.

> Regarding vehicle azimuth, is there any real use for it? I left it out deliberately.

I mostly added it because there was room for it now, and all the other layouts include it. I also find it unnecessary but I didn't realize it was left out deliberately, I was operating more on a "this should have everything the others do" mindset.

mlange-42:
Tried it in-game now. The re-arranged items at the top (speed, weariness, etc.) work ok. (Although I found the focus/speed/mood/save/stamina block more clear and structured before) (This:)

grafik

What I find a degradation is the stuff around MA, wielded and the compass.
I need 2 more rows now to see everything that was there before in the default setup.  
Namely, I need "Wielded" AND "MA Buffs", which were additionally moved out of the overmap/compass widget.
The sidebar was particularly designed for people playing on rel. low resolution or with a larget font size, so every row matters IMO. E.g. in my default setup (no MA buffs, 7x7 overmap, maximum possible minimap), I am already left with only 7 lines of log in the current version.

A particular problem IMO is that the new "Style/Wielded" widget does not use a separate sub-widget for ammo, so we get this when using it instead of the two mentioned above to save space:

grafik

So, no ammo there for weapons with long names. So wielded and ammo should be split here.

Unfortunately, the new "Style/Wielded" widget is probably too wide to put it back next to the overmap. And the space for the MA style in the health widget is gone. Therefore, maybe this could be a solution:
* Leave MA buffs as is for those who want to see the current buff (I personally don't care enough to sacrifice a line of log for it)
* Move "Wielded/mode/ammo" back next to the overmap (i.e. as it was before)
* Move the mood to the right of weariness, focus or temperature
* Add MA mode (no buffs) where the mood is now
* (Maybe?) Move the Weight widget where Speed is now, and move the two speed widgets one more to the right? Would alogn the speed and focus labels and help with structure (ok, honestly, I think the speed widget was better in its compact form, no added info by making it twice as wide)

What do you think?

grafik

mlange-42:
And a thought regarding the **learning curve** for sidebars...

I think the way to go here is not to add more labels to sidebars (that want to be super compact), but to help the player to understand sidebars from the game side.

I already startet this by adding the possibility to add a description to each sidebar (when scrolling through sidebars in the right panel). But information that can be provided there is limited of course. It could, however, at least be used to add a link to a documentation page for the sidebar.

Another idea that was already discussed was to add tooltips for widgets. With the challenge that they should also work for people that don't want to touch the mouse. Digged into that a bit already, but seemingly it was over my abilities.

Kamayana:
Thanks for the feedback, I changed this to a draft for when I get the time to work on it some more.

One thing I'll mention is that the mood should have at least 5 spaces to accommodate wider mutant faces like #62975. Maybe it could move all the way to the right, since changing to compact speed would give enough room for it. Here's a mockup of what that'd look like:

image

Or we could swap focus and mood:

image

Or we could change focus to "F 100" to give space for mood, which actually gets us a new space in the bottom right for something else (though we still need space to accommodate vehicle fuel).

image

Move weight back to the right and we could even bring back the wielded display:

image

mlange-42:
I would prefer the last version. Works much better for me this way.

Kamayana:
image

Here are two options; Option A squeezes the fuel percentage to the left of weight, Option B replaces wielded with the fuel percentage. I'm inclined towards B, especially since the wielded weapon will always be visible in the compass.

mlange-42:
> Here are two options; Option A squeezes the fuel percentage to the left of weight, Option B replaces wielded with the fuel percentage. I'm inclined towards B, especially since the wielded weapon will always be visible in the compass.

:+1: Would definitely also prefer B!

Kamayana:
All the feedback has been applied, and I updated the PR's description accordingly. I also added side view alternatives for the tiny bodygraphs.

bombasticSlacks:
As I understand it @mlange-42 you are the original author of the sidebar so if you are happy with the changes I'm going to merge this!

PR #57902, opened 2022-05-24
Desert Region Revival

FarkasKBence:
#### Summary
Mods "Desert region revival"

#### Purpose of change

Part 1/??? of the desert region revival project.

#### Describe the solution

The way I see it, there are three main reasons why the desert region mod doesn't garner enough interest: it doesn't have much to offer content-wise, it is unplayable due to the scorching heat, and it was obsoleted - simply put, it's an abandoned proof of concept mod that wasn't intended for a balanced playthrough.  This PR doesn't solve any of the mentioned issues yet ( this PR removes more content than it adds ), but it takes the mod one step closer to being enabled again.

This PR updates regional map settings, blacklists a few locations, organizes the mod with folders, replaces swamps with desert pavement, and adds mesas, canyons (that are a bit too circular), and dust storms ( the code of which I borrowed from portal storms ). ~~I've removed the animals that aren't native to North American deserts, like kangaroos and bush elephants~~, and added sandstone, which can be found in levels -1 and -2.

#### Describe alternatives you've considered

I'm still debating which North American desert this mod should be set in.  So far, I've got two picks: the Sonoran Desert, and the Mojave Desert.  AFAIK, the Mojave has cold winters, almost no precipitation, and its plant life isn't too varied. The Sonoran is hilly, with lots of rifts and faults, especially in the Colorado Desert. Its temperatures are warm year-round, and rainfall is infrequent and irregular. From the images I've seen, it seems to me that the Sonoran Desert is more lively, and the Mojave is just a wasteland with Las Vegas in the middle.  Content-wise, the Sonoran Desert seems to have a bit more to offer, so that's what I'm leaning towards for now.

#### Testing

Generated a new world with `desertpack` loaded.  Left the evac shelter, groundcover looked desert-ish, visited a nearby mesa and a canyon, it generated as intended.  Then teleported to about 100 overmaps, no error messages popped up.

Then I tested the new dust storm weather type.  I debug-spawned the Test Dust Storm Causer, activated it, and as expected, a dust storm appeared.  I wasn't wearing any environmental protection, so the dust gave me negative effects.  Then I waited a few days, and a dust storm arrived naturally.  It didn't seem to conflict with portal storms.

#### Additional context

<details>
  <summary>Screenshots:</summary>

The mesas and a canyon as seen from the map.
kép

A mesa and a canyon meeting:
kép

Dust storm warning:
kép

Stepping outside without environmental protection:
kép
kép

</details>

Things to be done:
[...]

PR #63124, opened 2023-01-15
Desert Region: world generation changes

FarkasKBence:
#### Summary
Mods "Desert Region: world generation changes"

#### Purpose of change

Continues desert region revival project.

#### Describe the solution

Changes are focused on improving world generation:

* Re-added rivers, made lakes and rivers completely dry
  * added mudcracks to indicate where water once was
  * ~~the implementation of this was somewhat primitive; right now every river and lake OM terrain uses a 24x24 chunk of mudcrack as its map, which looks... bad.~~
  * replaced hardcoded river banks and lake shores with nested chunks
* Renamed sparse forests to 'bushland', dense forests to 'woodland'
* Re-added forest trails with configured settings
* Ported newer city buildings from vanilla
  * except the textile mills because they're based on buildings from NE
* Improved mesa and canyon mutable generation
  * fixed issue mentioned in #57902, now they don't rely on automatic roofs
  * much smoother and natural-looking walls
  * linear canyons; still not perfect, but much better than the old, circular ones

~~I've tried using conditional nested chunks to fix shorelines, but that got way too convoluted thanks to all the variables.
My next idea was to create a map extra - similar to `mx_reed` - that converts water into mudcrack, and apply it to every river bank and lake shore. That requires C++ however.~~

#### Testing
Loaded a world with Desert Region.

#### Additional context

<details>

  <summary>Screenshots:</summary>
Forest trails, shrubland, woodland, dry rivers, improved canyon generation:

kép

Smoother mesa walls:

kép

Lake shore:

kép

</details>


data/mods/desert_region/mapgen/waterbody.json
```diff
@@ -0,0 +1,44 @@
+[
+  {
+    "type": "mapgen",
+    "method": "json",
+    "om_terrain": [
+      "lake_bed",
+      "lake_shore",
+      "river_center",
```
NetSysFire, line 8: I believe you could make the dried out rivers prettier by adding a nested mapgen of a blotch of clay or recesses, with a low chance of some puddles in the middle, only spawning in the rivers center. Same for lakes perhaps. Then making river or lake shores mostly mudcracks, sand and dirt with maybe a bit more than usual vegetation.

John-Candlebury:
Take note that Ravines and Rivers/lakes dont interact nicely with each other (since ravines are designed principally for the Aftershock Exoplanet which has no liquid water). 

You'd either need to expand the generation code so they merge properly, or disable one or the other.

data/mods/desert_region/mapgen/waterbody.json
LovamKicsiGazsii: I think the simplistic river centers look *okay* for now, I am more concerned about the overly rectangular shorelines. Turning the shore into a mixture of sand, dirt, and mudcrack would not fix the harsh seams between either the center and the shore or the shore and the land, which is my main issue.

LovamKicsiGazsii: It's a problem for another day then.  Thank you for your suggestion, I'll consider it.

LovamKicsiGazsii: Nevermind, just realized else_chunks exist.  I'll make a commit here once I'm done.

PR #55268, opened 2022-02-10
Change the workbench to drop its item form upon deconstruction

KittyTac:
#### Summary
Balance "Workbench now simple-deconstructable, drops its item form upon deconstruction"

#### Purpose of change
There was no way to pick up a workbench after it was placed, and I don't think you would ever need to deconstruct it into its raw materials and waste so much time (besides I think it can be disassembled anyways), while you could find a workbench "in the wild" and want to take it home.

#### Describe the solution
Make the workbench be able to be deconstructed without tools (10 minutes) and instead of the materials used to craft it it will drop itself in its item form which will then be able to be placed.

#### Describe alternatives you've considered
Adding a separate "Pick Up Workbench" construction but that clutters the menu and is inconsistent with the other furniture that works like this.

#### Testing
image
image

#### Additional context
N/A

PR #63128, opened 2023-01-16
Remove workbench construction vs crafting exploit

cake-pie:
#### Summary
Bugfixes "Remove workbench construction vs crafting exploit"

#### Purpose of change
If you wanted a make a `workbench` to put in your vehicle, you could use the item crafting recipe, which requires various proficiencies, takes 2 hours, and consumes 8 `pipe`, 2 `sheet_metal`, 4 `sheet_metal_small` and 4 `pipe_fittings`. This is a reversible recipe.

Or, you can construct the furniture `f_workbench`, which requires no proficiencies, takes just 1 hour, and consumes 8 `pipe`, 2 `sheet_metal`, 4 `sheet_metal_small` but does not need `pipe_fittings`. This has `EASY_DECONSTRUCT` flag and can be trivially taken down without tools using "Deconstruct Simple Furniture" to yield the item. Disassembling this item would yield extra 4 `pipe_fittings` in addition to the resources originally put into constructing it, due to the reversible item craft recipe.

#### Describe the solution
Delete the furniture construction "Build workbench" entirely. It will no longer be possible to directly construct the furniture `f_workbench`. Instead, one must first craft the item `workbench`, and then use "Place workbench" to install the item as furniture `f_workbench`.

Bring furniture construction requirements in line with item crafting recipe. However, that still would not include proficiencies.

#### Testing
Check that the option is gone from construction menu.

### DIFF
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -2222,19 +2222,6 @@
   },
   {
     "type": "construction",
-    "id": "constr_workbench",
-    "group": "build_workbench",
-    "category": "FURN",
-    "required_skills": [ [ "fabrication", 3 ] ],
-    "time": "60 m",
-    "using": [ [ "welding_standard", 5 ] ],
-    "components": [ [ [ "pipe", 8 ] ], [ [ "sheet_metal", 2 ] ], [ [ "sheet_metal_small", 4 ] ] ],
-    "pre_note": "Can be deconstructed without tools.",
-    "pre_special": "check_empty",
-    "post_terrain": "f_workbench"
-  },
-  {
-    "type": "construction",
     "id": "constr_place_workbench",
     "group": "place_workbench",
     "category": "FURN",
--- a/data/json/construction_group.json
+++ b/data/json/construction_group.json
@@ -886,11 +886,6 @@
   },
   {
     "type": "construction_group",
-    "id": "build_workbench",
-    "name": "Build Workbench"
-  },
-  {
-    "type": "construction_group",
     "id": "build_woven_wattle_fence",
     "name": "Build Woven Wattle Fence"
   },


Maleclypse:
How does this change basecamp construction if at all? I believe some base camp workshop or other uses that construction likely? It’s a good catch Im  curious if it’s a complete fix yet.

cake-pie:
This does not impact basecamp construction at all, there are no references to the exploitable construction anywhere else in the codebase.

```bash
redacted/Cataclysm-DDA/data/json$ grep -rn '"build_workbench"'
construction_group.json:889:    "id": "build_workbench",
construction.json:2226:    "group": "build_workbench",

redacted/Cataclysm-DDA/data/json$ grep -rn '"constr_workbench"'
construction.json:2225:    "id": "constr_workbench",
```

The furniture `f_workbench` does show up in basecamp mapgens, but those are built via upgrade mission rather than furniture construction. That process requires/consumes materials defined in upgrade mission recipes; I have not reviewed those for requirements mismatch vs the item craft, or other potential exploits.

Maleclypse:
Workbench requirements
I continue to have concerns given that the recipe used by the basecamp is the recipe you are removing.  Can you test building a basecamp workshop to the upgrade anvil and workbench stage and make sure it works or at least pulls the requirements using pipe fittings now?

Maleclypse:
I do prefer the recipe with pipe fittings to be used by the basecamp but I want to at least make sure that basecamp recipes don't break on this in some way our tests miss.

cake-pie:
Basecamp appears to implicitly use whatever construction it can find that will yield the corresponding furniture. Currently it is pulling requirements from "Build workbench" simply because that is the first entry listed in `construction.json` that results in `f_workbench`. With "Build workbench" removed, it will use "Place workbench" instead, which requires that the items first be crafted via the item crafting recipe (which will thus require the pipe fittings).

Before: 3× `constr_workbench` requirements
before

After: 3× `constr_place_workbench` requirements
after

Maleclypse (approved):

Maleclypse:
Thanks for doing this!

cake-pie:
> _not_ being able to build a workbench (say, in your garage like any other normal person) feels ... awkward.

Not sure what your complaint is; you still can:

1. bring all the necessary materials to the spot where you want the workbench
2. use the crafting menu (default: `&`) to craft the workbench item right there 
3. use the construction menu (default: `*`) to place the workbench on the desired spot

The only difference is that there are now separate steps where you first have to turn the raw materials into a "flatpacked" version of the furniture, and then subsequently assemble/deploy it.

> Perhaps an updated Construction menu recipe would have sufficed

No, it would not. As mentioned in "alternatives considered" in OP, even if we adding pipe fittings to the "Build workbench" construction and changed the time taken to 2h to be in line with the item craft recipe, it is not possible to specify proficiencies required for a construction. This would still constitute an exploit with which to bypass the item craft recipe proficiencies requirements.

> The constructed version could even be non-destructible

This would require built and placed workbenches to be defined as separate furniture defs.  
It would also defeat the whole point of #55268.

PR #63147, opened 2023-01-17
Fix vehicle::air_drag_coeff()

### LINKED ISSUE #63101: vehicle::coeff_air_drag() doesn't work as intended
### Describe the bug

There are mainly two point of this issue.
1. "being twice as long as wide mildly reduces air drag" isn't true in code.
```
        c_air_drag_c -= ( 2 * ( mount_max.x - mount_min.x ) > width ) ? c_air_mod : 0;
```
This means, not being twice as wide as long mildly reduces air drag.
2. "trunk doors and halfboards at the tail mildly reduce air drag" will hardly come true.
```
        c_air_drag_c -= ( dc.last == mount_min.x ) ? c_air_mod : 0;
```
Where dc.last is an integer not smaller than 0 (if there's "LOW_FINAL_AIR_DRAG" or "HALF_BOARD" in this column) or -122 (if there's not any). And mount_min.x is the position of the rear of the vehicle relative to the default mount_point ( the default highlighted tile when you "e"xamine the vehicle ). So this "reduce" will take effect only if you build the car from scrap and install new parts only in the front of the vehicle.

### Steps to reproduce

I don't think in-game testing is needed to prove the first part to be wrong.

I'll illustrate the second part.

1. Spawn a SUV, which obeys "trunk doors and halfboards at the tail" principle.
2. Add extra frames to the rear of the vehicle, which breaks the above principle.
3. Find out that the air drag coefficient won't change.

### Expected behavior

For the first part, I think the fix code could be like this:
```
         c_air_drag_c -= ( mount_max.x - mount_min.x + 1 >= 2 * width ) ? c_air_mod : 0;
```
For the second part, I think the fix code could be like this:
```
        c_air_drag_c -= ( dc.last == 0 ) ? c_air_mod : 0;
```
The reason I choose 0 is that mount_min.x is an integer for the whole vehicle, which doesn't change with different column. And the smallest valid number for a member variable of dc is 0, which means the part is at the rear of the whole vehicle. 

### Screenshots

_No response_

- OS: Windows
- Game Version: 8bf4ccc [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Cataclysm++ [Cata++]
]

### Additional context

_No response_

SurFlurer:
#### Summary
Bugfixes "Make vehicle::coeff_air_drag() work as intended, in aspect of vehicle length and low-air-drag vehicle tail parts."

#### Purpose of change
Fixes #63101 

#### Describe the solution
There are two parts of the solution.

1. Make actually being twice as long as wide mildly reduce air drag, not "not being twice as wide as long mildly reduces air drag".
2. Compare the last part of each column with the last part of that column that is "LOW_FINAL_AIR_DRAG" or "HALF_BOARD". If true, reduce air drag.

1. Leave it as is.
2. For part 1, I think it's actually too harsh to strictly follow the "being twice as long as wide" rule in game, as veh_parts that are peripheral (doors, e.g.) in real life aren't as wide as a whole tile. Maybe we should only compare the inner part, aka (length minus 2) versus 2*(width minus 2) ? For part 2, I think the original code's meaning was to check if the last part of **the whole vehicle** is a low-air-drag one. But I changed that to only check if the last part of **the very column** is a low-air-drag one. This change is to help helicopters, because they usually have wings, although they are rare.

#### Testing
It compiles and generates test data. In-game testing is also good.

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -6093,6 +6093,7 @@ void vehicle::refresh( const bool remove_fakes )
     check_environmental_effects = true;
     insides_dirty = true;
     zones_dirty = true;
+    coeff_air_dirty = true;
     invalidate_mass();
     occupied_cache_pos = { -1, -1, -1 };
     refresh_active_item_cache();


SurFlurer:
Well, what should I do when there are many https 403 errors in checks' logs? To wait for a while, then add a commit to make checks running again?

BrettDong:
You can fix AStyle code formatting errors first.

SurFlurer:
@dseguin @irwiss Please review, thank you!

irwiss (approved):
This seems to be correct

I'd also consider moving `coeff_air_dirty = true;` from vehicle::enable_refresh to vehicle.cpp around line 6095 as the comment at https://github.com/CleverRaven/Cataclysm-DDA/issues/63101#issuecomment-1383071952 exposes stale cache value

SurFlurer:
The reason to the failure seems to be an angry random number god. lol

PR #63148, opened 2023-01-17
Turn off Android native ui by default

Zero3496:
#### Summary
None

#### Purpose of change

The experience of "Android native ui" may not be so good, and I have observed that many players will find ways to close it again. I suggest that it be turned off by default

#### Describe the solution

Change the default values of "SplashScreen.java" and "options.cpp".


Rimplestilt:
I would fully agree with this.

I am still disappointed that the fullscreen settings cannot be changed after installing the game via options, but i suppose there is some reason for that limitation.

Zero3496:
Is it so hard to admit that your design has failed? I really appreciate BrettD's work on Android  , but this ui is a very bad design. BrettD could go straight back to the online community he left and ask if this was a good design

ZhilkinSerg:
No

PR #63149, opened 2023-01-17
map: don't cast vision through removed cameras

andrei8l:
#### Summary
None

#### Purpose of change
Vision is cast through vehicle cameras that are about to be removed
Installing new parts in a vehicle leaves fake parts in removal limbo

#### Describe the solution
1. don't cast vision through cameras that are marked for removal
2. clean up fake parts when installing new parts instead of leaving them in limbo (refresh adds them back right after)

N/A

#### Testing
New and existing tests still pass

Using save file provided by Stoopedhooy on discord:  [Franktown.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10436831/Franktown.zip)
Remove and reinstall the inboard mirror on your tile. Your vision outside the vehicle should not change. Verify that the camera system still works.

#### Additional context
~~Ideally, this needs a test unit. I will follow up with one when I have time.~~ done

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -1505,7 +1505,7 @@ int vehicle::install_part( const point &dp, const vehicle_part &new_part )
         }
     }
     // refresh will add them back if needed
-    remove_fake_parts( false );
+    remove_fake_parts( true );
     parts.push_back( new_part );
     vehicle_part &pt = parts.back();
     int new_part_index = parts.size() - 1;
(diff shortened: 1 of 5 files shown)


andrei8l:
I've fixed the real issue (see updated OP), left the original as a backup, and added a test unit that covers #59714 too.

src/vehicle.cpp
```diff
@@ -1505,7 +1505,7 @@ int vehicle::install_part( const point &dp, const vehicle_part &new_part )
         }
     }
     // refresh will add them back if needed
-    remove_fake_parts( false );
+    remove_fake_parts( true );
```
kevingranade, line 1508: I'm only passing on this right now because adjusting these remove parts calls when debugging the game parts feature caused no end of errors and i don't want to risk breaking things worse right before a release.

bombasticSlacks: @kevingranade are we good to merge this now then and see if there are cascading effects?

kevingranade: Yep,  just haven't gotten around to it yet.

PR #63150, opened 2023-01-17
Make tool belt a lot lighter and a bit smaller

Olaroll:
#### Summary
Balance "Rebalanced tool belt weight and volume"

#### Purpose of change
I noticed that an empty tool belt currently weighs 3.85kg, which is way too much imo, especially for the relatively small amount of storage it gives.

#### Describe the solution
To find the right weight, I tried looking for an IRL tool belt that has a similar amount of storage when compared to the one in-game. Overall, the lightest tool belts I found were around 0.5kg, while the heaviest were about 2.5kg.

Just for convenience, here are the in-game tool belt's pockets:
- 1x 1L Pocket
- 3x 1L Sheath/Belt Loop
- 2x 1.5L Sheath/Belt Loop

Looking at the pockets, you can see that the in-game tool belt mostly consists of belt loops for the tools, and would thus be quite light.

The tool belt I found that had the most similar storage specs ended up being [this one](https://akribisleather.com/product/es/). It's made out of thick leather, and weighs a total of 0.85kg. It doesn't include the belt, so I rounded it up to 1kg to compensate.

I also lowered the volume slightly. I just set it to what feels right, as volume isn't that important for clothing anyway. I did make sure its volume is a bit larger than the police belt, as it seems like a good reference point.

Leaving the tool belt unreasonably heavy.

#### Testing
~~None, because the JSON change is as simple as can be.~~

After changing the crafting recipe in a later commit, I also tried loading up the game and making sure crafting and disassembly works.

#### Additional context
None


andrei8l:
The default/disassembly components weigh 1.7kg so the recipe needs to be updated too

Screenshot from 2023-01-17 19-53-41

Olaroll:
Alright, I've adjusted the recipe in the following ways:
- Leather pouches needed 6 -> 2
- Leather patches needed 6 -> 12
- Clasps needed 2 -> 5
- Leather belts needed 0 -> 1

These are mostly arbitrary amounts, but they add up to 0.95kg. Is that alright?

The only non-arbitrary change was setting the leather pouches to 2. The reasoning is that the tool belt has 1L of non belt loop storage, and each leather pouch contributes 0.5L.

Knut-Aage-Hofseth (approved):
I think this is a good sanity fix on this item and recipe.

PR #63151, opened 2023-01-17
Add fentanyl

MNG-cataclysm:
#### Summary
Content "Add fentanyl"

#### Purpose of change
Add fentanyl, a highly addictive painkiller and anesthetic. 

#### Describe the solution
Used the JSON from this [reddit post](https://www.reddit.com/r/cataclysmdda/comments/yjzldj/is_fentanyl_in_the_game/?utm_source=share&utm_medium=ios_app&utm_name=iossmf) for fentanyl, and tweaked it so the player won't instantly overdose from consumption. Fentanyl has also been added to some item groups for painkillers and narcotics, where it should reasonably appear.

None.

#### Testing
Consumed fentanyl to verify it worked and went exploring to find it naturally. Everything works as intended.

#### Additional context
I would like to place the painkiller effects as close as possible to the overdose threshold that would result in death, to help make the effects of fentanyl more realistic. If possible, I have also considered adding different grades of fentanyl which assign a random potency for the painkiller effects, making it possible to overdose with a single tablet.
~~Unlike the suggestions in the reddit post, fentanyl cannot be used to kill zombified law enforcement.~~


Venera3:
The effect of fentanyl has the same high ceiling as morphine, the only difference being the pharmacokinetics. "Boo hoo it's 100x as potent" just means you use 1/100th of the dose both for legal as for illicit usage. In actual effect there is no real difference, unless we want to differentiate `painkiller` into separate effects to allow for minor PK differences.

MNG-cataclysm:
> The effect of fentanyl has the same high ceiling as morphine, the only difference being the pharmacokinetics. "Boo hoo it's 100x as potent" just means you use 1/100th of the dose both for legal as for illicit usage. In actual effect there is no real difference, unless we want to differentiate `painkiller` into separate effects to allow for minor PK differences.

I suppose I'd set the drug effects to be the same as morphine, with an increase in charges to represent the 1/100th dose?

Venera3:
No, differentiating between iv opiates has no real point.
Let's use the nice legal versions as an example: morphine comes in ~5-10 mg/ml ampoules (variable volume, but the concentration is usually consistent), fentanyl in 0,05-0,1 mg/ml concentration - one ml of each will effect you *roughly* the same, in the level of abstraction we're working at. Illegal sources will have more variable concentrations, but I'd expect most of the market to be fed by medicinal sources that fell off the back of a van instead of de novo synthesis (and distributing a cut that will kill all your customers is not good business).

Profit0004:
As almost all fentanyl used in america is illegally used, and the world of CDDA seems to be an alternate reality, maybe we can not have fentanyl in the game?  Already  there is a lot of things different from the current USA so maybe China (yes, I understand the irony) does not supply massive amounts of cheap opiod to south american countries and have the same near non existent border between them and the USA

kevingranade:
If you want an accurate depiction of street fentanyl, give it wildly variable strength per dose.
Even better would be a meta substance like, "opoids, from the packaging you can tell that much,  but you can't tell which one it is".

MNG-cataclysm:
> If you want an accurate depiction of street fentanyl, give it wildly variable strength per dose.
> Even better would be a meta substance like, "opoids, from the packaging you can tell that much, but you can't tell which one it is".

I was thinking about making this pharmaceutical fentanyl, which was either stolen or sold from the store and redistributed.

kevingranade:
Make it a variant if morphene instead of a new item then.

MNG-cataclysm:
@kevingranade 
Why close this?

Issue #58725, opened 2022-06-25
Spawned items somtetimes get processsed at wrong z-coordinates on start of the game


Hirmuolio:
### Describe the bug

Spawned items somtetimes get processsed at wrong z-coordinates.

### Steps to reproduce

1. Start new game in evac shelter.
2. Look at the protein bars on the tables of evac shelter.
3. See that they are not all at same temperature.

1. Add breakpoint to this line in code: https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/item.cpp#L12350. This part of the code runs only when an item that is located underground is processed.
4. Start new game in evac shelter.
5. Observe as protein bars get processed on all z-coordinates ranging from -10 to 0.

This will cause some of the protein bars to have wrong temperatures (underground temperature=43 F).

This wrong coordinate is only present on the first turn. After that the items get processed at correct coordinate.

### Expected behavior

Protein bars that were spawned on ground level should have their z-coordinate as 0.

### Screenshots

Items on table of evac shelter few turns after start of new game. They should all be at same temperature but since some of them were processed at wrong z-coordinate they are not all same.
![evac](https://user-images.githubusercontent.com/22011552/175793963-cce61147-473f-4f3e-a7c5-5d9ca4220338.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19043.1766 (21H1)
- Game Version: 5ea8d0c [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

andrei8l:
/confirmed this happens for any `tinymap::load()`
Possible patch:
```diff
diff --git a/src/map.cpp b/src/map.cpp
index 108f612236..1eeabc6536 100644
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -6903,10 +6903,12 @@ void map::load( const tripoint_abs_sm &w, const bool update_vehicle,
 
     // actualize after loading all submaps to prevent errors
     // with entities at the edges
+    int const zmax = zlevels ? OVERMAP_HEIGHT : w.z();
+    int const zmin = zlevels ? -OVERMAP_DEPTH : w.z();
     for( int gridx = 0; gridx < my_MAPSIZE; gridx++ ) {
[...]

PR #63154, opened 2023-01-17
Map optimizations

Farseer2:
#### Summary
Performance "Improve map code"

#### Purpose of change
Generally improve map's code

#### Describe the solution
1) Devirtualize `in_bounds`, as it is called _a lot_ and is duplicated in tiny_map. Also, move base functions to header file for inlining.
2) Mark outside_cache as non-dirty on negative z-levels as well to prevent memset loops.

#### Testing
Test suite

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -8454,17 +8454,6 @@ bool map::inbounds( const tripoint_abs_ms &p ) const
     return inbounds( getlocal( p ) );
 }
 
-bool map::inbounds( const tripoint &p ) const
-{
-    static constexpr tripoint map_boundary_min( 0, 0, -OVERMAP_DEPTH );
-    static constexpr tripoint map_boundary_max( MAPSIZE_Y, MAPSIZE_X, OVERMAP_HEIGHT + 1 );
-
-    static constexpr half_open_cuboid<tripoint> map_boundaries(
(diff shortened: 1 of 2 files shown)


andrei8l:
> Devirtualize in_bounds, as it is called a lot and is duplicated in tiny_map. Also, move base functions to header file for inlining.

Isn't this the compiler's job? Also `tinymap::inbounds()` is arguably wrong in both variants since tinymap doesn't have z-levels (see #58725 for example)

Farseer2:
MSVC is notoriously bad (In my experience) in devirtualizing, especially in cases like this where the difference is a const.
I'd also argue that this is more readable :)
Another options would be templating the sizes, if you want to keep it consts, but making map templated would require tons of changes around the code...

As for the bug - do you want me to change it to check using `zlevels`? Might cause some unexpected bugs

ehughsbaird:
Do you have any evidence this actually improves performance?

Farseer2:
Kind of hard to prove - `inbounds` is inlined and checking overall performance is very inconsistent.
Logically, though, de-virtualizing, allowing inlining and correctly marking a dirty bit should far surpass the maximum of 2 extra multiplications (`* my map size`) that can no longer be pre-calculated.

PR #63157, opened 2023-01-17
Add canine bite suits

MNG-cataclysm:
#### Summary
Content "Add canine bite suits"

#### Purpose of change
Add canine bite suits, a utility item used in dog training. Animal pounds and shelters seemed to be lacking this or something similar for their work.

#### Describe the solution
The bite suit is made of Kevlar, highly cut resistant. They will also spawn in animal shelters and pounds.

None.

#### Testing
Spawned a bite suit and tested it out on a zombie dog. It couldn't hit any limbs which the suit covered.

#### Additional context
None.


Drew4484:
You can remove the 'valid_mods' entry as clothing modification has been removed. Might be a nice extension to add just the sleeves and legs as options.

MNG-cataclysm:
> You can remove the 'valid_mods' entry as clothing modification has been removed. Might be a nice extension to add just the sleeves and legs as options.

Huh. Why was clothing modification removed?

Drew4484:
> > You can remove the 'valid_mods' entry as clothing modification has been removed. Might be a nice extension to add just the sleeves and legs as options.
> 
> Huh. Why was clothing modification removed?

Short version is that it wasn't realistic (How was the kevlar being attached to the plate mail? How did one line chainmail with fur?), could make armor more likely to be damaged (adding additional materials divided the protection values of the existing materials) and wasn't compatible with our new detailed armor layers. The alternative is what we have now, a variety of armor recipes for things like the jumpsuits. We're focusing on adding more armor options and versatility, so you should see a lot more options in the recipe lists.

Rllaman:
Shouldn't this be on the "outer" layer? From googling canine bite suits, it looks like it might be possible to wear normal clothing underneath it, but it would definitely be hard to wear with something like an overcoat or any type of armor.

PR #63165, opened 2023-01-18
During the Android CDDA installation process, the "config" folder is allowed to be installed

Zero3496:
#### Summary

None

#### Purpose of change

Allow config folder to be installed during Android installation. This will not have a disruptive impact on the game itself.  This can increase the customization of players for making "integration package" or "lazy man package".  Players can add config to apk to preset values, and then use other tools to sign.  After that, they can spread these versions normally. This can make them more distinctive.

#### Describe the solution

Add a new installable resource folder in "SplashScreen. java".

#### Testing

After testing, the contents in the config folder can be installed normally. The game is also set as expected.

### DIFF
--- a/android/app/src/main/java/com/cleverraven/cataclysmdda/SplashScreen.java
+++ b/android/app/src/main/java/com/cleverraven/cataclysmdda/SplashScreen.java
@@ -304,8 +304,7 @@ public class SplashScreen extends Activity {
                 deleteRecursive(assetManager, externalFilesDir, new File(externalFilesDir + "/data"));
                 deleteRecursive(assetManager, externalFilesDir, new File(externalFilesDir + "/gfx"));
                 deleteRecursive(assetManager, externalFilesDir, new File(externalFilesDir + "/lang"));
-                deleteRecursive(assetManager, externalFilesDir, new File(externalFilesDir + "/config"));
-                
+        
                 // Install the new data over the top
                 copyAssetFolder(assetManager, "data", externalFilesDir + "/data");
                 copyAssetFolder(assetManager, "gfx", externalFilesDir + "/gfx");


ZhilkinSerg:
People who will assemble custom APK can add this patch themselves.

android/app/src/main/java/com/cleverraven/cataclysmdda/SplashScreen.java
```diff
@@ -303,11 +304,13 @@ protected Boolean doInBackground(Void... params) {
                 deleteRecursive(assetManager, externalFilesDir, new File(externalFilesDir + "/data"));
                 deleteRecursive(assetManager, externalFilesDir, new File(externalFilesDir + "/gfx"));
                 deleteRecursive(assetManager, externalFilesDir, new File(externalFilesDir + "/lang"));
-
+                deleteRecursive(assetManager, externalFilesDir, new File(externalFilesDir + "/config"));
```
ZhilkinSerg, line 307: > This will not have a disruptive impact on the game itself.

I am pretty sure it does have disruptive impact - existing `config` folder will be unconditionally deleted during upgrade from a different version and it won't be replaced with anything as we do not bundle `config` folder by default.

Zero3496: OK. What I want to confirm is that no matter how I change it, will it not merge? Depending on the answer I will stop continuing

Zero3496: > > This will not have a disruptive impact on the game itself.
> 
> I am pretty sure it does have disruptive impact - existing `config` folder will be unconditionally deleted during upgrade from a different version and it won't be replaced with anything as we do not bundle `config` folder by default.

OK. What I want to confirm is that no matter how I change it, will it not merge? Depending on the answer I will stop continuing or you can close then.

Zero3496:
> People who will assemble custom APK can add this patch themselves.

With all due respect, I think there's a huge gap between you and the players. Objective reality has completely failed to guide you.  Kevin always says“special snowflake syndrome”. I'm afraid there's no more special snowflake than you and Kevin.

Zero3496:
Finally, it is hard for debate to have a real impact on these mature people. I won't waste my time on it.

Issue #62446, opened 2022-11-28
Item groups with type "distribution" lack proper consistency checks


mqrause:
### Describe the bug

There are item groups that can cause rare CI errors as already pointet out in other issues (#60784, #59873). They're hard to identify because consistency checks are incomplete. `Item_spawn_data::check_consistency` is supposed to create a maximized item group, but completely ignores it for groups with the type distribution.

https://github.com/CleverRaven/Cataclysm-DDA/blob/4034a38d4bae1abb7f208ba1a63630e05de02662/src/item_group.cpp#L701-L721

For collections it checks for `flags & spawn_flag::maximized` to spawn everything without randomness, but it's obviously not that easy for distributions where there are multiple paths that would all have to be checked.

### Steps to reproduce

See code and random CI errors.

I guess one could create a test group with a container and two possible content items of which one does and one doesn't fit into the container so that it fails the check 50% of the time depending on which one got selected, when it really should always fail.

### Expected behavior

Item groups don't have errors and creators are informed if there can be issues.

### Versions and configuration

4034a38d4bae1abb7f208ba1a63630e05de02662

PR #63174, opened 2023-01-18
Prevent loot-generation errors during ranged_balance_test

Kromgart:
#### Summary
Bugfixes "Prevent loot-generation during ranged weapons tests"

#### Purpose of change

Ranged balance test was occasionally failing due to rare loot-generation issues (details below). 
This test is about 'shooting' creatures and testing damage output, dispersion etc. The items generated on creature's death are not used in any way. Therefore I suggest to turn off item drops for these 'test monsters' and avoid random build errors in PR workflows (see this PR for example: #63158 )

#### Describe the solution

After the change (setting _monster.death_drops_ to false), this should not run, when the monster::die() is called (src/monster.cpp):
```
if( death_drops && !no_extra_death_drops ) {
     drop_items_on_death( corpse );
     spawn_dissectables_on_death( corpse );
 }

```

Otherwise the following can happen:
During loot generation, sometimes a container has too many items spawned for it (too much volume or weight). In game we get an error message saying something like:

ERROR : src/item.cpp:1495 [ret_val<void> item::put_in(const item &, item_pocket::pocket_type, const bool)] tried to put an item (creepy_doll) count (1) in a container (purse) that cannot contain it: pocket unacceptable because not enough space

This can happens even for regular zombies:
1. Zombie dies and its loot is generated
2. It can get  'small_bags' item_group, which has "contents-group": "default_zombie_items_bags_small"
3. 'small_bags' can become a 'purse' (5 L volume)
4. 'default_zombie_items_bags_small' in general has some small items + 50% chance for 'default_zombie_items_pockets'
5. 'default_zombie_items_pockets' also has small items and can get a 'softdrugs' category. Up to now the total volume of generated items very rarely reaches 3 L volume
6. But the 'softdrugs' item_group can potentially give us one of these:
  - 'alcohol_wipes' (which come in a small cardboard box (2.25 L)
  - 'used_1st_aid' item_group, which is a bunch of misc medical stuff with a 'container-item': '1st_aid_box' (2.9 L)  
7. When the code tries to put items into the purse, and there is not enough capacity, the game uses this debugmsg:

src/item.cpp:

```
ret_val<void> item::put_in( const item &payload, item_pocket::pocket_type pk_type, 
                            const bool unseal_pockets ) 
{      
    ret_val<item_pocket *> result = contents.insert_item( payload, pk_type ); 
    if( !result.success() ) { 
        debugmsg( "tried to put an item (%s) count (%d) in a container (%s) that cannot contain it: %s", 
                  payload.typeId().str(), payload.count(), typeId().str(), result.str() ); 
    } 
    ...

```

During the game it causes an error, but is a mild annoyance, the game can be continued fine (the excessive items are gone though).
But when this happens during tests, the PR workflow fails:

```
 ([slow] ~starting_items)=> (continued from above) ERROR : src/item.cpp:1495 [ret_val<void> item::put_in(const item &, item_pocket::pocket_type, const bool)] tried to put an item (creepy_doll) count (1) in a container (purse) that cannot contain it: pocket unacceptable because not enough space20.332 s: shot_features
([slow] ~starting_items)=> 18.893 s: shot_features_with_choke
([slow] ~starting_items)=> 154.325 s: starve_test
([slow] ~starting_items)=> 81.990 s: starve_test_hunger3
([slow] ~starting_items)=> 89.891 s: all_nutrition_starve_test
([slow] ~starting_items)=> ===============================================================================
([slow] ~starting_items)=> All tests passed (44056 assertions in 13 test cases)
([slow] ~starting_items)=> 
([slow] ~starting_items)=> 
([slow] ~starting_items)=> 01:26:41.610 INFO : Ended test at Wed Jan 18 01:26:41 2023
([slow] ~starting_items)=> 
([slow] ~starting_items)=> 01:26:41.611 INFO : The test took 1520.03 seconds
([slow] ~starting_items)=> 01:26:41.611 INFO : Randomness seeded to: 1674003632
([slow] ~starting_items)=> 01:26:41.611 INFO : Treating result as failure due to error logged during tests.
test exited with code 1
```

Remove the debugmsg in this case (maybe a warning could be logged somewhere?)
Instead of showing in-game error create a tiny static analysis tool to find 100% of such potentially problematic item_groups (when there is a non-zero probability of the container's capacity being overflown by the generated loot). This could be run anytime on the json data and report all such possible issues to the developer/modder instead of occasionally popping error messages to the players (and failing PR workflows). 

#### Testing

Assign custom drops with deliberately too much loot in some container for the test monster. Run the tests, observe no errors from ranged_balance_test

### DIFF
--- a/tests/map_helpers.cpp
+++ b/tests/map_helpers.cpp
@@ -142,9 +142,11 @@ void clear_map_and_put_player_underground()
     get_player_character().setpos( { 0, 0, -2 } );
 }
 
-monster &spawn_test_monster( const std::string &monster_type, const tripoint &start )
+monster &spawn_test_monster( const std::string &monster_type, const tripoint &start,
+                             const bool death_drops )
 {
     monster *const added = g->place_critter_at( mtype_id( monster_type ), start );
+    added->death_drops = death_drops;
     cata_assert( added );
     return *added;
 }
--- a/tests/map_helpers.h
+++ b/tests/map_helpers.h
@@ -21,7 +21,7 @@ void clear_map( int zmin = -2, int zmax = 0 );
 void clear_radiation();
 void clear_map_and_put_player_underground();
 monster &spawn_test_monster( const std::string &monster_type, const tripoint &start,
-                             const bool death_drops = true );
+                             bool death_drops = true );
 void clear_vehicles( map *target = nullptr );
 void build_test_map( const ter_id &terrain );
 void player_add_headlamp();
--- a/tests/map_helpers.h
+++ b/tests/map_helpers.h
@@ -20,7 +20,8 @@ void clear_zones();
 void clear_map( int zmin = -2, int zmax = 0 );
 void clear_radiation();
 void clear_map_and_put_player_underground();
-monster &spawn_test_monster( const std::string &monster_type, const tripoint &start );
+monster &spawn_test_monster( const std::string &monster_type, const tripoint &start,
+                             const bool death_drops = true );
 void clear_vehicles( map *target = nullptr );
 void build_test_map( const ter_id &terrain );
 void player_add_headlamp();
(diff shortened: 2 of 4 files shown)


mqrause:
I suggest adding a parameter to `spawn_test_monster` to turn death drops off, since I'm pretty sure some other tests (e.g. throwing tests) can also kill test monsters.

Also want to mention #62446 as the actual cause for the errors.

PR #63177, opened 2023-01-19
Remove Ultica_iso from overmap tilesets

irwiss:
#### Summary
None

#### Purpose of change

Isometric tilesets currently don't work with overmap, smashbutton was already disabled but new ultica iso could be chosen

#### Describe the solution

Disable any tileset where id ends with "_iso" from overmap tileset selection
(unfortunately the json with iso field isn't parsed at that point)

#### Testing

Try selecting smashbutton or ultica iso for overmap tileset, shouldn't work

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -2175,9 +2175,12 @@ void options_manager::add_options_graphics()
     get_option( "USE_TILES_OVERMAP" ).setPrerequisite( "USE_TILES" );
 
     std::vector<options_manager::id_and_option> om_tilesets = build_tilesets_list();
-    // filter out SmashButton_iso from overmap tilesets
+    // filter out iso tilesets from overmap tilesets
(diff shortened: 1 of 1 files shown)


Night-Pryanik:
I think it's better to somehow mark tileset as isometric (some kind of flag?), otherwise tilesets called not like `foo_iso` (take `foo_isometric` as an example) will still be available for overmap.

irwiss:
There's already a way - it's marked in the tileset's json, but that'll require parsing the json in options class and handling all the failure modes (bad json, io exceptions etc), refactoring more code options so build_tilesets_list also returns iso bool and the tileset options code handles it etc...

Iso tilesets are added once in a blue moon I don't think it's worth the effort
There's even non zero odds someone will make overmap handle iso mode sooner than someone creating the next iso tileset

PR #63178, opened 2023-01-19
Cutting logs with power tools

Lalomb0000:
#### Summary
Content "Cutting logs with power tools"

#### Purpose of change

To make use of the utility of power tools and a faster method of crafting

#### Describe the solution

Added a recipe for the crafting of planks from logs with power tools

Keep using axes

#### Testing

Tested and crafted the recipe, if possible, adding it to the camp crafting list
#### Additional context

None

### DIFF
--- a/data/json/recipes/other/parts_construction.json
+++ b/data/json/recipes/other/parts_construction.json
@@ -660,7 +660,7 @@
     "result": "2x4",
     "category": "CC_*",
     "subcategory": "CSC_*_NESTED",
-    "skill_used": "fabrication" ,
+    "skill_used": "fabrication",
     "difficulty": 2,
     "time": "10 m",
     "autolearn": true,


Karol1223:
You would not be getting 10 planks from this. A log with our current dimension would have a 10cm diameter meaning that it would be able to support two 2x4s in the center. It would definitely be able to support two more on the sides of that and I'd hazard a guess two more on top of that as well, you'd be looking at roughly 6-8 planks, though I'd be leaning more towards 6 from an unprocessed log.

I'd also highly suggest adding at least some splinters as byproducts to this since you'd be losing tons of mass with this recipe.

On top of that, I'd suggest making it autolearn at fab 2 or 1. Currently fab 2 is when the recipe to make long planks from heavy wooden beams is learned (though it may be better off as fab 1, I am not entirely sure there) and making planks from a raw unprocessed log is way harder.

Karol1223 (approved):

PR #63181, opened 2023-01-19
Use fake curtains to block vision

SurFlurer:
#### Summary
Bugfixes "Use fake curtains to block vision"

#### Purpose of change
Currently, there's no fake curtains generated when the vehicle is turning. This makes monsters able to see what's in the vehicle.

To test this, you can spawn a `surv_rv`, and drive northeast. 
Stop driving, and enter the center area of the RV, where there are closed curtains on both sides. Walk through every passable tile there, and notice that you are able to see outside the RV.

#### Describe the solution
Generate fake curtains, so that vision is correctly blocked.

#### Testing
Spawn a `surv_rv`, repeat the above steps, and find that you cannot see outside the RV.

### DIFF
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -6066,7 +6066,7 @@ void vehicle::refresh( const bool remove_fakes )
             add_fake_part( rp.first, "CAMERA" );
         }
         // add fake curtains so vision is correctly blocked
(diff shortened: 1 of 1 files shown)


SurFlurer:
To better illustrate this, I'll add some screenshots.
1. Expected vision in a surv_rv:
image
2. Unexpected vision in a slant surv_rv:
image
3. After this fix:
image

PR #63187, opened 2023-01-19
Mole Crickets and Molerats can give tetanus

Karol1223:
#### Summary
Balance "Mole Crickets and Molerats can now give tetanus with bites"

#### Purpose of change
I recently remembered how often I hear people locally say that Mole Cricket bites are dangerous primarily because they can give you tetanus if not cleaned, due to the nature of the animal and the infection. Coincidentally Mole Crickets in the game are really damn boring. A match made in heaven.

#### Describe the solution
Mole Crickets were given a bite attack that has a chance to inflict tetanus. Molerats have had their bites adjusted to be able to give tetanus.

I initially didn't include Molerats since they do not fit the region very well but as long as they're still in the game I don't care that much, I shouldn't exclude them just because they may be removed at some point. For Mole Crickets, I am not sure whether the bite is particularly balanced. @Venera3 I'd appreciate your feedback on that matter specifically, they're your children.

### DIFF
--- a/data/json/monsters/insect_spider.json
+++ b/data/json/monsters/insect_spider.json
@@ -2623,7 +2623,7 @@
     "melee_skill": 6,
     "melee_dice": 4,
     "melee_dice_sides": 6,
-    "melee_damage": [ { "damage_type": "cut", "amount": 10 } ],
+    "melee_damage": [ { "damage_type": "cut", "amount": 7 } ],
     "armor_bash": 15,
     "armor_cut": 18,
     "armor_bullet": 20,
@@ -2639,7 +2639,7 @@
     "special_attacks": [
       {
         "type": "bite",
-        "cooldown": 5,
+        "cooldown": 3,
         "damage_max_instance": [ { "damage_type": "stab", "amount": 20, "armor_penetration": 8 } ],
         "effects": [ { "id": "tetanus", "chance": 5, "duration": 2, "permanent": true } ]
       },


anoobindisguise:
Should tetanus require it to do damage to you to infect you? Looks like it will currently apply even if it's totally blocked/armored, right?

Karol1223:
> Should tetanus require it to do damage to you to infect you? Looks like it will currently apply even if it's totally blocked/armored, right? Or does the Bite attack have a specific provision that already handles that

I copied this pretty much 1:1 from centipede's venom. If tetanus does get applied even when the damage is blocked, which I do admit I haven't tested if it does, then there's a deeper problem at the core here I'm afraid.

Venera3:
Special attacks need to do damage to apply their effects (unless `effects_require_dmg` is set to false), so that should be fine. 
Mole crickets aren't mine for the record.

Karol1223:
> Mole crickets aren't mine for the record.
Yeah but you're the bug man and I know you like your balance so that's why I pinged you here to check if the bite made sense.

Venera3:
Yeah, 10-20/8 looks about right. I might bump their base damage down since you have bite for chomping and the impale for the stabby lil' clawses (it could use a lower cooldown, though).

Karol1223:
Alright took down their melee damage from 10 to 7 and changed the cooldown on the bite from 5 to 3, hopefully that's better.

Mister-Cultist:
thanks I hate it

PR #63189, opened 2023-01-19
Half Worm volume and mass changes

Karol1223:
#### Summary
Balance "Number changes for half worms"

#### Purpose of change
On the devcord it was lately brought up that it's better for a giant worm to split if you want to butcher it as two half worms give more resources than one giant worm - the laws of physics have been foiled again.

#### Describe the solution
Half worms had their numbers changed to roughly half of giant worm's (the mass is exactly half, the volume is rounded down) while the large worms now use the numbers half worms used to have (large worms always used them, anyway, as a part of copy_from)

### DIFF
--- a/data/json/monsters/insect_spider.json
+++ b/data/json/monsters/insect_spider.json
@@ -133,8 +133,8 @@
     "description": "A disturbingly large mutated subterranean creature similar to a nightcrawler.  It possesses a fanged mouth and a long slender body that comes up to your knees, with even more surely hiding underground.",
     "copy-from": "mon_halfworm",
     "looks_like": "mon_worm",
-    "volume": "30 L",
-    "weight": "40750 g",
+    "volume": "62500 ml",
+    "weight": "81500 g",
     "dissect": "dissect_troglobite_sample_single",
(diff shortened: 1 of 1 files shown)


Maleclypse:
I've set it to rerun all tests to see if that causes json styling test to run

Karol1223:
It doesn't appear to have changed anything... Weird, the only test lest is Windows Build but it doesn't add the specific label...

data/json/monsters/insect_spider.json
```diff
@@ -74,8 +74,8 @@
     "default_faction": "worm",
     "bodytype": "snake",
     "species": [ "WORM" ],
-    "volume": "62500 ml",
-    "weight": "81500 g",
+    "volume": "45 L",
+    "weight": "60 kg",
```
Fris0uman, line 78: But this is still more than half the full worm, no?

Karol1223: It is more than half the full _smallest_ worm, yes. That size variant does not split. When it grows it becomes 120kg and 92.5 l of volume and only then can it split.

PR #63192, opened 2023-01-19
[Backrooms] Backrooms artifacts

onura46:
#### Summary
Mods "[Backrooms] Adds 10 rare artifact variants to the Backrooms"

#### Purpose of change
Add 10 rare artifact variants to the Backrooms found strewn about the facility

#### Describe the solution
This PR adds 10 new artifacts which will very rarely be found laying on the ground and inside furniture in the Backrooms. The rate of spawning should be ~1:40 overmap tiles (assuming I did my math right), which should make them pretty rare. They're not intended to be as rare as normal CDDA relics, since they're more just like a natural side-effect of the Backrooms. I can rebalance if people complain of finding them too much, but my first intent is to add a little more mystery and character to the Backrooms and I hope players find at least one or two in a run.

Like my other additions to the mod, all of this is original content meant to present a Cataclysm-flavored experience with no attachment to other lore or canons. Some of the items are pretty spooky. Others are just a bit quirky and unexplainable.

<details><summary>SPOILERS: Backrooms Artifact Descriptions</summary>
<p>

Names used here come from IDs. All artifacts are known only as "strange objects" in-game and their descriptions all begin: "An object of uncertain origin, unusually heavy for its size."

- **The Stapler**: It looks like a glob of white plastic streaked with bits of red and black has melted over a jagged chunk of mixed scrap metals.  There are two large and sharp pieces of metal jutting out from the side in parallel.  The metal is very slightly, inexplicably slimy to the touch.  If you hold it up to your ear, it emits a low, oscillating hum very quietly.  Unsettling…
- **Hybrid Fossil**: It's a brown, flaky rock just bigger than your fist.  Horrifically, what seem to be small animal bones stick out of it at weird angles.  Something that looks suspiciously like blood very slowly oozes out from between the layers in the rock - almost unnoticeably slow - almost as if something living were freshly crushed between the sedimentary layers.  But that would be impossible…
- **Disappearing Box**: It's a small wooden cuboid with a large rectangular hole running from one side to the other.  The wood is well-treated and it seems completely mundane at first. If you look through the hole, everything looks normal.  It's only when you wave your hand in front that you realize that you can't see yourself!  If you stick your finger inside the hole, it doesn't seem to come out the other side.  How is that possible?!
- **Bong Chihuly**: You could be wrong, but it appears to be maybe 6 or 7 colorful glass bongs fused together in a ball-shaped mess of melted glass, reminding you strangely of an art exhibition you saw once.  When you pick up the object, you notice a pale purple ooze trapped in the middle of the glass meld.  It's iridescent, colloidal, and slightly viscous, although it doesn't remind you of any fluid you've ever seen before.
- **Demon Horn**: It's colored jet black and looks suspiciously like the ridged horn of a goat, only seemingly studded with red gems and carved with inscriptions.  You're no expert, but these inscriptions are unlike anything you've ever seen humans make, with many geometric shapes inside squares that could be words or characters.  There are thin, intricate lines weaving all over the horn, curving in a way that you're sure could only be achieved with laser engraving technology yet you feel certain they were not.
- **The Nineteenth Bible**: It appears like a normal book with a bright blue cover from a distance, but when you pick the object up you notice it's simply a thick slab of some dense material shaped like a book.  It's oddly stocky and brightly colored, textured with low-resolution images almost like an object that fell straight out of an old video game.  On the cover is just the number "19" and below it the words "A new bible by" but there's no author listed.  In fact, there's no other text on the entire object.
- **Dimensional Amber**: It seems to be a heavy chunk of light pink amber.  Suspended inside the amber are… fossils?  You're unsure what to call them. One appears to be a small ball of hair with copper wires reaching out in one direction, reminiscent of an amoeba reaching for something.  Another appears to be a tiny fern, only made out of a highly iridescent metal, with blades quite literally as thin as a razor.  A third, particularly horrifying one appears to be the shrunken head of a human with a network of dark red nerves instead of hair.
- **Meat Bop-It**: It seems almost like a children's toy has somehow melted into a solid chunk of brass.  There are bright red, green, and yellow plastic pieces fused into the sides and a very distinct purple button in the middle.  When you press it, however, there's a sickening squelch and you can't help but feel like you're pressing into a pile of ground meat protected by this metal outer shell.  Disgusting.
- **Velvet Rock**: It seems like a small, jagged rock uniformly and evenly covered in bright red velvet.  You feel certain that the rock came to be this way through some supernatural process.  When you hold it to your ear, it emits a very low, oscillating hum.  Very strange.
- **Impossible Rainstick**: It appears to be a small black box made out of some kind of hard metal.  Although it has no discernable openings, you call it a box because when you shake it you can feel things moving around inside.  Strangely, though, what's inside always seems to be… changing.  One moment it's a liquid, the next it feels like marbles or rocks, and then the next it feels like sand. Although probably inconsequential, the object emits a fairly strong odor of ginger at all times.

</p>
</details>

Oh yeah, and if you damage them they'll start emitting radiation. So be careful!

#### Testing
Turned up the spawn rate to test artifact effects, then turned it back down for a playtest to make sure they weren't too common. Everything seems to be working as expected.


anoobindisguise:
These are probably good enough to include in Vanilla imo.

PR #63193, opened 2023-01-20
use open glass door looks-like for open red & green glass door

MrDraMaster:
#### Summary
Bugfixes "Make open red & green glass doors look like open glass doors"

#### Purpose of change

Noticed that some glass doors didn't look "open" when opened.

#### Describe the solution

change the `_c` suffix on the faulty looks-like entries to `_o`

do nothing and keep wondering whether those doors are open or not

#### Testing

spawned in the closed door variant of each door, open & closed them repeatedly, watching the sprites follow along


NetSysFire (approved):
Nice catch. Looks like a copy-paste error by the one who added those.

Issue #63195, opened 2023-01-20
Enchantments increasing Max HP continually reset to pre-enchanted value


Standing-Storm:
### Describe the bug

1. When applying an enchant that increases HP through the MAX_HP ID value, HP resets to the pre-enchant base every few of in-game minutes. 

### Steps to reproduce

1) Start a world with Magiclysm
2) Debug in the Gaia's Chosen attunement, which uses an enchantment attached to it to increase max HP by 30%.

>   "values": [ { "value": "MAX_HP", "multiply": 0.3 } ], )

3) Debug in the Restoration druid spell and cast it
4) Watch as HP heals up slightly toward the new value but then resets down to pre-enchantment max

### Expected behavior

An enchantment that says it increases Max HP should allow HP to actually reach the new Max HP.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.2486 (21H2)
- Game Version: 950acde [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Magiclysm [magiclysm]
]

alp7292:
also check #61701

Issue #61701, opened 2022-10-15
Artifact passive  "MAX_HP" is not working


KymasTuran:
### Describe the bug

The extra hp show, but the hit point reset back to is original number after a turn. 

### Steps to reproduce

Example: you have 100 hp plus the bonus an artifact with this passive give like 110 in total, heal from any source, you got 105 hp. A few moment later you got 100 again.
You need : an artifact with MAX_HP and some way to fast heal.

### Expected behavior

The normal extra HP bonus without losing it few moment later.

### Versions and configuration

Windows 7
cdda-windows-tiles-x64-msvc-2022-10-11-1637

### Additional context

Still single.

alp7292:
MAX_HP doesnt works in any case for me my suggestion is fixing MAX_HP enchantment and applying into the traits and mutations instead of using hp_modifier_secondary and hp_modifier as they causes issues if two different mutation use hp_modifier or hp_modifier_secondary

GuardianDll:
Yep, just decided to test it, but something brick it up.  Regeneration also doesn't work
```json

  {
    "type": "mutation",
    "id": "DEBUG_HP",
    "name": { "str": "Debug 9000 HP" },
    "points": 99,
    "valid": false,
    "description": "If you need to survive the bug attacks.",
    "relic_data": { "passive_effects": [ { "values": [ { "value": "MAX_HP", "add": 9000 } ] } ] },
    "debug": true
  }

```


```json

  {
    "type": "mutation",
    "id": "DEBUG_HP_REGEN",
    "name": { "str": "Debug Regeneration" },
    "points": 99,
    "valid": false,
    "description": "Heal yourself from this stupid bug attacks.",
    "relic_data": { "passive_effects": [ { "values": [ { "value": "REGEN_HP", "multiply": 100 } ] } ] },
    "debug": true
  }
  
  ```

alp7292:
> Yep, just decided to test it, but something brick it up. Regeneration also doesn't work
> 
> ```json
>   {
>     "type": "mutation",
>     "id": "DEBUG_HP",
>     "name": { "str": "Debug 9000 HP" },
>     "points": 99,
>     "valid": false,
>     "description": "If you need to survive the bug attacks.",
>     "relic_data": { "passive_effects": [ { "values": [ { "value": "MAX_HP", "add": 9000 } ] } ] },
>     "debug": true
>   }
> ```
> 
> ```json
>   {
>     "type": "mutation",
[...]

PR #63197, opened 2023-01-20
Move Magiclysm Gaia's Chosen attunement HP increase from enchant to mutation

Standing-Storm:
#### Summary
Bugfixes "Since the MAX_HP function in enchantments is bugged as described in https://github.com/CleverRaven/Cataclysm-DDA/issues/61701 and https://github.com/CleverRaven/Cataclysm-DDA/issues/63195, this moves the HP increasing part of the Gaia's Chosen Magiclysm attunement from an attached enchantment to the Gaia's Chosen mutation itself."

#### Purpose of change

Workaround for MAX_HP not working, plus since mutations can already increase hp there's no reason not to use the existing function. This matches the Crusader attunement, which also uses hp_modifier directly on the CRUSADER mutation.

#### Describe the solution

Replace a MAX_HP enchantment with hp_modifier in the GAIAS_CHOSEN mutation.

None

#### Testing

Made the change in my own game first to make sure that it works. It does.

### DIFF
--- a/data/mods/Magiclysm/traits/attunements.json
+++ b/data/mods/Magiclysm/traits/attunements.json
@@ -415,6 +415,7 @@
     "valid": false,
     "description": "The Champion of the Earth, Gaia's Chosen is blessed on all things nature, animate and inanimate.",
     "enchantments": [ "GAIAS_CHOSEN" ],
+    "hp_modifier": 0.3,
     "prereqs": [ "DRUID", "EARTHSHAPER" ],
     "spells_learned": [ [ "gaiaschosen_samsara", 5 ] ],
     "cancels": [


GuardianDll (changes requested):
I'm sorry, i know this is annoying bug, but desire fix for this is not replacing the one enchantment effect (out of many) to not enchantment, but fix the enchantment itself. You can play with this fix for yourself, but i can not approve it for the game

alp7292:
if two mutations have hp_modifier then only one works currently code is mess swapping to hp_modifier will only cause more conflict with other mutations that use hp_modifier

Standing-Storm:
> if two mutations have hp_modifier then only one works currently code is mess swapping to hp_modifier will only cause more conflict with other mutations that use hp_modifier to solve the issue someone needs to fix MAX_HP and replace every mutation that uses hp_modifier and hp_modifier_secondary with MAX_HP and removing useless hp_modifier hp_modifier_secondary code from game

Ah, alright, I didn't realize they didn't stack. Never mind then.

PR #44373, opened 2020-09-23
Rename skills and merge speech-barter

I-am-Erk:
#### Summary

SUMMARY: Infrastructure "Rename skills in preparation of some changes; merge speech & barter."

#### Purpose of change

- Several skills (traps, lockpicking, speech, barter, and swimming) are noticeably more niche than others.
- Proficiencies create an environment where we want skills to represent transferable umbrella abilities, and proficiencies represent more specific subsets of knowledge. This encourages us to try to make most skills cover recognizable sets of transferable motor skills and things.

#### Describe the solution
Mostly, this PR just renames skills into the concepts I hope for them to encompass later.

- Chemistry > **Applied Science**. It took an embarrassingly long time to come up with this name, but it's what we want and it makes it clear that this isn't a research skill but a general skill for things ranging from organic chemistry to metallurgy to microbiology.
- Swimming > **Athletics**. For the moment, it's still just swimming, but in the future it could affect things like your ability to keep your balance on rough terrain, your stamina, and your ability to climb. If you have wings and can fly it might also affect how long you can manage that and how precisely.
- Cooking > **Food Handling**. This makes it less counterintuitive to move stuff like butchery in there (at least partly), and really brewing and canning are loose fits for cooking already.
- First Aid > **Health Care**. This skill can be used for more than just emergency medical treatment, especially as we get proficiencies for things like actual medicine.
- Driving > **Vehicles**. This is a strict name change, we're already using this skill for more than driving and have been for ages.

The biggest change is merging "speech" and "barter" into a single skill, "**Social**". In the future, things like bartering can be managed by proficiencies of Social. For now, I just made everything that refers to barter refer to Social instead. Both these skills are hard to train and find uses for even after all our NPC boosts, and *highly* transferable. I hope they won't be nearly the same level of controversy as my other ongoing skill merge project.

**OTHER CHANGES**
Adds an "obsolete" attribute (true/false, default false) to skill factory. Flagging this as true prevents skills from being listed in get_skills_sorted_by, which is a pretty root level function that I believe is called any time we want to list skills, so it should prevent obsolete skills from showing up anywhere at all.

Special thanks to KorGgenT for helping get me through that code.

#### Describe alternatives you've considered

I really want to change "mechanics" and "computers" into "grease monkey" and "code monkey", but THE MAN won't let me.

[...]

PR #63199, opened 2023-01-20
Remove athletics training for waiting in water

### LINKED ISSUE #63190: Waiting In Water Training
### Describe the bug

Waiting in water:

Trains athletics;
Does not cause activity;
Restores stamina.

While I'm all for it restoring stamina, at least with sufficient athletics, it should NOT train athletics without causing at least LIGHT activity.

### Steps to reproduce

Strip on a fresh character, dive into deep water, and use the wait command.

### Expected behavior

Either no training or a activity level.

### Screenshots

_No response_

- OS: Windows
- Game Version: 10bcab1 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
    Magiclysm [magiclysm],
    Generic Guns [generic_guns],
    Sky Island [skyisland]
]

### Additional context

_No response_

cake-pie:
#### Summary
Bugfixes "Remove athletics training for waiting in water"

#### Purpose of change
Resolves #63190 -- waiting in water trains athletics with neither physical activity nor stamina drain

#### Describe the solution
Remove training of "athletics" skill when waiting in water.

This is intended as a stopgap. It should be replaced with proficiency gain in a new proficiency specifically for swimming, but that will have to be added after the string freeze. See "Additional context" for further discussion.

Add physical activity and drain stamina. This would have undesired side effect as it removes the possibility to stop and tread water in order to recover stamina when having to swim long distances.  

See "Additional context" details.

#### Additional context
Digging into history, this piece of code has existed for a very long time. It should be noted that the skill id being practiced in the code here is `swimming` -- the original intention here was:
- there is a skill specific to swimming ability
  - i.e. knowing various swimming strokes and techniques v.s. just flailing or dog paddling
- moving in water ( = "swimming" ) would have checks against that skill, and drain stamina
- it is possible to wait in water ( = "treading water" )
  - this would be a way to stop and recover stamina when having to swim a long distance
  - treading water still improves skill at swimming (with respect to technique)

In #44373 the skill was renamed to "athletics" by changing the skill name string and description without modifying the skill id. See [here](https://github.com/CleverRaven/Cataclysm-DDA/pull/44373/files#diff-01cd937af3bbfec07928be7268f16a131ddd218f1f0304106ee5e6436373f893L63). The intention behind the change was to expand the scope:

> could affect things like your ability to keep your balance on rough terrain, your stamina, and your ability to climb. If you have wings and can fly it might also affect how long you can manage that and how precisely.

Since then, it has evolved into a catch-all for cardio fitness and stamina, while still retaining a reference to "swimming endurance":

https://github.com/CleverRaven/Cataclysm-DDA/blob/c0335423ae9aed250c392a323372d0dc9170c112/data/json/skills.json#L53-L60

IRL swimming is of course a great exercise for improving general cardio fitness and stamina, but simply treading water should not lead to significant gains in that department -- I seem to recall seeing someone exploit this in a youtube video by just entering a water tile and staying there for a long time for the skill gain.

Instead, we can achieve something closer to the original intent by decoupling technique vs endurance. To accomplish this, implement a swimming _proficiency_ for "swimming skill" in terms of good technique and knowledge of strokes and styles, as opposed to the athletics _skill_ which covers endurance and stamina.

Also, it'd probably be a good idea to migrate the skill id from `swimming` to `athletics` to reduce potential confusion when reading source code.

#### Testing
Sit in deep water and wait for a long time with no skill gain.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -11577,11 +11577,9 @@ void Character::pause()
     if( !in_vehicle && !get_map().has_flag_furn( "BRIDGE", pos( ) ) ) {
         if( underwater ) {
             // TODO: gain "swimming" proficiency but not "athletics" skill
-            // practice( skill_swimming, 1 );
             drench( 100, get_drenching_body_parts(), false );
         } else if( here.has_flag( ter_furn_flag::TFLAG_DEEP_WATER, pos() ) ) {
             // TODO: gain "swimming" proficiency but not "athletics" skill
-            // practice( skill_swimming, 1 );
             // Same as above, except no head/eyes/mouth
             drench( 100, get_drenching_body_parts( false ), false );
         } else if( here.has_flag( ter_furn_flag::TFLAG_SWIMMABLE, pos() ) ) {


Night-Pryanik:
No, don't leave commented code. Leave TODO, but remove the commented code.

alp7292:
how do we gain xp fot athlethic then

cake-pie:
> No, don't leave commented code. Leave TODO, but remove the commented code.

Done.

> how do we gain xp for athlethic then if there is no alternative added.

By actually swimming (*moving* in water) and not just waiting / passing time while remaining stationary in water.

https://github.com/CleverRaven/Cataclysm-DDA/blob/c0335423ae9aed250c392a323372d0dc9170c112/src/avatar_action.cpp#L612

PR #60750, opened 2022-09-04
Furniture Project: Seating

Karol1223:
#### Summary
Content "A lot of changes regarding seating furniture and some new one to boot"

#### Purpose of change
It was pointed out to me that at the moment mapgen lacks creative freedom, so to speak, in some definitions of this word, because we lack furniture variety. Adding new materials or colours to existing furniture is one thing but since a lot of current furniture is a bit of a mess cleaning it up can go a long way as well.

#### Describe the solution
**Chairs**
- [x] Rework existing foldable chairs to fold properly, make sense with flags, disassembly and materials
- [x] Rework existing wooden chairs to be able to be taken down into an item form
- [x]  Add a bunch of new chairs

**Sofas, Seats and Armchairs**
- [x] Rework sofas and armchairs so that their disassembly recipes make sense
- [x] Rework armchairs to be able to be taken down into item form
- [x] Remove sofa construction recipe
- [x] Rework the airplane seats to give more accurate resources when disassembled

**Stools**
- [x] Give the log and wooden stools appropriate crafting recipes and item forms
- [x] Rework the disassembly recipes for both of them to make more sense
- [x] Give the wooden stool comfort value
- [x] Add a bar stool

**Benches**
- [x] Remove the construction recipe for both the wooden and the metal bench
- [x] Rework the disassembly recipes for both benches to make more sense
- [x] Add a park bench and a craftable wooden bench

**Other changes**
- [x] Add some new wood resources and have them spawn in the world
- [x] Rework some existing wood resources to make them more in line with what they should be
- [x] Add black paint apparently, didn't plan for this one
- [x] Nest all the new furniture and woodworking crafting

#### Describe alternatives you've considered
Initially considered messing with mapgen to make new furniture spawn but I would probably not survive that. Also considered reworking the swivel chairs with this but the vehicle code scares me, that's for another day and hopefully another contributor. Other than that I've had a few more chair types in mind but I either couldn't find good references or decided not to so that if anyone wants to add the new furniture they have a less intimidating amount of variety to work with.

#### Testing
Everything appears to be working fine. No crashes anymore, nested crafting looks to be fine, too

#### Additional context
Big thanks to abchiptop/a-chancey for helping me out (pretty much doing all the hard work, all I did was add the code for it) with the woodworking side of this PR. Without their help, I would most likely just have given up and had the furniture give either not-fitting pieces or splintered wood.

PR #63201, opened 2023-01-20
Add's crafting item to camps for planks. 

Profit0004:
#### Summary
Content "Adds a plank recipe to the camp menu "

#### Purpose of change

This is to offer another method of crafting planks from logs rather than zones.

#### Describe the solution

It simply adds a recipe and links to it. 

Doing nothing?

#### Testing

I tested it in a save to make sure results were as expected. 

#### Additional context

This is less efficient than doing it yourself but I figure NPC's will be sleeping and taking breaks for other things.  about 15 minutes for player and 1 hour for NPC's .


Karol1223:
Look at #63178 which tackles the exact same problem and has its issues already fixed, please. I'd suggest changing this PR to adding the recipe from that PR to the faction camps (once that one gets merged), no point in two PRs adding a recipe to make planks from logs.

Brambor:
> Look at #63178 which tackles the exact same problem and has its issues already fixed, please. I'd suggest changing this PR to adding the recipe from that PR to the faction camps (once that one gets merged), no point in two PRs adding a recipe to make planks from logs.
> 
> Yes, yours does it with manual tools and not powertools, that's worse. You would not be cutting a log with a regular hand woodsaw into planks.

The methods of acquiring (planks) are different. If you don't have tools for one method, you can still do the other. There is no reason not to merge both PRs. And on top of that third can be added (powertools in fraction camps).

Karol1223:
The reason not to merge both PRs is it is unreasonable to make planks from logs with hand tools, as has been discussed in #60750 and as I mentioned in my previous comment.

Profit0004:
> The reason not to merge both PRs is it is unreasonable to make planks from logs with hand tools, as has been discussed in #60750 and as I mentioned in my previous comment.

But we CAN make planks from logs with hand tools already and so can NPC's using zones so that ship has sailed.   This is just a convenient method of doing so inside of a camp.    In the past planks were made by hand as well in the real world, usually by elevating it and having someone on top of the log sawing down on it, but I do not dispute the impracticality, just the fact that since it is already in the game it doesn't change anything and can be removed if the other methods are removed later.

Karol1223:
> But we CAN make planks from logs with hand tools already and so can NPC's using zones so that ship has sailed

The ship indeed has not sailed. If something unreasonable is in the game that is the ground to remove that thing, not to add other similarly unreasonable things. I have indeed made an issue about removing the tool qualities allowing for crafting of planks with hand tools, only today because your PR has reminded me I've never been able to remove it myself.

Karol1223:
After giving it more thought - I do admit that not everything I said was right.

Making planks with hand tools is very much possible, just not with _saws_. A wood saw would not be able to cut through a log, much less cut it into planks. You can, however, use the axe quality for this. The thing is that making planks with axes is extremely inefficient and time-consuming. You'd be looking at 3-5 planks from one log if anything and it'd take at least an hour, with a rough estimate. 

I am saying this because while I did file an issue about axes breaking the laws of physics, my initial issue was about them being able to make planks at all - which isn't an issue in theory. I am saying in theory because the quality of planks obviously wouldn't be comparable to powertool-cut ones but that is something we likely would not need to model. The issue is currently that is a hardcoded use-action which it has no reason to be - that use action has to be edited, and likely it should be changed into a regular recipe in the process. Which it wouldn't need to be if we already had those recipes present.

Profit0004:
> After giving it more thought - I do admit that not everything I said was right.
> 
> Making planks with hand tools is very much possible, just not with _saws_ (any saws, that is, of course there are specific saws designed to cut logs but they are not wieldy enough to be used for plank cutting). A wood saw would not be able to cut through a log, much less cut it into planks. You can, however, use the axe quality for this. The thing is that making planks with axes is extremely inefficient and time-consuming. You'd be looking at 3-5 planks from one log if anything and it'd take at least an hour, with a rough estimate.
> 
> I am saying this because while I did file an issue about axes breaking the laws of physics, my initial issue was about them being able to make planks at all - which isn't an issue in theory. I am saying in theory because the quality of planks obviously wouldn't be comparable to powertool-cut ones but that is something we likely would not need to model. The issue is currently that is a hardcoded use-action which it has no reason to be - that use action has to be edited, and likely it should be changed into a regular recipe in the process. Which it wouldn't need to be if we already had those recipes present.

That all seems extremely reasonable when laid out like that.  When the hand recipe is changed to something a little different, this can be changed with only a single edit as well.   I personally could probably make reasonable planks out of dried Catalpa wood and an axe, and I do make large thin sheets from logs of it as fire starters that are roughly 1 square foot of face at about 1/4 of an inch thick, but it certainly isn't a plank, more like a shingle, and while I can make 100+ an hour with a hatchet, once again they are not planks in the size I cut them.

Profit0004:
The irony that this kicked off such a discussion, I just was watching rycon on youtube and thought it was silly there was no such recipe in the camps rather than make zones which are so unintuitive so I would make one.   Took a bit to figure out how it all of it worked in camps and downloaded a new copy of VS to lint it and make sure it was good.    The discussion has actually been far more robust than any of my thinking and time when I made the change.

Maleclypse (approved):

PR #63216, opened 2023-01-21
Change tree felling to produce furniture `f_trunk` instead of overwriting terrain to `t_trunk`

### LINKED ISSUE #42428: Felling trees inappropriately creates dirt squares
# Describe the bug

Logs from a felled tree create dirt squares under themselves inappropriately overwriting other types of terrain. For instance, a tree growing on top of a LMOE shelter can fall on top of stairs and other trees creating dirt squares.

# Steps To Reproduce

1. Find an LMOE shelter or other raised structure with a tree on top
2. Fell the tree so that it falls on stairs/other trees
3. Chop the resulting logs and notice that dirt squares remain

# Expected behavior

A felled tree should not change terrain under itself.

# Screenshots

Screenshot_2020-07-24_11-45-16

# Versions and configuration

 - OS: Arch Linux
 - Game Version: O.E-2
 - Graphics version: Tiles
 - Ingame language: English
 - Mods loaded: Disable NPC Needs, No Rail Stations, sees-player icon (retrodays), Bionic slots, Alternative map keys

cake-pie:
#### Summary
Bugfixes "Change tree felling to produce furniture `f_trunk` instead of overwriting terrain to `t_trunk`"

#### Purpose of change
Resolves #42428

#### Describe the solution
- [x] add furniture `f_trunk`, it will exist alongside `t_trunk` for now
  - fundamentally identical; move cost adjusted down as it is now calculated in addition to underlying terrain's move cost
- [x] change tree felling so that it produces furniture `f_trunk` instead of overwriting terrain to `t_trunk`
  - [ ] if original tile is open air, should impact next valid z-level below rather than hang in mid-air
- [ ] modify existing methods of chopping up trunks so that both `f_trunk` and `t_trunk` are acceptable inputs (for the time being)

#### Future work
Will not be addressed in this PR

- jsonize deconstruction for `f_trunk`
  - eliminate hardcoded means for chopping up trunks, or  
convert them to wrappers for the jsonized deconstruction
- eliminate `t_trunk` from mapgens etc
- :grey_question: somehow migrate any existing `t_trunk` in existing saves :grey_question:
- eventual complete elimination of `t_trunk`

#### Testing
Not yet

Make the trunk an item rather than furniture, but I don't think it was intended to be something that could be moved around with ease.

#### Additional context
No new strings, as `f_trunk` is copied directly from `t_trunk`

### DIFF
--- a/data/json/furniture_and_terrain/furniture-flora.json
+++ b/data/json/furniture_and_terrain/furniture-flora.json
@@ -1,6 +1,24 @@
 [
   {
     "type": "furniture",
+    "id": "f_trunk",
+    "name": "tree trunk",
+    "description": "A section of trunk from a tree that has been cut down.  Can be sawn into logs with the right tool.",
+    "symbol": "1",
+    "color": "brown",
+    "move_cost_mod": 4,
+    "coverage": 45,
+    "flags": [ "TRANSPARENT", "FLAMMABLE_ASH", "REDUCE_SCENT", "MOUNTABLE", "SHORT" ],
+    "bash": {
+      "str_min": 80,
+      "str_max": 180,
+      "sound": "crunch!",
+      "sound_fail": "whack!",
+      "items": [ { "item": "splinter", "count": [ 5, 15 ] } ]
+    }
+  },
+  {
+    "type": "furniture",
     "id": "f_mutpoppy",
     "name": "mutated poppy flower",
     "description": "These strange flowers have appeared in the wake of the Cataclysm, and their buds can be used for medicinal purposes, like the seeds of the mundane poppy they're named after.  The dirt around them gently churns as their roots writhe beneath the soil, and it's surrounded by an overwhelming floral smell that makes you feel sleepy.",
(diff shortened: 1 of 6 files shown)


PatrikLundell:
Note that you'd have to introduce logic to handle the felling of trees onto destructible terrain (such as other trees and shrubs). Currently felling a tree onto another eliminates that other tree, and I'm unsure what happens if you instead try to place furniture there (could be silently both on the same tile, or could be yet another of these error messages complaining about terrain/furniture clashes).

cake-pie:
@PatrikLundell I was under the impression that the existing code already dealt with that?

https://github.com/CleverRaven/Cataclysm-DDA/blob/90ec3063688f1a08e43d7361837077d78f849036/src/activity_actor.cpp#L5607
https://github.com/CleverRaven/Cataclysm-DDA/blob/90ec3063688f1a08e43d7361837077d78f849036/src/faction_camp.cpp#L4256

PatrikLundell:
Yes, you're correct. That ought to take care of it.

Profit0004:
This may have weird effects when trees are dropped on impassable terrain such as when people drop them on hills, caves and walls.  They are edge cases but I am hoping they are addressed.

anothersimulacrum:
Stale for over a year

Issue #63203, opened 2023-01-20
Axes break the laws of physics when making planks


Karol1223:
### Is your feature request related to a problem? Please describe.

~~Recently #60750 changed some woodworking recipes, and while discussing stuff for that it was brought up how you'd not be making planks directly from logs with hand tools because that's simply unreasonable. Technically doable but the result would be low quality and it'd take way more effort. Even then, you'd not be doing that with an axe but a saw in the first place. Yet the axe did it perfectly fine, in 20 minutes somehow, and even managed to break the laws of physics in the process (10 planks do not fit with the dimensions of our logs). The same use action appears to be used by anything with wood sawing, which while saws can indeed cut wood into planks they can't do it very well on an entire damn log and they certainly cannot break the laws of physics in 20 minutes either. It also appears that same quality is used in faction camps / follower interactions and should be looked into as well.~~ 

Alright then, rewriting this and leaving the old one behind as a testament to my hubris. Making planks with hand tools is possible. Not good, but theoretically possible. The issue is currently axes make amazing planks faster than powertools and do so more efficiently as well. This should not be. Our logs allow for roughly 6-8 planks to fit in them with precision cuts. With an axe you'd be looking at 3-5 planks at best. It should also take far longer, I'd throw a rough estimate of an hour for one batch (it may need to be one hour for one plank, but I have not tried to make planks with an axe myself) instead of ~20 minutes it does now (I think 18 specifically, but that's a problem either way). Ideally, while we're at it, that should probably be moved into a recipe instead of a hardcoded use-action. It has no business being a use-action, much less with nested recipes being a thing now. When it comes to saws, though, they should not be able to turn logs into planks. You'd destroy a wood saw trying to do that.

### Solution you would like.

~~Remove the hardcoded 'a'ctivation use from the axes, wood saws and all the other stuff. Make NPCs not able to do that either.~~ 

Move the LUMBER activity into a recipe that needs an axe quality and provides a long and inefficient log-plank conversion rate. It could likely use a proficiency of its own as well, but I don't know if adding a proficiency specific to just one recipe is something that'd be approved.

### Describe alternatives you have considered.

~~I could handle the json part myself but there'd be dead leftover c++ code laying around still if I did that myself so I decided it's better to let someone who actually has c++ knowledge handle this and not split it into multiple PRs for no reason.~~ 

[...]

PR #57466, opened 2022-05-06
Migrate digging to constructions

andrei8l:
#### Summary
None

#### Purpose of change
Digging can't be paused or resumed very well

Pits can be filled for free

Deep pit is already a possible construction, used as a prerequisite for other constructions, but it uses different mechanics compared to the digging activity

Fixes #54511

#### Describe the solution
Migrate digging to constructions and add the missing bits:
1. add exertion level to constructions
2. add a `do_turn_special` to run arbitrary functions on each turn of the build activity
3. pass the actor to `post_special` and `do_turn_special` too instead of just the location
4. improve the improved `ACT_MULTIPLE_CONSTRUCTION` from #57327 to allow building prerequisites recursively

TODO:

- [x] migrate `ACT_FILL`
- [x] add byproducts to dig/mine down and constr_pit_underground
~maybe pass the actor to `pre_special`s too in order to fix being able to spot traps with the construction direction prompt~ this PR is already enormous

#### Describe alternatives you've considered
Various hacks to make the dig activity properly pausable (ex: spawning fake items to mark the activity status)

#### Testing
Two new unit tests for point 4 above

`iuse` digging from shovels should not be significantly different from before

Exhuming graves is still bad for morale, etc

#### Additional context
This looks pretty scary and risky for 0.G, but most of the code was just moved around and/or deduplicated. I think we should be safe thanks to the test units, but I'm not married to getting this into 0.G

PR #63219, opened 2023-01-21
Remove redundant `FARM_WOOD` constructions

cake-pie:
#### Summary
Content "Remove redundant `FARM_WOOD` constructions"

##### Remove construction id `constr_chop_trunk`
Contributes to #63203, part of a push to remove c++ hardcoded processes in favor of JSON crafting/construct/deconstruct recipes. Removing `constr_chop_trunk` allows us to eliminate `done_trunk_plank` as a construction `post_special` and corresponding c++ `construct::done_trunk_plank`. This sets up `iuse::cut_log_into_planks` as next on the chopping block.

Although this option has been removed from the construction menu, it does not cause any problems as the functionality is actually redundant -- it is also possible to accomplish the same task, by using an axe (by activating it). The aim is to eventually have axe use_action be superseded by a JSON deconstruct recipe rather than using hardcoded c++, and eliminate having the same processes being defined in multiple places that can have inconsistent/divergent requirements/outputs.

##### Remove construction id `constr_crt_chop_trunk`
Apart from the vanilla `constr_chop_trunk`, this mod construction is the only other member of `FARM_WOOD` construction category, which is woefully underused. This construction is similarly redundant to axe use_action; this again leads to potentially inconsistent requirements/outputs being defined in different places for the same thing and should thus be eliminated.

##### Remove construction category `FARM_WOOD`
This is an overspecific and underutilized category, by comparison to other categories. With the removal of both its constructions (that are supported in the official codebase) the category itself is obsolete and can also be deleted.

#### Describe the solution
- remove vanilla construction id `constr_chop_trunk` and construction group id `chop_tree_trunk_into_planks`
  - remove construction `post_special` `done_trunk_plank` and underlying C++ `construct::done_trunk_plank`
- remove CRIT mod construction id `constr_crt_chop_trunk` and construction group id `chop_tree_trunk_into_logs`
- remove construction category `FARM_WOOD`

Only delete `constr_chop_trunk` to facilitate removal of `done_trunk_plank`, which leaves the `FARM_WOOD` category empty unless playing with CRIT mod, in which case it only contains one construction.

#### Testing
`FARM_WOOD` category no longer shows up in construction menu; axe actions to chop trunk to logs and planks function as per normal.

### DIFF
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -73,18 +73,6 @@
   },
   {
     "type": "construction",
-    "id": "constr_chop_trunk",
-    "group": "chop_tree_trunk_into_planks",
-    "category": "FARM_WOOD",
-    "required_skills": [ [ "survival", 2 ] ],
-    "time": "60 m",
-    "qualities": [ [ { "id": "AXE", "level": 2 }, { "id": "SAW_W", "level": 2 } ] ],
-    "pre_terrain": "t_trunk",
-    "post_terrain": "t_dirt",
-    "post_special": "done_trunk_plank"
-  },
-  {
-    "type": "construction",
     "id": "constr_improvised_shelter",
     "group": "build_improvised_shelter",
     "category": "CONSTRUCT",
--- a/data/json/construction_category.json
+++ b/data/json/construction_category.json
@@ -42,11 +42,6 @@
   },
   {
     "type": "construction_category",
-    "id": "FARM_WOOD",
-    "name": "Farming and Woodcutting"
-  },
-  {
-    "type": "construction_category",
     "id": "TOOL",
     "name": "Tools"
   },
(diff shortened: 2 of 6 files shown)


andrei8l:
I very much object to removing the construction for chopping trunks, especially since it takes 60 minutes. Constructions offer a neat, interruptible and resumable framework for interacting with terrain and furniture that has not been easy to reproduce elsewhere.

IMO the better solution is to port `iuse::chop_logs` to use the construction (and deprecating `chop_logs_activity_actor` that's 90% duplicated boilerplate), in the same way that I've ported `iuse::dig`.

cake-pie:
> same way that I've ported `iuse::dig`

Could you point out the relevant PR? I'd check blame on iuse but github struggles with the large file.

andrei8l:
> Could you point out the relevant PR? I'd check blame on iuse but github struggles with the large file.

#57466

PR #63223, opened 2023-01-21
add large_tent_equivalent component group

Brambor:
#### Summary
None

#### Purpose of change

Story: _I wanted to add damaged_tent to components for constructing basecamp. Poking around I found that I can deconstruct them to obtain resources to build other tents, so I no longer need it._

While poking around I decided to make a component LIST / group to make such changes easier in the future.

#### Describe the solution
Refactor. Pull tent numbers into a single component group.

Delete branch. Add the damaged_tent to the list. See Aditional context.

#### Testing

Loaded the game, and looked at the base camp tent building, It looked the same.

#### Additional context

According to a discussion on discord, 2 damaged tents could be used as well.
In my opinion, materials obtained in deconstruction should be taken into account. And adjusted if they don't match.
image

And how is it possible that it takes the NPC **3 hours** to put up a tent?


Maleclypse (approved):

NetSysFire:
Why close?

Brambor:
My bad

PR #63227, opened 2023-01-21
Jsonifying of ladder use actions

Karol1223:
#### Summary
None 

#### Purpose of change
Stepladder used hardcoded use actions when the same could be achieved with the json alternatives. No point for that.

#### Describe the solution
Removes the LADDER use action and replaces it in stepladders with the use actions to deploy furniture. It should work exactly the same.

#### Additional context
~~I am only capable of doing json work and as such there's still c++ dead code leftover after this. I'm doing this PR as removing dead code should be a minute of work for someone capable of it and hopefully fixing this part would incentivize the removal of dead code.~~ Huge thanks to @cake-pie for removing the dead c++ code straight away! I'll look into how to set you as the co-author of the PR in a bit since I do believe you've done the harder part of the work here. EDIT: I have no idea how to do that. I thought it'd be simple and yet again GitHub manages to make this unnecessarily complicated. Apologies.

### DIFF
--- a/src/iuse.cpp
+++ b/src/iuse.cpp
@@ -239,8 +239,6 @@ static const efftype_id effect_weed_high( "weed_high" );
 
 static const flag_id json_flag_POWER_CORD( "POWER_CORD" );
 
-static const furn_str_id furn_f_aluminum_stepladder( "f_aluminum_stepladder" );
-static const furn_str_id furn_f_ladder( "f_ladder" );
 static const furn_str_id furn_f_translocator_buoy( "f_translocator_buoy" );
 
 static const harvest_drop_type_id harvest_drop_blood( "blood" );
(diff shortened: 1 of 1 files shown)


Night-Pryanik:
Please at least add a to-do comment in the code to remove this part later.

cake-pie:
> Please at least add a to-do comment in the code to remove this part later.

Dead code removal at Karol1223#1 to be merged into this PR.

cake-pie:
@Karol1223 please delete four more lines that I missed earlier:

https://github.com/CleverRaven/Cataclysm-DDA/blob/8c7e8194c8abee9da582a45dc1840fe15e414ddd/src/iuse.cpp#L242-L243

https://github.com/CleverRaven/Cataclysm-DDA/blob/8c7e8194c8abee9da582a45dc1840fe15e414ddd/src/iuse.cpp#L253

https://github.com/CleverRaven/Cataclysm-DDA/blob/8c7e8194c8abee9da582a45dc1840fe15e414ddd/src/iuse.cpp#L309

These were reported by the [Clang-tidy check](https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/3978991317/jobs/6821224722#step:7:1837) which I don't have locally. I'd use "suggest changes" functionality if I could but these lines are "too far away" from any of the preexisting changes for that to work so it's probably fastest if you just handled it directly.  
Thanks in advance.

> I have no idea how to do that.

I don't think you need to do anything? 8c22c8e63652ad240857b41e463b47602573ab0e is already authored by me.

Karol1223:
Alright good and done. I noticed the clang tidy reporting them but I was worried I may not be able to remove them from github website considering I've never tried touching cpp files before but it appears to have worked just fine.

> I don't think you need to do anything? https://github.com/CleverRaven/Cataclysm-DDA/commit/8c22c8e63652ad240857b41e463b47602573ab0e is already authored by me.

I also noticed that but it looks like the PR still only has me as its author. I hope it doesn't, though. At least in the commits it is still very visible that you authored that part of the PR.

PR #63230, opened 2023-01-22
Hide vehicle stats on sidebar unless relevant

Vaskritaya:
#### Summary
Interface "Sidebar vehicle stats are now hidden until relevant"

#### Purpose of change
The vast majority of the time, the player is spending time running around doing things on foot. If you have the vehicle stats enabled on your sidebar, they take up space doing nothing during this time. One could manually toggle the widget on and off as necessary, but that would be pretty silly.

#### Describe the solution
The sidebar UI already had a flag for hiding elements that were empty, such as the radiation badge entry if you lacked a badge. This PR just adds that flag to the relevant ui elements, causing them to be hidden unless they are filled. As an added bonus, this also keeps the pointless fuel meter hidden on fuelless vehicles.

None, this is very straightforwards

#### Testing
Spawned in. No vehicle stats in sidebar. Spawned bike, stepped on and off to see sidebar change. Same for a cube van. This was confirmed to work for both the labeled variants and the unlabeled variants.

### DIFF
--- a/data/json/ui/vehicle.json
+++ b/data/json/ui/vehicle.json
@@ -30,19 +30,19 @@
     "id": "vehicle_azimuth_desc_no_label",
     "type": "widget",
     "copy-from": "vehicle_azimuth_desc_label",
-    "flags": [ "W_LABEL_NONE", "W_DISABLED_WHEN_EMPTY" ]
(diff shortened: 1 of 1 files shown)


alp7292:
awesome change

NetSysFire (approved):

PR #63233, opened 2023-01-22
map/item: add only active contents to active cache

andrei8l:
#### Summary
None

#### Purpose of change
All contents of an item are processed as active items if any one of them is active

Follow up from #62916

#### Describe the solution
Partially revert #62916
When adding or removing items to/from the map (or vehicles), loop through contents and only cache the actually active items
Prune the active item and active submap caches while processing items instead of calling `active_item_cache::remove()` all over the place.

Items on characters still follow the old logic.

<details>
<summary> TODO (finished) </summary>

- [x] figure out how to handle `item::unseal()`: added pocket chain to the cache and reverted to calculating spoil modifier on the spot

</details>

N/A

#### Testing
The existing tests still pass, especially `inactive_container_with_active_contents` (from #62916) which targets this case.

New test unit for items rotting away in containers or (directly) on the map.

[Strathcona.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10473960/Strathcona.zip) In this save, there are 484 drums, each of which contains 599 rags and one active towel. Performance difference can be very clearly seen by waiting one hour. Verify that the towels are dry after one hour.

<details>
<summary>Profiling before</summary>

before

</details>

<details>
<summary>Profiling after</summary>

after

</details>

[Storden-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10713038/Storden-trimmed.tar.gz) This is a more realistic scenario created by starting the `Overrun` scenario and killing all monsters. Active items are on average only one container deep and don't share pockets with inactive items. We should still get a modest improvement due to the inactive containers not getting processed anymore.

<details>
<summary>debug timer & profiling before</summary>

before

profiling before

</details>

<details>
<summary>debug timer & profiling after</summary>

after

profiling after

</details>

#### Additional context
Fris0uman you might be interested in this

### DIFF
--- a/src/active_item_cache.cpp
+++ b/src/active_item_cache.cpp
@@ -19,17 +19,21 @@ void _remove_if( std::list<item_reference> &active_items, item const *it )
 
 } // namespace
 
+float item_reference::spoil_multiplier()
+{
+    return std::accumulate(
+               pocket_chain.begin(), pocket_chain.end(), 1.0F,
+    []( float a, item_pocket const * pk ) {
+        return a * pk->spoil_multiplier();
+    } );
+}
+
 void active_item_cache::remove( const item *it )
 {
-    int const ps = it->processing_speed();
-    if( ps == item::NO_PROCESSING ) {
-        // this item might have contained an active item so remove it from all queues
-        for( decltype( active_items )::value_type &al : active_items ) {
-            _remove_if( al.second, it );
-        }
-    } else {
-        _remove_if( active_items[ps], it );
+    for( item const *iter : it->all_items_ptr() ) {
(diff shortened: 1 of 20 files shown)


andrei8l:
I'm having some second thoughts here. There are many cases where the list of active submaps needs to be updated (such as the ubiquitous `visitable::remove_item()`). AFAICT there's no harm in simply [deleting](https://github.com/CleverRaven/Cataclysm-DDA/pull/63233/commits/9823fd648a76c2b5d76e78466e9f8478c132f6d4) the submaps from the list when they turn out to be empty, but I'll have to investigate more.

Let's keep this for after 0.G and maybe @jbytheway or @Hirmuolio can chime in

PR #34184, opened 2019-09-22
basecamps: Basecamps anywhere

mlangsdorf:
### Summary
```SUMMARY: Features "basecamps: Basecamps anywhere"```

#### Purpose of change
Add support for #34181 and allow faction camps to be built in any z-level location that has a recipe group, not just fields.

#### Describe the solution
Adjust recipe_groups to match terrains by the root terrain id, not including any directional suffix.

Add a new keyword to the at_om_location dialogue condition, FACTION_CAMP_START, which returns true if there is a faction camp recipe group that matches the om terrain id.

Adjust start_camp() to not fail if there are less than 4 fields adjacent to the camp.  Also, remove the warning if there aren't any swamps nearby because that warning was irrelevant and confusing.

Adjust the faction camp start response to check the NPC position and FACTION_CAMP_START instead of the player position and if the player was in a field.

Remove the Old Camp from the list of options; no one should be starting those anymore.

#### Additional context
Tested by applying #34181 and confirming that I could start a camp in a fire station.  I also confirmed that I still couldn't start a camp in a ranch camp field (which is ranch_camp_#, not field) and could still start a camp in an actual field.

Merging this is a prerequesite for testing and merging #34181

PR #29807, opened 2019-04-22
Delete "take over the camp" from NPC

kwl01skz:
The camp manager has been replaced by the bulletin board and two-way radio.
If you accidentally choose this conversation, it will cause the NPC to do something weird.

Regardless of how the faction camp will develop in the future, it is always reasonable to temporarily hide this dialogue.

#### Summary

SUMMARY: Bugfixes "Delete "take over the camp" from NPC."

#### Purpose of change

When doing this talk with the NPC, the NPC will run around and cannot return to the team.

#### Describe the solution

hide this dialogue.
#### Describe alternatives you've considered

Clean up all the code about this

PR #28742, opened 2019-03-15
Bulletin board to control faction camps, camp overseer resigns.

ghost:
#### Summary
```SUMMARY: Features "Bulletin board to control faction camps, camp overseer resigns."```

#### Purpose of change
As @mlangsdorf has requested, in his quest to entirely kill the camp overseer and then stomp on his remains.
The bulletin board will now be the place to look at camp missions and send companions on them.

#### Describe the solution
Re-enabled iexamine for bulletin board, made that display camp information like the overseer currently does.
added a bulletin board to all camp upgrades
Added dialogue to camp overseer so that they will offer to resign, then the camp will upgrade to the same level camp but with a bulletin board.

#### Describe alternatives you've considered
N/A

#### Additional context
N/A

PR #45572, opened 2020-11-22
Base camp reorganization

PatrikLundell:
#### Summary
SUMMARY: Infrastructure "Reorganized base camp hub and expansions to support new versions in parallel"

#### Purpose of change
The purpose is to prepare the ground for the introduction of new base camp versions as well as new expansion versions.
This is part of PR #45560, which was considered too large to be practical to review as a whole. 

#### Describe the solution
- Untangled dependencies between different base camps, expansions, and versions of those. This is only possible to a minor extent, due to backwards compatibility issues, but it's possible to get new uses use structures with lesser inter dependencies than legacy ones.
- Little is visible for the base camp. The structure prepares for the introduction of new versions, though, which is why the single modular one has been pushed down a folder level into a version_1 folder.
- The changes are more visible for expansions. Like base camps, the existing expansions are pushed down one level in preparation for the introduction of additional versions.
- A yet completely unnecessary step with a zero time recipe has been introduced, where the player has to select the single available version for construction. This is in preparation for multiple versions, where this recipe would let the player select which version, potentially among may, to construct. The intention is that this level would also provide some kind of description of each variant to help guide the selection. See the next section for alternatives.
- It should also be mentioned that a small number of minor adjustments have been made to existing base expansions. An extra cooking token for the canteen cooking facility (2 stoves), a well token for the kitchen well, an east-west copy paste error, "forage"->"forge". There might be something else along these lines as well.
- Differs from PR #45560 in that the top level expansion palette used has been factored out in a cleaner implementation. Also, the Saltworks, which PR #45560 doesn't touch, has been modified to use the same preparation logic as all the other expansions.

#### Describe alternatives you've considered
- A very clear alternative for expansion version selection implementation is to use the same logic as for the base camp, i.e. to use the top level recipe group to list all alternatives. The main reason this option was not chosen is that the list isn't sorted in any way that makes sense to humans (probably the order in which the recipes happen to be stored), which means the player will get a very poor overview of the available options. There's also no support for a descriptive text along the line of the recipe selections further down (where there's one "recipe name" line and a separate description line presented along with the recipe requirements). Sorting such that variants of the same expansion would end up grouped together would be sufficient to change the balance in favor of this implementation and prompt an adjustment of the logic.

#### Testing
[...]

PR #63235, opened 2023-01-22
Terminating support for obsolete primitive field camp

cake-pie:
#### Summary
Content "Terminating support for obsolete primitive field camp"

#### Purpose of change
Since build 9665 (Sep 24, 2019) it was no longer possible to start a faction camp using the legacy "primitive field camp" (aka "Old Camp") design. For people using stable builds, this took effect as of 0.E.

The source code contains hardcoded special casing required in order to support save games that still have faction camps of this variety, carried over from before its removal, such as:

https://github.com/CleverRaven/Cataclysm-DDA/blob/27d6b46dcda2ce6e0e02f7fdb8927093c23650b8/src/basecamp.cpp#L294-L299

Now that 0.G has dropped, we are at two full releases since 0.E; time to consider ending support for these legacy camps.

This will additionally allow us to eliminate other legacy code that is lurking around from various changes in the implementation of faction camps during the early stages of development.

#### Describe the solution

- Remove hardcoded special casing from `basecamp::has_water()`
  - primitive field camp will no longer support NPCs automatically using its well(s) to quench their thirst.
  - `fbmh_well_north` special case is retained to continue support for any that were constructed between b9162 (Jun 16, 2019) and b9644 (Sep 20, 2019), since the "Modular Field Camp version 1" is still available and supported. This special case could be removed if that type of camp was obsoleted, or when `has_water()` is modified to actually check for presence of terrain/furniture that acts as a water source rather than relying on JSON `blueprint_provides`.
- Document the purpose of `blueprint_provides` `water_well` in `BASECAMP.md`
- Remove hardcoded special case handling for primitive field camp's "Gather Materials" mission, which required extra machinery to allow it to return gathering yield that varies according to the camp level 1~20.
- Remove hardcoded special case handling for `faction_base_camp_1` mission (command tent) not requiring food
  - the recipe is being deleted so no other special cases remain, but the machinery is already there, so JSONize the food requirement waiver with a `NO_FOOD_REQ` flag anyway, in case it comes in useful in future.
  - an alternative approach here would have been to strip out all code related to waiving the food requirement entirely so that it is no longer possible
- Remove machinery related to "recover camp" which assigns a new NPC as overseer after the previous one died.
  - obsoleted after overseer replaced by bulletin board  
in #28742
  - associated talk option was removed later on  
in #29807
- Remove talk option to "remove overseer" and related machinery
  - was implemented to allow players to free up NPC from obsoleted overseer role  
in #28742
  - players have had since b8878 (experimental) / 0.E (stable) to switch from overseer to bulletin board
  - also remove stray "become overseer" left over  
after #28826
- Delete obsolete `TALK_CAMP_OVERSEER` dialogues, no longer needed without overseers
  - used to be [set as default topic when NPC became overseer](https://github.com/CleverRaven/Cataclysm-DDA/pull/28826/files#diff-f0de892fdc66c1d8d3a18d71a9f8263c954b66a51c62f5ac024b84f295c68640L524)
- Confirmed that it was already no longer possible to start the primitive field camp nor add its expansions:
  - `faction_base_camp_0` was [removed](https://github.com/CleverRaven/Cataclysm-DDA/pull/34184/files#diff-0e654486a3862f60925666d1294178a9ae95f3ce9ea15d0b77b379047ec0efdfL8) from `all_faction_base_types`  
in #34184
  - `faction_base_blacksmith_0` was [removed](https://github.com/CleverRaven/Cataclysm-DDA/pull/34411/files#diff-0e654486a3862f60925666d1294178a9ae95f3ce9ea15d0b77b379047ec0efdfL19) from `all_faction_base_expansions`  
in #34411
  - `faction_base_kitchen_0` was [removed](https://github.com/CleverRaven/Cataclysm-DDA/pull/34496/files#diff-0e654486a3862f60925666d1294178a9ae95f3ce9ea15d0b77b379047ec0efdfL18) from `all_faction_base_expansions`  
in #34496
- Deletion of obsolete json entries / files:
  - base building recipes for `faction_base_camp_0` and its upgrades, and associated update/nested mapgens
  - base building recipes for `faction_base_blacksmith_0` and its upgrades, and associated update/nested mapgens
  - base building recipes for `faction_base_kitchen_0` and its upgrades, and associated update/nested mapgens
- Added npc dialog to inform players about the changes / upcoming removal of support

#### Future work
Continue checking for legacy code, dead code and obsolete JSON for removal.

Final step of end-of-life for the primitive field camp would be to have a savegame migration process that deletes any remaining camp/expansion of these types from the camp/expansion list and reverts the overmap ids that they occupy to `field`. This would enable culling of all associated `"type": "overmap_terrain"` json entries. Prefer to defer this for one more version so people have a chance to salvage resources and abandon camp "in-game" rather than be forcibly ejected.

Leave it as it is.

#### Testing / checking
It is highly recommended to review this PR commit-by-commit rather than try to digest all the changes at once.

- `has_water()`: code inspection.
- gather mission: check that gather mission not broken for other camp types. For primitive field camp, change is semi-visible to player; people who are very observant may notice that the yields have changed, but otherwise it still functions
- `NO_FOOD_REQ`: add the flag to a long mission, e.g. one of the week-long ones that set up an expansion. Should be able to dispatch companion even with very low food stocks in the camp.
- recover camp, remove overseer, overseer talk topic: code inspection; these are basically dead code
- json deletions: follow the links I provided to previous PRs to confirm that the base/expansions can no longer be started

#### Additional context
I originally misdiagnosed the hardcoded `blueprint_provides` in `basecamp::has_water()` as a lack of consistent jsonization. Thanks to @PatrikLundell for pointing out the error.  
PR was repurposed; original OP preserved below.

---

<details>
<summary>Click for original OP</summary>

#### Summary
Infrastructure "Expand and standardize jsonization of `basecamp::has_water()`"

#### Purpose of change
Spotted this:
https://github.com/CleverRaven/Cataclysm-DDA/blob/27d6b46dcda2ce6e0e02f7fdb8927093c23650b8/src/basecamp.cpp#L294-L299

It does not make sense if every time someone added a new basecamp expansion -- a primarily JSON task -- that provides a water supply directly to the camp that they would have to come and add their corresponding `blueprint_provides` to this growing list -- in C++ code.

#### Describe the solution
`water_well` is already used for this purpose in some of the basecamp expansion defs. Make that a standard thing, document it, and add it to those that don't already have it. Remove the redundant entries from the C++.

Have some other id for this, like `water_source`, but it seems that all the existing ones are indeed wells and I don't think it's worth splitting hairs over what the id should be even if some hypothetical expansion supplies water to the camp by some means other than a well.

#### Testing
Hang around in a camp with a water well expansion and ensure that NPCs there continue to correctly get a drink automatically when thirsty enough (thirst > 40); they will [say `camp_water_thanks` line](https://github.com/CleverRaven/Cataclysm-DDA/blob/27d6b46dcda2ce6e0e02f7fdb8927093c23650b8/src/npcmove.cpp#L3772-L3776) when they do this. Can also check in periodically and check that no NPCs are building up high thirst levels.

</details>


PatrikLundell:
My understanding is that the hard coded faction camp pieces are for the (now very) old hard coded faction camp that can't be generated (but was retained for save compatibility, I presume). Newer JSON faction camp usage relies on the water_well token instead, and so haven't been added as direct entries here.
Given that the hard coded faction camp wasn't present at all in 0.F (I don't know if it was removed in 0.E or even earlier than that), it's probably time to remove support for it.

Note that I'd want to see the gradual removal of various faction camp tokens in favor of scanning for and detection of functionality providing assets in the base camp to allow them to be available regardless of whether they're produced via faction camp blueprints or player (ordered) construction.

cake-pie:
Well, this change doesn't preclude termination of support for the old pieces and would actually facilitate it; one less thing to have to deal with in addition to just dropping them from JSON.

PatrikLundell:
Agree. I was trying to provide some context.

The basecamp.md file seems to retain outdated, and possibly erroneous, information. The chop shop has been removed, for instance. I'm not sure how correct the others are.
The added description for "water_well" doesn't actually say what the token does, only point to a function that JSON editors aren't supposed to need to know what it's for. What it does provide (from a JSON crafter perspective) is unlimited amounts of water for faction camp crafting and other usages of water (and it can be noted that the water isn't clean, although I don't know if companions are affected by unpurified water).

Changing the code in basecamp.cpp causes existing (if any) instances of the old hard coded basecamp to cease to function. Modifying the JSON for the construction of these won't retroactively add the new token to them; it would only make a started but not finished base camp capable of (finally) adding these constructions with retained functionality, which is an extremely unlikely niche case.

I think it's time to get rid of the primitive_field files in version 1 of the hub, i.e. the recipe and blueprint ones (the ones in the garage and farm version 1 expansions should remain: they're primitive but should remain functional). As far as I can tell from a brief look they can't be accessed unless you started a basecamp at some time before 0.F, and I believe one version is the span obsolete functionality is to be supported.

cake-pie:
Okay, thanks for the elaboration, that was helpful. Are you going to make a PR that ends support for the old v1 camp pieces? In which case I'm happy to close this as superseded. Otherwise please advise how you'd like to proceed.

PatrikLundell:
This isn't really version 1, but rather version 0. It ended up in version 1 when version 2 was introduced and all that existed previously was moved into version 1, including the support for the obsolete version 0.

I'd recommend the removal of
- .../data/json/mapgen/basecamps/base/modular_hub/version_1/primitive_field.json and
- .../data/json/recipes/basecamps/base/recipe_modular_hub/version_1/recipe_primitive_field.json (the file you edited)

plus the code simplification you've made. My checks haven't found the tokens used outside of these files, with the exception of "faction_base_camp_0".
This token is used in .../data/json/overmap/overmap_terrain/overmap_terrain_faction_base.json, but there it's used as the base from which camp symbols are inherited, so I don't think it would be a good idea to try to remove it from there. The token is also used in .../data/json/mapgen/faction_buildings.json. I think much of this file is obsolete, in which case the obsolete tokens in it ought to be possible to remove from overmap_terrain_faction_base.json. However, blueprints for existing expansions are in this file, so clearing out this file of obsolete entries requires searching each entry for usages and then removing the obsolete ones together with their obsolete overmap terrain entries.
There's also a section for hide sites and remote wall construction that's still in use.

As you see, it isn't trivial to figure out what should be removed.

cake-pie:
> Given that the hard coded faction camp wasn't present at all in 0.F (I don't know if it was removed in 0.E or even earlier than that), it's probably time to remove support for it.

I've tracked it down to b9665 experimental and 0.E stable; repurposing PR into end of support for the primitive field camp.

PatrikLundell:
I believe the fbmh_well_north support can be dropped as well, for two reasons:
- While the base camp itself remains supported, that doesn't mean every iteration of it has to be. When sufficiently long time has passed, support can be dropped.
- There is a scanning functionality in place that should pick up the presence of a well even when not explicitly added by a blueprint. It's primarily there to support the bare bones basecamp, but should cover that corner case as well. If it doesn't, there is something missing either in the scanning code or in the application of its results.

cake-pie:
> "faction_base_camp_0" [...] is used in .../data/json/overmap/overmap_terrain/overmap_terrain_faction_base.json, but there it's used as the base from which camp symbols are inherited, so I don't think it would be a good idea to try to remove it

Oh, in any case we don't want to remove any `overmap_terrain` entries yet, doing so would make it impossible to load any save files with those overmap terrain ids. That'll happen in the final stage of EOL process when we cease to support even existence of legacy camps -- either save files with such camps will outright fail to load properly, or the camps get "deleted" during save game migration, with their overmap ids being migrated to something else (probably bare field). I think we should wait another version before forcibly removing any remaining camps in this manner.

For now, deletion of relevant basebuilding recipes and the related update/nested/regular mapgens would halt any further upgrades and expansions. Along with the C++ change that will cease support for automatic watering of followers, this should signal to anyone still playing on a save file that has been migrated from that long ago that it is time to abandon their legacy camps in favor of more modern variants. We could also do a bit more to alert players of the change, like mentioning it in the faction camp tutorial.

> There is a scanning functionality in place that should pick up the presence of a well

For crafting perhaps, but definitely not going to work for automatic watering of followers [ [1](https://github.com/CleverRaven/Cataclysm-DDA/blob/f097346492f1d1ecfd3210520810bb4d1bf4debd/src/npcmove.cpp#L3772-L3776), [2](https://github.com/CleverRaven/Cataclysm-DDA/blob/f097346492f1d1ecfd3210520810bb4d1bf4debd/src/faction_camp.cpp#L3285-L3287) ]

---
In `BASECAMP.md` under ["adding basecamp expansions"](../blob/a7cc07a70bc0015161c7cdf5779aea0635efe79e/doc/BASECAMP.md#adding-basecamp-expansions) it states that "../json/mapgen/faction_buildings.json also has to be updated to introduce an entry for the new expansion." @PatrikLundell you added this statement in #45572, but is it actually true? It looks like since #29523 basecamp construction has used update mapgens, which in turn may refer to nested mapgens -- these are `"type": "mapgen"` with `update_mapgen_id` or `nested_mapgen_id` json property respectively. I suspect the basic mapgen (with `"om_terrain"`) are obsolete; plenty of expansions don't have corresponding entries in there either.

PatrikLundell:
I believe you're correct regarding the faction_buildings.json entry. I believe that was the only pattern in use at the time this was written, and if I remember correctly my attempts to use "normal" definitions for the starting definition failed to work for unknown reasons (I think it only works with "rows" and not point/line/square, and has to be exactly 24*24, although I can't swear on either being correct). Thus, I tried to mention all the places I've found you had to add stuff for it to work, as some of them weren't exactly intuitively obvious to find.
I believe later someone (me or somebody else) figured out how to get rid of that pesky entry and move the data to where it belongs, but without any thought of updating the documentation (or that it might affect documentation at all).

cake-pie:
There's obviously more to be done; I've been turning up various bits of legacy support code while looking through the camp-related code. I'll stop here for now and defer further work on this to a subsequent PRs so that this one doesn't get too big and unmanageable. (See "future work" section in OP).

cake-pie:
Build failure unrelated; ef490d7 was fine and 48308a4 doesn't touch C++ code.

PR #63247, opened 2023-01-22
Integrated armor double dipping penalty fix

Dan1ss1mo:
#### Summary
Bugfixes "Weird interaction between skintight clothing and thick natural armor"

#### Purpose of change
Currently wearing skintight clothing with mutations like thick scales and shaggy fur is way more uncomfortable then normal later clothing as penalty is double dipping. Both should probably recive equal penalties for overlapping with you're natural armor

#### Describe the solution
Remove "SKINTIGHT" flag from natural armor with "NORMAL" flag. You would still be unable to wear any clothing under it and recive penalties from both normal and skintight layer. The only difference it that it would be equil for both, rather then doubled for skintight.

Adding custom variants for mutants, that would be copies of skintight items but for normal layer. Imo that seems a bit redundant, as this kind of stuff should probably already be handled by refitting\XL stuff

#### Testing
Works as intended: don't get double dipping penalties for skintight stuff,recive normal penalties for normal and skintight layer, still can't put clothing under natural armor

#### Additional context
Our scaly and chitinous friends deserve some warmth, you know


Dan1ss1mo:
> Conflicts on this currently

Ah, my bad, was taking a break from cdda for a bit. Will fix that in a moment

PR #63252, opened 2023-01-23
Fix display of byproduct charges in crafting gui

### LINKED ISSUE #63142: Crafting GUI reports incorrect amount of byproducts (for byproducts with charges)
### Describe the bug

The crafting GUI reports kerosene has a diesel byproduct of 40. It actually has a byproduct of 40\***250** = 10,000. The 10,000 value is correct, but the GUI is off. This is due to my changes to `crafting_gui.cpp` in #62148. That's my bad.

`recipe::get_byproducts` returns a map containing the `itype_id` and an integer corresponding to the amount produced. The crafting GUI has no idea whether that integer is a `count` (the old `byproducts` field), or a `charges` (definable exclusively via `byproduct_group`), or something else altogether. The function likely needs to be expanded to pass that information along. I don't know a good way to do that.

### Steps to reproduce

Open crafting menu, look at kerosene. See it says a byproduct of 40 diesel instead of 10,000. Absolute lies!

### Expected behavior

The UI should show the correct values for produced items. If nothing else, my changes to crafting_gui.cpp should be reverted.

### Screenshots

Test recipe containing both a `byproducts` and `byproduct_group` producing equal amounts. The display for byproducts should be equal in number(10,000). They are not.

Untitled

OS: Windows
OS Version: 10
Game Version: https://github.com/CleverRaven/Cataclysm-DDA/commit/d6d5349c992826d17eee88822464e1d83aa6ae70
Graphics Version: Tiles
Game Language: System language []
Mods loaded: [
Dark Days Ahead [dda],
]

### Additional context

The crafting GUI has no idea how to deal with a `count` in a defined `byproduct_group` either, although the actual crafting functions seem to handle them fine.

mqrause:
#### Summary
None

#### Purpose of change

Fixes #63142

#### Describe the solution

`byproducts` and `byproduct_group` are counting their results differently. This aligns `byproducts` with `byproduct_group` for the purposes of the crafting gui.

Unify `byproducts` and `byproduct_group` and make it item groups only (>300 json entries would need adjustment).

#### Testing

Check recipes for `kerosene` (byproduct with charges) and `canvas sack` (byproduct without charges) and craft them to make sure both display and result is correct.

### DIFF
--- a/src/recipe.cpp
+++ b/src/recipe.cpp
@@ -677,7 +677,10 @@ std::map<itype_id, int> recipe::get_byproducts() const
 {
     std::map<itype_id, int> ret;
     if( !byproducts.empty() ) {
-        ret.insert( byproducts.begin(), byproducts.end() );
+        for( const std::pair<const itype_id, int> &byproduct : byproducts ) {
+            int count = byproduct.first->count_by_charges() ? byproduct.first->charges_default() : 1;
+            ret.emplace( byproduct.first, byproduct.second * count );
+        }
     }
     if( !!byproduct_group ) {
         std::vector<item> tmp = item_group::items_from( *byproduct_group );


RenechCDDA:
Thank you for cleaning this up, it's hard for me to work on these things sometimes. I pulled your branch and just doubly confirmed it works, here's some pictures of the GUI display.

<details>
<summary>Images of crafts</summary>
Byproduct *group* with charges:

image
Byproduct with charges:

image
byproduct without charges:

image

</details>

PR #63256, opened 2023-01-23
Fix #63204: Make proficiency account for SKILL_TRAINING_SPEED

### LINKED ISSUE #63204: skill training speed should also affect proficiency xp

skill training speed affects xp from both training and books but not proficiency.  with  skill training speed either 0 or 100 multiplier proficiency xp gain is same

### Solution you would like.

skill training speed should also affect proficiency xp 

maybe add a seperate profeciency xp gain multiplier on debug settings

### Additional context

_No response_

Kromgart:
#### Summary
Bugfixes "Make proficiency account for SKILL_TRAINING_SPEED"

#### Purpose of change

Fix #63204 

#### Describe the solution

Proficiency has a field _time_to_learn. I checked the code and seems like this field is accessed only when creating proficiency from json, and then read only through corresponding getter time_to_learn(). I modified the getter, so that the value is adjusted accordingly to the SKILL_TRAINING_SPEED game option.

#### Testing

New test char. Made 5 Z corpses, dissected them with scalpel, one time with default SKILL_TRAINING_SPEED (1.0), and then with modified (0.2). Numbers seem to respect the option now, see screenshots.

SKILL_TRAINING_SPEED 0.2:

prof1a

prof1b

SKILL_TRAINING_SPEED 1.0:

prof2a

prof2b

#### Additional context

UPD: fixed screenshots order

### DIFF
--- a/src/proficiency.cpp
+++ b/src/proficiency.cpp
@@ -180,7 +180,7 @@ float proficiency::default_weakpoint_penalty() const
 time_duration proficiency::time_to_learn() const
 {
     const double scaling = get_option<float>( "SKILL_TRAINING_SPEED" );
-    if ( scaling != 1.0 && scaling != 0.0 ) {
+    if( scaling != 1.0 && scaling != 0.0 ) {
         return _time_to_learn / scaling;
     } else {
         return _time_to_learn;
--- a/src/proficiency.cpp
+++ b/src/proficiency.cpp
@@ -129,7 +129,12 @@ const std::vector<proficiency_category> &proficiency_category::get_all()
 
 bool proficiency::can_learn() const
 {
-    return _can_learn;
+    if( _can_learn ) {
+        const double scaling = get_option<float>( "SKILL_TRAINING_SPEED" );
+        return scaling != 0.0;
+    } else {
+        return false;
+    }
 }
 
 bool proficiency::ignore_focus() const
--- a/src/proficiency.cpp
+++ b/src/proficiency.cpp
@@ -179,7 +179,12 @@ float proficiency::default_weakpoint_penalty() const
 
 time_duration proficiency::time_to_learn() const
 {
-    return _time_to_learn;
+    const double scaling = get_option<float>( "SKILL_TRAINING_SPEED" );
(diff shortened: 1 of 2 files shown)


alp7292:
that was quick thanks for your efforts

Kromgart:
> That was quick thanks for your efforts does this also works with bigger numbers like 2.0 or 5.0. Skill training speed can be be between 0.0 to 100.0

Yes, prof's _time_to_learn will become smaller (less hours needed) if you use bigger number.

src/proficiency.cpp
```diff
@@ -179,7 +179,12 @@ float proficiency::default_weakpoint_penalty() const
 
 time_duration proficiency::time_to_learn() const
 {
-    return _time_to_learn;
+    const double scaling = get_option<float>( "SKILL_TRAINING_SPEED" );
+    if( scaling != 1.0 && scaling != 0.0 ) {
```
Fris0uman, line 183: Is the 0.0 option also ignored in the other uses of the setting? Shouldn't 0 mean no learning?

intergalacticdata:
Makes sense to me.

alp7292:
if we remove the "&& scaling != 0.0" would that work

alp7292:
if we remove the "&& scaling != 0.0" would that work

src/proficiency.cpp
alp7292: if we remove the "&& scaling != 0.0" would that work

alp7292: yes 0 should mean no learning it shouldnt ignored

Kromgart: Fixed with the next commit:  when the option is 0.0 there is no progress on proficiency anymore and the UI is more consistent.

src/proficiency.h
```diff
@@ -12,6 +12,7 @@
 #include "color.h"
 #include "flat_set.h"
 #include "optional.h"
+#include "options.h"
```
irwiss, line 15: If the stuff from options.h is only used in the cpp file it's best to move it there, so it doesn't get included everywhere proficiency.h is included

alp7292:
now it fails great

PR #63258, opened 2023-01-23
The great oversizing

Dan1ss1mo:
#### Summary
Bugfixes "Most belts, belted pouches and pockets should be oversize"

#### Purpose of change
Currently a lot of  belted pouches and pockets don't have oversize flag, which is weird, as there is no mutations i can think of that would conflict with them, unless it's remove the limb in question entierly. Some stuff like armband pouches are dropleg pouches even have varsize flag, which doesn't make any sense, as they are litteraly nothing but pockets with ADJUSTABLE straps. Same goes for belts, which don't have oversized flag at all. Just tighten\loose them, lmao. What make it even more weird, is that some stuff do have oversize as it supposed to.

#### Describe the solution
So, for the sake of consistency and common sense, I think it is save to assume that belts, pouches and pockets should be oversized. 

#### Testing
Everything seems to for fine, feel free to point out stuff that i may miss tho

#### Additional context
I deliberately avoided various bigger torso rigs, as they may indeed get in a way of some mutated appendages without more extensive modifications

### DIFF
--- a/data/json/items/armor/bespoke_armor/utility.json
+++ b/data/json/items/armor/bespoke_armor/utility.json
@@ -84,7 +84,7 @@
       }
     ],
     "use_action": { "type": "holster", "holster_prompt": "Store item", "holster_msg": "You put your %1$s in your %2$s" },
-    "flags": [ "WATER_FRIENDLY", "STURDY", "BELTED" ],
+    "flags": [ "WATER_FRIENDLY", "STURDY", "BELTED", "OVERSIZE" ],
     "armor": [ { "encumbrance": [ 2, 6 ], "coverage": 100, "covers": [ "torso" ], "specifically_covers": [ "torso_waist" ] } ]
   },
   {


PatrikLundell:
If you keep expanding the waist line you'll eventually find that you'll run out of belt to make new holes in, and similar considerations would affect things tied to limbs.
The straps of a leg pouch are probably not going to be long enough to actually be usable by super humanly large mutants, or, for that matter, obese people with huge amount of padding on their thighs. After all, you don't want long pieces of strap flapping around uselessly (and possibly getting in the way or snagging on things).

Now CDDA doesn't deal with ordinary clothing in various sizes, so this issue might be hand waved away as being on a scale that's not implemented.

Dan1ss1mo:
> If you keep expanding the waist line you'll eventually find that you'll run out of belt to make new holes in, and similar considerations would affect things tied to limbs. The straps of a leg pouch are probably not going to be long enough to actually be usable by super humanly large mutants, or, for that matter, obese people with huge amount of padding on their thighs. After all, you don't want long pieces of strap flapping around uselessly (and possibly getting in the way or snagging on things).
> 
> Now CDDA doesn't deal with ordinary clothing in various sizes, so this issue might be hand waved away as being on a scale that's not implemented.

According to the volume schanges for XL stuff, even huge mutants aren't that bigger then humans, and straps on most of this stuff tend to be retty damn long

Issue #62721, opened 2022-12-13
Make "5.56x45mm/.300BLK 10-round clip" useful. Allow speedloaders with non-empty magazines.


Zanthra:
### Describe the bug

If a magazine has any bullets loaded in it the game will exclude speedloaders from the ammunition avaialble to reload the magazine.

With common 30 round STANAG magazines, this means you can load only the first 10 rounds into the magazine using a speedloader.

Additionally, ammunition contained inside another magazine is also hidden from available ammunition. So any further 10 round clips are not available, and neither is the ammo inside them. To load the remaining 20 rounds into the magazine, you first have to unload the 10 round clips, then load the bullets into the magazine.

To fix this, the function item_pocket::can_reload_with in src/item_pocket.cpp should be changed to allow any speedloader that has ammo less than or equal to the remaining capacity in the magazine. This is the relevant check.

https://github.com/CleverRaven/Cataclysm-DDA/blob/2594fd915cfb8ce51066095790be037beb801b83/src/item_pocket.cpp#L1479-L1487

### Steps to reproduce

1. Have an empty STANAG magazine, and 3 "5.56x45mm/.300BLK 10-round clip"s with compatible ammo in them.
2. Reload the magazine, and observe the first reload gives the clips as valid reload options.
3. Attempt to load the next 10 bullets into the magazine, and observe that you have no valid ammo to reload the magazine with.

### Expected behavior

After loading the first 10 rounds into the magazine from the clip, I would expect to be able to load the next 10 rounds into the magazine from another clip.

### Versions and configuration

cdda-windows-tiles-sounds-x64-msvc-2022-12-02-0627

GuardianDll:
That's a good info about how to fix it, but i think its better to make a separate flag - you definitely can put few rounds from 556 clip or shotgun speedloader, but not from revolver one
@catdach you may be interested due the #58580

Zanthra:
I think that checking that the ammo in the speedloader is less than or equal to remaining ammo capacity in magazine is reasonable in that case, even for revolver or shotgun speedloaders.

Alternatively you can check the max ammo capacity of the speedloader against the remaining capacity in the magazine. These checks should not negatively affect current speedloaders (and could allow half revolver speedloaders).

The downside is that with the 5.56 speedloaders you cannot reload some ammo from the speedloader clip if the magazine does not have a multiple of 10 capacity remaining.

Zanthra:
[...]

PR #63264, opened 2023-01-24
Fixed 5.56 clips bug and fix revolver reloading

CutenessOverloaded:
#### Summary
Bugfixes "Fixed 5.56 clips bug and fix revolver reloading"
Closes https://github.com/CleverRaven/Cataclysm-DDA/issues/62721

#### Purpose of change
**For the clip:**
Yesterday in my run, I'm wondering why I can't reload a **"STANAG 30-round magazine"** with **"5.56x45mm/.300BLK 10-round clip"** if there has already some ammo inside the mag. Then, I wanna fix it. So, I don't have to unload & reload again and again.

**For the revolver:**
I accidently made it. ._.

#### Describe the solution
**For the clip:**
The bug is from **"SPEEDLOADER"** flag. In original (I guess), this flag is used for something like **revolver** that they don't have a magazine to reload quickly. So, the speed loader with the flag can allow these guns to reload faster. And because you can't really use a speed loader if there are still some ammo inside the gun. So, the codes have to make sure there are no longer ammo inside the gun, then reload.
But if the gun or magazine (e.g. **"STANAG 30-round magazine"**) has more capacity then the amount of ammo the speed loader with **"SPEEDLOADER"** flag (e.g. **"5.56x45mm/.300BLK 10-round clip"**) can reload, the **speed loader** can't reload the gun or magazine if there are still some ammo in it.
For example, you can reload a **empty "STANAG 30-round magazine"** with a **full "5.56x45mm/.300BLK 10-round clip"**, it loads 10 ammo in the magazine. When you try to reload the magazine with another 10-round clip, you can't but in real life you can.
So, I created a new flag **"SPEEDLOADER_CLIP"** to fix the problem. It let clips to reload gun or magazine that still have some ammo inside.

**For the revolver:**
Speed loaders which have **"SPEEDLOADER"** flag can now reload magazine or gun that have enough space to content the amount of ammo that the speed loader have.

**For the clip:**
Maybe make the **"SPEEDLOADER"** just like **"SPEEDLOADER_CLIP"** and add a new flag for guns like revolver that if you still want to reload the gun with the speed loader even though there were still some ammo inside, you will first drop all ammo on the ground then reload.

#### Testing
**For the clip:**
Before the fix, I took a **empty "STANAG 30-round magazine"** and successfully reload it with 1 **full "5.56x45mm/.300BLK 10-round clip"** but I can't reload it with the rest of the clips.
After the fix, I took a **empty "STANAG 30-round magazine"** and successfully reload it with 3 **full "5.56x45mm/.300BLK 10-round clip"**.

**For the revolver:**
Before, **5 round revolver** can reload with **speed loader** only if the revolver was emptied.
After, **5 round revolver** that have 3 remaining ammo capacity can reload with **speed loader** that have 3 or less ammo

#### Additional context
**"5 round revolver"** vvv
Screenshot (93)

Screenshot (94)

### DIFF
--- a/src/item_pocket.cpp
+++ b/src/item_pocket.cpp
@@ -1479,8 +1479,8 @@ bool item_pocket::can_reload_with( const item &ammo, const bool now ) const
         // The ammo in it needs to be compatible,
         // and the opcket needs to have enough space (except casings)
         return allows_speedloader( ammo.typeId() ) &&
-                is_compatible( ammo.loaded_ammo() ).success() &&
-                ( remaining_ammo_capacity( ammo.loaded_ammo().ammo_type() ) >= ammo.ammo_remaining() );
+               is_compatible( ammo.loaded_ammo() ).success() &&
+               ( remaining_ammo_capacity( ammo.loaded_ammo().ammo_type() ) >= ammo.ammo_remaining() );
     }
 
     if( ammo.has_flag( flag_SPEEDLOADER_CLIP ) ) {


TheShadowFerret:
Add "Closes #62721" to your summary, and that will automatically close that issue on merge of this PR.

GuardianDll:
Great work!
You can also replace the SPEEDLOADER flag with SPEEDLOADER_CLIP for shotguns, as they have similar problem - you can't reload them if they are not empty (syringe speedloader we emulate can do this irl)

data/json/flags.json
```diff
@@ -1689,6 +1689,10 @@
     "id": "SPEEDLOADER",
     "type": "json_flag"
   },
+  {
+    "id": "SPEEDLOADER_CLIP",
+    "type": "json_flag"
```
irwiss, line 1694: Might be worth adding `"info"` entry here to inform player e.g.; "speedloader_clip can be used to quickly reload magazines yadda yadda..." (without getting into clip vs magazine arguments 😂 ), although we're now in string freeze so that might be too annoying as that'd postpone the merge

Or update the docs https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_FLAGS.md#magazines (or both)

Dan1ss1mo:
Seems to work as intended. Testsed with 30 stanag and 3 clips, and stanag that filled woth 25 out of 30 rounds and 1 full clip.

CutenessOverloaded:
Thinking about to spread the "SPEEDLOADER" and "SPEEDLOADER_CLIP" into two PR.
Since, I think "SPEEDLOADER" should be enough and no need to have "SPEEDLOADER_CLIP".
Maybe I remove the "SPEEDLOADER_CLIP" and keep the **new** "SPEEDLOADER" for now.
I will work on it maybe 9 hours later. *Zzz...Zzz...*

CutenessOverloaded:
12 hours after, I made two pr which spread this pr into two ([About "SPEEDLOADER"](https://github.com/CleverRaven/Cataclysm-DDA/pull/63287) and [About "SPEEDLOADER_CLIP"](https://github.com/CleverRaven/Cataclysm-DDA/pull/63288)). So, now I am closing this pr.

PR #63272, opened 2023-01-24
Remove "vacuum pump" req from recipe for makeshift vacuum mold

crowder:
From a discussion with Aerin on Discord, the vacuum pump is large, some 22KG and is essentially a piece of heavy equipment not a part salvaged from a home vacuum. The makeshift vacuum mold is comparitively much smaller at 4.4 lbs or so.

It's possible that some "small vacuum pump" part should be added to the game, which could be salvaged from (say) the vacuum sealer tool.

As an interim stopgap, this PR simply removes the vacuum pump part from the recipe for the makeshift vacuum mold.
A link to the discussion in Discord:

https://discordapp.com/channels/598523535169945603/642349261320880128/1067531784687988736

Removing the "vacuum pump" requirement from the recipe for the "makeshift vacuum mold".

#### Summary
Bugfixes "Remove `vacuum pump` req from recipe for makeshift vacuum mold"

#### Purpose of change
The vacuum pump unit itself is apparently meant to be a piece of industrial machinery, not a salvaged pump from a sealer or home vacuum. This fixes that dependency.

#### Describe the solution
Simply removes the current dependency.

It's possible that a `small vacuum pump` part should be added to the game to replace this dependency.

#### Testing
Tested by deploying the same JSON files into my game directory in the correct locations and verifying that the recipe is altered as described.

### DIFF
--- a/data/json/recipes/tools/tools_electronic.json
+++ b/data/json/recipes/tools/tools_electronic.json
@@ -729,6 +729,7 @@
     "components": [
       [ [ "2x4", 4 ] ],
       [ [ "plastic_chunk", 2 ] ],
+      [ [ "motor_micro", 1 ] ],
       [ [ "nail", 16 ] ],
       [ [ "wood_panel", 1 ] ],
       [ [ "hose", 1 ] ],
--- a/data/mods/innawood/recipes/tools_electronic.json
+++ b/data/mods/innawood/recipes/tools_electronic.json
@@ -190,6 +190,7 @@
     "components": [
       [ [ "2x4", 4 ] ],
       [ [ "plastic_chunk", 2 ] ],
+      [ [ "motor_micro", 1 ] ],
       [ [ "nail", 16 ] ],
       [ [ "wood_panel", 1 ] ],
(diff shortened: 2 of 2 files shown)


Maleclypse (changes requested):
You'll need to add the part or figure out an existing in game part that won't make the recipe magic.

crowder:
Would a `micro motor` possibly be a reasonable replacement?

crowder:
I notice that the `makeshift vacuum sealer` recipe uses a `micro motor`, perhaps that would be reasonable here?

Maleclypse:
> I notice that the `makeshift vacuum sealer` recipe uses a `micro motor`, perhaps that would be reasonable here?

Perfect, I think. Fans deconstruct into micromotors so I think it works.

Maleclypse (approved):
Thanks for catching this!

crowder:
Are there steps I need to take to make this mergable?

Maleclypse:
> Are there steps I need to take to make this mergable?

I've approved it and it is currently mergeable.  Our mergers are very busy irl and this close to 0.G merging slows down.  I'll make a note this is ready for merge whenever one is available to do so.

PR #63278, opened 2023-01-25
Blocking fix

### LINKED ISSUE #63271: "prevent_weapon_blocking": true also prevent leg and arm block
### Describe the bug

Martial arts with "prevent_weapon_blocking": true flag make charecter to be unable to block at all, even while unarmed

### Steps to reproduce

1. Set melee and unarmed to 5 or more
2. learn Taekwondo
3. let somthing hit you for a while
4. Observe that you don't block anything

### Expected behavior

You should be able to block with legs and arms

### Screenshots

_No response_

Windows, no mods, d6d5349 experimental 
[Unarmed_test.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10493001/Unarmed_test.zip)

### Additional context

removing "prevent_weapon_blocking": true seems to fix the issue

Dan1ss1mo:
#### Summary
Bugfixes "prevent_weapon_blocking": true prevent blocking entierly"

#### Purpose of change
Taekwondo and forced unarmed currenly copletly disable charecter ability to block attacks.

#### Describe the solution
Shuffle allow_weapon_blocking check to be performed earlier as Venera3 suggested

#### Testing
Seems to work for me. Was able to blok with my limbs with Taekwondo both armed and unarmed, blocked with my weapom then style allow it, no blocks with Sōjutsu while unarmed, as expected

#### Additional context
fix #63271, thanks to @Termineitor244 for helping me to confirm the bug and @Venera3 for proposing better solution

### DIFF
--- a/src/melee.cpp
+++ b/src/melee.cpp
@@ -2207,7 +2207,7 @@ bool Character::block_hit( Creature *source, bodypart_id &bp_hit, damage_instanc
 
     /** @ARM_STR increases attack blocking effectiveness with a limb or worn/wielded item */
     /** @EFFECT_UNARMED increases attack blocking effectiveness with a limb or worn item */
-    if( unarmed || force_unarmed || worn_shield ) {
+    if( unarmed || force_unarmed || worn_shield || ( has_shield && !allow_weapon_blocking ) ) {
         arm_block = martial_arts_data->can_arm_block( *this );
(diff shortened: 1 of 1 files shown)


Venera3 (approved):
Shunts disallowed  weapon blocks into the unarmed/shield block route from the start, looks good.

PR #63285, opened 2023-01-25
Refactor unload and sort loot activity

Willenbrink:
#### Summary
None

#### Purpose of change
When sorting loot, the character annoyingly walks to an item even if it cannot be sorted (e.g. due to being a favorite).

#### Describe the solution
Refactor the loot moving activities. Specifically: `unload_loot_activity_actor` and `move_loot_activity_actor`. I found the previous code quite hard to read, therefore the refactoring changes them basically completely.

The code now works as follows:
- `loot_start`: Build a list of all interesting tiles. Filter them according to some basic rules (e.g. tile is ignored, on fire or inaccessible).
- `loot_unload_item`: Try to unload a specific item. Check if it can be unloaded according to all the complicated rules. If we can unload it and are not adjacent, move to the tile and return. Due to the way activity actors work, this will move the character to the tile where it will again attempt to unload the item, this time succeeding.
- `loot_iterate_tiles`: Iterate over all the interesting tiles and apply a `process_item` function (passed as argument) to every tile.
- Implement both `unload_loot_activity_actor` and `move_loot_activity_actor` using the functions above.

- Doing a minimal change to only fix the problem with walking to a tile without needing to.
- Refactoring a whole lot more. `activity_item_handling.cpp` is apparently completely full of similarly hard-to-read code and does not use the activity actor system. I wasn't motivated enough for that (:

#### Testing
The tests run, playing with the PR myself works as intended.

Haven't checked if the NPC behaviour is still completely correct as it previously did some weird stuff with reloading the NPCs. Hopefully I've reproduced the same behaviour.

#### Possible problems
- I do not know how `moves_left` really works. Should that be an accurate estimate? I set it to the number of tiles remaining, i.e. a reasonable approximation when compared to the total number of tiles inspected.
- I changed the fields in the activity actor. This also changes the serialization. Is that problematic? What happens if a json does not have the relevant field?
- I'm also not sure how activity ids work. I.e. `ACT_MOVE_LOOT` vs `player_activity( move_loot_activity_actor() ) )`
`


Willenbrink:
Did I just mess up my PR? I'm confused what just happened.

Willenbrink:
So I've got no clue how that happened but it should be fixed now. Somehow all recent commits from master got merged into my branch instead of rebasing my branch onto master. Sorry for pinging unrelated people :)

mqrause:
> * I do not know how `moves_left` really works. Should that be an accurate estimate? I set it to the number of tiles remaining, i.e. a reasonable approximation when compared to the total number of tiles inspected.

That depends on the activity definition in json, specifically the `based_on` property. If it's `speed` or `time`, it expects the accurate time because it automatically decreases `moves_left`. If it's `neither` you don't need to set it at all. Activities loop over `activity_actor::do_turn` until `moves_left` hits 0 at which point `activity_actor::finish` is called, or until the activity is cancelled.
I didn't really review this, yet, but chances are you don't need `moves_left` at all except for maybe setting it to 0 once all the tasks are done.

> * I changed the fields in the activity actor. This also changes the serialization. Is that problematic? What happens if a json does not have the relevant field?

You should test that by creating autosaves during these activities before your changes and loading them with your changes. But in general only the things you put in the deserialization functions get loaded, so chances are you're losing some data and can't process the old activity state. Best case the activity just stops, worst case it causes a crash.

> * I'm also not sure how activity ids work. I.e. `ACT_MOVE_LOOT` vs `player_activity( move_loot_activity_actor() ) )`

It's just that, an identifier. Using the activity actor constructor is preferred because it allows you a safer way to pass on data as opposed to constructing the activity first and then setting the data on that. The constructor using the id is just a legacy constructor to support the not yet migrated activities.

PatrikLundell:
As indicated, changes have to work with existing saves, so you may have to introduce code that detects the previous format and transforms it into the new one when reading a save. If such code is needed, I'd recommend the addition of a comment that indicates when the change was made (I'd suggest both date and game version cycle) so that it can be removed when no longer relevant (typically a version cycle later) without having to try to trace it via the revision history.

src/activity_actor.cpp
```diff
@@ -6171,334 +6192,506 @@ static void move_item( Character &you, item &it, const int quantity, const tripo
...
-            // skip unpickable liquid
-            if( thisitem.made_of_from_type( phase_id::LIQUID ) ) {
-                ++num_processed;
-                continue;
-            }
+void unload_loot_activity_actor::finish( player_activity &act, Character &who )
+{
+    add_msg( m_info, _( "%s sorted out every item possible." ), who.disp_name( false, true ) );
+    if( who.is_npc() ) {
+        npc *guy = dynamic_cast<npc *>( &who );
```
mqrause, line 6550:
```suggestion
        npc *guy = who.as_npc();
```

src/activity_actor.cpp
```diff
@@ -6171,334 +6192,506 @@ static void move_item( Character &you, item &it, const int quantity, const tripo
...
 
-    // If we got here without restarting the activity, it means we're done
-    add_msg( m_info, _( "%s sorted out every item possible." ), you.disp_name( false, true ) );
-    if( you.is_npc() ) {
-        npc *guy = dynamic_cast<npc *>( &you );
+void move_loot_activity_actor::finish( player_activity &act, Character &who )
+{
+    add_msg( m_info, _( "%s sorted out every item possible." ), who.disp_name( false, true ) );
+    if( who.is_npc() ) {
+        npc *guy = dynamic_cast<npc *>( &who );
```
mqrause, line 6694:
```suggestion
        npc *guy = who.as_npc();
```

mqrause:
I'm noticing quite a lot of similarities between the two activity actors. Especially the `do_turn` methods have the same structure but make use of lambdas for their specific differences. Especially with big lambdas like in the move_loot actor the code becomes a lot harder to follow. This could be improved with a generalized `do_turn` method in a common parent actor with the two actors then getting their own specific implementations for `process_item` and `filter_item_all`.

The improvements are huge as is already, so I wouldn't require that change in this PR. Especially since I don't know if or how other loot zone activities could also benefit from that when they get migrated to activity actors. But defnitely something to consider for future refactoring.

PR #63291, opened 2023-01-26
Boomer Glutton Spells and Stats Tweaks

SenpaiSlime:
#### Summary
Balance "Despammify Boomer Glutton Spells and update stats"

#### Purpose of change
The current implementation of the glutton is not as good as i would like it to be, i had forgotten to provide a fungalize and did not do enough to solidify its niche as a area denial zed. It also felt rather spammy as it had two special ranged attacks on 20ish turn cooldowns, which isnt great. I want to update this monster to be the threat it is supposed to be.

#### Describe the solution
I am updating the Boomer Glutton monster to better fit its niche, and to touch up on some parts that were overlooked. Its stats now mostly match the huge boomer, and its death is now a sludge explosion. I have also merged its two ranged special attacks into one while also increasing its actual field frequency, and gave it a variable intensity (its current attack is too weak) , to reduce the spammy feeling nature of the monster.

We can forgo its on death explosion change but i strongly think the despam and spell change is important for solidifying the threat of this monster

#### Testing
I have loaded in and fought a few of these monsters, they seem to be behaving much better in regards to how quickly they can make a room hard to navigate, and punish players without eye protection. Game loads and runs fine with them, spells behaving as expected 

#### Additional context
N/A


data/json/monster_special_attacks/spells.json
```diff
@@ -443,14 +443,17 @@
     "flags": [ "SILENT", "NO_HANDS", "NO_LEGS", "NO_EXPLOSION_SFX" ],
     "min_aoe": 45,
     "max_aoe": 45,
-    "base_casting_time": 75,
+    "base_casting_time": 50,
```
GuardianDll, line 446:
```suggestion
```
base_casting_time is not used for monster attacks

data/json/monster_special_attacks/spells.json
```diff
@@ -443,14 +443,17 @@
...
     "max_range": 4,
     "effect": "attack",
+    "effect_str": "boomered",
+    "min_duration": 2400,
+    "max_duration": 2400,
     "shape": "cone",
     "field_id": "fd_sludge",
-    "field_chance": 2,
-    "min_field_intensity": 2,
+    "field_chance": 1.5,
```
GuardianDll, line 455: field_chance works with float?

PR #63293, opened 2023-01-26
Update description of zipper bag (bag_zipper)

mikegogulski:
#### Summary
Content "Update zipper bag description"

#### Purpose of change

Update the description of the zipper bag to improve some awkward phrasing and a possible translation error from the original PR author's native language, with all due respect and apologies to @damalsk.

#### Describe the solution

Makes the description of the item read more fluently.

Went through a bunch of different phrasing possibilities before arriving at this. Recalled that per the project rules it's permissible to use brand names. Learned too much for one day about how zipper bags work. Checked the IRL dimensions of Ziploc sandwich bags against the defined 750 mL max volume to ensure verisimilitude and compatibility with the "roomy enough to hold a sandwich" phraseology. Recalled that people outside the US might have different ideas about the dimensions of a "standard sandwich", but calmed down by listening to Rammstein's "We All Live in America" 23 times at extreme volume.

#### Testing

Spawned the item into a test world and verified that the updated text appears.

image

### DIFF
--- a/data/json/items/containers.json
+++ b/data/json/items/containers.json
@@ -260,7 +260,7 @@
     "category": "container",
     "name": { "str": "zipper bag" },
     "looks_like": "bag_plastic",
-    "description": "A resealable transparent plastic bag roomy enough to hold a sandwich.  Known generically as a Ziploc bag, it can be sealed and opened with a press-together mechanism similar to a zipper.",
+    "description": "A resealable, watertight transparent plastic bag roomy enough to hold a sandwich.  Known generically as a Ziploc bag, it can be sealed and opened with a press-together mechanism similar to a zipper.",
     "weight": "2 g",
     "volume": "10 ml",
     "price": 0,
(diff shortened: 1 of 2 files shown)


KHeket:
Can you also add to discription, that zippers are waterproof?

mikegogulski:
@KHeket: A brief survey of search results gives conflicting information as to whether or not zip bags are waterproof, and to what degree they are. Based on the little bit that I've read, I'd only be comfortable saying things like "a zip bag picked at random from all zip bags on the market is more than 50% likely to be one of the types that will admit at least some water when fully immersed for 60 minutes".

All that being said, the item's `pocket_data` field includes the `watertight` flag set to `true`, so I will push a commit to add this to the description. Thanks!

bombasticSlacks:
Rebased your code so that tests will rerun :)

mikevendelux:
w00t, first merge :)

PR #63298, opened 2023-01-26
Alternate ironshod staff recipe, no forging required

Aerin-of-the-Toast:
#### Summary
None

#### Purpose of change

The ironshod quarterstaff is a fairly simple upgrade over a bo staff. It should be pretty simple to put steel end caps onto an already-made bo staff without needing an entire blacksmithing forge.

#### Describe the solution

Adds a new alternative ironshod quarterstaff recipe. This uses 2 small metal sheets (acting as the end caps), metal sawing and hammering 2 (to cut the metal sheets and shape them over the ends), and drilling 2 and some nuts and bolts (to drill a hole with which to bolt the end caps on).

Making this a new item entirely, but it felt to me like it'd be so similar to the current version that it'd just be a bit pointless to make the distinction.

#### Testing

Made sure the recipe showed up in game.

#### Additional context

While it is wrong that this outputs a staff that's half _iron_ instead of steel, the same could be said for the original recipe anyway that's explicitly using steel scraps, unless we're assuming the survivor somehow decarburizes the steel into wrought iron. Fixing that should be its own PR for later.

### DIFF
--- a/data/json/recipes/weapon/bashing.json
+++ b/data/json/recipes/weapon/bashing.json
@@ -841,7 +841,7 @@
     "time": "30 m",
     "autolearn": true,
     "proficiencies": [ { "proficiency": "prof_metalworking" } ],
-    "components": [ [ [ "sheet_metal_small", 2 ] ], [ [ "nuts_bolts", 4 ] ], [ [ "bo", 1 ] ] ],
+    "components": [ [ [ "sheet_metal_small", 6 ] ], [ [ "nuts_bolts", 4 ] ], [ [ "bo", 1 ] ] ],
     "qualities": [ { "id": "HAMMER", "level": 2 }, { "id": "SAW_M", "level": 2 }, { "id": "DRILL", "level": 2 } ]
   },
   {


data/json/recipes/weapon/bashing.json
```diff
@@ -829,6 +829,21 @@
...
+    "result": "i_staff",
+    "id_suffix": "with sheet metal",
+    "category": "CC_*",
+    "subcategory": "CSC_*_NESTED",
+    "skill_used": "fabrication",
+    "difficulty": 2,
+    "time": "30 m",
+    "autolearn": true,
+    "proficiencies": [ { "proficiency": "prof_metalworking" } ],
+    "components": [ [ [ "sheet_metal_small", 2 ] ], [ [ "nuts_bolts", 4 ] ], [ [ "bo", 1 ] ] ],
```
Maleclypse, line 844: how big is a sheet metal small?

Aerin-of-the-Toast: Anywhere from 15x15cm to 20x20cm, which is more than large enough for an end cap I think since they're not meant to be mace heads. I also checked the weight; a small metal sheet weighs 0.25kg and the ironshod staff is only 0.25kg heavier than a bo staff, so overall this actually loses a bit of weight (presumably from just the extra bits being cut down).

Maleclypse: I have doubts that is enough metal to make a quarterstaff do more damage?

Aerin-of-the-Toast: The same could be said for the other recipe, then. Also, my guess is the weight for the ironshod was originally based off the weight for the regular quarterstaff which it originally used as the component, which only weighs 0.61kg.

It's probably the ironshod needs a weight increase overall, or maybe the bo staff just weighs too much, but that's probably out of scope for the PR. Otherwise, if more weight is needed for this particular crafting recipe, I could just increase the amount of small metal sheets to 6 or so under the assumption they're being layered for the end caps.

Aerin-of-the-Toast: I increased the small metal sheet cost from 2 to 6.

PR #63302, opened 2023-01-26
Add information about migration and obsoletion into JSON_INFO.md

GuardianDll:
#### Summary
None
#### Describe the solution
Update the JSON_INFO.md with information about obsoletion and migration

### DIFF
--- a/doc/JSON_INFO.md
+++ b/doc/JSON_INFO.md
@@ -40,7 +40,6 @@ Use the `Home` key to return to the top.
     - [Item Category](#item-category)
     - [Item Properties](#item-properties)
     - [Item Variables](#item-variables)
-    - [Item Migrations](#item-migrations)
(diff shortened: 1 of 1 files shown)


irwiss:
There's also existing migration info in that file https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md#item-migrations might be worth consolidating into one spot

GuardianDll:
First is already here, i just forgot to remove it
About second, what exactly reset_item_vars is?

irwiss:
It deletes all the `item_vars` from the existing `item`, replacing them with the item vars defined in the item prototype (defined in `itype::item_variables`)
What i used it for - migrating the old "convertible" vehicles required deleting off old item vars, and setting fresh ones that define the vehicle parts inside an item type

PR #63324, opened 2023-01-28
Unarmed hits correctly gain unarmed experience

### LINKED ISSUE #63257: Unarmed combat skill leveling with unarmed weapons, gauntlets and bare-handed
### Describe the bug

Even after #62578, using an unarmed weapon or a gauntlet leads to both unarmed combat and bashing skill rising. Additionaly, bashing skill increases 2-3 times faster than unarmed combat.
When fighting bare-handed, unarmed requieres to kill almost twice the amount of targets to gain the same exp as fighting armed. This affects melee skill too, also gainig only half exp, compared to fighting armed.
The two issues voiced in #61682 are stil present:

1. Fighting unarmed your skills rise much slower, even though you spend more time/perform more hits. This affects both "melee" and "unarmed combat" skills.
2. Wearing unarmed weapons, gloves or boots (if you use style that has techniques using said vector), raises both bashing and unarmed (~3% bashing for 1% unarmed).

### Steps to reproduce

1. Make a character
2. Hit some monsters unarmed
3. Note the rate at wich you gain xp
4. Get a melee weapon, hit some more monsters
5. Notice xp % going up faster
6. Wear gauntlets or unarmed weapon and hit some more
7. Notice bashing rising, and rising faster than unarmed combat

### Expected behavior

1. Unarmed combat rising at the same rate as bashing **or** unarmed rising at the same rate as melee skill **or** some where inbetween 
2. Wearing unarmed weapons rising either only unarmed skill, or rising mainly unarmed skill + a bit of damage type skill

### Screenshots

_No response_

- OS: Windows
- Game Version: b96b791 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Standard combat test suite [standard_combat_test]
]

### Additional context

_No response_

ArgonMercury:
#### Summary

Bugfixes "Unarmed hits now correctly gain experience"

#### Purpose of change

Closes #63257 

Currently, landing an unarmed attack cannot gain you unarmed exp. Missing an unarmed attack can get you unarmed and melee exp up to level 6 as intended. Hitting with your bare fists gives you no exp, unarmed or melee. Hitting with a worn item (e.g. light gloves) will gain melee exp, and if the worn item has bash/cut/pierce values you will gain experience for that (even though those skills will not increase the damage dealt by those values), but you will not gain any unarmed exp.

#### Describe the solution

Currently there is a check that makes it so if you aren't hitting with a wielded weapon or a worn item you get no exp for either unarmed or melee; that is removed.

Secondly melee_train has had attack_vector input added and now checks if the attack vector is WEAPON. If it is not WEAPON, then it trains unarmed, otherwise it uses the values of the current weapon to train the relevant skills. Missing additionally has a fake attack vector set for WEAPON or HAND depending on if you are using a weapon to attack (this works properly with taekwondo).

Not using attack vector and putting the attack vector check before calling melee_train and having it set weapon to null item reference if attack vector isn't WEAPON. Putting attack vector in melee_train seems more robust and clarifies the intended behavior so hopefully this bug won't crop up yet again.

#### Testing

Compiled and attacked debug monster with several set ups and all had expected behavior.

No style and bare fists = Gain unarmed and melee for both hits and misses
No style and pipe mace wielded = Gain bashing and melee for both hits and misses
No style and light gloves worn = Gain unarmed and melee for both hits and misses
Taekwondo and pipe mace wielded = Gain unarmed and melee for both hits and misses (taekwondo forces you to use kicks)

#### Additional context

I have no idea what the intended functionality of attack_override is, but attack_overrides with no vector set have a vector of NONE so they will still correctly use and train unarmed skill even with their special use case removed.

### DIFF
--- a/src/melee.cpp
+++ b/src/melee.cpp
@@ -438,7 +438,8 @@ void Character::roll_all_damage( bool crit, damage_instance &di, bool average,
     roll_other_damage( crit, di, average, weap, attack_vector, crit_mod );
 }
 
-static void melee_train( Character &you, int lo, int hi, const item &weap )
+static void melee_train( Character &you, int lo, int hi, const item &weap,
+                         const std::string &attack_vector )
 {
     you.practice( skill_melee, std::ceil( rng( lo, hi ) / 2.0 ), hi );
 
@@ -451,7 +452,7 @@ static void melee_train( Character &you, int lo, int hi, const item &weap )
     float total = std::max( cut + stab + bash, 1 );
 
     // Unarmed may deal cut, stab, and bash damage depending on the weapon
-    if( weap.is_null() ) {
+    if( attack_vector != "WEAPON" ) {
         you.practice( skill_unarmed, std::ceil( 1 * rng( lo, hi ) ), hi );
     } else {
         you.practice( skill_cutting,  std::ceil( cut  / total * rng( lo, hi ) ), hi );
@@ -680,7 +681,8 @@ bool Character::melee_attack_abstract( Creature &t, bool allow_special,
 
         // Practice melee and relevant weapon skill (if any) except when using CQB bionic
         if( !has_active_bionic( bio_cqb ) && !t.is_hallucination() ) {
-            melee_train( *this, 2, std::min( 5, skill_training_cap ), cur_weap );
+            std::string attack_vector = cur_weapon ? "WEAPON" : "HAND";
+            melee_train( *this, 2, std::min( 5, skill_training_cap ), cur_weap, attack_vector );
         }
 
         // Cap stumble penalty, heavy weapons are quite weak already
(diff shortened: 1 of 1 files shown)


Heitomos:
I tried this, but unarmed hits still don't work. Misses do, but at Unarmed 6+ all XP stops.

ArgonMercury:
> I tried this, but unarmed hits still don't work. Misses do, but at Unarmed 6+ all XP stops.

Just tested with current experimental and I got to level 10 unarmed punching a kevlar hulk (with 2000 int). Since #63773 you need to hit monsters that have a high melee skill to gain higher levels of exp, which is what I'd assume to be your problem.

PR #63335, opened 2023-01-29
Yarn clothing disassembly rebalance

Karol1223:
#### Summary
None

#### Purpose of change
I was really sick of having to take 7 hours to take apart a turtleneck. Come on.

#### Describe the solution
Removed the 'reversible' line from most yarn clothing that had it. They have been instead given proper disassembly recipes that give almost all of their mass back in yarn but take significantly shorter. On top of that, the scarf was made 100g as it weighing less than a pair of gloves made no sense. The turtleneck now also takes less yarn to craft so that it doesn't lose 1/3 of it's mass when turned into an item.


Maleclypse (commented):
Watched a video on unraveling sweaters and this looks like it's in the right direction.

PR #63337, opened 2023-01-29
Oversized greatbows

Nebnis:
#### Summary
Balance "Added oversized tag to greatbow and compound greatbow"

#### Purpose of change
I believe greatbows are big enough to accommodate large mutants as they are 2 meters big and it makes them fit a nice niche with them as well.

#### Describe the solution
Added oversized flag to them.

#### Testing
Logged and equipped them everything is okay.


Nebnis:
Looks like there is a sling already that does what i want without doing this, plus discord raised some good points

PR #63355, opened 2023-01-30
Update professions.json

zeeb1337:
Made Shower Victim into -2 points instead of -1 since some other stronger starting professions are -1. It didn't make much sense to have it equally strong with much less available items and clothing.

### DIFF
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -2743,7 +2743,7 @@
     "id": "svictim",
     "name": "Shower Victim",
     "description": "You just stepped out of a nice, hot shower to find the world had ended.  You've got some soap, along with the most massively useful thing ever… a towel.",
-    "points": -1,
+    "points": -2,
     "flags": [ "NO_BONUS_ITEMS" ],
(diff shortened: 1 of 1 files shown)


intergalacticdata:
The other -2 professions have some kinda advanced problems built in. Shower victim is like: got no shoes...

Aerin-of-the-Toast:
The main issue here is you didn't follow the PR template at all. But also this doesn't matter much anyway. Adjusting point balance for professions is kinda like rearranging the deck chairs on the Titanic as points aren't going to stick around forever, and this doesn't even make any sense anyway since we don't have any other -2 point professions and Shower Victim isn't notably any more difficult to survive with than e.g. hobo, groom/bride, and especially Naked and Afraid which is just strictly worse since you don't even start with a towel and soap (But also not worse enough to be worth the extra point either).

Fris0uman:
On top of explanations given by others, point cost are mostly an indication of how hard a thing is. Changing a value by one really doesn't do anything, so I'm just going to close this.

zeeb92:
Fine by me.

PR #62936, opened 2022-12-30
[WIP] Scrutinize and fix densities of metal resource items

cake-pie:
#### Summary
Bugfixes "Scrutinize and fix densities of metal resource items"

#### Purpose of change

- Resolves #62928
  - rectify errors introduced in 8ccc9d98faecc852fe3108d6c290b8fd8689cdd4, 68052db00fcd341ba3a6e6875212ffea25f208e4 and propagated by other subsequent commits.
- general audit of all weight / volume / stack size in `data/json/items/resources/metal.json`
- reordering and reorganization so that related entries (material or form factor) are clustered together
- adopt `copy-from` for the many steel type variants to reduce redundancy
- take advantage of the change from unitless weights/volumes to improve the accuracy of some of these stats which in the past had to be rounded off, leading to pretty crude approximations in some cases (e.g. 2.445g per unit of bismuth rounded down to 2g, nearly 20% error!)

#### Describe the solution

Cross reference known good version of the file from before the regression, but not necessarily reverting to those values, since we can improve precision while we're at it. Base density figures via wikipedia infobox unless otherwise indicated.

Read further for full breakdown of changes.

#### Describe alternatives you've considered

Revert commits responsible for the regression, but it's already been a few months and there have been other intervening commits.

#### Testing

Not yet.

#### Additional context

Work in progress.

It may be hard to comprehend changes if looking at a consolidated view of all changes in the PR. Instead, in github's "Files changed" tab you can narrow down to changes from each commit. I try to organize work into small blocks for ease of review, so that should be a better way to step through and follow along with my edits.

Note that because many of these items have had way too small volumes, this fix may not be friendly to existing save files -- after the fix, volume of contents may exceed capacity of containers.

---

### Progress

<details>
<summary>Click to show/hide full breakdown of changes and remaining todos</summary>

#### Basic steel resources
Items representing plain old `steel` in unworked form; other form factors listed in a separate section.  
`steel_lump`, `steel_chunk`, `scrap` & `sheet_metal_small`

- [x] regression (wrong volume) fixed
- [x] weight stat precision improved
- :exclamation: recipes involving these items need to be checked for contagion arising from the regression
[...]

PR #62604, opened 2022-12-06
Reset daily health at the end of each day

fungamer2-2:
#### Summary
Bugfixes "Reset daily health at the end of each day"

#### Purpose of change

There were certain issues with the current system as stated in #59557, the main one being that daily health didn't reset every day, instead decaying slowly every 6 hours. As stated in the issue, this allowed small health changes to be amplified quickly.
So I decided that this might be a simple enough fix for stable

#### Describe the solution
Reset `daily_health` after adding to `health_tally`, and remove the thing where `daily_health` decays every 6 hours

#### Describe alternatives you've considered
Not fixing it

PR #63365, opened 2023-01-31
Rectify incorrect densities of steel resource items

### LINKED ISSUE #62928: Steel resource items have erroneous densities
### Describe the bug

Stats for steel metal in `data/json/items/resources/metal.json` are broken, yielding unrealistic items in game that are denser than a solid ingot of steel material, or indeed, denser than any known material on earth.

Certainly, different values for density of steel can be found depending on what sources you look up and the exact blend of steel alloy under discussion. I recall from working on #25032 and #25210 that someone had already selected 7.60g/cm³ for solid blocks of "normal" steel and also rationalized how crumpled fragments like scrap metal (`scrap`) takesup more space than solid blocks and have lower density in practice. You could try to make an argument for larger values up to around 8g/cm³ or slightly more, considering there are now even more different varieties of steel in the game, but that's kind of besides the point. The issue at hand is clearly due to some kind of error and not something that can be attributed to design or balancing decisions.

### Steps to reproduce

Look up the density of steel, survey multiple sources to get a feel of the range of values
Use debug to wish up various amounts of steel resource items
Do some math

### Expected behavior

item stats have some basis in reality

### Screenshots

scrap
20 g/cm³ for scrap metal that is crumpled fragments that don't stack well. > 2.5 × density of a solid block and 10 × intended

chunk
32.26 g/cm³ ... is denser than anything on earth. (Osmium and iridium are around 22.6 g/cm³)

irrelevant.

### Additional context

Scope of items afflicted is not clear. I have merely checked the basic `scrap`, `steel_chunk` and `steel_lump` and found all three to be in error. There are many more items corresponding to different blends of steel that need to be checked. I do not know if the problem extends beyond steel items.

From an initial brief inspection of git blame I am fairly certain that #56870 is one of the primary causes behind the problem, as it changed many item volumes -- that PR divided all volumes by `stack_size`, seems the author did not know that stack size is factored into the calculation. Another possible contributor is #60839, which despite being about ingots, also divided volumes of chunks by 2 without adequate explanation.

Suggest to start by scrutinizing changes from those two PRs, though a complete audit of `weight`, `volume` and `stack_size` throughout `metal.json` may be needed.

cake-pie:
#### Summary
Bugfixes "Rectify incorrect densities of steel resource items"

#### Purpose of change
Resolves #62928

- rectify errors introduced in 8ccc9d9, 68052db and propagated by other subsequent commits.
- ~reordering and reorganization of `data/json/items/resources/metal.json` so that related entries (material or form factor) are clustered together~
- ~move some misplaced items out of the file, to more appropriate locations~
- adopt copy-from for the many steel type variants to reduce redundancy

#### Describe the solution
Cross reference known good version of the file from before the regression.

Fix impacted budget steel smelting recipes [per discussion](https://github.com/CleverRaven/Cataclysm-DDA/pull/62936#issuecomment-1378478543).

~Oven components `fanblade_metal` and `heavy_wire_rack` moved to `appliance_parts.json`~
~Door hinge `hinge` moved to `home_improvement.json`~

Simply revert the commits responsible for the regression, but it's already been a few months and there have been other intervening commits.

#### Testing
Check in game that:
- the item stats are sane
- the budget steel smelting recipe inputs/outputs are sane

#### Additional context and future work
Part of #62936 which is being reorganized and re-done as it was starting to get out of hand.  
Bits of that PR will continue to be split off as smaller, more manageable chunks and dealt with separately.

It was originally intended in #62936 to take advantage of the change from unitless weights/volumes to improve the accuracy of some of these stats which in the past had to be rounded off, leading to pretty crude approximations in some cases. That is now being deferred as a non-critical enhancement.

For instance, based on preexisting comments, we know that steel was intended to be 7.6g/cm³, but in practice it is 8.0 due to rounding errors (in the age of unitless stats). That can be rectified by adjusting either mass or volume; it seems [preferable to leave mass untouched](https://github.com/CleverRaven/Cataclysm-DDA/pull/62936#issuecomment-1378343605) but further discussion may be warranted.

Note that steel ingots stats currently don't line up with the general density of other smaller steel resource items, as they appear to have used 7.859g/cm³. Leaving it alone for now as these are uncommon: they don't spawn in world, only player-craftable for bulk-storage and handling convenience. In any case, resolution of this discrepancy will depend on how steel density overall gets adjusted.


cake-pie:
@Drew4484 I hope ae932d9 is satisfactory

cake-pie:
Unrelated test failure likely originating from #62604

cake-pie:
Rebased to resolve conflicts.

bombasticSlacks:
Alright, so you need to move back hinges, blades, and wire (as well as any other moved items)

I understand messy organization can be frustrating, but moving the items means we lose **all the history of changes to the entries** which is a hard no for the project.

The same goes for grouping the items, moving them around messes with history (and also makes this PR very annoying to review). The reorganizing within the file causes the same issues. 

doing either of these things the code will be rejected because we lose the history of the items and it breaks all other concurrent work. 

Again sorry I know this must be frustrating and I wish I had noticed the issue sooner.

data/json/items/resources/appliance_parts.json
```diff
@@ -73,5 +73,35 @@
     "weight": "5 kg",
     "volume": "20 L",
     "longest_side": "69 cm"
+  },
+  {
+    "id": "fanblade_metal",
```
bombasticSlacks, line 78:
```suggestion
    "id": "fanblade_metal",
```

As commented, needs to be moved

bombasticSlacks (changes requested):
requesting changes so no one yolo merges this while it has these problems

cake-pie:
> we lose **all the history of changes to the entries**

???

This is untrue -- keeping history of changes is the whole point of version control systems like git. I've previously traced change history in this project across moved/renamed files, files that were split up, and other instances of c++ code or json entries moved across different files; no problem whatsoever.

I don't understand what your complaint is, please elaborate,

> very annoying to review

You do realize you can review changes commit-by-commit or for a subset of commits in the PR, right?

I make it a point to organize my work into commits that reflect coherent steps; in this case the sorting is in a separate commit of its own, which can be checked to ensure that entries were only moved as-is, without changes.

bombasticSlacks:
I spoke in simple terms because I didn't recognize your name, many people working on the project don't understand the nuances of GIT, so explaining things simply is important to get points across. 

It looks like you've been doing a lot of work over the last while and you understand how software works, so I'll speak clearly:

The project position is that JSON organization is pointless data churn and that it isn't worth even the minor inconvenience of burying git history over reorganization commits. Someone else is just going to come along and fuck it up by putting stuff in between, so it just makes data gathering annoying for 0 long-term benefit. 

>You do realize you can review changes commit-by-commit or for a subset of commits in the PR, right?

>I make it a point to organize my work into commits that reflect coherent steps; in this case the sorting is in a separate commit of its own, which can be checked to ensure that entries were only moved as-is, without changes.

150 PRs have been merged in the last 2 days. I would encourage you to have some perspective. 

Hopefully, that makes the position of the developers clearer, you should incorporate that instruction into this PR and others, or they will not get merged.

anothersimulacrum:
The merits of moving entities around aside (though I think it ought be avoided).

> in this case the sorting is in a separate commit of its own, which can be checked to ensure that entries were only moved as-is, without changes.

Human checking of this is rarely perfect, and effort intensive. The easier you make your PR to review, the faster it is to merge and less effort will be required of you.

cake-pie:
commits dropped and overall rebased

PR #63366, opened 2023-01-31
github: make save files mandatory for new bug reports

andrei8l:
#### Summary
None

#### Purpose of change
Help improve the quality of bug reports

#### Describe the solution

- Add a mandatory text area with instructions for #62966
- Replace wordy courteous sentence that no one ever reads with straight instructions

Adding the field but not making it mandatory: it's a half measure and people can circumvent the requirement anyway.

#### Testing

Can be tested on [my fork](https://github.com/andrei8l/Cataclysm-DDA/issues/new?template=bug_report.yaml) once [2023-01-31-0603](https://github.com/CleverRaven/Cataclysm-DDA/releases/tag/cdda-experimental-2023-01-31-0603) finishes building 

Screenshot from 2023-02-01 06-10-07

#### Additional context
N/A


Maleclypse:
I do have a question about how we'll add CI bugs though after this.

andrei8l:
The template doesn't specifically check for an uploaded file so you can:

- link a failed run, or
- write `N/A`, or
- write `I am not a bug report number. I'm a free man!`

All are valid options.

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -55,8 +69,7 @@ body:
     attributes:
       label: Versions and configuration
       description: |
-        The following information should consist only of the automated game report generated by the game; please put further information below.
-        To generate a report: "Main menu (ESC) > Debug Menu > Info > Generate game report".
+        `F1`Debug Menu -> `i`Info -> `r`Generate game report
```
Qrox, line 72: Maybe move the minimized save archive option to the main menu's debug menu? The real debug menu contains way too many options that are not supposed to be used in a normal gameplay.

andrei8l: That shows the full debug menu if the key is bound and I didn't realize there was a limited set always available. Thanks!

PR #63369, opened 2023-01-31
New missions for the heli crash scenario

Termineitor244:
#### Summary
Content "Adds 2 new simple missions to the helicopter crash scenario"

#### Purpose of change

A military squad crashing in the wilderness with only one survivor while in transport sounds like a very good scenario to have more content, or, in this case, more missions to explore these special circumstances a little.

#### Describe the solution
Adds 2 new missions, one for all the passengers of the helicopter to be able to return to their base of origin and another one for the pilot to search for a new ride in the supply point closest to the crash, just a pair of nice little flavor missions to spice the scenario.

To revamp a little the zombie spawns of the 2 locations while I was here, since they seem pretty... bland. Specially the military helipad where the game just spawns a bunch of zombie soldiers glued together in the same point, but that can always be done in a different PR if I find the time.

#### Testing
Tested the changes in my local copy of the game of my phone, the missions generate normally and everything works.

#### Additional context
I was able to make and test this in a cave! With a trash text editor! Meaning in my work, with my cellphone haha.

### DIFF
--- a/data/json/starting_missions.json
+++ b/data/json/starting_missions.json
@@ -640,7 +640,7 @@
       "assign_mission_target": {
         "om_terrain": "helipad_nw",
         "reveal_radius": 10,
-        "om_special": "helipad",
+        "om_special": "military helipad",
         "search_range": 70,
         "min_distance": 25,
         "z": 0
--- a/data/json/starting_missions.json
+++ b/data/json/starting_missions.json
@@ -628,20 +628,6 @@
     "origins": [ "ORIGIN_GAME_START" ]
   },
   {
-    "id": "MISSION_HELICOPTER_REFUEL",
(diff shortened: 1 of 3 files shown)


data/json/starting_missions.json
```diff
@@ -627,6 +627,82 @@
     },
     "origins": [ "ORIGIN_GAME_START" ]
   },
+  {
+    "id": "MISSION_HELICOPTER_REFUEL",
```
Night-Pryanik, line 631: This new mission doesn't seem to be set neither for profession nor for scenario?

Termineitor244: This mission got mistakenly copy pasted from just above in the doc when I passed the changes to my pc, I need to delete it.

PR #63375, opened 2023-01-31
Fix `zinc_metal` stats (and cascade effects)

cake-pie:
#### Summary
Bugfixes "Fix `zinc_metal` stats (and cascade effects)"

#### Purpose of change
Split off from #62936

The density of `zinc_metal` is way off. Fix it, and deal with cascading effects in recipes.

#### Describe the solution

Zinc has density 7.14g/cm³. That's 357g for 50ml volume, which divides into a stack of 200 charges, 1785mg per unit.
&nbsp;

Adjust the disassembly of 5L zinc ingot, applying a 5.5% lossiness -- as I work my way through the spinoffs of #62936, 5~6% loss rate that will be applied consistently to all 5L ingot disassembly.
&nbsp;

Migrated item zinc oxide with misleading id `chem_zinc` to `chem_zinc_oxide` to prevent confusion with actual elemental/metal zinc items. Drive-by adjustment to volume due to discrepancy in quoted density figure -- not expected to affect recipes, since those depend on the weight (which corresponds to 1/100 moles of chemical).
&nbsp;

**Edit:** adopted `coal_lump` and `charcoal` stats from #62180
~Fix weight of `coal_lump`, from 1.5g to 37.5g. There was a mistake in 5549b62 where divided by count (250) instead of stack size (10) leading to value off by a factor of ×25. This is required for sanity of the next step.~
~Check cascade effects of coal weight change; see [details below](#issuecomment-1411414456)~
&nbsp;

As a source of elemental carbon for chemistry, `coal_lump` treated as anthracite coal with 90% carbon content.  
For `charcoal`, carbon content of wood charcoal can vary drastically (sources give 50~95%). Picked a fairly generous value of ~80.56% for convenience, since that will allow `charcoal` and `coal_lump` to be used at parity.
&nbsp;

Fix recipe for smelting `zinc_metal` from ore `material_zincite` by reduction with carbon. 
- Quantities reworked on the basis of zinc content of zincite, to ensure balanced input and output.
  - Zincite composition: Zn<sub>0.9</sub>Mn<sup>2+</sup><sub>0.1</sub>O <small>[[source](https://www.webmineral.com/data/Zincite.shtml)]</small>
- Added hammer and chisel tool requirements to break ore and coal into smaller pieces before roasting; would not work if you just chucked a large chunk of ore in.
- Replace basic boil and contain requirements with crucible due to high temperatures. 
- Added fuel required for roasting. This is distinct from the coal being used as source of carbon reagent. Smelting roughly 1.5kg of mass in total, or 6 steel-chunk-equivalents.
&nbsp;

Fix recipe for further reduction of zinc oxide with carbon; this uses an extremely conservative approach that oversupplies carbon.
- takes a hint from [this](https://royalsocietypublishing.org/doi/10.1098/rsos.170710), "to achieve the complete reduction of ZnO, the amount of reducing agent was increased to two times the stoichiometric amount."
- Replace basic boil and contain requirements with crucible due to high temperatures. 
- Added fuel required for roasting. This is distinct from the coal being used as source of carbon reagent. Smelting roughly 500g of mass in total, or 2 steel-chunk-equivalents.
&nbsp;

Adjust recipe for grinding of `zinc_metal` into `chem_zinc_powder`; reworked to produce ~130g, on the same scale as aluminum powder grinding recipe (176g output) and maintaining roughly the same material inefficiency as before (was 1308g/1400g, 6.6% lossy). This only balances zinc input and output, other issues with this recipe are marked with TODO and deferred to be fixed as a package with other problems found with powdered metals in #62936
&nbsp;

Adjust battery recipes by scaling `zinc_metal` inputs to roughly preserve mass. However, proportion also needs to be preserved when comparing among those recipes, so it is a straightforward scaling by 4× rather than 3.922× followed by rounding. (This is naive multiplication only, I did not look into the scientific basis for these recipes.)
&nbsp;

This should cover all pertinent references to `zinc_metal` in `src/`, `data/json/` and `data/mods`.

#### Testing
Suggested checks:
- Double check my maths / chemistry derivations.
- Load in game and check
  - zinc metal has sane stats
  - 5L ingot disassembly recipe reports sane input/output in crafting menu
  - check zinc smelting recipe in crafting menu for sanity
  - check zinc powder grinding recipe in crafting menu for sanity
  - check battery recipes in crafting menu for changes applied correctly.


cake-pie:
Adjustments to usages of `coal_lump` going to be listed here because it's kind of a spinoff sub-issue so it's neater to keep it separated from the `zinc_metal` stuff.

---

( CaC<sub>2</sub> stuff goes here )

---

Magiclysm Nova Flare spell requirements
```diff
 {
     "id": "spell_components_nova_flare",
     "type": "requirement",
     "//": "Fire starting items suitable for the Nova Flare spell",
+    "//1": "1.3MJ energy content",
+    "components": [ [ [ "coal_lump", 1 ], [ "lamp_oil", 35 ] ] ]
-    "components": [ [ [ "coal_lump", 30 ], [ "lamp_oil", 5 ] ] ]
 }
```
1 `coal_lump` is 37.5g of anthracite coal and has 1.3MJ  
https://www.wolframalpha.com/input?i=37.5g+*+higher+heating+value+of+anthracite+coal

35 `lamp_oil` is 35ml or 35cm<sup>3</sup> of kerosene and has 1.33MJ  
https://www.wolframalpha.com/input?i=35cm3+*+density+of+kerosene+*+higher+heating+value+of+kerosene

PatrikLundell:
A note regarding ore processing: The hammer+chisel requirement isn't an issue here, but be careful when you get to iron ores, as we don't want to put Innawood in a catch-22 situation where you need the tools to process the ore you need to get the tools. For that case it might be useful to add an additional recipe that uses more primitive tools at the expense of it taking longer.

Karol1223:
> A note regarding ore processing: The hammer+chisel requirement isn't an issue here, but be careful when you get to iron ores, as we don't want to put Innawood in a catch-22 situation where you need the tools to process the ore you need to get the tools. For that case it might be useful to add an additional recipe that uses more primitive tools at the expense of it taking longer.

Couldn't Innawoods issues be handled in Innawoods-specific recipes? It's a bit weird to adjust mainline recipes specifically to avoid problems in Innawoods. I had the same thing happen when I changed the antenna recipe and made electronics not able to be learned in Innawoods anymore - I just ported the old recipe that allowed for it to be Innawoods-specific.

PatrikLundell:
Yes, I would suggest it would be an Innawood specific recipe. However, this PR is tagged with Innawood, so I didn't think to mention that specifically.

cake-pie:
The tagging happened automatically because in this PR I have adjusted the innawood version of the battery recipe alongside the mainline one when adjusting input amounts.

Noted the potential smelting catch-22. Depending on the ore's hardness, I guess you could just try smashing it with another rock of harder material, or something like that.

PatrikLundell:
Come to think of it, bog iron isn't actually huge chunks of rock, so it wouldn't need much or any bashing in the first place. Thus, hematite could use the regular breaking apart routine, while the Innawood specific bog iron might not actually need any tool change. Even if that's needed, pulverizing it with a rock (hammer 1) ought to be sufficient.

cake-pie:
Ready for review but not merging -- will rebase after #63365 is merged, dropping 21c4214.

anoobindisguise:
I already fixed the weight of coal in pending #62180. this will conflict.

cake-pie:
> I already fixed the weight of coal (and charcoal) in pending #62180. this will conflict.

Noted, thanks for the heads up

cake-pie:
Rebased to incorporate changes from #62180.

PR #63377, opened 2023-01-31
Remove bashing stat from (chemical) powders, pastes and liquids

cake-pie:
#### Summary
Bugfixes "Remove bashing stat from (chemical) powders, pastes and liquids"

#### Purpose of change
Several items in `data/json/items/chemicals_and_resources.json` (many are chemicals) that are described/defined to be in powder, paste or liquid form have a `bashing` stat for use as a melee weapon. This is generally =1 which is really poor and impractical for actual combat, but why even have the property in that case?

I can understand it for small stones and pebbles, but it just doesn't make sense for sand or anything finer grained than that.

#### Describe the solution
Remove the stat from those items that are in powder, paste or liquid form.

Drive-by change: set chunk of sulfur to material stone rather than powder (retains bashing).

### DIFF
--- a/data/mods/TEST_DATA/known_bad_density.json
+++ b/data/mods/TEST_DATA/known_bad_density.json
@@ -277,7 +277,6 @@
       "dogfood",
       "bullwhip",
       "book_nonf_hard_homemk_grtrms",
-      "chunk_sulfur",
       "recipe_fauxfur",
       "water",
       "freeze_dried_meal_hydrated",


Maleclypse:
Your change to the chunk of sulfur appears to have broken a density test.  

```Check failure on line 774 in tests/item_test.cpp

GitHub Actions
/ Basic Build and Test (Clang 6, Ubuntu, Curses)
2.0f > 2.8f
```

cake-pie:
More like fixed it!
```
(~[slow] ~[.],starting_items)=>   chunk_sulfur had its density fixed, remove it from the list in data/mods/
(~[slow] ~[.],starting_items)=>   TEST_DATA/known_bad_density.json
```

PR #63387, opened 2023-02-01
Fix batch crafting calories displayed wrong

### LINKED ISSUE #63176: Batch recipes with calories range are displayed wrong
### Describe the bug

Let number of Calories or vitamin percentage be value V, when batch crafting B recipes, then instead of V, V/B is displayed. See screenshots.
Actually, `cooked mushroom` is broken too.
Lots of recipes are broken.the 

### Steps to reproduce

1. Be able to craft recipes with calorie range.
2. Go to the crafting menu.
3. `B`atch craft.
4. As the number of results increases, calories, vitamins etc decrease.

### Expected behavior

Recipes show calories consistently.

### Screenshots

image
image
image
image

- OS: Windows
- Game Version: 0.F-13072-g763331f711-dirty [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

SurFlurer:
#### Summary
Bugfixes "Fix batch crafting calories displayed wrong"

#### Purpose of change
Fix #63176

#### Describe the solution
Check if it's using a recipe for calories calculation. If yes, omit batch number's effect to the item's charge.

Change how `food_info()` displays the calories. I think there's possibly precision loss when the calories is divided by big batch number, so it's not enough to just times batch number at the display part.

#### Testing
All tests passed. In-game testing is also good.

### DIFF
--- a/src/consumption.cpp
+++ b/src/consumption.cpp
@@ -351,7 +351,12 @@ std::pair<nutrients, nutrients> Character::compute_nutrient_range(
         tally_max -= byproduct_nutr;
     }
 
-    int charges = comest.count();
+    int charges = 0;
+    if( rec.makes_amount() == 0 ) {
+        charges = comest.charges;
+    } else {
+        charges = rec.makes_amount();
+    }
     return { tally_min / charges, tally_max / charges };
 }
 
--- a/src/consumption.cpp
+++ b/src/consumption.cpp
@@ -351,12 +351,7 @@ std::pair<nutrients, nutrients> Character::compute_nutrient_range(
         tally_max -= byproduct_nutr;
     }
 
-    int charges = 0;
-    if( rec.makes_amount() == 0 ) {
-        charges = comest.charges;
-    } else {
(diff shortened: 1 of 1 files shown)


src/consumption.cpp
```diff
@@ -351,7 +351,12 @@ std::pair<nutrients, nutrients> Character::compute_nutrient_range(
         tally_max -= byproduct_nutr;
     }
 
-    int charges = comest.count();
+    int charges = 0;
+    if( rec.makes_amount() == 0 ) {
+        charges = comest.charges;
+    } else {
+        charges = rec.makes_amount();
+    }
```
mqrause, line 359:
```suggestion
    int charges = rec.makes_amount();
```

`makes_amount` can't be 0, so this should be enough. For what it's worth it's also what I do in #60885 to fix that, though I didn't even realize it was broken before.

Brambor:
Will test. Did this fix vitamin amounts as well?

PR #63389, opened 2023-02-01
Fix platinum grille recipe

### LINKED ISSUE #56321: Platinum grille requires too much platinum
### Describe the bug

The crafting recipe for a platinum grille requires 100 platinum, which is 5 liters and 1 kg, to electroplate a steel grille with just enough to act as a catalyst in chemical reactions (and another oddity is that it's only used in nitric acid, where straight platinum is used in several other recipes as a catalyst but that's not the point of this issue specifically)

The stated weight difference between the steel grille and the platinum plated one is also only 60 grams, which still seems a bit high to just be a layer of plating

### Steps to reproduce

n/a

### Expected behavior

That a more reasonable amount would be required, especially with the amounts you get from breaking down platinum jewelry being comparatively low and mufflers (standing in for catalytic converters) being even lower

### Screenshots

_No response_

Latest experimental as pulled from Hitchhiker's Guide to the Cataclysm

### Additional context

[Platinum](https://nornagon.github.io/cdda-guide/#/item/platinum_small)
[Platinum grille](https://nornagon.github.io/cdda-guide/#/item/platinum_grille)
[Steel grille](https://nornagon.github.io/cdda-guide/#/item/steel_grille)

Comrade-Soupbeans:
#### Summary
Bugfixes "Reduce amount of platinum to craft platinum grille"

#### Purpose of change
Fixes #56321

#### Describe the solution
Reduce required platinum to 1 unit (5 grams)

Continue requiring half a kilo of platinum to electroplate a catalyst

#### Testing
Edited recipe, loaded the game, value changed

#### Additional context
The final result still gains 60 grams on the steel grille, which should probably also be changed. If no one has any problems with this i will probably do that next

### DIFF
--- a/data/json/recipes/recipe_others.json
+++ b/data/json/recipes/recipe_others.json
@@ -2878,7 +2878,7 @@
     "tools": [ [ [ "electrolysis_kit", 10 ] ], [ [ "surface_heat", 10, "LIST" ] ] ],
     "components": [
       [ [ "steel_grille", 1 ] ],
-      [ [ "platinum_small", 100 ] ],
+      [ [ "platinum_small", 1 ] ],
       [ [ "water", 3 ], [ "water_clean", 3 ] ],
       [ [ "any_strong_acid", 1, "LIST" ] ]
     ]


Profit0004:
This seems reasonable for modern manufacturing methods but is it realistic for the cataclysm? I would assume a catalytic grill would be more like the steel wool in a catalytic converter, and that can be 2-7 grams but does it work in a artisanal method?

YourCompanionCube:
the "artisanal" method is still just electroplating, bad efficiency is going to mostly mean using more electricity, it being slow.

YourCompanionCube:
also worth noting that the grille is only being used for one recipe at the moment and most of the ones that would are just using 1 raw platinum

PR #63390, opened 2023-02-01
Added presets for pocket settings

Skrassh:
#### Summary
Interface "Added Presets for pocket settings"

#### Purpose of change
Added the ability to save and load presets onto pockets in the pocket management window i->i. This makes it much simpler and faster to use the pocket system once you've set up your presets..

#### Describe the solution
I added menu options in the i->i context menu for saving, applying, and deleting presets from a pocket_presets.json file. I used the existing item_pocket::favorite_settings struct while adding a preset_name field so I could use the same serializing functions.

I tried to put the code farther up in inventory to avoid having to load all the presets from the file whenever you used the action but it felt far more complicated that way. I'm open to suggestions on where to move this.

#### Testing
I started a new character, set some whitelist, blacklist and priorities for pockets, Saved the preset, loaded it onto another pocket and then deleted the preset.

#### Additional context
image
image
image

### DIFF
--- a/src/item_contents.cpp
+++ b/src/item_contents.cpp
@@ -396,21 +396,21 @@ bool pocket_favorite_callback::key( const input_context &ctxt, const input_event
         if( !custom_preset_popup.canceled() ) {
             const std::string &rval = custom_preset_popup.text();
             // Check if already exists
-            selected_pocket->load_presets();
-            if( selected_pocket->has_preset( rval ) ) {
+            item_pocket::load_presets();
+            if( item_pocket::has_preset( rval ) ) {
                 if( query_yn( _( "Preset already exists, overwrite?" ) ) ) {
-                    selected_pocket->delete_preset( selected_pocket->find_preset( rval ) );
+                    item_pocket::delete_preset( item_pocket::find_preset( rval ) );
                     selected_pocket->settings.set_preset_name( rval );
-                    selected_pocket->add_preset( selected_pocket->settings );
+                    item_pocket::add_preset( selected_pocket->settings );
                 }
             } else {
                 selected_pocket->settings.set_preset_name( rval );
-                selected_pocket->add_preset( selected_pocket->settings );
+                item_pocket::add_preset( selected_pocket->settings );
             }
         }
         return true;
     } else if( action == "FAV_APPLY_PRESET" ) {
-        selected_pocket->load_presets();
+        item_pocket::load_presets();
         selector_menu.title = _( "Select a preset" );
         for( const item_pocket::favorite_settings &preset : item_pocket::pocket_presets ) {
             selector_menu.addentry( preset.get_preset_name().value() );
@@ -422,7 +422,7 @@ bool pocket_favorite_callback::key( const input_context &ctxt, const input_event
         }
         return true;
     } else if( action == "FAV_DEL_PRESET" ) {
-        selected_pocket->load_presets();
+        item_pocket::load_presets();
         for( const item_pocket::favorite_settings &preset : item_pocket::pocket_presets ) {
             selector_menu.addentry( preset.get_preset_name().value() );
         }
@@ -431,7 +431,7 @@ bool pocket_favorite_callback::key( const input_context &ctxt, const input_event
         if( selector_menu.ret >= 0 ) {
             if( query_yn( _( "Are you sure you wish to delete preset %s?" ),
                           item_pocket::pocket_presets[selector_menu.ret].get_preset_name().value() ) ) {
-                selected_pocket->delete_preset( item_pocket::pocket_presets.begin() + selector_menu.ret );
+                item_pocket::delete_preset( item_pocket::pocket_presets.begin() + selector_menu.ret );
             }
         }
         return true;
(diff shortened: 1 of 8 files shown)


src/item_pocket.h
```diff
@@ -381,6 +385,17 @@ class item_pocket
...
         favorite_settings settings;
 
+        // Pocket presets functions
+        void serialize_presets( JsonOut &json ) const;
+        void deserialize_presets( const JsonArray &ja );
+        void load_presets();
+        void add_preset( const item_pocket::favorite_settings preset );
+        void save_presets();
+        std::vector<item_pocket::favorite_settings>::iterator find_preset( const std::string s );
+        bool has_preset( const std::string s );
```
Qrox, line 395:
```suggestion
        std::vector<item_pocket::favorite_settings>::iterator find_preset( const std::string &s );
        bool has_preset( const std::string &s );
```
Use reference to avoid a copy.

src/item_contents.cpp
```diff
@@ -384,6 +384,54 @@ bool pocket_favorite_callback::key( const input_context &ctxt, const input_event
...
+                }
+            } else {
+                selected_pocket->settings.set_preset_name( rval );
+                selected_pocket->add_preset( selected_pocket->settings );
+            }
+        }
+        return true;
+    } else if( action == "FAV_APPLY_PRESET" ) {
+        selected_pocket->load_presets();
+        selector_menu.title = _( "Select a Preset" );
```
Night-Pryanik, line 411:
```suggestion
        selector_menu.title = _( "Select a preset" );
```

src/item_contents.cpp
```diff
@@ -384,6 +384,54 @@ bool pocket_favorite_callback::key( const input_context &ctxt, const input_event
...
+        return true;
+    } else if( action == "FAV_DEL_PRESET" ) {
+        selected_pocket->load_presets();
+        for( const item_pocket::favorite_settings &preset : selected_pocket->pocket_presets ) {
+            selector_menu.addentry( preset.get_preset_name() );
+        }
+        selector_menu.query();
+
+        if( selector_menu.ret >= 0 ) {
+            if( query_yn( _( "Are you sure wish to delete preset: %s" ),
```
Night-Pryanik, line 429:
```suggestion
            if( query_yn( _( "Are you sure you wish to delete preset %s?" ),
```

Skrassh:
Not sure how to fix this last clang tidy error, any recommendations?

src/item_pocket.cpp
```diff
@@ -2113,6 +2114,84 @@ void item_pocket::heat_up()
...
 
+void item_pocket::add_preset( const item_pocket::favorite_settings &preset )
+{
+    pocket_presets.emplace_back( preset );
+    save_presets();
+}
+
+void item_pocket::save_presets() const
+{
+    cata::ifstream fin;
```
Qrox, line 2125:
```suggestion
```
Does not seem to be used.

src/item_pocket.h
```diff
@@ -381,6 +385,17 @@ class item_pocket
...
+        // Pocket presets functions
+        void serialize_presets( JsonOut &json ) const;
+        void deserialize_presets( const JsonArray &ja );
+        void load_presets();
+        void add_preset( const item_pocket::favorite_settings &preset );
+        void save_presets() const;
+        std::vector<item_pocket::favorite_settings>::iterator find_preset( const std::string &s );
+        bool has_preset( const std::string &s );
+        void delete_preset( std::vector<item_pocket::favorite_settings>::iterator iter );
+        std::vector<item_pocket::favorite_settings> pocket_presets;
```
Qrox, line 397: To my understand this is supposed to be a global value rather than a per-pocket value? If so you should make it static to avoid wasting memory and triggering the clang-tidy warning.

src/item_pocket.cpp
```diff
@@ -2360,6 +2450,7 @@ void item_pocket::favorite_settings::info( std::vector<iteminfo> &info ) const
         info.emplace_back( "BASE", string_format(
                                _( "Items in this pocket <bad>won't be unloaded</bad> unless you manually drop them." ) ) );
     }
+    info.emplace_back( "BASE", string_format( _( "Preset Name: %s" ), preset_name ) );
```
Qrox, line 2453: This should be hidden if there is not a preset name. And since the preset name is optional it's better to change its type to `cata::optional<std::string>` to avoid handling the special case of empty string.

src/item_pocket.cpp
```diff
@@ -2450,7 +2451,10 @@ void item_pocket::favorite_settings::info( std::vector<iteminfo> &info ) const
         info.emplace_back( "BASE", string_format(
                                _( "Items in this pocket <bad>won't be unloaded</bad> unless you manually drop them." ) ) );
     }
-    info.emplace_back( "BASE", string_format( _( "Preset Name: %s" ), preset_name ) );
+    if( preset_name.value() != "" ) {
```
Qrox, line 2454:
```suggestion
    if( preset_name.has_value() ) {
```

src/item_pocket.cpp
```diff
@@ -2137,6 +2217,7 @@ units::volume pocket_data::max_contains_volume() const
 
 void item_pocket::favorite_settings::clear()
 {
+    preset_name = std::string();
```
Qrox, line 2220:
```suggestion
    preset_name = cata::nullopt;
```

PR #32454, opened 2019-07-17
[READY][Feature] Craftable Acetylene

kasanryukin:
#### Summary
```Summary: Content "Adds in the framework for creating calcium carbide as a precursor to acetylene as referenced by #32441"```

Closes #32441

#### Purpose of change
As proposed in #32441, this implements the crafting chain and associated tools, hardware, and furniture items in order to take raw materials (coal, limestone, water, and oxygen), and eventually end up with oxyacetylene for use with acetylene torches.

#### Describe the solution
Creation of a new furniture item, The Arc Smelter, in the c++ code was the first step in the process, since the creation of the first catalyst, calcium carbide, requires heat exceeding that which is capable by ordinary means available to the survivor. (An arc welder/welding rig could do it, and might be modified in the future to add a secondary recipe for small batch production.)   With furniture in hand, the rest was a series of json additions everywhere to build up the crafting chain, and add in necessary alterations in order to make everything work.  A list of changes follows:

- The Arc Furnace. A stationary furniture item that turns electricity into heat to make calcium carbide from water and quicklime
- Oxygen Cylinders.  High capacity sources of raw oxygen found in labs, and anywhere welding supplies can spawn.  Also modified the existing emergency oxygen tanks to also work as a source of oxygen.
- Tool additions.  Added the Acetylene-gas Machine as outlined in #32441. Modified the hand-pump to have a new quality, "PRESSURIZATION", to facilitate taking unpressurized "gas" and pressurizing it into usable fuel.

#### Describe alternatives you've considered
The only alternate I considered was creating another furniture item for the creation of pressurized fuel tanks, taking the three base components, acetylene, oxygen, and an empty tank, and spitting out a full tank of oxyacetylene.  This proved to be cumbersome and somewhat outside my ability to code most due to unfamiliarity with the c++.  Ultimately, I realized the same effect could be achieved using a simple compression style pump such as a bellows or a bicycle pump, and some behind the scenes wizardry with rubber hoses to achieve the same effect.  Most modern tanks have an internal check valve on the fill port to prevent back pressure, and a bellows has a diaphragm (while a bicycle pump is a sealed piston) to prevent fluid from escaping, so a positive pressure can be applied to slowly pump in the gasses to the tank under pressure.

#### Additional context
See #32441 for discussion.

Issue #32441, opened 2019-07-16
Craftable Acetylene


kevingranade:
**Describe the solution you'd like**  
After some digging, it looks like acetylene is eminently craftable, so I'm dropping my notes in case someone gets the urge to implement it.
Following the guide here: https://panam1901.org/miscbuildings/acetylene/acetylene_article.htm
Crafting Calcium Carbide (CAC2) requires:
`` Ground coke and lime are intimately mixed in the proper proportions and placed in the electric furnace-, the result is that fifty-six parts of lime and thirty-six of coke will make sixty-four parts calcium carbide and liberate twenty-eight parts carbon monoxide.``
The furnace is heated to ~4,500F, so it's a pretty serious furnace.
The resulting ingot is relatively stable except in contact with water, so it can be handled more or less normally with just a sealed container or even a wrapping.
So materials list:
Heavy-duty furnace
Coke
Lime
From there, generating and capturing Acetylene via CAC2 / H2O reaction is relatively straightforward as outlined here:
https://chestofbooks.com/home-improvement/repairs/Mechanics-Household/Types-Of-Acetylene-Generators.html
This part is potentially more involved as it involves some fairly heavy looking metalwork, but it results in a fairly renewable acetylene source for welding, and perhaps for other uses such as lighting and heating.

DeltaEpsilon7787:
Honestly, even if you do manage to synthesize acetylene, mixing it with oxygen is an entirely different matter altogether.

The best substitute by far would probably be an electrolysis cell with consumable electrodes to produce a perfect oxyhydrogen mix that works just as well as oxyacetylene. The torch itself probably doesn't care.

l29ah:
There also needs to be a way to pressurize the gas to fill the existing tank items. On the other hand, the torch can be hooked to the acetylene generator that consumes water and CaC2 without intermediate gas storage (but it's less controllable, and you need to make sure the gas generator doesn't explode due to underutilization).

l29ah:
Acetylene is mixed with oxygen in the torch itself. It's impossible to store acetylene and oxygen premixed, since they explode on contact.

DeltaEpsilon7787:
>Acetylene is mixed with oxygen in the torch itself.

Then somehow you need to connect an oxygen tank and an acetylene tank together. Guess something like a railgun that consumes both battery power and a rail rod, but with oxygen and acetylene.

DeltaEpsilon7787:
[...]

PR #63392, opened 2023-02-01
Calcium carbide production overhaul

cake-pie:
#### Summary
Bugfixes "Calcium carbide production overhaul"

#### Purpose of change
The production of calcium carbide by arc furnace as implemented in #32454 is very much a rough first prototype. C++ was adapted from charcoal kiln, unfinished CaC<sub>2</sub> was basically a clone of unfinished charcoal, [grinding requirements missing](https://github.com/CleverRaven/Cataclysm-DDA/pull/32454#discussion_r307986772), etc. Lacks scientific rigor.

Process yields 50±10% by volume compared to inputs, that's pretty much copied from charcoal firing and has no relation to the stochiometry of calcium carbide production CaO + 3 C -> CaC<sub>2</sub> + CO. 

#### Describe the solution

Calcium carbide's stats, like many chems, is standardized to 1 unit = 0.01 mole of the chemical. That's great, we'll start from that.

We have "calcium carbide premix", which is coke and lime that have been ground to fine powder and combined in the right proportion to be smelted into CaC<sub>2</sub>. For the sake of sanity in calculations and implementation, we'll set it equivalent to the end product, i.e. absent any loss/wastage, one unit of premix becomes one unit of CaC<sub>2</sub>. Do a bunch of math to determine what the weight / volume stats should be.

Adjusted the recipe for grinding coke and lime into calcium carbide premix to get inputs and outputs to match the required proportions as closely as possible. Make it require mortar and pestle for grinding and not just hammering.

Created a recipe that uses "electric arc furnace" `f_arc_furnace`; energy requirement for the reaction is 460 kJ per mole (plus overheads and wastage, 70% efficiency is generous, [source](http://dx.doi.org/10.1016/j.energy.2009.04.015) reports 40~75%). The smelting creates molten CaC<sub>2</sub> which is allowed to cool and solidify; this then has to be crushed and ground down to the powdered form that is `chem_carbide`.

Obsoleted "unfinished calcium carbide" which is what you would have found if you smashed apart an arc furnace after firing it up but before it is done. This was created when shoehorning the charcoal firing process for carbide production and no longer needs to be a thing. Delete cac2powder and unfinished_cac2 ammo types, and cac2powder material, which were "shoehorns".

For now this PR disables the `iexamine` function for `f_arcfurnace_empty` so the old process can no longer be started, but any savegames with the process underway can still get the produced carbide from the `f_arcfurnace_full` (after required time has elapsed), which turns the furniture back into a `f_arcfurnace_empty`. Will leave it like this for a transition period, after which the "arc furnace" (`f_arcfurnace_empty`, `f_arcfurnace_full`) and its `iexamine` functions will no longer be used at all and are removed as dead code.

Have not decided what to do with the furniture after its obsolescence. It is kind of redundant with `f_arc_furnace`, so one thought is to migrate/obsolete it. However, `f_arc_furnace` cannot be constructed, whereas the `f_arcfurnace_empty` can, with some special requirements like fire bricks. It is not clear to me if it is okay to merge these into one furniture def, or whether they should continue to exist separately, filling different gameplay roles.

For now, I've given `f_arcfurnace_empty` the same `fake_arc_furnace` as `f_arc_furnace`, so players who went to the trouble of constructing the former -- which existed specifically for making CaC<sub>2</sub> -- aren't left without a means of doing so.

1
Industrial processes also involve "electrode paste composed of coal tar pitch binder and anthracite coal" <small>[[source](https://www3.epa.gov/ttnchie1/ap42/ch11/final/c11s04.pdf)]</small> but we're going with a more artisanal approach per sources cited in #32441.
&nbsp;

2
The carbide smelting process liberates carbon monoxide, IRL this may be piped directly to other industrial process (e.g. lime kiln) that are colocated with the carbide arc furnace to take advantage of this. We could certainly come up with uses for carbon monoxide in-game -- there are at least a couple of recipes for extraction of metal from ores that could have a variation that uses CO.

One way to do it is to have a carbon monoxide item that is produced as by-product, but that requires the player to have some additional means to capture and bottle it. If the by-product can be made on an opt-in basis (or otherwise discarded), that would be more palatable than forcing players to have extra equipment to collect CO that they don't want just in order to make carbide.

Another idea is to add machinery in C++ to support having some kind of co-recipe that defines a co-process that can be crafted at the same time to take advantage of synergies. E.g. while making carbide, we can take the liberated carbon monoxide to reduce `chem_zinc_oxide` to `chem_zinc_powder`, perhaps even leverage some waste heat to lower the amount of fuel required.

#### Testing
Test the recipes for mix/grind coal+lime into premix and smelt/grind premix into carbide.

#### Additional context
Assumes that `coal_lump` has mass 37.5g per unit and not 1.5g -- this is bugfixed by 79cd0dcaf52a647867541fe98220dcd094f59ceb in #63375, no need to duplicate it here.

> Fix weight of `coal_lump`, from 1.5g to 37.5g. There was a mistake in 5549b62bdb85fdd9eab3f18d1e39a75c2ade1963 where divided by count (250) instead of stack size (10) leading to value off by a factor of ×25.


kasanryukin:
Just saw this when reviewing my old commits!  Much more elegant that what I did.  What I mocked up worked with the existing code (in a heavy handed way), but I always wanted to give it a second pass. I'm glad somebody did and did it well!

PR #63395, opened 2023-02-01
Show radiation level using radiation badge instead of character's rads

### LINKED ISSUE #63283: Radiation badges aren't supposed to reset. But in the game they do, sorta.
### Describe the bug

Some sidebars (custom, labels, labels narrow, thick and probably others I forgot) have a "Radiation" field. If you put on a radiation badge, this field shows the color of its film strip, which is very convenient.

However, the radiation badge isn't supposed to reset. If it was a brand new green-colored badge and you got a dose of radiation, it will turn yellow and *stay* yellow, unless you get even more radiation and it will go red/black. Bottom line, it will never become green again.

In the game, radiation badge (the item) doesn't reset. But the sidebar color does.

Here is my character's radiation badge - you can see it says "orange" at the bottom.

image

The rad exposure happened some time ago, and after a lot of sleep, water and Prussian blue tablets some of that radiation went away. The badge still says "orange", but the sidebar says "blue"

image

Tested it on several other sidebar layouts - everything that has "Radiation" field works the same. They don't show the badge's film stip color - they show your current radiation level instead.

### Steps to reproduce

1. Put on a radiation badge
2. Bathe in radiation until it changes colour
3. Walk away from the source of radiation and eat a dozen Prussian blue tablets
4. Watch how your sidebar "Radiation" field changes colour to less severe (yellow, then blue, then green) while your badge's colour remains orange/red/black

### Expected behavior

I expect the sidebar Radiation field to show the current colour of the badge

### Screenshots

_No response_

OS: Windows
OS Version: 10.0.22000.1455 (21H2)
Game Version: https://github.com/CleverRaven/Cataclysm-DDA/commit/0808fe73d76702995326fd16114e47aa60848a7d [64-bit]
Graphics Version: Tiles
Game Language: English [en]
Mods loaded: [
Dark Days Ahead [dda],
Bionic Slots [cbm_slots]
]

### Additional context

_No response_

dseguin:
#### Summary
None

#### Purpose of change
- Fixes #63283

#### Describe the solution
Determine the state of the sidebar radiation indicator using a worn or held radiation badge instead of using the character's current rad level.

#### Testing
Getting exposed to radiation while wearing the badge increases the radiation of the badge. Setting the player's rads to 50 (down from 180) doesn't affect the sidebar indicator, which shows as RED because the badge's irradiation is still 180:

rad_badge

value in memory

### DIFF
--- a/data/json/ui/radiation.json
+++ b/data/json/ui/radiation.json
@@ -10,7 +10,7 @@
         "color": "white_green",
         "condition": {
           "and": [
-            { "compare_int": [ { "u_val": "rad" }, "<=", { "const": 0 } ] },
+            { "compare_int": [ { "u_val": "item_rad", "flag": "RAD_DETECT", "aggregate": "min" }, "<=", { "const": 0 } ] },
             { "or": [ { "u_has_worn_with_flag": "RAD_DETECT" }, { "u_has_wielded_with_flag": "RAD_DETECT" } ] }
           ]
         }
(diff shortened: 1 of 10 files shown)


cake-pie:
Thanks for prompt attention!

One thought: if the character can have multiple badges, would it be more useful to show the highest or lowest color?
- highest = the worst you (or rather, that badge) has previously (recently?) been exposed to, vs
- lowest = the worst things could be *right now* if you don't have other ways to take a measurement

There's utility in the information offered in either case.

dseguin:
Instead of just getting the lowest rad value, I've added some infrastructure to choose any of these aggregate values:
- min
- max
- first
- last
- sum
- average

In this case I'm using `min`, so if the player is wearing more than one radiation badge the sidebar will show the state of the lowest one
```JSON
"compare_int": [ { "u_val": "item_rad", "flag": "RAD_DETECT", "aggregate": "min" }, "<=", { "const": 120 } ]
```

PR #57795, opened 2022-05-19
Add the feral burner

MNG-cataclysm:
#### Summary
Content "Add the feral burner"

#### Purpose of change
Add a new feral military soldier, the feral arsonist, and have it spawn with other feral soldiers.

#### Describe the solution
Adds a new monster, the feral arsonist. A military-clad zombie which holds a flamethrower and knows how to use it, the arsonist will spawn alongside zombie burners and feral soldiers wherever they occur. It is fairly uncommon, however.

#### Describe alternatives you've considered
Not adding the arsonist

#### Testing
Spawning the arsonist works as intended, they operate their weapon properly, and spawn in their monster groups.

#### Additional context
To prevent the random burning of buildings, the feral burner is required to target players before using their flamethrower.

PR #63396, opened 2023-02-02
Add the feral burner

MNG-cataclysm:
#### Summary
Content "Add the feral burner"

#### Purpose of change
Add a new feral militia soldier with a flamethrower, and have it spawn at mass graves starting ten days after the Cataclysm.

#### Describe the solution
Adds a new monster, the feral burner, as described above. It is fairly uncommon in mass grave sites when it does spawn.

None.

#### Testing
Everything works as intended. The monster spawns properly and attacks like it's supposed to.

#### Additional context
This is a reopening of #57795, which I was instructed to open again after it was accidentally merged.

### DIFF
--- a/data/json/monsters/feral_humans.json
+++ b/data/json/monsters/feral_humans.json
@@ -757,7 +757,7 @@
         "fake_per": 10,
         "ranges": [ [ 0, 6, "DEFAULT" ] ],
         "require_targeting_player": false,
-        "description": "The feral burner fires its' flamethrower."
+        "description": "The feral burner fires its flamethrower!"
       }
     ],
     "luminance": 8,


anoobindisguise:
flamethrower feral is probably a no go for the same reason that flamethrower zombies were removed, because the US military doesn't use flamethrowers. Why not do like, an acetylene torch feral that's still super dangerous in melee, but has a poor vision range and won't murk the player instantly due to its flamethrower?

MNG-cataclysm:
> flamethrower feral is probably a no go for the same reason that flamethrower zombies were removed, because the US military doesn't use flamethrowers. Why not do like, an acetylene torch feral that's still super dangerous in melee, but has a poor vision range and won't murk the player instantly due to its flamethrower?

I did some description revamping to imply that the flamer is/was not part of the U.S military, but part of an independent militia. As for the danger, I figured ten days would be enough time for a player to obtain a gun of some sort which can outrange the flamers' attack distance of six tiles.

anoobindisguise:
> > flamethrower feral is probably a no go for the same reason that flamethrower zombies were removed, because the US military doesn't use flamethrowers. Why not do like, an acetylene torch feral that's still super dangerous in melee, but has a poor vision range and won't murk the player instantly due to its flamethrower?
> 
> I did some description revamping to imply that the flamer is/was not part of the U.S military, but part of an independent militia. As for the danger, I figured ten days would be enough time for a player to obtain a gun of some sort which can outrange the flamers' attack distance of six tiles. I could go up to fifteen days in before it spawns if that'd balance it better.

But... how did they get the military flamethrower in the first place, how did they not use up all their fuel already after going feral, why are they still hanging around the mass grave, why did they go feral only after getting the flamethrower (everyone who was "chosen" by the blob to go feral started doing so just before the zombies really got going, how did the feral pick out a flamethrower and go out to burn stuff? by the time they had the circumstances to do that, everyone who could already went feral). Why do they only show up at the mass grave after 10+ days, despite "deploying" at the start of the cataclysm? There's no way that can be anything other than a highly illogical mercy/balance feature in favor of the player.

TheShadowFerret (commented):
I agree with @anoobindisguise that this is not justified to exist, however no failing your grammar 😛

MNG-cataclysm:
> But... how did they get the military flamethrower in the first place, how did they not use up all their fuel already after going feral, why are they still hanging around the mass grave, why did they go feral only after getting the flamethrower (everyone who was "chosen" by the blob to go feral started doing so just before the zombies really got going, how did the feral pick out a flamethrower and go out to burn stuff? by the time they had the circumstances to do that, everyone who could already went feral). Why do they only show up at the mass grave after 10+ days, despite "deploying" at the start of the cataclysm? There's no way that can be anything other than a highly illogical mercy/balance feature in favor of the player.

I’ll address these in order:
1. The zombie obtained the flamethrower through military surplus or some other manufacturer or distributor. You can buy flamethrowers online.
2. They just didn’t have time to operate the flamethrower in full before getting zombified. Probably getting mobbed and managing to escape before turning, or getting the flamethrower as they were slowly going feral.
3. They either got it shortly before or didn’t have time to operate it before going feral. That’s why they still have it.
4. They didn’t “deploy” at the start of the Cataclysm, but they formed up with a militia shortly after, and went feral that way.

GuardianDll:
The main question that should be asked is "why it stipp has some ammo in his tank, as it will probably be gone in the next few hours after the person goes mad?"

MNG-cataclysm:
> The main question that should be asked is "why it stipp has some" ammo" in his tank, as it will probably be gone in the next few hours after the person goes mad?"

It would still have ammo on the off chance it didn’t shoot anything while turning, as it would lose its reason to fire on the zombies.

kevingranade:
As outlined by others, flanethrowers are not pervasive in the setting, so this is a no go.
There are other issues,  but that is the insurmountable one.

PR #63400, opened 2023-02-02
Fix: activating a pill container consumes the whole pill stack

### LINKED ISSUE #62997: Using meds via activating their container consumes the whole stack
### Describe the bug

I am quite new and don't know the proper names for the different menus/actions, so I will also refer to them with their default keybindings.

When activating a container (i -> select the container -> RETURN -> a) which holds certain meds, namely pills, one of the listed actions is to consume said med. _Doing so will consume the entire stack of the med instead of one portion,_ contrary to what is shown in the activation menu (where it says -1).

This effects meds which do not have a specific target, such as pills and saline solution. Meds which are then applied to limbs, such as bandages and antiseptic, are not prone to this.

Selecting and activating the med from the inventory directly, or using the consumption menu (E) still works fine.

I happen to be playing on the Dec 27 experimental build, but I didn't see anything like this addressed in the changelogs since then. Please correct me if I'm wrong.

### Steps to reproduce

1. Pick up any container holding multiple aspirin
2. Open the inventory ( i )
3. Highlight the container
4. Select it (RETURN)
5. Activate (a)
6. Take the consume action 
The container will no longer have any aspirin, no matter how many it contained originally

### Expected behavior

I expect only one aspirin to be consumed, as the action is described with a -1 ( ex. Consume aspirin(8)(-1) )

### Screenshots

_No response_

OS: Windows 10 x64
Game Version: Experimental 2022-12-27-0658
Graphics Version: Tiles
Language: English
Mods: dda, no_npc_food, personal_portal_storms, no_fungal_growth, package_bionic_professions

### Additional context

I know the consume menu is better, still trying to break noob habits

emomicrowave:
#### Summary
Bugfixes "Fix a bug where activating a medical container and using pills would remove all the pills from the container"

#### Purpose of change

Fixes #62997

#### Describe the solution

Not sure it's an elegant solution, but it's an easy one and it will impact the least amount of other items and actions. 

Basically activating the pill directly, and activating the container are wildly different flows in the code. So, an easy fix was to add an additional check in the `item::ammo_consume( )` function, so that if it encounters a medical item with charges, it should substract the used charges. This works, because consuming an e.g. "aspirin" directly, doesn't even land you in the `ammo_consume` function, so this probably only for containers with usable meds.

As already mentioned, I wanted to find a common function, that is accessed by both (a)ctivating a container full of pills, and (a)ctivating the pill itself, and do some redirection there. But I went as far back as `do_regular_action()` in `handle_action.cpp`, but even this isn't a function that is triggered when activating a container.

Another idea was, similarly to drinks, to make the container unusable, but I wouldn't know where to start with this, and it would also break items like the First Aid Kit, which contain many different meds that are hidden in the inventory.

#### Testing

I spawned some medical items such as aspirins, and oxycodone, both inside and outside of containers. The oxycodone test made sure that activating the container still applies the effect of the drug. I also got a first aid kit from the Evacuation Center, and used up all the contents one-by-one. Finally, I (a)ctivated the jeans where all the pill containers were stored, and this also worked.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -10959,6 +10959,13 @@ int item::ammo_consume( int qty, const tripoint &pos, Character *carrier )
         qty -= contents.ammo_consume( qty, pos );
     }
 
+    // Dirty fix: activating a container of meds leads here, and used to use up all of the charges.
+    if( is_medication() && charges > 1 ) {
+        int charg_used = std::min( charges, qty );
+        charges -= charg_used;
+        qty -= charg_used;
+    }
+
     // Some weird internal non-item charges (used by grenades)
     if( is_tool() && type->tool->ammo_id.empty() ) {
         int charg_used = std::min( charges, qty );


emomicrowave:
The functionality that I touched should affect the failing test in `health_test.cpp` at all. It seems to also fail in other PRs.

mqrause:
I didn't investigate this bug, but if liquids aren't affected, #60885 most likely already fixes it by making pills not be charges anymore. Might be worth trying out, unless people want this to still go into 0.G stable.

emomicrowave:
Just did a quick test, and it seems liquids are affected as well. Tried with the `cough syrup` for example, and this PR also fixes that.

PR #28958, opened 2019-03-24
Add a methlab basement, new palette for residential basements

anothersimulacrum:
#### Summary
```SUMMARY: Content "Adds basement with a hidden methlab"```

#### Purpose of change
The game was previously lacking a methlab basement. This aims to fix it.

#### Describe the solution
Add a few new itemgroups, a new palette, and a new basement. Two varieties, one with methlab, and one without.

#### Additional context
I might clean up the `methlab` item group and some itemgroup stuff in a few later PR, and I plan on adding some more residential basements.
Pictures:
LivepeopleLab
RetrodaysLab
ASCIILab
LivepeopleNoLab
RetrodaysNoLab
LivepeopleNoLab2
RetrodaysNoLab2

PR #63403, opened 2023-02-02
Add a drug lab

MNG-cataclysm:
#### Summary
Content "Add a drug lab"

#### Purpose of change
Add a drug lab, as we do not have a stationary drug lab that can spawn outside of cities.

#### Describe the solution
Add a drug lab, based off of the moonshine stills' layout, Zombies can spawn there, and it contains meth, chemicals, and lab equipment.

Making multiple drug labs dedicated to one kind of drug, such as one for weed, meth, opioids, etc.

#### Testing
Spawned the drug lab, went to it, killed the inhabitants, and died to a meth overdose.

#### Additional context
Jessee, if you're reading this, we need to cook.

Screenshots:
image
drug lab 1


data/json/mapgen/drug_lab.json
```diff
@@ -0,0 +1,50 @@
+[
+  {
+    "method": "json",
+    "om_terrain": "drug_lab",
+    "type": "mapgen",
+    "weight": 100,
```
Night-Pryanik, line 6: Since at the moment there's only one variant of this location, this field effectively ain't doing anything useful right now. Not technically an error, just a nitpick.

Night-Pryanik:
Some screenshots?

JonathanLochridge:
Interesting. It seems it tends to spawn with random chemicals?

From what I can tell the "methlab" itemgroup also throws in some topical items.
I couldn't manage to find where that item group is actually defined though.

I noticed you mentioned making drug-specifc drug labs as an alternative. I overall agree that seperate maps for drug types probably isn't necessary.
Although, I do think having some variance built into a single map is a good idea though.
The simplest way I can think of would be to have the same map. But, have multiple loot tables flavored to a particular drug type. Then have it pick which one to use at certain rates.

Although, even adding just one in is interesting.

Maybe adding a "drug stash" item group to the loot tables might be a good idea?

FarkasKBence (commented):
Game loads without errors, not seeing any mapgen errors either.

Just about the only "large scale" suggestion I have is that this misses a roof.
I'd also recommend replacing `place_monsters` with `place_monster` in case the dice roll the wrong way. Also, having hazmat zombies spawning here instead of ordinary ones would stop children or dogs from spawning here.

data/json/overmap/overmap_special/specials.json
```diff
@@ -4437,6 +4437,16 @@
...
   },
+  {
+    "type": "overmap_special",
+    "id": "drug_lab",
+    "overmaps": [ { "point": [ 0, 0, 0 ], "overmap": "drug_lab_north" } ],
+    "locations": [ "forest" ],
+    "city_distance": [ 20, 60 ],
+    "city_sizes": [ 8, -1 ],
+    "occurrences": [ 0, 2 ],
+    "flags": [ "CLASSIC", "WILDERNESS", "MAN_MADE" ]
```
anothersimulacrum, line 4448: Numbers pulled a bit out of my ass, but this should be rare.
```suggestion
    "occurrences": [ 1, 5 ],
    "flags": [ "UNQIUE", "CLASSIC", "WILDERNESS", "MAN_MADE" ]
```

anothersimulacrum:
> Add a drug lab, as we didn't/do not have a stationary drug lab.

We do, but it's just _very_ rare https://github.com/CleverRaven/Cataclysm-DDA/pull/28958

PR #58181, opened 2022-06-05
Added avgas to list of valid fuel choices for gasoline engines

Ilurker:
#### Summary
Balance "Make avgas burnable by cars"

#### Purpose of change
Avgas is trade name for a variety of leaded, high octane gasoline which is usable in standard gasoline engines, but the game does not treat it as a valid fuel type. While avgas certainly isn't good for a typical engine, it's not immediately harmful and wouldn't produce issues across the typical lifespan of a CDDA player, much less their vehicles - much less than burning rendered spider fat would cause a humvee, for instance. The primary variety of avgas in usage today appears to be 110LL, which is much lower lead than typical. There are many varieties of unleaded avgas on and coming to market (UL102 and G100UL), and in the "near future" context with such a relatively high prevalence of solar polar, it is not outlandish to assume that most avgas is of these "green" varieties.

#### Describe the solution

Added avgas to the fuel options for gasoline engines

#### Describe alternatives you've considered

Finding lore reasons for players to adhere to the environmental prohibition on burning leaded fuel in standard engines.
Adding catalytic converters which must be removed to utilize avgas.
Having avgas and other makeshift fuel slowly damage engines.

#### Testing

Spawned a gasoline-powered vehicle and replaced its fuel with avgas. Verified that the vehicle was able to drive with avgas as its only fuel source and that the fuel selection dialogue existed and recognized avgas as an option.

#### Additional context

https://en.wikipedia.org/wiki/Avgas
https://www.ainonline.com/aviation-news/business-aviation/2016-03-30/faa-moves-next-phase-unleaded-avgas-testing
https://web.archive.org/web/20170415200334/https://www.lycoming.com/node/18093
https://bobistheoilguy.com/forums/threads/use-avgas-in-car.19939/
https://classicmotorsports.com/articles/fuel-facts-why-not-avgas-your-car/

PR #63416, opened 2023-02-04
Make avgas useable as gasoline

waylon531:
#### Summary
Balance "Make avgas usable as gasoline"

#### Purpose of change

This patch makes avgas more similar to gasoline. Compared to normal gasoline, avgas is higher octane and leaded, but still usable in cars. It will destroy your catalyctic converter after enough time, but if you're in the apocalypse there's no reason to care about car emissions and this part isn't modeled anyways in CDDA, so its not an issue.

#### Describe the solution

Edit the json entry for avgas, to make it look more like gas, and for combustion engines, to let them use avgas as fuel.

Avgas is colored the same as gasoline, to prevent confusion. However, it could be colored light blue as it's dyed this color in real life.

#### Testing

Tested in debug mode, I spawned a car in, and tested it with both normal gas and avgas.


RenechCDDA:
Per #58181 avgas should not be useable in gasoline engines.

waylon531:
Ah, dang. It's incredibly simple to remove a catalytic converter so it really shouldn't be an issue in game, but if you want to close it I guess I'll start making a mod/fork.

ZhilkinSerg:
Duplicate of https://github.com/CleverRaven/Cataclysm-DDA/pull/58181

PR #63418, opened 2023-02-04
Hallucinating taming attempts / success on hallucinated animals

cake-pie:
#### Summary
Content "Hallucinating taming attempts / success on hallucinated animals"

#### Purpose of change
Rather than simply have hallucinated animals always disappear upon taming attempts, have it possible to hallucinate the taming process and success.

#### Describe the solution
Only one in four chance that taming attempt will cause the animal to disappear, revealing that was hallucinated.
Rest of the time, all UI outputs to the player will be exactly what is expected from the taming process. However, the petfood is not consumed, only silently dropped to the ground without any feedback to the player. You might miss it if you aren't paying attention!

Also put in similar support for hallucinated interaction with hallucinated dogthing, that may or may not be a thing that could happen naturally in the game without debug intervention, but now we have the code to handle it.

#### Testing
Debug spawn hallucinated animal and corresponding petfood, try taming and not have it immediately disappear all the time.

### DIFF
--- a/src/iuse.cpp
+++ b/src/iuse.cpp
@@ -1608,7 +1608,9 @@ cata::optional<int> iuse::petfood( Character *p, item *it, bool, const tripoint 
             return cata::nullopt;
         }
 
-        if( mon->is_hallucination() ) {
+        bool halluc = mon->is_hallucination();
+
+        if( halluc && one_in( 4 ) ) {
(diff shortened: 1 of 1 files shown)


Raikiri:
what happens to the petfood that you use?

cake-pie:
> what happens to the petfood that you use?

> the petfood is not consumed, only silently dropped

PR #63420, opened 2023-02-04
fix skull strapped to head coverage

kevingranade:
#### Summary
None

#### Purpose of change
This item came up in discussion on discord,  it had slipped by me that this was made to be an effective piece of armor instead of a piece of a costume,  this does a small tweak to fix that. 

#### Describe the solution
Reduce the coverage to reflect that strapping a wolf skull to the top of your head is not going to block anything from the sides. 

It could happily go away instead,  but this at least keeps it from being something worth pursuing, which matches the original intent.

### DIFF
--- a/data/json/items/armor/helmets.json
+++ b/data/json/items/armor/helmets.json
@@ -734,7 +734,7 @@
     "color": "dark_gray",
     "material_thickness": 3,
     "flags": [ "WATER_FRIENDLY" ],
-    "armor": [ { "encumbrance_modifiers": [ "IMBALANCED" ], "coverage": 85, "covers": [ "head" ] } ]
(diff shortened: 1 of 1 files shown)


anoobindisguise:
coverage might even arguably be lower. Of the head locations of ears, crown, nape, neck and forehead it would cover what, the forehead, maybe a bit of the crown? Although it could arguably also cover eyes/mouth. Using sublocations that might be as little as 25% if we're generous and say 100% of the forehead and 25% of the crown.
canineSkull05_sculpt_725_MAIN

kevingranade:
It says the top of the head, which is more sane than trying to use it as a mask. I'm not too concerned with getting the coverage precisely correct, it's trash tier which is the critical piece.

PR #63428, opened 2023-02-04
Liquid bandage spray

cake-pie:
#### Summary
Content "Liquid bandage spray"

#### Purpose of change
:warning: Requires #63434, which was split off from this PR.

Demonstrate/leverage new feature added in #63434.

#### Describe the solution
Adds a new item, liquid bandage spray, a healing tool that uses charges.

- [x] todo: add to itemgroups, spawn the item where appropriate

#### Testing
Debug damage self, debug spawn liquid bandage spray, use liquid bandage spray.


Shodan14:
Would be cool to have some ad hoc wound wound supergluing as well down the line.

cake-pie:
> Would be cool to have some ad hoc wound wound supergluing as well down the line.

Only if `superglue` is actually cyanoacrylate. The item description "a tube of strong glue" kind of suggests it might be, but there is also a recipe to make it which definitely rules it out.

Shodan14:
> > Would be cool to have some ad hoc wound wound supergluing as well down the line.
> 
> Only if `superglue` is actually cyanoacrylate. The item description "a tube of strong glue" kind of suggests it might be, but there is also a recipe to make it which definitely rules it out.

Yeah, I guess there's a bunch of backend work for that. It really should be a liquid with each use using some amount rather than just "a superglue". Then you could also make it in reasonable amounts.

cake-pie:
No, I meant that the `superglue` that currently exists in game has a recipe that allows it to be made from bones or hide or pine pitch etc, therefore it is definitely not cyanoacrylate.

cake-pie:
I know, I looked at it and decided it was easier to just make a new distinct item for this rather than deal with the problems associated with making superglue into proper cyanoacrylate and giving it the heal action.

bombasticSlacks:
Conflicts with your other PRs where the spray is tucked in the middle of lists. Afterwards, this is g2g.

cake-pie:
Conflicts were expected and have been resolved.

PR #63430, opened 2023-02-05
Ask before lighting fires adjacent to fireplaces

### LINKED ISSUE #57835: Ask player to reconsider setting something on fire when a fireplace is on an adjacent tile

So you do not accidentally set your cupboard (and also potentially your house) on fire when you typo and meant your brazier.

### Solution you would like.

Search adjacent tiles for a thing that can contain fire, e.g a brazier, fire ring and similar. Then ask the player if they really mean to do that. There may also need to be an option to ignore that, so the game does not ask you again for the next N turns or real life minutes.

_No response_

### Additional context

_No response_

ehughsbaird:
#### Summary
Interface "Ask before lighting a fire adjacent to a fireplace"

#### Purpose of change
Resolves https://github.com/CleverRaven/Cataclysm-DDA/issues/57835
Nice quality-of-life feature.

#### Describe the solution
Check for a fireplace that is on any of the 8 adjacent squares if we're lighting a fire on a tile that is not a fireplace. If any we find a fireplace on any of those squares, ask once if they want to start a fire next to a fireplace.

#### Testing
image

1. Find or create two fireplaces next to each other.
2. Place flammable material (e.g. a plank) on a tile adjacent to both fireplaces.
3. Try to set the flammable material on fire.
4. Select no on the prompt, no fire is started.
5. Select yes on the prompt. You are only prompted once and a fire is created.

### DIFF
--- a/src/iuse_actor.cpp
+++ b/src/iuse_actor.cpp
@@ -1204,6 +1204,27 @@ bool firestarter_actor::prep_firestarter_use( const Character &p, tripoint_bub_m
             return false;
         }
     }
+    // Check for an adjacent fire container
+    for( const tripoint_bub_ms &query : here.points_in_radius( pos, 1 ) ) {
+        // Don't ask if we're setting a fire on top of a fireplace
+        // TODO: fix point types
+        if( here.has_flag_furn( "FIRE_CONTAINER", pos.raw() ) ) {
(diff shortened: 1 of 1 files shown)


ehughsbaird:
Whoops

Issue #63345, opened 2023-01-30
use:action { type: heal } does not respect the "SINGLE_USE" flag


Standing-Storm:
### Describe the bug

I'm trying to make a summonable bandage for a spell effect and want it to go away after use. Since comestibles cannot be ethereal and I don't want the player to be able to stockpile psychic bandages, I made it a tool with the heal use_action, but doing so means the item lasts the full duration of the spell and allows multiple bandage actions even though it's set with the flag SINGLE_USE

### Steps to reproduce

1) Create a spell to summon a single-use object, like "Summon Rollmat"
2) Cast the spell, use the summoned item, and notice that the rollmat disappears after use.
3) Copy the rollmat data into a new item and point the spell to that item, change the item's use_action to { "type": "heal", "bandages_power": 8, "bleed": 20, "move_cost": 100 }
4) Cast the spell, notice that you can bandage yourself every turn for as long as the spell duration lasts. 

### Expected behavior

A SINGLE_USE item should be single-use, regardless of what use it's put to.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.2486 (21H2)
- Game Version: 690e38e [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

### Additional context

My original idea was to have the summoned item last a single turn, but that causes the game to crash when it's used, presumably because it disappears and then the avatar tries to use the no-longer-extant item.

cake-pie:
This turned out to be fairly complex.

In most cases "tool-like" use actions return amount/charges consumed from their `use()` as soon as they are (a)ctivated. Consumed amount must be >0 to even reach the point in the code where `SINGLE_USE` is checked for and handled.

The catch here is that `heal_actor::use()` returns 0. because heal use action is not instantaneous. It assigns a first aid activity that takes time to complete, so whatever resources are being used cannot be consumed immediately upon (a)ctivating, but rather only upon completion of the activity -- and there is even a chance that the activity is cancelled or otherwise cannot be completed.

Which is why
> have the summoned item last a single turn,

won't work, because that's not enough time.
[...]

Issue #38698, opened 2020-03-10
Using a bandage with rigid table manners has food-related morale effects


wapcaplet:
# Describe the bug

When using a bandage with rigid table manners, morale effects are applied as if the character had just eaten something.

# Steps To Reproduce

1. Create a character with the "Rigid Table Manners" trait
2. Get some bandages
3. Hurt yourself
4. Use bandages with a nearby table and chair: "Ate with table"
5. Use bandages without a nearbly table and chair: "Ate like an animal"

# Expected behavior

Using bandages should not affect morale as if something were eaten.

# Screenshots

![image](https://user-images.githubusercontent.com/238395/76363300-59fe6380-62e8-11ea-8dbc-167a943a25a1.png)

![image](https://user-images.githubusercontent.com/238395/76363306-5e2a8100-62e8-11ea-911e-75b4b9f8275b.png)

# Additional context

Thanks to Chromosome Kun! on the discord chat for raising this. These morale effects are handled in consumption.cpp in [this section of Character::modify_morale](https://github.com/CleverRaven/Cataclysm-DDA/blob/2f83dbec457ec41ee15b55c05a9d922cf0349178/src/consumption.cpp#L982-L997) but I can't imagine why this would be getting invoked for bandages. I will take a crack at a fix.

KorGgenT:
perhaps the block of code with the table manners trait should exclude comestibles with NO_INGEST?

PR #63434, opened 2023-02-05
Charges / item consumption for non-comestible healing; first aid items from comestible to generic

cake-pie:
#### Summary
Infrastructure "Charges / item consumption for non-comestible healing"

#### Purpose of change
(Split from #63428)

While investigating #63345 I found that the "heal" use action only has code for consumption of comestibles, and doesn't handle anything else properly (nothing is deducted).

#### Describe the solution
If a tool has `charges_per_use`, deduct that much when it is used for healing. -- See #63428 for a demo/test item.

If an item has `SINGLE_USE`, delete the item when done with healing.
Leverage this new functionality to change first aid items that use the "heal" use action from comestibles to generic items. Adjust item stats, spawns/drops, recipes proportionately; put items on charge migration blacklist.

Driveby: two recipes with fake activity level changed to `LIGHT_EXERCISE`

#### Alternatives considered
Continue to have comestibles that are `NO_INGEST` and `EDIBLE_FROZEN`.
Continue to be prone to bugs due to using comestibles for things that aren't actually food. (e.g. #38698)

#### Testing
Debug damage self, debug spawn the first aid items, use them and note that they are correctly used up.
Compare vs prior version to check item stats and affected recipes.
Check professions that start with any of the changed items -- ensure that those that used to have `charges` in the item definition still spawning with correct number of items (now set with `count`)
Scrutinize JSON spawn/drop changes.


mqrause:
Considering these items don't seem to be using any tool mechanics, `generic` might be a better choice of type. Good stuff, though. Should be merged before #60885.

andrei8l:
UX-wise this is pretty awkward as some healing items (ex: bandage) are now in the `a`ctivate menu while others (ex: antiseptic) are in the `E`at menu. IMO you should change `comestible_inventory_preset` and related bits to preserve the old UX. That doesn't have to be in this PR though.

cake-pie:
> don't seem to be using any tool mechanics, `generic` might be a better choice

Hmm, you have a point; I guess I was a bit fixated on tools due to the context behind this (#63345, #63428)

---

> UX-wise this is pretty awkward as some healing items (ex: bandage) are now in the `a`ctivate menu while others (ex: antiseptic) are in the `E`at menu. IMO you should change `comestible_inventory_preset` and related bits to preserve the old UX.

The items changed were already in the `a`ctivate menu to begin with even before this PR, and remain accessible there. This has not changed. Other healing items like antiseptic, also do show up as "apply to wound" in the `a`ctivate menu. This is the logical way to access them.

The reason behind healing items (that are not actually food, or even edible) showing up in the `E`at menu is because of (ab)using comestible type for things that are not food. Seriously, we do not eat bad-aids, bandages, or medical gauze to treat a wound. Rather than trying to "preserve" such illogical UX, I think that we should instead be looking into how other non-food medical items can be taken out of comestibles as well.

andrei8l:
There is a dedicated `Cosume Medication` hotkey unbound by default.

What's logical is not very relevant for UI as people just expect their UX habits to keep working.

Finally, the `a`ctivate menu shows far more than just medication AND doesn't auto-return. So if you want to patch yourself up after a fight, you now need to repeat 6 times 1.`a`ctivate 2. `/`filter bandage 3. apply

andrei8l:
I've noticed that this makes alcohol wipes spawn as one per box. Example from `drugs_heal_simple`

Screenshot from 2023-02-16 18-01-05

cake-pie:
Ah, it needs the first aid kit treatment. Same for cotton balls. Thanks for catching this.

PR #63437, opened 2023-02-05
Hub greaves can be worn with tails

Dan1ss1mo:
#### Summary
Bugfixes "Hub greaves can now be worn with tails"

#### Purpose of change
Last time i checked, greaves are generally pretty far away from the place there tails comes out

#### Testing
can wear graves with tails now


actually-a-cat:
Hub greaves are already compatible with tails, since #61521 only items that cover hips conflict with tails.
image

Dan1ss1mo:
> Hub greaves are already compatible with tails, since #61521 only items that cover hips conflict with tails and `ALLOWS_TAIL` was removed from greaves in that PR as redundant. image

oh, my bad, seems like it only the case for arcana dragonblood tail, guess I need to change it instead

PR #63449, opened 2023-02-06
Cqb fix

Dan1ss1mo:
#### Summary
Bugfixes "Adds missing MA's to cqb"

#### Purpose of change
Consistency

#### Testing
[ initiate testing sequence... ]
[ boot up errorse: none ]
[ testing implant functionality: within expected parameters ]
[ testing sequence completed successfully ]


Termineitor244 (changes requested):
I don't think the Five Deadly Venoms should be learnable in an unmodded game, these martial arts are in the base game for compatibility sake with previous versions of the game, but they belong to the Mythical Martial Arts mods now, and can't be learned unless one uses the debug menu since the martial arts books of these styles (and the trait), are now only in the mod, if you want to include them with the bionic, you should probably rewrite the bionic if the mod is enabled.

Dan1ss1mo:
> I don't think the Five Deadly Venoms should be learnable in an unmodded game, these martial arts are in the base game for compatibility sake with previous versions of the game, but they belong to the Mythical Martial Arts mods now, and can't be learned unless one uses the debug menu since the martial arts books of these styles (and the trait), are now only in the mod, if you want to include them with the bionic, you should probably rewrite the bionic if the mod is enabled.

Yeah, seem like you can't get any of this books in vanilla, makes sense

PR #63450, opened 2023-02-06
Make the broker to accept dehydrated meat

AnotherSeawhite:
#### Summary
None

#### Purpose of change
It's very weird that the refugee center doesn't accept dehydrated meat. This PR fixes it

#### Describe the solution
Add `dry_meat` to the broker's list

Making a reason why they can't accept dehydrated meat, but I don't see why they can't

#### Testing
Created a new world, spawned into a refugee center, debug spawned 10 units of dehydrated meats, delivered the meat and got 12 merch in return.


Profit0004:
Been a minute since I have traded with them but that seems like a large amount of merch.

MNG-cataclysm (approved):

Issue #63452, opened 2023-02-06
Sorting zones freeze


Raikiri:
### Describe the bug

Sometimes my sorting zones enter a state where the game freezes permanently every time I try to auto sort them. It only happens when I accumulate enough crap, so it's very hard to track down exactly what it's related to. However, thanks to the recently added feature, I now can upload a trimmed down version of the saved game where the issue is still reproducible.

### Attach save file

[Columbia Heights-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10664147/Columbia.Heights-trimmed.tar.gz)

### Steps to reproduce

1. Shift-o
2. Select all zones
3. Witness a freeze

### Expected behavior

I expect the game to not freeze when I use zones

### Versions and configuration

Current experimental.

### Additional context

The bug seems to be related to the current location of my vehicle. If I move it to a different place, it no longer freezes.

andrei8l:
/confirmed It gets stuck on `MAG_DESTROY` items in the unload zone. Items are sourced from vehicle and map, but `MAG_DESTROY` items are only removed from the map.

<details>
<summary>Potential fix</summary>

```diff
diff --git a/src/activity_item_handling.cpp b/src/activity_item_handling.cpp
index 2957d47578..6fe67a3df1 100644
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -2323,7 +2323,11 @@ void activity_on_turn_move_loot( player_activity &act, Character &you )
                         }
                     }
                     if( it->first->has_flag( flag_MAG_DESTROY ) && it->first->ammo_remaining() == 0 ) {
-                        here.i_rem( src_loc, it->first );
+                        if( this_veh != nullptr ) {
+                            src_veh->remove_item( src_part, it->first );
+                        } else {
+                            here.i_rem( src_loc, it->first );
+                        }
                         num_processed = std::max( num_processed - 1, 0 );
                         return;
                     }
```

</details>

Raikiri:
Do I understand correctly that a workaround for this issue is to disable my unloading zone?

andrei8l:
Yes. Let me know if that's not sufficient.

Raikiri:
That character had an accident with a really unfortunate grenade throw tonight, so he's not too concerned with zone problems anymore.

[...]

PR #63456, opened 2023-02-06
activity: respect source vehicle when unloading

andrei8l:
#### Summary
None

#### Purpose of change
Move loot and unload activities don't use the source vehicle when unloading ammo belts

Fixes: #63452

#### Describe the solution
Use the vehicle if target items were sourced from it

N/A

#### Testing
Updated test unit

Game doesn't hang in the save file from #63452 after starting the sort all activity

#### Additional context
There's a lot of duplicated code between the move loot and unload activities


mqrause:
> There's a lot of duplicated code between the move loot and unload activities

#63285 addresses a lot of that.

PR #63460, opened 2023-02-06
Update landscaping.json -- Bill flags adjustment

HotCupOfJava:
#### Summary
Content "The bill weapon was historically used as farming equipment similarly to the machete and should be considered BLADED_FARMING."

#### Purpose of change
With both the war scythe, a modified farming tool for combat, and the machete, a bladed chopping weapon, being eligible for use under Eskrima with the BLADED_FARMING tag, it is only fitting that the Bill, a farming tool modified into a polearm be treated similarly.

#### Describe the solution

Added BLADED_FARMING to both the bill and the bronze bill

Add a non-polearm variant of the Bill that could substitute.

#### Testing

Equipped a Bill as a character with Melee 10 and the Eskrima martial art, the Bill functioned as expected


Maleclypse (commented):
I’m not sure that a polearm is equivalent to the other items on the current list and I’m not seeing any real equivalent weapon to the bill on this list. https://en.m.wikipedia.org/wiki/Arnis

kevingranade:
You can't just take the category name at face value,  you need to look at usage in martial arts,  where it's very clear that this group is light one handed bladed implements.

PR #12992, opened 2015-07-19
Medieval churl JSON

forkingpath:
A friend (who provided the lovely Middle English descriptions) was playing a conduct game as a medieval transplant, so I thought it might be a fun addition to the game proper.

Future work: make chat with NPCs unintelligible.

PR #63465, opened 2023-02-07
Make Churls totally illiterate and socially inept

onura46:
#### Summary
Features "Make Churls totally illiterate and socially inept"

#### Purpose of change
Make Churls unable to talk to NPCs except for some trading, and make Churl scenario-only

#### Describe the solution
Churls are _literally_ ripped from another dimension, and would realistically lack even basic literacy skills because of their intended time period/profession. Right now, "Churl" exists as a normal profession but also through two challenge scenarios: "Portal Dependent" and "Medieval Peasant". This PR removes the normal profession due to the extradimensional nature of the churl and the new, somewhat hardcore additions that follow...

This PR also makes Churls (mostly) unable to speak to NPCs. The original author of the churl PR indicated their intent to try something like this. **You are still able to trade with friendly NPCs if they have 9 or more intelligence** who can understand your archaic drawl. Yep, it's a huge debuff, but I would argue Churls are a roleplay-heavy outlier anyway, so it's worth embracing! 👨‍🌾

Capture2

The mere existence of churls is a little anatopistic, so this PR may literalize roleplay elements that don't really need it. Put another way, does it ruin immersion more to be a medieval peasant speaking in modern vernacular to all NPCs or to be prevented from talking to anyone at all? Perhaps churl roleplayers prefer to make these decisions on a case-by-case basis (whilst maintaining a thick, extradimensional Welsh accent at all times, surely).

#### Additional context
Aerin: `personally if i could be assed to learn a minimal amount of c++ I'd give the churl a unique trait that prevents them from talking to NPCs as well cuz you don't speak this universe's language. alas, i can not be assed.`

#12992 : `Future work: make chat with NPCs unintelligible.`

I have had suggestions to allow requests for NPCs to follow as well as replace more dialog with archaic nonsense (mixing up strings in dialog is, I think, close to what the original author was talking about). This is outside the scope of this PR, but interesting nonetheless!

### DIFF
--- a/data/json/npcs/common_chat/TALK_COMMON_OTHER.json
+++ b/data/json/npcs/common_chat/TALK_COMMON_OTHER.json
@@ -532,5 +532,11 @@
     "type": "talk_topic",
     "dynamic_line": "Pleasure doing business!",
     "responses": [ { "text": "You might be seeing more of me…", "topic": "TALK_DONE", "effect": "u_bulk_trade_accept" } ]
+  },
+  {
+    "id": "TALK_CHURL_FRIENDLY",
+    "type": "talk_topic",
+    "dynamic_line": "Excuse me?  Um… you mean you want to trade?",
+    "responses": [ { "text": "Ye!", "topic": "TALK_DONE", "effect": "start_trade" } ]
   }
 ]
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -706,7 +706,8 @@
       { "level": 1, "name": "fabrication" },
       { "level": 1, "name": "tailor" }
     ],
-    "traits": [ "PROF_CHURL", "ILLITERATE" ]
+    "traits": [ "PROF_CHURL", "ILLITERATE" ],
+    "flags": [ "SCEN_ONLY" ]
   },
   {
     "type": "profession",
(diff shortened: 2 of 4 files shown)


Qrox:
> Functionally, this works exactly like Mute and Deaf right now

There could be some special dialog to let the churl convey simple meaning by pointing at things and doing imitation. but that's probably out of the scope of this PR.

onura46:
> There could be some special dialog to let the churl convey simple meaning by pointing at things and doing imitation. but that's probably out of the scope of this PR.

I'm still trying to think of a way to possibly allow trading with friendly NPCs, maybe as a check against the player's intelligence and the NPC's trust? Maybe instead of trading just make them follow? Like, you can't give them commands or talk to them but if you got a freakishly good roll versus their trust level, then whatever, you've got a buddy! I don't really know how to do that without shortcutting other blocks in dialog, though. And I imagine the scope will creep, yeah...

I left this as a draft for now since I still need to test and refine it. I'm open to more ideas.

Nebnis:
> > There could be some special dialog to let the churl convey simple meaning by pointing at things and doing imitation. but that's probably out of the scope of this PR.
> 
> I'm still trying to think of a way to possibly allow trading with friendly NPCs, maybe as a check against the player's intelligence and the NPC's trust? Maybe instead of trading just make them follow? Like, you can't give them commands or talk to them but if you got a freakishly good roll versus their trust level, then whatever, you've got a buddy! I don't really know how to do that without shortcutting other blocks in dialog, though. And I imagine the scope will creep, yeah...
> 
> I left this as a draft for now since I still need to test and refine it. I'm open to more ideas.

something like trading should be very easy to someone convey even if they were deaf and mute. It should be okay to allow that and disable practically anything else like quests and other high level cognitive conversations.

PR #63472, opened 2023-02-07
Heat pack fixes

### LINKED ISSUE #63454: Cutting up heat packs yields cotton patches
### Describe the bug

As the title suggests, cutting up heat packs yields cotton patches somehow?

### Attach save file

Shouldn't be necessary. Not that I have any heat packs on my savefile left anyway.

### Steps to reproduce

1. Find a heat pack
2. Cut up the heat pack
3. Behold the glorious material transmutation

### Expected behavior

It is made of plastic. It has no connections to cotton. It should either not be cuttable (depending on the irl size, not sure here) or give plastic chunks when cut.

### Screenshots

obraz

- OS: Windows
- Game Version: 908472c [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

How, even? I've been trying to find ways to connect json to the cutting code and kept failing, how in tarnation do we have a bug that overcomes that?

Karol1223:
#### Summary
None

#### Purpose of change
Fixes #63454. Also fixes up the volume and weight for heatpacks

#### Describe the solution
Removes the cotton patch from heatpack crafting recipe. Also changes the volume and weight to be more inline with the real world counterpart.

#### Additional context
The weight comes from https://apps.dtic.mil/sti/pdfs/ADA382575.pdf
The volume, however, is calculated proportionally. The above file only gives the dimensions of the chemical heater pad inside the heat pack. I took the dimensions of the entire heat pack from https://www.mreinfo.com/mres/flameless-ration-heater/ and then calculated them proportionally to the weight given from the pdf file

### DIFF
--- a/data/json/items/containers.json
+++ b/data/json/items/containers.json
@@ -2817,7 +2817,7 @@
     "pocket_data": [
       {
         "pocket_type": "CONTAINER",
-        "max_contains_volume": "3029 ml",
+        "max_contains_volume": "2777 ml",
         "max_contains_weight": "4 kg",
         "airtight": true,
         "moves": 50,


data/json/items/containers.json
```diff
@@ -2817,7 +2817,7 @@
     "pocket_data": [
       {
         "pocket_type": "CONTAINER",
-        "max_contains_volume": "2749 ml",
+        "max_contains_volume": "3029 ml",
```
andrei8l, line 2820:
```suggestion
        "max_contains_volume": "2777 ml",
```
So they can spawn sealed, as verified with Debug Menu -> Info -> Test Item Group -> MRE

Karol1223: You are absolutely right. For some reason I decided to increase the container volume by 280ml instead of 28ml. I'll fix that right away.

Panicode:
why did you do this, why did you dig through 23 year old stuff just for a heat pack weight and volume
I admire it heavily but wtf

PR #63476, opened 2023-02-07
Fix: Prevent infinite loop on unload ammobelt during sort from shopcart

inogenous:
#### Summary
Bugfixes "Prevent infinite loop on unload ammobelt during sort from shopcart"

#### Purpose of change

* Prevents infinite loop when sorting items using mag-unload on ammo belt in shopping cart
* Previously reproduced as follows:

1. Have "Loot: Unsorted"-zone on shopping cart
2. Have "Unload: Always unload"-zone on shopping cart
3. Have `belt223` ("5.56x45mm ammo belt") in shopping cart
4. Press key for "Sort out my loot"
5. Game enters infinite loop when unloading ammo belt from shopping cart

#### Describe the solution

Method `activity_on_turn_move_loop` is supposed to destroy the `belt223`-item once the last ammo is unloaded from it, but previously failed to do so because the item is not present in `here` - it is instead present in the vehicle referenced by `this_veh`. Previously, this made the ammo belt item not get removed from the autosort loop, thus being evaluated again in the next evaluation of autosort, where the same thing happened yet again and thereby causing an infinite loop.

With this commit, the mag (ammo belt) is instead removed from its vehicle when empty if it originated from one.

**Alt1:** We could also extract these lines with ("if its a vehicle, remove from vehicle, otherwise remove from map") into a separate function that can also be used from inside `move_item` where the same check also happens. Something like:
```
void remove_item(item &it, const tripoint_bub_ms &src, vehicle *src_veh, int src_part) 
{
    // Remove from map or vehicle.
    if( src_veh ) {
        src_veh->remove_item( src_part, &it );
    } else {
        here.i_rem( src, &it );
    }
}
```

**Alt2:** We could also add an error msg in `map.cpp` in `map::i_rem( const tripoint &p, item *it )` if the item could not be found, alerting the player about the problem. This would however not resolve the infinite loop.

#### Testing

Have tested on a save where this happened before. With this commit, the game does not enter an infinite loop when loading that save. Instead, the ammo belt is sorted (and destroyed into its parts) correctly.

### DIFF
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -2323,7 +2323,11 @@ void activity_on_turn_move_loot( player_activity &act, Character &you )
                         }
                     }
                     if( it->first->has_flag( flag_MAG_DESTROY ) && it->first->ammo_remaining() == 0 ) {
-                        here.i_rem( src_loc, it->first );
+                        if( this_veh ) {
+                            this_veh->remove_item( this_part, it->first );
+                        } else {
+                            here.i_rem( src_loc, it->first );
+                        }
                         num_processed = std::max( num_processed - 1, 0 );
                         return;
                     }


andrei8l:
#63456

inogenous:
Oops, duplicate of #63456 . Closing.

PR #63479, opened 2023-02-08
Windshields aren't just glass

Golfavel:
#### Summary
Balance "Common windshields are made of tempered glass"

#### Purpose of change
Fixes #62998 

#### Describe the solution
Changed standard windshields construction to use tempered glass sheets instead of standard glass sheets
Increased standard and reinforced windshields by 5x based on various sources
([1](https://www.fgdglass.com/blog/what-is-tempered-glass/), [2](https://www.dillmeierglass.com/news/tempered-vs-annealed-glass-a-comparison), [3](https://www.ny-engineers.com/blog/tempered-glass-properties-and-applications))
Introduced makeshift windshield vehicle part that acts just like windshields before the changes

Only changing the construction cost and increasing the durability to not hit string freeze.
Only changing the construction cost.
Dedicating more time to investigate bulletproof glass to more accurately adjust the reinforced windshields

#### Testing
Vehicles spawn with normal, tempered glass windshields, installing said windshields requires tempered glass sheets, and standard glass sheets can be used instead for makeshift windshields. New windshields require significantly more damage 

#### Additional context
Needs more discussion with regards to durability changes.
More research shows that they aren't always tempered glass, and that instead they have an internal layer made of some form of plastic that actually prevents it from shattering.


Boonlike:
If you want to get really technical, all car windows except the windshield are tempered glass. Tempering makes the glass brittle, that's why they break into little tiny bits instead of shards. Windshields are made of laminated glass, two layers pressed together on a layer of plastic. This is why the windshield doesn't collapse when a big rock hits it and maintains its shape even when it has a hole.

In my opinion your solution has the same in game effect of making the windshield more complex to build and more durable. If you wanted to go full realism, there already is laminated glass in the game and could be used instead.

Golfavel:
Yeah, as I researched more I realized there was more to this than I had anticipated. I'll try looking at laminated glass when I get back to this.

PR #63484, opened 2023-02-08
Fix holes in libraries

### LINKED ISSUE #54237: Libraries have holes in the floor

ghost:
#### Summary
Bugfixes "Fix libraries spawning with large holes"

#### Purpose of change
Fix #54237

#### Describe the solution
Replace "set": "point": "bash" with "place_rubble"

#### Testing
visit some vandalized libraries and notice that their isn't giant holes in the ground

#### Additional context
after fix
Screenshot (16)
Screenshot (19)
Screenshot (20)

### DIFF
--- a/data/json/mapgen/library.json
+++ b/data/json/mapgen/library.json
@@ -71,7 +71,7 @@
         "........................"
       ],
       "palettes": [ "standard_building_palette_vandalized", "vandalized_library_palette" ],
-	  "place_rubble": [ { "x": [ 4, 21 ], "y": [ 5, 21 ], "repeat": [ 30, 100 ] } ],
+      "place_rubble": [ { "x": [ 4, 21 ], "y": [ 5, 21 ], "repeat": [ 30, 100 ] } ],
       "nested": { "|": { "chunks": [ [ "bile_field", 10 ], [ "shelter_graffiti", 5 ], [ "general_graffiti", 20 ], [ "null", 75 ] ] } },
       "place_nested": [ { "chunks": [ [ "null", 80 ], [ "fire_field", 20 ] ], "x": [ 10, 20 ], "y": [ 9, 19 ] } ]
     }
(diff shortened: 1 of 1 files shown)


Maleclypse (changes requested):
Can you show a picture of a few vandalized libraries in this PR after your change?

Paleiste:
> Can you show a picture of a few vandalized libraries in this PR after your change?

Done

Maleclypse:
Ok yeah that's what I worried it would look like.  It looks like what you need to do is replace the set point bash with the following. So you can basically run the below across the same places as bash does.
```
### Place rubble and smash existing terrain with "rubble"

Creates rubble and bashes existing terrain (this step is applied last, after other things like furniture/terrain have
been set). Creating rubble invokes the bashing function that can destroy terrain and cause structures to collapse.

| Field       | Description
| ---         | ---
| rubble_type | (optional, furniture id, default: `f_rubble`) the type of the created rubble.
| items       | (optional, bool, default: false) place items that result from bashing the structure.
| floor_type  | (optional, terrain id, default: `t_dirt`) only used if there is a non-bashable wall at the location or with overwrite = true.
| overwrite   | (optional, bool, default: false) if true it just writes on top of what currently exists.

To use this type with explicit coordinates use the name "place_rubble" (no plural) like this:

```json
"place_rubble": [
    { "x": 10, "y": 1 }
```

Paleiste:
> Ok yeah that's what I worried it would look like. It looks like what you need to do is replace the set point bash with the following. So you can basically run the below across the same places as bash does.

I added your suggestion and updated my PR with the new changes.

Maleclypse:
That looks good! I'll adjust my review after tests complete

Maleclypse (approved):
Thank you for doing this!

PR #63485, opened 2023-02-08
Baby Wildlife Spawns

Termineitor244:
#### Summary
Content "Adds new baby wildlife spawns around the world"

#### Purpose of change
The Tamable Wildlife mod allows for taming of most baby wildlife, you can even tame a bear if you catch it young enough! The difficult thing is finding a bear cub to tame... And if that wasn't difficult enough in real life, in the game they don't even spawn naturally! They need to be birthed by existing (and living!) bears!

That of course is in addition to the obvious fact that baby wildlife should exist at day one of the game, nature did not stop its natural course for several months before the cataclysm just because.

#### Describe the solution
Adds many new spawns for mammal baby wildlife in company of its parents, or in rare occasions even lost and alone! Perfect for brainwashing from a dedicated player that wants to create an animal army.

The spawn of babies decrease as time passes, in different time spawns for different species, a pig will find very difficult to maintain alive its offspring after a week of zombie persecution, but a bear will have an easier time until the sheer number (and evolution) of zombies ends up culling its numbers.

#### Testing
Started the game with the changes in them, searched around for natural wildlife and special locations changed, they correctly spawn baby AND adult wildlife, often both.

#### Additional context
imagen
imagen
imagen
imagen
imagen


LyleSY:
I tried to do something similar to this a while ago for DinoMod. The issue I ran into was that I couldn't spawn groups of different animals together, so in the wilderness list you would just see a random baby off by itself. The only workaround I could figure out was to do it in mapgen with like a map special or extra, which is why DinoMod has some many nest specials. That said, the farm and zoo spawns make sense and look good. I'd recommend holding off on the wilderness stuff unless you intend random babies by themselves. Maybe consider a map special or extra to handle that? I agree that no babies existing in the wild in the Spring doesn't make a ton of sense.

Termineitor244:
Well, I frankly do not see the problem with a few baby animals alone in the apocalypse, the chance is small enough that you will have to search a lot to even *see* them, I justify it in my head as offspring that was left alone after their parents died by zombies/mutant animals, but they could run away or they were in another place when the death happened.

If this is really a concern and we do not want to have baby wildlife spawning alone in the wild, I could follow your footsteps and add some map extras, but I don't really think this is much of a problem, at least for natural wildlife.

JonathanLochridge:
> Well, I frankly do not see the problem with a few baby animals alone in the apocalypse, the chance is small enough that you will have to search a lot to even _see_ them, I justify it in my head as offspring that was left alone after their parents died by zombies/mutant animals, but they could run away or they were in another place when the death happened.
> 
> If this is really a concern and we do not want to have baby wildlife spawning alone in the wild, I could follow your footsteps and add some map extras, but I don't really think this is much of a problem, at least for natural wildlife.

In my view, it would only really be a problem for animal types where it wouldn't make much sense for the baby to be completely alone. Like herd animals.
The chances of a mother animal dying and leaving a baby alone is pretty likely. But, a baby being alive when it's whole pack dies or leaves seems less likely. Although, it could still potentially happen.

I am also not sure what the likelihood of that happening actually is in play.

So, kid/baby bears, birds, goats, mountain lions, moose,etc. don't seem to be a problem for me.
But, say deer, wolves and such not as much.

bombasticSlacks:
Conflicts on this now, fix when you can 😄

Termineitor244:
> Conflicts on this now, fix when you can 😄 

I thought so after seeing the PR by LyleSY be merged, haha. I will fix the conflicts and make the proposed changes to the babies' spawn of pack animals as soon as I find the time.

kevingranade:
Baby animals are highly seasonal,  they need to only appear in at least approximately the right time of year.

Termineitor244:
> Baby animals are highly seasonal, they need to only appear in at least approximately the right time of year.

Ok, will do then, I thought to leave it non-seasonal since I considered baby animals as "A year or less old" (The bear and deer babies take approximately that long into reaching adulthood in the game for example), but well, I will make the requested changes so they have a stronger spawn chance depending of their mating season and less so in the next season, removing the chance to spawn in the other 2 seasons.

Putting this in a draft until I finish updating the PR.

Zireael07:
"A year or less" old are more accurately described as "young" animals ("baby" animals  made me think of Spring, too)

Termineitor244:
> "A year or less" old are more accurately described as "young" animals ("baby" animals made me think of Spring, too)

Uh, oh, wrong choice of words then, sorry.

Termineitor244:
Rebased the branch to the current master, I re-added a pair of mutant mammals as possible spawns since they seem to have been forgotten when #63513 was made (And I noticed while updating my PR), I will also add again the Cerbearus as a possible spawn (even though it cannot have spawn) since I'm already here and it cannot spawn anywhere currently, after which I will implement the suggested changes for seasonal spawns.

PR #53606, opened 2021-12-19
Allow character swap on death.

Ramza13:
#### Summary
Features "Allow character swap on death."

#### Purpose of change
First part of #53523
When you die, if you have followers you can become one.

PR #51450, opened 2021-09-07
Take control of and play as NPC followers

eltank:
#### Summary
Features "Take control of and play as NPC followers"

#### Purpose of change

Ever wanted to take control of a follower for a while in order to make use of their abilities? Or just continue playing as that NPC? Well now you can.
This PR adds infrastructure allowing the player to relinquish control of their current avatar character and permanently take control of a follower NPC.
This functionality could be used to give a new option on death (continue playing the character of a follower), or simply to change the way you play the game (switch to controlling a follower at any time no matter where they are).
But for now it's limited to the debug menu as a proof of concept.

#### Describe the solution

The key idea that enables all of this is the transfer of `Character` data between `avatar` and `npc` objects (both subclasses of `Character`) using the `Character` move-assignment operator.
Add a protected `Character::swap_character(Character&, Character&)` and a public `avatar::control_npc(npc&)` that uses it.
Add a `shadow_npc` field to `avatar` which stores the suppressed personality of the character being controlled by the avatar. A shadow npc is generated automatically for the "original character" the first time this feature is used. The default shadow NPC has a neutral personality (0 in every personality trait) and a high `op_of_u` (10 trust and 10 value).
Move the `followers` field from `Character` to `avatar`.
Add a debug menu entry to activate the feature (target any follower NPC including those outside the reality bubble).

#### Describe alternatives you've considered

Doing more with it, as described above. But I want to collect some feedback first.

Implementing a "possession" spell in Magiclysm. I still want to do that in another PR, but the way that will work will be fundamentally different. The avatar won't change, instead they'll be rooted in place while they "warg" a monster, similar to remote-controlling an RC car or vehicle but seeing through the target's eyes. The range will be limited (less than reality bubble radius). The player should be able to invoke the controlled monster's special attacks and cancel at any time, possibly with negative after-effects, especially if the target took damage. Of course, maintaining possession will also cost mana each turn. Possessing one's own familiar may be cheaper/free.

#### Testing

Start a new game as Debug. Debug all skills to 5. Using debug Mind Control and NPC spawning acquire 2 followers. Rename them A and B. Tell them to guard. Make note of the 2 NPCs personalities to make sure they remain unchanged.

image

image

Take over npc A. Save and quit. In the main menu see only 1 save for the test world, named "Debug" (the save name is unaffected by character change). Load.
Take over npc B, repeat above.
Inspect Debug's personality, check default shadow NPC personality.

image

[...]

PR #63488, opened 2023-02-08
Update events to account for ability to swap avatars (debug or upon death)

### LINKED ISSUE #63455: Some scores/achievements broken due to take over follower NPC upon death of starting character
### Describe the bug

Some scores and achievements rely on event statistics that perform a comparison against `avatar_id`, as a check to determine whether it was the player character (as opposed to some other NPC) who did/attained something. This `avatar_id` is recorded from the starting character upon `game_start` event.

If your starting character dies and you take control of a follower NPC, the stats break because the player character now has a different character ID.

### Attach save file

N/A -- bug identified by code inspection

### Steps to reproduce

1. Start a new game as a survivor in evac shelter
2. Make the shelter NPC into a follower
3. Check score screen for "0 monsters killed"
4. Debug spawn a suitable weapon
5. Debug spawn a chicken
6. Kill the chicken, and check score increased to 1 kill
7. Debug to cause yourself a bunch of damage, then spawn a ladder, climb on the roof and walk off it
8. Die from fall, take over the NPC
9. Grab the weapon, spawn a chicken, kill it
10. Kill count on score tab has stopped incrementing. Kill count on kills tab still works, though

### Expected behavior

Either:

All scores, achievements and conducts accrued so far continue to count and continue to be updated with respect to the new player character.

or

All scores, achievements and conducts are reset upon death and take over of a new character.

### Screenshots

_No response_

Bug reproduced in 4b6270c

### Additional context

_No response_

cake-pie:
#### Summary
Bugfixes "Update events to account for ability to swap avatars (debug or upon death)"

#### Purpose of change

Resolves #63455 

#51450 and #53606 gave the ability to switch to another character.
Preexisting game events (that feed into memorial, stats, achievements, conducts, scores) are not equipped to handle this, as they only recorded the starting character in `event_type::game_start` and then death and last words in `event_type::game_over`. This oversight causes some things to break.

#### Describe the solution
- [x] split `game_avatar_new` off from `game_start` and `game_avatar_death` from `game_over`
- [x] add migration for existing saves that retroactively populates `game_avatar_new` from available information
  - where `game_start` is available, get starting character information from it
    - if current character is different from starting character, add it as well, treating current turn as the time they took over
  - if `game_start` is also not available (pretty old save!) treat current character as starting character.
- [x] adjust existing event emissions (split into two events accordingly)
- [x] emit `game_avatar_new` at the end of `avatar::control_npc()` when new avatar takes over
- [x] show the "tombstone" screen and prompt for last words on death of all player characters (not just the final one when you run out of "continues" or choose not to do so)
- [x] emit `game_avatar_death` to record last words of all dead PCs
- [x] change `avatar_id` event statistic to pull its value from the most recent `game_avatar_new` event
- [x] memorial logger logs `game_avatar_death` instead of `game_over`
- [x] memorial logger logs `game_avatar_new` instead of `game_start`, and will have different text for a character swap rather than a brand new game
- [x] past games info will try to account for multiple characters over the course of a save, insofar as the information is available. It will pull the starting character name from the first `game_avatar_new` entry. Depending on how many `game_avatar_new` entries there are, it will append "et. al." to indicate that there was/were subsequent character(s) that took over from the starting character due to death or debug. If there are no `game_avatar_new` entries, it will fall back on the legacy approach and pull the starting character name from `game_start`.

#### Behavior of stats after this PR
- any statistic that doesn't rely on comparison against the `avatar_id` statistic continue working as they did
- any statistic that depends on `avatar_id` will switch to the "lifetime" stats of the new avatar when switching to them
  - this includes things they did as an NPC, e.g. any kills, skill gain etc that was captured by the events system

This is imperfect, but now we are at least capturing an event whenever player control switches from one character to another. This ensures that save games will contain the information needed when a fuller solution is developed.

Subsequent work can look into a different way for event transformation to filter events (e.g. kills) to match them against the player character `avatar_id` ***at the time the event occurred*** rather than the current avatar, so that the affected scores, achievements and conducts are on the basis of *player action*.

#### Additional context
I can't do anything about information that was lost between #51450 / #53606 and now, because it wasn't able to be captured.

~This will probably break some achievements in existing saves, because they'd look for `avatar_id` but be unable to resolve a value for it when there are no `game_avatar_new` events yet. There's no way in the event statistics system to "union" two event streams (`game_start` with `game_avatar_new`) and implementing such a feature looks like it will get really complex, so I don't think it's worth it just for the sake of this one unusual case. Perhaps some other migration strategy could be devised? In any case, the affected achievements are already broken anyway ( #63455 ) so I'd be satisfied with fixing things so that they work correctly for new saves henceforth.~

Hijack the existing `game_start` and `game_over` so that they are also emitted whenever there is a character swap, but that seems weird esp. with respect to the game version and total play time fields.

#### Testing

Simple test case: [TestLand.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10696161/TestLand.zip)

1. Build this PR, load the save in and verify that the score screen shows 5 monsters killed as the first stat on top. Spawn a slow, harmless monster e.g. water strider and kill it; check that the count has increased to 6.

1. Using the debug menu, `p`, `x` to switch to the follower NPC. Check the score screen, which should now shows 0 kills (Antonia Boswell's lifetime stats). Spawn water strider; kill; check that the count has increased to 1.

1. Switch back to the original character. Check score screen, which should list 6 kills.

1. Fire up a recent experimental build and load up the original save file again. Verify that the score screen shows 5 monsters killed.

1. Switch to the follower NPC. Check that score screen; it shows 5 kills. Spawn water strider; kill; check the count, which did not increase.

1. Save the game and transfer it to the PR build. The score screen should reflect Antonia's 1 kill.

1. Spawn water strider; kill; check count increased to 2.

1. Switch back to the original character. The score screen should reflect their 5 kills.

### DIFF
--- a/data/json/statistics.json
+++ b/data/json/statistics.json
@@ -2,8 +2,8 @@
   {
     "id": "avatar_id",
     "type": "event_statistic",
-    "stat_type": "unique_value",
-    "event_type": "game_start",
+    "stat_type": "last_value",
+    "event_type": "game_avatar_new",
     "field": "avatar_id"
   },
   {
--- a/data/json/statistics.json
+++ b/data/json/statistics.json
@@ -7,6 +7,13 @@
     "field": "avatar_id"
   },
   {
+    "id": "last_words",
+    "type": "event_statistic",
+    "stat_type": "last_value",
+    "event_type": "game_avatar_death",
+    "field": "last_words"
+  },
+  {
     "id": "avatar_wakes_up",
     "type": "event_transformation",
     "event_type": "character_wakes_up",
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -166,7 +166,7 @@ static void swap_npc( npc &one, npc &two, npc &tmp )
     two = std::move( tmp );
 }
 
-void avatar::control_npc( npc &np )
+void avatar::control_npc( npc &np, const bool debug )
 {
     if( !np.is_player_ally() ) {
         debugmsg( "control_npc() called on non-allied npc %s", np.name );
@@ -204,11 +204,11 @@ void avatar::control_npc( npc &np )
     character_mood_face( true );
 
     profession_id prof_id = prof ? prof->ident() : profession::generic()->ident();
-    get_event_bus().send<event_type::game_avatar_new>( false, getID(), name, male,
-            prof_id, custom_profession );
+    get_event_bus().send<event_type::game_avatar_new>( /*is_new_game=*/false, debug,
+            getID(), name, male, prof_id, custom_profession );
 }
 
-void avatar::control_npc_menu()
+void avatar::control_npc_menu( const bool debug )
 {
     std::vector<shared_ptr_fast<npc>> followers;
     uilist charmenu;
@@ -229,7 +229,7 @@ void avatar::control_npc_menu()
     if( charmenu.ret < 0 || static_cast<size_t>( charmenu.ret ) >= followers.size() ) {
         return;
     }
-    get_avatar().control_npc( *followers[charmenu.ret] );
+    get_avatar().control_npc( *followers[charmenu.ret], debug );
 }
 
 void avatar::longpull( const std::string &name )
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -203,8 +203,9 @@ void avatar::control_npc( npc &np )
     g->update_map( *this, z_level_changed );
     character_mood_face( true );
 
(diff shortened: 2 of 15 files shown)


cake-pie:
Added migration from existing saves that makes best-effort attempt to populate `game_avatar_new` retroactively with what information is available. Imperfect but better than nothing.

Ready for testing and review.

cake-pie:
Rebased to resolve conflicts.

PR #63492, opened 2023-02-08
Adds roads to roadless mansions

FarkasKBence:
#### Summary

None

#### Purpose of change

Wilderness mansions are weird: they're basically regular mansions but without the road connection.  They look out-of-place, given that they're disconnected from the rest of the world, only accessible by foot.

#### Describe the solution

Added a road connection to `Mansion_Wild` and `Mansion_WildAlt`. Renamed them to `Mansion_Road_1` and `Mansion_Road_Alt_1` to avoid confusion.

None.

#### Testing

Game loads without errors, debug-spawned the four variants.

#### Additional context

None.

### DIFF
--- a/data/json/starting_missions.json
+++ b/data/json/starting_missions.json
@@ -635,7 +635,7 @@
     "destination": "mansion_c3",
     "difficulty": 1,
     "value": 0,
-    "start": { "assign_mission_target": { "om_terrain": "mansion_c3", "om_special": "Mansion_Wild", "search_range": 200 } },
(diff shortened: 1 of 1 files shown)


LovamKicsiGazsii:
On further thought, giving wilderness mansions a road connection might be a better solution here. Just realized regular and wilderness mansions have some structural differences. My mistake.

Nebnis:
I assumed people got there by a helicopter

LovamKicsiGazsii:
> I assumed people got there by a helicopter

Even if there's enough room for a helicopter to land (which there isn't), I still think it'd make sense to add road connections here. You can't exactly haul barrels of wine, or large quantities of building material by helicopter, after all.

Maleclypse (approved):

PR #63500, opened 2023-02-09
inv_ui: show healing tools in consumption UIs

andrei8l:
#### Summary
None

#### Purpose of change
Follow up for #63434 to preserve the current UX: show all healing tools and medication in the consumption UIs, particularly in the dedicated `Consume Medication` UI

Depends on #63434 

#### Describe the solution
Add a bit of glue so that `heal` tools show up and can be used from the consumption UIs
Invoke heal method directly instead of passing through the consumption code path
Add more glue to ensure the `heal` method is invoked directly for tools with multiple uses (ex: rag)
Add `ALLOWS_REMOTE_USE` to all liquid healing tools to preserve old behaviour (and this can potentially be hardcoded)

Any other option breaks player habits and there's no good reason to do that here. *IF* the total UI revamp gains traction, we can consider discarding the bandage-is-comestible paradigm.

#### Testing
<details>
<summary>before #63434, for reference</summary>

Screenshot from 2023-02-09 08-05-39

</details>

<details>
<summary>with only #63434</summary>

where bandage

</details>

<details>
<summary>with both</summary>

Screenshot from 2023-02-09 10-21-16

Note that the consumption time now respects the definition in JSON.
Verify that you can actually use the bandages from this menu.

</details>

#### Additional context
N/A

### DIFF
--- a/data/json/items/comestibles/med.json
+++ b/data/json/items/comestibles/med.json
@@ -314,7 +314,7 @@
     "color": "white",
     "phase": "liquid",
     "container": "bottle_plastic_small",
-    "flags": [ "NO_INGEST", "WATER_DISSOLVE" ],
+    "flags": [ "NO_INGEST", "WATER_DISSOLVE", "ALLOWS_REMOTE_USE" ],
     "use_action": { "type": "heal", "disinfectant_power": 4, "bite": 0.95, "move_cost": 2000 }
   },
   {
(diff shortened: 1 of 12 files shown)


cake-pie:
Thanks for this, one less thing on my plate.

bombasticSlacks:
I was looking for a fix like this TY for your work as always.

PR #63513, opened 2023-02-10
wilderness spawn rework

LyleSY:
#### Summary
Balance "wilderness spawn rework"

#### Purpose of change

Make it simpler to balance and manage wilderness spawns

#### Describe the solution

Creates new spawn groups for the enormous FOREST monster spawn group and manages them so that insect and zombie spawns increase over time. 
Same for smaller SWAMP group
Removes a bunch of comment lines I added to document the old confusing spawn system that doesn't exist anymore
Removes "mon_null" spawns
Changes all "freq" to "weight"
Manages insect and zombie evolution and insect seasons and mammal decline with the groups
Removes lines dedicated to micromanaging spawns based on time of day, which is either entirely or mostly broken.
Uses existing cat and dog groups to manage cat and dog spawns instead of duplicating them here. 
Standardizes times to 3 days, 7 days, 28 days, and 90 days to make management possible.
Standardizes insect spawns to be consistent.
Assigns insects to either be on forest or swamp, not both
Real world colony insects like wasps, bees, and ants don't spawn randomly alone
Removes duplicate lines of monsters at different stages of mutation since that is now duplicative with the newish mutation system.
Day 1 forest spawns are now only living unmutated animals and harmless mutant insects. That becomes much less true over time.

Doing less

#### Testing

Game loads no errors, monsters spawn as intended

#### Additional context

Overdue review after the spawn group rework

### DIFF
--- a/data/json/monstergroups/wilderness.json
+++ b/data/json/monstergroups/wilderness.json
@@ -4,7 +4,6 @@
     "name": "GROUP_CAVE",
     "is_animal": true,
     "monsters": [
-      { "monster": "mon_null", "weight": 15 },
       { "monster": "mon_bat", "weight": 600, "cost_multiplier": 5, "pack_size": [ 6, 32 ] },
       { "monster": "mon_bear", "weight": 100, "cost_multiplier": 10, "pack_size": [ 1, 3 ] },
       { "monster": "mon_cougar", "weight": 100, "cost_multiplier": 20, "pack_size": [ 1, 2 ] },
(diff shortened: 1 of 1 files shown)


Karol1223:
I do think that adult antlions and stag beetles should still be able to spawn, which I do believe you removed in this PR. Makes no sense for only larval forms of them to exist in the world. Specifically with the antlions, where adults are less threatening than the larvae.

LyleSY:
Apologies, I thought they were duplicative, I didn't realize you were using age_grow there. You're right, the forest spawn group will have to manage this manually to get your intended result.

Venera3:
"Reddit complains about X" is on its own no real reason to change anything. Not to mention that this actually buffs insect spawns (those it doesn't remove, at least).

LyleSY:
Agreed, Reddit reference removed. It was confusing and unnecessary

Issue #63087, opened 2023-01-12
Weak light sources shouldn't ruin stealth


SentientBowtie:
### Is your feature request related to a problem? Please describe.

Having an active light source on your person makes you visible to enemies as if your tile were brightly lit, i.e. by daylight, even during the night or in an enclosed space with no lighting. This is a reasonable mechanic on its own, but the threshold at which this effect occurs is incredibly low, and several small light sources trigger it even when that doesn't make a lot of sense: Lit cigarettes, cigars, and joints, and active lightstrips, all of which provide a mere `LIGHT_4`, will render you just as visible in the darkness as an active flashlight or mining helmet.

### Solution you would like.

**Increase the threshold at which the** `haslight` **effect is applied to the player** (in [src/lightmap.cpp](https://github.com/CleverRaven/Cataclysm-DDA/blob/66746d1e9c84d522739c3acdbf9332e740f7c5e7/src/lightmap.cpp)) **from 4 to another number.** 

The simplest way to solve the issue would be raising it to 6, as that neatly excludes every problem object without altering anything else.

In the interest of furthering realism, I would alternatively suggest **raising it to at least 11** in order to exclude everything of `LIGHT_10` or less, for the following reasons.

Lit Up could currently be rationalized as enemies noticing bespoke sources of light and being drawn to them, which then allows them to notice the player when they get close. However, this applies to any `Hostile!` creature that would be able to see the player in bright lighting, even when it doesn't make sense for them to be attracted to light simply because it's there. Notable examples include nocturnal predators such as wolves and tigers, as well as creatures which _die when exposed to sunlight_ like the dark wyrm. Additionally, enemies will ignore active light sources unless they are being carried by the player; turning on a cell phone's flashlight and dropping it in the grass only matters for single turn the player is holding it, and as soon as any nearby zombies forget the player exists, they won't bother to investigate the light.

I have instead come up with another justification for the effect that I feel is more easily applicable to how it works mechanically: being Lit Up makes the player more easily detectable because their light source is illuminating them in such a way that their silhouette is mostly visible and identifiable against the surrounding darkness. Zombies see something bipedal that isn't part of the blob, triffids see something not part of the grove, the great mantis identifies small prey, and so on.

[...]

PR #63516, opened 2023-02-10
No more cigarette flashlights

Karol1223:
#### Summary
Bugfixes "Lit cigarettes no longer function as light sources"

#### Purpose of change
It was brought up on discord how some people use cigarettes to dissect in the dark. Need I say more?

#### Describe the solution
Change ``LIGHT_4`` flag to ``LIGHT_2`` on cigars, cigarettes, and joints. This means cigarettes will, when dropped on the ground, illuminate that tile but not enough to allow for stuff such as dissection or performing most crafts. It appears, however, that when 'worn' they produce no light whatsoever.

### DIFF
--- a/data/json/items/generic.json
+++ b/data/json/items/generic.json
@@ -3048,7 +3048,7 @@
         }
       ]
     },
-    "flags": [ "LITCIG", "TOBACCO", "LIGHT_4", "TRADER_AVOID", "FIRESTARTER", "FRAGILE", "OVERSIZE" ]
+    "flags": [ "LITCIG", "TOBACCO", "TRADER_AVOID", "FIRESTARTER", "FRAGILE", "OVERSIZE" ]
   },
   {
     "type": "GENERIC",
@@ -3095,7 +3095,7 @@
         }
       ]
     },
-    "flags": [ "LITCIG", "TOBACCO", "LIGHT_4", "TRADER_AVOID", "FIRESTARTER", "FRAGILE", "OVERSIZE" ]
+    "flags": [ "LITCIG", "TOBACCO", "TRADER_AVOID", "FIRESTARTER", "FRAGILE", "OVERSIZE" ]
   },
   {
     "type": "GENERIC",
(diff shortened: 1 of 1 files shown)


GuardianDll:
How does such small values work in game? I though the idea was to make it not bright enough to do anything useful, but enought to disable stealth with cigarette in the mouth

Karol1223:
> How does such small values work in game? I though the idea was to make it not bright enough to do anything useful, but enought to disable stealth with cigarette in the mouth

I am not sure. At ``LIGHT_4`` it was enough to light up the tile of the person the same way a lighstrip did aka allow to do anything on that one tile. It is possible that lowering it further may give stealth penalties without allowing for light-requiring activities but I honestly am not sure a cigarette is enough to compromise stealth when the enemy is a zombie - they're not attracted to the light, they're attracted to you in the light, and a cigarette isn't enough for them to see you in said light.

Night-Pryanik:
This is intentional - see #63087.

Zireael07:
@Night-Pryanik That issue is about stealth while IIUC this issue is about being able to do stuff requiring light

GrandpuhTy:
Should cigarettes have an affect on night vision if they're extremely dim? I'm not really sure they're bright enough and in the right color spectrum to have a significant impact on your night vision.

Karol1223:
> This is intentional - see #63087.

Right - this does still go back to the drawing board though. A cigarette should not allow you to be able to perform actions as if you were illuminated. If I had to choose between being stealthier and being able to perform an autopsy under cigarette light I'd rather choose the former. Do you have a suggestion as to how fix both the issues together?

Night-Pryanik:
I'd add check for very low light conditions, such as lit cigarettes, and forbid conducting most activities requiring proper light.

Karol1223:
> I'd add check for very low light conditions, such as lit cigarettes, and forbid conducting most activities requiring proper light.

This does seem like a good solution. Would you say then that the ``LIGHT`` flag should be given back to them? If so, should it still be at 4? I don't feel like a lightstrip and a cigarette give off the same amount of light...

Night-Pryanik:
Light level produced by lit cigarette could be lowered to 2 or 3 then.

Karol1223:
> Light level produced by lit cigarette could be lowered to 2 or 3 then.

2 looks to be a good middle ground. When carried in the mouth it does basically nothing, when dropped on the ground it gives enough light to illuminate but not enough to dissect or craft (most crafts, anyway, it allows for some still, not sure why).

Issue #63515, opened 2023-02-10
Blackpowder Fouling Extremely Rapid


Grubleafeater:
### Describe the bug

If placed in inventory, weapons that have been fired with EVEN ONCE with blackpowder will rust away in a matter of HOURS.

This is easiest to see with flintlock pistols or pistols you shoot with manually loaded black powder munitions, but happens to seemingly all guns.

### Attach save file

N/A

### Steps to reproduce

Spawn in pocket universe, 2 each of a handful of guns, and black powder ammo. Fire each once, drop 1 of the pair and put the other in pocket universe. Rest.

### Expected behavior

Having the weapon in inventory should not impact rust rate; also, _weapons should not rust to XX within 2 hours of being used once._

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.2364 (21H2)
- Game Version: 2ee4928 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Xedra Evolved [xedra_evolved],
    Magiclysm [magiclysm],
    Blaze Industries [blazeindustries],
    SpeedyDex [speedydex],
    Stats Through Kills [stats_through_kills],
    Stats Through Skills [StatsThroughSkills],
    My Sweet Cataclysm [my_sweet_cataclysm]
]

andrei8l:
/confirmed

Top-level items never process blackpowder fouling because the flag `BLACKPOWDER_FOULING_DAMAGE` doesn't mark them as active.

https://github.com/CleverRaven/Cataclysm-DDA/blob/43d20409b45fbb6b08cd7aaa39191e7f9d23a9db/src/item.cpp#L12522

Items stored in containers with at least one active item ignore that check and roll 1:2000 chance every turn for damage.

https://github.com/CleverRaven/Cataclysm-DDA/blob/43d20409b45fbb6b08cd7aaa39191e7f9d23a9db/src/item.cpp#L13389

#63233 will fix this bypass, but not for items in inventory.

So basically blackpowder fouling hasn't worked correctly at least since the pocket system was introduced.

anoobindisguise:
[...]

PR #63540, opened 2023-02-11
Blackpowder fix

Grubleafeater:
#### Summary
Balance "Less Rapid Rusting"

#### Purpose of change
Make blackpowder fouling less blatantly ridiculous. Currently, on average, any amount of blackpowder fouling will damage a gun every 30 minutes if it's in a active container.

#### Describe the solution
As I lack the expertise to do anything better, I'm inflating the damage rate to be more like once a day on average rather than every 30 minutes. Far more realistic, and far less infuriating to boot.

Someone more skilled than me could revamp blackpowder fouling to actually consider how many times you've fired it without cleaning it.

#### Additional context
#63515


kevingranade:
Discussion on discord has brought to light that this has zero testing and not even an estimate of the end results.

PR #63551, opened 2023-02-12
player can now use clean water in tempered steel crafts

Daved27hundred:
#### Summary
Bugfixes "fixed tempered steel crafts only taking unclean water as a tool"

#### Purpose of change

tempered steel arms and armor use water along with a 60L tank as a tool for quenching. For an odd reason it will only accept non-purified water and not clean water. I imagine this being annoying for someone who purifies every drop of water they get.
#### Describe the solution

added some code to all recipes that use the tank and water as a tool, basically everything that needed the "quenching and tempering" proficiency.

keeping it as it is.

### DIFF
--- a/data/json/recipes/armor/suit.json
+++ b/data/json/recipes/armor/suit.json
@@ -3279,12 +3279,7 @@
       { "proficiency": "prof_articulation" },
       { "proficiency": "prof_quenching", "fail_multiplier": 2 }
     ],
-    "tools": [
-      [ [ "tongs", -1 ] ],
-      [ [ "hotcut", -1 ] ],
-      [ [ "metal_tank", -1 ] ],
-      [ [ "water", -240 ], [ "water_clean", -240 ] ]
-    ]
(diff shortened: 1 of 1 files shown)


data/json/recipes/armor/suit.json
```diff
@@ -3279,7 +3279,12 @@
       { "proficiency": "prof_articulation" },
       { "proficiency": "prof_quenching", "fail_multiplier": 2 }
     ],
-    "tools": [ [ [ "tongs", -1 ] ], [ [ "hotcut", -1 ] ], [ [ "metal_tank", -1 ] ], [ [ "water", -240 ] ] ]
+    "tools": [
+      [ [ "tongs", -1 ] ],
+      [ [ "hotcut", -1 ] ],
+      [ [ "metal_tank", -1 ] ],
+      [ [ "water", -240 ], [ "water_clean", -240 ], [ "water_clean", -240 ] ]
```
dseguin, line 3286: Was it intentional to add `water_clean` twice here?

Daved27hundred: oh whoops. I’ll fix this at home

PR #63553, opened 2023-02-12
Add contested roll function to EOCs

### LINKED ISSUE #60773: Dialogue condition for contested rolls

I'd like to see dialogue conditions expanded to allow contested rolls.  This would allow us to create some random chance in areas of skill and stats in a way that probably replicates hacking and lockpicking but makes them expandable in a much more jsonized fashion than currently.

### Solution you would like.

I'd like to be able to select a skill and a stat and then roll them against a difficulty.   I'm open to how the math actually works but one example I'm envisioning is if the chosen skill and stat were combined and that was rolled as a percentage against the difficulty.  I'm not sure if that provides enough range for an average success rate maybe it could be theoretical +practical + stat all times 100 against difficulty times 100.  

Ideally the player would be able to see their chance of success before deciding.

Any other math that makes sense would also work.  Could do skill/stat against difficulty level as an option too instead of it being skill+stat against difficulty.

### Additional context

_No response_

Ramza13:
#### Summary
None

#### Purpose of change
Fix #60773
More json power

#### Describe the solution
Add a new function to allow EOC to use contested rolls.

#### Testing
Used this json
```
  {
    "id": "EOC_TEST",
    "condition": { "roll_contested": { "u_val": "strength" }, "difficulty": 12 },
    "effect": [
      { "u_message": "True" }
    ],
    "false_effect": {
      "u_message": "Fail." }
  }
```

#### Additional context
@Maleclypse let me know if this isn't what you wanted.

### DIFF
--- a/doc/NPCs.md
+++ b/doc/NPCs.md
@@ -977,7 +977,7 @@ Condition | Type | Description
 `"u_rule" or "npc_rule"` | string or [variable object](#variable-object) | `true` if u or the NPC follower AI rule for that matches string is set.
 `"u_override" or "npc_override"` | string or [variable object](#variable-object)| `true` if u or the NPC has an override for the string.
 `"has_pickup_list" or "u_has_pickup_list" or "npc_has_pickup_list"` | simple string | `true` if u or the NPC has a pickup list.
-`"roll_contested""`, `difficulty`: int or [variable object](#variable-object), (*optional* `static_diff : `bool ) | [int expression](#compare-integers-and-arithmetics) | Compares a roll against a difficulty. Returns true if a random number between 0 and the integer expression is greater than a random number between 0 and the `difficulty`.  If `static_diff` (defaults false) is true instead a random number between 0 and the integer expression will be compared to the `difficulty` value.  For example { "u_roll_contested": { "u_val": "strength" }, "difficulty": 4 } will return whether a random number between 0 and strength is greater than a random number between 0 and 4.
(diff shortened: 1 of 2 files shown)


PatrikLundell:
When possible, I would avoid using the same number for two parameters in tests, in order to be able to detect cases where the parameters are swapped or the same parameter is accidentally used for both purposes (such as when using copy/paste during coding and forgetting to replace the parameter used).

doc/NPCs.md
```diff
@@ -977,6 +977,7 @@ Condition | Type | Description
 `"u_rule" or "npc_rule"` | string or [variable object](#variable-object) | `true` if u or the NPC follower AI rule for that matches string is set.
 `"u_override" or "npc_override"` | string or [variable object](#variable-object)| `true` if u or the NPC has an override for the string.
 `"has_pickup_list" or "u_has_pickup_list" or "npc_has_pickup_list"` | simple string | `true` if u or the NPC has a pickup list.
+`"roll_contested""`, `difficulty`: int or [variable object](#variable-object), (*optional* `static_diff : `bool ) | [int expression](#compare-integers-and-arithmetics) | Compares a roll against a difficulty. Returns true if a random number between 0 and the integer expression is greater than a random number between 0 and the `difficulty`.  If `static_diff` (defaults false) is true instead a random number between 0 and the integer expression will be compared to the `difficulty` value.  For example { "u_roll_contested": { "u_val": "strength" }, "difficulty": 4 } will return whether a random number between 0 and strength is greater than a random number between 0 and 4.
```
Maleclypse, line 980: Does it make more sense to be pulling a random number from zero to the stat/skill level rather than being like a d10 plus the stat/skill to beat the difficulty?  
That's how I was thinking it would be easiest to balance when designing content.  This way you have currently might make more sense but I'm not sure of the math on it.

Ramza13:
> When possible, I would avoid using the same number for two parameters in tests, in order to be able to detect cases where the parameters are swapped or the same parameter is accidentally used for both purposes (such as when using copy/paste during coding and forgetting to replace the parameter used).

Good point.   Updated test json and tried out some more cases.

> Does it make more sense to be pulling a random number from zero to the stat/skill level rather than being like a d10 plus the stat/skill to beat the difficulty? That's how I was thinking it would be easiest to balance when designing content. This way you have currently might make more sense but I'm not sure of the math on it.

I had no idea what formula to use so guessed, I've updated it to use this formula instead.

PR #54344, opened 2022-01-13
More mapgen sanity checking and control

jbytheway:
#### Summary
Infrastructure "Better checking of mapgen definitions"

#### Purpose of change
We have a lot of mapgen that runs on top of existing maps.  This can happen due to nested chunks, map extras, predecessor mapgen, update mapgen, or some hardcoded things.

#### Describe the solution
This PR adds a sanity check for whenever mapgen changes the terrain on a tile where something (furniture, trap, or item) already exists.  This is often a sign of a problem.  When it happens, it flags a warning message.

It also provides a variety of solutions to this problem that mappers can use:
- Two new mapgen flags `ALLOW_TERRAIN_UNDER_OTHER_DATA` and `ERASE_ALL_BEFORE_PLACING_TERRAIN` which specify the desired default behaviour if a particular piece of mapgen wishes to use the same behaviour whenever this happens.  The first flag suppresses the warning while the second causes everything on the tile to be removed before setting the terrain (which is normally what you want).
- A new `remove_all` mapgen command that erases everything on a tile, for when the mapper needs to be more targeted in removing only in some cases.

The recently-added exhaustive mapgen unit test allows these problems to be quickly discovered, so I ran that a lot and fixed all the issues it found (although, as usual, there's probably a long tail of low-probability issues I didn't fix).

I only ran the tests for Magiclysm, not other mods, so there are probably issues there too.

I still need to write documentation, so opening this as a draft.

#### Describe alternatives you've considered
I considered enhancing the `terrain` definition itself to allow specifying how to handle this situation.  That might be better than adding the `remove_all` command for fixing these issues, but I think `remove_all` is probably something we want to have regardless.

#### Testing
Running the unit tests a lot.

#### Additional context
I started this work back in October, but realised it was infeasible to fix all the issues without having a unit test to help, so I have been diverted by adding that unit test in the time since and only recently returned to this.

Issue #57631, opened 2022-05-11
Tacoma farm map error report (flags...)


PatrikLundell:
### Describe the bug

 DEBUG    : In nested mapgen tacoma_commune_junk_shop_13 in unknown object in add_mapgen_update_func on ranch_camp_49_north, setting terrain to t_dirtfloor (from t_dirt) at (2,12) when item feces_manure existed.  Resolve this either by removing the terrain from this mapgen, adding suitable removal commands to the mapgen, or by adding a suitable flag to the innermost mapgen: either ERASE_ALL_BEFORE_PLACING_TERRAIN if you wish terrain to replace everything previously on the tile or ALLOW_TERRAIN_UNDER_OTHER_DATA if you wish the other items to be preserved

 FUNCTION : apply
 FILE     : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\mapgen.cpp
 LINE     : 2612
 VERSION  : 30c8dc3

### Steps to reproduce

1. Hand in the radio quest at the Tacoma farm.
2. Get the message when the map is modified.

### Expected behavior

No error report.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19043.1645 (21H1)
- Game Version: 30c8dc3 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]
+ JSON addition of wind mill recipe to workshop 2 blacksmith
+ JSON "mod" to disable portal storms as the PR mod providing that was rejected.

### Additional context

I'd recommend the ALLOW_... flag, as the player may potentially (and stupidly) place stuff in that location.

haveric:
I'll probably add the allow to everything in the Tacoma updates. I was hoping to be able to move some of them to different locations, but I don't think that has been implemented yet.

SamGondra:
this is something that happens (still) in 0.H stable
```
[...]

PR #63559, opened 2023-02-12
Finer-grained mapgen flags for terrain wiping

### LINKED ISSUE #57418: nested mapgen error when NPC comes back from creating basic farm plot mission
### Describe the bug

When recalling the NPC from the create farm plot expansion, this error shows up:
image

### Steps to reproduce

- Load the attached save
- Interact with the billboard on the tile southeast of the player
- Recall the NPC from the farm basic plots mission (which I made one maptile to the north)
- Observe error

[Yocemento.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/8624904/Yocemento.zip)

### Expected behavior

Create the basic farm plot without throwing the error

### Screenshots

This is the mission the NPC is recalled from:
image

After ignoring the error, the farm plots seem to be created at least:
image

- OS: Windows
    - OS Version: MINGW/CYGWIN/MSYS2 on unknown Windows version
- Game Version: f97c81e [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm]
]

### Additional context

The error in text form:
```
DEBUG    : In nested mapgen bc_prim_farm_long_fields in update mapgen faction_base_field_camp_farm_1 on faction_base_farm_0, setting terrain to t_dirtmound (from t_dirt) at (4,4) when item ash existed.  Resolve this either by removing the terrain from this mapgen, adding suitable removal commands to the mapgen, or by adding a suitable flag to the innermost mapgen: either ERASE_ALL_BEFORE_PLACING_TERRAIN if you wish terrain to replace everything previously on the tile or ALLOW_TERRAIN_UNDER_OTHER_DATA if you wish the other items to be preserved

 FUNCTION : virtual void jmapgen_terrain::apply(const mapgendata&, const jmapgen_int&, const jmapgen_int&, const string&) const
 FILE     : src/mapgen.cpp
 LINE     : 2604
 VERSION  : f97c81e
```

### LINKED ISSUE #63281: Basecamp build farm on flowers throws error
### Describe the bug

Making farm makes farm tiles. An error is thrown if farm tiles are to be made under spurge (and other) flowers.

The ground is still plowed though.

Errors from save below:
```
FIRST ERROR

 DEBUG    : In nested mapgen bc_prim_farm_long_fields in update mapgen faction_base_field_camp_farm_1 on faction_base_farm_0, setting terrain to t_dirtmound (from t_grass_dead) at (14,4) when furniture was f_flower_spurge.  Resolve this either by removing the terrain from this mapgen, adding suitable removal commands to the mapgen, or by adding a suitable flag to the innermost mapgen: either ERASE_ALL_BEFORE_PLACING_TERRAIN if you wish terrain to replace everything previously on the tile or ALLOW_TERRAIN_UNDER_OTHER_DATA if you wish the other items to be preserved

 FUNCTION : apply
 FILE     : C:\Users\Tonda\Documents\GitHub\Cataclysm-DDA\src\mapgen.cpp
 LINE     : 2791
 VERSION  : 0.F-13127-g9864b457e4-dirty

SECOND ERROR

 DEBUG    : In nested mapgen bc_prim_farm_long_fields in update mapgen faction_base_field_camp_farm_1 on faction_base_farm_0, setting terrain to t_dirtmound (from t_grass) at (22,4) when furniture was f_maianthemum_stellatum.  Resolve this either by removing the terrain from this mapgen, adding suitable removal commands to the mapgen, or by adding a suitable flag to the innermost mapgen: either ERASE_ALL_BEFORE_PLACING_TERRAIN if you wish terrain to replace everything previously on the tile or ALLOW_TERRAIN_UNDER_OTHER_DATA if you wish the other items to be preserved

 FUNCTION : apply
 FILE     : C:\Users\Tonda\Documents\GitHub\Cataclysm-DDA\src\mapgen.cpp
 LINE     : 2791
 VERSION  : 0.F-13127-g9864b457e4-dirty
```

### Steps to reproduce

Without the save (probably)
1. Fill a chunk with flowers.
2. Make a farm on that chunk.

With save:
1. Load save (below).
2. Retrieve the NPC making the farm.

### Expected behavior

No error.

- OS: Windows
- Game Version: 0.F-13127-g9864b457e4-dirty [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

[Ransom Canyon.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10501747/Ransom.Canyon.zip)

cake-pie:
#### Summary
Infrastructure "Finer-grained mapgen flags for terrain wiping"

#### Purpose of change
There are multiple outstanding issues related to this debug message:
https://github.com/CleverRaven/Cataclysm-DDA/blob/7216560e9acd3cbe7d2976a3489965fb3f481ada/src/mapgen.cpp#L2783-L2791

More options beyond "allow all" and "erase all" would allow finer grained control over the disposition of preexisting furniture, traps and items on tiles that are being changed in a mapgen.

Resolves #57418, resolves #63281

#### Describe the solution

A tile being changed by mapgen may contain preexisting furniture, traps and items. Existing infrastructure from #54344 provides flags `ALLOW_TERRAIN_UNDER_OTHER_DATA` and `ERASE_ALL_BEFORE_PLACING_TERRAIN` to either ignore all of them, or erase all of them before setting the new terrain.

Besides ignore/erase, this PR adds the possibility to "dismantle" furniture and traps. With this option, furniture would be deconstructed or bashed (preferring the former) while traps are disarmed; this would yield some items as compared to the outright magical deletion of the "erase" option.

A new flag `DISMANTLE_ALL_BEFORE_PLACING_TERRAIN` will tell mapgen to use the dismantle option on furniture and traps, and ignore (allow) items, including those produced from dismantling stuff.

The three aforementioned flags ("shorthand flags") are mutually exclusive.

A further eight flags allow finer control by setting the behavior for furniture, traps and items individually:
- `ALLOW_TERRAIN_UNDER_FURNITURE`, `DISMANTLE_FURNITURE_BEFORE_PLACING_TERRAIN` and `ERASE_FURNITURE_BEFORE_PLACING_TERRAIN` are mutually exclusive.
- `ALLOW_TERRAIN_UNDER_TRAP`, `DISMANTLE_TRAP_BEFORE_PLACING_TERRAIN` and `ERASE_TRAP_BEFORE_PLACING_TERRAIN` are mutually exclusive.
- `ALLOW_TERRAIN_UNDER_ITEMS` and `ERASE_ITEMS_BEFORE_PLACING_TERRAIN` are mutually exclusive.

These eight flags can be used in conjunction with any of the three shorthand flags to override behavior for furniture/traps/items specifically. Alternatively, omit the shorthand flags and specify the behavior for furniture/traps/items individually using a combination of the finer grained flags.

Not all combinations necessarily make sense, so a warning will be output via debugmsg if they are encountered.

As was already the case before, if there is no explicit behavior defined and mapgen encounters existing furniture/traps/items, it will trigger a debugmsg.

#### Additional details

Dismantle option ignores any tool/skill/strength/etc requirements for deconstruct/bash of furniture and disarming of traps; it is guaranteed to succeed.

Deconstruct/bash of furniture may result in a different furniture. To avoid any unexpected looping, dismantle will make at most 10 successive attempts, after which outright erasure is used if the tile still contains furniture. This should be more than enough even for multi-step deconstructions.

Added call to `map::delete_signage` when furniture is erased/dismantled, in case it was a sign. This is needed because signs actually store their writing data in submap, not in the furniture, so that needs to be removed to prevent the writing from mysteriously reappearing if/when a sign is rebuilt in the same spot later.

#### Todo
- [x] add/update documentation
- [x] apply to camp farm plot mission mapgens

Add only `DISMANTLE_ALL_BEFORE_PLACING_TERRAIN` to the existing two flags.

#### Testing
Unit tests aren't complaining.

Save files provided in #57418 and #63281 were tested, and no longer trigger a debug message.  
#63281 demonstrates successful removal of `f_flower_spurge` by bashing (no yield)  
#57418 demonstrates items (ash) allowed to remain on plowed farm plots

### DIFF
--- a/data/json/mapgen/basecamps/expansion/modular_farm/version_1/primitive_farm.json
+++ b/data/json/mapgen/basecamps/expansion/modular_farm/version_1/primitive_farm.json
@@ -36,6 +36,7 @@
         "      ",
         "      "
       ],
+      "flags": [ "DISMANTLE_ALL_BEFORE_PLACING_TERRAIN" ],
       "palettes": [ "acidia_camp_palette" ]
     }
   },
@@ -43,10 +44,15 @@
     "type": "mapgen",
     "method": "json",
     "nested_mapgen_id": "bc_prim_farm_short_fields",
-    "object": { "mapgensize": [ 2, 2 ], "rows": [
+    "object": {
+      "mapgensize": [ 2, 2 ],
+      "rows": [
         "mm",
         "mm"
-      ], "palettes": [ "acidia_camp_palette" ] }
+      ],
+      "flags": [ "DISMANTLE_ALL_BEFORE_PLACING_TERRAIN" ],
+      "palettes": [ "acidia_camp_palette" ]
+    }
   },
   {
     "type": "mapgen",
--- a/doc/MAPGEN.md
+++ b/doc/MAPGEN.md
@@ -417,22 +417,45 @@ Example:
 ```
 
 Currently the defined flags are as follows:
+- Clearing flags for layered mapgens: see [dedicated section below](#clearing-flags-for-layered-mapgens)
+- `NO_UNDERLYING_ROTATE` The map won't be rotated even if the underlying tile is.
+- `AVOID_CREATURES` If a creature is present, terrain, furniture and traps won't be placed.
 
-* `ERASE_ALL_BEFORE_PLACING_TERRAIN` and `ALLOW_TERRAIN_UNDER_OTHER_DATA` are
-  mutually exclusive flags that can be used with any mapgen which is layered on
-  top of existing terrain.  This can be update mapgen, nested mapgen, or
-  regular mapgen with a predecessor.  It specifies the behaviour to follow when
-  an existing terrain is changed by the update, but the tile has existing
-  items, trap, or furniture on it.  If neither flag is provided this is an
-  error.  If `ERASE_ALL_BEFORE_PLACING_TERRAIN` is given then any items, trap,
-  or furniture will be removed before changing the terrain.  If
-  `ALLOW_TERRAIN_UNDER_OTHER_DATA` is given then they will be retained without
-  an error.  If you require more fine-grained control over this behaviour than
-  can be provided by these flags, then each of these things can be removed
-  either individually or together.  See the other entries below, such as
-  `remove_all`.
-  `NO_UNDERLYING_ROTATE` The map won't be rotated even if the underlying tile is.
-  `AVOID_CREATURES` If a creature is present terrain, furniture and traps won't be placed.
+### Clearing flags for layered mapgens
+Some mapgens are intended to be layered on top of existing terrain.  This can be update mapgen,
+nested mapgen, or regular mapgen with a predecessor.  When the mapgen changes an existing terrain,
+the tile may already contain preexisting furniture, traps and items.  The following flags provide
+a mechanism for specifying the behaviour to follow in such situations.  It is an error if existing
+furniture, traps or items are encountered but no behaviour has been given.
+
+A blanket policy can be set using one of these three (mutually exclusive) shorthand flags:
+- `ALLOW_TERRAIN_UNDER_OTHER_DATA` retains preexisting furniture, traps and items without triggering
+  an error.
+- `DISMANTLE_ALL_BEFORE_PLACING_TERRAIN` causes any furniture to be deconstructed or bashed, while
+  traps are disarmed.  The outputs, along with any other preexisting items, are then retained.
+- `ERASE_ALL_BEFORE_PLACING_TERRAIN` removes all preexisting furniture, traps and items before
+  changing the terrain.
+
+For finer-grained control, the following flags can be used to set different behaviors for furniture, traps and items:
+- `ALLOW_TERRAIN_UNDER_FURNITURE`, `DISMANTLE_FURNITURE_BEFORE_PLACING_TERRAIN` and `ERASE_FURNITURE_BEFORE_PLACING_TERRAIN`
+  are mutually exclusive flags for determining the disposition of furniture.
+- `ALLOW_TERRAIN_UNDER_TRAP`, `DISMANTLE_TRAP_BEFORE_PLACING_TERRAIN` and `ERASE_TRAP_BEFORE_PLACING_TERRAIN`
+  are mutually exclusive flags for determining the disposition of traps.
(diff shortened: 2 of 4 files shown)


andrei8l:
Feels like this would work better with a dismantle op for setmap (similar to `JMAPGEN_SETMAP_BASH`) and chaining mapgens if the operation needs to run before the terrain or furniture phases.

Also, is there an use case for these? (actual or planned)

cake-pie:
I just did the obvious thing and expanded the existing implementation where it already existed.

Use cases #57418 #57631 #63281 to name a few, where it would make sense to have furniture and traps erased/dismantled but keep items, to not rob the player of potential resources.

PatrikLundell:
Where would you place the items resulting from dismantling? If they're left in place, they may well end up inside walls (although things are still accessible there, but may well be removed automatically by the wall placement logic...).

How would "furniture" without any dismantling descriptions be handled? Bashing? An example of such problematic "furniture" is the various kinds of plants that are treated as "furniture".

cake-pie:
> Where would you place the items resulting from dismantling? If they're left in place, they may well end up inside walls (although things are still accessible there, but may well be removed automatically by the wall placement logic...).

Dismantle outputs are left in place. I am aware of the [concerns you voiced](https://github.com/CleverRaven/Cataclysm-DDA/pull/54344#issuecomment-1011906560) in #54344 regarding camp constructions placing walls that stomp out stuff in loot zones;  this PR isn't equipped to address that yet. I'm not sure what the best approach would be -- do we move stuff "out of the way", and if so, how do we determine where is sufficiently "out of the way"?

FYI, the preexisting implementation checks if the mapgen intends to place new terrain that is a "boring wall" (`TFLAG_WALL` and not `TFLAG_PLACE_ITEM`) in which case it actually disregards whatever flags there are -- behavior is forced to `ERASE_ALL...` in that case. So yes, as things currently stand, stuff may end up being erased anyway.

What this PR delivers is an improvement over the status quo for use cases like farm plot missions. It also helps facilitate the task of creating a mechanism for "moving stuff out of the way" -- with the dismantling already taken care of, all that remains to be done is shunting a pile of items (question is, where to?)

> How would "furniture" without any dismantling descriptions be handled? Bashing? An example of such problematic "furniture" is the various kinds of plants that are treated as "furniture".

> deconstructed or bashed (preferring the former)

I deliberately use the word "dismantling" which is different from the actual processes we have -- deconstruct, bash, and disarming -- it encompasses all of those.

PR #63572, opened 2023-02-13
Fix AIM container view

SurFlurer:
#### Summary
Bugfixes "Fix AIM container view"

#### Purpose of change
Originally, the container view of AIM is malfunctioning, showing only the first item in the container. The fix is to make AIM correctly show all items in the container.

#### Describe the solution
Use the same way as dealing with the corpses.

No.

#### Testing
All tests passed. In-game testing is also good.

#### Additional context
Before:
image
image

After:
image

### DIFF
--- a/src/advanced_inv_pane.cpp
+++ b/src/advanced_inv_pane.cpp
@@ -224,7 +224,7 @@ void advanced_inventory_pane::add_items_from_area( advanced_inv_area &square,
                 item_location container = square.get_container( in_vehicle() );
                 size_t item_index = 0;
                 for( const std::vector<item_location> &it_stack : item_list_to_stack( container,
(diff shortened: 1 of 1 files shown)


SurFlurer:
Found that moving items out of the container is still malfunctioning. Will figure out.

SurFlurer:
It looks like that when moving items out, the legacy_front() of the container will mistakenly report a null item, causing the container to be considered invalid.

Qrox:
IIRC legacy_front is deprecated and should be gradually replaced with newer container methods. It was from before the nested container update when containers only contained a single item.

andrei8l:
#62547

SurFlurer:
@andrei8l Ohhhh I missed that. Thank you for reminding me! I'll close this then.

PR #63575, opened 2023-02-13
Keep updating sidebar with AIM active

### LINKED ISSUE #63368: AIM messages usually aren't printed to sidebar until it's closed
### Describe the bug

When moving individual items around using the AIM, messages are printed in the message box at the top of the AIM but not printed in the sidebar. When the AIM is closed, the sidebar is suddenly filled with all those messages at once. This does not occur when using Move All or when moving to All Surrounding Squares. It also doesn't occur if there's a nearby interruption, like an enemy, which thankfully makes this bug pretty minor.

This looks to be an unintended consequence of #62400 - it doesn't show up in the release immediately before that PR was merged.

### Steps to reproduce

1. Start default scenario.
2. Go next to some items and open the AIM. Make one pane the items' tile and the other either another tile or the inventory.
3. Move things around. Note that the AIM message box updates but the sidebar does not.
4. Close the AIM. The sidebar should suddenly update with all of the messages.
5. Open the AIM again and press ',' to move all items from one pane to another. Both the sidebar and the AIM messages should update properly without needing to close the AIM.
6. Press 'a' to make one pane All Locations. Move items from the other pane to a tile and it should update both places.
7. Debug spawn a zombie near the shelter NPC and move individual items. The sidebar should occasionally update as the zombie and the NPC fight, without needing to close the AIM.

### Expected behavior

I'd expect the sidebar to always update as things are moved.

### Screenshots

_No response_

- OS: Windows
- Game Version: 59f5604 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

Kamayana:
Add invalidate_main_ui_adaptor to AIM display

#### Summary
Interface "Keep updating sidebar with AIM active"

#### Purpose of change
Fixes #63368

#### Describe the solution
Implement's [andrei8l's suggestion](https://github.com/CleverRaven/Cataclysm-DDA/issues/63368#issuecomment-1410851907).

#### Testing
I did all of the steps I reported in that issue and saw that the sidebar was updating through all of it. Didn't notice any issues, in any sidebar layout. This change does increase ui_manager::redraw_invalidated()'s processing time from ~6ms to ~19ms, but updating the sidebar was bound to add some processing cost anyway, and it feels like an acceptably imperceptible increase to me. If I'm mistaken in that assumption let me know.

### DIFF
--- a/src/advanced_inv.cpp
+++ b/src/advanced_inv.cpp
@@ -1635,6 +1635,7 @@ void advanced_inventory::display()
 
         if( ui ) {
             ui->invalidate_ui();
+            g->invalidate_main_ui_adaptor();
             ui_manager::redraw_invalidated();
         }
 


Raikiri:
Huge. Transferring vehicles can take literally hours and not updating the minimap during that whole time can be disasterous.

PR #63579, opened 2023-02-13
item: expand contents name if they're all of the same type

andrei8l:
#### Summary
None

#### Purpose of change
Container names are expanded only if they contain a single item. With the upcoming demise of charges, container names become significantly less informative.

#### Describe the solution
Expand the container name if all of its contents are of the same type

N/A

#### Testing
I wrote this for #60885 but it's occasionally useful in current master too:

Before: the container has to be uncollapsed to see its single-type contents

Screenshot from 2023-02-13 20-31-57

After:  the container offers more information at a glance when it makes sense, even for contents without charges
Screenshot from 2023-02-13 20-31-51

#### Additional context
Testing shows this has a negligible performance impact.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -275,6 +275,20 @@ struct scoped_goes_bad_cache {
 
 const int item::INFINITE_CHARGES = INT_MAX;
 
+namespace
+{
+
+bool contents_only_one_type( item const *it )
+{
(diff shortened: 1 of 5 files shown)


Zireael07:
If charges are on the way out, how are you gonna represent lots of items, e.g. glass shards, ammo and the like?

PR #63580, opened 2023-02-13
Inventory UI performance pass

andrei8l:
#### Summary
None

#### Purpose of change
People often complain that the inventory UI is slow and the upcoming demise of charges is bound to make it worse

#### Describe the solution
Address the major bottlenecks identified thanks to #60885.
See commit messages for details as they're self-explanatory

N/A

#### Testing
Time to open the `pickup_all` UI, using the debug timer from 2c9c859e9d5da5063cc923088f05eb118313f89e and an `-O3` build:

[City of the Dalles.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10725265/City.of.the.Dalles.zip) Super-extreme scenario created by stacking all the loot from a Military Base warehouse into a 3x3 spot.
Before:
O3 extreme before
After:
O3 extreme after

[Strathcona.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10725305/Strathcona.zip) A more moderate scenario where the loot is roughly equivalent to a few boxes of sugar/flour/salt after #60885
Before:
O3 moderate before
After:
O3 moderate after

Verify that the pickup UI looks and functions exactly the same as before.

#### Additional context
N/A

### DIFF
--- a/src/inventory_ui.cpp
+++ b/src/inventory_ui.cpp
@@ -1985,14 +1985,18 @@ void inventory_selector::rearrange_columns( size_t client_width )
     const inventory_entry &prev_entry = get_highlighted();
     const item_location prev_selection = prev_entry.is_item() ?
                                          prev_entry.any_item() : item_location::nowhere;
-    while( is_overflown( client_width ) ) {
-        if( !own_gear_column.empty() ) {
-            own_gear_column.move_entries_to( own_inv_column );
-        } else if( !map_column.empty() ) {
-            map_column.move_entries_to( own_inv_column );
-        } else {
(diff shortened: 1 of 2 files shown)


andrei8l:
One of the performance increases turned out to be a little bit bogus due to a [really dumb mistake](https://github.com/CleverRaven/Cataclysm-DDA/compare/3471447196f6aaa5aa5d1d20930bee215d8571fb..ef5b2d34c583daab52724f58a8da1ae4d4fd917b) on my end, but most of the performance benefit is still there. I've updated the screenshots.

andrei8l:
Thanks for merging!

PR #63582, opened 2023-02-13
Obsoletes NPC campgrounds

FarkasKBence:
#### Summary

None

#### Purpose of change

A few years ago, I made a copy of the campground location for a planned NPC faction. That project was abandoned, but the location lingered on. This PR finally obsoletes it.

#### Describe the solution

Moves NPC campgrounds and related stuff to the obsoletion file. This also gets rid of the fan-favorite prime-condition RVs, so please don't hurt me for that.

None.

#### Testing

World loads without error.

#### Additional context

None.


MNG-cataclysm (approved):
> This also gets rid of the fan-favorite prime-condition RVs, so please don't hurt me for that.

Them's fightin' words! Stick up yer dukes and we'll settle this!

On a more serious note, the overmap special should probably be migrated to a regular campground in order to prevent conflict with old 0.F saves for the upcoming stable release.

I just remembered, No Hope has a copy of the mapgen, terrain, and specials contained within it. You should remove those too.

MNG-cataclysm:
Out of curiosity, what project was this going  to be a part of?

LovamKicsiGazsii:
> On a more serious note, the overmap special should probably be migrated to a regular campground in order to prevent conflict with old 0.F saves for the upcoming stable release.

I suppose this can get merged after the stable release. There never was a rush to get this obsoleted after all.

> I just remembered, No Hope has a copy of the mapgen, overmap terrain, and special contained within it. You should remove those too.

You're right. I always forget to check No Hope...
Although I don't seem to find the duplicate of the OM terrains or the special.

> Out of curiosity, what project was this going to be a part of?

It was for an early version of God's Community.

Raikiri:
I kind of liked those as-is. The archery spawn is thematic and very tactical, a bunch of vehicles to play with. The pristine RV is obviously out of place, but everything else I kind of liked.

LovamKicsiGazsii:
> Or wait, was only the copy of campgrounds obsoleted and does the original campgrounds still stay?

That's right.

PR #63587, opened 2023-02-14
Don't display unrepairable entries in repair ui

### LINKED ISSUE #63585: Disallow un-repairable or un-reinforcable items to show up in the tailor and sewing kit's repair menu

I need to scroll through a list of unrepairable or un-reinforcable items that are already at full durability when I intend to repair or reinforce items. This would be a QoL improvement to the sewing/repairing/reinforcing menu.

### Solution you would like.

All items that cannot be repaired or reinforced should be filtered out from the repair list when using the tailor or sewing kits

I could sort all the items that appear in the repair and reinforcement menu into another pile. This way will be a waste of time, but save me time when scrolling the repair menu.
I could just continue to press 10 keys to get to my item (pgdn->pgdn->pgdn->pgdn->dn->dn->dn->dn->enter->repair until reinforced).

### Additional context

_No response_

irwiss:
#### Summary
None

#### Purpose of change

Fixes #63585

#### Describe the solution

Use repair_item_actor to perform the check

#### Testing

Spawn tailor kit + cloth item - like a t-shirt, debug wish high tailor skill, reinforce an item, see it disappears from repair menu

### DIFF
--- a/src/activity_handlers.cpp
+++ b/src/activity_handlers.cpp
@@ -2239,8 +2239,8 @@ void repair_item_finish( player_activity *act, Character *you, bool no_menu )
         // Print message explaining why we stopped
         // But only if we didn't destroy the item (because then it's obvious)
         const bool destroyed = attempt == repair_item_actor::AS_DESTROYED;
-        const bool cannot_continue_repair = attempt == repair_item_actor::AS_CANT ||
-                                            destroyed || !actor->can_repair_target( *you, *fix_location, !destroyed );
+        const bool cannot_continue_repair = attempt == repair_item_actor::AS_CANT || destroyed ||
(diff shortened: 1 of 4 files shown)


Qrox:
Does this show an item when it can in theory be repaired but there is not enough material / the player has insufficient skills? I think it is better to display a denial message instead of hide the item in those cases.

irwiss:
> Does this show an item when it can in theory be repaired but there is not enough material / the player has insufficient skills? I think it is better to display a denial message instead of hide the item in those cases.

There don't seem to be skill checks but the material checking needed a skip

irwiss:
All the gun stuff is excluded from this repair ui into iuse::gun_repair
https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/game_inventory.cpp#L1829

PR #63594, opened 2023-02-14
item: cache inherited flags

andrei8l:
#### Summary
None

#### Purpose of change
Increase Inventory UI performance some more by caching item inherited flags.

#### Describe the solution
Cache the inherited flags when contents are changed. Don't inherit recursively since there are (currently) no uses for that.

N/A

#### Testing
Using #63580 as a baseline with [City of the Dalles.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10725265/City.of.the.Dalles.zip),

Opening the pick up all menu, before:
before

After:
Screenshot from 2023-02-14 21-07-00

A few unit tests rely on flag inheritance directly or indirectly.

Exosuit still (dis)inherits flags from modules.

UPS modded tools still draw power from UPS.

#### Additional context
N/A

### DIFF
--- a/src/character_attire.cpp
+++ b/src/character_attire.cpp
@@ -68,8 +68,7 @@ ret_val<void> Character::can_wear( const item &it, bool with_equip_change ) cons
     if( it.has_flag( flag_INTEGRATED ) ) {
         return ret_val<void>::make_success();
     }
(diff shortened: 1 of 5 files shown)


andrei8l:
Thanks for merging.

PR #63597, opened 2023-02-14
Unifies gutter terrains

FarkasKBence:
#### Summary

None

#### Purpose of change

I've always wondered why we were using four terrains — and therefore four different map symbols — with virtually identical definitions to add rain gutters.  Those map symbols could've been used for much better stuff if we really only need to use one.

#### Describe the solution

Merged the four rain gutters into one, unified gutter terrain: `t_gutter`. Added aliases to avoid a magnitude of errors. Applied changes to the subway roof as proof of concept.

Three extra map symbols may not sound much, but with more complicated locations, every symbol matters.

1. continue using four symbols to define one type of terrain
2. keeping the four terrains, but only using one of them

#### Testing

World loads without errors. Not seeing any visual errors either.

kép
kép


Procyonae:
Would a PR that changed every map using roof_palette to use "-" over "|", "2" and "3" be appreciated? (I'd probably write something external to do it unless I can figure out how to do some crazy REGEX replace in files through notepad++)

LovamKicsiGazsii:
Yes, please.

Issue #61886, opened 2022-10-23
Make waterwheels a power generating appliance


yoshiquest:
### Is your feature request related to a problem? Please describe.

Currently (assuming I haven't missed something) one is able to hook up a power generating waterwheel to a vehicle to generate power, but can't just construct it standalone right next to a river. You can even build water mills, which are very similar, but not power generating water wheels!

### Solution you would like.

Add the existing water wheel and large water wheel as buildable appliances next to the water. Pretty sure power generation can be kept the same too.

### Describe alternatives you have considered.

Continue having to make dummy vehicles I guess? But I'm pretty sure the appliance system was meant to prevent people from needing to do that....

### Additional context

This could also potentially be expanded upon with additional applications, such as an even larger appliance-only waterwheel, or even a buildable dam (might need a better simulation of water and pressure before doing that though).

PatrikLundell:
The reason I backed off on making a waterwheel appliance when I implemented a wind turbine one is that appliances are single tile "machines", and water wheels only work when placed over flowing water. When building a water wheel "vehicle" you have to start with a tile on land that acts as an anchor, but with an appliance you'd have to somehow allow it to just magically hover over the water without any support. The back off happened when I realized the issue during testing, as I didn't find levitating machinery to be appealing.

yoshiquest:
@PatrikLundell
Aren't there water mill constructions now though? Wouldn't it be possible to make it work similarly to those? (Haven't actually tried building a mill in game though)

PatrikLundell:
I think there are water mill constructions (I've never build one either), but I haven't investigated how they work. They're "furniture" however, not appliances.

kevingranade:
No apparent way forward here, needs a multi-tile appliance... somehow?

PR #63603, opened 2023-02-15
Adds Water Wheels & XL Water Wheels as appliances

KKDavion:
#### Summary
Content "Adds Water Wheels & XL Water Wheels as placeable appliances."

#### Purpose of change

Wind Turbines & Solar Panels (and their variants) can all be placed on vehicles to charge their batteries, as well as placed as appliances to power other appliances. As pointed out in the New Player Questions reddit thread and the Fixed issue #61886, Water Wheels can be placed on vehicles, but not placed as appliances. This fixes this discrepancy between water, wind, and fire.

#### Describe the solution

Adds entries to the relevant json files (appliances, construction, & construction_groups) to allow for valid placement of both the regular and XL water wheels in flowing shallow water. Copied-from the vehicle part where I could, and generally copied the text from the wind turbine & its XL variant.

Going with the lukewarm consensus in the issue and not adding said parts.

#### Testing

Made changes in my own build. Created a new world & custom character. Debugged in the skills, tools, & both wheels necessary to install the parts. Installed both in flowing shallow water and observed both having positive power. Plugged each into a battery to confirm that the battery was charging.

#### Additional context

The issue linked had two comments from PatrickLundell (and what seemed like a small agreement from Kevin) that floating, non attached water wheels were aesthetically unpleasant compared to the current needed method of placing a frame & the wheel down. While I agree on the aesthetics, the water mill item & furniture exists. The water mill takes equivalent crafting materials as the regular sized water wheel, is around the same size in weight & volume, and in reality is functionally the exact same *thing*, just using the rotational energy of the water wheel to power a grindstone or hammer, rather than rotating an electromagnet to generate electricity. In my opinion, while one exists, the other must as well, and rather than removing the mill for aesthetic reasons, I'd rather add the wheels.

I think, in the future, the wheels could be changed to be multi-tile furniture/appliances. The wheel placed in the water generates mechanical "power", that can then be used to power mills, grindstones, generators, etc.. Maybe that's too Dwarf Fortress inspired, but it's also beyond the scope of this simple PR and my limited coding abilities.

Additionally, I realized while making this PR that water wheels are *really strong*. They generate something like 80 and 160 power, compared to the 50 & 100 of solar panels, & the ~60 & ~160 of wind turbines in a *strong gale*. While turbines *can* match the power output of the wheels and are generally equivalent in crafting costs, you're not going to consistently get high winds unless you build high up. Water Wheels are more consistent and constant than solar panels, are easier to craft, and generate the power of a wind turbine in ~20 mph winds (10 m/s is considered a whole gale, which is ~22 mph). Maybe that's realistic in terms of efficiency when looking at turbines as small as we are (can't imagine the diameter of the blades/wheel being >10' or so), but it could be something to look at. Is anyone aware of the conversion ratio between a battery charge and kWh, if there is one? To keep the scope of this PR relatively small & focused I deliberately didn't touch any numbers & balance, but I might draft one up after some reading & if there's any real need to. 

Additionally additionally, the construction recipes for placing the water wheel, XL water wheel, & XL turbine could be adjusted to better match the installation requirements for the vehicle part (2,4, and 3 mechanics skill, respectively), but again changing existing parts is out of the scope here and I'd rather things be consistently wrong than inconsistently correct for now. That'll come with the above mentioned potential future PR, in all likelihood.


KKDavion:
> You are creating a pull request with the master branch as the head branch. This is likely a mistake unless you really know what you are doing. You may read https://docs.github.com/en/get-started/quickstart/contributing-to-projects#creating-a-branch-to-work-on for a typical workflow of contributing to a project on GitHub.

Aha, first time user issues. I see what I did wrong (should've created a branch for my changes in my repo rather than editing the master). I don't think that should be an issue for this single PR? But I'll keep that in mind for the future.

NetSysFire:
Yes, changing it now for this PR is pretty hard to impossible. But it will cause issues in the future.

KKDavion:
Understood. Should I close this PR and resubmit it using a branch then?

Also, not sure why it's failing to build, unless it's related to the above issue

NetSysFire:
You may leave this one until its merged. After that you will need to fix your master branch or otherwise it will be hard to keep it up to date.

Ignore the build failure, it looks unrelated.

```
0058:err:explorer:initialize_display_settings Failed to query current display settings for L"\\\\.\\DISPLAY1".
```

PR #43116, opened 2020-08-22
Make Mute also a player trait

Maleclypse:
#### Summary

SUMMARY: Features "Making Mute a player trait"

#### Purpose of change

In Magiclysm I created an aura that changes the players form to that of a werewolf.  As a con I wanted the player to be unable to communicate with NPCs for the time of the change.  I gave the aura the Mute trait.  However, mute only impacts NPC's with the trait.

#### Describe the solution

This PR seeks to make mute player impacting.  I've done very little in C++ before so I hope this is a good step in the right direction.

#### Describe alternatives you've considered

Learning nothing new ever again. Begging a dev to do this.  

#### Testing

TBD need to setup a compiling environment.

Issue #63584, opened 2023-02-14
Mute characters should at least have a basic way of communicating.


actually-the-artificer:
### Is your feature request related to a problem? Please describe.

I am always frustrated when my Mute character cannot communicate with NPCs on a rudimentary level. The feature was recently PR'd with the Churl, so why not allow Mute characters a similar way as well?

### Solution you would like.

Instead of locking a Mute character out of communication entirely, a Mute character should be able to use a phone/tablet/laptop computer to communicate with text-to-speech. Another way if the character has not yet found these devices is to use pen & paper.

### Describe alternatives you have considered.

Sign language is another option, but in the whole of the USA, less than 1% of people know sign language - 250,000 to 500,000 individuals. That number would be severely diminished to be vanishingly low Post-Cataclysm.

### Additional context

This feature will also allow a character to do something which someone would do in real life. There are multiple apps for smartphones that can do text-to-speech, and it is certain that someone who is mute would use it to speak when sign language is not understood by another party.

RenechCDDA:
You have it backwards. The recent PR to churls (still open; not merged) *adds restrictions* to their ability to communicate.

The entire point of the mute trait is the inability to communicate with NPCs. If you find this frustrating, *do not pick the mute trait*.

KHeket:
> You have it backwards. The recent PR to churls (still open; not merged) _adds restrictions_ to their ability to communicate.
> 
> The entire point of the mute trait is the inability to communicate with NPCs. If you find this frustrating, _do not pick the mute trait_.

This game is about realism, not about "If you take trait - your character magically cant do simple things". May be is will be ok for magiclysm as a magick curse, but not for non magic worlds like vanilla. So, if deaf/mute people can do such simple thing in real world - they should have ability to make it in game
And author of pr describe good solutions, I also want to add, that it will be good, if the distance of communication with phone/tablet/laptop computer/pen&paper will be reduced to 1-2 tiles, like in real world its will be very hard to read little letters on big distance

anoobindisguise:
> This game is about realism, not about "If you take trait - your character magically cant do simple things"

[...]

PR #63614, opened 2023-02-16
Add a new starting trait: Nonverbal

actually-the-artificer:
#### Summary
Content "Added Nonverbal, a new starting trait similar to Mute."

#### Purpose of change

A player who desires to play without NPCs but still be able to hear in-game has no option to do so, only being allowed to pick the "Deaf" trait. This PR remedies that by allowing a PC who can still hear, yet is unable to talk to NPCs.

#### Describe the solution

A new mutation, Nonverbal, is now in the game as a valid starting trait, with an appropriate point cost. It is -3 since it disallows talking to NPCs as is the case with Deaf, yet it still allows the PC to hear.

Not doing this - which would also render all PCs who cannot converse with NPCs deaf by default.

#### Testing

I loaded this up in my personal copy of the game and tested it works.

#### Additional context

The Mute trait  also works in game in unmodded CDDA even if added through debug, which can be seen below.
CaptureMute

CaptureMute2

see #43116 for the PR that made Mute a player trait, which inspired this PR.

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -7979,10 +7979,9 @@
     "type": "mutation",
     "id": "MUTE",
     "name": { "str": "Mute" },
-    "points": -3,
+    "points": 0,
     "valid": false,
-    "starting_trait": true,
-    "description": "Whether due to personal choice, injury, or childhood trauma, you're unable to speak.  This prevents you from interacting with NPCs."
+    "description": "Trait that makes it impossible to say anything."
   },
   {
     "type": "mutation",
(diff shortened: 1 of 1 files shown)


Maleclypse:
I’m opposed to this exact trait being a starter trait because it’s specifically designed to prevent contact with NPCs. If it becomes a starter trait people will ask for ways to contact NPCs while having the trait. I’m not opposed to a similar trait existing or migrating the existing mute trait to a different trait that makes it completely impossible to speak with NPCs but specifically this trait as it is right now shouldn’t be a permanent state of the PC.

Zireael07:
@Maleclypse people already did ask for Mute characters to communicate in #63584

Your comment above sounds completely different to the closing comment in there, which seems like you intend to leave Mute incapable of communicating while the comment above suggests you're in favor of renaming current Mute to something that does not have unfortunate ableist connotations

Maleclypse:
> @Maleclypse people already did ask for Mute characters to communicate in #63584
> 
> Your comment above sounds completely different to the closing comment in there, which seems like you intend to leave Mute incapable of communicating while the comment above suggests you're in favor of renaming current Mute to something that does not have unfortunate ableist connotations

To clarify, the current Mute should not be a permanent trait with ableist connotations. It should remain a temporary effect that completely blocks npc communication. I am fine with another similar trait that people can work around that is also a potential starting trait. It should be named something that is not Mute.

actually-the-artificer:
> > @Maleclypse people already did ask for Mute characters to communicate in #63584
> > Your comment above sounds completely different to the closing comment in there, which seems like you intend to leave Mute incapable of communicating while the comment above suggests you're in favor of renaming current Mute to something that does not have unfortunate ableist connotations
> 
> To clarify, the current Mute should not be a permanent trait with ableist connotations. It should remain a temporary effect that completely blocks npc communication. I am fine with another similar trait that people can work around that is also a potential starting trait. It should be named something that is not Mute.

No worries! I'll revert my changes and copy it to a new trait.

actually-the-artificer:
This will require more C++ work than I have knowledge of.

If someone else can actually pull this off, please feel free to do so.

PR #63621, opened 2023-02-16
Move Crazy Cata monsters to Crazy Cata files

Karol1223:
#### Summary
None

#### Purpose of change
I was incredibly annoyed that I kept having these popup for me in HHG. And honestly, it takes more backflips to have them be in mainline and not do anything than to just move them over to Crazy Cata?

#### Describe the solution
Moves the monsters and relevant spells connected to them to Crazy Cataclysm mod. The monstergroups which contained only these monsters have been removed and that one mannequin that had ``looks_like`` Shia LeBouf has been made to look like the Person instead.


NetSysFire:
I feel like the reason why this was left in is because the crazy cataclysm mod is pretty ancient and everyone is afraid to touch it. Especially because it contains those elusive granades.

Karol1223:
> I feel like the reason why this was left in is because the crazy cataclysm mod is pretty ancient and everyone is afraid to touch it. Especially because it contains those elusive granades.

To be honest, I looked around and 2/3 of these monsters didn't even appear in monstergroups in Crazy Cata. And it didn't look like Crazy Cata used the vanilla monstergroups that the monsters were unique to either. It's... weird, to say the least.

NetSysFire:
Did you mean to say "crazy" instead of "weird"?

PR #62460, opened 2022-11-29
Moved ammo, ammo types, and recipes of metal rails to Aftershock

Night-Pryanik:
#### Summary
None

#### Purpose of change
* Closes #60987.

#### Describe the solution
Moved ammo, ammo types, and recipes of metal rails to Aftershock.

#### Describe alternatives you've considered
Remove all stuff from vanilla, and call it a day.

#### Testing
None, obvious change.

#### Additional context
None.

PR #63622, opened 2023-02-16
Add ``rebar_rail`` to the obsoletion files

Karol1223:
#### Summary
Bugfixes "Add ``rebar_rail`` to the obsoletion files"

#### Purpose of change
Fixes... well, supposedly there is an issue opened about it. But I couldn't find it. Long story short there's an error message like this pestering me on my save
obraz

#### Describe the solution
Simply adds the copy of the item's json from Aftershock to vanilla obsoletion files to have it not cause errors in vanilla. Aftershock's should still overwrite it when it is loaded as the ids would be identical.

### DIFF
--- a/data/json/items/obsolete.json
+++ b/data/json/items/obsolete.json
@@ -1217,7 +1217,7 @@
     "type": "ammunition_type",
     "id": "metal_rail",
     "name": "ferrous rail projectile",
-    "default": "steel_rail"
+    "default": "rebar_rail"
   },
   {
     "id": "rebar_rail",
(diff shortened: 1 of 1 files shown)


dseguin:
Can you also add an obsolete entry for `steel_rail`? Both it and `rebar_rail` were moved to Aftershock in #62460

Karol1223:
> Can you also add an obsolete entry for `steel_rail`? Both it and `rebar_rail` were moved to Aftershock in #62460

Ah, fair point. I've not encountered ``steel_rail`` errors myself but I guess it's safe to assume they're both in the same boat when it comes to that. I'll get on it soon.

Karol1223:
@dseguin All done

PR #63623, opened 2023-02-16
Allow randomizing starting hour through options

Kromgart:
#### Summary
Features "Setting 'Initial time' world option to -1 randomizes that value"

#### Purpose of change

We could randomize player character and starting day but not the starting time. This PR adds such possibility.

#### Describe the solution

Allow setting the 'Initial time' world option to -1 (same as starting day). This forces the game to randomize the starting time, overriding scenario starting time (if necessary).

#### Testing

Generated several worlds with option set to -1 and to something else. Seems to behave properly.

#### Additional context

<details>
<summary>(done)</summary>
<del>I did not change the option text label. I'm not entirely clear on that 'string freeze' thing (would that prevent the PR to be merged?) and how the text is translated to other languages etc, so I set this PR currently as draft. Please let me know what else is needed here, or can I just update this line at options.cpp?</del>

```
to_translation( "Initial starting time of day on character generation." ),
```
</details>

### DIFF
--- a/src/scenario.cpp
+++ b/src/scenario.cpp
@@ -547,9 +547,15 @@ time_point scenario::start_of_game() const
 {
     time_point ret;
 
+    int options_start_hour = get_option<int>( "INITIAL_TIME" );
+    bool override_start_hour = options_start_hour == -1;
+    if( override_start_hour ) {
+        options_start_hour = rng( 0, 23 );
+    }
+
     if( custom_start_date() ) {
         ret = calendar::turn_zero
-              + 1_hours * start_hour()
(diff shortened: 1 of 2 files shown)


src/scenario.cpp
```diff
@@ -547,9 +547,15 @@ time_point scenario::start_of_game() const
 {
     time_point ret;
 
+    int options_start_hour = get_option<int>( "INITIAL_TIME" );
+    bool override_start_hour = options_start_hour == -1;
```
Night-Pryanik, line 551: No need for adding this new variable. You can check for `options_start_hour` directly.

Kromgart: Ok, I cleaned that up.

PR #63625, opened 2023-02-16
Globally Unique NPC Specials -REDUX-

Perset:
#### Summary
Content "Creates a mod that forces some important NPC-driven locations to spawn only once" 

#### Purpose of change

Eliminates the redundancy of several highly-unique, personalized locations with the same characters, missions, and features. Some of these locations have even been known to break each other if communicated with about the same missions. Saves map space and precious immersion.

#### Describe the solution

Adds the GLOBALLY_UNIQUE flag to the aforementioned locations, and widens the search_range for missions targeting those locations. Also increases HUB 01 and Tacoma to occurrence 50%, up from 33%. and sets the Exodii base to 50% as well.

#### Testing

I've spawned at least 10 worlds now and TP'd to various mission-givers, the main concern here being that the missions need to be able to find their now greatly endangered quarry. The specials themselves have, without fail, been unique when I searched large swathes of overmap for duplicates.

#### Additional context

This mod is not neat: it has no inheritance available to it and that means it will be finicky to maintain, and while mission search failures have been a constant problem throughout my experience playing CDDA, this mod may very well magnify them, despite my tests. To mitigate this I've also increased to the spawn likelihood of three (I think?) major specials, but the overall effect will be that these important locations will be more tightly clustered than in a vanilla world.

If all that doesn't deter you, then hopefully it will satisfy anyone who's sick of seeing Tacomas and Isherwoods and refugee centers all over their world, assuming of course that one actually plays a world long enough to notice such things, which is a pretty low bar, imo.

If you use this mod, my advice is: save before initiating at least the following missions: Deliver FEMA data, Find the Exodii base (from the liaison or the stone barn), and anything from the teamster. You can widen your search_range's found in the mission definitions file to your needs.

I've long been in the habit of saving before accepting missions anyway, because they have a long history of fucking up, in my experience.

Hopefully this will obsoleted elegantly and promptly.

Please leave any thoughts or critique, I'll be the first to say that I could use it. Here's hoping I haven't left anything embarrassing.


MNG-cataclysm:
The occupied scrap yard, homeless camp, lumbermill, and chem lab were designed to spawn multiple times, and I do not believe that these locations should be made globally unique.

Perset:
> The occupied scrap yard, homeless camp, lumbermill, and chem lab were designed to spawn multiple times, and I do not believe that these locations should be made globally unique.

Are they flagged with globally_unique?

MNG-cataclysm:
> Are they flagged with globally_unique?

No, I misread.

Perset:
> > Are they flagged with globally_unique?
> 
> No, I misread.

Yes, I remember your preoccupation with making sure your stuff spawns repeatedly in my last PR. Did you ever do anything to make them mutable or procedural in any way?

MNG-cataclysm:
> Yes, I remember your preoccupation with making sure your stuff spawns repeatedly in my last PR.
> 
> Did you ever do anything to make them mutable or procedural in any way?
> 
> Also, can you interact with multiples of the same npc type without breaking anything? Two quest lines for two identical lumbermills?

You can interact with multiple of the same NPC type, including doing missions for two different lumbermills.

Zireael07:
Why close?

PR #63627, opened 2023-02-17
Globally Unique NPC Specials -REDUX- Take Two

Perset:
#### Summary
Content "Creates a mod that forces some important NPC-driven locations to spawn only once"

#### Purpose of change

Eliminates the redundancy of several highly-unique, personalized locations with the same characters, missions, and features. Some of these locations have even been known to break each other if communicated with about the same missions. Saves map space and precious immersion.

#### Describe the solution

Adds the GLOBALLY_UNIQUE flag to the aforementioned locations, and widens the search_range for missions targeting those locations. Also increases all affected to locations to occurrence (EDIT) 100%(!!!). Also adds the (EDIT) GLOBALLY_UNIQUE flag to several more generic npc locations, reducing the chance of duplicates spawning side-by-side, though not eliminating it, sadly. These locations don't have breadcrumbs (I think), so I don't need to worry about search_range BS.

#### Testing

I've spawned at least 10 worlds now and TP'd to various mission-givers, the main concern here being that the missions need to be able to find their now greatly endangered quarry. The specials themselves have, without fail, been unique when I searched large swathes of overmap for duplicates.

#### Additional context

This mod is not neat: it has no inheritance available to it and that means it will be finicky to maintain, and while mission search failures have been a constant problem throughout my experience playing CDDA, this mod may very well magnify them, despite my tests. To mitigate this I've also increased to the spawn likelihood of three (I think?) major specials, but the overall effect will be that these important locations will be more tightly clustered than in a vanilla world.

If all that doesn't deter you, then hopefully it will satisfy anyone who's sick of seeing Tacomas and Isherwoods and refugee centers all over their world, assuming of course that one actually plays a world long enough to notice such things, which is a pretty low bar, imo.

If you use this mod, my advice is: save before initiating at least the following missions: Deliver FEMA data, Find the Exodii base (from the liaison or the stone barn), and anything from the teamster. You can widen your search_range's found in the mission definitions file to your needs.

I've long been in the habit of saving before accepting missions anyway, because they have a long history of fucking up, in my experience.

Hopefully this will obsoleted elegantly and promptly in 0.G...

Please leave any thoughts or critique, I'll be the first to say that I could use it. Here's hoping I haven't left anything embarrassing...again.


onura46:
I saw you posted on another older, related issue. #61257 is also relevant. 

Rather than dupe all this code, it seems more prudent to fix the inheritance/extensibility issues you were having with quests _and/or_, preferably, design a way to ensure that these globally unique locations won't have the problems that necessitated them being in multiple places in the first place, like ant hills, triffids, portal storms, etc. destroying them.

Perset:
> I saw you posted on another older, related issue. #61257 is also relevant.
> 
> Rather than dupe all this code, it seems more prudent to fix the inheritance/extensibility issues you were having with quests _and/or_, preferably, design a way to ensure that these globally unique locations won't have the problems that necessitated them being in multiple places in the first place, like ant hills, triffids, portal storms, etc. destroying them.

OK. I certainly will get on top of that, TY for pointing it out! 

In the meantime, why don't I just leave this here for anyone who isn't so compelled to meet the infinitely well-crafted NPC's of the game that they will see duplicates of said NPC's if they drive in a straight-ish line for three or more minutes?

data/mods/Globally_Unique_NPC_Specials/specials.json
```diff
@@ -0,0 +1,2935 @@
...
+      { "point": [ 0, 3, 4 ], "overmap": "exodii_base_x0y3z4_north" },
+      { "point": [ 1, 3, 4 ], "overmap": "exodii_base_x1y3z4_north" },
+      { "point": [ 2, 3, 4 ], "overmap": "exodii_base_x2y3z4_north" },
+      { "point": [ 3, 3, 4 ], "overmap": "exodii_base_x3y3z4_north" }
+    ],
+    "locations": [ "land", "swamp", "forest" ],
+    "city_distance": [ 2, 120 ],
+    "city_sizes": [ 1, 12 ],
+    "//": "Occurrences should drop a lot once we have some breadcrumbs leading to the base, but until then this is going to be the only way to get CBMs.",
+    "occurrences": [ 50, 100 ],
```
LovamKicsiGazsii, line 2932:
```suggestion
    "occurrences": [ 100, 100 ],
```

data/mods/Globally_Unique_NPC_Specials/specials.json
```diff
@@ -0,0 +1,2935 @@
...
+      { "point": [ 0, 1, 0 ], "overmap": "isolated_house_farm_blacksmith" },
+      { "point": [ 1, 0, 1 ], "overmap": "isolated_house_farm_roof_gunsmith" },
+      { "point": [ 0, 1, 1 ], "overmap": "isolated_house_farm_roof_blacksmith" },
+      { "point": [ 1, 1, 0 ], "overmap": "isolated_road_field_1" }
+    ],
+    "connections": [ { "point": [ 0, -1, 0 ], "terrain": "road", "connection": "local_road", "from": [ 0, 0, 0 ] } ],
+    "locations": [ "field" ],
+    "city_distance": [ 3, -1 ],
+    "city_sizes": [ 1, 16 ],
+    "occurrences": [ 50, 100 ],
```
LovamKicsiGazsii, line 2670:
```suggestion
    "occurrences": [ 100, 100 ],
```

data/mods/Globally_Unique_NPC_Specials/specials.json
```diff
@@ -0,0 +1,2935 @@
...
+      { "point": [ 11, -6, -2 ], "overmap": "microlab_sub_connector_north" }
+    ],
+    "connections": [
+      { "point": [ 10, 1, 0 ], "terrain": "road", "connection": "local_road", "from": [ 10, 2, 0 ] },
+      { "point": [ 4, -1, 0 ], "terrain": "road", "connection": "local_road", "from": [ 3, -1, 0 ] }
+    ],
+    "locations": [ "wilderness" ],
+    "city_distance": [ 3, -1 ],
+    "city_sizes": [ 1, 16 ],
+    "occurrences": [ 50, 100 ],
```
LovamKicsiGazsii, line 2651:
```suggestion
    "occurrences": [ 100, 100 ],
```

data/mods/Globally_Unique_NPC_Specials/specials.json
```diff
@@ -0,0 +1,2935 @@
...
+      { "point": [ 24, 29, 0 ], "overmap": "special_forest" },
+      { "point": [ 25, 29, 0 ], "overmap": "special_forest" },
+      { "point": [ 26, 29, 0 ], "overmap": "special_forest" },
+      { "point": [ 27, 29, 0 ], "overmap": "special_forest" },
+      { "point": [ 28, 29, 0 ], "overmap": "special_forest" },
+      { "point": [ 29, 29, 0 ], "overmap": "special_forest_thick" }
+    ],
+    "locations": [ "land", "road" ],
+    "occurrences": [ 33, 100 ],
+    "//": "Inflated chance, effective rate ~19% (n=726)",
```
LovamKicsiGazsii, line 2535:
```suggestion
    "occurrences": [ 100, 100 ],
```

data/mods/Globally_Unique_NPC_Specials/specials.json
```diff
@@ -0,0 +1,2935 @@
...
+      { "point": [ 12, 14, -4 ], "overmap": "refctr_SE3e_z-4_north" },
+      { "point": [ 13, 14, -4 ], "overmap": "refctr_SE4e_z-4_north" },
+      { "point": [ 14, 14, -4 ], "overmap": "refctr_SE5e_z-4_north" },
+      { "point": [ 7, 15, 0 ], "overmap": "road_end_south" }
+    ],
+    "connections": [ { "point": [ 7, 15, 0 ] } ],
+    "locations": [ "wilderness" ],
+    "city_distance": [ 3, -1 ],
+    "city_sizes": [ 4, -1 ],
+    "occurrences": [ 75, 100 ],
```
LovamKicsiGazsii, line 1606:
```suggestion
    "occurrences": [ 100, 100 ],
```

FarkasKBence (commented):
I made 15 test runs on different worlds to check how laggy loading a breadcrumb mission can get; the only time the game froze for >4s was when I activated the "Reach Refugee Center" mission. Everything else worked just fine. Still, I'd move these breadcrumbed locations closer together, just to make sure.

A few other notes:
* I'd include Lapin's Cabin, the biker dump, and Glooscap's Hunting Lodge in the mod. `Cabin_Lapin` is a breadcrumbed location, so I'd give it a 100/100 chance to spawn.
* Ranches probably don't need that `GLOBALLY_UNIQUE` flag. Since only one refugee center is spawning, only one ranch gets to be transformed into the Tacoma Commune.
* The pictograms at the unusual stone barns don't work if they're too far from Exodii. You either need to 1) prevent it from generating too far away from the spawn area, 2) remove the pictogram, 3) edit the mission given by the pictogram so that it wouldn't attempt to find the Exodii hub.

Perset:
> Also, I really don't want to summon a heated debate over this, but for a mod that's only purpose is to prevent NPC locations from spawning multiple times, I don't think it's unreasonable to give occupied scrap yards, homeless camps, lumber mills, and chem labs the `GLOBALLY_UNIQUE` flag. Even if they're a "bit generic", they still share the same locations, missions, and characters (although with randomized names) — which makes their repetition all the more immersion-breaking.

The UNIQUE->GLOBALLY_UNIQUE business is spicy, I guess we can mull it over. Maybe I'll msg MNG to give his two cents. I believe he made an issue about bringing mutability to npc missions/dialogue.

In general, if you think it's not too heavy handed to run them all up to 100/100, I'm not inclined to argue; did you test it out? Was there enough room on all your trials? I don't know exactly how they're distributed in the initial world spawn, does the game try to cram every one of them into your starting OM chunk or are they distributed across that chunk as well as all adjacent when forced to spawn?

As for the danger of search_range failures, you are correct in that it's still there. I kinda took this in stride, because they've been happening forever, I think. There has always been a danger of it, and the further your surround a location with explored tiles before you trigger the search, the worse it gets. Not unique to the mod, I promise, although I imagine that for these important locations it was less likely than for, say, Dana's bakery or the sex shop mission.

Also,  I would like to say TYVM for giving this "mod" your effort and opinion: I know it's ugly and low effort, but I'm pretty sure that from this point I'll be playing with some version of it on (possibly tweaked from w/e it ends up as here).

LovamKicsiGazsii:
> In general, if you think it's not too heavy handed to run them all up to 100/100, I'm not inclined to argue; did you test it out?

Yes, although I've only used it on the ones with breadcrumbs to avoid search_range failures & let the less essential ones disperse throughout the world.  It worked well, even with a pre-revealed map, though it's true that I didn't do as much testing as with previous settings.

> does the game try to cram every one of them into your starting OM chunk or are they distributed across that chunk as well as all adjacent when forced to spawn?

Strangely enough, I've never seen them naturally spawning in my starting overmap. It's always a neighbouring overmap where the five main hubs group up.

Perset:
> > In general, if you think it's not too heavy handed to run them all up to 100/100, I'm not inclined to argue; did you test it out?
> 
> Yes, although I've only used it on the ones with breadcrumbs to avoid search_range failures & let the less essential ones disperse throughout the world. It worked well, even with a pre-revealed map, though it's true that I didn't do as much testing as with previous settings.
> 
> > does the game try to cram every one of them into your starting OM chunk or are they distributed across that chunk as well as all adjacent when forced to spawn?
> 
> Strangely enough, I've never seen them naturally spawning in my starting overmap. It's always a neighbouring overmap where the five main hubs group up.

Probably some insight into how OM generation occurs, there. 

I think I've put most of what you suggested in, but I don't know what to do with the Exodii signposts. I want them to be able to do their job, but I don't know of any consistent way to restrict them to a close proximity to the possible spawn of the rocky butte unless I make them globally unique, which is slight overkill in this case lol, I like being able to find the Exodii. Is it possible to at least ensure you only get the mission once? That's only a compromise, though, and doesn't eliminate the possibility of failures.

stubkan:
This PR is badly needed.  Just started a new game, two refugee centers close enough to both be seen when zoomed out on one side.  On the other side, two isherwoods the same distance away from each other.  If I take an extended roadtrip in the country I will end up with half a dozen glooscaps.

Verisimiltude gone out the window and the game feels silly.  Arguments against this are not warranted enough IMO.

* "mission algorithm search failures have been a constant problem " - Limiting spawns won't affect this - if one spawns, the mission will find it, if two, three or more spawn, same result.  In fact. limiting spawns to 1, should reduce the odds of this bug, since the map won't be filled up with duplicate isherwoods/refugee centers freeing up space for further specials.

* "the problems that necessitated them being in multiple places in the first place, like ant hills, triffids, portal storms, etc. destroying them." As of 0.G it appears triffids, ants appear to have a vastly reduced area of effect (?) - but regardless I think the chance of a location being threatened should be a valid game outcome - the same way your character can die unexpectedly and brutally - its in the nature of a roguelike for random outcomes to severely hinder your game progress, leading to challenging gameplay choices rather than just packing up and going for a small drive when you don't like how something turned out.

Making sure there are enough npc pathways around to give breadcrumb location markers to important locations - such as the shelters around each town sending you directly to the closest refugee center would remove the need for having multiple specials to make sure the player doesn't drive away from one and end up never discovering them.

The only problem I see cropping up here would be where a player travels far away enough from the starting overmap that the mission special finder would start to have a difficult time locating an unique that spawned many overmap tiles back.  This could be mitigated by having the mission special finder check the list of generated unique specials as a first step, before attempting to seek manually in each overmap.  This PR doesn't affect code and last I saw the mission seeker has no such function...  Perhaps a future PR can introduce that.

AlexanderCecile:
> * "the problems that necessitated them being in multiple places in the first place, like ant hills, triffids, portal storms, etc. destroying them." As of 0.G it appears triffids, ants appear to have a vastly reduced area of effect (?) - but regardless I think the chance of a location being threatened should be a valid game outcome - the same way your character can die unexpectedly and brutally - its in the nature of a roguelike for random outcomes to severely hinder your game progress, leading to challenging gameplay choices rather than just packing up and going for a small drive when you don't like how something turned out.

I agree completely, it's strange how this argument keeps popping up.

Tread4:
Just bumping this PR - would hate to see it lost due to being stale since it is much needed imo.

Is there something stopping it from being merged? Is it still being worked on? Some status update would be great.

AlexanderCecile:
Worth noting: https://github.com/CleverRaven/Cataclysm-DDA/pull/66202

PR #56914, opened 2022-04-19
Adds a new evolved child zombie

DoctorBoomstick:
#### Summary
Content "Adds the swollen wretch as a new possible child zombie upgrade, plus fungus variant."

#### Purpose of change
This started life as a short joke concept on the discord chat, before I thought it’d be a genuinely possibly cool idea to add a new flying boomer-like zombie to the game, and subsequently took it upon myself to add this monster as an upgrade.

#### Describe the solution
I’ve added a zombie child upgrade called the swollen wretch, which is something of an amalgamation between a boomer zombie, a pupating zombie, and a regular child zombie. More than anything it’s a floating flesh raptor incubator, which traps decomposition caused gasses within itself, causes a chemical chain reaction where in the gas is burned within it and allows it to obtain a measure.of buoyancy, thus giving it a reason for why it can fly. It’s speed and health are very low, with the main threats being the large amount of flesh raptors it harbours within itself. In addition, I’ve also added it’s fungal form and the fungal flesh raptors that it spawns. I’ve included the regular zombie form of the wretch into the zombie upgrades file, under the child zombie grouping.

#### Describe alternatives you've considered
Letting a joke remain as a joke and not deciding to Frankenstein up this creature.

#### Testing
Spawned in a bunch of iterations of the creature, checked to see that all their properties function correctly, spawned in a few zombie children and a zombie master to observe if they would mutate into the wretch, and ran the same style of tests on the fungal form.

#### Additional context
No response

PR #63630, opened 2023-02-17
Revamps the swollen wretch.

DoctorBoomstick:
#### Summary
None
#### Purpose of change
In PR #56914, I added a new evolved child zombie dubbed the "swollen wretch." Largely speaking, this zombie had many, many flaws in both its overall design and implementation. To name but a few, issues included:
1.	It laughs heartily at physics and basic biology concepts.
2.	It’s simply a mishmash of all the cool features I thought would be cool in a new zombie variant, which I stapled together based on a Discord joke without actually considering the enemy in question or even trying to make it sensible... or really even trying to make something that wasn’t just an edgy cartoon balloon. This resulted in a flying, gas-spewing, glowing, raptor-spawning giant meat balloon that, for some ungodly reason, also exploded upon death.
3.	Its description is no less than 159 words long, which literally means it trails off screen for a good chunk and is simply the pinnacle example of edgy zombie descriptions that are overly long for the sake of being edgy.
4.	It's so out of place with reality that the graphics for it are difficult to create in such a way that it doesn't look like a cartoon hot air balloon.
For these and other reasons, I was always inclined to change the enemy to make it more in line with the theme of the game, rescuing it from being the dumbest, meme-worthy bag of floating skin this side of cataclysmic New England.
#### Describe the solution
I have reworked the enemy to become an immobile flesh raptor hatchery, adopting a similar style and principle to the bone hive SCP, although a comparison with the latter only came to mind after the rework had been completed. Fundamentally, the zombie makes use of its biomass to formulate flesh raptors within its chest cavity, utilizing organs, bones, and extraneous bits and pieces to put the raptors together. To this end, I have sawed away the zombie’s flying capabilities and explosive tendencies and removed its gas-producing potential. These changes hold true for the fungal variant as well.
I considered leaving the zombie as it is, but I simply hated it with a deep, deep passion.
#### Testing
As of yet, I haven’t observed the newly adjusted zombie within a game scenario, though I believe that all should be working fine.

### DIFF
--- a/data/json/monster_special_attacks/spells.json
+++ b/data/json/monster_special_attacks/spells.json
@@ -140,6 +140,22 @@
   },
   {
     "type": "SPELL",
+    "id": "wretch_raptor_spawn",
+    "name": { "str": "Summon Raptor" },
+    "description": "Summons a flesh raptor while dealing some damage to the zombie.",
+    "flags": [ "HOSTILE_SUMMON", "RANDOM_DAMAGE", "PERMANENT", "NO_EXPLOSION_SFX" ],
+    "valid_targets": [ "ground", "self" ],
+    "min_damage": 5,
+    "max_damage": 10,
+    "min_aoe": 3,
+    "max_aoe": 3,
+    "base_casting_time": 100,
+    "shape": "blast",
+    "effect": "summon",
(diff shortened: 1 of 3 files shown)


MNG-cataclysm (approved):
This is a more sensible change.

PR #63629, opened 2023-02-17
Remove unused enchants plus allow loading of relics after enchantments are loaded. 

Ramza13:
#### Summary
None

#### Purpose of change
Fix #63596
Fix #61701
Fix #63195
Fix several enchantments.
Clear unused code

#### Describe the solution
Fix several enchantments. 
Removed the remaining non implemented enchantments.

PR #63636, opened 2023-02-17
Pain immunity flag + flat pain modifier

Dan1ss1mo:
#### Summary
Features "Pain immunity flag and flat pain modifier for mutations"

#### Purpose of change
more json stuff is always nice

#### Describe the solution
~~Also~~ Added flat pain modifier as requested by Venera3 for future use in limb stuff and PAIN_IMMUNE flag

#### Testing
Everything works as intended, tested by inflicting pain with debug menu


alp7292:
what happens when there is two trait with pain modifier

Dan1ss1mo:
> what happens when there is two trait with pain modifier

They multiply with each other

Ramza13:
This looks like a good idea and good code.  

Awkwardly its happening at the same time I was about to add a working pain enchantment to #63629 and I think I will likely basically steal your code to do so.(I'll credit you for it).  In general when possible we want to be adding enchantments rather than fields to mutations directly because enchantments can be used by a bunch of different things like effects, bionics and items rather than just mutations.  I'm (very slowly) working on migrating the various fields from mutations/bionics into enchantments with the goal of someday having effects/mutations/bionics all be the same under the hood.

Dan1ss1mo:
> 

Well, enhancement can also add mutations :P, but any way, glad to help) I haven't look into enhancement code for now. I was just really confused than i first find out that this mutations are hard coded for some reason. I planning to also jsonify few more, but maybe i should also made them into enhancements?

Maleclypse:
> > 
> 
> Well, enhancement can also add mutations :P, but any way, glad to help) I haven't look into enhancement code for now. I was just really confused than i first find out that this mutations are hard coded for some reason. I planning to also jsonify few more, but maybe i should also made them into enhancements?

Changing them into enchantments is one of the ways to jsonify things.

Dan1ss1mo:
> > > 
> > 
> > 
> > Well, enhancement can also add mutations :P, but any way, glad to help) I haven't look into enhancement code for now. I was just really confused than i first find out that this mutations are hard coded for some reason. I planning to also jsonify few more, but maybe i should also made them into enhancements?
> 
> Changing them into enchantments is one of the ways to jsonify things.

I think having both is even better. Rather then having mutation that adds enhancement that actually does that mutation supposed to do why not have both dedicated mutation modifier and enhancment effect?

PatrikLundell:
@Dan1ss1mo: Just to clarify, the others talk about enchantments (as in magic enchantments), not enhancements (as in improvements). Enchantments are used as a way to achieve effects that are not limited to magic. Thus an enchantment could be used to provide mutation effects, magic spell effects (for suitable mods as well as portal magic), as well as enhancements from e.g. artifacts.

Dan1ss1mo:
> @Dan1ss1mo: Just to clarify, the others talk about enchantments (as in magic enchantments), not enhancements (as in improvements). Enchantments are used as a way to achieve effects that are not limited to magic. Thus an enchantment could be used to provide mutation effects, magic spell effects (for suitable mods as well as portal magic), as well as enhancements from e.g. artifacts.

Sure, i understand that "enhancment" is refering to in game mechanic)

Dan1ss1mo:
I found a few problems with flat pain modifier, will fix that in a moment

Computer-Engineer-Man:
Shouldn't this be included somewhere in the docs? Such as under JSON_FLAGS.md in the character section?

Dan1ss1mo:
> Shouldn't this be included somewhere in the docs? Such as under JSON_FLAGS.md in the character section?

Oh, I guess you are right

src/game_inventory.cpp
```diff
@@ -69,6 +69,8 @@ static const activity_id ACT_EAT_MENU( "ACT_EAT_MENU" );
 static const bionic_id bio_fitnessband( "bio_fitnessband" );
 static const bionic_id bio_painkiller( "bio_painkiller" );
 
+static const json_character_flag json_flag_PAIN_IMMUNE( "PAIN_IMMUNE" );
```
kevingranade, line 72:
```suggestion
```

PR #63639, opened 2023-02-17
Randomize freshness of comestibles on game start

Kromgart:
#### Summary
Features "Randomize age/freshness of comestibles generated at calendar::start_of_cataclysm"

#### Purpose of change

Currently all food generated on the map or dropped from monsters is considered to be created at the calendar::start_of_cataclysm: that is usually right before the game starts (unless you delay the start by scenario or world options). This is fine for canned/sealed food (it is reasonable to assume it was canned in a fresh state, and it doesn't spoil), but strange for everything else that can spoil/rot.

#### Describe the solution

When comestibles ar created on map generation, or on monster death (and probably any time the item_group creation is executed), and the item 'birthday' is calendar::start_of_cataclysm, the rot level of food (freshness) is randomized within the item's shelf_life range. The following distribution is used:
 
rot_distr

On average we should expect the following results:
50% of the food will have 25% rot (or less)
75% of the food will have 50% rot (or less)
 

Continue eating only butter for the first few weeks.

#### Testing

Generated new world, teleported into town, killed zombies and looted kitchens. Sealed/canned food is not affected, so I didn't pick it up for the screenshot. 

<details>
  <summary>Results:</summary>

list

</details>

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -7644,15 +7644,6 @@ void item::calc_rot( units::temperature temp, const float spoil_modifier,
         temp = std::min( temperatures::fridge, temp );
     }
 
-    // simulation of different age of food at the start of the game and good/bad storage
-    // conditions by applying starting variation bonus/penalty of +/- 20% of base shelf-life
-    // positive = food was produced some time before calendar::start and/or bad storage
-    // negative = food was stored in good conditions before calendar::start
-    if( last_temp_check <= calendar::start_of_cataclysm ) {
-        time_duration spoil_variation = get_shelf_life() * 0.2f;
-        rot += rng( -spoil_variation, spoil_variation );
-    }
-
     rot += factor * time_delta / 1_hours * calc_hourly_rotpoints_at_temp( temp ) * 1_turns;
 }
 
--- a/src/item_group.cpp
+++ b/src/item_group.cpp
@@ -720,8 +720,8 @@ std::size_t Item_group::create( Item_spawn_data::ItemList &list,
 
     if( birthday == calendar::start_of_cataclysm ) {
         // randomize rot value for comestibles generated at the 'start of cataclysm'
-        for( item &e: list ) {
-            e.visit_items( []( item *visit_itm, item * ) {
+        for( item &e : list ) {
+            e.visit_items( []( item * visit_itm, item * ) {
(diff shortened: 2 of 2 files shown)


Hirmuolio:
The starting rot of items with birth date at calendar::start_of_cataclysm is randomized when the item is processed for the first time after being created (should happen immediately on next turn after spawning).

https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/item.cpp#L7668

±20% with even distribution.

Kromgart:
> The starting rot of items with birth date at calendar::start_of_cataclysm is randomized when the item is processed for the first time after being created (should happen immediately on next turn after spawning).
> 
> https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/item.cpp#L7668
> 
> ±20% with even distribution. Probably should remove the old one to not do same thing in multiple places.

Thanks for pointing this out. I never suspected this was already happening. Probably because even the rare case of -20% shelflife at the game start was not noticable by the rough [Fresh / quite fresh / approaching midlife] gradation.

MochaModifier:
Some issues off hand, with what this results in:
1) people evidently walked around with candy or bags of chips in their pockets for months or years, and then became zombies - leading them to have snacks that are old or nearly rotten, days after the cataclysm occurs. Not for nothing, but chocolate candies say they spoil in a year. Why were so many people carrying the same candy in their pockets for me to be finding heaps of old candy like this?

>This is fine for canned/sealed food (it is reasonable to assume it was canned in a fresh state, and it doesn't spoil)

2) Shelf stable goods, i.e. 'sealed' goods, are being found old or near rotten. 
image

This is presumably unintended, but I keep finding sealed goods randomly (old). In some cases, they are so (old) that they rot within hours of opening.

MochaModifier:
Another thing I found, is that apparently things like jerky and pemmican aren't spawning on store shelves in sealed containers (?)

Likely worth keeping in mined that atleast in the US, stores will remove food from shelves that is past their best by date and destroy them; so it'd be unlikely to find food that lasts for a season or longer just sitting on a shelf about to rot mere days or weeks after the cataclysm hits.

I was also finding food in this state within sealed food crates - normally, food in the backroom is going to be fresh off a truck, just a few days, maybe weeks depending on the food. So much food has a short best by date, and backrooms being limited in size, retailers have to constantly cycle stock

PR #63641, opened 2023-02-18
The description for "raptor tail" makes no sense and should probably be changed

NerdCuddles:
The raptor tail is described as "lizard like" for some reason, even though dromeosaur tails are nothing like lizard tails. They have stiff interlocking bones in them, and are also a feathered fan. Neither of those are lizard like characteristics

### DIFF
--- a/data/mods/Magiclysm/mutations/debug_mutations.json
+++ b/data/mods/Magiclysm/mutations/debug_mutations.json
@@ -8,7 +8,9 @@
     "player_display": false,
     "description": "To cast bug killing spells.",
     "debug": false,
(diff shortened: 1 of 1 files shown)


Maleclypse:
Oof, you are going to want to rebase your changes on top of a current clone of the repo.  Currently you are changing infinity files in the game which is an error I rarely see.  You may want to ask for help in the #dev-help channel on the devcord.

Qrox:
The OP probably intended to create an issue because the branch of this PR is [pull/53254/head](https://github.com/CleverRaven/Cataclysm-DDA/tree/pull/53254/head) which has nothing to do with the proposed change.

dseguin:
Please open an issue instead

PR #63653, opened 2023-02-18
Sane person's lifespan

GuardianDll:
#### Summary
None
#### Purpose of change
Lifespan of person's was too long, and lot of people spotted they are often live long after the end of portal storm
#### Describe the solution
decrease their lifetime from 40- 60 minutes to 12-15
i'd like to make something so they can live only when portal storm exist, and instantly die after the end, but it seems not possible


PatrikLundell:
The solution description is misleading, as 60 is not a decrease from 40, but rather an increase. However, the actual change is a decrease from 40-60 to 12-15.

PR #63654, opened 2023-02-18
Misc debug stuff

GuardianDll:
#### Summary
None
#### Purpose of change
Few debug things that i made for my stuff
#### Describe the solution
Debug mutations for HP and HP regen, because high str mess this too much, and spell to kill any creature

### DIFF
--- a/data/json/item_spell_effects.json
+++ b/data/json/item_spell_effects.json
@@ -63,5 +63,20 @@
     "min_duration": 18000,
     "max_duration": 18000,
     "damage_type": "cold"
+  },
+  {
+    "id": "debug_kill_spell",
+    "type": "SPELL",
+    "name": "Debug Kill Creature",
+    "description": "Deal damage equal to 100% of monster's hp.",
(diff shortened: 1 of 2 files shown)


PatrikLundell:
What's wrong with the current debug means to kill off creatures? Wouldn't it be better to fix those if there's something wrong with them than introducing a new one?

GuardianDll:
It doesn't count as player kill, and i found impossible to obtain lot of experience because of this. Now i can just cast this spell to kill a person, and take a lot of xp instantly

PatrikLundell:
If you just want to get a lot of XP and kills and if there aren't useful means to do so, it would be better to create debug commands that allowed you to to gain XP and kills respectively, without creating an additional kill command with side effects. Test purposes usually benefit from as pure debug commands as possible.

GuardianDll:
Great idea, except i have no clue how to make it from code side

bombasticSlacks:
I think these are fine, you have to go looking for them 👍

PR #63657, opened 2023-02-18
Fixes fitness gym pool spawns

FarkasKBence:
#### Summary

None

#### Purpose of change

Fixes the occasionally absurd number of swimmer zombies at fitness gyms.

kép

#### Describe the solution

Replaced `place_monsters` with `place_monster`.

Repealing the no-pool-party policy.

#### Testing

Debug-spawned 9 fitness gyms.

#### Additional context

Most other locations should be using `place_monster` as well. It allows for more precise control over monster spawns.

### DIFF
--- a/data/json/mapgen/gym.json
+++ b/data/json/mapgen/gym.json
@@ -191,7 +191,7 @@
         "T": "f_toilet",
         "t": "f_sink"
       },
-      "place_monsters": [ { "monster": "GROUP_MAPGEN_POOL", "x": [ 12, 19 ], "y": [ 18, 21 ] } ],
+      "place_monster": [ { "group": "GROUP_MAPGEN_POOL", "x": [ 12, 19 ], "y": [ 18, 21 ], "repeat": [ 1, 5 ] } ],
       "toilets": { "T": {  } },
       "items": { "O": { "item": "gym", "chance": 80 }, "H": { "item": "vending_drink", "chance": 75, "repeat": [ 4, 8 ] } },
       "place_loot": [ { "group": "cash_register_random", "x": [ 17, 19 ], "y": 2 } ]


Termineitor244:
Very good! But this location (As well as some other similar ones) should rather make use of a different monster group (Or even better, use the current group and another one for the area outside the pool), I personally recommend GROUP_FOOTBALL because of the good amount of though, runner and fat zombies in there, or as a second option the GROUP_HOTEL_GYM, which is composed mainly of fat zombies...

But even better would be just a new monster group that includes said monsters (though, runner, fat) and some amount of normal zombies, so the location has better thematic monsters.

Maleclypse (approved):

LovamKicsiGazsii:
> Very good! But this location (As well as some other similar ones) should rather make use of a different monster group (Or even better, use the current group and another one for the area outside the pool), I personally recommend GROUP_FOOTBALL because of the good amount of though, runner and fat zombies in there, or as a second option the GROUP_HOTEL_GYM, which is composed mainly of fat zombies...
> 
> But even better would be just a new monster group that includes said monsters (though, runner, fat) and some amount of normal zombies, so the location has better thematic monsters.

Agreed. I didn't realize fitness gyms don't have regular zombies, nice catch!  Although I hope you don't mind if I leave that task to you. I know you're an expert at monster-spawning-related stuff, so I'm sure you'd do a much better (and less hasty) job than me.

Termineitor244:
> Agreed. I didn't realize fitness gyms don't have regular zombies, nice catch! Although I hope you don't mind if I leave that task to you. I know you're an expert at monster-spawning-related stuff, so I'm sure you'd do a much better (and less hasty) job than me.

Haha I wouldn't call me an expert, I just stare a lot to spawn groups/mapgen and like to tune them whenever I can, but thanks for the compliment! I will give gym related places a small monster spawn revamp this weekend then, it shouldn't take much time.

PR #63661, opened 2023-02-18
[Xedra Evolved] Random fixes for the voices in my head vol. 02

GuardianDll:
#### Summary
None
#### Purpose of change
Another random fixes i found and/or which someone point me in discord. string freeze edition
#### Describe the solution
added deduction to all inventor's spells
fixed bio detonator items - now they work properly, require warhead as expected, has more sane recipe (don't need to use brains now) and warhead is not 16L size
replace the mapgen modification effect to give artifact from spark of inspiration spell with much simpler u_spawn_item with use_item_group - now artifacts spawn right in the player's pockets
#### Testing
Yep, it works
#### Additional context
also listen Celldweller - First Person Shooter


Maleclypse (approved):
Thanks for these fixes!

PR #63662, opened 2023-02-19
Ferrous spawn revamp

Termineitor244:
#### Summary
Balance "Revamps the spawns of ferrous/rust zombies"

#### Purpose of change
Rust zombies currently are in a weird spot where their monstergroup spawns them as an all or nothing monster, where after a certain time every monster is a rust monster, but in early days they do not exist until that specific point in time where they replace all the other zombies.

#### Describe the solution
A small revamp of spawning for these monsters, making it a gradual process the replacement of normal zombies by rust ones, also changing the rust shell zombie into an evolution of the rust zombie, it seems like a reasonable outcome to a possible evolution of a zombie integrating metal into its body, losing ability to move and gaining a strong armor in its place... The zombies were also included in the steel mill and dump locations, makes sense.

To change the spawns of several locations that could make use of these kind of zombies other than the steel mill, but maybe later.

#### Testing
It works!

#### Additional context
The locations in which these monsters appear need a spawn revision in general, but I can't do it right now, maybe in another time, in another PR.

### DIFF
--- a/data/json/monstergroups/zombie_upgrades.json
+++ b/data/json/monstergroups/zombie_upgrades.json
@@ -184,10 +184,11 @@
     "name": "GROUP_FERROUS_UPGRADE",
     "type": "monstergroup",
     "monsters": [
-      { "monster": "mon_zombie_rust", "weight": 700 },
-      { "monster": "mon_zombie_plated", "weight": 100, "cost_multiplier": 2 },
-      { "monster": "mon_zombie_urchin", "weight": 100, "cost_multiplier": 2 },
-      { "monster": "mon_zombie_hammer_hands", "weight": 100, "cost_multiplier": 2 }
+      { "monster": "mon_zombie_rust", "weight": 600 },
+      { "monster": "mon_zombie_shell", "weight": 175 },
+      { "monster": "mon_zombie_plated", "weight": 75, "cost_multiplier": 2 },
+      { "monster": "mon_zombie_urchin", "weight": 75, "cost_multiplier": 2 },
+      { "monster": "mon_zombie_hammer_hands", "weight": 75, "cost_multiplier": 2 }
     ]
   },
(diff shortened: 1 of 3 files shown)


Maleclypse (approved):

Termineitor244:
A little comment for anyone that stumbles upon here: Pack sizes of 25 / 35 and other big quantities are not funny unless the rest of the spawn group has pack sizes of similar size, the possibility of a big number of enemies should be left to the mapgen definition with the "chance" and "repeat" field, using pack sizes this big only creates huge agglomerations of enemies in a reduced space, if you want 25 zombies, make a repeat of 25 or more in a sufficiently sized zone.

The steel mill needs a rework of its spawns because of this reason (and others), something that should be addressed in a different PR.

fairyarmadillo:
I have a suggestion that may help - make them a possible evolution route for zombie technicians, with the idea being that they magnetically pulled too many ferrous objects into their body.

There are some areas (TCL comes to mind, also open sewers etc) where groups of technicians eventually turn into groups of shockers, which are trivialized by a few common pieces of electricity-absorbing gear. Such a change would add some rusty bois to the mix and make the groups less boring and a bit more dangerous.

Termineitor244:
> I have a suggestion that may help - make them a possible evolution route for zombie technicians, with the idea being that they magnetically pulled too many ferrous objects into their body.
> 
> There are some areas (TCL comes to mind, also open sewers etc) where groups of technicians eventually turn into groups of shockers, which are trivialized by a few common pieces of electricity-absorbing gear. Such a change would add some rusty bois to the mix and make the groups less boring and a bit more dangerous.

That's actually a good idea, I will implement it in another PR though, zombie technicians and their crazy magnetism is on the chopping block so I want to hear more opinions about it with a dedicated PR.

PR #63664, opened 2023-02-19
Implement searching Chinese using pinyin

Water-Moon:
#### Summary
Interface "Pinyin matching for search in menus, to be used when the game is set to Chinese."

#### Purpose of change

The existing searching/matching function only match exactly the translation and the query string. This will work for most language from the western countries. However, it will be a problem for Chinese players, as we need input method editor (IME) to input Chinese characters. Furthermore, the game does not work well with such IMEs, and the word selection screen is invisible when typing in the game window. 

Beside this, it is also very inconvenient to switch IME during the game play. A Chinese IME usually changes the input key code received by the game, and prevent proper actions. As a result, players will need to turn their IME off when performing most interactions. It will be an inconvenience to switch IME each time a search is needed.

#### Describe the solution
The changes implemented matching via pinyin (拼音, pronunciation or spelling of Chinese characters). A pinyin match is appended at the end of [`cata_utility.cpp#lcmatch`](https://github.com/CleverRaven/Cataclysm-DDA/blob/05b28a6abcfda92273ebc1bcce629a6fa0b0b1d2/src/cata_utility.cpp#L80), which will match user's query string against all possible spelling combinations of each term. As a result, user will be able to search directly using pinyin, which is just plain letters, and thus bypassing the need to use an IME.

The search is performed by `pinyin.cpp` which is added in this PR. This file contain a lookup table for Chinese characters and their corresponding pinyin. The data is from (but not a copy of) [this repo](https://github.com/rime/rime-terra-pinyin). The pinyin data is hard-coded into `pinyin.cpp`, but since pronunciations of existing words most likely will not change, it should not be a problem.

Due to the amount of data lookup required, and the fact that some Chinese characters have multiple pronunciations, this matching can be slow. Therefore, a new item `USE_PINYIN_SEARCH` is added under the `interface` page of the options menu. This allow the user to turn it off (and it default to off) so player using other language won't have much performance impact from this new feature.

- Make the game compatible with IME so that word selection screen can show: not viable because of the multi-platform nature of this project.
- Make Chinese player use English: this is not feasible, especially since there are many terminologies that we only understand the best in Chinese.

#### Testing

Searching in keybinding menu: 
("物品" = "wupin")
image

Searching in debug menu:
("子弹" = "zidan")
image

Searching in overmap:
("图书馆" = "tushuguan"), partial match still work just like searching in English, so "tushug" can also work
image

In theory this should work everywhere, as all search call the `lcmatch` function.

Turning off the option in settings indeed disable this search:
image
image

#### Additional context
There could be a faster implementation, one that per-compute a separate search tree for all known item names. However, that might not work well for anything that has been renamed.

Also, this has been such a great game and I enjoy it so much. I hope this PR can help other Chinese players to enjoy this game even more.


Water-Moon:
Note: I am not sure if this should be delayed until after 0.G release, because it seems like 0.G string freeze is in effect and C++ PR might not be good for now. 

However I think this one did not touch much of existing code that could break, so maybe let this one go through if possible?

andrei8l:
> Beside this, it is also very inconvenient to switch IME during the game play. A Chinese IME usually changes the input key code received by the game, and prevent proper actions.

Keycode input mode was implemented exactly for this. You can use regular keybindings and still type IME'd text in string input boxes. Does it not cover your use case too?

Water-Moon:
Keycode input mode could solve some problem, but it behave inconsistently, it depend on the IME you use. Some IME keep previous key strikes and as a result the game will receive multiple input and do unexpected things, even though these input have correct key code. (The reason this happens is because multiple letters is required to input one Chinese character, and IME will have to keep the sequence to know what the user want to say)

Also even if that solve the problem with interaction, one would still not be able to use an IME properly in game to type Chinese character, the IME simply don't show. (I'm not sure if you know Chinese or not, but to simply explain this, different Chinese characters have same pronunciation, and the user will need to look at the IME window and select the correct word they want to input. So if the IME window don't show, even if we solve all other problem, it is still a pain to navigate through the vast amount of items this game provides.)

Qrox:
> Keycode input mode could solve some problem, but it behave inconsistently, it depend on the IME you use.

The key code mode is currently WIP and not implemented for some UIs, which might be where the inconsistency comes from. However, the key code mode disables IME input altogether in non-text-input UIs through the `SDL_StopTextInput` function so it should not depend on the IME used. If you have a list of IMEs that do not work with the key code input mode it would be nice if you could open an issue about it so someone can look into it.

> not be able to use an IME properly in game to type Chinese character, the IME simply don't show

This is a known problem of the official SDL build. We could build our patched version of SDL to fix it and show the IME candidate list, but I haven't got the time to look into it.

src/cata_utility.cpp
```diff
@@ -89,7 +90,14 @@ bool lcmatch( const std::string &str, const std::string &qry )
     }
     // Then try removing accents from str ONLY
     std::for_each( u32_str.begin(), u32_str.end(), remove_accent );
-    return u32_str.find( u32_qry ) != std::u32string::npos;
+    if( u32_str.find( u32_qry ) != std::u32string::npos ) {
+        return true;
+    }
+    if( get_option<bool>( "USE_PINYIN_SEARCH" ) ) {
```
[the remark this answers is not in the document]
Water-Moon: Done adding that, didn't know that was a thing, thanks for the advice!

src/pinyin.cpp
```diff
@@ -0,0 +1,493 @@
...
+    // else, there is already 1 or more entry in destination
+    // therefore append all possible combinations
+    std::vector<std::u32string> new_dest;
+
+    for( std::u32string &existing_dest_str : dest ) {
+        for( std::u32string &result_entry : result ) {
+            new_dest.push_back( existing_dest_str + result_entry );
+        }
+    }
+    dest = new_dest;
```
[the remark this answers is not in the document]
Water-Moon: Added a cache that is lazy-loaded when a string is processed once, so subsequent search will be faster. The cache then get reset when the world is unloading in case the user decide to load a different world with different mods.

It will still be slow on first search, but assuming a user does not search once per play session, it should overall bring some improvements.

Water-Moon:
Added a cache to make the search run faster on the same string after the first time. Tested on debug menu (because there is the most amount of item), and it only slow down on first key stroke, then it is fast until world is closed and opened again.

Water-Moon:
Oh and I have no idea what happened, but probably due to line ending or something it display that I changed a lot in the existing files. I'm trying to solve that now.

Zireael07:
Plz don't do anything now - your last commit shows a green tick, which means everything is ok from style point of view. IF you change things around now, you could break things again, preventing tests from working and getting a red cross again

Water-Moon:
Oh, ok

src/pinyin.cpp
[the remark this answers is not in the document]
Qrox: Well, caching does improve the speed but is going to exacerbate the memory problem because now the memory consumption not only increases exponentially with the number of multi-pronunciation characters in each string but also linearly with the number of searched strings... What I meant by lazy evaluation is that this list of possible combinations doesn't have to be computed all at once; you could calculate one at a time and compute the next only if the previous one does not match the query string. I don't know if that is needed though, maybe there are not enough edge case strings to trigger the problems I described.

Qrox:
That line ending problem is probably caused by the git line ending option. I think you should set it to commit Unix line endings but checkout Windows line endings.

src/pinyin.cpp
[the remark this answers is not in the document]
Water-Moon: Actually that is a good idea, I will try to do that, thanks for the advice!

Water-Moon:
This should do it, I implemented the way you suggested, so it will return if it already found a match.

Also I made the program build an unordered map with index being the characters. This way it don't need to loop through 400+ strings and run search on each of them.

Qrox:
I might not have made it clear enough. I'll explan with an example then.

For example, if the searched string is "长长长“, which could be any combination of a series of "chang" and "zhang", the first possible combination "changchangchang" is matched against the query string. If the query string is "changchang", it already matches and the function therefore returns.

But if the query string is "changzhang" which does not match, the next possible combination "changchangzhang" is searched. This is repeated until the query string matches one of the combinations or all combinations from "changchangchang" to "zhangzhangzhang" are searched and none matches.

This way the memory consumption only depends on the length of the searched string and not the number of possible combinations which grows exponentially when the searched string contains many characters with multiple pronunciations. However the worst search time is still exponential.

Water-Moon:
Thanks for providing the idea. Please check if this is what you mean

Water-Moon:
Actually, on my second thought, I don't know if I should just combine the option to use pinyin search with the language option. 

If I do that, then pinyin search will just be automatically enabled for ZH_CN and ZH_TW but disabled for other languages, and also that could remove the need for new translation string. It doesn't seems like too much of a performance hit, and since the game already have automatic adaption to letter with accent, it might just be natural to make pinyin search also automatic?

src/pinyin.cpp
```diff
@@ -0,0 +1,513 @@
...
+            * combination % list.size() will give one of the index for this list
+            * and then divide it by the list size ensure that for each of the next index,
+            * we have tried all possible combinations for this index
+            */
+            current_combination += cur_char_pinyin_list.at( combination % cur_char_pinyin_list.size() );
+            combination /= cur_char_pinyin_list.size();
+
+            //we count the amount of total combinations possible to determine when to stop
+            if( cur_char_pinyin_list.size() > 1 ) {
+                total_combination += cur_char_pinyin_list.size();
```
Qrox, line 498: Should this be multiplication instead of addition?

src/pinyin.cpp
```diff
@@ -0,0 +1,513 @@
...
+        for( int index = 0; index < str.length(); index++ ) {
+            std::vector<std::u32string> cur_char_pinyin_list;
+            try {
+                //try to find the pinyins for the current character
+                cur_char_pinyin_list = indexed_pinyin_map.at( str.at( index ) );
+            } catch( const std::out_of_range &err ) {
+                //not a known character
+                current_combination += str.at( index );
+                continue;
+            }
```
Qrox, line 485:
```suggestion
        for( const uint32_t ch : str ) {
            //try to find the pinyins for the current character
            if( indexed_pinyin_map.find( ch ) == indexed_pinyin_map.end() ) {
                //not a known character
                current_combination += ch;
                continue;
            }
            const std::vector<std::u32string> &cur_char_pinyin_list = indexed_pinyin_map.at( ch );
```
Uses a reference instead of a copy of `indexed_pinyin_map.at( ch )` to improve speed.

src/pinyin.cpp
```diff
@@ -0,0 +1,513 @@
...
+    // O(1) instead of O(n)
+    static std::unordered_map<char32_t, std::vector<std::u32string>> indexed_pinyin_map;
+
+    //Build the indexed map if not built yet
+    if( indexed_pinyin_map.empty() ) {
+        for( auto const &entry : pinyin_data ) {
+            //entry.first = a pinyin; entry.second = all characters that have this pronounciation
+            for( int index = 0; index < entry.second.length(); index++ ) {
+                //for each character, use it as the index and add its pinyin to the indexed map
+                char32_t current_char = entry.second.at( index );
```
Qrox, line 449:
```suggestion
            for( const char32_t current_char : entry.second ) {
                //for each character, use it as the index and add its pinyin to the indexed map
```

Qrox (commented):
Thanks for following up. Is there any noticeable improvement in speed though? There should be if there are enough multi-pronunciation characters in the strings, but I'm not sure if that's the case.

src/pinyin.cpp
Water-Moon: Oh, thank you for pointing this out, I didn't realize this is a permutation. It indeed should be a multiplication.

Water-Moon:
With most use cases, the speed improvement is not significant. However, I could verify that it does help with memory issue, as the game no longer freeze for several minute when using `d:` to search for description in recipe menu.

The most significant speed improvement happened when we migrate the look up table to a unordered map. I tested on a rather old computer (i3-5005 cpu with 4gb ram) and there is no significant freeze in most case, except the debug spawn item menu.

Water-Moon:
I have no idea why it build on my computer (and why VS studio say nothing about missing import), should be fixed now

Water-Moon:
OK, this should be the final fix needed, I tried compile with standalone `g++` and it did not report any syntax error. Sorry for all these mistakes

Water-Moon:
Oh no, formatting errors managed to slip through :(

This should fix it, hopefully this will be the last commit

I also performed a final test, verified that pinyin search is working as expected.

Water-Moon:
Oh no, formatting errors managed to slip through :(

This should fix it, hopefully this will be the last commit for this PR

I also performed a final test, verified that pinyin search is working as expected.

Water-Moon:
Oh no, formatting errors managed to slip through :(

This should fix it, hopefully this will be the last commit for this PR

I also performed a final test, verified that pinyin search is working as expected.

src/pinyin.cpp
```diff
@@ -0,0 +1,513 @@
+#include "pinyin.h"
+
+#include <algorithm>
+#include <map>
+#include <string>
+#include <cuchar>
```
BrettDong, line 6: `uchar.h` is not available on macOS, but I think it's also not needed as you are not using the conversion functions between UTF-8 and multi-byte char sequence.

src/pinyin.cpp
```diff
@@ -0,0 +1,513 @@
...
+#include <unordered_map>
+#include <vector>
+
+namespace pinyin
+{
+
+/* data is referenced from https://github.com/rime/rime-terra-pinyin (which is an IME setup for chinese input) */
```
BrettDong, line 13: Probably need to figure out if there is legal concern or consequence from directly embedding this LGPLv3 licensed data into the source code of this CC BY-SA 3.0 licensed project without a library boundary.

Qrox: Not sure why they chose LGPL for two yaml files which are not library source files. Perhaps we could turn these files into a library under the LPGL license and link them here but that sounds kinda ridiculous.

Also it seems that repo does not include some pronunciations of simplified characters (https://github.com/rime/rime-terra-pinyin/pull/49#issuecomment-1418266742) so we probably should use a different source anyway.

Perhaps we could use Unihan_Readings.txt from https://www.unicode.org/Public/UCD/latest/ucd/Unihan.zip which seems to have a compatible license (https://www.unicode.org/license.txt) but this files seems to have inconsistent format and needs normalization before we can use it.

src/pinyin.cpp
Water-Moon: Yeah, it should be not needed, thank you for the advice

src/pinyin.cpp
Water-Moon: To be honest I have no idea how I could make this work. Even if I generate data through other means, or by reading a dictionary and manually typing in the data, the result would mostly be the same, because the pronunciation of a character does not change no matter how the work is done...

I'll see if I can come up with alternative methods for this

Water-Moon: > Perhaps we could use the kHanyuPinyin field from Unihan_Readings.txt in https://www.unicode.org/Public/UCD/latest/ucd/Unihan.zip which seems to have a compatible license (https://www.unicode.org/license.txt) but this files seems to have inconsistent format and needs normalization before we can use it.

Yeah, I found that as well, but I might just format it like before (pinyin:characters map), I don't know if I should put that into a separate file in the third_party folder and attach the unicode's license to it

Water-Moon: > To be honest I have no idea how I could make this work. Even if I generate data through other means, or by reading a dictionary and manually typing in the data, the result would mostly be the same, because the pronunciation of a character does not change no matter how the work is done...

Please disregard that, I think if we attach the license for Unihan or Unicode it should be fine, it seems like most similar work are derived from the Unihan database that you just mentioned

Water-Moon:
I'm not sure if doing it this way will be OK. I don't know much about licenses so please help check this, thanks.

src/third-party/pinyin/pinyin_data.hpp
```diff
@@ -0,0 +1,454 @@
...
+    {U"mian", U"㒙㛯㝃㝰㡈㤁㨺㮌㰃㴐㻰䀎䃇䏃䛉䤄䩄䫵䯶䰓丏俛偭免冕冥勉勔喕娩婂媔嬵宀愐杣棉檰櫋汅沔泯渑湎湣澠牑眄眠瞑矈矊矏糆絻綿緜緡緬绵缅腼臱芇莬葂蝒蠠面靣靦鮸麪麫麵麺黽黾𡒳𡕢𡧍𡧒𡯫𢃮𢣔𣅍𣡠𣧾𥄝𥊿𥌂𥤵𥻩𥾝𦬛𦽃𧭇𧸨𨉥𨟺𨡞𨲛𨲾𩈹𩋠𩬞𩮉𩾃𪁼𪏊"},
+    {U"miao", U"㑤㠺㦝䁧䅺䏚䖢仯吵喵妙媌嫹庙庿廟彯描杪淼渺猫玅眇瞄秒竗篎紗緢緲繆缈苗藐蜱訬邈鱙鶓鹋𠋝𡙎𡡺𢆷𢚋𢤧𢷕𤾛𥭝𦳥𩳸𪃐𪃦"},
+    {U"mie", U"㒝㩢䁾䈼䌩䘊䩏乜吀咩咪哶孭幭懱搣櫗滅瀎灭烕眜篾羋蔑薎蠛衊覕鑖鱴鴓𠺗𡖺𡞙𡟬𢦼𢧞𢨖𤊾𤏿𥄲𥉓𥋚𥌨𥣫𥵒𥸴𥽘𥾝𦇪𧀅𧂝𨣱𩔠𩱷𪇴𪌺𪒍"},
+    {U"min", U"㞴㞶㟩㟭㢯㥸㨉㬆㳷䁕䂥䃉䋋䝧䟨䡑䡻䪸䲄僶冺刡勄厸呡垊姄岷崏忞怋悯惽愍慜憫抿捪敃敏敯旻旼暋民汶泯渂湏湣潣玟珉琘琝瑉痻皿盷盿眠砇碈笢笽簢緍緡繩缗罠苠蠠鈱錉鍲閔閩闵闽鰵鳘鴖黽𠊟𢼖𢽹𣱈𣱉𣷠𣹒𤇜𤛎𤸅𤺖𤿕𥜐𦈏𦌡𦫮𦳜𧁋𧌙𧲃𧵣𨏵𩔉𩭷𪂆𪄴𪉎"},
+    {U"ming", U"㝠㟰㫥䄙䆩䊅䏃䒌䫤䳟佲冥凕名命姳嫇慏掵明暝朙椧榠洺溟猽皿眀眳瞑茗萌蓂螟覭詺鄍酩銘铭鳴鸣𠋶𡥸𥌏𥥊𥹆𥿨𦡉𦫭𧟠𧱴𩣶𪂤𪗸"},
+    {U"miu", U"嘐繆謬谬𨱯"},
+    {U"mo", U"㱄㱳㵹㶬㷬㷵㹮䁼䁿䃺䊳䏞䒬䘃䜆䩋䬴䭩䮬䯢䱅䳮䴲万么伯佰冒劘劰勿唜嗼嘿嚤嚩嚰圽塻墨妺嫫嫼寞尛帓帕帞庅怽懡戂抹摩摸摹撫擵攠昧昩暯末枺模橅歾歿殁没沫湐漠瀎無爅狢獏瘼百皌眜眽眿瞐瞙砞磨礳秣粖糢絈絔縸纆耱膜艒茉莈莫蓦藐藦蘑蛨蟆蟔袜袹謨謩譕谟貃貈貉貊貌貘鄚銆鏌镆陌靺鞨饃饝馍驀髍魔魩魹麽麿默黙𠄮𠆮𠇱𠡞𠢓𠬛𠻚𡈗𡊉𡠜𡡉𡢜𡻟𡾉𢄏𢊗𢐖𢗿𢣗𢳀𣋟𣧒𣧣𣶊𣻕𤋂𤣻𤹴𤿖𥂓𥄕𥕓𥙎𥞪𥬎𥱹𥽘𥿉𦅔𦔭𦟟𦥦𦫕𦮅𧀅𧕤𧕥𧠓𧥟𧰱𧻙𧼟𧿴𨆽𨟖𨰞𨱱𩃁𩄻𩌧𩎟𩐻𩑦𩑷𩞁𩟠𩢖𩢷𩥔𩪮𩿣𪍇𪍤𪎠𪏟𪒂𪒇𬙊"},
+    {U"mou", U"㭌䋷䍒䍙䏬䗋䜼䥐䱕件侔劺厶呣哞堥婺恈敄某桙毋洠牟畝眸瞴繆缪蛑蟱袤謀谋踎鉾鍪鞪鴾麰𠀱𠥨𢃱𢜯𣫬𥆆𥿵𦊋𦊎𦋡𦭷𦳑𦺒𧎄𧎻𨴍𩢫𩶢"},
+    {U"mu", U"㙁㜈㟂㣎㧅㾇䀲䊾䑵䥈䱯亩仫凩募嘿坶墓墲姆姥娒婺峔幕幙慔慕拇暮木朰朷楘模樢母毣毪氁沐炑牟牡牧牳狇獏畆畒畝畞畮目睦砪穆縸繆胟艒苜茻莫莯萺蚞踇鉧鉬钼雮霂鞪鶩𠺖𡵬𢘃𢜯𢟨𣈊𣳠𤚅𤝂𤝕𤵝𥄈𥣸𥰻𦃤𦱒𧚀𧩒𧬏𧰷𧿹𨈶𨍎𨎸𨡭𨢢𩡨𩬍𩵦𩶖𩶩𪎫𬭁𬰃"},
+    {U"m̀", U"呣"},
```
Qrox, line 239:
```suggestion
```
Already in line 219.

src/third-party/pinyin/pinyin_data.hpp
```diff
@@ -0,0 +1,454 @@
...
+    {U"zuan", U"㸇䂎䌣䡽䤸䰖劗揝撮攢攥欑篹籫繤纂纉纘缵賺躜躦鑚鑽钻𡉺𣀶𣪁𦆈𦙉𨇃𨉖𨰭𩎑𩛻𩯳"},
+    {U"zui", U"㝡㠑㭰㰎䓱䘒䘹䮔䯿厜嗺嘴噿嫢嶉嶊嶵摧撮晬最朘枠栬槜槯樶檇檌欈濢璻睟祽稡穝絊纗罪羧蕝蕞蟕觜辠酔酨酻醉鋷錊雋𠲋𠾋𠿘𡙭𡡔𡽁𡽛𢈡𢊛𣖱𣩑𥍋𥳣𥳵𦆈𦈬𦏳𦙈𦸺𧎹𧻝𨢅𨻵𨿇𩚻𩣷𩲨𪋌𪓌"},
+    {U"zun", U"䔿僎僔噂墫壿奠尊嶟拵捘捽撙栫樽瀳繜罇袸譐跧踆蹲遵銌鐏鱒鳟鶎鷷𠟃𠱜𤮐𥊭𥞘𥢎𦢐𦨆𦪚𨱔𩯄"},
+    {U"zuo", U"㑅㘀㘴㘸㛗㝾㤰㩇㭮㵶㸲䋏䎰䔘䝫䞢䞰䟄䟭䟶䦈乍佐作侳做咗唑嘬坐岝岞左座怍挫捽撮昨柞柮椊琢砟祚秨稓笮筰糳繓胙苲莋葃葄蓙袏諎迮酢醋鈼鑿阼飵𠂇𠱯𠹠𡪳𡯨𡹥𢀡𢂃𢤎𣠹𣩈𣹧𤿀𥅁𥙀𥥏𥹁𥽿𦁎𦈛𦥬𦦹𧃘𧨀𧮙𧲭𨀨𨐳𨝨𨞒𩛠𪎇𪎲"},
+    {U",", U"，"},
+    {U".", U"。"},
+    {U"?", U"？"},
+    {U"!", U"！"},
+    {U"-", U"—"},
+    {U"\"", U"“”"}
```
Qrox, line 451:
```suggestion
    {U"\"", U"“”"},
    {U"(", U"（"},
    {U")", U"）"}
```
Full-width brackets are often used in Chinese item names.

src/third-party/pinyin/pinyin_data.hpp
```diff
@@ -0,0 +1,454 @@
...
+    {U"ning", U"㝕㣷㲰㿦䆨䔭䗿䭢佞侫倿儜冰凝咛嚀嬣宁寍寕寗寜寧年拧擰攘柠橣檸泞泥澝濘狞獰甯疑矃聍聹苧薴鑏鬡鬤鸋𡫃𡷪𣍆𤕦𤹧𤻝𥣗𥧤𥳥𦡲𦡼𧃱𧍖𧑗𧕝𧭈𩕳𫛢"},
+    {U"niu", U"㖻㺲㽱䂇䋴䏔䒜妞忸怓扭拗杻汼沑炄牛牜狃紐纽莥蚴鈕钮靵𢔟𣅚𣧊𣲶𥀝𥍳𥝦𧘥𨋀𨙺𨳞𨷁𩈇𩙷𩚖𩜷𩲍𩵠𪏲"},
+    {U"nong", U"㶶㺜䁸䢉䵜侬儂农咔哝噥弄憹挊挵檂欁浓濃燶癑禯秾穠繷脓膿莀蕽襛農辳醲齈𠌚𠘊𥂒𨑊𨠤𨲳𩅽𩇔𩟊𪆯𪒬𬪩"},
+    {U"nou", U"㜌㝹㳶䅶䘫䨲䰭啂嬬搙擩槈檽獳羺耨譨譳鎒鐞𠲴𡝦𡨻𡭾𢉕𢉚𣻖𤟦𥀫𧂦𧃨𧅘𩆟𩒔𪋺"},
+    {U"nu", U"㚢㣽仅伮傉努呶奴孥帑弩怒挐搙擩砮笯肭胬褥詉駑驽𠴂𢪦𢫓𢫭𥅄𥛑𥤨𥱂𧉭𧗈𧪅𧿔𪺹"},
+    {U"nuan", U"㬉䙇奻暖渜湪濡煖煗臑餪𪋐"},
+    {U"nun", U"媆黁"},
+    {U"nuo", U"㐡㑚㔮㖠㛂㡅㰙䇔䎠䚥傩儺呐哪喏堧娜媠愞懦懧挪掉掿搙搦搻梛榒橠毭濡稬穤糑糥糯耎袲袳諾诺蹃逽那郍鍩锘難需𠸱𠹈𡖔𡖫𡬥𡿊𢜪𢰜𢾲𣃽𣆚𤘟𥑽𥻾𦀨𦂍𦓢𦡃𦩜𧣚𧣺𨁌𨎭𩈺𩴓𩷁"},
+    {U"nv", U"㵖䏔䖡䘐䚼䶊女恧朒沑狃籹絮聏胬衂衄釹钕𥄋𥍞𦓕𦓖𧖷𩈢"},
+    {U"nve", U"䖈䖋䨋婩疟瘧硸虐𤷡𨵫"},
```
Qrox, line 267: I think "nue" is more common when typing with an IME? There are also other possible alternative spellings such as "lve" - "lue", "jv" - "ju" so we should probably create an alternative spelling map and use it when building the character-pinyin map in pinyin.cpp.

src/pinyin.cpp
```diff
@@ -1,472 +1,20 @@
...
-                    if( std::find( indexed_pinyin_map.at( current_char ).begin(),
-                                   indexed_pinyin_map.at( current_char ).end(),
-                                   entry.first ) == indexed_pinyin_map.at( current_char ).end() ) {
-                        indexed_pinyin_map.at( current_char ).push_back( entry.first );
-                    }
-                }
-            }
-        }
-    }
-
```
Qrox: Why is this removed? I think the map still needs to be built.

Water-Moon: oh sorry, something went very wrong when I upload my commit and I did not notice it

Water-Moon: I initially plan to just store the document as-is (character:pinyin map instead) but that turned out to be too big, but then I dont think I fully reverted my change

src/third-party/pinyin/pinyin_data.hpp
Water-Moon: This shouldn't even be here, I think I missed something when removing the accents

src/third-party/pinyin/pinyin_data.hpp
Water-Moon: Mapped ü to both u and v, hopefully that will make more sense to end user

src/third-party/pinyin/pinyin_data.hpp
```diff
@@ -8,6 +8,7 @@
         extracted kMandarin and kHanyuPinyin field
         converted to a map structure
         removed accents on letter
+        mapped "ü" to both "u" and "v"
```
Qrox, line 11: I'm afraid a simple mapping from `ü` to `u` and `v` won't work. There are `ü` sounds that are spelled as `u` in standard pinyin such as `ju` which can also be spelled as `jv` and some pinyin with the `ü` sound cannot be spelled with `u` (for example `lü` cannot be spelled as `lu`)

src/third-party/pinyin/pinyin_data.hpp
```diff
@@ -18,6 +19,7 @@
 #include <string>
 
 static const std::map<std::u32string, std::u32string> pinyin_data = {
+    {U"", U"㓘㟬㣇㤺㬈㱎㼔㼛䄯䇫䈧䣺䦕䩶冞劷叽埒宂宜岍嵮廾形惇憤戛揅斲曼枅椔檨殦涅潛瀛灊灰災牐犀珊甮竮籑緇罺聠舁茣荆蚈蚩蛌蛢裗豕賾跚鄑鄖鄛酁鐕韌頋飢鬒鼖𠂇𠒢𠓌𠢢𠣉𡙛𡥋𡬘𡶭𡽜𢍫𢔴𢛔𢻂𣝜𣽨𤕚𤜵𤳘𥌱𥖗𥟿𥩒𥪧𥳐𦈨𦣹𦥱𦫎𦾱𧈚𧗆𧙧𧢛𧲨𧷎𨊻𨋐𨗥𩄮𩲡𪆷"},
```
Qrox, line 22: Something probably went wrong here.

src/third-party/pinyin/pinyin_data.hpp
```diff
@@ -233,10 +236,10 @@ static const std::map<std::u32string, std::u32string> pinyin_data = {
     {U"min", U"㞴㞶㟩㟭㢯㥸㨉㬆㳷䁕䂥䃉䋋䝧䟨䡑䡻䪸䲄僶冺刡勄厸呡垊姄岷崏忞怋悯惽愍慜憫抿捪敃敏敯旻旼暋民汶泯渂湏湣潣玟珉琘琝瑉痻皿盷盿眠砇碈笢笽簢緍緡繩缗罠苠蠠鈱錉鍲閔閩闵闽鰵鳘鴖黽𠊟𢼖𢽹𣱈𣱉𣷠𣹒𤇜𤛎𤸅𤺖𤿕𥜐𦈏𦌡𦫮𦳜𧁋𧌙𧲃𧵣𨏵𩔉𩭷𪂆𪄴𪉎"},
     {U"ming", U"㝠㟰㫥䄙䆩䊅䏃䒌䫤䳟佲冥凕名命姳嫇慏掵明暝朙椧榠洺溟猽皿眀眳瞑茗萌蓂螟覭詺鄍酩銘铭鳴鸣𠋶𡥸𥌏𥥊𥹆𥿨𦡉𦫭𧟠𧱴𩣶𪂤𪗸"},
     {U"miu", U"嘐繆謬谬𨱯"},
+    {U"m", U"呣"},
```
Qrox, line 239: Still needs to be merged with line 222

src/third-party/pinyin/pinyin_data.hpp
Water-Moon: Oh, I just forgot to remove this line, it is caused by those data entries that have two index number but one pronounciation, such as this one: 
image
My conversion code was just a makeshift one so it happens. Removed now

src/third-party/pinyin/pinyin_data.hpp
Water-Moon: Thank you again for catching this, I did not expected it to happen

src/third-party/pinyin/pinyin_data.hpp
Water-Moon: Oh, didn't thought about this.. There are two cases with this problem, lu and nu, it should be fixed now. Thank you for pointing this out

PR #63671, opened 2023-02-19
Adjust quantities of some food generated at kitchens

Kromgart:
#### Summary
Balance "Randomize fixed amounts of certain foods found on kitchens"

#### Purpose of change

A lot of food on kitchens has random amount, but some are always found in the same fixed amounts: 33x tea, 32x butter, 8x cheeses, 100x salt/spices, full bottles of cooking oil and vinegar etc. This PR makes the amounts more varied.

#### Describe the solution

The following SUS groups were updated:

SUS_pantry
SUS_coffee_cupboard
SUS_breakfast_cupboard
SUS_spice_collection
SUS_fridge_stocked
SUS_fridge_bachelor
SUS_fridge_healthy_vegan
SUS_fridge_partial_empty
SUS_fridge_breakroom

#### Testing

Used debug menu 'Test item group' and observed the results.
Then teleported into a nearby town and went through ~15 houses checking the amounts generated.
There are still occasionally some cases of 100 seasoned salt lying on a table in a dining room, but those are most probably not using SUS itemgroups. Restaurants' kitchens also seem to be not affected by these changes etc.

But I think that the majority of cases of perfect cloned stacks of cheese/salt/butter/pepper/tea/cooking oil found in regular kitchens was handled.


Maleclypse:
Sealed containers should always be found full

kevingranade:
This is held up by the sealed container issue.

Issue #62247, opened 2022-11-16
Typing in the mobile version of the game is inconsistent


gettingusedto:
### Describe the bug

Typing in the mobile port with the on-screen keyboard is quite inconsistent. Some works as expected, like searching for a recipe in the crafting menu, meanwhile typing a letter in the diary screen closes my on-screen keyboard, making me press my "back" button to deploy the on-screen keyboard back and type another letter.

### Steps to reproduce

1. play the mobile port
2. get into a world
3. open up the diary
4. edit text
5. agony

### Expected behavior

On-screen keyboard not retracting every time i press a letter.

### Versions and configuration

- OS: Android
    - OS Version: Manufacturer: samsung; Model: SM-A710F; Release: 7.0; Incremental: A710FXXU2CRC2; 
- Game Version: 65e5b97 [32-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Magiclysm [magiclysm],
    Misc Sprites [misc_sprite],
    Misc Sprites: Magiclysm Edition [misc_sprite_magiclysm],
    Crow's Hairstles 2: Ears and Tails [hairmod2_earsAndTails],
    Portal Gatcha Mod [GATCHA_Portal],
    Stats Through Skills [StatsThroughSkills]
]

### Additional context

I assume using physical keyboard didn't have this issue since the issue is that the on-screen keyboard retracting back when a letter pressed in certain places.

WreckingAbandon:
I think there's an option to turn that off, the "auto-manage cirtual keyboard" in the Android tab in the options.

gettingusedto:
Ah, it works as expected now, though it's weird that turning on "auto-manage virtual keyboard" makes it acts weirdly in certain places.

Qrox:
I think this is probably what is happening: the following code tries to close virtual keyboard automatically in non-text-input UIs:

https://github.com/CleverRaven/Cataclysm-DDA/blob/d1f765d85f6cf7df6d0fe07fa676c3a639b3f70e/src/sdltiles.cpp#L2727-L2734

However, it uses hardcoded ids to tell whether a context is text input or not:

[...]

PR #63672, opened 2023-02-19
Display IME candidate list and composition text correctly on Windows

Qrox:
#### Summary
Bugfixes "Display IME candidate list and composition text correctly on Windows"

#### Purpose of change
SDL 2.0.20 has added a new hint `SDL_HINT_IME_SHOW_UI` that lets the IME UI show correctly on Windows. (Technically it was 2.0.18, but it seems it was not correctly implemented in that version)

SDL 2.0.22 also added a `SDL_HINT_IME_SUPPORT_EXTENDED_TEXT` hint to provide the full composition text when it is too long to be contained within the old text editing event.

Also should fix android soft keyboard in string editor (diary) window closing on keypress (#62247).

#### Describe the solution
Set the hints. Use SDL 2.26.2 in the MXE builds to be consistent with the MSVC builds. Update documentation.

#### Testing
Tested the game built by the release action [here](https://github.com/Qrox/Cataclysm-DDA/releases/tag/cdda-experimental-2023-02-21-1511). The MSVC and MinGW versions now display the candidate list in the lower right corner of the game window and also display long composition text correctly.

Credits to @BrettDong for building the updated MXE.

I can't get the android build to work in my repo so I'm going to test it after this is merged.

#### Additional context
I suppose this could be shipped with 0.G?

### DIFF
--- a/.github/workflows/matrix.yml
+++ b/.github/workflows/matrix.yml
@@ -169,9 +169,9 @@ jobs:
             sound: 1
             localize: 1
             libbacktrace: 1
-            title: GCC 11, Ubuntu cross-compile to MinGW-Win64, Tiles, Sound
+            title: GCC 12, Ubuntu cross-compile to MinGW-Win64, Tiles, Sound
             ldflags: -static-libgcc -static-libstdc++
-            mxe_target: x86_64-w64-mingw32.static.gcc11
+            mxe_target: x86_64-w64-mingw32.static.gcc12
             wine: wine
             # ~285MB in a clean build
             # ~36MB compressed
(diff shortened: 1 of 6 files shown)


BrettDong:
The latest SDL2 version in MXE repository is 2.0.20. I just made a build and uploaded to https://github.com/BrettDong/MXE-GCC/releases/tag/mxe-sdl-2-0-20.

Updating to a more recent SDL2 version is possible by patching MXE Makefile in [the build process](https://github.com/BrettDong/MXE-GCC/blob/master/.github/workflows/build-mxe.yml), but I don't have time to do so.

PR #63673, opened 2023-02-19
Fix npc needs pathfinding taking unreasonable amount of time

### LINKED ISSUE #54683: Game froze for several minutes trying to spawn an npc

### LINKED ISSUE #58533: Long hang while reading book
### Describe the bug

While reading a book as the sun was setting, the game hung for a very long time, and showed as "not responsive", before eventually showing me the "stop reading?" prompt.

### Steps to reproduce

I'm not sure how to reproduce this precisely, but I've attached a dump file created with Task Manager, which hopefully will show where the problem is.

### Expected behavior

The game shouldn't take 60+ seconds to do anything!

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19043.1706 (21H1)
- Game Version: 781a1fc [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    sees-player, run, crouch, bleed icons, retrodays [sees_player_retro]
]

### Additional context

Here's a link to the memory dump taken while the game was hung: https://drive.google.com/file/d/1papImzrLN-a2oNqxeZV4vCsVNkR0mTud/view?usp=sharing

### LINKED ISSUE #59912: Freezing when crafting with very low NPC spawn delay
### Describe the bug

Upon crafting I experience a freeze when NPC spawn delay is low enough

requirements for freeze:
NPC spawn delay need to be low (less than 0.1)

[Xedra-max-npc.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/9264246/Xedra-max-npc.zip)
### Steps to reproduce

1. Use the save file sent above
2. Start a long craft (enough resources are around to start crafting a short bow)
3. The game will freeze, usually around 4% completion of the shortbow for me

### Expected behavior

The game does not freeze, or the game prevents me from setting a NPC spawn delay low enough to crash the game

### Screenshots

_No response_

OS: Windows 10
Game Version: 8f8b8d9 (although I have experienced this bug on all the version of the experimental I have tried in the last 2 weeks)
Tiles: Yes
Mods: cdda, NPC_NO_NEEDs, Xedra_Evolved

### Additional context

My kneejerk reaction is the solution should be to create a minimum NPC spawn delay, but I think that might just mask the actual issue

### LINKED ISSUE #60371: Hang while sleeping
### Describe the bug

This save consistently hangs if you sleep immediately after loading.

[Bear Creek.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/9391499/Bear.Creek.zip)

### Steps to reproduce

1. Load save
2. Immediately sleep

The game will hang after a few seconds, becoming nonresponsive.

### Expected behavior

The game shouldn't hang.

### Screenshots

_No response_

Experimental - 630e438

### Additional context

_No response_

### LINKED ISSUE #63202: Game hangs with message "hang on a bit..." constantly
### Describe the bug

In my last two created words i get this weird behavior that during long activities (reading, sleeping, hacksawing etc.) almost always there is this message on top of the screen "hang on a bit..." which takes like ~5 real world minutes to go away. Game is literally unplayable like this. I've tried updating to the latest experimental 2023-01-19-0645 and also reinstalling the game completly but the problem is still there. I'm attaching my save file for the current world (still small as just started) and my debug.log (which doesn't seem to contain anything regarding this behavior).

[Sistersville.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10464427/Sistersville.zip)
[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/10464435/debug.log)

### Steps to reproduce

1. Load the save
2. Try long activities (hacksaw metal bars, sleeping etc)

### Expected behavior

Let me play without wasting 5 minutes of my time every couple minutes.

### Screenshots

_No response_

- Win 10
- Tiles
- Version: 2023-01-19-0645
- Lang: German
- Mods: "dda", "no_npc_food", "no_fungal_growth", "package_bionic_professions", "blazeindustries", "speedydex", "StatsThroughSkills", "personal_portal_storms", "stats_through_kills"

### Additional context

_No response_

irwiss:
#### Summary
None

#### Purpose of change

Fix hangs caused by npc pathfinding

Fixes #54683
Fixes #59912
Fixes #63202

There's probably a ton of "mystery hang" issues related to this because it's caused by random NPC spawns which will most often occur during activities and player sleep where turns advance quickly, I propose closing the likely earlier issues, especially considering most of them don't have much actionable information; for the player the game window just froze before spinner was added in an earlier PR 62985:
Fixes #58533
Fixes #60371

How npc needs pathfinding works;
`npc::set_omt_destination` can set random goal matching the npc's needs, it then tries to find nearest overmap tiles that can fulfill the need. To do so it calls `overmap_buffer.find_closest`, this function may randomly return tiles that are on zlevels other than z=0. After that `overmap_buffer.get_travel_path` is called, and it tries to pathfind exact route to the selected OM tile.

The problem is - the pathfinding scorer function is only able to travel up/down zlevels via ramp overmap tiles (currently those are bridge head tiles), so when the NPC is on z=0 (random npcs only spawn on z=0) and the target OM tile is in the basement or upstairs it cannot be reached, causing `pf::find_overmap_path` to run until it exhausts the search space (or it reaches max_search_count=100000) which can take unreasonable amount of time

#### Describe the solution

This PR limits `overmap_buffer.find_closest` to search on just z=0 when pathfinding to fulfill npc needs, this will almost always produce a reachable tile.

The alternative is making sure the overmap tiles are VERY carefully set to correct SOURCE_* flags, but for example house_farm_roof inherits flags from generic_rural_building causing a roof to be a valid target for npc needs pathfinding, and enforcing this seems close to impossible to be automated and improbable to be 100% correct if done by reviewers.

#### Testing

Apply the patch from #63067 - this will allow debug menu to spawn random NPCs and spawn 20 random NPCs.
Alternatively; set npc spawn rate to 0.01, wait 20 hours.

You will most likely see a spinner multiple times before applying this patch, with this patch applied pathfinding should be near instant.

#### Additional context

Before (notice how the pathfinding that takes 60+ seconds are trying to pathfind to a different zlevel, since they're trying to look for a non existent route via a ramp up/down):
image

After:
image

### DIFF
--- a/src/npcmove.cpp
+++ b/src/npcmove.cpp
@@ -4111,6 +4111,11 @@ void npc::set_omt_destination()
             // those having equal minimal distance
             find_params.search_range = 75;
             find_params.existing_only = false;
+            // force finding goal on the same zlevel (for random spawned NPCs that's z=0), otherwise
+            // we may target unreachable overmap tiles (no ramp up/down) which makes
+            // overmap_buffer.get_travel_path waste a lot of time
+            find_params.min_z = surface_omt_loc.z();
+            find_params.max_z = surface_omt_loc.z();
             goal = overmap_buffer.find_closest( surface_omt_loc, find_params );
             npc_need_goal_cache &cache = goal_cache[fulfill];
             cache.goal = goal;
--- a/src/overmapbuffer.cpp
+++ b/src/overmapbuffer.cpp
@@ -1196,20 +1196,20 @@ tripoint_abs_omt overmapbuffer::find_closest( const tripoint_abs_omt &origin,
     const int max_dist = params.search_range ? params.search_range : OMAPX * 5;
 
     std::vector<tripoint_abs_omt> result;
-    cata::optional<int> found_dist;
+    int found_dist = std::numeric_limits<int>::max();
 
     for( const point_abs_omt &loc_xy : closest_points_first( origin.xy(), min_dist, max_dist ) ) {
         const int dist_xy = square_dist( origin.xy(), loc_xy );
 
-        if( found_dist && *found_dist < dist_xy ) {
+        if( found_dist < dist_xy ) {
             break;
         }
 
-        for( int z = -OVERMAP_DEPTH; z <= OVERMAP_HEIGHT; z++ ) {
+        for( int z = params.min_z; z <= params.max_z; z++ ) {
             const tripoint_abs_omt loc( loc_xy, z );
             const int dist = square_dist( origin, loc );
 
-            if( found_dist && *found_dist < dist ) {
+            if( found_dist < dist ) {
                 continue;
             }
 
--- a/src/overmapbuffer.h
+++ b/src/overmapbuffer.h
@@ -137,6 +137,8 @@ struct omt_find_params {
     bool must_see = false;
     bool cant_see = false;
     bool existing_only = false;
+    int min_z = -OVERMAP_DEPTH;
+    int max_z = OVERMAP_HEIGHT;
     cata::optional<overmap_special_id> om_special = cata::nullopt;
 };
 


andrei8l:
Nice job, though I still have to wonder why this is slow only on msvc. gcc/clang builds also need to exhaust the search space, but the process finishes in milliseconds rather than seconds/minutes.
Screenshot from 2023-02-19 19-32-25
on a debug build, no less!

irwiss:
mqrause posted a profiler screenshot in https://github.com/CleverRaven/Cataclysm-DDA/issues/63202#issuecomment-1433893771, looks like the hottest parts are either the unordered_map (or node_address_hasher, but it looks like it calls into cata::hash64->hash64_detail::maybe_mix_bits which supposedly just returns value as is on 64 bit systems if enable_if works right)

PR #41269, opened 2020-06-13
Add Desert Island start to wilderness

cirops:
#### Summary
SUMMARY: Content "Add Desert Island start to wilderness"
#### Purpose of change
Given that islands are now a thing in the game, it makes sense to allow the player to start on a deserted island (no structures) on the wilderness scenario. There is already a lake cabin start in-game, but I'd like to start on a "completely" bare one (other than map extras).
#### Describe the solution
To create a new location, sloc_desert_island, which includes all current islands in data/json/mapgen/islands as allowed_locs, and then add sloc_desert_island to the Wilderness scenario.
#### Describe alternatives you've considered
Creating a mod for it with more stuff, but it makes sense to have this already existing mapgen feature and make it available as starting scenario.
#### Testing
Started a bunch of characters on the Wilderness scenario, choosing the starting location to be "Desert Island"
#### Additional context
I added the [WIP] because **starting_location.cpp** sometimes can't find a "perfect spot" when starting on a bare island,  and sends a debug message, and the comment on that file says it only happens on some exotic scenarios like deep science lab starts, so I'm not sure this is good enough as-is, or needs the starting-spot picker to be tweaked to go in (which I don't know how to do).

PS: This is my first PR, I'm sorry if I messed something and am totally ok if this gets shot down.

PR #63675, opened 2023-02-19
Fix: "Failed to Create Starting Location" Exception does not flush Buffers.

### LINKED ISSUE #62601: Make light amp and IR goggles fold up while not in active

Inactive light amp and IR goggles have an encumbrance of 40 while inactive and 20 while active. This makes wearing them inactive destroy accuracy and make them near unusable when they are not active. 
Most if not all Light amp goggles let you fold them up if you are not using them. This is not represented in cdda. 

### Solution you would like.

Possible solutions:
- the simplest solution would be to treat inactive goggles to be folded up and reduce/remove its eye encumbrance
- make goggles strapped, covering crown and forehead (like headlamp) with some head encumbrance (eg 10-20, headlamps have 0) and when active add the normal 20 encumbrance to the eyes
- make them properly foldable (like loosening scarf), though as I understand it activate and loosening would take the same event

Leaving as is and unequiping the goggles when I dont need them

### Additional context

_No response_

### LINKED ISSUE #63601: Crash when generating starting location after previously aborted failed attempt
### Describe the bug

https://github.com/CleverRaven/Cataclysm-DDA/blob/170527d6c15564fbfb806d272a69a16976e00c4e/src/mapgendata.cpp#L46-L58

It crashes trying to initialize `default_groundcover` in this constructor.

Callstack:
grafik

### Attach save file

n/a

### Steps to reproduce

1. Create a world with city size 0
2. Make a character in the wilderness scenario with desert island starting location and confirm
3. Failed to generate starting location
4. Select "no" when asked to try again
5. Load template from previous attempt (not sure if that's necessary, but that's what I did to reproduce)
6. Change starting location to one of the other options and confirm
7. Crash

### Expected behavior

No crash

### Screenshots

_No response_

Actually did this on the branch from #60885, but due to the nature of the crash I'm very confident it has nothing to do with that. That branch is built on 95521e0ea24b8e59b617d0d65105adb3dd89d48a.

### Additional context

_No response_

aahue1889:
#### Summary

Bugfixes "Flush map buffers after failing to create starting location"

#### Purpose of change

Fixes #63601

Steps to Repro:

> I can confirm and repro the issue: Reposting steps here for further clarity.
> 
> Create a world with city size 0 image
> 
> Make a character in the wilderness scenario with desert island starting location and confirm image
> 
> Failed to generate starting location image
> 
> Select "no" when asked to try again image
> 
> - Load template from previous attempt (not sure if that's necessary, but that's what I did to reproduce) **Change starting location** to one of the other options and confirm Crash
> 
 

#### Describe the solution

- I put the flush Mapbuffer and Overmap_buffer in the Else statement as well; not just the If statement.

image
---

A greater breakdown of why this was needed is here:

> ## Potential Fix:
> Hitting “NO” executes the else portion of code which DOES NOT flush the MAPBUFFER and Overmap_buffer (I added the highlighted yellow portion of code)
> image
> image
> If user later decides to attempt rebuilding the world after making changes to the Starting point; then the logic will occasionally use ‘unflushed’ values from the previous failed attempt to create the world:
> 
> Here is an Example of this behavior; a gigantic number was passed to boosted_percent_str which would later cause a failure in other functions
> 
> image
> 
> See a regular session in comparison that does not pass strange values
> 
> image
> 
> After I made the highlighted changes; I **was able to change the starting location without any crash**
> 

At first I was trying to see if I could fix the Exception that was causing users to potentially run into the crash to begin with; 

image

I was not able to find an appropriate solution to this; it would require me to wrap my head around how mapgen is working - in particular the logic around Desert Island as a start point. This seems to have been a known issue; https://github.com/CleverRaven/Cataclysm-DDA/pull/41269#issuecomment-643844040

#### Testing

Some context:

This code path is a unique scenario;
- In order for a user to run into this portion of code they would need to fail to acquire a valid starting location that they have chosen
- That being said I exhausted the test from the repro steps provided

I performed the following tests:

During World Creation I changed-up the Spacing across 4 worlds 

- Ran into the exception when I chose a Desert Island start
- Refused to retry 
- Attempted to recreate a start using a different point (Once for each environment)
- Confirmed Start Point Success

image

I repeated the 4 steps above across all 7 locations with Wilderness startpoint and was able to get them all to work, with the exception of 'Desert Island' which - if Set to City Size 0 will always fail at each attempt. But the root problem of this issue was the crash -> And this is no longer occurring.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -827,11 +827,12 @@ bool game::start_game()
             omtstart = start_loc.find_player_initial_location( u.world_origin.value_or( point_abs_om() ) );
         }
         if( omtstart == overmap::invalid_tripoint ) {
-            if( query_yn(
+
+            MAPBUFFER.clear();
+            overmap_buffer.clear();
+
+            if( !query_yn(
                     _( "Try again?\n\nIt may require several attempts until the game finds a valid starting location." ) ) ) {
-                MAPBUFFER.clear();
-                overmap_buffer.clear();
-            } else {
                 return false;
             }
         }
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -832,6 +832,8 @@ bool game::start_game()
                 MAPBUFFER.clear();
                 overmap_buffer.clear();
             } else {
+                MAPBUFFER.clear();
+                overmap_buffer.clear();
                 return false;
             }
         }


mqrause:
> #### Purpose of change
> 
> Fixes #62601 

You have a typo there, it's #63601 actually.

aahue1889:
Moved Summary up to align with the expression:

image

irwiss:
If both branches execute same code you can just pull the common parts out of the `if` instead of duplicating like so:
```c++
            MAPBUFFER.clear();
            overmap_buffer.clear();
            if( !query_yn(
                    _( "Try again?\n\nIt may require several attempts until the game finds a valid starting location." ) ) ) {
                return false;
            }
```
It changes order slightly but clearing those should be fast enough to not make a big difference (or store query_yn() result in a variable otherwise)
Also your PR is missing a title

PR #63677, opened 2023-02-19
Pain recovery jsonification

Dan1ss1mo:
#### Summary
Features "Effects of pain recovery traits now implemented with eocs"

#### Purpose of change
More jsonification

There are 2: first is to remove pain > 0 check like guatdiandll suggested, i mainly added for performance reason, but i have no clue if it actually better and how to test it. Another one is to leave current implementation if it more efficient and allow code to draw values from json files. Again I have now idea which of them would be better in terms of performance and what ever or not effect is actually big enough to care. I didn't noticed any slow downs during my testing.

#### Testing
Inflicted a bit of pain with debug menu on my charecter and checked the rate at which it was recovering


alp7292:
how fast pain recovery happens by looking at code (not knowledgeable about coding) recovering every minute seems too good

GuardianDll:
in code it was the same, but 10 times bigger - 10 minutes and 30 pain, for example

GuardianDll:
With #63629 fixes, shoud this pr be edited to use enchantment instead of EoC? 0 EoCs sound better than 3, when the result is same (but need to check how often the pain fade is applied naturally, to pick the right `PAIN_REMOVE` value)

Dan1ss1mo:
> With #63629 fixes, shoud this pr be edited to use enchantment instead of EoC? 0 EoCs sound better than 3, when the result is same (but need to check how often the pain fade is applied naturally, to pick the right `PAIN_REMOVE` value)

Yeah, that sounds nice, tho it is more 1 EoC vs 0, as 2 others will copletly inactive and would only start ticking once you get the trait

PR #63680, opened 2023-02-19
Adds Earthshaper spell to reveal world map

Standing-Storm:
#### Summary
Content "Adds a Tier 3 Earthshaper spell to reveal the world map" 

#### Purpose of change

To add a spell using the "map" function, which IIRC isn't used by any of the eight Magiclysm classes, and provide a unique capability to Earthshaper so that Technomancer doesn't become a lot of people's obvious choice. 

#### Describe the solution

Adding the spell, giving Earthshaper a unique capability.

Keeping the solution in my base game.

#### Testing

Been using this in my personal game for a while and it works.

### DIFF
--- a/data/mods/Magiclysm/Spells/earthshaper.json
+++ b/data/mods/Magiclysm/Spells/earthshaper.json
@@ -636,5 +636,24 @@
     "max_level": 10,
     "min_range": 1,
     "max_range": 1
+  },
+  {
+    "id": "earthshaper_reveal_world_map",
+    "type": "SPELL",
+    "name": "Reading the Earthbones",
+    "description": "You extend your senses into the ground, becoming aware of every hill and valley within a large area.",
+    "valid_targets": [ "none" ],
+    "flags": [ "NO_HANDS", "CONCENTRATE", "SILENT" ],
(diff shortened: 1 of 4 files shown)


GuardianDll:
The problem is it does not actually reveal the "earthbound", it reveal the exact Z level you are in. So if you stand in the roof, you will reveal there is a lot of `.` air around you, which pretty eeh
Also its not only `every hill and valley within a large area.`, its `every field, forest, house, road, shop, mall, fire station, dojo, basement, sewer, cathedral, subway station, lab etc etc etc`
Despite this good job, it think it fine to go into the mod

Standing-Storm:
If you prefer, I could try to set it up as an EoC that only works on Z level 0 or below and otherwise says "You have no connection with the living earth, and the spell fails" or something like that?

GuardianDll:
Oh, that's actually smart approach, yeah, would be nice.

Standing-Storm:
Alright, it's all in and works. Once spell levels can be passed through EoCs or EoCs accept variables for spell levels, I can make it a lot more elegant.

KorGgenT (approved):
looks great!

PR #63681, opened 2023-02-19
Reword existing trap message

Kamayana:
#### Summary
Interface "Reword existing trap message"

#### Purpose of change
The existing message for attempting to deploy a trap on a deployed trap displays "It contains a trap already" even when the trap is set as benign. Benign traps shouldn't be called "traps" in-game since they're used for deployable furniture and the like.

#### Describe the solution
Changed the message to say "a deployed object" if the trap is benign.
Changed the message to the more encompassing "It contains a deployed %s."

Use "It contains a deployed %s." for both traps and benigns. This has incorrect grammar for caltrops, however ("a deployed caltrops"), and there wasn't an simple fix for that.

#### Testing
Checked, it shows both messages for traps and benigns.

### DIFF
--- a/src/iuse_actor.cpp
+++ b/src/iuse_actor.cpp
@@ -3769,8 +3769,11 @@ bool place_trap_actor::is_allowed( Character &p, const tripoint &pos,
     const trap &existing_trap = here.tr_at( pos );
     if( !existing_trap.is_null() ) {
         if( existing_trap.can_see( pos, p ) ) {
-            p.add_msg_if_player( m_info, _( "You can't place a %s there.  It contains %s already." ),
-                                 name, existing_trap.is_benign() ? _( "a deployed object" ) : _( "a trap" ) );
+            p.add_msg_if_player( m_info,
+                                 existing_trap.is_benign()
+                                     ? _( "You can't place a %s there.  It contains a deployed object already." )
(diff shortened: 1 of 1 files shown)


src/iuse_actor.cpp
```diff
@@ -3769,8 +3769,8 @@ bool place_trap_actor::is_allowed( Character &p, const tripoint &pos,
     const trap &existing_trap = here.tr_at( pos );
     if( !existing_trap.is_null() ) {
         if( existing_trap.can_see( pos, p ) ) {
-            p.add_msg_if_player( m_info, _( "You can't place a %s there.  It contains a trap already." ),
-                                 name );
+            p.add_msg_if_player( m_info, _( "You can't place a %s there.  It contains %s already." ),
+                                 name, existing_trap.is_benign() ? _( "a deployed object" ) : _( "a trap" ) );
```
Qrox, line 3773:
```suggestion
            p.add_msg_if_player( m_info,
                                 existing_trap.is_benign()
                                     ? _( "You can't place a %s there.  It contains a deployed object already." )
                                     : _( "You can't place a %s there.  It contains a trap already." ),
                                 name );
```
Avoids conflicts with possible identical strings in other places and also gives translators more context.

Kamayana: Good idea, done.

PR #63687, opened 2023-02-20
Fix build without sound feature

CzBiX:
#### Summary
Build "Fix build without sound feature"

#### Purpose of change

Fix the build require `sdl2-mixer` when `make SOUND=0`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -92,6 +92,7 @@ jobs:
             mxe: none
             ext: zip
             content: application/zip
+            tiles: 1
             sound: 0
           - name: Windows Tiles Sounds x64 MSVC
             artifact: windows-tiles-sounds-x64-msvc
@@ -100,6 +101,7 @@ jobs:
             mxe: none
             ext: zip
             content: application/zip
+            tiles: 1
             sound: 1
           - name: Windows Tiles x32 MSVC
             artifact: windows-tiles-x32-msvc
@@ -108,6 +110,7 @@ jobs:
             mxe: none
             ext: zip
             content: application/zip
+            tiles: 1
             sound: 0
           - name: Windows Tiles Sounds x32 MSVC
             artifact: windows-tiles-sounds-x32-msvc
@@ -116,6 +119,7 @@ jobs:
             mxe: none
             ext: zip
             content: application/zip
+            tiles: 1
             sound: 1
           - name: Windows Tiles x64
             mxe: x86_64
@@ -123,6 +127,7 @@ jobs:
             os: ubuntu-latest
             ext: zip
             content: application/zip
+            tiles: 1
             sound: 0
           - name: Windows Tiles Sounds x64
             mxe: x86_64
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -281,14 +281,14 @@ jobs:
       - name: Build CDDA (linux)
         if: runner.os == 'Linux' && matrix.mxe == 'none' && matrix.android == 'none'
         run: |
-          make -j$((`nproc`+0)) TILES=${{ matrix.tiles }} SOUND=${{ matrix.tiles }} RELEASE=1 LOCALIZE=1 LANGUAGES=all BACKTRACE=0 PCH=0 bindist
+          make -j$((`nproc`+0)) TILES=${{ matrix.tiles }} SOUND=${{ matrix.sound }} RELEASE=1 LOCALIZE=1 LANGUAGES=all BACKTRACE=0 PCH=0 bindist
```
Qrox, line 284: I think the `sound` flag is supposed to enable shipping soundpacks with the game, not enabling the sound feature, see #57138.

.github/workflows/release.yml
```diff
@@ -281,14 +281,14 @@ jobs:
...
           mv cataclysmdda-0.F.tar.gz cdda-${{ matrix.artifact }}-${{ needs.release.outputs.timestamp }}.tar.gz
       - name: Build CDDA (windows mxe)
         if: matrix.mxe != 'none'
         env:
           PLATFORM: /opt/mxe/usr/bin/${{ matrix.mxe }}-w64-mingw32.static.gcc11-
         run: |
-          make -j$((`nproc`+0)) CROSS="${PLATFORM}" TILES=1 SOUND=1 RELEASE=1 LOCALIZE=1 LANGUAGES=all BACKTRACE=1 PCH=0 bindist
+          make -j$((`nproc`+0)) CROSS="${PLATFORM}" TILES=${{ matrix.tiles }} SOUND=${{ matrix.sound }} RELEASE=1 LOCALIZE=1 LANGUAGES=all BACKTRACE=1 PCH=0 bindist
```
Qrox, line 291: Windows builds always have tiles enabled. If you make this change please also add the tiles flag to the build parameters above (line 88-147).

.github/workflows/release.yml
CzBiX: I don't get it. Why we need to include the sound feature when there is no sound packs.

.github/workflows/release.yml
CzBiX: Fixed.

.github/workflows/release.yml
Qrox: Because the player can download a different soundpack themselves afterwards

CzBiX: ok, seems this PR is not useful.😅

Qrox: The changes to Makefile are still useful though in case someone wants to compile with sound off themselves.

CzBiX: I see. That will require two separate options for sound feature and sound pack, and should be in another PR.

PR #63688, opened 2023-02-20
[Magiclysm] some unarmed fixes

GuardianDll:
#### Summary
None
#### Purpose of change
Found there is some stuff in enchanted_unarmed, that no one touched for a long time, decided to fix it
#### Describe the solution
Added recipes for fire and pounding fist, fixed some their stuff
Obsoleted the shotcestus item and recipe
Added cestuces into +1 and +2 weapons loot groups, so Forge will trade them now
#### Additional context
Wonder how much stuff i can find in the game files, that is not used anywhere. Do we have some way to spot such examples?


SittingDucken:
Only thing I can think of atm is checking through history, finding things older than 2-3 years, and following references. That's a manual process, so it's not great. AFAIK everything that's jsonized has a key of "id": value, so there's potential to write a tool that parses a given id key:value and generates a reference count.

Cupidatis:
Did anyone find "greater bracer of lightning" and "bracer of greater defense" in game?

GuardianDll:
Very rarely they can spawn in the dragon lair, but that's it

SittingDucken:
See the flaming fist implemented, and it shows up at the Forge of Wonders shop, current item price is 0.

PR #63690, opened 2023-02-20
Tough feet trait jsonification

Dan1ss1mo:
#### Summary
Features "Tough feet trait effect now tied to flag rather then specific traits"

#### Purpose of change
More json for the json god, also i wanted to make some integrated armor to have this effect, becouse even tho it is considered skintight, it should still be tough enogh to keep youre soles safe

#### Describe the solution
Tough_feet flag that can be used on mutations and worn items

#### Testing
Barefooted penalties doen't apply with any of the changed traits or clothing with this flag


Maleclypse (approved):

bombasticSlacks:
Rebased so this would run again

PR #63691, opened 2023-02-20
Allow running EOCs as result of a recipe

Ramza13:
#### Summary
None

#### Purpose of change
Been requested on discord a bunch of times.
Json power

#### Describe the solution
Allow running EOCS on completion of a recipe.

#### Testing
Used this json
```
  {
    "type": "recipe",
    "name":"tester",
    "id":"test",
    "category": "CC_FOOD",
    "subcategory": "CSC_FOOD_MEAT",
    "skill_used": "cooking",
    "time": "1 m",
    "autolearn": true,
    "batch_time_factors": [ 67, 5 ],
    "qualities": [ { "id": "COOK", "level": 1 } ],
    "proficiencies": [ { "proficiency": "prof_food_prep" } ],
    "result_eocs": [ {"id": "TEST", "effect": { "u_message": "You feel Test" } } ]
  },
```

### DIFF
--- a/doc/JSON_INFO.md
+++ b/doc/JSON_INFO.md
@@ -1715,6 +1715,8 @@ Crafting recipes are defined as a JSON object with the following fields:
   "BLIND_EASY",
   "ANOTHERFLAG"
 ],
+"result_eocs": [ {"id": "TEST", "effect": { "u_message": "You feel Test" } } // List of inline effect_on_conditions or effect_on_condition ids that attempt to activate when this recipe is successfully finished.
+], 
 "construction_blueprint": "camp", // an optional string containing an update_mapgen_id.  Used by faction camps to upgrade their buildings
 "on_display": false,         // this is a hidden construction item, used by faction camps to calculate construction times but not available to the player
 "qualities": [               // Generic qualities of tools needed to craft
--- a/doc/JSON_INFO.md
+++ b/doc/JSON_INFO.md
@@ -1715,7 +1715,7 @@ Crafting recipes are defined as a JSON object with the following fields:
   "BLIND_EASY",
   "ANOTHERFLAG"
 ],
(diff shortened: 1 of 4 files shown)


mqrause:
Just wondering, what's the reason to make recipes that aren't practice and don't have a result? Is the intention to make items through the EOCs?

Ramza13:
> Just wondering, what's the reason to make recipes that aren't practice and don't have a result? Is the intention to make items through the EOCs?

Thats the main one yea, for things like artifacts.

mqrause:
Something to keep in mind is that as it is you can't learn recipes that produce no result. Might not be an issue for the intentions of this PR, but should probably get documented to avoid confusion if you wanna keep it like that.

Ramza13:
> Something to keep in mind is that as it is you can't learn recipes that produce no result. Might not be an issue for the intentions of this PR, but should probably get documented to avoid confusion if you wanna keep it like that.

I'm not sure I follow.  Practice recipes can be learned from books and there is a dialog command to learn recipes.  What scenario are you referring to?

mqrause:
I'm referring to this block: https://github.com/CleverRaven/Cataclysm-DDA/blob/47cf98ca587b8b1ad4a6767c8b26ee98084ffe72/src/crafting.cpp#L1266-L1286

You have a chance to learn the recipe when you finish the craft. Currently that only happens when you have an item resulting from the craft.

Ramza13:
> I'm referring to this block:
> 
> https://github.com/CleverRaven/Cataclysm-DDA/blob/47cf98ca587b8b1ad4a6767c8b26ee98084ffe72/src/crafting.cpp#L1266-L1286
> 
> You have a chance to learn the recipe when you finish the craft. Currently that only happens when you have an item resulting from the craft.

Thanks, updated it to fix that.

PR #63693, opened 2023-02-20
Add Magiclysm forage items to forage tables.

SittingDucken:
#### Summary
Features "Adds Magiclysm forage items to forage table, and allows rare items to be found via factionbase forage."

#### Purpose of change

Adding forage items to forage_<season> enables more interaction. Backrooms and Aftershock will now see forage items from Magiclysm. NPCs foraging via faction bases will collect Magiclysm forage items. Rare forage items are not added to table, as that can make them more common then intended. Anything found specifically in magic meadows only were note added.

#### Describe the solution

Feature adds common and uncommon forage items to the forage_<season> tables. This way when other mods or the faction base calls the loot table for forage_<season> the items will be available. Weighted percentages were determined by calculating their rough availability based on mapgen and compared to the Magiclysm forage rarity table (itemgroups.json, magic_plants_common). 

Could instead directly add the magic_plants_common to various locations instead of specifically extending the seasonal forage tables. However this would add out of season forage items to the table that normally couldn't be harvested. Could also only add to the faction base forage table, but that precludes them from showing up in mods that utilize the forage_<season> tables. 

Could also add the rare ingredients to the forage_season tables, however since they only appear in magic meadows, would dilute the rarity of the object substantially.

#### Testing
Implemented locally, loaded NPCs into a faction base, and had them forage. They were able to generate Magiclysm forage items at a low rate.

#### Additional context

Some items forage probabilities were changed based on comparing their intended rarity, and the feasibility of finding the item in the wild. Specifically, kampuch is considered uncommon, but has a high probability of being found in the wild. I've found this to occur in play too, not just in the data. 

I am concerned about the funnel cap drop amount as being too high. It is currently set to the actual harvest, and should be a good drop rate since the recipe usually calls for a fair amount of funnelcap_chunk.


SittingDucken:
Code Style Reviewer / style-code (pull_request_target) Failing after 52s ran once already and was accepted.

data/mods/Magiclysm/items/forage.json
```diff
@@ -0,0 +1,94 @@
...
+    "copy-from": "foraging_faction_camp_fall",
+    "extend": {
+      "entries": [
+        { "item": "golden_reed_stalk", "count": [ 1, 4 ], "prob": 1 },
+        { "item": "platonic_flower", "count": [ 1, 1 ], "prob": 4 },
+        { "item": "sojourners_flower", "count": [ 1, 5 ], "prob": 2 }
+      ]
+    }
+  }
+]
```
Maleclypse, line 94:
```suggestion
]

```

SittingDucken: Was reading the end line the secret sauce?

Maleclypse: Yes, you have to have an empty line at the end of the file and for whatever reason it must be manually added if it is missing.

Maleclypse (approved):

PR #63695, opened 2023-02-20
Add Environmental protection to "turtleneck (rolled)" and "turtleneck shirt (rolled)"

Alm999:
#### Summary
Balance "Add environmental protection to rolled "turtleneck" and "turtleneck shirt" -- the same amount as their unrolled variants"

#### Purpose of change

I find it strange that rolling the neck part makes whole upper torso unprotected against environmental attacks (somehow).

#### Describe the solution

I simply copy-pasted the settings from the base unrolled items.

None.

#### Testing

Tested the change in-game. At least now the "Protection" entry shows in the item description window.

#### Additional context

None.

### DIFF
--- a/data/json/items/tool_armor.json
+++ b/data/json/items/tool_armor.json
@@ -2730,6 +2730,7 @@
     "color": "white",
     "warmth": 20,
     "material_thickness": 0.5,
+    "environmental_protection": 1,
     "use_action": {
       "target": "turtleneck_shirt",
       "menu_text": "Unroll neck",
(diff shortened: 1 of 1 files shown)


Rivet-the-Zombie:
Sorry, misclicked.

bombasticSlacks:
I think the turtle-necks that are unrolled should have mouth coverage is the idea... So I would say either the rolling should be removed or they should be further distinguished.

Alm999:
> I think the turtle-necks that are unrolled should have mouth coverage is the idea... So I would say either the rolling should be removed or they should be further distinguished.

My understanding is that currently there is no way to individually set environmental protection for mouth and nothing else. A piece of clothing either has the protection, or don't. So, without reworking the system I found making "env" protection consistent between the modes is better than outright removing the protection.

Alm999:
Changing to dedicated branch.

PR #63699, opened 2023-02-20
Separate unused casings from used casings pool

### LINKED ISSUE #63000: Unused .22 casings spawning where they should realistically be used casings
### Describe the bug

The game scatters the map with random spent casings, probably representing gunfire during the chaos of the cataclysm. Take a walk outside and you'll see a few random casings scattered around. But you'll also find that the .22 casings you see are unused, despite representing the leftovers of past gunfire. I think this is an unintentional side-effect of #62562.

### Steps to reproduce

1. Take a scroll around some random fields, using 'V' to list the items in your field of vision.
2. You'll eventually find unused .22 casings, representing the now-forgotten art of firing a bullet without using its primer.

### Expected behavior

Used .22 casings should be spawning instead of unused .22 casings, because these casings are probably from past gunfire rather than Santa sprinkling bullet casings across the world as a gift to handloading enthusiasts.

### Screenshots

Took a single step outside from my starting evac shelter and immediately found an example.
Screenshot 2023-01-04 202517

- OS: Windows
- Game Version: a00a2ec [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

IcySlurpee:
#### Summary
Bugfixes "Separate unused casings from used casings pool"

#### Purpose of change
Sometimes in the fields and roads, the game generates a human corpse and some ammo casings to realistically portray events that happened during the cataclysm.  Unused .22 casings were being spawned in these groups. It doesn't make sense that unused rimfire casings would be spawning here; used casings should spawn instead.

Fixes #63000

#### Describe the solution
This separates out the unused .22 casings from the pool of used casings into its own pool.  The use case for a separate pool was to allow gunsmith stores (`s_gun_4`) to have unused casings, as a gunsmith would realistically have primed casings available for sale.  Prepper garages use the same item group for ammo casings as gunsmith stores, so I let unused casings spawn there as well. Prepper garages can also spawn gunsmith's tools at the same rate; a prepper would also realistically be (or at least practicing) handloading ammo.

I thought about adding the unused casings directly to the `ammo_casings_bulk` item group, but I couldn't get the probability in a spot where casings were an uncommon enough spawn.  Adding more entries to the bulk casings group would throw off the rates. Also, custom mods may use that item group for situations where unused casings shouldn't be spawned.

#### Testing
* Created a new world, debug spawned in `s_gun_4` and `house_garage_prepper` specials, verified unused casings spawned.
* Debug tested item group `field`, `forest`, `forest_trash` to ensure unused casings did not spawn.
* Debug tested item group `ammo_casings_gunsmith_bulk` to ensure unused casings were dropping at the same rates as a used .22 casing were before in `ammo_casings_bulk` (~2%).

#### Additional context
N/A

### DIFF
--- a/data/json/itemgroups/Weapons_Mods_Ammo/ammo.json
+++ b/data/json/itemgroups/Weapons_Mods_Ammo/ammo.json
@@ -713,7 +713,6 @@
     "entries": [
       { "item": "10mm_casing", "prob": 60 },
       { "item": "22_casing", "prob": 80 },
-      
       { "item": "223_casing", "prob": 100 },
       { "item": "270win_casing", "prob": 60 },
       { "item": "300_casing", "prob": 80 },
--- a/data/json/itemgroups/Weapons_Mods_Ammo/ammo.json
+++ b/data/json/itemgroups/Weapons_Mods_Ammo/ammo.json
@@ -713,6 +713,7 @@
     "entries": [
       { "item": "10mm_casing", "prob": 60 },
       { "item": "22_casing", "prob": 80 },
+      
       { "item": "223_casing", "prob": 100 },
       { "item": "270win_casing", "prob": 60 },
       { "item": "300_casing", "prob": 80 },
--- a/data/json/itemgroups/Weapons_Mods_Ammo/ammo.json
+++ b/data/json/itemgroups/Weapons_Mods_Ammo/ammo.json
@@ -813,6 +813,17 @@
     ]
   },
   {
+    "id": "ammo_casings_gunsmith_bulk",
+    "type": "item_group",
+    "subtype": "distribution",
+    "entries": [
+      { "group": "ammo_casings_gunsmith", "prob": 25, "charges-min": 50, "charges-max": 50 },
+      { "group": "ammo_casings_gunsmith", "prob": 10, "charges-min": 100, "charges-max": 100 },
+      { "group": "ammo_casings_gunsmith", "prob": 2, "charges-min": 250, "charges-max": 250 },
+      { "group": "ammo_casings_gunsmith", "prob": 1, "charges-min": 500, "charges-max": 500 }
+    ]
+  },
+  {
     "type": "item_group",
     "id": "ammo_light_batteries",
     "subtype": "distribution",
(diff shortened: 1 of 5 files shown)


data/json/itemgroups/Weapons_Mods_Ammo/ammo.json
```diff
@@ -766,6 +766,13 @@
       { "item": "shot_hull", "prob": 50, "count": [ 1, 3 ] }
     ]
   },
+  {
+    "id": "ammo_casings_gunsmith",
+    "//": "Casings found in prepper/gunsmith locations, including primed casings",
+    "type": "item_group",
+    "subtype": "distribution",
+    "entries": [ { "group": "ammo_casings", "prob": 98 }, { "item": "22_casing_new", "prob": 2 } ]
```
GuardianDll, line 774: What's the point to make a new group to add 22 instead of adding it directly to ammo_casings?

IcySlurpee: It being in `ammo_casings` causes the bug.  The `field` and `road` item groups use the `ammo_casings` group to portray where a human was killed by gunfire during the events of the cataclysm, an unused casing shouldn't appear in that situation.

data/json/itemgroups/Weapons_Mods_Ammo/ammo.json
```diff
@@ -712,7 +712,7 @@
     "subtype": "distribution",
     "entries": [
       { "item": "10mm_casing", "prob": 60 },
-      { "item": "22_casing_new", "prob": 80 },
+      { "item": "22_casing", "prob": 80 },
```
Maleclypse, line 715:
```suggestion
      { "item": "22_casing", "prob": 80 },
      
```

data/json/itemgroups/Weapons_Mods_Ammo/ammo.json
```diff
@@ -712,7 +712,8 @@
...
+
```
Maleclypse, line 716:
```suggestion
```

PR #63700, opened 2023-02-20
large metal box can hold taller items

anoobindisguise:
#### Summary
balance "large metal box can hold taller items inside it"

#### Purpose of change
I had a character who wanted to use the large metal box to haul around items, such as a stack of planks, but you can't fit tall items inside the metal crate. This doesn't make much sense since you could easily stack them vertically within the crate and unlike a sealed watertight container it isn't critical that the object fits fully within the bounding box of the item.

#### Describe the solution
Double the max length.
This may cause an issue where since the container is rigid, it won't update its length to reflect the planks or whatever in it, but the large metal box is so big that this will physically never be an issue because longest_side only matters for putting something inside of something else.

### DIFF
--- a/data/json/items/containers.json
+++ b/data/json/items/containers.json
@@ -1097,7 +1097,7 @@
     "weight": "48 kg",
     "volume": "330000 ml",
     "longest_side": "91 cm",
-    "//": "(60x60x90)",
+    "//": "(60x60x90) but you could easily fit taller objects in it",
     "price": 16000,
     "price_postapoc": 80,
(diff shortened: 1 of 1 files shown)


mqrause:
> This may cause an issue where since the container is rigid, it won't update its length to reflect the planks or whatever in it

I don't think it makes a difference if the container is rigid or not. Or at least it shouldn't.

kevingranade:
Were not making ad hoc exceptions for length limits.

PR #63702, opened 2023-02-20
Support Mac arm64

CzBiX:
#### Summary
Build "Support Mac arm64 build"

#### Purpose of change

Get native arm64 build for mac.

#### Describe the solution

Use framework for mac to avoid cross compile.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -188,6 +188,15 @@ jobs:
             artifact: osx-tiles-x64
             ext: dmg
             content: application/x-apple-diskimage
+          - name: OSX Tiles arm64
+            os: macos-11
+            arch: arm64
+            mxe: none
+            tiles: 1
+            sound: 0
+            artifact: osx-tiles-arm64
+            ext: dmg
+            content: application/x-apple-diskimage
           - name: Android x64
             os: ubuntu-latest
             mxe: none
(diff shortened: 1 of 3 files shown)


BrettDong:
As far as I know, an ARM binary compiled on a different machine won't run unless code signed with a valid Apple developer signature.

CzBiX:
yes, but there are two things.
The linker will automatically ad-hoc signing output binary, and which then can be run by "Open" in the context menu.
While I believe this is basic knowledge of using open source software on a mac, we are supposed to add it to the FAQ.
see: https://support.apple.com/guide/mac-help/open-a-mac-app-from-an-unidentified-developer-mh40616/mac

BrettDong:
What I was saying is that, the ad-hoc signature generated on machine A (in this case, GitHub Actions servers) is not accepted on another machine B (user's own Mac). When the user tries to run the binary on machine B, it will immediately be killed by the kernel (zsh: Killed). I'm not talking about the quarantine flag which can be workaround by right-click and select "open".

CzBiX:
I get it, I've confirmed it works.
if you have the ARM mac, I can upload what I built to you, so you can also verify it.

BrettDong:
If you confirm the binary compiled on GitHub Actions can run on a local ARM Mac then never mind about what I said. I don't have ARM Mac at hand now. When M1 just came out I got some and I could remember an ad-hoc signature is only valid on the same machine. Apple might have updated the security policy at some point later.

kevingranade:
See https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4271800540 for tge build errors.

.github/workflows/release.yml
```diff
@@ -262,8 +271,12 @@ jobs:
       - name: Install dependencies (mac)
         if: runner.os == 'macOS'
         run: |
-          HOMEBREW_NO_AUTO_UPDATE=yes HOMEBREW_NO_INSTALL_CLEANUP=yes brew install sdl2 sdl2_image sdl2_ttf sdl2_mixer gettext ccache parallel
+          brew install gettext ccache parallel
+          ~/build-script/get-mac-sdl2-framework.sh
```
kebkaldanil, line 275: typo: ~~build-script~~ **build-scripts**

LyleSY: Yep, no such file or directory, this is it

PR #63707, opened 2023-02-21
fix getting sick having the exponent first wrongly

anoobindisguise:
#### Summary
Bugfixes "get_sick exponent issue"

#### Purpose of change
Exponent was backwards. In #61058 it should be enviro prot to an exponent of 0.3 but because the two were reversed it was instead 0.3 to the power of the exponent causing more enviro resist to have a decreasing effect.

#### Describe the solution
put them in the right order.


anothersimulacrum:
> Exponent was backwards.

Based on what?

anoobindisguise:
> > Exponent was backwards.
> 
> Based on what?

It should be enviro prot to the power of 0.3, not 0.3 to the power of enviro prot. should be std::pow(base, exponent) when it was std::pow(exponent, base) which was backwards.

anothersimulacrum:
> It should be enviro prot to the power of 0.3

Why? What is your basis for this claim? I can do the math and see that it is correct, but please put that in your PR description (e.g. fill out the template) for this PR and future PRs.

anoobindisguise:
> > It should be enviro prot to the power of 0.3
> 
> Why? What is your basis for this claim? I can do the math and see that it is correct, but please put that in your PR description (e.g. fill out the template) for this PR and future PRs.

Alright, done.

ferociousdork:
The division by 2 is also in the wrong place, at least if you intend it to be like `((mouth environment protection)^0.3 / 2) +1` as in #61058

src/character.cpp
```diff
@@ -4964,7 +4964,7 @@ void Character::get_sick()
     // Health is in the range [-200,200].
     // Diseases are half as common for every 50 health you gain.
     float health_factor = std::pow( 2.0f, get_lifestyle() / 50.0f );
-    float env_factor = 1.0f + std::pow( 0.3f, get_env_resist( body_part_mouth ) / 2 );
+    float env_factor = 1.0f + std::pow( get_env_resist( body_part_mouth ) / 2, 0.3f );
```
kevingranade, line 4967:
```suggestion
    float env_factor = 1.0f + std::pow( get_env_resist( body_part_mouth ), 0.3f ) / 2.0;
```

PR #63709, opened 2023-02-21
Fix containers with non-CONTAINER pockets skipping active items

Kamayana:
#### Summary
Bugfixes "Fix containers with non-CONTAINER pockets skipping active items"

#### Purpose of change
Containers that start with a pocket that isn't type CONTAINER will have all following pockets skipped by visit_contents(), which notably breaks active item handling. A multi cooker, for instance, starts with a battery pocket, so active items placed inside it will not be processed. In this save, the player starts standing on top of an active grenade and with a multi cooker in their inventory with top pocket priority. All the player needs to do is pick the grenade up and it'll be put in the multi cooker and then never explode as long as it's kept in there.

[Grenade Cooking-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10788310/Grenade.Cooking-trimmed.tar.gz)

#### Describe the solution
The for loop of item_contents::visit_contents() checks to see if a pocket is type CONTAINER as it'll only check a pocket for active items if it's a CONTAINER type. However, on finding a pocket that is not a CONTAINER, the function uses `return VisitResponse::NEXT` to immediately exit the for loop, skipping all remaining pockets. I replaced that return with `continue` so the for loop could actually check through all the pockets. There's still a `return VisitResponse::NEXT` after the for loop if there are no CONTAINERs, so no functionality is lost.

I believe it'd actually be safe to also remove `if( !pocket.is_type( item_pocket::pocket_type::CONTAINER ) )`, as I don't see what it accomplishes other than potentially skipping active items in non-CONTAINER pockets, but since visit_internal() is used so widely I didn't want to risk overstepping.

#### Testing
I ran the save provided above, picked up the grenade, and it exploded despite being inside the multi cooker. I did other things with active items and such to look for any unintended consequences of editing visit_internal() but I didn't see anything. Regardless, it might be a good idea to not merge this until after the 0.G release, in case there's something I missed; visit_internal() is called from a lot of places so it'd make sense to play it safe.


mqrause (approved):

mqrause:
> I believe it'd actually be safe to also remove `if( !pocket.is_type( item_pocket::pocket_type::CONTAINER ) )`, as I don't see what it accomplishes other than potentially skipping active items in non-CONTAINER pockets, but since visit_internal() is used so widely I didn't want to risk overstepping.

The question there would be why active items would be in non-container pockets. I don't think it makes sense for any other pocket type. Opening visitable up to more pocket types would likely cause lots of (minor) issues with other stuff simply because it wasn't intended for that.

Issue #63715, opened 2023-02-21
Smokes do not trust me


BoruMaboru:
### Describe the bug

When I triggered the quest for between bringing the solar panels and deliver propectus to smokes. Suddenly smokes don't want to trade with me.

NOTE : After several debugging, smokes want to trade with me again after completing the 100 jar mission.

### Attach save file

https://send.vis.ee/download/47886e04b079723f/#AmqRxQE26lZzuctLqnrEZg

### Steps to reproduce

1. Do the quest of return prospectus or solar power for smokes.
2. Check the trade option dialogue.

### Expected behavior

Smoke can trade with me again.

### Screenshots

![hate smokes](https://user-images.githubusercontent.com/123331915/220330243-22b0c133-7665-456a-b662-55df94321d43.png)

### Versions and configuration

cdda-windows-tiles-x64-2023-02-21-0336

andrei8l:
I've played through Smoke's mission tree and couldn't reproduce. This needs a save file and/or more info.

Zireael07:
The recent similar issue I recall was due to the user picking up stuff that was the faction's...

BoruMaboru:
> I've played through Smoke's mission tree and couldn't reproduce. This needs a save file and/or more info.

I've added the save file recently

andrei8l:
Your trust with the Free Merchants is `-8` but there's no indication why.

PR #63717, opened 2023-02-21
debug/tgz: trim overmapbuffer in minimized archive

andrei8l:
#### Summary
None

#### Purpose of change
The debug minimized archive is not trimmed enough for long-running games - specifically the overmap is not trimmed at all.

Some edge cases aren't handled correctly: broken symlinks and paths longer than 100 chars.

#### Describe the solution
Trim the overmap buffer to current and adjacent overmaps and drop temporary files too.

Handle the edge cases even though we're unlikely to see them occur naturally.

N/A

#### Testing
The save folder from #63715 packs to a file smaller than 25Mb (~15Mb in this case). Verify that the visible overmap looks the same on reload.

Game doesn't crash when packing a save folder with broken symlinks.

Really long (100+ chars) paths and filenames are handled correctly.

#### Additional context
N/A


andrei8l:
Thanks for merging

PR #63733, opened 2023-02-22
Make MOLLE belts not conduct electricity

Inshal240:
#### Summary
Content "Make the MOLLE webbing belt non-conductive."

#### Purpose of change
The MOLLE belt description says it conducts electricity which does not match similar items like the leather belt or load bearing vests. The latter are made out of similar material and I could not think of a reason why the MOLLE belt should be different.

#### Describe the solution
I added the NONCONDUCTIVE flag to the belt in the json file. I am not sure if this is the correct solution as something might be reading the materials on the code side and assigning the property incorrectly? However, this seems like a simple fix.

None. I could look into the code but prefer to keep this change simple as it would be my first contribution.

#### Testing
Loaded the game and looked at the MOLLE webbing belt's description in-game to see that it no longer conducts electricity.

### DIFF
--- a/src/debug_menu.cpp
+++ b/src/debug_menu.cpp
@@ -244,6 +244,10 @@ struct fake_tripoint { // NOLINT(cata-xy)
     int x = 0, y = 0, z = 0;
 };
 
+struct OM_point { // NOLINT(cata-xy)
+    int x = 0, y = 0;
+};
+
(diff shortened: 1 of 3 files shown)


Inshal240:
Shoot, I wanted to move the change to its own branch as the message said

Inshal240:
Started another PR with the new branch here. PR#63755

PR #63735, opened 2023-02-23
Bionic fitness band

anoobindisguise:
#### Summary
Content "Calorie and heart rate tracker bionic"

#### Purpose of change
Fitness band was a good candidate to make into a bionic as it would be a minimally invasive installation and would have a clear advantage over a fragile exterior wristband.

#### Describe the solution
Fitness band is a new low-tier bionic, uses 1 torso slot and passively tells you your calorie info in the eat menu. ~~It can be activated to tell you your heart rate like the fitness band.~~

~~Find a way to make the bionic use 0 power to activate as it currently uses 1 mJ, the lowest possible amount.
Make it provide extra info of some kind.~~
Give it to the Bionic Athlete and Bionic Runner.

#### Testing
Tested and confirmed I couldn't see my calorie menu without the bionic, then could see it with the bionic. ~~Then activated it and confirmed it was showing me its info correctly.~~

#### Additional context

fitnessbandbionic0
fitnessbandbionic1

### DIFF
--- a/src/bionics.cpp
+++ b/src/bionics.cpp
@@ -91,6 +91,7 @@ static const bionic_id bio_blood_filter( "bio_blood_filter" );
 static const bionic_id bio_cqb( "bio_cqb" );
 static const bionic_id bio_emp( "bio_emp" );
 static const bionic_id bio_evap( "bio_evap" );
+static const bionic_id bio_fitnessband( "bio_fitnessband" );
 static const bionic_id bio_flashbang( "bio_flashbang" );
 static const bionic_id bio_geiger( "bio_geiger" );
 static const bionic_id bio_gills( "bio_gills" );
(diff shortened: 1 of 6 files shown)


src/bionics.cpp
```diff
@@ -938,6 +939,32 @@ bool Character::activate_bionic( bionic &bio, bool eff_only, bool *close_bionics
...
+                                   as_avatar()->get_daily_ingested_kcal( false ),
+                                   as_avatar()->get_daily_ingested_kcal( true ) ) );
+        msg.append( "\n" );
+        msg.append( "-> " );
+        msg.append( string_format( _( "You burned %d kcal today and %d kcal yesterday." ),
+                                   as_avatar()->get_daily_spent_kcal( false ),
+                                   as_avatar()->get_daily_spent_kcal( true ) ) );
+        //TODO add whatever else makes sense (steps, sleep quality, health level approximation?)
+        add_msg_if_player( m_neutral, msg );
+        popup( msg );
```
andrei8l, line 967: Can you not reuse the fitness band code here?
```suggestion
        item bioband( "bio_fitnessband" );
        iuse::fitness_check( this, &bioband, true, tripoint_zero );
```

anoobindisguise: Yeah that might actually be better. I wanted the bionic to not have the motivational flavor text and be more purely informational, and I also want it to be usable if you are illiterate but it would save a lot of code room.

AnotherSeawhite:
Maybe make it to use around 100J. Integrated Dosimeter and Blood Analysis CBMs use around 100J.

src/bionics.cpp
ehughsbaird: Extract the common parts to a function and use that instead.

PR #63736, opened 2023-02-23
buff the efficiency of the trickle charger CBM

anoobindisguise:
#### Summary
Balance "Make the trickle charger a little better"

#### Purpose of change
Trickle Charger bionic was a little on the weak side and not super worth it compared to installing a proper power gen bionic, it was just never quite enough for its slot cost.

#### Describe the solution
Went from 2 J trickle for 7 slots to 3 J trickle for 5 slots.

Don't buff it at all, because it's free power at no cost without Bionic Slots installed.


Raikiri:
I actually felt that with no bionic slots it's in a good place. Perfectly usable, has good utility for no investment. With bionic slots it's obviously not the case. So maybe the answer would be to just reduce the cost?

kevingranade:
No meaningful rationale.

PR #63737, opened 2023-02-23
Drag much heavier vehicles

anoobindisguise:
#### Summary
Balance "You can drag heavier vehicles"

#### Purpose of change
Vehicle weight drag limits were unrealistically low. There was also a hardcoded limit to max weight that was too easy to reach.
https://www.thearmwrestlingarchives.com/the-origin-of-the-wheelbarrow-contest.html
The article linked involves moving a wheelbarrow full of lead ingots over 25 feet without the wheelbarrow tipping over. This is an ideal scenario to map in-game vehicles to real life as the wheelbarrow has to be lifted and kept steady, thus giving us a "worst case" scenario as a wagon or a shopping cart would be much easier.
Vehicles in neutral are extremely easy to move also. While it's not as easy to actually angle the direction of a vehicle, the wheelbarrow example means you have to be able to lift one end of the vehicle you are dragging so it presents a sane limit (pick up the car by the bumper and rotate it just as you would a wheelbarrow full of lead) 

#### Describe the solution
The heaviest wheelbarrow pushed in the above example was over 4,000 lbs, though I'll lower it to 3,000 to account for more sustainable action and possible bumpy terrain. If we assume that person is a 20 strength individual then that would give us a required_strength of (vehicle weight in kg) / 75 instead of the current (vehicle weight in kg) / 25. There are examples in the article of people moving lighter loads (2,000 lbs) over as much as 1/3 a mile. In the extremely likely event that person is not 20 strength, it would be an even more generous formula such as (vehicle weight in kg)  / 100 or (vehicle weight in kg)  / 150 and I think there's a strong case to be made that should be the case especially if 20 is no longer the literal human limit.

 I also raised the max weight of vehicles from 1000 kg to 5000 kg. People in real life can drag firetrucks behind them so 1000 kg was absurdly low. 
 
 I also raised the range in which you can barely drag the vehicle, previously it was only 1 str point, now it is 3 str points of room to get slowdown while dragging a vehicle.
 
 Finally whether a vehicle had wheels or not had very little effect on whether it was draggable, only causing the strength requirement to increase by 1. I changed this to being 4x the strength requirement because wheels make a massive difference in your ability to move heavy loads.

Increase the strength limit even more.
Implement a more granular and harsher slowdown since the people moving 3000 lbs of lead in a wheelbarrow weren't moving very quickly. In game you'd be moving about 1 mph assuming you could barely move the vehicle.

#### Testing
Tested this out and confirmed the new weight limits are working as intended.

### DIFF
--- a/src/grab.cpp
+++ b/src/grab.cpp
@@ -79,18 +79,19 @@ bool game::grabbed_veh_move( const tripoint &dp )
 
     //vehicle movement: strength check
     int mc = 0;
-    int str_req = grabbed_vehicle->total_mass() / 25_kilogram; //strength required to move vehicle.
+    int str_req = grabbed_vehicle->total_mass() / 75_kilogram; //strength required to move vehicle.
+    
     // ARM_STR governs dragging heavy things
     int str = u.get_arm_str();
 
     //if vehicle is rollable we modify str_req based on a function of movecost per wheel.
 
-    // Vehicle just too big to grab & move; 41-45 lets folks have a bit of a window
-    // (Roughly 1.1K kg = danger zone; cube vans are about the max)
-    if( str_req > 45 ) {
+    // Very strong humans can even drag multiple fire trucks around, but orientation would be difficult.
+    // hard limit of 5000 kg here since mutants, hydraulic muscles, artifacts etc can get str quite high
+    if( str_req > 66 ) {
         add_msg( m_info, _( "The %s is too bulky for you to move by hand." ),
                  grabbed_vehicle->name );
-        return true; // No shoving around an RV.
+        return true; // No shoving around an APC.
     }
 
     const auto &wheel_indices = grabbed_vehicle->wheelcache;
@@ -110,7 +111,8 @@ bool game::grabbed_veh_move( const tripoint &dp )
             str_req = mc / wheel_indices.size() + 1;
         }
     } else {
-        str_req++;
+        //massive penalty if it doesn't have wheels
+        str_req*=4;
         //if vehicle has no wheels str_req make a noise.
         if( str_req <= str ) {
             sounds::sound( grabbed_vehicle->global_pos3(), str_req * 2, sounds::sound_t::movement,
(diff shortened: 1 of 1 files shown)


Raikiri:
I always had this RP pipe dream of roleplaying a hulk and tossing vehicles around to clear path for my deathmobile. Honestly I think you should be able to not only to drag them, but also to toss (if your strength vastly exceeds the requirement to drag).

PatrikLundell:
There are a couple of problems with pushing and throwing large vehicles around.
The first one is one of traction, i.e. being able to actually get enough "grip" on the ground to move the vehicle rather than pushing yourself away from it as your feet slip. This is particularly tricky on rough terrain where the weight of the vehicle increases the amount of purchase you need to be able to exert the power required.
The second one is one of balance: how would you get the combined vehicle/self center of gravity inside the area of support (i.e. the line between your legs) when trying to lift it? A bonus problem is to avoid being pushed into the ground when throwing the vehicle.

Raikiri:
The way I imagine throwing a vehicle is by flipping it over. We can't simulate actual flipping, but throwing is close enough.

Raikiri:
By the way, here's normal (presumably unmutated) people flipping some cars: https://www.youtube.com/watch?v=Y-fK3xj5Gak

Assume these are 14str peeps. Now why a 30str mutant should not be able to do the same with a bus?

SirPendrak:
In game we have not only mutations but also bionics, passive and active that cobined can realistically make you over 20 str. 
And i think some endgame Rivtech stimulants and healing inections do this too, not sure about effects of adrenaline. 
Add to this some mutagens giving additional stats. And artifacts. Cars should be flying when punched at this point.

kevingranade:
Your example is on a flat hard surface over a limited distance,  it doesn't remotely hold for moving on inclines, bumpy surfaces, soft ground, or over indefinitely long distances.

Raikiri:
:(

PR #54385, opened 2022-01-14
Aftershock Exoplanet: Glacial tunnels

John-Candlebury:
#### Summary
Mods "Aftershock Exoplanet: Glacial tunnels, a new low-risk area."

#### Purpose of change
These tunnel systems are intended to be low to mid-danger areas that you can explore and forage for food, not completely unlike vanilla forests. Later on they should also function as a refuge from hazardous weather conditions and as entry points for more dangerous and deeper dungeons

Note that this PR only adds the bulk of the  tunnel mapgen, and the tunnels currently have no monsters, harvestable plants or rewards of any sort. These would need to come in successive and hopefully much smaller PRs

#### Describe the solution
I  designed these to be relatively easy to expand in the future.  Different tunnel "biomes" can be easily added through the creation of new mapgen palletes, and all the map pieces are assembled piece-by-piece using chunks, with different chunks providing the base floor, then walls, and then special pieces like rivers, or large decorative elements. So adding future tunnel types would mostly be a matter of creating a few more chunks and altering the spawning order/frequency of these chunks.

Anyhow I did two biomes to test the concept,
Heres a Volcanic version of the tunnels, full of geothermal vents and jagged rocks.
image

And heres the Forested version of the tunnels, which are much more lush and full of plants:
image

And also made underground river maps to test the chunk layering idea:
image

#### Testing
This is a wip mostly becuase I still need to fix a few errors in the mutable overmap definition.

PR #63744, opened 2023-02-23
Adds reality distortion map extras.

DoctorBoomstick:
#### Summary
Content " A very heavily in progress and unfinished implementation of a new, portal storm-effected biome."

#### Purpose of change
Underlined by the existence of entities such as distorted labs and portal storms, it is evident that the cataclysm has destabilized the border between dimensions, allowing spillovers from alien realities. That being said, the player will rarely ever see such anomalous happenings: outside of government black sites, intermittent portal storms, the occasional tear in reality, and warped trees. I personally feel that, while these current game entities highlight the Lovecraftian elements of the setting, they do not encapsulate the whole extent of the reality-spanning damage caused by the blob’s invasion.
Areas with evident portal distortion, elements from other realities leaking into our own, and signs of the larger, dimension-spanning destruction of the Cataclysm, I feel, would assist in further flushing out the extent of the apocalypse’s effects. To a large extent, the superb work on the portal dungeon already takes steps into exploring the reality-twisting side effects of the cataclysm, though, in my opinion, adding some signs of further breakage outside of portal storms and things that signpost to the faits of other realities would be rather interesting. To this end, perhaps allowing the player to, under certain conditions, peek into the landscapes of other realities could open the door for interesting biomes, items, and storytelling opportunities while also cementing further the cosmic horror elements of the game.

#### Describe the solution
This is a heavily unfinished and still-in-progress draft of a few portal/reality distortion map extras. Specifically, this includes the basic framework for a subway survivor encampment falling victim to a spillover from another reality and experiencing a subsequent time looping event. Though the map extra itself is prone to a rework and some further changes, the monsters and terrains should be representative of the angle I am aiming for. There are a number of features left that I’d like to include, such as (ranging from the simplest to the most optimistically complicated),
1.	Adding extra nested maps, flushing out the camp’s structure, and creating more variants
2.	Adding extra crystalline furniture, monsters, and terrains
3.	By using notes and ambient storytelling, providing the player with an idea of what happened at the camp.
4.	adding a similar effect to the déjà vu portal storm effect (which causes the game to rewind terrain, furniture, and player placement several turns) that will activate while the player is under the influence of the map extra.
5.	Adding a trigger (similar to that found within the portal dungeon) that will briefly teleport the player to a very small portion of the origin of the alien anomaly
6.	Adding other map extras and exploring other forms of reality overlap and distortion

All of this being said, I anticipate a number of issues in creating these types of map extras, including,
1. Are these types of reality overlaps in keeping with the game’s tone, lore, and style?
2. Are these map extras too haphazardly laid out and incoherent?
3. Trying to ensure that direction-dependent extras, such as the present subway camp, spawn in the right direction
4. Should the player ever be allowed to see the faits  of other realities, and what would such realities even look like?
5. how to make this particular map feel less cramped.

1.	I originally considered making the anomaly a flesh biome, in the same spirit as SCP 610. While such a biome may still be in the works, as it would represent the eventual reality of an advanced, blob-infested reality, I personally find the concept already well implemented in existing areas of the game and other popular media. I thought that something like predatory crystalline beings would be more original and perhaps somewhat more unique.
2.	I am still, at this moment in time, contemplating reworking the map extra to be a larger, fully fledged map. The reason for this is that, as matters stand, the present camp feels rather cramped already, and half of the nested maps aren’t even included yet. If I am to rework the existing map extra, I believe I’d lean towards transferring it to an underground, anomalous fallout shelter or some other similar structure, while map extras would be relegated to smaller, more individualistic events.

#### Testing
WIP

#### Additional context
As a sidenote, the original handful of commits were an accidental merge from another PR of mine, which has since been removed.


John-Candlebury:
Hey this is pretty cool!
I had simmilar concepts I wanted to explore with mutable dungeons using the -4 subway and the Hub Ancilla. If you want to collaborate with these in the future feel free to message me in the dev discrod

DoctorBoomstick:
@John-Candlebury , thanks for the offer, it’d be great to collaborate on something of this nature in the future. Regarding this particular style of content, my current thoughts have developed in to the following,
1.	As I’m a bit wary of placing map extras in something like a subway tunnel, without a guaranty that the rotation will follow the configuration I want it to follow, and I don’t know if you can set EOCs in map extras, I am thinking of reducing this map extra to a temporary field encampment, with only a couple of tents. This camp will become something of a scavenger encampment, perhaps trying to access subway labs, hence their presence in the subways. Their downfall will have been caused by a portal storm, which caused some tasty interdimensional nightmare fuel to dramatically reduce their lifespan.
2.	I will likely move the crystal stuff to a different location, a large fallout shelter or farmhouse for example, which would have been the main base for the dead scavengers in the subway map extra. In the latter location, a map would more than likely be placed to point the player to the bunker or whatever building I finally settle upon. The base itself would be overgrown with this crystalline material, with all personnel experiencing the time anomaly described within this PR’s included monsters. An area within the location, the root source of the crystalline mass, will have an examinable tile which will transport the player to an alternate reality Earth, which has fallen victim to this overgrowth and the blob. My current inspiration for this alternative Earth is SCP 093, with some changes.
I think this implementation is a bit neater, and leaves room for some more interesting stuff to be included. That being said, I neither have ever tried creating something to this extent, though I think it can be done, nor do I know if this is a good idea in keeping with the game’s style, so any thoughts would be welcomed with open arms.

John-Candlebury:
My intention is to add labyrinth style dungeons that can spawn in the lab subways, their end form wouldn't be unlike the glacial tunnels of Aftershock, which you can see here #54385. All with the express intent of having more stuff down there that isn't just more labs or lab variations, as well as making travel in the subway more interesting.

Procyonae:
Any chance you'll pick this back up at any point?

DoctorBoomstick:
> Any chance you'll pick this back up at any point?

Unfortunately, like quite a few things, this is one of those PRs I’d really like to pick up again; however, it’s subject to time availability, skills, and a good plan. Currently, in hindsight, the map extra form of this layout is laughably dumb, and I always wanted to move it to a true map, but I feel like this would be far too similar to other content we already have.

PR #63751, opened 2023-02-23
Add real dark gray from the 256 color palette

### LINKED ISSUE #57385: Do not expect bold black to be drawn as dark gray
### Describe the bug

A user just asked for help in the IRC channel, stating they could not see the following colors because they are solid black:  c_black (normal); c_dark_gray (normal); i_black (inverted); i_dark_gray (inverted)

Ignore the black since it is supposed to be invisible, because dark_gray is the actual issue here. https://github.com/kovidgoyal/kitty/issues/197 demonstrates this quite well. Alacritty is affected, too but there is the `draw_bold_text_with_bright_colors: true` option, luckily.

image

This is a screenshot from kitty, where there is no option to control the behavior. Note the invisible roads, parking lots and weird rivers.

### Steps to reproduce

1. Start the game in either kitty or alacritty.
2. Check your map or any other UI element using dark gray.
3. Go quaff from a fountain or put on a ring of see invisible to be able to see those roads.

### Expected behavior

Dark gray is not bold black. Light blue is not bold blue. There are proper color codes luckily.

Relevant: #880

### Screenshots

_No response_

- experimental

### Additional context

_No response_

SJonesy:
#### Summary
Bugfixes "Fix dark gray in the ncurses client for terminal emulators that support 256 colors."

#### Purpose of change
Fixes #57385

#### Describe the solution
The ncurses client now uses number 237 on the 256 color palette (if available in the terminal emulator) for dark gray instead of using black with the intensity bit set. This makes it so many terminal emulators will no longer require using the "interpret bold colors as bright" setting globally. The ncurses client does this by detecting if the terminal emulator being used supports >= 256 colors and this is exposed in the catacurses namespace.

The dark gray color could realistically be anywhere in the range of like 235-239. I like 237 the best.

#### Testing
Opened the client and checked both Settings->Colors and a road in game to make sure the dark gray displayed correctly. I also verified a few other places like inside a dark house and the crafting menu.

#### Additional context
We have a color pair budget of 256 from ncurses, and this change adds the 88th.

### DIFF
--- a/src/ncurses_def.cpp
+++ b/src/ncurses_def.cpp
@@ -425,23 +425,6 @@ int nc_color::to_color_pair_index() const
 
 nc_color nc_color::bold() const
 {
-    // TODO check toggleable setting
-    // if (COLORS >= 256)
-    // {   
-    //     short fg, bg;
-    //     int color_pair_index = to_color_pair_index();
-    //     pair_content(color_pair_index, &fg, &bg);
-    //     if (fg == black)
-    //     {
-    //         switch (bg)
-    //         {
-    //         case:
-                
-
-    //         }
-    //     }
-    // }
-
     return nc_color( attribute_value | A_BOLD );
 }
 
--- a/src/ncurses_def.cpp
+++ b/src/ncurses_def.cpp
@@ -262,11 +262,11 @@ void catacurses::init_interface()
     set_escdelay( 10 ); // Make Escape actually responsive
     // TODO: error checking
     start_color();
-    init_colors();
-
+    // creates the color_manager and sets 256 color mode if available
     if (COLORS >= 256) {
         get_all_colors().enable_256_colors = true;
     }
+    init_colors();
 }
 
 void input_manager::pump_events()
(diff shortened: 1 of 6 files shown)


Zireael07:
Plz edit so that there is no fictious link to issue 237 (e.g. by putting a space between hash and number)

SJonesy:
> Plz edit so that there is no fictious link to issue 237 (e.g. by putting a space between hash and number)

Done!

src/ncurses_def.cpp
```diff
@@ -265,6 +265,11 @@ void catacurses::init_interface()
     init_colors();
 }
 
+bool catacurses::supports_256_colors()
+{
+    return( COLORS >= 256 );
```
ZhilkinSerg, line 270:
```suggestion
    return COLORS >= 256;
```

src/wincurse.cpp
```diff
@@ -656,6 +656,11 @@ void catacurses::init_interface()
     initialized = true;
 }
 
+bool catacurses::supports_256_colors()
+{
+    return( COLORS >= 256 );
```
ZhilkinSerg, line 661:
```suggestion
    return COLORS >= 256;
```

PR #63757, opened 2023-02-24
Fix helicopters to consume fuel at a realistic rate

### LINKED ISSUE #63419: Lower Helicopter Fuel Consumption Rate

After testing various helicopter flight profiles, I'm finding that helicopters across the board are burning fuel many times faster than their real life equivalents. The civilian helicopters are also too heavy since they're being built from steel frames.

Using a helicopter pilot with max vehicle skills and spawning in undamaged helicopters, I found the following:

BLACKHAWK:
- The fuel capacity is accurate. The weight and size are a bit low, but in the right ballpark
-Burns fuel at about 3x the IRL rate. In game they can fly for anywhere from ~1-1.5hr depending on their speed, but IRL the Blackhawk has an endurance closer to 3-4.5hrs.

MEDEVAC
- Fuel capacity is about right. Weight is about 2x high, should probably be made from lighter frames. Only military helicopters are made from steel frames, civilian models are usually made from composites or aluminum which are many times lighter.
-Burns fuel at about 6-8x the IRL rate. In game can only fly for 0.5hr at the most efficient cruising speed, but real 4-seater medevac models have an endurance in the range of 3-4hrs.

SMALL FLIER
- Too heavy and carries too much fuel. In-game version weighs 4749lbs and carries 400L of fuel. IRL single-seaters weigh in at 500-100lbs and carry 50-100L of fuel. Even the larger models (mostly military) top out around 2500lbs and 200L.
- Top speed is too high (225mph in game vs ~100mph IRL)
- Burns fuel at too high a rate. Even with the 400L fuel capacity, can still only fly 1-1.5hrs. Single seaters IRL are usually designed to an endurance time of 2-2.5hrs on a 50-100L tank.

ALL MODELS
- Burn way too much fuel at max speed, to the tune of several liters of fuel per second. IRL helicopter speed is limited by blade aerodynamics, not engine performance. Helicopters are only going to burn fuel at about twice the rate when flying at max speed compared to cruise. Eg. the Blackhawk in game can only fly for about 50s at max speed, but in real life could sustain max speed for close to an hour.

### Solution you would like.

- Fuel consumption rate of all motors should be decreased across the board. Most helicopters should be able to stay in the air 2.5-4.5hrs if flying at normal cruising speeds. They should be able to fly for roughly half that long when either hovering in place or flying at their max design speed
- Civilian models should be constructed from extra light frames to represent aluminum/composite construction, and the maximum flight weight of these vehicles should be adjusted to match their new weights
- Civilian models should have lower max speeds. While military combat helicopters can usually push speeds over 200mph, common civilian models usually top out somewhere in the 100-150mph range
- The single seater should probably carry less fuel (50-200L range), and have its own rotor with a lower lift capacity than the one used for the medevac

I did not run extensive tests for the Apache, the Osprey, or the 2 seater, but they have similar issues.

_No response_

### Additional context

See below spreadsheet for in-game flight time calculations, sorted by vehicle model and mode of flight.
[cdda_helicopter_flight_times.xlsx](https://github.com/CleverRaven/Cataclysm-DDA/files/10608266/cdda_helicopter_flight_times.xlsx)

crp763:
#### Summary
Category "Balance"

#### Purpose of change

Updated source code and JSON data files so that helicopters consume a realistic amount of fuel, fly at realistic speeds, and have more realistic designs. Helicopter performance is now realistically governed by the mass, lift, and drag of the vehicle.

Also made player messages more helpful when attempting to fly a helicopter (removed message about lacking wheels, added message for if helicopter is too heavy to take off).

Fixes #63419

#### Describe the solution

The original power vs velocity curve in vehicle_move.cpp had incorrect units, and all helicopters used the same curve regardless of their governing performance parameters. All helicopters were given a fixed max flight speed of 225mph in vehicle.cpp regardless of design, which meant that helicopters were almost always using 100% of their power to fly,. Helicopter designs in helicopters.json were unrealistic for civilian helicopters, since they were too heavy and had engines that were too powerful.

The code has been updated to use a more realistic power vs velocity curve, with corrected units. Helicopters now use an appropriate amount of power to fly, and their performance is governed by the correct performance parameters (lift, drag, mass). The JSON files have been updated improve civilian helicopter design.

Using actual helicopter performance equations would be possible, but would add a lot of complexity and performance overhead to the code. The updated code seems to do a good job of approximating performance by using only simple polynomials.

#### Testing

All helicopters that are available in-game without debug spawning (Blackhawk, medevac, 2 seater, small flier) were spawned and tested to ensure that they had realistic weight, lift, drag, speed, and fuel consumption. Also performed various tests to make sure new player messages work properly when helicopter takes off, or attempts to take off. Used debug messages during testing to ensure that the power vs velocity curve was working as expected.

The Matlab/Octave script in the zip file below contains the same performance equations as used in this PR, and will generate plots of helicopter power consumption and flight range as a function of mass, lift, and drag.
[cdda_helicopter_performance.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10819825/cdda_helicopter_performance.zip)


crp763:
Closing pull request: requested merge into wrong branch

PR #63761, opened 2023-02-24
Fix helicopter fuel use

### LINKED ISSUE #63419: Lower Helicopter Fuel Consumption Rate

crp763:
#### Summary
Balance "Fix helicopter fuel use"

#### Purpose of change

Updated source code and JSON data so that helicopters consume a realistic amount of fuel, fly at realistic speeds, and have more realistic designs. Helicopter performance is now correctly governed by the mass, lift, and drag of the vehicle.

Created more helpful player messages when attempting to fly a helicopter. Removed message about lacking wheels when taking off, and added message for if helicopter is too heavy to take off).

Fixes #63419

#### Describe the solution

The original power vs velocity curve in vehicle_move.cpp had incorrect units, and all helicopters used the same curve regardless of their governing performance parameters, which meant that helicopters were almost always using 100% of their power to fly. All helicopters were given a fixed max flight speed of 225mph in vehicle.cpp regardless of design. Helicopter designs in helicopters.json were unrealistic for civilian helicopters, as they were too heavy and used engines that were too powerful.

The code has been updated to use a more realistic power vs velocity curve, with corrected units. Helicopters now use an appropriate amount of power to fly, and their performance is governed by the correct performance parameters (lift, drag, mass). The JSON files have been updated to improve civilian helicopter design. The unit test for the small turbine engine has been updated to reflect the JSON changes.

Using actual textbook helicopter performance equations would be possible, but would add a lot of complexity and performance overhead to the code. The updated code seems to do a good job of approximating performance by using only simple polynomials.

#### Testing

All helicopters that are available in-game without debug spawning (Blackhawk, medevac, 2 seater, small flier) were spawned and tested to ensure that they had realistic weight, lift, drag, speed, and fuel consumption. Also performed various tests to make sure new player messages work properly when helicopter takes off, or attempts to take off. Used debug messages during testing to ensure that the power vs velocity curve was working as expected. Ran unit tests to ensure changes did not introduce any new failures.

The Matlab/Octave script in the zip file below contains the same performance equations as used in this PR, and will generate plots of helicopter power consumption and flight range as a function of mass, lift, and drag.
[cdda_helicopter_performance.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10819825/cdda_helicopter_performance.zip)


crp763:
Unsure why code style reviewer is failing on vehicles.cpp. Running AStyle v3.0.1 with all the correct options from CODE_STYLE.md shows no changes or issues.

src/vehicle.cpp
```diff
@@ -7727,9 +7755,9 @@ int vehicle::part_count( bool no_fake ) const
     return no_fake ? std::count_if( parts.begin(), parts.end(), []( const vehicle_part & vp ) {
         return !vp.is_fake;
     } ) : static_cast<int>( parts.size() );
-}
+    }
```
SurFlurer, line 7758: Are these spaces causing the astyle test to fail?

Zireael07: Most likely

crp763: Good catch, I think AStyle got confused by the format of that method. Thanks for taking a look!

data/json/items/vehicle/rotor.json
```diff
@@ -4,26 +4,42 @@
...
     "name": { "str": "heavy-duty military rotors", "str_pl": "sets of heavy-duty military rotors" },
-    "description": "A set of four rotor blades from a military attack helicopter.",
+    "description": "A large four blade rotor assembly from a military attack helicopter.",
     "price": 23400,
     "price_postapoc": 500,
-    "weight": "452000 g",
-    "volume": "452000 ml",
+    "weight": "1125000 g",
+    "//": "Increased weight from ~1000lbs to ~2500lbs to account for the mast, swashplates, transmission, pumps, and all other relevant parts of rotor assembly besides the blades",
+    "volume": "1125000 ml",
```
SurFlurer, line 12: Although there's no chance to obtain this item "as an item", but being larger than 1000L will prevent it from being spawned via debug. However, I don't think there's any other impact.

anoobindisguise:
You should probably introduce a fourth turbine instead of turning the 1350 hp turbine into a 400 hp turbine in case the M1 Abrams ever makes it into the game, because that's more or less what that engine was intended for and part of why it's way heavier than the 1900 and 6000 hp ones.

src/veh_interact.cpp
```diff
@@ -2720,6 +2720,13 @@ void veh_interact::display_stats() const
                     veh->coeff_air_drag() );
     i += 1;
 
+    if( veh->has_rotors() ) {
+        fold_and_print( w_stats, point( x[i], y[i] ), w[i], c_light_gray,
+                        _( "Max Lift:     <color_light_blue>%7.0f</color> lbs" ),
+                        veh->lift_thrust_of_rotorcraft( true ) * 0.2248 );
```
irwiss, line 2726: This needs to use convert_weight instead of hardcoding lbs, look at line 2670 for an example

taatu:
Title ought to be

`Balance "Title"`, not `Category "Balance"`

data/json/items/vehicle/rotor.json
crp763: I edited the weight and volume down to 1000kg and 1000L to fix the spawning issue. The change doesn't make much difference for the Blackhawk (250lbs difference to a 10,000lbs vehicle), and these weight/volume figures are educated guesses anyway.

src/vehicle_move.cpp
```diff
@@ -484,10 +484,60 @@ void vehicle::thrust( int thd, int z )
...
+        //     With a lift:weight ratio of 2, a speed of 202mph can be reached
+        // This curve allows a Blackhawk weighing 16,000lbs with 23,500lbs of lift and an air_drag of 1.0 to fly at a approx 180mph safely, vs a rated top speed of 183mph
+        //
         thrusting = true;
+        const double mass_kg = to_kilogram( total_mass() );
+        const double mass_lbs = mass_kg * 2.20462;
+        const double mass_scalar = std::pow( mass_lbs / 16000, 0.3 );
+        double velocity_mph = velocity * 0.01;
+        const double power_scalar = ( ( to_kilogram( total_mass() ) * 9.8 ) / lift_thrust_of_rotorcraft(
+                                          true ) ) * 2.0;
```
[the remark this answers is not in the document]
crp763: Suggested change has been committed

src/vehicle_move.cpp
```diff
@@ -484,10 +484,60 @@ void vehicle::thrust( int thd, int z )
...
+                    500;
+        } else {
+            // The inverse of this quadratic is solved for in vehicle.cpp to determine the max safe airspeed
+            // If you modify this equation, you must update the inverse equation
+            value = ( ( 1.0 / 18 ) + ( ( d - 1.0 )  / 50.0 ) ) * std::pow( velocity_mph, 2 ) +
+                    -10 * velocity_mph +
+                    750;
+        }
+        value *= power_scalar;
+        load = value;
```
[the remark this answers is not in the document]
crp763: Done and committed

kevingranade:
This is difficult to evaluate in its current state,  how did you reach your expeced outcomes,  and how did you test for them?
The primary metrics here are variations on flight time and fuel consumption rates,  more accurate weights make that easier to reason about,  but at the end of the day the important balance issue is around how expensive it is to operate one of these helicopters.

crp763:
> This is difficult to evaluate in its current state, how did you reach your expeced outcomes, and how did you test for them? The primary metrics here are variations on flight time and fuel consumption rates, more accurate weights make that easier to reason about, but at the end of the day the important balance issue is around how expensive it is to operate one of these helicopters.

I balanced the performance around the real world specs of the Blackhawk, S-70, and some other helicopters. But going through some more testing, it seems like smaller helicopters are too slow now, so I need to spend some more time refining this update.

kevingranade:
Never got more than hand-waving about the rationale for the changes, and it seems to be improperly labelled as a bugfix so it's not getting closed for staleness. This kind of thing very much needs a strong rationale that can be followed and documented so the number doesn't keep getting shuffled around in the future.

Issue #63759, opened 2023-02-24
Game crashes upon loading new areas


Sepentine:
### Describe the bug

Game crashes upon loading new area

### Attach save file

[Groveville-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10820038/Groveville-trimmed.tar.gz)

### Steps to reproduce

seems to just require that you load new areas and generate new map extras I think
So to replicate just walk around long enough

### Expected behavior

Ideally it would load the map without issue

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.2604 (21H2)
- Game Version: af2fe0d [32-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Portal Storms Ignore NPCs [personal_portal_storms],
    Bionic Professions [package_bionic_professions],
    SpeedyDex [speedydex],
    Stats Through Skills [StatsThroughSkills],
    Tamable Wildlife [Tamable_Wildlife]
]

### Additional context

[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/10820046/crash.log)
[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/10820047/debug.log)

Issue #63018, opened 2023-01-06
Travelling far distances causes a memory leak


Blueflowerss:
### Describe the bug

After (auto-)travelling on the overmap through many tiles the game experiences a memory leak issue.

### Steps to reproduce

1. Load up save.
2. Start travelling anywhere, seemingly doesn't matter if the terrain was saved beforehand.
3. See the game consumes way too much memory

### Expected behavior

I expect to be able to travel long distances without the game memory leaking.

### Screenshots

![image](https://user-images.githubusercontent.com/77662682/211053447-a76f7dc2-2ac7-432a-b2d5-f67c59e1d246.png)

### Versions and configuration

- OS: Linux
    - OS Version: Distributor ID: Linuxmint; Description: Linux Mint 20; Release: 20; Codename: ulyana; 
- Game Version: 28d1538 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Bionic Professions [package_bionic_professions],
    Tamable Wildlife [Tamable_Wildlife],
    SpeedyDex [speedydex],
    Xedra Evolved [xedra_evolved],
    No Hope [no_hope],
    Blaze Industries [blazeindustries],
    Bionic Slots [cbm_slots],
    Bee content [bee_content],
    Slime Ranching [slime_ranch]
]

### Additional context

I suspect it's related to the game not unloading the areas I already travelled through.
And it seems to be system specific, as no amount of updating the game or using different installations fixed it.

andrei8l:
> I suspect it's related to the game not unloading the areas I already travelled through.

Probably that. This very old comment here

https://github.com/CleverRaven/Cataclysm-DDA/blob/28d15385bb7756645aa286610a9e6b615ca4b446/src/mapbuffer.cpp#L168-L171

indicates that map buffer should be trimmed every (auto)save but the code below does not do that.

andrei8l:
My comment above was wrong. Map buffer gets trimmed every (auto)save. 

I've also tested by walking around for a couple of overmaps without saving and memory usage dropped back to start value after saving. ASan also did not find any leaks:

<details>
<summary>LeakSanitizer report</summary>

```gcc
==190832==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 5323 byte(s) in 255 object(s) allocated from:
[...]

PR #63762, opened 2023-02-24
submap: lazily allocate uniform submaps

andrei8l:
#### Summary
None

#### Purpose of change
Heap profiling shows that uniform submaps dominate memory usage.

See #63759 and #63018

#### Describe the solution
Allocate submap data only if/when changes are needed.

Virtual submap with 2 implementations: too much boilerplate
Sparse arrays: probably the better solution but I'm not convinced it's worth the code complexity for 12x12 arrays

#### Testing
Uniform submaps are still uniform.
Tests still pass.
Saving and reloading with mixed quads still works (ex: dig down then save).
Verify that mixed mapgen chunks/OM tiles like `office_tower_open_air_corner` save and reload without error.
Verify that submap reversion (portal storm dungeon) still works, including after a save/reload cycle.

Using [Groveville-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10822040/Groveville-trimmed.tar.gz) from #63759 and profiling with `heaptrack` by autodriving to the nearby ranch with autosave off.

<details>
<summary>Flame graph before</summary>

before

</details>

<details>
<summary>Flame graph after</summary>

after

</details>

#### Additional context
I had to simplify submap reversion to prevent duplicating effort there.

### DIFF
--- a/src/lightmap.cpp
+++ b/src/lightmap.cpp
@@ -155,7 +155,7 @@ bool map::build_transparency_cache( const int zlev )
                 return std::make_pair( value, value_wo_fields );
             };
 
-            if( cur_submap->is_uniform ) {
+            if( cur_submap->is_uniform() ) {
                 float value;
                 float dummy;
                 std::tie( value, dummy ) = calc_transp( sm_offset );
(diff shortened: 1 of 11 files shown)


andrei8l:
Thanks for merging!!

PR #63763, opened 2023-02-24
Lower damage for BB gun

Maddremor:
#### Summary
Balance "Remove bonus stab damage from BB gun"

#### Purpose of change

During discussions on the various Discords, the phenomena of the BB gun doing 4 damage total was brought to my attention. For something that is essentially a toy, this is quite a lot. It also seemed to be doing bleed damage, which is odd for something I would not expect to break skin in most cases.

#### Describe the solution

The BB gun had a stab bonus damage of two, which I removed with this PR.
The BB ammo itself had bullet damage 2, which I lowered to 1.

Making the BB gun into a pneumatic varmint rifle. This would either require multimag or make the BB gun into a single shot, single round capacity weapon similar to the bullet crossbow. This would probably be better served by a new item if anyone was inclined to add it.
#### Testing

I'll take the L on this one as I'm making it from a web UI lest I forget about it. I have not sat down and tested this proper. With such a trivial change, I assume it will not break things, but I have been wrong before. I will do proper testing if asked to.

#### Additional context

~~I'll probably lower the BB pellet damage from 2 to 1 in a separate PR as the web UI is not great for changing multiple files.~~
Never mind, I figured it out.

### DIFF
--- a/data/json/items/ammo.json
+++ b/data/json/items/ammo.json
@@ -617,7 +617,7 @@
     "volume": "200ml",
     "weight": "1 g",
     "ammo_type": "BB",
-    "damage": { "damage_type": "bullet", "amount": 2 },
+    "damage": { "damage_type": "bullet", "amount": 1 },
     "dispersion": 100,
     "count": 500,
     "stack_size": 200,
--- a/data/json/items/ranged/pneumatic.json
+++ b/data/json/items/ranged/pneumatic.json
@@ -21,7 +21,6 @@
     "bashing": 9,
     "to_hit": -1,
     "range": 12,
-    "ranged_damage": { "damage_type": "stab", "amount": 2 },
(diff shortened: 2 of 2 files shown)


actually-the-artificer:
I can try adding the Varmint rifle once my PR and this one goes through.

Shodan14:
Interesting, I always imagined it being the pneumatic version in my head.

Maddremor:
Okay, so I actually did the math on a lever operated spring BB guns (I used Daisy ones for reference) and with a muzzle velocity of 350 fps and a bullet weight of 5.1–5.4 gr [shooterscalculator](https://shooterscalculator.com/bullet-kinetic-energy.php) gives me a muzzle energy of 2 J which gives us a damage of √2 which in turn rounds down to 1.

Issue #11055, opened 2015-01-30
Quick Load


Gabriel-SE:
As contributor, there are plenty of situations where I must reload my save to get back to the initial state of debugging and a Quick Load hotkey will definitely help with that.
Make the Quick Load available in "List of all commands" if the game is compiled with "make DEV=1" so only developers have access to it.

Anyway, I'd like to hear other people's opinions before working on this issue since this topic is controversial.

KA101:
IME on windows, closing the window and relaunching is quick enough.

kevingranade:
I'm fine with it being in debug (non RELEASE=1) builds.  I'm not sure how
invasive it's going to be though, do you have the implementation planned
out, or just the idea?

Gabriel-SE:
Just the idea I guess, I can't say much without looking at the source code first.
But if the implementation is too invasive or requires many hacks to the source code I won't push this change.

bkentel:
You would have the hotkey function from the opening screen? Or would it be better as a command line parameter?

narc0tiq:
I understood it as an in-game hotkey at the same level as quicksave. You
can see how that would be a bit cheaty for release builds, but if you're
trying to reproduce a bug...

Gabriel-SE:
@bkentel 
> You would have the hotkey function from the opening screen? 

I usually load the same world and player name, hence the Quick Load name. It's the counterpart of Quick Save.
But another dev hotkey to load > World Name > Player Name function will be better as different name, "Load" maybe.

> Or would it be better as a command line parameter?

Command line parameter to load existing world will be nice, something like ./cataclysm --world world_name --name name
main.cpp has some sort of map sharing command line, but I'm not sure if it can be used to load existing world instantly without investigating it first.

@narc0tiq 
>I understood it as an in-game hotkey at the same level as quicksave.

Yeah, this is correct.

@kevingranade 
>I'm fine with it being in debug (non RELEASE=1) builds.

Debug build is kinda slow for me to compile, 3-4x slower than the speed of release build, so I want this on my release build but Quick Load definitely won't appear in RELEASE build of the bot.
It will only appear on _GLIBCXX_DEBUG (default debug mode) or when they compile the game with DEV=1.

kevingranade:
It's specifically _GLIBCXX_DEBUG that makes debug builds slow, try removing it.

PR #63767, opened 2023-02-24
Only allow quickload from debug menu

anothersimulacrum:
#### Summary
None

#### Purpose of change
https://github.com/CleverRaven/Cataclysm-DDA/pull/60587 enabled quickloading from the debug menu, and allowed you to bind a key to it (along with fixing it). As far as I'm aware, opinions haven't changed since https://github.com/CleverRaven/Cataclysm-DDA/issues/11055#issuecomment-72302531, so this should not be allowed.
Aside from that, quickload is expressly a debug feature, and it's use is expected to cause issues. Having a keybind for a feature that could break the game is not a good idea.

#### Describe the solution
Revert the removal of the check for `!defined(RELEASE)` from https://github.com/CleverRaven/Cataclysm-DDA/pull/60587.
Also add checks for the debug menu feature, but remove them in a second commit. I'm not sure if it's inclusion in the debug menu is approved, so the second commit can simply be dropped or reverted if it should not be there.

See above, also removing it from the debug menu in RELEASE builds.

#### Testing
Build the first commit, see that there are no quickload options in the debug menu or keybinds.
Add the second, see that you can quickload from the game portion of the debug menu.
Build with RELEASE=0, see that it is available as a keybind and in the debug menu.

#### Additional context
This issue also cropped up for the MSVC builds briefly, but was fixed by https://github.com/CleverRaven/Cataclysm-DDA/pull/58237.

### DIFF
--- a/src/debug_menu.cpp
+++ b/src/debug_menu.cpp
@@ -202,7 +202,9 @@ std::string enum_to_string<debug_menu::debug_menu_index>( debug_menu::debug_menu
         case debug_menu::debug_menu_index::PRINT_FACTION_INFO: return "PRINT_FACTION_INFO";
         case debug_menu::debug_menu_index::PRINT_NPC_MAGIC: return "PRINT_NPC_MAGIC";
         case debug_menu::debug_menu_index::QUIT_NOSAVE: return "QUIT_NOSAVE";
+#if !defined(RELEASE)
         case debug_menu::debug_menu_index::QUICKLOAD: return "QUICKLOAD";
+#endif
         case debug_menu::debug_menu_index::TEST_WEATHER: return "TEST_WEATHER";
         case debug_menu::debug_menu_index::WRITE_GLOBAL_EOCS: return "WRITE_GLOBAL_EOCS";
         case debug_menu::debug_menu_index::WRITE_GLOBAL_VARS: return "WRITE_GLOBAL_VARS";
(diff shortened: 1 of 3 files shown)


alp7292:
its unbinded and you need to search and bind a key to use it as a modder its very convenient to just click f3 to test changes as game reloads forcing quickload from menu would be inconvinient for testing

anothersimulacrum:
My reason for leaving it in the debug menu is for people (like modders) who do not compile their own game.
It's entirely inappropriate to have it just as a keybind without debug warnings, for reasons outlined in the PR body.

alp7292:
not only quickload but most of the debug options are also bindable and doesnt have debug warnings they are bindable because its easier to use. if your problem these actions not warning then add a warning not remove debugging feature

alp7292:
or we can simply rename this actions something like debug show radiation, debug quickload, debug lighting, etc to warn players

anothersimulacrum:
I'm not making a warning because I don't think that would be accepted.
I'm not aware of any other debug actions being bindable to keys except for opening the debug menu and turning on debug mode (pretty explicit) and harmless performance features like the hour timer.

angleKH:
The release builds ARE the builds intended for debugging, see CONTRIBUTING.md:
> Note on debug: You should probably always build with RELEASE=1 unless you experience segfaults and are willing to provide stack traces.

PR #63769, opened 2023-02-24
Fix flying insects not benefiting from their intended effect resistance

Venera3:
#### Summary
None

#### Purpose of change

Extends the downed immunity to `flying insect` bodytype as intended (instead of `insect_flying`).

#### Describe the solution

Changed the bodytype string that doesn't exist for the one that does.

Finally get around to per-mtype effect resistances, or make bodytypes their own struct holding that info. One of these days, but not in the freezes.

#### Testing

Spawned a giant wasp queen pre-fix, could sweep to down 100% of the time. After the fix it only happened less than half the time.

#### Additional context

It was only relevant to non-flying flying insects, so the giant wasp queen is the only one affected in vanilla.

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1591,7 +1591,7 @@ bool monster::is_immune_effect( const efftype_id &effect ) const
     }
 
     if( effect == effect_downed ) {
-        if( type->bodytype == "insect" || type->bodytype == "insect flying" || type->bodytype == "spider" ||
+        if( type->bodytype == "insect" || type->bodytype == "flying insect" || type->bodytype == "spider" ||
             type->bodytype == "crab" ) {
             return x_in_y( 3, 4 );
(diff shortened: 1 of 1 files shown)


src/monster.cpp
```diff
@@ -1591,7 +1591,7 @@ bool monster::is_immune_effect( const efftype_id &effect ) const
     }
 
     if( effect == effect_downed ) {
-        if( type->bodytype == "insect" || type->bodytype == "insect_flying" || type->bodytype == "spider" ||
+        if( type->bodytype == "insect" || type->bodytype == "insect flying" || type->bodytype == "spider" ||
```
RenechCDDA, line 1594:
```suggestion
        if( type->bodytype == "insect" || type->bodytype == "flying insect" || type->bodytype == "spider" ||
```

I assume this is what you meant based on your summary? Otherwise, another oopsie.

Venera3: Oh for crying out loud, kids, don't not sleep and commit.

PR #63772, opened 2023-02-24
give Elf-A antlers

anoobindisguise:
#### Summary
Balance "elf-a gets Antlers"

#### Purpose of change
No one had the Antlers mutation and I wanted to distribute it.

#### Describe the solution
Give it to Elf-A. Based on Huntress Wizard from Adventure Time.

Introduce a Deer mutation. Change Elf-A's mutagen creation requirements.

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -5430,9 +5430,9 @@
     "visibility": 3,
     "ugliness": 3,
     "description": "You have a pair of bumps on your skull.",
-    "types": [ "HORNS", "ELFA" ],
+    "types": [ "HORNS" ],
     "changes_to": [ "HORNS", "ANTENNAE" ],
-    "category": [ "CATTLE", "CHIMERA", "INSECT" ]
+    "category": [ "CATTLE", "CHIMERA", "INSECT", "ELFA" ]
   },
   {
     "type": "mutation",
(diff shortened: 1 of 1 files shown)


BlizzDA:
At the moment Antlers is one of those free-floating, 'raw mutagen' mutations, like the scorpion tail. They're obtainable but not part of a set line. A proper Cervid line is probably the ideal. While Elf-A is due for a rework (and someone might be working on it AFAIK), the design is more of a scientist's human/plant hybrid side-project, rather than an attempt at being an actual dryad. It's also gotten some critique for being a bit too fantasty-like and not mad science-y. 

Antlers on an Elf-A might be better as a goal for someone doing both Cervid (Moose, deer) and Elf-A mutagens. Encourages mixing. Horns are already a pre-tresh mutation for other lines, so it fits.

kevingranade:
Make elf-a characters look like a character from adventure time is not a valid rationale.

anoobindisguise:
> Make elf-a characters look like a character from adventure time is not a valid rationale.
Elf-A is completely up in the air as to what it represents, it is totally arbitrary. Pop culture examples simply give us a "basis" for stuff (like how Medical is a mix of Deadpool, Hellraiser and Jekyll and Hyde). The rationale is mostly "I think it would fit the nature dryad" theme of Elf-A, existing media properties just give an example of the "theme"

PR #63773, opened 2023-02-24
Automatically cap melee skill gain from monsters

Venera3:
#### Summary
Balance "Cap melee skill gain based on monster melee skill"

#### Purpose of change

Attacking monsters can train melee skills without limit unless explicitly specified by `melee_training_cap`, but that leads to whack-a-exploit with increasingly asspully reasonings for why monster X should give exactly Y skill XP. This introduces an automatic calculation for the cap unless otherwise specified.

#### Describe the solution

`melee_training_cap` defaults to `melee_skill + 2`.

TODO:
- [x] Force myself to add some unit tests, we'll see if I manage.

Base it on dodge, but that leads to some small critters like kittens becoming you ticket to the phat XP gainz.
Base it on difficulty but the difficulty calculation is very wonky (and has a way too broad range in any case).
Take the average of melee skill and dodge into account, but as a rule melee skill correlates way better with melee lethality (and presumably "things to learn while fighting this") than dodge.

#### Testing

Checked that the training cap does what it's supposed to. Added the automatic default, when not specified the cap is set to the desired level (checked both in debugger and with a debug mode readout I didn't commit because it technically adds a string).

Also did a unit test! A pretty clunky one, but hey.

### DIFF
--- a/doc/MONSTERS.md
+++ b/doc/MONSTERS.md
@@ -56,6 +56,7 @@ Monsters may also have any of these optional properties:
 | `melee_dice`             | (integer) Number of dice rolled on monster melee attack to determine bash damage
 | `melee_dice_sides`       | (integer) Number of sides on each die rolled by `melee_dice`
 | `grab_strength`          | (integer) Intensity of grab effect, from `1` to `n`, simulating `n` regular zombie grabs
+| `melee_training_cap`     | (integer) The maximum melee skill levels learnable by fighting this monster. If not defined defaults to `melee_skill + 2`.
 | `armor_bash`             | (integer) Monster's protection from bash damage
 | `armor_bullet`           | (integer) Monster's protection from bullet damage
 | `armor_cut`              | (integer) Monster's protection from cut damage
--- a/src/monstergenerator.cpp
+++ b/src/monstergenerator.cpp
@@ -1003,7 +1003,6 @@ void mtype::load( const JsonObject &jo, const std::string &src )
 
     optional( jo, was_loaded, "speed_description", speed_desc, speed_description_DEFAULT );
     optional( jo, was_loaded, "death_function", mdeath_effect );
-    optional( jo, was_loaded, "melee_training_cap", melee_training_cap, MAX_SKILL );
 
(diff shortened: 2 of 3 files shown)


anoobindisguise:
This is definitely more reasonable than making it based on target's dodge! And it will help quash some of the sillier melee training behaviors we have.

GuardianDll:
Can it picks the biggest from meele or dodge skill instead of flat melee?

adamkad1:
> Can it picks the biggest from meele or dodge skill instead of flat melee?

I think this is the right way tbh, you should be rewarded for hitting something that is hard to hit, even if its not exactly dangerous. And non dangerous things tend to run really fast or die quickly

Venera3:
So now the best way to train your fighting prowess is bothering a poor liquid kitten. I did consider a lot of permutations, as mentioned in alternatives, but flat melee skill had the least weird edge cases.

GuardianDll:
Edge cases can be handled by printing `melee_training_cap` manually? Also i don't think giving 8 dodge to semi-incorporeal monsters is good idea

Venera3:
Yes, and I'd rather start with the implementation that leaves the least edge cases. What dodgy critter should add more skill gain than its melee skill would imply?

Grubleafeater:
> Yes, and I'd rather start with the implementation that leaves the least edge cases. What dodgy critter should add more skill gain than its melee skill would imply?

I mean...all of them? I would assume that you'd get more training from trying to hit something small and weak(like a mouse) than something big and strong(like a hulk). Or trying to spear-fish.

Venera3:
You're not learning "hit small thing trying to run away", you're learning blocking, aiming for weakpoints, sweeps/precise strikes/whatever your martial art of fancy gates behind melee skill. Learning *that* needs capable opponents, not a reaction ball.

bombasticSlacks:
Venera's logic I think is correct on this.

A passive enemy that is superb at dodging may as well not be there as far as your swinging is concerned, so it would be the same value as training forms without an opponent or on a training dummy/speed ball.

PR #63775, opened 2023-02-24
Add "dense bones" mutation

anoobindisguise:
#### Summary
Content "New dense bones mutation"

#### Purpose of change
We could get lighter bones, but nothing in the other direction. This is logical as a necessary step to becoming a very large bearperson as you need stronger bones to support your body and all that muscular strength.

#### Describe the solution
Adds "dense bones" which does the following: 5% slower walking and melee attacks, -2 bash damage received, 20% harder swimming, -1 dodge, 5% more HP and 15% more carryweight.
It is a prereq of LARGE and is found on Chimera, Cattle, Ursine, Lizard, Beast and Alpha.

- Don't give it to Alpha. I was justifying it off the Halo lore of Spartans being sorta like Alpha mutants and having extremely strong bones. Obviously we're not Halo but mutations do draw from popular fiction a lot already.
- Don't make it mess with Dodge, because Light/Hollow bones don't affect Dodge either. Same for swim cost.
- Make it affect your body weight. There's no possibility of doing that right now and it would need to be c++'d in.
- Different amount of bash resistance. I'm not using EXTRA_BASH because I don't know if it works backwards and a flat modifier is much more useful for a player, and ARMOR_BASH applies before armor while the current implementation applies after armor.
- Make a second tier that is needed for HUGE.
- Make it increase arm strength because it affects the max power your muscles can put out, and do the inverse for Light/Hollow bones.
- Give it to other trees.
- Make it make you need more calcium.


Maleclypse:
If it’s a prerequisite for large you need to make sure you don’t break the dinomod trees and mastodon in aftershock.

Maleclypse:
> If it’s a prerequisite for large you need to make sure you don’t break the dinomod trees and mastodon in aftershock.

You had already done that I see now.

PatrikLundell:
Does this have any backing in reality? I was under the impression large animals just had larger bones, not denser, and that dinosaurs had bones more akin to those of birds (in terms of structure) than those of mammals, but I may be wrong.

SirPendrak:
My two cents would be that it does not rlly fit alpha, and belongs better to true tanky monster mutation trees or bulky animal trees. 
Also maybe it should give small bash bonus to unarmed melle attacks?

LyleSY:
Yes, larger animals do have stronger bones. The specific mechanics vary between mammals, birds, and other dinosaurs https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0237042

Nebnis:
> My two cents would be that it does not rlly fit alpha, and belongs better to true tanky monster mutation trees or bulky animal trees. Also maybe it should give small bash bonus to unarmed melle attacks?

Adding a 2 strength bonus should cover some of that and fit the rest of the tree evolution

BlizzDA:
I can see the reasoning for adding it as prereq for LARGE, and adding it to Beast as that's a general mammal trait line, but Alpha doesn't seem very fitting. That's more Xedra's foray into Human+, perfecting what's already there, rather than creating a Spartan-like supersoldier. Hence the Prime descriptions. (Plus Alpha's design being stat boosts across the board) It may fit better in fit a eventual new Prime + X mutagenic line, like how Elf-A is already a mix.

LyleSY (approved):

PR #63776, opened 2023-02-24
make most anatomy mutations neutral

anoobindisguise:
#### Summary
Balance "most anatomy mutations are 0 points"

#### Purpose of change
Under our current system I believe that "negative" mutations should exclusively be ones that are completely bad with no redeeming value such as frail, pain sensitive, nomad, very thirsty, bad back, etc. Ones that are currently labeled as "bad" but have a clear upside (Hooves etc) should be neutral, so they are possible to obtain regardless of instability.
From a lore perspective it's OK for mutations to start becoming malignant if you mutate too much, but it's weird that the mutagen arbitrarily decides that some mutations are "bad" and others are "good" (for example, paws are "bad" but bird talons are "good" despite having a similar consequence for your character, or how a fluffy tail is "good" but a fish tail is "bad". Compare this to bad back and strong back, one is clearly entirely detrimental and one is clearly entirely beneficial)

#### Describe the solution
Make the vast majority of previously-negative mutations that have an upside 0-cost.

This will increase the number of zero-cost mutations by a lot so it might be worth bumping the odds of a neutral mutation from 10% to 20%.


data/json/mutations/mutations.json
```diff
@@ -4008,7 +4008,7 @@
     "type": "mutation",
     "id": "HORNS_CURLED",
     "name": { "str": "Curled Horns" },
-    "points": -3,
+    "points": 0,
```
Maleclypse, line 4011: horns and curled horns both need to restrict gear on the head

anoobindisguise: Originally it was regular horns allow all headgear, curled horns allow no headgear and pointed horns/antennae/antlers allow soft headgear only, but I went ahead and made regular horns soft headgear only.

alp7292:
just by quick look i didnt looked all of them but
bird legs restrict gear
catte tail restricts gear
fat slows you down
tiny reduces your health
butterfly wings restrict gear and reduces dodge
paws and broad paws are terrible they give encumbrance crafting malus and  restrict gear
beaks restrict gear
these are negative with ranging from little to no benefit mutations i dont think these are fit to be "neutral traits" as you gain these even when you dont have genetic damage

Rabidweasel:
I like this idea in theory but some of the anatomy mutations you've mentioned are really, really bad (the various "paws" mutations are one of the worst, IMO) to the point where the logical conclusion from this is to just not mutate certain categories. These mutations were clearly designed to be negatives, hence their existing negative points cost.

I think that making these iconic mutations into a more integral part of each mutation category is a fantastic idea - being forced to use different gear due to slot restrictions is interesting even if many players try to avoid it - but the effect of the mutations should be reviewed to ensure that there's a more reasonable balance of positive and negative for each before categorising them as neutral.

alp7292:
a lot of mutations are added by different people with different time thus a lot of mutation effects and its points are inconsistent witch each other a better fix would be just checking and balancing every mutation in one go using some guideline to be consistent and well balanced but that requires a lot of time and patience and requires discussing as some people find changes good and balanced and some may not.

actually-a-cat:
> This will increase the number of zero-cost mutations by a lot so it might be worth bumping the odds of a neutral mutation from 10% to 20%.

fyi there is no longer a fixed chance for neutral mutations, see #63793, so this shouldn't be an issue.

Raikiri:
Characters that want to use normal armor are already limited to only a few trees that don't restrict their armor use, treating other trees as non-existent. In my opinion, this would just worsen this situation.

Venera3:
The munchkin perspective is not really relevant to this - mutations are supposed to be kinda unpredictable rolls on the eldritch slot machine living in your cells. 
In that reading, these counting as neutral mutations makes sense, since *the hooves* work just fine! That the ungrateful survivor attached to the hooves is pissy they don't get to wear their nice flip-flops anymore is an externality as far as the mutagen is concerned. Net negative traits tend to be the ones implying loss-of-function (or gain-of-cancer) changes that are objectively screwing you up.

Raikiri:
> The munchkin perspective is not really relevant to this - mutations are supposed to be kinda unpredictable rolls on the eldritch slot machine living in your cells.
The problem is that if it actually was the case, nobody would actually want to engage with this system except for RP or challenge reasons. If as you said you actually had a chance of getting an unpredictable mutation that you actually didn't want. Like right now you're describing how sleeping in a nuclear reactor works, or getting yugg'd or getting genetic chaos -- and there are not many volunteers to use those mechanics.

The reason people engage with the mutation system is because they're considering the upsides that it gives to be greater than the downsides. And if you say that from now on bestial mutations will have even more downsides for most people, they will just engage even less with those trees. People will just keep exploiting the "lazy" post-human trees.

Venera3:
Again, the *balance* of the distinct trees is not really a concern - Snail exists, for a reference. They should be internally consistent, not balanced against each other (or even against a baseline human with an M4, which is a mark you won't really reach with any of our mutations).

Raikiri:
> Again, the _balance_ of the distinct trees is not really a concern - Snail exists, for a reference. They should be internally consistent, not balanced against each other (or even against a baseline human with an M4, which is a mark you won't really reach with any of our mutations).

Okay, curious. So what's the reason why a baseline human with an m4 would ever want to take a snail mutagen if they're literally better off not doing so? And if the answer is "there's no reason", then what's the point of developing this deeply elaborate system that people don't have a reason to interact with?

anoobindisguise:
So one issue here is "how does the mutagen Know"? Like I'm not sure how the mutagen decides to classify hooves as being "bad" but spider mandibles as being "good" when it's an unthinking process. It doesn't have the luxury of a player to decide "oh, well Fluffy Tail is good enough to lose access to metal pants, but Cow Tail isn't good enough to be forced to wear fabric pants".

Raikiri:
By the way, the reason why it makes sense to give preference "good" mutations instead of "bad" ones, where "good" and "bad" come entirely from a human's perspective is that mutagens are synthesized while having certain goals in mind: their creators engineered it so that unwanted phenotypes are suppressed.

In this way of thinking that's the whole difference between getting yugg'd and taking a mutagen: the former just randomly alters your phenotype, the latter is engineered to be beneficial from the perspective of its designers. The reason that you can get unwanted effects at all might be because you're not following the protocol correctly and are not giving your body enough time to adapt to the mutagen, hence the whole instability mechanic.

anoobindisguise:
So, it's worth noting that for most of the trees that had Paws, Paws being negative didn't really do anything because you were going to get them as a prereq of Burrowing anyways. There are a lot of trees that give you "negatives" like that for a later post thresh positive.

Venera3:
> The reason that you can get unwanted effects at all might be because you're not following the protocol correctly and are not giving your body enough time to adapt to the mutagen, hence the whole instability mechanic.

No, mutagens are a very experimental, not at all understood (or, really, understand*able*) thing XEDRA was just beginning basic research on. 

> Okay, curious. So what's the reason why a baseline human with an m4 would ever want to take a snail mutagen if they're literally better off not doing so? 

Because you potentially get capabilities you otherwise wouldn't? In some cases, those capabilites will not necessarily be worth what you got in return, but that's just how the cookie crumbles. For the record, I'm interested in making mutations generally *better* in their niche, but removing the element of uncertainty of it is not the way to go - especially since the *intended* end point of every system is to be learnable in-game, not theorycraft your exact mutation path with external resources.

PR #63795, opened 2023-02-25
Mana softening

adamkad1:
#### Summary
Mods "Makes mana soft"

#### Purpose of change

Makes concentrated mana a soft material to unrigidify certain things. Stone gauntlets explicitly say they are flexible, and stormfist... how can lightning be rigid anyway? Bonus points for the fact that this change will make magic actually be able to put those things over rigid equipment instead of summoning to your hands because you are allready wearing armguards or something.  Kinda wish we had different ways to make things rigid/not rigid than 'Is the material soft?'

#### Describe the solution

To make magic arm things non rigid, mana has been made a soft material

None. If anyone else got better idea, you are welcome to replace this.

#### Testing

Do this, check in game, magic stuff arent rigid anymore

#### Additional context

Nothing here

### DIFF
--- a/data/mods/Magiclysm/materials.json
+++ b/data/mods/Magiclysm/materials.json
@@ -271,6 +271,7 @@
     "name": "Concentrated Mana",
     "density": 2,
     "reinforces": false,
+    "soft": true,
     "bash_resist": 8,
     "cut_resist": 13,
     "bullet_resist": 7,


GuardianDll:
Does it cause any problems with armor or stuff?

adamkad1:
> Does it cause any problems with armor or stuff?

It makes things comfortable but as far as I know, nothing else changes.

GuardianDll (approved):
Then i see no problem to make it soft

adamkad1:
> Then i see no problem to make it soft

If anything, someone who knows what they would be looking for could check it, but I didnt notice any stat changes

GuardianDll:
image
image

Now someone need to add length to all items that uses concentrated mana as material 😆

adamkad1:
> image image
> 
> Now someone need to add length to all items that uses concentrated mana as material 😆

XD putting your magic hammer in a bottle is the funniest shiz

PR #63824, opened 2023-02-27
Make adhesive bandages recipe craft multiples.

TaiGambol:
#### Summary
None

#### Purpose of change
Crafting adhesive bandages from medical gauze causes >95% of the gauze to be wasted. This changes things so that only ~33% is wasted. By contrast, crafting bandages from medical gauze wastes ~22% of the material, so for multiple smaller cuts, ~33% seems reasonable wastage.

#### Describe the solution
Edited the adhesive bandage crafting recipe to craft 30 bandages instead of 1.

Changing the recipe to use something other than medical gauze (perhaps a cotton patch? That would make these a field-craftable dressing, now that makeshift bandages are more difficult to craft on the spot)

#### Testing
Created new character with Health Care 2.
Checked crafting screen, found Adhesive Bandage recipe.
Recipe indicates 30 will be created.
Grabbed first aid kit from evac shelter basement.
Crafted adhesive bandages.
30 were crafted.

### DIFF
--- a/data/json/recipes/other/medical.json
+++ b/data/json/recipes/other/medical.json
@@ -70,9 +70,8 @@
     "subcategory": "CSC_*_NESTED",
     "skill_used": "firstaid",
     "difficulty": 1,
-    "time": "1 m",
+    "time": "20 m",
     "charges": 30,
-    "batch_time_factors": [ 50, 2 ],
     "autolearn": true,
     "qualities": [ { "id": "CUT", "level": 2 } ],
     "components": [ [ [ "medical_gauze", 1 ] ], [ [ "duct_tape", 8 ], [ "medical_tape", 10 ] ] ]
--- a/data/json/recipes/other/medical.json
+++ b/data/json/recipes/other/medical.json
@@ -71,7 +71,7 @@
     "skill_used": "firstaid",
     "difficulty": 1,
     "time": "1 m",
-    "charges": 30
+    "charges": 30,
     "batch_time_factors": [ 50, 2 ],
     "autolearn": true,
     "qualities": [ { "id": "CUT", "level": 2 } ],
--- a/data/json/recipes/other/medical.json
+++ b/data/json/recipes/other/medical.json
@@ -75,7 +75,7 @@
     "batch_time_factors": [ 50, 2 ],
     "autolearn": true,
     "qualities": [ { "id": "CUT", "level": 2 } ],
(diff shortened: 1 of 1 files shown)


Kamayana:
I rather like the idea of being able to make an adhesive bandage with a cotton patch, but that would be in addition to this PR rather than instead of, this making a bunch of them with gauze and tape also makes sense.

TaiGambol:
Not sure what failed, exactly? It looked like something to do with the mutation code, but I didn't touch that...

bombasticSlacks:
You probably need more duct tape/medical tape to make 30 bandages, though otherwise the rationale makes sense.

bombasticSlacks:
Will need to be restarted after #63856

TaiGambol:
How do I restart?

bombasticSlacks:
I'll get it

cake-pie:
Heads up: #63434 changes adhesive bandages from comestible to generic; it will no longer have charges. In case that PR is merged first, change `charges` to `result_mult`.

data/json/recipes/other/medical.json
```diff
@@ -71,10 +71,11 @@
     "skill_used": "firstaid",
     "difficulty": 1,
     "time": "1 m",
+    "charges": 30,
     "batch_time_factors": [ 50, 2 ],
```
cake-pie, line 75: Should probably adjust time and batch time factors as well

data/json/recipes/other/medical.json
```diff
@@ -71,10 +71,11 @@
     "skill_used": "firstaid",
     "difficulty": 1,
     "time": "1 m",
```
bombasticSlacks, line 73:
```suggestion
    "time": "20 m",
```

data/json/recipes/other/medical.json
bombasticSlacks, line 75:
```suggestion
```

bombasticSlacks (commented):
Going to increase time and merge it so that this can get in

data/json/recipes/other/medical.json
```diff
@@ -70,11 +70,11 @@
     "subcategory": "CSC_*_NESTED",
     "skill_used": "firstaid",
     "difficulty": 1,
-    "time": "1 m",
-    "batch_time_factors": [ 50, 2 ],
+    "time": "20 m",
+    "charges": 30,
```
bombasticSlacks, line 74:
```suggestion
    "result_mult": 30,
```

PR #63821, opened 2023-02-27
M1 mac fixes

LyleSY:
#### Summary
Build "Restore M1 build"

#### Purpose of change

Restores M1 build from #63702 with typo fixed

#### Describe the solution

Restores M1 build from #63702 and applies typo fix suggested by [kebkaldanil](https://github.com/kebkaldanil)

#### Describe alternatives you've considered

Remove homebrew environment variables entirely because I think they're redundant, remove new .sh script that isn't working

#### Testing

Appears to work and passes checks

#### Additional context

Part of push to get the new stable build working

PR #63825, opened 2023-02-27
Support Mac arm64

CzBiX:
#### Summary
Build "Support Mac arm64 build"

#### Purpose of change

Get native arm64 build for mac.

#### Describe the solution

Use framework for mac to avoid cross compile.

#### Additional context

reopen #63702 to fix typo


Maleclypse:
Is this fix covered by or in addition to #63821?

CzBiX:
@Maleclypse seems the same, but with clean commit messages.

LyleSY:
Isn't this missing the script and all of the other good things in the original that were reverted?

kevingranade:
You need to cherry pick the original commits or revert the revert.

PR #63832, opened 2023-02-28
Revert "Revert "Support Mac arm64""

kevingranade:
Reverts CleverRaven/Cataclysm-DDA#63797
Unverts CleverRaven/Cataclysm-DDA#63702

Applies fix from CleverRaven/Cataclysm-DDA#63825

#### Summary
None

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -188,6 +188,15 @@ jobs:
             artifact: osx-tiles-x64
             ext: dmg
             content: application/x-apple-diskimage
+          - name: OSX Tiles arm64
+            os: macos-11
+            arch: arm64
+            mxe: none
+            tiles: 1
+            sound: 0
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -262,8 +271,12 @@ jobs:
       - name: Install dependencies (mac)
         if: runner.os == 'macOS'
         run: |
-          HOMEBREW_NO_AUTO_UPDATE=yes HOMEBREW_NO_INSTALL_CLEANUP=yes brew install sdl2 sdl2_image sdl2_ttf sdl2_mixer gettext ccache parallel
+          brew install gettext ccache parallel
+          ~/build-script/get-mac-sdl2-framework.sh
```
kevingranade, line 275:
```suggestion
          ~/build-scripts/get-mac-sdl2-framework.sh
```

PR #63840, opened 2023-02-28
No more spiders in subways

MNG-cataclysm:
#### Summary
Balance "Remove spiders from subway systems."

#### Purpose of change
Spiders don't live in subways, as their prey doesn't live there. As such, they should be relocated to somewhere else.

#### Describe the solution
Remove spider spawn groups from the subway spawn groups.

Keeping spiders in subways, or adding prey insects to subways.

#### Testing
Went to a subway, looked around a bit, and didn't find any spiders.

#### Additional context
None.

### DIFF
--- a/data/json/monstergroups/misc.json
+++ b/data/json/monstergroups/misc.json
@@ -103,15 +103,6 @@
         "pack_size": [ 1, 2 ],
         "starts": "180 hours"
       },
-      { "monster": "mon_spider_cellar_small", "weight": 300, "pack_size": [ 1, 3 ], "ends": "14 days" },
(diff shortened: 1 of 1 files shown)


Karol1223:
Cellar spiders specifically like areas like this, please read up on the animals you're affecting before going through with the changes.

> These webs are constructed in dark and damp recesses such as in caves, under rocks and loose bark, and in abandoned mammal burrows. In areas of human habitation pholcids construct webs in undisturbed areas in buildings such as high corners, attics and cellars, hence the common name "cellar spider".

Literally from the first Wikipedia article.

PR #63849, opened 2023-03-01
prevent stored cal integer overflow

### LINKED ISSUE #63727: Instant death by integer overflow if stored kcal > ~2,100,000

Currently the integer limit to stored calories caps out around this number and causes it to become massively negative if you exceed it, killing you instantly. At that point you would be about 250 kg overweight which is dangerously high although probably not enough to kill you of a heart attack outright. (Note that the game thinks you are over BMI 400 because it calculates BMI incorrectly, you are in reality closer to BMI 100 - see #60597). Regardless of whether that's way more than anyone should ever achieve in the apocalypse it is clearly a bug as an integer overflow is not normal behavior.

### Solution you would like.

There are four solutions I can think of:
a) warn the player if they are above 2 million stored kcal and then cause them to die of a "scripted" heart attack if kcal > ~2,100,000
b) raise the limit on stored kcal somehow (probably not possible without breaking a save)
c) hardcap stored kcal at ~2,100,000 and don't allow it to become bigger than that
d) ignore this "problem" because it is unlikely to matter in normal gameplay.

see above

### Additional context

_No response_

anoobindisguise:
#### Summary
Bugfixes "stored calories cap instead of integer overflowing"

#### Purpose of change
Fixes #63727

#### Describe the solution
Simply limit max stored calories in the function that adjusts stored calories.
Should go along just fine with #60597.

#### Testing
Compiled and tested this by setting stored kcal to 2 million and guts kcal to 2 million and didn't see stored kcal exceed ~2.14 million after waiting over 24 hours repeatedly refreshing guts kcal to 2 million whenever it fell below 1 million kcal.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -4046,7 +4046,7 @@ void Character::mod_stored_calories( int ncal, const bool ignore_weariness )
     if( !ignore_weariness ) {
         activity_history.calorie_adjust( ncal );
     }
-    if( ncal < 0 || ncal + stored_calories > 0 ) {
+    if( ncal < 0 || std::numeric_limits<int>::max() - ncal > stored_calories ) {
         set_stored_calories( stored_calories + ncal );
     }
 }
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -4046,7 +4046,9 @@ void Character::mod_stored_calories( int ncal, const bool ignore_weariness )
     if( !ignore_weariness ) {
         activity_history.calorie_adjust( ncal );
     }
-    set_stored_calories( std::min( 2147000000, std::abs( stored_calories + ncal ) ) );
+    if( ncal < 0 || ncal + stored_calories > 0 ) {
+        set_stored_calories( stored_calories + ncal );
+    }
 }
 
 void Character::set_stored_kcal( int kcal )
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -4046,7 +4046,7 @@ void Character::mod_stored_calories( int ncal, const bool ignore_weariness )
     if( !ignore_weariness ) {
         activity_history.calorie_adjust( ncal );
     }
-    set_stored_calories( stored_calories + ncal );
+    set_stored_calories( std::min( 2147483645, stored_calories + ncal ) );
 }
 
 void Character::set_stored_kcal( int kcal )
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -4046,7 +4046,7 @@ void Character::mod_stored_calories( int ncal, const bool ignore_weariness )
     if( !ignore_weariness ) {
         activity_history.calorie_adjust( ncal );
     }
-    set_stored_calories( std::min( 2147483645, stored_calories + ncal ) );
+    set_stored_calories( std::min( 2147000000, std::abs( stored_calories + ncal ) ) );
 }
 
 void Character::set_stored_kcal( int kcal )


mqrause:
You need to check if `stored_calories + ncal` is safe to do without overflow. E.g. check if `INT_MAX - stored_calories > ncal`, otherwise just pass `INT_MAX` along.

A simple test case would be nice to have, too.

anoobindisguise:
> You need to check if...
shouldn't the std::abs do that anyways? if it overflows it will be -2147483647 which will correct it to a positive value, which is larger than the 2147000000 and thus min will reduce it to 2147000000. right?

mqrause:
There's one more negative number because positive numbers start at 0, so you could get undefined behavior. You could also end up with less calories than before if you do it this way, which might not be the biggest concern, but it's weird behavior anyway.

Qrox:
> if it overflows it will be -2147483647

Signed overflow is actually undefined by the standard, so it's best to check if it would overflow here.

src/character.cpp
```diff
@@ -4046,7 +4046,9 @@ void Character::mod_stored_calories( int ncal, const bool ignore_weariness )
     if( !ignore_weariness ) {
         activity_history.calorie_adjust( ncal );
     }
-    set_stored_calories( stored_calories + ncal );
+    if( ncal < 0 || ncal + stored_calories > 0 ) {
```
Qrox, line 4049: Integer overflow is undefined so even if `ncal + stored_caloreis > 0` it could still have overflowed (for example if the target system uses saturation instead of wrap around for overflow).

src/character.cpp
```diff
@@ -4046,7 +4046,9 @@ void Character::mod_stored_calories( int ncal, const bool ignore_weariness )
...
+        set_stored_calories( stored_calories + ncal );
+    }
```
Qrox, line 4051:
```suggestion
    static_assert( std::is_same<decltype( ncal ), decltype( stored_calories )>::value, "Expected the same type for ncal and stored_calories" );
    const decltype( ncal ) cal_max = std::numeric_limits<decltype( ncal )>::highest();
    const decltype( ncal ) cal_min = std::numeric_limits<decltype( ncal )>::lowest();
    if( ( ncal <= 0 && stored_calories >= 0 ) || ( ncal >= 0 && stored_calories <= 0 ) ) {
        // non-negative + non-positive must be no less than the lowest negative and no more than the highest positive and therefore never overflows
        set_stored_calories( stored_calories + ncal );
    } else if( ncal > 0 ) {
        // therefore stored_calories > 0
        if( cal_max - stored_calories >= ncal ) {
            set_stored_calories( stored_calories + ncal );
        } else {
            set_stored_calories( cal_max );
        }
    } else {
        // ncal < 0 and stored_calories < 0
        if( cal_min - stored_calories <= ncal ) {
            set_stored_calories( stored_calories + ncal );
        } else {
            set_stored_calories( cal_min );
        }
    }
```
This might look intimidating but I think it is the safest approach. I haven't compiled and tested it though.

anoobindisguise: Alright, I'll hold on for a senior dev on this.

mqrause: I think adding negatives isn't really relevant here since `stored_calories <= 0` means starving to death, so doing all of this might be a bit overkill. Although it might be useful to make it into a general "safe addition" helper function. I'm sure we have other places in the code with custom overflow guards.

irwiss:
I'll allow myself to voice an opinion as a normal contributor;
I feel this is over the top for the job it's supposed to be doing, in cata codebase there's tons of places where overflows may occur, in none of them we are doing even the trivial checks, it's also a non trivial issue for C++ newcomers to grasp. Some alternatives I can think of;
Use a uint32_t - makes it non-UB to overflow, keep same memory footprint but overflow will still kill the character, just at ~4b calories instead with the extra bit
Use a int64_t: give up 4 bytes per Character instance, but: over/underflow not practically reachable, calorie code doesn't seem to be on any hot paths, trivial for contributors
Make it a mix; widen to int64_t, do the addition/subtraction, check for overflow and clamp, cast back to int32 to store, still saves the 4 bytes per Character, which is overall worst of the bunch but still fairly trivial for contributors

src/character.cpp
kevingranade, line 4049: This overflow proofs it.
```suggestion
    if( ncal < 0 || std::numeric_limits<int>::max() - ncal > stored_calories ) {
```

src/character.cpp
kevingranade: This might be appropriate for a generic helper for "does_sum_overflow()", but is massively invertebrates for inlining.

PR #63853, opened 2023-03-01
Tune down trapdoor spider spawns

### LINKED ISSUE #61017: Are trapdoor spiders intended to migrate all over the place?
### Describe the bug

When I started the game I found trapdoor spider traps in a bunch of trees, and stayed well clear of those. However, after 18 days these buggers have dug mounds all over the place without actually placing any new trapdoors anywhere.

### Steps to reproduce

1. Start a [Innawood] game with trapdoor spiders somewhat near.
2. Detect the traps and stay well away from them and any spiders that may be visible.
3. Wait.
4. Found that the buggers have tilled themselves over several world tiles and threatening to surround or overwhelm you.

### Expected behavior

Trapdoor spiders should stay close to their doors, only leaving to chase prey, and then return to their doors, rather than chase animals all over the place (a guess: I've seen some animal corpses, but only near to their original location) and then stay there to migrate further.

### Screenshots

The spiders started in the wooded tile at the bottom right corner of the screenshot.

Screenshot (260)

- OS: Windows
    - OS Version: 10.0.19043.1889 (21H1)
- Game Version: 4f45027 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Innawood [innawood]
]

### Additional context

I sort of verified that the spiders were responsible by saving the game and then walk around the mounds (still well away from the original location), and had the companion grabbed by a spider. As expected, fighting it would result in death unless managing to run away (or save scum) when having neither armor or weapons (or skill).

On the other hand, there's another trapdoor spider group a bit further away that I've passed by somewhat frequently, but with no indication of mound generation (or corpses).

There's also a nearby mound filled area I've avoided because of the worm indication, but saving, walking around there, and save scumming has not given any indication of what caused them. I don't see those mounds spreading either.

Edit: Gave up the save after seeing a spider from the crafting spot. Getting overrun happened faster than expected...

Venera3:
#### Summary
Bugfixes "Adjust trapdoor spider dens"

#### Purpose of change

Trapdoor spider dens spawn two dozen giant trapdoor spiders who make a mess of your lawn and/or eat Day1 survivors for breakfast. This gets them down to smalled bunches of small spawns (at least at first), with some gibs strewn around for good measure to warn you.

#### Describe the solution

Reduced and randomized the trapdoor trap spawns, changed the spawn range and the spawned monster to the small version of the spooders - since they evolve instead of growing up, they will get bigger with time.

Rework them properly, but doing ambush monsters *right* would need a lot of infra we don't really have, and would definitely take more time than what I want to spend for the moment. One of these days.

#### Testing

Spawned the map extra prior to the changes, got torn apart.
Spawned the map extra after the changes, saw some mounds and bug bits lying suspiciously around, got torn apart slightly less.

#### Additional context

Fixes #61017


CMakePresets.json
```diff
@@ -14,41 +14,26 @@
       "condition": {
         "type": "allOf",
         "conditions": [
-          {
```
Maleclypse, line 17: Your gonna want to remove your linting of the cmake presets

org.cataclysmdda.CataclysmDDA.json
```diff
@@ -16,22 +16,10 @@
       "name": "cataclysm-tiles",
       "buildsystem": "simple",
       "build-options": {
-        "env": {
-          "MAKE_ARGS": "PREFIX=/app LANGUAGES=all USE_XDG_DIR=1 TILES=1 SOUND=1 RELEASE=1 RUNTESTS=0 ASTYLE=0 LINTJSON=0"
-        }
+        "env": { "MAKE_ARGS": "PREFIX=/app LANGUAGES=all USE_XDG_DIR=1 TILES=1 SOUND=1 RELEASE=1 RUNTESTS=0 ASTYLE=0 LINTJSON=0" }
```
Maleclypse, line 19: Also I don't believe you intended linting here either.

PatrikLundell:
It doesn't fix the problem, but it would reduce the consequences a bit. As mentioned, fixing it really would require a rework of ambush predators.

Is there a way to change the trap effect over time? As far as I understand the triggered trap would summon the small version regardless when it's triggered, while it really ought to summon small ones early on and larger ones later so it didn't matter whether the trap was triggered on day one and the spider remained in that position due to a lack of prey to pursue, or the trap was just triggered at the one year mark: you ideally would face the same type of spider in both cases.

Venera3:
>  since they evolve instead of growing up, they will get bigger with time

The small spiders will roll to evolve on spawn, just like every other `half_life` mob.

org.cataclysmdda.CataclysmDDA.json
Venera3: Good catch, thanks!

PR #63862, opened 2023-03-02
fix release version numbers

kevingranade:
#### Summary
 None

#### Purpose of change
The existing tag based versioning system broke at some point, so the version of the 0.G executables is just the commit hash. This PR does some quiche and dirty changes to produce the correct version number.


rumly111:
I downloaded 0.G release, and the game extracted to `cataclysmdda-0.F` directory . Is it fixed now?

PR #63868, opened 2023-03-02
remove hypervitaminosis reference from iron vitamin

nornagon:
The excess effect was never applied since there is no disease_excess threshold list.

#### Summary
None

#### Purpose of change

This seems to have been left over from earlier vitamin refactors.

#### Describe the solution

Given no other vitamin has the hypervitaminosis effect, I assume that the design intent here is to remove hypervitaminosis from the game (at least for now). Since this effect is never applied, I just removed the reference for now, and left the hypervitaminosis effect definition itself in place.


bombasticSlacks:
needs a rebase to work

Omnikron13:
On a pedantry note, I'm pretty sure that iron hypervitaminosis isn't a thing, either, on account of iron being rather more of an element/heavy metal than a vitamin. Iron toxicity would be a slightly odd thing to need to model for this kind of game too, imo, as I've only ever heard of it in relation to combining certain medications over a period of time? Perhaps there's more about than I'm aware of. *shrug*

(on a tangent, Hypervitaminosis A would make a lot of sense; more than one unfortunate (ant)arctic explorer have supposedly met a very unpleasant fate from mistakes such as eating the livers of their pack dogs when they ended up in a survival situation... I don't know that it has much in common with iron toxicity though.)

PR #63870, opened 2023-03-02
Fursuit rework

KittyTac:
#### Summary
Content "Separate fursuits into components, add fox fursuit"

#### Purpose of change
CDDA fursuits did not have gloves and boots unlike real fursuits, and had the head unable to be separated. This PR aims to fix that.

#### Describe the solution
I split every fursuit into a body, a head, a pair of gloves, and a pair of boots. They should spawn together. If I missed something, tell me please. Also I added a fox fursuit.

Leaving as is, adding partial fursuits.

#### Testing
image

I wasn't able to find any "in the wild" and I'm not sure how to test itemgroups.

#### Additional context
I plan on adding more types. Also, there should be sprites added for them.


Karol1223:
Normally I'd suggest turning them into variants of one 'fursuit' item but to be honest, this may not play nicely with splitting them into parts. Someone more experienced than me should probably comment on whether its viable to split them into variants and yet spawn a full set of them together as opposed to pure RNG madness.

KittyTac:
> Normally I'd suggest turning them into variants of one 'fursuit' item but to be honest, this may not play nicely with splitting them into parts. Someone more experienced than me should probably comment on whether its viable to split them into variants and yet spawn a full set of them together as opposed to pure RNG madness.

Yeah that. Also some types have different stats.

KittyTac:
I think I accidentally marked as resolved things that are unresolved.

KittyTac:
I'll finish this later.

KittyTac:
I'm a moron okay?

PR #63873, opened 2023-03-02
item_location: add a make_active() helper

### LINKED ISSUE #63894: Crash when inserting items into some worn items
### Describe the bug

Game crashed when I tried to insert makeshift wooden arrows into a worn large quiver. Total vanilla game without mods, clean install of 0.G.

### Attach save file

N/A
A similar thing happened earlier in the same game when i tried to insert something into a leather belt, but i can't reproduce it consistently, so the save game won't help, i think? There seems to be some other variable that I am not aware of yet, so the first time i ignored it, but now it happened again with the quiver, so I thought i'd create an issue at least so you can look into it.

### Steps to reproduce

1. Wear an item
2. Try to insert something via the item menu (enter + i)

### Expected behavior

The game not randomly crashing when using the insert function.

### Screenshots

_No response_

OS: Windows 10
Version: 0.G dafb974 (2023-03-02-1808)
Graphics: Tiles
Language: English
Mods: None/Standard

### Additional context

Both crashes prompted the same message: SIGABRT: Abnormal termination
[crashlog.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10877691/crashlog.zip)

### EDIT:
This also seems to happen when activating worn items like in the case of a scabbard or a belt, not using the insert function.

andrei8l:
#### Summary
None

#### Purpose of change
#62547 broke the `insert_item_activity_actor` by assuming the holster is always on the map
* Fixes: #63894

#### Describe the solution
Add a `make_active()` helper function to `item_location` that handles the low level bits.

N/A

#### Testing
Using the inventory UI and the AIM, insert an active grenade into a container on your avatar, on the map, and in a vehicle. Verify that the game doesn't crash and the grenade explodes after a few seconds.

Check again with nested containers.

#### Additional context
N/A

### DIFF
--- a/src/item_location.cpp
+++ b/src/item_location.cpp
@@ -543,6 +543,10 @@ class item_location::impl::item_on_vehicle : public item_location::impl
             cur.veh.invalidate_mass();
         }
 
+        void make_active( item_location &head ) {
+            cur.veh.make_active( head );
+        }
+
         units::volume volume_capacity() const override {
             return cur.veh.free_volume( cur.part );
         }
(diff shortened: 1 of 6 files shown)


Kamayana:
Oh hey, looks like we were both working on a fix for it. Here's my commit for reference: 00d6badd3e77544284c8aa91faf18997744bce97, though I prefer your solution; giving item_location a make_active function is a good call.

src/item_location.cpp
```diff
@@ -934,6 +938,31 @@ void item_location::on_contents_changed()
...
+        case type::container: {
+            parent_item().make_active();
+            break;
+        }
+        case type::map: {
+            get_map().make_active( *this );
+            break;
+        }
+        case type::vehicle: {
+            dynamic_cast<impl::item_on_vehicle *>( ptr.get() )->make_active( *this );
```
ZhilkinSerg, line 957: It looks like it is missing `break;` here.

PR #63876, opened 2023-03-02
Painted walls addition and tweaks

gettingusedto:
#### Summary
Content "More colour for the painted walls and tweaks to pre-existing painted walls."

#### Purpose of change

Adds more colour types to the walls and some tweaks for the existing painted wall constructions.

#### Describe the solution

1. Add more colour type to the existing painted walls due to #62957 .
2. Tweak the amount of paint used in pre existing painted walls constructions.

Not doing so.

#### Testing
Made a mod for it, i tested it in my world. the wall appears in game and can be painted onto wall, wood walls, and concrete walls as expected.

#### Additional context

Thank you @Karol1223 for the paint pr.


jbytheway:
If I understand correctly that this is adding more 'regular' walls, someone might want to add these new wall types to the options that are chosen from for house mapgen.  The distribution is [here](https://github.com/CleverRaven/Cataclysm-DDA/blob/a52f23da779f4f9edef8596594bff882dcd8d67a/data/json/mapgen_palettes/house_general_palette.json#L11-L17).

PR #63878, opened 2023-03-02
Mutant Cats & Dogs: A Tale of Nether & Confusion

Karol1223:
#### Summary
None

#### Purpose of change
As discussed on the devcord:
Our mutants primarily operate in the realm of possible but EXTREMELY unlikely changes in our real world. A deer crossed with a spider the way our spideer functions could theoretically operate with sufficient modifications. So could, theoretically, a three-headed bear and most of our giant insects (I'm looking at you Frigate Fly). And then we had cats and dogs. Who spontaneously decided that, actually, they're gonna go full SCP. No matter how hard you try you will not biologically explain an antimemetic dog. Just no. This PR instead makes them 'nether-touched' animals - animals who by one mean or another got in contact with the nether and changed. I'll let you figure out the details.

#### Describe the solution
I really wasn't sure how to go about defining this change...

- For starters, the dogs and cats have been moved to their own file. Because contributors getting confused is a big problem and if you don't believe this you have not been in all the annoying conversations on the devcord.
- These nether-touched creatures can no longer reproduce. The babies have been obsoleted. This makes them more rare and unique when found in the wild - having them reproduce took that factor away
- The cats no longer have useless code they're already copying via ``copy_from``. Dogs have been given ``copy_from`` from the labrador because they copied over 99% of the lines anyway. All this does is gives then 5 more aggression (again, to match the labradors)
- The dogs no longer give mutant meat. They're not mutants. Nether creatures should probably have their own unique factors for butchery but it's out of the scope of this PR
- The liquid cat has been made ``phase: LIQUID`` and no longer produces a corpse - instead it ``splashes one last time``. I DARE you to butcher a liquid
- The spawnrates for mutant dogs have been reworked. Overall the trio is now way rarer, mongrels are slightly (about 0.3%) more common but this change shouldn't be visible. Cats remain largely unchanged in this department.

Initially I considered making them connected to portal storms and not the nether but I don't really know how portal storms would affect animals. I also wanted to make them more discernible as not being mutants but I don't have a lot of ideas. People also suggested new nether-touched animals - again, out of scope for this and out of the scope of my interest. I just want to keep mutants make sense and this is what this PR does. What ya'll do with the concept of nether-touched animals is beyond my interest.


NetSysFire:
>The liquid cat has been made phase: LIQUID and no longer produces a corpse - instead it splashes one last time. I DARE you to butcher a liquid

good one!

data/json/monsters/nether_animal.json
```diff
@@ -0,0 +1,54 @@
...
+    "//": "Higher dodge due to the constantly changing fur pattern",
+    "dodge": 9
+  },
+  {
+    "id": "mon_cat_liquid",
+    "type": "MONSTER",
+    "copy-from": "mon_cat",
+    "name": { "str": "liquid cat" },
+    "description": "It is universally known that cats are liquid, but this cat's body is actually quivering while it walks around.  It also enjoys resting while twisting its body in ways that would break a normal cat's bones.",
+	"phase": "LIQUID",
```
NetSysFire, line 17: cant really make a multiline suggestion but this and the next line needs to be indented.

NetSysFire (commented):
I am sad to see the kittens go. Cant we make the reproducing much, much slower instead of non-existent?

data/json/monsters/nether_animal.json
Karol1223: Whoops! I forgot to lint! Thanks for catching this!

Karol1223:
> I am sad to see the kittens go. Cant we make the reproducing much, much slower instead of non-existent?

Personally I just think it doesn't feel the vibe of these animals. If the nether is to warp what it touches I think it's only normal for creatures affected by it to be sterile. I'll be honest if we had a way for the game to check if there are already two adults around to breed I'd have left it but the way our mitosis works doesn't feel like the creature is anything special.

Daved27hundred:
> I think the iridescent cat was meant to be based on something like this, rather than being a full on blob-mutant: https://www.bbc.com/news/science-environment-14882008

I think Karol said earlier today she wants to avoid making mutants that are loosely based on irl things.

Karol1223:
Mutants based on *existing* irl things are fine, as long as they are actually the same thing as the existing irl thing. What we had here was a cat that loosely referenced some existing irl thing, which didn't make it make any more sense. What it's based on doesn't matter. The execution is what does.

data/json/monsters/nether_animal.json
```diff
@@ -0,0 +1,54 @@
...
+	"harvest": "exempt",
```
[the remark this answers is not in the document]
Karol1223: You're LATE

bombasticSlacks:
In this case, it's fine since you've moved everything to copy from's anyway, so they are basically new entries and the obsoleting movement is always okay, but please avoid moving JSON entries to new files. It makes it difficult to track history and there is no way to enforce the organization long term.

Karol1223:
> In this case, it's fine since you've moved everything to copy from's anyway, so they are basically new entries and the obsoleting movement is always okay, but please avoid moving JSON entries to new files. It makes it difficult to track history and there is no way to enforce the organization long term.

Haha, yeah, I get this response every time. I am aware of the confusion it may cause and I generally avoid doing that unless I believe leaving it as is would cause more confusion - I explained in the PR description why I believe in this case this was the lesser of two evils, keeping them in a file dedicated for mutants would just cause a wave of confusion by anyone trying to add more to these.

PR #46082, opened 2020-12-16
Exploit human psychology

anothersimulacrum:
#### Summary
SUMMARY: None

#### Purpose of change
Frame proficiencies as bonuses to crafting time as opposed to maluses.

#### Describe the solution
Convert displays to percentages, adjust grammar, and apply some math to the numbers before display.

image
image
image

image
image

image
image
(JSON for this recipe edited)

<details>
<summary>Really old images</summary>

image
image
</details>

#### Describe alternatives you've considered
<details>
<summary>Even more ancient images</summary>
(The math may be wrong on some of these)
Keeping white text/using multiplicative:
image
Using inverted multiplicative:
image
Using inverted percentages:
image
Using fractions:
image
image
</details>

#### Testing
See images.

#### Additional context
I'm not a particularly huge fan of doing this, but it's probably a good idea.

Issue #44738, opened 2020-10-08
change crafting skill check from dice to a normal curve.


I-am-Erk:
# Is your feature request related to a problem? Please describe.
In https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/crafting.cpp, Character::crafting_success_roll is a pretty arbitrary high-side dice roll. This is actually kinda counterintuitive, and also not very good for simulating things.

# Describe the solution you'd like
This, and skill rolls like it, should be handled by a normal curve function.

**First**: Calculate what your `mean_roll` is. This is equal to your skill level plus/minus modifiers.
1. Your mean_roll equals: `3*(primary_skill_level)+(sum of all secondary_skill_levels)`/`(number of secondary skills +3 )`, or basically, a weighted average where the primary skill level is 3x as important as any secondary skills. This is a change from the current ratio where the primary skill is always 75% of your success: the more secondary skills are needed, the less weight the primary skill has. However when you have one primary and one secondary skill, it remains the same as current.
2. Add in modifiers as currently used in crafting_success_roll: For example, the current code penalizes electronics and tailoring crafts specifically based on myopic and hyperopic. This should probably become data-driven, with a recipe flag based on skill, but keep that for now. instead of a dice penalty it is a penalty to mean_roll.
3. Adjust a value based on intelligence. Currently intelligence increases the size of your dice. Instead, let's modify your mean_roll by `(int-8)/4`.
4. If focus >50, add 20% to mean_roll. If focus >100, add 40%.
5. Add crafting helper modifiers, same as current except instead of boosting your dice, they are increasing your mean_roll.
6. Divide mean_roll by any fail_multipliers in proficiencies for the recipe. 

**Second**: Determine where your roll actually falls on the normal curve.
use normal_roll function in rng.cpp, with mean_roll as the mean (obvs) and stdev of 1.

I thought this section would be larger but apparently this is quite easy.

**Finally**: Judge success
- If your final value after the `normal_roll` is greater than the difficulty of the recipe, you succeeded! if your mean roll equals the difficulty of the recipe, this is 50% of the time.
- if your final value is less than 1 points below the difficulty of the recipe, you fail but waste nothing. This represents about 34% of values if your mean_roll equals the difficulty.
- If your final value is more than 1 points below the difficulty of the recipe, you fail and waste some components. This represents about 13% of rolls if your mean_roll equals the difficulty.
[...]

PR #63886, opened 2023-03-02
Rewrite crafting skill checks to use normal roll

anothersimulacrum:
#### Summary
SUMMARY: Features "Tweak crafting failure formulas and display failure chances"

#### Purpose of change
See #46153 for maximal detail, this replaces that.
The goals of this are two-fold:

1. Make calculating failure chance at all reasonable
2. By making failure chance calculation possible, enable tweaking and testing of the crafting failure formula

There are also longer-term goals such as https://github.com/CleverRaven/Cataclysm-DDA/issues/44738, and enabling partial skill to play a role.

#### Describe the solution
From @I-am-Erk 's rewrite of the formula:

1. NPCs no longer have to be more skilled than you to increase your chance of success. Instead, each NPC able to help that also has all the prerequisites to be able to do the craft on their own gets a skill check just like you. Then we add those skill checks together using the formula:

```
final skill value = sqrt( your skill value^2 + sum (all contributing NPC skill values^2) )
```

This root-sum-of-squares is the same system used in summing morale: it creates diminishing returns the more contributors you have, and makes more skilled sources of help substantially outweigh smaller ones. Using this formula, for example, if you and a helper NPC both have an effective skill level of 7 once all effects are applied, your combined skill level is around 10. If you add another NPC of effective skill level 7, you get a total combined skill of 12.

In the end this makes NPC helpers much more useful (previously they just added a small flat bonus, and only if they were more skilled than you).

2. Changes the role of secondary and primary skills in your check, making secondary skills potentially more important. Previously the weight of primary and secondary skill requirements was flat: 3/4 of your skill check was from your primary skill and 1/4 was from the average of the secondary skills. Now it is a weighted average of: `2*primary skill*required level of primary skill, + sum of secondary skills * required level of secondary skills`. That means if a recipe asks for electronics 2 and cooking 1 (what the hell recipe is this??), your overall level in electronics will be twice as important to your final skill value. If that same recipe uses fabrication 2 as its primary skill, then fabrication will weigh in at 2x as important as electronics and 4x as important as cooking.

3. Moves the method of calculating your effective skill level to a new method to make it easier to determine your and your NPCs' effective skills with all modifiers included.

From my additional changes:

4. Calculate and display the failure chance for a recipe https://github.com/CleverRaven/Cataclysm-DDA/commit/9f2ee4d78fb41ebe55698bf629816cf6cd73b22d
   
5. Adjust the chance for destruction of components to follow a different formula https://github.com/CleverRaven/Cataclysm-DDA/commit/58f5676cf1e53dfbbc3d0e5c221656afff79ba14, and display that chance (catastrophic failure chance). This is just an easier version of the above formula for now, I just fudged it to get pretty numbers.

6. Add some tests to:
    1) generate failure chance for every recipe in the game for some given stats related to the recipes
    2) Check if the actual failure rate of a number of recipes matches the calculated value, and that both match a value input, to detect changes to the failure rates.

#### Testing
See tests.
image
image
image
image
(Ridiculous percentages due to debug set skills command) 

#### Additional context
This has ping-ponged between Erk and I for a while, and the commit history reflects that. Mergers may want to avoid squashing for history reasons.
I do not expect this to be the final formula(s), this just puts the infrastructure in place.
The proficiency info screen is lying about how the failure rates work from proficiencies. I want to fix that (included in https://github.com/CleverRaven/Cataclysm-DDA/pull/46082/commits), but it's a lot of lines, and so I think it's better separate.
Closes #46153


I-am-Erk:
thank you for this

I-am-Erk:
we've done it Sim, we've destroyed one of the core parts of the game. Bwahahaha.

src/crafting.cpp
```diff
@@ -1008,95 +1008,248 @@ bool Character::craft_proficiency_gain( const item &craft, const time_duration &
...
+    data.final_difficulty -= 1;
+    data.final_difficulty *= 0.25;
+    data.stddev *= 0.5;
+    return data;
+}
+
+float Character::crafting_success_roll( const recipe &making ) const
+{
+    craft_roll_data data = recipe_success_roll_data( making );
+    float craft_roll = std::max( normal_roll( data.center, data.stddev ), 0.0 );
```
andrei8l, line 1183: I get an assertion failure here
```gdb
/usr/include/c++/12.2.1/bits/random.h:1995: std::normal_distribution<_RealType>::param_type::param_type(_RealType, _RealType) [with _RealType = double]: Assertion '_M_stddev > _RealType(0)' failed.
```
Build with `-Wp,-D_GLIBCXX_ASSERTIONS` and run `cata_test proficiency_gain_short_crafts`

Issue #63136, opened 2023-01-16
Metal Smoking Rack and Metal Butchering Rack require absurd amounts of Pipe Fittings


Zourin2:
### Describe the bug

Pipe fittings are not that common. No other recipe requires more than 4-6 pipe fittings, including the folding table.  Manufacturing the combined 54 pipe fittings for one of each would take more than 8 straight days of crafting.

### Steps to reproduce

Metal Smoking Rack recipe requires 30 sets of pipe fittings with no alternative
Metal Butcher Rack recipe requires 24 sets of pipe fittings with no alternative

### Expected behavior

These recipes should be reduced by a factor of 5-6, requiring 4-6 pipe fittings each to craft or an acceptable alternative should be added.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.22621.1105 (22H2)
- Game Version: 10bcab1 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    SpeedyDex [speedydex],
    Stats Through Skills [StatsThroughSkills]
]

NetSysFire:
Actually a duplicate of #52624 but that one was closed as stale.

flybye3o6:
i was looking into this because the weight and volume of both racks are very off compared to the resources that go into making them, the pipes and pipe fittings will have to be reduced and the racks have more weight added to them 
i wasnt able to find reference so its kinda guess work

Vaskritaya:
the weight could be all over the place, due to being able to use either copper tubing or pipes which have drastically different weights. That being said, youd need to use a certain amount of pipes to make a functional frame regardless of material. 

It's a bit of a tricky situation, but either way the fittings requirement is insane and the whopping 15 pipes needed for a smoking rack are nearly 4 times the weight of the finished product all on their own!

flybye3o6:
the fittings say its a loose assortment of things, so dont know exactly how much that is but fair to say 1 unit of fittings can do more then 1 joint, it is 0.5kg so idk about 4 or more joints, idk what the nails and wire are for
[...]

PR #63890, opened 2023-03-03
Butchering Rack Recipe Tweak

Fitzolir:
#### Summary
Balance "Butchering Rack Recipe Tweak"

#### Purpose of change

I made this recipe tweak to correct the Metal Butchering Rack's Crafting Cost to better reflect reality. The cost of the butchering table was already in question, as shown in [#63136](https://github.com/CleverRaven/Cataclysm-DDA/issues/63136) which this PR somewhat addresses. Hopefully, these changes make the recipe a bit more manageable! :)

#### Describe the solution
 
I re-evaluated the size and cost of the butchering rack to determine a more accurate pipe fitting, pipe, and spike cost for the recipe. My thought process is detailed below. (Credit to MrDraMaster & maleclypse for assisting with this!)
 
A butchering rack needs to stand at 8 feet, and the 'longest side' of the butchering table, when folded, is 60 cm. Converting 8 feet to cm, we get 240 cm. 240/60 is 4, which means there are 4 pipes per leg. Each pipe requires a pipe fitting between them, meaning there are 3 fittings per leg. There are 4 legs, so 3*4 which gives us 12 fittings so far. Accounting for the crossbar that the hooks are attached to, there are 2 pipes and then some. These pipes need a fitting between them as well, and one on both ends to hold the legs and crossbar together. This gives us our final estimate and formula for pipe fittings, **(240/60)*3+3=15**.
Additionally, our count tells us there are 18 pipes involved, not 12. (4*4+2)
Lastly, I changed the spike cost from 2 to 4. This change is mostly based on the in-game sprite & real-life butchering racks.

I considered the semantics of pipe fittings, and how many pipe fitting 'sets' should truly be required. When you stretch the definition of 'set' how it's applied in game, this recipe could take as little as 4 fitting sets.
I also considered changing the recipe to require welding instead, like most modern butchering racks.

#### Testing
As far as I'm aware, this PR doesn't need to be tested in-game. The only code changes made are number swaps; 12 to 18, 2 to 4, and 24 to 15. If it needs to be tested, I'll be more than willing to do so! The only test this requires (as far as I'm aware) is checking the in-game recipe to make sure it reflects the updated values.

#### Additional context
Mock-up based on in-game sprite
image
Snippet of PR discussion in #cdda-development on the C:DDA Discord. Provides some context to the math.
image

### DIFF
--- a/data/json/recipes/recipe_others.json
+++ b/data/json/recipes/recipe_others.json
@@ -2769,12 +2769,12 @@
     ],
     "components": [
       [ [ "scrap", 6 ] ],
-      [ [ "pipe", 12 ], [ "cu_pipe", 12 ], [ "frame", 2 ] ],
+      [ [ "pipe", 18 ], [ "cu_pipe", 18 ], [ "frame", 3 ] ],
       [ [ "nail", 12 ] ],
       [ [ "wire", 6 ], [ "rope_6", 6 ], [ "vine_6", 6 ], [ "rope_makeshift_6", 6 ] ],
       [ [ "chain", 1 ], [ "rope_natural", 1, "LIST" ] ],
-      [ [ "spike", 2 ] ],
-      [ [ "pipe_fittings", 24 ] ]
+      [ [ "spike", 4 ] ],
+      [ [ "pipe_fittings", 19 ] ]
     ]
   },
   {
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):
Thanks for doing this!

PR #63900, opened 2023-03-03
Prompt player on where he wants to climb down to when pressed 'move down' button

### LINKED ISSUE #50940: The current way the game has you get down from a roof is incredibly unintuitive. New players miss it regularly.
I find it extremely unintuitive that I can simply push the “ascend” key to go up on a drainpipe, but the “descend” key along the edge, even adjacent to the drainpipe square, does nothing.
If you’re over a drainpipe top graphic, it should work more like stairs.
I died once when first starting to play the game because I couldn’t figure out to “examine” the empty air spaces to climb down. And that’s a very odd use of the examine command, I believe the only one where you functionally examine empty space. And it’s also kind of dumb to have to do it to jump down. And it pops up a selection box to choose between climb and jump down every time as well? Cumbersome and unintuitive.

Here's my thoughts on an improved system.
(I believe all of this could also apply to ladders in some situations. I don't know if ropes are implemented for climbing as yet, though.)

**Climbing down:**

1) if you are orthogonally adjacent to a drainpipe square and push the "descend" key, you should just climb down the drainpipe, ignoring the possibility of climbing down any other nearby edges. This makes it simpler for the player, since if they maneuvered to the square adjacent to the drainpipe there are few to no cases where they were not doing so in an interest of climbing down it.
Also, if you are diagonally adjacent to a drainpipe square, but **not** adjacent to any other edge square, it should act in exactly the same way.

2) if you are adjacent to an edge **_and_** diagonally adjacent to a drainpipe and push the "descend" key, a selection box asking to choose between "climb down nearby drainpipe" and "climb down" could pop up. If there is more than one choice of non-drainpipe square, a yellow square selection box should appear asking the player to choose the square to climb down, omitting the drainpipe from the selectable squares (kind of how examining the squares now has you select one of the squares that are highlighted.)
If there is only one possible square for selection, once the player has confirmed they wish to climb down, it should just do it .

3) if you use the “descend” key command at an edge **without** a drainpipe adjacent in any way, it should jump directly to the part in (2) above where it asks you if you want to climb down, then prompt you to select a space if there is more than one choice.
Again, if there is only one possible square for selection, once the player has confirmed they wish to climb down, it should just do it .

5) attempting to move off the edge without pressing the descend command key should indicate that the player wishes to jump.
They should get a selection box with the choices "jump down" and "jump off roof", and if there is a nearby roof, the second option should change to "jump across".
The player does not need to select a square option because the attempted move selected the target square/direction already.
However, it has been suggested that the option to "climb down" should **_also_** be offered in this circumstance as well, since players may not know of the usage of the "descend" key in this circumstance, or just wish to use an initially simpler method of choosing the target square by movement.

Anyway, all that would make it simpler for everyone and more intuitive for new players. In Discord and Forum posts about this, several players mentioned that they had this issue as well.

image

I would code it, but I don’t know how…

Night-Pryanik:
#### Summary
Features "Prompt player on where he wants to climb down to when pressed 'move down' button"

#### Purpose of change
* Closes #50940 (despite it's already closed).

#### Describe the solution
When player presses `MOVE_DOWN` button, prompt him on direction he wants to climb down to.

Make player character automatically prefer downspout tiles when pressed climb down. In my testing I found this behavior rather confusing despite I knew what was gonna happen. Also there are some rare situations when player might prefer climbing down not on downspout, but to some other place - for example, if there's monster standing adjacent to downspout.

#### Testing
While on the roof and standing next to ledge, pressed climb down. Was able to successfully climb down using this button.
Stepped away from the ledge, pressed climb down. Got "You can't go down here" message.

#### Additional context
изображение

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -11112,8 +11112,7 @@ cata::optional<tripoint> game::point_selection_menu( const std::vector<tripoint>
         const std::string &up_or_down = up ? _( "Climb up %s (%s)" ) : _( "Climb down %s (%s)" );
         // TODO: Inform player what is on said tile
         // But don't just print terrain name (in many cases it will be "open air")
-        pmenu.addentry( num++, true, MENU_AUTOASSIGN, _( "Climb %s %s (%s)" ), up_or_down, dir_name,
-                        dir_arrow );
+        pmenu.addentry( num++, true, MENU_AUTOASSIGN, up_or_down, dir_name, dir_arrow );
     }
 
     pmenu.query();
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -11109,10 +11109,11 @@ cata::optional<tripoint> game::point_selection_menu( const std::vector<tripoint>
         // TODO: Sort the menu so that it can be used with numpad directions
         const std::string &dir_arrow = direction_arrow( direction_from( upos.xy(), pt.xy() ) );
         const std::string &dir_name = direction_name( direction_from( upos.xy(), pt.xy() ) );
-        const std::string &up_or_down = up ? _( "Climb up %s (%s)" ) : _( "Climb down %s (%s)" );
+        const std::string &up_or_down = up ? _( "Climb up %s (%s)" ), dir_name, dir_arrow :
+                                        _( "Climb down %s (%s)" ), dir_name, dir_arrow;
         // TODO: Inform player what is on said tile
         // But don't just print terrain name (in many cases it will be "open air")
-        pmenu.addentry( num++, true, MENU_AUTOASSIGN, up_or_down, dir_name, dir_arrow );
+        pmenu.addentry( num++, true, MENU_AUTOASSIGN, up_or_down );
     }
 
     pmenu.query();
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -11090,7 +11090,7 @@ void game::fling_creature( Creature *c, const units::angle &dir, float flvel, bo
     }
 }
 
-static cata::optional<tripoint> point_selection_menu( const std::vector<tripoint> &pts )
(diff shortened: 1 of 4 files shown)


src/game.cpp
```diff
@@ -11109,9 +11109,11 @@ static cata::optional<tripoint> point_selection_menu( const std::vector<tripoint
         // TODO: Sort the menu so that it can be used with numpad directions
         const std::string &dir_arrow = direction_arrow( direction_from( upos.xy(), pt.xy() ) );
         const std::string &dir_name = direction_name( direction_from( upos.xy(), pt.xy() ) );
+        const std::string &up_or_down = up ? _( "Climb up %s (%s)" ), dir_name, dir_arrow :
+                                        _( "Climb down %s (%s)" ), dir_name, dir_arrow;
```
Qrox, line 11113:
```suggestion
        const std::string &up_or_down = up ? _( "Climb up %s (%s)" ) : _( "Climb down %s (%s)" );
```

src/game.cpp
```diff
@@ -11109,9 +11109,11 @@ static cata::optional<tripoint> point_selection_menu( const std::vector<tripoint
...
         // TODO: Inform player what is on said tile
         // But don't just print terrain name (in many cases it will be "open air")
-        pmenu.addentry( num++, true, MENU_AUTOASSIGN, _( "Climb %s (%s)" ), dir_name, dir_arrow );
+        pmenu.addentry( num++, true, MENU_AUTOASSIGN, up_or_down );
```
Qrox, line 11116:
```suggestion
        pmenu.addentry( num++, true, MENU_AUTOASSIGN, string_format( up_or_down, dir_name, dir_arrow ) );
```
Perhaps this will work.

PR #63905, opened 2023-03-03
Makes Avian Eyes have the prerequirement of Insanely Perceptive

Kamejeir:
#### Summary
None

#### Purpose of change
When mutating bird, it has always been that getting more than Perceptive (the current prereq of Avian Eyes), would afterwards often try to downgrade to Perceptive instead of always ticking up to Very, Extremely and Insanely Perceptive until one reached Avian Eyes, at which point it would work normally. I think this just adds a whole lot of unnecessary luck-based genetic instability, to which I want to change it.

#### Describe the solution
Changed the prereqs of Avian Eyes from Perceptive to Insanely Perceptive, in addition to swapping the leads_to between the two.

Removing the prereqs of Avian Eyes

#### Testing
Got a full bird mutation line, removed Avian Eyes and all perception mutations, then started the mutation process.
The perception mutations only ever went up, and gave me Avian Eyes after Insanely Perceptive.
lastly, tested with pre-threshold mutation and it gave me perceptive and very perceptive, properly not giving me avian eyes or above very perceptive, as well as stayed at very perceptive like I expected it to

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -2075,7 +2075,7 @@
     "visibility": 1,
     "description": "Your vision has become particularly acute: you suspect you could pick out zombies much farther away.  Perception +4.",
     "types": [ "EYES" ],
-    "prereqs": [ "PER_UP" ],
+    "prereqs": [ "PER_UP_4" ],
     "threshreq": [ "THRESH_BIRD" ],
     "category": [ "BIRD" ],
     "passive_mods": { "per_mod": 4 }
@@ -5236,7 +5236,6 @@
     "description": "Your senses are a little keener.  +1 Perception.",
     "types": [ "PER" ],
     "changes_to": [ "PER_UP_2" ],
-    "leads_to": [ "BIRD_EYE" ],
     "category": [ "ALPHA", "CHIMERA", "BATRACHIAN", "ELFA", "RAPTOR", "BIRD", "SLIME", "SPIDER" ],
     "passive_mods": { "per_mod": 1 }
   },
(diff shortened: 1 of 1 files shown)


data/json/mutations/mutations.json
```diff
@@ -5273,6 +5272,7 @@
     "description": "You can sense things you never imagined.  +7 Perception.",
     "types": [ "PER" ],
     "prereqs": [ "PER_UP_3" ],
+    "leads_to": [ "BIRD_EYE" ],
```
Maleclypse, line 5275: Shouldn’t this be changes too? Otherwise it will be Perception +11 total instead of 7

Maleclypse: No I get that you can mutate them both currently in game but that doesn't exactly make it intended.  What would make it intended if this is the most significant ability that bird gets far and away.  I'm inclined to believe that's likely true given that bird mutants don't regenerate or fly or any other big post thresh abilities that I saw looking around.  I would guess that Bird is probably in an interim state while waiting for a significant retooling because +11 Perception isn't a playstyle changing group of mutations.

anoobindisguise: yeah it's pretty clearly meant for Bird to get both for +11 perc total. Perc is the weakest stat anyways so it's not exactly harmful. Someone just made it Perceptive only and not Very/Extremely/Insanely.

Maleclypse: That's fair, I happen to think that puts this in line for a rework to in some way change the game loop.  "See's good" isn't that useful unless it greatly increases chance to hit the best weakpoints and even then it would be pretty niche unless you are playing Dark Days of the Dead, headshots only.

PR #63910, opened 2023-03-03
New Radiophile mutation

Kamejeir:
#### Summary
Content "Radiophile mutation"

#### Purpose of change
I had the experience that the Integrated Dosimeter was the only way to have a completely passive method of knowing when to put on the cleansuit. This new mutation would, in addition to solving the dependence on the Exodii regarding this need, would fit into the mishmash of miscellaneous augments that is the Medical mutation tree.

#### Describe the solution
New mutation:
- In the Medical and Slime mutation line
- Will warn you if you're currently being irradiated (argue about cosmic rays if you must)
- Gives a small bonus to morale for current absorbed dose
- Can be turned off to not spam the log

Make the Integrated Dosimeter less annoying to use, such as making it passively tell you your absorbed dose instead of your having to check (might actually get around to doing that anyway)

#### Testing
It properly gives me a log message exactly every three minutes any time there's ambient rads to be had, and turning it off works as expected. Also gives morale as expected.
Mutating it properly works through the standard method.

### DIFF
--- a/data/json/morale_types.json
+++ b/data/json/morale_types.json
@@ -264,6 +264,11 @@
     "type": "morale_type",
     "text": "Masochism"
   },
+   {
+    "id": "morale_perm_radsense",
+    "type": "morale_type",
+    "text": "Tingly feeling"
+  },
   {
     "id": "morale_perm_noface",
     "type": "morale_type",
--- a/data/json/morale_types.json
+++ b/data/json/morale_types.json
@@ -264,7 +264,7 @@
     "type": "morale_type",
     "text": "Masochism"
   },
-   {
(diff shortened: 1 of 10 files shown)


Venera3:
I'd add Radiogenic as a prerequisite, and flavor it as a nice-feeling feedback on where your new flesh wants to grow - also slap it on SLIME in that case.

Venera3:
> You mean, like turning the rad detect message into something like "Your flesh tingles. It feels nice."? Or perhaps you mean that if you have wounds in tandem with radiogenic, it would tell you something special?

Yes, only a flavor change.

PR #63925, opened 2023-03-04
Cephalopod additions

KittyTac:
#### Summary
Content "Add more mutations to Cephalopod path"

#### Purpose of change
The Cephalopod path needs more love IMO and also I'm on a project to do some JSON touchups to various mutation paths.

#### Describe the solution
I gave it a pre-thresh mutation that decreases stealth but turns into one that increases stealth post-thresh, as well as added a post-thresh mutation that functions as an even better Fast Learner ~~and Skilled Liar~~... but that gives you a stronger version of the Forgetful effect. This is to represent octopus camouflage and mental flexibility respectively. To balance this out, it now gets Pain Sensitive.

Leaving as is, delving into C++ code to add a paralyzing venom.

#### Testing
Tested and it works.

#### Additional context
N/A, expect more stuff like this coming soon.


Karol1223:
First of all, lint before the tests yell at you

Second of all
> added a post-thresh mutation that functions as an even better Fast Learner and Skilled Liar
Why and how? I just read through the code and Cephalopod Mind seems to imply that you specifically no longer can understand humans. If anything it should have the opposite effect than making you a *better* liar. Not understanding the person you're talking to would make it harder to lie effectively and persuade them.

KittyTac:
> First of all, lint before the tests yell at you
> 
> Second of all
> 
> > added a post-thresh mutation that functions as an even better Fast Learner and Skilled Liar
> 
> Why and how? I just read through the code and Cephalopod Mind seems to imply that you specifically no longer can understand humans. If anything it should have the opposite effect than making you a _better_ liar. Not understanding the person you're talking to would make it harder to lie effectively and persuade them. The learning part fits the current mutation line though so I have no complaints with that.

You can coldly manipulate them using your intelligence. I'm fine with reducing the buff or removing it altogether but I'll wait for more input.

Venera3:
I don't think Liar works for them either. You might be great at controlling your chromatophores but the apes don't understand those anyway and *their* skin doesn't tell you what they are thinking.

It would be interesting to add something like senescence as a drawback to them - maybe a Disintegration-equivalent on a timer you need to stave off with staying in the higher levels of the Ceph mutagen effect.

Karol1223:
Nah social buffs for being an octopus don't work whatsoever.

KittyTac:
> I don't think Liar works for them either. You might be great at controlling your chromatophores but the apes don't understand those anyway and _their_ skin doesn't tell you what they are thinking.
> 
> It would be interesting to add something like senescence as a drawback to them - maybe a Disintegration-equivalent on a timer you need to stave off with staying in the higher levels of the Ceph mutagen effect.

Yeah I'll just remove it.

KittyTac:
I'm an idiot, I thought my PR was just in queue for 2 weeks but actually it was because there was a conflict. Sorry I'm a relative noob to this.

PR #63927, opened 2023-03-04
Allow the morningstar to be clipped to an appropriate belt

Inshal240:
#### Summary
Content "Allow morningstars to be clipped onto a suitable belt"

#### Purpose of change
Allow the morningstar to be carried more easily without a backholster or vest with sling. It should be easy enough to carry at your side rather than on your back. 

#### Describe the solution
I added the BELT_CLIP flag to the JSON for the morningstar.

None.

#### Testing
Loaded the game and stored the morningstar in my webbing belt with the large belt loops that were added recently/yesterday.

### DIFF
--- a/data/json/items/melee/bludgeons.json
+++ b/data/json/items/melee/bludgeons.json
@@ -1076,7 +1076,7 @@
     "bashing": 38,
     "cutting": 8,
     "to_hit": { "grip": "weapon", "length": "short", "surface": "any", "balance": "neutral" },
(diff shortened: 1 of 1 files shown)


GuardianDll:
I don't think having thing, full of metal spikes on your belt is a good idea

andrei8l:
`SHEATH_AXE` is probably more appropriate so that it goes in an axe ring holster and the spiky bits aren't right next to your body.

Inshal240:
Makes sense, but then I should probably add support for the axe ring to be attached to the MOLLE belt

PR #63945, opened 2023-03-04
Migration: remove temperature tracking for former comestibles

### LINKED ISSUE #63922: Un-cosmetibled items are (active) for some reason
### Describe the bug

Related to #63434 and #62800

The title.

### Attach save file

N/A

### Steps to reproduce

1. Change the game version to prior to #63434
2. Create a world, debug summon a lot of bandages (roughly 200)
3. Update the game to recent versions
4. Load the save

### Expected behavior

I expected the bandages to not haunt me.

### Screenshots

IMG_20230304_152800

- OS: Android
    - OS Version: Manufacturer: samsung; Model: SM-G991N; Release: 13; Incremental: G991NKSU3EWB1; 
- Game Version: 4ea6fb5 [64-bit]
- Graphics Version: Tiles
]

### Additional context

_No response_

cake-pie:
#### Summary
Bugfixes "Migration: remove temperature tracking for former comestibles"

#### Purpose of change
Resolve #63922 

#### Describe the solution
Add migration code to check for items that `!has_temperature()` but also non-default `last_temp_check` value or one of the flags `HOT`, `COLD` or `FROZEN`, which indicates it used to have temperature tracking but is no longer a comestible.

Before deactivating the item, check against `temperature_removal_blacklist` which lists former comestible items for which it is safe to do a straightforward `active = false;`. However, if a former comestible item can also be active for other reasons, then special case handling will be needed -- no such item so far, but it's good to future-proof for that possibility.

If an unexpected item matches the criteria but isn't covered by the migration list (or future special cases) then print a warning and abort migration without changing `last_temp_check` or clearing the temperature flags. This maintains the ability to detect and migrate the item later while also avoid messing with unforeseen items that shouldn't be migrated.

Not have the blacklist, which opens up the possibility for:
- some item gets incorrectly migrated (without required special case handling), which gets harder to solve
- unexpected item that shouldn't be migrated matches the criteria and gets its `active`, `last_temp_check` and flags mangled

#### Testing
Former comestibles preexisting in older saves no longer show up as "(active)"

### DIFF
--- a/doc/JSON_INFO.md
+++ b/doc/JSON_INFO.md
@@ -5198,6 +5198,21 @@ For recipes, deleting the recipe is enough.
 
 For mods, you need to add an `"obsolete": true,` boolean into MOD_INFO, which prevent the mod from showing into the mod list.
 
+## Charge and temperature removal
+
+If an item that used to have charges (e.g. `AMMO` or `COMESTIBLE` types) is changed to another type that does not use charges, migration is needed to ensure correct behavior when loading from existing save files, and prevent spurious error messages from being shown to the player.  Migration lists for this are found in `data/json/obsoletion/charge_removal.json`.
+
+Such items may be added to one of the following:
+
+* `charge_migration_blacklist`: items in existing save files with `n` charges will be converted to `n` items with no charges.  This will preserve item count.
+* `charge_removal_blacklist`: items will simply have charges removed.
+
+Additionally, `COMESTIBLE` items have temperature and rot processing, and are thus set as always activated.  When an item is changed from `COMESTIBLE` to a different type, migration is needed to check and unset this if applicable:
+
+* In most cases, the item has no other features that require it to remain activated, in which case it can be simply added to `temperature_removal_blacklist`.  Items in this list will be deactivated and have temperature-related data cleared *without any further checks performed*.
+* In case of an item that may be active for additional reasons other than temperature/rot tracking, an instance of the item loaded from existing save file cannot be blindly deactivated -- additional checks are required to see if it should remain active.  Instead of adding to the above list, a separate special case should be added in `src/savegame_json.cpp` to implement the necessary item-specific deactivation logic.
+
+
 # Field types
 
 Fields can exist on top of terrain/furniture, and support different intensity levels. Each intensity level inherits its properties from the lower one, so any field not explicitly overwritten will carry over. They affect both characters and monsters, but a lot of fields have hardcoded effects that are potentially different for both (found in `map_field.cpp:creature_in_field`). Gaseous fields that have a chance to do so are spread depending on the local wind force when outside, preferring spreading down to on the same Z-level, which is preferred to spreading upwards. Indoors and by weak winds fields spread randomly.
--- a/src/savegame_json.cpp
+++ b/src/savegame_json.cpp
@@ -2894,6 +2894,16 @@ void load_charge_migration_blacklist( const JsonObject &jo, const std::string &/
     charge_migration_blacklist.insert( new_blacklist.begin(), new_blacklist.end() );
 }
 
+static std::set<itype_id> temperature_removal_blacklist;
+
+void load_temperature_removal_blacklist( const JsonObject &jo, const std::string &/*src*/ )
+{
+    jo.allow_omitted_members();
+    std::set<itype_id> new_blacklist;
+    jo.read( "list", new_blacklist );
+    temperature_removal_blacklist.insert( new_blacklist.begin(), new_blacklist.end() );
+}
+
 template<typename Archive>
 void item::io( Archive &archive )
 {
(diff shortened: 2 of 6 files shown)


src/savegame_json.cpp
```diff
@@ -3044,6 +3054,34 @@ void item::io( Archive &archive )
         // Some hot/cold items from legacy saves may be inactive
         active = true;
     }
+
+    // Former comestibles that no longer need temperature tracking
+    if( !has_temperature() && ( last_temp_check != calendar::turn_zero || has_own_flag( flag_HOT ) ||
+                                has_own_flag( flag_COLD ) || has_own_flag( flag_FROZEN ) ) ) {
+        bool abort = false;
+        if( active ) {
+            if( temperature_removal_blacklist.count( type->get_id() ) != 0 ) {
```
andrei8l, line 3063: It's called a blacklist but acts as a whitelist. The whitelist is too much extra baggage for a minor migration issue.

src/savegame_json.cpp
```diff
@@ -3044,6 +3054,34 @@ void item::io( Archive &archive )
...
+                // list of items that are safe for straightforward deactivation
+                active = false;
+            // *INDENT-OFF*
+            // } else if( type->get_id() == ... ) {
+                // special case handling: item-specific deactivation logic needed for
+                // items that may be active for reasons other than temperature tracking
+            // *INDENT-ON*
```
andrei8l, line 3070: Delete. Don't micromanage future contributors.

src/savegame_json.cpp
```diff
@@ -3044,6 +3054,34 @@ void item::io( Archive &archive )
...
+            } else {
+                // warn and bail in unexpected cases
+                debugmsg( "Item %s is active and tracking temperature, but it should not!", type->get_id().str() );
```
andrei8l, line 3073: Downgrade to a DebugLog. No need to alarm the player. Also, can't you simplify these branches to just

```
if( !has_temperature() &&...) {
  if( temperature_removal_blacklist.count( type->get_id() ) == 0 ) {
     ...migrate....
  else {
     ...debuglog...
  }
}
```

src/savegame_json.cpp
cake-pie: > It's called a blacklist but acts as a whitelist.

I know; however same is true for `charge_migration_blacklist` and `charge_removal_blacklist` so I'm just following the existing convention.  If you insist, I can rename all of them.

> too much extra baggage

It's insurance against unintended consequences.
Can you provide a strong guarantee that `!has_temperature() && ( last_temp_check != ... )` will never match some other item that we don't actually want to migrate?

Unintended migration stomps over `last_temp_check` and temperature flags. It would lead to incorrect behavior that is very hard to detect by players, because it is difficult to discern from other normal behavior. e.g. for an item that ought to have been `HOT` one might think that it had cooled down naturally. I'd rather try to avoid that.

src/savegame_json.cpp
cake-pie: > Downgrade to a DebugLog

If we reach here, it means that either
- matched some unintended item that isn't supposed to be migrated -- need to revisit/refine the targeting criteria
- some item was changed and is no longer comestible, but forgot to think about the temperature tracking aspect -- needs to be added to the list or be given its own special case handling

Similar to [this](https://github.com/CleverRaven/Cataclysm-DDA/blob/7762a549d992baf6d275678bafefc8884ed8007c/src/savegame_json.cpp#L3095-L3096) and [this](https://github.com/CleverRaven/Cataclysm-DDA/blob/7762a549d992baf6d275678bafefc8884ed8007c/src/mapgen.cpp#L2893), I think its preferable for players to see and report so that the problem can be fixed, rather than just logged in a place that players don't usually pay attention to, and thus going undetected.

---

> simplify these branches

It's structured as:
```
1. deactivation logic:
   a. trivial case (in the list), just deactivate
   b. special cases, can only deactivate if no other reasons to remain active
   c. (implicit, no-op) item that isn't active but somehow does have temp. tracking vars/flags
   x. unexpected match: warn and abort
2. variable reset / flag clearing -- common to a, b and c
```
Suppose discard skeleton of 1b, and fold 1a, 1c and 2 together; it would actually look like:
```
if( !has_temperature() &&...) {
  if( active && temperature_removal_blacklist.count( type->get_id() ) == 0 ) {
     // case x
     ... warn ...
  else {
     // cases 1a&c + 2
     ... deactivate ...
     ... variable reset / flag clearing ... 
  }
}
```
Certainly the logic is sound and it will behave correctly for the time being.

However, semantics are subtly different -- whereas the list is originally intended as "items that can be trivially deactivated during migration", it becomes easy to treat this as simply "items to migrate", leading to more complacent approach of just blindly adding to the list whenever changing item from comestible to something else without sufficiently careful consideration.

Suppose in the future an item comes along that should fall into special cases (1b), this would lead to erroneous migration, breaking item processing of some other feature that needed `active == true`. I'm trying to avoid this eventuality; if you're sure you don't want these safeguards, it can be removed.

cake-pie:
Re: review feedback -- I've elaborated further on the rationale behind implementation decisions, mostly defensive safeguards. If it can be guaranteed that such protection is not needed, or consensus that prevention of potential mistakes is not wanted, then suggested changes can certainly be enacted.

src/savegame_json.cpp
cake-pie: deleted

cake-pie:
Commented code removed, safeguards retained, documentation added.

doc/JSON_INFO.md
```diff
@@ -5198,6 +5198,21 @@ For recipes, deleting the recipe is enough.
 
 For mods, you need to add an `"obsolete": true,` boolean into MOD_INFO, which prevent the mod from showing into the mod list.
 
+## Charge and temperature removal
+
+If an item that used to have charges (e.g. `AMMO` or `COMESTIBLE` types) is changed to another type that does not use charges, migration is needed to ensure correct behavior when loading from existing save files, and prevent spurious error messages from being shown to the player.  Migration lists for this are found in `data/json/obsoletion/charge_removal.json`.
+
+Such items may be added to one of the following:
+
+* `charge_migration_blacklist`: items in existing save files with `n` charges will be converted to `n` items with no charges.  This will preserve item count.
```
mqrause, line 5207: You don't need to remove it, but just for clarification this is going away again with #60885, making charge migration the default behavior.

src/savegame_json.cpp
mqrause: I agree with you to keep the debug message, since it's meant to catch unintended behavior and a log message at that point will surely remain unnoticed. But as for simplifying the logic I agree with andrei8l. There is not much of a point to give it a complex structure for cases that aren't even checked. That's work that can be done once such cases are actually identified.

PR #63815, opened 2023-02-26
Allow use of faction trust for any faction in eocs

Ramza13:
#### Summary
None

#### Purpose of change
Fix #63602

#### Testing
Used this json
```
  {
    "type": "effect_on_condition",
    "id": "EOC_TEST",
    "condition":{ "compare_int": [ { "faction_trust": "free_merchants" }, "<=", { "const": 10 } ] },
    "effect": [ { "u_message": "You feel untrusted" }, { "arithmetic": [ { "faction_trust": "free_merchants" }, "+=", { "const": 2 } ] } ],
    "false_effect": [ { "u_message": "You feel trusted" } ]
  },
```

PR #63958, opened 2023-03-05
Add some reputation-related events

MNG-cataclysm:
#### Summary
Content "Begin adding some reputation-related events"

#### Purpose of change
#63815 and #63919 added some ways to use and alter faction standing in EOCs. This PR begins to add some world consequences based on the player's standing with existing factions.

#### Describe the solution
If you become sufficiently disliked by a faction beyond outright hostility, they will begin to make efforts to try and kill you. The Hell's Raiders will send thugs after you, in increasing numbers depending on how far out of favor you are. The Old Guard will consider you a high-priority target and send a kill team of two, heavily armed and skilled fighters. The Exodii and Hub 01 will attempt to kill you using a bomb, which can wreck player bases and even destroy vehicles. These will only occur every 1-2 weeks, and NPC fighters will not arrive in portal storms or in certain areas, such as faction bases. 

Not adding this, lowering the firepower of these events, or increasing the time between them.

#### Testing
I decreased my standing with these factions significantly and waited a while. Everything works as intended.

#### Additional context
All of these events require significant effort on part of the player to decrease standing that much. If needed, the requirements could be increased.

### DIFF
--- a/data/json/effect_on_condition.json
+++ b/data/json/effect_on_condition.json
@@ -639,7 +639,7 @@
     "id": "EOC_ROBOFAC_ASSASSIN",
     "recurrence": [ "7 days", "14 days" ],
     "condition": {
-      "and": [ { "compare_int": [ { "faction_like": "robofac" }, "<", { "const": -30 } ] }, { "days_since_cataclysm": 5 } ]
+      "and": [ { "compare_int": [ { "faction_like": "robofac" }, "<", { "const": -230 } ] }, { "days_since_cataclysm": 5 } ]
     },
     "effect": [ { "mapgen_update": "exodii_bomb_add" } ]
   }
--- a/data/json/effect_on_condition.json
+++ b/data/json/effect_on_condition.json
@@ -556,7 +556,7 @@
   },
   {
     "type": "effect_on_condition",
-    "id": "EOC_BANDIT_ASSASAIN",
+    "id": "EOC_BANDIT_ASSASSIN",
     "recurrence": [ "7 days", "14 days" ],
     "condition": {
       "and": [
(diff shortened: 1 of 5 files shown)


anoobindisguise:
Wait isn't Hub01 always an archenemy for weird technical reasons? Won't they just insta bomb you immediately?

andrei8l:
How are they tracking you?

MNG-cataclysm:
> Wait isn't Hub01 always an archenemy for weird technical reasons? Won't they just insta bomb you immediately?

Yes, they would. I've set the EOC condition for them to a new low, which exceeds that they have set as a default, so you won't instantly die.

MNG-cataclysm:
> How are they tracking you?

Drones, bounty hunters, private eyes, mainly traditional investigation methods.

andrei8l:
>  mainly traditional investigation methods.

They show a picture of you to the zombies and ask which way you headed last?

actually-the-artificer:
shouldn't these stop at some point? Like if you wipe all their best dudes they just give up?

MNG-cataclysm:
> shouldn't these stop at some point? Like if you wipe all their best dudes they just give up?

It seems to me that killing their best members would only empower the faction in question to continue sending assassins after you, and possibly redouble their efforts to kill you, as you would exemplify your danger to them. The attacks would only stop through peace negotiations; they could bargain with you after some number of assassins fail to kill you, or by you reaching out first.

actually-the-artificer:
Ah; I was talking about the fact that they should at some point run out of people to throw at you. Especially for the smaller factions.

Naadn:
I think factions should have stop loss point once they send a few bubbly and all of them didn't come back. As long as player don't mess them again they will stop bothering you.

PR #63960, opened 2023-03-05
Makes earthshaper cestus nonconductive

adamkad1:
#### Summary
Mods "Makes earthshaper cestus nonconductive"

#### Purpose of change

Normal cestus is non conductive. Nail knuckles are non conductive. I don't know where the scrap goes (probably into runes?), but it does not make sense for earthshaper cestus to be conductive. And with the added stone, its probably even less conductive than nail knuckles could be.

#### Describe the solution

Add nonconductive flag to earthshaper cestus.

idk

#### Testing

add flag, check in game, works.

#### Additional context

Nope


Maleclypse (approved):
Good reasoning.

PR #63970, opened 2023-03-05
Revisit compiler support decisions

jbytheway:
#### Summary
Infrastructure "Update compiler support.  Now supporting gcc 8.1+, clang 10+, XCode 10.1+"

#### Purpose of change
Having just released 0.G, it's a good time to revisit our complier support choices.

This PR is the first step in migrating in this direction, by reviewing the various operating systems and seeing what the oldest versions we should support are according to our support policies.

If these decisions are agreeable, the next steps (in separate PRs) are:
* Update CI to test newer compiler versions (and stop using Ubuntu Bionic, if we still are using that anywhere).
* Update compiler flags to C++17.
* Start taking advantage of new features!

#### Describe the solution
I decided to be a little aggressive and forward-date these decisions to May, when Ubuntu Bionic will be out of general support.  That moves our oldest gcc version up to 8.1 (rather than 7.3) and allows us to adopt most of C++17.  But there's actually only one added feature in gcc 8 (elementary string conversion for integers) so we can easily hold off on breaking Ubuntu Bionic until May by not using that for a little bit.

Sticking with gcc 7.3 for another couple of months.

#### Testing
View the rendered markdown [here](https://github.com/jbytheway/Cataclysm-DDA/blob/update_compiler_support_docs/doc/COMPILING/COMPILER_SUPPORT.md).

#### Additional context
This will be the first time we've adopted a new C++ standard with some provisos (i.e. some features we can't have).  This could be a minor challenge for contributors, but so long as we have robust CI on the oldest compiler versions I believe it will work out.

We really have no choice but to do this because some C++17 features are still not supported in some compilers even to this day.


akrieger:
We have github workflows & releases for Visual Studio now. Besides the question of whether to deprecate the cross build releases, we should at least add VS to the supported workflows list as well?

VS2019 supports C++17 decently well, VS2022 supports it very well, and in some ways supports C++20 better than gcc (libstdc++ specifically).

irwiss:
Is codeblocks actually "supported" or even used by anyone? I tried it super long time ago but the experience hasn't changed much; re-did a quick test on windows 10 and latest "stable" codeblocks from 2020 (version 20.03);

Codeblocks "no admin" installation is flat out broken - tries to install into windrive:\program files which fails as it requires admin

no-mingw-bundled installation fails to detect the downloadable version of mingw - which actually bundles gcc 12.2 rather than 8, but the "autodetector" inside codeblocks fails

mingw-bundled installation sorta opens the .cbp file, but compiling fails finding flatbuffers headers, requiring you to put `-isystem` pointing at src/thirdparty/ dir somewhere in the compiler settings, only to start failing on zlib and such, which further requires another `-I` into headers dir of vcpkg or whatever packager you're using... It then fails with more errors screaming at wincurse.cpp COLORS and flatbuffer ClassicLocale...

At this point my patience ran out, I'm guessing either no one compiled using codeblocks on windows at least since flatbuffers commit merged around september 2022 or didn't contribute changes needed to make CB able to compile "out of the box"

I'm not sure if there's a reason to use ancient mingw on linux with or without codeblocks, but there also aren't any docs for doing this - so effectively there is no "support" to begin with

jbytheway:
I would prefer not to support Code::Blocks. But since its supported features are on par with XCode right now I didn't want to argue for its removal. Since it's not in CI the support was always going to be flaky at best.

jbytheway:
> We have github workflows & releases for Visual Studio now. Besides the question of whether to deprecate the cross build releases,

As you know, I would like to switch to the non cross builds as default, but inbred that's separate from this issue of upsetting compiler requirements.

> we should at least add VS to the supported workflows list as well?

Agreed. I'll add that.

> VS2019 supports C++17 decently well, VS2022 supports it very well, and in some ways supports C++20 better than gcc (libstdc++ specifically). 

I guess we should bump the minimum supported VS to 2019, then?

esotericist:
> Is codeblocks actually "supported" or even used by anyone? I tried it super long time ago but the experience hasn't changed much; re-did a quick test on windows 10 and latest "stable" codeblocks from 2020 (version 20.03);

it's something that occasionally comes up, and we've even had a few members of the team try to use it in the past, but we always tried to encourage moving to something less... brittle.

but given the state it's been for years now, i'd like to suggest we pivot code::blocks from merely "not officially supported" to "officially unsupported", barring someone on that team actually putting in the effort to bring that mess at least vaguely into functionality that resembles parity with any other mainstream toolchain.

in my opinion, it's clearly no longer fit for purpose, and we shouldn't enable anyone using it.

akrieger:
> As you know, I would like to switch to the non cross builds as default, but inbred that's separate from this issue of upsetting compiler requirements.

I mean, I guess it is relevant iff it changes what minimum compiler version we need to/can support. I did go check though and the mxe gcc version is 12.1 though so the point is moot. Dropping those builds won't change the calculus for gcc.

> I guess we should bump the minimum supported VS to 2019, then?

We would have to, yes. I would almost want to suggest saying officially 2022 is minimum (it's freely accessible to anyone and still supports all supported Windowses) and it'll lessen the risk of fringe c++17 issues.

jbytheway:
I've bumped the minimum VS to 2019 and mentioned that we test that in CI.  We can switch to 2022 if there's need, but that should probably be accompanied by a change to CI, so I don't think it's a good idea to change that in this PR.

The consensus seems to be to drop Code::Blocks (which makes me happy), so I've removed that from the docs and also remove the `.cbp` project file.

kevingranade:
Also of note, github is in the process of deprecating their ubuntu 18.04 image, as in we are seeing outages right now.

esotericist:
> Also of note, github is in the process of deprecating their ubuntu 18.04 image, as in we are seeing outages right now.

i had been waffling on saying something about the 18.04 thing, but this makes it more pertinent. i would suggest we can as a matter of principle assume anything that is shifting tiers of support *between* stable cycles is fair game for us, even in advance of that date. it's experimental, after all.

PR #63974, opened 2023-03-05
Allow mutations to remove starting traits, unhardcode more mutation functionality

Venera3:
#### Summary
Features "Allow mutations to remove starting traits"

#### Purpose of change

Currently mutations are prevented from removing starting traits (aside from threshold mutations, but none use this functionality currently) - this is both easily gameable and not very flavorful. Change that, and expand the mutation system.

#### Describe the solution

- Went crazy and slapped a debug mode message on pretty much every single step of the mutation process to facilitate understanding the logic
- Added the `base_removal_chance` and `base_removal_cost_mul` variables to mutation categories - the former governs the chance of removing a conflicting starting trait per mutation attempt, the latter acts as a multiplier on the mutation cost if in the process of gaining the trait you gave up a starting one
- Hooked up the logic to remove the starting traits - traits removed like this will get purged from your starting trait list, so you're not getting them back with purifier after the fact
- Added `threshold_min` to categories to allow hand-picking the necessary primer level to break a threshold, mostly because it was a low-hanging unhardcoding to do

TODO:
- [x] Do some unit tests, probably
- [ ] Potentially unhardcode the message for removing a starting trait, we'll see if I can come up with something fun for Standard Mammal Categories 1-15

Different chances
Using a flat vitamin cost, but tying it to the end effect sounded more reasonable - as mutation costs get expanded it should differentiate between a huge change rewriting most of you insides and your fingernails not being that sharp
PR the threshold limit separately, but might as well `While I'm Here` it

#### Testing

Previously couldn't lose a starting trait (BADKNEES in the tests), either with purifier, mutagen, or the debug menu.
The debugmode messages show up and explain the logic in a pretty verbose way.
Post-changes purifier leaves my knees alone, the other options remove them without incident.
Cost multipliers and chances are read and applied as expected.

#### Additional context

Purifier can't remove starting traits, human-ish categories only remove 20% of the time at 3x cost multiplier, scaling to 100% chance and no multiplier for Slime/Chimera.

Random-type mutations get to remove starting traits currently, that can go either way but they are usually meant as a penalty for radiation/involuntary mutation.

<details>
  <summary>Debug mode messages</summary>

grafik

Too much? Nonsense!

grafik

</details>

### DIFF
--- a/data/json/mutations/mutation_category.json
+++ b/data/json/mutations/mutation_category.json
@@ -13,7 +13,9 @@
     "threshold_mut": "",
     "skip_test": true,
     "mutagen_message": "You feel the aberrance fade.",
-    "vitamin": "mutagen_human"
+    "vitamin": "mutagen_human",
+    "//": "Starting trait removal: Human-but categories: 20% chance 3x cost multiplier, vertebrates: 33% chance 2,5x cost, invertebrates: 50% chance 2x cost, non-Animalia: 75% chance 1.5x cost, extradimensional/generally a mess: 100% at base cost",
+    "base_removal_chance": 0
   },
   {
     "type": "mutation_category",
@@ -22,7 +24,9 @@
     "threshold_mut": "THRESH_PLANT",
     "mutagen_message": "You feel much closer to nature.  Your breath and thoughts slow.",
     "memorial_message": "Bloomed forth.",
-    "vitamin": "mutagen_plant"
+    "vitamin": "mutagen_plant",
+    "base_removal_chance": 75,
+    "base_removal_cost_mul": 1.5
   },
   {
     "type": "mutation_category",
@@ -31,7 +35,9 @@
     "threshold_mut": "THRESH_INSECT",
     "mutagen_message": "You hear buzzing, and feel your body harden.",
     "memorial_message": "Metamorphosed.",
-    "vitamin": "mutagen_insect"
+    "vitamin": "mutagen_insect",
+    "base_removal_chance": 50,
+    "base_removal_cost_mul": 2.0
   },
   {
     "type": "mutation_category",
@@ -40,7 +46,9 @@
     "threshold_mut": "THRESH_SPIDER",
     "mutagen_message": "You feel insidious, and your vision fragments for a moment.",
     "memorial_message": "Found a place in the web of life.",
-    "vitamin": "mutagen_spider"
+    "vitamin": "mutagen_spider",
+    "base_removal_chance": 50,
+    "base_removal_cost_mul": 2.0
   },
   {
     "type": "mutation_category",
@@ -49,7 +57,8 @@
     "threshold_mut": "THRESH_SLIME",
     "mutagen_message": "Your body loses all rigidity for a moment.",
     "memorial_message": "Gave up on rigid human norms.",
-    "vitamin": "mutagen_slime"
+    "vitamin": "mutagen_slime",
+    "base_removal_cost_mul": 1.0
   },
   {
     "type": "mutation_category",
@@ -58,7 +67,9 @@
     "threshold_mut": "THRESH_FISH",
     "mutagen_message": "You are overcome by an overwhelming longing for the ocean.",
     "memorial_message": "Went deep.",
-    "vitamin": "mutagen_fish"
+    "vitamin": "mutagen_fish",
+    "base_removal_chance": 33,
+    "base_removal_cost_mul": 2.5
   },
   {
     "type": "mutation_category",
@@ -67,7 +78,9 @@
     "threshold_mut": "THRESH_GASTROPOD",
     "mutagen_message": "Your split lower body starts to come back together.",
     "memorial_message": "Left a trail.",
-    "vitamin": "mutagen_gastropod"
+    "vitamin": "mutagen_gastropod",
+    "base_removal_chance": 75,
+    "base_removal_cost_mul": 1.5
   },
   {
     "type": "mutation_category",
@@ -76,7 +89,9 @@
     "threshold_mut": "THRESH_RAT",
     "mutagen_message": "You feel a momentary nausea.",
     "memorial_message": "Found that survival *is* everything.",
-    "vitamin": "mutagen_rat"
+    "vitamin": "mutagen_rat",
+    "base_removal_chance": 33,
+    "base_removal_cost_mul": 2.5
   },
   {
     "type": "mutation_category",
@@ -85,7 +100,9 @@
     "threshold_mut": "THRESH_BEAST",
     "mutagen_message": "Your heart races and you see blood for a moment.",
     "memorial_message": "Embraced their bestial nature.",
-    "vitamin": "mutagen_beast"
+    "vitamin": "mutagen_beast",
+    "base_removal_chance": 33,
+    "base_removal_cost_mul": 2.5
   },
   {
     "type": "mutation_category",
@@ -94,7 +111,9 @@
     "threshold_mut": "THRESH_URSINE",
     "mutagen_message": "You feel an urge to… patrol?  The forests?",
     "memorial_message": "Became one with the bears.",
-    "vitamin": "mutagen_ursine"
+    "vitamin": "mutagen_ursine",
+    "base_removal_chance": 33,
+    "base_removal_cost_mul": 2.5
   },
   {
     "type": "mutation_category",
@@ -103,7 +122,9 @@
     "threshold_mut": "THRESH_FELINE",
     "mutagen_message": "Your back arches and the hair stands up on the back of your neck.",
     "memorial_message": "Realized the dream.",
-    "vitamin": "mutagen_feline"
+    "vitamin": "mutagen_feline",
+    "base_removal_chance": 33,
+    "base_removal_cost_mul": 2.5
   },
   {
     "type": "mutation_category",
@@ -112,7 +133,9 @@
     "threshold_mut": "THRESH_LUPINE",
     "mutagen_message": "You feel an urge to mark your territory.  But then it passes.",
     "memorial_message": "Wolfed out.",
-    "vitamin": "mutagen_lupine"
+    "vitamin": "mutagen_lupine",
+    "base_removal_chance": 33,
+    "base_removal_cost_mul": 2.5
   },
   {
     "type": "mutation_category",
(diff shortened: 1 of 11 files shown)


Terrorforge:
When I discussed this with @Ilysen in #61036 [they said](https://github.com/CleverRaven/Cataclysm-DDA/pull/61036#issuecomment-1252954146) that the prevailing opinion at the time was that Purifier _should_ revert starting traits. That was almost six months ago, of course, so has it changed or is it still (eventually) the plan?

Venera3:
> When I discussed this with @Ilysen in #61036 [they said](https://github.com/CleverRaven/Cataclysm-DDA/pull/61036#issuecomment-1252954146) that the prevailing opinion at the time was that Purifier _should_ revert starting traits. That was almost six months ago, of course, so has it changed or is it still (eventually) the plan?

It's a good destination, but as it stands the removal won't cause any instability (since you don't *actually* gain any traits from purifier) - since you're safe from any consequence beyond the purifier effect excesses it would just be an easy freebie. It might become a thing with some work on instability as a system, and the actual removal-unlocking will be just a JSON edit away when that happens.

Raikiri:
I wish any of that debug information was relayed to the player in any way, shape or form. Currently there's absolutely no way to understand what to expect from the mutation system using just in-game data.

Venera3:
You get the Warping effect when you consume enough mutagen to qualify and if you try to mutate without a primer you explicitly get told to use it. It could always get better, but I feel like it's relatively well signposted. Exposing the whole word salad up there will not help in any case.

Raikiri:
The primer/catalyst part is relatively well telegraphed and is as discoverable as it gets.

However, all of the mechanics that separate using mutagens to your advantage vs gimping yourself have no in-game explanation:
- as basic as what negatives do you expect to get from a given mutation tree?
- how does instability influence the chance of gaining negative mutations?
- nothing in the game explains that if a post-threshold mutation has a negative mutation prerequisite then you can still get it with 0 instability. this is one of the most critical and at the same time the least intuitive interactions.
- nothing in the game tells you that "Light Eater" is of the same time as "Very Fast Metabolism" (the both belong to METABOLISM type) so they will cancel each other
- nothing in the game tells you what prerequisites a given mutation has and nothing explains to you that if you can't get a prerequisite then that mutation is blocked
- currently traits are in this weird spot when they have critical strategic importance to planning your build, the can't be removed, but they can be "Changed To". I guess you're planning to get rid of it so maybe it's going to have less impact, but currently it's one of those critically important mechanics that has no explanation.
- nothing tells you which CBM cancels which mutations. I understand that the whole CBM system is slated to undergo significant changes, but I don't think it's a good justification to just neglecting it until that happens.

I guess my greatest concern that this system used to be opaque but at least it functioned relatively simply. Over the last couple of years it got much more complicated (in my opinion, much more interesting overall), but there's no way an average player can actually appreciate it because of how opaque it is.

Venera3:
The core difference of opinion is I *don't* think planning mutations out should be encouraged as a game mechanic - it's a gamble, it should have high enough highs to make you roll the dice but not something you can plan around. Even the current system is hilariously straightforward, since you *never* gain any out-of-category mutations.

>     * as basic as what negatives do you expect to get from a given mutation tree?
> 

Cat will turn you into a sleepy dodgy carnivore. Bear will turn you into a tanky big boy who may or may not actl like he can hibernate. I don't see the point of exposing every single trait id, they should be intuitive from the category.


>     * how does instability influence the chance of gaining negative mutations?

Barely, but it should be signposted better. For the record, in my tests I regularly broke threshold with 0,3 bad trait chance.

>     * nothing in the game explains that if a post-threshold mutation has a negative mutation prerequisite then you can still get it with 0 instability. this is one of the most critical and at the same time the least intuitive interactions.

Again, only critical if you want this to be a 100% happy no bad traits ever experience you plan out from the word go, which is not really my idea of mutagen.

>     * nothing in the game tells you that "Light Eater" is of the same type as "Very Fast Metabolism" (they both belong to METABOLISM type) so they will cancel each other

What should tell you about it? Short of a dynamic pop-up on mutagen application going `WARNING, 12% chance of removing Trait X` I don't really see an open place in-game.

>     * nothing in the game tells you what prerequisites a given mutation has and nothing explains to you that if you can't get a prerequisite then that mutation is blocked

Again, what is the alternative? You gain the prereqs by mutating, the mutation takes care of giving you prereqs - if they don't that's a bug, but not a part of the system to shove in the player's face.


>     * nothing tells you which CBM cancels which mutations. I understand that the whole CBM system is slated to undergo significant changes, but I don't think it's a good justification to just neglecting it until that happens.

This is fair, and adding a special message is not put ot scope for this.


> Using the hitchhiker's guide is currently pretty much an obligatory thing to actually benefit from the mutation system. Not only that, you _also_ have to learn from somewhere how it works.

Relying on HHG is a problem we don't really get to solve, sadly, unless the game starts with a scrolling pop-up of /doc for everybody. For explaining the system an update of the mutation tutorial in the help menu is a question of text editing and would definitely be very welcome.

Zireael07:
> Relying on HHG is a problem we don't really get to solve, sadly, unless the game starts with a scrolling pop-up of /doc for everybody. For explaining the system an update of the mutation tutorial in the help menu is a question of text editing and would definitely be very welcome.

That is, unfortunately, a problem. Right now mutations are a game of chance you don't know the rules of, and end up force closing to "reroll"

Terrorforge:
> If a system requires 4 pages of explanation to be usable, it means that we probably need to start integrating this information into the game somehow before making it even more complicated.

This change makes the system less complicated, because it removes a weird special exception that starting traits have. I guess I don't really understand why starting traits need to be less likely and more expensive to remove, but knowing the exact % mechanics of how it works isn't terribly relevant to making effective use of mutagen.

> That is, unfortunately, a problem. Right now mutations are a game of chance you don't know the rules of, and end up force closing to "reroll"

Or you accept that when you roll the dice, sometimes you lose. There's a grand total of like two or three mutations you could reasonably consider run enders, everything is else is very dealwithable. I really don't understand why people keep insisting that mutations are useless unless you can game the system to get 100% pure upside. Imo, the weird restrictions are the most fun part of playing a mutant.

Granted, it _is_ a problem that any kind of anatomical mutation denies you access to basically the only practical way to get acid and electrical resist, but the fact that an activity suit is mandatory after week three is a problem with armor and encounter design

Zireael07:
@Terrorforge The problem is not "you lose", the problem is "you don't know the rules of the game"

Raikiri:
Venera3
> Even the current system is hilariously straightforward, since you never gain any out-of-category mutations.

I hard disagree with this statement. If you actually wanted a simple roulette-based mutation system, it could have been much simpler, much more intuitive with much fewer moving parts. For example, if it was the case, why do we even need to have instability at all, why do we even have different chances of getting positive/negative mutation, why do we even treat traits in any different way at all, why do we even have the purifier that works the way that it does, etc etc.

> Bear will turn you into a tanky big boy who may or may not actl like he can hibernate

Ok let's go there. I see 5 major ways to interact with the mutation system:
a) I just want to be a bear UwU
b) I want to become a bear that will most likely gimp my character but I'm willing to accept the challenge
c) I want to have bear strenth, but I also don't want to permanently gimp myself and become a carnivore that can no longer wear my arrow and who can't fix it due to paw encumbrance, so I'll need to start with an expanded digestive system CBM as well as arm plating CBM or alternatively I'll have to find fish mutagen to overwrite my paws with webbed hands and as for carnivore.. wait but actually bear seems to be the only exception that does not actually make me carnivore, it makes me a fake carnivore https://nornagon.github.io/cdda-guide/#/mutation/CARNIVORE_FAKE that's actually fine, however <....>
d) I would rather not even touch the system that's likely make me much weaker than I am right now.
e) I'll just a savescum the system that's unnecessarily overcomplicated

As far as I understand, you want the system overall to work like a). But in reality currently people either go b) for challenge purposes, they go c) if they have a PhD in cdda mutations or they just go d)/e) because they can't be bothered. While it might be possible to make more people go a), I don't even know what kind of changes would encourage that.

> What should tell you about it? Short of a dynamic pop-up on mutagen application going WARNING, 12% chance of removing Trait X I don't really see an open place in-game.

I think the only solution to make the system more intuitive is to just have less moving parts that need explanation before becoming useful.

Terrorforge
> This change makes the system less complicated, because it removes a weird special exception that starting traits have.
As far as I can see, this changes adds more parameters (base_removal_chance and base_removal_cost_mul, etc) hence making it more mechanically complex and introducing even more moving parts.

> Or you accept that when you roll the dice, sometimes you lose. 
I agree that it can work this way in the end (which would correspond to "a)" path above), but in practice it almost always results in suicide unless you know what consequences to expect. Things like being forced out of your armor, being practically unable to craft, being unable to use your vehicle, becoming a carnivore are guaranteed to give an unuspecting person much more problems than they offer benefits. "But I can hybernate now" just does not cut it in terms of cost/benefit.

> Granted, it is a problem that any kind of anatomical mutation denies you access to basically the only practical way to get acid and electrical resist, but the fact that an activity suit is mandatory after week three is a problem with armor and encounter design
It is A problem, but not The problem. There's multiple other dealbreakers like carnivore, 30% craft speed (combined with the need to craft mutant anatomy gear) vehicle problems.

AcidAntOnAMinefield:
I have to agree with the notion that there is pretty much no information conveyed to the player in-game. I used to take a dip into mutations before because the system was easier to understand, despite being a massive resource salad and time sink.

As it currently stands, it's a system that gives the player pretty much no feedback and no instructions - and if the player isn't one of the 4 or 5 people who worked on the code, they probably won't know how things work other than by lucky trial and error. This, of course, assuming they aren't given false or tampered information in the process. And since it was implemented, I tried it on a couple of different occasions but decided it isn't worth the time. The game does a poor job at explaining itself and there are no legitimate and up-to-date sources available for a player to even begin to try and understand it. Unless those sources are hidden behind a discord channel, which, if so, that's a really bad decision, given that information should be in a wikia or a pdf of sorts available to the players. This goes along the lines of the several occasions I've seen complaints about lack of design documents - which is very much a problem I can see. This system is, in my opinion, not good: it's poorly explained, it's over-complicated for the sake of being overly-complicated, due to the perception that mutations were too overpowered before.

In my opinion, the current system should go away, and mutagenic stuff should be limited to a faction as a reward for long-term help. Your "average joe" should never be able to just waltz into a lab, read a book for 4 or 5 days and suddenly come out of it with all the needed equipment to install in a vehicle to start moonshining the hell out of their DNA.
We currently have so many factions and little groups scattered around the world: woodcutters, religious nutcases, norse-wannabe nazis, fungal emissaries, an undertaker/seer. The Exodii were created as an excuse to move CBMs away from the world, which was a nice decision (pity for the fact that one can STILL get and install them without the Exodii, which makes them pretty much moot). Why not have the Hub01 be the place were mutations are being studied? It's an excuse to open up access to the actual underground of the building that exists solely to add insane lag to that area. The player could start helping them by bringing "samples" and tools and stuff for them. Hell it could even be revealed that some of the people within the Hub are *actually* mutants who are trying to investigate this issue - in the beginning they want to find a way to reverse it and cure people, but as they start seeing the potential, they start shifting towards helping survivors "adapt" to the new world, offering them a better chance to easily survive the dangers the current world offers.

PR #63977, opened 2023-03-05
Add Bungalows 15 and 16

LilShiningMan:
#### Summary
Content "add bungalow 15 and 16"

#### Purpose of change

increase house spawn variety

#### Testing
linted and spawned in-game.

#### Additional context
bungalow15 g
bungalow15 r
bungalow16 g
bungalow16 r


LovamKicsiGazsii:
Thanks for not forgetting about the gutter changes.

For bungalow 15, could you add a path to the shed and some sidewalk to that back exit like you did in a previous PR? Also, it seems that the x and y coordinates for the lawnmower are mixed-up.

And for bungalow 16, could you please add at least one  trash can somewhere (maybe in that empty corridor by the kitchen)?

Also, no need to define terrain under roof furniture once    `fill_ter` is defined.

LilShiningMan:
all done except the extra path in bungalow 15. I think I've been adding too much sidewalk to my houses, and I wanted to avoid that with this one. Instead I've added a bit of sidewalk by the shed door.

LilShiningMan:
Can someone tell me what these errors are all about? The only files I changed from the last commit, which passed all checks, were the same files I changed with this commit. One error mentions something about explosions and says something about "teleport"

PR #63978, opened 2023-03-05
Unify gun battery/ups/bionic energy consumption 

### LINKED ISSUE #63846: Finger-Mounted Laser CBM can keep firing with no power
### Describe the bug

The Finger-Mounted Laser CBM consume 30kj per shot. However, if you run out while firing, you can just keep shooting despite not having enough 

### Attach save file

Trivially reproducible

### Steps to reproduce

1. Install Finger-Mounted Laser and necessary Power Storage/Generation CBMs
2. Generate at least 30kj of Bionic power
3. Spawn a bunch of mobs to shoot at
4. Active the finger laser and just start shooting

### Expected behavior

Laser is not able to fire when you don't have enough power for a shot.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19045.2604 (22H2)
- Game Version: f1bc060 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Sky Island [skyisland]
]

### Additional context

_No response_

Hirmuolio:
#### Summary
Infrastructure "Unify gun battery/ups/bionic energy consumption"

#### Purpose of change

Fix: #63846

It was not possible for gun to consume only bionic power without roundabout "it consumes ammo/ups but actually bionic power".

#### Describe the solution

* Create `"energy_drain"` property.
  * `"energy_drain"` consumes electric energy. By default from battery but can also consume ups/bionic with appropriate flags.
* Change `"ups_charges"` to `"energy_drain"` with appropriate flags.
* Change bionic guns to use `"energy_drain"`. The energy consumption of the bionic now does nothing (use for display purposes only). The gun itself consumes bionic power.

There are some oddities with guns/tools behaving in different ways. I really do not want to touch the code that handles UPS consumption in crafting. As long as you do not use battery as ammo for guns it shouldn't be a problem.

Guns that use only electricity from battery require JSON that looks like this:
* Pocket that holds the battery.
* `"ammo": [ "battery" ]` (to avoid complaints about containing ammo that is not valid for this gun)
* `"ammo_to_fire": 0,` (because we aren't actually using any ammo)
* `"energy_drain": "30 kJ",` (how much energy per shot to consume)

Guns that use battery as ammo should not be created or used. That will cause problems if anyone ever tries to mod them with energy related mods.

Make `"energy_drain"` work on tools. But that would involve crafting code.

Disallow using battery as ammo directly on guns.

Rewrite ammo checks to allow energy powered guns to be loaded with batteries without listing batteries as ammo for them.

#### Testing

H&K G80 railgun, M134D-H Minigun and A7 laser rifle fire and uses UPS power. Do not fire without UPS.

Laser finger fires and uses bionic power.

Mechs fire their weapon and consume mech power. Do not fire with empty battery.

Vehicle mounted energy consuming guns consume energy and fire. Do not fire with insufficient charges.


doc/JSON_INFO.md
```diff
@@ -3423,7 +3423,7 @@ Guns can be defined like this:
 "blackpowder_tolerance": 8,// One in X chance to get clogged up (per shot) when firing blackpowder ammunition (higher is better). Optional, default is 8.
 "min_cycle_recoil": 0,     // Minimum ammo recoil for gun to be able to fire more than once per attack.
 "clip_size": 100,          // Maximum amount of ammo that can be loaded
-"ups_charges": 0,          // Additionally to the normal ammo (if any), a gun can require some charges from an UPS. This also works on mods. Attaching a mod with ups_charges will add/increase ups drain on the weapon.
+"energy_drain": "2 kJ",    // Additionally to the normal ammo (if any), a gun can require some electric energy. Drains from battery in gun. Use flags "USE_UPS" and "USES_BIONIC_POWER" to drain other sources. This also works on mods. Attaching a mod with ups_charges will add/increase ups drain on the weapon.
```
ehughsbaird, line 3426:
```suggestion
"energy_drain": "2 kJ",    // Additionally to the normal ammo (if any), a gun can require some electric energy. Drains from battery in gun. Use flags "USE_UPS" and "USES_BIONIC_POWER" to drain other sources. This also works on mods. Attaching a mod with energy_drain will add/increase ups drain on the weapon.
```

src/item.cpp
```diff
@@ -11015,21 +11050,66 @@ int item::ammo_consume( int qty, const tripoint &pos, Character *carrier )
         qty -= charg_used;
     }
 
-    // Consume UPS power from various sources
+    // Modded items can sonsume UPS/bionic energy instead of ammo.
```
ehughsbaird, line 11053:
```suggestion
    // Modded items can consume UPS/bionic energy instead of ammo.
```

data/json/bionics.json
```diff
@@ -588,9 +588,10 @@
     "name": { "str": "Chain Lightning" },
     "description": "Your body is equipped with a chain lightning generator, allowing you to emit a blast of lightning at a target, leaving a trail of lightning in its wake, jumping to additional targets within 4 tiles of the previous target.",
     "occupied_bodyparts": [ [ "torso", 16 ], [ "arm_l", 4 ] ],
+    "//": "Firing cost determined by item. This number is only for display.",
```
Fris0uman, line 591: It's a bit weird to have a value that doesn't do anything, couldn't the display code get the value from the item too instead of relying on manual input? This seem very brittle

data/mods/Xedra_Evolved/items/guns.json
```diff
@@ -86,11 +86,11 @@
     "dispersion": 358,
     "durability": 8,
     "modes": [ [ "DEFAULT", "burst", 10 ] ],
-    "ups_charges": 1,
+    "energy_drain": "1 kJ",
     "reload": 200,
     "default_mods": [ "bars" ],
     "valid_mod_locations": [ [ "accessories", 4 ], [ "sights", 1 ], [ "rail mount", 1 ] ],
-    "flags": [ "MOUNTED_GUN" ],
+    "flags": [ "MOUNTED_GUN", "USE_UPS" ],
```
Maleclypse, line 93: I have tried to follow along but I want to ask to make sure I understand.   "USE_UPS" is what allows the turret to draw energy from the vehicle battery when it fires?   This is a turret only gun so I want to make sure all it needs is to be on a vehicle turret mount, with a battery with sufficient energy and not a UPS.

PR #63987, opened 2023-03-06
Delete CRT

Maleclypse:
#### Summary
None

#### Purpose of change
Removes CRT mod from the repo

#### Describe the solution
Deletes the mod folder


MNG-cataclysm:
If I may ask, why is CRT getting removed? It seems to have had some recent activity in terms of development.

MNG-cataclysm (approved):

PR #64004, opened 2023-03-06
Made layering inside handle sub limbs

### LINKED ISSUE #60693: Missing encumbrance warning in item layering menu
### Describe the bug

When wearing a sheath (strapped layer) inside your knee pads (outer layer), even though the encumbrance penalty is applied properly, neither item is highlighted yellow in the item layering menu (`+`) and neither item shows the warning message in the middle column (I would expect it to say something like `wearing your knee pads outside your sheath is adding encumbrance there`).

Layered correctly: 
cata-layer0

Mislayered (sheath selected): 
cata-layer1

Mislayered (knee pads selected):
cata-layer2

### Steps to reproduce

1. Make a new character.
2. Remove all clothing from your legs.
3. Wear a sheath and knee pads.
4. Observe no penalty when the sheath is worn on the outside.
5. Observe a penalty, but no highlighting or explanation, when the sheath is worn on the inside. 

### Expected behavior

Whenever there's an extra encumbrance penalty, I expect the relevant items to be highlighted yellow in the left column, and the reason described in the middle column.

### Screenshots

_No response_

Linux, last commit is cb944bfe7b5bfbf08296133b7014f412d43d3bbf.

### Additional context

Issue #57888 is also be related as it also deals with layer conflicts.

bombasticSlacks:
#### Summary
None

#### Purpose of change
fixes #60693 which was functionality that just didn't exist but should

#### Describe the solution
Made it so that instead of checking per sublimb only sometimes in the conflicts checking code, always check for sublimbs and then make sure no redundant penalties are added.

I also made sublimbs mandatory so that some redundant code could be removed.

#### Testing
Tested it fixes the bug, unit tests will catch any other issues.

### DIFF
--- a/data/core/sentinels.json
+++ b/data/core/sentinels.json
@@ -16,7 +16,8 @@
     "side": "both",
     "base_hp": 60,
     "drench_capacity": 1,
-    "legacy_id": "NUM_BP"
+    "legacy_id": "NUM_BP",
+    "sub_parts": [ "sub_limb_debug" ]
   },
   {
     "id": "mx_null",
--- a/data/json/body_parts.json
+++ b/data/json/body_parts.json
@@ -794,15 +794,24 @@
     "temp_mod": [ 500, 500 ],
     "technique_encumbrance_limit": 10,
     "bmi_encumbrance_threshold": 15,
-    "bmi_encumbrance_scalar": 0.5
+    "bmi_encumbrance_scalar": 0.5,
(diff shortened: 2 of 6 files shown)


sparr:
clang-tidy failure was caused by another PR merged shortly before this Pr was tested and has since been fixed.
windows build failure is from vcpkg problem also since fixed.

`encumbrance_test` failure in general build matrix might be an actual problem

PR #64011, opened 2023-03-06
bump ubuntu image and minimum compiler versions

kevingranade:
#### Summary
None

#### Purpose of change
Github is in the process of EOLing ubuntu 18.04 https://github.com/actions/runner-images/issues/6002

In particular, we are seeing brownouts from github actions using this image, which causes waves of basic build failures across all PRs.

We're also already eyeballing newer compiler features, i e. C++17 https://github.com/CleverRaven/Cataclysm-DDA/pull/63970

### DIFF
--- a/.github/workflows/matrix.yml
+++ b/.github/workflows/matrix.yml
@@ -89,7 +89,7 @@ jobs:
             native: linux64
             archive-success: basic-build
             dont_skip_data_only_changes: 1
-            title: Basic Build and Test (Clang 7, Ubuntu, Curses)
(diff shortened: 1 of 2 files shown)


kevingranade:
This is blocked from merging because of branch protection rules that say "Basic Build and Test (Clang 6, Ubuntu, Curses)" must pass, and this changes the name of the test.
To push this change I'll have to disable that rule for a while until all the PRs that have already passed that test have been merged,  then we can turn it back on using the new name.

Issue #25285, opened 2018-08-31
The bottle jack is rated at 8 tons but cant be used to change a tire on a 10 ton vehicle.


Androwski3000:
**Describe the bug**  
As the headline says it. I worked as mechanic in at a John Deere dealership. A 8 ton jack can lift a 10 ton vehicle just fine, as it only has to lift ~50% of the weight on that specific axle. It does not have to lift the complete vehicle into the air LOL.
So if weight distribution between axles is 70% / 30 % that´s 7 tons on the heaviest axle. 7 tons /2 is 3,5t per wheel/side.
If you have bad luck, the weight on the side you want to change that wheel, you have to jack up 5 tons. (Like on a loaded trailer which dug a wheel into mud and then this wheel went flat, while the load like corn or wheat moved to the flat side.)

**To Reproduce**  
1. Get a 10 ton vehicle.
2. Try to change a tire on the vehicle with bottlejack in inventory.

**Expected behavior**  
With the bottlejack to change a tire is just EZ.

**Screenshots**  

![bug with veh tile 2](https://user-images.githubusercontent.com/42423698/44934916-03b57980-ad6f-11e8-8263-2f3ee4e8158d.jpg)
Check the weight at bottom left.

Androwski3000:
This could be made skill based. So a noob mechanic reads the limit on the jack and things: "Now way we gona fix thge tire with that." Whereas the skilled mechanic asks the noob to go aside while he puts his shirt sleeves up.

CoroNaut:
Technically if it has four wheels, you only need to offset <3 tons of weight to make up for the wheel. If it has more wheels, then it needs to offset even less weight.

Night-Pryanik:
No discussion in over half a year, closing.

PR #25193, opened 2018-08-28
Self Jacking and Self Lifting and Crane Improvements

mrkybe:
#### Summary
```SUMMARY: Content "Adds a vehicle mounted jack that provides jacking for that vehicle only.  Changes Lifting to be provided to the vehicle its mounted on, allowing a crane to lift the engine into a vehicle its mounted on.  Adds motorcycle kickstand."```

#### Purpose of change
Provides code to support fully fixing ```#22547```.

#### Describe the solution

**Self Jacking**
- When trying to install or remove a wheel, check the vehicle being operated upon for a part with the `SELF_JACK` flag.
- Convert this to the corresponding item that has `SELF_JACK` quality, then allow that to replace `JACK` quality if necessary.
- This means that for a part with `SELF_JACK` to be used, it must be attached to the vehicle it is being used on.

**Self Lifting**
- Previously, LIFT quality was cached when you first interacted with a vehicle.  It is now updated whenever you move the cursor over the vehicle frame.
- Cranes now provide LIFT to every point they can reach within 4 tiles, though are blocked by obstacles like boards or quarter panels.

#### Describe alternatives you've considered
None?  ~~Though it would be nice to extract the jacking/lifting calculation to a function, as it had to be placed in two places.~~  Done.

#### Additional context
There's also no crafting recipe for it or for the motorcycle kickstand.

Credit to @Rod995 for the Air Jack System JSON.

Closes https://github.com/CleverRaven/Cataclysm-DDA/issues/22547

PR #64015, opened 2023-03-07
Self-jacking of vehicles with multiple installed jacks

cake-pie:
#### Summary
Features "Self-jacking of vehicles with multiple installed jacks"

#### Purpose of change
Self-jacking code checks for a single self-jacking part capable of supporting the entire vehicle. This is unrealistic, such systems in RL (most commonly seen in motorsports) involve multiple pistons (usu. 3 to 4) per car.

Likewise, the air jack has `SELF_JACK` quality 17, This allows a *single* jack to support *any* vehicle up to *infinite weight*, which is not reasonable.

<details>
<summary>derivation</summary>
Each level of jacking is equiv to 500kg, 17 × 500kg = 8500kg

https://github.com/CleverRaven/Cataclysm-DDA/blob/a0af8a61227b08914ed42e07eb59a2916efece24/src/item.h#L3029-L3030

For jacking purposes, vehicle mass is capped at 8500kg, anything heavier is treated as 8500kg.

https://github.com/CleverRaven/Cataclysm-DDA/blob/a0af8a61227b08914ed42e07eb59a2916efece24/src/veh_interact.cpp#L110-L119
</details>

#### Describe the solution
Modify the self-jacking feature use take the sum of the self-jacking capacity of installed parts. Currently, this doesn't require realistic load distribution so it still works even if you installed all the jacks in one corner of the vehicle.

Rebalancing of `SELF_JACK` capable parts' capacity:

- Air jack part changed, will now support up to 1000kg each. This is still a bit high, as a bit of research into RL systems suggests that the value should be around 600~700kg.

- Motorcycle kickstand jacking capacity looks okay.

#### Testing
One air jack is insufficient for changing the tires on a sports car, but two is enough.

Handy test file: [Sabine-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10905345/Sabine-trimmed.tar.gz)

#### Potential future work
- Refine air jack capacity down further to around 600\~700kg -- this would better reflect RL values, and a typical car would then require 3\~4 jacks, which is what you'd expect. Doing this would involve changing `TOOL_LIFT_FACTOR` to allow finer granularity (currently in increments of 500kg) and thus requires rescaling of all `JACK`/`LIFT` tool qualities, and I don't feel like tackling that right now.

- `SELF_JACK` is both a flag and a tool quality. Get rid of the redundancy.

- More realistic load-distribution remains an aspirational goal. This was originally discussed in #25193 and would solve issues such as #25285 -- in practice, you don't need to lift the entire weight of a vehicle to change a single tire.

### DIFF
--- a/data/json/items/vehicle/engineering.json
+++ b/data/json/items/vehicle/engineering.json
@@ -68,7 +68,7 @@
     "price": 8500,
     "price_postapoc": 500,
     "volume": "8750 ml",
-    "qualities": [ [ "SELF_JACK", 17 ] ]
+    "qualities": [ [ "SELF_JACK", 2 ] ]
   },
   {
     "type": "GENERIC",
--- a/data/json/vehicleparts/engineering.json
+++ b/data/json/vehicleparts/engineering.json
@@ -168,7 +168,10 @@
       { "item": "scrap", "count": [ 8, 24 ] },
       { "item": "jack", "count": [ 1, 1 ] }
     ],
-    "qualities": [ [ "SELF_JACK", 17 ] ],
+    "qualities": [ [ "SELF_JACK", 2 ] ],
+    "//0": "jacking ability equiv. to 1 ton. compare to:",
+    "//1": "600kg ea @ 12bar -- https://www.nukeperformance.com/category.html/air-jacks",
+    "//2": "675kg ea @ 30bar -- https://apracing.com/race-car/air-jacks",
     "requirements": {
       "install": { "skills": [ [ "mechanics", 2 ] ], "time": "60 m", "using": [ [ "welding_standard", 5 ] ] },
       "removal": { "skills": [ [ "mechanics", 2 ] ], "time": "30 m", "using": [ [ "vehicle_weld_removal", 1 ] ] },
@@ -191,6 +194,7 @@
     "description": "A kickstand to keep the bike from falling over.  You could use this to lean it forward or backward to change a tire.",
     "breaks_into": [ { "item": "pipe", "count": [ 1, 3 ] }, { "item": "scrap", "count": [ 1, 3 ] } ],
     "qualities": [ [ "SELF_JACK", 1 ] ],
+    "//": "jacking ability 500kg; heaviest motocycles are 400~500 kg",
     "requirements": {
       "install": { "skills": [ [ "mechanics", 2 ] ], "time": "60 m", "using": [ [ "welding_standard", 5 ] ] },
       "removal": { "skills": [ [ "mechanics", 2 ] ], "time": "30 m", "using": [ [ "vehicle_weld_removal", 1 ] ] },
--- a/src/veh_interact.cpp
+++ b/src/veh_interact.cpp
@@ -749,14 +749,11 @@ bool veh_interact::is_drive_conflict()
 
 bool veh_interact::can_self_jack()
 {
-    int lvl = jack_quality( *veh );
-
+    int lvl = 0;
     for( const vpart_reference &vp : veh->get_avail_parts( "SELF_JACK" ) ) {
-        if( vp.part().base.has_quality( qual_SELF_JACK, lvl ) ) {
-            return true;
-        }
+        lvl += vp.part().base.get_quality( qual_SELF_JACK );
     }
-    return false;
+    return lvl >= jack_quality( *veh );
 }
 
 bool veh_interact::update_part_requirements()


cake-pie:
> Heavy duty airjacks are commonly used in garages servicing trucks

You are referring to some kind of pneumatic/hydraulic floor jacks, which are *tools* with `JACK` quality


This is about self jacking *vehicle parts* with `SELF_JACK` quality


If you want to examine some actual specs, see:
https://github.com/CleverRaven/Cataclysm-DDA/blob/5cea35104fcbc0644956e5dfb7919c9ad7edab45/data/json/vehicleparts/engineering.json#L173-L174


> only to unload the wheel spindle just enough to take off/replace

This is a different problem that actually exists, see #25285

irwiss:
I don't understand the difference you're trying to show here, the truck jack you linked is just one type of jack with a manual lever you pump, there's pneumatic/hydraulic jacks that have air/oil valves and you hook them up with a compressor to pump, there's combos where you have both a valve and a lever

If the difference is "bolted/welded to the car in some way" - there's tons of those fixtures too that you can buy or walk into local workshop to do it for you, even easier in CDDA as aesthetics are of little concern so just welding a latch to hold it is enough, and you will definitely need to hook up the compressor somehow (well, unironically welding it in the case of existing air jack)

The fact weight distribution isn't simulated is an issue but adding up the jacks isn't the solution, dumping 100 scissor jacks under a heavy truck won't let you jack it outside of extreme conditions even if you somehow manage to put all 100 jacks on the tiny metal rib designed for jacking

cake-pie:
> I don't understand the difference you're trying to show here

`JACK` and `SELF_JACK` are completely different.
`JACK` is for tools. `SELF_JACK` relates to vehicle parts that can be used in lieu of `JACK` tools.
Code paths for checking whether you have a suitable `JACK` tool of enough tool quality vs whether the vehicle has sufficient `SELF_JACK` installed are completely independent of one another. Either one can be used to satisfy `NEEDS_JACKING`, but you cannot combine the two types.

This PR does not change anything about `JACK` functionality or any items or tools with `JACK` of any kind be it scissor jacks or bottle jacks or whatever.


> the truck jack you linked is just one type of jack with a manual lever you pump, there's pneumatic/hydraulic jacks that have air/oil valves and you hook them up with a compressor to pump, there's combos where you have both a valve and a lever

...and the point is that none of those are relevant to this PR.


> The fact weight distribution isn't simulated is an issue but adding up the jacks isn't the solution,

Weight distribution is a different issue and this PR is *not* an attempt to try to solve that.

Yes, you are right that IRL to change a single tire you need to "only to unload the wheel spindle just enough". In #25285 a professional mechanic complained that a 8-ton jack should suffice to change a tire on a 10-ton vehicle. But the current, preexisting behavior in the game for both `JACK` and `SELF_JACK` is to check that jacking capacity is at least equal to weight of the entire vehicle. Unrealistic? Yes, but solving that would boil down to computing the weight distribution between the jacking tool and the other remaining wheels that won't get lifted off the ground, to derive a value for how much load the jack must withstand. Nice aspirational goal, but not what this PR is about.

What this PR does:
It is "adding up the jacks" -- not for `JACK` tools but rather only for `SELF_JACK` vehicle parts, e.g. pistons.

The picture of the race car precisely exemplifies the kind of piston that the vehicle part `"id": "air_jack"` is supposed to be. Realistically these are installed as a set of three or four. Physics dictates that the load is distributed among all of the pistons. From the manual for the nuke performance product:

>  these air jacks will lift 402 KG (886 LB) each, and in a 4-jack system, it results in a full 1608 KG (3545 LB)

So they definitely work by summing together and are intended to be used this way. All installed pistons are used simultaneously to lift the entire vehicle, facilitating quick tire changes in motorsports settings.

With this PR, let's say you have a vehicle; installing one piston on each of its four corners will give it sufficient `SELF_JACK` for a vehicle up to 4 tons. Caveats:
- It also works if you installed all four pistons in one corner of the vehicle and use that to change the tire on the opposite corner. This can be seen as a flaw in the implementation, but it goes back to the weight distribution issue, so it's not a regression compared to status quo (one unrealistically strong piston in one corner allows changing opposite tire).
- the intent is to reduce the capacity per piston to 600\~700kg, a more realistic number. With 3 to 4 pistons you would get capacity of around 2\~2.5tons; it would still capable of lifting sports cars and most "normal" cars in the game. (You should not expect to lift an overbuilt custom battlebus with one piston.)

PR #64034, opened 2023-03-07
Fix asrgs turning into lamps

anoobindisguise:
#### Summary
Bugfixes "fix id conflict with ap_compact_asrg_containment"

#### Purpose of change
The vp and the ap had the same id, so while the vehicle part could be installed the appliance turned into a lamp.

#### Describe the solution
change the vehicle part to v_ instead of ap_


irwiss:
fwiw vehicle parts don't have a prefix only appliances do (cause they're effectively vehicle parts and share same std::map key space), don't think it matters that much but tilesets will have a bit odd `vp_v_...` prefix for their id

Maleclypse (approved):

PR #64043, opened 2023-03-07
make ads bionic power cost linear to damage

anoobindisguise:
#### Summary
Balance "linear instead of quadratic relationship between damage and power cost for ADS bionic"

#### Purpose of change
ADS bionic power cost was reasonable for its high-end damage values (20-25 kJ to dampen a rifle bullet assuming 10% efficiency for the bionic) but it ended up being a bit too low for weaker attacks and thus the power cost became overly negligible for random fodder enemies.

#### Describe the solution
Make the relationship linear (instead of power cost being equal to 10 J * damage squared it is instead 500 J * damage) which keeps the same upper bound but raises the lower bound considerably which should make it more fair in terms of total power usage.

@Night-Pryanik your thoughts are appreciated.


Night-Pryanik:
Well, from "realism" standpoint, it's absolutely cool that low damage results in low energy cost. I think this bionic was _designed_ to be almost impenetrable for low damage attacks, assuming one has enough energy for it.
But from balance standpoint, this is not so cool, as being almost completely invincible becomes boring very quickly.
And I'm on the balance side here.

anothersimulacrum:
For bullets at least, damage is the sqrt of kinetic energy, so squaring the damage for energy cost seems appropriate.

kevingranade:
What kind of attacks are we talking about that are trivialized by ads but not trivialized by armor?

Maybe we can make things more interesting by ADS having more selective application, like perhaps it has different costs for different types of attack, ir even non functional against some types.

anoobindisguise:
> What kind of attacks are we talking about that are trivialized by ads but not trivialized by armor?
> 
> Maybe we can make things more interesting by ADS having more selective application, like perhaps it has different costs for different types of attack, ir even non functional against some types.

So the big one is chaff zombies, who hit you for ~12 damage, ADS gives you 50% reduction on bash damage so it takes you from 6 to 12 damage. Add for example the bionic armor suite (5 bash total) or some other low-level armor, and it's significantly easier to hit 6 bash than 12 bash. Cut is reduced to 1/3, so claw/bite attacks are also easier to ignore. If each zombie punching you took away 6 kJ of energy instead of 144j it would at least come with a significant drawback for using it to tank. It could maybe be changed to not even effect bash or have a tiny effect against it though.

PR #64046, opened 2023-03-07
Fix SDL audio glitching with steaming hot patches from Qrox

### LINKED ISSUE #63765: Police car's sound is distorted (@'s soundpack)
### Describe the bug

Sometimes CDDA plays strange clicking noises during its sound playback. Vormithrax encountered this problem a bunch of times on his streams. I can now repro this problem reliably in any police car when I drive.

### Attach save file

[Boulder Junction-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10824351/Boulder.Junction-trimmed.tar.gz)

### Steps to reproduce

1. Enter any police car
2. Turn on the engine
3. Wait a couple of sound loops
(tried with @'s soundpack)

### Expected behavior

I expect the sound to not be distorted and to not produce clicks.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19044.1706 (21H2)
- Game Version: c69ea7c [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

This started happening about a month ago, so the problem has to be somewhat recent

akrieger:
#### Summary
None

#### Purpose of change
Fixes (for MSVC only) #63765. An SDL change last year (https://github.com/libsdl-org/SDL/commit/111c3add738587f97f41f8eaea35c6ee0d084ca5) introduced audio glitches/clicking when resampling audio files like in @'s soundpack. I identified https://github.com/libsdl-org/SDL/commit/1e5e8e2fda3796e76e6f7b1c39683925a3e9fed9 as a possible mitigation in https://github.com/libsdl-org/SDL_mixer/issues/466. @Qrox then wrote a better fix and put it up as a PR in https://github.com/libsdl-org/SDL/pull/7435. I tested CDDA building against SDL with that patch and it seems to work without introducing more regressions.

#### Describe the solution
I forked SDL's github repo and backported the fix commit to the 2.26.x release branch. I created a new tag based off the most recent one (created in the last day) which includes the fix. I then created a 'portfile overlay' inside the CDDA repo which contains a copy of the portfile from vcpkg, based off the vcpkg commit we are now using, except pointed at the patched release in my SDL fork. This then builds against the patched SDL.

Hosting a vcpkg fork to use as an alternate repository, and adding the portfile patch as a proper commit in that vcpkg repository. This would involve two forks to manage, and honestly the repositories feature is meant more for internal packages than patching upstream ones. 'Rebasing' the fix over future sdl fixes would be extremely marginally easier this way, but twice the forks is twice the pain, for just one package.

#### Testing
Downloaded @'s soundpack, loaded the save from #63765, reproduced the glitching sound. Build with my fork, did not reproduce the glitching sound.
```
1>Building sdl2[base,core]:x64-windows-static...
1>-- Installing port from location: C:\open\personal\Cataclysm-DDA\msvc-full-features\../.github/vcpkg_ports\sdl2
1>-- Downloading https://github.com/akrieger/SDL/archive/release-2.26.4_patched2.tar.gz -> akrieger-SDL-release-2.26.4_patched2.tar.gz...
1>-- Extracting source C:/open/personal/vcpkg/downloads/akrieger-SDL-release-2.26.4_patched2.tar.gz
```

### DIFF
--- a/.github/vcpkg_ports/sdl2/portfile.cmake
+++ b/.github/vcpkg_ports/sdl2/portfile.cmake
@@ -1,0 +1,130 @@
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO akrieger/SDL
+    REF release-2.26.4_qrox_math
+    SHA512 aa8c0552ac92bbf85dd3f9339088ebd41e0d09b1d46492462b47b632bab5d850c62b7938950accbfb51798b9f5664c1896fc75bdb8dd42b2c1cee84eebd59e13
+    HEAD_REF main
+    PATCHES
+        deps.patch
+)
+
+string(COMPARE EQUAL "${VCPKG_LIBRARY_LINKAGE}" "static" SDL_STATIC)
+string(COMPARE EQUAL "${VCPKG_LIBRARY_LINKAGE}" "dynamic" SDL_SHARED)
+string(COMPARE EQUAL "${VCPKG_CRT_LINKAGE}" "static" FORCE_STATIC_VCRT)
+
+vcpkg_check_features(OUT_FEATURE_OPTIONS FEATURE_OPTIONS
+    FEATURES
+        vulkan   SDL_VULKAN
+        x11      SDL_X11
+        wayland  SDL_WAYLAND
+        samplerate SDL_LIBSAMPLERATE
+        ibus     SDL_IBUS
+)
+
+if ("x11" IN_LIST FEATURES)
+    message(WARNING "You will need to install Xorg dependencies to use feature x11:\nsudo apt install libx11-dev libxft-dev libxext-dev\n")
+endif()
+if ("wayland" IN_LIST FEATURES)
+    message(WARNING "You will need to install Wayland dependencies to use feature wayland:\nsudo apt install libwayland-dev libxkbcommon-dev libegl1-mesa-dev\n")
+endif()
+if ("ibus" IN_LIST FEATURES)
+    message(WARNING "You will need to install ibus dependencies to use feature ibus:\nsudo apt install libibus-1.0-dev\n")
+endif()
+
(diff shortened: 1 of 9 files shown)


.github/vcpkg_ports/sdl2/portfile.cmake
```diff
@@ -0,0 +1,130 @@
+vcpkg_from_github(
+    OUT_SOURCE_PATH SOURCE_PATH
+    REPO akrieger/SDL
+    REF release-2.26.4_patched2
+    SHA512 885243493d981eeeb87c116a9ad6c9ef66dd6f8fcc46023e6c539b72b8d454c051ce1a8d5a083b0494d888e0ea8a987ede7c0794f38bde8aaf8e7ae615740bdc
```
akrieger, line 5: These are the only changed lines compared to clean vcpkg.

Qrox:
> An SDL change last year introduced audio glitches/clicking in the msvc build

Is this bug present in the MinGW build? Asking because the MinGW CI build is recently updated to SDL 2.26.2.

akrieger:
Probably. I'll try to track down exactly what I (believe) introduced it based on what seems to fix it.

akrieger:
Yep, confirmed its in the mxe cross build.

Qrox:
I'm trying a fix that uses integer arithmetics to fix this bug. I'll let you know if it works.

akrieger:
In the meantime, should we take this anyway?

Qrox:
I think my fix works. The clicking sound does not appear and a unit test shows that the signal-to-noise ratio of resampling increased from ~25 dB (using float or double) to ~50dB and the maximum error decreased from ~100% (using float) and ~5% (using double) to ~0.25%. I'll post a patch soon after I clean up the code.

PR #64047, opened 2023-03-07
Update labeler to catch another file for EOC label

Maleclypse:
#### Summary
None

#### Purpose of change
Adds the conditions.c++ file to the labeler but with a broader brush.

#### Describe the solution
@casswedson  please let me know if I'm doing this right.

None


casswedson:
as the totally, 101% expert in the subject at hand I approve of this good and sound addition! :D

PR #64054, opened 2023-03-08
Add a pepperbox revovler

actually-the-artificer:
#### Summary
Content "Adds the pepperbox revolver"

#### Purpose of change
I wanted a new gun that could be crafted; this seemed like an obvious candidate.

#### Describe the solution
Added a Pepperbox revovler.

Not adding it.

#### Testing
This was part of a personal Mod I made; it works fine there.

#### Additional context
Need to figure out a crafting recipe and spawns

### DIFF
--- a/data/json/items/gun/flintlock.json
+++ b/data/json/items/gun/flintlock.json
@@ -127,5 +127,39 @@
     "price_postapoc": 2500,
     "relative": { "range": 2 },
     "proportional": { "dispersion": 0.5, "reload": 1.1 }
+  },
+  {
+    "id": "pepperbox_revolver",
+    "copy-from": "pistol_base",
+    "type": "GUN",
+    "name": { "str_sp": "pepperbox revolver" },
+    "description": "A crude pepperbox revolver made out of scrap.",
+    "weight": "850 g",
+    "volume": "300 ml",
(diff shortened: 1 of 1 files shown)


anoobindisguise:
-4 damage might actually be too low given the flintlock pistol is 345mm and has +2, six more damage than the pepperbox - barrel length might be lower, but it would be worth comparing how long you expect the barrel to be in practice compared to the existing flintlock guns.

Is there a way to make it resistant to blackpowder fouling? It has multiple barrels, but as it stands it will sustain standard fouling rates. This would probably require additional c++ infrastructure but it would be nice for some other guns like derringers and double-barrrel shotguns to represent them having more barrels and thus needing more bullets total to foul.

actually-the-artificer:
> Is there a way to make it resistant to blackpowder fouling? It has multiple barrels, but as it stands it will sustain standard fouling rates. This would probably require additional c++ infrastructure but it would be nice for some other guns like derringers and double-barrrel shotguns to represent them having more barrels and thus needing more bullets total to foul.

Hmm. I dunno. That can be done another time.

There would be two types of pepperbox pistol: the crude one and the fancy antique one. The crude one is easily craftable, the antique one spawns in museums and such.

PR #64056, opened 2023-03-08
Refine criteria that prohibit vehicle modification when moving

cake-pie:
#### Summary
Bugfixes "Refine criteria that prohibit vehicle modification when moving"

#### Purpose of change
This is the current check for whether a vehicle is moving before allowing modifications (e.g. adding / removing parts)
https://github.com/CleverRaven/Cataclysm-DDA/blob/524ad226a60197be93e9edf7e3f5a87ff221024c/src/veh_interact.cpp#L712-L714

It was adapted from the "Dive from moving vehicle?" check, so apparently a small enough non-zero velocity is fine! This should not be the case.

Meanwhile the check for whether the player is controlling the vehicle was intended as a guard against acceleration via cruise control. This isn't sufficient, especially also considering the development of autopilot, autodrive, and eventually looking toward things like NPC driving.

Here are a couple corner cases I can think of that would pass, but shouldn't:
- Drive a vehicle, get it up to a relatively high speed. Stop controlling the vehicle while it is moving; vehicle will continue to coast uncontrolled while gradually slowing down. Per the criteria, you'd be allowed to remove a tire etc, once it reaches a sufficiently low speed, but before it has come to a complete stop. (The window of opportunity to do this will depend on deceleration rate.)
- Some self-driving vehicle (auto-farm or auto-patrol) is moving sufficiently slowly, and you are able to get alongside it and `e`xamine as it passes by...

#### Describe the solution
- require zero velocity
- player controlling vehicle check replaced by actual check on cruise velocity
- additional check to catch aircraft in stationary hover, since helicopters are a thing now
- update the messages displayed to player

One could argue for allowing some simple things such as clock or seatbelt. That would to need special casing to flag what kind of work is allowed, with checks such as you actually being on board the vehicle and will continue to be able to access the mount point while vehicle continues to move. Seems pretty low utility, and out of scope.

#### Testing
Absolutely no changes allowed while vehicle is in motion, or aircraft in stationary hover.

### DIFF
--- a/src/veh_interact.cpp
+++ b/src/veh_interact.cpp
@@ -709,7 +709,7 @@ task_reason veh_interact::cant_do( char mode )
             return task_reason::UNKNOWN_TASK;
     }
 
-    if( std::abs( veh->velocity ) > 100 || player_character.controlling_vehicle ) {
+    if( veh->velocity || veh->cruise_velocity || veh->is_flying_in_air() ) {
         return task_reason::MOVING_VEHICLE;
     }
     if( !valid_target ) {
@@ -1010,7 +1010,7 @@ void veh_interact::do_install()
                     msg = _( "It's too dark to see what you are doing…" );
                     return;
                 case task_reason::MOVING_VEHICLE:
-                    msg = _( "You can't install parts while driving." );
+                    msg = _( "You can't install parts while the vehicle is in motion." );
                     return;
                 default:
                     break;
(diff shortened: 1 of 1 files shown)


PatrikLundell:
Please check whether you can modify stationary vehicles that have been towed to their current location. Currently* there is a bug that sets the cruise speed of such vehicles to their max speed (or something like that) resulting in heavy acceleration when you start one unless you remember to immediately reduce their speed to something reasonable. It seems that a check for a non zero cruise speed might fail on such vehicles, resulting in them not being possible to be modified for no apparent reason (from the player's perspective).

* "Currently" = last time I played with towed vehicles, probably half a year ago.

cake-pie:
Oof, yes that's an inconvenient bug.

PatrikLundell:
I suspect you'd also end up with a non zero target speed if the engine is turned off while moving or otherwise dies (e.g. due to running out of fuel or getting damaged). Note that I haven't verified this is the case, though.

PR #64074, opened 2023-03-09
EOC triggers for OMT entry and exit

bombasticSlacks:
#### Summary
Features "EOC triggers for OMT entry and exit"

#### Purpose of change
Something that was requested by the core team. Allows for efficient EOC use without running into issues where you have to constantly check what OMT you are in and then fire. Instead you can directly attach an EOC to entry or exit.

#### Describe the solution
Added code where the avatar move stuff is tracked in game.cpp

If the OMT has entry or exit EOC, it builds the fake dialogue and runs them.

#### Testing
Briefly tested with some fake EOC's I didn't include here.

### DIFF
--- a/doc/OVERMAP.md
+++ b/doc/OVERMAP.md
@@ -245,6 +245,8 @@ rotation for the referenced overmap terrains (e.g. the `_north` version for all)
 | `mapgen_tee`      | Specify a C++ mapgen function for a LINEAR feature variation. Prefer JSON instead.               |
 | `mapgen_four_way` | Specify a C++ mapgen function for a LINEAR feature variation. Prefer JSON instead.               |
 | `eoc`             | Supply an effect_on_condition id or an inline effect_on_condition.  The condition of the eoc will be tested to see if the special can be placed.  The effect of the eoc will be run when the special is placed.  See [effect_on_condition.md](effect_on_condition.md). |
+| `entry_eoc`       | An effect on condition ID that will run when you enter this location.                            |
+| `exit_eoc`        | An effect on condition ID that will run when you exit this location.                            |
 
 ### Example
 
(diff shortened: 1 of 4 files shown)


doc/OVERMAP.md
```diff
@@ -245,6 +245,8 @@ rotation for the referenced overmap terrains (e.g. the `_north` version for all)
 | `mapgen_tee`      | Specify a C++ mapgen function for a LINEAR feature variation. Prefer JSON instead.               |
 | `mapgen_four_way` | Specify a C++ mapgen function for a LINEAR feature variation. Prefer JSON instead.               |
 | `eoc`             | Supply an effect_on_condition id or an inline effect_on_condition.  The condition of the eoc will be tested to see if the special can be placed.  The effect of the eoc will be run when the special is placed.  See [effect_on_condition.md](effect_on_condition.md). |
+| `entry_eoc`       | An effect on condition ID that will run when you enter this location.                            |
+| `exit_eoc`        | An effect on condition ID that will run when you exit this location.                            |
```
dseguin, line 249: Small oopsies, but it looks like the case doesn't match between the docs and the strings loaded in `assign` ("entry_eoc" vs. "entry_EOC")

bombasticSlacks: good catch

src/overmap.cpp
```diff
@@ -816,6 +816,8 @@ void oter_type_t::load( const JsonObject &jo, const std::string &src )
     assign( jo, "travel_cost", travel_cost, strict );
     assign( jo, "extras", extras, strict );
     assign( jo, "mondensity", mondensity, strict );
+    assign( jo, "entry_EOC", entry_EOC, strict );
```
bombasticSlacks, line 819:
```suggestion
    assign( jo, "entry_eoc", entry_EOC, strict );
```

PR #64080, opened 2023-03-09
[Mods] Replace "fake" activity levels with appropriate ones

RenechCDDA:
#### Summary
Balance "Placeholder activity levels for many mod recipes replaced with appropriate values"

#### Purpose of change
A lot of recipes were using the activity level fake. Notably, this does not mean NO_EXERCISE. I did DDA stuff in #64007 so now the mods get a chance.

#### Describe the solution
Grep grep grep grep grep, lots of copy-pasting the appropriate string in after thinking about how hard it is to do the thing.

Stare at aftershock's wall of unintelligible recipes and cry

#### Testing
Linted, I think

#### Additional context
I didn't touch the TEST_DATA mod because I believe that's used for automated tests? So I just figured, better off not touching that right now, I don't want to deal with potentially making a mess.

I gave all the named weapons from magiclysm `EXTRA_EXERCISE` because I thought it was thematically appropriate. Normally I was giving blacksmithing only `ACTIVE_EXERCISE`, but hey - you gotta put your blood sweat and tears into making something that great, right?


data/mods/Aftershock/recipes/bionic_recipes.json
```diff
@@ -345,7 +345,7 @@
   },
   {
     "type": "recipe",
-    "activity_level": "fake",
+    "activity_level": "LIGHT_EXERCISE",
```
Maleclypse, line 348: I'm thinking this should maybe be a moderate exercise?

Maleclypse: Same for all the bio armors

RenechCDDA: Hammers and wrenches? Yeah, I agree. Moderate it is.

anothersimulacrum:
Does this get everything? If so, we can get rid of loading support for fake after this.

RenechCDDA:
>Does this get everything? If so, we can get rid of loading support for fake after this.

It leaves the TEST_DATA stuff, as well as the NPC recipes and nesting (category) ones mentioned in my previous PR. Those would be easy to clean up and shouldn't need any *specific* attention to what they get replaced with.

PR #64083, opened 2023-03-09
Rework Judo's passive bonus, limbify recovering from being downed

Venera3:
#### Summary
Features "Limbify downed recovery, rework Judo's passive bonus"

#### Purpose of change

Judo's passive bonus conferred grab resistance and being immune to downing, the second of which is hilariously wrong and the first is debatable. Also, more limby things are better than less.

#### Describe the solution

- Added a limb score component to recovering from being downed
- Deprecated `is_throw_immune` (it hasn't meant being immune to throws for a good while now)
- Added the ability to define and check character flags in MA buffs (no, they are not effects. don't ask why)
- Added the `DOWNED_RECOVERY` flag granting a greatly increased chance to stand up while downed

Include other scores, but balance+lift works pretty well and the charmod readout is getting busy.
Use a lower recovery bonus for judo, but you do learn how to fall and get up pretty well.

#### Testing

Got downed unencumbered and wearing EOD gear, the latter left me breakdance for a good few turns without getting up.
The flag triggered the custom message and the bonus was applied.

#### Additional context

Falling and rolls are of course very much a part of Judo, the reason being *you spend a lot of time on the ground on account of being thrown.*

### DIFF
--- a/src/character_escape.cpp
+++ b/src/character_escape.cpp
@@ -42,7 +42,7 @@ void Character::try_remove_downed()
     /** @EFFECT_DEX increases chance to stand up when knocked down */
     /** @EFFECT_ARM_STR increases chance to stand up when knocked down, slightly */
     // Downed reduces balance score to 10% unless resisted, multiply to compensate
-    int chance = ( get_dex() + get_arm_str() / 2.0 ) * get_limb_score( limb_score_balance ) * 10. 0;
+    int chance = ( get_dex() + get_arm_str() / 2.0 ) * get_limb_score( limb_score_balance ) * 10.0;
     // Always 2,5% chance to stand up
     chance += has_flag( json_flag_DOWNED_RECOVERY ) ? 20 : 1;
     if( !x_in_y( chance, 40 ) ) {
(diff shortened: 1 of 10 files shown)


src/character_escape.cpp
```diff
@@ -38,12 +40,17 @@ void Character::try_remove_downed()
 {
 
     /** @EFFECT_DEX increases chance to stand up when knocked down */
-
-    /** @EFFECT_STR increases chance to stand up when knocked down, slightly */
-    if( rng( 0, 40 ) > get_dex() + get_str() / 2 ) {
+    /** @EFFECT_ARM_STR increases chance to stand up when knocked down, slightly */
+    // Downed reduces balance score to 10% unless resisted, multiply to compensate
+    int chance = ( get_dex() + get_arm_str() / 2 ) * get_limb_score( limb_score_balance ) * 10;
```
kevingranade, line 45:
```suggestion
    int chance = ( get_dex() + get_arm_str() / 2.0 ) * get_limb_score( limb_score_balance ) * 10. 0;
```

kevingranade: Tell clang tidy to hush

src/martialarts.cpp
```diff
@@ -953,6 +948,16 @@ bool ma_buff::is_stealthy() const
     return stealthy;
 }
 
+bool ma_buff::has_flag( const json_character_flag &flag ) const
+{
+    for( const json_character_flag q : flags ) {
```
kevingranade, line 953:
```suggestion
    for( const json_character_flag &q : flags ) {
```

Issue #63683, opened 2023-02-19
Faction Camp Starting Scenario


JonathanLochridge:
### Is your feature request related to a problem? Please describe.

I think it would be cool to have a starting scenario that starts you with a really basic form of faction camp in the early stages. Obviously, it would cost a significant number of starting points. I also think adding a time delay so that you start with the evolution clock pushed back further and such could also be good to help with the trade-off. It might also place some restrictions on the starting professions allowed to prevent it being too cheesy.
Potentially, putting it towards the end of summer or fall could provide a bit of a natural rush against time to make sure you can get enough food to feed everyone long term.

### Solution you would like.

Add a new starting scenario.

PR #64086, opened 2023-03-10
Add a faction camp starting scenario

MNG-cataclysm:
#### Summary
Content "Add a faction camp starting scenario"

#### Purpose of change
Do something like #63683 and add a starting scenario that places a character in a faction camp.

#### Describe the solution
The player begins in a evac shelter with two companion NPCs, and a lot of materials and tools outside for use in the camp. This start occurs in the next fall of the Cataclysm, providing a challenge to scavenge enough food on time for winter and keep building up. It also costs 8 points to choose, making it very expensive and limiting starting build options, save for taking negative traits.

Starting in a ruined faction camp or one the player made themselves, but only NPC dialogue can designate someplace as a faction camp.

#### Testing
Spawned in multiple times and tested everything out. It all works well and looks good.

#### Additional context
I tried placing the faction camp itself through the starting EOC, but it only works through NPC dialogue. As such, one of the NPCs will talk to you and prompt the creation of the camp before you leave the shelter.


Termineitor244:
Good!! But I think an 8 point cost for something like this is fairly excessive, I mean, if the camp was already somewhat developed (one or two rooms already built, the well and / or the radio tower already in there...) I could understand, but as it is right now, the scenario just gives you an extra npc and some construction materials/tools, with the downside of starting with evolved monsters some time in the future, no food collected as well as no special bonuses...

I think a 4 point cost is more reasonable for the scenario as it is, but still, the scenario could give you some other materials necessary for the camp, like rocks or some amount of planks, a small amount of food would be reasonable too, the player could do with not being expected to go search for food as soon as they start, since they seem to have prepared all of these materials for the purpose of establishing a base there, and some food for the actual construction of the place is to be expected, not enough for winter, but enough to prepare for an excursion to the nearest city in a week or so.

adamkad1:
A variant of this starting in an open field or such for innawoods would be nice too.

PR #64087, opened 2023-03-10
Fix bug with hats providing coverage uniformly across the entire head

### LINKED ISSUE #63570: Turtlenecks are not hats
### Describe the bug

'Turtleneck' and 'Turtleneck shirt' causes additional encumbrance when wearing hats (any head clothing or armor on the normal layer)

When rolled, these turtlenecks don't cover the mouth, but still somehow covers the head. 

### Attach save file

[Platte Woods-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10718308/Platte.Woods-trimmed.tar.gz)

### Steps to reproduce

1. Acquire a 'turtleneck shirt' or 'turtleneck shirt (rolled)
2. Acquire any article of clothing or armor that occupies the head (normal layer)
3. wear both
4. character is affected with extra encumbrance from wearing two items on the head (normal layer)

### Expected behavior

Turtlenecks should not occupy the same space/layer as a hat - especially if you roll the neck down. 

When worn correctly, a turtleneck should either cover the torso, arms and neck, or just the torso and arms when the neck is rolled down.

### Screenshots

image
image

- OS: Windows
- Game Version: 7216560 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
]

### Additional context

Recreated issue on a 'Start Now!' save, spawning a random hat, and both turtleneck items.
Used latest experimental version, but issue has existed for some time. Noticed several issues with other common clothing items and layering issues - including bras

strategineer:
Fix bug with hats providing coverage uniformly across the entire head by modifying/adding the `coverage` and `specifically_covers` fields in data\json\items\armor\hats.json to line up a little more with reality.

This also will resolve the issue with wearing turtlenecks with hats leading to additional (nonsensical) encumbrance due to both the turtleneck and the hat covering the throat (which as of this commit, is/should-not-be not the case for any hat except for the summer hard hat).

Fixes #63570

#### Describe the solution

I determined reasonable values for these fields by looking at pictures of the various hats online while also trying to maintain similar total coverage values to what was present previously in most cases to avoid rocking the boat too much.

NB: An exception to this was the tin foil hat. 90% coverage across the entire head seemed a little weird... Seemed more like a tin foil helmet.

None

#### Testing

For #63570, I spawned/wore a straw hat and turtleneck at the same time, took note of the "Wearing multiple items of normal clothing on your head is adding encumbrance there." message. Then made these changes, and noted the disappearance of that message as well as a reduction in the encumbrance due to the straw hat's coverage no longer overlapping with the turtleneck's.

Also, I used the debug menu, spawned each hat that was changed and made sure that the coverage values for each one made sense and were being displayed correctly through the UI.

Save with all the hats on the floor below the player: [Tallulah.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10938718/Tallulah.zip)

### DIFF
--- a/data/json/items/armor/hats.json
+++ b/data/json/items/armor/hats.json
@@ -17,7 +17,14 @@
     "material_thickness": 0.3,
     "environmental_protection": 1,
     "flags": [ "VARSIZE", "SUN_GLASSES" ],
-    "armor": [ { "encumbrance_modifiers": [ "NONE" ], "coverage": 50, "covers": [ "head" ] } ]
+    "armor": [
+      {
+        "encumbrance_modifiers": [ "NONE" ],
+        "coverage": 100,
+        "covers": [ "head" ],
+        "specifically_covers": [ "head_crown", "head_forehead" ]
+      }
+    ]
   },
   {
     "id": "bandana_head",
@@ -42,7 +49,14 @@
     "warmth": 5,
     "material_thickness": 0.1,
     "environmental_protection": 1,
-    "armor": [ { "encumbrance_modifiers": [ "NONE" ], "coverage": 80, "covers": [ "head" ] } ]
+    "armor": [
+      {
+        "encumbrance_modifiers": [ "NONE" ],
+        "coverage": 100,
+        "covers": [ "head" ],
+        "specifically_covers": [ "head_forehead" ]
+      }
+    ]
   },
   {
     "id": "beret",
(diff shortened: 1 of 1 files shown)


Termineitor244:
Thank you!! I absolutely forgot to update hats and helmets to the new sublimbs coverage system, I wanted to do it after adding hard hats modularity, but I kept getting distracted by other things, I'm glad someone else has taken the time to tackle this!! Good job!

PR #64102, opened 2023-03-10
Disassembly doesn't return items with UNRECOVERABLE flag

mqrause:
#### Summary
Bugfixes "Disassembly doesn't return items with UNRECOVERABLE flag"

#### Purpose of change

Not sure if I can say it's a fix, but it addresses #51172. Apart from `UNRECOVERABLE` there is also `NO_RECOVER` in recipes which isn't addressed by this.

#### Describe the solution

Skip items with the flag.

Not adding them to components in the first place, but I'm not sure if that could cause other issues.

#### Testing

Craft and disassemble hand-crank charger, doesn't return superglue.

### DIFF
--- a/src/crafting.cpp
+++ b/src/crafting.cpp
@@ -2806,6 +2806,9 @@ void Character::complete_disassemble( item_location &target, const recipe &dis )
     // Roll skill and damage checks for successful recovery of each component
     for( const item_components::type_vector_pair &tvp : components ) {
         for( const item &newit : tvp.second ) {
+            if( newit.has_flag( flag_UNRECOVERABLE ) ) {
+                continue;
+            }
             // Use item type to index recover/destroy tallies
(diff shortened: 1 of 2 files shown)


Karol1223:
I'm not good with C++ so I'll ask - does that tackle only ``reversible`` recipes or manually defined disassemblies as well? Because it really should only tackle the former. If an unrecoverable item is defined in a manual disassembly recipe then it most likely has a reason to be defined there.

mqrause:
So for one, yes, this only fixes the issue for reversible recipes. Manually defined disassembly recipes already properly respected the flag, so what you're describing isn't possible already. If an item could be recovered from disassembly, it simply shouldn't have that flag. Instead the recipes should use the currently dysfunctional `NO_RECOVER` flag.

misterprimus:
Glad this is finally fixed. This problem has been around for a long, long time.

PR #64107, opened 2023-03-10
Add Chemistry Proficiencies

Alm999:
#### Summary
Balance "Add chemical related proficiencies to crafting recipes"

#### Purpose of change

Currently, despite being "learnable", all chemical proficiencies have no actual opportunities to be learned. They are not used in any recipes.

This also effectively blocks "Biochemistry" weakpoint from being obtainable, as it requires "Organic Chemistry" knowledge (which can not be learned via in-game activities).

P.S. OK, I've learned my lesson. No more nested branches for each change.

#### Describe the solution

Simply adding the proficiencies to recipes presents opportunities needed to learn them. My discretion at adding the proficiencies was as follows:
1) if a recipe requires any manipulation on chemical stuff -> "Principles of Chemistry";
2) if a recipe mandates operations on inorganic substances (saltpeter, lye, even salt water counts) -> "Inorganic Chemistry";
3) if a recipe needs organic substances (benzene, citric acid, even ethanol counts) -> "Organic Chemistry" (substances need not to be complex, just being a product of hydrocarbons counts);
4) if any distillation process is required (both normal and "fine" variants) -> "Distillation" (maybe adjusting for difficulty of fine distillation is needed as a time modifier);
5) if a product of a recipe is not an "end product" (like "superglue") but intended to be used as an ingredient for more complex chemical synthesis () -> "Principles of Chemical Synthesis";
6) for any high level mutagens -> "Complex Chemical Synthesis";
7) if a product is a drug meant for consumption -> "Pharmaceutical Production";
8) if a recipe requires any manipulation on a (recently) living creature including chemical extraction from its internal organs -> "Biochemistry".

Maybe it is a good idea to also add training recipes provided by chemistry textbooks.

#### Testing
Tested some (not all, bobviously) recipes in-game. Proficiencies successfully added.

#### Additional context

None.


Drew4484:
I like this concept. Practice recipes are a good idea even though you've been very thorough and added plenty of opportunities to learn each proficiency.

onura46:
Good stuff! Great work! I haven't checked every recipe, but I searched for a couple I knew should be included and they were. I appreciate your comments as well.

I feel like recipes in `data\json\recipes\food\distill.json` and `data\json\recipes\weapon\explosive.json` could probably use the same love. I'm a bit too ignorant about chemistry stuff to do that confidently myself, and there's no obligation to do that in this PR. But maybe down the road if you feel inclined. You already covered the big files.

Alm999:
> I feel like recipes in `data\json\recipes\food\distill.json` and `data\json\recipes\weapon\explosive.json` could probably use the same love. I'm a bit too ignorant about chemistry stuff to do that confidently myself, and there's no obligation to do that in this PR. But maybe down the road if you feel inclined. You already covered the big files.

The plan is to review the whole crafting database, yes. Starting from big hitters and slowly moving towards obscure places (like, some food handling recipes). :)

Aww, what I'm getting myself into… and all I wanted is to learn some Biochemistry…

data/json/recipes/recipe_medsandchemicals.json
```diff
@@ -603,6 +718,10 @@
     "autolearn": true,
     "batch_time_factors": [ 90, 4 ],
     "//": "reversal of the recipe above",
+    "//": [
```
Brambor, line 721:
```suggestion
    "//1": [
```

Alm999:
OK, that should be most of them.
I think it is ready for initial reviews.

data/json/recipes/other/power_supplies.json
```diff
@@ -763,6 +770,7 @@
     "skill_used": "mechanics",
     "time": "2 h 40 m",
     "autolearn": true,
+    "proficiencies": [ { "proficiency": "prof_intro_chemistry" } ],
```
onura46, line 773: I think this and the next three `methanol_tank` recipes just simulate tediously refilling a tank with a syringe, based on what I read in the original PR. Not sure if you'd be learning anything, but it's really your call.

onura46:
There's some things I'm not knowledgeable enough to make a judgement call on, but everything I do know about here seems pretty good to go. Great audit!

data/json/recipes/other/power_supplies.json
Alm999: Well, it is true that there are no reactions involved here. But methanol is a highly toxic chemical, so I think this activity can at least train a character safety measures and handling of chemical equipment (a syringe in this case).
I've assigned "Principles of Chemistry" to lots of such activities, and have given "(In)organic Chemistry" only to recipes in which actual reactions are taking place.

At least, that was my reasoning at the end of the process. Of course, the work started when lack of proficiencies actually penalized a character, so having un-learned "Principles of Chemistry" in a recipe was in fact worse than having no proficiency at all. It all changed with recent patches, so I understand that now some may see the addition as an attempt at aiding player (and we all know, player must suffer more :D , not less).

onura46: That all sounds good to me! That division between proficiencies makes sense. 👍

Alm999:
Now, that the zinc production recipes have been changed by https://github.com/CleverRaven/Cataclysm-DDA/pull/63375, I replaced chemistry-related proficiencies with the one I've intentionally avoided (as I intended to have a look at it after the chemistry was fixed) -- "Metallurgy".

Conflicts must be resolved, after all.

PR #64110, opened 2023-03-10
make rm13 armor a bit less strong

anoobindisguise:
#### Summary
Balance "tone down rm13 a little"

#### Purpose of change
I've tested RM13 on a few long term characters now and I'm feeling that I probably buffed it too much.

#### Describe the solution
Powering it on does not reduce encumbrance anymore. It had always worked like that, but the armor isn't assistive like the power armors so it probably should not be lowering encumbrance. Also, increased its power consumption from 1 charge per 6 turns to 250 W.
I also gave it unbreakable melee so that you won't accidentally destroy it if you try punching things while not wearing soft gloves on top of it (fingerless gloves, rubber gloves, light gloves etc to take the damage in its stead)

Unsure, but I felt that it was overcentralizing and overshadowed the phase immersion suit too much also. This should probably make it a bit less extremely good than it was.

### DIFF
--- a/data/json/items/tool_armor.json
+++ b/data/json/items/tool_armor.json
@@ -1189,7 +1189,8 @@
       "PADDED",
       "USE_UPS",
       "NO_UNLOAD",
-      "COMBAT_TOGGLEABLE"
+      "COMBAT_TOGGLEABLE",
+      "UNBREAKABLE_MELEE"
     ],
     "price": 50000000,
     "price_postapoc": 10000,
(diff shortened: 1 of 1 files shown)


SirPendrak:
> Powering it on does not reduce encumbrance anymore. It had always worked like that, but the armor isn't assistive like the power armors so it probably should not be lowering encumbrance.

I allways imagined that hyperweave is getting less stiff when powered

PR #64114, opened 2023-03-10
Gambesons rework (again)

anoobindisguise:
#### Summary
Content "Split gambesons into gambesons and aketons and model them more thoroughly"

#### Purpose of change
Gambesons were in need of some attention. They were overly thick for the normal layer, used ordinary cotton when better materials were available, and their creation process was a guess rather than one backed by data.

#### Describe the solution
Gambesons now follow a new paradigm. Heavily based on the work done in https://costumegirl.wordpress.com/2010/04/01/the-making-of-a-medieval-gambeson/.

1. All gambeson related items and recipes have been moved to their own individual JSON files.
2. Gambesons are split into Aketons (these are normal-layer only, supplementary armor, and are 3.5mm on torso, 3mm on arms and legs, 2mm on head and hands) and Gambesons (these are both normal and outer, standalone armor, and are 7.5mm on torso, 6mm on legs/arms, but their additional coverage over the aketon equivalents only has 95% instead of 100%). The 7.5mm gambeson mirrors the "30 layer jack" from historical documents as each canvas layer is 0.25mm.
3. There are wool, nylon, canvas and stuffed cotton varieties. Canvas is the heaviest, most encumbering, least breathable, and most protective, nylon is breathable, low encumbrance, bad vs bash, good vs ballistic, wool is a middle ground of protection aside from ballistic (its ballistic is poor), middle ground of encumbrance, moisture wicking, but also warmer. and not waterproof. Stuffed cotton is cheaply made and inferior in terms of protection while also being as heavy and encumbering as the canvas varieties. The existing gambesons and arming pants in game were made to be stuffed cotton. Kevlar gambesons have been obsoleted as there is too much overlap between the survivor jumpsuits and ballistic vests.
4. The material requirements were redone. Nylon was placed at 288 GSM and Canvas at 406 GSM to make their respective sheets each ~0.25mm. Body parts were estimated using the Rule of Nines and a total body square footage of 20 to give an accurate number of required 0.25mm 1x2' sheets. This base requirement was then increased by 20% or so to account for wastage from cutting the sheets into necessary shapes. The products require an additional "Fabric Quilting" proficiency and additional threading to quilt the sheets together. Byproducts are equal to the remaining weight of fabric and are divided into 75% scraps and 25% patches by weight. For Wool aketons/gambesons there was no easy sheet to work with as felt IRL is much less dense than sheet wool and is sold commercially in 1mm+ thickness sheets - its very low GSM compared to proper wool means it would make poor armor. Therefore these ones are made of actual quilts, assumed to be non-felted wool and thus would be of the necessary weight to make armor. The weight of these aketons/gambesons were set assuming they were made of hypothetical wool sheets with a density of 1.4 instead of 1.6 (canvas) or 1.14 (nylon), and the number of required quilts is equal to the minimum needed to match that number and the remainder weight becomes scraps/patches. Stuffed cotton gambesons are made of Sheets (easily available) and "gambeson batting" which is 50 grams of waste fabric or equivalent. This produces fairly heavy gambesons, but those weights line up with IRL figures (easily 10kg for a full gambeson IRL)
5. The time needed to make gambesons/aketons was increased to an order of multiple days as this is much more accurate to IRL production times. The required skill was also bumped up a bit (5 instead of 4 for wool/canvas and 6 for nylon). They are still autolearned however as there isn't any secret technique to their creation that wouldn't be familiar to a bookless seamster/ess. Joint Articulation was also added as a requirement where applicable.
6. Gambesons/aketons are made of a fake "x_quilted" material which like layered kevlar has higher defense than its constituent. this is necessary to make them protective to a level that matches IRL testing such as in my primary source. For Canvas this is 2/3/2, for Nylon this is 1.5/2/3, and for Wool/Cotton this is 2/2/1.5.
7. Aketons have Vest variants, while Gambesons have Vest and Thin-sleeved variants. There are no individual sleeves as those should typically need a device to anchor them to. 

Call Nylon Gambesons "Synthetic Fabric Gambesons" as despite being obviously nylon we still call the material synthetic fabric in game.
Add Denim Gambesons. These would be more breathable than Canvas gambesons but not waterproof and slightly less protective.
Make the recipes require wax as the source mentions waxing the needle/thread to make stitching faster. However wax is being reworked right now so I planned on holding off for the time being.

#### Testing
Loaded it up in game and checked the recipes and they seem to be working correctly.

#### Additional context
aketon


anothersimulacrum (changes requested):
> All gambeson related items and recipes have been moved to their own individual JSON files.

Either don't do this at all, or don't do it in this PR. It makes review impossible.

anoobindisguise:
> > All gambeson related items and recipes have been moved to their own individual JSON files.
> 
> Either don't do this at all, or don't do it in this PR. It makes review impossible.

I didn't realize this made it hard to review, but I believe this is 100% necessary because otherwise everything becomes way too cluttered and unnavigable.

Drew4484:
> > > All gambeson related items and recipes have been moved to their own individual JSON files.
> > 
> > 
> > Either don't do this at all, or don't do it in this PR. It makes review impossible.
> 
> I didn't realize this made it hard to review, but I believe this is 100% necessary because otherwise everything becomes way too cluttered and unnavigable.

I think the request is that if you do want to reorganize them, do that move in a separate PR so it can be easily blamed later.

anothersimulacrum:
It's a "do that move in a different PR so we can see what changes are here". Generally, we don't want to move data around. Moving data and changing it at the same time is pretty much flat-out not acceptable because of the difficulty in reviewing it.
If the author of the PR is well versed in git and wants to manage their commits, they can do a commit to move and then changes in follow-up commit(s), but this PR has 133 commits, so it's not possible to review in that way.

PR #62845, opened 2022-12-23
Cleanup prison_1.json

FarkasKBence:
#### Summary
Content "Minor prison mapgen update"

#### Purpose of change

prison_1.json is one of the largest files in `/mapgen`; it's 1636 lines long, and contains 3 locations, of which 2 are unpalettized.

#### Describe the solution

File changes:
* created a mapgen folder for prisons
* moved `prison_1,json` into the new folder
* split the file into 3, 1 for every prison
* created separate palettes for each prison and made a base palette

Mapgen changes:
* regionalized terrains
* added SUS itemgroups, expanded kitchen loot, overhauled kitchens
* added sinks to cells, trash cans to the land prison and Alcatraz
* added flagpoles to Alcatraz, standardized flag spawns
* overhauled land prison workstation with newer furniture
* added solitary cells and guard breakrooms to land prison
* expanded guard breakroom to Alcatraz
* removed most thin Unicode characters from prison_island
* replaced hardwood flooring with concrete
* moved item groups to a new file
* removed `SIDEWALK` tag from land prison

#### Describe alternatives you've considered

None.

#### Testing

Debug-spawned the 4 prisons. Peeked under furniture to check if they're on appropriate terrains.

#### Additional context

The palettes could've been optimized a bit more, but I was too tired to further mess with them.

For a follow-up PR: the watchtowers need to be a bit taller, and the land prison's road connection needs fixing, also the reception area should be expanded upon.

PR #64115, opened 2023-03-10
Cleanup after island prison cleanup

Night-Pryanik:
#### Summary
None

#### Purpose of change
#62845 brought several small issues to the island prison mapgen, like cooling unit standing on a barbed wire, or vending machine standing on a column, concrete floor in underground cavern instead of rock floor etc. But also several changes I don't like, like completely removing trees from the location, replacing metal benches with wood benches, replacing secret slope under the bookcase in one of the cells with stairs without the bookcase, and some other.

#### Describe the solution
Fixed most of the issues I found.

None.

#### Testing
Debug-spawned island prison location several times to check if everything is ok.

#### Additional context
None.


LovamKicsiGazsii:
Some details got lost in translation while untangling the behemoth that was prison_1.json. Thanks for catching these.

> like completely removing trees from the location

I actually think removing them makes perfect sense, as trees would block vision from the guard towers. But I get if you want them back for aesthetics reasons, I'm prone to doing these sorts of stuff myself.

> replacing metal benches with wood benches

I guess there's no reason why we wouldn't go 100% on this, replacing every wooden bench with metal ones in both the base palette and otherwise. From the images I've seen, US prisons don't seem to use wooden furniture around inmates.

Termineitor244:
> replacing secret slope under the bookcase in one of the cells with stairs without the bookcase

Huh, I was wondering why I found the secret escape route like that in the open. I thought it was a mapgen bug or something, glad to see it hidden again!

PR #64120, opened 2023-03-10
Improve modded weather

Ramza13:
#### Summary
None

#### Purpose of change
Modding weather was clunky and required overwriting regional_settings just to add or remove a type.

#### Describe the solution
Instead of having regional biome contain the order of weather loading add a parameter to weather to handle it.  Add a white list and black list to regional settings to make modding weather types easier.

### DIFF
--- a/doc/WEATHER_TYPE.md
+++ b/doc/WEATHER_TYPE.md
@@ -27,7 +27,7 @@ Each weather type is a type of weather that occurs, and what causes it. The only
 | `duration_max`                 | Optional, the upper bound on the amount of time this weather can last. Defaults to 5 minutes.
 | `weather_animation`            | Optional, Information controlling weather animations.  Members: factor, color and glyph |
 | `condition`                  | A dialog condition to determine if this weather is happening.  See Dialogue conditions section of [NPCs](NPCs.md)
-| `load_order`                  | An integer.  If the condition of multiple weather types are true the higher load order wins
+| `priority`                  | An integer.  If the condition of multiple weather types are true the higher load order wins
 |	`required_weathers`          | A string array of possible weathers, it is at this point in the loop. i.e. rain can only happen if the conditions for clouds light drizzle or drizzle are present.  Required weathers need to have lower load orders to be  |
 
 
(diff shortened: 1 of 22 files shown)


src/weather_type.cpp
```diff
@@ -110,6 +111,7 @@ void weather_type::load( const JsonObject &jo, const std::string & )
     mandatory( jo, was_loaded, "ranged_penalty", ranged_penalty );
     mandatory( jo, was_loaded, "sight_penalty", sight_penalty );
     mandatory( jo, was_loaded, "light_modifier", light_modifier );
+    mandatory( jo, was_loaded, "load_order", load_order );
```
anothersimulacrum, line 114: Is there a reason this is exposed to JSON instead of something like
```suggestion
    static int current_load_order = -1;
    load_order = current_load_order;
    ++current_load_order;
```

Ramza13: Yes, its the whole point of the PR in some ways.  The order a weather type is loaded in is arbitrary especially when dealing with mods.  But its not a meaningless field, it matters in the logic in terms of what weather it is.  If the conditions for two or more weather types are true then the one with the higher load order is what happens.

anothersimulacrum: Ah, my bad. Perhaps this field would be better named `priority`.

doc/WEATHER_TYPE.md
```diff
@@ -26,8 +26,10 @@ Each weather type is a type of weather that occurs, and what causes it. The only
 | `duration_min`                 | Optional, the lower bound on the amount of time this weather can last. Defaults to 5 minutes.
 | `duration_max`                 | Optional, the upper bound on the amount of time this weather can last. Defaults to 5 minutes.
 | `weather_animation`            | Optional, Information controlling weather animations.  Members: factor, color and glyph |
-|	`required_weathers`          | A string array of possible weathers, it is at this point in the loop. i.e. rain can only happen if the conditions for clouds light drizzle or drizzle are present |
 | `condition`                  | A dialog condition to determine if this weather is happening.  See Dialogue conditions section of [NPCs](NPCs.md)
+| `priority`                  | An integer.  If the condition of multiple weather types are true the higher load order wins
```
ehughsbaird, line 30:
```suggestion
| `priority`                  | An integer.  If the condition of multiple weather types are true the one with higher priority wins
```

PR #64123, opened 2023-03-10
Aftershock laser cannons

John-Candlebury:
#### Summary
Mods "Aftershock: Two new heavy laser cannons"

#### Purpose of change

All aftershock weapon categories should have heavy weapons, this adds the ones laser based ones. 

#### Describe the solution
Short description of the new weapons:

Izhevsk PLC-75 very big and supposedly antiquated laser cannon, inspired by the historic soviet laser pistol. Unique among other laser guns it uses expendable flashtubes as amunition instead of the usual power cartridge.  Combined with its very high per shot power, this pretty much makes it function like a very precise rocket launcher. 

Shikishma A-82 tactical laser a much more reasonable heavy laser weapon, and the one that I expect will find common use. On paper the stats are very close to those of the A7 rifle, but every fifth shot will cause an explosion in the target, which increases its turn damage considerably and allows it to damage foes that are usually immune to laser fire.  

Both spawn in the general aftershock armories. With the Shikishma being considerably more common.

#### Testing

Spawned both guns and tried them out for a bit

#### Additional context

Also noticed lots of ammo effects remain harcoded when they probably shouldn't, plenty of future opportunities there.


data/mods/Aftershock/ammo_effects.json
```diff
@@ -14,6 +20,12 @@
     "type": "ammo_effect",
     "aoe": { "field_type": "fd_elect_anomaly", "intensity_min": 2, "intensity_max": 3, "radius": 0 }
   },
+  {
+    "id": "TACTICAL_LASER_EXPLOSION",
```
Maleclypse, line 24: Both the ammo effects are quite neat!

Maleclypse (approved):
Thanks for also adding the json_info!

PR #64127, opened 2023-03-10
Re-add moose taming to tamable wildlife mod

adamkad1:
#### Summary
Mods "Re-adds moose taming to tameable wildlife mod"

#### Purpose of change

Because taming moose is cool.

#### Describe the solution

#64095 removed taming moose entirely which is not cool, so i added it to a mod instead.

Not having cool moose taming anywhere.

#### Testing

Get in game, feed moose, works.

#### Additional context

Someone probably should make moose milkable too since that seemingly was half the point of them being tameable. Not sure if I can. And I definitely don't know how much milk should a moose give

### DIFF
--- a/data/mods/Tamable_Wildlife/mammal.json
+++ b/data/mods/Tamable_Wildlife/mammal.json
@@ -350,7 +350,7 @@
     "name": { "str_sp": "reindeer" },
     "extend": { "flags": [ "CANPLAY", "PET_MOUNTABLE" ] }
   },
-    {
+  {
     "id": "mon_tusked_moose_calf",
     "type": "MONSTER",
     "copy-from": "mon_tusked_moose_calf",
--- a/data/mods/Tamable_Wildlife/mammal.json
+++ b/data/mods/Tamable_Wildlife/mammal.json
@@ -358,7 +358,7 @@
     "petfood": { "food": [ "CATTLEFOOD" ], "feed": "The %s seems to like you!  It lets you pat its head and seems friendly." },
     "extend": { "flags": [ "CANPLAY" ] }
   },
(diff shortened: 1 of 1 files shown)


Termineitor244:
Thank you! This was something I put in my back burner some time ago but never went around to do it.

adamkad1:
> Thank you! This was something I put in my back burner some time ago but never went around to do it.

Must have planned that since before the pull to remove it from base got merged eh?

Termineitor244:
> > Thank you! This was something I put in my back burner some time ago but never went around to do it.
> 
> Must have planned that since before the pull to remove it from base got merged eh?

Yeah, I wanted to add their young as tamable in the mod, since they didn't exist when I originally made it, I only remembered it because of the linked PR to remove the taming from vanilla.

adamkad1:
> > > Thank you! This was something I put in my back burner some time ago but never went around to do it.
> > 
> > 
> > Must have planned that since before the pull to remove it from base got merged eh?
> 
> Yeah, I wanted to add their young as tamable in the mod, since they didn't exist when I originally made it, I only remembered it because of the linked PR to remove the taming from vanilla.

Neat.

AcidAntOnAMinefield:
Thank you for this. I was wondering why it was completely removed instead of just being moved towards the mod in the first place.

adamkad1:
> Thank you for this. I was wondering why it was completely removed instead of just being moved towards the mod in the first place.

Laziness? Not caring? One can only speculate.

Karol1223:
> Not caring?

Yeah that one. I care more about mainline making sense than catering to mods, even if it's a mod I personally like.

PR #64128, opened 2023-03-10
Drop Windows 32 bit builds, msvc and cross build both.

akrieger:
#### Summary
None

#### Purpose of change
Windows 10 still allows a 32 bit OS install, but that doesn't mean we have to support it. According to the [steam hardware survey](https://store.steampowered.com/hwsurvey?platform=pc), only up to 0.22% of Windows installs (surveyed) can be 32 bit, the other 99.78% are definitely 64 bit.

#### Describe the solution
Don't build and ship any 32 bit variant of CDDA for Windows anymore. Remove them from the experimental release section. Keep the 32 bit arch in the solution file only for now as a fallback for compiling but expect it to be reaped soon too.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -101,22 +101,6 @@ jobs:
             ext: zip
             content: application/zip
             sound: 1
-          - name: Windows Tiles x32 MSVC
-            artifact: windows-tiles-x32-msvc
-            arch: x86
-            os: windows-2019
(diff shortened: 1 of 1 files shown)


kevingranade:
Ok so we were already not doing 32 bit PR builds, 32 bit linux builds were never in release.yml, and the build tools have dropped support for windows xp.

Yea it's time.

PR #64129, opened 2023-03-10
paws tweaks and improvements

anoobindisguise:
#### Summary
Balance "Paws increase punch damage, always require little paws"

#### Purpose of change
Paws were in need of tweaks. ~~They were overly without benefit~~ and also didn't require "little paws", instead your hands turned straight into paws without growing part of the way there first.

#### Describe the solution
~~- Little Paws no longer require Long Fingernails as it isn't a super logical requirement.~~
- Little Paws has been added as a mutation to every tree that has normal paws. It changes to normal paws, and normal paws has little paws as a requirement.  This will give paws trees more room to back out if they don't like the way their hands are going.
~~- Little Paws provides +1 bash damage for unarmed, normal paws provide +3, and large paws provide +6. As your hands are bigger, heavier, and stiffer, they become better tools to smack people, so this is a logical benefit for them to have.~~ Never mind, this is redacted.

Different values. 
Keep the long fingernail requirement to begin with, this would give more lead in to little paws (and every tree that gets regular+ paws also has long fingernails)

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -8261,6 +8261,7 @@
     "encumbrance_always": [ [ "hand_l", 5 ], [ "hand_r", 5 ] ],
     "craft_skill_bonus": [ [ "electronics", -1 ], [ "tailor", -1 ], [ "mechanics", -1 ] ],
     "types": [ "HANDS" ],
+    "prereqs": [ "NAILS" ],
     "cancels": [ "TALONS" ],
     "changes_to": [ "PAWS" ],
     "category": [ "RABBIT", "LUPINE", "RAT", "FELINE", "BEAST", "URSINE" ]
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -8264,6 +8264,7 @@
     "craft_skill_bonus": [ [ "electronics", -1 ], [ "tailor", -1 ], [ "mechanics", -1 ] ],
     "types": [ "HANDS" ],
     "cancels": [ "TALONS" ],
+    "changes_to": [ "PAWS" ],
     "category": [ "RABBIT", "LUPINE", "RAT", "FELINE", "BEAST", "URSINE" ],
     "bash_dmg_bonus": 1
   },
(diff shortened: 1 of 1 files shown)


Maleclypse:
Weren't little paws supposed to be the paws version for mutation trees that are shrinking from human size, such as rat, mouse and rabbit?  Instead of them being a step on the way to paws?

anoobindisguise:
> Weren't little paws supposed to be the paws version for mutation trees that are shrinking from human size, such as rat, mouse and rabbit? Instead of them being a step on the way to paws?

Currently only Rabbit has them, but Little Paws are still bigger than your original hands (hence the crafting difficulty and hand encumbrance)

kevingranade:
Please go touch a cat paw.
One, its nice and fuzzy.
Two, it will tell you why paws should not get bash damage bonuses.

Homonids actually have pretty outstanding punching ability, creatures with paws tend to use claws for damage, not bashing.

fairyarmadillo:
Little paws were implemented because rabbit was failing tests due to not having enough unique mutations, and to give rabbits access to the burrowing mutation. Long fingernails are a prereq because paws have longer nails than human hands. This PR would run counter to the intended design of little paws.

Rats don't use Little Paws because they get their own claw mutation that lets them burrow. The fact that mice don't get it could be an oversight or just an attempt to differentiate them from the other mutation lines - they're not as well known for burrowing as rats and rabbits.

Furthermore, Kevin's right. Paws of any sort would be neutral or a hindrance in punching. It's essentially making a less flexible hand/foot out of several fewer bones and they tend not to present a good impact surface like a knuckle. Since they can still grasp things I'm guessing it's a median state and not a full on cat paw, but I don't see this providing an advantage.

My recommendation is to close this.

anoobindisguise:
So paws bash damage was already removed, but I went ahead and made the little paws require long fingernails again.
Now all that changes is that you have to get Little Paws before you get regular Paws which I think is probably pretty uncontroversial, it's more steps of lead in into an anatomy mutation.

fairyarmadillo:
1) I don't see the purpose of turning little paws into a prereq for paws. The two mutations are functionally almost identical and little paws only exists because rabbit lacked unique mutations and was failing tests when the game tried to compile. 

2) Paws are not bigger than hands, and little paws are not a midpoint between hands and paws. One is a small mammal hand, the other is a cat/dog hand. It is only called little paws because rabbit mutants are smaller than feline/lupine mutants.

PR #64134, opened 2023-03-11
Add new basic military mine operations manual and buff the existing one

### LINKED ISSUE #63569: Update M 20-32 Mine/Countermine Operations manual to actually instruct on countermine operations

With the introduction of the proficiencies system, the check for disarming traps has been weighted such that your Trapping proficiency matters far more than your Devices skill. Even with 8 devices skill and no encumbrance, you're still more likely than not to blow yourself up attempting to disarm a landmine if you don't have any relevant proficiency.

This leaves the M 20-32 Mine/Countermine Operations in an awkward position. It only teaches the player up to level 3 devices skill, and the recipes for landmines and booby traps. Not only is this woefully insufficient for actually disarming landmines, a player who has learned everything the book has to offer has little reason to assume such a low level of skill is insufficient before they actually attempt to disarm a landmine, which will likely be the last thing they do with that character.

### Solution you would like.

Implement a Principles of Trapping proficiency practice recipe, and include it in the mine/countermine operations manual. Potentially also include a Trap Setting proficiency training recipe, though whether this would disrupt balance might be worth it's own discussion. Personally, I find it more realistic that someone would learn to set and disarm landmines by reading a US army manual on that exact subject than by placing and disarming 200 nailboard traps.

Add the following passage to the description of the M 20-32 Mine/Countermine Operations manual:
"The latter chapters of the book dealing with countermine operations have all been furiously torn out, and "JUST THROW ROCKS AT THEM!!!" has been shakily written in blood on the inverse of the back cover."

### Additional context

_No response_

strategineer:
#### Summary
Balance "Add new basic military mine operations manual and buff the existing one"

#### Purpose of change

Fixes #63569

#### Describe the solution

Buff military mine ops manual to make it as useful as it should be by:
- Increasing the traps/devices level it can teach from 3 to 5 (with a minimum devices level of 3 needed to read it)
- Adding two unique practice recipes granted upon reading the manual that allow the player character to gain the required proficiencies to set and disarm booby traps and mines (two recipes also learned by reading the same manual) (mostly?) safely.
- Adding another more basic version of this military trap manual to cover devices level 0 to 3 and act as a primer for the more advanced manual.

NB: The practice recipes require similar components to those needed by the booby trap and mine recipes excluding the explosive components for both realism (practicing with actual explosives seems like a bad idea) and balance (requiring an exorbitant amount of explosive components to do these practice recipes wouldn't be fun IMO because that would leave the player with less explosive components to make actual explosives with their newly gained skills) purposes.

None but I'm open to suggestions. It's hard for me to test this kind of balance change in a vacuum given my relative inexperience with the game (for now). One book that can bring a character from devices level 0 to 7 might be busted.

I thought about having a single recipe instead of two but putting three proficiencies into one practice recipe seemed to make the practice recipe take too long and made it harder to figure out a reasonable set of training components.

Three recipes might be overkill (eg. basics, setting, disarming) but let me know!

#### Testing

1. I spawned in both books on a fresh character.
2. Read the basic book until I reached devices level 3, ensuring that the first military trap recipe was learned upon reading the book the first time.
3. Placed some booby traps.
5. Spawned the needed ingredients and started continuously doing the newly added "military traps" training recipe whilst evaluating the resource expenditure, the rate at which the basic trap proficiency was gained, the time taken, etc. until I had gained the proficiency.
6. Read the advanced book until I reached devices level 6, ensuring that the advanced military trap recipe was learned upon reading the book the first time.
7. Did the same for the "military traps (advanced)" recipe until I had gained the next two trap proficiencies.
8. Spawned a bunch of booby traps and mines, placed them and disarmed them safely dozens of times (I'm not sure if theoretically disarming a mine at this point would possibly have it blow up in your face but if so, it seems like it would be quite rare).
9. Repeated this process a bunch until I got something that felt "right".


data/json/items/book/traps.json
```diff
@@ -74,7 +74,7 @@
     "symbol": "?",
     "color": "green",
     "skill": "traps",
-    "max_level": 3,
```
Maleclypse, line 77: I would suggest that if you are making the max 5 you need a minimum level of around 3.  Which might mean there needs to be a book that handles the lower level of traps.

strategineer: Makes sense, done!

"The Modern Trapper" book might already fill that niche? It sort of makes sense to start with making bear traps and then getting into explosive traps.

strategineer: Although I'll need to tweak the skill values in the practice recipes to account for this... And do some more testing :)

PatrikLundell:
While The Modern Trapper might fill the the low level skill needs from a technical point of view, military organizations tend to train people from scratch (or at least be set up to be able to do so), so a military basic training manual ought to exist.

Zireael07:
Mines aren't something you're training people from scratch on, though? That's something reserved for specialized forces

PatrikLundell:
I claim that yes, you need to train mine handling from scratch at some stage. Otherwise the special forces would have to only train people who somehow had gotten the basics from outside the military, complete with potential bad practices (probably hunters and farmers, as military recruits tend to be young, and thus rarely would have a prior career in using explosives for construction, mining, etc.). I would expect regular forces to get some basic training in explosives detection and brute force elimination (by setting them off/destroying them from a safe distance), but there would have to be additional training for people intended to actually be able to disarm explosive devices, and that would likely have to be done in several steps.

data/json/items/book/traps.json
strategineer: I've added a basic military trap book!

strategineer:
Thanks for the help y'all.

I've:
- Added a basic version of this military trap book.
- Lined up the values for the intelligence, max_level, time etc. for the military trap books based on the other books in the same file.
- Added the new basic military trap practice recipe to the basic book (the advanced book contains both).
- Set the max devices level that can be trained by the advanced recipe to 6 instead of 7 because that seems like it's enough to safely disarm land mines and lines up with max level you can gain from the book.

strategineer:
I tested this some more, should be good to go. Let me know if there's anything else!

PR #59061, opened 2022-07-08
Nyctophobia

Night-Pryanik:
#### Summary
Content "Added nyctophobia trait"

#### Purpose of change
AKA fear of dark. Add trait for RP and gameplay.

#### Describe the solution
- Players with nyctophobia can enter tiles with `cloudy` and `dark` light levels. Tiles with `very dark` light level are off-limits to them, unless player character is running or took a xanax.
- Players have a chance to see monsters (represented as hallucinations) in very dark tiles in 5-tile radius around the player, even if players themselves are standing in lit tile. The chance to spawn hallucinations is 2% per turn if player is standing in lit tile, and 10% if player is standing in very dark tile.
- Players have a chance to hear auditory hallucinations while staying in very dark tiles. The chance is 50% every 30 seconds.
- Players suffer several mental and physical effects if they are standing in very dark tiles. 2% chance per turn for every effect, i.e. they can happen all at once. Among them:
- - shaking hands (penalties to dexterity and strength);
- - trembling legs (player will fall down);
- - sudden suffocation attack (loss of stamina);
- - paralyze by fear (loss of turns).
- All negative effects are alleviated if player has taken xanax.

#### Describe alternatives you've considered
- Instead of strict blocking of entering very dark tiles, use a chance.
- Add more effects.

#### Testing
Got nyctophobia. While standing in lit tiles, looked at very dark corners. Eventually hallucinations monsters appear in these corners. Tested the same while standing in very dark tiles.
Walked over very dark tiles. Eventually was knocked down, got shaking hands, lost some stamina, got paralyzed for few turns.

#### Additional context
None.

PR #64141, opened 2023-03-11
greatly reduce the severity of nyctophobia, also removing hallucinations from it

esotericist:
#### Summary
Balance "greatly reduce the severity of nyctophobia, also removing hallucinations from it"

#### Purpose of change

nyctophobia as implemented in #[59061](https://github.com/CleverRaven/Cataclysm-DDA/pull/59061) is far more crippling than warranted, especially for its apparent intended severity.

most notably, many of the effects that can trigger can be fairly immediately lethal in a lot of circumstances, the mitigation available (via medication) isn't reliable (especially for an early game character), and the hallucinations in particular are extremely problematic: just being near darkness can result in arbitrary hallucinations, even if standing in full sunlight.

further, the standard scenario (evacuee) is always in darkness, which can instantly trigger multiple effects (and is all but certain to result in at least one hallucination) even if the player immediately turns on a light source.

#### Describe the solution

remove the hallucination effects from the trait entirely, alongside the downed and paralyzed effect.

slightly reduce the frequency of the shakes/stamina impacts (but also increase the maximum shakes duration).

ensured that the stamina impact cannot also occur in the same turn as shakes, also ensured the periodic fear notification won't occur in the same turn as shakes or the stamina impact.

maintained the movement restrictions unmodified; this at least was working in a reasonable fashion.

considered adjusting the point value of the trait down (from 3 to 2), but given the long-term plans to discard point values, and the highly subjective impact of the changes, it seemed superfluous.

considered adding a second, more severe version of the trait, but per kevin this isn't desirable.

longer term a more staged progression (using one of our many systems: vitamins, eocs, etc) could be beneficial (to allow the player to reliably spend just a turn or two running through a dark area without immediately being impacted), as well as allowing some of the effects to scale upwards. it could be unhardcoded at the same time; there is nothing about this trait that strictly requires c++. for now, as this is a 0.G candidate, i've opted to not do anything that could cause string shuffling (beyond the removed strings).

for anyone working on this in the future: please do not reintroduce visual hallucinations, they are too problematic from a gameplay standpoint for what this is clearly intended to represent. if auditory hallucinations or the other effects (downed/paralyzed) are reintroduced, they should be only at a high fear intensity (i.e. a long time spent in the dark).

#### Testing

made a character with nyctophobia, on the evacuee scenario. waited. done in 30 minute increments, a few shorter increments, and simply holding down `.`

notably, even waiting in place with the reduced frequency of the stamina impacts, it's possible for stamina to get fairly low. a character actively moving through darkness (which still _requires_ running) is very likely to have severe stamina problems in short order.

#### Additional context

screenshot showing incidence rate of remaining events in 30 minute wait time:
image

one of the lower stamina values observed while waiting.
image

### DIFF
--- a/src/suffer.cpp
+++ b/src/suffer.cpp
@@ -1675,56 +1675,30 @@ void suffer::from_tourniquet( Character &you )
 
 void suffer::from_nyctophobia( Character &you )
 {
-    std::vector<tripoint> dark_places;
     const float nyctophobia_threshold = LIGHT_AMBIENT_LIT - 3.0f;
 
-    for( const tripoint &dark_place : points_in_radius( you.pos(), 5 ) ) {
-        if( !you.sees( dark_place ) || get_map().ambient_light_at( dark_place ) >= nyctophobia_threshold ) {
-            continue;
-        }
-        dark_places.push_back( dark_place );
-    }
-
     const bool in_darkness = get_map().ambient_light_at( you.pos() ) < nyctophobia_threshold;
-    const int chance = in_darkness ? 10 : 50;
-
-    if( you.is_avatar() && !dark_places.empty() && one_in( chance ) ) {
-        g->spawn_hallucination( random_entry( dark_places ) );
-    }
-
     if( in_darkness ) {
+        if( one_in( 80 ) && !you.has_effect( effect_shakes ) ) {
+            you.add_msg_if_player( m_bad,
+                                   _( "Your fear of the dark is so intense that your hands start shaking uncontrollably." ) );
+            you.add_effect( effect_shakes, rng( 1_minutes, 3_minutes ) );
+
+            return;
+        }
+
+        if( one_in( 80 ) ) {
+            you.add_msg_if_player( m_bad,
+                                   _( "Your fear of the dark is so intense that you start breathing rapidly, and you feel like your heart is ready to jump out of your chest." ) );
+            you.mod_stamina( -500 * rng( 1, 3 ) );
+
+            return;
+        }
+
         if( one_turn_in( 5_minutes ) ) {
             you.add_msg_if_player( m_bad, _( "You feel a twinge of panic as darkness engulfs you." ) );
         }
 
-        if( one_in( 2 ) && one_turn_in( 30_seconds ) ) {
-            you.sound_hallu();
-        }
-
-        if( one_in( 50 ) && !you.is_on_ground() ) {
-            you.add_msg_if_player( m_bad,
-                                   _( "Your fear of the dark is so intense that your trembling legs fail you, and you fall to the ground." ) );
-            you.add_effect( effect_downed, rng( 1_minutes, 2_minutes ) );
-        }
-
-        if( one_in( 50 ) && !you.has_effect( effect_shakes ) ) {
-            you.add_msg_if_player( m_bad,
-                                   _( "Your fear of the dark is so intense that your hands start shaking uncontrollably." ) );
-            you.add_effect( effect_shakes, rng( 1_minutes, 2_minutes ) );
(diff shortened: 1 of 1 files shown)


Night-Pryanik:
I don't like this at all. You're basically destroying my work as nyctophobia was designed to be severe and crippling, and definitely not for new players or players with little game experience.

NetSysFire:
I feel like there can be two varities. This one and the more severe original. You could attain the latter by random mutations or maybe even specific mutagen flavors, but I cant think of any. Troglobite would never get any of this. Maybe could fit medical or something.

This would be a good compromise in my opinion.

esotericist:
> I feel like there can be two varities. This one and the more severe original. You could attain the latter by random mutations or maybe even specific mutagen flavors, but I cant think of any. Troglobite would never get any of this. Maybe could fit medical or something.
> 
> This would be a good compromise in my opinion.

i offered to do two traits with different severities and kevin told me no.

the long term plan is going to be to make this have progressive elements, so it gets more severe, as you spend more time int he dark. i just wanted to keep things simple for now, to address the most egregious problems from being briefly in the dark, or in the dark when you first start a run (as evacuee mandates).

MochaModifier:
The hallucinations were just obnoxious and nonsensical, and I'm supposing that's because it used the same mechanic as psychedelic's or kaluptic episodes
This makes sense if you took a pink tab and are riding the white rabbit - but nyctophobia should be characterized by _fear_.

Spending extended periods of time in the dark I can see causing progressively worse symptoms.
Being stunned, stramina loss etc. effectively seems like a full blown panic attack

I can understand what was attempted with hallucinations - but it seems to be re-using the existing hallucination mechanics, making it less scary and more obnoxious.
A more subtle pool of things, stuff you absolutely don't want to run into in the dark, things that _terrifies_ any survivor at night - thats the kind of 'hallucination' the player should see spawning.  Totally get that would likely require a whole new thing be coded, but it would ideally do nyctophobia justice if so.

If nyctophobia is fear of the dark, players shouldn't see a skeleton hulk day one and think "oh lol, whatever". It might fool a new player, but anyone who's played a while and knows where they are, wouldn't buy it.

But if instead, the character imagined a couple deranged axemen, a barghest and dog pack, grapplers etc. - that might prompt some 'irrational' responses to avoid danger. As zombies evolve, the player character would see all kinds of new terrors to imagine in the dark - and this would simply operate off the evolution mechanic in game, using monsters that are valid to spawn rather than the hallucinations pool.

- If I were to imagine Nyctophobia, it'd be a progression of maluses:

The time to start start seeing these should probably be a bit longer than the time it takes to go from house to house. Maybe a minute. 
Spending more than a minute or two should start damaging focus, harm your morale, quickly progressing to auditory hallucinations.
Then it progresses to weariness malus as the character starts hyperventilating. Moving becomes more tiring as you can't control your breathing. 
Jitters start to set in as oxygen rushes to your brain.


Spending more time in the dark, this progresses to your fight or flight instincts gong wild - imagining monsters in the dark, the feeling of being hunted by things that may or may not be there. 
You may start to yell or scream in teror.
Panic sets in; tremors, movement penalties, and if this continues - a full on panic attack. 
Knocked to the floor, loss of stamina, severe stat maluses, forced to go prone etc. You eventually recover - a burst of adrenaline; but this will keep happening if you can't get in the light to calm down.
You can't focus, fighting is near impossible and all you can think of is to run.

Turning the light on doesn't immediately make this go away either - coming down from that level of stress and adrenalin would take time. 
The longer you spent in the darkness, the longer it takes to come down - but still faster than it took to get to this point. You can see again - and can atleast tell real monsters from the shadows.

Issue #61505, opened 2022-10-06
"Monotask" ferals: an idea that solves minor consistency issues and opens a lot of design space


I-am-Erk:
### Is your feature request related to a problem? Please describe.

We have numerous spots in the game where a gameplay element is fun or good, and we don't want to remove it, but also makes less and less sense with the changing game world. For some of these we have ways to fix them later (cybernetic nurses will eventually be a thing inside of a netherum dungeon instead of a lab, and they'll fit fine there), but others are trickier. Roadblocks with eternally-powered turrets are one example, another is the eyebot-copbot combo. The copbot one is a favourite example for me because those guys are fun, kind of hilarious, and have some good gameplay, but they really don't make sense. Why aren't they harassing the zombies? How are they powered? How can the whole city be on fire but these copbots just endlessly show up?

### Solution you would like.

The lore around ferals gives us room for a good answer to most of these situations.

Note that this is **not** a lore rewrite. This doesn't change the timeline or anything already established, it just clarifies existing stuff around what we call "blob psychosis"

Add a new type of feral, let's call them Monotasks. These are people whose response to blob infection was to become unrelentingly fixated on a specific complex duty, to the exclusion of all else not related to continuing this job. This is inspired by classic Romero zombies, who had a similar trait, but we're adapting it to our world now. Monotasks are, like ferals, still alive, and they do necessary minimal scavenging to stay alive to do their fixated tasks.

The tasks a Monotask fixates on should fit certain criteria:
- they require your attention, but are kind of boring
- the Monotask knew how to do them well before being affected

Unlike other ferals, at least some Monotasks should be aware of the state they are in, but unable to change their behaviour. It could be possible to talk to them. Not all Monotasks are hostile to the player, especially the talking ones.

Monotasks may be less rare than ferals, particularly because they tend to be people who were holed up safely and engaged in less self-destructive behaviour. However, some specific monotask encounters should still be quite rare.

Examples of Monotask encounters:
- Road block: A military monotask is inside a humvee, refueling the generator for turrets and keeping things operational. The turrets are no longer hostile to zombies. This is a non-talking monotask.
- Guard duty: replacing turrets we have monotasks who will guard a zone, even long after it is destroyed. To aovid "haha you're dead" the most common form of these should probably be riot cops with batons and shields, not people with guns.
[...]

PR #64142, opened 2023-03-11
Introduce some neutral civilian monsters to better fit the initial chaos after the Cataclysm

NetSysFire:
#### Summary
Content "Addition of non-NPC civilians for lore reasons"

#### Purpose of change

This has been bugging me for a long time and is a more complex issue. It has been 5 days since the Cataclysm when you start a game and some scenarios, e.g in this case the firefighter are described as `On your way to respond to an emergency call, you nearly drove straight into a riot in the city.`, as if everything was sort-of normal until this today.
This PR introduces some passive monsters to add some initial chaos and action.
Also seals the fate of the feral humans in the far future by making them upgrade into zombies and stopping the spawning of them.

#### Describe the solution

This is still a draft and incomplete on purpose. I know this might be a more "controversial" PR and before wasting lots of time implementing things that will get rejected, I kept it lean on purpose.

Adds 3 new monsters, all are designed to die quickly and revive as zombies, because humanity is getting decimated and this should visualize this fact:

- The panicked person, only runs away and screams.
- The petrified person, just stands there and barely moves.
- The futile fighter, the only one which is not purely passive. It will attack zombies and also make noise to attract attention when it sees hostiles.

None of them will attack the player by default.
In this case, I placed more fighters in the prepper group because it seems appropriate that these are the ones getting cocky and want to fight back. Blob psychosis does the rest here. On the other hand, a school will have more petrified/panicking ones (something along the way of traumatized children, or something) and few fighters. Still subject to change, obviously.
All of this will stir things up a little, so when you go scavenging in the city early on, you will still see occasional fighting. This will die down (pun not intended) pretty quickly and stop affecting the game after maybe 14 days at most. In my opinion, this adds some much needed flavor, but there needs to be more chaos overall (e.g more debris and burning cars and so on) but this is out of scope for this one.

I made feral humans have a half life of 90 days until they "die" (upgrade). Very basic scavenging only gets you this far and clean water will be very hard to come by, adding lots of potential for untimely deaths. Plus all their cuts and bruises will not receive medical care at all. And they wander around with lots of toxic zombies. From acidic ones, to the toxic gas emitting ones to electric and potentially fiery explody ones.
I made the final deadline 380 days and 400 for sapiens.

Not making feral humans upgrade and letting them die instead, so they can properly revive.

#### Testing

None yet /o\\, I still need to compile my version of 0.G and the one shipped by my distro is already too old.


GuardianDll:
You may give them some fluff topics so you can talk with them, seaparting from common ferals

harakka:
These sound more like just regular people implemented as monsters, than _ferals_, as the concept is used in DDA.

NetSysFire:
To cite the [design document](https://cataclysmdda.org/lore.html):

>In about one quarter of the population, changes are basically unnoticeable.
>For about half the population, the blob causes an increase in risk-taking behaviour, ranging from mild unsafe practices all the way to people attacking hordes of zombies with a stick. This effect may be permanent or may be temporary, and resolves at varying speed by person. Some may be strongly enough affected to be almost indistinguishable from ferals, but not aligned with zombies.
>The remaining quarter see a heavy increase in aggressive, violent behaviour, often completely out of character. The worst of these, about 1/20 of the population, become “ferals”, a form of living zombie.

Currently we have ferals, zombies and thats it. Where did the remaining population go? 2/3 of my "ferals" are not very violent and yes, I agree that my additions are more like "regular people" to represent an unimplemented part of the population. I am fine with renaming them but they are still affected by blob psychosis, which is causing their behavior. Making them all NPCs is absolutely unfeasible (due to performance and complexity, so no talking with them) and there is a reason that the remaining survivors are so scarce.

See #61505. This is why I called them ferals for now. Also cc @I-am-Erk I guess.

harakka:
Just to clarify I'm fine with the concept, just wanted to clarify the term. As we use "feral" now, it specifically refers to the "remaining quarter".

Also, I believe we have functionality for talking to monsters now, so that is still doable with this implementation, it just isn't used anywhere currently.

TheShadowFerret:
Shouldn't these neutral ferals cause morale drop if you kill them without Psychopath?

~~Also more sources of human flesh nom nom~~

NetSysFire:
Yes, they should. Thanks for the idea!

RAldrich:
So, I like the general concept - but I'm wondering how believable this is going to come across in-game. For the first 90 days, whenever you enter a new city, you're more likely than not to find panicked civilians running in the streets? Who will then immediately get killed by the zombies? Why were they not killed before the player showed up to watch the scene? I think this will aggravate the issue of the world feeling static unless the player is observing it.

They're also going to come across as incredibly un-intelligent actors. What happens if the player saves them? They just stand in the street forever, never eating, never sleeping? I know that's what our current NPCs do, but at least they do it indoors where it makes sense to stay put.

I assumed that we don't see that 50% of the population because they ran panicked through the street and were killed by zombie hordes in the first day or two. Everyone who was unaffected enough to survive past that probably qualifies as a NPC.

PatrikLundell:
I don't see these behaviors as feral behavior, but rather various shock responses. Thus, I don't see any reason for zombies not to chase the ones running around screaming and kill the frozen ones once they catch their smell. Fighting back against zombies while they're nominally on the same team doesn't make sense either.

Monotask firefighters "putting out" fires with empty buckets or disconnected hoses would make a lot more sense.
The same goes for monotask police ticketing all the cars in the area (if that's a police task in the game's location).

NetSysFire:
>For the first 90 days, whenever you enter a new city, you're more likely than not to find panicked civilians running in the streets?

A week or two. They stop spawning after a couple of days and upgrade pretty quickly into proper ferals. I could add some more detail and give them a say 75% chance of dying and a 25% chance of turning feral.

>I don't see these behaviors as feral behavior, but rather various shock responses.

This is true but the passive effects of the blob cause a majority of the population to participate in more risky behavior. So in our case, a lot of people take a risk by going outside and trying to get away. They think they might make it.
Although I do realize that there is no real explanation for the ones just frozen in fear, that does seem a bit far fetched now after a good night of sleep.

>I don't see any reason for zombies not to chase the ones running around screaming and kill the frozen ones once they catch their smell.

They will. I set their faction to human on purpose.

PatrikLundell:
If the fleeing critters are set to belong to the human faction it seems the only reason they're "feral" is for technical implementation purposes, i.e. they are human lore and game play wise (apart from the no talk bit, but someone in shock running around screaming is likely not very easy to talk to in real life either), but technical constraints cause making them feral under the hood a more palatable implementation.

NetSysFire:
> but technical constraints cause making them feral under the hood a more palatable implementation.

I do not see any technical difficulties with not making them feral. The only thing making them feral right now is their ID and the file they are in. I am having a hard time finding a name for them. Civilian sounds weird, feral is not optimal either.

PatrikLundell:
What about shocked_locals?

fairyarmadillo:
If Exodii can be monsters (codewise), I don't see why irretrievably compromised humans can't be. This, like insects, will help the player feel like their character isn't the center of the universe. There's great potential for emergent storytelling too.

I think their name and descriptions should make it clear that they're beyond saving, and I think we should avoid terms that are used for standard mental illness to make it clear that these people are not saveable. Something like "hollow" in Dark Souls, but not that. Lost? Broken?

This makes the Exodii quadrupeds' origins a bit darker/sadder, too. Great stuff.

RAldrich:
> A week or two. They stop spawning after a couple of days and upgrade pretty quickly into proper ferals. I could add some more detail and give them a say 75% chance of dying and a 25% chance of turning feral.

I must have misunderstood the half life calculation on their evolution. Two weeks is not as bad as 90 days. I think the more you tune this down towards only spawning very early on, the less I object to it. Personally I'd target like, two or three days.

To be clear, I have zero issue with implementing them mechanically as monsters, so long as they don't persist long enough for that to become a problem. For example, the Exodii are simple to implement as monsters because they cannot (or will not) speak to the player, and don't (when you see them in game) appear to need to eat or otherwise do NPC-like stuff to stay alive.

NetSysFire:
>I must have misunderstood the half life calculation on their evolution.

Or I did /o\\. In any case, my target is this very timespan, 5 days max. I'll think I go with upgrading the monster into mon_null (death) 90% of the time and 10% into a feral. The petrified ones have a much lower chance of turning feral and a much higher of dying. The futile fighters on the other hand can turn feral more quickly. Or something.

My current technical concern is that when I let them upgrade into mon_null you will find corpses of them a year later when you explore other overmaps. I do not know how this will be handled and I hope those corpses will not show up at all.

anothersimulacrum:
Let them upgrade into zombies instead of just dying.

RAldrich:
> My current technical concern is that when I let them upgrade into mon_null you will find corpses of them a year later when you explore other overmaps. I do not know how this will be handled and I hope those corpses will not show up at all.

This at least is easy to test! Just fire up a new world with the start date cranked up to year two or three.

> Let them upgrade into zombies instead of just dying.

This is largely equivalent, as the corpses won't be pulped and so will re-animate as zombies.

PatrikLundell:
The corpses you come across that reanimate have very low health. so I would expect turning them into zombies directly would be better.

NetSysFire:
I made a feature request for this and until this is added, I will go with zombifying ferals directly

NetSysFire:
I updated the PR description and title and renamed these ferals to civilians. The work that I still need to do is to adjust monster groups, make sure I understand that half life function and some lore stuff. E.g putting a hint that these ones can not be saved anymore into the generic tips a NPC (e.g the one in the default evac shelter) can give you.

Besides that, I will probably also add some more variety. Things that come to my mind are people who hide (e.g in basements or on roofs), ones that decided to get high for their last hours of life and will behave erratically.

The more I think about it, the better some more NPCs sound, too. Ones with a mission of getting to a safe place perhaps but this is pretty advanced and probably out of scope here. Plus NPCs are not good for performance and can slow down things dramatically.

PatrikLundell:
It would be mildly embarrassing if someone else pointed out things missed in a finished PR. Detecting your own omissions in a draft is merely annoying...

NetSysFire:
Yes but these are simple errors I could avoid if I paid more attention instead of doing things in the deceptive free time period. Aka the time between going to bed, saying "nah just 5 more minutes" and actually going to bed.

data/json/monsters/feral_humans.json
```diff
@@ -45,6 +45,7 @@
       }
     ],
     "death_drops": "feral_humans_death_drops_pipe",
+    "upgrades": { "half_life": 90, "into": "mon_zombie" },
```
TheShadowFerret, line 48: Would a zombie master's forced upgrade make ferals auto-die from this? 🤔

NetSysFire: image

Wow, it looks like it behaves properly in regard with evolution. I spawned some ferals after setting the time to 5 years later and a zombie master next to them.

NetSysFire:
I have thought about all this a bit and have some (hopefully) cool ideas: Basically representing professions. E.g an overconfident officer or fatal firefighter (flavored zombie fighter). Of course, these will be rare and not all fighters. Especially the officer would be quite rare and/or designed to die quickly.

I think having a police officer with a gun would work quite well here, but balancing is key.
However, I am having a technical difficulty here: There are seven different guns a police officer can carry (item group `sidearms_cop`). Due to how monster ranged attacks work, one needs to hardcode a specific gun and ammo. Of course, I could make 7 variants of that monster but that seems excessive. I know next to nothing about guns, so I do not know what gun/ammo combination to pick and would like some help with this.

And rest assured, I do have balance in mind since I do not want an overpowered neutral monster taking out all the zombies for the player either.

NetSysFire:
I looked into being able to speak to these civilian monsters but that requires them to be friendly, which is not possible for a variety of reasons. Other than that, it should work fine.

NetSysFire:
By the way, I managed to find out why the officer did not shoot - they had no flags set and thus could not see.

bombasticSlacks:
Wanted to say this is a 👍 from me

bombasticSlacks:
> However, I am having a technical difficulty here: There are seven different guns a police officer can carry (item group sidearms_cop). Due to how monster ranged attacks work, one needs to hardcode a specific gun and ammo. Of course, I could make 7 variants of that monster but that seems excessive. I know next to nothing about guns, so I do not know what gun/ammo combination to pick and would like some help with this.

I don't think it ever says directly when the monster shoots so you can just pick whichever gun has a fair damage profile and then drop a modified cop drops on death that guarantees a gun and ammo IMO.

NetSysFire:
@I-am-Erk I have since renamed the whole thing because, as you and others pointed out, these are not really ferals.

These ones are representing the civilians which will die when the apocalypse starts and their death is enforced with upgrades. When the Cataclysm, aka the default start -5 days, struck, lets say that at least 90% of the population is already dead/zombified but there are still people running around, which will not survive for several possible reasons:

1. Simply bad luck.
2. State of panic.
3. Increased risk taking behavior (blob thing), see futile fighter.

1 would include:

- Being in an unfortunate place, e.g building in the city center or near a horde
- Being ill or otherwise in a bad state
- Being in an accident

2 is a bit trickier. How would the population react when more and more people die, revive as mindless zombies and actively attack people? Why did these riots, especially when unsupervised, not cause zombies to break down doors and attack people in houses earlier (**important-ish lore question**)?
In any case, many people will likely panic and try to flee (panicked person). Some might also simply be overwhelmed and freeze in their track (petrified person).

3 is mainly being cocky. The futile fighter or overconfident officer are the embodiment of this. But I think it might also include to try and mix a molotov at home, throwing it at zombies and ending up setting the house ablaze. Or trying to escape in a car, mowing down a tiny part of a horde, only to be stopped by the sheer mass.

So this is aimed at solving the "all people except 1% zombified at the same time" problem and will also add some twists and turns to city combat in the first 3-5 days.

The things which I am still working on for this PR is playtesting to determine the amount of civilians spawned and how they behave in combat. My specific concerns which I will look at are that the civilians are too distracting and make it trivial to go loot in the city, but as I said, I am onto this. This will take some more time because it is not a minor change and needs more extensive playtesting done by me (although you are welcome to help, too).

NetSysFire:
There are two issues that I have yet to fix, one when entering the game, which is a bit confusing since I did not change that part and I have no idea why this happens:

```
 DEBUG    : Error: Json error: data/json/feral_humans.json:93:33: Monster attacks must specify type and/or id

    "families": [ "prof_intro_biology", "prof_physiology" ],
    "special_attacks": [
      [ "PARROT_AT_DANGER", 0 ],
                                ^
      {
        "id": "stretch_attack",
        "cooldown": 5,


 FUNCTION : bool main_menu::new_character_tab()
 FILE     : src/main_menu.cpp
 LINE     : 983
 VERSION  : 060fe4f
```

And the tests. This happens due to `copy-from`. I tried appending `into_group: "GROUP_NULL"` but the above error kept me from actually testing that. Perhaps removing the upgrade with `delete` and then setting it anew might work.

```
(~[slow] ~[.],starting_items)=> (continued from above) ERROR : src/monstergenerator.cpp:1534 [void MonsterGenerator::check_monster_definitions() const] both into and into_group defined for monster mon_feral_human_pipe_fungal_infected
(~[slow] ~[.],starting_items)=> 09:10:11.534 ERROR : src/monstergenerator.cpp:1534 [void MonsterGenerator::check_monster_definitions() const] both into and into_group defined for monster mon_feral_human_crowbar_fungal_infected
(~[slow] ~[.],starting_items)=> 09:10:11.534 ERROR : src/monstergenerator.cpp:1534 [void MonsterGenerator::check_monster_definitions() const] both into and into_group defined for monster mon_feral_human_pipe_fungal_corpse
(~[slow] ~[.],starting_items)=> 09:10:11.535 ERROR : src/monstergenerator.cpp:1534 [void MonsterGenerator::check_monster_definitions() const] both into and into_group defined for monster mon_feral_human_crowbar_fungal_corpse
```

data/json/monsters/feral_humans.json
```diff
@@ -45,6 +45,7 @@
       }
     ],
     "death_drops": "feral_humans_death_drops_pipe",
+    "upgrades": { "half_life": 90, "into_group": "GROUP_ZOMBIE_UPGRADE" },
```
Maleclypse, line 48: Here is your problem.  You've chained copy from so that all your infected copy an `into_group` from here while also adding `into` in the first version of the `mon_civilian_panic`.  You'd need to extend delete this or change your into's to an into_group replacing the one copied from.

NetSysFire: Thanks for the hint. I tried but I am doing this blind since I can not ignore the `Monster attacks must specify type and/or id` error I get when I apply this PR locally. This iteration of blindly trying things has not worked, unfortunately.

Maleclypse: OK since the error didn't change that means you can't use delete to delete "into" that's good to learn.  The only thing you can do then is not copy-from a monster with `into` and make a monster that uses `into_group`.   Above are you saying if you don't use copy-from it breaks the PR locally?

NetSysFire: Okay, since I took a little break from working on this, I can now see that the error I could not ignore is no longer present in experimental. Hooray, I can continue testing!

I tried the following, which all do not work:

```json
"delete": { "upgrades": [] },
"delete": { "upgrades" },
"delete": { "upgrades": {} },
"delete": { "upgrades": { "into_group": "GROUP_ZOMBIE_UPGRADE", "half_life": 90 } },
```

NetSysFire: Irwiss suggested I use a `GROUP_NULL` override and it worked!

NetSysFire:
>some of the speeches of the portal storm memories.

Got an example?

>Also, would this affect some scenarios?

Yes, there is a mentioned issue just above which mentions how it positively affects city starts by making them more dynamic.

WreckingAbandon:
> > some of the speeches of the portal storm memories.
> 
> Got an example?


Yeah. Things like the "I don't wanna die!", "HELP! HEEELP!", perhaps the "*agonized screaming!*" one, "Get it off me!" might be good but it has to be in a specific context like getting grabbed by zeds. I'm just imagining this for the panicked ones though, dunno about those that decided they'd go down swinging.

NetSysFire:
Would be fun to have but the PARROT attack we are using for this can not do conditionals yet, so it is sadly impossible to implement right now.

WreckingAbandon:
Were futile fighters intended to spawn in the Refugee Center back bay? I don't think that would make sense, since Smokes explained that they only tossed their dead bodies there. Additionally, it also made missions easier since instead of spawning zeds civilians spawn instead, plus the ones they manage to kill.

PR #64152, opened 2023-03-11
Switch to C++17

jbytheway:
#### Summary
Infrastructure "Convert build to C++17 (from C++14)"

#### Purpose of change
Following up on #63970 and #64011.

We want to move to C++17 to get new features and generally keep the codebase up to date.

#### Describe the solution
Update each of the build files to use C++17 (`Makefile`, `CMakeLists.txt`, Visual Studio projects).

Fix the various compiler errors and clang-tidy warnings that arise.

~Waiting until #64011 was already merged.~  It is merged now.

#### Testing
Unit tests, clang-tidy.

Mostly this needs to be tested on CI, so I have to open it as a non-draft, but marking it WIP because there's bound to be issues on other compilers.

#### Additional context


irwiss:
Considering not all cpp17 features can be used; can structured binding be used? Asking both from compiler support and code style perspectives? E.g.
```
nutrients this_min;
nutrients this_max;
std::tie( this_min, this_max ) = compute_nutrient_range( result_it, rec, extra_flags );
```
into
`auto [this_min, this_max] = compute_nutrient_range(...`
(as you can see type information isn't visible)
Decomposing pairs is by far the most annoying part of using them

jbytheway:
From a compiler support perspective, yes, structured bindings are fine.

From a code style perspective, I think we will probably allow them in certain situations.  But since we do have the "almost always avoid `auto`" rule, we'll probably want to be cautious about how we adopt them.  Only time (and @kevingranade) will tell.

At the least I hope to use them for iterating over containers of pairs, which is a super common pattern in our codebase.

jbytheway:
For reference, the C++17 features which we should avoid are documented [here](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/COMPILING/COMPILER_SUPPORT.md),

BrettDong:
Maybe revert back to ghc file system and see if permission errors go away?

jbytheway:
> Maybe revert back to ghc file system and see if permission errors go away?

@akrieger is hoping to test locally and see what can be done.

kevingranade:
I saw an instance of this in an unrelated PR yesterday and it resolved with a rerun, so at a minimum I don't think this PR is the cause.

jbytheway:
Good to know, thanks!

RAldrich:
I would appreciate it if we could resolve #64210 before merging this. (Not going to fix / dropping Debian support is totally a possible resolution - I just think it would be best to make a decision before merging additional compiler changes).

jbytheway:
I think the issue in #64210 should be mostly orthogonal to this, but thanks for the heads up.

BrettDong:
Successfully reproduced the error. Built with VC2019 build tools.

image

BrettDong:
The error is caused by the trailing `/` in the path requested. I made it working in VC2019 by removing trailing slash:

image

```diff
diff --git a/src/filesystem.cpp b/src/filesystem.cpp
index e498da3736..d437f50470 100644
--- a/src/filesystem.cpp
+++ b/src/filesystem.cpp
@@ -21,7 +21,11 @@

 bool assure_dir_exist( const std::string &path )
 {
-    return assure_dir_exist( fs::u8path( path ) );
+    std::string p = path;
+    if (!p.empty() && p.back() == '/') {
+        p.pop_back();
+    }
+    return assure_dir_exist( fs::u8path( p ));
 }

 bool assure_dir_exist( const cata_path &path )
```

akrieger:
I saw something related to that in the some 2019 patch notes. I believe there are many fixed listed in https://learn.microsoft.com/en-us/cpp/overview/cpp-conformance-improvements-2019?view=msvc-170 which are backports of recent standards committee meetings, including stuff about trailing slashes like mentioned in https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87846

akrieger:
Specifically https://cplusplus.github.io/LWG/issue3096 is probably the relevant change.

src/filesystem.cpp
```diff
@@ -21,6 +21,15 @@
 
 bool assure_dir_exist( const std::string &path )
 {
+    // TEMPORARY until we drop VS2019 support
```
BrettDong, line 24: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=87846 says GCC was behaving the same way until the fix in GCC 8.3. Our documentation says minimum supported GCC is 8.1, so a bump in GCC support policy is also needed to remove this workaround in the future.

jbytheway: That's a doc mistake.  It's supposed to be minimum gcc 9.1 now since dropping Code::Blocks.  I'll PR a fix to the docs.

kevingranade:
Regarding the item density test, I thought it was something like that too, but it turns out it's much more prosaic, the clang++-12 workflow is the first one that runs with magiclysm loaded, and you were hitting a magiclysm-specific data error that was fixed in #64258
Other than that, it looks like this will be good to go with a rebase.

jbytheway:
OK, I've removed the item density fix and rebased on master.

Based on previous experience on this PR, the CI is going to take a long time to run (presumably because all the cached data is useless here).

tests/list_test.cpp
```diff
@@ -141,6 +140,7 @@ TEST_CASE( "list basics", "[list]" )
         SECTION( "reverse iterator count/access" ) {
```
akrieger, line 140: Either this is lying or the loop below is wrong.

src/tgz_archiver.h
```diff
@@ -2,6 +2,7 @@
 #define CATA_SRC_MAP_TGZ_ARCHIVER_H
 #include <array>
 #include <chrono>
+#include <filesystem>
```
akrieger, line 5: Use our filesystem header filesystem.h or <ghc/fs_std_fwd.h>. Although given it already resolves fs::file_time_type, one or both are already transitively provided.

src/harvest.cpp
```diff
@@ -43,20 +43,23 @@ bool string_id<harvest_drop_type>::is_valid() const
...
+    // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
+    return SNIPPET.random_from_category(
+               succeeded ? msg_butcher_success : msg_butcher_fail ).value_or( translation() );
 }
 
 translation harvest_drop_type::dissect_msg( bool succeeded ) const
 {
-    return SNIPPET.random_from_category( succeeded ? msg_dissect_success : msg_dissect_fail ).value_or(
-               translation() );
+    // NOLINTNEXTLINE(clang-analyzer-cplusplus.NewDeleteLeaks)
```
akrieger, line 60: ?

akrieger (approved):
A lot of this lint shouldve been caught pre-17, no?

tests/list_test.cpp
jbytheway: Yeah; this is a copy/paste error.  Will fix.

src/tgz_archiver.h
jbytheway: Weird.  I don't know how this ever worked on C++14, but I will fix.

src/harvest.cpp
jbytheway: Something about the implementation of `unique_ptr` has changed in C++17 which causes what I believe to be a clang-tidy false-positive here.  I tried some tweaks to fix it, but nothing worked, so in the end I chose to suppress it.

Part of me is a little worried that it's not a false-positive and there really is a potential memory leak here, but the `value_ptr` implementation looks pretty solid to me...

src/tgz_archiver.h
andrei8l: It was already included through `"filesystem.h"` -> `<ghc/fs_std_fwd.hpp>` -> `<filesystem>`

https://github.com/CleverRaven/Cataclysm-DDA/blob/8286f4d7d8f43e1d3b79125ceb2270facf9e0caf/src/tgz_archiver.h#L8

src/harvest.cpp
jbytheway: In case you're curious to analyse it yourself, here's the error message we get without the suppression:
```
.../Cataclysm-DDA/src/harvest.cpp:46:5: error: Potential leak of memory pointed to by field '_M_head_impl' [clang-analyzer-cplusplus.NewDeleteLeaks,-warnings-as-errors]
    return SNIPPET.random_from_category(
    ^
.../Cataclysm-DDA/src/harvest.cpp:47:16: note: Assuming 'succeeded' is false
               succeeded ? msg_fielddress_success : msg_fielddress_fail ).value_or( translation() );
               ^
.../Cataclysm-DDA/src/harvest.cpp:47:16: note: '?' condition is false
.../Cataclysm-DDA/src/harvest.cpp:46:12: note: Calling 'optional::value_or'
    return SNIPPET.random_from_category(
           ^
.../Cataclysm-DDA/src/optional.h:146:20: note: Assuming field 'full' is true
            return full ? get() : static_cast<T>( other );
                   ^
.../Cataclysm-DDA/src/optional.h:146:20: note: '?' condition is true
.../Cataclysm-DDA/src/optional.h:146:27: note: Calling implicit copy constructor for 'translation'
            return full ? get() : static_cast<T>( other );
                          ^
.../Cataclysm-DDA/src/optional.h:146:27: note: Calling copy constructor for 'value_ptr<std::basic_string<char>>'
.../Cataclysm-DDA/src/value_ptr.h:27:33: note: '?' condition is true
            std::unique_ptr<T>( other ? new T( *other ) : nullptr ) {}
                                ^
.../Cataclysm-DDA/src/value_ptr.h:27:41: note: Memory is allocated
            std::unique_ptr<T>( other ? new T( *other ) : nullptr ) {}
                                        ^
.../Cataclysm-DDA/src/optional.h:146:27: note: Returning from copy constructor for 'value_ptr<std::basic_string<char>>'
            return full ? get() : static_cast<T>( other );
                          ^
.../Cataclysm-DDA/src/optional.h:146:27: note: Returning from copy constructor for 'translation'
.../Cataclysm-DDA/src/harvest.cpp:46:12: note: Returned allocated memory
    return SNIPPET.random_from_category(
           ^
.../Cataclysm-DDA/src/harvest.cpp:46:5: note: Potential leak of memory pointed to by field '_M_head_impl'
    return SNIPPET.random_from_category(
    ^
```

jbytheway:
I think this is good to go, so I've removed the WIP label.

src/harvest.cpp
BrettDong: Looks similar to the type of false positive in https://bugs.llvm.org/show_bug.cgi?id=38176. Switching to `libc++` may help.

jbytheway: It might help, but I'd prefer not to do that, especially not in this PR.  I'd like to keep what clang-tidy is analyzing as close as possible to what we build in the rest of CI.  And I don't think we use libc++ except on Mac.

Longer term, if we add the clang-tidy check that forbids any use of raw new or delete, then we can probably just disable this check.

akrieger (approved):

src/harvest.cpp
akrieger: @jbytheway if you resolve this then someone can merge this ;)

PR #31720, opened 2019-06-22
fix obsolete recipe warnings

alanbrady:
#### Summary
SUMMARY: Bugfixes "Fix removed recipes issuing warning"

#### Purpose of change
Whenever someone makes a recipe obsolete the game issues a warning because it thinks the recipe is invalid but actually we intentionally removed it.

#### Describe the solution
Prevent removing the recipe from `recipe_dict.recipe` if it's obsolete and instead just prevent it from showing up.  We can't just delete it because we need to know later on when we try to make an `obj` from it if it actually doesn't exist or was just made obsolete.  We're already checking if the recipe was obsolete before trying to make an `obj` but it could never hit it because we were deleting it.  `blacklisted` recipes are still straight up deleted and will still issue a warning.

PR #64154, opened 2023-03-11
Squelch debugmsg from missing recipe_ids

irwiss:
#### Summary
None

#### Purpose of change

Squelches debugmsgs arriving from deserializing invalid `recipe_id`s (from obsoletion/deletion) for avatar recipe book and crafting results

#### Describe the solution

Demote debugmsg to DebugLog warning
`recipe_id::NULL_ID()` is used for id instead - no logic should have changed from this, it's equivalent to skipping the debugmsg
Deleted most recipe obsoletions (some are kept; generic guns mod abuses them to disable some recipes)

#### Testing

Load 0.G save where avatar has fabrication>=1 in latest experimental - debugmsg will pop for propane_cooker
Apply patch - should be no debugmsg


anothersimulacrum:
Introduced in https://github.com/CleverRaven/Cataclysm-DDA/pull/31720

I agree that there's no reason we can't just ignore these instead of requiring an entry for them.

PR #64156, opened 2023-03-11
Fix nested sealed containers rot

### LINKED ISSUE #64138: Sealed items spawn old or near rotten
### Describe the bug

Have been finding canned goods and other sealed food items which are listed as both sealed, but also old or going rotten.

Perhaps more nonsensically, you can open an MRE and find that the crackers or cookies have begun to spoil, but the lobster seafood medley is perfectly good. Or the reverse, and the ratatouille is about to go off.

### Attach save file

[Martyr-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10947972/Martyr-trimmed.tar.gz)

### Steps to reproduce

1. start looting food items that are sealed
2. you should eventually find some that are (old) but also still (sealed)

### Expected behavior

Sealed foodstuffs like MREs, canned goods, etc. should be in equally good condition.

I say this, keeping in mind that 'best by' dates in no way reflect the expected time for food to rot; sealed dry or canned goods especially remain shelf stable for years and years past whatever date the FDA mandated be printed on the can.

### Screenshots

image

- OS: Windows
    - OS Version: 10.0.19044.2604 (21H2)
- Game Version: 59c0d6e [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
]

### Additional context

_No response_

Kromgart:
#### Summary
Bugfixes "#64138"

#### Purpose of change

Fix #64138 
Nested item_groups could change the comestibles rot value before being placed in a higher level sealed container, so sealed MREs were getting old and rotten food inside.

#### Describe the solution

Apply rot randomization only on top-level containers after the item_group has been generated and assembled into its containers. Rot randomization is now called explicitly in mapgen's map::place_items and in monster::drop_items_on_death.

#### Testing

Debug-spawned zombie soldiers and killed them. Checked their MREs. Sealed ones now contain only 'fresh' food. Not sealed MREs can have the 'past midlife' / 'old' food.

Teleported into town, checked ~20 kitchens: sealed food was all 'fresh', non-sealed has varying freshness.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -7594,18 +7594,19 @@ void item::set_rot( time_duration val )
 
 void item::randomize_rot()
 {
-    visit_items( []( item * visit_itm, item * ) {
-        if( visit_itm->is_container() && visit_itm->all_pockets_sealed() ) {
-            return VisitResponse::SKIP;
-        } else if( visit_itm->is_comestible() && visit_itm->get_comestible()->spoils > 0_turns ) {
-            const double x_input = rng_float( 0.0, 1.0 );
-            const double k_rot = ( 1.0 - x_input ) / ( 1.0 + 2 * x_input );
-            visit_itm->set_rot( visit_itm->get_shelf_life() * k_rot );
-            return VisitResponse::SKIP;
-        }
+    if( is_comestible() && get_comestible()->spoils > 0_turns ) {
+        const double x_input = rng_float( 0.0, 1.0 );
+        const double k_rot = ( 1.0 - x_input ) / ( 1.0 + 2.0 * x_input );
+        set_rot( get_shelf_life() * k_rot );
+    }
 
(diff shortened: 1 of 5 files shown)


andrei8l:
This would be even better without visitable. You'd also be able to handle the edge case of only one sealed pocket, thought I'm pretty sure that can't happen naturally currently.

Kromgart:
Got rid of visitable.

Also changed the sealed check to use pocket->spoil_multiplier(). Some sealed containers do not provide 100% protection from rot (plastic bottle, gallon jug). And I think it is technically possible for non-sealed containers to provide 100% protection from rot (maybe in magiclysm?). 

pocket->spoil_multiplier() should account for such cases.

PR #64161, opened 2023-03-11
Soldier drop tweaks

IncoherentOrange:
#### Summary
Balance  "Give zombie soldier drops attachments & 40mm grenades"

#### Purpose of change

Make zombie soldiers drop weapons with attachments that match US Army doctrine more closely. Most of the soldiers we find after the Cataclysm probably belonged to infantry squads and would generally be equipped appropriately given the US' stockpiles of materiel. Each squad has two grenadiers out of nine troops, but none of the zombies drop weapons matching the grenadier role, and they don't drop weapons that have attachments that are ubiquitous - particularly the ACOG scope.

#### Describe the solution

Retaining the current rate of weapon drops for zombie soldiers and related monsters, modify drop rates and item groups to include M4 carbines with grenade launchers and a better than average chance for all dropped rifles and machine guns to have ACOG sights. Allow them also to drop 40mm HEDP grenades and smoke grenades in place of regular grenades. Grenade launchers are made as common as M249s, as a similar proportion of troops should have them.

Also add chances for soldiers to drop holographic sights and magnifiers, as well as small chances to drop M72 or other missile launchers (see pages 44 and 644 of below PDF) and to add the M145 Machine Gun Optic for the M249 instead of the ACOG, but I wanted to stick to the basics and focus on the most common weapons and attachments without adding confusion and redundancy.

#### Testing

Spawned several hundred zombie soldiers and sifted through their drops. Found M16 and M27 rifles with and without ACOG scopes, M4s with ACOGs and/or M203/M320 launchers. Found smoke bombs and regular grenades as well as 40mm HEDP in expected proportions.

#### Additional context

I got a lot of the information I based this change on from a US Army field manual currently available here: https://armypubs.army.mil/epubs/DR_pubs/DR_a/ARN13842-ATP_3-21.8-001-WEB-4.pdf
It contains squad composition and equipment information for an Army infantry platoon, and so seemed like a logical thing to base soldier drops on.

### DIFF
--- a/data/json/itemgroups/Weapons_Mods_Ammo/gunmod.json
+++ b/data/json/itemgroups/Weapons_Mods_Ammo/gunmod.json
@@ -149,6 +149,19 @@
     "subtype": "collection",
     "entries": [ { "item": "suppressor" }, { "item": "rifle_scope" }, { "item": "bipod" } ]
   },
+	{
+		"type": "item_group",
+		"id": "military_grenadier_mods",
+		"//": "Default mods for m4 carbines issued with grenade launchers",
+		"subtype": "distribution",
+		"entries": [{
+			"item": "m203",
+			"prob": 40
+		}, {
+			"item": "m320_mod",
+			"prob": 60
+		}]
+	},
   {
     "type": "item_group",
     "id": "vz58v_mods",
--- a/data/json/itemgroups/Weapons_Mods_Ammo/gunmod.json
+++ b/data/json/itemgroups/Weapons_Mods_Ammo/gunmod.json
@@ -149,19 +149,13 @@
     "subtype": "collection",
     "entries": [ { "item": "suppressor" }, { "item": "rifle_scope" }, { "item": "bipod" } ]
   },
-	{
-		"type": "item_group",
-		"id": "military_grenadier_mods",
-		"//": "Default mods for m4 carbines issued with grenade launchers",
-		"subtype": "distribution",
-		"entries": [{
-			"item": "m203",
-			"prob": 40
-		}, {
-			"item": "m320_mod",
-			"prob": 60
-		}]
-	},
+  {
+    "type": "item_group",
+    "id": "military_grenadier_mods",
+    "//": "Default mods for m4 carbines issued with grenade launchers",
+    "subtype": "distribution",
+    "entries": [ { "item": "m203", "prob": 40 }, { "item": "m320_mod", "prob": 60 } ]
+  },
   {
     "type": "item_group",
     "id": "vz58v_mods",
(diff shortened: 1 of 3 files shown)


Rllaman:
I think the reasoning behind this change might be faulty.

It's true that this matches US Army doctrine about how an infantry squad should be equipped in ideal situations. However, the Cataclysm is about as far from an ideal situation as you can get, and as a result it's unlikely that most zombie soldiers you'd encounter would be actual infantry.

When possible, the military would send combat arms out to fight first because they're the ones best equipped and trained to handle a threat. If the situation demands it, then anyone else available can get sent out to work as de facto infantry, even if their normal MOS is to be a cook, mechanic, administrative clerk, supply, etc. They'd be given whatever's available, which might be nothing more than what they've already been issued.

From the game's lore, it seems like there was slightly over a month of moderate unrest that progressively got worse, then 10 days of violent riots, then 5 days of national emergency before the portal storms struck, and then another 10 days until the government fell. 

In an apocalyptic scenario where mass violence breaks out in every town and city in the country, then whatever functioning command remains would probably have every soldier they possibly could pick up whatever they had and try to do something. And many of those would have older weaponry and lack any kind of explosives.

IncoherentOrange:
> I think the reasoning behind this change might be faulty.
Yeah, it definitely could be - the exact condition of the military when things got really bad (and when most of these soldiers died) could use some nailing down - someone could certainly do the napkin math to determine what proportion of zoldiers were probably what and dole out the weapons accordingly with non-combat MOS being much more likely to have naked M16s vs. sighted M4s (+~15% grenade launchers and ~15% M249s) for most regular troops. These figures can be easily modified if someone has the courage, but I feel it's a bit out of scope for the PR - all I really wanted to do is make it so any of these dropped at all.

For now it's about half that were equipped like real infantry and half that are equipped poorly - of the 10% of soldiers that drop weapons,10% of those dropped automatic rifleman kit before this change (well, an M249 at least). 10% will now drop grenadier kit. A few of them drop Marine weapons too (the M27, didn't mess with those much but the changes do make the M38 DMR somewhat rarer). While nowhere near half of troops are infantry, it's hard to say whether the sheer stock of equipment wouldn't counteract that somewhat. 

What I don't think is doubtful is the commonality of sights. Per Trijicon's website they sold their millionth ACOG in 2017, and a heck of a lot of those went to the Army and Marines - the Army appears to use them wherever possible, and the Marines wanted one for each rifle as far back as 2005 from what I can find. It's a fairly durable part due to its design not relying on large lenses like the M68 CCO red dot, as we should probably have a few of those kicking around on these too.

PR #64165, opened 2023-03-11
Add pockets/collar/hood items to body status for relevant parts

sparr:
#### Summary
Interface "Add pockets/collar/hood items to body status for relevant parts"

#### Purpose of change
When viewing the Body Status interface, mouth parts might be Comfortable or Warm in cold environments, while showing no gear covering them, due to some other gear having a collar. This is relatively opaque to the player. Ditto for pockets/hands and hood/head.

#### Describe the solution
This PR adds "nameofitem (collar)" to the "Worn" section when viewing a mouth part that is warmed by the collar without being covered by the item with the collar. Ditto "(pockets)" and "(hood)" for hand and head parts.

I couldn't think of another good way to communicate to the player why the body part in question was warmer than expected.

#### Testing
I've equipped various items with different body part coverage and observed the new entry appearing and disappearing as expected to match the effectiveness of the collar.

#### Additional context
Originally just COLLAR, added POCKETS and HOOD by request.

I also updated two bits of docs to more accurately describe the function of the COLLAR flag

### DIFF
--- a/doc/ARMOR_BALANCE_AND_DESIGN.md
+++ b/doc/ARMOR_BALANCE_AND_DESIGN.md
@@ -563,7 +563,7 @@ RAIN_PROOF | Wont get wet in rain
 HOOD | Keeps head warm if nothing on it
 POCKETS | Keeps hands warm if they are free
 BLOCK_WHILE_WORN | Can be used to block with while worn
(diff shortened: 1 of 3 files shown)


anothersimulacrum:
Would you mind doing this for POCKETS and HOOD too?

PR #64168, opened 2023-03-11
Fix far-sighted butlers spawning without reading glasses

jesse-r-s-hines:
#### Summary

Bugfixes "Fix far-sighted butlers spawning without reading glasses"

#### Purpose of change
If there are wear conflicts on profession starting items without `no_auto_equip`, the second item would be silently dropped from inventory. This is noticeable with a far-sighted butler, the butler's default monocle would conflict with the reading glasses and cause the reading glasses to be deleted. The monocle only corrects near-sighted, so the character can't read until they find glasses. The same will occur with near-sighted Butlers, but the since the monocle corrects Near-Sighted its not an issue.

#### Describe the solution

Change the `add_profession_items` to fall-back to putting items in inventory if we can't auto-wear them instead of silently deleting the items. 

This change does mean that "Near-Sighted" Butler's will spawn with both a monocle and redundant glasses now. We could simply remove the monocle from the Butler instead.

Or, better we could remove the monocle from the default profession items list and modify `profession_item_substitutions` to allow thematic profession substitutions so that "Near-Sighted" butlers spawn with a monocle instead of regular glasses.

### DIFF
--- a/src/newcharacter.cpp
+++ b/src/newcharacter.cpp
@@ -628,8 +628,11 @@ void avatar::add_profession_items()
                 inv->push_back( it );
             }
         } else if( it.is_armor() ) {
-            // TODO: debugmsg if wearing fails
-            wear_item( it, false, false );
+            if( can_wear( it ).success() ) {
+                wear_item( it, false, false );
+            } else {
+                inv->push_back( it );
+            }
         } else {
             inv->push_back( it );
         }


jesse-r-s-hines:
We could allow profession substitutions with something like so:
```json
  {
    "type": "profession_item_substitutions",
    "profession": "main",
    "sub": [
      { "item": "glasses_eye", "new": [ "glasses_monocle" ] },
      { "item": "glasses_bifocal", "new": [ "glasses_monocle", "glasses_reading" ] }
    ]
  },
```

PR #64170, opened 2023-03-11
Fixed pocket contents displaying each item seperately

Skrassh:
#### Summary
Interface "Fixed pocket contents displaying each item seperately"

#### Purpose of change
When in the pocket settings management window pocket contents displayed each item in a stack individually which could cause the contents to go off the bottom of the screen.

#### Describe the solution
Display the stack of items as a stack when in that window.

Do a quick and dirty solution that didn't show the plurals of item names.

#### Testing
Opened up the pocket settings window and looked at contents of the pocket.

#### Additional context
image
image

### DIFF
--- a/src/item_pocket.cpp
+++ b/src/item_pocket.cpp
@@ -1244,7 +1244,7 @@ void item_pocket::contents_info( std::vector<iteminfo> &info, int pocket_number,
                 }
             }
         }
-        for( auto &content : counted_contents ) {
+        for( std::pair<const std::string, int> &content : counted_contents ) {
             if( content.second > 1 ) {
                 info.emplace_back( "DESCRIPTION", space + std::to_string( content.second ) + " " + content.first );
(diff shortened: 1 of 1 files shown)


src/item_pocket.cpp
```diff
@@ -1212,6 +1213,8 @@ void item_pocket::contents_info( std::vector<iteminfo> &info, int pocket_number,
 
     // ablative pockets have their contents displayed earlier in the UI
     if( !is_ablative() ) {
+        // Create map here for item contents, use name as key and value as number held, after traversing the entire pocket dump the map into the vector.
+        std::unordered_map<std::string, int> counted_contents;
```
andrei8l, line 1217: It would be better to use `std::unordered_map<item const *, int>` here, then down below use `item::stacks_with()` instead of string matching, and finally colorize only at the end.

PR #64172, opened 2023-03-11
Remove find_part and engine_fuel_current

irwiss:
#### Summary
None

#### Purpose of change

Remove awkward functions off vehicle class

#### Describe the solution

Working on a bigger PR for vehicle code I'd like to get these 2 out of the way;

`vehicle::find_part` - goes on a roundabout searching via comparing base part pointers when a reference to the vehicle_part is directly available and `vehicle::index_of_part` exists

`vehicle::engine_fuel_current` - used once, seems to only exist because engine part indexing is awkward, and no reuse potential after the PR i'm working on


irwiss:
Obsoleted by https://github.com/CleverRaven/Cataclysm-DDA/pull/64179

PR #64174, opened 2023-03-12
Add two travel-related overmap keybinds for a better UX

strategineer:
#### Summary
Interface "Add two travel-related overmap keybinds for a better UX"

Add two new keybinds GO_TO_DESTINATION and CENTER_ON_DESTINATION that allow the player to:
- Go to the current travel destination without the cursor needing to be on it.
- Move the cursor to the current travel destination.

#### Purpose of change

I was getting annoyed about constantly having to open the overmap and pan over to my destination over and over again when traveling somewhere to resume traveling after enemies showed up.

#### Describe the solution

- Added the keybinds
- Refactored the code used by the CHOOSE_DESTINATION keybind into a function so that I can call it from the GO_TO_DESTINATION keybind code.

For next steps, I'd like to make the GO_TO_DESTINATION keybind work directly from the game world as well as the overmap. Although that seems like a bigger change.

Let me know if you've got a better idea for the default values for the keybinds.

#### Testing

- Tested the two keybinds and ensured that they're doing what's expected.
- Checked that the UI hints were displaying well.
- Ensured that the two keybinds' default values aren't overlapping with any other overmap keybinds. 

#### Additional context

Overmap screen with new keybinds being displayed:
cataclysm-tiles_kzvH8Yd1ed

GO_TO_DESTINATION key pressed when the cursor is NOT on the destination:
cataclysm-tiles_4fe0I3zeD3

GO_TO_DESTINATION or CHOOSE_DESTINATION key pressed when the cursor is already on the destination:
cataclysm-tiles_Ue8uNhXXyb

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -1209,6 +1209,18 @@
   },
   {
     "type": "keybinding",
+    "id": "GO_TO_DESTINATION",
+    "name": "Go to destination",
+    "bindings": [ { "input_method": "keyboard_char", "key": "g" }, { "input_method": "keyboard_code", "key": "g" } ]
+  },
+  {
+    "type": "keybinding",
+    "id": "CENTER_ON_DESTINATION",
+    "name": "Center on destination",
+    "bindings": [ { "input_method": "keyboard_char", "key": "c" }, { "input_method": "keyboard_code", "key": "c" } ]
+  },
+  {
+    "type": "keybinding",
     "id": "FIRE",
     "category": "TARGET",
     "name": "Fire weapon",
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -1211,13 +1211,13 @@
     "type": "keybinding",
     "id": "GO_TO_DESTINATION",
     "name": "Go to destination",
-    "bindings": [ { "input_method": "keyboard_char", "key": "g" }, { "input_method": "keyboard_code", "key": "g" } ]
+    "bindings": [ { "input_method": "keyboard_any", "key": "g" } ]
   },
   {
     "type": "keybinding",
     "id": "CENTER_ON_DESTINATION",
     "name": "Center on destination",
-    "bindings": [ { "input_method": "keyboard_char", "key": "c" }, { "input_method": "keyboard_code", "key": "c" } ]
+    "bindings": [ { "input_method": "keyboard_any", "key": "c" } ]
   },
   {
     "type": "keybinding",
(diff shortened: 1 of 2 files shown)


data/raw/keybindings.json
```diff
@@ -1207,6 +1207,18 @@
     "name": "Choose destination",
     "bindings": [ { "input_method": "keyboard_char", "key": "W" }, { "input_method": "keyboard_code", "key": "w", "mod": [ "shift" ] } ]
   },
+  {
+    "type": "keybinding",
+    "id": "GO_TO_DESTINATION",
+    "name": "Go to destination",
+    "bindings": [ { "input_method": "keyboard_char", "key": "g" }, { "input_method": "keyboard_code", "key": "g" } ]
```
Qrox, line 1214:
```suggestion
    "bindings": [ { "input_method": "keyboard_any", "key": "g" } ]
```

data/raw/keybindings.json
```diff
@@ -1207,6 +1207,18 @@
...
+  },
+  {
+    "type": "keybinding",
+    "id": "CENTER_ON_DESTINATION",
+    "name": "Center on destination",
+    "bindings": [ { "input_method": "keyboard_char", "key": "c" }, { "input_method": "keyboard_code", "key": "c" } ]
```
Qrox, line 1220:
```suggestion
    "bindings": [ { "input_method": "keyboard_any", "key": "c" } ]
```

src/overmap_ui.cpp
```diff
@@ -1695,6 +1697,44 @@ static std::vector<tripoint_abs_omt> get_overmap_path_to( const tripoint_abs_omt
...
+    if( query_yn( confirm_msg ) ) {
+        if( driving ) {
+            player_character.assign_activity( player_activity( autodrive_activity_actor() ) );
+        } else {
+            player_character.reset_move_mode();
+            player_character.assign_activity( ACT_TRAVELLING );
+        }
+        return "QUIT";
+    }
+    return nullptr;
```
BrettDong, line 1735: Why return a `nullptr` here?

strategineer: No particular reason (unfortunately). Looking back at the code I think it would be cleaner to return true/false in this function and set the `action` variable to "QUIT" conditionally. Although we've still got a function that returns a value and has a side-effect which isn't ideal... Let me type up a quick PR before I forget to refactor this.

BrettDong: I meant constructing a `std::string` from `nullptr` is going to crash the program here.

strategineer: Huh interesting, it wasn't crashing on my end when i tested it. But I'm not surprised that it could. Thanks for keeping an eye out!

BrettDong: If you press N to the prompt you'll reach here and access `nullptr` memory.

PR #64182, opened 2023-03-12
electric-jackhammer-power-tweaks

flybye3o6:
#### Summary
Balance "tweaks to electric jackhammer's power costs"

#### Purpose of change
a few constructions had there fuel cost set to 50 for both gas and electric jackhammers set out to find what the ratio for fuel cost should be for the 2 jackhammers

#### Describe the solution
the gasoline one seemed to be spot on so used that as the baseline
used a 2200w electric jackhammer common for demolition and construction for its stats
adjusted the power cost based on the gas powered one at a ratio of 5gas/66kj
set its battery capacity to 1 hour use and adjusted its action to 30min use same as the gas powered one

could set the max stored power to the same as the `medium storage battery` used in its crafting


RAldrich:
Would it make sense to eliminate the UPS from this and make it run on regular batteries while you're at it?

UPS is pretty archaic and is only used by this jackhammer, smartphones, and the combat exoskeletons. It makes no sense to have a wirelessly rechargable jackhammer.

flybye3o6:
> Would it make sense to eliminate the UPS from this and make it run on regular batteries while you're at it?

thats out of the scope of this PR, but ill keep it in mind
i did do a bit of research and there are cordless electric jackhammers but they are not great or common 
something i wanted to do is have the electric jackhammer 26kg you can find be corded needing a power grid to plug into
(people were asking for the ability for extension cords to plug into electric items in your inventory and plug into a grid so ill wait for that (if possible))
and a craft to have a makeshift modification to add a swappable power pack but have some draw backs (less efficiency, more weight and volume)
also there are smaller electric jackhammers 10kg i could add and use heavy tool batteries, they are not rated for much but they could be used to break locks and light crafting, they wouldnt be able to take down a concert wall in 30mins and only have enough power for about 15mins

RAldrich:
Sounds good.

If we get functionality for cords plugging into things, that should let us get rid of UPS entirely. My understanding is that UPS was originally created in large part because it was hard to design a system for "what is plugged into what" in inventory, and it was easier to have a magic power field that just recharges everything you're holding.

MoonlitRiverlight:
Battery jackhammers rely on having other batteries in rechargers and rotating them as you empty the one currently in the machine. They're not meant to do a huge contruction work in a single battery load. This might be a moot point for us since UPS is leaving at some point, but the interrupted construction system could be used to reflect the irl case. Do the battery load, get x% done, swap the batteries, continue.

PR #58180, opened 2022-06-05
Parametric basecamp expansions

jbytheway:
#### Summary
Features "Parametric basecamp expansion definitions"

#### Purpose of change
We have lots of basecamp expansion definitions that use almost exactly the same layout, but differ only in the materials used.

Currently each requires an entirely separate recipe and mapgen definition.

It would be nice to take advantage of parametric mapgen to consolidate these similar definitions into a single one.  This would dramatically cut down on JSON.

Fixes #45098.

#### Describe the solution
Allow the mapgen definitions for basecamp expansions to use parameters.

The human-readable versions of those parameters must be defined in the recipe to appear in the UI.

The game will consider each possible combination of values for the parameters, and the Cartesian product of these options will appear in the UI.  That has the potential to explode the options, but hopefully people won't abuse it.  I imagine in most cases there will be only a single parameter (which defines the palette).

I chose one pair of similar expansions (`fbmh_2_log_room_1_1` and `fbmh_2_resin_room_1_1`) and consolidated them into a single genericised version (`fbmh_2_generic_room_1_1`) as an example.

A `basecamp_upgrade` object now must contain `mapgen_arguments` to convey the particular choice made by the player.

I also fixed parametric sub-palettes used inside palettes.  This was theoretically supported before, but didn't actually work.

#### Describe alternatives you've considered
The use of recipes to represent basecamp upgrades is a bit of an abuse.  This change is probably making that particular mess worse.  It would be nice to factor out the common functionality and make two separate classes, but that would be a bigger change.

#### Testing
I tried the new upgrades.  They take a reasonable amount of time and consume resources as expected.  But there are a million potential corner cases, so more testing is probably warranted.

@PatrikLundell I believe it was you who was wanting to use this feature.  Are you able to take a look at this approach and see if it seems reasonable to you.  Even better, if you are able to build this branch perhaps you could try using it yourself and see if you run into any problems?

I still haven't written any documentation; feel free to tell me you're going to wait until that's done.

I don't expect this feature to get merged during feature freeze, so there's no particular hurry.

@curstwist I'm also interested in your opinion as our resident mapgen guru.

#### Additional context
Some before-and-after shots of the upgrade UI:
parametric-basecamp-before
parametric-basecamp-after
The UI for the new parametric case could certainly be improved, but that's a problem for a future PR.

PR #64183, opened 2023-03-12
parameterized field base camp version 2

PatrikLundell:
#### Summary

None

#### Purpose of change

Main purpose:
- Make use of the ability to use palette parameters in mapgen/recipes to reduce the size of the JSON for field base camp version 2.

Side purposes:
- Remove bed/cooking construction of the first building as prerequisite for further construction.
- Add the option of making makeshift beds (made easy by the palette parameters).

#### Describe the solution

Massaging the construction recipes into palette parameter based ones.

Leave the new functionality implemented for this purpose unused.

#### Testing

The main testing was performed while testing the functionality (#58180), but a quick sanity check constructing the whole base camp with "random" materials for different parts was done before producing this PR.

- Verify that all construction materials actually can be used (but possibly not for every individual segment, although that probably was done before the PR went into the deep freezer last year).
- Verify that all 3 first building cooking facilities can be constructed.
- Verify that all 3 bed options can be constructed. Every option for every position has not been tested, though, but rather all permutations for the first room and "random" ones for the rest of them.


PatrikLundell:
An unrelated comment: After finding that the binding of the JSON uglifier (json_formatter.exe) was missing from Visual Studio as the result of an upgrade that resulted in reinstall a while ago, I manually applied the uglifier to all of the files modified and pushed the result (which is what's in this PR). I then located [https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_STYLE.md](url) and followed its instruction for restoring the "Lint All JSON" tool and then applied the uglifier. The result was that a bunch of starting with .github/vcpkg_ports/sdl2/vcpkg.json was modified. Is this intended? What happens when I inevitably push those changes after having made JSON changes in a future PR?

jbytheway:
I think having those files linted is an unlucky coincidence of the current situation.  The powershell script lints any file that's different in your branch and either `master`, `origin/master`, or `upstream/master`.  Recently there was a PR to change some of the vcpkg stuff, so those files were different in one of those branches when they wouldn't normally be.  If you update `master` on your fork and rebase your branch then it shouldn't happen any more.

Still, I agree that behaviour isn't ideal.  If you know how to wrangle powershell then I'm sure a PR to improve the situation would be welcome :).

PatrikLundell:
Thanks for the answers.

I'll definitely update my master before making any new branches (that is, or should be, standard practice). In fact, I believe my branch was produced immediately after updating master (and I remember I did update it), so any external changes would have happened during a fairly short time measured in hours.

However, even if these normally stable files are changed, why aren't they're subjected to the formatting rules applied to everything else? If they were, no requirements to change the files would occur, and the only side effect might be additional time for the processing of these files (which is rather substantial anyway, so it's fired off and then it's time to do something else in the mean time).

I am, unfortunately, not good at using any kind of shell scripting. Looking at the script I can get a gist of approximately what it's doing, but that's about it.

PatrikLundell:
I very much doubt the field tests failed due to my addition of a word in a string (the tests succeeded before that commit)...

PR #64186, opened 2023-03-12
Correct description of 'Hooves' mutation

Terrorforge:
#### Summary
Bugfixes "Correct description of 'Hooves' mutation"

#### Purpose of change

The Hooves tooltip says that it "allows kicking attacks to do much more damage", which is incorrect and confusing. Hooves is an additional attack just like the other offensive mutations, it has no effect whatsoever on standard kicks.

Edit: the part about being "unable to wear shoes of any kind" is also wrong as you can wear OVERSIZE/XL shoes, so I changed that too.

#### Describe the solution
Changed the description to clarify that it lets you do a special attack instead of saying that it enhances anything.

Change the effect to match the description. This should probably happen, because it's super weird that you kick for like 7 damage then immediately super extra hoof kick for like 40 damage. But that's touching on the ways that mutation attacks are just super weird in general and would also be a lot more work, so for now I'm just changing the description to be less misleading.

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -3750,7 +3750,7 @@
     "visibility": 2,
     "ugliness": 2,
     "mixed_effect": true,
-    "description": "Your feet have fused into hooves.  This allows kicking attacks to do much more damage, provides natural armor, and removes the need to wear shoes; however, you cannot wear shoes of any kind.  Reduces wet effects.",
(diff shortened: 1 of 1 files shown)


adamkad1:
Honestly it should be a bonus to kick attacks but idk if mutations can do that. How do you do a crappy kick with a hoof?

Terrorforge:
Add an integrated armor that goes on the feet and has decent bash bonus.
It's not that complicated on a technical level, but there's lots of balance considerations. Like do you do that instead of the bonus attack? In addition? How much damage? How do you handle the fact that rigid foot armor really would make it impossible to wear shoes? I don't know the best way to implement that, but feel free to have a cracking if you think you've got a good idea.

Issue #51228, opened 2021-08-30
[0.F-0] Towed vehicles rear end the tower when stopping, and start out set at silly speed


PatrikLundell:
### Describe the bug

Two bugs that may well be related.
When towing a vehicle, the towed vehicle rear ends the tower as often as not when stopping, even if the speed is reduced gradually, or never having been above minimum.

When taking control of a formerly towed vehicle it tends to start out with a silly speed set, such as a value in the hundreds, rather than the expected 0.

### Steps To Reproduce

1. Connect a towing vehicle to a towed vehicle.
2. Drive straight ahead at minimum speed.
3. Stop after towing the vehicle for at least some distance.
4. Either get rear ended or try again until you do.
5. Enter the towed vehicle and start it. Check the speed it's set to reach: it's typically not zero.

### Expected behavior

When speed is reduced gradually, the towed vehicle should roll to a stop before reaching the towed vehicle on any surface (one could argue that grass would be rougher than furrowed earth, but the towed vehicle should stop in time even on a road).

The target speed for a towed vehicle should be reset to zero when the pulling force ceases (i.e. when the cable slackens).

### Versions and configuration

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: bc28921 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Bionic Professions [package_bionic_professions]
]

### Additional context

I've tried to increase the tower vehicles speed from 0 to minimum when the towee approaches to within two tiles to give the towee additional room to stop, but that does not help: the code probably doesn't model a slack cable not propagating any force. This is OK as long as the towee stops in time.

My last towing resulted in the rear ending vehicle hitting and damaging the tow cable part of the tower...

There are also wild swerving and strange catch ups of the towed vehicle (typically causing any wing mirrors on the towed vehicle to be lost by the end of the trip), but these are minor issues: it's impressing towing works at all.

Maleclypse:
This is only considered a bug if you are experiencing this when slowing down in increments of five mph or less.

PatrikLundell:
As mentioned, I've experienced it when never even increasing speed beyond minimum, and I've never slowed down faster than one step per tick (nor have I accelerated faster). The towing vehicles haven't been dragsters, and the speed increments seem to be about 6 km/h, which is a fair bit less than 5 mph.

PR #53596, opened 2021-12-19
Stop towed vehicles from rear-ending tow-ers

PatrikLundell:
#### Summary

Bugfixes "stop towed vehicles from rear-ending tow-ers when stopping"

#### Purpose of change

Fixes #51228

Towed vehicles were (almost?) always rear-ending the tow-ers regardless of the speed at which they were towed, as they rolled to a stop while slowing down very slowly when the tow-er was stationary. The reason was that code that normally added a constant braking factor wasn't applied to towed vehicles, In addition to that, vehicle wheel friction remains the same regardless of the terrain and the weight of the vehicle.

#### Describe the solution

This PR addresses the first problem above by applying the standard braking factor whenever the pulling cable wasn't fully extended. This also means you're less likely to be hit by the pulled vehicle while turning (the pulled vehicle still swings when you turn, as it should).

#### Testing

Pulling a road roller backwards (because that's how I happened to hook it up) with a tow truck and stop from minimum speed, driving around at comparatively high speed while turning (sensibly), and reducing speed gradually to see that the towed vehicle still behaved in a reasonable manner. Reducing speed quickly still gets you rear-ended, which is reasonable.

#### Additional context

I tried to address the issue of the cruise speed getting set at weird values in towed vehicles, but failed to find a good solution.
The cause of the problem is that the cruise control gets enabled when the tow cable starts to move the towed vehicle, at which time the cruise speed is set as well. The cruise speed doesn't seem to be changed from then on, and thus remains when the towing has ended.
Not enabling the cruise control results in a jerky movement where the vehicle jumps a number of tiles whenever the cable gets taut and then waits for it to tighten again, which isn't acceptable.
I've been unable to find a suitable trigger for disabling the cruise control, as stopping is basically an absence of actions.

PR #64188, opened 2023-03-12
Don't use `cruise_thrust()` on towed vehicles

cake-pie:
#### Summary
Bugfixes "Don't use `cruise_thrust()` on towed vehicles"

#### Purpose of change
Fix towed vehicles getting their cruise velocity set to ridiculous values, as reported in #51228:

> When taking control of a formerly towed vehicle it tends to start out with a silly speed set, such as a value in the hundreds, rather than the expected 0.

(H/T @PatrikLundell)

#### Describe the solution
Towed vehicles utilize `vehicle::selfdrive()`.
In `vehicle::selfdrive()` check for a towed vehicle in which case set `thrust()` rather than `cruise_thrust()`

Towing code could use some work, but this will have to do for now to clear a blocker for #64056

#### Testing
Debug spawn a tow cable and a couple of vehicles of appropriate size -- I picked a road roller and a beetle -- towed the latter with the former, then check that cruise velocity of the latter is zero and not some unexpected number.

### DIFF
--- a/src/vehicle_move.cpp
+++ b/src/vehicle_move.cpp
@@ -1310,7 +1310,7 @@ void vehicle::selfdrive( const point &p )
     }
     if( p.y != 0 ) {
         int thr_amount = 100 * ( std::abs( velocity ) < 2000 ? 4 : 5 );
-        if( cruise_on ) {
+        if( cruise_on && !is_towed() ) {
             cruise_thrust( -p.y * thr_amount );
         } else {
             thrust( -p.y );


PatrikLundell:
Thanks for getting rid of this bugger!

cake-pie:
Test failure unrelated -- [fail to cook oatmeal using vehicle rig](https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4398163395/jobs/7701882484#step:15:345).
Tests passed when run locally.

Raikiri:
If I understand correctly then in theory using cruising should preferrential when decelerating, because if you match the speed of the towed vehicle to the vehicle that tows, then they should decelerate at the same rate. However, if you only use thrust and the towing vehicle decelrates, the only thing stopping the towed vehicle would be friction which might be much slower.

So maybe it's worth investigating why the towed vehicles get ridiculous cruise speeds to start with, rather than changing it to a ridiculous thrust?

cake-pie:
There's plenty of room for improvement in towing code but this was not the PR for it.

This only sought to be a non-disruptive change that stops towing from modifying cruise velocity values in a way that breaks player expectations and other code.  In `selfdrive()`, `thrust()` is already used for non-towing contexts when cruise control is off, and towing behavior appears to work correctly with it substituted in place of `cruise_thrust()`.  Also, towed vehicles are presently being stopped by friction anyway (#53596) so there's no substantive change in that regard.

My best guess is that when towing was implemented, it just leveraged selfdrive as a convenient way to move the towed vehicle where it needed to go, without realizing that there were unintended side effects.

If cruise velocity were to be (ab)used for anything other than actually being set by a character who is driving the vehicle, then it needs to be reset back rather than left dangling at arbitrary values.

> changing it to a ridiculous thrust

Is that actually the case?
Towed vehicles aren't gaining ridiculous speed and careening into the back of the towing vehicle at the slightest tug of the tow cable.

Raikiri:
> Towed vehicles aren't gaining ridiculous speed and careening into the back of the towing vehicle at the slightest tug of the tow cable.

I'm just following what's said in #51228:

>When taking control of a formerly towed vehicle it tends to start out with a silly speed set, such as a value in the hundreds, rather than the expected 0.

So I assumed there was something wonky with how target cruise speed was calculated.

PatrikLundell:
I don't understand why the target speed is set high, but my guess is that it's done to ensure that brakes aren't applied. The only thing that should slow a towed vehicle is friction, so a sharp reduction of speed by the tower should  result in the tower being rear ended.
The vehicle is towed by a simple cable, and there's nobody in the towed vehicle that can apply the brakes when needed (or even steer the vehicle). This contrasts with simple trailer towing arrangements where there aren't any brake control in the trailer, but the connection is rigid and the trailer is much lighter than the tower (assuming the laws are adhered to), so the trailer is slowed simply through its failure to maintain momentum by pushing through the tower.

Raikiri:
I should say that towing a vehicle with a cable when there's nobody sitting in it is illegal for a good reason (because you can't make it not rear-end you). So I think it's more feasible to be simulating a tow bar rather an a tow cable.

image

PR #64193, opened 2023-03-12
[Magiclysm] Fix fist weapons

Terrorforge:
#### Summary
Bugfixes "[Magiclysm] Fix fist weapons"

#### Purpose of change
The enchanted unarmed weapons Stormfist and Flaming Fist aren't working properly. Their enchantment is applied even when the player is holding a weapon, and if you wear two, the damage bonus stacks.

#### Describe the solution
Made them both ONLY_ONE and added a condition so the enchantment is only active when the player is not wielding a weapon.

Also made Stonefist ONLY_ONE, mainly for consistency. It's just a gauntlet with a high damage bonus so wielding a weapon or wearing multiples doesn't cause problems like with Stormfist, but most summoned armor is ONLY_ONE and now so is Stormfist, so it seemed reasonable to make it work the same way.

Just let them do that, but it seems counter to the design intention, and also pretty busted.

#### Testing
Spawned the various items and some zombies to punch, checked to see if I could wear multiple, checked that I got a noticeable damage bonus when fighting bare-handed but not when smacking zombies with a spoon.

#### Additional context
May cause wonkiness when loading a save that already had one of these items. Seems existing items keep the old version of the enchantment.


Terrorforge:
Nvm, for some reason this crashes the game after you load and try to wear them. Making this a draft until I figure out what's going on

Terrorforge:
It turns out this is due to a deeper bug with the game not handling enchantments with dialog conditions properly. I'll undraft this after @Ramza13 submits their fix.

Terrorforge:
The problem has been fixed. Pelase merge this after #64206

PR #64197, opened 2023-03-12
Rework (and nerf) melee combat practice recipes.

Aerin-of-the-Toast:
#### Summary
Balance "Rework melee, unarmed, dodge, cutting, stabbing, and bashing practice recipes."

#### Purpose of change

The practice recipes for melee were *very* powerful. Being able to reach melee 3, piercing/cutting/bashing 4, and dodge 3 without any instruction is overkill. As they say, practice without guidance is just reinforcing your mistakes. On the opposite end of the spectrum, there was no way to practice from 0 to 1 in most of these skills without a book. This made a strange meta of ineffectually flailing at zombies with 0 skill until you got to 1, then running home to practice up to 3 or 4. Not only did this rocket your accuracy up a lot, but getting to skip to melee 3 unlocked Feint in the brawling tree, which is very powerful. This made most of the early game skill progression moot and you'd basically start every run moderately competent at melee.

#### Describe the solution

Melee, unarmed, and dodge can now be trained to 1 from the start. This is representing someone that's never done any fighting in their life taking some practice swings or practicing some side hops or dodge moves (they think) will be useful, but not knowing how to improve beyond "jump sideways without tripping over your own feet." While this is trivial to do now since practicing from 0 to 1 takes about 3 in-game minutes, this'll actually take a reasonable amount of time once skill gain XP overall gets reworked.

The melee 3 and unarmed 3 recipe now requires a book. This represents the book teaching general flow drills, stances, and other techniques a random office worker wouldn't know without guidance.

Dodge practice beyond 2 is only available in the dancing book and represents learning an advanced type of footwork that has some application in not falling over. Learning practical dodge beyond that should require active combat, or at least sparring with an active participant. We don't have a way to specify "friendly NPC" as a tool for crafting, or any easy "sparring" mechanic though, so sparring will have to wait for later. Playing wallball was removed as there's not much you could learn doing that you couldn't learn just practicing dodges normally or from your natural human instinct. Dodging in combat is a lot more complicated than just "move out of the way."

The 3 weapon-specific skills (stab cut and bash) all require one of their respective magazines or books now, and their recipes have been changed around a bit to be more interesting instead of the old formulaic "training weapon and dummy to 3, real weapon and armored dummy to 4" method. Notably, the beginner versions of the crafts only bring you to level 2, and all the level 2 versions of the skills are essentially the same as the melee 3 recipe, involving just drills in open air with suitable weapons. No dummies needed there. As a general rule, the level 4 practices need the proper manual while the level 2 practices can use the relevant magazine.

Cutting 4 now involves setting up ad-hoc cutting stands and slicing them with a variety of cutting tools, which is a traditional way to practice your edge alignment and cutting power. We're using rolled-up newspaper instead of tatami mats, but the principle is the same.

Bashing 4 requires a heavy bludgeon of some form and an armored training dummy to withstand the blows you're hitting it with.

Stabbing got quite a few recipes, since its one skill governing knives, rapiers, and spears. There's a level 2 version of each, each using appropriate books even when the IDs might make it seem funny ("manual_cutting" is a bunch of knife-fighting books, for instance, so the knives version of the stabbing training can use that one). The fencing manual is for learning the martial art, so there's only a knife and spear version of the level 4 practice recipes. Only a regular training dummy is needed here, since you're only poking at it with a stick or knife.

All recipes that require a book are also auto-learned at the maximum skill level they can give as well. E.G. The practice recipe that can get you to piercing 4 is also auto-learned at piercing 4. This is for skill rust reasons; if you fight enough zombies and get punched in the mouth enough to figure out what does and doesn't work well enough to hit a certain level in a skill, you should reasonably know what you're doing by that point to practice and keep your skill at that level even if you stop fighting for a bit.

TL;DR: Books, man. You need books for anything past melee 1, unarmed 1, or dodge 1.

Spending a bajillion hours either making a "knife" "sword" etc. quality for every compatible weapon for some of the practices and/or making the crafting list ultramassive to account for every knife you could train with. I tried to make sure the chosen weapons were all ones you could easily get or probably had if you were doing the practice, regardless of your circumstances, but there's always a risk of some dude stuck in a lab who only specifically has access to carving knives but not chef knives I guess.

#### Testing

Checked in-game to make sure the recipes were in properly, spawning from books, and working correctly.

#### Additional context

I generally tried to be as lenient as possible with the tools and components needed for the practice recipes. Under the old system it was mostly gated by your access to wood sawing to make a proper training dummy (even though a log would have probably worked), or getting a *real* weapon for the level 4 weapon skills. Now, the primary bottleneck should be the actual knowledge of what to do and not to do, not your access to a Real Spear instead of a makeshift spear, or locking dodge practice behind finding a soccer ball before you could learn how to jump sideways.


Termineitor244 (approved):
Looks great to me! A better work than what I could have probably changed, I was never fully satisfied with my work in the melee practice recipes, but I couldn't think of how to improve them...

Glad to see you improving them like this! Now we only need a way to decrease the Exp received from the recipes and they will be perfect.

TheShadowFerret:
>Stabbing got quite a few recipes, since its one skill governing knives, rapiers, and spears

Planning to add proficiencies to each of those if we get combat profs? 👀

Issue #64198, opened 2023-03-12
[Magiclysm] Summoned items don't have length


Terrorforge:
### Describe the bug

Despite having a longest_side specified in their item definitions, items that are summoned via magic have no length. This leads to absurd situations such as being able to put a "great sword of judgement" (same size as a zweihander) in your backpack.

### Attach save file

Trivially reproducible

### Steps to reproduce

1. Learn and use any spell that summons an item

### Expected behavior

Items have length, even when conjured magically.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.2604 (22H2)
- Game Version: eb8e722 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Magiclysm [magiclysm],
    Mythical Martial Arts [MMA]
]

GuardianDll:
Interestingly, but its not true, its something else
my game contain `"longest_side": "70 cm",` for `stormhammer`, but despite this, spawning it using debug shows it still has no length
![image](https://user-images.githubusercontent.com/67688115/224568604-cd93705c-3a75-4b03-a258-fb7df4cef44b.png)

Terrorforge:
Hrm. But there seems to be nothing else connecting them? And it's happening to at least items in ethereal_items.json and the sword of judgment in the Crusader spell file. But not every summoned item - the pocket sun has a length, despite having zero volume. Something wrong with the materials, maybe?

Rewryte:
Done some testing on this. The item::length() function will give a length of 0mm for all items that are made of a liquid material. As concentrated mana is considered a liquid, any items made of it will have 0 length. By changing the stormhammer to steel, I was able to summon one with the correct length of 70cm.

GuardianDll:
And liquid it because of what? This exact bug occured when hardened mana got `rigid:false` boolean, not sure how it's related to liquid materials

Rewryte:
Sorry spoke too soon. It's the `"soft": true` property. Turning that off in the concentrated mana material will make item length behave properly.

GuardianDll:
[...]

PR #64118, opened 2023-03-10
Refactor My Mutations

bombasticSlacks:
#### Summary
None

#### Purpose of change
Mutations given by enchantments weren't providing enchant effects.

#### Describe the solution
Added a helper function to grab both my_mutations and the enchant mutations your character has.

Added it to character routines

#### Describe alternatives you've considered
spreading this throughout mutation but as of right now it just seems like that is asking for trouble. I don't know the code well enough.

#### Testing
Still building draft for now

PR #64199, opened 2023-03-12
[Magiclysm] Make the Crusader attunement's Sword of Judgement sheathable

Terrorforge:
#### Summary
Bugfixes "Make the Crusader attunement's Sword of Judgement sheathable"

#### Purpose of change
None of the forms of the Sword of Judgement could fit into the normal sheaths/holsters for weapons of that type. Most summoned weapons can, so I see no reason this shouldn't be the case for the Sword of Judgement.

#### Describe the solution
Added the appropriate flags for the various forms. SHEATH_SWORD for the sword and great sword, SHEATH_KNIFE for for the dagger, SHEATH_SPEAR (spear strap) for qiang, BELT_CLIP for mace.

In order to avoid problems with transforming the weapon while it's in a sheath that can't hold its new form, I made the activations require wielding. This is probably a necessary change anyway, because it's already possible to put e.g. the dagger form in a bag and then transform it into a zweihander too big for the bag.

#### Testing
Made sure the various forms fit into appropriate holsters and can't be transformed while holstered or otherwise stored.

#### Additional context
Due to #64198, some of the forms can go in nonsensical places, but it seems that by coincidence they still fit in the correct sheathes due to weight and volume. The sword of judgement (longsword sized) needs a baldric or back scabbard, great sword needs a back scabbard, dagger (kukri-ish) fits in a sheath but not an ankle sheath, etc.

### DIFF
--- a/data/mods/Magiclysm/Spells/attunements/Crusader.json
+++ b/data/mods/Magiclysm/Spells/attunements/Crusader.json
@@ -117,7 +117,7 @@
     "material": [ "concentrated_mana" ],
     "weapon_category": [ "MEDIEVAL_SWORDS", "LONG_SWORDS", "GREAT_SWORDS" ],
     "techniques": [ "RSTRIKE", "WBLOCK_2", "BRUTAL", "RAPID" ],
-    "flags": [ "UNBREAKABLE_MELEE", "NONCONDUCTIVE", "NO_REPAIR", "NO_SALVAGE", "MAGIC_FOCUS", "TRADER_AVOID" ],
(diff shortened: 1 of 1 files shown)


GuardianDll:
Actually I'd prefer to have some integral pocket for this weapon, or integrated belt, but in this case the code think its a clothing, and equip it instead of spawning it in your hands

Terrorforge:
Add an extra_effects spell that summons a separate holster of some kind? Maybe give it some minor enchantment so it doesn't feel like it's just sitting there.

GuardianDll:
I know what to do, write it in 20 minutes

GuardianDll:
No

After some tests i found my solution doesn't work, and require #64118 first, so make your version at first, i'll do my additional version afterwards

Terrorforge:
Do you mean you want this version, that just makes them fit in normal sheaths, or the thing where it summons a separate holster? Because I don't think I have time to do the latter

GuardianDll (approved):
I mean "this one is good, don't worry much"

PR #64208, opened 2023-03-12
Fixing some book stuff

GuardianDll:
#### Summary
None
#### Purpose of change
Accidentally found some stuff has duplicated or wrong ids, and some stuff had a wrong order
#### Describe the solution
Replaced the ids with proper one (they are duplicated, so no reason to obsolete them), and added some books to fill holes in order (so mag_cutting, mag_cutting_1, mag_cutting_2, **mag_cutting_3**, mag_cutting_4 etc)
#### Additional context
I wonder why the game is able to spawn two variants with the same name, and do not throw duplicated id error


data/json/items/book/cutting.json
```diff
@@ -32,6 +32,16 @@
         "fun": 0,
         "weight": 1
       },
+      {
+        "id": "mag_cutting_3",
+        "name": { "str": "Basics Of Stage Combat", "str_pl": "copies of Basics Of Stage Combat" },
+        "description": "Foundation of using single handed sword on stade and on screen, with different tricks how to make a good show of it.",
```
GuardianDll, line 38:
```suggestion
        "description": "Foundation of using single handed sword on stage and on screen, with different tricks how to make a good show of it.",
```

PatrikLundell:
Does the game actually produce both items with the same name? In the past duplicate definitions just led to one of them overwriting the other randomly so you'd get one or the other (and it wouldn't consistently be one of them). Duplicate detection might have improved since then, though.

GuardianDll:
Yeah, thats the point - i was able to spawn both this books, and the game separated them just as "mag_dodge" and "book_nonf_hard_dodge_tlwd"

PR #64219, opened 2023-03-13
Display remaining weight/volume stats of gear/container when inserting into it

### LINKED ISSUE #64104: Show weight/volume info when inserting

Open up the inventory and select gear or a container that can hold items and insert items. This does not show any info regarding the items storage capacity.
Screenshot_20230310-092755_Cataclysm DDA (experimental)

### Solution you would like.

Show the capacity in the corner of the screen.

_No response_

### Additional context

_No response_

strategineer:
#### Summary
Interface "Display remaining weight/volume stats of gear/container when inserting into it"

#### Purpose of change

Seems like good UX to remind the player how much weight/volume the container they're trying to insert into can handle initially as well as when they start selecting items to insert.

Fixes #64104

#### Describe the solution

Nothing crazy here, just making use of existing code and displaying the weight/volume stats at the top-right of the insert inventory menu.

I do this by checking to see whether the selected items will be holstered or not and then update all the stats accordingly as the player selects items.

None although let me know if there's anything I can improve here.

#### Testing

Spawned various containers and tried inserting items into them and checked to see that this was working properly.

I tried inserting containers into each other with and without items, that seemed to work as expected as well.

I inserted plates into a ballistic vest, cross-references the displayed weight/volume values based on what was present in the .json files for the vest's pockets and that also worked as expected.

I inserted plates and other things into the MOLLE rucksack (which has a few different kinds of pockets), and that also worked as expected

Let me know if there are any edge cases you think I should test.

#### Additional context

cataclysm-tiles_gZctpoBLsw

### DIFF
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -1585,11 +1585,11 @@ drop_locations game_menus::inv::holster( avatar &you, const item_location &holst
 
     inventory_holster_preset holster_preset( holster, &get_avatar() );
 
-    inventory_drop_selector insert_menu( you, holster_preset, _( "ITEMS TO INSERT" ),
-                                         /*warn_liquid=*/false );
+    inventory_insert_selector insert_menu( you, holster_preset, _( "ITEMS TO INSERT" ),
+                                           /*warn_liquid=*/false );
     insert_menu.add_character_items( you );
     insert_menu.add_nearby_items( 1 );
-    insert_menu.set_display_stats( false );
+    insert_menu.set_display_stats( true );
 
     insert_menu.set_title( title );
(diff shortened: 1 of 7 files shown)


mqrause:
Inventory selector menus already have ways to display these stats dynamically. Check out #61974 for something similar I've done.

strategineer:
Ayyyyy, will do!

strategineer:
That did the trick! This still needs some work and more testing though.

I'm running into problems with containers that have pockets. For example, with a ballistic vest, it displays the max weight allowed as 0 but the current weight as more than zero (~7kg with 4 plates inserted) .... But that's a problem for another day.

PR #64223, opened 2023-03-13
Fix category spawn probability

Kromgart:
#### Summary
Bugfixes "Item category spawn roll"

#### Purpose of change

Item categories have a spawn_rate parameter, which specifies the probability to spawn items from that category. It is reasonable to assume that setting that value for category GUNS to 0.2 should make guns spawn 5 times rarer than usual (when the setting is used in the code of course).
 
But when category has a spawn_rate less than 1, the code was comparing that value to a roll of rng_float(0.1, 1.0) which is a uniform_real_distribution. 
This causes problems for small spawn_rate parameters. In the above example (0.2 spawn_rate for guns) the check **rng_float(0.1, 1.0) <= 0.2** would succeed in 0.1 / 0.9 ~ 11% cases (almost 2 times rarer than we have specified). Setting category FOOD to 0.15 would cause it to spawn in 0.05 / 0.9 ~ 5.6% cases (almost 3 times rarer).

#### Describe the solution

Made the roll from 0.0 to 1.0 to even out the actual spawn probability. Made spawn_rate of 0 a special check (never spawn).

#### Testing

I've been playing with this change for a week, using low spawn rates for some categories. Do not see any issues so far.


Night-Pryanik:
All this math stuff was never my strong side.

PR #47576, opened 2021-02-17
Tiring dodges

kevingranade:
#### Summary
None

#### Purpose of change
I noticed dodges are inexplicably "free" from a stamina point of view, so decided to add a stamina cost.
While I was looking at the code, I also noticed that the special attack code inexplicably doubles the skill cap for dodge training.

#### Describe the solution
Removed skill cap doubling for dodge training.
Added a small stamina cost when attempting to dodge.

#### Describe alternatives you've considered
I still need to fine tune the stamina cost, and add a weariness level increase. (DONE)

#### Testing
Spawn a monster with a frequently firing special attack, set dodge skill high, allow it to attack you repeatedly, and see if skill continues training.
Spawn a monster that attacks, wait near it to allow it to attack you, and check that stamina drops slowly.

PR #64224, opened 2023-03-13
Tiring dodges

Fris0uman:
#### Summary
None 

#### Purpose of change
Continuation of #47576

#### Describe the solution

- [x] Check for potential damage before deciding to dodge
- [x] Check for probability of dodge before trying to dodge
- [x] Move check_dodge out of monattack and  into Creature
- [x] Unify all the scattered `get_dodge` to use `dodge_check` every time an actual dodge is supposed to take place
- [x] Make use of `dodge_roll()`
- [x] Move ` dodges_left--` from `on_dodge` to `on_try_dodge`
- [x] Make sure `dodge_check` is called where needed
- [x] doudble check there's no free dodge left-overs
- [x] Make sure tests are still working
- [x] Projectile attacks

Also consume dodge_attempt inside `on_dodge_try` instead of on succesfull dodge which will probably reduce effective chance to dodge since you'll only get one attempt per turn instead of one succes per turn

#### Testing

- Get naked
- Get high dodge
- Get invincible
- Spawn a slime
- Wait and dodge
- See stamina go down
- When stamina get around two bars, the dodges don't happen anymore and the stamina doesn't get lower

Check that dodging projectile cares about stamina
image

Check that you get a message about not trying to doge special attacks
image

#### Additional context

We still need to unify dodges, this makes two parallele way to doge:
- Special attacks are dodged via `dodge_check` which only tries to dodge if we have 80% or more chance of succes (should make it 50%?) , and only if the attack would do non 0 damage. Alternatively if the attacks were to do more than half the current hp of the target part in damage it tries to doge no matter the chances
- Regular attack go through deal_melee_attack which always calls `on_try_dodge`

I didn't manage to unify everything because deal_melee_attack sometimes use flat hit roll value, and sometimes random value from dice rolls and sometimes random values from rng_normal and I have no idea how to unify those

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1415,7 +1415,12 @@ bool Creature::dodge_check( monster *z, bodypart_id bp, damage_instance dam_inst
         potential_damage += dmg.amount - guy_target->worn.damage_resist( dmg.type, bp );
 
     }
-    float dmg_ratio = potential_damage / guy_target->get_part_hp_cur( bp );
+    int part_hp = guy_target->get_part_hp_cur( bp );
+    float dmg_ratio = 0.0f; // if the part is already destroyed it can't take more damage
+    if( part_hp > 0 ) {
+        dmg_ratio = potential_damage / part_hp;
+    }
+
 
     //If attack might remove more than half of the part's hp, dodge no matter the odds
     if( dmg_ratio >= 0.5f ) {
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -723,14 +723,15 @@ float Creature::get_crit_factor( const bodypart_id &bp ) const
 
 int Creature::deal_melee_attack( Creature *source, int hitroll )
 {
+    const float dodge = dodge_roll();
+    int hit_chance = hitroll - size_melee_penalty();
+    int hit_spread = hit_chance - dodge;
+
     add_msg_debug( debugmode::DF_CREATURE, "Base hitroll %d",
                    hitroll );
-
-    float dodge = dodge_roll();
     add_msg_debug( debugmode::DF_CREATURE, "Dodge roll %.1f",
                    dodge );
 
-    int hit_spread = hitroll - dodge - size_melee_penalty();
     if( has_flag( MF_IMMOBILE ) ) {
         // Under normal circumstances, even a clumsy person would
         // not miss a turret.  It should, however, be possible to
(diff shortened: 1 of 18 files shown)


anoobindisguise:
So it doesn't look like it currently does,  but will dodging stamina use decrease with the dodging skill?

src/creature.cpp
```diff
@@ -1309,6 +1313,57 @@ void Creature::longpull( const std::string &name, const tripoint &p )
...
+    }
+
+    Character *guy_target = as_character();
+
+    float potential_damage = 0.0f;
+    for( damage_unit dmg : dam_inst.damage_units ) {
+        potential_damage += dmg.amount - guy_target->worn.damage_resist( dmg.type, bp );
+
+    }
+    float dmg_ratio = potential_damage / guy_target->get_part_hp_cur( bp );
```
fungamer2-2, line 1355: Wouldn't this lead to a division by zero when the limb is broken (0 HP)?

kholat:
The players stamina pool is a resource used for doing damage, movement, and in the case of dodge also damage mitigation.
What if a player thinks/decides/figures out that he would rather use stamina for attacking more times instead of dealing damage? On the other hand, what if he is anticipating a special attack at <2 stamina bars and wants to attempt to dodge it in the near future?

To say it shortly, shouldn't then there be a way for a player to decide if/when he wants to dodge or not, kind of like enabling skills to train to focus your 'focus' in a specific skill?

If the answer to the question is trivial, for instance "this is too metagamey, thats another button to press to enable/disable dodging now" or something similar, this leads to another question, how/why are the seemingly arbitrary number limits decided here and that loops back to the kind of "why cant the player manipulate these limits in some way?". For instance why never dodge if hit prob is >80%? What if theres a big attack that might come next soon and the player WANTS to attempt to dodge? Why always dodge if limb would lose >50% hp? What if that attack is 49% hp? An interesting case right off the bat in my eyes is some heavily armored bird mutant player, that will basically take zilch damage due to armor, but in theory pretty much any attack, if hit naked would do tons of damage. Some quick things to think about IMO at least.

Fris0uman:
quick note on the arbtrary number: they are only involved in special attacks, so they'll barely ever actually matter in game. Better stamina management code for dodge is welcome, it should manage itself as much as possible though, having to go through menus every time to switch the numbers would be a huge pain.

alp7292:
cant gain xp

PR #64234, opened 2023-03-13
Set pocket management menu message according to set keybindings

Qrox:
#### Summary
None

#### Purpose of change
Generate the keybinding message according to the currently set hotkeys, reduce code duplication by moving the code into a function, and other code cleanup.

#### Testing
Tested in game and the keybinding message was shown correctly when changing the pocket settings for a single item or the full inventory. Other parts of the UI also worked as expected.

#### Additional context
Currently keybinding messages of uilists are not updated dynamically so this one is not updated either. Perhaps we can change the message members of uilist into callbacks that take the input context and output the dynamic message but that is out of the scope of this PR.

### DIFF
--- a/src/item_contents.cpp
+++ b/src/item_contents.cpp
@@ -2441,7 +2441,7 @@ void item_contents::favorite_settings_menu( item *i )
     pocket_management_menu( remove_color_tags( i->display_name() ), to_organize );
 }
 
-void pocket_management_menu( const std::string title, const std::vector<item *> &to_organize )
+void pocket_management_menu( const std::string &title, const std::vector<item *> &to_organize )
 {
     static const std::string input_category = "INVENTORY";
     input_context ctxt( input_category );
--- a/src/item_contents.cpp
+++ b/src/item_contents.cpp
@@ -36,6 +36,34 @@
 
 static const flag_id json_flag_CASING( "CASING" );
 
+class pocket_favorite_callback : public uilist_callback
+{
+    private:
+        std::vector<std::tuple<item_pocket *, int, uilist_entry *>> saved_pockets;
+        // whitelist or blacklist, for interactions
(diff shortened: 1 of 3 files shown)


src/item_contents.cpp
```diff
@@ -2413,13 +2436,22 @@ void item_contents::info( std::vector<iteminfo> &info, const iteminfo_query *par
...
     std::vector<item *> to_organize;
-    uilist pocket_selector;
     to_organize.push_back( i );
-    pocket_favorite_callback cb( to_organize, pocket_selector );
-    pocket_selector.title = remove_color_tags( i->display_name() );
-    pocket_selector.text = cb.title;
+    pocket_management_menu( remove_color_tags( i->display_name() ), to_organize );
+}
+
+void pocket_management_menu( const std::string title, const std::vector<item *> &to_organize )
```
Qrox, line 2444:
```suggestion
void pocket_management_menu( const std::string &title, const std::vector<item *> &to_organize )
```

src/item_contents.h
```diff
@@ -375,32 +374,6 @@ class item_contents
...
-    public:
-        explicit pocket_favorite_callback( const std::vector<item *> &to_organize,
-                                           uilist &pocket_selector );
-        void refresh( uilist *menu ) override;
-        bool key( const input_context &, const input_event &event, int entnum, uilist *menu ) override;
-
-        const std::string title =
-            _( "Modify pocket settings and move items between pockets.  [<color_yellow>RETURN</color>] Context menu\n" );
-};
+void pocket_management_menu( const std::string title, const std::vector<item *> &to_organize );
```
Qrox, line 377:
```suggestion
void pocket_management_menu( const std::string &title, const std::vector<item *> &to_organize );
```

PR #64235, opened 2023-03-13
fix #64207

### LINKED ISSUE #64207: Faraday Chainmail Suit - Uncomfortable.
### Describe the bug

Was wanting to re-use the old issue but it is closed so a new one it is.

https://github.com/CleverRaven/Cataclysm-DDA/issues/61910

Despite the changes in https://github.com/CleverRaven/Cataclysm-DDA/pull/61066 , there still are issues in regards to this armour. Namely the fact it is uncomfortable in literally every part of your body. This is a bit weird because there are many body parts that do not have any "softer" alternatives to wear underneath this, such as *the eyes*. 
And getting -24% speed and constant (and increasing) pain per **each different limb** is a bit silly. 

### Attach save file

n/a - easy to reproduce

### Steps to reproduce

1. make a new world, create a new character;
2. spawn a faraday chainmail suit and wear it.
3. Notice it is "uncomfortable" (refer to previous issue's Screenshot)
4. open the @ menu and check effects. Notice the number of body parts affected.
5. scramble to find soft things to wear underneath the Suit.
6. Fail to find anything for the eyes because the crafting menu's search function does not allow you to search for things that aren't "head", "torso", "arms", "legs", "feet", "hands." and one does not have the patience to scroll through hundreds of items to figure out there aren't any soft-things that protect the eyes. 
6.1 Realize afterwards that for the other body parts, one has to choose between mostly high-warmth soft gear to wear underneath which makes no sense.

### Expected behavior

I would expect there would either be more low-warmth items *or* we would have the opportunity to use the tailoring kit to pad hard items with soft materials, at the expense of minimal warmth increase and encumbrance, as well as weight and volume.

### Screenshots

_No response_

- windows
- no relevant mods
- ce1081a

### Additional context

I was wanting to re use the older issue but I thought opening a new one would be better. I apologise if it isn't the proper course of action.

Kamejeir:
#### Summary
None

#### Purpose of change
fixes #64207, ~~in the way provided by @anoobindisguise, then agreed upon by the author~~ well yes, but actually no. it's a functionally identical fix with less materials.json bloat, provided by Drew4484

#### Describe the solution
~~creates a copy of qt_steel_chain, except it's comfortable, and applied that to the faraday chainmail suit's eyes and mouth.~~
gives the eyes and mouth of faraday chainmail suit an infinitesimally thin layer of cotton (not actually reflected in the recipe, but I hope we can pretend it doesn't exist because I try to reflect anoobindisguise saying it's a cage), and adjusted breathability to match

some convoluted way of telling certain clothing layers to act padded, rather than the whole outfit

#### Testing
nakey. faraday suit. uncomfortable, but not for eyes or mouth. the new stats show up in-game
image

#### Additional context
the size variants of the faraday chainmail suit has "copy-from" so I am not making a direct edit there

picture of one Anouk Wipprecht for reference of the new visualization of face cover (focus on the cage part). also removing the mouth encumbrance completely + lowering eye encumbrance because of this, as well as setting the ranged coverage to 10% on behalf of anoobindisguise. I also cite for the change from qt_steel to qt_steel_chain, as outlined by a comment in the issue
image
even after Drew's fix, it being a modelling of Anouk's work does not change. pretend it's a cage. 
yes, I'll try to keep up to date when future PRs allow us to get rid of this janky fix and set the two layers to comfortable.

### DIFF
--- a/data/json/items/armor/suits_protection.json
+++ b/data/json/items/armor/suits_protection.json
@@ -1981,18 +1981,20 @@
         "encumbrance": 30
       },
       {
-        "material": [ { "type": "pseudo_soft_qt_steel_chain", "covered_by_mat": 100, "thickness": 1.2 } ],
+        "material": [ { "type": "qt_steel_chain", "covered_by_mat": 100, "thickness": 1.2 }, { "type": "cotton", "covered_by_mat": 100, "thickness": 0.001 } ],
         "covers": [ "eyes" ],
         "coverage": 95,
         "cover_ranged": 10,
-        "encumbrance": 2
+        "encumbrance": 2,
+        "breathability": "GOOD"
       },
       {
-        "material": [ { "type": "pseudo_soft_qt_steel_chain", "covered_by_mat": 100, "thickness": 1.2 } ],
+        "material": [ { "type": "qt_steel_chain", "covered_by_mat": 100, "thickness": 1.2 }, { "type": "cotton", "covered_by_mat": 100, "thickness": 0.001 } ],
         "covers": [ "mouth" ],
         "coverage": 95,
         "cover_ranged": 10,
-        "encumbrance": 0
+        "encumbrance": 0,
+        "breathability": "GOOD"
       },
       {
         "covers": [ "foot_l", "foot_r" ],
--- a/data/json/items/armor/suits_protection.json
+++ b/data/json/items/armor/suits_protection.json
@@ -1981,7 +1981,10 @@
         "encumbrance": 30
       },
       {
-        "material": [ { "type": "qt_steel_chain", "covered_by_mat": 100, "thickness": 1.2 }, { "type": "cotton", "covered_by_mat": 100, "thickness": 0.001 } ],
+        "material": [
+          { "type": "qt_steel_chain", "covered_by_mat": 100, "thickness": 1.2 },
+          { "type": "cotton", "covered_by_mat": 100, "thickness": 0.001 }
+        ],
         "covers": [ "eyes" ],
         "coverage": 95,
         "cover_ranged": 10,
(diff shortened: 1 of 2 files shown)


Drew4484:
Why not just change the chainmail to be outer layer, like our other chainmails and to match the description?

Drew4484:
> > Why not just change the chainmail to be outer layer, like our other chainmails and to match the description?
> 
> Huh. I'll go test if it works.

When I updated the other chainmails I left the faraday and sharksuit alone because I didn't want to make variants for each steel grade for two more complete suits. Adding the "OUTER" flag would be sufficient for this quick change.

RAldrich:
The conflict in resolution is how one imagines the head-gear of the Faraday suit. In the issue where this solution was proposed, it's put forward as a "cage" that's not in direct contact with your face/eyes/mouth, and so should not be uncomfortable. If you envision it as more of a chain-mail veil, that would definitely be uncomfortable on your face, and could be mitigated with no changes by wearing (e.g.) safety glasses and a balaklava.

So - what does the Faraday suit item description say? If you want to press on with this change I would make sure that description matches. But I'm also fine with leaving this as-is (and maybe making the item description clear that it's an uncomfortable chain veil).

anoobindisguise:
Yeah I imagined it as something similar to Anouk Wipprecht's faraday chainmail for the face. So another thing that might be done is specifically giving it really poor ranged coverage now that I think about it.

Drew4484:
> Yeah I imagined it as something similar to Anouk Wipprecht's faraday chainmail for the face. So another thing that might be done is specifically giving it really poor ranged coverage now that I think about it.

If this is the route you're going, please just use flags to make the equipment not uncomfortable instead of adding a new material to maintain.

Drew4484:
> and making the entire suit comfortable is not desired citing anoob in the issue: "So chainmail is really uncomfortable to wear on your skin,"

So make a separate helmet/mask piece. These don't spawn anywhere, players will be making them anyway. A separate component is more desirable than an entirely new material that will be used for two items (plus XL/XS variants).

Raikiri:
I guess there's currently an issue with multi-piece suits giving electrical immunity. As in, there's currently no way to grant a certain amount of electrical resistance per armor piece and immunity if you have all of them. This leads to a whole range of problems associated with electric immunity items that are only allowed to be fullbody suits.

Drew4484:
> I guess there's currently an issue with multi-piece suits giving electrical immunity. As in, there's currently no way to grant a certain amount of electrical resistance per armor piece and immunity if you have all of them. This leads to a whole range of problems associated with electric immunity items that are only allowed to be fullbody suits.

This has to do with electrical ranged attacks like a shocker zombie or environmental hazard. For those you need 100% coverage. For melee against a shocker zombie, you only need a non-conductive weapon or glove to prevent zapback.

Raikiri:
Yes, but it's shock clouds that kill players most of the time, not melee attacks.

PR #64236, opened 2023-03-13
Update furniture-sleep.json

### LINKED ISSUE #64227: Straw Beds Give Just 5% of its Straw on Deconstruction

I started making my base in a Junkyard, which happens to have some straw beds in the lot, and wanted to move one inside. However, deconstructing a straw bed gave me just 8 piles of straw, out of 160 needed to create a new one. This doesn't make any sense - how would straw be lost in this process? It's practically amorphous - it can't get damaged or lost, it'd be like losing water from pouring it out of a bottle and into a bucket. I can't break water - maybe a small amount can evaporate, but straw doesn't evaporate, so it doesn't have that excuse.

### Solution you would like.

Deconstructing a straw bed should realistically give at least 80-90% of its resources, if indeed any are lost at all.

N/A

### Additional context

N/A

dgreenlee1:
#### Summary
Bugfixes "Straw Beds give more straw piles when deconstructed, up to 80%-90% of required instead of just 5%"

#### Purpose of change
This change Fixes https://github.com/CleverRaven/Cataclysm-DDA/issues/64227, a pile of straw is an amorphous mass that doesn't spill or become destroyed by forming it into different shapes such as a bedroll. When a straw bed is deconstructed, it should give back just about as much straw as was needed to make it. And it also makes more sense if carefully deconstructing the bed returns more materials than bashing it apart.

#### Describe the solution
I made it give more straw instead of less in the .json file in the data folder.

I also considered adjusting the amount of straw returned by bashing the bed apart, but that would be outside the scope of this change.

#### Testing
Making a straw bed and deconstructing it returns materials.

#### Additional context
N/A

### DIFF
--- a/data/json/furniture_and_terrain/furniture-sleep.json
+++ b/data/json/furniture_and_terrain/furniture-sleep.json
@@ -259,7 +259,7 @@
     "comfort": 2,
     "floor_bedding_warmth": 200,
     "required_str": -1,
-    "deconstruct": { "items": [ { "item": "stick", "count": 4 }, { "item": "straw_pile", "count": [ 7, 8 ] } ] },
+    "deconstruct": { "items": [ { "item": "stick", "count": 4 }, { "item": "straw_pile", "count": [ 120, 140 ] } ] },
     "flags": [ "TRANSPARENT", "FLAMMABLE_ASH", "ORGANIC", "MOUNTABLE", "SHORT", "EASY_DECONSTRUCT" ],
     "bash": {
       "str_min": 6,


PatrikLundell:
Straw does "spill" in the form of individual straws getting lost, and gets destroyed by straws breaking into smaller pieces with use. Thus, getting most of it, but not all, back seems reasonable.

SolventMercury:
> Straw does "spill" in the form of individual straws getting lost, and gets destroyed by straws breaking into smaller pieces with use. Thus, getting most of it, but not all, back seems reasonable.

The current changes appear to already incorporate this - giving a variable amount of straw rather than the exact amount that went in.

PatrikLundell:
Yes. My comment was on the text part of the PR which sort of says that there shouldn't be any losses. The second sentence was intended to state that I did agree with the implementation which does allow for some losses.

Thus, I think it's simply a minor matter of miscommunication.

flybye3o6:
also make sure you make new a "branch" under your "dgreenlee1:master" branch, so you can work on multiple projects at the same time

Issue #59708, opened 2022-07-28
Allow the player and companions to slaughter tame animals


PatrikLundell:
### Is your feature request related to a problem? Please describe.

Currently the way to "slaughter" animals is to attack them. However, that can result in the corpse getting sufficiently damaged that you get a reduced product yield out of it. This is particularly noticeable with birds (such as chicken), as they yield little to begin with. Loss of chucks of meat for scraps of meat is bad because scraps have far fewer uses than chunks do.

### Solution you would like.

Introduce a tame animal interaction of "slaughter" that kills the animal and leaves an undamaged corpse. It may require a suitable knife for larger animals (wringing the neck of birds presumably is done without any tools).

For companions you'd probably want a combined "slaughter and butcher" activity where you designate one or more targets (several may be appropriate for smaller animals such as birds). After that, the companion would kill and process one animal at a time until they've all been processed. Ideally, the activity should allow the designation of target locations (zones?) for the products into which they'd be stored one animal at a time. Further, an option to collect blood would be useful (although that requires that the issue of selecting containers for it is supported).

### Describe alternatives you have considered.

Ideally I'd complement the companion activity with a faction camp order where you'd order the same activity to be performed, but the companion goes off into the void with the animals and returns with the products when recalled from the board. This wouldn't require any storage designation as all the products would be delivered at once, but it does have the blood collection option issue.
The advantage of this version is that it wouldn't require the player to stay at the base while the potentially several hours long sequence of actions is performed.

### Additional context

Tangentially related to #59643, as this is a way to make slaughter and butchery less of a hassle for the player, to offset that suggestion's increase in effort to maintain the animals.

Zireael07:
Hm, slaughterhouse camp building coming when? (I imagine slaughtering things like horses or cows would require them being tied down or placed in such a building)

PatrikLundell:
In the real world slaughter houses are currently used for industrial scale slaughtering, but I don't think people who slaughter animals for their own use have/had dedicated buildings on their farms for it. In the richer parts of the world there are health (and [anti] animal cruelty) authorities that mandate slaughter conditions for commercial and sometimes private use (don't think it's legal where I live, with hunting being an obvious exception), but those authorities are somewhat impotent in the post apocalyptic world...

[...]

PR #64240, opened 2023-03-13
Cull pets

ekaratzas:
#### Summary
Features "Allow player to easily and cleanly kill friendly farm animals"

#### Purpose of change
This PR partially addresses #59708 by allowing the player to easily and cleanly slaughter friendly farm animals. Right now the only solution is to outright attack friendly farm animals in order to harvest them which is very clunky and often results in damaged corpses, even if you use something like a crossbow with small game bolts on friendly chickens. The PR addresses this issue for the player character but does not include a system for NPCs to do this work on behalf of the player.

#### Describe the solution
The feature works by introducing a new flag, CAN_BE_CULLED, that can be applied to monsters. When interacting with a friendly monster via 'e', it introduces a new menu option if this flag is present for that type of monster and the player now has the option to 'cull' their friendly monster which results in a dead pet with an undamaged corpse. The only requirements are that the player has a cutting tool on them when interacting with the friendly monster. This flag is currently only applied to common farm animals that can currently be tamed like chickens, cows, sheep and so on via the respective json files.

I've thought about some of the suggestions mentioned in the respective issue like not requiring a knife for culling chickens or requiring a bigger knife for bigger animals like cows but those ideas seemed a bit extreme to me, somewhat over-engineering the approach. This is a simple QoL improvement that a lot of farmer player will appreciate. At the moment I'm not sure how introducing an activity for NPCs to do this work for the player would work in the codebase so I'm introducing a simple fix for the player in the meantime as I'm tired of chasing fleeing chickens with my crossbow.

#### Testing

1. In debug mode spawned chicken/cow next to player.
2. Notice I can't interact with them before they are tamed.
3. Once they are tamed and I interact with them, the culling menu option shows up.
4. Culling the animal fails with the appropriate error message if the player does not have a tool with cutting quality on them.
5. Culling the animal works as expected if the player has a cutting tool. The process leaves behind a cleanly killed (undamaged) corpse.
6. Smoked chicken tastes nice.

### DIFF
--- a/data/json/monsters/bird.json
+++ b/data/json/monsters/bird.json
@@ -35,7 +35,7 @@
       "feed": "The %s seems to like you!  It runs around your legs and seems friendly.",
       "pet": "The %s runs around your leg."
     },
-    "flags": [ "SEES", "HEARS", "SMELLS", "ANIMAL", "PATH_AVOID_DANGER_1", "WARM", "CANPLAY", "SWARMS" ]
+    "flags": [ "SEES", "HEARS", "SMELLS", "ANIMAL", "PATH_AVOID_DANGER_1", "WARM", "CANPLAY", "SWARMS", "CAN_BE_CULLED" ]
   },
   {
     "id": "mon_grouse",
@@ -411,7 +411,7 @@
       "feed": "The %s seems to like you!  It runs around your legs and seems friendly.",
       "pet": "The %s runs around your leg."
     },
-    "flags": [ "SEES", "HEARS", "SMELLS", "ANIMAL", "PATH_AVOID_DANGER_1", "WARM", "CANPLAY", "SWARMS" ]
+    "flags": [ "SEES", "HEARS", "SMELLS", "ANIMAL", "PATH_AVOID_DANGER_1", "WARM", "CANPLAY", "SWARMS", "CAN_BE_CULLED" ]
   },
   {
     "id": "mon_grouse_chick",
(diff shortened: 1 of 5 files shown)


TheShadowFerret:
Does this bypass morale penalties for killing cute animals?

~~I wonder this for absolutely no future PR-related reasons 🙀 🐶~~

ekaratzas:
> Does this bypass morale penalties for killing cute animals?
> 

To my understanding, yes it bypasses those.

NetSysFire:
This might be harder to implement, but I would potentially not offer the player to cull e.g a chick as they are too small and literally only yields a single scrap of meat. Although this is probably desired in an emergency, I would maybe add a hint to the message somehow, e.g "This creature is immature and should probably be allowed to mature for best butchering results".

MoonlitRiverlight:
Irl cows are culled with a hammer (small sledgehammer) blow to the forehead. Pigs are culled with a normal knife by stabbing the heart once. Chicken can be culled by breaking the neck, yes, and it will, arguably, be more humane than decapitation. All of those rely on killing as fast as possible to diminish stress as it alters the meat, and not cause the animal to struggle. Decapitation has a tendency to leave the heads alive for a few seconds if they don't die of the shock, which can be stressful to the culler to watch.
In general a wound to the brain is more "humane" than to the heart, but the difference isn't big enough to prove a point. Either would work in our scenario, so it might be fine to have culling require either hammering 2 or cutting. If you were to implement a failure state where an unproficient culler would stab off-mark and cause the animal to bleed and flee or aggro til it dies, hammering might be the one option with lower failure rate so long as you have enough strength to use a hammer.

NetSysFire:
I would also require cutting 2 because cutting 1 usually means a sharp rock or a weapon that can cut but is not really ideal (e.g a glaive). It would be possible to kill a chick (or maybe even chicken as it still might be small enough) with hammering 1, which is basically a rock, but doing the same to a cow is cruel and inefficient. I could imagine an optional property in a monsters' json like this:

```json
  "cull": { "required_quality": "CUT", "value": 2 }
```

Now we have to decide how in-depth we want to go. Sure, you can kill a chick with both cutting 2 and hammering 1 and have that dict in an array for that purpose but in the end it is a rather niche function.

PR #64251, opened 2023-03-14
Obsolete plastic mold

Aerin-of-the-Toast:
#### Summary
None

#### Purpose of change

The "plastic mold" that was largely replaced by proper vacuum molders was still used in a couple of recipes.

#### Describe the solution

Move the plastic mold to the obsoletion file, remove the recipe for it entirely. Switch angle grinder to use a vacuum molder instead, remove mold and replace with needing boiling 1 for plastic chunk recipe (since they're irregularly-shaped chunks made of melted plastic forks you don't need precision) and the two gummy candies (which already required boiling 1 anyway).

None

#### Testing

Game runs with change and recipes updated

#### Additional context
None

### DIFF
--- a/data/json/items/tool/workshop.json
+++ b/data/json/items/tool/workshop.json
@@ -722,23 +722,6 @@
     "flags": [ "ALLOWS_REMOTE_USE" ]
   },
   {
-    "id": "mold_plastic",
-    "type": "TOOL",
-    "name": { "str": "plastic mold" },
-    "description": "A mold that can be used to form plastic chunks.",
-    "weight": "320 g",
-    "volume": "2500 ml",
-    "longest_side": "36 cm",
-    "price": 1000,
-    "price_postapoc": 50,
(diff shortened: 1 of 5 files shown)


Zireael07:
Changes view sucks. Which recipes still had it? The mold, as I envisioned it when I added it, was supposed to be used for plastic bottles/cups, that kind of simple items. Later people piggybacked on it, going as far as to making HULLS out of it... O_o

Aerin-of-the-Toast:
> Changes view sucks. Which recipes still had it? The mold, as I envisioned it when I added it, was supposed to be used for plastic bottles/cups, that kind of simple items. Later people piggybacked on it, going as far as to making HULLS out of it... O_o

Guess it got out of hand didn't it? At least now we have a proper vacuum molder :P The recipes that still needed the mold were the plastic chunk, angle grinder, and two gummy candies.

Issue #64221, opened 2023-03-13
Improving No NPC Needs


Maleclypse:
### Describe the bug

There are a number of things that cause NPCs to experience needs even with NPC Needs turned off but our tools for handling these types of things have improved since that mod was created.

### Attach save file

N/A

### Steps to reproduce

Gather NPC followers,  let them experience thirst and hunger but be unable to force them to eat and drink without taking over the NPCs directly.

### Expected behavior

The stated needs that are modded out to not be regularly impacting my NPCs

We have Effects on Condition now that could be created to reset hunger, thirst and calories to baseline every day at dawn.  This wouldn't even be a very complicated EOC probably.

### Versions and configuration

N/A

anoobindisguise:
This would be a good time to mention that turning off No NPC Needs makes it impossible for NPCs to lose instability.

Maleclypse:
> This would be a good time to mention that turning ON No NPC Needs makes it impossible for NPCs to lose instability.

Does that mean NPCs don't regenerate blood after losing it and will eventually die of hypovolemia if the take enough wounds even if the heal from them? Or are NPCs immune to blood loss other than hp damage?

PR #64274, opened 2023-03-15
[WIP] No NPC Needs EOC-ified

DPavonis:
#### Summary
Mods "Added EOC-based alternative to No NPC Food"

#### Purpose of change

Per #64221 I threw this together last night 

#### Describe the solution

Use a recurring EOC to globally set NPC needs to baseline

Letting someone else do it? Do nothing?

#### Additional context

Things TODO/I would like input on:

- Decide on how to filter out the player
- ~~Look into resetting hunger/stomach contents, or if that even matters~~ So far this doesn't really seem to matter
- Decide on appropriate time between resetting needs
- Do we want this to cover vitamins as well?


GuardianDll:
How does it works with camp? Do NPC require food in this case, or it is also effectively turned off?

DPavonis:
> How does it works with camp? Do NPC require food in this case, or it is also effectively turned off?

Some quick and probably not quite as thorough as it should be testing on a built-up camp in an in-progress game shows they mostly shouldn't. There was some initial eating after loading up the save, but once the EOC started affecting them, they stopped eating.

data/mods/No_NPC_Needs/effects_on_condition.json
```diff
@@ -0,0 +1,15 @@
+[
+  {
+    "type": "effect_on_condition",
+    "id": "reset_needs",
+    "recurrence": "4 hours",
```
Qrox, line 5: So this means during the 4 hour interval the NPCs' needs will be calculated as usual right? If so this might not work in some extreme conditions such as overheating where you can become thirsty pretty quick I think.

DPavonis: Correct, and good call on extreme conditions like that

DPavonis: Testing shows that NPCs don't actually care about temperature, and thus don't sweat when it's 50 C out ¯\_(ツ)_/¯

Qrox: Can EOCs be triggered when the needs change, so that you can reset hunger, thirst etc once they fall below a certain value?

Maleclypse:
> How does it works with camp? Do NPC require food in this case, or it is also effectively turned off?

So the calories NPCs take from the camp stores currently either just make them obese and they never lose weight or don't actually impact them at all.  I'm never sure because all my camp NPCs are obese and constantly scream about hunger randomly after a few months.

data/mods/No_NPC_Needs/effects_on_condition.json
```diff
@@ -0,0 +1,15 @@
...
+    "global": true,
+    "run_for_npcs": true,
+    "effect": [
+      { "arithmetic": [ { "u_val": "stored_kcal_percentage" }, "=", { "const": 25 } ] },
+      { "arithmetic": [ { "u_val": "thirst" }, "=", { "const": 0 } ] },
+      { "arithmetic": [ { "u_val": "fatigue" }, "=", { "const": 0 } ] },
+      { "arithmetic": [ { "u_val": "sleep_deprivation" }, "=", { "const": 0 } ] }
```
Maleclypse, line 12: I'd remove this line because currently NPCs do a great job of finding a bed and sleeping when they need to. It's just eating and drinking that are problems.
```suggestion
      { "arithmetic": [ { "u_val": "sleep_deprivation" }, "=", { "const": 0 } ] }
```

Maleclypse:
@Ramza13 can you confirm that this will only effect NPCs or if there is a way to make it only effect NPCs?

data/mods/No_NPC_Needs/effects_on_condition.json
DPavonis: Yeah, that was just me leaving it in there when I ~~copied~~ borrowed from the portal resurrection EOC for the purposes of testing because I was too lazy to use the debug mutation

Maleclypse: No shame in being a cargo cultist.  I'm a priest of the cargo cult. :)

DPavonis: Praise the cargo cult

DPavonis:
> @Ramza13 can you confirm that this will only effect NPCs or if there is a way to make it only effect NPCs?

Well if there wasn't before, there is now

PR #62662, opened 2022-12-10
Revisit Fire Drills (Refills)

oosyrag:
#### Summary
Balance "Removed fire drill refill recipes, added notched fire boards as ammo"

#### Purpose of change

Apparently a recurring source of confusion for new players is "Why does a fire drill need a fire drill to craft?", not understanding that it is a refill. Fixes #59093 by removing the refill recipes and adding new fire boards recipes (notched plank, notched stick) as ammo for fire drills. They function as disposable batteries.

Also plan to resolve #34063, increasing the base move time of starting a fire drill by a lot. Seems like code is available to account for a "long action".

Fixes #61076, fire drills no longer have BURNOUT flag.

#### Describe the solution

I think the confusion stems from being having access to the reloading recipes at survival 0, without having or seeing the base recipe yet, which is auto learned at survival 2. ~~Various solutions below, went with option "a" as the least invasive change, as fire drills are not found in the world, so having access to the reloading recipe before being able to make it doesn't matter anyway.~~

Ideally the ability to change the name of a recipe on the crafting screen would be nice, such as "Refill fire drill" that results in a new fire drill. However, the crafting screen has to use the resulting item name. The workaround could be "c" below, with a new item named "Pump/Bow Fire Drill Reloaded". But I struck that out as adding item bloat for a very niche item(s), and not respecting the current string freeze, and basically overkill.

Edit: Moving forward with option "e". Fire drills will be able to be (re)loaded with new notched plank and notched stick magazines, with notches as ammo.

- [x] Remove fire drill reload recipes
- [x] Remove plank requirement from fire drill recipes
- [x] Add recipes for notched stick and notched plank
- [x] Add notch ammo
- [x] Allow fire drills to take notched stick and plank as ammo
- [x] Add notched stick or plank to starting inventory for Innawoods
- [x] Test
- [x] Balance pass for weight/volume/length and other values
- [x] Lint
~~- [ ] Figure out how to display in inventory loaded fire drill and fire board ammo as "charges" (or at least "notches") instead of  blank and "notch" respectively~~

This seemingly simple change is a quite a bit more involved than I thought it would be. This is a largest pr I've done so far please let me know if I missed something or did something wrong or there are other things to consider that I didn't think of.

Balance:
Notched plank to 24 charges, notched stick 10 charges.
Bow fire drill base moves adjusted to 30000 (5 minutes), moves_slow to 60000 (10 minutes)
Pump fire drill base moves to 6000 (1 minutes) and moves_slow 12000 (2 minutes)
[...]

PR #64277, opened 2023-03-15
Makes tinder easier to craft

Light-Wave:
#### Summary
Balance "Makes tinder easier to craft"

#### Purpose of change
#62662 made it so that primitive fire drills requires tinder. By default, tinder requires a tool with cutting 2. In Innawood, you need to create a fire as early as possible, preferably before you spend time making a good cutting tool.
Besides, you probably don't need a good cutting tool to make tinder out of a newspaper.

#### Describe the solution
Split the tinder recipe into 3 parts depending on how good a knife you would need to create it.

The following items can be made into tinder without a knife: pine_bough, cotton_patchwork, paper, cardboard, rolling_paper, wrapper, straw_pile
The following items require cutting 1: splinter, withered, cattail_stalk
The following items still require cutting 2: stick, 2x4, birchbark, willowbark

I considered removing the softer ingredients from the cutting 2 recipe, but decided to keep the recipe as-is in order to not have to add the new recipes to the basecamp recipes, while still maintaining the ability to use your basecamp to turn paper into tinder.

#### Testing
I tried the recipes in-game, they seems to work fine.

### DIFF
--- a/data/json/recipes/recipe_ammo.json
+++ b/data/json/recipes/recipe_ammo.json
@@ -895,5 +895,48 @@
         [ "straw_pile", 1 ]
       ]
     ]
+  },
+  {
+    "type": "recipe",
+    "activity_level": "LIGHT_EXERCISE",
+    "result": "tinder",
+    "id_suffix": "softer",
+    "category": "CC_OTHER",
+    "subcategory": "CSC_OTHER_MATERIALS",
+    "skill_used": "survival",
+    "difficulty": 1,
+    "time": "1 m",
+    "autolearn": true,
+    "qualities": [ { "id": "CUT", "level": 1 } ],
+    "components": [
+      [
+        [ "splinter", 2 ],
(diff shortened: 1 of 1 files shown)


Maleclypse:
Did you mean to leave all the softer items as potential components for the recipe requiring cut 2?

Light-Wave:
> Did you mean to leave all the softer items as potential components for the recipe requiring cut 2?

Yes, because the basecamp recipes uses the cut 2 recipe.
If I removed the softer ones from cut 2, I would need to also add the other recipes to the basecamp crafting list.

I could do that, but I feel that would add excessive bloat to the list of basecamp recipes.

I could also add a hidden 'basecamp' recipe that is only used by the basecamp. But that seems hacky and would run the risk of not getting updated when the other recipes for tinder are changed.

Maleclypse (approved):
This makes sense.

Issue #64232, opened 2023-03-13
Microwave doesn't work when placed as an appliance


CansecoGPC:
### Describe the bug

When i try to craft or remove a microwave on the power grid it yields an error about pseudo_microwave not defined and it doesn't work to heat food. Works fine with batteries.

 DEBUG    : Missing item definition: pseudo_microwave

 FUNCTION : const itype* Item_factory::find_template(const itype_id&) const
 FILE     : src/item_factory.cpp
 LINE     : 2400
 VERSION  : a98d542

### Attach save file
[Kermit-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10960024/Kermit-trimmed.tar.gz)

### Steps to reproduce

1 Debug menu - Spawning - Spawn item - microwave
2 Place microwave using construction menu
3 Try to craft something, error shows up, microwave is not an option when selecting heat appliance/tool
4 Examine microwave, no heating option

### Expected behavior

A microwave with incorporated and charged heavy battery has actions to heat food, except battery can't be replaced unless charge is 0 or UPS conversion mod doesn't work, another bug report /request feature for later i guess.

### Versions and configuration

- OS: Linux
    - OS Version: LSB Version: n/a; Distributor ID: ManjaroLinux; Description: Manjaro Linux; Release: 22.0.4; Codename: Sikaris; 
- Game Version: a98d542 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

CansecoGPC:
Fixed recently, closing.

PR #60232, opened 2022-08-17
Fix examine actions and crafting qualities for several vehicle parts and appliances

ZeroInternalReflection:
#### Summary
Bugfixes "Fix examine actions and crafting qualities for several vehicle parts and appliances"

#### Purpose of change
The crafting pseudo item changes I made in #56991 broke crafting qualities (e.g. the chemistry lab providing "Chemical making 3") and examine actions (e.g. "`e`->purify water in tank" for a vehicle-mounted water purifier) for the vehicle parts/appliances changed.

Fixes #60144
Fixes #60150 
Fixes #60318

#### Describe the solution
Modify the crafting pseudo items to use both "copy-from" (for examine actions and qualities) and "sub" (for properly handling charge usage on the crafting menu).  Add "fake_item" properties (e.g. "PSEUDO" and "ZERO_WEIGHT" flags) individually or through the use of abstracts rather than a "copy-from" chain to "fake_item".
Most pseudo items have their own entry to allow them to be properly identified (e.g. "mounted electric forge" vs. "electric forge appliance").  Water purifiers are the exception, where all vehicle parts/appliances use pseudo_water_purifier, since the vehicle-handling code is currently set up to only check a single id for "purify tank" actions

#### Describe alternatives you've considered
- Change JSON handling to allow multiple inheritance (e.g. start with water purifier, then extend/replace with fake_item)

#### Testing
I have tested individual appliance/vehicle setups (just battery, part being tested, and perhaps a tank, set 7 squares away from the next one) for the following, confirming the following behaviours:

Oven appliance
 - `e`->use oven examine action
 - Crafting - provides oven
 - Can use full charges of the grid

Water purifier appliance
- `e`->use water purifier examine action
- `e`->purify in tank is available in UI, but I don't know of any way to attach a tank to purify
- Crafting - able to purify water
- Cannot test charge usage >50

Forge appliance
- Crafting - provides forge
- Can use full charges of the grid

Kiln appliance
- Crafting - provides kiln
- Can use full charges of the grid

FoodCo appliance
- `e`->use water purifier examine action
- `e`->purify in tank is available in UI, but I don't know of any way to attach a tank to purify
- Crafting - provides vacuum sealer
- Crafting - provides water purifier
- Crafting - provides dehydrator
- Crafting - provides food processor
- Crafting - provides hand press
- Disassembly - provides bullet pulling
- Can use full charges of the grid

Chemistry lab appliance
- `e`->use hotplate examine action
[...]

PR #64279, opened 2023-03-15
Fix app_microwave lack of microwave tool

GuardianDll:
#### Summary
None
#### Purpose of change
Appliance microwave doesn't work, because uses nonexisted/abstract `pseudo_microwave`
#### Describe the solution
~~Replace `pseudo_tools` `pseudo_microwave` with `microwave`~~ see comments
#### Testing
Before
image
After
image
#### Additional context
Maybe also remove `pseudo_microwave_abstract`?  I can't find any usage of it in the files

### DIFF
--- a/data/json/furniture_and_terrain/appliances.json
+++ b/data/json/furniture_and_terrain/appliances.json
@@ -958,7 +958,7 @@
       },
       "removal": { "time": "1 m" }
     },
-    "pseudo_tools": [ { "id": "pseudo_microwave" } ],
+    "pseudo_tools": [ { "id": "microwave" } ],
     "size": 20,
     "breaks_into": [
       { "item": "plastic_chunk", "count": [ 4, 7 ] },
--- a/data/json/furniture_and_terrain/appliances.json
+++ b/data/json/furniture_and_terrain/appliances.json
@@ -958,7 +958,7 @@
       },
       "removal": { "time": "1 m" }
     },
-    "pseudo_tools": [ { "id": "microwave" } ],
+    "pseudo_tools": [ { "id": "pseudo_microwave" } ],
     "size": 20,
     "breaks_into": [
       { "item": "plastic_chunk", "count": [ 4, 7 ] },
--- a/data/json/items/fake.json
+++ b/data/json/items/fake.json
@@ -397,7 +397,7 @@
     "name": "hotplate abstract"
   },
   {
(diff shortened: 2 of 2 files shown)


irwiss:
I can suggest trying to add `"sub": "microwave"` to pseudo_microwave

In https://github.com/CleverRaven/Cataclysm-DDA/pull/60232 the `pseudo_<tool>` stuff was added to work around the fact handheld tools take medium batteries with 500 charges max instead of making the vehicle's batteries available, they instead use a huge pseudo battery that can hold a bunch of charges ( no idea if the pseudo_microwave is set up correctly for this though )

GuardianDll:
I am actually lost, because `pseudo_microwave` do not exist at all, again, there is only `pseudo_microwave_abstract`, and using it do not change anything: the game just print `Missing item definition: pseudo_microwave_abstract`, same as before

irwiss:
Sorry I got distracted and then took a while to figure out another related bug

With changes in your PR batch of 41 pretzels (requiring 1025 charges, more than `microwave` with heavy battery has):
image
So it'll raise a query about making it anyway with insufficient energy

My idea of the fix is the following:
image

This allows up to ~100kJ battery charges used for it;
image

42 pretzels is unlikely case but batch crafting something that wants >25kJ can open up the incorrect confirmation dialog at lower batch counts

Related bug: Aftershock overrides `surface_heat` and that list doesn't have microwaves and a lot of other tools to begin with this is probably a bug

And pseudo tools should probably check for valid itype_ids in vpart_info::finalize() to prevent bugs like these

irwiss:
This tramples over updated `surface_heat` when aftershock is loaded https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/mods/Aftershock/recipes/requirements.json#L3
Same with water_boiling_heat

data/json/furniture_and_terrain/appliances.json
```diff
@@ -958,7 +958,7 @@
       },
       "removal": { "time": "1 m" }
     },
-    "pseudo_tools": [ { "id": "pseudo_microwave" } ],
+    "pseudo_tools": [ { "id": "microwave" } ],
```
GuardianDll, line 961:
```suggestion
    "pseudo_tools": [ { "id": "pseudo_microwave" } ],
```

CansecoGPC:
This patch solves this bug #64232 i reported days ago, thanks.

PR #64282, opened 2023-03-15
Aftershock - extend cooking requirements instead of replacing

irwiss:
#### Summary
None

#### Purpose of change

Aftershock overrides cooking / water heating requirements breaking half of https://github.com/CleverRaven/Cataclysm-DDA/pull/62180

#### Describe the solution

The purpose of the override seems to add afs_atompot to the lists, use copy-from and extend instead

#### Testing

After patch games with aftershock should receive the updated requirement lists

#### Additional context

Before:
image
After:
image


Maleclypse (approved):
Thanks!

PR #64288, opened 2023-03-16
Unique NPC Faction Specials Mod

anoobindisguise:
#### Summary
Mods "Unique NPC Faction Specials mod"

#### Purpose of change
I was unreasonably annoyed by clones of NPCs appearing in the world and it immediately took me out of any sense of immersion in the world. However, it's not a change we can make in the base game because making them unique would cause a ton of problems; I would personally rather deal with those problems than have copies of my NPCs.

#### Describe the solution
Make it a toggleable balance mod, obviously this is disabled by default and will exist under "you install this with the understanding it can break your game"

Did I make the mod correctly? Is there a better way to do this? I'm copying No Hope's implementation for specials but there might be a better way.

#### Testing
I am going to personally test this in my games for awhile before I take it out of draft.


LovamKicsiGazsii:
How is this different from #63627?

anoobindisguise:
> How is this different from #63627?

The primary difference is that I didn't know the other one already existed

PR #64296, opened 2023-03-16
Fix to pollution by stat modifications in newcharacter.

### LINKED ISSUE #64280: More character creation STATS panel bugs.
### Describe the bug

The STAT panel when creating a new character doesn't properly display the effects of the statistic you are inspecting.
With StatsThroughSkills enabled, stats can be added under the hood during character creation. Increasing your skills enough to add stats will start polluting the visuals for every button press in the menu. This is a cosmetic bug that doesn't affect gameplay.

This is the same issue as the BMI rework unearthed, Information displayed can be polluted by effects outside of the base statistics of the character. Looking into the avatar::reset_stats() function should reveal all potential edge cases, such as mutations. All changes inside that function that calls mod_stat_bonus(amount) seem to apply itself repeatedly with each button press. Replace 'stat' with str/dex/int/per.

### Attach save file

N/A

### Steps to reproduce

1. Create a world with only StatsThroughSkills as the only mod.
2. Skip all details and tab to skills, add 1(2) points into any two of these: Dodge, Piercing, Unarmed, Survival, Tailor, Devices, Vehicles.
3. Tab backwards to STAT page, and go between Strength and Dexterity. Observe that the bonuses of Dexterity will go up repeatedly.
4. Go forward to skills again and add 1(2) points into all skills. 
5. Tab backwards to the STAT page and flick through all statistics. Observe all stats gaining benefit.

### Expected behavior

Effects that are outside the scope of picking your stats should not be visible on your STAT page.

### Screenshots

image

- OS: Windows
- Game Version: b908848 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

Wyghab:
#### Summary
Bugfixes "Fixes the situation where stat mods affect the displayed effects of your base stats."

#### Purpose of change
Fixes #64280

#### Describe the solution
avatar::reset_stats() is accumulating stat mods each passthrough because character::reset_bonuses() is never called on the object. reset_stats() should never be called without first resetting bonuses as it can reapply additional bonuses without clearing the last ones. Ideally character::reset() should be used in these situations? Regardless, I find that no mod bonuses should affect the displayed effects of your base stats, hence I've placed reset_bonuses() after reset_stats() so it only displays the effect of your base stats, though it could still be polluted by other means as I've not investigated the functions that calculate the bonuses that are affected by the stats.

Calling reset() instead of reset_stats() or extracting the calculations used to determine stat effects and calculate the bonuses in-place instead of relying on the avatar object.

#### Testing
Tested #64280, while also including the mutation "compound eyes" as that mutation will continually mod perception by 2 while equipped in the same manner as StatsThroughSkills or the BMI rework.

#### Additional context
This could likely need some refactoring later. Since this fix is closer to the core issue than #64270 , maybe the fixes from there can be removed safely. If there are other places in code that calls reset_stats() haphazardly it could cause issues in other places.

### DIFF
--- a/src/newcharacter.cpp
+++ b/src/newcharacter.cpp
@@ -1125,6 +1125,7 @@ void set_stats( tab_manager &tabs, avatar &u, pool_type pool )
         werase( w_description );
         u.reset_stats();
         u.set_stored_kcal( u.get_healthy_kcal() );
+        u.reset_bonuses(); // Removes pollution of stats by modifications appearing inside reset_stats(). Is reset_stats() even necessary in this context?
         switch( sel ) {
             case 1:
                 mvwprintz( w, point( 2, 5 ), COL_SELECT, _( "Strength:" ) );


NetSysFire:
Feels like a good backport candidate.

Wyghab:
> Feels like a good backport candidate.

Just to keep the thread of change going for eventual refactor, this issue was caused by #57884. That PR assumed that using the avatar class to calculate everything and using reset_stats to update it to be less bug-prone, which seems untrue. So using the stat_max to calculate bonuses seem to be the way forward if what we want is for the stat panel to only care for your base stats when showing their effects.

PR #64301, opened 2023-03-16
Make vehicle mass aware of monster moving

SurFlurer:
#### Summary
Bugfixes "Make vehicle mass calculation aware of monster moving"

#### Purpose of change

At present, the vehicle doesn't know that a monster is moving onto it or away from it, so its mass will remain untouched unless "e"xamined. This makes it possible to ride a bike at 6km/h, leashing a friendly cow, and wait for the cow to step onto the bike's basket. Then it will settle there, without making the bicycle too heavy to ride.

#### Describe the solution
When the monster moves, check if there's vehicle at the origin tile and destination tile. If true ~~and the vehicle is moving ( to save CPU cycles when there's a lot of monsters and a lot of parked cars in the town )~~, mark the vehicle's mass as dirty, so it get recalculated.

Mark the mass as dirty, only if the vehicle is being driven by the player character.

#### Testing
All tests passed. 
Rode a bike at 6km/h, leashing a friendly cow, and waited for the cow to step onto the bike's basket. The bike complained for being too heavy to ride, and it slowed down to 0km/h. After the bike stopped, the cow stepped away, and the bike started moving as usual. No change of cruise speed setting during this, always 6km/h.

### DIFF
--- a/src/monmove.cpp
+++ b/src/monmove.cpp
@@ -1727,9 +1727,20 @@ bool monster::move_to( const tripoint &p, bool force, bool step_on_critter,
                                 has_flag( MF_AQUATIC ) ? _( "dives" ) : _( "sinks" ), here.tername( destination ) );
     }
 
+    optional_vpart_position vp_orig = here.veh_at( pos() );
+    if( vp_orig && vp_orig->vehicle().is_moving() ) {
+        vp_orig->vehicle().invalidate_mass();
+    }
+
     setpos( destination );
     footsteps( destination );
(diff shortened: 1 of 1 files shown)


irwiss:
I think the is_moving check can be skipped in case someone decides to cownap using a helicopter which has no lateral movement or just skipped as it has no relation to mass changing;
invalidate_mass is cheap as it only sets some dirty flags, it doesn't recalculate the mass until something queries for it (therefore needing it and/or center of mass to be correct)

PR #64304, opened 2023-03-16
parameterized field base camp ver 1

PatrikLundell:
#### Summary

None

#### Purpose of change

Main purpose:
- Make use of parameter palettes to reduce the volume of the JSON descriptions for field base camp hub version 1.

Secondary purposes:
- Add the "new" ALLOW_TERRAIN_UNDER_OTHER_DATA flag to building construction to make it up to the current standard.
- Add the makeshift bed to the set of bed options (essentially a copy of the corresponding version 2 implementation).
- Mirrored west side building bed arrangement in the same manner as was implemented in version 2 while at it.
- Added concrete as a construction material (again, the use of the version 2 implementation made this trivial, and maybe even easier than not adding it).

Side effects:
- Someone decided spaghetti coding using version 1 palettes for other purposes was a good idea. Thus, the modified names of the palettes had to be used in these places as well. Either these places should have their own palette(s), or the version 1 palettes should be moved to a central palette location (like the old acadia ones) and be given a name not directly tied to any specific camp implementation.
- For some reason a pile of JSON files in Master did not adhere to the mandatory JSON format pattern. These were changed by json_formatter.exe to conform to the only permitted format as a result of running the recommended Visual Studio external tool on the project.

#### Describe the solution

Port the solution used for version 2 in #65183 to version 1, plus the things mentioned above.

Leave it to someone else, at some other time.

#### Testing

- Built the buildings of a base from scratch using each of the "real" (i.e. non tent) construction materials.
- Built each kind of bed in each of the buildings used in the construction path above.
- Build each of the heating/cooking options for the first room in different instances of the first set of tests.
- Built the 6 tent version of the base with some kind of bed plus the central building from a "random" material, i.e. mi-go resin, since that's the easiest one to use.
- Built the base with tents in the top 2 and 4 positions with a "random" kind of bed for the "real" bedroom buildings (because these use their own mapgen definitions).
- Built the central building from 2, 3, and 4 construction steps (there are recipes that combine the two top quarters and the bottom quarters respectively) in the first step sequence.

During or after the construction of each test I walked around and looked at the base to verify that every room and room content looked as expected. Much of the time the walk around happened directly after a building the contents of a building.

Note that this means that every permutation hasn't been covered for:
- Beds in bedrooms that are specific for when tents are present (only one bed type tested for certain). No issues have been seen in any other case where this exact bed pattern has been used.
- Building materials of buildings specific for when tents are present (only mi-go resin tested). No issues have been seen in any other case where this exact building material pattern has been used.
- Central building construction of 2/3/4 steps combined with every construction material. Note that the recipes for the combination of two parts uses those exact same mapgen parts, so exactly those same palette parameters are used for both parts.

#### Additional context

The definition of the northwestern building is probably not quite right: it probably shouldn't replace the unused terrain with bare dirt to the north and to the east. Since it's not important and potentially might even have been intended (for some unknown reason) it wasn't changed.

Edit: Well, the stale-bot killed this PR.

It's a pity that the functionality developed to allow this PR was wasted, as those with the power to approve PRs have decided not to bother with reviewing this PR. As a result, the planned updates of other base camp components are now cancelled as they're expected to be ignored as well.


Maleclypse:
This is editing a couple of files I don't believe it intends to.  Can you correct and ping me when it's ready for review?

PatrikLundell:
I don't know which files you think aren't intended to be included.

Do you mean the JSON files I hadn't modified but the json formatter nevertheless changed to match the mandated format? Those changes were kept because every use of that tool would change those files into the mandated format for every sod unfortunate enough to have to use the tool. Is there a reason for them not to conform to the mandated format? I had thought kevingranade's changed the one that shouldn't be adjusted.

PR #64306, opened 2023-03-16
make Bulging Reservoir decrease your thirst, not increase

GuardianDll:
#### Summary
None
#### Purpose of change
Bulging Reservoir description says it should decrease the person's thirst, but instead it increases it for 75%
#### Describe the solution
swap 75% increase with 25% decrease, as intended
#### Testing
done
#### Additional context
8k strings in mutation.json is a mess

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -3296,7 +3296,7 @@
     "visibility": 4,
     "ugliness": 3,
     "description": "You develop a reservoir of fluids that bulges out of your abdomen.  Your body has adapted to containing extra liquids in event of shortfalls.",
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):

Maleclypse:
I'd rather this actually allowed the player to store more liquid but that would require c++ changes.

GuardianDll:
Depend on what you mean by "stored" - it can be a common integrated armor/pocket, which can contain only water

Maleclypse:
> Depend on what you mean by "stored" - it can be a common integrated armor/pocket, which can contain only water

I meant by moving the maximum negative thirst.

PR #58104, opened 2022-06-03
New Naval Monsters

DoctorBoomstick:
#### Summary
Content "Adds a few new naval themed zombies and feral humans to the aircraft carrier map."

#### Purpose of change
Continuing work started in PR #58049 , where new naval professions were added, this PR aims to make the aircraft carrier start and general location more unique, realistic, and varied. Previously, the carrier’s spawn group had included within it regular feral humans and military/ civilian zombies. On the first count, I felt like these feral humans, the axe, crowbar, pipe, hammer, and spear wielding feral humans, were, from a realism perspective, rather out of place to be abord a carrier which was on active deployment. Firstly, many of these enemies have a rock throwing capability, which doesn’t make sense for the context of a carrier due to the fact that there’s a slight shortage of large chunks of rock present on a military boat. Secondly, when killed all these enemies constantly drop civilian clothing. If one is to assume that the ship was on active deployment, as described within the scenario description, it would be safe to conclude that many of the people abord the ship would be naval personnel, marines or sailors. As such, it doesn’t make sense for such a large number of people abord to be dropping fully civilian gear and clothing all the time. On account of zombies, the second issue carries into the zombie contingent abord the ship as well, and whilst having stock standard zombie soldiers and uniformed zombies is perfectly fine and serviceable, I thought that it’d be fun to add a few specifically naval themed undead.
#### Describe the solution
I’ve added a fair few feral humans and zombies to replace the current aircraft carrier’s contingent of enemies.
[...]

PR #64309, opened 2023-03-16
Renames the Devil dog and twisted leatherneck zombies.

DoctorBoomstick:
#### Summary
None
#### Purpose of change
For the last few months, I’ve made note of many instances during which players showed a dislike for the names and descriptions assigned to the Devil Dog and Twisted Leatherneck, which I added in PR #58104. Most of the complaints revolved around the names being too cryptic and edgy for the monsters they represent, with several people not familiar with US military slang finding it necessary to resort to internet searches to discover the context behind the names. In addition, with respect to the devil dog in particular, I saw a handful of comments that questioned its design and reason for being, as it is ultimately just a redressed zombie soldier, with the specification of its Marine origins being unnecessary clutter.
#### Describe the solution
I have renamed both monsters in question, as well as trimming their respective descriptions and removing unneeded elements.
I considered removing them outright, as I honestly felt like they were more hassle than they were worth, but I thought that trying these smaller changes might serve as a midway point between a complete removal and an entire redesign.
#### Testing
The changes are all simple string alterations, save for a single additional line which ups the (formerly named) Twisted Leatherneck’s speed to bring it inline with what my original views were over the creature.


Karol1223:
During the last conversation about this topic, which I suspect is the one that sparked this PR, the most notable thing I kept repeating throughout is that variety is fun - its the confusion that isn't. What I'm saying here is leaving them in the game is a GOOD choice. Having a zombie that is functionally identical to something else but in a different flavor is good if the player can read what it is. I am a fan of more variety and I am glad you didn't opt to remove them altogether.

RenechCDDA:
As above, this seems like overkill. Just assign them differentiating symbols so they're easier to manage if the tilset doesn't have sprites for them.

Karol1223:
Upon reading the actual code - yeah, the name was changed to just be a zombie soldier and yeah I would call that a bit overkill.

Having them have a unique name is fine, just one that is easily readable. 'Zombie Marine' (I have no idea if they're Marines, I don't know shit about the army, this is just an example) is perfectly viable and adds that sliver more of variety to the world.

PR #64313, opened 2023-03-16
Remove unused colored stat display functions

sparr:
#### Summary
None

#### Purpose of change
The four stat text color functions have a lot of duplicated code.

#### Describe the solution
~~Move the duplicated code to a new function.~~
Removed the code as requested.

Remove the code since it seems to be unused since https://github.com/CleverRaven/Cataclysm-DDA/commit/adb20a2a9026dc390550dd700a2e577e2ac62057

### DIFF
--- a/src/display.cpp
+++ b/src/display.cpp
@@ -109,64 +109,41 @@ vehicle *display::vehicle_driven( const Character &u )
     return veh;
 }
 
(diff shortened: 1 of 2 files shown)


kevingranade:
If it's really dead code I'd rater it be gone.

PR #64314, opened 2023-03-16
Switch a time type to fs::

jbytheway:
#### Summary
None

#### Purpose of change
There were build errors on the OS X build.  See for example [this one](https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4437676544/jobs/7789080345#step:16:137).  They seemed to only appear on the release build, not the PR CI build.

We were using `std::filesystem::file_time_type` when we should have been using `fs::file_time_type`.  These types are the same on some platforms, but not on others.

#### Describe the solution
Switch to the correct type.

None.

#### Testing
I verified it still builds on Linux, but I can't test on OS X.  I'm not sure even CI will test this properly.  We might have to just merge and hope the release builds do better.

#### Additional context

### DIFF
--- a/src/flexbuffer_cache.cpp
+++ b/src/flexbuffer_cache.cpp
@@ -162,7 +162,7 @@ struct file_flexbuffer : parsed_flexbuffer {
 
         bool is_stale() const override {
             std::error_code ec;
-            std::filesystem::file_time_type mtime = fs::last_write_time( source_file_path_ );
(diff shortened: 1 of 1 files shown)


BrettDong:
The reason why it compiled on CI but failed on release build is because the target macOS version in CI build is 10.15 but it's 10.12 in release build.

`ghc` file system library delegates to std library on macOS >= 10.15, and uses its own implementation on older systems.

PR #64323, opened 2023-03-17
Make skill help spell learn time instead of hurt

sparr:
#### Summary
Bugfixes "Higher skill speeds up spell learn time instead of slowing down"

#### Purpose of change
Increased spellcraft (or other specific) skill causes spell learning to slow down. This appears to be an arithmetic/parentheses mistake.

#### Describe the solution
Added parentheses to calculation so that skill is added to the divisor instead of to the multiplier.

None

#### Testing
With the original code and 10 int, "One spell is all you need" book takes 78 minutes to read at spellcraft 0, 108 minutes at spellcraft 10.
With this patch, it still takes 78m at skill 0, but 58m40s at skill 10.

#### Additional context
It might make sense to allow high skill to reduce the time even further, possibly below the base time, but that would be a more significant change. I attempted here to follow what seems like the most likely intent of the original code, which only requires moving a parenthesis (although I added another pair for clarity as well).

### DIFF
--- a/src/magic.cpp
+++ b/src/magic.cpp
@@ -1936,8 +1936,9 @@ int known_magic::time_to_learn_spell( const Character &guy, const std::string &s
 int known_magic::time_to_learn_spell( const Character &guy, const spell_id &sp ) const
 {
     const int base_time = to_moves<int>( 30_minutes );
-    return base_time * ( 1.0 + sp->difficulty / ( 1.0 + ( guy.get_int() - 8.0 ) / 8.0 ) +
-                         ( guy.get_skill_level( sp->skill ) / 10.0 ) );
+    const double int_modifier = ( guy.get_int() - 8.0 ) / 8.0;
+    const double skill_modifier = guy.get_skill_level( sp->skill ) / 10.0;
+    return base_time * ( 1.0 + sp->difficulty / ( 1.0 + int_modifier + skill_modifier ) );
 }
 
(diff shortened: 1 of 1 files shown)


src/magic.cpp
```diff
@@ -1936,8 +1936,8 @@ int known_magic::time_to_learn_spell( const Character &guy, const std::string &s
 int known_magic::time_to_learn_spell( const Character &guy, const spell_id &sp ) const
 {
     const int base_time = to_moves<int>( 30_minutes );
-    return base_time * ( 1.0 + sp->difficulty / ( 1.0 + ( guy.get_int() - 8.0 ) / 8.0 ) +
-                         ( guy.get_skill_level( sp->skill ) / 10.0 ) );
+    return base_time * ( 1.0 + sp->difficulty / ( 1.0 + ( ( guy.get_int() - 8.0 ) / 8.0 ) +
+                         ( guy.get_skill_level( sp->skill ) / 10.0 ) ) );
```
irwiss, line 1940: May be taking it apart would make it more human readable e.g.
```suggestion
    const double int_modifier = ( guy.get_int() - 8.0 ) / 8.0;
    const double skill_modifier = guy.get_skill_level( sp->skill ) / 10.0;
    return base_time * ( 1.0 + sp->difficulty / ( 1.0 + int_modifier + skill_modifier ) );
```

sparr: Good call

PR #64324, opened 2023-03-17
Improve new character stats interface and code

sparr:
#### Summary
Interface "Add min/max feedback to new character stats interface"

#### Purpose of change
The new character stats interface has a lot of duplicated code, and unnecessary handling of non-zero-based indices.

#### Describe the solution
Deduplicated and simplified the code where possible.

#### Testing
Manipulated the interface to every extreme and combination I could think of.

#### Additional context
I also added a few new bits of feedback for when the stats are at min or max, which is the only user-facing change here.

I anticipate following up with a PR to deduplicate all the different places `_( "Strength ")` etc exist.

### DIFF
--- a/src/newcharacter.cpp
+++ b/src/newcharacter.cpp
@@ -1071,9 +1071,12 @@ void set_points( tab_manager &tabs, avatar &u, pool_type &pool )
 
 void set_stats( tab_manager &tabs, avatar &u, pool_type pool )
 {
+    // TODO: Move this out to a common header and eliminate other separate instances of these strings.
+    static const std::array<translation, 4> stat_labels = { to_translation( "Strength" ), to_translation( "Dexterity" ), to_translation( "Intelligence" ), to_translation( "Perception" ) };
     const int max_stat_points = pool == pool_type::FREEFORM ? 20 : MAX_STAT;
+    const int min_stat_points = 4;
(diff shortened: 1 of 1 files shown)


src/newcharacter.cpp
```diff
@@ -1063,11 +1063,14 @@ void set_points( tab_manager &tabs, avatar &u, pool_type &pool )
     } while( true );
 }
 
+static const std::array<std::string, 4> stat_labels = { _( "Strength" ), _( "Dexterity" ), _( "Intelligence" ), _( "Perception" ) };
```
kevingranade, line 1066: I think all you need to do here is move this inside set_stats(), then it will be a function level static with deferred initialization.

kevingranade (changes requested):

src/newcharacter.cpp
sparr: I intend to follow this up with consolidating all the places that reproduce this list. I've pushed a change that I think will work here, and indicating how I'd proceed on a larger scale if this is acceptable and works.

sparr: Please let me know if the update is also unacceptable. If so then I'll move it to a normal function level static and make the global a separate PR when I deduplicate all the uses.

PR #64327, opened 2023-03-17
Deduplicate book read time code

sparr:
#### Summary
None

#### Purpose of change
`avatar::read`, `Character::time_to_read`, and `npc::time_to_read` have duplicated code for calculating reading speed and some of the factors that contribute to it.

#### Describe the solution
Move some code out to helper functions, deduplicate uses of time_to_read and its helpers.

#### Testing
I have not come up with a good test case for reading with both npc helpers and learners, and would appreciate assistance in doing so.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -10786,6 +10786,18 @@ const Character *Character::get_book_reader( const item &book,
     return reader;
 }
 
+bool Character::read_effective_int( const Character &reader, const Character *learner ) const
+{
+    return std::min( { get_int(), reader.get_int(), learner ? learner->get_int() : INT_MAX } );
+}
+
+bool Character::complex_read_penalty( int book_intel, const Character &reader,
+                                      const Character *learner ) const
+{
+    return book_intel > read_effective_int( reader, learner ) &&
+           !reader.has_trait( trait_PROF_DICEMASTER );
+}
+
 time_duration Character::time_to_read( const item &book, const Character &reader,
                                        const Character *learner ) const
 {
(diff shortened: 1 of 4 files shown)


src/character.cpp
```diff
@@ -10803,9 +10815,9 @@ time_duration Character::time_to_read( const item &book, const Character &reader
     time_duration retval = type->time * reading_speed / 100;
     retval *= std::min( fine_detail_vision_mod(), reader.fine_detail_vision_mod() );
 
-    const int effective_int = std::min( { get_int(), reader.get_int(), learner ? learner->get_int() : INT_MAX } );
-    if( type->intel > effective_int && !reader.has_trait( trait_PROF_DICEMASTER ) ) {
-        retval += type->time * ( time_duration::from_seconds( type->intel - effective_int ) / 1_minutes );
+    if( complex_read_penalty( type->intel, reader, learner ) ) {
+        retval += type->time * ( time_duration::from_seconds( type->intel - read_effective_int( reader,
+                                 learner ) ) / 1_minutes );
```
irwiss, line 10820: Seems the right side is used as a multiplier rather than a time_duration, possible to divide by 60.0 to get same result?
```suggestion
        retval += type->time * ( type->intel - read_effective_int( reader, learner ) ) / 60.0;
```

sparr: This change results in minor differences in arithmetic results and [failed tests](https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4455938873/jobs/7826060866?pr=64327#step:15:305). The code in the PR is taken from an existing implementation, so I'd like to keep it unchanged.

irwiss: Grandfathering weird stuff because of snapshotted test feels weird but that's a fair argument for a refactor PR

Although I think the cause is time_duration doing integer math so it should be `/ 60` instead of `/ 60.0` or something similar

PR #64333, opened 2023-03-17
Migrate from cata::optional<T> to std::optional<T>

BrettDong:
#### Summary
None

#### Purpose of change
The `cata::optional<T>` polyfill is no longer necessary as we are in C++17 now.

#### Describe the solution
`cata::optional<T>` and `std::optional<T>` are functionally equivalent, but they have some minor language level differences. So all uses of `cata::optional<T>` and `cata::nullopt` are massively transformed into `std::optional<T>` and `std::nullopt` first, and then some small manual changes were done to fix compile errors and warnings:

* `cata::optional<T>` is not trivially copyable but `std::optional<T>` is, provided that its inner type `T` is trivially copy constructible. So all occurrences of `std::move()` from a `cata::optional<T>&&` to a `cata::optional<T>` object where `T` is trivially copy constructible are removed to solve `clang-tidy` warning
```
error: std::move of the variable 'p' of the trivially-copyable type 'std::optional<vpart_position>' has no effect; remove std::move() [performance-move-const-arg,-warnings-as-errors]
```
* `cata::optional<T>` permits certain level of implicit conversion on its inner type, but `std::optional<T>` is more strict about implicit type conversion. Some manual `static_cast<>` are added to solve MSVC warnings.
* `cata::optional<T>` has a user-defined default constructor, but in `glibcxx`, `std::optional<T>` has a defaulted default constructor, so Clang 12 refuses to generate a defaulted default constructor for a class with a `const std::optional<T>` member variable, although Clang 16 compiles fine without the probelm. Need to manually define the constructor for `struct item_tweaks`.
```
warning: explicitly defaulted default constructor is implicitly deleted [-Wdefaulted-function-deleted]
```
* Got some lifetime warnings from Clang 16, so manually edited a few uses of `wear_item()` in **test code only**.
```
../tests/item_autopickup_test.cpp:206:24: error: object backing the pointer will be destroyed at the end of the full-expression [-Werror,-Wdangling-gsl]
    item &backpack = **( they.wear_item( item( itype_backpack ) ) );
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../tests/item_autopickup_test.cpp:283:24: error: object backing the pointer will be destroyed at the end of the full-expression [-Werror,-Wdangling-gsl]
    item &backpack = **( they.wear_item( item( itype_backpack ) ) );
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../tests/item_autopickup_test.cpp:323:24: error: object backing the pointer will be destroyed at the end of the full-expression [-Werror,-Wdangling-gsl]
    item &backpack = **( they.wear_item( item( itype_backpack ) ) );
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../tests/item_autopickup_test.cpp:443:24: error: object backing the pointer will be destroyed at the end of the full-expression [-Werror,-Wdangling-gsl]
    item &backpack = **( they.wear_item( item( itype_backpack ) ) );
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../tests/item_autopickup_test.cpp:518:24: error: object backing the pointer will be destroyed at the end of the full-expression [-Werror,-Wdangling-gsl]
    item &backpack = **( they.wear_item( item( itype_backpack ) ) );
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

#### Testing
Wait for all CI configurations to build successfully.


kevingranade:
Looks good but I'd like to see all the builds complete.

BrettDong:
The clang-tidy error originated from #64174 and is not related to this pull request. Apart from that I see persistent errors related to vehicle passenger boarding but they only occur in MSVC and MinGW cross compile builds, not on Linux and macOS.

tests/optional_test.cpp
```diff
@@ -1,29 +1,29 @@
 #include <new>
```
Qrox, line 1: Tests in this file are no longer necesary after switching to `std::optional`.

BrettDong:
The same vehicle boarding error occurred in another pull request: https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4453248488/jobs/7821721631?pr=64352#step:15:934

So I'm convinced it's not caused by `optional<T>` switch.

jbytheway:
Both optional migration PRs have test failures related to folded vehicle masses, so that might be genuinely caused by the migration.

https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4450133109/jobs/7821018468?pr=64333#step:19:428

BrettDong:
`optional_vpart_position` looks suspicious. It inherits from `std::optional` which may cause unintended side effects depending on standard library implementation. I'll try refactor it to use composition and see if it's related to those vehicle test errors.

BrettDong:
It's strange that vehicle fold mass errors also occurred here in a non `optional` related pull request https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4459588843/jobs/7832205776?pr=64116

BrettDong:
So this is it: https://github.com/CleverRaven/Cataclysm-DDA/pull/64384

Vehicle boarding test errors are related to https://github.com/CleverRaven/Cataclysm-DDA/pull/64331

Issue #62416, opened 2022-11-26
Extension Cords and Wired Tools


Playyne:
### Is your feature request related to a problem? Please describe.

With the appliance system in a quite functional state at the moment, it feels kind of strange to not be able to plug certain tools into the household power grid like you can do in reality. It is possible with furniture or vehicle-sized appliances, but small kitchen appliances like food processors, water purifiers, hotplates, and so on still run on internal batteries. Battery-powered kitchen appliances do exist in reality but aren't particularly commonplace. This was I assume a compromise for gameplay purposes back in the day, since according to the story, the public power grid was out due to the cataclysm, and no system that simulated mains power existed yet. We have one now, so I think it would be a good addition for the game if it were possible to incorporate normally cord-powered tools into the appliance system.

### Solution you would like.

The way I think you'd go about implementing this would be to give certain wired tools a POWER_CORD flag where it would make sense (like hotplates, but not flashlights) that lets them draw power from a nearby appliance power network in range. But this might be a bit finicky if you don't want to use the appliance power in some cases. Another would be to implement a "kitchen counter top" appliance, or allow certain spaces or storage furniture tiles to be wired with an "extension cord hub" appliance, where the flagged tools can either be installed through some unique menu, or just placed into it as items, to draw power (and charge the internal battery if present).
Some tools could have their internal battery compartments removed, to make them purely grid-powered, like most kitchen appliances are, but that might be a harsh balance change many would dislike. Adding additional tool mods, like a power cord attachment to exclusively battery-powered tools, or a battery compartment to exclusively cord-powered tools would add some depth and customization.

### Describe alternatives you have considered.

The UPS conversion kit used to work similarly to my suggestion where you could leave modded tools in a recharge station on a vehicle and let it draw power from the vehicle grid. That quirk was removed around the time battery magazines were implemented, so restoring that feature in some way would be a quick and dirty solution.

ParaplegicRacehorse:
I think this is a great idea, especially if it lets me 'place [item]' on top of existing furniture.  Having to move or destroy useful countertops just to set a stereo on it is annoying.

For power-tools, batteries are the way to go, since the world is currently moving in that direction anyway.  Other items, and especially kitchen appliances, would benefit greatly from this.

(also, @Playyne, the hotplate would be corded while the mess kit would still be battery operated or burn some sort of fuel)

PR #64334, opened 2023-03-17
Connect items to electrical grid; improved cables

### LINKED ISSUE #64375: Extension cord breaks when going up stairs, but not down
### Describe the bug

I'm trying to connect the large storage batteries on the underground floor of my LMOE to the solar panels at the top of the little hill thing. If I start at the batteries and go up, "The over-extended cable breaks loose!" as soon as I go up the stairs, but if I start at the solar panels and go down to the batteries it works fine.

### Attach save file

[Magic-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11009684/Magic-trimmed.tar.gz)

### Steps to reproduce

Attempt to connect things on different z-levels, going both up and down.

### Expected behavior

Extensions cords are bidirectional.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19045.2006 (22H2)
- Game Version: 53f9146 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm],
    Mythical Martial Arts [MMA]
]

### Additional context

_No response_

### LINKED ISSUE #65265: Solitary power cord on vehicle cannot be removed.
### Describe the bug

If I plug one appliance into the vehicle, and something shoots at the appliance so that the power cord gets damaged and drops onto ground, the other end of power cord on the vehicle cannot be regularly removed. I have to bash it until it breaks or drops onto ground, too.

### Attach save file

[npcneed-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11303008/npcneed-trimmed.tar.gz)

### Steps to reproduce

1. Load the save file.
2. Shoot at the ASRG until the power cord breaks down. ( Note that auto turrets will do the same thing----not necessarily player's fault.)
3. Find that you cannot remove the power cord on the vehicle. And you cannot unrack the bike now.

### Expected behavior

1. I don't think the power cord should exist as an item, since it's somewhat "fake".
2. There should be a way to uninstall the power cord, besides violently bashing.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19045.2846 (22H2)
- Game Version: 0.G-1503-ge8a1eafffb [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

Kamayana:
#### Summary
Features "Connect items to electrical grid; improved cables"

#### Purpose of change
Electrical devices rely on a variety of methods to emulate power sources:
- Swappable batteries, which are completely unlike modern-day devices that keep their batteries as inaccessible as possible.
- Appliance electrical grids, which works great for big things like refrigerators but is very unsuited to little items like cellphones.
- Cable active items, like extension cords. Mostly just serve as supplements to the electrical grid and bionics systems.
- Unified Power Supplies, but they're meant to be a military-designed system for their purposes, making it unusual to find on consumer devices.

While all of these methods work great, none of them accurately match the experience of taking a device and plugging it into an outlet to charge. It's always seemed odd that an activity so common in real life had no representation in-game, and in some situations like UPS cellphones it can be debilitating to not have the ability to just plug the thing in.

Partly fulfills #62416. Fixes #65265. Partly fixes #64375.

#### Describe the solution
Items can now have a CABLE pocket that's mostly inaccessible... except for the new "link_up" use action. That action will create a new active cable item, attach it to the specified electrical source, and put it inside the item's CABLE pocket. It will then behave like extension cords do - remaining active as long as it's within a certain distance of its attached point. As long as its cable is active it will gain effects - specifically charging the item's batteries from the attached power grid's stores, and using the power grid's electricity for any electrical use functions, working even when there's no battery in the device. There is also the possibility to use a negative charge value that'll charge the connected electrical grid at the expense of the device's battery, though nothing in this PR implements that.

In addition, the preexisting cable use functions have been thoroughly rewritten and all cable items are more flexible as a result. Dropping an active one on the ground will no longer deactivate it, and moving their attached vehicle/appliance will update their connection point accordingly as long as it's still in range. On one test run I connected 100 laptops to a Blackhawk's dashboard and flew at 255 mph with no disconnections. Here are the notable changes:
- A warning will be displayed in the log if a cable is stretched to its breaking point. Previously, the player had no feedback so the cable would just suddenly break with no warning.
- The cable menus are arranged differently - disconnecting is no longer the default option.
- Tow cables can now connect the tower and towee in either order.
- Cables can no longer be connected anywhere on a vehicle - extension cords need to connect to the dashboard or electronic controls, and jumper cables need to connect to a battery. Some vehicles that have batteries located deeper inside than the hood will be harder to use jumper cables with, but I feel like it makes more sense this way.
- Cables that disconnect after only having been stretched a few squares will be respooled immediately (simulating the player character simply jerking the plug out and letting it dangle). Cables that disconnect after stretching a longer distance will be labeled "(unspooled)" and need to be manually spooled up. Previously, the time spent spooling was immediate and uninterruptible, which could cause problems if an extension cord broke off while in danger.

Note: this PR only actually adds link_up functionality for some electronic tools: laptop, smartphone, eink tablet, mp3 player, and the handheld game system. They still have their swappable batteries/UPS support, since I wanted to focus solely on adding new mechanics while leaving everything mostly untouched. However, I have confirmed that other items will work and I plan to make a separate PR to add more of them. 

Cables are dynamically created and destroyed with the action, and the CABLE pocket sets their weight to 0. Basically, the player is assumed to be keeping a device's charging cable with the item at all times, and thus its weight is included with the device. I initially decided on this because I thought cable management would be overly fiddly, but as I kept working on it I found myself thinking it actually wouldn't be too bad, especially since this system has potential for ALL rope-like items, not just electrical cables. Regardless, I went with automatic cables for simplicity right now, while leaving open the possibility for a toolmod-like system for associating separate cable items with devices in the future.

A major choice I made early on was to decide who "owns" the cable - the device or the electrical grid. In the end I handled almost everything inside the process_cable function of the device itself. This has the advantage of keeping things compact and efficient, with the downside of having things mostly inaccessible outside of it; for example, there's no way currently for a vehicle to find items that are plugged into it. If we ever need such a function we could add a specific cable cache like how explosives and corpses are handled, but I found I was able to do without it.

#### Testing
The most obvious test, "power_cable_stretch_disconnect", passes with all green. All the cable tests in "fueled bionics" also pass. I also ran the profiler during a stress test of 100 laptops plugged into a helicopter and waiting for 6 hours. There's definitely a CPU hit, but it's not that bad for such an intense example.

Profiler result: https://user-images.githubusercontent.com/6759778/236055448-75c9b73c-a07c-4f95-b114-eac96d4f2bac.png

#### Additional context
image


Kamayana:
Note that all the notes that say "TODOkama" are things I plan to address before this leaves draft. One un-noted change I plan to make is switching plugging in from iuse to a more general action, since it screws up the quick action menu for devices that previously only did one thing, like MP3 players, to now have two iuses.

PatrikLundell:
Swappable batteries isn't a weird concept. It was (and might still be) a contender for vehicle "refueling", and I've had a lawn mower with swappable batteries (which was needed, as one battery wasn't enough for the not particularly large lawn, especially as the batteries aged).

Kamayana:
> Swappable batteries isn't a weird concept. It was (and might still be) a contender for vehicle "refueling", and I've had a lawn mower with swappable batteries (which was needed, as one battery wasn't enough for the not particularly large lawn, especially as the batteries aged).

> 

To be clear, I agree. I didn't mean to call _all_ swappable batteries weird; a lot of devices, especially tools, rely on batteries and cables, some even using _only_ batteries without a cable. I was talking specifically about desktop devices like smartphones that make their batteries hard to access.

Kamayana:
Very significant optimizations just pushed. The 100 laptop test went from 1525 CPU units to 230-370. Almost all of the processing is now entirely inside process_cable, rather than split between cables and vehicles with the map as a middleman. To do this I needed vehicle pointers that would fail safely, and I was able to do it by giving vehicles safe_reference_anchor, which was previously only used for item_locations. It worked like a charm! Cables will also only calculate length changes when either side of the cable moves

Battery charging will also now take time elapsed outside the bubble into account and catch up on draining/charging when you return.

image

Kamayana:
Whoops, I thought rebases didn't clutter the PR like that. Oh well. Here's the comment I left since it got buried:

> Very significant optimizations just pushed. The 100 laptop test went from 1525 CPU units to 230-370. Almost all of the processing is now entirely inside process_cable, rather than split between cables and vehicles with the map as a middleman. To do this I needed vehicle pointers that would fail safely, and I was able to do it by giving vehicles safe_reference_anchor, which was previously only used for item_locations. It worked like a charm! Cables will also only calculate length changes when either side of the cable moves, to further speed things up.
> 
> Battery charging will also now take time elapsed outside the bubble into account and catch up on draining/charging when you return.
> 
> image

src/item.cpp
```diff
@@ -13002,6 +13246,20 @@ bool item::process_internal( map &here, Character *carrier, const tripoint &pos,
...
     }
 
+    if( contents_linked ) {
+        std::vector<item *> cables = contents.cables( true );
+        if( !cables.empty() ) {
+            for( item *cable : cables ) {
+                if( cable->process_cable( here, carrier, pos, this ) ) {
+                    remove_item( *cable );
+                }
+            }
```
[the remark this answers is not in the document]
Kamayana: The formatter is wrong here, process_internal should keep going so it can process the device itself if needed.

src/item.cpp
```diff
@@ -10290,6 +10296,22 @@ int item::ammo_remaining( const Character *carrier ) const
...
+            } else {
+                const optional_vpart_position vp = get_map().veh_at( cable->link->t_abs_pos );
+                if( vp ) {
+                    ret += vp->vehicle().connected_battery_power_level().first;
+                    break;
+                }
+            }
+        }
+    }
+
```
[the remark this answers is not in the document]
Kamayana: This suggestion would makes things harder to parse in my opinion, I'd prefer to keep those sections separated by a line break.

Kamayana:
This was another major update. I wasn't happy with having distinct states for every combination of cable targets, so I split them into one for the target and one for the source. This meant I had to rewrite the pre-existing cable functions, so I took the opportunity to combine them all into one big function, since there was a lot of reused code there before. I also used this opportunity to fix some bugs I noticed:

- Solarpacks weren't properly removing their "plugged in" item variable. Now they work like the UPS and will check their active state.
- Tow cables would disconnect when _any_ part of its vehicle was removed. Now it's only the tow cable part and the structural part its connected to that'll do that.
- Smashing a vehicle tile that contains a tow cable part could potentially drop a damaged tow cable on the ground... while keeping a broken tow cable vehicle part still on the vehicle, essentially duplicating the cable. Now it'll just drop a single undamaged cable on the ground.
- You can no longer plug the same two appliances together over and over forever.
- Connecting cables now checks if the cable is actually a valid vehicle part, fulfilling the TODO that asked for such a check.

It's been a while but it's very close to being finished, now. I think the only major thing remaining is re-implementing cable spooling time.

Kamayana:
One last update, mostly making sure I fixed the issues I promised I would. I also made it so jumper cables can only connect to car batteries specifically, and extension cords to dashboards specifically. The extension cord thing is a little odd since cars don't normally have normal electrical outlets that can handle plugs like that, but I didn't want to remove that functionality since it's a vital part of how people have their electrical grids set up and there isn't a proper replacement for it (that said, my own car IRL actually _does_ have a normal electrical outlet, so there's _some_ basis in reality there).

It's been a lot longer than I anticipated, but with this I'm finally ready to take this PR out of drafts.

Kamayana:
The remaining Clang-tidy errors are complaints about certain variables not being serialized, but those variables get generated elsewhere and would not benefit from serialization, so it shouldn't be an issue.

irwiss:
it'll be a problem because clang-tidy will fail the check for any
subsequent commit, you can use //NOLINT or //NOLINTNEXTLINE to annotate
that those get generated/shouldn't serialize

at a glance you can probably get rid of the new include in item.h too

Kamayana:
> it'll be a problem because clang-tidy will fail the check for any subsequent commit, you can use //NOLINT or //NOLINTNEXTLINE to annotate that those get generated/shouldn't serialize at a glance you can probably get rid of the new include in item.h too

Yeah, I actually realized that myself right after making that comment and I went through and added appropriate NOLINTs.

Good point on the include, it looks like I can replace it with a forward declaration just fine. I'll double check the other includes I added, too.

src/inventory_ui.cpp
```diff
@@ -1386,6 +1386,7 @@ void inventory_column::collate()
             if( e->is_item() && e->get_category_ptr() == outer->get_category_ptr() &&
                 e->any_item()->is_favorite == outer->any_item()->is_favorite &&
                 e->any_item()->typeId() == outer->any_item()->typeId() &&
+                !e->any_item()->contents_linked && !outer->any_item()->contents_linked &&
```
andrei8l, line 1389: Why? This seems like a good use case for collation.

<details>
<summary>With the current exclusion:</summary>

with

Imagine having 10 of them
</details>


<details>
<summary>Without it:</summary>

Screenshot from 2023-05-05 06-43-58

Screenshot from 2023-05-05 06-44-01


</details>

Kamayana: I intended to keep them separated in two separate collated groups so you could reliably do one or the other, but I didn't actually intend to make all plugged in devices completely separate like that. It's obviously overly strict, now that I'm looking back at it. Replacing the line with `e->any_item()->contents_linked == outer->any_item()->contents_linked &&` makes it work as I originally intended. I need to make it work like that for cable items like extension cords, too.

Also, the battery number showing the entire connected power supply is also unintentional, it should just be showing the device's battery so you can tell how charged it is. I'll take a look at that.

tools/spell_checker/dictionary.txt
```diff
@@ -2985,7 +2985,7 @@ unlivable
 unlively
 unlives
 unlooted
-unlubricated
```
Fris0uman, line 2988: Why remove this word?

Kamayana: Unintended, might've been a rebase issue. Just re-added it.

PR #64341, opened 2023-03-17
boiled makeshift bandage cost

anoobindisguise:
#### Summary
Bugfixes "Correct cost of boiled makeshift bandage"

#### Purpose of change
You have 3 units of water, to clean each it should be 1u of water_boiling_heat per. Adding in the bandages rounds up to 4u. The time to sterilize the bandages shouldn't be any higher than the time needed to sterilize the water (1 minute at boiling temp)

#### Describe the solution
4u water_boiling_heat. Also make 67% of the water return as a clean water byproduct.

No byproduct.


Maleclypse (approved):
Reasoning looks solid.

adamkad1:
Yeah, this seems fine, aside from the byproduc

Issue #60336, opened 2022-08-20
Epicuticle causes too much hand encumbrance to staunch blood loss.


rty275:
### Describe the bug

The epicuticle item causes too much encumbrance on your hands to make bare-handed attempts to put pressure on a bleeding wound. Given that it is "a thin, flexible layer of chitin", I'm unsure if this is intended or not. If it's purposeful, then that's good to know and definitely unexpected, but I'm not an entomologist, I wouldn't know one way or the other really.

### Steps to reproduce

1. Acquire Epicuticle mutation and remove all other hand coverings.
2. Receive bleeding wound.
3. Wait a turn with hands empty, receive "Your hands are too encumbered to effectively put pressure on the bleeding wound!" message.

### Expected behavior

I would expect to be able to put pressure on wounds, but I fully admit I am not an entomologist, and what's more, I'm not totally sure there *is* any information available about putting pressure on wounds in chitin. I would think based on the "thin, flexible" part, that it would be receptive to pressure, or perhaps a hand on either side of the breach in chitin holding it closed, something along those lines. 

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19043.1889 (21H1)
- Game Version: 5235d18 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda]
]

### Additional context

I generated a test world free of the admittedly pretty extensive list of mods the world I discovered this on had, and reproduced it successfully.

fungamer2-2:
Seems like a duplicate of #60100

Issue #60100, opened 2022-08-12
epicuticle causes hands to be "too encumbered" to apply pressure to bleeding wounds


yam1917:
### Describe the bug

epicuticle causes hands to be "too encumbered" to apply pressure to bleeding wounds

### Steps to reproduce

just spawn a character with epicuticle trait, get scratched by a zombie or something and then wait while bleeding (yes, i did make sure my hands are free)

### Expected behavior

pressure should be applied to bleeding wound

### Versions and configuration

-windows 10
- version: x64-2022-08-01-0506

hexagonrecursion:
/confirmed

- OS: Linux
    - OS Version: LSB Version: :core-4.1-amd64:core-4.1-noarch; Distributor ID: Fedora; Description: Fedora release 36 (Thirty Six); Release: 36; Codename: ThirtySix; 
- Game Version: 13fbcc2 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

yam1917:
i also think that epicuticle negates spines mutation, but im not sure if i was just really unlucky

PR #64344, opened 2023-03-17
Increase epicuticle encumbrance, lower breathability, add mouth coverage

anoobindisguise:
#### Summary
Balance "Epicuticle has worse encumbrance, defense and breathability"

#### Purpose of change
Due to #60100 and #60336 Epicuticle was buffed to have only 1 encumbrance in #60880 which made it basically cheat-level strong. However due to changes in #61726 that change is no longer necessary.

#### Describe the solution
Worked with @Venera3 a bit on this.
Bash defense of Epicuticle material reduced from 3 to 2. It gives it a defense of 5/5/2.5. Breathability is reduced to AVERAGE, as bugs aren't known for sweating and neither should bug skin be, but the primary reduction in sweat will come from the actual hard exoskeleton itself so your underlying skin still gets to breath a little.
Encumbrance raised to 2.
Added coverage to the mouth, at 85%. Encumbrance is only 1 there instead of 2 since it shouldn't affect your oxygen intake too bad (though if your parts are stiffer it will be very slightly more difficult to draw breaths)

Plenty,
~~- Make it 1 encumbrance still, but less thickness (1-2mm)~~
~~- Make it less thickness on hands and less encumbrance.~~
~~- Make it cover the mouth.~~


Maleclypse:
I think that creatures that have epicuticle don't sweat naturally so it might be even worse than that?

PR #64361, opened 2023-03-18
Make instability test more reliable

actually-a-cat:
#### Summary
None

#### Purpose of change
The instability test goes to 10,000, which doesn't make sense, because instability itself is capped at 8,000. This causes occasional test failures. The bad mutation vs instability curve at that point is very flat (the difference between 8k and theoretical 10k is within the test margin of error) which is why it only fails occasionally, but removing this should reduce spurious test failures.

#### Describe the solution
Only test up to 8k instability.

There is probably an altogether smarter way to do this test, but I'm not the person to figure that out.

#### Testing
The test works. Full disclosure, I'm not doing the trials needed to verify the effect on failure rates.

### DIFF
--- a/tests/mutation_test.cpp
+++ b/tests/mutation_test.cpp
@@ -581,8 +581,7 @@ TEST_CASE( "Chance of bad mutations vs instability", "[mutations][instability]" 
         {4000, 0.598f},
         {5000, 0.649f},
         {6000, 0.686f},
-        {8000, 0.737f},
-        {10000, 0.770f}
+        {8000, 0.737f}
     };
 
     const int tries = 1000;


src/character.h
```diff
@@ -1371,15 +1371,16 @@ class Character : public Creature, public visitable
         void mod_cost_timer( const trait_id &mut, time_duration mod );
 
         /** Picks a random valid mutation and gives it to the Character, possibly removing/changing others along the way */
-        void mutate( const int &true_random_chance, bool use_vitamins );
+        void mutate( const int &true_random_chance, bool use_vitamins, double ( *rng )( double,
```
anothersimulacrum, line 1374: Please don't call this rng, you're shadowing the rng function in a way that's pretty opaque in ~200 line function.
```suggestion
        void mutate( const int &true_random_chance, bool use_vitamins, double ( *rng_fn )( double,
```
Also, you should be able to use a std::function here instead of a function pointer.

actually-a-cat: Thank you for your input but I decided to undo that attempt, I don't think I want to mess with things like that. I reverted this PR back to just fixing the current test instead of replacing it.

PR #64362, opened 2023-03-18
Migrate to std::optional (part 1)

jbytheway:
#### Summary
Infrastructure "Migrate to std::optional"

#### Purpose of change
Now that we're on C++17 we don't need to maintain our own implementation of `cata::optional`; we can use `std::optional` instead.

Besides the code reduction, `std::optional` has small additional features.  For example, if `T` is trivially copyable, then `std::optional<T>` is too.  This can improve the performance of some operations.

#### Describe the solution
Make `cata::optional` an alias of `std::optional`.  Similarly for `nullopt` and `in_place`.

Fix or work around the resulting warnings and errors.

The biggest workaround required was making `achievement::time_bound` not be defined inside `achievement` any more.  This couldn't be emplaced into an optional for some reason; I think it's a GCC 12 compiler bug; I might dig into it further but this workaround works for now.

I could have tried to replace all uses of `cata::optional`, but that would be a much more invasive change prone to merge conflicts.  With this approach we can gradually migrate `cata::optional` to `std::optional` over time.

I also thought about removing the tests for `cata::optional`, but maybe we should keep them, at least until we're confident all our target platforms have a good `std::optional` implementation.

#### Testing
Unit tests and clang-tidy.

#### Additional context


andrei8l:
* #64333

jbytheway:
Superceded by #64333.

PR #64364, opened 2023-03-18
Set OSX_MIN on CI to match release workflow

jbytheway:
#### Summary
Infrastructure "Set OSX_MIN on CI to match release workflow"

#### Purpose of change
To avoid bugs appearing in release builds that were not detected on CI.

See for example #64314 which was such an issue.

#### Describe the solution
Update the CI build script to use the same OSX_MIN value as release builds.

This is my best guess as to the source of the discrepancy; they both build on version 10.15 build hosts.

#### Testing
Will need to be tested on CI.

#### Additional context

### DIFF
--- a/build-scripts/gha_compile_only.sh
+++ b/build-scripts/gha_compile_only.sh
@@ -62,7 +62,10 @@ then
         ..
     make -j$num_jobs
 else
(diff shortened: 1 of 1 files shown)


BrettDong:
I also see this cannot find passenger error in macOS and MinGW+wine tests in a few other pull requests, but I failed to reproduce the error locally.

PR #64367, opened 2023-03-18
Deduplicate code for drawing labeled bars of chars

sparr:
#### Summary
None

#### Purpose of change
Existing code for drawing bars like `Foo: [***--.   ]`, with or without color, is repetitive

#### Describe the solution
Move repetitive code out to helper functions

I started with less templating but didn't like the results. I did consider defining the printer functions rather than using lambdas and could do that instead if requested.

#### Testing
Played around with ranged fire interface in different situations


kevingranade:
This has several clang tidy warnings.

PR #64372, opened 2023-03-18
Menstrual products

Aerin-of-the-Toast:
#### Summary
Content "Menstrual products"

#### Purpose of change

Fluff. To expand out bathroom loot. So the AFAB population can stop implicitly suffering.

#### Describe the solution

Added in tampons (with snippets), menstrual pads (with snippets), and rare menstrual cups. Added common spawns to house bathrooms as well as slapped them in some bugout bags and the "drugs_heal_simple" item group so they'll spawn wherever that can spawn as well (such as pharmacies).

No, there are no menstruation *mechanics.* Pragmatically the only thing these are useful for are disassembly into cotton balls or trade.

Not having any snippets and just have a single generic description for each item, but if toilet paper can have snippets so can tampons. 

#### Testing

Made sure the game ran and checked out some houses to make sure they were spawning correctly.

#### Additional context

I am not AFAB so if there are any gross inaccuracies feel free to yell at me to fix them.

Before anyone brings up tampons being designed to plug bullet holes or something: they weren't, and I couldn't find reliable sources on whether they'd even be a good idea or a really bad one. I was advised to avoid giving them any bandaging/clotting properties and just letting them disassemble into cotton balls instead.

Lastly, please be mature about this.


data/json/itemgroups/stashes.json
```diff
@@ -266,7 +266,13 @@
     "id": "bugout_bad_pack",
     "subtype": "distribution",
     "//": "an item that wouldn't be super useful in a survival situation, recreational items and such.",
-    "entries": [ { "group": "SUS_book", "prob": 50 }, { "group": "games", "prob": 20 }, { "group": "bugout_memento", "prob": 50 } ]
+    "entries": [
+      { "group": "SUS_book", "prob": 50 },
+      { "group": "games", "prob": 20 },
+      { "group": "bugout_memento", "prob": 50 },
+      { "group": "tampon_box_used", "prob": 20 },
+      { "group": "menstrual_pad_box_used", "prob": 20 }
```
anothersimulacrum, line 274: I think it would more sense to grab a full box instead of a used box for a bug-out bag. I might have a distribution between full and used here.

Aerin-of-the-Toast: thanks, good catch

TheShadowFerret:
>Before anyone brings up tampons being designed to plug bullet holes or something: they weren't, and I couldn't find reliable sources on whether they'd even be a good idea or a really bad one.

https://www.youtube.com/watch?v=Brfeu2f-LFY

The answer is no 😛

Aerin-of-the-Toast:
> >Before anyone brings up tampons being designed to plug bullet holes or something: they weren't, and I couldn't find reliable sources on whether they'd even be **a good idea or a really bad one**.
> 
> https://www.youtube.com/watch?v=Brfeu2f-LFY (w/ live demo)
> https://www.crisis-medicine.com/heavy-flow-is-not-massive-hemorrhage-tampons-dont-belong-in-ifaks/
> https://havokjournal.com/fitness/medical/your-tactical-tampon-is-useless-for-life-threatening-hemorrhage/
> 
> The answer is no, they're too small to plug the wound cavity effectively nor have the absorption capacity for it(& absorbing blood so it doesn't stain the floor does nothing for living :p).

Guess making them unable to be used as bandages entirely was the right call then, thanks for the sources!

Raikiri:
I guess we don't have personal hygiene items such as toilet paper specifically because the point is to abstract their existence out of the game as irrelevant. If there was, say, toilet paper, one would expect to have a system where it's needed.

KittyTac:
> I guess we don't have personal hygiene items such as toilet paper specifically because the point is to abstract their existence out of the game as irrelevant. If there was, say, toilet paper, one would expect to have a system where it's needed.

We do have toilet paper though.

PR #64373, opened 2023-03-18
Update professions.json add a profession.

MdeeDev:
#### Summary
Content "Add electronics engineer profession"

#### Purpose of change
To be able to play the game as an electronics engineer.


kevingranade:
Who's carrying around electronic prototyping gear?
Skills,  proficiencies,  maybe a pocket reference or even better some ebooks on the phone,  but the electronics gear doesn't make sense.

MdeeDev:
If a lab technician can walk around with safety glasses, a lab coat and full chemistry set. Why not?
Also an electrician can carry around a bunch of stuff: Voltmeter, multimeter, solder (200), a soldering iron.

kevingranade:
> If a lab technician can walk around with safety glasses, a lab coat and full chemistry set.

Really? I'll track that down then.

PR #64380, opened 2023-03-19
Replace min(max()) and clamp() with std::clamp()

sparr:
#### Summary
None

#### Purpose of change
Our homegrown `clamp()` is obsoleted by C++17's `std::clamp`, and our use of variations on `std::min(a,std::max(b,c))` is relatively difficult to understand.

#### Describe the solution
Replaced all instances of `clamp(a,b,c)` (distinct from the unrelated two-parameter version used for 2d points) with `std::clamp(a,b,c)`. Replaced most instances of nested `min`/`max` with equivalent `std::clamp`. Some other closely related changes, and some minor related refactors.

Leaving the code as-is would reduce churn.

#### Testing
Ran test suite. Playing the game with it as long as this is a draft. Worked around a crash bug described in #64381.

### DIFF
--- a/src/vehicle_move.cpp
+++ b/src/vehicle_move.cpp
@@ -192,8 +192,8 @@ void vehicle::smart_controller_handle_turn( bool thrusting,
                                 max_battery_level;
 
     // ensure sane values
-    cfg.battery_hi = clamp( cfg.battery_hi, 0, 100 );
-    cfg.battery_lo = clamp( cfg.battery_lo, 0, cfg.battery_hi );
+    cfg.battery_hi = std::clamp( cfg.battery_hi, 0, 100 );
+    cfg.battery_lo = std::clamp( cfg.battery_lo, 0, cfg.battery_hi );
 
     // when battery > 90%, discharge is allowed
     // otherwise trying to charge battery to 90% within 30 minutes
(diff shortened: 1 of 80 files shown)


src/vehicle_move.cpp
```diff
@@ -586,9 +586,9 @@ void vehicle::thrust( int thd, int z )
         const int min_vel = max_reverse_velocity();
         if( vel_inc > 0 ) {
             // Don't allow braking by accelerating (could happen with damaged engines)
-            velocity = std::max( velocity, std::min( velocity + vel_inc, max_vel ) );
+            velocity = std::min( velocity + vel_inc, max_vel );
         } else {
-            velocity = std::min( velocity, std::max( velocity + vel_inc, min_vel ) );
+            velocity = std::max( velocity + vel_inc, min_vel );
```
mqrause, line 591: This possibly changed behavior and might be the cause for the test failures.

sparr: Good catch. I'm working my way through all the cases, and discovering some places where the existing clamp() or min(max()) use cases have the max lower than the min with unexpected results. I will be either correcting my change or fixing the scenario that leads to those outcomes.

PR #64384, opened 2023-03-19
(Don't merge) Revert "Convert anonymous vehicle_part enum to class enum (#64331)"

irwiss:
#### Summary
Don't merge (and making this comment not fit template so validator step fails)

#### Purpose of change

Trigger CI to probe if commit d31ab2d789559c31f794aa36ba98fd39018038cf is causing the board_vehicle test to fail

#### Describe the solution

Test seems to only fail in the clang12/macos/ubsan config, so I'm trying to test if that commit put in any UBs even though all information found indicates static_cast class enum with fixed underlying type  to it's underlying type and other way around is non-UB

#### Testing

And now we wait...

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -1220,13 +1220,13 @@ void map::board_vehicle( const tripoint &pos, Character *p )
         }
         return;
     }
-    if( vp->part().has_flag( vp_flag::passenger_flag ) ) {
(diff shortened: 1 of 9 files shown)


NetSysFire:
You could also make your PR a draft. Drafts can never be merged.

irwiss:
Drafts don't run the UBSan test unfortunately, they skip most of the matrix except for basic build and test

PR #64386, opened 2023-03-19
Jsonify damage types

dseguin:
#### Summary
None

#### Purpose of change
Allow new damage types to be created in JSON so mods can define new types of spell damage, armor resistances, mutation/bionic armor types, etc.

#### Describe the solution
Generalize most of the damage type handling so that new types can be handled automatically. This required some changes to some JSON types (`"bionic"`, `"material"`, `"MONSTER"`, and all item types) to allow them to define armor and resistances for new damage types.

The new damage type definition is basically:
```JSON
{
  "id": "stab",
  "type": "damage_type",
  "//": "true means this damage type is associated with melee attacks",
  "melee_only": true,
  "//": "represents damage that can be dealt by tangible means, including ranged damage"
  "physical": true,
  "//": "determines if this damage type comes from an edged surface, also checked by the CVD machine for diamond coating weapons",
  "edged": true,
  "//": "whether this damage type can be inflicted by the environment",
  "environmental": false,
  "//": "color of spells associated with this damage type",
  "magic_color": "light_red",
  "//": "translation-friendly damage type name",
  "name": "stab",
  "//": "the skill that dealing this damage type trains",
  "skill": "stabbing",
  "//": "the damage type that this type is based on, and relative effectiveness. (used for monster defs)",
  "derived_from": [ "cut", 0.8 ],
  "//": "if the character or monster has one of these flags, they fully resist damage of this type",
  "immune_flags": { "character": [ "STAB_IMMUNE" ], "monster": [ "STABPROOF" ] },
  "//": "true for damage types that cannot be defended against",
  "no_resist": false,
  "//": "list of EOCs that activate when a creature is hit with this damage type",
  "onhit_eocs": [ "EOC_STAB_ONHIT" ]
}
```

TODO:
- [x] Convert damage type infrastructure to be (mostly) data-driven, and genericize most damage handling logic
- [x] Resolve any test errors
- [x] Add more ways to create JSON-defined effects for certain damage types (EOCs :eyes:)
- [x] Documentation
- [x] Testing
  - [x] Melee unit tests
  - [x] Ranged unit tests
  - [x] Extensive ingame testing

##### Other notable changes
- Updated the format for bionic protection values to a generic "protec" field that can accept any damage type
- Updated the format for monster armor values to a generic "armor" field that can accept any damage type
- Derived values for monster protection (i.e.: "stab" resistance, if it's not defined) were originally floored to the nearest integer, but now the derived value is calculated and stored as a float
- Updated the format for item melee damage values to a generic "melee_damage" field that can accept any damage type
- Obsoleted the `STAB` flag, since piercing damage can just be defined in the "melee_damage" field using `"stab": 4`
- Item info protection values are generic, and the previously hardcoded damage types might display in a different order, or new types might show up if they're used in the armour definition
- Added a new `damage_info_order` object type that defines how damage types are represented in the item info UI (order and visibility)

#### Testing
Added a couple test cases that cover most bits of the changes: weapon damage, clothing damage absorption, monster & character damage resistance, ammo type damage, damage type EOCs, etc.

```
tests/cata_test "Damage type effectiveness vs. monster resistance,shot_custom_damage_type"
```

There will be bugs, but this should catch most of them

#### Additional context
There be dragons over in the "files changed" tab


GuardianDll:
How it shouls work with enchantments? For example, ARMOR_ACID, EXTRA_ACID (maybe it's better to get rid out of it tho), ITEM_DAMAGE_ACID and ITEM_ARMOR_ACID?

dseguin:
> How it shouls work with enchantments? For example, ARMOR_ACID, EXTRA_ACID (maybe it's better to get rid out of it tho), ITEM_DAMAGE_ACID and ITEM_ARMOR_ACID?

Those will have to be generalized separately as a Part 2. This is just for genericizing damage types directly. Also, adding anything else to this project might push it over the "10 000 lines changed" mark :p

src/melee.cpp
```diff
@@ -808,10 +818,16 @@ bool Character::melee_attack_abstract( Creature &t, bool allow_special,
...
+
+            bool has_edged_damage = false;
+            for( const damage_type &dt : damage_type::get_all() ) {
+                if( dt.melee_only && dt.edged && dealt_special_dam.type_damage( dt.id ) > 0 ) {
+                    has_edged_damage = true;
+                    break;
+                }
+            }
+
+            if( has_edged_damage || ( !cur_weapon && has_edged_damage ) ) {
```
ZhilkinSerg, line 830: This looks like there is a logical error in here - you check for `has_edged_damage` twice, so that check could be simply  omitted.

PR #64388, opened 2023-03-19
New arachnid exclusion mod

DoctorBoomstick:
#### Summary
Mods "Creates a new arachnophobia-friendly mod for CDDA, blacklisting spiders and spider related locations."
#### Purpose of change
If there’s one thing Cataclysm is stuffed with, apart from guns, it's various insects and arachnids. With respect to our spidery friends in particular, I’ve seen quite a few folks with arachnophobia being triggered upon encountering one of the many in-game spiders, with a given tile set’s respective depictions generally dictating the severity. While a simple ASCII S is, in my view, probably not going to trigger most arachnophobes, I figured that playing a game only to be unexpectedly booted by one’s phobia isn’t nice. As such, when somebody suggested that a simple arachnophobia-friendly mod could be created and there was some interest in the thought, I figured that it would be appropriate to pick up the idea.
#### Describe the solution
I’ve put together a very simple mod that blacklists spider enemies from the base game, as well as removing spider-related locations and map extras. The mod also removes the demon spider layer from Magiclysm.
When it came to deciding which enemies deserved the boot, I followed the notion that anything spider-related ought to be axed. Not being an arachnophobe myself, I don’t know if something like the arthropod mutant and spideer would prompt a reaction, but I figured that removal was the better part of thoroughness in this particular case. Plus, if they don’t in fact prompt a trigger, their reintroduction is a simple matter of removing a single line.
1.	Possibly removing spider mutagens, however, as pointed out by several people when workshopping the mod, the mutations themselves are versatile and non-intrusive enough to merit their staying. Plus, I don’t suppose that an arachnophobe would inject themselves with something explicitly meant to turn them into a spider.
2.	Also possibly removing some of the spider-like zombie variants and amalgamations, though I’m doubtful of the notion that they would prompt arachnophobia and felt that removing them would possibly effect the monster balance of certain locations.
#### Testing
I loaded up a fresh game with the mod installed and observed that no errors were generated before boarding a Humvee and driving around for a mighty long time and seeing that no spiders manifested themselves.

### DIFF
--- a/data/mods/No_Spiders/modinfo.json
+++ b/data/mods/No_Spiders/modinfo.json
@@ -1,0 +1,12 @@
+[
+  {
+    "type": "MOD_INFO",
+    "id": "No_spiders",
+    "name": "No Arachnid Monsters",
+    "authors": [ "DoctorBoomstick" ],
+    "maintainers": [ "DoctorBoomstick" ],
+    "description": "makes the game arachnophobe-friendly by taking all arachnid monsters and related locations from the base game, and nuking them from orbit.",
+    "category": "monster_exclude",
+    "dependencies": [ "dda" ]
+  }
+]
--- a/data/mods/No_Spiders/modinfo.json
+++ b/data/mods/No_Spiders/modinfo.json
@@ -5,7 +5,7 @@
     "name": "No Arachnid Monsters",
     "authors": [ "DoctorBoomstick" ],
     "maintainers": [ "DoctorBoomstick" ],
-    "description": "makes the game arachnophobe-friendly by taking all arachnid monsters and related locations from the base game, and nuking them from orbit.",
+    "description": "Makes the game arachnophobe-friendly by taking all arachnid monsters and related locations from the base game, and nuking them from orbit.",
     "category": "monster_exclude",
     "dependencies": [ "dda" ]
   }
--- a/data/mods/No_Spiders/blacklists/location_blacklist.json
+++ b/data/mods/No_Spiders/blacklists/location_blacklist.json
@@ -1,0 +1,14 @@
+[
+  {
+    "type": "overmap_special",
+    "id": "Spider Pit",
+    "overmaps": [  ],
+    "occurrences": [ 0, 0 ]
+  },
+  {
+    "type": "overmap_special",
+    "id": "demon_spider_lair",
+    "overmaps": [  ],
+    "occurrences": [ 0, 0 ]
+  }
+]
--- a/data/mods/No_Spiders/blacklists/map_extra_blacklist.json
+++ b/data/mods/No_Spiders/blacklists/map_extra_blacklist.json
@@ -1,0 +1,14 @@
+[
+  {
+    "type": "region_overlay",
+    "regions": [ "all" ],
+    "map_extras": {
+      "forest": { "extras": { "mx_spider": 0 } },
+      "forest_thick": { "extras": { "mx_spider": 0 } },
+      "forest_water": { "extras": { "mx_spider": 0 } },
+      "field": { "extras": { "mx_Trapdoor_spider_den": 0 } },
+      "agricultural": { "extras": { "mx_Trapdoor_spider_den": 0 } },
+      "build": { "extras": { "mx_house_spider": 0 } }
+    }
+  }
+]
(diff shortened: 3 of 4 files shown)


data/mods/No_Spiders/modinfo.json
```diff
@@ -0,0 +1,12 @@
+[
+  {
+    "type": "MOD_INFO",
+    "id": "No_spiders",
+    "name": "No Arachnid Monsters",
+    "authors": [ "DoctorBoomstick" ],
+    "maintainers": [ "DoctorBoomstick" ],
+    "description": "makes the game arachnophobe-friendly by taking all arachnid monsters and related locations from the base game, and nuking them from orbit.",
```
GuardianDll, line 8:
```suggestion
    "description": "Makes the game arachnophobe-friendly by taking all arachnid monsters and related locations from the base game, and nuking them from orbit.",
```

Venera3:
CRT (or Xedra, can't remember) has a blood spider virus thing as well iirc.

DoctorBoomstick:
> CRT (or Xedra, can't remember) has a blood spider virus thing as well iirc.

Unfortunately, I can’t blacklist those monsters as, if I blacklisted them and somebody used this mod without having Xedra also in the mod list, it throws up an error. This is why I didn’t blacklist the Magiclysm enemies. Locations on the other hand don’t throw up any issues, so I could blacklist them if I axed the places they spawn in, though I suspect that would mess with other content in Xedra apart for strictly the spiders.

anothersimulacrum (commented):
I think a better approach would be to turn these in to some other form of monster. I know there's no infrastructure for that right now.

DoctorBoomstick:
> I think a better approach would be to turn these in to some other form of monster. I know there's no infrastructure for that right now.

I believe one could do that simply enough with migrations, and I have thought over the possibility as I do like it better than just removing these monsters, but my question is: what could they be replaced with that wouldn't or shouldn’t just go into the base game anyway? It's either that or perhaps replacing them with another, already existing insect type, like cockroaches, but I don’t know how that would effect the areas these creatures spawn in. I’d be glad to replace them with something; I just don’t know what.

anothersimulacrum:
Monster migrations don't exist.

We can make up less phobic monsters with similar set-pieces as necessary, and have this mod just turn the spiders into those set pieces, I agree.

DoctorBoomstick:
I see.      It's unfortunate.      As things currently stand, I see three possibilities:

1. Creating some form of lese-phobic enemies, the details of which are still vague enough to escape me at present, but perhaps some form of nether insectoid, and including them into the spawn groups in place of these axed spider monsters (which is quite clunky and doesn’t serve to counter much, to be honest),
2. Leaving these monsters excluded, but reworking their map extras to have other insect set pieces
3. shutting down this PR until such a time when monster migrations might get added.

The issue I’ve got with the first two ideas is that replacing these spiders with anything original would bring into question the notion of: shouldn’t the original idea and extra content be included into the base game anyhow, rather than as a replacement for other content? To sidestep this, anything new would have to be out of line with the base idea of the game (a stupid example: replacing all spiders with Xenomorphs, for example), which at that point just becomes rather nonsensical and beyond the scope of what this PR is trying to achieve, in my opinion. I stand to be corrected.

anothersimulacrum:
The replacement could be in the base game too.

GuardianDll:
Looks_like is ignored, if tileset has a texture for a creature

Fris0uman:
Exclusion mods don't fit the criteria to be first party mods https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/IN_REPO_MODS.md 
You should publish it and share it as a third party thing.

I-am-Erk:
@DoctorBoomstick , this has come up in discussion and I'm not sure it should have been closed. Do you have an interest in helping to bring it back and maybe modify a few details?

DoctorBoomstick:
> @DoctorBoomstick , this has come up in discussion and I'm not sure it should have been closed. Do you have an interest in helping to bring it back and maybe modify a few details?

If there’s an interest in bringing the PR back and including it as a workable mod, while I can’t swear that I’ll get around to it in the near future (life’s busy and I’ve had to sideline my other WIP PRs), I’d be fully willing to pick this back up. I just need to read up on the discussion and any tweaks that should be made.

PR #64390, opened 2023-03-19
Finish the port of Crazy Cata monsters to Crazy Cata

Karol1223:
#### Summary
None

#### Purpose of change
During the initial move of these guys I didn't realise they had hardcoded interactions with the map. Which means people could extremely rarely get an error message about a non-defined monstergroup

#### Additional context
All of jsonifying was done by @GuardianDll, I simply removed the c++ leftovers

### DIFF
--- a/data/json/overmap/map_extras.json
+++ b/data/json/overmap/map_extras.json
@@ -254,17 +254,6 @@
     "min_max_zlevel": [ -2, 0 ]
   },
   {
-    "id": "mx_shia",
-    "type": "map_extra",
-    "name": { "str": "Shia LaBeouf" },
-    "description": "Cannibal is nearby.",
-    "generator": { "generator_method": "map_extra_function", "generator_id": "mx_shia" },
-    "min_max_zlevel": [ 0, 0 ],
-    "sym": "c",
-    "color": "red",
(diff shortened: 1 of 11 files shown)


mqrause:
Ideal solution would be to unhardcode this stuff, then you can also make it crazy cata exclusive. Other than that, `place_spawns` is for monstergroups only. For single monsters there is `add_spawn`, but that's supposed to be removed as per #3951, which is why those groups were created in the first place. So apart from unhardcoding the spawns, possibly empty monstergroups are an option, if that's allowed. And then override them in crazy cata.

Karol1223:
> Ideal solution would be to unhardcode this stuff, then you can also make it crazy cata exclusive. Other than that, `place_spawns` is for monstergroups only. For single monsters there is `add_spawn`, but that's supposed to be removed as per #3951, which is why those groups were created in the first place. So apart from unhardcoding the spawns, possibly empty monstergroups are an option, if that's allowed. And then override them in crazy cata.

I think I'll have to go about unhardcoding this, yeah. No reason to have code for this in basegame, even less so if it's c++ code. But I'm not looking forward to this since I've not touched specials before.

Issue #64302, opened 2023-03-16
Obsoletions are not in fact obsolete


irwiss:
### Describe the bug

Tons of ids currently sitting in obsoletion files are not in fact obsolete:
`data/json/obsoletion/**`
`data/json/obsolete.json`
`data/json/items/obsolete.json`

Lots of ids also rely on `data/json/items/migrations.json` file to move them to correct id (they're specified as old id in actual data files )

Many "obsolete" ids are still used in game files, like loose necropolis ids, are still used in overmap data

New PRs ( after 0.G ) already started adding new entries to those files, I expect postponing this to disentangle later will become a huge mess of old obsoletions and new mixed together

### Attach save file

N/A

### Steps to reproduce

Delete the above mentioned files and folders, the game will spew debugmsgs for a ton of invalid data and finally give up when it starts hitting missing overmap ids

### Expected behavior

Obsoletion files should be clean from pre-0.H entries

### Versions and configuration

N/A

irwiss:
This is practically done so I think it's okay to close, only 2 files 0.G remain and those mostly relate to steel quality changes

PR #64399, opened 2023-03-19
Remove more 0.G migrations

irwiss:
#### Summary
None

#### Purpose of change

Cull more stuff to clear https://github.com/CleverRaven/Cataclysm-DDA/issues/64302

#### Describe the solution

Stuff not resolved completely:

Untouchables: t_centrifuge, t_machinery_light etc... MUST sit in `data/json/obsoletion.json`, if you so much as move the file to obsoletion/ folder then the game will fail on loading, these are still widely used, probably [load order is messing up somehow](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/json/LOADING_ORDER.md).

Overmap stuff is complicated to remove; -snip- turns out a bad PR was messing with this, but it can be done in a later PR

Some stuff like obsolete monsters still sit in `data/json/obsoletion/obsolete_but_used_0_G_monsters.json`, removing them causes older games fail on load; when overmap and achievement try to resolve monster types, seems we have no migration path for those without raising debugmsgs

Survivor armor in `data\json\recipes\armor\bespoke_armor` also remains still

Stuff done:

Removes 0.G item/recipe migrations
Replaces a bunch of ids still used in game with their migrated id
Consolidates remaining 0.G obsolete entries under `data/json/obsoletion/*` (except for data/json/obsolete.json, see above)
Moves (or rather - copies for now) monsters like zombie burner to Xedra and No Hope (they're used there)

#### Testing

Loaded worlds with mods Aftershock, Xedra, No_Hope, Generic Guns without errors
Loaded a couple older games - there were no errors

Hopefully tests catch stray ids I missed


Maleclypse (commented):
Aftershock changes make sense

Brambor:
Why not mark this PR as a draft instead of naming it _(don't merge yet)_?

irwiss:
Drafts don't run all tests

Issue #64368, opened 2023-03-18
[Magiclysm] Mage blade and Mage Armor don't work


Terrorforge:
### Describe the bug

The scaling damage/armor on Mage Blade and Mage Armor isn't applied. Mage Blade only does it's base 12 damage and Mage Armor provides no protection from physical attacks whatsoever.

### Attach save file

Trivially reproducible

### Steps to reproduce

- Give self spells at max level
- Cast both
- Fight zombie

### Expected behavior

The enchantment is applied and the items work as described.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.2006 (22H2)
- Game Version: 53f9146 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Magiclysm [magiclysm],
    Mythical Martial Arts [MMA]
]

### Additional context

At first I thought the Mage Armor thing was because the sign was inverted (ARMOR_[BLANK] enchantments should be negative to reduce damage), but that should _increase_ damage rather than do nothing, and since the blade isn't working either I assume something more fundamental is broken.

GuardianDll:
\confirmed
![image](https://user-images.githubusercontent.com/67688115/226125786-a611ec2f-7ae5-4a3d-8275-75cd9b09c66e.png)

also heck i forgot i need negative value for armor, need to redo this part

SittingDucken:
Believe mage armor needs the relic_data k, v. As already mentioned the * const:-1 for damage reduction. And potentially items duplicated to account for mage armor/blade +.

GuardianDll:
Oh, that's right
i wonder why the game silenty ignored the fact there is no relic_data

SittingDucken:
Bet without digging.. well formed json is passing through the json check. Since it's a proper field, but doesn't apply to an object, there's no reason to fail when loading the object. Would need a check for dependency, or null application to validate.

GuardianDll:
The reason it silently fails is in #64438

SittingDucken:
Welp there it is 🗡️ 
Question then for these items not working. Would we see better functionality if we transfer the relic data to an enchantment similar to how Gaia's Chosen terra armor is working?
[...]

PR #62703, opened 2022-12-11
[Magiclysm] Lot of different stuff

GuardianDll:
#### Summary
None

#### Describe the solution

Sane some armor values of magic items, so they wont be on the same level as anti-tank armor
Add Gaia's Chosen spells - Terra Armor, that give [3*terra armor spell level] defense to all damage, and 0 lvl Dust Reborn spell, that create waypoint and ressurech character at this point afterwards
Replace old Frost Armor with new - almost no defence, but pretty nice damage
Sane the Heft effect to be not so crazy busted. Still good tho
Change the thunder pillar line attack to cone, so it can attack more enemies at once. Pretty busted, but will be fixed eventually with #62676
Move Soulfire EoC to enchantment folder, where all another EoC belongs
Tweaked the Earth Elemental attunement - now his spells do not require hands, can be casted faster, also the attunement got 3x  stamina and stamina regen boost, as his spells uses stamina intsead of a mana. Also earthquake now ignore walls and deal bash damage around, so be aware of your car nearby!
Added throwing knifes to knifeshot spell
Added NO_HANDS flag to knifeshot, because the fact you need to "throw" the knife with bare hands is stupid
Replaced the effect of holy armor and holy sword to new, conditional powered one
Added a cool integrated ethereal pocket for alchemist, that can store 10L or 20kg of mass for free

#### Testing

Everything works as intended, but im too lazy to describe each test

#### Additional context

(well, its Quake 35 lvl, you can't obtain such high level in the game currently, but it looks dope as for me)
image
image

PR #64405, opened 2023-03-20
[Magiclysm] Small fix

GuardianDll:
#### Summary
None
#### Purpose of change
Fixing small bug i found
#### Describe the solution
add relic_data to mage armor, boost the lumination effect, remove the intelligence dependence
#### Additional context
It doesn't fix the #64368 - mage blade, for example, has relic_data already, but test shows it doesn't work still

### DIFF
--- a/data/mods/Magiclysm/items/ethereal_items.json
+++ b/data/mods/Magiclysm/items/ethereal_items.json
@@ -885,7 +885,7 @@
     "category": "armor",
     "copy-from": "armor_lightplate",
     "name": { "str_sp": "mage armor" },
-    "description": "A summoned magic armor.  It glows shiny and get sturdier with each spell level.",
+    "description": "A summoned magic armor.  It gets shinier and sturdier with each spell level.",
     "material": [ "concentrated_mana" ],
     "warmth": 0,
     "material_thickness": 0,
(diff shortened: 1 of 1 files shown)


data/mods/Magiclysm/items/ethereal_items.json
```diff
@@ -885,39 +885,41 @@
     "category": "armor",
     "copy-from": "armor_lightplate",
     "name": { "str_sp": "mage armor" },
-    "description": "A summoned magic armor.  It glows shiny and get sturdier with each spell and your intelligence level.",
+    "description": "A summoned magic armor.  It glows shiny and get sturdier with each spell level.",
```
KorGgenT, line 888:
```suggestion
    "description": "A summoned magic armor.  It gets shinier and sturdier with each spell level.",
```

KorGgenT:
looks good other than minor grammar fix

Knut-Aage-Hofseth:
Since the armour now has a material thickness of 0 and work as an aura damage reduction rather than straight armour, should it still have significant weight? The version introduced in #62703 still seems to weight 11 kg.

GuardianDll:
good catch

Knut-Aage-Hofseth (approved):

Knut-Aage-Hofseth:
@GuardianDll I looked into the Mage Blade. Testing shows that when I remove `"cutting": 1,` from the item it gains cutting damage from the relic data and it looks to be about correct for the int and spell level

GuardianDll:
idk, i just removed cutting debugged 9999 int, but still can't deal any significant damage

Knut-Aage-Hofseth:
You are right, removing cutting just lets the longsword inherited cutting damage through. Damn, back to the drawing board. I really like the idea for the Mage Blade so I want to help figure out how to make it work.

Will test more thoroughly next time.

PR #64411, opened 2023-03-20
active_items: mark death drops as active items

### LINKED ISSUE #64404: food rotting after player pickup action
### Describe the bug

weird and inconsistent bug about rot timer calculation of food items drop from zombie kills... this issue is just a point where to start doing some investigations after this bug was mentioned on the devcord
image

### Attach save file

[Emerald Lake Hills-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11015532/Emerald.Lake.Hills-trimmed.tar.gz)

### Steps to reproduce

load save
kill zombies
hopefully more zombies will yield food items to test this

### Expected behavior

not sure what my stance on the expected behavior is... the food should drop already rotten? or it should stay the condition it was dropped in the first place? 
my main concern is that this leads to players unaware of this behaviour potentially picking up food that rots just a turn later, annoying to seay the least...

### Screenshots

image

Win 10
2023-03-20-0422
Tiles
English
mod list:[
  "dda",
  "no_npc_food",
  "no_fungal_growth",
  "Tamable_Wildlife",
  "stats_through_kills",
  "speedydex",
  "package_bionic_professions",
  "blazeindustries",
  "rub_reb",
  "No_Person",
  "extra_mut_scens",
  "MMA",
  "translate_dialogue",
  "No_Rail_Stations"
]

### Additional context

A recent PR randomized rot timer of food drops but i don't think it matters for this thing

don't mind the mod loading shouting, it's just a couple of personal JSON changes

andrei8l:
#### Summary
None

#### Purpose of change
Monster drops aren't processed as active items
* Fixes: #64404

#### Describe the solution
Add monster drops to the active item cache
Process them once to get rid of rotten items

N/A

#### Testing
New test unit

Advance time by 10 years and kill a bunch of zombies. They should not drop any unsealed perishable food.

#### Additional context
This bug was hidden before #63233 due to the order of operations in monster drop logic.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -4807,7 +4807,23 @@ units::volume map::free_volume( const tripoint_bub_ms &p )
     return free_volume( p.raw() );
 }
 
+item_location map::add_item_ret_loc( const tripoint &pos, item obj, bool overflow )
+{
+    std::pair<item *, tripoint> ret = _add_item_or_charges( pos, std::move( obj ), overflow );
+    if( ret.first != nullptr && !ret.first->is_null() ) {
+        return item_location { map_cursor{ ret.second }, ret.first };
+    }
+
+    return {};
+}
+
 item &map::add_item_or_charges( const tripoint &pos, item obj, bool overflow )
+{
+    return *_add_item_or_charges( pos, std::move( obj ), overflow ).first;
+}
+
+std::pair<item *, tripoint> map::_add_item_or_charges( const tripoint &pos, item obj,
+        bool overflow )
 {
     // Checks if item would not be destroyed if added to this tile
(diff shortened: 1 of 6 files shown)


Night-Pryanik:
Won't this make all grenades dropped by soldiers detonate on their death?

andrei8l:
If they are active, they will tick once which can make them explode if they're on the last turn. 

I'll see if I can figure out another solution (such as not spawning the rotten items in the first place), but I'll keep this one if the alternative needs extensive looping.

PatrikLundell:
It sounds tricky to determine whether something is merely getting old compared to being too old without spawning the item. A fully proper solution should remove the rotten items only if they're old enough to have rotted away.

PR #59620, opened 2022-07-26
Add a new faction: the Norse Evangelicals

MNG-cataclysm:
#### Summary
Content "Adds a new faction, a base for them, and several quests."

#### Purpose of change
Adds a new faction: the Norse Evangelicals. They are based in an old re-enactment set for Viking demonstrations, which they have fortified and made into a home. They are a blood-and-honor cult, mainly based around the xenophobic extermination of zombies and otherworldly creatures. They hold a special hatred for the Mycus and its disciples, considering them traitors to humanity. 

#### Describe the solution
The layout of their home fortress holds a henge in the middle, surrounded by a storehouse, a central longhouse containing the chief, homes, gardens, and mechanic shops. The fortress was intended to be a re-enactment site equipped with some modern amenities for the on-site staff. It was never finished before the Cataclysm; while the plumbing works, it has no electricity. 

The high gode (main leader) gives out combat-oriented quests centered around killing the Mycus, mi-gos, zombies, etc. 
The gode found at the henge offers more NPC-centric quests, like saving a raiding party or killing bandits.
The warehouse accountant is the local merchant, allowing players to buy bulk household items, medieval armor and weapons, and some assortment of guns. They offer far tamer missions than the leader; currently extending to fetch quests that open up bulk trade and loan opportunities later. 
The mechanics sell vehicles; specifically bikes, and provide basic fetch quests for automotive supplies.

#### Describe alternatives you've considered
I have considered rewriting the leaders' backstory to have the entire faction spawn on a delay, but I don't know how to do that. 

#### Testing
All dialogue options, quests, and trades work. No errors are given on loading a game with this content active.

#### Additional context
For some historical background, a 'gode' is a religious leader in the old Norse religions prior to the arrival of Christianity. The 'high gode' is of my own invention, seeing as how the current leader was a close friend of the cults' founder. "Reisendre" should mean "traveler" in Norse.
(SPOILER) -> The founder died to a hulk, revealed in the custom leader backstory.

If anyone knows otherwise, please correct me.

Current development plan:

- [x] Give the gode some quests
- [x] Give mechanics some jobs to do
- [x] Expand interactions with generic members
- [x] Expand quests for the high gode
- [x] Expand quests for the warehouse manager
- [x] Test code
- [x] Bugfix and lint
- [x] Finalize and un-draft for submission

PR #64417, opened 2023-03-20
Obsolete valhalla cultists

dseguin:
#### Summary
None

#### Purpose of change
The faction and maps added in #59620 have a number of issues. There's been a lot of discussion surrounding this content and how it should be either fixed or removed.

I don't think this content meshes well with the current design of the game so I've opted to remove it.

#### Describe the solution
Move the special and terrains to the obsoletion files and set the spawn occurrence to zero.

Fleshing out the faction and fixing the various issues, but no one has stepped up to the plate and I'm not interested in keeping them around.

#### Testing
Created a save beforehand with the valhallists spawned and visited the camp. After loading with the changes, the save produces no errors and the camp appears as it was.

Created a save after the changes and traveled 10 overmaps, the camp does not spawn.


Maleclypse:
I think it would be best to list the issues that are needing to be addressed to make it fit, as many/most of those conversations have happened outside of the sight lines for the majority of contributors?

dseguin:
> I think it would be best to list the issues that are needing to be addressed to make it fit, as many/most of those conversations have happened outside of the sight lines for the majority of contributors?

This is something that people have been badgering me about on Discord. Here's a list of complaints that I've heard:
- The camp spawns day one, which means the cultists have presumably constructed many buildings and gathered enough resources to rival the refugee center or the Hub
- None of the supplies in the camp are owned by the faction, so the player can just take everything right in front of them
- The npc interactions need to be fleshed out so that they sound like believable characters
- The theme of this faction could be seen as problematic, some people associate the old norse imagery and identity with hate groups. Also, the inclusion of slaves is a bit eyebrow-raising, even if the cultists are trying to rescue them

Maleclypse:
I'm responding to the list above so that if there are interested parties they can work on them.

> * None of the supplies in the camp are owned by the faction, so the player can just take everything right in front of them
1. Simple faction ownership change on mapgen.

>* The camp spawns day one, which means the cultists have presumably constructed many buildings and gathered enough resources to rival the refugee center or the Hub in an equally short amount of time
2. This needs to be much more like the Tacoma ranch in that at best there is a mostly undeveloped camp site that over time with player assistance can be built up into a semi fort campsite over several seasons of work.
>* The npc interactions need to be fleshed out so that they sound like believable characters
3. Probably best to have various interactions and shop lists and mission lists increase as time goes on so editing them with days past cataclysm type things, would also be a place where the camp can be developed as part of the character developments.  Also most of the characters should probably reference their life before joining this camp, some should be true believers and some should be like "I'm larping for safety", and other responses.
>* The theme of this faction could be seen as problematic, some people associate the old norse imagery and identity with hate groups. Also, the inclusion of slaves is a bit eyebrow-raising, even if the cultists are trying to rescue them. 
4.  If the other parts are completed I will personally write additional dialogue and encounters to make them antifascist, because I'm a firm believer in "No Nazis in Valhalla".  Rescuing captives from the Hell's Raiders is going to be a thing eventually no matter what, should they be renamed captives then to avoid a painful word?

dseguin:
If you can address those issues then that would be great, cheers.

Westbend:
Regarding the "Full camp at day 1" issue, isnt part of their backstory that they are a hobby group that built the replica viking fort/settlement before the disaster? I could have sworn thats what the leader says when you talk to him.

MNG-cataclysm:
> Regarding the "Full camp at day 1" issue, isnt part of their backstory that they are a hobby group that built the replica viking fort/settlement before the disaster? I could have sworn thats what the leader says when you talk to him.

Yes, that is part of their backstory. I initially included this as a reasonable means of explaining why they exist on the first day.

MNG-cataclysm:
The Valhallists don't own slaves or practice slavery either, the ones you rescue are emancipated upon mission completion.

MNG-cataclysm:
I did some testing with faction ownership, and the necessary code for item designations was in place. The problem seems to be a larger bug in #64599.

PR #64419, opened 2023-03-20
Tidy item examine display

irwiss:
#### Summary
None

#### Purpose of change

Make item examine text nicer - show spell class on scrolls and fold large walls of text

#### Describe the solution

First commit makes spell classes displayed the item info pane

Second commit adds `enumerate_lcsorted_with_limit` which takes a container, sorts via localized_compare, and forwards up to `max` entries to the normal `enumerate_as_string`.

This allows to fold large enumerations in item descriptions - mostly relevant to stuff that can store recipes or popular crafting components which dump out gigantic sheets of text into the info window, also refactors enumerate_as_string to use the container overload instead of iterators

Memorize all magiclysm spell classes and keep scrolling on

#### Testing

Poked a few items in my inventory, seems to work alright

#### Additional context
First commit adds the spell class to it's info screen:
image

Second commit folds large enumerations (slightly outdated screenshot, there is no second 'and' conjecture but i'm lazy to update it ☺):
image

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -414,6 +414,26 @@ static const item *get_most_rotten_component( const item &craft )
     return most_rotten;
 }
 
+// sorts with localized_compare, and enumerates entries, if more than \p max entries
+// the rest are abbreviated into " and %d more"
+static std::string enumerate_lcsorted_with_limit( const std::vector<std::string> &v, int max = 5 )
+{
+    std::vector<std::string> c = v; // copy so we can non destructively sort
+    std::sort( c.begin(), c.end(), localized_compare );
+    if( max < 1 ) {
+        debugmsg( "max expected to be 1 or more" );
+        return "";
+    }
+    if( max >= static_cast<int>( c.size() ) ) {
+        return enumerate_as_string( c );
+    }
+    const std::string res =  enumerate_as_string( c,
+    [&max]( const std::string & r ) {
+        return ( --max >= 0 ) ? r : "";
+    }, enumeration_conjunction::none );
+    return res + string_format( _( " and %d more" ), -max );
+}
+
 static time_duration get_shortest_lifespan_from_components( const item &craft )
 {
     const item *shortest_lifespan_component = nullptr;
@@ -742,13 +762,11 @@ item &item::ammo_set( const itype_id &ammo, int qty )
                 if( opts.empty() ) {
                     const std::string magazines_str = enumerate_as_string( mags,
                     []( const itype_id & mag ) {
-                        return string_format(
-                                   // NOLINTNEXTLINE(cata-translate-string-literal)
-                                   "%s (taking %s)", mag.str(),
-                                   enumerate_as_string( mag->magazine->type,
+                        const std::string ammotype_str = enumerate_as_string( mag->magazine->type,
                         []( const ammotype & a ) {
                             return a.str();
-                        } ) );
+                        } );
+                        return string_format( _( "%s (taking %s)" ), mag.str(), ammotype_str );
                     } );
                     debugmsg( "Cannot find magazine fitting %s with any capacity for ammo %s "
                               "(ammotype %s).  Magazines considered were %s",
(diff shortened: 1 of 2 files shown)


src/item.cpp
```diff
@@ -413,6 +413,26 @@ static const item *get_most_rotten_component( const item &craft )
...
+    }
+    if( max >= static_cast<int>( c.size() ) ) {
+        return enumerate_as_string( c );
+    }
+    const std::string res =  enumerate_as_string( c,
+    [&max]( const std::string & r ) {
+        return ( --max >= 0 ) ? r : "";
+    }, enumeration_conjunction::none );
+    return res + string_format( _( " and %d more" ), -max );
+}
```
Qrox, line 434:
```suggestion
static std::string enumerate_lcsorted_with_limit( const std::vector<std::string> &v, const int max = 5 )
{
    if( max < 1 ) {
        debugmsg( "Expected max to be 1 or more" );
        return std::string();
    }
    std::vector<std::string> c( std::min<size_t>( max, v.size() ) );
    std::partial_sort_copy( v.begin(), v.end(), c.begin(), c.end(), localized_compare );
    const int more = static_cast<int>( v.size() - c.size() );
    if( more == 0 ) {
        return enumerate_as_string( c );
    }
    const std::string res = enumerate_as_string( c, enumeration_conjunction::none );
    return string_format( max == 1 ? _( "%s and %d more" ) : _( "%s, and %d more" ), res, more );
}
```
Should be more memory and time efficient, please test if it works.

src/item.cpp
```diff
@@ -5798,35 +5750,18 @@ void item::final_info( std::vector<iteminfo> &info, const iteminfo_query *parts,
             std::sort( result_parts.begin(), result_parts.end(), localized_compare );
 
             // Endarken parts that can't be installed with the character's skills
```
Qrox, line 5752:
```suggestion
```

src/item.cpp
```diff
@@ -5757,37 +5722,24 @@ void item::final_info( std::vector<iteminfo> &info, const iteminfo_query *parts,
...
                 if( is_duplicate || !predicate( vp ) ) {
                     continue; // skip part variants, they have same part names
                 }
 
-                bool can_install = player_character.meets_skill_requirements( vp.install_skills );
+                const bool can_install = player_character.meets_skill_requirements( vp.install_skills );
+                const std::string name = vp.name();
 
-                result_parts.emplace_back( std::make_pair( vp.name(), can_install ) );
+                result_parts.emplace_back( can_install ? name : string_format( "<dark>%s</dark>", name ) );
```
Qrox, line 5742:
```suggestion
                 // Endarken parts that can't be installed with the character's skills
                result_parts.emplace_back( can_install ? name : string_format( "<dark>%s</dark>", name ) );
```

src/iuse_actor.cpp
```diff
@@ -2190,6 +2190,9 @@ void learn_spell_actor::info( const item &, std::vector<iteminfo> &dump ) const
     for( const std::string &sp_id_str : spells ) {
         const spell_id sp_id( sp_id_str );
         spell_text = sp_id.obj().name.translated();
+        if( sp_id->spell_class.is_valid() ) {
+            spell_text += ", " + sp_id->spell_class->name() + " ";
```
Qrox, line 2194:
```suggestion
            spell_text = string_format( pgettext( "spell name, spell class", "%s, %s" ), spell_text, sp_id->spell_class->name() );
```
Some languages need to localize comma and space. Also the trailing space seems unnecessary.

src/item.cpp
```diff
@@ -5676,71 +5676,36 @@ void item::final_info( std::vector<iteminfo> &info, const iteminfo_query *parts,
...
-
-            int processed_recipes = 0;
-
-            // Build result_names vector of at most the first `max_recipes` items
+            std::vector<std::string> crafts;
             for( const recipe *r : item_recipes ) {
-                // Break out early if we've hit our limit
-                if( processed_recipes++ > max_recipes ) {
-                    break;
-                }
```
Qrox, line 5703: I'm a bit worried about the speed of this because previously it only enumerates whatever the first 12 recipes are but now it enumerates all available recipes for their names. Is there any noticeable lag when you examine a smartphone with a very large number of stored recipes?

irwiss: I don't notice lag with fairly loaded smartphone (has most recipes from vanilla+afs+magiclysm i think), but the machine i'm testing on might be a bit beefier than average.

The heaviest operation in that loop seems to be evaluating crafting inventory, may be removing the graying out entirely is acceptable?
I don't believe people will consult these tooltip texts over opening the proper crafting menu

PR #64421, opened 2023-03-20
Attempts to fix nether monster generation

FarkasKBence:
#### Summary

None

#### Purpose of change

Basic Build and Test is erroring because of mutable shenanigans.

#### Describe the solution

Visualized how the mutable generates and found out that, on rare occasions, the head exceeds the safe-to-place zone.

<details>
<summary> </summary>

kép
kép
</details>

So I expanded the zone.

This PR also fixes stair misalignments and auto-roof generation. I also removed the ˙CLASSIC˙ flag from the location, cause it wouldn't make much sense in a classic zombie movie.

If it's still failing I can just further expand the zone check. Or shorten the tentacles & the necks, I guess?

#### Testing

Debug-spawned in the world. Didn't help much, as the game wasn't giving errors earlier either.

#### Additional context

None.


LovamKicsiGazsii:
Attempt 2 it is then.

kevingranade:
Looks like that did it, since it's breaking all over I'll go ahead and merge this and a later pr can try and re-add the tentacles in a way that doesn't break.

jbytheway:
FWIW I think the best fix for the neck placement issues would be to add an alternate decapitated version which spawns if there's not space to spawn the head.  But I realize that's more effort.

PR #64422, opened 2023-03-20
Palisade Rework

NotDucc:
#### Summary
Balance "Palisade Rework"

#### Purpose of change

The purpose of this change is to increase the bash minimum and maximum required to break palisade walls. The current strength of these walls are very low despite this style of fortification being incredibly durable due the timber being buried 5 feet into the ground, offering greater resistance through the earth's resistance.

A real world example of palisade wall strength would be the Battle of Ōhaeawai, where this fortification was used by the Māori defending against British troops. A quote from an article on the battle as British troops took down the walls, "Large fires were made against the stockade in several places, and where the posts were too thick to burn while standing upright, they were pulled down by main force. In some cases it took the whole strength of forty men, with ropes, to pull down one post, although much of the earth at the base had been previously dug away for the purpose of loosening it." Keep in mind, this was after their victory against the Māori and there was still difficulty taking down the walls.
#### Describe the solution

The solution to this was simple, I updated the palisade walls to have a bash minimum of 350 and a maximum of 500. Palisade gates and embrasures were given lower values with 300 minimum and 400 maximum. I also doubled the log requirements for palisade construction as the current log requirement didn't make a lot of sense, alongside giving it a wood sawing requirement to sharpen the timber.

Using lower values.
#### Testing

Tested in-game and it works as intended.
#### Additional context

The quoted article that goes over Māori use of palisade walls: 
https://nzetc.victoria.ac.nz/tm/scholarly/tei-BesPaMa-t1-body-d6-d1-d5.html

Screenshot (965)

### DIFF
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -1478,9 +1478,9 @@
     "category": "CONSTRUCT",
     "required_skills": [ [ "fabrication", 4 ], [ "survival", 2 ] ],
     "time": "120 m",
-    "qualities": [ [ { "id": "DIG", "level": 2 } ] ],
+    "qualities": [ [ { "id": "DIG", "level": 2 } ], [ { "id": "SAW_W", "level": 2 } ] ],
     "components": [
-      [ [ "log", 3 ] ],
+      [ [ "log", 6 ] ],
       [
         [ "wire", 6 ],
         [ "wire_barbed", 4 ],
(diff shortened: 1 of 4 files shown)


NetSysFire:
Tests fail because of something unrelated (again).

I feel like our palisades do differ from theirs. I read the article, which is a really interesting source by the way, and it looks like they are using whole tree trunks as individual "logs".

To construct a palisade ingame, one needs a pit, some wire/rope and 3 logs. These logs are 140 cm long. I do not know how deep `t_pit` is but I guess about 3 meters, guessing from the damage you get from falling in one. I think that means the recipe is woefully outdated and your change would fit if the recipe is updated, too.

The problem we have here is that we do not have longer logs. There are tree trunks but these are terrain. I think the easiest solution is to require at least double the logs (that means 6, so one you bury, the other one you stick on top by sawing an indentation so it fits together or something), require wood sawing (to actually sharpen the upper end of the logs and to maybe fit them together) and some additional earth to fill the pit back up.

Due to the hacks used I would reduce their bash protection a little bit but otherwise it should be fine. Also apply the changes to the palisade gates.

NotDucc:
Yep, figured it was something unrelated based off the error.

Also agree that the construction recipe is completely outdated and that doubling the log requirement + add "wood sawing" to the requirements should make the recipe a bit more realistic if you imagine it's just two large lumber and not separated logs (like you said, it's not possible to use the tree trunk terrain). I'll see about making that edit for the PR tomorrow morning.

As for the bash values... Not 100% sure on how much to lower it by if anything at all, they would still realistically have very high values that would make it near if not completely impossible to break down through melee damage, though still have the ability to be broken through explosives or fire.

NetSysFire:
Only thing that is still missing right now is some earth to fill the pit in the construction. But take your time, quality over quantity here, especially if you do some research for proper sources on palisade building.

Issue #64406, opened 2023-03-20
Hang on attempting wield with many items


Cenness:
### Describe the bug

If container has non-zero wield cost (pocket_data[].moves) and a lot of items inside, opening wield menu takes ages of irl time - more than 10 minutes in provided save.  
Changing cost to 0 makes menu open in under one minute.

### Attach save file

[wcost.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/11016568/wcost.zip)

### Steps to reproduce

1. in game\data\json\items\armor\storage.json L2400:
    change to
    ```
        "weight_multiplier": 0.00,
        "volume_multiplier": 0.00
    ```
    so character could actually stand,
2. in game\data\json\items\armor\storage.json L2398: change `"moves": 1,` to `"moves": 0,`
3. spawn and wear debug backpack
4. grant char every item
5. try to open wear menu - it opens almost instantly
6. try to open wield menu - it opens in a minute
7. save and exit
8. in game\data\json\items\armor\storage.json L2398: change `"moves": 0,` to `"moves": 1,`
9. load game - it will take a lot of time
10. try to open wear menu - it still opens almost instantly
11. try to open wield menu - game window hangs, ~10 minutes for the menu to show

### Expected behavior

Move cost shouldn't influence how long it takes to open a menu
Menu should open in less than a minute
Subsequent menu access should be instantenous - add per-item caching maybe?

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.2728 (22H2)
- Game Version: b738ae2 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda]
]

andrei8l:
You have ~9000 unique, non-stackable items in the pocket universe. The wield menu does open several orders of magnitude slower than the inventory so this does look like a real bug, but fixing it has mostly only academic value.

ZhilkinSerg:
> But why it's faster with zero move cost? Neither [item_handling_cost](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/character.cpp#L6706) or [obtain_cost](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/item_pocket.cpp#L2029) have special path for 0.

Because game is processing while items are being moved.

PR #64425, opened 2023-03-20
inv_ui: move cell cache to entry and delete redundant calls

andrei8l:
#### Summary
None

#### Purpose of change
Address another (~largely academic~) bottleneck in the inventory UI
* See #64406

EDIT: this turned out to have a positive perf impact when collation is enabled so it's no longer just academic

#### Describe the solution
for `item_location`: don't copy the target item willy-nilly when calculating obtain cost and also use the pocket shortcut
for `inventory_ui`: move cell cache to entry and remove/reorder redundant calls so that values are only calculated and cached once per execution
for `item`: don't restack charges when deserializing since they had already been stacked before serialization.

Fixing the lower-level things that made the wield UI slow. In this case that means improving `item::weight()` but I don't see a good/robust way without item backlinks. A project for later, maybe.

#### Testing
The inventory/wield/pickup UIs should look and function exactly the same as before. Filtering should still work (including on return in the `E`at menu), reflowing when (un)collapsing should be the same, etc.

Using the save file from #64406 and opening the `w`ield menu:
Before:

before

After:

after

It's still slow, but not quite enough for a coffee break anymore.

Load time of that save should be drastically faster with 7adcce0186e97b3104cb7bc73d613f0514bfc98c

In other cases, such as in my yardstick `City of the Dalles` save from other PRs, there should be no perf difference.

#### Additional context
N/A

### DIFF
--- a/src/inventory_ui.cpp
+++ b/src/inventory_ui.cpp
@@ -577,8 +577,11 @@ void inventory_entry::update_cache()
 void inventory_entry::cache_denial( inventory_selector_preset const &preset ) const
 {
     if( !denial ) {
-        denial = { preset.get_denial( *this ) };
+        denial.emplace( preset.get_denial( *this ) );
         enabled = denial->empty();
+    }
+    if( is_collation_header() ) {
+        collation_meta->enabled = denial->empty();
     }
 }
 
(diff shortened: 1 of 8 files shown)


andrei8l:
Thanks for merging.

PR #22378, opened 2017-11-07
Prompt when creating a character with the same name in a world.

OzoneH3:
Fixes #16419

PS: Another thing in this PR not really worth it's own.
* Removed the Y/N query on world selection. I selected the world for a reason in the world selection dialog, don't ask me again if I really want to use that world, just do it.

PR #64428, opened 2023-03-21
Fix warning prompt when creating a character with the same name in a world not appearing

### LINKED ISSUE #64413: A warning if you are going to overwrite a character

I have noticed that since I have set up a preferred name I always forget to change it when testing or checking things to do with other professions. If I make a character using the same name as another character in that world it does not warn that the other character will be replaced.

### Solution you would like.

The best thing I can think is having another "are you sure" popup when making a character, something along the lines of
`|------------------------------------------------------|`
`|                     /| WARNING |\                    |`
`|    [Character name] already exists in [World Name]!  |`
`|    Continuing will overwrite the previous character. |`
`|------------------------------------------------------|`

_No response_

### Additional context

_No response_

strategineer:
#### Summary
Bugfixes "Fix warning prompt when creating a character with the same name in a world not appearing"

#### Purpose of change
Fix warning prompt when creating a character with the same name in a world not appearing.

This feature was implemented here #22378

The regression seems to have happened here https://github.com/CleverRaven/Cataclysm-DDA/commit/a1bf4e975be#diff-da8a925e8abf39943d87477831cecd3686ad13b250b7bc1c4bd74ffeaaefce34L758

Fixes #64413

#### Describe the solution

This fixes the bug by inserting the logic that checks for the duplicate name in the right place (right before confirming that you are "SURE" that you're finished creating your character). Also, change ensures that we never pick a randomly generated name that already exists in the world to avoid overriding an existing character unintentionally.

None.

#### Testing

Tested all the different ways to create new characters and ensured that the warning showed up when appropriate and ensured that it reacted in the expected way when answering yes and no.

#### Additional context

WindowsTerminal_ohW0ZzsGHa

### DIFF
--- a/src/newcharacter.cpp
+++ b/src/newcharacter.cpp
@@ -665,9 +665,6 @@ bool avatar::create( character_type type, const std::string &tempname )
     prof = profession::generic();
     set_scenario( scenario::generic() );
 
-    const bool interactive = type != character_type::NOW &&
-                             type != character_type::FULL_RANDOM;
-
     std::vector<std::string> character_tabs = {
         _( "POINTS" ),
         _( "SCENARIO" ),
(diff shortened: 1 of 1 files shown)


strategineer:
I need to figure out why the windows build is consistently falling...

Night-Pryanik:

PR #64429, opened 2023-03-21
Removed the small lens from smartphone deconstruction because

### LINKED ISSUE #55182: Cell phones and small lenses

Modern smartphones generally don't have removable small high-quality lenses, especially one ones that are 1ml, or 20 grams(!?!) in size.

The entire camera assembly and board might be 1 or 2ml in volume.

I don't think (atomic) smartphones should let you get the lens as a discrete component.  They are generally something like 1 or 2mm on a side, and less than that thick, which is way to tiny to be able to extract, especially with just a screw driver.  (The piece of plastic on the outside of a phone is just a cover, afaik).  That's also way to tiny to be utilized for pretty much any recipe that needs a high-quality lens.  

Lastly, if you look at the list of items that you get small high-quality lenses from, all the rest of them have lens that are in line with 10ml, 20 gram lenses.  Telescope, microscopes, atomic headlamp(?).  The only one that isn't is the 'sensory array'.

### Solution you would like.

Remove small high-quality lenses from disassembly of smartphones, and atomic smartphones.  Probably the sensory array as well, but I know less of what that's actually supposed to be.

If needed in the future, change small high-quality lens from the smartphone into pinhole camera assembly, and let possible future recipes for things that need it use that instead.  It's not like you'd be able to mix&match random pinhole camera & lenses in the real world, anyways.  At least not without 300 page datasheets for the cameras, that require strict NDA's to get.*

Possibly entirely obsolete the small high-quality lens?  There's nowhere it's used (afaik) that it's not substitutable with a a normal high-quality lens.

### Additional context

Searching within this topic is absolutely polluted with clip-on lenses for cell phones.

*I once tried to get the datasheet for a cell phone camera component off digikey.  Uni EE project.  It was an absolute pain, even though getting the actual camera was pretty easy. It was also an abomination of mounting nightmares with ~25 contacts across 5 of the faces of a 2.5mm cube.

Cell Phone lens in much larger plastic housing, circa 2017:
<https://youtu.be/GMd77on_k0Q?t=200>

strategineer:
#### Summary
None

#### Purpose of change

(apparently) Modern smartphones don't have removable small high-quality lenses (or at least not ones of a similar kind to what a microscope would have).

Fixes #55182

#### Describe the solution

Self-explanatory.

Create a `lens_micro` component that could be obtained from deconstructing a smartphone, and recipes that would make use of it (a lot of work though).

#### Testing

Spawned a smartphone, deconstructed it and noted that no `lens_small` was produced.

### DIFF
--- a/data/json/recipes/recipe_deconstruction.json
+++ b/data/json/recipes/recipe_deconstruction.json
@@ -2218,8 +2218,7 @@
       [ [ "light_plus_battery_cell", 1 ] ],
       [ [ "small_lcd_screen", 1 ] ],
       [ [ "mobile_memory_card", 1 ] ],
-      [ [ "processor", 1 ] ],
-      [ [ "lens_small", 1 ] ]
+      [ [ "processor", 1 ] ]
     ]
   },
   {


chrispikula:
You're a champ!

MochaModifier:
I recall there being atleast one possible NPC quest, where they request multiple small lens.
Not sure which it was, meteorologist maybe(?), will they still be requesting this item, or is there an alternative quest item now?

chrispikula:
Several other times provide the small lens.

strategineer:
Ya it's the meteorologist. sigh :(

NPC stuff has not really been on my radar until now, I'll keep this in mind.

Not many items disassemble into the small lens though (the "afs_atomic_smartphone" does right now but it probably shouldn't if the regular smartphone doesn't). There's the microscope, microscope_dissecting, sensor_module, melting_point (whatever that is).

The meteorologist has the following dialog line:
> Try looking for cameras to dismantle.  Smart Phones, Telescopes or microscopes would also work.

I didn't see this... Now, there's two options here:
- Revert the change and let smartphones disassembly produce "lens_small" again.
- Make "afs_atomic_smartphone" disassembly not produce "lens_small", change all dialog line (forcing all the translations to be redone) and maybe reduce the number of requested "lens_small" the meteorologist is asking for.

What should I do?

anothersimulacrum:
Go for option 2. Cell phones shouldn't produce lenses like this.

MochaModifier:
Could change Meteorologist to require the larger lens as well. Otherwise, not sure where you could find the lens she needs outside raiding a lab - microscopes and telescopes are highly uncommon, yet they want half a dozen of them?

chrispikula:
What about adding them to the disassembly of professional cameras?

Another option would be for some sort of child's science toy to have them.  For making a periscope, or whatnot.

strategineer:
> Could change Meteorologist to require the larger lens as well. Otherwise, not sure where you could find the lens she needs outside raiding a lab - microscopes and telescopes are highly uncommon, yet they want half a dozen of them?

I'm probably the furthest thing from an expert on NPC mission balance so I can't speak to that.

But to clarify, the mission is asking for 5 small lenses but most things that you can disassemble into small lens(es) produce more than 1small lens.

A microscope produces 8 for example. Sensor modules produce 2.

PR #64431, opened 2023-03-21
Remove cata::void_t; migrate to std::void_t

jbytheway:
#### Summary
None

#### Purpose of change
Simplifying code and moving to standardized solutions.

Now we're in C++17 we can use `std::void_t` instead of reinventing it.

#### Describe the solution
Remove `cata::void_t`.

Migrate all uses to `std::void_t`.

None.

#### Testing
Compiled.

#### Additional context


anothersimulacrum (approved):

kevingranade:
Maybe rebase so we can see the build kick off.

PR #64434, opened 2023-03-21
Raise minimum supported macOS version to 10.13

BrettDong:
#### Summary
None

#### Purpose of change
```
Error: src/vpart_position.h:126:34: error: 'value' is unavailable: introduced in macOS 10.13
            return has_value() ? value().get_label() : std::nullopt;
                                 ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include/c++/v1/optional:950:33: note: 'value' has been explicitly marked unavailable here
    constexpr value_type const& value() const&
                                ^
1 error generated.
```

#### Describe the solution
* Update documentation to claim macOS >= 10.13 is supported
* Build with latest Xcode on macOS 12

Currently our GitHub Actions workflow runs on macOS 10.15 and uses Xcode 12.4. It claims `std::optional` is available on macOS 10.14 or later. But when I compile locally on macOS 12 with Xcode 14.2, it says `std::optional` is available on macOS 10.13 or later.

### DIFF
--- a/.github/workflows/matrix.yml
+++ b/.github/workflows/matrix.yml
@@ -148,7 +148,7 @@ jobs:
             ccache_limit: 4G
 
           - compiler: clang++
-            os: macos-10.15
+            os: macos-12
             release: 0
             cmake: 0
             native: osx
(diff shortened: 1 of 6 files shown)


kevingranade:
It looks like raising the supported version to 10.13 still doesn't violate our 5% market share threshold, and from what I can tell this is actually the version we were expecting to need to jump to to support c++17.

PR #64442, opened 2023-03-21
64366

Pointabc:
#### Summary
I18N "Fixed translation and double spaces"


data/mods/Xedra_Evolved/mutations/classes.json
```diff
@@ -4,7 +4,7 @@
     "id": "DREAMER",
     "name": "Dreamer",
     "points": 0,
-    "description": "Dreamers are people able to tap into the power of dreams and nearby dimensions and create fantastical effects as a result.  They can even pull some of this dream material into our world to build other things with it.  Gains a spell to summon dreamdross scraps on demand.",
+    "description": "Dreamers are people able to tap into the power of dreams and nearby dimensions and create fantastical effects as a result. They can even pull some of this dream material into our world to build other things with it. Gains a spell to summon dreamdross scraps on demand",
```
kevingranade, line 7:
```suggestion
    "description": "Dreamers are people able to tap into the power of dreams and nearby dimensions and create fantastical effects as a result.  They can even pull some of this dream material into our world to build other things with it.  Gains a spell to summon dreamdross scraps on demand",
```

data/mods/Xedra_Evolved/mutations/classes.json
```diff
@@ -17,7 +17,7 @@
     "id": "EATER",
     "name": "Eater of Dreams",
     "points": 0,
-    "description": "These people are aware of this cross-dimensional contamination and while they can't pull anything over and make new things, they can eat anything they find in our world that doesn't belong and gain abilities and powers from that consumption.  They have larger than normal mana pools but regen that mana much slower as a result, they can a spell to convert dreamdross scraps directly into mana.",
+    "description": "These people are aware of this cross-dimensional contamination and while they can't pull anything over and make new things, they can eat anything they find in our world that doesn't belong and gain abilities and powers from that consumption. They have larger than normal mana pools but regen that mana much slower as a result, they can cast a spell to convert dreamdross scraps directly into mana.",
```
kevingranade, line 20:
```suggestion
    "description": "These people are aware of this cross-dimensional contamination and while they can't pull anything over and make new things, they can eat anything they find in our world that doesn't belong and gain abilities and powers from that consumption.  They have larger than normal mana pools but regen that mana much slower as a result, they can cast a spell to convert dreamdross scraps directly into mana.",
```

data/mods/Xedra_Evolved/mutations/classes.json
BrettDong: Is removing the trailing period intended?

data/mods/Xedra_Evolved/mutations/classes.json
```diff
@@ -4,7 +4,7 @@
     "id": "DREAMER",
     "name": "Dreamer",
     "points": 0,
-    "description": "Dreamers are people able to tap into the power of dreams and nearby dimensions and create fantastical effects as a result.  They can even pull some of this dream material into our world to build other things with it.  Gains a spell to summon dreamdross scraps on demand.",
+    "description": "Dreamers are people able to tap into the power of dreams and nearby dimensions and create fantastical effects as a result.  They can even pull some of this dream material into our world to build other things with it.  Gains a spell to summon dreamdross scraps on demand",
```
Fris0uman, line 7:
```suggestion
    "description": "Dreamers are people able to tap into the power of dreams and nearby dimensions and create fantastical effects as a result.  They can even pull some of this dream material into our world to build other things with it.  Gains a spell to summon dreamdross scraps on demand.",
```

Pointabc: @Fris0uman if restore dot after demand, we will get double dots, because in code
if( cancel == sp->spell_class->cancels.back() ) {
                    trait_cancel += ".";
}
magic.cpp line 1748

Brambor: https://github.com/CleverRaven/Cataclysm-DDA/blob/71d39820d37071c161f248f4a0457f7f3bea177e/src/magic.cpp#L1748-L1750

The simplest fix would be to add a condition there and keep the string ending with `.`:

```cpp
if( cancel == sp->spell_class->cancels.back() && !trait_cancel.ends_with(".")) {
    trait_cancel += ".";
}
```

But maybe there is a better fix. I didn't look at what the code is supposed to do, or what strings get there.

Fris0uman: No but if the games adds it any way it's fine I guess. Sorry I didn't know that. It is a bit weird that the code does that since it doesn't for other descriptions though, but that's out of scope for this PR  I think

PR #61896, opened 2022-10-24
Various Additions way too late for Native American Heritage month

merrygin:
#### Summary
Content "Various additions and tweaks relating to Native American Heritage"

#### Purpose of change
November is Native American Heritage month, as some of you might know. That might be a nice opportunity to make additions that relate to Native American heritage and are beneficial to the game or should be represented in the game. For example, this would Fix #61663 by adding harvestable wild rice and make it an ingredient for several recipes.

#### Describe the solution
This is a list of things included or planned. You can refer to the numbers if you have specific comments/requests.

- [x] 1. Add Wild Rice (aka "Indian Wild Rice"), a staple of many Great Lakes and New England tribes in the past and still part of many tribes traditional cuisine. Also, native though increasingly rare plant in this area.
- [x] 2. Added dishes: Succotash, Frybread, "Indian Tacos", Hazelnut cakes, Three sisters stew, pemmican (only description change to highligh origin, also made available through recipe book, see below)
- [x] 3. Added a recipe book with Native American recipes.
- [postponed] 4. Added a simple variant of a maple spigot, which can be made by bark, as was traditionally done by tribes in New England, who invented that method.
- [x]  5.Changed Moccasin description and recipe (to make it craftable with leather and fabric cutting 1 instead of pelts and fab_cut 2)
- [x] 6. Make innawood starting scenario starting you of with historical native american gear and three sisters seeds.
- [x] 7. Add butternuts (different walnut species)? Or other species of plants native to new england that could have use to the player.
- [postponed] 8. Add some names (current or historical ones) of tribes of the area into the mix.
- [x] 9. Add squashes as seperate harvestable/plantable item, also with uses like making bottles out of it (tornikeo gave his consent to use #61007)
- [postponed] 10. more food recipes?

#### Describe alternatives you've considered
Different additions relating to Native American Heritage in New England that is beneficial to/should be represented in game? 

#### Testing
1) tested, seems to spawn and work fine now
2) all appear sensible, some remaing kcal+density issues upsetting tests, though
3) seems to work fine, spawns in shops not tested
_4) more involved solution needed than only new item, needs more time so Ill postpone this_
5) appears to work
6) appears to work
7) does not **not** work, but haven't seen one in the wild yet?! Very few/no walnut trees in general
_8) postponed_
9) all crafts work as intended, only thing not tested is if the gourd grows properly 

#### Additional context
[...]

PR #61746, opened 2022-10-16
Add more simple meals

GuardianDll:
#### Summary
Content "Add some simple, salt and pepper meals by u/MuscleAgile5442 aka Cosmic"

#### Purpose of change

Close #61686

#### Describe the solution

Added each type of meal dish variant with pepper and salt

#### Testing

The game opens, recipes looks good

#### Additional context

Let's be honest, i'm just too lazy to make proper requests, but still want to close 4 pr for october, and luckily Cosmic came to us to do cool stuff

PR #64443, opened 2023-03-21
fix more butane spawns and fix new PRs' foodcosts being wrong out of the gate

anoobindisguise:
#### Summary
Bugfixes "Butane spawns and new food PR cooking cost corrections"

#### Purpose of change
#61746 and #61896 were both merged after my food cooking costs PR but were using the wrong costs for foods. Also some butane cans were still spawning with 0-30 charges.

#### Describe the solution
Adjust cooking costs to proper values, Also fixes the spawns of butane canisters being wrong.

#61746 also messed with the weight of meat which throws off a ton of stuff in my system, but I'm working with @GuardianDll to find a resolution to it.


GuardianDll:
Re:meat weight, we actually can go into opposite direction, revert the weight back to 300g, but increase the volume to 275 ml

PR #64446, opened 2023-03-21
Timber!

NotDucc:
#### Summary
Content "Add Timber"

#### Purpose of change

While the main use of this is for use with palisade walls ( #64422 ), I can see this being useful is a few ways. Firstly, it can make transporting timber much quicker as the player won't be forced to chop every trunk and instead can easily "deconstruct" the trunk, carry the timber to their cargo space (or make a lumber stack and (G)rab four tied together) and transport it for later cutting. Secondly, this could likely be used for other large construction menu items or projects that don't really need cut logs but need timber instead. Lastly, this could be used as a "base" should trees ever get reworked to provide different timber (small, medium, or large).
#### Describe the solution

The solution was to add a new item, "timber", to the game. This is essentially 2.5 logs uncut, which I got the weight, volume, and length from averaging out how many logs you can get per trunk and combining the values. You can get this item through the "Gather timber" construction on tree trunks. If you want to cut it into logs and/or planks later, you can place it back on the ground as a tree trunk and cut it as normal. You can also construct a stack with 4 timber logs and 2 short ropes of any kind and (G)rab it using this method.

Altering values used for timber, a different name aside from "timber".
#### Testing

Tested in-game, every feature described works as intended.
#### Additional context

Screenshot (970)
Screenshot (978)
Screenshot (979)

### DIFF
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -7785,12 +7785,24 @@
   },
   {
     "type": "construction",
-    "id": "constr_place_lumber",
-    "group": "place_lumber",
+    "id": "constr_gather_timber",
+    "group": "gather_timber",
+    "category": "OTHER",
+    "required_skills": [ [ "survival", 0 ] ],
+    "time": "10 m",
+    "qualities": [ [ { "id": "TREE_CUTTING", "level": 1 } ] ],
+    "byproducts": [ { "item": "timber", "charges": 1 } ],
+    "pre_terrain": "t_trunk",
+    "post_terrain": "t_trunk"
+  },
+  {
+    "type": "construction",
+    "id": "constr_place_timber",
+    "group": "place_timber",
     "category": "OTHER",
     "required_skills": [ [ "fabrication", 0 ] ],
     "time": "15 s",
-    "components": [ [ [ "lumber", 1 ] ] ],
+    "components": [ [ [ "timber", 1 ] ] ],
     "pre_special": "check_empty",
     "dark_craftable": true,
     "post_terrain": "t_trunk",
@@ -7798,13 +7810,13 @@
   },
   {
     "type": "construction",
-    "id": "constr_stack_lumber",
-    "group": "build_stack_lumber",
+    "id": "constr_stack_timber",
+    "group": "build_stack_timber",
     "category": "FURN",
     "required_skills": [ [ "fabrication", 2 ], [ "survival", 1 ] ],
     "time": "10 m",
     "components": [ [ [ "lumber", 4 ] ], [ [ "rope_makeshift_6", 2 ], [ "rope_6", 2 ], [ "vine_6", 2 ] ] ],
     "pre_special": "check_empty",
-    "post_terrain": "f_stack_lumber"
+    "post_terrain": "f_stack_timber"
   }
 ]
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -7815,7 +7815,7 @@
     "category": "FURN",
     "required_skills": [ [ "fabrication", 2 ], [ "survival", 1 ] ],
     "time": "10 m",
-    "components": [ [ [ "lumber", 4 ] ], [ [ "rope_makeshift_6", 2 ], [ "rope_6", 2 ], [ "vine_6", 2 ] ] ],
+    "components": [ [ [ "timber", 4 ] ], [ [ "rope_makeshift_6", 2 ], [ "rope_6", 2 ], [ "vine_6", 2 ] ] ],
     "pre_special": "check_empty",
     "post_terrain": "f_stack_timber"
   }
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -7792,8 +7792,8 @@
     "time": "10 m",
     "qualities": [ [ { "id": "TREE_CUTTING", "level": 1 } ] ],
     "byproducts": [ { "item": "timber", "charges": 1 } ],
-    "pre_terrain": "t_trunk",
-    "post_terrain": "t_trunk"
+    "pre_terrain": "f_trunk",
+    "post_terrain": "f_null"
   },
   {
     "type": "construction",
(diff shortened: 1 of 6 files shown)


flybye3o6:
could you test t_trunk as a `furniture` or add a new furniture for lumber
if it was furniture it could be dragged so it could be moved without carrying it
also could use lumber as a vehicle part, slap it on the side of a car with some rope for transport and could act like log armor

NotDucc:
While true, this wouldn't solve the problem of it being unusable for construction. It also creates an issue with it being dragged against the ground at high speeds, chipping it away on whatever uneven or rocky ground it hits (including pavement). It being an item makes it usable for construction work and able to be safely stored in cargo space (or I guess a trunk, though the length would realistically cause issues if it's closed off unless tiles have an item length maximum)

flybye3o6:
carrying even welding 1 at a time seems unrealistic and also impossible for low straight characters
hauling is also unrealistic and would be a case of dragging it on the ground, along with all other items you collect along the way
(G)ragging if you flip the log furniture tile image sideways it could look less like your dragging it but rolling it
and can still keep the item version to be "picked up" for use in construction
could also make a furniture of multiple logs in a line to be rolled, and could use that for moving other things like large stones

NotDucc:
You're right and I did decide to test around with it. Main issue with converting tree trunks to furniture is that the functions that use tree trunks rely on it being a terrain rather than furniture. Setting it to a furniture created tons of errors and attempting to chop down trees made the invalid "nothing" terrain.

I could make a pallet furniture that can be constructed though. Thinking 2 short rope and 4 lumber to construct it would be reasonable.

kevingranade:
The word you're looking for is log. Lumber specifically means cut wood like planks and beams.
Cutting up a tree trunk into logs takes a lot of work,  it's intentionally not a simple deconstruct.

NotDucc:
Logs are already an item in the game, which was why I avoided using that, though I did realize later that timber would be a better name so I can change it to that if it works better. Used lumber originally because I was thinking of the UK definition, which I didn't realize was different from the US definition. English is fun.

As far as it being a simple deconstruct, my line of thinking was just picking up the tree trunk that the player fell or found in the woods, though I guess that only works if there is a single tree trunk. I can make a small construction for turning it to timber, thinking a 10 minute construction time and a `TREE_CUTTING` requirement should suffice, shouldn't take THAT much longer for only one cut.

flybye3o6:
lumber made sense to me, what your describing as "cut wood like planks and beams" is called timber for me
what is timber in US definition? is there UK English and US English translations for this game?

and yes the tree comes down as a whole so it should be cut up and require `TREE_CUTTING` 
what your trying to make is a log that is double the length of the normal log, maybe looking up logging terminology try to find some terms for sizes of log, i know there is for diameter idk about length

NotDucc:
Ah, yeah, should explain what I meant a little better: I'm switching it to the US definition (timber) because the game takes place in the US, so it would make more sense for the character themself to refer to it as timber. Timber typically refers to large uncut logs (the item in question), lumber is usually referred to for more processed wood. In the UK the definitions are flipped, lumber is used for large uncut logs and timber is usually processed.

cake-pie:
> just picking up the tree trunk

I'm not sure it was intended that felled trunks can simply be picked up and handled as-is -- considering that `t_trunk`s are created as a terrain instead of an item when a tree is felled. The latter would be simpler more straightforward, but as Kevin has already alluded to, it is intentional that cutting a non-portable trunk up into portable logs (or other smaller lumber) is a non-trivial and time-consuming task.

A quick search shows 75kg is in the ballpark of recommended deadlifting weight for male beginners, so it is a "liftable" weight, but I wonder if we really want it to be something that can practicably carried (wielded) around esp. over longer distances, which is what making it an item basically entails.

> t_trunk as a `furniture`

Since this was mentioned: trunks are being converted from terrain to furniture in #63216.
If a change is made such that 75kg trunks are considered portable items, then there's no point to having it as a terrain or furniture, might as well spawn as an item when felling trees (or directly in mapgen)

data/json/construction.json
```diff
@@ -7782,5 +7782,41 @@
...
+    "type": "construction",
+    "id": "constr_gather_timber",
+    "group": "gather_timber",
+    "category": "OTHER",
+    "required_skills": [ [ "survival", 0 ] ],
+    "time": "10 m",
+    "qualities": [ [ { "id": "TREE_CUTTING", "level": 1 } ] ],
+    "byproducts": [ { "item": "timber", "charges": 1 } ],
+    "pre_terrain": "t_trunk",
+    "post_terrain": "t_trunk"
```
cake-pie, line 7796: Created timber from nothing. You need to set post_terrain to something else, probably dirt -- but then this opens up a bug/exploit similar to #42428 allowing players to convert terrain at will.

NotDucc: > Since this was mentioned: trunks are being converted from terrain to furniture in #63216. If a change is made such that 75kg trunks are considered portable items, then there's no point to having it as a terrain or furniture, might as well spawn as an item when felling trees (or directly in mapgen)

Nice to hear it is getting a conversion to furniture. I'll fix the PR's definitions that use `t_trunk` in a bit for when it gets merged. 

However, as stated in the change's purpose, the reason behind making an item version is to make it usable with construction. Palisade in particular has the problem of using logs that are too short for the task they're meant for. Logs will still have the advantage of being more portable and actually usable in fireplaces and charcoal kilns, giving them their own use where these timber logs can't be used.

NotDucc:
I'm going to throw this back into draft. PR #63216 should help with some of the issues whenever it gets merged.

flybye3o6:
how i was imagining it was
cutdown tree = f_trunk, cutup trunk = f_timber + long sticks, simple deconstruct on f_timber = item timber, 3-4 item timber make f_timber_row
f_timber can be (G)rab where as f_trunk can not, and f_timber_row would allow you to (G)rab a whole tree worth of logs and "roll" it to its destination

NotDucc:
That's an interesting idea... I'll test a bit more with that another time, definitely when trunks are converted to furniture at the very least.

PR #64458, opened 2023-03-22
Add morale menu keybinding to the player info menu and remove the default config for the morale menu keybinding at the root level

### LINKED ISSUE #59677: Move the "morale" menu away from root level keybindings

This is part of a larger effort to identify places where our "top level" keybindings have been mapped to things that players do not need rapid, single-keypress access for. This is a discussion: all these issues are valid places to raise concerns about reasons a player should have immediate single-key access to this function.

The Morale menu is currently mapped to `V` by default, but I argue this probably makes it harder for new players to find, as it doesn't even appear to make sense as a menu in its own right.

### Solution you would like.

Move morale to the `@` menu. In the future it probably makes sense to bundle it with the same pane as effects, but prior to any full revamps of the layout of the `@` menu we could simply put a line somewhere obvious reading "press V to morale effects"

We could instead make the `V` menu more worthwhile, eg by moving active effects and things in there, but I would still argue that that stuff should be in the @ menu where players expect to find it.

In the more distant future, we should rework the entire @ menu to have a customizable tabbed layout, with morale and effects in a tab.

strategineer:
#### Summary
None

#### Purpose of change

Fixes #59677

#### Describe the solution

Self-explanatory.

None.

#### Testing

- Deleted config/keybinds.json
- Noticed that the root level morale menu keybinding was "Unbound locally" as expected.
- Noticed newly added "[m] Morale" hint in the player info menu.
- Notice the PLAYER_INFO level keybinding for view the morale menu was defaulted to "m".
- Pressing "m" brings up the morale menu on top of the player info menu as expected.

#### Additional context

player info menu (with morale keybind hint):
WindowsTerminal_9HShiIIWrg

morale menu overlaid over player info menu:
WindowsTerminal_0gPVzzyCYI

keybind added to PLAYER_INFO context: 
WindowsTerminal_ItphOd8hKj

root level morale menu keybind unbound by default:
WindowsTerminal_gLp9zXMO30

root level morale menu key pressed:
WindowsTerminal_HOUSxZwbJk

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -2611,10 +2611,9 @@
   },
   {
     "type": "keybinding",
-    "name": "View morale",
+    "name": "View character's morale",
     "category": "DEFAULTMODE",
-    "id": "morale",
-    "bindings": [ { "input_method": "keyboard_any", "key": "v" } ]
+    "id": "morale"
   },
   {
     "type": "keybinding",
@@ -4181,6 +4180,13 @@
     "id": "VIEW_BODYSTAT",
     "name": "View character's body status",
     "bindings": [ { "input_method": "keyboard_any", "key": "s" } ]
+  },
+  {
+    "type": "keybinding",
+    "name": "View character's morale",
+    "category": "PLAYER_INFO",
+    "id": "morale",
+    "bindings": [ { "input_method": "keyboard_any", "key": "m" } ]
   },
   {
     "type": "keybinding",
--- a/src/action.h
+++ b/src/action.h
@@ -257,7 +257,7 @@ enum action_id : int {
     ACTION_MISSIONS,
     /** Display factions screen */
     ACTION_FACTIONS,
-    /** Display morale effects screen */
+    /** Displays morale menu */
     ACTION_MORALE,
     /** Displays medical menu */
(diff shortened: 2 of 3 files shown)


strategineer:
I can't for the life of me figure out why a "morale" top level keybind is still showing up...

WindowsTerminal_B6PsWMfhBA

No matter what I do it's always there. Weird.

chphilli:
Why remove the global keybind? I'm certainly not against it being available / encouraged from the player menu -- that's a great place for people that don't know about it to learn about it. 

However, for more experienced players, this is just adding an extra menu-ing step to something that gets used fairly often. When I'm crafting or reading, I try to keep focus high by checking on morale often. Removing this keybind from the global context just adds a keystroke to something that I might do several dozen times in a crafting/reading heavy session.  

Is there a problem being caused by this keybind that this change solves?

strategineer:
I don't have a preference either way, I'm just implementing #59677. If someone else can speak to this, that would be great!

We could keep it at the root level and just remove the default keybinding, that'd be an easy alternative.

kevingranade:
Yea just removing it from the defaults is fine,  mainly we want to free up to level keybinds.

PR #64454, opened 2023-03-22
Weekly Changelog 2023-03-15 to 2023-03-22

kevingranade:
#### Summary
None

#### Additional context
`Automatically generated as a draft. Please copy-edit before merging.`

PR #64459, opened 2023-03-22
Fix changelog script not preserving leading spaces

dseguin:
#### Summary
None

#### Purpose of change
The `merge_summaries.sh` script discards leading spaces on existing lines, which changes a lot more of the resulting lines than intended. See https://github.com/CleverRaven/Cataclysm-DDA/pull/64454/commits/f0db93c185b881eb60aa298e2d13fc421d9bb604 (commit reversing the unintended changes)

#### Describe the solution
Empty `IFS` before running the script so that `read` doesn't try to format the line as it's reading it

### DIFF
--- a/tools/merge_summaries.sh
+++ b/tools/merge_summaries.sh
@@ -1,5 +1,8 @@
 #!/bin/bash
 
+# Preserve leading spaces
+IFS=''
+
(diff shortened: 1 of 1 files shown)


kevingranade:
Freaking IFS.

PR #63611, opened 2023-02-16
Revive part 2 of the netherium monster corpse project

MNG-cataclysm:
#### Summary
Content "Finish part 2 of the netherium monster corpse"

#### Purpose of change
I noticed that #54155 had been closed and marked as a desired PR, so I decided to finish it. 

#### Describe the solution
Finish the planned content from #54155 in the "Purpose of Change" category, and have it spawn. See the PR for full details.  

#### Describe alternatives you've considered
Not doing this or leaving it to someone else to do.

#### Testing
I spawned everything multiple times in testing to make sure it all worked, and everything worked great.

#### Additional context
I changed the C++ for the admire effect into a JSON-based EOC. Additionally, I did not include the finishing textures for the organs in the final draft, as I am unfamiliar with the processing of tilesets.

PR #64462, opened 2023-03-22
Revert "Attempts to fix nether monster generation"

kevingranade:
#### Summary
None

Reverts CleverRaven/Cataclysm-DDA#64421

Not fixed after all,  so now the plan is to revert both and get back to working.


irwiss:
Is it on purpose that this reverts just the one unsuccessful fix attempt commit?
The original PR changed 15 files https://github.com/CleverRaven/Cataclysm-DDA/pull/63611

kevingranade:
Yea I need to do this one first and then the other one since I'm doing it in my phone.

LovamKicsiGazsii:
That's a bummer, didn't expect the tentacles to error from beyond the grave. :/
So the overall plan is to prevent the mutable from spawning or removing it entirely?

GuardianDll:
Right now we remove it entirely, because both the game and our autotests are bricked because of it

PR #64463, opened 2023-03-22
Refactor and deduplicate UI input code

sparr:
#### Summary
Interface "Minor additions and changes to UI key mapping"

#### Purpose of change
The codebase contains dozens of different implementations of basic UI navigation such as going up and down in a list, or left/right or otherwise between tabs or categories.

#### Describe the solution
The primary change in this PR is to replace a significant number of disparate such implementations with a few new functions. First there are two small functions with various templated overloads, `increment_and_wrap` is effectively `(line+delta)%num_lines` but more robust, while `increment_and_clamp` ditto `std::clamp(line+delta,0,num_lines)`, both of which can be used to drop in for simple one-dimensional navigation of a collection of things (rows, lines, entries, tabs, categories, etc). In addition to standing alone, they also serve as the basis for `navigate_ui_list` which more broadly handles input context action strings, multi-line jumps such as for page_up and page_down, optional wrapping, etc.

The existing code could be left alone. A much larger refactor could be done, replacing more widely differing existing implementations.

#### Testing
I've tried to poke around in each of the UIs affected by this PR. Also as long as it stays in draft status I'll be playing the game with it.

#### Additional context
Along the way I fixed a few bugs, added and changed a bit of key mapping related to scrolling and changing tabs (hence the changlog entry), and made a few subjective stylistic choices.

### DIFF
--- a/src/armor_layers.cpp
+++ b/src/armor_layers.cpp
@@ -673,6 +673,8 @@ void outfit::sort_armor( Character &guy )
     ctxt.register_action( "QUIT" );
     ctxt.register_action( "PREV_TAB" );
     ctxt.register_action( "NEXT_TAB" );
+    ctxt.register_action( "PAGE_UP" );
+    ctxt.register_action( "PAGE_DOWN" );
     ctxt.register_action( "MOVE_ARMOR" );
     ctxt.register_action( "CHANGE_SIDE" );
     ctxt.register_action( "TOGGLE_CLOTH" );
@@ -996,24 +998,16 @@ void outfit::sort_armor( Character &guy )
                     popup( _( "Can't sort this under your integrated armor!" ) );
                 }
             }
-        } else if( action == "LEFT" ) {
+        } else if( action == "LEFT" || action == "PREV_TAB" || action == "RIGHT" || action == "NEXT_TAB" ) {
             mid_pane.offset = 0;
-            tabindex--;
-            if( tabindex < 0 ) {
-                tabindex = tabcount - 1;
-            }
+            tabindex = increment_and_wrap( tabindex, action == "RIGHT" || action == "NEXT_TAB", tabcount );
             leftListIndex = leftListOffset = 0;
             selected = -1;
-        } else if( action == "RIGHT" ) {
-            mid_pane.offset = 0;
-            tabindex = ( tabindex + 1 ) % tabcount;
-            leftListIndex = leftListOffset = 0;
-            selected = -1;
-        } else if( action == "NEXT_TAB" ) {
+        } else if( action == "PAGE_DOWN" ) {
             if( rightListOffset + rightListLines < rightListSize ) {
                 rightListOffset++;
             }
-        } else if( action == "PREV_TAB" ) {
+        } else if( action == "PAGE_UP" ) {
             if( rightListOffset > 0 ) {
                 rightListOffset--;
             }
--- a/src/auto_note.cpp
+++ b/src/auto_note.cpp
@@ -330,20 +330,18 @@ void auto_note_manager_gui::show()
     int startPosition = 0;
     int endPosition = 0;
 
-    input_context ctx{ "AUTO_NOTES" };
-    ctx.register_action( "QUIT" );
-    ctx.register_action( "SWITCH_AUTO_NOTE_OPTION" );
-    ctx.register_action( "HELP_KEYBINDINGS" );
-    ctx.register_action( "NEXT_TAB" );
+    input_context ctxt{ "AUTO_NOTES" };
+    ctxt.register_action( "QUIT" );
+    ctxt.register_action( "SWITCH_AUTO_NOTE_OPTION" );
+    ctxt.register_action( "HELP_KEYBINDINGS" );
+    ctxt.register_action( "NEXT_TAB" );
     if( !char_emptyMode || !global_emptyMode ) {
-        ctx.register_cardinal();
-        ctx.register_action( "PAGE_UP", to_translation( "Fast scroll up" ) );
-        ctx.register_action( "PAGE_DOWN", to_translation( "Fast scroll down" ) );
-        ctx.register_action( "CONFIRM" );
-        ctx.register_action( "QUIT" );
-        ctx.register_action( "ENABLE_MAPEXTRA_NOTE" );
-        ctx.register_action( "DISABLE_MAPEXTRA_NOTE" );
-        ctx.register_action( "CHANGE_MAPEXTRA_CHARACTER" );
+        ctxt.register_navigate_ui_list();
+        ctxt.register_action( "CONFIRM" );
+        ctxt.register_action( "QUIT" );
+        ctxt.register_action( "ENABLE_MAPEXTRA_NOTE" );
+        ctxt.register_action( "DISABLE_MAPEXTRA_NOTE" );
+        ctxt.register_action( "CHANGE_MAPEXTRA_CHARACTER" );
     }
 
     ui.on_redraw( [&]( const ui_adaptor & ) {
@@ -384,7 +382,7 @@ void auto_note_manager_gui::show()
 
         // TODO: Show info about custom symbols (hotkey, hint, state)
         std::string header_info_custom_symbols = string_format( _( "<color_light_green>%1$s</color> %2$s" ),
-                ctx.get_desc( "CHANGE_MAPEXTRA_CHARACTER" ), _( "Change a symbol for map extra" ) );
+                ctxt.get_desc( "CHANGE_MAPEXTRA_CHARACTER" ), _( "Change a symbol for map extra" ) );
         // NOLINTNEXTLINE(cata-use-named-point-constants)
         fold_and_print( w_header, point( 0, 1 ), FULL_SCREEN_WIDTH - 2, c_white,
                         header_info_custom_symbols );
@@ -480,15 +478,15 @@ void auto_note_manager_gui::show()
     while( true ) {
         ui_manager::redraw();
 
-        const std::string currentAction = ctx.handle_input();
+        const std::string action = ctxt.handle_input();
 
         // Actions that also work with no items to display
-        if( currentAction == "SWITCH_AUTO_NOTE_OPTION" ) {
+        if( action == "SWITCH_AUTO_NOTE_OPTION" ) {
             get_options().get_option( "AUTO_NOTES_MAP_EXTRAS" ).setNext();
             get_options().save();
-        } else if( currentAction == "QUIT" ) {
+        } else if( action == "QUIT" ) {
             break;
-        } else if( currentAction == "NEXT_TAB" ) {
+        } else if( action == "NEXT_TAB" ) {
             bCharacter = !bCharacter;
         }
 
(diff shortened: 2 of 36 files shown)


src/armor_layers.cpp
```diff
@@ -996,24 +998,16 @@ void outfit::sort_armor( Character &guy )
...
+            tabindex = increment_and_wrap( tabindex, action == "RIGHT" || action == "NEXT_TAB", tabcount );
             leftListIndex = leftListOffset = 0;
             selected = -1;
-        } else if( action == "NEXT_TAB" ) {
+        } else if( action == "PAGE_DOWN" ) {
             if( rightListOffset + rightListLines < rightListSize ) {
                 rightListOffset++;
             }
-        } else if( action == "PREV_TAB" ) {
+        } else if( action == "PAGE_UP" ) {
```
sparr, line 1010: I believe this more closely matches the norms in other UIs with multiple columns to scroll independently.

src/auto_note.cpp
```diff
@@ -330,20 +330,18 @@ void auto_note_manager_gui::show()
...
-        ctx.register_action( "QUIT" );
-        ctx.register_action( "ENABLE_MAPEXTRA_NOTE" );
-        ctx.register_action( "DISABLE_MAPEXTRA_NOTE" );
-        ctx.register_action( "CHANGE_MAPEXTRA_CHARACTER" );
+        ctxt.register_navigate_ui_list();
+        ctxt.register_action( "CONFIRM" );
+        ctxt.register_action( "QUIT" );
+        ctxt.register_action( "ENABLE_MAPEXTRA_NOTE" );
+        ctxt.register_action( "DISABLE_MAPEXTRA_NOTE" );
+        ctxt.register_action( "CHANGE_MAPEXTRA_CHARACTER" );
```
sparr, line 344: This rename is for consistency with other UI source files, the vast majority of which use this consistent name.

src/auto_note.cpp
```diff
@@ -503,41 +501,14 @@ void auto_note_manager_gui::show()
...
-        } else if( currentAction == "PAGE_UP" ) {
-            if( currentLine == 0 ) {
-                currentLine = cacheSize - 1;
-            } else if( currentLine <= scroll_rate ) {
-                currentLine = 0;
-            } else {
-                currentLine += -scroll_rate;
-            }
-        }  else if( currentAction == "ENABLE_MAPEXTRA_NOTE" ) {
+        if( navigate_ui_list( action, currentLine, scroll_rate, cacheSize, true ) ) {
```
sparr, line 504: The replacement above, like the many similar instances later, is the original and primary purpose of this PR.

src/auto_pickup.cpp
```diff
@@ -357,7 +356,7 @@ void user_interface::show()
         }
         mvwprintz( w_header, point( 1, 3 ), c_white, "#" );
         mvwprintz( w_header, point( 8, 3 ), c_white, _( "Rules" ) );
-        mvwprintz( w_header, point( 52, 3 ), c_white, _( "I/E" ) );
+        mvwprintz( w_header, point( 52, 3 ), c_white, _( "Inc/Exc" ) );
```
sparr, line 359: The column is this wide anyway to fit "Include" and "Exclude" in lower rows.

src/debug_menu.cpp
```diff
@@ -1017,7 +1018,7 @@ static void change_spells( Character &character )
             mvwprintz( w_name.window, point( 2, line_number ),
                        spell_color, splt.name.translated() );
             mvwprintz( w_level.window, point( 2, line_number++ ), spell_color,
-                       _( "%1$-3d/%2$3d" ), spell_level, splt.max_level );
+                       _( "%1$2d / %2$2d" ), spell_level, splt.max_level );
```
sparr, line 1021: right aligned numbers are easier to parse visually, but this leaves 3-digit spell levels (or negative with two digits) looking odd.

src/diary_ui.cpp
```diff
@@ -168,10 +168,13 @@ void diary::show_diary_ui( diary *c_diary )
     enum class window_mode : int {PAGE_WIN = 0, CHANGE_WIN, TEXT_WIN, NUM_WIN, FIRST_WIN = 0, LAST_WIN = NUM_WIN - 1};
     window_mode currwin = window_mode::PAGE_WIN;
 
-    std::map<window_mode, int> selected = { {window_mode::PAGE_WIN, 0}, {window_mode::CHANGE_WIN, 0}, {window_mode::TEXT_WIN, 0} };
+    std::map<window_mode, int> selected = { {window_mode::PAGE_WIN, std::max<size_t>( 0, c_diary->pages.size() - 1 )}, {window_mode::CHANGE_WIN, 0}, {window_mode::TEXT_WIN, 0} };
```
sparr, line 171: diary view will start on the latest page instead of the first page

src/distraction_manager.cpp
```diff
@@ -147,72 +132,16 @@ void distraction_manager_gui::show()
...
-                case 12:
-                    uistate.distraction_mutation = !uistate.distraction_mutation;
-                    break;
-                default:
-                    return;
-            }
+        if( navigate_ui_list( action, currentLine, 5, number_of_distractions, true ) ) {
+        } else if( action == "CONFIRM" ) {
+            *std::get<0>( configurable_distractions[currentLine] ) = !*std::get<0>
+                    ( configurable_distractions[currentLine] );
```
sparr, line 144: Should I put it in a local reference? Or just `^=1` it? Not sure of our style preferences for something like this.

src/gamemode_defense.cpp
```diff
@@ -500,16 +498,15 @@ void defense_game::setup()
     } );
     ui.mark_resize();
 
-    int selection = 1;
-    int selection_max = 20;
+    int selection = 0;
```
sparr, line 501: changing to zero based indexing caused a lot of churn below, but also simplified the arithmetic and is more consistent with other UI implementations

src/gamemode_defense.cpp
```diff
@@ -720,58 +664,58 @@ void defense_game::setup()
 
 void defense_game::refresh_setup( const catacurses::window &w, int selection )
 {
+    // *INDENT-OFF* to preserve column alignment and human readability
```
sparr, line 667: I see this used a few other places to keep "columns" of code aligned. I hope it's acceptable here.

src/diary_ui.cpp
Qrox: Won't the current code cause it to underflow to the maximum unsigned `size_t` when the size is zero?

src/player_display.cpp
```diff
@@ -1143,74 +1131,61 @@ static bool handle_player_display_action( Character &you, unsigned int &line,
         }
     };
 
-    unsigned int line_beg = 0;
-    unsigned int line_end = 0;
+    unsigned int line_count = 0;
     switch( curtab ) {
         case player_display_tab::stats:
-            line_end = 8;
+            line_count = 9;
```
sparr, line 1137: bug fix, couldn't previously reach the blood type line

src/ui.h
sparr, line 1: I suspect there's a more concise way to do all the templating and overloading in this file.

src/gamemode_defense.cpp
```diff
@@ -861,9 +805,9 @@ std::string defense_location_description( defense_location location )
         case DEFLOC_NULL:
             return "NULL Bug.  (defense.cpp:defense_location_description)";
         case DEFLOC_HOSPITAL:
-            return                 _( "One entrance and many rooms.  Some medical supplies." );
+            return                 _( "One entrance and many rooms. Some medical supplies." );
```
irwiss, line 808: These double spaces after `.` in sentences are on purpose for monospace text - [there's some debates](https://en.wikipedia.org/wiki/Sentence_spacing) but that's the style cata uses, you'll see them everywhere

src/newcharacter.cpp
```diff
@@ -2396,9 +2346,9 @@ void set_hobbies( tab_manager &tabs, avatar &u, pool_type pool )
         }
 
         //Draw options
-        calcStartPos( iStartPos, cur_id, iContentHeight, profs_length );
-        const int end_pos = iStartPos + ( ( iContentHeight > profs_length ) ?
-                                          profs_length : iContentHeight );
+        calcStartPos( iStartPos, cur_id, iContentHeight, hobbies_length );
+        const int end_pos = iStartPos + ( ( iContentHeight > hobbies_length ) ?
```
irwiss, line 2350: seems equivalent to std::min

src/ui.h
```diff
@@ -535,4 +535,145 @@ class pointmenu_cb : public uilist_callback
...
+{
+    if( size == 0 ) {
+        return 0;
+    }
+    if constexpr( std::is_unsigned_v<V> ) {
+        if( !inc && val == 0 ) {
+            return size ? size - 1 : 0;
+        }
+    }
+    return modulo( inc ? ( val + 1 ) : ( val - 1 ), size );
```
irwiss, line 554: These expressions (and a few similar below) seem more readable in this way, but probably a matter of taste
```suggestion
    return modulo( val + ( inc ? +1 : -1 ), size );
```

irwiss (commented):
That seems to cut down a lot of repetition, good job

src/auto_pickup.cpp
Brambor: Incl/Ex might be better
I read it as increment / execute (my dyslectic brain exc->exe).

src/player_display.cpp
```diff
@@ -1104,6 +1080,18 @@ static void draw_tip( const catacurses::window &w_tip, const Character &you,
     wnoutrefresh( w_tip );
 }
 
+inline void skip_skill_headers( const std::vector<HeaderSkill> &skillslist, unsigned int &line,
+                                bool inc, unsigned int line_count )
+{
+    const unsigned int prev_line = line;
+    while( skillslist[line].is_header ) {
```
Brambor, line 1087: Could this be an endless loop if the only lines are header lines?
Or if line = line_count = 0? - Actually, wouldn't it call for undefined skillslist[0]?

src/ui.h
Brambor: I am making commit improving it a bit.

Brambor: I made a PR into your fork improving it somewhat:
https://github.com/sparr/Cataclysm-DDA/pull/4

sparr: Thank you. I have some further changes to push, then I'll see if I can merge some or all of your PR before setting this for review.

src/gamemode_defense.cpp
sparr: The string was one character too long. I'll find a different way to shorten it.

src/player_display.cpp
sparr: Not an endless loop. The conditional break at the bottom of the loop will end it after one cycle.
Good point re line_count == 0, but this is one of many places where the original code has such* a problem and I don't want to try to fix them all in this PR.

\* code paths that could have bugs with arbitrary input but work with every input the game currently produces. In this case, the skillslist is never empty (further, it always starts with a header item, which some other old code takes advantage of and would break if it didn't)

src/diary_ui.cpp
```diff
@@ -168,10 +168,14 @@ void diary::show_diary_ui( diary *c_diary )
     enum class window_mode : int {PAGE_WIN = 0, CHANGE_WIN, TEXT_WIN, NUM_WIN, FIRST_WIN = 0, LAST_WIN = NUM_WIN - 1};
     window_mode currwin = window_mode::PAGE_WIN;
 
-    std::map<window_mode, int> selected = { {window_mode::PAGE_WIN, 0}, {window_mode::CHANGE_WIN, 0}, {window_mode::TEXT_WIN, 0} };
+    const int diary_last_page = c_diary->pages.size() ? c_diary->pages.size() - 1 : 0;
```
kevingranade, line 171:
```suggestion
    const int diary_last_page = c_diary->pages.empty() ? 0 : ( c_diary->pages.size() - 1 );
```

src/distraction_manager.cpp
```diff
@@ -15,27 +15,37 @@
...
-    {translate_marker( "Thirst" ),                       translate_marker( "This distraction will interrupt your activity when you're dangerously dehydrated." )},
-    {translate_marker( "Temperature" ),                  translate_marker( "This distraction will interrupt your activity when your temperature is very high or very low." )},
-    {translate_marker( "Mutation" ),                     translate_marker( "This distraction will interrupt your activity when you gain or lose a mutation." )},
+struct configurable_distraction {
+    bool *state;
+    std::string name;
+    std::string description;
 };
 
+const std::vector<configurable_distraction> &get_configurable_distractions()
```
kevingranade, line 24:
```suggestion
static const std::vector<configurable_distraction> &get_configurable_distractions()
```

PR #61087, opened 2022-09-18
Lock vehicle doors

Numbuh474:
#### Summary
Features "Make vehicle doors lockable and pickable"

#### Purpose of change

This is a contribution to #51383, addressing the ability to lock and unlock vehicle doors.
It also fixes #46160.

#### Describe the solution

Enable locking and unlocking of certain vehicle doors, adding the `LOCKABLE_DOOR` flag for valid doors and the `DOOR_LOCKING` flag for parts that allow locking and unlocking. A new item, the door lock, was given the DOOR_LOCKING flag and most metal doors and hatches were given the `LOCKABLE_DOOR` tag. A door must be a LOCKABLE_DOOR and have a part with the DOOR_LOCKING tag on the same frame to enable the feature. Modify relevant door opening code to respect locked vehicle door status for player and NPCs. Add lockpicking behavior for vehicle doors, allowing lockpick tools to pick open doors that are otherwise inaccessible.

Monsters and door-opening ferals aren't able to figure out locks, but NPCs will path through locked doors if given the option.  A new pair of behavior flags were added that allows you to make followers stop unlocking doors, or to also lock them if possible.

Behavior that protected door motors from damage if the door was closed was added to door locks and moved to the more generic `BOARD_INTERNAL` flag, which now also works with things installed in boards in case someone wants to add pipework or something. Doors and their locks can't be removed while a tile on the same square is in the "locked" state.

Door motor menu was modified to allow you to lock and unlock all closed doors if the door is lockable, and examining a tile with a lockable door now has options for modifying the lock state based on player position.

#### Describe alternatives you've considered

Doors could have been made intrinsically lockable without a corresponding DOOR_LOCKING part.
Part lock state currently has nothing to do with vehicle lock state, and these could be made to work together to do things like lock doors on vehicle spawn or trigger alarms. 
Lockpicking could have been excluded or made impossible for player-owned vehicles, as this gives "free" locked doors for training purposes.

#### Testing

Tested door opening and closing from inside and outside the vehicle, door locking and unlocking from both inside and outside the vehicle, door-opening monster behavior for closed vs. locked doors, and locking and unlocking via the door motor menu.

Tested having an NPC follow you into and out of a vehicle with a door lock while using the new rule toggles.

PR #64464, opened 2023-03-22
Turn person's ability to open doors back

GuardianDll:
#### Summary
None
#### Purpose of change/Describe the solution
Revert #63053
#### Additional context
Also should be removed from https://github.com/CleverRaven/Cataclysm-DDA/projects/59

### DIFF
--- a/data/json/monsters/nether.json
+++ b/data/json/monsters/nether.json
@@ -1406,7 +1406,7 @@
     "morale": 100,
     "vision_day": 60,
     "vision_night": 60,
-    "path_settings": { "max_dist": 60 },
+    "path_settings": { "max_dist": 60, "allow_open_doors": true },
     "death_function": { "corpse_type": "NO_CORPSE", "message": "it just disappeared." },
     "flags": [
       "NO_BREATHE",


KHeket:
May be its better to wait, until the https://github.com/CleverRaven/Cataclysm-DDA/pull/61087 will be merged? Because of persons, who can open doors - portal storms become deadly, if you are hiding in car or in houses without ability to block intrances with furniture, unkillable person will open door - and then portal monsters will go in, and I dont know, was it fixed, but few memory monsters can grab you and suffocate to dead. And there is no logic, that player cant prevent his own car doors from opening by the strange portal monster

RAldrich:
What's the intent behind this change? As pointed out, it dramatically increases the danger of portal storms, and that's not super fun given that we don't have mechanics in place to allow players to lock doors shut (or otherwise counter the Person in any way).

GuardianDll:
The idea is exactly this: make the portal storm more dangerous, in a reality breaking incident you can't just close the door and call it a day - you should prepare yourself, you should find the better place to hide, you should go underground, you may even abandon your vehicle for a while just to be safe
Sure, it will annoy people for a while, untill someone find and implement the door locking mechanism for terrain (and merge the vehicle doors stuff), but it doesn't mean we should hold it
Honestly i can't even say i like the current implementation, because you still can just put few bookcases and it will effectively prevent them to reach you

KHeket:
> The idea is exactly this: make the portal storm more dangerous, in a reality breaking incident you can't just close the door and call it a day - you should prepare yourself, you should find the better place to hide, you should go underground, you may even abandon your vehicle for a while just to be safe Sure, it will annoy people for a while, untill someone find and implement the door locking mechanism for terrain (and merge the vehicle doors stuff), but it doesn't mean we should hold it Honestly i can't even say i like the current implementation, because you still can just put few bookcases and it will effectively prevent them to reach you

Portal storm are already dangerous
1) Didnt find a good shelter in time - you will die (if there is no good funiture to block entrance - you will die)
2) Zombie attack at the shelter/car - you will die (hole in the wall of house or car - death)
3) If this pr will be merged before door locks - you will die, even if you have a good car. But in reality you can at least solve it by simple door bolt
There is a lot of frequenly spawning portal monsters, memory suffocaters and unkillable (why they are immortal and can open doors - I dunno) - so there is no need for now for this pr. And zombies are also became more deadly because of grab upgrage, so there is no reason to make game unrealistic harder

RAldrich:
The concern I have is that there is NO way to survive a portal storm while outside. You cannot defeat the storm monsters. The only way to engage with the storm right now is to hide indoors.

Adding this back in removes the "hide inside" option and makes the storm entirely fatal with nothing the player can do about it. The only way to survive a storm will be to hide in a basement (with z layer vision options disabled) or to use furniture to barricade the doors. This is both not fun and not realistic.

If you agree that being able to lock doors would be an acceptable countermeasure to the Person, why not wait until that is implemented, rather than making storms completely fatal in the meantime?

Speaking more generally, I don't think making portal storms more dangerous is going to help, I think we need more mechanisms to engage with the storm. Right now they're already maximally dangerous (especially with memories getting the grab strength buff), so hide or die is the only player choice. If anything they should probably be LESS dangerous so that going out to take a risk in one is a potential choice to be made.

GuardianDll:
> The only way to engage with the storm right now is to hide indoors.

It is actually work with any weather - you can't fight a storm, you can only hide

> why not wait until that is implemented

Person get the ability -> someone is pissed enough to implement a door lockage -> it goes into the game -> profit
Person do not get the ability -> no one is pissed off -> no one implement a door lockage -> no profit
(btw, ability to nail doors and windows are still here, but i found 0 mentions of it anywhere in the discussion of the "portal storm is too dangerous" theme)

> I don't think making portal storms more dangerous is going to help

Sure, the target was never to make portal storm harder, it was turning back a one interesting gimmick, that was removed specifically for better stable experience (same as was removed, in this case entirely, the ability to get one free resurrection in the portal dungeon)
FYI Ramza (and me, but not as much) is interested in making portal storm a better experience, and we would like to hear some good ideas how it can be made much more interesting and less annoying

KHeket:
> Person get the ability -> someone is pissed enough to implement a door lockage -> it goes into the game -> profit
> Person do not get the ability -> no one is pissed off -> no one implement a door lockage -> no profit

Ability to lock the doors was asked 1.5 year ago because of human animies able to open doors, and 1.5 year ago it was much more easier to repair you vehicle, you didnt need to waste resourses and there was no degradation of car parts, so players used their vahicles as zombie smashers. There is still a need for door locks because of mi go, bandit spawning npc and other cases, when anemies can open your car

> (btw, ability to nail doors and windows are still here, but i found 0 mentions of it anywhere in the discussion of the "portal storm is too dangerous" theme)

Nail window costs 20 minutes, nail doors 10 minutes. You also need to find resorses, and there can be more than 1-2 doors or broken windows, so its costs 20-40+ minutes (or more)- enough time to portal storm begun and let persons open doors. And have base, where you must nail all doors because of this for no reason unkillable monsters - not very good idea. And building at your base litlle room/shelter with only one door speacial for portal storm - not good idea too, very unrealistic. And we cant also nail metal door (because we need door bolt or locks mechanic)

Its okay, if this ability of persons will be added after door lock mechanic, but for now it just will let players to die because of missing mechanic or they will use bugs or tricks, like abusing furniture

RAldrich:
I think the difference is that none of the other weather is actively trying to kill you. Even a blizzard is perfectly survivable with the right preparation.

I take your point about this being a forcing function to incentivize someone to work on the lock system.

Regarding improvements to storms in general, this PR probably isn't the best place to record them, but my general impression is what I've said above: it needs more player interaction. Maybe we should start thinking about it less like weather and more like a timed special event that involves dimensional monsters showing up? Maybe make some of the monsters persist, as a bridge towards giving players a reason to build fortified bases with traps.

AcidAntOnAMinefield:
> Person get the ability -> someone is pissed enough to implement a door lockage -> it goes into the game -> profit

This is honestly rather insulting. 

So you really think that aggravating your playerbase with the offchance that someone will get pissed off to the point of wanting to create an entire new feature to deal with a problem you created out of spite is a good idea?

PR #64465, opened 2023-03-22
Throwable fire extinguishers can now be placed in grenade-only pockets

pjf:
#### Summary

Balance "Throwable fire extinguishers can now be placed in grenade-only pockets"

#### Purpose of change

Throwable fire extinguishers are "A fire extinguisher in grenade form". They're used just like grenades, except for fighting fires instead of zombies. It makes sense they can go in grenade-only pockets.

#### Describe the solution

Add the `GRENADE` flag to the item's JSON.

None.

#### Testing

Loaded a game with the changes applied. Added a throwable fire extinguisher to a grenade-only pocket when I couldn't before.

#### Additional context

None

### DIFF
--- a/data/json/items/tool/firefighting.json
+++ b/data/json/items/tool/firefighting.json
@@ -155,7 +155,7 @@
       "active": true,
       "type": "transform"
     },
-    "flags": [ "ACT_IN_FIRE", "BOMB" ]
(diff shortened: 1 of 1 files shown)


pjf:
Thanks for the quick merge!

PR #64469, opened 2023-03-22
[Magiclyms] Make black dragons Stylish

Terrorforge:
#### Summary
Content "Make black dragons Stylish"

#### Purpose of change
Dragons famously hoard treasure. This is true in Magiclysm as well, as dragon lairs contain both powerful magical items and mundane but shiny trinkets. However, there's no incentive for dragon mutants to do this outside of standard looting and/or pure roleplaying.

#### Describe the solution
Add the Stylish trait to the Black Dragon category. This trait rewards you for decking yourself out in gold jewelry and other expensive clothing, which gives you at least a little bit of incentive to knock over banks and jewelers as a proper dragon should.

Make a new trait that's a little more appropriately flavored, like maybe it's just gold and jewels that count. However, that would require a much larger intervention. Even just doing a version of Stylish with more dragon-y flavor text isn't feasible, as the functionality is hardcoded.

I could also just not add even more traits to what is already a very bloated tree, but I got really sad when my (prospective) dragon mutant found some really nice jewelry and I realized there wasn't even a token mechanical reward for roleplaying as Smaug.

### DIFF
--- a/data/mods/Magiclysm/mutations/mutations.json
+++ b/data/mods/Magiclysm/mutations/mutations.json
@@ -11,6 +11,12 @@
   },
   {
     "type": "mutation",
+    "id": "STYLISH",
+    "copy-from": "STYLISH",
+    "extend": { "category": [ "DRAGON_BLACK" ] }
+  },
+  {
+    "type": "mutation",
     "id": "PAINRESIST",
     "copy-from": "PAINRESIST",
     "extend": { "category": [ "DRAGON_BLACK" ] }


kevingranade:
> Make a new trait that's a little more appropriately flavored

Seems like a good way to go. @KorGgenT has the final say of course,  but it seems to me "piles of treasure in a cave" to "wearing fancy clothes" is quite a leap.

Terrorforge:
It's not that much of a leap if you consider that "fancy clothes" generally means "expensive clothes". Double especially because the majority of (super) fancy items, especially the ones that a Huge dragon can actually wear, is jewelry. There's admittedly some wonk, though. Notably, quite a few costume pieces are considered "fancy".

The other reason I went with Stylish is that wearing an item is one of the few meaningful metrics for whether that item actually belongs to you. "Owned" items may have just been picked up and dropped. It would be really cool if we had some sort of lair mechanics where you actually got bonuses based on how much crap you piled up in your home, but I'm not sure how you'd define "home". The only other alternative I can think of is I guess giving bonuses based on the weight/volume of items made of gems and precious metals you have on your person? You could base it on price I guess, but that could get real wonky real fast since there's some very big price differences in this game.

PR #64470, opened 2023-03-22
Remove achievements/scores/killcount top-level keybinding and move it into the diary menu (bound by default to ')')

### LINKED ISSUE #59685: Remove top-level keybinding for kill count

This is part of a larger effort to identify places where our "top level" keybindings have been mapped to things that players do not need rapid, single-keypress access for. This is a discussion: all these issues are valid places to raise concerns about reasons a player should have immediate single-key access to this function.

"Kill count" is a purely for-interest feature, and not one that needs to be accessed rapidly as a single keypress.

### Solution you would like.

I suggest moving Kill Count into the Diary - `K: view kill count` for example. 

Then, let's bind the Diary to `(` because it's a very nice feature that is currently not accessible through default keybindings and many people don't know about it.

The kill count could go in the `@` menu instead but I like the conceptual idea of keeping the diary to "fluff" stuff and the @ menu for "affects the game directly" stuff.

### Additional context

Ideally we'd also revamp the appearance of the Kill Count screen to match the diary page display but in the short term, it could just pop up overtop of the diary looking like it currently does.

### LINKED ISSUE #59686: Achievements do not need to be a top level keybinding

This is part of a larger effort to identify places where our "top level" keybindings have been mapped to things that players do not need rapid, single-keypress access for. This is a discussion: all these issues are valid places to raise concerns about reasons a player should have immediate single-key access to this function.

Achievements don't need to be accessed frequently, and so can be relegated to another submenu.

### Solution you would like.

Move the achievements key into Diary - the same as I have recommended for Kill Count, in https://github.com/CleverRaven/Cataclysm-DDA/issues/59685. In this context, Diary would now be mapped to a root level key like `(` giving a convenient single stop shop for all your fluff/lore needs. So while in diary, you can press eg. `A` to view your achievements.

Achievements aren't in-character lore so there's a decent argument to instead put them in the ESC menu, but I think they fit fine in the diary myself

### Additional context

Ideally we'd also revamp the appearance of the achievements screen to match the diary page display but in the short term, it could just pop up overtop of the diary looking like it currently does.

strategineer:
#### Summary
None

#### Purpose of change

See issues below for more details.

Fixes #59686
Fixes #59685 

#### Describe the solution

- Remove global "scores" keybind from code and replace it with a VIEW_SCORES keybind accessible only from the diary menu.
- Re-use "scores" keybind's default use of ')' for the key to access the diary by default.
- Implement VIEW_SCORES keybind and hint at its use from the diary menu.

None.

#### Testing

- Checked changed keybinds and ensured that they appeared in the right places and worked as expected.

#### Additional context

diary menu:
WindowsTerminal_nMG8Yy5khk

diary keybindings:
WindowsTerminal_hcPQPGeEah

achievements/scores/killcount menu showing overlaid on top of diary menu:
WindowsTerminal_VMXz7lZDck

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -2611,13 +2611,6 @@
   },
   {
     "type": "keybinding",
-    "name": "View achievements, scores, and kills",
-    "category": "DEFAULTMODE",
-    "id": "scores",
-    "bindings": [ { "input_method": "keyboard_char", "key": ")" }, { "input_method": "keyboard_code", "key": "0", "mod": [ "shift" ] } ]
-  },
-  {
-    "type": "keybinding",
     "name": "View morale",
     "category": "DEFAULTMODE",
     "id": "morale",
@@ -2986,7 +2979,8 @@
     "type": "keybinding",
     "id": "diary",
     "name": "Open diary",
-    "category": "DEFAULTMODE"
+    "category": "DEFAULTMODE",
+    "bindings": [ { "input_method": "keyboard_char", "key": ")" }, { "input_method": "keyboard_code", "key": "0", "mod": [ "shift" ] } ]
   },
   {
     "type": "keybinding",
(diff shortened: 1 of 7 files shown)


bombasticSlacks:
@I-am-Erk the code for this looks fine to me, but it's your issue, so figured you should take a look if this is what you wanted, or is at least a step in the right direction.

kevingranade:

https://discourse.cataclysmdda.org/t/game-continually-crashes-when-creature-dies/28632/6

Issue #58846, opened 2022-06-30
Keybinding display consistency


ZeroInternalReflection:
### Is your feature request related to a problem? Please describe.

Many of the UI screens in CDDA show keybindings for common actions, which is great.  However, there is very little consistency in how those keybindings are shown, meaning it's not always clear if something is a keybinding, or if that keybinding works in the current context.  For example...
Sometimes the keybindings are described with words rather than the actual key name:
![Crafting - Filter screen -- 103x34](https://user-images.githubusercontent.com/89038572/167521440-57208844-5756-4832-87fe-9ec5ae80cf9c.png)

(In this case, if the user changes the keybinding for this action, the hint is now incorrect.  It also means that the keybinding needs to be translated)

Sometimes the keybindings are flagged by being in a different colour, such as:
Green
![Actions - Combat -- 80x24](https://user-images.githubusercontent.com/89038572/167330265-48639872-96d2-4fe5-8a9b-c2b830f719f8.png) 

Yellow:
![Base camp - Inspect NPC -- 80x24](https://user-images.githubusercontent.com/89038572/169736605-09f3cb07-30d9-4745-8af1-4a466bedd82c.png)

Blue:
![Main menu - Help - Details -- 80x24](https://user-images.githubusercontent.com/89038572/166125099-c20d3392-5708-4d26-8c82-d4a3aa6c86d6.png)

White:
![Debug error prompt -- 80x24](https://user-images.githubusercontent.com/89038572/167521441-fda560e6-2e53-4ace-a216-bedc873e618d.png)

Sometimes the keybindings are left to match the adjacent text, in:
White: 
![Activate (a) -- 80x24](https://user-images.githubusercontent.com/89038572/166401118-34114026-4ca3-47a9-9f4a-9c32f32e8613.png)

Dark Gray (shows as magenta in my configuration):
![Activate - Gun modification - Risk damage -- 80x24](https://user-images.githubusercontent.com/89038572/170834473-dab1fcbc-2de8-4044-803a-c3def9681c2a.png)

Yellow:
![Examine vehicle - Crew -- 80x24](https://user-images.githubusercontent.com/89038572/169736589-d183c1c3-ef8d-49da-88a2-81ded29d0326.png)

Green:
![Look around (x) -- 80x24](https://user-images.githubusercontent.com/89038572/166250088-c351c1ed-286a-467b-a811-d9600f5c6195.png)

Red:
[...]

Issue #58776, opened 2022-06-27
Panels Block Minimap When Minimap Is Intended to Be Displayed


TheMurderUnicorn:
### Describe the bug

When you peek using the `X` button, the sidebar changes to show whatever terrain you're looking at.  This blocks the minimap, at least when using the classic sidebar.  There's even a line that says you can press N to toggle the minimap but the sidebar blocks the map no matter what.

### Steps to reproduce

I'm using the classic sidebar layout, more or less the default stuff displayed.

Press X to begin peeking

Observe there is no minimap or that you can see a few pixels of it beneath the new side display.

Press N to toggle minimap, it collapses it in the sidebar but the other menu is still on top blocking it.

I am on an older version from last week but it was confirmed by people in discord as still occurring on the newest version as of my posting this.

### Expected behavior

Having your minimap while peeking is sort of a big deal.  We had this issue a few years ago and it was changed to include the option to toggle the minimap for this exact reason but that no longer works. I would really like my minimap while peeking. 

Additionally, there is really no reason for this other menu to be so long.  Even when looking at an NPC or tile with a lot of information, it does not fill to the bottom of the screen.  In my opinion it takes up too much space, which has led to this issue.

### Screenshots

You can see the sliver of minimap where the panel stops at the bottom.
![image](https://user-images.githubusercontent.com/41651793/176038497-25f552d8-3898-421a-bfcf-f935129bc9ab.png)

Pressing the N key which is listed there as collapsing the minimap simply lowers the panel a few pixels to cover where those few lines of minimap were previously displayed.
![image](https://user-images.githubusercontent.com/41651793/176038588-63933d67-5538-4a6e-bb45-33cc60ca39d0.png)

So I believe the intention is for the Minimap to be displayed, since it tries to show it and there's control for doing so listed here.  Also because that's what I want to believe because I really like seeing the minimap. x_x

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.1766 (21H2)
- Game Version: 008354a [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda]
]

### Additional context

This does also occur with other instances of those sidebar panels.  For example, if you use short range teleport via the debug menu it displays the same thing, blocking the minimap.

hexagonrecursion:
/confirmed

Note: the "toggle the minimap" key is not bound by default

- OS: Linux
[...]

PR #64478, opened 2023-03-22
Display keybindings in UI consistently and add options to allow players to customize how the keybindings are displayed

strategineer:
#### Summary
Interface "Display keybindings in UI consistently and add options to allow players to customize how the keybindings are displayed"

#### Purpose of change

F*xes most of #58846 but not all...

This PR will:
- Make the code cleaner.
- Improve UX by making it (more) clear which keybindings the player can press at any time in a way that they're more likely to recognize at a glance.
- Ensure that we never display hardcoded keybindings to the player to prevent the situation where the player would rebind some keys but they're hardcoded in the UI so we'll end up displaying wrong keybindings (bad UX).
- Allow players to customize how keybindings are displayed throughout the UI to suit their needs (eg. "[x] Examine", "x: Examine", "(x) Examine", "x - Examine", "E[x]amine").
- Make extensive use of `fold_and_print` for displaying keybindings to allow them to show up well (in theory) no matter what language they're displayed in.

#### Describe the solution

- Create some `input_context::get_hint*` methods built on the existing `get_desc` methods that provide a more complete interface for generating the strings to display.
- Replace all existing uses of `get_desc` outside of the input.cpp with calls to the more complete `get_hint` methods (greatly simplifying UI code in the process).
- Find hardcoded keybindings and replace the vast majority of them with calls to `get_hint*` methods (passing any relevant input_context object around as needed) so that the keybindings are displayed consistently and always display an actual keybinding that the player has set.
- Insert extra bits of code into the existing `shortcut_{print|text}` and `press_x` methods that are harder to refactor to make the keybindings generated by them appear consistently with the keybindings generated through `get_hint*` methods  (This refactoring can be tackled in a subsequent PR to make this one more reasonable to review).
- Add `input_context::get_hint_color*` methods (NB: these should probably be in the `color.h` file) that take a state as a parameter (like ENABLED, DISABLED, TOGGLED_ON, TOGGLED_OFF, HIGHLIGHTED) and try to use them consistently throughout the code as needed such that we have one single place where we can swap all keybindings-related colors around.
- Fix minor UI issues/inconsistencies while I was in the neighborhood.

None.

#### Testing

- Each changed menu (except the basecamp UI code because I don't know how to trigger it :() was tested at least once after making changes to ensure that nothing was broken.
- WIP I'm doing a second pass now of all the changed screens (most of them) to make sure nothing is broken but any help with testing this would be appreciated.

#### Additional context

To Review:
The majority of the important changes can be found in `input.*`, `options.cpp` and `output.*`. All the other changes stem from there and, for the most part, are relatively simple. I've attached screenshots for each changed UI screen, you can see them by scrolling down!

Here are some screenshots for different keybinding hint settings:

Default settings:
WindowsTerminal_I837azrnMt
WindowsTerminal_Wb4qL7nOkJ

settings example \#1:
WindowsTerminal_y03FKJNmFe
WindowsTerminal_w4UwxOrAWj

settings example \#2:
WindowsTerminal_4ZpIxUVmye
WindowsTerminal_PxqIkHzFyK

settings example \#3:
WindowsTerminal_TwlGKwnL2A
WindowsTerminal_Udqlq5im11

### DIFF
--- a/src/action.cpp
+++ b/src/action.cpp
@@ -536,16 +536,6 @@ std::string press_x( action_id act, const std::string &key_bound_pre,
     input_context ctxt = get_default_mode_input_context();
     return ctxt.press_x( action_ident( act ), key_bound_pre, key_bound_suf, key_unbound );
 }
-std::optional<std::string> press_x_if_bound( action_id act )
-{
-    input_context ctxt = get_default_mode_input_context();
-    std::string description = action_ident( act );
-    if( ctxt.keys_bound_to( description, /*maximum_modifier_count=*/ -1,
-                            /*restrict_to_printable=*/false ).empty() ) {
-        return std::nullopt;
-    }
-    return press_x( act );
-}
 
 action_id get_movement_action_from_delta( const tripoint &d, const iso_rotate rot )
 {
--- a/src/advanced_inv.cpp
+++ b/src/advanced_inv.cpp
@@ -841,8 +841,8 @@ void advanced_inventory::redraw_pane( side p )
         wattroff( w, c_white );
     }
     if( !filter_edit && !pane.filter.empty() ) {
-        std::string fsuffix = string_format( " %s", ctxt.get_hint( "RESET_FILTER" ) );
-        print_colored_text( w, point( getmaxx( w ) - utf8_width( fsuffix ) - 2, getmaxy( w ) - 1 ),
+        std::string fsuffix = ctxt.get_hint( "RESET_FILTER" );
+        print_colored_text( w, point( getmaxx( w ) - utf8_width( fsuffix, true ) - 2, getmaxy( w ) - 1 ),
                             fsuffix );
     }
     wnoutrefresh( w );
--- a/src/advanced_inv.cpp
+++ b/src/advanced_inv.cpp
@@ -676,8 +676,9 @@ int advanced_inventory::print_header( advanced_inventory_pane &pane, aim_locatio
 
             bcolor = in_vehicle ? c_light_blue :
                      area == data_location || all_brackets ? c_light_gray : c_dark_gray;
-            kcolor = area == data_location ? c_white : sel == data_location ||
-                     container_loc == data_location ? c_light_gray : c_dark_gray;
+            kcolor = area == data_location ? input_context::get_hint_color_for_key() : sel == data_location ||
+                     container_loc == data_location ? input_context::get_hint_color_for_key(
+                         keybinding_hint_state::HIGHLIGHTED ) : input_context::get_hint_color_for_key();
         }
         const std::string key = get_location_key( static_cast<aim_location>( i ) );
         const point p( squares[i].hscreen + point( ofs, 0 ) );
@@ -817,8 +818,8 @@ void advanced_inventory::redraw_pane( side p )
     }
     // draw a darker border around the inactive pane
     draw_border( w, active ? BORDER_COLOR : c_dark_gray );
-    mvwprintw( w, point( 3, 0 ), _( "< [%s] Sort: %s >" ), ctxt.get_desc( "SORT" ),
-               get_sortname( pane.sortby ) );
+    print_colored_text( w, point( 3, 0 ), string_format( _( "< %s: %s >" ), ctxt.get_hint( "SORT" ),
+                        get_sortname( pane.sortby ) ) );
     int max = square.max_size;
     if( max > 0 ) {
         int itemcount = square.get_item_count();
@@ -827,23 +828,22 @@ void advanced_inventory::redraw_pane( side p )
         mvwprintw( w, point( w_width / 2 - fmtw, 0 ), "< %d/%d >", itemcount, max );
     }
 
-    std::string fprefix = string_format( _( "[%s] Filter" ), ctxt.get_desc( "FILTER" ) );
+    std::string fprefix = ctxt.get_hint( "FILTER" );
     if( !filter_edit ) {
         if( !pane.filter.empty() ) {
-            mvwprintw( w, point( 2, getmaxy( w ) - 1 ), "< %s: %s >", fprefix, pane.filter );
+            print_colored_text( w, point( 2, getmaxy( w ) - 1 ), string_format( "< %s: %s >", fprefix,
+                                pane.filter ) );
         } else {
-            mvwprintw( w, point( 2, getmaxy( w ) - 1 ), "< %s >", fprefix );
+            print_colored_text( w, point( 2, getmaxy( w ) - 1 ), string_format( "< %s >", fprefix ) );
         }
     }
     if( active ) {
         wattroff( w, c_white );
     }
     if( !filter_edit && !pane.filter.empty() ) {
-        std::string fsuffix = string_format( _( "[%s] Reset" ), ctxt.get_desc( "RESET_FILTER" ) );
-        mvwprintz( w, point( 6 + utf8_width( fprefix ), getmaxy( w ) - 1 ), c_white,
-                   pane.filter );
-        mvwprintz( w, point( getmaxx( w ) - utf8_width( fsuffix ) - 2, getmaxy( w ) - 1 ), c_white, "%s",
-                   fsuffix );
+        std::string fsuffix = string_format( " %s", ctxt.get_hint( "RESET_FILTER" ) );
+        print_colored_text( w, point( getmaxx( w ) - utf8_width( fsuffix ) - 2, getmaxy( w ) - 1 ),
+                            fsuffix );
     }
     wnoutrefresh( w );
 }
@@ -1165,8 +1165,8 @@ void advanced_inventory::redraw_sidebar()
     Messages::display_messages( head, 2, 1, w_width - 1, head_height - 2 );
     draw_minimap();
     right_print( head, 0, +3, c_white, string_format(
-                     _( "< [<color_yellow>%s</color>] keybindings >" ),
-                     ctxt.get_desc( "HELP_KEYBINDINGS" ) ) );
+                     _( "< %s >" ),
+                     ctxt.get_hint( "HELP_KEYBINDINGS" ) ) );
     if( get_player_character().has_watch() ) {
         const std::string time = to_string_time_of_day( calendar::turn );
         mvwprintz( head, point( 2, 0 ), c_white, time );
@@ -1791,7 +1791,9 @@ void query_destination_callback::draw_squares( const uilist *menu )
         // always show storage option for vehicle storage, if applicable
         bool canputitems = menu->entries[i - 1].enabled && square.canputitems();
         nc_color bcolor = canputitems ? sel == loc ? h_white : c_light_gray : c_red;
-        nc_color kcolor = canputitems ? sel == loc ? h_white : c_dark_gray : c_red;
+        nc_color kcolor = canputitems ? sel == loc ? input_context::get_hint_color_for_key(
+                              keybinding_hint_state::HIGHLIGHTED ) : input_context::get_hint_color_for_key() :
+                          input_context::get_hint_color( keybinding_hint_state::TOGGLED_OFF );
         const point p( square.hscreen + point( ofs, 5 ) );
         mvwprintz( menu->window, p, bcolor, "%c", bracket[0] );
         wprintz( menu->window, kcolor, "%s", key );
--- a/src/armor_layers.cpp
+++ b/src/armor_layers.cpp
@@ -1172,10 +1172,10 @@ void outfit::sort_armor( Character &guy )
             }
         } else if( action == "USAGE_HELP" ) {
             const std::vector<std::string> help_strings = {
-                ctxt.get_hint_pair( "UP", "DOWN", _( "to scroll the left pane (list of items)." ) ),
-                ctxt.get_hint_pair( "SCROLL_ITEM_INFO_UP", "SCROLL_ITEM_INFO_DOWN", _( "to scroll the middle pane (item information)." ) ),
-                ctxt.get_hint_pair( "PREV_TAB", "NEXT_TAB", _( "to scroll the right pane (items grouped by body part)." ) ),
-                ctxt.get_hint_pair( "LEFT", "RIGHT", _( "to limit the left pane to a particular body part." ) ),
+                ctxt.get_hints( { "UP", "DOWN"}, _( "to scroll the left pane (list of items)." ) ),
+                ctxt.get_hints( { "SCROLL_ITEM_INFO_UP", "SCROLL_ITEM_INFO_DOWN"}, _( "to scroll the middle pane (item information)." ) ),
+                ctxt.get_hints( { "PREV_TAB", "NEXT_TAB"}, _( "to scroll the right pane (items grouped by body part)." ) ),
+                ctxt.get_hints( { "LEFT", "RIGHT"}, _( "to limit the left pane to a particular body part." ) ),
                 ctxt.get_hint( "MOVE_ARMOR", _( "to select an item for reordering." ) ),
                 ctxt.get_hint( "ASSIGN_INVLETS", _( "to assign special inventory letters to clothing." ) ),
                 ctxt.get_hint( "CHANGE_SIDE", _( "to change the side on which item is worn." ) ),
--- a/src/armor_layers.cpp
+++ b/src/armor_layers.cpp
@@ -737,14 +737,11 @@ void outfit::sort_armor( Character &guy )
         std::string temp = bp != bodypart_id( "bp_null" ) ? body_part_name_as_heading( bp, 1 ) : _( "All" );
         wprintz( w_sort_cat, c_yellow, "  << %s >>", temp );
         right_print( w_sort_cat, 0, 0, c_white, string_format(
-                         _( "[<color_yellow>%s</color>] Hide sprite.  "
-                            "[<color_yellow>%s</color>] Change side.  "
-                            "Press [<color_yellow>%s</color>] for help.  "
-                            "Press [<color_yellow>%s</color>] to change keybindings." ),
-                         ctxt.get_desc( "TOGGLE_CLOTH" ),
-                         ctxt.get_desc( "CHANGE_SIDE" ),
-                         ctxt.get_desc( "USAGE_HELP" ),
-                         ctxt.get_desc( "HELP_KEYBINDINGS" ) ) );
+                         "%s %s %s %s",
+                         ctxt.get_hint( "TOGGLE_CLOTH" ),
+                         ctxt.get_hint( "CHANGE_SIDE" ),
+                         ctxt.get_hint( "USAGE_HELP" ),
+                         ctxt.get_hint( "HELP_KEYBINDINGS" ) ) );
 
         leftListSize = tmp_worn.size();
         if( leftListLines > leftListSize ) {
@@ -1175,49 +1172,32 @@ void outfit::sort_armor( Character &guy )
             }
         } else if( action == "USAGE_HELP" ) {
             const std::vector<std::string> help_strings = {
-                string_format( _( "[<color_yellow>%s</color>]/[<color_yellow>%s</color>] to scroll the left pane (list of items).\n" ),
-                               ctxt.get_desc( "UP" ), ctxt.get_desc( "DOWN" ) ),
-                string_format( _( "[<color_yellow>%s</color>]/[<color_yellow>%s</color>] to scroll the middle pane (item information).\n" ),
-                               ctxt.get_desc( "SCROLL_ITEM_INFO_UP" ), ctxt.get_desc( "SCROLL_ITEM_INFO_DOWN" ) ),
-                string_format( _( "[<color_yellow>%s</color>]/[<color_yellow>%s</color>] to scroll the right pane (items grouped by body part).\n" ),
-                               ctxt.get_desc( "PREV_TAB" ), ctxt.get_desc( "NEXT_TAB" ) ),
-                string_format( _( "[<color_yellow>%s</color>]/[<color_yellow>%s</color>] to limit the left pane to a particular body part.\n" ),
-                               ctxt.get_desc( "LEFT" ), ctxt.get_desc( "RIGHT" ) ),
-                string_format( _( "[<color_yellow>%s</color>] to select an item for reordering.\n" ),
-                               ctxt.get_desc( "MOVE_ARMOR" ) ),
-                string_format( _( "[<color_yellow>%s</color>] to assign special inventory letters to clothing.\n" ),
-                               ctxt.get_desc( "ASSIGN_INVLETS" ) ),
-                string_format( _( "[<color_yellow>%s</color>] to change the side on which item is worn.\n" ),
-                               ctxt.get_desc( "CHANGE_SIDE" ) ),
-                string_format( _( "[<color_yellow>%s</color>] to toggle item visibility on character sprite.\n" ),
-                               ctxt.get_desc( "TOGGLE_CLOTH" ) ),
-                string_format( _( "[<color_yellow>%s</color>] to sort worn items into natural layer order.\n" ),
-                               ctxt.get_desc( "SORT_ARMOR" ) ),
-                string_format( _( "[<color_yellow>%s</color>] to equip a new item.\n" ),
-                               ctxt.get_desc( "EQUIP_ARMOR" ) ),
-                string_format( _( "[<color_yellow>%s</color>] to equip a new item at the currently selected position.\n" ),
-                               ctxt.get_desc( "EQUIP_ARMOR_HERE" ) ),
-                string_format( _( "[<color_yellow>%s</color>] to remove selected item.\n" ),
-                               ctxt.get_desc( "REMOVE_ARMOR" ) ),
-                "\n",
-                _( "Encumbrance explanation:\n" ),
+                ctxt.get_hint_pair( "UP", "DOWN", _( "to scroll the left pane (list of items)." ) ),
+                ctxt.get_hint_pair( "SCROLL_ITEM_INFO_UP", "SCROLL_ITEM_INFO_DOWN", _( "to scroll the middle pane (item information)." ) ),
+                ctxt.get_hint_pair( "PREV_TAB", "NEXT_TAB", _( "to scroll the right pane (items grouped by body part)." ) ),
+                ctxt.get_hint_pair( "LEFT", "RIGHT", _( "to limit the left pane to a particular body part." ) ),
+                ctxt.get_hint( "MOVE_ARMOR", _( "to select an item for reordering." ) ),
+                ctxt.get_hint( "ASSIGN_INVLETS", _( "to assign special inventory letters to clothing." ) ),
+                ctxt.get_hint( "CHANGE_SIDE", _( "to change the side on which item is worn." ) ),
+                ctxt.get_hint( "TOGGLE_CLOTH", _( "to toggle item visibility on character sprite." ) ),
+                ctxt.get_hint( "SORT_ARMOR", _( "to sort worn items into natural layer order." ) ),
+                ctxt.get_hint( "EQUIP_ARMOR", _( "to equip a new item." ) ),
+                ctxt.get_hint( "EQUIP_ARMOR_HERE", _( "to equip a new item at the currently selected position." ) ),
+                ctxt.get_hint( "REMOVE_ARMOR", _( "to remove selected item." ) ),
+                _( "\nEncumbrance explanation:" ),
                 _( "<color_light_gray>The first number is the summed encumbrance from all clothing on that bodypart."
-                   "The second number is an additional encumbrance penalty caused by wearing either multiple items "
+                   "  The second number is an additional encumbrance penalty caused by wearing either multiple items "
                    "on one of the bodypart's layers or wearing items the wrong way (e.g. a shirt over a backpack)."
-                   "The sum of these values is the effective encumbrance value your character has for that bodypart."
+                   "  The sum of these values is the effective encumbrance value your character has for that bodypart."
                    "</color>" )
             };
-            std::string assembled_string;
-            for( const std::string &current_line : help_strings ) {
-                assembled_string += current_line;
-            }
-
             const auto new_win = []() {
                 return catacurses::newwin( FULL_SCREEN_HEIGHT, FULL_SCREEN_WIDTH,
                                            point( std::max( 0, ( TERMX - FULL_SCREEN_WIDTH ) / 2 ),
                                                   std::max( 0, ( TERMY - FULL_SCREEN_HEIGHT ) / 2 ) ) );
             };
-            scrollable_text( new_win, _( "Sort armor help" ), assembled_string );
+            scrollable_text( new_win, _( "Sort armor help" ), enumerate_as_string( help_strings,
+                             enumeration_conjunction::newline ) );
         } else if( action == "QUIT" ) {
             exit = true;
         }
--- a/src/auto_note.cpp
+++ b/src/auto_note.cpp
@@ -364,7 +364,7 @@ void auto_note_manager_gui::show()
         hints.push_back( ctxt.get_hint( "CONFIRM", _( "Toggle" ) ) );
         // TODO: Show info about custom symbols (hotkey, hint, state)
         hints.push_back( ctxt.get_hint( "CHANGE_MAPEXTRA_CHARACTER" ) );
-        hints.push_back( ctxt.get_hint_pair( "NEXT_TAB", "PREV_TAB", _( "Switch Page" ) ) );
+        hints.push_back( ctxt.get_hints( { "NEXT_TAB", "PREV_TAB"}, _( "Switch Page" ) ) );
         fold_and_print( w_header, point( 0, 0 ), getmaxx( w_header ) - 1, c_light_gray,
                         enumerate_as_string( hints, enumeration_conjunction::space ) );
 
--- a/src/auto_note.cpp
+++ b/src/auto_note.cpp
@@ -335,6 +335,7 @@ void auto_note_manager_gui::show()
     ctxt.register_action( "SWITCH_AUTO_NOTE_OPTION" );
     ctxt.register_action( "HELP_KEYBINDINGS" );
     ctxt.register_action( "NEXT_TAB" );
+    ctxt.register_action( "PREV_TAB" );
     if( !char_emptyMode || !global_emptyMode ) {
         ctxt.register_navigate_ui_list();
         ctxt.register_action( "CONFIRM" );
@@ -354,10 +355,18 @@ void auto_note_manager_gui::show()
         wnoutrefresh( w_border );
 
         // == Draw header
-        int tmpx = 0;
-        tmpx += shortcut_print( w_header, point( tmpx, 0 ), c_white, c_light_green, _( "<E>nable" ) ) + 2;
-        tmpx += shortcut_print( w_header, point( tmpx, 0 ), c_white, c_light_green, _( "<D>isable" ) ) + 2;
-        shortcut_print( w_header, point( tmpx, 0 ), c_white, c_light_green, _( "<Enter> - Toggle" ) );
+        std::vector<std::string> hints;
+        hints.push_back( ctxt.get_hint( "SWITCH_AUTO_NOTE_OPTION",
+                                        get_option<bool>( "AUTO_NOTES_MAP_EXTRAS" ) ? keybinding_hint_state::TOGGLED_ON :
+                                        keybinding_hint_state::TOGGLED_OFF ) );
+        hints.push_back( ctxt.get_hint( "ENABLE_MAPEXTRA_NOTE" ) );
+        hints.push_back( ctxt.get_hint( "DISABLE_MAPEXTRA_NOTE" ) );
+        hints.push_back( ctxt.get_hint( "CONFIRM", _( "Toggle" ) ) );
+        // TODO: Show info about custom symbols (hotkey, hint, state)
+        hints.push_back( ctxt.get_hint( "CHANGE_MAPEXTRA_CHARACTER" ) );
+        hints.push_back( ctxt.get_hint_pair( "NEXT_TAB", "PREV_TAB", _( "Switch Page" ) ) );
+        fold_and_print( w_header, point( 0, 0 ), getmaxx( w_header ) - 1, c_light_gray,
+                        enumerate_as_string( hints, enumeration_conjunction::space ) );
 
         // Draw horizontal line and corner pieces of the table
         for( int x = 0; x < 78; x++ ) {
@@ -368,11 +377,8 @@ void auto_note_manager_gui::show()
                 mvwputch( w_header, point( x, iHeaderHeight - 2 ), c_light_gray, LINE_OXOX );
             }
         }
-        tmpx = 17;
-        tmpx += shortcut_print( w_header, point( tmpx, iHeaderHeight - 2 ),
-                                bCharacter ? hilite( c_white ) : c_white, c_light_green, _( "Character" ) ) + 2;
-        shortcut_print( w_header, point( tmpx, iHeaderHeight - 2 ),
-                        !bCharacter ? hilite( c_white ) : c_white, c_light_green, _( "Global" ) );
+
+        print_global_and_character_mode_headers( w_header, point( 17, iHeaderHeight - 2 ), !bCharacter );
 
         mvwprintz( w_header, point( 1, iHeaderHeight - 1 ), c_white, _( "Map Extra" ) );
         mvwprintz( w_header, point( 53, iHeaderHeight - 1 ), c_white, _( "Symbol" ) );
@@ -380,31 +386,6 @@ void auto_note_manager_gui::show()
 
         wnoutrefresh( w_header );
 
-        // TODO: Show info about custom symbols (hotkey, hint, state)
-        std::string header_info_custom_symbols = string_format( _( "<color_light_green>%1$s</color> %2$s" ),
-                ctxt.get_desc( "CHANGE_MAPEXTRA_CHARACTER" ), _( "Change a symbol for map extra" ) );
-        // NOLINTNEXTLINE(cata-use-named-point-constants)
-        fold_and_print( w_header, point( 0, 1 ), FULL_SCREEN_WIDTH - 2, c_white,
-                        header_info_custom_symbols );
-
-        mvwprintz( w_header, point( 39, 1 ), c_white, _( "Auto notes enabled:" ) );
-
-        int currentX = 60;
-        mvwprintz( w_header, point( currentX, 1 ), c_white,
-                   std::string( FULL_SCREEN_WIDTH - 2 - currentX, ' ' ) );
-
-        const bool enabled_auto_notes_ME = get_option<bool>( "AUTO_NOTES_MAP_EXTRAS" );
-        currentX += shortcut_print( w_header, point( currentX, 1 ),
-                                    enabled_auto_notes_ME ? c_light_green : c_light_red, c_white,
-                                    enabled_auto_notes_ME ? _( "True" ) : _( "False" ) );
-
-        currentX += shortcut_print( w_header, point( currentX, 1 ), c_white, c_light_green, "  " );
-        shortcut_print( w_header, point( currentX, 1 ), c_white, c_light_green, _( "<S>witch " ) );
-        currentX = 39;
-        mvwprintz( w_header, point( currentX, 0 ), c_white, std::string( FULL_SCREEN_WIDTH - 2 - currentX,
-                   ' ' ) );
-        shortcut_print( w_header, point( currentX, 0 ), c_white, c_light_green,
-                        _( "<Tab> to change pages." ) );
 
         // Clear table
         for( int y = 0; y < iContentHeight; y++ ) {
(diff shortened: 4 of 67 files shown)


src/output.cpp
```diff
@@ -2699,9 +2717,22 @@ std::pair<std::string, nc_color> rad_badge_color( const int rad )
     return std::pair<std::string, nc_color>( _( values[i].second.first ), values[i].second.second );
 }
 
-std::string formatted_hotkey( const std::string &hotkey, const nc_color text_color )
+std::string keybind_hint( const std::string &hotkey )
+{
+    // todo Add option to select between [x], x: and e[x]amine?
```
Qrox, line 2722: FYI [this method of `input_context`](https://github.com/CleverRaven/Cataclysm-DDA/blob/dc9b4941013aafa6b8cae9dff13aa3695c5a75f2/src/input.cpp#L1115-L1120) have something similar to what you want here, but for registered input actions.

src/inventory_ui.cpp
```diff
@@ -299,12 +299,12 @@ std::string enum_to_string<inventory_selector::uimode>( inventory_selector::uimo
 {
     switch( mode ) {
         case inventory_selector::uimode::hierarchy:
-            return "hierarchy";
+            return "Hierarchy";
```
Qrox, line 302: This is used for serialization of the enum, and is not marked for translation. You'd better create a separate function for the translated strings.

src/worldfactory.cpp
```diff
@@ -1927,12 +1908,6 @@ int worldfactory::show_worldgen_basic( WORLD *world )
                 int lvl = wg_slevels[sel_opt - 1] + ( action == "LEFT" ? -1 : 1 );
                 wg_slevels[sel_opt - 1] = clamp<int>( lvl, 0, wg_sliders[sel_opt - 1]->count() - 1 );
                 wg_sliders[sel_opt - 1]->apply_opts( wg_slevels[sel_opt - 1], world->WORLD_OPTIONS );
-            } else if( sel_opt > static_cast<int>( wg_sliders.size() ) ) {
```
strategineer, line 1930: worldgen option "buttons" are displayed vertically now, so don't need this anymore

src/inventory_ui.cpp
```diff
@@ -560,11 +560,8 @@ nc_color inventory_entry::get_invlet_color() const
 {
     if( !is_selectable() ) {
         return c_dark_gray;
-    } else if( get_player_character().inv->assigned_invlet.count( get_invlet() ) ) {
-        return c_yellow;
-    } else {
-        return c_white;
     }
+    return c_yellow;
```
Qrox, line 564: I think yellow and white are used to distinguish user-assigned and auto-assigned inventory letters.

src/gamemode_defense.cpp
```diff
@@ -1228,9 +1232,10 @@ void draw_caravan_borders( const catacurses::window &w, int current_window )
     mvwputch( w, point( FULL_SCREEN_WIDTH - 1, 0 ), col, LINE_OOXX );
     mvwputch( w, point( FULL_SCREEN_WIDTH - 1, FULL_SCREEN_HEIGHT - 1 ), col, LINE_XOOX );
 
-    // Quick reminded about help.
+    // Quick reminder about help.
     // NOLINTNEXTLINE(cata-text-style): literal question mark
```
Qrox, line 1236:
```suggestion
```

src/inventory_ui.cpp
strategineer: I wasn't sure about this change because I don't play with the auto-assigned inventory letters option (sidenote: it assigns the keys I use for navigation and it assigns digits preventing me from dropping/inserting X items at a time making the feature less than useful. Although both of these problems are bugs that can be fixed separately IMO).

Is it useful to the player for us to distinguish between user-assigned and auto-assigned inventory letters/keybindings?

NB: Thanks for the help @Qrox, it's much appreciated!

Qrox: I haven't used it in a while but the last time I used it user-assigned ones behave differently. I don't remember the details but I think user-assigned ones persist even if you drop the item and pick it back up, while auto-assigned ones can change after dropping and picking up.

strategineer: ahhhhhh might as well keep the color scheme as-is then, I'm already making enough changes here :)

src/action.cpp
```diff
@@ -536,16 +536,6 @@ std::string press_x( action_id act, const std::string &key_bound_pre,
     input_context ctxt = get_default_mode_input_context();
     return ctxt.press_x( action_ident( act ), key_bound_pre, key_bound_suf, key_unbound );
 }
-std::optional<std::string> press_x_if_bound( action_id act )
```
strategineer, line 539: function used once in src/help.cpp, deleting it. It'll be easier to refactor the rest of these `press_x` functions later on.

src/input.cpp
```diff
@@ -1145,38 +1157,205 @@ std::string input_context::get_desc(
...
+    std::string c_key = get_tag_from_color( get_hint_color_for_key( state ) );
+    if( state != keybinding_hint_state::NONE ) {
+        std::string c_sep = get_tag_from_color( get_hint_color_for_separator( state ) );
+        std::string c = get_tag_from_color( get_hint_color( state ) );
+        inline_fmt = to_translation( "keybinding",
+                                     string_format( "%s%s</color>%s%s</color>%s%s</color>%s%s</color>%s%s</color>", c, "%1$s", c_sep,
+                                             lp_sep, c_key, "%2$s", c_sep, rp_sep, c, "%3$s" ) );
+        separate_fmt = to_translation( "keybinding",
+                                       string_format( "%s%s</color>%s%s</color>%s%s</color>%s%s</color>", c_sep, lb_sep, c_key, "%1$s",
+                                               c_sep, rb_sep, c, "%2$s" ) );
```
Qrox, line 1200: The arguments to `to_translation` have to be string literals or it won't work (the string extractor won't know what runtime values the format arguments will take).

Also formatting a string with format specifiers as arguments to form a new format string feels too hacky to me, and it will likely break if some of the arguments happen to contain a sequence that can be interpreted as a format specifier.

I suggest passing fixed format strings and the needed variables as arguments to the final `get_desc` call instead and format the text in one go there.

Might want to use `colorize` instead of color tag string too to avoid confusing the translators with unpaired color tags.

src/input.cpp
```diff
@@ -1145,38 +1157,205 @@ std::string input_context::get_desc(
...
+        //~ %3$s: action description text after key.
+        inline_fmt = to_translation( "keybinding", string_format( "%s%s%s%s</color>%s%s", "%1$s", lp_sep,
+                                     c_key,
+                                     "%2$s", rp_sep, "%3$s" ) );
+        // \u00A0 is the non-breaking space
+        //~ %1$s: key description,
+        //~ %2$s: action description.
+        separate_fmt = to_translation( "keybinding", string_format( "%s%s%s</color>%s%s", lb_sep, c_key,
+                                       "%1$s",
+                                       rb_sep, "%2$s" ) );
```
Qrox, line 1213: Same here. Also the comments starting with "//~" are intended for translators and only apply to the extracted format strings, so the translators will not know what "%1$s" etc refer to after this change.

src/input.cpp
```diff
@@ -1145,38 +1157,205 @@ std::string input_context::get_desc(
...
+        case keybinding_hint_state::HIGHLIGHTED:
+            return get_option<bool>( "KEYBINDINGS_ALTERNATE_COLOR_THEME" ) ? h_cyan : h_yellow;
+        default:
+            return get_option<bool>( "KEYBINDINGS_ALTERNATE_COLOR_THEME" ) ? c_cyan : c_yellow;
+    }
 }
 
+std::string replace_spaces_with_non_breaking( std::string s )
+{
+    return std::regex_replace( s, std::regex( " " ), "\u00A0" );
```
Qrox, line 1260: Use regex for a simple substring replacement might be overkill.

src/input.cpp
strategineer: Weird, I responded to you this morning but I guess I never hit return? AHHHHHHH

This code was absolutely atrocious :) I refactored it in the way you suggested, thanks!

src/input.cpp
strategineer: The problem I was running into is that `std::replace` doesn't handle strings, it only works with characters which is a problem when dealing with unicode values.

Should I replace this call to `regex_replace` with custom string manipulation code?

src/wish.cpp
strategineer, line 1: WindowsTerminal_6UAplGZAlc
WindowsTerminal_7SSSgAXStn

src/worldfactory.cpp
strategineer, line 1: image

src/veh_interact.cpp
strategineer, line 1: image

src/armor_layers.cpp
strategineer, line 1: image
image

src/advanced_inv.cpp
strategineer, line 1: image

image

src/auto_note.cpp
strategineer, line 1: image

src/auto_pickup.cpp
strategineer, line 1: image

src/avatar.cpp
strategineer, line 1: image

src/bionics_ui.cpp
strategineer, line 1: image

src/color.cpp
strategineer, line 1: image

src/construction.cpp
strategineer, line 1: image

src/crafting_gui.cpp
strategineer, line 1: image

src/debug_menu.cpp
strategineer, line 1: image

src/dialogue_win.cpp
strategineer, line 1: image

src/diary_ui.cpp
strategineer, line 1: image

src/editmap.cpp
strategineer, line 1: image

src/game.cpp
strategineer, line 1: image
image
image
image

src/help.cpp
strategineer, line 1: image

src/inventory_ui.cpp
strategineer, line 1: image

src/item_contents.cpp
strategineer, line 1: image

src/iuse_software_kitten.cpp
strategineer, line 1: image
image
image

src/iuse_software_snake.cpp
strategineer, line 1: image

src/iuse_software_lightson.cpp
strategineer, line 1: image

src/iuse_software_minesweeper.cpp
strategineer, line 1: image

src/iuse_software_sokoban.cpp
strategineer, line 1: image

src/magic.cpp
strategineer, line 1: image
image

src/main_menu.cpp
strategineer, line 1: image

src/medical_ui.cpp
strategineer, line 1: image

src/messages.cpp
strategineer, line 1: image

src/mutation_ui.cpp
strategineer, line 1: image
image

src/newcharacter.cpp
strategineer, line 1: image
image
image
image
image
image
image
image

src/npctalk.cpp
strategineer, line 1: image

src/options.cpp
strategineer, line 1: image

src/overmap_ui.cpp
strategineer, line 1: image
image

src/panels.cpp
strategineer, line 1: image

src/player_display.cpp
strategineer, line 1: image

src/popup.cpp
strategineer, line 1: image

src/proficiency_ui.cpp
strategineer, line 1: image

src/ranged.cpp
strategineer, line 1: image

strategineer: image

src/safemode_ui.cpp
strategineer, line 1: image

src/smart_controller_ui.cpp
strategineer, line 1: image

src/trade_ui.cpp
strategineer, line 1: image

src/ui.cpp
strategineer, line 1: image

src/gamemode_defense.cpp
strategineer, line 1: image

src/input.cpp
Qrox: Huh, didn't know that the standard library does not provide string replacement. Yeah I think a custom string manipulation function would be better here.

strategineer: Done!

Qrox:
Currently it looks a bit weird when using the inline format for descriptions containing spaces, especially when the descriptions are also separated by space. Perhaps disabling the inline format when a description contains space will make it look better?

src/ui.cpp
```diff
@@ -778,9 +779,13 @@ void uilist::show( ui_adaptor &ui )
 
             mvwprintz( window, point( pad_left + 1, estart + si ), co, padspaces );
             if( entries[ei].hotkey.has_value() && entries[ei].hotkey.value() != input_event() ) {
-                const nc_color hotkey_co = ei == selected ? hilight_color : hotkey_color;
-                mvwprintz( window, point( pad_left + 1, estart + si ), entries[ ei ].enabled ? hotkey_co : co,
-                           "%s", right_justify( entries[ei].hotkey.value().short_description(), 2 ) );
```
Qrox, line 783: The hotkey description could be two letters long in the keycode mode (for example ctrl+a will be displayed as `^a`):

https://github.com/CleverRaven/Cataclysm-DDA/blob/6d311a79b1ba922a0bd86282b3f2290e47a81428/src/input.cpp#L159-L163

So you should keep `right_justify` and add one more to pad_left in line 797 below.

src/worldfactory.cpp
```diff
@@ -1666,12 +1649,12 @@ int worldfactory::show_worldgen_basic( WORLD *world )
...
         // Hint text
         std::string hint_txt =
-            string_format( _( "Press [<color_yellow>%s</color>] to pick a random name for your world.\n"
-                              "Navigate options with [<color_yellow>directional keys</color>] "
-                              "and confirm with [<color_yellow>%s</color>].\n"
-                              "Press [<color_yellow>%s</color>] to see additional control information." ),
-                           ctxt.get_desc( "PICK_RANDOM_WORLDNAME", 1U ), ctxt.get_desc( "CONFIRM", 1U ),
-                           ctxt.get_desc( "HELP_KEYBINDINGS", 1U ) );
+            string_format( _( "Press %s to pick a random name for your world.\n"
+                              "Navigate options with directional keys "
```
Qrox, line 1653: Should print the actual hotkeys

src/worldfactory.cpp
```diff
@@ -1875,21 +1850,14 @@ void worldfactory::draw_modselection_borders( const catacurses::window &win,
...
-                       "[<color_yellow>%s</color>] = keybindings" ),
-                    ctxtp.get_desc( "SAVE_DEFAULT_MODS" ),
-                    ctxtp.get_desc( "PREV_TAB" ),
-                    ctxtp.get_desc( "NEXT_TAB" ),
-                    ctxtp.get_desc( "LEFT" ),
-                    ctxtp.get_desc( "RIGHT" ),
-                    ctxtp.get_desc( "PREV_CATEGORY_TAB" ),
-                    ctxtp.get_desc( "NEXT_CATEGORY_TAB" ),
-                    ctxtp.get_desc( "HELP_KEYBINDINGS" )
+                    _( "%s | %s | %s | %s | %s" ),
```
Qrox, line 1855: Maybe remove `|` to be consistent with the style of other keybinding text?

strategineer:
> Currently it looks a bit weird when using the inline format for descriptions containing spaces, especially when the descriptions are also separated by space. Perhaps it will look better if you disable the inline format when the description contains space and the hotkey is not the first character of the description?
> 
> The help menu also looks a bit messy with the inline format, might want to use separated format for it.
> 
> A space between the dialogue option hotkey and text might make it look better.

This should look better!

image


WindowsTerminal_qycuatLnpg

src/input.cpp
```diff
@@ -1112,71 +1117,291 @@ std::string input_context::get_desc( const std::string &action_descriptor,
...
+    const int n_max_inline = get_option<int>( "KEYBINDINGS_INLINE_UP_TO_X" );
+    const bool can_inline = n_max_inline == 0 || n_text < n_max_inline;
+    const auto &events = inp_mngr.get_input_for_action( action_descriptor, category );
     bool na = true;
     for( const input_event &evt : events ) {
         if( is_event_type_enabled( evt.type ) && evt_filter( evt ) ) {
             na = false;
+            if( n_text == 0 || !should_inline_if_possible || !can_inline ) {
+                break;
+            }
```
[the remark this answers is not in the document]
strategineer: I thought so too but I tried this a few days ago and realized that we're going through this loop for two reasons:
- to try and inline the key within the text
- to set the `na` variable to false if there's at least one key bound to the event

So moving this check outside the loop breaks the logic.

src/ranged.cpp
```diff
@@ -3646,62 +3631,51 @@ void target_ui::draw_controls_list( int text_y )
...
     if( shifting_view ) {
-        lines.push_back( {8, colored( col_move, _( "Shift view with directional keys" ) )} );
+        lines.push_back( {8, colorize( _( "Shift view with directional keys" ), input_context::get_hint_color( move_hint_state ) )} );
     } else {
-        lines.push_back( {8, colored( col_move, _( "Move cursor with directional keys" ) )} );
+        lines.push_back( {8, colorize( _( "Move cursor with directional keys" ), input_context::get_hint_color( move_hint_state ) ) } );
     }
     if( is_mouse_enabled() ) {
         std::string move = _( "Mouse: LMB: Target, Wheel: Cycle," );
         std::string fire = _( "RMB: Fire" );
```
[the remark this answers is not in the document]
strategineer: Bound to actions

I changed the code to this.
```
std::string move = string_format( _( "%s %s" ), ctxt.get_hint_mouse( "SELECT", _( "Target" ),
                                        move_hint_state ), ctxt.get_hint_mouse( "SCROLL_UP", _( "Cycle" ), move_hint_state ) );
        std::string fire = ctxt.get_hint_mouse( "FIRE", fire_hint_state );
```

and implemented the `get_hint_mouse` function in case we want to use them elsewhere eventually

src/ranged.cpp
```diff
@@ -3646,62 +3631,51 @@ void target_ui::draw_controls_list( int text_y )
...
-        std::string dropaim = string_format( _( "[%s] to stop aiming." ),
-                                             bound_key( "STOPAIM" ).short_description() );
 
-        lines.push_back( {2, colored( col_fire, aim )} );
-        lines.push_back( { 2, colored( col_fire, dropaim ) } );
-        lines.push_back( {4, colored( col_fire, aim_and_fire )} );
+        lines.push_back( {2, ctxt.get_hint( "AIM", _( "Steady your aim (10 moves)" ), fire_hint_state )} );
+        lines.push_back( {2, ctxt.get_hint( "STOPAIM", _( "Stop aiming" ), fire_hint_state )} );
+        lines.push_back( {4, _( "Aim and fire with:" )} );
+        lines.push_back( {5, string_format( "  (%s)", enumerate_as_string( aim_and_fire, enumeration_conjunction::space ) )} );
```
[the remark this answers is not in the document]
strategineer: huhmm I changed things around, it seems like it's working better now

image


image

src/ui.cpp
strategineer: I think this did the trick https://github.com/CleverRaven/Cataclysm-DDA/pull/64478/commits/cdf95fbd01f4cab869bb3ccf1b10cd97e65acadd

src/safemode_ui.cpp
```diff
@@ -144,51 +144,49 @@ void safemode::show( const std::string &custom_name_in, bool is_safemode_in )
                 translate_marker( "<T>est" )
             }
         };
```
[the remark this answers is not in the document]
strategineer: Good catch, it isn't. "<M>ove" is included, the action is called "SWAP_RULE_GLOBAL_CHAR"

src/options.cpp
```diff
@@ -1685,6 +1685,67 @@ void options_manager::add_options_interface()
...
+             false
+           );
+        add( "KEYBINDINGS_INLINE", page_id, to_translation( "Inline keybindings in hints" ),
+             to_translation( "Inline keybindings in hints (like 'E[x]amine') where possible." ),
+             true
+           );
+
+
+        add( "KEYBINDINGS_INLINE_UP_TO_X", page_id, to_translation( "Limit inlining of keybinding hints" ),
+             to_translation( "Inline keybindings in hints (like 'E(x)amine') if the length of the hint text is smaller than X characters. Disable this limit by setting this to 0." ),
```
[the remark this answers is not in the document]
strategineer: deleting this option as your proposal to only inline the keybind when there's no spaces in the suffix/text part makes this option obsolete

src/game.cpp
```diff
@@ -7333,24 +7328,16 @@ look_around_result game::look_around(
...
 
             if( !ctxt.keys_bound_to( "toggle_pixel_minimap" ).empty() ) {
-                std::string pixel_minimap_text = string_format( _( "%s - %s" ),
-                                                 ctxt.get_desc( "toggle_pixel_minimap" ),
-                                                 ctxt.get_action_name( "toggle_pixel_minimap" ) );
-                right_print( w_info, getmaxy( w_info ) - 1, 1, pixel_minimap_option ? c_light_green : c_green,
-                             pixel_minimap_text );
+                right_print( w_info, getmaxy( w_info ) - 1, 1, ctxt.get_hint( "toggle_pixel_minimap",
+                             pixel_minimap_option ? keybinding_hint_state::TOGGLED_ON : keybinding_hint_state::TOGGLED_OFF ) );
             }
```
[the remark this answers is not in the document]
strategineer: I added an ifdef above to solve this problem
```
#if defined(TILES)
    ctxt.register_action( "toggle_pixel_minimap" );
#endif
```

when playing on curses, the keybind doesn't show up in the keybinds menu or this menu.

NB: This feature is also broken https://github.com/CleverRaven/Cataclysm-DDA/issues/58776 so that's something we'll tackle in another PR.

src/armor_layers.cpp
```diff
@@ -1175,49 +1172,32 @@ void outfit::sort_armor( Character &guy )
...
+                ctxt.get_hint_pair( "PREV_TAB", "NEXT_TAB", _( "to scroll the right pane (items grouped by body part)." ) ),
+                ctxt.get_hint_pair( "LEFT", "RIGHT", _( "to limit the left pane to a particular body part." ) ),
+                ctxt.get_hint( "MOVE_ARMOR", _( "to select an item for reordering." ) ),
+                ctxt.get_hint( "ASSIGN_INVLETS", _( "to assign special inventory letters to clothing." ) ),
+                ctxt.get_hint( "CHANGE_SIDE", _( "to change the side on which item is worn." ) ),
+                ctxt.get_hint( "TOGGLE_CLOTH", _( "to toggle item visibility on character sprite." ) ),
+                ctxt.get_hint( "SORT_ARMOR", _( "to sort worn items into natural layer order." ) ),
+                ctxt.get_hint( "EQUIP_ARMOR", _( "to equip a new item." ) ),
+                ctxt.get_hint( "EQUIP_ARMOR_HERE", _( "to equip a new item at the currently selected position." ) ),
+                ctxt.get_hint( "REMOVE_ARMOR", _( "to remove selected item." ) ),
```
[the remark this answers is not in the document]
strategineer: I resolved this by implementing your proposal to not inline keys into sentences.

src/main_menu.cpp
```diff
@@ -348,7 +350,7 @@ void main_menu::print_menu( const catacurses::window &w_open, int iSel, const po
 
     int menu_length = 0;
     for( size_t i = 0; i < vMenuItems.size(); ++i ) {
-        menu_length += utf8_width_notags( vMenuItems[i].c_str() ) + 2;
+        menu_length += utf8_width_notags( vMenuItems[i].c_str() ) + 3;
```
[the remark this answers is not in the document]
strategineer: I think I was able to resolve this but i'll need to take another look.

src/overmap_ui.cpp
```diff
@@ -1158,27 +1158,33 @@ static void draw_om_sidebar(
         }
     }
 
-    mvwprintz( wbar, point( 1, 12 ), c_magenta, _( "Use movement keys to pan." ) );
-    mvwprintz( wbar, point( 1, 13 ), c_magenta, _( "Press W to preview route." ) );
-    mvwprintz( wbar, point( 1, 14 ), c_magenta, _( "Press again to confirm." ) );
+    mvwprintz( wbar, point( 1, 12 ), input_context::get_hint_color(),
+               _( "Use movement keys to pan." ) );
```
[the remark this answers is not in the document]
strategineer: image

src/ui.cpp
```diff
@@ -538,20 +543,21 @@ void uilist::calc_data()
...
             }
         }
         if( desc_enabled ) {
-            const int min_desc_width = std::min( max_desc_width, std::max( w_width, descwidth_final ) - 4 );
+            const int min_desc_width = std::min( max_desc_width, std::max( w_width,
+                                                 descwidth_final ) - WIDTH_BUFFER );
             int descwidth = find_minimum_fold_width( footer_text.empty() ? entries[i].desc : footer_text,
                             desc_lines, min_desc_width, max_desc_width );
-            descwidth += 4; // 2x border + 2x ' ' pad
+            descwidth += WIDTH_BUFFER; // 2x border + 4x ' ' pad
```
Qrox, line 560: Might not be needed because description does not start with a hotkey hint.

src/ui.cpp
```diff
@@ -607,15 +613,15 @@ void uilist::calc_data()
...
+                        w_width = realtextwidth + WIDTH_BUFFER;
                     }
                 }
             }
         } else if( textwidth != -1 ) {
             realtextwidth = textwidth;
-            if( realtextwidth + 4 > w_width ) {
-                w_width = realtextwidth + 4;
+            if( realtextwidth + WIDTH_BUFFER > w_width ) {
+                w_width = realtextwidth + WIDTH_BUFFER;
```
Qrox, line 624: I think `text` is printed before all uilsit entries and does not start with a hotkey hint, so changing 4 to `WIDTH_BUFFER` is not needed.

src/ui.cpp
```diff
@@ -627,9 +633,8 @@ void uilist::calc_data()
     if( desc_enabled ) {
         desc_lines = 0;
         for( const uilist_entry &ent : entries ) {
-            // -2 for borders, -2 for padding
             desc_lines = std::max<int>( desc_lines, foldstring( footer_text.empty() ? ent.desc : footer_text,
-                                        w_width - 4 ).size() );
+                                        w_width - WIDTH_BUFFER ).size() );
```
Qrox, line 637: Not needed for description.

src/ui.cpp
```diff
@@ -692,7 +697,7 @@ void uilist::reposition( ui_adaptor &ui )
 {
     setup();
     if( filter_popup ) {
-        filter_popup->window( window, point( 4, w_height - 1 ), w_width - 4 );
+        filter_popup->window( window, point( WIDTH_BUFFER, w_height - 1 ), w_width - WIDTH_BUFFER );
```
Qrox, line 700: Not needed for filter.

src/ui.cpp
```diff
@@ -742,7 +747,7 @@ void uilist::show( ui_adaptor &ui )
     int estart = 1;
     if( !textformatted.empty() ) {
         for( int i = 0; i < text_lines; i++ ) {
-            trim_and_print( window, point( 2, 1 + i ), getmaxx( window ) - 4,
+            trim_and_print( window, point( 2, 1 + i ), getmaxx( window ) - WIDTH_BUFFER,
```
Qrox, line 750: `text` starts one space after the left border and end one space before the right border, so no need to change this.

src/ui.cpp
```diff
@@ -836,7 +843,7 @@ void uilist::show( ui_adaptor &ui )
         }
 
         if( static_cast<size_t>( selected ) < entries.size() ) {
-            fold_and_print( window, point( 2, w_height - desc_lines - 1 ), w_width - 4, text_color,
+            fold_and_print( window, point( 2, w_height - desc_lines - 1 ), w_width - WIDTH_BUFFER, text_color,
```
Qrox, line 846: Not needed for description.

src/ui.cpp
Qrox: I did my best to mark which changes are probably not needed, but the uilist code is a mess and I'm tired analyzing magic number now. Might want to fine-tune it later instead.

src/npctalk.cpp
```diff
@@ -1818,7 +1818,7 @@ talk_data talk_response::create_option_line( const dialogue &d, const input_even
     }
     talk_data results;
     results.color = color;
-    results.hotkey_desc = right_justify( hotkey.short_description(), 2 );
+    results.hotkey_desc = hotkey.short_description();
```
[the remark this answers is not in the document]
strategineer: How would you get this to show a key with a modifier? It seems like keys show up alphabetically from a to z?

src/overmap_ui.cpp
```diff
@@ -1158,8 +1158,7 @@ static void draw_om_sidebar(
         }
     }
 
-    mvwprintz( wbar, point( 1, 12 ), input_context::get_hint_color(),
-               _( "Use movement keys to pan." ) );
+    print_colored_text( wbar, point( 1, 12 ), inp_ctxt->get_hint_directions( _( "Pan view" ) ) );
```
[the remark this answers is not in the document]
strategineer: Those are keys I'm using for moving around in cardinal directions (I've got a weird keyboard layout).

With default settings it shows the following:
image

Which seems reasonable although I suspect that most players don't play with vi-keys and use the keypad instead (or maybe even arrow keys and avoid moving diagonally). It's hard to know what players are actually using to move around and thus hard to know what we should display here.

I'm starting to think that because movement like this is the default "action" players are doing, we can avoid displaying any sort of keybinds related to this because if the player has gotten this far, they know how to move around their view/character (however they like to do so) and we probably don't need to remind them.

What do you think?

src/ui.cpp
```diff
@@ -783,20 +788,22 @@ void uilist::show( ui_adaptor &ui )
...
+                // cases printing starts at pad_left+1, here it starts at pad_left+WIDTH_BUFFER, so 3 cells less
                 // to be used.
                 const std::string &entry = ei == selected ? remove_color_tags( entries[ ei ].txt ) :
                                            entries[ ei ].txt;
-                point p( pad_left + 5, estart + si );
+                point p( pad_left + 6, estart + si );
                 entries[ei].drawn_rect =
-                    inclusive_rectangle<point>( p + point( -3, 0 ), p + point( -4 + pad_size, 0 ) );
+                    inclusive_rectangle<point>( p + point( -WIDTH_BUFFER, 0 ), p + point( -WIDTH_BUFFER + pad_size,
+                                                0 ) );
```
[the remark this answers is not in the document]
strategineer: WindowsTerminal_N7YfYSUBv7

It seems good!

src/ui.cpp
```diff
@@ -850,7 +857,7 @@ void uilist::show( ui_adaptor &ui )
     } else {
         if( !filter.empty() ) {
             mvwprintz( window, point( 2, w_height - 1 ), border_color, "< %s >", filter );
-            mvwprintz( window, point( 4, w_height - 1 ), text_color, filter );
+            mvwprintz( window, point( WIDTH_BUFFER, w_height - 1 ), text_color, filter );
```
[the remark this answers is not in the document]
strategineer: Yes I broke this definitely :).

src/ui.cpp
strategineer: Thanks for being so meticulous. This https://github.com/CleverRaven/Cataclysm-DDA/pull/64478/commits/2fbc662ed6e32eee9b009acc12f23fd661beb2c0 should resolve the issue. I replaced my WIDTH_BUFFER variable back to 4 everywhere one-by-one, recompiled and tested a few uilists until I found which ones needed to actually be 6 now :) due to the added padding

src/ui.cpp
```diff
@@ -495,14 +500,14 @@ void uilist::calc_data()
...
-        w_width = 4;
+        w_width = WIDTH_BUFFER;
         if( !title.empty() ) {
             w_width = utf8_width( title ) + 5;
         }
     } else {
         w_width = w_width_setup.fun();
     }
-    const int max_desc_width = w_auto ? TERMX - 4 : w_width - 4;
+    const int max_desc_width = w_auto ? TERMX - WIDTH_BUFFER : w_width - WIDTH_BUFFER;
```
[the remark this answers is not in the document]
strategineer: resolved by https://github.com/CleverRaven/Cataclysm-DDA/pull/64478/commits/2fbc662ed6e32eee9b009acc12f23fd661beb2c0

src/ui.cpp
strategineer: resolved by https://github.com/CleverRaven/Cataclysm-DDA/pull/64478/commits/2fbc662ed6e32eee9b009acc12f23fd661beb2c0

src/ui.cpp
strategineer: resolved by https://github.com/CleverRaven/Cataclysm-DDA/pull/64478/commits/2fbc662ed6e32eee9b009acc12f23fd661beb2c0

src/ui.cpp
strategineer: resolved by https://github.com/CleverRaven/Cataclysm-DDA/pull/64478/commits/2fbc662ed6e32eee9b009acc12f23fd661beb2c0

src/input.cpp
```diff
@@ -10,6 +10,7 @@
 #include <memory>
 #include <new>
 #include <optional>
+#include <regex>
```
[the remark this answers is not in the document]
strategineer: I thought I removed this... Sorry about that.

src/input.cpp
```diff
@@ -1105,7 +1110,20 @@ std::string input_context::get_desc( const std::string &action_descriptor,
     const std::string separator = inputs_to_show.size() > 2 ? _( ", or " ) : _( " or " );
     std::string rval;
     for( size_t i = 0; i < inputs_to_show.size(); ++i ) {
-        rval += inputs_to_show[i].long_description();
+        const input_event &event = inputs_to_show[i];
+        std::string description  = event.long_description();
```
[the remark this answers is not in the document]
strategineer: I reverted my hacky change here. I'll leave the strings as-is for now unless someone complains about "MOUSE_LEFT" looks weird in the UI.

src/popup.cpp
```diff
@@ -103,8 +103,13 @@ std::vector<std::vector<std::string>> query_popup::fold_query(
...
     for( const query_popup::query_option &opt : options ) {
-        const std::string &name = ctxt.get_action_name( opt.action );
-        const std::string &desc = ctxt.get_desc( opt.action, name, opt.filter );
+        // TODO: This function is only called once (on init) so we're not able
+        //   to properly color the key (colour tagging it prevents it from
+        //   being highlighted when selected)... This isn't a too much of
+        //   problem with the default settings because the key is separated
+        //   by parens but still...
+        const std::string &desc = ctxt.get_hint( opt.action, keybinding_hint_state::NONE_AT_ALL,
+                                  opt.filter );
```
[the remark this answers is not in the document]
strategineer: it's a little hacky but my solution here works https://github.com/CleverRaven/Cataclysm-DDA/pull/64478/commits/cdadc091f83a81f5d886ec7e31d3d06b68b4d80b

src/main_menu.cpp
```diff
@@ -350,14 +350,13 @@ void main_menu::print_menu( const catacurses::window &w_open, int iSel, const po
 
     int menu_length = 0;
     for( size_t i = 0; i < vMenuItems.size(); ++i ) {
-        menu_length += utf8_width_notags( vMenuItems[i].c_str() ) + 3;
-        if( !vMenuHotkeys[i].empty() ) {
-            menu_length += utf8_width( vMenuHotkeys[i][0] );
-        }
+        const std::string shortcut = remove_color_tags( shortcut_text( vMenuItems[i] ) );
+        const int shortcut_width = utf8_width_notags( shortcut.c_str() ) + 1;
+        menu_length += shortcut_width;
```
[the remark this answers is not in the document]
Qrox: Oh, nevermind then, not sure how I missed it.

src/main_menu.cpp
```diff
@@ -348,14 +351,13 @@ void main_menu::print_menu( const catacurses::window &w_open, int iSel, const po
 
     int menu_length = 0;
     for( size_t i = 0; i < vMenuItems.size(); ++i ) {
-        menu_length += utf8_width_notags( vMenuItems[i].c_str() ) + 2;
-        if( !vMenuHotkeys[i].empty() ) {
-            menu_length += utf8_width( vMenuHotkeys[i][0] );
-        }
+        const std::string shortcut = remove_color_tags( shortcut_text( vMenuItems[i] ) );
+        const int shortcut_width = utf8_width_notags( shortcut.c_str() ) + 1;
```
Qrox, line 355: The following might fix the centering problem:
```suggestion
        const int shortcut_width = utf8_width_notags( shortcut.c_str() );
```

src/main_menu.cpp
```diff
@@ -348,14 +351,13 @@ void main_menu::print_menu( const catacurses::window &w_open, int iSel, const po
...
+        menu_length += shortcut_width;
     }
     const int free_space = std::max( 0, window_width - menu_length - offset.x );
-    const int spacing = free_space / ( static_cast<int>( vMenuItems.size() ) + 1 );
-    const int width_of_spacing = spacing * ( vMenuItems.size() + 1 );
+    const int spacing = free_space / static_cast<int>( vMenuItems.size() );
+    const int width_of_spacing = spacing * vMenuItems.size();
```
Qrox, line 360:
```suggestion
    const int spacing = std::max( 1, free_space / static_cast<int>( vMenuItems.size() + 1 ) );
    const int width_of_spacing = spacing * ( vMenuItems.size() + 1 );
```

src/main_menu.cpp
```diff
@@ -119,8 +119,8 @@ std::vector<int> main_menu::print_menu_items( const catacurses::window &w_in,
         }
         ret.push_back( utf8_width_notags( text.c_str() ) );
 
-        std::string temp = shortcut_text( iSel == i ? hilite( c_yellow ) : c_yellow, vItems[i] );
-        text += string_format( "[%s]", colorize( temp, iSel == i ? hilite( c_white ) : c_white ) );
```
Qrox, line 123: The following call at line 131 should be updated to use the default separator (remove `']'`) because the items are now not separated by `]` anymore.

src/overmap_ui.cpp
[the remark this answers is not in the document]
Qrox: That's actually a good point.

Qrox (commented):
Some of these are responses to previous comments but somehow github decides to show them as standalone comments while burying the context in the comment history.

src/main_menu.cpp
strategineer: that combined with your other fix did the trick

image


it works! ;)

src/ranged.cpp
```diff
@@ -3534,63 +3519,48 @@ void target_ui::draw_controls_list( int text_y )
...
+    movement_text = colorize( movement_text, input_context::get_hint_color( move_hint_state ) );
+    lines.push_back( {9, string_format( "%s %s", ctxt.get_hint_directions( move_hint_state ), movement_text )} );
     if( is_mouse_enabled() ) {
-        std::string move = _( "Mouse: LMB: Target, Wheel: Cycle," );
-        std::string fire = _( "RMB: Fire" );
-        lines.push_back( {7, colored( col_move, move ) + " " + colored( col_fire, fire )} );
+        std::string move = string_format( _( "%s %s" ), ctxt.get_hint_mouse( "SELECT", _( "Target" ),
+                                          move_hint_state ), ctxt.get_hint_mouse( "SCROLL_UP", _( "Cycle" ), move_hint_state ) );
+        std::string fire = ctxt.get_hint_mouse( "FIRE", fire_hint_state );
+        lines.push_back( {8, string_format( "%s %s", move, fire )} );
```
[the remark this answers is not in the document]
strategineer: These hints were displaying mouse inputs specifically. I've retained that by limiting the hints in this way because the keys to move/fire with the keyboard are displayed further down. What would like me to do here?

src/ranged.cpp
```diff
@@ -3534,63 +3519,48 @@ void target_ui::draw_controls_list( int text_y )
...
 
     // Compile full list
-    if( shifting_view ) {
-        lines.push_back( {8, colored( col_move, _( "Shift view with directional keys" ) )} );
-    } else {
-        lines.push_back( {8, colored( col_move, _( "Move cursor with directional keys" ) )} );
-    }
+    std::string movement_text = shifting_view ? _( "Shift view" ) : _( "Move cursor" );
+    movement_text = colorize( movement_text, input_context::get_hint_color( move_hint_state ) );
+    lines.push_back( {9, string_format( "%s %s", ctxt.get_hint_directions( move_hint_state ), movement_text )} );
```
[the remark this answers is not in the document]
strategineer: Definitely!

src/overmap_ui.cpp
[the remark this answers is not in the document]
Qrox: I don't have a strong preference for either so I'll leave that up to you.

src/popup.cpp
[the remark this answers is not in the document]
strategineer: I refactored the code so it's a little less hacky now.

I couldn't quite get the other solution to work cleanly (although I understand it's possible). Do you think this can be good enough for now and I can come back to this in another PR? Or is this unmergeable in this state you think?

src/output.cpp
```diff
@@ -2608,20 +2608,18 @@ void print_global_and_character_mode_headers( const catacurses::window &w, const
...
-            return shortcut;
-        }
         std::string prestring = fmt.substr( 0, pos );
         std::string poststring = fmt.substr( pos_end + 1, std::string::npos );
 
-        std::string hint_text = string_format( "%s%s%s", prestring, shortcut,
+        std::string hint_text = string_format( "%s%s%s", prestring,
+                                               remove_shortcut_from_text ? "" : shortcut,
                                                poststring );
         return input_context::get_hint_basic( shortcut, hint_text, state );
```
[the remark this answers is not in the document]
strategineer: > Well, I thought removing the colon would make it replace the first letter with the hotkey even if the string contains spaces but I forgot you reverted that change.

It did do that. The menu looked OK but it used the inlined separators instead of the separate separators ("(a): Blah" instead of "[a]: Blah"). This wasn't consistent so I added this extra `remove_shortcut_from_text ` to the `shortcut_text` function as a way of dealing with this edge case. The code works as expected in every other context the `shortcut_text` function is used (I'm pretty sure) because the `remove_shortcut_from_text` param defaults to false.

For example, this is what the french menu looks like:
image

Just to clarify, are there other reasons to implement the solution you've described as opposed to what has already been implemented?

src/popup.cpp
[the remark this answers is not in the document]
Qrox: If you add a comment to the replacement function to warn that it could break in some edge cases I think it's good enough for now.

src/output.cpp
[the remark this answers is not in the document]
Qrox: Can you test the Japanese menu and let me know if it prints the correct strings? They translated `<W|w> world` as `<W|w> 世界` and I'm concerned it may become `[W] W 世界` with the current code.

src/ranged.cpp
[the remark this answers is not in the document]
Qrox: Previously I implemented selecting with the mouse cursor and a keyboard key by binding a non-mouse key to the "SELECT" action, which is different from using the keyboard hotkey to fire at the currently selected target. Could you try rebinding a keyboard hotkey to the "SELECT" action and see the hints make sense in that case?

src/output.cpp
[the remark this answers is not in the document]
strategineer: Good catch, ya this is broken.

image

This change should resolve the issue but I'll need to test this more. https://github.com/CleverRaven/Cataclysm-DDA/pull/64478/commits/394e371de9f1ff2c575251637b6d0b9aef1bae8f

There's a lot of edge cases ;)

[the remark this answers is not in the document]
Qrox: Please test the Chinese (simplified and traditional) menu with inline format disabled in the settings. They removed the space from the translation (so `<W|w> World` becomes `<W|w>世界` instead of `<W|w> 世界`) and I think the newly added checks will not work in that case.

Although my point is that it's hard to add checks for data that are not (fully) controled by the project itself so I think it may still be more approriate to only replace the hotkey tags with the formatted hotkeys in place and disregard the user's inline setting, at least for now until we fully migrate the main menu hotkey system to be more in line with the rest of the game.

Issue #63982, opened 2023-03-06
Enchantments are not considered in the player displayed DPS


Maleclypse:
### Describe the bug

Mirrored Estoc is both slower and deals less damage than the frostrimed estoc and yet the PC has no way of seeing this.  It will likely be equally true for almost all magical weapons.
```
{
    "id": "mirror_march_estoc",
    "type": "GENERIC",
    "symbol": "/",
    "color": "light_gray",
    "name": { "str": "mirrored estoc" },
    "description": "A thin piercing sword no wider than two fingers side by side, it's mirrored surface reflects the world back at it's wielders opponents and is difficult to see to parry.  You've never felt a sword like this before.",
    "price": 135000,
    "price_postapoc": 50000,
    "material": [ "moon_tears" ],
    "techniques": [ "WBLOCK_2", "PRECISE" ],
    "flags": [ "STAB", "DURABLE_MELEE", "SHEATH_SWORD" ],
    "weapon_category": [ "MEDIEVAL_SWORDS", "LONG_SWORDS", "FENCING_WEAPONRY" ],
    "weight": "1214 g",
    "volume": "2750 ml",
    "longest_side": "130 cm",
    "bashing": 8,
    "cutting": 42,
    "to_hit": { "grip": "weapon", "length": "long", "surface": "line", "balance": "good" },
    "category": "weapons",
    "qualities": [ [ "CUT", 1 ], [ "BUTCHER", 8 ] ]
  },
  {
    "id": "winter_lord_estoc",
    "type": "GENERIC",
    "name": { "str": "frostrimed estoc" },
    "copy-from": "mirror_march_estoc",
    "description": "This sword is covered in shifting frost and seems to move of its own volition in your hands.",
    "relic_data": {
      "passive_effects": [
        {
          "has": "WIELD",
          "condition": "ALWAYS",
          "values": [ { "value": "ITEM_DAMAGE_COLD", "add": 9 }, { "value": "ATTACK_SPEED", "add": -20 } ]
        }
      ]
    }
  },
```

### Attach save file

n/a

### Steps to reproduce

Goto spawn items with Xedra Evolved loaded.  Look at the two estocs from Xedra Evolved.  See that they show identical statblocks.  Compare the raw json files and see that they are not identical once enchantments are considered.

### Expected behavior

[...]

PR #64480, opened 2023-03-23
Apply Enchantments to Item Info

H3LL-J3F3:
#### Summary
Bugfixes "Enchantment effects on attack cost and dps now reflected in item info"

#### Purpose of change
Enchantments not reflected in an item's moves per attack and dps as mentioned in #63982. This change reflects these effects the same way a Stat change would affect dps. 

#### Describe the solution
When calculating moves per attack, also apply enchantment impact(s). When calculating dps, also apply enchantment damage.


H3LL-J3F3:
Think this is a non-issue. After further testing the user will see a different dps value once wielded due to the way enchantments are handled (the passive effects must be applied to the character before they can be reflected in the dps calculation).

PR #64485, opened 2023-03-23
Differentiate welding tool names

RenechCDDA:
#### Summary
None

#### Purpose of change
People have repeatedly, *repeatedly* gotten "welding rig" and "welding kit" mixed up.

#### Describe the solution
Change the display name so it's very obvious which is which, and that that the former 'welding kit' can do more, different things.

N/A

#### Testing
Github tests guide my commits!

#### Additional context
@Drew4484 How does this look? This is the difference between the two as far as I'm aware.

### DIFF
--- a/data/json/furniture_and_terrain/appliances.json
+++ b/data/json/furniture_and_terrain/appliances.json
@@ -225,7 +225,7 @@
   {
     "type": "vehicle_part",
     "id": "ap_weldkit",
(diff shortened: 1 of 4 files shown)


Drew4484:
This is fine, it definitely is distinct from the original. That it doesn't really describe the gear is probably less important than players being able to distinguished the two without knowing anything about welding.

Issue #64438, opened 2023-03-21
JSON Object unvisited member checking no longer works


anothersimulacrum:
### Describe the bug

JSON unvisited member checking was introduced in https://github.com/CleverRaven/Cataclysm-DDA/pull/34735 and turned on outside of tests in https://github.com/CleverRaven/Cataclysm-DDA/pull/45635.
It provides an error on loading the game when a JSON object has data in it that the game makes no attempt to read, which is a significant help with debugging.

A bisect gives the first bad commit as any of
97d6a4836dbe41f7d5155d028d5b38c537acae10
ec3f896d5b82938a91bdcbbf71a4e94837741fdf
2833d7a974f9ea8194b15ae5b4dcb13b109586bc
ca9428d3eb71d9dfa468bad3dae25df0282d9e99
898708829f0318152b17a7dd9dfe3a31f4780d14

Clearly, the new flexbuffer `JsonObject` visited member reporting is not working.

### Attach save file

N/A

### Steps to reproduce

Apply this patch
```diff
diff --git a/data/json/items/obsolete.json b/data/json/items/obsolete.json
index 2b023a1ab1..1f2afd98da 100644
--- a/data/json/items/obsolete.json
+++ b/data/json/items/obsolete.json
@@ -26,6 +26,7 @@
     "price": 500,
     "price_postapoc": 100,
     "weight": "300 g",
+    "BOB": "ALICE",
     "volume": "250 ml",
     "material": [ "plastic" ],
     "looks_like": "mre_beef_box",
diff --git a/data/json/monsters/mi-go.json b/data/json/monsters/mi-go.json
index c3208f5c72..777e919e3c 100644
--- a/data/json/monsters/mi-go.json
+++ b/data/json/monsters/mi-go.json
@@ -5,6 +5,7 @@
     "name": { "str": "mi-go" },
     "description": "An alien creature of uncertain origin.  Its shapeless pink body bears numerous sets of paired appendages of unknown function, and a pair of ribbed, membranous wings which seem to be quite useless.  Its odd, vaguely pyramid-shaped head bristles with numerous wavering antennae, and it moves with an uncanny fluidity on its many legs.",
     "default_faction": "mi-go",
+    "BAR": "FOO",
     "bodytype": "migo",
     "species": [ "MIGO" ],
     "volume": "92500 ml",
```
The game does not report that it fails to visit the `BAR` and `BOB` members, when there is no code reading them.
[...]

PR #64494, opened 2023-03-23
Fix unvisited JSON members

anothersimulacrum:
#### Summary
None

#### Purpose of change
Working towards re-enabling unvisited JSON member checking (see https://github.com/CleverRaven/Cataclysm-DDA/issues/64438).
Fix JSON errors.

#### Describe the solution
Enable unvisited member checking, game complains about these.

#### Additional context
Mods still need to be done.
Alerting the following people in case these data errors need other solutions
@dseguin @Venera3 https://github.com/CleverRaven/Cataclysm-DDA/commit/48b36faec137562960964170727325463a99c004
@MNG-cataclysm @LovamKicsiGazsii https://github.com/CleverRaven/Cataclysm-DDA/commit/a773ecec5e14e5d5419a68574ee70ebdf4af95a5 https://github.com/CleverRaven/Cataclysm-DDA/commit/70dc95c06db3e4659c1452609f8b25b362d15a39 https://github.com/CleverRaven/Cataclysm-DDA/commit/3aca0a2fd56ac5e6fc1c2a39d2247960b24edb14
@GuardianDll https://github.com/CleverRaven/Cataclysm-DDA/commit/8ad015a6af7073b1c0960bd46056cc690652ab6b https://github.com/CleverRaven/Cataclysm-DDA/commit/673a65cc15a0d10722ca0b0f5ac1e3b0a84467d8


Venera3 (approved):
Dodge encumb got scored properly, some dumdum forgot to remove the field. Thanks!

data/json/mutations/mutations.json
```diff
@@ -7959,7 +7959,6 @@
     "points": 99,
     "valid": false,
     "description": "If you need to survive the deadliest bug attacks.",
```
GuardianDll, line 7961:
```suggestion
    "description": "If you need to survive the deadliest bug attacks.",
    "enchantments": [ { "values": [ { "value": "MAX_HP", "add": 9000 } ] } ],
```

data/json/mutations/mutations.json
```diff
@@ -7969,7 +7968,6 @@
     "points": 99,
     "valid": false,
     "description": "Heal yourself from this stupid bug attacks.",
```
GuardianDll, line 7970:
```suggestion
    "description": "Heal yourself from this stupid bug attacks.",
    "enchantments": [ { "values": [ { "value": "REGEN_HP", "add": 100 } ] } ],
```

anothersimulacrum:
It will work with no `to` field defined, but it expects a string for that field if it is defined. I'll remove the field.

FarkasKBence (approved):
At first I was confused why you removed the `repeat`s from the private resort, but then it dawned on me that distribution doesn't work like that.🤦

GuardianDll (approved):
Kinda sad the variants do not accept variety in stats, but thats okay

PR #64496, opened 2023-03-23
Cut more 0.G obsoletions

irwiss:
#### Summary
None

#### Purpose of change

Remove more obsoleted stuff

#### Describe the solution

✂ all kinds of obsolete stuff

Some decisions made:
* Terrains the previously had bashed into t_console_broken + pile of part items now break into t_metal_floor + pile of part items
* Obsoleted item_group `vehicle_scrapped` is resurrected as it's being used, could possibly be folded into `metal_wreckage` but I'm not sure if they're strictly equivalent, either way this should be trivial find&replace in follow up PR
* `building_rubble` was still used - replaced with `collapsed_building_rubble` they seem fairly similar in use and this one is more complete
* Obsolete item `spidery_legs_big` - resurrected, used in uncraft of exodii corpse
* Bunch of remaining t_some_terrain_that_is_now_furniture was replaced with f_actual_furniture
* Computer interaction lookup for `t_plut_generator` was removed - just assumes there's a hidden generator somewhere

commits c6bd970a5332ac3e38fa98b0ec1e192cd3ac95af, 929560e8711b91f519f295d98798cffbdf7939f0, 7ba5a4a6ddfacbe803b5ef2cb3bcdbc23a3c8e55 add migrations:
* monsters deserialized with invalid type are converted to mon_breather (or game spams debugmsgs)
* monsters deserialized with invalid faction are converted to "factionless" (or game spams debugmsgs)
* invalid ter_ids in already-generated maps are converted to t_dirt (or game spams a tile of "Nothing.")

commit 0a927162a86a95f924bb1699baaddadaa0c65bc4 removes 9 years old tool (except for linting pass for flake test)

https://github.com/CleverRaven/Cataclysm-DDA/commit/c76a9a54c64f182e5af43b022d846b8faf837cc3 adds and updates some docs regarding migration/obsoletions locations and obsoletion of mapgen

https://github.com/CleverRaven/Cataclysm-DDA/pull/64496/commits/949cd83c0128abb31f51fd22b0b0b8373b7ea13f redirects "find / lead npc to farm" missions to the Mutable Farm

Other commits cut content

#### Testing

Local tests ran OK, loaded older save, moved around, long-teleported to a few towns


src/savegame_json.cpp
```diff
@@ -4999,10 +5009,22 @@ void submap::load( const JsonValue &jv, const std::string &member_name, int vers
                     if( !remaining ) {
                         JsonValue terrain_entry = terrain_json.next_value();
                         if( terrain_entry.test_string() ) {
-                            iid = ter_str_id( terrain_entry.get_string() ).id();
+                            const ter_str_id terstr( terrain_entry.get_string() );
```
anothersimulacrum, line 5012: I'd like if we could have some sort of JSON-defined migration for this, but I understand that's even more scope for this PR.

anothersimulacrum: I'd also appreciate if you could add a test for this (tests/submap_load_test.cpp), but it's by no means required.

irwiss: I added the test just in case, but I think migrating terrain ids is best left for a separate PR

anothersimulacrum: I agree. Thank you!

PR #64505, opened 2023-03-23
Make fitness band use UPs power

ghost:
#### Summary
Bugfixes "Fitness Bands now use UPS power"

#### Purpose of change
the ability to check the time, calories, exercise level and heartbeat without needing a single joule of power. Batteries were too cumbersome and annoying, so i went ahead with UPS instead.

#### Describe the solution
Take the fitness band and make it use UPS power, Now to use the features, except for the clock. you can activate the band, which uses a single charge out of 20 to tell you exercise level, heartbeat, And calories burnt & consumed.

debated just leaving it be, but it had too many features :p

#### Testing
Debugged spawned in a band, and works as intended

### DIFF
--- a/data/json/items/tool_armor.json
+++ b/data/json/items/tool_armor.json
@@ -526,10 +526,12 @@
     "price": 5000,
     "price_postapoc": 100,
     "material": [ "plastic" ],
-    "flags": [ "WATCH", "WATER_FRIENDLY", "BELTED", "FRAGILE", "ALLOWS_NATURAL_ATTACKS", "OVERSIZE" ],
+    "flags": [ "WATCH", "USE_UPS", "NO_UNLOAD", "NO_RELOAD", "WATER_BREAK", "BELTED", "FRAGILE", "ALLOWS_NATURAL_ATTACKS", "OVERSIZE", "ELECTRONIC" ],
     "symbol": "[",
+    "ammo": [ "battery" ],
+    "charges_per_use": 1,
     "use_action": [ "FITNESS_CHECK" ],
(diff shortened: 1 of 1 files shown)


KHeket:
Can you increase energy capacity to 100? Because in reality fitness bands can work without charging for a few days, and 20 chagres - not enough

RAldrich:
Making more things use UPS seems like a bad idea. The only things that use it now are smartphones and power armor, and the only reason smartphones use it is because we didn't have a reasonable simulation of power cords at the time. There's a PR out to enable items to use power plugs, and when that's merged we should look at eliminating UPS entirely.

flybye3o6:
i assumed the fitness band had a kinetic charger so it would gain charges as you moved
thats why it didnt have battery charges because it needed someone to work on items generate power from movement
my suggestion, add kinetic chargers to the game as a generator source converting character momentum into kj, add that ability to the fitness band with an unremovable battery, making the fitness band only be chargeable by running around a bit

andrewhr:
@RAldrich sorry to highjack here, but your comment about smartphones is something I've been wondering for a few days. My feeling is UPS is something that will eventually be deprecated, given better alternatives today, there is somewhere with more information about it to confirm or give more nuance? Thanks in advance!

RAldrich:
Check out #64334

Devengel:
> Making more things use UPS seems like a bad idea. The only things that use it now are smartphones and power armor, and the only reason smartphones use it is because we didn't have a reasonable simulation of power cords at the time. There's a PR out to enable items to use power plugs, and when that's merged we should look at eliminating UPS entirely.

ooh, that sounds good! i'll close the PR then. no point in messing with it

Issue #64287, opened 2023-03-16
Warning about liquid ammonia missing when starting crafting


TheDarkMaster13:
### Describe the bug

When you try to craft a recipe that uses liquid ammonia, you'll get a warning that it is missing and be asked to consume a new unit.  Saying yes will repeat the warning.  Saying no, but then resuming crafting, works as intended as a work around.

Possibly this is a canister issue?

### Attach save file

N/A

### Steps to reproduce

1. Get a liquid ammonia canister and the rest of the ingredients needed to craft something.  Ex: Mutagen.
2. Attempt to craft the recipe.  Saying yes repeats the prompt.
3. Say no, then work on the craft as a workaround.

### Expected behavior

The recipe should consume the ingredients as normal and proceed straight to crafting with no warning prompt.

### Screenshots

![image](https://user-images.githubusercontent.com/6277016/225507285-c27e52b5-1ba7-416f-be01-9f32d91fe293.png)

### Versions and configuration

- OS: Windows
    - OS Version: 6.1 7601
- Game Version: d6ec466 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Aftershock [aftershock],
    Generic Guns [generic_guns]
]

TheDarkMaster13:
Actually the problem is more confusing than I thought it was, as you sometimes do have to consume the ammonia and sometimes you don't.

RenechCDDA:
Is the ammonia actually being destroyed? There was a discourse "report" after this issue was previously cleared up with a different sort of reproduction: https://discourse.cataclysmdda.org/t/liquid-ammonia-evaporating-while-cooking-mutagen/28503/2

PatrikLundell:
The description sounds like a similar issue with clay canning pots. When using any recipe ingredients stored in a clay canning pot such a prompt appeared, and the ingredients were consumed twice: first when the crafting failed, and then again when it was resumed. It was reported that the cause was that the clay canning pot had some tool properties (I believe more items were affected, but that's the one I specifically know of).

RenechCDDA:
I've now specifically tested this and in no circumstance on d6ec466 (0.G) can I get this to happen. Please provide a reproduction save.

[...]

PR #64506, opened 2023-03-23
Fix underground emitter drain

irwiss:
#### Summary
None

#### Purpose of change

Fix battery drain for emitters and underground vehicles

#### Describe the solution

A few bugs clumped together and trigger a sanity check in https://github.com/CleverRaven/Cataclysm-DDA/issues/64287#issuecomment-1480617192 but all of them together sorta cancel each other so they were effectively invisible until the sanity check was added in #64097 to prevent "discharging" negative amounts

This may cause slightly more bug reports as I added debugmsgs when charging or discharging a negative amount, but driving around and crafting didn't seem to trigger these after the fix in this PR

This PR fixes:
* Appliances don't initialize `last_update` to current turn, creating massive time intervals between `calendar::turn` and `vehicle::update_time` on first turn - now new appliances will initialize it
* Vehicles located at zlevel < 0 don't update their `last_update` field, which makes the interval `vehicle::update_time` needs to simulate only grow upwards overflowing the int variables - this is now fixed and last_update will tick forward at any zlevel
* Parts which are both emitter and VPFLAG_ENABLED_DRAINS_EPOWER were being double dipped - batteries discharged once in `vehicle::power_parts()` with the rest of VPFLAG_ENABLED_DRAINS_EPOWER parts and once in `vehicle::update_time` effectively doubling power consumption, but the integer overflow canceled the drain in `vehicle::update_time` so power use was appearing "normal". This PR changes `vehicle::update_time` to only discharge the amount for simualted interval minus 1 turn - so if vehicle is in bubble and power_parts() is running it doesn't discharge anything, if vehicle comes into bubble with larger interval it should simulate discharge up to current turn, and then power_parts will do the last discharge.

Also PR moves time travel protection upwards so emitters are also guarded against going back in time

#### Testing

Start some scenario to put you in evac center at summer like normal (or some date, just make sure it's away from turn 0)
Enable debug mode, disable all filters, enable only DF_VEHICLE
Build a battery appliance + small space heater near each other, put some charge in battery via debug
Turn on heater and start looking at the debug message printouts, or put breakpoints in vehicle::update_time and see the intervals simulated and int overflow feeding negative amounts to discharge_batteries

May be automated tests will catch something

### DIFF
--- a/src/veh_appliance.cpp
+++ b/src/veh_appliance.cpp
@@ -68,6 +68,7 @@ void place_appliance( const tripoint &p, const vpart_id &vpart, const std::optio
         veh->install_part( point_zero, vpart );
     }
     veh->name = vpart->name();
+    veh->last_update = calendar::turn;
 
     // Update the vehicle cache immediately,
     // or the appliance will be invisible for the first couple of turns.
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -7456,6 +7456,7 @@ void vehicle::update_time( const time_point &update_to )
     }
 
     if( sm_pos.z < 0 ) {
+        last_update = update_to;
         return;
     }
 
--- a/src/vehicle.cpp
+++ b/src/vehicle.cpp
@@ -7431,14 +7431,18 @@ static bool is_sm_tile_outside( const tripoint &real_global_pos )
 
 void vehicle::update_time( const time_point &update_to )
 {
-    double muffle;
-    int exhaust_part;
-    std::tie( exhaust_part, muffle ) = get_exhaust_part();
+    const time_point update_from = last_update;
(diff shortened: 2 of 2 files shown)


NetSysFire:
Unsure if this might be a sensible thing to backport. It introduces a new string but debugmsgs should never be translated (right?).

irwiss:
I'm not sure how popular appliances in basements are but overflow will feel buggy, and will mutate the buggy behavior as calendar advances

If this is a valid candidate I suppose it's possible to do a smaller PR for backport without debugmsg, most of the changes are just more validation checks not neccesary in backport, the important changes are the one line veh_appliance.cpp change and 
a few lines in vehicle::update_time

anothersimulacrum:
These debug messages are not translated, no new strings would be added.

PR #64511, opened 2023-03-24
friendly monsters spawn friendly offspring

ekaratzas:
#### Summary
Features "friendly animals spawn friendly offspring"

#### Purpose of change

This change is a QoL improvement for farmer players. The idea is that if the parent animal is friendly to the player, their offspring should be friendly too. Right now players have to be extra careful while going through their pastures to not accidentally 'attack' some of the younger animals because there's a mixture of older/friendly animals and younger, neutral ones. This is especially bad when dealing with a chicken coop that has a grown a bit large. In the long run, this change might also help when dealing with dynamically created pastures of friendly animals.

#### Describe the solution

PR works by piggybacking on existing interfaces and functionality for the most part. Adding friendly offspring for mammals like cows was straightforward by just overriding the default false 'friendly' parameter of map::add_spawn() with the attitude of the parent towards the player.  Birds that hatch from eggs required an extra flag that is added to the egg item (flag_SPAWN_FRIENDLY) which is then used by map::rotten_item_spawn() to determine if the newly spawned creature should be friendly or not.

Existing behavior seemed super clunky to me as a player so this QoL made sense, to me at least. Code-wise, the solution seems like the most straightforward and least invasive way to go about it.

#### Testing

1. Spawned cows/chickens near player, neutral at first.
2. Locally edited monster::try_reproduction() and item::process_temperature_rot() in order to ensure that the creatures would spawn an egg or baby animal every time the submap would get loaded.
3. Player goes away, comes back, spawned offspring is neutral for cow / chicken.
4. Tried same experiment but starting with friendly parents. Player comes back to submap and offspring is generated friendly.
5. Also locally hacked the evolution code in monster.cpp in order to check whether their evolved variants would also retain the same attitude towards player and that worked as well.

### DIFF
--- a/data/json/flags.json
+++ b/data/json/flags.json
@@ -538,6 +538,11 @@
     "//": "This item will spawn active. Not advised for items that drain power when active."
   },
   {
+    "id": "SPAWN_FRIENDLY",
+    "type": "json_flag",
+    "//": "This item will spawn friendly. Used for pets and such."
+  },
+  {
     "id": "STAB",
     "type": "json_flag"
   },
--- a/src/flag.cpp
+++ b/src/flag.cpp
@@ -303,6 +303,7 @@ const flag_id flag_SOFT( "SOFT" );
 const flag_id flag_SOLARPACK( "SOLARPACK" );
 const flag_id flag_SOLARPACK_ON( "SOLARPACK_ON" );
 const flag_id flag_SPAWN_ACTIVE( "SPAWN_ACTIVE" );
+const flag_id flag_SPAWN_FRIENDLY( "SPAWN_FRIENDLY" );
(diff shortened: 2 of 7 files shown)


kevingranade:
Not autoattacking is a legit concern, but automatically marking an animal you may have never seen as friendly is not a reasonable solution.
Supressung autoattack on any creature you CAN tame seems like a good approach.

ekaratzas:
Outside of auto-attacking concerns, the alternative would be to semi-regularly have to manually tame all new spawned farm animals in order to interact with them. Seems very clunky from a player experience perspective but at the end of the day it's your game.

Zireael07:
That alternative sounds like tedium and I thought CDDA was against tedium?

PR #64520, opened 2023-03-24
bomb teleport wtf

kevingranade:
#### Summary
None

#### Purpose of change
Exodii teleporting a bomb at you if you make them mad doesn't work from a gameplay, thematic or believability point of view.

#### Describe the solution
Remove the bomb teleporting event.


MNG-cataclysm:
Did you try to rob Rubik's store again?

PR #64531, opened 2023-03-24
Rebalance Bordered scenario

Night-Pryanik:
#### Summary
Balance "Rebalance Bordered scenario"

#### Purpose of change
Judging by several videos of streamers playing this scenario, 180x180 OMTs is way too much for a "challenge" feel, and some players don't even realize that there's actually a Wall encircling their starting overmap.

#### Describe the solution
Instead of giving the player the whole overmap with its 180x180 OMTs, reduce available space to 100x100 OMTs, centered on player's location. In case player was spawned more or less close to overmap's border, the available space might be even lower, because now the game dynamically adjusts available space based on proximity to the overmap's borders.

Limit the space inside the Wall to 50x50. Maybe something in-between? I don't know what's better.

#### Testing
Started game with this scenario several times, checked that the Wall correctly spawns and correctly adjusts its size if player is too close to overmap's border.

#### Additional context
100x100 square
изображение

50x50 square
изображение


anoobindisguise:
Why not make all the surrounding terrain endless black nether glass? It might not need to get smaller even. But one of the big things I heard about the Bordered challenge was that you really were *not* supposed to leave the border and it would have been made of a bedrock equivalent if possible.

Night-Pryanik:
I think leaving an option to criss cross the border is a good thing. Players can play inside the perimeter AND can cross the border if they want. Two options are better than one.

PR #64534, opened 2023-03-24
Fix crash with preventing death

### LINKED ISSUE #64448: Game crashing on player death when using a custom EoC of type "PREVENT_DEATH"
### Describe the bug

I have written a simple test mod that uses the PREVENT_DEATH EoC to prevent dying and the game is crashing occasionally when it is triggered, in particular if a teleport is used (debug menu and u_teleport both exhibit this behaviour) before dying again. It appears to only happen when dying from a monster's attacks. It is inconsistent in it's occurance, but an almost sure-fire way to get it to happen is with the provided save file. I have also had this crash occur even without using a teleport, but that is much rarer and hard to reproduce. I have stripped down the player reviving EoC as much as possible, and the issue persists.

### Attach save file

[test-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11034781/test-trimmed.tar.gz)

### Steps to reproduce

1. Use long-range teleport to go to another overmap tile, with monsters present.
2. Allow yourself to die to the monsters.
3. In most cases, the game will crash upon death. Sometimes it does not happen unless multiple deaths are incurred. This save seems to fairly consistently cause the crash with this method.

### Expected behavior

The EoC should function consistently and not crash the game when being triggered.

### Screenshots

_No response_

OS: Windows 10 Pro 64-bit
Version: 0.G stable
Graphical Version: Tiles
Ingame Language: English
Mods Loaded: default mods and my test mod provided with the report.
[eoc_crash_test.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/11034804/eoc_crash_test.zip)

### Additional context

[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/11034805/crash.log)
[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/11034806/debug.log)

Ramza13:
#### Summary
None

#### Purpose of change
Fixes #64448

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1836,6 +1836,11 @@ void Creature::set_killer( Creature *const killer )
     }
 }
 
+void Creature::clear_killer()
+{
+    killer = nullptr;
+}
+
 int Creature::get_num_blocks() const
 {
     return num_blocks + num_blocks_bonus;
--- a/src/creature.h
+++ b/src/creature.h
@@ -1190,6 +1190,7 @@ class Creature : public viewer
         virtual bool is_symbol_highlighted() const;
 
         std::unordered_map<std::string, std::string> &get_values();
+        void clear_killer();
 
     protected:
         Creature *killer; // whoever killed us. this should be NULL unless we are dead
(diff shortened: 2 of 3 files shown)


NetSysFire:
This one looks pretty crucial to backport.

PR #64535, opened 2023-03-24
mercenary coat coverage fix

anoobindisguise:
#### Summary
Bugfixes "merc coat has the correct coverage"

#### Purpose of change
Mercenary coat per its comment is supposed to have the stats of a US ballistic vest and is made from a US ballistic vest but it did not have the same coverage, US ballistic vest has 95% while it only had 90%.

#### Describe the solution
Bump up to 95%.


NetSysFire:
Could maybe be backported, too.

PR #64539, opened 2023-03-25
Improve dish: succotash and add a deluxe version

NetSysFire:
#### Summary
Content "Improve Succotash and add a deluxe version"

#### Purpose of change

While I was doing unrelated research, I stumbled upon https://en.wikipedia.org/wiki/Succotash and I saw that the version ingame could have some tweaks.
Continuation of #64497 - I couldnt use that one anymore since I had to rebase and thus force push.

#### Describe the solution

- Improved description, making it more interesting.
- Made it spoil quicker because it contains much water.
- Because of the above, also add 1 quench.
- Reduced fun from 2 to 1. It is a filling meal and even better if its hot, but it will not make you very happy.
- Adjusted vitamins based on different sites I looked at.
- Moved recipe from recipe_food.json to vegetable_dishes.json
- Instead of beans you can also use lentils as they are also a commonly used legume. Its the apocalypse, its gonna work.
- Seasoning is now required.
- Better batch time savings, but recipe takes slightly longer to make now.
- The deluxe version is basically just like the basic one, but with higher enjoyability and calories at the cost of some more ingredients.

#### Testing

None yet since it is a quite simple addition.


NetSysFire:
It looks like the tests fail because of unrelated reasons, since they complain about dodging.

anoobindisguise:
You're not using the correct cooking costs, you can even see how they were defined when you deleted the recipe in recipe_food.

PR #64548, opened 2023-03-25
Protective Lenses buff (again because stalebot closed my first one)

anoobindisguise:
#### Summary
Balance "improve protective lenses CBM"

#### Purpose of change
Followup on #62834
Protective lenses were a little on the weak side to the point that you didn't really want to install this bionic.

#### Describe the solution
Encumbrance 5->1 (note that 1 eye encumbrance is still noticeable if it takes you down from a whole dodge number. 1 eye encumbrance is how much encumbrance sunglasses give, and these are highly personalized + similar to sunglasses)
Skintight->PERSONAL
Plastic 0.5mm->0.2mm + Aluminum 0.05mm (overall armor slightly lower)

Make it a full on "visor" as is common for various cyborgs, but this one specifically is a Molly Millions reference so I'd like to keep it that way.

#### Testing
Confirmed functional in game.

#### Additional context

lenses2


kevingranade:
Please leave this kind of thing as a more targeted change, the only meaningful change is to encumbrance,  so just make that change.

data/json/items/armor/integrated.json
```diff
@@ -938,20 +937,10 @@
...
-      "UNBREAKABLE",
-      "SKINTIGHT",
-      "SOFT",
-      "WATER_FRIENDLY",
-      "SUN_GLASSES",
-      "NO_REPAIR",
-      "SWIM_GOGGLES",
-      "NO_SALVAGE"
-    ],
+    "flags": [ "INTEGRATED", "UNBREAKABLE", "PERSONAL", "SOFT", "WATER_FRIENDLY", "SUN_GLASSES", "NO_REPAIR", "SWIM_GOGGLES" ],
```
anothersimulacrum, line 943: Why remove NO_SALVAGE?

anoobindisguise: I don't remember the original reason. I can put it back but integrated armors will automatically have NO_SALVAGE nowadays.

anothersimulacrum: https://github.com/CleverRaven/Cataclysm-DDA/pull/64449 suggests that is not the case?

anoobindisguise: You are right. I assumed that pr fixed it that way but it didn't.

anoobindisguise:
Other than that, does this seem too strong balance wise? Too much of a leap realism wise? I can tone stuff down or even reflavor it into some kind of cyber-visor if molly millions lenses aren't acceptable

kevingranade:
The encumbrance change is fine,  strip out everything else.

PR #64560, opened 2023-03-25
Fix diary text navigation wrapping

Qrox:
#### Summary
Bugfixes "Fix diary text navigation wrapping"

#### Purpose of change
It incorrectly uses the length of the diary text for the number of lines, causing it to wrap incorrectly.

#### Describe the solution
Calculate and use the number of lines instead. Also test size in `navigate_ui_list` to avoid unsigned integer underflow.

#### Testing
It now wraps correctly when pressing up/down and page up/down.

#### Additional context


NetSysFire:
Looks like another good backport candidate.

Qrox:
It was caused by a post-0.G change so no need for backport I think.

PR #64562, opened 2023-03-25
Add false eoc to Vampvirus checks in Xedra Evolved

Maleclypse:
#### Summary
None

#### Purpose of change
Eliminate the roll remainder messages that continue on after the player has gained all the abilities for that stage.

#### Describe the solution
Ramza said that I needed to use the false_eoc array and since I couldn't find any examples of it in code I've attempted to make one real quick.

None

#### Testing
Tested and it works better than before.


Maleclypse:
Candlebury suggested replacing the ticking of the effect decay with an eoc inside the existing daisy chain similar to 
```
"run_eocs": [
                      {
                        "id": "EOC_PERK_COMBAT_TEMPO_ADD_STACK",
                        "condition": {"or": [{ "u_has_effect": "tempo_counter", "intensity": 2 }, { "u_has_effect": "tempo_ready"} ]},
                        "effect": [
                            {"u_add_effect": "tempo_ready", "duration": "1 seconds"},
                            { "queue_eocs": "EOC_PERK_COMBAT_TEMPO_ADD_STACK", "time_in_future": [ "2 seconds", "2 seconds" ] }
                        ],
                        "false_effect": {  "u_add_effect": "tempo_counter", "duration": "2 seconds"}
                  } 
                ] 
```

Issue #57081, opened 2022-04-23
Mine filled with drive-through restaurants


actually-a-cat:
### Describe the bug

The underground portion of a mine generated filled with over a dozen copies of "abandoned drive-through". They're all identical, tightly packed, and empty - the only items and monsters on the level are in the mine entrance section.

### Steps to reproduce

Unclear. I checked out another mine that I think was also freshly generated on this build and it was normal. 

### Expected behavior

Something that makes sense in a mine.

If it's intentional and meant to be some sort of strange facility I don't understand, it looks much more like a bug than that.

### Screenshots

![mine1](https://user-images.githubusercontent.com/102094270/164935706-67f5e836-b9bc-43cf-9334-d131f5fc6a98.png)
![mine2](https://user-images.githubusercontent.com/102094270/164935715-949c23e8-19f0-46b1-91e7-87dcfc725a09.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19042.985 (20H2)
- Game Version: 2eef1a1 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth]
]

### Additional context

Save in the mine: 
[Marquette.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/8548194/Marquette.zip)

haveric:
Pretty sure the abandoned drive through is the fallback for when something doesn't exist, so it's either a bug or you had one generated before it got changed (assuming there might have been a bugfix after the initial version), which would have replaced it. Definitely not intentional though.

toweroglass:
I actually just saw this in my map when i was going up and down across z-levels and thought the bug was that it was labeled incorrectly. i didn't go down and see it for myself.

hexagonrecursion:
likely related: #56555

Night-Pryanik:
Have anyone encountered this issue since April?

Issue #45734, opened 2020-11-30
Corrupted save after updating. Overhauls the whole area


scraw-darkblade:
Corrupted save after updating. Overhauls the whole area

#After updating received couple of errors and when I loaded my game, It was a different area.
It is dark that time (like 4:00AM) and as you walk and discover the previous tiles it is replaced by new one's. Items that are supposed to be there is also removed.

Eg: Shrub that was supposed to be there (You can see it on unseen status, and when you walk to check it, it is gone. replaced by other tile. 

# Steps To Reproduce

1. Update the game from the recent one. 
2. Load the game.
3. Walk as you see terrain is changing while doing so.

# Expected behavior

It should work as it is no changes on tiles or any kind of change without manipulation by character, monster/NPC or natural occurences.

# Versions and configuration
- OS: Windows
    - OS Version: 10.0 1909
- Game Version: 0.E-7938-g8736e6b [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Bionic Professions [package_bionic_professions]
]
  

# Additional context

Limitations:
I am not able to check the whole world if it changed as well.
I am not able to load the previous save (forgot the build) but I provide both save.
[debug (New).log](https://github.com/CleverRaven/Cataclysm-DDA/files/5614806/debug.New.log)
[debug (old) .log](https://github.com/CleverRaven/Cataclysm-DDA/files/5614807/debug.old.log)
[Errors after update.txt](https://github.com/CleverRaven/Cataclysm-DDA/files/5614808/Errors.after.update.txt)

scraw-darkblade:
[save (new).zip](https://github.com/CleverRaven/Cataclysm-DDA/files/5614873/save.new.zip)
[save (old).zip](https://github.com/CleverRaven/Cataclysm-DDA/files/5614876/save.old.zip)

ZhilkinSerg:
If you will try to load your savegame in a previous game version (that worked properly previously) and it won't work that would mean your files are corrupted and there is nothing wrong with new game version.

Corrupted savegames is an issue in your environment we cannot help you with besides suggesting to do backups regularly.

jeremyshannon:
[...]

Issue #45631, opened 2020-11-25
Bad terrain when loading old saves


jeremyshannon:
# Describe the bug

Using saves from 0.E or earlier results in multiple
 "DEBUG    : Loaded bad ter! ter house_south"
and other bad terrain errors, including all the old random houses, 
"DEBUG    : Loaded bad ter!  ter haz_sar_b1_north"
 and other hazardous waste sarcophagus terrains.
 There may be other terrain  problems buried under the tons of house errors.

It then replaces all of these things with abandoned drive-ins.

# Steps To Reproduce

Load a save generated before random houses were removed.

# Expected behavior

It should be replacing those old tiles with modern versions, or a nearest equivalent.

# Screenshots
![good terrain](https://user-images.githubusercontent.com/23157047/100276093-d2a85d00-2f15-11eb-81c4-2c3070413027.png)
The town of Keene in 0.E

![bad_terrain](https://user-images.githubusercontent.com/23157047/100276125-ddfb8880-2f15-11eb-8dd9-f432f62af751.png)
The town of Keene in latest experimental.

# Versions and configuration

Linux, 64-bit, ascii, version tag b11188

# Additional context
Rob Keys has been driving east for three and a half years now, usually in various experimental versions. His overmap is a history of CDDA mapgen over that period, and I feel like it'd be a shame to have pieces get erased.

ZhilkinSerg:
Saves older than 0.E aren't really supported.

wapcaplet:
@jeremyshannon What version did you save your game in? Was it 0.E, or an earlier version? If you're using an older one like 0.D, you might try saving in 0.E or 0.E-2 first (making backups of course) before upgrading to experimental.

jeremyshannon:
I believe this particular save was last updated in experimentals just prior to the nested containers update. (I have a bunch of others similarly affected.) 
 Re-saving it in 0.E just now, and testing again in master, and this is interesting. In 0.E, I scrolled over the previously affected areas in the overmap to make sure everything was fine there, then saved. Then in current master, those areas are fine, but if I scroll just a little further to the west, the error recurs.
 The migration code is only affecting the area currently viewed, and isn't running on the rest of the map, which sorta makes sense, but it must be how this bug happened in the first place. I guess I would need to manually scroll over the whole thing in 0.E to  ensure it all gets migrated. That's kind of a PITA. Especially if I have to do it for every save.
[...]

PR #64564, opened 2023-03-25
Unhardcode oter_id migrations

### LINKED ISSUE #64592: Dirt roads and other terrain turning into abandoned drive-throughs at scale
### Describe the bug

Overland map tiles keep turning into abandoned drive-throughs . At first it was just one or two forest paths, but now almost every dirt road in a large farm has done it.

When terrain has already been visited, the drive-through shows on the overland map, but doesn't change the already generated terrain. For example, the yard outside my old base turned into a drive-through on the overland map, with no ill effects to the yard remaining a yard when visited on the zoomed-in map.

However if the terrain hasn't been visited before, it will generate actual drive-throughs, one after another. (See screenshot below)

It looks like any sort of terrain can flip, but roads and paths seem most common. Whenever a tile flips, it's *always* to a drive-through. No other weird terrain transformations have been seen.

### Attach save file

[NoHope-2023-03-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11071650/NoHope-2023-03-trimmed.tar.gz)

### Steps to reproduce

I wish I knew how to reproduce this! It seems to happen at random. However you can load the attached file to see an example of a shattered reality, with lines of drive-throughs intersecting still corn fields, fading away into the distance.

### Expected behavior

Not everything slowly becomes an abandoned drive-through.

### Screenshots

These were dirt roads in the farm before they transformed:

no-roads-only-drive-through

farm-of-movies

```
- OS: Linux
    - OS Version: Distributor ID: Ubuntu; Description: Ubuntu 20.04 LTS; Release: 20.04; Codename: focal; 
- Game Version: 0.G-934-g5a1e67f196 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Slots [cbm_slots],
    No Hope [no_hope]
]
```

Version is d6d06eb1146d46c8eb4e1f9f74f215f80f211380 (`cdda-experimental-2023-03-26-0619`) with some local patches applied, none of which have anything to do with mapgen.

### Additional context

While only a hunch, I suspect this may be related to the weird behaviour we've seen in #64588, #64552, and other issues that suggest we've got a stray pointer or other memory weirdness going on.

irwiss:
#### Summary
None

#### Purpose of change

Fixes #64592

Turns out `obsolete_terrain` entry in json wasn't a "jsonized" obsoletion, but a pre-filter list for the hardcoded C++ migration. This migration itself lives in C++ if/elseif/... ladder and has no `else` for undefined migration so effectively it's a no-op unless you define the migration in **both** json and C++, this produces no warning if there is no corresponding branch to take in C++.

The obsoleted overmap_terrain entries were still kept in data files so other places just looked them up and no errors raised from there either, removing them caused obsoleted_terrain entries from 0.F transform into "abandoned drive-through" since that happens to be assigned to ter_id 0.

Related issues: https://github.com/CleverRaven/Cataclysm-DDA/issues/57081 https://github.com/CleverRaven/Cataclysm-DDA/issues/45631 https://github.com/CleverRaven/Cataclysm-DDA/issues/45734 (although here it looks like a submap or several were lost)

#### Describe the solution

This PR refactors how oter_id should be obsoleted/migrated - you now remove the overmap_terrain entry entirely, and define an entry similar to the following:
```json
{
  "type": "oter_id_migration",
  "oter_ids": {
    "hive": "omt_obsolete",
    "hive2": "omt_obsolete",
    ...
  }
}
```
This is mentioned in updated docs. Missing entries will raise debugmsg and migrate to `omt_obsolete` as well

Hardcoded C++ migrations from 0.F were removed, the affected overmap:: functions and static storage was moved from savegame.cpp to overmap.cpp.

#### Testing

Saved a game using 199ac9b7c590e7110c4026997bb145918cdab098, loaded with the new patch, obsolete entries turned into `nether-chewed field` as specified in json, but no mass- drive-through creation happened.

#### Additional context
Before:
image
After:
image


doc/JSON_INFO.md
```diff
@@ -5197,8 +5197,17 @@ Obsoletion is used, when we want to remove the item entirely from the game, with
 
 For items, monsters, furniture, terrain, factions, loot groups and lot of similar stuff, you remove all places, where the entity can spawn (maps, palettes, NPCs etc), mark the item with "OBSOLETE" flag (optional), and move into `data/json/obsoletion/` or inside  - they will stay here till the next developement cycle, to make fluent transfer between one stable and another
 
-For maps, you also remove it from all the places it can occur, add the map into `data/json/obsoletion/`, and also add the overmap terriains into `data/json/obsoletion/obsolete_overmap_terrains.json`
-list.
+For maps, you need remove it from all the places it can spawn, remove the mapgen entries, and add the overmap terrain id into `data/json/obsoletion/obsolete_oter_ids.json`, to migrate oter_id `hive` and `hive2` into `omt_obsolete` add an entry similar to this, note that if mapgen has already generated this area this will only alter the tile shown on the overmap:
```
Brambor, line 5200:
```suggestion
For maps, you remove the item from all the places it can spawn, remove the mapgen entries, and add the overmap terrain id into `data/json/obsoletion/obsolete_oter_ids.json`, to migrate oter_id `hive` and `hive2` into `omt_obsolete` add an entry similar to this, note that if mapgen has already generated this area this will only alter the tile shown on the overmap:
```

`need` should have been `need to`. But since the previous paragraph is in the form `You do` and not `You need to do`, removing `need` altogether seems better to me.

src/savegame.cpp
```diff
@@ -48,6 +48,7 @@ static const oter_str_id oter_lake_bed( "lake_bed" );
 static const oter_str_id oter_lake_shore( "lake_shore" );
 static const oter_str_id oter_lake_surface( "lake_surface" );
 static const oter_str_id oter_lake_water_cube( "lake_water_cube" );
+static const oter_str_id oter_omt_obsolete( "omt_obsolete" );
 
 static const oter_type_str_id oter_type_bridge( "bridge" );
```
Brambor, line 53:
```suggestion
```

src/savegame.cpp
```diff
@@ -48,6 +48,7 @@ static const oter_str_id oter_lake_bed( "lake_bed" );
...
 static const oter_type_str_id oter_type_bridge_road( "bridge_road" );
```
Brambor, line 54:
```suggestion
```

Brambor (changes requested):

PR #64567, opened 2023-03-25
move warning suppression, fix clang-tidy

kevingranade:
#### Summary
None

#### Purpose of change
Suppress the warning

### DIFF
--- a/src/ui.cpp
+++ b/src/ui.cpp
@@ -1284,6 +1284,8 @@ void pointmenu_cb::select( uilist *const menu )
     impl->select( menu );
 }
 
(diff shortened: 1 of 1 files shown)


kevingranade:
That didn't do it,  maybe wrap the places where it's invoked?

anothersimulacrum:
This was introduced in clang-tidy 14 https://stackoverflow.com/questions/55165840/clang-tidy-nolint-for-multiple-lines

It looks like we're using 12?

anothersimulacrum:
Let's see if https://github.com/CleverRaven/Cataclysm-DDA/pull/64579 works.

kevingranade:
#64579 is the correct fix.

PR #64568, opened 2023-03-25
Small metals weight & recipes audit

Karol1223:
#### Summary
None

#### Purpose of change
Small metals were ALL OVER THE PLACE
- Copper was 8.86g per charge
- Silver was 3g per charge
- Lead was 3g per charge
- Gold was 5g per charge
- Platinum was 5g per charge

And pretty much every recipe with them assumed the charges were of different sizes. Not to mention volumes sucked as well. And prices. And everything, pretty much. Anyway, I wanted to fix it.

#### Describe the solution
**``COPPER``**
- Weight of one charge lowered to 5g
- Volume fixed to match weight and density
- Audited prices
- All* recipes edited to give and use the charge equivalents in the new weight
- Removed the recipe to craft neoprene sheets. No. Too much chemistry. And the PR adding that admitted to gameism (7 years ago)
- Changed the 5l copper ingot disassembly to give 95% of its mass as highlighted by @cake-pie 
- With copper this is pretty much it, it already was audited a while ago and I fixed the most notable problems in some of my previous PRs

**``BRONZE``**
- Weight increased to 300g per instance
- Volume fixed to match weight and density
- Audited prices
- 10l ingots can now be disassembled into 5l ingots, 5l ingots can now be disassembled into raw bronze
- No longer is ``AMMO`` as it has no reason to be so - charges have been removed
- Renamed to 'chunk of bronze'
- Recipe altered to take appropriate amounts of aluminum powder and aluminum foil as well as to not take aluminum ingots anymore (they were wayyyyyy too overkill in weight)

**``LEAD``**
- Weight of one charge increased to 5g
- Volume fixed to match weight and density
- Audited prices
- All* recipes edited to give and use the charge equivalents in the new weight
- Added 1l lead ingots that the 5l ingots disassemble into first before giving raw lead
- Car batteries manually edited to give 60% of their weight in lead when disassembled as that appears to be a common % shared across lead-and-acid batteries. For some this was a buff, for some this was a significant nerf. Also changed crafting recipes to reflect this.

**``SILVER``**
- Weight of one charge increased to 5g
- Volume fixed to match weight and density
- Audited prices
- All* recipes edited to give and use the charge equivalents in the new weight
- Added 1l silver ingots that the 5l ingots disassemble into first before giving raw silver
- 10l silver ingots can now be disassembled into 5l ingots
- Silver jewelry with gemstones no longer gives silver - they are way too small
- Silver hairpin disassembly removed - too small
- Silver watch disassembly brought in-line with the other precious metal watch disassembly
- Removed the recipe to craft hand mirrors with polished steel chunks
- Not directly but very related: Migrated Cybercop badge to Marshall badge, as well as the Cybercop trait to Fed trait. They're relics of ages past and we're way past them.

**``GOLD``**
- Volume fixed to match weight and density
- Audited prices
- All* recipes edited to give and use the charge equivalents of the weight present
- Added 1l gold ingots that the 5l ingots disassemble into first before giving raw gold
- 10l gold ingots can now be disassembled into 5l ingots
- Gold jewelry with gemstones no longer gives gold- they are way too small - EXCEPTION here are dental grills which have enough weight to keep giving gold
- Gold hairpin disassembly removed - too small

**``PLATINUM``**
- Volume fixed to match weight and density
- Audited prices
- All* recipes edited to give the charge equivalents of the weight present
- Added 1l platinum ingots that the 5l ingots disassemble into first before giving raw platinum
- 10l platinum ingots can now be disassembled into 5l ingots
- Platinum jewelry with gemstones no longer gives gold- they are way too small - EXCEPTION here are bracelets which have enough weight to keep giving platinum
- Platinum hairpin disassembly removed - too small

#### Additional context
- * All the recipes affected EXCEPT FOR AMMO AND FIREARMS. I do not know enough about them and there's a lot.


PatrikLundell:
While I agree silver (and probably other material, such as the gold and platinum ones not processed yet <at the time of the writing of this post>) hairpins and jewelry typically are too small to be disassembled into 5 g lumps, there still ought to be a way to use them, possibly by disassembly into <metal> scrap measured in much smaller units (milliliters?). Said scrap would then be possible to melt into larger units.

Karol1223:
> While I agree silver (and probably other material, such as the gold and platinum ones not processed yet (at the time of the writing of this post)) hairpins and jewelry typically are too small to be disassembled into 5 g lumps, there still ought to be a way to use them, possibly by disassembly into (metal) scrap measured in much smaller units (milliliters?). Said scrap would then be possible to melt into larger units.

I am not a fan of adding smaller units that'd both have to be measured in milligrams and have to rely on proper volume solely in the ammo/charges system (we can't define volume smaller than 1ml otherwise).
I did consider the possibility of making a recipe to melt down multiple hairpins into usable metal but that'd only apply to hairpin and not gemstone jewelry.

PatrikLundell:
The problem with melting multiple X into usable metal is that the crafting system isn't capable of combining items in recipes (no "mixed fruit" from some cherries, some apples, and some raspberries, for instance, to provide you with a requirement for 5 units of fruit in a recipe), so you can't take some rings of various types and some hair pins to make a unit of metal, but require a large number of exactly the same item for a recipe to work (I guess you could work around it by having multiple lines of "5 hair pins or fancy ring X or fancy ring Y or 3 small item of some other kind or ..." until you reach the amount required to actually provide a unit of metal).

pjf:
I always felt like small item disassembly should have stochastic returns. So if you disassemble a hairpin, you have (for example) a 1/30 chance of recovering a unit of metal. That avoids needing to track tiny specks of metal, but still allows "fair" salvaging of small items.

(But this PR already does so much for us! I'm not asking for stochastic salvaging to be added to it.)

Karol1223:
> I always felt like small item disassembly should have stochastic returns. So if you disassemble a hairpin, you have (for example) a 1/30 chance of recovering a unit of metal. That avoids needing to track tiny specks of metal, but still allows "fair" salvaging of small items.
> 
> (But this PR already does so much for us! I'm not asking for stochastic salvaging to be added to it.)

This is kind of a bandaid fix - this still means that every hairpin has a 1/30 chance to generate mass out of nowhere and I've been avoiding this like fire in all of my PRs. The answer here is probably a better smelting system, to be honest, but that's way beyond my capabilities.

Zireael07:
Uh, I think the hairpin problem would need a separate PR to fix (or hairpins could just be dropped from the game entirely :P ), this PR is a very large step in a GOOD direction

PR #64572, opened 2023-03-25
reduction of coal spawns in mines and coal/charcoal price

anoobindisguise:
#### Summary
Balance "less insane coal"

#### Purpose of change
Mines were spawning too much coal, in the order of hundreds of liters. This might be fitting for a coal mine but afaik NE isn't known for its present-day coal mines. coal and charcoal were also worth a bit too much.

#### Describe the solution
I reduced coal spawns quite a bit (factor of 10). Also lowered prices.


Fris0uman:
If there's no coal mine in current day NE maybe the coal spawn should just be 0? Having some sources to back up this change would be nice.

Zireael07:
No present-day coal mines doesn't mean no coal in the region - mines are being closed for a boatload of reasons.

Historically it seems there were some https://newenglandtowns.org/new-england-coal#:~:text=New%20England%20Coal%20and%20Anthracite%20Mines&text=New%20England%20has%20never%20been,considerable%20wealth%20in%20local%20areas. and there is a coal-fired power plant still active

And Pennsylvania next door *is* a coal mining hub

anothersimulacrum:
New York exists between New England and Pennsylvania.

Fris0uman:
link says ` America's early days small mines and quarries often successfully supplied local needs for many years`, our mines maps are not small mines abandonned since the 19th century so they should have no coal becasue they are not coal mines at all. And maybe if someone wants to add maps for old coal mines then there could be coal there. Unless modern mines happen to also dig out coal sometimes alongside the thign they are actually looking for

Zireael07:
@anothersimulacrum D'uh, geography fail - but still IIRC the game has elements of other NE states, not just New England proper

anoobindisguise:
Removing coal spawns entirely is probably out of scope as it will also require editing solid rock loot drops, I just wanted to make the amount of coal loot back to what it was instead of substantially more.

PR #64574, opened 2023-03-25
No more Take Your Dog to Work Day

FarkasKBence:
#### Summary

None

#### Purpose of change

Removes dogs (and children) from workplace environments.

#### Describe the solution

Replaced `GROUP_ZOMBIE` with a new monstergroup (`GROUP_CIVILIAN`) and `GROUP_SMALL_STATION`. I've also replaced `place_monsters` with `place_monster`. Showing the greatest of restraints, I did not make any other changes to mapgen.

Affected locations: office tower 1 & 2, office building, cubical office 1 & 2, town hall, light industry, pump station, sewage treatment plant, warehouse.

I don't know much about feral lore-wise, do they make sense to have in workplaces? If so, I can make a monstergroup without the feral-exclusion.

#### Testing

Debug-spawned affected locations in the world.

#### Additional context

None.

### DIFF
--- a/data/json/mapgen/pump_station.json
+++ b/data/json/mapgen/pump_station.json
@@ -105,7 +105,7 @@
         "o": { "item": "SUS_office_filing_cabinet", "chance": 66 }
       },
       "place_monster": [
-        { "group": "GROUP_SMALL_STATION", "x": [ 5, 21 ], "y": [ 8, 23 ], "chance": 50, "repeat": [ 1, 2 ] },
+        { "group": "GROUP_SMALL_STATION", "x": [ 5, 21 ], "y": [ 8, 23 ], "chance": 50 },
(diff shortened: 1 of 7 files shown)


anothersimulacrum:
People became "feral"s pretty much at random, so they could show up anywhere, but they'd be most likely to be rioters.

Termineitor244:
Oh shoot! Thanks for updating the light industries, I added the Civilian spawn group with the ferrous spawn update but forgot to add it when I was making at the same time the PR for updating the light industries...

And you should probably make another spawn group including the basic ferals, for these locations, as I said, this group was added with the ferrous update, for the FERROUS spawn group to naturally spawn non ferrous zombies that would end up "evolving" in spawn times into ferrous zombies... That spawn group is called only inside other spawn groups so the ferals were acounted for in those...

In short, I had no use in that PR for a similar group including ferals, but you can definitely make use of one in these locations since these would be the main spawn group for them! Ferals in office settings should definitely be a thing, if only we had ferals charging at you with a stapler, a heavy paperweight or some office scissors...

PR #62664, opened 2022-12-10
Item names now include the number of additional pockets attached to them

Ilysen:
#### Summary
Interface "Item names now include the number of pockets attached to them; for instance, a ballistic vest with three extra pouches might display as 'US ballistic vest+3'"

#### Purpose of change

MOLLE pockets are currently buried pretty deep in the interface; to see if an item has any attached pockets at all, you have to go into the item description and then scroll down to the storage section. There's no way to easily see at a glance how many pockets an item has attached.

#### Describe the solution

I appropriated the code used to display the number of mods on an item to also display the number of attached pockets. For instance, a US ballistic vest with four extra accessories would show up as `US ballistic vest+4`, in the same way that an M4 carbine with two mods would show up as `M4 carbine+2`. See below for screenies.

For items that can have both mods and attached pockets (I believe this is only applicable through modding right now), it will display the sum of mods and pockets. A modded gun with two mods and two pockets would show up as `modded gun+4`, and not `modded gun+2+2`.

#### Describe alternatives you've considered

I originally wanted this PR to be a general improvement to MOLLE pockets in that I tried to make them be filthy and damaged if the item they spawned attached to was also filthy and damaged. I couldn't actually figure out how to do that, though, so I scrapped it. Yay, rustiness!

I'd considered also making the number of attached pockets display as a unique thing on its own (such as `US ballistic vest+2 pockets` or `US ballistic vest (+2 pockets)` or something, but all the things I thought of seemed pretty jank to me. Including it with the number of existing mods seemed like the most natural display to me.

#### Testing

I spawned in some zombie soldiers and killed 'em using debug, then searched the bodies. Items with attached pockets showed up using the new names, while items without any did not. Attaching a new pocket to an item caused its name to update to include that pocket as `+1`, and taking all pockets off removed the suffix from an item entirely.

I made sure that items with just mods (and no attached pockets) appeared correctly; an M4 carbine with just a shoulder strap appeared as `M4 carbine+1` correctly and lost the suffix when I took the strap off.

#### Additional context

Some drops from a zombie soldier. Note the +4 on the load bearing vest.
cataclysm-tiles_Y8qBILmW3J

Looking further in the item card for the vest, we can see the pockets that were attached:
cataclysm-tiles_zzr13daQra

Bonus picture: my testing helper named Fred.
image

PR #64369, opened 2023-03-18
Item names now include the number of additional pockets attached to them

Dan1ss1mo:
#### Summary
Interface "Item names now include the number of pockets attached to them"

#### Purpose of change
MOLLE pockets are currently buried pretty deep in the interface; to see if an item has any attached pockets at all, you have to go into the item description and then scroll down to the storage section. There's no way to easily see at a glance how many pockets an item has attached

#### Testing
Works as intended

#### Additional context
#62664

PR #64582, opened 2023-03-26
Revive 62664

bombasticSlacks:
#### Summary
Interface "Item names now include the number of pockets attached to them"

#### Purpose of change
MOLLE pockets are currently buried pretty deep in the interface; to see if an item has any attached pockets at all, you have to go into the item description and then scroll down to the storage section. There's no way to easily see at a glance how many pockets an item has attached.

#### Describe the solution
Revived from #62664 I only added an additional unit test

from original PR "I appropriated the code used to display the number of mods on an item to also display the number of attached pockets. For instance, a US ballistic vest with four extra accessories would show up as US ballistic vest+4, in the same way that an M4 carbine with two mods would show up as M4 carbine+2. See below for screenies.

For items that can have both mods and attached pockets (I believe this is only applicable through modding right now), it will display the sum of mods and pockets. A modded gun with two mods and two pockets would show up as modded gun+4, and not modded gun+2+2."

#### Testing
Unit test added

#### Additional context
Some drops from a zombie soldier. Note the +4 on the load bearing vest.

image

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -6350,7 +6350,8 @@ std::string item::tname( unsigned int quantity, bool with_prefix, unsigned int t
 
     if( is_corpse() || typeId() == itype_blood || item_vars.find( "name" ) != item_vars.end() ) {
         maintext = type_name( quantity );
-    } else if( ( is_gun() || is_tool() || is_magazine() ) && !is_power_armor() ) {
+    } else if( ( ( is_gun() || is_tool() || is_magazine() ) && !is_power_armor() ) ||
+               contents.has_additional_pockets() ) {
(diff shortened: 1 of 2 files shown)


GuardianDll:
Cough cough #64369

anothersimulacrum:
https://github.com/CleverRaven/Cataclysm-DDA/pull/64369#issuecomment-1481917647

bombasticSlacks:
> Also please, make a separate item to test, it would we bad, if someone make a variant of a vest, and the test will fail because of this

it already uses a test vest, the same test vest used for the rest of the MOLLE attach tests

PR #64595, opened 2023-03-26
Rework anvil craft and add two new simpler anvils

Aerin-of-the-Toast:
#### Summary
Content "Rework anvil craft and add two new simpler anvils"

#### Purpose of change

The recipe for the anvil was taking a lot of liberties. It's also not strictly necessary to have a full, proper, modern anvil for most blacksmithing tasks the survivor is doing where simpler alternatives would generally work fine.

#### Describe the solution

Let's start with the anvil recipe. It was split into two; a recipe using modern tools and pre-made steel frames, and an innawoods-friendly one forging from chunks of steel. The "modern" craft is based off the method in this video: https://www.youtube.com/watch?v=fcE7kDpmzYM It's essentially just a lot of angle grinding and drilling metal into shape and welding all the bits together.

For the forged variant, I had to do a bit of educated guesswork since I cannot find a source on how to *smith* a full anvil from scratch; every site I saw was aimed at modern blacksmiths using angle grinders. While I can't guarantee the recipe is accurate, it's _more_ accurate than the previous version; it now actually costs a more reasonable amount of charcoal to forge (the old anvil took less charcoal to forge than small swords!), takes way longer (it's a BIG piece of metal; that's a lot of forging!) and needs more tools and proficiencies since it is a very complex craft. With no proficiencies at all, the recipe takes 6 days to complete; this might still be a low-ball, I'm not sure, but it's a lot more reasonable than making a 54kg anvil in 4 hours. The crucible requirement was also removed because you're not casting this thing. Notably, this anvil requires proper tools (and at least anvil 3) to make, since it's such a complex craft, so at minimum you'd need to have made one of the other anvils first to make this one.

To compensate, two new anvils were added; again they were essentially split into one made with modern power tools and one that can be bootstrapped innawoods if you desire. The first one is an anvil cut out of a railroad track; doing a bit of digging shows that this is actually a very common beginner anvil for blacksmiths, which is convenient because it meant there was a lot of sources on making one. Here's a source video I used, though I skipped the aesthetic parts: https://www.youtube.com/watch?v=UB1KBBWWeYg To put it simply, you use an angle grinder to cut it into the shape of an anvil, drill out a hardy hole and some holes in the bottom as attachment points, and bam. Anvil.

As an additional part of the railroad track anvil, I also added the ability to get lengths of railroad track directly by "cutting metal" on a railroad track with a hacksaw or angle grinder. While I was at it, I also reworked the bash drops for railroad tracks. There's like, 120 kilograms of medium carbon steel there per meter, just ripe for the taking! Alongside this I also added an uncraft recipe for the "length of railroad track" to disassemble it into 30 lumps of medium carbon steel, along the same lines as disassembling a heavy-duty frame for lumps of steel.

The last are a set of stake anvils; these are anvils that were used throughout much of history since they're suitable for most blacksmithing tasks and way easier to make and more portable than modern-style massive ACME anvils. Stake anvils were even used to make armor (https://www.anvilfire.com/anvils/donated_019.php) which is very complex for blacksmithing, so they're good enough for anything the survivor is doing. The ones I added were mostly based on the "viking anvil" How To Make Everything made. https://www.youtube.com/watch?v=fO8tvaaZUro The craft difficulty and time spent are a bit high for the stake anvils, but that's because the craft is assuming sub-optimal tools (hammering 2 and a bronze anvil).

Both new anvils need to be staked on a log in the construction menu before they can be used, they can't be used as raw items. For now, the new anvils work just the same as the old anvil, so the old one is only useful if you find it somewhere; as far as I'm aware there's nothing that *needs* a large, modern anvil, it's just more convenient. Maybe when we get the option to have better tools affect crafting speed/craft difficulty/exertion, modern anvils can be easier and faster to use than stake or track anvils.

There are A LOT of items that can work as an anvil; all you really need is thick, flat steel to hammer off of. Here's the video where I got the idea for the track anvil from that includes other examples: https://www.youtube.com/watch?v=ltBqRKKb0OQ Some notable alternatives include a truck axle or even a sledgehammer head if you could secure it.

I also considered options like locking tempered crafts behind a "proper" anvil or other ideas to keep the "proper" anvil relevant, but they felt too gamey and I couldn't find justification that a "proper" anvil is ever _strictly_ necessary.

#### Testing

Hopped in game, made sure the recipes and items spawned in right. Made sure the railroad tracks cut into lengths of track and bashed into lumps/chunks of medium steel properly.

#### Additional context

The two biggest effects this PR has on game balance are: the crucible is no longer needed at all for a blacksmithing setup (unless you're bootstrapping from nothing, in which case it's still needed for bronze casting first). And now there's a very easy source of basically limitless medium carbon steel for all your forging needs. But, like, railroad tracks _are_ made from medium carbon steel, and there's a LOT of steel.


kevingranade:
> far as I'm aware there's nothing that needs a large, modern anvil,

My understanding is this is not true,  you need a large anvil for large pieces and heavy operations such as forge welding. If you read the article about the stake anvil more closely,  you'll see it doesn't say it's used for general armor crafting,  it's used for shaping sheet metal.

https://www.anvilfire.com/21centbs/Selecting-an-Anvil.php

For our purposes,  if a recipe has ingots or scrap as an input,  it needs to be a heavy anvil.

Aerin-of-the-Toast:
> > far as I'm aware there's nothing that needs a large, modern anvil,
> 
> My understanding is this is not true, you need a large anvil for large pieces and heavy operations such as forge welding. If you read the article about the stake anvil more closely, you'll see it doesn't say it's used for general armor crafting, it's used for shaping sheet metal.
> 
> https://www.anvilfire.com/21centbs/Selecting-an-Anvil.php
> 
> For our purposes, if a recipe has ingots or scrap as an input, it needs to be a heavy anvil.

Ah, thanks for the source! Since like 95% of our blacksmithing recipes seem to involve forge welding I guess the stake anvil is insufficient. The railroad track anvil should still be fine though, right? Since it qualifies as a bench anvil or is on the lightest side of a farrier's anvil.

If I remove the stake anvils should I then make sure the regular forged anvil recipe is doable without needing an anvil first, even if it takes longer?

Brambor:
Regarding anvil "craft-ability":
I would say that the tools define the time needed.
If there are two different processes=recipes with different time and tool requirements, make the two recipes.
Maybe there is a reason to craft additional anvils even if you have an anvil. Like basecamp blacksmiths (if that is a thing) etc.

tenmillimaster:
> For our purposes, if a recipe has ingots or scrap as an input, it needs to be a heavy anvil.

Wouldn't it make sense then to have Anvil be a tool quality? 
Something like...
peening rivets - Anvil 1
work hardening bronze, shaping hot metal - anvil 2
forge welding large items - Anvil 3

Aerin-of-the-Toast:
> > For our purposes, if a recipe has ingots or scrap as an input, it needs to be a heavy anvil.
> 
> Wouldn't it make sense then to have Anvil be a tool quality? 
> Something like...
> peening rivets - Anvil 1
> work hardening bronze, shaping hot metal - anvil 2
> forge welding large items - Anvil 3

It already is. Anvil 1 is a rock for bushcraft stuff and copper, anvil 2 is a bronze anvil for bronzeworking and anvil 3 is a proper anvil for steelworking.

I suppose there's an argument some simpler steel crafts (like maybe nails) can be anvil 2 instead of 3 but that's out of scope for this pr.

kevingranade:
> The railroad track anvil should still be fine though, right?

It depends, a work piece over a certain size/heft,  say a big melee weapon,  would be out of the reach of a workbench anvil. 

> If I remove the stake anvils should I then make sure the regular forged anvil recipe is doable without needing an anvil first, even if it takes longer?

I need to look at what a forged anvil even entails,  I suspect a bigger anvil...

In general though, no it just works the way it works,  maybe firing an anvil is a pointless exercise. 

Yea quality is definitely the mechanism for this, increasing size would increase one quality, limiting larger firing operations to larger anvils,  while smaller anvils would have other qualities like fine hammering or silversmithing. Idk if we have a yse fir a jewelers anvil, probably not.

Zireael07:
> I would say that the tools define the time needed.

A topic frequently raised which IIRC has a dedicated issue somewhere in the tracker...

Brambor:
> > I would say that the tools define the time needed.
> 
> A topic frequently raised which IIRC has a dedicated issue somewhere in the tracker...

In this case, I meant completely different recipes that need different tools and happen to have different times, which happens to be better tool => better recipe => better time.

Drew4484:
Condensed thoughts here, longer arguments at the end.
Steel better than iron, forged better than cast. You do not always need a 100kg anvil. Low quality Anvil-Like Objects are valid.

Using the criteria above, I recommend ranking the existing anvils as follows:

| Name | Weight | Material | Current Quality | Suggested | Notes 
| :--- | :--- | :--- | :--- | :--- | :--- |
| Large Rock | 15.77 kg | Stone | 1 | 1 | Too light, too soft, and unstable. Barely qualifies.
| Boulder Anvil | Undefined | Stone | 1 | 1 | Same as above. Presumably heavier and more stable, but not dramatically so.
| Hard Plating | 34.64 kg | Tempered Carbon Steel | 1 | 2 | Heavy, solid, good steel. Would make a good countertop/bench anvil.
| Heavy-Duty Frame | 100 kg | Tempered CS/Steel | 1 | 1 | As above, but larger mass offset by awkward work surface if left intact.
| Counterweight | 750 kg | Iron | 1 | 2 | This appears to be a single bulk counterweight, these are typically squared off and would make a decent anvil surface. Raise to 2 due to sheer mass and at least one smooth flat surface as a face. (Should be changed to steel or lead though lead would be a worthless anvil). 
| Bronze Anvil | 2.18 kg | Bronze | 2 | 3 | Ignoring that the volume and weight suggest this is an anvil for ants (which should be corrected ASAP), bronze anvils work fine. Long term, bronze will not handle the temperatures and impacts associated with working steel, but a properly sized anvil made from bronze should work well enough for any survivor decides to go through the work to make one.
| Anvil | 54 kg | Steel | 3 | 3 | This currently represents a professional blacksmith's anvil as opposed to one found in a garage workshop or a barn.  A more 'consumer' anvil would range from 1/2 to 2/3rds this weight. I recommend keeping this quality but lowering the weight.

If we adjust the existing anvil, I propose a new Heavy Anvil ( ~225kg, Tempered Carbon Steel, Quality 4) to salvage from anywhere that would do serious metal fabrication. The large size allows for very heavy tools and typically the largest possible work area.

- Scrapyards: Sometimes people just throw anvils away. Maybe a grandparent died, maybe the shop finally closed. If no interested party buys it, a surprising number of anvils just get tossed. A 200kg anvil would make a great curb piece for the office.
- Military base: Our compact in-game base lacks heavy metalworking equipment, but this kind of thing wouldn't be out of place at a larger base, especially one with a significant vehicle detachment or a base established during/before WWII. While they wouldn't be making new parts or anything like that, makeshift repairs to get a unit moving again after recovery. 
- Exodii: They presumably do some amount of metal fabrication, and don't carry an entire foundry with them. While not all of their work needs to be done by hand, their salvage-centric existence likely means they could put a large anvil to work.
- Artisans: Cody is one of the few living people in the region who works heavy enough metal to need a heavy anvil. 
- Light Industry: Maybe, but low odds. These suffer from Mansion-itis already, becoming a low effort jackpot for any survivor with a truck.

- Materials and cast/forged: Steel is harder, which is desirable as you don't want the anvil absorbing the energy you're trying to put into the object you're making. It also makes the smith's life easier as the hammer bounces back up higher, saving a non-trivial amount of energy on long crafts. There are examples of iron anvils with steel benches, but if you can make that you can just make a steel anvil. A properly used anvil will last a very long time.
- Function of weight and size: A bigger anvil has many advantages and few disadvantages in use. Practically you only need a properly mounted anvil with a large enough face and a weight at least 50:1 the weight of your handheld striking tool, i.e. a hammer with a 1 kg head would call for at least a 50 kg anvil. However giant metal anvils are difficult to salvage, transport, and near impossible for a survivor to make. The anvil's face dimensions do not scale linearly with weight but generally a larger anvil will offer a larger face.
- Anvil as a Quality: An anvil's primary reason to exist is to be a solid surface to bang on. The Quality should represent the suitability of the face (or flat-ish part of the rock you're hammering against), how stable the anvil is (weight and mount impact this unless the anvil is extremely heavy anyway) and durable it is. All additional features like horns, steps, and Pritchel/Hardie holes are present on most manufactured anvils, but a heavy duty-frame will work just fine as a hard solid surface to hit hot metal on. Our higher quality crafts that can gain advantage from those additional features should be limited to the higher Quality anvils. 

(wrote this late at night, ping me on Discord with questions or disregard entirely as appropriate)

data/json/items/resources/metal.json
```diff
@@ -956,6 +956,23 @@
...
+    "weight": "30000 g",
+    "//0": "Modern railroad tracks are typically around 60kg per meter.",
+    "volume": "5 L",
+    "longest_side": "50 cm",
+    "price": 5000,
+    "price_postapoc": 50,
+    "material": [ "mc_steel" ],
+    "symbol": "=",
+    "color": "light_gray",
+    "qualities": [ [ "ANVIL", 2 ] ]
```
Brambor, line 974:
```suggestion
    "flags": [ "DURABLE_MELEE" ]
```

data/json/items/tool/metalworking.json
```diff
@@ -16,6 +16,38 @@
...
+    "weight": "20000 g",
+    "volume": "3 L",
+    "price": 10000,
+    "price_postapoc": 1000,
+    "to_hit": -5,
+    "bashing": 20,
+    "material": [ "mc_steel" ],
+    "symbol": ";",
+    "color": "dark_gray",
+    "flags": [ "DURABLE_MELEE" ]
```
Brambor, line 33:
```suggestion
    "flags": [ "DURABLE_MELEE" ],
    "qualities": [ [ "ANVIL", 3 ] ]
```

data/json/items/tool/metalworking.json
```diff
@@ -16,6 +16,38 @@
...
+    "weight": "4000 g",
+    "volume": "700 ml",
+    "price": 1400,
+    "price_postapoc": 1000,
+    "to_hit": { "grip": "solid", "length": "short", "surface": "any", "balance": "uneven" },
+    "bashing": 20,
+    "material": [ "steel" ],
+    "symbol": ";",
+    "color": "dark_gray",
+    "flags": [ "DURABLE_MELEE" ]
```
Brambor, line 49:
```suggestion
    "flags": [ "DURABLE_MELEE" ],
    "qualities": [ [ "ANVIL", 3 ] ]
```

Brambor (changes requested):
I'm not sure exactly about the proposed changes, but you seem to have swapped `railroad_track_item` and `anvil_track`.
And I'm not sure about the numbers either.
Just please check.
Maybe you meant to do it later...

Aerin-of-the-Toast:
> I'm not sure exactly about the proposed changes, but you seem to have swapped `railroad_track_item` and `anvil_track`.
> And I'm not sure about the numbers either.
> Just please check.
> Maybe you meant to do it later...

No, the anvil *items* don't have anvil quality because I wanted them to only be usable when placed as furniture. The main reason I didn't give them the same anvil quality as the raw track on its own is cuz I didn't want to confuse players who thought they were only supposed to have 2.

Aerin-of-the-Toast:
Oh I should've done this sooner oops
Paging @kevingranade , does Drew's proposal sound good for me to go ahead with?

Brambor:
> Oh I should've done this sooner oops Paging @kevingranade , does Drew's proposal sound good for me to go ahead with?

Nobody raised any issue and everyone is thumbing up first post. No need to ask kevin to make your PR from Draft to Open.

OmegaCenti:
I feel this needs to become a thing. I agree with so many points on this, and after watching the youtube videos, you definitely have my vote! unstale this post post-haste!

Miner239:
> All additional features like horns, steps, and Pritchel/Hardie holes are present on most manufactured anvils, but a heavy duty-frame will work just fine as a hard solid surface to hit hot metal on. Our higher quality crafts that can gain advantage from those additional features should be limited to the higher Quality anvils.

I think these additional features should be separated as their own qualities instead of merging it into higher anvil quality, so that they can later be assigned to other non-anvil items that also provide that quality.

kevingranade:
Yes i agree with everything drew said, do that.

data/json/items/fake.json
```diff
@@ -71,6 +71,20 @@
     "name": { "str": "boulder anvil" },
     "qualities": [ [ "ANVIL", 1 ] ]
   },
+  {
+    "id": "fake_anvil_track",
+    "type": "TOOL",
+    "copy-from": "fake_item",
+    "name": { "str": "stake" },
+    "qualities": [ [ "ANVIL", 3 ] ]
```
Drew4484, line 79:
```suggestion
    "qualities": [ [ "ANVIL", 2 ] ]
```

data/json/items/tool/metalworking.json
Drew4484: > 
```suggestion
    "flags": [ "DURABLE_MELEE" ],
    "qualities": [ [ "ANVIL", 2 ] ]
```

Aerin-of-the-Toast:
Obsoleted by #68386

PR #64597, opened 2023-03-26
refactor ui.h & fix clamp to min instead of 0

Brambor:
#### Summary
None

#### Purpose of change
Following #64463
Simplify the code: reuse the general case in specialized cases.
Fix one of the general cases. (But the fix will affect nothing right now.)
Streamline documentation.

#### Describe the solution
An entire day of debugging templates...
Simply implement most general functions and in the specialised functions call the general one.

Do something useful or impactful with my time. Like fixing filtering `m:no` in the crafting menu.

#### Testing
~I'm not sure whether the tests will pass. Clang-tidy 12. We shall see.~
Compiled on my machine, ran the game, and scrolled some menus, looked fine.
I couldn't test the bugfix, as it is not used anywhere.

### DIFF
--- a/src/ui.h
+++ b/src/ui.h
@@ -537,11 +537,11 @@ void kill_advanced_inv();
 
 /**
  * Helper for typical UI list navigation with wrap-around
- * Add delta to val, then wrap to the range [0,size)
+ * Add delta to val. If on bounds, wrap to other bound, otherwise clamp to the range [0,size)
  */
 template<typename V, typename S>
 inline typename std::enable_if < !std::is_enum<V>::value, V >::type
-increment_and_wrap( V val, int delta, S size )
+inc_clamp_wrap( V val, int delta, S size )
 {
     if( size == 0 ) {
         return 0;
@@ -565,9 +565,9 @@ increment_and_wrap( V val, int delta, S size )
  */
 template<typename V, typename S>
 inline typename std::enable_if < !std::is_enum<V>::value, V >::type
-increment_and_wrap( V val, bool inc, S size )
+inc_clamp_wrap( V val, bool inc, S size )
 {
-    return increment_and_wrap( val, static_cast<int>( inc ? 1 : -1 ), size );
+    return inc_clamp_wrap( val, static_cast<int>( inc ? 1 : -1 ), size );
 }
 
 /**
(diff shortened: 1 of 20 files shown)


Brambor:
I don't understand templates... Or whatever is causing the linker errors,

Brambor:
I believe it will work now.
Finally ready to be merged, I was going a bit crazy :')

src/ui.h
```diff
@@ -536,22 +536,38 @@ class pointmenu_cb : public uilist_callback
...
+        return size - 1;
+    }
+    // wrap to first
+    // Templating of existing `unsigned int` triggers linter rules against `unsigned long`
+    // NOLINTNEXTLINE(cata-no-long)
+    if( val >= static_cast<V>( size ) - 1 && delta > 0 ) {
+        return 0;
     }
-    return modulo( val + ( inc ? 1 : -1 ), size );
+    return std::clamp<V>( val + delta, 0, size - 1 );
```
kevingranade, line 559: So this will clamp to 0, size, then only wrap on the next increase? That's pretty counterintuitive.

kevingranade: If that is the intended behavior,  it needs a brief comment clarifying it,  and/or the function needs a new name. clamp_then_wrap() gets the idea across but it's awkward as hell.

Brambor:
Rebased and addresed kevingrade's review.

src/ui.h
Brambor: @kevingranade is it good enough now?

Brambor: I also thought about splitting them:
`inc_clamp_wrap( V val, int delta, S size )` needs the clamp in name, as it clamps
`inc_clamp_wrap( V val, bool inc, S size )` and `inc_clamp_wrap( T val, I inc, T size )` don't need clamp in name, as they only increase or decrease by one, so they can never be clamped. So rename them to `inc_wrap`.
Then, there would be three names: `inc_clamp_wrap`, `inc_clamp`, `inc_wrap`

Or maybe it would be better to make them all into one function? with parameters `bool clamp` `bool wrap`?
in there break into four cases
```
!clamp && !wrap {static assert("has to choose at least one `clamp` or `wrap`")}
clamp && !wrap { the inc_clamp code }
!clamp && wrap { the inc_clamp_wrap code - BUT not stopping at bounds }
clamp && wrap { the inc_clamp_wrap code }
```

Brambor: I consider this solved.

PR #64600, opened 2023-03-26
Support creating items from item groups with spells

mqrause:
#### Summary
Features "Support creating items from item groups with spells"

#### Purpose of change

Documentation was saying this is a thing already when it absolutely isn't.
I also want to reduce stray uses of `item::in_its_container` to have an easier time changing its behavior.

#### Describe the solution

- refactor `spell_effect::spawn_ethereal_item` to support `SPAWN_GROUP` flag
- remove `WITH_CONTAINER` flag because the same thing can be done with itemgroups
- UI has no proper way to display item group results, but I don't want to get too sidetracked with this

#### Testing

Changed a spell to use an item group and used it. It spawned the group (damage times), granted items were still ethereal if not comestible/permanent.


GuardianDll:
FYI EoC has such ability now, widly used in xedra mod

mqrause:
So you want to tell me I can just completely remove the "spawn_item" spell effect cause there are better ways to do it?

GuardianDll:
Sure not, having more ways to do a thing is always better

Fris0uman:
> Sure not, having more ways to do a thing is always better

No, it's confusing and brittle. Having one good centralized way to do things is better code wise

GuardianDll:
right now to spawn loot group from spell, you should make a spell, a loot group, and additional EoC that will spawn a loot group
why not get rid out of additional stuff, if we can afford it? its like saying "let's get rid out of damage dealing in spells, because we have guns and effects already, that can do the same"

mqrause:
Just to be clear, there's over 100 spells that spawn items and I definitely won't rewrite all of them to use EOCs to remove the spawn_item spell effect. That'd go far beyond what I wanted to achieve with this, which is to clean up a bit of code for some other stuff I'm working on.

PR #64610, opened 2023-03-26
Make all tile lights destroyable by smashing and shooting

teackot:
#### Summary
Features "Make all lights above floor tiles destroyable by smashing and shooting"

#### Purpose of change

Some tiles' lights can be destroyed without destroying the tile itself, t_thconc_floor_olight's light can also be destroyed by shooting, so it makes sense to implement it for other tiles with lights too.

#### Describe the solution

Add the "shoot" field to all tiles with lights and correct some "bash" fields to only destroy the light.

Remove the ability to destroy lights by shooting.

#### Testing

Edited the map to create all _olight tiles and successfully destroyed their lights by shooting. Made sure the correct tile without lights is set.


Maleclypse (approved):
Thanks for doing this!

PR #64612, opened 2023-03-27
Add all effects to player display speed tab

sparr:
#### Summary
Interface "Player display speed tab effect list is more comprehensive"

#### Purpose of change
The speed tab shows the player's current base move cost and speed, and some of the reasons those are not the default values. Many common reasons are not shown, leaving the player to guess, or spend time investigating, all the possible causes.

#### Describe the solution
This PR modifies the existing functions that calculate movement cost and speed to also emit a list of each modifier applied along with a description. The `draw_speed_tab` code that partially duplicates that behavior is then replaced with calls to those functions.

I discussed a few different ways to produce and pass the data with folks on Discord, but was encouraged to just get the feature working before doing too much optimization in that direction. It is probably advisable to avoid generating/processing the descriptions when they won't be required, when the functions are called in the normal gameplay loop.
I also considered ways to store *more* info about each modifier, for display in the info tab, but that can also happen later.

#### Testing
I have produced in-game situations triggering about half of the hard coded modifiers touched by this code, and a few dynamic modifiers such as martial arts. I haven't thoroughly tested every modifier to see how it looks in the interface.

#### Additional context
image

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -185,6 +185,8 @@ static const character_modifier_id character_modifier_limb_str_mod( "limb_str_mo
 static const character_modifier_id
 character_modifier_melee_stamina_cost_mod( "melee_stamina_cost_mod" );
 static const character_modifier_id
+character_modifier_move_mode_move_cost_mod( "move_mode_move_cost_mod" );
+static const character_modifier_id
 character_modifier_ranged_dispersion_vision_mod( "ranged_dispersion_vision_mod" );
 static const character_modifier_id
 character_modifier_stamina_move_cost_mod( "stamina_move_cost_mod" );
@@ -6442,7 +6444,8 @@ void Character::burn_move_stamina( int moves )
 
     burn_ratio *= move_mode->stamina_mult();
     mod_stamina( -( ( moves * burn_ratio ) / 100.0 ) * get_modifier(
-                     character_modifier_stamina_move_cost_mod ) );
+                     character_modifier_stamina_move_cost_mod ) * get_modifier(
+                     character_modifier_move_mode_move_cost_mod ) );
     add_msg_debug( debugmode::DF_CHARACTER, "Stamina burn: %d", -( ( moves * burn_ratio ) / 100 ) );
     // Chance to suffer pain if overburden and stamina runs out or has trait BADBACK
     // Starts at 1 in 25, goes down by 5 for every 50% more carried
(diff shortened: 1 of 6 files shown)


src/character.cpp
```diff
@@ -9263,75 +9297,90 @@ int Character::run_cost( int base_cost, bool diag ) const
...
+                                 _( "Obstacle Mutations" ) );
 
             if( has_proficiency( proficiency_prof_parkour ) ) {
-                movecost *= .5;
+                run_cost_effect_mul( movecost, effects, 0.5, _( "Parkour" ) );
             }
 
             if( movecost < 100 ) {
+                run_cost_effect effect { _( "Bonuses Capped" ) };
+                effects.push_back( effect );
```
sparr, line 9309: The three run_cost_effect entries above here are actually never used, because draw_speed_tab always calls this function with a base movecost of 100, but I wanted to be consistent with the later cases.

src/character.cpp
```diff
@@ -9263,75 +9297,90 @@ int Character::run_cost( int base_cost, bool diag ) const
...
+                    run_cost_effect_mul( movecost, effects, 0.85f, _( "Heelys (Running)" ) );
                 } else if( is_walking() ) {
-                    movecost *= 0.9f;
+                    run_cost_effect_mul( movecost, effects, 0.9f, _( "Heelys" ) );
                 }
             } else {
-                movecost *= 1.1f;
+                run_cost_effect_mul( movecost, effects, 1.1f, _( "Heelys (Offroad)" ) );
             }
         }
```
sparr, line 9385: The three blocks above could do with some deduplication. I can do that in this PR if requested.

src/character.cpp
```diff
@@ -9263,75 +9297,90 @@ int Character::run_cost( int base_cost, bool diag ) const
...
 
-        movecost *= get_modifier( is_prone() ? character_modifier_crawl_speed_movecost_mod :
-                                  character_modifier_limb_run_cost_mod );
+        if( is_prone() ) {
+            run_cost_effect_mul( movecost, effects, get_modifier( character_modifier_crawl_speed_movecost_mod ),
+                                 _( "Crawling (Modifier)" ) );
+        } else {
+            run_cost_effect_mul( movecost, effects, get_modifier( character_modifier_limb_run_cost_mod ),
+                                 _( "Encumbrance/Wounds" ) );
```
sparr, line 9325: I couldn't find a good way to split this into two separate entries.

Terrorforge:
There's no martial art that actually increases speed, is there? They all reduce the cost of actions, the same way modifiers to base move cost work.

(Incidentally the martial art .json confuses me, because it seems that when martial arts modify "movecost", it means the cost of attacking, but when mutations have a "movecost_modifier" it refers exclusively to modifying the Base Move Cost, i.e. overland movement. I have not tested to make sure this actually the case)

sparr:
@Terrorforge I, too, was confused when I saw [the existing check for martial arts](https://github.com/CleverRaven/Cataclysm-DDA/blob/78fd8cdc03c8014c63eb5b120558ad75c7db62db/src/character.cpp#L10864-L10865) in the speed calculation code. There appears to currently be just one martial art which provides an actual speed bonus, the [Ninjutsu onkill buff](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/json/martialarts.json#L975), and it happens to be the thing I had most recently tested when I took this screenshot.

ashGlaw:
Wonderful change! Speed calculations have vexed me for years. Thanks much

qoreQyaS:
idk if it's actually related to this PR but before opening an issue i post it here:
image

```yaml
- OS: Linux
    - OS Version: Distributor ID: Debian; Description: Debian GNU/Linux 12 (bookworm); Release: 12; Codename: bookworm; 
- Game Version: 3d15dcc [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
    My sidebar [sidebarfoo]
]
```

[Quivero-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11216037/Quivero-trimmed.tar.gz)

sparr:
@qoreQyaS I recommend opening an issue for that, which I'll attempt to investigate. There's some nontrivial overlap between my efforts to collect speed modifiers in `Character::mod_speed_bonus` and the speed-affecting entries in `Creature::effects`

PR #64614, opened 2023-03-27
Build Universal Binary on macOS

BrettDong:
#### Summary
Build "Build Universal Binary on macOS"

#### Purpose of change
Deliver native arm64 binaries for Apple Silicon users.

Second attempt to #63702, supersedes #63825.

#### Describe the solution
Add a `UNIVERSAL_BINARY` flag in `Makefile`, that when set to `1` and is compiling for macOS, add `-arch x86_64 -arch arm64` to `CXXFLAGS` and `LDFLAGS` to let the compiler and the linker to generate a dual architecture binary (Universal Binary).

Also changed the build matrix to build Universal Binary for macOS to test if this is working on GitHub Actions. I'm afraid enabling UBSan is going to spend too much time and blow up cache spaces so I turned off UBSan for macOS builds.

#### Testing
macOS build in general build matrix passes.

### DIFF
--- a/.github/workflows/matrix.yml
+++ b/.github/workflows/matrix.yml
@@ -89,6 +89,7 @@ jobs:
             localize: 1
             test-stage: 1
             native: linux64
+            pch: 1
             archive-success: basic-build
             dont_skip_data_only_changes: 1
             title: Basic Build and Test (Clang 10, Ubuntu, Curses)
@@ -103,6 +104,7 @@ jobs:
             release: 1
             cmake: 0
             native: osx
+            pch: 0
             tiles: 1
             sound: 1
             localize: 1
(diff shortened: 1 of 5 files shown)


CzBiX:
Build Universal Binary will required us add `LSArchitecturePriority` in Info.plist, otherwise it will still run in Intel mode.
see https://developer.apple.com/documentation/apple-silicon/building-a-universal-macos-binary#Specify-the-Launch-Behavior-of-Your-App.

> If an app doesn’t contain an executable binary, the system may run it under Rosetta translation as a precautionary measure to prevent potential runtime issues. For example, the system runs script-only apps under Rosetta translation. If you verified that your app runs correctly on both Apple silicon and Intel-based Mac computers, add the [LSArchitecturePriority](https://developer.apple.com/documentation/bundleresources/information_property_list/lsarchitecturepriority) key to your app’s Info.plist file and list the arm64 architecture first.

BrettDong:
Yes, I was just going to make a pull request to update the Info.plist metadata.

PR #64615, opened 2023-03-27
Avoid spilling items into the original item.

### LINKED ISSUE #64604: Crash when using a hotplate/similar tool while it's inside an open container with another object
### Describe the bug

This one's one of those weirdly specific bugs - If the player uses a tool to heat up food while said tool is in an open container (a pot, a can, anything that would spill) whilst there is at least _one other object in the container with it_ the game will crash once the food item is heated. If it's just the tool in the container alone it doesn't cause a crash.

### Attach save file

[Mountains Abound_copy-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11072919/Mountains.Abound_copy-trimmed.tar.gz)
this save is set up to perform the bug from the instant you spawn in.

### Steps to reproduce

1. obtain a container that has the pocket value `open_container` set to `true.` like a canning pot or similar.
2. use the `i` key and insert a tool with the `HOTPLATE` use function like a coffeemaker, along with any other item. 
3. activate the tool to heat up an item. if there is nothing to heat up the game will crash. if there is a food item to heat up the player will need to heat up the item to experience the crash.

### Expected behavior

This crash should not occur.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.18363.2274 (1909)
- Game Version: 0bba914 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Xedra Evolved [xedra_evolved]
]

### Additional context

no crash report was generated.

SurFlurer:
#### Summary
Bugfixes "Avoid spilling pocket contents into the origin item."

#### Purpose of change
Fix #64604

#### Describe the solution
Avoid spilling pocket contents into the origin item.

#### Testing
Avoids crash.

#### Additional context

If you try to heat up a bottle of water with an atomic coffeemaker inside a canning pot, there will be one non-ideal behaviour:
If the bottle is in the canning pot and you don't have enough storage volume to store the coffeemaker, the coffeemaker will drop to the ground and the water will not be heated. See the screenshot for better illustration.

Preparation:
image
After choosing to heat up the water:
image
Another attempt to heat up the water will generate the expected result.

However, before this PR it will lead to infinite loop and hang the game. So I don't think this is a regression.

### DIFF
--- a/src/avatar_action.cpp
+++ b/src/avatar_action.cpp
@@ -1180,7 +1180,7 @@ void avatar_action::use_item( avatar &you, item_location &loc, std::string const
         you.use( loc, pre_obtain_moves, method );
 
         if( parent_pocket && on_person && parent_pocket->will_spill() ) {
-            parent_pocket->handle_liquid_or_spill( you );
+            parent_pocket->handle_liquid_or_spill( you, loc.parent_item().get_item() );
         }
     }
 


NetSysFire:
Seems like a good backport candidate.

Issue #64528, opened 2023-03-24
RFC: Rework item repairs into faults


irwiss:
### Is your feature request related to a problem? Please describe.

Fwiw I've presented it to some devs on discord a while ago and the response was fairly positive, but likely not too accessible to people outside the timezone at the time; so I'm putting this as an issue as a RFC and polishing before I start working on it, or probe for potential blockers for this.

Right now item damage is abstracted into arbitrary number that you decrease using some magical toolkit (e.g. needle / tailor's kit / firearm repair kit etc) activator for the repair ui. This was pretty dodgy even before, but now with multi-material items like armor being widespread or complex repair kits made of mundane items (welding blanket) it creates even weirder situations.

Here are some primary examples of it:
* Vehicle parts had to reinvent repairing because the item repairing was lacking features, but now there are 2 separate flows that disagree with each other https://github.com/CleverRaven/Cataclysm-DDA/issues/24803 for example:
    * A quarterpanel on vehicle (based on metal sheet) requires 2 welding wire, 1 chunk of steel, 40 arc welder charges, welding goggles, hammer quality 2, mechanics 2.
    * A sheet metal item requires: 1 chunk of steel, 10 arc welder charges * failure chance... And that's it, not even welding goggles.
* There are 3 completely separate flows that do the same thing in different way: item repair, vehicle part repair, and mending.
* Repairing has very rudimentary support for multiple materials, in essene you can pick any material you want, so the player will always prefer to repair using the "cheaper" or more available material, for example: motorcycle boots are made from kevlar and cotton, you will never need kevlar sheets as long as you have (bulletresistant?) cotton patches in stock.

Secondary / minor issues:
* Magical "activators" which are amalgamation of tools into single item (instead of a containers + separate tools inside), for example firearm repair kit / gunsmith repair kit activator require mechanics overrides similar to #57398 or just funny situations like #53378. Or the magiclysm enchanted tailor's kit - https://github.com/CleverRaven/Cataclysm-DDA/issues/47849 which has it's item definition completely overriden due to json model limitations requires maintenance every new material added https://github.com/CleverRaven/Cataclysm-DDA/issues/47849 or a glue mod for modded materials.
[...]

PR #64622, opened 2023-03-27
Remove reinforcement / accurizing, clean up item damage

irwiss:
#### Summary
None

#### Purpose of change

Section 1 of https://github.com/CleverRaven/Cataclysm-DDA/issues/64528

#### Describe the solution

This PR removes the remains of reinforcement system and gun accurizing (see rationale in section 1 of linked issue).
As a result a bunch of stuff could be done:

Main stuff:
* item damage is clamped, it's range is [(degradation), itype::max_damage_] for non-count_by_charges and [0,1) for count_by_charges
* item::set_damage can no longer modify degradation, e.g. `it.set_degradation(3000); it.set_damage(0)` will no longer result in 0 degradation and 0 damage, instead degradation will stay and damage will be clamped to 3000.

Other:
* `itype::damage_min_`, `itype::damage_min()`, `item::min_damage()` are deleted
* `itype::damage_max_` is now `static constexpr int` with value 4000
* `item::damage_floor()` is removed - it is always equal to `item::degradation()`
* `itype/item::damage_levels()` now returns values [0 .. 5] instead of [-1 .. 4]

As a result a bunch of clamps to deal with negative damage were removed, some parts of the code are now much simpler, in particular item::mod_damage, item::damage_level() and everything that calls item::damage_level()

Item damage indicators and thresholds have changed as a result, the new thresholds are tested in degradation_test.cpp line 59+ and listed here:
|      damage | level | symbol | color         |
|------------:|:-----:|:------:|---------------|
|           0 |   0   |  `++`  | c_green       |
|     1 - 999 |   1   | `\|\|` | c_light_green |
| 1000 - 1999 |   2   |  `\|\` | c_yellow      |
| 2000 - 2999 |   3   |  `\|.` | c_light_red   |
| 3000 - 3999 |   4   |  `\.`  | c_red         |
|        4000 |   5   |  `XX`  | c_dark_gray   |

This means pristine items will appear as the previously-reinforced ++, || will be "slightly dinged" and so on

#### Testing

Some tests were adjusted:
* Tests that captured damage symbol/color inside text strings had them changed to new ones
* Tests that captured damage_level() value had them changed to new ones
* One degradation test had to be adjusted - it set degradation to over 4000, new behavior clamps it at 4000 causing expected mean degradation to be lower
* Some degradation tests assumed set_damage will modify degradation - they were modified as well
* New test added that matches [item damage : damage level : damage symbol : damage color] to the breakpoints

Overall though the test suite with the new test should catch the numeric-related changes.

~~For repair UI testing i'll be testing it manually~~ seems to be working for the most part


RenechCDDA:
On a related note it would help to remove this from vehicle parts too (assuming it isn't already).

You're not supposed to be able to *directly* reinforce them anymore but if you mininuke a vehicle you'll get weird overflows(underflows?) and end up with bits and pieces still alive but... reinforced.

Maleclypse:
So do all items have 4000 hp now? They might have all had the same before too I never knew how item hp truly worked.

irwiss:
Sorta... the item hp is weird - it's stored and acted upon in reverse; as damage not as hp, which makes some calculations a bit awkward, some stuff like vehicle parts have wrappers that invert it to count hp instead

but in short the absolute range of hp item can have is:
[0, 4000] - for non-count_by_charges items, when item reaches 4001 damage or -1 "hp" it is considered "destroyed" and the mod_damage caller should handle deleting it, when item had negative damage it was "reinforced" which everything interacting with item hp had to handle

or [0,1) for count_by_charges as they reduce charges instead of applying damage, charges -1 would signal mod_damage caller to "destroy" the item

as for old vs new; no changes for count_by_charges items, for normal items:
old - -1000 damage to -1 had negative damage_level and marked with ++, 0 damage marked with ||, 4000 - XX, 4001 damage destroyed
new - negative damage shouldn't be possible, 0 marked ++, 1-999 marked || etc... , 4000 - XX, 4001 damage destroyed

LordZanos:
Normally accurizing a weapon removes 30 dispersion. Might need a pass on weapon dispersion, not sure if the current weapon dispersion values are assuming that a gun actively being used would be accurized.

anothersimulacrum:
They are not.

Kromgart:
In recent experimentals people are encountering [an issue](https://old.reddit.com/r/cataclysmdda/comments/12f6y6s/does_uninstalling_vehicle_parts_supposed_to/) that uninstalling vehicle parts is 'damaging' them. Could this PR be the cause?
I cannot test this myself currently, otherwise I would make a proper issue.

Issue #58106, opened 2022-06-03
Add mouse support to all UI windows (where it is appropriate)


leemuar:
### Is your feature request related to a problem? Please describe.

Mouse is a very powerful tool, it increases usability, accessibility and generally could make things a lot easier for many players.

I want to continue a great work of [dseguin](https://github.com/dseguin) by adding mouse support to as many UI windows as I possibly can. This includes wheel scrolling, mouse hovers, mouse clicks and maybe even drag-n-drops.

This will be a project issue, to group all related issues together

Important: mouse features should not make UI unusable for people who has no mouse and old PC. Backward compatibility with old computers is an important feature, adding mouse should not break it

### Solution you would like.

Add mouse support to all UI windows

### Additional context

- [dseguin](https://github.com/dseguin) showed some examples on how to use mouse in UI elements: pull/57764
- [mqrause](https://github.com/mqrause) is working on new UI framework to ease UI handling: pull/56126

ZeroInternalReflection:
I've been making my own notes on UI consistency (raised on Discord, but not yet formalized in any issues).  Since you mentioned the to-do list growing, here are some of the screens that I've identified as not currently having any mouse support in tiles:

- Achievements (`)`)
- Attach gun modifications (`a`->gun modification->Select gun)
- Active world mods (`Esc`->Active world mods)
- AIM (`/`)
- Auto notes manager (`Esc`->Auto notes manager)
- Auto pickup manager (`Esc`->Auto pickup manager)
- Base camp (`e`->Bulletin board)
- Bionics (`p`)
- Body status (`@`->`s`)
- Compare (`I`->Select two items)
- Construction (`*`)
- Dialogue (`e`->NPC->Talk)
- Diary (unbound by default)
- Examine vehicle (`e`->vehicle)
- Factions (`#`)
- Medical (`N`)
- Mutations (`[`)
- Options (`Esc`->Options)
- Player info (`@`)
- Proficiencies (`@`->`p`)
- Safe mode manager (`Esc`->Safe mode manager)
- Sort armor (`+`)
- View items/monsters (`V`)

This game has a lot of different UI screens.
Mouse support in curses is of course, a more complicated issue.  Though it does work on the `f`ire screen...

Night-Pryanik:
Regarding your question in #58454:
>What UI window is this? i Inventory? /Extended Inventory?

[...]

PR #64632, opened 2023-03-28
Mouse hover support for inventory screens

katemonster33:
…inventory screen as well as the inventory multi-selector.

#### Summary
Features "Made it so that inventory items highlight when hovered over with the mouse"

#### Purpose of change

When I first installed the game, I was very confused by how the mouse worked as expected on the main menu, but not anywhere else. On the inventory screen, though I found that clicking items selects them as expected, without seeing the items become highlighted as the mouse is moved over them, it appeared that the dialog did not support the mouse at all. I feel this change adds UI consistency. Relevant to #58106, does not fully fix it.

#### Describe the solution

Registered actions for MOUSE_MOVE and COORDINATE on the action handler. COORDINATE turns on mouse hover events, but I found that I needed to register a separate action for MOUSE_MOVE to see the event propagate up to the on_input functions.

n/a

#### Testing
I loaded both the pick-selector and the multi-selector inventory views, by using the 'i' key and 'd' keys, respectively. I am not 100% familiar with all the screens in CDDA yet but it didn't look like this was used in any other places.

#### Additional context
n/a

### DIFF
--- a/src/inventory_ui.cpp
+++ b/src/inventory_ui.cpp
@@ -3371,8 +3371,8 @@ std::pair<const item *, const item *> inventory_compare_selector::execute()
 
         if( input.entry != nullptr ) {
             highlight( input.entry->any_item() );
-            if ( input.action == "SELECT" ) {
-                toggle_entry(input.entry);
+            if( input.action == "SELECT" ) {
+                toggle_entry( input.entry );
                 just_selected = input.entry;
             }
         } else if( input.action == "TOGGLE_ENTRY" ) {
--- a/src/inventory_ui.cpp
+++ b/src/inventory_ui.cpp
@@ -3972,7 +3972,7 @@ int inventory_examiner::execute()
             if( highlight( input.entry->any_item() ) ) {
                 ui_manager::redraw();
             }
-            if ( input.action == "SELECT" ) {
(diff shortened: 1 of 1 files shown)


dchieh222:
This seems like it'll finally lead to tooltips!

katemonster33:
Added a commit... noticed regressions in mouse support in pocket examiner (inventory_examiner) screen and the inventory compare screen, fixing the bugs has caused mouse hover support to now be working as intended in those screens as well.

src/inventory_ui.cpp
```diff
@@ -3525,7 +3531,9 @@ void inventory_multiselector::on_input( const inventory_input &input )
 
     if( input.entry != nullptr ) { // Single Item from mouse
         highlight( input.entry->any_item() );
-        toggle_entries( count );
+        if( input.action == "SELECT" ) {
+            toggle_entries( count );
+        }
     } else if( input.action == "TOGGLE_NON_FAVORITE" ) {
```
andrei8l, line 3537:
```suggestion
    }
    if( input.action == "TOGGLE_NON_FAVORITE" ) {
```
This should make clicking on chevron (from #64750) work for multiselectors too

leemuar:
well done, thanks! one step closer to better mouse support

PR #64633, opened 2023-03-28
Math expression parser

andrei8l:
#### Summary
None

#### Purpose of change
`arithmetic` and friends (`compare_num`, `u_val`, and `npc_val`) are too verbose and awkward to use for any real calculation in EOCs. Anything that needs more than one operator gets quickly untenable with curly braces and quotation marks.

#### Describe the solution
Quick example of math in JSON:
```json
  { 
    "id": "EOC_math_test",
    "condition": {
        "math": [ "u_val('stamina')", ">=", "500" ] // exactly the same keyword and interface everywhere
    },
    "effect": [
      { "u_message": "math test! "},
      { "math": [ "u_math_test", "=", "u_val('stamina')" ] }, // assign to variable `math_test` in scope `u` (alpha talker)
      { "math": [ "math_test", "=", "u_val('stamina') + 3900" ] }, // assign to variable `math_test` in global scope
      { "math": [ "u_val('pain')", "=", "60 + rng(0,20)" ] }  // assign to special value exposed through stringy function `val` in the `u` scope
    ]
  },
```
Add a validating math expression parser. It can parse any math expression you'd want with simple JSON, can interact with scoped dialogue variables, and run arbitrary functions to calculate or extract data from the game just like EOC functions. 

There's also an `u_val` shim (seen in the snippet above) that lets it reuse some EOC code and can be used to replace almost all instances of `u_val` and `npc_val`.

<details>
<summary>A couple of more examples, taken and from our data:</summary>

<details>
<summary>Instead of this</summary>

```json
{ "compare_num": [ { "u_val": "pos_z" }, ">=", { "const": 0 } ] }
```

</details>

<details>
<summary>You can now write this</summary>

```json
{ "math": [ "u_val('pos_z')", ">=", "0" ]}
```
</details>

<details>
<summary>Instead of this</summary>

```json
{ "value": "LUMINATION", "add": { "arithmetic": [ { "u_val": "morale" } ] } }
```

</details>

<details>
<summary>You can now write this</summary>

```json
{ "value": "LUMINATION", "add": { "math": [ "u_val('morale') * 3 - rng(0, 100)" ] } }
```
and do some math in there at the same time.
</details>

</details>

<details>
<summary>Expressions are validated with helpful messages when something is wrong and there should never be any crashes from bad syntax. Example:</summary>

Screenshot from 2023-03-28 06-20-49

</details>

Internally, this parser is based on shunting-yard with some bells and whistles: validation, unary operators, variadic, and stringy functions. The implementation is hidden from the interface so it could be swapped out theoretically.

TODO:
- [x] documentation
- [x] another test unit for JSON integration

Instead of stringy functions, I was originally accomplishing their goal with scoped variables like `u_spell_level_test_spell_pew` but that was both ambiguous and ugly (it's now `u_val('spell_level', 'spell: test_spell_pew')` through the `u_val` shim but I plan to flatten these kinds of things to `u_spell_level('test_spell_pew')`).

Other parsing algorithms are more efficient and/or more powerful, but their descriptions on Wikipedia weren't written for people like me.

Using a parser generator instead of hand-rolling one: this needs a build-time dependency and is possibly incomprehensible to the average contributor.

Using lua to parse math and/or replace EOC code: I got a mildly negative reaction to this when I proposed it a few months ago so I didn't commit and I don't know what it'd entail.

#### Testing
Test unit that provides 93% coverage for `math_parser`.
EOC test that targets all the integration options.

I've also been fuzzing the parser for a couple of days with libFuzzer and [this fuzz target](https://github.com/andrei8l/Cataclysm-DDA/commit/6334f693e49c4e1522f13bce16c4d10fe4c7efc1) with a seed corpus extracted from the test unit. It's not included in this PR because we're not set up for fuzzing.

#### Additional context
I've squashed the commits into logical units so this can be reviewed by commit.

Syntax is validated, but whitespace style is not. I don't have a solution for that.

If this is accepted, I plan to slowly replace and eventually remove `arithmetic`, `compare_num`, `u/npc_val`, and `compare_var`.

### DIFF
--- a/data/mods/TEST_DATA/EOC.json
+++ b/data/mods/TEST_DATA/EOC.json
@@ -12,5 +12,48 @@
         "success_message": "All at once your body feels like it's beeing pulled in 1000 different directions."
       }
     ]
+  },
+  {
+    "type": "effect_on_condition",
+    "id": "EOC_math_test_equals_assign",
+    "condition": { "math": [ "u_val('stamina')", "==", "500" ] },
+    "effect": [ { "math": [ "math_test", "=", "-3^2" ] } ]
+  },
+  {
+    "type": "effect_on_condition",
+    "id": "EOC_math_test_greater_increment",
+    "condition": { "math": [ "math_test", ">", "0" ] },
+    "effect": [ { "math": [ "math_test", "+=", "1" ] } ],
+    "false_effect": [ { "math": [ "math_test", "--" ] } ]
+  },
+  {
+    "type": "effect_on_condition",
+    "id": "EOC_math_switch_math",
+    "effect": {
+      "switch": { "math": [ "math_test" ] },
+      "cases": [
+        { "case": -1, "effect": { "math": [ "math_test_result", "=", "1" ] } },
+        { "case": 9, "effect": { "math": [ "math_test_result", "=", "2" ] } },
+        { "case": 10, "effect": { "math": [ "math_test_result", "=", "3" ] } }
+      ]
+    }
+  },
+  {
+    "type": "effect_on_condition",
+    "id": "EOC_math_diag_assign",
+    "effect": [ { "math": [ "u_val('stamina')", "/=", "2" ] } ]
+  },
+  {
+    "type": "effect_on_condition",
+    "id": "EOC_math_var",
+    "condition": { "days_since_cataclysm": { "math": [ "math_test_result" ] } },
+    "effect": [ { "u_message": "test var math" } ]
+  },
+  {
+    "type": "effect_on_condition",
+    "id": "EOC_math_duration",
+    "recurrence": { "math": [ "math_test_result" ] },
+    "global": true,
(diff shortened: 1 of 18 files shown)


Qrox:
For one second I though we were adding TeX to CDDA.

andrei8l:
@Ramza13 since you've been doing the overwhelming majority of EOC work, this change can only really succeed if you agree with it. Is there anything here you don't like, or that seems to be more difficult than with `arithmetic` (and friends)?

Ramza13:
This looks fantastic!  You already handled all the things I thought might be an issue.  
Now I feel good I never got around to converting everything over to compare_num and arithmetic.

andrei8l:
Thanks for merging!!

I-am-Erk:
Ahaaaaaaaah

Fine if you insist I guess I'll get back to contributing.

PR #64636, opened 2023-03-28
funiture desc changed to indicate deployability

dchieh222:
This changes the furniture descriptions which can be placed via the construction menu to match the description of the water and wind mill, which indicates that it can be placed via the construction menu.

#### Summary
Content "Furniture description update to indicate deployability"

#### Purpose of change

This helps indicate that the furniture can be deployed via the construction menu and moves it in line with the watermill and windmill objects to have consistent description.

#### Describe the solution

The change in the description guides the player to the construction menu.

The most significant effort, and I would have thought to be the superior one, was to add the use_action flag with type deploy_furn and then indicate the appropriate furniture type. This would also allow the deployed furniture to respond to examines with requests to be undeployed, which would provide more clarity. 

However, this approach bypasses the time, requirements and activity efforts triggered by the construction menu.  Since I am not certain how to edit the deploy_furn action, I am attempting a description change fix instead. I looked into EoC solutions but none appeared to fulfill what I needed to do(increment time, and add an activity level).

#### Testing

Executed and tested this with my own version of Cataclysm, and confirmed that the description appeared as expected with no consequences to the action. 

#### Additional context

I believe the appropriate solution is to have the correct ACT_BUILD function set for deploy_furn_actor::use in iuse_actor.cpp & iexamine::deployed_furniture in iexamine.cpp. This should pull from the JSON so there is consistent behavior all around. 

However, I'm not familiar with the code and act_build appears particularly gnarly.

Any advice on how to do this correctly, or better, would be appreciated.


Karol1223:
I am very much not a fan of this. We had a similar thing with the description of scissors for years and I hated it. This would be better off as being mentioned somewhere in the UI. The description should be something your character sees or knows, not game tips.

dchieh222:
Sometimes there's no better place to put "how to interact", I think. For this PR, I'll resume trying to do this with the deploy_furn action item, and eventually loop it to use the Build activity when I figure it out.

data/json/items/furniture.json
```diff
@@ -19,7 +19,7 @@
     "type": "GENERIC",
     "id": "armchair",
     "name": { "str": "armchair" },
-    "description": "A simple upholstered chair with armrests.  Soft and fairly warm, it definitely beats the ground for sleeping on, though not by much.",
+    "description": "A simple upholstered chair with armrests.  Soft and fairly warm, it definitely beats the ground for sleeping on, though not by much. Can be placed via the construction menu.",
```
Maleclypse, line 22:
```suggestion
    "description": "A simple upholstered chair with armrests.  Soft and fairly warm, it definitely beats the ground for sleeping on, though not by much.  Must be deployed upright to be used.",
```

Would this kind of language be better at maintaining 4th wall while also pointing out that something must be done to place it properly?  Also the item sprite could be drawn on it's side to represent the item not being upright.

narc0tiq: We could state (somewhat reasonably) that the item-form is packed for transport, and therefore it "Must be unpacked to be used"? Or even, in keeping with other messages, "You need to unpack this to sit on it" -- speaking to the character, not the player.

PR #64639, opened 2023-03-28
Implement unused CLOSE_ADV_INV option

### LINKED ISSUE #58249: "Close advanced inventory on move all" option does not close the advanced inventory.
### Describe the bug

Enabling the "close advanced inventory on move all" interface option appears to do nothing. When using the "move all items" command, the advanced inventory interface remains open after moving the items. This defeats what I presume to be the point of this option, eliminating an extra key press when moving lots of items over a distance.

### Steps to reproduce

1. Open the options menu and move to the "interface" section. On line 19, change "Close advanced inventory on move all" to "true". (Default is false).
2. During gameplay, open the advanced inventory menu (default: /) and use the "move all items" command (default: ,) to move some items.
3. The advanced inventory tab should still be open after the items are moved.

### Expected behavior

I expect that enabling this option would close the advanced inventory menu upon pressing the "move all items" key (default: ,). Either this option does not work, it's description is misleading or I am doing something wrong.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19043.1706 (21H1)
- Game Version: ca05dd4 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

strategineer:
#### Summary
None

#### Purpose of change

Fixes #58249

Implement unused CLOSE_ADV_INV option which closes the AIM menu when the player uses the "move all items" keybind and after they finish moving the items.

#### Describe the solution

The CLOSE_ADV_INV option wasn't doing anything (it probably worked at some point I'm guessing).

I've implemented the desired behavior for it. 

Deleting the option.

#### Testing

- I spawned 1000 glass shards (which are going to take a while to move)
- I set the CLOSE_ADV_INV option to true
- I used the "Move all items" key to move all the glass shards to another tile
- Noticed that the menu didn't close.
- I implemented my fix 
- I repeated these steps and noticed that after after pressing the "Move all items" key, AIM menu stayed up during the processing and when it was done, the menu closed as expected.
- Also, I repeated these steps with the  CLOSE_ADV_INV option to false and noticed that the menu did not close after moving the glass shards (as expected)

### DIFF
--- a/src/advanced_inv.cpp
+++ b/src/advanced_inv.cpp
@@ -1584,6 +1584,7 @@ void advanced_inventory::display()
         } );
     }
 
+
     while( !exit ) {
         if( player_character.moves < 0 ) {
             do_return_entry();
@@ -1605,6 +1606,10 @@ void advanced_inventory::display()
         // current item in source pane, might be null
         advanced_inv_listitem *sitem = spane.get_cur_item_ptr();
         aim_location changeSquare = NUM_AIM_LOCATIONS;
+
+        if( !is_processing() && move_all_items_and_waiting_to_quit ) {
+            break;
+        }
 
         const std::string action = is_processing() ? "MOVE_ALL_ITEMS" : ctxt.handle_input();
         if( action == "CATEGORY_SELECTION" ) {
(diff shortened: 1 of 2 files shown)


RenechCDDA:
Oops, I think that might've been a regression on my part...

PR #64650, opened 2023-03-28
fursuit zombie no longer weighs nothing and can no longer bite

Daved27hundred:
#### Summary
Bugfixes "Fixed malicious mane having 0 weight or volume"

#### Describe the solution

oversight in #61173 caused the malicious mane to have a weight of zero, and a volume of zero. While I was fixing that I also removed the bite attack - I have never seen a fursuit that has that capability, usually fursuit heads just a static face or at most a moving jaw that shouldn't do more than harmless "noms." 

remove the whole zombie? I don't want to, I think it's nice flavor.


MNG-cataclysm (approved):

anothersimulacrum (changes requested):
You should just apply the changes from https://github.com/CleverRaven/Cataclysm-DDA/pull/64494/commits/58294ca34f39fda9d091991c6de1b1dab1079df8

Daved27hundred:
done.

Issue #56237, opened 2022-03-20
More non-electric light-producing furniture


Sathra225:
### Is your feature request related to a problem? Please describe.

Creating light sources that don't use electricity seems to be limited to items or fireplaces. The handheld items have a limited range (basically only user tiles) and fireplaces are bulky and block movement and vision, or produce smoke.
More of a QoL thing for innawoods runs, but could be useful for other purposes.

### Solution you would like.

Either a built furniture or deployable item that is essentially a big oil lantern on a pole. Casing and reflectors (optional) around the flame to spread the light around and protect from wind. Consumes charges over time while generating light, can be refilled.
Might need new code though.

### Describe alternatives you have considered.

Darkness without large fires?
Or some kind of small, transparent fireplace that can be walked on the tile of without setting yourself on fire.

### Additional context

Houses before gas street lights could have oil lamps hanging above the entrances suspended from ironwork. These could light the footpaths but not much further due to not having reflectors. Not sure why, either cost or soot buildup.

Marrim:
Wall torches, too!

robotic-gamer:
some ideas
- oil street lamp (replica of the type used in towns before gasification, recipe found in history books?)
- glow in the dark objects, when in light they charge, when in dark they discharge
- emergency lights (found by exits and can be used as a long lasting torch or dismantled for motorbike battery)
- emergency signs light with tritium paint, a bulky nightlight that can't be turned off
- firefly "jar" uses fuel that can be made from water and sugar , needs firefly egg to craft)
- mirrors, provide light when light, could be used to move light around corners
- carbon arc lamp, primitive and energy hungry but very bright
-  various mushrooms that can be foraged and used as a temporary glowstick
- rattles that use sugar and triboluminescence to create a brief moment of light when activated, niche use to attract enemies
- craftable flares (could be made from saltpeter and charcoal)
- solar powered lawn light (uses a light battery as a battery and so could be unloaded to charge batteries)

Jarewill:
> Did some thinking.

I like the idea of turning the oil lamp into a furniture, however it really should have more range than just 3x3.
3x3 is the range of the oil lamp itself, which would make those sources not really worth making IMO, as you can just put an oil lamp down on the floor.
The OP mentioned using reflectors, and while I don't know how exactly those work, if they can extend the light range to even 5x5 (aisle lights range) it will be way better.

Sathra225:
[...]

PR #64653, opened 2023-03-28
Add harvestable bioluminescent organs to fireflies

teackot:
#### Summary
Content "Add harvestable bioluminescent organs to fireflies"

#### Purpose of change

This PR was inspired by #56237. Fireflies have their own non-electric light sources, so why not make use of them? It may also be useful for innawoods runs.

#### Describe the solution

Added a bioluminescent_sack item and its lit and empty stages. Added luciferin ammo for it. Added bioluminescent sack to the firefly harvest.

Luciferin could be harvested like blood, but it would require a separately crafted lamp and would also expose it to oxygen, starting the chemical reaction.

#### Testing

Spawned, killed and butchered a firefly, getting a bioluminescent sack. Activated it and waited for it stop glowing (~ 5 min).

#### Additional context

Fireflies emit light by controllably exposing luciferin in their abdomens to oxygen. I'm not sure if they have dedicated organs for this IRL, but even if they don't - it can be a mutation in the game.

### DIFF
--- a/data/json/harvest.json
+++ b/data/json/harvest.json
@@ -1298,7 +1298,8 @@
       { "drop": "mutant_bug_lungs", "type": "flesh", "mass_ratio": 0.0035 },
       { "drop": "mutant_bug_organs", "type": "offal", "mass_ratio": 0.015 },
       { "drop": "chitin_piece", "type": "bone", "mass_ratio": 0.01 },
-      { "drop": "egg_firefly", "type": "offal", "base_num": [ 0, 3 ] }
+      { "drop": "egg_firefly", "type": "offal", "base_num": [ 0, 3 ] },
+      { "drop": "bioluminescent_sack", "type": "offal", "base_num": [ 0, 3 ] }
     ]
   },
   {
--- a/data/json/items/ammo.json
+++ b/data/json/items/ammo.json
@@ -1476,6 +1476,18 @@
     "ammo_type": "glowstick_juice"
   },
   {
+    "id": "luciferin",
+    "type": "AMMO",
+    "name": { "str_sp": "luciferin" },
+    "symbol": "~",
+    "description": "A chemical used by fireflies to glow in the dark. Reacts with oxygen, producing light.",
+    "volume": "1 ml",
+    "weight": "1 mg",
+    "color": "green",
+    "flags": [ "TRADER_AVOID" ],
+    "ammo_type": "luciferin"
+  },
+  {
     "type": "AMMO",
     "id": "graphite",
     "name": { "str_sp": "graphite" },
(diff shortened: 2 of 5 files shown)


Karol1223:
The realism department is severely lacking here. @Venera3 should comment but to me the issues are:
- Luciferin decays in 3 weeks at -20, yours doesn't at all
- Luciferin reacts with oxygen only with an enzymatic catalyst which would go inert likely several minutes after death of the animal
- Tearing open the sack to activate it would let in enough oxygen to use up all the luciferin at once

Fris0uman:
What Karol said, bioluminscence overall needs living stuff to work. Irl bioluminescent lamps use bacteria, I know they exist, I don't know how well they work and how much of a pain they are to keep alive though, but I suspect it's not worth the effort for a survivor.

Karol1223:
Just to make sure the message got across though - I am not advocating for 'don't bother with this'. I am saying that this is prime content if you move it to Innawood. This is exactly the kind of vibe that Innawood gives off. Although this may need to be discussed with the maintainer beforehand, not sure what's the consensus from their point of view.

Light-Wave:
No, I've been trying to keep Innawood as close to mainline and as realistic as practically possible.
Maybe this could be used for Magiclysm somehow? I could see luciferin being used as an alternative to glowing mushrooms in potions of illumination for example.

Karol1223:
> Why not keep the firefly alive and keep it in a terrarium? Mr. Lapin has such a thing in his cabin. 
> Check out the firefly_terrarium item.

This is not a novel idea and every time it gets brought up someone has to come and say that the survivor is not making a self-sustaining terrarium for a mutant. Lapin is supposed to have the knowledge to pull it off, the survivor not. Plus the author of that item has been against it on every occasion possible.

AngularFruit:
> The realism department is severely lacking here. @Venera3 should comment but to me the issues are:
> 
> * Luciferin decays in 3 weeks at -20, yours doesn't at all
> * Luciferin reacts with oxygen only with an enzymatic catalyst which would go inert likely several minutes after death of the animal
> * Tearing open the sack to activate it would let in enough oxygen to use up all the luciferin at once (if the enzyme was still present)
> 
> This all looks to me like something completely nonviable for mainline. Innawoods would probably gladly adopt it though.

https://www.biwahaku.jp/smith/facts.html

https://hakaimagazine.com/features/secret-history-bioluminescence/

I'm not a contributer or anything, just browsing but I figured I'd share these links about Ostracods, their shells/bodies could be dried up and stored for years and then when mixed with water could stay lit for a while.

I know they aren't fireflies, but maybe instead of an organ you ground up it's chitin into a powder and it acted the same way? 

My apologies if this isn't the place for it.

kevingranade:
There are a bunch of light sources you can improvise,  but this isn't one of them.

As a general practice,  please don't reflexively shop changes around for different mods just because they don't fit in mainline.  Finding a place to put every possible change is way less important than different mods maintaining their own sense of direction.

PR #64654, opened 2023-03-28
Added missing nutrients to teas

Nebnis:
#### Summary
Balance "Added vitamins to a bunch of tea"

#### Purpose of change
Ever since the change that removed most healthy buffs to food, a bunch of them ended up having only calories even despite a bunch of their ingredients having vitamins. So i went ahead and added some vitamin c to them and some of the other missing teas as well as most teas are known to have vitamin C in them.

#### Describe the solution
Added vitamins to teas.

#### Testing
Logged and everything seems good, other than the percentage i added being slightly off by 1% for some reason

### DIFF
--- a/data/json/items/comestibles/drink.json
+++ b/data/json/items/comestibles/drink.json
@@ -187,6 +187,7 @@
     "description": "A healthy beverage made from chamomile flowers steeped in boiling water.  Can be used to treat insomnia.",
     "price": 100,
     "price_postapoc": 25,
+    "vitamins": [ [ "vitC", 10 ] ],
     "flags": [ "EATEN_HOT", "NUTRIENT_OVERRIDE", "NO_AUTO_CONSUME" ],
     "fun": 1
   },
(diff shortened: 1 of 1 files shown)


NetSysFire:
I'd prefer to have some sources to back up the realism here. If you can find some that would be amazing.

Nebnis:
> I'd prefer to have some sources to back up the realism here. If you can find some that would be amazing.

Would this suffice? https://ratetea.com/topic/vitamin-c-tea/62/#:~:text=Some%20herbal%20teas%2C%20including%20hibiscus,much%20less%20than%20fresh%20fruit.

I guess from reading black tea doesn't have vitamins C I mainly wanted to make Tea healthy again since they are known to be that with antioxidants overall.

fairyarmadillo:
Most teas have negligible nutritional value of any kind irl. There are a lot of bottled drinks containing tea which have added vitamins, and the fresh leaves may have some nutritional value, but looking over the nutrition facts for several different kinds of black, green, and even hibiscus tea bags, I don't see any vitamin C, nor anything else for that matter.

PR #64656, opened 2023-03-29
Refactor monster::plan to something more readable and encapsulated

KorGgenT:
#### Summary
Infrastructure "Refactor monster::plan to something more readable and encapsulated"

#### Purpose of change
The purpose of this is to reduce the length of a single function monster::plan(), by breaking out pieces of it, and by encapsulating those pieces into other functions, label them.

#### Describe the solution
the solution is the same as the description of the purpose

#### Testing
really just compile, nothing actually changed functionally at all. it's to do with monsters deciding whether they want to attack or not.

#### Additional context
after seeing the results, i became a little less confident that i was adding too much, since the number of additions is larger than deletions, but overall i decided that the point of this PR wasn't to reduce the number of lines in the file itself, but the number of lines in that specific function. it reduced by around 60, and there's probably some more that can go later.

there were a very large number of local variables, so to make it easier to pass them around into other functions, i created a struct to store them in.

### DIFF
--- a/src/monmove.cpp
+++ b/src/monmove.cpp
@@ -295,122 +295,184 @@ float monster::rate_target( Creature &c, float best, bool smart ) const
     return FLT_MAX;
 }
 
+struct monster_plan {
+    explicit monster_plan( const monster &mon );
+
+    // Bots are more intelligent than most living stuff
+    bool smart_planning;
+
+    bool fleeing;
+    bool docile;
+
+    const bool angers_hostile_weak;
+    const bool fears_hostile_weak;
+    const bool placate_hostile_weak;
+    const int angers_hostile_near;
+    const int angers_hostile_seen;
+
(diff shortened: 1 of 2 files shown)


src/monmove.cpp
```diff
@@ -295,122 +295,184 @@ float monster::rate_target( Creature &c, float best, bool smart ) const
...
+    bool smart_planning;
+
+    bool fleeing;
+    bool docile;
+
+    const bool angers_hostile_weak;
+    const bool fears_hostile_weak;
+    const bool placate_hostile_weak;
+    const bool angers_hostile_near;
+    const bool angers_hostile_seen;
```
ZhilkinSerg, line 311:
```suggestion
    const int angers_hostile_near;
    const int angers_hostile_seen;
```

PR #64659, opened 2023-03-29
Fix "Play Now" options bringing up the world generation prompt

### LINKED ISSUE #61273: Play Now interrupted by world creation
### Describe the bug

When choosing to Play Now (with default or random scenario), the player is put straight into a game... unless they don't have a world generated in which case they have to navigate the world generation dialog. I propose that this should not be the case, and that a world should be generated with the default settings if one doesn't exist.

### Steps to reproduce

1. Delete all worlds
2. Start a game with Play Now
3. Encounter world creation dialog

### Expected behavior

Game starts immediately

### Screenshots

_No response_

- OS: Linux
    - OS Version: LSB Version: n/a; Distributor ID: ManjaroLinux; Description: Manjaro Linux; Release: 22.0.0; Codename: Sikaris; 
- Game Version: 0.F recent master branch
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

strategineer:
#### Summary
None

#### Purpose of change

Fixes #61273

Regression introduced by https://github.com/CleverRaven/Cataclysm-DDA/commit/fdbe06d53a48e7a3191f7ffb23e5b49b6de7f146

#### Describe the solution

Pass in the correct parameters to the `world_generator.pick_world` function.

Not fixing this bug.

#### Testing

Started new games with the various options  and got the expected behavior.

### DIFF
--- a/src/main_menu.cpp
+++ b/src/main_menu.cpp
@@ -962,7 +962,8 @@ bool main_menu::new_character_tab()
         // First load the mods, this is done by
         // loading the world.
         // Pick a world, suppressing prompts if it's "play now" mode.
-        WORLD *world = world_generator->pick_world( true, sel2 == 3 || sel2 == 4 );
+        const bool is_play_now = sel2 == 3 || sel2 == 4;
+        WORLD *world = world_generator->pick_world( !is_play_now, is_play_now );
         if( world == nullptr ) {
             return false;
         }


src/main_menu.cpp
```diff
@@ -962,7 +962,8 @@ bool main_menu::new_character_tab()
         // First load the mods, this is done by
         // loading the world.
         // Pick a world, suppressing prompts if it's "play now" mode.
-        WORLD *world = world_generator->pick_world( true, sel2 == 3 || sel2 == 4 );
+        const bool is_play_now = sel2 == 3 || sel2 == 4;
+        WORLD *world = world_generator->pick_world( !is_play_now, is_play_now );
```
strategineer, line 966: ` WORLD *worldfactory::pick_world( bool show_prompt, bool empty_only )`

Maleclypse: Would it be easy to write a test for this so it doesn't regress again?

strategineer: Depends what you mean by easy :)

My gut says that I would have to fix this same bug a few more times before it would be worth it to write a good test for this.

Maleclypse: Makes sense.

PR #64662, opened 2023-03-29
feat: moar option groups

sonphantrung:
#### Summary
Interface "Add more option groups"

#### Purpose of change
This PR aims to fix the "bloated UI" thing that has haunted the repo for years (or just make the settings screen look much better)

#### Describe the solution
Add more option groups to options which are sticked together in a column ~~(I only add Measurement Units option group, will add more later)~~ 100% added now!

Not doing this and suffering from CHUNKY UI

#### Testing
It werks, check below

#### Additional context
Considering the fact that Night is currently under a hiatus, I think I maybe try to update his No Hope mod (and possibly, take ownership), but anyways, I LOVE UNFOLDING/FOLDING lines ;)
image


sonphantrung:
Damn, I have to git clone and then astyle manually now

src/options.cpp
```diff
@@ -1569,106 +1583,121 @@ void options_manager::add_options_general()
...
          0, 128, 100, COPT_NO_SOUND_HIDE
        );
 
     get_option( "SOUND_EFFECT_VOLUME" ).setPrerequisite( "SOUND_ENABLED" );
 
-    add( "AMBIENT_SOUND_VOLUME", "general", to_translation( "Ambient sound volume" ),
+    add( "AMBIENT_SOUND_VOLUME", page_id, to_translation( "Ambient sound volume" ),
          to_translation( "Adjust the volume of ambient sounds being played by the game." ),
          0, 128, 100, COPT_NO_SOUND_HIDE
        );
```
Qrox, line 1703: The previous opening brackets at line 1671-1673 are unclosed but I cannot add the suggestion below line 1705.

src/options.cpp
```diff
@@ -2008,30 +2059,33 @@ void options_manager::add_options_interface()
...
     "right"
        );
+    });
 
     add_empty_line();
-
+    add_option_group( "interface", Group( "mouse_cont_opts", to_translation( "Mouse Control Options" ),
+                                          to_translation( "Options regarding mouse control." ) ),
+    [&]( const std::string & page_id ) {
     add( "ENABLE_JOYSTICK", "interface", to_translation( "Enable joystick" ),
```
Qrox, line 2089: Should replace "interface" in this group with `page_id`

src/options.cpp
```diff
@@ -2065,79 +2120,89 @@ void options_manager::add_options_graphics()
...
        );
 
     get_option( "ALT_TITLE" ).setPrerequisite( "ENABLE_ASCII_TITLE" );
 
     add_empty_line();
 
+    add_option_group( "graphics", Group( "term_opts", to_translation( "Terminal Display Options" ),
+                                          to_translation( "Options regarding terminal display." ) ),
+    [&]( const std::string & page_id ) {
     add( "TERMINAL_X", "graphics", to_translation( "Terminal width" ),
```
Qrox, line 2206: Should replace "graphics" in this group with `page_id`. Also should add closing brackets to below line 2214.

src/options.cpp
```diff
@@ -2354,71 +2430,75 @@ void options_manager::add_options_graphics()
...
+    add( "PIXEL_MINIMAP_BG", page_id, to_translation( "Background color" ),
          to_translation( "What color the minimap background should be.  Either based on color theme or (0,0,0) black." ),
     {
         { "theme", to_translation( "Theme" ) },
         { "black", to_translation( "Black" ) }
     }, "black", COPT_CURSES_HIDE
        );
+    });
 
     get_option( "PIXEL_MINIMAP_BLINK" ).setPrerequisite( "PIXEL_MINIMAP" );
```
Qrox, line 2485:
```suggestion

    get_option( "PIXEL_MINIMAP_BLINK" ).setPrerequisite( "PIXEL_MINIMAP" );
    });
```

mqrause:
You're definitely missing a closing `}` somewhere.

ZhilkinSerg:
Have you tried to compile it locally?

src/options.cpp
```diff
@@ -1687,146 +1717,154 @@ void options_manager::add_options_interface()
...
+        },
+        "symbol"
+           );
 
-    add( "HIGHLIGHT_UNREAD_RECIPES", "interface",
-         to_translation( "Highlight unread recipes" ),
-         to_translation( "If true, highlight unread recipes to allow tracking of newly learned recipes." ),
-         true
-       );
+        get_option( "ALT_TITLE" ).setPrerequisite( "ENABLE_ASCII_TITLE" );
```
Qrox, line 1866: Apparently you commited some incorrect changes suggested by reviewdog.

I-am-Erk:
I'm not sure I see what you're going for here given the code stuff that's going on, so take this with a grain of salt, but I want to caution that adding groups and collapsible options isn't the solution to what we mean by option bloat.

I-am-Erk:
> > I'm not sure I see what you're going for here given the code stuff that's going on, so take this with a grain of salt, but I want to caution that adding groups and collapsible options isn't the solution to what we mean by option bloat.
> > It's not a bad thing to have some UX improvement, I just want to make sure we're on the same page here.
> 
> Hmm, I thought that UI bloat meant "there are too many options here, and it looks so confusing!"?

One thing it means is that our menus are overloaded with things, yes, but the problem is also that many of these things can, when adjusted to their limits, break the game in ways the player may not anticipate. For example, setting item spawns too low can make it impossible to complete some quests, and a player adjusting this value does not necessarily realize this. Option bloat refers to the combined problem that our menus are too full of poorly organized options, and also that we need better curation of those options. How exactly to curate them is still a matter of debate, and the more options we have the higher the chance that some of them will interact in ways that do unexpected harm to the game.

sonphantrung:
All (reviewdog's) convos have been resolved now

OmegaCenti:
Crow: :crowBongo: who collapsed my goddamn options menu
OmeegaCenti (Too many E's): yeah, I don't like that either crow
Crow: is there a button to uncollapse all at least?
OmeegaCenti (Too many E's): you have to uncollapse them, I think they stay uncollapsed at least
Crow: oh gosh it's all of them
Crow: :bcaShook:
Crow: mah MENUS
Cobweb [0.G Stable]: Somehow this is better
OmeegaCenti (Too many E's): yeah, it definitely wasn't hard to find things before without nested menus
Crow: agony
OmeegaCenti (Too many E's): now what took a heading and a number, now takes a heading, a subdivision AND a number
Crow: it re-collapses them on re-entering the menu
OmeegaCenti (Too many E's): you get more button presses and you get more button presses! everyone gets more button presses!

PR #51085, opened 2021-08-26
Added explicit num capture state to drop/clean menus to allow numpad navigation

StStep:
#### Summary
Bugfixes "Added explicit num capture state to drop/clean menus to allow numpad navigation"

#### Purpose of change
Fixes #50860 and Fixes #50819 

#### Describe the solution
As discussed on #50860, an explicit state was added that captured number keys. This state can only be entered if the new keybinding (MARK_WITH_COUNT) is bound to a key, I recommend '#' or '+', and when it is bound you can use numpad number keys for navigation within the affected menus (mutli drop and clean). If it isn't bound, which it isn't by default, then the menu behavior stays the same where it captures all number keys and it prevents numpad navigation. I've also added a display of the current number for both number entry methods where the filter text shows up. This means the legacy method of selecting uses <enter> instead of TOGGLE_ENTRY to choose the number, I feel this is appropriate now that the user can see the numbers they are typing.

#### Describe alternatives you've considered

As far as macro implementation:
Letting one of the other Drop/Clean menu PRs handle this. Potentially migrated inventory menus to use keycodes.

Micro implementation:
Rather than introducing a new bool variable to track num capture, do something fancy and reuse count as an indicator (setting to -1 or something).

#### Testing

With MARK_WITH_COUNT unbound:

1. Spawn character with multiple of an item
2. Enter drop menu, highlight item with quantity > 4
3. Move selector with arrow keys to select the item and type "4" and verify the count entry area shows a 4
4. Press enter to note that 4 of of the item have been selected for dropping
5. Perform a similar action with the clean menu, summoning 10 filthy toothbrushes, a washing kit, and 20 soap.

With MARK_WITH_COUNT bound to "+":

1. Spawn character with multiple of an item
2. Enter drop menu, highlight item with quantity > 4
3. Move selector with numpad "2" to select the item then press "+", verify count entry shows and type 4, verify it now shows 4.
4. Press enter to note that 4 of of the item have been selected for dropping
5. Perform a similar action with the clean menu, summoning 10 filthy toothbrushes, a washing kit, and 20 soap.

PR #64664, opened 2023-03-29
inventory_ui: improve pickup header

andrei8l:
#### Summary
None

#### Purpose of change
The instructions in the pickup header blend in and are too wordy to read so new players don't realize there are helpful keys available. The header also gets garbled on the smallest screens.

#### Describe the solution
Screenshot from 2023-03-29 12-44-11

<details>
<summary>Before, for reference</summary>

Screenshot from 2023-03-29 10-49-58

</details>

Also update the default keybindings to make this work. Namely, add a key for `Mark a specific amount of selected items` and remove the double keybindings from `e`xamine as `E` has been repurposed in #62864.

Some internal changes were needed to preserve numpad navigation mode, see https://github.com/CleverRaven/Cataclysm-DDA/pull/64664#issuecomment-1488371822.

N/A

#### Testing
Other inventory UIs aren't affected

#### Additional context
N/A

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -1954,6 +1954,12 @@
   },
   {
     "type": "keybinding",
+    "id": "TOGGLE_NUMPAD_NAVIGATION",
+    "name": "Toggle navigation with numpad numbers",
+    "bindings": [ { "input_method": "keyboard_any", "key": "F5" } ]
+  },
+  {
+    "type": "keybinding",
     "id": "INVENTORY_FILTER",
     "category": "INVENTORY",
     "name": "Set item filter",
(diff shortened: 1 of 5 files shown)


mqrause:
> add a key for `Mark a specific amount of selected items`

Doesn't this change behavior because you can't just type a number anymore?

andrei8l:
> > add a key for `Mark a specific amount of selected items`
> 
> Doesn't this change behavior because you can't just type a number anymore?

Weird. Why is that a thing? I can remove that check.

andrei8l:
I've removed the exclusion so you can now both `Mark a specific amount...` and type a number then toggle. I can't find any ill effects and it was probably a very old hack. I've also re-added a bit of text about typing number.

andrei8l:
Alright, I've implemented [Kevin's  suggestion](https://github.com/CleverRaven/Cataclysm-DDA/pull/51085#issuecomment-907557439). Namely:
1. There's a `TOGGLE_NUMPAD_NAVIGATION` key and associated `uistate` that defaults to false/off
2. If numpad navigation is off, typing a number opens the string input bar so you can see what you're typing. You can close the string input bar in this case with `TOGGLE_ENTRY` so there is no functional difference
3. If numpad navigation is on, the behaviour from 2 is disabled but you can still use `MARK_WITH_COUNT`
4. `MARK_WITH_COUNT` is always available and has the current behaviour

There's also an extra hint on the right side when numpad navigation is active
Screenshot from 2023-03-29 13-28-12

PR #64670, opened 2023-03-29
Prevent the gun shooting sound from being played when player using a bow with damper.

wwkk222208:
#### Summary
None

#### Purpose of change
The game will play the same sound when the player shoots a weapon with a sound suppression accessory installed. However, it is absurd that a bow with a damper sounds like a gun. The sound of the bow should be different from that of guns.

#### Describe the solution
The sound when firing a bow with a damper installed will sound the same as bow without damper.

None.

#### Testing
Run the game and tested.

### DIFF
--- a/src/sounds.cpp
+++ b/src/sounds.cpp
@@ -1292,7 +1292,7 @@ void sfx::generate_gun_sound( const Character &source_arg, const item &firing )
         if( std::any_of( mods.begin(), mods.end(),
         []( const item * e ) {
         return e->type->gunmod->loudness < 0;
-    } ) ) {
+    } ) && firing.gun_skill().str() != "archery" ) {
             weapon_id = itype_weapon_fire_suppressed;
         }
 


Fris0uman:
I'm pretty sure soundpacks can have specific sound for bows, if the bow sounds like a gun it's probably a mistake/oversight on the soundpack side

wwkk222208:
Yes

> I'm pretty sure soundpacks can have specific sound for bows, if the bow sounds like a gun it's probably a mistake/oversight on the soundpack side

Yes, soudpachs do have specific sound for bows. But it has no way to set specific sound for a bow with damper. I checked the source code and confirmed this fact

NetSysFire:
Could be a good backport candidate I'd wager.

Raikiri:
Not sure if it's relevant, but the basic bow shooting sound is personally my favourite sound in the entire @'s soundpack. It's just so satisfying. And ended up never installing the damper because it makes the sound much less juicy.

wwkk222208:
> Not sure if it's relevant, but the basic bow shooting sound is personally my favourite sound in the entire @'s soundpack. It's just so satisfying. I ended up never installing the damper because it makes the sound much less juicy.

You may have misunderstood this PR. This PR will let you enjoy the basic bow shooting sound instead of the sound of the gun, when you shoot a bow with a damper installed.

PR #64674, opened 2023-03-29
Make elephant gun a double rifle to match reality

Terrorforge:
#### Summary
Bugfixes "Make elephant gun a double rifle to match reality"

#### Purpose of change
After a [recent Reddit post](https://www.reddit.com/r/cataclysmdda/comments/124k5yh/first_time_finding_the_gun_that_holds_the_700_nx/) reminded me that the elephant gun exists (and is kind of an interesting option now that .50 cal is much harder to get), I got curious and looked up what a .700 NX elephant gun actually looks like, and it turned out that every single example I could find was not single-shot like the in-game item, but a two-barreled double rifle.

#### Describe the solution
I'm updating the in-game weapon to be a two-shot double rifle. Curiously, the weight and other stats already seem to match. Note that these all seem to be handmade custom guns so they're all going to be slightly different, but the current stats are in the right range.

References:
[Auction listing at Rock Island Auction Company with some stats and measurements](https://www.rockislandauction.com/detail/72/2345/engraved-b-searcy-co-double-rifle-in-700-nitro-express) 
[Long-winded article with lots of pictures from Revivaler](https://revivaler.com/holland-holland-phillippe-grifnee-signed-600-and-700-nitro-express-double-rifles/) 
[History of the round and rifles from American Rifleman](https://www.americanrifleman.org/content/holland-holland-s-mighty-700-nitro-express/) 
[Short video of someone shooting one](https://www.youtube.com/watch?v=NjIpxOlWTvE) 

In addition to increasing magazine size, I made the reload speed faster. I understand that it's probably a bit slow and awkward to reload just because of the weight, but ~6 seconds seemed a bit steep, especially considering it's a break-action where you can quite quickly load two rounds at once. At 2:52 in [this video](https://www.youtube.com/watch?v=GkX7izHG0xA&t=173s), we can see the guy very unhurriedly reloading one barrel in about 7 seconds. He could certainly do it faster under time pressure, and reloading the other barrel as well wouldn't be that much slower. I don't think there's a good mechanism for making a full reload faster per cartridge than a partial reload, so I've put it at a compromise rate of 300 moves, which probably makes reloading a single barrel slightly faster than it should be, but reloading 2 slightly slower, especially considering rummaging times.

The id trex_gun suggests this was originally intended to be chambered in [.577 Tyrannosaur](https://en.wikipedia.org/wiki/.577_Tyrannosaur), so I could've reverted it to that. But it turns out the only gun that uses that uses that round, the A-Square T-REX, isn't single-shot either (it's a 3+1 internal magazine bolt action), and the changes required would essentially involve making a whole new gun and cartridge from scratch, which seems unnecessary.

#### Testing
Spawned the gun and some ammo, checked that it shoots and reloads and all that.

#### Additional context
I hemmed and hawed over whether or not I should give it the double barrel shotgun's ability to fire both barrels simultaneously. While this would technically be possible, considering [what can happen](https://www.youtube.com/watch?v=CZ-yDHL_aYI) when you do that even with significantly less powerful .500 NX, I've deemed it too impractical to include. Even ignoring the fact that it risks knocking the shooter clean off their feet, it's not clear if the gun itelf could handle it without breaking. While it might indeed be doable by a highly trained shooter or sufficiently beefy mutant, we don't have mechanics for breaking the shoulders and buttstocks of anyone else who tries, so I've elected to simply disallow it.

### DIFF
--- a/data/json/items/gun/700nx.json
+++ b/data/json/items/gun/700nx.json
@@ -5,7 +5,7 @@
     "looks_like": "ar15",
     "type": "GUN",
     "name": { "str": "Elephant gun" },
-    "description": "A custom-made single shot rifle specially designed for the hunting of huge game.  You could obviously kill everything with this, EVERYTHING.  If you ever find enough ammo of course.",
+    "description": "A ornately engraved, custom-made double rifle specially designed for the hunting of huge game.  You could obviously kill everything with this, EVERYTHING.  If you ever find enough ammo of course.",
     "weight": "8082 g",
     "volume": "4500 ml",
     "longest_side": "110 cm",
@@ -19,8 +19,8 @@
     "dispersion": 90,
     "durability": 8,
     "blackpowder_tolerance": 56,
-    "clip_size": 1,
-    "reload": 600,
+    "clip_size": 2,
+    "reload": 300,
     "barrel_volume": "1000 ml",
     "valid_mod_locations": [
       [ "accessories", 4 ],
@@ -34,7 +34,7 @@
       [ "stock accessory", 2 ],
       [ "underbarrel", 1 ]
     ],
-    "flags": [ "NEVER_JAMS", "RELOAD_EJECT" ],
-    "pocket_data": [ { "pocket_type": "MAGAZINE", "rigid": true, "ammo_restriction": { "700nx": 1 } } ]
+    "flags": [ "NEVER_JAMS", "RELOAD_EJECT", "RELOAD_ONE" ],
+    "pocket_data": [ { "pocket_type": "MAGAZINE", "rigid": true, "ammo_restriction": { "700nx": 2 } } ]
   }
 ]


CansecoGPC:
There are examples off real life single barrel chambered with .700 ammo, like these rifles:

[Hunting rifles at a convention in 2014](http://fc08.deviantart.net/fs70/f/2014/085/2/7/006_iwa2014_b_uuml_chse_700_nitro_express_8bit_108_by_it_sniper-d7bqj0q.jpg)
[More examples on a nitro express dedicated forum](http://forums.nitroexpress.com/showflat.php?Number=28469)

Maybe it would be better to add the double barrel variant instead of removing the single one.

KheaHyena:
I thought there would be a mention specifically for the NX, but I don't see it (it's just called Elephant Gun).

There's a few _Elephant Guns_ out there, as already pointed out, with bolt action and such, all single barrel (which may also include the repurposed ones for military and such). And if we really go far in time, we even have Muskets.
Would it be better to specify or add?
Or just simplify, should the desire for only one Elephant Gun to exist?

KheaHyena:
> That second rifle is a bolt action with a three-round magzine. The first does seem to be a falling block single-shot, but I can find no further information about it.
> 
> This strikes me as the exception that proves the rule. Single shot .700 guns exist, but you have to go (even further) out of your way to find one.
> 
> And perhaps more importantly, I can't find stats for one, so I'm unable to verify if the length, weight etc. is appropriate. I can't find footage of one in use, so it's hard to estimate reload time. All in all, I think it still makes more sense for the default elephant gun to be a double rifle, if for no other reason than because it's easier to compare against reality

Yeah, I understand your point, more than valid! I do agree that going for a common version, that's also easier to evaluate stats, to fit best overall.

CansecoGPC:
That single barrel is harder to find on the internet doesn't make it a exception. I just did a basic search, i'm sure specs and more examples can be found, like this one.

[Single barrel bolt action rifle with cartridge 700 NX video](https://www.youtube.com/watch?v=D41NYBHkb9M)

The same video shows a double barrel with no cartridge, just two shots like a shotgun

CansecoGPC:
Found some rifles on the photos i posted with specs

[700 NE Hambrusch Hunting Weapons](http://www.ferlachguns.com/mm/mega-700.htm)

Terrorforge:
Oh, I just realized you're saying "single-barrel" and I'm saying "single-shot". The gun as it currently exists in the game is specifically single-_shot_, and that's the variant that seems rarest and that I'm having trouble sourcing any information about. I still think the double rifle should be the standard version, but the Hambrusch bolt-action seems well-documented enough that it could be implemented.

That's a bit out of scope for this PR, though. I'm just making a small edit to make the existing elephant gun more closely match the most easily verifiable real-world version.

Terrorforge:
> Found some single shot rifles on the photos i posted with specs
> 
> [700 NE Hambrusch Hunting Weapons](http://www.ferlachguns.com/mm/mega-700.htm)

The specs at the bottom seem to be for the bolt-action. Hard to tell, though, because the layout of that page isn't great.

PR #64661, opened 2023-03-29
[Xedra Evolved] Electronic home doors

Maleclypse:
#### Summary
None

#### Purpose of change
Adds some electronically locked doors to Xedra Evolved homes that can be opened if the player guesses the code, which is handled by a contested roll.

#### Describe the solution
Adds doors and ter transform and eocs.  Adds the new doors to existing palettes.

#### Describe alternatives you've considered
Adding the new doors as variant doors for vanilla and then overwriting them to have an eoc option in Xedra Evolved.

#### Testing
Success!

#### Additional context
failure
success

PR #64678, opened 2023-03-29
Standard house variant distribution, survivor homes, doormats

FarkasKBence:
#### Summary

Content "House cleanup, survivor homes, doormats"

#### Purpose of change

Begins moving houses to an externally defined distribution of palette variants, thus increasing their variability. In the process, I have made some minor cleanup to house mapgen.
Maintains the variant palettes with stuff that are already in the general palette.

#### Describe the solution

Cleanup:
Palettized homes to utilize  `standard_domestic_lino_kitchen`, `standard_domestic_lino_bathroom`, and `standard_domestic_landscaping_palette`, moving long lists of terrain definitions outside the mapgen file.
Removed map variants 'hoarder' and 'abandoned', and replaced them with external palettes.  Using new map variants is a particularly unsustainable solution to adding variability, as they have to be individually updated and it easily leads to bloated files.
I have also removed the 'construction' map variant.  No replacements for that yet.
Standardized symbols for linoleum `'` and concrete flooring `~`.  The end goal is to have dedicated symbols for linoleum and concrete tiles.
Unified gutter terrains and removed unneeded weights.

Palettes:
Created `domestic_general_and_variant_palette`, a palette distribution that I eventually want to apply to every house.  Its purpose is to convert homes to one of three variants: hoarder -- 0.4%, abandoned -- 2.8%, and survivor -- 1.8%.
I made the hoarder palette compatible with the standard domestic palette, so no need for a new map variant.
Added new furniture, like ladders, bedside tables, and freezers to the variant palettes.
Added gate parameters to drystone half-walls, and comments to clarify palette usage.
Created the survivor palette, a variant that fortifies homes.  It also has a chance to spawn traps by the entrances. which I'm using doormats as a pretense to place.  I found it to be wasteful to use a symbol `÷` only to place traps in only one variant.

Not removing internal wall definitions.  The palette's parameters add more variation to the house but it also removes some of its uniqueness.

#### Testing

Spawned all six houses in all four variants.

#### Additional context

Future me will have to figure out how to deal with basements and nested chunks.

<details>

<summary> Screenshots: </summary>

kép

General variant:
kép

Abandoned:
kép

Hoarder:
kép

Survivor:
kép

</details>


Maleclypse:
Is there a better way for me to edit palettes than what I've been doing in Xedra Evolved? #64661 would it make more sense for me to add those door variants into the base game but only have them have an examine action in XE?

LovamKicsiGazsii:
> Is there a better way for me to edit palettes than what I've been doing in Xedra Evolved? 

Unless there's a way to expand palettes like we can with item groups or region settings, I'm not quite sure.  My next best guess is something like

`````
  {
    "type": "palette",
    "id": "standard_domestic_palette",
    "palettes": [ "standard_domestic_palette" ]
    ...
  }
`````
...but recursions are scary.

> would it make more sense for me to add those door variants into the base game but only have them have an examine action in XE?

Electronic locks are a thing in real life, so why not?  Beats having duplicate palette ids.

PR #63718, opened 2023-02-21
Add beekeeping and beekeeping acessories

MNG-cataclysm:
#### Summary
Content "Allow players to keep bees"

#### Purpose of change
Add something like #57355, and add some bee-related things.

#### Describe the solution
Add wild honey bee hives, which can be harvested for honey, wax, and their queen. Using the queen, players can create primers which are used to seed their own bee colonies in apiaries, and extract honey from them seasonally, providing a renewable source of the substance for private use and/or profit.

#### Describe alternatives you've considered
Removing the primers and requiring direct culture in a hive, not making queen bees an item, or not adding normal-sized bees. Using an EOC-based function instead of `harvest_ter`.

#### Testing
I tested everything multiple times, and it all works well.

#### Additional context
With the current, non-EOC system, honey scrapers do not serve a role in this process, as I took the "simpler" route to implementation. You also don't need beekeeping gear to extract honey from hives at the present time, although many beekeepers don't use these things.

PR #64679, opened 2023-03-29
not to bee

### LINKED ISSUE #7862: bee fields can't merge.
i added bee fields as a source of danger for the granade item to balance the unintended stat-roulette effect it had, but I don't want to include them anywhere 'serious' in the game because there was a technical issue I couldn't solve.
I couldn't figure out how to get bee fields to merge without having them move at different speeds based on which direction they were moving.
specifically, if a field merges with a field south and or east of it, the field it merges into will be processed again, and move in the same direction, but if the field moves north or west this doesn't happen.
## <bountysource-plugin>

Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/2523808-bee-fields-can-t-merge?utm_campaign=plugin&utm_content=tracker%2F146201&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F146201&utm_medium=issues&utm_source=github).
</bountysource-plugin>

kevingranade:
#### Summary
None

#### Purpose of change
Ran across an ancient issue about bee fields and realized they've likely been removed entirely at some point,  but there is special handling for them still scattered around. 
Fixes #7862
#### Describe the solution
Remove bee fields.

### DIFF
--- a/data/json/effects.json
+++ b/data/json/effects.json
@@ -595,16 +595,6 @@
   },
   {
     "type": "effect_type",
-    "id": "stung",
-    "name": [ "Stung" ],
-    "desc": [ "You have been stung!" ],
-    "rating": "bad",
-    "base_mods": { "pain_min": [ 1 ] },
-    "show_in_info": true,
-    "blood_analysis_description": "Stung"
-  },
-  {
-    "type": "effect_type",
     "id": "staggered_character",
     "name": [ "Staggered" ],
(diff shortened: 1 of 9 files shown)


MNG-cataclysm:
What if we were to re-add bee fields for #63718, if/when it's merged?

ehughsbaird:
The issues with them would need to be fixed first.

PR #64681, opened 2023-03-29
Further mutant scales refinements

anoobindisguise:
#### Summary
Balance "one more stage of Scales mutations, adjust coverage, thresholdify sleek scales"

#### Purpose of change
I was discussing the matter with @Venera3 and it was noted that Scales mutations could use some more work.

#### Describe the solution
Sparse Scales has become a lower-tier "entry level" scales that is a requirement for "regular" Scales, which have more full coverage now. Sleek Scales is post thresh Fish only, similar to how Thick Scales is post thresh Lizard only. 
In addition, I changed the Bone/Chitin dual layers to a single layer of "Mutant Scales" which is defensively in between bone and chitin.

Do two layers of mutant scales with less thickness and coverage.
Do different coverage and values.

#### Testing
Loaded up in game and they seem to be working OK. The mutations chain through rough skin->sparse scales->scales->sleek/thick scales.

#### Additional context

sparsescales
scalesscales
lizardscales
fishscales


bombasticSlacks:
Got a materials conflict now 😢

PR #64682, opened 2023-03-30
Modular vehicle toolkit part

irwiss:
#### Summary
None

#### Purpose of change

Adds modular tool station vehicle parts
2 vehicle part definitions; mounted kitchen, mounted workshop and 2 equivalent appliances
Allows attaching tools to the above, allows draining first "ammo" with tools that have magazine wells.
This in turn allows using propane cookers

#### Describe the solution

Wanted to make propane cooker working with mounted propane tanks, but then realized i'd have to do a combinatorial json of every kitchen unit type with every combo of hotplate/induction/propane/oil/etc cooker instead of the old "hotplate", every welding kit with every welder part etc... So here we are with another refactor to make most tools usable as vehicle sorta-pseudo tool draining vehicle batteries/tanks.

The end goal is replace (or augment, but likely replace would be better) the pre-made kitchen unit / foodco / kitchenmaster / metalmaster etc parts with a workbench part that you "attach" pieces like a propane cooker and it'll hook them up to the vehicle tank or batteries for when requested via `e`xamine or as crafting inventory tools

Reuses looks_like of existing kitchen_unit / welding_rig
Instead of hotkey definitions in vpart json definition it uses the `invlet` from the item for menu hotkeys.

Currently the allowed attachments are mostly what's shared among various kitchen/welder parts.
Divided into 3 "categories";

mounted kitchen - low power stuff requiring fume hood/ventilation: all kinds of heaters/cookers, chemistry set, dehydrator etc
mounted workshop - high power stuff: welder, forge, kiln
common shared by both kitchen and workshop parts: everything low power and not requiring fume hood/ventilation, or passive tools like pans/pots

Not sure what magic prevents from putting a fume hood on the workshop but that's how it currently is.

Stuff that potentially fits the PR but I cut since that'd overstep this PR:
* acetylene_cooker - no tank parts available to mount, but theoretically should be able to drain
* gasoline_cooker - the item type represents on where magazine is integrated into the stove body
* oil_cooker (kerosene cooker) - the item type represents on where magazine is integrated into the stove body
* Various tools possibly fitting on the kitchen or workshop - are left for future PRs

More stuff left for future PRs:
Replacing existing parts with the new modular ones in vehicle definitions, adding migrations for already spawned vehicles - left for future PRs as current migration code doesn't support it, and these modular ones could use testing

#### Testing

Vehicle crafting tests should catch the form_from_map changes
Some ui changes have to be tested manually

#### Additional context
image
image

### DIFF
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -4770,6 +4770,30 @@
   },
   {
     "type": "construction",
+    "id": "app_veh_tools_kitchen",
+    "group": "place_veh_tools_kitchen",
+    "category": "APPLIANCE",
+    "required_skills": [ [ "fabrication", 0 ] ],
+    "time": "5 m",
+    "components": [ [ [ "veh_tools_kitchen", 1 ] ] ],
+    "pre_special": "check_empty",
+    "post_special": "done_appliance",
+    "activity_level": "BRISK_EXERCISE"
+  },
+  {
+    "type": "construction",
+    "id": "app_veh_tools_workshop",
+    "group": "place_veh_tools_workshop",
+    "category": "APPLIANCE",
+    "required_skills": [ [ "fabrication", 0 ] ],
+    "time": "5 m",
+    "components": [ [ [ "veh_tools_workshop", 1 ] ] ],
+    "pre_special": "check_empty",
+    "post_special": "done_appliance",
+    "activity_level": "BRISK_EXERCISE"
+  },
+  {
+    "type": "construction",
     "id": "app_fridge",
     "group": "place_fridge",
     "category": "APPLIANCE",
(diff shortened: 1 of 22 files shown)


Raikiri:
Giga pog. If you can make it work, it'll make it much better now and also future-proof, but I'm not sure what to do with the way tools look. Currently vehicle welding rig looks different from a cooker for example, but will they all look like workbenches after this change?

irwiss:
not sure about tileset solution yet, they're tightly coupled to vpart id, so either uncouple or hack together a solution to uncouple it from using id directly

or have 2 separate vparts like there is now, with pure magic:tm: preventing avatar connecting soldering iron to the kitchen's hotplate outlet

Raikiri:
I wonder if swappable battery case can or should follow the same mechanism

adamkad1:
Oven doesnt work from this thing.

PR #64685, opened 2023-03-30
create avatar_dies event

KorGgenT:
#### Summary
Features "Create an event that fires when the avatar dies"

#### Purpose of change
mainly this is a feature i'm backporting from my steam version. I needed an event to fire when the avatar dies to keep track of deaths, and i thought it might possibly be useful on the repo. it might make it a little easier for me to rebase future stable versions too, if only by one commit.

#### Describe the solution
add a new event type, and fire it in game::is_game_over. note, that the event does not fire when abandoning the character.

#### Testing
image

#### Additional context
eyyy, steam version directly contributing to more stuff in-repo! go me!


src/event.cpp
```diff
@@ -140,6 +141,7 @@ DEFINE_EVENT_FIELDS( activates_artifact )
 DEFINE_EVENT_FIELDS( administers_mutagen )
 DEFINE_EVENT_FIELDS( avatar_enters_omt )
 DEFINE_EVENT_FIELDS( avatar_moves )
+DEFINE_EVENT_FIELDS( avatar_dies )
```
dseguin, line 144:
```suggestion
```
Your new event type is defined as empty, so there's no fields to define here

src/memorial_logger.cpp
```diff
@@ -482,6 +482,9 @@ void memorial_logger::notify( const cata::event &e )
                  pgettext( "memorial_female", "Angered a group of amigara horrors!" ) );
             break;
         }
+        case event_type::avatar_dies: {
+            add( pgettext( "memorial_male", "Died" ), pgettext( "memorial_female", "Died" ) );
```
dseguin, line 486:
```suggestion
            add( pgettext( "memorial_male", "Died" ), pgettext( "memorial_female", "Died" ) );
            break;
```

PR #52877, opened 2021-11-18
Better Difficulty Info

dseguin:
#### Summary
None

#### Purpose of change
Reading through #37869, I liked the suggestion of showing how selected options on character creation affect gameplay difficulty in different ways. For example: Fast Learner may be a really good trait if you're focusing on crafting, but doesn't really help as much in combat.

#### Describe the solution
Here's a demonstration:

https://user-images.githubusercontent.com/12537966/142718331-3339cd29-d448-44d9-9f5e-23183f8bbf69.mp4

This change adds information to the character creation screen to estimate different aspects of difficulty. I've added 5 areas of difficulty (based on Erk's comments in the issue and the comments below):
- Combat (interacting with enemies)
- Mobility (speed and dexterity)
- Crafting (learning and applying skills)
- Environment (how safe you are when starting out)
- Social (interacting with NPCs)

These areas are defined in JSON using the `"difficulty_impact"` type:
```json
{
  "type": "difficulty_impact",
  "//": "Difficulty impacting melee and ranged combat",
  "id": "combat",
  "name": "Combat"
  "weight": {
    "scenario": 0.2,
    "profession": 1.0,
    "hobby": 0.2,
    "mutation": 0.5,
    "skills": 0.8,
    "strength": 0.6,
    "dexterity": 0.2,
    "intelligence": 0.0,
    "perception": 0.0
  }
}
```

Each impact of difficulty is weighted differently depending on the source of the difficulty. Environmental difficulty is mostly based on the scenario, while combat difficulty is based more on profession and mutations.

The difficulty rating varies from Very Easy to Very Hard. These difficulty ratings are defined using in JSON using the `"difficulty_opt"` type:
```json
{
  "type": "difficulty_opt",
  "id": "very_hard",
  "value": 5,
  "name": "Very Hard",
  "color": "light_red"
}
```
"value" represents the difficulty relative to other difficulty ratings (higher = harder). "color" is used to highlight the difficulty text in the character creation screen.

Difficulty can be defined for scenarios, professions, hobbies, and traits/mutations. Here's a sample definition from the "A Really Bad Day" scenario:
```json
{
  "type": "scenario",
  "id": "bad_day",
  "name": "Challenge - Really Bad Day",
  "points": -8,
  "difficulty": [
    [ "combat", "very_hard" ],
    [ "mobility", "very_hard" ],
    [ "crafting", "normal" ],
    [ "environment", "very_hard" ],
    [ "social", "normal" ]
  ],
  ...
}
```
[...]

PR #64697, opened 2023-03-30
Revamped Free Start

bombasticSlacks:
#### Summary
Features "New balanced free start for character creation"

#### Purpose of change
It's been well known for a while that point totals in multipool are not balanced and haven't been for a while. It's not something that the team wants to maintain long term, and we wish to encourage a wider variety of character types without forcing players to take challenging starts or loading up on weird non-downsides. 

#### Describe the solution
Introducing Survivor Mode, free form with guidelines. 

Revived the summary info from #52877 but address the issues by making it dynamic based on an average NPC.

Created a new helper singleton object called player_difficulty. This creates a very basic NPC which the avatar can compare themselves to for various metrics.

These metrics are split into two categories:
Difficulties - How challenging things will be in game. currently summarized as “OFFENCE”, “Defence”, “SOCIAL”

Power Level - How many points you have compared to a normal character. Currently summarized as "Genetics" and "Knowledge".

So difficulties go from trivial to impossible,
Power level go from underpowered to overpowered.

These are evaluated as:

Offence is calculated as your best weapon (similar to NPCs) out of a: machete, combat knife, and baseball bat (items easy to find early) or whatever your best starting weapon is compared to a basic NPC.

Defence is a combo of armour on all locations, dodge, and the ability to run away compared with a basic NPC.

Social is the ability to convince and the opinion NPCs will have of you compared with a basic NPC.

Genetics is a judgment on how many stats and traits you have compared to an average character from old multipool.

Knowledge is a judgment on how many skills and proficiencies you have compared to the average character from old multipool with an additional consideration for any speed learning and reading stuff you have.

Finally I cleaned up some nomenclature in the menus, including moving multipool and single pool to be called "Legacy"
and some UI weirdness like how we showed skills from professions.

Also refactored getting NPC opinion of, so it can be accessed from a constant. 

#### Testing
Tested the menu quite a bit. Numbers seem good to me for now.

#### Additional context
SCREENSHOTS:
NOTE Genetics has been renamed to Lifestyle
image

image

image

image


Nebnis:
Great PR. I have trouble balancing myself out so having some comparation with ingame things might prod me into the right direction

Terrorforge:
This may be nitpicky, but looking at this as someone who's never seen it before, I don't see the point of differentiating between and using different terminology for power level and difficulty. Maybe I'm just easily confused, but when I see "Knowledge: overpowered, Offense: trivial", I don't immediately grokk that these are representing nominally different types of information, I just think "wait, is 'trivial' stronger or weaker than 'overpowered'?"

I think it makes more sense to just use the power level terminology throughout. "Offense: overpowered" conveys the same information as "Offense: trivial", right?

TheShadowFerret:
Does having levels in added new skills(spellcraft/deducation/whatever third parties add) increase, dilute(thus lower), or ignore Knowledge?

bombasticSlacks:
> Does having levels in added new skills(spellcraft/deduction/whatever third parties add) increase or ignore Knowledge?

if non-combat they should yeah.

Raikiri:
I agree that adjectives "easy" and "trivial" simply don't land on the same scale that "underpowered" and "overpowered" do. The gradations should be either power levels "Feeble/Weak/Strong/Overpowered", capability compared to a normal baseline "Below average/Above average" or on gameplay difficulty "Trivial/Easy/Hard/Very Hard/Insane", however, you probably should not mix them.

pjf:
Can we use a different name than 'genetics'? Going to the gym (high strength), being a smoker, having the blob give you psychosis, or being spiritual don't really seem to fit the 'genetics' theme (although obviously some things do).

Perhaps 'background' or 'lifestyle'? I'm not attached to any term in particular, but 'genetics' doesn't feel right to me.

Terrorforge:
> The reason was that Trivial gives more of the feeling of "you are making the game too easy for yourself", whereas something like "powerful" could make a new player go "hell yeah I _am_ making a powerful character 😎"
> 
> With the swap to overpowered as the top word it might be okay, though.

Is that a bad thing? New players have it hard enough as it is, I don't actually think we need to discourage them from making strong characters. I think it would be difficult for a new player to make something that's truly so strong the game stops being fun and/or hard, anyway.

But if you do keep the current language, they need to be more clearly differentiated in the ui. Right now it looks and feels like they're using different words for the same thing, and it's confusing.

PatrikLundell:
Looks like a very ambitious task!

I'd suggest replacing words with numbers with a fairly large range (0-100 or 0-1000), as words are rather fluffy and easily misinterpreted (does "average" mean the mathematical average or "rather average" = fairly crappy? Is fluffy word X better or worse than fluffy word Y?). After all, there would have to be numbers as the basis for the words of it anyway. 

Also, it should be made clear where the game's balance point lies, as there is a lot more space for additions above the balance point than below it (if stats are balanced around 38 points, you can't go lower than to 4, but at least up to 80, if 20 is the maximum number, and it seems the balance point for skills is at 0, which is also the minimum).

There's also a need for a clear explanation that the comparison is for a starting character, rather than something reflecting the game difficulty. The last screen shot (at the time of this writing) has a character with a skill of 1 in 3 ranged skills described with offense and defense as "trivial" (based on starting equipment?) which a new player probably would find very misleading.

I agree with @pjf that "genetics" does a poor job of conveying what the category is. I'd suggest something like "properties" (unless the group is split up).

The last sentence of the screen displaying the Legacy Multiple pools start (at the time of this writing) ought to either be split into two sentences or separated into two parts.

Nebnis:
If i may throw my own idea. I would name it "Very weak, weak, average, strong, very strong".

PR #59884, opened 2022-08-04
REMOVE SKILL RUST FROM GAME OPTIONS

sonphantrung:
#### Summary
Infrastructure "REMOVE SKILL RUST FROM GAME OPTIONS"

#### Purpose of change
Resurrect #55173
Original desc:
Since skill rust was transformed into a useful and important gameplay mechanic in https://github.com/CleverRaven/Cataclysm-DDA/pull/50176, there is no good reason to keep it in the game debug options.

#### Describe the solution
Basically the same thing as the former repo (can't playtest rn, machine has no good capability of compilation)

#### Describe alternatives you've considered
NONE

#### Testing
The code is practically the same, so test results should be the same

#### Additional context
- Doing git apply/ git am is a PAINFUL process, so I basically just copied the same code ;)

PR #64708, opened 2023-03-30
remove skill rust option

kevingranade:
#### Summary
None

#### Purpose of change
As outlined in #59884 and elsewhere, the skill rust options are no longer necessary.  Skill rust no longer has the issues that made an option necessary. 

#### Describe the solution
Remove the option and the code handling it.

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -196,7 +196,7 @@ static const std::map<std::string, std::pair<std::string, std::map<std::string, 
 &get_migrated_options()
 {
     static const std::map<std::string, std::pair<std::string, std::map<std::string, std::string>>> opt
-    = { {"DELETE_WORLD", { "WORLD_END", { {"no", "keep" }, {"yes", "delete"} } } }
+    = { {"DELETE_WORLD", { "WORLD_END", { {"no", "keep" }, {"yes", "delete"} } } } };
     return opt;
 }
 
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -196,10 +196,7 @@ static const std::map<std::string, std::pair<std::string, std::map<std::string, 
 &get_migrated_options()
 {
     static const std::map<std::string, std::pair<std::string, std::map<std::string, std::string>>> opt
-    = {
-        {"DELETE_WORLD", { "WORLD_END", { {"no", "keep" }, {"yes", "delete"} } } },
-        {"SKILL_RUST", { "SKILL_RUST", { {"int", "vanilla" }, {"intcap", "capped"} } } }
-    };
+    = { {"DELETE_WORLD", { "WORLD_END", { {"no", "keep" }, {"yes", "delete"} } } }
     return opt;
 }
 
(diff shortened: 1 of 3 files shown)


src/options.cpp
```diff
@@ -196,10 +196,7 @@ static const std::map<std::string, std::pair<std::string, std::map<std::string,
 &get_migrated_options()
 {
     static const std::map<std::string, std::pair<std::string, std::map<std::string, std::string>>> opt
-    = {
-        {"DELETE_WORLD", { "WORLD_END", { {"no", "keep" }, {"yes", "delete"} } } },
-        {"SKILL_RUST", { "SKILL_RUST", { {"int", "vanilla" }, {"intcap", "capped"} } } }
-    };
+    = { {"DELETE_WORLD", { "WORLD_END", { {"no", "keep" }, {"yes", "delete"} } } }
```
kevingranade, line 199:
```suggestion
    = { {"DELETE_WORLD", { "WORLD_END", { {"no", "keep" }, {"yes", "delete"} } } } };
```

Maleclypse:
Would you want to remove the practical level loss notification as well while you are here?  Because the notification is annoying when everything else is on a percentage scaling impact.  I'm remembering correct that the changes are on a sliding scale and that there isn't a drastic difference in play between  level 1 and 99% and level 2?

cdru123:
> Would you want to remove the practical level loss notification as well while you are here? Because the notification is annoying when everything else is on a percentage scaling impact. I'm remembering correct that the changes are on a sliding scale and that there isn't a drastic difference in play between level 1 and 99% and level 2?

IIRC, a few things don't check for your fractional skill value, with claims that Dodge is impacted. In the same way, martial arts techniques have a minimum skill level, and thus get impacted by non-capped skill rust

sonphantrung:
There is also another problem:
- In StatsThroughSkills, skill drop can risk stat drop

Maleclypse:
> There is also another problem:
> 
> * In StatsThroughSkills, skill drop can risk stat drop

That's not a problem for this PR.  Whoever manages StatsThroughSkills has chosen for stats to be correlated to practical and not theoretical. And that choice isn't impacted in any way by this PR.  If you think it should be different I would make a PR altering the StatsThroughSkills choice but it has no bearing on skill rust itself.

Maleclypse:
> > Would you want to remove the practical level loss notification as well while you are here? Because the notification is annoying when everything else is on a percentage scaling impact. I'm remembering correct that the changes are on a sliding scale and that there isn't a drastic difference in play between level 1 and 99% and level 2?
> 
> IIRC, a few things don't check for your fractional skill value, with claims that Dodge is impacted. In the same way, martial arts techniques have a minimum skill level, and thus get impacted by non-capped skill rust

I tested Computers last night extensively and it goes by fractional, so you would need proof in code that dodge is impacted or that martial arts are pulling against practical rather than theoretical knowledge.

PR #64717, opened 2023-03-31
Use minimum spanning tree for road connections

sparr:
#### Summary
Bugfixes "Eliminate unnecessary long road connections"

#### Purpose of change
The greedy algorithm used to connect points on an overmap with roads (or other types of connections) occasionally requests connections that are unnecessarily far apart.

#### Describe the solution
Replace the greedy algorithm with a minimum spanning tree algorithm.

A greedy algorithm starting with the most isolated point would usually do better than the current algorithm, and be simpler than the one in this PR.

#### Testing
Generated a few dozen overmaps and visually confirmed connections between cities and overmap edges and across overmaps, while also running #64716

#### Additional context
Uses [Kruskal's algorithm](https://en.wikipedia.org/wiki/Kruskal%27s_algorithm)

### DIFF
--- a/src/overmap.cpp
+++ b/src/overmap.cpp
@@ -4868,62 +4868,50 @@ void overmap::place_roads( const overmap *north, const overmap *east, const over
 
     // At least 3 exit points, to guarantee road continuity across overmaps
     if( roads_out.size() < 3 ) {
-        std::vector<tripoint_om_omt> viable_roads;
-        tripoint_om_omt tmp;
-        // Populate viable_roads with one point for each neighborless side.
-        // Make sure these points don't conflict with rivers.
 
+        std::array<const overmap *, 4> neighbors = { east, south, west, north };
+        static constexpr std::array<point, 4> neighbor_deltas = {
+            point_east, point_south, point_west, point_north
+        };
+
+        // x and y coordinates for a point on the edge in each direction
+        // -1 represents a variable one dimensional coordinate along that edge
+        // east == point( OMAPX - 1, n ); north == point( n, 0 );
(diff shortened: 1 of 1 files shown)


anothersimulacrum:
Could you showcase what road generation looks like before and after this?

ZhilkinSerg:
before:
image

after:
image

sparr:
Before:
cataclysm_overmap_roads_before_4
After:
cataclysm_overmap_roads_after_4

sparr:
Before:
cataclysm_overmap_roads_before_3
After:
cataclysm_overmap_roads_after_3

sparr:
Before:
cataclysm_overmap_roads_before_2
After:
cataclysm_overmap_roads_after_2

sparr:
This one was unchanged. I suspect that, by chance, the greedy algorithm produced a similar enough graph to the minimum spanning tree that the pathfinder ended up choosing the same paths.
cataclysm_overmap_roads_before_1

src/overmap.cpp
```diff
@@ -5752,24 +5740,77 @@ void overmap::build_connection(
...
+                if( subgraphs[k] == dead_subtree ) {
+                    subgraphs[k] = subgraphs[i];
+                }
             }
+            // and connect them
+            connect = true;
+        } else if( one_in( 10 ) ) {
+            // the remaining case is that i and j are already in the same subgraph
+            // making this connection creates a loop
+            connect = true;
```
sparr, line 5809: This is new behavior. It feels reasonable to me to have occasional loops rather than the roads making a perfect tree (ignoring cities).

PR #64718, opened 2023-03-31
[CR] NPC spawn with reasonable stored_kcal

SurFlurer:
#### Summary
Bugfixes "NPC spawn with reasonable stored_kcal"

#### Purpose of change
When spawned, NPC's stored calories are hard set to 54'000'000 cal. With the new BMI system, this means that a majority of them are underweight.

#### Describe the solution
When randomizing npcs, set their stored_kcal to a time-related level. I also moved this part of code to before they mutate, so it wouldn't neutralize body-size mutation ( if there's any).

The details are:

I introduced a chi_squared_distribution roll to `rng.cpp` and `rng.h`. The distribution gradually transit from χ^2(6) to χ^2(3), in the period of 180 days.

`weight_percent` is the relative percentage against BMI 5.0 (healthy). So `weight_percent = 1.5` means that you have a BMI of 7.5.

At day 0, with the distribution of χ^2(6), 81% of NPCs have BMI more than 3.0 (aka "Normal") , 54% of them have BMI more than 5.0 (aka "Overweight"), and 12% of them have BMI more than 10.0 ( aka "Obese" ).

After day 180, with the distribution of χ^2(3), 57% of them have BMI more than 2.0 (aka "Underweight"), 39% of NPCs have BMI more than 3.0 (aka "Normal"), 17% of them have BMI more than 5.0 (aka "Overweight").

I also clamped the weight_percent from 0.2 to 5.0, making it fall between XS and XXXL size.

Use an alternative distribution.

#### Testing
Spawned many npcs at different time_points, confirmed that their stored_kcal fall in the right ballpark.

### DIFF
--- a/src/npc.cpp
+++ b/src/npc.cpp
@@ -838,6 +838,8 @@ void npc::randomize( const npc_class_id &type )
 
     set_body();
     recalc_hp();
+    randomize_height();
+    set_stored_kcal( get_healthy_kcal() );
 
     starting_weapon( myclass );
     starting_clothes( *this, myclass, male );
@@ -882,8 +884,6 @@ void npc::randomize( const npc_class_id &type )
     }
 
     set_base_age( rng( 18, 55 ) );
-    randomize_height();
-    set_stored_kcal( get_healthy_kcal() );
 
     // Add eocs
     effect_on_conditions::load_new_character( *this );
--- a/src/npc.cpp
+++ b/src/npc.cpp
@@ -745,7 +745,7 @@ void npc::load_npc_template( const string_id<npc_template> &ident )
 }
 
 npc::~npc() = default;
-#pragma optimize ("",off)
+
 void npc::randomize( const npc_class_id &type )
 {
     if( !getID().is_valid() ) {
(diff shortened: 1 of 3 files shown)


andrei8l:
It should be randomized with the distribution leaning towards underweight as time passes.

irwiss:
Tbh the time distribution seems weird, the avatar experiences no food shortages at all but NPCs will be starving after just 6 months by pure magic

Feels like if there's no specific backstory unique to particular NPC this shouldn't be the case

SurFlurer:
> Tbh the time distribution seems weird, the avatar experiences no food shortages at all but NPCs will be starving after just 6 months by pure magic
> 
> Feels like if there's no specific backstory unique to particular NPC this shouldn't be the case

Yeah I'm also faintly feeling that this process should be longer. Maybe we can turn on this function only if No_NPC_Need is activated?

bombasticSlacks:
Alright this is a step in the right direction, gonna merge it and it can be tweaked in the future but languishing on the back pages of git isn't doing any favours.

PR #64736, opened 2023-03-31
Remove stale warning about an older ccache version for visual studio.

akrieger:
#### Summary
None

#### Purpose of change
Stale warnings are more confusing than they are worth. ccache is up to version 4.8.x, this warning about one point rev of 4.6 needs to be taken to the farm upstate.

#### Describe the solution
Take the warning to the farm upstate.

#### Testing
👀


BrettDong:
I tried 4.8 on my computer last week, another new bug was introduced in 4.8 that makes ccache reject to cache anything: https://github.com/ccache/ccache/issues/1262

PR #64750, opened 2023-04-01
Enhanced mouse support for inventory screens

katemonster33:
#### Summary
Features "Added sorting, showing&hiding pocket contents, and scrolling via mouse in inventory screens"

#### Purpose of change

Per #58106 I am adding full mouse support to the inventory screens. My other PR adds mouse-hover support, this one adds using the expander button next to pockets to show/hide contents, as well as scrolling the screens using the scroll wheel. I also saw a suggestion for adding drag-drop support to the screen (inventory_pick_selector only) and felt it was a worthy addition.

#### Describe the solution

For clicking the expander, the GUI simply checks if the user has clicked the character just left of a selectable item, and if they have, the flow continues as if the user had pressed the '>' key.
For drag-drop, I registered a handler for CLICK_AND_DRAG for when the user presses left mouse down, and it marks what object was selected when left-click is pressed, and what item is selected when it released. If they are not the same item, and the destination is a pocket, the code attempts to place the object inside the pocket, throwing the same errors that the sorting view shows when it is not possible.
For mouse scrolling, I registered SCROLL_UP and SCROLL_DOWN actions and added a new function for handling scrolling which always goes to the next page in the inventory if there is one. This works in tandem with drag-drop when the user clicks an item on one page and drags it to a pocket on another.

#### Testing

I opened the inventory drop dialog, as well as the inventory list ('i') and the inventory compare view ('I') to verify that they are working as expected.

### DIFF
--- a/src/inventory_ui.cpp
+++ b/src/inventory_ui.cpp
@@ -3513,7 +3513,7 @@ std::pair<const item *, const item *> inventory_compare_selector::execute()
 
         if( input.entry != nullptr ) {
             highlight( input.entry->any_item() );
-            if( input.action == "SELECT" ) {
+            if( input.action == "SELECT" || input.action == "ANY_INPUT" ) {
                 toggle_entry( input.entry );
                 just_selected = input.entry;
             } else if( input.action != "MOUSE_MOVE" ) {
@@ -3673,7 +3673,7 @@ void inventory_multiselector::on_input( const inventory_input &input )
 {
     if( input.entry != nullptr ) { // Single Item from mouse
         highlight( input.entry->any_item() );
-        if( input.action == "SELECT" ) {
+        if( input.action == "SELECT" || input.action == "ANY_INPUT" ) {
             toggle_entries( count );
         }
     }
--- a/src/inventory_ui.cpp
+++ b/src/inventory_ui.cpp
@@ -595,8 +595,7 @@ const item_category *inventory_entry::get_category_ptr() const
     if( !is_item() ) {
         return nullptr;
     }
-    custom_category = &any_item()->get_category_of_contents();
-    return custom_category;
+    return &any_item()->get_category_of_contents();
 }
 
 bool inventory_column::activatable() const
@@ -919,6 +918,20 @@ void inventory_column::move_selection_page( scroll_direction dir )
     highlight( index, dir );
 }
 
+void inventory_column::scroll_selection_page( scroll_direction dir )
+{
+    size_t index = highlighted_index;
+
+    while( page_of( index = next_highlightable_index( index, dir ) ) == page_index() ) {
+
+        if( index == highlighted_index && page_of( index ) == page_index() ) {
+            break; // If flipped and still on the same page - no need to flip
+        }
+    }
+
+    highlight( index, dir );
+}
+
 size_t inventory_column::get_entry_cell_width( const inventory_entry &entry,
         size_t cell_index ) const
 {
@@ -1196,7 +1209,6 @@ void inventory_column::set_collapsed( inventory_entry &entry, const bool collaps
 
 void inventory_column::on_input( const inventory_input &input )
 {
-
     if( !empty() && active ) {
         if( input.action == "DOWN" ) {
             move_selection( scroll_direction::FORWARD );
@@ -1206,6 +1218,10 @@ void inventory_column::on_input( const inventory_input &input )
             move_selection_page( scroll_direction::FORWARD );
         } else if( input.action == "PAGE_UP" ) {
             move_selection_page( scroll_direction::BACKWARD );
+        } else if( input.action == "SCROLL_DOWN" ) {
+            scroll_selection_page( scroll_direction::FORWARD );
+        } else if( input.action == "SCROLL_UP" ) {
+            scroll_selection_page( scroll_direction::BACKWARD );
         } else if( input.action == "HOME" ) {
             highlight( 0, scroll_direction::FORWARD );
         } else if( input.action == "END" ) {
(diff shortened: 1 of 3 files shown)


src/inventory_ui.cpp
```diff
@@ -2763,6 +2811,15 @@ inventory_input inventory_selector::process_input( const std::string &action, in
...
                 }
+                if( res.entry == nullptr && res.action == "SELECT" ) {
+                    p.x++;
+                    res.entry = find_entry_by_coordinate( p );
+                    // if the user has clicked the coordinate just to the left of a pocket, they are clicking the pocket's expander icon
+                    if( res.entry != nullptr && res.entry->is_selectable() && res.entry->chevron ) {
+                        res.action = "SHOW_HIDE_CONTENTS";
+                        return res;
+                    }
+                }
```
andrei8l, line 2822: This won't work for `inventory_multiselector` https://github.com/CleverRaven/Cataclysm-DDA/blob/6ef8d469bf1d2e0f8e729bd372973d86fad53342/src/inventory_ui.cpp#L3553-L3556

Either add another special case there or move this code to `inventory_column::on_input()`

src/inventory_ui.cpp
```diff
@@ -3088,15 +3145,44 @@ item_location inventory_pick_selector::execute()
...
+                dragActive = false;
+                item *startDragItemCpy = startDragItem;
+                startDragItem = nullptr;
+                item *endDragItem = nullptr;
+                if( input.entry->any_item() != item_location::nowhere ) {
+                    endDragItem = input.entry->locations.front().get_item();
+                    if( startDragItemCpy != nullptr && endDragItem != nullptr && startDragItemCpy != endDragItem ) {
+                        if( drag_drop_item( startDragItemCpy, endDragItem ) ) {
+                            clear_items();
+                            add_character_items( u );
```
andrei8l, line 3174: You're assuming here that the pick selector was initialized with the character's inventory. Quick search shows that it's also used for reading ebooks, looking inside containers, and picking nearby CBMs for autodoc.

Atm I don't have a good solution to suggest. Typically we solve this by assigning a menu return activity.

src/inventory_ui.cpp
```diff
@@ -1894,6 +1911,34 @@ bool inventory_selector::add_entry_rec( inventory_column &entry_column,
...
+    auto pockets = destItem->get_all_contained_pockets();
+    if( pockets.empty() ) {
+        popup( "Destination has no pockets." );
+        return false;
+    }
+    for( auto pocket : pockets ) {
+        ret_val<item_pocket::contain_code> contain = pocket->can_contain( *sourceItem );
+        if( contain.success() ) {
+            // storage should mimick character inserting
+            get_avatar().as_character()->store( pocket, *sourceItem );
```
Qrox, line 1929: Does this handle the movement cost of moving items between pockets?

src/inventory_ui.cpp
katemonster33: maybe we can add some sort of flag to enable drag-drop so that it only works when the selector is populated with the character's inventory.

src/inventory_ui.cpp
katemonster33: I like the idea of moving the code into inventory_column::on_input, however find_entry_by_coordinate is a member of inventory_selector. We would need to add more members to inventory_input so that subclasses could handle this. If you find this approach preferable, I will go this route.

src/inventory_ui.cpp
katemonster33: So I copied this code from pocket_favorite_callback::move_item, the function used by the pocket organizer menu. This usage appears to hard-code a movement cost of 100. It looks like I could be calling item::insert_cost to handle this better, so if you think it's OK to change this usage in both places in this PR, I will do so.

src/inventory_ui.cpp
```diff
@@ -1894,6 +1911,34 @@ bool inventory_selector::add_entry_rec( inventory_column &entry_column,
     return vis_contents;
 }
 
+bool inventory_selector::drag_drop_item( item *sourceItem, item *destItem )
+{
+    if( sourceItem == destItem ) {
+        popup( "Can't move an item onto the same item!" );
```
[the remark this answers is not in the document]
katemonster33: Sounds good, committed.

src/inventory_ui.cpp
```diff
@@ -3088,15 +3145,44 @@ item_location inventory_pick_selector::execute()
 {
     shared_ptr_fast<ui_adaptor> ui = create_or_get_ui_adaptor();
     debug_print_timer( tp_start );
+    item *startDragItem = nullptr;
```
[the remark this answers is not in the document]
katemonster33: OK, done.

src/inventory_ui.cpp
Qrox: It would be nice to unify the cost, but before that I'm also concerned that it may not be using an activity to move the items from the sound of it, and fixing that would require more changes and is better done in a separate PR. If you move a large amount of items in these UIs and exit from them, do you have to wait for additional turns and can you interrupt the waiting?

src/inventory_ui.cpp
andrei8l: AFAICT you're solving this one in the hover PR and it only needs a tiny change there.

src/inventory_ui.cpp
katemonster33: Unfortunately these questions are above my current level of knowledge on how the game works. Maybe a separate issue to address both use cases would be best?

mqrause: I didn't get to actually test it, yet, but from the sounds of it I don't think this should go in like this without being an actual activity. Worst case you're dragging around a bunch of items in the menu trying to figure out some ideal layout and it stacks up moves that cause you to lose turns without any clear indication why.

Either check out how wield/wear in the pickup menu does it (close menu, do activity, reopen menu) or implement this as a sort of planning tool that lists all the movement actions you want to perform with the moves it will take that requires confirmation before it's executed (as an activity so it's interruptible). Inventory management in this game is neither meant to be free nor to lock your character from reacting to things happening around you.

katemonster33: What about showing a dialog upon releasing the left-click button that makes the user clicks the pocket the item would be going into that shows the movement cost? This way we would also handle things like backpacks with multiple pockets, the user could choose which pocket the item goes into while taking into consideration the cost.

mqrause: I currently have no real opinion on the ui side of the implementation. Just that the ui should communicate what's happening and that whatever is happening should be an activity.

src/inventory_ui.cpp
katemonster33: after merging in master with the mouse-hover branch changes, this appears to be fixed, resolving.

src/inventory_ui.cpp
katemonster33: I added a boolean in inventory_pick_selector for enabling drag-drop support, only set to true for the main 'i' inventory screen.

src/inventory_ui.cpp
katemonster33: I'm not sure what you mean by activity. As I said above the number of moves is currently hard-coded as 100 and not shown to the user. I will display the proper value. But the way it works here is the same way it works in the Inventory Organization screen. If we need to change the work-flow of how moving items between pockets is communicated to the user, I believe it would be more appropriate to create a new issue after this PR is merged to fix it here and in the Inventory Organization screen.

katemonster33:
Added a dialog which displays during a drag-drop event that prompts the user to select the pocket to place the item into. Currently it shows all pockets as selectable and throws errors when a pocket can't contain the item. Maybe this is too cumbersome. Let me know what other features are required for this to be acceptable.

src/inventory_ui.cpp
```diff
@@ -1894,6 +1911,49 @@ bool inventory_selector::add_entry_rec( inventory_column &entry_column,
...
+    for( item *parent : parentContainers ) {
+        base_move_cost += parent->obtain_cost( *sourceItem );
+    }
+    std::vector<item_pocket *> containable_pockets;
+    uilist action_menu;
+    action_menu.text = "Select pocket";
+    int index = 0;
+    for( item_pocket *pocket : pockets ) {
+        // NOLINTNEXTLINE(cata-translate-string-literal)
+        std::string pocket_text = string_format( "%s - %s/%s | %d moves",
```
Qrox, line 1935:
```suggestion
        std::string pocket_text = string_format( _( "%s - %s/%s | %d moves" ),
```

src/inventory_ui.cpp
```diff
@@ -1894,6 +1911,49 @@ bool inventory_selector::add_entry_rec( inventory_column &entry_column,
...
+        return false;
+    }
+    int base_move_cost = 0;
+    auto parentContainers = u.parents( *sourceItem );
+    for( item *parent : parentContainers ) {
+        base_move_cost += parent->obtain_cost( *sourceItem );
+    }
+    std::vector<item_pocket *> containable_pockets;
+    uilist action_menu;
+    action_menu.text = "Select pocket";
```
Qrox, line 1931:
```suggestion
    action_menu.text = _( "Select pocket" );
```

src/inventory_ui.cpp
```diff
@@ -1894,6 +1911,49 @@ bool inventory_selector::add_entry_rec( inventory_column &entry_column,
...
+    int index = 0;
+    for( item_pocket *pocket : pockets ) {
+        // NOLINTNEXTLINE(cata-translate-string-literal)
+        std::string pocket_text = string_format( "%s - %s/%s | %d moves",
+                                  pocket->get_description().translated(), vol_to_info( "", "", pocket->contains_volume() ).sValue,
+                                  vol_to_info( "", "", pocket->max_contains_volume() ).sValue, pocket->moves() + base_move_cost );
+        action_menu.addentry( index++, true, std::optional<input_event>(), pocket_text );
+    }
+    action_menu.query();
+    if( action_menu.ret >= 0 && action_menu.ret < int( pockets.size() ) ) {
```
Qrox, line 1941:
```suggestion
    if( action_menu.ret >= 0 && static_cast<size_t>( action_menu.ret ) < pockets.size() ) {
```

katemonster33:
Going to mark as draft for a bit so I can squash all the trivial commits

PR #64759, opened 2023-04-01
Replace count_*_flag with has_*_flag

sparr:
#### Summary
None

#### Purpose of change
Every current use of the flag counting functions is in a boolean context. This means we're wasting computation after the first matching flag is found.

#### Describe the solution
Rename `has_flag` to `count_flag`, make a new `has_flag` and `has_*_flag` alternatives to the `count_*_flag` functions, and use them in all the places the counts were previously used.

The count functions are now unused and could be removed?

#### Testing
Ran test suite. Played the game for a few minutes.

#### Additional context
This came up while investigating a performance regression in #62092 when using gcc, which I'll be making further PRs regarding soon.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -1735,10 +1735,10 @@ bool Character::uncanny_dodge()
 
     const bool can_dodge_bio = get_power_level() >= trigger_cost &&
                                has_active_bionic( bio_uncanny_dodge );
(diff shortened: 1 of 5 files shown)


Venera3:
Counts will be necessary for conditional limb flags - the limb project got shifted to this cycle so it's a bit of a stump, but the functionality should stick around.

andrei8l:
Some profiling that shows the perf improvement would be nice

PR #64760, opened 2023-04-01
(Perks mod) Create the trouble seeker perk

SariusSkelrets:
#### Summary
Mods "Add a new perk for the perk mod"

#### Purpose of change

Add a perk that will regularly spawn a zombie in the reality bubble when activated.

#### Describe the solution

Add that perk that spawns a zombie every ~4-7 hours in the reality bubble at a distance of 35-50 tiles from the player, for those who want more zombies to kill, loot and butcher.

Creating another perk

#### Testing

Made a new world.  Got the perk through experience.  Waited until the spawn. Worked as intended.

#### Additional context

Monster groups might work better than only spawning the regular zombie but only spawning one zombie reduces the loot range that can be obtained, facilitating balance.


GuardianDll:
Won't it just annoy you when you try to do some stuff, and will add almost nothing while active combat?

SariusSkelrets:
> Won't it just annoy you when you try to do some stuff, and will add almost nothing while active combat?

You can toggle the perk on and off with the mutation menu. It's opt-in and you can opt out when you want. I thought about adding that in the description but the recycler perk doesn't mention activation in its description so I followed its writting style. 
I can add it if needed

The main goal of the perk is to add a way to get zombie loot even if you cleared the whole area. Clothes, some materials, possibility of rare drops, tainted meat and bones, a small amount of xp, ect. It saves time when you're mass producing stuff such as mutagen in a safe spot as the main ingredient comes to you

The zombie will also fight non-zombies as expected so that's another niche use

The perk isn't powerful, but it got its uses which is the goal of perks

bombasticSlacks:
I like this idea, gotta take time to test and make sure it works. If it's based on the recycler perk it currently doesn't all work 😆

SariusSkelrets:
I'm just bringing it up to date, not closing it

bombasticSlacks:
Hey so feel free to ping me when you reopen, I need to find some time to rewrite the recycler stuff, then I can clean up the EOCs in this to follow a similar format. The problem with the current setup is you can queue it multiple times and stuff causing weirdness. 

Ive been very busy with a few other large projects and haven't had much time for the perk mod.

SariusSkelrets:
Ok I'll check that when I'll update my branch

Got to account for the new perks

PR #64773, opened 2023-04-02
Standarized Makaylas

John-Candlebury:
#### Summary
Content "Standarize Makayla's gear, skills and traits"

#### Purpose of change

I want to make Makayla spawn in different locations using EOCs, this doesn't really work very well if she's completely different every time you meet her.  Also making  NPCs  have better defined personalities usually works for the better, and this is harder if they are random.

#### Describe the solution
Gave Makayla a definite Inventory, skills and trait lists. Took the time to explain mechanically why she has mutated ears from day one too.

Unfortunately some things cannot still be set definitively, namely stats. And it doesn't seem like the game actually respects the Age and height settings included here.

At least these are relatively minor differences most players aren't likely to notice.

#### Testing
Spawned the center to no errors, swapped into Makayla to check the trait worked.

#### Additional context

I'll make a bug report for the Age and height settings being ignored with this PR as an example.

### DIFF
--- a/data/json/effect_on_condition.json
+++ b/data/json/effect_on_condition.json
@@ -385,6 +385,39 @@
   },
   {
     "type": "effect_on_condition",
+    "id": "MAKAYLA_MUTATOR_portal_storm",
+    "recurrence": "15 minutes",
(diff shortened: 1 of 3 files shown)


Fris0uman:
When the unfathomable eldritch weather decides you'll have cat ears and that's final. X)

PR #64774, opened 2023-04-02
Make makeshift bandages not require tailoring

KittyTac:
#### Summary
Balance "Make makeshift bandages not require tailoring, making them easier to craft at 0 tailoring"

#### Purpose of change
Makeshift bandages trained tailoring, which made no sense as it's basically looking at a cotton sheet and saying "this is a bandage". This also made them really hard to craft without skill which makes no sense. It doesn't even require cutting or anything.

#### Describe the solution
Remove the skill requirement, also reduce time to 1 second.

Somehow making this and some other crafts (like soap flakes) impossible to fail, IDK how though.

#### Testing
изображение

#### Additional context
N/A.

### DIFF
--- a/data/json/recipes/other/medical.json
+++ b/data/json/recipes/other/medical.json
@@ -28,10 +28,10 @@
     "activity_level": "NO_EXERCISE",
     "category": "CC_*",
     "subcategory": "CSC_*_NESTED",
-    "skill_used": "tailor",
-    "time": "9 s",
+    "time": "1 s",
     "autolearn": true,
     "flags": [ "BLIND_EASY" ],
+    "//": "Too simple to learn any tailoring or healthcare.",
     "components": [ [ [ "sheet_cotton", 1 ], [ "sheet_cotton_patchwork", 1 ] ] ]
   },
   {


PatrikLundell:
I agree with the logic. However, have you checked if there are sufficient remaining recipes that train tailoring? Recipes with reasonably available resources are needed both for illiterate and Innawood characters.

KittyTac:
> I agree with the logic. However, have you checked if there are sufficient remaining recipes that train tailoring? Recipes with reasonably available resources are needed both for illiterate and Innawood characters (as well as those who won't find appropriate books for whatever reason).

Some primitive clothing/basketweaving stuff trains tailoring so it's fine.

Fris0uman:
Why does it still display failure chance in your screenshot?

KittyTac:
> Why does it still display failure chance in your screenshot?

I don't know how to make it have no failure chance at all. Even soap flakes have a failure chance.
image

Note that on the screenshot I'm playing a cat mutant with paws so that adds some chance, but even a normal human will have a failure chance. If you know how to remove that, tell me please.

Issue #64771, opened 2023-04-02
NPC invited me on a quest, then stole all my stuff


nornagon:
### Describe the bug

I got the MISSION_GET_SAFE_BOX mission from my starting NPC, and once I closed the dialog, the NPC walked over to my stash of items (which was in the evac shelter) and started picking things up and eating my food.

### Attach save file

[Chanhassen-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11131117/Chanhassen-trimmed.tar.gz)

### Steps to reproduce

I haven't tried to reproduce this on a fresh character, but I'm guessing it would happen with this mission reliably?

### Expected behavior

The NPC shouldn't take my stuff without asking! And shouldn't eat when no_npc_food is loaded.

### Screenshots

![Screen Shot 2023-04-01 at 8 53 37 PM](https://user-images.githubusercontent.com/172800/229330397-a97dadb0-f1a9-4003-a075-e6d882bb3147.png)

![Screen Shot 2023-04-01 at 8 58 17 PM](https://user-images.githubusercontent.com/172800/229330505-8b9e3aa4-62cf-4f35-8c3e-9a7e47fa3f90.png)

### Versions and configuration

- OS: MacOs
    - OS Version: Darwin wilson.local 21.6.0 Darwin Kernel Version 21.6.0: Mon Dec 19 20:44:01 PST 2022; root:xnu-8020.240.18~2/RELEASE_X86_64 x86_64
- Game Version: 0.G-986-ga33c958e6f [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    sees-player, run, crouch, bleed icons, retrodays [sees_player_retro]
]

nornagon:
Found this in "NPC vars":

```
Character Name: Percy Sorenson
|;key;value;
|;was_sleeping;false;
|;sleep_health_mult;1;
|;quarter_stam_counter;0;
|;THIEF_MODE;THIEF_ASK;
```

it seems like THIEF_ASK is supposed to make the npc ask before taking stuff...?

nornagon:
seems similar to https://github.com/CleverRaven/Cataclysm-DDA/issues/54822

[...]

Issue #54822, opened 2022-01-26
NPC follower who agrees to follow, seems to ignore command "follow me" after set to guard


coyo7e:
### Describe the bug

I noticed this on a couple NPCs now, not sure if it's specific to the NPC or just because their wanted quest was not yet completed however, when you convince them to come with the player via a Social skill check, and then they get banged up (or are just bad in combat) and then they are asked to "follow me" after being set to guard, they ignore the "come with me" option, even after repeated attempts.

Unsure if this is related to their trust level or not however, it is annoying to set an NPC to guard a spot and then realize they won't come with me until I (assumedly) manage to complete their quest. Some quests take a lot of exploring and searching (such as the "relic" quest) so this can end with an NPC left behind accidentally, possibly in a dangerous area.

They seem to accept the order to come with player, but then continue to act in the same manner as a guard NPC and if you converse with them again, they will continue to refuse to leave the area or stop guarding.

Specifically for the instance I'm currently looking at, it's the NPC who asks the player to find a relic.

Even forcing a new social check would be helpful in this circumstance, and at least make the AI behavior seem more logical.

### Steps to reproduce

1. Spawn NPC.
2. Use Social check to convince them to travel with player.
3. Set NPC to guard.
4. Notice they will not re-follow player.

### Expected behavior

Expect a follower who chooses to come with the player, would also accept re-following the player when given the apropos command, regardless of trust levels etc.

### Screenshots

![image](https://user-images.githubusercontent.com/80022377/151242798-c19001e1-1265-4a9e-b856-bfdc34dd4afd.png)

![image](https://user-images.githubusercontent.com/80022377/151242849-fd9d91bf-9707-410d-8342-5c72bbc371e1.png)

![image](https://user-images.githubusercontent.com/80022377/151242893-bcb12ec4-4386-4fbc-8fdc-b451744ce577.png)

![image](https://user-images.githubusercontent.com/80022377/151242957-74d6a619-1423-4b82-9b16-29355d518aa0.png)

(NPC is set to "follow always" "within 4 squares" and does not do so)

![image](https://user-images.githubusercontent.com/80022377/151250008-c6527cdb-3659-44da-9938-4ba7c42ad566.png)

They also can be quite spammy and ignore commands like to not pick up items or use ranged/loud weapons, etc

![image](https://user-images.githubusercontent.com/80022377/151251186-0c949521-a060-41d5-8273-132a8ad9d390.png)

[...]

PR #64775, opened 2023-04-02
add force-drop option to debug menu

nornagon:
#### Summary
Interface "Added force drop option to player/NPC debug menu."

#### Purpose of change

This allows cleaning up bugs like #64771 and #54822, which are otherwise
difficult to fix.

#### Describe the solution

Just pops the multidrop menu for the NPC.

I thought about allowing "control NPC follower" to allow controlling any nearby NPC, but that seemed like a harder fix and I wasn't sure if it might have unintended consequences. This seems straightforward.

#### Testing

I tested this locally and it works.

### DIFF
--- a/src/debug_menu.cpp
+++ b/src/debug_menu.cpp
@@ -1811,7 +1811,7 @@ static void character_edit_menu()
     }
 
     enum {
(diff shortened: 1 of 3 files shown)


nornagon:
hm. this should maybe also remove item ownership from the dropped items? otherwise the game marks them as belonging to the npc.

PatrikLundell:
It's probably better to have a separate debug command to remove ownership.

Issue #64765, opened 2023-04-01
Performance regression related to limb scores when using gcc


sparr:
### Describe the bug

#62092 introduced a significant performance regression when I build with gcc 12.2.1, but not when I build with clang.

The poor performance is exhibited in almost all in-game interactions, but not in UI interactions. For testing purposes I focused on moving the cursor in the firing interface, which does not involve any time passing.

### Attach save file

N/A, reproduced with new games.

### Steps to reproduce

1. Build with gcc 12.2.1 (and possibly other parameters I haven't narrowed down)
2. Play the game and observe significantly longer delays when moving around, or when moving the firing cursor

### Expected behavior

No slowdown

### Versions and configuration

Compiled current git master, and bisected back to the PR linked above.
Sorry for lack of game report, I didn't keep those binaries around through my other testing and improvement efforts.

### Additional context

Profile from `perf` when moving the firing cursor. Almost all time spent in get_limb_score. I have made a small improvement in  #64761 that should slightly reduce total number of times through this call stack, but not nearly the factor of four that would be appropriate. I initiated some discussions on Discord about caching the duplicated function calls here, but had trouble implementing that myself. I suspect clang is optimizing these repeated calls away and making the impact of the problematic PR 1/4 as bad so I don't notice it with clang, while gcc is failing to optimize it so the calls get repeated.
![image](https://user-images.githubusercontent.com/121299/229323215-10cbd80b-a0e1-4b14-8cad-015ede3e7a24.png)

Profile from `perf` within get_limb_score, for which I have made very small improvements in #64759 
![image](https://user-images.githubusercontent.com/121299/229323196-cf9cbcbe-48f2-4da3-9991-42428a5a3c1e.png)

`calculate_ranged_chances` calls `predict_recoil` and `Character::gun_engagement_moves`, both of which call `Character::aim_per_move`, so caching/memoizing that result could cut the problem in half.

`Character::aim_per_move` calls `Character::fastest_aiming_method_speed` and `Character::most_accurate_aiming_method_limit`, both of which call `Character::effective_dispersion`, so caching/memoizing that result could cut the problem in half [again].

And, finally, `Character::effective_dispersion` calls a few things that eventually call `Character::get_limb_score` which could probably also be improved with caching.

PR #64776, opened 2023-04-02
Store limb_scores as map instead of vector

sparr:
#### Summary
None

#### Purpose of change
Checking for the presence of a limb_score currently requires a linear search of the limb scores for the body part in question, which is slightly wasteful, especially if the body part doesn't have that limb score.

#### Describe the solution
Change the vector of ids and scores to a map from id to scores.

#### Testing
Ran test suite. Broke an arm in the game and confirmed some lifting tasks still degrade as before.

#### Additional context
This is another small effort toward resolving #64765

### DIFF
--- a/src/bodypart.cpp
+++ b/src/bodypart.cpp
@@ -415,10 +415,10 @@ void body_part_type::load( const JsonObject &jo, const std::string & )
         const JsonArray &jarr = jo.get_array( "limb_scores" );
         for( const JsonValue jval : jarr ) {
             bp_limb_score bpls;
-            bpls.id = limb_score_id( jval.get_array().get_string( 0 ) );
+            const limb_score_id id = limb_score_id( jval.get_array().get_string( 0 ) );
             bpls.score = jval.get_array().get_float( 1 );
             bpls.max = jval.get_array().size() > 2 ? jval.get_array().get_float( 2 ) : bpls.score;
-            limb_scores.emplace_back( bpls );
+            limb_scores[id] = bpls;
         }
     }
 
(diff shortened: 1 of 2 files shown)


mqrause:
Maybe also change the json into a proper map while you're at it?

src/bodypart.cpp
```diff
@@ -561,32 +561,23 @@ void body_part_type::check() const
...
 float body_part_type::get_limb_score( const limb_score_id &id ) const
 {
-    for( const bp_limb_score &bpls : limb_scores ) {
-        if( bpls.id == id ) {
-            return bpls.score;
-        }
+    if( limb_scores.count( id ) ) {
+        return limb_scores.at( id ).score;
     }
     return 0.0f;
```
irwiss, line 567: might be worth grabbing iterator instead, considering the recent perf issues - you can cut another search then e.g.
```suggestion
    const auto it = limb_scores.find( id );
    return it == limb_scores.end() ? 0.0f : *it;
```

sparr: good call. required a minor syntax change, but this is in now.

sparr:
@mqrause I actually wrote the C++ to read a map from the json, but decided against committing it. That can go in a separate PR if it's desirable, but I don't think it made the json any easier to read or write for a human.

mqrause:
Maybe it's just a me thing, but I always find these arrays of values with different types strange and sometimes it can be very unclear what kind of values and how many of them are allowed.

sparr:
Changing it from `array<array<string,int[,int]>>` to `object<string:array<int[,int]>>` may help a little, but you're still left with the mystery of one vs two ints. Changing the inner array to `object<"score":int[,"max":int]>` to make it entirely self-documenting bloats the json and slows load time. We have to draw the line somewhere, and I'm comfortable sticking with the choice of the original devs of the feature unless there's more solid support for changing it.

mqrause:
Yeah, it's just a suggestion. If you don't want to do it that's fine. I'll mention we also have a pretty common setup where you can use either a simple value or an array/object to unbloat it. For example like this with example data from legs:
```
"limb_scores": { "manip": [ 0.1, 0.2 ], "move_speed": 0.5, "swim": 0.15, "block": 1, "crawl": 0.2 }
```

PR #64777, opened 2023-04-02
[Bombastic Perks] Mo' Perks & Thick Skull adjustments

TheShadowFerret:
#### Summary
Mods "More Bombastic Perks"

#### Purpose of change

Fresh mod needs more content + Thick Skull perk:
-Conflicts with skintight layer clothing when your skull's in your head. 😛
-Blocks rigid headgear such as army helmets.

#### Describe the solution
Existing Perks:
     -Hobbyist typo fixed
     -Thick Skull moved to personal layer & given SOFT flag so headgear cannot conflict with it.
New Perks:
     Way of the Closed Fist/Pinched Fingers/Open Palm:
           -All provide hand armor
           -Serve as bash/pierce/cut damage when punching w/o gloves.
           -Give (ballpeen? no claw end) hammer, small wrench, or knife qualities
           -Required unarmed 2 & their damage type 1
           -Mutually exclusive
           -Reference to Jade Empire :D

#### Describe alternatives you've considered

Move locked out perks to their own line instead of being combined with prereq skills' line.

#### Testing

Get a skill point with 0 in skills, see I cannot take Way of the X perks.
Get a skill point with prereq skills, see I can take Way of X perk. The other options disappear after I do.

#### Additional context
image

First usage of mutually exclusive & skill-gated perks :D

PR #64792, opened 2023-04-02
[bombastic perks] cleaned up prereqs stuff

bombasticSlacks:
#### Summary
Mods "cleaned up prereqs"

#### Purpose of change
Set up some standards for the style  of perks added in #64777 

#### Describe the solution
Added docs, migrated to a single entry with logic that respects mod settings.

#### Testing
Works in game

#### Additional context
@TheShadowFerret if you are interested in seeing what I cooked up for any of your additional perks, if they have reqs they should have the format in the doc in this PR. This way a player can view them (even if they can't select them), and there's a notification if it doesn't work.

### DIFF
--- a/data/mods/BombasticPerks/perkmenu.json
+++ b/data/mods/BombasticPerks/perkmenu.json
@@ -51,60 +51,16 @@
       },
       {
         "condition": {
-          "and": [
-            {
-              "not": {
-                "or": [
-                  { "u_has_trait": "perk_way_openpalm" },
-                  { "u_has_trait": "perk_way_closedfist" },
-                  { "u_has_trait": "perk_way_pinchedfingers" }
-                ]
-              }
-            },
-            { "u_has_skill": { "skill": "unarmed", "level": 2 } },
-            { "u_has_skill": { "skill": "cutting", "level": 1 } }
-          ]
+          "not": {
+            "or": [
+              { "u_has_trait": "perk_way_openpalm" },
+              { "u_has_trait": "perk_way_closedfist" },
+              { "u_has_trait": "perk_way_pinchedfingers" }
+            ]
+          }
         },
-        "text": "Gain [<trait_name:perk_way_openpalm>]",
+        "text": "Gain [Way Of The Fist]",
         "topic": "TALK_PERK_MENU_OPENPALM"
-      },
-      {
-        "condition": {
-          "and": [
-            {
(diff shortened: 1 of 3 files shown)


TheShadowFerret:
>@TheShadowFerret if you are interested in seeing what I cooked up for any of your additional perks, if they have reqs they should have the format in the doc in this PR. This way a player can view them (even if they can't select them), and there's a notification if it doesn't work.

I shall rewrite the two I made & retracted for these 🦦

data/mods/BombasticPerks/perkmenu.json
```diff
@@ -51,61 +51,17 @@
...
+            "or": [
+              { "u_has_trait": "perk_way_openpalm" },
+              { "u_has_trait": "perk_way_closedfist" },
+              { "u_has_trait": "perk_way_pinchedfingers" }
+            ]
+          }
         },
-        "text": "Gain [<trait_name:perk_way_openpalm>]",
+        "text": "Gain [Way Of The Fist]",
         "topic": "TALK_PERK_MENU_OPENPALM"
```
TheShadowFerret, line 63: Shouldn't this topic be renamed if it leads to more than just Way of the Open Palm 🤔

bombasticSlacks: probably, yeah

TheShadowFerret:
Do mutually exclusive perks count as variants if they're just the reverse of each other?

(Asking for an anti-Bloody Mess perk for skin & bionic harvesters)

bombasticSlacks:
> Do mutually exclusive perks count as variants if they're just the reverse of each other?
(Asking for an anti-Bloody Mess perk for skin & bionic harvesters)

I think, if possible, anything that is mutually exclusive should be on the same inner menu.

TheShadowFerret:
> I think, if possible, anything that is mutually exclusive should be on the same inner menu.

Should perks that 'stack'(ex. Stronger leads to Stronger...er 😛 I'm not going to PR that) or require some related perk also be in the same inner menu?

bombasticSlacks:
> Should perks that 'stack'(ex. Stronger leads to Stronger...er 😛 I'm not going to PR that) or require some related perk also be in the same inner menu?

I can't imagine a situation where that is necessary, and would prefer it's never the case, but in that case, for clarity they would have to be separate entries.

PR #64793, opened 2023-04-02
JSON sewer mapgen

### LINKED ISSUE #58404: Proper sewer mapgen

Sewers are super hacky currently. Their "base shape" is a 4 tile wide tunnel, which can connect to any side. Currently there is a map extra spawned on *every* overmap tile. Since it can connect to any side, there can't be any walls. I already tried adding walls because "walls" made of soil are annoying but this is simply not possible due to the issue I just described.

image

### Solution you would like.

Something akin to how subways are currently generated. Perhaps also a change in how these map extras are placed, could be done in json, so it only places most of them in the straight tunnels (e.g `sewer_ns`). The latter one could be done right now with a bit of effort and time I'd wager.

_No response_

### Additional context

_No response_

FarkasKBence:
#### Summary

Content "Sewer mapgen overhaul"

#### Purpose of change

Map extras were a rather restrictive way of adding decor to sewers.  For one there was no differentiation between curved, straight or fourway sewer tiles which prevented me from adding walls. Same goes for pipes and walkways; how I attempted to solve it was branch them out in all directions, hoping that they'd connect with a neighboring tile.  The technique started show its faults after a while.

Fixes #58404 (I hope)
Partially addresses #62747

#### Describe the solution

Removed sewer map extras and replaced them with regular JSON mapgen. These maps take sewer shape into account, making connected walls, pipes and walkways possible.  As an added bonus, this also frees up slots for new sewer extras.

I also edited locations with sewer accesses to have a seamless connection with the new design.
Dehardcoded manhole ladder placements for the extra flexibility.

Removing hardcoded mapgen. Not sure how to do that.

#### Testing

Auto-walked through what felt like 500 map tiles of sewers.  I also checked individual variants by inflating their weights and guaranteeing nested chunk placements.

#### Additional context

<details>
<summary> Screenshots: </summary>

kép

kép

kép

</details>


NetSysFire:
As the author of #58404, yes this fixes it. Very nice work!

PR #64795, opened 2023-04-02
Update Load menu world selection on create/delete

sparr:
#### Summary
None

#### Purpose of change
Creating or deleting a world in the world menu results in the Load menu selecting the wrong world by default for loading

#### Describe the solution
Move last_world_pos out to the main_menu class, and update it when creating or deleting worlds based on their order

Could pass last_world_pos into world_tab by reference instead of moving it to the class

#### Testing
Created and deleted worlds and went back and forth between menus

### DIFF
--- a/src/main_menu.cpp
+++ b/src/main_menu.cpp
@@ -609,7 +609,6 @@ bool main_menu::opening_screen()
     player_character = avatar();
 
     int sel_line = 0;
-    size_t last_world_pos = 0;
 
     // Make [Load Game] the default cursor position if there's game save available
     if( !world_generator->get_all_worlds().empty() ) {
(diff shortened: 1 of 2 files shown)


src/main_menu.cpp
```diff
@@ -1072,7 +1071,10 @@ void main_menu::world_tab( const std::string &worldname )
 {
     // Create world
     if( sel2 == 0 ) {
-        world_generator->make_new_world();
+        WORLD *world = world_generator->make_new_world();
+        if( world->world_name < world_generator->all_worldnames()[last_world_pos] ) {
```
kevingranade, line 1075: This seems like you want equality instead of less than? I think that will also silence the warning. 
```suggestion
        if( world->world_name != world_generator->all_worldnames()[last_world_pos] ) {
```

PR #64801, opened 2023-04-02
Monsters will now fire at moving vehicles

Maleclypse:
#### Summary
Features "Monsters with guntypes will now fire at moving vehicles."

#### Purpose of change

Implementation in the port lacks a number of key features for this to go into DDA as it is unfairly punishing on the player in nonsensical ways.  I however am unable to make the changes personally that would bring this into a mergeable state.
Ports the c++ from https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/4 which is a port of https://github.com/cataclysmbnteam/Cataclysm-BN/pull/2165 by Night_Pryanik and joveeater respectively.   Adds significantly more monster types to the monsters that will now shoot at a moving vehicle.   Also adds documentation of this so that new contributors won't have to search the c++.  Thus the two separate commits.

#### Describe the solution
C++ makes guntypes able to shoot at moving vehicles.  Json defines what monsters will now target moving vehicles.   Documentation explains the json object. Added it to gun toting monsters in Xedra Evolved and Dinomod.  

Waiting on others

#### Testing
It works really well.  I'm super happy with it.

#### Additional context
Car vs UGV


src/mattack_actors.cpp
```diff
@@ -847,20 +872,43 @@ bool gun_actor::call( monster &z ) const
...
+            vehicle *veh = find_target_vehicle( z, get_max_range() );
+            if( !veh ) {
+                return false;
+            }
+            for( const vpart_reference &vp : veh->get_avail_parts( "CONTROLS" ) ) {
+                if( z.sees( vp.pos() ) ) {
+                    aim_at = vp.pos();
+                    untargeted = true;
+                }
+            }
```
irwiss, line 893: this chunk feels sorta weird, unless i missed some checks (it is kinda late for me) this will turn monsters into clairvoyant snipers;
detecting moving vehicle doesn't seem to require sight or noise - monsters will shoot up to their max range anyway
for some reason they'll aim at controls - even migo/yrax or acid llama will know what CONTROLS look like
and only the first CONTROLS tile, creating a meme/meta for sacrificial set of armored controls

Maleclypse: I'm going to be honest here I am a cargo cultist so if there is a way to improve this, someone other than me will have to make the changes.  I agree with you it would make more sense for them to target randomly, wheels, engines, controls, everything else?  How would I go about making that happen?

irwiss: One way would be to grab vehicle->get_points() and roll via random_entry() from those points - that'll hit some random vehicle tile, or may be a more complex roll that weighs the front tiles more

a more weird thing is the lack of sight or noise checks, getting shot at by a feral or llama spitting acid at you while riding a silent bicycle in pitch dark seems odd, but may be i've missed the check for that or it happens earlier, or BN players don't care/complain loud enough

or just put it in as-is and try to balance/fix later when some gameplay testing occurs

PatrikLundell: I would expect a (functioning) human to aim at the driver or where the driver ought to be if not directly visible (i.e. at the controls), unless trying to take out the vehicle specifically by firing an RPG at the engine block or trying to take out the wheels. On a normal car you know where the driver seat is located, while the Cataclysm custom jobs may require some more intelligence (which the game doesn't have unless you use the controls as a proxy) to figure out where the driver might be located.

A non intelligent opponent would probably fire at the center, as perceived from their perspective, i.e. the center tile on the facing "side" (easy for orthogonal orientations, but trickier for the rest), with dispersion from that location hitting tiles further away to either side.

irwiss: Playtested it a bit;
Spawned m2hb turret 60 tiles away - tested both daylight and new moon (no ambient light), the turret had no vision indicator in either but; It will happily frolick while player is in control of velocity=0 bicycle, the moment bicycle starts moving it will shoot to beyond it's daylight sight range, the same happens with migo, acid llamas etc

So no vision or noise checks made means they'll detect moving vehicles in reality bubble - also behind several walls / smoke etc

Maleclypse: So then it definitely needs more work. I’ll leave it open for people to make suggestions and commit them but otherwise this probably isn’t mergeable.

Maleclypse:
@irwiss will this do what you were suggesting or get me closer to it?

```
std::vector<tripoint> valid_targets;
            for (const tripoint& p : veh->get_points()) {
                if (veh->part_with_feature(p, VPFLAG_CONTROLS, true) >= 0 &&
                    veh->part_with_feature(p, VPFLAG_ENGINE, true) >= 0 &&
                    veh->part_with_feature(p, VPFLAG_WHEEL, true) >= 0) {
                    valid_targets.push_back(p);
                }
            }
```

src/mattack_actors.cpp
```diff
@@ -847,20 +875,43 @@ bool gun_actor::call( monster &z ) const
...
+            if( !untargeted ) {
+                untargeted = true;
+                aim_at = veh->global_pos3();
+            }
+        } else {
+            aim_at = target->pos();
```
anothersimulacrum, line 902: Something like this will make it pick a nice random point
```suggestion
            std::vector<tripoint> valid_targets;
            std::vector<tripoint> visible_points;
            for( const tripoint &p : veh->get_points() ) {
                if( !z.sees( p ) ) {
                    continue;
                }
                visible_points.push_back( p );
                if( veh->part_with_feature( p, VPFLAG_CONTROLS, true ) >= 0 &&
                    veh->part_with_feature( p, VPFLAG_ENGINE, true ) >= 0 &&
                    veh->part_with_feature( p, VPFLAG_WHEEL, true ) >= 0) {
                    valid_targets.push_back( p );
                }
            }
            if( !targets.empty() ) {
                std::shuffle( valid_targets.begin(), valid_targets.end(), rng_get_seed() );
                aim_at = valid_targest.back();
            } else if( !visible_points.empty() ) {
                std::shuffle( visible_points.begin(), visible_points.end(), rng_get_seed() );
                aim_at = visible_points.back();
            } else {
                return false;
            }
```

src/mattack_actors.cpp
```diff
@@ -888,18 +888,28 @@ bool gun_actor::call( monster &z ) const
...
+                if( veh->part_with_feature( p, VPFLAG_CONTROLS, true ) >= 0 &&
+                    veh->part_with_feature( p, VPFLAG_ENGINE, true ) >= 0 &&
+                    veh->part_with_feature( p, VPFLAG_WHEEL, true ) >= 0) {
+                    valid_targets.push_back( p );
                 }
             }
-            if( !untargeted ) {
-                untargeted = true;
-                aim_at = veh->global_pos3();
+            if( !targets.empty() ) {
```
anothersimulacrum, line 904: Whoops
```suggestion
            if( !valid_targets.empty() ) {
```

src/mattack_actors.cpp
```diff
@@ -847,20 +875,51 @@ bool gun_actor::call( monster &z ) const
...
+                    valid_targets.push_back( p );
+                }
+            }
+            if( !valid_targets.empty() ) {
+                aim_at = random_entry( valid_targets, tripoint_zero );
+            } else if( !visible_points.empty() ) {
+                aim_at = random_entry( visible_points, tripoint_zero );
+            } else {
+                return false;
+            }
```
anothersimulacrum, line 910:
```suggestion
            }
            untargeted = true;
```

src/mattack_actors.cpp
```diff
@@ -847,20 +875,51 @@ bool gun_actor::call( monster &z ) const
...
+        aim_at = target->pos();
     } else {
         target = z.attack_target();
         if( !target || !z.sees( *target ) || ( !target->is_monster() && !z.aggro_character ) ) {
-            return false;
+            //return false;
+            if( !target_moving_vehicles ) {
+                return false;
+            }
+            //No living targets, try to find a moving car
```
anothersimulacrum, line 886:
```suggestion
            //No living targets, try to find a moving car
            untargeted = true;
```

src/mattack_actors.cpp
```diff
@@ -847,20 +875,51 @@ bool gun_actor::call( monster &z ) const
...
+            }
         }
     }
 
-    int dist = rl_dist( z.pos(), target->pos() );
-    add_msg_debug( debugmode::DF_MATTACK, "Target %s at range %d", target->disp_name(), dist );
+    int dist = rl_dist( z.pos(), aim_at );
+    if( target ) {
+        add_msg_debug( debugmode::DF_MATTACK, "Target %s at range %d", target->disp_name(), dist );
+    }
```
anothersimulacrum, line 917:
```suggestion
    } else {
        add_msg_debug( debugmode::DF_MATTACK, "Shooting at vehicle at range %d", dist );
    }
```

src/mattack_actors.cpp
```diff
@@ -847,20 +875,54 @@ bool gun_actor::call( monster &z ) const
...
+            std::vector<tripoint> valid_targets;
+            std::vector<tripoint> visible_points;
+            for( const tripoint &p : veh->get_points() ) {
+                if( !z.sees( p ) ) {
+                    continue;
+                }
+                visible_points.push_back( p );
+                if( veh->part_with_feature( p, VPFLAG_CONTROLS, true ) >= 0 &&
+                    veh->part_with_feature( p, VPFLAG_ENGINE, true ) >= 0 &&
+                    veh->part_with_feature( p, VPFLAG_WHEEL, true ) >= 0 ) {
```
anothersimulacrum, line 901: These want an index, not a point. Either an overload with a point needs to be added or this condition needs to iterate through the parts at this point and see if any of them have these flags.

src/mattack_actors.cpp
```diff
@@ -847,20 +875,54 @@ bool gun_actor::call( monster &z ) const
...
+                    valid_targets.push_back( p );
+                }
```
dseguin, line 903: Going with anothersim's suggestion. Try this

```suggestion
                for( const vpart_reference &vp : veh->get_all_parts() ) {
                    if( vp.pos() != p ) {
                        continue;
                    }
                    if( veh->part_with_feature( vp.part_index(), VPFLAG_CONTROLS, true ) >= 0 &&
                        veh->part_with_feature( vp.part_index(), VPFLAG_ENGINE, true ) >= 0 &&
                        veh->part_with_feature( vp.part_index(), VPFLAG_WHEEL, true ) >= 0 ) {
                        valid_targets.push_back( p );
                        break;
                    }
                 }
```

Maleclypse:
```../tests/monster_attack_test.cpp:223: FAILED:
(~[slow] ~[.],starting_items)=>   REQUIRE( attack->call( test_monster ) == true )
Error: (~[slow] ~[.],starting_items)=>   false == true```
Hmm I wonder if I borked this test?

**Comment by irwiss** [comment]
The test borked because `aim_at` is not initialized around line 880 when target != nullptr

I suggest the attached refactor below (applied on top of your changes);
Fixes the mentioned aim_at bug
Moves both parts of find-me-a-part into map, cutting the 2 extra includes
Iterates the parts once per vehicle; the current code will make points*part_count iterations, and each iteration will then iterate another vector of parts-in-this-square)
Filters based on vehicle part positions rather than vehicles; current code assumes if vehicle center is out of range then all parts are out of range, which is ok for long range turrets but looks bad for short range enemies

[Cataclysm-DDA-5ef3600-Refactor.patch](https://github.com/CleverRaven/Cataclysm-DDA/files/11163997/Cataclysm-DDA-5ef3600-Refactor.patch)

For convenience putting the diff here;

```patch
From bcb1ace95a522cad1ddb3d72b9b2361101aeb89e Mon Sep 17 00:00:00 2001
From: Alexey <irwiss@users.noreply.github.com>
Date: Thu, 6 Apr 2023 01:00:31 +0300
Subject: [PATCH] Refactor

---
 src/map.cpp            | 35 ++++++++++++++++++++++++
 src/map.h              |  4 +++
 src/mattack_actors.cpp | 62 +++++-------------------------------------
 3 files changed, 46 insertions(+), 55 deletions(-)

diff --git a/src/map.cpp b/src/map.cpp
index 726839e0eb..64d75795fc 100644
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -1122,6 +1122,41 @@ bool map::deregister_vehicle_zone( zone_data &zone ) const
     return false;
 }
 
+std::set<tripoint_bub_ms> map::get_moving_vehicle_targets( const Creature &z, int max_range )
+{
+    const tripoint_bub_ms zpos( z.pos() );
+    std::set<tripoint_bub_ms> priority;
+    std::set<tripoint_bub_ms> visible;
+    for( wrapped_vehicle &v : get_vehicles() ) {
+        if( !v.v->is_moving() ) {
+            continue;
+        }
+        if( !fov_3d && v.pos.z != zpos.z() ) {
+            continue;
+        }
+        if( rl_dist( zpos, tripoint_bub_ms( v.pos ) ) > max_range + 40 ) {
+            continue; // coarse distance filter, 40 = ~24 * sqrt(2) - rough max diameter of a vehicle
+        }
+        for( const vpart_reference &vpr : v.v->get_all_parts() ) {
+            const tripoint_bub_ms vppos = static_cast<tripoint_bub_ms>( vpr.pos() );
+            if( rl_dist( zpos, vppos ) > max_range ) {
+                continue;
+            }
+            if( !z.sees( vppos ) ) {
+                continue;
+            }
+            if( vpr.has_feature( VPFLAG_CONTROLS ) ||
+                vpr.has_feature( VPFLAG_ENGINE ) ||
+                vpr.has_feature( VPFLAG_WHEEL ) ) {
+                priority.emplace( vppos );
+            } else {
+                visible.emplace( vppos );
+            }
+        }
+    }
+    return !priority.empty() ? priority : visible;
+}
+
 // 3D vehicle functions
 
 VehicleList map::get_vehicles( const tripoint &start, const tripoint &end )
diff --git a/src/map.h b/src/map.h
index 959d02196b..e98630af5c 100644
--- a/src/map.h
+++ b/src/map.h
@@ -675,6 +675,10 @@ class map
         std::vector<zone_data *> get_vehicle_zones( int zlev );
         void register_vehicle_zone( vehicle *, int zlev );
         bool deregister_vehicle_zone( zone_data &zone ) const;
+        // returns a list of tripoints which contain parts from moving vehicles within \p max_range
+        // distance from \p source position, if any parts are CONTROLS, ENGINE or WHEELS returns a
+        // list of tripoints with exclusively such parts instead. Used for monster gun actor targeting.
+        std::set<tripoint_bub_ms> get_moving_vehicle_targets( const Creature &source, int max_range );
 
         // Removes vehicle from map and returns it in unique_ptr
         std::unique_ptr<vehicle> detach_vehicle( vehicle *veh );
diff --git a/src/mattack_actors.cpp b/src/mattack_actors.cpp
index 95949cee50..025aef2691 100644
--- a/src/mattack_actors.cpp
+++ b/src/mattack_actors.cpp
@@ -32,9 +32,7 @@
 #include "rng.h"
 #include "sounds.h"
 #include "translations.h"
-#include "vehicle.h"
 #include "viewer.h"
-#include "veh_type.h"
 
 
 static const efftype_id effect_badpoison( "badpoison" );
@@ -833,27 +831,6 @@ int gun_actor::get_max_range()  const
     return max_range;
 }
 
-static vehicle *find_target_vehicle( monster &z, int range )
-{
-    map &here = get_map();
-    vehicle *chosen = nullptr;
-    for( wrapped_vehicle &v : here.get_vehicles() ) {
-        if( !z.sees( v.pos ) ) {
-            continue;
-        }
-        if( !fov_3d && v.pos.z != z.pos().z ) {
-            continue;
-        }
-        int new_dist = rl_dist( z.pos(), v.pos );
-        if( v.v->velocity != 0 && new_dist < range ) {
-            chosen = v.v;
-            range = new_dist;
-        }
-    }
-    return chosen;
-}
-
-
 bool gun_actor::call( monster &z ) const
 {
     Creature *target;
@@ -878,47 +855,22 @@ bool gun_actor::call( monster &z ) const
         aim_at = target->pos();
     } else {
         target = z.attack_target();
+        aim_at = target ? target->pos() : tripoint_zero;
         if( !target || !z.sees( *target ) || ( !target->is_monster() && !z.aggro_character ) ) {
-            //return false;
             if( !target_moving_vehicles ) {
                 return false;
             }
-            //No living targets, try to find a moving car
-            untargeted = true;
-            vehicle *veh = find_target_vehicle( z, get_max_range() );
-            if( !veh ) {
-                return false;
-            }
-            std::vector<tripoint> valid_targets;
-            std::vector<tripoint> visible_points;
-            for( const tripoint &p : veh->get_points() ) {
-                if( !z.sees( p ) ) {
-                    continue;
-                }
-                visible_points.push_back( p );
-                for( const vpart_reference &vp : veh->get_all_parts() ) {
-                    if( vp.pos() != p ) {
-                        continue;
-                    }
-                    if( veh->part_with_feature( vp.part_index(), VPFLAG_CONTROLS, true ) >= 0 &&
-                        veh->part_with_feature( vp.part_index(), VPFLAG_ENGINE, true ) >= 0 &&
-                        veh->part_with_feature( vp.part_index(), VPFLAG_WHEEL, true ) >= 0 ) {
-                        valid_targets.push_back( p );
-                        break;
-                    }
-                }
-            }
-            if( !valid_targets.empty() ) {
-                aim_at = random_entry( valid_targets, tripoint_zero );
-            } else if( !visible_points.empty() ) {
-                aim_at = random_entry( visible_points, tripoint_zero );
-            } else {
+            untargeted = true; // no living targets, try to find moving car parts
+            const std::set<tripoint_bub_ms> moving_veh_parts = get_map()
+                    .get_moving_vehicle_targets( z, get_max_range() );
+            if( moving_veh_parts.empty() ) {
                 return false;
             }
+            aim_at = random_entry( moving_veh_parts, tripoint_bub_ms() ).raw();
         }
     }
 
-    int dist = rl_dist( z.pos(), aim_at );
+    const int dist = rl_dist( z.pos(), aim_at );
     if( target ) {
         add_msg_debug( debugmode::DF_MATTACK, "Target %s at range %d", target->disp_name(), dist );
     } else {
-- 
2.39.1.windows.1


```

Maleclypse:
Car vs CROWS

PR #64809, opened 2023-04-03
Deduplicate diary border drawing code

sparr:
#### Summary
None

#### Purpose of change
The diary ui code is very repetitive when drawing small blocks of ascii art

#### Describe the solution
Add a function to print multiple strings stacked.

I considered making the new function available more broadly as part of catacurses, but that seems unnecessary until it's used more places. I also considered passing the strings in a vector so the count wouldn't be needed, but that felt like unnecessary overhead.

#### Testing
Opened the game, looked at the diary.

### DIFF
--- a/src/diary_ui.cpp
+++ b/src/diary_ui.cpp
@@ -93,56 +93,75 @@ void print_list_scrollable( catacurses::window *win, const std::string &text, in
     print_list_scrollable( win, list, selection, active, border, color_error );
 }
 
-void draw_diary_border( catacurses::window *win, const nc_color &color = c_white )
+// Print multiple strings, stacked vertically downwards from the specified point
+template<typename ...Ts>
+void mvwprintwa( const catacurses::window &win, point p, Ts... args )
 {
-    wattron( *win, color );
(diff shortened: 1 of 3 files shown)


sparr:
Current version has undesired color inconsistency.
image

src/diary_ui.cpp
```diff
@@ -93,56 +93,75 @@ void print_list_scrollable( catacurses::window *win, const std::string &text, in
     print_list_scrollable( win, list, selection, active, border, color_error );
 }
 
-void draw_diary_border( catacurses::window *win, const nc_color &color = c_white )
+// Print multiple strings, stacked vertically downwards from the specified point
+template<typename ...Ts>
+void mvwprintwa( const catacurses::window &win, point p, Ts... args )
 {
-    wattron( *win, color );
```
sparr, line 98: I intend to bring back the configurable color (even though it wasn't used, I have a use in mind for it) in a future PR that eliminates a lot of unnecessary (and sometimes unwanted or counterproductive) color switching within functions like mvwputch and mvwhline.

PR #64810, opened 2023-04-03
Hopefully refine telefragging behavior

anoobindisguise:
#### Summary
Bugfixes "resolve issues with telefrag"

#### Purpose of change
Telefrag's replacement, which lacks an easy and consistent name ("teleport collision" isn't as nice as telefrag was) had some bugs. It could ignore the player's dimensional anchor, which I somehow didn't notice for literally over a dozen individual microlabs, and it sometimes left behind "phantom hitboxes" of enemies which went away after reloading the game but were impassible.

#### Describe the solution
Clean up the messy as hell code that was responsible for the new changes. Handle both parts of the teleport collision in one single block and add a check to break out if the victim in the targeted space is anchored.

#### Testing
Will require testing.


anoobindisguise:
I tested this and it seems to be functional, and tentatively it looks like the phantom hitbox bug may no longer be happening although i'm not 100% sure as it is very hard to reproduce.

PR #64812, opened 2023-04-03
Spawn graboids after 13 days

### LINKED ISSUE #64247: Graboids Should not Occur in Earlygame

Why are graboids? I can't outwalk them, I can't outrun them, they're "the size of semi trucks", can't be seen until they reveal themselves, and show up in innocuous places very early in the game. I just got murdered by one on day 2 in probably the cheapest death this game has given me because I thought the pile of dirt (which I didn't notice until it was very close already) would be a half worm or something. This is worse than a zombie hulk or shocker brute showing up, because these are both combat encounters you can avoid - you can spot them from a distance and, failing that, just start sprinting in the opposite direction, but unless you're very near some pavement, the graboid can chase you down and kill you, and you likely won't see it coming because it will run up to you outside your view distance. While new piles of dirt showing up would be obvious in most games, if it happens outside your zoom radius, you won't see an icon on the mini-map, you won't see an entity on your entity list, and it won't proc safe mode, so all of the tools designed to tell the player they are in danger fail them despite the fact that a character paying attention should, in universe, easily be able to see or hear it coming.

### Solution you would like.

Graboids should not show up until the game can reasonably assume they'd be able to deal with them. Making them the product of an evolution line rather than a day one spawn would be a very good idea. Might also be advisable to add sounds in-game for tunneling monsters moving through the ground (especially large ones) so the player could hear them coming instead of solely relying on visual cues.

Making the graboid weak enough for an early survivor to actually fight them would take away something from the game, but it's still possible. Could also limit spawn locations to places like caves where player would expect higher risks to occur.

### Additional context

_No response_

Leonkoithara:
#### Summary
Balance "Spawn graboids after 13 days"

#### Purpose of change
Currently graboids can spawn on day 1 with a 3% chance changed this to spawn only after 13 days. This Fixes #64247 

#### Describe the solution
Added the "starts" flag to monstergroups json file

Used 13 days because it is slightly weaker than a giant wasp queen which can start spawning after 15 days

### DIFF
--- a/data/json/monstergroups/bugs.json
+++ b/data/json/monstergroups/bugs.json
@@ -270,7 +270,7 @@
     "name": "GROUP_WORM",
     "is_animal": true,
     "monsters": [
-      { "monster": "mon_graboid", "weight": 30, "cost_multiplier": 20, "starts": "13 days"},
+      { "monster": "mon_graboid", "weight": 30, "cost_multiplier": 20, "starts": "13 days" },
       { "monster": "mon_halfworm", "weight": 0, "cost_multiplier": 0 },
       { "monster": "mon_worm_small", "weight": 970, "cost_multiplier": 0 }
     ]
--- a/data/json/monstergroups/bugs.json
+++ b/data/json/monstergroups/bugs.json
@@ -270,7 +270,7 @@
     "name": "GROUP_WORM",
     "is_animal": true,
     "monsters": [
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):

bombasticSlacks:
Rationale makes sense but this really shouldn't be spawning at all for months potentially a full year. However as you say other creatures are currently in the same boat.

PR #64813, opened 2023-04-03
add lamps to motels

qoreQyaS:
#### Summary
None

#### Purpose of change
There were no lamps in motels which doesn't make much sense irl.

#### Describe the solution
add lamps to motels in mapgen.

alone in the dark

#### Testing
I put the altered mapgen json into my running experimental, spawned the 5 motel types via debug and had look around.


LovamKicsiGazsii:
IMO this is a bit of an overkill.  Putting floor lamps in bedrooms is a welcome change, but I'm not quite sure about the bathroom ones.  Are floor lamps common bathroom furniture in the US? I'm especially surprised by the fancy lamp you put in the maintenance closet.

> There were no lamps in motels which doesn't make much sense irl.

I'd like to believe most lamps here are unrepresented overhead lights, sort of like in *every* building.

qoreQyaS:
i asked specifically about the overkill on discord.
[First message](https://discord.com/channels/598523535169945603/598535827169083403/1092460949111320707)
> **qoreQyaS**: I'd like to add floor lamps to motels, but i don't want it to be overkill. Do you think adding said lamps to the bathroom and motel rooms would be to much?
> **GuardianDll**: Rule of thumb: if it exist irl, the do it in such way
> **I-am-erk**: For something like floor lamps, if it's a commonplace item that's rare in game, it's not supposed to be rare in game
> **Ranik**: I just created a Start scenario involving motels. (A far less severe Really bad day where you are just drunk)  I thought they were oddly under-utilized. Floor lamps make sense i'd say

I've never been in a motel in the US, but:
My way of thinking here was that a room without light doesn't make much sense, that's why i put at least one lamp in every room (if i didn't forget any).
I would have been a lot happier with other means of lightning though, please shoot if ideas.

qoreQyaS:
the addition of the floor lamp furniture on different terrain then `fill_ter` was a nice learning experience for me, i'am fine with everything.
everyone who has an opinion on this please react to this message accordingly:
image lamps everywhere (bathrooms, motel rooms, maintenance closet and 4 of them in the reception hall)
image don't change the map at all
image add lamps only to motel rooms (1 for each room)

i'll change the PR in the light of the above in a week if the sky didn't fall on my head in meantime.

GuardianDll:
Well, it is indeed possible to put one here, but no one actually do it, because you should either put it in the place, where water can't hit it (which also means the lighting area would be really bad), or violate the common sense, and put the wired electric appliance right near the water sourse. None of this is a good options, therefore bathrooms (both motels and common one) uses integrated lamps, that are not open for short circuits due water. Therefore you should add something else here to be the source of light

qoreQyaS:
Can you please give me an example, i am quiet new to the game and don't know any other light sources.

GuardianDll:
Hard to say, it looks like we have lack of common light terrains, all i can find is `t_thconc_y_olight` and some variations, but they has `light_emitted`, which would be bad, because there is no electricity left. Either @LovamKicsiGazsii may know some, or you should make new terrain `ceiling lamp` and use it

harakka:
It's not necessary to have the bathrooms lit properly for this PR, it's enough that floor lamps are found in more places where you expect to find _floor lamps_ specifically. Other lighting needs can be tackled separately.

PR #64829, opened 2023-04-04
Enable auto pulping acidic zombies if acid protected

Rewryte:
#### Summary
Features "Enable auto pulping acidic zombies if acid protected"

#### Purpose of change
Auto pulping is disabled for all acidic zombies regardless of a character's immunities and protections.  This is a QoL change to allow acidic zombies to be pulped if the character is safe from acid puddles.

#### Describe the solution
Auto pulp conditions for acidic zombies should check for acid immunity, acid armor and environmental protection which are the same checks used by acid fields. This is done using existing functions "is_immune_damage( damage_type::ACID )" and "is_immune_field( fd_acid )".  If either function returns true, the acidic corpse is auto pulped.

#### Testing
Using the default scenario, I was able to auto pulp a spawned acidic zombie with either the acidproof mutation or the Hub 01 environmental protection suit (casual) equipped.

### DIFF
--- a/src/activity_handlers.cpp
+++ b/src/activity_handlers.cpp
@@ -190,7 +190,6 @@ static const itype_id itype_battery( "battery" );
 static const itype_id itype_burnt_out_bionic( "burnt_out_bionic" );
 static const itype_id itype_muscle( "muscle" );
 
-static const json_character_flag json_flag_ACID_IMMUNE( "ACID_IMMUNE" );
 static const json_character_flag json_flag_CANNIBAL( "CANNIBAL" );
 static const json_character_flag json_flag_PSYCHOPATH( "PSYCHOPATH" );
 static const json_character_flag json_flag_SAPIOVORE( "SAPIOVORE" );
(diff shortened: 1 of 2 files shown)


src/activity_handlers.cpp
```diff
@@ -1596,9 +1597,10 @@ void activity_handlers::pulp_do_turn( player_activity *act, Character *you )
     for( item &corpse : corpse_pile ) {
         const mtype *corpse_mtype = corpse.get_mtype();
         if( !corpse.is_corpse() || !corpse.can_revive() ||
-            ( std::find( act->str_values.begin(), act->str_values.end(), "auto_pulp_no_acid" ) !=
-              act->str_values.end() && corpse_mtype->bloodType().obj().has_acid ) ) {
-            // Don't smash non-rezing corpses //don't smash acid zombies when auto pulping
+            ( ( std::find( act->str_values.begin(), act->str_values.end(), "auto_pulp_no_acid" ) !=
```
irwiss, line 1600: you could pull out part of this into a variable e.g.;
`const bool acid_immune = you->is_immune_damage( damage_type::ACID ) && you->is_immune_field( fd_acid );`
then use it inside the if, the one below as well, and make this much more readable

json_flag_ACID_IMMUNE also seems unused in both files

Rewryte: ah right. forgot to remove that bit. ill push out another commit

irwiss: no need for re-review you can just resolve the convo if you addressed the comments in one way or another (the person who starts it can't do it)

PR #64830, opened 2023-04-04
setup catalytic cracking reactor as appliance

hfox:
#### Summary
Bugfixes "Adds catalytic cracking reactor as appliance, as told in its description"

#### Purpose of change
The catalytic cracking reactor says to plug it in/create an appliance, but that doesn't work and it does not provide qualities as furniture, only as a 500l, 22kg item. This fixes that.

#### Describe the solution
mostly just copying the hydraulic press, filling in values for the reactor instead, and hoping everything turns out as i hope.

waiting until someone else got annoyed enough at this.

#### Testing
copypasted to my phone and tested. works.[^1]

#### Additional context
500l diesel as byproduct means i can fuel my helicopter with zombies.
~~I also have no idea how much power is a reasonable draw to use it.~~[^1]

[^1]: doesn't work as I wanted it to(draw power when used), but as expected when thinking about it(require x power available to provide tool quality). I am not rebuilding how tool qualities work to make it work how i want it to.


data/json/items/fake.json
```diff
@@ -179,6 +179,14 @@
     "name": { "str": "jointer" },
     "copy-from": "fake_appliance_tool"
   },
+  {
+    "id": "catalytic_cracking_reactor_tool",
+    "type": "TOOL",
+    "copy-from": "fake_appliance_tool",
+    "name": { "str": "catalytic cracking reactor", "str_pl": "catalytic cracking reactors" },
+    "charges_per_use": 50,
```
hfox, line 187: Is this an ok amount?

RenechCDDA: I am not sure why it would use any? It does not use any (directly) in the existing form, why should the appliance form?

hfox: A real one I believe ought to use some for circulation of catalyst in the cases where the chemset creates the heat, otherwise heating too.
I realise now that correcting that might be out of scope for making it an appliance?

hfox: Actually testing shows that this doesn't draw any power when used, only makes it require that many available to be able to be used.

data/json/items/tool/science.json
```diff
@@ -355,20 +355,18 @@
...
-    "description": "A makeshift extraction system, or as makeshift as you can consider something that was purpose-built in a laboratory by a group of experts.  Meticulously disassembled and reassembled.",
-    "symbol": "U",
-    "color": "dark_gray",
-    "looks_like": "household_water_heater",
+    "type": "GENERIC",
+    "name": { "str": "disassembled catalytic cracking reactor" },
+    "description": "A makeshift extraction system, or as makeshift as you can consider something that was purpose-built in a laboratory by a group of experts.  Meticulously disassembled.  You can reassemble and plug it into your own power grid to use it.",
+    "symbol": "Y",
+    "color": "light_cyan",
+    "looks_like": "f_catalytic_cracking_reactor",
```
Karol1223, line 363: I am 99% sure this won't work.

data/json/items/fake.json
```diff
@@ -179,6 +179,15 @@
     "name": { "str": "jointer" },
     "copy-from": "fake_appliance_tool"
   },
+  {
+    "id": "catalytic_cracking_reactor_tool",
+    "type": "TOOL",
+    "copy-from": "fake_appliance_tool",
+    "name": { "str": "catalytic cracking reactor" },
+    "//": "there's a laptop im the disassembly, I assume it requires power.",
```
Karol1223, line 187: Typo?
```suggestion
    "//": "there's a laptop in the disassembly, I assume it requires power.",
```

hfox: Typo.

data/json/items/tool/science.json
hfox: No explicit errors/warnings when tested, but I'll look for something better.

Thanks.

data/json/furniture_and_terrain/appliances.json
```diff
@@ -388,6 +388,33 @@
     ],
     "variants": [ { "symbols": "7", "symbols_broken": "x" } ]
   },
+  {
+    "type": "vehicle_part",
+    "id": "ap_catalytic_cracking_reactor",
+    "name": { "str": "catalytic cracking reactor" },
+    "looks_like": "f_catalytic_cracking_reactor",
+    "symbol": "Y",
```
Maleclypse, line 396:
```suggestion
    "variants": [ { "symbols": "Y", "symbols_broken": "x" } ]
```

data/json/furniture_and_terrain/appliances.json
```diff
@@ -388,6 +388,33 @@
     ],
     "variants": [ { "symbols": "7", "symbols_broken": "x" } ]
   },
+  {
+    "type": "vehicle_part",
+    "id": "ap_catalytic_cracking_reactor",
+    "name": { "str": "catalytic cracking reactor" },
+    "looks_like": "f_catalytic_cracking_reactor",
+    "variants": [ { "symbols": "Y", "symbols_broken": "x" } ]
```
Karol1223, line 396:
```suggestion
    "variants": [ { "symbols": "Y", "symbols_broken": "x" } ],
```

Karol1223 (changes requested):

PR #64834, opened 2023-04-04
Replace short rope by actual seatbelt in vehicles

Fris0uman:
#### Summary
None

#### Purpose of change
Replace short rope by actual seatbelt in vehicles

#### Describe the solution

- [x] Add seatbelt item
- [x] Add seatbelt to similar requirements as short rope
- [x] Replace short rope by seatbelt in the seatbelt vehicle part
- [x] Have seatblet spawn in some places in the world
- [x] Add 5-point harness item
- [x] Make it spawn
- [x] Add a recipe

#### Testing

Load game without error
Spawn seatbelt without error
Get seatbelt for desinstalling seatbelt
Install seatblet using a seatbelt without error

#### Additional context

The item definition of seatbelt is mostly a copy-paste of the short rope I could use some advice there


GuardianDll:
Maybe make it not a "seat belt", but just a "belt"? It serves identical purpose to rope, but making it a "belt" limit it usage anywhere else

Fris0uman:
Where else would it be usefull? I added it to the same requirements as short rope but maybe I shouldn't because a belt is flat and might be harder to make strong knot with?

Brambor:
> I added it to the same requirements as short rope but maybe I shouldn't because a belt is flat and might be harder to make strong knot with?

As a climber, I would say flatness isn't such an issue. I will try a couple of knots on a seatbelt tomorrow if I don't forget :)

fairyarmadillo:
> Where else would it be usefull? I added it to the same requirements as short rope but maybe I shouldn't because a belt is flat and might be harder to make strong knot with?

I suspect Brambor will find it is a pretty poor tool for cordage. I think a seatbelt could work great as a bag strap, both for wearing and for like, strapping stuff to the bag. Ditto for any other piece of gear (ie leg guards) that strap to a character's body. I can't think of how you'd implement that without manually reworking every individual recipe though.

Brambor:
To get it off my chest:

I tried a simple knot and dragon bowline. The flatness truly makes it difficult. If I were to use it as a rope, I would sew together the seatbelt in half to make it twice less flat and twice as thick/cylindrical even.

When I was knotting, folding the seatbelt in half was the first step, but it didn't want to stay folded, therefore the stitching.

johnrdconnolly:
Removing a seatbelt with a screwdriver still results in a short rope. And there is no recipe for a seatbelt. I would make the argument that properly removing a seatbelt with a screwdriver and mechanic skill should result in a seatbelt instead of a short rope.

It could be further argued that attaching a seatbelt should take more than just screw driving. Perhaps also drilling 1? Drill holes for seatbelt, then attach.

To be fully comprehensive, the seatbelt item should have its own recipe that requires more than just a rope. Going off the fact that a seatbelt is both the belt and the mechanism that operates it. There should then also be a crude seatbelt, which would be just a tied rope. Crude seatbelt should be inferior in its abilities considering it is simply a tied rope in a car.

Fris0uman:
Can't reproduce: start new game, teleport in town, walk up to a car, remove the seatbelt, get a seatbelt item.

johnrdconnolly:
> Can't reproduce: start new game, teleport in town, walk up to a car, remove the seatbelt, get a seatbelt item.

You are right.. I just started a new game and it was fine. but still getting short rope on my other game that's been going for a while.... Perhaps I need to fine newly generated vehicles since updating..? And already generated ones have the rope as seatbelt already saved.?

Is there a recipe for a seatbelt?

Issue #64857, opened 2023-04-05
Extra key presses caused by #64819 or #64632


KHeket:
### Describe the bug

Before #64819 or #64632 if you bind items to any letter - when you press "a"ctivate and binded letter - it will be activated without extra key presses. Now, after this update, you need to additionaly press "enter", what is uncomfortable and create extra key presses

### Attach save file

N/A

### Steps to reproduce

Create character
Take activateable item
Bind it to any letter
press "a" and binded letter
See, that you need to extra press enter to activate it
Load any version before this, repeat and see, that you dont need to waist time for extra press enter

### Expected behavior

Hot keys working as before

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.2728 (22H2)
- Game Version: 6bd5d00 [64-bit]
- Graphics Version: Tiles
- Game Language: Системный язык []
- Mods loaded: [
    Тёмные Дни Впереди [dda],
    Отключить потребности у НПС [no_npc_food],
    Портальные штормы игнорируют НПС [personal_portal_storms],
    Без роста фунгалоидов [no_fungal_growth],
    Бионические профессии [package_bionic_professions]
]

KHeket:
@katemonster33 check pls

KHeket:
All is working good now, thx

PR #64862, opened 2023-04-05
Mouse hover hotkey regression

katemonster33:
#### Summary
Bugfixes "Fixed regression in inventory screens where pressing an item's hotkey does not work as expected"

#### Purpose of change
Fixing #64857 , regression caused by #64632 , pressing an item's hotkey currently only highlights the item.

#### Describe the solution
Screens now look for "ANY_INPUT" action when determining whether item passed in by input.entry should be selected.

#### Testing
Loaded the game and tested each screen with an item assigned with a hotkey.

### DIFF
--- a/src/inventory_ui.cpp
+++ b/src/inventory_ui.cpp
@@ -3098,7 +3098,7 @@ item_location inventory_pick_selector::execute()
             if( highlight( input.entry->any_item() ) ) {
                 ui_manager::redraw();
             }
-            if( input.action == "SELECT" ) {
+            if( input.action == "SELECT" || input.action == "ANY_INPUT" ) {
                 return input.entry->any_item();
             }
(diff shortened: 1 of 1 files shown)


src/inventory_ui.cpp
```diff
@@ -3098,7 +3098,7 @@ item_location inventory_pick_selector::execute()
             if( highlight( input.entry->any_item() ) ) {
                 ui_manager::redraw();
             }
-            if( input.action == "SELECT" ) {
+            if( input.action == "SELECT" || input.action == "ANY_INPUT" ) {
```
andrei8l, line 3101: This still doesn't work for invlets that shadow other actions, for example `q` (for quit) and `r` (for reset filter).

`if( input.action != "MOUSE_MOVE" && input.action != "COORDINATE" )` should restore the old behaviour in this case as well.

PR #64864, opened 2023-04-05
Add trap avoid json flag

### LINKED ISSUE #54849: Spiders fall into their own sinkholes and die
### Describe the bug

Spiders in a spider pit can fall into their own traps.

### Steps to reproduce

1. Teleport into a spider pit (surface).
2. Wait some turns.
3. The spiders trigger sinkholes.

image

(Also a minor case of #7533)

For more context, this is my debug playtesting world where all my stats are 25 which is why I spotted that sinkhole immediately and freely use the debug teleport ability.

### Expected behavior

Spiders should be immune to sinkholes or the sinkholes present should be turned into web-filled sinkholes, providing immunity to spiders but you get webbed and fall.

### Screenshots

_No response_

- experimental

### Additional context

_No response_

Leonkoithara:
#### Summary
Features "Add trap_avoids Json property to Monsters to control trap trigger"

#### Purpose of change
Some monsters set up traps which can be triggered by themselves(e.g Web spinning spiders and sinkholes Fixes #54849) and sometimes it doesn't make sense for certain traps to be triggered by monsters(e.g ferals and cot trap haven't added that to this PR can be a separate change)

#### Describe the solution
Create a new mtype attribute(trap_avoids) which is a vector of strings(trap_ids). These are populated from the json files during world creation.

Check for trap_id in monster::avoid_trap method if exists trap not triggered(Normally this function avoids trap only for fliers and diggers)

Hardcode monsters to individual traps however this seems a more generic approach and with more possibilities.

#### Testing
Added flags to monsters and made them walk through traps to see if they triggered them or not

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1045,7 +1045,7 @@ bool monster::avoid_trap( const tripoint & /* pos */, const trap &tr ) const
         return true;
     }
 
-    if( get_trap_avoids().count( tr.id ) > 0 ) {
+    if( type->trap_avoids.count( tr.id ) > 0 ) {
         return true;
     }
 
@@ -3594,9 +3594,4 @@ const pathfinding_settings &monster::get_pathfinding_settings() const
 std::set<tripoint> monster::get_path_avoid() const
 {
     return std::set<tripoint>();
-}
-
-const std::set<trap_str_id> &monster::get_trap_avoids() const
-{
-    return type->trap_avoids;
 }
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -62,6 +62,7 @@
 #include "text_snippets.h"
 #include "translations.h"
 #include "trap.h"
+#include "type_id.h"
 #include "units.h"
 #include "viewer.h"
 #include "weakpoint.h"
@@ -1044,8 +1045,7 @@ bool monster::avoid_trap( const tripoint & /* pos */, const trap &tr ) const
         return true;
     }
 
-    std::vector<std::string> tavoids = get_trap_avoids();
-    if (std::find(tavoids.begin(), tavoids.end(), tr.id.str()) != tavoids.end()) {
+    if( get_trap_avoids().count( tr.id ) > 0 ) {
         return true;
     }
 
@@ -3596,7 +3596,7 @@ std::set<tripoint> monster::get_path_avoid() const
     return std::set<tripoint>();
 }
 
-std::vector<std::string> monster::get_trap_avoids() const
+const std::set<trap_str_id> &monster::get_trap_avoids() const
 {
     return type->trap_avoids;
 }
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -1043,6 +1043,12 @@ bool monster::avoid_trap( const tripoint & /* pos */, const trap &tr ) const
     if( digging() || flies() ) {
         return true;
     }
+
+    std::vector<std::string> tavoids = get_trap_avoids();
+    if (std::find(tavoids.begin(), tavoids.end(), tr.id.str()) != tavoids.end()) {
+        return true;
+    }
+
     return dice( 3, type->sk_dodge + 1 ) >= dice( 3, tr.get_avoidance() );
 }
 
@@ -3588,4 +3594,9 @@ const pathfinding_settings &monster::get_pathfinding_settings() const
 std::set<tripoint> monster::get_path_avoid() const
 {
     return std::set<tripoint>();
+}
+
+std::vector<std::string> monster::get_trap_avoids() const
+{
+    return type->trap_avoids;
 }
--- a/src/monster.h
+++ b/src/monster.h
@@ -575,7 +575,6 @@ class monster : public Creature
         void on_load();
 
         const pathfinding_settings &get_pathfinding_settings() const override;
-        const std::set<trap_str_id> &get_trap_avoids() const;
         std::set<tripoint> get_path_avoid() const override;
         // summoned monsters via spells
         void set_summon_time( const time_duration &length );
--- a/src/monster.h
+++ b/src/monster.h
@@ -575,7 +575,7 @@ class monster : public Creature
         void on_load();
 
         const pathfinding_settings &get_pathfinding_settings() const override;
-        std::vector<std::string> get_trap_avoids() const;
+        const std::set<trap_str_id> &get_trap_avoids() const;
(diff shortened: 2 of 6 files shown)


src/monstergenerator.cpp
```diff
@@ -770,6 +770,16 @@ void mtype::load( const JsonObject &jo, const std::string &src )
...
 
+    if ( jo.has_array("trap_avoids") ) {
+        for( const JsonValue entry : jo.get_array( "trap_avoids" ) ) {
+            if( entry.test_string() ) {
+                trap_avoids.push_back( entry.get_string() );
+            } else {
+                entry.throw_error( "array element is not a string." );
+            }
+        }
+    }
```
anothersimulacrum, line 781: Doing it by hand is unnecessary and this has extra handy features.
```suggestion
  optional( jo, was_loaded, "trap_avoids", trap_avoids );
```

src/mtype.h
```diff
@@ -346,6 +346,9 @@ struct mtype {
         ::weakpoints weakpoints;
         weakpoint_families families;
 
+        // Traps avoided by this monster
+        std::vector<std::string> trap_avoids;
```
anothersimulacrum, line 350: Little more convenient for what you're using it for.
```suggestion
        std::set<std::string> trap_avoids;
```

src/monster.h
```diff
@@ -575,6 +575,7 @@ class monster : public Creature
         void on_load();
 
         const pathfinding_settings &get_pathfinding_settings() const override;
+        std::vector<std::string> get_trap_avoids() const;
```
anothersimulacrum, line 578: This makes a copy, that's expensive and unnecessary.
```suggestion
        const std::set<trap_id> &get_trap_avoids() const;
```

src/mtype.h
anothersimulacrum, line 350: A set is more convenient for lookup (and faster). Also, use the ids instead of raw strings.
Also, iterate through this in `MonsterGenerator::finalize_mtypes` and check that each `trap_id` `.is_valid()`.
```suggestion
        std::set<trap_id> trap_avoids;
```

src/monster.cpp
```diff
@@ -1043,6 +1043,12 @@ bool monster::avoid_trap( const tripoint & /* pos */, const trap &tr ) const
     if( digging() || flies() ) {
         return true;
     }
+
+    std::vector<std::string> tavoids = get_trap_avoids();
+    if (std::find(tavoids.begin(), tavoids.end(), tr.id.str()) != tavoids.end()) {
+        return true;
+    }
```
anothersimulacrum, line 1050:
```suggestion
    if( get_trap_avoids().count( tr.id ) > 0 ) {
        return true;
    }
```

src/monster.cpp
```diff
@@ -3589,3 +3595,8 @@ std::set<tripoint> monster::get_path_avoid() const
 {
     return std::set<tripoint>();
 }
+
+std::vector<std::string> monster::get_trap_avoids() const
```
anothersimulacrum, line 3599:
```suggestion
const std::set<trap_id> &monster::get_trap_avoids() const
```

irwiss: adding function to a popular class header whose only user is the class's own implementation, and it only saves typing like 7 characters seems not really necessary

src/monstergenerator.cpp
Leonkoithara: Resolved

src/mtype.h
Leonkoithara: Resolved

src/monster.h
Leonkoithara: Resolved

src/monster.cpp
Leonkoithara: I wanted to keep the code similar to what we already have such as get_pathfinding_settings etc.

Leonkoithara: Should I instead directly reference the attribute by trap->trap_avoids instead of the getter function?

src/mtype.h
Leonkoithara: Resolved

src/monster.cpp
Leonkoithara: Resolved

src/monster.cpp
irwiss: get_pathfinding_settings is abstract function from Creature and monster class must implement it, it's used in many places and has potential to be used even more as code evolves; while this field isn't "owned" by monster to begin with so it's at best a "nice shortcut"; but it's used only once, inside the class itself and very little potential for other classes being interested in it (and if they are they can just grab it via mtype)

Nebnis:
What about player made traps? The spider would know to avoid its own webs but not a bear trap

anothersimulacrum:
It's a list of traps to avoid, not a flag set to ignore all traps.

src/monstergenerator.cpp
```diff
@@ -770,6 +778,8 @@ void mtype::load( const JsonObject &jo, const std::string &src )
     assign( jo, "armor_pure", armor_pure, strict, 0 );
     assign( jo, "armor_biological", armor_biological, strict, 0 );
 
+    optional( jo, was_loaded, "trap_avoids", trap_avoids, string_id_reader<trap> {} );
```
anothersimulacrum, line 781: You probably don't need this reader (needed for int ids) but I don't know.

anothersimulacrum (approved):
Code looks good, but I agree with irwiss that the getter is a little superfluous.

src/monstergenerator.cpp
Leonkoithara: This is from where the Json is read and the trap_str_ids are fetched

src/monster.cpp
Leonkoithara: Removed superfluous getter function.

src/monstergenerator.cpp
Leonkoithara: The ids are needed to compare against the trap being triggered to check whether it gets triggered

anothersimulacrum: I'm saying this could probably just work as
```suggestion
    optional( jo, was_loaded, "trap_avoids", trap_avoids );
```
I'm fairly sure I've used it that way before in several other places.

But I can't test, and it doesn't hurt to have it here.

Leonkoithara: You are right that is not needed and have made that change

PR #64867, opened 2023-04-05
Drag heavier vehicles take 2

anoobindisguise:
#### Summary
Balance "player can drag heavier vehicles"

#### Purpose of change
Continue work on #63737

#### Describe the solution
Changes from the first PR:
- The strength requirement to drag a vehicle off road is now adjusted by the vehicle's off roading capability. This may mean a shopping cart will just get stuck if dragged onto grass.
- There is a strong stamina cost associated with moving a vehicle, given an athletic human has about 10k stamina, and are dragging a vehicle that has an appropriate weight for your strength, you can drag a vehicle for about 2 minutes straight. This could be increased further to a higher amount of stamina cost.
- Time to drag a heavy vehicle increases a lot. At a matching str requirement and your strength, you drag a thing about 20% of your normal speed.
- Removed the max strength limit as it is unnecessary, your own strength will set limits for this naturally. Either it does nothing because you weren't strong enough for it to matter anyways or it artificially prevents you from dragging something you are strong enough to drag. 

- Higher stamina cost.
- More qualifiers to reduce your ability to drag stuff.

#### Testing
Tried an 8 strength character dragging two different vehicles. Shopping cart (empty) and car (about 1000kg).
Shopping cart is slower than walking but it's not really tiring. Dragging it offroad makes it super slow but not strenuous.
Dragging a car is possible but very slow, exhausting, and often a little bit painful. It gets even slower off road but is still barely draggable, unless you get it into some bad terrain (such as a patch of tall grass) in which case it can get totally stuck.
Spawned a wrecked car with no wheels and couldn't drag it at all.
The character was unable at all to drag a vehicle that was much heavier (1500kg 4x4 car) but upon boosting strength to 20 it became easy and wasn't any more difficult offroad.

#### Additional context
@kevingranade


fairyarmadillo:
Just giving a quick heads up to make sure that a travois works on grass after this change, as that is the purpose of a travois.

I cannot wait to see people going bovine just so they can be an ox and pull their own cart!

anoobindisguise:
> Just giving a quick heads up to make sure that a travois works on grass after this change, as that is the purpose of a travois.
> 
> I cannot wait to see people going bovine just so they can be an ox and pull their own cart!

Travois might need to be changed into some kind of hybrid storage/wheel with off road capabilities yea

chrispikula:
> Travois might need to be changed into some kind of hybrid storage/wheel with off road capabilities yea

Maybe a 'wear-resistant support lever' wheel type object?  Not sure what the best term to describe it should be.  

It should also get damaged in use and when overloaded, but that's a ways off and beyond the scope of this PR.  Part of me is now also imagining that if it becomes a wheel, that somehow there's going to be a bug with people re-inflating it.

PR #64868, opened 2023-04-05
Apply the welding audit to welding recipes

anoobindisguise:
#### Summary
Balance "apply real world cm values to various welding recipes in game"

#### Purpose of change
Follow up on #60660
The costs in that PR were assumed to be per 1cm of weld (10 seconds, 1cm of welding wire, 20 kj of battery or 1 gram of acetylene (which is one charge) or 8 units of propane (which is 4 grams) or 40kj with the shitty makeshift welder.)
However actual welding recipes were assigned arbitrarily and were usually 1-4cm or so under the new system.

#### Describe the solution
Give "real"(ish) cm values to welding recipes based on what feels right. This is mostly guesswork so some of it may be wrong but it should be relatively accurate. Assigned some nuts_bolts costs to certain vehicle stuff like armor plates as those are usually affixed with a combo of riveting and welding.
It's ~15 seconds per welding_standard to attach something to the vehicle and ~100 seconds per unit of repair (each unit of welding repair represents 10cm because it needs a chunk of steel, so this is done to prevent you from needing equal chunks of steel per cm of weld)

I made arc welders use medium storage batteries instead of medium batteries, which isn't ideal but it is super necessary because you cannot swap batteries midway through vehicle crafts or construction projects and it will be physically impossible to do many crafts without a big enough battery. Currently they allow any battery both for save compatibility reasons (so that your existing medium battery welders don't cause errors on game load) but this should be done away with once we have the infrastructure to just plug stuff into a grid.

Different values.
**I still need to figure out the cost of cutting stuff with a torch.** I've been looking at the issue and while the underlying math is simple (acetylene torch is delivering 40% of its thermal energy to the metal it is welding, and steel needs to hit 1000 c or so to melt, then we should just be able to use the specific heat of steel to determine how much energy is needed) it raises questions of how many KG of steel you need to vaporize to cut through a door. Are you just torching the hinges and weak points? Are you merely weakening the structure rather than completely melting it so that it's more vulnerable to blunt force? I have not been able to find a specific source that says "In order to cut through a bank vault door we needed 50% of the contents of a MC cylinder" or anything specific that would denote how much fuel should be used. Please let me know if you have any information about this.

Edit: I put in values such that charges_per_use represents 1/2 kg of steel to vaporize. A door uses 14 uses to cut through, so it's assumed you're cutting 7 kg of hinges etc currently. This can totally be subject to change.

#### Testing
I compiled and loaded this in game and the recipes seem to be working OK since this is mostly just a gigantic pile of JSON churn.


anoobindisguise:
I can't seem to get those errors to go away so I'll try testing this on local.

anoobindisguise:
It compiles and runs fine with no errors. I don't know why the test is freaking out, I'm pretty sure I fixed it?

Maleclypse:
Can you resolve conflicts and once basic build passes again please post in ready to merge?

Calavera20012:
Hello, I would like to say that imo, these are very good changes, and very much needed. I have to ask, though, from a gameplay standpoint, how do these changes translate to the average survivor and their avaliable resources? I know we'll see a lot of people complain when this gets merged so I've been thinking about the implications of increasing the costs of welding by a factor of 10 or 50, in some cases. I am not extremely familiar with the vehicle system, only speaking as an average player here, but I am interested in attempting to contribute to make these changes integrate better for experimental players, who would see their resource count value 0.1x to 0.02x if they decide to update. This is to be expected in experimental, as some features get tweaked before others. But I think we have to consider the abstractions already made in other areas of the game, like resource avaliability, in both prices and spawning. I just wanted to ask you about some questions and raise some points for discussion:


1.- Should we also consider modifying the amounts these items spawn in? not speaking about adjusting the spawn locations and sources of the welding resources, but the amounts they spawn in. I don't know whether you have made any changes to this or not, so that's why I'm asking.

2.- Also, it might be good to re-evaluate the pricing of these resources in the game. The Refugee Centre is currently the most reliable way for players to access welding equipment. Changes in the usefulness of the material could make it so a change in their in-game cost and avaliable stocks to preserve the overall game economy.

Sorry about the long text, and keep in mind I am not certain my assumptions are correct, nor am I criticizing the effort put into this PR. I am not trying to be an armchair contributor here either. It was something that crossed my mind and something I would be interested in developing further myself, I don't have any JSON PRs under my name but I love the game and just want to try and understand the changes and how I can contribute to this aspect in particular.

data/json/vehicleparts/rams.json
```diff
@@ -5,11 +5,20 @@
...
-      "repair": { "skills": [ [ "mechanics", 3 ] ], "time": "60 m", "using": [ [ "repair_welding_standard", 1 ] ] }
+      "install": {
+        "skills": [ [ "mechanics", 2 ] ],
+        "time": "40 m",
+        "using": [ [ "welding_standard", 200 ], [ "vehicle_bolt_install", 2 ] ]
+      },
+      "removal": {
+        "skills": [ [ "mechanics", 2 ] ],
+        "time": "15 m",
+        "using": [ [ "vehicle_weld_removal", 1 ], [ "vehicle_wrench", 1 ] ]
```
nornagon, line 19: The `vehicle_wrench` requirement doesn't exist. Not sure why built-in tests don't catch this, but it broke the Hitchhiker's Guide 🙃

nornagon: https://github.com/CleverRaven/Cataclysm-DDA/pull/66168

anoobindisguise: oh whoops lmao. thanks for fixing that

kevingranade:

https://discourse.cataclysmdda.org/t/changes-to-vehicle-welding/28799/7

PR #64872, opened 2023-04-05
Move gambesons to their own file

anoobindisguise:
#### Summary
Infrastructure "Gambesons are in their own files"

#### Purpose of change
There are a ton of gambesons and it's only going to get more out of control if more are added (denim gambesons, armor attach points, etc) so these really ought to be put in their own files to make things cleaner.

#### Describe the solution
Put them back in their own files like my original PR did.


data/json/items/armor/gambesons.json
```diff
@@ -0,0 +1,1898 @@
...
+        "encumbrance": 25
+      },
+      {
+        "material": [ { "type": "cotton_quilted", "covered_by_mat": 100, "thickness": 2.0 } ],
+        "covers": [ "leg_l", "leg_r" ],
+        "specifically_covers": [ "leg_draped_l", "leg_draped_r" ],
+        "coverage": 72,
+        "encumbrance": 8,
+        "layers": [ "BELTED" ],
+        "warmth": 20
```
sparr, line 46: This was `"//warmth": 20` in the other file, so effectively commented out. Unclear whether this is leftover from a past feature or intended to support a new feature for this sort of item composition.

anoobindisguise: Yeah, I misunderstood that being something that can work - I'm gonna need to revise this more thoroughly, apologies.

anothersimulacrum: Commented out in: https://github.com/CleverRaven/Cataclysm-DDA/pull/64922

PR #64873, opened 2023-04-05
Remove SMG type from repeating crossbow

tenmillimaster:
#### Summary
Content "Makes Repeating Crossbow not a SMG"

#### Purpose of change
A repeating crossbow is not capable of the hundreds of RPM that the SMG skills is intended to help you mitigate, and is more in line with the RPM that are achievable from some rifles.

#### Describe the solution

Bring the Repeating Crossbow in line with most other crossbows in type by switching it to the rifle type, which is a more reasonable approximation of skillset.

A whole new skill...
a new fantastic skill for you
With me to tell you "no"
you'll cho ko know
that the SMG isn't 
the right skill

#### Testing

Loaded game, Loaded the crossbow. 

#### Additional context
None.

### DIFF
--- a/data/json/items/ranged/crossbows.json
+++ b/data/json/items/ranged/crossbows.json
@@ -534,7 +534,7 @@
     "price": 324000,
     "material": [ "steel", "wood" ],
     "flags": [ "FIRE_TWOHAND", "PRIMITIVE_RANGED_WEAPON", "TRADER_AVOID" ],
-    "skill": "smg",
+    "skill": "rifle",
     "ammo": [ "bolt" ],
     "weight": "3628 g",
     "volume": "2500 ml",


Sergiej-B:
Semiautomatic weapons like MP40 also use this skill.
I think it's part of a bigger issue - is handling a shotgun that much different than a rifle? Would've AKSU used SMG or rifle skill if it was in the game? Why would a long and heavy SMG like Thompson be different from a rifle when firing single shots?
Maybe there should be skills for one-handed, two-handed and heavy guns, with full-auto on all of them handled by a separate skill or a proficiency? Currently, probably it would have been better from a game balance perspective if repeating crossbow was a tool for SMG training. Yes, crossbows and bb guns are not very good for firearm training IRL, but currently in-game that's the biggest reason to use them. Nerfing repeating crossbow, while leaving the rest  as is, is not good game design, especially considering that SMGs are underpowered in comparison to rifles.

thethunderhawk:
I think this makes a lot of sense. However I’ll say, I suspect the Repeating Crossbow was under the SMG skill because IRL they’re fired from the hip. Whoever initially did the stats probably figured “ranged repeating weapons that you fire from the hip” are most alike an SMG.

That said, AFAIK they don’t train you to fire SMGs from the hip IRL even when firing full automatic extended bursts.

PR #64881, opened 2023-04-06
Renders the lighthouse truly safe and removes the lake cabin from the safe place scenario.

### LINKED ISSUE #64758: Safe Place scenario issue regarding Lake Cabin and Lighthouse Island.
### Describe the bug

Despite being a location available for the player to spawn in if they've chosen the Safe Place scenario, the locations of Lake Cabin and Lighthouse Island are populated with swimmer zombies and are thus, not safe. 

### Attach save file

N/A

### Steps to reproduce

1: Make a new character, and set the Scenario as Safe Place.
2: Choose either Lake Cabin or Lighthouse island as starting location
3: locate the zombies, 
4: (Optional) Show them who the boss is around here.

### Expected behavior

Since these are listed as locations in the Safe Place scenario, they are supposed to be safe.

### Screenshots

_No response_

Using the latest experimental build.  2023-04-01-1436 

### Additional context

_No response_

DoctorBoomstick:
#### Summary
Bugfixes "fixes #64758 by removing enemy spawns from the lighthouse, and removing the lake cabin from the safe place scenario."
#### Purpose of change
Issue #64758 pointed out that zombies were spawning at the lighthouse and lake cabin locations, despite them being starting locations for the safe place scenario.
#### Describe the solution
As recommended in the PR itself, I removed zombie spawns from the lighthouse locations, rendering them completely devoid of enemies. Apart from this, I removed the lake cabin from the starting list of possible locations for the safe place scenario, as I felt that zombie spawns within the property made sense and did not deserve removal. In addition, I changed the zombie groups within the latter area to pull from the standard zombie group rather than the beach group, as a bunch of randos tromping about somebody’s private home in their swimming gear didn’t make the greatest amount of sense in the world.
1.	Using EOCs to remove enemies from the scenario location upon game start would be a far more elegant solution; however, I am not aware of how one would go about implementing such an EOC nor if removing enemies is even possible in the first place.
2.	It was suggested to me that a boarded-up or safe variant of these two locations could be created for use by the scenario, which is a solution that might merit further thought.
#### Testing
The PR is a simple line removal, so no issues should be forthcoming.
#### Additional context
consult issue #64758 for the original issue.

### DIFF
--- a/data/json/mapgen/lake_buildings/cabin_lake.json
+++ b/data/json/mapgen/lake_buildings/cabin_lake.json
@@ -95,7 +95,7 @@
           { "item": "dishes_dining", "chance": 10, "repeat": [ 1, 2 ] }
         ]
       },
-      "place_monsters": [ { "monster": "GROUP_BEACH", "x": [ 1, 22 ], "y": [ 1, 22 ], "density": 0.1 } ]
+      "place_monsters": [ { "monster": "GROUP_ZOMBIE", "x": [ 1, 22 ], "y": [ 1, 22 ], "density": 0.1 } ]
     }
   },
   {
@@ -205,7 +205,7 @@
         "d": { "item": "fishing_items", "chance": 5, "repeat": [ 1, 2 ] },
         "M": { "item": "stash_wood", "chance": 100, "repeat": [ 6, 10 ] }
       },
-      "place_monsters": [ { "monster": "GROUP_BEACH", "x": [ 10, 22 ], "y": [ 15, 22 ], "density": 0.1 } ]
+      "place_monsters": [ { "monster": "GROUP_ZOMBIE", "x": [ 10, 22 ], "y": [ 15, 22 ], "density": 0.1 } ]
     }
   },
   {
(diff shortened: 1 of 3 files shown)


Fris0uman:
> as a bunch of randos tromping about somebody’s private home in their swimming gear didn’t make the greatest amount of sense in the world.

But it's a lake house so the rando in swiming suits are the owners, no? Or maybe some squatter there while the actual owner is away but it's still coherent with the place, no?

DoctorBoomstick:
> > as a bunch of randos tromping about somebody’s private home in their swimming gear didn’t make the greatest amount of sense in the world.
> 
> But it's a lake house so the rando in swiming suits are the owners, no? Or maybe some squatter there while the actual owner is away but it's still coherent with the place, no?
fair enough, though I’d personally question the universal constant of the owners always, in every possible iteration of the cabin, being kitted in their swim gear.

Fris0uman:
Yeah I guess that's also a fair point ^^"

PR #57940, opened 2022-05-26
Clothing and Armor Coverage Conversion Part 2

Reuvin:
The coat part of the coverage update.

#### Summary
Balance "Converts most of the armor and clothing still using the old limb system to the new system with sub-limbs, part 2: Cool Coats"

#### Purpose of change
There are still a number of armors making use of the old coverage system that uses the entire limb for coverage data, instead of the sub-limb system. This PR is meant to convert all the armors and clothing I can find that would have varied coverages that still use the old system, and then assigning the coverages to the specific sub-limbs instead.

#### Describe the solution
In part 2, I've gone through all the coats and converted as much of the coverage values as I felt necessary. If I found an item that needed changing, I would pull up a picture of the item in question and then assign coverages to the sub limbs of the body part it covers in accordance with the real item as best as I could.

#### Describe alternatives you've considered
Not doing this and saving myself a LOT of time.

#### Testing
Part 2 is good to go with no errors.

#### Additional context
If I finish more files before this gets merged, I'll add them to this PR and further update the description.

PR #57911, opened 2022-05-25
Clothing and Armor Coverage Conversion Part 1

Reuvin:
#### Summary
Balance "Converts most of the armor and clothing still using the old limb system to the new system with sub-limbs, part 1: Shoe sole boogaloo, and arm armor too"

#### Purpose of change
There are still a number of armors making use of the old coverage system that uses the entire limb for coverage data, instead of the sub-limb system. This PR is meant to convert all the armors and clothing I can find that would have varied coverages that still use the old system, and then assigning the coverages to the specific sub-limbs instead. 

#### Describe the solution
In part 1, I went through all the arm armor and footwear and converted as much of the coverage values as I felt necessary. If I found an item that needed changing, I would pull up a picture of the item in question and then assign coverages to the sub limbs of the body part it covers in accordance with the real item as best as I could. 

#### Describe alternatives you've considered
Not doing this and saving myself a LOT of time. 

#### Testing
Part 1 is good to go without any errors.

#### Additional context
For the moment this is still in progress. There's a LOT of armor and clothing out there that needs to be converted.

As this is ready to merge as is, I’ll instead do the coverage conversion over a few separate PRs instead of all at once. This part will only contain the arm armor and footwear items.

PR #64897, opened 2023-04-06
Clothing and Armor Coverage Conversion Part 3

Reuvin:
#### Summary
Balance "Converts most of the armor and clothing still using the old limb system to the new system with sub-limbs, part 3: Glorious gloves and snazzy smallclothes"

#### Purpose of change
There are still a number of armors making use of the old coverage system that uses the entire limb for coverage data, instead of the sub-limb system. This PR is meant to convert all the armors and clothing I can find that would have varied coverages that still use the old system, and then assigning the coverages to the specific sub-limbs instead. 

#### Describe the solution
In part 3, I went through all the gloves and undergarments (and swimwear but there was just one thing that was off so not much credit there) and converted as much of the coverage values as I felt necessary. If I found an item that needed changing, I would pull up a picture of the item in question and then assign coverages to the sub limbs of the body part it covers in accordance with the real item as best as I could. 

Not doing this and saving myself the time. 

#### Testing
Part 3 is good to go ingame without any errors.

#### Additional context
A continuation of my previous conversions #57911 and #57940

I return from my yearlong hiatus in the void.

### DIFF
--- a/data/json/items/armor/gloves.json
+++ b/data/json/items/armor/gloves.json
@@ -277,7 +277,19 @@
     "material_thickness": 4,
     "environmental_protection": 4,
     "flags": [ "STURDY", "DURABLE_MELEE" ],
-    "armor": [ { "encumbrance": 12, "coverage": 95, "covers": [ "hand_l", "hand_r" ] } ]
+    "armor": [
+      {
+        "encumbrance": 12,
+        "coverage": 95,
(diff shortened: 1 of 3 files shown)


pjf:
Just wanting to say thank you. Stuff like this is a *lot* of work, but it makes the game so much better! ♥

PR #64912, opened 2023-04-07
Added small chance to increase follower npc relations during casual chatting

stubkan:
#### Summary
Balance "Added small chance to increase follower npc relations during casual chatting"

#### Purpose of change

If you have a npc follower without any quests.  There is only one thing available to you to do, to make them like you better.  Nothing.  You just wait a week, not interacting with the npc at all, and it will ever so slowly begin to like you.

This adds an alternative and more realistic interactive pathway toward follower npc relations.  The original method of do-nothing-but-wait is still valid, and this new method is entirely optional.  But it in my opinion adds verisimiltude, since friendships irl are built by communication.

#### Describe the solution

In the function that gives you a morale boost for idle chatting to a npc for 10 mins, it now will give the npc a small chance (50%) of increasing one of the following values - trust, value or (reducing) anger by one point.  After the chat, there is a 6hr cooldown, during which you can't idle chat again.

Staying with do-nothing-but-wait method.

#### Testing

Ran ingame with a npc, recruited him and chatted over several days to observe relations slowly improving and the function appears to work well.

#### Additional context

If opinions on this differ and values / cooldowns, etc should be different, please let me know.   I think npc relations need more hands-on options than there are available.  

The chance breakdown is...  
33% chance of positive trust/value increase
16.5% chance of becoming less angry (or no change, if not angry)
50.5% chance of no change

Screenshots follow; 

Recruited a NPC, checked what his starting opinion was...

image

Chose chitchat for a while option.
image

Chitchat for a while option disappeared, after chitchat was completed.
image

It comes back after 6 hrs - I repeated this 3 times before a npc value increased (trust) from 5 to 6.
image

### DIFF
--- a/data/json/effects.json
+++ b/data/json/effects.json
@@ -146,6 +146,12 @@
   },
   {
     "type": "effect_type",
+    "id": "socialized_recently",
+    "name": [ "Socialized Recently" ],
+    "desc": [ "AI tag for asking to NPCs to socialize.  This is a bug if you have it." ]
+  },
+  {
+    "type": "effect_type",
     "id": "npc_run_away",
     "name": [ "Running Away" ],
     "desc": [ "AI tag to enable NPCs to flee.  This is a bug if you have it." ]
--- a/src/npctalk_funcs.cpp
+++ b/src/npctalk_funcs.cpp
@@ -91,6 +91,7 @@ static const efftype_id effect_lying_down( "lying_down" );
 static const efftype_id effect_npc_suspend( "npc_suspend" );
 static const efftype_id effect_pet( "pet" );
 static const efftype_id effect_sleep( "sleep" );
+static const efftype_id effect_socialized_recently( "socialized_recently" );
 
 static const faction_id faction_no_faction( "no_faction" );
 static const faction_id faction_your_followers( "your_followers" );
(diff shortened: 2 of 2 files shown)


PatrikLundell:
I'd probably re-weight it a bit. Something like 50% chance for an improvement roll result. If the improvement branch is reached, check if angry, and, if so, give that a 50% chance to reduce the value and 25% each for the other two values to be improved. If not angry, give it a 33% chance to improve Trust and 33% to improve Value, with the remaining 34% not changing anything.

The logic for it is that I'd find it reasonable to usually need to reduce the anger before being able to improve the other aspects, while it's more likely overall to improve things if the anger is present, as it's one more factor available for improvement. Of course, weights could be different (as could the logic).

stubkan:
> I'd probably re-weight it a bit. Something like 50% chance for an improvement roll result. If the improvement branch is reached, check if angry, and, if so, give that a 50% chance to reduce the value and 25% each for the other two values to be improved. If not angry, give it a 33% chance to improve Trust and 33% to improve Value, with the remaining 34% not changing anything.
> 
> The logic for it is that I'd find it reasonable to usually need to reduce the anger before being able to improve the other aspects, while it's more likely overall to improve things if the anger is present, as it's one more factor available for improvement. Of course, weights could be different (as could the logic).

I thought that someone would suggest making the chance to decrease the chances, not increase them!

They are quite high already IMO, the base game standard of increasing relations is averaged out at a 0-2 point increase in trust per 1 week.  This has the potential to increase trust/value by 0-2 per day.

PR #64920, opened 2023-04-07
ci: bump actions/checkout to v3

casswedson:
#### Summary
None

#### Purpose of change
v2 runs on a "deprecated" node version, resolve warnings fix fix fix

https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4602053354

also remove `depth` if it's not needed, 1 is default

*also*
bump comment and labeler actions

these are easy, just update versions; no need to worry about something
breaking
no need to make a different pr just for this

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -174,8 +174,10 @@
 "Map / Mapgen":
   - "**/**mapgen**"
   - "**/mapgen**"
+  - "**/mapgen/**"
   - "**/**overmap**"
   - "**/overmap**"
+  - "**/overmap/**"
 
 "Martial Arts":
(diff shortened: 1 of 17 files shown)


.github/labeler.yml
```diff
@@ -174,8 +174,10 @@
 "Map / Mapgen":
   - "**/**mapgen**"
   - "**/mapgen**"
+  - "**/mapgen/**"
   - "**/**overmap**"
   - "**/overmap**"
+  - "**/oevermap/**"
```
irwiss, line 180: typo or intentional?

casswedson: big typo; good catch

PR #64924, opened 2023-04-07
Keybindings screen mouse support

katemonster33:
#### Summary
Features "Added full mouse support to the keybindings screen"

#### Purpose of change
Per #58106 , adding mouse support to the keybindings screen

#### Describe the solution
The "legend" at the top of the screen that showed the user what buttons to press to use the screen, now doubles as buttons that can be highlighted and clicked in order to initiate adding or removing of keybindings.

Locations of the buttons on the screen is still up for debate. I am not sure buttons being at the top of the screen is intuitive enough.

#### Testing
I launched the screen and verified the buttons work and that you can still change keybindings using the keyboard.


src/input.cpp
```diff
@@ -1427,20 +1432,20 @@ action_id input_context::display_menu( const bool permit_execute_action )
 
     // colors of the keybindings
     static const nc_color global_key = c_light_gray;
+    static const nc_color h_global_key = h_light_gray;
     static const nc_color local_key = c_light_green;
+    static const nc_color h_local_key = h_light_green;
     static const nc_color unbound_key = c_light_red;
+    static const nc_color h_unbound_key = h_light_red;
+
+    enum { no_btn, remove_btn, add_local_btn, add_global_btn } highlighted_btn_index = no_btn;
```
leemuar, line 1441: Scoped enum would be better in general

katemonster33: ok fair enough, done.

PR #64926, opened 2023-04-08
Add percentage-translated statistic to language selection

jbytheway:
#### Summary
I18N "Add percentage-translated statistic to language selection"

#### Purpose of change
Some languages in CDDA are much more completely translated than others.  When selecting a language in the options menu there's no indication of how complete the translation is.

Providing some indication would help people who speak multiple languages choose which might work best for them.  It might also manage expectations and reduce the disappointment for languages where the translation is very minimal.

I would hope it might also inspire some players to help contribute to translation.

#### Describe the solution
In the options menu, when selecting a language, each is now annotated with the percentage of strings which are translated for that language.

These percentages are computed from language stats encoded in a new file, `src/lang_stats.inc`, which is a fragment included in `src/lang_stats.cpp`.  The `lang_stats.inc` is updated by `lang/merge_po.sh`, the same script which updates the `.po` files.  I'm hoping this updating will fit easily into the existing translation update workflow, so should be updated with no extra effort.

I took the opportunity to refactor `merge_po.sh` slightly to reduce code duplication, and fixed a couple of typos I found in other translation-related files.

Rather than having these stats hardcoded, I considered computing them on the fly by parsing the `.mo` files.  But currently those files are only parsed for the languages needed at runtime; it could be rather expensive to parse them for all languages every time the options menu is opened.

#### Testing
Looked at the new interface; verified that switching language still works.

#### Additional context
Currently the annotation is a simple parenthesized percentage suffix (see screenshot below).  I'm not sure if the meaning is sufficiently self-evident.  I also don't know if other languages do not conventionally present fractions as percentages, and this might be entirely unclear.

I welcome any feedback on how to better present the information.  I could add an explanation in English, but there's not much space, so that might entail improving how the top panel of the options menu gets sized.

image


src/lang_stats.cpp
```diff
@@ -0,0 +1,17 @@
+#include <initializer_list>
+#include <lang_stats.h>
+
+static constexpr std::initializer_list<lang_stats> all_lang_stats = {
+#include <lang_stats.inc>
+};
```
andrei8l, line 6: GCC 12 doesn't like this.
```gdb
../cataclysm-dda-git/src/lang_stats.cpp:32:1: error: modification of ‘<temporary>’ is not a constant expression
   32 | };
      | ^

```

`static constexpr std::array<lang_stats,27> all_lang_stats = {` works though, with `lang_stats{"ar", 5, 85479},` and so on.

src/options.cpp
```diff
@@ -1302,36 +1303,48 @@ std::vector<options_manager::id_and_option> options_manager::get_lang_options()
...
+            { "uk_UA", R"(український)" },
+            { "zh_CN", R"(中文 (天朝))" },
+            { "zh_TW", R"(中文 (台灣))" },
+        }
+    };
+
+    for( const auto& [lang_id, lang_name] : language_names ) {
+        std::string name = lang_name;
+        if( const lang_stats *stats = lang_stats_for( lang_id ) ) {
+            name += string_format( _( " (%.1f%%)" ), stats->percent_translated() );
```
andrei8l, line 1343: Some color or better formatting would be nice here. This is just a quick example:
```suggestion
            name += string_format( colorize( _( " (%.1f%%)" ), c_dark_gray ), stats->percent_translated() );
```
<details>
<summary>screenshot</summary>

Screenshot from 2023-04-08 06-30-18

</details>

But this needs either a change in order of initialization (`init_colors()` before `get_options().init()`) or a manual color tag.

Zireael07:
Hmm, I had no idea that the completion level was SO low for some... Why are they even offered as options? :thinking:   Most software will not allow you to pick a language that is below some completion threshold...

Qrox:
IIRC the translation workflow no longer uses merge_po.sh

src/lang_stats.cpp
jbytheway: Thanks.  I found a different workaround that doesn't require explicitly hardcoding the number of languages.

jbytheway:
> Hmm, I had no idea that the completion level was SO low for some... Why are they even offered as options? thinking Most software will not allow you to pick a language that is below some completion threshold...

I was also surprised.  With these stats we could now automatically do something similar (skip languages with very few translated strings) if we wanted.  But that's out of scope for this PR.

jbytheway:
> IIRC the translation workflow no longer uses merge_po.sh

Does anyone know how it does work?  I looked in the GitHub workflows but didn't see anything promising.

Based on who makes the PRs updating .po files, I will ping @kevingranade.

jbytheway:
I followed @andrei8l's suggestion of adding colour, and added an extra sentence of clarification, and increased the number of rows in the option tooltip section by one to make everything fit.  It now looks like this:
language-stats-with-color

src/options.cpp
jbytheway: Good idea.  I opted for a manual color tag since I'm too scared to rearrange initialization order.

Qrox:
> > IIRC the translation workflow no longer uses merge_po.sh
> 
> Does anyone know how it does work? I looked in the GitHub workflows but didn't see anything promising.
> 
> Based on who makes the PRs updating .po files, I will ping @kevingranade.

You can probably change compile_mo.sh, based on this line from the release action:

https://github.com/CleverRaven/Cataclysm-DDA/blob/b8a29ea8c27a7f8a9378288e8dd96512b7b69b2b/.github/workflows/release.yml#L257

Qrox:
It seems the script did not run properly and outputed -1 for number of strings in b4113c1caffa239d4a71cf3c32d4f099025af36a.

PR #64939, opened 2023-04-08
Amalgamations no longer avoid player made traps

Nebnis:
#### Summary
Balance "All amalgamations no longer avoid player made traps"

#### Purpose of change
According to the original amalgamation PR, they are supposed to be discardable drones revived from biological mass. It doesn't make sense that they can avoid player made traps while i can concede that they will avoid fire,holes like most common animals. Even beside that they are the only monster group in the game with avoid_danger 2 other than the godly being Yrax. 

#### Describe the solution
Changed PATH_AVOID_DANGER_2 to PATH_AVOID_DANGER_1

There are a specific avoid trap json now, so if there is anything specific they would avoid, it could be avoided now.

#### Testing

Logged, they don't avoid nailboard traps but won't jump into fire or holes.

### DIFF
--- a/data/json/monsters/zed_amalgamation.json
+++ b/data/json/monsters/zed_amalgamation.json
@@ -112,8 +112,8 @@
     "families": [ "prof_wp_amalgamation" ],
     "harvest": "zombie_amalgamation",
     "//": "TODO: When effects/spells can mess with path settings + flags move smartness to the effect of dedicated coordinatiors",
-    "path_settings": { "avoid_sharp": true, "avoid_traps": true, "max_dist": 400 },
-    "flags": [ "SEES", "SMELLS", "HEARS", "PATH_AVOID_DANGER_1", "REVIVES", "NO_BREATHE" ]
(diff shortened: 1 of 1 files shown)


Karol1223:
@Venera3 has blatantly said on discord that this was intentional and that if you wanted to change this you should not do this by changing the flag but by implementing a way for them to avoid dangerous fields while not avoiding man-made traps.

They have this flag so they can avoid dangerous fields other than fire. Them avoiding traps is a side effect.

Venera3 (changes requested):
Traps are not the problem, dangerous fields are. That needs to change if you want this in.

Nebnis:
> Traps are not the problem, dangerous fields are. That needs to change if you want this in.

They still avoid dangerous fields, they won't walk into fire or holes.

Venera3:
Fire is included in avoid_1, acid isn't. I stand by the Discord stance of "add trap/field differentiation or smartening effects and we can talk", going through three different instances of the same discussion is not going to change that.

PR #64943, opened 2023-04-08
Activities can fire EOCs

bombasticSlacks:
#### Summary
Features "Activities can fire EOCs"

#### Purpose of change
Wanted to be able to define an activity entirely in JSON, now you can.

#### Describe the solution
Added code for activities to define a do_turn_eoc and a completion_eoc. Each triggers when you would expect.

#### Testing
Unit tests have been added.

### DIFF
--- a/data/mods/TEST_DATA/EOC.json
+++ b/data/mods/TEST_DATA/EOC.json
@@ -55,5 +55,20 @@
     "recurrence": { "math": [ "math_test_result" ] },
     "global": true,
     "effect": { "u_message": "test recurrence" }
+  },
+  {
+    "id": "EOC_increment_var",
+    "//": "tests adding a variable to the player",
+    "effect": [
+      {
+        "arithmetic": [
+          { "u_val": "var", "var_name": "activitiy_incrementer" },
+          "=",
+          { "u_val": "var", "var_name": "activitiy_incrementer" },
+          "+",
+          { "const": 1 }
+        ]
+      }
+    ]
   }
 ]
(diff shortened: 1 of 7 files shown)


mqrause:
Unless I'm mistaken this only works for legacy activities and not activity actors.

bombasticSlacks:
> Unless I'm mistaken this only works for legacy activities and not activity actors.

My understanding of activity actors is that they require a c++ class. Since the goal of this is to all be JSON, yeah it works with the legacy stuff.

mqrause:
I don't exactly see why hooking it up with activity actors would be a bad thing, considering the activities they're used in use the same json definitions. Plus legacy activities are meant to be phased out, and this just adds more dependency on them which hinders that process. Implementing it in `player_activity::do_turn` should be a better option to do this. Then you can still leave figuring out jsonified activity actors for later.

bombasticSlacks:
>I don't exactly see why hooking it up with activity actors would be a bad thing, considering the activities they're used in use the same json definitions. Plus legacy activities are meant to be phased out, and this just adds more dependency on them which hinders that process. Implementing it in player_activity::do_turn should be a better option to do this. Then you can still leave figuring out jsonified activity actors for later.

I see your point, yeah I can move it out of the legacy functions.

data/mods/TEST_DATA/EOC.json
```diff
@@ -55,5 +55,21 @@
...
+      {
+        "arithmetic": [
+          { "u_val": "var", "var_name": "activitiy_incrementer" },
+          "=",
+          { "u_val": "var", "var_name": "activitiy_incrementer" },
+          "+",
+          { "const": 1 }
+        ]
+      }
+    ]
```
andrei8l, line 73:
```suggestion
    "effect": [ { "math": [ "u_activitiy_incrementer", "++" ] } ]
```

PR #64945, opened 2023-04-08
Add a rocky obstacle

NetSysFire:
#### Summary
Content "New construction: rocky obstacle"

#### Purpose of change

While playtesting, I got some large rocks from digging pits to fortify my base. And I did wonder if they could have some other use.

#### Describe the solution

Adds construction and terrain for the rocky obstacle. Note that this is different to rubble, which is more chaotic in nature and needs more energy and time to clear.

Could certainly be great for hastily constructed survivor camps, both player and mapgen-made.

Looking at rubble, I am unsure of my move_cost value. I was briefly thinking about letting payers construct rocky rubble, but pits have the same move cost value, making that both cheaper and easier to come by.

#### Testing

Loaded into the game locally, looks as expected.

### DIFF
--- a/data/json/construction.json
+++ b/data/json/construction.json
@@ -7840,7 +7840,7 @@
     "required_skills": [ [ "survival", 1 ] ],
     "qualities": [ [ { "id": "DIG", "level": 1 } ] ],
     "time": "5 m",
-    "components": [ [ [ "rock_large", 6 ] ] ],
+    "components": [ [ [ "rock_large", 22 ] ] ],
     "pre_terrain": "t_dirt",
     "post_terrain": "t_rockyobstacle"
   }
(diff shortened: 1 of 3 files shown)


data/json/construction.json
```diff
@@ -7831,5 +7831,17 @@
...
+  },
+  {
+    "type": "construction",
+    "id": "constr_obstacle_rock",
+    "group": "obstacle_rock",
+    "category": "CONSTRUCT",
+    "required_skills": [ [ "survival", 1 ] ],
+    "qualities": [ [ { "id": "DIG", "level": 1 } ] ],
+    "time": "5 m",
+    "components": [ [ [ "rock_large", 6 ] ] ],
```
anothersimulacrum, line 7843: I feel like this should be more than 36L of rocks, especially given a movecost of 4.

NetSysFire: How many would you use?

anothersimulacrum: 24 or so - assuming they're perfect spheres, should be about enough to cover the whole ground in rocks.

NetSysFire: I decided for 22, to make it uneven enough to slow people down. Also, me no good at math.

PR #64956, opened 2023-04-09
Update COMPILING.md

alef:
#### Summary
None

#### Purpose of change
Clean-up this document

#### Describe the solution
- Add unsupported `COMPILING-CMAKE-VCPKG.md`
- Fix links
- Remove backticks from one-liner to avoid shell command substitution
- Add `\` to long one-liner
- Remove `$`/`#` prompts to allow one-click copy

Maybe this document should be split


sparr:
Check failures appear to be caused by an unrelated json problem that was merged and fixed around the time this PR was created.

PR #64961, opened 2023-04-09
Bullet damage by barrel length

bombasticSlacks:
#### Summary
None

#### Purpose of change
Makes it so bullets can define damage per length of barrel, and weapons, and mods can define barrel length as well.

#### Describe the solution
Damage instances can now have a barrel_desc object that defines amounts by length,
Guns can have barrel lengths,
Barrel mods with an integral length can overwrite the guns length,
When guns calc damage, they consider their barrel length if they have one, and lerp the right value,
Inherited damage values for bullets are also propogated through the barrel description array for stuff like black poweder.

Plenty of this is now displayed in the UI as shown in the screenshot bellow.

#### Testing
Using some mock remington values rn

#### Additional context
image


src/damage.h
```diff
@@ -39,6 +40,13 @@ struct enum_traits<damage_type> {
     static constexpr damage_type last = damage_type::NUM;
 };
 
+struct barrel_desc {
+    units::length barrel_length;
+    float amount;
+
+    barrel_desc( units::length bl, float amt ) : barrel_length( bl ), amount( amt ) {}
```
anothersimulacrum, line 47:
```suggestion
    barrel_desc( units::length bl, float amt ) : barrel_length( bl ), amount( amt ) {}
    void deserialize( const JsonObject &jo ) {
            mandatory( jo, false, "barrel_length", barrel_length );
            mandatory( jo, false, "amount", amout );
    }
```

src/damage.cpp
```diff
@@ -410,7 +440,22 @@ static damage_unit load_damage_unit( const JsonObject &curr )
...
+
+    if( curr.has_array( "barrels" ) ) {
+        for( const JsonObject &barrel : curr.get_array( "barrels" ) ) {
+            units::length bl;
+            float amt;
+            mandatory( barrel, false, "barrel_length", bl );
+            mandatory( barrel, false, "amount", amt );
+            du.barrels.emplace_back( bl, amt );
+        }
+    }
```
anothersimulacrum, line 456:
```suggestion
    optional( curr, false, "barrels", du.barrels );
```

anothersimulacrum (commented):
A little cleaner

src/damage.h
bombasticSlacks: Your absolutely right, thanks I'll incorporate when I get a chance.

RenechCDDA:
Does this gel with the (somewhat recently) updated 'rules' for firearm damage? https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/GAME_BALANCE.md#ammo-stats

That seems to want to put damage modifiers *on the gun itself*, is this meant to supplant it? It *would* be easier to derive the values themselves in the cartridge so that each and every gun doesn't need to be set manually outside of defining the barrel length.

bombasticSlacks:
> That seems to want to put damage modifiers on the gun itself, is this meant to supplant it? It would be easier to derive the values themselves in the cartridge so that each and every gun doesn't need to be set manually outside of defining the barrel length.

That's what I was thinking, which is why I'm working in parallel with the person who updated the docs with those details 😄

tenmillimaster:
> Does this gel with the (somewhat recently) updated 'rules' for firearm damage? https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/GAME_BALANCE.md#ammo-stats
> 
> That seems to want to put damage modifiers *on the gun itself*, is this meant to supplant it? It *would* be easier to derive the values themselves in the cartridge so that each and every gun doesn't need to be set manually outside of defining the barrel length.

This will supplant that system. I'm working on regenerating the dataset (I have a large dataset of simulated projectile speeds and barrel lengths) and skipping the regression entirely. This would increase accuracy and simplify the reviewing and contributing process.

Many thanks to Slacks.

src/damage.h
```diff
@@ -39,6 +40,17 @@ struct enum_traits<damage_type> {
...
+    void deserialize( const JsonObject &jo ) {
```
anothersimulacrum, line 48: Probably best to stick this in the .cpp file, so we don't have to put generic_factory.h in a header, whoops.

src/item.cpp
```diff
@@ -10604,26 +10674,36 @@ itype_id item::ammo_default( bool conversion ) const
...
-        } );
+        if( gun ) {
+            return enumerate_as_string( type->magazine->cached_ammos[at],
+            [gun]( const itype_id & id ) {
+                return string_format( "<info>%s(%d)</info>", id->nname( 1 ),
+                                      static_cast<int>( gun->gun_damage( id ).total_damage() ) );
+            } );
+        } else {
+            return enumerate_as_string( type->magazine->cached_ammos[at],
+            [this]( const itype_id & id ) {
```
bombasticSlacks, line 10696:
```suggestion
            []( const itype_id & id ) {
```

tests/ammo_test.cpp
```diff
@@ -2,13 +2,16 @@
...
+#include "damage.h"
 #include "item.h"
 #include "type_id.h"
 
 static const ammotype ammo_762( "762" );
 static const ammotype ammo_9mm( "9mm" );
 static const ammotype ammo_battery( "battery" );
 
+static const itype_id itype_100mm_ammo( "test_100mm_ammo" );
+static const itype_id itype_100mm_ammo_relative( "test_100mm_ammo_relative" );
```
bombasticSlacks, line 14:
```suggestion
static const itype_id itype_test_100mm_ammo_relative( "test_100mm_ammo_relative" );
```

tests/ammo_test.cpp
```diff
@@ -2,13 +2,16 @@
...
 
+#include "damage.h"
 #include "item.h"
 #include "type_id.h"
 
 static const ammotype ammo_762( "762" );
 static const ammotype ammo_9mm( "9mm" );
 static const ammotype ammo_battery( "battery" );
 
+static const itype_id itype_100mm_ammo( "test_100mm_ammo" );
```
bombasticSlacks, line 13:
```suggestion
static const itype_id itype_test_100mm_ammo( "test_100mm_ammo" );
```

tests/ammo_test.cpp
```diff
@@ -196,3 +199,24 @@ TEST_CASE( "ammo default", "[ammo][ammo_default]" )
...
+        item base_gun( "test_glock_super_long" );
+        CHECK( base_gun.gun_damage( itype_100mm_ammo ).total_damage() == 65 );
+    }
+
+    SECTION( "basic ammo and mod length test" ) {
+        item base_gun( "test_glock_super_long" );
+        item gun_mod( "barrel_glock_short" );
+        REQUIRE( base_gun.put_in( gun_mod, item_pocket::pocket_type::MOD ).success() );
+        REQUIRE( base_gun.barrel_length().value() == 100 );
+        CHECK( base_gun.gun_damage( itype_100mm_ammo ).total_damage() == 60 );
```
bombasticSlacks, line 214:
```suggestion
        CHECK( base_gun.gun_damage( itype_test_100mm_ammo ).total_damage() == 60 );
```

tests/ammo_test.cpp
```diff
@@ -196,3 +199,24 @@ TEST_CASE( "ammo default", "[ammo][ammo_default]" )
...
+    }
+
+    SECTION( "inherited ammo and barrel length test" ) {
+        item base_gun( "test_glock_super_long" );
+        CHECK( base_gun.gun_damage( itype_100mm_ammo_relative ).total_damage() == 66 );
```
bombasticSlacks, line 219:
```suggestion
        CHECK( base_gun.gun_damage( itype_test_100mm_ammo_relative ).total_damage() == 66 );
```

tests/ammo_test.cpp
```diff
@@ -196,3 +199,24 @@ TEST_CASE( "ammo default", "[ammo][ammo_default]" )
     }
 }
 
+TEST_CASE( "barrel test", "[ammo][weapon]" )
+{
+    SECTION( "basic ammo and barrel length test" ) {
+        item base_gun( "test_glock_super_long" );
+        CHECK( base_gun.gun_damage( itype_100mm_ammo ).total_damage() == 65 );
```
bombasticSlacks, line 206:
```suggestion
        CHECK( base_gun.gun_damage( itype_test_100mm_ammo ).total_damage() == 65 );
```

PR #64965, opened 2023-04-09
fix: item spawning spells with no min/max damage

casswedson:
#### Summary
None

#### Purpose of change
if a spell with the effect `spawn_item` does not have a defined min/max damage it will not spawn anything at any level (given it doesn't have something like damage increment)

setting the values manually just makes sense, just in case, so it always spawns something at level 1

scraped a bunch of data in search of spells that spawn items and do not have min/max damage defined, there were some

purpose of change:
how _dare_ you make my favorite spell repelling arc useless as it does not spawn anything >:(


mqrause:
> how _dare_ you make my favorite spell repelling arc useless as it does not spawn anything >:(

Oops, that's an unintended side effect of  #64600. Instead of adding damage to all spells a small fix for this loop might be the better solution.
https://github.com/CleverRaven/Cataclysm-DDA/blob/e9ab8341f0b7964ade6e72c3ad6025a0f6b15e80/src/magic_spell_effect.cpp#L1071

PR #64971, opened 2023-04-10
Migrate some JSON APIs to string_view

jbytheway:
#### Summary
Infrastructure "Migrate some JSON APIs to string_view"

#### Purpose of change
Now we're on C++17 we have access to `std::string_view`.  This allows simplification of some APIs.  In particular, cases where we have previously had separate `char*` and `std::string` overloads can now be merged into one function.

There are quite a few such overload sets in the JSON code.

#### Describe the solution
Migrate a bunch of the JSON code and associated stuff to `string_view`.

This is just an initial effort; it's supposed to be a minimum viable initial step.

Note that the conversion is quite viral; it led me to convert all the APIs used by the JSON code to string_view too.

Also, fixed a random unrelated typo in the tests.

I'm using the convention that `string_view` should be passed by value, since it can be passed in registers on some platforms.  However, IIRC that's not true on Windows, so we might actually want to pass by `const string_view&` for best performance on Windows.  Still, it shouldn't make much difference either way, and this option is less verbose.

#### Testing
Unit tests.

#### Additional context
I plan to migrate more stuff to `string_view`.  This PR is partly to test the waters for how much appetite there is to do that.

### DIFF
--- a/src/flexbuffer_json.h
+++ b/src/flexbuffer_json.h
@@ -10,6 +10,7 @@
 
 #include "cata_bitset.h"
 #include "cata_small_literal_vector.h"
+#include "cata_utility.h"
 #include "flexbuffer_cache.h"
 #include "json.h"
 #include "json_error.h"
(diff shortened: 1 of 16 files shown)


src/flexbuffer_json.h
```diff
@@ -672,27 +647,19 @@ class JsonObject : JsonWithPath
         double get_float( const char *key, double fallback ) const;
```
andrei8l, line 647: Migrate these accessors too

jbytheway: I had planned to do those in a pater PR, but I guess since they are sandwiched between two other groups I migrated I should probably do them now.

kevingranade:
Lots of appetite for string view,  especially in json land.

I was going to ask about perf,  but then I realized all of these are going to be constructing a view from a const string so it's unlikely to make any difference yet.

Issue #64555, opened 2023-03-25
[Magiclysm] Magical auras destroyed by fire


Terrorforge:
### Describe the bug

Various magical auras, like Aura of Protection and Force Shield, can catch on fire and be destroyed. This is nonsensical, as these items represent intangible force fields and other magical effects that it shouldn't be possible to destroy through mundane damage. It also creates some gameplay problems, like Demon Spiders, an already very dangerous enemy, being able to instantly negate your most potent defenses.

This seems to affect items that are made of "magical material", but not ones made of "concentrated mana". Mage Armor isn't destroyed by fire, for example.

### Attach save file

trivially reproducible

### Steps to reproduce

1. Debug raise your strength to a couple hundred so you don't instantly die
2. Cast Aura of Protection and/or Force Shield
3. Spawn a demon spider
4. Wait for it to cast Burning Hands

### Expected behavior

Magical auras don't catch on fire and burn like mundane items.

### Screenshots

![image](https://user-images.githubusercontent.com/5107449/227710215-8c3177ad-8796-4bb1-948e-ed954aaa38ae.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.2728 (22H2)
- Game Version: 4ad1388 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Magiclysm [magiclysm],
    Mythical Martial Arts [MMA]
]

### Additional context

I tried fixing this by adding the "unbreakable" flag, but apparently that doesn't protect the item from burning, so I don't know how to fix this.

GuardianDll:
Probable fix: go to https://github.com/CleverRaven/Cataclysm-DDA/blob/66012021564e005fef9aaff3cdc713428abec3fd/data/mods/Magiclysm/materials.json#L194 and put `"burn_data": [ ]` to overwrite the cotton's inherited burn data

Terrorforge:
Nope, doesn't work. So far, the only fix I've found is giving the material a high fire_resist. But then if the garment has any amount of environmental protection, it also grants the player substantial fire resistance, so something liek Aura of Protection wouldn't really work.

Maleclypse:
[...]

PR #64979, opened 2023-04-10
Prevent fire damaging unbreakable items

Rewryte:
#### Summary
Bugfixes "Prevent fire damaging unbreakable items"

#### Purpose of change
Items flagged with "UNBREAKABLE" can be damaged by fire in two ways. Either by wearing them and getting set on fire, or by directly dropping the item into fd_fire. This might not be obvious because most "UNBREAKABLE" items are also "INTEGRATED" which, unlike the former flag, actually works to protect worn equipment from fire damage.

#### Describe the solution
The "item::burn" and "outfit::absorb_damage" functions will now check for the "UNBREAKABLE" flag and exempt flagged items from sustaining damage.

In addition to these changes, we can also remove damage immunity from "INTEGRATED" flagged items.  This opens the possibility of repairable bionic parts, but safeguards should be put in place to prevent them from being completely destroyed. Integrated parts with the "UNBREAKABLE" flag should retain their original invulnerability.

#### Testing
The "aura of protection" from the Magiclysm mod was modded to be "UNBREAKABLE" and unequipable. Its coverage and location in the outer aura slot makes it particularly vulnerable to fire
A character is given the aura of protection
fd_fire with intensity 1 is spawned next to the character
The character walks into fd_fire and waits for a few rounds
The aura of protection is undamaged (before the changes it would have been completely burnt)
The aura of protection is dropped into the fire
Waited a few more rounds
The aura on the burning ground remains undamaged

#### Additional context
Can be considered an intermediate step in addressing #64555 which I will work on after this.

### DIFF
--- a/src/character_attire.cpp
+++ b/src/character_attire.cpp
@@ -1759,7 +1759,8 @@ void outfit::absorb_damage( Character &guy, damage_unit &elem, bodypart_id bp,
         if( outermost && elem.type == damage_type::HEAT && elem.amount >= 1.0f ) {
             // TODO: Different fire intensity values based on damage
             fire_data frd{ 2 };
-            destroy = !armor.has_flag( flag_INTEGRATED ) && !armor.has_flag( flag_UNBREAKABLE ) && armor.burn( frd );
+            destroy = !armor.has_flag( flag_INTEGRATED ) && !armor.has_flag( flag_UNBREAKABLE ) &&
+                      armor.burn( frd );
             int fuel = roll_remainder( frd.fuel_produced );
             if( fuel > 0 ) {
                 guy.add_effect( effect_onfire, time_duration::from_turns( fuel + 1 ), bp, false, 0, false,
--- a/src/character_attire.cpp
+++ b/src/character_attire.cpp
@@ -1759,8 +1759,7 @@ void outfit::absorb_damage( Character &guy, damage_unit &elem, bodypart_id bp,
         if( outermost && elem.type == damage_type::HEAT && elem.amount >= 1.0f ) {
             // TODO: Different fire intensity values based on damage
             fire_data frd{ 2 };
-            destroy = !armor.has_flag( flag_INTEGRATED ) && !armor.has_flag( flag_UNBREAKABLE ) &&
-                      armor.burn( frd );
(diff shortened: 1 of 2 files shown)


src/character_attire.cpp
```diff
@@ -1759,7 +1759,8 @@ void outfit::absorb_damage( Character &guy, damage_unit &elem, bodypart_id bp,
         if( outermost && elem.type == damage_type::HEAT && elem.amount >= 1.0f ) {
             // TODO: Different fire intensity values based on damage
             fire_data frd{ 2 };
-            destroy = !armor.has_flag( flag_INTEGRATED ) && armor.burn( frd );
+            destroy = !armor.has_flag( flag_INTEGRATED ) && !armor.has_flag( flag_UNBREAKABLE ) &&
+                      armor.burn( frd );
```
irwiss, line 1763: This check likely shouldn't be here, if item::burn(...) below checks for UNBREAKABLE

src/item.cpp
```diff
@@ -11252,7 +11252,7 @@ bool item::burn( fire_data &frd )
 {
     float burn_added = simulate_burn( frd );
 
-    if( burn_added <= 0 ) {
+    if( burn_added <= 0 || has_flag( flag_UNBREAKABLE ) ) {
```
irwiss, line 11255: this check should be split off and pulled up into first line of the function because simulate_burn() mutates `frd` struct intensity/smoke etc for stuff that wasn't actually burnt

src/character_attire.cpp
Rewryte: Yeah, thought having the flags together would make it easier to phase out INTEGRATED invulnerability in the future, but it is unnecessarily redundant. I'll get rid of it.

src/item.cpp
Rewryte: Resolved

src/item.cpp
```diff
@@ -11250,6 +11250,10 @@ float item::simulate_burn( fire_data &frd ) const
 
 bool item::burn( fire_data &frd )
 {
+    if( has_flag( flag_UNBREAKABLE ) ) {
+        return false;
+    }
+
```
dseguin, line 11256:
```suggestion

```

PR #64984, opened 2023-04-10
math_parser: add more common math functions

andrei8l:
#### Summary
None

#### Purpose of change
Add more common math functions requested by contributors

#### Describe the solution
Add clamp, floor, trunc, ceil, and round

Also fix the argument order 🤦

N/A

#### Testing
```JSON
  {
    "id": "EOC_math_funcs",
    "effect": [ 
      { "math": [ "u_blorg", "=", "floor(1.4) + ceil(1.4) + trunc(-1.4) + clamp(1, 2, 3) + round(1.5)" ] },
      { "u_message": "<u_val:blorg>" }
    ]
  },
```
should output 6

#### Additional context
N/A

### DIFF
--- a/doc/NPCs.md
+++ b/doc/NPCs.md
@@ -1368,7 +1368,15 @@ The tokens `π` (and `pi` alias ) and `e` are recognized as mathematical constan
 #### Math functions
 Common math functions are supported: 
 
-`abs()`, `sqrt()`, `log()`, `sin()`, `cos()`, and `tan()` take one argument, for example `sin( 2 * pi + 1 )`.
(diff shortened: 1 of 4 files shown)


Ramza13:
Great stuff!  That said I can guess but I'm not actually sure exactly what ceiling or floor do, probably a good idea to document all of the new functions to be safe.

andrei8l:
> Great stuff! That said I can guess but I'm not actually sure exactly what ceiling or floor do, probably a good idea to document all of the new functions to be safe.

I've added some descriptions. Let me know if anything else in not clear.

PR #64986, opened 2023-04-10
Do not set charges to -1 on items that are not count by charges when filling a pocket

### LINKED ISSUE #64929: UPS weapons drain battery when in inventory
### Describe the bug

UPS weapons drain battery at a rate of 1 unit per turn when wielded and then stored to inventory. Also affects UPS CBM.

### Attach save file

[Tappen-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11182597/Tappen-trimmed.tar.gz)

### Steps to reproduce

1. Spawn any UPS weapon, a UPS, and a heavy battery.
2. Reload UPS with battery.
3. Wield UPS weapon and then store in inventory.
4. Watch as battery is drained each turn.

### Expected behavior

UPS weapons don't drain UPS battery when not in use.

### Screenshots

_No response_

- OS: Android
    - OS Version: Manufacturer: HUAWEI; Model: DBY-W09; Release: 10; Incremental: 102.0.0.277C185; 
- Game Version: 714aa12 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

Possibly related to #63978.

Hirmuolio:
#### Summary
Bugfixes "Do not set charges to -1 on items that are not count by charges when filling a pocket"

#### Purpose of change

Fix: #64929

When item that was not count by charges was placed in a holster pocket by unwielding it the item would get its charges set to -1.

#### Describe the solution

Keep the item charges as they are.
Move the check that should only apply to items counted by charges so that it does not apply to normal items which almost always have 0 charges.

#### Testing

No longer infinite UPS drain.
Tests pass.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -11621,7 +11621,7 @@ int item::fill_with( const item &contained, const int amount,
 
     item contained_item( contained );
     const bool count_by_charges = contained.count_by_charges();
-    contained_item.charges = count_by_charges ? 1 : -1;
+    contained_item.charges = count_by_charges ? 1 : contained.charges;
     item_location loc;
     item_pocket *pocket = nullptr;
 
@@ -11649,10 +11649,12 @@ int item::fill_with( const item &contained, const int amount,
                                                      pocket->charges_per_remaining_weight( contained_item )
                                                    } );
             }
+
+            if( contained_item.charges == 0 ) {
+                break;
+            }
         }
-        if( contained_item.charges == 0 ) {
-            break;
-        }
+
         if( !pocket->insert_item( contained_item ).success() ) {
             if( count_by_charges ) {
                 debugmsg( "charges per remaining pocket volume does not fit in that very volume" );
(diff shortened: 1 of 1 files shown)


src/item.cpp
```diff
@@ -11621,7 +11621,7 @@ int item::fill_with( const item &contained, const int amount,
 
     item contained_item( contained );
     const bool count_by_charges = contained.count_by_charges();
-    contained_item.charges = count_by_charges ? 1 : -1;
+    contained_item.charges = count_by_charges ? 1 : contained.charges;
```
mqrause, line 11624: You should test that, but it doesn't seem to me like this line is necessary at all. Charges are set to 1 at the beginning of the loop anyway.
And if it's necessary, it should just be
```
if ( count_by_charges ) {
    contained_item.charges = 1;
}
```

Hirmuolio: That is only for items that are counted by charges (ie. stacks of stuff like water).
Most items do not use charges (so it is usually 0). Some items use charges for other things (for example grenade ocuntdown).

PR #64988, opened 2023-04-10
Make refugee center guards move around in shifts.

Ramza13:
#### Summary
None

#### Purpose of change
Request by i-am-erk
Json power + npc more aliveness

#### Describe the solution
Added a EOC new function to allow changing NPC's guard pos and then a new mapgen way to place variables.  Then used them to set up a simple guard rotation.  Right now every 2 hours 7 of the guards move a short distance to a nearby spot and then two hours later move back.

#### Testing
Go to refugee center.  Wait two hours.

#### Additional context
This also makes the refugee center a globally unique spawn.  It won't work otherwise as it requires the guards to have unique_ids.  Pretty sure we wanted that to happen this version anyway.  The existing 'patrols' are quite minimal, this is meant more as a proof of concept than a full fledged living refugee center.


MNG-cataclysm:
Is there some way that could be programmed to make a position variable localized to an overmap tile or a range thereof? If so, we could use this for regular NPC routines for non-unique specials in the future.

You've done a great job with this, thank you.

Ramza13:
> Is there some way that could be programmed to make a position variable localized to an overmap tile or a range thereof? If so, we could use this for regular NPC routines for non-unique specials in the future.
> 
> You've done a great job with this, thank you.

How would you want that defined?  Where in the json?  I'm sure something could be done it just depends on how you would want to define it.

andrei8l:
Instead of relying on unique NPCs, why not add a new member to NPC definition like this
```JSON
    "patrol_points": [ "var1", "var2", "var3" ],
    "patrol_interval": [ rng_min, rng_max ]
```
and assign those variables on the NPC instead of globally.

src/mapgen.cpp
```diff
@@ -491,7 +491,7 @@ static mapgen_factory oter_mapgen;
 
 std::map<nested_mapgen_id, nested_mapgen> nested_mapgens;
 std::map<update_mapgen_id, update_mapgen> update_mapgens;
-
+static std::unordered_map<std::string, std::string> queued_points;
```
andrei8l, line 494:
```suggestion
static std::unordered_map<std::string, tripoint_abs_ms> queued_points;
```
And stringify only in `set_queued_points()`. Then you won't have to convert twice in `map::rotate()`

Ramza13:
> Instead of relying on unique NPCs, why not add a new member to NPC definition like this
> 
> ```json
>     "patrol_points": [ "var1", "var2", "var3" ],
>     "patrol_interval": [ rng_min, rng_max ]
> ```
> 
> and assign those variables on the NPC instead of globally.

The variables are still unique globally so that doesn't solve anything.

andrei8l:
They don't need to be global. And they probably shouldn't be anyway. You're currently rotating the points on any map rotation so if a special spawns rotated after `evac_center`, it'll mess up the patrol points.

Ramza13:
> 
Yes it would be possible to define the patrol the same as monster patrols are.  That approach does have the merit of being easier syntax at the cost of being less powerful.  Right now this patrol is just changing based on time but it could be based on anything, maybe after certain missions or monster attacks the refugee center adds more patrols for example.  Its a tradeoff, you get the power of EOC's but its a little clunkier.  

I'm invested in EOC's so I chose that approach, and having written it I'm honestly tired and don't want to change it up right now.  That said I could change up the syntax to what you propose, I dunno I'm biased as I said.

MNG-cataclysm:
> The variables are still unique globally so that doesn't solve anything.

What if we use an overmap range variable to solve this?

```JSON

    "patrol_points": [ "var1", "var2", "var3" ],

     "patrol_interval": [ rng_min, rng_max ]

"range": [ range_min, range_max ]

```
Range would go by overmap tile, and restrict how far the NPC in question would travel for rotations. A range of 0 restricts them to the overmap tile they're on. If they move out of the overmap tile in the process of attacking a zombie or something, they would travel back to the last spot.

Ramza13:
> Two comments:
> 
> * Have you tested that it works properly when guards are moved around by other causes (such as noise from clearing the back room and eliminating the spy?).
> * I'm not aware of the unique spawn problems having been addressed at all, and so think making the refugee center globally unique is premature.

My understanding is that right now they will behave the same as they do now in terms to their spawn point.  However next shift change they will correctly go to their shift.  So its not any worse than the current version and would requite tweaking NPC ai to fix, something worth doing for sure but not in the scope of this change IMO.

> 
> What if we use an overmap range variable to solve this?
> 
> ```json
> "patrol_points": [ "var1", "var2", "var3" ],
> "patrol_interval": [ rng_min, rng_max ]
> "range": [ range_min, range_max ]
> ```
> 
> Range would go by overmap tile, and restrict how far the NPC in question would travel for rotations. A range of 0 restricts them to the overmap tile they're on. If they move out of the overmap tile in the process of attacking a zombie or something, they would travel back to the last spot.

So the problem here is that right now when we run mapgen for a refugee center a global variable is being saved with the name GUARD1_First.  If you make a second refugee center then it gets overwritten.  A range would stop them from from going too far but then it still means that only one refugee center has a functioning shift change.  

My understanding was that we wanted to start enforcing global uniqueness in this version anyway so this was pulling the bandaid off.

PatrikLundell:
What I was after was verifying they'll actually move to their next point rather than do something undesirable like e.g. try to move relative to where they ended up, or get stuck trying to move through walls.

Just because we'd like to actually making uniques unique during this version doesn't mean we shouldn't make the changes that allow it to work properly first. Things should still be done in the proper order, and if appropriate prerequisites aren't produced (either due to lack of interest or failures of the attempts to work properly), the change should be put off until that's been taken care of (I'm aware there is a significant variation in the opinion of what's sufficient, including those who think the current situation is and was good enough).

MNG-cataclysm:
> So the problem here is that right now when we run mapgen for a refugee center a global variable is being saved with the name GUARD1_First.  If you make a second refugee center then it gets overwritten.  A range would stop them from from going too far but then it still means that only one refugee center has a functioning shift change.

What if we made the global variable a localized one? Is that a fairly easy thing, or would that be too difficult to do for the scope of this PR?

Ramza13:
> What I was after was verifying they'll actually move to their next point rather than do something undesirable like e.g. try to move relative to where they ended up, or get stuck trying to move through walls.
> 
> Just because we'd like to actually making uniques unique during this version doesn't mean we shouldn't make the changes that allow it to work properly first. Things should still be done in the proper order, and if appropriate prerequisites aren't produced (either due to lack of interest or failures of the attempts to work properly), the change should be put off until that's been taken care of (I'm aware there is a significant variation in the opinion of what's sufficient, including those who think the current situation is and was good enough).

They will in fact go to the next location, nothing about this is relative.  

As to the second part its been my experience that unless someone tries to make these things happen they don't.  If its decided there is more work to be done then this PR can wait until that's done or be modified to just not work fully until it happens.  It made this much easier to have that happen so I decided to give it a go.

> Would making the global variable a localized one fix that problem? Is that a fairly non-intrusive thing to do, or would that be too difficult to do for the scope of this PR?

So yes it would, it is technically possible but would be annoying.  There are a few ways to do it but none of them is totally trivial and to completely make all of it non unique would be a hassle and undo some of the benefit of this approach.  If its decided we aren't making this globally unique maybe I'll go about doing that but the prospect of another tangle with mapgen around the order various phases happens doesn't fill me with joy.

PR #64989, opened 2023-04-10
alternative fix: item spawning spells with no min/max damage

mqrause:
#### Summary
None

#### Purpose of change

Alternative to #64965, recreates the old behavior so it doesn't break existing json.

#### Describe the solution

Always spawn at least one item/item group, even with no damage defined.

#### Additional context

@casswedson probably of interest to you. I can't judge what the actually preferable solution would be since I don't know too much about spells.

### DIFF
--- a/src/magic_spell_effect.cpp
+++ b/src/magic_spell_effect.cpp
@@ -1068,7 +1068,8 @@ void spell_effect::spawn_ethereal_item( const spell &sp, Creature &caster, const
 
     std::vector<item> granted;
 
-    for( int i = 0; i < sp.damage( caster ); i++ ) {
+    int count = std::min( 1, sp.damage( caster ) );
+    for( int i = 0; i < count; i++ ) {
         if( sp.has_flag( spell_flag::SPAWN_GROUP ) ) {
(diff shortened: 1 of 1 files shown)


casswedson:
I created a new character and gave myself a level 4 repelling arc; it's not doing the thing
Screenshot from 2023-04-16 17-03-52

- OS: Linux
    - OS Version: Distributor ID: Pop; Description: Pop!_OS 22.04 LTS; Release: 22.04; Codename: jammy; 
- Game Version: afda5f6 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm]
]

mqrause:
I'm pretty sure that's just a UI bug. And I'm also actually fairly sure it's been like that before? Or it wasn't actually my fault and someone changed spell damage calculations.

casswedson:
the only thing I can say is that the spell does not work, if I cast it does not spawn anything

mqrause:
Oh god, yeah, I see my stupid mistake. Using `std::min` instead of `std::max`. Sorry about that.

PR #64993, opened 2023-04-11
Add bundling for synthetic fabric

pjf:
#### Summary

Balance "Add bundling for synethic fabric"

#### Purpose of change

Synthetic patches could not be bundled, making them extremely slow to move in bulk. This adds bundling for them, just like we have for leather, cotton, etc.

#### Describe the solution

JSON edits.

Using bundles of 100 rather than 50.

#### Testing

Applied changes to my current game. Bundled some synthetic fabric. Took it back to base. Unbundled it.

#### Additional context

None

### DIFF
--- a/data/json/items/generic.json
+++ b/data/json/items/generic.json
@@ -489,6 +489,22 @@
     "color": "white"
   },
   {
+    "id": "bundle_nylon",
+    "type": "GENERIC",
+    "category": "spare_parts",
+    "name": { "str": "bundle of synthetic fabric", "str_pl": "bundles of synthetic fabric" },
+    "description": "Fifty synthetic fabric patches, bundled tightly together for storage.  Disassemble to unpack.",
+    "//": "Weight is 50 x 5750mg per patch",
+    "weight": "287500 mg",
+    "//2": "Volume is 50 x 5ml per patch",
+    "volume": "250ml",
+    "price": 0,
+    "price_postapoc": 500,
+    "material": [ "nylon" ],
+    "symbol": ",",
+    "color": "white"
+  },
+  {
     "id": "bundle_plank",
     "type": "GENERIC",
     "category": "spare_parts",
--- a/data/json/recipes/recipe_deconstruction.json
+++ b/data/json/recipes/recipe_deconstruction.json
@@ -5093,15 +5093,6 @@
     "flags": [ "BLIND_EASY" ]
   },
   {
-    "result": "bundle_nylon",
-    "type": "uncraft",
-    "activity_level": "NO_EXERCISE",
-    "skill_used": "fabrication",
-    "qualities": [ { "id": "CUT", "level": 1 } ],
-    "components": [ [ [ "nylon", 50 ] ], [ [ "thread", 8 ] ] ],
-    "flags": [ "BLIND_EASY" ]
-  },
-  {
     "result": "nomex",
     "type": "uncraft",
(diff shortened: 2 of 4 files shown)


Maleclypse (approved):

Maleclypse (changes requested):
```/tests/item_test.cpp:739: FAILED:
(~[slow] ~[.],starting_items)=>   CHECK( item_density( target ) <= max_density )
Error: (~[slow] ~[.],starting_items)=>   1.152f <= 1.15f
(~[slow] ~[.],starting_items)=> with messages:
(~[slow] ~[.],starting_items)=>   target.typeId() := string_id( "bundle_nylon" )
(~[slow] ~[.],starting_items)=>   target.weight() := 288000mg
(~[slow] ~[.],starting_items)=>   target.volume() := 250ml
```

I think you've made it too dense  per the test or maybe not dense enough I'm not good with this particular test.

data/json/items/generic.json
```diff
@@ -488,6 +488,22 @@
...
     "color": "white"
   },
+  {
+    "id": "bundle_nylon",
+    "type": "GENERIC",
+    "category": "spare_parts",
+    "name": { "str": "bundle of synthetic fabric", "str_pl": "bundles of synthetic fabric" },
+    "description": "Fifty synthetic fabric patches, bundled tightly together for storage.  Disassemble to unpack.",
+    "//": "Weight is 50 x 5750mg per patch",
+    "weight": "288 g",
```
Karol1223, line 498: That equals to 287.5g, not 288, making your density test fail. Do that in milligrams and it'll be fine.

data/json/recipes/recipe_deconstruction.json
```diff
@@ -5092,6 +5092,15 @@
...
   },
+  {
+    "result": "bundle_nylon",
+    "type": "uncraft",
+    "activity_level": "NO_EXERCISE",
+    "skill_used": "fabrication",
+    "qualities": [ { "id": "CUT", "level": 1 } ],
+    "components": [ [ [ "nylon", 50 ] ], [ [ "thread", 8 ] ] ],
+    "flags": [ "BLIND_EASY" ]
+  },
```
Karol1223, line 5103: Don't do that. Make the recipe reversible. If you give it a dedicated uncraft like this for some reason people will be able to transform sinew/plant fiber/yarn into thread.

Karol1223 (changes requested):

data/json/items/generic.json
pjf: "Nobody will notice if I round this by half-a-gram, just for the thread..."  :)

Updated commit uploaded. Fingers crossed the density test will smile upon me.

data/json/recipes/recipe_deconstruction.json
pjf: Oh dang. All the other bundles have dedicated uncrafts, so I just went with consistency.

Making it reversible is much better. Fixing now.

Karol1223: Yeah old bundles started ages ago and then everyone just copied the same method. But reversible is the better solution - you can even make it take less than tying it together (I've been doing it for reversible recipes for some bags)

pjf:
Density tests look like they're now passing. I also crafted and uncrafted a bundle in-game to verify it's working with changes, so I think this is good to go! :tada:

Maleclypse (approved):
Thanks!

PR #64999, opened 2023-04-11
Rework grabs (most of them, at any rate)

Venera3:
#### Summary
Features "Unhardcode and rework grabs, ranged pulls"

#### Purpose of change

Hardcode = bad, mattacks = good
Opens up grabs and ranged pulls for JSON monster attacks, and reworks how grabs work.

#### Describe the solution

- Reworked and limbified grab logic
  - Instead of cumulative grabs being pooled in the torso grabs now target limbs normally (limit one grab per limb)
  - Debuffs handled by limb scores only (monsters are only held in place and opened for grab-dependent attacks), removed -block attempt and the hardcoded dodge debuff
  - Each limb defines its own grabbing effect to apply to the monster grabbing that limb for selective grab removal
- Reworked grab break calculations
  - Grabs are broken one-by-one, removing the grab in question (and the grabbing effect corresponding to the grabbed limb from the monster in question)
  - Grab strength scales from 1-100 for monsters, this gets turned into grab effect intensity 
  - Grab break strength depends on skill level, a custom character modifier and stats
  - Grab breaks don't confer a save against grabs hitting, but are treated as +10 on the grab break calculation
  - Pulled pockets increase grab break strength considerably
- Unhardcoded the grab effect, every effect with the flag `GRAB` will be treated as a grab for movement/breakage purposes
- Unhardcoded `GRAB` and introduced `grab_data` for mattack(JSON) monster attacks, allowing you to define e.g. pull data, grab strength, dedicated grab effects and messages
- Unhardcoded and expanded `RANGED_PULL`
  - Pulling depends on target/puller weight ratio, as defined in `grab_data` (defaults to 75% of the puller's weight)
  - Seatbelts still prevent pulling, the damage component was nixed
  - Pulls don't just magically break all grabs, instead they will attempt breaking all grabs one-by-one and fail unless they manage to break them
- Deprecated `MF_GRABS` for automatically grabbing on attack (it still handles monster descriptions, for now)
- Reworked crowd crush/suffocation code to care about breathing score, and increase crush level for adjacent impassable terrain 
- Added a ton of unit tests keeping the grab break chance honest (see Additional Context)

TODO:
- [x] Move monster attacks over to the JSON attacks and grab strengths
  - Added grabs and grab strengths to all monsters
- [x] Add some more unit tests for e.g. dedicated removal, pull weight limits and such
- [x] Some more testing

**NOT** touched (for now) are `GRAB_DRAG` and `NURESEBOT` attacks, this is pretty big already and I'm still missing 1K lines of JSON

Break this up, but the meat of the changes is the flagification of grabs and the rest doesn't really lend itself to chunking it (other than the actual pull code, but that's like 100 lines).
Allow more grabs per limb, but the dedicated removal code can't deal with that until effect sources track monsters.
Allow reducing the effect intensity on a "partial" success on grab break.
Different calcs, different scaling, different everything - I'm open to suggestions on the mathy front, as always.

#### Testing

Did some live testing on all functions.
Added a unit test for grab break stats, and reworked the suffocation test with the new changes (including wall crushing) in mind.
More to come

#### Additional context

Unit tests check for a Day1 skill 0 char and a "lategame" unarmed 8+Brawling+Grab break one, both at 8 STR agains 1/3 weak/mid-strength/max-strength grabs (20/50/100, corresponding to a base zombie, a Tier 2 grabber, and a Tier 3 grabber respectively).
The results are documented in the tests itself, relevant excerpts
- Starter characters have 40% chance to break a single weak grab every turn 5% to break *all three* grabs on a single turn (remember, grabs are tracked per-limb, so this misses partial breaks), with the single-grabber chance dropping to 15% and 7,5% at mid/max grab strength
- Lategame characters have 100% chance to break a weak grab and 99% to break three on a single turn, scaling to 80/40% for single grabbers and 25%/5% for three grabs

#### Monster attack changes

- Added ´bite_humanoid´ and ´scratch_humanoid´ special attacks for humanoid enemies and unhardcoded the bite/grab interaction on them - humanoid enemies will *require* a grab to bite, but *can't* scratch while grabbing
- Added ´bite_grab´ for monsters that bite you and don't let go - dogs, primarily, but also ants and similar critters. The grab strength of a dog is generally below a similarly evolved zombie for weight reasons (even though zanimals as a rule have extremely wacky weight progression). It will also open the door to some fun combo attacks down the line like crocodilian death spins and similar rending grab/chew things, but that's out of scope for now.
- Added ´scratch_grab_required´ to dogs and similar scratchy critters who need to hold you before they start scritching - it turned out to be a relatively short list but it felt neat.
- Unhardcoded ranged pulls and generally aimed to eyeball the weight ratios sensibly, but chonkier armored characters should have an easier time. I considered adding a backup grab to all pullers, but filtering it is not straightforward and too-short grab cooldowns can be problematic.
- Audited zombie grab capabilities - a *LOT* of zeds didn't grab at all despite being equipped to do so, and a good chunk of them did have a bite attack they never got to fire because of that. I refrained from eyeballing too much on the balance front and just slapped a base grab on most (apart from the shocker gang).
- Added ´grab_2´ attack - a *second* grab attack for extra-grabby critters (vine beasts and shelled mutants in vanilla) depending on a pre-existing grab by the same monster. It should be reasonably stackable in the future but I stayed at double grabs for now.


LyleSY:
Would it help for me to give you a list? Almost everything is using teeth. Only exceptions are these three zombies and their evolutions: mon_zerizinosaurus mon_zothronychus mon_zanzu . If you find it easier to give them all the same I'm happy to adjust later, those three could also easily grab with a bite.

Venera3:
> Would it help for me to give you a list? Almost everything is using teeth. Only exceptions are these three zombies and their evolutions: mon_zerizinosaurus mon_zothronychus mon_zanzu . If you find it easier to give them all the same I'm happy to adjust later, those three could also easily grab with a bite.

Okay, I can work with that - I'll probably eyeball grab strength from the weight but it should be a start.

LyleSY (approved):
I don't understand the error you're getting, but all of the DinoMod content looks correct to me. Well done!

PR #65004, opened 2023-04-11
Incremental skills

bombasticSlacks:
#### Summary
Balance "partial skill levels contribute to most game tests"

#### Purpose of change
Something that is obviously good as we stretch levels.

#### Describe the solution
Made character::get_skill_level return a float. Fixed every instance in the code base of it to still have correct behavior.

#### Testing
Haven't, CLANG is probably gonna have a field day with this one.


andrei8l:
I see a few `round()`s in there. Doesn't that flip the issue to overestimating the skills half the time?

bombasticSlacks:
> I see a few round()s in there. Doesn't that flip the issue to overestimating the skills half the time?

I used round when the value converts to a number of dice for further RNG. It does just shift the break point to the .5's rather than the 0's, but I think that should be okay. If it is undesirable I can instead either: 
* swap the rounds for static_cast<int> 
* do roll_remainder, but that would require making sure each isn't a roll that needs to be consistent.

anothersimulacrum:
Please no roll remainder, that makes any sort of testing of outcomes really awful.

bombasticSlacks:
I think I've settled on round actually being really good behaviour for the flat outcomes where it needs to be an int *because*:

Characters normally train up to a certain level and then plateau, so for example, practising dodge to 3, or throwing to 3. If we just bottom skills, then any amount of rust off your max practicable level means you lose the whole level. If we round, it gives you a comfortable buffer to keep those skills honed without introducing tedium to micro manage.

NetSysFire:
I think this would also make sense to extend to proficiencies but this might be out of scope here since there are already many changes here.

PR #65008, opened 2023-04-11
Clothing and Armor Coverage Conversion Part 4

Reuvin:
#### Summary
Balance "Converts most of the armor and clothing still using the old limb system to the new system with sub-limbs, part 4: Perfected pants and helpful helmets"

#### Purpose of change
There are still a number of armors making use of the old coverage system that uses the entire limb for coverage data, instead of the sub-limb system. This series of PRs is meant to convert all the armors and clothing I can find that would have varied coverages that still use the old system, and then assigning the coverages to the specific sub-limbs instead if it were less than 100%. 

#### Describe the solution
In part 4, I went through all the leg clothing, leg armor, and helmets, and converted as much of the coverage values as I felt necessary. If I found an item that needed changing, I would pull up a picture of the item in question and then assign coverages to the sub limbs of the body part it covers in accordance with the real item as best as I could. 

Not doing this and saving myself the time. 

#### Testing
Part 4 is good to go ingame without any errors.

#### Additional context
A continuation of my previous conversions #57911, #57940, and #64897

Helmets got a pretty big change. With the neck being a component of the head, a lot of helmets got lower overall coverage. Few helmets include complete nape protection, and far fewer include throat protection. Make sure to wear a coif underneath.

### DIFF
--- a/data/json/items/armor/helmets.json
+++ b/data/json/items/armor/helmets.json
@@ -18,7 +18,16 @@
     "material_thickness": 2,
     "qualities": [ [ "COOK", 3 ], [ "BOIL", 2 ], [ "CONTAIN", 1 ], [ "CHEM", 1 ] ],
     "techniques": [ "WBLOCK_1" ],
-    "armor": [ { "encumbrance_modifiers": [ "IMBALANCED" ], "coverage": 55, "covers": [ "head" ] } ]
+    "armor": [
+      {
+        "encumbrance_modifiers": [ "IMBALANCED" ],
+        "coverage": 100,
+        "covers": [ "head" ],
+        "specifically_covers": [ "head_crown" ]
+      },
+      { "coverage": 80, "covers": [ "head" ], "specifically_covers": [ "head_forehead" ] },
+      { "coverage": 20, "covers": [ "head" ], "specifically_covers": [ "head_ear_l", "head_ear_r" ] }
+    ]
   },
   {
     "id": "stockpot_helmet",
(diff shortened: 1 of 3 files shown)


Reuvin:
As an additional note, I have finished the helmets as well. I'd add them here to this PR, except there's a slight hiccup. Metal helmet eye slits. I'm not sure if I should set those to 95% universally for eyes because something COULD get through, or 100% to say that the threats you'll face in the apocalypse never will. Part of me wants to reserve 100% for high end helmets involving ballistic glass and other see through eye protection methods that actually completely protect the eyes, but short of fighting intelligent humans with proper equipment, nothing would feasibly be able to get through the eye slits anyhow. 

@bombasticSlacks if you could give a definitive answer on which route to take that'd be great.

bombasticSlacks:
> As an additional note, I have finished the helmets as well. I'd add them here to this PR, except there's a slight hiccup. Metal helmet eye slits. I'm not sure if I should set those to 95% universally for eyes because something COULD get through, or 100% to say that the threats you'll face in the apocalypse never will. Part of me wants to reserve 100% for high end helmets involving ballistic glass and other see through eye protection methods that actually completely protect the eyes, but short of fighting intelligent humans with proper equipment, nothing would feasibly be able to get through the eye slits anyhow.

I think 95% is more flavourful, I mean that's like a 1/1000+ chance to get hit in the eye slit still 😆

DillonSimeone:
If something hits those slits hard enough, pieces will go through into the eyes! Like a cheese grater at high velocity.

PR #65019, opened 2023-04-12
Fix tourniquet

HuajiSama:
#### Summary
None

#### Purpose of change
Missing tourniquet_lower_XS
Incorrect covers  of tourniquet_lower_XL
image

#### Describe the solution
Add tourniquet_lower_XS
Fix tourniquet_lower_XL covers
image

None


PatrikLundell:
The description is very poor. Just providing an image and letting the reader guess what in it you might think is wrong isn't particularly helpful.

PR #65026, opened 2023-04-12
Fix and enhance camp radio tower

SurFlurer:
#### Summary
Features "Fix and enhance camp radio tower"

#### Purpose of change
1. To make camp's radio tower more powerful. To make it act like a cellular phone base station. 
2. To fix the bug that `has_provides( "radio_tower" )` actually should be `has_provides( "radio" )`

#### Describe the solution
1. "Stand higher, call further" is still true.
2. Radio towers can relay radio calls, as detailed below:
- If there's a radio tower near the player, the player can call NPCs **through** that tower, and the effective distance (from the tower to the player, and from the tower to NPC ) gets boosted by the tower's enhanced sending and reception capabilities.
- If one tower isn't enough, and there's another radio tower near the NPC, the player can call that NPC through two towers. The range between those towers can be much longer ( up to 2500 overmap tiles). 

Only fix the bug, and keep radio tower specific to the very camp.

#### Testing
1. Direct call, the effective range is correct.
2. Relayed once, the effective range is also correct.
3. Relayed twice, the call is successful. 

#### Additional context
TODO: Add a path-finding ability to the radio system, making more relays possible.

### DIFF
--- a/src/faction.cpp
+++ b/src/faction.cpp
@@ -658,7 +658,7 @@ int npc::faction_display( const catacurses::window &fac_w, const int width ) con
                     bool camp_to_npc = false;
                     bool camp_to_camp = false;
                     for( const camp_reference &i : camps_near_player ) {
-                        if( !( i.camp->has_provides( "radio" ) ) ) {
+                        if( !i.camp->has_provides( "radio" ) ) {
                             continue;
                         }
                         if( camp_to_camp ||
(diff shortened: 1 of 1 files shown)


PatrikLundell:
I don't think you'd be able to use any radio towers "in the wild" as they wouldn't have any power (despite a Refugee Center mission to manipulate a tower, if I remember correctly). Thus, I think only working towers at/by active faction camps should be usable. This would exclude the radio tower base camp up until the point the player fixes the tower and include any non player factions that might have towers (I would expect the Exodii to have the capability, but they'd be unlikely to adhere to this world's cellphone protocols. I would expect the Refugee Center and Tacoma Range to want to build towers, although I haven't seen them doing so. Hub01 would probably use the tower they provide a quest for, although I would expect them to block it from being used by anyone else through encryption).

PR #65034, opened 2023-04-12
'variant' field should be optional in sound_effect and preload

### LINKED ISSUE #65031: Sounds not playing after update from 2023-04-12-0515 to 2023-04-12-1139
### Describe the bug

After switching from 2023-04-12-0515 to 2023-04-12-1139, the sound pack OTOPACK lost most of its played sounds. The Same is true for CC Sounds as well.

### Attach save file

N/A

### Steps to reproduce

Start game with 2023-04-12-1139 build and use OTOPACK or CC Sound.
Sounds are missing from both packs which worked clearly in the 2023-04-12-0515 build.

### Expected behavior

Sounds playing like in older builds.

### Screenshots

https://drive.google.com/drive/folders/1a8SJ9ZhiGLvqpEE7FVDLuXzG-TutaG5D?usp=share_link

Here are 4 videos i took where I compare the two builds and sound packs.

- OS: Windows
- Game Version: eb85d6b [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

irwiss:
#### Summary
None

#### Purpose of change

Fix #65031

#### Describe the solution

Actually make variant optional

Didn't notice get_as_string_array is making the member mandatory, since some sounds (that load before the variant-less json entry) playing gave the illusion of sounds working, but actually invalid and unvisited member json errors for soundpacks are silenced and shunted into debug.log only, instead of raising a debugmsg.

Also make soundpack errors raise debugmsg so invalid json is actually visible, but I'm not sure what's the reason it isn't so already

#### Testing

Select basic sounds, left-right arrows in main menu should bleep, debug.log should not have message about soundpack json errors

### DIFF
--- a/src/sdlsound.cpp
+++ b/src/sdlsound.cpp
@@ -564,9 +564,13 @@ void sfx::load_sound_effects( const JsonObject &jsobj )
         key.night = jsobj.get_bool( "is_night" );
     }
     const int volume = jsobj.get_int( "volume", 100 );
-    std::vector<std::string> variants = jsobj.get_as_string_array( "variant" );
-    if( variants.empty() ) {
-        variants.emplace_back( "default" );
+    std::vector<std::string> variants;
+    if( jsobj.has_array( "variant" ) ) {
+        variants = jsobj.get_string_array( "variant" );
+    } else if( jsobj.has_string( "variant" ) ) {
+        variants = { jsobj.get_string( "variant" ) };
+    } else {
+        variants = { "default" };
     }
     for( const std::string &variant : variants ) {
         key.variant = variant;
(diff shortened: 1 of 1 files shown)


Fris0uman:
no weay to use std::optional here?

Fris0uman:
I used the wrong word, I actually meant `optional` you know the dedicated thing from generic_factory to load optional fields?

irwiss:
Don't think it'll help much, converting loading sounds to generic factory is a much bigger PR and most of the LOCs is due to me not wanting to break existing json data single json field needs to read array/fallback to string/fallback to default string, which (afaik) generic_factory's optional doesn't have a nicer syntax for either, support for relative/proportional doesn't seem like it'll help much for sounds either

PR #65038, opened 2023-04-12
Magiclysm: Frost armor actually works now

Saint-of-Grey:
#### Summary
Bugfixes "Magiclysm: Made frost armor actually do something"

#### Purpose of change

I noticed that casting the frost armor spell would summon "0 frost armors". And indeed, when successfully cast, it eats your mana and does nothing. So I fixed that real quick.
#### Describe the solution

Added a missing `"min_damage" : 1` and `"max_damage": 1` to the frost armor spell and the upgraded version.

Ending all my kelvinist runs by crying myself to sleep.
#### Testing

Fired up a game, cast the spell, now it works.
#### Additional context

As far as I know, this wasn't reported, and I figured fixing it was just as much effort as writing a bug report for it.


casswedson:
#64989 and https://github.com/CleverRaven/Cataclysm-DDA/pull/64965 both address this

Saint-of-Grey:
So there's an underlying issue? That explains why I couldn't find anything about it specifically.

PR #64985, opened 2023-04-10
Put Soap in Mall Sinks

Aimless-dude:
#### Summary
Content "Put Soap In Mall Sinks"

#### Purpose of change
Currently, mall sinks do not spawn anything at all, which dosen't make much sense and also makes Mall starts a little harder, as you cannot wash any filthy gear easily.

#### Describe the solution

 Each sink will now spawn a small amount of soap, and also adds a itemgroup for spawning things in other Public sinks. 

#### Describe alternatives you've considered

Find the spawn tables for the campsite bathrooms if there are any, add towels or toilet paper, Give up on my dreams of hygiene.

#### Testing
Boot up a new save, select a mall start, and navigate to the nearest bathroom sink. Observe that each one holds an amount of soap.

PR #65039, opened 2023-04-12
prevent spam from linter bot reviews

casswedson:
#### Summary
None

#### Purpose of change
you can make the github actions bot spam reviews of the same starting errors over and over quite easily, just hit commit suggestion on each review individually

just make it cancel runs if they are running for the same pr

prevent this kind of stuff #64985 200+ comments, oh lord I am so sorry

#### Describe the solution
make it use a concurrency like so `workflow-name-unique-prnumberidentifier` so it cancels run of the same workflow running for a specific pr

#### Testing
I tried to trigger it here https://github.com/casswedson/Cataclysm-DDA/pull/101 it didn't spam and behaved a little better than expected *worried about that* but it does cancel runs like the release workflow does; just run on the latest commit please

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -15,7 +15,7 @@ on:
       - '**.c'
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.run_id }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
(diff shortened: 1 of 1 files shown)


Aimless-dude:
I'd like to formally apologize for said spam, I did not know it would do that.

casswedson:
it's more my bad for ehh... just not realizing the faults on my implementation; it's cool tho

PR #65058, opened 2023-04-13
math_parser/eoc: port u_val pain to math

andrei8l:
#### Summary
None

#### Purpose of change
Start porting `arithmetic` functions to `math`. bombasticSlacks used `"u_val": "pain"` in their latest devstream and it's a small enough target for this first port.

#### Describe the solution
Fix validation of number of arguments for nullary function.
Move the pain code to `math` and change all the JSON to match.
Also move the dialogue function bits to a new header to keep things clean.

N/A

#### Testing
There are several uses of `"u_val": "pain"` in test units, both for assignment and evaluation, so we're already fully covered.

#### Additional context
So instead of
`{ "arithmetic": [ { "u_val": "pain" }, "-=", { "const": 30 } ] }`
or
`{ "math": [ "u_val('pain')", "-=", "30" ]}` (with the shim)
you can now use
`{ "math": [ "u_pain()", "-=", "30" ] }`

The `min` and `max` parameters from the old pain function can be replaced by `min()` and `max()` math functions.

### DIFF
--- a/data/json/effect_on_condition.json
+++ b/data/json/effect_on_condition.json
@@ -259,7 +259,7 @@
     "effect": [
       { "u_message": "Your malfunctioning bionic causes you to spasm and fall to the floor!", "type": "bad" },
       { "sound_effect": "elec_crackle_high", "id": "bionics", "volume": 100 },
-      { "arithmetic": [ { "u_val": "pain" }, "++" ] },
+      { "math": [ "u_pain()", "++" ] },
       { "u_add_effect": "stunned", "duration": 1 },
       { "u_add_effect": "downed", "duration": 1, "force": true }
     ]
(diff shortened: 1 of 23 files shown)


Ramza13:
So its not a huge deal at the moment but I don't think this should be removed from the old method quite yet.  Both weighted eoc lists and the switch effect do still use the old method and so this is technically reducing json power.  Once everything is moved over those can be converted and then I'd say is the time to remove the old method.

andrei8l:
Both switch and weighted list are already supported.

 There's a test unit for switch here:
https://github.com/CleverRaven/Cataclysm-DDA/blob/2fcf05f39099a7428e62a328c50130f102cfc2d8/data/mods/TEST_DATA/EOC.json#L30-L40

Weighted list uses `get_get_dbl()` for the weight which supports `math`
https://github.com/CleverRaven/Cataclysm-DDA/blob/2fcf05f39099a7428e62a328c50130f102cfc2d8/src/condition.cpp#L2033
and apparently I forgot to add a test for it. I'll add one now.

andrei8l:
I've added a test unit for math-as-weight in `weighted_list_eocs`. Let me know if there are other issues please.

PR #65069, opened 2023-04-14
eoc: untemplate all EOC code

andrei8l:
#### Summary
None

#### Purpose of change
All EOC code is templated so it can work with either `dialogue` or `mission_goal_condition_context`, but `mission_goal_condition_context` is just a gimped `dialogue` used in only one place and the relevant code can be almost trivially ported to `dialogue`. Some functions are templated but use only `dialogue` anyway.

This templating makes the code hard to follow, especially while compiling and debugging, and needs either large headers or a lot of boilerplate from template instantiations.

For `math_parser` I want to use as little boilerplate as possible and not rely on gigantic monolithic if-else blocks like the current code for EOC functions (most significantly `get_get_dbl()` and `get_set_dbl()`).

#### Describe the solution
Trivially port `mission_goal_condition_context` code to `dialogue`

Semi-automatically untemplate all EOC code. I only had to intervene manually where code had to be moved around or out of headers (most notably in `math_parser_impl.h` and `dialogue_helpers.h`)

N/A

#### Testing
Since this is only a refactor, we should be good if the code compiles and the tests still pass.

#### Additional context
I would have done this before writing `math_parser` if I had known about it.

All tests need to pass for this one.

### DIFF
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -119,7 +119,7 @@ dbl_or_var_part get_dbl_or_var_part( const JsonValue &jv, const std::string &mem
     return ret_val;
 }
 
-dbl_or_var get_dbl_or_var( const JsonObject &jo, std::string member, bool required,
+dbl_or_var get_dbl_or_var( const JsonObject &jo, const std::string &member, bool required,
                            double default_val )
 {
     dbl_or_var ret_val;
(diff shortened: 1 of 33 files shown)


Ramza13:
Holy shit I love you and this!! The templating has been a huge pain everytime I touch things but it scared me too much to remove.

PR #65070, opened 2023-04-14
Claws rework

KittyTac:
#### Summary
Balance "Make claws integrated unarmed weapons, also make them stronger"

#### Purpose of change
Claws were unusually weak before, so I decided to buff them (on par with various kitchen knives now) and plug them into the integrated armor system.

#### Describe the solution
Make integrated armors for all the claw types (retractable claws uses normal claws).

Leaving as is.

#### Testing
~~The main issue: while the damage works correctly, the technique I added to the weapons, meant to slow them to 75 move cost instead of the unreasonably fast (for swiping) 50, simply doesn't work. A fix to make `"unarmed_weapons_allowed": true` actually work is needed, and that's C++ I am uncomfortable touching currently.~~
Tested and it works and it's properly slower than punches.

#### Additional context
N/A.


Karol1223:
I am really not convinced that claws should match kitchen knives. Especially not since it's a trait shared across trees like rat or lupine.

anothersimulacrum:
Please add these to the DPS tests.

Terrorforge:
Thanks for doing this, I was just getting annoyed about how useless claws are.

> I am really not convinced that claws should match kitchen knives. Especially not since it's a trait shared across trees like rat or lupine.

Honestly, [rat claws](https://i.pinimg.com/originals/c6/e1/6d/c6e16d6200ae48ad166456e9b2a3e5af.jpg) look like they'd be pretty nasty if they were human-sized. But Lupine probably shouln't have the Claws mutation at all, unless it's remade into something like Snout where it's just a prerequisite trait that barely does anything and Feline (and probably Beast/Ursine) gets its own version that does actual damage.

Anyway, I don't think a technique is a great solution here anyway, because you're not guaranteed to use it. Especially not if you're using a martial art, which someone intentionally relying on their claws almost certainly is. You could use an enchantment to apply an attack speed penalty if the player isn't wielding a weapon, maybe? I did a similar thing to make magical unarmed weapons only deal bonus damage while actually unarmed in #64193

But fwiw "unarmed_weapons_allowed": true means you're allowed to perform the technique  while wielding an unarmed weapon, like how knuckledusters used to be implemented. This isn't a thing any more, so yeah that flag does nothing. I guess there's no infrastructure for allowing the current style of unarmed weapons to give techniques? Other than WBLOCK in combination block_while_worn, naturally.

> Please add these to the DPS tests.

Can the DPS tests handle worn weapons?

KittyTac:
> I am really not convinced that claws should match kitchen knives. Especially not since it's a trait shared across trees like rat or lupine.

My reasoning: A cudgel does 10 damage. I think getting raked full-force with a full hand of claws sized for humans would hurt as much or a bit more as a whack with a cudgel.

Re: enchantments, I might try that one actually, gonna try tomorrow.

> Please add these to the DPS tests.

How? I'm new to item modding.

KittyTac:
Fuck it we ball, gonna do it today.

Karol1223:
> My reasoning: A cudgel does 10 damage. I think getting raked full-force with a full hand of claws sized for humans would hurt as much or a bit more as a whack with a cudgel.

'I think' is an extremely poor reasoning, really. You gotta compare to other similar stuff, which a cudgel isn't even remotely similar. How much it hurts is also irrelevant to the damage. If these were full-size tiger claws I'd be able to look past the kitchen knife damage. But the character isn't a tiger - they get tiger claws on their human size (except for ursine, which should likely get their own breed of the mutation given they mutate larger) which wouldn't be anywhere near comparable to that.

KittyTac:
> > My reasoning: A cudgel does 10 damage. I think getting raked full-force with a full hand of claws sized for humans would hurt as much or a bit more as a whack with a cudgel.
> 
> 'I think' is an extremely poor reasoning, really. You gotta compare to other similar stuff, which a cudgel isn't even remotely similar. How much it hurts is also irrelevant to the damage. If these were full-size tiger claws I'd be able to look past the kitchen knife damage. But the character isn't a tiger - they get tiger claws on their human size (except for ursine, which should likely get their own breed of the mutation given they mutate larger) which wouldn't be anywhere near comparable to that.

Fair enough. I did multiply steak knife damage by 1.5 because there were several claws and I felt that compensated for them being shorter-- I suppose I can just drop the values by 50%. Also, a tiger paw is larger than a human hand but not all that much.
image

KittyTac:
Welp, that's that, I still need help with the test stuff but overall after a bit of playtesting I'm happy with the balance here.

Terrorforge:
Just realized the enchantment option is going to apply regardless of whether you're using the claws or not. Like if you put on a knuckleduster, you're gonna be doing the knuckleduster damage but still suffer the speed penalty from the claws.

That's probably fine as part of the whole "you have weird mutant hands that make it hard to do certain things" thing, but it should maybe be mentioned in the desciption. "The akward anatomy makes unarmed attacks slower, whether you use your claws or not" or something.

(Also fwiw adding more blades tends to make weapons less damaging. The same amount of force gets distributed over a wider area, so each cut/stab is more shallow, and the seriousness of an injury is typically more dependent on how deep it is than how much surface area it covers. The few multi-pringed weapons that exist tend to be more for catching and holding than for added damage, e.g. tridents used for spearfishing, or indeed the claws of a cat, which normally aren't used directly as weapons, but to hold the prey in place for the killing bite.)

anoobindisguise:
Additive attack speed is bad btw. I don't know if you intended +20 attack speed to be negative or positive but it makes attacks take 20 more moves, just wanna make sure you're aware.

fairyarmadillo:
> Thanks for doing this, I was just getting annoyed about how useless claws are.
> 
> > I am really not convinced that claws should match kitchen knives. Especially not since it's a trait shared across trees like rat or lupine.
> 
> Honestly, [rat claws](https://i.pinimg.com/originals/c6/e1/6d/c6e16d6200ae48ad166456e9b2a3e5af.jpg) look like they'd be pretty nasty if they were human-sized. But Lupine probably shouln't have the Claws mutation at all, unless it's remade into something like Snout where it's just a prerequisite trait that barely does anything and Feline (and probably Beast/Ursine) gets its own version that does actual damage.

Real life rat claws are not weapons. They don't use them for hunting or combat, they're just for helping them dig, climb, or hold onto things. They have a pretty ferocious bite (about a pound of bite pressure, which is respectable at that size, and some very sharp teeth) which they use extensively in social combat, defense, and hunting, so that's really what rat mutants should be doing. Their claws don't seem very hard either, they feel more like fingernails. It is my firm opinion that rat claws should be renamed, restatted as one of the weakest types, and basically made a prereq for burrowing. If that's beyond the scope of the PR then fine, but it has always stuck out to me as a very poor reflection of how rats work.

As for the other mutants, claw buffs sound great to me.

anoobindisguise:
One other question, should talons actually cause hand encumbrance? Their description reads "...After a bit of practice, you find that this does not affect your dexterity..." and it's only one of your fingers on each hand. 7 is definitely a lot! On the bright side they can probably be made compatible with fingerless gloves now though that we have more advanced logic to handle such conflicts beyond "unless it's XL you can't wear it period".

fairyarmadillo:
I think birds should lose talons. Bird mutants have toe talons, so I don't know why they get finger ones too. Birds do not have clawed forelimbs. If it's supposed to be a dinosaur throwback thing, we have raptor for that.

It won't even really be a nerf to the mutation line - bird was always about using guns or their beak anyway.

Shodan14:
Should fingertip razors get a buff as well?

anothersimulacrum:
To add items to the DPS test, add them to the appropriate file here. https://github.com/CleverRaven/Cataclysm-DDA/tree/master/data/mods/TEST_DATA/expected_dps_data

KittyTac:
> Additive attack speed is bad btw. I don't know if you intended +20 attack speed to be negative or positive but it makes attacks take 20 more moves, just wanna make sure you're aware.

That was my point, to make it slower, as unarmed is too fast for swiping to make sense.


> One other question, should talons actually cause hand encumbrance? Their description reads "...After a bit of practice, you find that this does not affect your dexterity..." and it's only one of your fingers on each hand. 7 is definitely a lot! On the bright side they can probably be made compatible with fingerless gloves now though that we have more advanced logic to handle such conflicts beyond "unless it's XL you can't wear it period".

Yeah I'll change that, and maybe give Lupine just long fingernails but Bear stronger claws.



> Should fingertip razors get a buff as well?

Probably.



> To add items to the DPS test, add them to the appropriate file here. https://github.com/CleverRaven/Cataclysm-DDA/tree/master/data/mods/TEST_DATA/expected_dps_data
> 
> Use this as the authoritative source for balancing weapons.

kk thanks will do this evening alongside the other stuff.

Terrorforge:
Double-check that the DPS tests account for the weapon being worn (so it uses unarmed attack speed as a base rather than whatever the speed is of smacking the enemy while holding the item) and that the attack speed enchantment is applied correctly. Maybe triple-check with some extreme test values, because I think the speed of unarmed + enchantment penalty is very close to attacking with an object that weighs ~200g.

(I have no idea if they do account for this or not, it just seems like something that could potentially cause problems)

KittyTac:
Tomorrow, not in a contributor-y mood this evening sorry.

KittyTac:
I kinda lost the spoons to do this at the time, will pick this up this evening.

KittyTac:
I'm deciding to keep rat claws strong as I feel that's kind of outside the scope of this PR; rat would be pretty weak without them and I'd like to approach rat with a full rework. Lupine won't be hurt by losing its claws for nails, as a lupine you probably want to be using a big heavy weapon to take advantage of the insane stamina, not fast and light claws.

Consider the rat claws to be in limbo. I gave them to bear.

KittyTac:
I'm gonna make their weight the same for balance reasons, they ended up not being much better between tiers.

KittyTac:
Aaaand tests done. I need a dev to help me resolve the other stuff getting into the PR though... I suck at this.

KittyTac:
I think updating before continuing work was a mistake..

KittyTac:
Okay how tf did the mutations.json file replace the integrated armor file? I'll try to fix that.

KittyTac:
OK I was using different syntax that made it work right but it made it not build.

KittyTac:
Sorry for the delays guys, I lost the spoons to do this for a while. It's done now.

Nebnis:
> Sorry for the delays guys, I lost the spoons to do this for a while. It's done now.

looks like the dps test is failing?

PR #65071, opened 2023-04-14
Misc Comestible Fixes

Karol1223:
#### Summary
None

#### Purpose of change
There were several problems with comestibles that were brought to my attention. I also wanted to add salt as a new material to hopefully fix some densities. It seemed a bit much for the typical item fixes I do so it went into a separate branch.

#### Describe the solution
<details>
<summary>PEANUTBUTTER</summary>

-  Was given ``junk`` and ``nut`` as its material with a 50/50 proportion because I'm not a chef
- This should also affect the panutbutter spread and the imitation peanutbutter - making them properly problematic for carnivore mutants
</details>

<details>
<summary>SALT & SALT WATER</summary>

-  Added a new salt material
- The new material was applied to salt, seasoned salt, rock salt and salt water - lets hope this fixes some densities
</details>

<details>
<summary>SWEET WATER</summary>

-  Now has ``junk`` as part of its material to (hopefully) fix the density and represent the sugar used
- The density may still be broken because the laws of physics blur around food recipes
- Weight increased 260g -> 300g - the recipe uses 50g of sugar
</details>

<details>
<summary>COOKING OIL</summary>

-  Both types got 2x increase in commonality in pantries and seafood restaurants, 2.5x increase in normal restaurants
</details>

<details>
<summary>DEHYDRATED LOBSTER</summary>

-  Will now spawn in paper wrappers instead of laying loose in the pantries
</details>


mqrause:
> SWEET WATER
>     * Volume increased 250ml -> 265ml

I don't think that's a good change considering pretty much all liquid containers are multiples of 250ml.

Karol1223:
> > SWEET WATER
> > * Volume increased 250ml -> 265ml
> 
> I don't think that's a good change considering pretty much all liquid containers are multiples of 250ml.

I just realized - it's less than 'not good'. This will break it spawning in the default container... I'll revert it to 250, the problem here is that the density is going to implode... But I'll be lying if I was to say that I have a solution.

PR #65075, opened 2023-04-14
Make safemode safe again: don't attack neutral creatures via move

qoreQyaS:
#### Summary
Features "prevent attacking neutral critter via move in safemode"

#### Purpose of change
Due to popular demand, prevent attacking neutral creatures with the move command while safe mode is on.

#### Describe the solution
Upon moving to a tile which is occupied by a monster which has a neutral attitude towards the player in safemode, a warning message is displayed and the action is aborted.

Adding a separate option which allows a prompt/ignore/attack choice.
disregarded because of KISS and no-new-option-policy

#### Testing
spawn an exodii worker and move onto its tile with safemode turned on.

#### Additional context
thanks to irwiss and officer_dirtbag for halp

image

### DIFF
--- a/src/options.cpp
+++ b/src/options.cpp
@@ -1541,7 +1541,7 @@ void options_manager::add_options_general()
     add_empty_line();
 
     add( "SAFEMODE", "general", to_translation( "Safe mode" ),
-         to_translation( "If true, will hold the game and display a warning if a hostile monster/NPC is approaching. This will also prevent attacking neutral creatures on move. " ),
+         to_translation( "If true, will hold the game and display a warning if a hostile monster/NPC is approaching.  This will also prevent attacking neutral creatures on move." ),
          true
        );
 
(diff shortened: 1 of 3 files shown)


src/options.cpp
```diff
@@ -1541,7 +1541,7 @@ void options_manager::add_options_general()
     add_empty_line();
 
     add( "SAFEMODE", "general", to_translation( "Safe mode" ),
-         to_translation( "If true, will hold the game and display a warning if a hostile monster/NPC is approaching." ),
+         to_translation( "If true, will hold the game and display a warning if a hostile monster/NPC is approaching.  This will also prevent attacking neutral creatures on move." ),
```
irwiss, line 1544: imo don't put it here, players should not be normally messing with options for attacking neutral creatures and the message they're given already guides them to correct way to do it, you could show them their keybind like done in https://github.com/CleverRaven/Cataclysm-DDA/blob/3e920a092b4136c9ecec0b9f0ab3f90a4ced327f/src/game.cpp#L9930

PR #65077, opened 2023-04-14
obsolete the power armor bionics

anoobindisguise:
#### Summary
Content "obsolete the power armor bionics"

#### Purpose of change
power armor bionics no longer really make sense because bionics belong to the Exodii who would not have any CBMs designed specifically to interact with Cata-Earth technology. Further these bionics were super overpowered, taking a 4kW power armor cost down all the way to 1 kW / 0.5kW for the Mk.1/Mk.2 respectively compared to a standard UPS.

#### Describe the solution
Obsolete em.

I haven't removed the underlying functionality that checks for a flag from a bionic and edits power armor costs accordingly (preventing such bionics from draining unless you have active power armor and preventing power armor from draining if you have an active bionic).
I don't know if Aftershock still wants these because it largely replaced power armor with exosuits. But I'm planning to wait until the whole power armor system is torn out to remove the c++ code that handles bionics.


data/json/obsoletion/obsolete-0.G-items.json
```diff
@@ -262,5 +262,51 @@
         "encumbrance": 20
       }
     ]
+  },
+  {
+    "id": "bio_power_armor_interface",
```
irwiss, line 267: Moving stuff around in files is not going to make a difference, you need  to make 2 MIGRATION entries like described in https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md#obsoletion-and-migration to some inert item type id (say burnt bionic or whatever else), or alternatively another bionic id of same rarity and theme

Also these files are for the 0.G obsoletions that are left over still, new items shouldn't be put here

data/json/obsoletion/obsolete-0.G-items.json
```diff
@@ -262,5 +262,51 @@
...
+    "looks_like": "bio_int_enhancer",
+    "description": "Interfaces the user's bionic power system with the internal charging port on suits of power armor, allowing them to draw from the user's bionic power banks.  Twice as efficient as the Mk. I model.  The Mk. II was designed by DoubleTech Inc., to meet the popularity of the Mk. II power armor series.",
+    "price": 1000000,
+    "price_postapoc": 500,
+    "weight": "400 g",
+    "difficulty": 8,
+    "extend": { "flags": [ "OBSOLETE" ] }
+  },
+  {
+    "id": "bio_power_armor_interface",
```
irwiss, line 293: Same with actual bionic definitions - moving them around in files won't do much as characters will still have them as part of their character save json and at some point you'll have to delete them; you need to delete these entries and make a `bionic_migration` entry as described in the doc linked above to similar themed bionic or just nothing  to make them delete when loaded, similar to how it's done in https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/json/obsoletion/migration_bionics.json

data/json/obsoletion/obsolete-0.G-recipes.json
```diff
@@ -427,5 +427,15 @@
     "result": "xs_k_gambeson_hood",
     "type": "recipe",
     "obsolete": true
+  },
+  {
+    "type": "uncraft",
```
irwiss, line 432: Non-existing recipes just get deleted silently; you should be able to remove these entirely, (and these 0.G files shouldn't be updated for new obsoletions)

PR #65078, opened 2023-04-14
add second brighter mode to cranial flashlight

anoobindisguise:
#### Summary
Balance "Cranial Flashlight can get brighter"

#### Purpose of change
Cranial Flashlight was an oddly pointless bionic being extremely dim, only a bit brighter than a smartphone.

#### Describe the solution
Give it a secondary mode, which is as bright as a heavy duty flashlight. 

Mess with the power costs.
HD flashlight uses 15w, the bionic uses 12w. It could be changed to to use 15w. (Done)
The low vis mode is also a too-high power cost for its light level, being 1/5th the brightness of a headlamp but 1/3 the power cost so it could be lowered to 2w or even 1500 mW (and LED tattoo lowered further) (done, it is 2w for the normal mode)

#### Testing
Tested, made sure no errors, and it functions correctly.


nexusmrsep:
What you said about power cost is reasonable, I'd encourage you to go for it.

PR #65081, opened 2023-04-15
Deduplicate jackhammer and pickaxe use functions

sparr:
#### Summary
None

#### Purpose of change
iuse::jackhammer and iuse::pickaxe are almost entirely identical

#### Describe the solution
Move the duplicated functionality out to a helper function

#### Testing
test suite, and used both tools in-game in a few situations


sparr:
These shared a function a long time ago and it got duplicated back when there was less code and more dissimilarity. Since then the code has grown and the similarity has increased to the point that I think this consolidation makes sense again.

PR #65089, opened 2023-04-15
Port cata_utility to string_view

jbytheway:
#### Summary
Infrastructure "Port cata_utility functions to string_view"

#### Purpose of change
Modernization and simplification of code.

#### Describe the solution
Convert various of our string utility functions to use `string_view` arguments.

Doing more in this PR.

#### Testing
Unit tests.

#### Additional context
This was a fairly formulaic change.  None of these functions had `char*` overloads, so we don't get to simplify any overload sets.  This is primarily so that other functions using these utility functions can be ported to `string_view` going forwards.

### DIFF
--- a/src/cata_utility.h
+++ b/src/cata_utility.h
@@ -424,35 +424,18 @@ inline void deserialize_from_string( T &obj, const std::string &data )
 /**
  * \brief Returns true iff s1 starts with s2
  */
-bool string_starts_with( const std::string &s1, const std::string &s2 );
-
-/**
- * Returns true iff s1 starts with s2.
- * This version accepts constant string literals and is ≈1.5 times faster than std::string version.
- * Note: N is (size+1) for null-terminated strings.
- */
-template <std::size_t N>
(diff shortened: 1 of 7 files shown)


src/cata_utility.h
```diff
@@ -433,15 +433,15 @@ bool string_starts_with( const std::string &s1, const std::string &s2 );
  */
 template <std::size_t N>
 // NOLINTNEXTLINE(modernize-avoid-c-arrays)
-inline bool string_starts_with( const std::string &s1, const char( &s2 )[N] )
+inline bool string_starts_with( const std::string_view s1, const char( &s2 )[N] )
```
andrei8l, line 436: Is this still needed if you make the one above inline/constexpr? They look the same on godbolt https://godbolt.org/z/9K5dafK88

jbytheway: Looks like you're right (I checked they were still the same even on gcc 9.2).  I'll consolidate them.  Thanks for researching that.

PR #65094, opened 2023-04-15
[bombastic perks] updated the startup trigger for bombastic perks

### LINKED ISSUE #64832: [Bombastic Perks] Gaining any exp results in perk gain before menu initialization.
### Describe the bug

If you happen to kill enemies before 30 seconds pass and you can pick default values ( or custom ) for bombastic perks, you will receive a perk immediately. In one a hospital start using inflated enemy numbers I managed to get two perks before initialization.

### Attach save file

N/A

### Steps to reproduce

1. Start in a place where there might be enemies easily accessible
2. Manage to kill one within 30 turns (or the menu pops up for options)

### Expected behavior

I expect default values to be already in the system at launch.

### Screenshots

_No response_

- OS: Windows
- Game Version: e084982 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm]
]

### Additional context

_No response_

bombasticSlacks:
#### Summary
Mods "updated the startup for BP to be more reliable"

#### Purpose of change
Makes the config menu trigger immediately, I thought this would lead to breakage, but testing it, it seems fine.

fixes #64832

#### Describe the solution
Made the menu popup right away, added info about how to start the mod if it is broken.

### DIFF
--- a/data/mods/BombasticPerks/corefiles/core_eocs.json
+++ b/data/mods/BombasticPerks/corefiles/core_eocs.json
@@ -82,7 +82,9 @@
     "required_event": "game_start",
     "condition": { "and": [ { "not": { "u_has_trait": "perk_perk_menu" } } ] },
     "effect": [
-      { "u_message": "Bombastic Perks is loaded, configuration menu should appear.  If the mod is properly loaded you should have a new mutation called \"perk menu\".  If you do not, you can enable the mod manually by activating the \"EOC_bombastic_perks_try_configure\" EOC in debug." },
+      {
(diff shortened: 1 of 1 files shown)


GuardianDll:
Why make an EoC , that trigger another EoC, which make the config menu, rather than make a direct EoC, that instantly make the config menu?

bombasticSlacks:
Since the rules for non activation EOCs being queued might change in the future. 

Also the dialogue that spawns at start shouldn't be repeated if you somehow crash the dialogue menu.

PR #65095, opened 2023-04-15
Adding Makeshift Pressure Cooker

TaiGambol:
#### Summary
Content "Makes stock pots craftable, add craftable makeshift pressure cooker"

#### Purpose of change
Pressure cookers are rarer than they probably ought to be; now that they're required for food preservation, this is a headache, and from what I've been able to look up, they aren't too hard to DIY. Certainly not beyond the capabilities of a skilled survivor.

#### Describe the solution
Adds a Makeshift Pressure Cooker and accompanying crafting recipe, based loosely on this Youtube video https://www.youtube.com/watch?v=_Filk4I4Axg that demonstrates creating a homemade pressure cooker or still. The recipe requires a stock pot; while these seem much more common than pressure cookers, to prevent shifting the problem to searching for one of those instead, I've also added a crafting recipe for a stock pot, that's a direct clone of the canning pot recipe as everything I've looked up implies the two are not only mostly the same but often interchangeable.

Also added Makeshift Pressure Cooker to canning_heat_high and canning_metal so it's recognized as a suitable tool.

In comparison to the regular pressure cooker, the makeshift pressure cooker is nearly twice the bulk; if it should have any other drawbacks, I'm open to suggestions but unsure what or how to add them.

Adjusting the spawn rate of pressure cookers and stock pots, adding them to more spawn groups. I'm less sure how to do this than I am how to make a recipe, though, and nothing says I can't do both once I figure that out.

Considered using the Canning Pot as the item for crafting the pressure cooker, instead, as that's already craftable and would save making the stock pot recipe.

#### Testing

Loaded test game, no errors.
Gave character all recipes, searched for 'Stock Pot', stock pot shows up.
Searched for 'pressure', makeshift pressure cooker showed up.
Searched for 'apple cider', makeshift pressure cooker shows as a valid tool.
Searched for 't:makeshift pressure', assorted canning/jarring recipes showed up as expected.

#### Additional Contest
Shoutout to Candy on the community Discord server, whose terrible luck inspired me to make this PR.


sparr:
json build error seems real
json linting error also seems real

I think the canning pot would be a good component for the recipe here, perhaps as an alternative to the stock pot?

TaiGambol:
> json build error seems real json linting error also seems real
> 
> I think the canning pot would be a good component for the recipe here, perhaps as an alternative to the stock pot?

I'll have to fiddle with working out what's causing the errors.

My only concern with subbing in the canning pot is it's twice the size of the stock pot, so should logically result in the makeshift pressure cooker being twice the size too; I don't know how or even if I can make an item have different weights/sizes depending on the components it's made of?

TaiGambol:
Sorry for leaving this so long, I've fixed it now. Can I get a review?

Issue #63804, opened 2023-02-26
Ki Strike damage blocked by wearing gloves


Terrorforge:
### Describe the bug

If you have "Ki Strike" and wear any kind of gloves or unarmed weapon, the extra damage from Ki Strike does not get applied. This happens even if the "glove" in question is integrated armor.

Presumably this is a side effect of the changes to unarmed weapons, going from wielded weapons to worn gloves, but I haven't been able to find the exact point in the code where it breaks.

### Attach save file

[MA Test-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10833334/MA.Test-trimmed.tar.gz)

### Steps to reproduce

1. Give yourself Ki Strike and high Melee/Unarmed skill. Do not wear gloves.
2. Hit some zombies and note the damage.
3. Put on any gloves.
4. Hit some more zombies and note how much lower your damage is.

### Expected behavior

Ki Strike should still work while wearing gloves.

Based on references in the code (melee.cpp lines 1325-1328) Ki Strike is supposed to apply on top of unarmed weapon damage. And just from a gameplay perspective, it would be one thing if it couldn't be used in conjunction with unarmed weapons like skewer knuckles, but you can't wear gloves at all. And obviously it's complete nonsense that it doesn't work with integrated armor.

### Screenshots

![image](https://user-images.githubusercontent.com/5107449/221401733-8a358806-23c5-4261-88d9-3b47f78b9f21.png)
This is a character with Ki Strike, 10 in every skill and debug incincibility active. Note how when I put on the tactical gloves (middle of the log) I go from dealing 28 damage on a normal hit to struggling to even reach 24 on a critical.

![image](https://user-images.githubusercontent.com/5107449/221402059-0c92f160-5c3a-4e94-a7cd-10ec369e22e2.png)
Same with integrated armor.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.2604 (22H2)
- Game Version: f1bc060 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Mythical Martial Arts [MMA]
]

### Additional context

[...]

PR #65098, opened 2023-04-15
Clarify Ki strike description

adamkad1:
#### Summary
Mods "Clarify ki strike description"

#### Purpose of change

'Fighting unarmed' does not suggest that gloves impede the damage bonus, and I can clearly see my damage decrease when i wear gloves or unarmed weapons

#### Describe the solution

Clarify the description to make it more obvious how it works. If its not supposed to work that way though, someone needs to fix it though the mut effect is not a json thing.

N/A

#### Testing

Check game, description changed

#### Additional context

Barehanded implies no gloves, but some people are dumb I guess, so extra clarification doesnt hurt.

### DIFF
--- a/data/mods/MMA/mutations.json
+++ b/data/mods/MMA/mutations.json
@@ -44,7 +44,7 @@
     "id": "KI_STRIKE",
     "name": { "str": "Ki Strike" },
     "points": 2,
-    "description": "Who needs weapons?  You deal more melee damage while fighting barehanded (gloves included).  This damage improves as your unarmed skill increases.",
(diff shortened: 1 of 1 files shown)


Terrorforge:
Per comments from hymore246 on #63804, this is in fact unintended behavior. They said they were gonna fix it, but it'll be easy enough to change the description again when that happens.

adamkad1:
> Per comments from hymore246 on #63804, this is in fact unintended behavior. They said they were gonna fix it, but it'll be easy enough to change the description again when that happens.

Fair.

PR #65099, opened 2023-04-15
Make respirator bionic less hardcoded

anoobindisguise:
#### Summary
Balance "Respirator bionic jsonizing and balancing"

#### Purpose of change
Respirator bionic was weird. It had a highly esoteric function in c++ code and used an obscene amount of power (1 kJ per stamina point it was regenerating, meaning it could use 50-100kJ in a single second).

#### Describe the solution
Make the respirator bionic simply toggle a stamina regen enchantment and give you the REBREATHER flag. Adjust drowning code to check for if you has_flag( flag_REBREATHER ) instead of the Respirator bionic specifically.
Its interaction with asthma is still hardcoded.

This solution has a few downsides:
- It is no longer as convenient as you could previously just leave it running and it would only deduct stamina while you were using stamina, rather than idling at full stamina. This could be solved by either leaving that as hardcode and lowering the power cost, making it deactivate specifically if you are at full stamina, or specifically not consume power if you are at full stamina (hardcoding in bionics.cpp) - alternatively flags could be created to make many bionics not use power if the user doesn't need them under specific circumstances.
- It no longer auto prevents drowning, it must be activated preemptively. It might be possible to get the pointer to a gills bionic you have and forcibly activate it instead of taking drowning damage.

The power cost could also be changed, as could the stamina regen enchantment. Previously the bionic gave up to a 400% increase in regen speed according to the c++ code but that used a ridiculous amount of power and was unsustainable.

The hardcoded check in asthma suffering could also be removed and make it just cancel the asthma mutation, but we have no logic in place for making you get asthma back if it is removed. As a result it could also be changed into a permanent bionic that replaces your lungs outright and requires a baseline power cost to upkeep or else you'd suffocate.

### DIFF
--- a/data/json/bionics.json
+++ b/data/json/bionics.json
@@ -997,7 +997,18 @@
     "description": "A complex respiration augmentation system.  Improves respiration ability in air and allows breathing water.  Will automatically turn on when drowning.  Turn on to recharge stamina faster, at moderate power cost.  Asthmatics may also use it to stop asthma attacks.",
     "occupied_bodyparts": [ [ "torso", 8 ], [ "head", 2 ], [ "mouth", 2 ] ],
     "flags": [ "BIONIC_TOGGLED" ],
-    "trigger_cost": "25 kJ"
+    "act_cost": "250 J",
+    "react_cost": "250 J",
+    "time": "1 s",
+    "enchantments": [
+      {
+        "condition": "ACTIVE",
+        "values": [
+          { "value": "REGEN_STAMINA", "multiply": 1.0 }
+        ]
+      }
+    ],
+    "active_flags": [ "REBREATHER" ]
   },
   {
     "id": "bio_ground_sonar",
--- a/data/json/bionics.json
+++ b/data/json/bionics.json
@@ -1000,14 +1000,7 @@
     "act_cost": "250 J",
     "react_cost": "250 J",
     "time": "1 s",
-    "enchantments": [
-      {
-        "condition": "ACTIVE",
-        "values": [
-          { "value": "REGEN_STAMINA", "multiply": 1.0 }
-        ]
-      }
-    ],
+    "enchantments": [ { "condition": "ACTIVE", "values": [ { "value": "REGEN_STAMINA", "multiply": 1.0 } ] } ],
     "active_flags": [ "REBREATHER" ]
   },
   {
--- a/data/json/bionics.json
+++ b/data/json/bionics.json
@@ -999,6 +999,7 @@
     "flags": [ "BIONIC_TOGGLED" ],
     "act_cost": "250 J",
     "react_cost": "250 J",
+    "trigger_cost": "1 kJ",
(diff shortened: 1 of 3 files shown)


actually-a-cat:
The energy cost for respirator is actually 0.1 kJ per point, not 1 kJ, there's a division by 10 in there. That's still a lot, but IMO it's actually appropriate, at least from a balance perspective. Stamina regen is extremely powerful, so any bonus to it needs to be either difficult and expensive to use (which it is currently) or very small. +100% regen for a small power cost is just outrageously overpowered, ~doubling a characters melee potential for no downside with one bionic?

nexusmrsep:
I remember tweaking it's functions in the past. I want to offer some constructive feedback, but I realize that there might be a problem in the undefined lore of this bionic.

What is it exactly and how does it function? Are those simply artificial lungs? If so they should provide extra volume and augmented function, aka better air supply = stamina regen on land. Under water not so much, extra volume = extra long breath holding = only little more oxy reserve.

But if it has oxygen generation/storage properties, then it's a while new story. There are options here that perhaps deserve separate bionics.

[1] small oxy tank reserve, that is replenished on land by oxy concentration at cost of power and used when needed for ex. under water,

[2] rebreather oxy recycler, extracts carbon from carbon dioxide [CO2], returns O2 to the cycle, costs power

[3] electrolytic O2 supply, breaks H2O into expelled hydrogen and oxygen that is returned to the cycle, need power, need water = water tank [like generator CBMs] or direct supply when in water.

PatrikLundell:
Suffocation when out of power is really bad from a game play perspective. Activate some expensive bionic for too long? Die of suffocation. Run into magic critters that suck you power dry? Die of suffocation.

The only way to avoid the second fate is to be lucky enough to have some means of bionic power production with you at that time, and it can't be battery based, as those are drained by the same magic.

nexusmrsep:
@PatrikLundell - I wouldn't suggest becoming bionic dependent for life support functions, especially when EMP drainage is more common nowadays.

The reasonable approach for "artificial lungs" would be augmenting diaphragma rather then fully replacing it, so the bionic would function in two modes - standard (no stamina regen boost, bio operated) and supported (augmented stamina regen, active bionic power consumption).

This approach suggests that perhaps lung tissue replacement ("artificial lungs") could be also a separate bionic - a passive one that perhaps gives some reasonable bonus to max stamina and gives extra oxy reserve when diving, but maybe makes stamina less influenced by the athletics skill. Like power generation - this has the potential to become a whole branch of bionics that can coexist together.

fairyarmadillo:
No bionic short of artificial lungs should cure asthma, as the condition is caused by inflammation of the airways in the upper part of the lungs. Bionic lungs also wouldn't be removable, so that would solve the issue of no logic existing to reapply it if removed.

anoobindisguise:
Gonna close this for now, there are too many issues with it.

PR #65108, opened 2023-04-16
Hulks can see from afar

sonphantrung:
#### Summary
Balance "Improve LOS for Hulks"

#### Purpose of change
There's one thing that bugged me out: Why the heck do Mr. Hulks and his friends are the same sizes (aside from Kevlar Hulk), but their LOS are inconsistent? Plus, many people (including me), feels like we haven't had a real end-game threat for years.

#### Describe the solution
- Improve LOS of Hulks: Skeletal Juggernauts, Relentless Hulks now has 83 at daytime (heck, `No matter how far you run, it always seems to find you` now matters a lot more). Kevlar Hulks are 'smaller', so I buff his LOS only to 60 at daytime

Just not do this

#### Testing
Not yet tested, but changes are pretty simple, right?

#### Additional context
Guys, the Hunted scenario is 2x harder now, right?

### DIFF
--- a/data/json/monsters/zed_misc.json
+++ b/data/json/monsters/zed_misc.json
@@ -562,6 +562,7 @@
       { "name": "the head", "armor_mult": { "physical": 0.5 }, "coverage": 3 },
       { "name": "the eye", "armor_mult": { "physical": 0 }, "coverage": 1 }
     ],
+    "families": [ "prof_intro_biology", "prof_wp_zombie", "prof_wp_hulk" ], 
     "vision_day": 83,
     "vision_night": 8,
     "harvest": "zombie",
--- a/data/json/monsters/zed_misc.json
+++ b/data/json/monsters/zed_misc.json
@@ -562,7 +562,7 @@
       { "name": "the head", "armor_mult": { "physical": 0.5 }, "coverage": 3 },
       { "name": "the eye", "armor_mult": { "physical": 0 }, "coverage": 1 }
     ],
-    "vision_day": 50,
+    "vision_day": 83,
     "vision_night": 8,
     "harvest": "zombie",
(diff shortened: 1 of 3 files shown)


Venera3:
Relentless hulks always see you anyway (hence the `ALL_SEEING` flag), otherwise I don't really see how half a meter improves their vision by 50% above base zombies. 
On the balance side this means the big lads will be differentially kiteable and won't drag their smaller friends along, which will probably outweigh how useful it will be to them.

fairyarmadillo:
Fixing late game balance is probably going to come out of building more reasonable mechanics (ie the recent/coming grab updates) and nerfing things like the activity suit to be more reasonable, not just buffing random monsters.

Fris0uman (changes requested):
Just keep the proficiency edit, the rest doesn't really make sense as the previous comments have pointed out.

esotericist:
> Not yet tested, but changes are pretty simple, right?

independent of any other concern here (and there are few, including what venera mentioned) this isn't appropriate. you do need to test your changes, and you should at least *attempt* to validate that they actually do what you think they do. further, if you don't know how to test whether a given set of changes made an impact, you shouldn't be tampering with it.

as for the actual changes: it isn't clear what you're attempting to accomplish or how your changes actually further that objective, and without that (and the appropriate testing to demonstrate it's being accomplished), there isn't much to work with. what i can say is that the changes you describe in your PR body are unlikely to be accepted due to not having the impact you think they have.

the proficiency edit getting tossed in without a mention is less than ideal; try to document all the things you change in the future even if it's just a brief mention of additional alterations. for now, if you do want to pursue those sorts proficiency changes, they should happen in a discrete PR, as this pull request isn't going to be merged on the merits on hand.

PR #65109, opened 2023-04-16
Adds a key to toggle the language to English and back on the fly

MarcusAseth:
Adds to Keybind an entry called "Toggle language to English" (default assigned to the key "," for testing purposes). If from the game options a language different from English was selected, when using this key while playing the language will toggle between that language and English on the fly every time it is pressed.

#### Summary
Features "Adds a key to toggle the language to English and back on the fly while playing"

#### Purpose of change
This feature is of great value to two categories of people: **language learners** and **translators**.
**・For language learner** this would instantly turn the game into one of the best language learning games currently available. It's  often the case a language learner can read and maybe understand all the words in a sentence but fail to put the meaning together because of the grammar, and showing them the translation would bridge that gap. The ability of comparing two languages is found in many language learning platforms, and with 180.000 string of text, adding the power to compare two versions of a text could turn this game into a very powerful language learning tool for thousands of people if not more.
**・For translators**, I can tell you I was told someone in the Italian translation team was not Italian to begin with and was submitting horrible translations blatantly made using Google Translate, there is no telling to the damages already done from this individual and something similar might have already happened for the other translations as well. With this functionality when a translator spots an iffy translation all he has to do is press a key to be shown the source material, in turn pushing them towards continued improvement and polish of the translations, where otherwise they might not even have bothered to check.

#### Testing
I've tested it in game without any problems.

#### Additional context

Video of it working:
https://vimeo.com/820629407

### DIFF
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -2074,7 +2074,6 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
         item_info_data item_info_first;
         item_info_data item_info_second;
 
-        const int offset_y = confirm_message.empty() ? 0 : 3;
         int page_size = 0;
 
         int scroll_pos_first = 0;
@@ -2117,6 +2116,7 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
             ui.on_screen_resize( [&]( ui_adaptor & ui ) {
                 const int half_width = TERMX / 2;
                 const int height = TERMY;
+                const int offset_y = confirm_message.empty() ? 0 : 3;
                 page_size = TERMY - offset_y - 2;
                 wnd_first = catacurses::newwin( height - offset_y, half_width, point_zero );
                 wnd_second = catacurses::newwin( height - offset_y, half_width, point( half_width, 0 ) );
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -2074,8 +2074,6 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
         item_info_data item_info_first;
         item_info_data item_info_second;
 
-        const int half_width = TERMX / 2;
-        const int height = TERMY;
         const int offset_y = confirm_message.empty() ? 0 : 3;
         const int page_size = TERMY - offset_y - 2;
 
@@ -2117,6 +2115,8 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
 
             ui.reset();
             ui.on_screen_resize( [&]( ui_adaptor & ui ) {
+                const int half_width = TERMX / 2;
+                const int height = TERMY;
                 wnd_first = catacurses::newwin( height - offset_y, half_width, point_zero );
                 wnd_second = catacurses::newwin( height - offset_y, half_width, point( half_width, 0 ) );
 
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -2069,19 +2069,14 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
     std::string action;
     input_context ctxt;
     ui_adaptor ui;
+    item_info_data item_info_first;
+    item_info_data item_info_second;
+    int page_size = 0;
+    int scroll_pos_first = 0;
+    int scroll_pos_second = 0;
     bool first_execution = true;
+    static int lang_version = detail::get_current_language_version();
     do {
-
-        item_info_data item_info_first;
-        item_info_data item_info_second;
-
-        int page_size = 0;
-
-        int scroll_pos_first = 0;
-        int scroll_pos_second = 0;
-
-        static int lang_version = detail::get_current_language_version();
-
         if( first_execution || lang_version != detail::get_current_language_version() ) {
             std::vector<iteminfo> v_item_first;
             std::vector<iteminfo> v_item_second;
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -2069,6 +2069,7 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
     std::string action;
     input_context ctxt;
     ui_adaptor ui;
+    bool first_execution = true;
     do {
 
         item_info_data item_info_first;
@@ -2079,7 +2080,6 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
         int scroll_pos_first = 0;
         int scroll_pos_second = 0;
 
-        bool first_execution = true;
         static int lang_version = detail::get_current_language_version();
 
         if( first_execution || lang_version != detail::get_current_language_version() ) {
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -2075,7 +2075,7 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
         item_info_data item_info_second;
 
         const int offset_y = confirm_message.empty() ? 0 : 3;
-        const int page_size = TERMY - offset_y - 2;
+        int page_size = 0;
 
         int scroll_pos_first = 0;
         int scroll_pos_second = 0;
@@ -2117,6 +2117,7 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
             ui.on_screen_resize( [&]( ui_adaptor & ui ) {
                 const int half_width = TERMX / 2;
                 const int height = TERMY;
+                page_size = TERMY - offset_y - 2;
                 wnd_first = catacurses::newwin( height - offset_y, half_width, point_zero );
                 wnd_second = catacurses::newwin( height - offset_y, half_width, point( half_width, 0 ) );
 
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -2141,6 +2141,7 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
                 draw_item_info( wnd_second, item_info_second );
             } );
             lang_version = detail::get_current_language_version();
+            first_execution = false;
         }
 
         ui_manager::redraw();
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -2077,7 +2077,7 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
     bool first_execution = true;
     static int lang_version = detail::get_current_language_version();
     do {
-        //lang check here is needed to redraw the menu while using "Toggle language to English" option
+        //lang check here is needed to redraw the menu when using "Toggle language to English" option
         if( first_execution || lang_version != detail::get_current_language_version() ) {
             std::vector<iteminfo> v_item_first;
             std::vector<iteminfo> v_item_second;
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -2066,72 +2066,82 @@ drop_locations game_menus::inv::smoke_food( Character &you, units::volume total_
 bool game_menus::inv::compare_items( const item &first, const item &second,
                                      const std::string &confirm_message )
 {
-    std::vector<iteminfo> v_item_first;
-    std::vector<iteminfo> v_item_second;
-
-    first.info( true, v_item_first );
-    second.info( true, v_item_second );
-
-    int scroll_pos_first = 0;
-    int scroll_pos_second = 0;
-
-    item_info_data item_info_first( first.tname(), first.type_name(),
-                                    v_item_first, v_item_second, scroll_pos_first );
-
-    item_info_data item_info_second( second.tname(), second.type_name(),
-                                     v_item_second, v_item_first, scroll_pos_second );
-
-    item_info_first.without_getch = true;
-    item_info_second.without_getch = true;
-
-    input_context ctxt;
-    ctxt.register_action( "HELP_KEYBINDINGS" );
-    if( !confirm_message.empty() ) {
-        ctxt.register_action( "CONFIRM" );
-    }
-    ctxt.register_action( "QUIT" );
-    ctxt.register_action( "UP" );
-    ctxt.register_action( "DOWN" );
-    ctxt.register_action( "PAGE_UP" );
-    ctxt.register_action( "PAGE_DOWN" );
-
-    catacurses::window wnd_first;
-    catacurses::window wnd_second;
-    catacurses::window wnd_message;
-
-    const int half_width = TERMX / 2;
-    const int height = TERMY;
-    const int offset_y = confirm_message.empty() ? 0 : 3;
-    const int page_size = TERMY - offset_y - 2;
-
-    ui_adaptor ui;
-    ui.on_screen_resize( [&]( ui_adaptor & ui ) {
-        wnd_first = catacurses::newwin( height - offset_y, half_width, point_zero );
-        wnd_second = catacurses::newwin( height - offset_y, half_width, point( half_width, 0 ) );
-
-        if( !confirm_message.empty() ) {
-            wnd_message = catacurses::newwin( offset_y, TERMX, point( 0, height - offset_y ) );
-        }
-
-        ui.position( point_zero, point( half_width * 2, height ) );
-    } );
-    ui.mark_resize();
-    ui.on_redraw( [&]( const ui_adaptor & ) {
-        if( !confirm_message.empty() ) {
-            draw_border( wnd_message );
-            mvwputch( wnd_message, point( 3, 1 ), c_white, confirm_message
-                      + " " + ctxt.describe_key_and_name( "CONFIRM" )
-                      + " " + ctxt.describe_key_and_name( "QUIT" ) );
-            wnoutrefresh( wnd_message );
-        }
-
-        draw_item_info( wnd_first, item_info_first );
-        draw_item_info( wnd_second, item_info_second );
-    } );
-
     std::string action;
-
+    input_context ctxt;
+    ui_adaptor ui;
     do {
+
+        item_info_data item_info_first;
+        item_info_data item_info_second;
+
+        const int half_width = TERMX / 2;
+        const int height = TERMY;
+        const int offset_y = confirm_message.empty() ? 0 : 3;
+        const int page_size = TERMY - offset_y - 2;
+
+        int scroll_pos_first = 0;
+        int scroll_pos_second = 0;
+
+        bool first_execution = true;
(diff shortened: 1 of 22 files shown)


PatrikLundell:
It seems like a useful niche functionality. However, since it's a niche functionality, it shouldn't be bound to anything by default, but rather should have to be manually bound by those who want it. We're already short of available keys, and it doesn't help that some keys don't work with non English keyboards (they did in 0.E, but not in 0.F. I'm talking about all keys that are used to "construct" composite characters, and thus won't generate a visible character until pressed twice, such as the caret "^" used for vehicle interaction by default).

MarcusAseth:
I agree, I had it bound for test purpose then forgot to unbind, I've now unbounded it :)

src/translations.cpp
```diff
@@ -73,14 +73,19 @@ std::string locale_dir()
     return loc_dir;
 }
 
-void set_language()
+//if called without any parameters sets the language to the one selected in the options menu
```
BrettDong, line 76: This makes the function to serve multiple responsibilities at once:

- Get language option
- Set to a specific language

And you made this function to perform its different responsibilities depending on a default parameter.

It would be better if you could break them to their respective functions for better readability.

src/translations.cpp
```diff
@@ -73,14 +73,19 @@ std::string locale_dir()
...
+//if called with a parameter from code sets the language to that instead
+void set_language( std::string lang )
```
BrettDong, line 78: Passing a string by copy is not really necessary as you are not changing this variable in the function body.

src/translations.cpp
MarcusAseth: Earlier today I had it separated into two functions: "set_language()" currently present in the game and "toggle_language()" which essentially copied the body of the first function and did almost the same, but duplicated the size.
It looked like this:  https://pastebin.com/aW4k21Ah
Should I go back to that?

BrettDong: Of course duplicating code is bad. I mean separating the responsibilities. Currently you squeezed two different tasks in one function:

- Get language settings
- Set the game to a specific language

MarcusAseth: To be completely honest, I'm not entirely sure I'm following ^_^' the reason is that "set_language()" is currently part of the game and is doing those two things as well, is getting the language set in the option and is setting the language to it.
So I guess I should start splitting that into two before adding mine? I will try  :  )

BrettDong: It made sense before because it would always get language option and set to that language. But now you've introduced a defaulted parameter and depending on whether the parameter is provided or not, the function can either

- Get language option and set to that language
- Set language to the parameter provided

I don't feel it's still a clean and readable interface design.

MarcusAseth: I think I got you now, sorry if I am a bit slow to see these things, maybe it will come with time ^_^' 
I think I've made the correct changes, I'll compile and then then push it in 10 minutes

src/translations.h
```diff
@@ -97,5 +97,6 @@ inline const char *npgettext( const char *const context, const char *const msgid
 std::string locale_dir();
 
 void set_language();
+void change_language( std::string& lang );
```
BrettDong, line 100: It should be a const reference.

src/translations.cpp
```diff
@@ -79,8 +79,15 @@ void set_language()
     const std::string system_lang = SystemLocale::Language().value_or( "en" );
     std::string lang_opt = get_option<std::string>( "USE_LANG" ).empty() ? system_lang :
                            get_option<std::string>( "USE_LANG" );
-    DebugLog( D_INFO, D_MAIN ) << "Setting language to: '" << lang_opt << '\'';
-    TranslationManager::GetInstance().SetLanguage( lang_opt );
+    change_language( lang_opt );
+#endif
+}
+
+void change_language( std::string &lang )
```
BrettDong, line 86: You only changed declaration in header but not the implementation here.

MarcusAseth: oh, sorry about that x_x

kevingranade:

https://discourse.cataclysmdda.org/t/mod-suggestion-easily-turning-this-game-into-the-most-powerful-language-learning-game/28681/3

src/input.cpp
```diff
@@ -1189,6 +1189,25 @@ const std::string &input_context::handle_input( const int timeout )
...
+                if( !input.sequence.empty() ) {
+                    char keybind = input.sequence.front();
+                    if( *pressed_key.c_str() == keybind ) {
+                        g->toggle_language_to_en();
+                        g->invalidate_main_ui_adaptor();
+                        ui_manager::redraw_invalidated();
+                    }
+                }
+            }
+        } //
```
Qrox, line 1209: This feels too intrusive to me. It uses a keybind that is not registered to the specific context so there might be conflicts and people cannot reassign it for the specific context, unlike all other actions.

It might be better to make it work similar to "HELP_KEYBINDINGS", and you can probably register it by default if you don't feel like registering it manually everywhere `input_context` is used.

Also, comparing `text` with the hotkey is not guaranteed to work because some keys do not have a corresponding character, and in key code mode there will not be any text, but only keys. You should compare the action string instead.

Furthermore, currently only the main UI is invalidated, which I think happens to invalidate all UIs when a save is loaded because the main UI uses the whole screen area, but it might not work when a save is not loaded (for example, in the main menu) if I'm not mistaken.

Also, I think some UIs are storing the translated text so toggling the language won't work in these UIs, but that can be addressed in separate PRs.

MarcusAseth: The fact the key is not registered is incorrect, the keybind is indeed registered to a specific context, which is DEFAULTMODE - what the code above is doing is extending it's functionality across context, because if you have a function that quickly swap languages to either help you translate the game or learn another language, the last thing you want is to having to exit the item description menu, swap language, try to find the item and access again it's description and doing that back and forth - it wouldn't be user friendly at all, so the fact this key works globally and is not meant to be re-assigned is actually a feature of this functionality. Also, the key starts unbound.

I had attempted messing with HELP_KEYBINDINGS and all attempted have failed, so unless I'm shown exactly how, I can't really achieve an implementation better than this.
Attempting registering the same key to every single context I don't think it would work either because then you would have several "Toggle language to english" in the game's keybind option menu, and there is meant to be only one entry of that that controls it across the game.

Changing it to compare the action I can do! :)

Regarding some ui not updating certain elements (like in the inventory ui) I'm currently working on that, I was pointed towards "inventory_entry::make_entry_cell_cache()" and "Prepare_layout()" so I'll be checking into it today

src/input.cpp
```diff
@@ -1189,6 +1189,22 @@ const std::string &input_context::handle_input( const int timeout )
     next_action.type = input_event_t::error;
     const std::string *result = &CATA_ERROR;
     while( true ) {
+
+        //Allows "toggle_language_to_en" to also work on contexts other than "DEFAULTMODE"
+        if( !next_action.sequence.empty() ) {
+            auto attr = inp_mngr.get_action_attributes( "toggle_language_to_en", "DEFAULTMODE" );
+            if( !attr.input_events.empty() ) {
+                input_event input = attr.input_events.front();
```
BrettDong, line 1197: I don't think copying the input event object is really necessary.

MarcusAseth: const& would be ok, right?

src/input.cpp
Qrox: I think you can remove the `"DEFAULTMODE"` context so it becomes a global keybinding. Then you can register it in the constructors of `input_context` so every input context has it by default. Then you can handle it similar to how the code handles `"HELP_KEYBINDINGS"` below. The player can then register a global hotkey for the action using `=`, so it won't make them have to register it again unless they want to solve a conflict, and it will only be shown once in the keybindings menu, but in every input context so it can be rebound whenever you want.

MarcusAseth: > I think you can remove the "DEFAULTMODE" context so it becomes a global keybinding. 
If I remove the "ctxt.register_action( "toggle_language_to_en" );" under the DEFAULTMODE" context, then there is not an entry under keybind for this functionality. 

>What do you think about the UI invalidation problem I mentioned? Does it currently work in main menu, world creation UI, etc?
The main ui when you start the game has mixed results - The main menu itself (as soon as you launch the game) is not updating, while in the world generations menu and similar ui on top of the starting menu, it works but there you need also to move the mouse after you press the "toggle" key to make it refresh/redraw, it seems.
If you guys can tell me how to refresh that manually would be great, but honestly, even if this ended up not being solved I don't think it would be a reason to lose sleep on it, since the 99% of the reading is performed inside inventory and item description, that is where I'm focusing my attention first.  Currently, they key responds inside the inventory but not from inside an "item info" panel so that needs figuring out

Qrox: I mean removing the `"category": "DEFAULTMODE"` line from `keybindings.json`. An action without a category takes effect in all input contexts.

And if you want to invalidate all UIs, you can create a new function in `ui_manager.cpp` that sets the `invalidated` flag of all `ui_adaptor`s on `ui_stack` and call `invalidation_consistency_and_optimization` (i.e. similar to `ui_adaptor::invalidate` but invalidates all UIs). Then you can call that function when switching language so all UIs will be refreshed on the next redraw.

MarcusAseth: I've attempted removing the "category" entry from the json, it makes it disappear from the keybinds menu :\

MarcusAseth: Maybe is because I've also removed "ctxt.register_action( "toggle_language_to_en" );" using the "DEFAULTMODE" context.  
You said "Then you can register it in the constructors of input_context so every input context has it by default.", maybe that's the crucial step I'm missing, but I don't know where exactly I should do that because the input_context constructor is very strange, and what context I should use to register this

MarcusAseth: ok, I've attempted using the function register_action("toggle_language_to_en"); in what it seemed to me the point you've suggested, and it worked - although I'm starting a new branch because otherwise I'm bound to mix things up with the way it currently works, and if it fails I don't want to lose current progress x_x
Hope is ok if I'll be asking further questions

MarcusAseth: Can you go a bit into detail about what do you mean with "**Then you can handle it similar to how the code handles "HELP_KEYBINDINGS" below.** "

MarcusAseth: I will try :D

MarcusAseth: I think I've followed your direction correctly, although unfortunately the starting menu is still not getting refreshed :\

Qrox: It is all strings not changing or some strings not changing? The main menu stores some of its translated strings so these are expected to not change.

MarcusAseth: The only strings changing are the one showing the Version of the game, the one that says "Bugs?  Suggestions?  Use links in MOTD to report them." and the string saying "Tip of the day:" but not the tip itself.   Everything else stays unchanged.
But honestly, that's not a big, since this feature is mostly used/useful in inventory and item descriptions.  Currently with some help I've managed to make all the item strings in the inventory re-cache (is that the term?) so inventory is getting fully updated when swapping language, but I haven't figured out yet a good way to do this for the menu with the item description itself ( should be the one associated to the draw_item_info() function )  -  if you have any suggestion for that it would be much appreciated and likely would make this PR ready for merge ^_^

Qrox: Yeah, that's what I expected for the main menu because the other strings are stored as translated strings when entering the menu.

The item info also stores the translated strings so you probably need to regenerate the item info as a whole. I'd recommend adding a new overload of `draw_item_info` that takes a `std::function<item_info_data()>` instead of a `item_info_data`, and call that function at the start and when the language changes to get the updated item info data.

MarcusAseth: thank you for the suggestion, although I'm seeying it only now and in the meantime I was solving this differently, I've simply moved the code that draws the inventory_item_info panel into the loop below that code so that it can be executed again if the language changes

MarcusAseth:
Good news guys, I think this PR is completed and ready to be reviewed and merged :D
The function works correctly in game, inside inventory, inspecting/comparing items, inside building, crafting and player info menu.
Probably more that that but I'm not aware of all the menus in this game and those handle 99% of the usage, so those are the ones I've covered :)
Looking forward to study japanese with it, in the hope it will be merged ^_^

src/inventory_ui.cpp
```diff
@@ -986,11 +986,17 @@ void inventory_entry::reset_entry_cell_cache() const
 const inventory_entry::entry_cell_cache_t &inventory_entry::get_entry_cell_cache(
     inventory_selector_preset const &preset ) const
 {
+    bool has_invalid_lang_version = entry_cell_cache->lang_version !=
+                                    detail::get_current_language_version();
```
andrei8l, line 990: You're dereferencing `entry_cell_cache` here without verifying that it has a value. Just move it to the `if( !entry_cell_cache ...)` below. The second `if` is redundant anyway.

MarcusAseth: sorry andrei, I'm not 100% sure I'm following when you say "Just move it to the if( !entry_cell_cache ...) below." .  You mean moving that check at line 989 inside the first if block?  Unless I'm missing something that needs to stay outside, because there are cases in which the entry_cell_cache is valid so that if block would be skipped without that OR condition check.
Maybe I'm just misunderstanding what you meant, if it's not trouble could you show me the code you meant?

MarcusAseth: this is what I think you meant:

> if( !entry_cell_cache ||
>         entry_cell_cache->lang_version != detail::get_current_language_version() ){
>         ...
> }
Can you confirm it's correct? Even in this case aren't we dereferencing entry_cell_cache without checking it first? Or maybe is something about the order of the checks I'm not understanding very well. Anyway, can you confirm the change is correct? : )

BrettDong: `!entry_cell_cache` does the check, and if it does not contain a valid value, the or logic operator is short-circuited so the second condition will not be evaluated.

src/game_inventory.cpp
```diff
@@ -2076,72 +2076,82 @@ drop_locations game_menus::inv::smoke_food( Character &you, units::volume total_
...
-    second.info( true, v_item_second );
+        item_info_data item_info_first;
+        item_info_data item_info_second;
 
-    int scroll_pos_first = 0;
-    int scroll_pos_second = 0;
+        const int half_width = TERMX / 2;
+        const int height = TERMY;
+        const int offset_y = confirm_message.empty() ? 0 : 3;
+        const int page_size = TERMY - offset_y - 2;
```
Qrox, line 2090: While you're at it could you move these into the lambda inside the call to `on_screen_resize` below? These values depend on the window size so they should be updated when the window is resized.

src/input.cpp
Qrox: I was thinking about the item info panel opened by pressing `e` inside the invetory UI, which I think has its own input loop. Does that also work?

Qrox:
Another issue that I just realized is that only the current UI handling the input will refresh its cached strings, while other lower UIs will keep their original cache, unless the cache update code is moved inside the `on_redraw` callback of the `ui_adaptor`s.

src/game_inventory.cpp
MarcusAseth: sure, but I'm not proficient with lambda and I'm not sure what I'm doing here, could you tell me to which line I should move half_width and height to?   offset_y and page_size need to stay outside the if() block scope because page_size is used later on in the code

MarcusAseth:
@Qrox  can you give me an example of where this happen so I can see in-game?

src/game_inventory.cpp
```diff
@@ -2084,8 +2084,6 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
         item_info_data item_info_first;
         item_info_data item_info_second;
 
-        const int half_width = TERMX / 2;
-        const int height = TERMY;
         const int offset_y = confirm_message.empty() ? 0 : 3;
         const int page_size = TERMY - offset_y - 2;
```
Qrox, line 2088:
```suggestion
        int page_size = 0;
```

Qrox:
Unfortunately I cannot open that video for the moment due to network reasons. An easy way to check for the issue is to open the keybindings menu in any of the menus you changed and press the toggle language key, and see if the latter menu is refreshed while the keybindings menu is open.

src/game_inventory.cpp
```diff
@@ -2127,6 +2125,8 @@ bool game_menus::inv::compare_items( const item &first, const item &second,
 
             ui.reset();
             ui.on_screen_resize( [&]( ui_adaptor & ui ) {
+                const int half_width = TERMX / 2;
+                const int height = TERMY;
```
Qrox, line 2129:
```suggestion
                const int height = TERMY;
                page_size = TERMY - offset_y - 2;
```

MarcusAseth:
> Unfortunately I cannot open that video for the moment due to network reasons. An easy way to check for the issue is to open the keybindings menu in any of the menus that regenerate cached strings and press the toggle language key, and see if the latter menu is refreshed while the keybindings menu is open.

Ok, I see what you mean, I've tested it opening the keybind menu on top of the Compare menu and item_info menu, and as soon as you close the keybind menu those below are refreshed to the correct language. Honestly I would see this as a non issue both because how rare this situation seems to be but also because if you're using a toggle language feature then 99% of the cases you care for what you're reading, and not the menu below (otherwise you would have opened it in the menu below).
If I'm given instructions I can follow to solve this problem, I will solve it : )

Qrox:
I agree it is not a very big deal, and it might be tricky to fix it in some UIs so we can leave it for later.

MarcusAseth:
I've just noticed a bug caused by my code =_=    If I open the World Creation panel and then close it, the game throws an exception, even without using anything related to my keybind.  You guys have any idea where I might have caused this?  I suspect it might be in the "input_context::handle_input()" function - I'll try commenting out that code and see if the problem is there

MarcusAseth:
I had to rebase to current upstream/master because my branch was based on an old-buggy master

MarcusAseth:
guys if there is anything I can do to increase the chances of it being merged, or to speed up the process, let me know. That Clang-tidy build error I believe is not my fault, but if there is something I can do to clear it up let me know.

Qrox (approved):
It looks good to me, but someone with merge permission will have to test and merge it.

MarcusAseth:
In that case, I'll provide a clear explanation on how to use it:
1) From Keybinds, add a global keybind to "Toggle language to English"
2) From the option menu, change the language to something other than English
3) Use the keybind in game to toggle between choosen language and English

PR #65111, opened 2023-04-16
item_pocket: add support for default pocket settings

andrei8l:
#### Summary
None

#### Purpose of change
Allow item definitions to specify default priority and white/blacklists for pockets. This can be used to reduce the incidence of items getting picked up into snack boxes or canteens instead of backpacks.

#### Describe the solution
Add the minor infrastructure for default pocket settings.

Migrate all items with `COLLAPSE_CONTENTS` to default priority -1 and default collapsed

Add priority 1 to all backpack pockets that didn't have `ripoff` or `extra_encumbrance`

TODO:
- [x] add some default priorities to start things off
- [x] migrate `COLLAPSE_CONTENTS` to a pocket setting too

N/A

#### Testing
Test unit.

#### Additional context
This can be reviewed by commit.

### DIFF
--- a/data/json/items/armor/storage.json
+++ b/data/json/items/armor/storage.json
@@ -18,6 +18,7 @@
         "max_contains_volume": "25 L",
         "max_contains_weight": "30 kg",
         "max_item_length": "40 cm",
+        "priority": 1,
         "moves": 300
       },
       {
@@ -87,6 +88,7 @@
         "max_contains_weight": "50 kg",
         "max_item_length": "60 cm",
         "magazine_well": "12 L",
+        "priority": 1,
         "moves": 450
       },
       {
@@ -95,6 +97,7 @@
         "max_contains_weight": "8 kg",
         "max_item_length": "35 cm",
         "magazine_well": "1 L",
+        "priority": 1,
         "moves": 450
       },
       {
@@ -103,6 +106,7 @@
         "max_contains_weight": "2 kg",
         "max_item_length": "25 cm",
         "magazine_well": "500 ml",
+        "priority": 1,
         "moves": 450
       },
       {
@@ -111,6 +115,7 @@
         "max_contains_weight": "2 kg",
         "max_item_length": "25 cm",
         "magazine_well": "500 ml",
+        "priority": 1,
         "moves": 450
       },
       {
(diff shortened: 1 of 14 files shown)


irwiss:
May be have it as a separate object and reference instead of inline of pocket?
I imagine there'll be lots of reuse, even multiple pockets of same item could use the same default settings

andrei8l:
> May be have it as a separate object and reference instead of inline of pocket? I imagine there'll be lots of reuse, even multiple pockets of same item could use the same default settings

I've decided against that because it was too bulky and instead you can now specify either an array of `itype_id` or an `item_group_id`.
```JSON
        "whitelist": "test_tools_group",
        "blacklist": [ "fridge_test" ],
```

irwiss:
imo it's too bulky when you look at one item in isolation, if you start defining these on a per itype basis the bulk slims downs in relation to the individual work you need to do, e.g. if you'll now discover your newly added "(-1 priority)+collapsed" preset needs to add another field 
or a new category needs to be added you'll then have to find the 50ish places and edit each, instead of changing just 1 object, but i suppose it's up to you

bombasticSlacks:
I think we'll need to add some info to the item UI showing pocket white/blacklist. Since otherwise, a person might be confused why certain things don't fit in certain containers, or their character is showing available volume, but they can't fit the item.

This potentially should have a toggle, I feel like some veteran players aren't going to want these setting autopopulated for them.

Zireael07:
I think the UI concerns shouldn't be a blocker to this getting in, though - default pocket settings are SORELY needed

bombasticSlacks:
> I think the UI concerns shouldn't be a blocker to this getting in, though - default pocket settings are SORELY needed

I agree.

Kamayana:
> This potentially should have a toggle, I feel like some veteran players aren't going to want these setting autopopulated for them.

One idea I had was to add a "Packrat" trait during character creation that maintains the current "space efficiency at any cost" behavior, as an alternative to adding a setting.

Zireael07:
@Kamayana: Plz no hiding that stuff behind traits. It'd be a return of the Hoarder trait or whatever it was called, that rightfully got nuked with the move to pockets iirc

andrei8l:
I need to rethink this a little bit but don't have time/interest at the moment. I'll bring this back when I can.

PR #65112, opened 2023-04-16
Obsolete non-modular rigs

irwiss:
#### Summary
Interface "Make modular crafting rigs the default, obsolete old craft rigs"

#### Purpose of change

#64682 added modular craft rigs, this is a follow-up that migrates/obsoletes the previous kitchen unit/foodco/kitchenmaster and similar to the new modular ones.

#### Describe the solution

The "solution" is already available in latest experimental as another recipe/item/vehicle part, the testing section describes how to try them out; this is only the migration PR.

#### Testing

Craft/install a mounted workshop or mounted kitchen, attach a hotplate/coffeemaker or install filled propane tank + attach propane cooker and try heating/cooking, detach a tool, assign an invlet, attach it back - the invlet should appear as the tool's hotkey in examine menu.

Spawn a few vehicles with crafting rigs - RVs, welding carts, survivor rvs, luxury ones etc.
Save game, apply patch, load game - should be no errors on load (only a few warnings in debug.log) and older craft rigs should be converted to new ones, the older toolsets should spawn in the new modular rig tools, e.g. kitchen_unit should convert into veh_tools_kitchen and spawn/attach hotplate, pot and pan when game loads.


Maleclypse (commented):
Very cool,

PR #65125, opened 2023-04-16
Drop items off ledges

Leonkoithara:
#### Summary
Features "Functionality to drop items off ledges/cliffs onto creatures below"

#### Purpose of change
Add more realism and a fun mechanic of being able to crush any creature below

#### Describe the solution
It uses the multidrop function to directly move the item from the players inventory to the lowest valid z value and calculates any damage that needs to be dealt to any creature standing in that location.

Other than creating a whole different ui over the multidrop this PR seems to be quite whole

#### Testing
Tested on monsters, NPCs(currently they donot become aggro at this behaviour since they can't see who dropped the item) and vehicles(ensure the object doesn't get stashed).

#### Additional context
Thanks to whoever added the gore stuff to ultica tileset the crushing looks quite cool


GuardianDll:
It works only with items, or with furniture also?

src/game.cpp
```diff
@@ -8936,6 +8937,18 @@ void game::drop_in_direction( const tripoint &pnt )
     u.drop( game_menus::inv::multidrop( u ), pnt );
 }
 
+void game::drop_items_off_ledge( const tripoint &pnt, int ledge_height )
+{
+    units::mass wt_dropped = u.drop_items_off_ledge( game_menus::inv::multidrop( u ), pnt );
+
+    float damage = 10 * to_kilogram( wt_dropped ) * ledge_height;
```
sparr, line 8944: linear damage from height seems wrong. double height == double velocity == quadruple kinetic energy

Hirmuolio: Doubling height does not double velocity.

Remember conservation of energy.

Potential energy is directly propotional to height. Doubling height doubles potential energy.
Potential energy turns into kinetic energy in 1:1 ratio (ignoring air resistance).
So doubling height doubles kinetic energy.

sparr: Good catch. I forgot that v increases with sqrt() of d under accleration, which cancels out the v^2 for kinetic energy.

Theawesomeboophis:
Can't wait to drop a piano out of my window like in a cartoon and crush a shocker. Pretty cool feature that I could see lots of tactical potential for.

Leonkoithara:
Hi given the AIM/Drop in direction command currently behaves very differently to this option I will try to rewrite this PR to have a consistent behavior using the different drop methods. Unfortunately this means I'll have to take some more time to understand the code to add this functionality.

src/map.cpp
```diff
@@ -2726,22 +2726,41 @@ void map::drop_items( const tripoint &p )
...
         return;
     }
 
-    for( const item &i : items ) {
-        // TODO: Bash the item up before adding it
-        // TODO: Bash the creature, terrain, furniture and vehicles on the tile
+    float damage = 0.0f;
+    for( item &i : items ) {
+        units::mass wt_dropped = i.weight();
+        damage += 10 * to_kilogram( wt_dropped ) * sqrt( height_fallen );
```
irwiss, line 2742: Flat increase of damage over sum of weights doesn't make sense, this effectively grants armor penetration properties and will do more damage when selecting multiple items to drop than if you drop one at a time, unless monster has 0 bash resist

This also doesn't account for density(surface area for air resistance), collision type/momentum change over time etc

If 40 500g pillows do the same damage as 20kg battery, or 20kg mattress does the same damage as 20kg battery this seems more of a looney tunes meme than a game mechanic

src/map.cpp
```diff
@@ -2726,22 +2726,41 @@ void map::drop_items( const tripoint &p )
...
+
         add_item_or_charges( below, i );
     }
 
-    // Just to make a sound for now
-    bash( below, 1 );
+    // Bash creature standing below
+    Creature *creature = get_creature_tracker().creature_at( below );
+    if( creature ) {
+        creature->deal_damage( nullptr, bodypart_id( "head" ), damage_instance( damage_bash, damage ) );
```
irwiss, line 2750: 100% headshot chance with arbitrary "ammo" seems improbable to say the least

Leonkoithara: Yes I'm working to change this into a one_in chance of head/torso/arm_r/arm_l based on creature dexterity

src/map.cpp
Leonkoithara: This will not grant additional damage if dropped 1 at a time because damage is always 10*h^1/2*(x1 + x2 ... )

And armor penetration makes sense because this is bash damage like dropping a tank on top of an army helmet

Hmm perhaps air resistance is something that can be looked into, not sure if this is something we should add as it might cause all sorts of game aspects to go out of wack such as bow arrow movement, throwing objects, vehicle movement etc.

Also I think 40 500g pillows should do same damage as a 20kg battery if we ignore the air resistance

Leonkoithara:
Currently only works with items as extending it to furniture requires me to remove the check at ter_set() over open air the consequences of which I'm not sure.

src/map.cpp
kevingranade: Repeat that to yourself,  "40 500g pillows should deal the same damage as a 20kg battery"
That is absolutely not correct.
At falling speeds,  hardness and density are critical components of damage dealt,  this must be represented for this to be mergeable.

src/map.cpp
```diff
@@ -2726,22 +2726,59 @@ void map::drop_items( const tripoint &p )
...
 
-    // Just to make a sound for now
-    bash( below, 1 );
+    // Bash creature standing below
+    Creature *creature_below = get_creature_tracker().creature_at( below );
+    if( creature_below ) {
+        float dodge_mod = creature_below->dodge_roll();
+        dodge_mod = dodge_mod == 0 ? 1 : dodge_mod / 2;
+
+        int creature_hit_chance = rng( 0, 100 );
```
kevingranade, line 2753: The base hit chance needs to be scaled by target size,  we have a method that returns proportion of tile occupied.

src/map.cpp
```diff
@@ -2726,22 +2726,59 @@ void map::drop_items( const tripoint &p )
...
+        if( creature_hit_chance < 15 ) {
+            creature_below->deal_damage( nullptr, bodypart_id( "head" ), damage_instance( damage_bash,
+                                         damage ) );
+        } else if( creature_hit_chance < 30 ) {
+            creature_below->deal_damage( nullptr, bodypart_id( "torso" ), damage_instance( damage_bash,
+                                         damage ) );
+        } else if( creature_hit_chance < 65 ) {
+            creature_below->deal_damage( nullptr, bodypart_id( "arm_l" ), damage_instance( damage_bash,
+                                         damage ) );
+        } else if( creature_hit_chance < 100 ) {
```
kevingranade, line 2765: 100% chance of hitting a target that isn't dodging isn't a reasonable starting point.
I think starting with,  "you can make the item land at a random position in the tile at throwing 0" is a good starting point,  then the hit chance is proportion of tile occupied.
Higher throwing skill would increase this chance,  higher target dodge would reduce it.

src/map.cpp
```diff
@@ -2726,22 +2726,59 @@ void map::drop_items( const tripoint &p )
...
-    // Just to make a sound for now
-    bash( below, 1 );
+    // Bash creature standing below
+    Creature *creature_below = get_creature_tracker().creature_at( below );
+    if( creature_below ) {
+        float dodge_mod = creature_below->dodge_roll();
+        dodge_mod = dodge_mod == 0 ? 1 : dodge_mod / 2;
+
+        int creature_hit_chance = rng( 0, 100 );
+        creature_hit_chance *= dodge_mod;
```
kevingranade, line 2754: Multiplying by dodge is a bit odd here, especially since there's no way to counter it.  I think throwing - dodge to set targeting precision, then basing hit chance on that will be better.

kevingranade (changes requested):

src/map.cpp
Leonkoithara: I could add a bulk modulus component(how far the item travels after hitting another to come to a stop) to each item which could account for this but density definitely should not effect the damage.

src/map.cpp
Leonkoithara: Resolved

src/map.cpp
Leonkoithara: Resolved

bug-sniper:
Does this work with explosives and grenades too? And can I drop them down a staircase too? (Doing so would be an awesome way to unjam them, while also fixing one of my main gripes about the game)

Zireael07:
@bug-sniper  IIRC you can already drop grenades down staircases if you have z levels on

src/map.cpp
kevingranade: This is swinging too far in the other direction, a dodge_roll of 2 cuts the change to hit the target in half and it gets worse from there. Note, a monster "dodge" stat of 1 yields a dodge_roll() of 5, so it goes from "can't dodge at all" for about a 50% hit chance for most monsters to a 1/10 hit change for the tiniest amount of dodging ability, and likely several times that.

src/map.cpp
```diff
@@ -2726,22 +2727,59 @@ void map::drop_items( const tripoint &p )
...
     if( below == p ) {
         return;
     }
 
-    for( const item &i : items ) {
-        // TODO: Bash the item up before adding it
-        // TODO: Bash the creature, terrain, furniture and vehicles on the tile
+    float damage = 0.0f;
+    for( item &i : items ) {
+        units::mass wt_dropped = i.weight();
```
kevingranade, line 2742: This needs to be density and ideally factor in hardness, a really big pillow isn't going to do anything but a rock will.

src/map.cpp
Leonkoithara: Actually looking into this, since density is related to bulk modulus, using density to make this calculation instead

src/map.cpp
Leonkoithara: Added item's base material density as a damage modifier

src/map.cpp
Leonkoithara: Added character throwing as a chance modifier.

src/map.cpp
Leonkoithara: Fixed. According to my calculations:
t,d                 % hit chance
0,0                60
0,1                40
0,2                36
1,0                76
2,0                93
...

Maleclypse:
Looks like all the comments have been addressed.

PR #65136, opened 2023-04-17
Make weapons recognized as weapons

Termineitor244:
#### Summary
None

#### Purpose of change

Some objects are generally used as weapons but not classified as such by the game, the most obvious example being the cudgel which is classified as a tool but is always used as a weapon... This is problematic when you want to sort your items by category and expect to find your weapons with the, well, weapons pile.

#### Describe the solution
Makes the relevant items recognized as weapons.

#### Testing
Tested searching for the changed items in game, it works!


Karol1223:
I'm not sure about the axes being reclassified as weapons. They have good use as tools and they are, in concept, tools not weapons.

Termineitor244:
> I'm not sure about the axes being reclassified as weapons. They have good use as tools and they are, in concept, tools not weapons.

They are crafted from the weapons division on the crafting interface, and I mainly use them as weapons when playing the game, those 2 are the main reasons of the change, but if most other players use them more often as tools (I do not) I can move their crafting recipes instead so they are not crafted as weapons but classified as tools when actually made...

Karol1223:
> They are crafted from the weapons division on the crafting interface, and I mainly use them as weapons when playing the game, those 2 are the main reasons of the change, but if most other players use them more often as tools (I do not) I can move their crafting recipes instead so they are not crafted as weapons but classified as tools when actually made...

In my eyes, the category should be what the item actually is meant to be rather than what players use it as. This is also why I've started moving around some specific container 'clothing' to containers instead. So yes, I would move their crafting category. They're not really supposed to be weapons, they just happen to be good at that job similarly to stuff like wood axes or crowbars.

Termineitor244:
> In my eyes, the category should be what the item actually is meant to be rather than what players use it as. This is also why I've started moving around some specific container 'clothing' to containers instead. So yes, I would move their crafting category. They're not really supposed to be weapons, they just happen to be good at that job similarly to stuff like wood axes or crowbars.

I can follow that logic, so I will do that then when I come back from work.

bombasticSlacks:
> I can follow that logic, so I will do that then when I come back from work.

Am I still waiting on this next commit? No rush just ping me when it's done :)

Termineitor244:
> > I can follow that logic, so I will do that then when I come back from work.
> 
> Am I still waiting on this next commit? No rush just ping me when it's done :)

Yeah, sorry, I got sick yesterday and was unable to make the changes, I will ping you when they are done!

Termineitor244:
@bombasticSlacks done!!

PR #65139, opened 2023-04-17
Make the plural form of some book the same as its singular title.

kwl01skz:
#### Summary
None

#### Purpose of change
When we perform I18N, we can only see incomplete content on transifex, which is confusing.

#### Describe the solution
Make the plural form of these book the same as its singular title.

For all books, stop using plural forms.
either or

Pass on more information to transifex so that translators can understand it:
What is the original singular form of the plural entry that needs to be translated?

#### Additional context
No God From a Corpse
image

### DIFF
--- a/data/json/items/book/misc_pulp.json
+++ b/data/json/items/book/misc_pulp.json
@@ -10,7 +10,7 @@
   {
     "id": "book_fict_soft_pulp_venus",
     "type": "BOOK",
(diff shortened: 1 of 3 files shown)


Qrox:
> we can only see incomplete content on transifex

FYI you can click these buttons to switch between the singular and plural forms.

image

Issue #64916, opened 2023-04-07
Return option to disable skill rust


showervic:
### Is your feature request related to a problem? Please describe.

I was always frustrated to start each day with removing skills rust.
Having multiple skills at high levels, it was taking a hour or two each day to craft same recipes over again just to remove rust.
Maybe its realistic, but not a fun at all.
So it would be nice to return the option to disable skill rust completely. 

### Solution you would like.

Revert https://github.com/CleverRaven/Cataclysm-DDA/pull/64708

dseguin:
The option only existed because skill rust wasn't in a good enough state for regular gameplay, but it's been overhauled into a much more compelling feature since then, and the option to remove it is not coming back.

PR #65141, opened 2023-04-17
Add debug mutation to disable skill rust

neonspectra:
#### Summary
Infrastructure "Add debug mutation to disable skill rust"

#### Purpose of change

With the merge of #64708, there is no way to disable skill rust in-game without a mod. This PR adds a debug mutation to disable skill rust.

#### Describe the solution

A new mutation with ID `DEBUG_SKILL_RUST` has been added to `mutations.json`. This mutation sets `skill_rust_multiplier` to zero.

1. Return the global menu option to disable skill rust. (#64916)
    - Because this solution is a mutation that takes advantage of an existing property, it is purely json-based and does not add the extra overhead maintained by the previous debug option.
3. Create a mod that contains the added mutation instead of a debug option.
   - I think that being able to maintain a character without skill rust offers more utility as a toggle debug option than as a mod. It would make it easier to test things that require long wait time without having a character's skills rust.

#### Testing

1. I created a new character WITHOUT `DEBUG_SKILL_RUST` and WITH `DEBUG_LS` (so that I could wait without having to eat)
4. I practiced athletics. I waited one day and observed the athletics skill rust down as expected.
5. I debug-added the trait to the test character and waited a week. I confirmed that no skill rust occurred.

#### Additional context

Skill rust is and has been a polarising topic. This could easily be left as a mod, but it adds accessibility value for both testing purposes as well as for people who just don't like skill rust to include a debug option in this manner.

If the development team does not believe this should be merged into mainline, please just let me know and I'd be happy to withdraw this PR and re-implement this option as a standalone mod.

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -8099,6 +8099,18 @@
   },
   {
     "type": "mutation",
+    "id": "DEBUG_SKILL_RUST",
+    "name": { "str": "Debug Skill Rust" },
+    "points": 99,
+    "description": "Disable skill rust.",
+    "types": [ "MEMORY" ],
+    "skill_rust_multiplier": 0.0,
+    "valid": false,
+    "debug": true,
+    "cancels": [ "FORGETFUL", "GOODMEMORY" ]
+  },
+  {
+    "type": "mutation",
     "id": "SQUEAMISH",
     "name": { "str": "Squeamish" },
     "points": -1,


Fris0uman:
When do you need to turn skill rust off for debugging?

neonspectra:
> When do you need to turn skill rust off for debugging?

Personally? Pretty much never. That being said, it stands to reason that if you have a debug option to maintain a character's state (ala `DEBUG_LS`), it would stand to reason that you may want to preserve state in as much as it contains character skills as well.

---

Full cards on the table here, the main reason I'd like to merge this is to make it easier for myself and for others to choose to disable skill rust if they don't want it. I don't want there to be any misunderstandings regardng my intent of submitting this PR. I hope that's okay. 

It's easy enough to implement this is a standalone mod, and forgive my lack of understanding I suppose, but I just don't see why we *wouldn't* want a debug option here for this. If you have more insight on this, it would be very appreciated.

anothersimulacrum:
I don't think we need debug features that are not useful for debugging.

neonspectra:
Sounds good. Thanks for the feedback. I wasn't sure of the best way to submit this, so this was my first idea at an implementation that might provide some value. 

If I were to withdraw this PR and rework it as a balance mod, would the project be open to having that as a submission? I know that I'm not the only one who wants this as an option. I figured I'd at least try to make it as easy as possible for casual players to use by including it in the main repo before just going and doing my own thing with this. Any thoughts would be appreciated! 🙂

SariusSkelrets:
> If I were to withdraw this PR and rework it as a balance mod, would the project be open to having that as a submission? 

From my understanding, balance mods must do more than just remove a thing unless the thing needs improvements and even then mods for that will only remove a part of it (such as removing the fungus' exponential spread instead of the fungus itself) until a proper fix is available

If you aren't looking to do a content mod that includes disabling skill rust (like No Hope prevents zombie revive and Innawoods weakens portal storms) the available option would be preventing levelling down until the fix that makes level percentages count (so 2.99 isn't equal to 2) is ready

So your best chance IMO is to create a chargen and/or mutation trait pack containing a trait that disable skill rust. It could work as a content mod if you have some other traits with it and would allow you to create traits that wouldn't fit the main game

Just use preventing rust completly as an example of the desired power level (fully disabling skill rust will probably be way too strong an effect to fit in the perks mod) and that mod would fit next to the martial arts pack. However creating that mod will require someone to maintain it

anothersimulacrum:
No. https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/IN_REPO_MODS.md

It was removed as an option for a reason.

neonspectra:
> From my understanding, balance mods must do more than just remove a thing unless the thing needs improvements and even then mods for that will only remove a part of it (such as removing the fungus' exponential spread instead of the fungus itself) until a proper fix is available
> [...]
> So your best chance IMO is to create a chargen and/or mutation trait pack containing a trait that disable skill rust. It could work as a content mod if you have some other traits with it and would allow you to create traits that wouldn't fit the main game

Hey, thanks for the input. I don't think I have the ability to determine what is or isn't a "proper fix" because I think that it should be up to an individual to decide what the right way to play the game is.

I don't really want to make a full-on content mod because my goal here isn't to add more content. I just want to make it easy for people to use a feature that already works if they choose to do that.

neonspectra:
> It was removed as an option for a reason.

@anothersimulacrum Thanks for letting me know. I'll take this elsewhere. That being said, I'm very curious as to what the reason was. Why remove a completely optional setting when you know people want to use it?

anothersimulacrum:
I apologize, I have no interest in having that conversation right now. I realize this is not fair to you, but it has been well tread for me.

Past discussions on this topic, particularly relating to new features can provide you more information. Some handy filters in github search may be `is:unmerged` and `commenter:kevingranade`.

PR #65147, opened 2023-04-18
Fix maps changing their location

### LINKED ISSUE #61214: Trail guides and maps change their name if you travel to a different city

Inglonias:
#### Summary
Bugfixes "Fixes maps changing their location via travel"

#### Purpose of change

Fixes #61214

#### Describe the solution

This solution was stolen from one @harrison-salzverg . While they did create a fix and pull request for this issue, they unfortunately did so in their own forked repo, and this fix never made its way upstream.

Letting this continue when the solution is RIGHT. THERE.

#### Testing

Testing in progress...

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -9292,7 +9292,7 @@ std::string item::get_book_skill() const
 bool item::is_map() const
 {
     if( ( get_category_shallow().get_id() == item_category_maps ) ||
-        ( type->can_use("reveal_map") ) ) {
+        ( type->can_use( "reveal_map" ) ) ) {
         return true;
     } else {
         return false;
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -9292,7 +9292,7 @@ std::string item::get_book_skill() const
 bool item::is_map() const
 {
     if( ( get_category_shallow().get_id() == item_category_maps ) ||
-        ( get_action() == "reveal_map" ) ) {
+        ( can_use("reveal_map") ) ) {
         return true;
     } else {
         return false;
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -9291,7 +9291,11 @@ std::string item::get_book_skill() const
 
 bool item::is_map() const
 {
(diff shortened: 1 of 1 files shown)


anothersimulacrum:
Please put the author as a co-author in your commit(s).

anothersimulacrum:
https://github.com/CleverRaven/Cataclysm-DDA/pull/61439

Inglonias:
It looks like I added the co-author successfully.

I'm building the game to test this as we speak.

Inglonias:
Hrm. It looks like the `get_action` function doesn't exist anymore. I'm going to need to work out what, if anything, replaced it.

Inglonias:
Testing shows this doesn't work. Damnation.

Inglonias:
It looks like this is not the cause of the bug - I'm going to need to do a lot more digging to see what's happening here if I want to solve this.

Inglonias:
Upon further review, this is a necessary but not sufficient change to resolve this issue - This will fix the visitor's pass not counting as a map - nothing more. Why maps are wandering again, I still have yet to determine.

PR #65150, opened 2023-04-18
Forbid climbing chain fences indoors

sparr:
#### Summary
Balance "Forbid climbing chain link fences indoors"

#### Purpose of change
Most places with a chain link fence inside use them for security and they would go all the way to the ceiling. Being able to climb over the security fence in a warehouse or workshop or store mostly doesn't make sense.

#### Describe the solution
Made t_chainfence unclimbable when the player is indoors.

A new terrain type was suggested on discord, and I tried that route, but the complexity of dealing with roofs, indoors-ness, constructions, deconstructions, actions, etc was too daunting for me.

#### Testing
Tried out simple situations with fences and buildings.
Did NOT test every map that contains fences. There's a chance this change would make some starting scenario or quest impossible, if it relied on climbing over an indoor fence.

#### Additional context
Also added bad_knees penalties to more climbing situations.

### DIFF
--- a/src/iexamine.cpp
+++ b/src/iexamine.cpp
@@ -1457,6 +1457,13 @@ void iexamine::rubble( Character &you, const tripoint &examp )
  */
 void iexamine::chainfence( Character &you, const tripoint &examp )
 {
+    // If player is indoors, assume the fence goes to the ceiling and isn't climbable
+    map &here = get_map();
+    if( here.has_flag_ter( ter_furn_flag::TFLAG_INDOORS, you.pos() ) ) {
+        add_msg( m_info, _( "The ceiling prevents you from climbing that fence." ) );
+        return;
+    }
+
     // If player is grabbed, trapped, or somehow otherwise movement-impeded, first try to break free
     if( !you.move_effects( false ) ) {
         you.moves -= 100;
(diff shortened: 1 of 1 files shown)


src/iexamine.cpp
```diff
@@ -1488,17 +1495,26 @@ void iexamine::chainfence( Character &you, const tripoint &examp )
...
+            } else {
+                add_msg( _( "You vault over the obstacle with ease." ) );
+                move_cost = 100; // Not tall enough to warrant spider-climbing, so only relevant trait.
+            }
         } else {
-            add_msg( _( "You vault over the obstacle." ) );
-            move_cost = 300; // Most common move cost for barricades pre-change.
+            if( you.has_trait( trait_BADKNEES ) ) {
+                add_msg( _( "Your bad knees slow you down on the way over this obstacle." ) );
+                move_cost = 450;
```
irwiss, line 1513: Feels like this should use movecost_obstacle_modifier from the mutation and a generic faster/slower than normal message rather than try to invent new numbers or do per-mutation tree of messages ( or equivalent enchantment/limb score if one exists )

sparr: There's already a per-mutation tree of messages here, and more in slip_down. I like the flavor, and even added some to the slip_down in a recent PR.

sparr: I agree that movecost_obstacle_modifier would be more consistent and repeatable. I think it would change the existing behaviors here if we went that way, though.

LordZanos:
I don't think this makes sense. A quick google of "indoor fence" shows many examples, including of in warehouses and around things like furnaces, electrical boxes, and inside dog kennels, where the fence does not actually go to the ceiling. It shouldn't be assumed automatically that an indoor fence goes to the ceiling.

fairyarmadillo:
It really does feel like a separate terrain type would be the way to go here. This PR potentially creates as many problems as it solves.

Issue #63348, opened 2023-01-30
Ideas for future PRs 3: Modern Bowman and Incognito Police man/woman professions 


Termineitor244:
### Is your feature request related to a problem? Please describe.

More professions can always be good, right? We currently lack a profession making use of the multi-function quiver and the takedown recurve bow, so that is one easily done.

A Police man/woman in incognito, for use in the prison scenarios is also a good idea I think, so you can start with nearly nothing (dressing as a normal convict), but with knowledge and skills from a police background, even a little mission to finish your incognito investigation.

### Solution you would like.

**Modern Bowman:** Like the current bow hunter profession, but with a takedown bow and a multi-function quiver, some modern arrows and a mod or two stored in the quiver, a little more costly than the normal bow hunter, but you start with mod bows, which is pretty good since they are kind of rare at the start of most games.

**Incognito Police man/woman:** A "good/lawful" counterpart to the assassin convicts, an officer/special agent sent to infiltrate the circle of a known and powerful criminal, but that was interrupted by the cataclysm, now you need to search the body of your target for clues and try to escape a prison where it no longer matters that you are incognito, since there is nobody left to care if you are an officer or not... Common skills/proficiencies from the police officer professions, a little mission for flavor, and the police trait to be recognized as such if you find a badge.

### Additional context

Third take of my ideas while at work, this time a couple of new professions, another relatively simple change that I would like to do in the near future.

PatrikLundell:
I'd rather call the second one something like "Undercover cop", because infiltration is really an undercover operation, while being incognito just means you're not advertising who you are, while not really pretending to be someone else. The "cop" part would be to get rid of the annoyingly cumbersome political correctness "man/woman" description (which you actually don't need, so you can go with "Undercover Police" instead).
However, if you're trying to emulate reality, you'd actually need to differentiate between the genders, as prisons typically are gender separated (but the game doesn't really assign or keep genders on zombies, so a zombie may have a male sprite one time you encounter it and the other the next time if you've left the area in the mean time).

Termineitor244:
> I'd rather call the second one something like "Undercover cop", because infiltration is really an undercover operation, while being incognito just means you're not advertising who you are, while not really pretending to be someone else. The "cop" part would be to get rid of the annoyingly cumbersome political correctness "man/woman" description (which you actually don't need, so you can go with "Undercover Police" instead).
[...]

PR #65156, opened 2023-04-18
New profession: Modern Archer

Termineitor244:
#### Summary
Content "Adds a new profession called Modern Archer that uses the takedown bow and related quiver."

#### Purpose of change
Implements a part of #63348, adding a new profession to the game that makes use of the takedown bow: The Modern Archer.

#### Describe the solution
Adds the profession, an archer using modern arrows, quiver and bow, as well as some bow mods and a book about archery.

From the linked issue:

> **Modern Bowman**: Like the current bow hunter profession, but with a takedown bow and a multi-function quiver, some modern arrows and a mod or two stored in the quiver, a little more costly than the normal bow hunter, but you start with mod bows, which is pretty good since they are kind of rare at the start of most games.

#### Testing
It works!

### DIFF
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -2230,7 +2230,7 @@
     "type": "profession",
     "id": "modern_archer",
     "name": "Modern Archer",
-    "description": "A great fan of the archers of old, but at the same time a follower of the latest trends in archery.  You always tried to keep up with the times, adquiring a takedown bow as well as several accesories to complement it.  You can only hope all the training you did with them can help you survive in this trying times.",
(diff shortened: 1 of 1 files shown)


data/json/professions.json
```diff
@@ -2215,6 +2226,39 @@
       "female": [ "bra", "panties", "gold_ear" ]
     }
   },
+  {
+    "type": "profession",
+    "id": "modern_archer",
+    "name": "Modern Archer",
+    "description": "A great fan of the archers of old, but at the same time a follower of the latest trends in archery.  You always tried to keep up with the times, adquiring a takedown bow as well as several accesories to complement it.  You can only hope all the training you did with them can help you survive in this trying times.",
```
Termineitor244, line 2233:
```suggestion
    "description": "A great fan of the archers of old, but at the same time a follower of the latest trends in archery.  You always tried to keep up with the times, acquiring a takedown bow as well as several accessories to complement it.  You can only hope all the training you did with them can help you survive in this trying times.",
```

PR #65161, opened 2023-04-18
Disable android native ui default

Rimplestilt:
#### Summary
Interface "Android native UI default off"

#### Purpose of change
The initialization popup on Android has 4 options, one being "native Android UI". This is checked (enabled) by default. I highly doubt many use this feature and want to change it to default as unchecked.

#### Describe the solution
Just setting the option from true to false. (Can't test this right now. But I'm like 99 percent sure this simple fix works.

#### Testing
Have not tested. Can try later but this should be a very simple fix.

#### Additional context
Curious how many do use this feature.

### DIFF
--- a/android/app/src/main/java/com/cleverraven/cataclysmdda/SplashScreen.java
+++ b/android/app/src/main/java/com/cleverraven/cataclysmdda/SplashScreen.java
@@ -38,7 +38,7 @@ public class SplashScreen extends Activity {
     private AlertDialog accessibilityServicesAlert;
 
     public CharSequence[] mSettingsNames;
-    public boolean[] mSettingsValues = { false, false, true, true };
+    public boolean[] mSettingsValues = { false, false, true, false };
 
     private String getVersionName() {
         try {


GuardianDll:
#63148

Rimplestilt:
Oh ok... Well I was going to ask what the majority of android users prefer. What are the reasons for not wanting this?

I kind of figured it was there as an option for those who have issues using the standard UI. And as someone who has tried the native UI myself, it seems much more difficult to use and I'd only ever use it as a last resort

ZhilkinSerg:
> What are the reasons

We want this option to be enabled by default.

ZhilkinSerg:
> But why enabled? This is the 3rd time I think people are asking for it to be disabled...

Native controls are easier to use on Android devices w/o hardware keyboard.

Rimplestilt:
So, i do not want to be argumentative at all but i am genuinely curious. Do you use CDDA on android and do you use the native UI? I do not understand what this can enable/do/or make easier that the usual UI does not. It only shows up on menus, where the swiping and tapping of the screen works perfectly and feels much less clunky. It does not in any way make shortcuts or normal keybindings easier to use.

The main obstacle with CDDA on mobile are keybindings, yes. But not menus. The touch controls of being able to tap, swipe, and double tap are fully capable with menus. Any areas where greater control via keyboard bindings would help...native UI does nothing. 

If you could give an example of any menu or instance where the native UI is superior i would like it, because i can't find any. To be clear, the change i was proposing is a small one and doesn't matter that much, it's just a bit of a hassle to have to turn the option off everytime i update the game, and i figured if the majority felt the same, then it is logical to do that. Ideally the game should support mobile usage enough that native UI is only needed for those having trouble with the normal menus.

Rimplestilt:
I will say one addition that could really help mobile, would be a dedicated keyboard with a button to pop it open and closed. a small button you could drag around the screen that would open a custom qwerty keyboard.

But i think that would be difficult to implement.

Qrox:
The largest advantage I see in naitve UI is that you can drag more smoothly and tap on options directly. Apart from that the native UI breaks from the game's UI conventions and currently does not support many layouts and coloring of the game UIs. Perhaps instead of native UI, the android version should implement dragging and tapping controls similar to mouse controls in the game UIs.

> I will say one addition that could really help mobile, would be a dedicated keyboard with a button to pop it open and closed. a small button you could drag around the screen that would open a custom qwerty keyboard.

I think some software keyboards support floating keyboard which is close to what you want here.

ZhilkinSerg:
> I will say one addition that could really help mobile, would be a dedicated keyboard with a button to pop it open and closed. a small button you could drag around the screen that would open a custom qwerty keyboard.
> 
> But i think that would be difficult, if not impractical, to implement.

Yeah, there are keyboard apps that do that. From CDDA side we need to implement portrait mode layout - that would give space even for non-floating keyboards.

PR #65172, opened 2023-04-18
[CR] Graded Steel Frames

Drew4484:
#### Summary
Content "Graded Steel Frames"

#### Purpose of change
Vehicle frames should be an excellent supply of useful steel for a survivor. 

#### Describe the solution
Frames are now high carbon steel, meaning a survivor with access to a car frame can start making better equipment sooner. Heavy-duty frames are now fully tempered steel, which is not directly usable but can be recycled into useful lc_steel.

Adding more grades of steel to represent automotive steels.

#### Testing
WIP

#### Additional context
High carbon steel is not an exact 1:1 with automotive steels but like steel in general, automotive steel is a broad term with a variety of alloys used even on the same vehicle. Perfectly simulating this would require dozens of varieties and a truly ridiculous amount of labor.

Update, change in direction for this: All production vehicles use Factory Frame/Factory Heavy-Duty Frame. Those can be smashed by the player for Metal Scrap, which still has the progression path to graded steel. If they disassemble the Factory Frame instead, they get X Tempered Steel, Y High Carbon Steel, Z Aluminum, etc.


Drew4484:
> Assuming superseded by #71835?

Yep too many rebase conflicts for a simple change.

PR #39318, opened 2020-04-06
Generate z-levels for lakes

ralreegorganon:
#### Summary

SUMMARY: None

#### Purpose of change

This adds the minimal amount of content to start working on some underwater mechanics and content by having lakes generate multiple z-levels of empty water and a lake bed at the bottom. 

There's a lot of work between here and actual useful playable content, ranging from needing to fix player movement, revisiting the mechanics associated with the existing water terrains, dealing with all the wacky overmap generation that results from the many subsurface things assuming they're always working with solid rock (labs, ant tunnels, subways, etc), vehicles vanishing into thin air when they sink, better overmap generation (no reason we can't have various depths), better local mapgen (I spent so much time making the lake shores look nice, these hard edges look like garbage), and so much more.

I didn't want to add this until the start of a new release cycle so that we could have time to work on it, and I don't have the capacity to do all of the aforementioned work, but my hope is that with the locations existing in a crude form, the rest will be iterated on.

Also, there are no fluid mechanics here. You cannot drain the lake, you cannot divert it, you cannot flood locations. Those are neat ideas--I'm not going to work on them.

#### Describe the solution

* Add a new regional setting to the lakes section which defines the lake depth (all lakes are the same depth for now)
* Add two new overmap terrains representing the lake bed and the column of water that exists between the lake bed and surface
* Add two very basic mapgens for the aforementioned overmap terrains that just fill the location with the two existing terrains that we have for these purposes, `t_lake_bed` and `t_water_cube`
* Add code to the overmap generation to extend the lakes down using all the aforementioned elements

#### Testing

Generated new overmaps with lakes in them and verified the subsurface lake locations were created.

You can't actually do much of anything in these locations since the player mechanics don't even know what to do with a water column--you can't swim up or down.

#### Additional context

*a location with some lakes*
image

*sewers and ant tunnels adjacent to the lakes*
image

*sewers don't fill the entire overmap terrain and include some solid rock, so this works*
image

*the same can't always be said for ant tunnels*
image

*labs and subways adjacent to the lakes*
image

*lab needs an airlock...*
image

*subways will need some help*
image

PR #65174, opened 2023-04-18
Allow vertical movement for avatar in water

ralreegorganon:
#### Summary

None

#### Purpose of change

Allow the player to swim up and down in water.

#### Describe the solution

* Write new avatar movement tests for vertical movement in water.
* Add a new terrain flag that indicates the terrain is completely water such that the player could swim up out of the location.
* Update the avatar vertical movement code to handle both diving beneath the surface of water and moving up and down in the water column.

This set of changes could be so much bigger, but this is deliberately bite-sized.

When I added the z-levels to lakes in #39318, it was because the consensus at the time was that we had a bit of a chicken-and-egg problem where no one would work on underwater game code because there was no content, and no one would work on content because there was no game code. Since I created the lakes and knew mapgen well, it was easy to add z-levels to them to at least give the minimum content needed with a place to go so that someone could write some code... but it's been three years (almost to the day) and no one has tackled this.

There are endless things to still be addressed here: bugs this exposes, weird inconsistencies in how being "underwater" has worked up to this point, weird mapgen things, weird npc/monster things, missing items/vehicles/terrain/furniture, mechanics like buoyancy, and so on.

Much as I said in #39318, I don't really have the time to do all that. But I think this at least gets us closer.

#### Testing

Added the aforementioned tests, and also went diving in a lake.

### DIFF
--- a/data/json/furniture_and_terrain/terrain-liquids.json
+++ b/data/json/furniture_and_terrain/terrain-liquids.json
@@ -37,7 +37,7 @@
     "looks_like": "t_water_sh",
     "color": "blue",
     "move_cost": 8,
-    "flags": [ "TRANSPARENT", "LIQUID", "NO_SCENT", "SWIMMABLE", "DEEP_WATER", "FISHABLE" ],
+    "flags": [ "TRANSPARENT", "LIQUID", "NO_SCENT", "SWIMMABLE", "DEEP_WATER", "FISHABLE", "GOES_DOWN" ],
     "connect_groups": "WATER",
     "connects_to": "WATER",
     "examine_action": "water_source"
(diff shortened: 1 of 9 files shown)


MNG-cataclysm:
Possibly a fish mutant camp at the bottom of the lake or in a ship of some sort?

Maleclypse:
Ok so I was looking over this and don't see a doc entry for the `WATER_CUBE` flag in mapgen.   So I'm curious if I started adding underwater mapgen like a flooded cabin I'm thinking I'd create underwater (WATER_CUBE)  floor and window terrains, walls would be normal and then an underwater roof terrain unless it's only one floors worth of flooding in which case I'd have the roof just chilling in the middle of the lake letting you know it's there.   Or am I getting ahead of the implementation so far

PR #65177, opened 2023-04-19
Actually run tests with `make RUNTESTS=1`

sparr:
#### Summary
None

#### Purpose of change
`make RUNTESTS=1` builds the tests but doesn't run them

#### Describe the solution
Modify `tests/Makefile` so the default target also runs the tests in addition to building them

More complex updates to the two Makefiles could keep the default behavior of `make` in `tests/`, just compiling the tests, while allowing `make RUNTESTS=1` in the root dir to build and run the tests.

#### Testing
Unlike apparently every other time I've ever said this... Ran the tests!


actually-a-cat:
Does this leave a way to build tests but not run them automatically? Often I want to build tests but run them manually in a non-default way. (Sorry if it does and it's obvious, I struggle with reading makefiles)

sparr:
`make -C tests cata-tests` should do the trick

sparr:
I've updated the PR so there are now two separate make params, so that the existing CI workflows won't break.

TESTS=1 builds the tests (same as old RUNTESTS=1), 0 does not, and the default is 1.

RUNTESTS=1 runs the tests, 0 does not, and the default is 0.

Issue #49696, opened 2021-07-07
"Good Memory" and "Forgetful" traits are crippled by an optionality of Skill Rust


MomoiroShinobinomono:
# Is your feature request related to a problem?

I'm always frustrated when I cannot take "**Good memory**" trait as it is takes 3 whole points. 
My problem with it is that I never play with skill erosion and never did I see anyone doing it either.
To spend 3 points on a crippled trait half the functions of which I'm never going to access seems unreasonable, as Skill Rust is an **_optional_** addition to the gameplay that is even set at the Debug subsection of the Options menu.
Moreover, it looks especially unreasonable considering the fact the "**Good Memory**" and "**Forgetful**" traits are **_always_** there when you create a new character.

# Describe the solution

### Split the functions of the "**Good memory**" trait:
1. Take out a _slower erosion rate of skills_ part from the "**Good Memory**" trait and make it into its own trait which is only accesible in the character creation menu if the player has turned on Skill Rust in the Options.
 - It could be renamed as "**Mnemonist**" or left as is.
2. Merge the part about being able to remember more terrain from "Good Memory" trait with the "Scout" trait and up the price of "Scout" to **2** points. 

### Split the functions of the "**Forgetful**" trait:
1. Take out a _faster erosion rate of skills_ part from the "**Forgetful**" trait and make it into its own trait which is only accesible in the character creation menu if the player has turned on Skill Rust in the Options.
 - It could be renamed as "**Dementia**" or left as is.
2. Merge the part about being able to remember less terrain from "**Forgetful**" trait with the "**Topographagnosia**" trait and up the price of "**Topographagnosia**" to **-2** points.

**1**'s are going to ensure that the players who want to give their characters less or more memory map capacity will not be limited by an **_optional_** Skill Rust and also slightly tidy the Traits section of the character creation menu, while **2**'s will add more roleplay value to already existing traits.

Thanks.

actual-nh:
The map memory system may be reworked before too long - see #47253.

MomoiroShinobinomono:
I've got an additional idea about a memory map system, it would be really nice if players could designate 1 or 2 or couple more overmap squares as their "base" which will prevent a memory map of said squares from being forgotten.
Most complaints I saw in regards to the current system were exactly that characters easily forget layout of places they tend to spend weeks or even months at.
Which is understandably both jarring and unreasonable.
It could be called "**Mental note**".

anothersimulacrum:
[...]

PR #65178, opened 2023-04-19
Memory-Based Mutations Outside of Char Gen

BananaBonanza:
#### Summary
Content "Memory-Based Mutations Outside of Character Generation"

#### Purpose of change
With the recent patches and fixes to the memory system, I realized that the mutations for memory were a bit lacking. "Good Memory" wasn't available outside of character generation, "Forgetful" didn't have tags in the JSON to cancel "Good Memory", and there weren't any late-game mutations for further improving memory like CBMs have. This is tangentially related to #49696.

#### Describe the solution
JSON has been altered to cancel Forgetful/Good Memory. Good Memory has also become tied to the Cephalopod and Alpha mutation trees. In addition, two new mutations have been added to the Cephalopod tree.
**Distended Head:** A superficial mutation that doesn't add anything except beauty impact.
**Immutable Memory:** A upgrade to "Good Memory" that multiplies skill rust by a factor of 0.1. This has the prerequisite of "Good Memory" and "Distended Head".

The relationship between Good Memory and Forgetful seems mutually exclusive, so there isn't much of an alternative for them. 

Originally I wanted to include an additional mutations for faster skill rust, but I noticed that the "Fish" mutation doesn't have Forgetful as a mutation. Thus, it seemed odd to include a worse mutation for the Fish.

The modifier for the "Immutable Memory" I've gone back and forth on. Looking at issues #64916 and #65141, it seems that completely removing skill rust will be met with resistance. Thus, I have settled on a number that allows for a character concept I've played with in the past. Basically, a Cthulhu type that bunkers down for long stretches (technically I was an Ursine post-threshold with some Cephalopod mutations, but I was role-playing so leave me alone) only to return and spread chaos. The other alternative would be to remove skill rust when asleep, but that would then mean that about a third of in-game time wouldn't have skill rust. Hence why I have created this mutation as it is. However, I'm good with changing this multiplier if requested.

#### Testing
The JSON has been extensively tested on the most recent experimental build. The mutations have been added, removed, and forced removed by the forgetful trait.

#### Additional context
If additional mutations for a worse version of forgetful are reasonable, then I would create a new pull request for a mutation called "Dementia" with a skill rust modifier of 3 (or something like that). Possibly tied to the Chimera and/or Fish mutation trees.

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -859,8 +859,31 @@
     "types": [ "MEMORY" ],
     "skill_rust_multiplier": 0.66,
     "starting_trait": true,
-    "valid": false,
+    "changes_to": [ "IMMUTABLEMEMORY" ],
+    "category": [ "ALPHA", "CEPHALOPOD" ],
     "cancels": [ "FORGETFUL" ]
+  },
+  {
+    "type": "mutation",
+    "id": "DISTENDEDHEAD",
+    "name": { "str": "Distended Head" },
+    "points": 1,
+    "visibility": 3,
+    "ugliness": 3,
+    "description": "Your brain has grown to almost double the size and your head has distended to match. Despite its grotesque size, this hasn't resulted in any discernable differences to your mental capabilities.",
+    "leads_to": [ "IMMUTABLEMEMORY" ],
+    "category": [ "CEPHALOPOD" ]
+  },
+  {
+    "type": "mutation",
+    "id": "IMMUTABLEMEMORY",
+    "name": { "str": "Immutable Memory" },
+    "points": 4,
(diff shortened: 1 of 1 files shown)


fairyarmadillo:
I feel compelled to point out that cephalopods do not have brains in their head-feet, and in fact they have relatively small brains. Their intelligence comes partly from their thinking ability being distributed out among ganglia in their tentacles.

What I'm saying is the elongated head mutation is extraneous and not reflective of cephalopod anatomy.

BananaBonanza:
Looking into your suggestion, @fairyarmadillo, it looks like the Slime threshold already has a "Distributed Neurology" mutation that I don't just want to copy. I'm going to take this back to the drawing board to come up with a more comprehensive solution. Thanks!

PR #65183, opened 2023-04-19
Add no damage message for ranged weakpoint hits

Zephanos:
#### Summary
Interface "Adds a message that details no damage hits against weakpoints with ranged weapons."

#### Purpose of change
To provide additional context for players when a shot fails to do damage to a monster especially when using low-powered weapons/rounds(eg. slings or .22) or when hitting strongpoints. 

#### Describe the solution
Expanded the no damage line to check if a weakpoint is being hit and if so to display the appropriate message.

I had some trouble with phrasing to make it work with how weakpoints are named. I think what I got works but I'm down to change it if better or more desirable phrasing is conceived. 

#### Testing
Spawned in a variety of armored monsters then shot them with .22 at various skill levels then checked the log to make sure the messages displayed. Had several monsters and npcs shoot other armored monsters (mostly the Titan Stag Beetle) as well. Also shot some unarmored monsters to make sure I didn't break anything. 

#### Additional context
the shot

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -988,11 +988,17 @@ void Creature::messaging_projectile_attack( const Creature *source,
                 add_msg( source->is_avatar() ? _( "You miss!" ) : _( "The shot misses!" ) );
             }
         } else if( total_damage == 0 ) {
-            //~ 1$ - monster name, 2$ - character's bodypart or monster's skin/armor
-            add_msg( _( "The shot reflects off %1$s %2$s!" ), disp_name( true ),
-                     is_monster() ?
-                     skin_name() :
-                     body_part_name_accusative( hit_selection.bp_hit ) );
+            if( hit_selection.wp_hit.empty() ) {
+                //~ 1$ - monster name, 2$ - character's bodypart or monster's skin/armor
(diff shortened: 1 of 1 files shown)


src/creature.cpp
```diff
@@ -988,11 +988,17 @@ void Creature::messaging_projectile_attack( const Creature *source,
...
-                     is_monster() ?
-                     skin_name() :
-                     body_part_name_accusative( hit_selection.bp_hit ) );
+            if( hit_selection.wp_hit.empty() ) {
+                //~ 1$ - monster name, 2$ - character's bodypart or monster's skin/armor
+                add_msg( _( "The shot reflects off %1$s %2$s!" ), disp_name( true ),
+                         is_monster() ?
+                         skin_name() :
+                         body_part_name_accusative( hit_selection.bp_hit ) );
+            } else if( !hit_selection.wp_hit.empty() ) {
```
irwiss, line 997:
```suggestion
            } else {
```

DSeyka:
There's a double "the" in the example screenshot.

Zephanos:
> There's a double "the" in the example screenshot.

Whoops thanks for catching that

PR #65184, opened 2023-04-19
Add `longest_side` to bullets

GuardianDll:
#### Summary
None
#### Purpose of change
Doing some stuff, i found a need to calculate the length of the bullets, and found the game has lack of bullet length for almost all rounds. As i will use this info anyway, i decide to fix it also, as it may cause some issues, when charges went away
#### Describe the solution
Simply add `longest_side` to most bullet in the game - only 1 change per file, as bullets copy-from each other
The omitted bullet lengths are paper cartridges, some heavy stuff (as they have it already) and caseless rounds


data/json/items/containers/military.json
```diff
@@ -226,6 +226,7 @@
     "flags": [ "COLLAPSE_CONTENTS" ],
     "weight": "22 g",
     "volume": "480 ml",
+    "longest_side" "14 cm",
```
GuardianDll, line 229:
```suggestion
    "longest_side": "14 cm",
```

PR #65186, opened 2023-04-19
Quadrupedal movement with paws

Rewryte:
#### Summary
Features "Quadrupedal movement with paws"

#### Purpose of change
Paws are a pretty specialized piece of anatomy. Being adapted for weight bearing and movement comes at the price of the versatility of hands, represented in game by various penalties to crafting and combat, as well as restricting armor use. Yet using them of their intended purpose is currently not an option. Quadrupedal movement fills this gap. While it is mostly intended as a flavor feature for roleplaying a beast mutant, I can see a few playstyles getting some utility out of it.

#### Describe the solution
Quadrupedal movement uses a conditional enchantment which is activated when a character fulfills three conditions:

1. Has any paws mutation
2. Not wielding anything
3. Is crouching or prone

When triggered, it gives some flavor text as well as a move cost multiplier which negates the speed penalty of crouching. This allows a character to move at normal walking speed, yet maintain the lower profile of a crouch to represent quadrupedal movement. Prone receives a similar effect which leaves it at one-third walking speed. This is meant to emulate prowling.

Make a new movement mode, but that would make it available to all characters regardless of mutations, and cycling through the existing number of movement modes is already cumbersome enough so I did not want to clutter it further.

#### Testing
Tested the paws, broad paws, and little paws mutations on a fresh character by observing the move cost widget on the sidebar. No errors were encountered and movement cost when crouching and prone was calculated correctly.

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -6302,7 +6302,22 @@
       [
         {
           "condition": {
-            "and": [ { "or": [ { "u_has_move_mode": "crouch" }, { "u_has_move_mode": "prone" } ] }, { "not": "u_can_drop_weapon" }, { "u_has_any_trait": [ "THRESH_BEAST", "THRESH_FELINE", "THRESH_LUPINE", "THRESH_URSINE", "THRESH_RAT", "THRESH_CHIMERA", "THRESH_MOUSE", "THRESH_RABBIT" ] } ]
+            "and": [
+              { "or": [ { "u_has_move_mode": "crouch" }, { "u_has_move_mode": "prone" } ] },
+              { "not": "u_can_drop_weapon" },
+              {
+                "u_has_any_trait": [
+                  "THRESH_BEAST",
+                  "THRESH_FELINE",
+                  "THRESH_LUPINE",
+                  "THRESH_URSINE",
+                  "THRESH_RAT",
+                  "THRESH_CHIMERA",
+                  "THRESH_MOUSE",
+                  "THRESH_RABBIT"
+                ]
+              }
+            ]
           },
           "msg_on": { "text": "You drop down on all fours." },
           "msg_off": { "text": "You prance up and walk on your hind feet." }
@@ -6312,7 +6327,22 @@
     "enchantments": [
       {
         "condition": {
-          "and": [ { "or": [ { "u_has_move_mode": "crouch" }, { "u_has_move_mode": "prone" } ] }, { "not": "u_can_drop_weapon" }, { "u_has_any_trait": [ "THRESH_BEAST", "THRESH_FELINE", "THRESH_LUPINE", "THRESH_URSINE", "THRESH_RAT", "THRESH_CHIMERA", "THRESH_MOUSE", "THRESH_RABBIT" ] } ]
+          "and": [
+            { "or": [ { "u_has_move_mode": "crouch" }, { "u_has_move_mode": "prone" } ] },
+            { "not": "u_can_drop_weapon" },
+            {
+              "u_has_any_trait": [
+                "THRESH_BEAST",
+                "THRESH_FELINE",
+                "THRESH_LUPINE",
+                "THRESH_URSINE",
+                "THRESH_RAT",
+                "THRESH_CHIMERA",
+                "THRESH_MOUSE",
+                "THRESH_RABBIT"
+              ]
+            }
+          ]
         },
         "values": [ { "value": "MOVE_COST", "multiply": -0.5 } ]
       }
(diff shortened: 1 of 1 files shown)


drawsstuff:
This is fantastic! I absolutely adore the flavor and utility this adds to the game!

I _gotta_ ask, is there any plans to integrate a form of "fast" quadrupedal movement? An example would be dropping to all fours when breaking out into a sprint, should conditions be right. Flavor wise, doing so might imply a reversion to the limited bipedalism seen in many animals, where standing upright is reserved for fighting, fine manipulation, or surveying. Upright movement remains possible and is their primary way of interacting with the world around them, but need be, they can drop to all fours for the locomotion they need given the situation. 

Outside of the gameplay and roleplay aspects this could add, I'll just be frank as to why I'm asking: I think it's cool.

Rewryte:
@drawsstuff I did consider the idea of quadrupedal sprinting, though it is not something I am gunning for in the current iteration. Making new movement modes would be the most direct approach, but we don't have a way to tie them to specific mutations, and switching to them took more button presses than I think most people would bother with.

My current approach of using the enchantment system can't modify existing movement modes enough to, for example, give crouch LOS to sprinting. So this won't work either. Of course, once we start messing with the .cpp files, further possibilities might open up, but for now, this is the most elegant solution I can come up with.

drawsstuff:
> @drawsstuff I did consider the idea of quadrupedal sprinting, though it is not something I am gunning for in the current iteration. Making new movement modes would be the most direct approach, but we don't have a way to tie them to specific mutations, and switching to them took more button presses than I think most people would bother with.
> 
> My current approach of using the enchantment system can't modify existing movement modes enough to, for example, give crouch LOS to sprinting. So this won't work either. Of course, once we start messing with the .cpp files, further possibilities might open up, but for now, this is the most elegant solution I can come up with.

Makes sense! Thanks for replying! Looking forward to the day it becomes doable, but until then, this is a wonderful addition!

PatrikLundell:
For this to really work the mutations would have to modify limb lengths, as human arms are much shorter than the legs (at least in the general case: there are outliers, of course), resulting in a rather awkward four limbed stance. Also, the arms would probably need to be stronger to work as normal load bearing locomotion limbs.

Rewryte:
@PatrikLundell Tbh, having longer limbs was always how I envisioned the mutation. If the description of little paws mutation is anything to go by, it does state that "your hands have elongated".

fairyarmadillo:
Yeah I am pretty sure that somebody deep into Lupine isn't working with human proportions at all. You're not an anime wolf girl, you're a half-human abomination.

bombasticSlacks:
This is a cool addition; however, we don't have great ways to detect if a character has the anatomy to run on all fours. The best the dev team could think up that would be easy for you is making the speed boost also conditional on being post threshold in one of the relevant mutation branches. 

If you can add those conditionals, this will be mergable 😄

Rewryte:
@bombasticSlacks Sure... and done.  Side note, I did not actually add all the quadruped thresholds, only those with paws.  In hindsight, maybe I should not have tied quadrupedal movement to paws in the first place since lizards and frogs are quadrupeds but lack paws, but that is a discussion for another day.

fairyarmadillo:
> @bombasticSlacks Sure... and done. Side note, I did not actually add all the quadruped thresholds, only those with paws. In hindsight, maybe I should not have tied quadrupedal movement to paws in the first place since lizards and frogs are quadrupeds but lack paws, but that is a discussion for another day.

You may want to remove chimera from that list. Chimerae have hooves on their feet, but no hand paws. They actually have talons like a bird, which is obviously bipedal.

fairyarmadillo:
Sorry to double post. Per the reviews above, "prance" was being used incorrectly. To prance usually means to walk with a high-stepping gait, not to stand up.

There's one further issue and that's that I don't think the prone prowling thing is going to work. Characters with 0 leg HP are forced to be prone, and that's probably not the only effect that can or will knock a character down in this manner. It wouldn't make sense for a mutant to be prowling when they should be lying on the ground in agony. I also think it's redundant - crouching is already sneaking.

Rewryte:
@fairyarmadillo Forgive my rusty greek mythology, but shouldn't chimerae have paws on the front and hooves on the back?

I believe prancing is used to describe a quadrupedal animal shifting the weight to their hind feet and rearing up.  For instance, the Ferrari logo is described as a prancing horse.  I'm open to suggestions though.

About the prowling thing, I was just iterating though possibilities and stating the first reason on why it isn't a viable solution.  Like the first statement of an && operator you know?

fairyarmadillo:
The chimerae in the game aren't based on the one from greek mythology, it's just a word that means many creatures mixed together. A chimera mutant in CDDA winds up walking on cow hooves and clawing people with bird/raptor forearms. They do not have paws at all, nor do they have wings, a snake tail, multiple heads, fire breath, a lion head, or anything else from greek mythology.

I did submit ideas in the form of the review, which you can see slightly upstream in the comment thread. I swapped it out for  "msg_off": { "text": "You rise up to walk on your hind feet." } In American English, prancing is generally considered to mean a silly sort of walk, and that's how people are going to read it. Your call I guess, but it sounds strange to me.

I'm not sure what you mean about prowling. I'm saying that it's redundant and also causes a problem as you have implemented it. My suggestion would be to remove the prowling stuff entirely and just make the crouching changes, excluding chimera from the list of thresholds.

Rewryte:
@fairyarmadillo I guess you have a point about chimerae.  I don't mind changing the flavor text either.  I don't see your review anywhere though.  Neither in comments nor in the conversations menu of the files changed tab of this PR.  Can you see them on your end?

Thanks for clarifying about the prowling feature, didn't get what you meant at first.  Testing with my changes show that quadrupeds are not immune to becoming prone with broken legs, and still receive the substantially increased move cost and other combat penalties when prone.  They do however, reduce said move cost from 10x to 5x at zero pain (they still have two feet left after all).  If this still proves to be controversial, I can remove the prone changes.

MNG-cataclysm (approved):
Thank you for doing this.

fairyarmadillo:
> @fairyarmadillo I guess you have a point about chimerae. I don't mind changing the flavor text either. I don't see your review anywhere though. Neither in comments nor in the conversations menu of the files changed tab of this PR. Can you see them on your end?
> 
> Thanks for clarifying about the prowling feature, didn't get what you meant at first. Testing with my changes show that quadrupeds are not immune to becoming prone with broken legs, and still receive the substantially increased move cost and other combat penalties when prone. They do however, reduce said move cost from 10x to 5x at zero pain (they still have two feet left after all). If this still proves to be controversial, I can remove the prone changes.

I don't think that's going to work, because there's no check in place to see if they have working arms. So a quadruped can have two broken arms and one broken leg and still move faster than a human with one broken leg.

I can see the reviews on my end. I don't know why you can't. All they say is to change prance up to rise up, and to remove chimera.

Rewryte:
Merged fairyarmadillo's edits manually

PR #65190, opened 2023-04-19
fix: house32 glass door windows

FarkasKBence:
#### Summary

None

#### Purpose of change

kép

#### Describe the solution

Replaced glass doors with glass walls.

Other changes include:
* adding `standard_domestic_lino_bathroom` and `standard_domestic_landscaping_palette` to the palette list
* ordered file to the standard format
* adding bedside tables
* unifying gutter terrains
* removing unneeded weights

None.

#### Testing

Debug-spawned in the world.

#### Additional context

Non.


data/json/mapgen/house/house32.json
```diff
@@ -1,10 +1,9 @@
 [
   {
+    "type": "mapgen",
     "method": "json",
     "om_terrain": "house_32",
     "//": "used by lab basements.",
-    "type": "mapgen",
-    "weight": 100,
```
Maleclypse, line 7: Does it not need a weight?

LovamKicsiGazsii: As long as there's no other mapgen variant with the same id, it should be fine.

PR #65200, opened 2023-04-20
Refactor, consolidate, improve plant harvest funcs

sparr:
#### Summary
Features "Make plant harvesting more consistent"

#### Purpose of change
While deduplicating the code for harvesting plants I encountered many opportunities for small improvement and refactors and balance adjustment.

#### Describe the solution
* Move some repeated functionality into helpers
* Change harvest product placement from the ground to the player, or from the player to basecamp storage
* Add messages for more harvest action results
* Produce byproducts from all harvest methods
* Change harvest quantities to be more low and less at high skill, reducing the multiplier from 0 to 10 skill from 7.5 to ~3

#### Testing
I've been testing in-game and will continue as long as this remains a draft.
`tests/cata_test [rng]`

#### Additional context
See comments on code.
Replaced the roll_remainder implementation with one I wrote before I found it.

### DIFF
--- a/src/iexamine.cpp
+++ b/src/iexamine.cpp
@@ -2634,23 +2634,39 @@ void iexamine::harvest_plant( Character &you, const tripoint &examp, bool from_a
         }
     } else { // Generic seed, use the seed item data
         const itype &type = *seed->type;
-        here.i_clear( examp );
 
-        int skillLevel = round( you.get_skill_level( skill_survival ) );
         ///\EFFECT_SURVIVAL increases number of plants harvested from a seed
-        int plant_count = rng( skillLevel / 2, skillLevel );
-        plant_count *= here.furn( examp )->plant->harvest_multiplier;
-        plant_count = std::min( std::max( plant_count, 1 ), 12 );
-        const int seedCount = std::max( 1, rng( plant_count / 4, plant_count / 2 ) );
-        for( item &i : get_harvest_items( type, plant_count, seedCount, true ) ) {
-            if( from_activity ) {
-                i.set_var( "activity_var", you.name );
+        const int skill_level = round( you.get_skill_level( skill_survival ) );
+        std::list<item> results = harvest_items( type, here, examp, skill_level );
+        if( results.empty() ) {
+            item tmp = item( type.get_id(), calendar::turn );
+            add_msg( m_info, _( "You fail to harvest anything." ) );
+        } else {
+            for( item &i : results ) {
+                if( from_activity ) {
+                    i.set_var( "activity_var", you.name );
+                }
+                add_msg( m_info, _( "You harvest %1$s (%2$d)." ), i.type_name(), i.count() );
+                you.i_add_or_drop( i, 1 );
             }
-            here.add_item_or_charges( you.pos(), i );
         }
-        here.furn_set( examp, furn_str_id( here.furn( examp )->plant->transform ) );
+        practice_survival_while_gathering_from_plants( you, 3 );
         you.moves -= to_moves<int>( 10_seconds );
     }
+}
+
+std::list<item> iexamine::harvest_items( const itype &type, map &here, const tripoint &pos,
+        const int skill_level )
+{
+    std::list<item> items;
+    here.i_clear( pos );
+    const int plant_count = roll_remainder( std::max( rng_float( 4, skill_level ), 6.0 )
+                                            * here.furn( pos )->plant->harvest_multiplier );
+    const int seed_count = roll_remainder( std::max( 0.0,
+                                           rng_float( plant_count / 4.0, plant_count / 2.0 ) ) );
+    items.splice( items.begin(), get_harvest_items( type, plant_count, seed_count, true ) );
+    here.furn_set( pos, furn_str_id( here.furn( pos )->plant->transform ) );
+    return items;
 }
 
 ret_val<void> iexamine::can_fertilize( Character &you, const tripoint &tile,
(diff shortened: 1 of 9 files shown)


NetSysFire:
This might be out of scope, but there is also an issue about multiple terrain harvests per year (e.g young leaves from say hickory trees in the spring and actual nuts in the autumn) which you might want to look at. I got it while doing the wilderness overhaul but never got around to report that.

sparr:
@NetSysFire I'm open to hearing more about it, although it probably is out of scope since this PR has already grown so much from my original plan.

src/rng.h
```diff
@@ -196,4 +196,13 @@ std::optional<tripoint> random_point( const tripoint_range<tripoint> &range,
...
                                       const std::function<bool( const tripoint & )> &predicate );
 
+/**
+ * Round a value with the fractional part as the probability of rounding away from zero
+ * e.g. 4.2 rounds to 4 20% of the time or to 5 80% of the time
+ *
+ * @param val Value to be rounded.
+ * @return Rounded value.
+ */
+double round_rand( double val );
```
irwiss, line 206: may be just me, but this name seems not really descriptive of what it actually does - round away from 0 is likely not what most would expect from innocent named "round" function, it's also dissimilar (in name style) to almost identical roll_remainder

src/rng.cpp
```diff
@@ -178,3 +178,11 @@ std::string random_string( size_t length )
...
     return str;
 }
+
+double round_rand( double val )
+{
+    double mag = std::abs( val );
+    double frac = std::modf( mag, &mag );
+    mag += x_in_y( frac, 1.0 );
+    return copysign( mag, val );
+}
```
anothersimulacrum, line 188:
```suggestion
}

```

src/rng.cpp
```diff
@@ -178,3 +178,11 @@ std::string random_string( size_t length )
     std::generate_n( str.begin(), length, randchar );
     return str;
 }
+
+double round_rand( double val )
```
anothersimulacrum, line 182: Could we please not have (more) RNG based rounding?

src/rng.h
sparr: I spent at least five minutes searching through the existing rnd functions and failed to find roll_remainder, which I agree is almost identical. I'll switch to using that one.

src/mission_companion.cpp
```diff
@@ -1668,47 +1667,52 @@ void talk_function::field_harvest( npc &p, const std::string &place )
...
+        if( ( *i ).typeId() == plant_types[plant_index] ) {
+            number_plants += i->count_by_charges() ? i->charges : 1;
+            i = results.erase( i );
+            i--;
+        } else if( ( *i ).typeId() == seed_types[plant_index] ) {
+            number_seeds += i->count_by_charges() ? i->charges : 1;
+            i = results.erase( i );
+            i--;
+        }
+    }
```
sparr, line 1690: I'd eventually like to count each item, not just the primary plant and seed, and then give more options when [forced] selling things.

src/iexamine.cpp
```diff
@@ -3766,20 +3787,28 @@ static void pick_plant( Character &you, const tripoint &examp,
...
 
-    const float survival = you.get_skill_level( skill_survival );
-    you.practice( skill_survival, 6 );
+    const int survival = you.get_skill_level( skill_survival );
+    ///\EFFECT_INT Intelligence caps survival skill gains from foraging
+    const int max_forage_skill = you.int_cur / 5 + 1;
+    ///\EFFECT_SURVIVAL decreases survival skill gain from foraging (NEGATIVE)
+    const int max_exp = ( max_forage_skill - survival );
+    // Award experience for foraging attempt regardless of success
+    you.practice( skill_survival, rng( 1, max_exp ), max_forage_skill );
```
sparr, line 3796: This ended up duplicated with minor changes in enough places that I will probably move it out to a helper before un-drafting this PR.

src/iexamine.cpp
```diff
@@ -2634,25 +2634,46 @@ void iexamine::harvest_plant( Character &you, const tripoint &examp, bool from_a
...
         you.moves -= to_moves<int>( 10_seconds );
     }
 }
 
+std::list<item> iexamine::harvest_items( const itype &type, map &here, const tripoint &pos,
+        const int skill_level )
+{
+    std::list<item> items;
+    here.i_clear( pos );
+    const int plant_count = roll_remainder( std::max( rng_float( 4, skill_level ), 6.0 )
```
sparr, line 2668: This is a balance shift. 
```
// skill  0   1   2   3   4   5   6   7   8   9  10
// old   1.0 1.0 1.5 2.0 3.0 3.5 4.5 5.0 6.0 6.5 7.5
// new   2.0 2.5 3.0 3.5 4.0 4.5 5.0 5.3 5.5 5.6 5.7
```
Survival 10 producing 7.5x as much food as survival 1 doesn't make sense to me. Ditto every level of survival providing a linear increase in production.
This new formula increases the minimum production, decreases the maximum, and makes survival 7 produce 2.5x as much as survival 0, then a bit of diminishing returns up to survival 10.

src/mission_companion.cpp
```diff
@@ -1717,7 +1721,13 @@ void talk_function::field_harvest( npc &p, const std::string &place )
             tmp.charges = 1;
         }
         player_character.i_add_or_drop( tmp, number_seeds );
-        add_msg( _( "You receive %d %s…" ), number_seeds, tmp.type_name( 3 ) );
+        add_msg( _( "You receive %d %s…" ), number_seeds, tmp.type_name( number_seeds ) );
```
sparr, line 1724: I tried adding this same phrasing to `iexamine::harvest_plant` but realized it sounds bad with charged items (12 pumpkin is a single pumpkin) and items where a single item/charge is plural (e.g. seeds) and also sometimes the "handful of item" item names.
A few of us brainstorming on Discord couldn't come up with a good solution that covers all the bases. I don't like leaving these two cases different, so I'm going to think further on it and standardize them on... something.

Maleclypse:
Hi sparr,  I was working on trying to add ergot fungal infections as a thing the player could do to seedling forms of wheat/barley/other plants that ergot can infest and I discovered that you can't really touch plants when they are at the seedling stage and change them into something else.  My fallback option is to use a recipe to infest the grain seeds and then have that grow infested plants but someone mentioned that you were working on this space and I should ask you if there were a way to tie in to impacted at the seedling stage, I guess similar to how fertilizer maybe works (not sure we even still have fertilizer).  Thanks!

PR #65201, opened 2023-04-20
Rust UI refinements

I-am-Erk:
#### Summary
None

#### Purpose of change

The skill rust UI is getting more scrutiny now that more people are being dragged, kicking and screaming, into the century of the fruitbat. There are some elements in there that I just don't like and have wanted to fix for a while. Let's get to it.

#### Describe the solution

1. Change colour of skills to green if knowledge level > skill level. This is to indicate the selected skill is eligible for an xp boost because you _know_ more than you _have tried_ yet.
2. Change default listed level to your knowledge level, not your practical. The things you can do are defined by your knowledge. It's not helpful to show just the practical level in most cases. When you highlight the skill, you can see your skill level as you would have seen knowledge before. **NOTE**: I'd be open to switching this up for combat skills, but UI fixes are not my forte.
3. Remove skill rust warning messages about bionic memory banks. If you're an avid user of this CBM feel free to let me know, but it's my opinion this is now just another unnecessary message spam about rust.

I would like to add some information about how much of an XP boost you're getting from your knowledge::skill gap, but right now it is beyond my meager C++ skills on my very sleepy brain.

#### Testing

image

All works, but I wouldn't mind having one more kick at the can to display your XP boost before I finish this off.

#### Additional context

Some things I had planned to adjust on the existing rust system when I first made it, but haven't had the skill nor time for myself:

- If you have used a skill in the last 24(?) hours, it should not accumulate rust. This should apply even if you didn't actually use it enough to gain any XP. This change was beyond my code skills at the time, but I do think it is important, especially for skills that can be challenging to train at high levels.
- Related to above, if a skill is "turned off" for XP gain purposes we should still track its use and not rust it if it *would have* gained XP.
- The timer above could be set on a global variable, I wouldn't mind waiting more than 24 hours before the rust timer starts.
- Once those are implemented we should consider bumping up the rust ticker to every 12 hours, so it lags a little more before starting and once it does start it runs more often.

I think these are some very important changes and hope someone can help add them soon.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -3261,8 +3261,6 @@ void Character::do_skill_rust()
 
         const int rust_resist = enchantment_cache->modify_value( enchant_vals::mod::SKILL_RUST_RESIST, 0 );
         if( skill_level_obj.rust( rust_resist, mutation_value( "skill_rust_multiplier" ) ) ) {
-            add_msg_if_player( m_warning,
-                               _( "Your knowledge of %s begins to fade, but your memory banks retain it!" ), aSkill.name() );
             mod_power_level( -bio_memory->power_trigger );
         }
     }
--- a/src/player_display.cpp
+++ b/src/player_display.cpp
@@ -819,7 +819,8 @@ static void draw_skills_tab( ui_adaptor &ui, const catacurses::window &w_skills,
             const SkillLevel &level = you.get_skill_level_object( aSkill->ident() );
             const bool can_train = level.can_train();
             const bool training = level.isTraining();
-            const bool rusty = level.isRusty();
+            const bool skill_gap = level.knowledgeLevel() > level.level();
+            const bool skill_small_gap = level.knowledgeExperience() > level.exercise();
             int exercise = level.knowledgeExperience();
             int level_num = level.knowledgeLevel();
             bool locked = false;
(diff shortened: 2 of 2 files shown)


nexusmrsep:
How about presenting both theoretical and practical levels in numbers side by side, just like dodge skill shows applied level. This would require a way to show dodge applied love in other way unless it would be an exception. 
In my opinion showing both numbers would clear

GuardianDll:
Also what about bumping the trainings to give not, for example, lvl 3, but lvl 3 50%?

I-am-Erk:
> How about presenting both theoretical and practical levels in numbers side by side, just like dodge skill shows applied level. This would require a way to show dodge applied love in other way unless it would be an exception. In my opinion showing both numbers would clear a lot of misconception about what is what.

The only trouble is space. The melee skills have such long names that it starts to overlap even on the widest display.

nexusmrsep:
> The only trouble is space. The melee skills have such long names that it starts to overlap even on the widest display.

The entire menu is ripe for a redesign for this reason alone, but perhaps it could fit it in the meantime?

I-am-Erk:
> The entire menu is ripe for a redesign for this reason alone, but perhaps it could fit it in the meantime?

I'll definitely see what I can do.

Vaskritaya:
I don't know if green really reads as "you can get a skill boost" to me, I typically associate green with an existing bonus or enhancement, like if the skill was raised above its normal amount for some reason. Maybe cyan, since that's eye-grabbing and similar to the blue used for skills normally, but isn't hostile like the current pink?

Entirely personal opinion, I can't speak for anyone else's mental associations

I-am-Erk:
In the present version I'm looking at switching to cyan.

nexusmrsep:
> Also what about bumping the trainings to give not, for example, lvl 3, but lvl 3 50%?

Or at least +10% above the target level, thus less frequent "midnight disappointment" of level loss.

PR #65203, opened 2023-04-20
Add new types of artifacts

crypticcollaborator:
#### Summary
Content "Add new types of artifacts"

#### Purpose of change
Since the artifact refactor, the types of artifacts that can generate have greatly reduced in variety. In previous versions, artifacts could take the form of weapons and armor. These wearable and wieldable artifacts could be very engaging for roleplay and led to interesting trade-offs. Artifact weapons were effective but were at risk of being damaged or lost in combat and prevented wielding a gun or other ranged weapon. Artifact clothing only applied their powers when worn, leading to more control over when the effects applied. However, using artifact clothing forced users to manage their encumbrance and weigh how much they want to use the artifact's abilities versus the risk of the clothing being damaged in combat.

#### Describe the solution
This PR adds back the forms of artifacts that were available before the artifact refactor, including swords, spears, coats, helms, gloves, boots, and harps. They are largely identical to normal versions of the same object except for their artifact properties. Unlike previous versions in which descriptions and names were procedurally generated, these artifacts are a static list. Since these new artifacts are (probably) human made, artifact clothing and weapons only spawn in strange temples and mines. The nether and portals spawn malleable blob, teardrop, and cube artifacts as they do before this change.

This PR also adds a variety of wearable altered items, including a pair of sunglasses, an apron, and a hawaiian shirt. These spawn in microlabs in the same way as current altered items.

Wearable artifacts only apply their powers when worn. 

In the future, I'd like artifacts shaped like weapons to only apply their powers when wielded. I would also have liked to expand the types of powers that artifacts can have. In particular I'd like to return clairvoyance and invisibility, though in a more balanced form than the previous artifacts. I think both of these can be accomplished with JSON, but it would require more research and playtesting to understand how to get it to work.

I would also have liked to expand the number of artifacts to allow more variety beyond what the original artifact system had, but this is a good start.

#### Testing
I tested the artifact groups and found that my new artifacts spawned as expected with the normal variety of powers that other artifacts were.

I used map debug tools to test that artifacts were generated in the places that we expect them to be.

I tested that artifact clothing only applied their powers while worn. 

#### Additional context

An ancient bronze helmet artifact generated after dealing (messily) with amigara horrors.
image

A pair of altered sunglasses in the center of a microlab.
image


Daved27hundred:
shouldn't the sword and other weapons be a `WEAPON` instead of a `TOOL`?

crypticcollaborator:
I thought that as well, but it seems like there isn't a WEAPON type. Instead, weapons have `type: "TOOL"` and `category: "weapons"`. All the weapons in this PR inherit the category from the weapons they copy their attributes from.

anoobindisguise:
so just a heads up that altered_object can still only spawn with +- speed/attackspeed passives, is this your intention? I personally see no reason why these ones specifically have heavily limited passive effects. Might be out of scope for this PR though.

crypticcollaborator:
It's not my intention, but I'd consider it out of scope for this PR. I'll
ask Candlebury about it.

data/mods/TEST_DATA/known_bad_density.json
```diff
@@ -622,6 +622,7 @@
       "foodplace_food",
       "brew_dandelion_wine",
       "badge_deputy",
+      "art_badge",
```
mqrause, line 625: You're not supposed to add new things to this list. If an item fails density checks, you need to fix it by adjusting its weight, volume or possibly material.

crypticcollaborator: I fixed the offending badge_abstract base item and all its children, and removed them from known bad densities.

PR #65204, opened 2023-04-20
Confirmation before attacking neutral mobs

### LINKED ISSUE #55442: Make it harder to accidentally attack an exodii

It is too easy to accidentally attack an exodii. Unlike non-hostile NPCs, bumping into an exodii automatically attacks her or him. While exploring their main base I have accidentally attacked them one too many times and got murdered by a quadruped.

The base if packed full of them and they move - they might move into your path.

### Solution you would like.

The game should prompt me when I move into a non-hostile exodii similarly to how it prompts me when I walk into an NPC.

Walking with point-and-click is safe, but keyboard is not.

### Additional context

_No response_

### LINKED ISSUE #59021: Add some way to be able to get a prompt if the player would attack a non-hostile creature from a non-hostile faction

While I was walking around in the Exodii castle, I noticed that it was quite tedious and stressful to avoid walking into (and thus attaacking) the workers and quadrupeds

A single misshap could result in being blasted by a quadruped

### Solution you would like.

A way to get prompted if the player would accidentally attack a non-hostile to prevent being hurt by a now-hostile creature

Allowing to swap place with neutral creatures by walking in them and attacking by examining

### Additional context

_No response_

Rewryte:
#### Summary
Features "Confirmation before attacking neutral mobs"

#### Purpose of change
Since movement keys are also used for melee combat by walking into an occupied grid, holding down a movement key to move can result in a character melee attacking unintended targets and possibly drawing the aggression of otherwise neutral creatures. This is especially common in locations occupied by the Exodii faction where a large amount of neutral entities can be found.
Fixes #55442 
Fixes #59021

#### Describe the solution
Added a confirmation prompt which appears when attempting to melee attack a neutral creature by walking into them. This is similar to how attacking neutral NPCs are currently handled. Ranged attacks should be unaffected.

Disable melee attacking when holding down the key and requiring a separate button press.

#### Testing
Confirmation prompt appears correctly when walking into a grids occupied by neutral creatures. Tested with a chicken and an Exodii Quadruped. Walking into hostile creatures, friendly pets and NPCs are unchanged.

### DIFF
--- a/src/avatar_action.cpp
+++ b/src/avatar_action.cpp
@@ -360,9 +360,10 @@ bool avatar_action::move( avatar &you, map &m, const tripoint &d )
                 return false;
             }
 
+            // Ask for confirmation before attacking a neutral creature
             if( critter.attitude_to( you ) == Creature::Attitude::NEUTRAL &&
-            !query_yn( _( "You may be attacked!  Proceed?" ) ) ) {
-                	return false;
+                !query_yn( _( "You may be attacked!  Proceed?" ) ) ) {
+                return false;
             }
             you.melee_attack( critter, true );
             if( critter.is_hallucination() ) {
--- a/src/avatar_action.cpp
+++ b/src/avatar_action.cpp
@@ -361,13 +361,14 @@ bool avatar_action::move( avatar &you, map &m, const tripoint &d )
             }
 
             if( critter.attitude_to( you ) == Creature::Attitude::NEUTRAL &&
-                query_yn( _( "You may be attacked!  Proceed?" ) ) ) {
-                you.melee_attack( critter, true );
-                if( critter.is_hallucination() ) {
-                    critter.die( &you );
-                }
-                g->draw_hit_mon( dest_loc, critter, critter.is_dead() );
+            !query_yn( _( "You may be attacked!  Proceed?" ) ) ) {
+                	return false;
             }
+            you.melee_attack( critter, true );
+            if( critter.is_hallucination() ) {
+                critter.die( &you );
+            }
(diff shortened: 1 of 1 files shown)


irwiss:
https://github.com/CleverRaven/Cataclysm-DDA/pull/65075
If you're in safe mode = ON you shouldn't bonk neutrals already

Rewryte:
@irwiss Heh, never used safe mode so I guess that's on me. I'll let you maintainers decide if you want the extra security measure, otherwise feel free to close this.

irwiss:
You'll need to merge it with the check a couple lines above then, cause people who do use safe mode will now have to turn it off **and** do the y/n dance instead of just pressing ! they do now

Rewryte:
Hmm, while I can check if a player has safe mode off, there's no way to tell if they had just turned it off to attack a neutral, or never used safe mode in the first place. I can check for auto-safe mode though, but there's no guarantee that all safe mode users also use auto-safe mode. Thoughts?

irwiss:
you could probably try checking get_option<bool>( "SAFEMODE" )

Rewryte:
Thanks irwiss.  I'm working on something else right now, but I'll check it out when I'm done.

Rewryte:
So, apparently the safe mode option and safe mode hotkey are two different toggles. *mindblown*.  The hotkey still activates safe mode with the option off though.  Could be a bug, but then again I don't really use safe mode so what do I know.  Anyways, as irwiss suggested, I merged the two solutions so that safe mode enabled in options will use safe mode based prevention for attacking neutrals, and when disabled will use the confirmation prompts from this PR.

PR #65209, opened 2023-04-20
No more domestic tigers

NetSysFire:
#### Summary
Balance "No more domestic tigers"

#### Purpose of change

```
<m‌aleclypse> Ok so private/personal big cats are completely illegal in all of New England's states.  Indiana is the closest state you can even get a license for a tiger as a pet.  So it would make the most sense to remove tigers from every mongroup that isn't a zoo or circus mongroup.
```

Realism and balance. It does not make sense to encounter a tiger in a random barn in the middle of nowhere.

#### Describe the solution

Removing it from those monster groups.

#### Additional context

Introduced by @LyleSY in #41659.


Maleclypse:
I might add some nested drug dealers spawns with tigers because that is the source of most illicit tigers in America but that’s a task for later. Thanks for doing this!

LyleSY (approved):

PR #65213, opened 2023-04-21
[Xedra Evolved] SECRET flag dreamsmith recipes

Maleclypse:
#### Summary
None

#### Purpose of change
Hide dreamsmith recipes from people starting with deduction

#### Describe the solution
Adds Secret flag

none


Maleclypse:
@GuardianDll please confirm this won't interfere with your plans?

GuardianDll:
No problem for me

Maleclypse:
> No problem for me Why they are autolearned in a first place? The default value for autolearn is `true`, or i miss some functionality of a `SECRET` flag?

Autolearn defaults to false, so none of these recipes would be autolearned as you gain deduction skill in game.  However, characters who start the game with Deduction are assumed to know all non SECRET deduction recipes up to their starting skill level.  Same for all the skills.

PR #65219, opened 2023-04-21
[CR]Remove unnecessary prompt "soon!"

kwl01skz:
#### Summary
Interface "Show the expiration date of food, replacing the unreasonable "soon!"

#### Purpose of change
Show the expiration date of food, replacing the unreasonable "soon!

#### Describe the solution
When we check how long it is until food rots, we usually want to know when -- even if it's a ballpark figure -- not "soon!" Especially the "soon! It probably means 10 days.

`is_going_bad()` It will check to see if more than 90 percent of the food is rotten
The problem is, 100 x 0.1 = 10.
A food with a shelf life approaching 100 days tells us "soon!"  In fact, it may have a week's shelf life.

Instead of is_going_bad(), check the state of the food with another function, such as "4 hours from rot"

#### Testing
after
tt1

#### Additional context
before
tt2

### DIFF
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -833,11 +833,6 @@ class comestible_inventory_preset : public inventory_selector_preset
         }
 
         std::string get_time_left_rounded( const item_location &loc ) const {
-            const item &it = *loc;
-            if( it.is_going_bad() ) {
-                return _( "soon!" );
-            }
-
             time_duration time_left = get_time_left( loc );
             return to_string_approx( time_left );
         }


Zireael07:
I don't think an average person can tell whether it's going to expire in 20 or 50 minutes?

fairyarmadillo:
> Maybe we should add "maybe rotten" between "old" and "rotten."
> 
> * "Old" : Sure it's not rotting, but it won't last long.
> * "Maybe rotten" : Maybe a little bit rotten, maybe no rotten, I can't tell, eating it is risky (and warning), but not necessarily harmful.
> * "Rotten" : A glance to know that it is rotten, eating it must be harmful.

Yeah, something like this would be more appropriate. The purpose of Soon! is to cover up the fact that food goes from totally fine to irretrievably rotten in an instant.

A solution that best reflected reality I think would be a definite "best by" date, followed by a short window (a couple hours to a day?) wherein a food item is labeled as risky, but isn't visibly covered in mold or maggots, and then it's simply random chance that it makes you sick, maybe affected by lifestyle score and mutations. You could have a visible to-the-minute timer before food became risky, but risky's countdown should probably be obscured.

anoobindisguise:
Doesn't the player get more specific information than Soon! (tm) if they have high cooking/survival?

kwl01skz:
There's still a small problem here that hasn't been solved: when a soon is between 2 hours and 4 hours, we know it's maybe 3 hours. Its actual value is involved in sorting, in the food list.

PR #65226, opened 2023-04-21
Set Fixed ordering for skills in info menu, regardless of language chosen

### LINKED ISSUE #65218: Fixed order of skills in the player skill menu

Currently the skills in the Player skills panel (visible with @) are alphabetically sorted.
This means that the order vary wildly according on the language you're playing the game in.
No logical order is maintained, for example, in the italian translation "launchers" comes before all, "shotguns" comes before "rifles", "handguns" comes near the end and "throwing" is jammed inbetween firearms (between "rifles" and "submachineguns")

This is confusing at best, and extremely annoying to some.

### Solution you would like.

Remove the alphabetical ordering of skills.  Pick a sound order and have them appear in this fixed order.

_No response_

### Additional context

_No response_

MarcusAseth:
Adds an ordering value to the skills in skills.json, and uses it to control the order in which the skills appear in the player info menu

#### Summary
Interface "Adds an "ordering" value to each one of the skills inside the skills.json file. It then uses it to set a fixed order to how the skills should be desplayed in the Player Info panel"

#### Purpose of change
Fixes #65218

#### Describe the solution
The ordering value added to each skill in the json file controls the order in which the skills will be displayed, substituting the previous "alphabetical ordering" which led to random results depending on language selected.

#### Testing
I've launched the game and all looked good.

#### Additional context
An image of Before and After:
Cattura

### DIFF
--- a/src/skill.cpp
+++ b/src/skill.cpp
@@ -140,11 +140,11 @@ void Skill::load_skill( const JsonObject &jsobj )
     }
     skill_displayType_id display_type = skill_displayType_id( jsobj.get_string( "display_category" ) );
     Skill sk( ident, name, desc, jsobj.get_tags( "tags" ), display_type );
-    if( jsobj.has_int( "ordering" ) ) {
-        sk._ordering = jsobj.get_int( "ordering" );
+    if( jsobj.has_int( "sort_rank" ) ) {
+        sk._sort_rank = jsobj.get_int( "sort_rank" );
     } else {
-        sk._ordering = 1000000;
-        debugmsg( "skill '%s' missing 'ordering' field.", ident.str() );
+        sk._sort_rank = 1000000;
+        debugmsg( "skill '%s' missing 'sort_rank' field.", ident.str() );
     }
 
     sk._time_to_attack = time_to_attack;
--- a/src/skill.cpp
+++ b/src/skill.cpp
@@ -78,14 +78,14 @@ bool string_id<Skill>::is_valid() const
 
 Skill::Skill() : Skill( skill_id::NULL_ID(), to_translation( "nothing" ),
                             to_translation( "The zen-most skill there is." ),
-                            std::set<std::string> {}, skill_displayType_id::NULL_ID() )
+                            std::set<std::string> {}, skill_displayType_id::NULL_ID(), 100000000 )
 {
 }
 
 Skill::Skill( const skill_id &ident, const translation &name, const translation &description,
-              const std::set<std::string> &tags, skill_displayType_id display_type )
+              const std::set<std::string> &tags, skill_displayType_id display_type, int ordering )
     : _ident( ident ), _name( name ), _description( description ), _tags( tags ),
-      _display_type( display_type )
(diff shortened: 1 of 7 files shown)


NetSysFire:
For the future, you can batch apply those suggestions in the files tab, so you dont need 15 commits to apply them all.

src/skill.cpp
```diff
@@ -139,7 +139,8 @@ void Skill::load_skill( const JsonObject &jsobj )
         jso_tta.read( "time_reduction_per_level", time_to_attack.time_reduction_per_level );
     }
     skill_displayType_id display_type = skill_displayType_id( jsobj.get_string( "display_category" ) );
-    Skill sk( ident, name, desc, jsobj.get_tags( "tags" ), display_type );
+    int ordering = jsobj.get_int( "ordering", 100000000 );
+    Skill sk( ident, name, desc, jsobj.get_tags( "tags" ), display_type, ordering );
```
irwiss, line 143: You don't need any of the constructor changes, especially if you just feed it garbage value as a default anyway.

If you remove constructor changes you can just assign it like below.

Also a skippable alert so out-of-tree mod authors are aware they need to update their json would be useful because the garbage value will make multiple skills with the garbage value have undefined order, something similar to:
```suggestion
    Skill sk( ident, name, desc, jsobj.get_tags( "tags" ), display_type );
    if( jsobj.has_int( "ordering" ) ) {
        sk.ordering = jsobj.get_int( "ordering" );
    } else {
        sk.ordering = 1000000;
        debugmsg( "skill '%s' missing 'ordering' field.", ident.str() );
    }
```

The PR also misses updating in-tree mod skills - aftershock has smart guns and magiclysm has spellcraft skills, may be more

Also minor nitpick; "ordering" doesn't fit mostly-noun naming scheme of json fields

MarcusAseth: I've applied all changes except the renaming of the "ordering" field because I can't think of something better.  I'll change it to whatever you guys will tell me :)

MarcusAseth: following andrei8l suggested term, I've changed "ordering" into "sort_rank"

BrettDong:
Why not just sort based on lexical order of its English name? How is a manually added ordering field maintainable?

MarcusAseth:
I don't think alphabetical sorting is a good idea because the skill ends up arranged in an order that is not logical from a gameplay point of view.
Below you can see an image of "Before/After"
capture

you see we have "bashing weapon" then "cutting weapon" and "piercing weapon is far away on it's own, despite being logically close.
"melee" which govern all those skills in the section thus distinguished in a sense is neither on top or below, but in the middle - same for marksmanship.
With manually ordering, we can order guns in order of threat the same or similar order you would see them appear in any other game, no other game I know of presents you "launchers" before "handguns"

BrettDong:
If I am not mistaken, the sort ranking you added is global, not local to each skill category?

MarcusAseth:
> 

yes, but the numbers from skill to skill are 1000 apart, so there is plenty of space to add new stuff wherever people see (logically) fit

PR #65229, opened 2023-04-21
Expand mattacks to use the dialog condition system

Venera3:
#### Summary
Features "Allow JSON monster special attacks to use the dialog condition system"

#### Purpose of change

Rework the forbidden/required effect system into actuall using the dialog conditionals (the thing they were kinda sorta aping anyway).

#### Describe the solution

Added dialog conditionals to leap (self-only), spell (self if no target required, self and target otherwise), and melee(self and target) mattacks.
Removed the bulky effect checks.
Rewrote the unit test to utilize the dialog conditons logic/arguments.

This is necessary for some #64999 grabby stuff but it can stand on its own and I didn't want to make reviewers cry.
Bite the bullet and move the monster flags over to normal flag ids, but that felt like a huge refactor and there were concerns about ambiguous flags last time I raised it.

#### Testing

Tested charged and caustic amalgamations ( melee and spell conditions fired as expected)
Added a new unit test and retraced the old test to ensure parity.

#### Additional context

`u_has_flag` is only capable of checking for effect flags on the monsters, but that's all I need right now.

This is important to truly unhardcode grabbing - effect flags can't be checked on mosters otherwise.


Ramza13:
Nice, you don't need to but if you wanted to you could remove the range field and do that via condition as well.

Venera3:
> Nice, you don't need to but if you wanted to you could remove the range field and do that via condition as well.

Does that to obstruction checking?

PR #65232, opened 2023-04-21
[WIP]removed debug mode when NPC start infinite looping

andrewchen1216:
#### Summary
Category "Brief description"


irwiss:
imo leave the debugmsg up so it lands in the logs for triage
just remove the debugmode and debugmode filter adjustments so players don't get it turned on without knowing how to turn off

PR #65248, opened 2023-04-22
Refactor memory cards

irwiss:
#### Summary
None

#### Purpose of change

Allows mass-downloading of memory cards
Fight against the RSI insurance claims filed by CDDA players

#### Describe the solution

Uses inventory_multiselector to allow multiple choices
Refactors downloading into activity, now takes ~1 minute per card.
Unhardcodes some of the stuff involved with the cards roll chances/amounts, recipe categories
"Encrypted" cards are now a possible failure mode, no "hacking" is attempted with a phone or xp gained.
Changes item ids/names a bit to allow migrations to reset item vars just in case (started dehardcoding on top of item vars since original implementation used item vars, but then moved over to a struct tied to itype, so that savegames don't bloat with static data)
Updates docs with the new fields
A couple office and house item groups where clean cards were spawned are replaced with unread ones (avatar doesn't have knowledge if the card is clean by looking at the card).

Why waste 2 minutes doing the thing instead of 2 hours automating it?

#### Testing

Almost the entire changeset is manually tested UI code unfortunately, so I mostly poked the menus

#### Additional context

https://user-images.githubusercontent.com/6560075/233801512-0e7e8a68-69ec-4460-9778-04dc025c05d3.mp4


doc/JSON_INFO.md
```diff
@@ -3470,6 +3470,28 @@ Any Item can be a container. To add the ability to contain things to an item, yo
...
+  "on_read_convert_to": "memory_card", // converted to this itype_id on read
+  "photos_chance": 0.33,               // 33% chance to contain new photos
+  "photos_amount": 3,                  // contains between 1 and 3 new photos
+  "songs_chance": 0.33,                // 33% chance to contain new songs
+  "songs_amount": 4,                   // contains between 1 and 4 new songs
+  "recipes_chance": 0.33,              // 33% chance to contain new recipes
+  "recipes_amount": 5,                 // contains between 1 and 5 new recipes
+  "recipes_level_min": 4,              // recipes will have at least level 4
+  "recipes_level_max": 8,              // recipes will have at most level 8
+  "recipes_categories": [ "CC_FOOD" ]  // recipes from CC_FOOD category
```
Maleclypse, line 3492: Could there be a json object for `"SECRET": true` //recipes will come from secret recipes in this category?  That way we could have memory cards in laboratories with a chance to have a single mutagen recipe or something else hidden?

irwiss: do you mean filtering by recipe "flags"? that's also doable i suppose, if CC_CHEM and level>6 isn't a good enough filter

Zireael07:
RSI isn't a joking matter, tbh...

doc/JSON_INFO.md
Maleclypse: I don't want to request scope creep if I don't have a planned use case that I'm going to implement immediately.

irwiss: Alright, give a poke if you need help implementing some specific filter, should be fairly easy now - add fields to the struct, read them from json and check the predicate in the activity

Cupidatis:
Could you make that for book saving? I will be very happy for bulk book saving...

PR #65254, opened 2023-04-23
gambeson fixes

anoobindisguise:
#### Summary
none

#### Purpose of change
Gambesons still had some typos and balance issues.

#### Describe the solution
Fixed some typos regarding nylon gambs referring to canvas gambs. Also lowered the encumbrance of the draped-legs layer per @Terrorforge request. Also fixed padded gambesons having the wrong encumbrance for torso.

Find a way to deal with the warmth values for the draped layers being too high, but you cannot currently do per-part warmth so you'd have to probably increase coverage/thickness.

### DIFF
--- a/data/json/items/armor/suits_protection.json
+++ b/data/json/items/armor/suits_protection.json
@@ -2193,7 +2193,7 @@
         ],
         "covers": [ "torso" ],
         "coverage": 100,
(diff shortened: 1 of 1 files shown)


NetSysFire:
Could potentially be a backport candidate.

PR #64699, opened 2023-03-30
Upgrade clang-tidy to LLVM 16

BrettDong:
#### Summary
None

#### Purpose of change
We have custom `clang-tidy` checks, but the `clang-tidy` version we are using does not support plugins out-of-the-box. So we have to use a patched version of `clang-tidy` to run our custom checks: https://github.com/jbytheway/clang-tidy-plugin-support. This makes setting up `clang-tidy` checks in local development environments a little bit involved, especially for non-Debian/Ubuntu distro users.

Fortunately, plugin support landed in a later LLVM release: https://reviews.llvm.org/D111100. This simplifies `clang-tidy` workflow a lot. After building our custom checks as a shared library, we can just run the checks with a vanilla version of  `clang-tidy`.

#### Describe the solution

- [x] Make custom `clang-tidy` checks compile with LLVM 16 API
- [x] Ensure `lit` test suite of custom `clang-tidy` checks pass
- [ ] Run `clang-tidy` 16 on CDDA code base

PR #65272, opened 2023-04-23
Split clang-tidy workflow into two parts

jbytheway:
#### Summary
Infrastructure "Brief description"

#### Purpose of change
The full clang-tidy run takes too long when it has to run on everything, and causes false failures on PR CI.

#### Describe the solution
Split it into one part for src and one for everything else.

I borrowed this idea from #64699 which implemented it differently.  I prefer to have more logic in shell scripts and less in GitHub workflows.

#### Testing
I checked by running the build script locally, but it can only really be tested on CI because I can't reproduce everything.

#### Additional context
Thanks to @BrettDong for the idea of how to achieve this.

### DIFF
--- a/build-scripts/clang-tidy.sh
+++ b/build-scripts/clang-tidy.sh
@@ -121,6 +121,19 @@ else
     fi
 fi
 
+printf "Subset to analyze: '%s'\n" "$CATA_CLANG_TIDY_SUBSET"
+
+# We might need to analyze only a subset of the files if they have been split
+# into multiple jobs for efficiency
+case "$CATA_CLANG_TIDY_SUBSET" in
+    ( src )
+        tidyable_cpp_files=$(printf '%s\n' "$tidyable_cpp_files" | grep '/src/')
+        ;;
+    ( other )
+        tidyable_cpp_files=$(printf '%s\n' "$tidyable_cpp_files" | grep -v '/src/')
+        ;;
+esac
+
 function analyze_files_in_random_order
 {
     if [ -n "$1" ]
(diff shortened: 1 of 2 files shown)


BrettDong:
Looks like this breaks `clang-tidy` check on other pull requests, but it's not obvious why they failed.

https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4789780565/jobs/8518061193?pr=65109

https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4787792131/jobs/8513519115?pr=65291

build-scripts/clang-tidy.sh
```diff
@@ -121,6 +121,19 @@ else
...
+printf "Subset to analyze: '%s'\n" "$CATA_CLANG_TIDY_SUBSET"
+
+# We might need to analyze only a subset of the files if they have been split
+# into multiple jobs for efficiency
+case "$CATA_CLANG_TIDY_SUBSET" in
+    ( src )
+        tidyable_cpp_files=$(printf '%s\n' "$tidyable_cpp_files" | grep '/src/')
+        ;;
+    ( other )
+        tidyable_cpp_files=$(printf '%s\n' "$tidyable_cpp_files" | grep -v '/src/')
```
BrettDong, line 133: Now I know why. `grep` returns with exit code 1 if no file is matched, and this aborts the whole shell script.

Qrox: these PRs both changed files in the `src` folder though so it's strange for grep to fail in that case.

PR #65279, opened 2023-04-23
Fix spelling mistakes

kwl01skz:
#### Summary
None

#### Purpose of change
weilded->wielded

#### Additional context

./data/json/npcs/isolated_road/isolated_road_npcs.json

```
    "type": "item_group",
    "id": "CODY_MILLER_weild",
    
    "type": "item_group",
    "id": "JAY_RUCKERS_weild",

```
These two ids have the same problem,
But id misspellings don't affect the game itself.
Should we modify these two ids directly?
Will changing these two ids cause existing archived items to disappear?

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -4843,7 +4843,7 @@ void item::enchantment_info( std::vector<iteminfo> &info, const iteminfo_query *
         enchantment_info_helper( *this, enchantment::has::WIELD, always, active, inactive );
         if( !always.empty() || !active.empty() || !inactive.empty() ) {
             info.emplace_back( "DESCRIPTION",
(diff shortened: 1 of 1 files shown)


PatrikLundell:
Changing Ids will cause problems with loading saves containing these Ids, probably resulting in error messages and the loss of whatever the Ids were representing. Thus, there would be a need for migration of the old Id to the new one (and I can't describe how that would be done, unfortunately. An alternative is to retain the old Ids and mark them as obsolete so no new ones will be generated and the new ones would be generated instead. This may or may not require logic glue to ensure both the old and new Ids are compatible (e.g. in crafting recipes, which can lead to "duplicate" recipes or recipe entries).

PR #65290, opened 2023-04-24
Add 'bearable protein rations' and recipe for them

Grnkchi:
#### Summary
Content "Add 'bearable protein rations' and the recipe to make them"

#### Purpose of change

Protein rations are described as tasting horrible and having an enjoyment modifier to match. However, at this time there are no recipes to make them more palatable - in a survival scenario, even just the basic concept of drizzling some honey onto it salvaged from a kitchen would likely improve the taste if not the texture.

#### Describe the solution

Added a new item, the "bearable protein ration," made quickly and easily by combining protein rations and one of several sweetening ingredients, including both forms of honey, sugar, caramel, maple/coffee/beet syrup, etc. This marginally increases the amount of kcal per charge in most cases, save for caramel, which is the most efficient for kcal/volume, but more expensive than simply using sugar.

In the current version, the enjoyment and quench reductions are mitigated (from -5/-9 to -1/-5, respectively) but not removed, and the 'gross food' effect is taken off, as it is now - as the name says - bearable to eat.

Creating multiple variants of the 'bearable' version, including one which was faster, easier and cheaper to make but wouldn't rot and would have a reduced effect, and a second version which was more expensive in both time and materials while also having a spoil timer in exchange for actually reaching a morale-positive enjoyment level. This option was ultimately simplified for a single version, as the chosen ingredients usually have very long to indefinite shelf lives.

#### Testing

The game boots and loads into worlds without any errors, and the bearable protein ration is craftable immediately at game start, as pouring some honey or sprinkling some sugar in a protein ration is probably not difficult enough to require any skill at food handling.

### DIFF
--- a/data/json/items/comestibles/protein.json
+++ b/data/json/items/comestibles/protein.json
@@ -76,6 +76,26 @@
     "vitamins": [ [ "calcium", 30 ], [ "iron", 30 ], [ "vitC", 30 ], [ "bad_food", 5 ] ]
   },
   {
+    "id": "bearable_protein_ration",
+    "type": "COMESTIBLE",
+    "comestible_type": "FOOD",
+    "name": { "str": "bearable protein ration" },
+    "//": "Inspired by, but not based on, a true story",
+    "description": "A standard protein ration from an evac shelter, made significantly more bearable by sweetening it. Despite your efforts, the texture of the meal could not be saved.",
+    "weight": "160 g",
+    "volume": "230 ml",
+    "price": 500,
+    "price_postapoc": 60,
+    "flags": [ "EDIBLE_FROZEN" ],
+    "fun": -1,
+    "quench": -5,
+    "material": [ "fruit", "veggy" ],
+    "symbol": "%",
+    "color": "green",
+    "calories": 430,
+    "vitamins": [ [ "calcium", 35 ], [ "iron", 35 ], [ "vitC", 35 ] ]
+  },
+  {
     "id": "protein_shake",
     "copy-from": "protein_drink",
     "type": "COMESTIBLE",
--- a/data/json/recipes/food/other.json
+++ b/data/json/recipes/food/other.json
@@ -136,10 +136,19 @@
     "skill_used": "cooking",
     "time": "15 s",
     "autolearn": true,
-    "components": [ 
-		[ [ "protein_bar_evac", 1 ] ],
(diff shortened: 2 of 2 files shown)


PatrikLundell:
As a side note, it can be noted that mixing food ingredients with long shelf lives doesn't necessarily result in a product with a long shelf life. The longevity of honey, for instance, largely comes from the high sugar concentration, and diluting it may result in something that's suddenly very perishable when various organisms can start to make use of all these nutrients without suffering from the deadly dehydration effects.

Grnkchi:
To answer those concerns in the order they were presented:

While they are described as tasting so bad as to cause the company to go bankrupt, the player characters so far are apparently too oblivious as to the fact that it can be remedied to some extent, and this was simply the fastest and simplest way to do so I could think of. For example, recipes could be added to simply break/cut apart the bar and use it as an ingredient in other dishes to mask the taste, since this is a survival scenario, and the calories and nutrients it provides are too good to give up, regardless of the taste. However, if that answer is insufficient to convince you, then perhaps the Gross Food "vitamin" could be re-added at a reduced amount - the base Protein Ration has a value of 5, perhaps the bearable version could be set to somewhere in the 1-3 range.

The spoilage of mixed ingredients was briefly mentioned in the dev Discord regarding this PR, and a compromise was reached in the sense that the player character simply deciding to add the 'flavoring ingredient' to the ration when they eat it, as opposed to immediately pouring/adding whatever it is, and the recipe being an abstraction of that. While this still isn't perfectly realistic (for example, using chocolate bars preserves their calories indefinitely in the form of the flavored ration bar), I am patently unwilling to create several crafting recipes sorted by spoilage time with varying shelf lifes for what is meant to be a small change. 

And now for a question of my own as this is my first PR: how do I solve the two failed checks? Because I have no clue how to fix either of them, as of now.

Nebnis:
I believe the calories you added are different than the total ingredients.

anoobindisguise:
@I-am-Erk You said that the "Gross Food" vitamin was specifically supposed to be for these protein bars, and that it was probably an error that Hardtack had the Gross Food vitamin - is there something special about these bars that would make them specifically unsalvageable like this?

I-am-Erk:
Gross food was indeed meant to be explicit to these, which are meant to be so disgusting as to be a joke. I don't really think adding some flavours should help with that, as the whole point of adding that feature is to keep soypelusa bars from being too good of a long term nutrition source. They're extremely plentiful and nutritious, why do they need to be even better?

Grnkchi:
Alright, that sounds like a reasonable explanation as to why this won't be getting merged. I'll probably still disseminate it as a mod, but that's neither here nor there.

I-am-Erk:
Thanks for understanding :)

PR #65293, opened 2023-04-24
Vehicle interaction menu fixes

irwiss:
#### Summary
None

#### Purpose of change

Various vehicle UI fixes to make vehicle interaction menu less awkward - removes the invalid or weird options.

#### Describe the solution

* d8d7dec4db04330d8d66ba3a3379b8803d2c1cca Hides individual engine controls if there's only one engine
* 5c1860ebed85a20d86d1fe460cd0861cead0db20 Makes vehicle_part::fuel_current work correctly for muscle fuel, removing "turn on/off engine" from examining muscle vehicles like bicycles - This may cause side effects but seems most places that care about muscle engines also do `is_perpetual_type()` check which is what made them work before this change
* bf1ac8bfd34e70a4781945528988e427448d7848 Makes cruise control on the default and only option - there's slight difference in how controls will skip turns when it is off but it creates funny menus where canoes/wheelchairs have "cruise control" on them

#### Testing

Menus on vehicles like canoes/bicycle/wheelchair should be practically empty - examine/remember position, may be honk horn
No individual engine controls should appear unless more than 1 engine exists
Cruise control option is removed, all vehicles now behave as when it's "on"


PatrikLundell:
I don't know if it's an issue, but what happens if you have a vehicle with two engines, you turn one off, and then remove the other one (I don't want to have an engine turned off but being unable to turn it back on without having to install a second engine to turn it back on)?

irwiss:
nothing related to engine starting has changed, worst case start_engines has fallback to enable all if none are enabled

PR #64954, opened 2023-04-09
Fog and mist weather

GuardianDll:
#### Summary
Content "add new weather: mist and fog. maybe."
#### Purpose of change
I want to make a fog
#### Describe the solution
Currently fog and mist require extremely high humidity, more than 95%, small wind, and some level of temperature drop
Also @andrei8l fixed autotests, that require clean vision to pass, 
#### Testing
~~And here is the problem - i can't properly test this solution, because such high humidity happen only when it rains. Also i am not 100% sure i get how it should be done - from what i know there should be also a dew point calculation, but i still don't understand how it should be utilized, as when relative humidity is 100%, the dew point is equal the temperature itself~~ yep, now everything works as expected
image

PR #65298, opened 2023-04-25
math_parser/eoc: migrate arith. weather to math

andrei8l:
#### Summary
None

#### Purpose of change
Port another `arithmetic` function to `math`. `weather` was used recently for #64954 and it's a small enough target.

#### Describe the solution
Move the code from `arithmetic` to `math`.
Use Kelvin for temperature to match internal representation.
Add temperature conversion functions.

Fix the calculations and conditions for fog/mist: ported from EOC to `jmath_function`. Added reference for the formula we used and re-converted it to Celsius. Fixed the conditions for mist to better match reality.

Continuing to use Fahrenheit: Temperature is internally already Kelvin so it makes more sense to expose it to EOCs in the same unit. Weather aspects use a mixed bag of undocumented metric and non-metric units atm and I'd rather port them all to SI if possible.

#### Testing
Existing test unit.

Both mist and fog still occur in game. Mist/fog no longer occur by default on a fresh game simply because `dew_point_factor` hasn't been calculated yet.

#### Additional context
I promised I'd do this for #64954 but I got sidetracked IRL. Better late than never though.

### DIFF
--- a/data/json/npcs/TALK_TEST.json
+++ b/data/json/npcs/TALK_TEST.json
@@ -1144,22 +1144,22 @@
       {
         "text": "Temperature is 21.",
         "topic": "TALK_DONE",
-        "condition": { "compare_num": [ { "weather": "temperature" }, "=", { "const": 21 } ] }
+        "condition": { "math": [ "fahrenheit( weather('temperature') )", "==", "21" ] }
       },
       {
         "text": "Windpower is 15.",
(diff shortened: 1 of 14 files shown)


andrei8l:
I'm putting this in draft until #65379 gets merged so I can replace the dew point EOC with a jmath function.

andrei8l:
Thanks!

PR #65307, opened 2023-04-25
Context Vars

bombasticSlacks:
#### Summary
Infrastructure "EOCs can have Context Vars"

#### Purpose of change
EOCs should be able to have variables exclusive to them and any EOCs they call, keeping track of an internal context which doesn't need to be attached to the player, npcs, or the global state.

#### Describe the solution
Added a new context keymap attached to each dialogue, made dialogue non const everywhere it needed to be for this to work, added some unit tests.

#### Testing
Added unit tests.

#### Additional context
~~Going to add documentation and additional details when I hook this up to events.~~

It's hooked up, documentation is included

### DIFF
--- a/src/math_parser.cpp
+++ b/src/math_parser.cpp
@@ -183,7 +183,7 @@ bool is_assign_target( thingie const &thing )
 func::func( std::vector<thingie> &&params_, math_func::f_t f_ ) : params( params_ ),
     f( f_ ) {}
 
-double func::eval( dialogue const &d ) const
+double func::eval( dialogue &d ) const
 {
     std::vector<double> elems( params.size() );
     std::transform( params.begin(), params.end(), elems.begin(),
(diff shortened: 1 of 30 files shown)


Ramza13:
Between this, event EOCs and the template stuff going away some of the biggest problems I didn't see solutions for are going away.  The value of a fresh set of eyes can't be understated.  Thanks!

src/math_parser.cpp
```diff
@@ -549,6 +549,10 @@ void math_exp::math_exp_impl::new_var( std::string_view str )
                 debugmsg( "Unknown scope %c in variable %.*s", str[0], str.size(), str.data() );
         }
     }
+    if( str.size() > 1 && str[0] == '_' ) {
```
andrei8l, line 552:
```suggestion
    } else if( str.size() > 1 && str[0] == '_' ) {
```
Otherwise variables like `u__blorg` get scoped twice.

MNG-cataclysm:
This is great, but could you write some documentation for this in the respective file, please? That would help us use these new tools effectively.

PR #65308, opened 2023-04-26
remove gross food from hardtack

anoobindisguise:
#### Summary
Balance "hardtack doesn't have the gross food vitamin"

#### Purpose of change
Gross food is supposed to specifically represent protein rations doing psychic damage to the player and not show up on anything else per @I-am-Erk yet Hardtack was twice as gross per tack as protein bars were.

#### Describe the solution
Remove the vitamin.

### DIFF
--- a/data/json/items/comestibles/bread.json
+++ b/data/json/items/comestibles/bread.json
@@ -480,7 +480,7 @@
     "volume": "250 ml",
     "charges": 2,
     "flags": [ "EDIBLE_FROZEN" ],
-    "vitamins": [ [ "calcium", 2 ], [ "iron", 15 ], [ "bad_food", 10 ] ],
+    "vitamins": [ [ "calcium", 2 ], [ "iron", 15 ] ],
     "stack_size": 20,
     "fun": -1
   },
--- a/data/json/items/comestibles/bread.json
+++ b/data/json/items/comestibles/bread.json
@@ -459,7 +459,7 @@
     "volume": "250 ml",
     "charges": 2,
     "flags": [ "EDIBLE_FROZEN" ],
-    "vitamins": [ [ "calcium", 2 ], [ "iron", 15 ], [ "bad_food", 10 ] ],
+    "vitamins": [ [ "calcium", 2 ], [ "iron", 15 ] ],
     "stack_size": 20,
     "fun": -4
   },
--- a/data/json/items/comestibles/fruit_dishes.json
+++ b/data/json/items/comestibles/fruit_dishes.json
@@ -526,7 +526,7 @@
     "charges": 4,
     "phase": "liquid",
     "flags": [ "EATEN_HOT", "USE_EAT_VERB" ],
-    "vitamins": [ [ "vitC", 2 ], [ "iron", 15 ], [ "bad_food", 2 ] ]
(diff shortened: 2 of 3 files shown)


Maleclypse:
"Gnaw on a chunk of hardtack for 20-30 minutes if you’d rather not soak it. Dry hardtack is not meant to be eaten quickly. Nibble and gnaw at the edges and corners of the hardtack, then eat the small bits that break off in your mouth.[[12]](https://www.wikihow.com/Eat-Hardtack#_note-12)
Think of the hardtack as if it were a bread version of beef jerky. Work through the food slowly and don’t bite off too much at once, or you’ll be chewing 1 bite for 15 minutes."  Since the PC is only capable of eating hardtack all in one bite over in seconds I think psychic damage is likely.

fairyarmadillo:
The gross food debuff feels redundant in general imo. The negative fun factor on protein rations should be enough to convey that they suck. It's not like they're uniquely bad among foods in the cataclysm, considering you can eat raw mutant organs, rancid butter, and poisonous chemicals.

Vaskritaya:
> "Gnaw on a chunk of hardtack for 20-30 minutes if you’d rather not soak it. Dry hardtack is not meant to be eaten quickly. Nibble and gnaw at the edges and corners of the hardtack, then eat the small bits that break off in your mouth.[[12]](https://www.wikihow.com/Eat-Hardtack#_note-12) Think of the hardtack as if it were a bread version of beef jerky. Work through the food slowly and don’t bite off too much at once, or you’ll be chewing 1 bite for 15 minutes." Since the PC is only capable of eating hardtack all in one bite over in seconds I think psychic damage is likely.

I can't imagine this was ever meant to be played straight as the player character not understanding how to eat hardtack, and, in their desperation, ramming the entire piece into their mouth and swallowing as hard as they can with all the pain and choking that brings.

It's much more sensible to just see it as the classic reoccurring problem of high-time, low-input actions. Either you have it be unrealistically quick, or you have it take a realistic amount of time but demand 100% of the player character's attention that entire time. Both are silly, but you often have to pick one or the other. For a food you'd often be "working on" while in the field, the former is definitely the more sensible way of handling it.

anoobindisguise:
> "Gnaw on a chunk of hardtack for 20-30 minutes if you’d rather not soak it. Dry hardtack is not meant to be eaten quickly. Nibble and gnaw at the edges and corners of the hardtack, then eat the small bits that break off in your mouth.[[12]](https://www.wikihow.com/Eat-Hardtack#_note-12) Think of the hardtack as if it were a bread version of beef jerky. Work through the food slowly and don’t bite off too much at once, or you’ll be chewing 1 bite for 15 minutes." Since the PC is only capable of eating hardtack all in one bite over in seconds I think psychic damage is likely.

So the issue here is that while raw hardtack should be unfun to eat the vitamin inherits from anything you make with it and it was also assigned to stuff specifically designed to make hardtack tastier.

I-am-Erk:
I don't think the gross food feature is appropriate here. We can assume a player is munching their hardtack slowly in the field, as said there's just no simulation value to having some way to represent slowly chewing it.

It's really meant to be exclusive to the protein rations, which are intended to be comically disgusting as a play balance issue and a joke.

PR #65320, opened 2023-04-26
Do not cancel clang-tidy run when a subset finishes

Qrox:
#### Summary
None

#### Purpose of change
When working on #65223 I noticed that the recent change to divide clang-tidy run into two subsets has caused the `src` subset to always be canceled because the `tests` run is finished earlier. This prevents all warnings to be reported, and unlike the build matrix, the two subsets are checking different sources so they should be all completed IMO.

#### Describe the solution
Do not cancel clang-tidy runs when one subset finishes.

Not doing it.

#### Testing
Tested in #65223 and the two subsets finished without interruption.

### DIFF
--- a/.github/workflows/clang-tidy.yml
+++ b/.github/workflows/clang-tidy.yml
@@ -40,7 +40,7 @@ jobs:
     if: ${{ needs.skip-duplicates.outputs.should_skip != 'true' }}
 
     strategy:
-      fail-fast: true
+      fail-fast: false
       matrix:
         # To make the run finish in the run time limit, we split it up into two
         # parts: the src directory and everything else


andrei8l:
Still failed for me [here](https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4826902743/jobs/8599162737?pr=65352#step:7:1711), unless I misunderstand the issue.

Qrox:
In your run the `src` workflow failed but the `test` workflow completed after the failure instead of being canceled, which is what this PR intended to do.

According to this https://github.com/CleverRaven/Cataclysm-DDA/pull/65272#discussion_r1176334742 the failure of `src` itself is probably caused by `grep` failing because no file is matched, but since your PR also changes files in the `src` folder I'm not sure how that happened.

PR #65323, opened 2023-04-26
Fix vehicles' initial fuel spawns

### LINKED ISSUE #65063: Vehicles spawn without water
### Describe the bug

Vehicles that have tanks that should be filled with water or clean water spawn with empty "water tanks". 
Affected vehicles are:
- cabin_cruiser_wood 4 tanks with water_clean
- water_cart 1 tank_small with water_clean
- fire_engine 9 tanks with water
- van_motorhome 1 tank with water_clean
- lux_rv 1 tank with water_clean
- surv_rv 1 tank with water_clean
- rv 1 tank with water_clean
- bus_tour 1 tank with water_clean
- car_atomic_flame 1 tank with water, 1 tank with water_clean (CrazyCataclysm mod)

### Attach save file

N/A

### Steps to reproduce

1. Spawn undamaged fire_engine via Debug Menu
2. Examine vehicle

### Expected behavior

The tanks that have the fuel set to "water" or "water_clean" should spawn 80% full, just like the tanks that are marked with "gasoline" etc. In a perfect world I should be able to spawn a vehicle with any (comestible) liquid filled in the tanks for maximum modding support.

### Screenshots

_No response_

- OS: Linux
    - OS Version: Distributor ID: LinuxMint; Description: Linux Mint 19.2 Tina; Release: 19.2; Codename: tina; (Ubuntu 18.04 equivalent)
- Game Version: f65b2bc [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

second test with latest version:
- OS: Linux
    - OS Version: Distributor ID: LinuxMint; Description: Linux Mint 19.2 Tina; Release: 19.2; Codename: tina; 
- Game Version: 026046b-dirty [64-bit]
- Graphics Version: Curses
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Test Cart [cart]//custom test mod with different liquids in tanks
]

### Additional context

I've also tried looking for the problematic vehicles in the wild on the off chance that they have a bit of water, but to no avail. This issue seems systemic. All liquids of type "AMMO" (any defined in data/json/items/fuel.json) are spawned without any issue, only "COMESTIBLE" (like water) are silently dropped.

irwiss:
#### Summary
None

#### Purpose of change

Refactors fuel/liquid/plutonium spawns

Fixes #65063 - removes division by legacy_volume_factor that set water amount to 0
Fixes hybrid cars always spawning without fuel
Fixes all lightly damaged cars fuel capped between 1% and 7% of tank capacity
Fixes having tanks on mint vehicles spawning with just 48L filled out of 60L capacity

#### Describe the solution

This PR effectively increases (by a bit) the amount of fuel available in lightly damaged vehicles (the ones randomly spawned by mapgen) but most spawns should be below half full, if that's a major problem the mean value can be moved lower, but going lower seems extremely game-ified when reflected on the rest of the in-game world; mostly intact gas stations, enough fuel around to engorge packs of dozens of gasoline zombies etc.

Previous 1%-7% fuel in tanks IMO makes no sense (no used cars in the game universe having more than-), nor the (invalid/outdated) comment near veh_fuel_mult stating that it should actually be 7%-35%

This PR also stops reactors spawning full of plutonium (except for recators on mint vehicles) - plutonium follows normal distribution with same settings as well, this was probably a bug.

The random rolls for amount of fuel/battery/plutonium are now independent of each other

Make `vehicle_prototype::finalize()` assign fuel tanks as full, removing vehicle::init_state accessing a prototype field for it
Make `vehicle::init_state` randomize fuel amount for lightly damaged vehicles using normal distribution ( mean=0.3 stddev=0.15 ) clamped between 5% and 95% capacity

These parameters should let vehicles spawn with:
More than 3/4 full tanks ~0.13% rolls
More than 1/2 full tanks ~10% rolls
More than 1/4 filled ~63% rolls

Define fuel for tanks in 2 hybrid car prototypes so they have a chance to spawn fueled

#### Testing

Walk around triggering mapgen vehicles, check their fuel/battery/reactor fuels - they should make sense

Stuff that can be tested with debug:
Spawned fire engines should spawn with water
RVs should spawn with water


irwiss:
These pools aren't changed and some already act as described. A change to have some vehicles only "out of fuel" but intact fits in a diffeent PR because that makes a much bigger balance change; a canister of gas is trivial to acquire.

This PR only fixes a few bugs and makes initial car fuel a gaussian distribution around 30% mean, rather than every non-showroom car being limited to single digit percent of fuel amount, so for extremely lucky rolls you can get a decent amount of fuel and not top out at a fixed ~4.2L of fuel for 60L tank as the game universe's max allowed value

stubkan:
This makes the game way too easy.

Im constantly driving around 99% full, I have stockpiles of fuel every intersection inside abandoned cars.  I am never going to run out of fuel now.

I think this PR should be reversed or at least nerfed to a more sensible and balanced level.

Balance aside, what about npc looting?  People siphon fuel as gas stations are unmanned and this is 2 months after the cataclysm so quite a lot has gone on.  People would be trading gasoline like the arsonist.  Cars should be more empty as people zombify/die while driving and wouldn't have the werewithal to turn their engine off (especially if they couldnt even park their car but just drift to a stop in the middle of the road)

stubkan:
image

Screenshot shows how much gas is available (my vehicle is already full)

Bear in mind service stations need to be refilled once a week - more if in urban areas - or they run out of fuel.  The normal starting condition puts the game 2 months after the cataclysm began.  Unless the service stations were continuously filled during the apocalypse, fuel would have run out very quickly.

A second quirk that seems to have arisen from this PR is that every fuel tank on the vehicle is filled identically.  Previously you'd get multiple tanks of varying emptiness - lucky to get one filled - but now they appear to all be equally full.  I think having each tank filled separately like before would be better.  Since people use them separately, except for large vehicles such as fire trucks, trucks.  You would have a water tank and a fuel tank on a RV, or a reserve tank on your car that you'd fill separately depending on your uses - not have them both somehow identically filled even though you drive off one tank only.

PR #65324, opened 2023-04-26
Updated Phase Immersion Suit

halosammy:
#### Summary
Balance "Prevented melee from breaking the Phase Immersion Suit"

#### Purpose of change

Recent overhauls to other exoskeletons have added the UNBREAKABLE_MELEE tag to address the inherent issue of compromising protection around you entire body if you punch things a lot. Given that the Phase Immersion Suit is the only exoskeleton without such a property, it seemed appropriate to round things off unless someone has a compelling reason not to.

I considered the angle that the Phase Immersion Suit is more for environmental protection than combat and thus not strictly meant for melee, but the RM13 has the same flag and similar immunities, so there is precedent. Unlike other, HAZMAT-type suits, of course, this clothing is reinforced by several rigid components such as layered carbide and carbon steel.

#### Describe the solution

Simply added the same flag that the other exoskeletons have.

Remove UNBREAKABLE_MELEE from the exoskeletons if we want to go the route of assuming that damaging the hands of such armor can compromise its overall properties. Either that or overhauling the entire system so that armor like this can have its immunities “segmented” or somesuch. That seems like a whole different issue I’m not qualified to handle.

#### Testing

N/A

#### Additional context

N/A

### DIFF
--- a/data/json/items/tool_armor.json
+++ b/data/json/items/tool_armor.json
@@ -1466,7 +1466,8 @@
       "PADDED",
       "USE_UPS",
       "NO_UNLOAD",
-      "COMBAT_TOGGLEABLE"
+      "COMBAT_TOGGLEABLE",
+      "UNBREAKABLE_MELEE"
     ],
     "relic_data": { "passive_effects": [ { "id": "ench_climate_control_all" } ] },
     "armor": [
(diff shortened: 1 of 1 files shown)


SirPendrak:
Please add this flag to its "ON" version too

halosammy:
> Please add this flag to its "ON" version too

Fixed.

PR #65327, opened 2023-04-27
Document "location" for vehicle parts

RenechCDDA:
#### Summary
None

#### Purpose of change
I was investigating why certain vehicle parts couldn't be installed together and found that this field is not documented.

#### Describe the solution
Document the field.

#### Testing
Docs change only


hexagonrecursion (approved):

Fris0uman:
It'd be nice to add a list of valid locations ids

RenechCDDA:
Technically any string is a valid location. I know that's not *super* obvious from the text, but I didn't elaborate on it because it's very difficult to explain.

For the purposes of displaying which vehicle part is on top there are 9 or so specific locations that the game will consider, but there's also some logic checks elsewhere which look at random ones... the field is really a suggestion that most often trips you up in terms of part installing.

PR #65329, opened 2023-04-27
Addition of Profession requirements

RanikOrtega:
#### Summary
Content "Profession Requirements"

#### Purpose of change
Locks out some of the most heavily equipped starting professions behind location, kill count or survival time achievements.

#### Describe the solution
Made Professions require certain achievements to unlock. 

None

#### Testing
Tested new world and verified professions were locked. Spawned and ran to police station and unlocked achievement. Died and checked if new professions were available. 
CDDA-Professions
CDDA-Police Achievement
CDDA-Unlocked Police Professions


data/json/achievements.json
```diff
@@ -2202,5 +2216,19 @@
         "description": "Reach a Mine"
       }
     ]
+  },
+  {
+    "id": "achievement_reach_police_station",
+    "type": "achievement",
+    "name": "I shot the Sheriff!  What?!  He was a zombie!",
```
Fris0uman, line 2223: But this is not an achievement for killing a zombie cop though

data/json/professions.json
```diff
@@ -1593,6 +1607,7 @@
     "type": "profession",
     "id": "rescuer",
     "name": { "male": "Rescuer", "female": "Rescuer" },
+    "requirement": "achievement_kill_1000_monsters",
```
Fris0uman, line 1610: Shouldn't this rather be tied to finding the mi-go camp?

data/json/achievements.json
RanikOrtega: Zombie cops are a dime a dozen and spawn everywhere. It makes more sense to lock the profession behind the location rather than zombie cops.

data/json/professions.json
RanikOrtega: changing and reuploading

data/json/achievements.json
Fris0uman: Yes, I was more refering to the joke that's not so much related to the actual achievement

RanikOrtega: Ah I see your point. I may have to think of another description of the achievement and alter it in the next round of achievement additions.

Fris0uman: I'm not sure what you mean by next round, please fix it in this PR.

RanikOrtega: Changed to "Zombie PD, to protect and zombify!"

data/json/professions.json
```diff
@@ -1844,6 +1862,7 @@
     "type": "profession",
     "id": "heli_pilot",
     "name": "Helicopter Pilot",
+    "requirement": "achievement_reach_military_bunker",
```
Termineitor244, line 1865: Wouldn't this be a problem with the "Last Flight" scenario? I remember that the scenario in question only has this profession as an option...

And I don't think a normal pilot that does not spawn with weapons (or an helicopter if chosen outside the scenario) really needs to be locked behind reaching a military bunker, I feel like this one could be done with no requirement at all.

data/json/professions.json
```diff
@@ -1704,6 +1720,7 @@
     "type": "profession",
     "id": "security",
     "name": "Security Guard",
+    "requirement": "achievement_reach_megastore",
```
Termineitor244, line 1723: This profession is the only one allowed for the Trans-Coast Logistics Concourse scenario, which already has the `achievement_reach_trans_coast_lab` requirement, so it could end in the situation where the only profession is still locked behind an achievement even though you already unlocked the scenario...

An option to fix this would be to add the `labtech` and `medic` professions as option to the scenario in question, since those make sense in the context of the scenario and are not locked behind anything yet.

Termineitor244 (changes requested):

data/json/professions.json
RanikOrtega: Good eye. Need to remove the requirement from this profession.

data/json/professions.json
RanikOrtega: Changes Made

Issue #65330, opened 2023-04-27
cold resistance trait disables weak stomach


hr-shafiei:
### Describe the bug

cold resistance trait disables weak stomach.
is this unintentional or is that a thing ?
i mean, i myself am pretty strong against cold but im not sure about the quality of my stomach !

### Attach save file

-

### Steps to reproduce

-

### Expected behavior

-

### Versions and configuration

-

Fris0uman:
Fixed by https://github.com/CleverRaven/Cataclysm-DDA/issues/65330

PR #65335, opened 2023-04-27
Seperate weather tolerance from stomach related traits

aMegaSloth:
#### Summary
Bugfixes "Separates weather tolerance traits from stomach related ones"

#### Purpose of change
Addresses issue #65330. Seems like to me that being tolerant of weather shouldn't exclude stomach related traits.

#### Describe the solution
Separate the mutation category so they do not conflict.

### DIFF
--- a/data/json/mutations/mutation_type.json
+++ b/data/json/mutations/mutation_type.json
@@ -17,6 +17,10 @@
   },
   {
     "type": "mutation_type",
+    "id": "ACCLIMATIZATION"
+  },
+  {
+    "type": "mutation_type",
     "id": "ATTRACTIVENESS"
   },
   {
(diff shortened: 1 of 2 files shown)


hr-shafiei:
thanks, that was my issue
but is that possible to simply fix the trait in a way which it doesnt cancel the other one ?
instead of adding a new mutation type ?

aMegaSloth:
You need to add a new mutation type because as it stands they are in the same mutation 'type' as the stomach related ones as mutation categories are mutually exclusive.

hr-shafiei:
crushes the game !
at creating a new character

aMegaSloth:
Not sure where that crash is coming from. It loads just fine past creation for me.

Maleclypse:
You have errors that will prevent this from being merged on both the new mutations. 
IMG_0136

hexagonrecursion (approved):

anoobindisguise:
I think Cold Tolerance could have ECTOTHERMIC on it at least though.

aMegaSloth:
I've thought about expanding the category a little with a few more traits, but felt it was out of scope from a simple bug fix. I'll ponder it and probably include additional things in another PR.

PR #65341, opened 2023-04-27
Display why vehicle parts cannot be installed

RenechCDDA:
#### Summary
Interface "Display why vehicle parts cannot be installed"

#### Purpose of change
The rules for installing vehicle parts are incredibly arcane and hidden in `vehicle::can_mount`. Average players have little to no feedback for why parts just *disappear* from the list of parts to be installed.

#### Describe the solution
Stop filtering uninstallable parts. Add them to the list of parts that can be installed along with those missing tools/components, so they are displayed after parts that are ready to be installed. Add return values for `vehicle::can_mount` with strings for why a part can't be installed. Return those strings to the player in the same pane as other requirements are listed.

#### Testing
image

#### Additional context
Special thanks to @mqrause for great advice as always, especially on ret_val and examples of its implementation

### DIFF
--- a/src/veh_interact.cpp
+++ b/src/veh_interact.cpp
@@ -877,7 +877,8 @@ bool veh_interact::update_part_requirements()
                                   veh->part( cpart ).mount, sel_vpart_info->get_id() );
     if( !can_mount.success() ) {
         ok = false;
-        nmsg += _( "<color_white>Cannot install due to:</color>\n> " ) + can_mount.str() + "\n";
+        nmsg += _( "<color_white>Cannot install due to:</color>\n> " ) + colorize( can_mount.str(),
+                c_red ) + "\n";
     }
 
     sel_vpart_info->format_description( nmsg, c_light_gray, getmaxx( w_msg ) - 4 );
--- a/src/veh_interact.cpp
+++ b/src/veh_interact.cpp
@@ -873,6 +873,12 @@ bool veh_interact::update_part_requirements()
     }
     nmsg += res.second;
 
+    if( !veh->can_mount( veh->part( cpart ).mount, sel_vpart_info->get_id() ).success() ) {
+        ok = false;
+        nmsg += _( "<color_white>Cannot install due to:</color>\n" ) + veh->can_mount( veh->part(
+                    cpart ).mount, sel_vpart_info->get_id() ).str() + "\n";
+    }
+
     sel_vpart_info->format_description( nmsg, c_light_gray, getmaxx( w_msg ) - 4 );
 
     msg = colorize( nmsg, c_light_gray );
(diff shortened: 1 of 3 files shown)


src/vehicle.h
```diff
@@ -843,6 +843,35 @@ class vehicle
         vehicle &operator=( const vehicle & ) = default;
 
     public:
+        enum class install_code : int {
+            SUCCESS,
+            // part doesn't exist
+            ERR_INVALID,
```
irwiss, line 849: these ERR_ prefixes are duplicating the functionality of ret_val's success bool

also ret_val&lt;void&gt; will do the same thing ret_val&lt;install_code&gt; does but with less code and no extra enum, unless you have a plan for more functionality on top of the enum

RenechCDDA: Thank you greatly, these were all very good suggestions. I tried to enter you as a co-author on https://github.com/CleverRaven/Cataclysm-DDA/pull/65341/commits/28aa5852e863f7929745bf482372ac05fbb84954 after batching them (I thought the colorization would be more work than it was), but github didn't like my formatting. I hope that's fine?

irwiss: Don't worry about it, just a thanks is enough

pjf:
@RenechCDDA : Thank you *so much* for this. ♥

KHeket:
Now due this update you cant install frames no more because game saying, that

<details>
  <summary>Screenshot</summary>

image

image


  
</details>

RenechCDDA:
@KHeket Can you please open an issue with your save? I tried to reproduce and got.... a segfault crash, great. So I need to see if the behavior is different in different circumstances.

KHeket:
> @KHeket Can you please open an issue with your save? I tried to reproduce and got.... a segfault crash, great. So I need to see if the behavior is different in different circumstances.

Done https://github.com/CleverRaven/Cataclysm-DDA/issues/65470

src/veh_interact.cpp
```diff
@@ -873,6 +873,14 @@ bool veh_interact::update_part_requirements()
     }
     nmsg += res.second;
 
+    ret_val<void> can_mount = veh->can_mount(
+                                  veh->part( cpart ).mount, sel_vpart_info->get_id() );
```
irwiss, line 877: the main problem comes from here - cpart is only valid some of the time, when on empty square it's probably -1, you can test if veh_interact::dd is the mount point, i can't remember if you need to rotate() it to the correct orientation

looks like adding all parts to req_missing is also misbehaving but that's a smaller problem

RenechCDDA: Yeah looks like it's `dd.rotate( 2 )`. I was so pleased when I saw cpart so I 'could avoid that whole mess'. Shame on me, I suppose.

What's this about adding all parts to req_missing misbehaving?

irwiss: i only checked briefly and not on my pc right now, but some parts had white (installable as opposed to gray) names, but were sorted together with uninstallable ones, if you don't see them in your testing could be just a hiccup, didn't have time to dig into it

RenechCDDA: Oh no, I noticed that right before pushing the PR originally. It's tied to veh_interact::can_potentially_install being used in veh_interact::display_list. ~~I totally ignored that and slammed the PR ahead regardless~~ I gave it some thought over that night and figured it still does provide useful feedback to the player, although it is a slight change from previous behavior. Parts should only be getting colored white if you have all the components to install them, but it's agnostic of whether they're valid to install *on that tile* (would still be able to tell based on their placement in the list).

PR #65353, opened 2023-04-28
Lodge basement variant - normal

Maleclypse:
#### Summary
None

#### Purpose of change
Not every hunting lodge should have either a sex dungeon, drug trafficking, or TTPG lair.  Some should just have a normal basement.

#### Describe the solution
Adds a normal basement.

Making the laboratory entrance basement work.

#### Testing
tested and it works perfectly.

### DIFF
--- a/data/json/mapgen/hunting_lodge.json
+++ b/data/json/mapgen/hunting_lodge.json
@@ -233,6 +233,49 @@
   {
     "type": "mapgen",
     "method": "json",
+    "om_terrain": [ [ "lodge_basement_residential1", "lodge_basement_residential2" ] ],
+    "weight": 100,
+    "object": {
+      "fill_ter": "t_soil",
+      "rows": [
+        "                                                ",
+        "                                                ",
+        "                                                ",
+        "                                                ",
+        "                                                ",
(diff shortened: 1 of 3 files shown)


hexagonrecursion (approved):

FarkasKBence (changes requested):
Problem is that weights don't work well with multi-tile variants, leading to disjointed maps.

kép

You either need to give them separate ids or make their layouts compatible with each other.

Maleclypse:
> Problem is that weights don't work well with multi-tile variants, leading to disjointed maps.
> 
> kép
> 
> You either need to give them separate ids or make their layouts compatible with each other.

That was what I was worried would happen but didn't have availability to test at that moment.  Thanks!

Maleclypse:
> > #### Summary
> > None
> 
> I would have classified this as "content"

Anything that I work on that isn't big enough to go into a stable changelog I list as None. In part so I don't have to go back and try to remember how large a PR was when I'm working on the stable changelogs and now that we are trying to do an incremental weekly changelog to cut down on the amount of work the stable changelog is, I see the benefits of none on a weekly basis.

PR #65356, opened 2023-04-28
Rebalance magiclysm's mage armor spell to also scale with intelligence

zack-shoylev:
#### Summary
None

#### Purpose of change
Originally the spell was worded to scale with both spell level and intelligence, now it does so. See https://github.com/CleverRaven/Cataclysm-DDA/pull/64405

#### Describe the solution
Modifies the enchant JSON to use the proper arithmetic. The scaling can be additionally adjusted as desired.

Leaving as is (scaling only by spell level)

#### Testing
New JSON tested using debugging tools with a high-int low spell level character.

#### Additional context
None

### DIFF
--- a/data/mods/Magiclysm/items/ethereal_items.json
+++ b/data/mods/Magiclysm/items/ethereal_items.json
@@ -959,7 +959,7 @@
     "category": "armor",
     "copy-from": "armor_lightplate",
     "name": { "str_sp": "mage armor" },
-    "description": "A summoned magic armor.  It gets shinier and sturdier with each spell level.",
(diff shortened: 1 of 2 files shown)


GuardianDll:
I'd prefer if you use `math`, it will work much better, and math will replace the aritmetic eventually, but overall good

GuardianDll:
Made https://github.com/zack-shoylev/Cataclysm-DDA/pull/1, after merge it should be good to go

Fris0uman:
merged in #65508, authorship was preserved

PR #65360, opened 2023-04-28
Aquatic Slime

Jackody2:
creation of a new slime called water slime

#### Summary
Content "Creation of a new type of slime"

#### Purpose of change

expand the slime family
#### Describe the solution

create a new slime

none
#### Testing

Applied and tested locally

### DIFF
--- a/data/json/monstergroups/blob.json
+++ b/data/json/monstergroups/blob.json
@@ -19,7 +19,8 @@
       { "monster": "mon_blob_small", "weight": 200, "cost_multiplier": 0 },
       { "monster": "mon_blob_small", "weight": 100, "cost_multiplier": 0, "pack_size": [ 2, 12 ] },
       { "monster": "mon_blob", "weight": 695, "cost_multiplier": 3, "pack_size": [ 5, 10 ] },
-      { "monster": "mon_blob_large", "weight": 5, "cost_multiplier": 6, "pack_size": [ 5, 10 ] }
+      { "monster": "mon_blob_large", "weight": 5, "cost_multiplier": 6, "pack_size": [ 5, 10 ] },
+	  { "monster": "mon_aquatic_blob", "weight": 200, "cost_multiplier": 0, "pack_size": [5, 10] },
     ]
   }
 ]
--- a/data/json/monstergroups/blob.json
+++ b/data/json/monstergroups/blob.json
@@ -20,7 +20,7 @@
       { "monster": "mon_blob_small", "weight": 100, "cost_multiplier": 0, "pack_size": [ 2, 12 ] },
       { "monster": "mon_blob", "weight": 695, "cost_multiplier": 3, "pack_size": [ 5, 10 ] },
       { "monster": "mon_blob_large", "weight": 5, "cost_multiplier": 6, "pack_size": [ 5, 10 ] },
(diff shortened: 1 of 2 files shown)


Daved27hundred:
May I suggest changing the name to something like "aquatic slime" or "buoyant slime" ?

Jackody2:
yes it fits

data/json/monsters/slimes.json
```diff
@@ -231,5 +231,39 @@
...
+    "morale": 100,
+    "melee_skill": 5,
+    "melee_dice": 2,
+    "melee_dice_sides": 4,
+    "melee_damage": [ { "damage_type": "cut", "amount": 0 } ],
+    "bleed_rate": 30,
+    "harvest": "exempt",
+    "special_attacks": [ [ "FORMBLOB", 30 ] ],
+    "death_function": {
+      "message": "The %s splits in two!",
```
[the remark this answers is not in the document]
Jackody2: right, then it won't split in two

crypticcollaborator:
Is it intended that the aquatic blob is dramatically faster than the regular blob? 
If the aquatic blob has FORMBLOB and speed > 85, it can spawn additional small blobs over time, and these blobs won't be aquatic. That seems unintended.

Jackody2:
for this reason I removed the ability to split after death, otherwise they would have generated normal slimes

crypticcollaborator:
My understanding of the code is that formblob will make it generate slimes
before it does. It will also turn adjacent monsters into regular blobs as
well

Jackody2:
fixed

data/json/monsters/slimes.json
```diff
@@ -231,5 +231,33 @@
...
+    "aggression": 100,
+    "morale": 100,
+    "melee_skill": 6,
+    "melee_dice": 2,
+    "melee_dice_sides": 4,
+    "melee_damage": [ { "damage_type": "cut", "amount": 0 } ],
+    "bleed_rate": 30,
+    "harvest": "exempt",
+    "flags": [ "HEARS", "GOODHEARING", "NOHEAD", "POISON", "NO_BREATHE", "ACIDPROOF" ],
+    "armor": { "bash": 17, "cut": 9, "stab": 17, "acid": 12, "bullet": 7, "electric": 3 }
```
Fris0uman, line 261: should this have the same stats as a regular slime?

data/json/monstergroups/blob.json
```diff
@@ -19,7 +19,8 @@
       { "monster": "mon_blob_small", "weight": 200, "cost_multiplier": 0 },
       { "monster": "mon_blob_small", "weight": 100, "cost_multiplier": 0, "pack_size": [ 2, 12 ] },
       { "monster": "mon_blob", "weight": 695, "cost_multiplier": 3, "pack_size": [ 5, 10 ] },
-      { "monster": "mon_blob_large", "weight": 5, "cost_multiplier": 6, "pack_size": [ 5, 10 ] }
+      { "monster": "mon_blob_large", "weight": 5, "cost_multiplier": 6, "pack_size": [ 5, 10 ] },
+      { "monster": "mon_aquatic_blob", "weight": 200, "cost_multiplier": 0, "pack_size": [ 5, 10 ] }
```
Fris0uman, line 23: I don't see anything here to guarantee that they will spawn on water. As far as I can tell they'll just spawn among other slimes, so in slimes pits and labs pretty far from any pond.

PR #65363, opened 2023-04-28
Don't report json member errors if currently throwing an exception.

akrieger:
#### Summary
None

#### Purpose of change
The flexbuffer json pr introduced _two_ bugs with unvisited members. First it broke the actual check that members were visited. After that was fixed, people started reporting that it was too verbose and emitting pages of errors when an earlier issue is actually at fault. This is because when stack unwinding happens from an exception being thrown, the remaining eg. JsonObjects in a JsonArray will destruct without having been iterated.

#### Describe the solution
The old json implementation had a check for `!std::uncaught_exceptions()` to be true to report errors. So port that to the flexbuffer json implementation.

#### Testing
:eyes:

### DIFF
--- a/src/flexbuffer_json.cpp
+++ b/src/flexbuffer_json.cpp
@@ -240,7 +240,7 @@ bool JsonValue::read( std::string &s, bool throw_on_error ) const
 
 void JsonObject::report_unvisited() const
 {
-    if( !visited_fields_bitset_.all() ) {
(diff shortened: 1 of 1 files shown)


akrieger:
https://github.com/CleverRaven/Cataclysm-DDA/blob/094c9e46cef9d584532ed3b391d024b40925510b/src/json.cpp#L144 the check in original json.cpp.

PR #65366, opened 2023-04-29
Make some locations more rare

NetSysFire:
#### Summary
Balance "Make the following locations more rare: Hazardous waste sarcophagus, military outpost, military helipad, military bunker, LMOE shelter, convention center, bastion fort, cathedral"

#### Purpose of change

1. Convention centers are a new addition and while testing some changes of mine, I noticed them being weirdly common. There are not many in Massachusetts anyways it appears. Maybe a dozen in total.
2. New England only has a handful of bastion forts. See https://en.wikipedia.org/wiki/List_of_bastion_forts#United_States
3. Massachusetts has a ton of *former* military locations, but current ones, aka the non-defunct ones we have ingame, are quite rare: https://en.wikipedia.org/wiki/List_of_military_installations_in_Massachusetts
4. Ironically the only thing I can find about LMOE shelters in New England is this game. Fully built ones, since the apocalypse approached quite quickly, have to be rare.
5. There are maybe 6 cathedrals in total in Massachusetts. https://en.wikipedia.org/wiki/Category:Cathedrals_in_Massachusetts

#### Describe the solution

n/a

#### Testing

Applied locally. Looks fine so far.


PatrikLundell:
If I had or planned an LMOE shelter I certainly would attempt to avoid it being found via the Internet. Part of the corresponding mindset seems to be to protect yourself by not being found by desperate gun toting people trying to steal your survival hoard (or simply take over the shelter) after the apocalypse. Thus, I wouldn't expect an Internet search to give a reliable result.

BlizzDA:
The lore regarding LMOE shelters in 'larger' amounts is contributed to lore up to the Cataclysm, months before anything really hit the fan people were prepping. But as has been said, the point is also not reveal where you have your LMOE. A better change may be to have some LMOE that are mid-construction, or otherwise not quite furnished. But that may be for another PR.

BlizzDA:
> The lore regarding LMOE shelters in 'larger' amounts is contributed to lore up to the Cataclysm, months before anything really hit the fan people were prepping. But as has been said, the point is also not reveal where you have your LMOE. A better change may be to have some LMOE that are mid-construction, or otherwise not quite furnished. But that may be for another PR.

The other spawn changes are understandable, based on IRL numbers.

NetSysFire:
I do get your points. In my opinion there were waaaay too many LMOE shelters. If you have one every couple of overmap tiles, it really seems odd. What would your suggested value be?

crypticcollaborator:
Personally, I don't think LMOE shelters are necessarily too common. The thing that would probably help is to make them less visible somehow, since these shelters are meant to be difficult to find.

BlizzDA:
> I do get your points. In my opinion there were waaaay too many LMOE shelters. If you have one every couple of overmap tiles, it really seems odd. What would your suggested value be?
Thinking about it, I'm not sure myself actually. I do agree that they're presently too common checking on it in-game, even with the lore that people got really into prepping. They can spawn all over the overmap so if you're in a corner they can get rather grouped together. Since I can't think of a value I feel like can't really object there.

Vaskritaya:
> Personally, I don't think LMOE shelters are necessarily too common. The thing that would probably help is to make them less visible somehow, since these shelters are meant to be difficult to find.

I believe there's already specials in the game that just appear as "forest" in appearance and name. Seems like it would make a lot of sense for LMOE shelters, since they're made to just look like wooded knolls or whatever.

sadenar:
I believe that established lore does have some considerable buildup of shelters and small military installations in NE due to geopolitical reasons/XEDRA reasons no?
I do agree that there definitely shouldn't be a military base next to every town, but perhaps sticking to IRL numbers too conscientiously on this might be a bad move?

fairyarmadillo:
LMOEs are ridiculously common in the game to a point that badly stretches believability.

I would expect most preppers to be far less prepared. A multi-room complex hidden in the woods is quite an unusual thing in any circumstances, and even if there are a substantial number of preppers (I don't agree that there are), most of them wouldn't have the knowledge or resources to construct such a thing. We should see more sheds with guns and canned food and fewer endgame cataclysm bases.

I'd say for this PR, drastically reduce them and if someone wants to add a few scaled back ones, they can do that.

fairyarmadillo:
https://youtu.be/9uGRLKPHwIg

Here's an example of what a real LMOE shelter looks like. AFAIK LMOE is a CDDA term, it's not used in real life, but as you can see it's a pretty tiny structure.

NetSysFire:
>I believe there's already specials in the game that just appear as "forest" in appearance and name.

Correct. These are e.g lab entrances or black dragon lairs in magiclysm.

We have an either or here: LMOE shelters are frankly not that special. Couple underground rooms with loot and some machinery. Them being rare *and* invisible does not fit. For labs and the mentioned dragon lairs it makes sense because there is some insane loot inside. A single LMOE shelter appearing in 1/5 of all overmaps sounds reasonable to me. Leaving their spawn rate and making them invisible sounds fun, too but I prefer the former.

Nebnis:
You specifically mentioned Massachusetts, but i was under the impression that this game was generally located around the entire new england. Wouldn't that make the military locations way more common? As for LMOE, i feel like 1/5 is a bit too harsh, specially if you making them unique as well, but i definitely agree they are wayyyy too common as right now too. If you going to make them unique then there is no need to make them too rare.

NetSysFire:
I think you confuse the `UNIQUE` flag with `GLOBALLY_UNIQUE`. The latter means it can really only spawn one time, which is the case for the NPC occupied ones. The former means there can only be one occurance per overmap. And as you know, theres plenty of overmaps and every one has a 20% chance of containing a single LMOE shelter.

JReese3:
If you want more inspiration for LMOE's:

Rackstar under backyard bunker.
https://www.youtube.com/watch?v=ZQy89tZ-mRU

Especially take note the mad man using a guitar flamethrower and custom arm mounted rocket launcher in there...

Random Atlas Shelter (most common style of LMOE) based out of CA.
https://www.youtube.com/watch?v=rfMK7sutuNc

Another Atlas...:
https://www.youtube.com/watch?v=-Yg590JzYLo

MEMA!!!:
https://www.youtube.com/watch?v=0Dha50PgrbM

PR #65234, opened 2023-04-21
Rebalanced generic zombie drops and bugout bags

worm-girl:
#### Summary
Balance "Adjust the drop rates for bugout bags and basic zombies"

#### Purpose of change

Zombie loot has some real oddities going on, and presently, even vanilla zombies are so loaded that you can easily live off the unspoiled food and drinks they carry, to say nothing of the obscene trading profits you can make pilfering cigarettes and hard drugs off of them.

This is a multi-tiered problem that isn't unique to just zombies, but the zombie_default monsterdrops is as good of a place as any to start making changes.

#### Describe the solution

* Reduced the probability of union suits on male zombies from 10 to 1. The union suit is considered outmoded to the point of being comical, and should probably be much rarer, however probability requires an integer and it is not currently possible to reduce this to a fraction of a percent.

* Reduced the probability of the tuxedo to 1. This is also too high, as these are rare pieces of formalwear only used at specific events, but again it is not currently possible to make them less common.

* The thawb, cassock, and haori are likewise far too common, but there's not an easy way to make them more rare.

* Reduced the probability of finding stockings, tights, or garter belts on a zombie which was wearing pants to 1, 5, and 1 respectively. Wearing these items under pants is a conceivable possibility (and in some cases tights under pants is a decent way to keep warm), but they were much too common.

* Reduced the probability of the wedding dress and veil to 1. These items are generally something you wear once in your entire life, and even accounting for feral wives making bizarre sartorial decisions, there were way too many of them. **edit**: I went ahead and just removed these from the drop table. If we need zombrides for a specific location or something, that can probably be a unique enemy.

* Reduced long_glove_white's probability from 20 to 1. These are also an exceedingly rare piece of clothing, and are mostly seen as costume pieces.

* Reduced the probability of the biking distribution (helmet_bike and clothing_biker) to 3 from 10. Plenty of people ride two-wheeled vehicles, but most of them are not decked out for riding 24/7. This will make their protective gear a bit harder to find, but it should still be quite readily available in the early game.

* Reduced the probability of finding a vibrator in a small bag from 5 to 1. Most people do not carry sex toys around with them.

[...]

PR #65372, opened 2023-04-29
Additional ambient guns

bombasticSlacks:
#### Summary
Content "Paramilitary Zombies and everyday carry"

#### Purpose of change
After #65234 guns were taken out of the game, this makes sense, not all those guns should be in bags so I've moved them out of bags.

#### Describe the solution
Standard zombies now have a chance to be everyday carrying, they also have a chance to have scooped one of those ubiquitous "dropped guns" everyone is always talking about. These are from the new "everyday carry" spawn group, loaded guns that don't have spare mags. I know some people every day carry an extra mag but it makes this PR way more complicated and isn't super important.

Added paramilitary zombies. These are American citizens who went out geared once SHTF. Potentially wearing light military armour and carrying a long rifle and sidearm with mags (similar to how cops determine carry). They have similar spawn rates to survivor zombies in similar locations. This is a great spot to stuff impractical tacticool garbage and make subtle jabs at operator types (when are we getting wraparound oakleys)

Also updated a bugout bag related comment, someone added sanitation pads to bugouts but my OG comment calls that loot group "useless", sanitation products aren't useless so I changed the comment. 

*Someone* should probably go through and cook up some additional operator guns. Customized ARs for different "operations", some other guns popular with operators etc.

#### Testing
Everyone works in game

#### Additional context
image


fairyarmadillo:
Forgive me if my math is off here, but isn't this adding four times as many guns as were removed?

bombasticSlacks:
A bugout bag isn't concealed carry. This adds the potential for concealed carry using the numbers from your PR.

I'm not having this conversation with you again, there are more guns than people in the United States, if you want less guns drop your spawn rates.

fairyarmadillo:
i don't want less guns, I want the amount that there are in real life. If you believe that's what we've got here, then OK. I wasn't trying to start an argument about it.

data/json/monsters/zed_survivor.json
```diff
@@ -32,6 +32,18 @@
     "flags": [ "SEES", "HEARS", "WARM", "BASHES", "GROUP_BASH", "POISON", "NO_BREATHE", "REVIVES", "PUSH_MON", "FILTHY" ],
     "armor": { "bash": 7, "cut": 7, "stab": 5, "acid": 4, "bullet": 6, "electric": 2 }
   },
+  {
+    "id": "mon_zombie_paramilitary",
+    "type": "MONSTER",
+    "name": { "str": "paramilitary zombie" },
+    "description": "Wearing military surplus gear including an armoured vest with patches of local militias, cartoon women, obscene statements, or local political groups.  This person definitely thought they were ready for the apocalypse.",
```
TheShadowFerret, line 39:
```suggestion
    "description": "Wearing military surplus gear including an armored vest with patches of local militias, cartoon women, obscene statements, or local political groups.  This person definitely thought they were ready for the apocalypse.",
```

Daved27hundred:
what makes a paramilitary zombie different than a survivor zombie?

Nebnis:
> Forgive me if my math is off here, but isn't this adding twice times as many guns as were removed from the standard zombie loot pool? Previously you had basically a 1% chance to find a gun on a vanilla zombie. This was cut to .5% which is a lot more common than it sounds. But now there's an additional 1% chance for 1.5% total?
> 
> In the bugout bag PR, I put up statistics about how many people are likely to be conceal carrying in public on a daily basis, then pointed out that while about 1/100 people being armed might be roughly within expectations if you polled people outdoors, most zombies would have died at home, and their guns would more likely be left there (where they do currently spawn), not on ther persons.

Bomba still raised a good point that those guns shouldn't be phasing out of reality and all the riots happening before the cataclysm would pretty much make a considerable amount of people to be carrying guns. Hell if anything it should be MUCH higher than now. You were comparing data with time of peace instead.

fairyarmadillo:
> Bomba still raised a good point that those guns shouldn't be phasing out of reality and all the riots happening before the cataclysm would pretty much make a considerable amount of people to be carrying guns. Hell if anything it should be MUCH higher than now since you were comparing data with time of peace instead. I don't own guns but i would 100% get one if bloodlust crazy riots were happening around here.

The matter doesn't seem to be open for discussion, so there's not really any point trying to convince me.

Ruminator24:
I might be only a plebian, but is it possible to add stuff like "tacticool" items, such as impractically unsound boots, or cheap replica MOLLE copies and such, stuff that rips much easier than the actual stuff, much like how you can find 3 types of Katanas; The TRUE AND HONEST variant, the variant that's from Japan, but isn't as high quality as the original, and then the 10 dollar souvenir shop variant that snaps after one round of combat. So I'm wondering if the same could be applied to these paramilitary zombies, as well as tiers for them, with regular paramilitary zoms having the cheapest and worst equipment, very rarely having "quality" equipment, and "elite" paramilitary goons having better chances of higher quality equipment. It might also be a good idea to have "bad" attachments on these guns, or poorly made magazines which might help reduce the problem of there being "too" many guns everywhere. Of course, I'm just some dumbass who doesn't know how to code, so I don't know how impossible this is, but this might be a good compromise?

bombasticSlacks:
> might be a good compromise

No compromise needed. Players like loot, reality has guns, people who want less loot can modify world settings.

Generally though your suggestion is a good idea just because those things exist.  We could have multiple variants of some things to represent worse options. Unlike with our 6-8 variants of every melee weapon though it really needs a meaningful place or I'm considering it needless bloat. Pockets are a bad idea IMO because not much diferentiates a good pocket from a bad in this game, especially since the pockets end up obfuscated on the vests. Further these paramilitary zombies are high statted which implies they trained with their gear, so no dogshit amazon red dots cause if you actually drill with a gun you would start cutting your knockoff garbage.

We could have other zombies with cosplay tier gear, or put it in the standard zombie pool. 

Also after typing this all out, this current PR has the zombies with the worst vest in the game, low chance for head armor, and no mods for their guns so you are also theory crafting for a problem that doesn't exist yet.

Ruminator24:
Dang, that is true, it might be bad for there to be endless variants, although I did have the idea of there being different descriptions to show off the lower quality equipment, like the description mentioning "The fabric feels quite cheap and easy to rip" or "On closer inspection, you realize that the company's logo has been mispelled". However, I think I might be thinking of the wrong idea of how well equipped these paramilitary guys are, since I think your idea is more in line with them at least decently trained like those that might be found in self organized militias, while I'm thinking them as if they're just tacticool guys who browse army forums and think of themselves as tactical experts with the "best" equipment, which is just the front page stuff from amazon. 
In that case, I do apologize for the confusion.

Ruminator24:
I think it could be possible for me to try, perhaps I could try to do that, but I wonder if it's massively difficult to have one item that can have multiple descriptions, like how according to the hitchhiker's guide there's three different versions of the katana, each corresponding to a "tier" of said katana, instead of perhaps just having one katana with three... different subclasses? I think that's the correct word for it. But yeah, there's a good amount of cheap stuff that one could reasonably attach to a gun, from the $5 laser pointers to the top tier NV military stuff that only elite operators have, might be even good for me to try.

PR #65373, opened 2023-04-29
Fix calculation for inserting into nested containers

Rewryte:
#### Summary
Bugfixes "Fix calculation for inserting into nested containers"

#### Purpose of change
When trying to determine if an object will fit into a nested container, the function "parents_can_contain_recursive" is used.  It operates by checking if each of the parent containers "can_contain" an item.  This has 2 issues:

- Parent containers do not respect weight and volume multipliers of child containers.  This was reproduced by attempting to insert a pot helmet into a greater storage ring (Magiclysm) inside an armbag pouch.  The multipliers of the storage ring should allow the pot helmet to fit, but it is not the case
- Rigid containers ignore the weight capacity of parents.  The Insert menu of a metal tank (2L) inside a gallon-sized zipper bag would show a steel chain as an option, but this actually overloads the zipper bag and attempting it will result in an "aborting" message.

For objects with charges, the sister function "max_charges_by_parent_recursive" is used which exhibit similar issues.  Inserting oranges into a greater dimensional bag (Magiclysm) nested in an armband pouch, only one orange can be inserted at a time up to a maximum of 2.  The correct number of oranges that should fit is 6 and they should all be inserted on the first attempt.

#### Describe the solution
Made new calculations for the functions "parents_can_contain_recursive" and "max_charges_by_parent_recursive" which keeps track of weight, volume and length changes in items as they are modified by the weight / volume multiplier of each nested container.  It will continue to check weight capacity even after encountering a rigid container.  Code is commented for clarity.

#### Testing
Tests were performed by attempting to reproduce the aforementioned issues using their respective test scenarios as described above.  None of the issues were able to be reproduced.  Also tested various other use cases and so far no anomalies were encountered.

#### Additional context
These fixes target both the list of items displayed on the insert menu, and the actual act of inserting the item into the nested container.  The "pick all items" action do not use these functions and should show results consistent with older behavior.

### DIFF
--- a/src/item_location.cpp
+++ b/src/item_location.cpp
@@ -865,6 +865,7 @@ bool item_location::parents_can_contain_recursive( item *it ) const
     //item_location class cannot return current pocket so use first pocket for innermost container
     //Only used for weight and volume multipliers
     const item_pocket *current_pocket = get_item()->get_all_standard_pockets().front();
+    const pocket_data *current_pocket_data = current_pocket->get_pocket_data();
 
     //Repeat until top-most container reached
     while( current_location.has_parent() ) {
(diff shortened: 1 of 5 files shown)


src/item_location.cpp
```diff
@@ -877,13 +900,40 @@ int item_location::max_charges_by_parent_recursive( const item &it ) const
...
+        weight_multiplier = weight_multiplier * current_pocket->data->weight_multiplier;
+        volume_multiplier = volume_multiplier * current_pocket->data->volume_multiplier;
+        //Inserting into rigid pockets will not affect parent volume so stop keeping track
+        if( current_pocket->rigid() ) {
+            volume_multiplier = 0.0f;
+        };
+
+        //Calculate effective remaining weight and volume for current parent
+        units::mass temp_weight = weight_capacity() / weight_multiplier;
+        units::volume temp_volume = volume_capacity() / volume_multiplier;
```
irwiss, line 924: potentially you divide by 0 due to line 919 and there's items with weight multiplier set to 0 (ebook pockets, magiclysm various bags of holdings), float division won't abort (usually) but you'll get a nan or infinity

src/item_pocket.h
```diff
@@ -399,11 +399,11 @@ class item_pocket
 
         // should the name of this pocket be used as a description
         bool name_as_description = false; // NOLINT(cata-serialize)
+        const pocket_data *data = nullptr; // NOLINT(cata-serialize)
```
irwiss, line 402: there's a public get_pocket_data for const accessor

src/item_location.cpp
Rewryte: Thanks, can't believe I missed this

src/item_pocket.h
Rewryte: Thanks

PR #65379, opened 2023-04-29
math_parser: add support for JSON-defined functions

andrei8l:
#### Summary
None

#### Purpose of change
Add support for JSON defined math functions so that calculations can be easily reused. There is already a potential use case in the dew point calculation that currently uses a global EOC to update the value.

~~* Depends on #65307 and the first commit is that PR squashed~~

Math functions are defined like this:
```JSON
  {
    "type": "jmath_function",
    "id": "my_math_function",
    "num_args": 2,
    "return": "_0 * 2 + _1"
  },
```
_where `_0`, `_1`, etc are positional parameters (similar to `$1`, `$2`, etc in shell scripts)_
and can be used like regular math functions
```JSON
  {
    "id": "EOC_do_something",
    "effect": [ { "math": [ "secret_value", "=", "my_math_function( u_pain(), 500)" ] } ]
  },
```
_so `_0` takes the value of `u_pain()` and `_1` takes the value `500` inside `my_math_function()`_

Functions can be defined and used in any order thanks to deferred parsing.
Function composition with `jmath_function` is also supported, of course.

#### Describe the solution
Defer parsing math to finalization step.
Add the bits and bobs needed to load and use JSON-defined math functions.

TODO:
- [x] documentation: just copy-pasted the text above

N/A

#### Testing
Test unit

#### Additional context
N/A

### DIFF
--- a/data/mods/TEST_DATA/EOC.json
+++ b/data/mods/TEST_DATA/EOC.json
@@ -73,6 +73,23 @@
   },
   {
     "type": "effect_on_condition",
+    "id": "EOC_jmath_test",
+    "effect": [ { "math": [ "blorgy", "=", "jmath_test_blorg(1, 2)" ] } ]
+  },
+  {
+    "type": "jmath_function",
+    "id": "jmath_test_blorg",
+    "num_args": 2,
+    "return": "_0 + jmath_test_blorg2( _1 )"
+  },
+  {
+    "type": "jmath_function",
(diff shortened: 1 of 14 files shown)


andrei8l:
Thanks for merging.

zwa73:
Maybe you'll like this
https://github.com/zwa73/CDDA-EocScript
I made a compiler that compiles simple javascript code to eoc

PR #65380, opened 2023-04-29
Changed “Special Operator” to “Delta Operator”

halosammy:
#### Summary
Content "Renamed Special Operator"

#### Purpose of change

The “Special Operator” profession was conceived when it was the only one of its kind. Now that Navy SEALs are a distinct profession, it seemed best to set Operators by specifying that it’s Delta Force. Especially considering that the term of “operator” specifically applies to Delta, even though it originated from the Green Berets and became a colloquialism for all special forces in the 21st century.

#### Describe the solution

Changed the name.

Remove the SEAL profession if you want Special Operator to stand in for the whole spectrum of special forces.

#### Testing

N/A

### DIFF
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -1518,8 +1518,8 @@
   {
     "type": "profession",
     "id": "specops",
-    "name": "Special Operator",
-    "description": "You were the best of the best, the military's finest.  That's why you're still alive, even after all your comrades fell to the undead.  As far as you can tell, military command abandoned you in this hellhole when you missed the emergency evac.",
+    "name": "Delta Operator",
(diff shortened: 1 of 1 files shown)


Maleclypse:
I figured this would be easy to confirm and approve but apparently your premise is incorrect and this article when I searched for origin came with sources.  https://coffeeordie.com/origin-of-operator/

halosammy:
> I figured this would be easy to confirm and approve but apparently your premise is incorrect and this article when I searched for origin came with sources. https://coffeeordie.com/origin-of-operator/

It is? Oh well. Would you say changes aren’t applicable, then?

PR #57706, opened 2022-05-14
npctrade: rework price adjustment formula

andrei8l:
#### Summary
None

#### Purpose of change
Players can make a profit by trading a spoon for a spoon if their social skill is high enough.

Fixes: #47569

#### Describe the solution
Change the price formula so that the NPC always trades at an advantage. Specifically:
1. the NPC's price is `price * ( 1 + 0.25 * adjust )`
2. the player's price is `price * ( 1 - 0.25 / adjust )`

Where `adjust` is determined by the difference in social skill (capped at 5 points difference):
1. If difference is positive: `adjust = 0.045 * diff ^ 2 - 0.025 * diff + 1`. This is just a quadratic best fit of the old hardcoded values
2. If difference is negative: 1/`adjust`

In English, that means:
1. An NPC with no skill advantage sells at 1.25 and buys at 0.75
2. An NPC with 5 points skill advantage sells at 1.5 and buys at 0.5
4. An NPC with 5 points skill disadvantage sells at 1.125 and buys at 0.875

The intelligence stat has the same impact as before (doubles the adjustment at 20 stat point difference).

Additionally, I've incorporated the ideas from #57660 with my own tweaks:
1. Filthy items trade for 0.8 of their values. Non-faction squeamish NPCs refuse to buy filthy items entirely
2. Damage reduces the price by up to 80%

#### Describe alternatives you've considered
#57660: I've been sitting on this patch for some time and I wanted a different formula than used there. I did incorporate some of stubkan's ideas and added him as a co-author though. It doesn't matter in the end which PR gets merged as long as the changes are reasonable.

Other mitigation strategies:
1. Using the lowest/highest price or always the NPC's price if both parties have the item: this can be trivially exploited by trading all the stock in the cheaper direction then back in the other for profit
3. Using a price memory and refusing to buy below last sold price: if price memory is short enough, this is just 1 with a delay. If price memory is long enough, it encourages players to avoid trading until they level up their social skill.

#### Testing
Verify multipliers using a hammer (nominal price of 1$)

#### Additional context
N/A

PR #65389, opened 2023-04-30
npctrade: un-hardcode trade price formulas

andrei8l:
#### Summary
None

#### Purpose of change
* Move the trading price formulas from #57706 and #58119 to JSON, using jmath functions
~~* Depends on #65379 and will not compile without it~~

#### Describe the solution
Move the calculations to `jmath_function`s.

Instead of quadratic and linear best fits with branches for skill and int adjustments respectively, use a 6th-degree polynomial and an exponential fit.

TODO:
- [ ] document, somewhere

The polynomial and exponential functions are implementation details. The important bits are that the adjustment doubles/halves at 5 points of skill difference and again at 20 points of intelligence difference and there's a smooth curve in between, as described in #57706 and #58119. Instead of scary functions, I could have used:
- quadratic for skill and linear for intelligence (current hardcoded formulas): these need ternary operators for symmetry and I don't want to implement them at the moment (or ever)
- cubic for skill: the error is too high near the edges
- 4th or 5th degree polynomial: just as scary as a 6th degree but with worse fit so why not 6th?

#### Testing
Prices should be very similar before and after. Check at the extremes of skill/int difference and middle values too. They won't be exactly the same due to the different fits, but should be within a couple of cents for almost all items.

The `[trade]` test suite should still pass.

#### Additional context
The fits were calculated by WolframAlpha:
[skill adjustment](https://www.wolframalpha.com/input?i=polynomial+fit+%7B%7B-5%2C1%2F2%7D%2C%7B-4%2C1%2F1.65%7D%2C%7B-3%2C1%2F1.3%7D%2C%7B-2%2C1%2F1.15%7D%2C%7B-1%2C+1%2F1.05%7D%2C+%7B0%2C1%7D%2C%7B1%2C1.05%7D%2C%7B2%2C1.15%7D%2C%7B3%2C1.3%7D%2C%7B4%2C1.65%7D%2C%7B5%2C2%7D%7D+degree+6)
[int adjustment](https://www.wolframalpha.com/input?i=exponential+fit&assumption=%7B%22C%22%2C+%22exponential+fit%22%7D+-%3E+%7B%22Calculator%22%7D&assumption=%7B%22F%22%2C+%22ExponentialFitCalculator%22%2C+%22data2%22%7D+-%3E%22%7B%7B-20%2C+0.5%7D%2C+%7B0%2C+1%7D%2C+%7B20%2C+2%7D%7D%22)

### DIFF
--- a/data/json/npctrade.json
+++ b/data/json/npctrade.json
@@ -1,0 +1,46 @@
+[
+  {
+    "type": "jmath_function",
+    "id": "skill_adjustment",
+    "num_args": 1,
+    "//": "6th-degree polynomial fit for {{-5,1/2},{-4,1/1.65},{-3,1/1.3},{-2,1/1.15},{-1,1/1.05},{0,1},{1,1.05},{2,1.15},{3,1.3},{4,1.65},{5,2}}",
+    "return": "1.00342 + 0.0361699*_0 - 0.00316488*_0^2 + 0.00770736*_0^3 + 0.000952084*_0^4 - 0.000125351*_0^5 - 0.0000172177*_0^6"
+  },
+  {
+    "type": "jmath_function",
+    "id": "skill_adjustment_clamped",
+    "num_args": 1,
+    "return": "skill_adjustment(clamp(_0, -5, 5))"
+  },
+  {
+    "type": "jmath_function",
+    "id": "int_adjustment",
+    "num_args": 1,
+    "//": "exponential fit for {{-20, 0.5}, {0, 1}, {20, 2}}",
+    "return": "e^(0.0346574*clamp(_0, -19, 19))"
+  },
+  {
+    "type": "jmath_function",
+    "id": "npctrade_u_sell_adj",
+    "num_args": 1,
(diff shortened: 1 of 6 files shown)


andrei8l:
> imo this change is a bit weird; it takes some existing data points, moves them into a curve and into json, for someone to adjust but for what reason?

Same reason we push any numbers into JSON: so it's easier to modify by people who don't write code. Mods will be able to change the formulas right away, and with a bit of glue we can have individual traders have their own pricing schemes in addition to price rules.

> Practically any "useful" change in barter pricing will need ...  like account for faction-wide inventory, friendly/neighbor faction inventories, time the item is in/out of stock, world age etc.

And this prevents that how? We have faction price rules already that can cover things like that. They are not very dynamic at the moment but that will improve with jmath. This change is unrelated to that functionality.

> This change means instead of improving the barter prices you'll now be wasting time shoving all of this stuff into json instead first

Not really sure what you mean here. It's my time to waste in any case.

PR #65392, opened 2023-04-30
Horse Blinders

Geevies:
#### Summary
Content "Added 'Horse Blinders', an attachable for your mount that lets them get a bit closer without getting scared"

#### Purpose of change

The idea here is to make mounted combat more doable by letting the player ride up closer to the enemy. As mentioned in https://github.com/CleverRaven/Cataclysm-DDA/pull/64977, horses are skittish, so you simply won't be able to take non-warhorses up to a zombie for a sword-fight.

Blinders reduce the "scare" range from 10 tiles to 7, which is still a respectable risk-zone, but enable mounted archers to take potshots at zombies. Any closer and they risk getting thrown off still. Additionally, for non-trained mounts, the "no budge" range reduces from 15 to 12 tiles.

Blinders should never remove horses being scared, as the horse still needs to be able to see in front of them to move without tripping on a rock and exploding into a billion lego pieces. The reduced range is simply an abstraction for the comfort of peripherals being blocked, since we're dealing with a 360 degree view here.

#### Describe the solution

- Added a 'Horse Blinders' item, which can be attached to a mount provided you have at least one point in the survival skill.
- The blinders effect lets you take an untrained mount to a minimum of 12 (down from 15) tiles from an enemy before it starts refusing to budge.
- The blinders effect lets reduces the range at which a mount starts rolling for being scared to 7 (down from 10) tiles.
- Blinders can be crafted, the recipe being learned at the same place horse tacks are learned, or from having 3 tailoring and 2 fabrication at character creation.
- Additionally, blinders have been added to the farm riding supply loot pool.

- Not implementing blinders at all, but where's the fun in that?
- Making blinders make the mount not scared at all, at the cost of slowed movement. Discarded because a slow and blind mount sounds like a terrible idea.

#### Testing

- Spawned in a mount and blinders.
- Tested that attaching and removing blinders works.
- Tested that saving and loading the blinders works.
- Tested that the range for scaring and budging works.
- Tested that crafting works.

### DIFF
--- a/data/json/effects.json
+++ b/data/json/effects.json
@@ -215,6 +215,12 @@
   },
   {
     "type": "effect_type",
+    "id": "monster_blinders",
+    "name": [ "Has Blinders" ],
+    "desc": [ "AI tag used for critters wearing blinders.  This is a bug if you have it." ]
+  },
+  {
+    "type": "effect_type",
     "id": "monster_saddled",
     "name": [ "Has Saddle" ],
     "desc": [ "AI tag used for critters wearing a saddle.  This is a bug if you have it." ]
--- a/data/json/flags.json
+++ b/data/json/flags.json
@@ -416,6 +416,10 @@
     "info": "This gear is equipped with an accurate hygrometer (which is used to measure humidity)."
   },
   {
+    "id": "BLINDERS",
+    "type": "json_flag"
+  },
+  {
     "id": "BURNOUT",
     "type": "json_flag",
     "//": "You can visually inspect how much it is burned out (candle, torch)."
--- a/data/json/itemgroups/Locations_MapExtras/locations_commercial.json
+++ b/data/json/itemgroups/Locations_MapExtras/locations_commercial.json
@@ -2184,7 +2184,7 @@
     "id": "farm_supply_riding_gear",
     "type": "item_group",
     "subtype": "collection",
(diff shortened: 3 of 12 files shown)


Maleclypse:
I'm not sure that blinders should work on mounts that are combat trained.   When I look it up blinders were used on horses that engaged in support work (pulling wagons to the back of a battlefield) not combat work from what I can tell.

Geevies:
> I'm not sure that blinders should work on mounts that are combat trained. When I look it up blinders were used on horses that engaged in support work (pulling wagons to the back of a battlefield) not combat work from what I can tell.

I'm not 100% on this either. When I look up horse barding, a bunch of them have little eye-cups as well, but I reckon those are mostly for eye-protection rather than hiding a scary thing from the horse.

In the context of horse combat though, I think blinders are being used in desperation rather than function in this case, where you're pushing a normal modern day horse to the limit to gain an edge in an apocalypse, and you're using whatever tools you have to keep 'em from freaking out.

I'm open to any ideas / thoughts, though!

Maleclypse:
Would it be easy to make the blinder's effect and the combat mount effect not apply at the same time?  That way you can have one effect or the other basically.

kevingranade:
Blinders don't block all vision, so this wouldn't really do anything.

PR #65408, opened 2023-05-01
New animal and monster: the salamander

Jackody2:
introduction of a new amphibian, the salamander, and its mutated and zombified forms

#### Summary
Content "introduction of a new amphibian, the salamander, and its mutated and zombified forms"

#### Purpose of change

introduction of a new amphibian, the salamander
#### Describe the solution

edit the JSON files to make that change

none

#### Testing

Tested locally with the debug menu


data/json/monsters/reptile_amphibian.json
```diff
@@ -390,5 +390,103 @@
...
+  {
+    "id": "mon_gecko_mega",
+    "type": "MONSTER",
+    "name": { "str": "colossal gecko" },
+    "description": "A gecko bloated to titanic size. Run.",
+    "default_faction": "mutant",
+    "bodytype": "lizard",
+    "species": [ "REPTILE" ],
+    "volume": "92500 ml",
+    "weight": "120 kg",
```
Venera3, line 468: What is it made of? Fleshy things pretty much alwas end up at 1kg/l, so this has 30kg extra at least.

data/json/monsters/reptile_amphibian.json
```diff
@@ -390,5 +390,103 @@
...
+    "hp": 800,
```
Venera3, line 469: Way too high for 120 kg, especially for a living thing - this is half-ton hulk levels of HP.

data/json/monsters/reptile_amphibian.json
```diff
@@ -390,5 +390,103 @@
...
+    "speed": 100,
+    "material": [ "flesh" ],
+    "symbol": "G",
+    "color": "green",
+    "aggression": 5,
+    "morale": 100,
```
Venera3, line 475: Living things shouldn't have 100 morale. They are not automatons, if their prey proves too much they should scamper off.

data/json/monsters/reptile_amphibian.json
```diff
@@ -390,5 +390,103 @@
...
+    "aggro_character": false,
+    "melee_skill": 4,
+    "melee_dice": 6,
+    "melee_dice_sides": 10,
```
Venera3, line 479: Consider special attack instead of bump damage - a bite and maybe a tailslap instead of just generic hits.

data/json/monsters/reptile_amphibian.json
```diff
@@ -390,5 +390,103 @@
...
+    "melee_damage": [ { "damage_type": "cut", "amount": 1 } ],
+    "dodge": 3,
+    "vision_night": 10,
+    "regenerates": 4,
```
Venera3, line 483: Why the regen? It's not alluded to in any way in the description, and just because it's a lizard its wounds shouldn't close up instantly.

data/json/monsters/reptile_amphibian.json
```diff
@@ -390,5 +390,103 @@
...
+    "harvest": "mutant_animal_large_noskin",
+    "dissect": "dissect_batrachian_sample_huge",
```
Venera3, line 485: Lizard samples, surely.

data/json/monsters/reptile_amphibian.json
```diff
@@ -390,5 +390,103 @@
...
+    "path_settings": { "max_dist": 5 },
+    "zombify_into": "mon_evoluted_z_gecko",
```
Venera3, line 487: Giants mutant zombify preferentially into cocoons. I'm not 100% opposed to hand-picked zombification routes, but they need a bit more of a catch than "what if slower and angrier X".

data/json/monsters/zed_misc.json
```diff
@@ -1644,5 +1644,95 @@
...
+      "PUSH_MON"
+    ],
+    "upgrades": { "half_life": 34, "into": "mon_frog_mother" },
+    "armor": { "bash": 8, "cut": 2, "bullet": 1, "electric": 2 }
+  },
+  {
+    "id": "mon_evoluted_z_gecko",
+    "type": "MONSTER",
+    "name": { "str": "evoluted zombie gecko" },
+    "description": "a giant gecko with its mouth open in an ominous way and with its lifeless eyes stares into void, pray.",
```
Venera3, line 1696: The descriptions in general do a lot of prescribing ("pray, run, despair") instead of describing. I'd go lighter on the former.

Venera3 (changes requested):
There are no wild geckos in New England(in fact at a cursory glance the only established wild lizard is a skink). You could have them in a specific TCL testing enclosure or similar labby location, or add them to Tropicata if they fit better, but they shouldn't be vanilla wilderness spawns.

data/json/monsters/reptile_amphibian.json
Jackody2: what do you mean?

fairyarmadillo:
Salamanders should probably be giving batrachian samples, not lizard. Salamanders are amphibians despite sharing a similar body shape with lizards.

The little one also shouldn't have a tail whip attack, and I'm not sure they should be biting at all. All the salamander species in New England are very small and harmless to humans, even when provoked. The only threat they pose is that some of them are poisonous to eat.

Karol1223:
> In fact, a non-mutated salamander is so small that I think maybe they shouldn't be monsters

We have unmutated frogs so I'd say unmutated salamanders are fair game. I've had salamanders on my todo list for a while for additional aquatic fauna.

data/json/monsters/reptile_amphibian.json
```diff
@@ -390,5 +390,115 @@
...
+    "melee_dice": 6,
+    "melee_dice_sides": 10,
+    "melee_damage": [ { "damage_type": "cut", "amount": 1 } ],
+    "special_attacks": [
+      { "type": "tail_club", "cooldown": 4, "max_range": 5, "allow_no_target": true },
+      { "type": "stretch_bite", "cooldown": 4, "max_range": 5, "allow_no_target": true }
+    ],
+    "dodge": 3,
+    "vision_night": 10,
+    "regenerates": 4,
```
Karol1223, line 495: You didn't seem to fix anything here that Venera told you to.

Karol1223 (changes requested):

fairyarmadillo:
> > In fact, a non-mutated salamander is so small that I think maybe they shouldn't be monsters
> 
> We have unmutated frogs so I'd say unmutated salamanders are fair game. I've had salamanders on my todo list for a while for additional aquatic fauna.

Fair enough, though a bullfrog is a lot bigger than a northeastern salamander. I wonder what the cutoff should be.

Karol1223:
> I wonder what the cutoff should be.

For reference, we also have stuff like the tiniest Sunfish on this side of the continent and a *crayfish*. And a sparrow. There doesn't seem to be a cutoff as far as I can tell. It's kind of just common sense, I think.

PR #65409, opened 2023-05-01
Document burn_data for materials

RenechCDDA:
#### Summary
None

#### Purpose of change
Our docs need more details.

#### Describe the solution
Poke the relevant C++ code, read the code comments, set some things on fire (ingame!!), write docs.

#### Additional context
The fire code is really weird and I'm certain some things (like multiplying by charges' stack_size?!) is not a very good idea at all, and should probably change in the future. But for the moment, let's just have some documentation for our json contributors.


CuteCottonCandyCat:
Why not set things on fire out of game? Sacrifices are always good for development.

PR #65411, opened 2023-05-01
Give Oxy Torch Saw Metal Quality 5 (Correct Git Flow procedure this time I hope)

sam-trafton-dev:
When charged, Acetylene torch also has metal sawing quality that allows for certain vehicle recipes.

#### Summary
Balance "Give torch cutting ability beyond destructing fences."


Zireael07:
Why close?

PR #65413, opened 2023-05-01
[HEAVY WIP] Initial implementation of methane production

Karol1223:
#### Summary
Content "Methane production chain"

#### Purpose of change
Methane production has been suggested a lot and @Drew4484 managed to fool me into doing this.

#### Describe the solution
For starters, adds a new furniture type - the biogas digester. It works kind of like a charcoal kiln except differently. The code, under the hood, is very similar. It takes in stuff made from specific materials (most common comestibles, dry plant, feces - the list currently is WIP and will likely be expanded) and processes them into methane. It also has a logic that, hopefully, excludes corpses from being put there because I don't want to deal with the process of having corpses leave behind bones or chitin.

In future commits I intend to add valves and inner tubes to make the construction recipe for the digester more inline with the real counterpart.

#### Testing
Okay now it should *theoretically* work but still don't get your hopes too high

#### Additional context
Huge thanks to @Drew4484 and @RenechCDDA for walking me through this because if I tried to do this all myself I'd probably cry


RenechCDDA:
I am merely a humble contributor assisting others in copy pasting... I make no claims that I did anything *good* or *useful* here.

Karol1223:
Somehow, the biogas returned

kevingranade:
Unless I'm missing something this isn't producing solid waste? Is that just not done yet or had you missed that facet of a bigoas digester?

Karol1223:
> Unless I'm missing something this isn't producing solid waste? Is that just not done yet or had you missed that facet of a bigoas digester?

Honestly both. The code for this right now is a mess (the C++ currently does not work, because I just copied over the gas pump code to have it at hand). I intend to come back to this and first and foremost make some way to get the methane *out* of the digester, and after that make sure it matches realism.

PR #65415, opened 2023-05-01
New item: large shelter kit and the recipes for it.

Nebnis:
#### Summary
Content "Added a new larger shelter kit with its recipes"

#### Purpose of change
Added a new larger variant of the shelter kit made of skins and leathers. One tile tent makes grug very very sad

#### Describe the solution
Added recipes, and the leather/wood tent

Tents with new different shapes but i guess it is a whole new can of worm for me.

#### Testing
Logged in and everything worked exactly like the large tent except made of animal parts and wood.

#### Additional context
This was way harder than i thought it would have been...


irwiss:
A bunch of sticks and animal skins having same weight and volume as packed plastic/aluminum tent kit doesn't sound right - even the recipe turns 20 1.5kg sticks into 2.2 kg item
3 square meters of leather for a "large" tent also seems a bit weird, or stitching 128 leather patches which comes out to even less even assuming 0 overlap

Nebnis:
> A bunch of sticks and animal skins having same weight and volume as packed plastic/aluminum tent kit doesn't sound right - even the recipe turns 20 1.5kg sticks into 2.2 kg item 3 square meters of leather for a "large" tent also seems a bit weird, or stitching 128 leather patches which comes out to even less even assuming 0 overlap

I tried doubling the mats from the normal shelter kit since it is twice the size, so i assumed all those things were trimmed down during the process. It is hard to ballpark value i guess, so if you have any suggestion to how much i'm all ears.

Meanwhile i tried doing some research into it https://www.comitatus.net/romanleathertent.html. A tent this large would be around  like 45kg, and need 16 leather panels of 75 cm by 25 cm. In-game leather panels are 30 by 60 cm, so around 20 tanned hides? With the kg adding to 50 to 60kg. And the height stretching to 4 meters long.

PR #65416, opened 2023-05-01
Change anvil volume from 4L to 24L

Miner239:
#### Summary
Balance "Change anvil volume to a reasonable 24L"

#### Purpose of change

Anvil was 4L and 54kg. This is 13.5kg/L, which is heavier than a solid cube of lead of the same volume. 

#### Describe the solution

This anvilmaker (https://www.oldworldanvils.com/anvil-dimensions) lists three styles of anvils with its measurements and the 110lb versions measure out to a cube volume of roughly 19.8L, 23.2L, and 27.8L. Calculations for the three respective styles, converted into mL:

image

For a nice even number and considering the small weight difference between 54kg and 110lb, I decided on 24L for the final number for anvil volume.

I thought about making anvils lighter, but that might mean changing its recipe.

#### Testing

Minor JSON change.

### DIFF
--- a/data/json/items/tool/metalworking.json
+++ b/data/json/items/tool/metalworking.json
@@ -5,7 +5,7 @@
     "name": { "str": "anvil" },
     "description": "An enormously heavy block of oddly-shaped steel with a chisel-like projection set into the corner.  It's used in most metalworking fabrication recipes.",
     "weight": "54000 g",
-    "volume": "4 L",
+    "volume": "24 L",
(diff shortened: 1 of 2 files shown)


Karol1223:
You gotta remove the anvil from the known bad density list since it's no longer denser than the material its from.

Miner239:
There's a list???

Karol1223:
data/mods/TEST_DATA/known_bad_density.json

Issue #65417, opened 2023-05-01
Magiclysm rune spells gained above max level


Light-Wave:
### Describe the bug

When gaining the Magus Rune spell from gaining the Magus class, the spell is at level 2 despite having a max level of 0.

### Attach save file

N/A

### Steps to reproduce

1. Make a character as a Would-be Wizard
2. Learn Magic Missile
3. Open the spell list and note you know Magus Rune at level 2.

### Expected behavior

The spell should be at level 0

### Screenshots

![image](https://user-images.githubusercontent.com/66904273/235447433-134a884b-0fce-49e6-b2f0-c8f53d16fccb.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.2846 (22H2)
- Game Version: 3010c62 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Magiclysm [magiclysm]
]

### Additional context

This bug makes the changes in PR #65418 confusing for the player.

PR #65418, opened 2023-05-01
Caster Level Boosters

### LINKED ISSUE #65419: [Magiclysm] Various ways to temporarily boost caster level

Once #65418 gets merged it will be possible to give the player a temporary boost to caster level. This should be used for something so that we don't have unused code in the repository.

### Solution you would like.

I'm picturing a system where different Magiclysm classes have different ways to boost their caster level.
At first, you are limited to boosts by 1 level. Once you reach level 5 in a spell in a given class, and for every 5 levels after that, the maximum boost in that class increases by 1.
Domination-type schools will have relatively cheap craftable items that can be activated to boost your caster level in a specific school. Meanwhile, harmony-type schools will have environmental conditions that boosts your caster level.
Bombastic perks could have a perk that increases your caster level by 1 and is not subject to the maximum boost restriction.

- Animist gets boosted by standing in blood. More blood means a bigger boost.
- Magus can craft some arcane wind-up toy that stores force that can be released to boost caster level for a short duration. A primitive and less effective version made from sticks and twine exists for Innawood players.
- Stormshapers gets a caster level boost while outdoors during precipitation. The worse the weather is, the bigger the boost.
- Kelvinists can craft a magical candle that boosts caster levels while it burns. Can be turned on and off.
- Earthshapers gets a caster level boost while underground. The deeper they are, the bigger the boost. Standing in a pit and deep pit can also help.
- Technomancers can craft an electronic device that boosts caster level. It is also possible to craft magical clay tablets that boost caster level for a while after being broken.
- Druids gets a boost for standing on grass and moss, a bigger boost for standing on bushes and crops, and an even bigger boost for standing on young trees.
- Biomancer gets something they can inject with a syringe. A primitive bone syringe also exists for Innawood players. The content of the syringe is probably crafted from blood and/or flesh. Might cause pain and/or damage when injecting.

Maybe there could also be rings, potions, and staffs that increase caster level.

Maybe attunements should be boostable in some way, either adding the boosts from the two base classes they are based on together, or in some other maner.

There could be an attack spell that increases in damage and mana cost as it levels up, and each tome you cast it you get a temporary stacking boost to caster level, so the more you cast it during a short period, the harder you can blast.

There could be a spell that works in reverse of that, so that each time you cast it you temporarily reduce your caster level of that spell, and once you reach 0, the spell becomes fairly useless. That way your level in the spell kind of also acts as your mana for that spell.

I don't know enough about Xedra Evolved to give design advice, but those responsible for that mod might also want to get in on the action.

### Additional context

_No response_

Light-Wave:
#### Summary
Mods "[Magiclysm] Ways to boos your caster level"

#### Purpose of change

Adds unique ways for the Magiclysm classes to boost their caster level. Also adds some EOC functionality with documentation.
Fixes #65419

#### Describe the solution

The class specific caster level boost is limited to 1 + ( <the highest unboosted caster level you have learned in the relevant school> / 5 )
Control classes (magus, kelvinist, technomancer, and biomancer) gets items that can be used to boost caster level. The items are splitt up into tier 1 and tier 2.
Tier 1 can be crafted from sticks, rocks, plants, and the like. These are Innawood-friendly items that are meant to be items that have been used by casters al long as there have been wizards.
Tier 2 require metal and metalworking tools. They are meant to be generally superior, but requires the recipe to be found in a book. Most tier 2 items boosts caster level by 2 points at once.
You can stock up on as many tier 1 and tier 2 items you want, up to the limit mentioned above.

Animist get wound up devices that cost no resources to use, but requires winding up, and don't last all that long.
Kelvinists gets candles and lamp oils. They give off light and are not free to use; candles burns out and lamp oil requires fuel.
Technomancers gets clay tablets for tier 1 that they can break for a temporary boost. For tier 2, they get a battery powered doodad that is quite bulky. Unlike other classes tier 2 items, this item can boost your caster level upp to the limit on its own, but each time it does so, it draws 5 * <boost given from the doodad> charges from its battery.
Biomancers gets syringes, and have to also craft a booster slurry to use with the syringes. When used, the effect lasts a while, and the tier 2 syringe gives 2 boosts for the cost of 1. Using them causes some pain.

Meanwhile, harmony schools gets caster level boosts from circumstances.
Animist gets boosted by standing in blood. More blood means a higher boost, up to +3.
Stormshapers gets boosted from being outside during precipitation. Either 1 (very light), 3 (light), or 6 (heavy).
Earthshapers gets boosts from being underground. Each z-level below the surface gives a +1 bonus. They also gets bonuses from standing on natural earth, clay, sand, and the like, and gets a bigger bonus from being in a pit.
Druids gets a bonus from being on plants. Generally, the more overgrown the tile is, the bigger the boost, with young trees giving the highest boost.

I also increased the maximum level of the spawn class rune spells to 2 to circumvent issue #65417.

I added 2 new events, `opens_spellbook` and `spellcasting_finish`.
I added 1 more condition; `u_is_on_terrain_with_flag`
I added 2 new arithmetic read-only values; `charge_count` and `field_strength`.
Fixed a bug with u_item_count where it would count the charges in the battery in the items rather than the number if items
I added 1 new read-only value to `math`'s `weather`; `precipitation`
All of these changes have accompanying documentation.

I also add a Bombastic Perk that increases caster level in all schools by 1, which is not bound by the limit mentioned earlier.

Stormshapers might have trouble getting their bonus, and could perhaps either be given a way to make it rain, or given another way to boost their caster level.
Animist boosts might end up being a bit weak end-game. Perhaps there could be a way to boost it further by standing on corpses or in gore?
The tier 1 crafted items does not require a MANA_INFUSE tool because I want these recipes to be craftable before you get to metal in Innawood. This is intentional.
Perhaps a better solution could be found for the caster levels of the runes. I choose to increase the caster level because otherwise the UI would look bugged and the runes would be harder to cast than before. But this is a workaround. See issue #65417
The documentation for math's weather is not quite right now, as it shows all values to be both readable and writable, but one value is not writeable. Not sure how to best resolve that, perhaps break out precipitation to a new row? Or perhaps leave it the way it is right now.

#### Testing

I've been testing that the boosters works correctly that they only work while active, and that the maximum boost increases as your maximum unboosted caster level increases. I've taken a look at the crafting recipes in-game, and they look fine. The new c++ functionality currently has no automated tests written for it, but it's all used in this PR and seems to work perfectly.

I also tested the following code
```
[
  {
    "id": "EOC_test_spell_level",
    "eoc_type": "EVENT",
    "required_event": "opens_spellbook",
    "effect": [
      { "arithmetic": [ { "u_val": "spell_level_adjustment" }, "+=", { "const": 1 } ] },
      { "arithmetic": [ { "u_val": "spell_level_adjustment", "school": "MAGUS" }, "+=", { "const": 2 } ] },
      { "arithmetic": [ { "u_val": "spell_level_adjustment", "spell": "magic_missile" }, "+=", { "const": -15 } ] }
    ]
  }
]

```
image
Magic missile shows as expected a caster level of 0 when your base spell level is 0.

image
Magical Light shows a caster level of +1, and I tested that the spawned item had a duration that matches the increased caster level.

image
Phase door shows the expected caster level of 0 +3

image
After leveling up Magic Missile, it now displays as being level 1 -1, and is confirmed to still only do 1 damage.

image
I spawned in an NPC and gave him magic missile at level 20.

image
With a minus 15 caster level modifier, 6 damage from the NPC sounds about right.


Light-Wave:
Pinging @Ramza13

andrei8l:
The `context` class is pretty pointless since there's only one global instance. It also makes the adjustments apply to NPCs too even though you're scoping on the avatar in JSON (`u_val` instead of `npc_val` in your examples).

The adjustments should instead of part of the target talker's `known_magic` and use maps of `spell_id` and `trait_id`, for spells and schools respectively, instead of raw strings.

`arithmetic` is obsolete in the medium-long term and new functions should be added to `math` instead (in this case, in `math_parser_diag.h/cpp`). However, since the `spell_level` function hasn't been ported yet, this can also stay in `arithmetic` for now.

Light-Wave:
> The `context` class is pretty pointless since there's only one global instance. It also makes the adjustments apply to NPCs too even though you're scoping on the avatar in JSON (`u_val` instead of `npc_val` in your examples).
> 
> The adjustments should instead be part of the target talker's `known_magic` and use maps of `spell_id` and `trait_id`, for spells and schools respectively, instead of raw strings.
> 
> `arithmetic` is obsolete in the medium-long term and new functions should be added to `math` instead (in this case, in `math_parser_diag.h/cpp`). However, since the `spell_level` function hasn't been ported yet, this can also stay in `arithmetic` for now.

All right, that is all fair.
I removed the context class and moved the data storage to knows_magic, as suggested. I also switched to using spell_id and trait_id.
I also made it so that NPCs are affected the same way the player is.

The way EoCs that trigger on events works is that if it is an NPC that triggered the event, the NPC will be considered the primary speaker. Thus, you use u_val to modify the values of the caster regardless of if the character that opened the spellbook is a player or an NPC. I'm not sure if I should keep using u_val, or if I should treat it as a variable not connected to a speaker as far as JSON is concerned?

Ramza13:
Looks good to me!

Fris0uman:
Could you also add content that uses this feature please

Light-Wave:
> Could you also add content that uses this feature please

All right, I'll add that as well. I'll make this a draft while working on that.

Light-Wave:
This is now ready for review.

KorGgenT:
this is all pretty cool. i've got a little UX comment here - we need to display the final value of spell level. plus and minus should be in parenthesis off to the side somewhere. i don't want to have the user be forced to do math to figure out what spell level they're casting at.

Other than that minor thing, LGTM

PR #65422, opened 2023-05-01
added missing proficiencies to the modular kitchen station

Daved27hundred:
#### Summary
None

#### Purpose of change

the new modular kitchen added in #64682 uses welding tools but doesn't have any proficiencies needed in it's craft.

#### Describe the solution

added `principles of metalworking` and `principles of welding` to the recipe. I also added `general appliance repair` because I missed it in my previous PR

If there's a reason why it doesn't have these proficiencies let me know, I can close this.


irwiss:
The reason they don't have proficiencies is that the old rig recipes that I used as template had no proficiencies on them
No reason to not have some though

PR #65424, opened 2023-05-02
Turret Weakpoints & Review

SolventMercury:
#### Summary
Content "Gave Turrets Weakpoints"

#### Purpose of change
Part of my ongoing crusade to make proficiencies and weakpoints relevant.

#### Describe the solution
- Added a set of weakpoints for turrets, and a weakpoint set specific to the searchlight (mostly for flavor).
- Gave turrets higher stab armor, since stabbing would be relatively ineffective compared to its use against other kinds of enemies.
- Removed STUN_IMMUNE flag from turrets and stuff. Mostly to facilitate them getting locked up through weakpoint effects. I feel like a turret is light enough that you could screw it up just by hitting it real hard, so this doesn't strike me as an incorrect choice.

N/A

#### Testing
Spawned a lot of turrets in, tuned until the results felt right.


fairyarmadillo:
This will require players to make a real spear instead of just poking them with a stick, which is good

It would make sense for most robots and cyborgs to be extra resistant to stab, and might help with the present combat knife/bow supremacy. Something to consider if you or someone else winds up getting to those.

SolventMercury:
Ok for some reason the automatic formatter stuff didn't work. Will fix in a bit.

LordZanos:
Monsters automatically have an amount of stab armor equal to a percentage of their cut armor, which is why it wasn't set previously. Is there a particular reason turrets should be more resilient to stabbing then cutting?

SolventMercury:
> Monsters automatically have an amount of stab armor equal to a percentage of their cut armor, which is why it wasn't set previously. Is there a particular reason turrets should be more resilient to stabbing then cutting?

Yes. Turrets naturally will be very thin in most areas, and the rigid, rounded edges would tend to deflect the tips of pointed weapons. Their profiles are also just kinda narrow, especially since the weapon is always turning to face towards the player. Cutting weapons, however, do not have this issue, since they swing in arcs.

SolventMercury:
This is what a CROWS looks like.

image

I cannot see many good points to attack with a spear (besides the lens, which I've already got as a weak point). If I had like an axe though I'd like my options a bit more.

fairyarmadillo:
I would argue that nearly all mechanical enemies should be resistant to stab as they don't have soft tissue, blood, or vital organs, and are largely made of hard materials which are difficult to penetrate with low-energy weapons.

This applies to a lesser degree to cut, but cutting weapons are typically striking with more force and can do things like shred tires or sever cables, or simply batter the target with a hacking strike.

Not saying turrets or talons should be immune to spears, just that the normal logic of thrusting weapons penetrating armor to reach the soft insides doesn't hold up as well when the insides aren't soft.

SolventMercury:
Ok, can somebody tell me why the JSON check isn't passing? It's not invalid JSON and I did everything GitHub suggests.

fairyarmadillo:
> Ok, can somebody tell me why the JSON check isn't passing? It's not invalid JSON and I did everything the JSON bot suggests.

It needs linting, according to the log. Copy and paste the entire contents of the json files you're changing here https://dev.narc.ro/cataclysm/format.html and then repaste them over the jsons in the PR. It is probably just a line break or extra space somewhere.

SolventMercury:
> It needs linting, according to the log. Copy and paste the entire contents of the json files you're changing here https://dev.narc.ro/cataclysm/format.html and then repaste them over the jsons in the PR. It is probably just a line break or extra space somewhere.

Thanks, I hadn't been pointed to that link. That's a helpful resource.

NetSysFire:
I patched up your PR by resolving the conflicts and linting a file, now it should be able to get merged again.

PR #65432, opened 2023-05-02
Show NPC location when selecting NPC to chat with, guard, or follow

Qrox:
#### Summary
Interface "Show NPC location when selecting NPC to chat with, guard, or follow"

#### Purpose of change
So you know who is who when entering the refugee center.

#### Describe the solution
Use `pointmenu_cb` in the respective functions. Also fix a bug where selecting with mouse did not update the cursor position.

#### Testing
image


Karol1223:
As a person who spent my entire time in the refugee center yelling to smokes from across the room because I confused him for the doctor - thank you so much.

PR #65433, opened 2023-05-02
Floor craft difficulty at 1 to avoid zero'ing out player's calculated skill

### LINKED ISSUE #64241: Skill 0 and Training recipes are extremely hard regardless of Skill
### Describe the bug

0 skill recipes like clean water, heavy duty threads or lockpicks as well as all Practice crafts have a super high failure chance at all skill levels, l this is pretty extreme even at a default 8 int and just ridiculous at int 4.

### Attach save file

This can be seen in a fresh save file

### Steps to reproduce

1 - Make a new character with high skill
2 - Check recipes like clean water

### Expected behavior

Higher skill makes crafting easier even at int 4

### Screenshots

0 Skill 8 INT
Screenshot_1
10 Skill 8 INT, same success as if they had zero
Screenshot_2
Recipes that take 1 or more skill to craft works normally
Screenshot_3
Screenshot_4
Any craft or practice at int 4 becomes nearly impossible
Screenshot_5
Screenshot_6

- OS: Windows
- Game Version: a98d542 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
    Magiclysm [magiclysm],
    Portal Storms Ignore NPCs [personal_portal_storms]
]

### Additional context

The same thing happens with all training regardless of skill, this is definitely a bug since I've been playing with a 4 int survivor and around the time the new crafting odds display was introduced this started to happen, Why you can even fail in practice makes me confused to since you're not actually making anything most of the time and needing to replace lost components is just frustrating

RenechCDDA:
#### Summary
Bugfixes "Difficulty 0 recipes are no longer arbitrarily difficult"

#### Purpose of change
image
image

* Fixes #64241

#### Describe the solution
Make the player's skill not count as 0 just because the difficulty is 0

#### Testing
All tests were carried out by a 8/8/8/8 character with no impairments, using recipes that used no proficiencies or secondary skills.

test recipe (my skill / skill required)
% minor fail chance before --> % minor fail chance after
% catastrophic fail chance before --> % catastrophic fail chance after

clean water (skill 1/0)
50.00 --> 21.19
9.12 --> 0.07

clean water (skill **10**/0)
50.00 --> 0.00
9.12 --> 0.00

sock mitts (skill 0/0)
50.00 --> 50.00
9.12 --> 9.12

baked beans (skill 1/1)
50.00 --> 50.00
0.98 --> 0.98

gloves (skill 0/1)
71.61 --> 71.61
19.57 --> 19.57

As we can see, the only scenario in which this affects the failure chance is when being overskilled for difficulty 0 recipes.

#### Additional context

@anothersimulacrum @I-am-Erk 
Since this is your child, it only seems appropriate to let you know.

### DIFF
--- a/src/crafting.cpp
+++ b/src/crafting.cpp
@@ -997,9 +997,10 @@ float Character::get_recipe_weighted_skill_average( const recipe &making ) const
     // The primary required skill counts extra compared to the secondary skills, before factoring in the
     // weight added by the required level.
     const float weighted_skill_average =
-        ( ( 2.0f * making.difficulty * get_skill_level( making.skill_used ) ) + secondary_skill_total ) /
+        ( ( 2.0f * std::max( making.difficulty,
+                             1 ) * get_skill_level( making.skill_used ) ) + secondary_skill_total ) /
         // No DBZ
-        std::max( 1.f, ( 2.0f * making.difficulty + secondary_difficulty ) );
+        std::max( 1.f, ( 2.0f * std::max( making.difficulty, 1 ) + secondary_difficulty ) );
     add_msg_debug( debugmode::DF_CRAFTING, "Weighted skill average: %g", weighted_skill_average );
 
     float total_skill_modifiers = 0.0f;
--- a/tests/crafting_test.cpp
+++ b/tests/crafting_test.cpp
@@ -724,10 +724,10 @@ TEST_CASE( "crafting_failure_rates_match_calculated", "[crafting][random]" )
     const recipe_id armor_qt_lightplate( "armor_qt_lightplate_test_no_tools" );
 
     // Skill zero recipes
-    test_chances_for( makeshift_crowbar, 50.f, 50.f, 50.f, 50.f, 50.f, 50.f, 50.f );
-    test_chances_for( meat_cooked, 50.f, 50.f, 50.f, 50.f, 50.f, 50.f, 50.f );
-    test_chances_for( club_wooden_large, 50.f, 50.f, 50.f, 50.f, 50.f, 50.f, 50.f );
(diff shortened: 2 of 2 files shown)


I-am-Erk:
Since it's using a float, can you set the difficulty to 0.5, or is it rounded off somewhere?

RenechCDDA:
recipe::difficulty returns an int and std::max doesn't like the idea of comparing ints with non-ints. I can cast the recipe::difficulty to be a float but I haven't really worked with static casting before and don't know if that's safe. (I've asked in the dev discord and will update this if I get an answer)

I did try compiling and repeating the test with casting difficulty to a float (so `std::max( static_cast<float>(making.difficulty), 0.5f )` would have been the change) but that results in.... changes across the board for some reason?  FWIW I think this is probably just me not being familiar with casting.

Below is the test updated with effective difficulty of 0.5. This implementation is definitely not desired, as it makes changes outside of being overskilled for difficulty 0 recipes.

clean water (skill 1/0)
50.00 --> 28.39
9.12 --> 0.38

clean water (skill **10**/0)
50.00 --> 0.00
9.12 --> 0.00

sock mitts (skill 0/0)
50.00 --> 56.36
9.12 --> 16.85

baked* beans (skill 1/1)
50.00 --> 56.36
0.98 --> 2.74

gloves (skill 0/1)
71.61 --> 75.48
19.57 --> 29.06

RenechCDDA:
Yes, we could cast both to 0.5f. But that actually produces the same outcome as my original changes due to the std::max wrap in the second division part. It makes the code more difficult to read and in general I would be against it without a functional reason.

I am going ahead with the original changes, and we'll see if the mergers accept them.

PR #65434, opened 2023-05-02
enable and disable mutations EOCs

bombasticSlacks:
#### Summary
Infrastructure "Added enable and disable mutation EOCs"

#### Purpose of change
Something people desired, adds mutation enable and disable EOC. Also all Mutation EOCs now have acces to _this a context variable with the current mutation id stored in it.

#### Describe the solution
Added the new u_activate_trait and u_deactivate_trait stuff
and
the storage of _this

#### Testing
added robust unit tests

#### Additional context
Documentation is included

### DIFF
--- a/data/mods/TEST_DATA/EOC.json
+++ b/data/mods/TEST_DATA/EOC.json
@@ -146,12 +146,15 @@
   {
     "type": "effect_on_condition",
     "id": "EOC_activate_mutation_test",
-    "effect": [ { "u_deactivate_trait": {"context_val": "this" } } ]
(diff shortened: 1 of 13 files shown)


bombasticSlacks:
> Can we get a mutations, that can be activated, but only with this EoC then? Ie so player won't be able to turn it off manually, but only when certqin conditions are met? (or maybe we have such system already?) 

I don't think we have that exactly no, but I add it here :)

bombasticSlacks:
Actively working on it

PR #65447, opened 2023-05-03
Play with the corvids

### LINKED ISSUE #65368: Playing With Corvids

Raised several crows and ravens from eggs, was able to “play” under interaction menu with all in their youth stages, but not when they reached adulthood

### Solution you would like.

Add necessary tags to make play activity possible with adult crows and ravens

Adding tags to all tamable adult animals, removing tags from chicks (?)

### Additional context

While all real-life animals above a certain level of intelligence can and have benefited from play behaviors and environmental enrichment, corvids are noted to be extremely intelligent and social animals in their own right. Wild and, indeed, hand-reared birds do not “grow out of” play behaviors! Let’s give them some love.

Golfavel:
#### Summary
Balance "Fix bird chicks and playing with corvids"

#### Purpose of change
Fixes #65368 

#### Describe the solution
Adds feeding and playing to adult crows and ravens. Also fixes(?) issue where you can tame and play with *any* chick due to copy-from shenanigans by introducing a generic chick that you cannot tame.

Just adding playing to the corvids instead of tackling all of the chicks' copy-from.
Not allowing taming when a corvid is an adult.

#### Testing
Crows and ravens can be tamed with birdfood (albeit with some difficulty) and you can play with them afterwards. Chicks that you tame will grow up and can still be played with. You can no longer tame most chicks.

### DIFF
--- a/data/json/monsters/bird.json
+++ b/data/json/monsters/bird.json
@@ -458,7 +458,6 @@
     "id": "mon_raven_chick",
     "type": "MONSTER",
     "copy-from": "mon_crow_chick",
-    "name": { "str": "brown chick" },
     "volume": "750 ml",
     "weight": "60 g",
     "upgrades": { "age_grow": 35, "into": "mon_raven" }
@@ -466,8 +465,7 @@
   {
     "id": "mon_bluejay_chick",
     "type": "MONSTER",
-    "copy-from": "mon_crow_chick",
-    "name": { "str": "brown chick" },
+    "copy-from": "mon_generic_chick",
     "volume": "225 ml",
     "weight": "38 g",
     "upgrades": { "age_grow": 18, "into": "mon_bluejay" }
@@ -475,8 +473,7 @@
   {
     "id": "mon_cardinal_chick",
     "type": "MONSTER",
-    "copy-from": "mon_crow_chick",
-    "name": { "str": "brown chick" },
+    "copy-from": "mon_generic_chick",
     "volume": "200 ml",
     "weight": "15 g",
     "upgrades": { "age_grow": 11, "into": "mon_cardinal" }
@@ -484,8 +481,7 @@
   {
     "id": "mon_robin_chick",
     "type": "MONSTER",
-    "copy-from": "mon_crow_chick",
-    "name": { "str": "brown chick" },
+    "copy-from": "mon_generic_chick",
     "volume": "215 ml",
     "weight": "35 g",
     "upgrades": { "age_grow": 14, "into": "mon_robin" }
@@ -493,8 +489,7 @@
   {
     "id": "mon_sparrow_chick",
     "type": "MONSTER",
-    "copy-from": "mon_crow_chick",
-    "name": { "str": "brown chick" },
+    "copy-from": "mon_generic_chick",
     "volume": "175 ml",
     "weight": "12 g",
     "upgrades": { "age_grow": 15, "into": "mon_sparrow" }
--- a/data/json/monsters/bird.json
+++ b/data/json/monsters/bird.json
@@ -744,8 +744,7 @@
   {
     "id": "mon_hummingbird_chick",
     "type": "MONSTER",
-    "copy-from": "mon_crow_chick",
-    "name": { "str": "brown chick" },
+    "copy-from": "mon_generic_chick",
     "volume": "80 ml",
     "weight": "7 g",
     "upgrades": { "age_grow": 11, "into": "mon_hummingbird" }
@@ -753,8 +752,7 @@
   {
     "id": "mon_woodpecker_chick",
     "type": "MONSTER",
-    "copy-from": "mon_crow_chick",
-    "name": { "str": "brown chick" },
+    "copy-from": "mon_generic_chick",
     "volume": "200 ml",
     "weight": "15 g",
     "upgrades": { "age_grow": 11, "into": "mon_woodpecker" }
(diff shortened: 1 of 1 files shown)


Karol1223:
Is the generic chick supposed to just be used for ``copy-from``? Because if so you should make it an abstract not a monster

RenechCDDA:
I agree with Karol here, it can and should be an abstract.

Golfavel:
Done.

Karol1223 (changes requested):
Okay I looked at this again and this actually breaks things. By creating the abstract you made all chicks except the chicken and, for some reason, wild crows and ravens untameable. This means you can no longer tame ducks, turkeys or pheasants (which only would need to tolerate your presence for ''domestication'') but you can tame highly intelligent wild birds. I'm not a huge fan of this and I don't believe this should work like that. Crows being tameable was specifically a side effect of us having all chicks be tameable, having *them* be tamed but not having animals from species that actually have been domesticated is not a good thing. You even made ADULT crows and ravens tameable, and that's definitely not something that should be possible.

Granted that this still poses a problem because with ducks or turkeys we don't differentiate between domestic and wild populations, wild ones should still likely be untameable in mainline. But corvids definitely shouldn't be more tameable than mallards.

If the general consensus is that wild birds should not be tameable then corvids should not be tameable either - that should be moved to Tameable Wildlife. If the general consensus is that we make some exceptions then ducks and turkeys should be here first, corvids only as a secondary possible choice.

data/json/monsters/bird.json
```diff
@@ -82,7 +82,12 @@
     "reproduction": { "baby_egg": "egg_crow", "baby_count": 4, "baby_timer": 18 },
     "baby_flags": [ "SPRING" ],
     "biosignature": { "biosig_item": "feces_bird", "biosig_timer": 8 },
-    "flags": [ "SEES", "HEARS", "SMELLS", "ANIMAL", "PATH_AVOID_DANGER_1", "WARM", "FLIES", "SWARMS" ]
+    "petfood": {
```
Maleclypse, line 85:
```suggestion
```

data/json/monsters/bird.json
```diff
@@ -82,7 +82,12 @@
...
+      "food": [ "BIRDFOOD" ],
```
Maleclypse, line 86:
```suggestion
```

data/json/monsters/bird.json
```diff
@@ -82,7 +82,12 @@
...
+      "feed": "The %s seems to like you!  It runs around your legs and seems friendly.",
```
Maleclypse, line 87:
```suggestion
```

data/json/monsters/bird.json
```diff
@@ -82,7 +82,12 @@
...
+      "pet": "The %s runs around your leg."
```
Maleclypse, line 88:
```suggestion
```

data/json/monsters/bird.json
```diff
@@ -82,7 +82,12 @@
...
+    },
```
Maleclypse, line 89:
```suggestion
```

data/json/monsters/bird.json
```diff
@@ -114,7 +119,12 @@
     "reproduction": { "baby_egg": "egg_raven", "baby_count": 5, "baby_timer": 18 },
     "baby_flags": [ "SPRING" ],
     "biosignature": { "biosig_item": "feces_bird", "biosig_timer": 8 },
-    "flags": [ "SEES", "HEARS", "SMELLS", "ANIMAL", "PATH_AVOID_DANGER_1", "WARM", "FLIES", "SWARMS" ]
+    "petfood": {
```
Maleclypse, line 122:
```suggestion
```

data/json/monsters/bird.json
```diff
@@ -114,7 +119,12 @@
...
+      "food": [ "BIRDFOOD" ],
```
Maleclypse, line 123:
```suggestion
```

data/json/monsters/bird.json
```diff
@@ -114,7 +119,12 @@
...
+      "feed": "The %s seems to like you!  It runs around your legs and seems friendly.",
```
Maleclypse, line 124:
```suggestion
```

data/json/monsters/bird.json
```diff
@@ -114,7 +119,12 @@
...
+      "pet": "The %s runs around your leg."
```
Maleclypse, line 125:
```suggestion
```

data/json/monsters/bird.json
```diff
@@ -114,7 +119,12 @@
...
+    },
```
Maleclypse, line 126:
```suggestion
```

Maleclypse (changes requested):
Taming an adult corvid is much more involved than taming a corvid chick and raising it.  I would agree that adult corvids raised from chicks should have `CANPLAY`.

Golfavel:
Done.

Karol1223:
The problem with this PR completely axing the taming of all wild birds except for two arbitrarily chosen corvids still stands though.

If you go with corvid taming, go all in at least. But I still do believe that a corvid should not be easier to 'tame' than a mallard or a turkey. Their intelligence wouldn't make this easier and removing the domestication of those two in favor of crows seems extremely weird and unfitting.

Golfavel:
I'm unsure if you're wanting to have taming adult corvids restored or to restore taming for baby turkeys and pheasants (+anything else) (duck chicks have remained the same)?

Karol1223:
Adult taming of corvids should be gone as I suggested before already.

I do admit I did not see ducks had separately defined taming - that's on me. But pheasants, turkeys, and arguably geese should still be tameable if raised from an egg. And if you *are* going with corvid hatchling taming go all in and include bluejays.

Also, you've not touched even half of the bird hatchlings we have. Please review the entire file and audit the other chicks - you're leaving stuff like *hummingbirds* tameable.

Maleclypse (approved):
Thanks for doing this!

PR #65457, opened 2023-05-03
Add avatar swim tests

ralreegorganon:
#### Summary

None

#### Purpose of change

Add unit tests that assert our current swimming behavior so we have a baseline before changing any of the mechanics.

#### Describe the solution

Add the tests.

None

#### Testing

The added unit tests, as well as some ad-hoc testing of a subset of the scenarios being tested to compare the behavior in game versus the tests.

#### Additional context

The tests configure the avatar with different stats, gear, mutations, and skills and make them swim until they're winded. If they swim more than 9000 tiles it terminates--we can revisit that threshold whenever we revisit the mechanics but we need an upper bound. Because there are many cases that go above that threshold (and when unbounded go WAAAAAY above the threshold) we can't read too much into those cases for now...and this isn't really about whether the results make sense at this point, just that we have results we can look at. 

For fun, I've attached the CSVs the utility functions export. There are some interesting things to be observed even in just a casual review, such as:

* the movement mode interactions make no sense: the same basic avatar with no gear, skills, or traits, and just basic (8s) stats will only swim 3 tiles while running, 31 while walking, and 183 while prone...and of course the question of what those movement modes mean when in the water anyway.
* our aquatic professions aren't very good swimmers, and the migration of the swimming skill to a generalized athletics skill means that our athletes like the marathon runner and football player are just as good at swimming as our dedicated swimmers
* as independent variables, athletics skill has a much larger impact on swimming than strength or dexterity
* our mutations that impact swimming are pretty underwhelming compared to e.g. finding a pair of fins

[swim-dexterity.csv](https://github.com/CleverRaven/Cataclysm-DDA/files/11390196/swim-dexterity.csv)
[swim-profession.csv](https://github.com/CleverRaven/Cataclysm-DDA/files/11390197/swim-profession.csv)
[swim-scenarios.csv](https://github.com/CleverRaven/Cataclysm-DDA/files/11390198/swim-scenarios.csv)
[swim-skill.csv](https://github.com/CleverRaven/Cataclysm-DDA/files/11390199/swim-skill.csv)
[swim-strength.csv](https://github.com/CleverRaven/Cataclysm-DDA/files/11390200/swim-strength.csv)

### DIFF
--- a/tests/player_helpers.cpp
+++ b/tests/player_helpers.cpp
@@ -23,6 +23,7 @@
 #include "player_activity.h"
 #include "player_helpers.h"
 #include "point.h"
+#include "profession.h"
 #include "ret_val.h"
 #include "stomach.h"
 #include "type_id.h"
@@ -129,6 +130,8 @@ void clear_character( Character &dummy, bool skip_nutrition )
     dummy.set_per_bonus( 0 );
 
     dummy.cash = 0;
+
+    dummy.prof = profession::generic();
 
     const tripoint spot( 60, 60, 0 );
     dummy.setpos( spot );
--- a/tests/water_movement_test.cpp
+++ b/tests/water_movement_test.cpp
@@ -1,30 +1,42 @@
 #include <memory>
+#include <utility>
 
 #include "avatar.h"
+#include "avatar_action.h"
 #include "cata_catch.h"
 #include "creature.h"
 #include "game.h"
 #include "map.h"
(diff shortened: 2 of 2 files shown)


anoobindisguise:
Swimming is definitely screwy. If you have low athletics skill the difference between being able to swim and instantly drowning can literally be wearing a pair of shorts with 1 encumbrance.

PR #65398, opened 2023-04-30
Add Molting for Exoskeletons

worm-girl:
#### Summary
Features "Add molting for exoskeleton mutations"

#### Purpose of change

Real-life arthropods periodically have to molt their exoskeletons, usually a few times a year. This is a process that begins to slow them down, during which time they typically seek shelter. Once the molt begins, they are helpless for a short time, then have severely compromised armor until their new chitin can sclerotize and harden via exposure to air or water.

Simulating this process for arthropod mutants adds to the body horror aspect, opens the door to new gameplay considerations, and is just a neat way to flex EOCs.

#### Describe the solution

Using the EOC system, mutants with CHITIN2 or CHITIN3 will begin molting about every 5 weeks. They're given a small debuff for roughly 24 hours which advises them that the molt is coming, then a much worse debuff for an hour or two as their exoskeleton begins to lock up completely, with clear and direct warnings to seek shelter. After that, they are immobilized for 5 minutes (CHITIN2) or 15 minutes (CHITIN3) and get some body horror-y messaging about emerging from their own skin.

Once the above process has completed, they lose their CHITIN2, and CHITIN3 mutations and are given molted versions of the same, which have greatly diminished protection values on the associated integrated items, but a bit less encumbrance as well (halved for CHITIN2, 2/3rds for CHITIN3, rounding up). They receive a few chunks of chitin to represent their old exoskeleton. After a day or so, the new exoskeleton hardens and they are back to normal.

In real arthropods, this post-molt period often leaves them in a near-helpless state as their muscles require a rigid exoskeleton to move, however Cataclysm arthropods and arthropod mutants are using endoskeletons composed of bone or endochitin, so we probably don't need to go that far.

There's also a CHITIN_MOLTED, which is only there to prevent molting mutants from magically losing their chitin in the unlikely event that they drink mutagen mid-molt and it tries to give them CHITIN2 or CHITIN3.

Additionally, a preexisting bug was discovered during this PR. CHITIN_FUR2, CHITIN_FUR3, and CF_HAIR were conflicting with the chitin mutations, meaning they would cancel your armor entirely to give you hair. This PR introduces a fix to this issue by giving the chitin_fur mutations their own type, BODY_ARMOR_MOD. This type can be used for mutations which don't actually provide armor, but rather modify mutations which do. 

With these changes, the CHITIN_FUR mutations can properly cancel each other as the mutant becomes hairier, and are not lost when upgrading chitin tiers via mutation.

#### Describe alternatives you've considered

[...]

PR #65459, opened 2023-05-03
Redo molting with crabs

worm-girl:
#### Summary
Features "Add molting for exoskeleton mutations"

#### Purpose of change

Real-life arthropods periodically have to molt their exoskeletons, usually a few times a year. This is a process that begins to slow them down, during which time they typically seek shelter. Once the molt begins, they are helpless for a short time, then have severely compromised armor until their new chitin can sclerotize and harden via exposure to air or water.

Simulating this process for arthropod mutants adds to the body horror aspect, opens the door to new gameplay considerations, and is just a neat way to flex EOCs.

Presently, this PR won't be a huge gameplay change, however in the future, things like limb regeneration could be tied to molt cycles. This system could also serve as the basis for others - Mammal and maybe bird mutants could have winter coats or down that made life a little more bearable in summer, and plants and elves could shed their leaves. It might even be fun to give Chimerae EOCs that periodically change the distribution of materials on their patchwork armor.

#### Describe the solution

Using the EOC system, mutants with CHITIN2, CHITIN3, or CRUSTACEAN_CARAPACE will begin molting about every 4-5 weeks. They're given a small debuff for roughly 24 hours which advises them that the molt is coming, then a much worse debuff for an hour or two as their exoskeleton begins to lock up completely, with clear and direct warnings to seek shelter. After that, they are immobilized for 5 minutes (CHITIN2) or 15 minutes (CHITIN3, CHITIN_CARAPACE) and get some body horror-y messaging about emerging from their own skin.

Once the above process has completed, they lose their CHITIN2, CHITIN3, and CHITIN_CARAPACE mutations and are given molted versions of the same, which have greatly diminished protection values on the associated integrated items, but a bit less encumbrance as well (halved for CHITIN2, 2/3rds for CHITIN3 and CHITIN_CARAPACE, rounding up). They receive a few chunks of chitin to represent their old exoskeleton. After a day or so, the new exoskeleton hardens and they are back to normal.

In real arthropods, this post-molt period often leaves them in a near-helpless state as their muscles require a rigid exoskeleton to move, however Cataclysm arthropods and arthropod mutants are using endoskeletons composed of bone or endochitin, so we probably don't need to go that far.

There's also a CHITIN_MOLTED, which is only there to prevent molting mutants from magically losing their chitin in the unlikely event that they drink mutagen mid-molt and it tries to give them CHITIN2, CHITIN3, or CHITIN_CARAPACE.

Additionally, a preexisting bug was discovered during this PR. CHITIN_FUR2, CHITIN_FUR3, and CF_HAIR were conflicting with the chitin mutations, meaning they would cancel your armor entirely to give you hair. This PR introduces a fix to this issue by giving the chitin_fur mutations their own type, BODY_ARMOR_MOD. This type can be used for mutations which don't actually provide armor, but rather modify mutations which do.

With these changes, the CHITIN_FUR mutations can properly cancel each other as the mutant becomes hairier, and are not lost when upgrading chitin tiers via mutation.

This PR also gives the crustacean carapace mutation the 4 bonus bashing damage that its precursor sclerotin plate gets, as it looks to have been left out by accident in the PR which added it.

I'm not attached to any of the numbers I've listed here, they're just what looked good to me in testing. The integrated CHITIN2, CHITIN3, and CRUSTACEAN_CARAPACE armors are half thickness and made of leather_arthropod instead of chitin and sclerotin, respectively, as that seemed logical to me, but they could be weaker if needed. If you've ever eaten a soft-shelled crab, you probably understand how weak a molted (and cooked, to be fair) crustacean really is.

The debuffs on molting are meant to be minor, molting_imminent is meant to be fairly devastating. I pinned the molt cycles at roughly five weeks because we run into the portal storm problem if they're too common, where they stop being cool and start just being an interruption to gameplay. The fifteen minute molt for CHITIN3 and CRUSTACEAN_CARAPACE is based on a NOAA article about the Bering sea Tanner crab. https://www.fisheries.noaa.gov/alaska/science-data/alaska-shellfish-growth-research

CHITIN2 is faster to get out of because it's thinner, softer, and doesn't cover quite as much of the body.

#### Testing

I scaled all the molting timers down to be doable within a single day, then I successfully completed several molts with both CHITIN2 and CHITIN2+CHITIN3. I made sure that CHITIN1 wasn't molting on its own - it's soft like skin, so I imagine it can molt off in small fragments like human skin does.

I also gave CHITIN2 and CHITIN3 to an NPC and observed them successfully completing multiple molt cycles. After adding CHITIN_CARAPACE to this PR, I repeated all tests with it to see that it ran correctly.

I rescaled the molting timers to their current weeks-long cycles and used debug to advance time until a molt began, then watched it resolve in the expected timeframe.

I also tried drinking mutagen while molting to ensure that the mutagen wouldn't improperly replace a freshly molted exoskeleton with a new one, or double them up, or anything like that. The freshly molted versions of the exoskeleton mutations are not purifiable, and behaved as expected - you can't use mutagen or purifier to replace or remove your exoskeleton while they're still hardening, but afterwards they go back to normal.


fairyarmadillo:
This is an updated version of https://github.com/CleverRaven/Cataclysm-DDA/pull/65398 - It was easier to make a new PR than to try to shoehorn everything into the old one.

data/json/effects.json
```diff
@@ -4030,5 +4030,45 @@
     "desc": [ "You are covered in glowing nether spores!" ],
     "apply_message": "The glowing spores settle on you!",
     "rating": "bad"
+  },
+  {
+    "type": "effect_type",
+    "id": "molting",
+    "name": [ "Molting" ],
+    "desc": [ "You're exoskeleton feels uncomfortably tight.  Perhaps you should find somewhere to rest." ],
```
fungamer2-2, line 4038:
```suggestion
    "desc": [ "Your exoskeleton feels uncomfortably tight.  Perhaps you should find somewhere to rest." ],
```

fairyarmadillo:
lol the test had previously failed because the test character caught the flu while trying to swim, what are the chances?

Looks good now. Please let me know if there's something I need to change.

.gitignore
```diff
@@ -235,3 +235,4 @@ Gemfile.lock
 
 # VERSION.txt generated by our CMakeLists.txt
 VERSION.txt
+data/json/monsters/bird.json
```
kevingranade, line 238:
```suggestion
```

kevingranade:
Love it, just need to wait on the tests to re run because I had to remove an errant .gitignore entry. 

You missed out,  it could have been "remoulting".

Fun fact,  either spiders or scorpions, can't recall which,  moult the inside of their mouths and digestive systems.

PR #65461, opened 2023-05-04
fix matches/ref_matches category typo

nornagon:
#### Summary
None

#### Purpose of change

matches and ref_matches had "category": "containers", but there's no ITEM_CATEGORY with the id "containers". There is one with "container". That's probably the one the PR author meant.

#### Describe the solution

Fixed the typo.

#### Additional context

Discovered through HHG runtime error tracking.


Karol1223:
Oops! Sorry for that. Though it is kind of weird we don't have a test catching those. I thought that was under the unvisited members check?

Anyway, apologies for that.

RenechCDDA:
> thought that was under the unvisited members check?

From what I understand unvisited members would be for "category". Category is a member of the object. Whether that member has bad data is up to the member to deal with (each might handle it differently, like location for vehicle parts which just accepts whatever you give it)

nornagon:
yeah, afaik this property was still _visited_, it just referred to something incorrect :)

nornagon:
This is still relevant.

Karol1223:
It is relevant but it's missing the basic build. Though I've been looking at it a few times and can't tell why. It may just need a test kick.

data/json/items/tool/fire.json
```diff
@@ -161,7 +161,7 @@
     "type": "TOOL",
     "name": { "str": "matchbook" },
     "description": "A small book of matches with a coarse strike surface on the outside flap for lighting them.  Allows you to use various drugs like cigarettes, light explosives like Molotov cocktails, or light nearby items on fire.  Due to the matches being lightly glued to the matchbook, they can be removed, but not replaced on the book.",
-    "category": "containers",
+    "category": "container",
```
Maleclypse, line 164:
```suggestion
    "category": "container",
    "
```

data/json/items/tool/fire.json
```diff
@@ -161,7 +161,8 @@
...
+    "
```
Maleclypse, line 165:
```suggestion
```

Maleclypse:
Tests had sat too long to restart,  I've made these commits to restart them

data/json/items/tool/fire.json
Maleclypse, line 164:
```suggestion
    "category": "container"
```
Resetting test suite

data/json/items/tool/fire.json
```diff
@@ -161,7 +161,7 @@
     "type": "TOOL",
     "name": { "str": "matchbook" },
     "description": "A small book of matches with a coarse strike surface on the outside flap for lighting them.  Allows you to use various drugs like cigarettes, light explosives like Molotov cocktails, or light nearby items on fire.  Due to the matches being lightly glued to the matchbook, they can be removed, but not replaced on the book.",
-    "category": "containers",
+    "category": "container"
```
Maleclypse, line 164:
```suggestion
    "category": "container",
```

PR #65442, opened 2023-05-02
Fix crash: dangling pointer to faction

hexagonrecursion:
#### Summary

Bugfixes "Fix crash when monster attacks a dead NPC"

#### Purpose of change

Fixes #65440

#### Describe the solution

Clear the dangling pointer. ~Note: I am not sure if this fix is ok because the faction may still exist if it is a templated base faction~

Edit: this forgets which faction the NPC was from even if the faction still exists afterwards (has other members or is a templated base faction)

#### Describe alternatives you've considered

move `fac->remove_member( getID() )` to  `game::cleanup_dead()`

https://github.com/CleverRaven/Cataclysm-DDA/blob/d53bbbeae0455f9de1a871c5a47e0503047caa41/src/game.cpp#L4615-L4619

#### Testing

I have spawned a dozen NPCs, removed all their items to avoid triggering #65451 and watched them be devoured by a hoard of zombies -> no crash

Issue #65440, opened 2023-05-02
Crash or freeze when monster attacks a dead NPC


hexagonrecursion:
### Describe the bug

My game crashed when zombies killed an NPC

SIGSEGV

```c++
(gdb) bt
#0  0x00007f8a952b84a0 in __memcmp_avx2_movbe () at /lib64/libc.so.6
#1  0x00007f8a9559246e in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::compare(char const*) const () at /lib64/libstdc++.so.6
#2  0x0000000001660dd2 in std::operator==<char, std::char_traits<char>, std::allocator<char> >(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, char const*)
    (__rhs=<optimized out>, __lhs=<error: Cannot access memory at address 0xffffffffffffffff>) at /usr/include/c++/12/bits/basic_string.h:3599
#3  faction::has_relationship(string_id<faction> const&, npc_factions::relationship) const (this=0x22cccd10, guy_id=..., flag=flag@entry=npc_factions::kill_on_sight) at src/faction.cpp:373
#4  0x0000000001e1a913 in npc::has_faction_relationship(Character const&, npc_factions::relationship) const (this=this@entry=0x40844d70, you=..., flag=flag@entry=npc_factions::kill_on_sight) at src/npc.cpp:2416
#5  0x0000000001e1c437 in npc::attitude_to(Creature const&) const (this=this@entry=0x40844d70, other=...) at src/npc.cpp:2571
#6  0x0000000001c30e78 in melee_actor::on_damage(monster&, Creature&, dealt_damage_instance&) const (this=this@entry=0x7975f00, z=..., target=..., dealt=...) at src/mattack_actors.cpp:675
#7  0x0000000001c31169 in bite_actor::on_damage(monster&, Creature&, dealt_damage_instance&) const (this=0x7975f00, z=..., target=..., dealt=...) at src/mattack_actors.cpp:725
#8  0x0000000001c3272d in melee_actor::call(monster&) const (this=0x7975f00, z=...) at src/mattack_actors.cpp:616
#9  0x0000000001d37f0d in monster::move() (this=this@entry=0x472a6490) at src/monmove.cpp:801
#10 0x00000000015cfc91 in (anonymous namespace)::monmove() () at src/do_turn.cpp:457
[...]

PR #65463, opened 2023-05-04
Backport to 0.G: Fix crash: dangling pointer to faction

hexagonrecursion:
#### Summary

Bugfixes "Fix crash when monster attacks a dead NPC"

#### Purpose of change

this is a backport of #65442
original bug: #65440

#### Testing

I have already tested the patch in #65442

### DIFF
--- a/src/npc.cpp
+++ b/src/npc.cpp
@@ -2994,6 +2994,7 @@ void npc::die( Creature *nkiller )
                 }
             }
             my_fac->remove_member( getID() );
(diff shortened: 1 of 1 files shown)


I-am-Erk:
I'm not sure why this was missed, but since we're working on H-stable now I think it's unfortunately probably out dated. Sorry!

hexagonrecursion:
I think I know why this was missed - we simply never had a 0.G **point** release - so this was backport never became relevant.

PR #65476, opened 2023-05-04
Creatures can stumble into invisible players to discover them

Rewryte:
#### Summary
Features "Creatures can stumble into invisible players to discover them"

#### Purpose of change
Gameplay with invisibility effects could be more engaging.  The idea is to introduce mechanics that make invisibility feel like an alternative and more powerful form of stealth, rather than just a "do not disturb" button that makes things ignore the player.  Currently creatures can track invisible characters using senses such as hearing and scent, but only use these to follow them around.  We can build on top of that.

#### Describe the solution
Creatures that attempt to walk into an invisible player's location will stumble into them.  For a brief moment, a stumbled creature is aware of the player's existence and may attempt to attack their last known location.  If the creature in that state successfully lands an attack on the player (not dodged), the tactile feedback will continue give away the presence of the player, refreshing the effect's duration.  Players can easily break detection by moving out of their known location, but are susceptible to subsequent stumbles.

Unless the player charges straight into an encounter, hostile creatures will have to first track the player to stumble into them.  Understanding the detection abilities of different creatures as well as the stealth capabilities of your character becomes key to remaining undetected.

#### Testing
Ruining invisibility is the last thing I want to do, therefore all changes are tested with and balanced around a character with Debug Invisibility, no other traits, 100 speed, and 8 in all stats.  The character should be able with survive nearly any hostile encounter with their invisibility and situational awareness alone.  The threat of detection should always be looming when moving through a populated area, but there should not be a scenario where the player would not be able to escape barring extraordinary circumstances.

#### Additional context
From playtesting early iterations of this feature, I found that certain creatures such as barghests were able to track and destroy an invisible character with surprising efficiency due to their speed and scent.  To counteract this, a modification was made to the scent movement code which causes scent tracking creatures to zigzag and "sniff around" while moving towards the player instead of beelining straight at them in full speed, giving the player a chance to escape.  A nice touch to this is that the distinctive movement pattern can be used to tell if a creature is tracking by scent.

Update: I realized that if the player has been standing around for a while, creatures without the KEENNOSE flag that are tracking them by scent will stop right in front of them and idle.  This is due to the smell_threshold mechanic disabling scent entirely when smell is too strong.  I tweaked this so that creatures will wander / sniff around when this happens, giving them a good chance to stumble into the player.

Update 2: Added support for NPCs.  Situations where an NPC would tell an invisible player to let them pass would result in a stumble instead.  There are some NPC behaviors which wrongly ignore invisibility that are not introduced by this PR.  They are outside the scope of this PR, but I have fixed at least one occurrence.


Rewryte:
Mostly working at this point, but I'm keeping it in draft for more testing.

ashGlaw:
Really cool idea! Big boost to verisimilitude

ashGlaw:
Really cool idea! Big boost to verisimilitude

PR #65477, opened 2023-05-04
Fix unvisited member reporting from global destructors in atexit.

akrieger:
#### Summary
None

#### Purpose of change
When globals get destructed in atexit, `std::uncaught_exceptions` returns false. This means the fix from #65363 is insufficient depending on the situation, such as, in test suites when loading initial game data and there's a units error.

#### Describe the solution
I googled but couldn't find another way to detect if we're post-main. So just set up a sentinel guard variable that when destructed or unwound will disable reporting json member errors.

- :fire:
- Not really an alterative, but potentially could dismiss the guard on 'normal' shutdowns to catch places where things aren't cleaned up appropriately. But I'm not sure if we care so much in those cases.

#### Testing
Replicated the error from https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4884099474/jobs/8717107644?pr=65474 locally, validated that with my fix the spam is fixed but the error is still reported.

### DIFF
--- a/src/flexbuffer_json.cpp
+++ b/src/flexbuffer_json.cpp
@@ -1,5 +1,6 @@
 #include "flexbuffer_json.h"
 
+#include <atomic>
 #include <cstring>
 #include <istream>
 #include <optional>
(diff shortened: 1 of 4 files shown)


bombasticSlacks:
Clang is mad at you

PR #65480, opened 2023-05-05
Allow comestibles to inflict multiple addictions

dseguin:
#### Summary
None

#### Purpose of change
Maleclypse asked if it was possible to assign multiple addictions to a single item. Now it is :D

#### Describe the solution
The old way of specifying addictions for items is still valid:
```JSON
{
  "type": "COMESTIBLE",
  ...
  "addiction_potential": 5,
  "addiction_type": "alcohol"
}
```

But you could also specify multiple addiction types:
```JSON
{
  "type": "COMESTIBLE",
  ...
  "addiction_potential": 5,
  "addiction_type": [ "alcohol", { "addiction": "caffeine", "potential": 3 } ]
}
```

`"addiction_type"` can accept an array that contains either an object that defines an addiction and its potential, or a string that uses the default value from `"addiction_potential"`.

#### Testing
Added a simple test case:
```
tests/cata_test -d yes --rng-seed time "check that items can inflict multiple addictions"
```


data/mods/TEST_DATA/items.json
```diff
@@ -1536,6 +1536,34 @@
     "vitamins": [ [ "calcium", 3 ], [ "iron", 4 ] ],
     "rot_spawn_chance": 50
   },
+  {
+    "type": "COMESTIBLE",
+    "id": "test_whiskey_caffenated",
+    "name": { "str_sp": "caffenated whiskey" },
```
TheShadowFerret, line 1542:
```suggestion
    "id": "test_whiskey_caffeinated",
    "name": { "str_sp": "caffeinated whiskey" },
```

dseguin: This would break the unit test since the id would need to change there as well. Not that this data would ever be visible outside of unit tests anyway :p

PR #65487, opened 2023-05-05
barrel length for mods

bombasticSlacks:
#### Summary
None 

#### Purpose of change
lets random mods override barrel length.

#### Describe the solution
Adds it as an option for mods.

Added the logic for the sawed off too!

#### Testing
Letting unit tests run.


src/item.cpp
```diff
@@ -6928,6 +6928,12 @@ units::mass item::weight( bool include_contents, bool integral ) const
     return ret;
 }
 
+units::length sawn_off_reduction( const itype *type )
```
bombasticSlacks, line 6931:
```suggestion
static units::length sawn_off_reduction( const itype *type )
```

PR #65494, opened 2023-05-05
update and modernize lsurvivor_armor

anoobindisguise:
#### Summary
Balance "update light survivor body armor"

#### Purpose of change
Light survivor body armor was another one of the weird and old survivor armors but it was bad enough that it was inoffensive and went ignored.

#### Describe the solution
It is now combination armor made of a combat shirt and a kevlar vest, similar to how the mercenary coat is a combo of an army jacket and a us ballistic vest. It has pockets granting a total of 2.01 liters of storage spread across the arms and torso. Torso coverage for kevlar is 95% and 90% for arms. 3.3mm of kevlar for the torso and 2.2mm for the arms. Also renamed to "light handmade body armor".

Less coverage - the kevlar vest itself has a little less coverage than that but there are extra kevlar sheet layers to make up the difference. 
Move it out of cuttingroom.
Might still have a few too many pockets for normal layer but I've made them all slow access to make up for their laying space and the pockets are all individually fairly small.

### DIFF
--- a/data/json/items/armor/bespoke_armor/cuttingroom.json
+++ b/data/json/items/armor/bespoke_armor/cuttingroom.json
@@ -319,7 +319,7 @@
           { "type": "cotton", "covered_by_mat": 100, "thickness": 0.15 },
           { "type": "kevlar", "covered_by_mat": 95, "thickness": 3.3 }
         ],
-        "encumbrance": [ 4, 5 ]
+        "encumbrance": [ 4, 5 ],
         "coverage": 100,
         "cover_vitals": 70,
         "covers": [ "torso" ]
@@ -330,7 +330,7 @@
           { "type": "cotton", "covered_by_mat": 100, "thickness": 0.15 },
           { "type": "kevlar", "covered_by_mat": 90, "thickness": 2.2 }
         ],
-        "encumbrance": [ 4, 5 ]
+        "encumbrance": [ 4, 5 ],
         "coverage": 100,
         "cover_vitals": 30,
         "covers": [ "arm_l", "arm_r" ]
--- a/data/json/items/armor/bespoke_armor/cuttingroom.json
+++ b/data/json/items/armor/bespoke_armor/cuttingroom.json
@@ -302,29 +302,94 @@
     "id": "lsurvivor_armor",
     "type": "ARMOR",
     "category": "armor",
-    "name": { "str": "light survivor body armor" },
-    "description": "Lightweight, custom-built body armor made from Kevlar and tough fabric.  Mostly waterproof.",
+    "name": { "str": "light handmade body armor" },
+    "description": "Custom-made armor that combines a level II vest with a combat shirt and adds extra padding to the arms for safety.  Fairly lightweight and ergonomic and has numerous pockets for storage.",
     "weight": "3000 g",
     "volume": "6 L",
     "price": 60000,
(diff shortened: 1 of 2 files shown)


fairyarmadillo:
I think stripping "survivor" out of the title of these items is making them less descriptive, not more. Previously it was understood that these were hand-assembled pieces of kit by an experienced survivor, thus the name. Plus we have other named sets like the nomad and scavenger gear, why is survivor gear the exception?

anoobindisguise:
> I think stripping "survivor" out of the title of these items is making them less descriptive, not more. Previously it was understood that these were hand-assembled pieces of kit by an experienced survivor, thus the name. Plus we have other named sets like the nomad, mercenary, and scavenger gear, why is survivor gear the exception?
> 
> If there's some pressing need to eliminate the survivor title, I think "handmade armor" might not be the way to go, as these are factory-made pieces that have simply been customized, and not handmade at all. I'd suggest "custom light body armor" or something like that instead.

I can change it back and I don't mind it at all myself but there are a decent chunk of the devbase who really hate the word survivor being used in various gear pieces so I'll wait for more opinions.

bombasticSlacks:
> I think stripping "survivor" out of the title of these items is making them less descriptive, not more. Previously it was understood that these were hand-assembled pieces of kit by an experienced survivor, thus the name. Plus we have other named sets like the nomad, mercenary, and scavenger gear, why is survivor gear the exception?

There is a real stigma with the word survivor that it is an item that provides all the utility you want, and perfect defence. Since none of the items that originated as survivor items now do that, it makes sense to get rid of the qualifier. Anything else is being deceptive to players, setting them up with expectations that the core game no longer fills with any gear, especially hand made.

I'm sure you could get a dev on the discord to rant about how it's dumb in general, but I think that is a better way to frame it. Even though there are spiritual successors to some classic "survivor" items, the fundamental "survivor gear" power fantasy is dead, so there is no reason to trick people with a label that is no longer true.

I like handmade for this, but it could go under the mercenary classification, since it is other modified soldier gear. Custom would also be fine. I'm fine with merging this, so that means it should be taken out of cutting room otherwise someone is going to come along and delete it without consideration.

Zireael07:
> Anything else is being deceptive to players, setting them up with expectations

This is only true for players who played with those items pre survivor gear nerf. New players or those that never got around to survivor gear won't have such expectations. Hence I believe the premise is false

bombasticSlacks:
> This is only true for players who played with those items pre survivor gear nerf. 

They were nerfed with the name change iirc.

anoobindisguise:
> Since you are moving anyway, it would be worth considering fabric thicknesses for this PR. Your nylon and Kevlar I don't think are proper sheet multiples. It isn't enforced yet, but it would be worth considering 😄

Kevlar is the correct thickness IIRC. One sheet is 1/16th of 4.4mm (0.275mm) while Nylon is abouuuuut 0.25mm per my aketon/gambeson work and cotton is abouuuut 0.125mm (half the gsm of canvas which is 0.25mm) but presumably the combat shirt is some kind of nylon/cotton blend at a weird weighting (0,3mm total)
the extra nylon thickness on torso is coming from the kevlar vest itself.

ProjectileWaterfowl:
It might be a good idea to extend the changes to the XL and XS variants.

PR #65505, opened 2023-05-06
Initial Implementation Of Replacable Limbs

bombasticSlacks:
#### Summary
None

#### Purpose of change
#51294

#### Describe the solution
When a creature's body is checked for an id, the game will attempt to find the next best thing if it doesn't have the exact request.

#### Testing
Added big head mode, this probably breaks a pile of things still needs testing.

#### Additional context
The primary benefit of this system is it solves most of our problems while letting any individual system do more accurate simulation later on. However, for now, things should still for the most part make some sense. As an example if you were to step in acid and that should deal damage hard coded to the "FOOT_L" and "FOOT_R" well this will try to find the thing you are using as feet and damage it in its place.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1992,7 +1992,7 @@ bool Creature::has_part( const bodypart_id &id ) const
 
 bodypart *Creature::get_part( const bodypart_id &id )
 {
-    auto found = body.find( id.id() );
+    auto found = body.find( get_part_id( id ).id() );
     if( found == body.end() ) {
         debugmsg( "Could not find bodypart %s in %s's body", id.id().c_str(), get_name() );
         return nullptr;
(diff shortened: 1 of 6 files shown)


RenechCDDA:
image

src/creature.cpp
```diff
@@ -2045,6 +2045,40 @@ static void set_part_helper( Creature &c, const bodypart_id &id,
     }
 }
 
+bodypart_id Creature::get_part_id( const bodypart_id &id ) const
+{
+    auto found = body.find( id.id() );
+    if( found == body.end() ) {
+        // try to find an equivalent part in the body map
+        for( const auto &bp : body ) {
```
Fris0uman, line 2053: Are those types too complexe to be explicited?

Fris0uman: I think in general we avoid using auto as much as possible, usually makes the code easier to read

PR #65506, opened 2023-05-06
Add new holy symbol mission, minor technical improvements to other missions

NetSysFire:
#### Summary
Balance "Add a more accessible holy symbol mission, replacing the small relic one"

#### Purpose of change

Its a bit weird for a first mission. Also I made cathedrals more rare in #65366.

#### Describe the solution

New mission, copied from the small_relic one.

Also some misc improvements to other missions. No reason why the inhaler one had the MGOAL_CONDITION when MGOAL_FIND_ITEM is perfectly fine.

American flags are generic default gear a NPC can spawn with, so remove it when the NPC already has it.

This mission is probably vastly easier than others, admittedly. Please tell me your thoughts.

#### Testing

Loads and works properly.

image

image

### DIFF
--- a/data/json/npcs/missiondef.json
+++ b/data/json/npcs/missiondef.json
@@ -34,8 +34,8 @@
     "id": "MISSION_GET_INHALER",
     "type": "mission_definition",
     "name": { "str": "Find Inhaler" },
-    "goal": "MGOAL_CONDITION",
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):
Looks good!

Issue #58808, opened 2022-06-29
Collapsed baton disappears from inventory after loading save


jeanpaulrsoucy:
### Describe the bug

I was wielding a baton (extended) when I decided to wield another weapon. When prompted, I opted to place the baton (extended) into my inventory. I then activated the baton to collapse it. Later, I saved and reloaded the game and was shown the following error:

> DEBUG    : error: item baton cannot fit into pocket while loading: item is too short
>
> FUNCTION : void item_contents::combine(const item_contents&, bool)
> FILE     : src/item_contents.cpp
> LINE     : 599
> VERSION  : 80d1bee

After dismissing the error, the baton had disappeared from my inventory.

### Steps to reproduce

Load the attached save for Glinda 'Sixes' Moody.

More generally, it seems the error occurs when an item changes length after being inserted into a pocket.

### Expected behavior

Baton does not disappear from inventory.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19043.1706 (21H1)
- Game Version: 80d1bee [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

### Additional context

[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/9006925/debug.log)
[Pendleton Center.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/9006933/Pendleton.Center.zip)

PatrikLundell:
/confirmed

I believe similar issues have been reported for modding holstered weapons.

Move your character one step and the baton falls to the ground, so at least it isn't lost.

I-am-Erk:
as long as the items are falling to the ground and not being lost entirely, I think this can be deferred as a blocker. As with all deferred blockers, I still hope it gets fixed before stable, but I think there are some complex UI questions there. At the very least, it would be nice to have a warning prompt or an alert, as mentioned in #51457.

jeanpaulrsoucy:
This is a super old save, so the characteristics of the items in question have greatly changed, but I can confirm that the overflow functionality seems to be working as intended.

![overflow](https://github.com/CleverRaven/Cataclysm-DDA/assets/31170087/c6566b21-a103-426a-87b9-cc810cb51858)

[...]

PR #65518, opened 2023-05-07
Pockets overflow into their parent pockets if possible

mqrause:
#### Summary
Features "Pockets overflow into their parent pockets if possible"

#### Purpose of change

More working towards #60885. The migration makes this necessary so your foodstuffs won't be dropped everywhere, even out of sealed containers.

#### Describe the solution

Try to find a parent pocket the can hold the item recursively. Print a message of what moved to where to inform players something unintentional happened.

Could move the item to any pocket of the parents or even any pocket on the character (if it's on the character), but this should be a fairly rare occurence anyway, so no need to make it more complex than necessary.
This also doesn't account for possible partial overflow of stacked items, because #60885.

#### Testing

Added a simple test case. Then also tested it using an existing bug (#58808):
1. Spawn golf bag, fast draw holster and expandable baton
2. Wear golf bag
3. Organize it as golf bag > fast draw holster > expandable baton
4. Expand the baton (becomes too big for holster but stays there)
5. Wield golf bag to trigger inventory sanity checks (no to drawing from holster)
6. Baton drops into golf bag instead of the ground

grafik

#### Additional context

### DIFF
--- a/src/character_attire.cpp
+++ b/src/character_attire.cpp
@@ -1992,10 +1992,10 @@ void outfit::best_pocket( Character &guy, const item &it, const item *avoid,
     }
 }
 
-void outfit::overflow( const tripoint &pos )
+void outfit::overflow( Character &guy )
 {
-    for( item &clothing : worn ) {
-        clothing.overflow( pos );
+    for( item_location &clothing : top_items_loc( guy ) ) {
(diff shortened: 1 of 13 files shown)


AmarinReyny:
Does this affect liquids at all?

PR #65519, opened 2023-05-07
Allow mutants to walk underwater

worm-girl:
#### Summary
Features "Allow mutants to walk underwater"

#### Purpose of change

The new crustacean mutants were having a hard time in water because their shells are so heavy, often burning two whole bars of stamina moving a single tile. This change allows post-thresh crustaceans, cephalopods, and snails to walk along lake beds and other underwater surfaces without burning as much stamina as swimming, though still more than they burn on land.

#### Describe the solution

* Added CRAB_LEGS as a pre-threshold mutation. Currently this doesn't do much except make the mutant quieter, but it could have all kinds of effects, especially once wounds and better mutation limbification are in place.

* Added a new post-threshold mutation for crustaceans called SCUTTLE. Scuttle slightly reduces move cost over obstacles, as many crabs are good climbers.

* Added a flag to scuttle, tentacle bracing, and gastropod foot called WALK_UNDERWATER. Then added a check for it in burn_move_stamina. If the player has this flag and is standing in a tile that has the DEEP_WATER flag but not the GOES_DOWN flag, then they are exempted from the usual extra stamina cost that is normally applied by that function.

This does not make these mutants any faster underwater, but it does make it more feasible for them to hang out down there. This does not do anything if the mutant's feet can't touch the bottom, and it doesn't exempt them from other stamina costs associated with being in water. Walking around underwater will still drain stamina faster than it would on land, but not nearly as fast as swimming.

Other traits or character properties could be added to can_walk_underwater if anyone can think of them. Possibilities include a wearing a dive suit, some sort of ultraheavy bionic, or maybe even power armor.

I thought about removing athletics learning for walking around on the ocean floor, as well as blocking bonuses from items like swimming gloves (none of the applicable mutants can wear swim fins), but I don't think that's super necessary. Someone with human hands and crab legs could reasonably still get some benefit from wearing swimming gloves while moving along the ocean floor as they'd still be useful for "hopping" along, and a person who is moving around in water at all is dealing with things like current and water resistance, so the skill training doesn't feel too weird, especially as it's still burning more stamina than land walking, implying at least some physical exertion is taking place.

#### Testing

Mutated crustacean_carapace and tried walking around on a lake bed. Gave myself scuttle, tried again, and observed that stamina drain was much lower. Tried swimming above the lake bed and saw that stamina drain was appropriately not reduced.

Walked across a riverbed with scuttle and observed reduced stamina drain.

Repeated both tests with tentacle bracing and gastropod balance.

#### Additional context

image
A crustacean mutant doing battle with a giant carp at the bottom of a lake.

image
mr_krabs_walking_10hours.mp4


Venera3:
Instead of id-checking traits use a character flag please - that opens it for e.g. effects or other JSON bits to use the functionality down the line.

fairyarmadillo:
Oh cool, the tests passed this time! It's running on a JSON flag now instead of ID checking, as Venera3 requested, and it should be really easy to put the flag on artifacts/magiclysm spells/diving suits in the future.

PR #65528, opened 2023-05-07
Qiang Spear Strap Fix

NotDucc:
#### Summary
Balance "Qiang Spear Strap Fix"

#### Purpose of change
Currently the qiang, an ancient chinese spear, doesn't fit inside spear straps despite having the flag for it due to item length. This PR fixes that.

#### Describe the solution
The previous maximum item length for the spear strap was 200cm, while the qiang is 250cm long. The solution I used was increasing the maximum item length for the spear strap to 250cm.

None.

#### Testing
I tested after making the line change, now it fits the qiang perfectly fine.

#### Additional context
Screenshot (1026)


PatrikLundell:
Would that work in reality? The problem I suspect is that a spear that long would get caught in door openings, hit the ceiling when inside, catch in tree branches, etc.

fairyarmadillo:
> Would that work in reality? The problem I suspect is that a spear that long would get caught in door openings, hit the ceiling when inside, catch in tree branches, etc.

"According to the design of the 14th year of Qianlong.

Regulations of the dynasty; 
Tiger Spear Division's Tiger Spear: Spearhead made of forged steel. Overall 8 chi 3 cun long. Spearhead is 9 cun long, with triangular tip and center ridge. Shaft is 7 chi 4 cun long, made of white waxwood. At the top of the shaft are tied two transverse pieces of deer antler, 1 cun long. There is a horn ferrule at the end of the shaft. The spearhead is covered with a leather bag, wrapped with birch bark. It is carried by tying it to a leather belt.

Converted in cm
Overall: 290.5 cm
Head: 3.5 cm
Shaft: 259 cm"

source - https://www.mandarinmansion.com/article/spears-qing-dynasty

You could just angle it to get it through a doorway, it's not like a spear strap can't be adjusted. It's a fair point that the spear might be too big to comfortably stand indoors, but that problem really goes beyond one piece of gear - we have mech suits, horses, and all kinds of equipment and mutations that should have trouble in roofed areas, but no system to penalize or prevent that. I think this should be allowed and an issue created for checking character heights (including worn inventory length) when indoors and possibly forcing them to crouch to move or something, depending on the terrain type.

Zireael07:
Whoaaa... 3m long. Strap sounds like a requirement for carrying this in any sort of a way, and I'm guessing the historical users just did NOT use it indoors...

fairyarmadillo:
> Whoaaa... 3m long. Strap sounds like a requirement for carrying this in any sort of a way, and I'm guessing the historical users just did NOT use it indoors...

Our qiang is 250cm long, which I believe is the size used by modern day martial artists. I was just sharing the site because it had some interesting data.

lcy03406:
Qiang is the word of spear in chinese, its length varies from 2m to 6m, with specific names. Many types of infantry qiang have a soft wooden shaft so that the steel point can seize an area around the middle line in front of user.

PR #65531, opened 2023-05-07
Duct tape rework

Karol1223:
#### Summary
None

#### Purpose of change
I really hated how duct tape spawned around in free-floating strips.

#### Describe the solution
- Duct tape has had a thorough weight and volume rebalance to fix the density
- Price adjusted to what I found on amazon
- Added a duct tape roll item on which duct tape will now spawn. It is non-reloadable and non-unloadable.
- Hopefully duct tape for sales in bike stores should no longer be used
- Reworked the duct tape recipe, described more in detail below:

**Duct tape recipe details**
- Will now use rubber cement in place of superglue
- Will now use cotton sheets or patchwork cotton sheets in place of cotton patches
- Will now use plastic sheets in place of plastic chunks (shockingly, one regular plastic sheet is enough for the recipe)
- Can no longer be learned from mechanic textbooks
- No longer uses filament requirement group
- Uses fabrication skill and not chemistry
- Makes 40 charges instead of a full roll
- Recipe uses the assumption that since the roll of duct tape the item is now based off is 2 inches in width, the cotton sheet used (12x24 inches per JSON comment) is cut into 6 2x24 inch strips resulting in total 4 yards of length (1 charge = 0.1yard; 4 yards = 40 charges). This equals to about 0.186m^2, and with the assumption that a plastic sheet is 0.2m^2 in surface (because large plastic sheets are 1.6m^2 and they disassemble into 8 normal plastic sheets) this actually means only one plastic sheet is necessary. In fact, some of the plastic would be wasted here.
- **NOTE: AMOUNT OF RUBBER CEMENT AND THE ENERGY IT TAKES IS UNCHANGED AS I DO NOT KNOW HOW TO VERIFY THESE NUMBERS!!!**

I really wanted to make it so hardware stores spawned only full rolls of duct tape but it's a matter of them using weird itemgroups so it's beyond my paygrade

#### Testing
Checked that it spawns properly and checked that it can indeed be used while on the roll


data/json/items/containers.json
```diff
@@ -3500,5 +3500,21 @@
       }
     ],
     "flags": [ "COLLAPSE_CONTENTS" ]
+  },
+  {
+    "id": "duct_tape_roll",
+    "type": "GENERIC",
+    "category": "container",
+    "name": { "str": "duct tape roll" },
+    "description": "A small cardboard roll with duct tape rolled around it.",
```
casswedson, line 3509: I would make this a generic "tape roll" "cardboard roll", not sure how you refer to those in english, that way it'd be easy to swap the contents for medical tape, toilet paper etc

also the `other` category would fit better

Karol1223: To be fair - you're right! On both accounts. I was very close to naming it just a cardboard roll but I was worried that this way people might get confused. Though if we are going to reuse it, as we should admittedly, then this is the right approach. I'll correct those tomorrow once the tests inevitably yell at me for forgetting to remove duct tape from the known bad density list.

avianArsonist:
removing the crafting recipe seems like it would break a lot of Innawood crafting

Karol1223:
> removing the crafting recipe seems like it would break a lot of Innawood crafting

I'm not convinced so. How the hell Innawood gets superglue?

TheShadowFerret:
> > removing the crafting recipe seems like it would break a lot of Innawood crafting
> 
> I'm not convinced so. How the hell Innawood gets superglue?

Put it on autolearn like every other chem recipe when Innawoods is activated?

kilozombie:
>Removed the crafting recipe for duct tape because it was a bunch of bullshit. Plus... if you have superglue to make it, why would you make it? These two are interchangeable. It's not like you're in dire need of a duct tape wallet

This isn't really cogent. Unless you really think making makeshift tape is impossible, this is removing some key content for both innawoods and normal players. Duct tape is ammo for misc. repair kits and can be subdivided in recipes easier than superglue if you just need a little bit. If you hate this recipe, a real solution would be a complete refactor of superglue into an object of large quantities like duct tape currently is, and put it into bottles/containers.

Also... innawoods players can easily get superglue with the recipe to make superglue. Fire, water and bones are very common innawoods things.

kilozombie:
Also, speaking of misc. repair kits, how are you gonna reload them now?

Karol1223:
Ah true, I forgot about the misc repair kits. To be fair, they shouldn't work as they do anyway but I can't be asked to rework them right now if I can avoid it. I'll test if this still works later today.

For Innawood, people seem to think Innawood is as important as mainline. Which it is not. A mod should be additional content, not drag mainline behind. If all it'll take is giving Innawood players back the weird recipe, like I did with antennas a while back, that's doable. But I sincerely hope that's all it takes.

Qrox:
> Removed the crafting recipe for duct tape because it was a bunch of bullshit

I think a better solution than this is required. The current recipe is not ideal but it has been a good enough abstraction of what one would go about making duct tape substitute in real life, and many things in game probably do not require something as strong as duct tape either.

If we want to improve realism though, according to [Wikipedia](https://en.wikipedia.org/wiki/Chemistry_of_pressure-sensitive_adhesives#Composition),

> A typical PSA tape consists of a pressure-sensitive adhesive (the sticky part of the tape) coated to a backing material. To prevent the adhesive from sticking to the backing when wound in a roll, a [release agent](https://en.wikipedia.org/wiki/Release_agent) is applied to the backing or a [release liner](https://en.wikipedia.org/wiki/Release_liner) is placed on the adhesive. Sometimes a primer is coated between the adhesive and backing increasing the bond.

But some of these requirements could probably be relaxed a bit for homemade "duct tape", such as the pressure sensitivity and the release agent (we can use a release liner instead). That leaves us with a backing, a strong enough adhesive, and a release liner (or simply a surface that the adhesive does not stick to).

The superglue recipe in CDDA uses bones, hide, and slime blobs among other things, and we should probably check if adhesives made from some of these things can be made strong enough to be used for duct tape. Slime blob though could be handwaved as being strong enough itself to be used unaltered for duct tape.

I would also suggest increasing duct tape spawn because the current spawn rate seems to be a bit low.

> if you have superglue to make it, why would you make it?

In addition to how duct tape is used more frequently in game, in real life it is sometimes easier to use tape instead of glue, obviously.

kilozombie:
> For Innawood, people seem to think Innawood is as important as mainline. . .

Let me clarify. Like Qrox says, duct tape is sometimes just more helpful than superglue because you can use it in smaller quantities, and it can be used to very quickly bind two things. They are not equivalent things.

Karol1223:
> But some of these requirements could probably be relaxed a bit for homemade "duct tape", such as the pressure sensitivity and the release agent (we can use a release liner instead)

I would need a source of people actually making homemade duct tape like this, in this case. Because handwaving is something I am not going to do.

I am not against buffing spawns, though. As it stands duct tape is weirdly rare.

fairyarmadillo:
> The superglue recipe in CDDA uses bones, hide, and slime blobs among other things, and we should probably check if adhesives made from some of these things can be made strong enough to be used for duct tape. Slime blob though could be handwaved as being strong enough itself to be used unaltered for duct tape.

Superglue is not really interchangeable with the rubber adhesives used in duct tape. Superglue is an acrylic resin that cures almost instantly into a hard, crusty substance with zero flexibility. Duct tape glue basically never fully cures and this is important for it to remain flexible.

Bone glue is not interchangeable with superglue IRL. Bone glue is water soluble and not all that strong, and it's really only in use in places where it's anticipated that it will at some point need to be removed, such as in art restoration and constructing wooden instruments.

Per wikipedia: 

> During World War 2, Revolite, (then a division of Johnson & Johnson) developed an adhesive tape made from a rubber-based adhesive applied to a durable duck cloth backing. This tape resisted water and was used to seal some ammunition cases during that period.

I think the most appropriate recipe for duct tape would need a rubber glue, which we don't have yet. This could probably be made out of rubber chunks and solvent with some kind of chemical process that I suspect wouldn't be too difficult for a crafty survivor, then the recipe would call for that, plus canvas or cotton, a cutting tool, and some plastic sheets.

There's a video about the process here - https://www.youtube.com/watch?v=fVpNteozrO0

Karol1223:
Yep I've looked a bit into this more since my initial response and making homemade duct tape is possible but it'd call for some items we don't yet have. Personally I am of the opinion that if you have the tools, resources and time to make your own resources for duct tape you're better off just looting duct tape - so instead of adding the intermediate ingredients I'll boost duct tape spawnrates. It's a more sensible approach.

Having said that, I'd like to ask @Light-Wave for their opinions as to how this problem should be resolved within Innawood. Do you want me to throw you the unrealistic recipe?

Zireael07:
My feeling is that the superglue=boneglue thing as well as using either for duct tape is a remnant of the period when the game was much less realistic, and chemistry was an afterthought slapped onto cooking

Karol1223:
> Also, speaking of misc. repair kits, how are you gonna reload them now?

In a shocking twist I can confirm this does work.

Granted... it probably *shouldn't* - but apparently reloading with something from a container with ``NO_UNLOAD`` ignores that flag

Light-Wave:
Duct tape is needed for late-game Innawood progression. Specifically to craft the soldering iron and makeshift arc welder. Although there might be other recipes that requires duct tape as well.

I do like the idea of a more realistic recipe for duct tape. If you think it would be reasonable to update the recipe in a way that makes it still be doable in Innawood without electric equipment, that would be nice. I think it makes sense to have this crafting recipes in mainline even when it would be easier to loot the items in question.

Alternatively, the electric equipment that needs to be crafted with duct tape could have their recipes reworked. You shouldn't need them to be very flexible, so maybe superglue could be used here as an alternative. Although going this route would risk missing some important recipe.

If you don't wanna do either of that, I'm perfectly fine with you moving the old recipe to Innawood and leave it at that.
Either way, thanks for the consideration.

Karol1223:
> I do like the idea of a more realistic recipe for duct tape. If you think it would be reasonable to update the recipe in a way that makes it still be doable in Innawood without electric equipment, that would be nice.

It'd require a recipe to craft rubber glue, which in turn would require rubber. Is rubber obtainable in Innawood? Because if it is this may be possible but otherwise I doubt it'd work.

Karol1223:
> rubber glue

Does Innawood add it's own? Because that's not a vanilla item.

Either way, if these are obtainable in Innawood the recipe might be possible. And at that point, I'll be adding it to the mainline, too. It'll take a while before I get the time and energy to make it all work, though. New recipes are not exactly my strong suite.

Light-Wave:
Sorry, I was thinking about rubber cement.

Maddremor:
Speaking from memories of past conversations here, but I do recall people coming to the conclusion that while it might not be duct tape proper, textile with glue still does the job a lot of the times. Even if you make actual duct tape non-craftable, consider adding some kind of adhesive patch to fill the same niche. This could be useful for the fact that glue come in larger discrete units compared to the more granular duct tape. (Although perhaps you would prefer to make glue more granular too.)

RoyalFox2140:
There is a Youtube channel that does many glue tests you could easily re-create at home and at one point he makes rawhide glue with a 7 hour process, if you are worried about the strength of homemade glue to commercially available glue it may be relevant, it's a process of boiling down cut pieces of rawhide (dog bones in this case) to a paste. You would need a crane scale to re-create strength tests. Water solubility is a whole other issue than strength, a separate video tested that Elmer's glue was very strong but is known for being water soluble.

https://www.youtube.com/watch?v=sBi5KQ72k68

Karol1223:
> Even if you make actual duct tape non-craftable, consider adding some kind of adhesive patch to fill the same niche.

That's probably a better solution, true. Also would save me a lot of headache

When it comes to making glue more granular - this needs to happen at some point but I'm not feeling it. I lack the time nowadays to make such big projects. One day, though, hopefully.

fairyarmadillo:
> There is a Youtube channel that does many glue tests you could easily re-create at home and at one point he makes rawhide glue with a 7 hour process, if you are worried about the strength of homemade glue to commercially available glue it may be relevant

Nobody's arguing for the removal of bone/hide glue, it's just not an appropriate material for making duct tape as it fully cures and is water-soluble, two things that duct tape glue needs to not do.

Karol1223:
I'll come back to this PR in the second half of June if all goes well. It needs more time and research than I have to spare currently, unfortunately. I'll leave it up as a draft and hope with my whole heart that stalebot doesn't close it in the meantime.

Karol1223:
I'm closing this until I can actually return so I can reopen it myself when I'm back. If stalebot does it I won't be able to...

randomtyper:
> I think the most appropriate recipe for duct tape would need a rubber glue, which we don't have yet. This could probably be made out of rubber chunks and solvent with some kind of chemical process that I suspect wouldn't be too difficult for a crafty survivor, then the recipe would call for that, plus canvas or cotton, a cutting tool, and some plastic sheets.

We got rubber cement now, I think the duct tape recipe should use it instead of being thanos'd.

Karol1223:
> We got rubber cement now, I think the duct tape recipe could use it instead of just being thanos'd.

We've had it for years. This doesn't change much. Rubber glue =/= rubber cement, but I'll look into it specifically.

But for the love of god, I get it - I've had those suggestions for months here. Let me at least rebase.

Karol1223:
I believe it's in as good of a state as I can get it to.

After more review - the recipe stays but is heavily realistified (nerfed) as it no longer will take superglue, patches and plastic chunks - instead relying on rubber cement, cotton sheets and plastic sheets. It also will use fabrication instead of chemistry, won't be learnable via the mechanic textbook and produce significantly less... but hey, it won't use filament, so basically a win.

PR #65533, opened 2023-05-07
Smart watch

Ruminator24:
#### Summary 

Content "This adds a smart watch as an item."

#### Purpose of change 

Smart watches exist in real life.

#### Describe the solution 

This adds a smartwatch as an item, as well to the loot pool.

For some reason the game watch has a higher chance of being spawned, which doesn't seem right. I would edit it, but I believe someone else is already going to rebalance the chances anyways.

#### Testing

I did the 100 test, and have confirmed that it will spawn, and checked to see if it works properly.


data/json/itemgroups/Clothing_Gear/clothing.json
```diff
@@ -560,6 +560,7 @@
       { "item": "diving_watch", "prob": 20 },
       { "item": "game_watch", "prob": 20, "charges": [ 0, 50 ] },
       { "item": "fitness_band", "prob": 10 }
```
GuardianDll, line 562:
```suggestion
      { "item": "fitness_band", "prob": 10 },
```

PR #65536, opened 2023-05-07
Audit burn_data for vanilla materials

RenechCDDA:
#### Summary
Balance "Most materials now burn at least a little slower than gunpowder"

#### Purpose of change
Pretty much the entirety of our materials' burn_data entries were just copy-pasted (cargo culted) from one contributor to the next.  Most items are consumed by small fires in *seconds*, and the amount of fuel added to the fire would be very high, even for non-combustible materials.

#### Describe the solution
Go through materials.json. Do my best to research the relative flammability of a given material. Add comments. Comments comments comments. Burn everything. Burn it again to be sure. Remove burn_products from a bunch of materials which... literally don't burn and can't produce them.

I probably could have done something productive with my time instead...

#### Testing
I created a 1L, 1kg solid debug item and changed its material around before loading in to a preset testing setup and lighting fires then watching how well the objects burned. The testing setup consisted of 2x fire rings and 2x empty grass tiles, each of the 4 burn tiles had 1 tinder on them (required to start the fire) and was lit by a firestarter CBM. The player character was in the center of the burn tiles, which were located at the 7/9/3/1 positions on the numpad. No burn tiles were adjacent as that could skew the results.

For materials whose 'burn curve' varied significantly with the intensity of the fire, tests were repeated with the amount of tinder increased to 10000 units.

I repeated these tests for pretty much every material I modified except the copy-pastes (like the 6 different forms of chitin). For the liquid-expected materials (alcohol, hydrocarbons, etc) I duplicated the debug item into a liquid form and gave it a charges/stack_size of 1

#### Additional context
Wood was largely used a baseline and expressly was not touched.

This is probably going to have some knock-on effects and I have no doubt someone out there will be very upset you can no longer recreate dante's inferno by dropping a match onto a pile of zombie corpses and their loot.

Hats off to rubber for having actually tested their data before implementing. Looks like it works about how you wanted.

### DIFF
--- a/data/json/materials.json
+++ b/data/json/materials.json
@@ -1055,11 +1055,11 @@
     "bash_dmg_verb": "bruised",
     "cut_dmg_verb": "sliced",
     "vitamins": [ [ "calcium", 1 ], [ "iron", 1.3 ] ],
-    "//": "Copied from chitin",
+    "//": "FIXME if you see me",
     "burn_data": [
-      { "fuel": 0, "smoke": 0, "burn": 0.0001 },
-      { "fuel": 0.2, "smoke": 3, "burn": 0.005, "volume_per_turn": "200 ml" },
-      { "fuel": 0.35, "smoke": 5, "burn": 0.015, "volume_per_turn": "200 ml" }
+      { "fuel": 0, "smoke": 1, "burn": 0.001, "volume_per_turn": "200 ml" },
+      { "fuel": 0.2, "smoke": 3, "burn": 0.002, "volume_per_turn": "500 ml" },
+      { "fuel": 0.35, "smoke": 5, "burn": 0.004, "volume_per_turn": "1 L" }
     ],
     "burn_products": [ [ "corpse_ash", 0.035 ] ],
     "resist": { "bash": 1, "cut": 1, "acid": 1, "heat": 1, "bullet": 1 }
@@ -1265,7 +1265,7 @@
     "dmg_adj": [ "ripped", "torn", "shredded", "tattered" ],
     "bash_dmg_verb": "ripped",
     "cut_dmg_verb": "cut",
-    "//1": "Nylon/polyester react poorly to flame(they melt) but are hard to light. Doesn't break down easily unless exposed to a high-intensity fire.",
+    "//1": "Nylon/polyester reacts poorly to flame (it melts) but is hard to light. Doesn't break down easily unless exposed to a high-intensity fire.",
     "burn_data": [
       { "fuel": 0.05, "smoke": 1, "burn": 0.01 },
       { "fuel": 0.05, "smoke": 2, "burn": 0.03 },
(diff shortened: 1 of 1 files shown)


data/json/materials.json
```diff
@@ -1966,10 +1980,11 @@
     "dmg_adj": [ "ripped", "torn", "shredded", "tattered" ],
     "bash_dmg_verb": "ripped",
     "cut_dmg_verb": "cut",
+    "//1": "Nylon reacts poorly to flame(it melts) but is hard to light. Doesn't break down easily unless exposed to a high-intensity fire.",
```
NetSysFire, line 1983:
```suggestion
    "//1": "Nylon reacts poorly to flame (it melts) but is hard to light. Doesn't break down easily unless exposed to a high-intensity fire.",
```

data/json/materials.json
```diff
@@ -2226,7 +2246,13 @@
     "bash_dmg_verb": "squished",
     "cut_dmg_verb": "cut",
     "vitamins": [ [ "iron", 0.1 ] ],
-    "burn_data": [ { "fuel": 1, "smoke": 2, "burn": 1 }, { "fuel": 1, "smoke": 2, "burn": 2 }, { "fuel": 1, "smoke": 1, "burn": 3 } ],
+    "//": "This burn data assumes wheat and chaff. In this form, it burns pretty well.",
```
NetSysFire, line 2249: The wheat material is primarily used for food, be it toast ems or spaghetti.

NetSysFire (approved):
That are some badly needed changes. I'd have loved to see your sources as links somewhere but keeping track of all of them is pretty hard and some of them are implied anyways.

Karol1223:
I feel like removing burn items from metals may be a bad idea. As it stands, burning something which is part metal part plastic will result in the plastic burning off and the metal parts turning into scrap metal. Are you sure removing that is a good idea?

RenechCDDA:
>I feel like removing burn items from metals may be a bad idea. As it stands, burning something which is part metal part plastic will result in the plastic burning off and the metal parts turning into scrap metal. Are you sure removing that is a good idea?

They currently do not actually do anything. The item made of a metal material just burns and... disappears eventually (fires will destroy items even without burn data, *eventually*).

I am not opposed to metals breaking down into smaller scrap pieces (possibly with some slag) when burned but to do that they need actual burn data and, apparently, *to count as fuel*. Which... well metal doesn't really burn. Usually.

data/json/materials.json
RenechCDDA: Thanks for checking this one. Now that I think of it, harvested wheat would come with piles of straw as byproducts too. So we can safely drop this one down to essentially 0.

RenechCDDA:
Ok, following the conversation with karol (and a bunch more testing, using multi-material test items and upping the size and weight to 10kg) I've readded most of the burn products. Some nonsensical ones (like superalloy *unfolding itself into perfect sheets*) were still removed. Of the remaining ones, I've dropped everything down to a maximum return rate of 0.5 (down from 1 in many cases), to account for everything from boil off to slagging your metals (depending on the material). You've not getting perfectly divided specks of pure gold with 0 material loss by throwing a gold object into a raging inferno.

Charcoal and especially coal still seem to be underperforming given the values I've given them, and I'm not sure why. Still, previously they were copying from wood so even a little better is an improvement.

PR #65541, opened 2023-05-08
Give oxy torch Metal Cutting 5

sam-trafton-dev:
#### Summary
Balance "Makes the oxy torch behave more like it would in real life"

### DIFF
--- a/data/json/items/tool/workshop.json
+++ b/data/json/items/tool/workshop.json
@@ -744,7 +744,7 @@
     "ammo": [ "weldgas" ],
     "charges_per_use": 4,
     "use_action": [ "OXYTORCH" ],
-    "charged_qualities": [ [ "WELD", 2 ], [ "SAW_M", 5 ] ]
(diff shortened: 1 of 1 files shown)


Maleclypse:
You should test this change.  I tested it just now and it overrides the oxytorch ability to cut through doors and I was unable to cut through metal bar doors with an entire 1400 charges.

LITFAMWOKE93:
> You should test this change. I tested it just now and it overrides the oxytorch ability to cut through doors and I was unable to cut through metal bar doors with an entire 1400 charges.

Thank you. I'm super new to this but I'm trying to contribute haha. I'll run through it again and see why the override is happening.

Maleclypse:
> > You should test this change. I tested it just now and it overrides the oxytorch ability to cut through doors and I was unable to cut through metal bar doors with an entire 1400 charges.
> 
> Thank you. I'm super new to this but I'm trying to contribute haha. I'll run through it again and see why the override is happening.

:) Always glad to see new people.

Issue #65449, opened 2023-05-03
GCC13 build error: possibly dangling reference to a temporary


tim110011:
### Describe the bug

On fedora 38, build CDDA using GCC13 failed, clang works.
Make flags:
    -j8
    CLANG=0
    CCACHE=1
    RELEASE=1
    LTO=1
    TILES=1
    SOUND=1
    LOCALIZE=1
    LANGUAGES=zh_CN
    DYNAMIC_LINKING=0
    RUNTESTS=0
    -march=native

### Attach save file

src/action.cpp: In function ‘bool can_examine_at(const tripoint&, bool)’:
src/action.cpp:673:19: 错误：possibly dangling reference to a temporary [-Werror=dangling-reference]
  673 |     const furn_t &xfurn_t = here.furn( p ).obj();
      |                   ^~~~~~~
src/action.cpp:673:47: 附注：the temporary was destroyed at the end of the full expression ‘(& here)->map::furn((* & p)).int_id<furn_t>::obj()’
  673 |     const furn_t &xfurn_t = here.furn( p ).obj();
      |                             ~~~~~~~~~~~~~~~~~~^~
src/action.cpp:674:18: 错误：possibly dangling reference to a temporary [-Werror=dangling-reference]
  674 |     const ter_t &xter_t = here.ter( p ).obj();
      |                  ^~~~~~
src/action.cpp:674:44: 附注：the temporary was destroyed at the end of the full expression ‘(& here)->map::ter((* & p)).int_id<ter_t>::obj()’
  674 |     const ter_t &xter_t = here.ter( p ).obj();
      |                           ~~~~~~~~~~~~~~~~~^~
在文件作用域：
cc1plus: 附注：unrecognized command-line option ‘-Wno-unknown-warning-option’ may have been intended to silence earlier diagnostics
cc1plus：所有的警告都被当作是错误
make: *** [Makefile:975: obj/tiles/action.o] Error 1
make: *** 正在等待未完成的任务....
src/activity_item_handling.cpp: In function ‘activity_reason_info can_do_activity_there(const activity_id&, Character&, const tripoint_bub_ms&, int)’:
src/activity_item_handling.cpp:1344:31: 错误：possibly dangling reference to a temporary [-Werror=dangling-reference]
 1344 |                 const recipe &r = recipe_dictionary::get_uncraft( ( i.typeId() == itype_disassembly ) ?
      |                               ^
[...]

PR #65542, opened 2023-05-08
Disable buggy warnings

Arkaeriit:
#### Summary
Infrastructure "Fix GCC warnings"

#### Purpose of change
When compiling the game, a lot of warnings (especial dangling reference warnings) were triggered. This is probably because I am using GCC 13.

#### Describe the solution
I performed some slight modifications to the code that don't change the final behavior but that please the compiler.

#### Testing
I tested in game the result of those changes to ensure that the game behavior is unchanged.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,8 @@ WARNINGS = \
   -Wunused-macros \
   -Wzero-as-null-pointer-constant \
   -Wno-unknown-warning-option \
-  -Wno-dangling-reference
+  -Wno-dangling-reference \
+  -Wno-c++20-compat
 # Uncomment below to disable warnings
 #WARNINGS = -w
 DEBUGSYMS = -g
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,8 @@ WARNINGS = \
   -Wsuggest-override \
   -Wunused-macros \
   -Wzero-as-null-pointer-constant \
-  -Wno-unknown-warning-option
+  -Wno-unknown-warning-option \
+  -Wno-dangling-reference
 # Uncomment below to disable warnings
 #WARNINGS = -w
 DEBUGSYMS = -g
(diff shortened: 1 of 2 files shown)


irwiss:
How does this affect performance?

At a glance this will switch a ton of valid references (at least for the lifetime of their call site) to making copies, for example stuff like map::get_known_connections will now potentially copy ter_t(vector + 2 sets + minor stuff) in an inner loop and itself is called in a loop, the change in output.cpp seems just plain weird

Arkaeriit:
Hi irwiss.

I am quite a beginner at C++ so I might be doing some bad mistakes. If I understand the issue, GCC is not happy with using references from the chain of function calls as something needed for the rest of the function.

The performance cost of my solution is indeed quite bad. Thank you for pointing it out, I will try to find a better way.

What is the issue with the change in output.cpp? As I am not very accustomed with C++, there might be something I am missing.

irwiss:
string_id and int_id act like pointers in this codebase(react to operator* and arrow), they'll call the .obj() to return a reference to the pointed-at object, these objects are valid after json data is loaded (happens just before savegame data loads), until game quits to main menu, i don't know of any exceptions to this

I'm not sure what gcc13 complains about as i don't use it but likely the solution is to squelch that particular warning, either in makefile or somehow annotate .obj() function

the output.cpp change is weird because you change from (allegedly) dangling reference warning to (allegedly) a dangling pointer, if that pleases the compiler then the compiler seems to be wrong

Arkaeriit:
I disabled that warning in the makefile. Is that what you wanted?

irwiss:
That's one way although that completely eliminates even the may be valid results

May be wrapping the specific function that raises the false positives in push/ignored/pop pragmas is an option, cursory google search appears to return a bunch of projects that resorted to pragma wraps for view/reference wrapper kind of classes

i'm not sure why it's specifically ter/furn it complains about though, as it should complain about all types going through generic_factory::obj() 🤷‍♂️

andrei8l:
> That's one way although that completely eliminates even the may be valid results

`-Wdangling-reference` is very broken as evidenced by the numerous upstream bug reports and ASan will catch any valid issue anyway. Disabling the warning is the right solution for now.

Fixes #65449

I would also add `-Wno-c++20-compat` even though you've fixed the only trigger.

PR #65546, opened 2023-05-08
Changed Giant Pincer into a piece of integrated armor with qualities and some natural moves, also messed a little with the warmth values of crustacean carapace.

misora31:
#### Summary
Features "Changed Giant Pincer into a piece of integrated armor with Prying 1, Hammer 1, Cut 1, and Butchering -4 also decreased its encumbrance from 40 to 30. It also know has the built in moves Parry and Brutal Strike."

#### Purpose of change

To bring it inline with Sclerotin Claws that are a part of CHITIN3, those also have rapid strike built-in. Changes to warmth modifiers are because you have enough encumbrance already that wearing clothing to add warmth makes you extremely slow and vulnerable. Without this Crustacean mutants get frostbite in around 30 minutes during mildly cold periods.

Keeping these the same as they currently are.
#### Testing

Spawned in game and mutated myself, walked around and used the claw in combat and to hammer some stuff.

### DIFF
--- a/data/json/items/armor/integrated.json
+++ b/data/json/items/armor/integrated.json
@@ -777,7 +777,7 @@
     "type": "ARMOR",
     "category": "armor",
     "name": { "str_sp": "sclerotin right pincer" },
-    "description": "Useful for the hunt, suitable for maximum efficiency.",
+    "description": "A powerful tool for snatching fish or breaking open shells.",
     "weight": "900 g",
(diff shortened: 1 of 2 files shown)


fairyarmadillo:
I'd suggest changing the integrated claw item's description to something like "A powerful tool for snatching fish or breaking open shells."

PR #65554, opened 2023-05-09
Toggle language to English - Adds instructions for users

MarcusAseth:
#### Summary
Interface "Provides a clear instruction on how to use "Toggle language to English" option correctly"

#### Purpose of change
Currently, the option "Toggle language to English" is meant to receive a global keybind to work as intended because the keybind is not assigned to any context, thus assigning a local keybind will severely limit the functionality of the option.
User don't have a way to know this, so they will mistakenly assign a local keybind and then will get the impression that my addition is incomplete or poorly made, and I take issue with that.

#### Describe the solution
I've changed the keybind entry from "Toggle language to English" to "Toggle language to English (requires global keybind!)"
This way the keybind will be self-explanatory on how is meant to be used.

#### Additional context
I'm making this PR after a user was confused by this option, here's their message:
Cattura

That's what happens when a local keybind is assigned, it will work only outside of menus.

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -2102,7 +2102,7 @@
   },
   {
     "type": "keybinding",
-    "name": "Toggle language to English",
+    "name": "Toggle language to English (requires global keybind!)",
     "id": "toggle_language_to_en"
   },
   {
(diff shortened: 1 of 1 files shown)


ashGlaw:
Good fix! Awesome feature, by the way! I'm really glad you thought of this

MarcusAseth:
All the credit goes to Guild Wars 1, that game showed me how powerful his could be :)

PR #65557, opened 2023-05-09
[Xedra Evolved] Sane inventor's researches

GuardianDll:
#### Summary
None
#### Purpose of change
Accidentally found inventor's researches are too wild in current state
#### Describe the solution
Decrease the threshholds to research anything for a lot
Processing sparks not only transform `sparks` into `spark_research`, but also is boosted by the amount of your research done, so value of sparks you got passively is bigger
First two research recipes do not require software_math now


Maleclypse (approved):
Good catch!

GuardianDll:
oh, forgot to tweak recipes a bit

PR #65561, opened 2023-05-09
Wandering Hordes prefer to wander and recruit in open terrain

stubkan:
#### Summary
Balance "Wandering Hordes prefer to wander and recruit in open terrain"

#### Purpose of change

At the moment, wandering hordes move around and recruit zombies almost indiscriminately, mostly ignoring the terrain they are on.

This leads to unrealistic scenarios such as zombies appearing inside empty buildings when they wander in, and zombies that should be inside buildings (such as prison inmates) mysteriously teleporting to the open field when they become part of a horde.

This PR adds a check to a zombie's location before horde recruitment, to check if they are in open terrain before recruitment.  Also the horde move to preference is tweaked to much prefer open terrain (roads and fields) and everything else has a reduced preference.  A horde will still path through difficult terrain in order to reach cities and noises however.

#### Describe the solution

The first change is a check added to zombie horde recruitment to pass a check, which is;

> Does the overmap terrain this zombie is currently on ;
> contain the string "field" or "road" (to also check for farm field, dirt road)
> be prefixed by "swamp" or "road" (to cover all road types)

If it passes any one of these 4 checks, the zombie becomes eligible for wandering horde recruitment.

The second change is to the wandering horde's movement chance onto an overmap tile algorithm, which is ;

Deleting the old code, and using the more versalite is_ot_match() function to iterate through destination checks,
starting from the most preferred (1) (and common, for least overhead) terrain type to least preferred (25), 

> Does destination overmap tile string contain "road"?  Set preference to 1 if so,
> Does destination overmap tile string contain "field"?  Set preference to 3 if so,
> Is destination overmap tile string prefixed by "forest"/"swamp"/"bridge"?  Set preference to 6 if so,
> Else does the destination tile have the prefix "river"? Set preference to 10 if so.
> All other overmap destinations (which will most likely be a structure or obstacle) will be set to a preference of 25.

#### Testing

Disabling of Horde recruitment from inside Specials testing results.

Walked past a TCL, left the area...  Observed no hordes appearing.

image

Went back, entered the carpark and agitated some zeds to draw them outside the special into wilderness.

image

image

Left the area, then checked the map to see if any hordes have appeared.  Success apparent, as the hordes that appeared are only the zombies that were drawn outside the special.

image

#### Additional context

horde_test1_output

I will be doing more testing, as of now it seems to me that a preference of 25 for structures does allow the horde quite a high chance to wander through a house.  I will test with higher values, and may adjust the other preferences accordingly.

### DIFF
--- a/src/overmap.cpp
+++ b/src/overmap.cpp
@@ -3995,9 +3995,9 @@ void overmap::move_hordes()
             // Only monsters in the open (fields, forests, roads) are eligible to wander
             const oter_id &om_here = ter( project_to<coords::omt>( p ) );
             if( !is_ot_match( "field", om_here, ot_match_type::contains ) ) {
-                if( !is_ot_match( "forest", om_here, ot_match_type::prefix ) ) {
-                    if( !is_ot_match( "swamp", om_here, ot_match_type::prefix ) ) {
-                        if( !is_ot_match( "road", om_here, ot_match_type::contains ) ) {
+                if( !is_ot_match( "road", om_here, ot_match_type::contains ) ) {
+                    if( !is_ot_match( "forest", om_here, ot_match_type::prefix ) ) {
+                        if( !is_ot_match( "swamp", om_here, ot_match_type::prefix ) ) {
                             monster_map_it++;
                             continue;
                         }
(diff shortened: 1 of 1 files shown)


GuardianDll:
Is it possible to fix the issue, when horde can spawn inside the building, or its out of scope of this PR?

stubkan:
> Is it possible to fix the issue, when horde can spawn inside the building, or its out of scope of this PR?

If there is a reduced preference to path through buildings, that issue should come up much less. So, yeah it potentially is a step toward fixing that issue.

I dont like how they are still moving into buildings in my testing, so I think the numbers or algo needs another look at.

PR #63909, opened 2023-03-03
Misc Item Changes Again

Karol1223:
#### Summary
Balance "Yet more item cleanup and fixes"

#### Purpose of change
As always! There's always room for improvement and I'm here to get that room demolished!

#### Describe the solution

<details>
<summary>WEIGHT CHANGES</summary>

- Bow saw: 800g -> 750g
- Sewage water: 280g -> 262g
- Pair of copper earrings: 9g -> 8g
- Copper ring: 9g -> 8g
- Pot: 550g -> 880g
- Steel frying pan: 528g -> 660g
- Makeshift copper pot: 884g -> 300g
- Aluminum frying pan: 200g -> 400g
- Scrap/Chunk of copper: 432g -> 250g
- Copper knife: 600g -> 200g
- Wood axe: 2.267kg -> 1.1kg
- Copper axe: 3.7kg -> 1kg
- Tobacco pipe: 372g -> 100g
- Leather belt: 102g -> 150g
</details>

<details>
<summary>VOLUME CHANGES</summary>

- Wooden chair: 112l -> 30l
- Barstool: 75l -> 30l
- Plywood chair: 94l -> 30l
- Bow saw: 900ml -> 3.5l
- Copper bracelet: 5ml -> 10ml
- Pot: 2.1l -> 3.5l
- Copper pot: 2.1l -> 3.5l
- Steel frying pan: 1l -> 3l
- Copper frying pan: 1.05l -> 3l
- Aluminum frying pan: 1.05l -> 3l
- Makeshift copper pot: 1.05l -> 1.1l
- Scrap/Chunk of copper: 50ml -> 30ml
- Copper knife: 250ml -> 300ml
- Wood axe: 2.5l -> 1.5l
- Copper axe: 2.5l -> 1.5l
- Protective bark: 0.3l -> 7.5l
- Distributed bark: 0.3l -> 15l
- Consolidated bark: 0.3l -> 15l
- Latticed bark: 0.3l -> 15l
- Chemical reference-CLASSIFIED: 0.5l -> 2l
- FM 23-25 Army bayonet manual: 0.25l -> 0.5l
- Spetsnaz Knife Techniques: 0.25l -> 0.5l
- Tobacco pipe: 100ml -> 150ml
- Leather belt: 250ml -> 350ml
- Steel buckle: 5ml -> 10ml
</details>

<details>
<summary>LENGTH CHANGES</summary>

- Bow saw: 40cm -> 88cm
- Wood axe: 96cm -> 75cm
[...]

PR #65568, opened 2023-05-10
Fix weight, size, and damage for tool axes

worm-girl:
#### Summary
Balance "Adjusts weight, size, and damage for tool axes"

#### Purpose of change

I noticed the wood axe was weighing in at just a kilogram despite being described as a heavy wood-chopping axe with a 75cm handle, and was swinging faster than a baseball bat while dealing a whopping 31 damage per second(!), which is Hanzo steel territory. A brief foray into the jsons came up with several inconsistencies in axe size and weight, plus a few things like the copper axe that just didn't make a lot of sense.

Additionally, I discovered that it is very common for product listings online to only list the weight of the axe head and not the total weight of the item, which I suspect was an ongoing source of some of the confusion in statting these items. Therefore I've provided specific examples which either list total item weight or have separate data for heads and handles.

#### Describe the solution

**WOOD AXE**
I have used a medium sized axe that differentiates it from the hatchet (907g) and the fire axe (now 3448g). Both items come in a variety of sizes. All values taken from https://www.harborfreight.com/4-12-lb-hickory-axe-62473.html

LENGTH: 78 cm
HEAD: 2041 g

TOTAL WEIGHT: 2227 g

**FIRE AXE**
I used a larger axe to match its existing length and differentiate it from the wood axe. I increased the damage accordingly, which will make it better at bashing things, its primary purpose. All values taken from https://www.harborfreight.com/6-lb-firemans-axe-69246.html

LENGTH: 90 cm
HEAD: 2722 g

TOTAL WEIGHT: 3448 g

**COPPER AXE**
The description and crafting recipe suggest that this is a primitive item, not something like a modern beryllium copper tool. I went back and had a look at historical axes, and it doesn't seem like copper is a suitable material for the large modern-style wood axes this item was likely copied from. Therefore I've restatted it as a much smaller item, based on Otzi the Iceman's axe. [Link here (Contains a partial photo of a frozen copper age mummy so don't click if that will bother you)](https://www.cambridge.org/core/services/aop-cambridge-core/content/view/BE5499FB7D007EFEBCDAB315324772D2/S155192950005392Xa.pdf/metallography-for-the-european-copper-age-research-on-the-axe-blade-of-the-glacier-mummy-from-the-otztaler-alps-in-tyrol.pdf)

Otzi's axe, and all the other real-world copper axes I could find, are very light and way smaller than modern steel hatchets. As such I've reduced the damage and updated its name and description.

HEAD: 174 g (see above)
HANDLE: 205 g - The data above lacks a weight for Otzi's axe handle which was a stick made of yew. To approximate a weight, I've used this hatchet handle: https://www.amazon.com/20-Hickory-Clearing-Axe-Handle/dp/B002M164XC/754&psc=1
LENGTH: 35 cm

TOTAL WEIGHT: 379 g

**BRONZE AXE**
Unlike copper, bronze was used to make large axes roughly in the ballpark of wood axes we use today. See https://www.metmuseum.org/art/collection/search/548308

Head dimensions: 13.6 cm x 9.8 cm x 1 cm = 133.28cm^3 total volume.
Bronze density: 8.8 g/cm^3
Head Weight = 1199.52 g

The met museum does not list a total weight for the item, so I found a modern axe handle of roughly the same size. 

LENGTH: 56 cm (See above)
HEAD WEIGHT: 1120 g
HANDLE: 369 g https://www.amazon.com/20-Hickory-Clearing-Axe-Handle/dp/B002M164XC/754&psc=1

TOTAL WEIGHT: 1489 g

**STONE AXE**
This thing was super long before, as big as the fire axe. The existing stone axe head item was 453 g, which seems in line with sources like https://collections.peabody.harvard.edu/objects/details/58583 so I simply added the same handle from the bronze axe to its weight.

LENGTH: 56 cm (See above)
HEAD WEIGHT: 453 g
HANDLE: 369 g https://www.amazon.com/20-Hickory-Clearing-Axe-Handle/dp/B002M164XC/754&psc=1

TOTAL WEIGHT: 822 g

**STONE ADZE**
This was needlessly long. I've brought it in line with the other large primitive axe-type weapons using the same handle as the bronze and stone axes and head data from the British Museum.

LENGTH: 56 cm (See above)
HEAD: 581 g https://www.britishmuseum.org/collection/object/H_1958-0506-5616
HANDLE: 369 g https://www.amazon.com/20-Hickory-Clearing-Axe-Handle/dp/B002M164XC/754&psc=1

TOTAL WEIGHT: 950 g

**HATCHET AND CRASH AXE**
The first examples of these items that I googled were nearly identical to the existing game data, so I left them alone. Perfect, no notes.

Battle axes probably have some weirdness of their own going on, but getting data on them isn't quite as easy and they don't come up as often in play, so I didn't make it a priority.

#### Testing

I spawned into the world and tested that all axes have the expected weight, length, and damage values. I spawned many different axe-holding items and made sure they still hold axes as before. I tried using the fire axe in combat to make sure it was still effective at its new weight (it hits very very hard but is impractically slow when you're swarmed, which is about what you'd expect).


bombasticSlacks:
Changing damage as you have has messed up the DPS tests.

fairyarmadillo:
> Changing damage as you have has messed up the DPS tests.

I've just got to change the info in the tests, right? I'll try that in a bit unless there's some bigger issue with how I've gone about things.

bombasticSlacks:
If the description / size of an axe is meaningfully changing them the DPS tests should be updated, but for most of these the damage values provided should be moved to accommodate the tests. The tests are how they are for a reason and shouldn't be adjusted without very good reasoning.

taatu:
Wood axe was 2200 grams as recently as 0.G, but now it's only 1100 grams which definitely seems wrong.

It was changed in #63909 alongside many other items. I think this may be fixing a recent oversight?

fairyarmadillo:
I changed all the damage values to ones that should pass the tests. If they don't I'll come back and do it again.

I did change one test value, the copper axe. It was previously a two-handed axe and its test value was 13.5, but since the item is now a one-handed axe with a very narrow blade I've changed it to 12.0. It's an extremely primitive and lightweight copper tool and doesn't seem like much of a weapon at all.

> Wood axe was 2200 grams as recently as 0.G, but now it's only 1100 grams which definitely seems wrong. It was changed in #63909 alongside many other items. I think that may have been an oversight?

Seems to be the case, yeah.

fairyarmadillo:
I'm a little confused about the DPS tests. The test_guy has 10 in all stats and 4 in all skills, right?

When I test this in-game, the DPS listed for a character like that wielding the wood axe from this PR is 55.31, which is well above the 31 that the build test told me it got. At 8/8/8/8 and all 0s it's showing up as 20 on my end. How is it getting 31?

Karol1223:
If you wanna do all the tool axes you're still missing hatchets

fairyarmadillo:
> If you wanna do all the tool axes you're still missing hatchets

The first hatchet and crash axe I googled were near perfect matches to the ones in the game so I left them alone.

fairyarmadillo:
-------------------------------------------------------------------------------
Grab breaks against midline grabbers
  Lategame character against 1 midline grabber
-------------------------------------------------------------------------------
D:\a\Cataclysm-DDA\Cataclysm-DDA\tests\monster_attack_test.cpp(643)
...............................................................................

D:\a\Cataclysm-DDA\Cataclysm-DDA\tests\monster_attack_test.cpp(657): FAILED:
  CHECK( success == Approx( 800 ).margin( 50 ) )
with expansion:
  859 (0x35b) == Approx( 800.0 )
with message:
  80% chance to break a midline grab as a lategame character with grab breaks


This appears to be failing a grab test, which should be unrelated to my changes.

PR #65570, opened 2023-05-10
Doing parkour trains parkour

anoobindisguise:
#### Summary
Balance "Parkour can be trained by doing parkour"

#### Purpose of change
I was really annoyed that the only way to train parkour was to build specialized monkey bars and then use those.

#### Describe the solution
You can now learn parkour practically by moving over difficult terrain and climbing buildings.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -10365,7 +10365,8 @@ bool game::walk_move( const tripoint &dest_loc, const bool via_ramp, const bool 
             }
         }
         // learn a little about parkour
-        u.practice_proficiency( proficiency_prof_parkour, time_duration::from_turns( ( mcost_from + mcost_to ) / 2 ) );
+        u.practice_proficiency( proficiency_prof_parkour,
+                                time_duration::from_turns( ( mcost_from + mcost_to ) / 2 ) );
     }
     if( !u.is_mounted() && u.has_trait( trait_LEG_TENT_BRACE ) &&
         u.is_barefoot() ) {
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -10366,7 +10366,7 @@ bool game::walk_move( const tripoint &dest_loc, const bool via_ramp, const bool 
         }
         // learn a little about parkour
         // don't learn anything from traversing tall grass or dirt mounts
-        if ( !m.has_flag_ter_or_furn( ter_furn_flag::TFLAG_DIGGABLE ) {
+        if ( !m.has_flag_ter_or_furn( ter_furn_flag::TFLAG_DIGGABLE ) ) {
             u.practice_proficiency( proficiency_prof_parkour,
                                     time_duration::from_turns( ( mcost_from + mcost_to ) / 2 ) );
         }
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -10326,12 +10326,13 @@ bool game::walk_move( const tripoint &dest_loc, const bool via_ramp, const bool 
     u.recoil = MAX_RECOIL;
     u.last_target_pos = std::nullopt;
 
+    const bool parkour = u.has_proficiency( proficiency_prof_parkour );
     // Print a message if movement is slow
     const int mcost_to = m.move_cost( dest_loc ); //calculate this _after_ calling grabbed_move
     const bool fungus = m.has_flag_ter_or_furn( ter_furn_flag::TFLAG_FUNGUS, u.pos() ) ||
                         m.has_flag_ter_or_furn( ter_furn_flag::TFLAG_FUNGUS,
                                 dest_loc ); //fungal furniture has no slowing effect on Mycus characters
-    const bool slowed = ( ( !u.has_proficiency( proficiency_prof_parkour ) && ( mcost_to > 2 ||
+    const bool slowed = ( ( !parkour && ( mcost_to > 2 ||
                             mcost_from > 2 ) ) ||
                           mcost_to > 4 || mcost_from > 4 ) &&
                         !( u.has_trait( trait_M_IMMUNE ) && fungus );
@@ -10364,6 +10365,9 @@ bool game::walk_move( const tripoint &dest_loc, const bool via_ramp, const bool 
                 }
             }
         }
+        // learn a little about parkour
+        units::time{
+        u.practice_proficiency( proficiency_prof_parkour, time_duration::from_turns( ( mcost_from + mcost_to ) / 2 ) );
     }
     if( !u.is_mounted() && u.has_trait( trait_LEG_TENT_BRACE ) &&
         u.is_barefoot() ) {
@@ -12822,7 +12826,9 @@ bool game::slip_down( bool check_for_traps )
         slip *= 2;
         add_msg( m_info, _( "Your bad knees make it difficult to climb." ) );
     }
-
+    if( !parkour ) {
+        u.practice_proficiency( proficiency_prof_parkour, 10_seconds );
+    }
     add_msg_debug( debugmode::DF_GAME, "Slip chance after proficiency/trait modifiers %d%%", slip );
 
     // Climbing is difficult with wet hands and feet.
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -10326,13 +10326,12 @@ bool game::walk_move( const tripoint &dest_loc, const bool via_ramp, const bool 
     u.recoil = MAX_RECOIL;
     u.last_target_pos = std::nullopt;
 
-    const bool parkour = u.has_proficiency( proficiency_prof_parkour );
     // Print a message if movement is slow
     const int mcost_to = m.move_cost( dest_loc ); //calculate this _after_ calling grabbed_move
     const bool fungus = m.has_flag_ter_or_furn( ter_furn_flag::TFLAG_FUNGUS, u.pos() ) ||
                         m.has_flag_ter_or_furn( ter_furn_flag::TFLAG_FUNGUS,
                                 dest_loc ); //fungal furniture has no slowing effect on Mycus characters
-    const bool slowed = ( ( !parkour && ( mcost_to > 2 ||
+    const bool slowed = ( ( !u.has_proficiency( proficiency_prof_parkour ) && ( mcost_to > 2 ||
                             mcost_from > 2 ) ) ||
                           mcost_to > 4 || mcost_from > 4 ) &&
                         !( u.has_trait( trait_M_IMMUNE ) && fungus );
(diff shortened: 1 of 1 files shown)


Venera3:
This needs some better filtering. Vibrating between two dirt mounds for eight hours a traceur does not make, and at least with monkey bars you have a reasonable handwave of training different movesets.

anoobindisguise:
> This needs some better filtering. Vibrating between two dirt mounds for eight hours a traceur does not make, and at least with monkey bars you have a reasonable handwave of training different movesets.

OK, made it check for the DIGGABLE flag. That handles mounds of dirt, underbrush, etc, though it does have a few false positives like picket fences and graves. I don't know really an elegant solution to this but practicing vaulting through window frames and over tables etc should definitely be doing something.

Venera3:
Something, sure, but you can't really enforce any sort of variety with it (unless you add Principles of Vaulting Dirt Stuff, Principles of Vaulting over Yea-High Stuff, Principles of not Falling and Breaking Your Ankle etc. each tracked separately). Add to that the huge incentive this gives to just grind as opposed to routing through the practice recipe I'm not convinced organically learning it is desirable to be honest.

anoobindisguise:
> Something, sure, but you can't really enforce any sort of variety with it (unless you add Principles of Vaulting Dirt Stuff, Principles of Vaulting over Yea-High Stuff, Principles of not Falling and Breaking Your Ankle etc. each tracked separately). Add to that the huge incentive this gives to just grind as opposed to routing through the practice recipe I'm not convinced organically learning it is desirable to be honest.

That's fair, but if the player really wants to grind then they just level Fab up to 5 or whatever it was, dig a hole, and then practice, none of which requires fighting zombies. In my eyes this incentivizes meeting zombies sooner because it means that all those times you climb over a sofa or whatever do matter and contribute towards overall training. I don't think it really makes it that much easier to grind it to learning than what exists already, though not needing a welder does make things a bit more simple and mashing <--> over and over definitely sucks ass in terms of actual gameplay. Also, it's a little silly that the player needs handmade parkour bars specifically and not just like, a playground monkey bars which is mostly a game limitation (since it needs to provide a pseudo furniture for the recipe to work)

I agree that parkour is probably too all encompassing (handling cars, windows, shrubbery, climbing roofs, etc) though making that no longer the case might be out of scope.

fairyarmadillo:
Parkour isn't something you just pick up over time running around, it's a specific set of purposefully designed techniques that were modeled after martial arts training. I don't hate the parkour gym, but I think if we expand it, it should require a teacher or a book rather than just being something you can brainlessly grind in a field. It's not like you can learn Eskrima by stabbing zombies.

KHeket:
> Parkour isn't something you just pick up over time running around, it's a specific set of purposefully designed techniques that were modeled after martial arts training. I don't hate the parkour gym, but I think if we expand it, it should require a teacher or a book rather than just being something you can brainlessly grind in a field. It's not like you can learn Eskrima by stabbing zombies.

image

So, if the guy will just jump through the windows, fences, climbing/get down on/from roofs and making other parkour stuff without lessons and teacher - he will learn parkour in realife. Yes, it will be less effective and more trauma dangerous, but will be trainable alone in the field, so how it do lots of young guys in real life. You can also learn parkour technics in the woods, because fallen trees, different types of large stones and other items in wild life can be good simulators of training apparatus

fairyarmadillo:
You could apply that logic to martial arts. We don't accept it there and we shouldn't accept it here.

Zireael07:
I believe you CAN pick up at least the basics organically, so... maybe let doing parkour train parkour but only to some level.

NB: Parkour is in the same boat as social as in, opportunities/ways to train it are very thin on the ground, pun intended

fairyarmadillo:
And sorry to double post but if you're interested in pursuing it, OP, a neat compromise might be getting an introductory course from a book or NPC, which unlocks the ability to practice via the methods you're outlining here.

I think that would just involve checking for whether any % of parkour has been learned before allowing the player to learn on their own. This could later be applied to martial arts as well and would potentially be a lot better than our current system.

KHeket:
> You could apply that same logic to martial arts. We don't accept it there and we shouldn't accept it here.

No, you could not apply the same logic to martial arts. Parkour as you can read from wiki - getting from point a to point b. Martial arts are trained for centures or at least for few decades. It have special movements and you cant learn taekwondo by figthing enemies, because you will need a lot of time to create a fight technique in the field and it will be a new technique. But parkour is just moving through obstacles, we all have the same legs, arm and bodyes, obstacles also everywhere is the same. Yes, there is spechial movements, but if you dont know them - you can also be good in parkour. The same way most of parkour guys learning it in field in reallife. They dont use books, they dont read tons of information and teaching with teacher, they just fo in the wood/abandoned building/roofs/any other place with lots of abstacles - and just runing/jumping/rolling and making other parkour stuff. Yes, you can watch video to boost you leaning, wich will be much more faster, than reading or teaching with teacher, but I dont think, that most of guys will do it. And anyway, learn parkour by the climbing in the game also will take lots of time, so the changes is quite ok

fairyarmadillo:
> I believe you CAN pick up at least the basics organically, so... maybe let doing parkour train parkour but only to some level.
> 
> NB: Parkour is in the same boat as social as in, opportunities/ways to train it are very thin on the ground, pun intended

This is backwards from how I suggested and doesn't sound as reasonable to implement. My idea was to use the introductory course from a teacher or book as a way to open the door to the idea, which the student could then go and learn independently.

Nobody's going to invent taekwando wholecloth, but if a guy showed you the basics, you could go practice that in the field by yourself until you got passably good at it, given enough time and effort. Since parkour was devised as a nonviolent martial art, it stands to reason it would follow the same logic. If multiple lessons are deemed necessary, you could just have it be a multi-stage proficiency like archer's form.

And a lack of teachers is no excuse not to implement it. It would be very simple to spawn a couple of books in places and make one of our many static NPCs able to teach it.

Zireael07:
> Nobody's going to invent taekwando wholecloth, but if a guy showed you the basics, you could go practice that in the field by yourself until you got passably good at it, given enough time and effort

No you can't get passably good at it by practicing any of the martial arts alone. Unless you mean doing a rote kata. But that's not being good at any of the arts. You need a live opponent.

Source: a relative of mine used to practice karate

fairyarmadillo:
Alright but presently you read a book once and become a black belt. Let's think in terms of what is possible in our engine.

KHeket:
Anyway, leaning parkour by prkour actions is much more better and realistic mechanic, than just building parkour bars without any info and just practising on them without any teachers or books, and the changes un this PR also are **realistic** , because the same we see in reality

> Plenty of real life martial artists are self-taught using training materials and introductory lessons, and plenty of them do quite well this way. And the player is not lacking for live (or undead) opponents to practice with.

Making teachers, books, video to boost the parkour traning is also good idea, but it can be made later, step by step

Fris0uman:
To double down on what Venera said adding a "spend X hours climbing on and off a sofa" to people's playthrough is not super interesting. It would probably be a better idea to add more practice recipe instead. If we add a way to get a bonus by gridding stupid actions people are going to do that and it's not fun, see the years of meta of sitting in basement repairing socks and throwing rocks at the wall.

anoobindisguise:
> To double down on what Venera said adding a "spend X hours climbing on and off a sofa" to people's playthrough is not super interesting. It would probably be a better idea to add more practice recipe instead. If we add a way to get a bonus by gridding stupid actions people are going to do that and it's not fun, see the years of meta of sitting in basement repairing socks and throwing rocks at the wall.

This is my biggest concern. For example when Night made swimming training not work while standing still in water, all it did was make you need to press more keys. This isn't quite the same but it still lets you get an advantage by pressing keys over and over in boring as shit behavior, but unfortunately as long as we use the elder scrolls model of skill training it will probably be at least partially present. That said making it worse definitely isn't ideal and I don't know of a great way to simulate "climbing over fences a lot to escape the zombies makes you better at climbing fences" without also having the issue of "ok, now time to climb fences for 10 hours to get good at that"

Venera3:
The core problem is the (percieved) ease of learning skills/profs - if you introduce "do stupid repetitive irl thing for an in-game afternoon -> get cookie" you prime players for optimizing that as a subversion of the system. Now, if it was a month instead of 8 hours players *wouldn't* try to dodge learning organically quite so readily, unless you introduced "jump over couchx300" into everybody's earlygame routine. That works for swimming/athletics as well, incidentally.

fairyarmadillo:
What about putting a cooldown on it, so that you can only get a very small amount of the proficiency via movement per day? Jumping over the couch 300 times in one day to get the very desirable proficiency doesn't sound great, but making a habit out of doing something in order to eventually master it over weeks/months is a lot better.

bombasticSlacks:
I think tracking separately a few actions would be fine.


I also think a good compromise would be splitting parkour into 2 proficiencies, 

"principals of urban traversal" (or something)
and then full
"parkour"

Training passively would only get you to tier one as practice will help with urban traversal but without training some rolls and other movement techniques reinforcing bad form is dangerous.

Tread4:
> I also think a good compromise would be splitting parkour into 2 proficiencies,
> 
> "principals of urban traversal" (or something) and then full "parkour"
> 
> Training passively would only get you to tier one as practice will help with urban traversal but without training some rolls and other movement techniques reinforcing bad form is dangerous.

This. What bombasticSlacks described would be closest to IRL parkour. Source: did parkour for ~4 years in my teen years.
I would simply make "Basic Parkour" and "Parkour Expert" proficiencies (or "Parkour: Basics", "Parkour: Expert").

It is true you can learn a lot and "get better" by "just, running, jumping and climbing" over things. Similar as most other physical activities you simply get better in it and develop own techniques (what works for you etc.). 

Nonetheless, to get truly proficient you most likely need to learn some well established and proper techniques (and useful details) from someone (teacher, video, book). There are some principles that would be hard to "stumble upon" just by trying. Of course they might be known (or easier to come up with) to a gymnast or physical academy graduate. In such cases we can safely assume that player picks up proficiency at the start to role play. Nevertheless, average Joe will most likely not know or come up with them. In all fairness it is not a lot of "secret knowledge" so the potential book etc. can be made rather short.

Furthermore, as already mentioned, it is easy to develop "bad habits" by using sub-optimal technique instead of a proper (well-proven) one. Usually it is much harder to shake off a bad habit and learn the proper technique than learning it in the first place.

To support my stance a few examples:
- Making a proper landing roll - seems easy but a sub-optimal rolling technique is the biggest cause of serious injuries (legs, head or spine)
- Wall scaling gets a significant boost from having a proper "wall kick" technique (angle etc.)
- Jump height for some vaults or acrobatics is highly dependant on a technique as well (proper angles)
- Plenty of vaults are named and described. Anyone can "come up with" basic ones, but some more exotic ones will not come to mind while still being very useful.
- Even for the basic stuff there are some minor details and tricks that can make it easier, faster or more efficient.

anoobindisguise:
My plan for this is probably to convert Parkour into 3+ separate proficiencies, and then let them be trained by actions.
- Urban Traversal: this represents movement speed penalties for non natural terrain (windows, sofas, tables, etc)
- Bushwhacking: this represents movement speed penalties for natural terrain (bushes, long grass, etc)
- Climbing: This represents not injuring yourself from failing a climbing check.

PR #65571, opened 2023-05-10
Options accessible by EOC

bombasticSlacks:
#### Summary
Infrastructure "Options accessible by EOC"

#### Purpose of change
EOCs can now access options of string, numerical, or bool format.

#### Describe the solution
Added conditional for bool, mutator for string, and math function for numbers.

Added Documentation

#### Testing
Unit tests included.

### DIFF
--- a/doc/NPCs.md
+++ b/doc/NPCs.md
@@ -1082,6 +1082,7 @@ Condition | Type | Description
 --- | --- | ---
 `"mod_is_loaded"` | string or [variable object](#variable-object) | `true` if the mod with the given ID is loaded.
 `"get_condition"` | string or [variable object](#variable-object) | Runs the condition stored in the variable `get_condition` for the current dialogue.
+`"get_game_option"` | string or [variable object](#variable-object) | gets the true or false game option for the provided string.
 
 ---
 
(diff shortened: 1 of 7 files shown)


src/math_parser_diag.cpp
```diff
@@ -49,6 +50,21 @@ std::function<void( dialogue &, double )> u_val_ass( char scope,
...
+    return[option = params[0]]( dialogue const & d ) {
+        return get_option<float>( option );
+    };
+}
+
+std::function<void( dialogue &, double )> option_ass( char scope,
+        std::vector<std::string> const &/*params*/ )
+{
+    debugmsg( "assigning options is not currently supported in math" );
+    return []( dialogue const &, double ) {};
```
andrei8l, line 65:
```suggestion
    throw std::invalid_argument( "assigning options is not currently supported in math" );
```
Otherwise it passes validation in a bad PR. Or better yet, don't define the assignment function at all.

doc/NPCs.md
```diff
@@ -1432,6 +1434,7 @@ This section is a work in progress as functions are ported from `arithmetic` to
 |----------|------|--------|-------|-------------|
 | pain     |  ✅  |   ✅   | u, n  | Return or set pain<br/> Example:<br/>`{ "math": [ "n_pain()", "=", "u_pain() + 9000" ] }`|
 | skill    |  ✅  |   ✅   | u, n  | Return or set skill level<br/> Example:<br/>`"condition": { "math": [ "u_skill('driving')", ">=", "5"] }`|
+| game_option   |  ✅  |   ❌   | g     | Return the numerical value of a game option<br/> Example:<br/>`"condition": { "math": [ "game_option('NPC_SPAWNTIME')", ">=", "5"] }`|
```
andrei8l, line 1437: Should be sorted before pain. Also, I prefer `N/A<br/>global` for scope since it matches the prefix (and I wrote it that way for `weather`)

bombasticSlacks:
I'll try to find time tonight to do the fixes. Thanks for the review.

PR #65575, opened 2023-05-11
Arm Solar Generators CBM

anoobindisguise:
#### Summary
Content "Arm Solar Generators CBM"

#### Purpose of change
Augs for the aug god, chrome for the chrome throne! (Wait, wrong WH40K faction...) 
This one is probably possible in real life!!
https://pubmed.ncbi.nlm.nih.gov/27836601/

#### Describe the solution
The above source describes the functionality of photovoltaic cells implanted within the skin at about 9 mW per cm^2 (compare this to a real solar panel having nearly four times that at over 30 mW per cm^2)
The human body has a surface area of about 18000 cm^2. The arms and hands themselves make up about 17-20% of the total surface area of the body. Not all of the arms can be practically exposed to sunlight and obtain any reasonable level of efficiency, plus the hands are being ignored here, so in total about 1/3 of the arms have useful surface area for the purpose of powergen, or about 5% of the entire body, which is about 900 cm^2 of usable solar absorption. This means the bionic can produce a theoretical maximum output of about 8 watts of power.
HOWEVER
There are some limitations here. First of all the bionic represents solar irradiance, so in the early morning or in unideal conditions you are only getting 1-2 watts.
Further, wearing anything on your arms will reduce your power absorption, by a fraction of the amount your arms are covered. 100% coverage = no powergen. There was actually an entire function in bionics.cpp that was totally unused that can do this! I honestly couldn't believe my luck lmao.
Also they cost 1 encumbrance on the arms but that's not as important as the above two.

Lower the powergen even more.
Different versions of this, such as ones covering your neck or your back. These ones cover your arms but the code supports them being implanted anywhere in the body and looks like it will weight the power based on the body part's hit_size (so if it covers both torso and arms, the torso matters more than the arms)
More encumbrance.

#### Testing
I tested these bionics and found that they seem to be functional. I get no powergen while standing indoors, and much more power at 1 PM than I do at 8 AM. by putting on a shirt I get proportionally less power gen, or none at all if I wear a 100% coverage sleeve.


Qrox:
I think the overall idea is cool, but the source you mentioned measured the PV cell power under skin under a direct 100mW/cm^2 light source, which is close to the solar irradiance from the zenith point, so unless you move like a [Jiangshi](https://en.wikipedia.org/wiki/Jiangshi) with arms outstretched at noon near the equator you're not likely to get the 9 mW/cm^2 power density from that paper.

Hirmuolio:
Solar panel power output is scaled by sun angle and weather. The json value is treated as peak output when sun is it at 90° and there are no clouds so that is what the json should define.

Qrox:
That's right, but my point is that it is hard for someone to always keep their arms stretched. When the arms are in their natural position, sunlight will strike at a large angle which will cause the power density to fall considerably.

PatrikLundell:
My problem with this is that solar panels aren't known for their ability to take blows without taking damage. Unless you have self repairing magic in them, they should be rendered useless after the first hit to them, and as far as I know we don't have any CBM damage/repair system (apart from resetting used CBMs).

anoobindisguise:
too many issues to be viable. closing

PR #65581, opened 2023-05-11
recipe_food.json water_boiling_heat audit

Miner239:
#### Summary
Balance "water_boiling_heat audit"

#### Purpose of change

Stemming from [a reddit post](https://www.reddit.com/r/cataclysmdda/comments/13drvod/) that complained about the new cooking energy costs, I decided to take a look at any recipes that uses water_boiling_heat with first principles in mind.

#### Describe the solution

Derived from clean water recipe:
It takes 1u of heat to bring 1u of water or 250g of materials to boiling temperature. This takes a minimum of 5 minutes to do.
It takes 6u of heat to further boil 1u of water into vapor, for a total of 7u. Useful metric for distilling recipes.

Where does 6u comes from? Enthalpy of vaporization for a unit of water is equivalent to 540K of temperature change. Assuming that the water used in crafting starts at 10C, 1u of heat equals 90K difference. This conversion can be applied to vaporization energy: 540/90 = 6u of heat.

For each recipe a search of "boiling" hits:
1. Set boiling cost to the correct amount based on the ingredients and cooking method
2. Increase time cost if it is lower than "1u = 5 m" dictates

Removed erroneous maple syrup recipe energy cost reference. Maple syrup producers that uses evaporation method benefit from specialized equipment, allowing them to reach 70% efficiency even when using fuel combustion for energy source. However, requirements/cooking_tools.json already stated that the player's tools' combustion efficiency is 25%. 

A more based calculation was re-derived from https://mapleresearch.org/wp-content/uploads/wilmot_energy.pdf . 1 gallon of syrup requires 400kBtu / (139 kBtu/gal * 25%) = 11.51 gal of fuel oil. 
Using this 1:11.51 volume ratio, we can calculate that one batch of maple syrup requires 11.51 * 4 charge/batch * (250mL / 16 charge) * (1u / 12mL) = 60u of fuel oil per batch.

Also modifies tallow recipes to cost 2h and have batch time savings, because 10m is just absurd.

I considered also editing Aftershock recipes, but I think it's best to do them separately since the meats and bits recipes were also missed by the previous audit

#### Testing

JSON edits. Edited local copy, works.


Sumlord-047:
Maybe the recipe time for the 3L canned food should also be increased. In the case of canned meat, 40 minutes (just 10 minutes more than the 0.5L recipe) is absurd.

Issue #61970, opened 2022-10-28
Allow mods to define conflicting mods


NetSysFire:
### Is your feature request related to a problem? Please describe.

Some mods would greatly benefit from being able to use conflicts. E.g dark days of the dead does not work correctly with the *default* mods of bionic professions and no fungal growth.

### Solution you would like.

```json
[
  {
    "type": "MOD_INFO",
    "id": "demo",
    "name": "demo",
    "authors": [ "demo" ],
    "description": "demo.",
    "category": "total_conversion",
    "dependencies": [ "dda" ],
    "conflicts": [ "aftershock" ],
  }
]
```

However, it might make sense to let the player override those conflicts if they really know what they are doing™. Upon manually enabling a mod, it should say "demo conflicts with aftershock. Force loading? This will probably break stuff.".

### Describe alternatives you have considered.

Not offering the player a convenient method to override conflicts and force them to modify `modinfo.json` since they should already know about how mods work. Nonetheless, this is advanced stuff and deliberately accepting breakage is something for expert users only.

GuardianDll:
We can use the same syntaxys as for #48589, so i guess they are related

StaticSifting:
Hi, Im not like an active dev or anything but you were talking to me on discord today. I was wondering how you feel about maybe making only the modinfo.json file into an xml for the sake of utilizing the Xpath function for folder loading in the event that a contingency you prepared for has its requirments met. This is like super common in rimworld modding (sorry its my only frame of reference) but Im not sure how capable the current game is of reading xml files so maybe an alternative to look at is jsonPath?

BrettDong:
We don't use any XML in our game and we don't want to add a XML parsing library in our project either.

np4Huk:
Resolved by #80465.

PR #65584, opened 2023-05-11
Make the no rail stations mod default

NetSysFire:
#### Summary
Mods "The No Railstations mod is now loaded by default"

#### Purpose of change

I got a friend of mine into the game and he complained how non-immersive the rail stations are without proper rail connections. I agree on this and as long as we do not have proper railways, this mod should stay a default one.

The rail stations were always super weird and do not fit in with their rail connection going into nowhere.

#### Describe the solution

Add the mod to the default mods.

#### Testing

Created new world, mod is loaded as expected, world looks as expected, no rail stations.

### DIFF
--- a/data/mods/default.json
+++ b/data/mods/default.json
@@ -4,6 +4,6 @@
     "id": "dev:default",
     "name": "default",
     "description": "contains all the mods recommended by the developers",
-    "dependencies": [ "dda", "no_npc_food", "personal_portal_storms", "no_fungal_growth", "package_bionic_professions" ]
+    "dependencies": [ "dda", "no_npc_food", "personal_portal_storms", "no_fungal_growth", "package_bionic_professions", "No_Rail_Stations" ]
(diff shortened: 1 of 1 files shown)


Maleclypse:
@LyleSY mentioning you here in case Dinomod still has a mission hub out of a rail station.

LyleSY:
Yep, sure do. Thanks, this is a good incentive to get that moved.

NetSysFire:
Would be pretty handy if mods could define conflicting mods (see #61970)

Fris0uman:
Maybe we should just obsolete the rail station for good. And come back to it if we ever manage to get actual rails to generate?

Inglonias:
As previous features have shown, turning these things off by default causes dev work on that feature to grind to a halt. That said, these were slated for 0.D, and nobody has touched them since, so that may be a moot concern.

PR #65590, opened 2023-05-11
firemen don't carry flamethrowers jesus

Daved27hundred:
#### Summary
None

#### Purpose of change
 
Do I NEED to explain this?
Even if it could be argued that it's meant for making firebreaks to stop wildfires, There are actual tools for that (see: drip torch) and I am frankly sure the firefighters we have ingame are urban ones, not wildfire control.

#### Describe the solution

Removed the line in `fireman_gear` that allows flamethrowers to spawn.

#### Additional context

Yes I know what Fahrenheit 451 is, we already have a reference to that in the RM451.


andrei8l:
So add a drip torch

Issue #65578, opened 2023-05-11
Accessibility issue, the over map is a tedious muddle to navigate with screen-readers.


DoctorBoomstick:
### Describe the bug

By virtue of the fact that the map relies on visually displaying information with the use of ascii text, the process of a screen reader attempting to announce any usable information is hindered as it attempts to read all of the ascii graphics. This is incredibly tedious to work with, as a VI player is compelled to read the map at a snail’s pace, waiting for their screen reader to finish burbling out line after line of ascii symbols before it announces the name of the location the curser is centered upon. This makes reviewing even a small patch of the map a chore to get done, with a full exploration of the entire map requiring much time and waiting to be accomplished.

### Attach save file

N/A

### Steps to reproduce

1. Hop onto an installment of, in my case, Ubuntu 20.04.5.
2. Enable the natively installed Orca screen reader.
3. Boot up a pure text copy of the game, running in terminal, and enter a game session.
4. Bring up the map and utilize the numpad keys to shift the cursor about various locations, making note of the giant amount of ascii symbols that get announced before the name of the actual location.

### Expected behavior

Ideally, a screen reader would only announce the name of the location upon which the cursor is focused and not read the ascii. While I haven’t tried this personally, I’d suppose that replacing the symbols of all over map tiles with a blank space would solve the issue, i.e., replacing something like "sym": "E" with "sym": " ", as it would just up and remove the ascii outright.
While this is obviously unfeasible for general use, should it work, perhaps a dedicated accessibility toggle, which would convert these ascii symbol entries into a blank space, could be introduced.

### Screenshots

This is a recording I made showcasing how the map presently behaves with screen readers, in which I review my character’s start location before looking at the directly adjoining tiles and eventually following a small section of road while checking what tiles flank it
[Screen-Reader Map Demonstration.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/11450363/Screen-Reader.Map.Demonstration.zip)
.

### Versions and configuration

• OS: Ubuntu 20.04.5
• Game Version: A3dd8c6
• Graphics version: Terminal
• Ingame language: English

Qrox:
Based on previous feedback from screen reader users, placing the terminal cursor before the overmap tile name can probably solve this issue. Do you think it is likely going to work in your case? I can test the cursor position with MSYS2 emulating ncurses but the narrator that comes with Windows does not seem to understand how to read from the MSYS2 terminal.

[...]

Issue #65577, opened 2023-05-11
Accessibility issue, Character creation screen is a jumbled mess to parse with screen readers.


DoctorBoomstick:
### Describe the bug

This issue has been present for quite a while, but recently it has grown to encompass the entirety of the character creation process, from picking a scenario to defining skills. Throughout most of my time playing Cataclysm, navigating through any tab in the character creation screen has prompted the screen reader to narrate the specific information of any selected option while also interspersing it with the other options present in the menu. For instance, if I wish to browse through the professions tab and read the information associated with the hobo profession, the screen reader will read the relative information while also announcing the names of the other professions visible on the menu. This leads to a very confusing mix of irrelevant text mashed together.
While this is an issue in and of itself, in the past, a screen reader would reliably announce the name of the selected option as the first string of spoken text. This allowed a VI player to quickly parse what options were available before using the filter to search for and display only the particular option that interested them. This removes any cluttering text and allows one’s screen reader to narrate the text without any obstructions. This strategy, while clunky, worked with respect to the professions, hobbies, and traits tabs. However, when it came to scenarios, skills, and stats, the name of the chosen option was not the first element read by the screen reader and was, indeed, normally never announced.
As of late, this latter state of affairs now extends across all the tabs, making character creation an exercise in trying to parse important or needed information from a giant wall of random text while also trying to figure out which option is selected, as the name is never announced in a logical or dependable manner.

 

### Attach save file

N/A

### Steps to reproduce

1. Hop onto an installment of, in my case, Ubuntu 20.04.5.
2. Enable the natively installed Orca screen reader.
3. Boot up a pure text copy of the game, running in terminal.
4. attempt to create a character and be sufficiently confused in trying to figure out what the screen reader is meant to be announcing.

### Expected behavior

In an ideal world, it would be fabulous if a blind player could navigate over any given option and simply have their screen reader announce the related information without any intruding elements. Failing this, simply having the option name be the first element that a screen reader would present would be a massive improvement over the current state of affairs.

### Screenshots

This is a bundle of recordings I made to showcase how, at present, screen readers behave with the character creation screen. With respect to scenarios, I first attempt to read the information for the fungal-infected scenario before navigating down the list and trying to read the information for other scenarios. I do the same for both the professions and hobbies tabs, starting with the hobo profession in the former and sleeping pill dependence in the latter. Skills, traits, and stats work exactly the same as what’s presented here, so I didn’t include a recording of them
[...]

PR #65598, opened 2023-05-12
Fix UI and accessibility issues in the overmap UI and character creation menu

### LINKED ISSUE #65580: accessibility issue, any character creation option that isn’t the default custom creation selection will cause screen readers to bug out and not read anything.
### Describe the bug

Recently, it seems like any character creation option, barring the default custom creation style and play now, will cause screen readers to glitch and fail to read any text, even if the menu is exited. When I, for example, go to create a random character, the screen reader will comfortably read the final panel I’m dropped upon. However, should any key be pressed, the screen reader will cease reading any and all text from the game, requiring a reboot of the game in order to resume regular operations. This persists even if the screen reader itself is turned off and rebooted or the character creator is exited by finishing the generation process.
This makes the random character creator functionally unusable, as one cannot even tab back to review the selected options, pick a starting location, or, indeed, do anything in any way that uses a key press, including pressing tab to enter a game session. Indeed, the only way to both save a character generated in this manner and reacquire spoken feedback is to use hard muscle memory to quick save, save and quit, close the program, relaunch the game, and load into the saved character. Pre-saved character templates run into the same exact issues.

### Attach save file

N/A

### Steps to reproduce

1.	Hop onto an installment of, in my case, Ubuntu 20.04.5.
2.	Enable the natively installed Orca screen reader.
3.	Boot up a pure text copy of the game, running in terminal.
4.	Pick either the random character or the pre-saved character options when going to start a new game.
5. Wait until the data loads and you’re plopped upon the final panel of character creation. Press literally any button and try to navigate around, making note of the fact that the screen reader fails to read any and all text.

### Expected behavior

I expect that I’d be able to pick these options, press my keyboard’s not-so-shiny buttons, do game things, and have my screen reader read any information whatsoever.

### Screenshots

_No response_

• OS: Ubuntu 20.04.5
• Game Version: A3dd8c6
• Graphics version: Terminal
• Ingame language: English

### Additional context

_No response_

Qrox:
#### Summary
Bugfixes "Fix UI and accessibility issues in the overmap UI and character creation menu"

#### Purpose of change
Fixes #65580. Partially addresses #65578 and #65577.

#65580 is actually a UI issue, where the screen content is borked after the new character menu writes to a null window (`wprintz` with `w_hobbies` in small screen mode). It does not happen when entering custom character creation though, so there might some undefined behavior at play.

The cursor placement issues reported by #65578 and #65577 are fixed, although that does not solve all the problems so I'll not mark them as closed by this PR.

#### Describe the solution
1. Place cursor at the beginning of important text in the overmap and character creation UIs.
2. Add null window check to ncurses functions to not bork the screen when writing to a null window.
3. Skip drawing the null windows in character creation menu.
4. Update accessibility documentation with new information.

##### Other misc changes

- I needed to work around the locale dependent functions error on MSYS2 MSYS so I wrapped the error in a wrapper header and skipped the error on MSYS2 MSYS.

- In that process I also noticed that the Apache license of flatbuffer requires modifications to the original files to be noted in the respective files, so I check the git version history and added appropriate notices to the files.

- I also added a compile flag to turn on curses cursor display for debugging purposes.

There are a few other improvement options mentioned in #65578 but I'll leave them for later.

#### Testing
Tested on MSYS2 MSYS. The cursor is now placed right after the selected option, and random character creation does not break when pressing any key.

Also tested on Windows MinGW-w64 build and everything worked as before.

I also checked a flatbuffer cache file and its sha256sum is the same when produced by the MSYS and MinGW-w64 builds.

#### Additional context


Qrox:
I tested in an Ubuntu VM and it seemed to work better with the cursor placed at the start of the text as suggested by previous feedback: it prevents the screen reader from incorrectly announcing the next line if the end of the text is at exactly the right edge of the screen and the cursor is wrapped to the next line, and it's also easier for someone to read the nearby text by pressing numpad 6 to read the words one by one from the beginning. I've updated the code accordingly, but please let me know if my understanding is incorrect, because I'm not familiar with how people normally use a screen reader.

DoctorBoomstick:
I know this is already closed and merged, but wanted to chime in and say how fucking awesome it is that you put out these changes.

Zireael07:
CDDA is, from what I hear, one of the most accessible games in the genre. Great to see it strives to keep this distinction <3

PR #65599, opened 2023-05-12
Make makeshift sling not give survival skill

KittyTac:
#### Summary
Balance "Make makeshift sling not give survival skill"

#### Purpose of change
Can someone remind me:
1. Why does folding a blanket make you better at "survival"?
2. How did this survive (no pun intended) for years?

#### Describe the solution
Make it a skill-less craft like soap flakes and makeshift bandages.

Leaving as is, for some reason.

#### Testing
image

#### Additional context
Expect more crafting tweaks soon.

### DIFF
--- a/data/json/recipes/armor/storage.json
+++ b/data/json/recipes/armor/storage.json
@@ -1243,7 +1243,6 @@
     "activity_level": "NO_EXERCISE",
     "category": "CC_ARMOR",
     "subcategory": "CSC_ARMOR_STORAGE",
-    "skill_used": "survival",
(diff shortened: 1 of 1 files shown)


Karol1223:
Okay but does it even *work* if you have a recipe without a skill assigned? There's other methods to make it not train anything (though I am not sure how, I never bothered to check) but removing the assigned skill will likely just break this

RenechCDDA:
There are actually quite a few recipes without any skill assigned. The *swimming test* error is unelated.

PR #65493, opened 2023-05-05
Encumbrance fixes and PALS for some containers

anoobindisguise:
#### Summary
Balance "reduce extreme encumbrance values of certain items and give them PALS"

#### Purpose of change
Some items didn't have defined volume_encumber_modifier values or were set wildly high - this aims to correct that. This also adds PALS to a few items that seemingly should have them.

#### Describe the solution
- Ammo pouch: lower base encumbrance from 5 to 2, added mult of 0.3 (had none/1.0)
- Javelin Pouch: Lowered base encumbrance to 5 from 10, added mult of 0.3 (had none/1.0)
- Tac Vest: Previously had [2,14] for 5.4 liters of storage (which is an effective mult of 0.56) - reduced to 0.3 for holster, 0.25 for mag pouches and 0.2 for the shotshell container.
- Pistol Bandolier: added mult of 0.3 (had none/1.0) and PALS_SMALL
- Rifle Bandolier: added mult of 0.3 (had none/1.0) and PALS_MEDIUM
- Waist Shotgun Bandolier: added mult of 0.3 (had none/1.0)
- Waist Grenade Bandolier:  added mult of 0.3 (had none/1.0)
- Torso Shotgun Bandolier:  added mult of 0.3 (had none/1.0)
- (Leather) Stone Pouch:  added mult of 0.3 (had none/1.0)
- Large Grenade Pouch: added mult of 0.3 (had none/1.0)
- Cheap Ammo Bandolier: Lowered base encumbrance from 7 to 4 and added mult of 0.3 (had none/1.0)
- 556/308 belt pouches: Lowered base encumbrance from 7 to 4 and added mult of 0.3 (had none/1.0)
- Back Holster: Adjusted encumbrance from [3, 15] for 9 liters to 2 + mult 0.4
- Baldric Holster: added mult of 0.3 (had none/1.0)
- Belly Band Holster: Lowered encumbrance from [1, 6] to base of 0, 0.4 for the holster and 0.2 for the magazine (previously had an effective multiplier of 1.14)
- Large Holster: changed encumbrance from [2, 7] to [4, 5] (a standard holster had [3, 4]
- Axe Ring Holster: added mult of 0.3 (had none/1.0)
- Spear Strap: added mult of 0.5 (had none/1.0) and lowered base encumbrance from 2 to 1
- Book Strap: Lowered encumbrance from [0, 11] to base 0 and mult 0.35 (previously was effectively 0.92)

#### Describe alternatives you've considered
[...]

PR #65601, opened 2023-05-12
Axe ring empty encumberance to 0

Olaroll:
#### Summary
Balance "Axe ring empty encumberance to 0"

#### Purpose of change

An axe ring is just a little piece of leather and metal. I really don't see how the encumbrance would be any different to a regular leather belt, which is 0.

#### Describe the solution

I've set the empty encumbrance from 2 to 0. Inserting an axe still adds significant encumbrance.

Leaving it as is,

#### Testing

I tested it locally. Works as expected.

#### Additional context

Here is an image of an axe ring on a belt.
axering_small


anoobindisguise:
I actually already have a pending fix for it in #65493

bombasticSlacks:
I think both fixes are fine in parallel 😄

PR #65603, opened 2023-05-12
Bionic two way radio

anoobindisguise:
#### Summary
Content "Bionic radio system"

#### Purpose of change
Radio bionic is an obvious inclusion that was mentioned as being desirable by Erk.

#### Describe the solution
This bionic provides two-way radio functionality. Pretty simple, just lets you talk to your npc buddies without carrying a radio on your own.

I initially wanted to give it one-way-radio functionality (like the default "radio" item) but due to a number of factors this was FAR too much effort and hardcoded nonsense to enable what is basically just a flavortext functionality because the standard radio function is simultaneously half-baked and absurdly complicated and it led to all sorts of issues actually trying to implement it in a way that was functional.

I could also not call it "Infolink" as that's explicitly a Deus Ex thing, but I don't think that's a copyrighted term on DX's part and that's basically also what this is.

#### Testing
Installed the bionic. Gave an NPC buddy a two way radio and then traveled some distance. Confirmed I could chat with the NPC through my bionic.

### DIFF
--- a/data/json/bionics.json
+++ b/data/json/bionics.json
@@ -1348,6 +1348,15 @@
     "flags": [ "BIONIC_NPC_USABLE" ]
   },
   {
+    "id": "bio_radio",
+    "type": "bionic",
+    "name": { "str": "Infolink" },
+    "description": "A compact communications device has been built into your ears, enabling you to communicate over radio waves.",
+    "occupied_bodyparts": [ [ "head", 2 ] ],
+    "flags": [ "BIONIC_NPC_USABLE" ],
+    "passive_pseudo_items": [ "fake_radio" ]
+  },
+  {
     "id": "bio_radscrubber",
     "type": "bionic",
     "name": { "str": "Radiation Scrubber System" },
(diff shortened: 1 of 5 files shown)


PatrikLundell:
The only issue I have with this convenience CBM is how compatible an Exodii implementation would be with current real world ones, given that there are a few different ways to use radio waves to transfer information (FM, AM, etc.), as well as different ways to actually encode it. Also, they'd probably have a data transfer protocol (to e.g. transfer video or documents, etc.), although I guess they may have voice as a low end capability with is what the player would be able to make use of.
Come to think of it, they might actually not have digital, but instead be analog (but still be able to use old fashioned TV signals for video), given their preference for meat technology. Of course, you can still use analog signals to transfer digital data using modem technology.

bombasticSlacks:
> The only issue I have with this convenience CBM is how compatible an Exodii implementation would be with current real world ones, given that there are a few different ways to use radio waves to transfer information (FM, AM, etc.), as well as different ways to actually encode it. Also, they'd probably have a data transfer protocol (to e.g. transfer video or documents, etc.), although I guess they may have voice as a low end capability with is what the player would be able to make use of.

I think these points are fair, but until / if we ever get other aspects of ham radio, then we could look into adding scope to this.  As is, I think it's a fine inclusion.

PR #65605, opened 2023-05-12
Teargas spray

anoobindisguise:
#### Summary
Content "tear gas spray"

#### Purpose of change
Tear gas spray (high potency version of mace spray, bear spray, etc, using actual military tear gas instead of being oleoresin capsicum based "pepper" spray) is a real thing IRL and is civilian legal. I wanted to add some kind of pepper spray equivalent and this made the most sense to me, being fairly high power would mean it would be deployed against the "rioters" and it uses the pre existing tear gas effect instead of needing to invent a new pepper spray thing.

#### Describe the solution
Mostly copied from fire extinguisher implementation, it spawns a cloud at an adjacent tile. It takes 2 seconds to apply, but if you hit something that breathes (not a zombie, robot, alien, etc) it gets stunned for 3 seconds. Like fire extinguisher it has a 1/2 chance to be blinded for 1-2 minutes. There is a moderate intensity tear gas cloud that fades away quickly.
Most cartridges like this can be used for about 20-25 seconds before being depleted so I gave it 10 charges. It's basically pocket sized and it has belt clip, like many existing pepper or tear gas sprays I've seen sold on amazon (and dimensions/weight are based on such figures) 
It primarily spawns on law enforcement but also on various zombies as it is a sensible everyday carry.

Call it something else? Tear gas spray canister, bear spray, mace spray, etc. Unsure if I like "tear gas sprayer" or not.

#### Testing
Spawned the tear gas spray. Maced a bear, and watched as the bear suffered. Then tried to mace a zombie and just got mauled.


bombasticSlacks:
Over all this is 👍 , however I believe there would be value in moving this over to a spell/EOC, especially since this isn't even a generalized iuse to create a generic aerosol. If it is outside your capability, or there are spell/EOC functionality blockers that's a'okay just figured I would mention.

anoobindisguise:
> Over all this is 👍 , however I believe there would be value in moving this over to a spell/EOC, especially since this isn't even a generalized iuse to create a generic aerosol. If it is outside your capability, or there are spell/EOC functionality blockers that's a'okay just figured I would mention.

Yeah I'm not sure if "create cloud" action will work since the act of spawning the cloud has specific effects on the mob there such as blinding it or something like that. Also probably outside of my capacity lmao

PR #65606, opened 2023-05-12
Make portal storms mobile

Ramza13:
#### Summary
Features "Make portal storms mobile."

#### Purpose of change
Make portal storms avoidable.
Make portal storm dungeons no longer a matter of finding a structure.
Add EOC die command so nothing can stop dev's evil!!
Add first official means of escaping the cataclysm which doesn't technically kill you for some definitions of kill.

#### Describe the solution
Portal storms now  have a center and a mission to tell you where it is.  The center  periodically moves randomly.  There are 3 levels of storm depending on how close to the center you are.  They are similar with more of a chance of events closer to the center.  Portal storm dungeons are no longer accessed by a location.  Instead an effect in the center of the storm starts a conversation that can take you there.

#### Additional context
Feedback is always welcome.  The exact values of storm size, speed and whatnot are subject to change.


data/json/portal_storm_effect_on_condition.json
```diff
@@ -136,16 +136,47 @@
...
+    "condition": { "or": [ { "math": [ "cause_early_portal_storm", "==", "1" ] }, { "math": [ "cause_portal_storm", "==", "1" ] } ] },
+    "deactivate_condition": { "and": [ { "math": [ "cause_early_portal_storm", "!=", "1" ] }, { "math": [ "cause_portal_storm", "!=", "1" ] } ] },
+    "effect": [
+      {
+        "location_variable_adjust": { "global_val": "portal_storm_center" },
+        "overmap_tile": true,
+        "x_adjust": [ -1, 1 ],
+        "y_adjust": [ -1, 1 ]
+      },
+      { "u_message": "The portal storm shifts!", "type": "bad" },
```
bombasticSlacks, line 175: I'm a little worried this could get annoying.

bombasticSlacks:
Very cool!

data/json/portal_storm_effect_on_condition.json
Ramza13: Fair enough, I've removed the message.

Nebnis:
I'm not sure if this is the right place to portal storms feedback but here i go

Having played the new portal storms, i would say it is a lot better than before, it is almost perfect with them being mobile weather now. Having a bunch of its danger be immobile was also a really good idea which means you gotta pay attention where you going and a mistake will get you crippled bad, it seemed fun to be caught into a forest rather than "oh bother i gotta start another character"

In my opinion i feel the biggest issue is that the effects spawn TOO OFTEN, way way way too often and even an endgame character will get massively crippled if they spend even a few min outside much less the entire duration having to chase one to the center for the dungeon/rewards.

*The constant movement restriction from the effects happening often make too swingy about you getting fucked because you constantly getting caged by darkness or rewinded every 4~5 turns. Maybe i got lucky but if the frequency of this effect was toned down, this would not be an issue at all. Maybe it could be one of the raise ires thing?

*The random pain gotta go tbh, it just gotta go. It gives you massive pain out of nowhere with basically no counter play. Which snowballs a lot,this is hands down the worst effect from everything happening now.

*Shifting masses/chunks of unknown will decimate you with their toughness/durability  because of the outlined above since you can just randomly get stuck or get max tier pain out of nowhere which will spiral into death fast. They dont despawn either. Also they don't even drop artifacts anymore due to some bug, i killed dozens of them with no artifact drop.

Hopefully you will see this :P

bombasticSlacks:
Absolutely not the place for portal storm feedback lol. Even if people do read this, no one is going to think to check PR 65606 for portal storm feedback if working on it in the future

PR #65611, opened 2023-05-13
Post Thresh traits shouldn't be starting traits

Maleclypse:
#### Summary
None

#### Purpose of change
Someone mentioned to me that Adrenaline is a post thresh trait that is a starting trait also.  So I went and audited mutations in vanilla to see if there were others.  There are not so that's the only change.

#### Describe the solution
Removes starting trait true from Adrenaline

none

#### Testing
CI

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -983,7 +983,6 @@
     "name": { "str": "High Adrenaline" },
     "points": 1,
     "description": "If you are in a very dangerous situation, you may experience a temporary rush which increases your speed and strength significantly.",
-    "starting_trait": true,
     "changes_to": [ "ADRENALINE2" ],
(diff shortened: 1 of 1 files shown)


TaiGambol:
Just for the sake of argument, would making it not post-thresh instead be reasonable? Or are there reasons beyond 'this is post-thresh' to nix it from starting traits?

Maleclypse:
> Just for the sake of argument, would making it not post-thresh instead be reasonable? Or are there reasons beyond 'this is post-thresh' to nix it from starting traits?

I don't want to sound argumentative here because I don't actually care either way.  But if you put up a PR to do that you'll also need to adjust Adrenaline2 to be post thresh.  I have literally no plans to ever touch or comment on this PR again until I delete the branch after merge or when someone contacts me on discord to let me know that an alternative response has been merged.

PR #65613, opened 2023-05-13
Allow zones to interact across multiple Z levels

FatigueDev:
#### Summary
Features "Allow zones to interact across multiple Z levels"

#### Purpose of change
In the game as it is currently, zones that are created on the same tile can interact just fine; but they cannot interact if the zones are above or below on the Z level. This can allow players to create surface level bases, set mining jobs below and use tools in the attic to perform those tasks, if they are within reach.

#### Describe the solution
By removing a section of checks in `clzones.cpp` to make sure that the zone is on the same z level, it allows them to interact using the basic sorting jobs and stuff like that. The changes to `activity_item_handling.cpp` are required so that Players + NPCs can find their tool zones properly, as it uses it's own logic to find loot zones. Now players can tell NPCs to mine underground with tools from the surface.

N/A

#### Testing
- Tested organizing loot from z level 0 to z level -1 and vice-versa. No issues.
- Tested automated jobs, chopping wood using an axe from z level -1 and a tree on z level 0. No issues.
- Tested telling a miner (z level -1) to do a mining job (z level -1), but their tools were on z level 0. No issues; went up the stairs, came back down without a care in the world and got to work.

### DIFF
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -2699,7 +2699,8 @@ static requirement_check_result generic_multi_activity_check_requirement(
             loot_zone_spots.emplace_back( elem );
             combined_spots.emplace_back( elem );
         }
-        for( const tripoint_bub_ms &elem : here.points_in_radius( src_loc, PICKUP_RANGE - 1, PICKUP_RANGE - 1 ) ) {
+        for( const tripoint_bub_ms &elem : here.points_in_radius( src_loc, PICKUP_RANGE - 1,
+                PICKUP_RANGE - 1 ) ) {
             combined_spots.push_back( elem );
         }
         add_basecamp_storage_to_loot_zone_list( mgr, src_loc, you, loot_zone_spots, combined_spots );
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -970,7 +970,7 @@ static bool are_requirements_nearby(
     }
     // use nearby welding rig without needing to drag it or position yourself on the right side of the vehicle.
     if( !found_welder ) {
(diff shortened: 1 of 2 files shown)


PatrikLundell:
Sounds like a fantastic improvement I've wanted for a long time (but the suggestion has long since been buried by the stale-bot, as per usual, so there's no point in trying to find it and link to it).

I would like to have it verified that pathing works correctly in failing cases as well, such as e.g. construction zones where there is no path up to them, or no place to stand except the construction zone itself. What I'm worried about is the risk of eternal (or every turn) failures and retries causing trouble. We already have weird cases of companions selecting an activity site and the trying to walk through a wall (that's what I think they do, anyway) without reaching the destination, but also without cancelling or spamming errors, so I wouldn't consider such failures to be new.
There are probably additional problematic cases that ought to be tested, but I'm not great at coming up with tests, unfortunately.

FatigueDev:
> Sounds like a fantastic improvement I've wanted for a long time (but the suggestion has long since been buried by the stale-bot, as per usual, so there's no point in trying to find it and link to it).
> 
> I would like to have it verified that pathing works correctly in failing cases as well, such as e.g. construction zones where there is no path up to them, or no place to stand except the construction zone itself. What I'm worried about is the risk of eternal (or every turn) failures and retries causing trouble. We already have weird cases of companions selecting an activity site and the trying to walk through a wall (that's what I think they do, anyway) without reaching the destination, but also without cancelling or spamming errors, so I wouldn't consider such failures to be new. There are probably additional problematic cases that ought to be tested, but I'm not great at coming up with tests, unfortunately.

Luckily, it looks as though if there is a pathing failure.. They just can't find what they need for the job. If they cannot reach a zone physically, they skip it; much like how closed doors will stop you from sorting the whole zone! 
image

PR #65625, opened 2023-05-13
nonlethal tear gas

anoobindisguise:
#### Summary
Balance "Tear gas doesn't kill mobs"

#### Purpose of change
Tear gas is more or less non lethal except in extreme cases, and it didn't damage the player, but it dissolved enemies alive like it was a sulfuric cloud which is silly.

#### Describe the solution
Remove damage from all instances including the special-cased extra damage to VEGGY monsters.

Make it do a bit of damage but not much, and only at high intensities, as I've read CS gas can cause chemical burns, permanent blindness, RTIs and other unpleasantness. I spoke with @Venera3 who says symptoms like that are inappropriate for the tear gas clouds so I've left out any damage for now.

### DIFF
--- a/src/map_field.cpp
+++ b/src/map_field.cpp
@@ -1882,16 +1882,10 @@ void map::monster_in_field( monster &z )
             if( z.made_of_any( Creature::cmat_fleshnveg ) && !z.has_flag( MF_NO_BREATHE ) ) {
                 if( cur.get_field_intensity() == 3 ) {
                     z.add_effect( effect_stunned, rng( 1_minutes, 2_minutes ) );
-                    dam += rng( 4, 10 );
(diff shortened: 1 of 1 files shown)


Nebnis:
I agree with you that tear gas should have some small damage on high intensities to simulate your internal organs being burned and being in an enclosed area; People have died from breathing it in an enclosed area "https://www.opendemocracy.net/en/openindia/tear-gas-touted-safe-against-protesters-so-why-it-killing-so-many-kashmiris/". Wiki has more sources too. "https://en.wikipedia.org/wiki/CS_gas"

>It took several requests before they relented, giving the cause of death as an “acute lung injury” from inhaling toxic chemicals.

bombasticSlacks:
I think this is good as is, more work could be done in the future.

PR #65628, opened 2023-05-14
Burdock stalks

NetSysFire:
#### Summary
Content "Add burdock stalks"

#### Purpose of change

Realism. Both cattails and burdock plants have edible stalks. See e.g https://joshfecteau.com/foraging-wild-shoots-common-burdock/ and https://foragerchef.com/burdock-flower-stalks/

#### Describe the solution

Adds to new items: Burdock stalk and cooked burdock stalk. Both identical to their cattail counterparts.
Also does the same to the recipe. Adds a harvest entry for summer for the burdock plant, so you can get stalks from it that way.

Introducing a generic plant_stalk item. Would potentially require some migration and I am unsure about that one anyways. As you can see in https://www.eattheweeds.com/cattails-a-survival-dinner/, the lower stalks picture, they do look a lot like leek. Burdock stalks look much firmer and quite potentially have minor differences but there is no way to look them up for now as no one (e.g the USDA) formally looked into that yet.

#### Testing

Applied locally. Loads fine, I can harvest and process it just fine.


NetSysFire:
Of course I see 10 minutes after pressing submit that there are dozens of other (although smaller) plants with edible stalks. I'll literally sleep over it (although it is fine to merge as is, it is just the question of what is most efficient) I guess.

PatrikLundell:
Interesting use of the word "literally" in the comment. Using "literally" in a figurative sense that's actually not incorrect (unlike the horde of people who mistakenly think "literally" is to be used when "virtually" is not strong enough a word)...
My first reaction, however, was that you were sleeping outdoors (e.g. camping) on top of such plants.

NetSysFire:
No, I meant that I was, and unfortunately still am, trying to sleep. Fun pun though, did not think about that!

Maleclypse:
Interesting.  You've impacted the calorie test for Lobster Platter by adding burdock to the requirements.  It must have been on the edge of failing already.  
```
ms)=>   1260.0f <= 1259.3479591837
(~[slow] ~[.],starting_items)=> RecipeID: lobster_platter, default is 1680 Calories,
(~[slow] ~[.],starting_items)=> Current recipe range: 1099-[150](https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/4970985829/jobs/8895667715?pr=65628#step:16:151)0, Average 1259.3, Stddev 92.2
(~[slow] ~[.],starting_items)=> Average recipe Calories must fall within this range, derived from default Calories: 1260-2100
```

NetSysFire:
I saw that.

```
Error: (~[slow] ~[.],starting_items)=>   1260.0f <= 1259.3479591837
```

How do I even resolve that issue of 0.6520408163 missing calories?

Maleclypse:
> I saw that.
> 
> ```
> Error: (~[slow] ~[.],starting_items)=>   1260.0f <= 1259.3479591837
> ```
> 
> How do I even resolve that issue of 0.6520408163 missing calories?

Adjust the lobster platter item's calories to something more likely to pass like 1300.

Qrox:
> 1260.0f <= 1259.3479591837

That means it expected the calculated calories (the right side) to be greater than the lower bound (the left side), but the actual values did not meet the expectation, so the calorie value should be reduced not increased.

PR #65629, opened 2023-05-14
cloaks are belted layer

anoobindisguise:
#### Summary
Balance "cloaks go on the belted layer"

#### Purpose of change
Cloaks are meant to be thrown rather loosely over your body and should not realistically conflict with your worn armor. Meanwhile, they might actually conflict with exterior pockets since that means they can't lie flat against the body.

#### Describe the solution
BELTED layer for cloaks.

Mess with encumbrance.
Give cloaks the "draped over legs" layer for the legs.

### DIFF
--- a/data/json/items/armor/cloaks.json
+++ b/data/json/items/armor/cloaks.json
@@ -17,7 +17,7 @@
     "warmth": 20,
     "material_thickness": 0.3,
     "environmental_protection": 1,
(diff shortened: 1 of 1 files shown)


fairyarmadillo:
This will be a boon for spider, crab, snail, and cephalopod mutants who don't always have a reasonable way to keep warm.

Maleclypse (approved):
I think this makes sense

Issue #64778, opened 2023-04-02
[Mini-Project] Horsies


Procyonae:
### Is your feature request related to a problem? Please describe.

Horses want change

### Solution you would like.

Extend horses through a list of PR's performed collaboratively by myself and "horse expert" @Armokorguest , further suggestions are welcome, equally I'd love to know about anyone else's relevant PRs in the works

**DONE**
- [x] #65635 Revamp horse farm into a mutable map with new riding gear and additional animal related vehicles 

**IN-PROGRESS**

**TO-DO** (listed roughly in order likely to be implemented)
- [ ] Add more house/stable variants to the horse farm
- [ ] Add riding background, profession(s) and NPC(s) that spawn at the horse farm
- [ ] Horse breeds (implemented in the same way as dog and cat breeds, using monstergroups for spawns)
  - [ ] Megafauna changes to accommodate breeds?
- [ ] Spook alterations and implementation of mount obedience and training (training likely through recipes though minor gains through actual riding would be nice) Add tracking of mount on map.
- [ ] Riding proficiency/ies
- [ ] Trailer/articulated vehicle changes
- [ ] Animal inventory changes (single item removal, adv inventory support)
- [ ] Mutant horses if #63997 or a similar PR gets merged
- [ ] More map variants for the horse farm buildings and more horse related maps (stuff like a harness track w/ casino, large animal vet, rescue centre, riding school etc)

### Describe alternatives you have considered.

Letting the horses stew

jiornojiovanni:
Maybe some way to remember the horse last position? Even tho you can do that with the carriage

PatrikLundell:
Horses that can be ridden by people that aren't "height challenged" and aren't "of generous proportions" (how I hate politically correctness bullshit euphemisms!) would definitely be welcome.

Another horse related issue that might be worth looking at is to stop them from escaping their harnesses whenever spooked, as that's what caused me to give up trying to use horses altogether.

If a horse last position marker was used, it should be exactly that, rather than a GPS tracker using a GPS network that's no longer available.

fairyarmadillo:
Can't little people ride dogs, and big people ride larger animals, like cows and moose?

Something I'd really love to see is for certain constructions (IE palisades) to require an animal or vehicle and some rope to help pull/lift parts into place. Imagine using a donkey, a tree, and a pulley to lift an engine block so you could fix it. That'd be sweet as hell.

PatrikLundell:
[...]

PR #65635, opened 2023-05-14
Horse Farm Revamp

Procyonae:
#### Summary
Content "Make the horse farm mutable and add variance to it as well as flavour through the introduction of some riding gear and vehicle changes. Collab with @Armokorguest."

#### Purpose of change
The horse farm looks sad next to the new style mutable farms and there's a general lack of horse themed stuff, first part of #64778

#### Describe the solution
Horse farm is now a mutable map with the original being obsoleted but nearly all of the original map will be reused (The chicken coop hasn't been reused bc I used the already existing "yard" map which has a coop variant). The original map is still used on the Isherwood farm.
Adds a horse trailer, reinforced yoke and harness, riding helmet, riding crop and dressage whip.
Alters some animal-drawn vehicles to use reins over controls.
I appreciate this is alot of lines for what it is, I have tried to cut down where possible and alot of this should be reusable when someone next touches ranches.

- The forest potentially generated by the map is kinda ugly, see commit [7e05aee](https://github.com/CleverRaven/Cataclysm-DDA/pull/65635/commits/7e05aee6082f43a673ba845c941c9395c54262eb) for why I changed it to work this way.
- Flowchart style joins for the paddocks would be prettier in game but make the mutable way less readable and add to the line count even more, if this is wanted regardless I'm happy to implement them.
- I was planning on adding house and stable variants and adding a garage/garden spawn but I just want this done now, will most likely make a second PR addressing this eventually if noone else does first. The way I've set everything up aims to be very accessible if someone wants to just make a house/stable etc map and add it, happily contact me if you want help/explanations.
- Entrance roads should normally be made partially gravel/concrete instead of dirt but line count already be stoopid high.
- The paddock building blocks aren't super simple but I couldn't figure out a better solution while never ending up with doubled up fences and without creating an unintelligible use of nests based on neighbours.
- Only having the choice of poisson for max distributions is pretty restrictive I'm looking into making a PR expanding options such as uniform distribution option for [ 0, 1 ] type stuff, being able to do binomial and combinations of poisson/binomial with bounds would be nice too.
- Ideally I'd change over the farm house and stables to more standard palettes but given the Isherwood variant is staying as is there doesn't seem much point for now as it would just make any future mutual changes of the maps more awkward, if anyone adds house/stable variants etc in the future I wouldn't recommend using the palette.
- I considered adding a temporary fence type (the type that get moved around as needed) but couldn't decide how I'd want it to behave.
- The horse trailer has an extra wheel at the front where a coupler would be irl so it can be towed (vehicles need to have balanced wheels to be towed), I'm planning on looking into adding functional couplers in a future PR but I didn't want to add a dummy coupler part for now because I felt that would just confuse peeps.
- Irl horse trailers appear to use 15" wheels but I've used the generic 17" wheels instead of adding 15" trailer wheels or using the inappropriately specific 16" wheels. Equally it could have ramped doors on the back but that seems unnecessary.
- The farm vehicles json file also has all the vehicle blueprints facing north rather than east which I'm happy to switch in a separate PR if wanted.
- The riding helmet and crop use epoxy in place of fiberglass in line with the motorcycle helmet, I also calced the thicknesses from cross-section images which isn't exactly accurate but I'm not about to ship one to me and cut it open and couldn't find individual material thickness measures anywhere.
- The riding whip and crop serve no function (outside of being awful weapons) currently but will in a future PR.

#### Testing
Spawned the map lots, haven't tested old worlds yet. The mutable_overmap_placement test run on only this map failed with `assertions: 4689 | 4594 passed | 95 failed`. Unsure of the cause.

#### Additional context

Here's a graphic to try to show all the non extra tiles in the order they appear in the mutable file with each 3x3 of pixels representing an overmap:

Mutable Tiles With Text

The size of the generated map varies quite a bit depending on how blocked in it is and if it picks end tiles early, here's a few examples

image


data/json/items/vehicle/animals.json
```diff
@@ -17,5 +17,25 @@
     "width": 4,
     "looks_like": "foldframe",
     "melee_damage": { "bash": 3 }
+  },
+  {
+    "id": "yoke_harness_heavyduty",
+    "type": "WHEEL",
+    "name": { "str": "reinforced yoke and harness", "str_pl": "reinforced yokes and harnesses" },
+    "category": "veh_parts",
+    "description": "Heavy steel beams and harness to attach a sufficiently large creature to a wheeled vehicle, allowing them to pull it along, reinforced to withstand the rigours of the apocalypse and ill-adviced offroading. You would still need separate armor for said creature to wear.",
```
Maleclypse, line 26:
```suggestion
    "description": "Heavy steel beams and harness to attach a sufficiently large creature to a wheeled vehicle, allowing them to pull it along, reinforced to withstand the rigors of the apocalypse and ill-advised off-roading.  You would still need separate armor for said creature to wear.",
```

data/json/vehicleparts/vehicle_parts.json
```diff
@@ -129,6 +129,48 @@
...
+    "noise_factor": 10,
+    "fuel_type": "animal",
+    "durability": 600,
+    "item": "yoke_harness_heavyduty",
+    "m2c": 90,
+    "contact_area": 60,
+    "rolling_resistance": 1.5,
+    "wheel_offroad_rating": 0.7,
+    "wheel_terrain_modifiers": { "FLAT": [ 0, 3 ], "ROAD": [ 0, 1 ] },
+    "description": "Attach this part to a beast of burden to allow it to pull your vehicle. This one is reinforced with heavy chains and steel.",
```
Maleclypse, line 152:
```suggestion
    "description": "Attach this part to a beast of burden to allow it to pull your vehicle.  This one is reinforced with heavy chains and steel.",
```

Maleclypse (commented):
Cool stuff

Procyonae:
Do I need to rebase or is solving the merge conflict through the button below the tests enough?

PR #65638, opened 2023-05-14
Typo hunting 2

### LINKED ISSUE #65637: Translation errors
### Describe the bug

------------------------------------------------
strings has "your strike" instead of "you strike"
`Your attacks are a blur of hands and legs that become faster as your strike your opponents without rest.`

`-4 move cost.`
`Lasts 3 turns.  Stacks 4 times.`

------------------------------------------------
string mention "Pincer Strike", but I can't find other reference to it in other strings, therefore I believe this might
be a typo for the "Pincer Fist", for which there are other references. Not sure though, this require investigation.
`Rush forward and catch your prey!`

`+10% damage.`
`Enables "Pincer Strike" technique.`
`Lasts 2 turns.  Stacks 2 times.`

------------------------------------------------
string says "You are terrifying monster" instead of "You are a terrifying monster"
`You are terrifying monster on the battlefield!  Opponents will have a harder time fighting back.`

`+1 accuracy, +1.0 Dodging skill.`
`Lasts 3 turns.  Stacks 2 times.`

------------------------------------------------
string itself is correct, but the developer notes shows this is the description of a "mitre saw" that has been also applied to 'vehicle part "band saw". The "band saw" has it's own description so they should be separated.
`A circular sawblade on an arm that can slide and rotate in several directions, this is a staple tool for nearly any carpentry.  Plugged in and ready to go.`

`DEVELOPER NOTES`

`~ Description of vehicle part "mitre saw"`
`~ Description of vehicle part "band saw"`

------------------------------------------------
string says " 'e'xamine the tile" but I suspect the keybind e shouldn't be hardcoded into the string, in case people change keybinds
`A water purifier, powered by the vehicle's batteries.  You can use it to purify water in a container or in the vehicle's tanks.  'e'xamine the tile with the purifier to use it.`

------------------------------------------------
string should probably say "An electric forge" instead of "A electric forge" ? Not 100% sure, I'm not native english
`A electric forge for metalworking, powered by the vehicle's batteries.  With a hammer and other tools, you could use this for metalworking.  If you attempt to craft an item that needs a forge, you will be given the option of selecting it as a tool.`

------------------------------------------------
string says "You can take this down and plug it into a power grid to use it." but all other strings say "You can take this down and plug it into your own power grid to use it." so this one should be changed for consistency.
`A powerful drill mounted on a slide that lets it drop precisely down.  Useful in all kinds of projects from industrial fabrication to home woodworking.  You can take this down and plug it into a power grid to use it.`

------------------------------------------------
the 3 strings below are all associated to the description of a "table" so this is all the same item in game in the form of furniture, item and vehicle part.
Therefore "router" and "disconnected router" should become "router table" and "disconnected router table"
`router`
`router table`
`disconnected router`

------------------------------------------------
this 2 strings below disagree on what a power cord is, in the first case is a cord+plug and in the second is a cord+outlet.
I'm not saying this is wrong, maybe there are many types of power cords. I'm simply bringing up the fact I've noticed a discrepancy

`A power cord sticking out of an appliance.  You need to plug it into a powered grid for the appliance to work properly.`
`A short, multi-stranded copper cord with a power outlet at the end.`

------------------------------------------------
the two strings below are the description for the item "water wheel" and "large water wheel" - the fact they mention "recharge a vehicle" probably means they where copy-pasted descriptions from the vehichle parts, but this is probably inaccurate because I believe you can hook this to your power grid on your river house.
`A water wheel that will slowly recharge a vehicle's electrical power when built over shallow moving water.`
`A large water wheel with wooden supports.  It will recharge a vehicle's electrical power when built over shallow moving water.`

GuardianDll:
#### Summary
None
#### Purpose of change
Close #65637 and one that was mentioned in discord/localization


data/json/furniture_and_terrain/appliances.json
```diff
@@ -303,7 +303,7 @@
     "symbol": "7",
     "color": "yellow_cyan",
     "categories": [ "utility" ],
-    "description": "A circular sawblade on an arm that can slide and rotate in several directions, this is a staple tool for nearly any carpentry.  Plugged in and ready to go.",
+    "description": "Power saw with long, toothed blade, stretched between wheels.  Plugged in and ready to go.",
```
Karol1223, line 306:
```suggestion
    "description": "A power saw with long, toothed blade, stretched between wheels.  Plugged in and ready to go.",
```

PR #65639, opened 2023-05-14
Reintroduce light snares

RenechCDDA:
#### Summary
Content "Added craftable/findable light snares"

#### Purpose of change
I want to (force) feed the rabbit. The rabbit moves at 160(!!!) speed. I cannot catch the rabbit.

(Snares were removed in the great mod obsoleting of 2020, along with the 'More Survival Tools' mod. They were only moved to the mod because... apparently they had stopped working? I don't think that was a good reason but it happened.)

#### Describe the solution
Dust off some old code, write some new code. Do a bit of housecleaning wherever references to the lightsnare effect could be found. (Heavysnare *was not touched* and still lingers like a disease). Re-add the old item. Re-add the old trap. Redo all the values that don't make sense. Add new crafting recipe from scratch. Add item (commercial pre-made kits) to some appropriate existing item groups.

Maybe I'll do heavysnare next...

#### Testing
Youarealreadytamed.jpg
image

#### Additional context
We still have tileset support for the snare trap! That's awesome! (It's been wasted since 2020 but that's the tilesets' problem!)

In the process of adding it to item groups I replaced an existing entry in tools_survival. An acetylene gas machine is NOT a survival tool.

### DIFF
--- a/data/json/effects.json
+++ b/data/json/effects.json
@@ -4191,5 +4191,15 @@
     "desc": [ "You are covered in glowing nether spores!" ],
     "apply_message": "The glowing spores settle on you!",
     "rating": "bad"
+  },
+  {
+    "id": "lightsnare",
+    "type": "effect_type",
+    "name": [ "Snared" ],
+    "desc": [ "For such a tiny creature, you're in a very big pickle." ],
+    "rating": "bad",
+    "show_intensity": false,
+    "//": "Don't multiply speed by 0 or else the effect will never be removed for creatures with 0 damage(e.g. rabbits)",
+    "enchantments": [ { "values": [ { "value": "SPEED", "multiply": 0.01 } ] } ]
   }
 ]
--- a/data/json/recipes/tools/tools_primitive.json
+++ b/data/json/recipes/tools/tools_primitive.json
@@ -420,5 +420,33 @@
     "reversible": true,
     "autolearn": true,
     "components": [ [ [ "material_sand", 1 ] ], [ [ "leather", 1 ] ] ]
+  },
+  {
+    "type": "recipe",
+    "activity_level": "LIGHT_EXERCISE",
+    "result": "light_snare_kit",
+    "//": "Carving notches into the sticks (~15m), looping and knotting the string to sit securely in them (~30m), testing/misc/overhead (~5m)",
+    "category": "CC_OTHER",
+    "subcategory": "CSC_OTHER_TOOLS",
+    "skill_used": "survival",
+    "difficulty": 3,
+    "decomp_learn": 1,
+    "//2": "Pop culture magazine lacks most helpful information, requires inferring and therefore doesn't learn the recipe until survival 2",
+    "book_learn": [
+      [ "mag_survival", 2 ],
+      [ "pocket_survival", 1 ],
+      [ "manual_survival", 1 ],
+      [ "textbook_survival", 1 ],
+      [ "survival_book", 1 ]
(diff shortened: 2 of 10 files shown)


data/json/effects.json
```diff
@@ -4191,5 +4191,15 @@
...
+  },
+  {
+    "id": "lightsnare",
+    "type": "effect_type",
+    "name": [ "Snared" ],
+    "desc": [ "For such a tiny creature, you're in a very big pickle." ],
+    "rating": "bad",
+    "show_intensity": false,
+    "//": "Don't multiply speed by 0 or else the effect will never be removed for creatures with 0 damage(e.g. rabbits)",
+    "enchantments": [ { "values": [ { "value": "SPEED", "multiply": 0.01 } ] } ]
```
GuardianDll, line 4203: Enchantments doesn't work with monsters, use another way

RenechCDDA: They do though? The rabbits in testing are all sitting perfectly still.

RenechCDDA: https://github.com/CleverRaven/Cataclysm-DDA/assets/84619419/c5b52e19-80ab-430d-9b88-9543eecabac2

GuardianDll: Maybe speed is an exception?
IMG_20230514_193830

GuardianDll:
Btw, does it has a limitation on size, or you can trap a hulk using one?

Nebnis:
Please do heavy next, i wanna catch DEERS

RenechCDDA:
> Btw, does it has a limitation on size, or you can trap a hulk using one?

It is limited to creatures of size small and tiny, and they have a chance *per second* (not per turn! they get very few turns) of breaking free equal to their melee dice / 12. So if their melee dice is 2d3, their chance is equal to (2 * 3) / 12 = 0.5 (50%) per second.

Rabbits have 0d0, so rabbits are unfortunately stuck there until the effect wears off (100 turns later, or approximately 3 hours of struggling for most creatures).

data/json/recipes/tools/tools_primitive.json
```diff
@@ -420,5 +420,33 @@
...
+      [ "manual_survival", 1 ],
+      [ "textbook_survival", 1 ],
+      [ "survival_book", 1 ]
+    ],
+    "time": "50 m",
+    "qualities": [ { "id": "CUT", "level": 1 } ],
+    "proficiencies": [
+      { "proficiency": "prof_carving", "max_experience": "15 m" },
+      { "proficiency": "prof_fibers", "max_experience": "30 m" }
+    ],
```
TheShadowFerret, line 447: No Principles of trapping?

(Also Innawoods autolearn override plz ❤️)

RenechCDDA: >No 5m Principles of trapping("testing/misc/overhead (~5m)")?

You already get that exp during setting/removing the trap, and the proficiency says it's for doing that (and only that).

>(Also Innawoods autolearn override plz ❤️)

What do you mean? It already autolearns at survival 3. Should be true in innawoods too.

TheShadowFerret: >What do you mean? It already autolearns at survival 3. Should be true in innawoods too.

I can't read :D

TheShadowFerret: >It already autolearns at survival 3.

I can't read :(

Issue #63578, opened 2023-02-13
Low-tech variant of tree_spile / change tree tapping code?


merrygin:
### Is your feature request related to a problem? Please describe.

Tree spiles for tapping maples and harvesting maple sap should be possible to craft with relatively low technology available, i.e., without steelworking. 

It's a technique NE Native Americans used long before settlers arrived, probably using shaped bark or wood as spiles. So a) it makes sense to have it, but also b) seems like a fun addition for more wilderness oriented playthroughs. I originally planned to add it in #61896 but seemed too involved to get finished quickly and since I don't know enough c++.

The main problem is that the tapping functionality of the tree spile is coupled explicitely to the metal "tree_spile" item in iexamine.cpp.

### Solution you would like.

@flybye3o6 suggested on discord (see below) that you could just (mostly?) duplicate the tapping method in iexamine.cpp for a new tree_spile_2 item. This would be a good middle way between the two alternatives mentioned below, but maybe it is unnecessary amounts of additional code?

### Describe alternatives you have considered.

Easiest alternative solution would be to just make a crafting variant of the tree spile and change the description of the item so it doesn't say it is specifically made of metal. Downside is that the items material properties would not be representative of the actual crafting recipe used (you would probably change it from being "material: "steel"" to "material: steel, wood"). So this would definitely be only a provisional change.

Another option would be to totally change the way the tapping works to instead make the code look for an item flag/tool quality when the player examines a maple:

1. Pro would be greater codewise elegance and adaptability for other tapping uses in the future (e.g., tapping pines for tar, tapping birches for sap). 
2. Con would be that actual c++ is needed different than to the other solutions; especially the code for taking away the spile from the tree would need to change.

### Additional context

flybye306's verbatim from discord for future reference:
"you will probably need to make another terrain t_tree_maple_tapped to differ the 2
in iexamine.cpp line 3887  iexamine::tree_maple and line 3929 iexamine::tree_maple_tapped is what you wanna be talking a look at...
you will need the new tree_spile_2 to have a different result, otherwise when you go to take it off your cheap tree spike it turns into the default one. 
[...]

PR #65641, opened 2023-05-14
Use tool quality for maple tapping

ferociousdork:
#### Summary
Content "Use tool quality for maple tapping"

#### Purpose of change

Use a new tool quality for maple tapping, instead of the hardcoded itype. This allows for new tree tapping items, this PR also introduces a makeshift tree spile made of wood, and a way to craft it.

Implements #63578.

#### Describe the solution

- Introduce a new tool quality, TREE_TAP
- When examining a maple tree, allow the player to select an item with this quality to use as tap.
- The tap item is then put on the map tile of the tree.
- Removing the tap returns the original item.

I've added some QoL changes as well:
- Take the hammer and drill from crafting inventory instead of player inventory.
- Removing the container puts it on the tile instead of the player picking it up, as this would fail when the player had no room in their inventory. 

With regards to the makeshift tree spile: I've used https://bluehillheritagetrust.org/wp-content/uploads/2022/04/Staghorn-Sumac-Spile.pdf as a reference, and created a recipe based on what seemed reasonable. Since I have no real-life experience with maple tapping, I'd be happy to take input on this matter.

#### Testing

Tested adding / removing taps, adding / removing containers.
Tested crafting a makeshift tree spile.
Tested sap production on day 1 of Spring: after a day, 3 units of sap were produced with a makeshift tree spile.

#### Additional context

Tap selection menu:

Taps are visible on the tree:

### DIFF
--- a/data/json/items/generic.json
+++ b/data/json/items/generic.json
@@ -884,7 +884,18 @@
     "price_postapoc": 10,
     "material": [ "steel" ],
     "weight": "11 g",
-    "volume": "50 ml"
+    "volume": "50 ml",
+    "qualities": [ [ "TREE_TAP", 1 ] ]
+  },
+  {
+    "type": "GENERIC",
+    "id": "makeshift_tree_spile",
+    "copy-from": "tree_spile",
+    "color": "brown",
+    "name": { "str": "makeshift tree spile" },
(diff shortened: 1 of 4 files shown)


Maleclypse (approved):
Very cool.

Issue #65632, opened 2023-05-14
Misc maintenance/improvements to innawoods


RenechCDDA:
### Is your feature request related to a problem? Please describe.

This is mostly intended for @Light-Wave although other input is welcome. These are all things I plan to resolve myself, I just need direction on the preferred direction.

I'm going to list and number a series of suggestions and then provide my suggested changes in the next category. These have been written down to myself as notes, so some of them might be a bit rough.

1) Ceramic shards can be foraged from the trash_forest group, is this intended? It was added by @Light-Wave originally. https://github.com/CleverRaven/Cataclysm-DDA/blob/9a4259f56731c94ea315baa785d3804af24661dc/data/mods/innawood/itemgroups/trash_and_debris.json#L16

I believe they **may have been** originally added to allow basic butchery without making tools but ceramic shards no longer work for that purpose (cutting 2 is required).

They're a pretty rare find so if they were intended to be used for butchery I think their role can be replaced by sharp rocks (although the cutting 2 requirement being hardcoded is problematic)

2) Wood sawing 2 quality in innawoods requires steel tools(bow saw/wood saw), which means you cannot make screw press/still/**furniture** without massive infrastructure. 

3) The still also requires duct tape. This makes me very sad because duct tape is hard to acquire in innawoods.

4) Can't make spurge seeds from spurge plants

5) Wooden frames/etc require prying 1 to take apart, which requires.... hammer! STEEL hammer. So if you make a mistake on your dinky little vehicle put together with bronze nails you CANNOT take it apart until you get to the 'iron age'.

5a) The only way to make a steerable wheel mount (for a vehicle) needs wire so.... steel is definitely required

6) Bronze chisel for chiseling 2? used for breaking apart some ores into their component bits (manganese dioxide and zinc). Can be useful when someone is skilling into applied science faster than fabrication

7) Access to the fine cutting quality requires at least a bronze knife. Ok, reasonable, but... the bronze knife requires flatjaw tongs (at the least) to make. Very few of the bronze sandcast mold recipes require this. Why does the *knife* require tongs? Surely it can just be poured, cooled, and grinded down to a good edge afterwards?

Flatjaw tongs *require* multiple lumps of steel, although they can be forged with anvil 1(boulder)/hammering 2(stone hammer). It's not totally impossible, BUT....

7a) Fine cutting is required for making spear shafts. These are required to make proper spears, including stone, copper and brone spears.

[...]

PR #65660, opened 2023-05-15
Add a constructable brick oven

RenechCDDA:
#### Summary
Content "Constructable brick oven"

#### Purpose of change
Add things that exist IRL that can reasonably be built. Work towards #65632

#### Describe the solution
Do a bunch of research, add a new set of constructions, add the terrains those constructions produce. Brick oven terrain provides oven cooking 2, appropriate according to tool_qualities.json: `"oven 2 represents purpose built ovens with basic temperature reglation, including dutch ovens and similar.",`

I'm sure there are many other types of things that would function the same way but this is the one I was imagining.

#### Testing
Loaded it ingame, debug spawned items to build it, built it. Smashed it. Seems to work as I desired.
image
image

#### Additional context
The primary source used to 'design' this was:
 BOILY, Lise, and Jean-François Blanchette. 1979. [The Bread Ovens of Quebec](https://collections.historymuseum.ca/public/objects/common/webmedia.php?irn=5419521). Ottawa: National Museum of Man
It's available online! It can be downloaded individually or as a full PDF. Table of contents: https://www.historymuseum.ca/cmc/exhibitions/tresors/barbeau/mbp0501e.html

Specifically this is a brick oven built with clay mortar and no protecting shed (as we do not model weather effects on terrain). No indication was given that this is any less suitable than any other type of oven, and the use of this type of oven is well-attested to. For the finishing step the door can alternatively be made out of wood or welded sheet metal. There is some suggestion that even large rocks could be used, but these two doors were by far the best supported and reasonable.


data/json/construction.json
```diff
@@ -7867,5 +7867,62 @@
     "components": [ [ [ "rock_large", 22 ] ] ],
     "pre_terrain": "t_dirt",
     "post_terrain": "t_rockyobstacle"
+  },
+  {
+    "type": "construction",
+    "id": "constr_brick_oven_prep",
+    "group": "build_brick_oven",
+    "category": "CONSTRUCT",
+    "required_skills": [ [ "fabrication", 0 ] ],
```
Fris0uman, line 7876: Should probably be at least fab 1

data/json/construction.json
```diff
@@ -7867,5 +7867,62 @@
...
+    "time": "2 h",
+    "tools": [ [ [ "shovel", -1 ], [ "primitive_shovel", -1 ] ] ],
+    "qualities": [ [ { "id": "HAMMER", "level": 2 } ] ],
+    "components": [ [ [ "fire_brick", 30 ] ], [ [ "clay_lump", 5 ] ], [ [ "water", 2 ], [ "water_clean", 2 ] ] ],
```
Fris0uman, line 7880: There's no requirement already made that includes both clean and regular water you could use there instead of listing both? If not it's probably worth it to make that requirement

data/json/construction.json
```diff
@@ -7867,5 +7867,62 @@
...
+  {
+    "type": "construction",
+    "id": "constr_brick_oven_finisher",
+    "group": "build_brick_oven",
+    "category": "CONSTRUCT",
+    "required_skills": [ [ "fabrication", 3 ] ],
+    "time": "3 h",
+    "pre_note": "Carve a thick wooden door for the brick oven.",
+    "qualities": [ [ { "id": "AXE", "level": 2 } ], [ { "id": "CHISEL_WOOD", "level": 1 } ] ],
+    "components": [ [ [ "log", 1 ] ] ],
```
Fris0uman, line 7908: Couldn't you make it out of planks? Maybe it'd be more convenient to have the door be an actual item with a recipe and then have the consrtuction just require the door item.

data/json/furniture_and_terrain/terrain-manufactured.json
```diff
@@ -1484,5 +1484,63 @@
...
+      "items": [ { "item": "fire_brick", "count": [ 45, 75 ] } ]
+    }
+  },
+  {
+    "type": "terrain",
+    "id": "t_brick_oven",
+    "looks_like": "f_fireplace",
+    "name": "brick oven",
+    "symbol": "#",
+    "description": "A large brick oven that could make several pizzas at once.  Smells faintly like pizza.",
```
Fris0uman, line 1532:
```suggestion
    "description": "A large brick oven that could make several pizzas at once.",
```
Ince you can build it yourself I'm not sure the pizza smell makes sense

Fris0uman (commented):
How does it consume fuel? Is that handled automatically by the OVEN tool quality?

RenechCDDA:
> How does it consume fuel? Is that handled automatically by the OVEN tool quality?

Baked into(no pun intended) whatever recipe you're making. At the moment we don't represent scaling requirements for cooking based on batch size, even though it makes sense. A larger oven like this is definitely the most egregious case, as they're typically fired lightly for several hours before actual cooking begins... but there's no ramp-up or waiting time for the oven to get at temperature. It's just abstracted out into the recipe's cost.

I'll try to address the rest of your review sometime today when I get time.

data/json/construction.json
RenechCDDA: Yeah good catch, making it fab 2. Nothing fancy but still some opportunities to screw up.

data/json/furniture_and_terrain/terrain-manufactured.json
RenechCDDA: Fair enough. These sorts of large ovens can get some "seasoning" which leaves a residual smell (so I've been told) by means of fats/oils which melt and form a layer coating, much like the seasoning of cast iron cookware. But the player hasn't necessarily cooked anything which ends up smelling like pizza specifically.

data/json/construction.json
RenechCDDA: >Couldn't you make it out of planks?

That's a distinct 'maybe'. The book I used as reference does not elaborate on the construction of wooden doors. It does have a singular image of one (ch.1 pg 18), which appears to be several joined planks with a crossbeam and crude handle.

I think it could be added, I have nothing against it, I just don't feel the need/effort for a third option. I'll leave this conversation unresolved so you/the other mergers can decide if that should be added.

>Maybe it'd be more convenient to have the door be an actual item with a recipe and then have the consrtuction just require the door item.

Personally I do not like these very "bespoke" items with single purposes. 

Implementing it that way would mean 2 more entries in the crafting menu, and it'd still need an entry in the construction menu to change the terrain! More json, more definitions, little to gain IMO.

PR #65667, opened 2023-05-16
Altered lump of steel recipe to have time requirement comparable to similar recipes.

ajkrupka:
#### Summary
Balance change to bring lump of steel deconstruction in line with other parts

#### Purpose of change
Lumps of steel take 1 hour to deconstruct unlike similar products (metal frame -> lump, or deconstructing chunks) which take 15 minutes. This fixes lumps to deconstruct in 15m making deconstruct times for the entire product chain similar.

#### Describe the solution
Changed the timing in ./data/json/recipes/other/materials.json.
We're cutting a big piece of metal into somewhat smaller pieces, not doing surgery. This is a rapid deconstruction not precision work.

#### Testing
Tested by editing files in local updated client.

### DIFF
--- a/data/json/recipes/other/materials.json
+++ b/data/json/recipes/other/materials.json
@@ -859,7 +859,7 @@
     "subcategory": "CSC_*_NESTED",
     "skill_used": "fabrication",
     "difficulty": 2,
-    "time": "60 m",
+    "time": "90 m",
     "batch_time_factors": [ 90, 4 ],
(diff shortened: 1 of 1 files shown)


ajkrupka:
Uploaded the wrong file like a dummy.

Miner239:
You can just commit the correct file to overwrite the change... no need to remake the entire PR

PR #65668, opened 2023-05-16
Adjusted steel_lump deconstruct time.

ajkrupka:
#### Summary
Balance "adjustment to steel lump deconstruction."

#### Purpose of change
Lumps of steel deconstruct in 60 minutes. This is four times as much as chunks of steel. It is likely this is left over from a previous refactor. Deconstruction in this case is a rapid, coarse process designed to quickly break up a big piece into four small pieces, 60 minutes seems like overkill for what is quickly smashing up some parts.

#### Describe the solution
Added an entry to recipe_deconstruction.json

#### Testing
Tested on current build, fix works as intended.

### DIFF
--- a/data/json/recipes/recipe_deconstruction.json
+++ b/data/json/recipes/recipe_deconstruction.json
@@ -7595,5 +7595,13 @@
     "time": "2 m",
     "qualities": [ { "id": "FABRIC_CUT", "level": 1 } ],
     "components": [ [ [ "sheet_canvas", 3 ] ], [ [ "thread_canvas", 40 ] ] ]
+  },
+  {
+    "result": "steel_lump",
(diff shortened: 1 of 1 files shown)


Karol1223:
You're quite literally not doing anything at all you're claiming you're doing.

You're adjusting the recipe to CRAFT the lump, not to disassemble it.

Disassembling steel frames also doesn't take 15 minutes, it's more along the lines of several hours.

ajkrupka:
I understand; I'll make an un-crafting recipe instead. Thanks for the review.

ajkrupka:
I reverted the previous changes and updated the PR to include a new recipe in a different file. I hope this is more appropriate.

Karol1223 (approved):

PR #65675, opened 2023-05-16
vehicle_part cleanup

irwiss:
#### Summary
None

#### Purpose of change

Clean up some vehicle_part related things

#### Describe the solution

A bunch of changes that should clean up various areas related to vehicle_part; prunes constructor or function arguments that can be setter calls, and (hopefully) prevent some invalid vehicle configurations by making all install_part overloads check can_mount.

https://github.com/CleverRaven/Cataclysm-DDA/pull/65675/commits/40f01e46ef2456de526c56269d9104f2aa736d39 Removes operator bool()+default constructed hack and instead uses std::optional for vehicle tank selection function.

https://github.com/CleverRaven/Cataclysm-DDA/pull/65675/commits/15fa4bba16a82b527a60b73364361b2fd58f0daf Prunes vehicle_part constructors and arguments - leaving just the default one and the one taking type id and base item, the other 2 arguments aren't needed to construct a valid vehicle part and most places don't use them. The mount point argument isn't relevant until the part is actually placed by `vehicle::install_part` (which sets it), variant field is both valid when empty and can be set as a normal field at any time, instead of being a constructor argument.

Also removes `id` field making `vehicle_part::info_` field the single source of truth for vehicle part type, this allows replacing caching code from `vehicle_part::info()` with just a pointer dereference.

https://github.com/CleverRaven/Cataclysm-DDA/pull/65675/commits/b60d5ff164669a74671793b5cdd4f733a0ae7370 Prunes parameters from `vehicle::install_part` overloads; variant is a normal field that is both valid empty and can be set at any time and `force` argument that skips a check for what looks to be exclusively to allow power cable installations (more on this later).

https://github.com/CleverRaven/Cataclysm-DDA/pull/65675/commits/5c47680f245fc8bb90e2f74a74334110503c0caa - Makes all install_part overloads check can_mount for valid part mounting, and raise a debugmsg if the check fails (this commit will fail tests).

https://github.com/CleverRaven/Cataclysm-DDA/pull/65675/commits/6f935da37003cc93cace8f4d9f954bdc0d1a6b61 - Refactors \*INSTALL\* flags; as the previously mentioned `force` check was used to skip a failing can_mount case:

Power cords (/jumper cables) actually violate the "rules" of vehicles, they're set to NOINSTALL so can_mount check fails, but then circumvented this via the overload that didn't perform the check, this also violates the check of multiple parts of same type can't be mounted on same tile, (next is speculation:) to get around this the `force` argument was introduced to install_part to skip the check.

This commit cleans up this area;
* `NOINSTALL` flag is removed
* `NO_INSTALL_HIDDEN` will hide the part from install ui.
* `NO_INSTALL_PLAYER` will show in the install ui but won't let the player actually install the part.
Neither flag will block install_part or make can_mount fail and only prevent the UI from showing or allowing to install the part, this allows install_part to install jumper cables from code and still have a can_mount check

POWER_TRANSFER parts specifically are now allowed to stack same ids on same tile in `vehicle::can_mount` checks, while this isn't strictly correct (pretty sure i saw code somewhere that assumes unique ids) this is how it currently works.

Obsolete/migrated, special (mana frames, inflatable boat parts), all kinds of POWER_TRANSFER cables are set to `NO_INSTALL_HIDDEN`.
Helicopter rotors are currently the only parts which have `NO_INSTALL_PLAYER`.

`vehicle::can_mount` is also refactored to cut down on the number of invididual loops and various visual fluff for better ux (example in screenshot).

(This commit will still fail tests - as the wreck merging test still generates invalid vehicles, fixed in next commit)

https://github.com/CleverRaven/Cataclysm-DDA/pull/65675/commits/3d108a5c0f2b874735b99d0d32d0b01293780ef6 Blocks one of the cases where wrecks generate invalid vehicles; parts that would cause a conflict are dropped (quietly, only writing to debug.log), hopefully this will stop wrecks generating double-frames, double-seats and similar weird configurations which sometimes trigger errors when critters hop on them.

At this commit the tests pass

https://github.com/CleverRaven/Cataclysm-DDA/pull/65675/commits/d0305f2bcdf2df9fe3ce0fd5891ffa76eff986c6 Removes old migrations, one migration is installing a turret_mount for any turret that's loaded which causes an error for already-migrated vehicles since install_part raises debug_msg on failure now, but all of the removed migrations should be obsolete anyway.

At this commit existing saves should load without errors (as long as the part setup is valid and not debugmenu'd).

#### Testing

Tests should catch most of the changes, playtested the changes a bit

#### Additional context

Likely the only hotspot for install_part is vehicles::finalize_prototypes; but testing with afs, blazemod, magiclysm and xedra,
the time spent in it is almost the same as without this patch, or at least random jitter overwhelmed the difference, in total the finalize method runs in under 200ms or so.

TODO:
May be fix more weird stuff, like parts with "on_roof" location not actually requiring a roof, but that's a change for another pr, seems there's a ton of those; solar panels on various car prototypes, floodlights, turret mounts, emergency lights, security cameras on tractors etc...

image

### DIFF
--- a/data/json/furniture_and_terrain/appliances.json
+++ b/data/json/furniture_and_terrain/appliances.json
@@ -596,7 +596,7 @@
     "durability": 120,
     "description": "A power cord sticking out of an appliance.  You need to plug it into a powered grid for the appliance to work properly.",
     "item": "power_cord",
-    "flags": [ "NOINSTALL", "NO_UNINSTALL", "UNMOUNT_ON_DAMAGE", "UNMOUNT_ON_MOVE", "POWER_TRANSFER" ]
+    "flags": [ "NO_INSTALL_HIDDEN", "NO_UNINSTALL", "UNMOUNT_ON_DAMAGE", "UNMOUNT_ON_MOVE", "POWER_TRANSFER" ]
   },
   {
     "id": "ap_standing_lamp",
--- a/data/json/vehicleparts/vehicle_parts.json
+++ b/data/json/vehicleparts/vehicle_parts.json
@@ -84,7 +84,7 @@
     "broken_color": "red",
     "durability": 100,
     "item": "null",
-    "flags": [ "NOINSTALL" ],
+    "flags": [ "NO_INSTALL_HIDDEN" ],
     "breaks_into": [  ]
   },
   {
@@ -1158,7 +1158,7 @@
       "removal": { "skills": [ [ "mechanics", 2 ] ], "time": "30 m", "using": [ [ "vehicle_screw", 1 ] ] },
       "repair": { "skills": [ [ "mechanics", 4 ] ], "time": "60 m", "using": [ [ "adhesive", 2 ] ] }
     },
-    "flags": [ "MOUNTABLE", "BOARDABLE", "CARGO", "NO_INSTALL_PLAYER", "NO_UNINSTALL" ],
+    "flags": [ "MOUNTABLE", "BOARDABLE", "CARGO", "NO_INSTALL_HIDDEN", "NO_UNINSTALL" ],
     "breaks_into": [ { "item": "plastic_chunk", "count": [ 10, 20 ] } ],
     "damage_reduction": { "bash": 10 }
   },
@@ -1184,7 +1184,7 @@
       "removal": { "skills": [ [ "mechanics", 2 ] ], "time": "30 m", "using": [ [ "vehicle_screw", 1 ] ] },
       "repair": { "skills": [ [ "mechanics", 3 ] ], "time": "60 m", "using": [ [ "adhesive_rubber", 2 ], [ "tire_rubber", 2 ] ] }
     },
-    "flags": [ "FLOATS", "VARIABLE_SIZE", "NO_INSTALL_PLAYER", "NO_UNINSTALL" ],
+    "flags": [ "FLOATS", "VARIABLE_SIZE", "NO_INSTALL_HIDDEN", "NO_UNINSTALL" ],
     "breaks_into": [ { "item": "plastic_chunk", "count": [ 10, 20 ] } ],
     "damage_reduction": { "bash": 10 }
   },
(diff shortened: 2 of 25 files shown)


RenechCDDA:
>Power cords (/jumper cables) actually violate the "rules" of vehicles, they're set to NOINSTALL so can_mount check fails, but then circumvented this via the overload that didn't perform the check

Huh, so that explains some things.

>NO_INSTALL_PLAYER will show in the install ui but won't let the player actually install the part.

Epic, thank you! I hope my minor changes didn't cause you too much grief.

>vehicle::can_mount is also refactored to cut down on the number of invididual loops and various visual fluff for better ux (example in screenshot).

Another excellent change, especially listing out the parts which satisfy flag requirements.

irwiss:
> Epic, thank you! I hope my minor changes didn't cause you too much grief.

No I mostly took your changes as is only changing a bit, the diff looks bigger than it is cause I moved the stuff so it fits into one loop

PatrikLundell:
I may misunderstand things, but my understanding was that the part id and the tile is the tuple identifying a part (stored in an array at some index, but the index is mutable as parts are installed/removed before it in the array, and so the index can only be used very locally (but isn't, leading to occasional crashes when multiple companions dismantle vehicles, unless that's fixed)). This likely means that if you select a part that has multiple instances installed on a tile for removal, you're probably going to remove the first one encountered by the iteration, regardless of which one you selected (so regardless of which power cable you select, you're going to remove the one going to the same vehicle, and never the one going to the other vehicle).

irwiss:
Part id is still there, only removed as a separate field because id and info_cache.id were supposed to be the same, this patch removes the caching dance in info() function and makes info_ field a "read mostly" pointer (not const or const ref due to reasons listed in comment near it), performing the string_id->vpart_info lookup in constructor instead of re-checking it in every info() call

The assumption that [part id string,part mount] tuple is unique exists in (likely ineffective) hacks for npcs dismantling vehicles, this is kinda wrong even without the power cords violating this rule (since part mounts can shift), it's also unchanged by this patch

PR #65678, opened 2023-05-16
Correct weight of shark plushie

Inglonias:
#### Summary
Bugfixes "Corrects weight of shark plushie"

#### Purpose of change
The in-game shark plushie is, as far as I'm aware, modeled after the [Ikea BLÅHAJ](https://www.ikea.com/us/en/p/blahaj-soft-toy-shark-90373590/). According to the page linked, the packaging section of the site claims the BLÅHAJ weighs 1 lb, 6 oz, or ~625 grams.

#### Describe the solution
Change shark plushie's weight to be correct.

This is so trivial I may as well not bother. If it causes density tests to fail, I may just close this without a merge.

#### Testing
Automated testing should catch any failures. Only thing I can think of that could possibly fail is the density test.

### DIFF
--- a/data/json/items/toy.json
+++ b/data/json/items/toy.json
@@ -72,7 +72,7 @@
     "description": "A somewhat large, mass produced shark plushie.  Meant to provide comfort to people of all ages.",
     "category": "other",
     "material": [ "cotton" ],
-    "weight": "1945 g",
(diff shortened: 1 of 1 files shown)


Inglonias:
It looks like the density check only checks for maximum density, which is good for me.

PR #65681, opened 2023-05-16
fix exploits regarding the hm12

anoobindisguise:
#### Summary
Balance "HM12 is less exploitable"

#### Purpose of change
There were several unintentional exploits with the HM12 dazzle rifle. 
- It cost 1 kJ of charge making it the ideal weapon to train your rifles and marksman skill despite lacking most of the critical features of a gun that would need skill to handle.
- It could be modded with laser weapon mods. The High Density Capacitor adds 10 heat damage and 5 range at cost of 2x power consumption... for this weapon that means +1000% damage at +100% energy cost turning it into an exceedingly viable combat tool. There were also similar issues with Beam Scatterer and Focusing Lens (+10 damage and less range or +5 damage and more range). It also inherited automatic fire capability from its parent a7 laser rifle which gave it further combat capabilities despite it having no reason to do automatic fire in lore.

#### Describe the solution
- Semi auto only.
- Bump kJ cost from 1 to 20.
- Cannot be modded with emitter or lens.

Give it the WONT_TRAIN_MARKSMANSHIP flag instead of just powering up its cost. However it was still probably too viable for plinking things to death slowly 1 damage at a time as its battery consumption meant it had 1000-2500 damage per UPS which was still very useful and not its intended purpose (being an ideal varmint hunting tool for example)

#### Additional context
@John-Candlebury


John-Candlebury:
I think giving it the WONT_TRAIN_MARKSMANSHIP flag is a better solution overall. I'm concerned that removing burst fire might make it very difficult to use for its intended purpose.

Damage wise, it seems reasonable to make this gun deal a new "emp" damage type that only robotic enemies are vulnerable to. I think it should be possible now that damage types are data driven-

anoobindisguise:
> I think giving it the...

FWIW I've literally never used its auto mode for stunning the yrax. The auto mode was only useful for dealing damage it enemies in my experience. But if it no longer does 1 pure damage it wouldn't actually matter, I'd just need to figure out how new damagetypes work.

John-Candlebury:
> > I think giving it the...
> 
> FWIW I've literally never used its auto mode for stunning the yrax. The auto mode was only useful for dealing damage it enemies in my experience. But if it no longer does 1 pure damage it wouldn't actually matter, I'd just need to figure out how new damagetypes work.

Scratch the damage part thats actually really complex.

PR #65682, opened 2023-05-16
give nailgun WONT_TRAIN_MARKSMANSHIP

anoobindisguise:
#### Summary
Balance "nailgun doesn't train you to shoot a gun"

#### Purpose of change
Nailguns shouldn't teach you how to use a bullet gun.

#### Describe the solution
Nailguns don't teach you how to use a gun.

### DIFF
--- a/data/json/items/gun/nail.json
+++ b/data/json/items/gun/nail.json
@@ -24,7 +24,8 @@
     "reload": 50,
     "valid_mod_locations": [ [ "grip mount", 1 ], [ "rail mount", 1 ], [ "sights mount", 1 ], [ "stock mount", 1 ], [ "underbarrel mount", 1 ] ],
     "pocket_data": [ { "pocket_type": "MAGAZINE", "rigid": true, "ammo_restriction": { "nail": 20 } } ],
-    "melee_damage": { "bash": 6 }
+    "melee_damage": { "bash": 6 },
+    "flags": [ "WONT_TRAIN_MARKSMANSHIP" ]
   },
   {
     "id": "coilgun",


Maleclypse (approved):

Havok91:
I don't understand how it wouldn't train marksmanship. If you can hit a target with a weapon 100% not designed to be used in this manner, hell it's not even equipped with sights, then you should have a far easier time shooting a gun with the proper equipment for aiming at your target, IE: sights. As far as I am aware, marksmanship is fairly generic capability to hit a given target with a projectile based weapon, archery trains marksmanship (IIRC) and by your logic, bows shouldn't teach you how to aim guns better is a logical conclusion, but in reality, this doesn't make sense considering the much more complicated aiming process involved in archery. A similar situation occurs with the nail gun, it shouldn't raise handguns, even though that's the weapon it is most similar to, but marksmanship? It most certainly should. As of right now, with a lack of "Improvised Weapon" skill, using a nail gun for any period of time now nets you nothing skill wise, which takes away from the fun factor of using a pretty everyday tool as a weapon in a cataclysmic apocalypse.

anoobindisguise:
> I don't understand how it wouldn't train marksmanship...

You can give it more nuance if you want (like a "ONLY_TRAINS_MARKSMANSHIP_A_LITTLE") flag but as is it's a very exploitable and brainless way to become a master of firearms in a way that makes no sense (why doesn't the military use nailguns to train its recruits if it's so good at teaching you due? nails are cheaper than bullets.)

Havok91:
> You can give it more nuance if you want (like a "ONLY_TRAINS_MARKSMANSHIP_A_LITTLE") flag

Why should someone else have to go in and make a sensible change? There is a certain suspension of disbelief that is generally considered acceptable in video games, adding so much realism as to take away any incentive or reason whatsoever to use an item seems to starkly contrast that concept. As I stated in my first comment, there is no net gain or progression in any fashion by using the current iteration of an item that was purposely added as an improvised weapon. 

> why doesn't the military use nailguns to train its recruits if it's so good at teaching you due?

This is such an ignorant and gaslight-y question, thankfully I was in the military so I feel like I can actually answer this question with something more than "That's stupid" which would elicit a response along the lines of "That's why I did it" 

The military (U.S. at least) doesn't use improvised weapons for marksmanship training simply due to the fact that they don't have to improvise equipment for their soldiers. The military isn't currently in a post-apocalyptic survival scenario that would warrant learning to shoot something that isn't a service issued weapon. The U.S. military currently has a defense budget that is larger than some first-world countries' entire Gross Domestic Product, why the hell would they use bloody nail guns when they have hundreds of thousands of M9s (prior service pistol) and likely hundreds of thousands more of the new Sig Sauer service pistol? 

TLDR:
1. Don't gaslight me over a simple question that was anchored in game mechanics.
2. Based on what I've read in regards to the marksmanship skill, your commit is quite heavy handed.
3. I was trying to be nice about asking for either further explanation, or maybe requesting an edit to the commit that would be a little kinder and more game friendly, seeing as this is a game and all.
4. I don't know the code, nor do I actually know how to make a submission or request one, which is why I'm here contributing via the conversation feature rather than out writing code. I know weapons, physics and some other tidbits, not code though.

Qrox:
I feel this could use the proficiency system so using a nailgun trains you in nailgun and maybe a litte with a cap in handgun etc.

PR #65683, opened 2023-05-16
Tank top cleanup

Karol1223:
#### Summary
None

#### Purpose of change
Well, it annoyed me that the camo tops didn't have a disassembly since they weren't variants... and then it kind of got out of hand...

#### Describe the solution
- For starters, cheerleader tops and camo tops are now variants of the tank top item. This means they copy its values and can now be disassembled, I also made them be able to spawn in place of normal tank tops - camo has a 6% chance, cheerleader has a 4% chance
- Weight and volume of the tank top have been changed from 78g to 130g and from 250ml to 115ml
- Preapo price has been increased to 9$ from 5$
- Crafting recipe now has byproduct cotton scrap
- Disassembly no longer handled via reversible crafting to avoid mass generation. Will now give one less sheet and 10 less thread


data/json/professions.json
```diff
@@ -5381,7 +5354,10 @@
     "points": 2,
     "skills": [ { "level": 2, "name": "dodge" }, { "level": 3, "name": "speech" }, { "level": 3, "name": "swimming" } ],
     "items": {
-      "both": { "items": [ "pom_poms", "cheerleader_top", "socks_ankle", "sneakers" ] },
+      "both": {
+        "items": [ "pom_poms", "cheerleader_top", "socks_ankle", "sneakers" ],
```
Termineitor244, line 5358:
```suggestion
        "items": [ "pom_poms", "socks_ankle", "sneakers" ],
```

Termineitor244:
Thank you for taking the time to do this!

data/json/professions.json
Karol1223: Oops! Good catch, I had a hunch I forgot something...

data/json/uncraft/armor/torso.json
```diff
@@ -0,0 +1,10 @@
+[
+  {
+    "result": "pants_leather",
```
bombasticSlacks, line 3: I'm guessing this is a typo

Karol1223: Ah, yep, copied it over and forgot to change id. This is what I get for making PRs in the middle of the night and not testing the changes thinking it's too simple to fall apart. Give me a second.

bombasticSlacks: This is also a great example of why focused PRs have utility. If this PR touched a bunch of items, it would take me 10-20 times as long to notice an error like this.

Karol1223: Yup, that's why I actually applied your suggestion to make more smaller PRs instead of one massive one. This one's just one item because it was a lot of swapping around IDs...

PR #65685, opened 2023-05-16
Disarming telepad is trivial when anchored

anoobindisguise:
#### Summary
Balance "anchoring makes disarming telepad trivial"

#### Purpose of change
Teleport pads were obscenely difficult to disarm regardless of whether you had a dimensional anchor or not which made no sense, and it also turned them into the ideal skill-grinding method for Devices - simply attempt to disarm (and usually fail) 5000+ times while anchored and bam, 10 devices in the span of like 15 in game minutes (since failures train anyways but don't take time)

#### Describe the solution
Use the trivial take-down code for the telepad if you meet the conditions (the trap is a telepad and you are anchored) this makes it award no XP and always get disarmed.

You can still place the trap down for xp/profs which might make this pointless. I ought to remove that too since as is it means just more keypresses for the same result but there's fundamentally nothing that different about training trapping with some other thing and a telepad so it introduces some unpleasant inconsistencies. This could be done in iuse_actor though.

#### Testing
Tested and confirmed I could disarm without fail or gaining any knowledge while anchored but still floundered and got teleported every time when not. Unfortunately placing it still awards free XP, but... baby steps. At least placing traps takes time unlike disarming.
image


kilozombie:
Why are they so impossible to disarm in the game? It's completely anachronistic to the faux-realism of the labs elsewhere and impossible to imagine how the player would keep messing up in such a way that they get teleported. Surely you would just try to take the screws out or something.

Teleporters are very helpful for Magiclysm and telepads are the only reasonable source, but even after grinding to obscenely high Devices and training all my proficiencies related to trapping, they are impossible to take down. It's nonsense.

kilozombie:
I also want to bring up another point you had, which is totally correct:

>You can still place the trap down for xp/profs which might make this pointless. I ought to remove that too since as is it means just more keypresses for the same result but there's fundamentally nothing that different about training trapping with some other thing and a telepad so it introduces some unpleasant inconsistencies.

The solution _certainly_ does not have to be done via this specific PR, but trapping as a skill is in a really sorry state. Or, whatever, "devices". It's painful on the hands to train and unrewarding to have trained. Perfect candidate for more practice recipes and/or more uses of the skill, and this is just one example where the system breaks down.

Issue #57882, opened 2022-05-24
Debug "quit without saving" claims to disable achievements, but does not


Inglonias:
### Describe the bug

In the debug menu, the "Quit to Main Menu" option in the game menu claims that it will disable achievements. It doesn't. Probably because achievements being disabled isn't saved.

### Steps to reproduce

1. Start a game
2. Save the game
3. Use the debug "Quit to main menu" option.
4. Observe that it claims achievements will be disabled.
5. Do it anyway.
6. Reload the game
7. Achievements are still enabled

### Expected behavior

Either don't tell me that the option will disable achievements, or have it actually do so. The game lying to me is ever so slightly annoying.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.1706 (21H2)
- Game Version: 9dbad78 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    No Rail Stations [No_Rail_Stations],
    No Vault [No_Vault],
    ASCIITiles overmap color fix [colors_fix_asciitiles],
    Fewer Portal Storms [Fewer_Portal_Storms]
]

### Additional context

This is a low-priority issue, but I figured I would log it.

NetSysFire:
Can confirm. There are two issues here:

- Quit to main menu should not even be considered cheating and disable achievements.
- Quit to main menu also asks to quit without saving, which is why achievements do not get disabled: it is not written to disk and remains in-memory for this tiny amount of time between achievements being disabled and quitting without saving.

As you already said so yourself, this is low priority which is why I labelled as such. It will not break the game. The actual thing that needs to be fixed is the first thing I mentioned and might be an easy fix, maybe even a good first one.

PR #65686, opened 2023-05-17
Make Quickload and Quit To Main Menu not show "will disable achievements" warning

Kamayana:
#### Summary
Interface "Make Quickload and Quit To Main Menu not show "will disable achievements" warning"

#### Purpose of change
Quickload and Quit To Main Menu show the warning that using them will disable achievements, but this makes no sense since even if they _were_ cheats, they'd both reset the game to a point before they were used. It just adds an unnecessary popup to close through when testing people's saves.

#### Describe the solution
Adds them to the non_cheaty_options list.

#### Testing
I tested it, it works.


Inglonias:
This should close #57882 when merged.

PR #65690, opened 2023-05-17
Nerf Beehives

### LINKED ISSUE #65636: Beehives are completely unbalanced!

Beehives are currently found everywhere in the Forest. They are very close togheter and do not even need to be harvested. Just by smashing them you get around 10 honey combs for 1 Beehive and in the summer, only 1 is harvestable for around 170 honey combs! Honey combs have a very high calorie value and do not spoil. 
Beehives as they are now, make ANY OTHER Food source obsolete!

### Solution you would like.

The simplest solution would be reducing the amount of Beehives that spawn in the forest very significantly, making them rare.
Beehive are not this common in reality either, which should make this solution reasonable. 
From Venera3;
For reference, wild honeybee hives are about one per square kilometre in temperate NA - see https://resjournals.onlinelibrary.wiley.com/doi/10.1111/een.12715, supplementary table one. That comes out to about four per overmap.

_No response_

### Additional context

_No response_

worm-girl:
#### Summary
Balance "Nerf Beehives"

#### Purpose of change

Beehives were spawning everywhere and could be smashed for massive amounts of free honey year-round. 

closes #65636 

#### Describe the solution

Beehives were removed from the terrain spawn in mapgen and changed to a map special, which has a small chance of placing a hive in a forest and a slightly larger chance of placing a hive in a thick forest.

Smashing the hives was possible year-round, and was yielding a lot of honey, even though the hives aren't harvestable year-round. Additionally, wax and honey production were well above what google says you actually get from a beehive.

Wax was reduced to produce around a pound from a single harvest, and about half that if you smash the thing. The minimum honey harvest was reduced to 10 pounds, the maximum was left alone. If you smash the hive like a dumb beast instead of harvesting it, you will get some honeycomb year round (I don't think we have seasonal smashing) but most of it gets wasted.

This process needs several extra steps. While you can eat comb straight out of the hive, getting the full harvest irl requires a press, and getting clean, usable wax requires extensive filtering and rendering. There's currently no danger from the bees themselves, and the player should also be able to get edible brood and royal jelly.

This also sort of highlights one of the problems with our food system. We don't track macronutrients, so a person can be totally healthy just eating honey 24/7 and nothing else.

#### Testing

image

Wandered around in the woods for a really long time until I found a hive. Per @Venera3 they should be like 4 to an overmap tile. I didn't canvas an entire OMT (we don't have noclip so navigating forests is a pain) but they do seem to be in that ballpark.

Harvested and smashed some different hives at different skill levels, saw that the honey harvest were within expected bounds.


GuardianDll:
Fyi you can simply use autowalk (not autodrive) to navigate through the forest

Inglonias:
> Fyi you can simply use autowalk (not autodrive) to navigate through the forest

Even with autowalk, it would be a huge pain in the ass to canvas an entire OMT. This seems like something more suited for automated testing. That said, I don't think an automated test is worth the effort here.

Issue #65378, opened 2023-04-29
Change description of ceramic armor plate


RedBloodCat:
### Is your feature request related to a problem? Please describe.

The ceramic armor plate you can get is not an armor that the player can use, however the description makes it sound like it could be used as an armor plate. This is quite confusing and should be changed.

### Solution you would like.

Changed to something along the lines of "A ceramic armor plate, specifically in for use as armor. Not suitable for humans”

### Describe alternatives you have considered.

Na

PR #65691, opened 2023-05-17
Changes description of ceramic armor plate

WoodPollution:
#### Summary
None

#### Purpose of change
Makes its so that it's understood that ceramic armor plate is not for use as armor plates in vest.
Does my suggestion #65378

#### Describe the solution
Makes it clear that its not a ceramic plate for use in human body armor, to avoid confusion.

Do nothing

#### Testing
Small json tweak. Loaded the game up and it worked fine

### DIFF
--- a/data/json/items/generic.json
+++ b/data/json/items/generic.json
@@ -2062,7 +2062,7 @@
     "symbol": ",",
     "color": "yellow",
     "name": { "str": "ceramic armor plate" },
-    "description": "A ceramic armor plate, specifically engineered for use in a bullet-resistant vest.",
+    "description": "A ceramic armor plate, specifically engineered for use as armor. Not suitable for human use.",
     "price": 8000,
     "price_postapoc": 100,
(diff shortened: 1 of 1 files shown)


data/json/items/generic.json
```diff
@@ -2062,7 +2062,7 @@
     "symbol": ",",
     "color": "yellow",
     "name": { "str": "ceramic armor plate" },
-    "description": "A ceramic armor plate, specifically engineered for use in a bullet-resistant vest.",
+    "description": "A ceramic armor plate, specifically engineered for use as armor. Not suitable for human use.",
```
bombasticSlacks, line 2065:
```suggestion
    "description": "A ceramic armor plate, specifically engineered for use as armor.  Not suitable for human use.",
```

Daved27hundred:
If it weren't for the fact that mil. composite plating breaks into these (among other reasons) I'd just say obsolete it entirely.

Issue #65643, opened 2023-05-14
Segmentation Fault CTD when repairing using UPS-equipped tools


Spaceb0i:
### Describe the bug

Any time you repair something with a tool with the UPS conversion mod, it causing a crash to desktop and segmentation fault. This was tested both with a makeshift arc welder and a soldering iron, the former on a pipe and 1L engine, and the latter on a funnel. All of them caused the same CTD. 

### Attach save file

[Massachusetts-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11473103/Massachusetts-trimmed.tar.gz)
7zip claims "there are some data after the end of the payload data", not sure what this means but if that's some kind of save corruption it could be a problem (I doubt it, considering the rest of the game works fine).

### Steps to reproduce

1. Activate makeshift arc welder+1.
2. Attempt to repair pipe.
3. Observe a crash to desktop.

### Expected behavior

I would expect the item to be repaired and no CTD to happen.

### Screenshots

![image](https://github.com/CleverRaven/Cataclysm-DDA/assets/77634653/ced9441c-87a7-48af-906d-28bb419f1102)
Screenshot of repair screen taken immediately before crash.

### Versions and configuration

 OS: Windows
    - OS Version: 10.0.22621.1702 (22H2)
- Game Version: 9a4259f [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    sees-player, run, crouch, bleed icons, retrodays [sees_player_retro],
    Massachusetts [MA],
    Stats Through Kills [stats_through_kills]
]

### Additional context

[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/11473104/crash.log)
[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/11473105/debug.log)
I was initially on an older experimental build (version ea17d8e) which had the same issues.

Spaceb0i:
Source of the bug found. In line 2333 of activity_handlers.cpp there is the following line:
ammo_name = item::nname( used_tool->ammo_current() );
It gives the following reason for why there was an unhandled exception:
Item_factory::find_template(...)->ammo._Mypair._Myval2 was nullptr
[...]

PR #65693, opened 2023-05-17
CTD caused by repair attempts with UPS-equipped tools fixed

Spaceb0i:
#### Summary
Bugfixes "CTD caused by repair attempts with UPS-equipped tools fixed"

#### Purpose of change

This fixes a crash to desktop (null pointer error) described in Issue #65643. Simply put, attempting a repair using a UPS-equipped tool would cause a CTD, which is bad.

#### Describe the solution

repair_item_finish now checks if an item uses UPS and sets ammo type and name based on that, and inv_use_ups no longers tries to set ammo on UPS-equipped tools without their own battery.

A total rewrite of UPS function is not in my skillset (this is my first time using C++) and I'm not even sure it would do away with the need for all the manual checks everywhere.

#### Testing

Repairs were attempted with a UPS-equipped makeshift arc welder, a UPS-equipped soldering iron, and a normal soldering iron, all of which worked correctly with no errors. An empty UPS and a UPS with a dead battery, as expected, prevented the UPS tools from repairing. Additionally, a smartphone was used to test if inv_use_ups still worked, and it did, recharging the phone battery any time it dropped.

#### Additional context

Some old savegame conversion code meant to be deprecated after 0.F got removed, as it was related to the problem. I'm not sure how to get rid of the redundant commits that cancel each other out.

### DIFF
--- a/src/map_field.cpp
+++ b/src/map_field.cpp
@@ -1882,16 +1882,10 @@ void map::monster_in_field( monster &z )
             if( z.made_of_any( Creature::cmat_fleshnveg ) && !z.has_flag( MF_NO_BREATHE ) ) {
                 if( cur.get_field_intensity() == 3 ) {
                     z.add_effect( effect_stunned, rng( 1_minutes, 2_minutes ) );
-                    dam += rng( 4, 10 );
                 } else if( cur.get_field_intensity() == 2 ) {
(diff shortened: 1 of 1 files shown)


anoobindisguise:
This is probably what was breaking the test in #64868 I was struggling to figure that one out lol. Thanks for fixing!

PR #65700, opened 2023-05-17
START-I

Drew4484:
#### Summary
Content "Restrict mininuke availability"

#### Purpose of change
Mininukes fall under the same problem that the ASRGs do/did, they're far too common. Additionally they're one of the least realistic items in the game, and even if they were accurate, it's far too easy for a new player to kill, be killed, or both by one.

#### Describe the solution
Removes spawns of mininukes and mininuke hacks. Leaves recipes intact and does not remove the actual item from JSON.

Removing all pseudo-atomic devices.

#### Testing
WIP


anoobindisguise:
Now I can finally stop being so paranoid that a dispatch in a military outpost will send in a mininuke hack at the zombie soldiers and kill me from outside the structure. Don't forget about the hardcoded spawn of the damaged mininuke in old labs.

Issue #61690, opened 2022-10-14
Extreme basecamp woodcutting mission extention


PatrikLundell:
### Describe the bug

I send off my companion on a log cutting mission, with a time of about a week or slightly more than that (and just above the amount of stored calories I had, after having extended the store to be sufficient). Now, a day or two later, the expected remaining time is 10 weeks and 4 days.

### Steps to reproduce

1. Send off a companion on a wood cutting basecamp mission.
2. Repeat until the remaining time when checked later is extreme.

### Expected behavior

Given that the mission calculates the time needed when the companion is sent off, the mission should take that time. I don't even think there's any uncertainty in the calculation.

### Screenshots

![Screenshot (266)](https://user-images.githubusercontent.com/22739822/195894144-2b2aa2b6-9161-4b10-bc62-e6ccbc6a5842.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19043.2130 (21H1)
- Game Version: 64e17c6 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Innawood [innawood]
]
+ "Mod" to disable portal storms
+ Addition of bare bones basecamp recipes for lump of steel and plant fiber.

### Additional context

The only unusual thing I'm aware of is that I've visited the target area after sending of the companion and have foraged some plants on that tile (made visible to me, the player, thanks to the removal of many the trees). It may be the first time I've actually visited a logging tile while it's being logged.

I've also observed a cougar chasing a coromant across the area where I normally move to get to that part of the map (now taking a detour in order to avoid being killed).

I may have used the debug kill area on single tiles to kill off trapdoor spiders roaming the map instead of staying close to their traps as they should.

RenechCDDA:
/confirmed

I'm not sure I found the *root* cause of the bug, but the calculation for work/travel time is not deterministic. Repeatedly checking the estimated time for a given tile will produce different estimates, with wildly varying times in some cases. After actually sending them out, the time will be different from the last given estimate as well!

I prepared a save file to illustrate the issue. Load the save and give an order to perform a tree cutting mission targeting the forest 2 tiles to the east (which has been edited to be a square of nothing but oak trees). My given estimate was 1 week, 6 days. My return time was 8 weeks, 4 days!

[...]

PR #65705, opened 2023-05-17
Add exertion and modify trip length calculation for faction camps

FatigueDev:
#### Summary
Features "Add exertion to jobs in the faction camps and modified calculation for trip length"

#### Purpose of change
Upon creating a relay site, when the player tries to send any amount of supplies to that relay site, they will find that the calorie count and duration of the relay makes relay sites completely non-viable. Think moving 50 books taking 3 seasons levels of non-viable (just down the road, I might add).

#### Describe the solution
The solution to the issue was to take a good look into how the trips are calculated, as well as how calories are calculated by the faction camp code and doing a rework to consider the level of exertion as well. This means that the time spent cutting logs is what it was, but the calorie count should remain similar to what it was because it's exertion is very active, but doing some berry picking is rather low exertion so and so that is reflected as well. As for sending and receiving things via the relay, I have increased the 'fake' NPC calculation's gear from a makeshift sling to a duffel bag and increased it's carry weight from 30 kilograms to 60 kilograms. What this does is decrease the overall amount of trips while impacting other tasks minimally until we add some sort of "Assign vehicle to faction camp" solution instead. This change turns 3 seasons and 356,280 calories of moving one book at a time with nonsensical trip calculation code into something usable and even preferred; because now if you loot for example a section of a town, you can drop off the goods at a relay site and it should be moved back in a few hours, days or weeks depending on the weight and volume of the contents of the relay site. Basically- It makes pack rats happy; because now they can automate hauling stuff back to their base using a more robust calculation for a more sane amount of calories.

I've considered just hardcoding the relay site code, but that seemed like a half measure.

#### Testing
I have tested pretty much all cases in which the player will be interacting with the faction system, including but not limited to:
- Relaying a whole bunch of books (100+ liters) to a nearby site.
- Relaying a whole bunch of books (100+ liters) to a 3 jump relay site.
- Relaying a bunch of heavy stuff to neaby sites. Think 200 liter barrels and stuff. As expected, it takes a trip for each but it's a lot more manageable now.
- Sending NPC's out to cut wood and clear cut (I found a _new_ bug in here which has been raised [#61690](https://github.com/CleverRaven/Cataclysm-DDA/issues/61690)
- Doing repeating tasks like firewood, gathering plants and junk like that. Hunting as well.

#### Additional context
N/A


RenechCDDA:
Not sure what's going on with this, but I tried playtesting it. I debugged a simple camp with a board, added *100k* to the larder(confirmed only 100k there), and sent an NPC out. Here's what the window reports for a forest cut (not clearcut) about 6-8 tiles away:

image

Okay, 15k. So we should have 85k, or thereabouts, afterwards? We do not.

image

image

Now the reported time and calories has always been screwy, but this is now an order of magnitude mismatch.

RenechCDDA:
Looks like the culprit is `camp_food_supply`. I set a debugmsg and breakpoint, and it broke while reporting the calories were abysmally low. Reported exertion_level was 1, despite the fact they were on a log cutting mission. Here's the trace:

 	cataclysm-tiles.exe!time_to_food(time_duration work, float exertion_level) Line 4960	C++
	cataclysm-tiles.exe!camp_food_supply(time_duration work) Line 4954	C++
 	cataclysm-tiles.exe!basecamp::start_mission(const mission_id & miss_id, time_duration duration, bool must_feed, const std::string & desc, bool __formal, const std::vector<item *,std::allocator<item *>> & equipment, float exertion_level, const std::map<string_id<Skill>,int,std::less<string_id<Skill>>,std::allocator<std::pair<string_id<Skill> const ,int>>> & required_skills) Line 1752	C++
 	cataclysm-tiles.exe!basecamp::start_mission(const mission_id & miss_id, time_duration duration, bool must_feed, const std::string & desc, bool __formal, const std::vector<item *,std::allocator<item *>> & equipment, const string_id<Skill> & skill_tested, int skill_level, float exertion_level) Line 1731	C++
 	cataclysm-tiles.exe!basecamp::start_cut_logs(const mission_id & miss_id, float exertion_level) Line 2228	C++
 	cataclysm-tiles.exe!basecamp::handle_mission(const ui_mission_id & miss_id) Line 1616	C++
 	cataclysm-tiles.exe!iexamine::bulletin_board(Character & you, const tripoint & examp) Line 1931	C++
 	cataclysm-tiles.exe!game::examine(const tripoint & examp, bool with_pickup) Line 6067	C++

PR #65708, opened 2023-05-18
CTD caused by repair attempts with UPS-equipped tools fixed

Spaceb0i:
#### Summary
Bugfixes "CTD caused by repair attempts with UPS-equipped tools fixed"

#### Purpose of change
This fixes a crash to desktop (null pointer error) described in Issue https://github.com/CleverRaven/Cataclysm-DDA/issues/65643. Simply put, attempting a repair using a UPS-equipped tool would cause a CTD, which is bad.

#### Describe the solution
repair_item_finish now checks if an item uses UPS and sets ammo type and name based on that, and inv_use_ups no longers tries to set ammo on UPS-equipped tools without their own battery. This solution results in the a minor visual bug where the overall capacity of charges in the repair GUI is reported as 0 with UPS tools.

A total rewrite of UPS function is not in my skillset (this is my first time using C++) and I'm not even sure it would do away with the need for all the manual checks everywhere. Making ammo_capacity() consider the amount of available UPS capacity may be in a future PR, but is fraught with too many difficulties to prioritize over the fixing of a crash.

#### Testing
Repairs were attempted with a UPS-equipped makeshift arc welder, a UPS-equipped soldering iron, and a normal soldering iron, all of which worked correctly with no errors. An empty UPS and a UPS with a dead battery, as expected, prevented the UPS tools from repairing. Additionally, a smartphone was used to test if inv_use_ups still worked, and it did, recharging the phone battery any time it dropped.

#### Additional context
~~Some old savegame conversion code meant to be deprecated after 0.F got removed, as it was related to the problem. This is a remake of another PR that I messed up.~~ This was merged in another PR.

### DIFF
--- a/data/json/furniture_and_terrain/appliances.json
+++ b/data/json/furniture_and_terrain/appliances.json
@@ -596,7 +596,7 @@
     "durability": 120,
     "description": "A power cord sticking out of an appliance.  You need to plug it into a powered grid for the appliance to work properly.",
     "item": "power_cord",
-    "flags": [ "NOINSTALL", "NO_UNINSTALL", "UNMOUNT_ON_DAMAGE", "UNMOUNT_ON_MOVE", "POWER_TRANSFER" ]
+    "flags": [ "NO_INSTALL_HIDDEN", "NO_UNINSTALL", "UNMOUNT_ON_DAMAGE", "UNMOUNT_ON_MOVE", "POWER_TRANSFER" ]
   },
   {
     "id": "ap_standing_lamp",
--- a/data/json/items/generic.json
+++ b/data/json/items/generic.json
@@ -2084,7 +2084,7 @@
     "symbol": ",",
     "color": "yellow",
     "name": { "str": "ceramic armor plate" },
-    "description": "A ceramic armor plate, specifically engineered for use in a bullet-resistant vest.",
+    "description": "A ceramic armor plate, specifically engineered for use as armor.  Not suitable for human use.",
     "price": 8000,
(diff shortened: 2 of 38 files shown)


Hirmuolio:
item::ammo_capacity returning something else than item's ammo capacity sounds like very bad idea. 
The UI showing X/0 for available batteries is minor problem. A simple if/else to show something else instead of ammo capacity should work.

Spaceb0i:
> item::ammo_capacity returning something else than item's ammo capacity sounds like very bad idea.
This brings it in parity with ammo_remaining. If a function needs to search for exclusively a UPS-equipped item's internal ammo capacity, they can look at the ammo capacity of its magazine.

Hirmuolio:
ammo_remaining is used to see how much ammo there are in the item that can be used. Since UPS acts as ammo it is counted in it.

ammo_capacity is used to see how many ammo can be put into the item. You can't put ammo into item jus because somewhere else there is room in UPS. So UPS is not counted in it.

There are many places where the game checks for the ammo capacity and then shoves number of ammo into the item based on that. Adding other UPS capacity in ammo_capacity will beak all of them.

If you want to do something with available ups capacity you should write something new for that instead of using ammo_capacity.
This `item::ups_capacity` check would need to check if item is carried by a character. You can't just do `get_player_character()` for the carrier because the item may be on ground, carried by NPC, carried by player or not even exist.  
Probably easiest to have most of the logic in some sort of `charcter::ups_capacity`.
In some situations this capacity should also count UPS items on ground near the character that is being checked.

Issue #35665, opened 2019-11-23
Animals don't follow player up stairs


new02DAY:
# Describe the bug

I'm not sure if this is due to how AI and RNG works, but animals don't climb up stairs.
Save attached below.

# Steps To Reproduce

1. Create new world and character (evac).
2. Spawn cattle fodder and sheep.
3. Walk down the stairs and around until the sheep follows you down.
4. Go back up the stairs and walk around and try to get the sheep up.

# Expected behavior

Even with RNG and AI, I would expect the animal to follow up the stairs, especially in smaller stairwells (which would presumably increase the chance of the animal following).

# Screenshots

**If applicable**, add screenshots to help explain your problem.

# Versions and configuration

- OS: Windows
- OS Version: MINGW/CYGWIN/MSYS2 on unknown Windows version (Win7)
- Game Version: 0.D-9559-gcc3282c [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food]
]

# Additional context

[save_20191122201707.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/3882029/save_20191122201707.zip)
Save contains example of tamed animal in a house, and a shelter SW of current location.

KorGgenT:
are z levels on or off?

new02DAY:
> are z levels on or off?
Z levels are on, but align stairs are off.

mouseturd:
Confirmed I had this same issue with my pet horse. I don't have a save file to share though

yowshi:
yeah it is an issue with slimesprings as well

borazan:
I think align stairs might be causing this, haven't tried it myself tho

matt32106:
Same problem here, save attached
[save.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/4461086/save.zip)
 

- OS: Linux
    - OS Version: LSB Version: core-9.20170808ubuntu1-noarch:printing-9.20170808ubuntu1-noarch:security-9.20170808ubuntu1-noarch; Distributor ID: Ubuntu; Description: Ubuntu 18.04.4 LTS; Release: 18.04; Codename: bionic; 
- Game Version: 0.E-238-g6393a6d [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
[...]

PR #65720, opened 2023-05-18
Only reject air for pathfinding bounds, not pathable tiles without floors

### LINKED ISSUE #65593: Down stairs block pathfinding for purposes of crafting reach
### Describe the bug

Stairs leading down block pathfinding for purposes of crafting reach as if made of an impassable tile; the game cannot see crafting components if the path leading to the components is "blocked" by a down stairs tile. Up stairs are not affected by this issue.

### Attach save file

[Hibernia-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11470783/Hibernia-trimmed.tar.gz)

### Steps to reproduce

1. Get in your workshop
2. Surround yourself with down stairs
3. No crafting components will be visible in the crafting menu, even if normally within the 6-tile range

### Expected behavior

As a passable tile, down stairs should not block crafting component visibility

### Screenshots

_No response_

- OS: Windows 7 64bit
- Game Version: 0.G Stable / d6ec466 64bit
- Graphics Version: Tiles
- Ingame Language: English
- Mods loaded: dda, bionic professions, blaze industries, magiclysm, disable npc needs, portal storms ignore npcs, no fungal growth, tamable wildlife, translate complex dialogue, mythical martial arts, no rail stations, sees_player_retro

### Additional context

The issue also applies to wooden down stairs, ladders leading down, and up/down ladders.

RenechCDDA:
#### Summary
Bugfixes "Stairs doesn't block pathfinding for some purposes (e.g. crafting pickup range)"

#### Purpose of change
* Fixes #65593

#### Describe the solution
map::has_floor_or_support always returns false if going downwards is a valid move. Going downwards on stairs is a valid move, but not the *only* move. Therefore we need to check something more sane, like map::is_open_air.

There shouldn't be any scenarios where you can go downwards (map::has_floor_or_support returns true) that doesn't also return true for map::is_open_air, **except for** stairs.

I originally did this by checking map::passable and then realized that open air is, technically speaking, passable. Soooo that didn't work.

#### Testing
Compiled locally and confirmed I could find crafting components through the stairs. Debugged a new scenario(image below) where I was separated from components by air, confirmed I could not access them.

image

#### Additional context
map::has_floor_or_support is a very sussy name and a very sussy function.

This bug report came from 0.G so if this fix is acceptable it may be desirable to also backport it.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -7089,7 +7089,7 @@ void map::reachable_flood_steps( std::vector<tripoint> &reachable_pts, const tri
         const tripoint tp = { p.xy(), f.z };
         const int tp_cost = move_cost( tp );
         // rejection conditions
-        if( tp_cost < cost_min || tp_cost > cost_max || ( !has_floor_or_support( tp ) && tp != f ) ) {
+        if( tp_cost < cost_min || tp_cost > cost_max || is_open_air( tp ) ) {
             continue;
         }
         // set initial cost for grid point


GlitterLich:
nice sleuthing!! well done! 🎉

Qrox:
Sounds like this could fix #35665?

PatrikLundell:
What about an underwater case where you'd have to check that it isn't open water rather than open air? I think I've seen something about supporting walking on the bottom of bodies of water for some kinds of mutants, for instance.

RenechCDDA:
These are *rejection* criteria so if anything being able to path across water is too permissive. The pickup range is only 6 tiles though and I cannot imagine any scenario where that actually comes into play.

PR #65722, opened 2023-05-18
[Interface] Fix to unify tilesets behavior (part 1)

vetall812:
#### Summary
Interface "tilesets unification dirt connection behavior"

#### Purpose of change
Some terrain should connect to each other because they are the same material (like sand and sandmound) and some terrain should connect to each other because of common sense (like sand mound on the construction site should be "on the dirt" instead of forming a border. Those fixes unifies behavior for all tilesets (and didn't break anything).

#### Describe the solution
added some connection groups
some terrains assigned to those groups

leave as is

#### Testing
As you can see on the screenshots below dirt now have similar behavior in all tilesets (NOT forming grass border around sand in Ultica for example)

image
image
image


vetall812:
image
image

PR #65726, opened 2023-05-19
watercannon WONT_TRAIN_MARKSMANSHIP

anoobindisguise:
#### Summary
Balance "fire truck water cannon doesn't train skills"

#### Purpose of change
Splashing a zombie with a super soaker should not teach you how to use a rocket launcher. 

#### Describe the solution
This does mean that the launchers skill is virtually untrainable but tbf I'm not sure if that actually matters.

### DIFF
--- a/data/json/items/ranged/launchers.json
+++ b/data/json/items/ranged/launchers.json
@@ -92,7 +92,7 @@
     "description": "A water cannon from a fire truck.  Mounted on a vehicle, it could be used to stop fires or \"protesters\".",
     "price": 50000,
     "material": [ "steel" ],
-    "flags": [ "NEVER_JAMS", "MOUNTED_GUN", "NO_RELOAD", "NON_FOULING" ],
+    "flags": [ "NEVER_JAMS", "MOUNTED_GUN", "NO_RELOAD", "NON_FOULING", "WONT_TRAIN_MARKSMANSHIP" ],
     "ammo_effects": [ "JET", "BEANBAG", "NEVER_MISFIRES", "RECOVER_10" ],
     "skill": "launcher",
     "ammo": [ "water" ],


fairyarmadillo:
We need a potato gun and a practice action to train with it. I'm serious btw

Qrox:
Or rather we need a watercannon proficiency, similar to what I suggested in #65682 for nailguns. On a related note, flamethrowers also use the launcher skill but I'd imagine operating one is very different from operating a grenade launcher or an RPG, so we really need separate proficiencies for them. And correct me if I'm wrong, but actually I imagine operating a flamethrower will be more similar to operating a water cannon, since the flamethrower sends out liquid fuel similar to how a water cannon sends out water, but with additional burning.

Ichthysdekilt:
If very little trains the launchers skill, is there a reason it is a skill and not something like a proficiency? Or perhaps relies only on marksmanship? On the other hand, should each use provide a very large amount of XP since almost no one in or out of an apocalypse is going to fire thousands of rockets by hand?

Havok91:
And this is why I made the comment on the Nail Gun post, the only way to train these skills is by using valuable consumable items, or building obscenely expensive training equipment. Proficiencies in their current iteration don't make a whole lot of sense for weapons, as they currently only reduce crafting times and failure chances. However, allowing related items to be used to train a basic skill makes sense, give it diminishing returns so that beyond a certain point the XP gain becomes negligible. Or, an even better idea, make a craftable "snap cap" type of ammo that is used for training purposes, or a dummy round for launchers that is reloadable with minimal gunpowder so that we can use designated targets as training grounds for things like launchers or firearms so that it cannot be easily "exploited" by using a nail gun to get to level 10 marksmanship or whatever. Either way, completely removing the ability to practicably gain XP without a severe waste of resources is getting ridiculous.

fairyarmadillo:
The point is to make you spend your resources rather than maxing all skills in a risk-free cost-free environment. The world is awash in .22, there's no reason not to use it.

Havok91:
How is .22 going to raise launcher skill? How many rockets, grenades, etc. can I realistically find in a playthrough to make launchers more than laughably bad to even attempt to use? How badly are we going to nerf things into the ground before people start quitting the game because there is no reasonable ways left to get XP without scouring hundreds of square miles worth of map tiles to find enough resources to hopefully make the weapons we want to use not borderline suicidal? Using a water cannon to get the first couple of levels of launcher XP makes sense, it operates similarly enough to the flamethrower, not sure about comparing it to a bazooka or a 40mm grenade launcher, but that issue is more bundled in the suspension of disbelief aspect of a video game. The same way that learning to shoot a Sig Mosquito like a pro isn't going to magically transform me into a god when I switch over to a .50AE Desert Eagle, the guns are too dissimilar. This can be applied to almost all the weapon systems/types in the game. The solution isn't in this heavy handed approach though, maybe set a level cap, say 2, for improvised or strange weapons (IE Water Cannon, Nail Gun, etc) reduce the rate at which you gain marksmanship XP so that it's not drastically outpacing firearm skillsets, the list goes on for less harsh solutions, but we just keep seeing the heaviest of nerfs instead of balanced approaches, which is discouraging for everyone. Right now the meta seems to be find rifle and tons of ammo, shoot everything, have heaviest melee as backup. Pistols-basically useless except the biggest of calibers, Launchers-basically suicide unless you put points in the skill at CC, Shotguns-useless against armor and anything more than point blank, Throwing-actually useful-ish once you train up by pelting things with rocks (but this will likely be the next nerf, then it'll be same as launchers) I'm not asking for no-risk no-cost skill farming (which is not true anyhow, as time is always a cost because of evolution. Yes my launcher skill is level 9 due to the water cannon I used to hose 4 zeds against a wall with for the last 3 in-game weeks, step into a new tile with shit all gear and 2 grenades for a 40mm and get bitch slapped by a couple skelly-jugs and glowy-bois) but we need a way to make our weapons less suicidal the first time we walk out the door.

PR #65728, opened 2023-05-19
Cold tolerance cancels cold blooded, remove from plant

anoobindisguise:
#### Summary
Balance "cold tolerance mutation is exclusive from cold bloodedness and isn't a plant mutation"

#### Purpose of change
Cold tolerance seemed like it should be mutually exclusive with being cold blooded.

#### Describe the solution
Do that by adding "ECTOTHERMIC" to its category and also remove it from Plant (it was already exclusive with plant's heat tolerance, but it would make the issue worse by shuffling muts over and over)

Do direct "cancels" instead of adding the category.


Maleclypse (approved):

fairyarmadillo:
Please reconsider this. These mutations do totally different things and there are good arguments for not making them exclusive, namely that integrated armor prevents mutants from wearing warm clothing.

There were plans to give sea creatures cold tolerance, as these animals are cold blooded but live in very cold environments IRL. They are protected from injury by having anatomy which is harder for cold to damage, but would still be reasonably slowed down by the cold.

Plant in particular is a prime example of this. A tree in the wintertime goes dormant and is inactive, but it is at no risk of getting frostbite. Furthermore, the current solution is to wear a coat, but a coat would not actually warm an ectothermic organism.

anoobindisguise:
Part of the issue is again, that heat and cold tolerance conflict now (they didn't before when first added) so anything that had both in its tree would ricochet between the two. That could be fixed also.

KittyTac:
> Part of the issue is again, that heat and cold tolerance conflict now (they didn't before when first added) so anything that had both in its tree would ricochet between the two. That could be fixed also.

With my #65750 they will no longer conflict.

PR #65732, opened 2023-05-19
Faction succession without death

RenechCDDA:
#### Summary
Content "You can now choose a new leader for your faction without dying"

#### Purpose of change
We have succession on death! It's great! It's very cool! It offers a lot of roleplay and gameplay value!

But, what if there are times when the player character should hand the reigns over *without* dying? (Like, say, if we ever get long-term wounds and/or disabilities...)

#### Describe the solution
We make a way for the player to, legitimately, body swap. Once every three months at maximum. This PR adds a new 'mission' to all faction camp bulletin boards, and a series of menu selections for one of three choices.

-Autocratic succession: What it says on the tin. You choose.

-Sortition: A *completely* random NPC among all your followers is picked.

-Democratic: A random NPC (weighted towards higher social skills) is picked. (Specifically, the formula is currently set to `rng(0,5) + rng (0, social * 2)`. So it's pretty heavily weighted, but never guaranteed.)

Currently succession will *never* be called outside of the player selecting it as an option. To actually go through with it, you have to select the mission and accept a yes/no prompt. So it should be impossible to do this by accident.

I suppose I *could* just post a sign pointing to the debug menu...

#### Testing
It compiles! It works!

It's proofed against calling the succession even if you have no followers! (It will still go through, but it won't call nullptr and crash or anything like that)

<details>

image

image

image

image

</details>

#### Additional context
I haven't hooked the cooldown up yet and it's 0315 so I'm not going to figure out HOW to do that tonight. That's why it's a draft. *For now.*


Qrox:
Alternatively, add guillotines to the game and once you piss off your NPCs bad enough they build one and execute you.

RenechCDDA:
That's for the followup PRs!

(Time will tell how serious this statement is....)

PatrikLundell:
I'm not sure how you'd use a guillotine to facilitate succession without death... Well, I guess through "Step down, or else!", possibly.

RenechCDDA:
I've made an abomination but it's now a *fully working* abomination, with a 90-day cooldown and everything. And it tells you what the cooldown is!

I'm aware I could turn basically the whole thing into EOCs, but the main stumbling block with that is (re)writing the comparison logic for getting a list of follower NPCs+their social skill will probably be a nightmare to write. It already gave me a lot of trouble in C++, a language I've used at least a bit. I'm just... really not enthused about the idea.

So I'm marking this ready for review. It is functionally ready. I guess we'll see what the mergers think.

FickDiddler:
Crusader Kings in my Cataclysm? Nice!

src/faction_camp.cpp
```diff
@@ -1475,6 +1491,102 @@ void basecamp::get_available_missions( mission_data &mission_key )
...
+
+    if( choice < 0 || static_cast<size_t>( choice ) >= choices.size() ) {
+        popup( _( "You choose to wait…" ) );
+        return;
+    }
+
+    // Autocratic
+    if( choice == 0 ) {
+        if( !query_yn(
+                _( "As an experienced leader, only you know what will be required of future leaders.  You will choose.  \n\nIs this acceptable?" ) ) ) {
```
Brambor, line 1522:
```suggestion
                _( "As an experienced leader, only you know what will be required of future leaders.  You will choose.\n\nIs this acceptable?" ) ) ) {
```

src/faction_camp.cpp
```diff
@@ -1475,6 +1491,102 @@ void basecamp::get_available_missions( mission_data &mission_key )
...
+            // Yes this is a very barren representation of who gets elected in a democracy. Too bad!
+            int popularity = rng( 0, 5 ) + rng( 0, follower->get_skill_level( skill_speech ) * 2 );
+            followers.emplace_back( follower, popularity );
+            charnum++;
+        }
+    }
+    // Sortition
+    if( choice == 1 ) {
+        if( !query_yn(
+                _( "You will allow fate to choose the next leader.  Whether it's by dice, drawing straws, or picking names out of a hat, it will be purely random.  \n\nIs this acceptable?" ) ) ) {
```
Brambor, line 1550:
```suggestion
                _( "You will allow fate to choose the next leader.  Whether it's by dice, drawing straws, or picking names out of a hat, it will be purely random.\n\nIs this acceptable?" ) ) ) {
```

src/faction_camp.cpp
```diff
@@ -1475,6 +1491,102 @@ void basecamp::get_available_missions( mission_data &mission_key )
...
+            return;
+        }
+        int selected = rng( 0, charnum );
+        // Vector starts at 0, we inserted 'you' first, 0 will always be 'you' pre-sort (that's why we don't sort unless democracy is called)
+        if( selected == 0 ) {
+            popup( _( "Fate calls for you to remain in your role as leader... for now." ) );
```
Brambor, line 1556:
```suggestion
            popup( _( "Fate calls for you to remain in your role as leader… for now." ) );
```

src/faction_camp.cpp
```diff
@@ -1475,6 +1491,102 @@ void basecamp::get_available_missions( mission_data &mission_key )
...
+        npc_ptr chosen = followers.at( selected ).first;
+        popup( _( "Fate chooses %s to lead your faction." ), chosen->get_name() );
+        get_avatar().control_npc( *chosen, false );
+        return;
+    }
+
+    // Democratic
+    if( choice == 2 ) {
+        if( !query_yn(
+                _( "A leader can only lead those willing to follow.  Everyone must get a say in choosing the new leader.  \n\nIs this acceptable?" ) ) ) {
```
Brambor, line 1570:
```suggestion
                _( "A leader can only lead those willing to follow.  Everyone must get a say in choosing the new leader.\n\nIs this acceptable?" ) ) ) {
```

Brambor (changes requested):
There, solving all but two checks.

WreckingAbandon:
Would there be effects for each succession option? Like autocratic succession causing discontent which can be averted perhaps by current avatar's social skill?

Or maybe to make it simpler just grey out an option that does not meet specific requirements, like a charismatic avatar being needed to do the autocratic succession.

Or even more simpler, I guess just leave it to the imagination. 🤷‍♂️

RenechCDDA:
> Would there be effects for each succession option?

Not at the moment, maybe in the future. It's certainly something I've thought about. NPCs very much have a problem with being suicidally subservient to the player, so moving away from that direction is a good idea IMO. For example if you autocratically choose a new leader and people in your group barely trust you... they might think "Well why SHOULD I follow the new guy?".

Obviously any implementation of this would be in another PR at another time. Gotta get this fixed up right now, looks like it might be wrecking the tests with current implementation.

RenechCDDA:
Oookay, *this* should finally build. Hopefully. This was a massive pain in the ass to track down too, and even after researching what was going on I still don't know why this was only causing failures *here*.

tests/new_character_test.cpp
```diff
@@ -204,6 +205,9 @@ TEST_CASE( "Generated character with category mutations", "[mutation]" )
 {
     REQUIRE( !trait_TAIL_FLUFFY.obj().category.empty() );
     avatar &u = get_avatar();
+    clear_avatar();
+    u.clear_mutations();
+    u.mutation_category_level.clear();
```
mqrause, line 210: This looks like stuff that should be in `clear_avatar`.

RenechCDDA: Sadly it is definitely not because I re-ran the tests locally, adding each line until it stopped complaining.

...But if you mean in general "it should be in there!" I agree, absolutely.

mqrause: Yeah, if moving that in `clear_avatar` breaks too many other tests, that's a thing for a different PR.

RenechCDDA: I didn't actually test moving those lines into clear_avatar, but I'll try to give it a shake after this PR maybe.

RenechCDDA:
Another broken test, another fix. Why does modifying new_character_test.cpp break the starve_test in stomach_contents_test.cpp? Good question! 

I don't have an answer for you. I only know that it is now not resetting the activity properly and therefore I must bash it until it does, in order to preserve what the test actually does.

The last commit was actually a pared down version of a butchery, but it should still work. If not, there's always more force pushing! Weee!

tests/new_character_test.cpp
RenechCDDA: clear_mutations() is actually already in clear_avatar() (not sure why I thought otherwise) but mutation_category_level.clear() is something I'll try to move in there when my sanity can withstand more test dancing

Maleclypse:
Can you resolve the conflicts on this?  Ping me once you've done so and I can test.

RenechCDDA:
@Maleclypse You should be good to checkout the branch now. It's rebased onto master too, so hopefully it has no problems with tests.

RenechCDDA:
Looks like something changed on the EOC infrastructure side, the json fix is simple but I need to make sure the var still has the same name

RenechCDDA:
Err, okay, the deactivate condition no longer works but all that means really is a possible performance loss. It's a single value check, once every 5 minutes at most. So THIS should be good.

Weird that the json became invalid though. It was definitely working back when I committed it, and I don't see any changes which should have impacted it.

Maleclypse:
Ok so I compiled it and started testing it and had a faction camp with myself and a single follower.  When I sped time forward I was able to select the NPC.  Upon becoming the NPC a string of achievements fired off.  During one of the achievements the game ctd.  Will try again.

Maleclypse:
Yep every time I get a ctd.  If you can confirm you are not getting a ctd it's probably something with my dev build

RenechCDDA:
Can confirm I haven't experienced any CTDs. To be sure though I'd recommend posting the crash log? The code does have some nullpointers run around, but they're all sectioned out so they should never be accessed.

The act of switching characters is also giving a `false` argument for debug switching, so it's intended you are able to be able to keep getting achievements. Maybe there is something down the line I need to investigate.

kholat:
The 90 days cooldown makes this feature practically useless as a player.

RenechCDDA:
The cooldown is in json, you're welcome to mod it. Commenting on PRs that were merged months ago is definitely not the way to go.

kholat:
Do you sincerely believe that 90 days cooldown to swap the actor with an npc is what should be the default?

I am commenting because this is the PR that introduces this feature (well it was available before in the debug menu but I mean as a legit feature in the game), and I do not think that blindly starting a new PR to change the number without commenting here first would be a better idea; might as well start at the source. I am essentially asking from where does the 90 come from, why has it been decided that it should be 90, after the "months ago" implementation perhaps people have had some playtime and maybe can input some ideas if they think this is actually a good number or not. After all there is no explanation anywhere in the PR of why the number 90 was chosen.

As you can probably tell I think that this is a far too high number to be set as default (I know players can change it in json, this is not the point of my comment) mostly because it only acts as a hinderance to the player.

I imagine the player + 3 followers sitting at a campfire and them talking about who should lead now and then some god-being-from-above going "NO! YE SHALL WAIT FOR 87 MORE DAYS AND EXACTLY 87 DAYS!" (after the last leader was in charge for 3 days)

RenechCDDA:
> Do you sincerely believe that 90 days cooldown to swap the actor with an npc is what should be the default?

Yes, I sincerely believe that even a mere 90 days strains credibility. As it stands, people are far too willing to trust their life to you(the player) or the chosen 'leader' of your faction. You can recruit one of the beggars from the refugee center, walk back to camp, and immediately put them in charge. That is pretty silly. Being able to do that once a day? Farcical.

Now as I said, commenting on PRs closed months ago is not the way to go. If you have criticisms or suggestions, please make an issue or PR addressing them.

kholat:
Thats a good argument. I thought this feature was meant to be like in state of decay games where you can swap between characters as a way to kinda make the player more invested in them and to kinda run multiple different characters in the same game. Still the number 90 feels extremely arbitrary even taking into account your argument at least from my perspective. Ill think about it more and perhaps make a PR changing it if I have a seemingly good idea.

Merikona:
Hi, thank you for making this PR. I imagine this will be a huge QoL feature for me and I will be updating to experimental soon just to use it. One thing I struggle with in CDDA is that I want to see everything there is to see (and there is so much), but because of the permadeath nature of the game, the fact that I do not save-scum, and the time and resources you have to invest in a character to level their crafting skills, I'm also incredibly risk-averse. I'm fine with the 90 days cooldown for changing leaders (it makes sense you can't change leaders every other day), but do you think it would also be possible to swap characters but make it so you can't give orders to other NPCs in the meantime? The idea being you are merely controlling another member of your faction and playing out their story, but that member you're controlling doesn't have to be the leader. You wouldn't have your chemist or engineer leading a charge into a bandits camp, especially if they're the only ones with such expertise in your group, your group is large, and there are people more suitable for combat. You could eventually specialize NPCs. Some could have intelligence boosting bionics and mutations, and run the camp. Others could have combat bionics, chimaera/beast mutations, etc.

PR #61074, opened 2022-09-17
backport z-level draw

sonphantrung:
#### Summary
None

#### Purpose of change
I want to see zeds from below

#### Describe the solution
Port the change from [BN](https://github.com/cataclysmbnteam/Cataclysm-BN/pull/1743) by @joveeater

#### Describe alternatives you've considered
Not doing this

#### Testing
TBD, I don't have a good machine to compile

#### Additional context
my head is currently sleepy, I'll keep an eye on the code later ;)

PR #65738, opened 2023-05-19
3D vision of lower levels with distance fog

Rewryte:
#### Summary
Features "3D vision of lower levels with distance fog"

#### Purpose of change
The current method of drawing lower z-levels removes plenty of visual information that should otherwise be assessible to the player character.

before

#### Describe the solution
Draw lower z-levels.  Naturally, with the top-down perspective most tilesets use, a major issue would be that there is no way to tell which sprites belong to which z-level.  To solve this, `t_open_air` and other transparent terrain can now be assigned a semi-transparent sprite to create fog or other effects that emulate distance.  Tileset does not have a sprite?  The fallback graphic for transparent terrain now includes a semi-transparent overlay.  (The following picture uses clairvoyance for clarity.)

after

It adapts its brightness to ambient lighting giving night lights some extra flair.

light

Number of z-levels displayed is determined by 3D vision settings.  Setting vertical view distance to 0 will revert drawing behavior to that prior to this PR.

#### Testing

- LOS works as expected
- Works with night vision / clairvoyance
- Respects experimental 3d vision settings for spotting creatures
- Mostly works with map memory (There are a few spots on lower z-levels that do not get recalled when they should.  More feedback on this would be helpful.)
- Mostly working on isometric tilesets but disabled for now due to some issues with height_3d.  Future plan if this gets merged.

Performance test: https://github.com/CleverRaven/Cataclysm-DDA/pull/65738#issuecomment-1556103530
Modding example: https://github.com/CleverRaven/Cataclysm-DDA/pull/65738#issuecomment-1560693475

#### Additional context
This PR modifies tile drawing which indirectly affects the map memory system.  I would appreciate it if someone familiar with the memory code could look through this to make sure it wouldn't cause any complications.

The following changes will be made to non-isometric tilesets:

- `t_open_air` is now used as the fog between z-levels so if it is assigned a blank sprite, it will be unassigned.  Having a custom `t_open_air` is still recommended for a minor performance improvement.
- If `tr_ledge` is not assigned a sprite, it would be assigned a blank sprite, typically replacing `t_open_air`.  This is to prevent it getting replaced with an opaque placeholder.


RenechCDDA:
I'm not qualified to comment on the technical implementation, but are you aware of Kevin's previous concerns with similar features? (See: https://github.com/CleverRaven/Cataclysm-DDA/pull/61074#issuecomment-1257085629)

Rewryte:
@RenechCDDA Can't say that I am.  Thanks for bringing this up.  I'll look into it when I have the time.

src/cata_tiles.cpp
```diff
@@ -1572,27 +1573,67 @@ void cata_tiles::draw( const point &dest, const tripoint &center, int width, int
...
+
+                // Draw all layers for the bottom z-level
+                for( auto f : drawing_layers ) {
+                    ( this->*f )( p_draw, p.ll, cur_height_3d, p.invisible );
+                }
+
+                // Keep going up and drawing until current position
+                while( p_draw.z < p.pos.z ) {
+                    // Draws fog overlay to give the impression of distance
+                    draw_zlevel_overlay( p_draw, p.ll, color_blocks );
```
andrei8l, line 1627: You can compute the alpha and draw the overlay only once.
I used `int alpha = 77.3059 * std::log(3.9131 * std::abs( p.pos.z - lowest ));` as an approximation but you can also compute the real blended alpha with `dstA = srcA + (dstA * (1-srcA))` in a loop.
<details>
<summary>before, with multiple rects - staring down from z-level 6</summary>

6b


</details>

<details>
<summary>after, with alpha approximation</summary>

6a


</details>

<details>
<summary>patch</summary>

```diff
diff --git a/src/cata_tiles.cpp b/src/cata_tiles.cpp
index 2661c55e27..070c6105c7 100644
--- a/src/cata_tiles.cpp
+++ b/src/cata_tiles.cpp
@@ -1615,6 +1615,7 @@ void cata_tiles::draw( const point &dest, const tripoint &center, int width, int
                     p_draw.z -= 1;
                     cur_height_3d -= 1;
                 }
+                int const lowest = p_draw.z;
 
                 // Draw all layers for the bottom z-level
                 for( auto f : drawing_layers ) {
@@ -1623,14 +1624,18 @@ void cata_tiles::draw( const point &dest, const tripoint &center, int width, int
 
                 // Keep going up and drawing until current position
                 while( p_draw.z < p.pos.z ) {
-                    // Draws fog overlay to give the impression of distance
-                    draw_zlevel_overlay( p_draw, p.ll, color_blocks );
                     p_draw.z += 1;
                     cur_height_3d += 1;
                     for( auto f : drawing_layers ) {
                         ( this->*f )( p_draw, p.ll, cur_height_3d, p.invisible );
                     }
                 }
+                
+                if( lowest != p.pos.x ) {
+                    int alpha = 77.3059 * std::log(3.9131 * std::abs( p.pos.z - lowest ));
+                    // Draws fog overlay to give the impression of distance
+                    draw_zlevel_overlay( p_draw, p.ll, color_blocks, alpha );
+                }
             }
         }
 
@@ -3841,7 +3846,7 @@ bool cata_tiles::draw_zombie_revival_indicators( const tripoint &pos, const lit_
 }
 
 void cata_tiles::draw_zlevel_overlay( const tripoint &p, const lit_level ll,
-                                      color_block_overlay_container &color_blocks )
+                                      color_block_overlay_container &color_blocks, int alpha )
 {
     if( is_isometric() ) {
         return;
@@ -3862,7 +3867,7 @@ void cata_tiles::draw_zlevel_overlay( const tripoint &p, const lit_level ll,
             fog_color = curses_color_to_SDL( c_dark_gray );
         }
         // Setting for fog transparancy
-        fog_color.a = 100;
+        fog_color.a = alpha;
 
         // Transparancy will only work in blend mode
         color_blocks.first = SDL_BLENDMODE_BLEND;
diff --git a/src/cata_tiles.h b/src/cata_tiles.h
index 7b5093147d..5af01574dc 100644
--- a/src/cata_tiles.h
+++ b/src/cata_tiles.h
@@ -533,7 +533,7 @@ class cata_tiles
         bool draw_zombie_revival_indicators( const tripoint &pos, lit_level ll, int &height_3d,
                                              const std::array<bool, 5> &invisible );
         void draw_zlevel_overlay( const tripoint &p, lit_level ll,
-                                  color_block_overlay_container &color_blocks );
+                                  color_block_overlay_container &color_blocks, int alpha );
         void draw_entity_with_overlays( const Character &ch, const tripoint &p, lit_level ll,
                                         int &height_3d );
 
```

</details>

Also, the function is a little misnamed as it doesn't draw anything itself.

Rewryte: I actually did something similar in my earlier iterations, but I wanted to make sure that the z-level of flying creatures are accurately represented.  Now that I've done some testing, I realized that color blocks tend to draw at the top-most layer regardless, which is something I'll have to look into, but at least for custom fog textures, drawing multiple layers will allow us to accurately perceive airborne objects.

bee

I get that this will do a bit more drawing though.  Think it is worth it?

Qrox: You could probably draw the overlay only once with alpha approximation when there's actually something else to draw on a particular z-level, so when there's only one creature in between, only two overlay draws are need. Not sure how to check whether there's something to redraw though given the current code calls the redraw functions unconditionally.

src/cata_tiles.cpp
```diff
@@ -1572,27 +1573,67 @@ void cata_tiles::draw( const point &dest, const tripoint &center, int width, int
...
+                    p_draw.z += 1;
+                    cur_height_3d += 1;
```
Qrox, line 1629: Should `cur_height_3d` carry over to the next z-level as it currently does? I'm under the impression that `height_3d` represents the base height at which an object is drawn in iso mode and may be modified by the layer drawing functions, so it probably should either be reset to 0 or incremented by a fixed amount (ignoring the changes made by the overlay drawing functions) between z-levels.

src/cata_tiles.cpp
```diff
@@ -3799,6 +3840,37 @@ bool cata_tiles::draw_zombie_revival_indicators( const tripoint &pos, const lit_
...
 
+void cata_tiles::draw_zlevel_overlay( const tripoint &p, const lit_level ll,
+                                      color_block_overlay_container &color_blocks )
+{
+    if( is_isometric() ) {
+        return;
+    }
+    // Draw tileset fog sprite if exists
+    if( tileset_ptr->find_tile_type( "zlevel_fog" ) ) {
+        draw_from_id_string( "zlevel_fog", TILE_CATEGORY::NONE, empty_string, p, 0, 0, ll, false );
```
Qrox, line 3851: Does this support tile variants? A single fog sprite might look very repetitive.

src/cata_tiles.cpp
```diff
@@ -1377,6 +1377,7 @@ void cata_tiles::draw( const point &dest, const tripoint &center, int width, int
             const int &x = pos.x;
             const int &y = pos.y;
 
+            // light level is now used for choosing between grayscale filter and normal lit tiles.
```
Qrox, line 1380:
```suggestion
            // light level is used for choosing between grayscale filter and normal lit tiles.
```

src/cata_tiles.cpp
```diff
@@ -1572,27 +1573,67 @@ void cata_tiles::draw( const point &dest, const tripoint &center, int width, int
...
+            // Otherwise they will be replaced with an opaque fallback sprite
+            // 2. May worsen visibility issues in isometric tilesets with inconsistent terrain heights
             for( tile_render_info &p : draw_points ) {
-                ( this->*f )( p.pos, p.ll, p.height_3d, p.invisible );
+                for( auto f : drawing_layers_legacy ) {
+                    ( this->*f )( p.pos, p.ll, p.height_3d, p.invisible );
+                }
+            }
+        } else {
+            // For every point we drew terrain for ...
```
Qrox, line 1609:
```suggestion
```
Terrain is now drawn together with other layers so this comment no longer makes sense.

Qrox:
> Kevin's previous concerns with similar features? (See: https://github.com/CleverRaven/Cataclysm-DDA/pull/61074#issuecomment-1257085629)

If I may chime in, curses lacks full color support so the best that can be done on curses is probably to darken the symbols and/or use the `.` symbol one or two level below, and completely hide anything any further.

I'm not sure about the memory usage though, it could increase a bit due to caching and whatnot now that more sprites are drawn, but it probably would not be very significant because the current implementation, from what I can tell, draws the additional sprites with the existing tile context so at its worst it will be similar to the memory consumption when you zoom out.

@Rewryte My concern would be with the *time* it takes to redraw though, so it would be nice if you could run the framerate benchmark in the debug menu and post the results before and after this change.

Rewryte:
Thanks for your reviews @Qrox.  It might take me a while to go through them.  Still trying to get my compiler to make friends with curses.  I did a quick test on private working memory usage:

**Before changes**: 1189460K
**After changes**: 1202828K
**After changes (custom fog)**: 1234056K

Tried using the debug benchmark function but performance is capped to my framerate despite using exclusive fullscreen and disabling v-sync on my GPU.  Any ideas?

src/cata_tiles.cpp
Rewryte: It works in non-iso mode too.  I coded it that way so that two similar creatures, one on top of the other, would not overlap each other perfectly.

src/cata_tiles.cpp
Rewryte: I believe the draw functions are booleans which only return true if something is drawn, so perhaps it is possible.  Still, I'd like to find a way to accurately measure performance before I proceed with optimization.  Might experiment with the benchmark feature a bit.

Qrox:
> Tried using the debug benchmark function but performance is capped to my framerate despite using exclusive fullscreen and disabling v-sync on my GPU. Any ideas?

I think I saw a PR in the BN repo that solves this by removing the `refresh_display` call in the benchmark code.

Rewryte:
> I think I saw a PR in the BN repo that solves this by removing the `refresh_display` call in the benchmark code.

Yup, that was it.  Cheers.

Rewryte:
### Performance Testing Results

**Scenario 1**


**Scenario 2**


Performance is given by number of draws in 5 seconds
Memory usage is given by private working RAM in KB

**Without 3D Vision**

Scenario 1
Performance: 395 
Memory: 1187172K

Scenario 2
Performance: 477
Memory: 1040852K

**3D Vision (Color blocks)**

Scenario 1
Performance: 311 
Memory: 1202392K

Scenario 2
Performance: 413 
Memory: 1069640K

**3D Vision (Custom fog)**

Scenario 1
Performance: 308 
Memory: 1186452K

Scenario 2
Performance: 412 
Memory: 1085576K

**Conclusion:** 3D Vision increases draw times by 15-28%.  Might improve with further optimization.  Additional memory usage should be negligible.

src/cata_tiles.cpp
Rewryte: It does now.  Thanks for pointing it out.

bombasticSlacks:
Looking very promising 👍

Kamayana:
A minor thing after the importance of performance, but would it be possible to change the fog color to some kind of indigo, or allow customizing it (not via sprite, just by the color)? The white fog makes the world look very smokey. Previous attempts at this used a dark indigo for the fog and it looked nice:

Fog Example

Vaskritaya:
> A minor thing after the importance of performance, but would it be possible to change the fog color to some kind of indigo, or allow customizing it (not via sprite, just by the color)? The white fog makes the world look very smokey. Previous attempts at this used a dark indigo for the fog and it looked nice:
> 

I find this tinted example to be much easier to differentiate from the current level, as well. I'm not sure I could reliably tell if the pure grey fog is below me or not, at least not as quickly.

Rewryte:
I guess that's one way to do it.  Does the tint get stronger with each successive z-level?  In my mind, it would be difficult to tell the lower z-levels apart if they were all tinted similarly, but I might be wrong.

I could add new graphics options, but I'm under the impression that option bloat is highly discouraged.  The sprite based system already offers plenty of customization through graphics mods after all.  It supports animated sprites, variant sprites with different weights, multitiles with their corners and whatnot.  All without any hardcoded stuff.

Rewryte:
> I'm not qualified to comment on the technical implementation, but are you aware of Kevin's previous concerns with similar features? (See: [#61074 (comment)](https://github.com/CleverRaven/Cataclysm-DDA/pull/61074#issuecomment-1257085629))

> If I may chime in, curses lacks full color support so the best that can be done on curses is probably to darken the symbols and/or use the `.` symbol one or two level below, and completely hide anything any further.

Reckon there is not much more I can do for curses.  It already displays the terrain below.  It even shows us where creatures below are - a feature not currently found in tiles.  I think this is as close as we can get while maintaining readability.

src/cata_tiles.cpp
```diff
@@ -1572,27 +1573,66 @@ void cata_tiles::draw( const point &dest, const tripoint &center, int width, int
...
+                    ( this->*f )( p.pos, p.ll, p.height_3d, p.invisible );
+                }
+            }
+        } else {
             for( tile_render_info &p : draw_points ) {
-                ( this->*f )( p.pos, p.ll, p.height_3d, p.invisible );
+                // Find lowest z-level to draw
+                tripoint p_draw = p.pos;
+                int cur_height_3d = p.height_3d;
+                while( !here.dont_draw_lower_floor( p_draw ) && p_draw.z > -10 ) {
```
Night-Pryanik, line 1613: The same, but without magic number.
```suggestion
                while( !here.dont_draw_lower_floor( p_draw ) && p_draw.z > -OVERMAP_DEPTH) {
```

src/cata_tiles.cpp
```diff
@@ -3799,6 +3839,51 @@ bool cata_tiles::draw_zombie_revival_indicators( const tripoint &pos, const lit_
...
+        point fog_loc = player_to_screen( p.xy() );
+
+        // Overlay color is based on light level
+        SDL_Color fog_color = curses_color_to_SDL( c_black );
+        if( ll == lit_level::BRIGHT_ONLY || ll == lit_level::BRIGHT || ll == lit_level::LIT ) {
+            fog_color = curses_color_to_SDL( c_light_gray );
+        } else if( ll == lit_level::LOW ) {
+            fog_color = curses_color_to_SDL( c_dark_gray );
+        }
+        // Setting for fog transparancy
```
Night-Pryanik, line 3878:
```suggestion
        // Setting for fog transparency
```

src/cata_tiles.cpp
```diff
@@ -3799,6 +3839,51 @@ bool cata_tiles::draw_zombie_revival_indicators( const tripoint &pos, const lit_
...
+        fog_color.a = 100;
+
+        // Transparancy will only work in blend mode
```
Night-Pryanik, line 3881:
```suggestion
        // Transparency will only work in blend mode
```

Rewryte: Thanks

Kamayana:
> I guess that's one way to do it. Does the tint get stronger with each successive z-level? In my mind, it would be difficult to tell the lower z-levels apart if they were all tinted similarly, but I might be wrong.

I'd still want the transparency overlay like you're doing. A better image for what I'm imagining is how Dwarf Fortress does it (though I think indigo is a better overlay color than the cyan they use).

image
http://www.kitfoxgames.com/sendy/uploads/1586439223.gif

Rewryte:
Sure we can do that.  Tossing a semi-transparent purple sprite named zlevel_fog.png and a mod_tileset.json with

`[
  {
    "type": "mod_tileset",
    "compatibility": [ "UltimateCataclysm" ],
    "tiles-new": [
      { "file": "zlevel_fog.png", "tiles": [ { "id": "zlevel_fog", "fg": 0 } ] }
    ]
  }
]`

into a new or existing mod gets me this

test

src/cata_tiles.cpp
Rewryte: Tried to implement this but it didn't go well with custom sprites.  Forcing the alpha value of sprites to a predetermined value would simply cause them to display wrongly unless they were completely opaque in the first place.  My latest performance tests show that the majority of additional draw time is attributable to drawing the extra z-levels rather than the fog, so the performance improvement would have been limited anyways.

Rewryte:
Just a quick update on the optimization works.  Turns out that the single biggest cause of the performance drop from this PR is drawing t_open_air, making up an estimated 65% of the extra draw time.  Even though many tilesets do not even include a sprite for it, the terrain is processed just like any other, with subtiles and rotations calculated amongst other bells and whistles.  In fact, giving t_open_air a sprite actually speeds things up by 13% thanks to not having to look for a replacement sprite.  This is  likely why ChibiUltica actually performs better than Ultica in my tests.

So what now?  I could (1) modify the draw_terrain function to just quit for t_open_air, or I could (2) use t_open_air to replace zlevel_fog entirely and use the color_blocks fog as a fallback if no sprite is given.  I will be testing these to see which works best.  If anyone has any thoughts on this feel free to share them.

Rewryte:
Updated my previous performance test comment https://github.com/CleverRaven/Cataclysm-DDA/pull/65738#issuecomment-1556103530 with post-optimization numbers.

Qrox:
Just noticed there are other open air tiles such as [`t_open_air_rooved_outside`, `t_open_air_rooved`](https://github.com/CleverRaven/Cataclysm-DDA/blob/4ece8096c502090b4bc53ca3ecc28171e5022039/data/json/furniture_and_terrain/terrain-roofs.json#L3-L27), [`t_hole`](https://github.com/CleverRaven/Cataclysm-DDA/blob/4ece8096c502090b4bc53ca3ecc28171e5022039/data/json/furniture_and_terrain/terrain-traps.json#L3-L13), [`t_floor_brood_fake`](https://github.com/CleverRaven/Cataclysm-DDA/blob/4ece8096c502090b4bc53ca3ecc28171e5022039/data/json/furniture_and_terrain/terrain-floors-indoor.json#L1654-L1667), and [`t_skylight_frame`](https://github.com/CleverRaven/Cataclysm-DDA/blob/4ece8096c502090b4bc53ca3ecc28171e5022039/data/json/furniture_and_terrain/terrain-windows.json#L3995-L4018), and not all of them have `"looks_like": "t_open_air"` so testing a hardcoded value of `t_open_air` when the sprite is missing might not always work. I think we should rather draw the fog overlay as long as a tile is not the lowest tile drawn, and also draw the symbol (but without its original background color) in case some tile is supposed to be partially transparent, such as `t_skylight_frame`.

Also, if the lowest tile drawn is open air or an open air tile is drawn when z-levels is off, we should probably draw the tile's symbol (or an alternative opaque sprite) to avoid having a black background (or graphic artifact if it did not clear the background).

Also also, if `fov_3d_z_range` is zero we probably shouldn't draw the fog overlay in addition to calling the legacy `draw_*_below` functions.

Rewryte:
> Just noticed there are other open air tiles such as [`t_open_air_rooved_outside`, `t_open_air_rooved`](https://github.com/CleverRaven/Cataclysm-DDA/blob/4ece8096c502090b4bc53ca3ecc28171e5022039/data/json/furniture_and_terrain/terrain-roofs.json#L3-L27), [`t_hole`](https://github.com/CleverRaven/Cataclysm-DDA/blob/4ece8096c502090b4bc53ca3ecc28171e5022039/data/json/furniture_and_terrain/terrain-traps.json#L3-L13), [`t_floor_brood_fake`](https://github.com/CleverRaven/Cataclysm-DDA/blob/4ece8096c502090b4bc53ca3ecc28171e5022039/data/json/furniture_and_terrain/terrain-floors-indoor.json#L1654-L1667), and [`t_skylight_frame`](https://github.com/CleverRaven/Cataclysm-DDA/blob/4ece8096c502090b4bc53ca3ecc28171e5022039/data/json/furniture_and_terrain/terrain-windows.json#L3995-L4018), and not all of them have `"looks_like": "t_open_air"` so testing a hardcoded value of `t_open_air` when the sprite is missing might not always work. I think we should rather draw the fog overlay as long as a tile is not the lowest tile drawn, and also draw the symbol (but without its original background color) in case some tile is supposed to be partially transparent, such as `t_skylight_frame`.

You are right about the t_open_air variants, I'll see how to extend the fallback system to those.

My current idea is to let different semi-transparent terrain have their own sprites to give tilesets maximum control.  Having a universal zlevel_fog would interfere with situations such as semi-transparent water like in https://github.com/CleverRaven/Cataclysm-DDA/pull/65738#issuecomment-1560130635.  Terrain like water and t_skylight_frame won't be transparent until someone gives them an updated sprite though.

> Also, if the lowest tile drawn is open air or an open air tile is drawn when z-levels is off, we should probably draw the tile's symbol (or an alternative opaque sprite) to avoid having a black background (or graphic artifact if it did not clear the background).

The intention is to have the fog sprites show how deep the tile is, so unless the fov_3d_z_range is reduced, it should mostly be fog anyways.  I tried to use draw_terrain_below for the bottom but did not like how it looked, but if people want this, it should be an easy fix.

> Also also, if `fov_3d_z_range` is zero we probably shouldn't draw the fog overlay in addition to calling the legacy `draw_*_below` functions. I'd also be cautious about changing the order of `draw_*_below` functions because I suppose they were ordered that way to be drawn on top of non-transparent tiles such as `t_skylight_frame` in the current z-level.

Thanks, I fixed these earlier but have yet to upload the commits.  Will do it soon.

Qrox:
> My current idea is to let different semi-transparent terrain have their own sprites to give tilesets maximum control. Having a universal zlevel_fog would interfere with situations such as semi-transparent water like in https://github.com/CleverRaven/Cataclysm-DDA/pull/65738#issuecomment-1560130635. Terrain like water and t_skylight_frame won't be transparent until someone gives them an updated sprite though.

Yes, I'm aware of the need to have different sprites for different semi-transparent tiles, and the current code should work if they have sprites, but I was talking about the case when a sprite is missing. To make semi-transparent tiles draw correctly in that case, we should determine whether to draw the fog overlay by testing if the tile is not the lowest tile drawn rather than testing the `t_open_air` id. Since this may cause people to mistaken `t_skylight_frame` for `t_open_air`, I also suggested drawing the tile's symbol without its original background color on top of the fog overlay in that case.

> The intention is to have the fog sprites show how deep the tile is, so unless the fov_3d_z_range is reduced, it should mostly be fog anyways. I tried to use draw_terrain_below for the bottom but did not like how it looked, but if people want this, it should be an easy fix.

I figured `draw_terrain_below` may not look good, that's why I suggested using an alternative opaque sprite for the lowest semi-transparent tile drawn (and falling back to the tile's default symbol if a sprite is missing), so that tileset makers can use what they think best fit the aesthetics of their tilesets instead of being forced to have the tile be drawn on top of a black background. What I had in mind is to draw the lowest open air tile with a sky-blue sprite, for example. I do realize this might take more effort to implement so maybe it can be left for a later PR.

Rewryte:
> To make semi-transparent tiles draw correctly in that case, we should determine whether to draw the fog overlay by testing if the tile is not the lowest tile drawn rather than testing the `t_open_air` id. Since this may cause people to mistaken `t_skylight_frame` for `t_open_air`, I also suggested drawing the tile's symbol without its original background color on top of the fog overlay in that case.

Good idea.  Will try this.

gfx/ASCIITileset/tile_config.json
Qrox, line 1: I think tilesets are pulled from https://github.com/I-am-Erk/CDDA-Tilesets, so these changes will probably be overwritten after the next pull.

Qrox (commented):
Perhaps the 3D FOV depth option should mention the new 3D vision in its description. Also I think this might be a great chance to enable 3D FOV by default, since it has been working without noticeable issues for a while and there has been complaints that it isn't enabled by default.

gfx/ASCIITileset/tile_config.json
Rewryte: > Perhaps the 3D FOV depth option should mention the new 3D vision in its description. Also I think this might be a great chance to enable 3D FOV by default, since it has been working without noticeable issues for a while and there has been complaints that it isn't enabled by default.

Can't say I've much experience with it outside of playtesting this PR, but it does seem serviceable so far.  Should I also remove the bit where it says "Currently very bugged!"?

> I think tilesets are pulled from https://github.com/I-am-Erk/CDDA-Tilesets, so these changes will probably be overwritten after the next pull.

Thanks, I'll see if I can't post a PR there as well.

Qrox: > Can't say I've much experience with it outside of playtesting this PR, but it does seem serviceable so far. Should I also remove the bit where it says "Currently very bugged!"?

Yeah, that can be removed because it is no longer very bugged according to my (and a few other people's) experience with it.

Qrox: Also, just discovered that the automatic pull also pulls from https://github.com/pixel-32/CDDA-tileset so that needs a PR too.

Rewryte: Cuteclysm is fine off the bat.  I believe only ASCIITiles is unaccounted for, but I'm not sure if it is even built the same way as the other tilesets.

Rewryte: Yep, seems that changes to ASCIITiles are just made directly to its tile_config.json.  Marking this is as resolved.

Rewryte:
Looks like setting 3D FOV as default breaks tests.  No error message, just an indefinite freeze.  I'll take a look at the tests.  Probably will revert the option if it is not something I can fix in quick order.

Rewryte:
> This is looking good to me, does merging this without the tileset changes break the tilesets? Those as I understand it need to be made in the tileset repo / handled with the build scripts for the tilesets.

Nope.  They are for most part the same changes so these are just a temporary measure until the script built changes take over.

Fris0uman:
Would it possible to have other terrain render if the sprite is transparent? 
image
 Like here be able to see through t_gutter instead of black?

Fris0uman:
Never mind, the thing below gutter is a wall anyway

Qrox:
I think you can use connection groups to match the gutter sprite with the surrounding roof tiles instead. It might be possible to support non-open-air semi-transparent tiles by adding a flag similar to `NO_FLOOR`, but even then it will only draw the wall sprites below (as you've already mentioned) which might not match the roof sprites.

Rewryte:
> Would it possible to have other terrain render if the sprite is transparent?

Pretty much any terrain that returns false on the dont_draw_lower_floor function can be made transparent.  This is mostly the `NO_FLOOR` flag but adding more should be trivial.

PR #65741, opened 2023-05-19
Reworks to some survivor boots

anoobindisguise:
#### Summary
Balance "Rebalance survivor boots"

#### Purpose of change
Survivor boots were still using the old style of materials and coverage and needed to be modernized. They also did not reflect the survivor's needs accurately - there are plenty of boots lying around so anything the survivor makes should be specialized to deal with the post-apocalypse, namely acid puddles.

#### Describe the solution
The boots are now made of acid-resistant polymer and aramid materials. They have redone thickness, encumbrance, recipes and coverage. All the boots have rubber soles like other boots, a thin nomex liner - nomex is a fantastic material for its chemical resistance - and a waterproof/sunproof nylon outer layer. Nylon is also pretty acid resistant which will help keep the acid from seeping into the feet. As a result the boots have 5-6 environmental protection, not as good as full rubber boots (10) but much less encumbering.
Light Survivor Boots are 1.1mm of kevlar, 0.1mm of nomex and 0.3mm of nylon, with a 0.6mm rubber toe. Fitted encumbrance of 5.
Normal Survivor Boots are 2.2mm kevlar, 0.6mm nylon, 0.2mm nomex, with a 0.5mm steel toe. Fitted encumbrance of 10.
Winter Survivor Boots are 1.6mm kevlar, 2.5mm fur/fauxfur, 0.1mm nomex and 0.3mm nylon, with a 0.5mm steel toe. Fitted encumbrance of 25.

Heavy, Fire, and Wetsuit boots are untouched, but they were already planned for obsoletion.

Does this make acid zombies too weak and easy? (Certainly not any less than activity suit, right?)
Does this level of protection make sense for a survivor to stitch together? The underlying fabrics are pretty resistant to chemicals and whatnot but maybe we need a more advanced aramid sewing machine first?
Are these "too much better" than existing pre-cataclysm footwear? Most people don't need a kevlar-nomex aramid combo boot in their day to day life. These boots eschew leather, is there any particular reason a boot would be made of leather as opposed to an aramid of some sort?
Do the encumbrances seem fair? I based them roughly off existing items, though in the past we've typically done "you cannot have acid protection on feet without a shitton of encumbrance". Aside from game balance I'm not sure if that needs to be the case right?

I want to call them something other than "survivor". "Hazard boots" maybe?

#### Testing
Loaded up in game and confirmed without errors.

#### Additional context
Screenshot 2023-05-19 112516
Screenshot 2023-05-19 112606
Screenshot 2023-05-19 112640
Screenshot 2023-05-19 112745
Screenshot 2023-05-19 112804
Screenshot 2023-05-19 112833


fairyarmadillo:
Acid kind of sucks for a lot of reasons. It's a gear check that is everpresent in the world and forces players into a few pretty specific pieces of footwear. There's no real counterplay to it. You have acid shoes or you go home and sleep after every fight.

Frankly acid puddling on the floor should be the least of anyone's worries, it doesn't make sense that that's the primary concern when dealing with it - most shoes have thick soles and are at least somewhat waterproof. 

Acid should be damaging gear, causing acidosis, blinding people, sticking to the skin of your body, making fumes, etc. Acidproof clothing should look like goggles, masks, and smocks, not shoes.

randomtyper:
I'd say survivor wetsuit gear deserves some attention too, some players like to stay near water or just want to get wet buffs instead of just ignoring the debuff. Neoprene being pretty underused is a factor too. But that is likely work for another PR.

Survivor fireboots deserve to be obsoleted with this, though. Though I think IRL those survivor boots of yours would have their nylon outer layer burned off if one stayed close to fire for too long.

Speaking of naming those, how about "traveler's custom boots" or "custom survival boots"? Those are custom-made boots created for a specific purpose: ensure comfortable travel and wearer's survival. "Hazard" is a bit of a stretch imo.

Kantonine:
> Survivor fireboots deserve to be obsoleted with this, though. Though I think IRL those survivor boots of yours would have their nylon outer layer burned off if one stayed close to fire for too long.

it would make better sense to then have the nylon layer on the inside with the nomex/kevlar facing outwards

randomtyper:
Heh, kinda feels like the only thing that was seen in my comment was "not all nylon is really UV resistant". Can't really complain, though. 

If this article is to be believed https://www.leathermag.com/features/featureflammability-and-flame-retardancy-of-leather/, natural leather should resist direct fire for about 18 seconds, and even when ignited, burns (more like smoulders) slowly, unlike nylon (which burns brightly when ignited). Combined with natural water resistance and breathability of leather, that sounds like a pretty sound choice.

PR #65742, opened 2023-05-19
Make survivor zeds' drops more appropriate

anoobindisguise:
#### Summary
Balance "survivor zombies drop more and better loot"

#### Purpose of change
Survivor zombies were odd. They were described as being armored yet they almost always were just found in streetclothes and their loot was always inferior to regular zombies, you'd be lucky if you got a bottle of soda and a molotov from one. I wanted to fix that. Especially when you consider just how much *stuff* your character usually has when they die, it's really weird that survivor zombies were so stingy.

#### Describe the solution
Survivor zombies now drop a usually complete armor set.
- A jacket-type armor (torso AND arms) or torso only and arms only armor
- Protective gloves
- protective boots
- a decently protective normal layer shirt
- pants with good storage or utility (jeans, work pants, cargo pants etc)
- leg guards
- a helmet and eyewear/gasmask or a full-face helmet and no extras
These armor sets are constructed out of random parts of commonly available or low-skill-craft armor items. Damaged and filthy of course.
They also always spawn with a bugout bag, and usually extra meds, ammo, etc and a holstered pistol.

I haven't touched veteran survivor zombies yet, but I plan to if this passes muster.

#### Testing
Loaded up game, debug spawned several zurvivors, killed them, and checked out their loot.

#### Additional context
zurv1
zurv2
zurv3
zurv4
zurv5
zurv6
zurv7

### DIFF
--- a/data/json/monsterdrops/zombie_survivor.json
+++ b/data/json/monsterdrops/zombie_survivor.json
@@ -290,11 +290,13 @@
         "distribution": [
           { "group": "survivor_helmet_plus_mask", "prob": 60 },
           { "group": "survivor_basic_head_full", "prob": 40 }
+        ]
       },
       {
         "distribution": [
           { "group": "survivor_basic_with_jacket", "prob": 30 },
           { "group": "survivor_basic_individual", "prob": 70 }
+        ]
       }
     ]
   },
@@ -456,6 +458,7 @@
       [ "chaps_leather", 15 ],
       [ "chaps_cut_resistant", 10 ],
       [ "legguard_hard", 10 ]
+    ]
   },
   {
     "id": "survivor_basic_hands",
@@ -478,6 +481,7 @@
       [ "fire_gauntlets", 5 ],
       [ "gauntlet_fencing", 2 ],
       [ "gauntlet_fencing_l", 1 ]
+    ]
   },
   {
     "id": "survivor_basic_feet",
@@ -496,6 +500,7 @@
       [ "boots_fur", 10 ],
       [ "sneakers_steel", 10 ],
       [ "boots_rubber", 5 ]
+    ]
   },
   {
     "id": "survivor_basic_pants",
@@ -513,6 +518,7 @@
       [ "pants_faux_fur", 5 ],
       [ "pants_ski", 5 ],
       [ "pants", 5 ]
+    ]
   },
   {
     "id": "survivor_basic_shirt",
@@ -529,5 +535,6 @@
       [ "cloth_shirt_padded", 15 ],
       [ "combat_shirt", 5 ],
       [ "fencing_jacket", 1 ]
+    ]
   }
 ]
--- a/data/json/monsterdrops/zombie_survivor.json
+++ b/data/json/monsterdrops/zombie_survivor.json
@@ -5,28 +5,42 @@
     "id": "mon_zombie_survivor_death_drops",
     "magazine": 100,
     "entries": [
-      { "group": "clothing_outdoor_set", "damage": [ 1, 4 ] },
-      { "group": "bags", "damage": [ 1, 4 ], "prob": 10 },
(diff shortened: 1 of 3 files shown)


fairyarmadillo:
This is way too much stuff. The bags already have guns in them, and this enemy is pathetically easy.

anoobindisguise:
> This is way too much stuff. The bags already have guns in them, and this enemy is pathetically easy. Do we really want people copping 3 guns, ammo, MREs, multiple tools, medicine, supplies, and a full armor set off of one relatively common day 1 zombie?

The armor is beat to hell, filthy, and rarely actually good. You have to repair it first, if it's a kind that can be repaired, and clean it, which isn't really that big of an advantage given the amount of work you have to put in and that you can generally find replacements that are pristine elsewhere. The amount of drugs is on average less than that of a single bathroom. Food drops have literally never been a concern and also aren't here. Most of their loot isn't actually that helpful, aside from bullets, and yeah they reliably drop some bullets, but if you want a game where all the bullets in America mysteriously vanished then go play No Hope. Like it's really not that extreme.

Survivors pick up lots of loot. This isn't Fallout where you're the only one loaded down with lots of items but raiders drop like 10 bullets and maybe a single random other thing. Considering how much stuff the player accumulates this is still a pretty low amount of loot. 
FWIW survivor zombies are a midground between soldier and tough zombies as is. Yeah they're easy with a gun but what isn't.

Termineitor244:
Awesome! Good work on these spawns!

fairyarmadillo:
> if you want a game where all the bullets in America mysteriously vanished then go play No Hope. 

I don't think this combative tone is really necessary. I'm not making a bad faith argument, and if it's one you don't agree with, you can say as much without painting it that way.

I should also point out that New England has the lowest rate of gun ownership out of any region in the US, with many states at just under 15% - https://worldpopulationreview.com/state-rankings/gun-ownership-by-state

Venera3:
Don't forget to also update the monster to match their supposed armor...

anoobindisguise:
> Don't forget to also update the monster to match their supposed armor...

Yeah sure they currently have 7/7/5 defense, but I have a feeling that if I average out the overall protection of their gear it will be below that. Hard X guard is 8/6/6 for example and that's about the strongest thing they get. I'll do the math anyways tho.



> > if you want a game where all the bullets in America mysteriously vanished then go play No Hope.
> 
> I don't think this is really necessary. I'm not making a bad faith argument, and if it's one you don't agree with, you can say as much without painting it that way.
> 
> I should also point out that New England has the lowest rate of gun ownership out of any region in the US, with two states at just under 15% - https://worldpopulationreview.com/state-rankings/gun-ownership-by-state
> 
> In my current file I cannot run out of ammo even if I exclusively use guns, and no I'm not raiding military sites. Zombies will always drop bullets faster than I can fire them, and since I'm also constantly finding firearms, I never have a situation where I'm finding ammo I can't use. This does not seem correct, even in America.

I apologize if I'm being too combative here, but there are still easily several hundreds of bullets per person in civilan hands alone (estimates in the 10s or even 100s of billions of rounds in civilian ownership) and even in a state with comparatively few guns we're still far below reality. There still aren't even close to 1 gun per 7 zombies and that's under the very silly assumption that any given person only ever has 1 gun.

Daved27hundred:
I want to say a word, if these zombies are meant to be former survivors analogous to the player then it makes sense for them to be carrying a lot - what's your inventory look like when you go scavenging?

And it's not like they are a common appearance, I only encounter survivor zombies once in a blue moon.

bombasticSlacks:
As discussed on the discord a bit, this is, I think, a good quality change. Survivor zombies barely ever spawn, they are *very* dangerous in melee for basic survivors.  The armour values could potentially use tweaking however this zombie is used in the DPS tests so changing it would require a massive value overhaul to a pile of tests and, IMO, as Anid says it wouldn't be a huge change anyway.

PR #65747, opened 2023-05-20
Glassmaking audit

RenechCDDA:
#### Summary
None

#### Purpose of change
Found some things that were wildly unrealistic. Glass requires... sheet metal? I cannot find any rationale for this but it would appear to be as a working surface after pouring(molten glass). Any sufficiently strong, thermal-resistant surface would work for this.

#### Describe the solution
Make them more realistic. Added comments for the energy costs, with the calculations being in a show-your-work style.

For the tool changes the use of fire bricks is well-attested, sand-casting is as well (but not cited here), the use of sheet metal and anvils... is not specifically cited anywhere I can find but should work. Lotta ancient furnaces out there where glass would simply be cast into giant blocks and shipped out for later glassworking (which requires much less infrastructure).

Forges need my love too....

#### Testing
image

#### Additional context
Energy figures cited for glassmaking: https://www.energystar.gov/sites/default/files/buildings/tools/Glass-Guide.pdf

Glassworking(partly available online): http://oracc.museum.upenn.edu/glass/P394484  / *Glass and Glassmaking in Ancient Mesopotamia*, Oppenheim, A. Leo, Robert H. Brill, Dan Barag, and Axel Von Saldern. 1970

Specifically: `You grind (the mixture) once again. You (then) collect it within a clean dabtu-crucible, and you set down the mixture into a cold kiln chamber. You burn a good smokeless fire, and you cover it as soon as the mixture glows yellow. After (the mixture) has turned yellow, you transfer it onto the top of a fire brick. Its name is "lapis lazuli" (zukû-glass).`

and

`You (then) set down (the mixture) in a [harāgu-crucible] within the fire. You will see someone at the opening (stirring) with the tip of the rake. You apply heat to the glass-mixture within the “(slow) copper” and once the mixture takes on the appearance of mulled wine, you transfer it onto the top of a fire brick. Its name is tērsītu.`

"Forest" glass note is just adding a comment from the PR that added it. I checked out the PDF, it checks out and I don't have much more to say on that.

### DIFF
--- a/data/json/recipes/other/materials.json
+++ b/data/json/recipes/other/materials.json
@@ -980,7 +980,15 @@
       { "proficiency": "prof_inorganic_chemistry" }
     ],
     "result_mult": 4,
-    "tools": [ [ [ "crucible", -1 ], [ "crucible_clay", -1 ] ], [ [ "sheet_metal", -1 ] ], [ [ "forge", 25 ], [ "oxy_torch", 5 ] ] ],
+    "//": "This energy requirement is for charcoal forges, and should be revised when forges stop being copies of one another",
+    "//2": "For glass*making*, energy costs is about 2321MJ for ~907.2kg, or ~2.56MJ/kg. (cited figure: 2.2 MMBtu/one short ton of glass)",
+    "//3": "https://www.energystar.gov/sites/default/files/buildings/tools/Glass-Guide.pdf Optimally and not accounting for losses...",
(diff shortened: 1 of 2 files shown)


RenechCDDA:
@Light-Wave this touches innawoods and of course I only remember to @ you after I hit the button to post it

Maleclypse (approved):
Good catch. I look forward to your breaking out of the various forges!

PR #65748, opened 2023-05-20
VS Code: Preconfigured Workspace & Dev Containers

rtizzy:
#### Summary
Build "Adds VS Code Dev Containers & Workspace Config"

#### Purpose of change

- Allow users of VSCode to immediately clone, make changes, and build a copy of the source as easily as possible
- Ensure that VSCode is configured to properly format and style any changes. 
- Make it easier to contribute for less experienced individuals. 

1. Do nothing
2. Branching the `Dockerfile` off from `.devcontainer` into the root directory. This would likely make it easier for non-VSCode users to utilize the container for builds as well.  
3. Utilizing a different distro base. Ubuntu 22.04 was used as that appears to be what most of the GH Actions utilize.
4. `.vscode` has been removed from .gitignore. This is required to add recommended extensions and workspace settings The inclusion of this folder may not be desired. I could set this up as a `multi folder` workspace, which uses a separate file that exists in the root of the directory.

#### Testing

Tested on Arch Linux but instructions should apply to most distros. 

I was able to successfully build, verified that I could use `astyle` from within the editor, and confirmed that the `.editorconfig` file was respected. 

**Testing process**

1. Install VSCode & Docker
2. Clone the fork and checkout the branch `feat/devcontainer`
3. Add Cleverraven as the remote `upstream` with `git remote add upstream git@github.com:CleverRaven/Cataclysm-DDA.git` 
4. Open the folder in VSCode via the UI or by navigating to the directory in a terminal and typing `Code`
5. Install the recommended extensions from the bottom right
6. When prompted to "Reopen in container" select yes. 
7. Allow the container to build and for VSCode to Reopen 
8. Select the makefile extension on the bottom left and choose your desired configuration.
9. Press the "Play" button to build the project
10. ????
11. Profit.

#### Additional context

VSCode is my primary editor and I generally utilize dev containers when possible as it makes it easier for myself (and other devs) to be able to get an environment up and running quickly. 

I thought this would be another path for people who would like to contribute that has slightly less friction, particularly if VSCode is already installed. Configuring things with a workspace file will likely also lead to less frustration for the maintainers (Format your code!) and the contributor (How do I format my code?!


rtizzy:
This doesn't yet include cross compilation for Windows/OSX/etc. 

If there is traction on this and people are interested, I can see about adding the configs for that so people can quickly cross compile. 

This will leave us with a single container that people should be able to grab, regardless of their OS/Distro, and quickly compile a version that works on their container host.

rtizzy:
Regarding the GH Actions extension: The GH Actions extension (by default) will point towards `origin` 

The most common use case for contributions will be the creation of a fork and then a PR to upstream (here). 

In this case, the Actions are running on CleverRaven, not your fork and the GH Actions extension will show no feedback for actions. 

I've set the default `remote-name` for the Workspace to be `upstream` and manually added a remote pointing to Clever raven called `upstream` 

image

The issue:

1. Do we want this to be a manual step for the user
2. Do we want to have a way of automatically setting this up? 

I may be able to manage this with functionality in the VSCode Makefile extension but if someone has another idea, let me know.

rtizzy:
Alright, this should be closeenough.jpg for some basic testing and feedback. 

Aside from the Git thing mentioned above there are a few things that could be improved

1. Additional make configurations + Dockerfile changes for Cross Compilation.
2. Either remove -j/--jobs or utilize the preconfiguration feature to do something like `nproc - 2`
3. Basic documentation somewhere in `doc`, perhaps in `COMPILING/COMPILING.md`

.gitignore
```diff
@@ -62,7 +62,6 @@ json_formatter
 *.vcxproj.filters
 *.lastbuildstate
 .vs/
-.vscode/
```
ZhilkinSerg, line 65: What if there are some other unrelated files in that folder?

.vscode/extensions.json
```diff
@@ -0,0 +1,6 @@
+{
+    "recommendations": [
+        "ms-vscode-remote.remote-containers",
+        "ms-azuretools.vscode-docker"
+    ]
+}
```
ZhilkinSerg, line 6:
```suggestion
}

```

.vscode/settings.json
```diff
@@ -0,0 +1,32 @@
...
+                "CCACHE=1", 
+                "NATIVE=linux64", 
+                "TILES=1", 
+                "SOUND=1"
+            ]
+        }
+    ],
+    "cmake.configureOnOpen": false,
+    "github-actions.remote-name": "upstream",
+    // NOTE: This disable the plugins formatting so astyle is used.
```
ZhilkinSerg, line 30:
```suggestion
    "//": "NOTE: This disable the plugins formatting so astyle is used."
```

.gitignore
rtizzy: There are a few options:

1. I can whitelist only the currently required files (I believe that's possible with .gitignore). 
2. I can make it into a multifolder workspace which uses a separate file in the root of the directory, as mentioned in the OP.

rtizzy: I went with option one. The entire folder is ignored and we will explicitly allow that which should be checked in.

.vscode/settings.json
rtizzy: @ZhilkinSerg In VSCode JSON with comments are allowed in single line or multi-line format are supported.

https://code.visualstudio.com/docs/languages/json#_json-with-comments

> This mode is used for the VS Code configuration files such as settings.json, tasks.json, or launch.json. When in the JSON with Comments mode, you can use single line (//) as well as block comments (/* */) as used in JavaScript. The mode also accepts trailing commas, but they are discouraged and the editor will display a warning.

rtizzy:
@ZhilkinSerg Quick ping, any thing holding this up? I can pop in some docs or add multibuild support if you'd like.

rtizzy:
@kevingranade Giving you a ping on this (Sorry!) because Zhilk seems to be AFK for the past week or two. 

Are there any desired additions to this PR? Windows and OSX cross building can be included or we could get this merged and I can handle that in an additional PR.

.devcontainer/reinstall-cmake.sh
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env bash
```
alef, line 1: Did not work on Windows VS Code because 'bash\r' could not be found. Just converted the file from CRLF to LF fixed it.

snipercup:
@rtizzy Nice work on this PR. I am trying to adjust the dockerfile for cross-compilation to Windows, but I can't figure it out. Is there some way I can communicate with you about this? I'd like to be able to compile the windows build from my linux machine.

.devcontainer/reinstall-cmake.sh
rtizzy: Were you using a WSL setup or a non-WSL setup? WSL is probably the easier way IMO.

rtizzy:
@snipercup Open up a PR and ping me and I'll try and point you in the right direction. Thanks for writing up those docs!

PR #65750, opened 2023-05-20
Remove "Acclimatization" mutation type from normal Cold and Heat Tolerance (not the mutations themselves)

KittyTac:
#### Summary
Bugfixes "Remove 'Acclimatization' mutation type from normal Cold and Heat Tolerance (not the mutations themselves)"

#### Purpose of change
Some PR broke the ability to have both heat and cold tolerance. This doesn't really make much sense, heat and cold receptors are separate, and some people are just more resistant to extremes of temperature. I kept it on the post-thresh muts but I instead added a new mutation for Plant that combines both.

#### Describe the solution
Remove the mutation type flag. It doesn't bring anything when the traits aren't meant to be exclusive.

Leaving as is.

#### Testing
изображение

#### Additional context
N/A


Maleclypse:
Ok so my first thought on this is that in most animals this would tend towards a one or the other situation but that you could if there are examples of animals that have both make a post-thresh trait labeled "Expanded Acclimatization" or something similar that had both enchantments on it?

KittyTac:
> Ok so my first thought on this is that in most animals this would tend towards a one or the other situation but that you could if there are examples of animals that have both make a post-thresh trait labeled "Expanded Acclimatization" or something similar that had both enchantments on it?

That's an option. But it does mean you can't play normal human characters that can tolerate both climates. Maybe the type should only be restricted for the "High Tolerance" traits then?

anoobindisguise:
> That's an option. But it does mean you can't play normal human characters that can tolerate both climates. Maybe the type should only be restricted for the "High Tolerance" traits then?

Wouldn't them being post thresh only do that anyways?

KittyTac:
> > That's an option. But it does mean you can't play normal human characters that can tolerate both climates. Maybe the type should only be restricted for the "High Tolerance" traits then?
> 
> Wouldn't them being post thresh only do that anyways?

Plant has both.

KittyTac:
Why isn't this building?

fairyarmadillo:
> Why isn't this building?

(~[slow] ~[.],starting_items)=> All valid mutations can be purified
(~[slow] ~[.],starting_items)=>       Given: mutation of ID RESISTCHILL_RESISTWARM is valid and removable
(~[slow] ~[.],starting_items)=>        Then: a dummy mutation should cancel it or have the same type
(~[slow] ~[.],starting_items)=> -------------------------------------------------------------------------------
(~[slow] ~[.],starting_items)=> ../tests/mutation_test.cpp:538
(~[slow] ~[.],starting_items)=> ...............................................................................

The dummy mutations in the beginning of the list (human skin, human metabolism, etc) are what the game uses to purify mutations. If a mutation doesn't appear in one of these, either via its type or by being specifically called out in CANCELS, then it will throw this error.

Since you don't want it to have a type, the solution is to make HUMAN_STATS cancel this mutation.

KittyTac:
Gonna fix up this mfer as my return to CDDA PR work.

Issue #65679, opened 2023-05-16
Recipes that imply using a clay canning pot/canning pot don't work even if you fulfill all requirements


Kordian459:
### Describe the bug

I was trying to craft a 3L jug tomato sauce. 
Had all components and tools required  but the game doesn't allow me to start crafting.
Had the crafting book I needed in my inventory.
Tested with "nearby fire" and oven with full powered hotplate.
Also tested 0.5 L jug fruit jam, doesn't work.
"Byproduct: clean water" stays grey.

Tested similar recipes that use "pressure cooker" and have similar issues.

### Attach save file

N/A

### Steps to reproduce

1. get all tools needed for 3L jug tomato sauce
2. get all components needed for 3L jug tomato sauce
3. get an heating source (fire or electricity powered tools)
4. get a crafting book with 3L jug tomato sauce (like You Can Can At Home or Cucina Italiana)
5. try to craft 3L jug tomato sauce
6. game will show you that you have all requirements fulfilled (green lighted)
7. Byproduct line will stay grey
8. game won't allow you to craft this recipe
9. 

### Expected behavior

I'd expect to be able to craft my 3L tomato sauce jar

### Screenshots

![Immagine](https://github.com/CleverRaven/Cataclysm-DDA/assets/16650638/16f37285-93e5-4dc3-9611-bd9b1fb7a8f3)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.2965 (21H2)
- Game Version: a5245ca [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Mining Mod [Mining_Mod],
    Mythical Martial Arts [MMA],
    MSX Alternative Floor Tiles [UDP_Old_Floor],
    MSX Black Military Cars [UDP_BlackMilitaryCars],
    MSX Fake Snow [UDP_BN_FAKE_SNOW]
]

RenechCDDA:
No idea what is going on here but something that immediately strikes me is that the acid portion of the recipe is missing.

RenechCDDA:
Debugging hammerspace gives us an actual error message: 
[...]

PR #65761, opened 2023-05-20
Canning acid uses components and not tools

anoobindisguise:
#### Summary
Bugfixes "resolve issue that was making water bath canning recipes unstartable"

#### Purpose of change
Partially addresses #65679

#### Describe the solution
Make these components instead of tools. Does NOT fix the underlying issue.


RenechCDDA (approved):

anoobindisguise:
oh good that avg calories test failure isn't mine

PR #65762, opened 2023-05-20
Update magic_spell_effect.cpp

Kyoufi:
#### Summary
Features "Make spells hit walls instead of stopping on a tile in front of wall"

#### Purpose of change

It's just so stupid that i need to carry heavy sledgehammer everywhere just to occasionally shatter walls when my spells do damage similar to .50 BMG ammo. So i decided to make a mod that will add bash effects to spells that's supposed to have it. But there is a problem of the way spells interact with walls. If there is an impassable terrain, spell just stops on a tile in front of a wall, no matter its power. So i went even further and so here we are, with me making this pull request.

#### Describe the solution

So with adding one more iteration to few cycles spells will actually hit the walls, and if has enough power and bashing effect, shatter it.

Now im thinking about implementing the "BASH" spell flag that will add bash effect to spell, so we will no longer need to use consecutive casting for this.

#### Testing

Well i compiled game myself and all works fine for all spell shapes

### DIFF
--- a/src/magic_spell_effect.cpp
+++ b/src/magic_spell_effect.cpp
@@ -135,9 +135,13 @@ static void build_line( spell_detail::line_iterable line, const tripoint &source
                         const point &delta, const point &delta_perp, bool ( *test )( const tripoint & ),
                         std::set<tripoint> &result )
 {
+    bool sub = false;
     while( between_or_on( point_zero, delta, delta_perp, line.get() ) ) {
-        if( !test( source + line.get() ) ) {
+        if( sub == true ) {
             break;
+        }
+        if ( !test( source + line.get() ) ) {
+            sub = true;
         }
         result.emplace( source + line.get() );
         line.next();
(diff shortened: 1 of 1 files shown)


GuardianDll:
does any spell obtain an ability to bash terrain after this PR, or only one that already uses bash effect?

Kyoufi:
> does any spell obtain an ability to bash terrain after this PR, or only one that already uses bash effect?

Basically this just makes spells hit the walls to, not stopping in front of
To actually shatter walls we still do need bash spell effect

Maleclypse:
Looks like you've got a clang warning?

Kyoufi:
> Looks like you've got a clang warning?

Is it bad?

Maleclypse:
> > Looks like you've got a clang warning?
> 
> Is it bad? Will it prevent merging?

I think you would need to follow the instructions it gives 
```
use range-based for loop instead [modernize-loop-convert,-warnings-as-errors]
```

You can see what line it's in reference to on the files tab.

PR #65769, opened 2023-05-21
EOC u_sell_item() transfers items' ownership

SurFlurer:
#### Summary
Bugfixes "Make EOC u_sell_item() actually transfers the items' ownership"

#### Purpose of change
Current implementation of u_sell_item only put the item into the NPC's inventory, but the player still "owns" the item. This can be seem via debug menu.

One of the consequences is, when selling food to the refugee center's manager, if the manager's inventory is full, the item will drop onto the ground. With you still being the food's owner, you can easily pick it up and sell to the manager again.

#### Describe the solution
Transfer the item's ownership when calling u_sell_item().

#### Testing
It works, it compiles.


bombasticSlacks:
Nope usually as long as you have "basicBuildPassed" as a tag you are good to go

PR #65775, opened 2023-05-21
Gunsmith ammo convert json EOCify and cleanup

SurFlurer:
#### Summary
Balance "Gunsmith ammo convert json EOCify and cleanup"

#### Purpose of change
The gunsmith's ammo convert code is bulky and tedious.

There's also a problem that when carrying an ammo belt to the gunsmith, u_has_items will return true but u_sell_items will fail, because the game cannot extract ammo from the belt. This, sometimes can be an exploit ( giving you ammo for free ), and sometimes can be a bug ( not being able to finish the conversion, and not being able to start another, either ).

There's also a typo "ammount" in the original code.

#### Describe the solution

Utilize the shiny EOC functions and math parser, to make this part of code much much lighter.

Fix the exploit/bug via a lock.

Fix the typo while maintaining backward compatibility.

#### Testing
1. Spawn 200 `556`, and convert to 120 `76251`.
2. Spawn 800 `76251`, and convert to 1440 `545`.
3. Bring a .223 belt, and the game fails to take 556 from you, and you cannot choose what to convert to, as expected.
4. Create a 0.G save, ask the gunsmith to convert 200 `556` for me, save and quit; load it in experimental, advance 1day, and got 120 `76251`. Further conversion are still available.

#### Additional context
The variables' names are carefully crafted to provide backward compatibility.


bombasticSlacks:
Great job on this, I haven't fully reviewed it yet. Did you actually check this was backwards compatible?

bombasticSlacks:
> Could you please point me a way to migrate that u_val to use the new name? Thanks!

This will be a bombastic original, so it'll take me a few days to write it. I'll tag this PR and ping you when I find the time ☺️

bombasticSlacks:
#65810 if you need guidance on the formatting let me know

bombasticSlacks:
Thanks for doing this, really nice reorganization with some of the new features :)

Issue #55274, opened 2022-02-10
Flight and noise


chrispikula:
### Is your feature request related to a problem? Please describe.

So, quadcopters, which are robots that can hover, aren't especially quiet.  You can easily hear them buzzing when they are 10 meters above you.  They are also very light, and fragile.

Current flying robots are both weighty (5.4kg for the manhack, 62.6kg for the aftershock inactive eyebot, 31.5 for the broken mainline one), going by inactive chassis).  They are also quiet, they can surprise you when you walk around a corner.

Other large flier: (Hub 01 camspy, 30L (?)).  The rest of the bots are .75L quadcopters.

We don't have reactionless drives, even in aftershock (:-p), so moving that much thrust for that much lift is going to be noticeable.  Non-laminar air flow is noisy.

### Solution you would like.

Some flying things, eventually, should use wing-lift, but that's currently not a thing.  For stationary hovering?  That should produce excessive amounts of noise.  

### Describe alternatives you have considered.

We could just scrub almost all larger flying objects, but that seems needlessly petty. 

### Additional context

![image](https://user-images.githubusercontent.com/22281606/153504616-e69b04b8-7a99-4a39-8a18-ff9154f81a12.png)
https://hobbyhenry.com/are-drones-noisy-heres-what-the-neighbors-think/

PR #65781, opened 2023-05-21
Flight and Noise

matthew050603:
Manhacks are depicted as quadcopters but make no noise.

#### Summary
Balance "Added a parrot to manhacks so they make noise."

#### Purpose of change

Flight and Noise #55274 

#### Describe the solution

For convenience's sake, keep in mind that the following excludes the exodii drone. No edits to it were made.

drones.json is edited to add a parrot with a cooldown of 12 for all drones.

speech.json is edited to add a volume 20 sound for all manhacks, as well as a volume 25 sound for grenade, c4, and mininuke manhacks to indicate a heavy payload putting stress on the copter as well as add frequency.

I considered adding individual sounds to each manhack or making all of them sound the same, but settled on only making the ones with heavier payloads make sounds different from the rest.

#### Testing

I spawned in a LMOE shelter and debug spawned a manhack in a closed metal room. I stood outside and held 5 until the speech was confirmed, and then spawned a nuke manhack to ensure the other speech was confirmed as well.

#### Additional context

The original issue adds a source for the decibels.


matthew050603:
I changed the sound values to be a little bit closer.

fairyarmadillo:
Parrot is not the way to go here as it interrupts. We already have footstep sounds, just change manhacks to do something other than mechanical whirring.

A cooldown of 12 means these things are going to incessantly spam the shit out of players, and since they tend to spawn in large numbers in areas with multiple z levels (which parroting goes right through no problem) it will get very annoying. For comparison, prototype cyborgs have a parrot cooldown of eight hundred, and they're already sort of annoying despite appearing way less frequently.

matthew050603:
> Parrot is not the way to go here as it interrupts. We already have footstep sounds, just change manhacks to do something other than mechanical whirring. You can do this by creating a new species, such as ROBOT_FLYING, and looking at what's been done for frigate flies with INSECT_FLYING.
> 
> A cooldown of 12 means these things are going to incessantly spam the shit out of players, and since they tend to spawn in large numbers in areas with multiple z levels (which parroting goes right through no problem) it will get very annoying. For comparison, prototype cyborgs have a parrot cooldown of eight hundred, and they're already sort of annoying despite appearing way less frequently.

Will do. Thank you for the input! I'll redo it within a couple hours or so. Going through frigate flies will show me how to make new sounds, right? Mechanical whirring would work, but I _would_ like it to be unique just because of cyborgs and whatnot.

fairyarmadillo:
> > Parrot is not the way to go here as it interrupts. We already have footstep sounds, just change manhacks to do something other than mechanical whirring. You can do this by creating a new species, such as ROBOT_FLYING, and looking at what's been done for frigate flies with INSECT_FLYING.
> > A cooldown of 12 means these things are going to incessantly spam the shit out of players, and since they tend to spawn in large numbers in areas with multiple z levels (which parroting goes right through no problem) it will get very annoying. For comparison, prototype cyborgs have a parrot cooldown of eight hundred, and they're already sort of annoying despite appearing way less frequently.
> 
> Will do. Thank you for the input! I'll redo it within a couple hours or so. Going through frigate flies will show me how to make new sounds, right? Mechanical whirring would work, but I _would_ like it to be unique just because of cyborgs and whatnot.

If you look at frigate flies, they have the INSECT_FLYING species flag. If you check out that species in the species json, it will have a string for footsteps, for INSECT_FLYING I believe it's BUZZZZZ.

So what you do is make a new species called ROBOT_FLYING, copy everything from ROBOT, and change the footstep sounds from a mechanical whirring to whatever you want, then change the manhack monsters to have that species instead of ROBOT.

matthew050603:
> > Will do. Thank you for the input! I'll redo it within a couple hours or so. Going through frigate flies will show me how to make new sounds, right? Mechanical whirring would work, but I _would_ like it to be unique just because of cyborgs and whatnot.
> 
> If you look at frigate flies, they have the INSECT_FLYING species flag. If you check out that species in the species json, it will have a string for footsteps, for INSECT_FLYING I believe it's BUZZZZZ.
> 
> So what you do is make a new species called ROBOT_FLYING, copy everything from ROBOT, and change the footstep sounds from a mechanical whirring to whatever you want, then change the manhack monsters to have that species instead of ROBOT.

Sick, I appreciate the help a lot. I'll do that soon.

GuardianDll:
The probelm is that species do not specify the loudness of a movement, INSECT_FLYING uses LOUDMOVES flag to compensate it, but the only this flag do is adds additional 6 to movestep sound (which is 0 for flying creatures by default), so it would be much quieter than author made them

fairyarmadillo:
It wouldn't be too much work to add a flag that makes them even louder.

matthew050603:
> It wouldn't be too much work to add a flag that makes them even louder.

I did some experimenting with both, and I think the best option might be to do what it is now but with a slightly lower volume and more cooldown.

For the species, I edited it and tested with the frigate flies as a control and the manhacks as the actual testing group. Even with loud moves, because of the nature of footsteps, there was only 1-2 instances of the "footsteps" being heard out of about 10 minutes of game time.

When using the method I had originally used, I also noticed, and confirmed just now, that the whirring does not penetrate z-levels. With a distance of 10-12 and it not going through multiple z-levels, I don't think interruption should be too much of an issue. Additionally, I'm still tweaking it, but I think the cooldown I have it set to in my testing save works pretty well. In practice, walking around a medium sized lab room (10x9 tiles in this case) allows it to send out a parrot about every 2 laps around the outside of the wall. I think that's fair enough to allow early warning while also not being too intrusive. I'll commit it once I've figured out some BS that I ended up pushing by accident.

fairyarmadillo:
> When using the method I had originally used, I also noticed, and confirmed just now, that the whirring does not penetrate z-levels. 

Just to be sure, did you have experimental z level vision turned on in your debug options?

matthew050603:
> > When using the method I had originally used, I also noticed, and confirmed just now, that the whirring does not penetrate z-levels.
> 
> Just to be sure, did you have experimental z level vision turned on in your debug options?

I tried both with and without the experimental z-levels, and both there were no sounds, even directly above the manhacks on the next floor.

fairyarmadillo:
Well I think parrot is still the wrong solution, but it sounds like it's not the catastrophe I thought it would be, so I retract my objection.

fairyarmadillo:
@matthew050603 because there are several appropriate use cases for this (namely the huge flying insects, but also other stuff) I'm going to work on a solution. It shouldn't be too hard.

PR #65789, opened 2023-05-21
fix: wizard's apprentice achievement

casswedson:
#### Summary
None

#### Purpose of change
entering a school of magic in magiclysm gives you a default spell so you end up triggering the "learn 3 spells" after actually just learning 2 spells from different schools

change it to 5 to account for the default spells

### DIFF
--- a/src/newcharacter.cpp
+++ b/src/newcharacter.cpp
@@ -1778,7 +1778,7 @@ static std::string assemble_profession_details( const avatar &u, const input_con
     //Profession story
     assembled += "\n" + colorize( _( "Profession story:" ), COL_HEADER ) + "\n";
     if( !sorted_profs[cur_id]->can_pick().success() ) {
-        assembled += colorize( sorted_profs[cur_id]->can_pick().str(), c_red ) + "'n";
+        assembled += colorize( sorted_profs[cur_id]->can_pick().str(), c_red ) + "\n";
     }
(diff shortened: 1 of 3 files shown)


bombasticSlacks:
does changing the ID without some sort of migration break saves?

Maleclypse (approved):
Makes sense, good change!

edit:

Just saw the second file and I'm not sure what that is about.

src/newcharacter.cpp
```diff
@@ -1778,7 +1778,7 @@ static std::string assemble_profession_details( const avatar &u, const input_con
     //Profession story
     assembled += "\n" + colorize( _( "Profession story:" ), COL_HEADER ) + "\n";
     if( !sorted_profs[cur_id]->can_pick().success() ) {
-        assembled += colorize( sorted_profs[cur_id]->can_pick().str(), c_red ) + "'n";
+        assembled += colorize( sorted_profs[cur_id]->can_pick().str(), c_red ) + "\n";
```
Maleclypse, line 1781: What's this do?

dseguin: `\n` represents a line break, and `'n` does nothing (it was probably a typo)

casswedson: oh that's from another pr of mine; I mixed my commits somehow

PR #65804, opened 2023-05-22
[DinoMod] move Red

LyleSY:
#### Summary

Mods "[DinoMod] move Red"

#### Purpose of change

Bugfix

#### Describe the solution

Moves Old Guard questline from train station (which made no sense anyway) to mass graves. Necessary because train stations are getting hid by default by #65584

N/A

#### Testing

NPCs spawn no errors and behave as expected. Looks like they have the updated Center gear, which is not intended, they should have Old Guard equipment. Will fix in another PR.

#### Additional context

Thanks to @Maleclypse for catching this


Maleclypse:
Is this still supposed to be in draft?

LyleSY:
Yeah, just need a moment to test it. Should work fine though

PR #65806, opened 2023-05-22
Allow C++ standard includes in clang-tidy tests

jbytheway:
#### Summary
Infrastructure "Allow C++ standard includes in clang-tidy tests"

#### Purpose of change
All the clang-tidy tests, because they use `check_clang_tidy.py`, have been unable to use any C++ STL headers.  `check_clang_tidy.py` adds the `-nostdinc++` to prevent this.

Now that we are using our own copy of `check_clang_tidy.py` we can change this decision.  This will allow us to drop a bunch of duplicate code and ensure that our mock implementations in the tests do not diverge too much from the reality.

#### Describe the solution
Do this, via a new option `-allow-stdinc`, where tests can opt in if they wish.  Update all the tests that previously worked around this limitation to just use the regular includes instead.

This allows us to simplify `point.h`, which had previously had special treatment to allow including it without access to the C++ headers.

The main divergence I discovered that had already happened between the tests and main codebase was that the tests still used `JsonIn` in `deserialize` functions, which is no longer supported in the main code. Those tests updated accordingly.

The downside of this change is that the clang-tidy tests are now slower, so we could have kept some of the tests using mock implementations.  But they're still negligible in comparison to running clang-tidy on the rest of the code, so I'm not too worried.

#### Testing
Run clang-tidy tests.

#### Additional context


tools/clang-tidy-plugin/test/u8-path.cpp
```diff
@@ -145,9 +58,6 @@ void test_cases()
     // CHECK-MESSAGES: [[@LINE-1]]:54: warning: Construct `fs::path` by passing UTF-8 string to `fs::u8path` to ensure the correct path encoding.
     fs::path path_op_join_str_literal_path = ".." / path_u8_pchar;
     // CHECK-MESSAGES: [[@LINE-1]]:46: warning: Construct `fs::path` by passing UTF-8 string to `fs::u8path` to ensure the correct path encoding.
-    std::filesystem::operator/( "foo", "bar" );
-    // CHECK-MESSAGES: [[@LINE-1]]:33: warning: Construct `fs::path` by passing UTF-8 string to `fs::u8path` to ensure the correct path encoding.
-    // CHECK-MESSAGES: [[@LINE-2]]:40: warning: Construct `fs::path` by passing UTF-8 string to `fs::u8path` to ensure the correct path encoding.
```
Qrox, line 150: Is it removed because `std::filesystem` is not defined? If so it should be changed to `fs::` so `operator/` is still tested.

jbytheway: I tried that but there was no suitable definition of `operator/` in `fs::`.  I think we can't know reliably where it's defined, so I'm not sure we can expect to be able to call it in this way.

Qrox: It's defined here:
 https://github.com/CleverRaven/Cataclysm-DDA/blob/3d1627dab6da00a647dccd215552ce37cf16603e/src/third-party/ghc/filesystem.hpp#L628

PR #65810, opened 2023-05-23
Global and character variable migrations

bombasticSlacks:
#### Summary
None

#### Purpose of change
Something we are gonna need (like for #65775 )

#### Describe the solution
Added a new json entry for var_migration, similar to others uses to: and from:
checks on load for entries in the global and character vars to migrate, migrates their keys.

#### Testing
Tested it in game, I personally used these values lol:
```json
{
    "type": "var_migration",
    "from": "temp_var",
    "to": "hemp_var"
}
```

### DIFF
--- a/src/global_vars.h
+++ b/src/global_vars.h
@@ -47,7 +47,7 @@ class global_variables
         void serialize( JsonOut &jsout ) const;
 
         std::map<std::string, std::string> migrations; // NOLINT(cata-serialize)
-        static void load_migrations( const JsonObject &jo, const std::string_view src );
+        static void load_migrations( const JsonObject &jo, const std::string_view &src );
 
     private:
         std::unordered_map<std::string, std::string> global_values;
--- a/src/global_vars.h
+++ b/src/global_vars.h
@@ -45,6 +45,10 @@ class global_variables
         }
         void unserialize( JsonObject &jo );
         void serialize( JsonOut &jsout ) const;
+
+        std::map<std::string, std::string> migrations;
+        static void load_migrations( const JsonObject &jo, const std::string &src );
+
     private:
         std::unordered_map<std::string, std::string> global_values;
 };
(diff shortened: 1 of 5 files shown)


src/global_vars.h
```diff
@@ -45,6 +45,10 @@ class global_variables
         }
         void unserialize( JsonObject &jo );
         void serialize( JsonOut &jsout ) const;
+
+        std::map<std::string, std::string> migrations;
+        static void load_migrations( const JsonObject &jo, const std::string &src );
```
bombasticSlacks, line 50:
```suggestion
        static void load_migrations( const JsonObject &jo, const std::string_view src );
```

src/savegame.cpp
```diff
@@ -1484,6 +1492,15 @@ void global_variables::serialize( JsonOut &jsout ) const
     jsout.member( "global_vals", global_values );
 }
 
+void global_variables::load_migrations( const JsonObject &jo, const std::string &src )
```
bombasticSlacks, line 1495:
```suggestion
void global_variables::load_migrations( const JsonObject &jo, const std::string_view & )
```

src/global_vars.h
```diff
@@ -45,6 +45,10 @@ class global_variables
         }
         void unserialize( JsonObject &jo );
         void serialize( JsonOut &jsout ) const;
+
+        std::map<std::string, std::string> migrations;
```
bombasticSlacks, line 49:
```suggestion
        std::map<std::string, std::string> migrations; // NOLINT(cata-serialize)
```

src/global_vars.h
```diff
@@ -45,6 +45,10 @@ class global_variables
         }
         void unserialize( JsonObject &jo );
         void serialize( JsonOut &jsout ) const;
+
+        std::map<std::string, std::string> migrations; // NOLINT(cata-serialize)
+        static void load_migrations( const JsonObject &jo, const std::string_view src );
```
bombasticSlacks, line 50:
```suggestion
        static void load_migrations( const JsonObject &jo, const std::string_view &src );
```

PR #65818, opened 2023-05-23
lower fur bodyarmor encumbrance a bit

anoobindisguise:
#### Summary
Balance "fur body armor has slightly lower encumbrance"

#### Purpose of change
Fur body armor had extremely high encumbrance, and was especially bad if you stored anything in it (4 encumbrance for 1 liter of storage). Fur armor is itself a little ahistoric but we don't need to worry about that - it's not an issue if it's bad armor, but it's not an actual historical artifact, simply being a thrown-together primitive form of protection for wilderness survivors. That said it was probably too shitty.

#### Describe the solution
Give it the same encumbrance of leather armor, as it's about the same weight (but lower coverage). Make encumbrance not increase with storage, as there are many existing armors that also work this way (kevlar jumpsuits, nomad jumpsuit)

Fur is again, not exactly amazing armor so this armor existing is a little dubious. I don't think this is too dramatic of a change though and the armor as it was bad enough that you basically wouldn't ever bother, even in the one place it was "meant" to be used (no access to civilization armors)

### DIFF
--- a/data/json/items/armor/suits_protection.json
+++ b/data/json/items/armor/suits_protection.json
@@ -126,9 +126,9 @@
     "looks_like": "armor_larmor",
     "color": "brown",
     "armor": [
-      { "covers": [ "torso" ], "coverage": 95, "encumbrance": [ 21, 25 ] },
-      { "covers": [ "leg_l", "leg_r" ], "coverage": 85, "encumbrance": [ 21, 21 ] },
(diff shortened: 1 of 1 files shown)


fairyarmadillo:
At some point I think we could see these recipes redone as fur-lined or fur-trimmed versions of existing armor. Now that cloaks are strapped and can fit over armor, it'd probably be reasonable to make fur-lined armor way less furry than the current stuff and adjust the encumbrance to match.

PR #35500, opened 2019-11-13
Spawn loudspeaker at the roadblocks

Night-Pryanik:
#### Summary
```SUMMARY: Content "Spawn loudspeaker at the roadblocks."```

#### Purpose of change
Address some issues concerning insufficient warnings leading to cheap death by turrets.

#### Describe the solution
- Added a `loudspeaker` "monster". For now it will spawn only at the roadblocks and will continually shout a set of warnings which should warn player about serious threat up ahead.
- Warnings are written as snippets, so they are easily could be added or edited.
- Warnings are chosen randomly from the list.
- Loudness for warning sound is 60, so it should be heard from the border of reality bubble.
- Cooldown for another warning is 10 seconds.
- Loudspeaker will spawn at the center of the roadblock.
- Loudspeaker will drop some low-importance loot on its death.

#### Describe alternatives you've considered
- Use `parrot` monster attack function instead of creating a new one. But this function for some reason make random-duration delays between warnings while I wanted fixed ones.
- Use existing `speaker cabinet` furniture instead of a new monster. But that would require a lot of new code to make furniture shout warnings.

#### Testing
Debug-spawned roadblocks, walked around a bit, listened for warnings.

#### Additional context
None.

PR #65831, opened 2023-05-24
Rebalance and add variance to police/army road blocks

stubkan:
#### Summary
Balance "Rebalance and add variance to police/army road blocks"

#### Purpose of change

Road blocks have been pretty static for a long time, they have consistent structure and guaranteed loot.  Also they are quite easy to cheese for freebies.  Especially the military ones.

#### Describe the solution

This adds a new dual state for road blocks - they are now either 'live' roadblocks with live turrets and dead zombies or 'overrun' roadblocks with wrecked turrets and roaming zombies.

This makes the "dangerous" military roadblocks much more dangerous now, as the 'less than lethal' sponge turrets are replaced with m249 turrets instead.  Also the turrets and zombies will not be around at the same time to 'occupy' each other, making them easier to pick off anymore.

This also helps verisimiltude, as during the 2 months since the cataclysm occured - one of two things would have inevitably occured - the zombies would have killed the turrets, or the turrets would have killed the zombies.  It doesn't make sense to have that showdown occur at the precise moment you decide to turn up.

A few other small tweaks are also in the bag, to better balance the whole thing, and to retain verisimilitude as well as giving the player more varied and interesting results. 

* M249 CROWS turrets no longer have 1600 rounds of ammunition, but instead have 160.
* The riot control platform has 3 extra sponges, so it is slightly less easy to take out.
* The military roadblock has a new entity - a military loudspeaker that yells at you to stay away or get shot.
* The military loudspeaker is neutral so zombies do not destroy it.  If you capture it, it will revert to a normal loudspeaker.
* 'overrun' road blocks will have a random amount of zombies attracted to it as well as the normal soldier/cop zombies.
* The overrun military roadblock has more zombies in it, due to the loudspeaker attracting them.
* It is now possible to have much fewer army/cop zombies - as the loot available was quite high.
* The police roadblock has a chance to have no cars, one car, or two cars.
* The above 2 changes (fewer responders and fewer cars) are to reflect diminished manpower/resources in the face of the cataclysm.  It also makes the encounters more unpredictable and varied.

#### Testing

Cranked barricade chance way up and visited so many barricades to confirm changes were working as intended.

Live Police Barricade - with no car and one dead policeman, semi rare occurence 

image

Live Military Barricade - all zombies dead, but the hardware lives on - the loudspeaker broadcasting a warning

image

Overrun Police Barricade - the zombies long killed the turrets and the barricade 'picked up' two roamers

image

Overrun Military Barricade - this screenshot is from before I added picking up zombie roamers, there would be maybe a half dozen zombies added to those soldier ones, thanks to the loudspeaker making noise.

image

#### Additional context

I may do further edits ;

* the dead zombies will ressurect if they are not smashed - however they will ressurect as plain zombies, not soldiers/cops...  Perhaps I will figure out how to make them appear as dead soldiers/cops and ressurect that way.
* Also may add more flavour (regular dead zombies on the outside of the live barriers).
* I also want to remove spent casings from the tile that contains destroyed turrets, so the turrets are more visible.

I am open to feedback and ideas regarding the balances.


Night-Pryanik:
The "loudspeaker at roadblock" idea was rejected a long time ago - #35500.

stubkan:
> The "loudspeaker at roadblock" idea was rejected a long time ago - #35500.

Since the M249 CROWS doesn't have a warning message function like the riot platforms do, I put the loudspeaker in and copied it from already existing loudspeaker item - so it is a military roadblock only item.  So it technically functions the same as the original military barricade did by riot platforms warning people with a speaker.

Those arguments in #35500 they were giving against loudspeaker being that the barricades didnt want to broadcast loud messages to attract zombies - why does the riot platform have loud messages then?

Personally, I think, in the beginning of the cataclysm people were clueless as to what was going on - and setting up these roadblocks wouldn't have worked - so likewise, I doubt anyone setting up a roadblock back then would have had the foreknowledge to not make noise.  People setting up roadblocks did so to protect other people, and they would have wanted to warn people, so they didn't get shot unnecessarily, more so than being quiet.

I think in this case, it might be good as it makes the overrun military barricade interesting and dangerous - and that is explained away by the loudspeaker.  I'll get a screenshot of that tomorrow.

bombasticSlacks:
We just got through removing random gun turrets from the overworld. The government isn't putting automated turrets at roadblocks to shoot civilians lol. 

Anyways there is no reality where a turret beats the zombies, they get up tomorrow and it gets closer to having no ammo. 

Turrets have ammo based on reality, arbitrary changes are bad. 

There might be other issues but those immediately stick out.

anoobindisguise:
So the issue with the CROWS II roadblocks was that the military outfits these turrets with 1600 rounds of ammunition, not 160, and if there's a remote roadblock that hasn't started spending ammo on zombies yet (otherwise the turrets will spend all their bullets in an infinite cycle of shoot the zombies -> zombies revive) and run out of ammo (so "ammo-available") roadblocks should become rarer later into the cataclysm.
Balance wise turrets are substantially harder to cheese, now that they shoot at moving vehicles and you can't just crouch behind the dashboard. That said there was a reason outside of "too many easy bullets" that the turrets got removed from these locations so they may just be fundamentally unworkable, and should also definitely have specialized behavior instead of "it was foggy and I stumbled into its line of fire immediately and died before I realized"
Splitting them into either broken turrets, zombies, or no zombies, active turrets is probably a good idea. 
If anything the available loot could increase, if police officers brought some of their ginormous armories here to stockpile a defensive position (even if they dispersed and went elsewhere, they brought it here by car, and if they leave on foot they're not carrying it all). Then also make them just rarer in general because they're currently very common.

PatrikLundell:
Having "neutral" loudspeakers doesn't make sense. Noise makers attract zombies that try to smash whatever makes the noise (unless it's actually a zombie as well). This also means that any overrun location that had noise makers should spawn with smashed noise makers.

Nebnis:
Having these roadblocks with mob density like the corpse piles extra might tonne down the balance of them being easy to loot since they would really eventually reach a point where a lot of people would be there,both cops/military and civilians. I agree with making them rarer, they spawn every other road.

fairyarmadillo:
The riot turrets never made sense to me and I don't think anything would be lost if they were removed, especially since we could give their attack to a feral riot cop.

Something that would be nice but maybe outside of scope here would be putting protest/riot debris in the area. Used medical kits, food wrappers, handmade signs, flags, flyers, etc. Even peaceful protests tend to leave kind of a mess, let alone all-out riots.

Maleclypse (changes requested):
So the current trend is to move hardcoded map extras to json rather than adding new things to the hardcoding.  I believe that will keep this as it is currently written from getting merged.  Let me know once you've moved it over to json and I'll rereview

stubkan:
> So the current trend is to move hardcoded map extras to json rather than adding new things to the hardcoding. I believe that will keep this as it is currently written from getting merged. Let me know once you've moved it over to json and I'll rereview

How do you propose that is done?  As there would be 15 possible permutations (road directions) for each facing, times by 4 variants (police, military, live, overrun) so just for the maps required, that'd require about 60 json map files.  I think for the other context free map extras a json transfer is easy, but this special requires overmap context in order to know how to place itself - is the json functionality there?

Maleclypse:
> > So the current trend is to move hardcoded map extras to json rather than adding new things to the hardcoding. I believe that will keep this as it is currently written from getting merged. Let me know once you've moved it over to json and I'll rereview
> 
> How do you propose that is done? As there would be 15 possible permutations (road directions) for each facing, times by 4 variants (police, military, live, overrun) so just for the maps required, that'd require about 60 json map files. I think for the other context free map extras a json transfer is easy, but this special requires overmap context in order to know how to place itself - is the json functionality there?

I believe you started before #62407 was merged so that's what I'd look at.

stubkan:
> > > So the current trend is to move hardcoded map extras to json rather than adding new things to the hardcoding. I believe that will keep this as it is currently written from getting merged. Let me know once you've moved it over to json and I'll rereview
> > 
> > 
> > How do you propose that is done? As there would be 15 possible permutations (road directions) for each facing, times by 4 variants (police, military, live, overrun) so just for the maps required, that'd require about 60 json map files. I think for the other context free map extras a json transfer is easy, but this special requires overmap context in order to know how to place itself - is the json functionality there?
> 
> I believe you started before #62407 was merged so that's what I'd look at.

Oh man, thank you for linking that.  I should update my master branch it seems

Zireael07:
Definitely worth reviving imho

stubkan:
I quit this because the whole thing was moved to predefined json map files instead of being dynamically generated by code, so it would be pointless to implement my method.

Also, it looks like there is only one variation of roadblock now instead of 15??  unhardcoding it looks like that came with a big step backward

PR #46615, opened 2021-01-08
fix(npc, reading): fix NPC stops reading when player changes z-level

Insineer:
#### Summary

SUMMARY: Bugfixes "Fix NPC stops reading when the player changes z-level"

#### Purpose of change

Fix #43982

#### Describe the solution

Light caches work only for the player's submap and z-level, so we can't use them for NPC which isn't located on the same submap and z-layer as a player.

#### Testing

1. Spawn NPC and make friends with him (complete a quest, for example). Say him follow you.
2. Ask NPC to read a book.
3. Move to another z-level with stairs.
4. Move back to NPC and check that he is reading and not following you.

PR #65835, opened 2023-05-24
Fix stop of NPC disassembly job while on other Z-lvl (fix #52008)

### LINKED ISSUE #52008: NPCs stop disassembly work if player character changes Z-level
### Describe the bug

So, I've started to get enough NPCs at my new faction that I can spare a few for non-expansion-related things, and I decided to try out assigning them to jobs around the camp - especially the disassembly work I had suggested adding in a while back (thank you, by the way!). Unfortunately, I soon found out that the NPC immediately "completes" their disassembly work if I change Z-levels, despite not being finished with disassembling the items in the zone.

### Steps To Reproduce

1. Designate a disassembly work zone with loot to disassemble in it
2. Tell an NPC follower to disassemble the items
3. Walk up or down a ramp or set of stairs
4. Look at the message log, and/or return to the previous Z-level, and observe that the NPC supposedly "completed" their task without actually doing so

### Expected behavior

I expect NPCs assigned to do disassembly work will continue doing it regardless of whether or not I'm on the same Z-level.

### Screenshots

_No response_

OS: Windows
OS Version: Windows 8.1 Pro
Game Version: ccf5e15 [64-bit]
Graphics Version: Tiles
Game Language: System language []
Mods loaded: [
Dark Days Ahead [dda],
Aftershock [aftershock],
C.R.I.T Expansion Mod [crt_expansion],
Magiclysm [magiclysm],
XEAS [XEAS],
Gun Cotton [guncotton],
DinoMod [DinoMod],
Youkai Enemies [Youkai_Disco],
Mining Mod [Mining_Mod],
Mole Mutation [dig_a_tunna],
More Survivor Stuff [more_survivor_stuff],
Mythical Martial Arts [MMA],
More City Locations [cityside],
Compatible Nimian Manufacture Guns [NM_wpns],
Fuji's Military Profession Pack [fuji_mpp],
Magiclysm No Class Limit [magiclysm_no_class_limit],
Mutation Changes [Mutation_Changes],
Portable Autodoc [portable_autodoc],
SpeedyDex [speedydex],
Useful Helicopters [useful_helicopters],
Arcana and Magic Items [Arcana],
return_the_nut [return_the_nut],
Hydroponics [hydroponics],
Cooler Versatility [coolervers],
Chris K's Knives [ChrisKsKnives],
Custom Spells [Custom_Spells],
Reproduction mod [perverts],
Custom Bionics [custom_bionics],
Magiclysm Runes in Arcana Recipes [Custom_ArcanaRune],
Deconstruction Plus [Deconstruction_plus],
Core Data Changes [core_data_changes],
Arcadia Mapgen [custom_arcadia_mapgen]

### Additional context

_No response_

Tread4:
#### Summary
Bugfixes "Fix stop of NPC disassembly and crafting jobs while player changes Z-level"

#### Purpose of change
Fixes #52008

Resolves the issue of NPC companions stopping their disassembly activity when player changes Z-level. It most likely will resolve similar issue for crafting activities performed by NPCs

!!! See my post below for newest description of the solution - this one will stay to keep PR history intact !!!

#### Describe the solution
The underlying issue is the fact that no light level calculations are performed for Z-levels other than one player is on. NPCs performing an activity (or trying to start next one scheduled [e.g. next item disassembly]) were cancelling due to not meeting light level requirement (or craft speed requirement derived from light level).

Solution adds a "is_avatar()" check to those light level requirements to make them valid only for PC. NPCs are excluded and therefore can continue their assigned activity on different Z-level.

Solution has one major drawback: NPCs can disassemble in the dark - I am currently looking  for a better way to solve the issue and would gladly accept help/suggestions. Please take a look into alternatives I consider

Solution was based on comment found in #52008. It suggested that similar fix as #46615 might be needed in this case. I investigated and came to the conclusion that it indeed might work. Furthermore, a fact that similar solution was already merged for reading activity made it more probable as a valid approach.

Alternatives that solve Z-lvl issue but still forbid NPCs to craft/disassemble in the dark:
1) PREFERRED: Determine if workplace/NPC location is counted "inbounds" the same way terrain is checked for light calculations. If not use the override (to allow crafting while on different z-lvl). If place is "inbounds" it will follow all regular rules and not allow crafting in the dark. Additional advantage is that start of NPC activity has to be when light is sufficient. Once player changes z-lvl and light calculations cease for the workplace it is assumed that light is still present there (it was at the activity start).

I am currently looking into lightmap.cpp and character.cpp (fine_detail_vision_mod function) to somehow make it work. I think I keep failing at proper inbounds implementation.

2) Adding some check if the terrain has current light calculation and use override from original solution only when flag is absent. Result would be same as option 1) but it seems to be less efficient and elegant.

#### Testing
!!! I acknowledge that current solution might be unacceptable due to NPCs crafting in the dark - I still make this PR to give this issue a push and possibly receive a proper advice from someone more experienced with our C++ code !!!
I am fairly certain that it can be reasonably solved if I mange to properly check if NPC is in the area with calculated light or not.

- Compiled game with my code changes, 
- Started a new world and character
- Recruited starting NPC
- Spawned items to disassemble (smartphones) and necessary tools
- Set up proper tool and disassembly zones in unlit basement bathroom (starting shelter)
- Confirm there is no light
- Set up bare-bones basecamp in the basement, built the bulletin board
- Tried disassembly with PC (lack of light) and confirm it is impossible

1) Manual task assignment
- Asked NPC to disassemble items
- Confirmed NPC starts and continues with the job
- Went upstairs  and an overmap tile or two out of the shelter
- Waited 30 min
- Went back downstairs and confirmed that NPC is still busy and disassembled several items

2) Basecamp task
- Tell NPC to stop activity and follow me
- Set up basecamp zone (storage)
- Assigned NPC to basecamp as worker
- Set up Disassembly job priority for NPC (as highest)
- Observe NPC starting and making progress with the task
- Went upstairs, out of shelter, wait
- Went back and confirm NPC is still busy and disassembled several items

3) Notes
- Teleporting a few overmap tiles further stops the activity (separate function in character.cpp does it and I assume it is intended).
- Tests were performed with light on tiles as well; results as expected were the same and activity was performed
- Disassembly by PC in the presence of light was available as expected


Tread4:
Quick status update, without new code, as I am currently not near my private PC. I will update the PR with new code this night or tomorrow. Below you can find the description of new solution (possibly lacking a few details since I do not have code to look at atm.)

I searched for another solution that would solve the bug, while not introducing a new one (NPCs being able to craft/disassemble in the dark). I am quite certain that I found a better solution and testing so far proves promising. 

**Describe the solution**
- I rolled back the code changes from initial PR
- Adjusted "fine_detail_vision_mod()" function call in crafting.cpp to use NPC location as argument
- Added a check: "player z level == npc z level ? calculate NPC light : use arbitrary override value"

This allows for proper lighting checks when assigning the action to NPC (in practically all cases, I can imagine, player is on the same level). It should solve the new, potential, bug of NPCs crafting in the dark. 

At the same time using the arbitrary value once the player is on different Z-level prevent NPCs from cancelling action when player changes the Z-level. I think it is acceptable to assume that if NPCs started action when light was available it is available further (same as we do with reading).

**Describe alternatives you've considered**
The "best", or at least most realistic solution would be to calculate lighting in immediate NPCs area (therefore create a small, additional lightmap around them) if they perform action requiring it. It would however call for additional lighting calculations (with many NPCs potential performance hit with) and introduce new dilemes in-line of "NPC should be able to get new battery for flashlight, feed the fire etc.", all while being out of view. It also would be way more complex solution and most likely outside scope of my abilities.

**Testing**
See : Testing section in the first post - tests were following the same set up

Results:
- Task was assigned and delivered same result both for direct assignment and basecamp assignment
- NPC cannot perform task in the dark and immediately cancels it  (player on same Z-level)
- Above point is valid for both newly assigned tasks and for ongoing ones
- NPC properly works on the assigned task when light level meets the craft requirements (player on same Z-level)
- NPC continues to work on the task and makes expected progress once player is on another Z-level
- !!!   NPC cancels the tasks the very moment player gets on the same Z-level   !!!

The last point seems as obvious undesired behaviour. It is only called for if lighting conditions changed and are no longer met, however the same happens even if NPC is having sufficient light (atomic lamp in this case) all the time.

Most likely the culprit is the fact that in a first tick (or first few ones) the lighting map for z-level (on which NPC is, and which was just entered by PC) is not calculated yet but NPC action is processed already. It could be that NPC actions are processed first and then lightmap is generated (same turn). Either way, since player is already on the same level, the game expects lighting to be sufficient to continue NPC task and checks it against empty/default value lighting map. Returned result is "darkness" and NPC breaks the craft.

I am having a few ideas (although not sure I will manage to implement them) on how to work around or otherwise solve this edge-case issue. Even if it remains unresolved, I am fairly certain that results described above are a worthwhile improvement of the situation.

Two (in my opinion most promising) ideas to solve this "one turn lack of light" edge-case:
- add an int parameter to the PC that stores z-level it was on in the last turn; stop using NPC override only if PC is on the same level for 2+ turns (assuming 1 turn or less is the length of time without lightmap on z-lvl)
- add a bool or a table to NPC to store whether lighting level on the last (or last few) turns was sufficient to craft; break craft only if it was insufficient x number of turns (I do not find it a problem if NPC breaks craft 5 sec after light goes out but it might solve the issue I described)

PatrikLundell:
While it might be considered reasonable to allow a companion to finish a current task while in an unknown lighting situation (which can be questioned), I don't think it's reasonable to allow them to take on follow up tasks when dealing with long tasks (such as mining a tile and then switching to another tile upon completion). However, that would kill off sorting as soon as the PC leaves the level. Thus, I think the lighting calculation problem would have to be solved for this to work properly in an acceptable manner.

Tread4:
I fully understand the reasoning and in principle agree. It would be best if we can always check lighting conditions. System would benefit from allowing NPCs to influence lighting situation with available tools (moving/carrying lamps, flashlights etc., turning them on/off, changing batteries). Otherwise we run into risk of NPCs breaking the action due to stepping out of the lighted area or running out of battery while having required resources at hand.

Similar issue would regard natural lighting - it would be useful if NPC "pauses" the action once lighting is insufficient and resumes it once it is at correct level again. Similar as solar panels were tackled. Otherwise assigning NPCs any long tasks is moot since they will stop the first evening / dead battery.

I am however unable to code such a systems and can see a risk  of performance hit.

**This PR aims to influence only the disassembly and crafting actions by NPC (no mining, construction, sorting etc. activities).**

The logic behind this workaround is this:
- I have a perfect "crafter" friend, who is terrible in the field
- we agree that I make a 3-day excursion for resources, while he disassembles loot for components we need (staying in safe-house) 
- [I know, he would stop due to other checks if I go too far but we can assume other situation: stay in our basement base and disassemble things, while I spend three days farming above"]
- in reality I expect him to make reasonable progress once I am back 3 days later 
- currently, he will disassemble one or no items during those 3 days
- NPCs as an idea simulate living, thinking people and those per se will look for a way to mitigate/adjust to lack of lighting (use lamps and batteries, change activity place, start a fire etc.).

I doubt that current situation is any more realistic than lighting check override. I know that override is "gamey" solution and fully agree that proper system would be better. On the other hand it seems much more realistic to me that NPC (person) would find some way to continue with a task than "oh, he went upstairs - I'll take a nap till he is back in 3 days" ;)
Furthermore, it would simply make NPCs more useful and "working as intended". I am certain that multiple players keep running into issue #52008 and either not knowing why activity breaks or being disappointed that "it should work, but it doesn't" (as I did). 

I am totally open to making such override that would influence (negatively) crafting speed for "overridden" NPCs. It might simulate "downtime" of natural light or activities NPC performs to get proper light. It would also "penalise" the player who tries to "game" the system too much. Of course the downside is that it would penalise the players with good "base setup" (atomic lamps, lamps connected to recharging grid/vehicle etc.). It is a minor issue nevertheless.

I know that approach described is not ideal, although personally I think it is an acceptable workaround of the current issue. If someone is able and willing to assist me with devising a better and more realistic system I am open to it. Nonetheless, I am sure it would turn into me assisting this someone since my coding skills are rather limited.

PatrikLundell:
The current system has companions cease their activities on the next check when the light level is too low. This causes them to break off their activities and claim to be done. The point at which the "next check" occurs differs between tasks, so some are broken off with partial progress (such as construction), while others are broken off when the companion should move on to the next step (such as having sorted one item moving on to the next one, or vehicle disassembly when one part has been removed, before moving on to the next one).
It can be noted that some activities support PC departure and later return by "fast forwarding" what would have happened in the during that time (such as vehicle disassembly), while others don't (e.g. construction: I don't remember if the companion moves on to the next construction at the time of return or claims to be finished, but I would expect the latter to usually be the case anyway, due to the reality bubble usually not covering everything needed).

Tread4:
I do understand that it is how the current system works. Therefore I am trying to tackle a small chunk of the bigger issue and allow disassembly task that started (and in plenty of situations is performed) in the proper lighting conditions, just at different Z-level.

I am trying not to change the "disassembly <-> light lvl" in general, just tackle the z-level problem. Furthermore, the issue exist for years 

**I also do not delve into reality bubble check - it is a separate one and will still pause those activities.**

At the moment, from player perspective, it is not the lack of light that breaks those actions but a PC z-level change.

It is true that lighting calculations are an underlying issue, but I see no feasible and performance friendly way of tackling it. The only  "true" solution coming to my mind is constantly calculating lightmaps for every submap on which some NPC is performing/starting an activity. Simulating a few tiles is usually not enough (and I am still unable to code it) since light propagates and can originate from a point much further (or from z-level above).

The override I propose is a kind of a crutch. I hope that it will be "good enough for now" since it brings improvement (at least in my opinion). **If it is not acceptable, I hope that prodding this issue will motivate someone more knowledgeable to pick up and devise the realistic and acceptable system (and I will be happy to assist to the level of my abilities).**

On conceptual level it is not hard to describe "what would be perfect solution". It is however hard to implement such system and not risk performance. I am straightforward about not feeling up to such task.

Tread4:
Promised code update - took a bit longer since I decided to tackled the "stop action as soon as player returns to my z-level" issue I described above.

Current solution is based on description I posted on 25th May. It includes a fix to the "edge-case" I described there as well. To fix it I took the first approach, from the ones I described at the end of that post.

**Describe the solution**
- I rolled back the code changes from initial PR
- Adjusted "fine_detail_vision_mod()" function call in crafting.cpp to use NPC location as argument
- Added a check: "player z level == npc z level ? calculate NPC light : use arbitrary override value"
- Added a character variable that saves its z-level from last turn (and update it every turn)
- Added a check: "player z level == player z level a turn ago" - solves edge-case when on first turn after getting back to NPCs Z-level the task would stop (imo. NPC action processed before lightmap generates).

**Testing**
See : Testing section in the first post - tests were following the same set up

Results:

- Task was assigned and delivered same result both for direct assignment and basecamp assignment
- NPC cannot perform task in the dark and immediately cancels it (player on same Z-level)
- Above point is valid for both newly assigned tasks and for ongoing ones
- NPC properly works on the assigned task when light level meets the craft requirements (player on same Z-level)
- NPC continues to work on the task and makes expected progress once player is on another Z-level
- NPC no longer cancels the tasks the very moment player gets on the same Z-level (extra var and check solved it)
- Both proper going up and down the stairs and debug teleporting PC give the same results
- NPC pauses and "stores" the assignment once out of reality bubble - it properly resumes once player returns (other chunk of code regulates that  and I treat this behaviour "Working as intended"

**Request for review**
I would be grateful if someone much more knowledgable (than me) in Cata codebase and C++ can review those changes and when needed propose changes. It would be tremendous help if any code change requests are paired with either instruction or good pointers to where I can find similar code. My C++ skills are extremely rusty and I just started looking into Cata C++ code. Without a push in the right direction I might be unable to implement  potential requests.

PR #65846, opened 2023-05-25
Base repair difficulty on material instead of recipe

anoobindisguise:
#### Summary
Balance "difficulty to repair depends on what the thing is made of instead of its crafting difficulty"

#### Purpose of change
The current repair system had some negative aspects to it. First of all certain items were much more difficult to repair for no real good reason - for example it was much easier to stitch a tshirt back together than it was to stitch a sweatshirt back together despite them both just being cotton. In addition it created an endless headache of balancing "repairs_like" fields in various items to ensure they were properly repairable because if no recipe was found it was automatically difficulty 10. Numerous items in game became much harder to repair because recipes they repaired like got obsoleted, or just because they didn't have a repairs_like to begin with. Also just in general, repairing something is substantially easier than making a new one.

#### Describe the solution
Repair difficulty now depends on the material. Materials have repair_difficulty defined in materials.json.
First check if an item has "repairs_like". If it does, use the materials of the item it repairs_like. Otherwise, find the most difficult material the thing is made of, and use that to determine difficulty.

This will probably require some pruning of the "repairs_like" field. For example the work tshirt is canvas but repairs like a regular tshirt and so it's easier than normal (canvas is 2, but cotton is 1).

Also, it should probably also be applied to some items to make them more difficult to repair, such as a hazmat suit currently repairing like a survivor firesuit (which gives it difficulty of 4, it should probably be higher than that). Or hazmat suits could be made of something other than common plastic 

Should it be possible to set a direct repair_difficulty override in the item itself?

Of course any of these values are easy to change in JSON. If Kevlar should be 5 and not 4 that's a simple fix.

I originally wanted knowing a recipe to make a thing to give -1 to difficulty (or not knowing a recipe to give you +1) but it causes issues such as knowing how to make a sleeveless duster (trivial) making it easier to repair a sleeveless duster. I don't think any advantages it adds would be worth the problems it would introduce.

#### Testing
Loaded up in game and confirmed it seems to be working correctly. 

#### Additional context
image


PatrikLundell:
I think the argument made concerning the patching of a T-shirt vs a sweater applies to recipe knowledge as well: repair is rather different from construction from scratch, so recipe knowledge would only be useful for a rather small number of exceptions.
I do think it would be useful to allow for either overriding the repair difficulty or to modify it for a small number of items. The hazmat suit has to not just be patched up so the hole is covered/rent stitched together, but the result also has to be impervious to air and water, which adds to the difficulty.

Maleclypse (commented):
So a thing I noticed is that you didn't set a repair difficulty for Aftershock and Xedra materials.  Do they not need it because it defaults to something or do they not need it because they are copying from other materials that have repair difficulty set?

Issue #65849, opened 2023-05-25
"Psychopath" trait shouldnt disable "Killer Drive"


harmez7:
### Describe the bug

they disable each other so you cant have both of them on your character

### Steps to reproduce

try to have both "Psychopath" and "Killer Drive" trait on your character

### Expected behavior

i think you can be a Psychopath & enjoy killing at the same time

GuardianDll:
FYI if you plan to make a pull request, it's not necessary to make an issue

harmez7:
ok thanks

harmez7:
it was a wrong issue anyways

PR #65850, opened 2023-05-25
removed type category from "Killer Drive" trait

voiddeal:
#### Summary
Bugfixes "fixed "Killer Drive" trait"

#### Purpose of change

Addresses issue #65849. i think you can have both traits without conflict

#### Describe the solution

removed `type` from "Killer Drive" trait so it doesnt conflict with "Psychopath" and stand out of the `HUMAN_EMPATHY` mutation-type
also removed its `prereqs`

#### Testing

small json changes

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -1147,8 +1147,6 @@
     "starting_trait": true,
     "valid": false,
     "social_modifiers": { "intimidate": 10 },
-    "types": [ "HUMAN_EMPATHY" ],
-    "prereqs": [ "PSYCHOPATH" ],
     "flags": [ "CANNIBAL" ]
   },
   {


Qrox:
What exactly does psychopath offer that is not offered by killer drive though? Judging from the json code, killer drive is an upgrade to psychopath and is strictly superior to psychopath. If there's any hardcoded effect that is not present in killer drive, it should probably be updated in the c++ code to use a psychopath flag instead. (There's an open PR replacing hardcoded trait ids of psychopath with flags #65849)

harmez7:
yes, killer drive seems like an upgrade to psychopath
but i not sure if killer drive has the benefit of psychopath as well since they both share a same point cost (2)
going to check it out...

Venera3:
All traits need a base human trait category to be cancelable, this will not pass the unit test. Also, the traits are both mental and build on each other.

harmez7:
> All traits need a base human trait category

you mean `type` ?
if so well, not all traits have their own type according to the json file. "fast learner" for example

Venera3:
Yes, don't remove the type. 
Point cost is pretty irrelevant in the first place, especially with the new character creation

harmez7:
ok thanks

Issue #65847, opened 2023-05-25
Guns too tiny for carry holster(And other problems?)


Ruminator24:
### Describe the bug

The new carry guns that the zombies have is cool, but there's a few oddities with them. One oddity is that some guns will not spawn in their holster, but instead in the zombie's inventory. Another oddity is that when a gun does spawn in the holster, it will have no magazine if it's a semiautomatic.

### Attach save file

N/A

### Steps to reproduce

1. spawn a ton of zombies and then kill them all
2. check zombies for concealed carry and notice that it's a bit wonky

### Expected behavior

I would expect holstered guns to at least have their magazines, even if it's empty, or that the guns spawned with the zombie to have a fitting holster.

### Screenshots

![too smalle](https://github.com/CleverRaven/Cataclysm-DDA/assets/131492998/0816f3db-540f-4f43-bec1-e84ad870eb34)
![no magazine](https://github.com/CleverRaven/Cataclysm-DDA/assets/131492998/ef829e37-bc8a-4568-8468-43f49c8839a8)

### Versions and configuration

- Windows 10
- ddd274c
- tiles
- english
- basegame start

### Additional context

I'd assume the problem would be remedied by simply adding more types of holsters that can spawn with the zombie for concealed carry, especially for smaller guns. I'm not sure what causes the magazine problem, other than there being a problem with how charges are set.

RenechCDDA:
There's a debug message if the container-item of an item group is too small to contain its contents. For the first image you posted, the Ruger is actually **too small** (160mL vs 250mL minimum) to fit in the holster (so the opposite problem, and one that does not have a debug message afaik).

I can't confirm whether they're actually from the same item group at the moment.

TimPollard:
Yeah, it probably needs a re-work to get two lists of guns, one for full size holsters, and one for small holsters. (Alternatively the holsters should be re-worked to not have such small minimums).

The issue with lack of ammo/mags was also reported in #65412, though my attempts to fix that haven't gone anywhere. (I'm clearly misunderstanding the Class and Inheritance structure of Cataclysm somewhere)

PR #65852, opened 2023-05-25
Using default mags to calculate max capacity for guns with random ammo

### LINKED ISSUE #65412: Carry gun on regular zombie dont have magazine nor ammo
### Describe the bug

For some reason, the guns carried by regular zeds don't have magazines. I get it if it has empty mag, but i dont get when the whole magazine has gone as if they just drop the mag on the ground

### Attach save file

N/A

### Steps to reproduce

1.Kill zombies till you found the one that carried a handgun
2. see that it doesnt have magazine nor ammo

### Expected behavior

Having a magazine, preferably with some ammo inside.

### Screenshots

Screenshot_20230501-211053

- OS: Android
- Game Version: 3010c62 [32-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm],
    Crazy Cataclysm [crazy_cataclysm]
]

### Additional context

_No response_

TimPollard:
#### Summary
Bugfixes "Using default magazine to calculate maximum capacity for guns with random ammo"

#### Purpose of change

This should fix #65412 (which is also discussed in #65847)

`doc/ITEM_SPAWN.md` states that "Setting only min and not max will make the game calculate the max charges based on container or ammo/magazine capacity."

The problem was that for magazine fed guns it would calculate the max charges based on the assumption it didn't have a magazine, which would result in a max of 0, which doesn't seem to match the documentation (and also doesn't match how it works with revolvers).

#### Describe the solution

It now will calculate the max charges based on the capacity of the item's default magazine (if one exists).

To fix the specific issue in #65412 I also considered a few other approaches:

1. Manually specifying maximum capacities for 20 or 30 guns in the JSON.
2. Tweaking the code so that `magazine` chance applies before calculating ammo capacity and adding a magazine chance to the top of the item group.

Both of these still result in: 
* semi-automatics acting differently to revolvers
* and code that doesn't match the docs

I also considered implementing one of the above, and changing the docs, but it seems unlikely that anyone would specify `charges-min` and really mean that they want exactly 0 charges (you'd expect them to specify `charges` then).

#### Testing

I added the following as a death drop for a debug monster and checked the results before before any after my change:

```
{
  "type": "item_group",
  "subtype": "collection",
  "id": "debug_death_drops_ammo_changes",
  "entries": [
    { "item": "glock_17",     "charges-min": 0 },
    { "item": "glock_19",     "charges-min": 5 },
    { "item": "glock_21",     "charges-min": 0, "charges-max": 13 },
    { "item": "glock_22",                       "charges-max": 15 },
    { "item": "glock_31" },

    { "item": "m1911mag", "charges-min": 0 },
    { "item": "m9mag",    "charges-min": 5 },
    { "item": "kp32mag",  "charges-min": 0, "charges-max": 7 },
    { "item": "px4_40mag",                  "charges-max": 14 },
    { "item": "px4mag" }
  ]
}
```

This had the following results:
| item | before change | after change |
| --- | --- | --- |
| Glock 17 | 0 ammo | 0 to 17 ammo |
| Glock 19 | 0 ammo | 0 to 15 ammo |
| Glock 21 | 0 to 13 ammo | 0 to 13 ammo |
| Glock 22 | 0 to 15 ammo | 0 to 15 ammo |
| Glock 31 | 0 ammo | 0 ammo |
| 1911 Mag | 0 ammo | 0 to 7 ammo |
| M9 Mag | 0 ammo | 0 to 15 ammo |
| P-32 Mag | 0 to 7 ammo | 0 to 7 ammo |
| Px4 .40 Mag | 0 to 14 ammo | 0 to 14 ammo |
| Px4 9mm Mag | 0 ammo | 0 ammo |

I also temporarily turned up the drop rate of EDC guns and checked that zombies where dropping guns that generally had ammo.

### DIFF
--- a/src/item_group.cpp
+++ b/src/item_group.cpp
@@ -425,9 +425,22 @@ void Item_modifier::modify( item &new_item, const std::string &context ) const
     }
 
     int max_capacity = -1;
-    if( charges.first != -1 && charges.second == -1 && new_item.is_magazine() ) {
-        const int max_ammo = new_item.ammo_capacity( item_controller->find_template(
-                                 new_item.ammo_default() )->ammo->type );
+
+    if( charges.first != -1 && charges.second == -1 && ( new_item.is_magazine() ||
+            new_item.uses_magazine() ) ) {
+        int max_ammo = 0;
+
+        if( new_item.is_magazine() ) {
+            // Get the ammo capacity of the new item itself
+            max_ammo = new_item.ammo_capacity( item_controller->find_template(
+                                                   new_item.ammo_default() )->ammo->type );
+        } else if( !new_item.magazine_default().is_null() ) {
+            // Get the capacity of the item's default magazine
+            max_ammo = item_controller->find_template( new_item.magazine_default() )->magazine->capacity;
+        }
+        // Don't change the ammo capacity from 0 if the item isn't a magazine
+        // and doesn't have a default magazine with a capacity
+
         if( max_ammo > 0 ) {
             max_capacity = max_ammo;
         }


Ruminator24:
You're doing gods work here, i thank you very much

Ruminator24:
Do you mind checking out the gun show location as well? I believe it has the same problem, as none of the guns that use magazines have any. I can understand them being empty since it's a gun show, but no magazines exist for those weapons as well.

fairyarmadillo:
Looks great! Will this affect gun store guns? They should remain empty while on the shelves, since the ammo's locked in the back.

TimPollard:
@fairyarmadillo and @Ruminator24, this doesn't appear to have affected either the Gun Stores or the Gun Show in my quick debug testing.

It should have only changed things in the specific case of someone specifying a `min-charges` for a gun, but not a `max-charges`. If the person who wrote the JSON for a specific item group either didn't specify charges at all, or specified a specific number of charges it won't have changed the behaviour.

It seems whoever set up the JSON for the item groups that are used for both gun stores and gun shows specifically wanted them to generate without any ammo, so they set both `charges-min` and `charges-max` to 0. (I'm not sure why they didn't just set `charges` to 0, but perhaps I'm missing something)

PR #65857, opened 2023-05-25
Fix energy consumption and size of few mini appliances

KHeket:
#### Summary
Bugfixes "Fixing energy consumption and size of few mini appliances"

#### Purpose of change
Size of minifreezer and minifridge is 20L, but it can hold 75L, the same problem with small dishwasher and small washing_machin, and also energy comnsumtion of mini freezer/fridged is too big

#### Describe the solution
Increased size of this appliances and decreased energy comsumption of mini fridge and mini freezers

No alternatives 

#### Testing
Downloaded last version of the game, copy-pasterd edited json files, launched the game - size is normal now, energy comsumption in normal too

#### Additional context
I made some calculations, and wanted to fix this problems

EC - energy consumption. In brackets "( )" will be energy comsumption after changes
Volume - how much can hold fridge/freezer or its mini variant
Surface area - area of cold air volume (500/450L/75L)
It - Insulation thickness (I will just type, how much plastic chunks was used for crafting)
|   | Fridge | Minifridge | Freezer| Minifreezer |
| ------------- | ------------- | ------------- | ------------- | ------------- |
| Volume  | 500  | 75  | 450  | 75  |
| Surface area  | 377  | 106  | 352  | 106  |
| EC  | 30  | 30(10)  | 35  | 45(12)  |
| It  | 16  | 8  | 20  | 10  |

Energy comsumption of fridges and freezers depends on the volume of cooling air, surface area, bigger surface area - more contact area with air around, cooled air inside will heat up faster and will be needed to cooled more, and insulation thickness. More insulation thickness > less heat from outside > less energy comsumption for colling

As you can see in table, volume of minifridge is 6,66 time less, than its bigger variant, and for freezers its 6 times
And also, surface area  area only is ~ 3-4 time bigges for big variant.
Plastic chunks(which is used for insulation) for freezer is 20, for minifreezers is 10, and for frudges is 16, for minifridges is 8, the difference is 2 times, mnivariants have very, very good insulation for their size
So we have decrasing volume by ~ 6 time, surface area ~ 3-4 times and insulation thickness only by 2 time
And by all this parameters, I think, decrasing energy comsumption in 3 times from their big variants is good enogh, because of little size and very thick  insulation (also check comments to see limk to real existing variant)


data/json/vehicleparts/vehicle_parts.json
```diff
@@ -602,8 +602,8 @@
     "damage_modifier": 80,
     "durability": 100,
     "description": "A small fridge.  When turned on, it will cool the food inside, extending the time until the food spoils.",
-    "//": "Use average consumption, not the max on the appliance rating plate. 30W ~ 260kWh per annum",
-    "epower": "-30 W",
+    "//": "Use average consumption, not the max on the appliance rating plate. For big fridge 30W ~ 260kWh per annum is good, but for mini variant with decrased volume by ~ 6 time, surface area ~ 3-4 times and insulation thickness only by 2 time 5W is quite good ",
```
ehughsbaird, line 605: We don't need to do the math you did to determine this, we can just look at similar minifridges (75L ~= 2.6 cubic ft):
https://www.amazon.com/Frestec-Fridge-FR-250-WH/dp/B0BBVRMWWP
https://www.amazon.com/Upstreman-Refrigerator-Adjustable-Thermostat-Red-CR25/dp/B0BVLPFWWP
Both of these list an energy consumption of 210kWh per annum, so that is... [24W](https://www.wolframalpha.com/input?i=210kWh+%2F+%28year+in+seconds%29).

data/json/vehicleparts/vehicle_parts.json
```diff
@@ -639,7 +639,7 @@
     "looks_like": "minifridge",
     "damage_modifier": 80,
     "durability": 100,
-    "epower": "-45 W",
+    "epower": "-7 W",
```
ehughsbaird, line 642: I can't find numbers for this quite so easily, but I would not expect a reduction like this given the minifridge results.

data/json/vehicleparts/vehicle_parts.json
KHeket: If big fridge comsumption has been looked for many time
image
So, **everage** consumption cannot be decreased only by 20%, when you decrasing volume by ~ 6 time, surface area ~ 3-4 times and insulation thickness only by 2 time. At least energy consumtion in this case should be decresed by 4-5 times
I also found lots of different models, like https://www.amazon.com/DOMETIC-75-Liter-Portable-Refrigerator-Freezer/dp/B09FS4YJ8F which is Refrigerator+freezer and consuming ultra low. If I am right, it have ~ 86 kWH per annum, what is 10W. And our ingame mini fridges have good insulation as I can understand. Thicker insulation - less energy wasting. 
If the ingame fridge looks like cheap fridge from amazon - it doesnt mean, that it will work like that

Its hard to find good examples, but if you check oficcial ENERGYSTAR site, you can see, that there is refrigerators, which can hold more litres and comsuming less, than linked by you from amazon (if you make sorting by annual energy use)
https://www.energystar.gov/productfinder/product/certified-residential-refrigerators/

ehughsbaird: I'm not sure where the numbers come from for the amazon link, the only one I see there is "0.01 Kilowatt Hours" which doesn't make much sense to me. Past that, I personally would not say that is a minifridge.
Looking at the 49 2.6 cubic feet (~75L) models on the energystar website, I find the following frequency for energy consumption values (in kWh):
```
      1 208
      7 214
     36 215
      2 216
      1 228
      1 253
      1 259
```
I don't really think it's worth the time to think about a hypothetical fridge, when we have all this data for fridges that match this one.

KHeket: Its worth the time. Because in game we have tech specs of fridge (volume, recipe parts and energy comsumption) and can calculate , when at sites lots of this parameters will be very different or hidden
About energy star, if the sorting by annual energy use is enable, so


<details>
  <summary>Spoiler warning</summary>
  
image
image
image

</details>
Yes, they can hold more, but if the bigger fridge can do that, so do and less bigger

Insulation is playing very important role in energy consumption and cols saving
<details>
  <summary>Spoiler warning</summary>
  
image

</details>

We can make it that way, that the mini freezers crafted by player (and rarely finded in cars) have thick insulation and that why they need less energy, what is vital in the world without standart power generation (and even when you are traveling by car). And if somebody want to make cheap variant like the low price variants from amazon - he can do it. And the cheapest variant can be less bigger bacause of less insulation thickness. I also made the logic in that way, that have 1 big fridge(500L/30W)/freezer(450L/35W) is energy economical than have 6,6 minifridgres (~500L/33W) and 1 nig freezer (450L/35W) vs 6 minifreezers (450L/42W) , because of the calculation in the pr

ehughsbaird:
I don't think we should assume the crafting recipe is how the energy usage stats should be determined when [we have real world analogues](https://github.com/CleverRaven/Cataclysm-DDA/pull/65857#discussion_r1207553516).

KHeket:
> I don't think we should assume the crafting recipe is how the energy usage stats should be determined when [we have real world analogues](https://github.com/CleverRaven/Cataclysm-DDA/pull/65857#discussion_r1207553516).

I also showed you real world analogues, which consuming less energy than your examples, and also show you, that construction and what is more important insulation playing big role in energy loss

KHeket:
I think for some time, trying to find some energy economic fridges, and found
https://subcold.com/products/75-litre-mini-fridge-eco75
image
What is ~ 11,8W
I dunno, why my logic dont work here, but it seems, bigger fridge is more economic, than mini wariant, may be in future I will create some variants of minifridges, but for now I will make close to this
Its size: 486W x 500D x 627H in mm, what is ~ 150 L
So I will make nubers closer to it

data/json/vehicleparts/vehicle_parts.json
```diff
@@ -639,7 +639,7 @@
     "looks_like": "minifridge",
     "damage_modifier": 80,
     "durability": 100,
-    "epower": "-45 W",
+    "epower": "-12 W",
```
kevingranade, line 642:
```suggestion
    "epower": "-36 W",
```

kevingranade:
Ehugsbaird is right here, you're being incredibly optomistic in your analysis and coming up with an unbelievably good outcome. This is the reason we just go with published figures instead of estimating from first principles,  there are huge numbers of variables at play here and it's too easy to make a reasonable looking argument that has some bad assumptions baked into it.

data/json/vehicleparts/vehicle_parts.json
```diff
@@ -602,8 +602,8 @@
     "damage_modifier": 80,
     "durability": 100,
     "description": "A small fridge.  When turned on, it will cool the food inside, extending the time until the food spoils.",
-    "//": "Use average consumption, not the max on the appliance rating plate. 30W ~ 260kWh per annum",
-    "epower": "-30 W",
+    "//": "Use average consumption, not the max on the appliance rating plate. For big fridge 30W ~ 260kWh per annum is good, but for mini variant with decrased volume by ~ 6 time, surface area ~ 3-4 times and insulation thickness only by 2 time 10W is good too and very close to real world variant ",
+    "epower": "-10 W",
```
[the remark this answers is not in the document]
KHeket: 24W comnsumption is good for very cheap fridge
https://subcold.com/products/75-litre-mini-fridge-eco75 - here is 11.8W variant
https://www.amazon.com/DOMETIC-75-Liter-Portable-Refrigerator-Freezer/dp/B09FS4YJ8F here is 10W
And there was two very cheap 24W variants
Make at leat 15W for minifridges, because it will be ~ middle value of bad and good mini fridges

data/json/vehicleparts/vehicle_parts.json
KHeket: While 35W is enough good for 450L freezer, mini variant cant techically consume more power
As variant https://www.osculati.com/en/11211-50.042.00/isotherm-refrigeratorfreezer-bi75-75-l
image
350w/24 = 14.5W
So make at least 2 times lesser, than its big variant. 17W is quite ok even for hand made fridge

KHeket:
> This is the reason we just go with published figures

Yeah, I am agree with you. But when we have lots of published figures, its better to take middle values

PR #65858, opened 2023-05-26
[Innawoods] Yeet the professions, add more professions, make Portal Dependent possible w/ metaprog on

TheShadowFerret:
#### Summary
Mods "[Innawoods]: Obsolete professions w/ implausible tech, make Portal Dependent possible w/ meta prog on"

#### Purpose of change

Scenarios: Portal Dependent can't be unlocked w/ meta progression active due to no labs.
Professions: There are a lot of futuretech ones that need to go.

#### Describe the solution

-Change Portal Dependent's meta requirement to surviving a portal storm.
-Add profession whitelist to all non-"came from portal storm" scenarios consisting of all Innawoods overrides/additions, Naked, Shower Victim, and Survivalist(none of those three have futuretech).
-Lower tech levels on Goat Man.
-Add override of Farmer & Trapper
-Add tribal swordsman & arbalist as Innawood military analogues
-Add fisher who has a canoe
-Add carpenter & bricklayers as starts geared towards constructing a house instead of living in a cave.

~~Avoid Fris0uman killing me for not allowing Foodperson~~
Starting car/heli crash in a portal storm ~~to give a reason for those to exist~~ cuz I'm mean 😈

Using new IDs for the overrides so Car/Helicopter Crash can use those.
#### Testing

Load game, receive no errors.
Scroll over Portal Dependent, see updated meta progression.
Pick Canoe Fisherman, find & drag around the canoe.

#### Additional context

Canoe Fisherman gets no boat if you start too far from water, sorry 😛

People longing for more professions after all the futuretech ones disappear was entirely expected, fostering future PRs to fix this was an intended effect.


actually-the-artificer:
I think the description of Car Crash/Heli Crash leaves it ambiguous whether it was portal storm or not. Though it could be a good change and add an extra layer of challenge!

Light-Wave:
The idea behind the car and chopper crash is that you drove through a portal storm and ended up transported to the Innawood universe. So you can leave those in.

NetSysFire:
Tears in reality (and similar phenomena) can be caused by portal storms but also do persist when the weather is not a portal storm, so enforcing the weather to be a portal storm is indeed something that does not need to be done.

PR #65865, opened 2023-05-26
Overmap special flag audit

FarkasKBence:
#### Summary

None

#### Purpose of change

I was messing around with a mod, adding location blacklists & whitelists, when I noticed some inconsistencies with overmap special flags.

#### Describe the solution

Added a "MAN_MADE" flag to locations that were made by humans, and the "FARM" flag for the farm mutable and the abandoned farm.

The "WILDERNESS" flag has no description in JSON_FLAGS.md and it's used very inconsistently, so I gave it a description and revised flags accordingly.

I also removed the "CLASSIC" flag for the nether monster corpse as it doesn't make sense in a classic zombie movie.

None.

#### Testing

Vanilla, mods seem to load okay.

#### Additional context

Putting "URBAN" and "MAN_MADE" together might have been a bit overkill.

### DIFF
--- a/data/json/overmap/overmap_mutable/homeless_camp_mutable.json
+++ b/data/json/overmap/overmap_mutable/homeless_camp_mutable.json
@@ -7,7 +7,7 @@
     "city_distance": [ 0, 24 ],
     "city_sizes": [ 6, -1 ],
     "occurrences": [ 0, 3 ],
-    "flags": [ "CLASSIC", "URBAN" ],
+    "flags": [ "CLASSIC", "URBAN", "MAN_MADE" ],
     "check_for_locations": [
       [ [ 0, 0, 0 ], [ "forest" ] ],
       [ [ 1, 0, 0 ], [ "forest" ] ],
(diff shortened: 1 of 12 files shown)


data/json/overmap/overmap_mutable/homeless_camp_mutable.json
```diff
@@ -7,7 +7,7 @@
     "city_distance": [ 0, 24 ],
     "city_sizes": [ 6, -1 ],
     "occurrences": [ 0, 3 ],
-    "flags": [ "CLASSIC", "URBAN" ],
+    "flags": [ "CLASSIC", "URBAN", "MAN_MADE" ],
```
Maleclypse, line 10: Wouldn't homeless camps be WIlderness instead of urban since there shouldn't be road connections to a homeless encampment in general?

LovamKicsiGazsii: Yes, that was an oversight.

Maleclypse (approved):
I think all of these make sense it was a little hard to review because github kept moving me up whenever I'd expand text to see what the special was so I probably missed some.  Thank you for adding a doc explanation for WILDERNESS

data/json/overmap/overmap_mutable/homeless_camp_mutable.json
Maleclypse, line 10:
```suggestion
    "flags": [ "CLASSIC", "WILDERNESS", "MAN_MADE" ],
```

PR #65870, opened 2023-05-27
[CR] Overhaul basecamp handling resources

### LINKED ISSUE #63052: Basecamp crafting consumes no ingredients
### Describe the bug

This is pretty much the first time i'm dealing with the basecamp system so apologies if i'm doing it wrong.

I have two NPCs with me. I've set up a basecamp. Assigned both to the camp.
I have a fridge with 34 cattail rhizomes.
I sent one of my npcs to convert the rhizomes to starch (because i'd rather not do it myself).
When they finished, i had the 17 starches (as expected), but also the original 34 cattail rhizomes.

Something something infinite food glitch.

Tangentially: the resulting starch is dropped onto the *floor* where the fridge is, not into the fridge itself. Which is both unfortunate and easy to overlook, but also unrelated to the issue at hand.

### Steps to reproduce

- Load attached savefile
- Inspect the fridge, see 34 cattail rhizomes + 17 starches
- Go to builletin board (bit to the west), ask to make `[B] Cook: Cattail Starch`
- When prompted for the amount, enter the maximum of 17
- `|` -> wait 6 hours until finished
- Inspect the fridge again, see 34 cattail rhizomes + 17 starches + 17 more starches

### Expected behavior

materials get consumed

### Screenshots

_No response_

- OS: Windows
- Game Version: ba42c0a [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

Savefile: [Winters-starch.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/10376935/Winters-starch.zip)

### LINKED ISSUE #63722: Base Camp projects won't consume resources
### Describe the bug

Base Camp projects won't consume resources

### Attach save file

[Quakertown-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/10797597/Quakertown-trimmed.tar.gz)

### Steps to reproduce

1- go to the main board
2 - start a project supposed to consume resources
3 - the resources are still there
4 - optionally start another project with the same resources that should be now depleted...

### Expected behavior

resources are depleted during the building process

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.22621.1265 (22H2)
- Game Version: c571723 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm],
    Arcana and Magic Items [Arcana],
    Arcana/Magiclysm Patchmod [Arcana_Magiclysm_Patch],
    SpeedyDex [speedydex],
    Owlfren's Map Additions [owlfren_mapping],
    Tamable Wildlife [Tamable_Wildlife]
]

### Additional context

_No response_

SurFlurer:
#### Summary
Bugfixes "Overhaul basecamp consuming resources"

#### Purpose of change
Fix #63722
Fix #63052

There are multiple issues complaining about basecamp missions do not consume resources. The current implementation does have several inconsistencies. 

1. If the player accesses the camp directly (not via radio), the code searchs for storage zones in a range of 60, centering on the player's location. When forming crafting inventory to determine which camp recipes are available, it search for materials in a range of 60, centering on the nearest tile that's inside a storage zone ( aka the dumping_spot ). And it search for fuels in a range of 20, centering on the dumping_spot. When actually consuming resources, it search for resources in a range of 20, centering on the dumping_spot. This is likely to neglect some stockpiles, if there are multiple storage zones, or the zone is quite big.

2. If the player accesses the camp via radio, the code only loads a tinymap, making it possible that the dumping spot is outside that tinymap.

3. The player can call back follower NPCs by using `#` to view follower npcs immediately. However, using the camp ui to call back assignees requires the mission has been started at least 24 hours ago.

4. The NPC called back by `#` will not place results into a storage zone.

There's also a minor display bug that is fixed by this PR.
image

#### Describe the solution
1. Create variants of the resource consuming functions that I need, like `Character::consume_items()`, `Character::consume_tools()`, `map::use_charges()`, `map::use_amount()`, and `map::consume_ups()`, to search inside a specific set of tiles, instead of searching in a range.

2. When calling the camp via radio, load a full-sized map (11x11 submaps) of the camp, to ensure a consistent player experience ( available missions, resource consumption, etc. ). That map is kept during camp missions navigation and mission handling ( to minimize overhead ), and gets unloaded once the call is finished ( to avoid unsafely overlapping with the main map).

3. Eliminate the usage of inv_range. Always use 60 as the range when validating and storing storage zones. Storing all available storage zone tiles positions in the basecamp object, and use that set of tripoints to form crafting inventory and consume resources. Refresh that set only if nearby storage zones changes.

4. When calling back NPCs that have camp missions, call the camp UI, to make sure the time limit is applied, and results are placed. This also imitates the NPC with radio at the camp being able to distribute PC's commands to other NPCs that do not have a radio.

Do not touch the resource consuming functions, and just brute search in a range of 60 for resources.

#### Testing
It fixes #63722. I continued assigning basecamp missions, and resources were correctly consumed.

A more conprehensive testing:
Create a basecamp, create 5 storage zones (one beside the bulletin board (aka BB), then one 45 tiles away in each direction of east, west, north, south ). Place a knife in the nearest zone. Place 5 planks in each of the zone.

First, examine the BB, and assign a follower to craft 6 pointy sticks, for 4 times. Each time, the resources are correctly consumed.

Then teleport to several overmap tiles away, and use the radio to call my follower to craft 6 wood sticks, for 4 times. Each time, the resources are correctly consumed. Teleport back, and I can see the pointy sticks lying in the storage zone.

#### Additional context
Performance:
I tested this on the provided save file of #63722, which looks like a long-running and fabulously rich basecamp ( 10+ followers, a 30T+ vehicle, a big warehouse full of resources. )

When calling the camp by radio, loading the whole 132x132 tiles (11x11 submaps) basecamp map as if the player character was standing by the side of the bulletin board eats around 100MB of RAM.

Then I teleported away from the camp, saved and quit, then loaded into the world again. This is to measure the CPU pressure of loading the camp map. The results are: When the player character is away from the camp and the camp isn't loaded, [this part](https://github.com/SurFlurer/Cataclysm-DDA/blob/843472f3efde38c355f07dbd357dcca642d75828/src/faction_camp.cpp#L713-L731) consumes ~260ms of CPU time.

Here's the flame graph:

image

So loading non-cached camp's map from disk takes around 100ms.

Call that NPC again ( now the submaps has been cached), and the loading of map takes around 60ms:

image


SurFlurer:
When calling the camp by radio, loading the whole 60×60 basecamp map as if the player character was standing by the side of the bulletin board eats around 100MB memory, for a long-running and fabulously rich basecamp.

SurFlurer:
Many thanks for merging!

src/basecamp.cpp
```diff
@@ -657,20 +664,44 @@ std::list<item> basecamp::use_charges( const itype_id &fake_id, int &quantity )
...
+    if( here.check_vehicle_zones( here.get_abs_sub().z() ) ) {
+        mgr.cache_vzones();
+    }
+    tripoint src_loc = here.getlocal( bb_pos ) + point_north;
+    if( mgr.has_near( zone_type_CAMP_STORAGE, abspos, 60 ) ) {
+        const std::vector<const zone_data *> zones = mgr.get_near_zones( zone_type_CAMP_STORAGE, abspos,
+                60 );
+        // Find the nearest unsorted zone to dump objects at
+        if( !zones.empty() ) {
+            if( zones != storage_zones ) {
```
sparr, line 679: This fails to detect if the storage zone has been edited (same pointer, so the vectors of pointers compare equal). Fix incoming.

PR #65872, opened 2023-05-27
fix missing include in crash.cpp

andrejlevkovitch:
g++ (Ubuntu 10.3.0-1ubuntu1~20.04) 10.3.0

#### Summary
Bugfixes "fix missing include in crash.cpp"

#### Purpose of change
make it compilable

#### Describe the solution
add `cstring` include

#### Additional context
lsb_release:
```
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.2 LTS
Release:	20.04
Codename:	focal
```

g++ --version:
```
g++ (Ubuntu 10.3.0-1ubuntu1~20.04) 10.3.0
Copyright (C) 2020 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

### DIFF
--- a/src/crash.cpp
+++ b/src/crash.cpp
@@ -7,6 +7,7 @@
 #include <csignal>
 #include <cstdio>
 #include <cstdlib>
+#include <cstring>
 #include <exception>
 #include <initializer_list>
 #include <iostream>


Irets:
I noticed the same issue on Gentoo Linux.
Didn't think it was worth bringing up but it's interesting to see the same issue affects Ubuntu.

My system compiles developement version of cdda with a build script using the package manage.
I've used **sed** to include cstring in crash.cpp in the script before the compiler starts. This has been my local fix for now.

PR #65873, opened 2023-05-27
Refugee Center dialogue fix - Rhyzaea Johnny - Slovakia/Slovenia

Tread4:
#### Summary
Bugfixes "Refugee Center NPC visited coastal forests in landlocked Slovakia (I believe it should be Slovenia)"

#### Purpose of change
Rhyzaea Johnny mentions visiting coastal forests in Slovakia. It is a landlocked country, so if in world of Cataclysm (before Cataclysm, so he had chance to visit) nothing majored happened, that gave Slovakia access to see, it should be straight impossible.

I suppose it might be the case of simple mix-up of Slovakia and Slovenia. This second country is indeed coastal and as far as I checked it have some coastal forests. If there was an underlying reason to write Slovakia (or I misunderstand "coastal forest") this PR can be scrapped, but in my opinion we should make it consistent with reality.

#### Describe the solution
Very minor dialogue text adjustment.

None
If PR gets merged it should be repeated for translations (or otherwise translators should be notified of the change - no idea how system works currently)

#### Testing
- Went to Refugee Center and had conversation with Johnny
- Confirmed that now he mentions Slovenia in place of Slovakia

#### Additional context
Any other, country that fits, what Johnny says, can be picked up as well. I went for Slovenia as it both should fit and possibly was an initial idea of the dialogue author (potential mix-up of similar names).

### DIFF
--- a/data/json/npcs/refugee_center/surface_refugees/NPC_Rhyzaea_Johnny.json
+++ b/data/json/npcs/refugee_center/surface_refugees/NPC_Rhyzaea_Johnny.json
@@ -157,7 +157,7 @@
   {
     "type": "talk_topic",
     "id": "TALK_REFUGEE_Rhyzaea_Band",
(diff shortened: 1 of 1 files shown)


RenechCDDA (approved):
Reasoning checks out! These two countries do get mixed up a lot (https://www.bbc.com/news/world-europe-43419675). And Slovakia definitely doesn't have a coastline.

>If PR gets merged it should be repeated for translations (or otherwise translators should be notified of the change - no idea how system works currently)

This should happen automatically. Transifex regularly checks for new strings (even an old string that's been changed is a "new" string) and puts it in their system for translation work.

kholat:
POG im from slovenia!

PR #65874, opened 2023-05-27
Forest floor should look like moss and not tall grass

KittyTac:
#### Summary
Bugfixes "Forest floor should look like moss and not tall grass"

#### Purpose of change
Forest floor, described as being covered with twigs and dead leaves, but not grass, looks like tall grass, and does not slow movement. This affects tactics and makes it pretty confusing to tell at a glance.

#### Describe the solution
Change one line.

Making a new tile for it (out of scope, this is just changing a placeholder for a less bad placeholder)

#### Testing
Before:
изображение

After:
изображение

#### Additional context
This is probably my smallest PR.

### DIFF
--- a/data/json/furniture_and_terrain/terrain-floors-outdoors.json
+++ b/data/json/furniture_and_terrain/terrain-floors-outdoors.json
@@ -19,7 +19,7 @@
     "description": "Covered in twigs and lots of brown, decaying leaves.",
     "symbol": ".",
     "color": "brown",
-    "looks_like": "t_grass_tall",
+    "looks_like": "t_dirt",
     "move_cost": 2,
     "flags": [ "TRANSPARENT", "DIGGABLE", "FLAT" ],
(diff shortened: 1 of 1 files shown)


NetSysFire:
Fun fact: I actually tried this first and it made the forest look very sad. Using tall grass made it look a bit like before and less barren.

KittyTac:
> Fun fact: I actually tried this first and it made the forest look very sad. Using tall grass made it look a bit like before and less barren.

Yeah but as it doesn't slow anything down unlike tall grass, it's actually very detrimental to gameplay as you can't easily see what slows monsters down. Maybe someone will make a better tile but for now dirt is the least bad choice.

TaiGambol:
Could short grass be used instead, to get the best of both worlds?

KittyTac:
> Could short grass be used instead, to get the best of both worlds? Like, the grass shown in the second image around the dirt patch.

That's moss. I could make it moss.

KittyTac:
...why doesn't it build?

NetSysFire:
Math. Dont worry about it, its unrelated and the required build has passed.

PR #65876, opened 2023-05-27
Moving enchantments from Magiclysm classes to separated JSON and repl…

LightDimf:
…acing "arithmetic" with "math"

#### Summary
Mods "Magiclysm class enchantments fixes"

#### Purpose of change
Crash on load when loading a mod that tries to overwrite magiclysm classes that have enchantments declared inside due to the fact that game tries to dynamically create 2 identical enchantments.
Also enchantments still uses outdated "arithmetic" function.

#### Describe the solution
Declare enchantments separately in their own file and put enchantment id in classes. Replace "arithmetic" function with "math".


GuardianDll:
Apply the bot's suggestions please

GuardianDll:
It seems you can't apply the last suggestion, as it is about the end od line issue. In this case you need to lint Basic_Classes.json manually, using https://dev.narc.ro/cataclysm/format.html

PR #53821, opened 2021-12-27
Adding gun mod to use m240 and m249 in semi-auto mode

GuardianDll:
#### Summary
Content "Adding gun modification to use m240 and m249 in semi-auto mode"

#### Purpose of change
There is no way to use them in semi-auto mode

#### Describe the solution
According to #50832, maded a mod that allows to convert m240 and m249 into semi-auto mode

#### Describe alternatives you've considered
not doing this

#### Testing
works fine

PR #65883, opened 2023-05-27
M249/M240 semi triggers give select fire

anoobindisguise:
#### Summary
Bugfixes "Machine gun semi auto mods don't override the default burst"

#### Purpose of change
The PR #53821 that introduced trigger mods for M249 and M240 references the patent https://patents.google.com/patent/US8087343B2/en which states it enables select fire, but in-game if you install the mod it actually only allows semi automatic fire.

#### Describe the solution
Make it select fire. 
ALSO, makes the recipe need the gunsmithing prof when previously it was just welding.

Decide that it's intentional that it's semi auto only as a more improvised conversion done by the survivor rather than referencing a professionally made mechanism from a manual.


TheShadowFerret:
Shouldn't this be called "select fire trigger" then if it allows both 🤔

PR #65888, opened 2023-05-28
Civilians: bugfixes

### LINKED ISSUE #65796: New Civilian Monsters are too Plentiful

As of #64142 being merged there are a few new neutral civilian monsters that are meant to depict humans that haven't gone feral but haven't evacuated cities or have remained trapped in them. I'm posting this issue to hopefully invite further civil discussion on the implementation.

My main issue with these new monsters is that there's an abundance of them. I play pretty frequently and have already put a relatively significant time into the version with these new monsters, even though it was merged only hours ago. I've already played several city starts, and it needs to be said that for every 5 road tiles in cities, there's about 5 or more panicked/petrified civilians.

This seems bizarre considering the lore as well as gameplay implications. The half-life of these civilians evolving into feral humans is half a day, which also means that if you don't spawn in a city, approaching one will be met with more than double the amount of ferals you'd usually see.

### Solution you would like.

I'd like to see the spawning of these new monsters be greatly reduced. It doesn't make sense in terms of lore or gameplay for there to be hordes of non-feral humans roaming cities just for them to evolve the next day. I'd also like to see the areas they can spawn and their behavior tweaked a bit.

I'm open to hearing what others have to say about alternatives and compromises, but personally I think the spawning just needs to be tweaked to make more sense and be less intrusive. 

### Additional context

I'd like to state just for the record that I actually love this merge. I almost exclusively play city starts and I think it brings extra strategic movement and intrigue to an otherwise monotonous game of running from downspout to downspout. I think the author of the merge has great ideas and implemented them beautifully other than the amount spawned.

NetSysFire:
#### Summary
Bugfixes "Misc bugfixes targeting the new civilian monsters"

#### Purpose of change

Fixes #65796. Also addresses the comments in #64142 that were made after it was merged.

#### Describe the solution

- The back bay of the refugee center was using the monster group `GROUP_MALL`. No wonder there were civilians in there. I changed this to a monster group consisting solely of human zombies. 
- Also adjusts `GROUP_MALL` to also spawn ferals. Seems off for malls to be devoid of them.
- Added a new monster group: `GROUP_FERAL`. I use this for civilian evolution and deduplicating of entries in other monster groups. I am aware there is already `FERAL_HUMANS` but it includes some tougher ferals that should definitely not spawn anymore, most notably the soldier, militia and biker varieties.
- Nerfed the panicked person to be slower and more shortsighted. This will cause it to not run as far I hope.
- Adjusted spawning and evolution. They will be noticably less common after a few days. I uh might not know the implications of using *both* `age_grow` and `half_life` but it appears to work and thats what counts, right?

Renaming `FERAL_HUMANS` to `GROUP_FARM_SUPPLY_STORE` while I was at it. I do not know how many mods use that group though.
Given that this location is rural and full of materials needed for longer term survival, it does make sense in this location. Re-using that group anywhere else is tough since it is rather specific but its name promises something generic.

#### Testing

Loads fine.

- Teleported around at different days in the game. Population of civilians looks as expected.
- Checked the back bay of the refugee center. All as expected.
- Observed the behavior of a panicked person running away. It should be much better now since they will lose sight of the zombie that is chasing them frequently and not see obstacles such as other hordes ahead of them. As a bonus, this helps displaying their erratic and panicked nature.

### DIFF
--- a/data/json/mapgen/refugee_center/refugee_center.json
+++ b/data/json/mapgen/refugee_center/refugee_center.json
@@ -151,7 +151,7 @@
         { "class": "guard", "x": 54, "y": 20 },
         { "class": "guard", "x": 63, "y": 15 }
       ],
-      "place_monsters": [ { "monster": "GROUP_MALL", "x": [ 74, 87 ], "y": [ 4, 20 ], "density": 0.6 } ]
+      "place_monsters": [ { "monster": "GROUP_VANILLA_ONLY_HUMANS", "x": [ 74, 87 ], "y": [ 4, 20 ], "density": 0.6 } ]
     }
   },
   {
--- a/data/json/monstergroups/zombies.json
+++ b/data/json/monstergroups/zombies.json
@@ -10,10 +10,7 @@
       { "monster": "mon_zombie_fat", "weight": 50 },
       { "monster": "mon_zombie_rot", "weight": 50 },
       { "monster": "mon_zombie_runner", "weight": 20, "cost_multiplier": 2 },
-      { "monster": "mon_feral_human_pipe", "weight": 40 },
-      { "monster": "mon_feral_human_crowbar", "weight": 40 },
-      { "monster": "mon_feral_human_axe", "weight": 20, "cost_multiplier": 2 },
-      { "monster": "mon_feral_sapien_spear", "weight": 14, "cost_multiplier": 2 },
+      { "group": "GROUP_FERAL", "weight": 115 },
       { "monster": "mon_zombie_crawler", "weight": 25 },
       { "monster": "mon_zombie_brainless", "weight": 25 },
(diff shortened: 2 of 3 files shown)


AcidAntOnAMinefield:
I'd suggest giving the civilians some actual armour, given they all have a chance at "dropping" leather things as well as motorcycle things and rarely a Kevlar Vest. That or give them a specific drop list of items that offer no protection whatsoever.

NetSysFire:
The reason they drop that is because I am using the `default_zombie_death_drops`. Unfortunately there is no way to calculate armor from their death drops, which can range from a torn tank top to a beefy kevlar vest. I mean sure, I could copy the armor value from `mon_zombie`, but I am not sure if I should. They are meant to be vulnerable and super easy to kill. And especially because they can suffer from pain (although there is no mechanic for it) in theory, their vulnerability is justified in my opinion. I am open to other ideas, as always.

PR #65889, opened 2023-05-28
Add new stable wheels + proto-implementation of one-tile track wheels

a-white-guy-why:
#### Summary
Content "New wheels with STABLE flag: off-road wheelchair wheels, pair of [banded] wooden wheels, makeshift chain tracks and recipes for them"

#### Purpose of change
I was in existential denial when I realized there was only few stable wheels, where the best one in the game "wheelchair wheels" had non-existent off-road capabilities. And when I coded all of suitable wheels, I got an idea about chain tracks and made it too.

#### Describe the solution
Adds new variants with STABLE flag of [banded] wooden wheels and wheelchair wheels made of two off-road bicycle wheels that are a lot more capable off-road than normal wheelchair wheels. Also adds experimental chain track wheels in the game: has the best off-road capabilities in the game, but requires substantially more power to roll them than normal wheels. And of course added recipes for all of them.
I had no idea how to measure rolling resistance of tracks, so I just made it up in a way that makes sense.
Also made minor changes in code style and descriptions.

Create C++ code to emulate behaviour of track chains or just remove the idea of track wheels from my dreams in unforeseen future.
If we want to emulate it, the main problem is with measuring rolling resistance for tracks, as it requires a lot more brain power than with radial wheels. I personally gave up on that, if someone is able to somehow measure it - please do it.

#### Testing
Made a 150 kg electro-bicycle loaded with 300 kg with tracks and a cargo space, everything works fine and is able to go through rough terrain even when grabbed.
image
As I changed little about wooden wheels and wheelchair wheels they're probably working fine too.

#### Additional context
Creating/finding a calculator of rolling resistance would be neat if we plan to streamline wheel code as it will help a lot not just for people that aren't able to calculate it, but also those who are able to, as it will take less time to calculate.
In the future I want to add new items that represent parts of a track to make repairing and creating the track itself more sense and the factory-made version of small chain track.

### DIFF
--- a/data/json/items/vehicle/wheel.json
+++ b/data/json/items/vehicle/wheel.json
@@ -657,10 +657,7 @@
     "id": "wheel_wheelchair",
     "type": "WHEEL",
     "category": "veh_parts",
-    "name": {
-      "str": "pair of wheelchair wheels",
-      "str_pl": "pairs of wheelchair wheels"
-    },
+    "name": { "str": "pair of wheelchair wheels", "str_pl": "pairs of wheelchair wheels" },
     "description": "A pair of wheels for a wheelchair.",
     "weight": "3000 g",
     "volume": "12500 ml",
--- a/data/json/items/vehicle/wheel.json
+++ b/data/json/items/vehicle/wheel.json
@@ -33,7 +33,10 @@
     "id": "wheel_mount_medium",
     "type": "GENERIC",
     "category": "veh_parts",
-    "name": { "str": "medium wheel hub assembly", "str_pl": "medium wheel hub assemblies" },
+    "name": {
+      "str": "medium wheel hub assembly",
+      "str_pl": "medium wheel hub assemblies"
+    },
     "description": "A metal assembly that allows bolting a wheel onto a car.  Fit for a car wheel.",
     "weight": "5 kg",
     "volume": "6 L",
@@ -50,7 +53,10 @@
     "id": "wheel_mount_heavy",
     "type": "GENERIC",
     "category": "veh_parts",
-    "name": { "str": "heavy wheel hub assembly", "str_pl": "heavy wheel hub assemblies" },
+    "name": {
+      "str": "heavy wheel hub assembly",
+      "str_pl": "heavy wheel hub assemblies"
+    },
     "description": "A heavy metal assembly that allows bolting a wheel onto a car.  Fit for a large car wheel.",
     "weight": "6 kg",
     "volume": "7 L",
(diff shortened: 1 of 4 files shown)


Drew4484:
Lightweight answer for rolling resistance, straight from the cat's mouth: <https://www.cat.com/en_US/articles/ci-articles/rolling-resistance-factors.html>. A rubber track (or steel track with rubber pads attached) has significantly lower rolling resistance than a steel track.

Rolling resistance aside, the big obstacle to overcome with tracks is that we currently don't have a system to require continuous track sections (i.e. tracks extending along the length of the vehicle (=====) instead of installed anywhere like standard wheels (=o=o=). So if you aren't intending to tackle that issue with this PR, I would consider implementing something like these <https://www.mattracks.co/tracks/trucks/>.

a-white-guy-why:
> Rolling resistance aside, the big obstacle to overcome with tracks is that we currently don't have a system to require continuous track sections (i.e. tracks extending along the length of the vehicle (=====) instead of installed anywhere like standard wheels (=o=o=). So if you aren't intending to tackle that issue with this PR, I would consider implementing something like these https://www.mattracks.co/tracks/trucks/.

Of course I currently don't want to create multiple-tile track chains, this PR addresses the lack of off-road wheels with STABLE flag. I added small chain tracks as a mid-to-late game wheels for one-tile vehicles to carry heavy loads through forests and such. I will consider about adding unstable ones that you linked as they look sick! (and useful in game)
"Proto-implementation" means here as a proof-of-concept, as this PR doesn't add more mechanics but builds it on the current code.

irwiss:
Not a fan of some this stuff;

Nothing about offroad wheelchair wheels makes any sense:
Bicycle wheels description is changed to claim their design is "similar" but wheelchair wheels have push rims, wheelchair wheel won't fit in the bicycle fork either or be off-center if you somehow hammer it in, what is similar in their design besides being wheel shaped?

The recipe doesn't make sense too (although this was also the case for normal ones before your PR highlighted it) - a 25 minute DIY-something-with-2-pipes-and-welder thing which makes no sense - why only 2 wheels? Where are the front caster wheels and what are the pipes for? It also changes 27" wheels into 20" somehow

Makeshift treads also seem a bit bonkers:

The craft recipe is effectively (for mildly skilled survivor, without the advanced proficiencies) ~4 day DIY project that converts a pile of garbage into something that is also all of: twice more sturdy/durable than wide offroad wheel, only 300 gram heavier than wide offroad wheel and is practically discarding all traction mechanics at ~twice the fuel consumption for a scooter or ~4x the consumption on a semi truck, the linked factory made "treads" are around 200kg of weight
This sounds way over the top even giving it a pass on random survivor crafting "makeshift" tank treads that don't fall apart the moment any torque is applied

You should either use copy-from(which implicitly defines looks_like) or define looks_like, so tilesets don't fallback to ascii, removing `relative` and redefining weight to be the same number also seems a bit weird

Also probably a nitpick but some of the changed descriptions are changed from observation-styled language to recommendation or even meme style that doesn't fit the rest of the game or even the other wheel descriptions that you added a few lines below them

PR #65900, opened 2023-05-28
Makes match triggers better

WoodPollution:
#### Summary
Balance "Make match trigger more usefull"

#### Purpose of change
Match trigger as it is now is awful, makes it better.

#### Describe the solution
After asking on discord i got recommended to put it to 20, thought it might be a bit too high so ended up with 15. This might be a bit high/low, so input is appreciated.

#### Testing
A save loads up with no errors.

### DIFF
--- a/data/json/items/gunmod/mechanism.json
+++ b/data/json/items/gunmod/mechanism.json
@@ -16,7 +16,7 @@
     "location": "mechanism",
     "mod_targets": [ "smg", "rifle", "pistol", "shotgun", "launcher" ],
     "install_time": "30 m",
-    "dispersion_modifier": -1,
+    "dispersion_modifier": -15,
     "min_skills": [ [ "weapon", 4 ], [ "mechanics", 3 ], [ "gun", 2 ] ]
   },
   {


Tread4:
From experience I would say that lighter and smoother trigger does help, but the gains are diminishing on higher skill levels. All in all the difference is noticeable and the feeling is definitely better. Cannot provide any numbers from experience so looked for data:

A quick search thrown this text (with a bit of data) my way:
https://www.tierthreetactical.com/does-a-light-trigger-increase-accuracy-research-and-analysis/
it basically describes this "research" (too small test group to treat it as proper research, yet imo. still some usable data in there) 
https://mtntactical.com/knowledge/trigger-pull-study-initial-findings/

I suppose that grabbing some average value of dispersion on the guns and applying ~5% dispersion reduction bonus sounds reasonable.

Running the received value through "reality check" against other values in-game is still advised (for example if calculation turns -50 bonus it is unreasonable in comparison to other game numbers). No idea what numbers such math exercise will give us since I do not remember nor checked current in-game values.

kevingranade:
This is an absolutely insane value that needs very convincing evidence.
Unfortunately Tread4's studies aren't great either as they don't say what they mean by x% improvement,  lower dispersion? Higher score? Who knows.

Issue #65448, opened 2023-05-03
Small monsters can double as pack-mules


Golfavel:
### Describe the bug

You can attach a bag of any size and weight (e.g., a stuffed backpack) to a friendly creature of any size and strength (e.g., a chicken)

### Attach save file

[Daytona Beach Shores-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11378558/Daytona.Beach.Shores-trimmed.tar.gz)

The raven named "1" holds a military rucksack containing a 60L metal tank filled with water.

### Steps to reproduce

 1. Find a tameable creature that shouldn't be capable of holding anything worth attaching (e.g., a raven chick, a chicken)
 2. Make the creature friendly
 3. Attach a bag of any size and weight to the creature
 4. Consider if the creature is actually a misshaped pack-mule

### Expected behavior

Small, weak creatures can not hold large, heavy bags.

### Versions and configuration

- OS: Linux
    - OS Version: Distributor ID: Ubuntu; Description: Ubuntu 20.04.6 LTS; Release: 20.04; Codename: focal; 
- Game Version: 058d7a2 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    No Fungal Growth [no_fungal_growth],
    Bionic Slots [cbm_slots],
    Portal Storms Ignore NPCs [personal_portal_storms],
    Disable NPC Needs [no_npc_food]
]

RenechCDDA:
Confirming this.

`attach_bag_to`(link below) needs to check against monster::weight_capacity

https://github.com/CleverRaven/Cataclysm-DDA/blob/d2b90616077c1b6105b567ab7859d6d8be151abb/src/monexamine.cpp#L167-L192

PR #65901, opened 2023-05-28
Not attach the bag if too heavy

alef:
#### Summary
Bugfixes "#65448 Small monsters can double as pack-mules"

#### Purpose of change
#65448

#### Describe the solution
Simple check between total bag weight and monster's weight capacity, as suggested by @RenechCDDA

#### Testing
Spawned a chicken, an horse, a saddle bag, some bronze anvils.
Tested mounting the bag on the chicken or horse, before and after the changes.
Tested loading the anvils with bag mounted, or filling the bag before mounting.

#### Additional context
Did no read the proper weight values, just used some common sense.
Did not found a test for `attach_bag_to`.

### DIFF
--- a/src/monexamine.cpp
+++ b/src/monexamine.cpp
@@ -182,6 +182,10 @@ void attach_bag_to( monster &z )
     }
 
     item &it = *loc;
+    if (it.weight() > z.weight_capacity()) {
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):

Maleclypse:
Would you mind screenshotting the failure state to help reviewers looking over this?

PR #65902, opened 2023-05-28
Enable clang-analyzer checks

jbytheway:
#### Summary
None

#### Purpose of change
More static analysis.

The `clang-analyzer` checks do more complicated flow-control analysis.  This catches more subtle bugs but is also more prone to false-positives.

#### Describe the solution
Enable three `clang-analyzer` checks that were disabled in the move to LLVM16:

* `clang-analyzer-cplusplus.NewDeleteLeaks`
* `clang-analyzer-cplusplus.StringChecker`
* `clang-analyzer-optin.cplusplus.UninitializedObject`

These caused four warnings: two were genuine (failing to check the return value from `getenv` for being NULL); the other two seem to be false-positives which I've suppressed.

Disabling some of these checks for generating too many false-positives.

#### Testing
Run clang-tidy; unit tests.

#### Additional context

### DIFF
--- a/src/mapsharing.cpp
+++ b/src/mapsharing.cpp
@@ -96,8 +96,10 @@ void MAP_SHARING::setDefaults()
     MAP_SHARING::setCompetitive( false );
     MAP_SHARING::setWorldmenu( true );
     MAP_SHARING::setUsername( "" );
-    if( MAP_SHARING::getUsername().empty() && getenv( "USER" ) ) {
-        MAP_SHARING::setUsername( getenv( "USER" ) );
+    if( const char *user = getenv( "USER" ) ) {
+        MAP_SHARING::setUsername( user );
+    } else {
+        MAP_SHARING::setUsername( "" );
     }
     MAP_SHARING::addAdmin( "admin" );
 }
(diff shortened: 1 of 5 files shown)


src/mapsharing.cpp
```diff
@@ -96,8 +96,10 @@ void MAP_SHARING::setDefaults()
     MAP_SHARING::setCompetitive( false );
     MAP_SHARING::setWorldmenu( true );
     MAP_SHARING::setUsername( "" );
-    if( MAP_SHARING::getUsername().empty() && getenv( "USER" ) ) {
-        MAP_SHARING::setUsername( getenv( "USER" ) );
+    if( const char *user = getenv( "USER" ) ) {
+        MAP_SHARING::setUsername( user );
+    } else {
+        MAP_SHARING::setUsername( "" );
```
Qrox, line 102: Since it is already set to empty in line 98 I think this line is not needed?

src/mapsharing.cpp
```diff
@@ -96,8 +96,10 @@ void MAP_SHARING::setDefaults()
     MAP_SHARING::setCompetitive( false );
     MAP_SHARING::setWorldmenu( true );
     MAP_SHARING::setUsername( "" );
```
jbytheway, line 98:
```suggestion
```

PR #60275, opened 2022-08-18
Add messy and untidy house mapgen palettes

tjole:
#### Summary
Content "Adds new mapgen palettes to add various levels of untidiness to houses"

#### Purpose of change
The aim of this PR is to add variety and realism to houses by altering item spawns to give a more natural look, in some cases.

Currently, items spawns are very similar between houses. Items generally spawn in the place they belong, eg dishes spawn on tables, clothes spawn in dressers. In real life, many houses have varying levels of clutter, where items are not where they belong or storage is overfilled. For example, dirty dishes may be found on coffee tables, jackets may be thrown over the back of a chair, a laundry basket with clothes may have been put down somewhere and forgotten about, or a bookshelf might be stuffed with books and trinkets. In-game, all houses are very tidy and neat which is not reflective of real life, and can give a sense of uniformity.

#### Describe the solution
I have introduced an untidy and a messy palette, to be used with the standard_domestic_palette. The untidy palette aims to represent a house that is a bit cluttered, with some floor-spawns and some items spawning in places they don't belong. The messy palette aims to further this, adding some dirty item spawns and increasing the amount of out-of-place and floor spawns.

I am tentatively planning for the untidy palette to be used ~15% of the time with the messy palette being used ~5% of the time. This may change during testing, or from feedback.

#### Describe alternatives you've considered
Creating variants for each house, with specific item-spawns. This would be a lot of work, and doesn't make use of the palette functionality.

The palettes I've added are designed to be loaded after the standard_domestic_palette but I am having trouble getting multiple palettes to load (see comment). I have considered fleshing out these new palettes to not require the s_d_p but be loaded instead of it. I'd like to avoid this, however, as I think it leads to unnecessary code duplication.

I am opening this in draft to see if anyone has any ideas for loading multiple palettes.

#### Testing
This is still a WIP, and item groups and palette item spawns will continue to be tweaked.

I have used the debug overmap editor to spawn large numbers of different houses to test these palettes, and have made tweaks to the palettes to get closer to the desired effect. As I add the palette to houses, I will spawn a 3x3 grid of the house to test each palette. Finally, once I'm happy with the palettes and that they work in the houses they've been added to, I'll create a map and teleport around cities to ensure spawn rates for untidy/messy houses are not excessive.

#### Additional context

This is still a WIP and I will continue to change items groups and spawn rates - these screenshots are more concept than final result.

House33 - default palette, for comparison
house33_default

House33 - using the untidy palette
house33_untidy

House33 - using the messy palette
house33_messy

PR #65905, opened 2023-05-28
Residential-Decay

RanikOrtega:
#### Summary
Content "Mapgen changes"

#### Purpose of change
Somehow a few days into the Cataclysm buildings are still pristine even with riots and Zombies. Dirty up residential structures with Door and window damage. Blood, bile and body spawns as well as trash and graffiti. 

#### Describe the solution
Edited house-general_pallet and aux_nested files for small chance of décor spawning. Windows and doors have a chance to generate broken versions. Floor tiles in houses have a small chance for blood, bodies or trash to spawn settings relatively conservative currently. Could be adjusted higher or lower. Houses with dozens of zombies in / around them shouldn't be pristine. 

Leaving the mapgen as it was. 

#### Testing
Repeatedly generated maps and tested frequency of additions. 

#### Additional context
CDDA-Decay1
CDDA-decay2
CDDA-Decay3
CDDA-Decay4
CDDA-Decay5

### DIFF
--- a/data/json/mapgen/nested/aux_nested.json
+++ b/data/json/mapgen/nested/aux_nested.json
@@ -7,6 +7,15 @@
   },
   {
     "type": "mapgen",
+    "nested_mapgen_id": "blood_field",
+    "method": "json",
+    "object": {
+      "mapgensize": [ 1, 1 ],
+      "place_fields": [ { "field": "fd_blood", "x": 0, "y": 0, "intensity": [ 1, 3 ], "age": 10 } ]
+    }
+  },
+  {
+    "type": "mapgen",
     "method": "json",
     "nested_mapgen_id": "fire_field",
     "object": { "mapgensize": [ 1, 1 ], "place_fields": [ { "field": "fd_fire", "x": 0, "y": 0, "intensity": 2, "age": 10 } ] }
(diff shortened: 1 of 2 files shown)


Maleclypse:
@LovamKicsiGazsii I'd appreciate if you take a second to review this.

LovamKicsiGazsii:
Not much objectively wrong with this, except for the absence of glass shards near broken windows. That could be solved by using a modified version of `destroyed_glass_door_wall` in nested/aux_nested. Also, glass walls could use some bashing as well.

On a more personal note, I wish not all houses got vandalized the same way.  The fact that every house gets the same amount of destruction, the same amount of blood, trash, and corpses makes things look all too artificial.  I would prefer if the solution was implemented similarly to #60275, with houses having varying stages of destruction, starting from completely intact all the way to how it is in the screenshots.

Procyonae:
I agree, stages of destruction would be much more interesting, here's some doc links that might interest you https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/MAPGEN.md#mapgen-values https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/MAPGEN.md#place-rubble-and-smash-existing-terrain-with-rubble

RanikOrtega:
> @LovamKicsiGazsii  Let me have a look at 60275. I'm not at all attached to the current changes as much as I am to proper implementation of wear and tear on houses ESPECIALLY in Cities which should be very damaged. If it can be done better via 60275's methods then I may move it more towards that.

RanikOrtega:
> Not much objectively wrong with this, except for the absence of glass shards near broken windows. That could be solved by using a modified version of `destroyed_glass_door_wall` in nested/aux_nested. Also, glass walls could use some bashing as well.
> 
> On a more personal note, I wish not all houses got vandalized the same way. The fact that every house gets the same amount of destruction, the same amount of blood, trash, and corpses makes things look all too artificial. I would prefer if the solution was implemented similarly to #60275, with houses having varying stages of destruction, starting from completely intact all the way to how it is in the screenshots.

Not sure implimentation from #60275 is even working. Testing palette distribution between standard and hoarder versions isn't even working at the moment.  The following code seems to do nothing if changed. 

EX: "palettes": [
      {
        "distribution": [
          [ "standard_domestic_palette", 9 ],
          [ "standard_domestic_palette_abandoned", 280 ],
          [ "house_hoarder_palette", 40 ],
          [ "house_survivor_palette", 180 ]
        ]
      }
    ]

PR #64934, opened 2023-04-08
New guns, calibres, and conversion kits, fixed PR.

DoctorBoomstick:
#### Summary
Content "Migrates the changes from PR #64743. Adds 24 caliber conversion kits for various weapons, 13 new guns, 3 new calibres, and 3 professions, and migrates over the abandoned changes left behind in PR #63074."
#### Purpose of change
Following some rebase and merge errors, I had to close #64743, as it had become what could be generously described as a crap sandwich of errors and conflicts. This PR simply rescues and migrates over the files left behind in that PR. For the sake of ease of reference, the PR overview has been copied over.
Within the limits of real-world firearms’ use, caliber conversion kits for rifles (most notably the AR-15 platform) are rather prolific, with a number of such kits being in production and serving shooters within the remits of the roles they’re engineered for. This being said, such retool kits have been, for one reason or another, very thin on the ground within CDDA, with, at present, only a single .300 AAC Blackout kit existing in the whole game. This has led to guns like the abovementioned AR-15 rifles, as well as other arms like Glock pistols, not possessing the types of caliber swapping capabilities that their modular designs are prominently known for.
Personally, I believe that an effective implementation of such a mechanic could provide interesting tactical opportunities, as is the case with the Hub HWP. Apart from that, it could also assist in helping to elevate some firearms above their peers as far as individual utility goes and help them stand out from the current hordes of metal pue pue sticks that dump archaically named bullets into zombies.
Unrelated to caliber conversion kits, in PR #63074 I started to rework firearm damage modifiers in order to bring them up to speed with the currently established formula for determining such statistics. Due to various merge errors, rebase mistakes, and a large amount of IQ/coffee deprivation on my behalf, the entire PR got somewhat imploded following the release of 0.G, prompting me to shut it down.
#### Describe the solution
With respect to the caliber conversion kits, I have based their function on the current implementation of the HWP attachment system. Fundamentally, this involves having the base firearm itself simply be an empty frame, lacking any defined ammunition type, range value, damage amount, etc. Such values are only introduced with various removable and installable conversion kits that provide the weapons with their gun-related functionality.
[...]

PR #65918, opened 2023-05-29
Add Kel-Tec PLR16

CatWithFourEyes:
#### Summary

 Content "Add new gun, Kel-Tec PLR16"

#### Purpose of change

Add a new gun, compact rifle caliber pistol using common ammo, and trade some damage for small size.

#### Describe the solution

Add the gun, place it in items groups with probability similar to OA-93 in the same categories as AR Pistol, since PLR16 still in production. 

Finetune parameters
Change spawn chances
Not add it

#### Testing

Spawn gun, shoot from it, install and detach some mods, spawn gunstores to confirm it can be found. 

#### Additional context

Size and volume:
https://www.keltecweapons.com/firearm/pistols/plr16/
image
According to https://www.americanrifleman.org/content/kel-tec-plr-16-5-56-223-semi-automatic-pistol/ 
width is 1,7 inch (43mm)
Handguard based on Kel-Tec hanguard for PLR16 and needed to install some attachments (weight is added to gun).


data/json/items/gun/223.json
```diff
@@ -132,6 +132,69 @@
     ],
     "melee_damage": { "bash": 9 }
   },
+  {
+    "id": "plr16",
+    "copy-from": "rifle_semi",
+    "looks_like": "skorpion_61",
+    "type": "GUN",
+    "name": { "str": "Kel-Tec PLR16" },
+    "description": "Semi-automatic pistol, chambered in 5.56 NATO, using gas piston to reduce lenght and frame constructed from reinforced polymer to make the it both durable and light-weight. This one comes with plastic handguard.",
```
DoctorBoomstick, line 141:
```suggestion
    "description": "This is a semi-automatic pistol chambered in 5.56 NATO, using a gas piston to reduce length and a frame constructed from reinforced polymer to make it both durable and light-weight.  This one comes with a plastic handguard.",
```

data/json/items/gun/223.json
```diff
@@ -132,6 +132,69 @@
...
+          "stanag90",
+          "stanag100",
+          "stanag100drum",
+          "stanag150",
+          "survivor223mag"
+        ]
+      }
+    ],
+    "melee_damage": { "bash": 9 }
+  },
```
DoctorBoomstick, line 197: Could you also add a generic gun variant for use when the branded gun names option isn’t toggled to true? It would save some work later on when a clean sweep of the guns are made and more such variants are included. You can see how the Remington 700 in .270 Winchester does it for an example.

data/json/itemgroups/Weapons_Mods_Ammo/guns.json
```diff
@@ -189,6 +189,7 @@
       { "group": "nested_TDI_10", "prob": 4 },
       { "group": "nested_draco", "prob": 20 },
       { "group": "nested_ar_pistol", "prob": 25 },
+      { "group": "nested_plr16", "prob": 25 },
```
DoctorBoomstick, line 192: When I went looking on gun broker I found 1081 results for AR-15 pistols chambered in 5.56. this got narrowed down to 55 results when I searched for guns with a 7-inch long barrel as is represented in game. When I looked for PLR16s in 5.56x45mm, I only got 4 results. The numbers in the item groups should probably stand to be tweaked with these rough ratios in mind.

DoctorBoomstick (changes requested):

DoctorBoomstick:
You also need to add the gun’s id into the gg_firearms_migrationfile within the generic guns mod for it to pass tests.

data/json/items/gun/223.json
TheShadowFerret:
```suggestion
    "description": "A semi-automatic pistol, chambered in 5.56 NATO, using a gas piston to reduce length and a frame constructed from reinforced polymer to make the it both durable and lightweight. This one comes with a plastic handguard.",
```
We don't generally do start with "this" in item descriptions, as you're obviously looking at whatever the thing is.
(See "Item Description Cleanup Project" PRs by FoolsGold45)

TheShadowFerret:
This seems like it should be a lower instead of a complete firearm to use the caliber conversion kits from #64934.

DoctorBoomstick:
If I’m reading this right and the weapon’s meant to come attached with the polymer fore-end, shouldn’t it also have an under-barrel slot, unless that’s what the second rail slot is meant to represent?

> This fore-end also provides a second Picatinny rail below the barrel for attaching lights, lasers, rail-mounted bipods and similar accessories.

> Shouldn't this be a lower instead of a complete firearm to use the caliber conversion kits from #64934?


Honestly, I don’t know enough about the PLR to say if it could even accept AR-15 upper receivers.

CatWithFourEyes:
> Shouldn't this be a lower instead of a complete firearm to use the caliber conversion kits from #64934?

No. There is no changeable uppers, and i'm not sure, that polymer case is durable enough. 
Source: official user manual: https://media.keltecweapons.com/wp-content/uploads/2021/11/plr-16_manual-Optimized-1-1.pdf
For something like "customizable AR pistol" i remember Fightlite SCR Raider,  if you don't like buffer spring (like me).

data/json/itemgroups/Weapons_Mods_Ammo/guns.json
CatWithFourEyes: Check it, and i can't say it THAT bad ratio. ~Half of them in non intermediate cartige, and there is many 12+ barrel ones (not sure, what to say about 5-5.5 inches ones, they are more similar to OA-93 imo, and there is like 200+ of them). Also found 5 PLR16's, after searching by manufacturer. Looks like, there is not 1000 to 4, but ~500-600 to 5 or more in worst case.  
Also, maybe OA-93 would probably be the most appropriate example in terms of spawnrate? On GB it's 3 OA-93 and 1 OA-98, but since PLR16 still in production, it stays in same spawnpools as AR pistol?

CatWithFourEyes:
> If I’m reading this right and the weapon’s meant to come attached with the polymer fore-end, shouldn’t it also have an under-barrel slot, unless that’s what the second rail slot is meant to represent?

Reasonable. Added.

data/json/itemgroups/Weapons_Mods_Ammo/guns.json
CatWithFourEyes: Spawrate changed.

tenmillimaster:
> probability equal to AR Pistol

Can you please try to find its probability relative to other guns? Ar pistols are pretty common, keltec's weirder stuff is not.

I didn't find many hits on completed GunBroker.







Plr-16 weight = P50 * 14/476 
Or
Plr-16 weight= Sub2000 * 14/3869

You can try a few other hits to average the weights, but using the P50 seems good to me.

CatWithFourEyes:
> Can you please try to find its probability relative to other guns? Ar pistols are pretty common, keltec's weirder stuff is not.

Already changed with probability based on OA-93, with exeption, that PLR16 still in production.
https://github.com/CleverRaven/Cataclysm-DDA/pull/65918/files/db0c147abb8bfd5ddca961741c3314dae3738167#diff-1d7770ba5c6241f42a533e7474f9e4d8408cede560bbde6dcbddea8af5a113ec

tenmillimaster:
That's a little conservative, but I suppose it's OK.

PR #65932, opened 2023-05-30
Spell Event

### LINKED ISSUE #62488: [Magiclysm/Xedra Evolved] Training spells is **boring**

As the title says, it is hella boring to train your spells. The only valid level right now is to lock yourself somewhere with a food supply, and read books for a weeks, interrupting only for sleep and some food management. This is not okay, so here is the two ways that i see to resolve it, easy and hard one:

### Solution you would like.

Easy way:
Make spell cast give not the flat number of XP, but percentage amount. This will make training spell by casting them more useful, especially on the high level, but not as good at low level - so the main strategy would be "read book to some level, and then just use the spell in practice" the amount of xp may be discussed, but i believe 3-4% is okay value for start, but we definitely need a field for spell to defy it, so some spells may be easier to train by cast, and some is not

Hard way:
- Make an additional statistic, that each spell may track - amount of mana, spent on spell, amount of all damage, dealt by spell, amount of summoned items/creatures, length of all effects etc. Each time the spell reach some threshold, it gives a level (or some amount of xp, or some amount of percentage xp): spending 5000 mana on spell, cast spell 100 times, deal 5000 damage total, deal 1000 damage at once (for AoE), summon 50 creatures etc etc. I believe most of them may be automatic - taking a "mana spent", it may be "x15 of mana the spell uses, x50, x100, x200, x500, five levels in total - yet having ability to tweak it or at least turn it off for some spells would be nice. It may be bonded to unhardcoded/EoCed achievements, but it is hell of a work to do

_No response_

### Additional context

_No response_

bombasticSlacks:
#### Summary
Features "Spell casting now triggers an event"

#### Purpose of change
Resolves #62488

#### Describe the solution
Added new event type provides the vars in the docs

#### Testing
Unit tests provided


Maleclypse:
Looks like this needs a rebase

PR #65935, opened 2023-05-30
Allow NPCs to read E-books

Tread4:
#### Summary
Features "Allow NPCs to read E-books - new activity added"

#### Purpose of change
From now on friendly NPCs (followers) are able to read e-books from suitable devices. They can be ordered to do so in similar manner to reading books. A new dialogue option was added to handle e-readers.

It seemed to me that NPCs should be able to use advantages of e-reader the same way player does. Devices are widespread enough and scanning and copying a rare book so a group can read it at the same time seems reasonable use case IRL.

#### Describe the solution
- I have added a new dialogue option to NPCs that order them to read an e-book (same manner as existing read a book option).
- New inventory preset and menu were created to facilitate finding all E-readers in NPCs inventory and choosing the e-reader to use.
- After selcting of e-reader an existing e-book menu is started and allows selection of a book to read.
- If e-reader is broken or without battery charges it cannot be selected (proper message in selection menu).
- If e-reader stores no e-books a fitting pop-up shows up and exits menus.
- If it is too dark to read it is impossible to choose a book (all of them have proper text mentioning it)
- NPCs **DO NOT** turn on flashlight, screen etc. the same way player does (in order to read e-book) 
[this functionality can be added in the future - I took quite a bite (with my meager C++ skills) to code what is in this PR and by making it more complex I would fear choking on it ;) ].
- After e-book selction NPC proceeds to read a book as usual (using the same function he did with regular books)
- NPCs reading function was slightly adjusted (now it is called with bool parameter) to cater for choosing between regular book or e-book as reading material.

Not doing this PR

#### Testing
- I have reused a world with my "disassemble in the dark" test NPC
- Spawned a bunch of "e-reader" devices (smartphone, e-ink PC, laptop)
- Spawned a number of books
- Loaded different books to devices (some devices were left empty)
- Read book as PC (avatar) and confirm it still works as before
- Read e-book from a device as PC (avatar) and confirm it still works as before
- Gave NPC regular book, asked him to read it, confirm it works as before
- Gave NPC charged "e-reader" with some books and asked him to read
- Confirmed that NPC performs task as intended
- Gave NPC more devices (various types, including non "e-reader" electronics) ["e-readers" stored various books]
- Confirmed all three, checked "e-reader" device types worked and no other devices showed up in "Select e-reader" menu
- Confirmed that "e-reader" without battery cannot be used and player sees proper message
- Confirmed that broken (by water) "e-reader" cannot be used and player sees proper message
- Confirmed that "e-reader" with no books stored does not open "Select e-book" menu but escapes with proper pop-up
- Confirmed that NPC cannot start reading in the dark
- Confirmed that both "Select e-reader" and "Select e-book" menus can be escaped from (with esc) and close properly (had a crash here while working on PR but it seems I solved it)

#### Additional context
TODO (in oher PRs - I consider this one finished): 
- Allow NPCs to use in-built lighting/screen options in the same way player can
- I believe that NPCs reading e-books currently does not use device charges, I suppose it should be added. (ATM. I am afraid it is too big chunk to bite but I think it can be looked into in follow-up PR, any assistance here is also appreciated)

### DIFF
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -1267,8 +1267,8 @@ class ereader_inventory_preset : public pickup_inventory_preset
 
 item_location game_menus::inv::ereader_to_use( Character &you )
 {
-	const std::string msg = _( "You don't have any e-readers you can use." );
-    return inv_internal( you, ereader_inventory_preset( you ), _( "Select e-reader." ), 1, msg );					 
+    const std::string msg = _( "You don't have any e-readers you can use." );
+    return inv_internal( you, ereader_inventory_preset( you ), _( "Select e-reader." ), 1, msg );
 }
 
 class read_inventory_preset: public pickup_inventory_preset
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -1237,10 +1237,9 @@ class ereader_inventory_preset : public pickup_inventory_preset
 
         std::string get_denial( const item_location &loc ) const override {
             const item &it = *loc;
-            //const auto &uses = it.type->use_methods;
 
             if( it.is_broken() ) {
-                return /*string_format(*/ _( "E-reader is broken and won't turn on." ) /*)*/;
+                return _( "E-reader is broken and won't turn on." );
             }
 
             if( !it.ammo_sufficient( &you, "EBOOKREAD" ) ) {
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -1237,7 +1237,7 @@ class ereader_inventory_preset : public pickup_inventory_preset
 
         std::string get_denial( const item_location &loc ) const override {
             const item &it = *loc;
-            const auto &uses = it.type->use_methods;
+            //const auto &uses = it.type->use_methods;
 
             if( it.is_broken() ) {
                 return /*string_format(*/ _( "E-reader is broken and won't turn on." ) /*)*/;
--- a/src/game_inventory.cpp
+++ b/src/game_inventory.cpp
@@ -1240,13 +1240,13 @@ class ereader_inventory_preset : public pickup_inventory_preset
             const auto &uses = it.type->use_methods;
 
             if( it.is_broken() ) {
-                return string_format( _( "E-reader is broken and won't turn on." ), it.tname() );
+                return /*string_format(*/ _( "E-reader is broken and won't turn on." ) /*)*/;
             }
 
-            if( !it.ammo_sufficient( &you, uses.count( "EBOOKREAD" ) ) ) {
+            if( !it.ammo_sufficient( &you, "EBOOKREAD" ) ) {
                 return string_format(
-                           n_gettext( "Needs at least %d charge",
-                                      "Needs at least %d charges", loc->ammo_required() ),
+                           n_gettext( "Needs at least %d charge.",
+                                      "Needs at least %d charges.", loc->ammo_required() ),
                            loc->ammo_required() );
             }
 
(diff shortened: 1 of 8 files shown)


Night-Pryanik:
If you're actively working on the PR, consider converting it to a draft.

Tread4:
> If you're actively working on the PR, consider converting it to a draft.

I apologize if my activities cause any issue - I deem it finished, although tests have returned a few errors that I fix one by one now. Should I convert into draft until I see no errors related to my code?

I am fairly new and rather inexperienced when it comes to C++, Git and CDDA code. If I make a process mistake by not having it as a draft during the "bug stomping" please advise on good moment to convert to / from draft.

All advice is welcome and I apologize for any inconvenience caused.

PatrikLundell:
I'd recommend testing and bug stomping before creating the PR. It's normal to have people make comments that prompt changes and further testing (own or automated) to prompt additions(modifications, and I wouldn't put a PR into draft for these things, but I have marked PRs as draft when I've found a bad bug in it and want to make sure it's not merged before that was fixed.

Basically, if you're still working on it (including testing) and the PR is produced to receive comments I'd mark it as a draft, but if it's "done" with no changes except what comments would lead to I'd leave it as a "live" PR.

Note that this is my opinion, and it doesn't have any official weight.

Tread4:
> I'd recommend testing and bug stomping before creating the PR.

That is what I tried doing. I have uploaded this as PR once I was satisfied with functionality in game, had everything I thought of tested etc. and saw no bugs myself. Right now I am stomping bugs that were reported by our automated tests here on GitHub. I suppose there is a way to run tests in my local environment, but I do not know how (as I said, I am still a beginner).

Thank you for explanation and pointers on when to swap in and out of draft. I think I will keep my PRs as drafts as long as all required tests are not passed (and as you wrote, until I am satisfied with my code and see no bugs/pit-falls I accept).

Night-Pryanik:
> I apologize if my activities cause any issue

No issues on its own, but in general it's a good practice to put PR in a draft state if you're _actively_ working on it.

> I deem it finished

Then I recommend you to clean up your changes - for example, I saw some commented out code which you shouldn't include in your PR.

> I think I will keep my PRs as drafts as long as all required tests are not passed

The thing is, PRs in draft state don't trigger some of the tests, so it might be better to make all the work on your local machine and push a PR only when you really consider it finished. Some tests failing is not a reason on _per se_ to convert PR to draft, but if it's required to make some major changes, then converting to draft might be a thing to consider.

src/game_inventory.cpp
```diff
@@ -1218,6 +1218,59 @@ item_location game_menus::inv::gun_to_modify( Character &you, const item &gunmod
...
+            if( get_option<bool>( "INV_USE_ACTION_NAMES" ) ) {
+                append_cell( [ this ]( const item_location & loc ) {
+                    return string_format( "<color_light_green>%s</color>", get_action_name( *loc ) );
+                }, _( "ACTION" ) );
+            }
+        }
+
+        bool is_shown( const item_location &loc ) const override {
+            return loc->is_ebook_storage();
+        }
```
Night-Pryanik, line 1236: This isn't used anywhere?

Tread4: I "mimicked" some other "inventory_preset" class and in my understanding this part is responsible for marking which inventory items show or do not show on the related menu list.

I can be mistaken and can test locally what happens if I comment this part out.

Tread4: I did test it now - without this part  of the code "E-reader menu" shows incompatible items. Therefore this code is needed.
image

Maleclypse:
I spawned in an e-ink and then spawned in a first aid book and copied it into the e-reader.   I then spawned in an NPC and made them a follower, traded them the e-reader, after trading them the e-reader I was able to instruct them to read the ebook and they did so until they leveled healthcare to 1.

PR #65936, opened 2023-05-30
[Deadly Bites] Antivirals only spawn in medical contexts

dseguin:
#### Summary
None

#### Purpose of change
A little birdy informed me that antivirals were too easy to find :P

#### Describe the solution
Remove antivirals from item groups that spawn outside of medical contexts. You can still find them in pharmacies, hospitals, doctor's offices, and certain labs

#### Testing
Loaded the game and went spelunking for antivirals. No more house bathroom finds

### DIFF
--- a/data/mods/deadly_bites/itemgroups.json
+++ b/data/mods/deadly_bites/itemgroups.json
@@ -1,49 +1,9 @@
 [
   {
-    "id": "SUS_bathroom_medicine",
-    "type": "item_group",
-    "subtype": "collection",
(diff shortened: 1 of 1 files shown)


AcidAntOnAMinefield:
What is considered an anti-viral in CDDA? Were there ever anti-viral medications in the game, aside the Flu-Shot?

Because if you are considering antibiotics (narrow, broad-spectrum, etc) as "antivirals", that is wrong - antibiotics and antivirals are two completely different things.

PR #65940, opened 2023-05-30
Animal Feet mutation.json

GalacticApple:
#### Summary
Content "Adds the Animal Feet mutation which restricts footwear and reduces carry-weight"

#### Purpose of change
I made this for myself, but I thought others might like it. I didn't want this to just be an aesthetic thing, so I made it reduce the player's carry-weight since having a bend in your legs would make it more difficult to lift stuff. I also gave the Beast category the Padded Feet mutation so they can get this. 

#### Testing
I used this on my personal copy of CDDA, so it should work.


kilozombie:
Surely more informative to write it as "digitigrade legs"

PR #65941, opened 2023-05-30
[ Xedra Evolved ] Banishes are dreamer spells instead of eater spells

Maleclypse:
#### Summary
None

#### Purpose of change
Moves the banish spells to the dreamer class instead of the eater class.  They make more thematic sense as the opposite of summons which is the purview of the dreamer class.

#### Describe the solution
Move json

None


adamkad1:
One one hand, summoners would likely know how to banish summons, but on another hand, so would anti-dreamers

PR #65942, opened 2023-05-31
Digitigrade mutation

GalacticApple:
#### Summary
Content "Adds the Digitigrade Legs mutation which increases crouching move-speed and restricts footwear."

#### Purpose of change
I made this for myself, but I thought others might like it. I didn't want this to just be an aesthetic thing, so I made it give a movement bonus for crouching (stacks with Paws) while also restricting carry weight. I also gave the Beast category the Padded Feet mutation so they can get this (along with Lupine and Feline).

It might make sense to have this be a requirement for some of the other feet/leg mutations. It could be something like 
Padded_Feet -> Strong_Legs -> Animal_Feet/Rabbit_Feet -> Leaping_Legs/Feline_Leap.

#### Testing
It works on my personal copy of CDDA, so it should be fine. 

#### Additional context
The carry-weight penalty turns into a slight bonus if you are crouching without anything in hand. If you have PAWS, then the bonus is applied if you are running as well.


kilozombie:
Still with the same name?

Also, just a cursory thought, it's not hard to imagine this being a prerequisite for all kinds of mutated abilities like the Feline Leap, since things that require animalistic digitigrade legs would provide a justification for this otherwise-detrimental trait.

kilozombie:
All good, git always takes a while to get back into for me. Nuking everything and restarting is a very legitimate tactic if things go wrong, don't let anyone tell you otherwise 😆

Daved27hundred:
Can I suggest the name "digitigrade legs" instead? that's the term for those kinds of feet.

CatWithFourEyes:
Bears and _spiders_ are not digitigrade. In my opinion, this mutation is more about "your leg have some common traits with your mutation line and not harmed by walking on rough surfaces without boots, while boots become a hindrance (10% speed bonus without them)", and not "this is a specific type of leg/locomotion style", leaving to player imagination what exactly does their character look like.
And carry weight penalty... Any good sources about it? Digitigrades are mostly about energy efficiency while running, and plantigrades have some advantages with walking. Mostly, thats all i can find.

CatWithFourEyes:
> but there is a reason human legs are straight

Because human ancestors was unguligrade apes, and they locomotion type was much closer to plantigrade, in addition to the need to stand upright to see farther and save energy during a long movement. In theory, digitigrade mutant must be somewhat shorter, than human, or his legs are _very long_ (or angle in not so big, and very likely allow to use "locking knee" feature like humans to reduce energy consumption while standing)

Most drawbacks of digitigrade posture, like greater need for food or less time of daily activity are already in game - Beast have "Very Fast Metabolism" and Feline's are "Very Sleepy".

Also carry weight penalty will lead to a very controversial situation when mutant, moving on all four limbs ( https://cdda-guide.nornagon.net/mutation/PAWS ) can carry LESS then human walking on only two legs. And carry weight penalty for mutants in strenght-oriented trees (Lupine and Beast) is illogical in my opinion. 

From ingame perspective, this change will turn one of beneficial mutations in a very detrimental one. And technicallyBird and Raptor are digitigrades too. So, even more penalty to carry weight for bird?

Against this idea. Current realisation isn't great, but this isn't better.

CatWithFourEyes:
> If nothing else, this would reduce their effective carry-weight while standing upright.

Arguable point, that this is the consequences of less footprint:
https://www.researchgate.net/figure/Redistribution-of-the-weight-of-the-body-on-the-plantar-area-of-the-foot-a-in-case-of_fig1_325359275 - humans use not so much of foot area contacting ground. BUT they have two "point of pressure" - first is heel, second near toes.
Most obvious downside is "knockdown vulnerability" (maybe rollers can be source of effect for it?), for effectiveli having one, since heel is off-ground.

fairyarmadillo:
I'm not sure what this is adding to the game, and it makes no sense that being a skater would make you better at walking around on dog feet.

kilozombie:
It isn't strictly necessary for mutations to have negative effects, or even any effects at all. I'm not saying that it's a bad idea to make this mutation do something, but as flavor it works just fine.

data/json/mutations/mutations.json
```diff
@@ -3901,6 +3901,49 @@
...
+    "type": "mutation",
+    "id": "ANIMAL_FEET",
+    "name": { "str": "Digitigrade Legs" },
+    "points": -1,
+    "visibility": 2,
+    "ugliness": 2,
+    "description": "The heels of your feet have shifted up off the ground in a rather animalistic fashion.  While *some people* might think this looks cool, it mostly just means you can't wear shoes.  At least you can move faster when crouching.",
+    "types": [ "FEET" ],
+    "prereqs": [ "PADDED_FEET" ],
+    "category": [ "LUPINE", "FELINE", "BEAST" ],
```
Drew4484, line 3914:
```suggestion
    "category": [ "LUPINE", "FELINE", "BEAST", RABBIT ],
```

Drew4484:
> hold up, wouldn't including RABBIT as a category cancel out the RABBIT_FEET mutation? both are "FEET" types.

I'd go the route of Plantigrade Legs>Digitigrade Legs>(Specialized Leg/Foot)

Drew4484:
In consideration of the previous discussion on bonus/malus, a reduced (~0.8) carry weight would be reasonable. This isn't a gameplay balance issue, rather a mechanical issue. Standing on only the two digitigrade legs is a lot of work, which is why we don't see many non-avian digitigrade animals walking on their hind legs for extended periods of time. If you really wanted to get into the weeds, you could remove that carry weight penalty while they're on all fours as that might even provide a carry weight bonus.

data/json/mutations/mutations.json
```diff
@@ -3901,6 +3901,62 @@
...
+      {
+        "condition": {
+          "and": [
+            { "u_has_move_mode": "run" },
+            { "not": "u_can_drop_weapon" },
+            { "u_has_any_trait": [ "THRESH_LUPINE", "THRESH_FELINE", "THRESH_BEAST" ] },
+            { "u_has_trait": "PAWS" }
+          ]
+        },
+        "values": [ { "value": "CARRY_WEIGHT", "multiply": 0.35 } ]
```
Drew4484, line 3956: Does this still give the basic run speed modifier?

Drew4484: Since the bipedal mode gets a 15, maybe a 10 here.

Drew4484 (approved):

CatWithFourEyes:
3rd x0.8 carry weight penalty for Rabbit, without anyting in return is intentional?

data/json/mutations/mutations.json
```diff
@@ -3954,6 +3954,79 @@
...
+    "enchantments": [
+      {
+        "condition": {
+          "and": [
+            { "u_has_move_mode": "crouch" },
+            { "not": "u_can_drop_weapon" },
+            { "u_has_any_trait": [ "THRESH_LUPINE", "THRESH_FELINE", "THRESH_BEAST" ] }
+          ]
+        },
+        "values": [ { "value": "MOVE_COST", "multiply": -0.15 }, { "value": "CARRY_WEIGHT", "multiply": 0.35 } ]
```
Maleclypse, line 3998: So this reduces your carry_weight by 65%? Am I reading that right?

fairyarmadillo: The fact that you explained that and I still don't understand it, and I have no idea what happens if the character has this mutation but not PAWS (which will be a pretty common situation) suggests to me that you might have overthought this mutation. Players should be able to understand what is happening and how when something this intrinsic to their character is being changed.

fairyarmadillo: Then you probably want to remove this line: "At least you can move faster when crouching." from the description.

fairyarmadillo: If you don't want to leave it blank, I would suggest clarifying it even a little more - "These are a clumsy match for human arms, but for someone with paws up front, they'd be quite helpful."

fairyarmadillo:
Why did you change the description back?

fairyarmadillo:
What are you talking about with mindset? Why is it so long? What I wrote would have been fine, as would leaving it blank. This mutation by itself has nothing to do with walking on all fours, it's only set to confer a small benefit to the existing paws quadrupedalism, which the player will almost certainly not see before they get this mutation.

You're making it needlessly confusing by adding a bunch of information which is irrelevant to inexperienced players and which experienced players will already be aware of.

fairyarmadillo:
"description": "The heels of your feet have shifted up off the ground in a rather animalistic fashion, reducing your carried weight limit and preventing you from wearing shoes.  These hindfeet are a bad match for human arms, but would pair well with matching forepaws."

fairyarmadillo:
Mutations are always either considered ugly or ignored by NPCs. A line about some people thinking they're cool is misleading considering how damaging visible mutations are to your social interactions.

PR #65948, opened 2023-05-31
Simply Grits

LordBarkBread:
Adds in Grits and Cheese Grits.

#### Summary
Content "Adds Grits"

#### Purpose of change
Grits is a easy thing, this adds in the simple two recipes of making grits from cornmeal and water (and a version with some cheese).

#### Describe the solution
Added in Grits food item and recipes to make them.

Not having grits and just eating oatmeal like a Cataclysm peasant.


Knut-Aage-Hofseth:
I was under the belief that unless one is talking about instant grits it takes more than 5 minutes to cook grits, but that the recipe takes about the same amount of time for a large batch as for a single serving if it is cooked all at once in a large pot.

fairyarmadillo:
> I was under the belief that unless one is talking about instant grits it takes more than 5 minutes to cook grits, but that the recipe takes about the same amount of time for a large batch as for a single serving if it is cooked all at once in a large pot.

https://www.bonappetit.com/recipes/article/grits

You appear to be correct. Bon appetit says you're looking at like 40 minutes if you're just boiling cornmeal on the stove.

Issue #64972, opened 2023-04-10
Freshly dug shallow pits outside viewing range glitch


NetSysFire:
### Describe the bug

![image](https://user-images.githubusercontent.com/59517351/230807972-64954898-1356-4aa3-8190-7da17941d7a5.png)

Not a new bug but I finally got around to reporting it. Tileset is RetroDays, but this is not tileset specific and looks the same in e.g Ultica. When going outside to look at the pits, they do look correctly.

### Attach save file

[Playtesting-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11186872/Playtesting-trimmed.tar.gz)

Note: If you can safely ignore any errors about civilian monsters that do not exist because as the save file name says, I have been playtesting a PR of mine.

### Steps to reproduce

1. Dig some pits outside a building, e.g the evac shelter.
2. Go inside, so the pits will remain in map memory only.
3. They corrupt, but only freshly dug ones?

### Expected behavior

They look the same as when they are actively in your line of sight.

Minor graphical bug. Not gamebreaking or too annoying.

### Versions and configuration

- OS: Linux
    - OS Version: LSB Version: n/a; Distributor ID: Arch; Description: Arch Linux; Release: rolling; Codename: n/a; 
- Game Version: 0.G [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

TealcOneill:
@GuardianDll Can't replicate, looks like it was fixed by #65954.

NetSysFire:
I saw this just recently in a run of mine. I am still not sure of the exact replication steps but I am fairly sure its still there.

TealcOneill:
I'll contact you in the future! Thank you.
That's interesting, I wasn't able to replicate this issue with your save or a new one of my own. I don't know why it wouldn't happen in my saves.

- OS: Windows
    - OS Version: 10.0.19045.4291 (22H2)
- Game Version: cdda-experimental-2024-05-15-1833 7f06e9f [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
[...]

PR #65954, opened 2023-05-31
Improve map memory visuals

irwiss:
#### Summary
None

#### Purpose of change

Make map memory remember terrain separately from furniture, traps and vparts; makes aesthetics more consistent, since drawing e.g. a trap or vehicle part from map memory is doing so on an ugly fallback tile.

#### Describe the solution

PR makes map memory remember both the terrain and the "decoration" (either of furniture/trap/vpart) for each tile.

Old map memory will be loaded but will have the ugly fallbacks until player gets sight on each tile.

#### Testing

For curses nothing "should" change except a bit more memory used, as it only draws one symbol per tile.
For tiles visually:
Smashbutton is barely changed
Ultica_iso/Ultica/MSXotto start rendering terrain under vehicles/furniture/traps

Render time is almost unaffected - or at least the noise of gpu regulating itself had more effect than the difference (tested with vsync flag compiled out):
Before patch 5 seconds: 899 frames / 180 fps average
Before patch 60 seconds: 10797 frames / 179 fps average
After patch run 1 5 seconds: 886 frames / 177 fps average
After patch run 1 60 seconds: 10755 frames / 179 fps average
After patch run 2 5 seconds: 923 frames / 184 fps average
After patch run 2 60 seconds: 11054 frames / 184 fps average

The change in memory usage seems to be swallowed in the noise of all the other allocations, worst case the map chunks can be dumped to disk and discarded from memory more often if it becomes a problem

#### Additional context

Before (notice black splotches on wall wires, vehicle parts, flower/rock "furniture", less prominent but also the tables/shelves inside the shelter
pre
After:
post

Another example, street lights:
image
image

### DIFF
--- a/src/map_memory.h
+++ b/src/map_memory.h
@@ -9,84 +9,48 @@
 #include "memory_fast.h"
 #include "point.h" // IWYU pragma: keep
 
-class JsonOut;
 class JsonObject;
 class JsonOut;
 class JsonValue;
 
-struct memorized_terrain_tile {
-    std::string tile;
-    int subtile;
-    int rotation;
+class memorized_tile
+{
+    public:
+        std::string tile;
+        int subtile = 0;
+        int rotation = 0;
 
-    inline bool operator==( const memorized_terrain_tile &rhs ) const {
-        return ( rotation == rhs.rotation ) && ( subtile == rhs.subtile ) && ( tile == rhs.tile );
-    }
-
-    inline bool operator!=( const memorized_terrain_tile &rhs ) const {
-        return !( *this == rhs );
-    }
+        bool operator==( const memorized_tile &rhs ) const;
+        bool operator!=( const memorized_tile &rhs ) const {
+            return !( *this == rhs );
+        }
 };
 
 /** Represent a submap-sized chunk of tile memory. */
 struct mm_submap {
     public:
-        static const memorized_terrain_tile default_tile;
+        static const memorized_tile default_tile;
         static const int default_symbol;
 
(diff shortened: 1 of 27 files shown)


src/map_memory.h
```diff
@@ -9,87 +9,51 @@
...
-        }
-
-        inline int symbol( const point &p ) const {
-            if( symbols.empty() ) {
-                return default_symbol;
-            } else {
-                return symbols[p.y * SEEX + p.x];
-            }
-        }
+        // @returns true if mm_submap not empty; empty submaps are skipped during saving.
```
Brambor, line 41:
```suggestion
        // @returns true if mm_submap is empty; empty submaps are skipped during saving.
```

Fris0uman:
> Not really alternatives, reminders for possible future PRs:
Tripoints should be converted to tripoint_abs_ms
Using an actual compressor for map memory files in addition/instead of RLE will likely give better gains
Some of the slack can be cut down - e.g. a 4 byte rotation int, that only contains 0-360 angle in integer degrees, and even just 4 90-degree intervals could be compressed to a byte or less, same with subtile which only has a couple dozen values.
Compressed rotation/subtile to int8_ts
Compressed subtile to int8_t
Vehicle part rotations converted to fit all other rotations
Terrain ids now use string_ids so the strings might get interned, fingers crossed

This should probably go into an issue instead to not be lost

NetSysFire:
This looks like it would fix #64972, but I can not test right meow.

PR #65966, opened 2023-06-01
[bombastic perks] Perk Fixes

bombasticSlacks:
#### Summary
None

#### Purpose of change
Small oopsie about games in progress.


adamkad1:
Does the mod have ability to reduce the xp requirement? Last i've seen the options only increased, which sucks when you misstap or smth

bombasticSlacks:
You should try the mod it would answer your questions 🙂

PR #65969, opened 2023-06-02
Reduce Helmet DPS

Maleclypse:
#### Summary
None

#### Purpose of change
The average helmet was assumed to add 10 bash to whatever your base melee damage numbers worked out to.  Some added 20 bash.   Helmets can be swung but not very effectively or you'd read about soldiers acting like our npcs and dropping other weapons for them.  They were ridiculous and this PR attempts to make this more sane.

#### Describe the solution
Reduce all helmet damage by 80% or more.

None


Ruminator24:
Is this why npcs would drop their weapons for helmets? Maybe they were onto something...

jeffplata:
Hopefully, this would stop them from their helmet craze. Belts next?

PR #65980, opened 2023-06-03
math_parser: improve handling of dialogue functions

andrei8l:
#### Summary
None

#### Purpose of change
Dialogue functions in the math parser are currently a bit of a hack that circumvents most of the parser and so they don't support almost any of its features.

Some EOC functions can't be ported cleanly without optional named parameters.

#### Describe the solution
Integrate dialogue functions more tightly into the math parser:
- allow doubles and math sub-expressions to be passed as parameters, instead of just strings and variables
- add support for optional named parameters (kwargs)

An example with the new features:
```JSON
"condition": { "math": [ "u_monsters_nearby( 'radius': sin(u_somevar * pi/2) + 5, 'location': some_loc_var )", ">", "10" ] }
```

Also did a bit of housekeeping on the diag interface that should reduce churn in the future:
- dialogue function declarations on the C++ side now use a typedef
- functions are stored in a map

Instead of real kwargs I could have promoted string processing similar to the existing processing of string tags like `<global_val:blorg>`: I'd rather have all the parsing done in one place with one codebase. Also, putting a parser in a parser in a parser is a little too much even if you really like parsleyrs.

#### Testing
Updated test unit with that provides 95% coverage for `math_parser`.

Also fuzzed for about a day with this fuzz target https://github.com/andrei8l/Cataclysm-DDA/commit/869e7246cc8d56fc7056ddbaaa3b9daa636466ec.

#### Additional context
I also ported the `u_val` shim to real kwargs, but decided against including it in order to reduce churn. I will instead focus on porting the few EOC functions that need kwargs ASAP, then remove the fake kwargs from the shim entirely.

### DIFF
--- a/data/mods/BombasticPerks/corefiles/welcomemenu.json
+++ b/data/mods/BombasticPerks/corefiles/welcomemenu.json
@@ -20,7 +20,7 @@
       },
       {
         "text": "Custom Value",
-        "effect": [ { "math": [ "u_exp_to_perk", "=", "num_input('Exp until first level?', '100')" ] } ],
+        "effect": [ { "math": [ "u_exp_to_perk", "=", "num_input('Exp until first level?', 100)" ] } ],
         "topic": "TALK_PERK_MENU_WELCOME_EXP_INCREMENT"
       },
       { "text": "Back", "topic": "TALK_PERK_MENU_WELCOME" }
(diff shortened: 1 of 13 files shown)


andrei8l:
Thanks!

PR #65984, opened 2023-06-03
Fix throw miss-hitting a critter causes crash

### LINKED ISSUE #65834: Throwing items can cause crash
### Describe the bug

Throwing an item past an enemy causes a segmentation fault CTD.
When throwing an item at an enemy, it works as expected an nothing really happens. Though, if you aim just past the enemy, the game freezes and crashes with a segmentation fault if it happens to hit the monster/npc. If the item misses, it lands on the floor behind them as expected. This was tested using several different items on both a monster and npc.

### Attach save file

Attached below with crash logs

### Steps to reproduce

Find an npc/monster
Find a rock or something similar
Aim one tile past the npc/monster
Throw item, and repeat until crash.

### Expected behavior

Item should either miss, or impact and resolve as if you had aimed at the creature.

### Screenshots

_No response_

OS: Windows 10 x64
Version: 09b378d x64 
Graphics: Tiles
Language: English
Mods: DDA

### Additional context

Zip file containing a testing save and relevant .log files attached below.
[Throwing Crash.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/11554029/Throwing.Crash.zip)

SurFlurer:
#### Summary
none

#### Purpose of change
Fix #65834

#### Describe the solution
Validate critter.

#### Testing
Crash fixed.

### DIFF
--- a/src/ranged.cpp
+++ b/src/ranged.cpp
@@ -1325,11 +1325,13 @@ dealt_projectile_attack Character::throw_item( const tripoint &target, const ite
                                            this, nullptr, wp_attack );
 
     const double missed_by = dealt_attack.missed_by;
-    if( missed_by <= 0.1 && dealt_attack.hit_critter != nullptr && !critter->has_flag( MF_IMMOBILE ) ) {
+
+    if( critter && dealt_attack.hit_critter != nullptr && missed_by <= 0.1 &&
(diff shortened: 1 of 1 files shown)


SurFlurer:
Thanks for merging.

PR #65988, opened 2023-06-03
Allow large characters through windows.cpp

GalacticApple:
#### Summary
Balance "Allow Large characters to fit in windows"

#### Purpose of change
Large characters can fit in cars, wear human clothes and use human equipment. However, they cannot fit through windows. While windows can come in many shapes and sizes, I would say most are about as wide as your average door, so even if the opening is smaller than a door, such characters should still be able to pass through them (though perhaps with more difficulty). 

#### Describe the solution
I increased the creature size threshold from "medium" to "large" for allowing passage through windows. 

Maybe the move cost should be increased when passing through.

#### Testing
Works on my own machine.

#### Additional context
Large monsters and Huge characters still won't fit.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -10134,7 +10134,7 @@ std::vector<std::string> game::get_dangerous_tile( const tripoint &dest_loc ) co
 bool game::walk_move( const tripoint &dest_loc, const bool via_ramp, const bool furniture_move )
 {
     if( m.has_flag_ter( ter_furn_flag::TFLAG_SMALL_PASSAGE, dest_loc ) ) {
-        if( u.get_size() > creature_size::medium ) {
+        if( u.get_size() > creature_size::large ) {
             add_msg( m_warning, _( "You can't fit there." ) );
             return false; // character too large to fit through a tight passage
         }


Venera3:
Windows tend not to be full-height ones in most houses. I'm also not clear on why large monsters don't fit while characters of the same category do, since this is a creature-level check.

fairyarmadillo:
Being large should have more gameplay impact, not less. Squeezing through a window is difficult for a human, a giant should either have difficulty or be unable to do it.

fairyarmadillo:
Windows are much smaller than doors, which is why they already have a move cost to get through, so that argument doesn't really hold up. A person with the large trait is generally going to be between seven and eight feet tall and, per the description, bulkier than a human of average size. Given that the average double-hung window is only 36 inches wide, and that many windows only open halfway, we're dealing with an aperture that is often quite small. Unmutated people average around 14-16 inches in the shoulder, so squeezing through windows is not out of the question for them, but if you're an eight foot post-thresh bear mutant, no I don't think you're getting in. Remember that Large is not purely a height difference, the description of inconveniently large (the prereq) says that you're also much broader.

You seem to be selectively applying logic or comparing unlike things. A car is not like a window. A doorway is not like a window. A large zombie is of a similar size to a large survivor, and generally zombies are a lot more determined to get places than survivors are, even if they're a bit clumsier, and as discussed, not everything large is a zombie.

Large is also a highly advantageous mutation that commonly appears on powergamey mutation lists (see https://www.reddit.com/r/cataclysmdda/comments/118on78/a_guide_to_making_the_strongest_mutant_in_0g/ ). I don't think its few drawbacks should be removed. The point of the mutation is that you are bigger and stronger but that's coming at a cost, and the existing logic is more sound than this PR.

An alternative I might suggest would be to allow large characters (but not huge) to travel through empty window frames, but not open windows. This would need to be extended to large (but not huge) enemies as well.

Zireael07:
Part of the problem is that CDDA tiles are so very vague.... usually they're 5x5ft IIRC but sometimes not, and heights are even more vague

PR #65995, opened 2023-06-03
sprayable chloramine actually uses bleach/ammonia

anoobindisguise:
#### Summary
Bugfixes "Chloramine is made of liquid cleaner instead of powders"

#### Purpose of change
Sprayable toxic gas is chloramine, made of ammonia hydroxide and bleach, but the in game recipe was made of powdered lye and "oxidizer powder" (which is an ambiguous old material). As far as I can tell mixing powders wouldn't produce a toxic gas. 

#### Describe the solution
Previously the recipe used 125 ml each of oxidizer and lye powder. Now it's bleach and ammonia solution, 1 each (each unit is 0.125 liters as a gallon jug has 30)

- Liquid lye and bleach? The ratio would need to be bumped up because lye liquid is 0.25 liters.
- Liquid Ammonia and bleach? (So actual chlorine gas instead of chloramine, to justify the lethality of poison gas clouds against mobs)
- Is the surface heat needed? you're just pressurizing some gas and then spraying it and you don't need to hotplate the bleach and ammonia to get the gas.

### DIFF
--- a/data/json/recipes/recipe_ammo.json
+++ b/data/json/recipes/recipe_ammo.json
@@ -875,10 +875,10 @@
       [ "recipe_labchem", 3 ],
       [ "textbook_anarch", 3 ]
     ],
-    "proficiencies": [ { "proficiency": "prof_intro_chemistry" }, { "proficiency": "prof_inorganic_chemistry" } ],
+    "proficiencies": [ { "proficiency": "prof_intro_chemistry" } ],
     "qualities": [ { "id": "CHEM", "level": 1 } ],
     "tools": [ [ [ "surface_heat", 5, "LIST" ] ] ],
-    "components": [ [ [ "oxy_powder", 100 ] ], [ [ "lye_powder", 100 ] ] ]
+    "components": [ [ [ "bleach", 1 ] ], [ [ "ammonia_hydroxide", 1 ] ] ]
   },
   {
     "type": "recipe",
--- a/data/json/recipes/recipe_ammo.json
+++ b/data/json/recipes/recipe_ammo.json
@@ -877,7 +877,6 @@
     ],
     "proficiencies": [ { "proficiency": "prof_intro_chemistry" } ],
     "qualities": [ { "id": "CHEM", "level": 1 } ],
-    "tools": [ [ [ "surface_heat", 5, "LIST" ] ] ],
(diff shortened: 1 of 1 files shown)


data/json/recipes/recipe_ammo.json
```diff
@@ -875,10 +875,10 @@
       [ "recipe_labchem", 3 ],
       [ "textbook_anarch", 3 ]
     ],
-    "proficiencies": [ { "proficiency": "prof_intro_chemistry" }, { "proficiency": "prof_inorganic_chemistry" } ],
+    "proficiencies": [ { "proficiency": "prof_intro_chemistry" } ],
     "qualities": [ { "id": "CHEM", "level": 1 } ],
     "tools": [ [ [ "surface_heat", 5, "LIST" ] ] ],
```
Maleclypse, line 880:
```suggestion
```
Reaction happens at room temperature.  No heat needed.

randomtyper:
> * Liquid lye and bleach? The ratio would need to be bumped up because lye liquid is 0.25 liters.

I am all for alternative recipes. Good to have multiple ways of solving a problem. That said, given chloramine is a gas, shouldn't the recipe require something to put it into some container with? Like an air pump. And gas protection. In fact, I am really sceptical about handling gasses in survival conditions unless it is done with immediate dissolution of the gas in question into some liquid. And if survivor is making liquid, then it is not sprayable gas, but sprayable toxic liquid. Or aerosol. And as you said, chloramine's lethality is... Questionable. 

Also, per Wikipedia: 

"Chloramine is a highly unstable compound in concentrated form. Pure chloramine decomposes violently above −40 °C (−40 °F). Gaseous chloramine at low pressures and low concentrations of chloramine in aqueous solution are thermally slightly more stable. Chloramine is readily soluble in water and [ether](https://en.wikipedia.org/wiki/Diethyl_ether), but less soluble in chloroform and carbon tetrachloride".

randomtyper:
As for actual chlorine, there are simpler ways to obtain it than through rare ammonia (that is pretty hard to work with anyways imo). 

Simple Wiki check will say that electrolysis is a much more widespread method, and likely much more accessible to survivor, given that it does not involve handling ammonia. And it produces lye.

https://en.wikipedia.org/wiki/Chlorine#Occurrence_and_production

kevingranade:
Looks like electrolysis of salt water was the production method used in WWI, so yea that's a good starting point, though yield is going to be pretty bad.

PR #65998, opened 2023-06-03
NPC shooting improvement

Maleclypse:
#### Summary
Bugfixes "Improving NPC shooting frequency"

#### Purpose of change
Making use of KheirFerrum's work on NPC shooter AI and taking his suggestion to pull this line over.  This appears to improve NP shooting frequency anecdotally. 

#### Describe the solution
I've copied this change in so that NPC's increase the amount they shoot.

Waiting on other people to do these things.

#### Testing
I loaded in several bandit locations and wandered to them seeing if NPC's would shoot at me.  It happened very rarely.   After change compiled I went through the same three bandit sites.  PCs were shot to death repeatedly.  

Followed KheirFerrum's suggested method for testing and observed significantly less "couldn't decide on an action" messages.

### DIFF
--- a/src/npcmove.cpp
+++ b/src/npcmove.cpp
@@ -2074,9 +2074,7 @@ int npc::confident_shoot_range( const item &it, int recoil ) const
     if( !it.is_gun() ) {
         return res;
     }
-    if( confident_range_cache ) {
-        return *confident_range_cache;
-    }
+
     for( const auto &m : it.gun_all_modes() ) {
         res = std::max( res, confident_gun_mode_range( m.second, recoil ) );
     }


KheirFerrum:
You can confirm this change by turning on debug mode and checking the logs. Prior to this change what you should typically see is a lot of "NPC cannot decide what to do!" leading into "NPC has chosen action melee.". After this change instead you should see "NPC has chosen action aim" or similar lines.

Maleclypse:
> You can confirm this change by turning on debug mode and checking the logs. Prior to this change what you should typically see is a lot of "NPC cannot decide what to do!" leading into "NPC has chosen action melee.". After this change instead you should see "NPC has chosen action aim" or similar lines.

Thank you I'll try that when I get back to a computer tonight :)

Issue #63586, opened 2023-02-14
Slowdown when crafting items with lots of surrounding items in sight.


CoroNaut:
### Describe the bug

When crafting objects with none, or very few items in view range around the player, there is little to no process slowdown. Essentially the time it takes to pass 5 in-game minutes is for example, 1 second real-time. Crafting objects while there are lots of items, 500+ in sight (inside the "list all objects around player" menu, ctrl-v) will essentially add real-time to the crafting. So 5 in-game minutes will become 2, 3, 5 seconds real-time. 

I believe this is due to the crafting process needing to continually check every in-game second if the required materials are in range. If I however move all the required tools and materials around a wall, such that these 500+ items are out of view range, but still in the same map chunk, the crafting takes much less real-time.

For crafts taking more than 5 hours in-game, it can take a while as you see the crafting progress tick up 1% or less every real second.

### Attach save file

N/A. Anyone can plop down lots of loot and see the performance slowdown.

### Steps to reproduce

1. collect and store lots of items in one small area, perhaps 50 tiles over the course of a playthrough.
2. try to craft something that takes longer than 30 minutes while in sight range of all your exploits
3. try to craft that same item but take all the required tools and materials around a corner
4. compare real-time to complete crafts between steps 2 and 3.

### Expected behavior

I expect the crafting process to not check every second if it is in range of tools and materials. The materials can just be consumed into the crafting recipe, while the tools only need to be checked once upon starting the craft from 0%, or any%. If the player needs to  stop the craft and restart later, they set the in-progress craft down, then pick it up and re-check if the tools are present. Adding more loot around the player inside their view range should not increase the real-time to craft objects.

The player is locked on the spot while crafting. So to prevent the tools from wandering away, they can be put into an invisible inventory, or list, or something similar so that npc's cant take them, or they cant get destroyed by a meteor, or zombies don't break down walls and they get caved in on.

Ofcourse I haven't looked into the code as to how its handled. My observations may not represent the game's actual process for handling crafting. This is just how it looks and feels to a gamer.

### Versions and configuration

- OS: Windows
    - OS Version: MINGW/CYGWIN/MSYS2 on Windows 7
- Game Version: 18389fc [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    No Fungal Growth [no_fungal_growth]
]

ZhilkinSerg:
[...]

PR #66005, opened 2023-06-04
Refresh crafting speed less frequently

SurFlurer:
#### Summary
Performance "Refresh crafting speed less frequently"

#### Purpose of change
Mitigate #63586.

#### Describe the solution
Refresh crafting speed every 1, 10 or 100 turns, according to recipe time. For recipes that need 20min or less to craft, the behavior remains unchanged. For recipes that need 20min to around 3 hrs, refresh crafting speed every 10 turns. For recipes that need more 3 hrs, refresh crafting speed every 100 turns. Because every factor of crafting speed (lighting, workbench, morale and mutation) is a gradually transitioning or mostly consistent thing. So I think this approximation is not that unacceptable. 

Using figures from [here](https://github.com/CleverRaven/Cataclysm-DDA/issues/63586#issuecomment-1470034659), ~130 seconds is taken when checking everything every turn, and ~12s is taken when checking a portion of things every turn. So I think at least 130-12=118 seconds are consumed to actually check everything. 

For a 7hr30min in-game recipe, everything are checked 27000 times (turns.) Now they are checked 90% or 99% lesser frequently. This is especially good for long recipes.

Refresh crafting speed every 1% or 0.1% of progress. This is not ideal, because for extremely long recipes ( that take seasons or years to craft, 0.1% might still be a too big step.

#### Testing

Using the save from [here](https://github.com/CleverRaven/Cataclysm-DDA/issues/63586#issuecomment-1470034659).

Crafting a steel plate ( 9 hr recipe ) when both doors are closed consumed 7.7s.

Crafting a steel plate when both doors are open consumed ~16s. Far better than 2min-2min40s.

Crafting a lutefisk ( 18 min recipe ) when both doors are open consumed 9s.

Crafting a meat_tainted ( 1hr 30min recipe ) when both doors are open consumed 6.7s.

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -3200,6 +3200,7 @@ craft_activity_actor::craft_activity_actor( item_location &it, const bool is_lon
     cached_assistants = 0;
     cached_base_total_moves = 1;
     cached_cur_total_moves = 1;
+    refresh_speed = true;
 }
 
 bool craft_activity_actor::check_if_craft_okay( item_location &craft_item, Character &crafter )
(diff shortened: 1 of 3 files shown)


Zireael07:
Why close?

SurFlurer:
> Why close?

I found out this isn't easy to handle. There will be instant interrupts (as the test cases indicate.) And it's hard to separate workbench_crafting_speed_multiplier() with other multipliers.

I think I'll come back later and look into it further.

PR #66013, opened 2023-06-05
Aventails, Turban Helmets and Updated Nasal Helmets

MorvarchPrincess:
#### Summary
Content "Added graded steel turban/nasal helmets, chainmail aventails that attach to turban and nasal helmets."

#### Purpose of change

We had norse style nasal helms, but those were often worn with a chain aventail. 

Also added a turkish style turban helm too to add some lighter colder armor.

While I was at it, I upgraded the nasal helmets to use graded steel, and added a generic tempered steel material group.

#### Describe the solution

Adds four sets of items:

Turban helmets are slightly lighter and less warm than nasal helmets, but are mostly similar. They've been added to the japanese arms and armor, as out current source of "eastern" armors.

I added two types of aventails. "Partial" and full. There didn't seem to be a difference in naming IRL between the shorter one in the picture with the turban helmet and the longer one with the norse nasal helmet.
The Partial covers nape, throat and mouth. The full also covers the shoulders. 
I did the attachment via an ablative pocket added to the nasal and turban helmets. 
I added them to both the eastern and western armor books.

I also added nasal helmets to the graded steel list.

Everything is using the nested crafting menu framework.

Turban helmets could also go in the chinese armor book?

Would love extra scrutiny on the numbers of the new stuff.

#### Testing

I confimed the items existing in both crafting and spawning them in.

Confirmed that the helmets have the right looking stats in game, and that I can attach an aventail to the helmets properly.

#### Additional context

Viking helmet with aventail:
https://www.medievalcollectibles.com/product/viking-helmet-with-aventail-16-gauge/

Turban helmet with aventail:
https://www.metmuseum.org/art/collection/search/24813


Drew4484:
While not absolutely necessary, are you intending to add case hardened versions?

sadenar:
I think if you're willing to add more stuff than just this in the future this could be added to both the Chinese and a new Near Eastern Armor book.
Adding these recipes to the Chinese armor book makes more sense due to much greater historical cultural interaction with the Muslim world and Turkic peoples while adding it to the Japanese one makes no sense.

randomtyper:
First off, this is glorious. Nice ideas, OP. Now, for my inputs.

1. I agree with the opinion that turban helm should be in Chinese armor manual, not Japanese. Hell, even European armorsmithing book would be pretty logical choice (Kievan Rus was a major employer of Eastern style arms and armor), but it is already overflowing with recipes anyways. And for the future additions, Middle East has quite some unique weapon and armor designs (like mirror armor, yatagan short swords and all sorts of sabers) to fill out a whole new manual.
2. Remove leather padding from nose. Cut the coverage of leather on the cap piece to about 50% and thickness to 1 mm, as the leather mostly serves in suspending mechanism, not as protective padding. But add "PADDED" flag, as it has inner webbing and should cause no sores to head. Adjust weight to about 1.5 kg, as 1 mm thick helmet here https://www.battlemerchant.com/en/norman-helmet-st.-wencelas-1-mm-steel?c=307 weighs more than 1 kg while having thickness of only 1 mm.
3. Ideas for new pieces (mostly just want to lay out ideas I got from reading PR, no pressure on OP):
- Face mask like those of cumans https://www.artstation.com/artwork/nY6Z3o Mountable on the turban helm's nose piece, can be worn with aventails (it's custom-made by survivor anyways). "FANCY" tag (I dare you to say it's not).
- Open face aventails.
- Spangenhelm - same as nasal helms, but without the nosepiece. Possible slots for cheek pieces, nape cover and maybe faceplate (https://en.wikipedia.org/wiki/Sutton_Hoo_helmet is what I have in mind, though it is not a spangenhelm per se).
- Skullcap - close to skin spangenhelm without padding.

data/json/items/armor/helmets.json
```diff
@@ -2289,5 +2606,269 @@
...
     ],
     "melee_damage": { "bash": 1 }
+  },
+  {
+    "id": "lc_helmet_turban",
+    "type": "ARMOR",
+    "category": "armor",
+    "name": { "str": "mild steel turban helm" },
+    "description": "A middle eastern helmet, cylindrical on the top with protection for the nose and ears but nothing lower. This one is made in modern times out of mild steel.",
+    "weight": "800g",
```
randomtyper, line 2616: I would advise to adjust weight value for turban helm, too. Just as I said in my previous comment: 1 kg per 1 mm of thickness for total of about 2.5 kg +- 100 g. Otherwise, I approve of the changes and stats.

MorvarchPrincess: Fixed. Made it 2.4 as its just metal with no leather or anything.

bombasticSlacks:
@Drew4484

Drew4484 (approved):

Drew4484 (approved):

MorvarchPrincess (commented):
Cancel review accidently started

PR #66018, opened 2023-06-05
Cache crafting speed's workbench multiplier

SurFlurer:
#### Summary
Performance "Cache crafting speed's workbench multiplier"

#### Purpose of change
Mitigate https://github.com/CleverRaven/Cataclysm-DDA/issues/63586.

Using figures from https://github.com/CleverRaven/Cataclysm-DDA/issues/63586#issuecomment-1470034659, ~130 seconds is taken when checking every item surrounding every turn, and ~12s is taken when checking a portion of things every turn. So I think at least 130-12=118 seconds are consumed to actually check everything surrounding.

For a 7hr30min in-game recipe, every surrounding items are checked 27000 times (turns.) 

#### Describe the solution
Cache the workbench multiplier in `craft_activity_actor`.

Workbench conditions mostly remain consistent during craft activities. So refresh it per 5% process.

I also added the cache for disassemble actor.

Refresh it more frequently, like per 1% or 0.1%.

#### Testing
Using the save from https://github.com/CleverRaven/Cataclysm-DDA/issues/63586#issuecomment-1470034659.

Craft a steel plate ( 9 hr recipe ) when both doors are closed consumed 10.5s.

Craft a steel plate when both doors are open consumed 13s. Much better than 2min or 2min40s.

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -3200,7 +3200,7 @@ craft_activity_actor::craft_activity_actor( item_location &it, const bool is_lon
     cached_assistants = 0;
     cached_base_total_moves = 1;
     cached_cur_total_moves = 1;
-    cached_workbench_multipler = -1;
+    cached_workbench_multiplier = -1;
 }
 
 bool craft_activity_actor::check_if_craft_okay( item_location &craft_item, Character &crafter )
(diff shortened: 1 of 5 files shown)


bombasticSlacks (approved):
LGTM, it might be worth it to have bool for recaching since having to manually set re-caching to -1 in tests is a bit iffy.

bombasticSlacks:
The goal is to make it obvious what is happening to anyone else who looks at the code, using "-1" to signify recache and a null ptr to signify don't cache isn't very obvious especially without proper commenting.

I'm not sure if this is a place elegance is critical, and I would personally have just added a verbose bool.

SurFlurer:
So I decide to pre-calculate that multiplier. Also add a bool as an indicator for re-caching.

SurFlurer:
Thanks for merging!

PR #66020, opened 2023-06-05
UPS no longer automatically gets drained

### LINKED ISSUE #64890: Prevent UPS from auto-charging looted smartphones.

I use a UPS to power all of my electronics, and I also have smartphones on autopickup for the high-capacity batteries and SD-Memory cards (which may contain rare crafting recipes) you get from disassembling them. But when I pick a few up into my backpack from dead zeds, they over the course of just a few turns drain my UPS completely, leaving no power for the stuff I actually need it for. Instead, if I want the bits inside, I have to manually disassemble every smartphone when I find it which can be dangerous and easily interruptible, and takes ages compared to bulk disassembling them when I get home. If there's already a way to prevent this, please help.

### Solution you would like.

I had a few ideas as to how this could be solved:

1: Add a toggle to the pocket menu that determines if items inside are charged by a UPS. (Like the auto-insert toggle)

2: Make it so the UPS only charges favorited items.

3: Make it so you have to manually connect items to your UPS before it will charge them.

4: Add a blacklist/whitelist to the UPS itself which determines what it's allowed to charge.

5: Make smartphones use regular batteries until modified by the player.

6: Make smartphones charge much, much slower, giving the player time to put them in a shopping cart or something before their UPS runs dry.

Dealing with it. Giving up on the convenience of UPS and switching back to regular batteries for my small electronics (doesn't help with power armor and the like). Stopping picking up smartphones forever. Slowly and manually diassembling every single smartphone in the middle of the street where I find them.

### Additional context

_No response_

Ruminator24:
#### Summary

Bugfixes "UPS no longer gets automatically drained when in inventory"

#### Purpose of change

This helps fixes #64890, and makes sure the UPS only spawns as off.

#### Describe the solution

I simply added a second UPS item for when it's turned off, much like other electronic items.

N/A

#### Testing

Works fine when charges, doesn't charge when turned off, and spawns normally.

#### Additional context

I would remove it's ability to remove one charge when turned on, but I really don't wanna mess with what happens if the charger is at zero when I turn it on.


Hirmuolio:
I think you can use `"need_charges": 1` to require the item to have charges to be turned on but not consume charges for the action of turning on.

Ruminator24:
Yep, I'm in the middle of doing that, i just need to find where the npcs use UPS_off, as well as for tests

data/json/items/tool/electronics.json
```diff
@@ -738,6 +738,71 @@
...
+      {
+        "pocket_type": "MAGAZINE_WELL",
+        "flag_restriction": [ "BATTERY_HEAVY" ],
+        "default_magazine": "heavy_plus_battery_cell"
+      }
+    ],
+    "charges_per_use": 1,
+    "use_action": {
+      "ammo_scale": 0,
+      "target": "UPS_on",
```
Drew4484, line 797: Attempting to transform to UPS_on

PR #66024, opened 2023-06-05
Validate firstaid_activity_actor's item_location and add per-turn check

### LINKED ISSUE #66017: Taking away bandages from friendly NPC followers bug
### Describe the bug

Taking away bandages from friendly NPC followers while they are bandaging themselves will cause the game to crash.

### Attach save file

N/A

### Steps to reproduce

1. Give the friendly NPC some bandages and antiseptic
2. Make the friendly take damage from zombies
3. The NPC will proceed to bandage himself ( might require the npc to have a bit of healthcare skill, the friendly NPC that triggered this bug had 8 in healthcare )
4. Kill zombie (optional)
5. Take away those bandages and antiseptics BEFORE the NPC finishes using "first aid "
6. Walk around a bit
7. Game crash

### Expected behavior

I expected the the NPC to stop applying first aid and go on as normal 

### Screenshots

Seconds before the crash.
image
What pops up after the crash 
image

Windows 10
Using the latest Stable 0.G Gaiman
Using the Built in MSXOTTO tileset

### Additional context

[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/11650113/crash.log)
[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/11650115/debug.log)

SurFlurer:
#### Summary
none

#### Purpose of change
Fix #66017 .

#### Describe the solution
Validate firstaid_activity_actor's item_location and add per-turn check

Consume the bandage ( or other first-aid tools ) at the start of the activity. Because you won't have an intact and uncontaminated bandage if the first-aid gets interrupted. It's possible that it has been soaked with blood, or fallen onto ground if you decide to do some other things. 

Please comment on whether this alternative is better.

#### Testing
Fixed the crash.

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -5991,7 +5991,7 @@ std::unique_ptr<activity_actor> clear_rubble_activity_actor::deserialize( JsonVa
 void firstaid_activity_actor::do_turn( player_activity &act, Character & )
 {
     item_location it = act.targets.front();
-    if( !it.get_item() ) {
+    if( !it ) {
         // Erase activity and values.
         act.set_to_null();
         act.values.clear();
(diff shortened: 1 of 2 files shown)


PatrikLundell:
I agree with the logic that starting the activity should destroy or degrade the consumable, but only if breaking off the activity doesn't result in accomplishing the task faster than letting it run its course (i.e. interrupting the activity should at best allow it to be resumed or produce a partial result, but typically no progress at all and never full progress).
Immediate consumption with no progress on cancellation is good enough. Everything else are improvements on that.

SurFlurer:
Closing this for now. When I have the motivation to finish this I will reopen.

PR #66046, opened 2023-06-06
expand common items found in vehicles

casswedson:
#### Summary
Content "increase items that spawn inside vehicles"

#### Purpose of change
I am bored with only finding jacks in the back of cars and _rarely_ large amounts of chocolate covered snacks

make a vehicle specific "toolbox" with different spawn logic and items that I a non-vehicle owner would find useful, like screwdrivers, pliers _and_ wrenches that spawn together cause they make a good team, duct tape, lighting tools, cutting tools

expand the existing cat kit itemgroup to spawn be more interesting ie spawn jerrycans along with funnels, but don't exclude a hose, include the toolbox specified above and generally increase spawn changes

expand the misc cat itemgroup: increase spawn chances, split groups of similar items into distributions

#### Testing
some samples
Screenshot from 2023-06-06 15-37-00
Screenshot from 2023-06-06 15-37-17
Screenshot from 2023-06-06 15-38-51
Screenshot from 2023-06-06 15-39-12


sadenar:
Random trash would be good too, like a random novel the owner left there, a flyer they dumped on the floor.
That or just them being a slob and leaving empty/half-eaten stuff in their car.

PR #65924, opened 2023-05-30
Grain farming expansion and seed adjustments

worm-girl:
#### Summary
Features "Expand grain farming and adjust seeds"

#### Purpose of change

https://github.com/CleverRaven/Cataclysm-DDA/issues/64807 and the ensuing discussion outline several issues with farming, specifically that it's a fire-and-forget process with almost no character-side labor time involved. In reality, pre-industrial farming was backbreaking work that demanded enormous man-hours to keep people fed.

Previously, it was possible to plant a few wheat or barley seeds in a matter of minutes and return later in the year to a field which yielded tens of thousands of calories of unspoilable food which required no processing or even cooking. Picking a plot of crops was nearly instantaneous and had no calorie cost. This skipped many important steps in the real-life process and trivialized food acquisition, which especially late in the mid to late game should be an important survival concern.

https://github.com/CleverRaven/Cataclysm-DDA/discussions/49327 identifies farming, base management, and food security as avenues for building out the mid game. While this PR doesn't address any of the concerns specific to growing and tending plants (that's beyond me, codewise) it does try to build a much more robust and accurate simulation of small-scale grain production, which I hope will spur further development in this area.

Lastly, the biggest justification in my mind for complexifying this process is that many players, myself included, enjoy the homesteading aspect of the game. I personally don't see farming as a button you push to receive food, it's fun part of the game in itself, food just happens to be the reward.

#### Describe the solution

**GRAINS HAVE BEEN OVERHAULED**

Someone overhauled corn a while back, but the other grains were left undone. I have largely left corn alone and approached the rest of the gains with an eye toward creating parity with corn.

* Wheat, barley, wild rice, oats, buckwheat, and canola must be harvested with a tool that has the grass cutting quality. At the time of this writing, that includes: bill, bronze bill, combat machete, machete, makeshift machete, stone sickle, tempered combat machete, bronze sickle, scythe, scythe, sickle. 

There's presently no difference between these tools, if it cuts grass it'll do the job. Cutting wheat with hand tools seems tricky, but it's a pretty fast process once people get going. I don't know that there's much point in getting more granular than that.

Cut grain can be eaten by cattle mutants immediately, it also works as cattle fodder.

* Wheat, barley, wild rice, oats, and buckwheat must be dried before use.

[...]

PR #66054, opened 2023-06-07
Fix map::grow_plant debug message when seed item is not found

Qrox:
#### Summary
None

#### Purpose of change
Triggered in #65924.

The overmap test creates submaps outside the main map with `tinymap`, which has only one z-level. However, `map::load` actualizes the `tinymap` starting from the lowest z-level -10, which in effect actualizes the only z-level for 21 times (!!!). Therefore, it encounters invalid plant tiles in the first actualization and prints an error for z-level -10, whereas in subsequent actualizations the offending plant tiles are already removed, therefore no error is generated for z-levels other than -10.

#### Describe the solution
Only actualize once on the map z-level if there is only one z-level.

Also add more info to the message to help debugging in the future.

#### Testing
The test now prints the correct coordinates for #65924. Loaded the game and teleported to unexplored overmaps and the buildings are still correctly generated.

#### Additional context
I also notice the map code is not respecting the `zlevels` flag in many other places and is probably screwed in one way or another, but for the sake of my sanity I'll just pretend I didn't see them for now.

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -7353,7 +7353,9 @@ void map::load( const tripoint_abs_sm &w, const bool update_vehicle,
     // with entities at the edges
     for( int gridx = 0; gridx < my_MAPSIZE; gridx++ ) {
         for( int gridy = 0; gridy < my_MAPSIZE; gridy++ ) {
-            for( int gridz = -OVERMAP_DEPTH; gridz <= OVERMAP_HEIGHT; gridz++ ) {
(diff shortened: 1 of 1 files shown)


jbytheway:
LGTM.

Does this mean the overmap coverage test will be substantially faster now?

Qrox:
It should be faster in theory but my tests and the CI tests are both showing inconsistent results, so I guess the time it was using was not significant anyway.

PR #66056, opened 2023-06-07
Give range bonuses to 9mm smgs

anoobindisguise:
#### Summary
Balance "9mm smgs have more range"

#### Purpose of change
9mm SMGs have a significantly higher effective range than many equivalent pistols - for example a Glock 17 or Beretta M9 has an effective range of 50 meters, but Uzis and MP5A2s are listed at 200m. Some of this may be simply due to a greater volume of bullets = higher chance of getting at least 1 hit but there's no question that their longer barrels are contributing somewhat just as they improve muzzle energy and thus damage.

#### Describe the solution
Provide slightly arbitrary adjustments to maximum range (default of 9mm is 14 tiles). The Cx4 Storm has +20 range already, so I gave extra ranges going up to +14 tiles from the SMGs with the longest barrels (some have less, such as only +8 or +12).

Completely redo the system of range compression and the reality bubble to give each gun an effective range exactly equal to its real life version.


data/json/items/gun/9mm.json
```diff
@@ -251,6 +252,7 @@
     "longest_side": "377 mm",
     "skill": "pistol",
     "modes": [ [ "DEFAULT", "semi-auto", 1 ] ],
+    "range": 14,
```
TheShadowFerret, line 255:
```suggestion
```
Copies from MP5, Clang is mad you defined duplicate value 😛

anoobindisguise: Thanks!

PR #66057, opened 2023-06-07
Gas Mask Filters Update

Drew4484:
Changes gas mask filters to be a single filter with charges instead of a reloadable pseudo magazine.

#### Summary
Content "Changes mask and rebreather filters to a single cartridge with charges."

#### Purpose of change
Currently, gas masks are loaded with charges of gas filter, instead of a distinct cartridge that is used and disposed of. The same issue applies to the rebreather.

#### Describe the solution
Adds a new item that represents a disposable gas mask filter cartridge and one for a rebreather cartridge. 

None.

#### Testing
Spawned gas mask, unloaded it. Spawned new cartridge, loaded it. Activated mask. Spawned thick smoke, walked around in it for a few turns. Checked that charges were consumed. Repeated test with rebreather and water.


macrosblackd:
The Filter Mask also has the same issue - not sure if you've included that in your changes.

Drew4484:
> The Filter Mask also has the same issue - not sure if you've included that in your changes.

Yep got the filter mask and all the variants as well, you can see in the edited files under tool armor.

randomtyper:
Rebreather filters are refillable IRL, and have a bit different recipe. Will make an issue in a while.

Drew4484:
> Rebreather cartridges (not filters, they either scrub out exhaled air or regenereate O2 from CO2) are refillable IRL, and have a bit different recipe. Will make an issue in a while.

You can just post the details here, it won't take much to update the cartridges to be refillable.

randomtyper:
> > Rebreather cartridges (not filters, they either scrub out exhaled air or regenereate O2 from CO2) are refillable IRL, and have a bit different recipe. Will make an issue in a while.
> 
> You can just post the details here, it won't take much to update the cartridges to be refillable.

As a stopgap, yes, making rebreather cartridges refillable will be quick and easy. But as a realistic implementation... Not really.

Drew4484:
> Conflicts on this PR

That's what I get for not refreshing after the tests run.

randomtyper:
Survivor divemask was not updated in this PR and uses the old cartridges.

PR #66063, opened 2023-06-08
Oversize Single-strap Bags

Swootch:
Adds the OVERSIZE flag to the duffel bag.

#### Summary
Balance "Allows the duffel bag to be worn by characters with the Large and Huge mutations, along with their variants."

#### Purpose of change
The duffel bag is worn via a single large adjustable sling, which isn't any more constraining than similar backpacks and bags that are wearable by large and huge mutated characters.

#### Describe the solution
Allowing the duffel bag to be worn by Large and Huge characters.

Not allowing duffel bags to be worn by Large and Huge characters.

#### Testing
Made a character, mutated the full crustacean line which includes Huge, spawned a duffel bag and wore it.

#### Additional context
Wearing a duffel bag while mutating into a size category that forbids wearing the duffel bag without the oversize flag has it shredded into nothingness, which feels wrong, since at best the adjustable sling would snap.

### DIFF
--- a/data/json/items/armor/storage.json
+++ b/data/json/items/armor/storage.json
@@ -1597,7 +1597,7 @@
       }
     ],
     "material_thickness": 1,
-    "flags": [ "BELTED", "WATER_FRIENDLY" ],
+    "flags": [ "BELTED", "WATER_FRIENDLY", "OVERSIZE" ],
     "armor": [
       {
         "encumbrance": 4,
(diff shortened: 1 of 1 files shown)


fairyarmadillo:
Pretty much every single-strap bag should be set to oversize IMO. That's purses, messenger bags, duffel bags, swag bags, what else?

Swootch:
Sling Packs and uhh the Debug Pocket Universe? Which gets also destroyed when mutating into a huge creature if worn. That's it as far as I can see, most of what makes sense already is wearable.

PR #66067, opened 2023-06-08
Make all claws outer layer

KittyTac:
#### Summary
Bugfixes "Make all claws outer layer"

#### Purpose of change
During playtesting I noticed that normal claws, which were personal-layer, were under fur and thus didn't work as melee weapons.

#### Describe the solution
Add OUTER flag to all claws.

Making a flag that makes them usable with fur.

#### Testing
Tested and the claws work correctly.

#### Additional context
N/A

### DIFF
--- a/data/json/items/armor/integrated.json
+++ b/data/json/items/armor/integrated.json
@@ -1395,7 +1395,7 @@
     "material": [ "chitin" ],
     "symbol": ";",
     "color": "dark_gray",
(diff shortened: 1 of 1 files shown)


anoobindisguise:
Normal claws could be skintight instead so that they still fit under gloves. If they're meant to not be as long and bulky.

KittyTac:
> Normal claws could be skintight instead so that they still fit under gloves. If they're meant to not be as long and bulky.

Yeah but making them skintight would make them unusable with most integrated armor.

Anyways I tested it and you can still wear gloves with the claws so that's not a concern.

PR #66068, opened 2023-06-08
Fix skin and armor mutations

### LINKED ISSUE #66049: Can lose Epicuticle while retaining Chitin Charapace and Feathers underneath.
### Describe the bug

The Feathers mutation from the bird tree can override/replace your Epicuticle without doing anything to your Chitin Charapace, leaving you with a full-body-uncomfortable armor you can't unequip (Chitin Charapace is integrated armor). You also apparently then have feather underneath that Charapace. **Furthermore** your Chitin Charapace will eventually molt, which causes your Epicuticle to reappear, without making you lose your Feathers mutation. Trying to fix this with in-game means proved to be very difficult because trying to remove Feathers leaves you with Rough Skin, which is a pre-requisite for Epicuticle, but you already have Epicuticle (ordinarily you lose Rough Skin when it turns into Epicuticle).

### Attach save file

N/A

### Steps to reproduce

1. Get Insect mutations Epicuticle and Chitin Charapace.
2. Take Bird mutagen and happen to get the Feathers mutation.
3. Feathers replaces your Epicuticle, leaving the Charapace intact for further complications.

### Expected behavior

Chitin Charapace is lost before Epicuticle, since one is presumably growing out of the other.
Alternatively, the Charapace is simply destroyed upon molting the next time, since it presumably can't regrow without an epicuticle.

### Screenshots

_No response_

- OS: Windows
- Game Version: 4602415 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Bombastic Perks [bombastic_perks]
]

### Additional context

_No response_

worm-girl:
#### Summary
Bugfixes "Fix improperly overlapping skin and armor mutations"

#### Purpose of change

fixes #66049 

Several skin mutations shared a common precursor, rough skin, which only had the SKIN type. This meant that you could acquire epicuticle and chitin carapace, then mutate to feathers, losing epicuticle while keeping chitin carapace, despite the former being a prerequisite for the latter. This was also true of the shell mutations, which are supposed to require thick skin. This could cause all kinds of conflicts between different types of integrated armor competing for the same layer, integrated armor without padding, etc.

#### Describe the solution

Rough skin had no effect and was previously the sole precursor to a bunch of skin mutations, including patchy scales, patchwork skin, lightly furred, epicuticle, and phelloderm. I decided to split it out into six precursor mutations with mild effects designed to promote the body horror aspect of mutation. Most of the effects are somewhat negative, which will hopefully prompt would-be mutants to take more mutagen to evolve past them.

Peeling skin - 2 ugliness, no other effect. Leads to patchy scales.
Jaundice - ugliness, no other effect. Leads to phelloder..
Smooth skin - 1 ugliness, very mild wet effect reduction, -1 cut armor. Leads to epicuticle.
Rashy skin - 1 ugliness, randomly procs the formication effect for ten minutes a couple of times a day. Leads to patchwork skin.
Hirsute - No effect whatsoever. Leads to lightly furred.
Goosebumps - 1 ugliness. Leads to feathered.

Each of these mutations has the SKIN, BODY_ARMOR, TORSO_ARMOR, and BODY_ARMOR_MOD, and ARMOR types. These mutations are not kept when advancing to their next stage. 

* Specifically canceling undesirable mutations in each mutation's json entry - This is painstaking and would create a lot of work for future contributors.

* Just making rough skin have all those types so it cancels everything. - Where's the fun in that? The method I've used also sets up possible future changes such as allowing two different types of skin to both have access to the same armor mutation without losing it when you mutate between them.

#### Testing

* Spawned into a world, gave myself epicuticle and chitin carapace.
* Mutated feathers, saw that I lost both epicuticle and chitin carapace in the process
* Gave myself rashy skin. Waited a few hours until I started itching. Saw that itching lasted a reasonable length of time and was not excessive in frequency
* Gave myself all the second stage skin mutations, saw that they all were appropriately gained and lost
* Completed a molt cycle with chitin carapace and hairy chitin to make sure that hairy chitin wasn't somehow going to get lost in the shuffle

#### Additional context

muties rule

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -1963,7 +1963,7 @@
     "visibility": 1,
     "ugliness": 1,
     "description": "It's subtle, but your skin has definitely taken on a sickly yellow-green hue.",
-    "types": [ "SKIN", "BODY_ARMOR" ],
+    "types": [ "SKIN", "BODY_ARMOR", "TORSO_ARMOR", "BODY_ARMOR_MOD", "ARMOR" ],
     "changes_to": [ "PLANTSKIN" ],
     "category": [
       "PLANT",
@@ -1978,7 +1978,7 @@
     "visibility": 2,
     "ugliness": 1,
     "description": "You have goosebumps that never seem to go away.  Are they getting bigger?",
-    "types": [ "SKIN", "BODY_ARMOR" ],
+    "types": [ "SKIN", "BODY_ARMOR", "TORSO_ARMOR", "BODY_ARMOR_MOD", "ARMOR" ],
     "changes_to": [ "FEATHERS" ],
     "category": [ "BIRD" ]
   },
@@ -1988,9 +1988,9 @@
     "name": { "str": "Rashy Skin" },
     "points": -1,
     "visibility": 2,
-    "ugliness": 2,
+    "ugliness": 1,
     "description": "You've developed a persistent skin rash.  Sometimes it's fine, other times it itches like hell.",
-    "types": [ "SKIN", "BODY_ARMOR" ],
+    "types": [ "SKIN", "BODY_ARMOR", "TORSO_ARMOR", "BODY_ARMOR_MOD", "ARMOR" ],
     "changes_to": ["PATCHSKIN1" ],
     "category": [ "CHIMERA" ]
   },
@@ -2000,7 +2000,7 @@
     "name": { "str": "Hirsute" },
     "points": 0,
     "description": "Your body hair is growing in thicker than it used to.  It's almost like you're going through a second puberty or something.",
-    "types": [ "SKIN", "BODY_ARMOR" ],
+    "types": [ "SKIN", "BODY_ARMOR", "TORSO_ARMOR", "BODY_ARMOR_MOD", "ARMOR" ],
     "changes_to": [ "LIGHTFUR" ],
     "category": [
       "MOUSE",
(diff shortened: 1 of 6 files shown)


Maleclypse (commented):
I think this is good but I'm falling asleep reviewing it.  I'll try and come back to this tomorrow.

data/json/mutations/mutations.json
```diff
@@ -1955,35 +1955,87 @@
...
+    "changes_to": [ "PATCHSKIN1" ],
+    "category": [ "CHIMERA" ]
+  },
+  {
+    "type": "mutation",
+    "id": "HIRSUTE",
+    "name": { "str": "Hirsute" },
+    "points": 0,
+    "description": "Your body hair is growing in thicker than it used to.  It's almost like you're going through a second puberty or something.",
+    "types": [ "SKIN", "BODY_ARMOR", "ARMOR", "BODY_ARMOR_MOD" ],
```
Maleclypse, line 2000: What's the benefit of grabbing all these mutation types?  Nothing currently uses this combination.

fairyarmadillo: That's the entire point of the PR. Without those types attached, you suffer the problem described in the associated issue, where you can, for instance, lose CHITIN1 but keep CHITIN2, which is unintended and leads to all kinds of problems.

Most of ths mutations using those flags are designed to assume that there's a specific mutation type beneath them. If you'rs still not following, please follow the steps in the attached issue to see what the original problem was and then try that again with these changes.

Maleclypse: I forgot that no one has moved the major body transformations behind the Thresholds.  That was discussed previously and would have made this less necessary. At least in some cases.

PR #56345, opened 2022-03-25
[CR] Cleanup basecamp primitive farm. Fix Plow fields

haveric:
#### Summary
Bugfixes "Cleanup basecamp primitive farm. Fix Plow fields"

#### Purpose of change

Cleanup and improve the primitive farm basecamp construction. (May pull some of this out as the scope of this PR has changed)
Fix the basecamp overmap names not being updated on upgrades. (farm survey -> farm)
Attempting to fix the "Plow Fields" functionality for both basecamp farms.
Should fully fix #55739

#### Describe the solution

- Added a new json attribute that controls the overmap updating:
- Move upgrade rotation (and mirroring ideally) to the overmap to reduce duplication of flags

```
  "blueprint_update_oter": "bar"
  "blueprint_update_oter": [ { "pre": "any", "id": "bar" } ] // Same as above
  "blueprint_update_oter": [ { "pre": "foo", "id": "bar" } ]
  "blueprint_update_oter": [ { "pre": [ "foo", "foo2" ], "id": "bar" } ]
  "blueprint_update_oter": [ { "pre": [ "foo" ], "id": "bar" }, { "pre": [ "foo2" ], "id": "bar2" } ]
  ```
- Updated most of the primitive farm requirements and times to match the current state of building (Mostly shorter times, but more materials). I also chose to use the existing dirt floor w/ roof instead of the non-existent log roof, which changed some requirements. (May revert some of this as the scope of the PR has changed drastically)

TODO:
- [x] Fix rotation for the overmap
- [ ] Implement mirror_horizontal and mirror_vertical for overmaps (before rotation, such as `overmap_name_mh_north` or `overmap_name_mv_north`
- [ ] Convert existing overmaps in order to preserve rotation for upgrades
- [ ] Make sure other expansions are updating the overmap name correctly

#### Describe alternatives you've considered

- Remove all missions for plowing fields and only let the "Plow Fields" action plow them
  - Optionally with a number input to plow less
- Figure out how these used to work. I'm pretty sure it used to go off of the result, but I can't find when "Plow Fields" stopped working. Result also isn't consistently being used to reference an overmap, which is why I chose to add a new field.
- Let someone better at maps fix this

#### Testing

In progress.

#### Additional context

~I'm struggling to get the overmap rotation to work correctly. Hoping someone can take a look at my changes and make sense of what I'm trying to do.~ Overmap rotation is working now, but mirroring them is another can of worms that I wasn't planning on having to implement. If anyone wants to help with this, I would love some assistance here.

PR #66070, opened 2023-06-08
Fix basecamp farm's "plow field" mission

SurFlurer:
#### Summary
none

#### Purpose of change
Found during my attempt to further extend #65870.

It's impossible to plow the fields again, once the dirt mounds are all converted to t_dirt by harvesting. The reason is that the plowing function is trying to infer what the farm should look like according to the OMT type of the farm, but it never gets updated via upgrading. So the function only gets a view of a desolate farm survey" map, and thinks no field can be plowed. 

Also fix a test mistakenly tampered by me in #66018. Sincerely apologize for that.

#### Describe the solution
Update the overmap terrain type when finishing camp upgrade.

#### Testing
Fully upgraded the farm, the omt type did get updated.

Plow, plant, harvest. Plow, plant, harvest. All cycling perfectly now.

### DIFF
--- a/src/faction_camp.cpp
+++ b/src/faction_camp.cpp
@@ -3381,7 +3381,7 @@ bool basecamp::upgrade_return( const mission_id &miss_id )
     }
     update_provides( bldg, e->second );
     update_resources( bldg );
-
+    overmap_buffer.ter_set( upos, oter_id( bldg ) );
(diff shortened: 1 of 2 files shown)


SurFlurer:
Oh I noticed #56345 which encapsulates this PR. However that one hasn't had any activity in a year.

Issue #64552, opened 2023-03-25
segfault when reloading an item


rubenpd:
### Describe the bug

I had this bug twice or thrice between yesterday and 2 days ago, but only now i managed to get a 100% reproduction rate save.

The game just segfaults when i'm reloading my tailor kit, unfortunately i don't know more this time

### Attach save file

[Emerald Lake Hills-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11068455/Emerald.Lake.Hills-trimmed.tar.gz)

### Steps to reproduce

1 load save
2 hit "reload" button
3 select the tailor kit and hit enter
4 game should segfault

### Expected behavior

reload action being executed succesfully and no crash 

### Versions and configuration

Windows 10
build 2023-03-25-0547 msvc
Tiles
English
mods:
[
  "dda",
  "no_npc_food",
  "no_fungal_growth",
  "Tamable_Wildlife",
  "stats_through_kills",
  "speedydex",
  "package_bionic_professions",
  "blazeindustries",
  "rub_reb",
  "No_Person",
  "extra_mut_scens",
  "MMA",
  "translate_dialogue",
  "No_Rail_Stations"
]

Wyghab:
I had the same thing when reloading an Mp5 near a bunch of ammo on the ground. When I later picked up the ammo, half of it disappeared into the void, as if it never having existed. Maybe this is related?

andrei8l:
/confirmed, but it only crashes with the msvc builds, ASan doesn't catch anything, and the crash log is not very illuminating so I can't help.

SurFlurer:
Trimmed file size and scenario (much much less items, only a tailoring kit and 7 kinds of sewing thread nearby) for easier debugging.
[Emerald Lake Hills-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11068818/Emerald.Lake.Hills-trimmed.tar.gz)

c0bnut:
I think I've found a workaround for this bug. Take exactly the number of charges you need and walk away from your main stockpile. If you reload with the correct amount in your inventory it doesn't seem to crash.

PR #66073, opened 2023-06-08
Optimize the way map::process_items() handles vehicles

SurFlurer:
#### Summary
none

#### Purpose of change
There's a non-negligible portion of CPU cycles consumed on this function when crafting a sword inside a vehicle with tons of items ( using the save file attached at https://github.com/CleverRaven/Cataclysm-DDA/issues/64552#issue-1640442831 ), as can be seen on the flame graph:
捕获

#### Describe the solution
Grab the submaps first, then process each submap, which is more intuitive than the current implementation. 

Frankly, although I haven't figured out why it becomes faster, the flame graph does show improvement:

捕获1

The biggest part( map::process_items() ) mysteriously eliminates.

#### Testing

The game does run faster.


Zireael07:
Why close?

PR #66033, opened 2023-06-06
Gastropod Foot limb

Maleclypse:
#### Summary
Content "Gastropod Foot Limb"

#### Purpose of change
First of the replacement limbs.

#### Describe the solution
Uses #65505 as a model of how to replace limbs with their mutated counterparts.  The Gastropod foot merges both legs and feet into a single limb that slowly propels the character across the earth.   

#### Describe alternatives you've considered
I'm open to suggestions on improving this implementation.

PR #66079, opened 2023-06-08
Legs dehardcode for walking

bombasticSlacks:
#### Summary
None

#### Purpose of change
Needed for #66033

#### Describe the solution
Dehardcoded the legs checks for walking.

Added new rules that you need all your primary legs to be functioning to not struggle with walking.

This is however in its own function, in case people wish to extend it later.

#### Testing
Should just work :tm:

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -1373,7 +1373,7 @@ int Character::swim_speed() const
 
 bool Character::is_on_ground() const
 {
-    return ( get_working_leg_count() < 2 && !weapon.has_flag( flag_CRUTCHES ) ) ||
+    return ( !enough_working_legs() && !weapon.has_flag( flag_CRUTCHES ) ) ||
            has_effect( effect_downed ) || is_prone();
 }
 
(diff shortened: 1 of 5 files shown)


src/character.cpp
```diff
@@ -1941,26 +1941,35 @@ int Character::get_working_arm_count() const
...
+        if( part->primary_limb_type() == body_part_type::type::leg ) {
+            limb_count++;
+            if( !is_limb_broken( part ) ) {
+                working_limb_count++;
+            }
+        }
     }
-    return limb_count;
+
+    return working_limb_count == limb_count;
```
Brambor, line 1957:
```suggestion
    return true;
```

src/character.cpp
```diff
@@ -1941,26 +1941,35 @@ int Character::get_working_arm_count() const
...
-    if( has_limb( body_part_leg_r ) && !is_limb_broken( body_part_leg_r ) ) {
-        limb_count++;
+    int working_limb_count = 0;
+    for( const bodypart_id &part : get_all_body_parts() ) {
+        if( part->primary_limb_type() == body_part_type::type::leg ) {
+            limb_count++;
+            if( !is_limb_broken( part ) ) {
+                working_limb_count++;
+            }
+        }
```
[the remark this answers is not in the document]
bombasticSlacks: In the future, some body types might be able to function on only a portion of their limbs. Counting the limbs allows for easy extension of this based on body type in the future.

src/character.h
```diff
@@ -1283,9 +1283,10 @@ class Character : public Creature, public visitable
         // Return all the limb special attacks the character has, if the parent limb isn't too encumbered
         std::set<matec_id> get_limb_techs() const;
         int get_working_arm_count() const;
+        /** Returns true if at least half your legs are working */
```
SurFlurer, line 1286: However, if I understood correctly, `return working_limb_count == limb_count;` in `enough_working_legs()` means all legs are working.

SurFlurer: Thanks!

bombasticSlacks: Thanks for the review 😄

Maleclypse:
I've tested this with https://github.com/CleverRaven/Cataclysm-DDA/pull/66033 and all the errors that were happening are fixed.  Thank you!

PR #66082, opened 2023-06-08
Update seed.json

Froffy025:
#### Summary
Features "Dried lentils and dried beans can be crafted into their respected seeds."

#### Purpose of change
Easier to make seeds for these staple crops, rather than having to find unprocessed, raw varieties. I've grown sprouts from dried beans & lentils from the supermarket before -- this change makes it possible in game!

#### Describe the solution
Adds dried beans and lentils as components to the recipes for lentil and bean seeds. 

I could have made a new recipe that required water as a component, as would be more "realistic," but I believe that planting and growing these dried seeds means that they would get that water from rain, irrigation, etc. 

#### Testing
Added recipe, played with it in my own game, crafted both. Seemed to work. 

#### Additional context
image
image

### DIFF
--- a/data/json/recipes/food/seed.json
+++ b/data/json/recipes/food/seed.json
@@ -325,7 +325,7 @@
     "result": "seed_beans",
     "type": "recipe",
     "copy-from": "seed_extraction_base",
-    "components": [ [ [ "raw_beans", 1 ], ["dry_beans",	1] ] ],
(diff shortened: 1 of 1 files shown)


fairyarmadillo:
This is already handled by #65924

Froffy025:
my bad !!

PR #66083, opened 2023-06-09
Fix vitamins test flaking out, fingers crossed

irwiss:
#### Summary
None

#### Purpose of change

Fix a side effect of https://github.com/CleverRaven/Cataclysm-DDA/pull/66061 that only happens some times but not others 🤔

https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/5215165200/jobs/9412368189?pr=66078#step:16:88

#### Describe the solution

Make update_body( x, x ) reset last_updated when x is the same turn value

Can't quite pinpoint it as last_update mechanics seem weird and most tests don't reset it. Running `[effect][vitamins]` individually often randomly fails on commits even before #66061, likely related to test order somehow

#### Testing

Tests should pass

### DIFF
--- a/src/character_body.cpp
+++ b/src/character_body.cpp
@@ -196,6 +196,7 @@ void Character::update_body( const time_point &from, const time_point &to )
 {
     // Early return if we already did update previously on the same turn (e.g. when loading savegame).
     if( to <= last_updated ) {
+        last_updated = to;
         return;
     }
     if( !is_npc() ) {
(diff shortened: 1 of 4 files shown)


RenechCDDA:
I see you're touching stomach_contents_test even more, did you see my changes to it in #65732? Reconstructing avatar(the current behavior, unrelated to your PRs) is both weirdly unnecessary and seems to be the cause of some weirdness with setting up other tests.

irwiss:
The update_body() / stomach / vitamins code is weird, it both supports arbitrary time period by it's input params (which some code uses to simulate stuff) and uses ticks_between to handle it, but at the same time it doesn't - as some of it relies on calendar::once_every which reads the static calendar::turn which must match an exactly aligned number

There's good odds your change for the tests either won't be needed if merged after this or just needs to set calendar::turn to some on-the-hour number so the calendar checks trigger correctly

The real solution would be to alter update_body to not use any calendar:: funcs but it's a much larger change and this blocks tests right now

PR #66085, opened 2023-06-09
Make turret cpu npc persistent

SurFlurer:
#### Summary
none

#### Purpose of change
Turret cpu ( a fake npc ) is regenerated everytime `vehicle::get_targeting_npc()` is called. Also it is returning value instead of a reference, which is expensive. This function takes more than 2000ms of CPU time when the player character sleeps for 10 hours in a vehicle with one auto-firing turret.

#### Describe the solution
Introduce a `turret_cpu` class which contains a unique_ptr to a NPC.

Add a `turret_cpu` member to `vehicle_part` class, used for storing the turret control AI ( a fake npc ).

Modify `vehicle::get_targeting_npc()` so that it returns a reference.

1. Make the fake npc static and return a reference.
2. Do not make it static, only change the function to return a reference. This sole change contributes about 70 percent of the total improvement.

#### Testing
This function's consumption of CPU cycles shortened to less than 500ms in the aforementioned scenario.

#### Additional context
Thanks for @Qrox helping me!


src/turret.cpp
```diff
@@ -544,28 +544,30 @@ void vehicle::turrets_set_mode()
...
 }
 
-npc vehicle::get_targeting_npc( const vehicle_part &pt ) const
+npc &vehicle::get_targeting_npc( const vehicle_part &pt ) const
 {
     // Make a fake NPC to represent the targeting system
-    npc cpu;
-    cpu.set_body();
-    cpu.set_fake( true );
+    static npc cpu;
```
Qrox, line 550: This might cause issues if the calling code tries to get targeting NPCs for two vehicles/parts at the same time. It might be better to cache the NPC in the vehicle part instead. Or, alternatively, return a `shared_ptr` of the NPC and return a newly constructed NPC if the previous pointer is still in use by outside code, if we want to save memory.

SurFlurer: There are only two usages of `vehicle::get_targeting_npc()`, both of them are safe ( the turret is fired immediately following
the request of cpu ). The only problem l can imagine is when someone wants concurrency, calling two cpus in parallel.

Qrox: The potential of misusing the function is what I'm worried about. Updating the function documentation or name to reflect that it is returning a reference to a single npc might help too, although I'd prefer changing the name because who reads the documentation.

SurFlurer: `class npc` has an explicitly deleted `operator=()` function, making it a hassle when wanting to store an instance of NPC inside a vehicle ( maybe also vp ).

Qrox: You can create a `targeting_npc` class that stores a `unique_ptr` to an NPC and resets the cached pointer on copy-assignment and copy-construction, so it can create a different NPC after being copied into another vehicle part.

SurFlurer: Thanks, this is really inspiring to me. I'll see if giving every turret a dedicated NPC still has better performance than creating a new one every time.

SurFlurer: Wow, it's actually better than using a static one.

src/vehicle.h
```diff
@@ -233,6 +233,17 @@ enum class vp_flag : uint32_t {
...
 };
 
+class turret_cpu
+{
+    public:
+        std::unique_ptr<npc> brain;
+        turret_cpu() = default;
+        turret_cpu( const turret_cpu & ) {};
+        turret_cpu &operator=( const turret_cpu & ) {
+            return *this;
```
Qrox, line 243:
```suggestion
            brain.reset();
            return *this;
```
If the copy assignment function is called then presumably its vehicle part is being assigned with a different value, so the NPC should be invalidated.

I also suggest changing `turret_cpu` to be a private inner class of `vehicle_part`, making `turret_cpu cpu` private, and changing `get_targeting_npc` to be a member function of `vehicle_part` that takes a `vehicle` parameter instead, to keep `turret_cpu` as an implementation detail of `vehicle_part`.

src/vehicle.h
```diff
@@ -32,6 +32,7 @@
 #include "item_location.h"
 #include "item_stack.h"
 #include "line.h"
+#include "npc.h"
```
Qrox, line 35: I think you can avoid including `npc.h` and keep the forward declaration of `class npc` by declaring `~turrent_cpu()` in `turret_cpu` and defining it with `turret_cpu::~turret_cpu() = default;` in vehicle.cpp.

src/vehicle.h
```diff
@@ -327,6 +338,7 @@ struct vehicle_part {
         /* @return true if part in current state be reloaded optionally with specific itype_id */
         bool can_reload( const item &obj = item() ) const;
 
+        turret_cpu cpu;
```
Qrox, line 341:
```suggestion
        // No need to serialize this, it can simply be reinitialized after starting a new game.
        turret_cpu cpu; //NOLINT(cata-serialize)
```

src/vehicle.h
SurFlurer: Thanks for your suggestions. I think `turret_cpu` has already been private in `vehicle_part` ( the default access modifier ).

src/vehicle.h
SurFlurer: It refuses to compile :P

Qrox (approved):
Thanks!

src/vehicle.h
```diff
@@ -463,6 +480,13 @@ struct vehicle_part {
         /** parts are available if they aren't unavailable */
         bool is_available( bool carried = true ) const;
 
+        /*
+         * @param pt the vehicle part containing the turret we're trying to target.
```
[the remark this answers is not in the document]
SurFlurer: Oh my god I forgot to check :)

src/vehicle.h
SurFlurer: Thanks for approving, but I cannot figure out how to make it compile again without including `npc.h`. Should I just revert the change and include `npc.h`?

SurFlurer: OK I figured it out.

PR #66086, opened 2023-06-09
Clearer Crashed Airliner OMTs

Procyonae:
#### Summary
Bugfixes "Makes Crashed Airliner OMTs clearer"

#### Purpose of change

The Crashed Airliner only displays a single "A" on the map despite taking up a large amount of room.

#### Describe the solution

The Crashed Airliner now has 3 red "A"s for the body, 2 light gray "A"s for wreckage and the large dirt trail behind the crash is marked by ","s. I also adjusted the see cost of the tiles with little to no wreck.

The trail left by the aircraft isn't that obvious still, replacing it with underscores or "o"'s would be clearer still but at least you can see how far around you need to drive to avoid it now.

#### Testing

Genned a world with the changes and found the maps this affects.

#### Additional context
image
More accurately represented airliner crash


Maleclypse:
@LovamKicsiGazsii which part of this PR are you opposed to?

LovamKicsiGazsii:
The inconsistency between the homeless camp and the homeless shelter, the fact that there is already a wilderness location with the same symbol, and that I intentionally made the homeless camp hard to see among the forest because the camp is fairly camouflaged from up above.

I'm all good with the other stuff, I think the changes with the crashliner were neat.

Doesn't affect anything I'm doing long-term, I'm just grumpy around my creations.

Procyonae:
> The inconsistency between the homeless camp and the homeless shelter, the fact that there is already a wilderness location with the same symbol, and that I intentionally made the homeless camp hard to see among the forest because the camp is fairly camouflaged from up above.

I forgot the camp was separate tbh and didn't realise it was intentional when afaik nothing else tries to hide stuff, I'll reverse the change ^^

PR #65622, opened 2023-05-13
Add support for running and smashing animations

Rewryte:
#### Summary
Features "Add support for running and smashing animations"

#### Purpose of change
Certain actions could benefit from some extra animations for improved reactivity.  Sprinting and smashing are some of those.

#### Describe the solution
Adds new sprite IDs which allow tilesets to assign sprites for a motion trail when a character is running, as well as for various smashing animations.  Sprinting animations are available in 8 directions, while smashing animations come in ineffective smashing, effective smashing, and complete smashing (destruction).

Tileset makers have full control on how these are used.  For example, you could reuse 3 sprites between the 8 sprinting directions, show cracks for effective smashing, and no sprite at all for ineffective smashing.

This modifies the function cata_tiles::draw_custom_explosion_frame, which normally falls back to 3 explosion sprites determined by nc_color if the target sprite is missing.  It now supports nc_black which will result in no fallback image.  This adds some flexibility for more custom animations in the future.

#### Testing
The video below shows these animations at work using ChibiUltica as the tileset.  "fd_smoke" is used for both sprinting and smashing destruction, while "crack_glass_center_tall" is used for effective smashing.

https://github.com/CleverRaven/Cataclysm-DDA/assets/129854247/662adaf5-2626-4627-8b01-dffd17c60c7e

#### Additional context
The new sprite IDs can be found in TILESET.md under Hardcoded IDs.

PR #66087, opened 2023-06-09
Asynchronous animation type including sprinting and smashing

### LINKED ISSUE #65666: `Hang on a bit…` everytime i run
### Describe the bug

Whenever im running, `Hang on a bit…` popped up for a brief moment. I don't think it did that in the past.

### Attach save file

N/A

### Steps to reproduce

1. Change the movement mode to run
2. Run around
3. `Hang on a bit…`

### Expected behavior

`Hang on a bit…` not popping up when im running.

### Screenshots

_No response_

- OS: Android
- Game Version: c2f139b [32-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm],
    Crazy Cataclysm [crazy_cataclysm]
]

### Additional context

_No response_

Rewryte:
#### Summary
Features "Asynchronous animations including sprinting and smashing"

#### Purpose of change
#65622 uses the custom explosion system to introduce some new animations.  While it achieves its purpose, it is hardly the most efficient way to draw animations.  Said animations are also synchronous - the game basically waits for the `ANIMATION_DELAY` to elapse before handing control back to the player.  #65666 also pointed out that those animations can cause excessive pop-up messages.

#### Describe the solution
Introduce asynchronous animations, a new type of animation that does not pause the game while running.  It follows existing animation settings and works both tiles and curses.  Sprinting and smashing animations are changed to use this new animation type instead.  Fixes #65666.

Usage:
`g->draw_async_anim( p, tile_id, ncstr, nc_color );`
`p` is the location in tripoint form
`tile_id` is the sprite name as a string. If invalid, nothing is displayed (tiles only)
`ncstr` is the character to display as a string (curses only) (optional)
`nccol` is the color of the displayed character as nc_color (curses only) (optional)

#### Testing
Sprinting and smashing animations tested to be working in both tiles and curses.  Animations are modified by animation settings as expected.

### DIFF
--- a/src/animation.cpp
+++ b/src/animation.cpp
@@ -881,7 +881,7 @@ void game::draw_async_anim( const tripoint &p, const std::string &tile_id, const
 
     if( !use_tiles ) {
         if( !ncstr.empty() ) {
-            hit_animation( get_avatar(), p, nccol, ncstr );
+            g->init_draw_async_anim_curses( p, ncstr, nccol );
         }
         return;
     }
@@ -894,7 +894,7 @@ void game::draw_async_anim( const tripoint &p, const std::string &, const std::s
                             const nc_color &nccol )
 {
     if( !ncstr.empty() ) {
-        hit_animation( get_avatar(), p, nccol, ncstr );
+        g->init_draw_async_anim_curses( p, ncstr, nccol );
     }
 }
 #endif
--- a/src/animation.cpp
+++ b/src/animation.cpp
@@ -863,8 +863,8 @@ void game::draw_zones( const tripoint &start, const tripoint &end, const tripoin
 #endif
 
 #if defined(TILES)
-void game::draw_async_anim( const tripoint &p, const std::string tile_id, const std::string ncstr,
-                            const nc_color nccol )
+void game::draw_async_anim( const tripoint &p, const std::string &tile_id, const std::string &ncstr,
+                            const nc_color &nccol )
 {
     if( test_mode ) {
         // avoid segfault from null tilecontext in tests
@@ -890,8 +890,8 @@ void game::draw_async_anim( const tripoint &p, const std::string tile_id, const 
     g->invalidate_main_ui_adaptor();
 }
 #else
-void game::draw_async_anim( const tripoint &p, const std::string, const std::string ncstr,
-                            const nc_color nccol )
+void game::draw_async_anim( const tripoint &p, const std::string &, const std::string &ncstr,
+                            const nc_color &nccol )
 {
     if( ncstr != "" ) {
         hit_animation( get_avatar(), p, nccol, ncstr );
--- a/src/animation.cpp
+++ b/src/animation.cpp
@@ -886,11 +886,8 @@ void game::draw_async_anim( const tripoint &p, const std::string tile_id, const 
         return;
     }
 
-    shared_ptr_fast<draw_callback_t> async_anim_cb = make_shared_fast<draw_callback_t>( [&]() {
-        tilecontext->init_draw_async_anim( p, tile_id );
-    } );
-    add_draw_callback( async_anim_cb );
-    ui_manager::redraw();
+    tilecontext->init_draw_async_anim( p, tile_id );
+    g->invalidate_main_ui_adaptor();
 }
 #else
 void game::draw_async_anim( const tripoint &p, const std::string, const std::string ncstr,
--- a/src/animation.cpp
+++ b/src/animation.cpp
@@ -880,7 +880,7 @@ void game::draw_async_anim( const tripoint &p, const std::string &tile_id, const
     }
 
     if( !use_tiles ) {
-        if( ncstr != "" ) {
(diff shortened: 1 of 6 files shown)


src/animation.cpp
```diff
@@ -862,6 +862,46 @@ void game::draw_zones( const tripoint &start, const tripoint &end, const tripoin
...
+            hit_animation( get_avatar(), p, nccol, ncstr );
+        }
+        return;
+    }
+
+    shared_ptr_fast<draw_callback_t> async_anim_cb = make_shared_fast<draw_callback_t>( [&]() {
+        tilecontext->init_draw_async_anim( p, tile_id );
+    } );
+    add_draw_callback( async_anim_cb );
+    ui_manager::redraw();
```
Qrox, line 893: Is this call needed here? From what I understand, the game will handle inputs and call `ui_manager::redraw` after the calls to `draw_async_anim`, so simply calling `init_draw_asyn_anim` and `g->invalidate_main_ui_adaptor()` here will work.

src/animation.cpp
```diff
@@ -862,6 +862,46 @@ void game::draw_zones( const tripoint &start, const tripoint &end, const tripoin
...
+}
+#else
+void game::draw_async_anim( const tripoint &p, const std::string, const std::string ncstr,
+                            const nc_color nccol )
+{
+    if( ncstr != "" ) {
+        hit_animation( get_avatar(), p, nccol, ncstr );
```
Qrox, line 900: Does this mean it will still wait for the animation to finish in curses?

Rewryte: Nope.  It will cancel any running animations when an input is received.

Qrox: Well, you're right it will, but it will also ignore the first input when the animation is running:

https://github.com/CleverRaven/Cataclysm-DDA/blob/9e4ff331dfed15b45abd55012951dcc847c36b44/src/animation.cpp#L547-L550

It should probably be made consistent with tiles (cancel the animation and handle the input action)

src/animation.cpp
Rewryte: Yup it wasn't.  Integrated your suggestions.

src/animation.cpp
Rewryte: Apparently, handling the input by passing an `action_id` from `look_up_action( handle_input() )` to `game::handle_action( action_id act )` leads to a bunch of issues like actions not taking any time.  I'm not exactly familiar with the input loop, but I guess it has something to do with handling actions outside of do_turn().  I don't suppose you know a better way to go about this?

Qrox: I think you can store and use the animation data in curses similar to how `init_draw_async_anim` does on tiles. `add_redraw_callback` may be used but you'll have to store the `shared_ptr`s somewhere because `class game` only stores weak pointers to the callbacks.

Rewryte: Done.  I skipped the callbacks bit and just drew from an "animation layer" var like how it is done with tiles.  Works good so far on no-tiles mode.  Will test on real curses once I have my virtual machine handy.

Rewryte: Working now.  All that is left is the serialization warning for `async_anim_layer_curses`.  I don't intend to store it in the save game.  Not sure if there is a better place to store it than `class game` though.

src/handle_action.cpp
```diff
@@ -2848,11 +2849,12 @@ bool game::handle_action()
         ctxt = get_player_input( action );
     }
 
-#if defined(TILES)
     // Remove asynchronous animations if any action taken before the input timeout
     // Otherwise repeated input can cause animations to accumulate as the timeout is never reached
+#if defined(TILES)
     tilecontext->void_async_anim();
 #endif
+    g->void_async_anim_curses();
```
Qrox, line 2857: And here.

src/handle_action.cpp
```diff
@@ -386,12 +386,13 @@ input_context game::get_player_input( std::string &action )
...
             }
 
-#if defined(TILES)
             // Remove asynchronous animations after animation delay if no input
             if( current_turn.async_anim_timeout() ) {
+#if defined(TILES)
                 tilecontext->void_async_anim();
-            }
 #endif
+                g->void_async_anim_curses();
```
Qrox, line 394: Does it redraw in curses if you call `g->invalidate_main_ui_adaptor()` here instead of calling `ui_manager::redraw()` in `void_async_anim_curses`?

src/animation.cpp
Qrox: You can add `//NOLINT(cata-serailize)` after the line to suppress the warning.

Rewryte: Thanks.  Should have noticed how this is a standard practice.

src/handle_action.cpp
Rewryte: Looks like they do.  Added your changes.

PR #66101, opened 2023-06-10
fix: dry firing practice recipes aren't tiring

casswedson:
#### Summary
None

#### Purpose of change
I think it makes sense to remove the BRISK_EXERCISE activity level from dry firing practice recipes; what were you doing with those guns


Maleclypse (commented):
I could see there being some exercise given that you are often going to want to train moving and firing.  But Brisk is definitely too much

PR #66105, opened 2023-06-10
Roof Palette Porting

Procyonae:
#### Summary
Bugfixes "Ports more rooves to roof_palette"

#### Purpose of change

Palettes are cool

#### Describe the solution

Used the palette cause it were cool

Not using the palette and being lame

Alot of these would benefit from using character based item placement over place_items and some need nest spawns but one thing at a time


Procyonae:
While doing this I found a palette that isn't used by any vanilla/repo mod/obseleted map, [cground](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/json/mapgen_palettes/cground.json), that seems to have be an old version of [godco_camp](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/json/mapgen_palettes/church_retreat.json)

Should I just remove it?

PR #66107, opened 2023-06-10
Badges don't cover anything again

### LINKED ISSUE #65339: Badges are Rigid and add extra Encumbrance
### Describe the bug

Most badges, notably the Detective Badge, Swat Badge, etc... when worn seem to now cover Upper Torso on the Outer layer, thus conflicting with any clothing\armor on said layer and increasing encumbrance by a relative amount. In addition they are also rigid, which means you can't wear more than one, or any other type for that matter.

I am unsure if it's a new\unexpected feature or a bug, but decided to report it since I haven't seen anything related to this change.

### Attach save file

N/A

### Steps to reproduce

1. Pick up some of the mentioned Badges
2. Wear any clothing or armor that covers the Outer layer on the Upper Torso or more
3. Wear a Badge alongside the clothing or armor
4. Notice how encumbrance rises due to conflicting layers and how you are unable to add more Badges

### Expected behavior

It didn't do it before, nor have I seen any specific detail about this change, other than the regular updating of Clothing and Armor layers to the new format.
While mostly irrelevant to gameplay, it is still kinda silly how such a small badge can add so much encumbrance, not to mention they prevent the usage of any extra Badge as if they took all the available space.
I'd expect them to cover what most flavour and fancy items do, the _Nothing_ layer, which they did before. They do have a coverage value of 3% and some armor value as well, which is odd to me, but disallowing the use of more Badges is surely excessive.

And well, I like wearing pointless things so uh... my character can feel important. Although the Doctor Badge is handy at times.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19045.2846 (22H2)
- Game Version: 01203da [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    SpeedyDex [speedydex],
]

### Additional context

_No response_

anoobindisguise:
#### Summary
Bugfixes "badges don't get in the way of equipment"

#### Purpose of change
Fixes #65339
Badges covered the Torso on OUTER, which meant they conflicted with basically all armor and couldn't be worn alongside something rigid. They were also comically heavy being nearly a pound each.

#### Describe the solution
Remove coverage. Changed layer from OUTER to BELTED (this doesn't matter, but this is where this would belong if it had coverage). We could introduce a special "badge" location for badges specifically but this is so minor that wearing a badge doesn't seem like it should conflict with anything (why would a badge conflict with a load bearing vest for example?)
Also, lowered the weight back (real badges are 1-4 oz).

We don't need no STINKING badges!!


Maleclypse (approved):
Good catch!

PR #66112, opened 2023-06-10
Cache Character's radiation leak level

SurFlurer:
#### Summary
none

#### Purpose of change
Improve performance. 

`leak_level()` call is expensive, especially it has to be called every turn on every character. So cache it and re-cache on inventory change.

The impact of `leak_level()` on performance can be a lot more severe once #60885 gets merged.

#### Describe the solution
Cache the leak level when character inventory changes. 

Add a dedicated set of functions to get and set the value, to our code spaghetti. This also makes it more easy to potentially add a feature that radioactive items contribute to the environmental leak level.

Mostly, where invalidate_inventory_cache() is called, this leak level cache should also be called. But I avoided reusing that variable, to maintain clarity. I'm unsure whether this is a good practice for this case?

#### Testing

Test results are get by sleeping 9 hours with reasonably many items in character's inventory.

image

image

Before:

image

After:

image

### DIFF
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -1343,6 +1343,7 @@ bool avatar::wield( item_location target )
 bool avatar::wield( item &target )
 {
     invalidate_inventory_validity_cache();
+    invalidate_leak_level_cache();
     return wield( target,
                   item_handling_cost( target, true,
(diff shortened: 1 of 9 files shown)


SurFlurer:
Wow nothing really failed, including the msvc one. I'm thinking about smashing my PC and buying a new one lol. 😆

Issue #65735, opened 2023-05-19
Lighters and matchboxes are unable to light items/tiles ablaze


Irets:
### Describe the bug

On the latest developement version of cdda, matchboxes and lighters are unable to light items ablaze by [a]ctivating them and choosing an adjacent tile.

### Attach save file

[Helvete-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11515705/Helvete-trimmed.tar.gz)
Here is a save of mine where on your right is a pile of butchery refuse, leaves and a zombie child corpse.
I currently have a matchbox & lighter with me which are unable to light the fire, as it says "There's nothing to light there".

A quick way of testing this out on your system is to create a new world, choose Punk Rock Dude as your character as he spawns with a lighter and then attempt to light something flammable on fire through the [a]ctivation menu.

### Steps to reproduce

1. Spawn into a world with either matchbox or a lighter
2. [a]ctivate said item and attempt to light something adjacent to you ablaze
3. Fail the above and be met with "There's nothing to light there".

![maim_1684489690](https://github.com/CleverRaven/Cataclysm-DDA/assets/50279814/65b6c0ee-c5ae-4119-b08e-e7208d889c44)

### Expected behavior

I except items or the adjacent tile to light on fire.

### Screenshots

![maim_1684489690](https://github.com/CleverRaven/Cataclysm-DDA/assets/50279814/43826558-f2a7-40f4-81a7-3330fa0491b6)
![maim_1684490329](https://github.com/CleverRaven/Cataclysm-DDA/assets/50279814/ea037bd8-e285-4c26-96b5-99c038b97b1c)

### Versions and configuration

OS : Gentoo Linux
Game version : 0.G-1998-gc3ffd0cbf0-dirty
Graphics : either
Ingame language : English
Mods loaded : dda, no_npc_food, personal_portal_storms, no_fungal_growth, package_bionic_professions

### Additional context

I am very certain that in a previous version of the game, I was able to burn houses and light piles of butchery refuse on fire, but not in this latest developement version.

NetSysFire:
I wonder if this could be related to #65536.

Irets:
Through additional testing, it became apparent that the issue doesn't affect trees.
Those can be lit with either matches or a lighter, while most other items or tiles can not.

RenechCDDA:
[...]

PR #66116, opened 2023-06-10
Update flammability of some items

RenechCDDA:
#### Summary
None

#### Purpose of change
Partially addresses #65735
Does NOT close the issue.

Adjust some other items not directly mentioned.

#### Describe the solution
-Fix paper material's burn_data
-Add TINDER flag to a bunch of stuff that should have it
-Remove TINDER from the book abstracts because 90% of the entries are various holy books and that looks *really bad*  (All books can still be disassembled into individual bits of paper)

We could add TINDER to all books but I don't think that's really appropriate.

#### Testing
Loaded game with changes applied, paper can be lit with lighters.

#### Additional context
I don't believe the holy book abstract was intended to cause offense, but when you see this page on HHG you'll see why it should change.
image


Maleclypse (approved):
This makes sense

PR #66124, opened 2023-06-10
Make the kerosene recipe more accurate

anoobindisguise:
#### Summary
Balance "Kerosene recipe is less absurd"

#### Purpose of change
The kerosene recipe takes a black box lab equipment piece and turns fat into oil through a known chemical process, fair enough. Problem is it didn't follow this process very much.
- Requisite fat input was wildly variable with you needing 24x as much fat using chunks of fat as you did using tallow. 
- Washing soda was <1% soda by weight when it needed to be 10%.
- Power cost was hilariously low. Fluidic catalytic cracking reactors operate at 10-15 KW and bring their contents to 500c or higher. The cost was only 40 surface_heats.
- Oil production violated conservation of mass. Using 10 tallow, which is 550 grams of fat, it produced 2500 ml of kerosene  and 10000 ml of diesel which is nearly 20x as much oil mass as the input fat.

#### Describe the solution
Fat requirement is standardized at ~10 kilograms. Using raw chunks of fat is slightly less efficient (~10% ish) but not something dramatic like 2400%. Washing soda requirement bumped up to 40 units (1 kg). Water usage lowered from 25 kg water to 10 kg water as I read figures suggesting equal weight water:fat. Power cost increased to 50 MJ which is a far more appropriate representation of 11 kg of material being brought to 550c. The reactor now provides power instead of surface_heat because heating up the reactor with a hotplate was very weird (this reactor is its own thing that uses like 10-15 kw so presumably it has its own heat supply). It is fueled with Large and Very Large storage batteries for the time being (although I believe there are plans to make it griddable)

more or less water usage. even higher power cost relative to the inputs. 50 MJ might be low balling it a little but I wanted it to fit within a storage battery. There isn't a hard-and-fast given number for how much energy and how long it takes to process how much material into hydrocarbons and my current estimate is a little generous though probably within the correct order of magnitude at least.

#### Additional context
Heavily based on https://www.sciencedirect.com/science/article/abs/pii/S0165237016300778?via%3Dihub which is the article the original catalytic cracking PR references.


data/json/items/tool/science.json
```diff
@@ -364,13 +364,16 @@
     "symbol": "U",
     "color": "dark_gray",
     "looks_like": "household_water_heater",
-    "weight": "22700 g",
+    "weight": "500 kg",
```
Karol1223, line 367: At this point, shouldn't it become an appliance? Especially since you made it take large storage batteries.

anoobindisguise: I'm pretty sure someone else is doing that. #64830
hopefully this won't make a conflict for em lol.

PR #66127, opened 2023-06-11
Small changes: to improve perf

SurFlurer:
#### Summary
none

#### Purpose of change
Improve bits of performance.

This is to accumulate some random small changes that do improve performance, but do not deserve a separate PR ( improvement not big enough. )

#### Describe the solution
Use map::find() to replace an iterator.

Also return vectors of reference_wrapper to avoid copying.

Change veh_interact's `inventory` to a const pointer to save space and avoid copying.

Leave it as it is.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1772,8 +1772,9 @@ std::vector<effect> Creature::get_effects_from_bp( const bodypart_id &bp ) const
 {
     std::vector<effect> effs;
     for( auto &elem : *effects ) {
-        if( elem.second.find( bp ) != elem.second.end() ) {
-            effs.push_back( elem.second[bp] );
+        const auto iter = elem.second.find( bp );
+        if( iter != elem.second.end() ) {
+            effs.push_back( iter->second );
         }
     }
     return effs;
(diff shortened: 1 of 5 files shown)


src/creature.cpp
```diff
@@ -1772,10 +1772,8 @@ std::vector<effect> Creature::get_effects_from_bp( const bodypart_id &bp ) const
 {
     std::vector<effect> effs;
     for( auto &elem : *effects ) {
-        for( const std::pair<const bodypart_id, effect> &_it : elem.second ) {
-            if( _it.first == bp ) {
-                effs.push_back( _it.second );
-            }
+        if( elem.second.find( bp ) != elem.second.end() ) {
+            effs.push_back( elem.second[bp] );
```
irwiss, line 1776: If this affects perf you can shave off another lookup for "free"
```suggestion
        const auto iter = elem.second.find( bp );
        if( iter != elem.second.end() ) {
            effs.push_back( iter->second );
```

returning pointers or std::reference_wrapper instead of copying over `effect`s is also an option if callers can be modified to use em

SurFlurer: Thanks!

PR #52242, opened 2021-10-11
Fungal spread forever

kevingranade:
#### Summary
Features "Fungal spread overhaul"

#### Purpose of change
Handling of fungal areas has been problematic for roughly forever due to the requirement that spread only occurs when near the player is nearby, triggered by monster specials, therefore necessitating that the spread occurs FAST.  As a result, spread scales based on how much time the player spends nearby. This sets up a perverse incentive to simply stay away from fungal areas at all times since dealing with them is self-defeating unless you have overwhelming resources.

#### Describe the solution
This completely overhauls how fungal spread occurs, instead treating the entire area as a single massive organism. This necessitates adding some new infrastructure for such a coherent area  at the overmap level so that it can be updated even when the player is not present.  Compensating for the constant spread, it can now occur very slowly and still make meaningful progress.

The growth model is constant rate addition of area, which means that as the area continues to grow, the rate of increase of the *radius* drops off sharply.
The first week of growth is illustrative, assuming one tile per 10 minutes.
Day one, 144 tiles, 6.7 tiles of radius.
Day two, 288 tiles, 9.5 tiles of radius.
Day three, 432 tiles, 11.7 tiles radius.
Day four, 576 tiles, 13.5 tiles radius.
Day five, 720 tiles, 15.1 tiles radius.
Day six, 864 tiles, 16.5 tiles radius.
Day seven, 1008 tiles, 17.9 tiles radius.
Day 30, 4320 tiles, 37 tiles radius.
Day 365, 52,560 tiles, 129.3 tile radius.

The way this growth is realized is somewhat indirect, there is a counter that literally tallies up a count that is "time since cataclysm" divided by some number, something in the ballpark of 10 minutes per tick (configurable in json).  This area is then used to determine the area of effect as a radius, which is then used to determine whether a given tile has been altered.
In order to not simply have a boring perfectly circular area, I've added a feathering function that hashes the absolute coordinates of individual tiles to determine how far the growth needs to advance before that tile is altered.  If you're familiar with the concept this is essentially a noise function.

Independent of the growth rate, a rather complex part of the system is dealing with damage.
Modified tiles are treated as a resource, both absorbing damage and being expended in order to attack the player (this is not implemented yet).  In the fully realized system this is where all fungal "creatures" will come from, a number of tiles will go into some kind of dormant state, and a fungal creature will be produced.  The larger the creature the larger the number of tiles would be expended to make it happen.
[...]

PR #66135, opened 2023-06-11
Rename `No Fungal Growth` mod to `Slowdown Fungal Growth`

GuardianDll:
#### Summary
None
#### Purpose of change
People always think No Fungal Growth means total removing fungal from the game, which is not. Of course no one read the description also.
#### Describe the solution
Rename the mod name in attempt to highlight this
Drop it
#### Additional context
There is probably a better name we can use, but i can't find it

### DIFF
--- a/data/mods/no_fungal_growth/modinfo.json
+++ b/data/mods/no_fungal_growth/modinfo.json
@@ -2,7 +2,7 @@
   {
     "type": "MOD_INFO",
     "id": "no_fungal_growth",
-    "name": "No Fungal Growth",
+    "name": "Slowdown Fungal Growth",
     "authors": [ "eltank" ],
     "description": "Removes the exponential growth ability of fungaloids.",
     "category": "rebalance",


Karol1223:
Why not just 'no exponential fungal growth? like in the description? I feel like that explains it perfectly

GuardianDll:
Because people are not aware fungals has ability to spawn more fungaloids exponentialy, for them "no grow" means "this shitty terrain do not spawn around"
and when they see it indeed spawn, they don't think "hey, i guess i read the mod info wrong", they go to our discord and ask "ayo wtf no fungal mod doesn't work"? (btw mod id should be changed also, but its kinda another pain in ass to deal with)

NetSysFire:
Any reason this is still a mod and not in the base game?

GuardianDll:
We want to improve the fungal grow in the future, and remove this mod entirely, but the task seems to be extremely complicated, see #52242

PR #66140, opened 2023-06-11
gutter_unify tool

Procyonae:
#### Summary
None

#### Purpose of change

#63597 merged gutters but kept the old ids using alias, this tool will be the first step towards removing the ids altogether starting by allowing "|", "2" and "3" definitions to be removed from maps using roof_palette. Will make follow up PRs if reviewer is happy that the tool works in all use cases, applying this to a reasonable amount of files at a time as the total changes are roughly in this region - [branch comparison](https://github.com/Procyonae/Cataclysm-DDA/compare/GutterSymbolMerge...Procyonae:Cataclysm-DDA:GSMTest2)
image

#### Describe the solution

 - The tool takes all json files in the directory given then checks each object in the file
- Next it checks if the object has been deserialized as a dict to avoid files like mods/replacement.json from passing lists to the next step
- Then it checks that the object is a mapgen, declares at least one palette, defines it's rows and that roof_palette is used
- Next if roof_palette isn't the only palette used it adds the map to an message to be send at the end to be manually changed
Otherwise it declares the terms we're looking for, "|", "2" and "3" and if there's a terrain object we check if the palette definitions are overridden for any of them in which case we don't want to change that symbol and remove it from terms
- Next as long as all terms weren't overridden we form a regex pattern and iterate through the rows doing a regex search for any of the terms and then replacing them with "-" and marking the file as changed
- Finally it returns the altered file if it's changed to be serialised and formatted

#### Testing

Ran on /json and /mods and checked a good portion of results as well as manually testing override cases

### DIFF
--- a/tools/json_tools/gutter_unify.py
+++ b/tools/json_tools/gutter_unify.py
@@ -1,0 +1,55 @@
+#!/usr/bin/env python3
+
+import argparse
+import json
+import os
+
+args = argparse.ArgumentParser()
+args.add_argument("dir", action="store", help="specify json directory")
+args_dict = vars(args.parse_args())
+
+
+def gen_new(path):
+    change = False
+    with open(path, "r", encoding="utf-8") as json_file:
+        try:
+            json_data = json.load(json_file)
+        except json.JSONDecodeError:
+            print(f"Json Decode Error at {path}")
+            print("Ensure that the file is a JSON file consisting of"
+                  " an array of objects!")
+            return None
+        for jo in json_data:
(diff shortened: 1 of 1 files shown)


andrei8l:
Since this is a migration tool that only needs to run once, it doesn't need to be included in the repo. Just PR the results and link the code in the description.

Procyonae:
I were including it bc it can be changed to do other maps (2nd floors etc) in the future and used for out of repo stuff currently, if that's not relevant I'll happily close this though ^^

bombasticSlacks:
We have lots of other tools lingering from years past, you've been on a real tear going through maps, modders may stumble upon this and use it. 

So I'll merge it.

PR #66154, opened 2023-06-12
Shorten molt effect descriptions

### LINKED ISSUE #66143: Molting Effects Cut Off
### Describe the bug

The molting process added by #65459 added some molting related effects to inform the player they are going to molt and to get to a safe place.  Unfortunately, they are all cut off in the effects window, and so a player will not know the danger of molting.  There is no way to scroll this effect to read it, and as far as I know, effects do not appear elsewhere for the player to see.  Screenshots below.

### Attach save file

N/A

### Steps to reproduce

Debug one of the Carapace mutations
Use debug time travel and move forward until you get a message in the log that you will need to molt soon.
Check new molting related effect in the character menu
Observe that it appears to be only given one line and that it is too long to be read, even at normal scaling/font settings
Be sad

### Expected behavior

The point of the effect is to relay to the player that they will be immobilized by the molting process, as being immobilized for 5-15 minutes is pretty devastating if you're out in the zombie apocalypse.  But because it's cut off, players will not know this.  This will inevitably lead to people being killed and frustrated by the molting system -- which is a really neat mechanic that we don't want people to be mad about.

### Screenshots

image
image

- OS: Windows
- Game Version: 615c246 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda]
]

### Additional context

There also appears to be a typo in the regular molting effect, as seen in the screenshot above it says "somewheret" It's either a typo or there's supposed to be a linebreak and so it jammed the two words together.

Additionally, for those of us that scale our games, this lack of scrolling will be even more of an issue, as we'll see even less of the effect description.

worm-girl:
#### Summary
Bugfixes "Shorten molt effect descriptions"

#### Purpose of change

There's a six line limit in effect descriptions, and five lines are taken up by the stat and limb score mods in the molting and molting_imminent effects. This was causing the description to not fit in the @ screen. Since this description contains important gameplay information, it needs to be readable.

closes #66143 

#### Describe the solution

I shortened the description to be a single line long, even at 2x magnification.

Figuring out a way to make it so that pressing enter on an effect opens a more detailed view. But I'm bad at C++ so I'm not doing that.

#### Testing

Opened the game, checked out the description.

### DIFF
--- a/data/json/effects.json
+++ b/data/json/effects.json
@@ -4219,7 +4219,7 @@
     "type": "effect_type",
     "id": "molting",
     "name": [ "Molting" ],
-    "desc": [ "Your exoskeleton feels uncomfortably tight.  Perhaps you should find somewhere to rest." ],
+    "desc": [ "You feel a strong urge to hide somewhere safe." ],
     "apply_message": "Your chitin plating itches against your body.",
     "rating": "bad",
     "base_mods": {
@@ -4239,7 +4239,7 @@
     "id": "molting_imminent",
     "name": [ "Imminent Molting" ],
     "desc": [
-      "Your exoskeleton has grown so tight and rigid that it grips you like a vise.  You must seek shelter for the coming molt."
+      "Seek shelter, you will be helpless when it begins."
     ],
     "apply_message": "You can hardly move in your tightening exoskeleton, and you feel a strong impulse to hunker down somewhere safe.",
     "rating": "bad",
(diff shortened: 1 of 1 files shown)


GuardianDll:
It just partially fixes the issue, i guess there could be more of such long effect descriptions, especially for small screens

fairyarmadillo:
> It just partially fixes the issue, i guess there could be more of such long effect descriptions, especially for small screens

This effect will straight up kill you if you are in a dangerous area when it finishes. It's critical that the player is able to look at it and understand that immediately, so a fix needs to get merged ASAP.

I have suggested a more holistic fix here in the PR and in the attached issue, but given that most other effects aren't major emergencies like this one, I don't think we should delay fixing this issue until somebody gets around to making a major change to how the stat screen works.

PR #66155, opened 2023-06-12
added zombie sheep and goat

Skelly-Boi:
#### Summary
Content "Add a zombie sheep and goat"

#### Purpose of change
The sheep and goat lacked a zombie variant, despite being big enough to turn into one.

#### Describe the solution
Added a zombie variant and 'zombify_into' to the animals 

Changing the armor or other values around. I tried not to make them diverge too much from the original animal. The reason the sheep's vision is much lower than the normal zombie's is due to it's fur and filth covering it's eyes.

#### Testing
Spawned a goat and sheep, killed them, waited, and they revived without issue.


data/json/monsters/zed-animal.json
```diff
@@ -465,6 +465,68 @@
     "dodge": 0,
     "armor": { "bash": 1, "electric": 2 }
   },
+  {
+    "id": "mon_zombie_sheep",
+    "type": "MONSTER",
+    "name": { "str": "zombie sheep" },
```
Qrox, line 471:
```suggestion
    "name": { "str_sp": "zombie sheep" },
```

Qrox:
Perhaps we should also have an upgraded version of zombie goats that have long tongues and pull attack :wink:

harmez7:
do I need to start a new game to see these new zeds or i can encounter them in my current save ?

PR #66157, opened 2023-06-12
[Innawood] Fix for arsonist's quest

Light-Wave:
#### Summary
None

#### Purpose of change
The Innawood arsonist quest forgot to overwrite a piece of dialogue, so the quest could not be completed without gasoline.

#### Describe the solution
Overwrite the missing dialogue piece so that the quest requires tinder as intended.

### DIFF
--- a/data/mods/innawood/npcs/missiondef.json
+++ b/data/mods/innawood/npcs/missiondef.json
@@ -497,5 +497,19 @@
       "place_item": [ { "item": "tinder", "x": [ 5, 20 ], "y": [ 5, 20 ], "amount": 5, "repeat": 40 } ],
       "place_fields": [ { "field": "fd_fire", "x": [ 5, 20 ], "y": [ 5, 20 ], "repeat": 40 } ]
     }
+  },
+  {
+    "//": "MISSION_PYROMANIAC mission set this topic as starting topic for mission giver.",
+    "type": "talk_topic",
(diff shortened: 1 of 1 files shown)


mqrause:
Tried it out and I confirm it lets me complete the mission. Apparently also on the tiles next to the mission target, but that might be intended?
However it also gives me triple the options for "we're getting there soon". Not sure if that's because I already had the mission before I applied this fix. (Also "can't wait to burn that building".)
grafik

Plus completing the mission spreads gasoline around to fuel the fires.

Maleclypse:
> Tried it out and I confirm it lets me complete the mission. Apparently also on the tiles next to the mission target, but that might be intended? However it also gives me triple the options for "we're getting there soon". Not sure if that's because I already had the mission before I applied this fix. (Also "can't wait to burn that building".) grafik
> 
> Plus completing the mission spreads gasoline around to fuel the fires.

So do you feel this fixes the quest or creates more bugs?  Open question to anyone.

mqrause:
I'd say the bugs are pretty minor compared to being able to complete the mission at all.

Light-Wave:
Good catches. I changed it so that no gasoline spawn and you only get two options to tell him to be patient. Not sure where the second remaining one comes from. Also, he now talks about that tree instead of that building.

I think that's good enough for now.

PR #66158, opened 2023-06-12
Delete cground.json

Procyonae:
#### Summary
None

#### Purpose of change

cground.json contains a single palette cground which isn't used or referenced by anything in repo inc mods and obsoleted stuff

#### Describe the solution

Just deletes it


Maleclypse:
Searched using sublime.  No other uses of "cground" appeared.

JReese3:
Found something regarding constructing natural ground, search turned up reddit which is down, can't confirm at this time.

PR #66159, opened 2023-06-12
Unhardcode refugee centre locate mission

Procyonae:
#### Summary
Infrastructure "Unhardcodes refugee centre locate mission and adds a new field to mission definitions start"

#### Purpose of change

Unhardcode = good?

#### Describe the solution

Who knows


Procyonae:
I need a C++ course or something closing bc I'm smol brain, will probably come back to it at some point unless anyone else wants a crack, I just wanted to add a new reveal road field to guarantee revealing road rather than use the 1/3 chance in order to mimic current hardcoded behaviour (might have changed it into an array with [ om, one in x chance ] once I had it working and removed the 1/3 chance entirely.

Issue #65114, opened 2023-04-16
Unify items with delayed actions


Hirmuolio:
### Is your feature request related to a problem? Please describe.

Items can have delayed actions where after X time passes from item being activated something happens.

The problem is that there are currently 3.5 different ways to define this in json.

The different methods of delayed actions:

**1\. Using `transform` with charges.** This is used by most grenades.
The ability to do item transform makes it very versatile.
item -> activated by player -> item transforms -> timer starts -> `use_action` on item activates

Problems: 
* Using charges for this is pretty bad thing and it would be great if charges were not used like this.
* the timer runs only while the item is in reality bubble.
* Can't start on its own and must be started by player activating the item (?).

```
"use_action": {
      "need_wielding": true,
      "target": "grenade_act",
      "msg": "You pull the pin on the grenade.",
      "target_charges": 5,
      "active": true,
      "menu_text": "Pull pin",
      "type": "transform"
    },
```

**2\. Using `delayed_transform`.** This is mostly used by cooking things that "mature" on their own.
item is in active state -> age is checked -> item transforms.

This fixes some of the above problems.
* It is based on item age. So no charges involved and time passes out of reality bubble.
* Does not require player activation.

But overall it can't be used in place of `transform` because it can't trigger `use_action`. It can only transform.

```
"use_action": {
      "target": "milk_cream_jar",
      "msg": "You examine the batch and see that the cream has risen to the top, so you seal the jar up for storage.",
      "moves": 50,
      "type": "delayed_transform",
      "transform_age": 21600,
      "not_ready_msg": "The cream has not risen yet."
    }
```

**3\. `countdown_interval`.** Used by few bombs, mi-go armor and flaming arrow.
item is in active state -> timer starts -> specified action is triggered.

This almost has feature parity with the item transform method. Main benefit being that it does not require player activation.

Problems: 
* Can't require wielding the item before activation.
* Transforming item can add many new features to the item (name change, description change, activation msg change etc.).
* Still only runs in reality bubble,

```
"countdown_interval": 86400,
"countdown_action": { "type": "transform", "target": "migo_plate_overgrown", "msg": "The iridescent carapace plate expands as it grows." },
```

**3.5\. `"type": "countdown"`.** There is literally only one item in the whole repository that uses this (dragon hand scroll in MMA mod).
Before anyone says that one of the grenades also uses it: It does not use it because the copy-from always overrides that action.

[...]

PR #66161, opened 2023-06-12
Refactor timer items to be a bit more time based

Hirmuolio:
#### Summary
Infrastructure "Refactor timer items to be a bit more time based"

#### Purpose of change

 #65114

Items that activate with timer (grenades and such) use item charges to keep track of their timer. This is bad. Use time for tracking time instead.

#### Describe the solution

Use time_points instead of item charges to determine when item timer should activate.

Since it is time based it won't lose track of time out of reality bubble. They still won't explode when they are outside. But if you set C4 timer to 1 day, leave, and come back 1 day later it will explode at the correct time.

Two ways to define this timer: Adding `"target_timer"` in item transform starts timer when player activates item. `"countdown_interval"` starts timer when item is created.

Removed `"countdown"` item use action. Having multiple ways to start timer was messy. Only one item used it and the functionaly can be replicated with item transform simply by  transforming item into itself.

Since delayed action is no more same as player activation a new `"message"`action was added. Purely cosmetic thing used to tell the player they can't pull the pin on active grenades. Could be removed if it is considered unneeded.

`revert_to` could be defined for tools only. It defines what tool reverts to when to runs out of charges or completes its timer. Now it can be defined to all items and defines what they may revert to after timer.
Items without `revert_to` are deleted at the end of their action. Items that should just stay can revert into themselves.

`"countdown_action"` (probably) works only with actions defined in `iuse_actor.cpp`. The actions defined in `iuse.cpp` interpt the meaning `bool t` differently so they don't work.  
So `"explosion"` and `"transform"` work but `"MOLOTOV"` does not work. Hence those items will still use the old system.

Minor regression: In the old system item with explosion action could emit noise during the explosion countdown. That is not possible anymore as the explosion action only happens at the end of timer instead of happening every turn. So few explosives where this sound was important were not migrated to the new system. Explosives where the sound was not particularly important were migrated and sound was removed.

Rewrite `itype::tick` and item processing that follows from there so that player activation and item processing activation are separate things and stop causing problems to each other.

#### Testing

Grenades, molotovs and basic tools seem to work fine.
Active grenades and iridescent carapace plate migrate to the new system on save load.
Item timers stay in save-load cycle.

#### Additional context

Item activation system is one massive sphagetti mess where one thing can be used in two different places for two different purposes and all three systems are intertwined.


data/json/items/tool/explosives.json
```diff
@@ -108,17 +108,10 @@
     "material": [ "paper", "powder" ],
     "symbol": "*",
     "color": "red",
-    "turns_per_charge": 1,
+    "use_action": { "type": "message", "message": "You've already lit the %s, try throwing it instead.", "name": "Light fuse" },
     "explode_in_fire": true,
     "explosion": { "power": 938 },
```
Brambor, line 113:
```suggestion
    "explosion": { "power": 938 },
    "flags": [ "TRADER_AVOID", "BOMB" ],
```

data/json/items/tool/explosives.json
```diff
@@ -164,17 +156,10 @@
     "material": [ "paper", "powder" ],
     "symbol": "*",
     "color": "red",
-    "turns_per_charge": 1,
+    "use_action": { "type": "message", "message": "You've already lit the %s, try throwing it instead.", "name": "Pull pin" },
     "explode_in_fire": true,
     "explosion": { "power": 938, "shrapnel": { "casing_mass": 1597, "fragment_mass": 0.5 } },
```
Brambor, line 161:
```suggestion
    "explosion": { "power": 938, "shrapnel": { "casing_mass": 1597, "fragment_mass": 0.5 } },
    "flags": [ "TRADER_AVOID", "BOMB" ],
```

Brambor (commented):
I think you didn't mean to remove the "flags".

Hirmuolio:
I think this is ready now.

andrei8l:
Needs a test unit

src/item.cpp
```diff
@@ -13231,9 +13239,19 @@ bool item::process_internal( map &here, Character *carrier, const tripoint &pos,
...
+            if( type->countdown_action ) {
+                type->countdown_action.call( carrier ? *carrier : get_avatar(), *this, false, pos );
+            }
+            countdown_point = calendar::turn_max;
+            if( type->revert_to ) {
+                convert( *type->revert_to );
+
+                if( !active ) {
+                    active = needs_processing();
+                }
```
andrei8l, line 13253: What's the point of this? `needs_processing()` returns true if `active` is true so this is circular. The other conditions don't need to be mirrored in `active`.

kevingranade:

https://discourse.cataclysmdda.org/t/id-like-to-ask-some-questions-about-the-latest-experimental-version-of-debug/28938/2

PR #66168, opened 2023-06-13
vehicle_wrench => vehicle_wrench_2 in rams

nornagon:
#### Summary
Bugfixes "Fixed vehicle rams not requiring a wrench to install"

#### Purpose of change

There is no "vehicle_wrench" requirement.

#### Describe the solution

These are big rams so probably you need a big wrench to install and remove them.

vehicle_wrench_1 maybe?


irwiss (approved):

anoobindisguise:
I'll try to keep an eye out in case I screwed up anything else when fixing merge conflicts. thanks for fixing this, again.

PR #66169, opened 2023-06-13
Changing EOD helmet to normal layer to prevent wearing gaskmasks underneath the faceshield.

FFValor:
Changed the EOD helmet being outer layer due to the ability to wear a gasmask underneath the face mask of the helmet.

#### Summary
Features "EOD helmet layer change"

#### Purpose of change

EOD helmets able to be worn with gasmasks without conflict, makes no sense.

#### Describe the solution
Changed the OUTER tag to NORMAL. This prevents gas masks and EOD helmets being worn at the same time.

#### Testing
Spawned and wore the full EOD suit, no conflicts.

The layer change of the EOD helmet may have strange or unintended effects due to its built in nape protection.

### DIFF
--- a/data/json/items/tool_armor.json
+++ b/data/json/items/tool_armor.json
@@ -1201,7 +1201,7 @@
       "need_charges_msg": "\"Illumination disabled, low power.\""
     },
     "material_thickness": 15,
-    "flags": [ "STURDY", "OUTER", "RAINPROOF", "TWO_WAY_RADIO", "WATCH", "ALARMCLOCK" ]
+    "flags": [ "STURDY", "NORMAL", "RAINPROOF", "TWO_WAY_RADIO", "WATCH", "ALARMCLOCK" ]
   },
   {
     "id": "helmet_eod_on",


OmegaCenti:
EOD-9-Breathing-Apparatus-Faceshield-Med-Eng-SCBA-Forward

harakka:
Linking to the source of that image would have been a lot more useful than just dropping the image: https://www.med-eng.com/product/eod-9-breathing-apparatus-visor/
So what that is, is a replacement visor for a particular EOD suit that is designed to accomodate (specific) gas masks and breathing apparatus. That doesn't mean every random EOD suit will accomodate any gas mask.

randomtyper:
Given the thicknesses of actual EOD helmets (more than 3 mm of hard shell + 10+ mm of padding + air space for ventilation and onskin headwear like balaclavas), it should take both outer and normal layers when covering the head. The face cover is actually quite far from the face (helmet thickness + at least 2 cm), so it is reasonable to assume one can fit at least a slim gas mask underneath.

bombasticSlacks:
Solved in #66179

PR #66177, opened 2023-06-13
fix vpart removal requirement check

nornagon:
#### Summary
None

#### Purpose of change

Looks like vehicle_part's removal requirements weren't being checked. Instead, install requirements were being checked twice.

#### Describe the solution

Fixed the typo.

### DIFF
--- a/src/veh_type.cpp
+++ b/src/veh_type.cpp
@@ -827,7 +827,7 @@ void vpart_info::check()
             }
         }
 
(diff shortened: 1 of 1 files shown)


nornagon:
should prevent things like https://github.com/CleverRaven/Cataclysm-DDA/pull/66168 from happening again.

RenechCDDA:
Looks to have been obsoleted by #66295

PR #66180, opened 2023-06-13
math_parser: implement ternary operator

andrei8l:
#### Summary
None

#### Purpose of change
* @Light-Wave asked for inline boolean operators in `math`
* I also considered using ternary operators for #65389

#### Describe the solution
Implement. The ternary operator is right-associative and should work exactly like the C equivalent.

Simple examples:
```JSON
    "//0": "returns 5 if u_blorg is greater than 4, otherwise 0",
    { "math": [ "( u_blorg > 4 ) * 5" ] },
    "//1": "returns rng( 0.5, 5 ) if u_blorg is greater than 5, otherwise rand(100)"
    { "math": [ "u_blorg > 5 ? rng( 0.5, 5 ) : rand(100)" ] },
```

Piece-wise functions in `eoc_math`? They'd be bulkier for simple things, but much more versatile and easier to read. Can still be done in addition to this in the future, if the need arises.

#### Testing
Updated test unit

Only fuzzed for a few hours this time.

#### Additional context
N/A

### DIFF
--- a/doc/NPCs.md
+++ b/doc/NPCs.md
@@ -1430,6 +1430,19 @@ Common math functions are supported:
 
 Function composition is also supported, for example `sin( rng(0, max( 0.5, u_sin_var ) ) )`
 
+#### Ternary and inline boolean operators
+Inline comparison operators evaluate as 1 for true and 0 for false.
+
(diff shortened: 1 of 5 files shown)


andrei8l:
Thanks for merging!

PR #66181, opened 2023-06-13
Expand mutation transforms

Venera3:
#### Summary
Features "Allow mutation transformation using the normal mutation rules"

#### Purpose of change

While starting on limb stuff proper I decided to lean heavily onto (among other things) `condition->transform` to handle the trait progressions. For that to work it needed some more infrastructure.

#### Describe the solution

Added the `safe` variable to mutation `transform` data - when true instead of flatly unsetting the base trait and adding the new one it routes through the actual mutation logic, removing conflicting traits/checking thresholds etc.

Include it in an upcoming limby PR but it can stand on its own (and is small and non-limby enough)

#### Testing

Added `safe` to a transformation, it failed to give me the desired traits when missing the required threshold, gave the trait when all requirements were met.

### DIFF
--- a/doc/MUTATIONS.md
+++ b/doc/MUTATIONS.md
@@ -213,7 +213,8 @@ Note that **all new traits that can be obtained through mutation must be purifia
     "target": "BIOLUM1",                      // Trait_id of the mutation this one will transform into.
     "msg_transform": "You turn your photophore OFF.", // Message displayed upon transformation.
     "active": false,                          // If true, mutation will start powered when activated (turn ON).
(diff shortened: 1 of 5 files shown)


fairyarmadillo:
I have #66068 in to solve what I think is an issue related to this. That PR might not be necessary if this fixes things, unless people like the flavor.

Venera3:
This is explicitly for the `transform:` array, so it wouldn't fix that - and that PR looks like a nice bunch of body horror in any case.

Issue #66172, opened 2023-06-13
Better cache vehicle mass center


SurFlurer:
### Is your feature request related to a problem? Please describe.

`map::vehmove()` has to call `vehicle::calc_mass_center()` multiple times in a turn. This results in around 20% of total overhead for driving large deathmobiles. It seems that the code already handles some cache of the mass center. However, it is clearly not particularly good at reducing overhead.

![image](https://github.com/CleverRaven/Cataclysm-DDA/assets/22912139/5df94dea-ee77-4310-be6a-e98e00003cbc)

![image](https://github.com/CleverRaven/Cataclysm-DDA/assets/22912139/7ebda5ee-eeed-479c-a77e-8174fd41c635)

### Solution you would like.

Have a better cache of the mass center. This change can hopefully save 75% of this function's overhead per turn.

Maybe we should add a set of different functions to "cheaply" handle mass changes caused by fuel drain?

### Additional context

This is somewhat meaningful to finish this to smooth the transition towards the removal of charge from solid comestibles.

SurFlurer:
Now what's left to do is to add a fast way to recalc mass center for every-turn consumption of fuel.

PR #66188, opened 2023-06-14
Better cache vehicle mass center

SurFlurer:
#### Summary
none

#### Purpose of change
Address #66172.

#### Describe the solution
Rearrange program flow to reduce calls on that func.

1. Change `vehicle::total_mass()` to be equal to `vehicle::local_center_of_mass()`. 

The reason is that both of these two functions are called more frequently than `vehicle::rotated_center_of_mass()`, the latter only called on collision. So this change can better utilize the cache.

2. Fix a bug that thrusting vehicles have to pay for alternator fuel consumption twice. 

`vehicle::thrust()` in `vehicle::gain_moves()` has already consumed fuel for alternator loads, so do not consume it again in `vehicle::idle()`. This change also makes vehicle mass get invalidated less frequently, thus  `sfx::do_vehicle_engine_sfx()` do not have to refresh the mass center.

#### Testing

The game goes more fluently while driving.


bombasticSlacks:
Why did the tests need updating?

SurFlurer:
> Why did the tests need updating?

Because this pr: 
> Fix a bug that thrusting vehicles have to pay for alternator fuel consumption twice.

SurFlurer:
Thanks for merging!

PR #66202, opened 2023-06-14
Make hub01 globally unique

### LINKED ISSUE #50861: Unknown Recipient mission spawned two Hub 01 structures
### Describe the bug

When I started the Unknown Recipient mission, I was given the coordinates of a location that was very far away from the refugee center, labelled Hub 01. On my way there, I stumbled upon another place, also named Hub 01, where I was able to turn in the FEMA data as if it was the real hub. Turning in the data did not update the quest marker.

### Steps To Reproduce

Not sure; maybe just try starting the quest again?

### Expected behavior

One of those Hub 01s shouldn't exist.

### Screenshots

fakeHub
The location I was originally told to go to
realHub
The closer hub

- OS: Windows 10
- Game Version: 9a10b0c
- Graphics version: Tiles
- Ingame language: English
- Mods loaded: Default

### Additional context

_No response_

### LINKED ISSUE #57987: Completing Missions at Multiple Hub 01s gets "Retrieve Field Data" Stuck and Prevents Interacting with Intercom.
### Describe the bug

Somewhere along the way as I sporadically worked on the Hub 01 mission chain, I found myself unable to interact with the Intercoms at any new Hub 01s I encountered. Only my most-progressed Hub 01 was working as normal, providing trade and quest options after pushing the intercom button. 
Instead, I get the dialog prompt shown in screenshot 1 whenever I talk to any other intercom. Presenting a Prototype I/O Recorder will allow interaction with the Hub as normal, but leaving dialog and coming back will result in the same request. The mission Return Field Data does not mark as completed after presenting the item at the Hub 01 that initiated that instance of the quest and all other Intercoms also request the item, rendering all but the first Hub 01 nonfunctional. For reproduction steps I provide the approximate order I did the Hub 01 quests, in case it had something to do with causing this issue.

### Steps to reproduce

1. Complete Hub 01 Missions Return Field Data, Iron Safari and Hazard Meteorology, then start Steal a Dead Man's Last Words. all at a single Hub 01 [Hub 1].
2. Visit another Hub 01 and undertake mission Return Field Data. Obtain and turn in the Prototype I/O Recorder.
3. Complete Steal a Dead Man's Last Words, Active Noise Control, and then Light Retrieval at Hub 1. Wait one day and obtain Hub 01 Military armor pieces.
5. Attempt to interact with any Hub 01 instance except Hub 1.

### Expected behavior

Approach any previously not interacted Hub 01 instance and be provided with quest Return Field Data. Return Prototype I/O Recorder to Intercom to enable trading and other unlocked dialog options like asking for HWP magazines.

### Screenshots

image

- OS: Windows
    - OS Version: 10.0.19044.1706 (21H2)
- Game Version: 218edd5 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Translate Complex Dialogue [translate_dialogue]
]

### Additional context

Debug menu cannot complete the Return Field Data mission. Deleting or failing the mission through debug results in its reissuing at any Hub 01 as normal, but the new Return Field Data mission likewise will not be completed when the recorder is turned in and the problem will persist.

### LINKED ISSUE #61289: Permanently broke my Robofac interaction by interacting with a second robofac
### Describe the bug

When finding a second Hub01 after having interacted with the first, they will require you to do the prototype robot quest again in order to talk to you. If you accept this quest, then it will permanently break all future interaction with Hubs01, as the quest will never complete and they will request another copy of the prototype I/O recorder every time you speak with them, and as a result you will only ever be able to speak with them once after doing this.
I also found that, at least in this specific state of broken, you can pass the 10 int "the armor seems warped [...]" check an unlimited number of times and receive an unlimited number of prototype vambraces and greaves. 

### Steps to reproduce

1. Complete the Hub01 prototype robot quest.
2. go find a second hub01.
3. find they want you to do the prototype robot quest again in order to interact with that specific hub01.
4. Accept the quest, bring back the I/O recorder, find out that ALL hub01s will demand another I/O recorder and the quest will never complete, permanently breaking all interactions with Hub01s.

### Expected behavior

Tbh hub01 should be a global exclusive, as unlike the refugee center there's no issue with migos/fungus/triffids spawning right next to them and ruining the location.

### Screenshots

_No response_

- OS: Windows
- Game Version: 90fc2e4 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Slots [cbm_slots]
]

### Additional context

_No response_

anoobindisguise:
#### Summary
Bugfixes "make Hub01 globally unique"

#### Purpose of change
Fixes #61289
Fixes #50861
Fixes #57987
There is now precedent for globally unique factions (Refugee center) and Hub is one of the least affected by the whims of RNG surrounding them in fungals, ants or triffids due to being an impenetrable underground bunker. Multiple hubs caused tons of issues, such as the above, or the development of the ancilla bar (only one single hub can ever get the bar and it can easily be one of the many hubs which you cannot interact with)

#### Describe the solution
Make it globally unique.

Fix all the bugs associated with multiple hubs and then decide to live with multiple hubs.

#### Testing
Some of the missions pointing to the hub might need to have their search range expanded.

### DIFF
--- a/data/json/overmap/overmap_special/specials.json
+++ b/data/json/overmap/overmap_special/specials.json
@@ -4572,7 +4572,7 @@
     "city_sizes": [ 1, 16 ],
     "occurrences": [ 33, 100 ],
     "rotate": false,
-    "flags": [ "UNIQUE", "SAFE_AT_WORLDGEN", "MAN_MADE" ]
+    "flags": [ "UNIQUE", "SAFE_AT_WORLDGEN", "MAN_MADE", "GLOBALLY_UNIQUE" ]
   },
   {
     "type": "overmap_special",


bombasticSlacks:
@John-Candlebury you should weigh in

For what my opinion is worth (not much), I think this is good.

PatrikLundell:
I maintain my position that globally unique shouldn't be used until the environmental issues have been solved.

While the bunker itself would be safe from fungi, the above ground structures would not. They'd destroy the terminal, and thus disable all useful interactions with Hub01 (I assume you can still invade the bunker and kill everyone, but I don't consider that to be a useful interaction).
Even if the terminal was somehow rendered immune to the destructive power of the fungus, the terminal would effectively be unreachable until means to exterminate the fungus at a rate higher than its spread and regrowth was implemented, and that would probably be available only to late game players anyway.

The quests would probably be broken by the fungus, and the NPCs would die shortly after the reality bubble covers them.

Knut-Aage-Hofseth:
I agree with PatrikLundell. I want Hub01 to be globally unique, but the issues that might make the hub inaccessible in a run needs to be resolved before that is done.

bombasticSlacks:
Did either of you read the attached issues? That already happens because interaction with hub 01 is already globally unique, just more buildings spawn with a non functional intercom and NPC duplicates.

PatrikLundell:
When is this "ancilla bar" thing generated? At the time the facility is placed on the overmap or at the time the map actually is generated (i.e. when it enters the reality bubble)? If it's somehow generated only in the first map generated, the logic behind that might allow for the generation of wrecked instances for all subsequent ones (until such a time uniques can be generated a single time only without risk of it being wrecked by the RNG).

If it is the former I agree multiple instances are useless, while if it is the latter you'll get it generated at one you can actually reach.

All of the issues caused by trying to interact with multiple instances of the "same thing" can be worked around by only interacting with one and sticking to that one, while a unique instance that can't be reached and interacted with is completely lost. It's bad that new players can get screwed by trying to interact with redundant instances, but it's even worse that everyone can get screwed by an unlucky world generation.

Knut-Aage-Hofseth:
> Did either of you read the attached issues? That already happens because interaction with hub 01 is already globally unique, just more buildings spawn with a non functional intercom and NPC duplicates.

I did. I want the building to be global unique. I am worried about Hub01 being broken by a nearby fungal spawn or something similar that destroys something important there. If something like this can't happen, I have no more worries. Today, if the first Hub01 you find is destroyed before you interact with it, you can just look for a new one and start interacting with that instead.

I might be completely wrong. If so I apologize for derailing.

John-Candlebury:
Although its not a complete solution I think this is the best path forward at the time, after all the faction is already partially unique as was mentioned before.

For the cases when the surface is dangerous both Hub and The Terminal should be accessible from the underground subway, I can see to making both access points  more fortified and also adding a functional intercom at the subway level.

John-Candlebury (approved):

stubkan:
Re:  Some of the missions pointing to the hub might need to have their search range expanded.

Since there is already a record of the HUB being generated (since the game tracks that it was generated) might it be an idea to add its coordinates in that record and have mission code check for this before the intensive scanning of overmaps - it seems this would optimize things for future missions after unique locations are generated once.

PR #66204, opened 2023-06-14
Woodlouse harvests for crustacean samples

anoobindisguise:
#### Summary
Balance "woodlouse harvests for crustacean samples"

#### Purpose of change
Woodlice (pillbugs) are crustaceans, not insects, despite being visually similar to insects.

#### Describe the solution
Make them dissect for crustacean samples.

This is HIGHLY debatable and I figured I'd pose the question here in the form of a PR. Should woodlice actually dissect for crustacean samples? They are not really similar to crabs and if you mutated into a human-pillbug hybrid most people would say that you were a bug person, not a crab person. But taxonomically speaking they aren't insects...

### DIFF
--- a/data/json/monsters/insect_spider.json
+++ b/data/json/monsters/insect_spider.json
@@ -3597,7 +3597,7 @@
     "vision_day": 10,
     "vision_night": 15,
     "harvest": "mutant_shellfish",
-    "dissect": "dissect_insect_sample_single",
+    "dissect": "dissect_crustacean_single",
     "weakpoint_sets": [ "wps_arthropod", "wps_arthropod_beetle" ],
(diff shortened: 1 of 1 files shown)


Maleclypse (approved):

PatrikLundell:
A lot of people would say arachnids are insects as well...

I think the change is a good one.

fairyarmadillo:
90% of the crustacean mutation is gills, a tough outer shell, multiple legs, and the ability to eat garbage. It's very crab-flavored, but I don't think lack of claws should disqualify pillbugs, they're crustaceans after all. Cephalopod mutants have a nautilus shell but the monsters you get them from mostly don't.

There is a broader conversation to be had about what should drop samples as it seems very arbitrary, but I'm not sure there's a good and logical answer that would provide players with enough samples.

PR #66206, opened 2023-06-15
Add obsidian config to the ignore list

bombasticSlacks:
#### Summary
None

#### Purpose of change
I use [obsidian](https://obsidian.md/) for documentation. Don't want to mistakenly merge my config files.

#### Describe the solution
Add .obsidian to the ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -144,6 +144,9 @@ __pycache__
 # vimdiff backup
 *.orig
 
+# obsidian.md files
+.obsidian
+
 # OSX
 Cataclysm.app
 build-data/osx/AppIcon.icns


Procyonae:
What's the problem with just adding it to your local gitignore? (Coming from a position of ignorance not suggesting this shouldn't happen)

bombasticSlacks:
> What's the problem with just adding it to your local gitignore? (Coming from a position of ignorance not suggesting this shouldn't happen)

Honestly, I thought the local git ignore was global, which was a non-starter since other repos I use keep an obsidian config. That said, I was wrong about that.

However, Obsidian is a popular MD editor so I think this is worth including :)

PR #66208, opened 2023-06-15
Changed order of fields to appease clang-tidy.

katemonster33:
#### Summary
Bugfixes "clang-tidy complained of field order issues in struct mtype in mtype.h, I have re-arranged them."

#### Purpose of change

To fix an issue thrown by clang-tidy that was missed the last time this file was merged. I see the issue because I have a different PR up that has modified CMakeLists.txt and clang-tidy wants to check all files.

#### Describe the solution
Change the order of the fields

Could have NOLINT'ed the warning, I saw that done in another class, but it looks like mtype is used often.

#### Testing
I compiled the code. I can't test clang-tidy locally.

### DIFF
--- a/src/mtype.h
+++ b/src/mtype.h
@@ -265,66 +265,130 @@ struct mount_item_data {
 struct mtype {
     private:
         friend class MonsterGenerator;
-        translation name;
-        translation description;
-
-        ascii_art_id picture_id;
-
-        enum_bitset<m_flag> flags;
 
         enum_bitset<mon_trigger> anger;
         enum_bitset<mon_trigger> fear;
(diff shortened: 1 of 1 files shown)


irwiss:
mtypes are passed by ref almost exclusively and there's ~1000 instances if you load all major mods at 32 byte padding is savings of like 30kb of memory for the price of messing up field order, imo NOLINT is way more practical

Not that the field order is super ordered to begin with but 🤷

katemonster33:
I don't disagree, but we clearly had done the field sorting on this file once before. I prefer readability myself but I am just trying to follow what others have done before me.

PR #66209, opened 2023-06-15
Adding relevant recipe requirements to atomic butter

PolygonalPerson:
#### Summary
Bugfixes "Adds missing recipe requirments to atomic butter"

#### Purpose of change
Allows atomic butter to act as a kind of butter.

#### Describe the solution
Added any_butter & any_butter_or_oil to the requirements, and assigned atomic butter to these, allowing it to be used by any recipe that requires those tags.

#### Testing
Loaded into CDDA, checked meat confit for the any_butter tag-- worked! Then check with glazed tenderloins for that any_butter_or_oil tag-- worked! Atomic butter was successfully assigned to both tags.

### DIFF
--- a/data/mods/Aftershock/requirements.json
+++ b/data/mods/Aftershock/requirements.json
@@ -37,6 +37,18 @@
     "extend": { "components": [ [ [ "frost_human_lard", 1 ], [ "alien_lard", 1 ] ] ] }
   },
   {
+    "id": "any_butter",
+    "type": "requirement",
+    "//": "Butter of all types and other future butter substitutes",
+    "components": [ [ [ "atomic_butter", 1 ] ] ]
+  },
+  {
+    "id": "any_butter_or_oil",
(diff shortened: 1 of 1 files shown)


Maleclypse:
So looking at how the others do it 
```
  {
    "id": "edible_lard",
    "type": "requirement",
    "//": "Lard that is 'safe' for consumption.",
    "extend": { "components": [ [ [ "frost_human_lard", 1 ], [ "alien_lard", 1 ] ] ] }
  },
```
Did you check that regular butter could still be used?   I thought that requirements implicitly extended but I wasn't sure about that.

PolygonalPerson:
Good catch! Meat confit seems to work, but all other instances of any_butter or any_butter_or_oil didn't-- atomic butter overwrote it. I'll add extend to each, testing out the change shows that it's playing nice once that's added o7

Maleclypse (approved):

PR #66216, opened 2023-06-15
Process EoCs on non-activated traits as well

Venera3:
#### Summary
None

#### Purpose of change

Minor infrastructure change to allow non-activatable traits to process their EoCs.

#### Describe the solution

Added a check to test activatability of the trait.

#### Testing

Added processed EoCs to a trait before the changes, it did not trigger.
After the change the EoC triggers every `time` as intended.

#### Additional context

More limby infrastructure tweaking.


bombasticSlacks:
Does this break any current mutations? Also needs docs updating.

bombasticSlacks:
Also can't this be handled by having the mutation be forced to always on? Which also conveys it is actively doing something.

Venera3:
Not if I need triggers to trigger the transformation.

Unless I misread the existing processed mutations are all activated - I'll check them explicitly.

Issue #66213, opened 2023-06-15
Basecamp construction attached zones do not respect mirroring and/or rotation


PatrikLundell:
### Describe the bug

When trying to change the version 2 of the livestock expansion into using palette parameters I verified what I've observed in game play earlier, namely that the "place_zones" for NPC_NO_INVESTIGATE zones are placed in nonsensical locations when rotation and/or mirroring of the layout is used.

My verification during my work was to create the livestock 2 chicken coop in a North expansion, verify that it did end up beside the coop, and then create the same building in the NE expansion (no rotation or mirroring), verifying that it did end up on top of the coop.

### Attach save file

N/A

### Steps to reproduce

1. Create a livestock area version 2 in the North position and then build a chicken coop.
2. Bring up the Zone manager and check the location of the NPC ignore zone.
3. Verify that the zone is on the east side of the expansion rather than the west side (where the building ends up when not mirrored).

### Expected behavior

Zone placement should respect rotation and mirroring.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.3086 (22H2)
- Game Version: 0.G-2341-g4cd1bf1a1d [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Bionic Professions [package_bionic_professions]
]

### Additional context

I haven't actually verified that it fails on rotation, but I assume the cause of the error is that the code doesn't handle either mirroring or rotation.

SurFlurer:
Can /confirm . This happends on 0.G too.

PatrikLundell:
My guess is that it is something that simply wasn't thought of when the zone parameter was added, and so wasn't tested. Alternatively, it was missed when rotation/mirroring was added, if it was possible prior to that.

PR #66217, opened 2023-06-15
Validate the updated camp expansion OMT type

### LINKED ISSUE #66215: Faction camp expansion construction result in harmless(?) error messages
### Describe the bug

The following error message is produced when trying to build a log chicken coop in livestock expansion 1: 
DEBUG    : invalid overmap terrain id "faction_base_modular_livestock_coop_log_northeast"

 FUNCTION : convert
 FILE     : C:\Cataclysm-DDA\src\generic_factory.h
 LINE     : 479
 VERSION  : 0.G-2341-g4cd1bf1a1d

A similar message is produced when trying to build a concrete coop in livestock expansion 2 during work to use parameterized palettes, complaining about the new name of the expansion.

Edit:
The same thing happens when expanding a faction base, complete with the replacement of the overmap description with "% nothing":
 DEBUG    : invalid overmap terrain id "faction_base_modular_hub_room1_log_northeast"

 FUNCTION : convert
 FILE     : C:\Cataclysm-DDA\src\generic_factory.h
 LINE     : 479
 VERSION  : 0.G-2341-g4cd1bf1a1d

### Attach save file

N/A

### Steps to reproduce

1. Create a livestock 1 basecamp expansion.
2. Build a (log) chicken coop.
3. Get the error message.
4. See that the expansion name on the overmap has changed from the name given when the expansion was created to "% nothing".

### Expected behavior

Construction of buildings on an expansion should not result in the replacement of the expansion type on the overmap.

### Screenshots

_No response_

- OS: Windows
- Game Version: 0.G-2341-g4cd1bf1a1d [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

I'm modifying livestock expansion version 2 (JSON files only) and saw the error message and the destroyed overmap description and then went on to test with the corresponding construction of the unmodified livestock expansion 1 and observed the same issue. Thus, I don't think the error is related to my work.

I assume any companion recipe construction work (as opposed to zone construction, which probably isn't affected, although I wouldn't bet on that either) in any base camp and expansion to be subject to this.

SurFlurer:
#### Summary
none

#### Purpose of change
Fix #66215.

It turns out that I tested that plow-fix PR too insufficiently. That only fixed plow mission for farm type 1. Farm type 2 is probably still having trouble to start plow field missions.

#### Describe the solution

Verify that the blueprint's "result" string can be a valid oter_str_id. This fix can stop players getting invalid OMT terrain.

Propose a more comprehensive solution towards fixing plow-field missions.  That should be desirable, however that's unlikely to be composed and merged in a short time. To minimize negative impacts on players' save files, I submitted this fix.

#### Testing
Build a farm expansion type 1, and confirm plow fields mission is still available.

Create a livestock 1 basecamp expansion, build a (log) chicken coop, and get no error message.

### DIFF
--- a/src/faction_camp.cpp
+++ b/src/faction_camp.cpp
@@ -3372,7 +3372,9 @@ bool basecamp::upgrade_return( const mission_id &miss_id )
     }
     update_provides( bldg, e->second );
     update_resources( bldg );
-    overmap_buffer.ter_set( upos, oter_id( bldg ) );
+    if( oter_str_id( bldg ).is_valid() ) {
+        overmap_buffer.ter_set( upos, oter_id( bldg ) );
+    }
     const std::string msg = _( "returns from upgrading the camp having earned a bit of "
                                "experience…" );
 


PatrikLundell:
If I understand it correctly, you update the overland type to allow the plowing code to find where the fields are supposed to be located.

I can think of two ways to address this issue:
1. Modify the JSON format of the construction to indicate that the id should modify the terrain, e.g. "update_mapgen_id" getting a variant e.g. "update_and_set_mapgen_id", where a failure to find the id should be indicated as an error. Note that this method has the inherent disadvantage of requiring a rigid progression where each step requires the previous step to be included.
2. Use zones instead, so the plowing is done where such zones are placed (the zones would be placed as part of the initial construction, as for the non investigation zones for the livestock expansions). That would have the advantage that the command could be used for player defined zones as well, as long as the command is made available on the bulletin board. The disadvantage is that the zones currently don't respect rotation/mirroring (#66213).
If this solution was used, it might be possible to detect the presence of farming zones on expansions and add the corresponding farming options to the bulletin board when at least one corresponding zone is present. A complication would be when zones cross expansion boundaries and when zones are present on unallocated expansions (should they be ignored by the bulletin board or should they default to appear for the base camp itself?).

PR #66219, opened 2023-06-15
parameterized livestock 2 palettes

PatrikLundell:
#### Summary

None

#### Purpose of change

Primary purpose: Decrease the size of the JSON and simplify the usage of multiple construction materials.
Bycatches:
- Allowing coexistence between items and constructions (stopping error messages from appearing)
- Changing the generation of no investigation zones to cover the whole expansion tile both to work around the bug that causes zones to not respect rotation/mirroring, but primarily to cover the pasture areas.

#### Describe the solution

Change of the JSON files to use the palette parameterization functionality, which is largely a copy/paste activity based on the pattern established by earlier changes of this type.

If it ain't broke don't fix it.

#### Testing

Created the whole expansion in concrete in the NE orientation and verified that it looked the way it should. Killed the game and repeated with a new orientation and construction material until all materials and orientations are covered (happen to be the same number of 8). The subsequent tests did not build the "common" parts (fences), as those are the same regardless of material, and aren't prerequisites for anything that has a parametrized material.
I didn't verify that the non investigation zone actually works (I assume the creators of the zone functionality did that), only that it covers the whole tile for a few "random" expansions.

#### Additional context

The common palette was moved to the palette file, and the unused "common" palette there was removed as a bit of random cleanup.
Also changed the names of the material palettes as it was both the easiest to make the change and the new names make more sense (the old ones were the result of search&replace gone mildly wrong, having "_2_2_" rather than "_2_" in them. It didn't matter as the same thing was applied to the places where the palettes were referenced, so it was consistent).


Maleclypse:
If the test fails again try rebasing to include the newest merges in.  The error you are encountering has been fixed in a recent merge I believe.

PR #66222, opened 2023-06-15
EOC Selector

bombasticSlacks:
#### Summary
None

#### Purpose of change
Allows the player to select an EOC option from a uilist

#### Describe the solution
Creates a UI list with the options and trigger the EOC based on it.

#### Testing
used the test selector EOCs which I've included.

#### Additional context
image


src/npctalk.cpp
```diff
@@ -4095,6 +4095,125 @@ void talk_effect_fun_t::set_run_eocs( const JsonObject &jo, const std::string_vi
...
+            } else {
+                display = true;
+            }
+
+            if( hide_failing && !display ) {
+                // skip hidden entries
+                continue;
+            }
+
+            std::string description = "";
```
bombasticSlacks, line 4188:
```suggestion
            std::string description;
```

PR #66224, opened 2023-06-15
Hands encumbrance no longer affects crafting speed, except when it's encumbrance from integrated armor or when gloves are a part of whole-body suit

Night-Pryanik:
#### Summary
Features "Hands encumbrance no longer affects crafting speed, except when it's encumbrance from integrated armor"

#### Purpose of change
Currently player has to manually take off gloves or similar hands-covering clothing/armor before crafting (and wear them once again after the crafting), which only increases keypresses. Also it differs from our other conventions, such as virtual walking in 6-tile radius during crafting, or that we don't need to take off masks and other mouth-covering clothing/armor during eating.

#### Describe the solution
Ported https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/18: 

- only calculate decrease of crafting speed due to hands encumbrance if it's the encumbrance from integrated armor. Otherwise, presume that player character automatically takes off gloves and the like before crafting and wears them back after crafting.

None.

#### Testing
Wore mittens, checked amount of time required to craft zweitimber (no increase), crafted it, noticed that time spent is correct.
Took off mittens, debug-get myself bark mutation, checked amount of time required to craft zweitimber (30% increase), crafted it, noticed that time spent is correct.

#### Additional context
None.

### DIFF
--- a/data/json/flags.json
+++ b/data/json/flags.json
@@ -2179,5 +2179,9 @@
     "id": "BANK_NOTE_STRAP_SHAPED",
     "type": "json_flag",
     "restriction": "Item must be shaped like a bank note strap"
+  },
+  {
+    "id": "WHOLE_BODY_SUIT",
+    "type": "json_flag"
   }
 ]
--- a/data/json/items/armor/power_armor.json
+++ b/data/json/items/armor/power_armor.json
@@ -90,7 +90,7 @@
     "warmth": 50,
     "material_thickness": 8,
     "environmental_protection": 6,
-    "flags": [ "WATERPROOF", "STURDY", "UNBREAKABLE_MELEE" ],
+    "flags": [ "WATERPROOF", "STURDY", "UNBREAKABLE_MELEE", "WHOLE_BODY_SUIT" ],
     "armor": [
       {
         "encumbrance": 40,
@@ -217,7 +217,7 @@
     "environmental_protection": 16,
     "ammo": "battery",
     "use_action": { "type": "transform", "msg": "The %s engages.", "target": "power_armor_basic_on", "active": true },
-    "flags": [ "WATERPROOF", "STURDY", "ELECTRIC_IMMUNE", "COMBAT_TOGGLEABLE", "UNBREAKABLE_MELEE" ],
+    "flags": [ "WATERPROOF", "STURDY", "ELECTRIC_IMMUNE", "COMBAT_TOGGLEABLE", "UNBREAKABLE_MELEE", "WHOLE_BODY_SUIT" ],
     "armor": [
       {
         "encumbrance": 50,
@@ -234,7 +234,16 @@
     "type": "TOOL_ARMOR",
     "name": { "str": "combat exoskeleton (on)", "str_pl": "combat exoskeletons (on)" },
     "description": "These were the second wave of military combat exoskeleton, and got a lot of media attention, with popular Navy commercials featuring them heavily.  It consists of a muscle-boosting exoskeleton frame with overlaid segmented alloy plating.  Despite advancements over the original bulky 'tank suits', the wearer still cannot easily fit through narrow spaces, or sit down comfortably (and it ruins upholstery).  There is an integrated chemical-resistant bodyglove that precludes wearing other clothing.  It is turned on and continually drawing power.  Use it to turn it off.",
-    "flags": [ "USE_UPS", "WATERPROOF", "STURDY", "ELECTRIC_IMMUNE", "TRADER_AVOID", "COMBAT_TOGGLEABLE", "UNBREAKABLE_MELEE" ],
+    "flags": [
+      "USE_UPS",
+      "WATERPROOF",
+      "STURDY",
+      "ELECTRIC_IMMUNE",
+      "TRADER_AVOID",
+      "COMBAT_TOGGLEABLE",
+      "UNBREAKABLE_MELEE",
+      "WHOLE_BODY_SUIT"
+    ],
     "power_draw": "4 kW",
     "revert_to": "power_armor_basic",
     "use_action": { "type": "transform", "menu_text": "Turn off", "msg": "The %s armor disengages.", "target": "power_armor_basic" },
--- a/data/json/items/armor/suits_clothes.json
+++ b/data/json/items/armor/suits_clothes.json
@@ -37,7 +37,7 @@
     "color": "blue",
     "warmth": 15,
     "material_thickness": 1,
-    "flags": [ "VARSIZE", "SKINTIGHT", "WATER_FRIENDLY" ],
+    "flags": [ "VARSIZE", "SKINTIGHT", "WATER_FRIENDLY", "WHOLE_BODY_SUIT" ],
     "armor": [
       {
         "coverage": 100,
@@ -573,7 +573,7 @@
     "color": "dark_gray",
     "warmth": 20,
     "material_thickness": 0.1,
-    "flags": [ "VARSIZE", "SKINTIGHT", "WATER_FRIENDLY" ],
+    "flags": [ "VARSIZE", "SKINTIGHT", "WATER_FRIENDLY", "WHOLE_BODY_SUIT" ],
     "armor": [
       {
         "coverage": 100,
--- a/data/json/items/armor/suits_protection.json
+++ b/data/json/items/armor/suits_protection.json
@@ -4701,7 +4701,7 @@
         "coverage": 100
       }
     ],
-    "extend": { "flags": [ "ELECTRIC_IMMUNE" ] }
+    "extend": { "flags": [ "ELECTRIC_IMMUNE", "WHOLE_BODY_SUIT" ] }
   },
   {
     "id": "xl_chainmail_suit_faraday",
@@ -4859,7 +4859,7 @@
     "longest_side": "30 cm",
     "material_thickness": 5,
     "environmental_protection": 20,
-    "flags": [ "VARSIZE", "WATERPROOF", "RAINPROOF", "GAS_PROOF", "STURDY", "OUTER" ],
+    "flags": [ "VARSIZE", "WATERPROOF", "RAINPROOF", "GAS_PROOF", "STURDY", "OUTER", "WHOLE_BODY_SUIT" ],
     "armor": [
       {
         "encumbrance": 50,
@@ -4963,7 +4963,17 @@
     "warmth": 40,
     "material_thickness": 2,
     "environmental_protection": 20,
-    "flags": [ "VARSIZE", "WATERPROOF", "RAINPROOF", "GAS_PROOF", "RAD_PROOF", "ELECTRIC_IMMUNE", "OUTER", "SOFT" ],
+    "flags": [
+      "VARSIZE",
+      "WATERPROOF",
+      "RAINPROOF",
+      "GAS_PROOF",
+      "RAD_PROOF",
+      "ELECTRIC_IMMUNE",
+      "OUTER",
+      "SOFT",
+      "WHOLE_BODY_SUIT"
+    ],
     "armor": [
       {
         "encumbrance": 37,
@@ -5159,7 +5169,7 @@
       "msg": "You slip out of the top of the suit and tie the sleeves around your waist.",
       "target": "robofac_enviro_suit_casual"
     },
-    "flags": [ "VARSIZE", "WATERPROOF", "RAINPROOF", "GAS_PROOF", "RAD_PROOF", "ELECTRIC_IMMUNE", "STURDY" ]
+    "flags": [ "VARSIZE", "WATERPROOF", "RAINPROOF", "GAS_PROOF", "RAD_PROOF", "ELECTRIC_IMMUNE", "STURDY", "WHOLE_BODY_SUIT" ]
   },
   {
     "id": "robofac_enviro_suit_casual",
--- a/data/json/items/armor/swimming.json
+++ b/data/json/items/armor/swimming.json
@@ -970,7 +970,7 @@
     "color": "light_red",
     "warmth": 5,
     "material_thickness": 3,
-    "flags": [ "VARSIZE", "STURDY", "WATER_FRIENDLY", "ELECTRIC_IMMUNE" ],
+    "flags": [ "VARSIZE", "STURDY", "WATER_FRIENDLY", "ELECTRIC_IMMUNE", "WHOLE_BODY_SUIT" ],
     "armor": [
       {
         "encumbrance": 30,
@@ -1002,7 +1002,7 @@
     "color": "light_red",
     "warmth": 5,
     "material_thickness": 3,
-    "flags": [ "VARSIZE", "STURDY", "WATER_FRIENDLY" ],
+    "flags": [ "VARSIZE", "STURDY", "WATER_FRIENDLY", "WHOLE_BODY_SUIT" ],
     "armor": [
       {
         "encumbrance": 30,
--- a/doc/JSON_FLAGS.md
+++ b/doc/JSON_FLAGS.md
@@ -387,6 +387,7 @@ Some armor flags, such as `WATCH` and `ALARMCLOCK` are compatible with other ite
 - ```WATCH``` Acts as a watch and allows the player to see actual time.
 - ```WATERPROOF``` Prevents the covered body-part(s) from getting wet in any circumstance.
 - ```WATER_FRIENDLY``` Prevents the item from making the body part count as unfriendly to water and thus reducing morale from being wet.
+- ```WHOLE_BODY_SUIT``` This item covers torso, hands, and legs (and sometimes head). Character can't free his hands or legs without taking off the suit.
 
 
 ## Bionics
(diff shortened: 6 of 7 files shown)


fairyarmadillo:
This change is putting convenience ahead of simulation in a way I don't think is in line with our design doc. There are plenty of times when the player is wearing gloves and wants to do a craft without taking them off, such as hurredly carving a cudgel in the cold in a dangerous area, or while wearing a full-body radiation suit in an irradiated area.

OtpalTS:
Is there a way to remove the crafting malus only when using a workbench? But really to take the gloves off takes about 200 moves. That isnt much, if you have arterial bleeding and need to craft a bandage then maybe but at that point it might be just safer to use the cotton sheet.

alp7292:
making game convinient to play is always better. i agree with author taking off gloves to craft is a mindless task which only wastes your time makes game boring. i imagine the days where you couldn't unload items with zoning spending minutes to unload items one by one.

fairyarmadillo:
> Is there a way to remove the crafting malus only when using a workbench? But really to take the gloves off takes about 200 moves. 

That's only true if you're healthy and not slowed down and you're wearing gloves.

Venera3:
This does not take non-basic hands into account, either in code nor in design. That needs to change.

Night-Pryanik:
> This does not take non-basic hands into account, either in code nor in design. That needs to change.

I'm not familiar with non-basic hands. I suppose right bow we can work only with basic hands and add a TODO for non-basic ones for future implementation.

Night-Pryanik:
> while wearing a full-body radiation suit in an irradiated area

> Furthermore it fails to account for the removal, storage, and re-wearing time of those items

We have many things we can do in full-body suit without taking it off: we can eat, drink, and consume meds without taking the helmet of the suit; we can wear and take off any "under-the-suit-layer" clothes; we can apply bandages or antiseptics or inject drugs directly to the limb.

Full-body suit is an edge case, and while ideally we should handle this case, I think it isn't a big deal if we don't simulate it "properly" right now since it's rather rare case.

Calavera20012:
Thank you, this is sorely needed. I've seen many complaints on how inconvenient the hands encumbrance is, personally, this is a really good change.

Qrox:
> Code limitations are not a justification for purposefully adding code to make the situation worse.

A better justification would be consistency. If someone wants to make it so that every action involving a bare part of the body requires taking off the clothing on that body part, that can always be added as a whole later (preferably automatically), it's not like it's the end of the world if we temporarily make the game experience more consistent.

Night-Pryanik:
> Code limitations are not a justification for purposefully adding code to make the situation worse.

How is game becoming a bit more convenient (a bit less grindy) equals to "make the situation worse"?

I admit, this is not a 100% ideal and flawless solution. But in my opinion even in this state it makes the game better, not worse.

Also, not requiring to take off clothes before making many things is not a "code limitations". Nothing prevents us from adding such a thing if someone wishes for it. Code absolutely allows this stuff. It's just no one bothered about this kind of stuff to this moment.

> This is not the way we should be going about things.

We went this way literally hundreds of times. We implemented stuff in non-ideal state and fixed and enhanced it in subsequent PRs. Had we choose some other way, we'd never have filthiness mechanics, pockets mechanics, portal storms and many many other things.

fairyarmadillo:
> > Code limitations are not a justification for purposefully adding code to make the situation worse.
> 
> A better justification would be consistency. If someone wants to make it so that every action involving a bare part of the body requires taking off the clothing on that body part, that can always be added as a whole later (preferably by automatically taking off and putting back the clothing involved), it's not like it's the end of the world if we temporarily make the game experience more consistent.

I can make it so that eating and drinking in a full body suit is impossible in an afternoon, and then when that gets merged, this justification will fall apart. All this is doing is actively creating more work for someone to come and undo later for an extremely marginal benefit.

And nobody's explained why it would be OK to get full crafting speed in -20c without taking hand damage.

Night-Pryanik:
> And nobody's explained why it would be OK to get full crafting speed in -20c without taking hand damage.

Crafting already has tons of implied, but not simulated mechanics. I suppose you don't think that character is standing in one place all the time when he's crafting e.g. naginata, and doesn't even sit down a single time for the whole dozens of hours?

This PR implies that character is smart enough to take off and wear gloves back during crafting _by necessity_. It's too cold? Ok, character wears gloves behind the scenes for some time to keep his hands warm, and takes them off when the hands are warm once again.

> extremely marginal benefit

Removing half a dozen of unneeded keypresses doesn't seem like a extremely marginal benefit to me.

fairyarmadillo:
> > And nobody's explained why it would be OK to get full crafting speed in -20c without taking hand damage.
> 
> Crafting already has tons of implied, but not simulated mechanics. I suppose you don't think that character is standing in one place all the time when he's crafting e.g. naginata, and doesn't even sit down a single time for the whole dozens of hours?

I'm not sure how this relates to taking off protective clothing for long periods of time and not getting frostbite or rad poisoning. If the person is constantly having to doff and don their gloves, that's going to slow down the process considerably compared to just doing it in a comfortable non-irradiated environment, and wouldn't you know it that's already handled by encumbrance speed penalties.

kilozombie:
It's worth remembering that Cataclysm: DDA is a video game and is already prone to some of the most rigorously annoying tedium imaginable in all of gaming. We put up with this because of its ability to simulate and amaze with its realism and potential in systems-oriented gameplay.

Having to take off your gloves to craft seems cool but it's a tradeoff. I like this version because having to take off your combat gloves to craft is annoying key presses, but sure--if you're in an extremely cold environment and taking off your gloves would make you freeze, that's a sacrifice too.

If people are insistent on the best of both worlds, a pop-up asking the player if they'd like to remove hand-encumbering gear temporarily would be nice. Maybe make it extra bonus contextual and only ask them if the outside temperature is low or if the gear encumbers more than just the hands (like a bodysuit).

andrei8l:
Maybe check if the local environment is safe-ish before skipping the penalty. This is a nice QoL for vanilla, but for Aftershock Exoplanet it's basically a cheat.

kilozombie:
> Maybe check if the local environment is safe-ish before skipping the penalty. This is a nice QoL for vanilla, but for Aftershock Exoplanet it's basically a cheat.

This would be nice too. Drop a message ingame without stopping gameplay that says "Taking off gloves automatically to craft faster..." if the penalty is being skipped, and don't print that message if the penalty is still being applied (in cold or radioactive environs).

fairyarmadillo:
> Maybe check if the local environment is safe-ish before skipping the penalty. This is a nice QoL for vanilla, but for Aftershock Exoplanet it's basically a cheat.

There's a problem with this solution because if it can check for radiation, then you can use it to check for radiation in an area without the proper tools to do so, and if you can't it's either going to expose you to radiation when you don't want it to, or magically let you get crafting speed despite the presence of rads.

anoobindisguise:
Yeah there are several places where this gives you an undue advantage. Here are some that have happened to me specifically:
- I used C-4 to enter a hazardous waste sarcophagus but I didn't realize there wasn't enough scaffolding to climb out, so I had to build a stepladder using bits of wood and nails scavenged from down there. The hazmat suit made it harder, but I couldn't exactly safely take it off.
- One of the consequences of artifact resonance is radiation, which can be entirely or mostly blocked by armor depending on its severity. Ditto the above.
- Crafting stuff in a frozen lab while in climate controlled bulky armor.
Etc. There are few cases where this really matters, but when it does, it REALLY matters.
Maybe a simpler solution would be to prompt the player to take off their gloves if feasible, and let the player say yes/no. Optionally cache the worn glove during the craft and then try to wear it on completion or interruption.

fairyarmadillo:
>   Maybe a simpler solution would be to prompt the player to take off their gloves if feasible, and let the player say yes/no. Optionally cache the worn glove during the craft and then try to wear it on completion or interruption.

This would be super annoying the 99% of the time you do want to take your gloves off, or for people who have only a small encumbrance penalty and don't care, and would defeat the stated purpose of the PR, which is to reduce keypresses.

Nebnis:
What about only allowing to do this near a workbench/table?

Qrox:
> I can make it so that eating and drinking in a full body suit is impossible in an afternoon, and then when that gets merged, this justification will fall apart.

Thats still only makes it half realistic and nowhere near consistent unless you address the other problems Night-Pryanik mentioned about bare body parts. If you plan to make all actions involving a bare body part require taking off the clothing on that body part, it's a great step towards what I described and I'm all for it.

> All this is doing is actively creating more work for someone to come and undo later for an extremely marginal benefit.

Even if it has marginal benefit, it's also trivial to undo, which seems fair enough to me.

> And nobody's explained why it would be OK to get full crafting speed in -20c without taking hand damage.

Tradeoff between verisimilitude and gameplay (and coding difficulty, and lack of people who are willing to take time to propose and implement a perfect solution) seem like an explanation to me, even if you don't agree with it.

> This would be super annoying the 99% of the time you do want to take your gloves off, or for people who have only a small encumbrance penalty and don't care, and would defeat the stated purpose of the PR, which is to reduce keypresses.

We could check the crafting penalty and only prompt if it makes crafting slow by more than a given amount, similar to how we prompt about being tired or attacking with items with super long attacking time.

kilozombie:
> I can make it so that eating and drinking in a full body suit is impossible in an afternoon, and then when that gets merged, this justification will fall apart. All this is doing is actively creating more work for someone to come and undo later for an extremely marginal benefit.
> 
> And nobody's explained why it would be OK to get full crafting speed in -20c without taking hand damage.

I think using this kind of intense language for a small, helpful, well-meaning PR isn't such a good idea. The average player already has a pretty dire view on the coders that make the game run and bikeshedding to this degree feels like a waste of everybody's time.

It's understandable to dislike the PR for making some niche scenarios more gamey than you're fond of, but it's not understandable to have this kind of tone. The PR is in good faith and the benefits are being weighed against the drawbacks.

anoobindisguise:
I do think this is a good change for convenience's sake and there's obviously a difference between wearing a full body radsuit sans gloves and no radsuit at all. It's just that the potential edge case exploits are a little concerning, but maybe they don't come into play often enough to be worried about those?

Maleclypse:
> > while wearing a full-body radiation suit in an irradiated area
> 
> > Furthermore it fails to account for the removal, storage, and re-wearing time of those items
> 
> We have many things we can do in full-body suit without taking it off: we can eat, drink, and consume meds without taking the helmet of the suit; we can wear and take off any "under-the-suit-layer" clothes; we can apply bandages or antiseptics or inject drugs directly to the limb.
> 
> Full-body suit is an edge case, and while ideally we should handle this case, I think it isn't a big deal if we don't simulate it "properly" right now since it's rather rare case.

I agree that a full-body suit is an edge case that can be addressed later.  Is it possible to add 2 seconds to the front and end of a craft to represent taking off and putting gloves back on if you were them?  That way it's not fully a handwave.  PC is crafting food in a kitchen of mostly cleared neighborhood and gloves are autoremoved, zombie bursts in the window and crafting is interrupted and now PC has gloves on.  If they had to take a second or so to be fully armored back up that'd make sense.  Would it be at least possible to prevent the auto removal in hostile environments like radiation or poison fields, ideally also extreme cold/heat for this PR or would that need to be a followup PR?

fairyarmadillo:
> I do think this is a good change for convenience's sake and there's obviously a difference between wearing a full body radsuit sans gloves and no radsuit at all. It's just that the potential edge case exploits are a little concerning, but maybe they don't come into play often enough to be worried about those?

I think they come up often enough that this PR shouldn't be merged unless it can satisfactorily cover at least a majority of the possible cases, and I don't see a way for that to happen. Night doesn't have a solution for nonstandard hands, cold and heat are a common issue even in the base game, and while radiation is a rarer concern, someone has already spoken up to share an example of a time when it would have mattered. And any solution that did cover all these bases would need to have a way to optionally disable it that wasn't obnoxious to deal with.

kilozombie:
Probably important to consider the reasons a survivor wears gloves in the first place. I've been coming at this issue from the perspective of someone playing mostly vanilla DDA and my playthroughs rarely go past summer, so I only give my survivors gloves as armor or, in niche circumstances, warmth or rad protection. But for colder playthroughs or especially in dangerous areas, that might be the _only_ reason to wear gloves.

It's only the armored gloves scenario where this PR helps, otherwise it is a bit gamey. I can't really say I ever take off my survivor's gloves in the first place to craft, but that's also because I tend to wear really light gloves to avoid encumbrance anyway. There's no real solution that gets to all the needs of a player, although I think making a full workbench take away the penalty will cut away most of the use cases. If you're crafting a sword in your big hulking workshop, and for some reason it's freezing cold, you might just take your gloves off for any precision tasks, not the long stretches of repetitive pounding on metal, etc.

fairyarmadillo:
> Is it possible to add 2 seconds to the front and end of a craft to represent taking off and putting gloves back on if you were them?

Is it the case that all gloves in all cases take two seconds to put on and remove? That doesn't seem right, shouldn't it be pulling the actual removal/wear time?

Termineitor244:
I think the suggestion of @Maleclypse is the best/easier to implement, but as comments say, it should probably take the taking off time of the clothing in question.

About the problem of where to automatically remove the gloves... There are some options too, you can make it an option in the game menu, something like:

"Automatic gloves removal": "Always", "Ask first", "Only in crafting zone", "Never"

Always removes them in every situation, ask first would be the option with a promt asking you if you really want to remove them before starting the craft (as suggested above in the comments), "Only in crafting zone" would always remove them but only when you have an special "zone" (the ones you use for sorting loot and the like) in range, meaning, only in your base (Where you shouldn't have to worry about radiation, acid, and hopefully cold), and never just ignores this new feature. 

If game options are not wanted then the ask first or only in crafting zone options could be turned into a simple utility mod, we always need more of these since there are too many content mods but to little utility ones.

Night-Pryanik:
Ok, I added a new `WHOLE_BODY_SUIT` flag to several pieces of clothing and armor: hazmat suit, power armor, faraday suit, swimming suit and some others. Game now checks for this flag and doesn't automatically neglect hands encumbrance (aka "automatically take off gloves") in this case.
Also added a todo comment in the code about non-basic hands.

rohanlean:
This is a simple change that greatly enhances the game and does not hinder any future improvements.  Assuming that the implementation is correct, it should be merged as is.  Those who are working on non-standard limbs can create a follow-up PR that handles those too.

To those who are upset that this violates realism: everything in this game is an imperfect simulation.  Some systems are simulated well, but most are not.  Players are willing to suspend their disbelief as long as the simulation produces results that are generally intuitive and never glaringly wrong.  The whole crafting process is highly abstracted.  I  find it more intuitive if the character intelligently manages their hand encumbrance, as abstracted by this PR.  The examples where this gives an “unfair” advantage are contrived, and the advantage gained seems marginal.  On the other hand, the current system regularly produces missed optimisations that the *character* would not have let happen, but that slipped through the player's attention.

There are many actual examples where poor simulations in the game produce unintuitive and tedious results.  For example it is optimal to have a bipod (handguard) on most firearms and switch to prone position for almost every shot, even at point blank range.  The only notable exception is if one plans to dodge an attack before firing.  This patch on the other hand does *not* produce any such degenerate results, as far as I can see.  If anything, it removes them.

Finally, there is a lot of low-hanging fruit for improvement, especially when it comes to making the game more challenging, entertaining, and believable past the first couple of in-game hours.  I do not understand why some are threatening to introduce even more meaningless tedium in response to this PR.  Why not spend time on a meaningful issue instead?

rohanlean:
I think the suit exception makes this patch worse.  The same rationale for the discount applies to suits.  A whole body suit is also what most characters currently tend to wear after a few hours.

kevingranade:
This is the wrong way to go about this on several levels.
If it's beneficial and safe to unencumber yourself to do an activity, we can streamline that, but just pretending it happened at no cost causes exploits.
The code bypasses the encumbrance check and hardcodes an outcome in a very inappropriate place.
This provides no feedback to the player about what is happening.

rohanlean:
@kevingranade: Apologies for commenting on a closed PR, but the underlying issue seems to bother many, and I am not yet convinced that your comment is actionable.  I think we all agree that your proposed solution would fix the problem; but how do you intend to check whether it is safe to disrobe, without querying the player or revealing information?  Do you even intend to tackle this yourself?

I am also unsure what the exploits are supposed to be.  The effect that hand encumbrance has on crafting times is an arbitrary approximation.  Base crafting times themselves are a somewhat arbitrary abstraction.  Yes, if a character is wearing some heavily encumbering suit and cannot safely remove it (a rare occurrence), then their crafting would realistically be slowed.  But why draw the line here in particular?  Should the player also have to regularly instruct their character to disrobe and relieve themselves or suffer discomfort / filthy clothing?  (I hope not!)  Compared to the proposed discount, there are already much worse (and often annoying) violations of believability in the game.  The current crafting system itself allows one to retrieve items from under a hostile turret.  And while some interactions can be improved, everything is a compromise, and player tedium has to be part of the consideration.

Zireael07:
Kevin clearly said the problem is that it bypassed wear/wield costs. Not checking whether it's safe.

PR #66239, opened 2023-06-16
more police armory adjustments

anoobindisguise:
#### Summary
none

#### Purpose of change
Few more things to clean up with that PR.

#### Describe the solution
- Spawn regular ammo and not silenced ammo in the regular fiveseven magazine group.
- Fix ammo spawning outside of lockers in one police station type.
- Fix sig mags spawning in glock groups.
- Make glock 9mm the most common of the 9mm pistol variants.
- Reduce ammo a bit overall and especially in prisons
- Spawn empty boxes more consistently

#### Testing
Open to feedback regarding ammo drops. The issue may be that police stations are too common, rather than that their armories are too stocked.


GuardianDll:
Should mention, that unhardcoded mx_roadblock spawns loot from police lockers at the vehicle's trunks

PR #66243, opened 2023-06-16
adjust bedding spawn weighting

anoobindisguise:
#### Summary
Balance "Less body pillows and shark plushies"

#### Purpose of change
There were way way way too many dakimakuras and animal plushies.

#### Describe the solution
Massively bump up the weight of regular stuff to balance out the body pillows etc.
The blahaj subreddit has about 80k users. Not all blahaj owners use that subreddit but there is a considerable overlap between people who are online enough to subscribe to a subreddit and people who would own a blahaj due to its prevalence in specific communities. Even if we assume some users have more than one blahaj and they all live in america,  I'd still only expect one in like 3000 people to have a blahaj in their bed.
I've seen variable figures regarding anime consumption. Some sources say about 2/3 of americans watch anime periodically while others cite 15-20% of americans being "anime fans". I can't find a specific figure for this but I'm going to assume that maybe 1/1000 anime fans have anime body pillows/dakimakuras so they might have a frequency of something like 1/5000.
Teddy bears are much more common than any other kind of stuffed animal. Apparently teddy bear ownership is relatively high among adults in the USA so I've left teddy bear frequency somewhat high, although I have adjusted the frequencies such that teddy bears spawn much more often than any other animal plushie.

Different frequencies. Teddy bears could definitely be higher. Dakimakura spawns might still be too high.

### DIFF
--- a/data/json/itemgroups/furniture.json
+++ b/data/json/itemgroups/furniture.json
@@ -32,16 +32,17 @@
     "id": "bed",
     "subtype": "distribution",
     "entries": [
-      { "item": "blanket", "prob": 25 },
-      { "item": "down_blanket", "prob": 13 },
-      { "item": "pillow", "prob": 30 },
-      { "item": "down_pillow", "prob": 15 },
-      { "item": "electric_blanket", "prob": 4 },
-      { "item": "quilt", "prob": 10 },
-      { "item": "quilt_patchwork", "prob": 8 },
-      { "item": "teddy_bear", "prob": 2 },
-      { "item": "shark_plushie", "prob": 2 },
-      { "item": "bodypillow", "prob": 1 }
+      { "item": "blanket", "prob": 2500 },
+      { "item": "down_blanket", "prob": 1300 },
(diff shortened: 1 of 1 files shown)


Procyonae:
I agree with the change in general but your nonsense statistics are pretty funny 😂

anoobindisguise:
> I agree with the change in general but your nonsense statistics are pretty funny 😂

Yeah they're very close to being just made up. I couldn't find any hard numbers unfortunately so it's an estimate based on what "feels about right"

Procyonae:
> Yeah they're very close to being just made up. I couldn't find any hard numbers unfortunately so it's an estimate based on what "feels about right"
I'm not expecting you to find actual statistics like just the way you worded it like [actual kind of eh statistic] and then lets just idk divide by 1000 ;D

Maleclypse:
Totally a different PR but I believe each bed should have a 90% chance to include a sheet and a pillow at minimum.  Probably best setup as a SUS collection and easiest to add in palettes.

Issue #52595, opened 2021-11-02
Shotgun speedloader chute not work


AlexNTFox:
### Describe the bug

Weapon addon and shotgun speedloader chute not work. Gun don't see not 6, nor 8. They just keeping used shells, menu not open

### Steps To Reproduce

Select Shotgun who support this addon
Unload gun
install speedloader chute
load chute (any type of 12 ammo)
Try to reload with chute

### Expected behavior

I expected it to work the same way as with revolver speedloader, but gun doesn't see these speedloaders

### Versions and configuration

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: a37d197 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Professions [package_bionic_professions],
    Aftershock [aftershock],
    Blaze Industries [blazeindustries],
    Magiclysm [magiclysm],
    C.R.I.T Expansion Mod [crt_expansion],
    Hair and Skin Color Cosmetics, UDP and MXotto+ version p10 [mxoudpcosmetics],
    Zetsukaze's Hair - 0.E EDIT [zets_hair_extensions_undead],
    Mythical Martial Arts [MMA],
    Rummaging [Rummaging],
    Crazy Cataclysm [crazy_cataclysm],
    Extra Mutated Scenarios [extra_mut_scens],
    Graphical Overmap [Graphical_Overmap],
    Whaleys Hair Mod (UDP Tweaks) [WHALEYHAIR_UDP],
    Graphical Overmap Vanilla Additions [Graphical_Overmap_SDG],
    Stats Through Skills [StatsThroughSkills],
    Stats Through Kills [stats_through_kills],
    SpeedyDex [speedydex],
    Fuji's Military Profession Pack [fuji_mpp]
]

ektoplazm:
can confirm on

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: 70623c6 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    No Fungal Growth [no_fungal_growth],
    Disable NPC Needs [no_npc_food],
    Bionic Professions [package_bionic_professions]
]

tinukedaya:
This one still doesn't work... I played around with the JSON, but it's above my level of knowledge to fix it. :D 

[...]

PR #66246, opened 2023-06-16
Allow shotguns to use speedloaders

voiddeal:
#### Summary
Bugfixes "Allow 6/8 round shotguns to use speedloaders"

#### Purpose of change
shotgun speedloaders don't work.
issue #52595 

#### Describe the solution

shotgun speedloadres are out of order for a while and I noticed some PRs tried to fix it but didn't.
as for myself I wasn't able to fix it too(!) so decide to simply just allow 6/8 round shotguns to use speedloaders without "chute".
and probably temporarily remove the item "chute" and its recipe as well? IDK
until the problem is fixed by someone capable.

leave it as is

#### Testing
no problem, can use speeloaders with or without the chute.

### DIFF
--- a/data/json/items/gun/shot.json
+++ b/data/json/items/gun/shot.json
@@ -223,7 +223,8 @@
     "pocket_data": [
       {
         "pocket_type": "MAGAZINE",
-        "rigid": true, "ammo_restriction": { "shot": 6 },
+        "rigid": true,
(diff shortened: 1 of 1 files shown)


randomtyper:
There are speedloaders that do not require a chute. So we can assume that ingame ones are of that variety.

https://www.youtube.com/watch?v=mEAmqEZdkPA

PR #66260, opened 2023-06-17
item: add auto whitelists to spawned containers

andrei8l:
#### Summary
None

#### Purpose of change
1. People hate it when items get picked up into a box of cookies.
2. People especially hate it when that container is collapsed by default so now both the new and old contents are hidden.

The first thing I do in ALL of my games is apply whitelists to my drug containers (painkillers, alcohol wipes, antibiotics, etc). This PR automates that and should mostly address the issues above too.

#### Describe the solution
When spawning containers with a single container pocket and a single content type, automatically add that type to the whitelist. This whitelist persists until the container is emptied (by any means), unless the player manually edits the pocket settings.

Add a whitelist/blacklist indicator (supercript plus) to the item name:
Screenshot from 2023-06-17 21-19-27
The mark for automated whitelist is dark gray, the player edited one is light gray. It's meant to be as unobtrusive as possible in both cases.

Default pocket settings defined in JSON (#65111): there may still be a use case for that, but it doesn't really address the issues above. Content creators would have to add specific containers for each content type in order  to accomplish the same thing.

#### Testing
Test units

Manual testing:
- Teleport around and verify that all single-content-type containers have whitelists (whether the contents have charges or not).
- Unload the containers by various means and verify that the whitelists get removed.
- Edit the pocket settings for a container in any way, then unload it and verify that the whitelist persists.

#### Additional context
Test unit revealed a bug with active items spawned by mapgen in bounds of the main map and e1441476e6260429cbd9e2816515da789dfc5f2d is the fix.

### DIFF
--- a/data/mods/TEST_DATA/mapgen-test.json
+++ b/data/mods/TEST_DATA/mapgen-test.json
@@ -305,5 +305,31 @@
     "type": "map_extra",
     "copy-from": "mx_crater",
     "generator": {  }
+  },
+  {
+    "id": "test_spawn_w_modifier",
+    "type": "item_group",
+    "subtype": "distribution",
+    "entries": [ { "item": "butter", "prob": 100, "charges-min": 1, "charges-max": 1, "container-item": "wrapper" } ]
+  },
+  {
+    "id": "test_spawn_custom_container",
+    "type": "item_group",
+    "subtype": "collection",
+    "container-item": "box_small",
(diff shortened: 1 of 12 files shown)


bombasticSlacks:
Nice clean solution 👍 I like the UI inclusion.

andrei8l:
Thanks for merging.

PR #66266, opened 2023-06-18
reduce police roadblock monster spawns

casswedson:
#### Summary
None

#### Purpose of change
police roadblocks can spawn 20+ zed cops swat zeds etc combined with the distraction offered by the nearby turrets and decent supply of guns and ammo around in corpses and the back of police cars makes for a super easy loot party, pretty much everyone in the road block is easy prey for someone with half a working brain

reduce the zed spawns by half, the special still gives a good amount of loot, the cars and corpses have good changes of setting up for a good gunfight, just not as much anymore

make every monster in the road block friendly towards each other, could be interesting

#### Additional context
vorm: you guys have too much stuff

### DIFF
--- a/data/json/mapgen/map_extras/roadblock.json
+++ b/data/json/mapgen/map_extras/roadblock.json
@@ -47,7 +47,7 @@
         { "vehicle": "policecar", "x": 18, "y": 10, "chance": 100, "status": 1 },
         { "vehicle": "policecar", "x": 16, "y": 17, "chance": 100, "status": 1 }
       ],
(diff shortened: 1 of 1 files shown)


GuardianDll:
Btw, isn't our turrets suppose to be inanimate objects, that do not trigger zeds attacks (ie use a monfaction, neutral to zeds?)

bombasticSlacks:
> Btw, isn't our turrets suppose to be inanimate objects, that do not trigger zeds attacks (ie use a monfaction, neutral to zeds?)

No search lights are in that category, things making a lot of noise and attacking IMO would still be a target for Zeds

PR #66267, opened 2023-06-18
reduce loot in cars a bit

casswedson:
#### Summary
None

#### Purpose of change
quite a few hours of play testing deemed the chances of first aid kits and survival kits in cars a little too much; I live on the run myself just hopping from place to place in my good old car, I just found too much stuff in other cars, so much I don't bother hitting specific places for med supplies

tone down the chances I added in my expansion of stuff spawning in cars


Ruminator24:
Yeah this is fair, i keep stumbling onto medkits in a bunch of cars, it is nice tho that it gives the jerrycan a better chance of spawning at least

PR #66272, opened 2023-06-18
Eoc selector patch

bombasticSlacks:
#### Summary
None

#### Purpose of change
Realized selectors might want to run with their own separate variables.

#### Describe the solution
Made variables an array, handle no entry, 1 entry, n entries.

#### Testing
Included my test EOCs


bombasticSlacks:
Closing this adding it to a larger chunk of EOCs stuff

PR #66274, opened 2023-06-18
[Bombasticperks] Tiny fix

Light-Wave:
#### Summary
None

#### Purpose of change
Bombasticperks seems to have had a bug that doesn't cause any faulty behaviour. Nerveless, this PR fixes it.

#### Describe the solution
changes from `u_exp` to the seemingly intentional `u_val('exp')`

#### Additional context
The bug didn't cause any issues since `EOC_on_kill_check_exp` will run and set `u_available_exp` to a correct value before it is checked the next time. But it confused me when I was trying to read the code.

### DIFF
--- a/data/mods/BombasticPerks/corefiles/core_eocs.json
+++ b/data/mods/BombasticPerks/corefiles/core_eocs.json
@@ -11,7 +11,7 @@
     "effect": [
       { "run_eocs": "EOC_give_perk_point" },
       { "math": [ "u_used_exp", "+=", "u_exp_to_perk" ] },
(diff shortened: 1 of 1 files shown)


SurFlurer:
AFAIK, these two spellings mean the same thing, especially in `math` context.

Light-Wave:
> AFAIK, these two spellings mean the same thing, especially in `math` context.

I _think_ u_exp would refer to a variable called exp scoped on the alpha talker, while u_val('exp') would actually return how much exp the talker has.

SurFlurer:
Thanks for correcting me.

bombasticSlacks:
Good find thank you!

PR #66285, opened 2023-06-18
Very minor pillow fort fix

### LINKED ISSUE #65616: Allow Pillow fort to be build on any type of floor not just on "floor".

I always want to build pillow fort for decoration purpose, but it need specific type of floor to be build on and it just only one. I don't want to build an entire floor just for pillow fort.

### Solution you would like.

Allow it to be build on every type of floors not just one.

_No response_

### Additional context

_No response_

Procyonae:
#### Summary
None

#### Purpose of change

Allows pillows on all floor types
Fixes #65616

#### Describe the solution

Just removes the unnecessary extra check for t_floor

#### Testing

Checked they can be placed on other terrains

#### Additional context

image
Pillow forts on other terrains


Maleclypse:
Most important fix in a generation

Issue #58069, opened 2022-06-01
We need a "hold your breath" action


GuardianDll:
### Is your feature request related to a problem? Please describe.

We know that situation, when you need to run through the dangerous place with some toxic gas or smoke, and we know what we can do to make this a bit easier - we can wet rag as sorta of filter mask, and we can, yknow, just hold the breath

### Solution you would like.

New action, in same place as excescises, that will prevent player to inhale bad stuff, in cost of huge increasing the stamina consumption. 

### Describe alternatives you have considered.

Thinking the player is dumb enough to breath such things

### Additional context

I know most of gases attacks not only lungs, but also eyes/skin, but at least its better than go to cloud and breath bad things as normal air. Maybe we can also add some gases that target not only lungs but also skin and eyes, yeah

fairyarmadillo:
I feel like this could get annoying to micromanage. Is there any reason not to automate it? I can't think of a situation IRL where you'd want to suck down a lungful of smoke instead of holding your breath. 

This makes sense though. There are probably some mutants that would reasonably be able to hold their breath better or worse, too. Lizards and plants come to mind. Frogs probably couldn't properly hold their breath because they partially breathe through their skin.

KittyTac:
> I feel like this could get annoying to micromanage. Is there any reason not to automate it? I can't think of a situation IRL where you'd want to suck down a lungful of smoke instead of holding your breath.
> 
> This makes sense though. There are probably some mutants that would reasonably be able to hold their breath better or worse, too. Lizards and plants come to mind. Frogs probably couldn't properly hold their breath because they partially breathe through their skin.

It would add a good reason to go Lizard, too! Lizard is generally considered lackluster and giving it "gas mask at home" would help.

Stone94:
You should be able to prone through thin smoke layers as well, smoke rises and one of the things I was told countless times if I were ever in a house fire was to crawl on the ground where the smoke was thinner.

PR #66291, opened 2023-06-18
Prompt player to hold his breath upon entering tiles with a smoke

Night-Pryanik:
#### Summary
Features "Prompt player to hold his breath upon entering tiles with a smoke"

#### Purpose of change
Holding breath should be enough to safely traverse tiles with a smoke. Inspired by https://github.com/CleverRaven/Cataclysm-DDA/issues/58069#issuecomment-1143333891.

#### Describe the solution
Ported https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/61 and https://github.com/AtomicFox556/Cataclysm-EOD/pull/97/commits/f63e2fe697b80781105a530aede43cb22b912c4d.
- Added a prompt upon entering tile with a smoke. If player answers yes, then player character automatically holds his breath until he quits tile with a smoke or his oxygen runs out
- While standing in that tile, oxygen doesn't recover
- After exit from that tile, oxygen starts to recover as usual
- If oxygen runs out while standing in that tile, player character automatically inhales deeply, resulting in lungs full of smoke
- The feature only works for smoke fields

None.

#### Testing
Debug-spawned a smoke vent to get a stable smoke field. Tried to enter the smoke, got a prompt. Answered yes, entered the tile. No inhaling the smoke since character holds his breath. Waited ~1 minute for oxygen to run out. Got a warning about low oxygen supply, then waited some more. Character inhaled deeply.

#### Additional context
изображение

изображение

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -260,6 +260,7 @@ static const efftype_id effect_ridden( "ridden" );
 static const efftype_id effect_riding( "riding" );
 static const efftype_id effect_sleep( "sleep" );
 static const efftype_id effect_slept_through_alarm( "slept_through_alarm" );
+static const efftype_id effect_smoke( "smoke" );
 static const efftype_id effect_stumbled_into_invisible( "stumbled_into_invisible" );
 static const efftype_id effect_stunned( "stunned" );
 static const efftype_id effect_tapeworm( "tapeworm" );
@@ -299,6 +300,7 @@ static const json_character_flag json_flag_ECTOTHERM( "ECTOTHERM" );
 static const json_character_flag json_flag_ENHANCED_VISION( "ENHANCED_VISION" );
 static const json_character_flag json_flag_EYE_MEMBRANE( "EYE_MEMBRANE" );
 static const json_character_flag json_flag_FEATHER_FALL( "FEATHER_FALL" );
+static const json_character_flag json_flag_GILLS( "GILLS" );
 static const json_character_flag json_flag_GRAB( "GRAB" );
 static const json_character_flag json_flag_HEAL_OVERRIDE( "HEAL_OVERRIDE" );
 static const json_character_flag json_flag_HEATSINK( "HEATSINK" );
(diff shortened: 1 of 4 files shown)


GuardianDll:
image
<https://discord.com/channels/598523535169945603/598535827169083403/1108861431576789083>

randomtyper:
> image https://discord.com/channels/598523535169945603/598535827169083403/1108861431576789083

I am pretty sure holding breath is being taken as granted. Inhaling a lungful of smoke like the character does in game is pretty stupid. 

https://www.pratt.edu/resources/emergency-response-guide-fire-safety/

If this change does not get through, then the impact of smoke should be decreased at least somewhat.

That said, I have my own concerns that this change will render the filter masks essentially useless unless an increase in filter time is applied (and that is hardcoded IIRC).

fairyarmadillo:
I like the idea of just making the first couple of seconds of smoke exposure not as impactful. That strikes a balance between sucking in a huge lungful immediately and potentially annoying UI prompts.

randomtyper:
> I like the idea of just making the first five or so seconds of smoke exposure not as impactful, maybe with a stamina check to simulate being unable to hold your breath when breathing heavily. That strikes a balance between sucking in a huge lungful immediately and potentially annoying UI prompts.

The UI prompt to enter smoke would still stay, though. Or no?

fairyarmadillo:
> The UI prompt to enter smoke would still stay, though. Or no?
> 
> Then there is issue of making an effect where first few moments of exposure are different in impact from the long duration.

Kevin seemed to indicate that a UI prompt is not going to happen, and I have to agree that a prompt would be the wrong way to go.

It wouldn't be hard to make the effect interact with an EOC that steps up severity after a few turns of exposure. This could also check stamina and if it was too low it would simply apply the full effect immediately, no extra input from the player required, just a small buffer if they're not tired.

bombasticSlacks:
Guardian grabbed the wrong part of that convo so I'll paraphrase the actual part that matters:

Smoke also gets in your eyes and causes irritation.

If we want to allow holding breath we need to first split it into multiple severities of smoke since you can't just stand upright in smoke from a burning building, hold your breath and keep acting normally.

If we did split smoke, the player should hold their breath automatically. *Why would you ever want to eat the smoke right away*, but at that point why are we tracking breath holding for campfires or whatever, it's just busy work, weak smoke clouds could easily just not cause irritation.

Qrox:
> Smoke also gets in your eyes and causes irritation.

That's a valid point, but it currently only mentions lung irritation in the smoked effect description/messages, so that should be changed before someone gets around to actually separate eye irritation from lung irritation.

https://github.com/CleverRaven/Cataclysm-DDA/blob/c3e2de83c58fa96b79136273728a8f52f1a636dc/data/json/effects.json#L1935-L1939

bombasticSlacks:
> That's a valid point, but it currently only mentions lung irritation in the smoked effect description/messages, so that should be changed before someone gets around to actually separate eye irritation from lung irritation.

Yeah, the message is definitely misleading and should also make it clear it's all forms of irritation, or as you say it could be split.

PR #66292, opened 2023-06-18
Carrier for items on ground is nullptr

Hirmuolio:
#### Summary
Infrastructure "Carrier for items on ground is nullptr"

#### Purpose of change

Active items in active item processing always had player as the item carrier if the item was on ground.
This added unnecessary complications to item processing code as you had to do extra check to see if the carrier actually carried the item.

#65114

#### Describe the solution

If item is on ground it has no carrier (nullptr).

Some of the item actions did not handle not having carrier well. Adjusted them to work well.
If an action would never not be activated by player I left them be. Things like bionic installing or eating meds. They should never be triggered in active item processing.

Item actions in iuse_actor.cpp used `Character &` but apparently "reference cannot be bound to dereferenced null pointer in well-defined C++ code". And if nobody carries the item it would be null. So I changed it to use `Character *`. About 90% of this pull request is hundreds of small changes to make that happen. If that part compiles it is probably fine I hope.

Added tests for `ACT_ON_RANGED_HIT`.

#### Additional context

`iuse_transform::use()` contains checks to see wether player can do the action. Most of those should probably be moved into `iuse_transform::can_use()`.  
Some other use actions contain branches that can never be reached because `can_use()` prevents the action from being done but then `use()` handles those situations where it can't be done again. 

Did you know that when a flaslight runs out of power it first turns into `flaslight_off` and after that it turns into `flashlight_off`. Well now you know.

Acid bomb is the only item in existence to use `ACT_ON_RANGED_HIT`. So 100% of the items that use that feature are covered by the automated tests.


src/iuse_actor.cpp
```diff
@@ -231,77 +231,77 @@ std::optional<int> iuse_transform::use( Character &p, item &it, bool t, const tr
 
     int result = 0;
 
-    const bool possess = p.has_item( it ) ||
-                         ( it.has_flag( flag_ALLOWS_REMOTE_USE ) && square_dist( p.pos(), pos ) <= 1 );
+    const bool possess = p->has_item( it ) ||
```
andrei8l, line 234: I would prefer a solution with less churn, something like
```c++
std::optional<int> iuse_transform::use( Character *c, item &it, bool t, const tripoint &pos ) const
    if( c == nullptr ) {
        debugmsg( "yo dawg that's not right" );
        return std::nullopt;
    }
    Character &p = *c;
```
then you won't have to change every instance of `p`. You can drop the nullptr check for all current functions too.

PatrikLundell:
Does the first sentence in the solution description have too many negations (not on ground, no carrier)? It seems from the description that items on the ground do not need any carriers, while those elsewhere require them. Basically that the first "not" should be removed.

src/iuse_actor.cpp
Hirmuolio: I don't see what you are doing here. Changing `iuse_transform::use( Character &c` to `iuse_transform::use( Character *c` is all of the churn and without that churn the check you are suggesting is not possible.

Any kind of null check I tried to do with `Character &` resulted in `error: reference cannot be bound to dereferenced null pointer in well-defined C++ code; pointer may be assumed to always convert to true [clang-diagnostic-undefined-bool-conversion,-warnings-as-errors]` from our code analyzers that run on Github.

PR #66294, opened 2023-06-18
Cockroach Bonanza

### LINKED ISSUE #66131: EAT_FOOD cannot eat food
### Describe the bug

Monsters with the special attack EAT_FOOD will never eat anything, regardless of whether they can see the player, how close the player is, or what the food item is.

I have extensively tested this and come up with a few observations:

```
                if( item.count_by_charges() ) {
                    here.use_charges( p, 0, item.type->get_id(), consumed );
                }
```

Removing this check results in the monsters successfully eating items, however it will cause them to eat whole stacks of things.

```
                if( item.count_by_charges() ) {
                    here.use_charges( p, 0, item.type->get_id(), consumed );
                    add_msg_if_player_sees( *z, _( "The %s nibbles on some food." ), z->name() );
                } else {
```

If this message is present, it will pop up periodically when the monster is near the food, but the food will not be eaten.

```
                if( item.count_by_charges() ) {
                    here.use_charges( p, 1, item.type->get_id(), consumed );
                }
```

Extending the range to 1 here allows them to successfully eat single charges of food, but searching github shows that this special attack was working with a range of 0 at least back in December 2020, so I'm not sure this is the right solution. In testing with the message present and the range set to 1, a cockroach trapped inside a car produced the "the cockroach nibbles on some food." message repeatedly as it tried and failed to eat a piece of food which was on the ground beneath the rear hatch.

### Attach save file

n/a

### Steps to reproduce

1) Make a character
2) Debug invisibility, deodorizer, silent walk
3) Spawn a food item, place it on the ground
4) Move some distance away from it and spawn some giant cockroaches around the food item
5) Observe that it does not get eaten

### Expected behavior

The food should be eaten.

### Screenshots

_No response_

- OS: Windows
- Game Version: 0.G-2240-g602e9a2a65-dirty [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

I'm guessing this is an indirect result of count_by_charges or use_charges getting messed with at some point.

worm-girl:
#### Summary
Features "Cockroaches flee light, hide under furniture, eat corpses, and breed faster when fed"

#### Purpose of change

There's currently not much to animal husbandry in Cataclysm, but the pieces are all there to make it work. This PR is a project to make cockroaches a bit more interesting and in doing lay the groundwork for better interactions with livestock. It also helps expand conflicting factional interests between roaches and zombies and creates a tug-of-war that clever players can try to exploit, though this can backfire hilariously as well.

fixes #66131 

#### Describe the solution

* Added the EAT_CARRION monattack. This causes monsters to gnaw on a corpse, dealing some damage to it.
* Gave monsters a stomach and the ability to fill it by eating.
- * This currently only affects cockroaches. Other animals that eat things currently use the old method.
- * Monsters can now have a STOMACH_SIZE, which is filled by 1 every time they gnaw a corpse or eat an item.
- * Monsters digest 1 food from their stomach every day (checked on load).
- * Monsters with full stomachs will not gnaw or eat anything.
- * Monsters with the EATS flag (currently only roaches) have a lower chance to reproduce if their stomach is empty.
- * Monsters never _need_ to eat, it just helps them if they do.
* Added the BRIGHT_LIGHT trigger, which can be used for fear, anger, or pacification. All roaches except the plague vector (the strongest one) are afraid of bright light.
* Added the SMALL_HIDER flag, which lets creatures hide under furniture with the SMALL_HIDE flag. This currently includes all roaches except the plague vector and pregnant cockroach, who is too fat to scurry under the fridge.
- * This flag can be used in the future for any animal that could reasonably hide under low furniture, like rats and stuff. It might also help wildlife seem a little more realistic if little animals can hide in bushes etc.
* Added the CORNERED_FIGHTER flag. Monsters with this flag will stop fleeing and attack if a hostile player or monster gets within two tiles of them. This only happens after they've had a turn or so to back off, so they can still flee from melee on open ground if the player does not pursue them, or if they are faster. This flag should solve a lot of problems that plague our current morale system. Once this PR gets merged it can be stuck on pretty much all large mammals, mi-go, and anything else that would reasonably realize when it was cornered.
* The giant cockroach, skittering plague, and plague vector were all the same size. I made the weaker two a bit smaller so that they produce less meat and chitin. This might make them a little harder to hit, but that comes with the player also being able to scare them off with a flashlight, so a small buff is probably fine.

To do before merge:
- [x] Make sure only corpses made of flesh, iflesh, veggy, etc. are edible.
- [x] Turn up the procreation rate on roaches to correct for the overall debuff.
- [x] Perform a proper ecological survey of how this plays out in the wild. Roaches should not generally be overpowering and eating all the zombies, but early game they should make a dent in them, and in doing so create a lot more roaches.
- [x] Make it so that butchery refuse is edible for scavengers but feces are not.
- [x] Give the SMALL_HIDE flag to undergrowth and tall grass.
- [x] Figure out if pulped corpses need to be inedible.

* ~I think we need a way for cornered monsters to realize there's nowhere to flee to and fight back, but coding that is beyond me.~ No it's not. I did it!
* It would also be great if monsters could eat food in non-sealed containers. Currently chip bags and sealed protein rations are safe, and as someone who has owned a dog, I'm not sure they should be.
* I think that at midnight the game should reload all monsters in the player's vicinity. It might be a good idea to redraw the map too, that way monsters can evolve, plants can grow, etc. without relying on the player leaving and returning.

#### Testing

image
Created two test chambers, with food in the bottom but not in the top, and observed them over a long period of time. The lower roaches produced more eggs than the upper roaches, but both groups did lay eggs.

Observed that SMALL_HIDERs can hide under low furniture, only becoming visible when the player is adjacent.

Observed that bright light was frightening roaches after a few turns.

image

Spawned a giant cockroach in this basement, which was lit by a flashlight. After a few turns, the cockroach began fleeing. I approached it, and it attacked me. I fled and it resumed fleeing. I spawned a zombie. The zombie attacked the fleeing cockroach, which turned around and fought back until it killed the zombie, whereupon it resumed fleeing.

Fleeing roaches will tend to hang out in their new hiding places, which was an unexpected but really nice surprise. They'll now jump out and ambush players who get too close. This makes roach-infested areas feel a lot creepier than they used to, and players will want to keep an eye out for roach poop and other signs that there may be creepy crawlies hidden around the place.


Venera3:
Nice!

ABSORB has a lot of JSON infrastructure already in place dealing with unhardcoded item filtering, and it's integrated to monster goals - I'd hook this into that system mostly. I'll give it a proper review soonish.

Karol1223:
As someone 100% C++ illiterate I cannot properly review this but I will ask some questions and possibly give suggestions. Though, due to my code illiteracy, please don't be angry at me if the suggestions are idiotic, already implemented and I didn't notice or impossible to implement.

> Monsters can now have a STOMACH_SIZE, which is filled by 1 every time they gnaw a corpse or eat an item.

Does this utilize volume or just numbers? Because it *sounds* like just numbers - it using volume would be better but also I reckon harder to implement. The problem with this here is one chomp of a hulk sounds like it'd fill them by as much as one chomp of a normal zombie, which isn't exactly the case.

> Added the SMALL_HIDER flag, which lets creatures hide under furniture with the SMALL_HIDE flag. This currently includes all roaches except the plague vector and pregnant cockroach, who is too fat to scurry under the fridge.

While I LOVE the implementation - there's no way in hell a 30l cockroach is hiding under a fridge. This should likely be only reserved for nymphs or somehow reworked into multiple flags. 

These are the main concerns I have with this. The rest seems great - having proper diet choices would be swell but it's not necessary for roaches so this could come down the line whenever we actually use them somewhere.

Bonus points if you manage to have corpses be eaten but leave behind whatever ``bone`` parts of the harvestlist is associated with the monster, but I reckon that's a lot of work for very little returns.

fairyarmadillo:
> As someone 100% C++ illiterate I cannot properly review this but I will ask some questions and possibly give suggestions. Though, due to my code illiteracy, please don't be angry at me if the suggestions are idiotic, already implemented and I didn't notice or impossible to implement.
> 
> > Monsters can now have a STOMACH_SIZE, which is filled by 1 every time they gnaw a corpse or eat an item.
> 
> Does this utilize volume or just numbers? Because it _sounds_ like just numbers - it using volume would be better but also I reckon harder to implement. The problem with this here is one chomp of a hulk sounds like it'd fill them by as much as one chomp of a normal zombie, which isn't exactly the case.
> 
> > Added the SMALL_HIDER flag, which lets creatures hide under furniture with the SMALL_HIDE flag. This currently includes all roaches except the plague vector and pregnant cockroach, who is too fat to scurry under the fridge.
> 
> While I LOVE the implementation - there's no way in hell a 30l cockroach is hiding under a fridge. This should likely be only reserved for nymphs or somehow reworked into multiple flags.
> 
> These are the main concerns I have with this. The rest seems great - having proper diet choices would be swell but it's not necessary for roaches so this could come down the line whenever we actually use them somewhere.
> 
> Bonus points if you manage to have corpses be eaten but leave behind whatever `bone` parts of the harvestlist is associated with the monster, but I reckon that's a lot of work for very little returns.

Amount_eaten is a simple integer that goes up by one when the animal eats. Volume wouldn't be strictly appropriate either as corpses are not the only thing we are going to be measuring with this - it's going to be allowing horses to eat hay (extremely low calorie to volume ratio) and fruit (high calorie to volume ratio).

There's no way to pull the calorie count from a corpse, and something like a zombie has very different calories from a triffid.

This needs to be as simple as possible because the player is not going to want to micromanage domestic animal feed, and neutral/enemy monsters are only on-screen for a short amount of time. Attempting to perfectly model the amount of calories they're getting would result in all of them being underfed because they have no ability to intelligently seek out food and their lives aren't simulated while the player is gone. I believe a simplified method is the best solution for a couple of reasons:

1) It's mostly invisible to the player.
2) Since animals can't intelligently find food and are limited to only acting when they're in the reality bubble, we must assume that they are taking care of themselves to some extent even when the player isn't around, otherwise they should starve to death.
3) The result of the current system is that when the player observes roaches eating, regardless of what they're eating, the roaches in that area will eventually lay more eggs. There is not a directly observable relationship between one and the other, so players are not going to be able to notice that the roaches are not achieving their DRVs.
4) I'm pretty sure that large corpses have more HP, so a hulk corpse does provide proportionally more food than a child zombie corpse.

The bone issue is something that rotting should be doing, but currently doesn't. It's out of scope at this time.

As for roaches squeezing into places, it's something they do in real life: https://www.nationalgeographic.com/science/article/watch-amazing-video-reveals-why-roaches-are-so-hard-to-squish#:~:text=Not%20only%20can%20roaches%20fit,the%20National%20Academy%20of%20Sciences.

Cockroaches can flatten their exoskeleton to an unbelievable degree and fit through incredibly tiny gaps, When doing this, their legs splay out almost completely horizontally, allowing them to fit through or hide under pretty much anything comfortably. 30 liters is not terrifically large when you pancake it like that. I've seen a four foot nile monitor hide under a fridge, and while they're not quite as bulky, they don't have special adaptations for flattening themselves into small gaps.

Karol1223:
> Amount_eaten is a simple integer that goes up by one when the animal eats. Volume wouldn't be strictly appropriate either as corpses are not the only thing we are going to be measuring with this - it's going to be allowing horses to eat hay (extremely low calorie to volume ratio) and fruit (high calorie to volume ratio).

I both vastly agree and disagree with this.

Stomach content - while not necessary here - is going to be VERY important for player-farmed animals. Why? Because, for example, horses have terribly low stomach volume compared to their size and dietary needs ERGO horses need higher quality feed than for example cows or other ruminants which can engorge themselves as needed.

The specifics of diets of various animals are almost certainly too much - but stomach contents are pretty important if we want realistic animal husbandry. I'll just assume that this is out of the scope of this PR, though.

fairyarmadillo:
> Stomach content - while not necessary here - is going to be VERY important for player-farmed animals. Why? Because, for example, horses have terribly low stomach volume compared to their size and dietary needs ERGO horses need higher quality feed than for example cows or other ruminants which can engorge themselves as needed.
> 
> The specifics of diets of various animals are almost certainly too much - but stomach contents are pretty important if we want realistic animal husbandry. I'll just assume that this is out of the scope of this PR, though.

The purpose of this PR is to lay the groundwork for animal husbandry, so I think we ought to get it right the first time. My primary goal for animal husbandry is to not leash the player to an annoying set of daily chores, and for that reason I think a simpler system is better. I also think putting calorie costs on animal fullness will unduly burden the extremely simple monster AI and the result would be either that they never hit their targets, or that the targets had to be set so low that they might as well just be the current integer system.

Karol1223:
> My primary goal for animal husbandry is to not leash the player to an annoying set of daily chores, and for that reason I think a simpler system is better. I also think putting calorie costs on animal fullness will unduly burden the extremely simple monster AI and the result would be either that they never hit their targets, or that the targets had to be set so low that they might as well just be the current integer system.

That honestly sounds like a mediocre at best bandaid fix to me, but like I said I'm not experienced in C++ at all so I'll trust your judgment with what's better.

fairyarmadillo:
> That honestly sounds like a mediocre at best bandaid fix to me, but like I said I'm not experienced in C++ at all so I'll trust your judgment with what's better.

I do not think that forcing the player to micro their animals' daily calorie requirements would go over very well with anyone except people who do not want ranching to be a thing at all.

Karol1223:
> I do not think that forcing the player to micro their animals' daily calorie requirements would go over very well with anyone except people who do not want ranching to be a thing at all.

It all boils down to how much realism are we willing to sacrifice in the name of player comfort. I'm of the opinion that animal husbandry should be difficult and realistic. If it means I have to go out of my way to have my animals be fed specifically how they should be - I am all in. Is it micromanagement to make sure your horse gets high caloric density food? Yes. But you should be considering that regardless if you're using that horse as a mount since it'd obviously burn more calories.

> Enjoyment has nothing to do with calories and the enjoyment values on food are assuming a human palate. Cockroaches and cows like different foods from humans and each other.

Hence me saying we should likely have it different per animal. I gave a few examples, though I believe I worded it wrong. Rather than defining it in the material json it should probably be defined as a diet somewhere that can be connected to monster's JSON later. So we could have a ``carnivore`` diet that'd prefer flesh>insect flesh>tainted flesh for example, and have some other comestible materials blacklisted. This is far more work, though.

fairyarmadillo:
> It all boils down to how much realism are we willing to sacrifice in the name of player comfort. I'm of the opinion that animal husbandry should be difficult and realistic. If it means I have to go out of my way to have my animals be fed specifically how they should be - I am all in. Is it micromanagement to make sure your horse gets high caloric density food? Yes. But you should be considering that regardless if you're using that horse as a mount since it'd obviously burn more calories.
> 
> Hence me saying we should likely have it different per animal. I gave a few examples, though I believe I worded it wrong. Rather than defining it in the material json it should probably be defined as a diet somewhere that can be connected to monster's JSON later. So we could have a `carnivore` diet that'd prefer flesh>insect flesh>tainted flesh for example, and have some other comestible materials blacklisted. This is far more work, though.

If this was merely a matter of player comfort I would agree with you, but I simply do not think there is currently anything approaching a good way to give animals the same level of detailed needs as the player. We can't even properly do it with NPCs yet and they have all the code that should be required.

I don't think distinct diet entries for individual monsters is any good either, as it would rapidly become so granular as to create an undue burden for future contributors. I also don't think we have anything like the capability for monsters to assess all the food in an area, choose their favorite out of the bunch, and go eat that. In any case it would be beyond my abilities.

Karol1223:
> I also don't think we have anything like the capability for monsters to assess all the food in an area, choose their favorite out of the bunch, and go eat that. In any case it would be beyond my abilities.

Yeah, hence me saying 'for future reference'. I don't think this is possible now - and even if it was, likely still out of the scope of this PR.

> We can't even properly do it with NPCs yet and they have working AI and all the code that should be required.

THAT is a valid concern. The main problem here is that most animals would need custom caloric requirements and that's a mess in and of itself. Though I suppose it'd theoretically be possible to have a default one that gets applied to all animals and then species-specific ones for domestic animals that'd get the full scope of this. Regardless - your point about not being able to simulate it in NPCs is valid. Let's hope it doesn't stay valid forever, though. It'd be nice to finally see food become somewhat of a bigger concern than it is...

fairyarmadillo:
> It'd be nice to finally see food become somewhat of a bigger concern than it is...

That is ultimately the point of the PRs I'm working on. If I can get something across the finish line that works so that cows aren't free 24/7 milk factories that need zero care, I'll be happy. If someone else is able to refine that later, all the better.

anoobindisguise:
Even if it's pretty trivial to keep cows fed by letting them wander around an enclosed field and letting them munch on grass it's still a dramatic improvement over the "underground cow leashed to a 1x1 pen". At least that way the cows sre exposed to danger

anoobindisguise:
> Nice!
> 
> ABSORB has a lot of JSON infrastructure already in place dealing with unhardcoded item filtering, and it's integrated to monster goals - I'd hook this into that system mostly. I'll give it a proper review soonish.

I'm not sure if this function will be useful, since Shoggoths (the primary user of that attack) rely on it being hardcoded in some specific ways (the cockroach will eat a stack of 1000 food one at a time, or damage a corpse once, while a shoggoth will instantly eat the whole thing). Using JSON filtering to detect what materials a thing will eat will probably be helpful though (like maybe a cow won't eat FLESH but will eat VEGGY)

fairyarmadillo:
I'm happy with this now, at least as a foundation for future PRs. Roaches remain a minor presence in towns and are not outcompeting zombies. Their reproduction does not seem unduly impacted by the changes, but the player should notice them getting into food more and occasionally pulping corpses. The biggest changes the player will notice are that they'll be out in the daylight less often and early game they'll cause a little chaos as they run around looking for shelter and alerting zombies, similar to the panicked people.

If we want to make roaches have a bigger impact, the easiest method would be to turn up the size of their stomachs.

Once this gets merged we can look at adding EAT_CARRION to other animals (crows, flies, etc).

@Venera3 requesting that you take another look at this. As @anoobindisguise said, I don't think the shoggoth ABSORB is going to work here as it operates in a completely orthogonal matter to the existing EAT_FOOD ability and the new EATCARRION one.

Venera3:
Please document the flags + attack behavior, otherwise the munching looks good to go - I'd be happy to see some size-dependent corpse damaging and food filter in the future but this is a completely viable addition on its own.

Note that the bright light trigger will still make them flee from their enemy they can see, disregarding the local light level. I'd also add a lower limit to the morale impact, -15 per tick is a hefty accumulation especially wrt getting to corner them with a flashlight and whacking away. The former is less of a problem, but the latter makes it unworkable for e.g. threatening mobs angered by light that chill when out of it.

fairyarmadillo:
Oh, good catch. I intended for fear to be generated when the roach was standing in light, not when its enemies were. I think I can just do like this to fix it?

```
    process_trigger( mon_trigger::BRIGHT_LIGHT {
        int ret = 0;
        const tripoint lightcheck = here;
        static const int dim_light = round( .75 * default_daylight_level() );
        int light = round( get_map().ambient_light_at( lightcheck ) );
        if( light >= ( dim_light ) ) {
            ret += 5;
        }
        return ret;
    } );
```

I'll try and figure that out once the game is working again.

Generally I think we need a way for monsters to realize there's nowhere to run to so they can turn and fight when cornered. I am not sure if it would be possible to do that simply by making proximity and hurting anger them enough to override their fear, but I'll mess around with it.

Switching this over to draft mode until I have that sorted.

fairyarmadillo:
@Venera3 I changed the light check so the roach now looks in its own tile instead of anybody else's. I also made a new flag that will fix the "chase the roach into a corner with a flashlight and beat it to death while it refuses to fight back" thing and the results are really entertaining so far. I've updated the PR text to describe what I've done, but essentially creatures with the flag who are fleeing will stop fleeing and fight back if the player or a hostile monster gets within spear range.

I left the fear value very high on the bright light check to simulate the "turn on the light and all the roaches run away" thing that famously happens IRL. But with this new flag the player can only keep them at bay this way if they don't get near them. Since roaches can now hide under furniture, this means they'll jump out and bite you as you walk by certain tiles, and they'll stay and fight if you don't give them some space. The big roaches don't care about light.

As soon as this gets merged I want to slap it on bears, boars, moose, and mi-go so they stop turning into helpless weenies as soon as the player chases them into a corner.

fairyarmadillo:
Whew. Apparently someone went and changed how monster flags worked since the time I started this PR, but it's all caught up and working now.

Maleclypse:
There are merge conflicts and once they are resolved and test has run again please ping me.

PR #66302, opened 2023-06-19
No deduplication on rebuilding vehicle active items cache

SurFlurer:
#### Summary
none

#### Purpose of change

In [this senario](https://github.com/CleverRaven/Cataclysm-DDA/pull/60885#issuecomment-1596180470), I found the game almost unplayable, because loading the save, starting or stopping the vehicle's engine all take minutes irl.

The reason is, in `active_item_cache::add()`:

https://github.com/CleverRaven/Cataclysm-DDA/blob/b43755d5e75e5041f8f7a26a3d07f563bc64b51f/src/active_item_cache.cpp#L66-L73

This part is expensive, especially when refreshing active item cache:

https://github.com/CleverRaven/Cataclysm-DDA/blob/b43755d5e75e5041f8f7a26a3d07f563bc64b51f/src/vehicle.cpp#L5794-L5798

Which means if there are 30000 comestibles to add( that's not impossible once #60885 gets merged ), the code needs to iterate over 30000 items for 30000 times.

#### Describe the solution

Add a param to `active_item_cache::add()` to skip checking duplicate active item entries, and only use that when fully rebuilding the cache.

Add a function for clearing the active items cache, which is needed to get rid of deduplication.

Separate the active item cache between vparts, instead of storing all things together. This might also be helpful, but doesn't provide as much improvement as this one ( the time complexity is still O(n^2) for rebuilding the cache, although `n` is smaller ).

#### Testing
The game is playable again in [this senario](https://github.com/CleverRaven/Cataclysm-DDA/pull/60885#issuecomment-1596180470).

### DIFF
--- a/src/active_item_cache.cpp
+++ b/src/active_item_cache.cpp
@@ -49,27 +49,34 @@ void active_item_cache::remove( const item *it )
     }
 }
 
+void active_item_cache::clear()
+{
+    active_items.clear();
+}
+
 bool active_item_cache::add( item &it, point location, item *parent,
-                             std::vector<item_pocket const *> const &pocket_chain )
+                             std::vector<item_pocket const *> const &pocket_chain, bool deduplicate )
 {
     std::vector<item_pocket const *> pockets = pocket_chain;
     bool ret = false;
     for( item_pocket *pk : it.get_all_standard_pockets() ) {
         pockets.emplace_back( pk );
         for( item *pkit : pk->all_items_top() ) {
-            ret |= add( *pkit, location, &it, pockets );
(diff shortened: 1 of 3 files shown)


andrei8l:
Doesn't this mess up the scheduling? If you clear and rebuild, only the items at the front of the list will ever get processed.

https://github.com/CleverRaven/Cataclysm-DDA/blob/8846e84f1d71438582caaa4e728fabb3fb791d79/src/active_item_cache.cpp#L129-L131

SurFlurer:
Thanks for reminding me, I indeed forgot to check about that. I'll close this then.

PR #66310, opened 2023-06-19
Use unordered_set to deduplicate active items

SurFlurer:
#### Summary
none

#### Purpose of change

Same as #66302 . Copy-pasted below.

In [this senario](https://github.com/CleverRaven/Cataclysm-DDA/pull/60885#issuecomment-1596180470), I found the game almost unplayable, because loading the save, starting or stopping the vehicle's engine all take minutes irl.

The reason is, in `active_item_cache::add()`:

https://github.com/CleverRaven/Cataclysm-DDA/blob/b43755d5e75e5041f8f7a26a3d07f563bc64b51f/src/active_item_cache.cpp#L66-L73

This part is expensive, especially when refreshing active item cache:

https://github.com/CleverRaven/Cataclysm-DDA/blob/b43755d5e75e5041f8f7a26a3d07f563bc64b51f/src/vehicle.cpp#L5794-L5798

Which means if there are 30000 comestibles to add( that's not impossible once #60885 gets merged ), the code needs to iterate over 30000 items for 30000 times.

#### Describe the solution

Use an `unordered_set<safe_reference>` to check if the item to be cached has already been cached, instead of iterating over the whole list to search for this item.

Add operator==() and a hash function to `safe_reference` class using [andrei's patch](https://github.com/CleverRaven/Cataclysm-DDA/pull/66310#issuecomment-1598447220), to be able to compare `safe_reference` and use that as index, to gain the ability to check for deleted items in the index.

Remove unused `remove()` function of `active_item_cache` class.

Separate the active item cache between vparts, instead of storing all things together. This might also be helpful, but doesn't provide as much improvement as this one ( the time complexity is still O(n^2) for rebuilding the cache, although `n` is smaller ).

#### Testing
The game is playable again in [this senario](https://github.com/CleverRaven/Cataclysm-DDA/pull/60885#issuecomment-1596180470).


andrei8l:
> That unordered_set is only for this purpose, and those pointers are not actually accessed. So I think it's safe to store bare pointers.

I'm not sure this is correct. What happens if an item gets deleted between `get_for_processing()` and `add()`, and a new one gets allocated at the same memory address? Won't it get falsely rejected as a duplicate?

`safe_reference<item>` might work. I haven't tried to fix `remove()` because it's dead code and I'm not completely confident this is correct either:

<details>
<summary>safe_reference patch</summary>

```diff
diff --git a/src/active_item_cache.cpp b/src/active_item_cache.cpp
index 74f387a605..540f4c7791 100644
--- a/src/active_item_cache.cpp
+++ b/src/active_item_cache.cpp
@@ -9,7 +9,7 @@
 namespace
 {
 
-void _remove_if( std::unordered_set<const item *> &active_index,
+void _remove_if( std::unordered_set<safe_reference<item>> &active_index,
                  std::list<item_reference> &active_items, const item *it )
 {
     active_index.erase( it );
@@ -68,16 +68,16 @@ bool active_item_cache::add( item &it, point location, item *parent,
     if( speed == item::NO_PROCESSING ) {
         return ret;
     }
-    std::unordered_set<const item *> &target_index = active_items_index[speed];
+    std::unordered_set<safe_reference<item>> &target_index = active_items_index[speed];
     std::list<item_reference> &target_list = active_items[speed];
     if( target_index.empty() ) {
         // If the index has been cleared, rebuild it first.
         for( item_reference &iter : target_list ) {
-            target_index.emplace( iter.item_ref.get() );
+            target_index.emplace( iter.item_ref );
         }
     }
     // If the item is already in the cache for some reason, don't add a second reference
-    if( target_index.find( &it ) != target_index.end() ) {
+    if( target_index.find( it.get_safe_reference() ) != target_index.end() ) {
         return true;
     }
     item_reference ref{ location, it.get_safe_reference(), parent, pocket_chain };
@@ -88,7 +88,7 @@ bool active_item_cache::add( item &it, point location, item *parent,
         special_items[special_item_type::explosive].emplace_back( ref );
     }
     target_list.emplace_back( std::move( ref ) );
-    target_index.emplace( &it );
+    target_index.emplace( it.get_safe_reference() );
     return true;
 }
 
diff --git a/src/active_item_cache.h b/src/active_item_cache.h
index aef733f3da..efca746ff6 100644
--- a/src/active_item_cache.h
+++ b/src/active_item_cache.h
@@ -46,7 +46,7 @@ class active_item_cache
     private:
         std::unordered_map<int, std::list<item_reference>> active_items;
         std::unordered_map<special_item_type, std::list<item_reference>> special_items;
-        std::unordered_map<int, std::unordered_set<const item *>> active_items_index;
+        std::unordered_map<int, std::unordered_set<safe_reference<item>>> active_items_index;
 
     public:
         /**
diff --git a/src/safe_reference.h b/src/safe_reference.h
index cdb1488ea1..a0a7294c75 100644
--- a/src/safe_reference.h
+++ b/src/safe_reference.h
@@ -53,6 +53,21 @@ class safe_reference
         std::weak_ptr<T> impl;
 };
 
+template<typename T>
+constexpr bool operator==( safe_reference<T> const&lhs, safe_reference<T> const &rhs )
+{
+    return lhs && rhs && lhs.get() == rhs.get();
+}
+
+template<typename T>
+struct std::hash<safe_reference<T>>
+{
+    std::size_t operator()(safe_reference<T> const& s) const noexcept
+    {
+        return std::hash<T *>{}( s.get() );
+    }
+};
+
 class safe_reference_anchor
 {
     public:
```

</details>

So in the scenario above, `operator==()` should return false even if the dead and new items have the same address.

SurFlurer:
Thanks for your patch! Now that `remove()` is dead code, is it fine to just get rid of it?

SurFlurer:
Clang is complaining about the modification of std namespace. I guess fully specializing it ( making a `struct std::hash<safe_reference<item>>` ) and move it to `active_item_cache.h` will satisfy clang? 

@andrei8l I'm looking forward to hear your valuable opinions.

andrei8l:
Oh right. That's a bug in clang-tidy since specializations of `std::hash` are allowed in spec. You can add `// NOLINTNEXTLINE(cert-dcl58-cpp)` like elsewhere in code. There's another example of a fully specialized `std::hash` in `active_item_cache.h` so your solution probably works too (I can't trigger this error locally to verify).

src/safe_reference.h
```diff
@@ -53,6 +53,12 @@ class safe_reference
         std::weak_ptr<T> impl;
 };
 
+template<typename T>
+constexpr bool operator==( safe_reference<T> const &lhs, safe_reference<T> const &rhs )
+{
+    return lhs && rhs && lhs.get() == rhs.get();
```
BrettDong, line 59: Two safe references pointing to different items can compare equal after both items have been destroyed.

src/active_item_cache.h
```diff
@@ -38,22 +39,23 @@ struct hash<special_item_type> {
         return static_cast<size_t>( k );
     }
 };
+
+template<>
+struct hash<safe_reference<item>> {
+    std::size_t operator()( safe_reference<item > const &s ) const noexcept {
+        return hash<item * > {}( s.get() );
```
BrettDong, line 46: The same safe reference object can generate different hash values before and after the item has been destroyed.

SurFlurer:
I'm trying to implement this cache using an `unordered_map`. Current diff:

```diff
diff --git a/src/active_item_cache.cpp b/src/active_item_cache.cpp
index aeaa623f04..e9a9a146dc 100644
--- a/src/active_item_cache.cpp
+++ b/src/active_item_cache.cpp
@@ -60,15 +60,22 @@ bool active_item_cache::add( item &it, point location, item *parent,
             ret |= add( *pkit, location, &it, pockets );
         }
     }
-    if( it.processing_speed() == item::NO_PROCESSING ) {
+    int speed = it.processing_speed();
+    if (speed == item::NO_PROCESSING) {
         return ret;
     }
-    std::list<item_reference> &target_list = active_items[it.processing_speed()];
+    std::unordered_map<item*, safe_reference<item>>& target_index = active_items_index[speed];
+    std::list<item_reference>& target_list = active_items[speed];
+    if (target_index.empty()) {
+        // If the index has been cleared, rebuild it first.
+        for (item_reference& iter : target_list) {
+            target_index.emplace(iter.item_ref.get(),iter.item_ref);
+        }
+    }
     // If the item is already in the cache for some reason, don't add a second reference
-    if( std::find_if( target_list.begin(),
-    target_list.end(), [&it]( const item_reference & active_item_ref ) {
-    return &it == active_item_ref.item_ref.get();
-    } ) != target_list.end() ) {
+    auto iter = target_index.find(&it);
+    if (iter != target_index.end()) {
+        if(iter->second && iter->second.get() == &it)
         return true;
     }
     item_reference ref{ location, it.get_safe_reference(), parent, pocket_chain };
@@ -79,6 +86,7 @@ bool active_item_cache::add( item &it, point location, item *parent,
         special_items[special_item_type::explosive].emplace_back( ref );
     }
     target_list.emplace_back( std::move( ref ) );
+    target_index.emplace(&it,it.get_safe_reference());
     return true;
 }
 
@@ -98,6 +106,7 @@ std::vector<item_reference> active_item_cache::get()
                 all_cached_items.emplace_back( *it );
                 ++it;
             } else {
+                active_items_index[kv.first].clear();
                 it = kv.second.erase( it );
             }
         }
@@ -123,6 +132,7 @@ std::vector<item_reference> active_item_cache::get_for_processing()
                 ++it;
             } else {
                 // The item has been destroyed, so remove the reference from the cache
+                active_items_index[kv.first].clear();
                 it = kv.second.erase( it );
             }
         }
diff --git a/src/active_item_cache.h b/src/active_item_cache.h
index 5401b05eba..858100a6cb 100644
--- a/src/active_item_cache.h
+++ b/src/active_item_cache.h
@@ -30,6 +30,12 @@ enum class special_item_type : int {
     explosive
 };
 
+template<typename T>
+constexpr bool operator==(safe_reference<T> const& lhs, safe_reference<T> const& rhs)
+{
+    return !!lhs && !!rhs && lhs.get() == rhs.get();
+}
+
 namespace std
 {
 template <>
@@ -45,7 +51,7 @@ class active_item_cache
     private:
         std::unordered_map<int, std::list<item_reference>> active_items;
         std::unordered_map<special_item_type, std::list<item_reference>> special_items;
-
+        std::unordered_map<int, std::unordered_map<item*,safe_reference<item>>> active_items_index;
     public:
         /**
          * Removes the item if it is in the cache. Does nothing if the item is not in the cache.

```

The points I've thought of:

1. use `!!lhs && !!rhs` to check if the safe_reference has expired.
2. Only hash the item* pointer.
3. If there's a collision of memory address, the index will be checked to hold the very item, and has not expired.
4. The expiry of item_reference stored by the list should be equivalent to the expiry of safe_reference stored by index.
5. There will be more memory consumption, but the overall overhead shouldn't be large, as these objects mostly just hold pointers.

Please comment if I missed anything.

PR #66315, opened 2023-06-19
Fix clang-tidy tests on Windows

Qrox:
#### Summary
None

#### Purpose of change
Make clang-tidy work again on Windows after #65806.

#### Describe the solution
1. Remove `gettext` from clang-tidy checks and mark `gettext` and `ngettext` as `[[deprecated]]` to avoid accidentally using them when `libintl.h` is included by another system header. (Cannot use `=delete` because c++ does not allow redeclaring a function as deleted.)
2. Match both `int` and `long` in the random number generator check test due to MinGW-w64 using `int` instead of `long` for the rngs.
3. Add a parameter to `check_clang_tidy.py` to allow specifying the clang-tidy executable path on Windows.
4. Fix an error in the doc string of `check_clang_tidy.py`.
5. Add modification notice to `check_clang_tidy.py` to conform to its license.

None

#### Testing
Tested locally and all lit tests passed. Checked a few files with the compiled `CataAnalyzer` executable and everything worked as intended.

#### Additional context
None

### DIFF
--- a/tools/clang-tidy-plugin/test/check_clang_tidy.py
+++ b/tools/clang-tidy-plugin/test/check_clang_tidy.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 #
+# Modified by the Cataclysm: Dark Days Ahead project.
+#
 #===- check_clang_tidy.py - ClangTidy Test Helper ------------*- python -*--===#
 #
 # Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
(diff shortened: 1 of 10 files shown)


tools/clang-tidy-plugin/test/check_clang_tidy.py
```diff
@@ -20,8 +22,10 @@
     [-check-suffix=<comma-separated-file-check-suffixes>] \
     [-check-suffixes=<comma-separated-file-check-suffixes>] \
     [-std=c++(98|11|14|17|20)[-or-later]] \
+    [-clang-tidy-binary=<path/to/clang-tidy>] \
+    [optional clang-tidy arguments]
     <source-file> <check-name> <temp-file> \
-    -- [optional clang-tidy arguments]
+    -- [optional clang arguments]
```
jbytheway, line 28: I think this was correct; it's really `clang-tidy` arguments.

Qrox: https://github.com/Qrox/Cataclysm-DDA/blob/4f3e2aa0d1873c07013340f4a2347dadf85fb0b8/tools/clang-tidy-plugin/test/check_clang_tidy.py#L105-L110

https://github.com/Qrox/Cataclysm-DDA/blob/4f3e2aa0d1873c07013340f4a2347dadf85fb0b8/tools/clang-tidy-plugin/test/check_clang_tidy.py#L175-L176

Arguments after `--` are passed as clang arguments to `clang-tidy`, only the arguments before `--` are clang-tidy parameters per-se.

PR #66319, opened 2023-06-19
Lower volume of MRE packages

azraneth:
#### Summary
Bugfixes "Lower volume of MRE packages"

#### Purpose of change

MRE Packages are [~2.4 L in volume and ~800g in weight](https://www.dla.mil/Troop-Support/Subsistence/Operational-rations/mre/). In game, they are instead over 3 L weighing  approximately 1500g. This is unrealistic and also inconveniences the player since they don't really fit in pockets they're supposed to fit in.

#### Describe the solution

The MRE specific container beverage_bag is a non-rigid container with a base volume of 0.26 L and a pocket of 0.25 L, meaning a filled bag is twice the volume it's meant to be. This change migrates beverage_bag into zipper_bag, since every image I could find of an MRE beverage bag also included a zipper, meaning they're functionally identical to zipper_bag, 

Keeping beverage_bag as a near-copy of zipper_bag, but figured I might as well reduce bloat while I was at it.

Also reducing weight, but could not find measurements on reasonable weights on the containers or their contents.

#### Testing

Spawned in zombie soldiers and killed them to drop MRE packages and made sure nothing spilled and MREs still spawn sealed when full.
Teleported to military surplus and observed a spawned full MRE package
Loaded in a save that had beverage bags to make sure they migrated properly.

### DIFF
--- a/data/json/itemgroups/SUS/mre_packages.json
+++ b/data/json/itemgroups/SUS/mre_packages.json
@@ -67,18 +67,18 @@
       { "item": "matches", "charges": 20 },
       { "item": "napkin", "count": 8 },
       { "item": "gum", "charges": 2 },
-      { "item": "pur_tablets", "charges": 6, "container-item": "beverage_bag" },
-      { "item": "sugar", "charges": 2, "container-item": "beverage_bag" },
+      { "item": "pur_tablets", "charges": 6, "container-item": "bag_zipper" },
+      { "item": "sugar", "charges": 2, "container-item": "bag_zipper" },
       { "item": "spork" },
       {
         "distribution": [
-          { "item": "instant_coffee", "count": 2, "container-item": "beverage_bag" },
-          { "item": "herbal_tea_bag", "count": 2, "container-item": "beverage_bag" },
-          { "item": "tea_bag", "count": 2, "container-item": "beverage_bag" },
-          { "item": "tea_bag_chamomile", "count": 2, "container-item": "beverage_bag" },
-          { "item": "tea_bag_dandelion", "count": 2, "container-item": "beverage_bag" },
-          { "item": "tea_fruit_bag", "count": 2, "container-item": "beverage_bag" },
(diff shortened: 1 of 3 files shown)


Karol1223:
I don't see why you're removing the beverage bag? It is a real item. Just because it's not used in MREs doesn't mean it shouldn't be in the game. It could find use elsewhere.

Karol1223:
Regardless of our opinions your PR also doesn't migrate anything. You just removed an item. Without migrating or obsoleting it. This will cause errors in saves that already have this item.

azraneth:
That's a valid point, I'd love to take a guide on how to migrate items properly.

GuardianDll:
https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md#obsoletion-and-migration

casswedson:
do make sure the full mres still spawn sealed; due to some eh... code stuff only **full** containers spawn sealed, gotta match the container's capacity and the content's volume to the ml

bombasticSlacks:
> do make sure the full mres still spawn sealed; due to some eh... code stuff only full containers spawn sealed, gotta match the container's capacity and the content's volume to the ml

agreed, changing MRE info can lead to some oddities.

azraneth:
after a shrinking of 1ml (turns out 2*4 is 8 and not 7) things do spawn sealed again

Maleclypse:
I'm confused, none of the bags in an MRE are resealable from what I know.  So adding zipper bags seems off.

azraneth:
Specifically beverage bags made for being filled with water, a heater, and a flavoring packet are resealable, since otherwise they wouldn't work very well ^^. The only reason I'd see to not make them just a zipper bag is for them to be more durable in some way since they are military grade, but afaik there's no reason for that either.

Maleclypse:
Cool deal.  Can you rebase or otherwise resolve the conflicts that have occurred? Then this will be ready for final review/merge.

PR #66320, opened 2023-06-19
Whitelist Deserter Bandit camp from overmap_test.cpp

Maleclypse:
#### Summary
None

#### Purpose of change
Globally unique locations tend to fail the placement test unless you've made them high enough weight that they will always spawn in first few maps.

#### Describe the solution
Adds new bandit location to whitelist for spawning test

Making globally unique locations opt out of the placement test but I feel like that's a worse idea than adding this to the white list.

#### Testing
CI


bombasticSlacks:
Need to alphabetize the list

PR #66322, opened 2023-06-19
alter/dilute everyday carry drops

anoobindisguise:
#### Summary
Balance "zombies everyday carry less lethal items too"

#### Purpose of change
Not everyone carries a gun for self defense and this would probably still extend to the early days of the apocalypse.

#### Describe the solution
Add stun guns and tear gas spray to the pool of items.
Also, make these items more likely to be damaged as they could very easily have taken a beating.

More less lethal stuff.
Adjust other drops/rates.

### DIFF
--- a/data/json/monsterdrops/zombie_default.json
+++ b/data/json/monsterdrops/zombie_default.json
@@ -265,11 +265,10 @@
       { "group": "default_zombie_items_pockets", "prob": 20 },
       {
         "distribution": [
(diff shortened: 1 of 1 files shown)


bombasticSlacks:
For this and the other PR boosting the chance of gun not in holster seems bad. We'll run into more "where were they keeping it?" Some chance is fine, but normalizing it seems loose.

bombasticSlacks:
That works too

PR #66324, opened 2023-06-19
adjust zombie food to be more often perishables

anoobindisguise:
#### Summary
Balance "When zombies drop food, it's more frequently perishable"

#### Purpose of change
Zombies only ever carried shelf stable foods with them, but not everyone in a disaster would grab something that would last a long time.

#### Describe the solution
Make the zombie food drop (5% chance of a vending machine food) instead a 5% chance of a larger variety of food which includes vending machine food, but also sandwiches, stuff from their fridge, etc. Much of which will spoil quickly.

Unsure if zombies should carry that much food with them but this should at least have a gameplay effect of zombie-based food delivery becoming less potent as the days creep on. It might be worth applying these to bugout bags too.

### DIFF
--- a/data/json/monsterdrops/zombie_default.json
+++ b/data/json/monsterdrops/zombie_default.json
@@ -249,7 +249,7 @@
       { "group": "softdrugs", "prob": 10 },
       { "group": "harddrugs", "prob": 5 },
       { "group": "phones", "prob": 85 },
-      { "group": "vending_food_items", "prob": 5 }
+      { "group": "zombie_pocket_food", "prob": 5 }
     ]
   },
   {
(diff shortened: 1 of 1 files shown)


Ruminator24:
https://bugoutbagacademy.com/best-bug-out-bag-food/ I would disagree with putting perishable foods in bugout bags, as this site recommends non perishable foods, such as cliff bars, dehydrated foods, mres, etc etc, but it is a possibility that someone could have put some perishable foods they were intent on eating, as long as it's portable; someone's dinner in a plastic bag, half drunk fruit juice, or something similar to that, but mostly for bugout bags, i believe they would mostly only contain non perishable goods.

PR #66327, opened 2023-06-19
Fix map memory not cleared when dragging

irwiss:
#### Summary
None

#### Purpose of change

Fix dragging furniture with no vision leaving memory mark trails

#### Describe the solution

Should clear any decoration id when dragging non-vehicles, oops

#### Testing

Drag furniture around without blindfold, then with, in both cases old memory should be cleared

#### Additional context

https://discord.com/channels/598523535169945603/642349261320880128/1120331594595565598

### DIFF
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -301,7 +301,7 @@ void avatar::memorize_symbol( const tripoint &p, char32_t symbol )
 
 void avatar::memorize_clear_vehicles( const tripoint &p )
 {
-    player_map_memory->clear_tile_vehicles( p );
+    player_map_memory->clear_tile_decoration( p, /* prefix = */ "vp_" );
 }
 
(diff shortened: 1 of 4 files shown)


Qrox:
This does not erase the memory if the furniture is not dragged, but deconstructed for example. I think the proper solution would be clearing the decoration memory if no decoration is drawn on a tile when rendering.

irwiss:
https://github.com/CleverRaven/Cataclysm-DDA/pull/27457 put a cache for it so clearing every draw is probably not an option, even though terrain, by far the most common scenario, is bypassing it, not sure how much of that cache is still in working order

Seems these bugs (as in, decon/destroyed furniture/traps) behave the same since 0.G or earlier, https://github.com/CleverRaven/Cataclysm-DDA/pull/66399 makes an attempt at patching the cache invalidations in case we can get away with just that

PR #66328, opened 2023-06-19
Gutter Unify Tool Part 3

Procyonae:
#### Summary
None

#### Purpose of change

Begins to make the aliases for gutter on roof palette redundant

#### Describe the solution

Runs https://github.com/CleverRaven/Cataclysm-DDA/pull/66140
Aiming to do this in about 5 PRs as opposed to 1 huge PR to hopefully get it through reasonably fast without running into too many merge conflicts

#### Testing

Game loads with no errors
Manually checked the file changes expanding each to check for any edge cases the tool doesn't handle like items placed on one side of gutter through the symbol etc


LovamKicsiGazsii:
To think this would've taken me literal hours to accomplish... thanks for doing these!

Procyonae:
> To think this would've taken me literal hours to accomplish... thanks for doing these!

If there's anything else you want doing that could benefit from some automation I'd be happy to have a crack at it ^-^
I'll have a play and see if I can figure out tweaking other palettes and individual maps with this change too once I've gone through all the roof_palette ones

PR #66329, opened 2023-06-19
Typo fix

Maeyanie:
#### Summary
None

#### Purpose of change

Minor typo fix.

#### Describe the solution

Change "apply them the in a real fight" to "apply them in a real fight"

After reading this section's title I briefly considered changing it to "in the real fight apply it" but then I remembered this isn't written by Yoda.

#### Testing

I mean, I could run the test suite on it if you want, but it's one word in a json file...

Edit: Yes, apparently I should have, as this will need localization file changes too. I'll close this for now and make a new one when I get around to fixing it all.

#### Additional context

image

### DIFF
--- a/data/mods/MMA/martial.json
+++ b/data/mods/MMA/martial.json
@@ -50,7 +50,7 @@
     "looks_like": "manual_karate",
     "type": "BOOK",
     "name": { "str": "Pocket Monster Encyclopedia", "str_pl": "copies of Pocket Monster Encyclopedia" },
-    "description": "This encyclopedia contains a detailed listing of the strengths and techniques of various fictional monsters and how to apply them the in a real fight.",
+    "description": "This encyclopedia contains a detailed listing of the strengths and techniques of various fictional monsters and how to apply them in a real fight.",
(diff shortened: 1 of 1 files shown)


Qrox:
> apparently I should have, as this will need localization file changes too

If by localization files you mean the `.po` files, these are automatically generated and shound NOT be changed.

Maeyanie:
> > apparently I should have, as this will need localization file changes too
> 
> If by localization files you mean the `.po` files, these are automatically generated and shound NOT be changed.

I did, didn't realize that. Is this actually done correctly then?

NetSysFire (approved):
This is the kind of PR where you would use None as a summary by the way.

Maeyanie:
> This is the kind of PR where you would use None as a summary by the way.

Okay, changed. The instructions in the comment don't even mention a "None" type, but I did notice it on others when I was trying to figure out exactly how to make the auto-check happy.

NetSysFire:
Yeah it is not intuitive. I wish we could have PR templates as we have issue templates, but github does not support that one yet.

PR #66331, opened 2023-06-20
Wooden Zombie Line

Drew4484:
#### Summary
Content "New type of zombie, infused with wood."

#### Purpose of change
I wanted to add a slower tankier zombie type without resorting to metal zombies. 

#### Describe the solution
Adds basic, brute, and hulk versions of a wooden zombie. Descriptions note the appearance of a bark-like covering on their skin, advancing to full wooden skin in the hulk stage. A bit tougher and harder hitting than a Kevlar variant, these are slower and have worse attack accuracy. They drop basic zombie gear and assorted tree/wood bits.

None.

#### Testing
WIP


data/json/monsters/zed_wooden.json
```diff
@@ -0,0 +1,153 @@
...
+      [ "stick", 80 ],
+      [ "acorns", 20 ],
+      [ "pinecone", 20 ],
+      { "item": "withered", "prob": 80, "count-min": 1, "count-max": 10 },
+      { "item": "plant_fibre", "prob": 80, "count-min": 1, "count-max": 10 },
+      [ "pine_bough", 10 ],
+      [ "splinter", 90 ],
+      [ "tinder", 50 ]
+    ]
+  }
```
TheShadowFerret, line 152: These don't belong in the monster file 😛
Into monsterdrops/zombie they go

fairyarmadillo:
I don't completely hate the idea, but what's the lore explanation here? I thought the blob didn't really mess with plant matter except when specifically induced to, and most zombies don't follow the mutagenic format of human+other organism*

*Except the thorny shamblers, but those are supposed to be parasitized by triffids.

data/json/monsters/zed_wooden.json
Drew4484: TY, first monster I've made!

Drew4484:
> I don't dislike the idea, but what's the lore explanation here? I thought the blob didn't really mess with plant matter except when specifically induced to, and most zombies don't follow the mutagenic format of human+other organism*
> 
> *Except the thorny shamblers, but those are supposed to be parasitized by triffids.

The idea here is to expand on the theme of blob creatures absorbing a material and reproducing it like the Kevlar zombie family. Trees are plentiful, tough, and I'm sure there are plenty of zombies raised from people who thought to escape into a nearby forest.

fairyarmadillo:
Makes sense. Houses are often made of wood too, so it's not like there's any shortage of it around. Thanks for explaining.

Karol1223:
I'd say that given all of this they should not be dropping pine boughs, pinecones or acorns. If the zombies are reproducing the material same way Kevlar's function, the odds of them growing species-specific parts are low or near zero.

Drew4484:
> I'd say that given all of this they should not be dropping pine boughs, pinecones or acorns. If the zombies are reproducing the material same way Kevlar's function, the odds of them growing species-specific parts are low or near zero.

Fair point, changed that.

Daved27hundred:
Feel free to disregard this, but I want to suggest giving these guys more bash resistance and less cut resistance. Not only because it fits thematically (you chop trees, you chop zombie trees. same thing.) but because the other naturally armored zeds like the skeletal and kevlar lines have high resistance against cutting and piercing damage but are weak to bashing damage, and a switch would be interesting.

fairyarmadillo:
I agree they need to be differentiated from skeletals (and they are already somewhat because they dont have hardtoshoot), but their armor values should match the wood material in materials.json.

alp7292:
why wood zombies are tougher than kevlar. and they should take high fire damage

ilovehans10:
I don't know if this is possible with the infrastructure currently in place for corpses, but I feel like it makes sense for burning the corpse to yield at least some wood ash along with the corpse ash.

Karol1223:
> I don't know if this is possible with the infrastructure currently in place for corpses, but I feel like it makes sense for burning the corpse to yield at least some wood ash along with the corpse ash.

That'd be handled via material I believe. If the corpse material is partially wood it should yield wood ash.

PR #66332, opened 2023-06-20
Fix large sprites not displaying correctly

### LINKED ISSUE #66280: Large sprites are not displayed correctly.
### Describe the bug

Part of a large sprite that occupy a right in-game cell from their origin does not show up:
image
image
Tileset doesn't matter. Renderer doesn't matter. Scaling factor, scaling mode, batching does not matter as well.
Even when sprite is flipped:
image

_Noticed same behavior on screenshots from other people in the discord channel._
Harder to notice with non monster sprites, but if you'll look closely can be noticed on trees like pines and such.

### Attach save file

No special save required.

### Steps to reproduce

1. Have tileset with large sprites like MSXotto+ or UltiCa selected.
2. Spawn a large creature like frogmother, moose, skeletal juggernaut, etc.
3. Observe.

### Expected behavior

Should render that corner tile ¯\_(ツ)_/¯

### Screenshots

_No response_

- OS: Windows 10
- Game Version: b0bf856
- Graphics version: Tiles
- Ingame language: English
- Mods loaded: Dark Days Ahead [dda]

### Additional context

_No response_

Rewryte:
#### Summary
Bugfixes "Fix large sprites not displaying correctly"

#### Purpose of change
Fix #66280

#### Describe the solution
Correct the order in which sprites are drawn.  Row and draw depth data are now stored with `draw_points` and `tile_render_info` respectively to accommodate for this.  Care had been taken to avoid the reintroduction of other draw order related bugs.

#### Testing
Builds ok.
Large sprites now display correctly.
Fog is drawn on top of large sprites correctly.
Creatures are drawn behind tall sprites in front of them as they should.
No detectable increase in draw time.
Screenshot below uses debug clairvoyance.

Spathi Pkeloucht _2023-06-20T12-21-24

### DIFF
--- a/src/cata_tiles.cpp
+++ b/src/cata_tiles.cpp
@@ -1209,7 +1209,7 @@ struct tile_render_info {
     int draw_min_z = -OVERMAP_DEPTH;
     tile_render_info( const tripoint &pos, const int height_3d, const lit_level ll,
                       const std::array<bool, 5> &inv, const int draw_min_z )
-    : pos( pos ), height_3d( height_3d ), ll( ll ), invisible( inv ), draw_min_z( draw_min_z ) {
+        : pos( pos ), height_3d( height_3d ), ll( ll ), invisible( inv ), draw_min_z( draw_min_z ) {
     }
 };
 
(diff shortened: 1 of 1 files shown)


src/cata_tiles.cpp
```diff
@@ -1599,69 +1611,64 @@ void cata_tiles::draw( const point &dest, const tripoint &center, int width, int
...
-            while( !here.dont_draw_lower_floor( p_draw ) && p.pos.z - p_draw.z < max_draw_depth ) {
-                p_draw.z -= 1;
-            }
-            draw_points_3d[p_draw.z].emplace_back( p );
-        }
-
         // Start drawing from the bottom-most z-level
         int cur_zlevel = -OVERMAP_DEPTH;
         do {
             int cur_height_3d = ( cur_zlevel - center.z ) * height_3d_mult;
```
Qrox, line 1629: I think this should be using the `height_3d` member of `tile_render_info`, because otherwise the 3d height will accumulate across different tiles.

Rewryte: It is part of the WIP height_3d changes for isometric 3D that got pushed out early.  Isometric tilesets shouldn't get to that point so it should be harmless, but I could comment it out nevertheless.

bombasticSlacks:
All these renderer changes are much appreciated 😄

PR #66333, opened 2023-06-20
Various perf improvements

SurFlurer:
#### Summary
none

#### Purpose of change
Faster, faster, faster.

#### Describe the solution
1. When deserializing the vehicle parts of a given vehicle, reserve `parts` to avoid  emplace-reallocation.
2. When refreshing the vehicle, when generating fake parts, move the temporary vparts instead of copying them to avoid deconstruction and copy-construction.
3. Early exit when there's definitely nothing to butcher to avoid checking all items nearby for tools.

none

#### Testing
CI should tell.

#### Additional context

When generating fake parts, the item colonies of the real parts are also copied. This becomes an issue for Blaze Industry mod, because there's a 2000L cargo space vp that is also obstacle, so the copy overhead becomes large. Should we implement a dedicated copy-ctor that does not copy the item colony to generate fake parts?

### DIFF
--- a/src/action.cpp
+++ b/src/action.cpp
@@ -619,7 +619,7 @@ bool can_butcher_at( const tripoint &p )
 {
     map_stack items = get_map().i_at( p );
     // Early exit when there's definitely nothing to butcher
-    if( items.size() == 0 ) {
+    if( items.empty() ) {
         return false;
     }
     Character &player_character = get_player_character();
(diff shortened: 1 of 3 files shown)


irwiss:
making the items colony a shared ptr is also an option

SurFlurer:
Thanks for merging!

PR #66334, opened 2023-06-20
Player character now can get sick with common cold or flu only after contacts with sick NPCs or ferals

### LINKED ISSUE #65881: [Port Request] TISH Stuff

I saw three Prs for Tish which I thought would make perfect additions to dda and wanted to put them here:
https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/71
https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/66
https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/68

### Solution you would like.

I think these are great gameplay additions and I would like to see them ported to dda in some form, especially the npc command because it opens up a lot of tactical possibilities during combat encounters.

_No response_

### Additional context

_No response_

Night-Pryanik:
#### Summary
Features "Player character now can get sick with common cold or flu only after contacts with NPCs or ferals"

#### Purpose of change
* Closes #65881.

#### Describe the solution
Ported https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/66, https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/68 and https://github.com/AtomicFox556/Cataclysm-EOD/pull/97/commits/fe9c08d700ed343f164a7d5149aee512f4855712:

- All ferals now have a newly added `FERAL` species type.
- Every feral or _random_ NPC now have a 1% chance to spawn sick with common cold or flu.
- Every 30-300 seconds these sick characters may cough (50% chance).
- If you're standing right next to them right when the the "time has come", you have a chance to get sick too, unless you have a `Disease Immune` trait.
- After you're cured from these diseases, you get an immunity for them. On default season length (91 days) the immunity duration is ~120 days.
- Neither ferals nor random NPCs don't suffer from these diseases (yet), the only symptom is their (non-painful) cough.

None.

#### Testing
Built a cage for me and NPC with the effect. Waited for NPC to cough on me. After some coughs I got sick too.

#### Additional context
None.


Raikiri:
you absolutely can get common cold/flu if you just hang out in bad weather in a forest without contacting anybody whatsoever. i guess even without exposure viral particles just live inside of you and wait for your immune system to become weak enough, you don't necessarily need to come into a direct contact with a carrier.

Night-Pryanik:
> you absolutely can get common cold/flu if you just hang out in bad weather in a forest without contacting anybody whatsoever

This case most probably means that you were already infected prior to this forest voyage, and bad weather lowered your immune response which allowed infection to manifest itself.

> viral particles just live inside of you

I'm not aware of fact that rhinoviruses (up to 80% of colds caused by them) can live inside a human body for prolonged time periods without being killed by immune system. Do you have any sources of that?

I've read of polar explorers who live and work for prolonged time periods in really harsh conditions of Arctic/Antarctic. I think no one will argue that these regions are places which put a great strain on your immune system every day. Despite that, these explorers almost never get cold or flu during their expeditions. But after their return to home (and corresponding explosive growth of contacts with other people) they begin to get sick with cold and flu, and their illnesses are often much more severe than usual.

fairyarmadillo:
> you absolutely can get common cold/flu if you just hang out in bad weather in a forest without contacting anybody whatsoever. i guess even without exposure viral particles just live inside of you and wait for your immune system to become weak enough, you don't necessarily need to come into a direct contact with a carrier.

Night is correct, in cases like this it is because you had exposure prior to leaving town. The incubation period can sometimes be up to two weeks since contact. Cold and flu viruses can sometimes also live up to 24 hours on surfaces, so there's a rare chance you'll touch something that someone else touched without the two of you seeing each other. That probably isn't worth modeling in CDDA though, as there's no way to track where zombies have been and it's pretty rare IRL anyway.

@Night-Pryanik - I think this would be better if some percentage of "coughs" were invisible, and if sometimes ferals (but not NPCs) coughed despite not being sick. Otherwise the player will be able to recognize and identify every sick person they see, which is not the case IRL. Many times you get sick from people who appear to be healthy.

Night-Pryanik:
I'm afraid these "invisible coughs" might be the source of frustration and questions like "where did I get that cold from?" from players.

rselias:
Planning to make dust/filter/gas masks have any effect on this, or leaving that for another PR?

Hirmuolio:
Are zombies too "not human" to act as carriers for human diseases?

Maleclypse:
> Are zombies too "not human" to act as carriers for human diseases?

There's been at least some discussion of having a "coughing/infectious" line of zombies.

Night-Pryanik:
> Planning to make dust/filter/gas masks have any effect on this, or leaving that for another PR?

This is already a thing. When sick character coughs on you, the environmental protection of your mouth takes a step in to check if you get sick.

Night-Pryanik:
> Are zombies too "not human" to act as carriers for human diseases?

I assume zombies are "too dead" to carry such diseases.

Night-Pryanik:
Ok, now both ferals and NPCs will cough only 50% of time. They still may infect you when you're in close contact with them, even if they don't cough at you.

Maleclypse:
Hi Night, this is good for merge if the immunity goes away since usually people are carrying different virus strains and the flu vaccine works by targeting over a hundred variant viruses.

Night-Pryanik:
You mean completely remove the immunity mechanics? This will change our status quo, and players might get sick almost unlimited number of times per year.

Maleclypse:
> You mean completely remove the immunity mechanics? This will change our status quo, and players might get sick almost unlimited number of times per year.

That is accurate that it could possibly happen, another option would be to create a test that gives us an understanding of how many times a year the average PC would catch a cold/flu and then we could model contagiousness to an average of 3-4 colds per year and a flu every other year?

Night-Pryanik:
> we could model contagiousness to an average of 3-4 colds per year

Unless I'm understanding you wrong, our current implementation is already modeling 3 colds per year for non-disease-resistant characters.

Maleclypse:
> > we could model contagiousness to an average of 3-4 colds per year
> 
> Unless I'm understanding you wrong, our current implementation is already modeling 3 colds per year for non-disease-resistant characters.

Yes and I respect that, it accomplishes the same thing as I think is s desired state. When I was getting opinions on what this needed prior to merge the consensus was to remove the immunity because people can and do catch nearly back to back colds. So my thought process was ok what if we remove the immunity but then try to tweak the infection rates so that the change maybe doesn’t require weeks of people complaining about “overtuned colds” before a new solution is proposed.

Night-Pryanik:
The thing is that in my proposed solution immunity _is_ the mechanism to tweak infection rates. Removing the immunity equals to removing this mechanism, and we'll no longer be able to tweak infection rates.

Maleclypse:
> The thing is that in my proposed solution immunity _is_ the mechanism to tweak infection rates. Removing the immunity equals to removing this mechanism, and we'll no longer be able to tweak infection rates.

Ok so NPC's that spawn with the disease, do they get better from it as time passes and have chances to be reinfected by ferals or other npcs or are they perpetually sick and will infect the PC as soon as its immunity ends?

Procyonae:
Immunity seems important otherwise if you play with NPCs as soon as one NPC gets it then they'll all catch it and then when the first one recovers they'd catch the same cold again off the other NPCs, the only way around that would be to make peeps stop being contagious before they fully recover (unless it already works like that and I'm blind)

fairyarmadillo:
Given that diseases are now passing between small groups of people in a relatively small area, I think temporary immunity is a good enough model. It's not like ferals are taking intercontinental flights and bringing foreign germs home, it's likely the same one or two cold and flu viruses that are simply mutating over time to reinfect people.

Night-Pryanik:
> Ok so NPC's that spawn with the disease, do they get better from it as time passes and have chances to be reinfected by ferals or other npcs or are they perpetually sick and will infect the PC as soon as its immunity ends?

Infected NPCs have the same disease duration as the player (1-14 days for cold and 3-10 days for flu). After that they are considered cured and won't ever get sick again (yet). Only freshly spawned NPCs have a chance to spawn infected.

Night-Pryanik:
> as soon as one NPC gets it then they'll all catch it

No, at the moment NPCs don't catch cold or flu from other infected characters.

Maleclypse:
Looks like you rebased onto the broken test. Needs a rebase.

PR #66339, opened 2023-06-20
Hound afterimages also copy their host nicknames

John-Candlebury:
#### Summary
Balance "Hound afterimages also copy their host nicknames"

#### Purpose of change
Giving a nickname to the main hound so you can easily tell it apart is an exploit, this fixes it.

#### Describe the solution
image

#### Testing
See image


GuardianDll:
how evilish XD

PR #66347, opened 2023-06-20
Guns need more skill, performing much worse at low skills

### LINKED ISSUE #64990: Shooting is far too accurate

Within about 12 tiles against a human-sized enemy, guns in single-fire mode almost never seem to miss if you aim, regardless of light or weather conditions. This makes combat incredibly easy once you find an M4 or other rifle with plentiful ammunition.

https://youtu.be/LyWozhgx83Q?t=1807 - Here is an example of what I'm talking about. In this video at the timestamp, a character with 0 marksmanship or gun skill, 10 perception and dexterity, in total darkness, is able to score an incredibly high hit percentage and wipe out a horde of zombies at close range while taking very little damage. I don't think the situation in this example should be unwinnable, just that the hit rate is too high.

This applies to both players and monsters, and seems to extend to throwing as well. People commonly complain that ferals, for instance, are much too accurate, but they use similar math to player characters who see similar success rates at 1 rank of throwing or higher.

### Solution you would like.

https://americanshootingjournal.com/nypd-accepts-the-reality-that-sight-shooting-is-not-used-in-cqb/ This article from the American Shooting Journal uses NYPD data to show that across 36 close-quarters shootings where a subject was hit, officers fired 311 times and only 36 of those shots struck the target, for a hit rate of 12%.

I'm not suggesting a ~90% reduction in accuracy. These officers were not always aiming precisely, the subjects may have used cover, and police are taught to fire multiple times in succession rather than aim between each shot, which is currently the dominating meta in Cataclysm (the article has further data relevant to this). But these are trained professionals, not random survivors, and to me these numbers suggest that:

* The accuracy rate against moving targets, even when precisely aiming, should be much lower than it is.
* Precisely aiming should be more difficult and time consuming, especially at low skill.

I think also lighting should count for more than it does currently. In the linked video, the character has normal human night vision and it doesn't seem to impact anything other than her visual radius. In reality, it would be harder to hit targets you couldn't clearly see even if you were on the upper end of natural human night vision capability. Obviously if you have mutant vision, bionics, or goggles that might change things.

If I had to chalk the issue up to any one thing, I would say that precisely aiming is too fast and easy, and it's nearly always the best idea to do it unless something is in melee range with you, in which case you can just sort of walk a few steps away.

Further penalties could be given to low-skill characters. If you handed a 0 skill person a gun, I would expect them to not intuit how iron sights work, for instance, and they would be badly hampered by poor stance, unfamiliarity with things like the safety, etc.

Barrel length could reduce accuracy in close range.

A point shooting or CQB proficiency could be introduced, with a penalty to aim speed in close range if the character doesn't have it.

Grabs should badly penalize the use of long guns, but not so much for handguns.

Unskilled characters could fire prematurely while aiming, but that might be stupid/annoying. That might be a neat thing for the shakes effect though.

### Additional context

Shooting little enemies like wasps and flesh raptors feels about right, so any changes made to general accuracy might need to tweak size penalties so that these enemies do not become unhittable.

There's also a helpful graphic here, provided by the New York Times, that demonstrates firing accuracy in police encounters at ranges relevant to our player characters: http://www.wehaitians.com/a%20hail%20of%20bullets%20a%20heap%20of%20uncertainty.html

Further data about police accuracy during use of deadly force at different ranges can be found here https://www.politifact.com/factchecks/2018/may/25/shannon-watts/do-more-7-10-police-bullets-miss-their-mark-gun-co/

According to the above data, firing accuracy by police in close quarters deadly force scenarios is usually around roughly 35 percent.

### LINKED ISSUE #66326: Aiming at low skill levels should be considerably less powerful

Currently the player is a crack shot even at zero skill and can flawlessly hit every shot on enemy zombies if they precise aim. Their aiming is too fast, and their odds of landing a hit are too high (no matter how much a novice aims their gun, they're never going to get a solid hit %).
Police in general have pretty low hit rates, from 20%-33% or so and are trained with their guns. That alone implies that the current hit rate (90-100%) See some stats listed in #64990.

### Solution you would like.

Base aiming speed should be lower at low skill, and scale upwards towards normal levels at higher guns/marksman skill. Hit rates should be lowered (a simple fix would be to adjust the size-based hit rates so that normal sized enemies have a lower % of hit rate). Enemies moving and especially leaping should have a more serious effect of throwing off your aim. For a very low skill shooter, most shots should be misses (which is currently not the case at all)

There are a million ways to work around this and I don't have any definite solution. I am not super familiar with aiming code.

### Additional context

_No response_

anoobindisguise:
#### Summary
Balance "The player is substantially less effective with guns at low skill values"

#### Purpose of change
Fixes #64990
Fixes #66326
The player was too much of an instant expert when it came to guns. Even with zero skill they could easily hit 100% hit rates with precise shots at reasonably distant ranges and could aim quickly enough to gun down an approaching horde without risk. While guns are "the great equalizer" for a reason, and definitely need less training to be effective than bows or melee weapons,  firearm novices are not this effective IRL, and even trained combatants like police officers would rarely have 100% hit rate vs even a stationary target in practice, much less in the field against moving and hostile targets. 0 skill is meant to represent total unfamiliarity, i.e. never fired a gun before, and while learning to shoot is a skill you can pick the basics of up in a day or so and get to a decent level of competency in a reasonable time frame compared to mastering fencing or boxing or whatever, I wanted the learning process to be a lot more painful.

#### Describe the solution
- Increase penalty for having less than 10 skill from 3 dispersion per skill point to 10 dispersion per skill point.
- Previously the above penalty was added to the weapon's base dispersion multiplied by (1.25 + the amount of skill you are below 5 times 0.75). Now it is (1.25 + the amount of skill you are below 5 times 2).
- So for example if a gun had 200 base dispersion and you had 2 skill, its base dispersion of 200 was multiplied by (1.25 + 3 * 0.75 ) = 3.5x = 700 and then added to 3 * 7 (721).
- Now, if you had the same 200 dispersion gun and you had 2 skill, its base dispersion of 200 was multiplied by (1.25 + 3 * 2 ) = 7.25x = 1450 and then added to 10 * 7 (1520). These penalties level off fairly quickly as you approach 5 skill.
- In case of 0 skill and a 200 dispersion gun the penalty used to be 1030. Now it is 2350.
- Aiming a gun is no longer just manipulation score. It is now 60% lifting score (arms), 20% grip score (hands) and 20% manipulation (hands)
- Normally re-orienting your aim always scales your "amount of being aimed" linearly down to 0 at 180 degrees in rotation. Now, that 180 degrees is reduced by lower marksmanship skill, being divided by the amount your skill is less than 5 (so at 0 skill changing your aim radius by 30 degrees is enough that you have to completely re aim from scratch)
- Aiming speed is much more influenced by gun skill. Previously, base aim speed cap ranged linearly from 20-30 from 0-10 gun skill. Now It is a base of 5, with +4 speed per skill level up to level 5, then +1 per level as normal up to 10, giving a range of 5-30 and reaching 25 at skill 5 just like before. This means if you "cheat" your way to fast aim speeds by having a high dexterity and target acquisition tools such as a laser sight, you will still be limited somewhat by your skill. 
- Total base aim speed is reduced by 60% at skill zero, as skill "double dips" to pave over many of the other bonuses to aim speed and prevent you from making use of them. This debuff lessens in potency linearly to 7.5 skill, at which point it is completely non existent.

Oh my god there are so many I don't even know to start. Open to suggestions here.

#### Testing
I loaded this up in game without errors, and brought a dude into town with a debug spawned ar-15 and lots of bullets. I found things to be considerably more difficult than before, but not to a degree that it felt absurd, and was basically no different once I set my skills to be high (5+) using debug. This is super subjective though and this will almost certainly require other people to experience the changes.
I still think that if you have 0 skills whatsoever a gun is going to be by far the most effective weapon you can use, but it's considerably more painful to do now that you cannot just murderize everything the moment the gun is in your hands.

### DIFF
--- a/tests/ranged_balance_test.cpp
+++ b/tests/ranged_balance_test.cpp
@@ -170,7 +170,7 @@ static void test_shooting_scenario( npc &shooter, const int min_quickdraw_range,
         CAPTURE( shooter.get_modifier( character_modifier_ranged_dispersion_manip_mod ) );
         CAPTURE( good_stats.n() );
         CAPTURE( good_stats.margin_of_error() );
-        CHECK( good_stats.avg() > 0.05 );
+        CHECK( good_stats.avg() >= 0.05 );
     }
     {
         const dispersion_sources dispersion = get_dispersion( shooter, 500, max_good_range );
--- a/tests/ranged_balance_test.cpp
+++ b/tests/ranged_balance_test.cpp
@@ -162,7 +162,7 @@ static void test_shooting_scenario( npc &shooter, const int min_quickdraw_range,
     {
         const dispersion_sources dispersion = get_dispersion( shooter, 300, min_good_range );
         firing_statistics good_stats = firing_test( dispersion, min_good_range, Threshold( accuracy_goodhit,
-                                       0.5 ) );
+                                       0.05 ) );
         INFO( dispersion );
         INFO( "Range: " << min_good_range );
         INFO( "Max aim speed: " << shooter.aim_per_move( *shooter.get_wielded_item(), MAX_RECOIL ) );
@@ -190,7 +190,7 @@ static void test_shooting_scenario( npc &shooter, const int min_quickdraw_range,
 static void test_fast_shooting( npc &shooter, const int moves, float hit_rate )
 {
     const int fast_shooting_range = 3;
-    const float hit_rate_cap = hit_rate + 0.3f;
+    const float hit_rate_cap = hit_rate + 0.5f;
     const dispersion_sources dispersion = get_dispersion( shooter, moves, fast_shooting_range );
     firing_statistics fast_stats = firing_test( dispersion, fast_shooting_range,
                                    Threshold( accuracy_standard, hit_rate ) );
@@ -249,32 +249,32 @@ TEST_CASE( "unskilled_shooter_accuracy", "[ranged] [balance] [slow]" )
     SECTION( "an unskilled shooter with a common pistol" ) {
         arm_shooter( shooter, "glock_19" );
         test_shooting_scenario( shooter, 4, 5, 17 );
-        test_fast_shooting( shooter, 60, 0.3 );
+        test_fast_shooting( shooter, 60, 0.15 );
     }
     SECTION( "an unskilled archer with a common bow" ) {
         arm_shooter( shooter, "shortbow", { "bow_sight_pin" }, "arrow_field_point_fletched" );
         test_shooting_scenario( shooter, 4, 4, 13 );
-        test_fast_shooting( shooter, 90, 0.3 );
+        test_fast_shooting( shooter, 90, 0.1 );
     }
     SECTION( "an unskilled archer with a common crossbow" ) {
         arm_shooter( shooter, "crossbow", {}, "bolt_makeshift" );
         test_shooting_scenario( shooter, 4, 5, 17 );
-        test_fast_shooting( shooter, 80, 0.3 );
+        test_fast_shooting( shooter, 80, 0.15 );
     }
     SECTION( "an unskilled shooter with a common shotgun" ) {
         arm_shooter( shooter, "remington_870" );
         test_shooting_scenario( shooter, 4, 4, 19 );
-        test_fast_shooting( shooter, 80, 0.3 );
+        test_fast_shooting( shooter, 80, 0.15 );
     }
     SECTION( "an unskilled shooter with a common smg" ) {
         arm_shooter( shooter, "mp40semi" );
         test_shooting_scenario( shooter, 4, 5, 18 );
-        test_fast_shooting( shooter, 80, 0.3 );
+        test_fast_shooting( shooter, 80, 0.15 );
     }
     SECTION( "an unskilled shooter with a common rifle" ) {
         arm_shooter( shooter, "ar15" );
         test_shooting_scenario( shooter, 5, 5, 25 );
-        test_fast_shooting( shooter, 100, 0.3 );
+        test_fast_shooting( shooter, 100, 0.15 );
     }
 }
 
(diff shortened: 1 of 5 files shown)


Maleclypse:
Confirming that this will affect NPCs and monsters using the fake NPC as well?

anoobindisguise:
> Confirming that this will affect NPCs and monsters using the fake NPC as well?

I assume NPCs use ranged.cpp but in honesty I have no clue. Monsters should be totally unaffected but the enemies who have guns are either robots (and accuracy is warranted) or ferals (who have a massive debuff to accuracy as is).
I don't know how NPCs aim. is it different than the player?

tests/ranged_balance_test.cpp
```diff
@@ -170,7 +170,7 @@ static void test_shooting_scenario( npc &shooter, const int min_quickdraw_range,
         CAPTURE( shooter.get_modifier( character_modifier_ranged_dispersion_manip_mod ) );
         CAPTURE( good_stats.n() );
         CAPTURE( good_stats.margin_of_error() );
-        CHECK( good_stats.avg() > 0.5 );
+        CHECK( good_stats.avg() > 0.05 );
```
bombasticSlacks, line 173: might want to widen this a bit?

fairyarmadillo:
8 stat 0 skill combat report: In these tests I will be teleporting into the middle of town with multiple glock 22s and attempting to defeat as many enemies as possible. I'll move around obstacles to try to funnel enemies, but I won't be using things like roofs, basements, or making a full retreat. I'll be standing my ground until everything is dead or my character is. Hordes are off.

The expected outcome here is that I make a dent in the zombies but can't kill them so quickly that they don't eventually get me.

image
ATTEMPT 1: I was able to kill 8 zombies before a pack of dogs surrounded me and killed me.

image
ATTEMPT 2: Using this fenced in area, I was able to corral and evade the zombies long enough to kill 24 of them. I had ample opportunity to escape and could have survived to fight another day, but wound up sticking around so long that the noise attracted a pretty ungodly horde.

image
ATTEMPT 3: The evac shelter was unexpectedly attacked by 4 mi-go. Using my twin glocks, I was able to repel all four of them, but not before sustaining heavy damage. My NPC companion was killed in the fighting.

Undeterred, I pursued my quarry to the mi-go scout tower nearby and defeated two more before succumbing to my wounds.

image
ATTEMPT 4: In Amity, I was only able to defeat 4 enemies before being surrounded. This is mostly because I ran into a police station for cover, not realizing I was boxing myself in.

This feels about right so far. You can do a lot of damage with guns and no skill, but you're no longer able to just completely stomp hordes out with no skill and a couple of pistols. Dogs have turned out to be the real threat to gun users, as they're fast and hard to hit, but that is appropriate - police use dogs against armed suspects for exactly this reason IRL.

fairyarmadillo:
TEST ROUND 2: These are all with 3 marks/rifles/dodge and a starting survivor equipped with an AR-15 and plenty of spare mags.

image
Luigi was doing pretty well but then 5 feral humans came out of nowhere and rocked him to death because he had no armor. Hnnngh.

image
Filiberto racked up 10 kills before getting absolutely swarmed. He also had plenty of time to retreat and places to go to, but for the purposes of this test he just wandered around in the intersection capping zombies in broad daylight in the middle of a big town.

image
Houston did a bit better. Without any terrain to trip enemies up on, he got 26 kills before being overwhelmed.

Overall both tests feel about right. A gun will put down a given zombie very quickly, but it will not on its own answer for a crowd. You need use confined spaces, obstacles, rooftops, basements, etc. to keep from getting swarmed, or you need auto-fire, grenades, or some other tool to thin crowds. You also need armor so that you don't get too banged up while mowing them down.

fairyarmadillo:
Testing with a compound bow and wooden broadhead arrows:
0 skills, 10 str - You basically don't crit, but can land good hits for up to 40 damage. Precise shots take just over 6 seconds.
3 skills, 10 str - Precise shots take just over 4 seconds and crit about 1/3 of the time. This is a significant debuff from before, were basically every precise shot within 5 tiles was a crit, and they were much faster. The bow still feels very usable as a tool for stealth kills, but it no longer seems feasable to use it as a primary weapon when multiple enemies know where you are. This is probably a good thing. Creping around at the edge of town in broad daylight, I was able to kill any lone or pair of zombies I found, but once I got discovered and swarmed I had to drop the bow and run or die.

fairyarmadillo:
image
The speed and accuracy of vehicle turrets are unaffected. I assume they use the monster gun system. 

I think they should be less accurate, but I don't know how CROWS systems perform IRL compared to human shooters.

tests/ranged_balance_test.cpp
anoobindisguise: Ok so, "good_stats.avg()" used to need to hit 0.5, but the test failure showed 0.15-0.49 or so with one outlier (bow and arrow) at 0.09. The low results came from being unskilled. I'm not 100% sure how it is calculated but I can try and investigate it. It is wildly more variable than it used to be for sure so it may no longer even mean anything as a test statistic.

BjoHart:
Maybe someone could work this in?

"Fire Menu" New keybindings just for aim, then shoot manual #66829 

Is your feature request related to a problem? Please describe.

When aiming with ranged weapons, i very often have to press the button for steady my aim, to fire when I am exactly ready.
This is the only way for me to deal with many zombies in a firefight.
Solution you would like.

I would like to have an option to jump to the three different shooting modes quickly, but not to fire, just to aim.
Then I can make fine adjustments if I want to.
Just some key bindings would do the job.
Describe alternatives you have considered.

Pressing the button for steading the aim 100000 times more........
Additional context

No response

PR #66348, opened 2023-06-21
Added power consumption for generator UPS

### LINKED ISSUE #60613: Material energy consumption and efficiency

We have a few non-CBM non-vehicle power generators now, primarily used on the various exoskeletons/exosuits. For the ones that consume fuel, the fuel is consumed as a single charge to generate power, rather than using the appropriate material energy. Additionally, there is currently no way to modify the efficiency of that consumption. 

### Solution you would like.

Two additional fields, something along the lines of
"consume-as-fuel": "true"
"fuel-efficiency": "40"

Continuing to burn lamp oil and ammonia at the same rate.

### Additional context

List of known items this would help with:
power_armor_generator
ice_utility_exoskeleton_on
exo_back_gasoline
exo_back_fuelcell

bombasticSlacks:
#### Summary
None

#### Purpose of change
fixes #60613

#### Describe the solution
Adds fuel efficiency
When checking UPS's if they have efficiency and fuel then pass that info to contents::consume_ammo
in consume_ammo do the math for how much energy the fuel makes.

#### Testing
Still WIP

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -8838,7 +8838,12 @@ units::energy Character::consume_ups( units::energy qty, const int radius )
     if( qty != 0_kJ ) {
         std::vector<const item *> ups_items = all_items_with_flag( flag_IS_UPS );
         for( const item *i : ups_items ) {
-            qty -= const_cast<item *>( i )->energy_consume( qty, tripoint_zero, nullptr );
+            if( i->is_tool() && i->type->tool->fuel_efficiency >= 0 ) {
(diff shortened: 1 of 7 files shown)


SurFlurer:
Maybe this needs to be documented as well? Since this field is compulsory for generator UPS items to work properly.

bombasticSlacks:
Wtf I swear I documented it... 🤔 Must have gotten lost in a stash. Yeah I can do that.

PR #66352, opened 2023-06-21
Burning down forests for fun and profit (and wood ash)

### LINKED ISSUE #56242: Burning building results in spam about terrain t_open_air where furniture is f_rubble
### Describe the bug

 DEBUG    : Setting terrain t_open_air at (127,42,1) where furniture is f_rubble.  Terrain is_open_air
If this is intentional, set the ALLOW_ON_OPEN_AIR flag on the furniture

 FUNCTION : ter_set
 FILE     : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\map.cpp
 LINE     : 1769
 VERSION  : 53f9bd3

This happens a lot when I accidentally set an apartment tower on fire (I intended to set fire to a hedge maze adjacent to it).

### Steps to reproduce

Partial a guess:

1. Set fire to a building higher than one story.
2. Watch spam until you give up and ignore it.

### Expected behavior

No spam when buildings burn and things cave in (a guess is that zombies smashing through buildings ought to be capable of generating similar spam as well).

### Screenshots

Offset screen image of the hedge maze and adjacent apartment tower:

Screenshot (226)

- OS: Windows
    - OS Version: 10.0.19043.1586 (21H1)
- Game Version: 53f9bd3 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]
+ "mod" in the form of hacking portal storms to do nothing as the actual mod to do so was rejected.

### Additional context

_No response_

### LINKED ISSUE #65041: Plants do not leave wood ashes despite there being patches of burned vegetation

So the wood-ash patches strongly hint that they are patches of forest burned to leave only the residue. But if I play pyromaniac with a forest for real, setting a tree alight and watching it burn with glee, nothing is left behind! Where is my wood ash?

### Solution you would like.

Burning a tree should leave wood ash - that is, setting fire to it with a lighter or other firestarter and letting it burn to completion.

_No response_

### Additional context

_No response_

RenechCDDA:
#### Summary
Bugfixes "Trees and other FLAMMABLE_ASH terrain leaves behind ash when burned down"

#### Purpose of change
* Fixes #56242
* Fixes #65041

#### Describe the solution
-Tell fires to actually leave ash for FLAMMABLE_ASH terrain. RNG from 10-1000 charges, or approximately 16-1640mL.

-**Increased most ash left by the burned ground map extra by a factor of 10**. This still leaves... probably not enough ash, but it leaves *more*. Previously it would spawn only a maximum of 100 charges of wood ash(~164mL) for an entire tree.

-Add a terrain transformation to rock when terrain is burned on Z -2 or lower (previously turned into dirt, bit of a 🤔)

-Delete furniture before applying terrain transformation to avoid floating furniture & related errors (#56242)

field_processor_fd_fire should probably process furniture burning *before* terrain burning, to avoid instances where flammable terrain in a fire is set to null, but it's a very minor edge case that only applies on the turn of destruction

I decided not to mess with that as it could create more bugs and would result in some churn without any real benefit

#### Testing
Ran around burning down forests and homes like some sort of pyromaniac. Confirmed that previously burning down the forests left nothing, confirmed that after changes it left ash. Burn down the same home repeatedly pre changes, got errors about setting furniture (rubble) on open air, burned them down afterwards without a peep.

#### Additional context

### DIFF
--- a/src/map_field.cpp
+++ b/src/map_field.cpp
@@ -1046,10 +1046,17 @@ void field_processor_fd_fire( const tripoint &p, field_entry &cur, field_proc_da
             smoke += static_cast<int>( windpower / 5 );
             if( cur.get_field_intensity() > 1 &&
                 one_in( 200 - cur.get_field_intensity() * 50 ) ) {
+                here.spawn_item( p, "ash", 1, rng( 10, 1000 ) );
                 if( p.z > 0 ) {
-                    // We're in the air
+                    // We're in the air. Need to invalidate the furniture otherwise it'll cause problems
+                    here.furn_set( p, f_null );
                     here.ter_set( p, t_open_air );
+                } else if( p.z < -1 ) {
+                    // We're deep underground, in bedrock. Whatever terrain was here is burned to the ground, leaving only the carved out rock (including ceiling)
+                    here.ter_set( p, t_rock_floor );
                 } else {
+                    // Need to invalidate the furniture otherwise it'll cause problems when supporting terrain collapses
+                    here.furn_set( p, f_null );
                     here.ter_set( p, t_dirt );
                 }
             }
--- a/src/map_extras.cpp
+++ b/src/map_extras.cpp
@@ -1563,7 +1563,7 @@ static void burned_ground_parser( map &m, const tripoint &loc )
     if( iter != dies_into.end() ) {
         if( one_in( 6 ) ) {
             m.ter_set( loc, t_dirt );
-            m.spawn_item( loc, itype_ash, 1, rng( 1, 5 ) );
+            m.spawn_item( loc, itype_ash, 1, rng( 10, 50 ) );
         } else if( one_in( 10 ) ) {
             // do nothing, save some spots from fire
         } else {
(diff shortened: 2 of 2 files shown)


src/map_field.cpp
```diff
@@ -1046,10 +1046,17 @@ void field_processor_fd_fire( const tripoint &p, field_entry &cur, field_proc_da
             smoke += static_cast<int>( windpower / 5 );
             if( cur.get_field_intensity() > 1 &&
                 one_in( 200 - cur.get_field_intensity() * 50 ) ) {
+                here.spawn_item( p, "ash", 1, rng( 10, 1000 ) );
                 if( p.z > 0 ) {
-                    // We're in the air
+                    // We're in the air. Need to invalidate the furniture otherwise it'll cause problems
+                    here.furn_set( p, f_null );
```
NetSysFire, line 1052: Is there a way to bash the furniture and put the result of the bash on the z-level below? That might be a nicer solution.

RenechCDDA: Bash the furniture, yes. There's *too many* functions to do that.

Figuring out if the results are placeable below probably requires a new function. I'll take a look at it sometime soon

anoobindisguise:
Woot, it's potassium lye's time to shine!

PR #66359, opened 2023-06-21
main: fix cmdline parsing

andrei8l:
#### Summary
None

#### Purpose of change
#66330 removed a cmdline argument but didn't decrease the container size so some c-style strings are zero-initialized and crash the game when accessed.

#### Describe the solution
Reduce container size
Change the relevant structs to use `std::string_view` instead of C-style strings, so even if this happens again it won't crash the game.

N/A

#### Testing
`./cataclysm-tiles --datadir $(pwd)/data/` doesn't crash anymore
`./cataclysm-tiles --datadir` still halts on missing parameter
`./cataclysm-tiles --help` still displays the same thing

#### Additional context
N/A

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -163,10 +163,10 @@ struct arg_handler {
     //! consumed by the call or -1 to indicate that a required argument was missing.
     using handler_method = std::function<int ( int, const char ** )>;
 
-    const char *flag;  //!< The commandline parameter to handle (e.g., "--seed").
-    const char *param_documentation;  //!< Human readable description of this arguments parameter.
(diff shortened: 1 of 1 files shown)


irwiss:
Welp, seems i assumed it's a vector rather than std::array and forgot to scroll up to check, thanks for fixing

andrei8l:
> On second thought, what's the benefit of std::array here? thinking

`std::array` can be allocated on the stack (or not at all), but that's an insignificant distinction here. I've changed it to vector; thanks.

NetSysFire:
I wonder if we could have an extremely simple test checking if `--help` is displaying as expected.

PR #66360, opened 2023-06-21
Revert "Use unordered_set to deduplicate active items (#66310)"

SurFlurer:
This reverts commit 49b2acbda66ee722df380ff7a54af3864e70c0f3.

#### Summary
none

#### Purpose of change

#66310 is bug-prone, and introduced instability, which led to those Windows CI tests crash.

See discussion on that thread for details.

#### Describe the solution

Revert it.

Patch it. Eventually there should be some sort of cache, but it turns out that properly caching is pretty hard. Let's stabilize the game first.

### DIFF
--- a/src/active_item_cache.cpp
+++ b/src/active_item_cache.cpp
@@ -6,6 +6,19 @@
 #include "item.h"
 #include "safe_reference.h"
 
+namespace
+{
+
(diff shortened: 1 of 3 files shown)


SunkenLotus:
Sorry, not used to Github. Still getting the segmentation fault when hauling after the merge. Also crash on loading.

PR #65822, opened 2023-05-23
CMake install() changes, static linking build, mark directories with fs::path{}

alef:
#### Summary
None

#### Purpose of change
Changes how CMake and some code handle paths.
Edit: Not fixing problems here. It is more about expanding what configurations can be build through CMake definitions: extended explanations on task list below.

#### Describe the solution
- [x] Use CMake's TYPE install paths. It could be useful in future for creating packages or installers
- [x] Build static without VCPKG on Ubuntu (see notes below)
- [x] Use `pkg-config` where possible. This allows to discover static library files in Ubuntu which are not by packages' own `.cmake` config files
- [x] C++17: Change some path handling functions. Try avoid handling post '/' '\\\\' manually by just appending `fs::path{}`
- [x] Maybe fix #64619 ?

#### Describe alternatives you've considered
A non-dynamic executable cannot really be built as in VCPKG, because not all requirements are there. It reduces anyway the dynamic dependencies.

#### Testing
- [x] Built and run on VS 2019 and VS 2022
- [x] Built via vscode on WSL2/Docker Ubuntu image
Can't get SDL/TILES to run via WSL2's pulseaudio/CygwinX/Wayland transports.

#### Additional context
Will start with draft PR, responding to C/I issues

PR #66369, opened 2023-06-21
Improve CMake behavior on Unix and macOS

perryprog:
#### Summary
None

#### Purpose of change
I noticed a few build and install issues popping up after #65822 for me on macOS. Namely:
1. I was getting a linker error on compilation due to CoreFoundation being used but not linked in CMake
2. Installation locations were hecked up when using `-DUSE_PREFIX_DATA_DIR=Yes` because... okay, I'm not sure why but stuff would be installed to `/usr/local` instead of within `/usr/local/share/cataclysm-dda` as you might expect.

#### Describe the solution
1. Add CoreFoundation as a framework when building on macOS.
2. Set `CMAKE_INSTALL_DATAROOTDIR` to include the `/share/cataclysm-dda` suffix when building with `USE_PREFIX_DATA_DIR`. I'm uncertain if this is the ideal solution—@alef, do you know?
    - I also wonder if USE_PREFIX_DATA_DIR should just be on by default on Unix since I'm _pretty_ sure this is what most Unix people would want, but I'm content to leave it as is for now.

#### Testing
Stuff builds and installs in the right location. I'm hoping CI will catch any other obvious issues...


perryprog:
Going to try making `USE_PREFIX_DATA_DIR` the default to see if that makes CI happy. (And because it's probably better for Unix users.)

perryprog:
I'm confident enough that this should be good to merge despite the flaked CI failure stopping the CMake job from actually running.

andrei8l:
> I'm confident enough that this should be good to merge despite the flaked CI failure stopping the CMake job from actually running.

The CMake job is already broken from the last CMake PR. Let's make certain it works this time

perryprog:
> The CMake job is already broken from the last CMake PR. Let's make certain it works this time

Yeaaaaah, you're right. My impatience was just getting to me. 🙃

PR #66371, opened 2023-06-21
Increase default mountable weight ratio

randomtyper:
#### Summary
Balance "Increases default mountable weight ratio"

#### Purpose of change

Default mountable_weight_ratio at 20% is very low and represents "the safe maximum" that was relevant in pre-apoc world, but the good times are already past and I doubt the survivor would have much concern for recommendations and safety when zombies are near.

#### Describe the solution

Increases the default mountable_weight_ratio to 0.35.

Leaving stuff as is.

#### Testing

None for now.

#### Additional context

Sources:

https://www.helpfulhorsehints.com/how-much-weight-can-a-pony-carry/

Quote: "Many modern studies have indicated that any load heavier than 30% of the animal’s body weight is likely detrimental in the long run". Survivor has totally different priorities than pre-Cataclysm horse breeders/riders/racers, mind you, most likely caring more about his survivor even at the expense of wellbeing of his mount.

https://www.napoleon-series.org/military-info/organization/France/Artillery/c_artilleryhorses.html

Quote: "...a horse carrying 75.6 kg  and drawing 315 kg  could travel on average 20 miles a day". The artillery wheel horses (ones with the most load) were at least 15.2 hands high, and weigh about 905 to 1,215 pounds (410.5 – 551 kg) at the listed height. Horses can draw about 1.5 times their weight, and for horse weight of 500 kg the pull weight is about 750 kg. That means 75 kg of carried weight is about equal to 300 kg of draft for a horse of 500 kg -> horse effectively carries about 150 kg of weight, giving us the 0.3 load-to-weight ratio.

Then there is this: https://en.wikipedia.org/wiki/Common_ostrich#Racing

Common ostrich weighs up to 140 kg, but usually less. That gives more than 50% load if a 70 kg human rides a large one, racing into 70% if ostrich is not of the largest variety. The race happens over somewhat short distance, though.

### DIFF
--- a/src/mtype.h
+++ b/src/mtype.h
@@ -399,8 +399,8 @@ struct mtype {
 
         // Number of hitpoints regenerated per turn.
         int regenerates = 0;
-        // mountable ratio for rider weight vs. mount weight, default 0.2
-        float mountable_weight_ratio = 0.2f;
+        // mountable ratio for rider weight vs. mount weight, default 0.35
+        float mountable_weight_ratio = 0.35f;
 
         int attack_cost = 100;  /** moves per regular attack */
         int melee_skill = 0;    /** melee hit skill, 20 is superhuman hitting abilities */


RenechCDDA:
Most animals cannot carry 35% of their bodyweight for any significant period of time. 20% was arrived at after some discussion in the discord https://discord.com/channels/598523535169945603/598607195021508638/636426632151695360

Or in regards to why it *might be possible* for animals to carry more for short periods, but the game won't let you... "The thing is, we don't have a system for handling long term riding vs short term riding, or injury based on overloading your horse" https://discord.com/channels/598523535169945603/598607195021508638/636422220230557696

If there are specific animals that should have >20%, they can have their weight ratio modified rather than modifying *everything's*.

randomtyper:
> If there are specific animals that should have >20%, they can have their weight ratio modified rather than modifying _everything's_.

Aight, If the 0.2 is default and everything else from that are JSON overrides, I can work with that. I would say the horses could have their weight ratio expanded to 25%, though, per the article on artillery horses I linked.

PR #66373, opened 2023-06-22
Add the Hi-Point 3895, a .380 ACP PCC

anoobindisguise:
#### Summary
Content "add .380 ACP carbine"

#### Purpose of change
I want more PCCs. Bigger damage number give me the good brain chemical.

#### Describe the solution
Adds the aforementioned gun. It is fairly rare but shows up anywhere standard guns can be found. Its magazine is not compatible with the Hi-Point CF380 pistol magazine. It gets +2 damage, this is per the muzzle velocity of a 95g bullet found in https://rangehot.com/380-auto-pistol-vs-carbine/ compared to that of a standard round. Other stuff is mostly based on manufacturer's specs. Volume of the gun is a guess based on scaling up a regular hi-point CF380 pistol to the new length/weight, which means it might be wrong. Dispersion and range bonus are based on the Cx4 Storm (9mm has 14 range and Cx4 gives +20 range. 380 has 13 range so I gave the 3895 +17 range. Regular 9mm has 480 dispersion but the Cx4 has 180, so since regular 380 also has 480 dispersion I gave the 3895 180 dispersion)

Give it +3 damage. It's pretty close and would round up but my assumption is that we floor all fractional bullet damages.

#### Testing
Loaded up in game and took some screenies. No errors.

#### Additional context

3895mag


data/json/items/gun/380.json
```diff
@@ -171,5 +171,55 @@
...
+    "min_cycle_recoil": 270,
+    "ranged_damage": { "damage_type": "bullet", "amount": 2 },
+    "valid_mod_locations": [
+      [ "barrel", 1 ],
+      [ "bore", 1 ],
+      [ "brass catcher", 1 ],
+      [ "grip", 1 ],
+      [ "mechanism", 4 ],
+      [ "muzzle", 1 ],
+      [ "rail", 1 ],
```
TheShadowFerret, line 214:
```suggestion
      [ "rail mount", 1 ],
```
The 3895 has bottom and top rails, but not side ones.

anoobindisguise: thanks, I just sorta copypasted mod options. I'll definitely be doing more of these.

PR #66380, opened 2023-06-22
Cast iron

Karol1223:
#### Summary
Content "Actual cast iron resources"

#### Purpose of change
Cast iron stuff gave steel chunks and... wasn't even called cast iron...

#### Describe the solution
- Changed material name from ``Iron`` to ``Cast iron`` and changed its density appropriately
- Added scrap, chunk and lump items for cast iron
- Brought up the disassembly recipe for cast iron pot and pan up to speed with tool qualities and skills used
- Cast iron pot and pan as well as the manhole cover will now give cast iron instead of steel when disassembled
- Added a disassembly recipe for the mechanical pump
- Edited iron ingots (and Magiclysm golems) to take density from cast iron and not regular iron


Maleclypse:
@Drew4484 i’m not sure if this is your wheelhouse but wanted to get your opinion in review, if you feel comfortable with it.

Drew4484:
If you're going to change Iron to Cast Iron instead of just adding Cast Iron as it's own material, you'll need to update ~150 items that currently use Iron. Otherwise I agree with the goal of this PR, cast iron is most common form of non-steel iron the average person is going to interact with.

Karol1223:
> If you're going to change Iron to Cast Iron instead of just adding Cast Iron as it's own material, you'll need to update ~150 items that currently use Iron

The problem is that according to what Kevin himself said in the devcord a few months back... that's already a thing. Our "Iron" was apparently never supposed to be Iron but Cast Iron. But nobody ever answered me when I asked why was it never renamed so I took matters into my own hands here.

Drew4484:
> > If you're going to change Iron to Cast Iron instead of just adding Cast Iron as it's own material, you'll need to update ~150 items that currently use Iron
> 
> The problem is that according to what Kevin himself said in the devcord a few months back... that's already a thing. Our "Iron" was apparently never supposed to be Iron but Cast Iron. But nobody ever answered me when I asked why was it never renamed so I took matters into my own hands here.

I say go for it then. There will be a few oddities but most armor that needs updating anyway.

PR #66383, opened 2023-06-22
3D vision for isometric tilesets

Rewryte:
#### Summary
Features "3D vision for isometric tilesets"

#### Purpose of change
In #65738, 3D vision was disabled for isometric tilesets because of several visual bugs, mostly involving height_3d.

#### Describe the solution
**Update**: This PR has been merged but the feature will not be fully functional until https://github.com/I-am-Erk/CDDA-Tilesets/pull/2036 is merged as well and the next tileset update takes place.

This fixes said issues and enables 3D vision for all tilesets.  The height_3d feature should now be fully compatible with 3D vision.  The apparent height of each z-level can be tweaked by changing the `zlevel_height` property in the tile_info.json / tile_config.json of the target tileset. Cross z-level map memory still does not work properly so I have disabled it for now.  That will be worked on next.
iso1
iso2

#### Testing
Builds ok.
Tested with various scenarios involving different elevations using both HitButton_iso and Ultica_iso tilesets.  Renders fine so far.
Map memory for the same z-level works as expected.
Updated compose.py built new tilesets successfully.

### DIFF
--- a/src/cata_tiles.cpp
+++ b/src/cata_tiles.cpp
@@ -598,6 +598,7 @@ void cata_tiles::set_draw_scale( int scale )
     cata_assert( tileset_ptr );
     tile_width = tileset_ptr->get_tile_width() * tileset_ptr->get_tile_pixelscale() * scale / 16;
     tile_height = tileset_ptr->get_tile_height() * tileset_ptr->get_tile_pixelscale() * scale / 16;
+    zlevel_height = tileset_ptr->get_zlevel_height();
 
     tile_ratiox = ( static_cast<float>( tile_width ) / static_cast<float>( fontwidth ) );
     tile_ratioy = ( static_cast<float>( tile_height ) / static_cast<float>( fontheight ) );
@@ -660,6 +661,7 @@ void tileset_cache::loader::load( const std::string &tileset_id, const bool prec
     for( const JsonObject curr_info : config.get_array( "tile_info" ) ) {
         ts.tile_height = curr_info.get_int( "height" );
         ts.tile_width = curr_info.get_int( "width" );
+        ts.zlevel_height = curr_info.get_int( "zlevel_height", 0 );
         ts.tile_isometric = curr_info.get_bool( "iso", false );
         ts.tile_pixelscale = curr_info.get_float( "pixelscale", 1.0f );
         ts.prevent_occlusion_min_dist = curr_info.get_float( "retract_dist_min", -1.0f );
@@ -1611,7 +1613,6 @@ void cata_tiles::draw( const point &dest, const tripoint &center, int width, int
         }
     };
 
-    const int height_3d_mult = is_isometric() ? 10 : 0;
     if( max_draw_depth <= 0 ) {
         // Legacy draw mode
         for( int row = min_row; row < max_row; row ++ ) {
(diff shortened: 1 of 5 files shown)


vetall812:
Can't wait for it!

Zupercharged:
From the screenshots it doesn't render the wall tiles on the layer below, would it be possible to modify it slightly to render the walls beneath ledges as well?

src/cata_tiles.cpp
```diff
@@ -3827,7 +3838,8 @@ void cata_tiles::draw_zlevel_overlay( const tripoint &p, const lit_level ll, int
         fog_color = curses_color_to_SDL( c_dark_gray );
     }
     // Setting for fog transparency
-    fog_color.a = 100;
+    // On isometric tilesets, fog intensity scales with zlevel_height in tile_config.json
+    fog_color.a = is_isometric() ? 50 + zlevel_height / 2 : 100;
```
Qrox, line 3842: This should be capped at 255.

(Or changed to `255.0 - 255.0 * pow(155.0 / 255.0, (z_level_height / some_reference_value))`, if we want it to be mathematically accurate. It should be cached in that case to improve speed though.)

Qrox (commented):
Looking very nice!

Qrox:
> render the walls beneath ledges as well?

I think it can be achieved by making the unseen tile sprite have height if it represents a wall (so the unseen walls look like vertical shadow instead of horizontal shadow shown in the screenshot).

src/cata_tiles.cpp
Rewryte: Good catch.  Thanks!

vetall812:
This PR (right above this comment) will help with the fog.
I'd recommend to use it also for `unknown` tiles on the same level as player character. So it'll create consistent fog even in currently invisible areas that contains stuff from the levels below.

vetall812:
> We could improve it if someone comes up with a specialized algorithm to determine what should be drawn from an isometric perspective.

We can draw tiles that below current level if it touches any one of eight neighboring visible/remembered tiles. But i'm not a programmer, so it's just a guess.

Rewryte:
> is that due to the overlay rendering function rendering open air tiles as rectangles rather than rhombuses?

Yup.  Personally, I like how it gives fog a bit of volume especially with the smaller tiles of SmashButton, but the jaggedness is more pronounced in Ultica.

> This PR (right above this comment) will help with the fog.

Thanks, I'll check it out.

> We can draw tiles that below current level if it touches any one of eight neighboring visible/remembered tiles. But i'm not a programmer, so it's just a guess.

That would be performance unfriendly I'm afraid.  Other than fog, what we usually draw are a single z-level's worth of layers.  Drawing an additional level underneath could double draw time.  We would still have the same problem on buildings taller than 2 levels too.

Zupercharged:
are the fog sprites for open air trimmed to match the grid pattern of the tile? it looks like theyre just normal square sprites which creates the overlapping and jagged effect when aligned to an iso grid

Rewryte:
> you can assign specific sprites for ledges which could be used to do a similar effect more intentionally and cleanly

`tr_ledge` is only spawned in a small radius around the player so fog further away would not have the effect.  Might be able to achieve this with the multi-tile system though.

Rewryte:
Not sure if I should include the `t_open_air` and `tr_ledge` tile_config.json changes this time round.  Without it, the feature wouldn't be fully functional until the next tileset update.  With it, there will be merge conflicts every tileset update until this PR itself is merged.

bombasticSlacks:
> Not sure if I should include the t_open_air and tr_ledge tile_config.json changes this time round. Without it, the feature wouldn't be fully functional until the next tileset update. With it, there will be merge conflicts every tileset update until this PR itself is merged.

ahhh I was wondering about this.

Rewryte:
Yeah, I guess I should put a disclaimer on the original post.

bombasticSlacks:
Don't worry about it, I should have read the conversations before testing.

Either way, it's in now 😄

PR #66397, opened 2023-06-23
recruiting static NPCs makes them lose their static traits

Daved27hundred:
#### Summary
None

#### Purpose of change

The NPC traits "ignore sounds" and "will not bash" were not set to be invisible in the traits menu, meaning NPCs with them that could be recruited (Meteorologist mostly) would display them if you asked to know about them.

#### Describe the solution

~~added `player_display: false` to both traits~~ Maleclypse suggested I have the NPCs lose said traits once they're recruited.

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -8327,7 +8327,7 @@
     "name": { "str": "Ignores Sounds" },
     "points": 99,
     "valid": false,
-    "description": "NPC ignores sounds and does not investigate them.",
+    "description": "NPC ignores sounds and does not investigate them.  If you see this, it's is a bug.",
     "purifiable": false
   },
   {
@@ -8336,7 +8336,7 @@
     "name": { "str": "Will Not Bash" },
     "points": 99,
     "valid": false,
-    "description": "NPC will not bash to reach things.",
+    "description": "NPC will not bash to reach things.  If you see this, it's a bug.",
     "purifiable": false
   },
   {
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -8327,7 +8327,6 @@
     "name": { "str": "Ignores Sounds" },
     "points": 99,
     "valid": false,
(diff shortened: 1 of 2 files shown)


Maleclypse:
The correct fix for this would be to remove those traits when the character becomes a follower.   This would break things further by making follower NPCs have weird behaviour that is impossible to easily identify.  Those two traits break followers regularly.  Sorry :(

Maleclypse (changes requested):
Do not merge as is.

Daved27hundred:
Understood, what do you suggest I do instead?

Maleclypse:
```
"success": {
          "topic": "TALK_AGREE_FOLLOW",
          "effect": "follow",
          "opinion": { "trust": 1, "value": 1 },
          "npc_lose_trait": [ "RETURN_TO_START_POS", "NO_BASH", "IGNORE_SOUND" ]
        },
```
Update the meteorologist follow dialogue to include the npc_lose_trait line.  Same for any other characters that have any of  those three traits.  Make sure to only remove the traits they already have.

Daved27hundred:
got it - I'll start this tomorrow, it's quite late.

Daved27hundred:
image
@Maleclypse Is this where it should go? I'm not very familiar with NPC code.

Maleclypse:
> 

Ah ha.  So you're going to want to go to `TALK_COMMON_OTHER.json` and then any effect that includes `follow` your going to want to add `"npc_lose_trait": [ "RETURN_TO_START_POS", "NO_BASH", "IGNORE_SOUND" ]` apparently I didn't realize when I did the first one that the different reasons to persuade an npc to follow you used different talk topics.  Which I should have. Sorry

Daved27hundred:
Should I remove the `player_display: false` lines I added?

Maleclypse:
> Should I remove the `player_display: false` lines I added?

Yes please.  That way if other people add new follows we'll know something has gone wrong.

PR #66403, opened 2023-06-23
palettized canteen 1

PatrikLundell:
#### Summary

None

#### Purpose of change

Make use of palette parameters to cut down expansion JSON footprint.

Bycatches:
Addition of concrete as a construction material.
Palette cleanup.
Allowing items to remain when constructing on top of them to avoid error messages.

#### Describe the solution

Make use of the same logic as has been applied to other base/expansions, i.e. use palette parameters for constructions.
Pruned back palettes by removing unused entries.
Added concrete as a material (essentially free using the copy-paste-modify method).

The construction palettes contained invariant elements. Those were moved to the common palette, but could instead have been moved to a separate palette.

#### Testing

Built the complete expansion in the NW expansion out of concrete and then built it out of each of the other materials in other expansion slots (doesn't really do anything other than potentially test furniture/item collisions). The second one also constructed the alternative cooking furniture that wasn't used in the first one for completeness. After construction the building was inspected visually to see it looked OK, and then the game was terminated, after which it was restarted for the construction of a new expansion/material.

#### Additional context

The check for used palette entries resulted in the reordering of palette elements that might make it slightly harder to check differences (elements found to be used were moved to a new section at the beginning, sorting in alphabetic order, with those not matched deleted).

The expansion's metal version uses scrap metal rather than "real" metal. That remains unchanged.

It can be noted that this may be the first usage of a parameterized palette together with a fixed one. It works.


Maleclypse:
Thank you for doing this.  By doing this it makes future faction camp additions easier for people to make.  In case I hadn't mentioned it before.

PR #66407, opened 2023-06-23
Adding ranged damage enchantment and Old guns perk to bombastic perks

MorvarchPrincess:
#### Summary
Mods "Adds a new perk giving increased gun damage for pre WWII guns to bombastic perks. Also adds ranged damage enchantment"

#### Purpose of change

Figured a perk that incentivizes using the older guns would be fun, so added a perk that buffs damage of all WWII and before guns by 25%. Its a lifestyle perk, so one of the expensive ones. Also this adds a ranged damage enchantment to the code, so other places can buff ranged damage as well.

#### Describe the solution

The enchantment works by intercepting the projectile generated when firing a ranged weapon and applying the enchantment to its damage.

The perk works by adding an "OLD_GUN" flag to a whole bunch of guns, and then adding a perk that adds an enchantment of 25% ranged damage when wielding an OLD_GUN.

Naming guns individually in the perk was an option, but would be messy due to how many there are, so I went with a flag.

I also considered adding the flag to modern day versions of older guns, like modern revolvers or modern recreations of old guns, but decided that it made more sense to be actual antiques that get the bonus.

#### Testing

Manually tested with perk on and off. I confirmed that the M1911 got the damage bonus, and an M9A1 didn't. I also confirmed this was the case via breakpoints in code.

#### Additional context


data/json/flags.json
```diff
@@ -2223,5 +2223,10 @@
     "id": "REBREATHER_CART",
     "type": "json_flag",
     "info": "A filter for a rebreather."
+  },
+  {
+    "id": "OLD_GUN",
+    "type": "json_flag",
+    "info": "A flag for older guns, primarily used for an old guns perk for the perks mod right now."
```
bombasticSlacks, line 2230: This shows up in game, so should def be less meta

bombasticSlacks (commented):
All of the code you've added to the base game should be in the bombastic perks mod itself (flag and the additions to the guns).

Specifically a file called:

perkdata/olde_guns.json

and for the items with flags you should do exactly what the Tamable_Wildlife mod does: https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/mods/Tamable_Wildlife/reptile_amphibian.json#L21 

you just do "id" + "type" + "copy-from" + "name" and then add the "OLD_GUN" flag with extends like they do.

TheShadowFerret:
>I also considered adding the flag to modern day versions of older guns, like modern revolvers or modern recreations of old guns, but decided that it made more sense to be actual antiques that get the bonus.

Ok 🤔

Remove the flag from .700 Nitro Express, .40 S&W, 10mm, .38 Super, .44 Magnum, and .460 Rowland guns as those calibers didn't exist yet + LeMat revolver(it's a Pietta reprod)

Add the PPsh-41.

data/json/items/gun/45colt.json
```diff
@@ -96,5 +96,28 @@
     "proportional": { "reload": 1.5 },
     "pocket_data": [ { "pocket_type": "MAGAZINE", "rigid": true, "ammo_restriction": { "45colt": 6 } } ],
     "melee_damage": { "bash": 8 }
+  },
+  {
+    "id": "colt_saa_old",
```
TheShadowFerret, line 101: This would be a variant in main game, it should be in the mod's files.

MorvarchPrincess: I'm not quite sure I follow. I added to the main files as it seemed like it would be fine as a main game addition.

TheShadowFerret: Because in main game these two would have the exact same stats, they should have this on a single item's definition:
```
"variant_type": "gun",
    "variants": [
      {
        "id": "colt_saa_old",
        "name": { "str_sp": "Colt Single-Action Army Revolver" },
        "description": "This is an original antique of the legendary Colt Single Action Army, or Colt Peacemaker, one of the first revolvers to use a modern self-contained cartridge.  Made famous by Westerns, it is still in demand for Cowboy Action Shooting, reenactors, and collectors.  Unlike modern revolvers, the cylinder cannot swing out for loading, and spent brass must be ejected one at a time."
        "weight": 1
      },
      {
        "id": "uberti",
        "name": { "str": "Uberti Cattleman" },
        "description": "This 7.5\" barreled Uberti Cattleman is a modern reproduction of the legendary Colt Single Action Army, or Colt Peacemaker, one of the first revolvers to use a modern self-contained cartridge.  Made famous by Westerns, it is still in demand for Cowboy Action Shooting, reenactors, and collectors.  Unlike modern revolvers, the cylinder cannot swing out for loading, and spent brass must be ejected one at a time."
      }
    ],
    ```
    
    Flags can't change between variants, so having a version that has OLD_GUN on it requires two item definitions.

bombasticSlacks:
If you want to PR a historical Colt it would be better in a separate PR, it'll hold up reviewing this one (which aside from the styling issues is good to go).

alp7292:
shouldn't multiply, multiply the value so for a gun with 100 damage multiplied by 0.25 equals to 25

data/json/items/gun/45colt.json
MorvarchPrincess: Went ahead and removed it.

data/json/flags.json
MorvarchPrincess: Done

Paprika-H:
You'd also want to add the M2 carbine, M1911A1, Colt Army, 1887 bootleg shotgun, and maybe all of the Exodii rifles, as the designs, at least by current earth standards they would be pretty old.

PR #66409, opened 2023-06-23
Mapgen weight as a function of time since cata

Procyonae:
#### Summary
None

#### Purpose of change
Just putting this here but I don't get what I'm doing so this probably won't get done at least soon.
(Just astlying it at least then I'll close)

Aims to add the ability to use a weight function from a json defined list alongside the weight of a map to allow the weight to change at runtime dependant on variables (just time since cata to begin with), allowing different maps to only spawn after a certain time or to stop spawning after a certain time or less simplistic functions like making the map gradually decrease in weight to a minimum quadratically. Could be used in initial implementation to say have "untouched" variants of buildings decline over time, heavily looted/destroyed variants to increase over time and partially looted/destroyed variants to increase then decrease. Could also be used in the future to include other values to say increase the likelihood of running into the faction you've helped the most.

#### Describe the solution
What solution?

Sticking to json stuff

### DIFF
--- a/src/mapgen.cpp
+++ b/src/mapgen.cpp
@@ -109,21 +109,6 @@ static const itype_id itype_diesel( "diesel" );
 static const itype_id itype_gasoline( "gasoline" );
 static const itype_id itype_jp8( "jp8" );
 
-static const jmath_func_id jmath_func_weight_start("weight_start"); // I think there must be a way to just parse this
-static const jmath_func_id jmath_func_weight_end("weight_end");
-static const jmath_func_id jmath_func_weight_start_min("weight_start_min");
-static const jmath_func_id jmath_func_weight_end_min("weight_end_min");
-static const jmath_func_id jmath_func_weight_linear_pos("weight_linear_pos");
-static const jmath_func_id jmath_func_weight_linear_neg("weight_linear_neg");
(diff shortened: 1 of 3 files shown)


src/mapgen.cpp
```diff
@@ -813,8 +822,8 @@ mapgen_function_json_base::mapgen_function_json_base(
 
 mapgen_function_json_base::~mapgen_function_json_base() = default;
 
-mapgen_function_json::mapgen_function_json( const JsonObject &jsobj, const int w,
-        const std::string &context, const point &grid_offset, const point &grid_total )
+mapgen_function_json::mapgen_function_json( const JsonObject &jsobj, const dbl_or_var w,
+    const std::string& context, const point& grid_offset, const point& grid_total, const JsonObject &weightfunc )
```
Procyonae, line 826:
```suggestion
        const std::string &context, const point &grid_offset, const point &grid_total )
```

Procyonae:
> Nice PR! Why keep close/reopening?

I were just trying to style it before I closed in case anyone wants to pick it up but kept missing bits >_>

Procyonae:
Tried again here but I'm not really getting anywhere https://github.com/Procyonae/Cataclysm-DDA/tree/MapWeightFuncs

PR #66414, opened 2023-06-23
JSON formatter html build

irwiss:
#### Summary
None

#### Purpose of change

Toyed with emscripten a bit - made a non-cgi formatter for one-file-at-a-time json formatter flow

https://irwiss.github.io/Cataclysm-DDA/formatter.html

#### Describe the solution

Compiles the formatter with emscripten, sprinkles some html/js and the static page can lint in the format cata wants (example above)

It's a bit bigger (~120kB gzipped vs a few kB) than the narc.ro but I think it's worth the tradeoffs: it doesn't hit a cgi server with a large json trout and client side can be snappier if network isn't great. I didn't go into polyfills and the rest of js shenanigans, likely stuff requires es6 - but testing on up to date chromium/firefox it works including mobile

Also uses browser's JSON.parse to early detect some errors cata formatter barfs on (e.g. type in 12345 on narc.ro)

The automated PR maker job needs a personal access token in TX_PR_CREATOR_JSON_LINTER, fine grained worked for me with permissions for "Content" read/write, "Metadata" read-only, "Pull Requests" read/write, not sure if there's a better way to auto PR to gh-pages, stole the action from weekly changelog, although weirdly it starts unshallowing and churning GBs of deltas for just 1 file PR...

#### Testing

The PR creation likely won't work until someone with permissions adds a key and reruns actions, not 100% solid on how it works, although considering the amount of hassle with this action  just manually PRing to gh-pages starts to sound more and more attractive

### DIFF
--- a/.github/workflows/format_emscripten.yml
+++ b/.github/workflows/format_emscripten.yml
@@ -1,0 +1,63 @@
+name: JSON Formatter in HTML
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+    - 'tools/format/**'
+    - 'tools/format_emscripten/**'
+    - '.github/workflows/format_emscripten.yml'
+
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: checkout repository
+      uses: actions/checkout@v3
+
+    - name: install dependencies (ubuntu)
+      run: |
+          sudo apt-get update
+          sudo apt-get install emscripten
+
+    - name: Build with emcc (emscripten)
+      run: tools/format_emscripten/build.sh
+
+    - name: Display files
+      run: ls -Rah
+
+    - name: Upload zipped html as artifact
+      uses: actions/upload-artifact@v3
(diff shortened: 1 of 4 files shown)


build-scripts/formatter-emscripten/shell.html
```diff
@@ -0,0 +1,155 @@
...
+                json_error_alert.style.display = "block";
+                json_output.value =
+                    "Cataclysm JSON must start with array or object";
+            };
+        });
+
+    </script>
+</body>
+
+</html>
```
ZhilkinSerg, line 155:
```suggestion
</html>
```

irwiss: Seems vscode default html formatter trims it regardless of .editorconfig
I added it manually as github ui doesn't let applying [whitespace changes](https://cdn.discordapp.com/attachments/861650119899873321/1122115925706031145/image.png)

I had to move files anyway - changes in build-scripts seem to trigger a full clang-tidy check

PR #66418, opened 2023-06-23
Reduce hangtime of shocker sparks

worm-girl:
#### Summary
Balance "Reduce hangtime of shocker sparks"

#### Purpose of change

Shockers are ubiquitous in the world and basically force survivors to wear activity suits, get the capacitance CBM, or perish. While gear checks are well and good, this one is extremely punishing, partially because the sparks hang in the air for such a long time, resulting in the player being struck many times by a single bolt of lightning that can somehow hang around for several seconds.

Real-life electrical arcs do not hover in midair like a gas. They jump instantaneously from one surface toward another and then blip out of existence. This isn't completely possible with how we've modeled electricity in this game, but it can be better.

#### Describe the solution

This PR reduces the half-life of both types of spark fields from 4 to 1. The result is that it is just as easy to get hit by a shocker's initial attack or the odd stray bolt, but you are less likely to get spammed with hits from the same source unless you decide to just stand there and take it. Even if you do, it will result in fewer overall hits.

The spark field is partially hardcoded and I considered messing with that, but it may not be necessary for a small nerf like this. A more comprehensive fix might be to make it so that if a field spreads to a tile which already contains a field, it should not refresh that tile's lifespan.

#### Testing

I made a test character and spawned a shocker, an incandescent husk, and a shocker brute, and made sure their attacks all looked and acted correctly. I observed that the lightning looks a bit more like lightning now, in a bolt with forks, rather than a big blue and white cloud of death.

### DIFF
--- a/data/json/field_type.json
+++ b/data/json/field_type.json
@@ -808,7 +808,7 @@
     "description_affix": "illuminated_by",
     "has_elec": true,
     "priority": 4,
-    "half_life": "4 turns",
+    "half_life": "1 turn",
     "phase": "plasma",
     "display_items": false,
     "display_field": true
@@ -1625,7 +1625,7 @@
     "description_affix": "illuminated_by",
     "has_elec": true,
     "priority": 4,
-    "half_life": "4 turns",
+    "half_life": "1 turn",
     "phase": "plasma",
     "display_items": false,
     "display_field": false,


ashGlaw (approved):
Great change. This makes more sense.

Procyonae:
Just checking but they always still at least hit you at least once (if they don't miss) before expiring right?

fairyarmadillo:
> Just checking but they always still at least hit you at least once (if they don't miss) before expiring right?

Yes. There is no circumstance where sparks enter your tile but do not damage you, short of you wearing faraday armor. Lightning bolts last about 3-4 seconds overall and can still hit you multiple times if you were cleanly hit and don't (or can't) move, but if the shocker misses or you were only partially hit, you can move out of the danger zone easier than before.

Shockers are still very dangerous in melee or if you are being held in place by another monster, but you're now less likely to get randomly brought to half health on all limbs by a single attack from a monster on the first turn you see each other. You'll usually only lose a pip of health.

PR #66429, opened 2023-06-24
Counter gate move cost

### LINKED ISSUE #66423: An open counter gate has the same move cost as a closed counter gate
### Describe the bug

When moving through a counter gate, the character automatically opens it.
This spends extra time, but an opened counter gate has the same move cost as a closed one (200), or a counter without gate. This makes going through a gate slower than going over a regular counter.

### Attach save file

N/A

### Steps to reproduce

1. Spawn a closed counter gate and a regular counter
2. Go through a closed counter gate
3. Character spends 100 moves opening the gate and 161 going through the open gate
4. Go through a regular counter—spend only 161

### Expected behavior

After the character spends time opening a counter gate, going through it should not take the same time as going through a closed one.

### Screenshots

screenshot_2023-06-24_11-11-47_064698193
screenshot_2023-06-24_11-12-56_430014190

- OS: Linux
- Game Version: 8776427 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm]
]

### Additional context

_No response_

Procyonae:
#### Summary
Bugfixes "Open counter gates are faster to navigate than counters"

#### Purpose of change

Partially addresses #66423

#### Describe the solution

Open counter gate now has no move cost penalty over the base terrain

Making the closed gate -10 mod instead but them it says "Impassable" which doesn't make much sense so I just moved it in line with the open gate

Making it 1 move cost mod instead but as it is now with opening taking 100 it still takes about the same to open and move through the gate as going over the counter so 0 move cost mod seems to make more sense

Giving counter gates the ability to be examine climbed for the same movecost as navigating the regular counter but the only climb action costs 400 and I don't fancy making a c++ change for something so minor

I looked at counters and cupboards because they have kind of odd similarities/differences but I couldn't decide what to do with them so I left them as is


data/json/furniture_and_terrain/furniture-surfaces.json
```diff
@@ -66,7 +66,7 @@
     "description": "A commercial-quality swinging door made of wood that allows passage behind counters.  It's currently closed.",
     "symbol": "+",
     "color": "blue",
-    "move_cost_mod": 2,
+    "move_cost_mod": 0,
```
Demivan, line 69:
```suggestion
    "move_cost_mod": 2,
```

Closed counter door should still have a movement penalty

Demivan (changes requested):

Demivan:
It is a good addition, but it does not fully solve https://github.com/CleverRaven/Cataclysm-DDA/issues/66423. I think it is better to keep the issue open.

data/json/furniture_and_terrain/furniture-surfaces.json
Procyonae: I'm not making it crossable so that's irrelevant

Demivan: Yes, but tooltip is still shown. Could be confusing if it is different from a regular counter

Procyonae: Why would it be confusing?

Demivan: When you inspect it with `X`, it would say that movement cost is 100, like it is not there.

So if there is no difference between open and closed gate, why open it then :)

PR #66431, opened 2023-06-24
Check map bounds before checking Character::sees()

### LINKED ISSUE #66427: Game crashes when trying to start a run in a particular world.
### Describe the bug

Generated a new world and tried to start a game with a new character. Game crashed three times when I tried to start the Evacuee scenario. Other worlds give no such error.

### Attach save file

N/A

### Steps to reproduce

1. Generate a world.
2. Create a character.
3. Try to start the world in the Evacuee scenario
4. Game crashes immediately without loading in to the world.

### Expected behavior

For the game to start after character creation is complete.

### Screenshots

_No response_

- OS: Windows
- Game Version: 8776427 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Generic Guns [generic_guns],
]

### Additional context

[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/11857215/debug.log)
[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/11857216/crash.log)

irwiss:
#### Summary

None

#### Purpose of change

Fixes https://github.com/CleverRaven/Cataclysm-DDA/issues/66427

#### Describe the solution

Makes pl_sees overload on tinymap/fake_map return false without accessing lightmap

Doesn't solve the issue completely - the out of bounds read can still be reached just not from this chunk of code, but this will stop the random crash.

#### Testing

Mapgen is triggering it randomly, spawning default evacuee in shelter a few times triggers this bug on MSVC

### DIFF
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -1709,7 +1709,7 @@ bool map::furn_set( const tripoint &p, const furn_id &new_furniture, const bool 
     invalidate_max_populated_zlev( p.z );
 
     set_memory_seen_cache_dirty( p );
-    if( player_character.sees( p ) ) {
+    if( pl_sees( p, player_character.sight_max ) ) {
         player_character.memorize_clear_decoration( getabs( p ), "f_" );
     }
 
@@ -2153,7 +2153,7 @@ bool map::ter_set( const tripoint &p, const ter_id &new_terrain, bool avoid_crea
 
     set_memory_seen_cache_dirty( p );
     avatar &player_character = get_avatar();
-    if( player_character.sees( p ) ) {
+    if( pl_sees( p, player_character.sight_max ) ) {
         player_character.memorize_clear_decoration( getabs( p ), "t_" );
     }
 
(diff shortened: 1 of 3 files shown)


src/map.cpp
```diff
@@ -1709,7 +1709,7 @@ bool map::furn_set( const tripoint &p, const furn_id &new_furniture, const bool
     invalidate_max_populated_zlev( p.z );
 
     set_memory_seen_cache_dirty( p );
-    if( player_character.sees( p ) ) {
+    if( inbounds( player_character.pos_bub() ) && player_character.sees( p ) ) {
```
andrei8l, line 1712: I don't think this is correct. `pos_bub()` returns position relative to the main map, but the map running this function at mapgen is just a temporary tinymap. So the coordinate origins don't necessarily line up.

`player_character.sees( p )` also only makes sense if this map is the main map. In both cases the tripoint needs to be converted to absolute from this map, then to local coordinates for `get_map()`

irwiss: I agree this isn't technically "correct", but I don't see an obvious way to call sees() with the bounds check but without making every non-mapgen caller pay the perf cost for it; Character::sees() calls into get_map().pl_sees() that reads out of bounds at line 799 during mapgen because the avatar isn't yet initialized
https://github.com/CleverRaven/Cataclysm-DDA/blob/d505f22d68e4ea1f0bb5630de74365c06f241900/src/lightmap.cpp#L792-L799

pos_bub will return relative to main map but that is what that function is checking

When ter_set/furn_set is called during mapgen the avatar position isn't in bounds of any map including the main one (it's hardcoded 20,10,-500 in Creature ctor, and at some point before world loads changes to -4768,-4634,-500)

Another option is a dynamic_cast/typeid hack, since the map/tinymap inheritance is somehow flipped around

Or just paying the price of a bounds check inside, but have it useless for most calls after game is past loading

Do you or anyone else have a better solution or opinion on one?

Issue #65162, opened 2023-04-18
Json errors upon loading character. From #64922 likely


Rimplestilt:
### Describe the bug

Issue began in 2023-04-09-0816 for me. Likely #64922 specifically. All errors seem to be in flexbuffer_json.cpp

So many errors. A few of the What feels like 100 messages i have to get through each time i load now.  May just be an issue on my end?

![Screenshot_20230418-090643_Cataclysm DDA (experimental)](https://user-images.githubusercontent.com/120688336/232787138-e23d82b3-d453-4dc1-b0f8-0674049c7b58.jpg)
![Screenshot_20230418-090707_Cataclysm DDA (experimental)](https://user-images.githubusercontent.com/120688336/232787158-32c84755-5989-4150-a0d1-66d76f09f060.jpg)
![Screenshot_20230418-090713_Cataclysm DDA (experimental)](https://user-images.githubusercontent.com/120688336/232787188-0bc79dda-0123-4e7e-9ac5-811e5783f36e.jpg)

### Attach save file

N/a? Happens to any character i play or create

### Steps to reproduce

start game

### Expected behavior

No errors

### Versions and configuration

- OS: Android
    - OS Version: Manufacturer: samsung; Model: SM-S136DL; Release: 12; Incremental: S136DLUDU2BVE8; 
- Game Version: 060fe4f [32-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Stats Through Kills [stats_through_kills],
    Bombastic Perks [bombastic_perks],
    Magiclysm [magiclysm]
]

Theawesomeboophis:
/confirmed.

Theawesomeboophis:
Confirmed, doesn't occur on windows which is strange.

firegaul2077:
yeah, I compiled this on a rpi4 Ubuntu 22.04 64bit with tiles and it worked fine...  but android version keeps popping the same issue.

Empyrean-Heaven:
Still a problem, more than a month later. ㅠㅠ

mqrause:
Is there some other, different error reported before all these? There was a very similar issue before which supposedly was fixed by #65477. But that requires an actual error to trigger this kind of spam of not actual errors.

[...]

PR #66438, opened 2023-06-25
Pre-strip comment members before deciding whether to create TextJson

akrieger:
#### Summary
None

#### Purpose of change
Re-enabling skipped member checking was a good idea. It also exposed the fact that not erroring on comments worked... but only after the expensive part of the flow already happened. Creating the TextJsonIn and advancing it to the erroring location consumed 1/3 the cpu used in my test case of loading a save, a full 10+s. Lets, uh, fix that.

#### Describe the solution
After parsing the visited member bitmask into indices, remove indices that map to comment type members (start with `//`). Then don't call the expensive function if there are no indices left.

Not.

#### Testing
Perf test the same save. Profiler is a little bunk right now but, left is before right is after. `basic_istream` self cpu usage no longer top of functions.
image

 `JsonObject::report_unvisited` total cpu zeroes out after.
image

Introduce a new random member in an object and verified game still fires on it.
image

### DIFF
--- a/src/flexbuffer_json.cpp
+++ b/src/flexbuffer_json.cpp
@@ -279,7 +279,16 @@ void JsonObject::report_unvisited() const
             mask >>= 1;
         }
 
-        error_skipped_members( skipped_members );
+        // Don't error on skipped comments.
+        skipped_members.erase( std::remove_if( skipped_members.begin(),
+        skipped_members.end(), [this]( size_t idx ) {
+            flexbuffers::String name = keys_[idx].AsString();
+            return strncmp( "//", name.c_str(), 2 ) == 0;
+        } ), skipped_members.end() );
+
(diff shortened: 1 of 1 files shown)


akrieger:
The expensive call was `advance_jsin` on line 321/330.

akrieger:
clang-tidy!!!! *Shakes fist*

RenechCDDA:
Am I correct in understanding that this should only have a performance impact, not a functional impact?

Also given what's going on in this PR, can I bring your attention to #65162? A very few people across multiple OSes with nothing apparent in common report skipped member errors on perfectly valid json and a found workaround is... removing comments.

akrieger:
Yes, this should only be a performance change. I can't really explain what's happening in that issue, but I can investigate. Seems odd to only be affecting few people.

PR #66439, opened 2023-06-25
Made Slingshots more Tacticool

ashGlaw:
#### Summary
Content "Made Slingshots more Tacticool"

#### Purpose of change

Since the wrist rocket has a wrist brace, but isn't craftable, I wanted the slingshot you can craft to have an optional wrist brace, too. 

#### Describe the solution

I added the brace through the replacement stock mount and the tail hook stock. It improves the slingshot, but even paired with the underbarrel laser sight it's still weaker than the wrist rocket in every regard. 

I could make the wrist rocket craftable, but I think it would be difficult to make one as good as modern manufacture can. 

#### Testing

I spawned the mods and applied them, then compared their stats with unmodded slingshots. The mods went on like normal

#### Additional context

This started because I wanted to see a laser sight on a slingshot that actually helped it work better. I didn't realize that it was already possible, it just required the addition of a bottom mount. Now that I understand that system it's what I used to expand access to the tail hook stock. I also removed the grip mount mod slot from the slingshot and wrist rocket. It seemed unnecessary, as the ergonomic grip does literally nothing for slingshots but make them heavier. I'd hoped it would do something like reduce the stamina cost of shooting, but no dice yet. Maybe I can add that in eventually.

### DIFF
--- a/data/json/items/gunmod/mount.json
+++ b/data/json/items/gunmod/mount.json
@@ -109,7 +109,7 @@
     "symbol": ":",
     "color": "light_gray",
     "location": "stock mount",
-    "mod_targets": [ "smg", "rifle", "pistol", "shotgun", "crossbow", "launcher" ],
+    "mod_targets": [ "smg", "rifle", "pistol", "shotgun", "slingshot", "crossbow", "launcher" ],
     "install_time": "20 m",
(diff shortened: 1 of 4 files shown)


ashGlaw:
Posted from the wrong place. Trying again

PR #66446, opened 2023-06-25
Autostyle on save for VSC

bombasticSlacks:
#### Summary
None

#### Purpose of change
Something I've wanted for a while, autostyling in VSC.

#### Describe the solution
Irwiss actually set up a plugin for VSC that does this, so I just added it to the recommended plugins.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -34,7 +34,12 @@
             {
                 "match": "\\.json$",
                 "isAsync": true,
-                "cmd": "${workspaceFolder}/tools/format/json_formatter ${file}"
+                "cmd": "${workspaceFolder}/tools/format/json_formatter ${file}",
+            },
+            {
+                "match": "\\.json$",
+                "isAsync": true,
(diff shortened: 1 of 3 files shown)


.vscode/settings.json
```diff
@@ -29,4 +29,13 @@
...
     // NOTE: This disable the plugins formatting so astyle is used.
     "C_Cpp.formatting": "disabled",
+    "emeraldwalk.runonsave": {
+        "commands": [
+            {
+                "match": "\\.json$",
+                "isAsync": true,
+                "cmd": "${workspaceFolder}/tools/format/json_formatter ${file}"
+            },
+        ]
```
irwiss, line 39: On linux the formatter gets .cgi extension, seems the extension silently executes them both so probably ok to have both
```suggestion
        "commands": [
            {
                "match": "\\.json$",
                "isAsync": true,
                "cmd": "${workspaceFolder}/tools/format/json_formatter ${file}",
            },
            {
                "match": "\\.json$",
                "isAsync": true,
                "cmd": "${workspaceFolder}/tools/format/json_formatter.cgi ${file}",
            },
        ]
```

PR #66449, opened 2023-06-25
EOC Run Until

bombasticSlacks:
#### Summary
Infrastructure "EOC run until"

#### Purpose of change
Wanted to be able to call an EOC multiple times without creating a massive call stack, seemed messy and inefficient.

#### Describe the solution
Adds new call eoc_run_until
you provide the name of a condition in the current context and an EOC and it runs it until the condition is false.

#### Testing
Unit test included.

### DIFF
--- a/src/npctalk.cpp
+++ b/src/npctalk.cpp
@@ -4136,14 +4136,34 @@ void talk_effect_fun_t::set_run_eoc_until( const JsonObject &jo, const std::stri
 
     str_or_var condition = get_str_or_var( jo.get_member( "condition" ), "condition" );
 
+    dbl_or_var itteration_count;
+    if( jo.has_member( "itteration_count" ) ) {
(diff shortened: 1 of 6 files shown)


src/npctalk.cpp
```diff
@@ -4130,6 +4130,25 @@ void talk_effect_fun_t::set_run_eocs( const JsonObject &jo, const std::string_vi
...
+    str_or_var condition = get_str_or_var( jo.get_member( "condition" ), "condition" );
+
+
+    function = [eoc, condition]( dialogue & d ) {
+        auto itt = d.get_conditionals().find( condition.evaluate( d ) );
+        if( itt == d.get_conditionals().end() ) {
+            debugmsg( string_format( "No condition with the name %s", condition.evaluate( d ) ) );
+            return;
+        }
+        while( itt->second( d ) ) {
```
andrei8l, line 4146: This needs a timeout or iteration limit as a safeguard. It can be an optional parameter to `run_eoc_until` with some sane default like 5 seconds.

PR #66454, opened 2023-06-25
New Worlds

bombasticSlacks:
#### Summary
None

#### Purpose of change
Let the character create new worlds and locations that don't just exist at -10 z level.

#### Describe the solution
Currently working through needed fixes. So far
* Swaps the stored location to have a prefix
* Prefix is accessible through EOC teleport

#### Testing
Very not done, as usual PRing it so that I don't forget about it.

### DIFF
--- a/data/json/example_eocs.json
+++ b/data/json/example_eocs.json
@@ -46,5 +46,20 @@
         "hide_failing": true
       }
     ]
+  },
+  {
+    "type": "effect_on_condition",
+    "id": "EOC_world_test",
+    "//": "shifts the character to a new world",
+    "global": true,
+    "effect": [
+      { "u_location_variable": { "u_val": "tele_test" }, "x_adjust": 1, "y_adjust": 1 },
+      {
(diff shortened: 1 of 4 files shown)


ashGlaw:
This is infrastructure that would allow the player character to be transported to another reality?

b3brodie:
To clarify, by new worlds is this an entirely new unexplored map? (How would factions, evolution, etc be handled?)  Or is a better way for creating more restricted dungeon maps like for the portal storm dungeons that no longer relies on just creating an area at a normally inaccessible z level?

bombasticSlacks:
Since there are now two, I want to make it clear I have no interest in answering any gameplay questions about this feature in this PR. When it is done, basically nothing will be possible aside from some unit tests proving it works, a follow-up PR will start to implement uses for this infrastructure, others could implement things in whichever way they wish as well. 

Github is not a place for QA on features, the general answer is: Anything would be possible with enough work, but that was true before I started on this too.

Issue #55094, opened 2022-02-04
Remove the ability to create character templates from existing characers


hexagonrecursion:
### Is your feature request related to a problem? Please describe.

It is currently possible to create a template based on an existing character. Unfortunately this feature is difficult to implement correctly and is prone to regressions.

1. Currently the character creation UI always creates a full character behind the scenes to store your choices and every template is also a full character. This leads to many bugs (TODO: add examples). #47462 has suggested a less bug-prone approach that is incompatible with this feature.
2. Any post-processing applied during normal chargen is also applied when you create a character based on a template based on an existing character causing bugs like #53777 and #51128

I am opening this thread to collect information on how and why this feature is used. My hope is that it can be replaced with something simpler and less bug-prone.

It looks like the feature does not do a good job of satisfying the original request that motivated it. #33716 asked to allow the player to move the character, items, vehicles and followers (i.e. to preserve progress), but templates only let you copy the character (and a small number of items).

@FONZACUS, could you please shed some light on why you wanted to "move region"? Is this for plot/roleplay reasons? It is for technical reasons?
> should help shave a ton of save data

Just how big is are your save files? Just how small is your hard drive?

> update map contents to newer builds

Moving in one direction until you hit freshly generated terrain does this too.

> up/downloading saved games.

Ok. Worlds can grow to a point where sending them over the internet is slow. Is this the only reason you wanted this? There might be an easier way to accomplish this.

@OzoneH3, do you create templates based on existing characters? For what purpose?

### Solution you would like.

1. [ ] Find out how and why this feature is used
2. [ ] Add replacements that better suit the discovered use-cases
3. [ ] Remove the ability to create character templates from existing characers

### Describe alternatives you have considered.

Keep things as is and squash the bugs as they creep in.

hexagonrecursion:
related #54960 #43217 #43141

anothersimulacrum:
Also (fixed) https://github.com/CleverRaven/Cataclysm-DDA/issues/46995

hexagonrecursion:
https://github.com/CleverRaven/Cataclysm-DDA/issues/47462 would fix #50687 too - no mana pool in the template if the template is not a full character

chrispikula:
[...]

PR #66473, opened 2023-06-26
Implement export&import of the protagonist and follower NPCs

hexagonrecursion:
#### Summary

Features "Implement export&import of the protagonist and follower NPCs"

#### Purpose of change

This is intended as a replacement for the "character to template" feature as described here #55094. Note: the "character to template" feature will **not** be removed until it is confirmed that this is a satisfactory replacement.

#### Describe the solution

Implement 3 new debug commands:

- Export a follower - exports an NPC from your faction
- Import a follower - spawns a new NPC based on a file and adds her or him to your faction
- Export self - a shortcut to export the protagonist if you don't have a follower to switch to

1. I did not add an "Import self" command as it appears to be redundant: the same effect can be achieved with "Import a follower" followed by "Control NPC follower"
2. Export&import of followers is not required to replace the "character to template" feature, but the extra features are useful to me personally (original motivation for this PR). I hope other players will find them useful too
3. I have considered not hiding this in the debug menu, but I can not imagine any way to replace the "character to template" feature that would not be some form of "cheating"

#### Testing

Tried exporting and importing characters - appears to work.

#### Additional context

Please reopen #55094. The button is unavailable to me. :(

This is still raw-ish. There are likely edge cases I have not considered. As we know, devs make poor testers. It would be great if someone who was not involved in writing this tried to break it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,6 +33,7 @@
 /src/prefix.h
 /sound/
 /templates/
+/export_dir/
 /tools/format/json_formatter.cgi
 CataclysmWin.cscope_file_list
 CataclysmWin.depend
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -166,12 +166,6 @@ void avatar::control_npc( npc &np, const bool debug )
         debugmsg( "control_npc() called on non-allied npc %s", np.name );
         return;
     }
-    if( !shadow_npc ) {
-        shadow_npc = std::make_unique<npc>();
-        shadow_npc->op_of_u.trust = 10;
-        shadow_npc->op_of_u.value = 10;
-        shadow_npc->set_attitude( NPCATT_FOLLOW );
-    }
     character_id new_character = np.getID();
     const std::function<void( npc & )> update_npc = [new_character]( npc & guy ) {
         guy.update_missions_target( get_avatar().getID(), new_character );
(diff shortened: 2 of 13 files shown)


Maleclypse:
Reopened

hexagonrecursion:
I am considering expanding this: 
- allow exporting any character - not just followers
- let the player choose npc faction when importing

RenechCDDA:
> * allow exporting any character - not just followers

Would the only determination for valid exports be 'they are of the character class'? I think that would be a valuable addition, but I wonder if there are cases where it would be difficult to maintain. (None of this is meant to discourage you, but since you posted about it I assume you to discuss it.)

For example, there's two characters I can think of that have (flavor-important!) EoCs attached. Would the EoCs migrate with them? Would it be *reasonably possible* to do that? (Is it a good idea for the EoCs to follow them? I don't know!) It's one thing to export followers, it's another thing to yoink any NPC out of the world.

Makayla Sanchez (the refugee center arsonist) has... some sort of EOCs which force certain mutant traits to appear on them during portal storms, or something. (It's not 100% clear to me *how this would ever be encountered in gameplay* but it seems very intentional! Going from static NPC --> follower ---> avatar seems like the intended route, and I can see ways this would break if they never became a follower. Link for reference.)
https://github.com/CleverRaven/Cataclysm-DDA/blob/1560777d9d2ad6a3ad5cc429692fb80680e31b1d/data/json/effect_on_condition.json#L408-L411

On the flipside, there's an ancilla bar merc which disappears after you talk to them, this is accomplished by forcing a mapgen update which removes any NPC(?) of that class. I can imagine scenarios where you exported the character previously, the class is inherited, and then mapgen comes along, sees your avatar is technically a member of the character class and you inherited the `BEM_anchor_seller` class, and then crazy stuff happens! (link for reference.)
https://github.com/CleverRaven/Cataclysm-DDA/blob/1560777d9d2ad6a3ad5cc429692fb80680e31b1d/data/json/npcs/robofac/robofac_ancilla_npcs/BAR_ENCOUNTER_MERCENARIES/BEM_anchor_seller.json#L43-L49

And then there's Apis, ~~who has a bunch of hardcoded exceptions~~ apparently is fully json nowadays, but could otherwise produce weird results.

As CDDA continues to take contributions I expect to see *more* cases where content involves bounds and bounds of EoCs and other data attached to characters that is hard to classify or automatically migrate without human intervention.

bombasticSlacks:
This looks good as is, since it's only debug it'll be fun to see what weird stuff most people find.

PR #66478, opened 2023-06-26
Adjust collapsible solar arrays

Alm999:
#### Summary
Balance "Adjust collapsible solar arrays"

#### Purpose of change

Currently, the collapsible solar arrays are nearly useless, as (for some reasons) they lose 60% of their performance compared to the base un-foldable version. This is:
a) not justified because unlike the reinforced type, there is no additional cover blocking sunlight, so the 60% power loss would be either from additional wiring (but  in the recipe we are adding only 2 additional wires, so a loss this huge is not feasible), or
b) physically cutting off 60% of the array's area (which is not true, as we are getting back the full array after disassembly).

I've done some (very basic)  research and found out that 50W collapsible arrays are in fact exist and not that big, like this  one:
https://aliexpress.ru/item/1005002748733505.html?sku_id=12000021954970351
It has dimensions of 10.5x16.5x1.5cm (in a folded state), close to what we have in the game.

#### Describe the solution

I've removed the 0.4 modifier for the collapsible solar arrays (vehicle parts) and increased items' mass a little bit (from 14 to 14.5 kg) to compensate for some added wiring, scrap and nuts and bolts.

If anything, the stationary solar arrays are vastly under-powered in the game. We could replace a roof-mounted solar arrays with something like this:
https://aliexpress.ru/item/1005003588693144.html?sku_id=12000026395151709
a 300W rigid panel with 21% conversion efficiency (a cheap monocrystalline model).
If the power output between stationary and vehicle-mounted panels is the idea (like with freezer and workbench volumes), we can introduce the "stationary solar array" (analogous to "stationary water purifier").
These "stationary solar arrays" could consist of 4…6 regular (vehicle-mounted) solar arrays we currently have.

#### Testing

The changes work in-game (both mass, and power output ones).

#### Additional context

Maybe it is a typo, but polycrystalline panels shall be "advanced" (as their layered structure enables conversion in a broad light spectrum, not in a narrow wavelength), not the other way around. See Scott Manley's video on the matter:
https://www.youtube.com/watch?v=VBIVlE0Kd1M

### DIFF
--- a/data/json/vehicleparts/vehicle_parts.json
+++ b/data/json/vehicleparts/vehicle_parts.json
@@ -1791,7 +1791,7 @@
     "name": { "str": "collapsible solar array" },
     "color": "light_blue",
     "broken_color": "light_gray",
-    "proportional": { "epower": 0.4, "durability": 0.5 },
+    "proportional": { "durability": 0.5 },
     "folded_volume": "2 L",
     "description": "A small array of solar panels designed to be able to fold into one another.",
     "item": "folding_solar_panel"
--- a/data/json/vehicleparts/vehicle_parts.json
+++ b/data/json/vehicleparts/vehicle_parts.json
@@ -1791,7 +1791,7 @@
     "name": { "str": "collapsible solar array" },
     "color": "light_blue",
     "broken_color": "light_gray",
-    "proportional": { "durability": 0.5 },
+    "proportional": { "epower": 0.68, "durability": 0.5 },
     "folded_volume": "2 L",
     "description": "A small array of solar panels designed to be able to fold into one another.",
     "item": "folding_solar_panel"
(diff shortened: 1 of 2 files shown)


ashGlaw:
I agree that the solar panels in the game are we compared to commercially available solar panels in real life. Maybe your idea of making the vehicle ones and stationary ones separate objects is a good one.  In Massachusetts about 10% of homes have solar panels all over their roof's southern exposure. It would be great to see that represented in the game.

data/json/vehicleparts/vehicle_parts.json
```diff
@@ -1791,7 +1791,7 @@
     "name": { "str": "collapsible solar array" },
     "color": "light_blue",
     "broken_color": "light_gray",
-    "proportional": { "epower": 0.4, "durability": 0.5 },
+    "proportional": { "durability": 0.5 },
```
Maleclypse, line 1794:
```suggestion
    "proportional": { "epower": 0.68, "durability": 0.5 },
```
Quick google shows collapsible panels average efficiency of 7%-15%.  Non foldable 15-22%.  I went ahead and improved the efficiency difference to 15/22.  68%.

Alm999:
Is there a reason epower of the "collapsible solar array" was changed back to 0.68?

PR #66482, opened 2023-06-26
Camps display required calories for crafting and building upgrades

### LINKED ISSUE #66144: "Not enough food stored" when trying to upgrade radio tower basecamp
### Describe the bug

When trying to do the wooden shack upgrade, despite having enough food the game tells me I do not have enough food stored for the task, but I supposedly have 22 days of calories while the building requires 4 days.

### Attach save file

[Fiskdale-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11725105/Fiskdale-trimmed.tar.gz)

### Steps to reproduce

1. Make a basecamp at a radio station, the version without a building next to the tower.
2. get resources for the wooden shack upgrade
3. get enough food to supposedly do the task
4. get a companion to do the task
5. error for not enough food pops up despite having enough food

### Expected behavior

Should be doable.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19044.2965 (21H2)
- Game Version: ec7537a [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm],
    No Hope [no_hope],
    Bionic Slots [cbm_slots],
    <color_cyan>Arcana and Magic Items</color> [Arcana],
    <color_cyan>Arcana</color>/Magiclysm Patchmod [Arcana_Magiclysm_Patch]
]

### Additional context

N/A

RenechCDDA:
#### Summary
Bugfixes "Camps properly display required calories for building and crafting"

#### Purpose of change
* Fixes #66144

#### Describe the solution
Since calories are no longer 2500 per workday but variable depending on exertion level, the display needs to reflect that. Now it does.

~~I'd like to colorize the calorie display depending on whether or not you have enough, but shoving even more conditions into basecamp::get_available_missions_by_dir seems unnecessary. Ideally it'd go into basecamp::om_upgrade_description but~~ ~~basecamp doesn't have access to time_to_food and dealing with all that seems very unnecessary (https://discord.com/channels/598523535169945603/598529174302490644/1122957912550146238)~~
Implemented that! 

#### Testing
<details>
 <summary>Screenshots of it at work</summary>

image
image
image
[increase stored calories to 65000]
image
image

</details>

#### Additional context
Known issues:
-*Crafting*(not building) display sometimes displays as 0, this is due to the use of `fake` activity levels for NPC-only recipes. (Yes, I will eventually make a *third* PR to resolve that)

-Calories may not actually be consumed when crafting. Unrelated to this PR, will be resolved separately.

~~-Display does not properly dim possible upgrades if you don't have enough calories. Will be resolved before undrafting.~~

### DIFF
--- a/src/faction_camp.cpp
+++ b/src/faction_camp.cpp
@@ -783,12 +783,16 @@ void basecamp::get_available_missions_by_dir( mission_data &mission_key, const p
             if( npc_list.empty() ) {
                 std::string display_name = name_display_of( miss_id );
                 const recipe &making = *recipe_id( miss_id.parameters );
+                const int foodcost = time_to_food( base_camps::to_workdays( time_duration::from_moves(
+                                                       making.blueprint_build_reqs().reqs_by_parameters.find( miss_id.mapgen_args )->second.time ) ),
+                                                   BRISK_EXERCISE );
+                bool can_upgrade = upgrade.avail;
                 entry = om_upgrade_description( upgrade.bldg, upgrade.args );
-                entry += string_format( _( "Total calorie cost: %s\n" ),
-                                        time_to_food( base_camps::to_workdays( time_duration::from_moves(
-                                                making.blueprint_build_reqs().reqs_by_parameters.find( miss_id.mapgen_args )->second.time ) ),
-                                                      BRISK_EXERCISE ) );
-                mission_key.add_start( miss_id, display_name, entry, upgrade.avail );
+                entry += string_format( _( "Total calorie cost: %s\n" ), foodcost );
+                if( foodcost > camp_food_supply( 0, false ) ) {
+                    can_upgrade = false;
+                }
+                mission_key.add_start( miss_id, display_name, entry, can_upgrade );
             } else {
                 entry = action_of( miss_id.id );
                 bool avail = update_time_left( entry, npc_list );
--- a/src/faction_camp.cpp
+++ b/src/faction_camp.cpp
@@ -791,13 +791,13 @@ void basecamp::get_available_missions_by_dir( mission_data &mission_key, const p
                 entry = om_upgrade_description( upgrade.bldg, upgrade.args );
                 if( foodcost > available_calories ) {
                     can_upgrade = false;
-                    entry += string_format( _( "Total calorie cost: " ) ) +
-                             string_format( colorize( std::to_string( foodcost ), c_red ) ) +
-                             string_format( _( "(have %s)" ), available_calories );
+                    entry += string_format( _( "Total calorie cost: %s (have %d)" ),
+                                            colorize( std::to_string( foodcost ), c_red ),
+                                            available_calories );
                 } else {
-                    entry += string_format( _( "Total calorie cost: " ) ) +
-                             string_format( colorize( std::to_string( foodcost ), c_green ) ) +
-                             string_format( _( "(have %s)" ), available_calories );
+                    entry += string_format( _( "Total calorie cost: %s (have %d)" ),
+                                            colorize( std::to_string( foodcost ), c_green ),
+                                            available_calories );
                 }
                 mission_key.add_start( miss_id, display_name, entry, can_upgrade );
             } else {
--- a/src/faction_camp.cpp
+++ b/src/faction_camp.cpp
@@ -782,7 +782,12 @@ void basecamp::get_available_missions_by_dir( mission_data &mission_key, const p
 
             if( npc_list.empty() ) {
                 std::string display_name = name_display_of( miss_id );
+                const recipe &making = *recipe_id( miss_id.parameters );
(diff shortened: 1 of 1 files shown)


RenechCDDA:
Now properly dimming.
image
image

PatrikLundell:
I'd say the calorie cost ought to follow the same format as other entries, i.e. color coded red (done) or green (not mentioned), and also show the calorie store like other entries show what's available.

e.g. Total calorie cost: 60000 (have 50000)

Also, calorie storage measured in (kilo)calories make technical sense, but it's a rather inconvenient measure since you quickly get lots of zeroes involved and then have to try to convert that into something that makes sense to a human. It doesn't help that calorie consumption for the PC is extreme (and I don't know if that also applies to companions).

RenechCDDA:
> I'd say the calorie cost ought to follow the same format as other entries, i.e. color coded red (done) or green (not mentioned), and also show the calorie store like other entries show what's available.
>
> e.g. Total calorie cost: 60000 (have 50000)
>
> Also, calorie storage measured in (kilo)calories make technical sense, but it's a rather inconvenient measure since you quickly get lots of zeroes involved and then have to try to convert that into something that makes sense to a human. It doesn't help that calorie consumption for the PC is extreme (and I don't know if that also applies to companions).

Adding the current larder calories would be easy.

Currently the entry `Total calorie cost:` is colored red(if insufficient) or white(otherwise) depending. Yes, ideally we'd like to apply the coloring to the *number* but the way `colorize` takes parameters is difficult. Maybe it can be done simply, I'll just poke it some more.

For the number count, I don't know of any easy way to insert commas as appropriate. It should also be noted that some languages/cultures/countries/*places that I don't live* use different systems (like a period instead, or according to google, a space!).
Could reduce it down to "workdays", which would produce a more human readable number. Under the old system, 60,000 (kilo)calories would be 24 days. The problem with that is that one (BRISK_EXERCISE) workday is no longer equivalent to one (MODERATE_EXERCISE) workday...

RenechCDDA:
All right @PatrikLundell here's your new and improved display.
image
image

PatrikLundell:
That's an improvement.

I'd rather see the text colored as well, but if the interface forces you to chose between coloring the number or coloring the text (with both not really an option), selecting the number as you have done is the better option in my opinion.
Feel free to disregard this if I'm too bothersome: on problem with making UI is that everyone and their dog have opinions about them...

PatrikLundell:
No, I meant the "Total calorie cost:" part, although my ideal would be to have the whole line colored, including the "(have ####) bit)".
Personally I care more about the UI looking good than the ugly code needed to produce it, but ideally there should be support operations allowing you to produce good looking UI elements without repeated ugly formatting workarounds.

RenechCDDA:
That's easy enough, but it would differ in behavior to the other displays?

For examples, "Tools required:" is not colored here.
https://user-images.githubusercontent.com/84619419/248922533-41baea46-4030-4c70-9ffa-a46b778ce617.png

RenechCDDA:
> Unfortunately, now when comparing things I saw that the image posted earlier said "6000(have 57028)", that is a missing blank between "0" and "(".

I don't think it will actually let me start a string with a space, so I was ~~hoping you wouldn't notice~~ going to ignore that, honestly.

PatrikLundell:
If the string support doesn't allow the blank then we'll have to live without it, but it's still good to know it's not a mistake, but rather a result of poor infrastructure support.
I'd still make a last ditch attempt to see if it's possible to escape a blank in, if escape character logic is available.

src/faction_camp.cpp
```diff
@@ -782,8 +782,24 @@ void basecamp::get_available_missions_by_dir( mission_data &mission_key, const p
...
+                                                   BRISK_EXERCISE );
+                const int available_calories = camp_food_supply( 0, false );
+                bool can_upgrade = upgrade.avail;
                 entry = om_upgrade_description( upgrade.bldg, upgrade.args );
-                mission_key.add_start( miss_id, display_name, entry, upgrade.avail );
+                if( foodcost > available_calories ) {
+                    can_upgrade = false;
+                    entry += string_format( _( "Total calorie cost: " ) ) +
+                             string_format( colorize( std::to_string( foodcost ), c_red ) ) +
+                             string_format( _( "(have %s)" ), available_calories );
```
dseguin, line 796:
```suggestion
                    entry += string_format( _( "Total calorie cost: %s (have %d)" ),
                                            colorize( std::to_string( foodcost ), c_red ),
                                            available_calories );
```

RenechCDDA: Ah yeah that is better, thank you. I totally forgot that inputs other than `%s` existed, because of the ability for code to automatically convert between types.

PR #66493, opened 2023-06-27
remove fine screwdriving from multitool

anoobindisguise:
#### Summary
Balance "multitool no longer has fine screwdriving"

#### Purpose of change
Fine screw driving represents having a bunch of screwdrivers for the numerous proprietary and unusual screws, of which there are dozens, and while you can often approximate it isn't appropriate for even an expensive multitool to have.

#### Describe the solution
Remove that quality from the multitool.

Remove it from the bionic multitool too.

### DIFF
--- a/data/json/items/tool/workshop.json
+++ b/data/json/items/tool/workshop.json
@@ -725,7 +725,7 @@
     "material": [ "steel" ],
     "symbol": ";",
     "color": "light_gray",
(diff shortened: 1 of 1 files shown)


Zireael07:
> Fine screw driving represents having a bunch of screwdrivers for the numerous proprietary and unusual screws,

This should be documented somewhere visible, e.g. in JSON_INFO.md.

PR #66494, opened 2023-06-27
reduce impact of bmi on healthiness

anoobindisguise:
#### Summary
Balance "BMI has a less all consuming impact on how healthy you are"

#### Purpose of change
BMI influence on healthiness was broken by the lifestyle system. If you had a BMI in an unhealthy range, it caused your daily healthiness to be lowered by that much each day (in many cases this was like -100 healthiness per day) which was extreme (fat people don't heal from injuries 90% slower and get out of bed every day feeling like they're dying) and also made your lifestyle basically meaningless. If you were mildly obese you may have had an unhealthiness factor of like -25 which means unless you can get more than 25 points of healthiness per day (basically impossible without chugging panaceii) your health would slowly tick down to -200. So being healthy didn't even matter, because no matter what you did you would become as unhealthy as possible very quickly. Since you were already at minimum healthiness, you can do all the mutagens, candy, hard drugs, etc you want with no consequences because it can't get any worse, and nothing you can do except losing weight would make it better.

"Feels awful" should basically only come from eating lots of mutant meat or abusing mutagens. Unless you're so fat that you're literally dying (800+ lbs) you shouldn't "feel awful" unless you're living a garbage lifestyle on top of that.

#### Describe the solution
No longer limit daily_health to your BMI factor.
Your health tally (the net result of your daily healthiness compounding over days to weeks) is "effectively" lower due to BMI factor, this is -5 points per BMI over Obese. There are comments documenting this in the code but in summary being fat means you "feel" worse than you would be otherwise, but with enough exercise and healthy eating you can still be healthy, it is just harder. Your daily healthiness (the summary of your activities, vitamin consumption, etc) is unaffected by BMI. 
So if you have an XXL char you're about 50 points of healthiness lower than usual. This makes your range of healthiness -200 -> 150 (your absolute best healthiness isn't as high as a thin person). Technically goes down to -250 as it's always a 400 point scale but it is capped at -200. Upon starting the game you have -50 healthiness but if you're healthy you can feel good, it just takes more effort.

Different scaling for the formula.

#### Testing
Compiled and tested and confirmed it's working as I want now.

#### Additional context
This also fixes obesity health not impacting you until Midnight #1. You start the game unhealthy now.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -4538,10 +4538,8 @@ void Character::update_health()
 {
     // Limit daily_health to [-200, 200].
     // This also sets approximate bounds for the character's health.
-    // if( get_daily_health() > 200 ) {
-
-    if( get_daily_health() > get_max_healthy() ) {
-        set_daily_health( get_max_healthy() );
+    if( get_daily_health() > 200 ) {
+        set_daily_health( 200 );
     } else if( get_daily_health() < -200 ) {
         set_daily_health( -200 );
     }
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -3943,7 +3943,7 @@ int Character::get_lifestyle() const
                                             5 * ( bmi - character_weight_category::obese ) ) );
     int under_factor = std::round( std::max( 0.0f,
                                    50 * ( character_weight_category::normal - bmi ) ) );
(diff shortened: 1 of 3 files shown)


anoobindisguise:
I still cannot for the life of me figure out why this caused natural healing to be 15% better.

PR #66496, opened 2023-06-27
Readd Glorious to Elf-A

anoobindisguise:
#### Summary
Balance "Elf-As are Glorious again"

#### Purpose of change
Elf-A had its beauty mutations taken away because it was going to be given "mind control pollen" powers but the person who was going to do that abandoned their project and nothing has been done since to add those powers. Further, being impossibly beautiful was kind of the most famous thing about elf-a and it's not mutually incompatible with mind control pollen.

#### Describe the solution
Re add beauty muts to Elf-A.

Remove Elf-A.


Fris0uman:
I think the change was to move it from Elf-A to Alpha, so if you revert it maybe it should not be Alpha thing anymore?

fairyarmadillo:
IMO they should both have it. Alpha is the pinnacle of human beauty, Elf-A is strange and alien but also beautiful.

RenechCDDA:
> IMO they should both have it. Alpha is the pinnacle of human beauty, Elf-A is strange and alien but also beautiful. They're also related projects per the lore (thus the similar names).

There's no way alpha should get the *fourth tier* of attractiveness mutations. https://discourse.cataclysmdda.org/t/mutation-category-expansion/3755

`This also has 0 Ugliness, and in fact improves one’s looks: even mutant-haters could pass/convince themselves it’s not “really” a mutation.`

Issue #59781, opened 2022-07-31
Incorrect crater map extra generation


GuardianDll:
### Describe the bug

Sometimes the game spawn bomb on the edge of the overmap tile, and it cant interact with not loaded adjacent overmap tile correctly 
(#59717 may be related for this)

### Steps to reproduce

Walk on the map
find sometimes craters spawns incorrectly

### Expected behavior

Correct crater spawn

### Screenshots

![image](https://user-images.githubusercontent.com/67688115/182021542-8abbf51b-f34c-4a63-8bf2-4e7d43360199.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.1826 (21H2)
- Game Version: 3b41d64 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Massachusetts [MA],
    SpeedyDex [speedydex],
    Stats Through Kills [stats_through_kills],
    Stats Through Skills [StatsThroughSkills],
    Magiclysm [magiclysm],
    Mythical Martial Arts [MMA],
    Xedra Evolved [xedra_evolved]
]

Night-Pryanik:
The most easiest solution is to place bomb in the center of OMT. This will work for medium-power bombs (such as on crater map extra), with explosion radius not crossing the boundaries of OMT. Consequently, this solution won't work for high-power bombs, as even one bomb with enough destructive power can have very big explosion radius which will cross the boundaries of OMT.

The proper solution would be to rework explosion mechanic so it can handle crossing the boundaries of OMT.

Procyonae:
I don't think this is anything to do with the explosion mechanic as such it's just that the crater bomb goes off at the edge of your reality bubble so the other OMT isn't loaded, if you just spawn bombs in they work fine over OMT and even 180x180 OM tile borders.

Procyonae:
Ok scrap that there's something weirder going on I'm getting square edges consistently without them even crossing any OM boundary
![image](https://github.com/CleverRaven/Cataclysm-DDA/assets/45432782/eb9d2596-555d-4978-85dd-a6ebdcb413cc)

Brambor:
Solved by
 - #75567

Issue #66269, opened 2023-06-18
multi-stage rubble piles, and rubble generation


flybye3o6:
### Is your feature request related to a problem? Please describe.

demolish a whole 5 lvl office tower and there is barley any rubble created
rubble spawned in the air and lands on rubble on the ground deletes the falling one
most other furniture when falling on rubble destroys it(not going to change just pointing out it contributes to less rubble)
rubble has no verticality to it, is it a light covering on the ground or a full pile you cant see over?
dumpsites only cover being fridges washers and cars that spawn

### Solution you would like.

already testing the multi-stage tiles and mapgen, seems promising, dumpsites seem more interesting
need help with
- _bash transforming full f_rubble into lower stage f_rubble_ (<done) and spawning the extra as f_rubble in an adjacent tile (ideally in the opposite direction of the bash, meaning you could use this to knock over a pile on top of an enemy on the other side)
- updating f_rubble stages when other rubble mite land on them transforming it into the next stage up
- if f_rubble cant transform into the next stage up (rubble of different types) skip over to another adjacent tile until it can spawn new f_rubble or update another pile
- damage mechanics for rubble falling on player and creatures (where is the code for that so i can look at it)
- furniture sealed container (when clearing f_rubble stage the items hidden underneath get kicked out to an empty space, if you are standing on f_rubble when clearing it throws up an error, i want the items already hidden underneath to be put in the new stage, _bashing causes the spawned items to be dropped in the new stage_)
- ~~suggestions for names and descriptions of tiles, not going to leave it as 1/4 2/4 3/4 full~~ _using pile<mound<heap<mountain adjectives_
- suggestions for buildings and methods that consistently cause rubble to spawn for testing on (big bomb on big building spawns no rubble)(some older mapgens use tiles on higher zlvls that turn into t_dirt when destroyed)
- ~~suggestions on how/where to find mutable dumpsites (when i debug spawn them they dont look right)~~ _overmap spawns need to be done in an unloaded area_
[...]

PR #54149, opened 2022-01-07
[Innawood] Various improvements based on feedback

Light-Wave:
#### Summary
Mods "Various Innawood improvements"

#### Purpose of change

Includes various improvements to the Innawood mod found by playtesting both by myself and by others. Mainly discussed at #53960

#### Describe the solution

All changes are to the Innawood mod only.
- Adds mineral loot to more caves
- Makes craters not be radioactive
- Makes cotton repairable with rag again
- Makes plank armguards and plank sword craftable with sticks
- Makes most brewing and distilling recipes autolearn
- Makes many cooking recipes using foraging components autolearn
- Makes water faucet and pilot light auto learn
- Makes fermenting barrels no longer require water faucets

#### Describe alternatives you've considered

Remove craters altogether. Right now the vague lore is that either a meteorite crashed in the large craters, or maybe Mi-Go dropped bombs on fungi structures. It is intentionally left vague, but I think there is enough non-human reasons why craters might exists.
Rework cotton repairing mechanics. Before this pull request, many items could be crafted without cotton, but required cotton to repair. Since cotton is basically non-existent in this mod, it meant many of the clothing items you could craft could not be repaired. Preferably, an item should be repairable by the clothing items used to craft it. But until that has been fully implemented, this temporary fix makes it possible to repair your items for now.
Alternatively, make wild cotton spawn, or make it possible to acquire cotton seeds. Sadly, cotton is not native to New England, which is still the growth zone this mod is taking place in.
Being able to mix seeds with mutagen to create other seeds could be an interesting way to gain access to cultivated crops such as cotton, maybe?
Make separate wooden armguards with different flavour text that describes the item being made out of sticks instead of planks. Maybe make bark arm guards as a stand-in for paper armguards as well.
Instead of removing water faucets from fermenting barrels, a way to make water faucets could be implemented. Currently they require plastic, which is not craftable without vinegar, which is not craftable without a fermenting barrel. Rubber could be used as a substitute, but that is not craftable in game, and all recepies I found requires vinegar or some other hard to get resource.
Again, having a research system for recipes as an in-between between autolearn and only learnable from books would be neat.

#### Testing

Made a character and warped around to different caves to make sure cave material would drop properly. They did. Checked that cotton clothing could be repaired by rags. They could.

PR #66499, opened 2023-06-27
Unhardcode "crater", improve appearance of mutable craters

Procyonae:
#### Summary
Content "Mutable (the large not extra ones) craters are now round"

#### Purpose of change

Hardcoding bad
Square craters ugly

#### Describe the solution

Unhardcodes "crater" and removes all references to it in the c++
New map creates round craters with a single map for the edges using neighbour based nests
Also added a small version for some more variety, annoyingly #59781 means I can't use less bombs without resulting in straight borders on the explosion (tried 1 and 2)
All the crater maps now use fallback_predecessor_mapgen allowing the mutable to spawn over roads sensibly and resulting in items where stuff like trees and road are destroyed
Slightly reduces the occurrences to make up for the fact they're bigger now
All unsolved joins are resolved over already altered terrain so should never cause issues

Not ruining the beautifully simplistic mutable by making the terrain's radius small enough it wouldn't need to be changed to accommodate corners
Sadly while vehicles that spawn in the blast get decimated as expected, ones that spawn over the place_rubble get damaged very inconsistently and I'm getting the occasional crash that appears to be from that so I'm removing all vehicles outside of the small area of no rubble which basically means you'll never see vehicles getting wrecked v_v
Ideally it would be able to spawn anywhere but that's not viable presently
I might make a follow up PR where I:
- Make it more randomly patchy at the edges (see the wibbly wobbly commit for the general idea)
- Allow it to spawn over more exclusively z0 maps like fields/pastures etc
- Add sensible items to the crater, maybe a fighter crash site variant as an alternate choice with no join in the first phase so it only makes one crater

#### Testing

Compiled and checked every phase variant spawned by making them blue one at a time

<details>
  <summary>Testing Screenshot</summary>

image
  
</details>

#### Additional context

<details>

image
Example over road

image
Example in forest

image
Very rare vehicle getting destroyed

image
Map screenshot to show potential sizes and shapes, minimum size is one 3x3 crater (Obv this many don't spawn)
  
</details>


ashGlaw:
Presently, are craters just sheer pits? Might make some sense for some ramps to spawn, that way the craters can be explored and don't provide some handy zombie storage

Procyonae:
> Presently, are craters just sheer pits? Might make some sense for some ramps to spawn, that way the craters can be explored and don't provide some handy zombie storage

They're currently (and still will be) explosions that occur when they enter your reality bubble the same as the extra currently which makes them fairly random without needing to do anything clever but at the same time means adding ramps would be very awkward if not impossible. Feasibly the fake explosive could be replaced with an EOC that causes an explosion and then creates ramps/slopes where appropriate but that's very much out of scope.
I could swap the explosives out for "set": "bash" or "rubble" nests with slopes at the sides but that would be messy and neither of those are particularly robust at destroying terrain, or I could just make holes and manually fill them with expected items but that's far from ideal.
I will say that they don't currently place any items themselves other than a very underwhelming "wreckage" itemgroup around the actual blast so there's no purpose to exploring them particularly but the zombie storage problem is pretty silly.
Also just generally I'd argue destroyed natural terrain should result in ramps/slopes but the current implementation of ramps/slopes is really iffy already so I defo won't be attempting to add that.

data/json/mapgen/crater.json
```diff
@@ -1,4 +1,15 @@
...
+  {
+    "type": "palette",
+    "id": "crater_palette",
+    "terrain": {
+      ".": "t_region_groundcover_barren",
+      "?": [ [ "t_region_groundcover_barren", 3 ], "t_null" ],
+      "0": "t_region_groundcover_barren"
+    },
+    "items": { ".": { "item": "wreckage", "chance": 2 }, "?": { "item": "wreckage", "chance": 1 } },
+    "nested": { ".": { "chunks": [ [ "rubble", 10 ], [ "null", 3 ] ] }, "?": { "chunks": [ [ "rubble", 3 ], [ "null", 10 ] ] } }
```
GuardianDll, line 11: What's the point to create 1x1 nests, if you can simply 
```json
"furniture": {
      ".": [ [ "f_rubble", 10 ], [ "null", 3 ] ]
      "?": [ [ "f_rubble", 3 ], [ "null", 10 ] ]
    },
```

Procyonae: place_rubble bashes terrain/furniture which is somewhat relevant bc it's using fallback_predecessor_mapgen so there could be road/trees/street lamps etc on the ?, the current way . is set up that stuff would've already been dismantled so it'd be useless there but I might still tweak the terrain distributions anyway I'm getting the shapes how I want first

Procyonae: There is also this #66269 which might eventually lead to rubble and set bash being more appropriate than manually placing rubble for general use too

LovamKicsiGazsii:
The screenshot's looking seriously awesome, well done so far!

Procyonae:
Innawood's crater was added in #54149, hasn't been touched since and only appears to exist to remove the radioactivity and that's been removed from the vanilla crater anyway to save you some time ^^

Light-Wave (approved):

PR #66502, opened 2023-06-28
Redo nomad gear to be more sensible

anoobindisguise:
#### Summary
Balance "nomad gear is denim based instead of lycra based"

#### Purpose of change
Nomad jumpsuits were made of lycra, which was not realistic for multiple reasons.
- Lycra is light and not protective whatsoever and sucks as a base material for sewing things onto it (such as leather) because it is flimsy and stretchy and a stiff padding material would ruin it.
- The jumpsuits were way too heavy for being some insubstantial stretch fabric + a bit of padding.
- Lycra is not a reasonable material for the Exodii to base their designs on because many worlds they visit (such as their previous jump) are not technologically advanced. Presumably the Exodii still parlay with such people because even a medieval society has plenty of things to offer them, and as such the designs they create should be universally applicable rather than require something that has existed for less than 0.1% of all of written history.

#### Describe the solution
These jumpsuits are now cotton based, denim specifically. Canvas might be slightly more appropriate but its usage in game represents a waterproofed fabric and twill weave is still basically as protective while being more breathable, which is noted as a desirable quality of the jumpsuits. Recipes have been adjusted accordingly.
Leather coverage is now present on both jumpsuits and is low (30-45%) representing reinforcing specific critical locations while still maintaining breathability and flexibility overall.
I also fixed Marloss NPCs wearing Nomad gear (they now just wear a robe) and gave these recipes the SECRET flag so that starting with high skill does not automatically teach you the recipes. This is debatable, but this armor is highly optimized over decades or centuries of refinement and thus has pretty high defense:encumbrance ratios and also covers stuff like CBM integration.
For the crafting recipes I allowed jeans and jean jackets to be used substituted for some of the denim, and while doing so I noticed that the flannel jacket specifically is made of denim so I changed its material to denim and allowed it to be used in the recipe.

I did not touch the bodyglove or the nomad plate armor. I'm not sure what to do about those.
If we had Hemp, Silk or Bamboo I'd probably make it out of those to increase breathability, but since only cotton is available Denim felt the most appropriate.

#### Testing
Loaded in game, took some screenies, and simplified the recipes once I realized they created too many variants.

#### Additional context
I've been mucking around with the recipes a little to try and keep them within the 105 variant limit. They're pushing it for sure. I think they currently barely fit.
light nomad js
nomad js
nomad cowl
nomad cowl 2


fairyarmadillo:
The bodyglove and bodymesh could be their preferred designs when more technologically advanced worlds are encountered, using aramids and lycra instead of natural materials.

Aerin-of-the-Toast:
finally, the jomad jumpsuit

anoobindisguise:
I got some feedback regarding balance concerns and 2mm of denim probably being too much for an airy suit, and since I wasn't sure about the previous mm thickness myself, I went ahead and lowered by half the denim thickness. 0.5mm for light and 1mm for normal.

fairyarmadillo:
This looks great. It's not magically more protective than an Aketon, but it's got a lot more utility. A+

PR #66505, opened 2023-06-28
Fix and re-enable 3D map memory

Rewryte:
#### Summary
Bugfixes "Fix and re-enable 3D map memory"

#### Purpose of change
Map memory always had buggy interactions with 3D vision since #65738.  Cross z-level map memory was disabled in #66383 due to memory tiles being drawn on the wrong z-levels being more prominent on isometric tilesets.

#### Describe the solution
Map memory itself is stored in 3D, but the caching and retrieval of submaps are performed on a single z-level.  The relevant code was revised and cross z-level map memory is re-enabled.

#### Testing
Builds ok
Map memory now works as expected
Memory tiles are drawn on the correct z-level on isometric tilesets
Tested on various tilesets and 3D settings
Chiquita Waters _2023-06-28T11-40-44

### DIFF
--- a/src/cata_tiles.cpp
+++ b/src/cata_tiles.cpp
@@ -2936,7 +2936,7 @@ bool cata_tiles::draw_terrain( const tripoint &p, const lit_level ll, int &heigh
             get_terrain_orientation( p, rotation, subtile, {}, invisible, rotate_group );
             // do something to get other terrain orientation values
         }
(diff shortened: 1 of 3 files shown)


bombasticSlacks:
LGTM but gonna let a few of the usual suspects put eyes on it before merging 😄

PR #66509, opened 2023-06-28
Fix generator UPS not consuming fuel

SurFlurer:
#### Summary
none

#### Purpose of change
Since #66348, generator UPS items need a `fuel_efficiency` field to properly consume fuel.

#### Describe the solution
Add fuel efficiency data to generator ups items.

The efficiency of power armor generator frame was copied from diesel vehicle engines.

The efficiency of gas charger was somewhat arbitrarily chosen, as I cannot find reliable data from the Internet for such an appliance's efficiency. If anyone would like to point me a way to get realistic data, I will be happy to commit them.

#### Testing
They are correctly consuming fuel ( although rather roughly because the energy consumption on item level is counted as kj, and the consumed charges are rounded up. )

### DIFF
--- a/data/json/items/armor/power_armor.json
+++ b/data/json/items/armor/power_armor.json
@@ -294,6 +294,7 @@
     "looks_like": "backpack_tactical_large",
     "color": "light_gray",
     "ammo": [ "jp8" ],
+    "fuel_efficiency": 0.4,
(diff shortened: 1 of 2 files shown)


Drew4484:
Thank you for grabbing this!

PR #66544, opened 2023-06-30
Park benches in parks!? More likely than you may think!

Karol1223:
#### Summary
None

#### Purpose of change
Aaaaaaages ago I added some new furniture in hope that mapgen people would use it. They didn't. I'm doing it now.

#### Describe the solution
Swapped the benches in park palettes into the unused park benches. Changed some mapgen for parks around so that regular benches are still used near tables.

I wanted to make plywood chairs spawn in schools but schools aren't palettized so it's a bit out of my skillset

#### Testing
Spawned the mapgen and lo and behold - park benches

#### Additional context
@GuardianDll brother do me a solid and look at this to make sure I didn't break something horribly


Procyonae:
Make picnic benches and tables too :^)

Karol1223:
> Make picnic benches and tables too :^)

I do want to get more PRs for the furniture project going but they take a lot of looking at IKEA schematics and I value my sanity a little more than that. But yeah, I'll get there. They're just at the back of the todo list.

GuardianDll (approved):
Looks good for me

data/json/mapgen/park.json
```diff
@@ -401,14 +401,14 @@
...
+        "zftfz qq z t  cTc    t z",
+        "zfffz qq zz   cTc      z",
         "zfffz qq! zz       !   z",
         "zfffz qqq  zz          z",
         "zftfz  qqq,,,,,        z",
-        "zfffzz  qqq zz &   bb  z",
+        "zfffzz  qqq zz &   cc  z",
         "zzzzzzz  qq  z     TT  z",
-        "         qqq zz    bb  z",
+        "         qqq zz    cc  z",
```
Maleclypse, line 411: So below you make the b furniture a park bench and then up here in the park you make the mapgen use c which is a non park bench?

Karol1223: Indeed! This is partially because I am an idiot and partially because I didn't notice this not only added more bloat but didn't even resolve the issue. I'll fix it in a sec.

PR #61614, opened 2022-10-11
Fishy Cleanup

Karol1223:
#### Summary
Balance "Remove some fish, fix up some others"

#### Purpose of change
The fish were kind of a mess. Some were specified up to species level, some were generic 'Salmon'. Their sizes also didn't always reflect the real-world average for a particular species. Overall it was a mess I wanted to clean up.

#### Describe the solution
**REMOVAL**
- [x] Trout was removed on account of being a filler, nondescript trout while we had all other trout species from the region
- [x] Steelhead Trout was removed on account of being literally the same animal as the Rainbow Trout we already have
- [x] Chinook Salmon, Coho Salmon, White Bass, Longear Sunfish and Redear Sunfish were removed on account of not being present in New England according to iNaturalist

**NAME CHANGES**
- [x] Salmon -> Atlantic Salmon
- [x] Whitefish -> Lake Whitefish
- [x] Perch -> Yellow Perch
- [x] Sunfish -> Banded Sunfish
- [x] Pumpkinseed Sunfish -> Pumpkinseed
- [x] Calico Bass -> Black Crappie
- [x] Bullhead -> Brown Bullhead
- [x] Pike -> Northern Pike
- [ ] Pickerel -> Chain Pickerel
- [x] Carp -> European Carp
- [x] Freshwater Eel -> American Eel

**SIZE CHANGES**
- [x] Trout, Salmon and Whitefish
- [x] Bass
- [x] Perches
- [x] Sunfish and Crappie
- [x] Catfish
- [x] Pikes
- [x] Cyprinids
- [x] The three at the end of the list
- [x] Crayfish

Also changed the Fallfish' description to not reference Gar. They're not related to Gar unless we're talking about different fish I'm not aware of. They're cyprinids. Carp.

ALSO ran the dictionary file through an autosorter so things are alphabetical now

#### Describe alternatives you've considered
Changing the name of the Kokanee Salmon to Sockeye Salmon but both work, not changing the Pumpkinseed's name since it's also applicable but its just uncomfortably long right now.

#### Additional context
**CALCULATIONS**
[...]

PR #66548, opened 2023-06-30
Fish size cleanup

Karol1223:
#### Summary
None

#### Purpose of change
In #61614 I audited fish sizes. And like an absolute moron, I calculated them as CUBES. We all know fish are cylinders and not cubes. This is unacceptable. My immersion is ruined.

#### Describe the solution
Recalculated all the fish as cylinders and not cubes - most now have lower volume, except a few where I miscalculated it in the original PR. A few have changed weights since back then I didn't pick out the correct average weights. Pickerel renamed to Chain pickerel since I forgot to do it the last time.

Also apparently threw the dictionary through a tool to make it alphabetical. Smh people.


Procyonae:
They should be two back to back cones smh /s

PR #66550, opened 2023-06-30
Update basecamp fake char kiln's ammo restriction

SurFlurer:
#### Summary
none

#### Purpose of change
Charcoal's volume per charge has been changed. Update the fake kiln to be in line with the new data. Otherwise it can cause error when finishing the transformation of materials to charcoal.
```
 DEBUG    : tried to put an item (charcoal) count (20000) in a container (fake_char_kiln) that cannot contain it: pocket unacceptable because tried to put too many charges of ammo in item

 FUNCTION : put_in
 FILE     : E:\Cataclysm-DDA\src\item.cpp
 LINE     : 1575
 VERSION  : 0.G-2634-geddf858866-dirty

```

#### Describe the solution
Update the ammo restriction.

#### Testing
No more error.

### DIFF
--- a/data/json/items/basecamp.json
+++ b/data/json/items/basecamp.json
@@ -9,7 +9,8 @@
     "sub": "char_kiln",
     "extend": { "flags": [ "ALLOWS_REMOTE_USE" ] },
     "ammo": [ "charcoal" ],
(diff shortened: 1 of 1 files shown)


RenechCDDA (approved):
A quick blame shows that this was not updated at the time of charcoal changes. Good catch.

PR #66551, opened 2023-06-30
Active backup generator ---> Functional generator appliance, useable in player grids

RenechCDDA:
#### Summary
Content "Active backup generator can now be transformed into an appliance"

#### Purpose of change
~~Fixes several issues (WIP on links)~~
Apparently resolves no feature requests that I can find! But this has been repeatedly brought up in the development discord, and has provided a good deal of confusion. It's a generator. It *should* work as one!

#### Describe the solution
Make generator into appliance, mirror the energy output and fuel consumption of an actual vehicle with constituent parts attached. Specifically we are emulating a vehicle which is simply a steel frame, an I4 diesel engine, and a truck alternator.

Add recipe to construct your own (it's just frame+engine+alternator)

Add deconstruct recipe and clean up interactions with the furniture version now that it has an item form

#### Testing
Appliance is fully functional, recipe and disassembly work as written.

Save provided with all parts spawned/available: 
[Asharoken-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11918824/Asharoken-trimmed.tar.gz)

Includes nearby bicycle with I4 diesel engine and truck alternator (which generator is based off of) for comparison.

#### Additional context
This should NOT be backported to 0.G - the vehicle code has changed significantly since then and I sincerely doubt it will work on that version

### DIFF
--- a/data/json/furniture_and_terrain/appliances.json
+++ b/data/json/furniture_and_terrain/appliances.json
@@ -979,7 +979,7 @@
     "location": "structure",
     "damage_modifier": 80,
     "durability": 400,
-    "description": "FIXME",
+    "description": "A ubiquitous piece of compact machinery meant for running anything from the bathroom lights to electric stoves and other necessities when grid electricity is unavailable.",
     "//1": "BIZARRE FIELD NAMES HERE! Weird stuff happening! Epower determines how fast it is converted, power determines the ratio of diesel(mL) to battery charges output. In other words, *power* is *fuel efficiency*.",
     "//2": "On a normal vehicle 250mL of diesel in an I4 with a truck alternator (what the active backup gen uses) should provide ~1692 charges (kW?) of power in about 24 minutes based on a test setup. That's a little less than 7W power, rounded up as I assume a larger dedicated generator is more efficient.",
     "power": "7 W",
--- a/data/json/furniture_and_terrain/appliances.json
+++ b/data/json/furniture_and_terrain/appliances.json
@@ -969,6 +969,44 @@
     "variants": [ { "symbols": "0", "symbols_broken": "#" } ]
   },
   {
+    "type": "vehicle_part",
+    "id": "ap_active_backup_generator",
+    "name": { "str": "active backup generator" },
+    "looks_like": "f_active_backup_generator",
+    "categories": [ "energy" ],
+    "color": "green_white",
+    "broken_color": "blue",
+    "location": "structure",
+    "damage_modifier": 80,
+    "durability": 400,
+    "description": "FIXME",
+    "//1": "BIZARRE FIELD NAMES HERE! Weird stuff happening! Epower determines how fast it is converted, power determines the ratio of diesel(mL) to battery charges output. In other words, *power* is *fuel efficiency*.",
+    "//2": "On a normal vehicle 250mL of diesel in an I4 with a truck alternator (what the active backup gen uses) should provide ~1692 charges (kW?) of power in about 24 minutes based on a test setup. That's a little less than 7W power, rounded up as I assume a larger dedicated generator is more efficient.",
+    "power": "7 W",
+    "//3": "Truck alternator value",
+    "epower": "1320 W",
+    "//4": "This value doesn't seem to do anything for the appliance version? Let's keep it commented out for future reference.",
+    "//energy_consumption": "1000 W",
+    "m2c": 5,
+    "//5": "Actually accepts any diesel-type fuel but does NOT account for their relative fuel values! For diesel this is not a huge deal, only a few % inaccuracy, but for e.g. gas turbines this would be very bad!",
+    "fuel_type": "diesel",
+    "//6": "Appliance engine does NOT accept having these fields.",
+    "//fuel_options": [ "diesel", "biodiesel", "lamp_oil", "motor_oil", "jp8" ],
+    "//damaged_power_factor": 0.25,
+    "item": "active_backup_generator",
+    "requirements": {
+      "repair": {
+        "skills": [ [ "mechanics", 4 ] ],
+        "time": "60 m",
+        "using": [ [ "repair_welding_standard", 5 ], [ "soldering_standard", 5 ] ]
+      }
+    },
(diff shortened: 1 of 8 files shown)


GuardianDll:
Can you please also replace this entry with new generator-item?

https://github.com/CleverRaven/Cataclysm-DDA/blob/83224ec6845602a87f0ddf8fb3d52b691c1cf122/data/json/itemgroups/Locations_MapExtras/airdrop.json#L103-L118

RenechCDDA:
> Can you please also replace this entry with new generator-item?

Done. I also threw in the floodlight from the field manual since we have items for that, if that's fine?

data/json/furniture_and_terrain/appliances.json
```diff
@@ -968,6 +968,44 @@
...
+    "categories": [ "energy" ],
+    "color": "green_white",
+    "broken_color": "blue",
+    "location": "structure",
+    "damage_modifier": 80,
+    "durability": 400,
+    "description": "A ubiquitous piece of compact machinery meant for running anything from the bathroom lights to electric stoves and other necessities when grid electricity is unavailable.",
+    "//1": "BIZARRE FIELD NAMES HERE! Weird stuff happening! Epower determines how fast it is converted, power determines the ratio of diesel(mL) to battery charges output. In other words, *power* is *fuel efficiency*.",
+    "//2": "On a normal vehicle 250mL of diesel in an I4 with a truck alternator (what the active backup gen uses) should provide ~1692 charges (kW?) of power in about 24 minutes based on a test setup. That's a little less than 7W power, rounded up as I assume a larger dedicated generator is more efficient.",
+    "power": "7 W",
```
SurFlurer, line 985: This is a typo, should be 7 kw at least ( also for the comment above ), otherwise this appliance will drink up 20L of diesel in less than 2 minutes. According to https://en.wikipedia.org/wiki/Diesel_generator and https://www.generatorsource.com/Diesel_Fuel_Consumption.aspx , I think the reasonable number is around 10 kw.

RenechCDDA: I'm not sure I understand. Increasing the `power` field will result in *generating more energy for a given amount of diesel*. (Yes that's weird!)

Using the table provided in the 2nd link, a 20kW generator is expected to use 1.6gal/hr (~6.05L). We're using about 625mL/hr (~10.3% of that), and our expected power output would thus be about 2.06kW.

Our actual power output, as defined by epower, is *1320 W*(1.32kW). So, we're off by a decent amount but it perfectly mirrors an existing ingame setup (the truck alternator+i4 diesel engine in a static frame).

We're certainly not off by *three orders of magnitude*, though?

SurFlurer: This is a math problem ^^. With `power: 7 W`, You need 188ml of diesel to generate 1320 joule of energy. So 188ml of diesel for one second! I'm not complaining about "power in the physical sense", but is really the matter of fuel efficiency.

RenechCDDA: > So 188ml of diesel for one second!

This is what I am not understanding. What do you mean 'for one second'?

So, we have battery charges. Those are... *kilojoules*, yeah? (I know my comment says kW, power units are a bane!). And 1 kJ == 1kW*s (one kilowatt-second)?

188mL of diesel should produce about ~1272 charges (kJ) of battery according to the tests I did.

RenechCDDA: > Oh probably I misunderstood. I'll check it and respond soon. Sorry.

Nothing to apologize for, please do let me know if you find an error. This stuff is confusing at the best of times, and I am not the best at writing comments for it!

SurFlurer: Ah I know what I got wrong. I took it for granted that "power": "7 W" means 1ml of diesel will produce 7 joules of energy. But in fact that's 7 kilo-joules. The unit of it made me dizzy. Sorry for wasting your time 😢

RenechCDDA: Again, no worries!

VincentRoyceMorgan: There's an existing vehicle called "portable_generator" that uses a 7.5kW generator instead of an alternator and a way smaller 0.6L V2 gasoline engine. I'm going to assume that the values they use are considered balanced and accurate so I'd suggest using those numbers for this generator, or at least a more powerful version.
Reason being that the 7.5kW generator requires 10444 engine power and the small V2 engine already provides 37300 of it.
Meanwhile a truck alternator requires 2238 engine power and both versions of the V4 inline produce 93250 power while consuming over twice the fuel. That's quite the large disparity in engine power provided versus necessary.
There's two even smaller 1-cylinder engines (25cc and 50cc) that could be used for smaller, weaker generators with car and truck alternators. Though the smallest engine alone could power multiples of them, if we're to follow the numbers.
Only issue is that all 3 engines I mentioned are gasoline only. Having diesel versions (for appliance generators) would be nice, although I'm not sure if 1-cylinders are available in diesel. The 0.6 V2 does happen to have real-world diesel counterparts.

FayFlightee: > There's an existing vehicle called "portable_generator" that uses a 7.5kW generator instead of an alternator and a way smaller 0.6L V2 gasoline engine. I'm going to assume that the values they use are considered balanced and accurate so I'd suggest using those numbers for this generator, or at least a more powerful version.
> 
> Reason being that the 7.5kW generator requires 10444 engine power and the small V2 engine already provides 37300 of it.
> 
> Meanwhile a truck alternator requires 2238 engine power and both versions of the V4 inline produce 93250 power while consuming over twice the fuel. That's quite the large disparity in engine power provided versus necessary.
> 
> There's two even smaller 1-cylinder engines (25cc and 50cc) that could be used for smaller, weaker generators with car and truck alternators. Though the smallest engine alone could power multiples of them, if we're to follow the numbers.
> 
> Only issue is that all 3 engines I mentioned are gasoline only. Having diesel versions (for appliance generators) would be nice, although I'm not sure if 1-cylinders are available in diesel. The 0.6 V2 does happen to have real-world diesel counterparts.

Yeah, this has been brought up in devcord too. It provides so little power for quite a large engine used in it, ESPECIALLY when compared to already existing Portable Generator.

Issue #66545, opened 2023-06-30
Add possibility of overriding overmap tiles using `om_terrain_match_type` variable.


VladMasarik:
### Is your feature request related to a problem? Please describe.

Related to #66142 it has been suggested by @Procyonae that using `om_terrain_match_type` attribute with `PREFIX` could simplify the Alternative map keys mod. Currently the mod is defining every possible tile one by one, and overriding it, eg. ranch has up to 70 entries in the JSON
https://github.com/CleverRaven/Cataclysm-DDA/blob/83224ec6845602a87f0ddf8fb3d52b691c1cf122/data/mods/alt_map_key/overmap_terrain.json#L3149-L3164

AFAIK the `om_terrain_match_type` attribute is currently used in cases like mission definitions and starting locations, eg. https://github.com/CleverRaven/Cataclysm-DDA/blob/83224ec6845602a87f0ddf8fb3d52b691c1cf122/doc/JSON_INFO.md#terrain-1
When I tried to use it in the mod I got errors and crashes complaining that it cannot parse the JSON because an unknown field is used. Which is I guess expected since I didn't see any similar examples of using it in the overmap terrain. It is also possible that I just didn't define it properly, but I tried multiple definitions.

### Solution you would like.

Use just one entry instead, which would override majority of the tiles without needing to specify them manually, eg.
```
  {
    "type": "overmap_terrain",
    "id": "ranch_camp",
    "copy-from": "ranch_camp",
    "name": "field",
    "sym": "r",
    "color": "i_brown",
    "terrain": [ "field", { "om_terrain": "ranch_camp", "om_terrain_match_type": "PREFIX" } ],
  },
```
Note: I dont know if the `id` field would make sense in that case, as I would imagine everything with the `ranch_camp` prefix would start using the brown `r` symbols.

### Describe alternatives you have considered.

Instead of defining `ranch_camp_21` and ever other entry as a single entity, it is possible to define them using a multiple `id`s. eg. 
https://github.com/CleverRaven/Cataclysm-DDA/blob/83224ec6845602a87f0ddf8fb3d52b691c1cf122/data/json/overmap/overmap_terrain/overmap_terrain_commercial.json#L11-L19
This is currently the way I am improving the mod (PR to be yet published), however using a `PREFIX` would simplify the maintenance of the mod as adding any building alternatives would be covered by default, and not have to be added manually.

### Additional context

[...]

PR #66554, opened 2023-06-30
[Alternative Map Key] Building alternatives now show correctly on the map

### LINKED ISSUE #66142: Alternative map keys mod not working on all tiles
### Describe the bug

When using the the Alternative map keys mod, some tiles on the map should be changed from regular `^` character to a first letter in the name, eg. `h` for hardware store. However, this doesn't work for all variations of the buildings, eg. `^` is replaced with `h` for `s_hardware` tiles, but `s_hardware_1` is still kept as `^`, 

### Attach save file

[Boynton Beach-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11715693/Boynton.Beach-trimmed.tar.gz)

### Steps to reproduce

1. Disable tilesets for the overmap, eg. set it to `false` and use pure ASCII
2. Generate new map with the `alt_map_keys` mod enabled
3. Reveal the map using debug menu
4. Search for hardware stores, some of them will have just `h`, and some of them will be marked as `^`
5. In debug > teleport > long range > when you however over the `h` tiles, you can see it is just a `s_hardware` tile, while the `^` tile will be `s_hardware_1` or other variations.

### Expected behavior

The mod tries to make some locations more readable by replacing the `^` tiles. However, this is not done in all cases, and I would expect every variation of the hardware store, and other buildings to have their appropriate tiles, eg. `h` in case of the hardware store. Without it, the map is inconsistent.

### How it is currently:
before

### With a custom fix:
after

```
- OS: Linux
    - OS Version: Distributor ID: Ubuntu; Description: Ubuntu 20.04.6 LTS; Release: 20.04; Codename: focal; 
- Game Version: 602e9a2 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Alternative Map Key [alt_map_key]
]
```

### Additional context

As you can see in the screenshots, I was able to fix the issue myself in the case of hardware stores. I changed these lines https://github.com/CleverRaven/Cataclysm-DDA/blob/e40864f9989a843c7f6525084e0ec87a9c30c674/data/mods/alt_map_key/overmap_terrain.json#L537-L544

To this:
```
  {
    "type": "overmap_terrain",
    "id": ["s_hardware", "s_hardware_1", "s_hardware_2"],
    "copy-from": "s_hardware",
    "name": "hardware store",
    "sym": "h",
    "color": "light_cyan"
  },
```

There is a fix, however it seems that it would require including every possible `id` for each variety of the tile. **I am happy to create a PR for a fix myself**. My single question though is, is there any other way to include all possible variations without having to specify each `id` of the variation?

I thought that the `looks_like` field would solve this issue by default, but I am opening this issue because it doesn't seem to. https://github.com/CleverRaven/Cataclysm-DDA/blob/e40864f9989a843c7f6525084e0ec87a9c30c674/data/json/overmap/overmap_terrain/overmap_terrain_commercial.json#L250-L258

VladMasarik:
#### Summary
Mods "[Alternative Map Key] Building alternatives now show correctly on the map"

#### Purpose of change
Partially fixes #66142 where, when using the Alternative Map Key mod, it would not show proper map symbols on the map, and instead it would show the "old" symbols, i.e. symbols from the base game.

#### Describe the solution
I have added the building alternatives, as included in the base game, to the mod itself. So now it includes alternatives for the few fixed building.
https://github.com/CleverRaven/Cataclysm-DDA/blob/e40864f9989a843c7f6525084e0ec87a9c30c674/data/json/overmap/overmap_terrain/overmap_terrain_commercial.json#L250-L258

Specifically, I have added alternatives for these buildings:
- hardware
- clothing store
- restaurant
- tea shop
- fast food
- bar
- coffee shop
- bookstore
- electronics store
- pizza parlor
- gun store
- bank
- pawn shop
- mil. surplus
- butcher shop
- bike shop
- grocery store
- pharmacy
- doctor's office
- laundromat
- dispensary
- pet supply
- dollar store
- candy shop

There is an alternative described in #66545 however I dont know whether it is possible, nor how long it would take to implement. 

#### Testing
1. Update the Alternative Map Key mod locally
2. Start the game
3. Create new world
4. Reveal map > teleport to an adjacent overmap > reveal map > repeat 20 times > view map
5. Didn't throw any errors, and I saw the correct tiles on the map

#### Additional context
The screenshots show the whole map just in case, but the red square is the most obvious change.

NOTE! I did not fix every incorrect building. There are loads of them that are still missing, however I would like to add them in increments as some have "special cases", and some have to be added as completely new entities into the mod, which requires discussion on what symbols, and colors they should have. Plus, majority of them were from two files `overmap_terrain_recreational.json` and `overmap_terrain_commercial.json`, so I wanted to keep it to those two files for now only. More will follow if this PR is accepted.

Before:
before1

After:
after1

### DIFF
--- a/data/mods/alt_map_key/overmap_terrain.json
+++ b/data/mods/alt_map_key/overmap_terrain.json
@@ -320,7 +320,7 @@
   },
   {
     "type": "overmap_terrain",
-    "id": "s_pharm",
+    "id": [ "s_pharm", "s_pharm_1" ],
     "copy-from": "s_pharm",
     "name": "pharmacy",
     "sym": "p",
@@ -328,7 +328,7 @@
   },
   {
     "type": "overmap_terrain",
-    "id": "office_doctor",
+    "id": [ "office_doctor", "office_doctor_1", "office_doctor_2" ],
     "copy-from": "office_doctor",
     "name": "doctor's office",
     "sym": "d",
@@ -528,7 +528,7 @@
   },
   {
     "type": "overmap_terrain",
-    "id": "s_grocery",
+    "id": [ "s_grocery", "s_grocery_1" ],
     "copy-from": "s_grocery",
     "name": "grocery store",
     "sym": "g",
@@ -536,7 +536,7 @@
   },
   {
     "type": "overmap_terrain",
-    "id": "s_hardware",
+    "id": [ "s_hardware", "s_hardware_1", "s_hardware_2", "s_hardware_3" ],
     "copy-from": "s_hardware",
     "name": "hardware store",
     "sym": "h",
@@ -544,7 +544,7 @@
   },
   {
     "type": "overmap_terrain",
-    "id": "s_electronics",
+    "id": [ "s_electronics", "s_electronics_1", "s_electronicstore" ],
     "copy-from": "s_electronics",
     "name": "electronics store",
     "sym": "e",
@@ -568,7 +568,18 @@
   },
   {
     "type": "overmap_terrain",
-    "id": "s_gun",
+    "id": [
+      "s_gun",
+      "s_gun_looted",
+      "s_gun_1",
+      "s_gun_2",
+      "s_gun_3",
+      "s_gun_3_looted",
+      "s_gun_4",
+      "s_gun_4_looted",
+      "s_gunstore",
+      "s_gunstore_looted"
+    ],
     "copy-from": "s_gun",
     "name": "gun store",
     "sym": "g",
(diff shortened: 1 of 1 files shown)


VladMasarik:
I am not sure what to do about the failing test:
> In nested mapgen playset_4x4_1 in mapgen house_23 on house_23_west, setting terrain to t_monkey_bars (from t_dirt) at (4,19) when item spray_can existed.

I am not editing the houses eg. `house_23`, nor spawning items :thinking:

Maleclypse:
Thanks for doing this!

PR #66561, opened 2023-06-30
Update clear_character to remove mutation categories

RenechCDDA:
#### Summary
None

#### Purpose of change
Followup on https://github.com/CleverRaven/Cataclysm-DDA/pull/65732#discussion_r1204629324

#### Describe the solution
Move the act of clearing from the test to clear_character
Shake the testing tree, see what falls out

This could be put into Character::clear_mutations instead?

#### Testing
Github tests


RenechCDDA:
...Ok I guess that works

PR #66563, opened 2023-06-30
s_restaurant merge

Procyonae:
#### Summary
None

#### Purpose of change
There are 3 versions of the downstairs of s_restaurant and they don't even use a shared palette

#### Describe the solution
Merges the 3 versions of s_restaurant's downstairs into one
Fixes kitchen floor
Makes the NPC/roach variants mutually exclusive and adds a mutually exclusive version with lots of zeds
NPC version is also mutually exclusive with foodplace bc it doesn't make much sense to have both NPCs in there imo
Fixes a couple of wrong terrain/item definitions in the foodplace upstairs like a ladder appearing under the bin
The rare foodperson versions have been boosted so the average player might actually encounter them
Various other tweaks, most coord based stuff is now symbol based etc

Making Foodperson's "lair" more "lairy"
Doing something about the fact the side windows can spawn directly against a wall

#### Testing

Spawned several in


bombasticSlacks:
With the foodperson lair moved to a nest can you still guarantee it for the Foodperson starting scenario?

Procyonae:
> With the foodperson lair moved to a nest can you still guarantee it for the Foodperson starting scenario?

The upper floor/lair isn't a nest it's still it's own map it just uses the same lower floor map so the starting scenario still works the same

bombasticSlacks:
Thanks for answering that makes sense you just reused the bottom floor, just wanted to make sure.

PR #66575, opened 2023-07-01
Add ability to cull more farmed/bred animals

RenechCDDA:
#### Summary
Balance "Most friendly breedable animals can now be culled without having to attack them"

#### Purpose of change
There exists many animals which can be tamed and bred, but not culled, even if that is reasonably possible.

#### Describe the solution
Add culling to the following animals:
-rat
-rabbit
-pigeon
-cat (and anything inheriting from cat)
-dog (and anything inheriting from dog)
-horse (foal/child only. Adults remain unable to be culled)
-ferrets
-Sheep(children, adults could already be culled)
-Goats(both children and adults)

#### Testing
Linted

#### Additional context
I believe this leaves only adult horses as a tameable that's not also cullable. I left them out on purpose, as they are very difficult animals to work with. (i.e. I don't know if it's appropriate to be able to cull them)

### DIFF
--- a/data/json/monsters/bird.json
+++ b/data/json/monsters/bird.json
@@ -640,7 +640,7 @@
       "feed": "The %s seems to like you!  It runs around your legs and seems friendly.",
       "pet": "The %s runs around your leg."
     },
-    "flags": [ "SEES", "HEARS", "SMELLS", "ANIMAL", "PATH_AVOID_DANGER_1", "WARM", "FLIES", "SWARMS", "CANPLAY" ]
+    "flags": [ "SEES", "HEARS", "SMELLS", "ANIMAL", "PATH_AVOID_DANGER_1", "WARM", "FLIES", "SWARMS", "CANPLAY", "CAN_BE_CULLED" ]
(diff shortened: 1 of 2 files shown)


PatrikLundell:
Cows and sheep can be tamed, although I've never seen a calf or a lamb. Pigs can be tamed (as piglets), and I've done that in the past. Chicken can be tamed (as could other birds in the past, which may have changed). Goats can theoretically be tamed (theoretically, because you'd have to corner them to get near enough to attempt to tame them, as with cats).

RenechCDDA:
> Cows and sheep can be tamed, although I've never seen a calf or a lamb. Pigs can be tamed (as piglets), and I've done that in the past. Chicken can be tamed (as could other birds in the past, which may have changed). Goats can theoretically be tamed (theoretically, because you'd have to corner them to get near enough to attempt to tame them, as with cats).

All of these were made cullable in the original addition except... apparently, goats! They're using copy-from for about a single field (petfood) which is why I didn't spot them.

PR #66579, opened 2023-07-01
electric-jackhammer-tweaks-2

flybye3o6:
#### Summary
Content "changes electric jackhammer to use a cable and adds a battery jackhammer"

#### Purpose of change
more work that i said i would do in #64182

#### Describe the solution

electric jackhammer changed to 26kg
removed UPS, changed crafting recipe
has 12m long cable that can be pugged into a power grid

added battery jackhammer with a swappable power pack
less efficiency (about 20% rn), more weight and volume from the battery
added battery version to loot spawns, want it to be even more rare then the others

undecided on cable length, less efficiency amount, crafting requirements and components, loot spawn rates
looking into if can use TOOLMOD with extension cables and battery pack, then wont need the 2nd jackhammer just use tool mods to change between extension cable, outdoor extension cable and storage battery

#### Testing

cable version will draw power from attached grid with use action, crafting and construction
can not start craft or construction with pocket battery set to 0 power, set to 3960 for now
`"pocket_data": [ { "pocket_type": "MAGAZINE", "rigid": true, "ammo_restriction": { "battery": 0 } } ],`
also has an error if pocket battery set to 0
battery will not charge when connected to grid (not a problem if i can get it to work as intended)


Maleclypse:
Is this ready for review?

flybye3o6:
> Is this ready for review?

one critical bug im tying to workout
cant start construction with no battery power, work around is to have it set to 3960 but if someone uses the jackhammer without it plugged in it will drain the battery which cant be recharged and then can no longer be used to start constructions again like "dig downstair" workaround for that is to set it back to using UPS ("use_action" DRILL is unaffected and works regardless)
electric forge appliance can start crafts without the required power and also checks grid power, so looking into how that works to see if that can be setup for items and constructions

errors (and potentially test changes) if i set the electric jackhammer to have no battery, need to be worked out

also one hurdle to get past with TOOLMOD
if TOOLMOD could update/overwrite "use_action" extension cables could be turned into battery mods and used on all electric tools
"modify attached equipment" is what GUNMODS use, could try to use it on tools for "Plug in / Unplug" and "Detach" so looking into that

flybye3o6:
looking at crafting.cpp
https://github.com/CleverRaven/Cataclysm-DDA/blob/a51439692d74f549ab4da28f1c348c706925aad5/src/crafting.cpp#L2020-L2056
can i make a "player_can_build"  for something checking map_inv for charges
https://github.com/CleverRaven/Cataclysm-DDA/blob/55f0271f0b25ef7dd6433c5a5e208c19a3e831a6/src/construction.cpp#L865-L878
am i looking in the right place

PR #66585, opened 2023-07-01
update to kerosene spawns and refillable lighter

Daved27hundred:
#### Summary
Features "Refillable lighters use kerosene instead of gasoline + touched up kerosene spawns"

#### Purpose of change

I was just gonna make the gasoline lighter use kerosene because I'm pretty sure it using gasoline is a relic from another time, before we had kerosene, but I noticed that kerosene doesn't spawn as much as I thought it would - not only that but it spawns in plastic bottles. like water bottles. I recall seeing tins of fuel in the Canadian Tire I once worked at, so I'll make it spawn in those.

#### Describe the solution

- change ammo type of the `ref_lighter` to `lamp_oil` and make it take both camphine and kerosene
- create the fuel tin and set it as the default container of kerosene
- add kerosene spawns to hardware stores and increase how often they spawn in outdoorsman/hunting stores

I had thought of making a new fuel for the `ammunition_type` kerosene but thought that would be too much hassle.

#### Testing

not done yet

#### Additional context

thanks Tonkatsu for helping me find resources for this


uap4:
Indeed, for liquid-fueled lighters, fuel that burns without soot is strongly recommended. But a good lighter can even work on cologne.
It would be interesting to make the possibility of refueling with different fuels. With some fuel, the lighter's fire would burn badly and quickly go out, like matches in the wind, and the lighter would deteriorate like a firearm due to black powder.

Also, lighters filled with liquid fuel have a drawback - the fuel evaporates all the time, especially when not vertical. Reminds me of tinderbox_on.

PR #66588, opened 2023-07-01
A bunch of random item reworks: 7

Karol1223:
#### Summary
None

#### Describe the solution

<details>
<summary>BLANKET</summary>

- Weight: 1123g -> 848g
- Volume: 15l -> 15.5l
- ``price_preapoc``: 55$ -> 15$
- ``to_hit`` values now use modifiers
- Disassembly yield decreased to account for weight decrease - still generates mass, will need to be addressed in a separate PR later on when I double check the weight of thread

</details>

<details>
<summary>WINE GLASS</summary>

- Weight: 161g -> 130g
- Volume: 262ml -> 420ml
- Longest side: N/A -> 20cm
- ``price_preapoc``: 14$ -> 13$
- ``to_hit`` values now use modifiers
- Disassembly yield decreased to account for weight decrease 
- Pocket max weight decreased from 1kg to 500g

</details>

<details>
<summary>CAN OPENER</summary>

- Weight: 158g -> 204g
- Volume: 75ml -> 364ml
- Longest side: N/A -> 21cm
- ``price_preapoc``: 2$ -> 9$
- ``to_hit`` values now use modifiers
- Disassembly yield increased to account for weight increase

</details>

<details>
<summary>CAST-IRON FRYING PAN</summary>

- Weight: 2628g -> 2427g
- Volume: 1335ml -> 2940ml
- Longest side: 36cm -> 41cm
- ``price_preapoc``: 45$ -> 20$
- ``to_hit`` values now use modifiers
- Disassembly yield decreased to account for weight decrease 
- Pocket max weight decreased from 1kg to 500g
- Pocket max volume decreased from 1l to 500ml

</details>

<details>
<summary>SPONGE</summary>

- Weight: 40g -> 17g
- Volume: 150ml -> 156ml
- ``price_preapoc``: 0$ -> 0.66$
- Removed the snippet referencing sea sponges - not only are sponges made of those very rare, they're also definitely not plastic by material - not material for a snippet

</details>

<details>
<summary>DISH TOWEL</summary>

- Category changed to ``other`` from ``spare parts``
</details>

### DIFF
--- a/data/json/items/generic/bedding.json
+++ b/data/json/items/generic/bedding.json
@@ -77,11 +77,11 @@
     "type": "ARMOR",
     "name": { "str": "blanket" },
     "description": "Hiding under here will not protect you from the monsters.",
-    "weight": "1123 g",
-    "volume": "15 L",
-    "price": 5500,
+    "weight": "848 g",
+    "volume": "15500 ml",
+    "price": "15 USD",
     "price_postapoc": 50,
-    "to_hit": -1,
+    "to_hit": { "grip": "none", "length": "long", "surface": "any", "balance": "clumsy" },
     "material": [ "cotton" ],
     "symbol": "[",
     "looks_like": "towel",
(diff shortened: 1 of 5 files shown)


anoobindisguise:
Why are blankets so big? You can usually compress them down pretty well.

Karol1223:
> Why are blankets so big? You can usually compress them down pretty well.

My experience folding blankets wouldn't really agree... though I do think that 15l+ is overkill, honestly, but I'm not sure how to better calculate it. The amazon product doesn't provide item thickness, so I assumed 0.8cm (which is a bit overkill, frankly, but in a haphazardly folded blanket a survivor would throw in their backpack and not a neatly folded one maybe it's not?)

Ruminator24:
Maybe we need a compressed blankets item, like with how cardboard boxes can be folded

PR #66589, opened 2023-07-01
Half chausses

sadenar:
#### Summary
Content "Added a lighter, less costly but less protective alternative to chainmail chausses"

#### Purpose of change
Adding a new chainmail armor type for feet along with XS/XL versions and all 5 grades, as well as adding them to new nested groups.

#### Describe the solution
Added the items, recipes and nested groups for these new armor pieces, I am however still too new to contributing to find out how to make it so that the items require the leather lace these require to be kept in place and not just drop off from your foot, 2 short laces (1 for each individual half-chausse) would probably be enough seeing as those are supposed to be 15cm long.
As I bumped up the "strap_small" ingredient for all the recipes by one to reflect higher leather requirement the lace would involve while being unaware the laces existed, anyone who wishes to fix my work by adding the lace(s) should probably make all those identical to the normal chausses again.

- Not adding the chausses.
- Choosing other weight/coverage values instead of settling on overall removing 25% on most values from the full chausses.
- Considered making a new, larger nested group they would occupy together with the groups for the normal chausses, but didn't feel confident enough to not mess it up.
- I think the current sublimbs put in there are not ideal to properly simulate how this would be useless to protect from attacks hitting the back of the foot, so there may be something to improve there.

#### Testing
Spawned the items in game, checked their values, used the recipes, checked that the nested groups were working correctly, asked for feedback on the coverage errors I was getting and after fixing those am no longer getting any error message while loading the game.

#### Additional context
This manga page was what mainly prompted me to add them, I'm under the impression our chausses do not go as far up as those, but the potential was there to make them so I made them.
inspiration
Ironskin (an online merchant site for chainmail armor) also seems to agree that such early chausses did exist historically.

### DIFF
--- a/data/json/items/armor/boots.json
+++ b/data/json/items/armor/boots.json
@@ -2702,7 +2702,7 @@
     "type": "ARMOR",
     "category": "armor",
     "name": { "str": "pair of mild steel chainmail half-chausses", "str_pl": "pairs of mild steel chainmail half-chausses" },
-    "description": "Customized chainmail half-chausses that only cover the front and sides of the foot, leaving the back and sole unprotected.",
+    "description": "Customized chainmail half-chausses that only cover the front and sides of the foot and are maintained in place by a zigzaging leather lace on the unprotected back.",
     "weight": "1505 g",
     "volume": "375 ml",
     "price": 3508,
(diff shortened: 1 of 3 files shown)


sadenar:
I looked again at what was already in the game and had a think about this and am of the opinion that with our current implementation of chausses and leggings as separate armor pieces, which seems to differ a lot from historical chainmail leg armor, it seems that it would be better for me to either:
- Create half-chausses AND "half-leggings", both of them having open backs.
- Consider the exposed ankle of the half-chausse as negligible and only add "half-leggings" with their backs open, which would be closer to the niche of armor I was trying to add (lighter, cheaper, less protective) while taking into account the existing implementation of chainmail armor in the game.

sadenar:
Closed by #66665

PR #66590, opened 2023-07-01
Reduce airdrop frequency

worm-girl:
#### Summary
Bugfixes "Reduce airdrop frequency"

#### Purpose of change

The timeline states that there's been a steady breakdown of society for months by the time the game starts, and that the social order has been breaking down worldwide. However, it also states that a national state of emergency was only declared on April 10th, 5 days before the portal storms began (which surely would have made airdrops difficult if not impossible) and 15 days before the government entirely lost control. The government would not have started mass relief efforts prior to that declaration, and those fifteen days would not have left a lot of room for humanitarian aid, given that they couldn't even finish filling in the mass graves.

While it makes sense that what's left of the government might airdrop supplies to its troops and civilians in an attempt to save lives, there are some issues with the current implementation:

1) There's not enough food to effectively stock the emergency shelters that people were supposed to be evacuated to. However the supply drops contain many thousands of protein bars each and were fairly frequent, occurring about once per OMT (Square kilometer?). Given that the entire world is in the middle of an ongoing food crisis, where are all these DaiZoom bars coming from? If an OMT is 10 KM squared, that's still like 49 billion bars in New England alone. Even a hundredth of that strains credulity.

2) We can assume that these supply drops were not meant to be secretive, if they were intended to reach civilians. There were and are a lot of very angry and very hungry people everywhere in the world who have access to radios and would be actively looking for these things. Given that we have a faction of raiders hoarding resources, they'd be a prime target.

3) Why would these supply drops even be made, considering how much food and water are readily available on the ground?

4) A state of emergency is a specific and special declaration that gives local and federal government agencies access to funds and powers required to enact wide-scale humanitarian aid efforts, and this was only declared on April 10th, suggesting an incredibly sluggish and ineffective response from the government. Given that by April 10th the military is tearing itself apart and machine-gunning civilians, who exactly is making the airdrops? Who's coordinating the supplies? Where are they coming from? Blob psychosis is supposed to be making people angry, stupid, and selfish.

5) Why airdrops? We didn't do mass airdrops for Hurricane Katrina, which was a total breakdown of the social order in a single location. We (eventually) sent personnel and civilians in to distribute relief packages, which were very late and woefully insufficient even during a time of stability and prosperity for the rest of the nation, when nobody's brains were being turned into goo by Azathoth and there wasn't a civil war underway. There is (or was) aid on the ground at the FEMA camps, evac shelters, and refugee center, I don't think these would have been deemed necessary until it was far too late to get them organized.

6) The frequency of these airdrops obviates the supposed scarcity of supplies that drives many of the game's quests, including many of those surrounding the Refugee Center and Tacoma Ranch. Why is anyone farming or buying jerky when they can just drive around for an hour or two and collect enough protein bars and bottled water to feed everyone for years? It's not like the roads are particularly dangerous in the first two weeks or so.

7) There would have been no airdrops made after April 25th, when the final message from the government about no help coming is broadcasted. This means that the last supply drops would have been made at minimum five days prior (and probably longer), likely with their locations broadcasted either to the general public or to the military. Many groups would have been competing for these supplies for a week or more even on the player's first day.

#### Describe the solution

I removed airdrops from forests entirely and sharply reduced their frequency everywhere else. I also made them more likely to be looted.

Removing them altogether, changing their contents to be more often things which are not game-solving, etc. This is a recent PR however and the contents themselves seem supported by research, so neither of these seem appropriate.

There are plans to eventually have areas become depleted over time as the ambient NPC population rolls through and loots stuff. At that point it may be a good idea to increase spawns a tiny bit and then have them slowly get looted by the Hell's Raiders or whoever.

#### Testing

Created a new world and wandered around, noting that there were not millions of calories of unguarded nonperishable food lying out in the open everywhere. On the rare occasion that a player discovers one, it will be a spectacular find and not a common and expected occurrence.


GuardianDll:
> drops contain many thousands of protein bars each
And how much people for how long period of time this bars could support people? What sound like a lot for 1 person, for a group of hundreds of people is literally nothing
All another questions are legit, and i support the changes in this PR

fairyarmadillo:
> And how much people for how long period of time this bars could support people? What sound like a lot for 1 person, for a group of hundreds of people is literally nothing

7000 protein bars is 2.8 million calories, which will last 100 people about two weeks. If there's 1 drop of these things every 10km, that's effectively solved world hunger. If they're more spaced out, then as you say it's not an issue.

PR #66592, opened 2023-07-01
This should add hm-12 to robofac trades

Vgoloshivskiy:
#### Summary
Content "Add hm12 to possible trades"

#### Purpose of change
add hm12 to trades because if it falls of or get run over by car it is forever lost

#### Describe the solution
copypasted code to add it to possible items that trader spawn

1.create new not prototype weapon hm13 and sell it / also if notes for favorite loot are enabled it should note on losing items if they fall off during grab escape or dropping them into car. /  an ability to order it like an HWP

#### Testing
didn't do any testing 0 idea how just lost rifle and couldn't find it so this is my hotfix of active save

#### Additional context
:weary: after i lose hm12 in new run

### DIFF
--- a/data/json/itemgroups/Locations_MapExtras/robofac_trade.json
+++ b/data/json/itemgroups/Locations_MapExtras/robofac_trade.json
@@ -10,6 +10,7 @@
       { "group": "robofac_gun_exodii" },
       { "group": "robofac_gun_kit" },
       { "item": "robofac_enviro_suit" },
+      { "item": "hm12" },
       { "item": "robofac_enviro_suit_casual" }
     ]
   },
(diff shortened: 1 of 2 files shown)


bombasticSlacks:
@John-Candlebury

John-Candlebury:
I think its reasonable to have a replacement for sale like this, you'll have to check to make sure the trade value is reasonable tho, Im not sure if its set to anything atm.

Vgoloshivskiy:
> I think its reasonable to have a replacement for sale like this, you'll have to check to make sure the trade value is reasonable tho, Im not sure if its set to anything atm.

i m pretty sure it cost 150 normal price as by me

PR #66593, opened 2023-07-01
Fixes bridgehead connections, fixes 100 bridges adjacent, but all specials cut off by water get no road

Procyonae:
#### Summary
None

#### Purpose of change
Specials use bridgeheads as valid connections
A ridiculous amount of bridges can spawn next to each other

#### Describe the solution
Limits road connections to everything but bridgeheads
Only cities can make bridges
However all specials cut off by water get no road

#### Additional context
Just posting this here as a point of curiosity


PatrikLundell:
Only cities can make bridges?
That would mean that there would be no way to get across water without having to go through a city, which means not only being unable to bypass cities, but also that you'd have to hunt down a city to get across any water.

While cities are often founded close to rivers, there are lots of bridges built across rivers outside of cities in the real world.

Procyonae:
> Only cities can make bridges? That would mean that there would be no way to get across water without having to go through a city, which means not only being unable to bypass cities, but also that you'd have to hunt down a city to get across any water.
> 
> While cities are often founded close to rivers, there are lots of bridges built across rivers outside of cities in the real world.
> 
> This also interacts very poorly with the decision to make unique locations actually unique prematurely.

I meant the roads cities gen between themselves, cities very rarely currently span rivers in game. Also if you read this is only to show it's possible I'm not suggesting this actually get merged with the current state of road and river gen, just showing what can be done with a small amount of JSON. This makes specials have to connect to roads on their side of the river which makes more sense and you don't end up with loads of nonsense bridges, admittedly there's no way in JSON to encourage more city bridges so the end number of bridges in busy areas is lower than realistic but if you test the changes you'll see that in general it is more realistic, there's also nothing stopping specific specials like globally unique ones forming bridges if you wanted them to be more accessible.

PR #66595, opened 2023-07-02
Add ability to make slime globs into slime samples

randomtyper:
#### Summary
Content "Makes slime samples craftable from slime globs"

#### Purpose of change

The mobs that drop the slime samples are rare, and the slime globs have little use for survivor. At the same time, I think it would be reasonable if there was a way to make the slime samples from slime globs.

Also, microcentrifuge didn't use power on use, so I fixed that.

Granted the concentration quality to fractional distillation apparatus (as distilling a liquid can make it concentrated).

#### Describe the solution

Adds a pretty pricey slime sample recipe that utilises gear used for creating primers. The globs can be obtained in quantity in labs, and gear requirement makes it less usable as main source of slime mutagen (at least, in the beginning).

Made microcentrifuge use charges on use.

Pretend that slime globs are not related to slimes. Continue to have no use for said globs. Removing battery slot from microcentrifuge.

#### Testing

None! For now.

#### Additional context

None.

### DIFF
--- a/data/json/items/tool/science.json
+++ b/data/json/items/tool/science.json
@@ -1203,6 +1203,7 @@
     "symbol": "n",
     "color": "white",
     "ammo": [ "battery" ],
+    "charges_per_use": 5,
     "flags": [ "WATER_BREAK", "ELECTRONIC" ],
     "qualities": [ [ "CONCENTRATE", 1 ] ],
     "pocket_data": [
--- a/data/json/items/tool/science.json
+++ b/data/json/items/tool/science.json
@@ -1149,7 +1149,17 @@
     "flags": [ "WATER_BREAK", "ELECTRONIC" ],
     "symbol": "R",
     "color": "white",
-    "qualities": [ [ "CONCENTRATE", 1 ] ],
+    "charges_per_use": 5,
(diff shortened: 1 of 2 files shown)


fairyarmadillo:
Looks great, but IMO this recipe should only be in RECIPE_CREEPY and RECIPE_MAIAR. Nobody outside of XEDRA knew of the existence of slimes nor any scientific process involving them.

Even if this is meant to be like a general scientific process which could logically be applied to slimes, having it in standard chem books may give the wrong impression.

randomtyper:
> Looks great, but IMO this recipe should only be in RECIPE_CREEPY and RECIPE_MAIAR. Nobody outside of XEDRA knew of the existence of slimes nor any scientific process involving them.
> 
> Even if this is meant to be like a general scientific process which could logically be applied to slimes, having it in standard chem books may give the wrong impression.

Read the commits closely. The sample recipe is written exactly where you suggested.

fairyarmadillo:
I'm sorry, I misread the proficiency lines as recipe books.

randomtyper:
> 

Happens with everyone, don't worry.

PR #58810, opened 2022-06-29
CR Skateboard vehicle 

Maleclypse:
#### Summary
Content "Skateboard vehicle"

#### Purpose of change
Adds a skateboard vehicle. I'm still working on it and I'm not sure that it won't require changes to vehicle c++ to work.  I'm fine to cancel this if it requires more change than I'm capable of currently.

#### Describe the solution
Working on adding all the parts needed.

#### Describe alternatives you've considered
None

#### Testing
TBD

PR #66596, opened 2023-07-02
Add skateboard vehicle

balinck:
#### Summary
Content "Adds new skateboard vehicle"

#### Purpose of change
Adds skateboards to the game. Previously (see #58810) it hasn't been possible to properly model skateboards as vehicles. So this PR also includes some code support to make the skateboards feel like skateboards.

#### Describe the solution
- Skateboards are added to the game as a foldable one-tile muscle-powered vehicle.
- A new vehicle flag "SEAT_REQUIRES_BALANCE" functions as a vehicle equivalent to the "REQUIRES_BALANCE" worn flag. If you take damage and fail the stability roll, you fall over, just like with rollerblades.
- "SEAT_REQUIRES_BALANCE" also applies a penalty to move_resist in shake_vehicle(). So you're much more likely to get thrown off your seat in the event of a collision.

It might be better to have a more generalized way to apply move_resist penalties to seats, like with a JSON property, so that we can get more realistic behavior out of bikes, wheelchairs, etc.

#### Testing

- Spawned a skateboard on a test character. Rode skateboard repeatedly into a parked van. At 8MPH, my character was being thrown off the skateboard on impact consistently. When I added the "Skater" and "Deft" traits to the character, they only wiped out around 12MPH.
- Proceeded to ride skateboard into some zombies a bunch of times. Getting hit while on the skateboard added the Downed status and knocked me off my board. Direct collisions with zombies knocked me off a lot, too.


Maleclypse:
Thanks for doing this!

Issue #66452, opened 2023-06-25
Removing an integrated pocket deletes its contents


Standing-Storm:
### Describe the bug

If an integrated armor pocket (such as the Holdout Pocket from Bombastic Perks) is ever removed, the contents vanish rather than falling out onto the floor.

### Attach save file

N/A

### Steps to reproduce

1) Start a game with Bombastic Perks enabled.
2) Using the debug menu, grant yourself the Holdout Pocket mutation
3) Place something inside the holdout pocket
4) Remove the Holdout Pocket mutation
5) Item is completely gone.

### Expected behavior

Item falls on the floor, the way it does if an update changes a pocket size or possible contents. 

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.3086 (22H2)
- Game Version: 7871941 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    Slowdown Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    No Rail Stations [No_Rail_Stations],
    Bombastic Perks [bombastic_perks]
]

### Additional context

I'm not sure this scenario can come up during normal play (maybe if you have a battery in an internal compartment and then ask Rubik to remove the internal battery CBM?), but I ran into it while working on Mind Over Matter

RenechCDDA:
Confirmed it happens in the Rubik scenario.

Standing-Storm:
Solved by #75565

PR #66597, opened 2023-07-02
[MoM] Replace current Telekinetic passive with Integrated-Pocket-based Field

Standing-Storm:
#### Summary
Mods "Replace the +10% carry weight current bonus with an expandable 'Lifting Field' "

#### Purpose of change

Ten percent bonus weight was always a placeholder. When I first started working on Mind Over Matter, I knew much less about CDDA ins and outs and after a while of trying a summonable bag, in the interests of moving things along I added Telekinetic Carry and said that one day I'd come back to it. Now I know enough to actually implement the solution I was originally looking for.

#### Describe the solution

Obsolete Telekinetic Carry and replace it with Lifting Field, a mutation that comes in multiple levels, each tied to an integrated pocket. Telekinetics start with the ability to carry 2 kg and, based on the the level of their highest telekinetic power, can meditate to increase that capacity. I originally had the bag auto-update, but thanks to https://github.com/CleverRaven/Cataclysm-DDA/issues/66452 I swapped it over to requiring a recipe, which I think is a better implementation anyway because even if the bug is fixed it ensures that nothing will ever fall out of the Lifting Field at an inopportune time. Thanks to that bug, the recipe currently contains a warning message to empty the Lifting Field before upgrading it. 

Doing nothing.

#### Testing

I tested the auto-replace EoC, the recipe (both from +1 power level and +many power levels), the use of the pocket, starting as a telekinetic, and all of them work as expected.


data/mods/MindOverMatter/mutations/psi_passives.json
```diff
@@ -98,14 +98,333 @@
...
+    "id": "TELEKINETIC_LIFTER_1",
+    "name": { "str": "Lifting Hand" },
     "points": 0,
-    "description": "You are using your psionic powers to ease your carry burden.",
-    "weight_capacity_modifier": 1.1,
-    "valid": false,
-    "player_display": true,
-    "purifiable": false
+    "description": "With a modicum of concentration you can carry a single item with your powers.  As your powers increase, you may meditate on it to expand its capacity.",
+		"valid": false,
```
GuardianDll, line 105: this one has lack of lint

PR #66601, opened 2023-07-02
Create new lable for Mind Over Matter mod

GuardianDll:
#### Summary
None
#### Purpose of change
New mod was created, i guess it would be nice to lable it separately
#### Additional context
@Standing-Storm if you're interested, i can also add you into reviewers.yml, so you'll get pinged into every pull request, that edit mind over matter mod

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -257,6 +257,9 @@
 "Mods: Xedra Evolved":
   - data/mods/Xedra_Evolved/**/*
 
+"Mods: Mind Over Matter":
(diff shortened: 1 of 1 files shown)


RenechCDDA:
I believe that the label also needs to be created *on github* by someone with relevant permissions?

Otherwise, looks good.

Standing-Storm:
> @Standing-Storm if you're interested, i can also add you into reviewers.yml, so you'll get pinged into every pull request, that edit mind over matter mod

I would appreciate that, thank you.

PR #66603, opened 2023-07-02
[Bombastic Perks] Fix for perk_olde_guns

GuardianDll:
#### Summary
None
#### Purpose of change
`perk_olde_guns` accidentally used `global_val` instead of `context_val`, which make it impossible to read the description of the trait
#### Describe the solution
replace `global_val` with `context_val`
#### Additional context
If you read the descriptions from another perks, their context variables would be used
image
If you did not, nothing would be written
image


bombasticSlacks:
Thanks for fixing this 😄

PR #66623, opened 2023-07-02
Improve output for JSON errors

irwiss:
#### Summary

None

#### Purpose of change

Json error output is more confusing than it can be(in my opinion); it breaks line at the error position making it harder for humans to correlate the printed output with the actual line in a file, instead this keeps the full line intact.

Most of the "viewers"; game debugmsg in curses or sdl mode, github actions, local terminals for tooling are able to display colors, and since we have the capacity for colors might as well use them.

For before/after examples see screenshots in text folds at bottom.

#### Describe the solution

* Stops breaking the line at cursor - print the full line, then print the cursor underneath, this is similar to how clang or other tooling prints errors ( with a slight discrepancy of empty lines if the line is super long, but I don't wanna complicate it by probing for terminal size, it shows correct column except for the github actions bug described in additional context section ).
* Colorize the output where possible - debugmsg can colorize via `<color_foo>` tags, on stdout unless redirected and in github actions the normal ANSI color escapes work.
* Make the cursor 3 bold arrows, instead of just 1 caret character.
* `make style-all-parallel` is redirecting stdout when running in a github action so `error_log_json_force_color` will force the color on when `CI` env variable is set ([github sets it by default](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)).
* Feed the filename to JsonTextIn in the format_main.cpp  - this replaces `<unknown source file>` with actual filename instead of printing filename separately

Flatbuffers unfortunately doesn't display the error context at all in certain circumstances (doesn't rewind stream? Not sure yet), but that's unrelated to this PR.
<details>
  <summary>Example for flatbuffers erroring out</summary>
  
  image
</details>

#### Testing

Mostly manual as this needs to have errors in the json to show up.

#### Additional context

Examples under folds

<details>
  <summary>Curses</summary>
  
  curses_old curses_new
</details>

<details>
  <summary>SDL</summary>
  
  tiles_old tiles_new
</details>

<details>
  <summary>make style-all-parallel</summary>
  
  style-all-parallel_old style-all-parallel_new
</details>

<details>
  <summary>math parser debugmsg</summary>
  
  image
</details>

Github actions has an issue with long lines ( see links below ) but this exists in old format as well as new, seems github will trim whitespace even in raw logs, running `make style-all-parallel` locally will show cursor at the correct column

[Buggy output in github actions for old error printout](https://github.com/irwiss/Cataclysm-DDA/actions/runs/5438431534/jobs/9889662343?pr=27)
[Buggy output for new error printout (no change)](https://github.com/irwiss/Cataclysm-DDA/actions/runs/5438428479/jobs/9889657684?pr=25)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -857,9 +857,9 @@ THIRD_PARTY_SOURCES := $(wildcard $(SRC_DIR)/third-party/flatbuffers/*.cpp)
 HEADERS := $(wildcard $(SRC_DIR)/*.h)
 TESTSRC := $(wildcard tests/*.cpp)
 TESTHDR := $(wildcard tests/*.h)
-JSON_FORMATTER_SOURCES := $(wildcard tools/format/*.cpp) src/json.cpp
+JSON_FORMATTER_SOURCES := $(wildcard tools/format/*.cpp) src/wcwidth.cpp src/json.cpp
 JSON_FORMATTER_HEADERS := $(wildcard tools/format/*.h)
-CHKJSON_SOURCES := $(wildcard src/chkjson/*.cpp) src/json.cpp
+CHKJSON_SOURCES := $(wildcard src/chkjson/*.cpp) src/wcwidth.cpp src/json.cpp
 CLANG_TIDY_PLUGIN_SOURCES := \
   $(wildcard tools/clang-tidy-plugin/*.cpp tools/clang-tidy-plugin/*/*.cpp)
 CLANG_TIDY_PLUGIN_HEADERS := \
--- a/src/json.cpp
+++ b/src/json.cpp
@@ -2053,6 +2053,14 @@ void TextJsonIn::error( int offset, const std::string &message )
     const int max_context_lines = 3;   // limits context length to this many lines
     const int max_context_chars = 480; // limits context length to this many chars
     stream->seekg( offset, std::istream::cur );
+
+    // if offset points to middle of a codepoint then rewind stream to codepoint start
+    // this is so utf8_width below has correct utf-8 to count length.
+    // pattern of 10xxxxxx means the byte is utf-8 continuation byte.
+    while( ( stream->tellg() > 0 ) && ( ( stream->peek() & 0xC0 ) == 0x80 ) ) {
+        stream->seekg( -1, std::istream::cur );
+    }
+
     // remember positions of several places to print a few lines of context
     const size_t cursor_pos = tell();               // exact position of error
     rewind( 1, max_context_chars );                 // rewind to start of line
--- a/src/json.cpp
+++ b/src/json.cpp
@@ -21,9 +21,12 @@
 #include "cached_options.h"
 #include "cata_scope_helpers.h"
 #include "cata_utility.h"
+#include "wcwidth.h"
 #include "debug.h"
 #include "output.h"
 #include "string_formatter.h"
+
+json_error_output_colors_t json_error_output_colors = json_error_output_colors_t::unset;
 
 // JSON parsing and serialization tools for Cataclysm-DDA.
 // For documentation, see the included header, json.h.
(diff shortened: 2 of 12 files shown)


andrei8l:
Can you please change the caret in the math parser's error message too?

https://github.com/CleverRaven/Cataclysm-DDA/blob/458a53eb98fb222620315b0105e39abcc08fd921/src/math_parser.cpp#L722

src/json.cpp
```diff
@@ -1917,57 +1950,58 @@ void TextJsonIn::error( int offset, const std::string &message )
...
+        // print error line
+        std::string buffer( end_of_line - start_of_line, '\0' );
+        stream->read( buffer.data(), end_of_line - start_of_line );
+        err << color_error << buffer << color_end << "\n";
+    }
+    {
+        // display a cursor at the position if possible, or at start of line
+        err << color_highlight;
+        if( const int padding = cursor_pos - start_of_line - 1; padding > 0 ) {
+            err << std::string( padding, ' ' );
```
Qrox, line 2002: I believe part of the reason the line was broken into two parts was that, 1. it allows seeing where the error is even if the caret is misplaced, because you won't know the exact relative width of characters in a terminal (or a text editor if the output is redirected) if there are unicode characters in the message, and 2. it allows to easier see where the error is if the text before the caret spans multiple lines.

For 1, `utf8_width` should be used to at least make it work in-game and with monospace font with the same relative width as the in-game code. For 2, I'm not aware of any better solutions than the previous way of breaking the line though.

Makefile
```diff
@@ -857,9 +857,9 @@ THIRD_PARTY_SOURCES := $(wildcard $(SRC_DIR)/third-party/flatbuffers/*.cpp)
 HEADERS := $(wildcard $(SRC_DIR)/*.h)
 TESTSRC := $(wildcard tests/*.cpp)
 TESTHDR := $(wildcard tests/*.h)
-JSON_FORMATTER_SOURCES := $(wildcard tools/format/*.cpp) src/json.cpp
+JSON_FORMATTER_SOURCES := $(wildcard tools/format/*.cpp) src/cached_options.cpp src/json.cpp
```
Qrox, line 860: I don't feel like it's a good idea to introduce unrelated game constants into the formatter, which might make it more difficult to solve source dependency in the future and cause problems if someone accidentally uses an uninitialized constant.

Maybe it will be better to add `error_log_json_force_color` to `json.h/cpp` instead. (A better solution would of course be disentangling this mess, but that's out of the scope of this PR.)

tests/json_test.cpp
```diff
@@ -243,10 +245,9 @@ TEST_CASE( "translation_text_style_check", "[json][translation]" )
             R"(    Suggested fix: insert " ")" "\n"
             R"(    At the following position (marked with caret))" "\n"
             R"()" "\n"
-            R"("…foo.)" "\n"
-            R"(       ^)" "\n"
-            R"(         bar.")" "\n" ),
-        R"("…foo. bar.")" ); // NOLINT(cata-text-style)
+            R"("…foo. bar.")" "\n"
+            R"(      ▲▲▲)" "\n" ),
```
Qrox, line 249: This shows that when there are unicode characters in the message, the previous method allows seeing the exact position of the error by breaking the line.

Makefile
irwiss: This sounds more correct, not sure why I was so fixated putting it together with the log format, moved it to json.h

tests/json_test.cpp
irwiss: The previous method didn't work for unicode either as most json code appears to assume all-ascii-content - it'll happily break line in the middle of a codepoint, e.g. place `"type": 两` in a json file and try to run any formatter or execute the game you'll see broken text:

image

I think optimizing for the edge case which doesn't get fully solved isn't going to benefit the main target group (json writers) that much, instead we can optimize for the common case; json errors with mostly latin or non-"duo"spaced/cjk content and leave the outliers where they are now (complex errors are usually resolved in discord anyway)

src/json.cpp
irwiss: Did the (1) - I had to pull a copy of `UTF8_getch` and `utf8_width` from cata_charset.h/cpp to json.cpp, otherwise it pulled output.h for remove_color_tags() and that pulled string formatting sources... and probably more dependencies, this will need a follow up PR to sort it out

For 2 - afaict the editors and fonts that support monospace "correctly" will work even for CJK glyphs (even though there's close to no CJK or double-width codepoints in json files), even the ones that don't quite work correctly can be made to work using a specific font, for example using Inconsolata ( free font with fixed pitch ) renders CJK correctly everywhere I tested

Qrox: > For 2 - afaict the editors and fonts that support monospace "correctly" will work even for CJK glyphs

There are "ambiguous width" glyphs which can be either 1 or 2 cells wide so it might be difficult to find a font that matches the in-game width exactly for every glyph, but I agree using `utf8_width` is good enough for most cases.

tests/json_test.cpp
Qrox: I think it is trying to read in the first character to determine the value type, so it assumes the character is ascii and breaks the line in the wrong place.

I don't quite follow your second paragraph though, but there are mod writers who writes text primarily in Chinese (and presumably other languages) so I think it's important that it handles unicode characters correctly.

irwiss: Yep it does, that's what i was trying to say in the second paragraph - instead of assuming the terminal/editor is broken and using line breaks as cursor (which performs worse than incorrectly placed cursor in case of problems) we could assume most users have their terminal correctly configured for their locale and will place the cursor at or very near the correct spot.

In current proposed code it does so when testing with some CJK, cyrillic and ukrainian symbols with pretty much 0 config, it's likely impossible to exactly match the width of every codepoint on every font but it should be close enough for most users

There's also the fallback option of using line/column numbers (some editors will even follow them automatically) or using the game to render which will make the width match exactly

src/json.cpp
```diff
@@ -1917,57 +2049,61 @@ void TextJsonIn::error( int offset, const std::string &message )
...
+    }
+    {
+        // print error line
+        std::string buffer( end_of_line - start_of_line, '\0' );
+        stream->read( buffer.data(), end_of_line - start_of_line );
+        err << color_error << buffer << color_end << "\n";
+    }
+    {
+        seek( start_of_line );
+        std::string buffer( cursor_pos - start_of_line, '\0' );
```
Qrox, line 2093: I think it might be safer to check if `cursor_pos` points to a starting byte of a utf8 sequence and increment it to point to the next starting byte if it doesn't, because otherwise it may lead to incorrect width and out-of-bounds access:

If `cursor_pos` points to the middle of a utf8 sequence here, `buffer` will contain an incomplete utf8 sequence which might cause `utf8_width` to calculate the wrong width. Furthermore, if the incomplete sequence contains only one byte, `UTF8_getch` will try to access the second byte without checking for the buffer length, causing out-of-bounds access (which, on a side note, sounds pretty grim considering the original comment in `catacharset.cpp` suggests it was originally from SDL2... hope they either fixed it or are only using it for c-style strings.)

irwiss: Went with rewinding to start of a codepoint (or stream start) instead - the cursor makes a bit more sense then

TextJsonIn::rewind seems to be a bit borked - `rewind( /* max_lines = -1 */, /* max_chars = 1 */ )` will rewind twice instead of once

PR #66633, opened 2023-07-03
palettized modular shelter

PatrikLundell:
#### Summary

None

#### Purpose of change

Make use of palette parameters to shrink the size of the JSON contents for the modular shelter.

#### Describe the solution

Use palette parameters to consolidate beds into a single set of blueprints (instead of one set per type of bed), as well as doing the same with the construction materials.

Add additional construction materials. As there's already a non standard "normal" wall in the mix the result wouldn't have been a standard set anyway.

Add logic to deconstruct furniture rather than just replace it. Considered to be out of scope.

Not make use of fbmc_modular_shack.json palettes for beds. Since the files are called "fbmc" it was thought to be reasonable to make use of this somewhat out of scope definitions.

#### Testing

Constructed the full set of facilities except the radio tower (too much work to test something that remains unchanged) using log, brazier, and makshift beds on the first pass, inspecting things as construction progresses and killing the game after the final inspection. Repeat with rock, fireplace, and regular beds but no other common things, then "standard", wood stove, and makeshift beds, followed by just the next construction material (and brazier, as the cooking facility is a requirement to unlock construction), repeated for the last material.

#### Additional context

Had to add flags to specify items are allowed, as the door of the first room replaces a cabinet that has items in it. It's possible some of the other rooms have the same issue, but the flag was added to all of them (and the player may well drop things anyway).

It seems someone has made down mattresses useless cluttering red herrings by not allowing them to be used in bed construction any more. I'm quite sure regular and down mattresses used to be interchangeable.


Knut-Aage-Hofseth:
#66457 changes down mattresses. It seems a down bed deconstructed into a regular mattress.

> Describe the solution
Create new construction options for making "down beds" and "down bunkbeds"
Remove down mattresses from the options for making "beds" and "bunkbeds"

PatrikLundell:
Thanks for the info. I appreciate the effort.

I disagree with the solution used, though, as that path leads to a multiplication of the number of kinds of items and thus recipes for functionally equivalent objects/furniture that then would have to be maintained separately when things used are changed.

The proper solution, in my view, would be for items/furniture/terrain to retain information of what went into their construction unless marked to forget it (as a new parameter) so the proper items could be extracted on deconstruction/smashing. That's a large undertaking, though.

An inferior alternative would be if multiple recipes could generate the "same" product and have logic that allowed the recipes to be collected under the same headline (with a suffix or some such differentiating the items under the hood), but you would still have the issue when multiple components entries logically could make use of multiple different alternative components (all the different alternatives for short ropes combined with all the various alternatives for stick, combined with...).

There's a large number of items that take "proper" and "makeshift" versions of items as their components, for instance, and I don't think we have a shortage of recipe glut.

Maleclypse:
Is this good to merge or does it need a change after the down mattress changes were merged?

PatrikLundell:
I think it's good to go.

Adding the new down mattress bed to fbmc_modular_shack.json's bed blueprint would affect all users of the bed modules downstream (all of them would need to specify the additional parameter value). However, this particular user doesn't use those, but only the palettes, so adding a palette that wasn't used (yet) in that file would be possible, although it would probably cause people looking at it some confusion over why it was present but not used locally.

PR #66638, opened 2023-07-03
Assorted tile rendering fixes

Qrox:
#### Summary
Bugfixes "Tile rendering fixes"

#### Purpose of change
Fix several issues I noticed about tile rendering. Each commit is explained below.

1. The code was drawing off-screen tiles when memorizing them after all on-screen tiles are drawn, which caused incorrect order of large sprites.
2. Fixing 1 revealed that the drawing range for on-screen tiles was incorrect, so sometimes the last row of sprites in isometric mode were drawn by the memorizing code rather than the main drawing loop, which lacks some of the main drawing loop's layers.
3. Item highlight were drawn without 3d height in isometric mode, so highlights in other z-levels were drawn in the current z-level.
4. Vision effects (the `lighting_xxx` sprites) were drawn along with terrain in the same tile before the 3d vision change, but drawn before all terrain sprites after the 3d vision range, which made the sprites have wrong occlusion orders (which only becomes apparent if you make the `lighting_xxx` sprites have extra height.)
4.1. The code was using the `hidden` sprite for unseen and unmemorized tiles, but `dark` for tiles outside maximum vision range, which caused inconsistent appearance of unseen tiles.
5. The game was using the `hidden` sprite for unseen tiles when boomered, but the `boomered_dark` sprite for tiles outside maximum vision range, which was inconsistent.
6. The code was iterating from the lowest z-level even if the maximum 3d depth is low.
7. Some map memory was not loaded when looking at faraway tiles in the west in isometric mode (i.e. towards the bottom left of screen).
8. None
9. You could not see the upper body of a hulk if its lower body is off-screen, a.k.a. the code did not consider the maximum sprite extent when determining the range of tiles to draw.
10. None
11. None
12. Some top tiles in lower levels are not drawn in isometric mode.

#### Describe the solution
1. Memorize off-screen tiles without actually drawing. This should also improve drawing speed a bit.
2. Unify the mapping between player and screen coordinates into functions and fix errors.
3. Add 3d height to item highlights.
4. Store information required to draw the vision effects and draw them along with terrain in the actual drawing loop.
4.1 See 5 below.
5. Use `hidden` for tiles outisde maximum vision range, and `boomered_dark` for unseen tiles when boomered.
6. Pre-calculate the minimum z-level before the main drawing loop.
7. Calculate the rectangular region containing the on-screen tiles and load map memory from the region, using the unified coordinate mapping code.
8. None.
9. Calculate the maximum sprite extent when loading a tileset, and use this extent to calculate the maximum range of tiles that can possibly be on-screen.
10. None
11. None
12. Calculate range of on-screen tiles for respective z-levels.

#### Testing
All changes worked. Tested multiple tilesets in isometric and non-isometric mode and the tiles were correctly drawn.

Also tested framerate benchmarks before and after these changes on three tilesets in a steel mill. The changes improved drawing speed a little bit for the isometric mode, especially when drawing a single level, whereas the speed of non-isometric mode seems unaffected overall apart from some random fluctuations.

| scenario | ultica-iso | ultica | smashbutton-iso |
| -- | -- | -- | -- |
| ground level (1) | 34.435 -> 38.508 | 43.339 -> 43.660 | 33.094 -> 35.786 |
| (2) | 34.283 -> 39.466 | 43.574 -> 42.854 | 33.247 -> 36.356 |
| 3 levels (1) | 24.990 -> 25.304 | 38.177 -> 36.342 | 19.124 -> 20.800 |
| (2) | 25.000 -> 25.289 | 37.725 -> 36.941 | 18.474 -> 20.639 |
| 3 levels & 4x zoom out (1) | 9.408 ->  9.539 | 12.537 -> 12.168 | 6.853 ->  7.671 |
| (2) | 9.198 -> 9.856 | 12.460 -> 12.166 | 6.989 -> 7.571 |

#### Additional context

### DIFF
--- a/src/cata_tiles.cpp
+++ b/src/cata_tiles.cpp
@@ -12,6 +12,7 @@
 #include <stdexcept>
 #include <tuple>
 #include <unordered_set>
+#include <variant>
 
 #include "action.h"
 #include "avatar.h"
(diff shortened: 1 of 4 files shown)


bombasticSlacks:
Great comments for all the new stuff 😄

Issue #65173, opened 2023-04-18
Enemy armor penetration is too high


anoobindisguise:
### Is your feature request related to a problem? Please describe.

Many enemies have ridiculously high flat armor penetration values. The issue is that any amount of flat armor penetration applies to each layer of armor individually, meaning basically anything less than an ESAPI plate or power armor get penetrated like they're not even there.
Example: you are fighting a Giant Wasp, which has an armor penetration of 25 on its sting. You are wearing Light Tempered Plate Armor and a Canvas Aketon and it stings your torso (the aketon has 10.5 cut prot and the cuirass has 22). Due to its armor penetration you have effectively no armor whatsoever and it does full damage. Its bite also has an armor penetration of 20, which means it still does 8 damage to you out of the total 10 its bite inflicts.
Other enemies also possess huge armor penetration values, although even low values such as 5-10 are still extremely strong because they make any armor that isn't individually extremely strong moot. Some examples:
- Brutish Amalgam (15 armor pen on bite)
- Arthropod Mutant (15 armor pen on bite)
- Camo Mutant (25 armor pen on bite)
- Root Runner / Leech Plant (5-15 random on EVOLVE_KILL_STRIKE)
- Unseen Hunter (10 on basic attacks and 30 on its peck)
- Pack Rat and Tunnel Runner (12 on bite)
- Wasp Queen (25/20 like above)
- Giant Wasp Queen (60 on sting and 30 on bite)
- Dire Wolf (10 on bite)
- Anything with the `impale` special attack (8 pen plus 0.5x multiplier)

This leads to armor being almost or entirely useless against certain enemy attacks to a highly illogical extent (you can wear 20 ballistic vests and a giant wasp would still sting you through all of them effortlessly) and forces the player to rely on defenses that bypass the absurd armor pen values, namely blocking, defensive martial arts and the ADS bionic. This also affects bullets - vs a m249 CROWs, anything weaker than 7 armor does nothing against 5.56 rounds, meaning the incidental armor you pick up from sources are vastly diminished since 6 pen becomes effectively 18 pen against 3 layers of armor.

### Solution you would like.

[...]

PR #66653, opened 2023-07-04
armor penetration gets used up like damage

anoobindisguise:
#### Summary
Bugfixes "flat armor penetration is spread across all armor layers instead of applying its full value to each"

#### Purpose of change
Partially address #65173
Armor penetration is broken, both because it has absurdly high values and also because it renders single armor layers useless (you could wear 20 individual hypothetical 1mm kevlar shirts and still take full damage from a 5.56 nato round, which has 6 armor penetration, but if you had a single 8mm thick kevlar shirt, it would stop the bullet almost entirely)

#### Describe the solution
Armor penetration is "used up" by the amount it penetrated a single armor layer. If there is a flat armor penetration in addition to a multiplier armor penetration, the actual defense code always applies flat before multiplier, so the same thing is true here (even if the attack has 0.5x armor mult, a 10-armor armor will always reduce the total flat penetration in the attack for the next layer by 10)

I have not thought of any.

#### Testing
Compiled and tested using a special Giant Wasp (mon_wasp) which had its normal attack reduced to 1 damage and its special attacks changed to always do 10 damage instead of a variable amount and wore increasingly large numbers of phelloderms to test its damage. The bite has 20 pen and does cut damage, and the sting has 25 and does stab damage. Phelloderm has 3 cut and 2.4 stab protection.
- At 6 or fewer phelloderms, I took full damage from each attack (10)
- At 7 phelloderms I have 21 cut and 16.8 stab prot. The bite did 9 damage and the sting did 10 damage.
- At 8 phelloderms, I have 24 cut and 19.2 stab prot. The bite did 6 damage and the sting did 10 damage.
- At 9 phelloderms, I have 27 cut and 21.6 stab prot. The bite did 3 damage and the sting did 10 damage.
- At 10 phelloderms, I have 30 cut and 24 stab prot. The bite did 0 damage and the sting did 10 damage.
- At 11 phelloderms, I have 33 cut and 26.4 stab prot. The bite did 0 damage and sting did 9.
- At 13 phelloderms, I have 39 cut and 31.2 stab prot. The bite did 0 damage and the sting did 3.
- At 14 phelloderms, I have 42 cut and 33.6 stab prot. The bite did 0 damage and the sting did 1.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -8843,6 +8843,8 @@ void item::mitigate_damage( damage_unit &du, const bodypart_id &bp, int roll ) c
 {
     const resistances res = resistances( *this, false, roll, bp );
     const float mitigation = res.get_effective_resist( du );
+    // get_effective_resist subtracts the flat penetration value before multiplying the remaining armor.
+    // therefore, res_pen is reduced by the full value of the item's armor value even though mitigation might be smaller (such as an attack with a 0.5 armor multiplier)
     du.res_pen -= res.type_resist( du.type );
     du.res_pen = std::max( 0.0f, du.res_pen );
     du.amount -= mitigation;
@@ -8853,6 +8855,8 @@ void item::mitigate_damage( damage_unit &du, const sub_bodypart_id &bp, int roll
 {
     const resistances res = resistances( *this, false, roll, bp );
     const float mitigation = res.get_effective_resist( du );
+    // get_effective_resist subtracts the flat penetration value before multiplying the remaining armor.
+    // therefore, res_pen is reduced by the full value of the item's armor value even though mitigation might be smaller (such as an attack with a 0.5 armor multiplier)
     du.res_pen -= res.type_resist( du.type );
     du.res_pen = std::max( 0.0f, du.res_pen );
     du.amount -= mitigation;
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -8843,6 +8843,8 @@ void item::mitigate_damage( damage_unit &du, const bodypart_id &bp, int roll ) c
 {
     const resistances res = resistances( *this, false, roll, bp );
     const float mitigation = res.get_effective_resist( du );
+    du.res_pen -= res.type_resist( du.type );
+    du.res_pen = std::max( 0.0f, du.res_pen );
     du.amount -= mitigation;
     du.amount = std::max( 0.0f, du.amount );
 }
@@ -8851,6 +8853,8 @@ void item::mitigate_damage( damage_unit &du, const sub_bodypart_id &bp, int roll
 {
     const resistances res = resistances( *this, false, roll, bp );
     const float mitigation = res.get_effective_resist( du );
+    du.res_pen -= res.type_resist( du.type );
+    du.res_pen = std::max( 0.0f, du.res_pen );
     du.amount -= mitigation;
     du.amount = std::max( 0.0f, du.amount );
 }


Venera3:
Broad strokes I'm for this, but I'd expect the multiplier to be applied before the flat AP - in your example the 10 AP should get reduced to 5 instead of zeroed out.

anoobindisguise:
> Broad strokes I'm for this, but I'd expect the multiplier to be applied before the flat AP - in your example the 10 AP should get reduced to 5 instead of zeroed out.

Yeah so how it works currently, is that the flat amount is applied to your armor, then the remainder is multiplied. So if you have 20 armor and face a 10 damage attack with 10 pen, and 0.5x multiplier, you take 5 damage if the multiplication is done after the flat reduction, and 0 if the multiplier is applied before the flat damage. IMO it should work like the former way, rather than the multiplier doubling the effectiveness of all flat armor penetration too.

Venera3:
Whatever you're doing on your attack that makes the armor half as effective should make armor half as effective, not half as effective unless it's about defeating armor penetration. 

Regardless, whichever you go with should be documented explicitly.

bombasticSlacks:
> 
    At 6 or fewer phelloderms, I took full damage from each attack (10)
    At 7 phelloderms I have 21 cut and 16.8 stab prot. The bite did 9 damage and the sting did 10 damage.
    At 8 phelloderms, I have 24 cut and 19.2 stab prot. The bite did 6 damage and the sting did 10 damage.
    At 9 phelloderms, I have 27 cut and 21.6 stab prot. The bite did 3 damage and the sting did 10 damage.
    At 10 phelloderms, I have 30 cut and 24 stab prot. The bite did 0 damage and the sting did 10 damage.
    At 11 phelloderms, I have 33 cut and 26.4 stab prot. The bite did 0 damage and sting did 9.
    At 13 phelloderms, I have 39 cut and 31.2 stab prot. The bite did 0 damage and the sting did 3.
    At 14 phelloderms, I have 42 cut and 33.6 stab prot. The bite did 0 damage and the sting did 1.

Gotta get you writing unit tests 😁

anoobindisguise:
So the issue is that in damage.cpp you have 
```
float resistances::get_effective_resist( const damage_unit &du ) const
{
    return std::max( type_resist( du.type ) - du.res_pen,
                     0.0f ) * du.res_mult * du.unconditional_res_mult;
}
```
So if I made penetration reduction apply after the multiplier, it would introduce an incongruity. I mean that can be rewritten too but as is right now, you need an armor with 15 resistance to tank a 10 damage, 10 arpen, 0.5x multiplier attack, not 20.

bombasticSlacks:
So I'm good to merge this as it is now, I think any math changes would need more comprehensive work. Just waiting on some form of documentation, it looks like none of this is documented at all right now, so adding it somewhere would be appreciated. A comment in the cpp file about how armor pen gets used up would make sense too I think.

anoobindisguise:
> Also, if mult applies before flat pen, there's no reason you wouldn't just get rid of the mult and double the flat pen in JSON. Under the current proposed system having a flat pen and mult leads to a different scaling system, where the remainder of whatever the flat pen couldn't defeat is multiplied by a fraction, which means you need less armor for your armor to start mattering, but your armor matters less, rather than your armor mattering the same amount, but needing more armor for it to start mattering. If that makes sense

I just realized this isn't true 100% of the time. Gonna put this in draft for a bit because that throws off my assumption for why this is this way

bombasticSlacks:
> I just realized this isn't true 100% of the time. Gonna put this in draft for a bit because that throws off my assumption for why this is this way

That's true but I think this is still a step in the right direction.

anoobindisguise:
OK so. I can change the line ` du.res_pen -= res.type_resist( du.type )` to be ` du.res_pen -= ( res.type_resist( du.type ) * du.res_mult )`.

Scenario1: 40 damage attack, 0.5x armor multiplier, 5 flat pen. You have a cuirass with 20 armor which is hit. Mitigation = 20 - 5 = 15, 15 *0.5 =7.5 protection, so you take 32.5 damage. Mitigation = 7.5. Arpen is reduced by 20 (becoming 0). This is how it would currently work.  If the change is made, instead it works like this: Mitigation still = 20 - 5 = 15 * 0.5 = 7.5, but arpen is reduced by 10, becoming 0 again. There is no difference.

Scenario2: 40 damage attack, 0.5x armor multiplier, 15 flat pen. You have a cuirass with 20 armor which is hit. Mitigation = 20 - 15 = 5 * 0.5 = 2.5, so you take 37.5 damage. Arpen is reduced to 0, because 20 armor > 15 flat pen. If the change is made, it instead works like this: Mitigation is still 2.5 total, but arpen is reduced by 10 instead and 5 flat pen remains to apply to any other layers. There was enough armor to still have some mitigation against the attack, but despite that there was some armor penetration that gets passed to the next layer. This strikes me as wrong.

Scenario3: 40 damage attack, 0.5x armor multiplier, 25 flat pen. You have a cuirass with 20 armor which is hit. Mitigation = 20 - 25 = 0. All 20 of the armor is used up, leaving 5 flat armor pen remaining and the full 40 damage still incoming.  If the change is made, instead it works like this: Mitigation still = 20 - 25 = -5 * 0.5 = 0, but arpen is reduced by 10, meaning 15 arpen gets through. The difference is instead of reducing incoming armor penetration by 20, it's instead reduced by 10.

PR #65549, opened 2023-05-09
Apply different messages per intensity level

Maleclypse:
#### Summary
Features "Apply different messages at different intensity levels"

#### Purpose of change
I want to apply different messages at different intensity levels. Mirroring how decay messages work. Required for #65311 

#### Describe the solution
I have cargo culted the decay message bool and removed the std:: for apply message.

Issue #66255, opened 2023-06-17
Crash when opening character info


SurFlurer:
### Describe the bug

Playing with an old save migrated to latest master, when I press `@` to see the character's info, the game crashes.

### Attach save file

https://github.com/CleverRaven/Cataclysm-DDA/files/10797597/Quakertown-trimmed.tar.gz from https://github.com/CleverRaven/Cataclysm-DDA/issues/63722#issue-1594082023

### Steps to reproduce

1. Load the world.
2. Ignore those warnings.
3. Press @
4. The game crashes

### Expected behavior

The game should not crash.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.3086 (22H2)
- Game Version: 0.G-2380-ga272beb2b3 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    Slowdown Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Magiclysm [magiclysm],
    No name () [],
    No name () [],
    SpeedyDex [speedydex],
    Stats Through Skills [StatsThroughSkills],
    Blaze Industries [blazeindustries],
    No name () [],
    Tamable Wildlife [Tamable_Wildlife]
]

### Additional context
Suspicious warnings generated while loading the world:
```
 DEBUG    : invalid effect type id rural_church_aware

 FUNCTION : obj
 FILE     : E:\Cataclysm-DDA\src\effect.cpp
 LINE     : 94
 VERSION  : 0.G-2380-ga272beb2b3

```
```
 DEBUG    : deserialized bionic id '' doesn't exist and has no migration

 FUNCTION : deserialize
 FILE     : E:\Cataclysm-DDA\src\bionics.cpp
 LINE     : 3144
 VERSION  : 0.G-2380-ga272beb2b3

```
```
 DEBUG    : No names for effect type, ID: 

 FUNCTION : disp_name
 FILE     : E:\Cataclysm-DDA\src\effect.cpp
 LINE     : 685
 VERSION  : 0.G-2380-ga272beb2b3

```

SurFlurer:
Looks like the translation code is trying to translate a nullptr.

SurFlurer:
The reason is the vector of translations for an effect's description is empty.

PR #66660, opened 2023-07-04
Update smoke effect

GuardianDll:
#### Summary
None
#### Purpose of change
Smoke filed apply different intensities of smoke effect on player. However, the effect contain no intensities, so it is not used. While here, polish the effect and make it better overall
#### Describe the solution
Now there is two effects, first respond for eye irritating, and can be countered with eye protection, and second for breathing issues, that can be countered by mouth protection
Each effect apply it's own set of effects, that progress the worse smoke field you happen to enter
Effect uses #65549 to write different messages for different intensities, and #66683 for multiple `immunity_data`s


Qrox:
While you are at it could you also address https://github.com/CleverRaven/Cataclysm-DDA/pull/66291#issuecomment-1596409761? I.e. the smoked effect should also mention eye irritation in addition to lung irritation.

GuardianDll:
That's actually the reason i didn't make it sooner: fd_smoke specify only `mouth` bodypart in immunity data
https://github.com/CleverRaven/Cataclysm-DDA/blob/a71a91179f84802a3c081ffd05339936355abd69/data/json/field_type.json#L505
If there was a way to split effect into moutn and eye parts, where eye effect checks against env stat of eyes, and mouth effect checks against env stat of mouth, it would work much better. But, from what i know, it's not possible, and i don't really know should i make this changes or wait when such thing would be possible

data/json/effects.json
```diff
@@ -1779,15 +1779,70 @@
...
+      "Your eyes tear and stings heavily.",
+      "Your eyes burn out of smoke, and you hardly see anything."
+    ],
+    "apply_message": [
+      [ "Your eyes mildly irritate.", "bad" ],
+      [ "Your eyes watering out of smoke.", "bad" ],
+      [ "Your eyes tear and stings heavily.", "bad" ],
+      [ "Your eyes burn out of smoke, and you hardly see anything.", "bad" ]
+    ],
+    "rating": "bad",
```
GuardianDll, line 1801:
```suggestion
```

data/json/effects.json
```diff
@@ -1779,15 +1779,70 @@
...
+      "Dense cloud of smoke enter your lungs.",
+      "You inhale an overwhelming amount of smoke."
+    ],
+    "apply_message": [
+      [ "You draw in a faint whiff of smoke.", "bad" ],
+      [ "Noticeable amount of smoke fills your lungs.", "bad" ],
+      [ "Dense cloud of smoke enter your lungs.", "bad" ],
+      [ "You inhale an overwhelming amount of smoke.", "bad" ]
+    ],
     "rating": "bad",
```
GuardianDll, line 1831:
```suggestion
```

data/json/field_type.json
```diff
@@ -454,11 +454,21 @@
         "translucency": 1,
         "effects": [
           {
-            "effect_id": "smoke",
+            "effect_id": "smoke_eyes",
+            "body_part": "eyes",
+            "intensity": 1,
+            "min_duration": "2 seconds",
+            "max_duration": "2 seconds",
+            "immunity_data": { "body_part_env_resistance": [ [ "eyes", 3 ] ] },
```
GuardianDll, line 462:
```suggestion
            "immunity_data": { "body_part_env_resistance": [ [ "sensor", 3 ] ] },
```

data/json/field_type.json
```diff
@@ -470,11 +480,21 @@
         "translucency": 10,
         "effects": [
           {
-            "effect_id": "smoke",
+            "effect_id": "smoke_eyes",
+            "body_part": "eyes",
+            "intensity": 2,
+            "min_duration": "7 seconds",
+            "max_duration": "7 seconds",
+            "immunity_data": { "body_part_env_resistance": [ [ "eyes", 5 ] ] },
```
GuardianDll, line 488:
```suggestion
            "immunity_data": { "body_part_env_resistance": [ [ "sensor", 5 ] ] },
```

data/json/field_type.json
```diff
@@ -486,11 +506,21 @@
         "translucency": 0,
         "effects": [
           {
-            "effect_id": "smoke",
+            "effect_id": "smoke_eyes",
+            "body_part": "eyes",
+            "intensity": 4,
+            "min_duration": "15 seconds",
+            "max_duration": "15 seconds",
+            "immunity_data": { "body_part_env_resistance": [ [ "eyes", 7 ] ] },
```
GuardianDll, line 514:
```suggestion
            "immunity_data": { "body_part_env_resistance": [ [ "sensor", 7 ] ] },
```

Kamayana:
I think this PR removing the `"smoke"` effect is causing crashes (#66980 #67005), as it seems like the game is looking for an effect that doesn't exist, and it doesn't look like there's a migration of `"smoke"` to the new `"smoke_lung"` (though I'm not sure if migrating smoke effects is actually possible).

SurFlurer:
A no-longer-exist effect will cause crash when the game wanna translate its descriptions / messages. Preventing this kind of crash will solve #66255.

PR #66639, opened 2023-07-03
Make horses milkable

linewriter1024:
#### Summary
Features "Make horses milkable"

#### Purpose of change

Horses, like other mammals, produce milk. [Humans drink this milk](https://www.york.ac.uk/news-and-events/news/2020/research/early-evidence-horse-milk-consumption/). This PR makes DDA horses produce milk like sheep, cattle, etc. already do.

#### Describe the solution

A lactating mare around the weight of a DDA horse might produce [11 liters or more](https://ker.com/equinews/high-quality-milk-essential-foals/) of milk per day. This PR makes horses milkable at a little less that half that, 5 L per day (half the production of a DDA cow), since DDA can't yet simulate all the conditions that go into actual milk production.

#### Describe alternatives you've considered

Not being able to ride around like some kind of post-apocalyptic Genghis Khan full of horse milk, horse cheese, and haggis.

#### Testing

Spawn horse, tame horse, milk horse. Receive 5L of milk.

PR #66663, opened 2023-07-04
reign in horse milk yield

kevingranade:
#### Summary
None

#### Purpose of change
#66639 was a bit optimistic about milk yield from horses, this reigns it in. 

#### Describe the solution
Drop production to a litre a day.

This is still about 4x too high as mares don't produce without a goal which means something like 3 months of production a year. 
Mares also produce a very different milk than cows,  so ideally this would be a different milk.

#### Additional context
https://www.npr.org/sections/thesalt/2018/07/12/627454097/mares-milk-for-health-europeans-look-to-horses-for-ancient-remedy
After some searching I was able to find a source for yield instead of raw production, which indicates about a litre per day.

### DIFF
--- a/data/json/monsters/mammal.json
+++ b/data/json/monsters/mammal.json
@@ -2040,7 +2040,7 @@
     "material": [ "flesh" ],
     "symbol": "H",
     "color": "brown",
-    "starting_ammo": { "milk_raw": 20 },
+    "starting_ammo": { "milk_raw": 4 },
     "morale": 20,
     "aggro_character": false,
     "melee_skill": 6,


uap4:
At the link "organic" production.

https://journals.nauka-nanrk.kz/bulletin-science/article/view/1457/1338
Lactation is more than 3 months, but the milk is really taken about in second, third and fourth months. About 5 liters are taken these days. 5*(3/12)=1.25 liters per day averaged per year.
This is without harm to the foal and only in 3 months. And to milk in the next months, about the same, then in total there will be up to two average liters per day for a year.

kevingranade:
Using an intensive farming process with extremely frequent milkings,  premium food for both the mare and foal, etc. The paper is remarkably unclear about whether yields are gross (including what the foal consumes) or harvested.  From what I can tell from some offhand comments it looks like harvested yield peaks at about 3-4 litres a day for about 3 months. That obviously doesn't come out to 2l a day year round.
Also averaging per year is misleading,  if you could average it like that it would be invaluable,  but you can't, instead you get seasonal peaks you have to preserve.

uap4:
I wanted to write a lot, but I'm tired of fighting against Google Translate.

Typically, milk production is indicated from a biological point of view. Not economic.

The link first indicates the full daily volume, but the volume of the taken part is also indicated, but it looks like a bad translation there. And in the end, too, there is the volume of all milk for the entire time and the volume of the selected part. The method here is the most common. Nothing extreme.

If do not feed the mares and foals, but leave them to graze, then I also saw this in a scientific journal. https://panor.ru/en/articles/milk-yield-and-quality-of-milk-of-mares-of-the-tuvan-breed/74623.html . But I did not find a free English version.
 In this case, 150-200 liters of milk are obtained and koumiss is produced from it. Lactation is 5 months, which means milking is 4 months. This is almost 1.5 liters every day for those 4 months.

And here is another article considering such farms from an economic point of view. There can see the proportions in the herd and how much koumiss is produced. https://doi.org/10.26897/1728-7936-2018-2-55-62 . But all this pile of materials is in Russian.

PR #66664, opened 2023-07-04
Repair Firearm Repair Kit Repair

ashGlaw:
#### Summary
Balance "Firearm repair kits used too much energy compared to its own components."

#### Purpose of change

It makes no sense that a soldering iron would use 1 charge per use, a Gunsmith Repair Kit uses 25, but a Firearm Repair Kit uses 100, given that the only electric tools in either kit *are* a soldering iron.

#### Describe the solution

Brought the costs down in line with the Soldering Iron. 

I considered that the cost difference was put there as balance, but it seems substantially less necessary now that the kits can be plugged into vehicles anyway.

#### Testing

I spawned in a firearms repair kit, a gunsmith repair kit, some batteries, and a Glock. I bashed things with the pistol until it was very damaged. I used the kits to repair the gun. Each used only 1 charge.


GuardianDll:
it looks like none of them have any battery-powered tools except soldering iron, so i recommend to make activation cost for both of them to be in line with iron

PR #66666, opened 2023-07-04
Riverside boats

harakka:
#### Summary
Content "Add boats to river banks"

#### Purpose of change
Add boats to riverbanks. Condition varies, with some of the boats still being in regular use, and others having been abandoned there for a long time.

#### Describe the solution
Place boats on river banks as map extras using nests to control what kinds of riverbank tiles boats can spawn on.

This has some downsides: riverbank overmap terrains are hardcoded. There are n/e/s/w rotations and then same for inner and outer corners. These hardcoded variations can't be targeted on regional settings level. Because of that, nested mapgen is used to make sure boats only spawn on straight riverbank segments (tile has riverbank neighbors on both sides). Allowing spawning on all riverbank tiles would result in them showing up in inner and outer riverbank corners off to the water in weird rotations, instead of spawning on the actual bank.

One downside of this approach is that we can't enable autonotes for the boats, because the actual placement on an eligible tile is determined by nested mapgen within the map extra mapgen. If river bank terrains get unhardcoded in the future, and I've been told they will likely be, I will update the mapgen to get proper mapnotes.

If lack of autonotes kills this, let me know and I'll happily re-draft this until river mapgen is unharcoded.

Every river bank tile has 1/100 chance of being considered for boat placement. This is still constrained by the nested mapgen rules, so for example inner and outer corners will never spawn a boat. Actual odds will therefore depend on river wriggliness. I can increase the odds if this feels tuned too low.

Have boats potentially spawn on all riverbank tiles, including the corners where they end up in some unexpected rotation off to the water, but I considered that too wonky looking.

Stuff I left out on purpose to keep scope minimal: making boats spawn together with reeds, adding more/better boat types, boats on lakes, more piers.

#### Testing
Run around as a debug character to expose a lot of shoreline, observe boats being present.

#### Additional context
Example of how often boats spawn with the current odds. Both sides of the.river were scouted. Every red circle is a boat:
boat_likelihood

I has a boat
image

🤘

### DIFF
--- a/data/json/mapgen/map_extras/riverside_boats.json
+++ b/data/json/mapgen/map_extras/riverside_boats.json
@@ -1,0 +1,154 @@
+{
+  "type": "mapgen",
+  "method": "json",
+  "update_mapgen_id": "mx_riverside_boat",
+  "object": {
+    "place_nested": [
+      { "chunks": [ "riverside_boat" ], "neighbors": { "west": [ "river_n" ], "east": [ "river_n" ] }, "x": 0, "y": 0 },
+      {
+        "chunks": [ "riverside_boat" ],
+        "neighbors": { "west": [ "river_n" ], "east": [ "river_e" ] },
+        "x": 0,
+        "y": 0
+      },
+      {
+        "chunks": [ "riverside_boat" ],
+        "neighbors": { "west": [ "river_n" ], "east": [ "river_s" ] },
+        "x": 0,
+        "y": 0
+      },
+      {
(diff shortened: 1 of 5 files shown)


Procyonae:
You could potentially get away with having the corner ones in the water with a fence post and a rope on the land, I wouldn't bother trying to make them work perfectly too hard bc the rivers need smoothing anyways

kilozombie:
This makes me think that driftwood and pieces of broken ships on the shores of the ocean/rivers would also be really cool.

Also, nice issue number!

Issue #66649, opened 2023-07-03
Allergen flags need to be updated to account for some existent materials


RenechCDDA:
### Describe the bug

![image](https://github.com/CleverRaven/Cataclysm-DDA/assets/84619419/bf8b27b2-cad5-43c5-bd4a-a13d402f866c)

Carnivore (missing bread, `dried_vegetable`):
https://github.com/CleverRaven/Cataclysm-DDA/blob/19858bf5bd4de6f75df3d96265d578a2b8dd4398/src/consumption.cpp#L155-L158

Vegan (missing cheese):
https://github.com/CleverRaven/Cataclysm-DDA/blob/19858bf5bd4de6f75df3d96265d578a2b8dd4398/src/consumption.cpp#L831-L835

Flags assigned here:
https://github.com/CleverRaven/Cataclysm-DDA/blob/19858bf5bd4de6f75df3d96265d578a2b8dd4398/src/item_factory.cpp#L3470-L3504

Any new allergen flags need to be defined in data/json/flags.json

### Attach save file

N/A

### Steps to reproduce

Debug carnivore and eat these materials ingame:

bread (found on caviar & egg sandwich)
dried_vegetable (not functionally used on any current items but exists so should be classified accordingly)

Debug vegan and eat these materials ingame:

cheese (not functionally used on any current items but exists so should be classified accordingly)

### Expected behavior

Carnivores shouldn't be able to eat non-meat materials

### Versions and configuration

- OS: Windows
    - OS Version: 10
- Game Version: c94dc03 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    Slowdown Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    No Rail Stations [No_Rail_Stations]
]

fyhannes:
<img width="1376" alt="Screenshot 2023-07-03 at 11 01 03 PM" src="https://github.com/CleverRaven/Cataclysm-DDA/assets/29219984/28566892-81e2-48ff-a90a-d79e982d234d">
Could you clarify what the issue is with Vegan and cheese? I spawned a character, debug-added Vegan and tried to eat a spawned Cheese but could not.

fyhannes:
Can close, resolved with #66669

PR #66669, opened 2023-07-04
Added new allergen flags and updated carnivore blacklist 

fyhannes:
#### Summary
Features "Added allergen flags for cheese, dried vegetables, and bread"

#### Purpose of change
#66649 Currently, carnivores can eat anything that has egg or milk since they've been whitelisted in `set_allergy_flags`. Also, materials such as "dried_vegetable", "cheese", and "bread" do not have allergy flags defined, nor are they accounted for in the carnivore and vegan blacklist. This means that food made out of bread could be consumed by carnivores and food made out of cheese can be consumed by vegans.

#### Describe the solution
Added the respective flags and updated the blacklists to account for these new allergens. Removed { material_egg , flag_CARNIVORE_OK }, and { material_milk, flag_CARNIVORE_OK } since the inclusion of such flags can inadvertently allow any carnivore to eat any item as long as its material is composed of egg or milk.

None

#### Testing
Steps: Debug Carnivore and try to eat Caviar and Egg. 
Expected Result: Player should be unable to consume Caviar and Egg.


Maleclypse:
So egg is in fact a major food for many carnivore species, such as snakes.   Snakes are obligate carnivores.  Milk sure remove that from obligate carnivores though not all obligate carnivores are lactose intolerant.

fyhannes:
```
{ material_milk, flag_CARNIVORE_OK },
{ material_egg, flag_CARNIVORE_OK },
```

I assume you're referring to the removal of these two flags above. The reason I removed it was because any food such as [Caviar and Egg t](https://github.com/CleverRaven/Cataclysm-DDA/blob/7920ed62f2b05e29248a943faeb94f7f8f54ba46/data/json/items/comestibles/egg.json#L979) that is made out of "bread" and "egg" should not be edible by carnivores, but the flag `{ material_egg, flag_CARNIVORE_OK },` will allow the Caviar and Egg to be eaten. To generalize it, as long as any food contains `milk` or `egg`, it doesn't matter what other materials it's made out of because the carnivore can eat it. You can test it by trying to eat [buttered_toast](https://github.com/CleverRaven/Cataclysm-DDA/blob/7920ed62f2b05e29248a943faeb94f7f8f54ba46/data/json/items/comestibles/bread.json#L261) or [buttered_toast_wheat_free](https://github.com/CleverRaven/Cataclysm-DDA/blob/7920ed62f2b05e29248a943faeb94f7f8f54ba46/data/json/items/comestibles/bread.json#L271) as a carnivore (without my changes). The alternative could be to just edit the json entries for each of these food types but I don't think it's future-proof.

Unless I'm mistaken, isn't the point of `set_allergy_flags` to assign allergens to specific food types? If that's the case, then removing `{ material_milk, flag_CARNIVORE_OK },` and `{ material_egg, flag_CARNIVORE_OK },` shouldn't be an issue because they aren't in the [carnivore_blacklist]({ material_milk, flag_CARNIVORE_OK },
{ material_egg, flag_CARNIVORE_OK },) and the carnivore_blacklist is what really controls the carnivore's diet.

dseguin:
Doesn't look like the new allergens are being added to any food items, which I'm pretty sure means that eating bread with a bread allergy does nothing. Are you planning to add those in a follow up PR?

fyhannes:
Is that not what `set_allergy_flags` is for?

dseguin:
You're absolutely right, sorry for doubting :+1:

fyhannes:
No worries! I'm happy to have other eyes on this.

PR #66672, opened 2023-07-04
Implement all-wooden fermenting vat (see #65632).

db47h:
#### Summary
Content "Add an all-wooden fermenting vat"

#### Purpose of change

As discussed in #65632, this is a rather dirty implementation of an all-wooden fermenting vat, providing an option for players without easy access to steel and plastic to build a fermenting vat entirely made out of wood.

#### Describe the solution

Regarding the crafting recipe, I took this research document as reference: https://journals.openedition.org/gallia/5242#tocto1n3
the 6 long sticks (spit along the length), should be enough to form about 12 wooden hoops holding the barrel staves together, as well as 6 short ropes cut in half to tie up each hoop together. Short sticks seemed too short. I might have been a bit heavy handed on the short ropes: while 12 short strings are clearly not enough, may be 24 would do. I deliberately did not add the option to use short cordage as I felt it would be too easy to build, but I have no issue with adding it. The wood chiseling requirement is simply for realism (the staves had grooves near the ends and had obviously been cut in shape with some kind of planer) and can be achieved with a stone chisel.

The recipe is placed in its own group so that players can choose which variant to build and not inadvertently waste sheet metal.

The code changes are too extensive to my taste since I had to add special case handling between `fvat` and `fvat_wooden` in order to switch the furniture between its empty and full state. In order to slightly simplify this, I've added static functions to this effect in iexamine.cpp and ideally they should be moved elsewhere so that they could be called from `activity_handlers::fill_liquid_do_turn()` for example.

However, as discussed in #65632, the plan is to add more fermenting containers (one of which will probably terrain), so we will end up making more changes, so these ad-hock functions and special case handling will very likely go away and I suggest to leave this as-is for the time being (famous last words).

Add an alternative recipe for the fermenting vat, but since furniture does not keep track of the components used to build it, deconstruction any vat would yield small metal sheets, which is not desirable.

#### Testing

Built wooden vat, filled with unfermented vinegar, advanced time by two weeks, removed contents. I did this with a partially filled vat and fully filled, also tested partial removal of contents upon completion of the fermenting process.


db47h:
I forgot to mention that I doubled the time required to build compared to the regular vat. This is purely a guesstimate, but it is meant to take into account the fact that the hoops need to be crafted (unless the original recipe considers that the small steel sheets need to be shaped as well?). In any case, 90 minutes felt too quick.

db47h:
Don't merge just yet. I did not update the disassembly yields in the last commit. Also after a bit of experimenting in Innawood, it turns out that not allowing cordage rope makes the craft almost as long as the sheet metal version. I'll need to add another version of the vat with cordage.

db47h:
It should be ready to merge, unless you want me to shrink the history into a single commit.

Maleclypse:
I've kicked the tests so hopefully it'll be ready for merge in a few hours once the required tests have passed.

Maleclypse:
> It should be ready to merge, unless you want me to shrink the history into a single commit.

We squash on merge usually, so you don't have to.

PR #66684, opened 2023-07-05
Fix: show slowed message on slow terrain

dseguin:
#### Summary
None

#### Purpose of change
Someone brought up that the "slow terrain" message was no longer displaying when walking over slow terrain. #66582 tried to fix a logic bug, but in the process made it so that the slow terrain message is only displayed when walking over fungal terrain.

#### Describe the solution
Instead of checking for "slow terrain AND fungal terrain", check for "slow terrain OR fungal terrain".


RenechCDDA:
Oops, I thought I checked that.

Evil evil boolean, it is. Thanks for the catch!

PR #66689, opened 2023-07-05
30gal_drum craft time fix

anoobindisguise:
#### Summary
Bugfixes "30 gal drum crafts in the correct time"

#### Purpose of change
30 gal drum accidentally was set to 60 s instead of 60 m (compare 80 m for 55gal drum)

#### Describe the solution
s->m

### DIFF
--- a/data/json/recipes/tools/containers.json
+++ b/data/json/recipes/tools/containers.json
@@ -330,7 +330,7 @@
     "subcategory": "CSC_OTHER_CONTAINERS",
     "skill_used": "fabrication",
     "difficulty": 8,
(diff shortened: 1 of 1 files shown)


Karol1223:
Can you do me a solid and apply this to the aluminum tank as well? I think I gave it some arbitrary numbers instead back then...

anoobindisguise:
> Can you do me a solid and apply this to the aluminum tank as well? I think I gave it some arbitrary numbers instead back then...

That one only has a blacksmithing recipe, not a welding one, so its time is nominally correct (not 60x faster)

PR #66696, opened 2023-07-05
make fuel bunker use a regular pocket

### LINKED ISSUE #66651: Weird things about fuel bunker
### Describe the bug

There are a bunch of inconsistencies about the fuel bunker, especially after the charcoal overhaul.

1. Filling this vehicle part up with coal, charcoal, and carbon electrode rods respectively, you can see that the displayed content volume is different, which is counterintuitive. See the screenshots below.
2. In the vehicle interaction interface, the content volume is displayed in two places, and the two places have different values. See the screenshots below.

### Attach save file

[vehtest-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11943746/vehtest-trimmed.tar.gz)

### Steps to reproduce

Load the save, and examine the vehicle nearby.

### Expected behavior

I expect they have consistent displayed volume, and consistent content volume between different kinds of fuel.

### Screenshots

2.5L coal lump VS 12.5L charcoal vs 500.0L carbon electrode rod.

12.0 L at left VS 12.5L at right.

2.0 L at left VS 2.5L at right.

- OS: Linux
    - OS Version: <unknown>
- Game Version: 19858bf5 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

anoobindisguise:
#### Summary
Bugfixes "fuel bunker storage uses a pocket instead of a magazine"

#### Purpose of change
Fixes #66651

#### Describe the solution
Make it use a 50 liter pocket (base size is 55 liters) for storage like a 60 liter tank does, rather than a magazine. 

Carbon electrode rods probably shouldn't be usable fuel but I decided not to fix that for now as that's a separate issue.

### DIFF
--- a/data/json/items/vehicle/fuel_storage.json
+++ b/data/json/items/vehicle/fuel_storage.json
@@ -16,13 +16,6 @@
     "ammo_type": [ "charcoal" ],
     "price": 3000,
     "price_postapoc": 100,
-    "pocket_data": [
-      {
-        "pocket_type": "CONTAINER",
-        "rigid": true,
-        "max_contains_volume": "50 L",
-        "max_contains_weight": "100 kg"
-      }
-    ]
+    "pocket_data": [ { "pocket_type": "CONTAINER", "rigid": true, "max_contains_volume": "50 L", "max_contains_weight": "100 kg" } ]
   }
 ]
--- a/data/json/items/vehicle/fuel_storage.json
+++ b/data/json/items/vehicle/fuel_storage.json
@@ -11,9 +11,7 @@
     "category": "veh_parts",
     "weight": "22000 g",
     "volume": "55 L",
-    "capacity": 2000,
     "//": "mods will need to expand",
-    "ammo_type": [ "charcoal" ],
     "price": 3000,
     "price_postapoc": 100,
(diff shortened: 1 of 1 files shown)


RenechCDDA:
> Carbon electrode rods probably shouldn't be usable fuel

Hah, because they have charcoal as a material? That's a fun exploit.

anoobindisguise:
so this is completely fucky. without some fairly extensive c++ rewrites, the fuel bunker has a really hard time functioning.
- If it can share coal and charcoal randomly, the menu doesn't report its fuel correctly (for example if you have 10L charcoal and then add 5 coal to it, it now says you have 2L coal, although it still has the total accurate number of charges). And again whenever the magazine changes between coal and charcoal the engine dies and you have to restart it.
- It cannot be a liquid tank, it just does not work. The refueling menu is hardcoded to only work with liquids and even if it weren't it still couldn't parse having two different kinds of fuel in the tank.
- If you make the magazine a holster so that you can't mix fuel, the game will still let you try but throw an error every time it does and fail to mix the fuel. Maybe the magazine could be used to accept different charges of each? Like coal: 50000 and charcoal: 10000 I'd have to find a specific example in code of that happening. In any case the fuel bunker literally cannot function if it's not a magazine.

PR #66697, opened 2023-07-05
soapy water can't be drunk

anoobindisguise:
#### Summary
Bugfixes "you can't drink soapy water anymore"

#### Purpose of change
Soapy water purely was negative, so there was no reason for it to be drinkable except for you to consume by accident.

#### Describe the solution
INVALID type.


TheEskir:
Bit late to the party, only saw this in the changelog.

Why is this a problem? You can also drink bleach/ammonia - I think wr should be able to drink more random liquids, rather than less. Make gasoline drinkable.

PR #66698, opened 2023-07-05
Add antacid tablets

anoobindisguise:
#### Summary
Content "add a tablet equivalent to heartburn medicine"

#### Purpose of change
We had pepto-bismol but no tums.

#### Describe the solution
Adds tablets anywhere heartburn medicine can be found. It's basically the same thing but a pill instead of a liquid. Convenient for the maniacs who like to mix all their pills together in one single container.

Add a chemical synthesis recipe for it, as it's pretty survivor doable.
I gave it 1 joy because I was inspired to do this by a bottle of berry-flavored tablets I had which tasted sorta good but it doesn't have to have any.


TheShadowFerret:
>Add a chemical synthesis recipe for it, as it's pretty survivor doable.

Do it 👀

Maleclypse:
Do you want this merged or wait til you add the recipe?

PR #66709, opened 2023-07-06
Various perf improvements for moving the firing cursor

### LINKED ISSUE #64765: Performance regression related to limb scores when using gcc

SurFlurer:
#### Summary
none

#### Purpose of change
Address #64765 .

#### Describe the solution
This can be reviewed by commits.

1. Speed up Character::has_flag() by caching whether bionics has certain json flags. The reason I didn't do the same thing for other aspects ( mabuff, effect, trait... ) is that tracking their changes is beyond my capabilities. Effects are inherited from `Creature`, mabuffs are basically effects, and traits are deeply entangled with enchantments.
2. Reduce dupliate calls of get_character_parallax() for performance. `effective_dispersion()` just calls `get_character_parallax()` and adds a number to it, so I manually inlined that func to improve perf. For heavily modified guns, there can be many gunmods, but there will only be two cases for `get_character_parallax()`, so cache the result and reuse when possible.
3. Move computational heavy operations outside the aiming loop. To do this, I created 2 structs to cache variables that are actually constant during the aiming loop, and made a func to initialize those variables. Then, I passed them down the call chain.

Venera once talked about caching results of `get_modifiers()`, but the same reason for not caching flags for traits, mabuffs and effects prevented me from doing so. I couldn't find out when to refresh the cache to not break swimming tests.

#### Testing
All tests passed.

Some perf data:

Before: 
Around 60% to 80% of cpu cycles are consumed by `get_limb_scores()` when moving the firing cursor.
https://user-images.githubusercontent.com/22912139/244914636-cd491b34-06c3-48d0-9b2b-315fb44d53de.png

After: 
With a late-game character, aiming with a HWP+9:
image

image

So most cpu cycles are consumed by UI drawing, which is expected.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -929,6 +929,9 @@ double Character::fastest_aiming_method_speed( const item &gun, double recoil,
     const float light_limit = 120.0f;
     bool laser_light_available = target_attributes.range <= ( base_distance + per_cur ) * std::max(
                                      1.0f - target_attributes.light / light_limit, 0.0f ) && target_attributes.visible;
+    // There are only two kinds of parallaxes, one with zoom and one without. So cache them.
+    std::vector<std::optional<int>> parallaxes;
+    parallaxes.resize( 2 );
     for( const item *e : gun.gunmods() ) {
         const islot_gunmod &mod = *e->type->gunmod;
         if( mod.sight_dispersion < 0 || mod.field_of_view <= 0 ) {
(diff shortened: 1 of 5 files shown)


src/character.cpp
```diff
@@ -892,8 +892,17 @@ int Character::point_shooting_limit( const item &gun )const
     }
 }
 
+aim_mods_cache Character::gen_aim_mods_cache( const item &gun )const
+{
+    parallax_cache parallaxes{ get_character_parallax( true ), get_character_parallax( false ) };
+    auto parallaxes_opt = std::make_optional( std::ref( parallaxes ) );
```
BrettDong, line 898: I think you're creating dangling references here.

PR #66717, opened 2023-07-06
Buffed range of .22LR rifles

BjoHart:
Added some extra range to the varmint rifle, small game rifle and the American180

#### Summary

Balance "Buffed range of .22LR rifles"

#### Purpose of change

The two .22LR rifles were hardly usable for anything in my eyes.
Here is a table with the values of 3 weapons. Once with effective range in real life and once with the range in fields in the game.

Wepon_________RL_____Fields
M16___________550m____36
Varmint rifle_____175m____13
CompoundBow___60m____18
Describe the solution

I have given the  .22LR rifles a range bonus.
In my opinion, the rifles should have a better place in the game.

Wepon_________RL_____Fields
M16___________550m____36
Varmint rifle_____175m____“24”
CompoundBow___60m____18

#### Describe the solution

Buffed range of .22LR rifles

None

#### Testing

Made a character
Gave him the small game hunting rifle
Tested the shoot distance
Works fine

#### Additional context

None

### DIFF
--- a/data/json/items/gun/22.json
+++ b/data/json/items/gun/22.json
@@ -7,6 +7,7 @@
     "name": { "str": ".22 caliber submachine gun" },
     "description": "A dramatically uncommon automatic weapon, making use of high-capacity pan magazines and firing the low power .22 calibre cartridge: an unusual ammunition choice for a submachine gun.  With negligible recoil on account of its modest cartridge and a suppressively high rate of fire, a burst of .22 LR rounds from this little machine gun can be best likened with a swarm of hornets… an incredibly angry swarm of hornets.",
     "variant_type": "gun",
+    "range": 7,
     "variants": [
       {
         "id": "american_180",
@@ -60,6 +61,7 @@
     "name": { "str": "small game rifle" },
     "description": "A venerable .22 caliber lever-action rifle, with a 19-round tube magazine and a classic design that dates back over a century.  Generally used for the hunting of small game like rabbits, the strict taxonomic classification of zombies as 'small game' is debatable, but this decades-old weapon can still be counted upon to help manage the undead population.",
     "variant_type": "gun",
+    "range": 11,
     "variants": [
       {
         "id": "marlin_9a",
@@ -205,6 +207,7 @@
     "name": { "str": "varmint rifle" },
     "description": "A lightweight and modular pistol caliber hunting carbine, firing the low power .22 LR cartridge.  Before the Cataclysm, a mix of high customizability, cheep ammunition, and low recoil made this a very popular varmint rifle amongst civilian shooters.",
     "variant_type": "gun",
+    "range": 11,
     "variants": [
       {
         "id": "ruger_1022",


BjoHart:
Is there anything i can do with the workflows?
Im bloody new to github and i can't find anything that explains it to me...

GuardianDll:
First time contributors need someone from the side to start their tests, someone with specific permissions. You just wait when such person start them for you

CatWithFourEyes:
Range compression for guns is /12, so, as example, effective range of 120m will give you 10 tiles ingame. And ~doubling of range, will set it on same level with M1 Carbine (25 range, ~300m effective fire distance, 1300J of muzzle energy), or with .50 Beowulf/.450 Bushmaster cartriges on short barrels, in terms of range.
I'm not sure, it's a good idea. 22LR is weak cartrige after all.

BjoHart:
> Range compression for guns is /12, so, as example, effective range of 120m will give you 10 tiles ingame. And ~doubling of range, will set it on same level with M1 Carbine (25 range, ~300m effective fire distance, 1300J of muzzle energy), or with .50 Beowulf/.450 Bushmaster cartriges on short barrels, in terms of range. I'm not sure, it's a good idea. 22LR is weak cartrige after all.

A simple range compressionrate is good for most of the weapons. But if a rifle is useles with this compression, i think someone has to adjust this.

I'm using a compound bow in the game, and the fact that the bow shoots more tiles then the .22LR rifle makes me mad. 
Played the game a while with the range of the rifles set to 24, and i dont think they are overpowerd. You need 5-6 shoots even for a standard zombie.

Zireael07:
https://archeryheaven.com/compound-bow-range/ says bow effective ranges are 60/70 yds, so 54-60m according to an online converter.

TLDR bow ranges need to be nerfed, not rifle buffed

CatWithFourEyes:
> TLDR bow ranges need to be nerfed, not rifle buffed

Also, bows use different from guns system of damage calculation, based on "what you can hunt with it" and making damage of x10 crit comparable with this animal HP ingame.

CatWithFourEyes:
> A simple range compressionrate is good for most of the weapons. But if a rifle is useles with this compression, i think someone has to adjust this.
> 
> I'm using a compound bow in the game, and the fact that the bow shoots more tiles then the .22LR rifle makes me mad. Played the game a while with the range of the rifles set to 24, and i dont think they are overpowerd. You need 5-6 shoots even for a standard zombie.

It doesn't need to be "balanced", and making precedents just "because it's useless" (and it's not, 22LR rifles can be silenced to ZERO loudness, and with 22LR accuracy and firerate - extremely good at night raiding in early/early-mid game) - isn't good idea. You see "rifle", but ammunition comes first, and 22LR is mostly good for training and small game. And gamedesight goal is making thing work close to RL as possible, with engine limitations.
I'm not saying, that "decomperssinon" of range for guns is bad. But not with current reality bubble size.
Bows and guns are in somewhat different categories, even they DMG calculated with different approaches, so can't say about "why they have X range", but link above giving very different numbers, and "estimated effective range" imo, about "you can make precise shot at hunt". Still, can't say about bows, better to wit for someone with knoledge of this part of game.

BjoHart:
True to most of your arguments Cat and Zireael.
But if you nerf bows in range, you can't hunt anything with them.
With the compound bow you already can't hunt small game, even at night they flee long before you get in range. You can hunt a deer in the dark, but if the range is only one or two tiles less, they also run away...

CatWithFourEyes:
> But if you nerf bows in range, you can't hunt anything with them.

i'm not talking about the bows (no offence) and, i understand your point in this case. But bows not equal with guns in current implementation, and i see them as different _type_ of ranged weapon, Like, no one blaming, that grenades can be tossed for ~10 tiles. But it's still "ranged" attack and 10 tiles = 120m for guns. 
Can't say, that bows and X-bows are good currenly, but also, not sure, that thay need to reworked **in current state of the game**. If reality bubble become, like 120 tiles - yes, definately. Currently? Not sure.

fairyarmadillo:
The bow's ballistic performance is a bit different than the gun's, and its weird range is partly a product of that. It can technically fire out to 18 tiles, but you are generally not going to do any damage past 6 or so. Reducing the bow's maximum range would likely make it only usable one or two tiles out unless its dispersion was also sharply reduced. This would unfortunately make it do waaay more damage, so it's kind of stuck in a weird spot.

Guns on the other hand can still do decent damage at their farthest range because they're not relying on a 10x crit modifier to do damage, effectively having way less falloff in terms of damage. So yes a bow can technically select a farther tile than a .22 to shoot to, but it's not going to kill anything at that range, and a fix would probably mean completely overhauling bows in general.

BjoHart:
Ok ok, forget the bows.

Compare the range of the .22LR rifles to handguns. They are about the same.
Shouldn't a rifle have more range than a handgun?

CatWithFourEyes:
> Shouldn't a rifle have more range than a handgun?

generic rifle Vs generic handgun? Yes. 
But even 9x19Para have at least 2x more muzzle energy, than 22LR. 
Rifle or not - isn't SO matter. First thing - cartrige. And 22LR use very light bullets, they are bad at conserving energy. As example, 9x19 bullets have more weight, faster, AND typycally, batter ballistic coefficient, so it not only better at hitting things, but at rateining energy to (so it losing speed slower, than 22LR).
Most of handgun ammunition _designed_ to kill or at least incapacitate humans. 22LR is training and small game, cheap and simple ammunition (yes, it can kill. But it would be much harder with 22lr, than with 9x19). 

TLDR: No, it shouldn't.

randomtyper:
The problem with this PR is that "Varmint rifle" (shown as example in table) covers a quite wide variety of calibers in-between big game rounds and rimfires. A .22 LR is a varmint. A .223 Remington is a varmint round (guess bald apes are varmints too, heh). A .220 Swift is a varmint round. One is weaker than 9 mm Para, one is civilian 5.56x45 NATO and one is a necked-down 6mm Lee Navy with bullet speed of 1200+ m/s.

https://en.wikipedia.org/wiki/Varmint_rifle

Some proper sources on specific weapons and their effective ranges (preferably ones represented in game) would help a lot with approval of this PR. While i do not need much persuasion to agree with that full-size rifles should have longer range than pistols, if not by much, the other members of the audience... You see. 

I hope author sees this and takes necessary actions to reinforce his argument. It is his PR and if I was him, I would not count on someone doing the work of providing proofs for me.

PR #66722, opened 2023-07-07
Jsonify monster flags

dseguin:
#### Summary
None

#### Purpose of change
When folks would like to add new flags for monsters, usually as part of a JSON content addition, they need to add those flags to the hardcoded enum in mtype.h. Since many people working on JSON are not comfortable compiling the game, this is a pretty serious obstacle.

The other part of this is that mod authors can't distribute mods with new monster flags unless those new flags get incorporated into the game itself.

#### Describe the solution
Replace the hardcoded `m_flag` enum with a JSON defined `"monster_flag"` object. The object is dead simple, only an "id" and a "type", since all that's needed is an id to point to for things like damage type immunity or effects that apply to specific monster flags.

The flag definition:
```JSON
{
  "id": "IRONWROUGHT",
  "type": "monster_flag",
  "//": "Immune to cold iron and does cold iron damage instead of any cutting damage",
}
```

Usage of the flag for damage immunity:
```JSON
{
  "id": "xe_cold_iron_cut_damage",
  "type": "damage_type",
  ...
  "immune_flags": { "character": [ "IRONSWORN" ], "monster": [ "IRONWROUGHT" ] }
}
```

#### Testing
There are several tests that rely on monster flags. If they pass, then the transition should be pretty seemless.

Also tested in-game by playing a couple runs with and without mods.

##### Profiling
Set up a save with the character standing in a lab, and profiled while waiting 5 minutes in-game. Profiled with `perf`, looking at `monster::has_flag`.

**Before, using hardcoded enums**
prof_before

**After converting to string_ids/int_ids**
prof_after2

**After removing the debug monster flag stats**
prof_after3

So it looks like it does carry a significant performance impact for areas with many creatures, which is unavoidable, but it's getting better with every optimization.

#### Additional context
TODO:
- [x] Profile performance
- [x] Optimizations
- [x] Some more testing


dseguin:
For mergers: Just undrafting to let the test suite run in its entirety. Please don't merge until all tests are green

PR #66726, opened 2023-07-07
Obsolete the survivor headlamp

Karol1223:
#### Summary
None

#### Purpose of change
The survivor headlamp, with the power of survival skill and soldering an amplifier onto a flashlight, created a headlamp that was better in every aspect than the regular headlamp.

#### Describe the solution
Obsolete it and everything connected


bombasticSlacks:
Recipes need a migration too

Karol1223:
> Recipes need a migration too

Are you sure? I could swear irwiss took care of that months ago. We don't have a current-cycle recipe migration/obsoletion file.

bombasticSlacks:
hmm it looks like you are right #64012

PR #66729, opened 2023-07-07
[CR] Very rough draft of a maybe better tutorial

NetSysFire:
#### Summary
None

#### Purpose of change

The tutorial we currently have is very lacking and I think a more practical approach is better.
This would be supposed to happen after the tutorial we currently have is finished by the player. Basically "Step onto this trap to get teleported into the next area", which also wipes the players inventory.

This much more open approach is *supposed* to highlight that the game gives you the tools, but in the end you are the ones to decide what to make with them. However, the guiding signs should prevent frustration caused by "I do not know what to do next and I am stuck."
I mainly have posted this here already so this can get discussed.

#### Describe the solution

So, I made a tiny python script to not make the rectangles by hand.

```python
def is_boundary(x, y, curx, cury):
    if curx == 0 or cury == 0 or curx == x or cury == y:
        return True
    return False

def gen_rect(x, y, xoffset=0, yoffset=0, border="tutorial_rock", filling="tutorial_field"):
    for coordy in range(0,y):
        for coordx in range(0,x):
            mapname=filling
            if is_boundary(x-1, y-1, coordx, coordy):
                mapname=border
            if mapname:
              print('{ "point": [ '+ str(coordx + xoffset) + ', ' + str(coordy + yoffset) + ', 0 ], "overmap": "' + mapname + '" },')

gen_rect(8, 8, filling=None)
gen_rect(6, 6, xoffset=1, yoffset=1, filling="tutorial_forest_thick", border="tutorial_forest")
gen_rect(8, 8, xoffset=8, filling="tutorial_forest")
```

image

I edited the results to add the landing area (sign1), the gate, the second sign and the shack.
The signs can and maybe should be replaced with these fancy popup traps. Because this is an open area, it will be much harder for the player to trigger it though since the trap is invisible. And yes, I am aware of our keybind variables which are used by the aforementioned traps, but I was not sure if they and color would work in signs and it is a detail which does not matter in this initial stage yet.

There is no mechanism to unlock the locked metal doors yet, but it would be something like "There are monsters in the next area. Proceed?" or "Wield your weapon of choice, then retry.", "A long pointy stick is an adequate weapon.  Gates unlocked."

A handler for player death would also need to be implemented, so the area would be reset and the player respawned.

As suggested earlier, wiping the inventory also means that they player will have no bag. Maybe a blanket sling would work really well here.

You usually start in an evac shelter, so why the heck do I drop a new player into a safe wilderness? Because dropping the player into an annotated evac shelter instead is basically duplicating early game content and making it even more boring. And even if you are naked and have literally nothing on you, you can build your way up by just using the resources in forests.
And foraging to train your survival and to make some initial gear is something you do in evac shelters, too.

#### Testing

There is one error I still have not figured out:

```
 DEBUG    : No mapgen terrain exists for "tutorial_practical_part".

 FUNCTION : void overmap_terrains::check_consistency()
 FILE     : src/overmap.cpp
 LINE     : 995
 VERSION  : 47f8c6e
```

#### Additional context

This is in a very early stage and I am unsure how much I am can work on it due to my current personal situation.
But the basic incomplete roadmap for this would be like this:

- [x] Initial peaceful "explore the forest" area
- [x] Shack inside a forest with weak zombies and some more "advanced" supplies
- [ ] Maybe an actual house or similar with some normal zeds and also things like pits, to aid kiting.
- [ ] Vehicles, basically "There are fast enemies up ahead.  Use this bicycle or, if you desire, try and build an even better one from the parts over there."
- [ ] Grand finale: An enclosed, small city with lots of gameplay mechanics highlighted and real death reactiviated. As in "You are now ready.  Permanent death has been reactivated and up ahead is a chaotic area with lots of fun things to find and do.  Upon death you will be dropped into the main menu, where you have can have an unrestricted and even more fun playthrough." - Fun things: Mini nukes, death mobiles, one of those fancy mech suits, stronger monsters.


bombasticSlacks:
> Mini nukes, death mobiles, one of those fancy mech suits, stronger monsters.

Mini nukes got nixed, but yeah, I think this is a great progression. Putting players in practical environments makes sense and gets them used to trying new things. It might be worth boosting skills as this progresses too so that they can take on the city confidently.

I-am-Erk:
This is something I have some thoughts on but can't comment now, I'll try to return

kevingranade:
I'm not against you pursuing this,  but the direction that I think fits the game better is a mixture of achievements and missions that can be layered on top of any relatively safe start that guide the player through the basics of movement,  resource gathering,  crafting,  combat, driving,  etc.

Issue #66510, opened 2023-06-28
Cannot make colored text in npc dialogue


gettingusedto:
### Describe the bug

I once had an idea to color the one part of the pyromaniac mission text to highlight the fact that you need to bring gasoline. after i covered said part of the text with `<color_yellow>` & `</color>` and ran the game to check it, it does not appear yellow.

### Attach save file

N/A

### Steps to reproduce

1. cover a npc text with the aforementioned `<color_(colorname)> [NPC_Text] </color>` in the json.
2. run the game and check said npc text
3. it does not get colored.

### Expected behavior

The npc dialogue text being colorable.

### Versions and configuration

- OS: Android
    - OS Version: Manufacturer: samsung; Model: SM-A710F; Release: 7.0; Incremental: A710FXXU2CRC2; 
- Game Version: 30f23ec [32-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Slowdown Fungal Growth [no_fungal_growth],
    Portal Storms Ignore NPCs [personal_portal_storms],
    Magiclysm [magiclysm],
    Xedra Evolved [xedra_evolved],
    Aftershock [aftershock],
    cable stuff [cablestuff],
    LuckySlotmachine [luckyslotmachine]
]

ZeroInternalReflection:
I've been unable to replicate this.  Adding colour tags to npc dialogue in the json appears to work properly.  In the tests below, "Emergency" and "Panic" had red tags added, "Help/Tutorial" had yellow added.  The text colour gets changed in both the dialogue selection panel and the message history panel.

![Talking_to_ally](https://github.com/CleverRaven/Cataclysm-DDA/assets/89038572/a05abc8d-50d0-437c-b636-56db2e38df0a)
![Talking_to_pc](https://github.com/CleverRaven/Cataclysm-DDA/assets/89038572/b5329473-aeca-4f4c-b848-b5573faa6f0b)


>NPC dialog window uses multiline_list::print_line which strips color tags from the output text (and does some other color manipulations):

It took me a minute to remember what was going on there, but the remove_color_tags() line is not actually used in the dialogue screen.  It was added as infrastructure so that I could convert some of the character creation screens to multiline_list (which hasn't happened yet, and probably needs some separate work).  The only colour change that multiline_list should be doing to the text is highlighting the selected line.


[...]

PR #66733, opened 2023-07-07
Make portal storm ending more apparent to the player

Daved27hundred:
#### Summary
None

#### Purpose of change

Some people were unaware that accepting the offer from the weird portal storm creature would end your game. 

#### Describe the solution

This PR adds a note to the dialogue option so the player knows what to expect when they choose it.

Maybe make it more apparent in the dialogue itself?

### DIFF
--- a/data/json/npcs/EOC_talkers/portal_storm.json
+++ b/data/json/npcs/EOC_talkers/portal_storm.json
@@ -89,7 +89,7 @@
     "responses": [
       { "text": "No ", "topic": "TALK_PORTAL_STORM_DO_FOR_ME" },
       {
-        "text": "Will it.",
+        "text": "Will it.  (This will end your game.  Do not choose this option if you wish to continue playing.)",
         "topic": "TALK_PORTAL_STORM_WILL_IT",
         "effect": { "give_achievement": "escaped_the_cataclysm" }
       },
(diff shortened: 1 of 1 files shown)


RenechCDDA:
I think it might be better to simply add an extra line to the *dialogue* (not response) very strongly suggesting it, and not, you know, bludgeoning the player over the head with the consequences.

So with some(warning but not spoiling) color and italicization...
`<color_yellow>*You get the feeling there's no coming back from this decision.*</color>`

Daved27hundred:
I don't think you can colour text in dialogue.

GuardianDll:
You can't, according to #66510

RenechCDDA:
Yeah just confirmed that myself. Still, for immersive stuff like this it needs to be handled pretty carefully. 

@Ramza13 Notifying you so you can provide any input, if desired?

Daved27hundred:
I'll gladly accept any changes, dialogue code is not my expertise

Ramza13:
So I'm conflicted on this.  I added the option as a surprise and to screw you over a bit.  If you know its coming there is no real point to it existing.  Maybe thats a bad design and I can see why people wouldn't like it, but I also feel the mood of cataclsym is such that accepting an obviously messed up offer from a strange voice is something you shouldn't do.

data/json/npcs/EOC_talkers/portal_storm.json
```diff
@@ -89,7 +89,7 @@
     "responses": [
       { "text": "No ", "topic": "TALK_PORTAL_STORM_DO_FOR_ME" },
       {
-        "text": "Will it.",
+        "text": "Will it.  (This will end your game.  Do not choose this option if you wish to continue playing.)",
```
Maleclypse, line 92:
```suggestion
        "text": "Will it.  'Your gut feels like this is jumping out of a plane without a parachute.'",
```

Something like this?

Issue #66434, opened 2023-06-24
MELEE_DAMAGE enchantment should affect bashing damage to tile


alp7292:
### Describe the bug

by using magic in magicalysm you can deal insane damage in melee but that doesnt affects your bashing damage on tile.
 for example you can break the reinforced glass with homewrecker but not with pipe mace but if you use magic to multiply its damage by 1.5 you should be able to break it.

### Attach save file

not needed

### Steps to reproduce

get Gravity Control and bash a terrain and realize it deals same damage.

### Expected behavior

MELEE_DAMAGE enchantment should affect bashing damage to tile

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19045.3086 (22H2)
- Game Version: 7070a09 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    <color_cyan>Arcana and Magic Items</color> [Arcana],
    Aftershock [aftershock],
    Bionic Professions [package_bionic_professions],
    Blaze Industries [blazeindustries],
    Magiclysm [magiclysm],
    Xedra Evolved [xedra_evolved],
    MST Extra [MST_Extra],
    DinoMod [DinoMod],
    <color_cyan>Arcana</color>/Aftershock Patchmod [Arcana_aftershock_Patch],
    <color_cyan>Arcana</color>/DinoMod Patchmod [Arcana_Dinomod_Patch],
    <color_cyan>Arcana</color>/Magiclysm Patchmod [Arcana_Magiclysm_Patch],
    Bombastic Perks [bombastic_perks],
    Translate Complex Dialogue [translate_dialogue],
    No Rail Stations [No_Rail_Stations],
    sees-player, run, crouch, bleed icons, retrodays [sees_player_retro],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    Slowdown Fungal Growth [no_fungal_growth],
    Tamable Wildlife [Tamable_Wildlife]
]

RenechCDDA:
Need to tie the enchantments into the bash function

https://github.com/CleverRaven/Cataclysm-DDA/blob/1b983260a71eb6df6b9109eb18b3c3cd0255887b/src/map.cpp#L2915-L2956

fyhannes:
Merged with #66735.

alp7292:
ok

PR #66735, opened 2023-07-07
MELEE_DAMAGE and ITEM_DAMAGE_BASH enchantments should affect bashing damage to tile

fyhannes:
#### Summary
Bugfixes "Smashing now incorporates any MELEE_DAMAGE and STRENGTH enchantments"

#### Purpose of change
#66434. STRENGTH and ITEM_DAMAGE_BASH enchantments should influence SMASH capabilities but currently only takes in arm_strength and the weapon's bash damage.

#### Describe the solution
Use `calculate_by_enchantment` to help contribute to the amount of smash damage done.

I initially thought about just linearly adding the STRENGTH and MELEE_DAMAGE bonuses to the smash damage, but I figured that `calculate_by_enchantment` was better suited to handle the smash damage calculation.  

#### Testing
1. Create a world with Magiclysm mod. Set character with strength of 4.
2. Locate retirement home and teleport to outside. There should be reinforced glass where the home faces the retirement home  parking lot.
3. Debug spawn pipe_mace and add TACTILE_KINESIS
4. Cast TACTILE_KINESIS and wield pipe mace
5. ~Terrorize senior citizens~ Smash reinforced glass until it breaks.

Reinforced glass has a bash_min of 40 and a pipe_mace has a bash rating of 24, so casting TACTILE KINESIS should allow the player to smash the window with a STRENGTH of 4

#### Additional context
Screenshot 2023-07-07 at 1 02 44 PM
Here's an example of reinforced glass breaking. My character has 4 strength, 4 dexterity, and has the GRAVITY_CONTROL effect. My character is also wielding a pipe mace and generic survivor clothing.


GuardianDll:
Strength doesn't influence it? I thought it add points to your character strength, which affect the bash output?
Anyway i think ITEM_DAMAGE_BASH should also be added here

fyhannes:
> Strength doesn't influence it? I thought it add points to your character strength, which affect the bash output?
Yeah it does. I'll remove the STRENGTH enchantment calculation I added since STRENGTH enchantments will affect `cur_str` which is used in calculating the bash output. Thanks for the catch.

>Anyway i think ITEM_DAMAGE_BASH should also be added here
Good point on ITEM_DAMAGE_BASH, I'll add that as well.

src/handle_action.cpp
```diff
@@ -741,19 +741,20 @@ static void smash()
...
+        smashskill += mon->mech_str_addition() + mon->type->melee_dice * mon->type->melee_sides;
         mech_smash = true;
-    } else {
-        smashskill = player_character.get_arm_str();
-        if( player_character.get_wielded_item() ) {
-            smashskill += player_character.get_wielded_item()->damage_melee( damage_bash );
-        }
-    }
+    } else if( player_character.get_wielded_item() ) {
+        smashskill += player_character.calculate_by_enchantment(
```
bombasticSlacks, line 751: Why only if the player is weilding a weapon?

fyhannes: If you mean "why do I only factor in MELEE_DAMAGE enchantment if the player is wielding a weapon", then you're right, this is a mistake. I should be linearly adding smashskill by the bash rating of the weapon they're currently holding and then perform the enchantment calculation afterwards.

I think there are two ways I could do it from there: the first is to call 

```
smashskill = player_character.calculate_by_enchantment( smashskill, enchant_vals::mod::MELEE_DAMAGE,
                  true );
smashskill = player_character.calculate_by_enchantment( smashskill, enchant_vals::mod::ITEM_DAMAGE_BASH,
                  true );
```

or, I could do something like

```
int melee_damage_enchantment =  player_character.calculate_by_enchantment( smashskill, enchant_vals::mod::MELEE_DAMAGE,
                  true );
int item_damage_bash_enchantment = player_character.calculate_by_enchantment( smashskill, enchant_vals::mod::ITEM_DAMAGE_BASH,
                  true );
smashskill += melee_damage_enchantment + item_damage_bash_enchantment
```

I'm leaning towards the second one but would like to hear your thoughts.

bombasticSlacks: I think #1 is easier to read 😄

src/handle_action.cpp
```diff
@@ -741,19 +741,20 @@ static void smash()
...
-        if( player_character.get_wielded_item() ) {
-            smashskill += player_character.get_wielded_item()->damage_melee( damage_bash );
-        }
-    }
+    } else if( player_character.get_wielded_item() ) {
+        smashskill += player_character.get_wielded_item()->damage_melee( damage_bash );
+    }
+    smashskill = player_character.calculate_by_enchantment( smashskill, enchant_vals::mod::MELEE_DAMAGE,
+                 true );
+    smashskill = player_character.calculate_by_enchantment( smashskill,
```
dseguin, line 755: Is the assignment operator intentional? Looks like it should be a `+=` instead

fyhannes: Yeah, it's intentional. `calculate_by_enchantment`'s first argument is the value we want to modify, and then it returns that value after adding/multiplying enchantment bonuses to it.

dseguin: I see, cheers :+1:

PR #66738, opened 2023-07-07
Document more fields for our monsters

RenechCDDA:
#### Summary
None

#### Purpose of change
Naughty contributors add json data but don't document their fields

#### Describe the solution
I went through monstergenerator.cpp and added anything that I didn't already find in the docs. Had to look up where some things were used but that wasn't too bad.

Cleaned up the explanation for attack_cost while I was in there since the current one wasn't very helpful.

Removing `status_chance_multiplier` because it's just a bit too long to fit in the column and nobody will miss it (it is currently unused)

#### Testing
Docs change only


bombasticSlacks:
The table didn't have leading and trailing "|" so my editor didn't pick it up as a table to modify so that extra character will have to survive.

PR #66746, opened 2023-07-08
sort first time contributors for test approval

casswedson:
#### Summary
None

#### Purpose of change
@Maleclypse requested this feature, add a `new contributor` label to first time contributors so their PRs don't get lost

#### Describe the solution
```bash
# fetch the number of merged prs of the author that triggered this workflow run
author=$(jq -r '.pull_request.user.login' $GITHUB_EVENT_PATH)
pr_count=$(gh pr list --state merged --author $author --json number | jq 'length'
# set the total count as an output we can later reference
echo "pr_count=$pr_count" >> $GITHUB_OUTPUT

# it the count is 0 aka this pr needs workflow approval from a maintaner label it as 'new contributor'
# not totally acurate from a naming point of view but that's okay
if: steps.pr-check.outputs.pr_count == 0
run: gh issue edit "$ISSUE_NUMBER" --add-label "new contributor"
```

#### Testing
this time it's interactive; make a pr to this repo to see this working https://github.com/casswedson/issue-labeling

~~draft cause I'd make the label beforehand (yellow is my favorite color btw) and we could review the label name itself~~

~~how this works: the code checks if you have merged prs in this repo, if not you are a new contributor, slap a label on it - I am the owner of that repo but I do not have any merged prs so I count
https://github.com/casswedson/issue-labeling/pull/25 runs on https://github.com/casswedson/issue-labeling/commit/98e1b9f9bd1d77ac8537b0aac3dae54354ebfa34~~

### DIFF
--- a/.github/workflows/label-first-time-contributor.yml
+++ b/.github/workflows/label-first-time-contributor.yml
@@ -1,0 +1,30 @@
+name: Sort new contributors for test approval
+
+on:
+  pull_request_target:
+    types:
+      - opened
+
+env:
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  ISSUE_NUMBER: ${{ github.event.issue.number }}
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Check if user has any merged PRs in this reposiory
+        id: pr-check
+        run: |
+          author=$(jq -r '.pull_request.user.login' $GITHUB_EVENT_PATH)
+          pr_count=$(gh pr list --state merged --author $author --json number | jq 'length')
+
+          echo "Debug: $author with $pr_count merged PRs."
+          echo "pr_count=$pr_count" >> $GITHUB_OUTPUT
+
+      - name: Label contributors with no merged PRs
+        if: steps.pr-check.outputs.pr_count == 0
+        run: |
+          gh issue edit "$ISSUE_NUMBER" --add-label "new contributor"


.github/workflows/label-first-time-contributor.yml
```diff
@@ -0,0 +1,38 @@
...
+              ...context.issue,
+              creator,
+              state: 'all'
+            })
+
+            const labelToAdd = 'new contributor';
+            const issues = await github.paginate(opts)
+
+            for (const issue of issues) {
+              if (issue.number === context.issue.number) {
```
Qrox, line 30: Unless I missed something, this iterates through a list of all issues created by the PR opener and adds a label to the PR if it is in the list, which is equivalent to just adding the label to every created PR.

casswedson: I honestly don't know what to say; it seems to work here
https://github.com/ankidroid/Anki-Android/blob/89441d7936cee653d61fbd42881d2e5be2ce361e/.github/workflows/label.yml#L154-L184
https://github.com/ankidroid/Anki-Android/pull/14118 the action runs here and does not label it as from a new contributor
https://github.com/ankidroid/Anki-Android/actions/runs/5489823104/jobs/10004466562?pr=14118#step:2:4

.github/workflows/label-first-time-contributor.yml
```diff
@@ -0,0 +1,44 @@
...
+                continue
+              }
+
+              if (issue.pull_request) {
+                return // Creator is already a contributor.
```
Qrox, line 35: This only adds the new contributer label to the first PR of an author*, but the author can create multiple PRs before any of them is merged, in which case all of these PRs will require approval. I think the correct logic is to add the label if the author does not have any merged PR when a PR is created.

(* Or, strictly speaking, this only adds the label if it's the author's only PR when the workflow runs, so if the author creates another PR before the workflow runs, which can be several hours after the PR is created if github action is busy, none of the PRs will have the label added.)

It would also be nice if the label is removed after the workflows are approved, maybe by adding a new workflow that only runs with approval which removes the label.

casswedson: I'd need help sorting those things out

Qrox: You can refer to the [`octokit`](https://octokit.github.io/octokit.js/v18/) documention on how to list PRs and check their merged status, but after some thought I think it's very tricky to get it right because I don't know how exactly github requests/dismisses workflow approvals, and the workflows can also be started with a delay and in different orders than their creation order, which makes it even more difficult to analyze.

I also thought about using a dummy workflow that triggers on pull request creation and another workflow that triggers on the former workflow's creation/start using the [`workflow_run` event](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run) to probe a PR's approval status, but it also turned out to be trickier than I had expected because the event data for the `workflow_run` event does not contain the exact id of the PR that triggered the dummy workflow, only the PRs that have the same head commit as the commit that triggered the dummy workflow, which may include other PRs.

Perhaps it'll be more sane if we just settle for a new contributor label, so we only need to solve the problem mentioned in the second paragraph of my previous comment, by checking that the author's other PRs are created after the PR triggering the workflow.

PR #66749, opened 2023-07-08
Sharper vector font display with subpixel rendering

Qrox:
#### Summary
Interface "Sharper vector font display with subpixel rendering"

#### Purpose of change
Currently, the game renders fonts either unblended (which looks terrible) or blended with the background at each pixel (not terribly terrible, but still terrible if you use a thinner font).

Fixes #66016.

#### Describe the solution
Use sub-pixel rendering which has been around for ages but only [added to SDL2_ttf last year](https://github.com/libsdl-org/SDL_ttf/releases/tag/release-2.20.0).

In a nut shell, it lights up the red/green/blue subpixels on your screen individually while doing some math to make sure they still blend to show the desired text color, so you now have 3x resolution in the direction of the subpixels.

This PR adds a new blending mode, "lcd", to render fonts using subpixel rendering, and automatically detects whether the library/renderer support all functions required by the new blending mode, and falls back to "blended" if they do not. However, SDL does not support blending subpixel-rendered font with arbitrary background natively, so I had to approximate it using two render-copy calls, which should be accurate to 1/255 (rounding error of two render-copy calls), but may make rendering slower.

Also changed the build matrix to use SDL2_TTF 2.20.2 instead. To maintain back-compatibility with older SDL2_TTF versions, the new code is guarded with a version test macro.

Also reduced code duplication by using `create_surface_32` and fixed an inversed debug message in `ColorModulatedGeometryRenderer`.

Invest in retina display.

#### Testing
Started the game and text looked much crisper. Some screenshots are included in the "additional context" section. Switched blending mode settings and the respective blending modes were correctly used.

Tested every renderer, and apart from "software" which does not support custom blending mode and "opengles2" which causes black screen unrelated to this PR, all other renderers available on my system (direct3d, direct3d11, direct3d12, opengl) rendered the fonts correctly. "software" fell back to "blended" instead.

The code still compiled with SDL2_TTF 2.0.15, in which case it fell back to "blended" when "lcd" was selected.

#### Additional context
Some screenshots before and after, using `SourceCodePro-Regular`:

Before ("blended", color manager):

image

After ("lcd", color manager):

image

Before ("blended", debug tempearture overlay)

image

After ("lcd", debug temperature overlay)

image

Still need to compare for rendering speed.

According to SDL2 TTF and freetype docs, this is mainly for LCD, but I imagine it also works as long as the subpixels are spatially separated. I'm not sure how if it will work for unconventional subpixel layout though, because freetype requires the layout to be specified via a function call but SDL2 TTF does not seem to do so. I'm also unsure how it looks when there are no spatially separated subpixels. Feedback would be appreciated if anyone is using non-LCD display or display with non-conventional subpixel layout.

Currently there is some green color bias for Chinese characters from unifont, but I'm not entirely sure why that happens and how to fix it. I'll probably add a per-font blending mode setting if I cannot think of a better solution.

Also, I'm not sure what has been added to SDL2 TTF, but it has increased from less than 1 Mb (2.0.15) to 43 Mb (2.20.2)... I think it's still worth the increased size for better display but I can also revert the release workflow changes if it is not desired.


harakka:
Did you check if this causes any regressions with bitmap style fonts that aren't meant to be antialiased at all, such as Terminus?

Qrox:
Good question. I'd hope freetype to handle it correctly but I'll make sure.

Qrox:
SDL2 ttf returns error when using bitmap font such as Terminus so I made it fall back to "blended" in that case.

Unifont is having some green color bias but I'm not sure it's due to the blending algorithm or it's something with my screen. It's probably not related to SDL2_ttf though, because it also has similar bias when viewed from the Windows font preview at 12 pts (16 px). I updated unifont to the newest version and modified it to use a bitmap version at 16 px to work around this problem.

The TTF_HINTING_LIGHT hinting mode produced best quality display in my test. Other modes either had severe color bias or rendered some characters at weird positions.

The extra size of SDL2_ttf is mostly due to it being unstripped so I've updated the workflows and scripts to strip the dll.

Qrox:
Tested the speed by disabling tileset and running the framerate benchmark twice. FPS before with blending was 42.183, 41.849, and FPS after with LCD blending was 34.193, 34.193

The speed of LCD mode is slightly lower than the blended mode, so I've made the option default to 'Solid' to match the bitmap fonts (Terminus and unifont at 16px) the game uses by default. The old option is migrated from "false" and "true" to "solid" and "lcd" respectively, because presumably if someone enables blending, they are using some vector font so "lcd" will look better for them.

Qrox:
SDL2_ttf 2.20.2 is exporting __some__ of zlib's symbols which causes duplicate definition errors during linking, but I could not fix it by removing the `-lz` flags because __not all__ symbols are exported. I had to use `objcopy` to rename the conflicting symbols, which fixed the linking errors.

Qrox:
I tested rotating the screenshots 90 degrees and observed color fringes, which means this will not work for vertical rgb geometry, and presumably other subpixel geometries, so I'm marking this as draft until I think of a solution.

> It seems the clang tidy was skipped, and I can't seem to find a build where it wasn't.

The third-to-last commit passed the clang-tidy checks, and the following commits' changed files do not seem to affect clang-tidy so the skip-duplicate workflows might have kicked in.

bombasticSlacks:
Okay, once it's at a place you are happy with I'll re-review thanks 😄

Night-Pryanik:

PR #66753, opened 2023-07-08
Iterate over corpses in active items instead of map tiles for resurrect check

SurFlurer:
#### Summary
none

#### Purpose of change
Improve performance when there are necromancers in the reality bubble.

#### Describe the solution
Use `get_active_items_in_radius()` to grab nearby corpses only ( neglect other items, and do not iterate over tiles ). This has the advantage of less `sees()` and `is_empty()` check, which are both somewhat expensive. This also makes this function not linearly slower when there are large quantities of unrelated items nearby. ( Thanks andrei8l !)

I also made a cache for related map tile attributes to reduce calls of `sees()` and `is_empty()` if there are many corpses in one tile.

#### Testing
In the refugee center, with one necromancer in the backyard zombie horde, waiting for 30 minutes:
1. With this change:
image

2. Without this change:
image

### DIFF
--- a/src/monattack.cpp
+++ b/src/monattack.cpp
@@ -1019,36 +1019,45 @@ bool mattack::resurrect( monster *z )
     bool found_eligible_corpse = false;
     int lowest_raise_score = INT_MAX;
     map &here = get_map();
-    for( const tripoint &p : here.points_in_radius( z->pos(), range ) ) {
-        if( !g->is_empty( p ) || here.get_field_intensity( p, fd_fire ) > 1 ||
-            !here.sees( z->pos(), p, -1 ) ) {
+    // Cache map stats.
+    std::unordered_map<tripoint, bool> sees_and_is_empty_cache;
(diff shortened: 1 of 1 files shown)


andrei8l:
You can probably do even better by only looking at corpses, which are cached separately. NPC AI code does that:

https://github.com/CleverRaven/Cataclysm-DDA/blob/b69ae48e9c29f079fe427dad026fef8410390af5/src/npcmove.cpp#L3229-L3230

SurFlurer:
> You can probably do even better by only looking at corpses, which are cached separately. NPC AI code does that:
> 
> https://github.com/CleverRaven/Cataclysm-DDA/blob/b69ae48e9c29f079fe427dad026fef8410390af5/src/npcmove.cpp#L3229-L3230

Great thanks! I was thinking about using the active items cache, but didn't make it to find example usage of it. This is exactly what I need.

PR #66761, opened 2023-07-08
Lower # of gas zombies

anoobindisguise:
#### Summary
Balance "fewer gas zombies"

#### Purpose of change
After a month of in game time, every single gas station had 5-10 gasoline zombies every single time. This was a stupidly large number of gasoline zombies especially with how badly they grief the map.

#### Describe the solution
lower number to 1-2 zombies 25% of the time instead of 5-10 zombies 100% of the time.

remove them entirely

### DIFF
--- a/data/json/mapgen/gas_stations/s_gas.json
+++ b/data/json/mapgen/gas_stations/s_gas.json
@@ -50,7 +50,7 @@
         { "liquid": "gasoline", "x": 2, "y": 3, "repeat": [ 200, 1075 ] },
         { "liquid": "diesel", "x": 4, "y": 3, "repeat": [ 200, 1075 ] }
       ],
-      "place_monster": [ { "group": "GROUP_ZOMBIE_GAS", "x": [ 5, 10 ], "y": [ 6, 10 ], "repeat": [ 5, 10 ] } ]
+      "place_monster": [ { "group": "GROUP_ZOMBIE_GAS", "x": [ 5, 10 ], "y": [ 6, 10 ], "repeat": [ 1, 2 ], "chance": 25 } ]
     }
   },
   {
@@ -157,7 +157,7 @@
           "y": [ 15, 16 ]
         }
       ],
-      "place_monster": [ { "group": "GROUP_ZOMBIE_GAS", "x": [ 5, 10 ], "y": [ 6, 10 ], "repeat": [ 5, 10 ] } ]
+      "place_monster": [ { "group": "GROUP_ZOMBIE_GAS", "x": [ 5, 10 ], "y": [ 6, 10 ], "repeat": [ 1, 2 ], "chance": 25 } ]
     }
   },
   {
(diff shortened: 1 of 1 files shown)


Procyonae:
I'm surprised this is getting so many +1s, you're reducing the average from 7.5 to 0.375, a 95% decrease on a unique enemy type that barely spawns outside of this change with the justification that they cause too much damage when they move at 25 speed so can be outmanoeuvred without having to explode them ridiculously easy, outside of stuff like fungals/ants killing them there's no particular danger in them exploding and I'd argue lowering the amount that spawn doesn't change their destructiveness because of this.

alp7292:
its already a unique enemy and what they are griefing anyway compared to other zombies and fungals

kholat:
Leave gas zombies alone.

fairyarmadillo:
This enemy should probably just be obsoleted. They pose no threat to anyone who knows what they are doing and their primary purpose appears to be farming infinite gas using the door/bathtub trick.

Procyonae:
> This enemy should probably just be obsoleted. They make very little sense (how did they even get the gasoline from the pump? Did they pay for it?), they pose no threat to anyone who knows what they are doing and their primary purpose appears to be farming infinite gas using the door/bathtub trick.

My headcanon is it's from standing around in the fumes for too long and I don't really see why an enemy needs to be dangerous to belong in the game, it's uniqueness makes it interesting in itself

fairyarmadillo:
They are filled with many gallons of gasoline which confers no apparent advantage (quite the opposite) and somehow have constantly lit pilot lights in their mouths which cause them to blow up when they die even if it's to venom or having rocks thrown at them. 99% of the time they just get other zombies killed.

I know zombies get to ignore conservation of mass/energy due to blob magic, but "a fat guy with a portal to an infinite gasoline dimension in his stomach" is an extremely goofy enemy with an ability almost as bad as the old zombie grenadier's death grenade that makes no sense and isn't even cool or effective like the shockers.

Procyonae:
> They are filled with many gallons of gasoline which confers no apparent advantage (quite the opposite) 

Again I don't see why all zed differences need to give them objective advantages "hulkier hulk with hulks for arms" isn't interesting enemy design, I'm not suggesting that gas zeds are like the pinnacle of enemy design but they are at least unique

> somehow have constantly lit pilot lights in their mouths which cause them to blow up when they die even if it's to venom or having rocks thrown at them. 

All other boomers do the same, that warrants mechanical changes not yeeting a specific one, having ways to kill boomers without them booming sounds like a fun addition

> 99% of the time they just get other zombies killed.

You could say the same about incandescent husks and acid zeds, I don't see a problem with this, the player being able to use their knowledge to use enemies abilities against them is good game design imo

> Removing them would not remove the player's ability to make fire fields with molotovs, arson, flamethrowers, etc.

I don't even see how this is relevant

> I know zombies get to ignore conservation of mass/energy due to blob magic, but "a fat guy with a portal to an infinite gasoline dimension in his stomach" is an extremely goofy enemy with an ability almost as bad as the old zombie grenadier's death grenade that makes no sense and isn't even cool or effective like the shockers.

Acid zeds make infinite acid out of nowhere and all boomers are fat goofy exploding enemies, if you want to suggest a rework for all boomers to make them less goofy without making them boring I'd be interested

> Fire is also very broken and is pretty much invariably fatal to the player.

I don't see how fire being broken is a reason to remove unique zeds, it's a reason to improve fire just like how acid just being a rubber boot check wasn't a reason to remove acid zeds it was a reason to improve acid

Procyonae:
I will add that I'm not against the spawn rate being tweaked, very few things spawn in 5-10 anything with 100% chance but just lowering the chance to like 50% and leaving the numbers the same would be plenty imo, some way to make the 0 -> everything at 7 days less jarring would be nice with a halfway evolution that could be mixed in or at least having regular zeds spawn there before 7 days to explain where the gas zeds are magically appearing from

fairyarmadillo:
Boomers pop like balloons and the effect is relatively mild. This is very different from somehow causing a combustion reaction that turns like a 50 meter circle into a raging inferno.

BjoHart:
I would also like to see less gasser zeds...

zachary-kaelan:
I remember my first time making it to a hospital it was already 50% on fire due to stupid gasoline zombies roaming the area and tripping over boxes of matches or something. I will gladly watch as their numbers get cut.

PR #66764, opened 2023-07-08
Digitigrade Mutation Line-of-Sight Fix

GalacticApple:
#### Summary
Bugfixes "Updates the mutation so that players can't see over objects while running on all-fours"

#### Purpose of change
The mutation has it so that you run on all-fours while running. However, you could still see over objects like you were standing upright.

#### Describe the solution
I made some changes to the character.cpp/h files and the lightmap.cpp file. Specifically, I added in a is_runallfours function which is then used in the lightmap.cpp code for limiting sight while crouching. 

#### Testing
It works on my personal build of CDDA.


alp7292:
does it also hides you when in cover like crouching

PR #66765, opened 2023-07-08
Update claw bar volume to be consistent with crowbar

ampersand55:
#### Summary
Balance "Update claw bar volume to be consistent with crowbar"

#### Purpose of change

A claw bar is basically a smaller crowbar, so it should have less volume, but it currently has a volume of 350 ml (which makes it not fit into a tactical tool pouch) compared to the crowbar which has a volume of 100 ml. This currently makes it completely useless.

#### Describe the solution

This sets claw bar volume to 64 ml, 64 % of the crow bar, which is the same proportion as the weight.

#### Testing

None, just linted as it only changes a number in json.

### DIFF
--- a/data/json/items/tool/workshop.json
+++ b/data/json/items/tool/workshop.json
@@ -253,7 +253,7 @@
     "qualities": [ [ "PRY", 2 ], [ "HAMMER", 1 ], [ "PRYING_NAIL", 1 ] ],
     "use_action": [ "CROWBAR" ],
     "flags": [ "BELT_CLIP" ],
-    "melee_damage": { "bash": 10, "cut": 1 }
+    "melee_damage": { "bash": 9, "cut": 1 }
   },
   {
     "id": "con_mix",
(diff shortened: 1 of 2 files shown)


Karol1223:
Tbh I am willing to bet money that it's because the crowbar's volume is fucked. 100ml for a crowbar seems ungodly low.

ampersand55:
I reverted the test change and decreased the damage to match the test. 

I don't know how volume is considered. Steel has a density of 8kg/l and as they are L-shaped you can fit them snugly in a corner of a box-shaped storage space and not take up much room.

Heavy crow bar has the same density of 5.5, and makeshift crowbar has a density of 4.5, but Halligan bar has a density exceeding that of steel (4.77 kg/0.55 l = 8.67 kg/l) and should probably have its volume increased. Heavy crow bar should probably also have prying 4 similar to a Halligan bar as it's 1.5 meters long. Perhaps even a reach-attack.

But those changes are beyond the scope of this PR.

Karol1223:
> But those changes are beyond the scope of this PR.

You're right. I'll take a look at crowbars sometime soon and audit their numbers to be based on actual real-world items but until now this is an okay compromise.

PR #66781, opened 2023-07-09
Better scenario worldbuilding by giving PC knowledge of the nearest city on start

RenechCDDA:
#### Summary
Content "Most scenarios now provide vision of the nearest city on start"

#### Purpose of change
Some starts spawn outside of cities, but are not like... ~~innawoods players~~ detached from society. Presumably the PC had a home before the cataclysm, and some sort of education, and at least knew where the mcdonalds down the street was. The game doesn't currently represent that - it only plops down a big (15-tile default) radius of sight around you at the start, and if there's no civilization in sight, too bad, guess your character was living off the grid!

#### Describe the solution
Add a json toggle defined per-scenario (**defaults true** so it doesn't need to be added to 99% of starts) whether this vision should be granted.

Draw a path from the nearest road to the city center, revealing an area around it.

Reveal an area around the city center, scaling with the CITY_SIZE setting.

The current method of scaling with city size doesn't seem to be perfect? Sometimes cities are massive, sometimes cities are small, and the vision might either be insufficient or too much. But then again, I wouldn't expect a city's resident to know *the entire* city...

#### Testing

Here is an average evac shelter start:
image

Here is **basically the worst-case scenario**:
image

#### Additional context
Disabled this for the following scenarios:
Medieval peasant(DDA)
Lab patient(DDA)
Island Prison(DDA)
Mi-Go prisoner(DDA)
Portal dependent(DDA)
Aircraft carrier start(DDA)
Middle of nowhere(DDA)
Lost Underground(DDA)
*All innawoods starts*(just in case)
Wizard's Vacation(Magiclysm)
Exile (Magiclysm)

### DIFF
--- a/data/json/scenarios.json
+++ b/data/json/scenarios.json
@@ -406,6 +406,7 @@
     "forced_traits": [ "ILLITERATE", "ANTIJUNK", "STRONGSTOMACH", "SPIRITUAL" ],
     "forbidden_traits": [ "FASTREADER", "SLOWREADER" ],
     "requirement": "achievement_witnessed_portal_storm",
+    "reveal_locale": false,
     "flags": [ "CHALLENGE", "LONE_START" ]
   },
   {
@@ -459,6 +460,7 @@
       "ELECTRORECEPTORS"
     ],
     "requirement": "achievement_reach_lab_finale",
+    "reveal_locale": false,
     "flags": [ "CHALLENGE", "CITY_START", "LONE_START" ]
   },
   {
@@ -627,6 +629,7 @@
       "cop"
     ],
     "requirement": "achievement_reach_island_prison",
+    "reveal_locale": false,
     "flags": [ "LONE_START", "CHALLENGE" ]
   },
   {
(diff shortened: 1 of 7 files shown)


RenechCDDA:
@ Mod maintainers & those more versed in the scenarios than I am, I would be interested in hearing which scenarios might need to be updated to toggle this off?

ashGlaw:
Love this idea

GuardianDll:
I wonder how it would work with alcatraz, as you can't draw a road here

> which scenarios might need to be updated to toggle this off?

For magiclysm: lake_retreat, exile
For xedra: none

RenechCDDA:
If it can't find a road it actually draws nothing (silently fails).

alp7292:
does this spawn world items. for example if you go to a unvisited place and load back save items and zombies will be different. but if you go to evac shelter and learn refugee center location everything on tha path is generated and if you follow the road and load save back everything will be same. it most of the time generates places i will never visit and by generating it.  it saving and loading times longer.

RenechCDDA:
No, it doesn't load any of the local maps.

image

Light-Wave (approved):

Issue #58025, opened 2022-05-30
New industrial professions and items


Termineitor244:
### Is your feature request related to a problem? Please describe.

There is a lack of proper PPE in the game, even though this kind of equipment would be items that are very useful to have in an apocalypse... In the same vein, there is a lack of professions with proper personal protection equipment.

### Solution you would like.

More hard hats! Maybe, the hard hats used in industrial environments come in many varieties, not just in color but in protection too.

Hard hats add-ons: Hard hats are often used in conjunction with certain items to better protect the user, why not use them in CDDA too?

Coverall/overall/boiler suit: These suits cover the whole body (excluding the head/hands and feet), commonly used to protect the user (and the user clothes) from chemical splatter and to proportionate a comfortable protective garment, they come in many varieties in color but in use too, most are fire-resistant and others have certain properties that make them very useful in dangerous situations. This item could be improved with some kevlar (like 50 / 75 % of kevlar coverage if you want to mantain the mobility of the suit) for better protection, it certainly works as a good alternative to a survivor duster or whatever, since this is an item designed to cover all your body comfortably, and has some thickness by itself.

New professions: Industrial engineers/workers, with all kinds of PPE and related proficiencies.

And even more...

### Describe alternatives you have considered.

To only add the hard hats related changes, like in #58020

anothersimulacrum:
The bionic starts are already a mod for this reason.

Inglonias:
I forgot that had been moved to a mod.

Night-Pryanik:
Closing as stale, since stalebot can't do it by itself.

PR #66783, opened 2023-07-09
Add work coveralls

anoobindisguise:
#### Summary
Content "work coveralls"

#### Purpose of change
I wanted to add work coveralls to the game. Partially this is because I think they would make a good basis for survivor suits but they're also a thing people wear IRL for protection and the existing in game "jumpsuit" was a poor representation of that.

#### Describe the solution
Based on products such as https://www.amazon.com/gp/product/B00V2VXJRW/
Add two versions, a simple 0.5mm version and a 0.5mm+0.5mm cotton liner version. 0.5mm is roughly 8oz fabric, while the product advertised is roughly 10 oz + 4oz cotton which works out to about 0.65mm canvas + 0.25mm cotton I decided to make the numbers just 0.5mm canvas + 0.5mm cotton which works out to about the same expected protection values. Plus it means if these do become a basis for survivor suits you could use them interchangeably if you stripped out the liner on the heavy version.
Encumbrance is 2 on arms + 2-4 on legs/torso on the light version, while the heavy version is 3 + 3-5. I think that's about fair.
100% coverage seems justified based on the reference images, especially with the collar they clearly cover up to the neck.

Different pockets. There was previously an effort to neuter normal layer pockets but these jumpsuit pockets are just the pockets of pre existing items merged together (work pants + work shirt) so it's not like they break established precedent.

#### Testing
loaded in game and didn't have errors, confirmed stats look good, made sure it spawns in the world properly.

#### Additional context

coveralls1

coveralls2

coveralls3


TheShadowFerret:
>these jumpsuit pockets are just the pockets of **pre existing items merged together (work pants + work shirt)**


Can we have leather-padded overalls from this like we do pants 👀

anoobindisguise:
> > these jumpsuit pockets are just the pockets of **pre existing items merged together (work pants + work shirt)**
> 
> Can we have leather-padded overalls from this like we do pants 👀

I plan on it!

Termineitor244:
Pretty neat!! I have wanted them since #58025, but never went around adding it...

Just a question, I use several of these at my work and I think they should be considered OUTER layer gear, since they are supposed to be worn over your every day clothes, the pants and shirts versions of these item are indeed normal layer, but the coverall itself is loose enough to fit some pants inside (And for normal layer reinforced clothing we already have work shirts and pants).

anoobindisguise:
> Pretty neat!! I have wanted them since #58025, but never went around adding it...

Thanks!

> Just a question, I use several of these at my work and I think they should be considered OUTER layer gear, since they are supposed to be worn over your every day clothes, the pants and shirts versions of these item are indeed normal layer, but the coverall itself is loose enough to fit some pants inside (And for normal layer reinforced clothing we already have work shirts and pants).

I'm torn on this. On one hand that makes sense to me but I also want them to be compatible with knee/shoulder pads, chaps, etc, without extra conflicts. Maybe allow them to be swapped between layers?

Termineitor244:
> I'm torn on this. On one hand that makes sense to me but I also want them to be compatible with knee/shoulder pads, chaps, etc, without extra conflicts. Maybe allow them to be swapped between layers?

That is one option, but I have to admit that it doesn't totally convince me... One point in favor of this is that it would allow the items to work as a base for custom armor in both layers I guess, but I don't really think these pieces of clothing can really be "adjusted" (like, for example, corsets can) to fit closer/looser in the body...

This is mainly a problem of these other items being difficult to adjust around because of their nature as outer armor even though they are worn strapped, I guess the more logical solution to all of this would be for *those* to be on an additional layer that does not conflict, or *them* being able to adjust how close to the body they are, since they are all, well, strapped...

A though choice whatever it is I guess.

fairyarmadillo:
What if the thick/padded one was outer, and the regular one was normal? Still imperfect but it makes some sense, and gives you a few options.

anoobindisguise:
I've gotten some feedback from the devcord and I've decided to make these normal layer since that was the majority opinion.

PR #66785, opened 2023-07-09
Placeable ladders

Procyonae:
#### Summary
Content "Adds ladder items that can be used to create temporary z level transitions"

#### Purpose of change

For some reason everyone who knows a man with a ladder only knows a man with a stepladder in cata

#### Describe the solution

Adds straight wooden, extending aluminium, extending fibreglass and telescopic aluminium ladders that can be placed via the construction menu from above or below to from terrain ladders functionally identical to t_ladder_up and t_ladder_down.
Picking the ladder back up from above and below are separate constructions due to the needed checks not fitting with simple deconstruction, but all the constructions/deconstructions are in one category to reduce ui clutter (this does mean if there are multiple valid ladder items nearby there's no way to control which you place)
Ladder requires a single orthogonal support at the base.
The _up will easily breaks into it's item and the _down can easily be bashed into nothing as a hacky solution for no multi tile bash support.
I've added spawns for the aluminium and fibreglass ladders but not gone too crazy, the wooden one is only craftable for now bc wooden ladders that size don't seem popular in the US

Half of this could probably be refactored if I were more C++ literate, moving the unordered_maps to the header or even the top of the document is resulting in them throwing out_of_range errors for some reason, check if I did something dumb in [ca06369](https://github.com/CleverRaven/Cataclysm-DDA/pull/66785/commits/ca063699ccb503a48611c62533f4020582c999b8)

Figuring out how to make a flag for furniture that supports actors so they can be furniture instead
Trying to figure out map_bash_info C++ to support bashing removing both _up and _down for a single item regardless of which is bashed or asking bombastic to add EOC functionality to map_bash_info

Changing stepladders
Adding in situ constructions/deconstructions for the steel fixed t_ladder_up/_down/_up_down ladders.
Making a difference between hanging fixed ladders and those fully attached to the wall.
Adding more/less ladders
Adding recipies for the non wooden ladders
Rope ladders (not that difficult to do but meh)

#### Testing

Placement, deconstruction and bashing work as intended

#### Additional context

Not super happy with the ids but I want to leave room for future additions (though hopefully they'll all become furniture anyway)

check_support_singular() is necessary rather than adding a default parameter supports_required due to the map pre_special_map requiring the functions to only have one parameter


Procyonae:
Honestly I'm not even sure I want these to merge as terrain, their removal resulting in t_floor just sucks, redrafting and I'll have another look to see if I can figure a furniture ACTS_AS_FLOOR flag out

PR #66793, opened 2023-07-10
NPC selects martial arts

lispcoc:
#### Summary
Interface "NPC selects melee style"

#### Purpose of change

NPCs can learn martial arts, but cannnot select their style.

#### Describe the solution

Add new talk function "pick_style".

#### Testing

image
image


Maleclypse:
Ok but will they actually use that style now? Because my current understanding is that they never use a martial art even if they only have one.

lispcoc:
> Ok but will they actually use that style now? Because my current understanding is that they never use a martial art even if they only have one.
Yes, but message is a little weird. I will try to fix it.
image

Theawesomeboophis:
Yay, I can finally reenact Cobra Kai with my npcs in the local dojo, this is a very cool addition and I can't wait to see it in game, glad npcs have been getting some updates the last couple of years, they feel more like people than ever and I hope this trend continues.

PR #66794, opened 2023-07-10
Debug menu quick setup option for manual testing

### LINKED ISSUE #66945: Both info and import are bound to "i" by default in debug menu
### Describe the bug

When opening the debug menu, both info and import are bound to the "i" key. This makes it so that the key binds listed in the attach save file can cause confusion when hitting i and the import menu opens instead of the info menu.

### Attach save file

N/A

### Steps to reproduce

1. Open debug menu
2. hit "i" key
3. have import menu open instead of info

### Expected behavior

In the bug submitting form, one of the steps is attach save file, which lists that hitting "i" will open the info menu. This can easily be figured out, but it is still a discrepancy I believe should either be fixed in the game or on GitHub.

### Screenshots

image

- OS: Linux
- Game Version: 0.G-3026-g30fe61a8fd [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Aftershock [aftershock],
    Magiclysm [magiclysm],
    SpeedyDex [speedydex],
]

### Additional context

_No response_

RenechCDDA:
#### Summary
Infrastructure "Quick setup for manual testing in the debug menu"

#### Purpose of change
* Closes #66945

I use the debug menu a lot when I'm tracking down bugs or testing my own work. In particular, several of the debug mutations (HVAC, life support, invincibility, invisibility) are key to avoiding interruptions. It's much faster to toggle off one or two mutations for testing (quick setup-->remove hammerspace--->test crafting) than it is to toggle on all the other ones. So, it would be nice to have a way to quickly do that...

#### Describe the solution
-*Move character import/export to the player submenu* (IMO should not be accessible from the top-level menu, for initial testing that's fine but now that it's merged it should not be used that often)

-Add a new top-level option called `Quick setup...` which does exactly what it says on the tin.

-Quick setup resets a bunch of values to default and otherwise makes the player character normal, as well as setting a bunch of debug mutations which are often desired when testing via the debug menu. (specifically, targets player character only)

-Also add a new debug trait, DEBUG_SPEED. Multiplies speed by 10. The faster you move, the quicker it is to travel short distances without waffling through the teleport menu (essentially, monsters and items process up to 10 times less often = big wins).

I am sure this does not cover all common test cases, but it is intended to serve as a general option to avoid lots of menu waffling. There are some commented out lines with notes for things that could possibly be added.

#### Testing
It works, but set_mutations() does not do some things like add messages to the log for new mutations. (Probably for the best when adding a dozen mutations at once?)


GuardianDll:
Actually thought having such option in main menu, similar to `Start now!` option, but instead of spawning in world with random scenario, it spawn in default scenario with debug traits
i use the similar approach, when i just spawn a character in debug world, give em a mutations, and save it (still need to recreate world time to time)

alp7292:
adding export/import to submenu is good but adding quick setup to top of the menu is very personal and most people wont use it. at least add it into submenu.

ilovehans10:
> Actually thought having such option in main menu, similar to `Start now!` option, but instead of spawning in world with random scenario, it spawn in default scenario with debug traits i use the similar approach, when i just spawn a character in debug world, give em a mutations, and save it (still need to recreate world time to time)

Would it work to add a compile flag to compile a “debug build” where the only difference is that there is the start now option. This way, your standard player wouldn't find it, but it would still be accessible to developers. It could even be set up so that making a debug build makes the game automatically chooses the start now option on the main menu so that developers can bypass the main menu altogether.

If you ask me, there should be a command line options for the game to make it so you can launch into your most recently played save on game launch. E.g., `cataclysm-tiles --load-last` or `cataclysm-tiles --load-save="myworld"`, but command line options haven't really ever seemed to catch traction with this game.

RenechCDDA:
> If you ask me, there should be a command line options for the game to make it so you can launch into your most recently played save on game launch. E.g., `cataclysm-tiles --load-last` or `cataclysm-tiles --load-save="myworld"`, but command line options haven't really ever seemed to catch traction with this game.

This already exists, but is also way off-topic.

https://github.com/CleverRaven/Cataclysm-DDA/blob/237374d4ec0196725844a6f430b02bccc96368c3/src/main.cpp#L317-L319

PR #66797, opened 2023-07-10
fix: missing spaces

casswedson:
#### Summary
None

#### Purpose of change
text gets double spaces after sentence end; fix spaces in the crit message for NPCs
Screenshot from 2023-07-09 20-04-00

add spaces after the numbers in this Protection line
Screenshot from 2023-07-09 20-28-38

Screenshot from 2023-07-10 09-41-04

### DIFF
--- a/src/item_pocket.cpp
+++ b/src/item_pocket.cpp
@@ -1196,7 +1196,7 @@ void item_pocket::contents_info( std::vector<iteminfo> &info, int pocket_number,
             const std::vector<damage_info_order> &all_ablate = damage_info_order::get_all(
                         damage_info_order::info_type::ABLATE );
             for( const damage_info_order &dio : all_ablate ) {
-                std::string label = string_format( idx == 0 ? _( "<bold>Protection</bold>: %s: " ) : " %s: ",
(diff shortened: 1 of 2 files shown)


src/item_pocket.cpp
```diff
@@ -1196,7 +1196,7 @@ void item_pocket::contents_info( std::vector<iteminfo> &info, int pocket_number,
             const std::vector<damage_info_order> &all_ablate = damage_info_order::get_all(
                         damage_info_order::info_type::ABLATE );
             for( const damage_info_order &dio : all_ablate ) {
-                std::string label = string_format( idx == 0 ? _( "<bold>Protection</bold>: %s: " ) : "%s: ",
+                std::string label = string_format( idx == 0 ? _( "<bold>Protection</bold>: %s: " ) : " %s: ",
```
Qrox, line 1199:
```suggestion
                std::string label = string_format( idx == 0 ? _( "<bold>Protection</bold>: %s: " ) : pgettext( "protection info", " %s: " ),
```

Just in case some languages don't want those spaces in translation.

PR #66806, opened 2023-07-11
Fix typo in src/third_party/CMakeLists.txt

Qrox:
#### Summary
None

#### Purpose of change
Fix a typo

#### Describe the solution
Fix the typo

Remove the line since it did not cause compilation errors, meaning it might be not needed at all?

#### Testing
Compiled the game without any problem.

#### Additional context
I originally PR'd this to #65709, but that probably needs more work before it is mergeable so I decided to PR this separately.


akrieger:
> it might be not needed at all?

I think it's only relevant in the context of project generators that need to explicitly list sources/headers in a project (eg. msvc/xcode). The target_include_directories call at line 22 is the functional component for the cli builds we use cmake for.

PR #66808, opened 2023-07-11
More large building and  up above start locations

ashGlaw:
#### Summary
Content "More large building and  up above start locations"

#### Purpose of change

I noticed some newer tall buildings hadn't been added to the appropriate starting scenarios.

#### Describe the solution

Made a few start_locations, added them to scenarios.json

Not let players start atop an apartment complex, school, convention center, etc.

#### Testing

I started new characters at each of the new start locations in the scenarios I added to. I got a "Could not find a good starting place for character" error a few times, but ignoring it started the game all the same. I don't think anything I added touches on that problem.

#### Additional context

N/A


Procyonae:
The only issue with this is the spawn location only tries to look for an indoors tile afaik so if it can't find one it can just dump you in open air and you can fall several stories or even die within the first second of starting which while amusing isn't a great thing for new players to experience, I'm not suggesting this can't go through, the convention centre start would be especially nice bc it's so darn rare but it would be nice if some more logic went into the spawn location

ashGlaw:
I agree completely, though I'm not certain how to address that. Probably beyond the scope of this PR, tho

PR #65300, opened 2023-04-25
Obsolete "accessories" gunmod slot

irwiss:
#### Summary
None

#### Purpose of change

Fixes #65317

It's far past it's usefulness period, the only gunmod left hanging on it is a "belt clip" (and it didn't spawn in game before this PR added it to the itemgroup), this takes it out of it's misery, freeing up players from wondering "what goes in this 0/4 slot" and saves 2 lines in item info screen

#### Describe the solution

Moves belt clips to it's own slot only present on pistol_backup and the 2 pistol_base that can fit it

Belt clips added to spawn list, should actually spawn now.

Also fixes gun/toolmods that change flags not updating correctly when adding/removing gunmods and toolmods similar to what's done in https://github.com/CleverRaven/Cataclysm-DDA/pull/64949

#### Testing

Spawn a P-32 pistol, belt clip, attach it and try to wear

PR #66830, opened 2023-07-11
Accessories obsoletion strikes back

GuardianDll:
#### Summary
None
#### Purpose of change
Continue #65300
#### Describe the solution
Obsolete even more accessories
Also check the mod folder for guarantee obsoletion
Also fix belt clip slot being named as `belt_clip`

### DIFF
--- a/data/json/items/classes/gun.json
+++ b/data/json/items/classes/gun.json
@@ -100,7 +100,7 @@
     "//": "diminutive backup pistols that lack underbarrel rails, sight dovetails and decent places to affix stocks.",
     "valid_mod_locations": [
       [ "barrel", 1 ],
-      [ "belt_clip", 1 ],
+      [ "belt clip", 1 ],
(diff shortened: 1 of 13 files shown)


Standing-Storm (approved):

data/json/items/classes/gun.json
```diff
@@ -100,7 +100,7 @@
     "//": "diminutive backup pistols that lack underbarrel rails, sight dovetails and decent places to affix stocks.",
     "valid_mod_locations": [
       [ "barrel", 1 ],
-      [ "belt_clip", 1 ],
+      [ "belt clip", 1 ],
```
Maleclypse, line 103: Does belt clip id use space instead of underscore?

GuardianDll: For mod locations name = id, and belt clip slot in game is named as `belt_clip`. I just removed the underscore

PR #66834, opened 2023-07-11
Adjust the way Intelligence affects focus (multiplier instead of flat bonus)

anoobindisguise:
#### Summary
Balance "intelligence provides a multiplier to current focus rather than adding to its value"

#### Purpose of change
Intelligence made focus essentially irrelevant. At 10 focus, if you had 10 int, you learned twice as fast as a character at 8 int. This made int-stacking very broken and led to dramatic differences in learning rates between characters with normal amounts of int and characters with lots of it - in particular having >8 intelligence lets you power through having 1 focus and just spam your skills up anyways.

#### Describe the solution
Intelligence now provides a 1 + ((int-8) * 0.05) multiplier to learn rates rather than just +5. Fast Learner trait is unaffected, as is the bionic memory unit cbm. What this means is that 10 focus, with 10 int, is now 11 focus, rather than 20 focus. However this isn't *quite* as punitive as it sounds because having higher intelligence makes focus drain faster, so you will often bottom out at a value higher than 1. Nonetheless this is going to dramatically slow down the rate at which you learn skills; you won't be able to get basically all skills to 10 within the first month of starting the game anymore.

Increase the INT_BASED_LEARNING_FOCUS_ADJUSTMENT to be a greater value such as 10. 
Make the bionic memory unit and fast learner not give flat bonuses.

#### Testing
Need to compile this and test it.


anoobindisguise:
I think this is working now. It compiled and at least appeared to function the way I wanted it to.

PR #66836, opened 2023-07-12
Titanium skeletal bracing provides some bash armor

anoobindisguise:
#### Summary
Balance "titanium skeletal bracing provides bash armor"

#### Purpose of change
After the subdermal carbon filament got its bashing armor removed it freed up a niche for a bionic to provide bashing armor, and a titanium reinforcement to your skeleton would be an good way to do that which would reduce the odds of joint dislocation or fractures from trauma. This CBM was also a little on the weak side if you had slots enabled so giving it a buff feels fair to me.

#### Describe the solution
Provide 2 bash armor to the torso, arms, legs. Head, mouth, eyes, hands and feet are unaffected. I chose 2 bash armor as this is both the same as what was removed from subdermal carbon filament and also the same amount that Dense Bones gives you.

Make it also allow you to recover your limbs without a splint if your bones break.
Give armor to feet/hands.


Maddremor:
A stronger skeleton would actually make blunt force more dangerous in some circumstances. It would be a harder anvil to the hammer of the attack. You would also see it be most noticeable when taking very small amounts of damage when that would not differ from regular bones.

Maleclypse:
My ideal setup.  Carbon fiber should cover whole protect against cut, probably not pierce and little to no ballistic, no bash.  Bio_armor line should be reframed as ceramic plates that  and have bash and ballistic, only protect against pierce and cutting  a percentage of the time.  Probably workshopped a little more but basically.

PR #66840, opened 2023-07-12
Empty containers are not fuel-air bombs

### LINKED ISSUE #66760: butane explosions seem too deadly
### Describe the bug

I just had a character die from a butane can exploding in fire at their feet. This character was in RM13 armor and had the active defense bionic and had 209 HP but still died instantly. This seemed weird to me so I tested it in debug by setting my strength to 9999 and found that the butane explosion (0.5 liter canister of butane, 400 units of butane) did 30,000 damage to each limb. Oddly even an empty butane can still did over 1000 damage to each limb.

### Attach save file

n/a

### Steps to reproduce

1. set hp to an arbitrarily high number.
2. light something on fire and step into the fire (wear Rm13 or something to be immune to fire)
3. drop a butane can at your feet and observe the damage it deals.

### Expected behavior

A 0.5 liter canister of butane probably shouldn't be stronger than a mininuke.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.22621.1848 (22H2)
- Game Version: b69ae48 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Bionic Slots [cbm_slots]
]

### Additional context

_No response_

RenechCDDA:
#### Summary
None

#### Purpose of change
* Fixes #66760

A full 0.5L can of butane definitely does not have the TNT-equivalence of 4.5kg of TNT(!?)

(In testing, a 0.5L can of butane **destroyed the ground underneath it**, leaving a ~2m tall crater in the ground!!!)

An empty one definitely doesn't have any TNT-equivalence, being that it is a chunk of aluminum (and air)

There is maybe some argument to be made for a container of fuel to become an improvised explosive device but if butane refills were better grenades than military hand grenades you wouldn't be able to buy them at the convenience store. The current behavior also contrasts with every other container (and fuel container) in the game, which do not become improvised explosive devices when set on fire (with or without fuel)

#### Describe the solution
Remove ability for empty fuel cans to explode

(Butane tank is simply the butane can's values scaled up, so removed from there too)

#### Testing
I set full cans of butane on fire, they were destroyed in the fire but I did not become shredded by shrapnel.

### DIFF
--- a/data/json/items/tool/misc.json
+++ b/data/json/items/tool/misc.json
@@ -61,12 +61,6 @@
     "pocket_data": [ { "pocket_type": "MAGAZINE", "watertight": true, "rigid": true, "ammo_restriction": { "butane": 400 } } ],
     "material": [ "aluminum" ],
     "flags": [ "NO_RELOAD", "GAS_TANK" ],
-    "explosion": {
-      "power": 4503,
-      "distance_factor": 0.1,
-      "fire": true,
-      "shrapnel": { "casing_mass": 50, "fragment_mass": 0.025, "recovery": 0 }
-    },
     "explode_in_fire": true
   },
   {
--- a/data/json/items/tool/science.json
+++ b/data/json/items/tool/science.json
@@ -587,12 +587,6 @@
     ],
     "looks_like": "oxygen_tank",
     "flags": [  ],
-    "explosion": {
-      "power": 90100,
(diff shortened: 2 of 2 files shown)


Maleclypse:
Needs rebase to restart tests

PR #66847, opened 2023-07-12
Add more link ups

### LINKED ISSUE #66824: Allow even more commonly-used tools and appliances to be plugged into electrical grids

There are a bunch more frequently-used (for crafting or gear repair/maintenance) survivor tools that should be pluggable into an electrical grid, but can't be. These include:
- Space heaters
- Induction hotplate and other hotplate-using tools
- Soldering iron
- Electric firestarter
- Gunsmith/firearm repair kit
- Radio
- Electrolysis kit
- Electric forge/kiln
- Food dehydrator
- Aim game target practice kit

### Solution you would like.

Add the feature to plug such appliances into an electrical grid, as was recently added to existing items with that feature.

Doing this for EVERY electrical tool. 

Adding a wireless recharger tool which you can plug in, which then acts as a battery recharger for any tools on the same space/container or gives other tools the 'plugged in' feature.

### Additional context

_No response_

GuardianDll:
#### Summary
Features "More tools can be connected to grid directly"
#### Purpose of change
Close #66824
#### Describe the solution
Add even more link_up tools where applicable
#### Additional context
Some are specifically omitted, because they should get a tool version, that work from the grid, for example `cordless drill` and `cordless impact wrench`
Sorry @ Kamayana

### DIFF
--- a/data/json/items/tool/cooking.json
+++ b/data/json/items/tool/cooking.json
@@ -301,7 +301,7 @@
         "type": "link_up",
         "menu_text": "Plug in / Unplug",
         "ammo_scale": 0,
-        "cable_length": 1,
+        "cable_length": 2,
         "charge_rate": "1 kW"
       }
     ],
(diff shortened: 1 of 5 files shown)


Kamayana:
Ha, it's all good.

Kamayana:
One thing I'll suggest: I think cables should have a max_length of 2 at minimum - a max_length of 1 means the player will get a "cable is stretching to its limit" warning immediately on plugging a device in, which feels unusual. I found it feels better to give a bit more leeway than going with a straight "1 square == 1 meter" conversion.

ashGlaw:
Good job getting these. Great contribution. 👍

Which ones did you intentionally avoid?

GuardianDll:
The one that specifically say they are battery powered, and one that exist in both corded and cordless variants (usually it's the same tools, i just don't have the list) - they should get a different treatment, which is "make a versions of this tools, that can work only using cord (i believe Kamayana made this possible) and add them to related loot groups

PR #66849, opened 2023-07-12
Add some splintmail additions: Splint mail greaves and vambraces

MorvarchPrincess:
#### Summary
Content "Added Splintmail Greaves and Vambraces"
#### Purpose of change

Apparently I was missing one last piece of the varangian guard cosplay -> Splintmail greaves and vambraces. Partially added because there were no steel greaves to wear with lamellar (lamellar was in the awkward situation of covering upper arms along with the torso, and most arm armors are both upper and lower arms and would conflict).

#### Describe the solution

Added splintmail greaves and vambraces. These are basically identical to the leg guard and arm guards but only cover the lower arms/legs instead of all of the arms/legs. I adjusted recipe costs and time to compensate, but otherwise the stats should be the same.

Not adding the greaves -> These are less important from an armor variety standpoint, but still made sense to add while I was at it.

#### Testing

Booted up the game and confirmed it loaded, I could spawn and create the items, and that everything looked good both recipe and item wise.

#### Additional context

[Picture of varangian guard armor for reference](https://qph.cf2.quoracdn.net/main-qimg-847cd58b669c01b1eee9b24811096244-lq)


randomtyper:
We already have splint leg and arm guards, no?

https://cdda-guide.nornagon.net/item/legguard_lc_brigandine

https://cdda-guide.nornagon.net/item/armguard_qt_brigandine

MorvarchPrincess:
> We already have splint leg and arm guards, no?
> 
> https://cdda-guide.nornagon.net/item/legguard_lc_brigandine
> 
> https://cdda-guide.nornagon.net/item/armguard_qt_brigandine

We do, but if you look those cover both upper and lower. These new additions are just lower.

PR #66856, opened 2023-07-12
Some brain bionics cannot be removed

anoobindisguise:
#### Summary
Balance "you can't uninstall bio_int_enhancer or bio_memory"

#### Purpose of change
Some bionics that directly augment your cognitive processes do not make sense to be removed. In case of the memory especially, if you're storing your memories and thoughts on a drive you'd lose access to them when you took it out of your head. For cerebral booster especially I wanted to make it spookier and more body horror-y (some of your more useless neurons are getting yanked out and replaced with better stuff)

#### Describe the solution
Add cant_remove_reasons to them.

Do the same for other brain implants (synaptic regen, wired reflexes, soporific inducer, remote controller, endocrine enervator) for a lot of those they're less "hardwired" and are just nerve stimulators so not as bad maybe.
I do think Muscle Augmentation could be changed to be irremovable and lower your Strength to 0 but give a flat str bonus of like 14-16 or something because it wholly replaces your muscles with a myomer graft. Same thing for Wired Reflexes which replaces most of your nervous system with actual wires. (And maybe synaptic accelerator too?)


sadenar:
Maybe warn players of the fact these can't be removed after installation in the autodoc menu too if possible?

ashGlaw (approved):

anoobindisguise:
> Maybe warn players of the fact these can't be removed after installation in the autodoc menu too if possible?

That would require some hardcoded extras into the autodoc menu, but I did update the item descriptions so if the player reads those before installing them then they should be sufficiently prepared.

PR #45868, opened 2020-12-06
Redesigning arthropods, Pt 2: Centipedes

Venera3:
#### Summary

SUMMARY: Content "Centipede redesign"

#### Purpose of change

So centipedes are cool, right? They are one of the oldest predatory arthropods with a 400 million-year fossil record, the oldest terrestrial users of venom and have a presence on all continents aside from Antarctica, feeding on everything they can get their venom claws into ( a list which includes birds, frogs and slower rats). Cata doesn't nearly do them justice, so let's change that! Part of #45030.

#### Describe the solution

- Added a custom species and monster faction to centipedes, so that they aren't as skittish and hunt appropriate prey
- Added a life cycle to centipedes and a dedicated reproductive variant, ~~with the goal of incentivising the player to root out a centipede infestation lest they spread to the surrounding area~~
- Redesigned the mega variant into a late-game, (almost) bulletproof menace with enough AP to open power armor
- Added eggs as a butchery result and redid the insect eggs to use a common abstract instead of copying from spiders and updated them with toxins
- Adjusted swamp spawn rates slightly to accomodate their arrival and fix some weirdness
- ~~Included the JSON-defined special attacks and some undocumented fields, and clarified some others in MONSTERS.md~~ Became its own PR

TODO:

- [x] Combat testing

- [x] Custom harvest lists and snippets 

- [x] Reproduction fine tuning

#### Describe alternatives you've considered

Letting their venom be their main threat factor, but those mouthparts look dangerous enough to warrant high damage and justify armor piercing. 
~~Their high reproduction rate is a way to keep swamps alive even with the interspecies violence, and to force players to patrol the chunks of swamp loaded into the reality bubble along their most frequented routes, but that might not pan out as expected and requires more testing.~~ See comments
I considered adding them/ a hibernating variant to the caves in the winter, but since almost all caves use hand-picked spawns I decided against it. I might do a dedicated PR in the future to that effect.
I'd like to redo chitin crafting slightly to use rarer chitinous plates or something similar, so you're not set up to craft a whole armor set the moment you cheese your first dark wyrm, but that can wait.

#### Testing

Monsters spawn, behave as expected, reproduce and hunt.

Reproduction testing:
[...]

PR #66858, opened 2023-07-12
(WIP) Oversized mutant arthropod rework

randomtyper:
#### Summary
Balance "Reworks huge mutant arthropods to be less bland"

#### Purpose of change

Mutant arthropods on evolving to their final form almost universally get their speed reduced by factor of 4. Meaning that a bug with 100 speed (can catch up to a walking healthy human) evolves into one with 25 speed, almost as low as a goddamn crawling zombie! 

In realism terms, such low speed would mean quick death from starvation, at least for carnivores. Also, increased size means increased limb size and stride length, which should at least compensate for general increase of sluggishness, if not eliminate it. In game balance terms, such "final" forms are less dangerous to a player than "non-evolved" ones, defying game balance logic, as player can just... walk away from those super mutants, even if he is severely wounded, while smaller, "less advanced" forms can catch up to player and deal some damage at least in some cases.

At the same time, reworked warehouse spider enjoys nice 120 speed, mommy long-legs, which is encumbered by spiderlings, has 85 speed. Colossal croaker, despite weighing 800 kg and having whopping 1200 HP, has 80 speed and retains leap ability (sure, croaker may be a bit overtuned, but still)!

Finally, the names and performance do not really match up with such low speeds. Missile spider is less of missile, but more of a jumping mine. Great mantis may be capable of killing a moose... If it can reach it at all. The wyrd spider is just about the only one to make some sense, "depending on its webs to prevent prey from fleeing", and even then I don't think it would really help, and even then, its speed is 27, which, ironically, is on the higher side for mega mutant arthropods.

---------------------

I'll leave the old reasons for changes here, but now the main reason for rework is that the mega versions of most arthropods are just beefed up mutants that are not really enjoyable to fight and are pretty bland with little difference between eachother. Most follow the pattern "scary in melee chunks of chitin that can be easily walked away from", and that is neither fun to encounter nor enjoyable to fight.

#### Describe the solution

Up to heavy rework. Increases proportional speed multiplier for most mega arthropods from 0.25 to 0.5 as a stopgap measure.

Plans include: branching out the arthropod trees for increase in variety, redesigns of existing non-rebalanced arthropods.

Making a small sweeping change to speed, which was not well accepted.

Leaving everything as be.

#### Testing
None.

#### Additional context
None.


PatrikLundell:
The only reason these monstrosities are bearable is that you can avoid them unless you absolutely have to eliminate them. Having these absurd armor and HP mountains chase you around would certainly reduce the number of available bullets in the game as you'd have to spend them all on trying to whittle these things down to death, but it would definitely not be fun to have to try to deal with all of them.

randomtyper:
> The only reason these monstrosities are bearable is that you can avoid them unless you absolutely have to eliminate them. Having these absurd armor and HP mountains chase you around would certainly reduce the number of available bullets in the game as you'd have to spend them all on trying to whittle these things down to death, but it would definitely not be fun to have to try to deal with all of them.

The problem is that those monstrosities are weaker im many aspects than current faster enemies (colossal croaker, pretty freshly reworked warehouse spider, compare those, say, to great mantis). Their HP is no longer as stupidly high as it once was, most having 400-500 HP at best. As of now, I'd say the less evolved versions somehow pose more threat to survivors than the bigger ones outside certain circunstances, while almost every other evolution line gives an increase in actual threat level every evolution (example: zombie -> tough zombie -> zombie brute -> zombie hulk).

Sure, they are allowed to be easier to avoid than their smaller versions (or not, if we are talking realism, which is deemed more important in the current state of game's design than creatures being bearable), but not to the point where they can barely outrun a crawling (!) zombie! They are not capable of catching up to any survivor and posing any threat unless one is already on brink of death or commited some extreme mistakes that would be already punished if one's opponent is just a bit faster than the mega artropod. With my edit, they will have about 50-60 speed, which is still slower than a basic zombie! And those are supposed to be fodder and tarpit mobs, not posing significant threat to survivor unless one screws up. Arthropod finale evolutions deserve MUCH more than this, being pinnacles of their respective lines.

As for available bullets... There is plenty of ways to get those. Not to mention most current mega arthropods are easily cheesable with long reach weapons such as spears, so your point about ammo is moot.

fairyarmadillo:
These animals are incredibly slow precisely because they're so large. Several of their descriptions even call that out:

> "description": "An enormous hairy mutated spider the size of a cow, with strong forelegs and a raised, bloated thorax.  Its vast increase in size has caused it to lose much of its mobility, and when near it you can smell rotten eggs for some reason…",

>  "description": "A cow-sized mutated grass spider with long spindly legs, light gray in color.  It moves ponderously, depending on its webs to prevent prey from fleeing.",

>  "description": "A cow-sized, hairy mutant wolf spider.  Its mouthparts have noticeably changed, and it moves much more slowly now with the additional weight it needs to carry.",

These are not perfectly evolved predators, they're bloated mutants with runaway genomes that are evolving into dead-ends that will invariably kill them. Their existence serves the blob's needs perfectly - They blow up to incredible size, making the wilds inhospitable for humans fleeing from populated areas, then they cause a biosphere collapse and massive die-off that produces tons of usable biomatter for amalgamations. These animals are not the new rulers of the earth, they're the last generation, just like the player. 

As for what they're eating, many are eating zombies, which willingly walk right up to them. Some are eating other gigantic insects, others (such as the alpha bee) are eating nectar from giant triffid flowers. The giant flies are eating carrion. Additionally several of them, such as the mantis, trapdoor spider, black widow, and locust, are very slow-moving animals in their natural real-life forms. They rely on stealth, webs, or their leaping ability to get around their poor mobility. These abilities are reflected by existing mechanics. Spiders have webs, locusts can jump, mantises have a reach attack.

"Game balance" is not really a great reason to be artificially adjusting their speed up, and I agree with @PatrikLundell that even if it was, it would be the wrong call. If you look at the PRs that added them, the largest mutant varieties are intended to be avoidable, while the intermediate forms are not: https://github.com/CleverRaven/Cataclysm-DDA/pull/45868

fairyarmadillo:
> The biosphere collapse is supposed to happen a tad bit later than less than 1 year in lore, but by your logic it seems to be the opposite. As it is right now in game, by the end of first year many creatures will evolve into their current final form, filling wilderness with those slow-moving gargantuan forms, making that window of existence of actually dangerous forms pretty damn narrow. Exodii lore states the world has at least a decade (and often a few) before it becomes absolutely inhospitable. The time period of game is still in stage where blob needs, or at least desires super critters that can hunt something else than munch on zeds. Still, even if they have to be slower than previous evolutions, the 4x decrease in speed is a little bit too harsh at the very least.

A few species of insect becoming unsustainably large does not immediately trigger a biosphere collapse, it just starts the ball rolling. I agree that the current rate of evolution is too fast, but that's not a problem with the giant insects themselves, more that the midgame needs to be built out so that their evolution can be delayed without ensuring that the player will never see them.

> See my previous statement. Blob is unlikely to be interested in bioweapon that only can effectively munch on zeds and deter survivors instead of being proficient at hunting them.

The blob is not intelligently designing these creatures with that much precision, and it certainly does not think or care about the player. Some zombies getting eaten is not a huge loss.

The fact that they're eating zombies is an example of how their existing biology adapts to the new world, and is reflected in player mutations like rat and chimera, which are also perfectly happy eating tainted flesh.

> That's where your argument contradicts your other arguments. Venera's reworks introduced armored centipedes (110 speed, MUCH faster than any of creatures edited by me) and warehouse spiders (120 speed) as ends of their respective creature lines. Those both are extremely dangerous opponents player usually has to sprint away from, not just leisurely walk away. If going by your arguments, those do not deserve their speed.

I didn't see that armored centipedes got reworked. I have a big problem with that given that they're one of the worst enemies in the game to deal with and their stats feel artificially inflated for video game reasons. That's an issue for another time I guess.

randomtyper:
> I didn't see that armored centipedes got reworked. I have a big problem with that given that they're one of the worst enemies in the game to deal with and their stats feel artificially inflated for video game reasons. That's an issue for another time I guess.

Reworked by person (Venera) whom you described as making evolved critters slow, mind you. I am merely looking at his work and making a stopgap solution in bringing those somewhat up to speed until he makes em even more scary. In fact, why not invite him to the discussion? He did some relevant reworks, and I would love to hear his opinion, and maybe even plans on the matter. @Venera3, I would like to hear from you on the matter of bringing mega mutant arthropods up to the standard you set up in your Centipede, Cellar spider and other reworks. Is speed buff enough, does something else need to be changed, or you better reworking those yourself?

Venera3:
The core problem of the megavolutions is that they are *boring* and omnipresent (since they use evolution every wolf spider will turn into a -*sigh*- necrowolf spider by next year). Bumping up their speed across the board makes them slightly more scary, but still mind-numbingly boring - I'm not a fan of this sweeping `yeah that feels about right` change across the board.

The actual solution is to rework them properly - there's a lot of open design space for interesting/scary/weird monsters, but most megamorphs are just slightly pointless statblock mobs as they stand. FWIW if it's up to me a good half of the remaining megas will get yote somewhere along the line.

randomtyper:
> The core problem of the megavolutions is that they are _boring_ and omnipresent (since they use evolution every wolf spider will turn into a -_sigh_- necrowolf spider by next year). Bumping up their speed across the board makes them slightly more scary, but still mind-numbingly boring - I'm not a fan of this sweeping `yeah that feels about right` change across the board.

So, needs more branching out and variety? So one does not fight exclusively one kind of bugs and then "slightly" larger bugs, but, for example, has to prioritize one kind of bug (centipede matriarch) so infestation does not expand?

> The actual solution is to rework them properly - there's a lot of open design space for interesting/scary/weird monsters, but most megamorphs are just slightly pointless statblock mobs as they stand. FWIW if it's up to me a good half of the remaining megas will get yote somewhere along the line.

There is a decent amount of good beginnings at least in some mobs, for example the missile spider, I'd say. It still needs work, though, and I am not sure I am qualified for it. Well, then, unless you want to rework/yeet those soon, I will try to spice the trees up a bit, if/when I will have time/inspiration.

Venera3:
Keep in mind not every species would need a fuckhuge variant in the first place. If you can come up with a design that needs it, good, if not that's also fine. Other than that and having a design for them I've become fond of using them as rare dead-end offshoots to keep the variety even at later dates.

randomtyper:
> Keep in mind not every species would need a fuckhuge variant in the first place. If you can come up with a design that needs it, good, if not that's also fine. Other than that and having a design for them I've become fond of using them as rare dead-end offshoots to keep the variety even at later dates.

Yeah, I saw that you decreased size for the cellar spider in your rework. Pretty sure I'll do the same to some extent for most varieties (if not yeeting them).

Alright, this issue will stay in draft and under work for now until I will be ready to return to creatures in question and take a good look at them.

PR #66866, opened 2023-07-13
Tick_action as a separate thing from use_action

### LINKED ISSUE #66946: Dropping an active (on) chainsaw, and then doing anything else crashes the game
### Describe the bug

Dropping an active (on) chainsaw, and then doing anything else, crashes the game.

### Attach save file

N/A really, reproduction is per the steps, but i made a test char for a clean world with a chainsaw in hand.
[Testing-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12064531/Testing-trimmed.tar.gz)

### Steps to reproduce

1: Spawn in either a gasoline or electric chainsaw that is off. Fuel\battery if you need.
2: Wield it, and turn it on.
3: Drop it.
4: Move, pause, or do anything that takes time.
5: Game crashes.

### Expected behavior

I expected the game not to crash ^^

### Screenshots

_No response_

[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/12064517/debug.log)

### Additional context

[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/12064518/crash.log)

### LINKED ISSUE #67066: "Unload nearby containers" on a dropped active mp3-player cause SEGFAULT.
### Describe the bug

"Unload all" on a dropped active mp3 player cause segfault.

### Attach save file

N/A

### Steps to reproduce

1. Make player with mp3 player (and light battery).
2. Turn on mp3-player and drop it on the ground.
3. Make a Personal-zone with "onload all" near the player from the zone menu.
4. Select "unload all nearby containers" from the sort zone menu.
5. SEGFAULT crash.

### Expected behavior

N/A

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.14393.6085 (1607)
- Game Version: 2c4c7ee [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda]
]

### Additional context

_No response_

Hirmuolio:
#### Summary
Infrastructure "Tick_action as a separate thing from use_action"

#### Purpose of change
Fix #66946
Fix #67066

`use_action` is triggered when player activates item.  
The same `use_action` is triggered every turn when active tool is processed.
The same `use_action` is also triggered when tool runs out of power.

This results in item actions that must handle multiple different things in single action.

For example `explosion`.

* If activated by player `explosion` displays message to the player on why they can't activate the item.
* If activated in item processing `explosion` displays message, makes noise and consumes charge.
* Finally `explosion` can also create explosions. But most of the time it does the above two.

#### Describe the solution

`tick_action` that is only triggered on item processing. This allows the individual actions to be simpler as they only need to handle one situation.

Migrates the following actions to the new system:
* Toolweapons (chainsaws etc.)
* Molotov cocktail
* Explosions
* Fire weapons
* Music instruments
* Radio
* granade
* firecrackers
* emf detector

Since the processing is separate from player activation the `bool t` will be unused and will be removed later.

The old method of using `use_action` still works so items can be migrated to the new system one by one in smaller pull requests.

`"sound"` and `"manualnoise"` are very similar and maybe just one of them should be used.


Hirmuolio:
I think this is ready for merging now. It does not update all items to use the new system so some potential for conflict exists. But I don't think there are any items left that rely on the old "invoke tool when ammo runs out" anymore which would be the biggest compatibility breaking change.

Individual items can be updated afterwards in smaller simpler pulls. This avoids ballooning this pull request any further.

SurFlurer:
Also will fix #67045.

PR #66871, opened 2023-07-13
Cable rework part 2: Extension cords and more

### LINKED ISSUE #66944: Plugged in tools cause crash when repairing
### Describe the bug

When I try to repair using a soldering iron/arc welder that is plugged in, I get a seg fault. If I repeat the same steps with either tool using batteries, it works as intended. 

### Attach save file

[Tonica-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12064447/Tonica-trimmed.tar.gz)

### Steps to reproduce

1. Have broken item
2. Have arc welder plugged into stationary power grid
3. use the action menu to activate the welder then choose repair
4. choose the item to repair and have the game segfault

### Expected behavior

I should be able to repair with plugged in appliances the same way I can with ones on battery. 

### Screenshots

_No response_

- OS: Linux
- Game Version: 0.G-3026-g30fe61a8fd [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Aftershock [aftershock],
    Magiclysm [magiclysm],
    SpeedyDex [speedydex],
]

### Additional context

[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/12064460/crash.log)

Kamayana:
#### Summary
Features "Cable rework part 2: Adds extending cables with extension cords; adds disconnecting cables from examined vehicle part; improves cable status in item name; other cable improvements"

#### Purpose of change
I didn't implement everything I'd planned for cables in #64334, and there were some issues with the system that came to light after seeing it in action for a few months. This PR is a followup to improve the cable system further.

##### Player-facing changes:

1. Cables and devices can now be extended by extension cords with the "can_extend" parameter in their link_up action. This adds the extension's length to the cable/device, letting it connect from further away. You can keep attaching more cables to make it even longer, and there is no limit to this, though the efficiency of all cables is multiplied together so the more cables attached, the less efficient it gets. Extended cables can be used to connect vehicles as well - you can daisy chain jumper cables to connect vehicles that are further apart, for instance.
2. Linked up items now show more information in inventory menus: the length, max length, and number of extensions, all color coded to how much length is remaining. Unconnected, extended cables show a `-` as the current length, unspooled cables show a `×`. Cables consolidate based on their length and status.
image
3. The "unplug power connections" option when examining appliances has been expanded to be usable on vehicle parts as well, though with vehicles you will only disconnect cables attached to the specific tile you're examining. There are also two separate actions for removing cables and power cables now, since we wouldn't want to disconnect a battery from the entire power grid just to unplug a laptop. The default hotkey for this action is `c`, with the "Fill container with water from faucet" action's hotkey being changed to capital `F`.
4. The "stretched to its limit!" warning will now work better with circular distances. Before, it was possible to break a cable with zero warning, but now you should always get one unless you're moving multiple squares a turn.
5. After detaching/re-spooling a cable, the menu will reopen for convenience, since you will generally only manually detach/respool a cable when you want to reconnect it to something.
6. A cable will show its attached vehicle in the menu for attaching it, if any.
7. JUMPSTART now needs a jumper cable specifically; a pile of meat doesn't have an outlet to plug an extension cord into.
8. Fixes #66944.
9. Disconnecting vehicle cables will now place the cable right in the player's inventory if possible. This does not apply to removing cable parts through the vehicle menu, as that would have to be a more general change for all vehicle part interactions.

##### Back-end changes:

1. Electrical devices now contain their link_data directly, rather than creating an invisible cable item in their CABLE pocket to carry the link_data. This greatly simplifies how they're handled, and the CABLE pocket has been re-worked as the container for cable extensions. This does mean that the "generic_device_cable" item has to be deprecated, so I added an item migration that will turn them into oxygen that'll sit silently inside the CABLE pocket indefinitely, getting deleted if the item ever removes any cable extensions. I included a CABLE_SPOOL flag check in the `cables()` function to prevent cabled items from erroneously showing that oxygen as a cable extension, but that flag check could be safely removed eventually since there's no other way for a non-cable to get added to that pocket than this item migration.
2. When processing a cable that's attached to a vehicle that's out of bounds, the game now loads up the vehicle's submap to assign the cable's vehicle pointer right away, rather than fudge the numbers until the vehicle is loaded by entering the reality bubble. This was accomplished by making the `find_vehicle()` function public.
3. A new activity `invoke_item_activity_actor` was added, which is a simple activity that invokes a specific item, and optionally a specific method, and ends the activity.
4. link_up actions now use `Plug in / Manage cables` as the default menu text, and use an ammo_scale of 0 by default. This lets json for link_up actions be much more concise.
5. `item_contents::clear_pockets_if` was added, a function that lets you easily clear an item's pockets if they meet a chosen lambda function.
6. Fixed a bug that made cable vehicle parts break off the turn AFTER they should have. Fixing this revealed that the power_cable_stretch_disconnect test was relying on that bug to return correct results, so I had to adjust things there as well.
7. `remove_remote_part()` reworked, now there's a `get_remote_part()` function that returns both the remote vehicle and the part as a pair. `remove_remote_part()` now uses that new function to delete the part.
8. `find_vehicle()` was made public, an easy way to load a vehicle outside the reality bubble into memory.
10. `properties_to_item()` was renamed to `part_to_item()` and moved from being a `vehicle_part` function to a `vehicle` one, as it now needs to know the absolute position of the part, which the vehicle keeps track of. All uses of it were replaced appropriately.
11. `shed_loose_parts()` was heavily reworked, as it didn't actually work properly for cables attached to a mount other than [0,0]. It's now far more robust, accurately updating connection points between vehicles when either of them moves or rotates, at any speed, and even beyond reality bubble boundaries.

1. Apart from the efficiency loss, there's no downside to connecting different kinds of cables together. In real life, connecting cables of different gauges/amperage ratings is dangerous and will probably destroy the cords eventually, or even start a fire. It would be possible to simulate such things, but for this PR I mostly just wanted to make a suitable replacement for the vehicle connection daisy-chaining people were doing to connect disparate electric grids.
2. The new disconnect cable option on the vehicle interaction menu uses the `c` hotkey, but when interacting with an appliance it will use the fallback hotkey assignment, which will probably assign it to `a`. `veh_app_interact::populate_app_actions()` has a set of specific commands it assigns hotkeys to, but assigns fallback alphabetical keys to everything else. Now that unplugging cables works for both vehicles and appliances, it can't be included in that list of specific commands anymore, and thus gets a fallback hotkey. I was able to make a fix for this, but it was hacky and felt like taking a shotgun to an issue that needs more nuance, so I ultimately let it be.
3. When reopening the cable interaction menu, sometimes the progress menu will be stuck at the top of the screen. This seems to be an issue related to actions that are based on speed instead of time, and fixing it would be out of scope.

#### Testing
[Device Cables Stress Test-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12037143/Device.Cables.Stress.Test-trimmed.tar.gz)
A helicopter with many, many cables attached, used for testing that everything stays connected even when flying at high speeds, and stress testing to see the CPU impact. The results after this PR were much the same as the results for the first cable PR.

[Device Cables All Connections-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12037182/Device.Cables.All.Connections-trimmed.tar.gz)
A save created in a recent experimental version (e284941) with every kind of cable connection active. Used to ensure this PR's changes can handle saves made after #64334. Vehicle connections and cable items stay linked, but electrical devices do not. This was an expected outcome, and any attempt to somehow migrate the links wouldn't be worth it in my opinion, since players could easily just re-link the devices.

#### Additional context

This PR is in draft so after #66847 gets merged I can adjust it accordingly. My sincere thanks to everyone who added link_up actions to items, something I said I would do myself before getting lost in the weeds of making this PR. I made sure to not invalidate their work.


Nebnis:
This was an amazing and very much needed PR for the electrical grid thank you very much for doing it :). It will make doing static bases a lot more fun now

data/raw/keybindings/vehicle.json
```diff
@@ -137,7 +137,7 @@
     "type": "keybinding",
     "category": "VEHICLE",
     "name": "Fill container with water from faucet",
-    "bindings": [ { "input_method": "keyboard_any", "key": "c" } ]
+    "bindings": [ { "input_method": "keyboard_any", "key": "F" }, { "input_method": "keyboard_code", "key": "f", "mod": [ "shift" ] } ]
```
Qrox, line 140:
```suggestion
    "bindings": [ { "input_method": "keyboard_char", "key": "F" }, { "input_method": "keyboard_code", "key": "f", "mod": [ "shift" ] } ]
```

src/iuse_actor.cpp
```diff
@@ -4317,24 +4318,41 @@ void link_up_actor::info( const item &, std::vector<iteminfo> &dump ) const
...
+        //~ Power in Watts. %+4.1f is a 4 digit number with 1 decimal point (ex: 4737.3 W)
+        dump.emplace_back( "TOOL", _( "<bold>Wattage</bold>: " ), string_format( _( "%+4.1f W" ),
+                           units::to_milliwatt( charge_rate ) / 1000.f ) );
+    }
+
+    if( !can_extend.empty() ) {
+        std::vector<std::string> cable_types;
+        cable_types.reserve( can_extend.size() );
+        for( const std::string &cable_type : can_extend ) {
+            cable_types.emplace_back( cable_type == "ELECTRICAL_DEVICES" ? "electrical device cables" :
```
Qrox, line 4351:
```suggestion
            cable_types.emplace_back( cable_type == "ELECTRICAL_DEVICES" ? _( "electrical device cables" ) :
```

bombasticSlacks:
Just reviewed the code and everything LGTM one conflict to be resolved and then I'm happy to merge :)

andrei8l:
<details>
<summary>I used an extension cord to extend a smartphone's cable and the smartphone didn't change size or weight.</summary>

Screenshot from 2023-07-21 08-08-01

</details>

src/inventory_ui.cpp
```diff
@@ -86,9 +86,12 @@ item_name_t &get_cached_name( item const *it )
 {
     auto iter = item_name_cache.find( it );
     if( iter == item_name_cache.end() ) {
+        const std::string name = remove_color_tags( it->tname( 1, false, 0, true, false ) );
+        const std::string name_full = string_format( "%s%d%d",
+                                      remove_color_tags( it->tname( 1, true, 0, true, false ) ),
+                                      it->max_link_length(), it->link_length() );
```
andrei8l, line 92: Isn't this redundant? The items that have a link up already include link length/max length in their name and the other ones get stacked so sorting never comes into play.

src/map.cpp
```diff
@@ -1305,6 +1305,32 @@ vehicle *map::veh_at_internal( const tripoint &p, int &part_num )
     return const_cast<vehicle *>( const_cast<const map *>( this )->veh_at_internal( p, part_num ) );
 }
 
+vehicle *map::find_vehicle( const tripoint_abs_ms &where ) const
```
andrei8l, line 1308: Why move this? It's kinda deceptive in `map` if it actually looks in the global mapbuffer. It was fine as a static function in vehicle.

(also, this function is pretty freaking crazy, but that's not on you)

src/item.cpp
```diff
@@ -13047,151 +13111,169 @@ bool item::process_cable( map &here, Character *carrier, const tripoint &pos, it
...
         // Around 85% efficient by default; a few of the discharges don't actually recharge
-        if( battery_deficit == 0 && !( short_time_passed && one_in( link->charge_efficiency ) ) ) {
-            linked_item.ammo_set( itype_battery, linked_item.ammo_remaining() + transfer_total );
+        if( battery_deficit == 0 && !( short_time_passed && rng_float( 0.0, 1.0 ) > link->efficiency ) ) {
+            ammo_set( itype_battery, ammo_remaining() + transfer_total );
         }
     } else {
         // Around 85% efficient by default; a few of the discharges don't actually charge
-        if( !( short_time_passed && one_in( link->charge_efficiency ) ) ) {
+        if( !( short_time_passed && rng_float( 0.0, 1.0 ) > link->efficiency ) ) {
```
andrei8l, line 13195:
```suggestion
        if(  !short_time_passed || rng_float( 0.0, 1.0 ) <= link->efficiency  ) {
```
Reference: https://clang.llvm.org/extra/clang-tidy/checks/readability/simplify-boolean-expr.html

src/item.cpp
```diff
@@ -13047,151 +13111,169 @@ bool item::process_cable( map &here, Character *carrier, const tripoint &pos, it
...
     int transfer_total = short_time_passed ? 1 :
-                         ( turns_elapsed * 1.0f / link->charge_interval ) * ( 1.0 - 1.0 / link->charge_efficiency );
+                         ( turns_elapsed * 1.0f / link->charge_interval ) * link->charge_interval;
 
     if( power_in ) {
         const int battery_deficit = linked_veh.discharge_battery( transfer_total, true );
         // Around 85% efficient by default; a few of the discharges don't actually recharge
-        if( battery_deficit == 0 && !( short_time_passed && one_in( link->charge_efficiency ) ) ) {
-            linked_item.ammo_set( itype_battery, linked_item.ammo_remaining() + transfer_total );
+        if( battery_deficit == 0 && !( short_time_passed && rng_float( 0.0, 1.0 ) > link->efficiency ) ) {
```
andrei8l, line 13190:
```suggestion
        if( battery_deficit == 0 && ( !short_time_passed || rng_float( 0.0, 1.0 ) <= link->efficiency ) ) {
```

src/item.cpp
```diff
@@ -10348,20 +10367,13 @@ int item::ammo_remaining( const Character *carrier, bool cable_links ) const
...
-            if( cable->link->t_veh_safe ) {
-                ret += cable->link->t_veh_safe->connected_battery_power_level().first;
-                continue;
-            } else {
-                const optional_vpart_position vp = get_map().veh_at( cable->link->t_abs_pos );
-                if( vp ) {
-                    ret += vp->vehicle().connected_battery_power_level().first;
-                    continue;
-                }
+    if( include_linked && link_length() >= 0 && link->efficiency >= 0.001f ) {
```
andrei8l, line 10370: minor nitpick, but please make the `0.001f` a constexpr somewhere since it's used in several places with no obvious justification.

src/inventory_ui.cpp
Kamayana: The length/max_length aren't actually a part of tname, they're an inventory-only addition to the name like ammo/max_ammo, so I need to add them to the internal name so the inventory only consolidates cables with the same values. Maybe I should add a comment explaining that?

andrei8l: Ok, I thought the cable length bit was in `tname()` instead of `display_name()`. But still:

You're handling cable lengths for stacking [here](https://github.com/CleverRaven/Cataclysm-DDA/blob/823fa0be255841a2d24dd89ca42f83276c40ce0f/src/inventory_ui.cpp#L1280-L1281) and for collation [here](https://github.com/CleverRaven/Cataclysm-DDA/blob/823fa0be255841a2d24dd89ca42f83276c40ce0f/src/inventory_ui.cpp#L1398-L1399C58).

This cached name is only used as a sort key and has no impact on the above two. If you want to sort by remaining length, then this is the wrong place for it because it's ASCII sorting so it looks like this
Screenshot from 2023-07-21 20-25-03
It can be done properly in the preset's `sort_compare()`.

Kamayana:
Cable extensions now add to volume and weight. If this makes a device too large/heavy for its pocket, it'll first try to relocate to another pocket. If that fails, it'll ask if the player wants to continue, and then spill the device into the parent container or onto the floor.

I'm converting back to draft for the time being because refining the inventory sorting is going to take a little longer.

bombasticSlacks:
Thanks for your continued work on this 😄

Kamayana:
Linked items now sort by link status first and then by length from shortest to longest:

image

Once I take a look at `find_vehicle()` I'll open it up for review again (I may have forgotten about that for a moment, oops).

Kamayana:
> > Linked items now sort by link status first, then by maximum length, then by current length from shortest to longest:
> > image
> 
> Is this intended to still be in draft or is it ready for merge?

It's intended, there's still more incoming. It started as just moving find_vehicle back to a static vehicle function as [andrei8l](https://github.com/andrei8l) suggested but it snowballed into more significant rewrites as I discovered out-of-bubble vehicle connections didn't actually work for anything but appliances. Nearing completion on that, though, and then I can undraft this.

Kamayana:
Significant reworks here:
1. Finding the maximum length of an unconnected, extended cable used to be laborious, having to fetch the link_up iuse actor data, then iterate through all the contained cables and adding their lengths. It's now much more efficient, as it keeps the link_data active for any item with extensions and just returns link->max_length.
2. `remove_remote_part()` reworked, now there's a `get_remote_part()` function that returns both the remote vehicle and the part as a pair. `remove_remote_part()` now uses that new function to delete the part.
3. `find_vehicle()` was switched back from map to a static vehicle function, though it's now public.
4. `properties_to_item()` was renamed to `part_to_item()` and moved from being a `vehicle_part` function to a `vehicle` one, as it now needs to know the absolute position of the part, which the vehicle keeps track of. All uses of it were replaced appropriately.
5. `shed_loose_parts()` was heavily reworked, as it didn't actually work properly for cables attached to a mount other than [0,0]. It's now far more robust, accurately updating connection points between vehicles when either of them moves or rotates, at any speed, and even beyond reality bubble boundaries.
6. Disconnecting cables will now place the cable right in the player's inventory if possible. This does not apply to removing cable parts through the vehicle menu, as that would have to be a more general change for all vehicle part interactions.

I think this is ready now, assuming the tests are green. Probably needs another round of reviews on the new changes, though.

PR #66872, opened 2023-07-13
Police Robot Family Obsoletion

Karol1223:
#### Summary
None

#### Purpose of change
For a longer while there have been discussions of how our police bots not only no longer fit the setting at all but also work in ways that don't make sense if they did fit in the first place. Granted, the second point alone wouldn't be enough to obsolete them but as it stands they are currently relics of the old setting.

#### Describe the solution
**MONSTERS**
- [x] Obsolete the Eyebot
- [x] Make sure the monsters that copy from Eyebot aren't immediately broken (Couldn't check all of their functionality but what they previously copied by hand they now have copied into their own JSON. Odds of this somehow breaking are nigh zero)
- [x] Make sure the Eyebots aren't present in mongroups and mapgen
- [x] Obsolete the Police Bot
- [x] Make sure Police Bots aren't present in mongroups and mapgen
- [x] Obsolete the Riot Bot
- [x] Make sure Riot Bots aren't present in mongroups and mapgen
- [x] Make Hub's Camspies deactivate into their own item instead of copying the Eyebot

**ITEMS**
- [x] Obsolete the items connected to Eyebots
- [x] Obsolete the items connected to Police Bots
- [x] Obsolete the items connected to Riot Bots
- [x] Add a new inactive item for the Camspy

**CPP**
- [x] Obsolete the hardcoded mentions of Eyebots
- [x] Obsolete the hardcoded mentions of Police Bots
- [x] Obsolete the hardcoded mentions of Riot Bots
- [x] Remove code responsible for ``WANTED`` and ``ROBOT_ATTACK`` events

**MOD CONTENT**
- [x] Make sure obsoleting mainline content doesn't break anything
- [x] Port the robots and their connected items to Aftershock

**MISCELLANEOUS**
- [x] Remove some ``PARROT`` snippets that feel like they're very much RoboCop - without Eyebots they make no sense for MiGos to use those
- [x] Remove some flier, newspaper and tip snippets referencing Eyebots
- [x] Changed the mission dialogue referring to Camspies as "Eyebots with a different coat of paint"
- [x] Edit the NPC background referencing Eyebots
- [x] Make the broken item for the Camspy ``NO_SALVAGE``

There's been many discussions of including more variants of camspies or drones, however, the existence of the Eyebot is not necessary (and I'd hazard a guess moving the functions to the Hub camspy is actually gonna make further conversions more comfortable). I opted not to do any of those, though I hope that I can do this PR in such a way where further theoretical implementation of them won't require digging through git blame to find dead code.

#### Additional context
The world would be a better place if people threw their dictionary edits through an autosorter so my PRs didn't look like a mosaic whenever I do that

Big thanks to @chichit1044 for doing the backstory as I'd have gone insane trying to make that work myself


NetSysFire:
What about the riot bot? The one that shocks you, emits relaxation gas and tries to arrest you

Karol1223:
> What about the riot bot? The one that shocks you, emits relaxation gas and tries to arrest you

The current goal list is very much WIP of what I just could think of on top of my head. I know there's AT LEAST three monsters I need to obsolete (Riot Bot was the third one whose name I forgot, unironically thank you for reminding me the name lol), alongside more items, likely some snippets and a plethora of hardcoded bullshit. I'll add most of them to the list as I go around finishing them up or remember specific things I need to note down not to forget.

Karol1223:
> Current code makes `timed_event_type::WANTED` do nothing. Just out of curiosity, are you planning to add some other consequences for the player character becoming "wanted", or just remove that event and all related code?

Realistically, I can't think of any possible use cases for it *unless* we make some NPCs with specific backstories have different responses? I'm likely going to remove it as a whole, though, unless someone steps in and tells me to keep it.

CocoBatPone:
I'll very much miss the robots, I've always liked DDA being somewhat futuristics and the robots and such have always been a part of that for me, will these re-appear in a mod like aftershock, or a seperate mod?

Karol1223:
> I know the PR says they could be moved to Aftershock, but if that never happens I'd like them somewhere else

I got greenlight from Aftershock maintainers to move them there so I will. This is not an ideal solution for Aftershock but it has a lot of content that requires these robots to exist so for the time being they'll be ported over in their entirety.

NetSysFire:
Ah so the other bots, mainly the utility ones, already had their description altered and are made by Uncanny now.  Though what of the hazmat and mole bot? The former is mainly a relic of the hazardous waste sarcophagus afaik.

Karol1223:
> Ah so the other bots, mainly the utility ones, already had their description altered and are made by Uncanny now. Though what of the hazmat and mole bot? The former is mainly a relic of the hazardous waste sarcophagus afaik.

I'm very meh about most of the robots we have but I'm only tackling the police lads this PR. If I tried to do all of them not only would I have to think more but the reviewers would probably avoid this like hellfire.

bombasticSlacks (approved):
The C++ looks reasonable for this, I don't think these are serialized in a way that things can break.

Issue #50279, opened 2021-07-28
[0.F stable] Map generate lamp posts blocking dump gates


PatrikLundell:
### Describe the bug

A dump site found had a vehicle in it (behind locked gates). The vehicle was functional, so I picked the gates and drove it out, resulting in a crash as it hit the lamp post on the sidewalk outside.

### Steps To Reproduce

1. Generate a world
2. Find a dump site at a city street

or look at the attacked save

### Expected behavior

Map generation shouldn't generate combinations that conflict.

### Versions and configuration

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: bc28921 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Bionic Professions [package_bionic_professions]
]

### Additional context

The entrance to the dump should probably be at one end of the tile rather than the center. I'd also suggest examining other tiles with exits to verify they don't have those at the center.

PatrikLundell:
Another example of a different clash from the same game (but different town):
![Screenshot (1)](https://user-images.githubusercontent.com/22739822/127573919-07674ce4-4eff-4c47-b656-2934c36931b1.png)
In this case it's possible to pass by to the side of the street lamp, but it still doesn't make real world sense,

actual-nh:
@eltank: I wonder if the obstacle detection code in the automatic-driving PR could be used to help detect situations like this (and others like that one with gates too narrow for farm vehicles or whatever it was) - prior to putting in things like crashed vehicles that could reasonably have happened later, make sure all the places that intact vehicles can spawn actually have a way to get them out and, say, to the nearest road (assuming road or amphibious).

eltank:
While trying to understand/reproduce the bug, I found a pickup truck in a dump site. With careful maneuvering I was able to get it out without crashing into the light post or anything else.
Autodrive will eventually do this careful maneuvering for you (right now autodrive is disabled when you're not standing on a road tile to begin with).
But even if I had to smash some fences to get the vehicle out, I would only consider that a minor inconvenience if I was in desperate need of a vehicle.

> I wonder if the obstacle detection code in the automatic-driving PR could be used to help detect situations like this

Maybe, but that's pretty expensive in terms of CPU cycles and also far from trivial.

[...]

PR #66873, opened 2023-07-13
Allow nests to check for neighbouring flags

### LINKED ISSUE #66759: Nest placement to be able to check for overmap terrain flags such as SIDEWALK

I'm trying my hand at road's maps unhardcoding and adding some new stuff like roundabouts but without a way to detect SIDEWALK in json it currently needs multiple declarations of arrays with every SIDEWALK map in to work which obviously isn't sensible

### Solution you would like.

Either adding it as it's own method of checking like joins are currently or adding it as a possible way of searching to ot_match_type would work though the latter applies in more places for better and worse
Example syntax A as an addition to the ot_match_type enum:
```
"neighbors": { "north_east": { "om_terrain": "SIDEWALK", "om_terrain_match_type": "FLAG" } }
```
Example syntax B like "joins":
```
"flags": { "north_east": { "SIDEWALK" } }
```

_No response_

### Additional context

_No response_

Procyonae:
#### Summary
None

#### Purpose of change

Checking flags with nests is necessary for road map unhardcoding to check SIDEWALK and should allow some other interesting possibilities

Fixes #66759

#### Describe the solution

Syntax is identical to `"neighbors"`, you just use `"flags": { "north_east": [ "WILDERNESS", "RIVER" ], "below": [ "LAB" ] }` as an example

I were strugglebussing but @bombasticSlacks came to the rescue so the function is entirely their work, if they want me to close this so they can post it that's cool

I might have a crack at allowing combinations of &= and |= rather than it just always being |= maybe using syntax like
`"flags": { "north_east":  [ { "flag": "FLAGA" }, { "flag": "FLAGB", "operator": "AND" }, { "flag": "FLAGC", "operator": "OR" } ]`
to achieve `( ( 0 | FLAGA ) & FLAGB ) | FLAGC`
Equally being able to optionally have `||`s in `neighbor_oters.test( dat ) && neighbor_joins.test( dat ) && neighbor_flags.test( dat )` could be useful

#### Testing

Tested basic functionality locally

#### Additional context

Removed the .check() because flags get checked when they're initially read so unless I'm stoopid there's no way an invalid one could get this far

image
Pre-existing error message if you use a nonsense flag

### DIFF
--- a/doc/MAPGEN.md
+++ b/doc/MAPGEN.md
@@ -1136,6 +1136,7 @@ Place_nested allows for limited conditional spawning of chunks based on the `"id
 | x and y            | (required, int) the cardinal position in which the chunk will be placed.
 | neighbors          | (optional) Any of the neighboring overmaps that should be checked before placing the chunk.  Each direction is associated with a list of overmap `"id"` substrings.
 | joins              | (optional) Any mutable overmap special joins that should be checked before placing the chunk.  Each direction is associated with a list of join `"id"` strings.
+| flags              | (optional) Any overmap terrain flags that should be checked before placing the chunk.  Each direction is associated with a list of `oter_flags` flags.
 
 
 The adjacent overmaps which can be checked in this manner are:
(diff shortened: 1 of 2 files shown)


ashGlaw:
What sorts of possibilities does this open up?

Procyonae:
> What sorts of possibilities does this open up?

I don't think it technically makes anything possible that wasn't already it just provides a different way to do it and makes some previously unmergeable stuff mergeable. Looking at #66941 for instance every place"flags": { "direction": [ "SIDEWALK" ] } is used would instead of had to be "neighbors": { "direction": [ "map_with_sidewalk", "another_map_with_sidewalk" ... "millionth_map_with_sidewalk" ] } which obviously isn't sensible nor maintainable. I'm looking to use it specifically to unhardcode the road maps and solve to #50279 but it could be used in mutable specials like new labs to save a lot of time and make things easier to understand and maintain.

PR #66877, opened 2023-07-13
Fix pathfinding around vehicles

ehughsbaird:
#### Summary
Bugfixes "Fix pathfinding around vehicles"

#### Purpose of change
I noticed that my NPC companions were no longer very good at getting in vehicles when I entered them. Sure, they would walk up to the side of the vehicle, but then they would sit there and make no attempt to go through the doors to get inside.

It turns out, pathfinding was now considering every vehicle tile a door, so my companions were trying to open and walk through the walls of my vehicle to get to me.

#### Describe the solution
This is due to a bug in https://github.com/CleverRaven/Cataclysm-DDA/pull/61087, where the return value of `vehicle::next_part_to_open` and `vehicle::next_part_to_unlock` is treated as though it will be 0 (false) when the tile cannot be opened/unlocked. However, these functions return a vehicle part index, and as 0 is a valid vehicle part index, they return -1 when the tile cannot be opened/unlocked.

To fix, check for the correct value, instead of a non-zero value.

#### Testing
This behavior is easily demonstrated with a test vehicle: (I used an ambulance)
```
 """"  " = windshield, + = door, - = quarterpanel, # = seat
 +!#+  ! =  player character, @ = npc companion
@----
```

A test has been added to demonstrate this scenario.

To get into the seat next to the player, the NPC will try to walk through the quarterpanel to the east of it, when they actually need to go around the vehicle and enter through the other door.

### DIFF
--- a/src/pathfinding.cpp
+++ b/src/pathfinding.cpp
@@ -331,10 +331,10 @@ std::vector<tripoint> map::route( const tripoint &f, const tripoint &t,
                         int dummy = -1;
                         const bool is_outside_veh = veh_at_internal( cur, dummy ) != veh;
 
-                        if( doors && veh->next_part_to_open( part, is_outside_veh ) ) {
+                        if( doors && veh->next_part_to_open( part, is_outside_veh ) != -1 ) {
                             // Handle car doors, but don't try to path through curtains
                             newg += 10; // One turn to open, 4 to move there
-                        } else if( locks && veh->next_part_to_unlock( part, is_outside_veh ) ) {
+                        } else if( locks && veh->next_part_to_unlock( part, is_outside_veh ) != -1 ) {
                             newg += 12; // 2 turns to open, 4 to move there
                         } else if( part >= 0 && bash > 0 ) {
                             // Car obstacle that isn't a door
--- a/tests/npc_test.cpp
+++ b/tests/npc_test.cpp
@@ -22,6 +22,7 @@
 #include "memory_fast.h"
 #include "npc.h"
 #include "npc_class.h"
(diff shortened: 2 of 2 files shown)


Kamayana:
The `npc-board-player-vehicle` test looks like it's consistently failing in the MacOS Clang suite:
[Clang 14, macOS 12, Tiles, Sound, x64 and arm64 Universal Binary](https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/5562065102/jobs/10160134757?pr=66920#logs)
[Clang 14, macOS 12, Tiles, Sound, x64 and arm64 Universal Binary](https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/5563626062/jobs/10162699417?pr=65549#logs)
[Clang 14, macOS 12, Tiles, Sound, x64 and arm64 Universal Binary](https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/5563666647/jobs/10162866673#logs)

```
(~[slow] ~[.],starting_items)=> ../tests/npc_test.cpp:386: FAILED:
(~[slow] ~[.],starting_items)=>   REQUIRE( companion->pos() == player_pos + point( 2, 0 ) )
Error: (~[slow] ~[.],starting_items)=>   (61,62,0) == (62,60,0)
(~[slow] ~[.],starting_items)=> Log messages during failed test:
(~[slow] ~[.],starting_items)=> 12:00:33AM: Beta NPC dies!
(~[slow] ~[.],starting_items)=> 12:00:33AM: Unisex Family eats a holy SPAM of debugging (fresh).
```

PR #66752, opened 2023-07-08
on_damage EOCs

bombasticSlacks:
#### Summary
None

#### Purpose of change
Request from discord, adds an on damage EOCs for damage types

#### Describe the solution
Added it post absorb where monsters and characters have been handled.
Added documentation.
Added tests.

#### Testing
Unit tests

PR #66879, opened 2023-07-13
[MoM Add telepathic and telekinetic damage types with appropriate immunities

Standing-Storm:
#### Summary
Mods "Add telepathic and telekinetic damage types"

#### Purpose of change

Telepathy should realistically not be usable on a big chunk of the monsters survivors fight in CDDA--zombies, robots, some nether monsters, the mycus, etc. There's no easy way to implement this and this PR doesn't go all the way to do so, but it does provide a partial solution.

#### Describe the solution

Implement two damage types: telepathic and telekinetic.  Telekinetic damage is derived from bash damage, though a few monsters have higher levels of telekinetic armor (shoggoths, flying polyps, telekinetic ferals, mice of NEMA, the thing, etc.). Telepathic damage has no derivation but quite a few monsters are flatly immune to it, including species-level flag immunities for robots, zombies, mycus, leech plants, slimes, nether emanations (portal storm creatures), and worms, plus copy-from immunities for many other nether creatures.  You can still telepathically assault blank bodies, krecks, gracken (you monster) and a few others.  Damage was also distributed to some monsters that should use it.

- [X] Distribute damage to remaining monsters.

Pyrokinesis did not receive its own damage type because heat already handles that.

#### Testing

Telepathic Shield makes you immune to telepathic damage, Inertial Barrier makes you immune to telekinetic damage.
Appropriate monsters are armored against appropriate damage types. on_damage EoCs work (though they still seem to apply the effects even against targets that are immune to the damage). 

#### Additional context

https://github.com/CleverRaven/Cataclysm-DDA/pull/66752 is merged so this should be ready.


Maleclypse:
So this may not be how you view it, but I would give Mycus half damage from telepathic damage.   They are a hive mind that communicates with it's multitude of bodies in some kind of way and/or give telepathic attack spells a telepathic disruption stunning flag  that will stun things with the species flag such as the mycus.  There's a couple of different directions that could go and you may hate the suggestion. Let me know if you want to ignore or think about the suggestion but still merge this PR.  I gotta say that you inspire me as a modder myself with your output and ideas.

Standing-Storm:
My initial idea was based precisely on the idea that since the mycus is a hive-mind, the overwhelming psychic force of the hive mind would make it immune to telepathy, but maybe it makes more sense to have a hierarchy where (frex) the  giant fungal blossom is immune to telepathy but fungal-infected ferals have a small amount of telepathic armor--fungal-infected creatures have the lowest telepathic armor, then actual fungaloids, scaling up to the queen.

Since I gave all telepathic attacks a chance to daze or stun, as long as the target takes any damage at all they are subject to that chance, which in humans would represent having their mind momentarily scrambled but here could be a brief disruption of their connection to the hive-mind. 

Or I could add a "HIVE_MIND" flag to mycus (and mi-go--aren't they mentally connected, or am I misremembering?) and make them immune to the normal chance to stun _but_ give them a separate chance to be stunned that, if it works, lasts for longer, representing their confusion and disorientation at the hive-mind suddenly falling silent. 

...I like that last one a lot, actually. Time to get to work.

Standing-Storm:
Alright, the mycus and migo are now resistant, but not immune, to telepathic damage. 

There's a little weirdness with the way ondamage_eocs trigger but the test_data file I included is so I can do a proper bug report around it.

PR #66880, opened 2023-07-13
Make mattresses deployable in tents

ashGlaw:
#### Summary
Features "Make mattresses deployable in tents"

#### Purpose of change

It bothered me that you could put out a cot or a rollmat in a tent, but you couldn't just slap a mattress down on the groundcloth of the tent. They're not queen or king sized mattresses, they'd fit in a tent of the sizes we represent in game.

#### Describe the solution

It seemed strange to me that some furniture could be placed in a tent, but most couldn't. Lo and behold, none can. The ones that can go in tents? They're traps! Not furniture at all. This game is such a kludge sometimes.

So I added a new trap version of the mattress, using the rollmat and cot as examples and cross referencing the furniture version of a mattress, to ensure they're similar as I can get them. Then I changed the entry for mattresses in data/json/items/resources/misc.json to have it's use_action make a trap instead of furniture. I didn't realize at first that the mattress needed the single_use flag, which lead to deploying the trap not removing the mattress item you used to deploy it. Whoops! Fixed that. 

I did down mattresses as well, because why not? I also added longest_side dimensions to both mattresses, after a quick google search. May as well while I'm here, right?

I went with this option because the alternative would be to change the Cpp to make furniture a vector, like itemstack. But that would be a BIG pain in the ass because of how integrated furniture is. It would be totally doable, with some help, but I only sorta understand the concepts to begin with, so going with traps seemed much more expedient. A big change like that would require notice and discussion and buyin that's far, FAR beyond the scope of this little PR.

#### Testing

I spawned the mattresses and a large tent. I deployed the tent and then deployed in each mattress. I tried to sleep in each and got the "this is a comfortable place to sleep" message with each. The mattress traps come down when undeployed and become items once more.

#### Additional context

N/A


Night-Pryanik:
While the results are satisfactory, I think that the whole idea of benign traps was flawed from the start. All these cots and funnels (previously also braziers) - and now also mattresses - implemented as traps is a sign of bad design. I think we should rework them all into proper entities and remove benign traps as a class, not add new ones. That's only my opinion, though.

ashGlaw:
I absolutely agree in principle. I'm adding technical debt, sadly. I recognize that, but I'm not prepared yet to take on fixing a problem of that magnitude.

PR #66889, opened 2023-07-13
Inline `sm_to_ms_copy` for multi% cpu wins in congested places.

akrieger:
#### Summary
None

#### Purpose of change

I noticed during a perf trace of a different PR that these functions were showing up kinda high up in total usage. @SurFlurer tried inlining them and found that it seemed to basically eliminate their cpu overhead. I poked at it a bunch until I felt confident the effect was real and not a cost shift. Also SurFlurer did the legwork to validate the extra include would be a noop in terms of extra build complexity.

#### Describe the solution
Inline them in the header.

#### Testing
I created a repro condition with a fresh game. Spawned in a setting with a bunch of NPCs. Debug killed all monsters. Created a cage of reinforced glass doors outside the shelter. Spawned two wasps inside the cage. Saved, then instrumented waiting here for 3 hours 7 times before/after. Discarded the fastest and slowest one for each based on total samples inside `map::getlocal` (the primary caller to `sm_to_ms_copy`.). Then compared the cost of callers of `map::getlocal` because the effect was so dramatic that I wanted to validate higher order effects. Concluded my samples were reproducible enough to feel confident in the results. Savings in this test case were ~3.7% total cpu from `Creature::pos` (essentially the only caller of `map::getlocal`), 2.5% from `npc::assess_danger` in this example. I also anecdotally saw similar but more noisy results with the same setup but with cows instead of npcs.

#### Additional context
This is a comparsion before/after, comparing the 'fastest' before case to the 'slowest' before case, so any other comparison would show more savings than pictured.
image

### DIFF
--- a/src/coordinate_conversions.cpp
+++ b/src/coordinate_conversions.cpp
@@ -143,16 +143,6 @@ point ms_to_sm_remain( int &x, int &y )
     return point( divide( x, SEEX, x ), divide( y, SEEY, y ) );
 }
 
-point sm_to_ms_copy( const point &p )
-{
-    return point( p.x * SEEX, p.y * SEEY );
-}
-
-tripoint sm_to_ms_copy( const tripoint &p )
-{
-    return tripoint( p.x * SEEX, p.y * SEEY, p.z );
-}
-
 void sm_to_ms( int &x, int &y )
 {
     x *= SEEX;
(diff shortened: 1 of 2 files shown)


akrieger:
Sure, I can do that tomorrow.

SurFlurer:
Thank you!

akrieger:
Done.

PR #66890, opened 2023-07-13
Jmath functions for mapgen

bombasticSlacks:
#### Summary
None

#### Purpose of change
revives #66409

#### Describe the solution
Adds the ability for mapgen to calc weights of mapgen_functions on the fly using a jmath function to effect the base weight. 

#### Testing
Tested with 

```
{
    "type": "jmath_function",
    "id": "weight_test",
    "num_args": 0,
    "return": "10000"
  },
```
and a junk spawn location.


ashGlaw:
So, I'm curious, what does this do? It allows for the changing of the probability of certain over map tiles to spawned based on what's already spawned? I'm not sure I get it.

Procyonae:
> So, I'm curious, what does this do? It allows for the changing of the probability of certain over map tiles to spawned based on what's already spawned? I'm not sure I get it.

Basically you can define a jmath function that can change a map variants weight (the chance for it to spawn compared to another using the same id) using variables like time_since_cataclysm or global variables from quests/faction interaction to either stop/start a map variant spawning entirely depending on the variables or to change the ratio of map variants over time. Ideally I'll figure out how to port it over to a few other places now bombasticSlacks has done the hard bit.
It's a little more rudimentary than what I were aiming for because it doesn't allow arguments but that just means the jmath functions have to be more specific and less generalised.

ashGlaw:
Thanks! That's a great add!

bombasticSlacks:
Gotta figure out a linking error

Issue #62739, opened 2022-12-15
Nested recipe lag


oosyrag:
### Describe the bug

There is a noticeable stutter/freeze when scrolling through recipes upon passing a nested recipe group in the crafting. I'm completely ignorant of the code behind displaying the nested recipes, but I can't help but wonder if there is some optimization to be found here. It's more pronounced on my phone (Android, OnePlus 7t, 2019) than my desktop (Win10, i3-12100F, 2022), but still noticeable even on desktop. I suspect it will be worse on lower end systems.

### Steps to reproduce

Open crafting menu, hold down a key to scroll through normal recipes until a nested recipe is passed. Witness stutter/lag/freeze. It is worse for multiple nests or nests with many items inside, such as steel plate armors and dusters. It happens regardless if the nested recipes are collapsed or expanded.

### Expected behavior

No noticeable lag would be nice. 

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.2251 (21H2)
- Game Version: e40157c [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

mqrause:
/confirmed

![grafik](https://user-images.githubusercontent.com/38702195/207823195-5148fc0a-1dea-4cba-9524-8f333624d05c.png)

It's this lookup when getting the description of the nest. The fewer recipes from that nest you know, the worse it gets. If you know all the recipes, it actually scrolls faster than normal recipes.

PR #66894, opened 2023-07-14
Fix Nested List lag in crafting menu

lispcoc:
#### Summary
Performance "Fix Nested List lag in crafting menu"

#### Purpose of change

Showing nested recipe does not use cache in spite of cache exists.

#### Describe the solution
use cache available_recipes.

#### Testing
Recipe is displayed collectly. Performance is improved.


RenechCDDA:
I believe that this might resolve #62739?

PR #66897, opened 2023-07-14
Generate vehicle prototypes from vehicles in game

SurFlurer:
#### Summary
Features "Generate vehicle prototypes from in-game vehicles"

#### Purpose of change
Make it easier for modders and contributors to add / edit vehicle prototypes. Having a way to visually create / edit vehicle prototypes is mentally more comfortable and accelerates iteration for modding.

#### Describe the solution
Make a debug menu entry to generate a vehicle prototype from any vehicles in the game.

- [x]  Make the debug menu entry.
- [x] Support exporting the blueprint (appearance) of a vehicle.
- [x] Support exporting the vehicle parts of a vehicle.
- [x] Support for exporting top-level items in the vehicle. ( items in containers are ignored. )
- [x]  Support exporting loot zones bound to the vehicle.
- [x] Add a test unit.
- [x] Add doc.

#### Testing
Spawned a surv_car, export it as a prototype, then try to spawn the exported prototype. The two vehicles are identical except for there's no items in the exported one ( as expected ).

#### Additional context

Exporting itemgroups seems to be impossible, so I'm not doing this.

Exported prototype needs to be formatted by the json formatter. Is there a way to format that while printing?


Procyonae:
Would making a 1 symbol blueprint of just "at least one vehicle part"/"no vehicle parts" work to encourage peeps to make a blueprint? You could potentially make the 0,0 it's own symbol too if that's not the top left corner. Even just an empty blueprint to encourage peeps to fill it in might be better than nothing.

SurFlurer:
> Would making a 1 symbol blueprint of just "at least one vehicle part"/"no vehicle parts" work to encourage peeps to make a blueprint? You could potentially make the 0,0 it's own symbol too if that's not the top left corner. Even just an empty blueprint to encourage peeps to fill it in might be better than nothing.

Yeah, I'm just printing an empty blueprint field like this : ` "blueprint": [ ]`. Thank you for reminding me to update the PR description to make it clear ^^

irwiss:
I'd consider moving the thing out of vehicle class into debug menu or wherever, as it doesn't seem to be related and it doesn't look like you need access to any internal parts of the class

Blueprints shouldn't be too complex to generate - `vehicle::get_display_of_tile` should get you the (utf32) codepoint for a mount point, `utf32_to_utf8` to get a utf8 string you can write to json and `vehicle::mount_max/mount_min` (after refresh() call) have the range of mount points to iterate

Also another thing that can make stuff more (human) readable (and is used in current vehicle defs) is splitting parts on a mount point so that formatter won't chop the array with newlines, so instead of X parts in one array you can have multiple objects on same x/y each defining chunks of the X. It's usually easier to read than the first one after formatter breaks the first array into separate lines, without doing what the formatter does you don't know exact the length, but even a rough id length sum + some leeway for the json syntax should work for most cases

SurFlurer:
Thanks for reviewing. Moving this out of vehicle class makes sense, I'll do that. 

I didn't realized that `get_display_of_tile()` returns utf32 chars. 😄 That would make printing the blueprint a lot easier.

As for splitting the parts at the same mount point, in my opinion, it's more cognitively consistent to put them in one array instead of multiple. But...let's see how far can I go in doing versatile formatting.

SurFlurer:
I decided to put the function inside vehicle_prototype class as a static member. Making it stay close to `load()` function seems reasonable to me.

SurFlurer:
Thanks for merging!

PR #66898, opened 2023-07-14
Allow owned vehicle cargo items in barter

irwiss:
#### Summary

None

#### Purpose of change

Allow owned vehicle cargo items in barter

#### Describe the solution

What it says on the tin

Also makes taking ownership of a vehicle (after theft timer expires) assign cargo items' owner - otherwise default spawned items remain owner-less until player picks up/drops them

#### Testing

Spawn a vehicle, take ownership (e.g. examine it or try to drive), talk to npc and try to barter - vehicle cargo should be available in trade screen

### DIFF
--- a/src/trade_ui.cpp
+++ b/src/trade_ui.cpp
@@ -111,9 +111,9 @@ trade_ui::trade_ui( party_t &you, npc &trader, currency_t cost, std::string titl
       _parties{ &trader, &you }, _title( std::move( title ) )
 
 {
-    map &here = get_map();
     _panes[_you]->add_character_items( you );
     _panes[_you]->add_nearby_items( 1 );
+    map &here = get_map();
     for( const wrapped_vehicle &wv : here.get_vehicles() ) {
         if( wv.v->is_owned_by( you ) ) {
             for( const tripoint pos : wv.v->get_points() ) {
--- a/src/trade_ui.cpp
+++ b/src/trade_ui.cpp
@@ -19,6 +19,7 @@
 #include "point.h"
 #include "string_formatter.h"
 #include "type_id.h"
+#include "vehicle.h"
 
 static const flag_id json_flag_NO_UNWIELD( "NO_UNWIELD" );
 static const item_category_id item_category_ITEMS_WORN( "ITEMS_WORN" );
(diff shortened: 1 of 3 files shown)


src/trade_ui.cpp
```diff
@@ -112,6 +113,16 @@ trade_ui::trade_ui( party_t &you, npc &trader, currency_t cost, std::string titl
...
+    map &here = get_map();
+    for( const wrapped_vehicle &wv : here.get_vehicles() ) {
+        if( wv.v->is_owned_by( you ) ) {
+            for( const tripoint pos : wv.v->get_points() ) {
+                if( here.inbounds( pos ) && square_dist( you.pos(), pos ) > 1 ) {
+                    _panes[_you]->add_vehicle_items( pos );
+                }
+            }
+        }
+    }
```
andrei8l, line 125: This adds the items even if the vehicle is on the other side of the map or inaccessible from your current position. I suggest limiting to `PICKUP_RANGE` (=6 tiles) and reusing `add_nearby_items()` with some new bools so you'd have either
```c++
    _panes[_you]->add_nearby_items( 1, true, false ); // map-only items
    _panes[_you]->add_nearby_items( PICKUP_RANGE, false, true ); // veh-only items
```
or just use the legacy trader logic and add everything in range
```c++
    _panes[_you]->add_nearby_items( PICKUP_RANGE );
```
(I hate this one)

src/trade_ui.cpp
```diff
@@ -112,6 +113,16 @@ trade_ui::trade_ui( party_t &you, npc &trader, currency_t cost, std::string titl
 {
     _panes[_you]->add_character_items( you );
     _panes[_you]->add_nearby_items( 1 );
+    map &here = get_map();
+    for( const wrapped_vehicle &wv : here.get_vehicles() ) {
+        if( wv.v->is_owned_by( you ) ) {
+            for( const tripoint pos : wv.v->get_points() ) {
+                if( here.inbounds( pos ) && square_dist( you.pos(), pos ) > 1 ) {
+                    _panes[_you]->add_vehicle_items( pos );
```
Qrox, line 121: If I use a vehicle scoop to pick up someone else's items, will this allow me to sell these items back to the person?

(Also, does scooping up own items count as stealing right now?)

PR #66902, opened 2023-07-14
Put Some Clothes On There Are Survivors Looking

### LINKED ISSUE #66861: Locations shouldn't be spawning naked human corpses that resurrect as naked zombies

There are several locations in the game that spawn completely naked dead "human" corpses. These almost instantly resurrect into actual zombies even upon freshly starting the game but have no clothes, because they weren't spawned with an itemgroup. Examples of this include mass grave sites, apartment complexes and looted gun stores. It doesn't make much sense that these people would be naked - maybe their backpacks were taken away and they were stripped of valuables but the clothes on their back shouldn't be gone too.

### Solution you would like.

Locations like this should just spawn living zombies instead of human corpses which resurrect almost instantly into naked zombies, which would give those zombies actual clothes instead of having nothing.

Remove the corpses entirely?

### Additional context

_No response_

Procyonae:
#### Summary
Bugfixes "Spawned corpses should now spawn with and contain their clothes"

#### Purpose of change

Fixes #66861

#### Describe the solution

I think this is every human corpse spawn in the game but I wouldn't be surprised if I missed some

For whatever reason due to corpses being partially hardcoded they can only be used as container-item outside of entries/items/groups so some of this looks messy but is just working around that.

Example working container:
```
{
  "id": "corpse_generic_human",
  "type": "item_group",
  "subtype": "distribution",
  "container-item": "corpse_generic_human",
  "entries": [ { "group": "default_zombie_death_drops" } ]
},
```

Example non working containers:

```
{
  "id": "corpse_generic_human",
  "type": "item_group",
  "subtype": "distribution",
  "entries": [ { "group": "default_zombie_death_drops", "container-item": "corpse_generic_human" } ]
}
```

```
{
  "id": "corpses",
  "type": "item_group",
  "subtype": "distribution",
  "entries": [
    { "group": "default_zombie_death_drops", "container-item": "corpse_generic_human" },
    { "group": "default_zombie_death_drops", "container-item": "corpse_generic_male" },
    ...
  ]
}
```

Yeeting some of the weirder corpses
Making a looted corpse drops item palette or something to differentiate between basic drops
Giving a chance for other categories of drops (survivor zed etc)

Making container-group able be used outside of entries/items/groups

#### Additional context

Will look into making mapgen corpses take a more random length of time to rez so they don't all stand up really fast, though more are damaged now so it's less of a problem outside of mass graves
Zeds still rez while inside bodybags not sure if there's a good way to stop that though

### DIFF
--- a/data/json/itemgroups/Locations_MapExtras/locations.json
+++ b/data/json/itemgroups/Locations_MapExtras/locations.json
@@ -2600,9 +2600,17 @@
   {
     "id": "mil_base_corpse_soldier",
     "type": "item_group",
-    "//": "Soldier corpse with clothes, without gear.",
+    "//": "Nested to damage corpse.",
     "subtype": "collection",
-    "entries": [ { "group": "mil_base_clothes_corpse_soldier", "damage": [ 1, 4 ] }, { "item": "corpse_generic_human" } ]
+    "entries": [ { "group": "mil_base_corpse_corpse_soldier", "damage": [ 2, 4 ] } ]
+  },
+  {
+    "id": "mil_base_corpse_corpse_soldier",
+    "type": "item_group",
+    "//": "Nested to damage all gear.",
+    "subtype": "collection",
+    "container-item": "corpse_generic_human",
+    "entries": [ { "group": "mil_base_clothes_corpse_soldier", "damage": [ 1, 4 ] } ]
   },
   {
     "id": "mil_base_clothes_corpse_soldier",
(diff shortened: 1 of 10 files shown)


Procyonae:
I've had a look through the source to try to improve the instant rev situation, mapgen placed corpses have their own starting age (which can seemingly even be influenced by a JSON "age" field) `std::max( calendar::turn - age, calendar::start_of_cataclysm )`
except afaik start_of_cataclysm means 60 days by default so it resolves a minimum of 60 days which means they're always ready to rez immediately, which while it makes sense isn't particularly interesting. The other place to affect it would be to change all rezzing behaviour or pass something telling the game it was manually placed and giving it different behaviour based on that.

Not really sure what to do, if anyone has suggestions as to more interesting behaviour I'm open to suggestions

Also I'm unsure what the purpose of L586 of item.cpp is but it references "corpse_generic_human" and not the other human corpses which could feasibly be causing (unknown) issues

PR #66910, opened 2023-07-14
make arbitrarily undodgeable attacks dodgeable

anoobindisguise:
#### Summary
Balance "certain previously non dodgeable attacks can be dodged"

#### Purpose of change
Certain attacks (The tail slam of a brutish amalgamation, the tail sting from arthropod mutants, and the hoof kick of ungulate mutants) were arbitrarily chosen to be completely undodgeable without the Uncanny Dodge cbm despite having no reason to be so. This feels like adversarial DMing and there's no good reason for it to be this way.

#### Describe the solution
Make them dodgeable.


RenechCDDA (approved):

Venera3:
Fair enough on the tail club, but I'm not a fan of nerfing either the horsey kick nor the sting. Also not a fan of `adversial DMing` for that matter.

RenechCDDA:
> Fair enough on the tail club, but I'm not a fan of nerfing either the horsey kick nor the sting. Also not a fan of `adversial DMing` for that matter.

I am curious on your reasoning? These attacks have no relevant preconditions (like being grabbed/held in place) that would warrant a 100% attack success rate even against uncanny dodge. ~~All of them are~~ mutant_stinger is 2 range, even, which would suggest they should probably be ~less accurate than melee attacks and not 100% accurate?

alp7292:
is grabs dodgeable i always get caught

PR #66915, opened 2023-07-15
Remove Pilot light from electric oven

Maleclypse:
#### Summary
None

#### Purpose of change
Remove pilot light from electric oven

#### Describe the solution
Remove pilot light from electric oven


RenechCDDA (approved):
OK, I saw this before a nap and I had to really really think about it after I woke up. For those wondering in the future: It's an **electric** oven. There's no need for a pilot light to light the gas. It's... electric.

Might just be me that took a while to figure it out, but hope it helps.

ashGlaw:
Makes sense to me! Good fix

Issue #65522, opened 2023-05-07
Infinite(?) Portal Storms


ManOWam:
### Describe the bug

I was just playing the game when a Portal Storm set in, which seemed normal enough for the time being, but the second the storm ended, I got the message for a new one starting, and then when the second one ended, a third began instantly, I kept going to see if it stops on it's own and got up to 10 with no signs of it slowing down

### Attach save file

[Nature-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11414446/Nature-trimmed.tar.gz)

### Steps to reproduce

1.Play the game until you get a Portal Storm
2.Wait until it ends
3.See if a new one starts immediately after?

### Expected behavior

I expected that the Portal Storm would conclude with just the one, instead of a seemingly infinite amount of them in a row, because that's what has happened every single time before

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.18363.1556 (1909)
- Game Version: 5cc5f31 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Innawood [innawood],
    Magiclysm [magiclysm],
    Blaze Industries [blazeindustries],
    DinoMod [DinoMod],
    Megafauna [megafauna],
    Bombastic Perks [bombastic_perks]
]

RenechCDDA:
I am very unsure how this has happened but it appears ps_min_woc and ps_max_woc have both been set to incredibly small values, causing the portal storm loop to run indefinitely. Here's a trimmed list of active EOCs as according to debug menu, and a trimmed list of global vars:

```global
id;timepoint;recurring
queued eocs:
EOC_PORTAL_EFFECTS_IRE;2 seconds
EOC_PORTAL_EFFECTS_ACTIVE;4 seconds
**EOC_PORTAL_STORM_WARN_OR_CAUSE_RECURRING;4 seconds**
EOC_PORTAL_EFFECTS_PASSIVE;57 seconds
theresa_remove;1 minute and 5 seconds
record_portal_storm_data;3 minutes and 47 seconds
MAKAYLA_MUTATOR_portal_storm;8 minutes and 50 seconds
[...]

PR #66934, opened 2023-07-16
Jsonify succession cooldown + fix some bad bugs!

RenechCDDA:
#### Summary
Mods "Faction succession cooldown is now a json value, and can be modded as such"

#### Purpose of change
More power to json

Fix some pretty nasty bugs I found in my recent work

#### Describe the solution
This PR is (currently) 4 commits, each commit should build properly and function individually as advertised.

6436583a2d3326c4dad71e4b78e853e7792d5a9d: Move the succession cooldown to json. Very importantly it must not be read as an integer value (std::stoi) or else we wind up with a regression of #65522

7db200d1b58aea1eac631905cc48844905038554: Extends the fix for reading very large numbers to the existing dialog var `timer_time_of_last_succession`

26c370854b9218057d44d3fdbed648503ef35d43: Makes EOC_FACTION_SUCCESSION_CHANGE_INIT only fire once, on the initial game start. Previously if you saved the game and reloaded it would fire again, setting your cooldown to 90 days even if you had been off cooldown!

7e0a890f9b32d446dd4e79e83154c5b6bdd40754: Just cleans up EOC_FACTION_SUCCESSION_IMMEDIATE to make it easier to mod or manipulate the cooldown otherwise (and also looks much better with the sexy math functions)

#### Testing
Reloaded and started new games under a variety of conditions to make sure each of these commits works

Recommend testing: Start new game, confirm global var for cooldown timer is set. Forward time 10 days to lower remaining cooldown to 80 days. Save, quit, load to check again and confirm remaining cooldown is still 80 days and not reset to 90. Then break character's limb and confirm cooldown is instantly refreshed.

### DIFF
--- a/data/json/npcs/missiondef.json
+++ b/data/json/npcs/missiondef.json
@@ -976,6 +976,15 @@
   },
   {
     "type": "effect_on_condition",
+    "id": "EOC_FACTION_SUCCESSION_COOLDOWN",
+    "eoc_type": "EVENT",
+    "required_event": "game_begin",
+    "//": "If you're looking to change the cooldown simply replace the below value with how long the cooldown should be, in seconds.",
+    "//2": "7,776,000 seconds = 90 days",
+    "effect": [ { "math": [ "time_between_succession", "=", "7776000" ] } ]
+  },
+  {
+    "type": "effect_on_condition",
     "id": "EOC_FACTION_SUCCESSION_IMMEDIATE",
     "eoc_type": "EVENT",
     "required_event": "broken_bone",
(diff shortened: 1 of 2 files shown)


data/json/npcs/missiondef.json
```diff
@@ -974,21 +974,30 @@
...
     ]
   },
+  {
+    "type": "effect_on_condition",
+    "id": "EOC_FACTION_SUCCESSION_COOLDOWN",
+    "eoc_type": "EVENT",
+    "required_event": "game_begin",
+    "//": "If you're looking to change the cooldown simply replace the below value with how long the cooldown should be, in seconds.",
+    "//2": "7,776,000 seconds = 90 days",
+    "effect": [ { "math": [ "time_between_succession", "=", "7776000" ] } ]
```
andrei8l, line 984:
```suggestion
    "effect": [ { "math": [ "time_between_succession", "=", "u_val('time: 90 d')" ] } ]
```

RenechCDDA: Ah that's neat, thank you.

PR #66940, opened 2023-07-16
lockpicking proficiency prevents damaging lockpick

LukeBaden:
#### Summary
Features "lockpicking proficiency prevents damaging lockpick"

#### Purpose of change
In real life a complete novice who understands the basics of lockpicking rarely damages their picks. Even picks made out of bobby pins are pretty hard to break.

#### Describe the solution
I made the lockpicking action check if the player has the lockpicking proficiency and if so disables the chance of damaging the pick, defaulting to the basic failure.

I thought about lowering the chance in general of breaking lockpicks but, someone who has no understanding of the mechanics of lockpicking could easily break lockpicks by jamming them into locks randomly.

#### Testing
 a basic survivor with no experience lockpicking and 0 devices skill will break multiple rudimentary/improvised lockpicks before becoming proficent and a survivor with proficiency will not break their lockpick.

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -2546,7 +2546,7 @@ void lockpick_activity_actor::finish( player_activity &act, Character &who )
     } else if( furn_type == f_gunsafe_ml && lock_roll > ( 3 * pick_roll ) ) {
         who.add_msg_if_player( m_bad, _( "Your clumsy attempt jams the lock!" ) );
         here.furn_set( target, furn_f_gunsafe_mj );
-    } else if( lock_roll > ( 1.5 * pick_roll ) ) {
+    } else if( lock_roll > ( 1.5 * pick_roll ) && !who.has_proficiency( proficiency_prof_lockpicking )  ) {
         if( it->inc_damage() ) {
             who.add_msg_if_player( m_bad,
                                    _( "The lock stumps your efforts to pick it, and you destroy your tool." ) );


src/activity_actor.cpp
```diff
@@ -2546,7 +2546,7 @@ void lockpick_activity_actor::finish( player_activity &act, Character &who )
     } else if( furn_type == f_gunsafe_ml && lock_roll > ( 3 * pick_roll ) ) {
         who.add_msg_if_player( m_bad, _( "Your clumsy attempt jams the lock!" ) );
         here.furn_set( target, furn_f_gunsafe_mj );
-    } else if( lock_roll > ( 1.5 * pick_roll ) ) {
+    } else if( lock_roll > ( 1.5 * pick_roll ) && !who.has_proficiency( proficiency_prof_lockpicking )  ) {
```
RenechCDDA, line 2549: So, this makes it completely impossible to damage your tools (no matter how difficult a lock) if you have proficiency_prof_lockpicking, not even prof_lockpicking_expert.

Given that the chance to damage is already skill-dependent, I think this needs to be *at least* checking for the lockpicking expert proficiency to say there's no chance to damage tools, ever.

LukeBaden: hmmm, you're right. 
but a survivor with lockpicking expert probably has a high enough devices skill not to run into this problem. so it might be a better idea to come up with an overhaul/upgrade to the lockpicking system instead of a balance change. I'll workshop some ideas on the forum.

PR #66941, opened 2023-07-16
Roundabouts

### LINKED ISSUE #66759: Nest placement to be able to check for overmap terrain flags such as SIDEWALK

Procyonae:
#### Summary
Content "Adds small roundabouts"

#### Purpose of change

Roundabouts exist
I made one while trying to fix #66759

#### Describe the solution

Unless I hit a roadblock I'm closing this and just putting it in my road unhardcoding PR

Adds rare small roundabouts that spawn as specials over 3way and 4way intersections, correctly spawning SIDEWALK where appropriate
If they spawn surrounded by SIDEWALK they spawn stuff in the middle like flowers, a statue or a "fountain"

I'll remove the specials once I unhardcode the road maps and just make them an alternate 3way and 4way intersection

I tried spawning larger roundabout where you get 4 adjacent road from connections messing up out of cities but for whatever reason they wouldn't spawn

I appreciate these are too small realistically but I think they're better than not having any roundabouts

I might look at allowing signage on things other than f_sign so for example the statue could have writing on it (like a plaque), just assigning the intended flag "SIGN" doesn't work currently

#### Testing

Sometimes `"existing": true` doesn't work correctly (see screenshots below) but this is unrelated to the map itself afaik
Out of city spawns are very rare but do occur, I don't fully understand how placement nor placement order works but I played with trying to manipulate the order and that didn't change anything

#### Additional context
<details>

image
3way in a city

image
4way in a city

image
Another 4way in a city

image
Example of `"existing": true` connections being bugged

</details>


PatrikLundell:
I can't say I'm a fan of making it even harder to drive around...

Procyonae:
> I can't say I'm a fan of making it even harder to drive around...

How does this make it harder to drive?

PatrikLundell:
Turning vehicles is always an issue, and with the center of the roundabout blocked by "stuff" you'll have to drive in a hemi circle, risking to hit broken vehicles and chased down by zombies as you can't maintain a reasonable speed. Towing would probably demand that the center was cleared of "stuff" to be successful.

Procyonae:
> Turning vehicles is always an issue, and with the center of the roundabout blocked by "stuff" you'll have to drive in a hemi circle, risking to hit broken vehicles and chased down by zombies as you can't maintain a reasonable speed. Towing would probably demand that the center was cleared of "stuff" to be successful.

I mean they spawn rarely, only get something in the centre if there's SIDEWALK at all 4 corners and even then only 2/5 spawn with stuff that's going to damage your vehicle. Admittedly the fountain is kinda awkward but the statue is only 2 tiles (it should be one when I figure out how to extend "signage" functionality).

bombasticSlacks:
See your trying to get a shoutout from https://en.wikipedia.org/wiki/Roundabout_Appreciation_Society

Procyonae:
> See your trying to get a shoutout from https://en.wikipedia.org/wiki/Roundabout_Appreciation_Society

I didn't know that was a thing, I'll have to find out how to become a member :^)

Ochrad:
Great addition. Though, I was sure that in game driving side is on the right-hand side of the road, since cars have steering wheel on the left. Therefore, the stop lines supposed to be flipped on the other lane to create a counter-clock wise motion. (have to drive through a roundabout every day IRL..)

Procyonae:
> Great addition. Though, I was sure that in game driving side is on the right-hand side of the road, since cars have steering wheel on the left. Therefore, the stop lines supposed to be flipped on the other lane to create a counter-clock wise motion. (have to drive through a roundabout every day IRL..)

Ye I realised I messed that up while doing the road maps, I'm not sure I'm going to bother merging this bc I'm making good progress with that so just waiting and adding it as a crossroad alternative probably makes more sense

Procyonae:
Closing bc road unhardcoding is on the way soon™

PR #66955, opened 2023-07-16
Pstorm distance fix

### LINKED ISSUE #66598: Portal Storms Stop Occurring after 1st Storm
### Describe the bug

After the first portal storm a player experiences, no more portal storms will correctly spawn. The player receives the 'reality is breaking' message, but no actual storm effects such as monster spawning or even the portal storm portal overlay appears.  Eventually the reality is healed message appears, but this does nothing as no portal storm was actually there.  Furthermore, while the center of the portal storm mission does spawn, moving to the center does nothing, with no text dialogue prompt that there is supposed to be.

### Attach save file

Save file is in the middle of a false portal storm.  Mission to find center is active, 'reality is breaking message' appears, but there is no actual portal storm.

[Demopolis.zip](https://github.com/CleverRaven/Cataclysm-DDA/files/11928724/Demopolis.zip)

### Steps to reproduce

1. Use debug menu to set time to 1 year to guarantee first portal storm / get past the no portal storm time period.
2. Wait for first portal storm to finish.
3. Use debug menu to pass more time / manually wait for a second portal storm.
4. See that despite a second portal storm 'beginning' nothing actually happens.

### Expected behavior

Portal storm should begin, portal storm center should give the dialogue option, etc...

### Screenshots

_No response_

- OS: Windows
- Game Version: b7db497 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

While I used debug commands to efficiently debug the issue, I have frequently encountered this issue in normal gameplay without debug commands to set time.

Zapafaz:
#### Summary
Bugfixes "Add an EOC for default portal storm distance values"

#### Purpose of change
Portal storms stop changing weather outside of a very small area (starts at 1 square) after EOC_PORTAL_EXPAND runs at the end of a storm for the first time. This adds an EOC to fix that issue. Fixes #66598

#### Describe the solution
Added an EOC to EOC_PORTAL_STORM_WARN_OR_CAUSE_RECURRING that checks if any of portal_storm_distant_distance, portal_storm_close_distance, and portal_storm_distance are lower than or equal to the default values (200, 100, 50 respectively) in weather_type.json. If they are, sets them to those values.

Check and set these default values elsewhere. I am not very familiar with the codebase so I didn't do this.

In addition to that, include some kind of check for the current value of portal_storm_distance etc. for already existing games that have those values at e.g. 5 or more. Did not do this since it'd only be relevant for in-progress games that have had many portal storms already and would add clutter that has no effect on new games.

#### Testing
Created a new world, started a new character, debug ran EOC_PORTAL_STORM_WARN_OR_CAUSE_RECURRING. Debug checked var_list.output for expected portal_storm_distance etc. values. Debug ran EOC_PORTAL_EXPAND. Checked var_list for expected values again. Debug ran EOC_PORTAL_STORM_WARN_OR_CAUSE_RECURRING and checked var_list for expected values again. All values and functionality is as expected.

#### Additional context
This should work for new games and for existing games since the EOC checks if it should run every time EOC_PORTAL_STORM_WARN_OR_CAUSE_RECURRING runs, which is what kicks off a portal storm every time. Scenarios that start with a portal storm via other EOCs use the default values already and should have the appropriate defaults set the next time a portal storm starts.

### DIFF
--- a/data/json/portal_storm_effect_on_condition.json
+++ b/data/json/portal_storm_effect_on_condition.json
@@ -26,7 +26,28 @@
     "id": "EOC_PORTAL_STORM_WARN_OR_CAUSE_RECURRING",
     "recurrence": [ { "global_val": "ps_min_woc", "default": "5 days" }, { "global_val": "ps_max_woc", "default": "18 days" } ],
     "global": true,
-    "effect": { "run_eocs": "EOC_CAUSE_EARLY_PORTAL_STORM" }
+    "effect": {
+      "run_eocs": [
+        {
+          "id": "EOC_CAUSE_EARLY_PORTAL_STORM"
+        },
+        {
+          "id": "EOC_FIRST_PORTAL_STORM",
+          "condition": {
+            "or": [
+              { "math": [ "portal_storm_distant_distance", "<=", "200" ] },
+              { "math": [ "portal_storm_close_distance", "<=", "100" ] },
+              { "math": [ "portal_storm_distance", "<=", "50" ] }
+            ]
+          },
+          "effect": [
+            { "math": [ "portal_storm_distant_distance", "=", "200" ] },
+            { "math": [ "portal_storm_close_distance", "=", "100" ] },
+            { "math": [ "portal_storm_distance", "=", "50" ] }
+          ]
+        }
+      ]
+    } 
   },
   {
     "type": "effect_on_condition",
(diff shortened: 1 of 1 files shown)


Zapafaz:
@Ramza13 beat me to it with #66953

PR #66956, opened 2023-07-16
Make all graded weapons use copy-from [Part 1 Swords and Blades]

Procyonae:
#### Summary
None

#### Purpose of change

First part of #67189

#### Describe the solution

See the linked issue

This includes my first item migration so if someone could specifically check that's correct it'd be appreciated ^-^

Make craftable makeshifts use copy-from but that doesn't seem appropriate when they're only meant to vaguely represent said weapon in terms of technique not build

#### Additional context

Any comments relating to the whole change should be posted on the linked issue


TheShadowFerret:
Time to use `"replace_materials"` from #66757 👀?

Procyonae:
> Time to use also replace their materials field with `"replace_materials"` from #66757 👀?

I can if that's preferred

PR #66976, opened 2023-07-17
Add recipe to craft 2 sticks out of a long stick

ghost:
#### Summary
Content "Add recipe to craft 2 sticks out of a long stick"
as its says on the tin

#### Purpose of change
at 5 strength it is very hard to snap a stick in half, below that and its practically impossible, this fixes that by using a cutting tool, a few minutes of your time.


data/json/recipes/other/parts.json
```diff
@@ -926,5 +926,19 @@
     "proficiencies": [ { "proficiency": "prof_knapping", "time_multiplier": 1.25, "skill_penalty": 0.1 } ],
     "qualities": [ { "id": "HAMMER", "level": 1 } ],
     "components": [ [ [ "rock_flaking", 1 ] ] ]
+  },
+  {
+    "type": "recipe",
+    "activity_level": "LIGHT_EXERCISE",
+    "id_suffix": "innawood",
+    "result": "stick",
```
Procyonae, line 934:
```suggestion
    "activity_level": "LIGHT_EXERCISE",
    "result": "stick",
```

RenechCDDA:
That's true, no chance to turn something into splinters while cutting it hope (I hope?).

data/mods/innawood/recipes/other/parts.json
```diff
@@ -262,4 +262,4 @@
     "qualities": [ { "id": "SAW_M", "level": 1 } ],
     "components": [ [ [ "steel_chunk", 1 ], [ "scrap", 3 ], [ "knife_butter", 2 ] ] ]
   }
-]
+]
```
Procyonae, line 265:
```suggestion
]

```
File ends need a newline the bot's not broke

Light-Wave:
This can already be done by deconstructing long strings. I am not against having a crafting recipe that does the exact same thing, but it should match the deconstruction recipe found [here.](https://github.com/CleverRaven/Cataclysm-DDA/blob/237374d4ec0196725844a6f430b02bccc96368c3/data/json/recipes/recipe_deconstruction.json#L4088-L4093)
More specifically, the time taken should be 1 minute. instead of 3.

RenechCDDA (approved):
Equivalent to existing deconstruction recipe

Procyonae:
> This can already be done by deconstructing long strings. I am not against having a crafting recipe that does the exact same thing, but it should match the deconstruction recipe found [here.](https://github.com/CleverRaven/Cataclysm-DDA/blob/237374d4ec0196725844a6f430b02bccc96368c3/data/json/recipes/recipe_deconstruction.json#L4088-L4093) More specifically, the time taken should be 1 minute. instead of 3.

I actually disagree that disassembly existing is a good reason not to have a crafting recipe that does the exact same, why should this recipe specifically have a duplicate? In general this defo shouldn't be allowed bc it's just UI bloat

Light-Wave:
> I actually disagree that disassembly existing is a good reason not to have a crafting recipe that does the exact same, why should this recipe specifically have a duplicate? In general this defo shouldn't be allowed bc it's just UI bloat

I did not intend to make any judgement on whether or not this change should be merged, only that _if_ it should get merged, then the recipes should be consistent.
I can see some value with merging this, and I can see some value with not doing so. I have no strong feelings one way or another.

PR #66985, opened 2023-07-18
Fix message in ter_furn_data don't be translated

lispcoc:
#### Summary
Bugfixes "Fix message in ter_furn_data don't be translated"

#### Purpose of change
Message in ter_furn_data is not translated, in spite of translation exists.

#### Describe the solution
convert type of message from std::string to translation.

#### Testing
Game runs, message is translated actually.

### DIFF
--- a/src/magic_ter_furn_transform.h
+++ b/src/magic_ter_furn_transform.h
@@ -21,10 +21,10 @@ class ter_furn_data
 {
     public:
         weighted_int_list<T> list;
-        std::optional<translation> message;
+        translation message;
         bool message_good = false;
         ter_furn_data() = default;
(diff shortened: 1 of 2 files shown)


src/magic_ter_furn_transform.h
```diff
@@ -21,10 +21,10 @@ class ter_furn_data
 {
     public:
         weighted_int_list<T> list;
-        std::string message;
+        std::optional<translation> message;
```
Qrox, line 24: `translation` already has a null state which can be tested by `empty()`, so `std::optional` is not needed here.

Qrox (approved):
Thanks!

PR #66990, opened 2023-07-18
Refactor test for npcs boarding player vehicles

### LINKED ISSUE #66968: npc_test frequent test fail

npc_test is failing the test case npc-board-player-vehicle very frequently with ```(~[slow] ~[.],starting_items)=> ../tests/npc_test.cpp:386: FAILED:
(~[slow] ~[.],starting_items)=>   REQUIRE( companion->pos() == player_pos + point( 2, 0 ) )
Error: (~[slow] ~[.],starting_items)=>   (61,62,0) == (62,60,0)``` 

### Solution you would like.

Tests not to fail all the time
~~As far as I can tell this is happening bc the NPC decides to get in the back reclining seat of the ambulance rather than the passenger seat because it's a shorter distance. I have no idea whether the test intentionally only wants them to go for the passenger seat or not.~~
No idea what the problem is see the PR below for a failed fix

~~Just allowing the other seat in the test.~~
Removing the test until someone finds a proper solution

### Additional context

_No response_

mqrause:
#### Summary
None

#### Purpose of change

Fixes #66968

#### Describe the solution

- jsonzie the test for fun and easier possible future expansion (like vehicles with z-levels)
- replace the ambulance with a custom test vehicle
- dirty pathfinding cache when placing or removing vehicles
- remove comments that aren't really necessary or helpful

Only replacing the vehicle, dirtying the cache and adjusting the points accordingly.

#### Testing

Ran `./Cataclysm-test-vcpkg-static-Release-x64.exe activity_interruption_by_distractions,npc-board-player-vehicle --order lex` and it passes. Without dirtying the cache it fails. Thanks a ton to @andrei8l for identifying this.

### DIFF
--- a/tests/npc_test.cpp
+++ b/tests/npc_test.cpp
@@ -24,9 +24,11 @@
 #include "npc_class.h"
 #include "npctalk.h"
 #include "overmapbuffer.h"
+#include "pathfinding.h"
 #include "pimpl.h"
 #include "player_helpers.h"
 #include "point.h"
+#include "test_data.h"
 #include "text_snippets.h"
 #include "type_id.h"
 #include "units.h"
@@ -44,7 +46,6 @@ static const trait_id trait_WEB_WEAVER( "WEB_WEAVER" );
 static const vpart_id vpart_frame( "frame" );
 static const vpart_id vpart_seat( "seat" );
 
-static const vproto_id vehicle_prototype_ambulance( "ambulance" );
 static const vproto_id vehicle_prototype_none( "none" );
 
 static void on_load_test( npc &who, const time_duration &from, const time_duration &to )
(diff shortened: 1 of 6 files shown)


Maleclypse:
I'm intrigued that this one got farther than other ones seem to be getting.  Is it a random failure that is extremely common?

mqrause:
I'm still not sure about randomness. It consistently passed basic build even before swapping the vehicle and with the swap it consistently passes clang 14. And at least in this PR it consistently failed GCC 9. I'm still in the process of figuring out helpful output to see why it actually fails.

mqrause:
So from what I can see, the GCC 9 failure comes from some issue with pathfinding. The vehicle tile the npc tries to path through seems to be correctly processed, but is still picked to path through for some reason. Since I lack the capability to compile with GCC 9, I can't reproduce the issue myself and thus also can't fix it.

andrei8l:
Try this
```diff
diff --git a/tests/npc_test.cpp b/tests/npc_test.cpp
index 8bd7246bb3..7d57fb1be8 100644
--- a/tests/npc_test.cpp
+++ b/tests/npc_test.cpp
@@ -347,6 +347,7 @@ TEST_CASE( "npc-board-player-vehicle" )
 
             vehicle *veh = here.add_vehicle( data.veh_prototype, data.player_pos, 0_degrees, 100, 2 );
             REQUIRE( veh != nullptr );
+            here.set_pathfinding_cache_dirty( data.player_pos.z );
             here.board_vehicle( data.player_pos, &pc );
             REQUIRE( pc.in_vehicle );
 
```
It's pathing through the windshield because [it tries the trivial straight line first](https://github.com/CleverRaven/Cataclysm-DDA/blob/1ba9bc838ddaba2fe3941e3c9d4337da24d996c2/src/pathfinding.cpp#L187-L203) and placing vehicles doesn't mark the pathfinding cache as dirty. It probably should, but I'll leave that up to you.

tests/npc_test.cpp
```diff
@@ -315,75 +316,105 @@ static void check_npc_movement( const tripoint &origin )
...
     g->load_npcs();
     guy->companion_mission_role_id.clear();
-    // Ensure they aren't trying to guard anything
     guy->guard_pos = std::nullopt;
-    // Reset the NPC, so they don't need to eat, or want to play with their items
     clear_character( *guy );
-    // But set them back to the desired position, because clear_character also sets position
     guy->setpos( npc_pos );
-
-    // Make the NPC a follower of the player
```
ehughsbaird, line 354: I don't think these comments are without value. They're saying why these things are being done because they're specific problems that came up in testing, and I think are thus important things to understand when writing a test using this.
In particularly, I think the setpos comment is important (if I'm reading this code and seeing that we're spawning at a precise location, then setting the position back to that again, I have a question. The comment answers that without spending my time).

tests/npc_test.cpp
```diff
@@ -315,75 +316,105 @@ static void check_npc_movement( const tripoint &origin )
...
+
+    for( std::pair<const std::string, npc_boarding_test_data> &given : test_data::npc_boarding_data ) {
+        GIVEN( given.first ) {
+            npc_boarding_test_data &data = given.second;
+            g->place_player( data.player_pos );
+            clear_map();
+            map &here = get_map();
+            Character &pc = get_player_character();
+
+            vehicle *veh = here.add_vehicle( data.veh_prototype, data.player_pos, 0_degrees, 100, 2 );
```
ehughsbaird, line 348: The assumption that we want the player to be at the center point of the vehicle should be documented. This is true for the test vehicles added, but not all vehicles are constructed like this.

ehughsbaird (commented):
Thank you for the cleanup and fixes! Sorry that I didn't catch this.

mqrause:
> It's pathing through the windshield because [it tries the trivial straight line first](https://github.com/CleverRaven/Cataclysm-DDA/blob/1ba9bc838ddaba2fe3941e3c9d4337da24d996c2/src/pathfinding.cpp#L187-L203) and placing vehicles doesn't mark the pathfinding cache as dirty. It probably should, but I'll leave that up to you.

Huh, but why is that only an issue with GCC? When does the cache even get built? Clearing it when placing a vehicle does sound sensible, especially if something like unfolding a folded vehicle could cause an issue like that. But if it's really only a test issue, clearing the cache in `clear_map` might be enough?

tests/npc_test.cpp
mqrause: The test is meant to use custom vehicles tailored to the test case. If we really want a vehicle with a different center than the player position, adding a separate vehicle spawn point is fairly trivial.

andrei8l:
> But if it's really only a test issue, clearing the cache in `clear_map` might be enough?

`clear_map()` normally does set the dirty flag through `map::ter_set()` and `map::furn_set()`, unless the map has already been cleared. This is most likely a test order issue as the test doesn't fail if ran on its own, even in LTO builds, and ASan+UBsan doesn't find anything.

tests/npc_test.cpp
mqrause: Okay, for one I'm mostly not a fan of comments in code, so there should be a minimal amount of that. That said especially the `setpos` is pretty unintuitive, so some explanation might be worthwhile. But the reason I extracted that to it's own method is so people using it don't have to think about it and just get an NPC companion at the point passed to it.

PR #66996, opened 2023-07-18
Bionics can require another bionic

anoobindisguise:
#### Summary
Content "bionics can require another bionic"

#### Purpose of change
Some bionics didn't make sense to be installed without other bionics present. I wanted to change that, but the existing "upgrades" structure for bionics is for replacing bionics with new versions, rather than installing them side by side.
This will probably help us do "framework" bionics where you can install certain bionics like a bionic arm, and then use that as a framework for for example the bionic monoblade or shotgun.

#### Describe the solution
You can now specify exactly one bionic that is needed to be installed in order to install a specific bionic.
As an example of the system, I made Hydraulic Muscles require Titanium Skeletal Bracing, as I think this one would be pretty uncontroversial (you're so strong under hydraulic muscles you would probably break your own bones)

I've seen some concern that such a system might not be beneficial to the game - bionics are designed to be "plug and play" and making too many bionics require other bionics will reduce build variety by forcing everyone to get specific "base" bionics in order to get the ones they want. However I don't really agree with that and think that as long as this system isn't used excessively bionics can progress in a "tree" which still offers plenty of variety. I also hope that this system will allow the addition of new bionics which we couldn't have before because there was no way to ensure you had a secondary bionic which would be necessary to make it function.

I debated allowing multiple requirements but I think this is better, as it reads more nicely than a list of different bionics, and is easier for players, and won't shoehorn too many bionics into chaining together and reducing build variety.

#### Testing
Compiled, tested, confirmed functional.

#### Additional context
The bionic that has a missing prereq shows up as gray in the inventory rather than the usual green. You can also see in the item description which bionic is needed as a prereq. The bionic install menu shows the reason it can't be installed. After installing the prereq, it shows as being green in inventory because its prereq is met, and it can be installed. Also, the required cbm cannot be uninstalled while the cbm is installed, but once it is removed it can also be removed again.

reqbionic1
reqbionic5
reqbionic2
reqbionic3
reqbionic4

### DIFF
--- a/data/json/bionics.json
+++ b/data/json/bionics.json
@@ -1097,7 +1097,8 @@
     "flags": [ "BIONIC_TOGGLED", "BIONIC_NPC_USABLE" ],
     "act_cost": "10 kJ",
     "react_cost": "10 kJ",
-    "time": "1 s"
+    "time": "1 s",
+    "required_bionic": "bio_weight"
   },
   {
     "id": "bio_infrared",
--- a/data/json/bionics.json
+++ b/data/json/bionics.json
@@ -1097,8 +1097,7 @@
     "flags": [ "BIONIC_TOGGLED", "BIONIC_NPC_USABLE" ],
     "act_cost": "10 kJ",
     "react_cost": "10 kJ",
-    "time": "1 s",
-    "required_bionic": "bio_weight"
+    "time": "1 s"
   },
   {
     "id": "bio_infrared",
--- a/data/json/npcs/exodii/exodii_merchant_itemlist.json
+++ b/data/json/npcs/exodii/exodii_merchant_itemlist.json
@@ -197,7 +197,10 @@
       [ "bio_evap", 10 ],
       [ "bio_face_mask", 10 ],
       [ "bio_gills", 10 ],
-      { "collection": [ { "item": "bio_hydraulics", "prob": 100 }, { "item": "bio_weight", "prob": 100 } ], "prob": 10 },
+      {
+        "collection": [ { "item": "bio_hydraulics", "prob": 100 }, { "item": "bio_weight", "prob": 100 } ],
+        "prob": 10
+      },
       [ "bio_membrane", 10 ],
       [ "bio_painkiller", 10 ],
       [ "bio_radscrubber", 10 ],
--- a/data/json/npcs/exodii/exodii_merchant_itemlist.json
+++ b/data/json/npcs/exodii/exodii_merchant_itemlist.json
@@ -197,7 +197,7 @@
       [ "bio_evap", 10 ],
       [ "bio_face_mask", 10 ],
       [ "bio_gills", 10 ],
-      [ "bio_hydraulics", 10 ],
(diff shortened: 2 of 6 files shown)


fairyarmadillo:
This is a bad idea and moreover a bad precedent. We are always going to have bionics which are more desirable than others and hooking them into a system of prereqs means that there is going to be a substantial set of bionics which are seen as necessary just because the good stuff is gated behind them.

This is the kind of thing that creates cookie cutter builds. The beauty of the existing mutation and bionic systems are that you have a lot of freedom to come up with intetesting builds, but there's been a creeping trend to curb that freedom. While some of it makes sense, like joint servos being incompatible with octopus tentacles, changes like this one seem to be monkeying with the system just for the sake of doing it. The result is all cost and no benefit.

The hydraulic muscles CBM was not envisioned as something that would require a special skeleton to work. That's an invented justification strictly for the sake of this PR. While it's a minor enough change, it creates a precedent that will prompt further drift in this direction as people's natural tendency to want to sort things into subcategories kicks in.

Procyonae:
As a single level do you have X I see this as good but turning this into "trees" by nesting this behaviour in the future as you've suggested sounds awful, even more so while CBMs are limited primarily to Rubik

anoobindisguise:
Rubik being the primary source of CBMs is definitely a problem for this PR. Quite frankly they suck and getting a set of bionics you want requires several weeks of going back and forth between their shop even if you savescum heavily to get them to sell you the bionics you want (which by default, they usually won't). I can see how needing Bionic X to install Bionic Y would be really goddamn annoying if the primary source of CBMs spends several months deciding not to sell you Bionic X due to bad RNG.

anoobindisguise:
I've gone ahead and removed the requirement of hydraulic muscles needing titanium skeletal bracing for now. The question of whether that should happen can wait until later, for now I know the framework is functional but I want to just introduce it alone without any other changes.

Maleclypse:
> > Is that really only a "right now" problem though? I don't think tgere are plans to make CBMs commonly available. There might be more zomborg locations added as they get migrated out of labs, but that's not really going to substantially change the rate the player comes across them is it?
> 
> I asked about adding zomborg locations and lore peeps said that wasn't even wanted 🤷

It's not wanted because Zomborgs aren't the desired state for advanced cbm drops.  What are desired are specific high difficulty missions from Rubik with the ability to choose one of two advanced CBMs from as a reward and Labyrinthine structures which would be mutable mapgen points that are basically dungeons into Nether locations that have weird/advanced cbms in them.

Procyonae:
> It's not wanted because Zomborgs aren't the desired state for advanced cbm drops. What are desired are specific high difficulty missions from Rubik with the ability to choose one of two advanced CBMs from as a reward and Labyrinthine structures which would be mutable mapgen points that are basically dungeons into Nether locations that have weird/advanced cbms in them.

I'm kind of into the idea of just unhardcoding and improving as much existing mapgen as possible rn but I'd happily make a map like that at some point if someone else comes up with suitable lore with an outline of the quest structure

Issue #66947, opened 2023-07-16
Submachine guns need a range bonus review.


randomtyper:
### Is your feature request related to a problem? Please describe.

In game, all MP5 series guns have a bonus range of 14. The MP5A2 (225 mm barrel), MP5K-PDW (140 mm barrel length, some pistols have longer ones) and MP5SD (a proprietary silencer that slows bullet to subsonic speed), despite all of them have differing barrel lengths and constructions, all of those increase range of a 9x19 mm FMJ bullet by factor of 2 from 14 to 28. This translates to ~330 m effective range if going by tiles*12 formula, despite IRL the effective range of MP5A2 is stated to be about 150-200 m. The effective range for MP5K is even lower at 100 m.

Sources:
https://weaponsystems.net/system/227-Heckler+&+Koch+MP5
https://weaponsystems.net/system/229-Heckler%20&%20Koch%20MP5-K
https://www.militarytoday.com/firearms/hk_mp5.htm

At the same time, P90, which has the same stated effective range of 200 m (with caveats: it defeats Level 3 body armor at 200 m while having "maximum effective range" of 400 m, at which bullet delivers 85 J of energy to target) as full-sized MP5A2 (200 m).
Despite that, the ingame P90, which has a barrel 264 mm long has the same bonus range (0 tiles) as a FN 5-7 with barrel length of 122 mm.

Quoting https://www.remtek.com/arms/fn/p90/data/concept.htm:
[...]

PR #66999, opened 2023-07-18
adjust 9mm gun range bonus

anoobindisguise:
#### Summary
Balance "9mm smgs and carbines have somewhat less range"

#### Purpose of change
Partially resolve issues in #66947
9mm PCC ranges were probably too large to begin with as we cannot directly translate effective range into tiles range in game.

#### Describe the solution
Lower bonuses by 33%-50% or so and also bugfixes the MP5 PDW and MP5K having the same range as the MP5A2 when they weren't supposed to.

Do other calibers too. 
Redo the way range works period?


randomtyper:
Should have asked ya to not try to fix things yourself as I don't like to roll over recent changes, but whatever. Thanks for the effort anyways 👍

PR #67003, opened 2023-07-18
Add a spell-using feral human to Magiclysm

MNG-cataclysm:
#### Summary
Content "Add a spell-using feral human to Magiclysm."

#### Purpose of change
Add a new feral enemy which can use spells.

#### Describe the solution
The feral magician is an extremely rare enemy which spawns with other ferals, and is capable of using Blinding Flash, Ethereal Grasp, and Point Flare, casting with the same frequency as the lizardfolk shaman. In addition to the usual loot of ferals, you can sometimes find minor magical items or wands on it when killed.

Not doing this.

#### Testing
Spawned one in and fought it. However, for some reason, neither it nor the lizardfold shaman would cast any of their magic, just rushing straight to melee. I believe this might be caused by something outside of the PR which I don't know of.

#### Additional context
None at the moment.

### DIFF
--- a/data/mods/Magiclysm/Spells/monsterspells.json
+++ b/data/mods/Magiclysm/Spells/monsterspells.json
@@ -236,10 +236,6 @@
     "min_damage": 1,
     "max_damage": 1,
     "message": "",
-    "extra_effects": [
-      { "id": "blinding_flash" },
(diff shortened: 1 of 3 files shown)


Maleclypse:
Love it!

PR #67004, opened 2023-07-18
change chewing gum to not be a stimulant

Zapafaz:
#### Summary
Bugfixes "Chewing gum is (not) a stimulant"

#### Purpose of change
Chewing gum counted as a stimulant; given the description this was unexpected. Now it's not a stimulant.

#### Describe the solution
Removed `"stim": 1,`  from chewing gum in med.json.

#### Testing
Loaded game started before the fix, consumed chewing gum, no stimulant effect. Made new character, spawned and consumed chewing gum, no stimulant effect.

### DIFF
--- a/data/json/items/comestibles/med.json
+++ b/data/json/items/comestibles/med.json
@@ -693,7 +693,6 @@
     "stack_size": 100,
     "symbol": "*",
     "color": "pink",
-    "stim": 1,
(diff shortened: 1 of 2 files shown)


Qrox:
Might want to also change its comestible type to `FOOD` and move the definition to another file such as `other.json`.

PR #67007, opened 2023-07-18
make mattress traps look like furniture

anoobindisguise:
#### Summary
Bugfixes "mattress traps didn't have a tile"

#### Purpose of change
Mattresses were changed to traps, which means they had new IDs and no sprite. Truly the crimes against comfy know no bounds.

#### Describe the solution
Give them looks_likes.

This won't work 100% because traps can't be dragged and don't have a facing value like furniture does. I think this means you can also deploy them on toilets and stuff too? And you can't drag them anymore? Should they really be traps?

### DIFF
--- a/data/json/traps.json
+++ b/data/json/traps.json
@@ -85,7 +85,8 @@
     "floor_bedding_warmth": 800,
     "action": "none",
     "drops": [ "mattress" ],
-    "benign": true
+    "benign": true,
+    "looks_like": "f_floor_mattress"
   },
   {
     "type": "trap",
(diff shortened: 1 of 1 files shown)


anoobindisguise:
wait traps can't have looks_like? WHY

ashGlaw:
> wait traps can't have looks_like? WHY

That's why I didn't add those. I didn't replace anything, you can still set up mattresses his furniture. I just wanted to be able to deploy them inside tents. It's a kludge, but that's the game

anoobindisguise:
> > wait traps can't have looks_like? WHY
> 
> That's why I didn't add those. I didn't replace anything, you can still set up mattresses up as furniture. I just wanted to be able to deploy them inside tents. It's a kludge, but that's the game

How do you do that? I can't seem to place them anymore, activating them just makes a trap and there's no construction menu for them.

ashGlaw:
Hmm. No, you're right. It's out. My bad

PR #67008, opened 2023-07-18
Synthetic Lungs CBM

anoobindisguise:
#### Summary
Content "synthetic lungs bionic"

#### Purpose of change
I wanted to add a bionic lung option, primarily inspired by Cyberpunk (Edgerunners).

#### Describe the solution
Synthetic Lungs CBM can be installed but not uninstalled. It automatically cancels any Indefatigable, Asthmatic, and other respiratory-related mutations. While installed, it makes your Cardio automatically 4,000 and independent of anything you do (health, exercise, etc) Normally the human limit for Cardio is 3000, + 25% from indefatigable and 10% from athletics skill and 20% from lifestyle (for a total of 4650 cardio) and of course mutations can change this too. While it is installed it also makes you no longer benefit from the stamina regeneration properties of stimulants and the fast metabolism line of mutations. This means a mutant who is also in great physical form can surpass the cardio from this bionic and regenerate more stamina overall, though you really have to work hard to do better than it.
It has a base power cost of 3w. This is based on the statistic that our lungs use about 50 kcal per day to keep us breathing. 
If it is ever deactivated, it automatically reactivates the next turn. Otherwise if you have no power it deducts 2000 stamina per turn and spams you with warning messages and prevents you from intaking oxygen. This won't directly kill you but it will probably get you killed.
Edit: lowered it to 3000. This is enough that if you want to munchkin your character you probably don't want this bionic, it's an immediate benefit over starting levels of cardio but even without indefatigable you can still go as high as 3900.
Edit2: Gave it a health malus over time. Install Leukocyte to disregard it.

I wanted to make it just straight up kill like crush suffocation or drowning but that's probably too mean and it makes sense it has some kind of emergency "low power" mode to prevent its user from dying in an accident of some kind involving forgetting to drink your morning gasoline.

#### Testing
Compiled it and confirmed it worked the way I wanted it to.


Procyonae:
Relevant comments by erk a couple hours ago
image

Personally I'd be down for Rubik to offer you this exclusively if you're asthmatic or have similar lung problems tho

Venera3:
Raising effective cardiac output is not primarily lung-limited, so replacing your lungs should be a way smaller boost than "best case peak human" - you do see VC and compliance increase in response to exercise but the meat of cardio comes from, well, cardiovascular changes.
 
For running out of juice I agree on not necessarily killing you, but I'd zero stamina and perma-wind you - your lungs not working is not a small problem [citation needed].

Venera3:
Somewhere in the 2,500 - 2800 ballpark, tops. It's basically irrelevant how well you oxygenate if your heart can't keep up.

kevingranade:
Setting aside effacacy of the replacement,  I'm also really skeptical of taking out the lungs at all while retaining a mostly human body otherwise.

Lungs seem like one of the last organs you would replace in the process of discarding your entire body in pursuit of a mostly robotic one, and the performance angle wouldn't really be an issue.

fairyarmadillo:
Would this make more sense as a heart? What physiologically is changing in a natural human to boost their cardio fitness? Is it too holistic to be a cbm?

I really like this idea, and it'll be very neat when we have permanent injuries. I'd love to see it work somehow if possible.

anoobindisguise:
So part of Venera's point is that just replacing your lungs alone might not be enough to make your cardiovascular system strong. That introduces an incongruity where the cbm is totally overriding your cardio, but also it can't make your cardio good because it alone isn't 100% of what you need for good cardio. 
So if it overwrites the heart as well, such that it can pump or circulate blood better, then that would be a good way to justify it making your cardio static.
I should probably add a character flag to STOP_HEART_ATTACK or something, even if it totally prevents you from being affected by stim/depressant due to a mechanical heart, there are still a couple hardcoded deaths-by-heart-attack which will need to be edited to respect your organs.

anoobindisguise:
Venera thinks this should be at most lungs only for the time being.

anoobindisguise:
I gave it a health malus to represent the immune system cost. Kevin mentioned it would make you more vulnerable to infection.

PR #67013, opened 2023-07-19
New bionic: "Super Stun Gun"

anoobindisguise:
#### Summary
Content "stun gun bionic"

#### Purpose of change
I want more bionics!!!!

#### Describe the solution
Implements a new bionic, the Super Stun Gun. It is explicitly heavily powered up over a regular handheld one, by the same logic that gave us MagnaVolt in whatever cyberpunk world it came from.
It uses the spell system rather than something like the stun gun item, melee range and deals 15 electric damage for 5 kj.

I had to add a check in bionics.cpp to make this work otherwise it's entirely JSON. if a bionic has a spell on activation, it qualifies as an activated bionic now, when previously it also had to have an activation cost, the BIONIC_TOGGLED flag, a time duration, or fire a bionic gun, and none of those worked for an "activate a spell and nothing else" bionic (you wouldn't want to pay the power cost up front since a spell might not end up firing) 

- Don't call it Super Stun Gun. It's a bit silly but I imagined it was a brand name at some point.
- Lean into the "branding" and have its description mention a manufacturer such as like:
> The MagnaVolt Super Stun Gun was the hottest implant in the personal defense field for those who chose not to carry, at least until the zombies showed up. Even now, brought to your world by the Exodii, it's a pretty potent device. Implanting into your hand, this CBM lets you unleash a highly dangerous shock in melee range which is perfectly capable of inflicting electrical burns or even cardiac arrest.

- Don't put the spell in bionics.json. I can definitely move this somewhere else and it's a little ugly.

#### Testing
Compiled in game and tinkered with it until it started working the way I wanted it to.


Maleclypse:
Move the spell. If there’s a file with cbm spells in it already put it there or make a new file I suppose?

PR #67022, opened 2023-07-19
Fixed randomization of monster special attack cooldown during monster generation and loading of save

### LINKED ISSUE #64907: Monster special attacks' cooldown is randomised on first use
### Describe the bug

When a monster is spawned in, when it first uses it's special attack, the cooldown value appears randomised between 1 and the actual cooldown value specified in the json. Every subsequent special attack is given the correct cooldown value. 

### Attach save file

N/A, easily reproducible.

### Steps to reproduce

1. Use debug menu or an EOC function to spawn a monster with a defined special attack.
2. Activate debug mode filtering for MATTACK and check monster cooldown value.

### Expected behavior

The cooldown value should be respected on the first usage by that monster.

### Screenshots

randomcooldownbug

Win 10 64-bit
cdda-windows-tiles-sounds-x64-2023-04-06-1455 (tested across multiple experimentals, this is the latest I have tried)
Graphical Tiles
Default mods, clean install

### Additional context

This was tested with various different monsters with both spell and regular special attack definitions. In the attached screenshot, 9 Vigilant Amalgamations were spawned at the same time within their sight range to me, and one turn has passed. Note also the messages showing some of the monsters attempting to use the special attack multiple times at once.

WhiteCloud0123:
#### Summary
None

#### Purpose of change
Fix #64907

#### Describe the solution
When a monster is first generated, the special attack cooldown  uses the JSON defined value. When loading save, the values are also not being randomized in a range.

#### Testing
Special attack cooldown was tested before and after loading the save and the results met expectations.

Picture 1 shows that when a monster is first generated, zombie launched an attack.The cooldown time for special attacks is correct. 
ET`WVPD8FPL%W{0K)_V4){5
Picture 2 shows that after loading the save, zombie launched an attack.The cooldown time for monster special attacks is correct.
(T)D9R9}C ~TJ9U9~V3T K8

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -280,7 +280,7 @@ monster::monster( const mtype_id &id ) : monster()
     hp = type->hp;
     for( const auto &sa : type->special_attacks ) {
         mon_special_attack &entry = special_attacks[sa.first];
-        entry.cooldown = rng( 0, sa.second->cooldown );
+        entry.cooldown = sa.second->cooldown;
     }
     anger = type->agro;
     morale = type->morale;
--- a/src/savegame_json.cpp
+++ b/src/savegame_json.cpp
@@ -2645,7 +2645,7 @@ void monster::load( const JsonObject &data )
         const std::string &aname = sa.first;
         if( special_attacks.find( aname ) == special_attacks.end() ) {
             auto &entry = special_attacks[aname];
-            entry.cooldown = rng( 0, sa.second->cooldown );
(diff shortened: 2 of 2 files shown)


WhiteCloud0123:
This may involve feedback related to monster battles in the game, and the experience brought about by such modifications also requires stable investigation. The author's original intention also needs to be carefully distinguished and clarified, so it will be closed.

PR #67025, opened 2023-07-19
reduce unseen hunter bullet armor

anoobindisguise:
#### Summary
Balance "reduce unseen hunter bullet armor"

#### Purpose of change
There was a note in unseen hunters stating specifically to reduce their armor when players started missing more. That's true now.

#### Describe the solution
Lower bullet armor from 30 to 12.

@Venera3 your thoughts wanted.


Venera3 (approved):
The armor was mostly a hacky thing to cut down on evacuee glock supremacy - I'd like these boys to be harder to hit even for trained shooters but it's all right to tone it down in the meantime.

PR #67026, opened 2023-07-19
fix toolset prying power usage

### LINKED ISSUE #66875: Opening crates with integrated toolset uses absurd power
### Describe the bug

The power usage of opening a crate is in excess of 200 kJ  _per crate_

### Attach save file

N/A

### Steps to reproduce

1. Make character with integrated toolset and at least 500 kJ power storage
2. Find a crate
3. Open it
4. See your stored power

### Expected behavior

That opening a crate would not have similar power usage of the probability travel CBM.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0.19044.3086 (21H2)
- Game Version: 634a254 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    No name () [],
]

### Additional context

_No response_

anoobindisguise:
#### Summary
Bugfixes "prying with integrated multitool doesn't use power"

#### Purpose of change
Fixes #66875

#### Describe the solution
This does NOT work. I'm opening it so I can remind myself to do this. Crowbar use actions are hardcoded in a very weird way such that they don't actually respect the ammo_scale settings you gave them (why does crowbar prying have hardcode for using power anyways??) so this fix will likely require c++ code.

#### Testing
Again this does NOT work yet.

### DIFF
--- a/data/json/items/tool/integrated.json
+++ b/data/json/items/tool/integrated.json
@@ -150,7 +150,7 @@
       [ "BUTCHER", -15 ]
     ],
     "use_action": [
-      { "type": "CROWBAR_WEAK", "ammo_scale": 0 },
+      { "type": "CROWBAR", "ammo_scale": 0 },
       {
         "type": "repair_item",
         "item_action_type": "repair_metal",
--- a/data/json/items/tool/integrated.json
+++ b/data/json/items/tool/integrated.json
@@ -150,10 +150,7 @@
       [ "BUTCHER", -15 ]
     ],
     "use_action": [
-      {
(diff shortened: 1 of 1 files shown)


anoobindisguise:
OK turns out the iuse doesn't respect ammo_scale if it uses CROWBAR_WEAK but it _does_ if it's just CROWBAR. I wanted to make it crowbar_weak because that only tries to pry stuff like crates, but that's not possible. It's working now though.

Maleclypse:
I've converted to draft so it's not picked up in my to review filter.

PR #67038, opened 2023-07-20
Damage types: armour/material fixes

dseguin:
#### Summary
None

#### Purpose of change
There are a few bugs in the way items calculate resistance to certain damage types:
1. Item resistance for damage types that are derived from a different damage type (e.g. "stab") use the derived value regardless of whether the resistance is explicitly defined. So for a material you can set a "cut" resistance of 2 and a "stab" resistance of 10, but the final "stab" resistance for the item will be based on 1.6 (0.8 x "cut").
2. The material resistance map wasn't being properly processed by `finalize_damage_map()`. It turns out `generic_factory::finalize()` isn't calling `material_type::finalize()`.

#### Describe the solution
1. Don't bother with recursion in `item::_resist()`. Instead, handle the derived resistance values for each material.
2. Call `finalize_damage_map()` in `material_type::finalize_all()`.

Instead of deriving the resistance at the item level, we could make the material store the derived values. Might be a bit misleading if someone omits material resistance values.

#### Testing
Added a couple of test cases for sanity checking.

Also loaded up the game with a lycra stab resistance set to 500:

Screenshot from 2023-07-20 01-09-39

### DIFF
--- a/data/mods/TEST_DATA/items.json
+++ b/data/mods/TEST_DATA/items.json
@@ -1343,6 +1343,29 @@
     ]
   },
   {
+    "id": "test_zentai_resist_stab_cut",
+    "type": "ARMOR",
+    "name": { "str": "test zentai" },
+    "description": "Full-body suit completely covering all body parts, including the head, mouth, and eyes.",
+    "weight": "300 g",
+    "volume": "500 ml",
+    "price": 2500,
+    "price_postapoc": 50,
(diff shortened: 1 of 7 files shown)


Termineitor244:
Hahahaha nice! And it makes me laugh to see your test case of a superman suit with 500 pierce resistant hahaha, very fitting for the man of steel!

PR #67043, opened 2023-07-20
[Magiclysm] Fix "Monsters don't cast spells"

### LINKED ISSUE #67032: [Magiclysm] Monsters don't cast spells
### Describe the bug

Monsters which are supposed to cast spells don't seem to be able to, and will not do so under any circumstances. This can be strange behaviors for creatures designed around magic, and remove the challenge of some fights.

### Attach save file

Tested on multiple worlds, this seems to transcend save files.

### Steps to reproduce

1. Spawn a monster that is supposed to cast spells (I used Lizardfolk Shamans and Feral Magicians)
2. Go to fight them, or pit them against another enemy.
3. They don't cast spells, and use melee instead.

### Expected behavior

I expect that monsters will use their magic.

### Screenshots

_No response_

-OS: Windows 11
-Game Version: cdda-windows-tiles-x64-2023-07-09-0709
-Graphics Version: Tiles
-Ingame Language: English (US)
-Mods: default mods + magiclysm

### Additional context

_No response_

WhiteCloud0123:
#### Summary
None

#### Purpose of change
Fix #67032 

#### Describe the solution
In the game, If the caster's target is not oneself and the distance between the caster and the target is greater than the release distance of spells, then the release will fail. `spell_shaman_base` and ` spell_feral_wizard_base` did not specify a range for spell release. So the repair method is very simple, which is to specify the release range of a spell.

Additionally, I personally believe that there is a problem with the shaman summoning the permanent crocodile. Because if shamans can summon permanent crocodiles without paying any price, then these crocodiles can become a very powerful productivity for their population, and zombies and other creatures are not worth mentioning, which is obviously not very appropriate. So I transformed them into short-term existence.

I have set the casting distance for shamans and wizards to 10. Their cooling time is  set to  10. I think this is more appropriate.

#### Testing
The behavior of releasing spells is normal.


GuardianDll (approved):

GuardianDll:
Actually wait, you forgot to change feral mage spell <https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/mods/Magiclysm/Spells/monsterspells.json#L228>

WhiteCloud0123:
> Actually wait, you forgot to change feral mage spell https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/mods/Magiclysm/Spells/monsterspells.json#L228

Thank you for the review. Perhaps you didn't see it, it has already been modified

GuardianDll:
oh, it's in the monsterspells.json, somehow i miss it

WhiteCloud0123:
I updated the PR description and there was a typo. In the description of PR, it should be "if the target of the caster is not oneself" instead of "the target of the caster is oneself".

PR #67044, opened 2023-07-20
A bunch of random item reworks: 9

Karol1223:
#### Summary
None

#### Describe the solution

<details>
<summary>LUG WRENCH</summary>

- Weight: 454g -> 1565g
- Volume: 1500ml -> 460ml
- ``price_preapoc``: 12$ -> 23$
- ``to-hit`` values now use modifiers
- Disassembly yield increased to account for weight increase
- Crafting recipe requires more steel to account for weight increase

</details>

<details>
<summary>CROWBAR</summary>

- Weight: 550g -> 1456g
- Volume: 100ml -> 520ml
- Longest side: 45cm -> 46cm
- ``price_preapoc``: 13$ -> 26$
- Crafting recipe is no longer reversible - introduced a manually defined disassembly recipe

</details>

<details>
<summary>CLAW BAR</summary>

- Weight: 350g -> 318g
- Volume: 64ml -> 470ml
- Longest side: 25cm -> 24cm
- ``price_preapoc``: 10$ -> 19$
- ``to-hit`` values now use modifiers
- Crafting recipe requires less steel, partially to account for weight decrease and partially because the amount was copied from the crowbar so there was 1.5kg+ of material loss
- Crafting recipe is no longer reversible - introduced a manually defined disassembly recipe

</details>

<details>
<summary>HEAVY CROWBAR</summary>

- Weight: 4kg -> 6201g
- Volume: 725ml -> 1740ml
- Longest side: 150cm -> 152cm
- ``price_preapoc``: 100$ -> 90$
- Disassembly yield increased to account for weight increase
- Disassembly now is ``BRISK`` instead of ``MODERATE`` exercise level
- Crafting recipe requires more steel to account for weight increase

</details>

#### Additional context
Only 4 items instead of the usual 5 because these 4 have already inflated the line limit I typically go with for those PRs


BjoHart:
I checked a crowbar in RL. It weights over 1kg...
Why reduce the weight of it even more?

Karol1223:
> I checked a crowbar in RL. It weights over 1kg... Why reduce the weight of it even more?

Check the specific amazon listing

BjoHart:
Description was updated. Nevermind...

Karol1223:
When I first post those PRs I copy paste the description from the last one so for the first few minutes they're a lie. Sorry about the confusion!

Karol1223:
> so we cant use crowbar as a weapon ?

I don't understand where this question comes from. What about this PR makes you think the crowbar can no longer be used as a weapon?

alp7292:
i hate how weight directly affects attack speed before patch you could use crowbar as a good melee weapon now makeshift crowbar (literally a pipe smashed by rock ) is better weapon than real crowbar. i think we need some weapon speed modifier for items because something beign heavy doesnt necessarily means you swing it slower ( it affects but it also depends on the handle and center of mass) for example i can swing a hammer and a crowbar at about the same speed but one would be more tiring.

Karol1223:
It swings slower but it does the exact same DPS. I don't believe there's an issue in that, given previously the crowbar for some reason had less volume than some of our knives.

PR #67048, opened 2023-07-20
Make padded feet quieter

randomtyper:
#### Summary
Balance "Makes padded feet quieter"

#### Purpose of change

Padded feet mutation description implies there is quite a bit of tissue to absorb impact and dampen noise, but there is no mechanical reflection of that in game. Animals with padded feet and retractable claws can move pretty damn quiet, after all (example: almost all felines, low noise is one of reasons they are so good at ambush hunting).

#### Describe the solution
Mentioned mutation decreases walking noise by about 50% of default now.

Also rewrite description (will likely do it RN).

Rework other mutations that should only affect footstep noise to use enchantment that only affects footstep noise instead of changing overall noise modifier ("noise_modifier").

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -3947,7 +3947,7 @@
     "name": { "str": "Padded Feet" },
     "points": 1,
     "visibility": 1,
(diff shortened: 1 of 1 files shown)


Rivet-the-Zombie:
Sorry, misclicked.

PR #67049, opened 2023-07-20
Fix handflare burning out crash

### LINKED ISSUE #67045: Game crashes shortly after a precise amount of time passes since updating to latest experimental
### Describe the bug

An error appears upon loading the save, once you ignore the error and the save has been loaded the game will crash 35 (in game) seconds after loading no matter what you do or where you go in the game.

### Attach save file

[Harveysburg-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12104924/Harveysburg-trimmed.tar.gz)

### Steps to reproduce

1. Load up save
2. Have 35 seconds of in game time pass through any method
3. Upon reaching 7:06:44 AM the game will crash, this has happened 100% of the time with a dozen or so attempts 

### Expected behavior

For the game not to crash of course 

### Screenshots

_No response_

OS: Windows 10
Game Version: f83ffe1
Graphics version: Tiles
Ingame language: English
Mods loaded: dda, no_npc_food, personal_portal_storms, no_fungal_growth, package_bionic_professions

### Additional context

 DEBUG    : invalid effect type id 

 FUNCTION : obj
 FILE     : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\effect.cpp
 LINE     : 94
 VERSION  : f93ffe1
image
Crash log: 
[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/12105040/crash.log)

SurFlurer:
#### Summary
none

#### Purpose of change

Fix #67045

When the handflare burns out, it will be "invoked" that eventually calls `firestarter_actor::use`. This behavior has been there before 0.G. But this will crash the game now, as `firestarter_actor::use` will have a null character pointer.

#66866 will also solve this. 

#### Describe the solution

If invoked from active item processing ( with a null character *p), just do nothing and return.

Wait for #66866 to be merged.

But this kind of validation should exist anyway, as the pointer CAN be nullptr.

#### Testing
Crash fixed.

### DIFF
--- a/src/iuse_actor.cpp
+++ b/src/iuse_actor.cpp
@@ -1314,11 +1314,10 @@ std::optional<int> firestarter_actor::use( Character *p, item &it, bool t,
 
     // TODO: fix point types
     tripoint_bub_ms pos( spos );
-    float light = light_mod( p->pos() );
-    if( !prep_firestarter_use( *p, pos ) ) {
+    if( !p || !prep_firestarter_use( *p, pos ) ) {  // if invoked from active item processing, do nothing.
         return std::nullopt;
     }
-
+    float light = light_mod( p->pos() );
     double skill_level = p->get_skill_level( skill_survival );
     /** @EFFECT_SURVIVAL speeds up fire starting */
     float moves_modifier = std::pow( 0.8, std::min( 5.0, skill_level ) );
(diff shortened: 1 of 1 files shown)


Hirmuolio:
After  #66866 is merged the pointer _should_ never be nullptr on actions that require player.
Maybe after  #66866 there should be a debugmsg like "[item] called [action] that requires player but player was not present" so bad json would give useful error message instead of crashing and requiring reading call stack.

PR #67055, opened 2023-07-20
"relative": can be used with qualities

Procyonae:
#### Summary
None

#### Purpose of change

```
"delete": { "qualities": [ [ "BUTCHER", 8 ] ] },
"extend": { "qualities": [ [ "BUTCHER", 10 ] ] },
```
looks stupid and isn't even that useful bc you have to specify the correct value you inherited

#### Describe the solution

You can now change inherited qualities values like `"relative": { "qualities": [ [ "HAMMER", -1 ], [ "SAW_M", 1 ] ] }` to add 1 metal sawing and removing 1 hammering
You also don't need to specify a quality level when deleting an inherited quality so `"delete": { "qualities": [ [ "COOK", 3 ], [ "DRILL", 1 ] ] },` becomes `"delete": { "qualities": [ "COOK", "DRILL" ] },`

Attempted to find everywhere this is being used inc. mods and change it, tests should throw any others anyway tho

#### Testing

Checked positive and negative numbers work as expected
Checked new delete method works

### DIFF
--- a/data/json/items/generic/dining_kitchen.json
+++ b/data/json/items/generic/dining_kitchen.json
@@ -1011,8 +1011,8 @@
         "rigid": true
       }
     ],
-    "delete": { "qualities": [ [ "COOK", 3 ] ] },
-    "extend": { "qualities": [ [ "HAMMER", 1 ], [ "COOK", 2 ] ] },
+    "extend": { "qualities": [ [ "HAMMER", 1 ] ] },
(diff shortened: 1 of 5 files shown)


Procyonae:
Realised I should probably make delete work without a known value too

Procyonae:
(Bumping bc I accidently got the draft -> open free pass on the tests)

PR #67058, opened 2023-07-20
auto eat/drink oldest item first

### LINKED ISSUE #67034: Auto Drink / Auto Eat zones should prioritize older food

Right now, comestibles in Auto Eat / Auto Drink zones are consumed in the order they're stored in the tile item list. This can lead to eating items with a long shelf life first, which wastes calories by allowing the stuff that's closer to rotting to rot before being eaten.

### Solution you would like.

When consuming from an Auto zone, the least fresh non-rotten item should be eaten first.

_No response_

### Additional context

_No response_

nornagon:
#### Summary
Features "Auto Eat/Drink zones now prioritize less-fresh items"

#### Purpose of change

Closes #67034

#### Describe the solution

This changes the way items are visited when considering them for auto eat/drink zones to find the "best" item to consume according to some comparison function, instead of the first available. We visit all items in all zones and save the best one we've seen, then consume that one.

I've copied over _almost_ all the logic from `comestible_inventory_preset::sort_compare`, with the exception that for items which are equally valid for consumption, the new logic doesn't discriminate between them (the existing logic sorts by name). Ideally this would match `comestible_inventory_preset::sort_compare` exactly, or even better, use a factored-out common implementation, such that the "first item consumed" is always the first one in the (E)at menu, which I think would aid understanding of the feature. I haven't done that yet because I'm not sure about the performance characteristics of the name cache which `inventory_preset` uses and whether that's relevant here.

I've also added a new `visit_item_contents` function that preserves `item_location` while recursively descending through item pockets. This should probably move to `visitable`, but since it requires a "root" `item_location` to start with, which `visitable` doesn't have, I've kept it separate for now.

This algorithm has the same worst-case time complexity as the previous algorithm, in the case that there are no consumable items, all items in all zones will be visited. The previous algorithm created a copy of the item list for each map tile in a zone, but stopped as soon as an item was found in a given tile, so in the best case the previous algorithm was faster. The new algorithm has Θ(d) space complexity where _d_ is the depth of the most deeply-nested item, as it uses a recursive function to traverse items, vs Θ(n) for the old algorithm, where _n_ is the number of items in the map tile. So the new algorithm allocates much less.

It could be worth also including distance to zone in the calculation, though it's unclear how to weight that against freshness.

#### Testing

I've tested this manually but not extensively.


src/activity_item_handling.cpp
```diff
@@ -3177,6 +3180,82 @@ static std::optional<tripoint_bub_ms> find_refuel_spot_trap(
     return {};
 }
 
+// Visits an item and all its contents.
```
Maleclypse, line 3183:
```suggestion
// Visits an item and all its contents.,
```
Resetting test suite

src/activity_item_handling.cpp
```diff
@@ -3177,6 +3180,82 @@ static std::optional<tripoint_bub_ms> find_refuel_spot_trap(
     return {};
 }
 
+// Visits an item and all its contents.,
```
Maleclypse, line 3183:
```suggestion
// Visits an item and all its contents.
```

Issue #62544, opened 2022-12-03
A dock is generated in wrong direction


vsisko:
### Describe the bug

I started on an island near with a dock building. But the dock is generated in wrong direction with square lake toward the ground.

### Steps to reproduce

I'm not sure that the next steps can help.
2. Check a generated dock building.

### Expected behavior

Expected orientation of the dock in accordance with generated terrain.

### Screenshots

![Даррон Полсон _2022-12-03T15-47-00](https://user-images.githubusercontent.com/5163151/205442066-7b9109c6-feed-4cb8-ab64-a03541ffaedd.png)
![Dock map](https://user-images.githubusercontent.com/5163151/205442070-a6caab20-8aeb-4bfa-9c0e-43899b2046d5.png)

### Versions and configuration

OS: Win 10
Game version: ec0f3a4
Graphic version: Tiles
Ingame lanquage: System (Russian)
Mods: "dda", "no_npc_food", "no_fungal_growth", "package_bionic_professions", "translate_dialogue", "sees_player_retro"

PR #38894, opened 2020-03-19
Rework river generation

Diabolus-Engi:
#### Summary

```SUMMARY: Bugfixes "Reworks river generation on overmap."```

#### Purpose of change

Reworks river generation so that they cross overmap boundaries better and flow more realistically. 

Fixes: #36630
Fixes: #38036

#### Describe the solution

1. Adds river nodes for the overmap to track entry/exit points of rivers so that they can be continued with more accuracy to adjacent overmaps. Also includes the size of the river. This provides some method continuity for rivers and can later be used to maintain other properties for the river such as fish populations, contaminates perhaps.

2. Reworks plotting of the river and generation by using a quadratic bezier curve, coupled with part of the original weighted random walk method that helps to create varied river sizes while the bezier curve allows bends in the river. The control points probably still need some work, but hoping to catch the eye of someone who knows more math than myself for that.

3. Changes the polishing of rivers to apply terrain mask based on adjacent tiles. I found this implementation to be a little cleaned than the original.

4. Rivers have a chance to generate branches of a smaller scale that will merge into the river further down and also generate branches to a random nearby point.

5. Rivers will 'pool up' creating small lakes if they haven't terminated on the overmap boundary.

**TODO:** 
- Remember river scales between overmaps to allow different river sizes to cross.
- Add streams, smaller versions of rivers. Requires mapgen work in a future PR.

#### Describe alternatives you've considered

Well, I originally set out just to fix overmap river borders. After many days of wrestling with the current implementation and testing various alternative methods, I found it was unfeasible to continue without rewriting a lot of the river generation.

Once elevations are added, then we can advance further into erosion models etc that would provide realistic rivers. While I played with the idea of using perlin noise, I opted against it in favor of waiting for elevations.

#### Testing

Generate a new map, teleport to adjacent overmaps and follow the rivers between overmaps.

#### Additional context
(Slightly outdated pre-random branches)
o_10

(Updated)
o_10

PR #67065, opened 2023-07-21
[Innawoods] - Added mutable stream

SandwichPie:
#### Summary
Content "Adds a mutable stream to the mapgen"

#### Purpose of change

Rivers are large and cumbersome, but most bodies of inland water are shallow and small. Especially in a wilderness mod like innawoods a stream/brook should be a point of interest where foliage and animals gather. It should add some visual interest and a respite from rolling fields and forests.

#### Describe the solution

Most map features are hardcoded, but I made this one take advantage of the mapgen mutables feature. This allows a lot of control over how the map elements generate and should be quick and easy to iterate over.

Mutables have several limitations however, making it not behave ideally in several situations.
+ Firstly, you cannot select connections to anything other than specific hardcoded points, so I could not have it flow into river or pond tiles.
+ Secondly, mutables seem to lack a feature to check if they are generating inside another mutable of the same type, so having two long streams close can cause map generation failures. Due to this I can only spawn one per overmap chunk.
+ Mutables cannot spread across chunk boundries.
+ Mutables cannot be set to terminate on certain conditions, like meeting a chunk boundry. Because of this the stream will prefer to deflect off chunk boundries instead of setting an end point.
+ And I was unable to allow prior map generation to remain when a stream passes through, though I believe that was a failure on my part. This means that trees are destroyed when the stream passes through a forest.

That being said, I believe that since so much mapgen was removed from innawoods it can get quite repetitive, and the interest this adds back into to the maps makes up for the shortcomings of the implementation.

I do not believe that with mutables in this state that it would be appropiate to add larger mutable elements like this to be core game, but with a little more work I'm hopeful it would be very possible to replace the hardcoded map generation like rivers and roads.

+ Hardcoded mapgen. It'd be a lot harder to implement, and we want less hardcoding and more JSON anyway.
+ Not having streams.

#### Testing

Generated and ran across miles and miles of streams with slightly different settings. The monster spawn rate is probably too low, but it's better too low than too high. This can increased as needed.

The willingness for the stream to double back on itself could certainly be decresed, but I am happy with the current values. I don't want it to be too straightforward as this way it makes it a little more interesting.

#### Additional context

2023-07-21-090759_291x986_scrot
2023-07-21-090843_621x1007_scrot


Procyonae:
A few thoughts, firstly big thumbs up for adding streams in general, I like the meandering and the way they're spawning on the overmap looks good ^^
Rivers are very much hardcoded but you could still make it so you have an end tile that can spawn over the bank tiles with nests to make it fit in (not saying this is easy and it could totally be added later).
I agree that this shouldn't be added to the base game but more bc I'd prefer river gen itself get reworked to be more sensible say by someone rezzing #38894 than by just adding streams and having an extra excuse not to touch river gen, while more innawood mapgen seems great in general due to the current lack of variety like you say.
While unhardcoding river with a mutable would potentially be possible, the chunk boundary stuff and lack of advanced logic mean they would look even worse than our current rivers so realistically we'd need multi tile wide connection support that I doubt will come any time remotely soon and would be a pain in the ass to make work. The only other way I can think of is having them be 1 tile wide connections but spawning them from and to the same point multiple times with no preference to spawn over themselves to make them seem wider and add a hack to make the boundaries wider, and I doubt that would look good either.
What issue(s) did you run into using fallback_predecessor to make it fit in with forests/swamps? I'm unhardcoding road and were quite surprised that I've managed to get away with fallback_predecessor there so I imagine it's possible to work around whatever hurdle you hit.
From you're description I'm not sure you realised special_locations isn't hardcoded in any way and there's no harm in adding extras if you want to try to some more fun stuff.
Just generally adding more nests in future and allowing them to cross trails with nests if you can get fallback_predecessor working (with a stepping stone or little z0 bridge for example) to add some variety would be cool.
I'm not suggesting this can't be merged in it's current state in any way though it already looks like a great addition just wanted to share my thoughts.

data/mods/innawood/overmap/overmap_terrarin/overmap_terrain_stream.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  {
+    "type": "overmap_terrain",
+    "id": "stream_corner",
+    "copy-from": "generic_open_land",
+    "name": "stream",
+    "sym": ".",
+    "color": "light_blue"
+  }
+]
```
Procyonae, line 26:
```suggestion
[
  {
    "type": "overmap_terrain",
    "id": [ "stream", "stream_end", "stream_corner" ],
    "copy-from": "generic_open_land",
    "name": "stream",
    "sym": ".",
    "color": "light_blue"
  }
]
```
Am I missing some reason why it's necessary to split these up?

SandwichPie: No, you aren't. I did not realise that the IDs could be put into a list like that. I'll do this in the future.

SandwichPie:
Thank you Procyonae, I agree entirely with your suggestions.

My problem with  'fallback_predecessor' was me mistaking a flag as replacing it's function, but your comment let me find the relivent documentation. I've fixed the issue with the forests being overwrote now, I'll commit that change shortly after a couple more quick tests.

I know overmap specials are not hardcoded, but I wanted to keep this commit short becuase I have a habbit of making mods and never pushing them. I plan to followup this commit with some nested clay banks, cattails and better monster spawn rules. I'm not familiar with nesting yet though, so I thought it would be better to do that seperately.

I'm not sure how you are suggesting I connect the stream to a river using nesting though. I'm willing to look into it but I am at a loss with how that would be done right now. According to the docs, the connections option won't connect to rivers and I don't think there is a method to select certain overmaps if it detects a river bank. I'd be happy to be shown otherwise though.

And I also appreciate the surrounding context of the river generation problems, it was very interesting.

Procyonae:
Nice, having it fit in with the terrain will make it really pretty ^-^

Small incremental PRs are defo sensible, excited to see where you take it \:D

If you want any help with nesting (or any other mapping) at any point you can dm me on discord (or just mention me on the devcord) at the same username

Once this is merged I might have a play with the connecting to river thing, it might well not be possible anyway thinking about it bc river banks seem really trolly (I've tried to fix the seemingly easy issue #62544 multiple times and can't get anything to work) so I wouldn't worry about it for now

Light-Wave:
I'm liking this a great deal. I agree with the previously made points that small incremental PRs are good here, and that it can stay in Innawood for now. Although it would be neat if it could enter mainline some day in the perhaps distant future.
I have been meaning to give Innawood players a way to more reliably get domestic animals, and this seems like a good opportunity to do so.
I'm 80% sure that a chance of 0.7 gets cast to an int and rounded down to 0 [here](https://github.com/CleverRaven/Cataclysm-DDA/blob/2c4c7ee663d5234ea400878b0d479fce8c51b5fa/src/mapgen.cpp#L2399).
I played around with it a little, and I think setting "GROUP_DOMESTIC" spawn chance to around 3 and "GROUP_FOREST" to maybe 25 would be more reasonable values.

SandwichPie:
I did see a few instances of horses spawning, but that could have been from the wilderness spawn group so thanks for letting me know that can be a problem since that's a very important thing to note.

In regards to monster spawns, I'm already starting on a followup PR to address that. I'm remarkably displeased with how they are set up currently. I don't feel like the wildreness spawn group is weighted correctly for a stream and I want to be a lot more careful with how many monsters spawn for each species.

To address the rounding down issue, the followup PR may have only certain tiles have a chance to spawn domestic animals or I'll just try to weight it better with the custom spawn groups. The latter is probably the more sensible approach.

I'll change the spawn values in the current PR to what you suggeted for now though.

Light-Wave (approved):

Procyonae (approved):

NetSysFire:
This would fit the main game, too with some slight adjustments.

Light-Wave:
> This would fit the main game, too with some slight adjustments.

I absolutely support putting these into mainline once they have gotten some more improvements. If you want to work towards that goal, then feel free to 👍

PR #67074, opened 2023-07-21
Fix: parse missing fields for translation

### LINKED ISSUE #67072: Translatable strings don't get generated for some EOC related texts
### Describe the bug

A translator reported on dev Discord [(related discussion and some preliminary analysis)](https://discord.com/channels/598523535169945603/873624399998234675/1131897803908591679) that some translatable strings were missing for Sky Island mod. On further investigation, at least the following don't get parsed by the translation scripts:

* u_query EOC, example: https://github.com/CleverRaven/Cataclysm-DDA/blob/2c4c7ee663d5234ea400878b0d479fce8c51b5fa/data/mods/Sky_Island/EOCs.json#L20
* item use_action EOC description, example: https://github.com/CleverRaven/Cataclysm-DDA/blob/2c4c7ee663d5234ea400878b0d479fce8c51b5fa/data/mods/Sky_Island/warpitems.json#L9
* name and description of result_eocs using recipes, example: https://github.com/CleverRaven/Cataclysm-DDA/blob/2c4c7ee663d5234ea400878b0d479fce8c51b5fa/data/mods/Sky_Island/recipes.json#L239

### Attach save file

N/A

### Steps to reproduce

1. `make localization LANGUAGES=all`
2. `grep "Begin searching for a spot to teleport to" lang/po/*`
3. `grep "Forgiveness to the innocent" lang/po/*`
3. `grep "materialize random artifact" lang/po/*`
4. Observe that greps fail

### Expected behavior

Strings to be present in translation files

### Screenshots

_No response_

N/A

### Additional context

N/A

dseguin:
#### Summary
None

#### Purpose of change
- Fixes #67072

#### Describe the solution
Added some parsing for various fields that were missing in the pot file.

Also made some code corrections to account for those translations.

#### Testing
Running `lang/update_pot.sh` reads the missing fields and adds them to `lang/po/cataclysm-dda.pot`.

Also tested in-game by adding a French translation for a `"u_query"` message:

Screenshot from 2023-07-21 12-12-28

### DIFF
--- a/lang/string_extractor/parsers/condition.py
+++ b/lang/string_extractor/parsers/condition.py
@@ -1,0 +1,20 @@
+from ..write_text import write_text
+
+
+def parse_condition(conditions, origin):
+    if not type(conditions) is list:
+        conditions = [conditions]
+    for cond in conditions:
+        if type(cond) is dict:
+            if "u_query" in cond:
+                write_text(cond["u_query"], origin,
+                           comment="Query message shown in a popup")
+            if "npc_query" in cond:
+                write_text(cond["npc_query"], origin,
+                           comment="Query message shown in a popup")
+            if "and" in cond:
+                parse_condition(cond["and"], origin)
+            if "or" in cond:
+                parse_condition(cond["or"], origin)
+            if "not" in cond:
+                parse_condition(cond["not"], origin)
--- a/lang/string_extractor/parsers/condition.py
+++ b/lang/string_extractor/parsers/condition.py
@@ -6,10 +6,11 @@ def parse_condition(conditions, origin):
         conditions = [conditions]
     for cond in conditions:
         if type(cond) is dict:
-            if "u_query" in cond:
+            # if *_query is an object, the message is taken from elsewhere
+            if "u_query" in cond and type(cond["u_query"]) is str:
                 write_text(cond["u_query"], origin,
                            comment="Query message shown in a popup")
-            if "npc_query" in cond:
+            if "npc_query" in cond and type(cond["npc_query"]) is str:
                 write_text(cond["npc_query"], origin,
                            comment="Query message shown in a popup")
             if "and" in cond:
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -243,6 +243,30 @@ str_or_var get_str_or_var( const JsonValue &jv, const std::string &member, bool 
     return ret_val;
 }
 
+translation_or_var get_translation_or_var( const JsonValue &jv, const std::string &member,
+        bool required, const translation &default_val )
+{
+    translation_or_var ret_val;
+    translation str_val;
+    if( jv.read( str_val ) ) {
+        ret_val.str_val = str_val;
+    } else if( jv.test_object() ) {
+        const JsonObject &jo = jv.get_object();
+        if( jo.has_member( "mutator" ) ) {
+            // if we have a mutator then process that here.
+            ret_val.function = conditional_t::get_get_string( jo );
+        } else {
+            ret_val.var_val = read_var_info( jo );
+            ret_val.default_val = default_val;
+        }
+    } else if( required ) {
+        jv.throw_error( "No valid value for " + member );
+    } else {
+        ret_val.str_val = default_val;
+    }
+    return ret_val;
+}
+
 tripoint_abs_ms get_tripoint_from_var( std::optional<var_info> var, dialogue const &d )
 {
     tripoint_abs_ms target_pos = get_map().getglobal( d.actor( false )->pos() );
@@ -1269,12 +1293,12 @@ void conditional_t::set_one_in_chance( const JsonObject &jo, const std::string &
 
 void conditional_t::set_query( const JsonObject &jo, const std::string &member, bool is_npc )
 {
-    str_or_var message = get_str_or_var( jo.get_member( member ), member, true );
+    translation_or_var message = get_translation_or_var( jo.get_member( member ), member, true );
     bool default_val = jo.get_bool( "default" );
     condition = [message, default_val, is_npc]( dialogue const & d ) {
         const talker *actor = d.actor( is_npc );
         if( actor->get_character() && actor->get_character()->is_avatar() ) {
-            std::string translated_message = _( message.evaluate( d ) );
+            std::string translated_message = message.evaluate( d );
             return query_yn( translated_message );
         } else {
             return default_val;
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -254,7 +254,7 @@ translation_or_var get_translation_or_var( const JsonValue &jv, const std::strin
         const JsonObject &jo = jv.get_object();
         if( jo.has_member( "mutator" ) ) {
             // if we have a mutator then process that here.
-            ret_val.function = conditional_t::get_get_string( jo );
+            ret_val.function = conditional_t::get_get_translation( jo );
         } else {
             ret_val.var_val = read_var_info( jo );
             ret_val.default_val = default_val;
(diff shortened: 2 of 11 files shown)


src/condition.cpp
```diff
@@ -1274,7 +1274,7 @@ void conditional_t::set_query( const JsonObject &jo, const std::string &member,
     condition = [message, default_val, is_npc]( dialogue const & d ) {
         const talker *actor = d.actor( is_npc );
         if( actor->get_character() && actor->get_character()->is_avatar() ) {
-            std::string translated_message = _( message.evaluate( d ) );
+            std::string translated_message = to_translation( message.evaluate( d ) ).translated();
```
Qrox, line 1277: This is equivalent to the previous line and not necessary, although I prefer `message.evaluate( d )` to be `translation` so it supports the translation context and text style checks.

dseguin: Cheers, I didn't realize it worked the same.

Tangent:
I just tried using a translation equivalent of `str_or_var` like you suggested, but it doesn't seem to pick up the translated text.

<details>
<summary>Tested with this diff</summary>

```diff
diff --git a/src/condition.cpp b/src/condition.cpp
index cfb2b275f0..0cb008af86 100644
--- a/src/condition.cpp
+++ b/src/condition.cpp
@@ -221,11 +221,14 @@ duration_or_var get_duration_or_var( const JsonObject &jo, const std::string &me
 }
 
 str_or_var get_str_or_var( const JsonValue &jv, const std::string &member, bool required,
-                           const std::string &default_val )
+                           const std::string &default_val, bool as_translation )
 {
     str_or_var ret_val;
     if( jv.test_string() ) {
         ret_val.str_val = jv.get_string();
+        if( as_translation && ret_val.str_val.has_value() ) {
+            ret_val.trans_str_val = to_translation( ret_val.str_val.value() );
+        }
     } else if( jv.test_object() ) {
         const JsonObject &jo = jv.get_object();
         if( jo.has_member( "mutator" ) ) {
@@ -239,6 +242,9 @@ str_or_var get_str_or_var( const JsonValue &jv, const std::string &member, bool
         jv.throw_error( "No valid value for " + member );
     } else {
         ret_val.str_val = default_val;
+        if( as_translation ) {
+            ret_val.trans_str_val = to_translation( default_val );
+        }
     }
     return ret_val;
 }
@@ -1269,12 +1275,12 @@ void conditional_t::set_one_in_chance( const JsonObject &jo, const std::string &
 
 void conditional_t::set_query( const JsonObject &jo, const std::string &member, bool is_npc )
 {
-    str_or_var message = get_str_or_var( jo.get_member( member ), member, true );
+    str_or_var message = get_str_or_var( jo.get_member( member ), member, true, "", true );
     bool default_val = jo.get_bool( "default" );
     condition = [message, default_val, is_npc]( dialogue const & d ) {
         const talker *actor = d.actor( is_npc );
         if( actor->get_character() && actor->get_character()->is_avatar() ) {
-            std::string translated_message = to_translation( message.evaluate( d ) ).translated();
+            std::string translated_message = message.evaluate( d );
             return query_yn( translated_message );
         } else {
             return default_val;
diff --git a/src/condition.h b/src/condition.h
index d3ec60f639..2ee85e4e93 100644
--- a/src/condition.h
+++ b/src/condition.h
@@ -56,7 +56,7 @@ const std::unordered_set<std::string> complex_conds = { {
 } // namespace dialogue_data
 
 str_or_var get_str_or_var( const JsonValue &jv, const std::string &member, bool required = true,
-                           const std::string &default_val = "" );
+                           const std::string &default_val = "", bool as_translation = false );
 dbl_or_var get_dbl_or_var( const JsonObject &jo, const std::string &member, bool required = true,
                            double default_val = 0.0 );
 dbl_or_var_part get_dbl_or_var_part( const JsonValue &jv, const std::string &member,
diff --git a/src/dialogue_helpers.cpp b/src/dialogue_helpers.cpp
index a5e189aae5..daf79c0841 100644
--- a/src/dialogue_helpers.cpp
+++ b/src/dialogue_helpers.cpp
@@ -79,6 +79,9 @@ std::string str_or_var::evaluate( dialogue const &d ) const
     if( function.has_value() ) {
         return function.value()( d );
     }
+    if( !trans_str_val.empty() ) {
+        return trans_str_val.translated();
+    }
     if( str_val.has_value() ) {
         return str_val.value();
     }
diff --git a/src/dialogue_helpers.h b/src/dialogue_helpers.h
index 929b9cb1b5..12f505e759 100644
--- a/src/dialogue_helpers.h
+++ b/src/dialogue_helpers.h
@@ -8,6 +8,7 @@
 #include "global_vars.h"
 #include "math_parser.h"
 #include "rng.h"
+#include "translation.h"
 #include "type_id.h"
 
 struct dialogue;
@@ -150,6 +151,7 @@ var_info process_variable( const std::string &type );
 
 struct str_or_var {
     std::optional<std::string> str_val;
+    translation trans_str_val;
     std::optional<var_info> var_val;
     std::optional<std::string> default_val;
     std::optional<std::function<std::string( const dialogue & )>> function;
```

</details>

But if I wrap `message.evaluate( d )` with `_( )` or `to_translation( ).translated()`, it suddenly translates correctly again. `message` gets a translation object with the english text, but it looks like it's loaded before translation strings? And recreating the translation when the condition gets called seems to get it cached properly?

Qrox: What I had in mind is to make a separate class (`translation_or_var`) that reads in `translation` directly using `JsonValue::read()` because using `JsonValue::get_string()` will only read a string without checking for the text style and cannot read an object that contains the translation context.

I cannot immediately see why it does not translate the string correctly in your code though, `translation` objects only stores the raw string so calling `transation::translated` should return the translated text when called right before the `query_yn` call.

dseguin: Oh, it looks like `read()` was the key there. Followed your advice using a dedicated `translation_or_var` object, and everything seems to be working :)

src/dialogue_helpers.h
```diff
@@ -147,13 +148,24 @@ std::string read_var_value( const var_info &info, const dialogue &d );
...
+    }
+};
+
+struct str_or_var : abstract_str_or_var<std::string> {
+    std::string evaluate( dialogue const &d ) const override;
+};
+
+struct translation_or_var : abstract_str_or_var<translation> {
+    std::string evaluate( dialogue const &d ) const override;
 };
```
Qrox, line 169: AFAICT `str_or_val` and `translation_or_val` are not used dynamically so you can probably define this class statically and define `str_or_var` and `translation_or_var` to be specializations of the template, like this:

```suggestion
    std::string evaluate( dialogue const &d ) const;
};

using str_or_var = abstract_str_or_var<std::string>;
using translation_or_var = abstract_str_or_var<translation>;
```

src/condition.cpp
```diff
@@ -243,6 +243,30 @@ str_or_var get_str_or_var( const JsonValue &jv, const std::string &member, bool
...
+{
+    translation_or_var ret_val;
+    translation str_val;
+    if( jv.read( str_val ) ) {
+        ret_val.str_val = str_val;
+    } else if( jv.test_object() ) {
+        const JsonObject &jo = jv.get_object();
+        if( jo.has_member( "mutator" ) ) {
+            // if we have a mutator then process that here.
+            ret_val.function = conditional_t::get_get_string( jo );
```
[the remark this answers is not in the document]
andrei8l: All the current mutators read `type_id`s or tripoints which are not translatable so string is fine.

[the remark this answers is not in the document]
Qrox: I think `translation_or_val` should not contain a `function` value then, which can be implemented later if needed, because if it's ever going to be used it should be translatable.

lang/string_extractor/parsers/condition.py
```diff
@@ -0,0 +1,20 @@
+from ..write_text import write_text
+
+
+def parse_condition(conditions, origin):
+    if not type(conditions) is list:
+        conditions = [conditions]
+    for cond in conditions:
+        if type(cond) is dict:
+            if "u_query" in cond:
+                write_text(cond["u_query"], origin,
```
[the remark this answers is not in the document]
dseguin: Cheers. I just added a string check because anything from an object currently will be pulling a message from elsewhere

src/condition.cpp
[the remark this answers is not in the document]
dseguin: It turns out that there are values of "mutator" that return translated strings, so I went ahead and split `get_get_string` into translatable and non-translatable components, so that `get_get_translation` can get the translation object from these cases.

Tested with:
```json
"u_query": { "mutator": "ma_technique_name", "matec_id": "WBLOCK_1" }
```
In French, it'll use "bloquer" in the query message.

src/condition.cpp
```diff
@@ -1504,40 +1530,89 @@ std::function<std::string( const dialogue & )> conditional_t::get_get_string( co
...
-        return "INVALID";
-    };
+    return nullptr;
+}
+
+std::function<translation( const dialogue & )> conditional_t::get_get_translation(
+    const JsonObject &jo )
+{
+    auto ret_func = get_get_str_<translation>( jo, []( const std::string & s ) {
+        return to_translation( s );
```
Qrox, line 1579:
```suggestion
        return no_translation( s );
```

Qrox (commented):
Thanks!

PR #67079, opened 2023-07-21
fix: reduce charge in trash bound items

casswedson:
#### Summary
None

#### Purpose of change
found some dolls with full charge in the trash, reduce their battery charge for inmmersion


anoobindisguise:
I think the creepy doll should have 99 charges. 'Cause they used it once and then threw it away. It would make it too obvious which is which maybe but it would be some fun worldbuilding

PR #67084, opened 2023-07-21
EMT, Paramedic, and Paramedic Firefighter adjustment

ghost:
#### Summary
None

#### Purpose of change

Input from an EMT working in NE suggested that our professions for EMT, Paramedic, and EMT firefighter were a touch out of step; an EMT is probably an EMT-b, which has less qualifications than an EMT-Paramedic, who are in turn the only ones allowed to be firefighters.

#### Describe the solution

-EMT firefighter is now Paramedic Firefighter
-EMTs have a slightly lower first aid skill, and don't have advanced wound care, or human dissection but do have an epipen and some asprin.
-Paramedic firefighters have a single shot of morphine and an epipen.
-Paramedics have had epipens added to their kit.


data/json/professions.json
```diff
@@ -2704,7 +2711,10 @@
           "mbag",
           "nomex_socks",
           "water_clean",
-          "pocketwatch"
+          "pocketwatch",
```
Maleclypse, line 2714: Why would I firefighter carry `"id": "pocketwatch",
    "type": "TOOL",
    "name": { "str": "pocket watch", "str_pl": "pocket watches" },
    "description": "An old-fashioned pocket watch.  Tells the time and looks good doing it.  Can also be disassembled into some useful parts.",`?

Maleclypse, line 2714:
```suggestion
          "wristwatch",
```

Maleclypse:
Decided to be the change I wished to see in the world.

chichit1044:
yeah that is genuinely baffling to me on discovery. what _possible_ use would a firefighter have for a pocketwatch? it is a net loss of safety and usefulness from having no watch at all. the chain would get caught on things!

data/json/professions.json
```diff
@@ -5231,8 +5240,7 @@
           { "item": "bandages", "count": 3 },
           "stethoscope",
           "scissors",
-          "syringe",
-          "morphine"
+          "adrenaline_injector"
```
Procyonae, line 5243: They did explicitly say the Paramedic having morphine is chill if you wanted to add that back on top of their epipen https://discord.com/channels/598523535169945603/598523535169945607/1131880091996147712

chichit1044: oh dang, I think I axed that because I started adding med supplies before noticing the medkit was there. back in it goes!

PR #67085, opened 2023-07-21
Split kevlar into two types (stab and ballistic)

### LINKED ISSUE #66762: stab weave kevlar vs ballistic weave kevlar

IRL there are like 50 billion versions of kevlar to choose from with specialized purposes but there are two main distinctions - ballistic weave (extremely tight machine precise weave) and stab weave (loose weave). Both kinds are rather poor at their opposite - stab weave doesn't do much vs bullets and ballistic weave doesn't do much vs knives.

In game however there's only one kind of kevlar, which appears to be a ballistic weave based on its stats (2.4 stab and 5 ballistic).

### Solution you would like.

Kevlar could probably be split into two different versions, with ballistic weave for example being largely the same or having slightly lower cut/stab while stab weave kevlar could have the inverse, 5 stab prot and only like 2 ballistic prot. Many pieces of gear could be converted to this - the robofac MDS, cut resistant gear, dog bite suits, stab vests, etc. While ballistic vests would remain ballistic kevlar and be the primary source of that resource.

Continue wearing bulletproof dog bite suit to gunfights.

_No response_

### Additional context

_No response_

anoobindisguise:
#### Summary
Balance "Stab weave and ballistic weave kevlar are now distinct"

#### Purpose of change
Fixes #66762
Previously there was only one type of Kevlar (ballistic). The only distinction was whether it was 4.4mm thick or not. This isn't how Kevlar works in real life, and it led to tons of stuff being made out of ballistic Kevlar when it made no sense, such as the dog bite suit or cut resistant sleeves. 

#### Describe the solution
Layered Kevlar is now "Ballistic Kevlar". Almost every single instance of ballistic gear in game already used Layered Kevlar, so it was an easy conversion to make. Ballistic Kevlar panels cannot be crafted anymore, only salvaged from ballistic armor. Repairs are still done with brand new panels since once a panel absorbs a bullet it very quickly loses its capacity to absorb more IRL, the entire weave is ruined in order to stop the bullet. This doesn't make it degrade any faster as it probably should though. The Kevlar Vest (the underarmor one) was converted to Ballistic Kevlar and it is the only Ballistic Kevlar item aside from some old survivor stuff to not be 4.4mm thick, but I feel this is an acceptable trade-off for simplicity's sake (rather than introducing an otherwise incompatible material for only one item). Ballistic Kevlar has had its bash protection lowered to 1.5 and cut prot lowered to 2. Ballistic vests IRL do very little to stop knives, the tight weave is good against bullets only. It still retains a bashing advantage over standard kevlar because of the tighter weave being more rigid and the thickness acting as a force multiplier on bashing protection, but this means ballistic vests are no longer amazing vs stabbings and thrown rocks (they went from 9.8 bash, 13 cut, 10.4 stab, 25 ballistic, to 7 bash, 9 cut, 7.5 stab and 25 ballistic)

Standard kevlar now acts as a loose woven "stab" variety, and again it cannot be turned into ballistic panels. Basically every single instance of regular Kevlar made more sense as a "stab" weave rather than a ballistic weave, so it made sense to do this as a sweeping change. This form of Kevlar is fairly flexible to work with, and it now salvages into patches, which can be turned into patchwork sheets (unlike for ballistic Kevlar, the way the loose weave stops knives IRL does not require long single lengths of thread to stretch, and it's ok to have little individual surface areas patchworked together). These sheets can also be woven due to a relatively looser tolerance and a lack of need to get a machine-precise uniformly extremely tight weave. This Kevlar has 1 bash, 3 cut, 5 stab and 2 ballistic, being the first material to use a stab protection value that isn't automatically 0.8x the cut prot (thanks to #67038 for making that possible!)

Not doing this.
Making the kevlar undervest work differently (maybe NO_SALVAGE to represent its odd panel thickness?)
Give the "stab vest" a thermoplastic resin insert as that's very common.

#### Testing
Loaded in game and didn't have errors, but when I did so the fix for stab protection not working right wasn't in yet. It is now though so everything should be good.


fairyarmadillo:
I think there should be a way to uncraft ballistic kevlar and get a new type of item - ballistic kevlar sheet, which could be sewn into the currently existing panels. AFAIK there's nothing special about how these panels are assembled, the layers of fabric are just laid flat on top of each other in many layers and sewn together like you would with an aketon.

The use cases for this would be finding bolts of ballistic kevlar cloth to make things out of, which should be rare but the stuff does exist, and uncrafting damaged body armor to salvage sheets from a compromised panel. I understand if that's out of scope for now.

data/json/items/resources/tailoring.json
```diff
@@ -392,6 +393,37 @@
...
+    "looks_like": "sheet_kevlar",
+    "symbol": "=",
+    "color": "dark_gray"
+  },
+  {
+    "id": "kevlar_patch",
+    "type": "GENERIC",
+    "category": "spare_parts",
+    "name": { "str": "Kevlar patch", "str_pl": "Kevlar patches" },
+    "description": "A 6-inch square patch of loose-woven kevlar, suitable for repairing cut and stab resistant clothing or reassembling into sheet for sewing.",
```
TheShadowFerret, line 417:
```suggestion
    "description": "A 6-inch square patch of loose-woven Kevlar, suitable for repairing cut and stab resistant clothing or reassembling into sheet for sewing.",
```
*Angry DuPont noises*

PR #67086, opened 2023-07-22
allow RM13/Phase Immersion to be used as PPE

anoobindisguise:
#### Summary
Balance "hypertech armors work as PPE"

#### Purpose of change
RM13 and Phase Immersion armors were more than suited for PPE work, offering immunity to static discharge, acid, gas, radiation, etc. But in game they could not be used as actual lab PPE.

#### Describe the solution
Allow them to be used as PPE.

### DIFF
--- a/data/json/requirements/toolsets.json
+++ b/data/json/requirements/toolsets.json
@@ -421,7 +421,18 @@
     "id": "ppe_gloves",
     "type": "requirement",
     "//": "Disposable or reusable gloves.",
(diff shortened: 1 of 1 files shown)


John-Candlebury:
Even though it makes sense isnt this a bit of a weird spoiler?

anoobindisguise:
> Even though it makes sense isnt this a bit of a weird spoiler?

You're right, it really is. Not sure how to do that, maybe create a PPE quality and hand it out?

John-Candlebury:
> > Even though it makes sense isnt this a bit of a weird spoiler?
> 
> You're right, it really is. Not sure how to do that, maybe create a PPE quality and hand it out?

Welding googles already work like that IRC so yeah I think that could fit.

PR #67089, opened 2023-07-22
Robofac robots and gear

John-Candlebury:
#### Summary
Content "Hub 01 robots and gear"

#### Purpose of change
Expand the list of robots Hub 01 has for sale and more importantly add a somewhat convenient way to carry them around.

#### Describe the solution
The list of new content and changes is:

-Drone-tech harness: A modified army harness that can store one skitterbot sized robot and four quadcopters for little encumbrance. Has a small regular storage capacity and a pair of tool clips in addition to that. For sale after Iron Safari
-Hub 01 Exterminatron: A skitterbot mounting a semi-auto 7.62 "rifle". For sale after Iron Safari.
-Hub sells tazerhacks.
-Adjust skitterbot weight and volume stats so they are roughly Spot sized.

#### Testing
Loaded and used the gear.


data/json/flags.json
```diff
@@ -2141,6 +2141,16 @@
     "type": "json_flag",
     "info": "This is meant for exosuit foot plating."
   },
+  {
```
Karol1223, line 2144: Am I missing something here or are those just used to group the robots into the pocket whitelist? If so... why bother with flags and not just manually define a whitelist for those pockets with specific item IDs?

John-Candlebury: Np. Mostly because I intend to end up with 15 sellable robots or so, and dont want to copy paste a list of that size around.

PR #67093, opened 2023-07-22
flags_any Helper Check To Make Road Unhardcoding More Readable

Procyonae:
#### Summary
None

#### Purpose of change

While I'm able to hack this already with a combination of #66873 and nests for every individual direction to check, there are multiple instances where I need to just check if there is at least one cardinally adjacent SIDEWALK which currently needs 8 nests that just do "if there's SIDEWALK this way do the thing, else carry on checking" that passes to the next nest that tests the next direction and so on, which makes it both seem more intimidating than it needs to for future contributors while also making it less legible and more lines for reviewers

#### Describe the solution

Just adds a flags_any field that works identically to flags except it only needs one of the given directions to have the flags specified, can be used alongside the existing checks
Adds documentation for it
The new check is purposefully separate from flags so you can combine the two until a more complete solution is in place

I'd rather add a more comprehensive syntax for all possible logic with the checks but I can't come up with anything sensible, if someone can think of something now or in the future please replace this

#### Testing

Checked it works in game by trying to place a random 24x24 nest over a map with

```
"place_nested": [
  { "chunks": [ "24x24_fence1" ], "x": 0, "y": 0, "flags_any": { "north": [ "SIDEWALK" ], "east": [ "SIDEWALK" ], "south": [ "SIDEWALK" ], "west": [ "SIDEWALK" ] } }
],
```

### DIFF
--- a/src/mapgen.cpp
+++ b/src/mapgen.cpp
@@ -3458,7 +3458,7 @@ class jmapgen_nested : public jmapgen_piece
                                     JsonObject jo = entry.get_object();
                                     dir_neighbor.first = jo.get_string( "om_terrain" );
                                     dir_neighbor.second = jo.get_enum_value<ot_match_type>( "om_terrain_match_type",
-                                                           ot_match_type::contains );
+                                                          ot_match_type::contains );
                                 }
                                 dir_neighbors.insert( dir_neighbor );
                             }
--- a/src/mapgen.cpp
+++ b/src/mapgen.cpp
@@ -3571,6 +3571,41 @@ class jmapgen_nested : public jmapgen_piece
                     return true;
                 }
         };
+        
+        // TO DO: Remove this and replace it with a smarter logical syntax for neighbor_join_check and neighbor_flag_check
+        class neighbor_flag_any_check
+        {
+            private:
+                std::unordered_map<direction, cata::flat_set<oter_flags>> neighbors;
+            public:
+                explicit neighbor_flag_any_check( const JsonObject &jsi ) {
+                    for( direction dir : all_enum_values<direction>() ) {
+                        cata::flat_set<oter_flags> dir_neighbours;
+                        std::string location = io::enum_to_string( dir );
+                        optional( jsi, false, location, dir_neighbours );
+                        if( !dir_neighbours.empty() ) { 
+                            neighbors[dir] = std::move( dir_neighbours );
+                        }
+                    }
+                }
+
+                bool test( const mapgendata &dat ) const {
+                    bool any_direction_has_flag = false;
+                    for( const std::pair<const direction, cata::flat_set<oter_flags>> &p :
+                         neighbors ) {
+                        const direction dir = p.first;
+                        const cata::flat_set<oter_flags> &allowed_flags = p.second;
+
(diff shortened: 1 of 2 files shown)


src/mapgen.cpp
```diff
@@ -3571,17 +3571,53 @@ class jmapgen_nested : public jmapgen_piece
...
         };
+        // TO DO: Remove this and replace it with a smarter logical syntax for neighbor_join_check and neighbor_flag_check
+        class neighbor_flag_any_check
+        {
+            private:
+                std::unordered_map<direction, cata::flat_set<oter_flags>> neighbors;
+            public:
+                explicit neighbor_flag_any_check( const JsonObject &jsi ) {
+                    for( direction dir : all_enum_values<direction>() ) {
+                        cata::flat_set<oter_flags> dir_neighbours;
```
Numbuh474, line 3582: Nitpick: use consistent spelling of neighbo(u)rs.

src/mapgen.cpp
```diff
@@ -3621,7 +3657,8 @@ class jmapgen_nested : public jmapgen_piece
         }
         const weighted_int_list<mapgen_value<nested_mapgen_id>> &get_entries(
         const mapgendata &dat ) const {
-            if( neighbor_oters.test( dat ) && neighbor_joins.test( dat ) && neighbor_flags.test( dat ) ) {
+            if( neighbor_oters.test( dat ) && neighbor_joins.test( dat ) && neighbor_flags.test( dat ) &&
+                neighbor_flags_any.test( dat ) ) {
```
Numbuh474, line 3661: `neighbor_flags && neighbor_flags_any` is redundant: `x AND (x OR y)` simplifies to `x AND x `which simplifies to `x`.

src/mapgen.cpp
```diff
@@ -3571,17 +3571,53 @@ class jmapgen_nested : public jmapgen_piece
...
+                        const cata::flat_set<oter_flags> &allowed_flags = p.second;
+
+                        cata_assert( !allowed_flags.empty() );
+
+                        for( const oter_flags &allowed_flag : allowed_flags ) {
+                            any_direction_has_flag |=
+                                dat.neighbor_at( dir )->has_flag( allowed_flag );
+                        }
+                    }
+                    return any_direction_has_flag;
```
Numbuh474, line 3605: Optional: you could return true if `dat.neighbor_at( dir )->has_flag( allowed_flag )` to skip the rest of the nested for loop iterations.

src/mapgen.cpp
Procyonae: The bane of trying to abide by Americanisms when I'm not American

src/mapgen.cpp
Procyonae: This isn't redundant they're entirely separate checks that can contain different flags for different directions and be combined eg `x AND (y or z)`

Numbuh474: Apologies, I must have misunderstood the change intent.

Procyonae: You can look at nest4's placement in the documentation I changed for an example where it needs to be &&

PR #67094, opened 2023-07-22
Non f_sign Signage

Procyonae:
#### Summary
None

#### Purpose of change

Only being able to use f_sign for signage is silly

#### Describe the solution

signage now takes an optional field furniture which places that furniture instead of a sign, the furniture needs the existing SIGN flag and the sign examine action
SIGN_ALWAYS is a new flag that provides the messages for blank signs and adding to signs that already exist, didn't make sense for everything you could mark with this to inform you of this imo
Replaces the silo's f_sign overwriting the f_sign_warning with the new syntax

Having the flag work the other way around (so SOMETIMES_SIGN or something)

#### Testing

Tested using statues with and without signage to check the iexamine changes as well as checking it defaults to f_sign.
The current rewriting sign code L4567 of iexamine.cpp doesn't seem to work with or without this change, could do with someone checking I'm not missing something and if not I'll open an issue for it


doc/JSON_FLAGS.md
```diff
@@ -701,6 +701,7 @@ List of known flags, used in both `terrain.json` and `furniture.json`.
 - ```SHARP``` May do minor damage to players/monsters passing through it.
 - ```SHORT``` Feature too short to collide with vehicle protrusions. (mirrors, blades).
 - ```SIGN``` Show written message on examine.
+- ```SIGN_ALWAYS``` Shows a message to indicate nothing is written here and let's you add a message if examined without a signage/snippet present.
```
Night-Pryanik, line 704:
```suggestion
- ```SIGN_ALWAYS``` Shows a message to indicate nothing is written here and lets you add a message if examined without a signage/snippet present.
```

PR #67096, opened 2023-07-22
[Xedra Evolved] Low Intelligence Mana Perk

Maleclypse:
#### Summary
None

#### Purpose of change
Adds a perk that grants additional Mana to characters with intelligence below 14 capping at 6.  This should allow lower int characters to be viable casters.

#### Describe the solution
Adds perk trait using formula from andrei.

Might also make it a character creation trait but that seems weird.

#### Testing
Works like a charm

### DIFF
--- a/data/mods/Xedra_Evolved/perks/perks.json
+++ b/data/mods/Xedra_Evolved/perks/perks.json
@@ -58,5 +58,14 @@
     "category": [ "perk" ],
     "active": true,
     "activated_eocs": [ "EOC_ALCHEMY5" ]
+  },
+  {
+    "type": "mutation",
+    "id": "perk_LOW_INT_MANA",
+    "name": { "str": "Mana for Morons" },
+    "points": 0,
+    "description": "Maybe you aren't a genius, but something inside of you grants previously hidden reserves of mana.",
+    "category": [ "perk" ],
+    "enchantments": [ { "condition": "ALWAYS", "values": [ { "math": [ "u_val('mana')", "=", "( 16 - clamp(u_val('intelligence'), 8, 16) ) * 62.5" ] } ]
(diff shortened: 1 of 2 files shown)


data/mods/Xedra_Evolved/perks/perks.json
```diff
@@ -58,5 +58,14 @@
...
     "activated_eocs": [ "EOC_ALCHEMY5" ]
+  },
+  {
+    "type": "mutation",
+    "id": "perk_LOW_INT_MANA",
+    "name": { "str": "Mana for Morons" },
+    "points": 0,
+    "description": "Maybe you aren't a genius, but something inside of you grants previously hidden reserves of mana.",
+    "category": [ "perk" ],
+    "enchantments": [ { "condition": "ALWAYS", "values": [ { "math": [ "u_val('mana')", "=", "( 14 - clamp(u_val('intelligence'), 6, 14) ) * 62.5" ] } ] } ]
```
GuardianDll, line 69:
```suggestion
    "enchantments": [
      {
        "condition": "ALWAYS",
        "values": [ { "value": "MAX_MANA", "add": { "math": [ "( 16 - clamp(u_val('intelligence'), 8, 16) ) * 62.5" ] } } ]
      }
    ]
```

Maleclypse:
@TheShadowFerret This is part of my long term desire to move away from mana as the power source for Xedra Evolved spells. My ideal setup would involve not having the mana substitute pool being entirely governed by intelligence.  Guardian had an idea with each class having their own mana pool derived from it's own stat.  Either way mana isn't jsonized yet but I figured I'd explain some of the ideas behind this so if it's ridiculous I can hear some counter arguments.

PR #67103, opened 2023-07-22
Big Blacksmithing Bonanza

Karol1223:
#### Summary
Balance "Rebalance most of our blacksmithing recipes"

#### Purpose of change
It started by me noticing we can use kitchen tongs in blacksmithing and spiraled out of control. So apparently none of our recipes followed the requirement json note with how much energy it should take per chunk. Yeah it's fun like that.

#### Describe the solution
- **Rebalanced our power costs for blacksmithing recipes (that use steel) to go in hand with the JSON comment** - NOTE: THIS DOES NOT MEAN THE COSTS ARE ACCURATE TO THE REAL WORLD! I am not a blacksmith and I do NOT know if the power outlined by the requirement is accurate - but I know that if someone was to double check that after my changes they'd only have to alter two requirement groups instead of *checks notes* 768 json hits
- **Removed cooking tongs from as many blacksmithing, redsmithing, and glassworking recipes as I could find** - Cooking tongs are fun but not designed for this. They maybe could work for redsmithing but their shape wouldn't help a lot so nah
- **Removed excess mentions of tool qualities handled by the requirements** - as it says on the lid. No more weapons that need ANVIL 3 twice
- **Added a new requirement group rebalanced to use proportionally little power to one scrap instead of one chunk**
- **Swapped a lot of recipes from ``forging_standard`` to ``blacksmithing_standard/scrap``** - a lot of our forging recipes had manually added tools/qualities from ``blacksmithing_standard``. There are some edge cases where this will make recipes harder, but for the most part, all it does is it means the kitchen tongs can't be used and that you need a metal fileset.
- **Renamed metal tongs to kitchen tongs to avoid confusion in the future**
- **Added both kitchen tongs and flatjaw tongs to the modular vehicle/appliance rigs**

**So what's the TLDR? Or otherwise: notes for players**
- Kitchen tongs can no longer be used in blacksmithing. Get flatjaw tongs
- A lot of recipes had their power costs rebalanced BUT NOT ALL TO BE HIGHER! There's been a lot of recipes that used too little power, but almost equally as many that used too much power. Look around and find out
- Since many recipes that used to use forging (with attached tool qualities...) now use blacksmithing instead - GET A METAL FILESET! You'll need it for a lot more nowadays

- If I could do it I'd change the ID for kitchen tongs because as it stands, their ingame name is dangerously close to flatjaw's ID so I can see how they could get confused by a contributor. Alas, no can do...
- I considered making kitchen tongs unusable for plasticworking and the crafting of carbon rods, though I simply don't know if they'd be reasonably usable for that purpose or not so I opted to leave these recipes alone

#### Additional context
I did not rebalance costs for stuff that used copper, bronze, gold, silver or items that weren't clearly raw steel resources. Not everything here is perfect because I'm not capable enough to recalculate everything. I did as much as I could.

ALSO! FUN FACT: One of the power cell recipes had its power cost increased by 10000% in this PR. It's honestly quite incredible.

### DIFF
--- a/data/json/items/tool/metalworking.json
+++ b/data/json/items/tool/metalworking.json
@@ -349,7 +349,7 @@
   {
     "id": "tongs",
     "type": "TOOL",
-    "name": { "str": "pair of metal tongs", "str_pl": "pairs of metal tongs" },
+    "name": { "str": "pair of kitchen tongs", "str_pl": "pairs of kitchen tongs" },
     "description": "A pair of long, metal tongs.  They are commonly used for cooking.",
     "weight": "167 g",
     "volume": "210 ml",
--- a/data/json/recipes/ammo/weldgas.json
+++ b/data/json/recipes/ammo/weldgas.json
@@ -72,7 +72,7 @@
     "autolearn": true,
     "qualities": [ { "id": "HAMMER", "level": 2 }, { "id": "ANVIL", "level": 1 }, { "id": "SAW_M", "level": 2 } ],
     "tools": [
-      [ [ "tongs", -1 ], [ "metalworking_tongs", -1 ] ],
+      [ [ "metalworking_tongs", -1 ] ],
       [ [ "crucible", -1 ], [ "crucible_clay", -1 ] ],
       [ [ "forge", 30 ] ],
       [ [ "rock_quern", -1 ], [ "clay_quern", -1 ], [ "mortar_pestle", -1 ], [ "food_processor", 10 ] ]
@@ -112,7 +112,7 @@
       { "id": "CHEM", "level": 1 }
     ],
     "tools": [
-      [ [ "tongs", -1 ], [ "metalworking_tongs", -1 ] ],
+      [ [ "metalworking_tongs", -1 ] ],
       [ [ "crucible", -1 ], [ "crucible_clay", -1 ] ],
       [ [ "forge", 350 ] ],
       [ [ "rock_quern", -1 ], [ "clay_quern", -1 ], [ "mortar_pestle", -1 ], [ "food_processor", 25 ] ]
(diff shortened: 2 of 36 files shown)


PatrikLundell:
Wouldn't it be possible to address the kitchen tong id by creating a new item with the more appropriate name and obsolete the old one (obviously replacing all item groups etc. entries of the old one with the new one)? It would also require all recipes that need it to list both (and updating of those when the old item is eventually removed). I don't know it it's possible to add id replacement to JSON loading, but it might be possible (in which case the old id might not be needed in recipes). If so, it should probably be its own PR.

Karol1223:
> Wouldn't it be possible to address the kitchen tong id by creating a new item with the more appropriate name and obsolete the old one (obviously replacing all item groups etc. entries of the old one with the new one)? It would also require all recipes that need it to list both (and updating of those when the old item is eventually removed). I don't know it it's possible to add id replacement to JSON loading, but it might be possible (in which case the old id might not be needed in recipes). If so, it should probably be its own PR.

It would theoretically be possible to dynamically migrate them into flatjaws, yes, but there's some problems with this:
1. Kitchen tongs are still properly used in many cooking recipes. Migrating them would mean all of them in the current saves would turn into metalworking tongs unusable in those recipes.
2. If the recipes still weren't manually edited, they'd call for flatjaw tongs twice because they are also present in the ``blacksmithing_standard`` requirement group
All in all it's simply easier to go through it the way I did, as weird as it sounds...

PatrikLundell:
What I tried to suggest was actually to migrate them into kitchen tongs, so they'd be removed from the blacksmithing recipes and remain in the cooking ones. However, that would cause some existing save blacksmithing capability setups to suddenly fail to work as one critical tool was suddenly transformed into being usable in the kitchen only.
I don't have any problem with you finding such a change too unwieldy or unsuitable, though.

Karol1223:
> that would cause some existing save blacksmithing capability setups to suddenly fail to work as one critical tool was suddenly transformed into being usable in the kitchen only

That's still gonna be the case with or without migrating after this PR, unfortunately, so it's nothing that should be a blocker. I'll consider your idea because it'll likely be easier in the long run (if merge people approve this).

Karol1223:
@Drew4484 I'd appreciate it if you could look at this to make sure I didn't break some of your work horribly

Drew4484:
> @Drew4484 I'd appreciate it if you could look at this to make sure I didn't break some of your work horribly

Looks great! There are changes in this that will make future work a lot easier and overall it drags metalworking in a more realistic direction which I'm all for. Even if it doesn't touch non-steel it sets some framing for other metalwork.

PR #67107, opened 2023-07-23
Clean Conscience trait

ghost:
#### Summary
None

#### Purpose of change

The word "psychopath" has a lot of unpleasant baggage in mental health, similar to the old trait name "schizophrenia". More than that, there are a range of reasons why someone might not take a morale penalty from murder (frontline soldier with multiple hot combat deployments, etc; any situation in which someone has already achieved moral parity with ending a sapient life).

For now, this is just a minor fluff adjustment to make it gel better. 

#### Describe the solution

Changed Psychopath trait name to Clean Conscience and edited the description to reflect a more neutral view of it.

#### Additional context

I also believe that being okay with cannibalism should be its own trait line, rather than one linked to the Psychopath tree. And it would be useful to have a way to acquire it after game start. If this direction is approved, I'll figure out how.


data/json/mutations/mutations.json
masema, line 1: While I agree this change needs done, I don't think the changes here are the best option. The word Callous doesn't quite give the impression of what the trait does and the description only describes part of the trait. 

My suggestion would be to leave the description alone and change the name to either "Guiltless" or "No Conscience"

chichit1044: on it.

Venera3:
If it's just `Get out of morale penalties free` you also need to scrub all the weird edgy interactions with e.g. cannibal/spiritualist etc.

The way to aquire cannibalism post-game is getting the Predator line of traits - that or adding a proper acclimation/burnout/trauma system that can do coping well.

chichit1044:
> you also need to scrub all the weird edgy interactions with e.g. cannibal/spiritualist etc.

I definitely was considering removing the cannibalism tag, but didn't want to do that without something to replace it. Making it come up after game start via EoCs that keep track of how many living humans you've killed is the only way I can think of to simulate a "toughening up" mechanic.

Venera3:
>Making it come up after game start via EoCs that keep track of how many living humans you've killed is the only way I can think of to simulate a "toughening up" mechanic.

That's not how this works, and that won't get accepted. I really urge you to check out the past issues/PRs regarding this stuff.

fairyarmadillo:
I support changing the name but I think clean conscience is the wrong way to go, as it doesn't carry the negative connotations this trait is meant to imply. Maybe "detached" or something?

Daved27hundred:
"Apathetic" would be a better word.

chichit1044:
The fundamental problem is, I don't think that being okay with killing other sapient beings if the circumstances demand it should come with a sense of in-game moral condemnation. If you have internalized the idea that sometimes it's you or the other guy, that doesn't make you "apathetic" to the idea of murder. It means that you can do it necessary and not become completely non-functional for several hours afterwards.

But, as this is a personal opinion that doesn't reflect larger design goals, I'll keep this closed.

PR #67120, opened 2023-07-23
Add some dialog function

lispcoc:
#### Summary
Features "Add some dialog function"

#### Purpose of change
We can not use max HP value in dialog function. Also we can not use topic_item id outside dynamic line.

#### Describe the solution
Add variable and mutator.

**math function "u_hp_max('bodypart')"**
Make getting max HP easily.

**mutator "topic_item"**
Make it  enable to store "topic_item" into variable. See [Repeat Responses](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/NPCs.md#repeat-responses)

#### Testing
mutator test case passed.
I can get max HP in dialog function.


src/condition.cpp
```diff
@@ -1567,6 +1567,10 @@ static std::function<T( const dialogue & )> get_get_translation_( const JsonObje
         return [ma, ret_func]( const dialogue & d ) {
             return ret_func( matec_id( ma.evaluate( d ) )->name );
         };
+    } else if( jo.get_string( "mutator" ) == "topic_item" ) {
+        return []( const dialogue & d ) {
+            return d.cur_item.str();
```
bombasticSlacks, line 1572: Do you want to return the ID of the item? or it's in game name? If you want the in game name you need to do something like "cur_item->nname(1);" and if you want the ID this should be moved above to get_str from get_translation

src/condition.cpp
```diff
@@ -1785,6 +1789,15 @@ std::function<double( dialogue & )> conditional_t::get_get_dbl( J const &jo )
                 bodypart_id bid = bp.value_or( get_bp_from_str( d.reason ) );
                 return d.actor( is_npc )->get_cur_hp( bid );
             };
+        } else if( checked_value == "hp_max" ) {
```
bombasticSlacks, line 1792: This is the legacy way to add accesors, you should instead place it in a math function in: math_parser_diag.h there are similar examples in there for stuff, armor has examples of getting the BP.

src/condition.cpp
lispcoc: I want ID, THX!

src/condition.cpp
lispcoc: it's moved  to math_parser_diag

PR #67137, opened 2023-07-24
Hair variants fix

Jesper-Johansson:
#### Summary
Bugfixes "Made old hair colors into mutation variants instead of being their own, separate mutations."

#### Purpose of change

Hairstyles were defined in two separate json files; `mutation_appearance.json` and `appearance_hair_styles.json`. Recent additions have gone into the latter file, with the benefit of every hair color becoming variants of the same mutation as opposed to becoming their own mutation.

Having the hair colors be variants of a mutation is good because it lets us use [`pick_variant()`](https://github.com/CleverRaven/Cataclysm-DDA/blob/e5a74fb42b8da530d4d00456bc3e4970942231af/src/mutation_data.cpp#L831) and [`pick_variant_menu()` ](https://github.com/CleverRaven/Cataclysm-DDA/blob/e5a74fb42b8da530d4d00456bc3e4970942231af/src/mutation_data.cpp#L856)in `mutation_data.cpp`.
This is currently used when the player opens up the @ menu and hovers over traits like eye/hair color, then presses Shift + S, which opens up a menu that lets them change colors. This feature was not working for the old hair colors in mutation_appearance.json (they didn't appear in the change hair color menu, and if your character started with an old hair color the game wouldn't let you change it at all). This PR makes them work, which means the functions can also be used in more places (like the create custom character menu, which was the thing I was originally trying to improve before I stumbled upon this problem).

#### Describe the solution

Remove the old hairstyles from mutation.appearance.json and put them into appearance_hair_styles.json instead, matching that file's format. Update references to their old IDs in `appearance_trait_groups.json` and delete references to their old IDs in `mutation_ordering.json` (since they're no longer separate mutations that need to be tracked).

I also manually edited the `tile_config.json` files of a couple of popular tilesets with the new IDs of the moved hairstyles so I could test my changes. The tilesets were UltiCa, MshockXotto+, and Chibi_Ultica. The files are included in this PR, but their compositing tilesets will still need to be updated to reflect these changes, otherwise they'll spit out faulty tile_conifgs again the next time they're updated. The same goes for all tilesets. I would gladly help with this.

Changing the source code to account for both ways of implementing hair colors, but it's better to improve the foundation instead of finding workarounds like that.

#### Testing

**Playtesting:** Making characters, selecting various options in the character creation to make sure they work, using in-game keybinds to switch between different colors to make sure they display correctly and that all variants are selectable, also switching between both genders.

**Logging:** I made use of how `set_traits()` in `newcharacter.cpp` filters out everything that's not a starting trait and put some write to file commands there, writing out all found variants in one file and all found mutations in another. The goal was to have hair color variants only appear in the first file and base hair styles appear in the other. 
Before my changes, the old hair colors appeared in the second log next to the base hair styles, and didn't appear in the found variants log. After my changes, the log files look as expected (I'll provide them on request).

#### Additional context

First PR! Go extra hard on me.


Procyonae:
Why close?

Jesper-Johansson:
Closed and made a new pull request because I messed up some commits and just wanted to start clean on a new branch

New PR is here https://github.com/CleverRaven/Cataclysm-DDA/pull/67141, have one commit I need to do before it's ready for re-review

PR #67139, opened 2023-07-24
Blacksmithing followup: crafting tools for blacksmithing

### LINKED ISSUE #67157: Impossible to start a blacksmithing career from scratch after the Bonanza update
### Describe the bug

After the Big Blacksmithing Bonanza update (67103), it became impossible to start blacksmithing without looting instruments. Before that, one could craft his own instruments one by one, getting a full set needed for the more complicated jobs. That meant that one could start crafting metal gear only by having access to metal and a forge. 

Now, some instruments needed in the recipes are only available by looting, since they need themselves as a tool for crafting. Maybe it is realistic this way and it's intended behavior, but then i ask myself how did blacksmithing as a profession be available in the real life, since they would've not been able to craft their own tools...

### Attach save file

N/A

### Steps to reproduce

1. Make any player and get enough fabrication to access blacksmithing recipes.
2. try to make a full set of tools for the most complicated crafts.

### Expected behavior

I expected it would be possible to either craft those tools without themselves as a requirement or to be able to use more mundane tools, craftable without overly complicated requirements.

### Screenshots

_No response_

Version a91199f

### Additional context

image

Karol1223:
#### Summary
None

#### Purpose of change
While I do consider #67103 a good PR, I overlooked some things that meant some of the blacksmithing tools you couldn't craft if you didn't already have one of them. Now, that's not bad for mainline - go out and loot the damn hardware stores instead of bushcrafting pliers you weirdos.

The problem arises with Innawood. As it stands, according to @Light-Wave the blacksmithing progression wasn't much broken due to Innawood using outdated recipes, but since the idea was to have Innawood as close to mainline as possible we came to the conclusion that the best way would be to include ways to get those tools crafted in mainline.

Should close #67157 when its done

#### Describe the solution
- [x] Substitute metal fileset for metal chisel in the metal fileset recipe
- [x] Add a bronze hammer that can be used as a substitute to the regular hammer but crafted more easily - ALSO can use nail punch instead of a drift, bronze is softer than steel so this is viable
- [x] Add a new recipe to cast flatjaw tongs
- [x] Add a new ``forging_scrap`` requirement group to handle metalworking recipes slightly easier in the future
- [x] Make the old flatjaw tongs recipe use significantly less metal and be able to substitute the nail for a nut and bolt
- [x] Metalworking chisel now uses forging requirement and takes slightly less steel to make but is no longer reversible (why was it reversible and not a manual disassembly?)
- [x] Add new recipes to craft scrap metal and steel chunks in Innawood
- [x] Make the steel lumps from bog iron in Innawood no longer reversible

#### Additional context
- The bronze hammer almost certainly has the numbers wrong. They're placeholders that copy from and balance around our existing hammer. These numbers will be corrected as I rework the regular hammer to match.
- The metalworking chisel can no longer be currently disassembled. I intend to give it a manual disassembly recipe in my next generic item fix PR, since I believe this is not with the themeing of this PR


PatrikLundell:
#67157 concerns this very issue (but it seems to have been created after this PR, so it's not exactly a surprise this PR doesn't reference it).

Quoting my comment from that issue to reduce the need to look at it over there:

"I agree that there should be a way to progress blacksmithing from scratch. This can be achieved by replacing the requirement of a tool's own tool quality to make it with a requirement for a lower level of quality that then has to be provided by a simpler, possibly makeshift, tool (that obviously can be made by simpler tools still).
Another way to achieve this is to provide an additional recipe for the tool that requires "lesser" tools but requires considerably longer time. An example of an alternative tool is the polishing stone, which is a pain to make, but it provided a way to climb the tool ladder (I haven't checked whether this remains the case, of it it was broken as well).

Once the progression path has been restored, the workshop expansions need to be updated to provide the recipes to create all the tools (at least workshop expansion 2, which was tuned to provide such a path)."

Karol1223:
I intend to handle the production of basic tools as outlined here. Specifically the fileset and a hammer and tongs alternative. But in no situation should it be easier to craft those tools than to make them from scratch. I am less than not interested in catering to the dreams of restarting every production chain to exist from nothing.

When it comes to basecamp expansions I have no idea how their code words and I may not be able to do much there. We'll see how that goes, regardless that's out of scope for this PR.

PatrikLundell:
The workshop expansions work by having recipe groups that are made available by the production of successive facility constructions. Each construction step typically places tools in the workshop, so the process would be to make sure the tools placed are ones that can be produced by the recipes enabled by previous steps (which may also require the recipe groups to be expanded to contain new tools).
Regardless, I can probably do that once you've finished your work (provided, of course, that your less than none interest hasn't made the restoration task impossible).

Karol1223:
> provided, of course, that your less than none interest hasn't made the restoration task impossible

It shouldn't have. This only applies to mainline and I'm working on Innawood. Mainline is gonna get the results as a byproduct so it should still be workable. I still think people should just stop trying to play Minecraft and go out and loot the tools, though.

PatrikLundell:
Sometimes you're consistently unlucky with your looting, and sometimes you need to update your gear to be able to take on the hordes blocking your access to the loot (I'm not as affected as others in the latter regard, as I tune down the number of zombies from the silly level to a somewhat reasonable one).

Karol1223:
> Sometimes you're consistently unlucky with your looting, and sometimes you need to update your gear to be able to take on the hordes blocking your access to the loot (I'm not as affected as others in the latter regard, as I tune down the number of zombies from the silly level to a somewhat reasonable one).

I don't disagree, but I will always put realism above gameplay. It is more reasonable to loot the tools. Unless you're in a life or death scenario where you can't go out and loot but can, somehow, put together enough stuff to make your way from bronzeworking to making a better weapon (again, somehow crafting a better weapon yourself?) then this shouldn't apply in mainline.

These are scenarios that should realistically be extremely rare - mainly because realistically the odds that this should work are pretty low.

Karol1223:
@Light-Wave this should be ready to go now, I'd appreciate it if you looked through this and checked if this works for Innawood.

And if it does, lemme know if I should cut out the old recipes Innawood currently uses or not.

Light-Wave:
For `metal_file`, you can copy in your recipe into Innawood files [here](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/mods/innawood/recipes/tools_hand.json#L25C1-L41), but alter it so that it can be crafted from a single `steel_lump` as well, since there is no guarantee that you can get `scrap` in Innawood that early.

Innawood's recipe for `metalworking_tongs` (found just above) can be removed now.

The hammer could still be difficult to make since it needs a drift which requires hammering 3. Do you think it would be possible to make the hammer without that tool?

I will check out the rest of the recipes shortly and let you know if I spot any other problems.

Karol1223:
> The bronze hammer could still be difficult to make since it needs a drift which requires hammering 3. Do you think it would be possible to make the hammer without that tool?

I'll draft it until I figure that part out (probably gonna have to be tomorrow unfortunately) but the rest should be easy enough to do so I'll handle them right away.

Karol1223:
> [The metalworking chisel in tools_hand.json](https://github.com/CleverRaven/Cataclysm-DDA/blob/cdda-experimental-2023-07-25-0102/data/json/recipes/tools/tools_hand.json#L150) will need to be updated to accept `bronze_tongs`.

That actually... needs cleanup in mainline. I missed that one in the original blacksmithing PR, apparently... So now I gotta figure out how to make this recipe realistic without creating a vicious circle with the fileset again...

Dlightfull:
> > Sometimes you're consistently unlucky with your looting, and sometimes you need to update your gear to be able to take on the hordes blocking your access to the loot (I'm not as affected as others in the latter regard, as I tune down the number of zombies from the silly level to a somewhat reasonable one).
> 
> I don't disagree, but I will always put realism above gameplay. It is more reasonable to loot the tools. Unless you're in a life or death scenario where you can't go out and loot but can, somehow, put together enough stuff to make your way from bronzeworking to making a better weapon (again, somehow crafting a better weapon yourself? not to mention the time it'd take so "life or death scenarios" would likely still not apply here) then this shouldn't apply in mainline.
> 
> These are scenarios that should realistically be extremely rare - mainly because realistically the odds that this should work are pretty low.

Realistically speaking, some time ago in human history, somebody had to make the first flatjaw tongues without having one, so that should definitely be possible in game as well.

Light-Wave:
> Realistically speaking, some time ago in human history, somebody had to make the first flatjaw tongues without having one, so that should definitely be possible in game as well.

Yeah we're working on it. It should soon be possible to do so again. Although it will require bronze tools.

Karol1223:
> Realistically speaking, some time ago in human history, somebody had to make the first flatjaw tongues without having one, so that should definitely be possible in game as well.

That person wasn't a lone survivor melting down things in their garage. My opinion on how useful this is in mainline remains the same. If you want to challenge yourself by going through history be my guest, but it's, realistically, a more than suboptimal approach.

Dlightfull:
> That person wasn't a lone survivor melting down things in their garage. Getting from nothing to steel took centuries in history, mind you. My opinion on how useful this is in mainline remains the same. If you want to challenge yourself by going through history be my guest, but it's, realistically, a more than suboptimal approach. If you wanna go this way really just play Innawood. That's the reason we're working on this in the first place.
> 
> And I swear if I hear another argument that's just "actually people definitely made those tools without having them because history" I'll just close this fucking PR at that point. I've heard it too much today. Metallurgy isn't just bash rocks together until things work and I'm tired of hearing arguments that simplify it to just that. We're ALREADY dumbing down the process a lot. If we tried to take into account things like tongs for holding the crucible this entire thing falls apart at the core.
> 
> We can't even simulate this properly because we can't have tools get damaged as they're used. This is an uphill battle, don't make this less bearable than it already is.

I'm sorry if I sounded rude, I didn't mean to. I like realism, so if it's not possible for a lone survivor to do it, then I vote against the idea (I know my vote leans nothing, but whatever). 

I mean this lone survivor has access to more knowledge than his historic predecessors, so I thought he'd be able to pull this off. If that is not the case, then that's it. I'm by no means an expert in blacksmithing, so I'll better keep my mouth shut.

Karol1223:
Okay so after sleeping on this - first things first apologies for lashing out. I've just had a horrible day and the entire time discord (GitHub to much smaller extent comparable to that) has been bugging me about tools and blacksmithing and I've had to retrace the same conversation at least four times. I ran out of patience but me running out of patience shouldn't impact genuine concerns about crafting tools.

> I mean this lone survivor has access to more knowledge than his historic predecessors, so I thought he'd be able to pull this off. If that is not the case, then that's it.

You're not wrong with the knowledge part, and I don't disagree even remotely. The issue isn't that the survivor will never be able to craft their own tools, it's that people are very much set on making their own instead of looting. The issue at hand is that if (in mainline, for the purpose of this I'm not counting Innawood) it's easier for people to set up a blacksmithing toolset from scratch instead of looting a pair of tongs there's a fundamental design flaw that has to be addressed and it's way beyond the scope of just crafting recipes.

There are ways to craft flatjaw tongs without flatjaw tongs, historical or not:
- Using another tool as a substitute. There's some other type of tongs (NOT kitchen tongs) or even pliers that theoretically could, albeit less efficiently, get the job done well enough to get you pliers. The issue here is this does not resolve the elephant in the room that is Innawood recipes, and you'd still have to luck out on RNG for looting those tools instead
- Bronze cast tongs. This should be a shitty alternative that gets damaged as you use them. Unfortunately we can't simulate that. If/When someone implements tongs for holding a crucible this recipe also falls apart
- Normal flatjaws BUT made from a longer stock so you can work with only a part of them. Normal blacksmithing recipe that doesn't take tongs, would need a rebar or so as an ingredient, give some lumps as a byproduct, take longer. Probably gonna be the choice I end up going for now

Karol1223:
> There is a way to model extreme damage to tools by making the tools "components" rather than "tools". If that approach was used, I'd suggest that the recipe would produce scraps as a byproduct. That would make the tools single use rather than damaging them, but if these makeshift tools are used to produce their proper version I don't think it's much of an issue, since you've got no reason to use the makeshift version once you've produced the real deal.

This has been suggested many times on Discord and unfortunately, it's a very hacky solution. Another solution that was suggested but is even more hacky is stealing the "charge" system the fire drill uses. Both would work but both are very suboptimal and really work on tools that'd be damaged as they're used should, ideally, be on hold until we can simulate this properly without bandaid solutions. As it stands it's easier to go the long stock route and if Innawood can indeed get rebar without tongs then this would solve all the issues at hand.

Karol1223:
So because things in this PR change on the fly with me changing the tongs recipe like 3 times today I'm saying fuck casting bronze tongs we ball and cast steel

Because apparently, that's actually a semi-viable solution that I overlooked entirely. It's gonna be worse, it's gonna be unlockable at a higher level. But it's gonna *work*.

PatrikLundell:
The only potential problem I see is if the changes would end up making it impossible to grind the skill/proficiency levels because things end up gated behind a tool you can't reach. Given the number of recipes available I don't think that's an actual issue.

Karol1223:
> The only potential problem I see is if the changes would end up making it impossible to grind the skill/proficiency levels because things end up gated behind a tool you can't reach. Given the number of recipes available I don't think that's an actual issue.

For skill there's no way this should be a problem because it's fabrication so there's a bazillion recipes that use it. For proficiencies, blacksmithing will likely be slightly harder to grind but I'd hazard a guess that adequately so - if blacksmithing is harder the proficiency should be harder as well. Though it could also likely use a practice recipe if that becomes an issue.

data/json/recipes/tools/tools_hand.json
```diff
@@ -161,9 +159,10 @@
       [ "textbook_armwest", 5 ],
       [ "textbook_weparabic", 5 ]
     ],
-    "using": [ [ "steel_standard", 1 ] ],
+    "using": [ [ "forging_standard", 3 ], [ "steel_tiny", 3 ] ],
     "qualities": [ { "id": "ANVIL", "level": 3 }, { "id": "HAMMER", "level": 3 }, { "id": "GRIND", "level": 2 } ],
-    "tools": [ [ [ "forge", 20 ], [ "oxy_torch", 20 ], [ "fake_arc_furnace", 10 ] ], [ [ "metalworking_tongs", -1 ] ] ],
+    "//": "Specifically avoids blacksmithing requirement groups so that it's craftable without the metal fileset - this process would still be possibly, just realistically take slightly longer.",
```
Brambor, line 164:
```suggestion
    "//": "Specifically avoids blacksmithing requirement groups so that it's craftable without the metal fileset - this process would still be possible, just realistically take slightly longer.",
```

Karol1223:
I got impatient. It SHOULD be done now...

@Light-Wave apologies for pinging you so much under this but I have to ask you to review this again. I've done some changes, should hopefully now be Innawood-friendly. The only problem I can think of is the recipes using chunks instead of lumps - can Innawood get those at this point of the progression?
- tongs can now be cast, no more bronze tongs - should be better compatibility
- bronze hammer can substitute drift with a nail punch
- metalworking chisel got an updated recipe - it still requires tongs but tongs should be now obtainable earlier so I don't think this is a problem now

Light-Wave:
No worries for the pings, I appreciate being kept informed. :)
Looks like flatjaw tongs require metal fileset which require flatjaw tongs. If flatjaw tongs must be filed, perhaps `grinding 2` could be used instead?
It also looks like bronze hammers can still be used for prying purposes. At least it still has the action to do so. Not sure how that works when the item does not have a prying quality. Perhaps delete does not work for use_actions, so you might have to redefine the whole item instead of using copy-from.
Innawood also can not gain scrap metals that early right now, but that can be fixed by adding some more recipes over [here](https://github.com/CleverRaven/Cataclysm-DDA/blob/e91293b05bb364763fffae160332386756a75bae/data/mods/innawood/mining/recipes/smelting.json) to let you process the ore to smaller chunks instead of to just steel_lump. I could fix that myself if you prefer.

Karol1223:
Okay, @Light-Wave attempt nr 3

- I swapped the tongs to be able to use grinding 2 for the cast version as you suggested. It makes sense. The only issue is the grinding stone maybe should not be grinding 2 but that's technically out of scope of this
- I added a null use action to the bronze hammer, you were right that ``delete`` doesn't work with use actions. I should know by now not to trust it
- I added a scrap recipe analogous to the lump recipe for Innawood - I recalculated the time and ``bloomery`` needed by dividing it by 17 (since lump uses 17x the weight in ore, 17x the weight of scrap would be 0.85kg but one ore weighs 1kg. I think it's good enough to abstract added mass?)

What I do wanna know is if with those changes the recipes for affected tools for Innawood should be deleted? I think with the new recipe for scrap, the custom recipe for the fileset for example isn't necessary? Also, on a completely unrelated note... is the smelting ore into lump *supposed* to be reversible?

Light-Wave:
They are absolutely not meant to be reversible, lol.

So in Innawood there are two paths to steel; iron ore (from bog iron, rewards hard work), and hematite (from craters and caves, rewards exploration). If you'd be so kind, could you make it so that both sources can be smelted into any size of steel bits? I've been meaning to do it myself, but I've never had enough of a reason before.

So that means
- making steel_chunk from iron_ore
-  making chunks and scraps from hematite. ( One ore of hematite is worth 2 chunks and 3 lumps)

I suggest that you keep the inputs and time the same, and instead scale up the outputs. Otherwise players will try to game the system by only smelting in the most efficient way, and I think that would take away some enjoyment. Doing it this way also means you won't have to recalculate batch_time_factors to keep them fair.

The tongs and hammer seems to work.

Light-Wave:
As a side note, you seem to have removed reversible from chisel. Doesn't make much of a difference, but thought I'd let you know in case it was a mistake.

Karol1223:
Yeah I think I can do the metal recipes. That makes sense.

Regarding the chisel - it's intentional. Reversible blacksmithing would create really awkward disassembly recipes. I intend to add a manual disassembly for it in a future PR.

Karol1223:
Okay, added the new recipes for scrap and chunks - I'd like it if you could look at it to make sure I didn't mess up the amount given.
I also removed the ``reversible`` from smelting bog iron into lumps given what you said and removed the custom metal fileset recipe because the mainline one should now be sufficient with the new changes.

Is that all good?

Light-Wave:
Yeah that all looks good. Let's get it merged. Good job.

Light-Wave (approved):

PR #67148, opened 2023-07-24
Documentation: New TESTING_YOUR_CHANGES.md

NetSysFire:
#### Summary
Infrastructure "New documentation on how to test proposed changes"

#### Purpose of change

Documentation! So far I have seen none about this.
My intent is to document the workflow to make it a less ardous task for a new contributor.
Also maybe you have some better testing steps than I do?

#### Describe the solution

Adds some docs and also modifies the PR template to point to it.

Different file name? Merging with TESTING.md?

#### Testing

Uhhh. This is ironic.

### DIFF
--- a/doc/TESTING_YOUR_CHANGES.md
+++ b/doc/TESTING_YOUR_CHANGES.md
@@ -12,8 +12,6 @@ so read this one first.
 
 For JSON changes, read [JSON_STYLE.md](JSON_STYLE.md).
 For C++ changes, read [CODE_STYLE.md](CODE_STYLE.md).
-If tinkering with C++, you may also want to look at
-[TESTING.md](TESTING.md) for our unit tests.
 
 All released builds include a pre-compiled `json_formatter.cgi`, which
 is a handy tool to automatically format any JSON file you point it at
@@ -39,6 +37,8 @@ Cataclysm or reload your save, whatever is more convenient.
 
 For C++ changes, you will need to compile the game yourself.
 See [COMPILING.md](COMPILING/COMPILING.md).
+You may also want to look at [TESTING.md](TESTING.md) for our unit
+tests.
 
 ## Using the debug menu
 
--- a/doc/TESTING_YOUR_CHANGES.md
+++ b/doc/TESTING_YOUR_CHANGES.md
@@ -12,6 +12,8 @@ so read this one first.
 
 For JSON changes, read [JSON_STYLE.md](JSON_STYLE.md).
 For C++ changes, read [CODE_STYLE.md](CODE_STYLE.md).
+If tinkering with C++, you may also want to look at
+[TESTING.md](TESTING.md) for our unit tests.
 
 All released builds include a pre-compiled `json_formatter.cgi`, which
 is a handy tool to automatically format any JSON file you point it at
--- a/doc/TESTING_YOUR_CHANGES.md
+++ b/doc/TESTING_YOUR_CHANGES.md
@@ -62,6 +62,14 @@ they will not be applied.
 With *Spawning > Spawn monster*, you can spawn an individual monster
 or an entire horde of them.
 This is very useful to test e.g monster faction and combat behavior.
-
 Unfortunately there is no way to spawn a monster group (`mongroup`)
 yet, as of writing.
+
+For more advanced playtesting, use loadouts from the Standard Combat
+Testing Mod (included in the Misc category of built-in mods) and document
+your results in the Testing section of your PR.
+
+Important test tips:
+ - A spawned and saved monster always stays the same, even if you change the underlying monster definition.  Always use freshly spawned monsters!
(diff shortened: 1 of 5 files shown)


GuardianDll:
Recommend to add somewhere at the beginning a note that we use doublespace after a comma (maybe even add a simple regex to find where your json has lack of, but it's probably too much)

Procyonae:
"Spawning individual pieces of overmap terrain is usually not as commonly used and less useful in that sense."
I somewhat disagree with this, placing singular pieces lets you place them right next to yourself which is much more convenient for testing as you make changes and the regenerate option is very useful for checking random chance stuff like nest and vehicle placement very quickly.

NetSysFire:
Would you be so kind and make a suggestion via the review feature? Since you worked a lot with mapgen stuff it would be cool if you can word it yourself so I do not mess it up.

Venera3:
I added a part about testing monsters to `MONSTERS.md` a while ago, you could move that over.

doc/TESTING_YOUR_CHANGES.md
```diff
@@ -0,0 +1,67 @@
...
+A very simple way is to simply place your Cataclysm executable into your
+git repository.
+It will automagically use this altered path to load resources.
+For more advanced situations, there are commandline parameters such as
+`--datadir`, which you can point at your git respository.
+Check `--help` for more information.
+
+Exiting to the main menu and loading your save again will re-read most
+JSON files, so after making changes, make sure to either relaunch
+Cataclysm or reload your save, whatever is more convenient.
```
Qrox, line 19:
```suggestion
If you need to test JSON-only changes, a very simple way is to simply place your Cataclysm executable into your git repository.
It will use this altered path to load resources.
For more advanced situations, there are commandline parameters such as
`--datadir`, which you can point at your git respository.
Check `--help` for more information.
You should also use an up-to-date version of Cataclysm and ensure that the git revision matches the Cataclysm executable version to avoid conflicts due to other people's changes to the JSON handling code.

Exiting to the main menu and loading your save again will re-read most
JSON files, so after making changes, make sure to either relaunch
Cataclysm or reload your save, whatever is more convenient.

For C++ changes, you'll need to compile the game, which is covered by `doc/COMPILING`.
```

Added some distinction between JSON and C++ changes. Simply placing the executable in the git repository has the potential of version mismatch so a note is also added.

For the altered path loading, it is a convention for most programs to use the resources in the directory they are located so there's nothing "automagical" about it really.

doc/TESTING_YOUR_CHANGES.md
```diff
@@ -0,0 +1,67 @@
...
+
+The *Map editor* can be used to alter small amounts of terrain and also
+apply a different overmap terrain without actually touching its ID.
+
+The *Overmap editor* is mainly useful to spawn buildings and other specials.
+For spawning entire buildings or other things occupying multiple overmap
+tiles, type `s` to spawn a special. Spawning individual pieces of overmap
+terrain is usually not as commonly used and less useful in that sense.
+
+Make sure to only spawn overmap specials in the non-red areas, or
```
Qrox, line 57:
```suggestion
Make sure to only spawn overmap specials in the non-highlighted areas, or
```
IIRC the already generated areas are marked with the highlight tile in the graphical overmap.

Qrox:
> Recommend to add somewhere at the beginning a note that we use doublespace after a dot (maybe even add a simple regex to find where your json has lack of, but it's probably too much)

It will be caught by the text style check anyway, so a reference to `MANUAL_OF_STYLE.md` should be enough. (Might want to also mention `CODE_STYLE.md` and `JSON_STYLE.md` in that case.)

I also recommend mentioning `TESTING.md` for how to run the unit test.

NetSysFire:
>I also recommend mentioning TESTING.md for how to run the unit test.

I am unfamiliar with running tests manually. Why and when would you do that?

doc/TESTING_YOUR_CHANGES.md
```diff
@@ -12,6 +12,8 @@ so read this one first.
 
 For JSON changes, read [JSON_STYLE.md](JSON_STYLE.md).
 For C++ changes, read [CODE_STYLE.md](CODE_STYLE.md).
+If tinkering with C++, you may also want to look at
+[TESTING.md](TESTING.md) for our unit tests.
```
Qrox, line 16: Please move this to the "applying your changes locally" section, after the mention of `COMPILING.md`.

> I am unfamiliar with running tests manually. Why and when would you do that?

When you change the unit test code itself, change C++ or JSON code covered by the test cases, or need to fix failing CI unit tests caused by your PR.

It is not uncommon for people to submit their changes without running the unit test though, and the CI unit test should catch any problems in that case, so a simple mention that the unit test exists should be enough.

Procyonae:
> Would you be so kind and make a suggestion via the review feature? Since you worked a lot with mapgen stuff it would be cool if you can word it yourself so I do not mess it up.

I keep forgetting about this 🤦 I think I'll wait until this is merged and then add to it properly

Maleclypse:
Hi Netsysfire, is this ready for merge review?

NetSysFire:
Yes. Its all done now.

PR #67155, opened 2023-07-25
Nylon quiver

Lalomb0000:
#### Summary
Content "adding nylon quivers with MOLLE capabilities"

#### Purpose of change
To add a nylon quiver and test the usability of a quiver as a molle attachment 

#### Describe the solution
To not add them

Leaving nylon undisturbed

#### Testing
Recipe added and tested

#### Additional context
no spawn, only craftable


TheShadowFerret:
Arrows(76cm) being longer than what the large holster allows(45cm), I think this should be a PALS_LARGE item.

Aeiouuu01234:
> ~PALS_HUGE when~

I want to attach a backpack in my backpack

PR #67166, opened 2023-07-25
[MoM] Add u_message to "extremely important note"

Standing-Storm:
#### Summary
Mods "Add u_message to the extremely important note"

#### Purpose of change

Someone on Discord mentioned that picking a starting psionic profession, they weren't really clear why the extremely important note was so important

#### Describe the solution

Adds a brief message that hints at the note's purpose and provides a pointer to the recipes tab.

Letting someone else do it or waiting until it's possible to make professions start with recipes


GuardianDll:
...why not just add u_message into recipes?

Standing-Storm:
> ...why not just add u_message into recipes?

I call those same EoCs to teach the recipes during the matrix crystal and portal storm awakenings, so adding a message about reading a note directly to those EoCs wouldn't make sense.

PR #67169, opened 2023-07-25
Update BFR

Holli-Git:
#### Summary
Bugfixes "Adjusts 45-70 BFR"

#### Purpose of change

Wanted to put a BFR in a large holster, thought I could just add barrel volume and saw it off. Made me realize that something was off, and with the help of Krita remeasured the feret length. New length, adds barrel volume, adjusts damage. 

#### Describe the solution

Adjusts the BFR.

Stuff. 

#### Testing

Tested locally, works fine.

#### Additional context
Here's my calcs and train of thought.
~~550 pixels for barrel of 7.5 inches. Meaning 73 ⅓ pixels per inch. Feret length is 1216.9 pixels, meaning 16.6 inches, or 422 mm. Barrel diameter is 54.5 pixels, meaning .3716 inches of barrel radius, meaning 3.25 in^3 or 53 ¼ milliliters of barrel volume.~~
Redid it all in ImageJ 
image
image


tenmillimaster:
- Revolver barrel lengths don't include the cylinder. You'd need to add the cylinder length for your calculations. In reviewing your work I've gone through the imagej stuff myself and calculated the length to be 10.621".
- Checking your calculations, it seems like you'd like to cut the entire barrel off? That'd remove the ejector rod and would drastically slow down reloads, and also drastically change the damage that'd be appropriate.
- Sawing off the barrel removes the front sight; we'd have to have the game remove the sights if we let someone cut the barrel off and this with it. 

BFR measurements
Here are my measurements, along with a more conservative length to saw off that preserves the ejector rod and ejector rod assembly, but still removes the front sight.

Paprika-H:
> * Revolver barrel lengths don't include the cylinder. You'd need to add the cylinder length for your calculations. In reviewing your work I've gone through the imagej stuff myself and calculated the length to be 10.621".
> * Checking your calculations, it seems like you'd like to cut the entire barrel off? That'd remove the ejector rod and would drastically slow down reloads, and also drastically change the damage that'd be appropriate.
> * Sawing off the barrel removes the front sight; we'd have to have the game remove the sights if we let someone cut the barrel off and this with it.
> 
> BFR measurements Here are my measurements, along with a more conservative length to saw off that preserves the ejector rod and ejector rod assembly, but still removes the front sight (but keeps it included in the calculations, assuming you're gonna reattach it somehow). It's always a bit of a guess as to what length "overall length comes from" but usually bounding box length is used. To verify I measured out the barrel length stated and it comes out to be exactly where their stated length is.
> 
> Lengths from https://sportsmansfinest.com/bfr-revolver-45-70-ss-7-5-l/, image from https://cdn11.bigcommerce.com/s-kp97pt369w/images/stencil/1280x1280/attribute_rule_images/3359_source_1559746034.jpg

Removing the barrel volume, as tbh I didn't think too hard while doing that. But as far as I can tell, you use feret diameter for length, not the width or height. Also did pythag which gave 410mm, but going with the 422 as I believe it is more accurate, due to not knowing what exactly the company measured, as I measured from top of the sight to the bottom left of the grip. Also for the cylinder length, is that for the damage? Because while yes, it does add about a half inch, which is equal to another point of damage, revolvers also lose a good bit of energy from the cylinder gap, which I assume the two would almost cancel out.  And the image in the 2nd, length is a 10" barrel, not a 7.5", and also uses a bisley grip, when the gun in game uses the standard grip. Got the numbers for pythag from <https://shopkahrfirearmsgroup.com/bfr-45-70-stainless-steel/> which is directly from Magnum Research.

PR #67173, opened 2023-07-25
use nested mapgen on zombie cages, lakeside cabin special.

ghost:
#### Summary
Bugfixes "special lakeside cabin version has a lower chance to spawn"

#### Purpose of change
There's a lakeside cabin with zombie cages and a corpse in the freezer...i'm not really sure if this is a reference to something
instead of removing the cages (and the corpse), i used nested mapgen so the cages have a small chance to spawn, rather than always, and lowered the chance for the corpse to spawn.

#### Describe the solution
occurence and flag change

Expanding on it somehow? i'm not really sure, as i said, i don't know if its a reference.

#### Testing
a quick overmap hop to make sure i didn't flop this, and it works fine.

#### Additional context
cabin in question (had to force spawn it hence why weird map gen)
image

### DIFF
--- a/data/json/mapgen/lake_buildings/lakeside_cabin.json
+++ b/data/json/mapgen/lake_buildings/lakeside_cabin.json
@@ -456,26 +456,22 @@
         "      ~  #B.=..=.s#     ",
         "      ~  wB.#..#.T#     ",
         "      ~  #D.#..#&h#     ",
-        " ~~~~~~  ####ww####     ",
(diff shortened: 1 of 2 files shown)


Procyonae:
You'll need to exclude this from the overmap test or it'll fail very frequently, also if no other non weird version exists you could just make the weird stuff a rare nest instead and leave the spawn rate alone

PR #67184, opened 2023-07-26
Added solar panels to the roof of LMOE bunkers

ActualDio:
#### Summary
None

#### Purpose of change
It doesn't make much sense that a fully independent bunker wouldn't have at least one solar panel, and rely exclusively on a gas generator.

#### Describe the solution
Simply added a new entry for a solar panel in the bunker_ext json pallete and then put four of them at the center of the lmoe roof grid

None


Maleclypse (changes requested):
This would be better as a nest that has a chance to spawn rather than every single LMOE no coming with 4 solar panels, especially since solar panels would make an LMOE easier to find.

Maleclypse (approved):

data/json/mapgen/nested/lmoe_nested.json
```diff
@@ -84,5 +84,15 @@
         }
       ]
     }
+  },
+  {
+    "type": "mapgen",
+    "method": "json",
+    "//": "This is done to give a chance for lmoe bunkers to spawn with solar panels on the roof",
+    "nested_mapgen_id": "lmoe_solar",
+    "object": { "mapgensize": [ 2, 2 ], "rows": [
```
Maleclypse, line 93: You'll need to define `M` here though

Maleclypse: @Procyonae You've been working with nested mapgen a lot lately.  Nests can't use palettes right? Or am I remembering incorrectly?

Procyonae: Nests can use palettes just fine

data/json/mapgen/nested/lmoe_nested.json
```diff
@@ -84,5 +84,15 @@
...
+        "MM",
+        "MM"
+      ], "palettes": "roof_pallete" }
```
Procyonae, line 96:
```suggestion
      ],
      "palettes": [ "roof_pallete" ]
    }
```
Palettes has to be an array

PR #67186, opened 2023-07-26
More monsters fight back if cornered

worm-girl:
#### Summary
Balance "More monsters fight back if cornered"

#### Purpose of change

#66294 introduced the CORNERED_FIGHTER flag, which gives monsters the ability to switch back to hostile if they're unable to escape an enemy they want to flee. It had several limitations, such as not working on NPCs and in some cases not giving the monster enough time to escape.

#### Describe the solution

* Improves the CORNERED_FIGHTER code in monster.cpp so that cornered monsters will now spend 3 turns trying to flee (or tracking, if that's what their AI calls for) before turning and fighting. They can attempt to flee again 9 turns later. This is done via a pair of AI effects, spooked and spooked_recent. The code can now also check for NPCs as well as the player and other monsters.
* Adds CORNERED_FIGHTER to all mi-go except the scouts.
* Adds CORNERED_FIGHTER to bears, boars, moose, reindeer, pigs, cows, coyotes, some dogs, and llamas.
* Does NOT add CORNERED_FIGHTER to any animal which uses HIT_AND_RUN.

This is a marked improvement to monster behavior, but there are still some issues. Mi-go in the daytime for instance will often run away for ten seconds and then come sprinting right back. They did this before this PR though, so one thing at a time.

#### Testing

Spawned mi-go and tried fighting them. When wounded they would switch to flee or tracking as expected, but if an attacker stayed near or pursued them, they would flip hositle again after a couple of seconds, rather than standing there cowering while getting beaten down.

Went into a cockroach basement with a flashlight out and spooked the little guys away, then pursued them into corners to make sure they'd still go hostile.

Spawned a bear and tried attacking it. Once it tried to flee, if I stayed near it it would go hostile.

Walked up to a moose and punched it. The moose punched back.

Walked up to some boars. They quickly fled, but fought back when engaged in melee.

Spawned mi-go and zombies to watch them duke it out. The mi-go would try to flee when injured, but would fight back if it couldn't.

Attacked some dogs and pigs. They still are pretty skittish, but would bite me once or twice if they couldn't get away.

### DIFF
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -3188,7 +3188,6 @@ void monster::process_effects()
             }
         }
     }
-    
 
 
     if( has_flag( mon_flag_PHOTOPHOBIC ) && get_map().ambient_light_at( pos() ) >= 30.0f ) {
--- a/src/monster.cpp
+++ b/src/monster.cpp
@@ -3164,19 +3164,24 @@ void monster::process_effects()
                         add_effect( effect_spooked, 3_turns, false );
                         add_effect( effect_spooked_recent, 9_turns, false );
                     }
-                    if( guy ) {
-                        monster_attitude att = attitude( guy );
-                        if( ( friendly == 0 ) && ( att == MATT_FOLLOW || att == MATT_FLEE ) && !has_effect( effect_spooked ) ) {
-                            if( !has_effect( effect_spooked_recent ) ) {
-                                add_effect( effect_spooked, 3_turns, false );
-                                add_effect( effect_spooked_recent, 9_turns, false );
-                            }
-                            else {
-                                if( morale < type->morale ) {
-                                    morale = type->morale;
-                                    anger = type->agro;
-                                }           
-                            }
+                } else {
+                    if( morale < type->morale ) {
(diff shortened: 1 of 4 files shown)


src/monster.cpp
```diff
@@ -3149,28 +3151,45 @@ void monster::process_effects()
...
-                if( mon->faction->attitude( faction ) != MFA_FRIENDLY ) {
+            const Character *const guy = creatures.creature_at<Character>( p );
+            if( mon && mon != this && mon->faction->attitude( faction ) != MFA_FRIENDLY &&
+                !has_effect( effect_spooked ) && morale <= 0 ) {
+                if( !has_effect( effect_spooked_recent ) ) {
+                    if( !has_effect( effect_spooked_recent ) ) {
+                        add_effect( effect_spooked, 3_turns, false );
+                        add_effect( effect_spooked_recent, 9_turns, false );
+                    }
+                } else {
```
ZhilkinSerg, line 3167: You are checking for same `effect_spooked_recent` twice here. Must be a mistake.

PR #67168, opened 2023-07-25
house roadmap: update house_01 in-construction variant

ghost:
#### Summary
Content "house roadmap: update house_01 in-construction variant to use modern code"

#### Purpose of change
this is my effort to help #62747, i would like to eventually update all old mapgen.

#### Describe the solution
use of palettes, regionalized terrain, and walls no longer being always wood.

Also added a new palette, which will be used for all in-construction homes and other buildings (when i get to them at least.)

#### Describe alternatives you've considered
None

#### Testing

before version is locked to always wood, and uses non-regionalized terrain, and no palette.
image

image

#### Additional context
i will touch the trashed variant in a different PR. i didn't want to do too much in one PR

PR #67187, opened 2023-07-26
House palette improvements

### LINKED ISSUE #67116: More Dynamic Residential Decay And Messy Houses

#65905 added basic random damage to every map using standard_domestic_palette which while better than no damage could do much to make the world feel more real
#60275 wanted to add untidy and messy variations to houses which I think would also be a great addition

### Solution you would like.

I'm looking to combine both into the existing but underused domestic_general_and_variant_palette we have and expand the usage of the palette to cover more houses that are currently using standard_domestic_palette.

The palette currently looks like this:
```
{
    "type": "palette",
    "id": "domestic_general_and_variant_palette",
    "//": "Standard variant distribution for homes: General: 95%, Abandoned: 2.80%, Survivor: 1.80%, Hoarder: 0.40%",
    "//2": "Please don't forget to apply your changes to each house palette.",
    "palettes": [
      {
        "distribution": [
          [ "standard_domestic_palette", 9500 ],
          [ "standard_domestic_palette_abandoned", 280 ],
          [ "house_hoarder_palette", 40 ],
          [ "house_survivor_palette", 180 ]
        ]
      }
    ]
  },
```
It works that if you use this as a maps palette that you randomly get one of the above based on their respective weights.

I'd be looking to revert #65905's changes to standard_domestic_palette and replace the use here for a copy that inherits from it (to avoid doubling maintenance and allowing pristine properties to still exist where wanted) and then nesting similar distributions for tidy(no palette)/_untidy/_messy and pristine(no palette)/_damaged/_heavily_damaged in it so that you could have an untidy heavily damaged house for example.

~~The only restricting factor code wise is there's no way to scope these distributions afaik unlike other mapgen values so the selection will be different between floors and this can only be used with houses that don't heavily rely on nests, but that seems like something that can be dealt with later.~~ On a second look it looks like [PatrikLundell](https://github.com/PatrikLundell) has been using the mapgen value method to define scoped palettes for basecamp stuff so ig the reports of that method not working are wrong and I'll apply that method instead which is great bc it could eventually be extended to house nests

I'll probably do this in 2 PRs, one to add the existing palette to more maps and one to add the new untidy/messy/damaged/heavily_damaged palettes to it, with the order not mattering.

Any thoughts/suggestions would be very welcome (Better names than damaged/heavily_damaged especially)

A similar chance for a squatter nest in the abandoned palette and an overrun one for the survivor palette could be cool
Ideally ground floors would have more damage than other floors and basements would get changes but I won't be dealing with that for now

### Additional context

While this is a basic sweeping change, a more personalised touch like #60379 was going for for individual properties would still be cool to see for stuff beyond what palette swaps can achieve

Procyonae:
#### Summary
None

#### Purpose of change

Improves on #65905
Rezzes #60275
Closes #67116

#### Describe the solution

Adds untidy and messy house palettes from #60275 to simulate pre-cata disorganisation/clutter

Adds damage and damage_large (please suggest better names) palettes in the spirit of #65905 to simulate damage from zeds and blood/corpses from peeps trying to escape their homes etc

Makes the survivor and hoarder palettes inherit standard_domestic_palette to make them more readable and maintainable

Expands domestic_general_and_variant_palette's usage to all non nest heavy houses (nests don't break the palette in any way but just lead to less sensical results and should be handled separately, the parameters can scope to the whole building including nests so there's no reason this can't be extended to them it's just very much out of scope for this PR)

Adds standard_domestic_migration_palette that applies the damage and tidyness palettes randomly on top of standard_domestic_palette, only used in domestic_general_and_variant_palette. I did this so standard_domestic_palette can shift to be standard_domestic_pristine_palette in future for maps that aren't suited to domestic_general_and_variant_palette. To this end reverts #65905 change to standard_domestic_palette

Current weights:
Tidy 6 / Untidy 3 / Messy 1
Undamaged 2 / Damaged 3 / Very Damaged 1

Tweaking the palettes in future should be nice and easy

Will make a squatter palette for abandoned houses in a separate PR
Might add a looted palette for regular houses in a separate PR
Might expand domestic_general_and_variant_palette to nests in a separate PR (Not for some time though)
Might expand tidyness and damage params to survivor palette in a separate PR
Ideally I'd split the palette for ground floor and upper floors so upper floors have less damage but I'm not sure how I'd want to approach that
An in-construction palette as a possible result of domestic_general_and_variant_palette trying to make a generalised representation of what Devengel has been doing in #67168 would be cool for a separate PR
Using bash on the furniture rather than the transforms but bash doesn't work great so I'd rather not apply it to so many buildings

#### Testing

Went through every file to check the palette's only being applied to what look to be sensible picks
Went around lots of cities to check results look decent

#### Additional context

<details>

image

Tidy + Damage

image

Tidy + Damage large

image

Untidy + Damage

image

Untidy + Damage large

image

Messy + Damage

image

Messy + Damage

image

Messy + Damage large

image

Hoarder (existing palette new houses)

</details>

#67116 contains additional explanation

I did want to split this into 2 PRs, one making the palette changes and one applying it to more maps but it didn't really make much sense in the end imo, if the reviewer would rather I split this tho that's cool


Ranik-Ortega:
Would it be possible to get some screenshots of it in action? I'm curious of the changes compared to 65905

Procyonae:
> Would it be possible to get some screenshots of it in action? I'm curious of the changes compared to 65905


Added screenshots at the bottom ^^

RanikOrtega:
> > Would it be possible to get some screenshots of it in action? I'm curious of the changes compared to 65905
> 
> Added screenshots at the bottom ^^

Thanks. They all look great. Do  "Normal" pallet houses still have a small chance of window / door / graffiti damage? I think even the "undamaged" buildings having small changes like in 65905 helps keep things consistent and reduces the number of "pristine" buildings in a post apocalypse setting.

Procyonae:
> Thanks. They all look great. Do "Normal" pallet houses still have a small chance of window / door / graffiti damage? I think even the "undamaged" buildings having small changes like in 65905 helps keep things consistent and reduces the number of "pristine" buildings in a post apocalypse setting.

No but if you want to add that in a follow-up PR you're welcome to, the numbers probably want tweaking generally but I'd rather wait for player feedback

data/json/mapgen_palettes/house_general_palette.json
```diff
@@ -141,9 +170,8 @@
       "u": "t_region_groundcover_urban",
       "X": "t_region_groundcover_urban",
       ".": "t_region_groundcover_urban",
-      "+": [ [ "t_door_b", 2 ], [ "t_door_frame", 2 ], [ "t_door_c", 5 ], [ "t_door_o", 5 ], [ "t_door_locked_interior", 1 ] ],
+      "+": [ [ "t_door_c", 5 ], [ "t_door_o", 5 ], [ "t_door_locked_interior", 1 ] ],
```
RanikOrtega, line 173: Reccommend keeping but toning down this. 

"+": [ [ "t_door_b", 1 ], [ "t_door_frame", 1 ], [ "t_door_c", 5 ], [ "t_door_o", 5 ], [ "t_door_locked_interior", 1 ] ],

data/json/mapgen_palettes/house_general_palette.json
```diff
@@ -160,7 +188,6 @@
       ":": "t_wall_glass",
       "-": "t_sidewalk",
       "o": [
-        [ "t_window_frame", 4 ],
```
RanikOrtega, line 163: Recommend keeping this but toning down to  [ "t_window_frame", 2 ],

Procyonae: Personally I don't understand why you want there to be consistent damage across all buildings as opposed to variation, like I said if you want to make a PR arguing for this go ahead but I don't agree. Making pristine buildings rarer would be fine but like I said I want to wait for player feedback. Equally this is representing the starting day situation, I'd be down to reduce pristine variants over time once I get round to using and expanding on #66890

PR #67196, opened 2023-07-26
allow digging 2 for brick ovens

nornagon:
#### Summary
None

#### Purpose of change

This recipe specified shovel or wooden shovel, but I don't think there's a reason that an entrenching tool, lobotomizer or rat/rabbit/wolf/bear paws (fake_burrowing) wouldn't work just fine.

#### Additional context

cc @RenechCDDA


RenechCDDA (approved):
Good change, wholeheartedly agree.

PR #67199, opened 2023-07-26
Make bone broth better by extracting more calcium using acified water.

ampersand55:
#### Summary
Content "Make bone broth better by extracting more calcium using acified water"

#### Purpose of change

There are few good natural sources of calcium in the game and I wanted to wanted to make bone broth a better source of calcium. 

In many cultures (such as many Asian cultures) they prescribe the use vinegar to extract more calcium from bone broth, especially for pregnancy and postpartum periods.

This has been confirmed to work in several studies. The one I use for the numbers in this PR is "Essential and toxic metals in animal bone broths":

* https://doi.org/10.1080%2F16546628.2017.1347478
* https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5533136/ (alt link)

See Figure 1. for the dramatic increase using acified water compared to un-acified:

https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5533136/figure/F0001/

#### Describe the solution

TL;DR:

We add 1 unit of vinegar per 1 unit bone to the recipe to make it extract more calcium from the bones as per the study.

Math:

The acified water used in the study:

> acidified water that was made by mixing 20 ml of table vinegar with 1 l of deionized (DI)

One serving of vinegar in game is 2500 / 160 = 15.625 ml, which is close to the 20 ml used in the study, so we can just add 1 unit of vinegar to the recipe.

The cooking time in the study I looked at was 2h, which is about the cooking time of 9x non-canned bone broth in game. This seems reasonable to me. 

The bones used in the study was:

> The animal bones in this evaluation were the rear leg (femur) and rib bones of domestic pigs (both white and black) and bovine leg (femur) bones, all of which were purchased from a local meat market. The white and black pigs were raised in Taiwan and typically fed on forage and food waste, respectively, while the bovine bone was imported from Australia, which was the most common source of local supplies.

This seems reasonably close to the bones we would get in game to me.

Results for boiling in acidic water for 2 hours (Table 1. in the study) in mg/kg.

```
Pig (white) 241 ± 11.8
Pig (black) 252 ± 19.05
Bovine      274 ± 48.3
```

The average of the three means values is 255.66 mg/kg.

The serving size in game is 267 mg, so 255.66 / 1000 * 267 ~= 68 mg calcium per serving. The RDA for calcium is 1000 mg for adults (according to the study), so it would be 68/1000 = 6.8 % RDA calcium per serving size, which we can round to 7 %.

They also studied the effect on iron, but it was not that much different than from non-acified water (Figure 1). At around 200 μg/kg it's pretty negligible. 

I considered making new recipes instead of updating the old ones. But then we would have 12 bone broth recipes including the canned versions, which seems kind of much for such a niche recipe.

#### Testing

Game loads and recipe works as expected.

### DIFF
--- a/data/json/items/comestibles/soup.json
+++ b/data/json/items/comestibles/soup.json
@@ -40,7 +40,7 @@
     "symbol": "~",
     "quench": 30,
     "calories": 43,
-    "description": "A tasty and nutritious broth made from bones.",
+    "description": "A tasty and nutritious broth made from bones, with added vinegar to extract more calcium.",
     "price": 350,
     "price_postapoc": 25,
     "material": [ "bone" ],
@@ -50,7 +50,7 @@
     "phase": "liquid",
     "flags": [ "EATEN_HOT", "NUTRIENT_OVERRIDE", "NO_SALVAGE" ],
     "fun": 10,
-    "vitamins": [ [ "calcium", 1 ], [ "iron", 1 ] ]
+    "vitamins": [ [ "calcium", 7 ], [ "iron", 1 ] ]
   },
   {
     "type": "COMESTIBLE",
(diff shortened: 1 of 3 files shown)


ampersand55:
It's of course better to just eat the bones raw. 

A character can consume a whole 250 ml raw bone in 50 seconds for 96 % calcium with no enjoyment penalty and no toxins. This can't be what's intended.

Maleclypse:
Thanks for doing this!

Issue #67172, opened 2023-07-25
Scrap Cuirass not repairable, unlike XL or XS Variants


Dandy1096:
### Describe the bug

The Scrap Cuirass cannot be repaired, but XL or XS variants can.

### Attach save file

N/A

### Steps to reproduce

1. Obtain crafting recipe of Scrap Cuirass and XL or XS Scrap Cuirass 
2. Scrap Cuirass cannot be repaired, but XL or XS variants can.

### Expected behavior

I suspect that this is an oversight, since the Scrap Cuirass is likely a much older item than the XL or XS variants. It should be repairable using the same items and tools- Bone Needle, Needle, Wooden Needle, Sewing Kit, or Tailor's Kit with Cotton and Budget Steel.

### Versions and configuration

Version 2023-07-20-0251. No mods or alternate configurations.

RenechCDDA:
The issue is that copy-from isn't copying the `repairs_with` field.

Also not sure `repairs_with` is working correctly - it's pointing at budget_steel which defines scrap as a repair item(below) but it's reporting not repairable.

https://github.com/CleverRaven/Cataclysm-DDA/blob/15f392c113c6900d2f930a63d0a957fd66fcf3c8/data/json/materials.json#L355

Procyonae:
Found the problem no tools can repair budget_steel so it needs another mat to provide tools, will make a PR to fix

PR #67202, opened 2023-07-26
Adds inheritance and replace_materials support to repairs_with

Procyonae:
#### Summary
None

#### Purpose of change

More inheritance good

#### Describe the solution

If you have a base item with say `mc_steel` and `leather` materials that can only be repaired with `mc_steel` and copy from it and use `"replace_materials": { "mc_steel": "ch_steel", "leather": "kevlar" }` before it would've been repairable by `kevlar` and `ch_steel` whereas now it should only be repairable by `ch_steel`

#### Testing

Works good, checked combinations of replace_materials, extend and delete

### DIFF
--- a/src/item_factory.cpp
+++ b/src/item_factory.cpp
@@ -3955,8 +3955,17 @@ static void replace_materials( const JsonObject &jo, itype &def )
         material_id replacement = material_id( jv.get_string() );
 
         units::mass original_weight = def.weight;
-
         int mat_portion = def.materials[to_replace];
+
+        // make repairs_with act the same way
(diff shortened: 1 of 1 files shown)


RenechCDDA:
Partially addresses #67172 it seems?

Procyonae:
> Partially addresses #67172 it seems?

Weirdly no

I just got confused why the budget_steel version that was part of my testing was irreparable so thanks for linking that I'll have a look

Procyonae:
Realised that assign seems to magically provide extend/delete support without having to anything, I'll have to refactor my redundant #66960 lol

PR #67205, opened 2023-07-26
Bovine Bulk mutation makes you stronger

GalacticApple:
#### Summary
Balance "Makes Bovine Bulk and Bull Roids increase your strength"

#### Purpose of change
These mutations are described as giving you a large amount of muscle mass; so much so that it makes it more difficult for you to swim. I think it makes sense then for these to actually make you stronger on top of the health bonus. 

#### Describe the solution
Just a simple JSON edit

#### Testing
Works on my CDDA copy.


fairyarmadillo:
Bovine already gets a shitload of strength, more than almost any other mutation. This is probably overdoing it.

anoobindisguise:
Alternatively you could reflavor these mutations to be more than just muscle, like fat and stuff. Cows tend to be really fuckin big in  general. Though that would also probably mean it shouldn't make you drown as easily.

fairyarmadillo:
Bovine already gets ponderous which greatly lowers its base movement speed, it doesn't need any more of that either.

anoobindisguise:
> Bovine already gets ponderous which greatly lowers its base movement speed, it doesn't need any more of that either.

I'd be interested in it having movement speed lost from mixed effects, rather than just negative ones, so you get more good with bad and can't just be a fast cow by not having instability (but that's kind of a separate issue)

fairyarmadillo:
Agreed but I don't think we should start getting rid of red mutations like that until we have a new model for mutating, as some of the proposed models require red mutations to work right.

ampersand55:
Perhaps bovine bulk should come with some integrated armor of "fat" beneath the skin that adds weight, encumbrance, cold resist and bash resist. 

I agree bovine needs some serious buffs though, the Cattle mutation line seems like a worse version of Ursine, but I have a hard time coming up with something besides a post thresh better horns attack.

fairyarmadillo:
> Perhaps bovine bulk should come with some integrated armor of "fat" beneath the skin that adds weight, encumbrance, cold resist and bash resist.
> 
> I agree bovine needs some serious buffs though, the Cattle mutation line seems like a worse version of Ursine, but I have a hard time coming up with something besides a post thresh better horns attack.

Cattle is hands down one of the best mutations in the game. Hooves alone is worth the price of admission but you also get a 60% HP buff on top of +8 strength, horns, and high cold tolerance. They already get two forms of integrated armor and have one of the best dietary mutations, allowing them to live off of grass and bushes.

anoobindisguise:
> Perhaps bovine bulk should come with some integrated armor of "fat" beneath the skin that adds weight, encumbrance, cold resist and bash resist.

You could add a "mutant fat" material and give that to an integrated armor for the fat deposits mutation, something with like pretty low bash resistance (but relatively thick) and no bullet/cut/stab protection but give the integrated armor some critical resistance to represent the known tendency of fat to prevent people and animals from dying to otherwise fatal bullets.

ampersand55:
> > Perhaps bovine bulk should come with some integrated armor of "fat" beneath the skin that adds weight, encumbrance, cold resist and bash resist.
> 
> You could add a "mutant fat" material and give that to an integrated armor for the fat deposits mutation, something with like pretty low bash resistance (but relatively thick) and no bullet/cut/stab protection but give the integrated armor some critical resistance to represent the known tendency of fat to prevent people and animals from dying to otherwise fatal bullets.

Yea, that was what I was thinking.

fairyarmadillo:
> Honestly, I was thinking an alternative could be to just give Cattle mutants Insanely Strong and then make it a requirement for this.
> 
> However, maybe the extra strength wouldn't be _that_ overkill. Does Ponderous make your attacks slower? If so, then maybe having OP levels of strength would be needed. If you can't kill something in a couple of hits you'd be dunzo. If not, then maybe the Bovine Bulk mutations could be re-oriented to not be about you _getting swol_, but more about you being able to more efficiently throw your mass around with a fatigue bonus.

Cattle doesn't need insanely strong, it's already better than the mutants which have it. Those mutants are pound for pound much stronger animals than cattle irl, and they have significant drawbacks that cattle mutants do not have to balance them out.

Ponderous only slows your move speed, not your attacks. Adding more strength on top of the compounding HP buffs and natural weapons (the best in the game, plus they derive their damage from strength) that cattle gets would be ridiculous.

PR #67206, opened 2023-07-27
voice remodulator conflicts SCREECH

anoobindisguise:
#### Summary
Bugfixes "Screeching voice is removed by the voice remodulator"

#### Purpose of change
Most voice mutations weren't removed by the remodulator but bird voice (SCREECH) wasn't.

#### Describe the solution
Add conflict.

Make the remodulator be unremovable since it may replace your voicebox?
Make it compatible with these mutations but override them so you can't just install it and uninstall it to remove the muts permanently?
Make it unremovable, but allow it to upgrade to a regular modulator that has your old voice instead of a creepy robot one

### DIFF
--- a/data/json/bionics.json
+++ b/data/json/bionics.json
@@ -1712,7 +1712,9 @@
     "occupied_bodyparts": [ [ "torso", 2 ], [ "mouth", 1 ] ],
     "flags": [ "BIONIC_FAULTY" ],
     "social_modifiers": { "persuade": -20, "lie": 10, "intimidate": 20 },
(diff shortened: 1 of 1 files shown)


fairyarmadillo:
There may have been some reasoning behind this in that birds produce vocalizations in the syrinx, a structure located in the top of their lungs, rather than in vocal cords in their throat like mammals. 

The remodulator should certainly prevent SCREECH from appearing, but I don't think you'd be able to install it in someone who had this mutation as they use an entirely different method to vocalize.

PR #67207, opened 2023-07-27
Typo Correction in Blaze industries.

Profit0004:
#### Summary
Bugfixes "Typo Correction in Blaze industries mod so shelving matches intended volume."

#### Purpose of change
Shelving is supposed to have 2000L volume, it had 200. 

#### Describe the solution

added a single zero.

### DIFF
--- a/data/mods/BlazeIndustries/items/vehicle/blaze_other.json
+++ b/data/mods/BlazeIndustries/items/vehicle/blaze_other.json
@@ -9,7 +9,7 @@
     "color": "cyan",
     "symbol": "]",
     "material": [ "steel", "plastic" ],
-    "volume": "200 L",
+    "volume": "2000 L",
     "category": "veh_parts",
     "price": 80000,
     "price_postapoc": 1000,


irwiss:
Did you test this change?
IIRC items above 1000L will not be able to spawn on map, so dropping it will make it disappear

ampersand55:
> Did you test this change? IIRC items above 1000L will not be able to spawn on map, so dropping it will make it disappear

Can confirm.

If you construct it on a table or crafting spot you get a "could not place vehicle shelving on map near (x,y)" error, unless you are near a 2000 L+ storage such as a warehouse shelf.

If you construct it without a table or crafting spot you get it in your hands if they're empty, and the same error otherwise.

If a game tile is a maximum of 1x1 meter, then the shelf has to be at least 2 m tall to fit 2000 liters, assuming it's a rectangular prism. Maybe set the `volume` to 1000 l and the `longest_side` to 200 cm.

SurFlurer:
The shelves do not need to be as big as their storage space. They, as items, can be folded or something, making them easier to be carried around.

Profit0004:
I did not test the change, the current shelves cannot store 2000L so I assumed it was a minor bug.   I feel foolish having not tested it, but I was using the GIT web interface at the time, and figured it was just a typo.

Profit0004:
As you can see in this screenshot, I changed it because it was misbehaving, my apologies for not testing the change ->
image

Profit0004:
Hard to believe I have played this game this much and never knew that behavior existed.   I wish it would kick out a warning stating as such.    I can understand the behavior but not the lack of feedback.  Thank you all.

Issue #58222, opened 2022-06-06
Crafting of an item with variants ignores component variant


TheShadowFerret:
### Describe the bug

Crafting an item with variants that itself comes from a variant item ignores the original.

### Steps to reproduce

1. Find or spawn a hard hat or hooded hard hat.
2. Craft a hooded hard hat or disassemble one.
3. See the variant switches.

### Expected behavior

Cartoon hard hats create cartoon hooded hard hats.

### Screenshots

![image](https://user-images.githubusercontent.com/99621099/172257721-4a489e6a-ccc5-4f74-928d-94bc153a9e8a.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.22000.675 (21H2)
- Game Version: 74a67c4 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    No Fungal Growth [no_fungal_growth],
    Bionic Slots [cbm_slots]
]

Termineitor244:
Sadly...
/confirmed

But a system for keeping the variants could make use of the variants IDs, the hard hat variants for example use the same IDs for both item variations.

PR #67209, opened 2023-07-27
Recipe: variants for results

### LINKED ISSUE #67191: Support for crafting variants

There have been several occasions when during planning out PRs I (and a few other contributors I've seen this happen to on the devcord as well) into the issue of having to make an item that was IDENTICAL to another but named differently. It'd be a perfect use case for our variant system, which is amazing and handles it like a champ... except it doesn't support crafting variants. You could add a "makeshift" variant that doesn't spawn anywhere, but you can't make a recipe that crafts specifically this variant, so this entire process is pointless.

I believe the easiest to point out current in-game victim of this is the makeshift body pillow, which could easily be a body pillow variant but isn't due to this limitation.

### Solution you would like.

A field in the recipe to define what variant of the parent item is being crafted. This would solve the aforementioned issue and allow us to handle this with a little more grace in the future.

_No response_

### Additional context
This just came up in a discussion about steel rods, which theoretically could easily be variants of our rebar, but the crafting process would have to be way different.

dseguin:
#### Summary
None

#### Purpose of change
- Resolves #67191

Folks are requesting the ability to make a recipe result in a specific variant of an item.

#### Describe the solution
Adds an optional field to the recipe definition called `"variant"`, which specifies a variant id of the result item. The recipe will always produce this variant when crafted.

If a variant is defined, the recipe's `ident` is appended with the variant id (with a `_` before the variant string). The variant string comes before the `id_suffix`. The resulting ident appears as `<recipe_ident><variant_id><id_suffix>`

When crafting a recipe with a variant, the name listed in the recipe list is pulled from the variant, as well as the item's info on the right:

```json
{
  "//": "Variant version",
  "result": "waist_apron_long",
  "type": "recipe",
  "variant": "pink_apron_cotton",
  "category": "CC_ARMOR",
  "subcategory": "CSC_ARMOR_TORSO",
  "skill_used": "tailor",
  "difficulty": 1,
  "time": "2 h",
  "autolearn": true,
  "using": [ [ "tailoring_cotton_patchwork", 6 ] ]
}
```

Screenshot from 2023-07-27 01-56-12

#### Testing
Added a test case to compare results between non-variant and variant recipes:
```
tests/cata_test "variant_crafting_recipes"
```

Also tested in-game with the JSON and screenshot posted above.

### DIFF
--- a/data/mods/TEST_DATA/items.json
+++ b/data/mods/TEST_DATA/items.json
@@ -1456,6 +1456,116 @@
     ]
   },
   {
+    "id": "test_waist_apron_long",
+    "type": "ARMOR",
+    "name": { "str": "long waist apron" },
+    "description": "A long waist apron made of cotton.",
+    "price_postapoc": 50,
+    "to_hit": 1,
+    "symbol": "[",
+    "looks_like": "vest_leather",
+    "volume": "1750 ml",
+    "weight": "226 g",
+    "price": 1000,
+    "material": [ "cotton" ],
+    "color": "white",
+    "warmth": 2,
+    "material_thickness": 2,
+    "environmental_protection": 1,
+    "flags": [ "POCKETS", "BELTED" ],
+    "armor": [
+      { "covers": [ "torso" ], "specifically_covers": [ "torso_waist" ], "coverage": 100, "encumbrance": [ 1, 1 ] },
+      {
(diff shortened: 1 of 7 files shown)


TheShadowFerret:
How about variants as components determining results for #58222? 👀

PR #67211, opened 2023-07-27
Electric masonry saw

Alm999:
#### Summary
Content "Add Electric Masonry Saw"

#### Purpose of change

I find the fact that there is no electric version of a masonry saw (especially considering we have electric versions of chainsaw and jackhammer) quite strange. Electric masonry saws exist in real life, so why not?

#### Describe the solution

I've added the (on/off) variants by mashing ordinary masonry saw and some data from electric jackhammer and chainsaw. Considering with the jackhammer we have the 13.2 to 1 for battery/gasoline conversion rate, I've set the electric masonry saw consumption to be 1320 charges per usage. The maximum capacity is also scaled. On top of that, as we do not have batteries potent enough to hold 1320+ charges (not counting disposable ones, as there is not enough supply of those), plus changing battery after each use is inconvenient, I've added UPS and plug-in capability.

I see no downsides to having alternatives to scavenging for gasoline for one particular tool…

#### Testing

I've added the changes to game files and tested. It holds charge, it is capable of being turned on and used as a weapon, it charges using plug-in mechanic (can be even plugged in and turned on at the same time!) and can be used in construction blueprints.

I haven't tested its spawn rate at appropriate places, hope the itemgroup list works as I understand.

#### Additional context

None.


Alm999:
Additional note: it seems both electrical
https://econosuperstore.com/itm/14-portable-concrete-saw-3200w-corded-electric-4100-rpm-w-water-pump/
and gasoline masonry saws
https://r1equip.com/product/14-gas-demo-saw-mp/
with 14" blade diameter fluctuate around 10kg (19-21 LBS) in weight. We have a masonry saw with only 4.5 kg. Maybe a weight adjustment is in order?

Knut-Aage-Hofseth:
I think there is a move to get away from UPS and battery powered versions of tools that are corded in real life. There is work being done to make corded tools work. The trick is then to get a storage battery to where you need the tool, plop it down as an appliance and connect the tool to it.

Alm999:
> I think there is a move to get away from UPS and battery powered versions of tools that are corded in real life. There is work being done to make corded tools work. The trick is then to get a storage battery to where you need the tool, plop it down as an appliance and connect the tool to it.
Well, yes. But this still has some issues. If, for example, I am to get this tool into underground lab to remove some walls, then I need at least some way to power this thing up. Getting portable generator is not an option (it can not be moved through z-levels) and I can not plug into the lab's power system (because, like all "hostile appliances", it is magical and does not really exist while still being able to power up those utility lights).

So, apart from constructing some huge monstrosity of 4x4 foldable extra-light frames filled with tool battery mounts and electronics control panel (because, for some reason, installing a storage battery case makes the contraption non-foldable) and a bunch of heavy batteries, having a small integrated battery is the only working solution for now.

So for now, I'll leave the job of getting rid of UPS for someone else (and some other dedicated PR).

gisaku33:
> So, apart from constructing some huge monstrosity of 4x4 foldable extra-light frames filled with tool battery mounts and electronics control panel (because, for some reason, installing a storage battery case makes the contraption non-foldable) and a bunch of heavy batteries, having a small integrated battery is the only working solution for now.

They were explaining that with the current appliance system, you don't need to do all that. All you need is any size of storage battery, place it down as an appliance via the construction menu, and then you can plug in electric devices to it.

Alm999:
> They were explaining that with the current appliance system, you don't need to do all that. All you need is any size of storage battery, place it down as an appliance via the construction menu, and then you can plug in electric devices to it.

I agree, this can work. Somewhat. But, and it is a huge "but", small storage battery contains only 1000 charge (not enough for a single use, unless one counts swinging the thing around as "usage"), car battery weights 20kg while holding only 3000 charges (two uses, yay!), and even a medium storage battery is only 10000 charges (7 uses) for 15 kg and not that easy to come by in the first place.

Now, compared to that, a normal gasoline masonry saw:
1) holds 450 charges (4 uses) of fuel without any additional weight;
2) has portable lightweight canisters with more fuel easily available: a single gallon jug holds 3000+ charges (30 uses!) of gasoline while weighting only 3 kg;
3) ironically, does not poison air with toxic fumes when being used indoors (and IRL you really don't want to use anything with a combustion engine indoors).

Hawing all that, we have in an "electric masonry saw" a vastly inferior tool.
Don't take me wrong, I'm all in for introduction of portable power sources. Plug-in mechanic works great (while I'm in my electric loot car I can use hotplate, water purifier and other appliances without the need for "workshops" or battery chargers). But this is not the purpose of this PR. I want to introduce the "missing piece" of equipment in order to complete the set of construction tools. If someone wishes to re-work the electric tools in a way that they would need a "portable backup generator" and would not have any built-in storage capacity on their own, fine by me. But I think is is a matter for another PR and all of the tools (angle grinder, chainsaw, jackhammer etc.) should be reworked in bulk.

Alm999:
Oh, and unless I'm wrong, all that "plug-in" mechanic does is refill tool's built-in battery according to the "charge_rate" setting. It does not enable direct power consumption from storage, thus a tool won't work if it does not have charges in the first place no matter plugged-in or not.
For "direct battery power utilization" one still needs to have a dedicated "welding station" vehicle connected to storage with "extension cord" (or having a swappable storage battery case built-in directly).

Alm999:
It seems the buzzing function for tools is hardcoded. I've added some copy-paste into the code, but for now I shall really leave the code stuff to the code guys/girls. AFAIK, they wanna un-hardcode these anyways.

Issue #63362, opened 2023-01-31
Roadmap of the Hell's Raiders


MNG-cataclysm:
### Is your feature request related to a problem? Please describe.

I'm mainly making this as a way to keep track of my ideas for the Hell's Raiders, but also as a way for discussion, feedback, and brainstorming on expanding the faction as a whole. This is meant to be over the long term.

### Solution you would like.

Current Plan:

- [ ] More bandit-related locations and places for them to spawn, such as new forts, outposts, plantations, drug labs, etc.
- [ ] Some sort of NPC hierarchy, from slaves -> foot soldiers -> capos -> underbosses -> bosses -> commissioners. Their rarity would be based on their standing, with underbosses and bosses only showing up in heavily fortified locations. The top leadership may or may not appear at all in the game.
- [ ] Some meaningful way of joining the Hell's Raiders. Presently, I think of joining by referral through non-hostile members, such as spy rings or stranded foot soldiers. Joining could include shadowy dealings with unscrupulous merchants to gain favor with the faction, but it shouldn't be simple to do or completely RNG-based. 
- [ ] Missions from both common soldiers up to the top leadership. Lower-tier jobs from common individuals could just be a handful of the random missions we have now, while jobs from higher ranking members would be more meaningful escapades, with lots of risk. These could include attacking settlements to conquer them, raiding caravans from other factions, or skirmishing with rival bandits. Less violent jobs could include settling internal disputes, managing slave populations, or making an example of somebody.
- [ ] Some way of rising in status and rank. Currying favor with the higher-ups through dangerous heists or by simply doing lots of small work could give the player an in on new intel, ownership of several slaves, or the sanction to lead their own crew for the faction. Advancement high enough could gain the player operation of an entire fortress or drug lab, with the Raiders getting a portion of the proceeds. Advancement generally shouldn't come though killing someone of the rank you want; there's too few people like that in the world. However, a possible job for the purposes of gaining a title might include killing a wayward member of high status.

### Describe alternatives you have considered.

There's a bunch of different ways to do these sorts of things, but further discussion on which way would be the best should occur below. 

### Additional context

[...]

PR #63300, opened 2023-01-26
Expand the Hells' Raiders 

MNG-cataclysm:
#### Summary
Content "Expand the Hells' Raiders as a faction"

#### Purpose of change
Currently, the Hell's Raiders is the only NPC faction that cannot be interacted with in a non-hostile manner. This aims to change that.

#### Describe the solution
Make it possible to join the Raiders by talking to the handler of the spy at the Refugee Center. Give bandits and thugs speech topics. Do some redesigns of the bandit camp to add a storage area, two merchants, and a local leader. 

#### Describe alternatives you've considered
None.

#### Testing
I tested everything multiple times, and it all works great.

#### Additional context
Plan:

- [x] Give basic NPC's talk topics
- [x] Add traders
- [x] Add vehicle merchants
- [x] Add a leader of some sort
- [x] Lint all JSON
- [x] Test

Revamps/Maps:
- [x] Improve the bandit camp mapgen
- [x] Improve the bandit garage mapgen
- [x] Slaveholding forts

Some parts of the original plan have been moved to #63362 in favor of a future PR.

PR #67212, opened 2023-07-27
Expand the Hells Raiders: Round 2

MNG-cataclysm:
#### Summary
Content "Expand the Hells' Raiders"

#### Purpose of change
Despite the desire to join them not being considered good in #63300, I still think that the Hell's Raiders could use some expanding.

#### Describe the solution
Adds two new bandit-related locations, the slave farm and a drug lab, which contain a variety of loot and NPCs. Existing bandit locations have also been updated with some new mapgen and additional NPCs, to make them feel more alive and working. The larger locations also contain a local gang boss who runs the operations in that area and can be fought. None of these NPCs will talk to you however, as their dialogue has been cut from the last PR. These NPCs, including a gunrunner and quartermaster, bring with them new loot to these highly dangerous structures, providing a high risk-reward tradeoff for approaching them.

The ability to join the Hells Raiders has also been cut from the PR, though non-hostile interactions with the group may come back in the future.

This is the alternative.

#### Testing
I tested and loaded everything; it all works like a charm.

#### Additional context
This is some work on #63362, aiming to provide new content to the Raiders.


Procyonae:
All these locations should probably be globally unique, making the regular bandit maps you're changing a non globally unique version would be good too

MNG-cataclysm:
> Seeing how you mentioned all these NPC's are hostile, what about the slaves they are/were being held by the Hells Raiders? Can we 'the player' talk to them and free them sort of in the same way we can free prisoners from the Mi-Go's?

Yes, the player can talk to and free the slaves they encounter.

MNG-cataclysm:
> All these locations should probably be globally unique, making the regular bandit maps you're changing a non globally unique version would be good too

They're supposed to be regular bandit locations, I don't see why they would need to be globally unique.

Procyonae:
> They're supposed to be regular bandit locations, I don't see why they would need to be globally unique.

Nvm I must have been tripping earlier I thought the higher ups were unique NPCs

Daved27hundred:
I might be in the minority but I feel like the hell's raiders owning slave farms and slaves in general is kind of...dumb, is the best word I can use? Ignoring the touchiness of the subject it doesn't even feel like something they should be able to do.

They're not a kingdom like Ancient Egypt or Greece with the power and wealth behind them to support an operation like this, they're a group of thugs. I don't know how organized they are intended to be but something like that would be a lot of effort to maintain without the slaves either revolting or the leadership falling apart, especially considering there's ten slaves on that farm you have and only six men overseeing it.

I guess this is less a criticism of your PR and more one of the hell's raiders in general.

Procyonae:
> I might be in the minority but I feel like the hell's raiders owning slave farms and slaves in general is kind of...dumb, is the best word I can use? Ignoring the touchiness of the subject it doesn't even feel like something they should be able to do.
> 
> They're not a kingdom like Ancient Egypt or Greece with the power and wealth behind them to support an operation like this, they're a group of thugs. I don't know how organized they are intended to be but something like that would be a lot of effort to maintain without the slaves either revolting or the leadership falling apart, especially considering there's ten slaves on that farm you have and only six men overseeing it.
> 
> I guess this is less a criticism of your PR and more one of the hell's raiders in general.

I feel like it could be written better personally, like rather than them being captured slaves they could be peeps that don't want to risk life and limb so they bought the bandits protection in exchange for stuff they had and now do hard labour in exchange for continued protection?

I-am-Erk:
The narrative around hell's raiders and slaves is supposed to be a pretty complex one, and I pretty strongly feel that it's a very bad place to start with the faction. it sends the message that this is what they are, and at the very least, runs the strong risk of just turning them into a budget bin Caesar's Legion from New Vegas.

I'm not opposed to adding HR content. This just isn't a good place to begin it. Generally, if we were to add a HR base of some sort in the current game, we should probably start with a fortified encampment not to far off the other faction bases. They'd be hostile by default, but once in, we'd add a few quests that would allow you to get in and parlay with them.

Just like every faction, the Raiders should have elements that keep them from being universally 'the baddies'. There are things they're right about and in some ways they've got good points. Opening them up as "these are the guys who attack you and keep slaves" will prevent much development in that direction.

I-am-Erk (changes requested):
See comment above: I think the content here needs to be done a little more cautiously. In general, I think you have a tendency to choose some very complex factions to write about, and I strongly recommend you try to organize your thoughts in a planning issue before you do all the writing and stuff.

MNG-cataclysm:
So, for the time being, I've scrapped the slaves at the farm, and turned the crops into something more reasonable for the Raiders to grow.
@I-am-Erk #63362 outlines much of what I was planning, I just intend for this PR to add some locations instead of meaty NPC content or lore.

PR #67218, opened 2023-07-27
Uncanny dodge spell fix

### LINKED ISSUE #66100: UNCANNY_DODGE + Spell Effect SEGFAULT
### Describe the bug

When the character has the uncanny dodge effect, such as through the BIOTEK attunement from Magiclysm, any spells that buff the character, such as restoration or ogre's strength, cause an immediate SEGFAULT.

### Attach save file

N/A

### Steps to reproduce

1. Install Magiclysm
2. Debug BIOTEK attunement with Mutation menu
3. Debug Ogre's Strength spell
4. Cast Spell, crash.

### Expected behavior

No SEGFAULT

### Screenshots

_No response_

- OS: Windows
- Game Version: f9dc830 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm]
]

### Additional context

Even without the UNCANNY_DODGE flag enabled, the character still tries to dodge spell effects, visible via the dodge message as well as reduced stamina.  These don't cause SEGFAULTS, however, but the character shouldn't be dodging their own buff spells / effects.

[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/11710836/crash.log)

deepvelocity:
#### Summary
Bugfixes "Stop dodging good spells, fix uncanny dodge spell crash"

#### Purpose of change
Fixes #66100

Right now, game attempts to calculate damage mitigation values for any spell, including non-damaging ones and healing spells. Since dodge check actually triggers uncanny dodge, this would lead to moving to a different tile during the execution, then trying to apply effect to a tripoint (which character moved from). Result: segfault. 

#### Describe the solution
1. Add effects to the spell targets before dodge can happen 
2. Do not attempt to dodge if spell is not harmful

We no longer calculate mitigation values for non-damaging spells. This solves the dodge attempt issue. Spell effects are now applied before calculating the mitigation values, which solves the segfault. 

1. Disabling dodge calculations for spells completely.
2. Using effect rating (bad/good) to determine whether the spell should be attempted to dodge.

#### Testing
Local build with tests enabled successful. Launched the game, killed boomer with uncanny dodge enabled, no segfault (expected behavior).
Ran Magiclysm mod, casted Ogre's strength: no dodge attempt noticed.


AmarinReyny:
There are certain self-targeting spells that do cause damage to the player character, IIRC (I think one of the Magiclysm attunements, overclocker maybe, has those?). The "bad/good" alternative might still be useful to implement in such cases, or maybe have a "no dodge" flag for specific spells that self-damage.

deepvelocity:
@AmarinReyny would you mind providing the spell names for testing?

AmarinReyny:
> @AmarinReyny would you mind providing the spell names for testing?

I tried looking it up after posting that reply, and if I'm reading it correctly, the Overdrive spell from the Overclocker attunement self-damages. There are also at least two druid spells, Sacrificial Regrowth and Sacrificial Healing, which self-damage. That's just what I've found so far; I don't know if there are any more.

AmarinReyny:
Also, how would this interact with harmful spells, like what the Boomer does?

deepvelocity:
> Local build with tests enabled successful. Launched the game, killed boomer with uncanny dodge enabled, no segfault (expected behavior). Ran Magiclysm mod, casted Ogre's strength: no dodge attempt noticed.

To answer your question, boomers are the primary goal for me.
The spells you mentioned do not deal damage as the spell itself, so no issues with those. Just tested them all, no issues.

Looked into effects: there can be a scenario when multiple effects apply, with both negative and positive effects, and some things don't have any effects at all, making it less reliable for our purposes.

I am strongly against using "undodgeable" flag, as it would introduce another factor of complexity, leading to bugs in the future.

AmarinReyny:
> > Local build with tests enabled successful. Launched the game, killed boomer with uncanny dodge enabled, no segfault (expected behavior). Ran Magiclysm mod, casted Ogre's strength: no dodge attempt noticed.
> 
> To answer your question, boomers are the primary goal for me. The spells you mentioned do not deal damage as the spell itself, so no issues with those. Just tested them all.
> 
> Looked into effects: there can be a scenario when multiple effects apply, with both negative and positive effects, and some things don't have any effects at all, making it less reliable for our purposes.
> 
> I am strongly against using "undodgeable" flag, as it would introduce another factor of complexity, leading to bugs in the future.

Ah, sorry. I must have missed that part of testing that mentioned the boomer. I might need new glasses already... that, and sleep. Again, my apologies.

Good to know the spells I was worried about don't have any problems, though. Thank you for letting me know!

PR #67220, opened 2023-07-27
Integrated multitool fix

### LINKED ISSUE #67216: Integrated Multitool CBM needs power even on tasks using qualities that don't say they need power.
### Describe the bug

Attempting to pry the lid off of a crate with an integrated multitool gets interrupted with warnings that it's run out of charges, despite prying not being on the list of tool actions that would need power.

### Attach save file

[Langes Corners-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12187858/Langes.Corners-trimmed.tar.gz)

### Steps to reproduce

Use the Bionic Prepper profession, have zero power, find a crate, fail to pry it open.

### Expected behavior

I would expect that the integrated multitool would succeed at a task that doesn't require power without claiming power issues.

### Screenshots

image
image

- OS: Windows
- Game Version: 520db0c [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm],
    SpeedyDex [speedydex],
]

### Additional context

_No response_

deepvelocity:
#### Summary
Bugfixes "Integrated Multitool consumes power only when welding"

#### Purpose of change

Fixes #67216 

#### Describe the solution
From my research, it is not possible to specify charge use only for a specific action/quality. I decided to split integrated multitool into 2 items: multitool and mini welder (help with naming would be appreciated).

The weight and volume was split between the two, while only multitool retained armor values (so nothing gained, nothing lost from the change). 

1. Introduce another CBM with integrated welder.
2. Rework charge consumption and consume charges only for actions with qualities that require item being charged. This would be a big overhaul, so I refrained.

#### Testing

Local build with tests passed. Launched the game, had two tools: multitool and mini welder.


PatrikLundell:
My suggestion would be to call the welder part "integrated welder" instead of "mini welder".

deepvelocity:
@PatrikLundell the in-game name is integrated welder. Miniwelder is the name used for the item id. Since it's a part of integrated.json, using "integrated" seems wrong (multitool was called toolset, for instance). Maybe welder or internal_welder thou, thoughts?

anoobindisguise:
#67026 fixed this already though?

PatrikLundell:
Internal names aren't as important as the player facing ones, but if the integrated multitool is called "multitool", I'd suggest the integrated welder would be called "welder". It can also be noted that JSON doesn't care which file stuff is in (a bit like C), so the "integrated" part would only be visible to anyone looking at that file, but not anywhere else where the identifier is used.

deepvelocity:
> #67026 fixed this already though?

This is a partial fix, please try using any other multitool action: it will consume the same 20KJ for any other action. I tested locally and the issue persists.

deepvelocity:
> Internal names aren't as important as the player facing ones, but if the integrated multitool is called "multitool", I'd suggest the integrated welder would be called "welder". It can also be noted that JSON doesn't care which file stuff is in (a bit like C), so the "integrated" part would only be visible to anyone looking at that file, but not anywhere else where the identifier is used.
> 
> The reason I don't like the "mini" part is that it implies it behaves differently from a regular welder in some way (apart from being integrated, of course), presumably with some diminished capabilities, which I don't think is the case.

Agreed, though integrated multitool uses "toolset" for some reason. Pushed change for the id.

deepvelocity:
Update on id: I missed the fact that IDs are universal, so welder is not an available option. After some consideration, I decided to do "integrated_welder" to avoid other collisions. Input is welcome.

PR #67222, opened 2023-07-27
Bovine Bulk description edit

GalacticApple:
#### Summary
Balance "Adjusts Bovine Bulk description to portray it as a defensive mutation"

#### Purpose of change
This is a follow-up to a previous pull request where I thought that since the mutation described you as gaining lots of muscle, it should also make you stronger. However, CATTLE are already very, very strong, so that wouldn't work for balance reasons. 

For this request then, the changes made to the descriptions should indicate a general increase in toughness. Also, you now need Dense Bones to get the initial Bovine Bulk.

#### Describe the solution
JSON edits

It was mentioned before that this mutation could grant you a mutant fat integrated armor with high critical defense. 

#### Testing
Works on my CDDA copy


fairyarmadillo:
Adding the prereq just makes it easier to get extremely strong, which is undesirable as people have already been heavily using this mutation for easy access to +strength mutations. I'd suggest leaving it where it is or putting it behind dense bones instead.

fairyarmadillo:
> What about large or huge? These are the mutations where you really get bulked.

Huge is post-threshold. Moving the mutations to have it as a prereq would make them post-threshold as well, which would functionally create the same problem as sticking them behind extremely strong. They're also not at all impressive or even particularly interesting in the way post-threshold mutations generally are meant to be.

Honestly this all feels like a solution in search of a problem.

PR #67223, opened 2023-07-27
Jsonify road

Procyonae:
#### Summary
Content "Adds curved road bends, rotaries and rest stops/laybys"

#### Purpose of change

Hardcoded stuff bad

#### Describe the solution

Removes all regular road hardcoding (and sewer hardcoding that isn't doing anything bc LovamKicsiGazsii already unhardcoded them in #64793) and replaces it with JSON maps and nests

Doesn't touch bridges

Manholes are still partially hardcoded (Will deal with in a follow-up PR)

Generally I've stuck to how roads were already working down to exact chances but I have made a few changes:

- Singular _curved spawn as actual curves
<details>

image

image

image

</details>

this also fixes the diagonal removing the internal sidewalk like this

<details>
  <summary>Screenshot</summary>
  
image

</details>

- When road placement messes up and you get 4 adjacent road (not a new problem) it forms a rest stop/layby if suitable
<details>
  <summary>Screenshot</summary>

image
  
</details>

- Rotaries (roundabouts/traffic circles) with random centre nests rarely spawn instead of four_way s. 
<details>
  <summary>Screenshot</summary>
  
image

</details>

- Roads use fallback_predecessor now so cut through forests more smoothly and can even form intersections with trails (Could be made cooler in a separate PR)
<details>
  <summary>Screenshot</summary>
  
image

</details>

- Pileups are meant to spawn between 5 and 12 vehicles linearly, that logic would be way messier than making it binomial though so there's still a min and max of 5 and 12 but 8 is more likely than 5 or 12 for example.

- I had to add a small function to make LINEAR maps work with fallback predecessor because otherwise if say a bend is replaced with a tee then places one on top of the other, the function should make it so LINEAR maps only care about the bottom most predecessor but it doesn't work perfectly in cities (enable the food cart variant of 24x24_sidewalk_curved_curvy_outer and witness food carts spawning on other variants where it's been placed over)

Deletes ancient plaza and highway code:

- Plazas haven't spawned since 0.C bc the road formation they need to spawn hasn't been able to occur, and afaik they only existed to make those road formations look less dumb (I went back to 0.C and they're not very thrilling). I'm looking to a couple of city buildings that are plaza-esque, a city square and a pedestrianised mall but won't be replicating the old behaviour in any way. If connection code got changed to allow them to spawn again they could be added in JSON as conditional nests similar to how the layby works.

- Highways are almost a myth as they haven't spawned since pre 0.A and if they ever get reintroduced it should be as a separate JSON connection (I am playing with a highway mutable but even if I make it work I probably wouldn't merge it)

- Removing hardcoded manholes and instead spawning manholes under any road over a sewer with a % chance in exchange for losing the overmap symbol differences and known down flag

- Turning 4 adjacent road into a cooler intersection rather than a layby (or just removing the extra tile but then the overmap would be lying). I'm not a big fan of the layby but hopefully it encourages someone to fix the connection code so that it can't occur in the first place :^) (I will make a follow up PR adding proper 1 lane road nests for the tees to connect to it in a follow up PR I just didn't want to add more conditional nest bloat)

- Splitting the road palette into a basic one and a decorative one to use alongside it

- Adding a t_region_groundcover border to the road for cutting through forest/swamp

- Making the different curvy_outer sidewalk nests nested nests instead of variants

- Leaving the vehicle placement alone as it was already partially unhardcoded, but eventually that whole system should be obsolete anyway

- Mimicking the current method, the straight/tee/four_way tiles before diagonals get extra road to start the diagonal wider which sucks imo bc it makes loads of extra edge cases for future expansion on road connections and intersections with other connections. I tried multiple different ways to make diagonals that didn't require this, see commit history or d3acd93c75 (final attempt that was probably objectively best) for my attempts if you want to have a crack at this yourself or make suggestions
<details>
  <summary>Screenshot</summary>

BadBendB
  
</details>

- Changing the workbench side boards on the food + taco trucks to a shutter window or something so the layby one can actually get some custom (Might do in another PR)

- Adding more furniture and stuff to sidewalks but I'll save that for a separate PR

- Vehicle placement could generally be more fitting and more interesting with more angles and such but it's a PITA to get all the coords right and spawning vehicles over each other to form wrecks is semi unreliable as is, will make a follow up PR to add bend/rotary specific spawns at the very least

- The way the upper half road_vehicles_nested file is structured is questionable as opposed to putting all the weights in a single nest like I did with No Hope but I'm leaving it this way until I've dealt with additional spawns as described above

- If vehicle groups worked as mapgen value parameters and/or if random chunks could have independant coords/coord offsets that could save quite a few vehicle placement nests each

- Idk if I'm just derping but I wanted logic so that roads that double back to connect to overmap specials would be curved on the end but while I thought I'd figured it out I can't seem to distinguish between these two scenarios now where the blue is the overmap special, pink is a curve and purple is a diagonal with the awkward restraints of not being able to tell what orientation the bend in question is or some flag applied to all overmap special connection tiles so I've scrapped double back logic entirely.
<details>
  <summary>Screenshot</summary>
  
image

</details>

- The way I handled No Hope's vehicle placement results in some regression bc I didn't bother applying their unique more segmented placement across each tile

#### Testing

Link to the exe to save peeps having to compile (ignore the source code stuff, just checkout this PR locally and use the exe) https://github.com/Procyonae/Cataclysm-DDA/releases/tag/JsonRoad

Someone who knows how should probably run a few performance tests before this gets merged

Spawned a ton of roads
Checked (lack of any) migration works using several stable saves

#### Additional context

Removes a "road" palette that was only used for a single map


fairyarmadillo:
Out of curiosity, will this let us put map extras on road tiles in town? That would be really nice.

Procyonae:
> Out of curiosity, will this let us put map extras on road tiles in town? That would be really nice.

I'll be real I hadn't even realised that was an issue, so I have no clue ;D
If they do there'd be no way to differentiate between in city road and out of city road though, I'll make sure to test that at some point (Road map extras have been spawning normally though I've not broken them)

data/json/vehicle_groups.json
```diff
@@ -88,6 +88,15 @@
...
     ]
   },
+  {
+    "type": "vehicle_group",
+    "id": "food_trucks",
+    "vehicles": [ 
+      [ "icecream_truck", 10 ],
+      [ "food_truck", 10 ],
+      [ "food_truck_taco", 10 ],
+      [ "wienermobile", 1 ] ]
```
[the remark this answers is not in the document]
Procyonae: Too slow bot

tests/mapgen_function_test.cpp
```diff
@@ -1,8 +1,60 @@
...
+        // East
+        case 1:
+            if( there == oter_sewer_ew || there == oter_sewer_ne || there == oter_sewer_es ||
+                there == oter_sewer_nes || there == oter_sewer_new || there == oter_sewer_esw ||
+                there == oter_sewer_nesw ) {
+                return true;
+            }
+            return false;
+        default:
+            debugmsg( "Connects_to with dir of %d", dir );
```
Procyonae, line 53:
```suggestion
```

tests/mapgen_function_test.cpp
```diff
@@ -1,8 +1,60 @@
...
+static const oter_str_id oter_sewer_ne( "sewer_ne" );
+static const oter_str_id oter_sewer_nes( "sewer_nes" );
+static const oter_str_id oter_sewer_nesw( "sewer_nesw" );
+static const oter_str_id oter_sewer_new( "sewer_new" );
+static const oter_str_id oter_sewer_ns( "sewer_ns" );
+static const oter_str_id oter_sewer_nsw( "sewer_nsw" );
+static const oter_str_id oter_sewer_sw( "sewer_sw" );
+static const oter_str_id oter_sewer_wn( "sewer_wn" );
+
+bool connects_to( const oter_id &there, int dir )
```
Procyonae, line 17:
```suggestion
static bool connects_to( const oter_id &there, int dir )
```

Procyonae:
> Out of curiosity, will this let us put map extras on road tiles in town? That would be really nice.

image

Extras do spawn in cities now

fairyarmadillo:
I don't want to clutter up your lovely PR, but pretend I posted the guy standing behind a tree rubbing his hands together excitedly meme.

Zireael07:
YAAAY!

sparr:
@Procyonae I want to revive #61342 which will get us gentle curves outside of cities, but it has a lot of conflicts with your work here. I tagged you on Discord, can we chat/correspond before I dive in?

PR #67230, opened 2023-07-28
Stop aphids eating their own honeydew

worm-girl:
#### Summary
Bugfixes "Stop aphids from eating their own honeydew"

#### Purpose of change

Aphids were eating their own honeydew as soon as they produced it, because they have EAT_FOOD. This was not a new issue, but nobody ever noticed it because EAT_FOOD was completely not working for a very long time.

#### Describe the solution

Removes EAT_FOOD from aphids. They have EAT_CROP, so they're fine. Real-life aphids don't eat your food, they just eat live plants.

Make EAT_FOOD ignore honeydew somehow. This would be undesirable as ants should be eating honeydew via EAT_FOOD.

Added messaging to EAT_FOOD. I had already added it in #66294 but I forgot to display it in some cases.

#### Testing

Spawned some aphids and ants. Tossed some honeydew around. The aphids ignored it and the ants ate it.

### DIFF
--- a/data/json/monsters/insect_spider.json
+++ b/data/json/monsters/insect_spider.json
@@ -2578,7 +2578,7 @@
     "proportional": { "hp": 0.5, "speed": 1.1, "morale": 0.67, "melee_dice_sides": 0.5 },
     "volume": "96 ml",
     "weight": "96 g",
-    "special_attacks": [ [ "EAT_FOOD", 400 ], [ "EAT_CROP", 600 ] ],
+    "special_attacks": [ [ "EAT_CROP", 600 ] ],
     "delete": {
       "reproduction": { "baby_monster": "mon_aphid", "baby_count": 1, "baby_timer": 20 },
       "baby_flags": [ "SPRING", "SUMMER" ]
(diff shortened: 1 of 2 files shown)


Fris0uman:
We sure that aphids don't eat the dew irl?

fairyarmadillo:
> We sure that aphids don't eat the dew irl?

> Xylem sap consumption may be related to [osmoregulation](https://en.wikipedia.org/wiki/Osmoregulation).[[37]](https://en.wikipedia.org/wiki/Aphid#cite_note-Pompon-38) High osmotic pressure in the stomach, caused by high sucrose concentration, can lead to water transfer from the hemolymph to the stomach, thus resulting in hyperosmotic stress and eventually to the death of the insect. Aphids avoid this fate by osmoregulating through several processes. Sucrose concentration is directly reduced by assimilating sucrose toward metabolism and by synthesizing [oligosaccharides](https://en.wikipedia.org/wiki/Oligosaccharide) from several sucrose [molecules](https://en.wikipedia.org/wiki/Molecule), thus reducing the solute concentration and consequently the osmotic pressure.[[39]](https://en.wikipedia.org/wiki/Aphid#cite_note-40)[[40]](https://en.wikipedia.org/wiki/Aphid#cite_note-41) Oligosaccharides are then excreted through honeydew, explaining its high sugar concentrations, which can then be used by other animals such as ants.

It's a waste product that they excrete so they don't die from eating a diet of almost pure sugar. I can find no mention anywhere of them eating it.

Fris0uman:
Nice, okay this makes sense to me.

anoobindisguise:
Wonderful, now I can resume eating the tasty bug poop.

PR #67239, opened 2023-07-28
Unhardcode mx_clearcut

Procyonae:
#### Summary
None

#### Purpose of change

Hardcoded stuff bad

#### Describe the solution

Should work identically

#### Testing

image
(Boosted rate)


NetSysFire (approved):
Fancy.

Procyonae:
Shame "set"'s "square" doesn't work with ter_furn_transforms bc then it wouldn't even need to define rows

PR #67240, opened 2023-07-28
make bone inedible

ampersand55:
#### Summary
Balance "Make bone inedible"

#### Purpose of change

Currently you can eat a whole raw 2.25 kg 40 cm bone in 50 seconds with no enjoyment penalties or toxins for 96 % calcium RDA. This seems unrealistic.

#### Describe the solution

Set material "bone" to inedible in json.

N/A

#### Testing

Loads and works as expected. Bone is inedible, but bone meal tablets can be eaten.

#### Additional context

N/A

### DIFF
--- a/data/json/materials.json
+++ b/data/json/materials.json
@@ -259,7 +259,7 @@
     "specific_heat_liquid": 1.2,
     "specific_heat_solid": 1.2,
     "latent_heat": 200,
-    "edible": true,
+    "edible": false,
     "rotting": true,
     "chip_resist": 8,
     "wind_resist": 90,


sadenar:
Could you make it edible to post tresh Lupine, Beast, Ursine, Feline and Reptile? That are lock it behind specific jaw mutations? I agree a normal human shouldnt just be able to munch bone, but it would not make a lot of sense for your crocodile jaw to struggle with it.

fairyarmadillo:
> Could you make it edible to post tresh Lupine, Beast, Ursine, Feline and Reptile? That or lock it behind specific jaw mutations? I agree a normal human shouldnt just be able to munch bone, but it would not make a lot of sense for your crocodile jaw to struggle with it.

These animals are capable of swallowing bones and that should be accounted for, but I don't think they digest them for 94% of the calcium they need. They just pass them, right?

Currently if the material is set as inedible, nobody can eat it. You can instead leave it edible, but add the INEDIBLE flag to all items made of bone. If you wanted to let lupines eat them, you'd add the LUPINE flag to the item as well.

ampersand55:
> These animals are capable of swallowing bones and that should be accounted for, but I don't think they digest them for 94% of the calcium they need. They just pass them, right?

The whole thing is messy. 

* Bone as a material has a calcium content of 2% per 250 ml.
* A bone is 2500 ml, so it should have had a calcium content of 20%, but instead it has a calcium content of 96%. It's currently edible. 
* Bone meal made from bone has a calcium content of 0 %.  It's currently inedible. 
* Bone meal tablets has a calcium content of 24%, which does match the 96% from bone as you get 1 bone meal tablet per bone meal and 4 bone meals per bone.

There's no nutrient inheritance if calcium in either of them. 

That we can't eat bone marrow is consistent with other powdered ingredients. We also can't eat e.g. sugar, flours, salt or spices in their raw form even though they're obviously consumable.

PR #67246, opened 2023-07-28
Fix kevlar requirements of survivor gear crafts

anoobindisguise:
#### Summary
Bugfixes "kevlar jumpsuits use the correct requirement"

#### Purpose of change
Fixes #a-complaint-on-reddit
Survivor gear was using only regular sheets, rather than patchwork sheets OR regular sheets.

#### Describe the solution
Change requirements to be either.
Also added many deconstruction recipes for various kevlar stuff. This is roughly their weight in sheets, with a higher time to represent the careful deconstruction. 
Fixed a couple kevlar items that were wrong. Chainsaw Chaps were 1mm while their weight suggested they were 3mm thick, and 3mm was how thick the leather chaps were too so I changed them to be 3mm. Cut resistant sleeves were too light, being the weight of 0.5mm while being 2mm thick. I quadrupled its weight so that it properly matches the weight of 2mm thick kevlar sleeves.

Get rid of the lycra underlayer and continue to tweak survivor suit balance in general.


a-chancey:
Values look good to me!

PR #67254, opened 2023-07-28
AIM: Fix some instances of stack changing list positions when moving items

Kamayana:
#### Summary
Bugfixes "AIM: Fix some instances of stack changing list positions while moving items"

#### Purpose of change
In the AIM, sometimes when moving items around the selected item stack would inexplicably change positions on the list, making you now select a different item and potentially leading to mistakes. This only occurs when you move one item of a selected item stack of 2 and there's a similar item listed next to it (like a battery with a different charge count).

#### Describe the solution
One cause of this problem is that `advanced_inv_listitem.name` uses `tname()` with a variable `quantity`. For example, a stack of two medium batteries is internally sorted as "medium batteries", which is alphabetically listed before "medium battery", so when the stack count changes from 2 to 1 it also changes its sorting order, making it move around. This plural string is not actually visible to the user at all, as `advanced_inv_listitem.name` is used only for sorting and the AIM itself doesn't use plurals, so there's no reason to use a plural name for sorting. The fix is to always use a quantity of 1 for the sort name.

This PR doesn't fix every instance of items changing list positions; it can also be caused by the internal order of the items. For example, if you move four batteries of differing charges to a location in this order: 500, 499, 500, 499, then the 500 charge battery will still change positions underneath the 499 as they are moved. Fixing that would take a lot more work, though, and I wanted to just fix one specific instance of the issue with this PR.

Items with charges use a different `advanced_inv_listitem.name` assignment that also sends stack count to `tname()`, so I was going to change that too, but on testing it seems like charge items never use the plural name anyway - a stack of nuts and bolts always uses the sort string of "nut and bolt" no matter how many charges are in the stack - so I left it alone.

This PR will actually be necessary for #66871, so cables can be sorted properly in the AIM. I didn't want to stuff this bugfix into an already-large PR, though.

#### Testing
Didn't see any weirdness when moving things around.

#### Additional context
image

### DIFF
--- a/src/advanced_inv_listitem.cpp
+++ b/src/advanced_inv_listitem.cpp
@@ -28,7 +28,7 @@ advanced_inv_listitem::advanced_inv_listitem( const std::vector<item_location> &
     area( area ),
     id( list.front()->typeId() ),
     items( list ),
-    name( list.front()->tname( list.size() ) ),
+    name( list.front()->tname( 1 ) ),
     name_without_prefix( list.front()->tname( 1, false ) ),
     autopickup( get_auto_pickup().has_rule( & * list.front() ) ),
     stacks( list.size() ),


Kamayana:
Looks like this isn't as simple as I thought, unfortunately.

PR #67255, opened 2023-07-28
Player activity related event

lispcoc:
#### Summary
Features "Player activity related event"

#### Purpose of change
Makes using player activity more advanced from EOC. 

#### Describe the solution
Add following player activity related events.
`character_starts_activity`
`character_finished_activity`

Monitoring player activity from EOC every turn. It takes extra load and makes processing become unnecessarily complicated.

#### Testing
`character_starts_activity` and `character_finished_activity` events are captured from EOC.

### DIFF
--- a/data/mods/TEST_DATA/EOC.json
+++ b/data/mods/TEST_DATA/EOC.json
@@ -330,6 +330,28 @@
   },
   {
     "type": "effect_on_condition",
+    "id": "EOC_character_finished_activity_event_test",
+    "eoc_type": "EVENT",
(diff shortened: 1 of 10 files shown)


Maleclypse:
So if you aren't adding something that uses the feature right now can you add a test to confirm that it works so that when json only people start using it they can reliably know that it works?

lispcoc:
> So if you aren't adding something that uses the feature right now can you add a test to confirm that it works so that when json only people start using it they can reliably know that it works?

OK, I added I simple tests.

PR #67258, opened 2023-07-29
Adjust charcoal production from filled charcoal kiln

Numbuh474:
#### Summary
Bugfixes "apply new energy audit to filled charcoal kiln"

#### Purpose of change

The filled charcoal kiln still only provides 200 charcoal on disassembly, which is not up to date with the current charcoal production/consumption values from #62180.

#### Describe the solution

Increase value to 10k charcoal units, which is the same amount produced by the metal charcoal kiln furniture item for 50L of wood (1 log).

N/A

#### Testing

Provides correct amount on disassembly.


Maleclypse:
@anoobindisguise Can you review and approve this if it looks good?

Issue #67029, opened 2023-07-19
Amphibian Additions


TheSaddestGoomba:
### Is your feature request related to a problem? Please describe.

Amphibians feel lacking as both monsters and natural wildlife. The natural varieties are lackluster compared to the diversity of birds and fish in game. Unlike those groups, amphibians have mutant monsters but those also feel under-developed.

### Solution you would like.

Add more specific natural species that function like wildlife and expand mutant varieties beyond size-growth.

### Describe alternatives you have considered.

Let the bugs rule the swamp.

Perhaps splitting the first phase into two. Starting with natural species and then handling frog/toad mutants as a second stage.

### Additional context

Here's what I've planned out so far.
Add in regional natural species that lay their own eggs, hatch into their unique larval forms and grow into adults.
Natural species will also rarely lay unusual eggs.
I'd aim to start with two natural species from each of these broad groupings: frogs, treefrogs, toads, and salamanders.
Additional natural species should be rather simple to implement thereafter.

Mixed in with the natural species are unusual ones (strange bullfrog, odd toad, etc.) that will grow into mutant and monstrous species.
These mutants will then "breed true", laying exclusively unusual eggs of their own at varying stages.
These mutants can zombify upon death and I plan to add a variety of new zombie forms for them.
Burned zombie versions are simplified down to three, assigned based mostly upon body size.

Frogs

- Bullfrogs will remain largely as-is, with their mutants occupying the ever-increasing-size niche.
- Giant Bullfrogs lay Giant Eggs which hatch into Giant Tadpoles. Most become Giant Frogs and continue the cycle. Others will grow into larger tadpoles as big as alligator gar. The oldest of these pack hunters may grow larger still.
- The new species will be Leopard Frogs and a mutant strain that develops electric eel like bioelectricity. Staticskins cause moderate shocks on contact, while Stuntongues can launch their tongue like a taser. Their shared zombie is the Shocker Frog.
- Both frog lines share a new zombie species for their smallest versions, the Frog Jockey. Frog Jocks can partly fuse with other zombphibians granting them rapid regeneration and possibly reduced cooldowns.

Toads
- Fowler's Toads and Eastern American Toads are the choice natural species.
- The existing size-growing toad line will be associated with the American Toad and will be given a 'mega' evolution. The Warted Monstrosity is comparable to the Colossal Croaker but the intent is to develop it into its own niche/combat style. 
[...]

PR #67259, opened 2023-07-29
Amphibian Additions - Phase 1 - Non-Monstrous Frogs, Toads, and Treefrogs

TheSaddestGoomba:
#### Summary
Content "Add diversity to amphibian species to reflect other animal varieties in game and set space to expand monstrous amphibian mutants (and their zombified remains)."

#### Purpose of change
Add more specific natural species that function like wildlife and expand mutant varieties beyond size-growth.
Expounded on and first implementation of #67029 

#### Describe the solution
Adds 6 new frog species, a strange (non-monstrous) version of each, and tadpoles and eggs for all 12 varieties.
The strange versions are for flavor and potential future development where they will progress into more mutated species.

~~Removing the existing frog, toad, and tadpoles. I still intend to eventually. My previous attempt to work them into this PR broke something so I simply worked around them and have left them for the time being.~~
This is done in the newest commit. Everything looks clean from my initial test.

~~Spawns and group weights still need a lot of testing and adjustment.~~
These have been updated and I'm happy with the current balance.

Create a weakpoints set for frog-a-forms and/or unique harvest results/items/descriptions. Neither seems very relevant with these little fellers. I expect to add them in the future with mutants the player might actually fight.

#### Testing
Created a world and character, teleported around checking spawns. Frog and tadpole varieties appear to be spawning in desired locations. Old entities are completely absent and the number and placement of new species feels appropriate.


data/json/monsters/reptile_amphibian.json
```diff
@@ -471,5 +471,269 @@
...
+  },
+  {
+    "id": "mon_shift_tadpole",
+    "type": "MONSTER",
+    "copy-from": "mon_gray_tadpole",
+    "name": { "str": "tadpole" },
+    "upgrades": { "age_grow": 60, "into": "mon_shift_frog" }
+  },
+  {
+    "id": "mon_proxy_frog",
```
Karol1223, line 605:
```suggestion
    "abstract": "mon_proxy_frog",
```
This should be an abstract instead of a full monster

data/json/monstergroups/eggs.json
```diff
@@ -495,5 +495,75 @@
     "name": "GROUP_EGG_FALLFISH",
     "type": "monstergroup",
     "monsters": [ { "monster": "mon_fry_fallfish" } ]
+  },
+  {
+    "name": "GROUP_EGG_PROXY_FROG",
+    "type": "monstergroup",
+    "monsters": [ { "monster": "mon_proxy_tadpole" } ]
+  },
```
Karol1223, line 503: Is this intended to exist? ``proxy_frog`` looks to be your abstract, why does it have functional egg mongroups?

data/json/monsters/reptile_amphibian.json
```diff
@@ -471,5 +471,269 @@
...
+  {
+    "id": "mon_bullfrog_frog",
+    "type": "MONSTER",
+    "copy-from": "mon_proxy_frog",
+    "description": "The American bullfrog, in its natural habitat.  It feeds on insects, mice, lizards and any other living thing it can stuff down its gullet.",
+    "name": { "str": "bullfrog" },
+    "volume": "750 ml",
+    "weight": "500 g",
+    "hp": 4,
+    "reproduction": { "baby_egg": "egg_bullfrog_frog", "baby_count": 16, "baby_timer": 60 }
```
Karol1223, line 657: We already have a bullfrog ingame. ``mon_frog``. You're making a duplicate

Karol1223 (changes requested):

data/json/monsters/reptile_amphibian.json
TheSaddestGoomba: Commited, I also changed the tadpole proxy to match.

data/json/monsters/reptile_amphibian.json
```diff
@@ -471,5 +404,270 @@
...
+    "hp": 2,
+    "reproduction": { "baby_egg": "egg_gray_frog", "baby_count": 8, "baby_timer": 26 },
+    "baby_flags": [ "SPRING", "SUMMER" ],
+    "extend": { "flags": [ "CAMOUFLAGE" ] }
+  },
+  {
+    "id": "mon_shift_frog",
+    "type": "MONSTER",
+    "copy-from": "mon_gray_frog",
+    "description": "You've seen it wrinkle its skin into bark and disappear on a tree; settle on a toadstool and take its spots.  It's no trick of the light, this frog's skin shifts not just color but texture.",
```
Karol1223, line 669: Is this supposed to be like how octopus camouflage works? I initially typed this thinking you went too deep into the nether territory but if it is supposed to be octopus-esque I think it's fine as a mutation. The only issue I have with this and all the other weird frogs is they produce non-mutant meat. They likely should not, even if they're just mild mutants.

data/json/items/comestibles/egg.json
```diff
@@ -1169,5 +1151,105 @@
     "material": [ "egg", "veggy" ],
     "fun": 5,
     "flags": [ "EATEN_HOT", "FREEZERBURN" ]
+  },
+  {
+    "type": "COMESTIBLE",
+    "id": "egg_proxy_frog",
```
Karol1223, line 1157:
```suggestion
    "abstract": "egg_proxy_frog",
```

I'm assuming this is supposed to be an abstract?

data/json/items/comestibles/egg.json
```diff
@@ -1169,5 +1151,105 @@
...
+    "copy-from": "egg_chicken",
```
Karol1223, line 1158: This is both good and bad. You're copying stuff from an egg which makes sense - but I am 99% sure caloric contents would be different and barter value should definitely be lower. If anything I'd advise copying from roe and adjusting price to be lower. Also, while I'm here, are you sure those are safe to consume? I vaguely recall hearing toad eggs specifically were poisonous?

Karol1223 (changes requested):

Karol1223 (approved):

Karol1223:
Not entirely sure what's going on with test failures but it looks like you have issues with the mods since you removed normal frogs and toads. On top of my head this breaks Tameable Wildlife, I'd assume TropiCata is affected as well but you should double check all the mods to see if they don't reference things you just removed.

On top of that, and I only realized it now, you're removing the eggs instead of obsoleting them which will cause errors on any saves that have those items. Either migrate them or obsolete them, please.

TheSaddestGoomba:
Sorry about that. The newest commit restores all of the deleted entries flagged and sorted for obsoletion. If I understand correctly, mods will still be able to call to the entries from the obsoletion folder. That should fix any issues of dead references, though Tamable Wildlife would need a future PR for the new species.
The old entities have already been removed from all monstergroups in mainline so shouldn't spawn there.
I think that should leave everything sorted now.

Karol1223:
Don't obsolete monsters. We don't do that. They'll just turn to breathers silently.

TheSaddestGoomba:
Ah ok, monsters were listed among other categories in the documentation. That would explain why there were none already in the folder. So is the solution for monsters to simply remove them from spawning? If so, I can remove the obsolete flag from the 4 monsters and move them back to reptiles_amphibians.json and that should fix the mod issues.

To clarify, I'm referring to making the new creatures tamable in a future PR, not fixing the errors. I wasn't assuming someone else would do that, I intend to.
I don't mind including the mod integration in this one, and I don't find it annoying; it will simply take more time.
I do however think it's a large barrier to contributing if any new creature addition is expected to also be integrated into mod features out the gate.

Karol1223:
> Ah ok, monsters were listed among other categories in the documentation. That would explain why there were none already in the folder. So is the solution for monsters to simply remove them from spawning? If so, I can remove the obsolete flag from the 4 monsters and move them back to reptiles_amphibians.json and that should fix the mod issues.

Just remove the monsters entirely. No obsoletion, no moving around. Just remove them completely - them not needing obsoletion is a relatively new thing. If they're unused you should definitely not keep them around, much less so in a normal file.

Realistically you should just remove those and fix the mod conflicts *in the mod files* and then it should be good to go. Keeping unused content in mainline to fix mod errors isn't the way to go.

TheSaddestGoomba:
> Just remove the monsters entirely. No obsoletion, no moving around. Just remove them completely - them not needing obsoletion is a relatively new thing. If they're unused you should definitely not keep them around, much less so in a normal file.

> Realistically you should just remove those and fix the mod conflicts _in the mod files_ and then it should be good to go. Keeping unused content in mainline to fix mod errors isn't the way to go.

> Integration isn't necessary for anything, I am only talking about fixing the mod conflicts. Tameable Wildlife has its own monsters that copy from vanilla monsters you removed. You should either remove them from Tameable Wildlife as well or make them copy from your replacements you introduced here. Having said that - other mods have errors, too. I think TropiCata animals copy from the amphibians you removed as well. Hence your test errors. 

Ok, I've done all of that that in the latest commit. Old monsters have been removed entirely.
I believe I got everything in TropiCata and Tamable Wildlife updated with new references. Megafauna also had some references to mon_frog in monstergroups.

> Integrating things into mods isn't a necessity, I think we just had a misunderstanding in what we were talking about here.

I think so, sorry for my part in it.

data/mods/Megafauna/monstergroups/wilderness.json
```diff
@@ -2117,22 +2117,23 @@
         "conditions": [ "SPRING", "SUMMER", "AUTUMN" ]
       },
       {
+        "group": "GROUP_FROGS_ALL",
         "monster": "mon_frog",
```
Karol1223, line 2121:
```suggestion
        "group": "GROUP_FROGS_ALL",
```
I'm assuming you missed that one?

Karol1223 (changes requested):

Karol1223 (approved):
Looks good to me so if the tests pass this looks acceptable

Karol1223:
Right that's an awkward test to fail. Your PR Summary isn't written correctly. You only have the content in ``"``, without the word prefix you gotta choose from

TheSaddestGoomba:
Ok, I think that's all that was needed

Karol1223 (approved):

PR #67261, opened 2023-07-29
Two new civilians: parent and hospitalized

NetSysFire:
#### Summary
Content "Two new civilians: Parent and hospitalized varieties"

#### Purpose of change

Flavor and making it seem a bit more real.
We already have the scenario of being left behind and there are plenty of people who were hurt due to or during the Cataclysm, some severely enough to need intensive care.

And people looking for left behind loved ones are also a scenario I consider quite plausible, thus the new parent civilian.

#### Describe the solution

- Hospitalized variant:
  - Blind, deaf and immobile, will always upgrade to a crawling zombie.
  - Can only be found in hospitals and nursing homes for obvious reasons.
- Parent variant:
  - Standard civilian, basically a very slightly modified panicked person.
  - Why would people not get to safety? Because they are not leaving without their loved ones.
  - I kept the dialogue intentionally "vague" because there is no way to store misc per-monster variables and they would otherwise randomly switch between e.g having a son and a daughter.

Not using these silly alliterations.

There is a tiny issue I found: Some of the hospital beds are actual 2x1 `f_bed` and there is a small possibility that two of them may spawn on the same bed. However, this is not worth the effort to fix imo.

#### Testing

Applied changes locally. Also, thank you RNG for producing this dark piece:

image

### DIFF
--- a/data/json/mapgen/hospital.json
+++ b/data/json/mapgen/hospital.json
@@ -194,6 +194,7 @@
           { "item": "surgery", "chance": 60 }
         ]
       },
+      "monster": { "B": { "monster": "mon_civilian_icu", "chance": 10 } },
       "place_monster": [
         { "group": "GROUP_HOSPITAL", "x": [ 0, 23 ], "y": [ 0, 23 ], "repeat": [ 15, 20 ] },
         { "group": "GROUP_HOSPITAL", "x": [ 25, 46 ], "y": [ 1, 22 ], "repeat": [ 15, 20 ] },
--- a/data/json/mapgen/nursing_home.json
+++ b/data/json/mapgen/nursing_home.json
@@ -192,6 +192,7 @@
           { "item": "magazines", "chance": 20, "repeat": [ 0, 1 ] }
         ]
       },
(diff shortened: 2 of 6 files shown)


Zireael07:
Having all the hospitalized humans be blind and deaf is a bit over the top. I'd randomize the chances for those two but keep the immobile tag.

NetSysFire:
This is for technical reasons only. Otherwise the monster will be in the fleeing state, which I want to avoid. Randomizing it is not possible afaik.

anoobindisguise:
It seems like the parent NPCs are really common. I see them everywhere even in early summer despite the panicked person being long gone. I can make an issue for this if you want but I don't have access to the game to make a version report right now.

Maleclypse:
> It seems like the parent NPCs are really common. I see them everywhere even in early summer despite the panicked person being long gone. I can make an issue for this if you want but I don't have access to the game to make a version report right now.

Please make an issue so discussion can happen on the issue.  Merged PRs are invisible to most people working on the game.

NetSysFire:
Generally true but I found the issue, missing "ends" parameter. I'll fix this real quick.

PR #67262, opened 2023-07-29
Remove leather material from tourniquet

Holli-Git:
#### Summary
Bugfixes "Change Tourniquet Materials"

#### Purpose of change

Realized to my horror that vegan characters can't use tourniquets cause it's made of leather. Even though majority of the ingredients are cotton, and that most IRL are nylon or similar, in game they're made with leather and wood. 

#### Describe the solution
Removes the leather material, replaces it with cotton.

Make a version that's not handmade that is standard plastic and nylon and the handmade one with wood and cotton. Or just not do that and have it as wood and nylon.

#### Testing
Tested. Works fine


Drew4484:
Rather than replace, why not make a separate version?

Zireael07:
Why would veganism prevent using leather TOOLS (as opposed to foods)?

Karol1223:
> Why would veganism prevent using leather TOOLS (as opposed to foods)?

I believe the intention is that the trait doesn't allow you to wear animal-product-clothing (leather definitely, I think chitin and wool may fall under this as well?) and the tourniquet is used by... wearing it. So this kinda fell apart.

Maleclypse:
This also breaks the density tests for the tourniquets.

Paprika-H:
> Why would veganism prevent using leather TOOLS (as opposed to foods)?

Because it's a dead animal. Vegans are against any sort of animal products. Also that's just how I coded it, can't exactly change that

Paprika-H:
> Rather than replace, why not make a separate version?

One that's made with a leather belt and one thats not?

Karol1223:
> This also breaks the density tests for the tourniquets.

Correction - this FIXES the density tests for the tourniquets. Just remove them from the known bad density list and it should be good to go, looks like

I-am-Erk:
I'd love to merge this as soon as that test thing is fixed :)

PR #67275, opened 2023-07-29
Solar Farm

ashGlaw:
#### Summary
Content "Solar Farm"

#### Purpose of change

Solar farms are all over New England, and solar panels are much harder to find in game than is realistic.

#### Describe the solution

I made a single tile map with solar panels and some electrical equipment, surrounded by a chain link fence. Technician zombies can spawn around it.

Making a much, much larger solar farm, with hundreds of panels. Maybe later

#### Testing

I got the location spawning in game and teleported to it. It's oriented correctly and everything. Monsters spawn. Doesn't crash the game.

#### Additional context

My first mapgen foray. Totally up for constructive criticism.
image


data/json/mapgen/solar_farm.json
```diff
@@ -0,0 +1,49 @@
...
+        "...------------++++--...",
+        "........................"
+      ],
+      "terrain": {
+        "*": "t_concrete",
+        "+": "t_chaingate_c",
+        "-": "t_chainfence_h",
+        "|": "t_chainfence_v",
+        ".": "t_grass",
+        "#": "t_grass",
```
Procyonae, line 41:
```suggestion
        ".": "t_region_groundcover",
        "#": "t_region_groundcover",
```

ashGlaw: Awesome! Thanks

sadenar:
Are molten salt solar farms a thing at all in NE?

ashGlaw:
Not that I'm aware of, no

data/json/overmap/overmap_special/specials.json
```diff
@@ -3503,6 +3503,18 @@
...
   },
+  {
+    "type": "overmap_special",
+    "id": "solar_farm",
+    "overmaps": [ { "point": [ 0, 0, 0 ], "overmap": "solar_farm_north" } ],
+    "connections": [ { "point": [ 0, 1, 0 ], "terrain": "road" } ],
+    "locations": [ "land" ],
+    "city_distance": [ -1, 2 ],
+    "city_sizes": [ 4, -1 ],
+    "occurrences": [ 0, 4 ],
```
Maleclypse, line 3514:
```suggestion
    "occurrences": [ 0, 2 ],
```

Maleclypse: Given the size of our overmaps this is probably still too common but it's a good number for testing.

fairyarmadillo:
My only suggestion is to add a few variants with missing or damaged solar panels to suggest a lack of maintenence and possible looting/vandalism/mindless monster damage since a bit before the Cataclysm began.

PatrikLundell:
Are there really mismanaged solar farms in the area? Firstly, they should all be comparatively new, and secondly, it doesn't make sense for a company to mismanage their expensive investment unless electricity prices drop so low they're tethering on the brink of bankruptcy.

I can see zombie damage as a thing, and I can see looting to be a thing (in which case many or all panels would just be missing, as I don't really see someone just taking one or two random panels). I can also see the possibility for vandal activities, in which case most or all panels would be smashed.

Procyonae:
I actually thought I'd use this as a great quick first use of #66890 to make a damaged and looted variant that become increasingly more likely to spawn over time once it merges (If you want to make said variants ashGlaw they'd just need to be another couple of maps with the same id and lower weight but I'm happy to, dw about the linked PR at all tho I haven't documented it yet)

ashGlaw:
You feel free, Procyonae. Seeing what you did will be a great way to learn how it's done

fairyarmadillo:
> Are there really mismanaged solar farms in the area? Firstly, they should all be comparatively new, and secondly, it doesn't make sense for a company to mismanage their expensive investment unless electricity prices drop so low they're tethering on the brink of bankruptcy.
> 
> I can see zombie damage as a thing, and I can see looting to be a thing (in which case many or all panels would just be missing, as I don't really see someone just taking one or two random panels). I can also see the possibility for vandal activities, in which case most or all panels would be smashed.

No, but there's no Cataclysm in real life. On day 1 of gameplay the world has been completely fucked for weeks and there have been massive portal storms causing everything from localized acid hurricaines to kaiju. For about two months, society has been in such a state that people mostly would not have been going into their jobs or doing them properly.

Add to that the fact that most of the solar farms the player finds will be encountered well into the cataclysm. Solar panels are delicate and need maintenence, and there are angry zombies at some of these sites.

Also people definitely would steal one or two panels. The way the player tends to play the game with a ridiculously sized vehicle and mass hoarding doesn't really fit with other characters or real life. People would definitely just grab one or two to replace a broken one on their car or hideout. It's not like the panels are going anywhere, they can always come back if they need more.

PR #67281, opened 2023-07-30
dodge skill affects dodge stamina cost

### LINKED ISSUE #66050: Dodging skill should lower the stamina cost of dodging

Currently dodging always takes 6 * BASE_BURN_RATE stamina as of #64224. However more skilled dodgers should be able to avoid overcommitting and making more precise motions which avoid attacks while expending less energy.

### Solution you would like.

Dodging skill should reduce stamina costs. for example the formula could be (8 * base_burn_rate) / (1 + dodge skill) 
this would be the BASE dodge skill, not the modifiers it gained from dexterity, mutations, etc.

_No response_

### Additional context

_No response_

anoobindisguise:
#### Summary
Balance "higher dodge skill lowers the stamina cost to dodge"

#### Purpose of change
Fixes #66050
Raw dodge skill should affect amount of stamina needed to dodge, because skilled dodgers know how to dodge with shorter and less exertive movements. Currently in game the stamina cost of dodging was very high, especially when coupled with martial arts that give you more than one dodge, leading to dodge honestly being actively detrimental for how quickly it exhausts you in combat compared to just letting enemies hit your armor over and over. Thus you probably wanted to actually turn off XP for the dodge skill so that you wouldn't start wasting more stamina on dodging... 

#### Describe the solution
Dodge skill now reduces stamina cost linearly to half the base (6 * base_burn_rate -> 3 * base_burn_rate). This still matters as you're being nickel-and-dimed on stamina on top of your melee attacks, grab breaks, etc so it still adds up but dodging should no longer account for quite as much stamina input in combat as before. This is only for the dodge skill, not the raw dodge score. Growing a tail or becoming more dextrous won't teach you how to dodge, it just amplifies your existing knowledge. Only the actual skill itself matters for the calcs.

Different values.

#### Testing
Compiled and confirmed functional. Set my dex to 30000 and sat in front of a zombie. I became too exhausted to dodge after 2 minutes or so of continuous dodging at 0 skill (not making any attacks, just waiting in place) but the drain gradually went down as dodge skill increased.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -1839,7 +1839,8 @@ void Character::on_try_dodge()
     consume_dodge_attempts();
 
     const int base_burn_rate = get_option<int>( STATIC( "PLAYER_BASE_STAMINA_BURN_RATE" ) );
-    mod_stamina( -base_burn_rate * 6 );
+    const float dodge_skill_modifier = ( 20.0f - get_skill_level( skill_dodge ) ) / 20.0f;
+    mod_stamina( std::floor( -static_cast<float>( base_burn_rate ) * 6.0f * dodge_skill_modifier ) );
     set_activity_level( EXTRA_EXERCISE );
 }
 


BjoHart:
Thank you anoobindisguise for all the good stuff your doing!

PR #67282, opened 2023-07-30
Adjust the fried seeds recipe and the plants which grow seeds for it

anoobindisguise:
#### Summary
Balance "fried seeds uses oil and the plants that make fryable seeds make more"

#### Purpose of change
Fried seeds were fun but the plants that made it were really undertuned. A pumpkin made 6 whole seeds items, which amounted to a measly 100 calories. IRL a pumpkin easily has 500 seeds, which is ~660 calories. Sunflowers and hemp plants had similar issues.

#### Describe the solution
I made the yields much greater as a (90 seeds items from disassembly of a pumpkin) 
I also adjusted sunflower and hemp seeds along similar lines each of which can have hundreds or thousands of seeds. Hemp is more variable here, so sunflowers are the primary seed producer if you really want lots of edible seeds.
Calorie counts were adjusted from 17 calories each to 8 for pumpkin seeds, 11 for hemp seeds and 12 for sunflower seeds according to figures I found online (each seeds item is 2 grams)
Frying the seeds uses 50 grams of seeds and 15 grams of oil to produce 64 grams of fried seeds (2 charges). This uses oil because they are "fried" seeds and not "toasted" seeds.

Toasted seeds.


fairyarmadillo:
You should not make the proposed changes to seed counts. A seeds item represents 250ml of seeds, not a single seed. It is abstracted for reasons discussed in #65924 , specifically in the comments (I believe Karol discussed it), and also serves important game and realism balance functions.

Maleclypse:
I think this is ready for rebase/review?

PR #67291, opened 2023-07-30
Mutant Salmon rework

Karol1223:
#### Summary
Content "Mutant Salmon rework"

#### Purpose of change
I've been aiming to touch mutant fish for a while and before I get anywhere new I gotta clean up our old mess

#### Describe the solution
- Rebalanced our current Giant Salmon to have less HP but actual armor, do slightly more damage, not smell the player anymore, be a little faster (roughly ~16.7% faster), and - most importantly - not be a blind killing machine. Salmon of this size would not turn Jaws. Let his evolved brother take that niche
- Added a new Huge Salmon - Giant's big brother that's roughly as heavy as an alligator, more health, more damage, same speed because I'm not convinced this guy would be much faster with the weight increase, angrier now, too
- Added the Leaper Salmon - a slightly further mutated Huge which is a little bit larger, faster, slightly angrier AND IT LEAPS. ~~I need to double-check if it leaps on land but if it does this is my masterpiece~~ It DOES leap on land!
- While I was here I handled the plural forms of gar and sturgeon. Yeah, they can have plural forms if it's used between different species but in the case of how it worked ingame it always referred to only one species at a time so meh
- Added new abstract mutant fish roe and fry for use in future mutant fish
- Added corresponding mutant salmon roe and fry
- Made it so non-mutant trout, salmon and whitefish reproduce specifically in Autumn, as that's when real life Salmon lay roe

#### Additional context
The leaper salmon's method of propulsion is based off an actual biologically present one of the Flying Squids, specifically https://en.wikipedia.org/wiki/Japanese_flying_squid. I'm leaving it here to show that yes, it's a biological mutation but no, the player character should not know the exact mechanisms of it (i.e. the ins-and-outs should not be in the creature description)


data/json/monsters/fish.json
```diff
@@ -1162,31 +1162,62 @@
...
+    "species": [ "FISH" ],
+    "volume": "312 L",
+    "weight": "265 kg",
+    "hp": 225,
+    "armor": { "bash": 5, "cut": 5, "bullet": 3 },
+    "speed": 175,
+    "material": [ "flesh" ],
+    "symbol": "S",
+    "color": "white_magenta",
+    "aggression": 0,
```
Procyonae, line 1208:
```suggestion
    "color": "white_magenta",
```

Karol1223: Yeah I know I've seen the error. I'm ignoring it for now until I figure out something for the next mutant size, then I'll handle it. Thanks for catching it, though!

Karol1223:
I think this is fineeee?

I'm not making them spawn because the base forms I modified already do spawn in the world, so they should just upgrade as the game time progresses.

@Venera3 I'd appreciate if you could take a look at this to make sure I'm not stupid.

PR #67293, opened 2023-07-30
Use mutation visibility as condition

### LINKED ISSUE #67152: Allow mutation visibility values to be referenced in dialogue conditions
### Describe the bug

In the mutation documentation, a visibility entry is referenced for use in NPC dialogue, which dictates how visible your mutations are. Despite my research, this value cannot be referenced in dialogue conditions for use as it was intended.

### Attach save file

N/A

### Steps to reproduce

1. Make new dialogue
2. Try to reference it in a condition
3. Fail miserably

### Expected behavior

I expect to be able to call on this when I need to.

### Screenshots

_No response_

OS: Windows 11
-Game Version: cdda-windows-tiles-x64-2023-07-09-0709
-Graphics Version: Tiles
-Ingame Language: English (US)
-Mods: Default Mods

### Additional context

If nobody wants to code this, I suppose I could make an EOC that adjusts a personal variable that governs total mutation visibility, and use that value instead.

SurFlurer:
#### Summary
none

#### Purpose of change
Address #67152

#### Describe the solution
As there are visible mutations displayed in the NPC's info window, you can use them as condition for dialogue.
image

Also make the arsonist dialogue use this new condition to determine the responses.

Hopefully this has the prospects to improve immersion and enhance role playing.

Allow using mutation visibility values in eoc context for more customized judgement.

#### Testing
1. Go to the arsonist, find that I can see they have feline ears, and the corresponding response shows up.
2. Set my character's per to 0, now their feline ears trait doesn't show up in the NPC info window, and the corresponding response doesn't show up.

#### Additional context
1. Should I add a `sees()` check to this?
2. I believe more conversation can be ported to use this instead of viewing from God's perspective and use `u_has_trait`.
3. Should I imitate `u_has_any_trait` and implement something like `u_has_any_visible_trait`?


MNG-cataclysm:
In addition to this, would it be possible to use mutation viability as a `u_val` when checking for just any visible mutations?

MNG-cataclysm:
Thank you for doing this, it looks great!

MNG-cataclysm:
I've been playing around with this new code for some of my things, and I have a (I think) small request. Could we use the visibility output that the C++ checks for as part of a math function, so we could check for generally any visible mutations?

`{ arithmetic: [ { u_val: visibility }, "=" { const: 15 } ] }` for example.

PR #67308, opened 2023-07-30
Better description for diary keybinding

jbytheway:
#### Summary
None

#### Purpose of change
At some point the achievements / scores / kills window was moved behind the diary window.  But the keybinding is just called "open diary".  This makes it essentially impossible to figure out how to get to the achievements and scores unless you already know.

#### Describe the solution

Rename the keybinding to mention achievements and scores too, so that it can be found more easily.

Also mentioning 'kills'.

#### Testing
Searched in the keybindings screen in-game (see screenshot below).

#### Additional context
I've been confused by this, and I've seen at least one YouTuber be confused by it too.

achievements-in-keybindings

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -3025,7 +3025,7 @@
   {
     "type": "keybinding",
     "id": "diary",
-    "name": "Open diary",
+    "name": "Open diary / achievements / scores",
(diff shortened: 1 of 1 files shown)


ashGlaw:
Good idea. I know where the menu is, and I've forgotten on occasion. This makes it super clear

PR #67309, opened 2023-07-30
More consistent key highlighting

jbytheway:
#### Summary
None

#### Purpose of change
Input contexts have a get_desc option intended to get a description of a particular command with the key for that command highlighted in it.

There is an option to pass in formatting information to control exactly how these strings are formatted.

The default formatting is somewhat boring (no colour).  Two places I found override this default formatting: the crafting menu highlights the relevant letter in yellow, and the vehicle interaction menu highlights it in green.  The vehicle menu doesn't use any brackets around the relevant letter like the other uses do.

This seems unnecessarily inconsistent.

#### Describe the solution
I changed the default to use colour, but green instead of yellow.  I think green is the more common colour used for keys the player can press across the CDDA UI (and I prefer it personally).

I then stopped the crafting menu from overriding the default.

I did not change the vehicle menu, because what it's doing is a bit more complex, and changing to the default might make the options not fit on smaller screens.

I could have changed the vehicle menu too, or more.  There are still other places in the UI using yellow (like the main menu) which I have not changed.  I believe they don't use `get_desc`.  Perhaps they should, but I'm treating that as out of scope for this change.

#### Testing
Looked at stuff in-game.  See screenshots below.

#### Additional context
Here are four examples from before the change:
get_desc-vehicle-before
get_desc-crafting-before
get_desc-diary-before
get_desc-dialog-before

The vehicle UI is unchanged; here are the other three after the change:
get_desc-crafting-after
get_desc-diary-after
get_desc-dialog-after

You can see that there's one small issue where the blue highlighting on the selected option in the dialog box is dropped for the one green letter.  ~That might be an issue with the code that interprets colour tags; I could look into that but it seemed mostly orthogonal to this PR.~  That's now fixed (see below).

### DIFF
--- a/src/crafting_gui.cpp
+++ b/src/crafting_gui.cpp
@@ -1148,20 +1148,10 @@ const recipe *select_crafting_recipe( int &batch_size_out, const recipe_id &goto
         const unsigned int header_info_width = std::max( width / 4, width - FULL_SCREEN_WIDTH - 1 );
         const int wStart = ( TERMX - width ) / 2;
 
-        // Keybinding tips
-        static const translation inline_fmt = to_translation(
-                //~ %1$s: action description text before key,
-                //~ %2$s: key description,
-                //~ %3$s: action description text after key.
-                "keybinding", "%1$s[<color_yellow>%2$s</color>]%3$s" );
-        static const translation separate_fmt = to_translation(
-                //~ %1$s: key description,
-                //~ %2$s: action description.
-                "keybinding", "[<color_yellow>%1$s</color>]%2$s" );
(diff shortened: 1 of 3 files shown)


jbytheway:
So it turns out that fixing the highlighting on the dialog button selections was pretty easy, so I've added that too.
get_desc-dialog-highlighting-fix

alef:
`git` blame lead me here. This is what I see in `Attach as gunmod` action from `Use item` menu after `Select gun to modify`:
image

jbytheway:
Thanks for letting me know.  Hopefully fixed in #67778

PR #67327, opened 2023-07-31
[Sky Island] "choose your difficulty" uses selector now

GuardianDll:
#### Summary
None
#### Purpose of change
It annoyed me
#### Describe the solution
replace old queue of picking the difficulty to use a new, cool selector EoC
image


GuardianDll:
done

PR #67342, opened 2023-07-31
Fix use menu denial message and invoking items for comestibles without charges

### LINKED ISSUE #67330: Cattle fodder(and pills ) need charge in use menu
### Describe the bug

report
as the screenshot 
pills can be use in eat menu so it is not that a problem

### Attach save file

N/A

### Steps to reproduce

Just get a Cattle fodder and press a

### Expected behavior

It is not a charge item

### Screenshots

_No response_

version 2023-07-31 1024

### Additional context

_No response_

mqrause:
#### Summary
None

#### Purpose of change

Fixes #67330

#### Describe the solution

Exit early when checking `item::ammo_sufficient` for comestibles, so the menu and invoking items accepts comestibles as having the required charge.

This whole chain of `item::ammo_sufficient` > `item::shots_remaining` > `item::ammo_required` > `item::ammo_remaining` is really ugly, because it shouldn't even have anything to do with comestibles. But it's still checking them, expects comestibles to use 1 charge, but now obviously fails for uncharged comestibles. This really should get a rework to properly split things up, but that requires quite a bit more work.

#### Testing

Activated a caffein pill from the use menu.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -10559,6 +10559,10 @@ bool item::ammo_sufficient( const Character *carrier, int qty ) const
         return ammo_remaining( carrier, true ) >= qty;
     }
 
+    if( is_comestible() ) {
+        return true;
+    }
+
     return shots_remaining( carrier ) >= qty;
 }
 


mqrause:
Looks like this won't work, unfortunately.

Hirmuolio:
I feel like adding extra checks for this in `ammo_sufficient` is not good. I think it would be better to have that check somewhere around https://github.com/CleverRaven/Cataclysm-DDA/blob/b4c54696208229b548294f9de17226f2476b77e9/src/game_inventory.cpp#L1105

mqrause:
I don't like the whole setup either, but it's what `ammo_sufficient` actually checks and it does so at multiple places. Before the charge removal, it'd go deeper into the chain and check if `charges` is larger than `qty`. Which is always true for comestibles if I'm not mistaken because `qty` will always be 1 and `charges` will always be >= 1 if the item actually exists.

It'd be great to clean this all up, but I'm currently not willing to dive into that. Most important is making the use menu functional with comestibles and actually allow comestibles with use actions to be actually used.

PR #67349, opened 2023-08-01
Add functions for efficiently handling massive amounts of identical items

Kamayana:
#### Summary
Performance "Add functions for efficiently creating massive amounts of identical items and finding pockets for them, use to greatly speed up debug spawning"

#### Purpose of change
The game has a lot of trouble handling massive numbers of individual items, and the removal of consumable charges in #60885 has made this issue even more difficult to ignore. I did some profiler tests of areas that were particularly laggy, and one bottleneck comes from `can_stash()` and associated functions. When large amounts of an item are created via crafting or the wish menu, the game laboriously checks if the player can hold every single item individually, even though every item being created is exactly the same.

#### Describe the solution
I made a duplicate of `Character::can_stash` named `Character::can_stash_copies`. This function has a `int copies` parameter that lets you ask "can this character stash `copies` number of `it` items in their pockets?", and it returns the number of remaining copies that won't be able to fit. So if it returns `0`, that means the player has enough room in their inventory to hold everything. If it returns the same value as `copies`, then the player has no room for any of them. Now instead of calling `can_stash` 1941 separate times, we call `can_stash_copies` one single time, which almost entirely removes the CPU% demand.

- [x] Make a version of `can_stash` that's optimized for mass item duplicates.
- [x] Make a version of `i_add` that's optimized for mass item duplicates.
- [x] Make a version of `map::add_item_or_charges` that's optimized for mass item duplicates, and use for `i_add`.
- [x] ~~Check to see how crafting could be optimized with these functions.~~
- [x] See if `can_stash` and `can_stash_copies` could be consolidated to have less repeat code.

I crossed out the "Check to see how crafting could be optimized with these functions" task because I think it'd be better to keep this PR narrowly focused on improving debug spawn speed, so these functions can be available sooner for other optimization PRs.

For map item spawning, I first attempted `current_submap->get_items( l ).insert( copies, new_item )`, but for some reason the game doesn't like using that version of the function for item stack colonies. The errors implied it wasn't an easy fix, so I went with a simple loop instead, which worked well enough.

### Before
This profiler result is for debug wishing in 1941 caffeine pills (the same number as a craft) while wearing a single debug pocket universe. `i_add` is the worst bottleneck, but `can_stash` is right behind it:

debug_spawn_1941_pills_0

### After
The exact same test after the changes now has so little impact it doesn't show up in the profiler - the only CPU usage here is from drawing the window.

image

Here is a much more intense test: spawning 100,000 pills, while carrying a huge number of pockets. It's still 16 times faster than the easy test on the vanilla version. Note that doing this will still bring the game to a crawl afterwards as everything else has trouble processing that many items, but at least creating them isn't a problem.

image

I've also done tests for spawning items while not equipped with any pockets and it is similarly significantly sped up to debug spawn items onto the ground.


ampersand55:
Unrelated to the performance issues, but while you are working on this would it be possible to implement #67385?

mqrause:
That's very much out of scope for this PR.

ampersand55:
I wasn't thinking of this PR. I just thought I'd mention it as possibly related to the infrastructure for handling massive amounts of identical items.

Kamayana:
> I wasn't thinking of this PR. I just thought I'd mention it as possibly related to the infrastructure for handling massive amounts of identical items.

It's something I've been thinking about. This PR is focusing on debug item spawning, which is kinda easy mode; since all items are freshly created it's fine to treat them as identical in all aspects. It becomes trickier when managing items that already exist, as they might have individual differences, like food items with different rot values. At the same time, though, different rot values shouldn't matter for something like checking how many will fit in a pocket, so I think this might still be useful for optimizing that area.

Kamayana:
Marking this as ready for review. The test I was using now has so little impact it doesn't show up in the profiler - the only CPU usage here is from drawing the window.

image

Here is a much more intense test: spawning 100,000 pills, while carrying a huge number of pockets. It's still 16 times faster than the easy test on the vanilla version. Note that doing this will still bring the game to a crawl afterwards as everything else has trouble processing that many items, but at least creating them isn't a problem.

image

I've also done tests for spawning items while not equipped with any pockets and it is similarly significantly sped up to debug spawn items onto the ground.

Issue #64799, opened 2023-04-02
NPCs assigned to faction camp construction tasks appear to cause the game to run incredibly slowly.


Froffy025:
### Describe the bug

Title. My game is running slowly ever since I assigned my NPC to perform a construction task. Message report shows that NPC is "trying to find necessary items to do the job" 300-400 times every time I move.
Telling the NPC to stand guard and follow causes the issue to stop.

### Attach save file

[Monarch_copy-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/11133553/Monarch_copy-trimmed.tar.gz)

### Steps to reproduce

1. Create player & NPC
2. Create faction camp
3. Use construction zones to try to create project (in my case, a log cabin with furniture, roof, etc.), basecamp storage to store some of these materials
4. Assign NPC to faction camp to do construction work

### Expected behavior

I expected the NPC to either find the materials in storage, or not be able to find them and give up for a while. However, they instead lagged my game by checking more than once per turn.

### Screenshots

![image](https://user-images.githubusercontent.com/78222136/229384142-b833ce50-c4f9-43f5-af76-c88392db5a70.png)

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.19044.2728 (21H2)
- Game Version: a33c958 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    No Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions],
    Stats Through Skills [StatsThroughSkills],
    Stats Through Kills [stats_through_kills],
    SpeedyDex [speedydex],
    Blaze Industries [blazeindustries],
    Bombastic Perks [bombastic_perks],
    Translate Complex Dialogue [translate_dialogue]
]

PatrikLundell:
[...]

PR #67350, opened 2023-08-01
Reduce NPC faction camp task slowdown (Part 1)

lispcoc:
#### Summary
Performance "Fix NPC faction camp task slowdown"

#### Purpose of change
This is a step towards resolving #64799.

In #64799, @SurFlurer raised three issues as causes.
> 1. The infinite loop.
> 2. Low efficiency of that function.
> 3. The NPC silently fails to retrive resources for no obvious reason.

1 is caused by NPCs fails to fetch for task infinitely. They dont remember that fetch is failed and tries to continue task, So they infinitely repeats fetching that are bound to fail.
2. `get_point_set_loot` function returns tripoints that are in zones and in range. but this function checking too many out-of-zone points unnecessarily.
I try to make an attempt to fix these two problems.

#### Describe the solution
1. Make NPC remember what things already tried to fetch. They give up a task  for a while if they failed already.
2. Dont search out of zones. and stop to use `get_zone_at`, that takes a while and need to call many times.

Fetch consumes very short In-game time (about 1/100 turn), so NPCs consume VERY long real time because they try to fetch hundreds of time per turns. Makeing fetch take some time reduces slowdown. This is simple solution, but NPC may not move on to the next task for a long time.

#### Testing
Slowdown is reduced in savedata of #64799.
Basic npc multiple construction task test passed.

#### Additional context
Multiple activity is very hard to understand. I may have overlooked something, so please point it out if you notice something.


src/handle_action.cpp
```diff
@@ -1401,6 +1401,9 @@ static void loot()
     menu.query();
     flags = ( menu.ret >= 0 ) ? menu.ret : None;
     bool recache = false;
+    // clear fetch history previous multiple activity.
+    player_character.fetch_history.clear();
```
SurFlurer, line 1405: The fetch history should belong to NPC instead of player character.

lispcoc:
> I think you can store the fetch history as a unordered_map<time_point, string>, so instead of defining a 10 minute refresh rate, you can look up if there has already been a fetch attempt with the same goal in the same turn. Also, storing the history in player_activity seems to be more reasonable than storing that in Character. ( when assigning a new fetch activity, check the previous activity's fetch history )

It is good idea, but player_activity is supposed to be removed many times in multi activity cycle. So fetch history is to be removed together.
I'm looking into this solution, but it might be difficult.

Maleclypse:
> > I think you can store the fetch history as a unordered_map<time_point, string>, so instead of defining a 10 minute refresh rate, you can look up if there has already been a fetch attempt with the same goal in the same turn. Also, storing the history in player_activity seems to be more reasonable than storing that in Character. ( when assigning a new fetch activity, check the previous activity's fetch history )
> 
> It is good idea, but player_activity is supposed to be removed many times in multi activity cycle. So fetch history is to be removed together. I'm looking into this solution, but it might be difficult.

So should this be merged as is and you'll PR the improvement or this is waiting until you've decided on the other option?

lispcoc:
> > > I think you can store the fetch history as a unordered_map<time_point, string>, so instead of defining a 10 minute refresh rate, you can look up if there has already been a fetch attempt with the same goal in the same turn. Also, storing the history in player_activity seems to be more reasonable than storing that in Character. ( when assigning a new fetch activity, check the previous activity's fetch history )
> > 
> > 
> > It is good idea, but player_activity is supposed to be removed many times in multi activity cycle. So fetch history is to be removed together. I'm looking into this solution, but it might be difficult.
> 
> So should this be merged as is and you'll PR the improvement or this is waiting until you've decided on the other option?

I think this is fatal problem and needs workaround. I will make PR improve this. Please merge if possible.

src/handle_action.cpp
lispcoc: fetch history now belong to NPC. In the future these should belong to player_activity but it is needed to clean up codes , because multi_activity is not supposed to resume.

PR #67352, opened 2023-08-01
Update all active items to new tick action system and remove old system

Hirmuolio:
#### Summary
Infrastructure "Update all active items to new tick action system and remove old system"

#### Purpose of change

* Get rid of `use_action`  being called in active tool processing.
* Get rid of `bool t`.

#### Describe the solution

Add some check for `!p`. That should never happen on those actions but bad json can cause it. More helpful than crash or nothing happening.

Some iuse handled both "player activated item" and "active item processing". Split those actions into two separate actions:
* multicooker
* remoteveh
* radiocaron
* radiocontrol
* foodperson
* ehandcuffs
* einktabletpc
* gasmask
* dive_tank
* mp3_on
* geiger
* noise_emitter_on
* fish_trap

#### Testing

The above uses were tested. Basic "run it in inventory", "drop it while active" and "wait until it runs out of charges".

Playing music that was copied from memory cards is weird functionality and does not work well on some more complex computers included in Aftershock (Clash with flashlight form, they have MP3 functionality that works and can be used instead).


src/iuse.cpp
```diff
@@ -5676,21 +5681,22 @@ void item::extended_photo_def::serialize( JsonOut &jsout ) const
...
-std::optional<int> iuse::einktabletpc( Character *p, item *it, bool t, const tripoint &pos )
+std::optional<int> iuse::epic_music( Character *p, item *it, const tripoint &pos )
 {
-    if( t ) {
-        if( !it->get_var( "EIPC_MUSIC_ON" ).empty() &&
-            it->ammo_sufficient( p ) ) {
-            if( calendar::once_every( 5_minutes ) ) {
-                it->ammo_consume( 1, p->pos(), p );
-            }
+    if( !it->get_var( "EIPC_MUSIC_ON" ).empty() &&
```
Maleclypse, line 5686:
```suggestion
    if( !it->get_var( "EPIC_MUSIC_ON" ).empty() &&
```
Typo or am I messing things up?

Hirmuolio: That typo already existed before me. AFAIK working as intended.

https://github.com/CleverRaven/Cataclysm-DDA/blob/383e125594ba442751c94176c948d6501398a274/src/iuse.cpp#L5682

PR #67383, opened 2023-08-02
carpet construction is easier

anoobindisguise:
#### Summary
Balance "placing carpet needs 0 fab instead of 1 and takes 1 minute instead of 30"

#### Purpose of change
Laying down a 1 meter square section of carpet doesn't take 30 minutes of hard work.

#### Describe the solution
Make it 1 minute, and get rid of the fab 1 requirement. Laying down a rug or whatever doesn't seem especially challenging.

Lower time even more?


ampersand55:
You missed the fabrication requirement on `carpet_floor_purple`

Kamayana:
Laying a carpet so it's properly flat isn't just a matter of plopping it down - note that you use a hammer and nails for it. I agree that 30 minutes is probably too much but 1 minute is too little I think. It's sorta hard to judge the actual time it should take because carpet is placed all at once in the real world but in-game it's like a jury-rigged tile-by-tile construction, sorta like carpeting a room using a bunch of individual carpet samples.

Maleclypse:
I've kicked carpet (professional term) before for money and 30 minutes is right assuming that we've abstracted away the stretching tool.  It's longer without a stretching tool.  You can get faster if it's your every day job but I don't think the survivor will ever reach that.

Maleclypse:
I guess if we are considering them 1 meter square tiles of carpet we are nailing down something less than 30 mins makes sense I also don't think it's 1 min though

data/json/construction.json
```diff
@@ -3637,8 +3637,8 @@
     "id": "constr_carpet_red",
     "group": "carpet_floor_red",
     "category": "DECORATE",
-    "required_skills": [ [ "fabrication", 1 ] ],
-    "time": "30 m",
+    "required_skills": [ [ "fabrication", 0 ] ],
+    "time": "1 m",
```
I-am-Erk, line 3641: Per discussion I'd suggest this split the diff. Half an hour per segment is excessive, but a 5mx5m room also takes much longer than 25 mins to lay carpet. Especially if you're dropping the requirements to fab0, so you could easily have someone doing this who doesn't know how to read a tape measure.
```suggestion
    "time": "15 m",
```

I-am-Erk (changes requested):

I-am-Erk:
(I had meant for that review comment to apply to all the recipes)

PR #67384, opened 2023-08-02
Vehicle improvements

deepvelocity:
#### Summary
Features "Vehicle improvements: togglable pre-collision system and foldable turrets"

#### Purpose of change

1. Autopilot self-driving features can behave in a weird manner due to pre-collision check. Example: vehicle will refuse to go up the ramp on a bridge (viewing ramp as a collision candidate). It also prevents the player from building blade-covered mowing bots, kamikaze bots and probably other fun stuff.
2. We have foldable robot driving unit and turret mount, but not a single foldable mounted gun. Also, unfoldable autopilot. This PR addresses this problem.

#### Describe the solution
1. Separate the pre-collision check into a separate method, introduce a separate variable to enable or disable pre-collision system. This way, existing code base will not suffer changes and the player will have more room for customization (finally "programmable" part of the autopilot will make sense!)
2. Add option to enable/disable this feature in autopilot menu.
3. Make every Pistol and SMG mounted gun foldable using the gun's item volume.

The only concern I have is around the definition of foldable. While it appears that generally Cataclysm DDA refers to things as foldable when they can actually be folded, it also doesn't seem logical that the player cannot pick up a single tile vehicle using its full volume for the final item volume. Making only SMGs and Pistols foldable seems like a good middle ground.

#### Testing
Local build with tests enabled passed. All features (togglable CAS system, foldable turrets/autopilot, remote control selection skip, sound beeps) work as intended.

Also tested with animal-powered vehicle to confirm that those do not produce the same sound.


Qrox:
Why closed?

deepvelocity:
> Why closed?

This PR contained changes from another PR I raised, which had not been merged yet at the moment. Re-opened PR.

data/json/vehicleparts/vehicle_parts.json
```diff
@@ -468,6 +468,7 @@
     "broken_color": "red",
     "damage_modifier": 80,
     "durability": 100,
+    "folded_volume": "2500 ml",
```
Maleclypse, line 471:
```suggestion
    "folded_volume": "2500 ml"
```

Maleclypse: Resetting test suite

data/json/vehicleparts/vehicle_parts.json
```diff
@@ -468,6 +468,7 @@
     "broken_color": "red",
     "damage_modifier": 80,
     "durability": 100,
+    "folded_volume": "2500 ml"
```
Maleclypse, line 471:
```suggestion
    "folded_volume": "2500 ml",
```

Issue #67326, opened 2023-07-31
Charge removal side effects 3: unmigrated item spawns


mqrause:
### Describe the bug

There are countless places that allow inline definitions of itemgroups or other shorthand definitions that were not caught by the automated script, so all of these are now broken and only spawn single items instead of the defined charges or default charges of the item.

Where to find these cases:
1. Inside regular itemgroups
 - "distribution" or "collection" members inside the "items" or "entries" array
 - simple itype entries inside the "items" array 
2. Professions
 - lots of simple itype entries, although probably not many that spawn with comestibles
3. Mapgen (and vehicles?)
 - this one is the most tricky, because there are many different ways to define spawns and I definitely don't know all of them
4. Recipe byproducts
 - haven't actually confirmed that one, but it's probably an issue?
5. Probably other places I forgot
 - comments appreciated

### Attach save file

n/a

### Steps to reproduce

n/a

### Expected behavior

n/a

### Versions and configuration

6088293934801590f503ff9d5d961e708af564ff

PR #67410, opened 2023-08-03
[CR] Fix "*_inf" item groups and check for unmigrated item spawns

### LINKED ISSUE #67325: Charge removal side effects 2: Broken itemgroups named `*_inf`
### Describe the bug

These automatically created groups are intended to fill up their container with a randomized number of charges between the min and the container capacity. Because this functionality does not exist for non-charge items, currently all these groups always only spawn the minimum amount of items.

### Attach save file

n/a

### Steps to reproduce

n/a

### Expected behavior

There are two possible solutions to this:
1. The json only way is to manually calculate and define a maximum amount for all the groups. While it's perfectly functional, it's quite a bit of work (45 groups like that in total), and any changes to the items or their containers could throw off the group. A list of all the automatically created groups is here: https://gist.github.com/mqrause/dc234737bfd5c4309aa4dc496d17bd53#file-itemgroup_migration-txt
2. Add code support that emulates the previous charge behavior for filling containers with non-charge items. Definitely the preferred solution, but requires C++.

### Screenshots

_No response_

6088293934801590f503ff9d5d961e708af564ff

### Additional context

_No response_

SurFlurer:
#### Summary
none

#### Purpose of change
Fix #67325
Partially address #67326

#### Describe the solution

Add a finalizing step for item groups that sanitize the usages of charge and count in item groups. Calculates the maximum number of one type of items that can be stored in the "closest" container ( in the nested container chain ). The reason I decide to calculate it at finalizing instead of run time is less overhead and less alteration to other parts of code.

I saw that those single_item_creator objects created by `add_entry()` have an unused `container_item` variables, so I decided to a field in itemgroup entry called "entry-wrapper" to represent that all spawn results of this entry should be wrapped in a provided wrapper/container.

~~This is heavily WIP. Hopefully CI logs can be used to identify those that needs migration ( although only item names of item spawn entries are provided in logs, since I haven't been able to get the readable itemgroup id whose item spawn entries need migration, so manual inspection are needed after searching for such an item spawn entry in VSCode ).~~

Most identified broken itemspawns in core game has been fixed ( hopefully ), mods included. But there are leftovers as mqrause said : https://github.com/CleverRaven/Cataclysm-DDA/pull/67410#issuecomment-1664274443 . 

I've tried my best to make my modified entries work "as intended", which means to set entry-wrapper to the item's default container and container-item to null. For follow-up contributors to deal with these leftovers, it's may be needed to explicitly set `container-item` field to be `null`, otherwise the item_modifier will try to wrap the item in its default container.

Separate this to be two steps of finalizing and verifying. But I think these checks are closely connected and some verification needs to be performed before deducing count-max.

Moving those newly created item groups that were used for migration back to their inline positions. This can be done later. 

Should I make those debugmsg test_mode only?

#### Testing
Todo


SurFlurer:
Thanks, pr description updated.

src/item_factory.cpp
```diff
@@ -1319,8 +1319,11 @@ void Item_factory::finalize()
             it->second.recipes.push_back( p.first );
         }
     }
+    for( auto &e : m_template_groups ) {
+        auto &isd = e.second;
+        isd->finalize( itype_id::NULL_ID() );
```
mqrause, line 1324: Did a small test with passing `e.first.str()` to `finalize` and it gave me output like this:

```
03:21:13.768 ERROR : [finalize] entry in itemgroup toxic_waste_mx_radioactive for spawning item cobalt_60 defined charges but can't have any
03:21:13.768 ERROR : [finalize] entry in itemgroup toxic_waste_mx_radioactive for spawning item plutonium defined charges but can't have any
03:21:13.783 ERROR : [finalize] entry in itemgroup zombie_reenactor_drops for spawning item tobacco defined charges but can't have any
03:21:13.783 ERROR : [finalize] entry in itemgroup zombie_reenactor_drops for spawning item hardtack defined charges but can't have any
03:21:13.783 ERROR : [finalize] entry in itemgroup ÃÀ 116 for spawning item cigar defined charges but can't have any
03:21:13.798 ERROR : [finalize] entry in itemgroup ÃÀ 1200 for spawning item 2x4 defined charges but can't have any
03:21:13.798 ERROR : [finalize] entry in itemgroup ÃÀ 1230 for spawning item pipe defined charges but can't have any
03:21:13.798 ERROR : [finalize] entry in itemgroup ÃÀ 1260 for spawning item splinter defined charges but can't have any
```

Doesn't help with the completely anonymous groups, but it's at least something.

SurFlurer: Thanks, I'd like to implement this but I'm pretty busy irl, and I believe these debugmsg will hardly appear in the future ( sorry to 3rd-party mod maintainers ), so...perhaps this can be kept for future.

PR #50262, opened 2021-07-27
Hobbies in Character Creation

Saelfen:
#### Summary
Features "Hobbies in Character Creation"

#### Purpose of change
Implement #45031
Players need flexibility in starting proficiency choices along with more specificity to character identity.

#### Describe the solution
Add Hobbies in character creation that act as multi-selectable professions which offer a package of proficiencies, skills, traits, and addictions.  Any number of hobbies can be chosen, giving more flexibility for character identity. This PR will be more sparse on the hobbies side, but there will be more to come.

image

On the skills side, hobbies provide additional experience instead of flat levels as a balancer against multiple hobby selections. The percentage and final level is reported in the description tab
image

#### Describe alternatives you've considered
Create a new hobbies object (and this is what I did to start), but it's essentially a direct copy of the existing professions object. So use that instead, and throw in a boolean to declare whether or not it's a hobby. There is no need to re-invent the wheel.

#### Testing
- Make sure tab order works with new Hobbies selection.
- Check Multi-Select works properly with increment/decrement points, skills, traits, final character description, and actually creating the character and starting the game. This includes loading proficiencies, skills, traits, and addictions.
- Make sure templates save and load properly.

#### Additional context
More hobbies to come in a second PR

PR #67412, opened 2023-08-03
More granularity and higher levels in backgrounds

Aerin-of-the-Toast:
#### Summary
Content "Add a bunch of new backgrounds that are tiers of existing ones."

#### Purpose of change

The primary reason is more granularity in character creation for the roleplayers out there. If there's only one option for each background, then both "dad who played football with his kids" and "Peyton Manning" are sharing the same "Football" background. 

Fittingly, many of the new backgrounds are higher tiers of old ones (with a few lower tiers as well), since another thing I wanted to address was a lack of options for people wanting to play "experts" instead of the usual kinda-bad-at-everything survivors most players default to.

The last reason is tutorializing; as a new player it can be hard to parse what exactly "4 throwing" or etc. can entail, so hopefully by adding more concrete examples that are more specific than vaguely "you played basketball, here's 2 throwing" new players can see "Oh, a beginner archer would have 2 archery, an intermediate archer would have 4, and a master archer would have 8."

#### Describe the solution

A shitload of new background options, denoted generally with (Beginner) for <3, (Intermediate) for 3-5, (Expert) for 6-7, and (Master) for 8-10. I believe this adds up to about 25 new backgrounds in total. In particular, the backgrounds I've added grades to were Lockpicking, Swimming, Cooking, DIY Crafting, Sewing, Baseball, Basketball, Football, Wrestling, Street Fighting, Fencing, Shooting (previously Target Shooting) and Melee Weapon Training. I've also added upgraded versions of other traits that use proprietary names instead of just (Expert) or the like. These ones are Racing (upgrade of Motorsports) and Black Belt (an upgraded version of Self Defense and/or Martial Arts, based on the profession of the same name).

So much number fiddling...

#### Testing

Checked and made sure the game loaded without errors and the new backgrounds were functioning.

#### Additional context

This isn't an exhaustive list by any stretch, but I'm hoping others will follow suit for any backgrounds they might have an interest in.

It seems skills are labelled as "advanced" in the backgrounds menu for any skill 4 or higher, but that's not really something I can fix.

Lastly, I'd make the traits of the same type (i.e. Archery (Beginner) and Archery (Intermediate)) conflict and be unable to be taken together, but there doesn't seem to be that functionality that I could tell.

### DIFF
--- a/data/json/hobbies.json
+++ b/data/json/hobbies.json
@@ -201,7 +201,7 @@
     "subtype": "hobby",
     "id": "archery_master",
     "name": "Archery (Master)",
-    "description": "You're a modern day Robin Hood and probably have hella back muscles.  Perhaps you're an olympic archer in training or maybe you're descended from the longbowmen of old.  In either case, sniping zombie heads should be as easy as pinning apples to trees for you.",
+    "description": "You're a modern day Robin Hood and probably have hella back muscles.  Perhaps you're an Olympic archer in training or maybe you're descended from the longbowmen of old.  In either case, sniping zombie heads should be as easy as pinning apples to trees for you.",
     "points": 8,
     "skills": [ { "level": 8, "name": "archery" }, { "level": 5, "name": "gun" } ],
     "proficiencies": [ "prof_bow_basic", "prof_bow_expert", "prof_bow_master" ]
--- a/data/json/hobbies.json
+++ b/data/json/hobbies.json
@@ -100,11 +100,21 @@
     "type": "profession",
     "subtype": "hobby",
     "id": "lockpicking",
-    "name": "Lockpicking",
+    "name": "Lockpicking (Beginner)",
     "description": "Whether it was for fun, innocent utility, or criminal intent, you learned the basics of how to silently pick locks.  Now, being able to get into secured places just got a lot more useful.",
     "points": 1,
     "proficiencies": [ "prof_lockpicking" ],
     "skills": [ { "level": 1, "name": "traps" } ]
+  },
+  {
+    "type": "profession",
+    "subtype": "hobby",
+    "id": "lockpicking_expert",
+    "name": "Lockpicking (Expert)",
+    "description": "Maybe you're a certified locksmith, a career burglar, or an intense hobbyist.  Regardless, you've spent so long picking locks you could pop most open in under a minute easy, with the right tools.",
+    "points": 3,
+    "proficiencies": [ "prof_lockpicking", "prof_lockpicking_expert" ],
+    "skills": [ { "level": 4, "name": "traps" } ]
   },
   {
     "type": "profession",
@@ -170,17 +180,46 @@
     "type": "profession",
     "subtype": "hobby",
     "id": "archery",
-    "name": "Archery",
-    "description": "You enjoy nocking back an arrow and hearing it whistle through the air.  If you can get the proper gear together and stay on-target, you'll have a good, quiet alternative to using guns.",
-    "points": 3,
-    "skills": [ { "level": 3, "name": "archery" }, { "level": 2, "name": "gun" } ],
+    "name": "Archery (Beginner)",
+    "description": "You've used a bow before and maybe taken a class or two.  You won't be pulling off any trick shots but at least you can usually get an arrow to fly forward instead of spinning off to the side.",
+    "points": 2,
+    "skills": [ { "level": 2, "name": "archery" }, { "level": 2, "name": "gun" } ],
     "proficiencies": [ "prof_bow_basic" ]
   },
   {
     "type": "profession",
     "subtype": "hobby",
+    "id": "archery_expert",
+    "name": "Archery (Intermediate)",
+    "description": "You enjoy nocking back an arrow and hearing it whistle through the air.  If you can get the proper gear together and stay on-target, you'll have a good, quiet alternative to using guns.",
+    "points": 4,
+    "skills": [ { "level": 4, "name": "archery" }, { "level": 3, "name": "gun" } ],
+    "proficiencies": [ "prof_bow_basic", "prof_bow_expert" ]
+  },
+  {
+    "type": "profession",
+    "subtype": "hobby",
+    "id": "archery_master",
+    "name": "Archery (Master)",
+    "description": "You're a modern day Robin Hood and probably have hella back muscles.  Perhaps you're an olympic archer in training or maybe you're descended from the longbowmen of old.  In either case, sniping zombie heads should be as easy as pinning apples to trees for you.",
+    "points": 8,
+    "skills": [ { "level": 8, "name": "archery" }, { "level": 5, "name": "gun" } ],
+    "proficiencies": [ "prof_bow_basic", "prof_bow_expert", "prof_bow_master" ]
+  },
+  {
+    "type": "profession",
+    "subtype": "hobby",
+    "id": "swimming_beginner",
+    "name": "Swimming (Beginner)",
+    "description": "You can swim.  Probably not very well, and you won't break any speed records, but, you CAN swim.",
+    "points": 1,
+    "skills": [ { "level": 2, "name": "swimming" } ]
+  },
+  {
+    "type": "profession",
+    "subtype": "hobby",
     "id": "swimming",
-    "name": "Swimming",
+    "name": "Swimming (Intermediate)",
     "description": "You're a strong swimmer, and you're pretty sure zombies aren't.  You can't swim forever, but it's certainly an advantage over the undead.",
     "points": 2,
     "skills": [ { "level": 4, "name": "swimming" } ],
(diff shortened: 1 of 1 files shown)


data/json/hobbies.json
```diff
@@ -170,42 +180,127 @@
...
+    "skills": [ { "level": 4, "name": "bashing" }, { "level": 7, "name": "throw" }, { "level": 3, "name": "swimming" } ]
+  },
+  {
+    "type": "profession",
+    "subtype": "hobby",
+    "id": "baseball_expert",
+    "name": "Baseball (Expert)",
+    "description": "You didn't quite make it into the big leagues, but you were close.  Now it's time to go for the grand slam or you'll be dealing with worse than a strike out.",
+    "points": 6,
+    "skills": [ { "level": 6, "name": "bashing" }, { "level": 5, "name": "throw" }, { "level": 3, "name": "swimming" } ],
```
snipercup, line 272:
```suggestion
    "skills": [ { "level": 6, "name": "bashing" }, { "level": 5, "name": "throw" } ],
```
I guess you forgot to remove the swimming skill? I don't see why a baseball_expert would have swimming 3

data/json/hobbies.json
```diff
@@ -170,42 +180,127 @@
...
+    "skills": [ { "level": 4, "name": "bashing" }, { "level": 4, "name": "throw" }, { "level": 1, "name": "swimming" } ]
+  },
+  {
+    "type": "profession",
+    "subtype": "hobby",
+    "id": "baseball_pitcher",
+    "name": "Baseball Pitcher (Expert)",
+    "description": "You're accustomed to striking out poor batters at the plate and can probably throw a mean knuckleball.  Now you get to practice your fastball, but perhaps with rocks or grenades this time.",
+    "points": 6,
+    "skills": [ { "level": 4, "name": "bashing" }, { "level": 7, "name": "throw" }, { "level": 3, "name": "swimming" } ]
```
snipercup, line 263:
```suggestion
    "skills": [ { "level": 4, "name": "bashing" }, { "level": 7, "name": "throw" } ]
```
I guess you forgot to remove the swimming skill? I don't see why a baseball_pitcher would have swimming 3

data/json/hobbies.json
```diff
@@ -342,20 +503,51 @@
...
+    "skills": [ { "level": 3, "name": "stabbing" }, { "level": 3, "name": "melee" }, { "level": 3, "name": "dodge" } ],
+    "traits": [ "MARTIAL_FENCING" ]
+  },
+  {
+    "type": "profession",
+    "subtype": "hobby",
+    "id": "fencing_expert",
+    "name": "Fencing (Expert)",
+    "description": "You've spent so long practicing your fencing you wouldn't look out of place in the Renaissance dueling with ne'er-do-wells and rapscallions.  En Garde!",
+    "points": 10,
```
snipercup, line 549: 10 points is a lot when compared to the 8 of diy_crafts_expert when you don't even get any proficency. Are these skills really good enough to justify 10 points? My suggestion is to lower it to 9 or raise the points of `diy_crafts_expert`. Alternatively, add a nice fencing sword or other items to ```fencing_expert```

data/json/hobbies.json
```diff
@@ -397,21 +598,51 @@
...
+    "subtype": "hobby",
+    "id": "blackbelt_hobby",
+    "name": "Black Belt",
+    "description": "You've been doing martial arts for years and have earned the coveted black belt in a martial art of your choice.",
+    "points": 10,
+    "skills": [
+      { "level": 5, "name": "melee" },
+      { "level": 5, "name": "unarmed" },
+      { "level": 5, "name": "dodge" },
+      { "level": 3, "name": "swimming" }
```
snipercup, line 612:
```suggestion
      { "level": 5, "name": "dodge" }
```
Probably forgot to remove swimming after copy-paste

data/json/hobbies.json
```diff
@@ -397,21 +598,51 @@
...
     ],
     "traits": [ "MARTIAL_ARTS5" ]
   },
+  {
+    "type": "profession",
+    "subtype": "hobby",
+    "id": "melee_training_expert",
+    "name": "Melee Training (Expert)",
+    "description": "Despite firearms making traditional melee weapons largely obsolete, you have extensive experience fighting in hand-to-hand combat with tools of war.  Maybe it was full-contact HEMA or SCA sparring, or maybe you're just a badass commando with a machete.  You start with your choice of Barbaran Montante, Bōjutsu, Eskrima, Fencing, Fior Di Battaglia, Medieval Swordsmanship, Niten Ichi-Ryu, Pencak Silat, or Sōjutsu.",
+    "points": 14,
```
snipercup, line 637: This is a whole lot of points when you aren't even getting any proficiency. You are getting a martial arts. I feel like 12 points should be enough unless you add a weapon. The description seems to suggest a machete.

data/json/hobbies.json
```diff
@@ -570,33 +801,92 @@
...
+    "traits": [ "TOUGH", "FLEET" ]
+  },
+  {
+    "type": "profession",
+    "subtype": "hobby",
+    "id": "football_expert",
+    "name": "Football (Expert)",
+    "description": "You played college ball, ya know.  You could have gone pro if the world hadn't ended.",
+    "points": 8,
+    "skills": [ { "level": 4, "name": "throw" }, { "level": 5, "name": "dodge" }, { "level": 3, "name": "swimming" } ],
```
snipercup, line 858:
```suggestion
    "skills": [ { "level": 4, "name": "throw" }, { "level": 5, "name": "dodge" } ],
```
Forgot to remove swimming?

snipercup:
Good work! You put a lot of effort into this. In general I'd say that some professions could do with some items, especially where the description suggests you have something, like the `airsoft gun` in the `plinking` profession. That's not really the focus of this PR though so I understand why no items are included here.

If you're talking about tutorializing, maybe it's a good idea to gate some professions behind achievements? You could do that with the `requirement` property. Before a player invests too much in swimming, I feel like it's not a bad idea to require a swimming achievement so the player has an idea what they're getting into. The same for some of the other professions. It's just a suggestion.

data/json/hobbies.json
GuardianDll: I remind you that `swimming` is id of athletics skill

Aerin-of-the-Toast:
> Good work! You put a lot of effort into this. In general I'd say that some professions could do with some items, especially where the description suggests you have something, like the `airsoft gun` in the `plinking` profession. That's not really the focus of this PR though so I understand why no items are included here.
> 
> If you're talking about tutorializing, maybe it's a good idea to gate some professions behind achievements? You could do that with the `requirement` property. Before a player invests too much in swimming, I feel like it's not a bad idea to require a swimming achievement so the player has an idea what they're getting into. The same for some of the other professions. It's just a suggestion.

Don't think we have any support for backgrounds giving items in the first place, only professions. Also, dunno if backgrounds can be locked behind achievements either. And I don't have the C++ chops to enable that, yeah.

data/json/hobbies.json
Aerin-of-the-Toast: As per Guardian, "swimming" is the id of Athletics, and I gave sporty backgrounds some athletics skill because, well, they're athletes.

data/json/hobbies.json
Aerin-of-the-Toast: Expert fencing and expert melee training have higher point values simply because starting with higher levels in combat skills is VASTLY more useful than crafting skills. Not that points matter anymore beyond just sorting the menu, but I digress. A character that starts with 5 in their melee skills (which previously was firmly mid to early lategame level of skill) can absolutely trounce early game enemies like zombies, even with no armor and early weapons. Add on how much of a power magnifier melee martial arts are and that's why the point costs are high.

Aerin-of-the-Toast:
> > > Good work! You put a lot of effort into this. In general I'd say that some professions could do with some items, especially where the description suggests you have something, like the `airsoft gun` in the `plinking` profession. That's not really the focus of this PR though so I understand why no items are included here.
> > > If you're talking about tutorializing, maybe it's a good idea to gate some professions behind achievements? You could do that with the `requirement` property. Before a player invests too much in swimming, I feel like it's not a bad idea to require a swimming achievement so the player has an idea what they're getting into. The same for some of the other professions. It's just a suggestion.
> > 
> > Don't think we have any support for backgrounds giving items in the first place, only professions. Also, dunno if backgrounds can be locked behind achievements either. And I don't have the C++ chops to enable that, yeah.
> 
> I was going off of what https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md describes and since you expressed doubt, I tested it locally. You can use items and requirements for hobbies, it just doesn't display items and requirements in the 'backgrounds' tab in-game. 
> 
> Also, if you look at the PR that introduced hobbies (#50262), it is specifically designed to select multiple hobbies. Moreover,  ```On the skills side, hobbies provide additional experience instead of flat levels as a balancer against multiple hobby selections.```. That makes sense if you look at the hobbies file, since there is no skill level over 4 currently.
> 
> My suggestion would be to adjust the professions instead of the hobbies because they are better suited for your use case.

Oh, you can add items to backgrounds? Neat, TIL! I still dunno if i want to add any this time, but maybe as a separate PR. Would be good for stuff like "CCL: you start with a handgun and holster" etc. 

As for backgrounds vs professions i mostly just disagree that backgrounds should be low. You can spend a lot of years doing more than one thing, for instance, and you can be an expert at something without it being your day job. If anything i just think professions *also* need a pass (some professions just give too few points in general but also it would be nice to have grades instead of most professions assuming you just started).

 As for hobby stacking the only real difference between someone taking just basketball (expert) vs taking all tiers of basketball is gonna be like 30% or so since exp requirements per level go up really high as skill numbers raise, so the xp to get 2 throwing will barely impact someone already at 6 throwing.

Issue #66040, opened 2023-06-06
3D vision lets you see unseen tiles below


Qrox:
### Describe the bug

It seems the newly added 3D vision (#65738) does not take the visibility of tiles below into account, letting you see what is below even if the tiles cannot be seen from the player character's position.

Pinging @Rewryte as the author of the change.

### Attach save file

N/A

### Steps to reproduce

1. Turn 3D fov on
2. Go into look around mode and look from z-levels above
3. Notice that the tiles not seen on the current z-level can be seen from above

### Expected behavior

Unseen tiles are hidden and map memory is displayed instead if available

### Screenshots

Unseen tiles in the current z-level:

![image](https://github.com/CleverRaven/Cataclysm-DDA/assets/11890223/b5da3a40-cf24-422b-bd18-704f92cefe4b)

The tiles are visible from two z-levels above:

![image](https://github.com/CleverRaven/Cataclysm-DDA/assets/11890223/4f2ecd68-d6b6-416f-b4df-01f13d874e07)

### Versions and configuration

bd8b8abd02b4d4d0022dd048e706344e884e4789

Rewryte:
Thanks, I'll take a look.

SurFlurer:
It's the problem that you cannot see the terrain below a furniture, which will be addressed by #65794

Qrox:
> It's the problem that you cannot see the terrain below a furniture, which will be addressed by https://github.com/CleverRaven/Cataclysm-DDA/pull/65954

Oh, I was not talking about that tile though, but the entire memorized but unseen area around it being visible from above.

SurFlurer:
Oh,sorry, I misunderstood.

Rewryte:
Looks like "looking around" gives players better visibility on higher z-levels.  3D vision inherits this when drawing lower z-levels.  We could calculate visibility for every z-level, but that might be computationally expensive and should have minimal benefit outside of "looking around".  I'll try to think an alternative.

Qrox:
The same happens when you are on a roof two z-levels above ground where you can see the unseen ground tiles near the walls even outside of look around mode.

SurFlurer:
Is #65205 related?

NetSysFire:
@Qrox Can you still reproduce? This looks like the kind of issue that might still be valid.

Qrox:
I haven't seen any PR that may fix this issue. Since the fix is likely non-trivial it's unlikely I missed it either.

[...]

PR #67434, opened 2023-08-04
Ledges provide sight coverage

Rewryte:
#### Summary
Features "Ledges provide sight coverage"

#### Purpose of change
Creatures ignore walls on their z-level when spotting characters on the z-level above.  A zombie staring at a wall should not be able to see a character on the roof at the opposite end.

#### Describe the solution
ledge
Introduce overly complex calculations to simulate how much of a character is obscured by a ledge when viewed by a creature below.  It takes into account the height of the ledge, distances of both creatures from the ledge, creature sizes of both viewer and target, and any furniture they are standing on.  The angle of the viewer's eye height to the ledge is extrapolated to the distance of the target to determine how much height the ledge covers at that point.

Part of the existing coverage system has been refactored to allow both ledge coverage and crouch/furniture coverage to interact nicely with each other.  Crouching and lying prone allows a character to remain hidden when ledge coverage is insufficient to cover their full body height.  Ledge and furniture coverage do not stack and only the highest of both is used.

#### Testing
Builds successfully.
New character created in the "Surrounded" scenario.
Teleported character to a rooftop.
Creatures on z-level below that are too close to the wall are unable to see the character, while those further away can.
Moving closer to the ledge make the character visible to more creatures.
Crouching and lying prone makes the character visible to less creatures.
No detectable increase in draw times.


Nebnis:
Next level update, thank you Rewryte for all the amazing QOL you been putting out for the game.

ashGlaw:
Gree! This is clutch. Way to go!

andrei8l:
Since our cover system is all-or-nothing wrt vision, this needs to be reciprocal. Meaning that I should not be able to see this zombie if it can't see me.
Screenshot from 2023-08-04 20-05-40
This should properly be solved from shadow casting code, but no one has been able to improve that in years and this solution works so whatevs 🤷.

Rewryte:
> Since our cover system is all-or-nothing wrt vision, this needs to be reciprocal. Meaning that I should not be able to see this zombie if it can't see me.

Fair enough.  I always had a mind to update the light / shadow system to work better with 3D vision, but ended up deferring the task.  Just making ends of the code is an undertaking on its own.  I'm working on something else at the moment but I'll have another go then I'm done.

kevingranade:
The shadowcasting test needs to be updated to mesh with this.
Thanks a ton for this,  the way this works has really bugged me for a long time but I haven't been able to get around to it.

Rewryte:
> The shadowcasting test needs to be updated to mesh with this.

Sure, will get to this soon.  Currently looking into #66040 which is hampering the new lightmap changes.

Rewryte:
This is fully functioning now.  However, there is a noticeable performance impact on elevated areas such as open rooftops as ledge coverage is calculated for each grid in the area below.  Looking into ways in which performance can be improved.

Rewryte:
Seems to be some discrepancy in visibility between creatures looking up a ledge, and the player lightmap when looking down at the same creature.  Looking into this.

Rewryte:
Should be good to go.  Shadowcasting might sporadically stop working for tiles 2 or more z-levels below the player but such is the behavior prior to this PR.  Have yet to ascertain the cause.  The same area also seems to have its own ledge shadowcasting unrelated to mine.  This behavior will be effectively obscured by the seen_cache post-processing introduced in this PR.

PR #67441, opened 2023-08-04
Improve turn performance when having many solar panels

inogenous:
#### Summary
Performance "Improve turn performance when having many solar panels"

#### Purpose of change

When having many solar panels, each turn can take a long time, most noticeable when sleeping or waiting. This change attempts to reduce the number of calls during each turn without changing any behaviour.

This was verified using a setup of 20 adjacent solar panels in a x3 grid, as pictured below.

rumney-layout

A callgrind trace was done to measure which parts of `do_turn()` take up work, pictured below. The callgrind trace shows that **31%** of the work in `do_turn()` comes from `map::vehmove()`, which in turn calls `vehicle::search_connected_vehicles`. In effect, `search_connected_vehicles` takes up 27% of the work in `do_turn()`. The callgrind trace measured 177 calls to `vehmove`, which in turn resulted in 531693 calls to `find_vehicle` - which gives approximately **3004** calls to `find_vehicle` for each turn.

callgrind out rumney-master

This issue relates to #57152 , which was resolved in #60776 , but the issue seems to have reappeared since then.

#### Describe the solution

The existing method `vehicle::search_connected_vehicles` uses a variant of Dijkstra's algorithm to find the shortest path to each vehicle. When doing this, it traverses many of the vehicles multiple times, comparing the found path lengths each time and choosing the shorter one. But the calls from `vehmove` shown in the callgrind trace above, do not in fact require us to find the shortest path - it only needs the list of unique vehicles. This makes it possible to simplify the vehicle traversal done in `map::vehmove` to only step through each vehicle once.

The simpler algorithm that only finds unique connected vehicles is implemented in a new function `get_connected_vehicles`, which populates the supplied set.

With this change applied, a callgrind trace was done using the same map as above. The new callgrind trace now shows that **9%** of the work in `do_turn()` now comes from `map::vehmove()` (was 31% before). The callgrind trace measured 277 calls to `vehmove`, which in turn resulted in 172026 calls to `find_vehicle` - which gives approximately **621** calls to `find_vehicle` for each turn (was 3004 before, as shown above).

callgrind out rumney-get_connected_vehicles

* Merging adjacent solarpanels into a single but larger vehicle, as done in #60681
* The current implementation of Dijkstra's algorithm in `vehicle::search_connected_vehicles` might be incorrect. Other implementations usually start each step of the loop by finding the first unvisited node with the lowest weight. The current implementation seems to pick any that was put on the queue. This might make it traverse each node multiple times, which is not needed if implemented correctly.
* It is unclear whether the calls to `search_connected_vehicles`/`get_connected_vehicles` done by `vehicle::gain_moves` each 5 turns are actually needed. It might be possible to just remove those calls.

#### Testing

* Verified that batteries are still being charged/discharged, appliances still work etc

#### Additional context

Hour debug timer before:

rumney-hourtimer-before

After:

rumney-hourtimer-after

### DIFF
--- a/src/iuse_actor.cpp
+++ b/src/iuse_actor.cpp
@@ -4811,7 +4811,7 @@ std::optional<int> link_up_actor::link_to_veh_app( Character *p, item &it,
                     ( it.link->t_abs_pos + prev_veh->coord_translate( it.link->t_mount ) ).raw(),
                     it.link->t_abs_pos.raw() );
 
-        for( const vpart_reference &vpr : prev_veh->get_any_parts( "POWER_TRANSFER" ) ) {
+        for( const vpart_reference &vpr : prev_veh->get_any_parts( VPFLAG_POWER_TRANSFER ) ) {
             if( vpr.part().target.first == prev_part_target.first &&
                 vpr.part().target.second == prev_part_target.second ) {
                 p->add_msg_if_player( m_warning, _( "The %1$s and %2$s are already connected." ),
--- a/src/map.cpp
+++ b/src/map.cpp
@@ -635,6 +635,7 @@ void map::vehmove()
     }
     dirty_vehicle_list.clear();
     std::map<vehicle *, bool> vehs; // value true means in on map
+    std::unordered_set<vehicle *> connected_vehs;
     for( int zlev = minz; zlev <= maxz; ++zlev ) {
         const level_cache *cache = get_cache_lazy( zlev );
         if( !cache ) {
@@ -642,10 +643,11 @@ void map::vehmove()
         }
         for( vehicle *veh : cache->vehicle_list ) {
             vehs[veh] = true; // force on map vehicles to true
-            for( const std::pair<vehicle *const, float> &pair : veh->search_connected_vehicles() ) {
-                vehs.emplace( pair.first, false ); // add with 'false' if does not exist (off map)
-            }
+            veh->get_connected_vehicles( connected_vehs );
         }
+    }
+    for( vehicle *connected_veh : connected_vehs ) {
+        vehs.emplace( connected_veh, false ); // add with 'false' if does not exist (off map)
     }
     for( const std::pair<vehicle *const, bool> &veh_pair : vehs ) {
         veh_pair.first->idle( /* on_map = */ veh_pair.second );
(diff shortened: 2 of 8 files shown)


ashGlaw:
Great idea! Anything that makes solar panels more cpu efficient is welcome

anoobindisguise:
Does a single vehicle process all of its solarpanels at once as a single block? So you still might want to just use a vehicle instead of appliance panels despite these very awesome improvements?

inogenous:
> Does a single vehicle process all of its solarpanels at once as a single block? So you still might want to just use a vehicle instead of appliance panels despite these very awesome improvements?

@anoobindisguise thank you for your question, it prompted me to take a closer look on things that were not clear, at least not to me, before.

So looking at the map pictured above with the 20 solar panels, let's say topleft is (0,0). At the moment, these 20 solar panels are currently implemented in the game as 20 vehicles. Each vehicle has a number of parts. In particular, each vehicle has a part with flag  `POWER_TRANSFER` for each connection it has to other vehicles. Think extension cords. So for example the vehicle at (0,0) has **four** vehicle parts:
* one obviously for (0,0) which is the panel itself
* but it also has (invisible) parts at (0,1) with flag `POWER_TRANSFER` 
* (1,0)
* and (1,1)

Each of these  `POWER_TRANSFER` vehicle parts are connectors that connect the vehicle to other vehicles (other solar panels). So the vehicle at (0,0) has four parts, right. 

The vehicle to the right of it, at (1,0), has six vehicle parts: at (1,0) obviously, and `POWER_TRANSFER` parts at (2,0), (2,1), (1,1), (1,0) and (0,0).

So each solar panel is represented as the panel itself, but also several parts for `POWER_TRANSFER` connectors. Also, please note that each solar panel, even if adjacent, has connector parts going both ways: for example, the vehicle at (0,0) has a part just to the right of it at (1,0) to the vehicle at that exact location, (1,0). And the vehicle at (1,0) has a connector part just to the left of it at (0,0) that connects to the vehicle at (0,0). 

So with all of this, the 20 solar panels pictured above will result in 129 vehicle parts if i counted correctly. Additionally, as stated above, each part connects both ways, so the number of connections between parts (# if edges in the graph) is something approximately like 200 for 20 panels (# of nodes in the graph).

The current way of representing the solar panels as separate vehicles requires us to keep in memory far more parts and connectors than number of solar panels itself, in this case 10x the number of panels. If we could represent these adjacent solar panels as a single vehicle with separate parts, i would *guess* that there would be lots of performance-related improvements to be made, possibly also in other places of the code.

But that's just a guess, it's an interesting direction you suggest @anoobindisguise . Let's see what #60681 brings, which looks fscking awesome :heart:  .

PR #67459, opened 2023-08-05
Adds proper one lane nests for the rest stop/layby

Procyonae:
#### Summary
None

#### Purpose of change
Adds some nests to make the laybys fit in better that I cut from #67223

#### Describe the solution
image
Before

image
After


ashGlaw:
What's a layby? That's not a term I'm familiar with

Procyonae:
> What's a layby? That's not a term I'm familiar with

Nice I somehow managed to miss that that's a UK term "A paved area at the side of a highway designated for drivers to stop in, for emergency parking, or where vehicles can wait, with larger lay-bys possibly having facilities like food vendors or public telephones."
It's not named in game anyway though

Zireael07:
@Procyonae The game is set in the USA, so UK terms are... oopsie imho

Procyonae:
> @Procyonae The game is set in the USA, so UK terms are... oopsie imho

Ik, like I said the word doesn't appear in game so what's your point?

witchwrenna:
I would expect an American to say a rest area or service area. See e.g. https://www.interstaterestareas.com

Zireael07:
@Procyonae My point is people used to USA spellings won't know what this is (heck, I'm an English major and I didn't know what a layby is, but I know the USA term)

Procyonae:
> I would expect an American to say a rest area or service area. See e.g. https://www.interstaterestareas.com/

They don't appear synonymous because that appears to mean there's a building and that it's off a primary road, we use that term here too this is just a little place to pull over. Also we already have a rest area in game so that would be confusing too. If you're that upset over it I suggest you fix connection code so it doesn't occur in the first place 😉

witchwrenna:
Alright yeah you're right there's a bigger difference here than I thought.

The american rest stop is more synonymous with the British motorway service area, which is what's already in the game as you mentioned. A place with toilets, a gas station, food, etc. Point is, it's a little more built up with facilities.

When it's smaller, not really any facilities, usually just a bit of gravel or the asphalt is extended out a bit, in america it may simply be "the shoulder" of the road, a scenic area, a turnout or a pullout. I don't know enough about american highway culture to really say what's the best word to use here!

Not trying to be pedantic, I just legit never heard layby until today and got curious. If I knew any C++ maybe I could help with the connection code but I have to admit my only experience is a semester long class around 10 years ago.

PR #67467, opened 2023-08-05
Add more stuff to Alpha

anoobindisguise:
#### Summary
Content "Make Alpha more cool & flavorful"

#### Purpose of change
Alpha mutants should be more full of themselves. Currently you just sort of get the single real downside of disintegration. I wanted to add more stuff to remedy that.

#### Describe the solution
- Adds two vanilla mutations to Alpha: Insomnia (side effect of you becoming less needful of sleep, plus you spend all night worrying about the future and/or scheming due to your grandiose designs implied by your dreams) and Bad Temper (You're perfect, so everything should just be going your way. If that's not happening, it means the world must be conspiring against you!)
- Adds a new mutation, which is also choosable as a starter mutation: Picky Eater (Prevents you from eating anything with negative joy unless you are Famished or hungrier). It's pre threshold.
- Adds a new mutation, Highly Vain: You love your face so much that you get sad when you cover it up and you don't get to show it off. Morale penalty for wearing mouth/eyes coverage. Requires Beautiful or Very Beautiful.
- Adds a new mutation, Superiority Complex: You hate stooping to the level of other people (who are all worse than you and don't deserve that level of respect). -persuasion, +intimidate/lie. Post thresh only, requires 2/4 prime mutations.
- Makes Disintegration post thresh, and require 1/4 prime mutations. Increased damage. This is based on the terminal entry stating that it arises from "mutagen overdose".
- Makes Prime Intelligence cause you sadness when you fail a craft with a specially coded message. Destroying components will increase the penalty over just progress loss.

The goal is to make Alpha more risky the more you try to get out of it. 

- Make Gourmand conflict with Picky Eater.
- Make Superiority Complex conflict with Truth Teller.

#### Testing
Confirmed things work in game after compiling!

#### Additional context
Here's an example of how Picky Eater works.
picky eater


sadenar:
Thoughts about having picky eater specifically only happen above a certain hunger/body weight threshold? I think everything here is very cool and definitely solves the problem of Alpha just having way too few mutations, but I think if an Alpha was literally starving they'd still be smart enough to just suck it up but take a massive morale loss for it.

I just think it doesn't make much sense for an Alpha mutant to just starve themselves to death rather than eat yucky food.

anoobindisguise:
> Thoughts about having picky eater specifically only happen above a certain hunger/body weight threshold? I think everything here is very cool and definitely solves the problem of Alpha just having way too few mutations, but I think if an Alpha was literally starving they'd still be smart enough to just suck it up but take a massive morale loss for it.
> 
> I just think it doesn't make much sense for an Alpha mutant to just starve themselves to death rather than eat yucky food.

I actually already did this, if you are Famished or hungrier (which requires you to be Underweight or more) the trait "turns off". I could add a morale penalty for it on top of that or make it only apply at Emaciated instead of Underweight.

andrei8l:
`Highly Vain` and `Superiority Complex` sound like stereotypical Elf attributes rather those of a "perfect human".

Nebnis:
A perfect human would have followers maybe then a boost to having them and a negative for doesn't?

fairyarmadillo:
> `Highly Vain` and `Superiority Complex` sound like stereotypical Elf attributes rather those of a "perfect human".

Alpha mutants are defined by their superiority complex in every bit of lore that discusses them, particularly the transcendent mutant and mutant child. Elf-A is a modified version of Alpha that has less of its superhumanity in exchange for some mild plant, slime and bird traits, and its stated purpose is to allow the last survivors to form small communities in the wilderness. I don't think either trait would be super appropriate for Elf-A as it is not a stereotypical fantasy elf, just an example of what a designer mutation might look like had the project been allowed to continue.

These traits were introduced in order to give Alpha some unique downsides. Splashing it out to every other mutant because Elves are jerks in D&D is a bad idea.

ampersand55:
What about adding Psychopath / Sapiovore as a tie-in to Superiority Complex?

Profugo-Barbatus:
Psychopath might work in a sort of "No longer sees them as equivalent to you" sorta way, how normal survivors dismiss zombies as not really human/worthy of guilt in most cases. Sapiovore sounds like a stretch though - Its one thing to not think much about killing the ignorant fool standing in the way of a god, its another to turn them into a sandwich. If anything, I'd expect an Alpha with an inflated ego to have more moral restrictions on food - After all, you're the best of the best, why shouldn't you be eating like a king?

fairyarmadillo:
Sapiovore is a mutation that happens because your brain is 50% bear and bears eat people. A ton of mutants already have it and splashing it out to questionable ones is only going to make it less special.

Psychopath is wrong too I think. Nothing about Alpha suggests it's cold and unfeeling, it actually seems quite the opposite what with all the vanity.

Zireael07:
"Vanity" and "attracts followers" do not seem (at least to me, a non-native English speaker) like they belong in the same sentence

fairyarmadillo:
> "Vanity" and "attracts followers" do not seem (at least to me, a non-native English speaker) like they belong in the same sentence

Vanity just means you value yourself more highly than others. Vain people often need adulation and respect in order to feel secure. Just look at the kings and dictators throughout history who have built up huge cults of personality despite being horrible people.

data/json/mutations/mutations.json
```diff
@@ -9145,5 +9163,37 @@
     "category": [ "CEPHALOPOD" ],
     "prereqs": [ "INT_UP_4" ],
     "threshreq": [ "THRESH_CEPHALOPOD" ]
+  },
+  {
+    "type": "mutation",
+    "id": "VANITY",
+    "name": { "str": "Highly Vain" },
+    "points": -1,
+    "description": "You know you're beautiful, and the world deserves to know it too.  You prefer to show off your stunningly good looks to the world and will be unhappy if you cover up your face.",
```
Maleclypse, line 9172: This is an actual reason why many people refuse to wear masks during the pandemic. Great addition.

data/json/mutations/mutations.json
```diff
@@ -9145,5 +9163,37 @@
...
+    "description": "You're a very picky eater.  You refuse to eat any food that tastes bad unless you're starving.",
+    "category": [ "ALPHA" ],
+    "starting_trait": true
+  },
+  {
+    "type": "mutation",
+    "id": "ALPHA_COMPLEX",
+    "name": { "str": "Superiority Complex" },
+    "points": 0,
+    "description": "You can't stand to prostrate yourself before your lessers and hate it when they don't immediately do what you say.  Persuasion is much more difficult, but you're better at lying and intimidating.",
```
Maleclypse, line 9190: Ooh now we could add an EOC that checks followers for this trait and randomly causes them to shift towards mutiny.

Maleclypse (commented):
Great additions!

PR #67470, opened 2023-08-05
Jsonify trail's maps

Procyonae:
#### Summary
Content "More trail variety"

#### Purpose of change

Hardcode bad
Fixes trailheads not rotating

#### Describe the solution

Unhardcodes trail maps, doesn't unhardcode placement (May implement in a future PR using #67736)
Adds more trail variety (Though not enough, please suggest more variants for a followup PR)
Adds trail connections to additional appropriate maps
Adds small and large animal dens for groundhogs/hares/foxes/bears

Some of the region settings have been replaced with palettes/overridable nests, look at the included desert mod changes for a basic example of changing the trail terrain and size in accordance to it's old region settings, happily DM/ping me here or on discord if you're confused about that or can't figure out how to override the nests.

Due to the change to fallback predecessor (from what was essentially predecessor forest_thick) the horse farms trail is kinda borked now bc it often spawns on field. The simple solution of making it spawn forest first doesn't work bc it errors at placing two maps at the same coords. Can't think of a good alternative other than making a JSON forest map and using a nest version, replicating forest in JSON isn't ideal though bc it's heavily dependant on region settings iirc.

Adding more nests to forest_trail_intersection

Replacing path dirt with mud rarely in swamps to make them more different but I want to make changing the path terrain piss easy for modders so I'd rather not

#### Testing

Tele'd around lots

#### Additional context
<details>

image

image

image

image

image

image

image

image

image

</details>


Procyonae:
I have a minor commit to do before this gets merged I just want to let the tests finish first

Maleclypse:
> I have a minor commit to do before this gets merged I just want to let the tests finish first

Just let me know when it's ready. In discord if possible.  if not @ me here.

Maleclypse:
I hope you come back to this. It's pretty cool!

Maleclypse:
You asked for bugs, here is a tiny one :)

Procyonae:
This wants more variation but it's better than what we have now at least

PR #67476, opened 2023-08-06
Canceling "save zone" query returns you to zone manager instead of discarding changes

Qrox:
#### Summary
Bugfixes "Canceling 'save zone' query returns you to zone manager instead of discarding changes"

#### Purpose of change
Give you one last chance when you misclick.

#### Describe the solution
1. Create a function `query_ynq` similar to `query_yn` that allows and defaults to canceling.
2. Change the zone manager code to support canceling the save zone query and use `query_ynq` for the save zone query.
3. Use `query_ynq` in the diary UI for the save diary query.
4. Change description of "QUIT" of the "YESNOQUIT" action to "Cancel" to reduce confusion.

#### Testing
Zone manager was displayed correctly and pressing ESC showed the save zone query, and selecting yes, no, and cancel saved changes, discarded changes, and returned to the zone manager respectively.

Selecting yes, not, and cancel in the save diary query saved changes, discarded changes, and returned to the string editor window respectively.

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -6859,7 +6859,6 @@ void game::zones_manager()
     } );
     ui.mark_resize();
 
-    std::string action;
     input_context ctxt( "ZONES_MANAGER" );
     ctxt.register_navigate_ui_list();
     ctxt.register_action( "CONFIRM" );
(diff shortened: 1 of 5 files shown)


src/output.h
```diff
@@ -428,6 +428,16 @@ inline bool query_yn( const char *const msg, Args &&... args )
     return query_yn( string_format( msg, std::forward<Args>( args )... ) );
 }
 
+enum class query_ynq_result {
```
inogenous, line 431: I think your idea is great. Minor nitpick: how would you feel about consolidating "cancel" and "q"? Maybe `query_ync` instead?

Qrox: Codewise it uses the `QUIT` action and the user-facing description is `Cancel` so I feel it's fine this way.

Qrox: I was trying to keep the function name consistent with the input category/action names, so I'll probably just rename "cancel" to "quit" here. Making the code consistent with the user-facing description requires much more changes and I doubt it's worth the effort though.

src/game.cpp
```diff
@@ -7186,6 +7210,25 @@ void game::zones_manager()
             .edit( facname );
             zones_faction = faction_id( facname );
             zones = get_zones();
+        } else if( action == "QUIT" ) {
+            if( stuff_changed ) {
+                const query_ynq_result res = query_ynq( _( "Save changes?" ) );
+                switch( res ) {
+                    case query_ynq_result::cancel:
```
Qrox, line 7217:
```suggestion
                    case query_ynq_result::quit:
```

src/output.cpp
```diff
@@ -849,6 +849,35 @@ bool query_yn( const std::string &text )
...
+        .cursor( 2 )
+        .default_color( c_light_red )
+        .query()
+        .action;
+    if( action == "NO" ) {
+        return query_ynq_result::no;
+    } else if( action == "YES" ) {
+        return query_ynq_result::yes;
+    }
+    return query_ynq_result::cancel;
```
Qrox, line 878:
```suggestion
    return query_ynq_result::quit;
```

src/output.h
```diff
@@ -428,6 +428,16 @@ inline bool query_yn( const char *const msg, Args &&... args )
...
+    cancel, no, yes
```
Qrox, line 432:
```suggestion
    quit, no, yes
```

PR #67482, opened 2023-08-06
bio_cqb won't provide things it can't possibly know about anymore.

waxcatape:
#### Summary
Content "Removes all martial arts from the Close Quarters Battle bionic except for biojutsu"

#### Purpose of change

This is extradimensional technology that is somehow aware of this Earth's martial arts and combat sports. In light of the current lore, the inclusion of almost all of the martial arts in there doesn't make sense. 

#### Describe the solution

Remove everything except for Bionic Combatives. This leaves the CQB bionic with its actual unique style, one that would actually make sense for it to feature. Some changes to the descriptions of the techniques have also been made. The original ones were jokey (which isn't in itself a problem) and referenced way too much of current Earth (which is).

Make the bionic read the contents of any integrated ebook storage on you and any martial arts found in any martial arts manuals to the list of styles it provides. I couldn't make it work. If someone wants to try, I believe that it'd be a straight upgrade over this pr.

#### Testing

Made the changes, spawned and installed the bionic, checked that the only style provided is Bionic Combatives and that the initial message has changed.


masema:
While I agree that the CBM shouldn't know about our earth's culture, considering that Exodus comes from a dimension quite like ours, it's nearly inevitable that they would develop martial arts quite similar to ours. In my opinion, this isn't anything lore breaking.

waxcatape:
> While I agree that the CBM shouldn't know about our earth's culture, considering that Exodus comes from a dimension quite like ours, it's nearly inevitable that they would develop martial arts quite similar to ours. In my opinion, this isn't anything lore breaking.

The Exodii don't come from any one dimension and the dimensions can get very different. We can assume that whatever place they picked this CBM up has martial arts. It wouldn't be outlandish to think that these martial arts feature striking, takedowns and grappling. But otherwise you can't really say anything with a whole lot of certainty. Out of the two Exodii we've met so far, one is from some kind of English speaking culture, but Anglic is so different from modern English that you can't even reliably call the two languages mutually intelligible. Liliya you can't really talk to at all, even if you happen to know some Arabic. So it's safe to say some major cultural differences apply. Not to mention that the CBMs are whatever technology they pick along the way. This thing could've been from anywhere.

masema:
>  We can assume that whatever place they picked this CBM up has martial arts. It wouldn't be outlandish to think that these martial arts feature striking, takedowns and grappling. 

So, it's quite possible that they would be aware of, just to pick two examples of martial arts you've removed from the cbm, bojutsu (the art of using a bo staff) or swordsmanship. In fact, it's guaranteed they would know of swordsmanship because one of the other CBMs available is a sword. 

My point is, no single martial art that is provided by the CBM is something that they couldn't have found in another dimension.

waxcatape:
> So, it's quite possible that they would be aware of, just to pick two examples of martial arts you've removed from the cbm, bojutsu (the art of using a bo staff) or swordsmanship. In fact, it's guaranteed they would know of swordsmanship because one of the other CBMs available is a sword.
> 
> My point is, no single martial art that is provided by the CBM is something that they couldn't have found in another dimension.

For one thing, swordsmanship is not some generic swordsmanship, it's medieval swordsmanship explicitly stated to be based on German and Italian traditions. Which may or may not exist in wherever that thing came from. Same with the bo staff. How can we know that the origin dimension ever had one? Something similar, sure. But if you're going to provide some certain martial arts and not the others, then which ones are we to provide?

Zireael07:
I'd change the lore, making the CBM NOT be extraterrestrial in origin. This neatly solves the problem without nerfing one of the more interesting CBMs (many are just straight buffs to a certain playstyle, and this one is unique in _opening_ many options instead)

masema:
> For one thing, swordsmanship is not some generic swordsmanship, it's medieval swordsmanship explicitly stated to be based on German and Italian traditions. 

Was not aware of this bit, thank you for clarifying 

> Which may or may not exist in wherever that thing came from. Same with the bo staff. How can we know that the origin dimension ever had one?

We can't know that the origin dimension didn't have one.

> Something similar, sure.

Similar enough that an average person could confuse the two?

> But if you're going to provide some certain martial arts and not the others, then which ones are we to provide?

What I'm trying to say is, in my opinion, the list of martial arts is not broken. It should remain exactly as it is, without modifications.

waxcatape:
> What I'm trying to say is, in my opinion, the list of martial arts is not broken. It should remain exactly as it is, without modifications.
I get that, I just don't agree.

> Similar enough that an average person could confuse the two?
An average person could confuse Muay Thai and Karate. That's not an argument.

> We can't know that the origin dimension didn't have one.
True but if we were to provide everything we can't know this thing is aware of, we'd have to provide every martial art that has ever existed. And a lot of them that never did as well, that dimension is bound to have a few MAs that nobody ever developed around here. The end result simply reads gamey and not thought through lorewise in my opinion.

Maleclypse:
@dseguin Do you have any ideas how to add an ebook pocket to this CBM so that the player can add any martial arts they to this CBM's repertoire?

Profugo-Barbatus:
> So it's safe to say some major cultural differences apply. Not to mention that the CBMs are whatever technology they pick along the way. This thing could've been from anywhere.

As long as the beings on the other side are human, the unarmed martial arts would only vary from ours in name, and perhaps a couple specific moves that are more or less commonly spread out. Its less of a cultural thing and more of a biomechanics thing - There's only so many effective ways to generate lethal force with human anatomy, and only so many effective ways to transition between and defend from those actions.

> But if you're going to provide some certain martial arts and not the others, then which ones are we to provide?

For armed martial arts, I'd agree with this more - The design of a weapon massively influences the combat arts that are available to it, along with local culture, customs and availability regarding armor, scale of combat, etc. Even with a sword that mostly resembles an earth one, the way you use it would be radically different if you were expected to use it in honorable one on one duels with the enemy leadership, or if you were expected to use it to slay ten foot tall gigachad bear variants, or anything else that might be different under the sun.

If I were to wave a magic wand, I'd have the module provide the non-fantasy martial arts, bionic combatives, and then either make combatives encompass proper weapons of war, or create an "Amalgamation" style or some such that represents the CBM blending a whole bunch of potentially alien techniques to try and make use of most weapons. It should be inferior to bionic combatics and 'proper' techniques for our weapons, but better than just brawling with a longsword - After all, maybe that sweep style move was from a culture who's heavily armored foes lead them to forsake thrusting tips in favor of hooked ends on their blades to bring the foe down for a dagger strike. Without that, its harder to pull the move off.

So in the end, its best to use it with bionic weapons, it knows all the proper body movements for effective strikes in techniques similar enough to what we know unarmed, and 'tries its best' with what we've got in the rest.

waxcatape:
@Maleclypse the problem I had wasn't adding a pocket, it was getting a list of ebooks out of it. I'm not sure why but I just couldn't coerce it into giving me any kind of container of ebooks in the pocket that I could iterate through. Tried both with a dedicated MA pocket and with making the CQB CBM dependent on the Integrated AR System and using that CBMs ebook pocket.

John-Candlebury:
I dont think this is a good solution to a pretty minor lore problem. 

he venoms arent real martial arts and are still in the game AFAIK. I would replace the martial arts given by this bionic with those.

waxcatape:
I've thought a bit abt this and it just feels like a cop out. If I ever get to making the alternative work, I'll try making a PR that properly reworks this CBM into something interesting. In the meantime, this should be closed

LlubNek:
If the CBM can acquire martial arts from ebooks (nvm electronic compatibility and language issues...), then it should probably come loaded with a random assortment of martial arts, rather than a complete set or just one.

PR #67483, opened 2023-08-06
Mutation selector mod

Rewryte:
#### Summary
Mods "Mutation selector mod"

#### Purpose of change
The mutation system leaves plenty of decisions to RNG.  I think some players would appreciate being offered mutation choices for story and roleplay purposes, or to better suit their playstyle.

#### Describe the solution
When the player character acquires a mutation, a selector menu appears which allows players to choose from a list of possible mutations.  Only mutations which the player is eligible for is shown on the list, and mutation category restrictions still apply.  It works with "true random" mutations which can result in a bigger list of mutations to choose from.  As in the base game, certain mutations, such as purifier mutations, may have to be mutated more than once to achieve the desired outcome as conflicting mutations are gradually removed.

Capture

This feature is a balance mod which has to be added into a world before it is enabled.  It is safe to add to an existing save.

Make an option instead of a mod.
Remove flavor text to make the menu much smaller.
Allow mutations to be selected for NPCs.

#### Testing
Builds successfully.
A fresh character was given mutagen and primer.
Waited in game until mutation happens.
Mutation selection menu appears correctly when mod is activated.
Mutations with missing prerequisites do not appear on list.
Mutations with threshold requirements only appear when threshold unlocked.
Mutations with bionic conflicts do not appear on list.
Purifier mutations only appear when applicable and correctly downgrades a conflicting mutation by one level.
Only mutations with the correct primer category is shown on list.
List does not show duplicate entries or existing mutations.
Mutations are applied correctly upon selection.

### DIFF
--- a/data/core/game_balance.json
+++ b/data/core/game_balance.json
@@ -355,5 +355,12 @@
     "info": "Sets eternal weather.  Possible values are 'normal', 'clear', 'cloudy', 'light_drizzle', 'drizzle', 'rain', 'rainstorm', 'thunder', 'lightning', 'flurries', 'snowing', 'snowstorm', 'early_portal_storm', 'portal_storm'.  'normal' clears all eternal weather overrides and sets normal weather pattern.  Requires restart of the game after changing value to take effect.",
     "stype": "string_input",
     "value": "normal"
+  },
+  {
+    "type": "EXTERNAL_OPTION",
+    "name": "SHOW_MUTATION_SELECTOR",
+    "info": "When mutating, displays a menu which allows players to pick from a list of possible mutations.",
+    "stype": "bool",
+    "value": false
   }
 ]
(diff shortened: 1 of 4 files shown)


alp7292:
is it compatible with othermods lile dinomod magicalysm

Rewryte:
> is it compatible with othermods lile dinomod magicalysm aftershock xedra? awesome work btw

Good question.  This should work with pretty much everything.

baldamundo:
Really excited to see this! Have been playing Caves of Qud recently and really enjoying the mutation system in that and was just thinking it would be cool for Cataclysm to have an option that works more similarly. In Qud, whenever you buy a new mutation, you get to pick one out of a random shortlist of three different mutations selected by RNG - which i wonder if it might be a fun option here as a halfway house between full RNG and totally free selection.

Zireael07:
I like the compromise @baldamundo suggested <3

Rewryte:
Believe it or not, I only found CDDA because I was looking for something like Qud.  But yeah, even in that game I never really used the random mutation system because RNG really isn't my thing.  I'll consider a pick-one system as a future option for this, but it probably won't be the default because the whole purpose of this mod is to remove the RNG element.

PR #67489, opened 2023-08-06
Fix a couple of drug spawns

Maleclypse:
#### Summary
None

#### Purpose of change
Fix a couple of food spawns

#### Describe the solution
Converting from charges to count in comestibles

Waiting for issues to get made

### DIFF
--- a/data/json/itemgroups/Drugs_Tobacco_Alcohol/drugs.json
+++ b/data/json/itemgroups/Drugs_Tobacco_Alcohol/drugs.json
@@ -74,9 +74,27 @@
     "type": "item_group",
     "//": "Painkillers balanced according to rarity excluding illicit drugs",
     "distribution": [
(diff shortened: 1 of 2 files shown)


Maleclypse:
too many casserole pots
@mqrause I've clearly made these incorrect.  Can you let me know what I should be doing instead?

mqrause:
Yeah, `count` doesn't play well with `container-item` or default containers currently. You have to make groups like this one for example:

https://github.com/CleverRaven/Cataclysm-DDA/blob/9e021bfcac7af3380da0dee6eea75ede409fad8b/data/json/itemgroups/collections_domestic.json#L1768-L1775

So separate group with it's own dedicated container and the quantities of the item you want. And if a default container exists, override it with null, or you'll get nested containers.

If you find that annoying and convoluted, you can wait for #67410 to present an easier option.

PR #67492, opened 2023-08-06
Portal lightnings

GuardianDll:
#### Summary
None
#### Purpose of change
So we have portal storm, but what a storm it is without lightnings?
#### Describe the solution
Simple EoC, that draw a clairvoyance line between randomly picked points around playet
#### Issues
For unknown for me reason, only furniture has a special treatment, that allow it to use `f_null` as `allowed to spawn on all types of furniture`, and both `t_null` or, which is important for this PR, `fd_null` cannot be used for this purposes. I believe to cover this i need to copy this piece of code <https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/magic_ter_fur_transform.cpp#L304>, but i can't do it for myself
#### Screenshots
<details>
  <summary>Because of the issue i have only close visualisation of the effect, that spawn tatami mats, without actual clairvoyance</summary>

screenshot is made from zlevel 3

image

</details>

### DIFF
--- a/data/json/portal_storm_effect_on_condition.json
+++ b/data/json/portal_storm_effect_on_condition.json
@@ -305,6 +305,7 @@
     "effect": {
       "weighted_list_eocs": [
         [ "EOC_PORTAL_GRASS_CHANGE", { "const": 10 } ],
+        [ "EOC_PORTAL_LIGHNING", { "const": 6 } ],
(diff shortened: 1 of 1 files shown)


GuardianDll:
i'll reopen it when i'll make something i'll be satisfied with

PR #67493, opened 2023-08-06
Map Editor displays ter_str_id/furn_str_id instead of names

Procyonae:
#### Summary
None

#### Purpose of change
Ids are more useful than names

Having both but the text already overflows the box sometimes

#### Testing

image


ZhilkinSerg:
We should definitely expand these boxes someday for non-80x25 terminals.

NetSysFire (approved):
Small thing that bothered me, too. Thanks!

PR #67494, opened 2023-08-06
Enable/disable showing several non-player-related messages in the log

Night-Pryanik:
#### Summary
Interface "Enable/disable showing several non-player-related messages in the log"

#### Purpose of change
Currently message log is spammed by tons of potentially unneeded information, such as "monster misses other monster" or "duck flies over the fence". I am very annoyed by such message spam, so I added an option group where player can enable or disable showing of such information.

#### Describe the solution
Ported https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/45:
- Added 'Additional messages in the log' option group
- Added options for showing/hiding messages for
- - "You see here x items" when player steps on tile with items
- - "Monster flies over the fence"-type of messages
- - "Monster leaps!"
- - Monster struggles to stand up after being knocked down
- - Monster/NPC hits/misses/bites/lunges/claws/throws other monster/NPC
- - Monster/NPC avoids incoming projectile from other monster/NPC
- All options defaults to true.

None.

#### Testing
Turned all options off, teleported into a city. No more message spam, until occasional "monster dies" or "you hear a crash".

#### Additional context
изображение

If there's some other message you want to hide, please tell me, I'll try to hide it under an option.

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -1128,13 +1128,14 @@ void Creature::deal_projectile_attack( Creature *source, dealt_projectile_attack
             add_msg_player_or_npc(
                 m_warning,
                 _( "You avoid %s projectile!" ),
-                _( "<npcname> avoids %s projectile." ),
+                get_option<bool>( "LOG_MONSTER_ATTACK_MONSTER" ) ? _( "<npcname> avoids %s projectile." ) : "",
                 source->disp_name( true ) );
         } else {
             add_msg_player_or_npc(
                 m_warning,
                 _( "You avoid an incoming projectile!" ),
-                _( "<npcname> avoids an incoming projectile." ) );
+                get_option<bool>( "LOG_MONSTER_ATTACK_MONSTER" ) ? _( "<npcname> avoids an incoming projectile." ) :
+                "" );
         }
         return;
     }
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -10896,19 +10896,22 @@ point game::place_player( const tripoint &dest_loc, bool quick )
                         and_the_rest += counts[i];
                     }
                 }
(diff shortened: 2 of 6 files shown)


new02DAY:
Would it be possible to add "DESCRIBE_MUSIC_FREQUENCY" from game_balance.json? It would be nice to not have to copy/edit the option after every update.

Night-Pryanik:
Technically it's possible, yes, but it's not likely to be approved by project leader.

I can recommend you to create a very small mod only with this option so you don't have to overwrite it with every update.

new02DAY:
Can I ask why it wouldn't be likely approved? I think it'd be a good fit for this PR, since it deals with the music flavor text "spam" from mp3/smartphones/etc. in the log. I'll cool with using a mod, but I wouldn't have discovered the option, if I hadn't been the one to ask for the ability, so I think it would be great for new players to have it in the game menu.

Also, thank you for adding it (just realized you were the author of that PR)! I never thought anyone would go to the trouble of doing it.

Night-Pryanik:
> Can I ask why it wouldn't be likely approved?

Current project management's policy is to avoid adding new options whenever possible (and I don't know the why of it). If there's a necessity to add a new option, developers are encouraged to add an external option instead. So, I'm not even sure that this PR will be accepted as a game option rather than as an external option.

> Also, thank you for adding it (just realized you were the author of that PR)!

You are welcome!

Maleclypse:
For me because this is an option that the PC can turn on and off mid play without breaking anything I think it's a good addition.  My understanding of option removal is that it's intended mostly in places where the player can break their game in ways that may not be immediately clear to the player.  Based on that interpretation I've merged this.

PR #67499, opened 2023-08-07
decrease frequency of tileset updates

casswedson:
#### Summary
None

#### Purpose of change
tileset development activity is sorta low at the moment, I'd be making stuff up if I started presenting numbers, but I think reducing the frequency of the update PRs would be OK

if anyone wants weekly updates, they are available at the upstream repositories

this now runs once a month

### DIFF
--- a/.github/workflows/update-tilesets.yml
+++ b/.github/workflows/update-tilesets.yml
@@ -1,9 +1,10 @@
-# This workflow runs at 23:00 UTC sundays
+# This workflow runs at midnight on the 1st day of every month
 name: "Pull Tileset Updates"
 
 on:
   schedule:
-    - cron: "0 23 * * sun"
+    - cron: "0 0 1 * *"
+  workflow_dispatch:
 
 env:
   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


Maleclypse:
I'm ok with merging this if frequent updatess causing issues but if the only issue is one PR a week I'm not sure that's an issue?

casswedson:
let's see, this is not about fixing a problem, it's more about tweaking the balance between sprites added and the regularity of the updates -- currently the balance is towards regular updates, not sprites added; _now_ the updates consist of a few sprites and lots of script generated json modifications (the script really likes to move json around)

this akin to making lots of small PRs instead of a big one, my motivation is to make the update PRs less bare

alp7292:
i dont think its an issue and affects anything

Maleclypse:
On discord this morning
andrei8l — Today at 2:19 AM
it'd make more sense to import the tilesets as a build step, just like CC-Sounds, instead of pulling the very noisy and bulky patches into the main repo

We are far outside of my realm of expertise here

casswedson:
CC sounds has the particular position of not partaking in the repo's data beforehand at all and being _massive_

adapting the model of cc sounds means removing all tilesets that are being worked on, cause why keep them if they will be outdated, leaving only tynitile in the repo

downloading tilesets (from upstream or a packed release) because the repo no longer contains the main ones and setting git ignores for the ones I use is not convenient, updating occasionally was the working method for a long time, I automated the updates to help fris, I want to tweak the rate of updates, again the updates come along every week and are small content-wise

Rimplestilt:
I really think this was objectively a bad PR to add into experimental. It makes no sense, unless the action of merging sprites is a lot of work.

For example new sprites were merged December 1st, and  a bunch of new sprites were uploaded less than a day after. But we have to wait an entire month for those  to appear in a daily updated, open source experimental game?

Maybe sprite development is slow enough to warrant it, but personally I miss seeing the weekly ruleset updates, and the monthly thing has yet to feel any better.

PR #67501, opened 2023-08-07
Flesh out the Free Merchant representative for the God's Chosen

MNG-cataclysm:
#### Summary
Content "Flesh out the Free Merchant representative for the God's Chosen"

#### Purpose of change
Add some new content to the God's Chosen and expand their inter-faction relations with the Free Merchants.

#### Describe the solution
Adds two new NPCs that will appear as representatives for the Free Merchants when they've met the New England Community Church. These two are unsavory individuals and are looking to make some money by price gouging the NECC on their goods and keeping it secret from the Merchants. The player can get involved in this whole scheme, either by demanding payments and keeping quiet or reporting the fraud to the teamster as part of a quest, more of an espionage type thing.

Meeting the merchant and inquiring about their high prices allows the player to talk to the teamster about it and can take the optional job of investigating the matter by collecting some evidence (accounting records) and bringing them back. The guard for the representative is in on their plot to price gouge and handles all the money in the scheme. However, they are willing to sell out their partner and run for the hills with the loot. The player can get the needed evidence in a few ways through them:
I. Pose as a errand boy and ask for them. You'll be told to come back in two days, which gets you a fixed version of the papers.
II. Threaten the partner into giving them up. You'll need to be ruthless, because they might attack you if you fail.
III. Bribe them for the papers, which involves negotiating a price. Skilled negotiators can get as low as $20, but the price can be as high as $150. Offering more up front makes it easier. The player can come back later to get the dirty versions if they can't pay then and there.
Once the papers are obtained, take them back to the teamster. Showing the dirty papers completes the quest with a reward and replaces the duo with an honest pair of traders. Showing the fixed version fails the job and allows the scheme to continue, so it's more expensive to buy stuff from them.

Not doing this at all, or making the two a different kind of people.

#### Testing
Tested numerous times, everything works just like I want it to.

#### Additional context
I plan on having the Gods Chosen members talk about the pair and the various outcomes, but that's for a future PR.


Fris0uman:
This is a bit dumb but why call them god's chosen in the ids if they're actually called NECC everywhere else in game?

MNG-cataclysm:
> This is a bit dumb but why call them god's chosen in the ids if they're actually called NECC everywhere else in game?

God's Chosen is shorter. Primarily, I can't remember what I called them in-game vs what I remember them as. I'll try to start calling them the NECC more often.

PR #67503, opened 2023-08-07
[MoM] Mind Over Matter-specific Research facility overhaul

Standing-Storm:
#### Summary
Mods "Mind Over Matter-specific Research facility overhaul"

#### Purpose of change

The Research Facility exists mostly to be a building on top of an old-style lab, but I want to revamp it as a core Mind Over Matter location.

#### Describe the solution

Prevent the Research Facility from spawning when Mind Over Matter is enabled, but copy the maps, change the IDs, and re-enable spawning of the new Project PHAVIAN lab (still called "Research Facility"). This also prevents the old-style lab from spawning under it, which will allow me to add a basement in the future. Remove mutagen loot from the lab and implement psionics research loot. Add the alien flora, water, and crystal mapgen as possible specials that spawn in the Project PHAVIAN lab. 

Future PRs will add more rooms, themed offices, special snippets, put a basement underneath, further tweak enemy lists, etc. 

Rarity is the same as the old Research Facility. 

Continually trying to smash the existing Research Facility into being what I want it to be.

#### Testing

Project PHAVIAN lab spawns in with appropriate specials. Old Research Facility does not spawn. 

#### Additional context

The nether map specials can spawn partially in the air, but I actually like this as their existence is supposed to be the result of dimensional nonsense having occurred around the lab. Having chunks of black stone and glowing crystal hovering unsupported is fine.


Maleclypse:
Does this prevent the catalytic cracker from spawning in games with MoM loaded?

Standing-Storm:
> Does this prevent the catalytic cracker from spawning in games with MoM loaded?

I left that room as-is, so it does not.

PR #67508, opened 2023-08-07
Solar panels rebalance

### LINKED ISSUE #67498: Solar panel technology has evolved a lot since solar panels were first implemented into the game and it's in need of an update.

Solar panel technology has evolved a lot since solar panels were first implemented into the game and I think it's time for an update, especially since the recent increased power cost to use many electric appliances. See e.g. #67433. 

### Solution you would like.

Modern solar panels for home use are rated at up to 400W, which is 8 times that of regular solar panels and 4 times that of advanced solar panels in game. Advanced solar panels in the game is described as monocrystalline panels, but that's just the standard now as opposed to older polycrystalline solar panels.

As there's an increased power cost to use many electric appliances now I believe this is a welcome change.

300 W solar panels are about 36 inches by 65 inches= 2340 in^2 / 1.5 m^2, weigh 10 lbs / 4.5 kg (aluminum frame) and 1.5 inches thick, volume = 65 * 36* 1.5 = 3510 in^3 / 57.5 liters. If tilted at a recommended 40 degree angle the ground cross section would be sin 40 * 65 * 36 = 41.7 x 36 = 1504 in^2 or 0.97 m^2.

https://www.ecowatch.com/solar/300-watt-solar-panels

400 W solar panels are about 45 inches by 75 inches = 3375 in^2 / 2.1 m^2 and weigh 45 lbs / 20 kg (steel frame) and 1.5 inches thick, volume = 75 * 40 * 1.5 = 4500 in^3 / 81 liters. If tilted at a recommended 40 degree angle the ground cross section would be sin 40 * 75* 40 = 48.2 x 45= 2169 in^2 or 1.4 m^2.

https://www.ecowatch.com/solar/400-watt-solar-panels

I propose we increase the output of solar panels by a factor of 3 or 4 across the board.

Table of proposed changes:

| Quantity | Current (50W/100W) | Modern 300 W | Modern 400 W |
|---|---|---|---|
| Weight | 14 kg |  4.5 kg | 20 kg |
| Volume |3 L | 57.5 L | 81 L |
| Longest side | 16 cm | 65 cm | 75 cm |
| Ground cross section|  N/A | 0.97 m^2 | 1.4 m^2 |

I also propose we update the name of advanced solar panel to "modern solar panel".

Another option is to make the regular solar cells monocrystalline cells and the advanced solar cells perovskite-on-silicon tandem cells, an emerging cutting-edge new solar technology which is expected to be available for market in 2025. Those can capture 30% of solar energy instead of the 20% of modern monocrystalline cells. This would fit the loot categories better, I think.

* https://en.wikipedia.org/wiki/Perovskite_solar_cell
* https://en.wikipedia.org/wiki/Solar-cell_efficiency#/media/File:NREL_PV_Cell_Record_Efficiency_Chart.png
* https://cosmosmagazine.com/technology/materials/perovskite-set-to-transform-solar-generated-electricity/

### Additional context

If it's just json-edits I could make the PR, I just thought I'd get some input/review first.

Alm999:
#### Summary
Balance "Solar panels rebalance"

#### Purpose of change

With recent changes to electric energy consumption current solar panels have become inadequate. They are too heavy for their output.

resolves #67498

#### Describe the solution

I've adjusted the panels based on what others have find ( https://github.com/CleverRaven/Cataclysm-DDA/issues/67498 ) and my own findings.

The changes are:
1) basic "solar cell" has an output of 25W, is made primarily of silicon (density of 2.33 kg/l) and has a mass of 1.1 kg and a volume of 470 ml;
2) basic "solar array" is made of 8 solar cells, has electric output of 200 W and can be mounted onto vehicle (mass and volume have been adjusted based on the "solar cell" data);
3) a "stationary solar array" consists of 16 "solar cells", has output of 400 W and can **not** be mounted onto vehicle rooftop;
4) all derived types have been adjusted proportionally, with the exception of "foldable" ones. These **shall** have the same performance as the non-foldable ones because they are **literally** made of them, have the same amount of cells and some added mass (extra hinges, locks, frames and wires).

As was noted, we do not currently have IRL analogues to "advanced" types commercially available. This requires further investigation and adjustments.

#### Testing

The items are in the game. I still can not deploy "stationary solar array" from a construction menu, even if the "disconnected stationary solar array" (E)xamine window says I can place it as an appliance. Needs further work.

UPDATE: Construction recipes added to the (*) menu

#### Additional context

TODO:
1) review the crafting recipes: maybe we need a proper steel frame for "stationary solar panel"; -- DONE: steel frame is too heavy (20 kg), aluminum one shall do;
2) define max. length and form or the panels. Assume a cell is square? If so, a 4x2 for a basic solar panel and 8x2 for a stationary one? -- DONE: settled for a 0.44 x 0.3 m rectangle and 3.5 mm thickness for a solar cell; panel shapes are derived from this;
3) make a bogus "f_stationary_solar_panel" furniture item that could replace current fake solar panels atop of some stores / houses (not all of them, though. I think that lone solar panel on an evac. shelter's roof shall stay :) ) -- DONE;
4) fix tests. -- DONE.


data/json/items/vehicle/solar.json
```diff
@@ -4,25 +4,57 @@
...
     "category": "veh_parts",
     "price": 90000,
     "price_postapoc": 1000,
     "melee_damage": { "bash": 1 }
   },
+  {
+    "type": "GENERIC",
+    "id": "stationary_solar_panel",
+    "name": { "str": "disconnected stationary solar panel" },
+    "description": "An electronic device that can convert solar radiation into electric power. It is mot mountable on a vehicle and is intended to be mounted on a rooftop as an appliance.",
```
GuardianDll, line 23:
```suggestion
    "description": "An electronic device that can convert solar radiation into electric power.  It is mot mountable on a vehicle and is intended to be mounted on a rooftop as an appliance.",
```

data/json/items/vehicle/solar.json
```diff
@@ -4,25 +4,57 @@
...
+    "flags": [ "NO_REPAIR", "WATER_BREAK" ],
+    "price": 180000,
+    "price_postapoc": 2000,
+    "melee_damage": { "bash": 1 }
+  },
+  {
+    "type": "GENERIC",
+    "id": "stationary_solar_panel_v2",
+    "name": { "str": "disconnected stationary advanced solar panel" },
+    "description": "An electronic device that can convert solar radiation into electric power. It is mot mountable on a vehicle and is intended to be mounted on a rooftop as an appliance. This model uses high-performance monocrystalline silicon cells.",
```
GuardianDll, line 39:
```suggestion
    "description": "An electronic device that can convert solar radiation into electric power.  It is mot mountable on a vehicle and is intended to be mounted on a rooftop as an appliance.  This model uses high-performance monocrystalline silicon cells.",
```

data/json/items/vehicle/solar.json
Alm999: Additional white-space or new paragraph?

Calavera20012:
Hope a rework for wind power comes at some point too (maybe they are accurate just not very efficient? idk) to put it on par with this. Awesome work!

ampersand55:
> Hope a rework for wind power comes at some point too (maybe they are accurate just not very efficient? idk) to put it on par with this. Awesome work!

I looked into it and you are right. See #67533.

GuardianDll:
any reason it's still draft?

Alm999:
> any reason it's still draft?

I want others to speak out and possibly verify my changes. Maybe there are some stupid oversights. Other than that, I can convert if everything is OK.

snipercup:
> make a bogus "f_stationary_solar_panel" furniture item that could replace current fake solar panels atop of some stores / houses (not all of them, though. I think that lone solar panel on an evac. shelter's roof shall stay :) )

What do you mean by this? Shouldn't the solar panels that are mounted on top of roofs return the new ```stationary_solar_panel``` instead of the current ```solar_panel```?

Alm999:
I've made some final tests (I knew it! There was a small bug with a missing name!). I think this PR is ready.

kevingranade (changes requested):
You need to show your work that justifies quadrupling power output of the existing solar panel part and demonstrate that your changes result in correct power production given new England area latitude and flat mounting. I'm very skeptical that the status quo is that far off.

Your statements about advancements in panel technology and changes to appliance power draw are irrelevant, I just need to see your justification for how much power these would produce after your changes.

Zireael07:
Real life solar panels apparently have a wattage of between 250 W and 400 W, so both numbers check out.
https://news.energysage.com/what-is-the-power-output-of-a-solar-panel/

(Real life panels have 60 cells or 72 cells apparently)

Solar Atlas info for New England: https://globalsolaratlas.info/map?c=44.205708,-70.753784,5&s=44.205708,-70.753784&m=site

Here come lots of deets: https://globalsolaratlas.info/detail?c=44.205708,-70.753784,5&s=44.205708,-70.753784&m=site&pv=small,180,39,1

I can't quite figure out how to apply this to the game since Solar Atlas gives amounts per day or year.

PVOUT seems to be around 3.5-3.9 kWh/kWp

ampersand55:
> You need to show your work that justifies quadrupling power output of the existing solar panel part and demonstrate that your changes result in correct power production given new England area latitude and flat mounting. I'm very skeptical that the status quo is that far off.
> 
> Your statements about advancements in panel technology and changes to appliance power draw are irrelevant, I just need to see your justification for how much power these would produce after your changes.

Here is an attempt at a justification:

> There are a couple of ways to measure solar potential. One standard measures how much electricity a square meter solar panel will produce in a day if installed pointing directly up. By this metric, the cloudy New England states [are near the bottom](https://www.nrel.gov/docs/fy16osti/65298.pdf), according to the National Renewable Energy Laboratory. Most of New England sits between 4 kWh and 4.5 kWh per square meter per day. At the other end of the spectrum, the same square meter solar panel in New Mexico and Arizona will likely produce closer to 7 kWh.

https://www.cnet.com/home/energy-and-utilities/new-england-solar-panels/

The solar panels with 2x8 cells are close to 1 square meter (4 x 2 x 0.44 m x 0.3 m=1.056 m^2), so they should produce 4.25 kWh per day (4250 Wh /  24 h = 177.1 W average). 

Note that solar panels:

1. Only produce in daylight 
2. Have a lower production when not sunny weather 
3. installed (non-vehicle) solar panels are likely angled towards the sun for better efficiency. 

I did a quick test in game where I let a single solar panel run for one in game week in spring (with portal storms disabled) and it produced 40120 battery charge (kJ), which is 11.144 kWh, or 1592 kWh per day or 66.31 W average, which is only 37.44 % of the value reported by the National Renewable Energy Laboratory for New England. This would mean that the in-game solar panels should be closer to 534 W to match real-life values.

kevingranade:
> Real life solar panels apparently have a wattage of between 250 W and 400 W, so both numbers check out.

It very much does not check out,  we're looking at output per square meter, not per "panel" which is variable in size. The 250W panels from your sources are close to one square meter,  and the 400W panels are about 2 square meters, which means the current values are correct, not the ones from this PR.

> Most of New England sits between 4 kWh and 4.5 kWh per square meter per day.

Please point to where in the research paper this is substantiated. The closest I can find is a chart included from another paper showing *tilted* panel production per square meter, which is a completely different thing.

ampersand55:
> Please point to where in the research paper this is substantiated. The closest I can find is a chart included from another paper showing _tilted_ panel production per square meter, which is a completely different thing.

The National Renewable Energy Laboratory has a calculator for what effect to expect:

https://pvwatts.nrel.gov/pvwatts.php

Optimistic:

* Location: Boston
* DC System Size (kW): 0.2
* Module type: Premium
* Array Type: Fixed (roof mount)
* System Losses (%): 0
* Tilt (deg): 0
* Total: = 261 kWh / y = 29.79 W average.

Pessimistic:

* Location: New Hampshire
* DC System Size (kW): 0.2
* Module type: Standard
* Array Type: Fixed (roof mount)
* System Losses (%): 14.08
* Tilt (deg): 0
* Total: = 215 kWh / y = 24.54 W average.

With a 30 degree tilt you can add 19% output.


In my in game test a 200 W panel produced an average of 66.31 W, The discrepancy can partly be explained with 200 W panels having a smaller area than 1 square meter and that calculate use AC/DC conversion and inversion losses which wouldn't be applicable in a vehicle. 

In my research in #67498, the most common panels being sold were 400W and 1.4 square meter, with the 300W versions being 0.97 m^2. Most of the 400 W were intended for house roofs and most 300 W were for RV use with light-weight aluminum frames.

kevingranade:
Your testing indicates that 200W is at least 2x as high as it should be, and I would argue that you're actually looking at something like a 140W panel on one tile of a car, meaning the right epower value to set is right around 50W.

The same applies to the fixed panels, they can be a full square meter and tilted, but you're still looking at an in-game setting around 150W or so to get the year round powet generation to be in the right neighborhood.

ampersand55:
In my testing, a 200 W panel produced 66.31 W average, about 1/3rd of its rating. Assuming it's linear, it would mean that the current 50 W solar panels produce 16.66 W average over a day which is quite a bit lower than the 25-30 W average of the predicted by the site.

The site assumes a 1.2 DC to AC Size Ratio and 96% inverter efficiency, so that would be a direct 25 % increase as we're not converting to AC for vehicles. 

If we allow the panels to be 1 square meter then the 300 W versions would be more applicable for modern panels. Then, if we're generous and go by the optimist scenario, the average output would be 30 * 1.25 * 1.5 = 56.25 W avg power, which isn't that far from the 66.31 W measured in game. 

If we allow solar panels installed as appliances to use DC and to have a 30 degree tilt then it pretty much exactly matches the measured value.

Alm999:
@kevingranade
> Your testing indicates that 200W is at least 2x as high as it should be, and I would argue that you're actually looking at something like a 140W panel on one tile of a car, meaning the right epower value to set is right around 50W.

I don't think it works that way. "epower" is *maximum* power output at full solar exposure, not the average power produced by the "tile".
A common solar panel has 1.056 m2 of area. According to https://en.wikipedia.org/wiki/Direct_insolation the simplified formula for " [Direct normal irradiance](https://en.wikipedia.org/wiki/Direct_insolation) " is

I = 1.353 * 0.7 ^ (AM^0.678) ,
AM = 1 / cos LAT
Boston is at LAT = 42N, so AM = 1.346.
I = 0.875 kW/m2
For a solar panel area the input is 0.875*1.056=0.92 kW.
If we take the energy conversion efficiency of 0.21 then we get the 0.92*0.21=0.193 kW, or 193 W of "maximum" output.

Alm999:
@kevingranade 
Care to show your calculations confirming current solar panels are adequate?

So far I've not seen any explanations the 50W output from a 14kg panel is justified… Only your word.

pwalsh0:
The figures are really obviously too low, by a factor of x3 at least. 

We all have these frickin things on our roofs now and know how much they produce.

The in-game ones currently produce 50W - which would correspond to a miserable 0.3m2 panel for little babies.

Alm999:
Kevin doesn't see a problem. Apparently, if he thinks current solar panels are adequate, that's it. No way of arguing.

GuardianDll:
Isn't #68603 is essentially what Kevin told about?

Alm999:
As far as I can understand, the https://github.com/CleverRaven/Cataclysm-DDA/pull/67508 introduces "arrays" consisting of 4 ordinary panels.
There seems to be a mismatch between item and appliance naming:
solar panel (item) -> "small solar panel" (appliance)
solar panel array (item) -> "solar panel" (appliance)

Other than that, it seems to be just what was proposed in this PR (roughly). Oh, and foldable solar panels are still trash for no reason.

This was probably merged because Kevin haven't seen it (he is not in the reviewers).

PR #67516, opened 2023-08-07
Show inventory selector for tinder when starting fires with tools that require tinder

mqrause:
#### Summary
Features "Show inventory selector for tinder when starting fires with tools that require tinder"

#### Purpose of change

It's annoying to have to put tinder on a fireplace manually.

#### Describe the solution

Show inventory selector to select tinder from when available and no tinder is on the fireplace yet.

Automatically pick one from a crafting inventory for even less key presses, but there's no easy way to prefer one item over the other.

#### Testing

Examined a fire ring with a firewood zone, a loaded pump fire drill, and some tinder, straw, and withered plants nearby. Selecting one moves one charge/item to the fire ring and starts the fire.

### DIFF
--- a/src/activity_handlers.cpp
+++ b/src/activity_handlers.cpp
@@ -1834,9 +1834,32 @@ void activity_handlers::start_fire_do_turn( player_activity *act, Character *you
     item &firestarter = *act->targets.front();
     if( firestarter.has_flag( flag_REQUIRES_TINDER ) ) {
         if( !here.tinder_at( where ) ) {
-            you->add_msg_if_player( m_info, _( "This item requires tinder to light." ) );
-            you->cancel_activity();
-            return;
+            inventory_filter_preset preset( []( const item_location & loc ) {
+                return loc->has_flag( flag_TINDER );
+            } );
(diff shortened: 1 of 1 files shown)


Nebnis:
>Automatically pick one from a crafting inventory for even less key presses, but there's no easy way to prefer one item over the other.

No way to do this even in a firewood zone? Could automatically select  a tinder there to start the fire but then never consume it to actually power the fire. Or...maybe instead of just reloading the drills with just "notched wood", it could be "notched wood with tinder"

Either way thank you for this, i also found really annoying to manually add tinder.

mqrause:
There are many possible ways to do this, and this is definitely not the best way. Doing it properly with drills would require support for multimags, but it'd probably be the best solution. Doing it with the firewood zone runs into issues with using tinder as firewood and again (lack of) preference of one kind of tinder over another, even if most people would probably just put one kind in there.

Nebnis:
i didnt mean multi mag but a whole new item that is called "notched wood with tinder" Notched wood has no other use so you might as well include the tinder right? And both comes from same source to boot since you would be getting some tinder from notching wood

PR #67521, opened 2023-08-08
[MoM] Add VERDANT HAND Combat Protocol

Standing-Storm:
#### Summary
Mods "Add VERDANT HAND Combat Protocol"

#### Purpose of change

Biokinetic ferals can kick the crap out of you with nothing but their bare hands.  It only makes sense that PC biokinetics can do the same.  Plus, finding neat loot is fun. 

#### Describe the solution

Adds a recipe that allows players to gain the hardcoded Ki Strike mutation (renamed "VERDANT HAND Combat Protocols" for the purposes of Mind Over Matter). At the moment, since it's a constant effect the requirements are very high--high proficiency in several biokinetic powers and passing two checks while performing the recipe. Should the Ki Strike mutation ever become an effect, these requirements will be relaxed (and the VERDANT HAND Combat Protocols will be re-implemented as a power you learn from studying the recipe instead). 

#### Testing

The book exists, shows up in loot tables, and the recipe is learnable. It is possible to succeed at it and gain the VERDANT HAND Combat Protocols.

### DIFF
--- a/data/mods/MindOverMatter/effects/effects_penalty.json
+++ b/data/mods/MindOverMatter/effects/effects_penalty.json
@@ -288,7 +288,7 @@
     "limb_score_mods": [ { "limb_score": "breathing", "modifier": 0.8 }, { "limb_score": "manip", "modifier": 0.8 } ],
     "flags": [ "EFFECT_LIMB_SCORE_MOD" ]
   },
-   {
+  {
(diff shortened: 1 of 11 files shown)


Standing-Storm:
Commits got messed up somehow.  Will retry.

PR #67526, opened 2023-08-08
JSONify Defense Mode

MNG-cataclysm:
#### Summary
Content "JSONify Defense Mode"

#### Purpose of change
Defense Mode is the neglected child of this game and hasn't been worked on in years. The barrier to entry on this mode is that it's entirely programmed in C++, which only a small part of the contributor base knows and understands. As such, I aim to fix this issue and lower the barrier for programming entry on Defense Mode, and later add on to and expand it and allow for mod compatibility, even for third-party, out-of-repo systems.

#### Describe the solution
Remove most of the C++ that can be replicated in JSON and make Defense Mode a mod. This lets player mix-and-match what content they'd want to see in their game and can even allow for modded enemies to appear. 

8/8/23: Currently, the basic infrastructure of map generation, buildings, prebuilt game settings, scenarios, and enemy type selection are ready to roll, but the EOCs that control wave spawning, and caravans still needs to be put together.

8/11/23: The EOCs that control everything is ready to go, including wave spawning and caravans. I still need to create EOCs that allow more random NPCs to appear and add some way to recruit mercenaries so the player can truly build a base as they want.
The old C++ has been removed, and everything appears to be cleaned and spruced up.

Not doing this or removing Defense Mode entirely without replacing it.

#### Testing
Loaded in the game multiple times, everything works well.

#### Additional context
8/8/23: This will require some C++ support in the form of new functions, and implementing the current code that governs monsters moving to your location in a more generalized way, such as in the form that missions assign targets. This could have a lot of other uses if used that way, such as allowing NPCs to spawn in and track the player's current position and home in on it, or with the horde mechanics.

8/11/23: From my testing with combinations of `ALL_SEEING` and `NEMESIS` flags for monsters, it seems that they can reliably tack the player over great distances. Some more playtesting is needed to fully confirm that it works without problems, but I think that it'll go well.


ZhilkinSerg:
Would you please use mod id without whitespace - like `defense_mode` or something?

MNG-cataclysm:
> Would you please use mod ~~id~~folder name without whitespace - like `defense_mode` or something?

Sure can do.

MNG-cataclysm:
I think this PR is getting a bit too large for anything more, so I think I'll mark it down as ready, and add the rest of the code in a different one.

Maleclypse:
That was super fun! I'm looking forward to seeing what more you do with it!

PR #67535, opened 2023-08-08
Make waiting in npc heavy areas up to 2x faster

akrieger:
#### Summary
Performance "Optimize eoc processing and other fixes to speed up waiting near many npcs"

#### Purpose of change
Waiting in npc heavy areas is slow. Very slow. Excruciatingly slow. The hot spots that stood out could be optimized to make things better. Better things are better.

#### Describe the solution
- `npc::is_dead` took up 25% of all cpu time just recalculating if an npc was dead every turn for ai planning purposes. This 'dead state' be cached and invalidated only when the relevant conditions to kill a character occur. This cost is essentially eliminated.
- `effect_on_conditions::process_effect_on_conditions` took up ~46% of all cpu time. Approximately 3% was an unnecessary `string_format` call just parsing for a single string printf formatter which could be replaced with a direct string concatenate. Of the remaining 43%, ~18.5% was effectively allocator overhead and eliminated through a variety of caching and lazy initialization techniques.
  - `process_eocs` was allocating (and reallocating) and deallocating a heavyweight vector every call. Caching this as a function static amortizes the cost to zero. This memory management of a transient vector cost 2.75% of cpu time.
  - A certain `dialogue::dialogue` constructor was creating temporary empty maps as arguments, default initializing member temporary empty maps, then copy-assigning from the temporary arguments to the member variables, resulting in net two wasted map allocations out of three. Likewise a map was being copied instead of moved-from during eoc queue processing. Optimizing the above to avoid the unnecessary allocations saved approximately .75% of cpu time.
  - The eoc queue structure was a `std::priority_queue` which uses a `std::vector` as storage. A contiguous storage data structure is essentially required by `std::priority_queue` (technically, it requires a random access data structure). `priority_queue` operations result in moving and swapping values around, which requires constructing a temporary value to swap into, resulting in more wasted allocations from temporary maps, every time a value is pushed and popped. Reimplementing the queue as a pair of a `std::list<queued_eoc>` and a `std::priority_queue<std::list<queued_eoc>::iterator>` made the expensive queue operations allocation-free, as the iterators are trivial. The list is only pushed and popped when an eoc actually is processed and dequeued, and if it is re-scheduled then the timestamp is updated in place instead of making (another) copy. This saves 3.5% of cpu time.
  - Despite optimizing how many copies happen during eoc processing, there are temporary `dialogue` objects created during eoc processing as copies of existing dialogues. This seems to be to provide a 'scratch' space for eoc logic to write and read context vars, among other things. Most of the time this resulted in just copying empty maps. Instead, introduce a `lazy<>` wrapper which allows transparently deferring initialization of a type until it is actually needed. Optimize the dialogue constructor to avoid unnecessarily instantiating or copying empty context or conditional maps. Move the callstack debug variable out of the context map and into a string member in dialog, to avoid instantiating a map for a single entry. Doing the above optimized out the remaining overhead, approximately 9.5% of total cpu time.
  
Between both fixes, time spent waiting eg. in the refugee center start was reduced by almost half, ~46.5%. This value is bigger the more npcs are in the bubble.

- :elmofire:

#### Testing
Wait 1 hour at the refugee center start, with a couple wasps in a glass door cage for extra npc ai pain just to help find hot spots.
Quick wait in MSVC debug mode to verify STL debug iterators don't immediately fire exceptions (care was taken to use apis that preserve iterators).

image
image
image
image

#### Additional context
TODO:
- [x] Need to ask implementors if the callstack *must* be in the context eg. when being written out to a debug file, or if its acceptable to stash as a side variable as it is.
- [x] Need to fix tests.

### DIFF
--- a/src/cata_lazy.h
+++ b/src/cata_lazy.h
@@ -1,0 +1,81 @@
+#pragma once
+#ifndef CATA_SRC_CATA_LAZY_H
+#define CATA_SRC_CATA_LAZY_H
+
+#include <optional>
+#include <type_traits>
+
+/**
+ * A wrapper for a type that lazily initializes the underlying type as needed. Use to
+ * save unnecessary allocation costs for eg. std containers in transiently allocated types.
+ *
+ * T must be default initializable.
+ */
+template<typename T, typename = std::enable_if_t<std::is_default_constructible_v<T>>>
+struct lazy {
+    lazy() = default;
+
+    // std::optional-like api for testing without allocating
+
+    bool has_value() const {
+        return actual.has_value();
+    }
+
+    // Transparent value operators that proxy to the actual value.
+
+    template<typename Arg, typename ...Args>
+    // NOLINTNEXTLINE(google-explicit-constructor)
+    lazy( Arg &&arg, Args &&...args ) : actual{std::in_place, std::forward<Arg>( arg ), std::forward<Args>( args )...} {}
+
+    template<typename U>
+    // NOLINTNEXTLINE(misc-unconventional-assign-operator)
+    T &operator=( U &&u ) {
+        get() = std::forward<U>( u );
+        return get();
+    }
+
+    // NOLINTNEXTLINE(google-explicit-constructor)
+    operator T &() {
+        return get();
+    }
+
+    // NOLINTNEXTLINE(google-explicit-constructor)
+    operator T const &() const {
+        return get();
+    }
+
+    T &operator*() {
+        return get();
+    }
+
+    const T &operator*() const {
+        return get();
+    }
+
+    T *operator->() {
+        return &get();
+    }
+
+    const T *operator->() const {
+        return &get();
+    }
+
+
+    template<typename Arg>
+    auto &&operator[]( Arg &&arg ) {
+        return get()[std::forward<Arg>( arg )];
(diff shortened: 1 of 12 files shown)


akrieger:
@Ramza13 you added the dialogue callstack variable in #65510, does it particularly need to be in the context map for any reason (afaict it's always transiently allocated and dropped inside copies of dialogues in eoc processing and never persisted), is it ok to move it to a separate variable? Without further changes it won't get serialized everywhere that serializes the context, not sure what that might break or not...

Zireael07:
Great catch with the 'dead' check :)

akrieger:
Hm. I don't understand why the lto build is segfaulting here, but not the other tests. It doesn't seem anywhere close to anything I touched. I'll dig into it.

akrieger:
Oh never mind it happens everywhere. https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/5813361680/job/15763737746#step:16:1

akrieger:
clang-tidy timed out but would've failed at least one file because of how pervasively transitively cata_lazy.h ends up being included.

akrieger:
Fun side effect is this should make just stepping around twice as fast too.

Issue #67324, opened 2023-07-31
Charge removal side effects 1: tiny items


mqrause:
### Describe the bug

A few items had really tiny charges smaller than 1ml. Because we have no unit smaller than that, all of them were rounded up to 1ml and most likely could use an audit to check their stats.

The list of affected items can be found here: https://gist.github.com/mqrause/dc234737bfd5c4309aa4dc496d17bd53#file-oversized_charge_removal-txt

### Attach save file

n/a

### Steps to reproduce

n/a

### Expected behavior

n/a

### Versions and configuration

6088293934801590f503ff9d5d961e708af564ff

PR #67539, opened 2023-08-09
Recharge sulfur powder and make it not a comestible

anoobindisguise:
#### Summary
bugfixes "sulfur is treated like a chem rather than a comestible"

#### Purpose of change
Sulfur powder was caught by the automated changeover of charge removal from comestibles. This is probably in error because it was not applied to most of the other chem items, but sulfur was a comestible specifically because you could rub it on your skin to cure a fungal infection. It probably should not work that way.

#### Describe the solution
This makes it an ammo and not a comestible. This causes an error as existing sulfurs become sulfur (active) which errors on load, I don't know how to fix this. Future sulfurs will not have the issue though. Sulfur should not be tracking temperature and doing all the other things a comestible does, it's needless overhead on the game. It also obviously removes its status as a high grade fungicide with no health penalty.

I get that we're trying to move away from charges but sulfur frequently spawns in the thousands (and breaking up a single chunk of sulphur makes 1500+ individual sulfurs) so it is a particularly egregious case that can probably be left for de charging after we fix some more of the underlying issues with charges that affect comestibles.

#### Testing
Works in game but has the abovementioned errors.


Maleclypse:
Are those errors skippable or crash?

mqrause:
> This causes an error as existing sulfurs become sulfur (active) which errors on load, I don't know how to fix this. Future sulfurs will not have the issue though.

You probably just need to add it to `obsoletion/blacklist_temperature_removal.json`.

Also considering sulfur is one of the cases from #67324, is it possible to rebalance the charge amount and size in a sensible way to avoid the same issue once ammo gets their charges removed?

anoobindisguise:
> > This causes an error as existing sulfurs become sulfur (active) which errors on load, I don't know how to fix this. Future sulfurs will not have the issue though.
> 
> You probably just need to add it to `obsoletion/blacklist_temperature_removal.json`.
> 
> Also considering sulfur is one of the cases from #67324, is it possible to rebalance the charge amount and size in a sensible way to avoid the same issue once ammo gets their charges removed?

Chemicals are measured as 1/100th mol per charge so unless that paradigm changes probably not.

mqrause:
> Chemicals are measured as 1/100th mol per charge so unless that paradigm changes probably not.

Huh, I didn't know that. But that'll probably mean a lot of the chemicals will break when ammo charges get removed, so we need to figure something out for that at some point.

PatrikLundell:
There is probably a need for a system for powders and liquids that tracks how large one lot of it is, regardless of whether you call it charges or anything else (and powders should probably require containers or be spilled. Spills should probably be partially recoverable into containers at some point).

PR #67542, opened 2023-08-09
[MoM] Allow Premonition to detect dangerous world events

Standing-Storm:
#### Summary
Mods "Allow Premonition to detect dangerous world events"

#### Purpose of change

Clairsentients should be able to tell if something is gunning for them.

#### Describe the solution

Copy and edit the hostile-spawning EoCs and insert a check in between to see if the player has the Premonition power.  If they don't, run the EoC as normal.  If they do, delay the EoC by a short time period, scaling higher with levels in Premonition

#### Testing

Works perfectly with Shadow.  Need to implement it with other EoCs.

#### Additional context

I don't know everything that should go here--the assassins sent by factions, maybe a warning when you descend into the bottom of the Amigara mine, but I'm sure there are things I'm missing.


Cupidatis:
Portal storm detection?

Standing-Storm:
> Portal storm detection?

I've been going back and forth on that because the early portal storm weather is already a warning system.  But you're right, in-world it does make sense that a clairsentient's powers would warn them a portal storm was about to break on them

Cupidatis:
Maybe some kind of modificator of radius at area of early portal storm and some message for this. Don't what be better. 
Sorry for my bad english

MNG-cataclysm:
I think that the assassin EOCs should be included under Premonition, possibly with a warning message about some mysterious individuals coming.

Standing-Storm:
This works so far--better to merge it and then I can add more than let it sit while I try to to think of everything that could possibly be included.

PR #67556, opened 2023-08-09
Remove Mutation Selector mod

kevingranade:
#### Summary
None

#### Purpose of change
As a feature toggle, this mod does not meet the inclusion criteria for the repository.
https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/IN_REPO_MODS.md

#### Describe the solution
Deletes the mod.
The infrastructure it came with in #67483 are usable for both in and out of repository mods, so I'm not reverting the whole change.


bombasticSlacks:
@Rewryte Pinging you, so it doesn't seem like this was done in secret. As posted above, single option mods like this don't meet repo inclusion.  You are welcome to distribute just this change separately. Sorry, it got merged and then reverted, there should have been more discussion on the PR proper.

I'm also going to include this infra as a perk in the BombasticPerks mod (with a bit of additional C++ extension), I think this is something players would really like, and would introduce additional build direction for. As you say, plenty of people hate the randomness and basically don't play with mutations because of it. Aftershock, I believe, will also be incorporating this.

I'm not sure on the current direction of mutations in vanilla but adding this functionality as a high-end consumable or furniture might be possible too. Potentially even, changing the current mutation stuff to just allow more choice, I'm not really dialed in on the current vanilla plans, though, so I can't say how that suggestion would be received.

Rewryte:
I kind of gathered what was going on, but I appreciate the gesture.  I could have looked at the docs more closely so this one's on me.  Glad you and the other authors found the code helpful.

PR #50361, opened 2021-07-31
Some of the hard-coded talk topics can now be described in NPC's json.

ti360gh:
#### Summary
 Features "Some of the hard-coded talk topics can now be described in NPC's json."

#### Purpose of change
 Some of the talk-topics of NPCs are currently the same for all NPCs in the game.
 To better characterize NPCs, the following talk topics now can be described in json.

- TALK_FRIEND / TALK_RADIO / TALK_LEADER / TALK_STOLE_ITEM / TALK_WAKE_UP / TALK_FRIEND_GURAD
- TALK_MUG / TALK_STRANGER_AGGRESSIVE / TALK_STRANGER_SCARED / TALK_STRANGER_WARY / TALK_STRANGER_FRIENDLY / TALK_STRANGER_NEUTRAL

 This will make NPCs more unique and will help to create a kind of NPC-driven story mods or NPC heroes.

follower#1
image
follower#2
image

#### Describe the solution
 - The above hard-coded topics are now can read from npc's json.
 - The custome topic is handling with npc.chatbin.<same_name_of_topic> in game.
 - Hard-corded topic replaced to chatbin.<same_name_of_topic>.
 - save/load that topic names.

The following topics seems system messages, so I don't touch it.

 - TALK_REBOOT / TALK_SEDATED 
 
The mission dialog also a "common", but this can be avoided by creating a dialog chain. 

#### Describe alternatives you've considered
 Nothing.
 I wanted to replace the hard-coded values with JSON.

#### Testing
- Setup
I created the scenario with 2 NPC instances.
One of NPC have all customized topics in json.
The other one have nothing above fields in json.

- Testing
 All topic was displayed by in-game debugging or save editing.
  - TALK_FRIEND
  Talk to follower.
  Assign NPCs as camp worker/re-join to follower and talk.
  - TALK_FRIEND_GUARD
    Assign NPCs as camp worker and talk.
  - TALK_WAKE_UP
  Set the sleep effect in the In-game debug menu and talk to the NPCs that is sleeping.
  - TALK_RADIO
  Assign NPCs as camp worker with 2-way radios. leave camp and talk them using "#" menu.
  - TALK_LEADER / TALK_STOLEN_ITEM
  Set NPC `attitude` with 5/18 in json, start a game and ~~save/exit~~ talk them.
  - TALK_MUG / TALK_STRANGER_ stuff.
[...]

PR #67561, opened 2023-08-09
Begin giving Luo unique dialogue after recruitment

I-am-Erk:
#### Summary
Content "Give Luo unique dialogue after recruitment"

#### Purpose of change
You can recruit Luo Meizhen and the other beggars to join your camp, but once you do htey're just subpar NPCs

#### Describe the solution
#50361 added the ability to customize a follower's dialogue and slipped right under my radar.
This uses that functionality to give Luo a fully custom follower dialogue. Currently some of it is just copied from the default, and it still uses a few default topics. Adds specific discussion allowing her to remind you of her background as well.
Of note, Luo will teach you without you needing to persuade her.

Could add different dialogue or pick a different NPC.

#### Testing
Loaded up and tested, it all works as intended. Getting to her conversation bits is a little more subtle than I'd like, but I can fix that later.
image
image

#### Additional context
It would be nice if we could use an efffect to change her display name.


anoobindisguise:
Should she say "you yeeted us out of there"? Not "yote"? Or something non yeet-related? 
This is cool btw I love getting to have unique dialogue with my minions.

I-am-Erk:
I'm over 40 years old, I do not know the correct conjugations of the verb "to yeet". However I suspect Luo also does not, because she's a massive nerd.

I-am-Erk:
I might work on this more tonight. It should get tested. However I'm leaving on a trip shortly... On the off chance I vanish, this is probably mostly safe to merge.

PR #67592, opened 2023-08-10
Nerf phaseimmersion drop from xedra merc

anoobindisguise:
#### Summary
Balance "Phase immersion boss drop needs extra work to be usable"

#### Purpose of change
Currently the by far most easy and effective way to get good armor is to kill the xedra merc boss, as it drops a fully functional suit of phase immersion armor which handles night vision, defense, temperature, electricity, radiation, acid, gas, etc... I've also never had it even degrade on me once even after being subjected to months of abuse. Balance wise, it's simply too strong and too easy to get - you just need a good rifle and can kill the boss very quickly, the hardest part is getting there since the hallway can be a little chaotic especially if leechplants invade.
Realism wise, is the nether spitting out a perfect "copy" of the merc's gear in mint condition? Or is the merc dying and dropping the suit, which should realistically be riddled with bulletholes from me killing the merc? And they were killed in some sort of dimensional breach so logically there should be a bit more wear and tear on their gear.

#### Describe the solution
The phase immersion suit now spawns with a random amount of damage and the BROKEN flag. Both of these can be fixed by Hub01 and should create a more interesting "mini quest" - I have this cool suit of hypertech armor, but it's trashed, so how do I fix it? Then you get a big reward for doing so (the armor)

Don't give it the "broken" flag. I'm not sure if a nanofabricator will fix it but I know hub will remove it. Alternatively, make nanofabs hardcoded to remove that flag. It just feels weird to have your suit be mangled but still say `"Initiating...Running suit integrity diagnostics…All systems operational."`

### DIFF
--- a/data/json/npcs/EOC_talkers/xedra_merc.json
+++ b/data/json/npcs/EOC_talkers/xedra_merc.json
@@ -47,9 +47,9 @@
       },
       { "group": "clothing_tactical_leg" },
       { "group": "clothing_watch" },
-      { "item": "phase_immersion_suit", "damage: [ 1, 4 ], "custom-flags": [ "BROKEN" ] },
+      { "item": "phase_immersion_suit", "damage": [ 1, 4 ], "custom-flags": [ "BROKEN" ] },
       { "item": "id_science_security_black" },
-      { "item": "survivor_machete_qt", "variant": "xedra_machete", "container-item": "scabbard", "damage: [ 1, 2 ] },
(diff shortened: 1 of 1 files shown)


John-Candlebury:
If "game balance" will orient the main argument I rather you make the fight more engaging difficult than damage the dropped suit, its not that its a bad idea, but repairing the suit is not a particularly dangerous task.

Take note that this is also serves as the labs "finale" its just been made more engaging that the classic room full of sci-fi loot.

PR #29867, opened 2019-04-23
Add NPC retreat, no investigate, and investigate limit zones

mlangsdorf:
#### Summary
```SUMMARY: Features "Add NPC retreat, no investigate, and investigate limit zones"```

#### Purpose of change
Part of #29333 
Add 3 new zones for friendly NPCs:
* Retreat zones give the NPC a preferred location to retreat to if the zone is within 60 tiles and the NPC is fleeing threats or a fire
* No investigate noises prevent the NPC from investigating sounds coming from inside the zone.
* Investigate limit zones prevent the NPC from investigating sounds coming from outside the zone.

#### Describe the solution
Add two new functions to the zone manager: get_nearest returns the nearest point of a zone type within a specified distance, and has_defined returns true if there is a cached zone type.

In npcmove.cpp, good_escape_dir will attempt to path to the nearest retreat zone in preference to using the threat map.

in npctalk.cpp, handle_sounds will skip investigating sounds that come from inside a no investigate zone, or if a investigate limit zone is defined, from sounds coming from outside of it.

Update the in-dialogue tutorial to include the new features.

PR #67593, opened 2023-08-10
NPCs actually flee - playtesting requested

I-am-Erk:
#### Summary
Features "NPC fleeing behaviour adjusted and slightly improved"

#### Purpose of change
in #29867, @mlangsdorf added some NPC retreat zones and associated retreat behaviour. This was a start meant to be expanded but, well, we haven't done so. NPCs seem to never retreat.

#### Describe the solution
Initially I thought this was because we never use retreat zones, so I set up retreat zones to test, but NPCs still didn't retreat to them. I added some debug messages to determine danger assessment and found that NPCs were much too brave.

I have adjusted the bravery calculation and made NPCs noticeably more cowardly. Initially I also added a fallback behaviour, where NPCs would try to flee to your location. However then I realized that there was an implemented threat map that gave them a significantly smarter flee route, it just wasn't triggering because they were so fatally courageous.

So, I spent a while adjusting the calculations for bravery, making the personality trait "bravery" more important and making threats much higher. I also significantly extended the duration they'll flee, so instead of dancing a few tiles back and charging in, they'll actually run a fairly long way away. Honestly they could stand to run further but I think this works pretty well.

Edit: after some playtesting reports from @Maleclypse I found that regular zombies, even hordes of them, were still not triggering a flee response. Specifically, it seems like threats of large groups are not scaled up enough: one zombie presents a threat of 5, six zombies presented a threat of 13. One migo presents a threat of 27, four migo present a threat of 55. I don't understand the code well enough to see why this is, but I added what may be just a simple hack to quickly count the number of hostiles and the number of friendlies in view, and scale up assessed threat if the friendly team is outnumbered. Now, six zombies present a threat of 50, ten fold what a single zombie does, which feels way more appropriate. This did seem to work: NPCs now run if there are more than a couple zombies coming at them. I would like some testing to see if I've made them too chicken, but I am on the way out on a trip tomorrow. I'd love some comments on the balance, but I think if this was merged as it is, it wouldn't be a total flop.

- more testing to see if the bravery calculations are appropriate. NPCs should be cowardly compared to the player, but not so cowardly as to be useless
- we could consider some more customization. You could tell your NPCs to run sooner, which would decrease their bravery roll, or tell them to run further when they decide to run. You could also have an override to tell them to regroup on your position instead of running away, and then we could implement the behaviour I had to begin with.
- I think we still should have the ability to give a RUN AWAY command that activates `effect_npc_run_away`, and a REGROUP command that causes the NPC to path to  you, regardless of what they're doing. It would also be nice to have a "HOLD THE LINE" command that decreases their chance to flee, but doesn't remove it entirely.

#### Testing
Loaded game. Spawned NPC ally and zomborg. NPC fought zomborg and was winning. Spawned another zomborg. NPC continued to fight. Spawned a third zomborg. NPC ran like a frightened rabbit.

Spawned a skeletal juggernaut. NPC fired until their gun ran out of ammo, then reassessed the threat, realized they stood no chance, and ran away.

Honestly it worked really well. Mark's AI stuff was pretty well done, it just wasn't triggering.

### DIFF
--- a/src/npcmove.cpp
+++ b/src/npcmove.cpp
@@ -425,6 +425,8 @@ void npc::assess_danger()
 {
     float assessment = 0.0f;
     float highest_priority = 1.0f;
+    int hostile_count = 0;
+    int friendly_count = 1; // count yourself as a friendly
     int def_radius = rules.has_flag( ally_rule::follow_close ) ? follow_distance() : 6;
 
     if( !confident_range_cache ) {
@@ -530,6 +532,7 @@ void npc::assess_danger()
         Creature::Attitude att = critter.attitude_to( *this );
         if( att == Attitude::FRIENDLY ) {
             ai_cache.friends.emplace_back( g->shared_from( critter ) );
+            friendly_count += 1;
             continue;
         }
         if( att != Attitude::HOSTILE && ( critter.friendly || !is_enemy() ) ) {
@@ -542,6 +545,7 @@ void npc::assess_danger()
 
         ai_cache.hostile_guys.emplace_back( g->shared_from( critter ) );
         float critter_threat = evaluate_enemy( critter );
+        hostile_count += 1;
         // warn and consider the odds for distant enemies
         int dist = rl_dist( pos(), critter.pos() );
         if( is_enemy() || !critter.friendly ) {
@@ -599,6 +603,10 @@ void npc::assess_danger()
     if( assessment == 0.0 && ai_cache.hostile_guys.empty() ) {
         ai_cache.danger_assessment = assessment;
         return;
+    }
+    // being outnumbered is serious.  Scale up your assessment if you're outnumbered.
+    if ( hostile_count > friendly_count ) {
+        assessment *= ( hostile_count / friendly_count );
     }
     
     const auto handle_hostile = [&]( const Character & foe, float foe_threat,
--- a/src/npcmove.cpp
+++ b/src/npcmove.cpp
@@ -524,7 +524,7 @@ void npc::assess_danger()
         // Unlike allies, hostile npcs should not see invisible players
         ai_cache.hostile_guys.emplace_back( g->shared_from( player_character ) );
     }
-    
+
     for( const monster &critter : g->all_monsters() ) {
         if( !clairvoyant && !here.has_potential_los( pos(), critter.pos() ) ) {
             continue;
@@ -561,7 +561,7 @@ void npc::assess_danger()
         if( !clear_shot_reach( pos(), critter.pos(), false ) ) {
             continue;
         }
-        
+
         float scaled_distance = std::max( 1.0f, dist / critter.speed_rating() );
         float hp_percent = 1.0f - static_cast<float>( critter.get_hp() ) / critter.get_hp_max();
         float critter_danger = std::max( critter_threat * ( hp_percent * 0.5f + 0.5f ),
@@ -599,16 +599,16 @@ void npc::assess_danger()
             ai_cache.danger = critter_danger;
         }
     }
-    
+
     if( assessment == 0.0 && ai_cache.hostile_guys.empty() ) {
         ai_cache.danger_assessment = assessment;
         return;
     }
     // being outnumbered is serious.  Scale up your assessment if you're outnumbered.
-    if ( hostile_count > friendly_count ) {
+    if( hostile_count > friendly_count ) {
         assessment *= ( hostile_count / friendly_count );
     }
-    
+
     const auto handle_hostile = [&]( const Character & foe, float foe_threat,
     const std::string & bogey, const std::string & warning ) {
         int dist = rl_dist( pos(), foe.pos() );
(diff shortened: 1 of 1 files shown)


Maleclypse:
Just compiled and we better hope NPCs don't have souls our there's gonna be about a dozen just waiting for me in hell.  Couldn't get a one to retreat to me or a zone.

PatrikLundell:
My main problem with companion uselessness in combat has been their inability to run after the PC when the PC runs and they're ordered to stay close, as walking companions weren't fast enough to be able to retreat from the enemies. The second issue has been them not following orders, such as not attacking when they should, but just stand there inactive while the PC is in melee.

However, fleeing (at a run) companions might improve their survivability and also improve their illusion of having a mind of their own.

I-am-Erk:
The original pr here had them running to you when they fled, but that concept has been eclipsed by the need to fix fleeing at all. Once it is going, we could look more at getting them to stick with you more

fairyarmadillo:
I believe your tests are skewed because you are looking at early game situations. There is a major issue with how NPCs calculate threats that even now make them completely useless past the first summer. They appear to rate enemies by their difficulty score with no respect for their or the player's gear, skill, or stats.

Over the course of a very long playthrough where I tried to extensively use NPCs, I found that even highly skilled NPCs would freak out and run away from basically any fight once hulk-tier enemies started showing up. These enemies are scary, but a veteran survivor with expertly crafted armor, a gun, and a tempered steel melee weapon they're very good at using is scarier, especially when they have a friend nearby.

Retreating generally spells death for NPCs, as they lack the AI to effectively navigate the many hazards enemies start spitting out lategame without direct orders from the player. I think that making them even more cowardly than they already are will have a massively negative impact on their use lategame, which is unfortunate because that's when you have the resources to train them up and are facing enemies that really require help to bring down.

PatrikLundell:
Sounds like important context.

However, I would claim that the very early game is also a point where companion support is extremely useful, as the player has no skill, no protection, and crappy weapons and thus needs support even against a single basic enemy to at least reduce injuries taken. Thus, balancing ought to allow for companions to be useful both in the very early game and in the late game (as well as the stretch in between those extremes).

I-am-Erk:
NPC's absolutely do rate their and your equipment. A no-go, for example, rates at threat 27. An NPC with a loaded hunting rifle rates itself at threat 400, and correctly decides that it can handle the mi-go.

Feel free to compile and let me know if I'm incorrect.

That said, NPC's running away is generally a more correct behaviour than NPC's fighting to the death always. If that needs tuning the other way, that's ok.

PatrikLundell:
There's a difference between retreating and blindly running in terror. A retreat ought to firstly retreat towards the PC (if the PC is further away from the threat and there's a clear path), and then the way they came after getting further away from the threat than the PC is, unless there are reasons to move in another direction. Regrouping by the car seems like a reasonable strategy when a vehicle is used, although the companion may well stop well before reaching it if out of range to draw in enemies (with some safety margin so random enemy movement won't cause them to start to approach again).

If I had the choice between companions fighting to the death and randomly drawing in more enemies causing even more trouble than you had in the first place I would take the first one. Although it probably wouldn't matter, as I'd probably continue to not use companions in either case, as both extremes are too broken to be useful.

fairyarmadillo:
> NPC's absolutely do rate their and your equipment. A mi-go, for example, rates at threat 27. An NPC with a loaded hunting rifle rates itself at threat 400, and correctly decides that it can handle the mi-go.

"A mi-go" is not a common lategame scenario, that's a day one enemy you can beat to death with a stick. A common lategame scenario is dozens of end-stage zombie evolutions on-screen at the same time. A properly trained and equipped survivor (or two of them) can handle this if they're able to act rationally, but in my experience under the existing system, ie BEFORE you have made NPCs more cowardly, they will still frequently blindly run away, which is almost invariably fatal for them when standing their ground would not have been.

I-am-Erk:
My point is not that NPCs in mid to late game are correctly tuned, but that your assessment just isn't correct. They do weigh their, and your, equipment into their calculations. If you want me to be able to assess whether or not there is a problem, I need actual play test information, ideally with the debug messages for what their threat assessment reads. Then there's a really good chance I can fix it... But when there's a straight factual error in your initial statement, there's no way I can act.

I-am-Erk:
> A retreat ought to firstly retreat towards the PC

Like I said, once I can confirm retreat behaviour is actually working well, I'd be very happy to work on implementing this. The code to do so was very easy.

I-am-Erk:
@fairyarmadillo sorry if that came across a bit too grouchy. I woke up with a splitting headache before a very long drive, it has nothing to do with you.

fairyarmadillo:
> @fairyarmadillo sorry if that came across a bit too grouchy. I woke up with a splitting headache before a very long drive, it has nothing to do with you.

I understand the frustration over what might have seemed like an off the cuff complaint. I do not have empirical data that uses your PR as a basis, I only have anecdotal data about how the NPC fleeing mechanic worked as of about a year ago. I'm aware that is less useful, but it can hopefully shed some light on a point where the system was not doing so well, and could use improvement.

I-am-Erk:
this has been waiting two weeks and I've not been able to recruit any further testers. I think we may need to merge it and test through the exp branch

PR #67595, opened 2023-08-11
Only transparent pockets allow light to pass through them

### LINKED ISSUE #16461: Changes in source-of-light system
The main idea is: any source of ligh(SoL) don't emit light when it in inventory. If you want light – equip SoL in hand like swag bag (if it don't wearable already, like head torch) or like weapon if it neccesary in current game situation. Exeptions(optional): glowsticks and lightstrip – these two in my opinion should be worn like holy symbol. Additionally, you can forbid activation in inventory and putting in inventory activated SoL that using fire – for natural roleplay causes.
Pros:
- More realistic.
- It will add roleplay point in head torches and mining helmets: savings two units of volume in pockets is a bad purpose for the thing which is designed to free up your hands.
- It is possible to hide SoL that can't be turned off "in pockets" therefor able to safe carry it in night city.
Cons:
- A little more in-game hassle than it was (almost every "mechanic change" update do so).

mqrause:
#### Summary
Features "Only transparent pockets allow light to pass through them"

#### Purpose of change

It's how light behaves. Also more uses for transparent pockets.
Fixes #16461

#### Describe the solution

- Replace the old `has_items_with` call with a more nuanced `visit_items` call that only checks nested items in transparent pockets.
- Items on the map also check their pockets for light sources.

Todo:

 - [ ] item info for transparency
 - [ ] pocket prefence for light sources

Not doing it as a recursive lambda, but it's small enough (for now).

#### Testing

1. Spawn active atomic headlamp and messenger bag.
2. Go to dark room
3. Wear both-> worn headlamp illuminates room
4. Wield headlamp -> still illuminates room
5. Put headlamp in messenger bag -> no more light
6. Make messenger bag pocket transparent -> illuminates room again

Also more or less the same test just with dropping the messenger bag with headlamp on the ground isntead of wearing/wielding it.

#### Additional context

On hold until not direcly related requirements are in the game, e.g. dual wielding.

### DIFF
--- a/doc/JSON_INFO.md
+++ b/doc/JSON_INFO.md
@@ -3604,9 +3604,10 @@ Any Item can be a container. To add the ability to contain things to an item, yo
     "magazine_well": "0 ml",          // Amount of space you can put items in the pocket before it starts expanding.  Only works if rigid = false.
     "watertight": false,              // Default false. If true, can contain liquid.
     "airtight": false,                // Default false. If true, can contain gas.
-    "ablative": false,                 // Default false. If true, this item holds a single ablative plate. Make sure to include a flag_restriction on the type of plate that can be added.
+    "ablative": false,                // Default false. If true, this item holds a single ablative plate. Make sure to include a flag_restriction on the type of plate that can be added.
     "holster": false,                 // Default false. If true, only one stack of items can be placed inside this pocket, or one item if that item is not count_by_charges.
     "open_container": false,          // Default false. If true, the contents of this pocket will spill if this item is placed into another item.
+    "transparent": false,             // Default false. If true, the contents of this pocket are visible from a distance and lights inside can illuminate the surroundings.
     "fire_protection": false,         // Default false. If true, the pocket protects the contained items from exploding if tossed into a fire.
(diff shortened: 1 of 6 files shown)


GuardianDll:
How the game decide which pocket is transparent and which is not?

Karol1223:
> How the game decide which pocket is transparent and which is not?

I was about to ask the same question. Normally I'd assume it means that it works with open pockets, but what about stuff like zipper bags which evidently would let light through?

To add to that, does it work with items which are not in the players inventory? Previously putting a lit candle in a candlestick, from what I heard (haven't tested myself, though) would not cause any light if the candlestick was put on the ground.

mqrause:
> How the game decide which pocket is transparent and which is not?

`"transparent": true` in the pocket definition. Currently only corpses have that, because the only thing it does right now is enable to see the contents from a distance. And I just realized I didn't add that to docs when I initially implemented that. Will add it to this PR later.

kevingranade:
On so basically the change after this PR is only wearable or wielded lights actually work? That's a too far for right now.

mqrause:
What would be necessary prerequisites?

mqrause:
> To add to that, does it work with items which are not in the players inventory? Previously putting a lit candle in a candlestick, from what I heard (haven't tested myself, though) would not cause any light if the candlestick was put on the ground.

Also fixes items on the map now. Still requires transparent pockets, so a bunch of items will need adjustments whenever this gets merged.

Night-Pryanik:

PR #67598, opened 2023-08-11
[MoM] Simplify night eyes json

GuardianDll:
#### Summary
None
#### Purpose of change
Night Eyes spell was heavily overengineered, and i didn't like it
#### Describe the solution
Now spell call 1 EoC, that apply only 1 effect, and it's strength scale with intensity (which scales from player's level). 
The only reason it still call EoC is because spells can't apply effect of specific intensity
#### Additional context
image


Standing-Storm:
I tested this out and the additional intensity doesn't seem to be applied: 
At level 1: 
image
At level 10: 
image

I have one other concern on the approach--currently, all long-duration powers use intensity to act as a rough guide for the power's duration for the player, so they know how long the power will last, as well as use the decay_messages to warn them that the power will expire.  With no int_dur_factor this change will make Night Eyes the only power that doesn't do that.  Is there a way to retain at the warning messages when the power is about to expire so the player knows to re-channel it?

GuardianDll:
\> additional intensity doesn't seem to be applied

oh right, i forgot that default value for intensity is 1, i can fix it

\> long-duration powers use intensity to act as a rough guide for the power's duration for the player

that's surely a regression, but don't int_dur_factor make effect decay after some time? `"int_dur_factor": "68 s"` should mean effect start to decay 68 seconds after applying, and as spells apply the effect of intensity 1 no matter what, it should remove effect 68 seconds after applying, isn't it? or it works in some another way that i don't understand?

data/mods/MindOverMatter/effects/effects_psionic.json
```diff
@@ -352,143 +352,8 @@
     ],
     "rating": "good",
     "max_duration": " 120 minutes",
-    "max_intensity": 106,
     "dur_add_perc": 10,
```
GuardianDll, line 355:
```suggestion
    "max_intensity": 106,
    "dur_add_perc": 10,
    "int_dur_factor": "68 s",
```

Standing-Storm:
> that's surely a regression, but don't int_dur_factor make effect decay after some time? `"int_dur_factor": "68 s"` should mean effect start to decay 68 seconds after applying, and as spells apply the effect of intensity 1 no matter what, it should remove effect 68 seconds after applying, isn't it? or it works in some another way that i don't understand?

It does, but it also changes the normal intensity function--int_dur_factor overrides normal intensity and makes it so intensity is entirely determined by time, so the starting intensity is always duration divided by int_dur_factor and then it decays from there and gives "decay_message" warnings when it's about to expire.  If you use intensity to set the power rather than duration here, is there a way to retain the warnings that the effect will expire?

GuardianDll:
guess i'll close it untill we find a better way to handle effects

Issue #67451, opened 2023-08-05
MoM: radiation based powers


ilovehans10:
### Is your feature request related to a problem? Please describe.

Lack of mods that interact with radiation in major ways.

### Solution you would like.

I just got super irradiated, and I thought to myself it could be cool to make a radiation shielding/radiation vision powers. When I say radiation vision, it would be a way to see where radiation is by having either a Geiger counter sound or some sort of visual cue. I know that any sorry of radiation removal power would probably be op, but that could possibly be balanced by making the character go through a lot of pain all at once, which could be explained as purging all the radiation at once.

### Describe alternatives you have considered.

Possibly including these in a class like Bio/Vitakinesis as some of the powers would be similar to what those allow.

### Additional context

At the end of the day, I know there are the rules for new powers, and I don't believe this idea violates that at all. I am not well versed enough in the spell json yet to do this in a timely manner, so this idea is adoptable. If no one else takes this on, I plan on pursuing it some day, but I am too busy, probably until mid-winter.

Standing-Storm:
It definitely doesn't violate it because the biokinetic power Sealed System provides radiation resistance (which I should make clearer in the description). 

I have an idea for how to include this in Clairsentience (seeing) and Vitakinesis (curing), and I can definitely get to it before midwinter. IIRC visions options are limited without C++ but there's some things I can do.

PR #67604, opened 2023-08-11
[MoM] New powers part 4

Standing-Storm:
#### Summary
Mods "Add new powers, part 4"

#### Purpose of change

The final adding of new powers (for now) because I really should start concentrating on other things

#### Describe the solution

Adds four new powers:

1. Slowfall (Telekinesis): Greatly reduces falling damage, scaling to zero with higher power levels, and increases swimming buoyancy.  If player flying every becomes possible, the power that allows it will branch off from here.
2. Incandescent Lance (Pyrokinesis): Use pyrokinesis to cut through metal doors or demolish metal furniture or as a welding device. 

And also, two powers that don't come from the learning web, but have to be discovered and researched from old Project PHAVIAN files:

3. Radiation Decontamination (Vitakinesis): Purge radiation from your system, also possibly causing pain, reduced health, reduced stats, reduced stamina, increased fatigue, and vomiting, with severity determining how much radiation is being removed.
4. Radiation Sense (Clairsentience): Determine if you yourself have absorbed radiation or if your surroundings are dangerous.  The latter uses the radiation badge functionality because radiation detection is hard-coded.

<details> 
  <summary>How Radiation Sense Works </summary>
  It's an effect with a period enchantment that summons a fake rad badge every 10 seconds, so if there's radiation it'll change from green to blue than back to green continuously, and otherwise stay green.  That lets the power be useful for its entire duration instead of having to throw away your "radiation sense" and use it again.
</details>

Implements https://github.com/CleverRaven/Cataclysm-DDA/issues/67451

#### Testing

Powers work! The research recipe for discovering the powers from PHAVIAN research does too.

#### Additional context

While most powers will be learned through other powers, I'm going to implement other ways to find them like PHAVIAN research here, visiting particular locations, meditating during a portal storm, finding some weird hermit who took their powers in a different direction and can teach you if you do a quest, etc, because hunting for loot is fun.


Cupidatis:
@Standing-Storm has USB (for radiation powers) itemgroups for spawning in world? Never found this...

PR #67606, opened 2023-08-11
Add transparency to some pockets

mqrause:
#### Summary
Bugfixes "Add transparency to some pockets"

#### Purpose of change

Pocket transparency is a somewhat new concept that will get more use in #67595. It allows seeing the contents of the pocket from a distance (`V` menu), and will be expanded to allow light sources inside them to illuminate the surroundings. But there aren't any pockets actually using it, yet, apart from corpses.

#### Describe the solution

Made a script to determine a baseline of pockets that could or should be transparent.
That currently includes
- pockets restricted to the flags `BELT_CLIP` or `HELMET_HEAD_ATTACHMENT` (flags of the flashlight item)
- pockets with `ripoff` and `extra_encumberance` values > 0

Identify more conditions that justify inclusion.

### DIFF
--- a/data/json/items/armor/ammo_pouch.json
+++ b/data/json/items/armor/ammo_pouch.json
@@ -1335,7 +1335,8 @@
         "max_item_length": "30 cm",
         "moves": 50,
         "flag_restriction": [ "BELT_CLIP" ],
-        "volume_encumber_modifier": 0.3
+        "volume_encumber_modifier": 0.3,
+        "transparent": true
       }
     ],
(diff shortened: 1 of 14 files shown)


anoobindisguise:
what about the flashlight_pouch? isn't that supposed to be a dedicated "strap flashlight here" hands free option?

mqrause:
According to the link provided in the item definition, it's just to keep it easily accessible while you don't need it. But this PR is just meant to get the word out there that pocket transparency exists and to apply it a bunch of pockets automatically. A lot of stuff will be missed and some stuff might be inaccurate.

PR #67608, opened 2023-08-11
Manually fix some (but probably not all) of the broken itemspawns

anoobindisguise:
#### Summary
Bugfixes "correct some of the item spawns that were only 1"

#### Purpose of change
Work towards resolving #67326

#### Describe the solution
Manually adjust a lot of itemgroups and monsterdrops.

Suffer even more to finish the job.


Maleclypse:
Double checking have you tested the spawns to make sure that the pills aren't spawning however many count of bottles each with one pill?

anoobindisguise:
> Double checking have you tested the spawns to make sure that the pills aren't spawning however many count of bottles each with one pill?

nope lmao. gonna have to do the whole thing over now.

SurFlurer:
Thank you for doing this!

I'm working in #67410 to provide a C++ way to make inline itemgroup with container possible. But there are leftovers as mqrause has said in https://github.com/CleverRaven/Cataclysm-DDA/pull/67410#issuecomment-1664274443 .

PR #67617, opened 2023-08-12
Add electronics control unit to Street-Legal Dirt Bike

marimarigi:
#### Summary
None

#### Purpose of change
This bike doesn't have a control unit for turning on the headlight, added it.

#### Testing
Spawn Street-Legal Dirt Bike, examine it, turn on headlight.


Maleclypse:
I believe you've made this so that maintainers can't edit which means I can't fix your error and get this in a mergeable state for you.   You'll have to lint it yourself at https://dev.narc.ro/cataclysm/format.html.

PR #67628, opened 2023-08-13
Use llvm tools to cut windows build noop critical path by 85% or more

akrieger:
#### Summary
Build "Fast Windows iteration with llvm-lib and lld-link"

#### Purpose of change
A lot of the time in the windows build is spent re-assembling the cata lib static archive, which is largely i/o bound. llvm tools support creating 'thin' archives which are indexes of objects instead of copies of objects, making this step almost instantaneous. To use this though we have to also use lld-link. lld-link is generally faster than link.exe anyway, although link.exe is more competitive nowadays. Most people are doing 'release' builds locally which are typically not set up for the fastest iteration speeds. We can safely allow people to opt-in to this behavior without sacrificing anything about the usual build process.

#### Describe the solution
Add a new property people can set in Directory.Build.props `CDDA_ENABLE_THIN_ARCHIVES`, which accepts any nonempty value, to opt into the new behavior. When set, override the lib and link tools to be llvm-lib and lld-link, respectively. Either use the default VS installed llvm tools or read a user provided path from `CDDA_LLVM_LIB_PATH` and `CDDA_LLD_LINK_PATH` (separate for debugging one binary or another).

Naively enabling llvm-lib caused segfaults when used in combination with ccache, because ccache was sharing stdafx.obj between different projects and this tickled llvm-lib the wrong way (it was reading the 'wrong' source file name from the object file which messed up some internal maps). So, gate off the stdafx source from being included at all when ccache is used.

lld-link doesn't work with vcpkg's autolink/integration because vcpkg passes libs via a glob pattern like `vcpkg_installed/triple/triple/*.lib`. So we have to manually list all the libs in the folders, with the right libpath, and also their windows lib dependencies. But it work.

#### Testing
Build locally. See that CI has a much faster lib step and potentially faster link step too.

### DIFF
--- a/msvc-full-features/Cataclysm-common.props
+++ b/msvc-full-features/Cataclysm-common.props
@@ -13,12 +13,30 @@
     <_CDDA_BACKTRACE Condition="'$(BACKTRACE)'!=''">true</_CDDA_BACKTRACE>
     <_CDDA_RELEASE_BUILD>false</_CDDA_RELEASE_BUILD>
     <_CDDA_RELEASE_BUILD Condition="'$(CDDA_RELEASE_BUILD)'!=''">true</_CDDA_RELEASE_BUILD>
+    <CDDA_LLVM_LIB_PATH Condition="'$(CDDA_LLVM_LIB_PATH)'==''">$(LLVMInstallDir)</CDDA_LLVM_LIB_PATH>
+    <CDDA_LLD_LINK_PATH Condition="'$(CDDA_LLD_LINK_PATH)'==''">$(LLVMInstallDir)</CDDA_LLD_LINK_PATH>
+    <_CDDA_ENABLE_THIN_ARCHIVES>false</_CDDA_ENABLE_THIN_ARCHIVES>
+    <_CDDA_ENABLE_THIN_ARCHIVES Condition="'$(CDDA_ENABLE_THIN_ARCHIVES)'!=''">true</_CDDA_ENABLE_THIN_ARCHIVES>
+    <_CDDA_USE_LLD_LINK>false</_CDDA_USE_LLD_LINK>
+    <_CDDA_USE_LLD_LINK Condition="'$(CDDA_USE_LLD_LINK)'!='' Or $(_CDDA_ENABLE_THIN_ARCHIVES)">true</_CDDA_USE_LLD_LINK>
   </PropertyGroup>
   <PropertyGroup Label="Vcpkg">
     <VcpkgAdditionalInstallOptions>--clean-after-build</VcpkgAdditionalInstallOptions>
   </PropertyGroup>
   <PropertyGroup Condition="$(_CDDA_USE_CCACHE)">
     <ClToolPath>$(CDDA_CCACHE_PATH)</ClToolPath>
+  </PropertyGroup>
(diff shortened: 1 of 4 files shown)


akrieger:
Well thats awkward, that llvm-lib is also crashing. Rip.

akrieger:
Drafting so I can try to debug wtf the github action is on.

akrieger:
The bitter irony is that although this speeds up local builds a nontrivial amount for me, it prevents using `/DEBUG:FASTLINK` on github which is dramatically faster than LLD can do with full link. So enabling this for our workflows would be a net loss. We'd save ~40-60s on the lib step but lose the same amount on both the full binary and tests binary link steps, on top of slowdowns for object creator and json formatter. RIP.

Maleclypse:
There's a conflict and I'm not sure from your comments whether you even want this to be merged?

akrieger:
Ah, it conflicts with the other PR. I'll resolve.

I do want it merged, I think it's strictly better than the other PR for people willing to set it up (and it's very easy to do so). It's just funny this in particular is not beneficial for CI because despite all attempts to do so I can't convince msbuild to link all the binaries in parallel, so the speed wins are lost.

msvc-full-features/Cataclysm-common.props
```diff
@@ -67,6 +88,56 @@
...
     </Link>
+    <Link Condition="$(_CDDA_USE_LLD_LINK)">
+      <AdditionalLibraryDirectories Condition="$(Configuration.StartsWith(Debug))">
+        $(ProjectDir)vcpkg_installed\$(VcpkgTripet)\$(VcpkgTriplet)\debug\lib;%(AdditionalLibraryDirectories)
+      </AdditionalLibraryDirectories>
+      <AdditionalLibraryDirectories Condition="$(Configuration.StartsWith(Release))">
+        $(ProjectDir)vcpkg_installed\$(VcpkgTripet)\$(VcpkgTriplet)\lib;%(AdditionalLibraryDirectories)
+      </AdditionalLibraryDirectories>
+      <AdditionalDependencies>
+        brotlicommon-static.lib;
```
ZhilkinSerg, line 99: I had to remove `-static` to link on latest vcpkg master.

msvc-full-features/Cataclysm-common.props
```diff
@@ -67,6 +88,56 @@
...
+        brotlidec-static.lib;
+        brotlienc-static.lib;
+        bz2$(VcpkgLibSuffix).lib;
+        FLAC.lib;
```
ZhilkinSerg, line 103: We should probably hide SDL sound related libs behind `SDL_SOUND`

msvc-full-features/Cataclysm-common.props
akrieger: Unfortunately I haven't figured out a way to get this to figure out whether the vcpkg checkout is 'new enough' or not so that it can name the libs differently. So for now I keep it up to date with the documented 'known good' vcpkg revision.

msvc-full-features/Cataclysm-common.props
akrieger: If the symbols aren't needed because they're not compiled into the binary, then listing these extra libs should have no effect on the binary.

PR #67631, opened 2023-08-13
Optimize pocket overflow function

Kamayana:
#### Summary
Performance "Optimize pocket overflow function"

#### Purpose of change
`item_pocket::overflow` is a common sight at the top of my profiler's CPU% results, a function that spends a lot of CPU time checking `can_contain` on every individual item in a pocket whenever the player's inventory validity is checked, which is fairly often. This intensive check is often not even necessary, as it's specifically checking for items that should never have been in the pocket in the first place.

#### Describe the solution
Add an unordered map that caches all of the `can_contain` results for every item type, and just use the cached result for all other items of that type.

The consolidation of all contents into a map of types is something that could be useful for other optimizations, so in the future it might be good to have a function for creating that map.

`process_items` is still the elephant in the room, optimization-wise, but I wanted to take on this first as a stepping stone towards that.

#### Testing
Tested enough to get profiler results. Tested around with frozen liquids, as that's the one case where you can't necessarily treat items the same just because they share a type. Didn't notice anything wrong occur, but I could've missed something, it wasn't the easiest thing to manually test.

##### Before
Results right after spawning 6000 pills into a pocket universe (keep in mind that Visual Studio is giving the red highlight to the line AFTER a bottleneck):

image

##### After
Results after the same test - note how `drop_invalid_inventory` isn't even using 1 CPU unit anymore:

image

### DIFF
--- a/src/character_attire.cpp
+++ b/src/character_attire.cpp
@@ -1168,7 +1168,9 @@ units::mass outfit::weight_carried_with_tweaks( const std::map<const item *, int
 {
     units::mass ret = 0_gram;
     for( const item &i : worn ) {
-        if( !without.count( &i ) ) {
+        if( without.empty() ) {
+            ret += i.weight();
+        } else if( !without.count( &i ) ) {
             for( const item *j : i.all_items_ptr( item_pocket::pocket_type::CONTAINER ) ) {
                 if( j->count_by_charges() ) {
                     ret -= get_selected_stack_weight( j, without );
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -1282,11 +1282,13 @@ void item::update_modified_pockets()
     contents.update_modified_pockets( mag_or_mag_well, container_pockets );
 }
 
-int item::charges_per_volume( const units::volume &vol ) const
+int item::charges_per_volume( const units::volume &vol, bool suppress_warning ) const
 {
     if( count_by_charges() ) {
         if( type->volume == 0_ml ) {
-            debugmsg( "Item '%s' with zero volume", tname() );
+            if( !suppress_warning ) {
+                debugmsg( "Item '%s' with zero volume", tname() );
+            }
             return INFINITE_CHARGES;
         }
         // Type cast to prevent integer overflow with large volume containers like the cargo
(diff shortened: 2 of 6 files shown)


PatrikLundell:
Are weapons (and other things) with mods an issue? I believe the volume and length differs depending on the mods attached, so the actual modded versions have to be treated as different templates.

Kamayana:
> Are weapons (and other things) with mods an issue? I believe the volume and length differs depending on the mods attached, so the actual modded versions have to be treated as different templates.

Yeah, good call, I'll make it so items with filled mod/cable pockets get checked individually.

PatrikLundell:
Your response caused me to realize usage of "magazines" should probably result in exclusion as well. While fully internal magazines won't affect the volume, they still affect the weight, and "magazines" containing physical resources (as opposed to electric charge, whose weight effects are extremely small), consumption of bullets or welding gas ought to reduce the weight of the item taken as a whole.

Kamayana:
> Your response caused me to realize usage of "magazines" should probably result in exclusion as well. While fully internal magazines won't affect the volume, they still affect the weight, and "magazines" containing physical resources (as opposed to electric charge, whose weight effects are extremely small), consumption of bullets or welding gas ought to reduce the weight of the item taken as a whole.

This particular check isn't actually checking for an invalid width/volume, it's only checking if the item type doesn't belong there at all. So a mod making the item too long for the pocket needs to be checked, but the weight/volume is irrelevant since that's handled later.

Kamayana:
Actually, on further examination, it looks like you're actually right - ERR_TOO_HEAVY is a different error from ERR_CANNOT_SUPPORT, so it should actually still be checked, along with volume-related ones. I might have to have special handling for all items with any contents.

Kamayana:
`overflow` will now check items individually if they have any contents, which unfortunately returned things like water bottles to the status quo, with spawning 10000 of them taking extremely long to process. However, I realized that, because the weight and volume checks are ignored in `overflow`, I could make a version of `can_contain` that skips those checks entirely. The weight and volume checks are the most intensive part of the function, so after these changes containers once again processed significantly faster. The cached check that's used for empty/non-containers is still much faster, but this is an acceptable alternative.

Kamayana:
I'm changing this back into a draft because it turns out the major optimizations I was putting off look like they'll actually be pretty easy to implement, so it wouldn't be increasing this PR's scope too much to include them.

Kamayana:
That's all the optimizations I think could fit into this PR. The biggest CPU hog now are all the item visit functions - adding caches to those would be a huge improvement, but that's well beyond the scope here.

PR #67640, opened 2023-08-13
add in new item category and zones: corpses

ghost:
#### Summary
Features "Corpses are now their own separate category and have a zone"

#### Purpose of change
being able to automatically sort corpses would be pretty nice

#### Describe the solution
addition of category and zone, and a category: corpses in the generic corpse code

None

#### Testing
loaded in, killed a bunch of random monsters, they're all in the correct category
also spawned the map-gen specific corpses, they're all in the correct category

#### Additional context
NOTE: this excludes broken robots and inactive robots, i don't think these count as corpses?

### DIFF
--- a/data/json/item_category.json
+++ b/data/json/item_category.json
@@ -183,6 +183,13 @@
     "sort_rank": 11
   },
   {
+    "id": "corpses",
+    "type": "ITEM_CATEGORY",
+    "name": { "str": "CORPSES" },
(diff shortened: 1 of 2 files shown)


PatrikLundell:
Given that "further processing" of corpses is butchering and dissection, while "further processing" of broken and inactivated robots is dismantling and reactivation (and possibly repair?), I think it's reasonable to not include robots in the corpse category.

PR #67656, opened 2023-08-14
Add some mod compatibility to Defense Mode

MNG-cataclysm:
#### Summary
Mods "Add some mod compatibility to Defense Mode."

#### Purpose of change
Add some cross-mod compatibility to Defense Mode.

#### Describe the solution
This has been desired for a long time, and with the new JSON changes, it can be done. Monsters from DinoMod and Magiclysm can now appear in Defense Mode waves, if you so choose. The supporting EOCs, species changes, and monster alterations are included within the Defense Mode mod itself, as I couldn't find a way to have two instances of an EOC combine across mods.

Not doing this.

#### Testing
Played through Defense Mode for a bit, everything works well.

#### Additional context
Following this example, any other mod creator should be able to hook up their own creations to Defense Mode and have their own creatures appear in waves. We might be able to fight alien magic dinosaurs from another reality with wizard powers and ray guns at some point with the right combinations.

Additionally, if a way exists to have two of the same instances of an EOC merge in their functions while going across mods, these functions could be exported to their respective mods to help cut down on file size and expand mod options to include out-of-repo things.


LyleSY:
This looks like a great approach.  Am I right that it would be possible to add other DinoMod monster groups like zombies or bio-operators or zombie bio-operators? I'm guessing fungals wouldn't make any sense for this game style.

MNG-cataclysm:
> This looks like a great approach. Am I right that it would be possible to add other DinoMod monster groups like zombies or bio-operators or zombie bio-operators? I'm guessing fungals wouldn't make any sense for this game style.

Yes, it's possible to add any other monsters you'd like. I considered fungal enemies for Defense Mode and spoke about it on Discord, but they'd be too powerful and quickly overwhelm the player with spores in a static game.

LyleSY (approved):
DinoMod content looks great

PR #67657, opened 2023-08-14
Obsolete RPG-7 and trim zombie survivor launchers item group

kevingranade:
#### Summary
None

#### Purpose of change
I was reminded that the RPG-7 spawns in a few places in the game and there is no rationale for this happening as this weapon system is not in used by any US armed force, it's expensive, it's unusable for any purpose, and it's heavily regulated.

#### Describe the solution
Remove the spawns and move the related definitions to the appropriate obsoletion files.
I also noticed the M79 and flamethrower in the survivor zombie loot table, which are likewise too rare for this loot table. These items spawn elsewhere though.

It's not *impossible* for a RPG to appear somewhere, but that should be some specific and rare scenario, not a general loot table.
I'm not personally interested in chasing down such a scenario.
I'm looking askance at the flamethrower and M79 spawns, but I don't have time to chase them down right now.


Karol1223:
> I also noticed the M79 and flamethrower in the survivor zombie loot table, which are likewise too rare for this loot table

I believe I handled the flamethrower in #67646. M79 is *not* in that group, though, but I suppose that's half of the problem gone at least.

PR #67662, opened 2023-08-14
New item: autopsy kit

ampersand55:
#### Summary
Content "New item: Autopsy kit"

#### Purpose of change

We have scalpel with fine cutting 3 and autonomous surgical scalpels with fine cutting 5, and it would be nice to have something in between to smooth out progression.

* https://cdda-guide.nornagon.net/tool_quality/CUT_FINE

#### Describe the solution

Introducing the autopsy kit, something that is commonly used by veterinarians but will probably work on zombies, cyborgs and other stuff too to learn about their biology and extract CBM's. 

I used these for reference:

* http://www.snaavet.com/post-mortem-set-for-veterinary
* https://kruuse.com/products/kruuse-post-mortem-kit
* https://cti-surgivet.com/shop/post-mortem-set-veterinary/

It consists of:

| Item | Volume | Weight | Comment | 
|---|---|---|---|
| Scalpel | 0.01 L | 0.02 kg | Basic fine cutting tool | 
| Pair of medical scissors | 0.10 L | 0.05 kg | - | 
| Pair of medical gloves | 0.01 L | 0.00 kg | Not strictly necessary, but you would use these when doing an autopsy. |
| Hunting knife | 0.25 L | 0.39 kg |Analogous to a boning knife and other knives you find in most kits |
| Hacksaw | 1.00 L | 0.95 kg | Analogous to a bone saw you find in most kits |
| Tin snips| 0.50 L | 0.40 kg | Analogous to a bone-cutting shears you find in most kits |
| Total | 1.87 L | 1.81 kg | | 

Length 30 cm from the longest items, price_postapoc set to $15 as it's imho a pretty valuable tool.

Added to item groups "surgery" being 8% as common as scalpel, and the veterinarian "vet_softdrug" being 20% as common as scalpel.

You could add a hatchet or meat cleaver as most irl kits use one of those, but that didn't seem like it would help increase the chance to extract CBMs or learn about biology. 

I also thought about adding syringe, pliers, blood draw kit, awl but decided against it as mot irl kits doesn't include those.

I tried to make it be loaded by medical gloves to use as charges, but I couldn't make it work. 

#### Testing

Loads and works as intended.

#### Additional context

image

image


ampersand55:
The differences in dissecting time is very slim and should imho be updated, but I can't compile c++ on this machine so I probably shouldn't do the update. The time is scaled by (100 - fine cutting quality), so the differences are minimal. I propose something like (100 - 10 * fine cutting quality), i.e. 44% quicker at 5 fine cutting quality.

It takes 51 minutes to dissect a regular zombie with fine cutting 4 compared to 53 minutes with fine cutting 1.

ampersand55:
I fully agree that that there are things about kits that are problematic, but I think having item synergies or combined item qualities in a sub-optimal way is better than not having them at all. 

Kits are easily revertible once/if we get proper item synergies. They don't require any infrastructure or introduce any dependencies. 

Plus, it's fun to go around searching for items to be able to assemble a kit.

BjoHart:
Thinking of someone doing a full body autopsy with only one scalpel is weird in my mind. 
Having a full set of tools would be more realistic imo.

Karol1223:
> Sounds reasonable, but how would we make it apparent to the player that they would dissect better if they had a combination of items? Putting it in the butchering menu would be a bit cluttered.

...a flag? Flags can print messages in item descriptions, like ``DURABLE_MELEE`` does. I'm brainstorming here, because this is a valid concern. This should be transparent to the player, no reason to hide this knowledge.

ampersand55:
You mean a flag on each item that helps with dissecting? That would be a good solution if dissection quality increased incrementally for every dissection item you have, but perhaps not if all items had to be present. 

As we're brain storming, we could have incremental tool qualities, like if you had one tool with `DISSECTION_HELP 1` and another with `DISSECTION_HELP 2` you'd have `DISSECTION_HELP 3` to factor in when dissecting.

Karol1223:
> You mean a flag on each item that helps with dissecting?

Yes but there's more to that. First of all it'd have into account only having one of this specific item so you can't just use 20 scalpels at once to become a medical god. Second of all, you'd have to somehow sort out items serving the same role (without adding 20 new qualities/flags). This is very tricky.

ampersand55:
If it's < 20 qualities or flags, it might be easier to just check for them than to add a "tricky" check for dupes, though I see how that would go against one's instincts as a programmer.

Also, this is obviously out of scope for this PR. I humbly request we add this kit in the interim, and remove it and any other assembled kits when we have item synergies.

fairyarmadillo:
> Sounds reasonable, but how would we make it apparent to the player that they would dissect better if they had a combination of items? Putting it in the butchering menu would be a bit cluttered.

Note that flat surfaces currently help with dissection but are not required. Specifically they reduce the work time. If that information isn't displayed in the butchery menu it probably should be.

ampersand55:
I agree. Would be easy to just add that it to the dissect menu msg.

Fris0uman (changes requested):
We're moving away from weird one-item-kits, we have actual boxes now so a kit is the box with the things in it. If you want to go and make dissection use or benefit from having multiple tools I thinks it's a good idea but we should not start adding back one-item-kits

ampersand55:
I don't have any experience with c++, so a multi-quality check is not something I can do in this PR. 

Would it be fine if I removed the player recipe and just added the kit?

Fris0uman:
I think the kit itself being one item is the issue, if you want dissection to need more tools you could do that through requirements I think, but I'm prety sure you could only make it be a on/off thing and not havign a bonus from having more tools.

PR #67678, opened 2023-08-15
Rearranged main menu to show Tutorial on top level

### LINKED ISSUE #51413: Tutorial Uses Default Mod List, Should Force Vanilla
### Describe the bug

Trying to boot up the tutorial loads your default mod list.  If you have broken mods in your default mod list, this breaks the tutorial's ability to load.  Updating the default mod list does not fix this problem, you actually have to delete the tutorial world AND update your default mod list before attempting to load the tutorial.  This will not be obvious to new players who jump into the game, may possibly use broken mods, and then find themselves locked out of the tutorial.

### Steps To Reproduce

1. Have a default mod list
2. Load Tutorial
3. Observe that it attempts to load your default mod list
4. Update default mod list in the world creation menu
5. Reload tutorial (by using the special options -> tutorial)
6. Observe that it tries to load your original mod list (because the tutorial world was already generated as soon as you tried to load it the first time)
7. The only way to fix this is to delete the world that is generated from the World-menu AND update your default mod list to not include broken mods, something that is not obvious to a new player.  This will generate the tutorial world again with your new default mod list.

### Expected behavior

I believe that the tutorial should force vanilla only.  There will inevitably be players who set up a world, save defaults with broken mods, and then try the tutorial, which will then not load.  I know this is a bit niche but it's an issue.

### Screenshots

_No response_

- OS: Windows
    - OS Version: 10.0 2009
- Game Version: 0138041 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Caches [P_Caches],
    Oa's Additional Buildings mod [oa_additional_buildings_mod],
    StatsThoughSkills IV [StatsThroughSkillsFour]
]

### Additional context

I'm not sure if there's support at the moment for forcing vanilla or not.  The only solution I can think of is to force use only DDA when starting the tutorial.  I can't think of a way to illustrate to new players that they will need to delete the TUTORIAL world from their world-menu and fix their default mods to solve the tutorial loading issue.

ZhilkinSerg:
#### Summary
Interface "Show Tutorial on top level in main menu"

#### Purpose of change
More visibility for tutorial.
Also fixes #51413

#### Describe the solution

Removed remaining code references to DEFENSE game mode and moved TUTORIAL game mode to top level.
Use only `dda` mod in tutorial.

We should eventually get rid of hardcoded special game modes and move tutorial to EOC driven mod, that can be updated more easily.

#### Testing
Run tutorial from main menu.

#### Additional context

image

### DIFF
--- a/src/enums.h
+++ b/src/enums.h
@@ -150,7 +150,6 @@ struct enum_traits<ot_match_type> {
 enum class special_game_type : int {
     NONE = 0,
     TUTORIAL,
-    DEFENSE,
     NUM_SPECIAL_GAME_TYPES
 };
 
--- a/src/main_menu.cpp
+++ b/src/main_menu.cpp
@@ -54,7 +54,7 @@ enum class main_menu_opts : int {
     NEWCHAR,
     LOADCHAR,
     WORLD,
-    SPECIAL,
+    TUTORIAL,
     SETTINGS,
     HELP,
     CREDITS,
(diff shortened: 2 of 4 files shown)


ZhilkinSerg:
> Spell checker encountered unrecognized words in the in-game text added in this pull request. See below for details.
> Click to expand
> 
> This alert is automatically generated. You can simply disregard if this is inaccurate, or (optionally) you can also add the new words to `tools/spell_checker/dictionary.txt` so they will not trigger an alert next time.

Yep, it would look better, but will probably do that in a separate PR - I'm having a stroke looking through main menu code lol.

image

PR #67682, opened 2023-08-15
[Mod] Lights Out: PVE survival in permanent night

Fris0uman:
#### Summary
Mods "Lights Out: PVE survival in permanent night"

#### Purpose of change

- Adds a small mod where you're all alone in a permanent night, there's no monsters and no NPCs, only you, the leftover robots and the wildlife
-  adds and use an external option to disable static NPCs

#### Describe the solution

- [x] Basic jsons, scenarios and blacklists
- [x] Disable NPCs
- [ ] Change the seasons, they should go something like spring> autumn> winter > more winter

#### Testing

Start the game with the mod, no error, no NPC, no sunlight

#### Additional context

This is a pretty small mod, maybe it's not worth it to put it in the repo. Putting it up as a draft still as I htink of other things to flesh out this setting

### DIFF
--- a/data/core/game_balance.json
+++ b/data/core/game_balance.json
@@ -362,5 +362,12 @@
     "info": "When mutating, displays a menu which allows players to pick from a list of possible mutations.",
     "stype": "bool",
     "value": false
+  },
+  {
+    "type": "EXTERNAL_OPTION",
+    "name": "STATIC_NPC",
+    "info": "Spawn static NPCs.  Does not affect the starting NPC or random NPC.",
+    "stype": "bool",
+    "value": true
   }
 ]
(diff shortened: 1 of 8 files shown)


anoobindisguise:
Does the "shadow" lieutenant still spawn? Would be fitting given it's kind of like a grue and Lights Out is a term from Don't Starve, I assume it's an intentional reference

Nebnis:
I think it would be more fun if you left the monsters in, otherwise this feels like an easier innawoods where you can just go live in the cities without all the monsters. Also since there is no more sun, no more crops either.

I like where you going with this idea, reminds me of the long winter from Game of thrones.

Fris0uman:
I don't actually have the energy to make this interesting. I think it'd require at least some darkness themed monsters that are atracted to light/ agressive toward light but passive and/or blind in darkness

PR #67683, opened 2023-08-15
Add ETERNAL_TIME_OF_DAY to the list of external options

Fris0uman:
#### Summary
None

#### Purpose of change

Add ETERNAL_TIME_OF_DAY to the list of external options, to document it.
Maybe it's normal that it's not there because it's actual an option that can be tweaked from within the game and putting it there will mess that up

### DIFF
--- a/data/core/game_balance.json
+++ b/data/core/game_balance.json
@@ -372,6 +372,13 @@
   },
   {
     "type": "EXTERNAL_OPTION",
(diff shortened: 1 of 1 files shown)


Fris0uman:
setting the option in the json, overrides the option in the menu, and I don't see any entry for EXTERNAL_OPTION in the doc. So unless we want to remove the option from the menu to move it to json I'm closing this

PR #67705, opened 2023-08-16
Make Yrax Trifacet untameable

Karol1223:
#### Summary
None

#### Purpose of change
~~We had a person come to the discord with a screenshot of a trifacet they somehow reactivated and got a friendly one. Why is that even possible, in the first place? Reactivating a robot you cannot fully comprehend is bullshit as is, reprogramming it is entirely impossible.~~

#### Describe the solution
~~Remove the use action of the inactive trifacet. The player should not be able to handle technology of this level.~~ Check comments

~~Keeping it activable but always spawning a hostile, but really the player should not have a set of controls or a guide on how to work with an Yrax. Deactivating it in the first place is already iffy but removing that would break Hub quests.~~

#### Additional context
I imagine the Hub with enough figuring out would likely find a way to reactivate it, but this does not in any way mean the player should be capable of doing so. And the question of whether the Hub could reprogram it is a different one entirely

### DIFF
--- a/data/json/effect_on_condition.json
+++ b/data/json/effect_on_condition.json
@@ -1333,5 +1333,13 @@
       { "u_add_trait": "CHITIN2" },
       { "u_add_trait": "CRUSTACEAN_CARAPACE" }
     ]
+  },
+  {
+    "type": "effect_on_condition",
+    "id": "yrax_trifacet_activation",
+    "effect": [
+      { "u_spawn_monster": "mon_yrax_trifacet", "real_count": 1, "min_radius": 1, "max_radius": 2 },
+      { "u_message": "The hostile trifacet violently unfolds just clear of your hand!", "type": "bad" }
+    ]
   }
 ]
--- a/data/json/items/corpses/inactive_bots.json
+++ b/data/json/items/corpses/inactive_bots.json
@@ -751,16 +751,7 @@
     "material": [ "steel", "ceramic" ],
     "symbol": ",",
     "color": "cyan",
-    "use_action": {
-      "type": "place_monster",
-      "monster_id": "mon_yrax_trifacet",
-      "friendly_msg": "Your attempt at reprogramming the trifacet is somehow successful!",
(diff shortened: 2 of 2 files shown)


John-Candlebury:
Keep it activable as it can have interesting(dangerous) tactical uses, and you can actually get more through some obscure hard way.
I didnt consider the 12 sill requirement would be achievable, just raise the value to something that is actually impossible.

Karol1223:
> Keep it activable as it can have interesting(dangerous) tactical uses, and you can actually get more through some obscure hard way. I didnt consider the 12 sill requirement would be achievable, just raise the value to something that is actually impossible.

Yeah honestly I wouldn't expect the 12 skill requirement to be achievable either but now I'm not sure what *is* impossibly high anymore. 20 is probably a good bet, though, I suppose.

Still not very convinced about the activation at all, but I believe it's your content so I'll just go with it. I'll handle this a bit later today.

Karol1223:
> I thought skills could only get to 10, how did someone get an electronics/computer skill of 12?? Also, why go with 20 when something like 100 is definitely safe, and makes it clear that this is not supposed to ever happen?

I think, if it works similar to how recipes work, it may actually allow chances with skill levels two below which means a skill level 10 would technically have a shot. Not sure, though.

Karol1223:
> Unless there's a cap putting it even higher to get around mods would probably be good, it might be worth testing in game quick that more than 10 doesn't default to 0 or something silly too

That's actually a fair point, I'll test it to be sure. I don't think the code supports levels higher than 11/14, though, so I don't think mods can go above this either.

Karol1223:
You were right. Third attempt...

image

Karol1223:
Fun fact! At difficulty 100 it is STILL POSSIBLE! What's going on?!

John-Candlebury:
Hmm ping me over in discord if you can't get it to work by day's end. It might need some c++ changes.

Karol1223:
Now this works. I tested it. With the powers of EOCs.

PR #67717, opened 2023-08-17
Speed up new character screen, particularly when many recipes are known

Kamayana:
#### Summary
Performance "Speed up new character screen, particularly when many recipes are known"

#### Purpose of change
The new character menu can become exceedingly sluggish when knowledge skills are leveled up and more recipes become known. This is due to the difficulty display: it's constantly initializing a character in the background to get values from, and part of the initialize function is learning recipes. This means the game is constantly looping through the entire library of recipes for no reason, as the character's recipe knowledge isn't used for calculating the difficulty ratings.

#### Describe the solution
Add a default true `learn_recipes` boolean parameter to `Character::initialize` that enables the recipe loop, make the call in `player_difficulty::npc_from_avatar` give it false to skip the loop.

#### Testing
Made the change, notice how much faster the screen runs, especially when many recipes are known. Adjusted knowledge skills to see that the known recipes list still adds and removes them accurately, and that the difficulty rating isn't thrown off. Actually started the game and saw that my character knew all the recipes they should.

#### Additional context
The profiler results from before my changes:

image


Maleclypse:
There are a couple of the below warnings popping up
```
Check failure on line 80 in src/newcharacter.cpp

GitHub
Actions
/ build (src)
Declaration of string_id 'prof_group_id_adult_basic_background' should be named 'profession_group_adult_basic_background'. [cata-static-string_id-constants,-warnings-as-errors]
```

Kamayana:
> There are a couple of the below warnings popping up
> 
> ```
> Check failure on line 80 in src/newcharacter.cpp
> 
> GitHub
> Actions
> / build (src)
> Declaration of string_id 'prof_group_id_adult_basic_background' should be named 'profession_group_adult_basic_background'. [cata-static-string_id-constants,-warnings-as-errors]
> ```

I'm pretty sure that's unrelated, I didn't touch those strings. I rebased to master anyway in case that'd fix it.

PR #67718, opened 2023-08-17
Multiselector menu for selecting seeds for basecamp farm missions + fix for seed duplication

mqrause:
#### Summary
Interface "Multiselector menu for selecting seeds for basecamp farm missions"

#### Purpose of change

Turned out the charge removal broke a couple things with basecamp farming. This aims to fix them.

#### Describe the solution

- added a multiselector for seed selection
   - includes a method to add items from basecamp storage to inventory selector menus
- seeds are consumed upon assigning the mission and removed from the npc inventory after planting

#### Testing

Assigned and completed a farming mission. Seeds got used up, planting worked as expected.

### DIFF
--- a/src/basecamp.h
+++ b/src/basecamp.h
@@ -340,7 +340,8 @@ class basecamp
                             float exertion_level );
         ///Display items listed in @ref equipment to let the player pick what to give the departing
         ///NPC, loops until quit or empty.
-        std::vector<item *> give_equipment( std::vector<item *> equipment, const std::string &msg );
+        drop_locations give_basecamp_equipment( inventory_filter_preset &preset, const std::string &title,
+                                                const std::string &column_title, const std::string &msg_empty ) const;
         drop_locations give_equipment( Character *pc, const inventory_filter_preset &preset,
                                        const std::string &msg, const std::string &title, units::volume &total_volume,
                                        units::mass &total_mass );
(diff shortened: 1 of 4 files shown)


PatrikLundell:
I tried (and failed) to improve this in the past (well before charges were removed), and had issues with seeds when the stack size was larger than the number of seeds required. I believe there was duplication in some scenarios, and if I recall correctly, the behavior was different between getting seeds from the PC's inventory compared to from the ground where the PC stood. I think some of the issue was what to do with the remainder of the seed stack, ensuring it wasn't duplicated or deleted. I think there were issues with full PC inventories and lack of room at the PC's feet as well.

This is a long winded way of saying that I think the solution requires testing of a lot of weird cases, although the solution may well have addressed the root of the problem, so symptom handling wouldn't be needed.

mqrause:
All this stuff is definitely worth a refactoring, and I can definitely imagine some issues still remain. I just want to focus on fixing charge removal issues for now (in fact the behavior for possible seeds that still use charges (none in this repo I think) should be more or less the same as before). Best make some separate issues when you notice something, even if it's just minor.

PatrikLundell:
A refactoring of the complex to use the newer type that contains a reference to the data intended to be used rather than the current one that essentially uses a look alike and then tries to find the original by a second search would be useful. However, as you said, that would be out of scope for this issue and it looks like it would require quite a lot of expertise and knowledge to succeed (how would you even (de)serialize such a thing, for instance?).

PR #67720, opened 2023-08-17
Finalise vanilla gutter unification

Procyonae:
#### Summary
None

#### Purpose of change

~~Finalise gutter unification :D~~
Finalise vanilla gutter unification, doesn't remove the alias until I've dealt with mods

#### Describe the solution

Includes a couple of minor fixes to the mall while I were at it, I want to move the roof storage and delivery parking to a basement in a separate PR so I haven't nulled + migrated the empty z-1 maps
~~Removes NoHopes mall variants bc they're not being maintained inline with the vanilla map and it's way too much to go through when NoHope barely changes the maps~~ Separates the NoHope mall and school palettes for now instead 

#### Additional context

~~Haven't touched mods yet, opening rather than drafting so the tests catch any I miss~~
Holy line count nvm I'll do mods separately


Procyonae:
*that line diff is on just the second floor

PR #67723, opened 2023-08-17
Adds priority flags for special placement

Procyonae:
#### Summary
None

#### Purpose of change

Not having priorities is limiting

#### Describe the solution

Adds 2 new special flags PRIORITY_HIGH and PRIORITY_LOW which result in attempting to place the special before/after other specials
Isn't really ideal because it still checks all mandatory special occurrences first which I think means UNIQUE and GLOBALLY_UNIQUE only try after all of the mandatory occurrences
Applies the flags to some maps that probably benefit from it including most `"existing": true` road connections (which should hopefully help with clumping when they get shuffled to the front when there aren't many roads), removes the ability for isherwood farm to spawn over roads in exchange for giving it high priority

Making priority a numerical field

#### Testing

Idk how to test this if someone could suggest a proper test or a way to test locally that'd be great

### DIFF
--- a/data/json/overmap/overmap_special/specials.json
+++ b/data/json/overmap/overmap_special/specials.json
@@ -16,7 +16,7 @@
     "locations": [ "land" ],
     "city_distance": [ 8, 30 ],
     "occurrences": [ 0, 2 ],
-    "flags": [ "CLASSIC", "MAN_MADE" ]
+    "flags": [ "CLASSIC", "MAN_MADE", "PRIORITY_LOW" ]
   },
   {
     "type": "overmap_special",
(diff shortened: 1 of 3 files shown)


Maleclypse:
Hás conflicts to resolve.

lispcoc:
I think this is an important feature to prevent refugee centers from being generated too far away, but are there any plans to resume development?
If there is no one else to start, I am thinking of taking over this.

Maleclypse:
> I think this is an important feature to prevent refugee centers from being generated too far away, but are there any plans to resume development? If there is no one else to start, I am thinking of taking over this.

I believe the person who was working on this has experienced a life event such as a move and it is unknown when they will return.   I think if they don't respond today you can probably work on it.

lispcoc:
I got it. If I don't get a response today, I'll try to reimplement it.

Procyonae:
lispcoc is implementing this in #68389 instead

PR #66123, opened 2023-06-10
Standardise Special Terrain

Procyonae:
#### Summary
Performance "Removes the ludicrous amount of OMs the refugee and research centres define"

#### Purpose of change

Make the aforementioned maps less ridiculous and cut line count.
Fixes #66008

#### Describe the solution

All filler maps that solely exist afaik to stop other maps and map specials spawning on them have been merged into a new file with one OM for each terrain type that is just predecessor_terrain ed to what it's would be. The current identical maps are then replaced 1:1 with these currently.
The 1575 "unique" OMs that the refugee centre defines have been reduced down to 48 if I can count.
I also made the OMTs for the refugee centre different to mark the outer fence and differ the terrain from regular field.
The refugee centre stuff was added in #31221 and while I agree with preventing hostiles on the z0 level and like the sound of proposed additions manually adding in individually named tiles as wanted isn't a big deal and given that noone's made use of it that doesn't seem like good reasoning to keep it around.

#### Describe alternatives you've considered

The refugee centre claims a 15x15x7 (Z+2 to Z-4) cuboid of space for itself which seems completely unnecessary, we have no floating maps for starters and outside of slowdown if a lab spawns directly under the actual building claiming the underground terrain seems overkill too.
The upper levels of the research centre have several maps that could be replaced by special_air but that would require changing loads of nest coords and I'm lazy. 

#### Testing

Found the maps and checked everything is still kosher. I've not changed any of the actual building's ids or maps and I checked the stairs for the research centre still line up.

#### Additional context

image

Screenie just to show the OMT symbol changes.

PR #67726, opened 2023-08-17
Delete unnecessary om placements from refugee centre

Procyonae:
#### Summary
None

#### Purpose of change

Scrolling is pain

#### Describe the solution

Doubles down on #66123 and removes all special_air because it doesn't do anything and deletes all special_earth, special_rock and special_deep_rock that aren't in reality bubble of the building
image
!s are the special terrain

Removing all the underground special terrain in exchange for labs being able to spawn under it

#### Testing

Checked I got all the right oms by changing the special terrain symbol as shown above

#### Additional context

(The additions are just organising the order a lil)


I-am-Erk:
These are present for expansion of the building into the deeper layers, and to prevent labs from going there. You can remove a bit of the air, but it's fairly pointless to remove the earth part

Zireael07:
@I-am-Erk Worth a comment in the code so that the next contributor doesn't try the same sort of "clean-up" again

I-am-Erk:
Or worth me doing some mapping at last

Procyonae:
> These are present for expansion of the building into the deeper layers, and to prevent labs from going there. You can remove a bit of the air, but it's fairly pointless to remove the earth part

"Present for expansion" is a nonsense reason, only new maps would gen with the basement when it gets added anyway
I left the ground part within reality bubble range in specifically so labs can't spawn down there
I disagree that removing 1145 pointless om placements is in itself pointless

I-am-Erk:
Oh... I see. Someone "simplified" and changed these all to have a generic name.

These were originally given specific map names so that update mapgen could be run to spawn new structures within the area. That hasn't happened yet obviously, but that's the intent. They make no sense with the generic names, I agree... I would prefer they be reverted to having specific names so we have that option without rebuilding the world, but I don't intend to get around to using them any time immediately either so I'm not going to argue at length about it. If you really want you can delete it and I'll replace them when I get the bunker added in.

PR #67736, opened 2023-08-17
[Doesn't work, pls help] Json connection origins

Procyonae:
#### Summary
None

#### Purpose of change
Unhardcoding good
~~The hardcoded function needs changing for sub_connector unhardcoding~~ Makes #68496 alot less wonky
Allows for adding custom connections like a new `rural_road` connection that paths to the nearest road.
Allows separating z-2 and z-4 subway into different connections, with the z-4 pathing to ramps and z-2 pathing to surface subway stations, with fallbacks to join to the closest subway if none found.
Potentially could allow streams to path to rivers if reworked into a connection.

#### Describe the solution
Connection targets can be defined in JSON in their respective `overmap_connection` object under a new "origin" field
Accepts `om_terrain`s optionally with accompanying match types (defaults to type), city true/false (defaults to false) and a max distance that specifies how far it should look for each origin point. Atm it just picks the closest one it finds.
If it fails to find any of the specified origins or there aren't any it tries to find the nearest connection subtype terrain within 50 tiles and joins to that.
(I presume I need to make it select an appropriate adjacent tile to the target but I just want it to work first, it'll need #67723 to work reliably with specials anyway)

Ideally boundary joins would be selectable too (and jsonified themselves)
Ideally you'd be able to join all of one type together to ensure full connectivity where desired (like cities join rn)

#### Testing
Currently doesn't work, adding `"origin": [ { "city": true } ]` to `local_road` loads but doesn't function properly and anything else (eg `{ "om_terrain": "forest_thick" }`) hangs the game when building a world.

### DIFF
--- a/src/overmap.cpp
+++ b/src/overmap.cpp
@@ -1399,17 +1399,18 @@ struct fixed_overmap_special_data : overmap_special_data {
                 }
 
                 const overmap_connection &connection = *elem.connection;
-                cata::flat_set<std::pair<std::pair<std::string, ot_match_type>, int>> origin_terrains = connection.origin_terrains;
+                cata::flat_set<std::pair<std::pair<std::string, ot_match_type>, int>> origin_terrains =
+                            connection.origin_terrains;
                 point_om_omt connection_point;
                 point_om_omt origin_connection_point;
                 point_om_omt city_connection_point;
                 int distance_to_origin = INT_MAX;
                 int distance_to_city = INT_MAX;
-                
-                auto find_city = [&] () {
+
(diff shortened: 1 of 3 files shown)


Maleclypse:
@jbytheway this seems like something you would probably understand better than most others. Sorry if I’m incorrect.

jbytheway:
Yes, I would like to take a look at this, but I probably won't get the chance until next weekend.

@Procyonae it would be helpful if you could make one JSON change demonstrating the expected use of this new feature. Or maybe more than one change if necessary to demonstrate the different ways it can be used.

PR #67737, opened 2023-08-17
Fix bugs that cause eoc duplication

akrieger:
#### Summary
None

#### Purpose of change
A particular bug would cause certain eocs to multiply exponentially, like the portal storm stuck effect. When the eoc is non-recurring, an `erase` call was missing to remove it from the underlying storage list. When the `queued_eoc` instance with this list is assigned to another `queued_eoc` struct, the 'zombie' eoc in the list is re-added to the priority queue.

There was also a non-relevant bug where the object's existing queue was not cleared on assignment, which would cause segfaults if it was not empty, as the entries would point into the free'd storage.

#### Describe the solution
- Add the missing erase.
- Explicitly zero the queue on assignment.

#### Testing
Load a problematic save, see the eoc list appropriately shrinks when processing the stuck portal storm effect.

### DIFF
--- a/src/character.h
+++ b/src/character.h
@@ -325,6 +325,8 @@ struct queued_eocs {
 
     queued_eocs &operator=( const queued_eocs &rhs ) {
         list = rhs.list;
+        // Why doesn't std::priority_queue have a clear() function.
+        queue = {};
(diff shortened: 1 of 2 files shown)


anoobindisguise:
Nice, I thought I was going crazy!

PR #67742, opened 2023-08-18
No more sealed half-full cans

### LINKED ISSUE #67634: Spawns sealed food container with insufficient contents
### Describe the bug

In low "Item spawn scaling factor" world, spawns sealed food containers with insufficient contents such as one canned tomato.

### Attach save file

N/A

### Steps to reproduce

1. Create a world with low "Item spawn scaling factor". (e.g. 0.10)
2. Start the game in that world.
3. Explore some houses, find sealed food containers.

### Expected behavior

Either a sealed food containers with full contents is spawned or nothing is spawned.

### Screenshots

referenceImage

- OS: Windows
    - OS Version: 10.0.19045.3324 (22H2)
- Game Version: f2dedcb [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

Probably related to #60885.
I think this is because the result of '"charges": X' is either 0 or X, but the result of '"count": X' is spread from 0 to X.

SurFlurer:
#### Summary
none

#### Purpose of change
~~#67634~~
Just to make sure no container is sealed when it's not full.

#### Describe the solution
When checking container fullness in item::seal(), allow checking buckets, so spillable containers are not always considered full.

#### Testing
Actually there are no more sealed half-full cans. They are now unsealed.

#### Additional context

There are two problems that relate to this PR but remain unaddressed :

1. There are many cans that seems to be full, but actually cannot be sealed due to volume rounding issues with the contents, including player crafted ones ( see #71878 ).

2. There are containers that should be sealed but currently have insufficient content, needing itemgroup audits.

Problem 1 is making problem 2 harder to be distinguished and fixed, but problem 2 technically is more easy to fix.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -9436,7 +9436,8 @@ bool item::is_map() const
 
 bool item::seal()
 {
-    if( is_container_full() ) {
+    // Check the actual fullness of all kinds of containers.
+    if( is_container_full( /*allow_bucket = */ true ) ) {
         return contents.seal_all_pockets();
     } else {
         return false;
--- a/tests/unseal_and_spill_test.cpp
+++ b/tests/unseal_and_spill_test.cpp
@@ -339,6 +339,12 @@ void test_scenario::run()
                                     {}
                                 }
                             }
+                        },
+                        initialization {
+                            itype_test_solid_1ml,
+                            true,
+                            false,
+                            {}
                         }
                     }
                 }
@@ -591,6 +597,12 @@ void test_scenario::run()
                             false,
                             false,
                             {}
+                        },
+                        final_result {
+                            itype_test_solid_1ml,
+                            false,
+                            false,
+                            {}
                         }
                     }
                 };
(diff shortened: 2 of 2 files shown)


SurFlurer:
While playing with this I found some items that were supposed to be spawned or crafted sealed were not sealed anymore. This behaviour isn't introduced by my code but I'll try to expand this pr and fix them.

Maleclypse:
Is this ready for testing?

SurFlurer:
During the first few weeks of my inactivity ( and in recent days) I tried to address the first problem ( see the Additional Context of pr description ) locally but failed to think up a solution that is acceptable to myself. 

I decided to put an end to this. Honestly, successfully solving problem 1 can possibly make this pr unnecessary. So I'll be happy to close this if someone makes up their mind to fix problem 1.

Maleclypse:
> During the first few weeks of my inactivity ( and in recent days) I tried to address the first problem ( see the Additional Context of pr description ) locally but failed to think up a solution that is acceptable to myself.
> 
> I decided to put an end to this. Honestly, successfully solving problem 1 can possibly make this pr unnecessary. So I'll be happy to close this if someone makes up their mind to fix problem 1.

So am I understanding you that there are multiple itemgroups such as canned liver which cannot seal because the container for it required 1000ML but when you put as many livers in it as will fit you only get 970ML?    Not an accurate example but trying to understand what you are saying.

SurFlurer:
> So am I understanding you that there are multiple itemgroups such as canned liver which cannot seal because the container for it required 1000ML but when you put as many livers in it as will fit you only get 970ML? Not an accurate example but trying to understand what you are saying.

Yes this is what I mean. In fact a unit of canned liver is 62 ml, and you can only put 6 units in a 0.5L jar, resulting in a total of 496 ml which doesn't qualify for sealable in current code.

PR #67756, opened 2023-08-18
Added profession age bounds for randomly generated characters

### LINKED ISSUE #66521: Senior citizen age is randomized which can result in hilariously young characters.
### Describe the bug

This is probably a design problem rather than bug and it is pretty self-explanatory. Here in this screenshot, my senior citizen character is 18 years old. I don't know how retirement works abroad, but here you can only retire after age of 65 (this changes according to your job).

### Attach save file

None

### Steps to reproduce

I did randomized start (Play now!). So, I guess you need to choose particular scenario.

### Expected behavior

Age should be appropriately determined according to realistic standards.

### Screenshots

IMG_20230629_133004

I play on mobile and I use the experimental build from couple days ago.

### Additional context

_No response_

venvenvenvenvenvenven:
#### Summary
Bugfixes "Randomly generated characters are now aged appropriately to their profession"

#### Purpose of change
When starting a game through Play Now, I've noticed that in the some scenarios, characters have nonsensical age, such as Senior Citizens that are 30 years old.

#### Describe the solution
Profession JSON can now have lower and upper age bounds defined. These do not limit player choice (if you want to play a 16-year-old Major General, you still can), they only affect random generation. For most professions (where bounds aren't defined explicitly), the default lower bound is 21, and the upper bound is 55 (max age).

Tie this into profession groups, but this would probably bloat groups.
Restrict not just random gen, but also manual age input. I don't really want to limit player options with my first PR.
Raise maximum character age so that Senior Citizens could be older than 55. Out of scope of this PR.

#### Testing
Pressed Play Now a few times. Schoolkids generated with ages that are below 18, as they should. An Old Veteran generated with age 52, as she should have.

#### Additional context
I don't know much about US military forces, so age brackets could be defined for them in another PR by someone else.
Fixes #66521

### DIFF
--- a/src/profession.cpp
+++ b/src/profession.cpp
@@ -195,16 +195,8 @@ void profession::load( const JsonObject &jo, const std::string_view )
         _description_male = to_translation( "prof_desc_male", desc_male );
         _description_female = to_translation( "prof_desc_female", desc_female );
     }
-    if( jo.has_int( "age_lower" ) ) {
-        age_lower = jo.get_int( "age_lower" );
-    } else {
-        age_lower = 21;
-    }
-    if( jo.has_int( "age_upper" ) ) {
-        age_upper = jo.get_int( "age_upper" );
-    } else {
-        age_upper = 55;
-    }
+    age_lower = jo.get_int( "age_lower", 21 );
+    age_upper = jo.get_int( "age_upper", 55 );
 
     if( jo.has_string( "vehicle" ) ) {
         _starting_vehicle = vproto_id( jo.get_string( "vehicle" ) );
--- a/src/profession.cpp
+++ b/src/profession.cpp
@@ -195,6 +195,12 @@ void profession::load( const JsonObject &jo, const std::string_view )
         _description_male = to_translation( "prof_desc_male", desc_male );
         _description_female = to_translation( "prof_desc_female", desc_female );
     }
+    if(jo.has_int("age_lower")){
+        age_lower = jo.get_int("age_lower");
+    }else{age_lower = 21;}
+    if(jo.has_int("age_upper")){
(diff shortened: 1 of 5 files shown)


src/profession.cpp
```diff
@@ -195,6 +195,16 @@ void profession::load( const JsonObject &jo, const std::string_view )
...
+    if( jo.has_int( "age_lower" ) ) {
+        age_lower = jo.get_int( "age_lower" );
+    } else {
+        age_lower = 21;
+    }
+    if( jo.has_int( "age_upper" ) ) {
+        age_upper = jo.get_int( "age_upper" );
+    } else {
+        age_upper = 55;
+    }
```
irwiss, line 207:
```suggestion
    age_lower = jo.get_int( "age_lower", 21 );
    age_upper = jo.get_int( "age_upper", 55 );
```
`optional(...)` would also work

ZhilkinSerg, line 207:
```suggestion
    optional( jo, was_loaded, "age_lower", age_lower );
    optional( jo, was_loaded, "age_upper", age_upper );
```

src/profession.h
```diff
@@ -119,6 +119,8 @@ class profession
         std::vector<proficiency_id> proficiencies() const;
         StartingSkillList skills() const;
         const std::vector<mission_type_id> &missions() const;
+        int age_lower;
+        int age_upper;
```
ZhilkinSerg, line 123:
```suggestion
        int age_lower = 22;
        int age_upper = 55;
```

venvenvenvenvenvenven:
Thanks for the help, I'm a bit new so it's appreciated.

ilovehans10:
This should fix #66521

PR #67758, opened 2023-08-18
Add better infrastructure to support Aftershock Skullgun

anoobindisguise:
#### Summary
Content "Make the aftershock skullgun cbm cooler"

#### Purpose of change
I love the Skullgun CBM from Aftershock because it's just perfect, it's so unbelievably shitty that it's incredibly funny. At the same time the Aftershock CBM didn't work correctly - it did damage to your torso whenever you extended it, rather than doing damage to your head when fired like you'd expect. It also gave you your intelligence back when you uninstalled it which was sad. It also took up your hands when you wanted to fire it, when as a skull gun it should be hands free.

#### Describe the solution
- Skullgun CBM has been rechambered from 40 s&w to 10mm auto as that's a cooler more scifi pistol cartridge.
- It has a base dispersion of 855 and a damage malus of -6. The bionic encumbers your head by 5.
- I created a new JSON field in guns called "hurt_part_when_fired" which allows guns to deal damage to one or more body parts when you fire them. The skullgun does 10 damage to your head every time you shoot something with it.
- I created a new JSON flag for guns called "ALWAYS_AIMED". This causes a gun to always have 0 recoil when aiming, so that you don't need to take time to aim it as the gun starts off already as aimed as possible when you go to fire it. This is because the skullgun would logically already be aimed at whatever you're looking at without needing to lift your arms and point it at something while lining up the sights - though it has horrendous dispersion so it's already very inaccurate, "aimed" might be something of a misnomer here. 
- The skullgun can be fired while holding something else in your hands. It has the RELOAD_AND_SHOOT flag, however since it is fully aimed and can be fired instantly after being loaded, you pay the move cost to reload it only after you actually fire it, so it has the advantage of being able to be fired immediately as soon as you have a target.
- I added a new JSON entry to CBMs called "give_mut_on_removal" which lets you make bionics add mutations to your character when removed. I plan on using this for bionic limbs so that it can add a "no left arm" or "no eyes" mutation where appropriate. In the case of the skullgun it gives you a mutation which lowers your int by 4 and perc by 2, which is otherwise unpurifiable, and then removes that mutation if you install it for a second time, so you are permanently saddled with those stat penalties.

#### Testing
I have tested extensively and confirmed it works the way I want it to.

#### Additional context

skullgun1
skullgun2


kevingranade:
Nope.

anoobindisguise:
> Nope.

Can you elaborate? I've been workshopping this with Maleclypse pretty extensively and he had ok'd my process thus far. What's it missing exactly

Maleclypse:
Yes, I had felt like it seemed reasonable given that Erk has been aiming for weird nethereum driven CBMs and since he’s been pushing the nether as a transitory place that seems to be influenced by human thoughts and ideas having the Skullgun as a common trope would make it more likely to be one of the weird CBMs produced by nethereum exposure.

kevingranade:
Reskinned I could see something similar working,  but a gun you put in your skull is just too far, like where does the displaced brain matter go,  it's not exactly microscopic. It needs to be at least phtsically feasible even if it's not required to be a commercially useful product.

anoobindisguise:
> Reskinned I could see something similar working, but a gun you put in your skull is just too far, like where does the displaced brain matter go, it's not exactly microscopic. It needs to be at least phtsically feasible even if it's not required to be a commercially useful product.

The brain matter gets scooped out, never to return. That's why you lose as much as 50% of your total logical capacity and as much as 25% of your visual processing capacity. Presumably its inventor company had access to studies on what parts of the brain can be "safely" removed, and people IRL do sometimes live with pretty serious TBIs or lobotomies. It has an install difficulty of 16, which is the highest in the game (higher than even the cerebral booster cbm). 
These things said I could make it into a 22 pistol with a -2 damage mod (like the Ruger LCR) if you think a cut down 9mm pistol couldn't possibly fit even given a vastly shorter barrel and no grip or magwell.

Maleclypse:
Reopening to target towards AFS.  Thanks!

data/mods/Aftershock/player/bionics.json
```diff
@@ -358,14 +358,14 @@
...
     "encumbrance": [ [ "head", 5 ] ],
-    "act_cost": "35 J",
     "fake_weapon": "bio_skullgun_gun",
-    "flags": [ "BIONIC_TOGGLED", "BIONIC_WEAPON", "NO_UNWIELD" ],
+    "flags": [ "BIONIC_GUN" ],
     "stat_bonus": [ [ "INT", -4 ], [ "PER", -2 ] ],
-    "canceled_mutations": [ "INT_UP", "INT_UP_2", "INT_UP_3", "INT_UP_4", "INT_ALPHA" ],
-    "spell_on_activation": { "id": "skullgun_snapback", "hit_self": true }
+    "canceled_mutations": [ "INT_UP", "INT_UP_2", "INT_UP_3", "INT_UP_4", "INT_ALPHA", "SKULLGUN_STUPID" ],
+    "give_mut_on_removal": [ "SKULLGUN_STUPID" ],
```
Maleclypse, line 368: Can you add this to the documentation in bionic fields?

ilovehans10:
With the ability to add mutations like no arm when a bionic is removed, wouldn't it also make sense to produce human flesh/bone when installing bionics that remove limbs?

anoobindisguise:
> With the ability to add mutations like no arm when a bionic is removed, wouldn't it also make sense to produce human flesh/bone when installing bionics that remove limbs?

I want to do this at some point yes. Specifically for the cerebral booster, synthetic lungs and expanded digestive system cbms as they exist now.

PR #67768, opened 2023-08-19
Add dogs barks

marimarigi:
#### Summary
None

#### Purpose of change
I noticed that some of dogs have a "speech" in speech.json but neither PARROT nor PARROT_AT_DANGER.
Conversely, some others have PARROT_AT_DANGER but no "speech".
Add them to bark properly.

#### Describe the solution
Samoyed, mongrel and Loper add "speech".
For dogs without PARROT or PARROT_AT_DANGER, I think it's because they are special_attacks isn't "extend", so add it.

#### Testing
I spawned those dogs with zombies to fight them and confirm they barked.

### DIFF
--- a/data/json/speech.json
+++ b/data/json/speech.json
@@ -2528,19 +2528,7 @@
   },
   {
     "type": "speech",
-    "speaker": [ "mon_dog_samoyed", "mon_mi_go", "mon_mi_go_slaver", "mon_mi_go_myrmidon" ],
-    "sound": "BARK!",
-    "volume": 40
-  },
-  {
-    "type": "speech",
-    "speaker": [ "mon_dog_mutant_mongrel", "mon_mi_go", "mon_mi_go_slaver", "mon_mi_go_myrmidon" ],
-    "sound": "BARK!",
-    "volume": 40
-  },
-  {
-    "type": "speech",
-    "speaker": [ "mon_dog_mutant_loper", "mon_mi_go", "mon_mi_go_slaver", "mon_mi_go_myrmidon" ],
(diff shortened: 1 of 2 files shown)


Fris0uman:
Did you read this https://github.com/CleverRaven/Cataclysm-DDA/pull/32831
I think you might be trying to make dogs bark constantly which would be awfull

Karol1223:
Out of pure curiosity, not meant to complain about how you did this, but I do have to ask - is there any reason why we have multiple speeches with the exact same string and volume that only differ by what monsters use them? Can't all dogs that ``BARK!`` use one speech? I'm genuinely curious now, as I see that's been the case for all the other dogs prior to your changes as well and I'm lost as to what's the benefit.

marimarigi:
The dogs were already separated so I did the same, but I don't know if there is any benefit.

data/json/speech.json
```diff
@@ -2526,6 +2526,24 @@
...
+    "speaker": [ "mon_dog_mutant_mongrel", "mon_mi_go", "mon_mi_go_slaver", "mon_mi_go_myrmidon" ],
+    "sound": "BARK!",
+    "volume": 40
+  },
+  {
+    "type": "speech",
+    "speaker": [ "mon_dog_mutant_loper", "mon_mi_go", "mon_mi_go_slaver", "mon_mi_go_myrmidon" ],
+    "sound": "BARK!",
+    "volume": 40
+  },
```
MNG-cataclysm, line 2546:
```suggestion
  {
    "type": "speech",
    "speaker": [ "mon_dog_samoyed", "mon_dog_mutant_mongrel", "mon_dog_mutant_loper", "mon_mi_go", "mon_mi_go_slaver", "mon_mi_go_myrmidon" ],
    "sound": "BARK!",
    "volume": 40
  },
```
Why not consolidate all of these into a single brick of code? That'd cut down on space and make things easier.

MNG-cataclysm (approved):
You've done a great job with this. Thank you.

PR #67771, opened 2023-08-19
Add a recipe filter for activity level

mqrause:
#### Summary
Features "Add a recipe filter for activity level"

#### Purpose of change

I mostly want to filter for the none activity level to find crafts for when I'm too exhausted for everything else.

#### Describe the solution

Add a new filter.

#### Testing

Tried some searches and results looked correct.

### DIFF
--- a/src/crafting_gui.cpp
+++ b/src/crafting_gui.cpp
@@ -941,6 +941,10 @@ static recipe_subset filter_recipes( const recipe_subset &available_recipes,
                     break;
                 }
 
+                case 'a':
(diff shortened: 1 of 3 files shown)


Nebnis:
If you don't mind me asking, i would love a filter for armor by layers(outer,normal,inner, etc)

mqrause:
> If you don't mind me asking, i would love a filter for armor by layers(outer,normal,inner, etc)

I'm not interested in adding that right now, but it'd be mostly the same as I'm doing here, so it's an easy C++ PR.

PR #67788, opened 2023-08-20
explain what the starting dates do

ghost:
#### Summary
None

#### Purpose of change
i guess its an oversight, but starting date of cataclysm & game are not explained, one of them (start of game) can alter the game's difficulty dramatically 

#### Describe the solution
see commit

#### Testing
looked alright when i loaded in new character

#### Additional context
as far as my testing went, no rotten food spawned when i changed the years in both cataclysm & game dates, only evolution changed after changing game date
SIDENOTE: setting starting game date at 35 years makes the game extremely laggy when you move.
SIDENOTE2: changing the start of cataclysm also changes start of game. seems to be only a visual bug, because evolution does not advance when you do it. even if you go ahead and try to change it manually as well. i do not know how to fix that sorry.

### DIFF
--- a/src/newcharacter.cpp
+++ b/src/newcharacter.cpp
@@ -3108,10 +3108,12 @@ static std::string assemble_scenario_details( const avatar &u, const input_conte
         assembled += current_scenario->vehicle()->name + "\n";
     }
 
-    assembled += "\n" + colorize( _( "Start of cataclysm:" ), COL_HEADER ) + "\n";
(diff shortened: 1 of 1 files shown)


Devengel:
Food rot does not change at all from what i saw, evolution changes as intended. but there is a number of visual bugs that need to be fixed, as much as i want to fix them i don't know how, the reason i did this C++ is because i copied what was written and hoped it'd work. i hope the information i found is helpful to whoever finds this closed PR.

Devengel:
see #67790

PR #67789, opened 2023-08-20
Insecticide audit

Karol1223:
#### Summary
None

#### Purpose of change
image
My computer froze for 10 fucking minutes trying to handle this

Insecticide was a comestible, so after the comestible charge removal it fell apart. Why was it comestible? Well... because there wasn't a reason for it not to be, I suppose? It was ``INVALID``, so inedible, and it had no use action. It also was 1g per item because... 
image
Yeah. This wasn't a problem before the charge removal, now it is...

#### Describe the solution
- change insecticide to ``type: GENERIC``, make it 50g per item instead of 1g
- audit the recipes to use ``result_mult`` appropriately and take it in batches of 50g instead of 50x50g
- audit the itemgroups so that... well, first of all they work, because from what I saw most were broken. But also to spawn 450g (~1lb) per instance. Was it necessary? No, but 400g was a weird number. Amazon showed me results sold in multiples of pounds, so I went with it

...keep it comestible and add a way for insect mutants to poison themselves? No, thanks. I'm good.

#### Additional context
This means all stashes of insecticide on existing saves will expand 50x. I can't change that, I'm afraid. On my main save I'm gonna have over 1 ton of insecticide, now... Save me from this...

### DIFF
--- a/data/json/itemgroups/Agriculture_Forage_Excavation/agriculture.json
+++ b/data/json/itemgroups/Agriculture_Forage_Excavation/agriculture.json
@@ -54,7 +54,7 @@
       [ "fertilizer_liquid", 20 ],
       [ "fertilizer_commercial", 30 ],
       [ "fungicide", 20 ],
-      { "item": "insecticide", "prob": 30, "count": 9 },
+      { "group": "insecticide_bag_plastic", "prob": 30 },
       { "item": "chem_saltpetre", "prob": 15, "charges-min": 10, "charges-max": 50 },
       [ "sickle", 40 ],
       { "group": "tools_toolbox", "prob": 10 },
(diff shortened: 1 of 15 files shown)


mqrause:
> * audit the recipes to use `result_mult` appropriately

Just an FYI, `charges` and `result_mult` in recipes are identical for non-comestibles as long as no containers are involved.

Karol1223:
> Just an FYI, charges and result_mult in recipes are identical for non-comestibles as long as no containers are involved.

Oh huh. Didn't know that. I could make it use charges again but I think it's not necessary now? I'd have to edit the numbers to 8 again anyway

mqrause:
Yeah, it doesn't matter either way, so just leave it as is.

PR #67791, opened 2023-08-20
Code optimizations reported by static code analysis (2023-08-20)

ZhilkinSerg:
#### Summary

Performance "Code optimizations reported by static code analysis"

#### Purpose of change

Cleanup code while increasing performance and fixing minor errors.

#### Describe the solution

See commit messages.


alp7292:
very good. how much of an improvement is this?

src/activity_handlers.cpp
```diff
@@ -3033,7 +3033,7 @@ void activity_handlers::operation_do_turn( player_activity *act, Character *you
             }
             if( !bps.empty() ) {
                 for( const bodypart_id &bp : bps ) {
-                    you->add_effect( effect_bleed,  1_minutes * difficulty, bp, false, true );
```
akrieger, line 3036: This seems to have been a wrong refactor in https://github.com/CleverRaven/Cataclysm-DDA/commit/50bba798fea84be2fe5ff3baea2ad9c223410e7c#diff-5ee2564389e1b0f512bd67486baff7a9488627beef1efe92530b6926449a6ee9.

It used to be `you->make_bleed( effect_source::empty(), bp, 1_turns, difficulty, true );` which was a call to (afaict)
```
        virtual void make_bleed( const effect_source &source, const bodypart_id &bp, time_duration duration,
                                 int intensity = 1, bool permanent = false, bool force = false, bool defferred = false ) = 0;
```

so the trailing `true` was 'permanent'.

The `add_effect` signature is
```
        void add_effect( const efftype_id &eff_id, const time_duration &dur, bodypart_id bp,
                         bool permanent = false, int intensity = 0, bool force = false, bool deferred = false ) {
```

It was a behavioral change to make the duration dependent on difficulty instead of the intensity.

akrieger (approved):
lgtm except the surgery bleed effect. tidy error is preexisting.

src/activity_handlers.cpp
ZhilkinSerg: Changed it to:
`you->add_effect( effect_bleed,  1_minutes * difficulty, bp, true, 1 );`
and
`you->add_effect( effect_bleed,  1_minutes * difficulty, bodypart_str_id::NULL_ID(), true, 1 );`

PR #67796, opened 2023-08-20
Folding Seat belts and tweaks

gadgetpatch:
#### Summary
Features "Added BELTABLE to folding seat, tweaked folding seat and saddle properties, and adjusted related factory unfoldables."

#### Purpose of change
A minor balance and sanity check for Folding Seats and Saddles:
1. There were no foldable vehicle seats that could attach a seatbelt, despite seatbelts having a folded volume. Folding wheelchairs often have attached belts or harnesses for securing infirm patients, so the Folding Seat seemed a logical place to let players attach them. 

2. Folding Seats are built of just a single sheet, but their volume was anomalously large at 2.5L. That's 5x the volume of the sheet, or the same as adding an entirely new folding frame tile to the vehicle, despite just being loose, collapsible fabric.

3. Folding Seats were somehow less comfortable to sleep on than a bicycle saddle, at Comfort 0. The same as trying to sleep while shackled to a steel prison bench. This is all very silly.

4. "Folding Seat" is part of the "Cloth [part]" series of folding vehicle parts, made of single fabric sheets attached to the frame, but it uses a unique naming convention. This may confuse players when browsing through folding cloth parts.

5. Bike Saddles inherited their warmth from car seats, despite only insulating a rider at their butt and upper legs.

#### Describe the solution
1) Added BELTABLE to Folding Seat (and removed NONBELTABLE).
2) Reduced folded_volume of Folding Seat to 1L (formerly 2.5L)
3) Changed in-game name to "Cloth Seat", and updated description.
4) Folding Seat now inherits Comfort of 1 from Saddle (formerly 0).
5) Reduced warmth of Saddle to 50 (formally inherited 200 from base car seat)

6) UPDATE: Adjusted the factory unfoldables for the wheelchair and inflatable boat to match new folded volume.

Making the folding seat just the size of the additional sheet.
Making saddles an even worse place to sleep.
A proper audit of the folded_volume of vehicle parts, which takes materials, flexibility, how they're mounted or folded, and other mechanical considerations into account.

#### Testing
Spawned a mountain bike and folding wheelchair. Successfully attached a seatbelt and 5-point harness to the wheelchair seat, and was able to sleep in either when sufficiently tired. New name and description read as expected, and the new folded volume was reflected.

#### Additional context
folding_seatbelts1


Maleclypse:
You'll have to adjust the folding vehicle tests it looks like.

PR #67801, opened 2023-08-21
[Magiclysm] Add Whisper of the Leaves druid spell

Standing-Storm:
#### Summary
Mods "Add Whisper of the Leaves spell"

#### Purpose of change

I'm pretty focused on MoM at the moment and it's going to be a long time before my magical innawood mod ever gets finished, so I might as well add some of the stuff I worked on to Magiclysm.

#### Describe the solution

Adds the Whisper of the Leaves spell.  It temporarily gives you the Mycorrhizal Communion mutation (using EoC) so you can talk to the trees.  The actual spell has a very short duration, but that doesn't matter--having tested it, an EoC taking away your Mycorrhizal Communion mutation during the process of communing doesn't cause any problems.

If you already have the Mycorrhizal Communion mutation, you get a message about how the spell won't make you any more in tune with the forest than you already are.

#### Testing

It works

#### Additional context

This should be merged before https://github.com/CleverRaven/Cataclysm-DDA/pull/67814


ashGlaw:
Cool spell. Good idea. And great idea testing what happens when you lose the mutation while communing 👍

Maleclypse:
Needs a rebase.

PR #67808, opened 2023-08-21
[MoM] Adjust stamina costs

Standing-Storm:
#### Summary
Mods "Adjust MoM stamina costs"

#### Purpose of change

Several of the Stamina costs were balanced early in the process, when I was assuming that the base of 8500 Stamina could increase maybe 50% through mutation and before I knew that through cardio it could almost double, not even counting traits.

#### Describe the solution

Increase stamina costs in many cases.  Sometimes the increase is modest, sometimes it's much larger  on very low level powers that had an extremely low cost on the high end, like Fountain of Flames (previously 300 at max level, now 600) or Blink (previously 50 at max level, now 250).  In many cases the cost reduction per power level was also slowed.

Drain is unaffected.

Allowing the Warp to overtake you.  It is a good pain.

#### Testing

Numbers work.


Dr-Expendable:
I was actually about to debug remove Fountain of Flames from new games simply because it was so efficient that it was obsoleting all other forms of combat, so this is very welcome.

Standing-Storm:
For obvious reasons most of my testing is on new characters in new worlds, but I debugged in pyrokinesis onto my long-time character (with huge cardio) and noticed that in most cases it was more efficient to just Fountain of Flames everyone individually unless a horde was truly overwhelming. Pyrokinesis is supposed to be primarily destructive, but it should require a range of powers to be so.

PR #67815, opened 2023-08-21
Fix compilation error on GCC 7 and 8

sethsimon:
#### Summary
Build "Fix compilation error on GCC 7 and 8"

#### Purpose of change

It's been quite awhile since I've compiled Cataclysm, and the other day I received a bunch of compilation errors with g++ 8.3.0 running under Debian GNU/Linux (64 bit). The errors are due to fancy <filesystem> stuff in src/third-party/ghc, and I got it to compile by adding `-lstdc++fs`. You might want to reference this [interesting discussion](https://github.com/avast/retdec/issues/842) involving the same issue. Below is the tail end of the compilation errors; the entire thing is 235 lines.

```
/usr/bin/ld: obj/tgz_archiver.o: in function `tgz_archiver::_gen_tar_header(std::filesystem::__cxx11::path const&, std::filesystem::__cxx11::path const&, std::filesystem::__cxx11::path const&, long)':
tgz_archiver.cpp:(.text+0xda): undefined reference to `std::filesystem::status(std::filesystem::__cxx11::path const&)'
/usr/bin/ld: tgz_archiver.cpp:(.text+0xe8): undefined reference to `std::filesystem::status(std::filesystem::__cxx11::path const&)'
/usr/bin/ld: tgz_archiver.cpp:(.text+0x108): undefined reference to `std::filesystem::symlink_status(std::filesystem::__cxx11::path const&)'
/usr/bin/ld: tgz_archiver.cpp:(.text+0x114): undefined reference to `std::filesystem::last_write_time(std::filesystem::__cxx11::path const&)'
/usr/bin/ld: tgz_archiver.cpp:(.text+0x134): undefined reference to `std::filesystem::status(std::filesystem::__cxx11::path const&)'
/usr/bin/ld: obj/tgz_archiver.o: in function `tgz_archiver::add_file(std::filesystem::__cxx11::path const&, std::filesystem::__cxx11::path const&)':
tgz_archiver.cpp:(.text+0x6a2): undefined reference to `std::filesystem::__cxx11::path::has_root_directory() const'
/usr/bin/ld: tgz_archiver.cpp:(.text+0x6d9): undefined reference to `std::filesystem::__cxx11::path::has_filename() const'
/usr/bin/ld: tgz_archiver.cpp:(.text+0x711): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
/usr/bin/ld: tgz_archiver.cpp:(.text+0x76a): undefined reference to `std::filesystem::__cxx11::path::lexically_relative(std::filesystem::__cxx11::path const&) const'
/usr/bin/ld: tgz_archiver.cpp:(.text+0x832): undefined reference to `std::filesystem::__cxx11::path::_M_split_cmpts()'
/usr/bin/ld: tgz_archiver.cpp:(.text+0x887): undefined reference to `std::filesystem::status(std::filesystem::__cxx11::path const&)'
/usr/bin/ld: tgz_archiver.cpp:(.text+0x94e): undefined reference to `std::filesystem::status(std::filesystem::__cxx11::path const&)'
/usr/bin/ld: obj/translation_document.o: in function `TranslationDocument::TranslationDocument(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
translation_document.cpp:(.text+0x6d1): undefined reference to `std::filesystem::file_size(std::filesystem::__cxx11::path const&)'
collect2: error: ld returned 1 exit status
make: *** [Makefile:956: cataclysm] Error 1
```

#### Describe the solution

I've modified the makefile to do a couple of minimal test compiles that use a `<filesystem>` function; if the one without the `-lstdc++fs` flag works (which I suppose would be the case for everyone but me), then all is well. If it fails, but adding the flag fixes it, then `LDFLAGS` is modified accordingly. On the other hand, if both of them fail, it will err. 

Abandoning support for GCC 7 and 8 is an option. Another option would be to check for g++ version 7 or 8 instead of checking via a test compile, but I think this way is cleaner.

#### Testing

It took some experimentation to get `FS_TEST_FILE` right; I can't say I've ever had to escape an `#` before since quotes usually do the job. I also examined the stdout and stderr of the test compilation programs to make sure that the first compilation was failing for the right reasons. And, of course, I compiled the game to make sure I didn't get any other errors or anything. I **haven't** tested under any other operating systems or with any other compilers.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -505,6 +505,15 @@ ifeq ($(PCH), 1)
   endif
 endif
 
+# Workaround for GCC 7 and 8.
+FS_TEST_FILE := $(shell printf "\#include <filesystem>\\\n int main() { std::filesystem::path p; p.parent_path(); }")
+ifeq   ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 -            >/dev/null 2>&1 || echo fail),fail)
+  ifeq ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 - -lstdc++fs >/dev/null 2>&1 || echo fail),fail)
+    $(error "Unable to compile test filesystem program")
+  endif
+  LDFLAGS += -lstdc++fs
+endif
+
 CPPFLAGS += -Isrc -isystem ${SRC_DIR}/third-party
 CXXFLAGS += $(WARNINGS) $(DEBUG) $(DEBUGSYMS) $(PROFILE) $(OTHERS)
 TOOL_CXXFLAGS = -DCATA_IN_TOOL
--- a/Makefile
+++ b/Makefile
@@ -507,8 +507,11 @@ endif
 
 # Workaround for GCC 7 and 8.
 FS_TEST_FILE := $(shell printf "\#include <filesystem>\\\n int main() { std::filesystem::path p; p.parent_path(); }")
-ifeq   ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 -            >/dev/null 2>&1 || echo fail),fail)
-  ifeq ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 - -lstdc++fs >/dev/null 2>&1 || echo fail),fail)
+$(shell echo "Debug: CXX='$(CXX)'" 1>&2)
+#ifeq   ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 -            >/dev/null 2>&1 || echo fail),fail)
+  #ifeq ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 - -lstdc++fs >/dev/null 2>&1 || echo fail),fail)
+ifeq   ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 -            || echo fail),fail)
+  ifeq ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 - -lstdc++fs || echo fail),fail)
     $(error "Unable to compile test filesystem program")
   endif
   LDFLAGS += -lstdc++fs
--- a/Makefile
+++ b/Makefile
@@ -506,8 +506,9 @@ ifeq ($(PCH), 1)
 endif
 
 # Workaround for GCC 7 and 8.
-FS_TEST_FILE := $(shell printf "\#include <filesystem>\\\n int main() { std::filesystem::path p; p.parent_path(); }")
+FS_TEST_FILE := $$(printf "\#include <filesystem>\\\n int main() { std::filesystem::path p; p.parent_path(); }")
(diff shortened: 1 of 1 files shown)


irwiss:
fwiw gcc 7 and 8 are effectively abandoned https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/COMPILING/COMPILER_SUPPORT.md

Not sure what kind of debian you're running but afaik even oldstable comes with gcc 10

akrieger:
If you can make the change in a way that doesn't inline invoke a compiler on every make invocation, which works on every platform (the makefile works for windows, mac, and linux, in both msys and mingw environments on Windows), then sure. But as per the compiler support doc irwiss linked, gcc 7 and 8 are EOL for this project and the desire to include workarounds is functionally zero.

sethsimon:
Interesting, I wasn't aware that it's not officially supported anymore. This version of gcc has always felt modern to me, and I'm surprised to find out that it's considered outdated. If I recall correctly, it's what came with Debian Buster (oldoldstable) when I installed it a few years ago, and since I've never run into anything that complained about it, I never bothered updating it. Personally I don't see the problem with using old compilers forever, and I have no plans to upgrade since it still works fine as long as I add that `-lstdc++fs` flag, but I respect the project's stance.

Anyway, if the two test compilations are objectionable (at 0.75 seconds each, I'd tend to agree), I could check for it with something like `$(CXX) --version | fgrep g++ | grep ' [78]'` (0.006s). The full output of `g++ --version` is
```
g++ (Debian 8.3.0-6) 8.3.0
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

Or better yet, since the makefile is already well over 1000 lines, then in the interest of simplicity, I could avoid adding even more to it, especially for a case that's not even supported.

Makefile
```diff
@@ -505,6 +505,19 @@ ifeq ($(PCH), 1)
...
+FS_TEST_FILE := $$(printf "\#include <filesystem>\\\n int main() { std::filesystem::path p; p.parent_path(); }")
+$(shell echo "Debug: CXX='$(CXX)'" 1>&2)
+$(shell echo "Debug: FS_TEST_FILE='$(FS_TEST_FILE)'" 1>&2)
+#ifeq   ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 -            >/dev/null 2>&1 || echo fail),fail)
+  #ifeq ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 - -lstdc++fs >/dev/null 2>&1 || echo fail),fail)
+ifeq   ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 -            || echo fail),fail)
+  ifeq ($(shell echo "$(FS_TEST_FILE)" | $(CXX) -o /dev/null -x c++ -std=c++17 - -lstdc++fs || echo fail),fail)
+    $(error "Unable to compile test filesystem program")
+  endif
+  LDFLAGS += -lstdc++fs
```
Qrox, line 518: Maybe you can simply specify this flag when invoking `make` in your local workflow.

sethsimon:
Agreed, that's what I would've done in the first place if I'd read the compiler support document.

PR #67826, opened 2023-08-22
[innawood] add alternative recipe to office scissors

ghost:
#### Summary
None

#### Purpose of change
current recipe for office scissors is difficult, requiring plastic chunks which is hard to make 

#### Describe the solution
adds an alternative recipe, one that uses wood for the handle, and add carving as a proficiency in recipe

#### Testing
opened the game and the recipe works as intended


Light-Wave:
Good change, although I think you can put this in mainline instead.

Karol1223:
I wouldn't mainline this because you're effectively swapping materials around. The resulting item will still have plastic as its material, despite being crafted with wood

Light-Wave:
> I wouldn't mainline this because you're effectively swapping materials around. The resulting item will still have plastic as its material, despite being crafted with wood

That is fair. Yeah, keep it in Innawood then.

PR #67835, opened 2023-08-23
Cache results of item search functions

Kamayana:
#### Summary
Performance "Cache results of item search functions"

#### Purpose of change
One of the larger CPU hogs when moving around with extremely large inventories is `all_items_with_flag` and similar inventory searching functions. These iterate through every single item in the character's inventory to find specific ones. These functions are called very often, and sometimes for fairly minor reasons such as looking for a watch to determine if the sidebar should show the exact time or the sundial. As these results aren't cached, every function like this will do their own iteration through the entire inventory every turn - some even do this every frame.

#### Describe the solution
Adds `inv_search_cache` to the character class. This map is used to cache the results of a new suite of functions that search through the inventory for certain items: `cache_visit_items_with` for applying code to the filtered items, `cache_has_item_with` for getting a boolean result, and `cache_get_items_with` for creating a vector. These functions can filter items based on having a specified item type, having a specified flag on their type, having a true result from a specified simple boolean item function, or a combination of the 3. All of these results are cached, making it extremely fast to request the same filters in the future. In addition to the filter parameters, the functions also accept a lambda function parameter that can be run on all the filtered items to further filter the results. Though the lambda's results are not cached, the preceding cached filters can help narrow things down so the lambda will only have to process a fraction of a character's items rather than all of them. The non-cache-using `has_item_with_flag` function is still around, as there are item instance flags like `WET` or `ITEM_BROKEN` that shouldn't be cached.

I reworked everything from my first attempt based on feedback received, so there was already an alternative I considered so much it became the real thing.

The UI functions that call item searches every draw frame seem like they could be optimized further - for example, the game is updating the lightmap every frame, and that update includes searching the player's inventory for their most luminous item. In practice this seems unnecessary, as even if the player drops said item they'll keep their same "lit_up" buff until the game turn ticks over to the next. It might work to only update the character's maximum inventory luminosity every turn instead, rather than every frame. However... the cached item results work well enough already that their CPU impact is almost entirely reduced, so I deemed it unnecessary.

#### Testing
Added and removed inventory items in all available ways to ensure they didn't crash. Loaded a worn pocket universe with 10,000 caffeine pills, 10,000 bottles of water, and 10,000 wrist watches, and noted that the inv_search_cache functions had little notable CPU impact. Ran the full test suite locally and got all green results.

#### Additional context
- [x] Rework `items_with` to also work with the cache.
- [x] Use said rework for lights instead of the special case it's using now.
- [x] Implement it for other uses of `items_with`.

I didn't actually end up using multiple filter parameters at once in the new functions, but I feel like they could be useful somewhere so I kept them.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -2938,7 +2938,7 @@ std::list<item *> Character::get_dependent_worn_items( const item &it )
 item Character::remove_weapon()
 {
     item tmp = weapon;
-    remove_from_flagged_item_cache( weapon );
+    remove_from_inv_search_cache( weapon );
     weapon = item();
     get_event_bus().send<event_type::character_wields_item>( getID(), weapon.typeId() );
     cached_info.erase( "weapon_value" );
@@ -8895,10 +8895,11 @@ bool Character::has_item_with_flag( const flag_id &flag, bool need_charges ) con
 
 std::set<item *> &Character::all_items_with_flag( const flag_id &flag ) const
 {
-    auto iter = flagged_item_cache.find( flag );
+    std::string flag_string = flag.c_str();
+    auto iter = inv_search_cache.find( flag_string );
 
     // If the flag set already exists in the cache, return it.
-    if( iter != flagged_item_cache.end() ) {
+    if( iter != inv_search_cache.end() ) {
         return iter->second;
     }
 
@@ -8914,31 +8915,31 @@ std::set<item *> &Character::all_items_with_flag( const flag_id &flag ) const
         } );
     }
     for( const item *it : items_with_flag ) {
-        flagged_item_cache[flag].insert( const_cast<item *>( it ) );
+        inv_search_cache[flag_string].insert( const_cast<item *>( it ) );
     }
-    return flagged_item_cache[flag];
+    return inv_search_cache[flag_string];
 }
 
-void Character::add_to_flagged_item_cache( item &it ) const
+void Character::add_to_inv_search_cache( item &it ) const
 {
-    for( auto &flagged_items : flagged_item_cache ) {
-        if( flagged_items.first == flag_LIGHT ) {
+    for( auto &cached_items : inv_search_cache ) {
+        if( cached_items.first == flag_LIGHT.c_str() ) {
             if( it.type->light_emission > 0 ) {
-                flagged_items.second.insert( &it );
+                cached_items.second.insert( &it );
             }
-        } else if( it.has_flag( flagged_items.first ) ) {
-            flagged_items.second.insert( &it );
+        } else if( it.has_flag( flag_id( cached_items.first ) ) ) {
+            cached_items.second.insert( &it );
         }
     }
 }
 
-void Character::remove_from_flagged_item_cache( item &it ) const
+void Character::remove_from_inv_search_cache( item &it ) const
 {
-    if( flagged_item_cache.empty() ) {
+    if( inv_search_cache.empty() ) {
         return;
     }
-    for( auto &flagged_items : flagged_item_cache ) {
-        flagged_items.second.erase( &it );
+    for( auto &cached_items : inv_search_cache ) {
+        cached_items.second.erase( &it );
     }
 }
 
@@ -9321,7 +9322,7 @@ void Character::on_item_acquire( const item &it )
     bool update_overmap_seen = false;
 
     it.visit_items( [this, &check_for_zoom, &update_overmap_seen]( item * cont_it, item * ) {
-        add_to_flagged_item_cache( *cont_it );
+        add_to_inv_search_cache( *cont_it );
         if( check_for_zoom && !update_overmap_seen && cont_it->has_flag( flag_ZOOM ) ) {
             update_overmap_seen = true;
         }
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -8946,14 +8946,13 @@ void Character::add_to_inv_search_caches( item &it ) const
 void Character::remove_from_inv_search_caches( item &it ) const
 {
     it.visit_items( [this]( item * i, item * ) {
-        for( auto &cache : inv_search_caches ) {
-            if( cache.second.flag.is_valid() && !i->has_flag( cache.second.flag ) ) {
-                continue;
+        for( auto iter = inv_search_caches.begin(); iter != inv_search_caches.end(); ) {
+            if( ( !iter->second.flag.is_valid() || i->has_flag( iter->second.flag ) ) &&
+                ( iter->second.filter_func == nullptr || ( i->*iter->second.filter_func )() ) ) {
+                iter = inv_search_caches.erase( iter );
+            } else {
+                ++iter;
             }
-            if( cache.second.filter_func && !( i->*cache.second.filter_func )() ) {
-                continue;
-            }
-            inv_search_caches.erase( cache.first );
         }
         return VisitResponse::NEXT;
     } );
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -8989,20 +8989,20 @@ bool Character::has_any_item_with_flag_and_charges( const flag_id &flag ) const
     } );
 }
 
-std::vector<item *> Character::get_items_with( const flag_id &flag,
+std::vector<item *> Character::all_items_with( const flag_id &flag,
         const std::function<bool( item & )> &do_and_check_func )
 {
-    get_items_with( "HAS FLAG " + flag.str(), flag, nullptr, do_and_check_func );
+    all_items_with( "HAS FLAG " + flag.str(), flag, nullptr, do_and_check_func );
 }
 
-std::vector<item *> Character::get_items_with( const std::string &key,
+std::vector<item *> Character::all_items_with( const std::string &key,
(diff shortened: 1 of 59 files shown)


Nebnis:
Long live and god bless for you for making all these optimization and QOL kamayana

akrieger:
nb. you almost never want `std::set` and `std::map`, instead use `std::unordered_set` and `std::unordered_map`. Only use the former two if you also need the items sorted.

Is the test case just, have stuff in inventory and walk around? Does simply waiting exercise the same codepaths?

akrieger:
Instead of adding explicit cache management calls, would a memoization pattern work better? Create the cache on demand when called, drop all caches when any inventory mutation happens? Could we hook into the existing ways that items are added/removed from inventory to do that instead of having to add explicit calls to new functions to keep caches up to date?

Kamayana:
I did some testing, and it looks like normal sets are actually a better choice here than unordered sets. Ordered sets iterate faster, which is important for how often it will be done. The test I did was spawning 100,000 lit candles and checking the optimization time of this `active_light` code:
```int maxlum = 0;
for( item* flagged_item : all_items_with( "is_emissive", &item::is_emissive ) ) {
    const int lumit = flagged_item->getlight_emit();
    if( maxlum < lumit ) {
        maxlum = lumit;
    }
}
```
The results with ordered sets, on the left, were around 50 ms faster on average:

image

akrieger:
Just for academic curiosity then, how important is it to be a set instead of a vector?

Maleclypse:
FYI the below error is consistently popping up so I think it's likely to do with these changes.  
```
 cata_test is a Catch v2.13.7 host application.
(~[slow] ~[.],starting_items)=> Run with -? for options
(~[slow] ~[.],starting_items)=> Randomness seeded to: 1693079606
(~[slow] ~[.],starting_items)=> -------------------------------------------------------------------------------
(~[slow] ~[.],starting_items)=> fueled_bionics
(~[slow] ~[.],starting_items)=>   bio_cable ups
(~[slow] ~[.],starting_items)=> -------------------------------------------------------------------------------
(~[slow] ~[.],starting_items)=> ../tests/bionics_test.cpp:460
(~[slow] ~[.],starting_items)=> ...............................................................................
(~[slow] ~[.],starting_items)=> ../tests/bionics_test.cpp:484: FAILED:
(~[slow] ~[.],starting_items)=>   {Unknown expression after the reported line}
(~[slow] ~[.],starting_items)=> due to a fatal error condition:
Error: (~[slow] ~[.],starting_items)=>   SIGSEGV - Segmentation violation signal
(~[slow] ~[.],starting_items)=> Log messages during failed test:
(~[slow] ~[.],starting_items)=> 12:00:00AM: Your Cable Charger System does not have enough fuel to start.
```

Kamayana:
> FYI the below error is consistently popping up so I think it's likely to do with these changes.

Yeah, I'm pretty sure you're correct, but I'm unable to replicate the error on my machine since all tests run green, so it's not an easy bug for me to fix. It's very possible Akrieger's correct and it's a problem with sets of pointers and I'll have to switch to using vectors that get recreated every time an item is removed, which would be a bit of a bummer.

Kamayana:
> Just for academic curiosity then, how important is it to be a set instead of a vector? Is iteration speed the dominant factor? There's also the funny pointer comparison bug in gcc which meant storing pointers in ordered containers was... not a good idea. But I think we're now officially only supporting gcc versions with the fix in it. https://quuxplusone.github.io/blog/2019/01/20/std-less-nightmare/

The idea was to make it fast to both add and remove pointers from the cache, but it's honestly pretty minor since vectors should also iterate quickly. I'll upload a commit that uses vectors and if that's able to compile then it's settled.

Kamayana:
...welp, still failing, albeit differently. I have no idea what the problem is, I still can't replicate it.

src/character.cpp
```diff
@@ -8884,19 +8888,103 @@ bool Character::in_sleep_state() const
...
+    it.visit_items( [this]( item * i, item * ) {
+        for( auto &cache : inv_search_caches ) {
+            if( cache.second.flag.is_valid() && !i->has_flag( cache.second.flag ) ) {
+                continue;
+            }
+            if( cache.second.filter_func && !( i->*cache.second.filter_func )() ) {
+                continue;
+            }
+            inv_search_caches.erase( cache.first );
+        }
```
andrei8l, line 8986: It crashes here. Example fix:

```suggestion
        for( auto it = inv_search_caches.begin(); it != inv_search_caches.end(); ) {
            if( ( !it->second.flag.is_valid() || i->has_flag( it->second.flag ) ) &&
                ( it->second.filter_func == nullptr || ( i->*it->second.filter_func )() ) ) {
                it = inv_search_caches.erase( it );
            } else {
                ++it;
            }
        }
```

src/character.cpp
```diff
@@ -8884,19 +8888,103 @@ bool Character::in_sleep_state() const
...
+    // Otherwise, add a new cache and populate with all appropriate items in the inventory. Empty sets are still created.
+    inv_search_caches[key].flag = flag;
+    visit_items( [this, &key, &flag]( item * it, item * ) {
+        if( it->has_flag( flag ) ) {
+            inv_search_caches[key].items.push_back( it );
+        }
+        return VisitResponse::NEXT;
+    } );
+
+    return inv_search_caches[key].items;
```
andrei8l, line 8918: Deduplicate code, see suggestions below too

```suggestion
    return all_items_with( "HAS FLAG " + flag.str(), flag, nullptr );
```

src/character.cpp
```diff
@@ -8884,19 +8888,103 @@ bool Character::in_sleep_state() const
...
+    // Otherwise, add a new cache and populate with all appropriate items in the inventory. Empty sets are still created.
+    inv_search_caches[key].filter_func = filter_func;
+    visit_items( [this, &key, &filter_func]( item * it, item * ) {
+        if( ( it->*filter_func )() ) {
+            inv_search_caches[key].items.push_back( it );
+        }
+        return VisitResponse::NEXT;
+    } );
+
+    return inv_search_caches[key].items;
```
andrei8l, line 8938:
```suggestion
    return all_items_with( key, {}, filter_func );
```

src/character.cpp
```diff
@@ -8884,19 +8888,103 @@ bool Character::in_sleep_state() const
...
+    auto iter = inv_search_caches.find( key );
+    if( iter != inv_search_caches.end() ) {
+        // If the cache already exists, use it.
+        return iter->second.items;
+    }
+    // Otherwise, add a new cache and populate with all appropriate items in the inventory. Empty sets are still created.
+    inv_search_caches[key].flag = flag;
+    inv_search_caches[key].filter_func = filter_func;
+    visit_items( [this, &key, &flag, &filter_func]( item * it, item * ) {
+        if( it->has_flag( flag ) && ( it->*filter_func )() ) {
```
andrei8l, line 8953:
```suggestion
        if( ( !flag.is_valid() || it->has_flag( flag ) ) &&
            ( filter_func == nullptr || ( it->*filter_func )() ) ) {
```

andrei8l (commented):
Quick review only.

I'm not too happy with `remove_from_inv_search_caches()`. Consider using the same pattern as in the active item cache: store `safe_reference<item>` instead of `item *` and prune the obsolete entries when the cache is accessed. In this case that means iterating over the cache in `all_items_with()` which might have an unwanted performance impact.

akrieger:
Oh true, these lists can mutate at random points not just beginning/end. set<> it is.

Maleclypse:
Is this intended to still be in draft?

Kamayana:
> Is this intended to still be in draft?

Yes. I just pushed a significant rework. I took [andrei8l](https://github.com/andrei8l)'s comments to heart, and I've now gotten something I'm happier with that's more robust and faster. The cache is now an unordered map of lists of safe item references, which I've determined through testing to be the fastest and safest system to use here (it doesn't have to go as complex as the active item cache's triple-reference-map setup). Items are no longer specifically removed from the cache, their references just become invalid and get cleaned up naturally if the game ever iterates through them.

Instead of the one-size-fits-all function `all_items_with` there's now `cache_visit_items_with` for applying code to the cached items, `cache_has_item_with` for getting a boolean result, and `cache_get_items_with` for creating a vector. I also, amusingly, brought back the `has_item_with_flag` function this whole PR was created to replace, as there ended up still being a need for a non-cache-using function that can look up the un-cacheable temporary flags like `WET` and `ITEM_BROKEN`.

While I'm finished with it now, I'm going to be out of town for a few days, so I can't finish cleaning things up, and even if it was all fine I'd feel a little uneasy merging it while I'm away and unable to help in case any issues bubble up.

Kamayana:
Alright, I'm home and all tests are green, time to open this for review.

One thing I forgot to mention: the cache adds 64 bytes to the 3992 bytes of a character instance, so it gives characters a larger memory footprint. I hope that's acceptable - it would be possible to make the cache an avatar-only variable, but `active_light` is still called for every NPC every draw, so they could still benefit from the cache.

Maleclypse:
Compiled locally and tested following the attached instructions.  Worked as described.

PR #67870, opened 2023-08-24
NPC AI/ fleeing adjustments 2: Come back, little NPC

I-am-Erk:
#### Summary
Features "Tweak NPC fleeing AI based on playtesting"

#### Purpose of change
This continues to work on NPC fleeing behaviour based on playtests from my prior PR, which mostly made NPCs more cowardly. 

#### Describe the solution
1. NPC Max Danger: break this into "monster" and "character" max danger. This allows NPCs to value your and their danger ratings higher than any monster can ever reach. This is pretty important I think, because you'll almost always be seriously outnumbered but far better equipped than your foes. The NPCs should be able to tell that. I believe the lack of distinction is why there are reports that late-game NPCs will never stand their ground even when supremely well equipped. This is handled somewhat by the way NPCs multiply assessments (something I've also made accessible to JSON now), but by splitting characters out from monsters, it means your NPC followers will also appreciate that a bandit with a sniper rifle is a bigger concern than even a very tough zombie.
2. Add an additional coarse boost to assessed allied player threat when close to the player. If the NPC is at your side, they consider your combat ability more relevant than they do if you're far away. This is already figured in during threat evaluation, but there's an extra buff. This should help stop the NPC from fleeing when in formation with the "party leader". In the future I would like to make this part influenced by their trust in you.
~~3. The big part so far: expose several critical NPC AI settings to global_options.json, allowing them to be adjusted without compiling. I will need some help with this to ensure they only check at runtime and aren't being constantly referred to, as I believe that will be potentially slow AF.~~ caching was more fuss than I was willing to try
4. NPCs actually feel pain! NPC pain is figured into their decision to flee, effectively lowering their personality trait for bravery. This is part of a random roll, so sometimes they can hold it together against pain.
5. Fleeign NPC regroups on your position if they get too far away, this is determined by their individual settings on how close you've told them to stay.

NOTE: Currently the 'regroup' effect has them move to the same tile you are in, rather than near you. I'd love for them to move to 1 tile behind you relative to the assessed threat so that they don't get stuck telling you to get out of the way, but I don't understand position stuff nearly well enough for this.

**end result**: The effect I am hoping to get from combining the changes to fleeing behaviour with a higher "in formation" boost is that ally NPCs may choose the safest path to escape from a swarm of enemies, but then they will turn and rejoin you as they move too far away. Once they move to your side, the formation should buff their courage, so if you're staying and fighting they'll choose to as well. On the other hand, if you are running, they'll run with you. It seems to play this way in tests, but this kind of complex behaviour is heavily dependent on terrain and spread of enemies, so it's very tough to test short of just playing the game for several hours.

Infinite. I could be looking more into how NPCs evaluate danger instead of adding coarse adjustments. maybe I still will.
I plan to add some more tweaks to this or a follow up when I can.

#### Testing

Testing that player character proximity drops threat significantly:
image

However if you're far away the threats are still rated highly:
image

When they do run away, they check their rules. If you've told them to stay close, then they'll flee towards you rather than off into the distance.
image

There is a fair bit of jank here but I think I've got it working a lot better.

#### Things I need help on
- I would like the NPC to track if they've been trying to flee for >2 turns and not moved, and if that's the case, decrease assessed danger and attempt to fight. This should be recursive, so if they still say "this is way too dangerous" they'll try another 2 turns and then reassess if they still can't escape. So if the enemy massively outclasses them they will struggle to escape a grab rather than standing their ground, but if it's a closer call, they'll change their mind and fight.
- Similarly if an enemy moves adjacent to them while they're fleeing they should reassess their decision to flee and drop their assessment of the threat, as this could mean they're slower than the thing they're running from and should decide to stand their ground.
~~- Currently the 'regroup' effect has them move to the same tile you are in, rather than *near* you. I'd love for them to move to 1 tile behind you relative to the assessed threat so that they don't get stuck telling you to get out of the way.~~
- NPCs with guns don't seem to be smart enough to back away when the enemy gets close. I think fleeing code could be used for this but I don't know how. We wouldn't want to use quite the same code block because we don't want them shouting that they're running away or anything.

Edit: Akrieger showed me how to expose constants to JSON properly but I'm going to save that for a follow up PR so that we can get this one in


data/core/game_balance.json
```diff
@@ -76,6 +76,41 @@
...
+    "type": "EXTERNAL_OPTION",
+    "name": "NPC_CROWD_BRAVADO",
+    "info": "This determines how much the NPC hates being outnumbered by enemies.  They will increase their perceived number of allies by this value, so the higher it is, the less they care about a dozen zombies coming their way.  Don't set this to a negative number unless you like dividing by zero.",
+    "stype": "int",
+    "value": 0
+  },
+  {
+    "type": "EXTERNAL_OPTION",
+    "name": "NPC_COWARDICE_MODIFIER",
+    "info": "This is a multiplier applied to an NPC's assessment of overall threat. The lower it is, the less likely they are to run away from their problems.  It has been tested between 0.1 and 0.5 pretty effectively.",
```
I-am-Erk, line 110:
```suggestion
    "info": "This is a multiplier applied to an NPC's assessment of overall threat.  The lower it is, the less likely they are to run away from their problems.  It has been tested between 0.1 and 0.5 pretty effectively.",
```

I-am-Erk:
This could be merged as is, but there are a few things I'd love to refine if someone with better code skills is around.

src/npcmove.cpp
```diff
@@ -428,6 +445,10 @@ void npc::assess_danger()
     int hostile_count = 0;
     int friendly_count = 1; // count yourself as a friendly
     int def_radius = rules.has_flag( ally_rule::follow_close ) ? follow_distance() : 6;
+    float NPC_COWARDICE_MODIFIER = get_option<float>( NPC_COWARDICE_MODIFIER_OPT );
+    float NPC_MONSTER_DANGER_MAX = get_option<float>( NPC_MONSTER_DANGER_MAX_OPT );
+    float NPC_CROWD_BRAVADO = get_option<int>( NPC_CROWD_BRAVADO_OPT );
+    float NPC_DANGER_VERY_LOW = get_option<float>( NPC_DANGER_VERY_LOW_OPT );
```
Qrox, line 451: Looks like this function could be called very often so it might make sense to cache these option values.

I-am-Erk: Yes, I agree. I don't know how to do that unfortunately.

Qrox: You can add the options to `cached_options.h/cpp` and cache the options in the following function:

https://github.com/CleverRaven/Cataclysm-DDA/blob/f5fe4d99f38193623fcc202ee1e3051760ba29fa/src/options.cpp#L3995-L4027

src/npcmove.cpp
```diff
@@ -669,24 +689,37 @@ void npc::assess_danger()
         assessment = std::max( min_danger, assessment - guy_threat * 0.5f );
     }
 
+    // being outnumbered is serious.  Do a flat scale up your assessment if you're outnumbered.
+    // This is a coarse tool that might be better handled
+    if( hostile_count > friendly_count + NPC_CROWD_BRAVADO ) {
+        assessment *= std::min( ( hostile_count / ( friendly_count + NPC_CROWD_BRAVADO ) ), 1.0f );
```
gadgetpatch, line 695: The integer division here will chop off the decimal precision and round the ratio down, before it gets passed to the "assessment" float. If you want to preserve the precision, cast one of the components as a float before dividing.
```suggestion
        assessment *= std::min( ( hostile_count / static_cast<float>( friendly_count + NPC_CROWD_BRAVADO ) ), 1.0f );
```

src/npcmove.cpp
I-am-Erk: Thanks, I should be able to cargo-cult it from there when I get home.

src/npcmove.cpp
I-am-Erk: Woops, thanks. I knew about that one but I still miss them sometimes.

PR #48330, opened 2021-04-02
[Magiclysm] New Spell: Nature's Commune

Termineitor244:
#### Summary
Mods "Adds a new druidic spell giving the effects of Animal Empathy and a little health"

#### Purpose of change

Druids are all about their relation with nature, but they currently don't have any means to be in peace with that same nature (fauna), so why don't give them a spell with that effect?

#### Describe the solution

A new druidic spell "Nature's Commune", that makes you enter a magical state of communion with the nature around you, calming the natural animals and making you a little healthier, since you are "embraced" by the mother nature, the health effect is mainly of real (immediate) health, so given enough time without another commune, your body will tend towards its old, unhealthy state.

#### Describe alternatives you've considered

Two different spells, or a healing/stamina regeneration effect.

#### Testing

Made the json changes, tested, they give the health in the proportions I want, I then compiled the game, and the effect in wildlife works as it should, a bunch of rabbits stay in my vicinity if I'm under the effect of the spell, and run away the moment the spell duration ends.

#### Additional context

This was originally a regeneration buff spell, but I discarded that idea and ended with this (in my opinion) better and needed alternative.

This is a draft because I still have to add the spell to books that seem appropriate, but if anyone wants to make changes to the spell itself, I'm open to suggestions.

Edit: Added the spell to the basic druidic book, to a new scroll, and to the Ecoterrorist profession (seems appropriate).

PR #67872, opened 2023-08-24
[Magiclysm] Edit Nature's Commune to provide morale and only be castable outdoors

Standing-Storm:
#### Summary
Mods "Edit Nature's Commune"

#### Purpose of change

The Nature's Commune druid spell had a note that it should provide a morale effect once that's possible through effects. Well, it's not possible through effects, but it is through EoCs

#### Describe the solution

Edit the Nature's Commune to an EoC, which provides the original Nature's Commune effect (Health bonus) and also provide a small morale bonus (+10, max +15, in line with original comment). I also changed it so you can only cast it outdoors and not within 1 overmap tile of a road, which should prevent it from being cast in cities or underground. In exchange, I made the duration much longer (max 1 hour, from around 8 minutes) and changed the health_tick of the effect to 30 (from 1). 

#### Testing

It works!  Go out there and frolic beneath the boughs!

#### Additional context

The spell also says it makes animals calmer but at the moment this is still impossible without giving the caster the hardcoded Animal Empathy mutations.

### DIFF
--- a/data/mods/Magiclysm/Spells/druid.json
+++ b/data/mods/Magiclysm/Spells/druid.json
@@ -479,10 +479,12 @@
   {
     "type": "effect_on_condition":,
     "id": "EOC_NATURE_COMMUNE",
+    "condition": "u_is_outside",
     "effect": [
       { "u_add_effect": "natures_commune", "duration": { "math": [ "( 10000 + (u_val('spell_level', 'spell: druid_natures_commune') * 2000) )" ] } },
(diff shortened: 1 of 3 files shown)


Termineitor244:
Thank you for this! EOCs were still something new (Or non existent I think). Back when I added the spell, thanks for bringing it up to date.

Oh, and the spell does calms the regular wildlife in the same way as the trait, I had to make it hard-coded for the triggered effect of the spell, using a little bit of C++, but it worked just fine!

The original PR that added the spell is: #48330

Standing-Storm:
> Oh, and the spell does calms the regular wildlife in the same way as the trait, I had to make it hard-coded for the triggered effect of the spell, using a little bit of C++, but it worked just fine!

Oh, perfect! No further work needed then!

PR #67876, opened 2023-08-25
Jsonize helicopter piloting requirement

lispcoc:
#### Summary
Infrastructure "Jsonize helicopter piloting requirement"

#### Purpose of change
The conditions for piloting the helicopter are hardcoded.

#### Describe the solution
It should be jsonized in case we decide to implement another specialized vehicle in the future.

#### Testing
Spawn AH-64 Apache
- Player has no Helicopter Piloting proficiency cannot control this.
- Player has Helicopter Piloting proficiency can control this.

remove "control_requirements" from Json
- Player has no Helicopter Piloting proficiency can control this.
- Player has Helicopter Piloting proficiency can control this.

Spawn Atomic car
- Player has no driving skill can control this.

add "control_requirements" to Json
```
"control_requirements": {
  "land": { "skills": [ [ "driving", 1 ] ] }
},
```
- Player has no driving skill cannot control this.
- Player has driving skill can control this.

#### Additional context
- Is it better to separate ground control and water control?
- Is it appropriate to make having a rotor a condition of being a helicopter?


PatrikLundell:
Gyrocopters have rotors but are not quite helicopters. I still think they would classify as helicopters if you don't add a specific gyrocopter classification. You nowadays also have drones that have rotors, but if it's large enough to be piloted I'd classify it as a helicopter. 

Sometimes in the future it may be useful to add an override telling you that the rotor doesn't require helicopter skills because electronics "translates" the rotor control logic into easy to use logic where you tell the controls what to achieve rather than control various parts to do their things to try to achieve what you want to do (which I think is what drone controls do).

gadgetpatch:
Oh, we might want to coordinate a bit: #67854

I just put together a system for land and water vehicle proficiencies. It allows non-proficient drivers, but causes over or understeering when turning, reduces handling in rough conditions, and reduces effective driving skill by a percentage equal to proficiency progress (0% effective skill at 0% proficiency, 50% at half proficiency, etc.

lispcoc:
> There shouldn't be any conflicts, but we might want to coordinate a bit: #67854
> 
> I just put together a system for land and water vehicle proficiencies. TL;DR, it assumes we still allow non-proficient drivers in most cases (special vehicles could still require proficiencies, like tanks or aircraft), and non-proficiency instead causes over or understeering when turning, reduces handling in rough conditions, and reduces effective driving skill by a percentage equal to proficiency progress (0% effective skill at 0% proficiency, 50% at half proficiency, etc.). Proficiency trains by operating a vehicle on land or water, so people without car experience can fumble their way into learning the basic controls.
> 
> I was considering expanding it to cover helicopters (and eventually any small planes as well) ...but considering this PR, it might be best as a 2-tiered piloting proficiency? I'm thinking you couldn't get any aircraft off the ground without at least basic Aircraft Training. A tier-2 Helicopter Pilot (or Plane Pilot) proficiency would prevent errors while flying rotary (or fixed wing) craft, which you'd learn via flying for long enough (or via starting profession). Something like that.

Is it an idea to divide the current helicopter proficiency into two, for example, "prof_basic_aircraft_pilot" and "prof_helicopter_pilot", and make "prof_basic_aircraft_pilot" the piloting requirement? 
I think it's an interesting idea to implement, but I'm not sure how to translate it into this PR.
If possible, it would be very helpful if you could show me a few lines of JSON code that you would ideally like. (sorry for my poor english)

gadgetpatch:
> Is it an idea to divide the current helicopter proficiency into two, for example, "prof_basic_aircraft_pilot" and "prof_helicopter_pilot", and make "prof_basic_aircraft_pilot" the piloting requirement? I think it's an interesting idea to implement, but I'm not sure how to translate it into this PR. If possible, it would be very helpful if you could show me a few lines of JSON code that you would ideally like. (sorry for my poor english)

Oh, I'm sorry for any confusion! I think this looks fine as it is. We can think about adding airplane proficiencies in a future Pull Request. I was just letting you know we were working on similar things!

It does look like we both edited vehicles.cpp and vehicles.h though, so I'm sorry if I've made merging a little more complicated!

PR #67877, opened 2023-08-25
[Magiclysm] Rework Nature's Bow to require ammo, add spell to create ammo

Standing-Storm:
#### Summary
Mods "Rework Nature's Bow to require ammo, add spell to create ammo"

#### Purpose of change

Nature's Bow was never intended to be a infinite magical gun long term (see [here](https://www.reddit.com/r/cataclysmdda/comments/cbhr08/comment/eth00c1/)), and it's so powerful that for a long time it was _the_ reason for people to pick druid. Also, it rendered the druid's craftable bow useless out of the gate because Nature's Bow was better in every way except that you couldn't cast spells while using it, but you didn't need to because you could just use Nature's Bow. 

#### Describe the solution

Change the spell so it summons a magical bow, a magical quiver, and a set of magical arrows (the number of arrows increase with spell level). Similarly change the improved Nature's Bow so that it summons even more arrows and a quiver that holds even more arrows. Nature's Bow has a quiver that holds 60 arrows and Improved Nature's Bow has a quiver that holds 120.

Add an additional spell, Harvest of the Hunter, that creates arrows out of sticks, bark, or boughs.  These arrows have the TANGLE special property.

#### Testing

Everything works! The quiver is summoned already worn and the arrows are summoned in the quiver.

#### Additional context

This is where I burn all the good-will I've built up so far.


ashGlaw:
If you split it into two spells, can they both be learned from the book of trees? Or is there a different source?

Standing-Storm:
> If you split it into two spells, can they both be learned from the book of trees? Or is there a different source?

I didn't split it exactly--Nature's Bow also summons arrows to go with the bow (that should be recoverable just like normal arrows), I just added an additional source of arrows through an additional spell that works for any bow,

Termineitor244:
It's a pretty good change! Infinite arrows were never a good thing to have with this bow.

Just one thing, the Archer Druid profession should start with this new arrow spell. It makes sense, I think. Maybe the Moose rider, too? But that one can be argued to just craft the arrows the old, mundane way.

Standing-Storm:
Redrafting this because the quiver test is constantly failing for no obvious reason

Standing-Storm:
Alright, this is finally ready.

PR #67891, opened 2023-08-25
[MoM] Make psionic powers unteachable

Standing-Storm:
#### Summary
Mods "Make psionic powers unteachable"

#### Purpose of change

The big thing preventing me from adding random NPCs with psionics is that you could have had them teaching you how to read minds like it was a pottery class.  Now that "teachable" is a field, I can change that and then go add random NPC psions.

#### Describe the solution

Add "Teachable: false" to every single power.

#### Additional context

This doesn't mean that in lore it's impossible to teach powers, just that it's not the kind of thing you can do in 3-6 hours while resting between lab raids. 

If conditions are ever added to NPC teaching ( some way to say like { "u_has_trait": "BIOKINETIC" } ) I'll revisit this.


GuardianDll:
> If conditions are ever added to NPC teaching ( some way to say like { "u_has_trait": "BIOKINETIC" } )

There is? I miss what's the problem

Standing-Storm:
> There is? I miss what's the problem

Through the automatically follower training? Starting the seminar is all talk topics, but IIRC the list of options chosen that the NPC can teach is set through C++

Issue #67653, opened 2023-08-13
CORNERED_FIGHTER appears to be broken


NetSysFire:
### Describe the bug

The critters affected by it do not seem to fight back.

### Attach save file

n/a

### Steps to reproduce

1. Spawn a bear in an enclosed space where it can not run.
2. Corner it.
3. Nothing happens.
4. Hit it.
5. Nothing happens again.

I might be severely dumb but if I saw that correctly the bear does not even aggro under normal circumstances if you hit it.
I also tested with `mon_cat` and added `CORNERED_FIGHTER` into its JSON, still no dice.

### Expected behavior

The creature gets aggressive.

I wonder if this is another case of a feature silently breaking, like the MEAT `placate_trigger`.

### Versions and configuration

- OS: Linux
    - OS Version: LSB Version: n/a; Distributor ID: Arch; Description: Arch Linux; Release: rolling; Codename: n/a; 
- Game Version: 5cca213 [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Slowdown Fungal Growth [no_fungal_growth],
    Bionic Professions [package_bionic_professions]
]

fairyarmadillo:
This is not a problem with CORNERED_FIGHTER as far as I can tell, it's a problem with how aggression is set up. If you test this with cockroaches they work fine.

Animals which are not extremely aggressive tend to get stuck in "tracking" because of this code here:

```
    if( effective_morale < 0 ) {
        if( effective_morale + effective_anger > 0 && get_hp() > get_hp_max() / 3 ) {
            return MATT_FOLLOW;
        }
        return MATT_FLEE;
    }

    if( effective_anger <= 0 ) {
        if( get_hp() <= 0.6 * get_hp_max() ) {
            return MATT_FLEE;
        } else {
            return MATT_IGNORE;
        }
    }

    if( effective_anger < 10 ) {
        return MATT_FOLLOW;
    }

```

The problem appears to be that there's a hard check for whether they're below 1/3 HP or not. If they aren't, they just flip to tracking. Furthermore, black bears have -10 aggression and 60 morale. Hurting them raises their aggression, but not quickly enough to really get them to react before you've had time to beat them into the dirt.

[...]

PR #67892, opened 2023-08-25
Make some dangerous herbivores and their upgrades actually dangerous.

Nebnis:
#### Summary
Balance "Made some herbivores and their upgrades more dangerous"

#### Purpose of change
Somewhat adress #67653. Bear and mooses in-game are total wimps, they die and get stuck tracking before they even think about attacking you, someone with a kitchen knife and no melee skills can easily kill them cause their hp gets reduced too fast.

#### Describe the solution
Buffed the bear family hp and gave them a bite attack. Added player close anger trigger to the normal bear, you still have some amount of time to run away in case you run into them accidentally.

Mooses got a stomp attack since irl they are know to use their hooves with great effect to kick whatever presents a threat to them.

Let people continue to tango melee wise with a bear without consequences.

#### Testing
I think I've managed to strike a good balance with them now. You need an actual gun to kill them without any danger and trying to melee them as inexperienced survivor and no prep gets you in big time trouble.


Nebnis:
Well looks like i might need some assistance with this. I got no idea how to fix the archery test that is preventing me from raising the bear and moose hp to 200+ and kevin got the okay to fix it by adding a crit and my weakpoint to make it pass i think. This way i could probably put them around 250 or so to feel balanced

Issue #8417, opened 2014-08-03
Fix signal handler


elfring:
The functions "[erase](http://invisible-island.net/ncurses/man/curs_clear.3x.html)" and "[exit](http://pubs.opengroup.org/onlinepubs/9699919799/functions/exit.html)" do not belong to the list of [async-signal-safe functions](https://www.securecoding.cert.org/confluence/display/seccode/SIG30-C.+Call+only+asynchronous-safe+functions+within+signal+handlers).
I guess that a different program design will be needed for your function "[exit_handler](https://github.com/CleverRaven/Cataclysm-DDA/blob/4079fcfeb82ef2aea4e64ba69dd2125136c68b1f/src/main.cpp#L326)".
##

kevingranade:
Instead of calling exit(), follow this article https://www.cons.org/cracauer/sigint.html

Instead of calling clear(), invoke the ncurses clean up handler as mentioned here https://stackoverflow.com/questions/44084793/handle-signals-in-ncurses

The windows terminal cleanup calls are probably also a problem.

PR #67893, opened 2023-08-25
Instead of calling `exit()`, reset and signal the default handler

alef:
#### Summary
Bugfixes "Fix signal handler: call to exit()"

#### Purpose of change
Related to #8417 (yes, I sorted by oldest issues). Changes to the code as by the article mentioned [here](https://github.com/CleverRaven/Cataclysm-DDA/issues/8417#issuecomment-394116340).

#### Describe the solution
If the signal handler was raised by `SIGINT` (2), the handler reset the signal disposition to `SIG_DFL` and send the same signal to itself (the process).

* ✔ Use `sigaction()` instead of plain `signal()` inside the handler, but wasn't sure passing the `struct sigaction` object from the handler stack would survive unwinding, or whether `sigaction()` would care at all.
* ❌ Transform `exit_handler` to a variadic template function just to get rid of this `-999` calls. Could not get around _gcc_ warnings.
* ❌ Add the default value `-999` to `exit_handler` and change the other calls.
* ❌ Replace `2` with `SIGINT`.
* Write a PR text shorter then the patch itself.

#### Testing
* Linux Curses, `CTRL+C`, `[N]o`, `CTRL+C` again, `[Y]es`, shell return code was `130`.
* Linux Tiles, `CTRL+C` from the terminal, `[N]o`, menu opens, `Esc`, `CTRL+C` from the terminal, `[Y]es`, shell return code was `130`.
* Windows builds and runs fine.
* Can't test the Android `#ifdef` section, nor anything MacOS.

The _menu opens_ above is maybe a bug in my system using WSLg, or the SDL input functions eating the keypress twice. Not a regression since the same happens without this PR changes.

#### Additional context
`object_creator/creator_main.cpp` should receive the same changes.

### DIFF
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -56,7 +56,7 @@
 #include "type_id.h"
 #include "ui_manager.h"
 #if defined(MACOSX)
-#   include <unitstd.h> // getpid()
+#   include <unistd.h> // getpid()
 #endif
 
 #if defined(PREFIX)
--- a/src/main.cpp
+++ b/src/main.cpp
@@ -151,7 +151,11 @@ void exit_handler( int s )
 
 #if !defined(_WIN32)
         if( s == 2 ) {
-            signal( s, SIG_DFL );
+            struct sigaction sigIntHandler;
+            sigIntHandler.sa_handler = SIG_DFL;
+            sigemptyset( &sigIntHandler.sa_mask );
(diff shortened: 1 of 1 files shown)


Maleclypse:
I think you have a clang error

alef:
> I think you have a clang error

I run the clang-tidy locally and it passes.

PR #67907, opened 2023-08-26
Default loot zone

smileynet:
#### Summary
Features "Add 'default' lowest priority loot zone"

#### Purpose of change

Adds a default loot zone to use as a catch-all when more specific zones are not found.

Currently the only way to accomplish this currently is using a custom zone, which takes top priority, making it override everything (and can cause user frustration when left enabled by accident on say a vehicle cargo zone :flushed:) and requires manual negative filtering to exclude categories of items.
#### Describe the solution

The default loot zone is selected if no other loot zones match the item being sorted. 

Creating the ability to set the priority of custom zones to be higher or lower than static loot zone types. This would be a more extensive change requiring additional menu questions when creating loot zones and potentially be a more confusing solution.

Otherwise, continue to be frustrated when I forget to disable a custom loot zone in a vehicle, and manually subtract categories when initially building out a base.
#### Testing

1/ Tested in a fresh world, shelter start. Verified default zone is only used when a more specific zone is not present. Static and personal versions both tested.

2/ Tested on a existing save to ensure expected behavior for both types.
#### Additional context

Currently there is no pass-through logic for if a specific zone is full to use a less specific zone. Based on my limited testing this currently seems to be the case for custom zones as well (although I'm not 100% certain about it). I plan to explore having the Default loot zone optionally absorb overflow items when more specific zones are full in a future PR.


PatrikLundell:
It can be noted that there is (or was) a partial implementation of this, with perishable food/drink ending up in the corresponding unspecified zone, and corpses (and probably other things) ending up in the "others" zone.

A fallback zone is useful, and would be even more useful if it supports exclusion of things you don't want to move (such as spending the next two days moving shards of glass one at a time when you don't actually want the blasted things anyway, trying to organize things from a lab).

smileynet:
> It can be noted that there is (or was) a partial implementation of this, with perishable food/drink ending up in the corresponding unspecified zone, and corpses (and probably other things) ending up in the "others" zone.
> 
> A fallback zone is useful, and would be even more useful if it supports exclusion of things you don't want to move (such as spending the next two days moving shards of glass one at a time when you don't actually want the blasted things anyway, trying to organize things from a lab).

The handling for firewood, corpses, and perishable priority are just if statements in the same function that return sooner based on specific conditions... Filthy is somewhere else, but handled similarly. But all of them are specific special cases. 

Because of the current side effect of ignoring items with existing zones, even if they're full, you could just make a custom zone set to all your least favorite things and keep it full from the start. I have been thinking about how to manage the tedium of lots of small things though, and could see creating an insertion zone (or maybe the logic would be better elsewhere) that uses bags/ containers to corral said small, irksome things (I've yet to find which change made moving that stuff soooooo much slower...)

At some point I could see trying to streamline this with a better priority based system, that could handle other cases more easily (the continued growth of incredibly specific categories has left me wanting to add some top level super categories). But I'm trying to do this in baby steps 😅

PR #67914, opened 2023-08-27
[Magiclysm] Add seven more spells (+ monsters + lab special)

Standing-Storm:
#### Summary
Mods "Add even more Magiclysm spells (+ lab special)"

#### Purpose of change

I still have more spell ideas, and Magiclysm's magical elements need more blending into the pre-Cataclysm world

#### Describe the solution

Add seven more spells:

1. Restore Genetic Stability (Biomancer): This spell lowers your instability from mutating. True to its function, it's not found anywhere on the outer world--it's found in a microlab special, because only XEDRA would have any reason to develop this spell.  I also added two monsters--the feral research thaumaturge and the feral radiation mage, along with an enchanted lab coat for your scientist-wizard.
2. Carrion Feast (Biomancer): Cast this spell can you can eat rotten food! It doesn't taste any better, though...
3. Knowing the Day and the Hour (Classless): Tells you the exact time. Amazing innawood, useless in the modern world. I thought of this as an example of a spell that absolutely would have existed in the past that modern technology rendered completely obsolete.
4. Magus's Mark (Magus): Metaphysically mark a single location, allowing you to teleport back to it with:
5. Word of Recall (Magus): Return to your Magus's Mark.  You may only return to a single location--casting a new Magus's Mark overwrites the previous Mark. 
6. Summon Watcher Spirit (Animist): Summon a spirit called a watcher, through whose eyes you can see. It's unobtrusive but not totally invisible. 
7. Beguiling the Savage Beast (Druid): Charm a natural animal.  Includes mutated animals, but not insects. I was originally going to have it not work on nether-touched animals, but the utility and compatibility (with Megafauna, for example) of being able to write "targeted_monster_species": [ "MAMMAL", "BIRD" ] overrides that.  Maybe someone will split species into more fine-tuned categories in the future. 

Add one microlab special, containing some feral wizards, a scroll of Restore Genetic Stability, and some magical items.

Add two monsters, as above--the feral research thaumaturge (magus spells), and the feral radation mage, which hurls radioactive gas around and summons zombies (because you need to be an animist to be a radiation mage). 

#### Testing

Spells all do what I want them to. Magus's Mark accurately only allows a single location and future castings overwrite that location, Carrion Feast lets you eat rotten or raw meat without a problem, you can see through the eyes of watcher spirits if you're close enough, Beguiling the Savage Beast tames animals, and Knowing the Day and the Hour would be amazing if smartphones didn't exist.

Feral research thaumaturges and radiation mages show up occasionally in labs but pretty rarely.

#### Additional context

I'm planning on including even more everyday connections to the pre-Cataclysm magical world in the future--XEDRA would absolutely have used magic during their experiments.


TheShadowFerret:
What disadvantage does Magus' Mark have to Translocate Self(other than "screw Animists")? It's lower level & doesn't require building a gate/finding the basement special with one.

Standing-Storm:
Magus's Mark only lets you pick a single destination to return to--if you cast it again, it overwrites the precious choice, so there's no way to set up a network like Translocate Self lets you do.

Standing-Storm:
Alright, this should be ready.

Feral radiation mage balance is about where I want it for a mage--if you shoot them in a head immediately you'll be fine, if you let them start casting a few spells you're in big trouble.

Termineitor244:
Good stuff!!

The only change I would like in this is for the teleportation spells to be known to the Dimensionalist Mage profession, since it would be weird for it to not know the spells.

The druid spell and the hour an day one could be used by some of the druids professions too, if you feel inclined to give them to them.

Standing-Storm:
I admit I had no idea that there was so many Magiclysm professions! I'll have to be sure to check them in the future when I'm adding more stuff.

Termineitor244:
> I admit I had no idea that there was so many Magiclysm professions! I'll have to be sure to check them in the future when I'm adding more stuff.

Several of them are fairly new! But yeah, we have a good number of them, more variety is always a good thing.

Maleclypse:
Just realized I should have left this up for that change discussed above.

Standing-Storm:
Adding spells to professions was the last commit I made, so there's no problem

PR #67920, opened 2023-08-27
Bugfix: Exploding 💥 zombie cows with landmines

### LINKED ISSUE #60363: Zombie cows do not take damage from land mines
### Describe the bug

I just walked two zombie cows over multiple land mines and neither of them took any damage

### Steps to reproduce

1. Make a zombie cow walk over a land mine

### Expected behavior

Zombie cow go boom

### Screenshots

_No response_

Experimental 2022-08-21-1332

### Additional context

_No response_

inogenous:
#### Summary
Bugfixes "Exploding zombie cows with landmines"

#### Purpose of change

* Fixes an issue where a zombie cow stepping on a landmine would previously not damage the cow at all. The mine exploded but the cow took no damage.
* Fixes #60363

#### Describe the solution

* The issue before was
	* The method `map::build_obstacle_cache` sets initial values for `.density`
	* Because of the cow size, the `density` on the tile of the cow got set to `0`. [Link](https://github.com/CleverRaven/Cataclysm-DDA/blob/cdda-experimental-2023-08-27-0712/src/map.cpp#L8998)
	* When `density` is zero, no damage will be done.
* The intention with `initial_cloud` seems to be that the density instead should be equal to the fragment count, and decreasing outward.
* Therefore, this change simply sets the initial value for `.density` regardless of the size of any creature that might be placed there.

#### Testing

1. Place landmine.
2. Debug spawn zombie cow
3. Make zombie cow step on landmine
4. Notice whether the cow took any damage. With the suggested fix, it now does.

Below are tables of the values for `density` for each tile surrounding the cow stepping on the landmine. The cow steps on the tile center.

Zombie cow on landmine, density before:
```
|  0  |  0  |  0  |
|  0  |  0  |  0  |
|  0  |  0  |  0  |
```

Zombie cow on landmine, density after:
```
|  0  |  0  |  0  |
|  0  | 160 |  0  |
|  0  |  0  |  0  |
```

Regular zombie on landmine, density before:
```
|  15 |  15 |  15 |
|  15 | 160 |  15 |
|  15 |  15 |  15 |
```

Regular zombie on landmine, density after (no change, still works):
```
|  15 |  15 |  15 |
|  15 | 160 |  15 |
|  15 |  15 |  15 |
```

### DIFF
--- a/src/explosion.cpp
+++ b/src/explosion.cpp
@@ -415,6 +415,7 @@ static std::vector<tripoint> shrapnel( const Creature *source, const tripoint &s
     fragment_cloud initial_cloud = accumulate_fragment_cloud( obstacle_cache[src.x][src.y],
     { fragment_velocity, static_cast<float>( fragment_count ) }, 1 );
     visited_cache[src.x][src.y] = initial_cloud;
+    visited_cache[src.x][src.y].density = static_cast<float>( fragment_count );
 
     castLightAll<fragment_cloud, fragment_cloud, shrapnel_calc, shrapnel_check,
                  update_fragment_cloud, accumulate_fragment_cloud>


PatrikLundell:
What about a regular cow on a land mine?

inogenous:
> What about a regular cow on a land mine?

Also did not take damage before. With this suggested change, it does.

Before:
exploding-cow-before

After:
exploding-cow-after

PR #67924, opened 2023-08-27
[Magiclysm] Add RNG to crystallized mana spawns

Standing-Storm:
#### Summary
Mods "Add RNG to crystallized mana spawns"

#### Purpose of change

There's a spell to create crystallized mana, but since crystallized mana spawns didn't have a range defined they all spawned with the max possible mana so you never need to use it.

#### Describe the solution

Add a range to existing crystallized mana spawns. I set it from 1-10 because the Crystallized Mana spell only produces 1 at a time (5 for manatouched mutants), so there was probably an industrial bottleneck in terms of the number of mages willing to be sleepy 24/7 to churn the stuff out. Smashing big mana crystals still gives you dozens of crystallized mana as before--since that already had a range implemented I didn't change it.

#### Testing

It works

### DIFF
--- a/data/mods/Magiclysm/itemgroups/itemgroups.json
+++ b/data/mods/Magiclysm/itemgroups/itemgroups.json
@@ -328,7 +328,7 @@
       { "group": "potions_rare", "prob": 5 },
       { "group": "alchemy_items", "prob": 30 },
       { "group": "magical_reagents", "prob": 20 },
-      [ "crystallized_mana", 55 ],
+      { "item": "crystallized_mana", "prob": 55, "charges-min": 5, "charges-max": 150 },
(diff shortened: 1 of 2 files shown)


GuardianDll:
Actually, with magic wands using 1 crystal per spell, i think original 1 crystal / 5 crystals with manatouched mutation is correct amount, it's just the game spawn is busted. Can you please sane it? Either by decreasing item groups where one spawn or by changing the crystal count or even by removing charges entirely, all approaches would work

Standing-Storm:
I think it has to keep charges since it's an ammo type for the wands, but I could reduce it to 1-10 per spawn point.  Does that sound like a reasonable amount?

PR #67932, opened 2023-08-28
Unload sparse items only zone option

smileynet:
#### Summary
Features "Unload sparse items only zone option"

#### Purpose of change

This PR is intended to address two problems I encounter while playing:
1/ With recent change to movement of small items like powders, it has become highly undesirable to accidentally unload bags full of spices or tea leaves commonly found in kitchens. Doing so can cause players to inadvertently spend large amounts of in game time shuffling around high quantities of small items with little benefit. 

2/ I keep a 9 square personal unload zone for easy use of unloading stuff in the field, but sometimes forget to toggle it off in my home area, causing me to unload containers filled with large amounts of items such as chitin, patches, etc.

#### Describe the solution

The goal is to create on option to allow me to predictably unload certain containers while avoiding others by introducing another dialog option along with the existing 'Loot: Unload Everything' questions about molle, mods, and unload everything, asking if the players want to avoid unloading items in quantities. The threshold is set to 20 by default, which will allow wallets to be fully unloaded by default based on current spawn rates, but is customizable in game options should players prefer a different level. 

Additionally, I've applied a consistent formatting to the option questions, and added a cautionary note to the remove mods option about not necessarily being able to reattach mods that are removed to help new players avoid stripping a found tacticool rifle by accident. 

This stemmed from me taking some pretty elaborate steps to avoid unloading bags of salt, tea, etc, which are incredibly common. Additionally, with the max item count for storage mechanic, not inadvertently unloading containers full of fabric patches or chitin is a plus as well.

I think providing better strategies for moving large quantities of items could help mitigate this, but would have a much larger impact on gameplay that I am not looking to wade in to at the moment. I am considering some kind of loot insertion zone to proactively move large quantities of items to containers, but it is going to be a more extensive change, and would be complementary to this anyway.
#### Testing

Tested to confirm that items using charges (i.e. charcoal) are not impacted by this option, but that items in stacks (i.e. salt) are. Got a plastic bag of 100 salt, verified that it does not unload when the spare only option is selected for the unloading zone. Set the threshold to 100 and made sure the bag was unloaded. Set back to 20 to confirm it did not unload again. Tested with regular and personal zones.

### DIFF
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -2236,7 +2236,8 @@ void activity_on_turn_move_loot( player_activity &act, Character &you )
                         for( item *contained : it->first->all_items_top( item_pocket::pocket_type::CONTAINER ) ) {
                             // no liquids don't want to spill stuff
                             if( !contained->made_of( phase_id::LIQUID ) && !contained->made_of( phase_id::GAS ) ) {
-                                if ( unload_sparse_only && item_counts[contained->typeId().str()] > get_option<int>("SPARSE_ITEM_THRESHOLD") ) {
+                                if( unload_sparse_only &&
+                                    item_counts[contained->typeId().str()] > get_option<int>( "SPARSE_ITEM_THRESHOLD" ) ) {
                                     continue;
                                 }
                                 move_item( you, *contained, contained->count(), src_loc, src_loc, vpr_src );
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -2236,7 +2236,7 @@ void activity_on_turn_move_loot( player_activity &act, Character &you )
                         for( item *contained : it->first->all_items_top( item_pocket::pocket_type::CONTAINER ) ) {
                             // no liquids don't want to spill stuff
                             if( !contained->made_of( phase_id::LIQUID ) && !contained->made_of( phase_id::GAS ) ) {
-                                if ( unload_sparse_only && item_counts[contained->typeId().str()] > 20 ) {
+                                if ( unload_sparse_only && item_counts[contained->typeId().str()] > get_option<int>("SPARSE_ITEM_THRESHOLD") ) {
                                     continue;
                                 }
                                 move_item( you, *contained, contained->count(), src_loc, src_loc, vpr_src );
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -2155,6 +2155,7 @@ void activity_on_turn_move_loot( player_activity &act, Character &you )
 
         bool unload_mods = false;
         bool unload_molle = false;
+        bool unload_sparse_only = false;
         bool unload_always = false;
 
         std::vector<zone_data const *> const zones = mgr.get_zones_at( src, zone_type_zone_unload_all,
@@ -2165,6 +2166,7 @@ void activity_on_turn_move_loot( player_activity &act, Character &you )
             unload_options const &options = dynamic_cast<const unload_options &>( zone->get_options() );
             unload_molle |= options.unload_molle();
             unload_mods |= options.unload_mods();
+            unload_sparse_only |= options.unload_sparse_only();
             unload_always |= options.unload_always();
         }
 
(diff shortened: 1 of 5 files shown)


GuardianDll:
Can we make it unload if it's more than 1 layer deep? So corpse > backpack > bottle of pills > pills will unload corpse and backpack, but not pills?

smileynet:
> Can we make a setting to unload if it's more than 1 layer deep? So corpse > backpack > bottle of pills > pills will unload corpse and backpack, but not pills?

Just tested to confirm that this is the case based on the existing, iterative way unloading happens. The sparse only logic checks a given pocket at the time the parent container is being unloaded, which only occurs when said parent container has been unloaded from other containers.

LMK if there's a corner case I should check for 😄

bombasticSlacks:
We are trying to cut down on menu options. The zone builder should probably just prompt for a count limit, and store it, rather than a boolean value.

Best UI option would be to ask if you want to limit, then prompt for the value. Under the hood saying no could just set a value of -1 or an arbitrarily large value.

I also feel like we should have a way of counting the number of something in a container, instead of doing it manually like you are, but I could be wrong.

smileynet:
> We are trying to cut down on menu options. The zone builder should probably just prompt for a count limit, and store it, rather than a boolean value.
> 
> Best UI option would be to ask if you want to limit, then prompt for the value. Under the hood saying no could just set a value of -1 or an arbitrarily large value.
> 
> I also feel like we should have a way of counting the number of something in a container, instead of doing it manually like you are, but I could be wrong.

I've moved the sparse threshold from general options to zone specific options.

As far as stack count, the closest I found was one related to [AIM](https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/advanced_inv_pane.cpp#L111). I've made a note to circle back and derive a more general helper function when time and wisdom allow, assuming there's not something else I've missed that someone calls out in the mean time 😁

src/clzones.cpp
```diff
@@ -42,6 +42,7 @@
 #include "vehicle.h"
 #include "visitable.h"
 #include "vpart_position.h"
+#include "options.h"
```
bombasticSlacks, line 45: Don't need this header any more I don't think.

src/activity_item_handling.cpp
```diff
@@ -2199,9 +2203,21 @@ void activity_on_turn_move_loot( player_activity &act, Character &you )
                 if( dest_set.empty() || unload_always ) {
                     if( you.rate_action_unload( *it->first ) == hint_rating::good &&
                         !it->first->any_pockets_sealed() ) {
+                        std::unordered_map<std::string, int> item_counts;
+                        if( unload_sparse_only ) {
+                            for( item *contained : it->first->all_items_top( item_pocket::pocket_type::CONTAINER ) ) {
+                                if( !contained->made_of( phase_id::LIQUID ) && !contained->made_of( phase_id::GAS ) ) {
+                                    item_counts[contained->typeId().str()]++;
```
bombasticSlacks, line 2210: Might be more efficient to use the int ID rather than string ID. Unless this is used to handle renamed items 🤔

kevingranade:
"Avoid unloading things like salt" vs "put a limit on how much of an item you unload".

Does this need to be an option at all? Should we never unload e.g. powders? Could we just have a flag in addition to powder material that means "this preferrs to stay in a container"?

Because of this disparity between goal and solution,  it seems unlikely that people are going to be able to discern what the option is for.

src/clzones.cpp
smileynet: Removed

src/activity_item_handling.cpp
smileynet: Resolved.

smileynet:
> "Avoid unloading things like salt" vs "put a limit on how much of an item you unload".
> 
> Does this need to be an option at all? Should we never unload e.g. powders? Could we just have a flag in addition to powder material that means "this preferrs to stay in a container"?
> 
> Because of this disparity between goal and solution, it seems unlikely that people are going to be able to discern what the option is for.

Re-pasting original response in Discord for continuity:

I originally started with "don't unload powders"... which, maybe that should still just be a thing, but my other common issue was unloading bags/boxes of things like patches I'd stashed by inadvertently unloading them in a personal unloading zone (I usually set it as the 9 squares around my char). Not to mention that when you consider other similar things, like tea leaves, which are not as distinctly tagged, the sprawl goes wide fast if you try to go by tag/ category. So ultimately this ended up addressing both cases in what was the most consistent behavior I could imagine.

Discussion notes:
kevingranade's main concerns were that there were alternative ways to solve this problem that might be preferable (such as special handling for powders and similar items via a flag that prevent them from being unloaded in a zone), and that options such as this needed to continue to add sufficient value to warrant being added/ maintained in the code base.

Other things that came out of the discussion were: 1/ Exploring treating powders more like liquids in terms of behavior (can't unload on to floor, must transfer between containers), 2/ Reworking strip corpse to be a more generic "de-nest containers" i.e. unload all containers of containers but not containers filled with end items, 3/ Explore zone to aggregate container contents from many small to few larger containers, 4/ Explore making personal zones optionally persist across characters.

PR #67933, opened 2023-08-28
unimpaired_range: Minor docs fixes

Brambor:
#### Summary
None

#### Purpose of change
~Fix FIXME.~
Fix grammar.
Finish a docs sentence, probably right.

#### Testing
Looked around instances where unimpaired_range is used to make sure it is't weird after the rename.

### DIFF
--- a/src/creature.h
+++ b/src/creature.h
@@ -391,7 +391,7 @@ class Creature : public viewer
         /*@}*/
 
         /**
-         * How far the creature sees under the given light. Places outside this range can
+         * How far the creature sees under the given light. Creature cannot see places outside this range.
          * @param light_level See @ref game::light_level.
          */
         virtual int sight_range( float light_level ) const = 0;
--- a/src/map.h
+++ b/src/map.h
@@ -366,7 +366,7 @@ class map
 
         /**
          * A pre-filter for bresenham LOS.
-         * true, if there might be is a potential bresenham path between two points.
+         * true, if there might be a potential bresenham path between two points.
          * false, if such path definitely not possible.
          */
         bool has_potential_los( const tripoint &from, const tripoint &to,
--- a/tests/char_sight_test.cpp
+++ b/tests/char_sight_test.cpp
@@ -265,10 +265,10 @@ TEST_CASE( "character_sight_limits", "[character][sight][vision]" )
 // equivalent to being nearsighted, which can be corrected with glasses. However, they have a
 // nighttime vision range that exceeds that of normal characters.
 //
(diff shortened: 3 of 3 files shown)


PatrikLundell:
I'm afraid I have to disagree with the change. The range seems to be the range as it would be while unimpaired by environmental factors, i.e. before dynamic effects such as smoke, low visibility, and obstacles (and, I assume, being lit up and the like which would counteract some of the negative factors within the max range) are taken into account.
See e.g. "/** Returns the player maximum vision range factoring in mutations, diseases, and other effects */".

Zireael07:
Huh. I have to say, as a non-native English speaker, that unimpaired_range is a mouthful and easily confusable.

Probably worth renaming to... "normal_range"? "expected_range"? "unimpeded_range" (impeded without the "un" is almost never used)

PatrikLundell:
I'm all for a change to a different word that conveys the correct meaning, although I can't come up with a really good alternative myself.

Brambor:
> Huh. I have to say, as a non-native English speaker, that unimpaired_range is a mouthful and easily confusable.
> 
> Probably worth renaming to... "normal_range"? "expected_range"? "unimpeded_range" (impeded without the "un" is almost never used)

I like `unimpended_range`, but I had to translate it.
What about `unobscured_range`? It's more widely known.

I-am-Erk:
I appreciate where your head is at, but I generally recommend against small tweaks to the names of things on the back end. It can cause confusion to programmers who already know the systems, and as we see here, it's not necessarily clear which is the better name anyway. It mostly just confuses the git blame without changing anything.

Brambor:
> I appreciate where your head is at, but I generally recommend against small tweaks to the names of things on the back end. It can cause confusion to programmers who already know the systems, and as we see here, it's not necessarily clear which is the better name anyway. It mostly just confuses the git blame without changing anything.

Git blame, I wouldn't take it as a good enough argument.

Btw, 3 people for unobscured, the `1`s are me setting up for mobile.

You would recommend leaving it there. Including the FIXME? Revert everything but the grammar fix?

I-am-Erk:
That's my 2 cents yes. You could change that comment "contrary to its name" to be clearer eg. "Unimpaired_range() returns the amount the character can see clearly once all impairments are taken into effect" or something.

Maleclypse:
Moving this to draft until resolved to prevent accidental merging.

Brambor:
I reduced it to the docs changes and it is ready for review / merge

Brambor:
Basic tests passed, I am not gonna rerun this to check if my comment changes broke something 🤣

PR #67936, opened 2023-08-28
Character creation: small screen fixes

### LINKED ISSUE #67777: [Suggestion] Add way to scroll on stats screen.

Now that stats display descriptive effects on game play now and the fact that I play on a Chromebook most of the time nowadays an issue became apparent to me, there is no way to scroll the stats screen. This is not a huge issue, but when you have a smaller screen such as my Chromebook or phone, there's a cutoff by the screen which you can't see past.

### Solution you would like.

I would like to be able to scroll the info on the stats screen like you can with the profession and scenario tabs.

Brambor:
#### Summary
Interface "Character creation: small screen UI fixes"

#### Purpose of change
Fix #67777

POINTS
image

STATS
image

TRAITS
image

SKILLS
image

DESCRIPTION
image

visible:
- [x] fix: SKILLS The help text doesn't update when keybinding is changed.
- [x] fix: help overflow (POINTS, STATS, SKILLS, DESCRIPTION)
- [x] fix: scrollable details (STATS, TRAITS)
- [x] fix: There should be space between the border and the text. (SKILLS)
- [x] fix: manny missing `┬`

non-visible:
- [x] fix: DESCRIPTION "switch gender" help string wasn't visible
  - but I removed it since it takes up precious space on small screens
  - position was hardcoded, so adding was hard, the position isn't hardcoded anymore
- [x] fix: SKILLS scroll_up or _down would trigger again if it scrolled on a header (in character creation)
  - It now moves only by one extra line
- [x] fix: scroll_up or _down could wrap, but shouldn't (in character creation on the SKILLS tab & in the @ menu in-game)

not fixed:
- [ ] SKILLS Cannot scroll up to view "Melee weapons" as it is not selectable entry in a list
  image  

- [ ] DESCRIPTION
  image
  image

- [ ] Everywhere Summary overflow
  image

#### Describe the solution
Apply the style of SCENARIO, PROFESSION, and BACKGROUND to TRAITS, and SKILLS descriptions.
Renamed variables and moved code to make it more clear that the code is the same.
Unify `player_display.cpp` & `newcharacter.cpp` skill listing into new `skill_ui.h`.
Detect DESCRIPTION's help height and make as big a window as needed to be.

Named some constants.

Less renaming & code movement.
Put `const int iHeaderHeight = 6` to global scope instead of defining it 6 times across different functions.

Maybe, make DESCRIPTION one window that can be scrolled. And move things around, so that everything with potentially
more entries is on that window. Including when the window isn't wide (2 cols).

#### Testing
On refactoring, switch between new and master .exe and check that the display is the same for all values.

Scroll things until you find something that looks weird.

### DIFF
--- a/src/newcharacter.cpp
+++ b/src/newcharacter.cpp
@@ -970,12 +970,20 @@ static const char *g_switch_msg( const avatar &u )
 
 void set_points( tab_manager &tabs, avatar &u, pool_type &pool )
 {
+    const int iSecondColumn = 31;
+    const int iHeaderHeight = 6;
+    // guessing most likely, but it doesn't matter, it will be recalculated if wrong
+    int iHelpHeight = 3;
+
     ui_adaptor ui;
     catacurses::window w;
     catacurses::window w_description;
     const auto init_windows = [&]( ui_adaptor & ui ) {
+        const int freeWidth = TERMX - FULL_SCREEN_WIDTH;
+        isWide = freeWidth > 15;
         w = catacurses::newwin( TERMY, TERMX, point_zero );
-        w_description = catacurses::newwin( TERMY - 11, TERMX - 35, point( 31, 6 ) );
+        w_description = catacurses::newwin( TERMY - iHeaderHeight - iHelpHeight - 1,
+                                            TERMX - iSecondColumn - 1, point( iSecondColumn, iHeaderHeight ) );
         ui.position_from_window( w );
     };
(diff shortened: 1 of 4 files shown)


Brambor:
I didn't fix everything I set out to, but I consider it finished now.

The rest is too big of an overhaul for too little gain.

Brambor:
restarting tests

PR #67952, opened 2023-08-30
[Magiclysm] Add everburning torch to cooking requirements list

Standing-Storm:
#### Summary
Mods "Add everburning torch to cooking requirements list"

#### Purpose of change

Someone on Discord mentioned it, and while it would be a little inconvenient to cook with you could certainly do it.

#### Describe the solution

Add the everburning torch (lit) to the surface heat and water boiling requirements list.

### DIFF
--- a/data/mods/Magiclysm/requirements/cooking_components.json
+++ b/data/mods/Magiclysm/requirements/cooking_components.json
@@ -3,5 +3,15 @@
     "id": "meat_raw_steak",
     "type": "requirement",
     "extend": { "components": [ [ [ "purified_meat", 1 ], [ "impure_meat", 1 ] ] ] }
+  },
+  {
+    "id": "surface_heat",
+    "type": "requirement",
+    "extend": { "tools": [ [ [ "mtorch_everburning_lit", -1 ] ] ] }
+  },
+  {
+    "id": "water_boiling_heat",
+    "type": "requirement",
+    "extend": { "tools": [ [ [ "mtorch_everburning_lit", -1 ] ] ] }
   }
 ]


Maleclypse:
I thought the everything torch turned off when dropped or does it start flammable fires when dropped?

Standing-Storm:
It's the latter, I think. It has the flags 

> "FIRE", "LIGHT_240", "FLAMING", "TRADER_AVOID", "WATER_EXTINGUISH"

...the same as the regular torch minus that one's BURNOUT and CHARGEDIM flags. A regular torch can't be used for cooking because the heat is lower and it only burns for an hour or so, but this is a magical torch that burns forever so inefficiency and extinguishing isn't a concern. 

Is your argument that without a way for tool qualities to represent efficiency, a torch shouldn't provide the same cooking time as a campfire/bonfire?

Zireael07:
I think his argument is that it only burns in your hand, so when you put it down (to cook) it's no longer burning.

Standing-Storm:
I really thought it still burned when you dropped it like the finger firelighter.  I'll have to check later

Though if it does go out, I would argue you should still be able to boil water manually using it (hold pot with water in one hand, hold torch in the other) even if not passively.

Maleclypse:
> I really thought it still burned when you dropped it like the finger firelighter. I'll have to check later
> 
> Though if it does go out, I would argue you should still be able to boil water manually using it (hold pot with water in one hand, hold torch in the other) even if not passively.

That’s the other problem. You’d have to rig up some way for it to heat a surface. If it’s torch sized it’s over a foot long. How will you hold it under what you want to heat? If you hold it sideways now the heat is in your face close to your body unless you have the pan extended way out so now you are cooking with both arms extended so how do you flip/stir whatever food stuff your cooking. And if you don’t stir then things just end up both burned and half cooked.

Standing-Storm:
Yeah, my idea is mostly predicated on the idea that it stays lit when you put it down and then you treat it like a magical campfire. If it goes out when you put it down (which I think is actually a reasonable safety measure and honestly surprising considering wizards are usually out there doing things like creating owlbears) then I agree cooking doesn't work.

rty275:
I know it does not currently go out if dropped, that's how I usually set up my infinite light in my workshop bases.

Maleclypse:
Ok ideally it would go out when dropped or set things on fire.  One option is to make it go out when dropped but add a construction for like a wall sconce or something it can be added to in order to not go out.  I realize that would be scope creep for this PR. Thoughts?

casswedson:
#62129 by the way

Standing-Storm:
You're right. Well, I don't know what those Discord people were talking about: 
image

I'll close this and move on to my idea for everburning torch furniture (letting you re-create those candelabras or put torch sconces on the walls like Maleclypse suggested).

PR #67955, opened 2023-08-30
Object creator merge windows into a tabbed view

snipercup:
#### Summary
Interface "Merge Object Creator windows into a tabbed view"

#### Purpose of change
To be able to switch between editors more easily. Before this change, you had to re-open the object creator to pick a different window. 

#### Describe the solution
- Remove everything from the main window and replace it with a tabbed view
- Have the separate windows be a tab in the tabbed view
- This is mostly just a functional change. I basically didn't do anything on the existing UI elements

- Instead of a tabbed view, I tried to do a menu bar with a 'navigation' option. When a option from the navigation menu was selected, it would open the window corresponding to the selected menu item. You could have 1 window open at a time (like the tabbed view) but the new windows would change position and size and it's not as convenient as the tabbed view.
- In the long term, I'd like it to work like a web browser, where you have your tabbed view and it lets you pop-out one of the tabs so you can get the most out of your multi-screen setup.

#### Testing
On windows and linux:
- Checked that all the controls are functional
- Mods are loaded in correctly

On windows:

Spell tab:

image

Item group tab:

image

Mod selection tab:

image

On Linux:
[Screencast from 30.08.2023 20:43:24.webm](https://github.com/CleverRaven/Cataclysm-DDA/assets/50166150/01425f35-cdec-4650-a892-6f1a5cafb752)

#### Additional context
I looked back at KorgGent's work but couldn't find anything about the broaded design of the Object Creator itself. I have some ideas, but if you think about what the Object Creator can do for us and how it will grow, I feel like it will be a good idea to come up with some good design concepts. If someone wants to exchange ideas I'm all for it.

### DIFF
--- a/object_creator/creator_main_window.cpp
+++ b/object_creator/creator_main_window.cpp
@@ -5,14 +5,13 @@
 #include "mod_selection_window.h"
 #include "enum_conversions.h"
 #include "translations.h"
-#include "worldfactory.h"
-#include "mod_manager.h"
 
 #include <QtWidgets/qapplication.h>
 #include <QtWidgets/qmainwindow.h>
 #include <QtWidgets/qpushbutton.h>
-#include <QtCore/QSettings>
 #include <QtWidgets/qstackedwidget.h>
+#include <QtWidgets/qmenubar.h>
+
 
 namespace io
 {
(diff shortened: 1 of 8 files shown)


Maleclypse:
It looks very cool

PR #67959, opened 2023-08-31
It gets hot (Overheat mechanics for energy guns)

John-Candlebury:
#### Summary
Features  "Add overheat mechanics to energy guns."

#### Purpose of change

Energy weapons, with their lack of recoil and generally deep magazines are pretty hard to balance properly, especially as they become more powerfull, since the player can just stand somewhere and turn into a deadly accurate floodlamp of death after a few seconds of aiming.  Overheating is a good way to limit this ability.

Also anything close to a combat pulse laser would make tons of heat as I understand it.

#### Describe the solution

Guns can now define the fields `heat_per_shot` `overheat_threshold` and `cooling_value`  and the faults included here to display safe or potentially deadly overheat effects (which are mostly explained in the fault descriptions themselves). Overall, they can be used to create the standard overheat events of FPS's, cheap guns that melt during sustained fire,  and 40K style exploding plasma rifles.

Includes a demo Plasma gun for Aftershock and the vanilla lasers have been expanded to also overheat (mostly safely).

#### Testing
Used both overheat prone guns and normal ones.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -6420,9 +6420,9 @@ std::string item::overheat_symbol() const
         case 3:
             return string_format( _( "<color_light_red>\u2585HEAT </color>" ) );
         case 4:
-            return string_format( _( "<color_red>\u2587HEAT! </color>" ) );
+            return string_format( _( "<color_red>\u2587HEAT! </color>" ) ); // NOLINT(cata-text-style)
         case 5:
-            return string_format( _( "<color_pink>\u2588HEAT!! </color>" ) );
+            return string_format( _( "<color_pink>\u2588HEAT!! </color>" ) ); // NOLINT(cata-text-style)
         default:
             return "";
     }
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -138,6 +138,8 @@ static const efftype_id effect_shakes( "shakes" );
 static const efftype_id effect_sleep( "sleep" );
 static const efftype_id effect_weed_high( "weed_high" );
 
+static const fault_id fault_overheat_safety( "fault_overheat_safety" );
+
 static const furn_str_id furn_f_metal_smoking_rack_active( "f_metal_smoking_rack_active" );
 static const furn_str_id furn_f_smoking_rack_active( "f_smoking_rack_active" );
 static const furn_str_id furn_f_water_mill_active( "f_water_mill_active" );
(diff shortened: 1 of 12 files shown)


Raikiri:
Do vanilla laser rifles now overheat or does the change only affect them if aftershock is enabled? I surely hope it applies to vanilla.

Profugo-Barbatus:
Love the sound of this. Now my energy weapons get downgraded to 'free, constrained problem solver' and I still break out the loaded mags and battle rattle if I'm planning to try and get into a mall or the like.

src/item.cpp
```diff
@@ -6397,6 +6402,32 @@ std::string item::dirt_symbol() const
...
+    switch( std::min( 5, static_cast<int>( get_var( "gun_heat",
+                                           0 ) / ( type->gun->overheat_threshold ) * 5.0 ) ) ) {
+        case 1:
+            return "";
+        case 2:
+            return string_format( _( "<color_yellow>\u2583HEAT </color>" ) );
+        case 3:
+            return string_format( _( "<color_light_red>\u2585HEAT </color>" ) );
+        case 4:
+            return string_format( _( "<color_red>\u2587HEAT! </color>" ) );
```
ehughsbaird, line 6423: I believe this is how you can disable that.
```suggestion
            return string_format( _( "<color_red>\u2587HEAT! </color>" ) ); // NOLINT(cata-text-style)
```

ilovehans10:
This makes me think of a mod for another game back in the day (might have been fallout), because they had a similar idea with a little twist.

They had a similar solution to this where laser weapons get hot, but they took it a step further and made it so that before the weapon would totally shut down it would first become inaccurate from heat. The reasoning was that the heat would warp focusing crystals and parts of the gun just enough to make it less accurate before it would just outright stop working. Idk how feasible that idea is, but I thought I should bring it up.

SurFlurer:
Do you think RMES and CMES from mechs are immune from overheating? I guess they should also suffer from it.

John-Candlebury:
Left them out since they have a whole mech attached that can potentially handle that better than just making the gun handheld. 

Afaik you cant fire them independently so I didnt consider adding overheats necessary.

SurFlurer:
> Afaik you cant fire them independently so I didnt consider adding overheats necessary.

 They can be installed on a turret, making an extremely efficient killing machine.

PR #67965, opened 2023-08-31
Add splash screen to Object Creator

snipercup:
#### Summary
Features "Add spash screen to Object Creator"

#### Purpose of change
Show to the user that the application is loading. Before this change nothing would show up until the application is fully loaded.

#### Describe the solution
It's a grey rectangle with black text in it that says 'Initializing Object Creator' and it is visible while the application is starting. There is a second splash screen that shows when the main window is showing. It disappears when the application is loaded. It's not much but it's better then nothing.
Removing 1 of the splashscreens will mean that the user sees nothing instead of that splash screen. The two splash screens cannot overlap that is why we need both of them.

Also, as part of the troubleshooting I changed the declaration of the itemgroup window and the spell window to be in line with the mod selection window. 

Show a image splash screen instead of a plain text one. I even had an image ready:
image

The problem is that I don't know where to put it on the filesystem. The /object_creator/ folder is not there in the release version of the game. All the other folders do not seem an appropriate place to put it.

Another alternative is a loading bar or at least a % completed, but I lost the tutorial for it. 

#### Testing
On both Windows and Linux:
- Start the object_creator
- See splash screen:
image

- See the second splash screen before the main menu loads:
image


ashGlaw:
What is Object Creator?

snipercup:
It's a tool you can use instead of a text editor to modify JSON files. You can see what it looks like in #67955

snipercup:
It doesn't quite work as well on Linux (ubuntu). I made some changes and now it shows the splash screen briefly but hardly long enough. This will need some more work.

GuardianDll:
The actual big problem of object creator is that people, who can use it, unable to obtain one because it required to be compiled
and people who can compile it don't need to use it already

snipercup:
> The actual big problem of object creator is that people, who can use it, unable to obtain one because it required to be compiled and people who can compile it don't need to use it already

I agree and that should change. I have some thoughts on this. Do you mind if we take the discussion to Discord? My DM's are open.

snipercup:
Now the .sln build fails because I imported unistd. I will have to find a fix for that now. Do not merge (I removed the draft status too soon)

snipercup:
Thanks for re-running the test. I checked it and it fails on something related to eternal_season. That has nothing to do with this PR. I consider this PR finished, feel free to merge.

snipercup:
Thank you for merging. In the future if someone looks back on this and wants to improve it, we can:
- Increase the font size and colors of the text. We'd need to go pack to a painter like in one of the early commits of this pr (easiest)
- Find a way to ship an image with the object creator itself so we can show that image instead
- Add a loading % to the splash screen. The trouble here is that it's hard to tell what the progress is if you're waiting for one big function to finish
- Get rid of the splashscreen by instantly showing the main window and then indicate that the components in the window are loading but this would be very difficult.

PR #67982, opened 2023-09-01
[DinoMod] more interesting behavior part 2

LyleSY:
#### Summary
Mods "[DinoMod] use EAT_CARRION"

#### Purpose of change

Make dinosaurs seem more alive and realistic and interesting

#### Describe the solution

Applies new EATS_CARRION special attack from #66294 and the EATS flag and new stomach_size entry to all except the tiniest meat eating dinos. Adds to pterosaurus and mosasaurus also.

N/A

#### Testing

Game loads no errors. Monsters spawn and behave as expected. Predators continue to murder as long as they can murder, only stopping to eat when all possible targets are gone. Not ideal for monsters attacking to eat, but definitely better.

Screenshot 2023-09-03 at 8 37 24 PM

#### Additional context

N/A


LyleSY:
@fairyarmadillo I didn't see stomach_size values documented so the values are a best guess based on monster weight. Am I in the right ballpark?

fairyarmadillo:
> @fairyarmadillo I didn't see stomach_size values documented so the values are a best guess based on monster weight. Am I in the right ballpark?

Correct. Presently.they and the amount tbe monsters are actually consuming is totally arbitrary.

PR #66532, opened 2023-06-29
More cars!

randomtyper:
#### Summary
Content "Adds new cars, big and small"

#### Purpose of change

Expanding the variety of cars in game. Generic cars that are currently in game are almost exclusively station wagons with changed bells and whistles and small trunks.

#### Describe the solution

Car parts changes:
- [x] Add sedan trunk;
- [x] Increase the size of station wagon (walk-through) trunk;
- [x] Increase storage size of hatchback hatch.

Rename old car into station wagon.

Add more cars!

- [x] Generic sedans (diesel, gas, engine variations);
- [x] Microcars (like Smart, variations: gas, hybrid, electric);
- [x] Generic coupes (diesel, gas, v8 muscle cars, v12 sport cars, v6 sport cars);
- [x] Supercompacts (gas (?), hybrid, electric, tandems, wheel amount variations);
- [x] Town cars (5-tile wide sedan monstrosities, v6-v8);
- [x] Shooting brakes (sporty coupe station wagons);
- [x] Trunk-forward sportcars;
- [x] Three-wheeled motorbikes (like those made by Can-Am).

- [ ] Addition of all those new cars into spawngroups and into spawns.

#### Describe alternatives you've considered

Living with the utter lack of variety among the in-game cars despite the amount of RL car types and subtypes.

#### Testing

None for now.

#### Additional context

Sedan trunk sizes:
https://www.edmunds.com/sedan/before-buy.html#:~:text=Subcompacts%20offer%2010%2D12%20cubic,large%20sedans%20offer%2017%2D21.

Station wagon trunk sizes (about 600 L):
https://www.autoexpress.co.uk/ford/focus/estate/practicality
https://www.autoexpress.co.uk/volkswagen/golf/estate

Hatchback trunk sizes (about 400 L, and those are medium ones, can reach into 500s):
https://www.whichcar.com.au/car-advice/boot-sizes-of-best-selling-hatchbacks
https://www.carbase.co.uk/news-and-features/used-car-buying-guide/best-family-hatchbacks-with-big-boots/

PR #67986, opened 2023-09-02
More cars strikes back

GuardianDll:
#### Summary
Content "Added more vehicles into the game"
#### Purpose of change
Picking up #66532, because author do not respond, and i don't like to left it abandoned
Closes #66532 (sorry, mate, i just like this PR a lot!)
#### Describe the solution
all same as previous PR, but worse:
vehicle groups was simplified - author tried to use nested vehicle groups, which doesn't exist as a feature; now all groups just got straight added new cars
removed solar version of new cars - right now amount of cars that has solar roof is not that big to represent it in any meaningful way
Author come back, fix it's own PR for themselves, everybody happy
#### Testing
Surprisingly good, everything works as intended


turco-ai:
Couldn't rare cars be put in the car showroom as a rare spawn? Seems like the perfect place for them, instead of being removed.

PR #67047, opened 2023-07-20
Concrete Mixer audit

Karol1223:
#### Summary
None

#### Purpose of change
Our tiny tabletop concrete mixer didn't really fit with realism that well

#### Describe the solution

<details>
<summary>CONCRETE MIXER ITEM CHANGES</summary>

- Weight increased from ~9kg to ~41kg
- Volume increased from 5l to 250l
- Added a manually defined longest side of 118cm
- Decreased pre-apo price from 500$ to 400$
- ``to-hit`` values now use modifiers
- Material now has 50% plastic (based on a concrete mixer with a plastic drum)
- Added an open pocket that can be used to store things
- Removed the ``HOTPLATE`` use action because why the fuck can we heat up lunch in a concrete mixer
- Made it an unfoldable vehicle and added a pseudotool - it now can only be used when unfolded into vehicle form

</details>

<details>
<summary>CONCRETE MIXER RECIPE CHANGES</summary>

- No longer takes raw steel, 2l metal tank, or heating elements
- Will now take 2 small wheels and a 30 gallon plastic barrel
- Tripled the number of pipes needed
- No longer has metalworking proficiency
- Will now take an extension cord instead of a bunch of loose cables
- Will take a small motor instead of a tiny motor
- Added some nuts and bolts as per the user manual of the one I'm basing this off
- Added some small metal sheets (for the blades to mix the concrete as well as to cover the motor)
- The weld length has been recalculated and increased over four times
- Now needs a metal sawing quality of 2 instead of 1, multitool won't cut it

</details>

<details>
<summary>30 GALLON PLASTIC BARREL CHANGES</summary>

- Volume increased from ~118l to ~165l
- Pocket volume increased slightly (by ~1l)
- Added a manually defined longest side of 72cm
- Increased pre-apo price from 50$ to 71$
- Renamed to clearly show it's made of plastic
- ``to-hit`` values now use modifiers
- Recipe fundamentally changed to be melding the plastic barrel instead of duct-taping plastic jerrycans together (?)
</details>

<details>
<summary>MISC CHANGES</summary>

- Concrete mixer can no longer be attached to the modular rigs (vehicle & appliance)

</details>

#### Testing
~~Although most of the things work as expected, the PR is blocked by #71108~~
No longer blocked

#### Additional context
The recipe isn't perfect and generates roughly 7kg of mass, but I cannot make it better without adding bullshit requirements.

PR #15744, opened 2016-03-11
[CR]Vehicle part weight/efficiency adjustments

Coolthulhu:
Changes to batteries, engines and frames.

Batteries got a giant nerf - their capacities got cut by an order of magnitude, their weights got increased with no compensation to their stats at all.

| Battery | Capacity | Weight in kg | Capacity-per-kg |
| --- | --- | --- | --- |
| Motorbike | 2k | 6kg | 333 |
| Car | 4.5k | 18kg | 250 |
| Truck | 8k | 30kg | 267 |
| Small storage | 0.5k | 2.5kg | 200 |
| Medium storage | 7k | 31kg | 226 |
| Big storage | 40k | 150kg | 267 |

Storage batteries also got their volumes increased, but that isn't too relevant for vehicles which ignore volumes anyway.

Electric engines got buffed to compensate:

| Engine | Power | E-power | E-power per power |
| --- | --- | --- | --- |
| Small | 20 | 5000 | 250 |
| Medium | 100 | 20000 | 200 |
| Big | 400 | 120000 | 300 |
| Enhanced | 500 | 200000 | 400 |

Frames got rebalanced (mostly nerfed).

| Frame | Durability | Weight in kg | Durability per kg |
| --- | --- | --- | --- |
| Handle | 30 | 1.25 | 24 |
| XLight | 150 | 5 | 30 |
| Frame | 400 | 20 | 20 |
| HDuty | 1200 | 60 | 20 |
| Foldable XL | 50 | 5 | 10 |
| Wooden | 150 | 8 | 18.8 |
| Light wooden | 100 | 4 | 25 |
| Foldable wooden | 15 | 4 | 3.75 |

Light frames lost a small bit of HP, regular frames got a big increase in weight, heavy duty frames almost doubled in weight.

Resulting vehicles have weights more comparable to the IRL ones.

Other adjustments:
- Solar panels lose ~60% of their efficiency. Except quantum ones, those lose ~46% instead
- Removed AISLE from cargo holder. It can't have both AISLE and volume 1000 of stuff on it
- Increased cargo holder weight with heavy frame. Now it weights 80 kg and is not as much of a no-brainer.
- Made bone and chitin armor kits cheaper, easier to craft, lighter and generally better to make them more sensible to craft to armor very light vehicles
- Made water purifier take "anywhere" instead of "center"
- Decreased reactor install difficulty to 6 (from 8). Now that batteries aren't amazing, it might see some use.
[...]

PR #67987, opened 2023-09-02
Pipe audit

Karol1223:
#### Summary
None

#### Purpose of change
So we all like pipes, yes? Ours were weird. Double the expected weight, for the record.

The last changes to pipe numbers I am aware of are #47825. And there's a source, too. This is good because it allowed me to look for a specific set of pipes for numbers (60cm, 28mm diameter, 1.2mm wall thickness). Now, I didn't find anything matching the wall thickness so instead went for 1.5mm, which is close enough (?) but the issue at hand was... weight. Even with 1.5mm thickness, which is MORE than what the old PR assumed, the weight came out at 600g, while ours were 2x+ of that. So here we are.

Since @RenechCDDA is dying to prove me wrong and that our pipes indeed should weigh over 1kg:
- Calculating two cylinders: first with 60cm height and 28mm diameter (369.45ml), second with 60cm and 25mm diameter, to remove wall thickness that is 1.5mm x 2 (294.52ml)
- Substract the second cylinder volume from the first: 369.45ml - 294.52ml = 74.93ml
- Multiply volume by the density of steel we assume as a material (8g/ml): 74.93ml * 8g/ml = 599.44g

#### Describe the solution
- Changed the weight of pipes and all pipe-based items (sharpened pipe, pipe maces, pipe spear, pipe staff, bundle of pipes, makeshift crowbar; **NOT PIPE FITTINGS!**)
- Changed the pre-apo price for the pipe to match the reference, changed the pre-apo price for the bundle of pipes to be 10x that instead of just 1 dollar
- Changed the lengths of some pipe-based weaponry to be *number of pipes used x 60cm* because not all applied to that (this lowers it for any affected, for the record)
- Changed the volume and weight of the makeshift crowbar to... actually make sense. I mean, look - it for some reason lost over 2/3 of its volume because it got hammered on the sides? Nah. Same with the weight. Also got ``FRAGILE_MELEE``
- Edited the pipe disassembly recipe to bring it up to speed

**DPS CHANGES:**
| NAME | PRE-AUDIT | POST-AUDIT |
|--------|--------|--------|
| Pipe | **30.40** | **30.08** |
| Sharpened Pipe | **60.39** | **59.61** |
| Basic Pipe Spear | **60.08** | **61.49** |
| Pipe Spear | **70.13** | ~~**69.52**~~ idk anymore I lost this absurd testing character, higher now to please the tests |
| Pipe Staff | **32.98** | **33.15** |
| Pipe Mace | **41.20** | **40.22** |
| Great Pipe Mace | **41.38** | **41.86** |
| Makeshift Crowbar | **40.68** | **32.03** |

^ These values are bullshit, by the way. They're not the expected DPS of these weapons. I have no idea why my test character gets this sort of insane DPS. And you know what? I won't question it. I don't need exact same DPS values as the tests, I just need to know how much I'm changing them in comparison.

I briefly considered obsoleting the makeshift crowbar altogether but it was rightfully pointed out to me the issue isn't its existence. The issue (if any) is how useful it is. And I don't wanna touch that. Scary

#### Additional context
1. Makeshift crowbar and sharpened pipe, despite using a single pipe, have a slightly lower weight with the assumption that some material got chipped off in the process of creating it
2. Some weaponry that used pipe fittings had their weight lowered to not use 100% of the weight of the pipe fittings - our pipe fittings weigh 0.5kg for some reason and you're definitely not using 0.5kg of pipe fittings to hold 3 pipes together
3. Wanna hear a fun fact? I did this to help me handle the weight gap in the recipe on #67047... and I made it worse...


GuardianDll:
for pipe weight actual person, that made it 1.25 kilo is Erk, in #32218. No reasoning was provided

Karol1223:
> for pipe weight actual person, that made it 1.25 kilo is Erk, in #32218. No reasoning was provided

I don't believe that's right. The pipe already weighed 1.25kg in that PR, Erk just moved it to a different file. Or am I reading it wrong?

Aerin-of-the-Toast:
What's the new attack speeds for the pipe and the pipe weaponry? I'm a bit worried this will make the already strong pipe weapons into DPS powerhouses

Karol1223:
> What's the new attack speeds for the pipe and the pipe weaponry? I'm a bit worried this will make the already strong pipe weapons into DPS powerhouses

I'll be real I was hoping they were in the DPS test so I could adjust them accordingly but now I'm kinda fucked. I'll have to manually compare them.

GuardianDll:
> I don't believe that's right. The pipe already weighed 1.25kg in that PR, Erk just moved it to a different file. Or am I reading it wrong?

oops, my bad, actual pipe weight goes back to #15744, when it was adjusted to sane frame weight

Karol1223:
> oops, my bad, actual pipe weight goes back to https://github.com/CleverRaven/Cataclysm-DDA/pull/15744, when it was adjusted to sane frame weight

I really hate that we use regular pipes as vehicle parts. Really complicates things, given that they had to have a weight adjustment solely for this reason and even in this PR I had to fight against the automated vehicle tests...

PR #67990, opened 2023-09-02
Adds Karate manual to Novice Martial Artist starting items

Hymore246:
#### Summary
Balance "Adds Karate manual to Novice Martial Artist starting items"

#### Purpose of change

Adds Karate manual to Novice Martial Artist profession

#### Describe the solution

This change makes the Novice Martial Artist start the game holding The Shōtōkan Karate Handbook. This allows them to learn Karate by successfully reading the book.

Novice Martial Artist offers very little to players. Its skills are negligible and its subpar gear makes it arguably worse than a default Survivor. The profession needs something to make it stand out and giving it a martial arts manual at the start and increasing its cost to one point will do just that. This change adds more depth the Novice Martial Artist and gives players a reason to actually use the profession instead of ignoring it in favor of taking a martial arts trait instead. This may seem like a lot at first glance since martial arts usually cost 3 points to take at chargen but consider the following:

* You cannot choose which martial art you want. It must be Karate.
* You need to read the book to learn the style. Depending on your starting location this could be easy or very difficult.
* Your lack of storage means you need to be careful with the manual before you learn the style. You might find yourself in a situation where you need to drop the book to keep yourself alive.
* Even after you learn the style, you still need to learn unarmed and melee skills to use it at full strength.

If the cost is too low, it can be increased to two points. Higher than that you would be better off just taking a martial arts trait.


oosyrag:
Why karate? And why would they have a book?

Seems like it would serve the profession better if it were to actually start in a dojo as described, maybe move it to a specific scenario.

Hymore246:
Karate was selected as the style because it is a common and well known martial art. From a gameplay perspective, it is a hybrid style that can be used from beginning to end game without becoming useless. 

As for why they have the book that could be because they were bring it to their class in the hopes of learning the "cool moves" in the manual. This makes sense even for non-karate classes since a novice might not know the difference between Karate and another style. This is just one of the possible roleplay reasons the profession might have.

ampersand55:
Maybe we can change the name to "karate white belt" or similar to more precisely reflect the profession?

Fris0uman:
I don't see any issue with the proposed change. Just a quick reminder though that points are going away and that there's no point in fiddling with it.

Fris0uman:
I think you need to update some tests, looks like the blackbelt_novice profession is used in Tests

Hymore246:
> I think you need to update some tests, looks like the blackbelt_novice profession is used in Tests

Judging from the errors, the game can't give an inhaler to the profession because it doesn't have inventory space. I assume it was putting into the character's hands previously. So, how is this corrected for professions that intentionally cannot carry any more items?

PR #67994, opened 2023-09-03
Updated camp help to reflect removal of bed restrictions on expansion

PatrikLundell:
#### Summary

None

#### Purpose of change

As per the title.
It was pointed out this hadn't been done when the restrictions were lifted.

#### Describe the solution

Update the description texts to match the current state.

#### Testing

Started the game with a save and asked the companion about camps to look at all the descriptions to see that they seem to be OK. Cannot guarantee the set available in the game is the one in the file, though...
Also noted a formatting error on point 7 of the changes topic, so I added that blank character without actually testing it afterwards.

#### Additional context

- A number of camps talk about immediate access to cooking. Is that still true after the ability to contain fire on standard stoves was removed?
- Does the sealab base camp still exist? I don't recall seeing it on my trips through facilities.


PatrikLundell:
I'll disregard the dictionary complaint about "blacksmithy" as that's not something I've added.

Issue #15854, opened 2016-03-21
"access_cache called with invalid z-level" on savegame load, corrupted minimap


mutability:
(on 26c1f77055b6817ddc762e548c5e51528d99e843)

Loading a save reliably produces "DEBUG: access_cache called with invalid z-level" while loading, and the initial minimap is corrupt (just garbage).

gdb says this is the callpath:

```
(gdb) bt
#0  map::access_cache (this=0x11fc0f90, zlev=-247598940) at src/map.cpp:8065
#1  0x000000000055769c in cata_tiles::draw_minimap (this=0x19633e0, destx=1424, desty=608, center=..., width=width@entry=360, height=height@entry=352) at src/cata_tiles.cpp:1229
#2  0x0000000000e6563f in curses_drawwindow (win=0x1bcd85e0) at src/sdltiles.cpp:800
#3  0x0000000000614289 in wrefresh (win=<optimised out>) at src/cursesport.cpp:267
#4  0x00000000006ba9be in game::draw_pixel_minimap (this=this@entry=0x1c03ed70) at src/game.cpp:5074
#5  0x00000000006fbb35 in game::draw_sidebar (this=this@entry=0x1c03ed70) at src/game.cpp:5213
#6  0x00000000006ff2d0 in game::draw (this=this@entry=0x1c03ed70) at src/game.cpp:5056
#7  0x0000000000734c4b in game::load (this=this@entry=0x1c03ed70, worldname="Northway Village", name="#Q3VydGlzIEZlbGRtYW4=") at src/game.cpp:3514
#8  0x0000000000947b12 in game::opening_screen (this=0x1c03ed70) at src/main_menu.cpp:681
#9  0x000000000094189f in main (argc=0, argv=0x7fffffffdbc0) at src/main.cpp:431
```

Looks like fallout from a26fddfea332695e47978d36c3bcd6fb27127086: ter_view_z is only set in draw_ter(), and following that commit draw_sidebar() is called before the first call to draw_ter()

Issue #15842, opened 2016-03-20
Minimap trouble (flickering, distortions)


Zireael07:
When walking around, I see the minimap flicker. At night, the flicker is where I believe Full Night Vision would let me see, every few tiles. During the day, the flicker happens every time I go past a building - the indoor tiles are displayed for a second or two and then hidden by the proper fog of war.

What is going on? Some kind of caching being done too late?

Visible in build g26c1f77. I ran the last version I have downloaded to check, nothing such in g0c77ec5. This means whatever happened, happened in the last three days.

Zireael07:
![cata distorted minimap](https://cloud.githubusercontent.com/assets/4525388/13906498/cec72fe6-eed7-11e5-8d65-c189d4a4a9c6.png)

AArgh. See the screen. It's definitely more than just indoor tiles being problematic.

-> Black block on the right hand part
-> Stripes above the house at the top of the minimap

codemime:
I've turned on the minimap and can confirm that. I have the same issue.

lucianposton:
a26fddfea332695e47978d36c3bcd6fb27127086 is the first bad commit according to bisect.

a26fddf  2 days ago  Rearange main rendering method to place curses cursor on @ at the end.

codemime:
Reversing #15828 solved the issue for me.

Beanzy:
#15828 Also breaks the UI refresh when I use SDL ASCII with non default fonts. Prior to #15828 there was no issue.

![cdda issue](https://cloud.githubusercontent.com/assets/15885173/13907707/c8e66c2a-eeca-11e5-87b8-9f47a5624bcd.PNG)

In the picture above the advanced inventory screen was opened and then closed, however the image still remains on screen.

DanmakuDan:
It looks like with the rearrangement of the draw cycle, the minimap is getting drawn with the previous submap offset before the game updates the location where the player is. It looks like it also breaks the minimap when traveling between floors.

kevingranade:
I think there is some caching weirdnwss, but I also see some issues caused
by game::ter_view_z not being initialized

DanmakuDan:
Looks like the minimap draw was implicitly depending on `void map::draw( WINDOW* w, const tripoint &center )` inside `draw_ter()` updating the visibility cache (line 5849 in map.cpp):
```
    visibility_variables cache;
    update_visibility_cache( cache, center.z );
```

[...]

PR #67997, opened 2023-09-03
Fix visible tiles revealing invisible tiles below

### LINKED ISSUE #66040: 3D vision lets you see unseen tiles below

Rewryte:
#### Summary
Bugfixes "Fix visible tiles revealing invisible tiles below"

#### Purpose of change
The way 3D vision currently handles tile visibility is by calculating visibility variables for the current z-level and have every z-level underneath inherit the same values.  In cases where a tile on the current z-level should be visible but the tile underneath should not, the player will be able to see both tiles due to the non-visible tile underneath inheriting the visibility variables of the visible tile above it.  Correcting this will fix #66040.  Also a prerequisite for the lightmap changes in #67434.

#### Describe the solution
There are two parts to this solution:

1. Instead on updating the visibility cache on the current z-level only, we update the z-levels below as well, limited by the 3D vertical range setting.
2. In the tileset code, pull new variables from the visibility cache on each z-level to determine light level.

Naturally, these make processing visibility somewhat more CPU intensive.  To compensate, map cache and visibility cache are no longer calculated every draw.  Any event that alters these can simply call `map::build_map_cache` and `map::update_visibility_cache` instead.  This involves the removal of the temporary fix introduced in #15868.  The issues targeted by the fix (#15842 and #15854) are not reproducible.

Overall, these changes have a positive effect on draw performance leading to an 11% decrease in draw times.

#### Testing
Compiles ok.
Confirmed that invisible tiles no longer show as visible when viewing visible tiles on a higher z-level.

Before:
before
After:
after

### DIFF
--- a/src/player_activity.cpp
+++ b/src/player_activity.cpp
@@ -240,23 +240,28 @@ void player_activity::do_turn( Character &you )
     }
     // Only do once every two minutes to loosely simulate consume times,
     // the exact amount of time is added correctly below, here we just want to prevent eating something every second
-    if( calendar::once_every( 2_minutes ) && *this && !you.is_npc() && type->valid_auto_needs() &&
-        !you.has_effect( effect_nausea ) ) {
-        if( you.stomach.contains() <= you.stomach.capacity( you ) / 4 && you.get_kcal_percent() < 0.95f &&
-            !no_food_nearby_for_auto_consume ) {
-            int consume_moves = get_auto_consume_moves( you, true );
-            moves_left += consume_moves;
-            if( consume_moves == 0 ) {
-                no_food_nearby_for_auto_consume = true;
+    if( calendar::once_every( 2_minutes ) && *this && !you.is_npc() ) {
+        if( type->valid_auto_needs() && !you.has_effect( effect_nausea ) ) {
+            if( you.stomach.contains() <= you.stomach.capacity( you ) / 4 && you.get_kcal_percent() < 0.95f &&
+                !no_food_nearby_for_auto_consume ) {
+                int consume_moves = get_auto_consume_moves( you, true );
+                moves_left += consume_moves;
+                if( consume_moves == 0 ) {
+                    no_food_nearby_for_auto_consume = true;
+                }
+            }
+            if( you.get_thirst() > 130 && !no_drink_nearby_for_auto_consume ) {
+                int consume_moves = get_auto_consume_moves( you, false );
+                moves_left += consume_moves;
+                if( consume_moves == 0 ) {
+                    no_drink_nearby_for_auto_consume = true;
+                }
             }
         }
-        if( you.get_thirst() > 130 && !no_drink_nearby_for_auto_consume ) {
-            int consume_moves = get_auto_consume_moves( you, false );
-            moves_left += consume_moves;
-            if( consume_moves == 0 ) {
-                no_drink_nearby_for_auto_consume = true;
-            }
-        }
+        // Also update map and visibility caches every 2 minutes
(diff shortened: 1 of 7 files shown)


Rewryte:
This will be in draft until optimizations are complete.  Currently seeing a 57.6% increase in draw times.

Rewryte:
Currently fixing visual bugs resulting from the removal of map cache and visibility cache calculations from `game::draw`

Maleclypse:
Is this intended to still be in draft?  No stress if yes. It just reads as if it's ready to go possibly.

Rewryte:
> Is this intended to still be in draft? No stress if yes. It just reads as if it's ready to go possibly.

It's more or less done.  Tests are passing on my end (MSYS2).  Just pushed out one more fix to prevent cache bloat during tests.

ehughsbaird:
FYI, this resulted in a massive increase in test run times https://github.com/CleverRaven/Cataclysm-DDA/issues/68154.

src/player_activity.cpp
```diff
@@ -240,23 +240,28 @@ void player_activity::do_turn( Character &you )
...
-                no_drink_nearby_for_auto_consume = true;
+            if( you.get_thirst() > 130 && !no_drink_nearby_for_auto_consume ) {
+                int consume_moves = get_auto_consume_moves( you, false );
+                moves_left += consume_moves;
+                if( consume_moves == 0 ) {
+                    no_drink_nearby_for_auto_consume = true;
+                }
             }
         }
+        // Also update map and visibility caches every 2 minutes
```
ZhilkinSerg, line 261: Why?

Rewryte: Lighting conditions can change while the player is performing an activity, such as a construction that takes place through sunrise.  Now that visibility is no longer calculated every draw, it should be done here for the player to perceive a lighting change during the activity.  Doing so every turn would seem excessive given the speed of activities, so I opted for a calculation every 2 minutes, similar to hunger.

ZhilkinSerg: I don't think it is a good approach. We can update player hunger every 2 minutes, but it is not good for visibility or lighting. Lighting cache should be updated whenever lighting conditions change, visibility cache should be updated when visibility changes - e.g. when lighting condition change, when player moves, when player constructs something, when fields around player change their state, etc.

Brambor: Can't you get when to update from world time implicating light level change throughout sunset / sunrise?
I don't know if it is a good solution, but it would react to the change directly and should be faster if the brightness doesn't change.
It would still be rebuilding unnecessarily when player is underground and doesn't see natural light.

Rewryte: Current plan is to make the conditions more general rather than specific, so mainly position and time.  Time in particular, should cover nearly all the conditions mentioned here.  To keep everything in one place, I would advice that further suggestions be made in #68138.  Thanks.

Issue #67385, opened 2023-08-02
Batch time saving cost for moving stacks of items.


ampersand55:
### Is your feature request related to a problem? Please describe.

Moving large stacks of items have been unrealistically slow and laggy for a long time, but it was made more apparent when items with charges were turned into stack of items. When you craft e.g. sugar, salt, spices or flour in batch you are easily dealing with stacks of several thousands that you need to move around. 

Currently it takes ~50 minutes of in-game time to drop 1000 units of sugar from a duffel bag, which isn't realistic. 

### Solution you would like.

Batch time saving cost for moving stacks of the same item, and handling the dropping of said stack as one action so the game doesn't freeze when you handle large stacks.

Perhaps just a simple square root of dropping time, so if dropping 1 item takes 1 second, then dropping a stack of 10 would take ~3 seconds, dropping 100 would take 10 seconds, dropping 10000 would take 100 seconds, and so on.

### Describe alternatives you have considered.

Make powdered items use charges.

lispcoc:
I'm currently thinking about solving this problem.
I think the idea of ​​making the cost proportional to the square root of the number of items is very difficult to realize, because you don't know how many items you can move until you actually move the items.
Zombies may get in your way while you work, or items may be stolen by NPCs along the way.

Actually, I think the only option is to apply an expression like a linear function that can be calculated sequentially, but do you have any good ideas for this?

ampersand55:
If we want to calculate dropping time for one item at a time we could use a slow-growing non-converging infinite series, such as the harmonic series, where the first drop takes 1 second, the second takes 1/3 seconds, the forth takes 1/4 seconds and so on.

If we want the dropping advantage to be slower we could use e.g. some form of Σ k / (n+k-1) where k is a constant.

items dropped | square root | harmonic series Σ 1/n | Σ 10/(n+9)
---|--:|--:|--:
1 | 1  | 1 | 1
2 | 1.41 | 1.5 | 1.91
10 | 3.16 | 2.93 | 7.19
100 | 10 | 5.19 | 24.44
1000 |  31.62|   7.49 | 46.65
10000 | 100 |  9.79 | 69.60

lispcoc:
Σ 10/(n+9) seems good.
There are some other decisions to be made as well.
- We should be able to save time like this when dropping items, but should we also apply them when picking them up?
- The time to move items from adjacent tiles includes the time to move tiles, but should this cost be applied only once when moving items all at once?

[...]

PR #68004, opened 2023-09-04
[CR]Cost reduction of batch item picking and dropping

lispcoc:
#### Summary
Features "Batch time saving cost for moving stacks of items"

#### Purpose of change
Implement https://github.com/CleverRaven/Cataclysm-DDA/issues/67385

Moving n items takes n times as long as moving one item now.
In reality, no one drops 1 kg of salt on the floor in 1 gram portions.

#### Describe the solution
- Apply the formula suggested on https://github.com/CleverRaven/Cataclysm-DDA/issues/67385.

items dropped | Σ 10/(n+9)
-- | --
1 | 1
2 |1.91
10 | 7.19
100 | 24.44
1000 | 46.65
10000 | 69.60
- Apply the same formula both when picking up and dropping items.
- Movement costs when picking up items on adjacent tiles only apply to the first one.

#### Testing
drop 1000 salts to floor from backpack
Before: 25:00
After: 1:05

pick 1000 salts from floor to backpack
Before: 13:20
After: 0:32

#### Additional context
- Please comment if you have any thoughts on this proposal.
- There may be a more efficient way to examine stacked items.
- There are cases where the cost to move small items becomes 0. Is that appropriate?
- I think it would be better to review the entire process and make it more efficient, but I don't have the time to tackle that. If anyone wants to work on this issue, I'll close the PR and leave it to you.

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -2184,7 +2184,8 @@ void pickup_activity_actor::do_turn( player_activity &, Character &who )
     }
 
     // False indicates that the player canceled pickup when met with some prompt
-    const bool keep_going = Pickup::do_pickup( target_items, quantities, autopickup, stash_successful );
+    const bool keep_going = Pickup::do_pickup( target_items, quantities, autopickup, stash_successful,
+                            batch_num );
 
     // If there are items left we ran out of moves, so continue the activity
     // Otherwise, we are done.
@@ -3709,7 +3710,7 @@ static void debug_drop_list( const std::vector<drop_or_stash_item_info> &items )
 static std::list<item> obtain_activity_items(
     std::vector<drop_or_stash_item_info> &items,
     contents_change_handler &handler,
-    Character &who )
+    Character &who, int &batch_num )
 {
     std::list<item> res;
 
@@ -3723,7 +3724,8 @@ static std::list<item> obtain_activity_items(
             continue;
         }
 
-        const int consumed_moves = loc.obtain_cost( who, it->count() );
+        bool next_batch_ok = it + 1 != items.end() ? can_batch_move( loc, ( it + 1 )->loc() ) : false;
+        const int consumed_moves = loc.obtain_cost( who, it->count(), batch_num );
         if( !who.is_npc() && who.moves <= 0 && consumed_moves > 0 ) {
             break;
         }
@@ -3759,6 +3761,7 @@ static std::list<item> obtain_activity_items(
         } else {
             res.push_back( who.i_rem( &*loc ) );
         }
+        batch_num = next_batch_ok ? batch_num + 1 : 1;
     }
 
     // Remove handled items from activity
@@ -3785,7 +3788,7 @@ void drop_activity_actor::do_turn( player_activity &, Character &who )
     const tripoint_bub_ms pos = placement + who.pos_bub();
     who.invalidate_weight_carried_cache();
     put_into_vehicle_or_drop( who, item_drop_reason::deliberate,
-                              obtain_activity_items( items, handler, who ),
+                              obtain_activity_items( items, handler, who, batch_num ),
                               pos, force_ground );
     // Cancel activity if items is empty. Otherwise, we modified in place and we will continue
     // to resolve the drop next turn. This is different from the pickup logic which creates
@@ -3949,7 +3952,7 @@ void stash_activity_actor::do_turn( player_activity &, Character &who )
 
     monster *pet = get_creature_tracker().creature_at<monster>( pos );
     if( pet != nullptr && pet->has_effect( effect_pet ) ) {
-        stash_on_pet( obtain_activity_items( items, handler, who ),
+        stash_on_pet( obtain_activity_items( items, handler, who, batch_num ),
                       *pet, who );
         if( items.empty() ) {
             who.cancel_activity();
--- a/src/activity_actor_definitions.h
+++ b/src/activity_actor_definitions.h
@@ -503,6 +503,7 @@ class pickup_activity_actor : public activity_actor
          * if not grabbing from the ground.
          */
         std::optional<tripoint> starting_pos;
+        int batch_num = 1;
 
     public:
         pickup_activity_actor( const std::vector<item_location> &target_items,
@@ -1024,6 +1025,7 @@ class drop_activity_actor : public activity_actor
         contents_change_handler handler;
         tripoint placement;
         bool force_ground = false;
+        int batch_num = 1;
 };
 
 class stash_activity_actor: public activity_actor
@@ -1058,6 +1060,7 @@ class stash_activity_actor: public activity_actor
         std::vector<drop_or_stash_item_info> items;
         contents_change_handler handler;
         tripoint placement;
+        int batch_num = 1;
 };
 
 class harvest_activity_actor : public activity_actor
--- a/src/item_location.cpp
+++ b/src/item_location.cpp
@@ -63,6 +63,18 @@ static item *retrieve_index( const T &sel, int idx )
     return obj;
 }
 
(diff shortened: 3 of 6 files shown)


GuardianDll:
Few thoughts:
It seems this change doesn't care about the size of items whatsoever, so moving thousands of apples (182 kilo) and 1000 units of salt (3,5 kilo) is equal. not good
next, it also ignore the fact human hand is pretty limited tool to move things - it can move either a bunch of small items, or one really big, and it's range is heavily depends on the size of the item
and the last, it completely ignore your own ability to move stuff - in most cases, when person need to move huge amount of items, it doesn't move every item on the ground, sure, but it also doesn't teleport it tile after tile - usually people just put stuff into bigger containers and move it after

My own thougts revolve around human hand as pocket (and maybe further mutations for different shapes of arms?) and some caching the player's free space in the inventory and between the items they move, so the first move calculate time to virtually "stuff" items you have in "free" containers, and all next moves cost you the "optimized" amount of time

For example, survivor may find 50 L of apples, they have their own hands that can store 2 apples, half-empty backpack (10L), and few sacks at the ground and/or in their inventory. The first move the game calculate time to put all apples in all possible containers, then process to move this stuff one tile, save the value, and every next move it checks is there new items in the pile (if you pick additional 10 kilo of apples, the game checks do you have enough space, recalculate additional time to stuff, and allow to move further), making you able to move a lot of stuff really fast, if you have enough bags

I don't really know how working this solution, i guess it require someone like Kevin to proceed

RenechCDDA:
My concerns are basically along GuardianDll's first point, moving 1000 rocks (about hand-sized) shouldn't be any faster than 1-at-a-time, but I'm very unsure what a good way to break it down would be. Items above a certain individual volume/weight are exempt?

Kamayana:
Apples/rocks are too big to move handfuls of, yes, but they're also big enough to move _armfuls_ of, so they shouldn't be one-at-a-time either. An object has to be actually pretty big for me to not even try to move multiple of them IRL; I think watermelon-size would be my limit.

This also prompts the question of: should the player need two free hands to move/grab/drop objects at top speed? I want to say yes, because the scenario of getting a nasty surprise by an enemy with your weapon on the ground because you were greedily grabbing handfuls of bullets is too good to pass up, but it also seems like it would be an annoying thing to demand/prompt the player every time they move items around. Could maybe have a threshold and say "that will take a while without two free hands, holster/drop your weapon?".

lispcoc:
I think GuardianDll's opinion is right from a game balance and realism perspective. I don't think the process of moving items should be any more complicated, although I'll see if it can be achieved with a simple implementation.
I also agree that this change should have a project manager decide on the policy since it has a big impact on game balance.

lispcoc:
I've thought about it, I think the simplest solution is to limit the mass and volume of items that can be batch moving. The way hands are treated as a container is very interesting, but I have no idea how to implement it.

@kevingranade It would be helpful if you could give me your opinion.

CHCOOH:
I would very much like to agree with you regarding this request.
However, I also understand the opinions of GuardianDll and RenechCDDA, so I have been thinking about the real cause of this problem.

One is that there may be a problem with putting "thousands of salt" or "1000s of apples" on the ground as it is.
It is natural to find a crate containing 50 liters of apples and put it in a backpack to keep it for yourself.
On the other hand, when you try to get "salt" or "meat" by crafting or dismantling, you drop it on the ground or table.
When processing rock salt into salt, you only need 100 per lump, but when crafting sulfur from a lump of sulfur, for example, you get 1562 pieces of sulfur per lump.
Just trying to pack these thousands of particles dropped on the table into something is very time consuming.

Another thing is that there is no separation of thought regarding the sturdiness of what is being handled.
With apples, you would be very careful when transferring them from the backpack to the crate, because the apples would be damaged if the bag was overturned.
But one would not in reality transfer salt, nails, stones, and bullets one by one from one container to another.
Nor would you use your hands to clean up glass shards if you could. Whether it is a shovel or a broom, anything that you are not worried about damaging is moved all at once using some kind of tool.
However, this idea is impractical in terms of implementation because it would be like applying a specific process to each and every item in the end.

So, I would propose the following compromise,
Allow the user to specify the container when crafting or dismantling to reduce the impact of this issue.
Shorten the usual time for transferring from one container to another.
This would be the case.

However, I think simply segregating the characteristics of items according to the size of the volume handled, as suggested by lispcoc, would be a broader solution.

ampersand55:
In moving very large quantities of apples or rocks you wouldn't handle them individually, but rather pour it from the container if you were dropping them, scooping them up with the container or shoving them into the container if you were filling something or shoving them if you were moving them from one tile to another.

alp7292:
i dont think we need more ways to waste time this works fine in most cases as we dont play with debug packpack to haul 1000 rocks in normal playthrough

RDru:
It should work like pumping/siphoning liquid to vehicle. In batches.

It should also be based on volume not amount.

kevingranade:
I agree with the insight that this is really about container loading/unloading. If you're moving around a lot of salt, you are not picking it up off the ground and putting it in a general-purpose container,  you're losing it into a bag/canister for salt.
This applies equally for things like apples, sand,  gravel and ammunition.

In short, a "batch cost reduction" for "moving items" is misplaced,  the action you're looking for is "bulk loading/unloading a dedicated container".

lispcoc:
I think it's sharp point that what we're really looking for is "bulk loading/unloading a dedicated container". To achieve this, there are several problems that must be resolved before implementing cost reduction.

1. The feature to pack items into containers such as plastic bags is difficult to access.
This feature should be accessible with a single keystroke, similar to [r]eload.
2. How “dedicated” should a dedicated container be?
I think it's nonsense to have separate loaders for sand and gravel, like the bullet speed loaders that already exist. Having separate containers for apples and pears seems to be just to increase the variety of items. Should items be broadly categorized as liquids or solids?
3. While picking up items, things end up in plastic bags or other containers in our backpack without we realizing it.

The issue of cost reduction should be discussed after these issues are resolved, so I would like to keep this PR pending.

Nebnis:
For point 2, i don't think  kevin meant that specifically. Just that if you have a bag of salt, you just gonna dump that salt in another bag rather than picking grain by grain, thus saving time. if there is a pile of salt in the ground, you just gonna swoop that with a sack grabbing a lot of salt in one go. Thus being a "dedicated batch load/unload" Similarly for the other items, if there is a box of apples you wanna move, you just gonna stick your 50 liter backpack into the apple box and try to grab as many as you can at once to save time

For point 3, you can whitelist and black list pockets to avoid items falling where they shouldn't

lispcoc:
Oh, It seems I was mistaken. (I'm not good at English, so it would be helpful if you could explain it to me.)
Regarding point 3, there is a problem that setting up the whitelist itself is a hassle. However, I feel like solving point 1 will naturally eliminate the problem, so I will solve this first.

Nebnis:
Ah i'm not sure if i can explain this better, maybe with an image? I will try to be through.  

Basically, you're not using your hands to move a lot of salt,sand,gravel, at once, you're using another bag, a backpack or cup or something that is large to grab the salt.  You could also be  dumping from one bag to another. You would save a lot of time grabbing a ton of rocks if you're just sweeping all into your sack rather than grab one by one with your hands. Moving things faster with bags.

Maybe someone can explain this better than me.

image

lispcoc:
What I wanted to achieve with this PR was achieved with #68214, so I will close this PR for now. 
In the future, I think it would be possible to improve the consistency of the process and reduce time by replacing picking up items with inserting them into the container and reimplementing it.
But it requires a drastic remodel and I don't have enough time to work on it : because a non-transparent function called Pickup::do_pickup is applied to each item to pick up, so you won't know which item goes into which container until you actually put it in.

PR #68020, opened 2023-09-06
Added cordless drill to "power tools" itemgroup

dchieh222:
#### Summary
Content

#### Purpose of change

The "power tools" itemgroup did not have cordless drills, which caused anomalies such as the entire hardware store as well as the hardware superstore never spawning cordless drills, despite it being a common item.

#### Describe the solution

This adds cordless drills to the "power tools" itemgroup so it will spawn where the itemgroup is used.

Accept that drills never appear in hardware stores, or convince myself that a cordless drill is not actually a power tool.

#### Testing

Ran and tested it on a local system, it spawned the cordless power drill as expected. 

#### Additional context

Quick search of "cordless power drill" at local hardware & home improvement stores such as Home Depot & Lowes shows that they are available for pickup. They are also defined as a power tool.


dchieh222:
Need to remove unneeded changes.

PR #68023, opened 2023-09-06
Replace attack_chance with condition

marimarigi:
#### Summary
None

#### Purpose of change
Since #65229, "attack_chance" seems to be unusable, so replace it.

#### Describe the solution
It has been replaced by the equivalent "condition".
Also, the part about "attack_chace" in the DOC has been removed as it is misleading.

#### Testing
I checked it by fighting monsters with modified "special_attack".


Maleclypse:
So it was named one way but documented another?

marimarigi:
After it was documented, the code was changed so that it no longer worked, but it was still in the documentation.

ZhilkinSerg:
This change breaks some tests.

PR #68026, opened 2023-09-06
Debug menu option to adjust NPC personality

RenechCDDA:
#### Summary
None

#### Purpose of change

Working on NPCs, need to test things with their personalities but I can't edit it from the debug menu!

#### Describe the solution
Copy much of the code for changing opinion, modify it to work on their personality instead

Maybe this could be sandwiched into the same menu but /shrug

#### Testing
image
image
image
image

#### Additional context
There's no cap on the values you can input even though personality values fall within the -10,10 range, so you can debug yourself into situations that can't be encountered in normal gameplay but I think that's probably fine? It is debug after all.

### DIFF
--- a/src/debug_menu.cpp
+++ b/src/debug_menu.cpp
@@ -1696,6 +1696,43 @@ static void character_edit_opinion_menu( npc *np )
     }
 }
 
+static void character_edit_personality_menu( npc *np )
+{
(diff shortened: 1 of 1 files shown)


ashGlaw:
I'm always here for more debug options. Cheers!

PR #68029, opened 2023-09-06
Cleanup arrows

GuardianDll:
#### Summary
None
#### Purpose of change
Found some arrows we have in game are not used anywhere (and another one was used in one place in aftershock, duh)
Decide to turn it back
Then decide to remove charges also
Then decide to simplify syntax using copy-from
#### Describe the solution
Arrows syntaxys was shortened using copy-from
`arrow_metal_sharpened_fletched` was close to removing, but then i decide to change it into just wooden arrow with hand-made grinded head
`arrow_plastic` same close to removing, but decide to left as fluff item
All arrows lost their `count` field (but because they are ammo, they are still count as charges, but still); values was tweaked respectively
`archery_ammo` got updated with all arrows, that could be sold (so both uncraftable, like carbon, and one, that require lot of effort like anvil, like wooden broadhead), and new `makeshift_archery_ammo` was created to store both makeshift and craftable arrows (maybe it was a bad decision, and i should make three groups: makeshift, craftable, manufactured?)
groups and maps that uses single arrows got replaced with this two loot groups respectively
#### Testing
The game starts, profs that spawn bow and arrow spawn proper amount of things, bows still shoot arrows
#### Additional context
No damage was changed whatsoever, so in theory dps test should be satisfied
didn't know what to do with explosive and flammable arrows, so they are mostly untouched


PatrikLundell:
Have you tested the effects on existing saves (to ensure things like stack of 20 not turning into a single arrow, etc.)?

I would call the craftable arrow group "craftable" rather than "makeshift" if it's to remain.

I can't really think of a location that would benefit from only proper craftable arrows (a fletcher's workshop possibly), but if we're talking survivor camps they might contain locally made makeshift arrows as well as store bought arrows, which could be any type, but I could see some only containing makeshift ones. I can't really see a location that would contain craftable arrows only, as I don't envision a case where a survivor advanced sufficiently to have access to production of all craftable arrows but was excluded from store bought ones.

GuardianDll:
despite i said "remove charges", they are still work as charges because they have type:AMMO - i just made so if we remove it from the type, arrows won't break apart

PatrikLundell:
I understood the "no change in behavior" part, but I'm always skeptical towards untested claims that "this shouldn't affect that". I've screwed up on that too many times...

PR #68039, opened 2023-09-07
Add three new professions (+some items maybe), branch the priest profession

KeremBabaG:
#### Summary

Content "Add three new professions"

#### Purpose of change

Adds three new professions that I think may be interesting, and branches the priest profession to three different instances: protestant, catholic and orthodox. Adds items if need be.

#### Describe the solution

The New Professions:

RV hunter: Self explanatory. Adds a rifle hunter (at least skill wise) with an RV, that lost their gun to balance the advantage. Will add a mission to find a new gun. Basic engine and driving proficiencies. I'm also thinking of adding the nomad trait. A good start for deathmobile enthusiasts.

Heist Driver: A cheesy Drive reference, but also adds a vehicle and combat focused start which is kinda lacking outside military professions imo. High vehicles and mechanics skill, moderate unarmed and melee skill and/or gun skills. Spawns with muscle car. Thinking of adding the gosling jacket lol

Prostitute: Alright this one is kinda vulgar but makes sense. I don't know if it is legal in New England but US in general have decent amount of sex workers to my knowledge. It is a social skill only skill start without the benefits of the religious professions. Maybe give a switchblade for defensive purposes.

And branch the priest profession into two others: Our current priest starts with the King James bible and a catholic symbol for some reason and idk if this is intentional. This aims to create seperate Catholic and Protestant(unitarian? whatever) priests. New England has appearently one of the highest catholic demographics os this makes sense, and fixes the current priest to be less offensive(?) and more realistic.

Rethink or don't add the new professions (pls don't take my Drive reference away ): )

#### Testing

In Progress

#### Additional context

Waiting for devs to review this because two of these are kinda absurd

### DIFF
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -3,27 +3,55 @@
     "type": "item_group",
     "subtype": "collection",
     "id": "holster_supp_MEU",
-    "entries": [ { "item": "m1911", "variant": "m1911_MEU", "ammo-item": "45_acp", "charges": 7, "contents-item": [ "suppressor" ] } ]
+    "entries": [
+      {
+        "item": "m1911",
+        "variant": "m1911_MEU",
+        "ammo-item": "45_acp",
+        "charges": 7,
+        "contents-item": [ "suppressor" ]
+      }
+    ]
   },
   {
     "type": "item_group",
     "subtype": "collection",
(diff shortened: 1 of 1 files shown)


KeremBabaG:
fuck forgot to update the thing

Issue #65305, opened 2023-04-25
Dormant zombies: a huge game changer that we're ready for now.


I-am-Erk:
### Is your feature request related to a problem? Please describe.

#36832 outlines a previous version of this but we now have tools to make it much more accessible.

Many monsters, including zombies, robots, and magiclysm golems, should not simply wander around all the time in brownian motion when not being observed. They should enter a dormant mode.

I will write this issue looking principally at zombies. We have a hard cap on the number of zombies we can tolerate in a location before they smash everything around them and cause lag, as well as just becoming silly from a game perspective.

### Solution you would like.

We should create "dormant" monsters, a form of trap that represents zombies lying around acting like human corpses. Imagine this as a "low power" mode, but it's also a way to keep run of the mill zombies present and dangerous, give them a way to ambush players, make night raiding scarier, discourage gunfire, but *also* encourage players to venture into more populated areas.

##### Implementation recommendation
1. Create a new item, "`corpse_zombie_dormant`". This is indistinguishable from the regular human corpse. (Note in the hopes of discouraging more repetition in the comments: it's not all that important if players can figure out which corpses are dormant zombies and which aren't.)
2. Create a trap, "`tr_zombie_dormant`". Add a bit of code somewhere to ensure that when `tr_zombie_dormant` is placed on the map, an associated `corpse_zombie_dormant` is placed. We may also want a sanity check that places the same trap down whenever the corpse item is placed, so you can never have one without the other. (See comments for some other ways this might be implemented)
3. When the trap is triggered, run an EOC or spell that revives the `corpse_zombie_dormant` in the same tile. More specifically, just delete the corpse and spawn a zombie there... we don't need to use the actual revive code.
4. Error trapping: the trap should trigger when you're in an adjacent tile, not just when you step on the tile, and the trap should trigger if you interact with or smash the `corpse_zombie_dormant`, or if the `corpse_zombie_dormant` is damaged but not destroyed (eg you drive on it with a car). The trap should ideally not be triggered by other zombies.
5. Stretch goals: it would be very cool if a damaged `corpse_zombie_dormant` revived as a zombie with fewer HP. However, the number of times you're giong to be able to damage one partly before it wakes up are pretty few so this isn't necessary on a first pass.

Once we have the basic ability to step on a zombie corpse and have it wake up and menace you, we need to work on the zombie hive mind a bit and make these guys really scary.
[...]

PR #68047, opened 2023-09-08
Sound-triggered traps

psuedolet:
Sound-triggered traps. Add traps to sound processing, add sound_threshold json key for traps, and add a test noise-triggered trap.

#### Summary
Features "Sound-triggered traps"

#### Purpose of change
Create the possibility to have things happen from sounds. One step towards the possibility of dormant zombies #65305

#### Describe the solution
The map now stores types of traps that have a sound trigger. When processing sounds, we iterate over these traps, similar to creatures. Sound trigger can be added to traps with the json key "sound_threshold", which defines the minimum volume of noise that will trigger the trap.

Add items to sound processing, so corpses can be used for dormant zombies. This seems like a potentially large performance hit and large amount of code work.
Pass the sound to the trap act function, so traps can do things like repeat the sound louder or only trigger for speech. This may be a potential future PR, but I wanted to get the basic functionality working.

#### Testing
Start a new game. Add the test trap. Yelling triggered the trap. Walking near it sometimes triggered the trap. The noise from the trap triggering sometimes triggered the trap.

#### Additional context
The new noise trigger is an addition to the standard enter-same-tile trigger. It does not remove the standard trigger.
If a triggered trap makes a noise it can trigger other noise-triggered traps, including itself!

### DIFF
--- a/src/sounds.cpp
+++ b/src/sounds.cpp
@@ -37,6 +37,7 @@
 #include "safemode_ui.h"
 #include "string_formatter.h"
 #include "translations.h"
+#include "trap.h"
 #include "type_id.h"
 #include "uistate.h"
 #include "units.h"
@@ -479,6 +480,16 @@ void sounds::process_sounds()
             if( vol * 2 > dist ) {
                 // Exclude monsters that certainly won't hear the sound
                 critter.hear_sound( source, vol, dist, this_centroid.provocative );
+            }
+        }
+        // Trigger sound-triggered traps
+        for( const trap *trapType : trap::get_sound_triggered_traps() ) {
+            for( tripoint tp : get_map().trap_locations( trapType->id ) ) {
+                const int dist = sound_distance( source, tp );
+                trap tr = get_map().tr_at( tp );
+                if( tr.triggered_by_sound( vol, dist ) ) {
+                    tr.trigger( tp );
+                }
             }
         }
     }
--- a/src/sounds.cpp
+++ b/src/sounds.cpp
@@ -484,9 +484,9 @@ void sounds::process_sounds()
         }
         // Trigger sound-triggered traps
         for( const trap *trapType : trap::get_sound_triggered_traps() ) {
-            for( tripoint tp : get_map().trap_locations( trapType->id ) ) {
+            for( const tripoint &tp : get_map().trap_locations( trapType->id ) ) {
                 const int dist = sound_distance( source, tp );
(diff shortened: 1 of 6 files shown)


src/sounds.cpp
```diff
@@ -481,6 +482,16 @@ void sounds::process_sounds()
                 critter.hear_sound( source, vol, dist, this_centroid.provocative );
             }
         }
+        // Trigger sound-triggered traps
+        for( const trap *trapType : trap::get_sound_triggered_traps() ) {
+            for( tripoint tp : get_map().trap_locations( trapType->id ) ) {
+                const int dist = sound_distance( source, tp );
+                trap tr = get_map().tr_at( tp );
```
ehughsbaird, line 489: No need to copy these.
```suggestion
            for( const tripoint &tp : get_map().trap_locations( trapType->id ) ) {
                const int dist = sound_distance( source, tp );
                const trap &tr = get_map().tr_at( tp );
```

src/trap.h
```diff
@@ -331,6 +338,13 @@ struct trap {
         static const std::vector<const trap *> &get_funnels();
         /*@}*/
 
+        /*
+         * Can the trap be triggered by sounds?
+         */
+        bool has_sound_trigger() const;
+        static const std::vector<const trap *> &get_sound_triggered_traps();
+        bool triggered_by_sound( const int vol, const int dist );
```
ehughsbaird, line 346: As the CI asks
```suggestion
        bool triggered_by_sound( int vol, int dist );
```

GuardianDll:
Please document your change in https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md#traps

ilovehans10:
> The noise from the trap triggering sometimes triggered the trap.

Couldn't this cause issues in some cases where a trap triggers itself indefinitely. I can see how that feature has some good uses, but would adding an optional cooldown time variable to traps allow for a cheap and easy way to avoid recursive traps.

psuedolet:
> 



> > The noise from the trap triggering sometimes triggered the trap.
> 
> Couldn't this cause issues in some cases where a trap triggers itself indefinitely. I can see how that feature has some good uses, but would adding an optional cooldown time variable to traps allow for a cheap and easy way to avoid recursive traps.

That would only happen if 
1. The trap has both a sound trigger and an effect that produces a sound
2. The sound produced is at least as loud as the sound trigger threshold
3. The trap persists after triggering

PR #68048, opened 2023-09-08
Allow item transform to pick a specified or random variant

RenechCDDA:
#### Summary
Features "Item transformation can now pick variant items to transform into"

#### Purpose of change
This feature was requested on the development discord and expands support for variants, which otherwise are not available except through item group spawns.

#### Describe the solution
Normal transform function calls item::set_itype_variant (previously transformed item did not get a variant at all). set_itype_variant takes a new defined `variant_type` json entry to determine what the variant should be, defaulting to `<all>` (produces random weighted variant).

#### Testing
No variant given: Random weighted variant

Specific variant given: That variant

`<any>` given: Random weighted variant

Specific variant given (undefined/incorrect id/etc): Debugmsg, as before


src/item.cpp
```diff
@@ -9207,6 +9207,27 @@ const itype_variant_data &item::itype_variant() const
...
+
+    // There's probably a better way to do this...
+    std::vector<itype_variant_data> variants;
+    int variant_count;
+    for( const itype_variant_data &option : type->variants ) {
+        variant_count++;
+        variants.emplace_back( option );
+    }
+
+    // Vectors start at 0.  I'm sure there's a simpler way to do this though...
```
ehughsbaird, line 9225: https://github.com/CleverRaven/Cataclysm-DDA/blob/e0694d002ca10f5329132a8bd9d91ed0aca1d5bf/src/rng.h#L91-L103

src/item.cpp
```diff
@@ -9207,6 +9207,27 @@ const itype_variant_data &item::itype_variant() const
     return *_itype_variant;
 }
 
+std::string item::pick_random_variant()
```
ehughsbaird, line 9210: You could probably use this instead:
https://github.com/CleverRaven/Cataclysm-DDA/blob/e0694d002ca10f5329132a8bd9d91ed0aca1d5bf/src/item.cpp#L9170-L9184

src/iuse_actor.cpp
```diff
@@ -277,14 +278,19 @@ std::optional<int> iuse_transform::use( Character *p, item &it, const tripoint &
...
+void iuse_transform::do_transform( Character *p, item &it, std::string variant_type ) const
 {
     item obj_copy( it );
     item *obj;
     // defined here to allow making a new item assigned to the pointer
     item obj_it;
     if( container.is_empty() ) {
         obj = &it.convert( target );
+        if( variant_type == "<any>" ) {
+            obj->set_itype_variant( obj->pick_random_variant() );
```
ehughsbaird, line 290: This gives a debugmsg if they item doesn't have variants, but you don't check for that.

src/item.cpp
RenechCDDA: Good call, thanks!

src/item.cpp
RenechCDDA: Good call again. Both of your remaining points should be resolved now. Going to give it some testing (and time for you to respond in case I really screwed the pooch) and will be undrafting

src/iuse_actor.h
```diff
@@ -56,6 +56,9 @@ class iuse_transform : public iuse_actor
         /** if set transform item to container and place new item (of type @ref target) inside */
         itype_id container;
 
+        /** if set choose this variant when transforming */
+        std::string variant_type;
```
ehughsbaird, line 60: One last thing! It looks like a variant should always be rolled (since one is rolled on save/load if it lacks one).
```suggestion
        std::string variant_type = "<any>";
```

MrDraMaster:
playing around with the EXE you provided, the `"variant_type": "<any>"`, `"variant_type": "[actual variant ID]"`, and `"variant_type": "[unknown variant ID]"` worked as intended :+1: 

but there's a UI issue I caught in the item description: the "Turns into: [target]" entry rolls the displayed target each time the item description is opened.
grafik
grafik

I would hope that displaying an explicit variant (so the `"variant_type": "[actual variant ID]"` case) is easy(ish), but I have no idea what the display should show in case of the `"variant_type": "<any>"` setting...

doc/JSON_INFO.md
```diff
@@ -4011,6 +4011,7 @@ The contents of use_action fields can either be a string indicating a built-in f
 "use_action": {
     "type": "transform",  // The type of method, in this case one that transforms the item.
     "target": "gasoline_lantern_on", // The item to transform to.
+    "variant_type": "condom_plain", // (Optional) Variant type to set for the transformed item. Special string `<any>` will pick a random variant from all available variants.
```
MrDraMaster, line 4014:
```suggestion
    "variant_type": "condom_plain", // Variant type to set for the transformed item. Defaults to the special string `<any>` that will pick a variant according to the weighted list of all variants. This does nothing for targets without variants.
```

RenechCDDA: I disagree with the wording you've given but I pushed a commit for this. It's already a rather verbose explanation, would like to try to keep it reasonable.

src/iuse_actor.cpp
```diff
@@ -277,14 +278,15 @@ std::optional<int> iuse_transform::use( Character *p, item &it, const tripoint &
     return result;
 }
 
-void iuse_transform::do_transform( Character *p, item &it ) const
+void iuse_transform::do_transform( Character *p, item &it, const std::string variant_type ) const
```
ehughsbaird, line 281: You need the & or else it copies.
```suggestion
void iuse_transform::do_transform( Character *p, item &it, const std::string &variant_type ) const
```

src/iuse_actor.h
```diff
@@ -114,7 +117,7 @@ class iuse_transform : public iuse_actor
         void finalize( const itype_id &my_item_type ) override;
         void info( const item &, std::vector<iteminfo> & ) const override;
     private:
-        void do_transform( Character *p, item &it ) const;
+        void do_transform( Character *p, item &it, const std::string variant_type ) const;
```
ehughsbaird, line 120:
```suggestion
        void do_transform( Character *p, item &it, const std::string &variant_type ) const;
```

PR #68049, opened 2023-09-08
[Magiclysm] More wizard furniture

Standing-Storm:
#### Summary
Mods "More wizard furniture for Magiclysm"

#### Purpose of change

You can wear a robe and wizard hat and look like a wizard in Magiclysm, but you can't set up a wizard's laboratory.

#### Describe the solution

Add construction recipes for an everburning torch sconce (that attaches to a wall like a wall light), an everburning torch candelabra that can be set up to provide omnidirectional light, and the ability to make your own magic circles. There are two grades of magic circle--one uses copper and provides the Mana Infusing and Mana Focusing qualities at 1, and the improved version uses silver and provides them at 2.

Also adds a construction recipe to assemble an alembic (from parts--transporting it as one unit seems unfeasible) and to turn a orrery into a transportable item and move it elsewhere. 

#### Testing

Everburning torch sconces and candelabras work as expected (though sconces also spawn power cords, they provide their light without drawing power). Alembics can be assembled and orreries can be placed. 

#### Additional context

I don't agree with a single magical torch producing 20% more light than an electric floor lamp, but those were the existing numbers so I just mirrored them.


ilovehans10:
This might be a separate issue, but would using everburning torches as headlights make sense? They're similar to atomic lamps, they wouldn't take any power, but it would make sense if they used some power to turn on/off (motorized light shield).

Standing-Storm:
> but would using everburning torches as headlights make sense

I do think it's reasonable to have them as vehicle mounted items (like on your ox-cart, for example), but I'm going to leave them out for this PR since I haven't ever touched non-appliance vehicle parts.

PR #68051, opened 2023-09-08
Optimize loading of pockets with many items inside

Kamayana:
#### Summary
Performance "Optimize loading of pockets with many items inside"

#### Purpose of change
When pockets have a large number of items inside the game can take a very long time to load them, as the game runs a `can_contain` check on every individual item inside, each of which checks against all the other items inside, giving it an exponential complexity growth.

#### Describe the solution
Use the simplified `can_contain` check created in #67631 to greatly reduce the processing time of the `combine` in `item::deserialize`. This simplified check ignores other contents of the pocket and only checks if the item could fit in the pocket even if it were empty. There's not really a reason to do the more intensive `can_contain` check, since if it was invalid it would've been handled before getting serialized in the first place.

It's possible there are other calls of `can_contain` that could use the simplified version, but I'm just doing it on a case-by-case basis for whatever needs optimization.

#### Testing
This save has a pocket universe filled with 100,000 pills: [Long Load-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12564070/Long.Load-trimmed.tar.gz)

1. Attempt to load that save in the vanilla version. When it gets to "Character save", the part where it's loading items, it hangs for so long that I eventually have to give up and force quit.
2. Apply changes.
3. Load the save in the new version. It only spends a few seconds on "Character save".
4. Check inventory to see that everything still loaded properly.

### DIFF
--- a/src/item_pocket.cpp
+++ b/src/item_pocket.cpp
@@ -1228,6 +1228,11 @@ void item_pocket::contents_info( std::vector<iteminfo> &info, int pocket_number,
                                            contains_weight() ) );
     }
 
+    if( remaining_volume() < 0_ml || remaining_weight() < 0_gram ) {
+        info.emplace_back( "DESCRIPTION",
+                           colorize( _( "This pocket is over capacity and will spill if moved!" ), c_red ) );
+    }
+
     // ablative pockets have their contents displayed earlier in the UI
     if( !is_ablative() ) {
         std::vector<std::pair<item const *, int>> counted_contents;
--- a/src/item_pocket.cpp
+++ b/src/item_pocket.cpp
@@ -1403,26 +1403,20 @@ ret_val<item_pocket::contain_code> item_pocket::is_compatible( const item &it ) 
 }
 
 ret_val<item_pocket::contain_code> item_pocket::can_contain( const item &it,
-        int &copies_remaining ) const
+        int &copies_remaining, bool ignore_contents ) const
 {
-    return _can_contain( it, copies_remaining, true );
+    return _can_contain( it, copies_remaining, ignore_contents );
 }
 
-ret_val<item_pocket::contain_code> item_pocket::can_contain( const item &it ) const
+ret_val<item_pocket::contain_code> item_pocket::can_contain( const item &it,
+        bool ignore_contents ) const
 {
     int copies = 1;
-    return _can_contain( it, copies, true );
-}
-
-ret_val<item_pocket::contain_code> item_pocket::can_contain_skip_space_checks(
-    const item &it ) const
-{
-    int copies = 1;
-    return _can_contain( it, copies, false );
+    return _can_contain( it, copies, ignore_contents );
 }
 
 ret_val<item_pocket::contain_code> item_pocket::_can_contain( const item &it,
-        int &copies_remaining, const bool check_for_enough_space ) const
+        int &copies_remaining, const bool ignore_contents ) const
 {
     ret_val<item_pocket::contain_code> compatible = is_compatible( it );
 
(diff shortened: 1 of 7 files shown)


KHeket:
Bless you saint Kamayana, goddes of optimization 🙏

Qrox:
> if it was invalid it would've been handled before getting serialized in the first place.

The size of items could change between serialization and deserialization due to JSON changes so I think you should test what happens in that case. Ideally the items should spill out of the pocket but I suppose it's fine to keep them inside the pocket as long as it does not generate any errors.

Kamayana:
> > if it was invalid it would've been handled before getting serialized in the first place.
> 
> The size of items could change between serialization and deserialization due to JSON changes so I think you should test what happens in that case. Ideally the items should spill out of the pocket but I suppose it's fine to keep them inside the pocket as long as it does not generate any errors.

Tested it, the container will sit in the world undisturbed until the player picks it up, which will spill the items out. No errors. However, if the container is non-rigid and wearable, examining it will show the "volume exceeds capacity" error from `relative_encumbrance`, so there's something to fix there. I also think it might be possible to efficiently spill too-full containers on load, so I can look into that.

Kamayana:
So I tested having overfilled pockets spill when loaded, but it felt weird for items to spill items out all on their own. The way the game currently works, with overfilled containers only spilling when you pick them up, seemed like something that could actually make sense in-universe, so I decided to see how it'd work to replace the error message with a note in the examine menu about it. I made it so overfilled pockets spill if they get hauled as well.

Here's a comparison of what happens when loading a save with containers that are now overfilled because of json changes:
#### Vanilla:

1. **Shows an error during loading ("error: item apple cannot fit into pocket while loading: not enough space").**
2. The game is loaded.
3. After a single turn, carried overfilled containers spill.
4. Overfilled containers on the map do not spill on their own.
5. Examining the overfilled container on the ground shows an incorrect number of items, falsely decrementing the number to the 88 items (29.92 L) the pocket can hold, even though there are actually 90 inside: image
7. On picking up the overfilled container, its contents spill.

#### With my previous changes:

1. No error during loading.
2. The game is loaded.
3. After a single turn, carried overfilled containers spill.
4. Overfilled containers on the map do not spill on their own.
5. On examining the overfilled container, **an error is shown ("volume exceeds capacity (30600ml > 30000ml) ).**
6. Examining the overfilled container on the ground correctly shows there are 90 apples currently inside the pocket:
image
8. On picking up the overfilled container, its contents spill.

#### With my new changes:

1. No error during loading.
2. The game is loaded.
3. After a single turn, carried overfilled containers spill.
4. Overfilled containers on the map do not spill on their own.
5. Examining the overfilled container on the ground correctly shows there are 90 apples currently inside the pocket, but makes a note that the container is overfilled and will spill if moved:
image

src/item_pocket.cpp
```diff
@@ -1504,17 +1503,17 @@ ret_val<item_pocket::contain_code> item_pocket::_can_contain( const item &it,
...
     }
 
-    if( data->ablative && !contents.empty() ) {
-        if( contents.front().can_combine( it ) ) {
-            // Only items with charges can succeed here.
-            return ret_val<item_pocket::contain_code>::make_success();
-        } else {
-            return ret_val<item_pocket::contain_code>::make_failure(
-                       contain_code::ERR_NO_SPACE, _( "ablative pocket already contains a plate" ) );
+    if( data->ablative ) {
```
Brambor, line 1506: Is there a reason to split one `if` into two?

CAHEK7: I guess the reason is to combine following check at the line 1517.
Transform this code
```C++
 if( data->ablative && !contents.empty() ) {
 }

 if( data->ablative ) {
    if( it.is_rigid() ) {
    }
 } 
```
into that:
```C++
 if( data->ablative ) {
    if ( !contents.empty() ) {
    }
    if( it.is_rigid() ) {
    }
 } 
```

Kamayana: Yes, it's to compact the code a bit, but it's also because I found out the hard way that the section of code around the `!contents.empty` check must be called before the `is_rigid` section or else there's a risk of causing an infinite loop. Combining them into one large block makes it more clear the order should be kept like that.

PR #68068, opened 2023-09-09
Restore the craft amount for arrows to 10

snipercup:
#### Summary
Balance "Restore the craft amount for arrows to 10"

#### Purpose of change
PR #68029 removed charges from arrows. This caused the recipes for the arrows to put out 1 arrow instead of 10. This pr changes the recipes to they put out 10 again.

#### Describe the solution
Add 'charges: 10' to the relevant recipes

Re-balance the recipe some other way. Reducing the craft time by a factor of 10 is possible, but you can't reduce the crafting materials by a factor of 10.

#### Testing
Loaded up the game and crafted wooden broadhead arrows. It produces 10 instead of 1
image


dchieh222:
Archers celebrate.

PR #68097, opened 2023-09-11
[DONE] Visitable flood fill

kevingranade:
#### Summary
Performance "Precalculate visitable zones to optimize inter-monster aggression checks"

#### Purpose of change
When there are a large number of monsters of opposing factions in the reality bubble, the checks for vision etc between them can get expensive, especially when we are trying to run the simulation forward quickly when e.g. crafting, waiting, or sleeping.

#### Describe the solution
This adds a concept I'm calling a "visitable region" and calculates it for each monster. The core idea is it performs a flood-fill across every tile that a monster *might* be able to traverse OR see. The reason to make it so permissive is so that it is a proper partition of the game space, every monster is in exactly one zone, so at monster AI calculation time all the monster needs to do is compare its zone number with an opposing monster's zone number to determine if there is any possibility of them interacting.
For the same reason, it looks at terrain-based transparency, ignoring things like field effects which change frequently and invalidate the map.
In principle, this map would only need to be re-calculated once every 24 turns or so when the player triggers a map shift, and almost never when the player is not moving.

As a type of cache, there's always an option to do nothing, but inter-monster agression performance is a pretty hard nut to crack.

#### Testing
So far I've been focusing on whether the system is even a net win and whether it's doing the appropriate work. I have added some debug statements that summarize the extent of the zones that are produced and they look like they are the appropriate sizes.
This really needs some unit tests with some different scenarios in it to demonstrate that the flood filling is working as intended.

#### Additional Context
This is based on pathfinding caching algorithms, but the main difference is uninioning traversibility with vision such that the cache remains stable and handles all different movement/vision modes.

#### Known Issues
This is currently a net negative as the flood fill is quite slow and I don't have any invalidation-based caching in place yet.  I've been focusing on making the flood-fill itself reasonably performant first before messing with caching.
There are also some potentially large optimizations I might be able to add to further optimize it.
1. Invalidation based caching might be all that is needed since it should run very infrequently, almost never when the player isn't moving. It invalidates on map shift, when the passability state of a tile changes, and when field-oblivious transparency changes. I did a quick test of this by generating the cache once and removing the invalidation logic and it results in eliminating about 10% of total overhead in a pathological case of waiting above an uncleared lab.
2. This is likely to have noticeably bad overhead when moving quickly, e.g. auto-drive. If so it might be mitigable by only performing a flood fill if the player stays in the same location for several turns.
3. Searching for monsters to mark is a pretty dominant operation, as it flood fills it searches for monsters and the negative case where there's no monster present is eating us alive. We could stick a bitset in front of the find operation in creature_at<>() and potentially save a ton of hashing.
4. All the uniform above-ground submaps could be handled specially and most likely cut overhead in half.

### DIFF
--- a/src/creature.h
+++ b/src/creature.h
@@ -742,8 +742,20 @@ class Creature : public viewer
         virtual bool uncanny_dodge() {
             return false;
         }
+        void set_reachable_zone( int zone ) {
+            reachable_zone = zone;
+        }
+        int get_reachable_zone() const {
+            return reachable_zone;
+        }
 
     private:
+        /** This number establishes a partition of zones on the map that have shared
+         * reachability/visibility. In short, if you aren't in the same zone as some other monster,
+         * you can ignore them since you won't be able to reach them by any combination of
+         * regular movement and vision. **/
+        int reachable_zone = 0;
+
         /** The creature's position in absolute coordinates */
         tripoint_abs_ms location;
     protected:
--- a/src/flood_fill.h
+++ b/src/flood_fill.h
@@ -50,6 +50,163 @@ std::vector<Point> point_flood_fill_4_connected( const Point &starting_point,
 
     return filled_points;
 }
+
+struct span {
+    span() : startX( 0 ), endX( 0 ), y( 0 ), dy( 0 ), z( 0 ),
+        dz( 0 ) {}
+    span( uint8_t p_startX, uint8_t p_endX, uint8_t p_y, int8_t p_dy, int8_t p_z,
+          int8_t p_dz ) : startX( p_startX ), endX( p_endX ), y( p_y ), dy( p_dy ), z( p_z ),
+        dz( p_dz ) {}
+
+    uint8_t startX;
+    uint8_t endX;
+    uint8_t y;
+    int8_t dy;
+    int8_t z;
+    int8_t dz;
+};
+
+inline void add_span( std::vector<span> &spans, int startX, int endX,
+                      int y, int dy, int z, int dz )
+{
+    const int clamped_z = clamp<int8_t>( z, -OVERMAP_DEPTH, OVERMAP_HEIGHT );
+    spans.emplace_back( std::max( startX, 0 ), std::min( endX, MAPSIZE_X - 1 ),
+                        clamp( y, 0, MAPSIZE_Y - 1 ), dy, clamped_z, dz );
+}
+
+
+template<typename UnaryPredicate, typename UnaryVisitor>
+void flood_fill_visit_10_connected( const tripoint_bub_ms &starting_point, UnaryPredicate predicate,
+                                    UnaryVisitor visitor )
+{
+    std::array<std::array<std::bitset<MAPSIZE_Y>, MAPSIZE_X>, OVERMAP_LAYERS> visited;
+    std::array<std::array<std::bitset<MAPSIZE_Y>, MAPSIZE_X>, OVERMAP_LAYERS> visited_vertically;
+    std::array<std::vector<span>, OVERMAP_LAYERS> spans_to_process;
+    int current_z = starting_point.z();
+    add_span( spans_to_process[current_z + OVERMAP_DEPTH], starting_point.x(), starting_point.x(),
+              starting_point.y(), 1, starting_point.z(), 0 );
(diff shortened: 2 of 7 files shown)


PatrikLundell:
NPCs and monsters changing the terrain need to be accounted for as well, such as a hulk smashing through a wall because it heard noise on the other side (as noise is another monster activity trigger, beside vision).

kevingranade:
Implemented several of the performance improvements, unfortunately the creature_at() bitset has a bug I need to track down that triggers when a monster is unloaded from the map, and I'm not actually sure that the bitmap is being handled correctly when the map is shifted.

kevingranade:
Changed my mind about the bitset thing, removed it instead as I'm pretty sure it's not necessary.
This needs testing to verify that it's assigning monsters to zones correctly, then it's ready to go.

kevingranade:
Alright, wrapped up a test suite and shook out a few issues, AFAICT this is good to go.

src/creature.h
```diff
@@ -742,8 +742,20 @@ class Creature : public viewer
         virtual bool uncanny_dodge() {
             return false;
         }
+        void set_reachable_zone( int zone ) {
```
kevingranade, line 745:
```suggestion
        void set_reachable_zone( int zone ) const {
```

src/creature.h
```diff
@@ -742,8 +742,20 @@ class Creature : public viewer
...
+            reachable_zone = zone;
+        }
+        int get_reachable_zone() {
```
kevingranade, line 748:
```suggestion
        int get_reachable_zone() const {
```

src/flood_fill.h
```diff
@@ -50,6 +50,163 @@ std::vector<Point> point_flood_fill_4_connected( const Point &starting_point,
...
+
+    uint8_t startX;
+    uint8_t endX;
+    uint8_t y;
+    int8_t dy;
+    int8_t z;
+    int8_t dz;
+};
+
+static void add_span( std::vector<span> &spans, int startX, int endX,
```
kevingranade, line 69:
```suggestion
void add_span( std::vector<span> &spans, int startX, int endX,
```

src/map.h
```diff
@@ -349,6 +349,12 @@ class map
         void set_outside_cache_dirty( int zlev );
         void set_floor_cache_dirty( int zlev );
         void set_pathfinding_cache_dirty( int zlev );
+        void set_visitable_zones_cache_dirty( bool dirty = true ) {
+            visitable_cache_dirty = dirty;
+        };
+        bool get_visitable_zones_cache_dirty() {
```
kevingranade, line 355:
```suggestion
        bool get_visitable_zones_cache_dirty() const {
```

src/point.h
```diff
@@ -352,4 +352,10 @@ inline const std::array<tripoint, 8> eight_horizontal_neighbors = { {
     }
 };
 
+inline const std::array<tripoint, 3> three_vertical_directions{{
```
kevingranade, line 355:
```suggestion
```

src/flood_fill.h
```diff
@@ -50,6 +50,163 @@ std::vector<Point> point_flood_fill_4_connected( const Point &starting_point,
...
+
+    uint8_t startX;
+    uint8_t endX;
+    uint8_t y;
+    int8_t dy;
+    int8_t z;
+    int8_t dz;
+};
+
+static inline void add_span( std::vector<span> &spans, int startX, int endX,
```
kevingranade, line 69:
```suggestion
inline void add_span( std::vector<span> &spans, int startX, int endX,
```

src/flood_fill.h
```diff
@@ -50,6 +50,163 @@ std::vector<Point> point_flood_fill_4_connected( const Point &starting_point,
...
+                }
+            }
+            if( current_z < -OVERMAP_DEPTH ) {
+                break;
+            }
+        }
+        tripoint_bub_ms current_point{ static_cast<int>( current_span.startX ), static_cast <int>( current_span.y ), static_cast<int>( current_span.z ) };
+        // Special handling for spans with a vertical offset.
+        if( current_span.dz != 0 ) {
+            bool span_found = 0;
```
kevingranade, line 123:
```suggestion
            bool span_found = false;
```

Maleclypse:
Pre-change TCL times
Pre-#68097 timer in TCL

Maleclypse:
Post-change TCL times
Post #68097

PR #68103, opened 2023-09-11
[Magiclysm] Everburning Torch rebalancing

Standing-Storm:
#### Summary
Mods "Everburning Torch rebalancing"

#### Purpose of change

I was very surprised when I looked at the everburning torch and saw it was set at LIGHT_240, which is more than the standing lamp's "bonus": 200.  It seemed pretty weird to me that a magical torch should give off more light than an electric lamp, especially with Magiclysm's premise that technology is often more convenient and sometimes better than magic, and when I looked up the source for everburning torch (D&D), it said that

> This otherwise normal torch has a [continual flame](https://www.dandwiki.com/wiki/SRD:Continual_Flame) spell cast upon it. An everburning torch clearly illuminates a 20-foot radius and provides shadowy illumination out to a 40-foot radius.

And if you look up a torch it says

> A torch burns for 1 hour, clearly illuminating a 20-foot radius and providing shadowy illumination out to a 40- foot radius. 

So, an everburning torch is a torch that burns forever but illuminates the same area, and we already have a light value for torches.

#### Describe the solution

Change the everburning torch so it matches the torch's light value (LIGHT_60). Do the same to the everburning sconce and change the everburning candelabra to LIGHT_120

#### Testing

Light values are reflected in game.

#### Additional context

I was going to try to make a dropped everburning torch start fires but it doesn't look like actual torches do that either: 
image

### DIFF
--- a/data/mods/Magiclysm/furniture.json
+++ b/data/mods/Magiclysm/furniture.json
@@ -342,7 +342,7 @@
     "color": "red",
     "move_cost_mod": 2,
     "required_str": 12,
-    "light_emitted": 240,
+    "light_emitted": 120,
     "flags": [ "TRANSPARENT", "EASY_DECONSTRUCT" ],
     "deconstruct": { "items": [ { "item": "everburning_torch_candelabra", "count": 1 } ] },
     "bash": {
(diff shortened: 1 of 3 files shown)


cskyaluna:
I will say that the everburning torch should not start fires - in D&D, it's a flame that creates no heat.

Standing-Storm:
> I will say that the everburning torch should not start fires - in D&D, it's a flame that creates no heat.

Huh...it's currently possible to cook with it, but if there's no heat that shouldn't be possible either.

cskyaluna:
Yeah, in CDDA it produces heat, but to be accurate to source material,
maybe it shouldn't. Would be a good nerf.

PR #68119, opened 2023-09-13
[MoM] Add two psionic animals

Standing-Storm:
#### Summary
Mods "Add two psionic animals"

#### Purpose of change

Humans aren't the only ones that gained psionic powers as a result of the Cataclysm, and MoM could use some more interesting creatures out there.

#### Describe the solution

Add two new monsters--the hidebehind and the bandersnatch.

The hidebehind is a psionic cougar that uses telepathy to convince its target that there's nothing there:

> Even looking directly at this cougar and knowing that it is there, a part of your brain simply refuses to see it.  It flickers in and out of sight in a way that makes your eyes hurt, and if you turn your attention away from it even for a moment, it disappears completely.

Mechanically, it gains the CAMOUFLAGE and NIGHT_INVISIBILITY flags.

The bandersnatch is a telekinetic bear.  Its powers grant it more armor against bullets and bashing weapons, allow it to pull targets close, and:

> An American black bear. A large omnivorous scavenger, it has powerful claws and jaws, and is an effective ambush hunter.  This one is distinguished from its fellows by its intent expression, its lack of caution in your presence, and the fact that it's hovering six inches off the ground.

It also has very slow regeneration. Thanks to these powers, it's more aggressive than a normal bear.

Bandersnatches and hidebehinds both start spawning 45 days after the Cataclysm. 

#### Testing

Animals behave the way I want them to.

#### Additional context
Having Telepathic Shield up should let you see the hidebehind just fine, but there's unfortunately no way to do that.


MNG-cataclysm:
Is the hidebender inspired by the [Displacer Beast](https://5e.tools/bestiary/displacer-beast-mm.html)?

Standing-Storm:
It's taken directly from [American folklore](https://en.wikipedia.org/wiki/Hidebehind)!

MNG-cataclysm:
I never knew such a thing existed in folklore, that's pretty interesting.

MNG-cataclysm (approved):
Everything looks good to me, content-wise.

Issue #67447, opened 2023-08-04
Grappling hooks should help with more descents.


MrHrulgin:
### Is your feature request related to a problem? Please describe.

I ran across a crater, and I could see a rail line down at the bottom of it. I had a grappling hook on me, and wanted to use it to descend to the bottom of the pit. But when I use the grappling hook at the edge of the pit it (somehow) extends upwards into space, and when I use it on the open air at the edge, I get:  

DEBUG    : Setting furniture f_rope_up at (66,64,0) where terrain is t_open_air (which is_open_air)
If this is intentional, set the ALLOW_ON_OPEN_AIR flag on the furniture

 FUNCTION : furn_set
 FILE     : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\map.cpp
 LINE     : 1726
 VERSION  : 6c7384d

and if I ignore the message, I see the sad remains of my grappling hook at the bottom of the pit.

If instead I step off the edge or climb down off of it, I fall freely to the floor without the grappling hook doing anything.

### Solution you would like.

I'd like it if I could deploy a grappling hook into open air to allow me to descend safely, and/or it would save me if I was going to take a fall.

### Describe alternatives you have considered.

Not doing anything about it.

teapot156:
As they are now it's kind of strange. It's like a portable ladder. 

So you can place it on the ground and use it to climb up as if you were throwing it up there but when you get to the top, it's still stuck below and now can no longer help you climb. I haven't tested it but I think a rope/grappling hook helps you "climb down", so you'd need 2 grappling hooks to successfully climb up and down safely. It'd be great if the item was looked it since it's very cool to run around with one or have one on you when you need it.

AmarinReyny:
Maybe having the grappling hook create a "hook end" and "rope end" when deployed would help with this? And if deployed on an open air tile, it would have the hook end made on that tile, and a rope end on the tile below it. Also, maybe there could be a way to prevent it from falling, as long as the hook end is adjacent to a tile that isn't open air.

It would also be cool if characters could be prevented from falling if they're on the rope end tile (albeit with a risk of losing one's grip and falling), even if it's on open air, so it'd be like using the grappling hook to go to the floor below, instead of immediately falling through all open air tiles and only landing (probably painfully) on whatever ground tile they reach first.

EvanBalster:
[...]

PR #54555, opened 2022-01-18
Slightly rework ledges, allow Web Diver to rappel down Z-levels, add a character flag to climb up/down walls

Venera3:
#### Summary
Features "Allow Web Diver to rappel, add infrastructure for wall crawling"

#### Purpose of change

Limb-proofs vertical movement, allows for wall crawling down the line, and stops Web Diver's description from lying.
Fixes #54561

#### Describe the solution

- Rework ledge/vertical movement logic. Currently the safes way to climb off a roof is by climbing down an empty side (*not* on a downspout), since it bypasses `slip_down` completely and instead relied on a very generous check of `frng( 8.0, 1.0 )> falling_mod` to bypass one Z-level's worth of falling. Instead, now vertical movement and climbing down ledges universally check for `slip_down`.
- Climbing up Z-levels (that is, "moving up a Z-level without a stair/ramp/etc" ) now universally takes at least 5 seconds, counts as an extreme activity and costs at least 500 stamina
- Added a new character modifier - `limb_fall_mod` to influence the amount of falling damage dealt. Previously it used `limb_movecost_modifier`, which was counterproductive since it meant that higher encumbrance made you take less falling damage.
- Slightly modified `slip_down`' to take arm encumbrance and stamina ratio into consideration as well. Also fixed the weight ratio calc not doing anything and redid the math to use percent chances of slipping instead of one-in-X.
- Added the `WEB_RAPPEL` character flag to allow for safe descent from a ledge across any number of Z-levels and added it to Web Diver's new activatable mode - at the cost of a good chunk of calories
- Added the `WALL_CLING` character flag to allow climbing up/down z-levels as long as the point above is adjacent to at least one wall. Costs 10 seconds and 1000 stamina and triggers a `slip_down` check every level to keep it reasonable unsafe. For now it only lives on the Debug Tail, since the Arachnid Arms' drawbacks don't really justify that kind of buff (yet.).

#### Describe alternatives you've considered

Pull the fixing/rearranging stuff into a separate PR, but it felt related enough.
Allow for horizontal movement while wall clinging, but this works for a proof of concept.
Use some sort of effect to confer immunity to falling while adjacent to a wall, but that felt a bit hacky (not that the game forgetting to check if you're falling between moves isn't janky, but at least it's not jank I introduced).

#### Testing

[...]

PR #40035, opened 2020-04-30
Slipping when climbing down may happen only on unstable terrain below

Night-Pryanik:
#### Summary
SUMMARY: Balance "Slipping when climbing down may happen only on unstable terrain below."

#### Purpose of change
Closes #39892 - Can fall when climbing down a stepladder.

#### Describe the solution
Added a check for unstable terrain below.

#### Describe alternatives you've considered
None.

#### Testing
Climbed up and down on stable terrain (stepladder) and on non-stable terrain (downspouts), also simply jumped down.

#### Additional context
I hope you're satisfied with this solution @sick-trigger.

Issue #39892, opened 2020-04-25
Can fall when climbing down a stepladder


sick-trigger:
# Describe the bug
You can slip and fuck yourself up when climbing down onto a stepladder, despite it saying you should be able to climb back up easily.

# Steps To Reproduce
1. place stepladder next to building
2. climb up and down it over and over until you fall
3. break legs, rage quit

# Expected behavior
using a stepladder to climb down a roof should be 100% safe. When climbing down from a roof via an unsafe method the game should /very clearly/ indicate you might fall off for massive damage.

# Versions and configuration
- OS: Windows
    - OS Version: 10.0 1809
- Game Version: 0.E-1182-g9f6568b [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food]
]

# Additional context
The changes to climbing down added in #39635 were incredibly rude and falling off a roof is infuriating when it happens.

A better solution is to put it back how it was, where you can either safely climb down somewhere or you cannot, without any RNG,  and with a clear message indicating that you're about to fall, and then just remove climbing up downspouts altogether, with maybe some exceptions for the parkour trait.

Night-Pryanik:
> You can slip and fuck yourself up when climbing down onto a stepladder, despite it saying you should be able to climb back up easily.

And where's the contradiction? You really can climb back **up** easily, it doesn't say anything about climbing **down** easily or safely.

> using a stepladder to climb down a roof should be 100% safe.

When you're climbing down from any roof, you're **NOT** using stepladder, or downspouts, or any other furniture or terrain. It's simply not implemented at the moment. Every time you climb down from a roof, in fact you're climbing down, using only your hands, that's why you can really easily slip down and fall off.

> When climbing down from a roof via an unsafe method the game should /very clearly/ indicate you might fall off for massive damage.

Isn't that obvious? Climbing down from a roof is dangerous as hell.

>rage quit
>infuriating

Maybe you should calm down a little? Even if it's a bug and must be fixed, it's not a reason to go into berserk mode.

sick-trigger:
when you try climbing down somewhere without something below to climb down you get a message warning you that it might hurt, which gives the impression you don't have to worry about damage when the message is only concerned with how difficult it'll be to get back up (which is how it used to behave)

[...]

PR #68120, opened 2023-09-13
Climbing Down: Safe Ladders + Risk Level in Prompts

EvanBalster:
#### Summary
Balance "Climbing down stepladders is now safe; climbing down ledges tells you how risky it is."

#### Purpose of change

1: Descending to a lower Z-levels is a [longstanding source of player frustration](https://github.com/CleverRaven/Cataclysm-DDA/issues/39892).  The current prompt indicates whether a player can climb back up, but not the degree of risk involved in climbing down or the nasty consequences of a slip.  The game does not mention the effect of low stamina on climbing success rates.  In particular, the current prompt "You should be able to climb back up easily." creates the mistaken impression that it is safe to climb down onto a downspout, fence or ladder.

2: Currently, characters ignore stepladders and do not use them when climbing down, which can result in unexpected falls.  An earlier PR https://github.com/CleverRaven/Cataclysm-DDA/pull/40035 was intended to make ladders safe to descend, and was merged, but appears to have been faulty.  A later PR https://github.com/CleverRaven/Cataclysm-DDA/pull/54555 appears to have overwritten that code.

#### Describe the solution

- Climbing down 1 level into a space containing a placed ladder (LADDER flag) is now 100% safe, like using a grappling hook.
  - The GOES_UP flag also qualifies the space as a ladder for purposes of this check.
  - CLIMBABLE objects such as downspouts, fences and [placed grappling hooks](https://github.com/CleverRaven/Cataclysm-DDA/issues/67447) are no safer than before.
- Add messages when the player's chance to slip is affected by stamina.
  - Limit the effect of stamina to at most a 10x increase in the chance to fall (previously it was unlimited and could divide by zero).
- Messages related to slip chance are now printed before showing the "Climb down?" prompt, giving players some hints about how to avoid a fall.
- The "climb down?" prompt now describes the risk of slipping, the estimated injury from slipping, and the expected difficulty of climbing back up.  Each of these is color-coded.
  - Risk ranges from `seems safe` to `doesn't seem possible` in five steps.
  - Severity ranges from `Falling wouldn't hurt much.` to `Falling would kill you.` and is based on a simplified version of the `Character::impact` math without considering the character's body or equipment.  Previous code checked fall modifier but not height or anything else.
  - Chance to climb back up is the same as before, but with coloring and slight rephrasing.
  - When contemplating a multi-story ledge, the player is warned that they'll fall the remaining N-1 stories even after a successful descent (unless using webs).
- Some logic from `game::slip_down` was extracted to a new method `game::slip_down_chance` to facilitate the changes above.
- A message is displayed upon *successfully* climbing down a ladder or a ledge.  This makes it clearer that downspouts are not helpful for climbing down.

New thresholds and messages for risk level:

| Slipping Chance | Description |
| --- | --- |
| 2% or less | "It seems safe to climb down like this." |
| 3% to 7% | "It seems a bit tricky to climb down like this." |
| 8% to 19% † | "It seems somewhat risky to climb down like this." |
| 21% to 60% | "It seems very risky to climb down like this." |
| 60% or more | "It looks like you'll slip if you climb down like this." |
| 80% or more | "It doesn't seem possible to climb down safely." |

† (marks the most common message for a default survivor examining a one-story ledge)

| Estimated Damage | Description |
| --- | --- |
| less than 5 | "Falling wouldn't hurt much." |
| 5 to 14 † | "Falling would hurt." |
| 15 to 29 | "Falling would hurt badly." |
| 30 to 59 | "Falling would break bones." |
| 61 to 99 | "Falling could cripple or kill you." |
| 100 or more | "Falling would kill you." |

**Leave a chance to slip from ladders.**  I would _prefer_ to have done it this way — climbing down a ladder still takes care and shouldn't be done with full hands, empty stamina or a heavy load.  Unfortunately this would require further changes to the `slip_down` function signature (the method doesn't know the direction or location of climbing).  It would also make stepladders less safe than grappling hooks, which is counterintuitive.  In the future it may be interesting if grappling hooks and ladders were negative modifiers (~25%) to `slip_down_chance`.

**Less elaborate climb-down prompt.**  I chose to indicate both the chance of a slip and the severity of a (typical unarmored) injury resulting from the fall.  Perhaps this is more info than is needed.  I left a little bit of ambiguity when risk is close to 0%.

**Clearly distinguish zero and non-zero slip chance.**  Currently a zero chance is described as a "very small risk".  Giving players perfect knowledge didn't seem quite in the spirit of this game but I'm interested in input from others.  The way I've done it — only indicating certain safety if the probability is negative 5% or less — was based on the idea that there might be negative modifiers in the future.

**Breaking fall damage calculation out of `Character::impact` for accurate injury estimate.**  Seemed like too much disruption for this change.  The estimated fall damage is calculated as if an unarmored human fell the given distance and only factors in fall damage multiplier (the previous code *only* checked this multiplier, and did not consider height).

#### Testing

- Climbing up/down evac shelter with & without wooden stepladder, at various levels of stamina.
- Tested successful use of ladder and unsuccessful ledge dismount due to low stamina.
- Tested climbing down by `e`xamine and `>`.
- Tested with multi-story buildings.
- Tested with grapple hook (single- and multi-story climbs).

#### Additional context

image


src/game.cpp
```diff
@@ -13138,31 +13156,79 @@ void game::climb_down( const tripoint &examp )
...
                 has_grapnel = false;
             } else {
                 web_rappel = false;
             }
         }
 
-        if( !has_grapnel ) {
+        if( !can_use_ladder && !has_grapnel ) {
             const char *query;
+            const char *climb_back_up = "";
```
Night-Pryanik, line 13177: Does it really need to be a pointer char instead of a string?

Night-Pryanik:
Players don't know that "injury" in message like "Climbing down runs a risk of injury" is actually a degree of severity. At the moment it looks like a generic info message with players' reaction like "yeah, thanks for stating the obvious, game".
I feel it needs some color-coding, or some other means to make players clearly recognize that there are several degrees of severity.

src/game.cpp
EvanBalster: Query was const char*, so I followed that convention.  I agree that std::string is better in this situation.  I'll change it.  Leaving `query` untouched though.

EvanBalster:
> Players don't know that "injury" in message like "Climbing down runs a risk of injury" is actually a degree of severity. At the moment it looks like a generic info message with players' reaction like "yeah, thanks for stating the obvious, game". I feel it needs some color-coding, or some other means to make players clearly recognize that there are several degrees of severity.

My experience as a player has been that *any* injury resulting from a fall is a pretty serious setback due to the severe pain inflicted, so I'm hesitant to telegraph even the lowest level of injury as mild.  I can see what you mean about text coloring calling the injury level out as _variable_, though!

I will add some color coding to call out the variables in the text.

EvanBalster:
Color coding implemented at @Night-Pryanik's suggestion.

Slip chance: lgreen-lgray-yellow-red-pink-pink
Injury severity: red-red-pink-pink-pink
Climb back up chance: red-ltblue-yellow

Test images below.  I'm not familiar with the colors or any of CDDA's code so please double check that my tags are correct.

image

EvanBalster:
Looks like there are some failing `clang-tidy` tests, but these relate to old-style foreach-loops in code I did not touch with this PR.

EvanBalster:
Here's a table of the messages shown for various thresholds of probability and estimated fall damage.  I've also added this to the summary above.

| Slipping Chance | Description |
| --- | --- |
| -4% or less † | "perfectly safe" |
| -3% to 2% | "very small risk" |
| 3% to 6% | "small risk" |
| 7% to 20% | "medium risk" |
| 21% to 60% | "large risk" |
| 60% or more | "overwhelming risk" |

† As a special case, if estimated damage is less than 5, the "perfectly safe" message will be displayed.  (Falls with less than 5 force deal no damage).

| Estimated Damage | Description |
| --- | --- |
| less than 5 | n/a ("perfectly safe") |
| 5 to 19 | "painful injury" |
| 20 to 39 | "severe injury" |
| 40 to 59 | "grievous injury" |
| 61 to 99 | "life-threatening injury" |
| 100 or more | "instant death" |

Notes:

* There are currently no negative modifiers, so the message will be "very small risk" even if the chance of slipping is in fact zero.
* Fall damage estimation is a simplified version of the `Character::impact` calculation.  It doesn't fully account for armor/bionics/mutations, only those which impact the fall damage multiplier.  By default, fall damage is 10 per Z-level.

akrieger:
Peanut gallery opinion - personally the options are a bit too granular, too close together at the bottom end. It doesn't read like a self-evaluation of risk/danger. What do you think of something like this with a more log scale like stepping. There's 3 steps to 20% risk instead of 4, and 3 steps above that instead of 2.

| risk | message |
| --- | --- |
| <5% | It seems safe to climb down here. |
| 5-10% | It feels a little tricky to climb down here. |
| 10-20% | It seems risky to climb down here. |
| 20-50% | You risk falling if you climb down here. |
| 50-80% | You'll probably fall if you climb down here. |
| 80-100% | It doesn't seem possible to safely climb down here. |

And for damage messaging
| damage | message | 
| --- | --- |
| < 5 | Falling wouldn't be bad. |
| 5 to 19 | Falling would hurt. |
| 20 to 39 | Falling would hurt badly. |
| 40 to 59 | Falling would probably break a limb. |
| 61 to 99 | Falling could cripple or kill you. |
| 100 or more | Falling would kill you. |

Nebnis:
> 10-20%	It seems risky to climb down here.
> 20-50%	You risk falling if you climb down here.
> 50-80%	You'll probably fall if you climb down here.

All sounds the same to me. I would write to

> 10-20%	It seems a bit risky to climb down here.
> 20-50%	Climbing down here feels somewhat risky.
> 50-80%	You'll probably fall if you climb down here.

akrieger:
Could also make it 20-40 & 40-80 for the dividers.

EvanBalster:
I like @akrieger's suggestions, especially the phrasing!  But I'm inclined to use a lower threshold for "seems safe" to avoid tormenting players too much.  Heights are spooky, and there's a significant jump between a 1/100 chance, a 1/50 chance and a 1/33 chance.  Here's how I might tune those numbers:

risk	message
0-2%	It seems safe to climb down here.
3-7%	It seems a little tricky to climb down here.
8-19%	It seems somewhat risky to climb down here.
20-44%	It seems very risky to climb down here.
45-80%	You'll probably slip if you climb down here.
80-100%	It doesn't seem possible to climb down safely.

Remember that the consequences of a first-story fall are pretty harsh for a default character (max pain and about 20% HP).  Based on what I've seen so far, a starting survivor will usually have a slip chance around 9-13% in "good" conditions.

EvanBalster:
I made the changes described above, along with some additional ones:

* Made text and threshold changes as discussed above.
* Broke the prompt message into multiple lines.
* Added a line when grappling or climbing down from multi-story ledges:  `Even if you climb down safely, you will fall at least N stories` (see image)
  * This replaces the previous, slightly misleading message `Looks like N stories.  Jump?`
  * I also consolidated the code for prompting about for multi-story descent with webs.
* Reduced the threshold for "Falling would hurt badly" to 15, so that this message will typically appear when examining a two-story drop.
* Adjusted text color to be more similar to the green-yellow-red scale used for HP.  Pink is still used for extreme risk.

image

While working, I noticed two issues probably outside the scope of this PR:

* Walking into a ledge should probably raise the same prompt as examining a ledge, rather than prompting to jump.
* Already-placed grappling hooks are not used for climbing down — in this state they are as useless as downspouts.  The `has_grapnel` flag in `climb_down` is only set to true if the player is initially holding a grappling hook.

EvanBalster:
Assuming this passes tests/review it should be ready to go now.

Aside from https://github.com/CleverRaven/Cataclysm-DDA/pull/68150 which I've already cooked up, a good follow-up PR would be to rework the `slip_down` functions to accept two new arguments describing what we're using to climb (eg: hook, ladder, webs, downspout, etc) and the type of maneuver (up, down or vaulting).  This would allow grappling hooks and ladders to factor into slip chance rather than bypassing it completely, and would probably facilitate some code cleanup.  It would be much easier to develop/add new climbing functionality that way too.

src/game.cpp
```diff
@@ -13116,53 +13161,116 @@ void game::climb_down( const tripoint &examp )
...
+            // We don't assess the expected damage but this simple warning is provided.
+            hint_remaining_fall = "\n";
+            hint_remaining_fall += string_format( n_gettext(
+                    "Even if you climb down safely, you will fall <color_yellow>at least %d story</color>.",
+                    "Even if you climb down safely, you will fall <color_red>at least %d stories</color>.",
+                    height - 1 ), height - 1 );
+        }
+
+        if( can_use_ladder ) {
+            if( !query_yn( _( "Use the ladder to climb down?" ) ) ) {
```
akrieger, line 13191: I would use `"Climb down the ladder?"`, because the negation to that is 'don't climb <down the ladder>', which is what happens. Negating `"Use the ladder to climb down?"` means 'Don't use the ladder to climb down' but *does* mean to climb down some other way, which is not what happens.

Or even `"There is a ladder below. Climb down the ladder?"` to make it more obvious.

src/game.cpp
```diff
@@ -13116,53 +13161,116 @@ void game::climb_down( const tripoint &examp )
...
-        if( !has_grapnel ) {
-            const char *query;
+        if( !can_use_ladder && !has_grapnel ) {
+            std::string query;
+            std::string hint_slip_risk;
+            std::string hint_fall_damage;
+            std::string hint_climb_back;
             if( web_rappel ) {
-                query = _( "Use your webs to descend?" );
+                query = string_format( n_gettext( "Use your webs to descend?",
```
akrieger, line 13213: I don't think this is an appropriate use of `n_gettext`. Since the singular form does not have any actual variables in it, it won't get translated in any plural way. The implementation of n_gettext has several places that might assert if a plural form can't be found.

You need to break this into separate translation calls, one for N=1 which uses the singular form exclusively, and one n_gettext call for 2+, even though the 'singular' form will never actually get used here.

src/game.cpp
```diff
@@ -13116,53 +13161,116 @@ void game::climb_down( const tripoint &examp )
...
+                if( slip_chance < 3 ) {
+                    hint_slip_risk = _( "It <color_green>seems safe</color> to climb down like this." );
+                } else if( slip_chance < 8 ) {
+                    hint_slip_risk = _( "It <color_yellow>seems a bit tricky</color> to climb down like this." );
+                } else if( slip_chance < 20 ) {
+                    hint_slip_risk = _( "It <color_yellow>seems somewhat risky</color> to climb down like this." );
+                } else if( slip_chance < 50 ) {
+                    hint_slip_risk = _( "It <color_red>seems very risky</color> to climb down like this." );
+                } else if( slip_chance < 80 ) {
+                    hint_slip_risk = _( "It <color_pink>looks like you'll slip</color> if you climb down like this." );
```
akrieger, line 13244: The "like this" phrasing only makes sense contextually if you see the prompts for using a ladder / webs / grapple and refuse them. If you have no ladder, webs, or grapple, then it sounds kind of out of place imo.

akrieger (changes requested):

src/game.cpp
```diff
@@ -13116,53 +13161,116 @@ void game::climb_down( const tripoint &examp )
...
-                               height );
-    } else {
-        query_str = n_gettext( "Looks like %d story.  Nothing your webs can't handle.  Descend?",
-                               "Looks like %d stories.  Nothing your webs can't handle.  Descend?", height );
-    }
 
-    if( height > 1 && !query_yn( query_str, height ) ) {
-        return;
-    } else if( height == 1 ) {
+    {
```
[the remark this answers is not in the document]
EvanBalster: It's a remnant from then the whole block was conditional.  I could remove it; would just change variable scoping a bit.

src/game.cpp
```diff
@@ -13116,53 +13161,116 @@ void game::climb_down( const tripoint &examp )
...
             } else {
-                if( climb_cost <= 0 && fall_mod > 0.8 ) {
-                    query = _( "You probably won't be able to get up and jumping down may hurt.  Jump?" );
-                } else if( climb_cost <= 0 ) {
-                    query = _( "You probably won't be able to get back up.  Climb down?" );
+                // The most common query reads as follows:
+                //   It [seems somewhat risky] to climb down like this.
+                //   Falling [would hurt].
+                //   You [probably won't be able to climb back up].
+                query = "%s%s\n%s\n%s\n";
```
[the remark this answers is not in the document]
EvanBalster: Alternatives were to include a bunch of %s in one of the strings and use it as base (which is not an improvement) or to use std::string concatenation instead of printf-style formatting.  I don't use printf-like functions in my own work so I've got a weak grasp on them.  Open to suggestions.

src/game.cpp
```diff
@@ -13054,21 +13066,54 @@ bool game::slip_down( bool check_for_traps )
...
+            }
+        } else if( stamina_ratio > 0.2 ) {
+            if( show_messages ) {
+                add_msg( m_info, _( "You can't catch your breath, which makes it much more difficult to climb." ) );
+            }
+        } else {
+            if( show_messages ) {
+                add_msg( m_info, _( "You feel faint and can't keep your balance." ) );
+            }
+        }
```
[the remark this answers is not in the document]
EvanBalster: Sensible.  My inclination when I wrote this was to replace all these `if ( show_messages ) add_msg(...)` with a lambda but I get the sense this is discouraged by the current guidelines.  I'll change it as you suggest.

src/game.cpp
EvanBalster: Good suggestion.  The existing implementation handled grappling hooks and webs by prompting about whether you wanted to use them, and then prompting to climb down by hand if you declined.  My phrasing was based on that, but your suggestion is better.

src/game.cpp
EvanBalster: That's how it worked before.  Makes sense.

src/game.cpp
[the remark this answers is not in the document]
akrieger: Lambda would be fine too if typed correctly to accept the post-translation string.

src/game.cpp
akrieger: Which line? I tried to find the original. If it's the same then ignore me cause it clearly works well enough.

src/game.cpp
EvanBalster: I chose "like this" rather than "here" to draw attention to the player's encumbrance, stamina, wetness, free hands and other slip chance factors.  These are printed to the message log before the prompt, giving the player a chance to realize their mistake before attempting e.g. a low-stamina climb.

src/game.cpp
akrieger: If you're talking about line 13131, the original included "Looks like %d story." in the singular form, which would be fine.

src/game.cpp
akrieger: TBH not sure how much someone will notice the extra info in the sidebar even if they have the message log visible by default, but okay.

src/game.cpp
```diff
@@ -13054,21 +13066,54 @@ bool game::slip_down( bool check_for_traps )
...
     }
     add_msg_debug( debugmode::DF_GAME, "Stamina ratio %.2f, final slip chance %d%%",
                    stamina_ratio, slip );
 
     if( weight_ratio >= 1 ) {
```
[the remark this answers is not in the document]
EvanBalster: I don't understand this comment.  Is it about spacing?  CDDA's style doesn't put a space after `if`.

src/game.cpp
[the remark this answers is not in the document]
EvanBalster: After discussing with @akrieger I added a lambda called `add_slip_chance_msg`.  This replaces all the if statements and calls to add_msg.

src/game.cpp
EvanBalster: Yeah, the line "Looks like %d story." was in the `n_gettext` call but would never be displayed in practice, as you suggested.  It was there before so I'll just restore it here.  The player will still see the "use your webs to climb down?" message.

src/game.cpp
EvanBalster: Yeah.  It would be interesting if there were some way to put the log messages from `slip_down_chance` into the prompt in the future.

src/game.cpp
```diff
@@ -13021,11 +13027,17 @@ bool game::slip_down( bool check_for_traps )
...
-        add_msg( m_info, _( "Your wet feet make it harder to climb." ) );
+        if( show_messages ) {
+            add_msg( m_info, _( "Your wet feet make it harder to climb." ) );
+        }
     } else if( wet_hands ) {
-        add_msg( m_info, _( "Your wet hands make it harder to climb." ) );
+        if( show_messages ) {
+            add_msg( m_info, _( "Your wet hands make it harder to climb." ) );
+        }
     }
```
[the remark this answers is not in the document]
EvanBalster: Implemented this suggestion and similarly restructured the other `if( show_messages )` blocks.

src/game.cpp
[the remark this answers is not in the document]
EvanBalster: After FURTHER discussion, no lambda, just tidier conditionals...

EvanBalster:
Rebased on the latest master, which addresses some kind of lighting issue that was making tests very very slow.

Leaving @akrieger's comments about message text (https://github.com/CleverRaven/Cataclysm-DDA/pull/68120#discussion_r1327725440 and https://github.com/CleverRaven/Cataclysm-DDA/pull/68120#discussion_r1327741394) open in case anybody else wants to weigh in but I think the current text is adequate.

EvanBalster:
All tests pass except for two issues that seem unrelated to my changes.

One is a code formatting issue unrelated to my commits.

The other looks like a rare error with a parse function in combat.  Maybe a NaN.  Here's an abridged copy of the text from the log, without all the prefixes...
```
(~[slow] ~[.],starting_items)=> cata_test-tiles is a Catch v2.13.7 host application.
Run with -? for options
Randomness seeded to: 1694934370
Damage_type_EOCs
Attacking a monster
-------
/home/runner/work/Cataclysm-DDA/Cataclysm-DDA/tests/melee_test.cpp:459
.......
/home/runner/work/Cataclysm-DDA/Cataclysm-DDA/tests/melee_test.cpp:365: FAILED:
Unknown expression after the reported line}
due to unexpected exception with message:
   stod

Log messages during failed test:
12:00:00PM: TestCharacter swings wildly and misses.

07:13:47.414 WARNING : opendir [cache] failed with "No such file or directory".282.832 s: starting_items

07:19:54.494 INFO : [i18n] Scanning core translations from lang/mo/30.104 s: weather_realism
30.073 s: eternal_season
```

PR #68125, opened 2023-09-13
[MoM] Add nospell mutation to the anchoring crown

SariusSkelrets:
#### Summary
Mods "[MoM] Wearing the anchoring crown suppresses your spellcasting as intended"

#### Purpose of change

According to the comment in the anchoring crown's code, it's supposed to shut down spellcasting but doesn't since there's no way to do that

Turns out there's a way to do that

#### Describe the solution

Wearing the crown now gives a mutation disabling all spellcasting until you take it off

Tried to place the NO_SPELLCASTING flag directly on the crown but it didn't worked

#### Testing

Wear crown, try to cast spells, can't cast spells

Remove crown, try to cast spells, spells are cast

#### Additional context

None


data/mods/MindOverMatter/items/armor/head.json
```diff
@@ -15,6 +14,7 @@
     "color": "blue",
     "material_thickness": 1,
     "flags": [ "DIMENSIONAL_ANCHOR" ],
+    "relic_data": { "passive_effects": [ { "has": "WORN", "condition": "ALWAYS", "mutations": [ "ANCHORCROWN_NOSPELL" ] } ] },
```
GuardianDll, line 17: Are you sure it works? I recall a longstanding issue, that mutations, given by enchantment (which relic_data is) is shown in @ menu, but have no effect whatsoever

Standing-Storm:
Could you also document the NO_SPELLCASTING flag, please? It’s not in any of the docs it looks like

data/mods/MindOverMatter/items/armor/head.json
SariusSkelrets: I did test it and it worked. Wearing the crown prevented me from using any spell at all, then removing it gave me my spellcasting back

Standing-Storm: Working on #68140, it looks like enchantment mutations given by items work, but enchantment mutations given by effects do not.

Issue #50805, opened 2021-08-15
Faction camp food stores always uses default calories of comestible, not actual calories


RenechCDDA:
### Describe the bug

See steps to reproduce

### Steps To Reproduce

Debugged campfire hot dog is 130 calories
Campfire hot dog made out of debugged bratwurst is 297 calories (inherits calories from sole component)
Put bratwurst-turned-hot-dog(297 calories) into camp food stores, you get only 130 calories
You are missing 167 calories

### Expected behavior

You are not missing 167 calories

### Screenshots
These two campfire hot dogs give the same amount of calories(130 each) to the camp food stores, despite having different calorie amounts when consumed.
![Untitled](https://user-images.githubusercontent.com/84619419/129469468-f429eab1-4618-4f60-a0fc-efe99c802684.png)

### Versions and configuration

70df41c
(0.F-1)

### Additional context

I believe this is the issue:

https://github.com/CleverRaven/Cataclysm-DDA/blob/ee0b7b6d6bc1f0ed6d1ed947fe16b4ba7ff9b5a2/src/faction_camp.cpp#L3823

Instead of `get_comestible()->default_nutrition.kcal()` we need something like `Character::compute_effective_nutrients`.  Except it probably shouldn't take from the avatar classes at all, because you're feeding generic NPCs. Assuming they are 'normal' is a decent approximation.

PR #68127, opened 2023-09-13
[CR] Camp calorie test

RenechCDDA:
#### Summary
None

#### Purpose of change
Camps are woefully underserved by our test suite.

#### Describe the solution
Make tests for things that have caused issues in the past and could come up again (e.g. #50805)

Let someone else write the tests? Hasn't worked so far!

#### Testing
This runs and works locally, both as an individual test, running just those two tests(this one and EOC_transform_radius) together, and running the whole suite.

#### Additional context
Ran into some problems with test ordering, see comments.

### DIFF
--- a/tests/eoc_test.cpp
+++ b/tests/eoc_test.cpp
@@ -249,15 +249,6 @@ TEST_CASE( "EOC_transform_radius", "[eoc][timed_event]" )
     calendar::turn += 2_seconds;
     get_timed_events().process();
     check_ter_in_radius( start, eoc_range, t_grass );
-    CHECK( "Cleaning up avatar's pos and location after successful test…" );
-    // Bizarre point handling ahead, sanity beware...
-    // Without this, we have an offset of {-12,-12, 0 } between player's pos and location.
-    tripoint_abs_omt return_omt;
-    point_omt_ms return_square;
-    std::tie( return_omt, return_square ) = project_remain<coords::omt>( start );
-    g->place_player_overmap( return_omt );
-    g->place_player( tripoint {start.x() + return_square.x(), start.y() + return_square.y(), 0} );
-    REQUIRE( get_map().inbounds( start ) );
 }
 
 TEST_CASE( "EOC_transform_line", "[eoc][timed_event]" )
--- a/tests/eoc_test.cpp
+++ b/tests/eoc_test.cpp
@@ -249,6 +249,15 @@ TEST_CASE( "EOC_transform_radius", "[eoc][timed_event]" )
     calendar::turn += 2_seconds;
     get_timed_events().process();
     check_ter_in_radius( start, eoc_range, t_grass );
+    CHECK( "Cleaning up avatar's pos and location after successful test..." );
+    // Bizarre point handling ahead, sanity beware...
+    // Without this, we have an offset of {-12,-12, 0 } between player's pos and location.
+    tripoint_abs_omt return_omt;
+    point_omt_ms return_square;
+    std::tie( return_omt, return_square ) = project_remain<coords::omt>( start );
+    g->place_player_overmap( return_omt );
+    g->place_player( tripoint {start.x() + return_square.x(), start.y() + return_square.y(), 0} );
+    REQUIRE( get_map().inbounds( start ) );
 }
 
 TEST_CASE( "EOC_transform_line", "[eoc][timed_event]" )
--- a/tests/eoc_test.cpp
+++ b/tests/eoc_test.cpp
@@ -249,7 +249,7 @@ TEST_CASE( "EOC_transform_radius", "[eoc][timed_event]" )
     calendar::turn += 2_seconds;
     get_timed_events().process();
     check_ter_in_radius( start, eoc_range, t_grass );
-    CHECK( "Cleaning up avatar's pos and location after successful test..." );
+    CHECK( "Cleaning up avatar's pos and location after successful test…" );
     // Bizarre point handling ahead, sanity beware...
     // Without this, we have an offset of {-12,-12, 0 } between player's pos and location.
     tripoint_abs_omt return_omt;
--- a/tests/faction_camp_test.cpp
+++ b/tests/faction_camp_test.cpp
@@ -22,22 +22,23 @@ TEST_CASE( "camp_calorie_counting", "[camp]" )
     clear_avatar();
     clear_map();
     map &m = get_map();
-    const tripoint zone_loc = m.getabs( tripoint{ 5, 5, 0 } );
-    mapgen_place_zone( zone_loc, zone_loc, zone_type_CAMP_FOOD, your_fac, {},
+    const tripoint_abs_ms zone_loc = m.getglobal( tripoint{ 5, 5, 0 } );
+    mapgen_place_zone( zone_loc.raw(), zone_loc.raw(), zone_type_CAMP_FOOD, your_fac, {},
                        "food" );
-    mapgen_place_zone( zone_loc, zone_loc, zone_type_CAMP_STORAGE, your_fac, {},
+    mapgen_place_zone( zone_loc.raw(), zone_loc.raw(), zone_type_CAMP_STORAGE, your_fac, {},
                        "storage" );
     faction *camp_faction = get_player_character().get_faction();
-    const tripoint_abs_omt this_omt = project_to<coords::omt>( m.getglobal( zone_loc ) );
+    const tripoint_abs_omt this_omt = project_to<coords::omt>( zone_loc );
     m.add_camp( this_omt, "faction_camp" );
     std::optional<basecamp *> bcp = overmap_buffer.find_camp( this_omt.xy() );
     basecamp *test_camp = *bcp;
     WHEN( "a base item is added to larder" ) {
         camp_faction->food_supply = 0;
         item test_100_kcal( "test_100_kcal" );
-        m.i_clear( zone_loc );
-        m.add_item_or_charges( zone_loc, test_100_kcal );
-        REQUIRE( m.has_items( zone_loc ) );
+        tripoint_bub_ms zone_local = m.bub_from_abs( zone_loc );
+        m.i_clear( zone_local.raw() );
+        m.add_item_or_charges( zone_local, test_100_kcal );
+        REQUIRE( m.has_items( zone_local ) );
         test_camp->distribute_food();
         CHECK( camp_faction->food_supply == 100 );
     }
@@ -51,8 +52,9 @@ TEST_CASE( "camp_calorie_counting", "[camp]" )
         made_of.add( test_100_kcal );
         // Setting the actual components. This will return 185 unless it's actually made up of two 100kcal components!
         test_200_kcal.components = made_of;
-        m.i_clear( zone_loc );
-        m.add_item_or_charges( zone_loc, test_200_kcal );
+        tripoint_bub_ms zone_local = m.bub_from_abs( zone_loc );
+        m.i_clear( zone_local.raw() );
+        m.add_item_or_charges( zone_local, test_200_kcal );
         test_camp->distribute_food();
         CHECK( camp_faction->food_supply == 200 );
     }
--- a/tests/faction_camp_test.cpp
+++ b/tests/faction_camp_test.cpp
@@ -19,6 +19,7 @@ static const zone_type_id zone_type_CAMP_STORAGE( "CAMP_STORAGE" );
 
 TEST_CASE( "camp_calorie_counting", "[camp]" )
 {
+    clear_avatar();
     clear_map();
     map &m = get_map();
     const tripoint zone_loc = m.getabs( tripoint{ 5, 5, 0 } );
@@ -26,7 +27,6 @@ TEST_CASE( "camp_calorie_counting", "[camp]" )
                        "food" );
     mapgen_place_zone( zone_loc, zone_loc, zone_type_CAMP_STORAGE, your_fac, {},
                        "storage" );
-    get_player_character().setpos( zone_loc );
     faction *camp_faction = get_player_character().get_faction();
     const tripoint_abs_omt this_omt = project_to<coords::omt>( m.getglobal( zone_loc ) );
     m.add_camp( this_omt, "faction_camp" );
--- a/tests/faction_camp_test.cpp
+++ b/tests/faction_camp_test.cpp
@@ -4,7 +4,6 @@
 #include <vector>
 
 #include "basecamp.h"
-#include "calendar.h"
 #include "cata_catch.h"
 #include "clzones.h"
 #include "coordinates.h"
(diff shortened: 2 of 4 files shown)


RenechCDDA:
I cannot get the character to craft, and I do not know why. I haven't pushed these commits but I've tried:
Giving the items directly to their inventory
Giving them a debug backpack to make sure they have space
Checking they did in fact get the component items (they did)

Any help would be appreciated.

PatrikLundell:
I don't know anything about the testing stuff, so it might not be any help...

My interpretation of the code is that it starts crafting, but does not advance time to or go through any process that would finish it before trying to use the product. There is time advancement AFTER the call to setup_test_craft, but that would be too late.

RenechCDDA:
> I don't know anything about the testing stuff, so it might not be any help...
> 
> Edit: Replaced what was below with something that might be less incorrect.
> 
> My interpretation of the code is that it starts crafting and verified the in process craft is wielded. It then just advances the time, but doesn't make any call that would actually cause the crafting to finalize.

It actually did try to start the craft, it just repeatedly refused to acknowledge that the character had components on hand. It was very bizarre.

Regardless, I have abandoned the "actually craft it" approach in favor of just inserting the components into the item via code.

RenechCDDA:
Aaaand it's not playing well on CI. Worked fine on my machine!

This is going to be an annoying one, isn't it?

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+        made_of.add( test_100_kcal );
+        // Setting the actual components. This will return 185 unless it's actually made up of two 100kcal components!
+        test_200_kcal.components = made_of;
+        m.i_clear( zone_loc );
+        m.add_item_or_charges( zone_loc, test_200_kcal );
+        test_camp->distribute_food();
+        CHECK( camp_faction->food_supply == 200 );
+    }
+}
+// TODO: Tests for: Check calorie display at various activity levels, camp crafting works as expected (consumes inputs, returns outputs+byproducts, costs calories)
```
ehughsbaird, line 60:
```suggestion
// TODO: Tests for: Check calorie display at various activity levels, camp crafting works as expected (consumes inputs, returns outputs+byproducts, costs calories)

```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "player_helpers.h"
+
+static const zone_type_id zone_type_CAMP_FOOD( "CAMP_FOOD" );
+static const zone_type_id zone_type_CAMP_STORAGE( "CAMP_STORAGE" );
+
+TEST_CASE( "camp_calorie_counting", "[camp]" )
+{
+    clear_map();
+    map &m = get_map();
+    tripoint const zone_loc = m.getabs( tripoint{ 5, 5, 0 } );
```
ehughsbaird, line 24: For consistent style
```suggestion
    const tripoint zone_loc = m.getabs( tripoint{ 5, 5, 0 } );
```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+    mapgen_place_zone( zone_loc, zone_loc, zone_type_CAMP_FOOD, your_fac, {},
+                       "food" );
+    mapgen_place_zone( zone_loc, zone_loc, zone_type_CAMP_STORAGE, your_fac, {},
+                       "storage" );
+    get_player_character().setpos( zone_loc );
```
ehughsbaird, line 29: I think this could cause a map shift, meaning the `zone_loc` used below is in the wrong place. You should get the global coords before doing this.

RenechCDDA:
I assumed that the failure to place items into `zone_loc` was possibly due to it being unloaded (player character not necessarily in the same map, although weird that it works reliably on my machine). Setting the player's pos was supposed to resolve that, but clearly it hasn't.... I wonder if this a test ordering issue?

ehughsbaird:
With https://github.com/CleverRaven/Cataclysm-DDA/issues/68154 figured out, it's a conflict with the `EOC_transform_radius` test. `tests/cata_test "EOC_transform_radius,camp_calorie_counting"` will consistently reproduce it.

RenechCDDA:
Looks like it specifically places the player out of bounds during that test, yeah. I am open to suggestions on how to resolve this that isn't "Just run the faction camp test first".

ehughsbaird:
That test should probably place the player back in bounds (do this), or this test should place the player inbounds first (leaving that test to potentially break other tests).

RenechCDDA:
That does sound like a fine solution when I think about it hah. I'll poke it sometime today.

RenechCDDA:
Ok so this runs and works locally, both as an individual test, running just those two tests together, and running the whole suite (sans the expected translation failures).

But I would appreciate a second look so it doesn't look like I'm doing something really nuts.

tests/eoc_test.cpp
```diff
@@ -246,6 +246,15 @@ TEST_CASE( "EOC_transform_radius", "[eoc][timed_event]" )
     calendar::turn += 2_seconds;
     get_timed_events().process();
     check_ter_in_radius( start, eoc_range, t_grass );
+    CHECK( "Cleaning up avatar's pos and location after successful test..." );
```
ehughsbaird, line 249: Are you meaning to use something other than CHECK? This is an odd use of it.

tests/eoc_test.cpp
```diff
@@ -246,6 +246,15 @@ TEST_CASE( "EOC_transform_radius", "[eoc][timed_event]" )
...
+    // Bizarre point handling ahead, sanity beware...
+    // Without this, we have an offset of {-12,-12, 0 } between player's pos and location.
+    tripoint_abs_omt return_omt;
+    point_omt_ms return_square;
+    std::tie( return_omt, return_square ) = project_remain<coords::omt>( start );
+    g->place_player_overmap( return_omt );
+    g->place_player( tripoint {start.x() + return_square.x(), start.y() + return_square.y(), 0} );
```
ehughsbaird, line 256: I wonder if you can just use setpos? I haven't looked at what these functions do.

tests/eoc_test.cpp
RenechCDDA: In case it fails due to future changes, this was intended to be somewhat informative output. (I meant to use CHECK_NOFAIL, oops)

tests/eoc_test.cpp
RenechCDDA: You cannot. I do not really understand why, but pos and location remain desynced with setpos. The 12 coordinate difference is (as far as I can tell) due to the conversion to OMT scale at line 240.

mqrause: What do you mean pos and location are desynced? Aren't they the same thing?

tests/eoc_test.cpp
RenechCDDA: I'm marking this resolved because we seem to have cleared that up and the change to CHECK_NOFAIL is only relevant to style, but *please do* open another review/comment if you have any concern. It is greatly appreciated.

dseguin, line 249:
```suggestion
    CHECK( "Cleaning up avatar's pos and location after successful test…" );
```

RenechCDDA:
OK, so now it consistently passes on windows and consistently fails on... whatever the `GCC 9, Curses, LTO ` step runs.

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+static const zone_type_id zone_type_CAMP_FOOD( "CAMP_FOOD" );
+static const zone_type_id zone_type_CAMP_STORAGE( "CAMP_STORAGE" );
+
+TEST_CASE( "camp_calorie_counting", "[camp]" )
+{
+    clear_avatar();
+    clear_map();
+    map &m = get_map();
+    const tripoint zone_loc = m.getabs( tripoint{ 5, 5, 0 } );
```
kevingranade, line 25: This is an abs tripoint and needs to be typed as such,  tripoint_abs_ms.
If it had been,  the compiler could have told you about the errors on lines 38-40 abd 54-55 where you pass an abs coord to local methods.

kevingranade, line 25:
```suggestion
    const tripoint_abs_ms zone_loc = m.getabs( tripoint{ 5, 5, 0 } );
```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+    const tripoint_abs_omt this_omt = project_to<coords::omt>( m.getglobal( zone_loc ) );
+    m.add_camp( this_omt, "faction_camp" );
+    std::optional<basecamp *> bcp = overmap_buffer.find_camp( this_omt.xy() );
+    basecamp *test_camp = *bcp;
+    WHEN( "a base item is added to larder" ) {
+        camp_faction->food_supply = 0;
+        item test_100_kcal( "test_100_kcal" );
+        m.i_clear( zone_loc );
+        m.add_item_or_charges( zone_loc, test_100_kcal );
+        REQUIRE( m.has_items( zone_loc ) );
```
kevingranade, line 40:
```suggestion
        tripoint_bub_ms zone_local = m.getlocal( zone_loc );
        m.i_clear( zone_local );
        m.add_item_or_charges( zone_local, test_100_kcal );
        REQUIRE( m.has_items( zone_local ) );
```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+        camp_faction->food_supply = 0;
+        item test_100_kcal( "test_100_kcal" );
+        item test_200_kcal( "test_200_kcal" );
+        item_components made_of;
+        made_of.add( test_100_kcal );
+        made_of.add( test_100_kcal );
+        // Setting the actual components. This will return 185 unless it's actually made up of two 100kcal components!
+        test_200_kcal.components = made_of;
+        m.i_clear( zone_loc );
+        m.add_item_or_charges( zone_loc, test_200_kcal );
```
kevingranade, line 55:
```suggestion
        tripoint_bub_ms zone_local = m.getlocal( zone_loc );
        m.i_clear( zone_local );
        m.add_item_or_charges( zone_local, test_200_kcal );
```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+static const zone_type_id zone_type_CAMP_FOOD( "CAMP_FOOD" );
+static const zone_type_id zone_type_CAMP_STORAGE( "CAMP_STORAGE" );
+
+TEST_CASE( "camp_calorie_counting", "[camp]" )
+{
+    clear_avatar();
+    clear_map();
+    map &m = get_map();
+    const tripoint_abs_ms zone_loc = m.getabs( tripoint{ 5, 5, 0 } );
```
kevingranade, line 25:
```suggestion
    const tripoint_abs_ms zone_loc = m.getglobal( tripoint{ 5, 5, 0 } );
```

tests/faction_camp_test.cpp
RenechCDDA: Hm I copied that wholesale from zones_custom_test.cpp, it seemed suspect to me but that hasn't been having any issues that I've seen.

https://github.com/CleverRaven/Cataclysm-DDA/blob/078a7b43be36c11f2431bf52f5f00d2d41c9a53a/tests/zones_custom_test.cpp#L14-L15

They seem to be doing it for the same reason - `mapgen_place_zone` only accepts raw tripoints and not typed points(!!!). (Not at my desktop right now so dumb question time) Is it possible to take our typed tripoint (tripoint_abs_ms/tripoint_bub_ms/etc) and turn it into a raw tripoint? There was some sticking point I can't recall, but it seems like something that should be easy.

Also judging by your messages in the dev discord's #technical-discussion channel I assume the chances to eoc_test.cpp should be reverted?

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+static const zone_type_id zone_type_CAMP_FOOD( "CAMP_FOOD" );
+static const zone_type_id zone_type_CAMP_STORAGE( "CAMP_STORAGE" );
+
+TEST_CASE( "camp_calorie_counting", "[camp]" )
+{
+    clear_avatar();
+    clear_map();
+    map &m = get_map();
+    const tripoint_abs_ms zone_loc = m.getglobal( tripoint{ 5, 5, 0 } );
+    mapgen_place_zone( zone_loc, zone_loc, zone_type_CAMP_FOOD, your_fac, {},
```
kevingranade, line 26:
```suggestion
    mapgen_place_zone( zone_loc.raw(), zone_loc.raw(), zone_type_CAMP_FOOD, your_fac, {},
```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+                       "food" );
+    mapgen_place_zone( zone_loc, zone_loc, zone_type_CAMP_STORAGE, your_fac, {},
```
kevingranade, line 28:
```suggestion
    mapgen_place_zone( zone_loc.raw(), zone_loc.raw(), zone_type_CAMP_STORAGE, your_fac, {},
```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+                       "storage" );
+    faction *camp_faction = get_player_character().get_faction();
+    const tripoint_abs_omt this_omt = project_to<coords::omt>( m.getglobal( zone_loc ) );
```
kevingranade, line 31:
```suggestion
    const tripoint_abs_omt this_omt = project_to<coords::omt>( zone_loc );
```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+    m.add_camp( this_omt, "faction_camp" );
+    std::optional<basecamp *> bcp = overmap_buffer.find_camp( this_omt.xy() );
+    basecamp *test_camp = *bcp;
+    WHEN( "a base item is added to larder" ) {
+        camp_faction->food_supply = 0;
+        item test_100_kcal( "test_100_kcal" );
+        tripoint_bub_ms zone_local = m.getlocal( zone_loc );
```
kevingranade, line 38:
```suggestion
        tripoint_bub_ms zone_local = m.bub_from_abs( zone_loc );
```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+        m.i_clear( zone_local );
```
kevingranade, line 39:
```suggestion
        m.i_clear( zone_local.raw() );
```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+    WHEN( "an item with inherited components is added to larder" ) {
+        camp_faction->food_supply = 0;
+        item test_100_kcal( "test_100_kcal" );
+        item test_200_kcal( "test_200_kcal" );
+        item_components made_of;
+        made_of.add( test_100_kcal );
+        made_of.add( test_100_kcal );
+        // Setting the actual components. This will return 185 unless it's actually made up of two 100kcal components!
+        test_200_kcal.components = made_of;
+        tripoint_bub_ms zone_local = m.getlocal( zone_loc );
```
kevingranade, line 55:
```suggestion
        tripoint_bub_ms zone_local = m.bub_from_abs( zone_loc );
```

tests/faction_camp_test.cpp
```diff
@@ -0,0 +1,62 @@
...
+        m.i_clear( zone_local );
```
kevingranade, line 56:
```suggestion
        m.i_clear( zone_local.raw() );
```

tests/eoc_test.cpp
```diff
@@ -249,6 +249,15 @@ TEST_CASE( "EOC_transform_radius", "[eoc][timed_event]" )
...
     check_ter_in_radius( start, eoc_range, t_grass );
+    CHECK( "Cleaning up avatar's pos and location after successful test…" );
+    // Bizarre point handling ahead, sanity beware...
+    // Without this, we have an offset of {-12,-12, 0 } between player's pos and location.
+    tripoint_abs_omt return_omt;
+    point_omt_ms return_square;
+    std::tie( return_omt, return_square ) = project_remain<coords::omt>( start );
+    g->place_player_overmap( return_omt );
+    g->place_player( tripoint {start.x() + return_square.x(), start.y() + return_square.y(), 0} );
+    REQUIRE( get_map().inbounds( start ) );
```
kevingranade, line 260:
```suggestion
```

PR #68128, opened 2023-09-13
[Magiclysm] Add dispel magic spells

Standing-Storm:
#### Summary
Mods "Add dispel magic spells"

#### Purpose of change

I keep seeing people posting about stacking a ton of buffs and ~~I am here to ruin their fun~~what kind of magical world doesn't have wizard duels where removing enchantments is part of the duel? 

#### Describe the solution

Add Dispel Magic and Disjunction spells. Dispel Magic costs less and is easier to cast but doesn't work on spells Difficulty 7 or above or any spell from an attunement. Disjunction works on anything. This distinction doesn't matter that much now but as more enemy wizards get added it should be more important.

In addition to scrolls for these spells, I added dispel magic as a possible spell for the feral wizard and disjunction as a spell for the feral radiation magic (and am planning to add it to the mi-go nethermancer).

Disjunction does not destroy your enchanted items because it would take 100s (1000s?) of lines of EoC to set that up.

Both spells are implemented as effects so you could theoretically have some kind of "anti-dispel shield" that prevented you from having Dispel Magic cast on you and require casting, I don't know, "Pierce Shield" or something first, but I haven't implemented any of that. 

#### Testing

Spawned a feral wizard and he dispelled my darksight.  Cast darksight and dispelled my own darksight.  Cast haste and dispel didn't work to dispel it. 

#### Additional context

Despite all the fun I had in Baldur's Gate II with the Sword Coast Stratagems mod, I decided against various different dispel spells and effects vulnerable to certain spells but not others because enemies' AI can't handle that.  Maybe in the future.


Zireael07:
Sword Coast Stratagems aside, most fantasy RPGs only have one dispel magic spell and it's enough <3

MNG-cataclysm:
If an effect from an attunement is removed, will it eventually return, or is it permanently gone? If it's gone forever, such an effect would be far too powerful and frustrating because of the one-time nature of attunement application.

GuardianDll:
Gone until you recast it

Standing-Storm:
> If an effect from an attunement is removed, will it eventually return, or is it permanently gone? If it's gone forever, such an effect would be far too powerful and frustrating because of the one-time nature of attunement application.

I believe I only targeted active buffs (that you have to cast). If not, it's unintended and should be fixed.

PR #68133, opened 2023-09-14
Make loaded stick ``looks_like`` the shillelagh

Karol1223:
#### Summary
None

#### Purpose of change
I've been annoying MSX tileset artists for months to give loaded stick a wielded sprite when the truth is shillelagh already had one and the two would look almost identical anyway...


gettingusedto:
Can ya also add the looks_like for `shillelagh_fake` cuz that also didnt looks like shillelagh.

PR #68138, opened 2023-09-14
Fix visibility not updating properly when moving, peeking and autodriving

### LINKED ISSUE #68121: Player character occasionally see through walls and reveal the stuff behind it
### Describe the bug

For some reason, player can seem to reveal stuff behind wall and get the map memory from it. It can happen by just walking beside the walls.

### Attach save file

[Bates-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12596220/Bates-trimmed.tar.gz)

### Steps to reproduce

1. open the attached save
2. walk beside the wall for a bit
3. see that it reveals some part of the dirt wall beyond the concrete walls

### Expected behavior

Player not having a crappy version of clairvoyance by default.

### Screenshots

Before (world used here is the one i attached to this bug report
Screenshot_2023-09-13_17-03-57

After
Screenshot_2023-09-13_17-05-17

- OS: Android
- Game Version: f12487c [32-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Custom Frequencey For Portal Storms [custom_freq_ps],
    Custom Portal Storm Size [custom_size_ps]
]

### Additional context

Also in my current run, i noticed that the line of sight when autowalking looks quite weird. This could be related to this issue, though i cant replicate it in the new fresh world somehow.
Screenshot_2023-09-13_17-20-34

My current run
[Cowell-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12596314/Cowell-trimmed.tar.gz)

### LINKED ISSUE #68134: No vision when Peeking
### Describe the bug

when "peeking" to check around corners or up/down elevations, the player's vision does not update, effectively breaking the feature.

### Attach save file

[Ropesville-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12609927/Ropesville-trimmed.tar.gz)

### Steps to reproduce

1.spawn into the world
2. find a wall to peek around or stairs to peek down/up where vision is blocked by the object in front of the player
3. press 'X' to peek and pick a direction that would remove the obstruction

### Expected behavior

When peeking around corners and up/down elevation, the player should be able to see in that direction.

### Screenshots

peeking
not peeking

- OS: Windows
    - OS Version: 10.0.19044.3086 (21H2)
- Game Version: e8a6beb [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

probably related to issue https://github.com/CleverRaven/Cataclysm-DDA/issues/68121 and could be caused by https://github.com/CleverRaven/Cataclysm-DDA/pull/67997

### LINKED ISSUE #68174: Visibility does not update while reading
### Describe the bug

Visibility does not update while reading.

### Attach save file

[Lorman-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12642109/Lorman-trimmed.tar.gz)

### Steps to reproduce

1. Start reading a book shortly before sundown.
2. Keep reading well into the night despite the fact that it should be too dark.
3. Stop reading in the middle of the night.
4. Try to read again.
5. Fail.

### Expected behavior

Visibility should update while reading to prevent the ability to keep reading into the night despite it being too dark to do so.

### Screenshots

daylightnoupdate

- OS: Windows
- Game Version: bb0bdee [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

Only recent item I found that could be related is: https://github.com/CleverRaven/Cataclysm-DDA/pull/68155

Rewryte:
#### Summary
Bugfixes "Fix visibility not updating properly when moving, peeking, and autodriving"

#### Purpose of change
#67997 changes map and visibility cache updates from being called every draw to being called on input.  Unfortunately, after movement, the cache update does not happen fast enough and a draw is done using an outdated visibility cache which can reveal previously unseen tiles.
Visibility is also not updated when peeking and autodriving.
Fixes #68121. Fixes #68134. Fixes #68174.

#### Describe the solution
Add a map and visibility cache update right after movement, while peeking, and while autodriving.

#### Testing
Tested with save files attached with #68121 and #68134.
Walking clairvoyance is fixed. Visibility during activities, driving, peeking, and looking around is updating correctly.


andrei8l:
I hate this solution. These are basically special cases peppered across the code.

Why not put `map::update_visibility_cache()` back in the draw call and add a cache invalidation check in `map::update_visibility_cache()` similar to the one in [`map::build_map_cache()`](https://github.com/CleverRaven/Cataclysm-DDA/blob/d6d150c655ea73070f68f7a8855173cc2f565092/src/map.cpp#L9200)? And check for a couple of other things like turn and weather.

Rewryte:
> Why not put `map::update_visibility_cache()` back in the draw call and add a cache invalidation check in `map::update_visibility_cache()` similar to the one in [`map::build_map_cache()`](https://github.com/CleverRaven/Cataclysm-DDA/blob/d6d150c655ea73070f68f7a8855173cc2f565092/src/map.cpp#L9200)?

Wouldn't that just pepper the same spots with invalidate functions instead?  Like how `map::invalidate_map_cache` is used.

andrei8l:
> Wouldn't that just pepper the same spots with invalidate functions instead? Like how `map::invalidate_map_cache` is used.

No, I don't think so, since all the cases here and from #67997 would be covered by position and time checks.  `map::invalidate_map_cache()` is primarily used when the map changes in some way and I suspect some of the other uses are deprecated.

Rewryte:
> all the cases here and from #67997 would be covered by position and time checks

Ok, I see where you're going with this now.  I'll put it into code when I find the time.

misterprimus:
Related to https://github.com/CleverRaven/Cataclysm-DDA/issues/68174 ?

Rewryte:
> Related to #68174 ?

I anticipated such an issue in https://github.com/CleverRaven/Cataclysm-DDA/pull/68155#issuecomment-1721641465.
It should be resolved once I reimplement this fix as andrei8l suggested.

Ivan-Shestakov:
Looks well, I'm running build from this branch and so far is fine...

KestasV:
> Looks well, I'm running build from this branch and so far is fine...

Dunno what you're talking about, i can still see through walls and areas i shouldn't see in.

Leonkoithara:
My testing:
1. Performing actions now do update the visibility cache. So issues like #68174 should be fixed.
2. Driving vehicles now do update the visibility fast enough that the issue where the character lags behind point of view is fixed.
3. Peeking is not fixed.
4. Viewing areas that should not be visibile is still persistant.
5. Jumps in point of view also happens sometimes(possibly what causes 4).

Rewryte:
Commits b52cd59 and 37f66fd were part of an experimental optimization that failed to achieve the expected performance improvement and has been reverted.

This branch as of commit 8d4f586 should be bug-free but unoptimized as map_cache and visibility_cache are calculated every draw (This was the default behavior before #67997).

Calling `draw_map_cache` every draw will result in a performance loss even when no caches are dirty.  More testing is required to determine if we could call it elsewhere without it being everywhere.

Almost all cases where `update_visibility_cache` is needed can be covered by calling it when a seen_cache is found to be dirty.  One exception would be during player activities, so it is not exactly the silver bullet I'd hoped it would be.

Brambor:
If it is only player activity, can something be changed in the activity handler?
https://github.com/CleverRaven/Cataclysm-DDA/blob/83978be0a463679b9178df0a9b2067dd04c294ee/src/activity_type.cpp#L155-L162

I don't know much about the code, but I expect this to be handling all player activity. So if there were separate problems with driving, sunrise etc. which all happened during player activity, could a change in this one place solve it?

I don't know whether it changes how activities are written, if you need to make them into a generator (executing only so much code before yielding to a redraw), or if it is just a simple change. Or whether it's impossible at all. But I wouldn't want you to miss this, since you worked on this for so long.

Rewryte:
> If it is only player activity, can something be changed in the activity handler?

Think the issue is that there is a need for a separate call in the first place.  Inefficient as it may, nothing beats the simplicity of brute forcing calculations every single draw.  The more specific our conditions for recalculations, the more separate calls become needed.

For now I'll just follow andrei8l's suggestion for position and time checks and see where to go from there.

misterprimus:
> > If it is only player activity, can something be changed in the activity handler?
> 
> Think the issue is that there is a need for a separate call in the first place. Inefficient as it may, nothing beats the simplicity of brute forcing calculations every single draw. The more specific our conditions for recalculations, the more separate calls become needed.
> 
> For now I'll just follow andrei8l's suggestion for position and time checks and see where to go from there.

Yes it's better to just revert it for now. The performance loss was significant, but not immense, and is better than something as important as vision cache being bugged. This part of the code doesn't get edited very often, so perhaps it's best to revert for now and then spend time testing a permanent solution.

Maleclypse:
> > If it is only player activity, can something be changed in the activity handler?
> 
> Think the issue is that there is a need for a separate call in the first place. Inefficient as it may, nothing beats the simplicity of brute forcing calculations every single draw. The more specific our conditions for recalculations, the more separate calls become needed.
> 
> For now I'll just follow andrei8l's suggestion for position and time checks and see where to go from there.

This seems to be a good path to follow.  Thanks!

Rewryte:
Ok, should be good to go.

Optimizations to map cache has been reverted.
Visibility cache now uses an invalidation system similar to the other caches.  It is recalculated on each turn, player location change, peeking, and looking around.

Walking clairvoyance is fixed.  Visibility during activities, driving, peeking, and looking around is updating correctly.

ehughsbaird (approved):
Tests regression fixed:
Test took 12 minutes to run
Performance analysis does not show visibility cache as contributor

Maleclypse:
Tested and it works.  But ODB has already merged :)

misterprimus:
> Tested and it works. But ODB has already merged :)

ODB?

PR #68139, opened 2023-09-14
Compare recipe result with inventory item

inogenous:
#### Summary
Interface "Compare recipe result with inventory item"

#### Purpose of change

* Adds ability to press `I` (capital i) when in `&` crafting screen to compare the result of a recipe with an item from the inventory.
* Use case: "Would this new thing be better than what I have now?"

#### Describe the solution

* Uses `I` keybinding, which is the same keybind as the regular "compare two items".
* Moves code from `recipe_result_info_cache::get_result_data` into new function `get_recipe_result_item` so that the comparison can use the correct (fitting) variant of clothing items.
* Reuses existing code in `inventory_pick_selector` for picking an items from inventory to compare to.
* Reuses existing code in `game_menus::inv::compare_items` to show the comparison.

* We might also want to compare recipes not only to items in inventory but also to adjacent items.
* Might also be interesting to compare two recipes to each other, but that might need a totally different ui.

#### Testing

* Compared some recipes to items in inventory
* Immediately cancelling the popup that asks for which item to compare: works.
* Pressing `I` when on a nested crafting category: prompts for selecting a recipe
* Pressing `I` when completely naked and no items in inventory: shows a popup stating that.

compare-recipe-to-item-1

compare-recipe-to-item-2

compare-recipe-to-item-3

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -758,6 +758,13 @@
   },
   {
     "type": "keybinding",
+    "id": "COMPARE",
+    "category": "CRAFTING",
+    "name": "Compare recipe to item in inventory",
+    "bindings": [ { "input_method": "keyboard_char", "key": "I" }, { "input_method": "keyboard_code", "key": "i", "mod": [ "shift" ] } ]
+  },
+  {
+    "type": "keybinding",
     "id": "LEVEL_UP",
     "name": "Go up",
     "bindings": [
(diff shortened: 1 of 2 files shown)


ilovehans10:
This is a top-notch addition! As an idea, something similar would be great for the bartering screen.

pjf:
I never even knew I wanted this, but it turns out I did. THANK YOU, @inogenous!

PR #68146, opened 2023-09-15
Remove token bash damage from knives

Maleclypse:
#### Summary
None

#### Purpose of change
Removes token bash damage from knives.  

#### Describe the solution
Removes bash from low weight knives where it makes up a third or less of the damage.

None


Fris0uman:
Looks like you need to update some tests

Maleclypse:
> Looks like you need to update some tests

Yeah, I'm trying to decide if I want to update the tests, add additional pierce damage, or a combination of both.

fairyarmadillo:
> > Looks like you need to update some tests
> 
> Yeah, I'm trying to decide if I want to update the tests, add additional pierce damage, or a combination of both.

Most of the knives already do a crapton of stab damage, and could stand to be reduced.

Karol1223:
Could you edit the affected weapons for ferals as well, while you're at it? I know this is going to affect the combat knife special attack, not sure if the other knife is going to be affected or not.

Maleclypse:
> Could you edit the affected weapons for ferals as well, while you're at it? I know this is going to affect the combat knife special attack, not sure if the other knife is going to be affected or not.

Feral soldier doesn't have any bash damage on him?
```
"melee_skill": 6,
    "melee_dice": 2,
    "melee_dice_sides": 2,
    "melee_damage": [ { "damage_type": "cut", "amount": 6 } ],
```

Issue #68024, opened 2023-09-06
Enervating Touch spell (from Mind Over Matter) is crashing the game when cast


rockygr:
### Describe the bug

When using Enervating Touch on an NPC, the game will crash

### Attach save file

[gooda-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12539443/gooda-trimmed.tar.gz)

### Steps to reproduce

Open spellcasting menu;
Activate Enervating Touch;
Target something (a Speaker is near, helpfully);
The game will crash.

### Expected behavior

The game would not crash when casting a spell.

### Versions and configuration

- OS: Linux
    - OS Version: Distributor ID: Linuxmint; Description: Linux Mint 21.2; Release: 21.2; Codename: victoria; 
- Game Version: 35b4e10 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    Slowdown Fungal Growth [no_fungal_growth],
    Stats Through Kills [stats_through_kills],
    Aftershock [aftershock],
    Mind Over Matter [mindovermatter],
    Xedra Evolved [xedra_evolved],
    Bombastic Perks [bombastic_perks],
    Translate Complex Dialogue [translate_dialogue]
]

### Additional context

[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/12539439/debug.log)

Maleclypse:
@Standing-Storm

Standing-Storm:
Just loaded up the game and I can't reproduce: 
![image](https://github.com/CleverRaven/Cataclysm-DDA/assets/120433252/1475f7ec-4134-4a6a-9a58-7879ff2bc3eb)
![image](https://github.com/CleverRaven/Cataclysm-DDA/assets/120433252/4a54d861-c4bd-4e84-a71a-27e6607c0986)

shadowsapex:
@Standing-Storm i can confirm this bug. after applying the effect, pass a turn. the game will crash. my save file is attached, it only requires passing a turn to see the bug

[Escobares-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12615603/Escobares-trimmed.tar.gz)

i tried it on the starting npc too but it seemed to work fine

Standing-Storm:
[...]

PR #68148, opened 2023-09-15
[MoM] Change Enervating Touch damage type

Standing-Storm:
#### Summary
Bugfixes "Change Enervating Touch damage type"

#### Purpose of change

Re: #68024, people are getting crashes when using the Vitakinetic power Enervating Touch and changing the damage type fixes the crashes if not the underlying problem.

#### Describe the solution

Change damage type to pure.

#### Testing

Loaded up a game, Enervating Touched a zombie and ran from it for 15 seconds, no crash.


Maleclypse:
@dseguin is this a jsonized damage type problem?

PR #29588, opened 2019-04-14
Expanded ledge examine action with jump over command

ZhilkinSerg:
#### Summary

```SUMMARY: Features "Expanded ledge examine action with jump over command"```

#### Purpose of change

Allow players to jump over single tile ledges.

Partially addresses #28564.

#### Describe the solution

Added menu to examine action on ledge terrains which allow whether you want to jump over or climb down the ledge.

image

Player cannot jump if there is something in the way on jump destination (terrain/monster/furniture).

image

Player cannot jump if destination is ledge too (thus player can only jump over a single tile ledge).

image

#### Additional context

Jumping over the ledges won't work for automove and NPC.

PR #68150, opened 2023-09-15
Multiple-choice prompt when walking into ledges.

EvanBalster:
#### Summary
Interface "Walking into ledges examines them."

#### Purpose of change

Currently the game treats ledges as traps and issues a generic prompt "Really step into ledge?" when the player attempts to walk off one.  This "Wile E. Coyote maneuver" is usually not the safest or most desirable interaction.

I was inspired to make this change while working on a more complex PR https://github.com/CleverRaven/Cataclysm-DDA/pull/68120 which makes climbing down a more approachable game mechanic.  This PR and that PR are independent (UX vs balance) but work nicely together.

#### Describe the solution

- Walking into a ledge now invokes `iexamine::ledge` as if the player used the `e`xamine key.
- The options list is ordered as follows:
  - `[p] -- Peek down.`
  - `[C] -- Crawl down.`  (only listed if player has the WALL_CLING flag).
  - `[c] -- Climb down.`
  - `[j] -- Jump across.`  (disabled with explanatory note if there isn't a one-tile gap).
  - `[f] -- Fall down.`
- When choosing to fall down, the player is prompted `Climbing might be safer. Really fall from the ledge?`

In my first revision I did not include an option to fall.    On further consideration I believe this option is still useful when the player wants to move quickly, dive into water or intentionally self-harm.

Currently, "crawl down" is an special case in the `iexamine::ledge` menu but it may be cleaner to integrate it into the `climb_down` function in the future.

#### Testing

- Tested walking into a ledge.
- Tested climbing down via this interaction: by hand, with grappling hook, and with ladder.
- Tested walking into a ledge / climbing off ledge with a workbench grabbed.
- Tested together with changes from https://github.com/CleverRaven/Cataclysm-DDA/pull/68120

#### Additional context

image


harakka:
This is a welcome addition. I think peek down should be the default selected though, because it is the safest, and I don't think there is a "the most obvious use case" for walking into a ledge.

EvanBalster:
Adjusted the menu after some playtesting in my own campaign.

Implemented @harakka's suggestion and put `[p] Peek down` first.  Added `[f] Fall down` as the last option, restoring the option to walk off a ledge (which is faster, and may be sensible when there's water below).

Code formatting:  Fixed spacing issues.  Modified `iexamine::ledge` to use an enum rather than integers for menu options (readability).  Restored the original ordering of the switch/case to improve diff readability.

EvanBalster:
Clang 12 seems to be unhappy with this fallthrough code:

```
            /* Otherwise, intentional fallthrough (no pun intended). */
        }
        default:
            popup( _( "You decided to step back from the ledge." ) );
            break;
```

Peculiar thing is, I see these all over the codebase annotated with `/* fallthrough */` or `/* intentional fallthrough */`.  I'll try changing the comment.  Sounds like these are "special" comments commonly used with Clang.

EvanBalster:
Greyed out the menu item for jumping when there isn't a single-tile gap to jump across.  This is mostly intended to draw attention to the conditions for ledge hopping.

I notice the language for jumping between ledges is pretty weird.  The gap is referred to as "an obstacle" but I think this is the only two-tile horizontal jump in the game.  My reckoning as to why we can only perform two-tile moves in this specific situation is that the player is jumping, catching on the opposite ledge and climbing up.  Notably this action is 100% safe.  I made the action a bit more strenuous than it was before.

image

I notice this change was suggested by @kevingranade some four years ago when reviewing the PR that added jumping across ledges:  https://github.com/CleverRaven/Cataclysm-DDA/pull/29588#issuecomment-483504219

EvanBalster:
I've been playtesting with this change for a few days and it seems to be working well.  Polishing it up and making supplications to the review gods.

image

Asterix199:
possible error in the line:
expectation: Looks like **1%d** story.  Nothing your webs can't handle.  Descend?
in reality:      Looks like **% d** story.  Nothing your webs can't handle.  Descend?

EvanBalster:
@Asterix199 Looks like I introduced that problem in #68120.  If I recall correctly, the message in question will never appear because the singular version is skipped over in favor of a simpler message.

I'm totally reworking the `climb_down` function and prompts in #68190 and in the process that malformed message is removed completely.

PR #68155, opened 2023-09-15
Do not update visibility cache every two minutes

### LINKED ISSUE #68154: 15x (!) increase in test run time
### Describe the bug

After https://github.com/CleverRaven/Cataclysm-DDA/pull/67997, running the full suite of tests takes 3 hours. If that is reverted, it takes 12 minutes.

### Attach save file

N/A

### Steps to reproduce

1. Build tests
2. Run tests
3. Wait 3 hours for tests to complete
4. Revert f12487cefcdd8c18f70ce68f73e09826bdced18d
5. Build tests
6. Run tests
7. Wait ~12 minutes for tests to complete

### Expected behavior

Tests take much less than 3 hours to run.

### Screenshots

`perf` profile:
image

Commit d6d150c655ea73070f68f7a8855173cc2f565092

### Additional context

_No response_

ZhilkinSerg:
#### Summary
None

#### Purpose of change

Fixes #68154

#### Describe the solution

Do not update visibility cache every two minutes

### DIFF
--- a/src/player_activity.cpp
+++ b/src/player_activity.cpp
@@ -258,10 +258,6 @@ void player_activity::do_turn( Character &you )
                 }
             }
         }
-        // Also update map and visibility caches every 2 minutes
-        map &here = get_map();
-        here.build_map_cache( you.pos().z );
-        here.update_visibility_cache( you.pos().z );
     }
     const float activity_mult = you.exertion_adjusted_move_multiplier( exertion_level() );
     if( type->based_on() == based_on_type::TIME ) {


Rewryte:
Thanks for identifying the lines that was causing the test slowdown.  I suggest adding a check for test mode before updating map and visibility caches to allow the player to perceive lighting changes while performing an activity.

ehughsbaird:
I think the better fix for that would be to update the visibility cache on lighting changes, this doesn't just slow down the game in `test_mode`.

ZhilkinSerg:
> Thanks for identifying the lines that was causing the test slowdown. I suggest adding a check for test mode before updating map and visibility caches to allow the player to perceive lighting changes while performing an activity.

Test mode should not disable any game logic. It can only suppress things like draw calls or UI interactions, but not visibility cache updates.

ZhilkinSerg:
> I think the better fix for that would be to update the visibility cache on lighting changes, this doesn't just slow down the game in `test_mode`.

Yes, cache update needs to be revisited in a separate PR. Right now I just to remove an unnecessary slowdown.

PR #68159, opened 2023-09-15
[MoM] Add zombie blank

Standing-Storm:
#### Summary
Mods " Add zombie blank to Mind Over Matter"

#### Purpose of change

Thanks to the discovery of the NO_SPELLCASTING flag, I finally had a good idea for a zombie evolution--the anti-psi zombie.

#### Describe the solution

Add the "zombie blank," a zombie that emits a field that inhibits psionic activity by disrupting the local connection with the Nether. Add two upgrades, the zombie null and the nether-void, who are faster and have a melee attack that can scramble the target's psionic abilities, removing all psionic effects they have on them. These zombies are also immune to basically all psionics (except telekinetic push/pull, that would take C++).

Other than their anti-psi powers and being fast, they don't have any particular special abilities that work on the player. The null and the nether-void can banish Nether beings (re: Nether connection disruption), so luring shoggoths to them is a viable strategy.  The nether-void has very long arms and can grab and drag you, but their base damage is low and dodge is average.

I know I have a tendency to, in reaction to the ubiquity of Use Gun advice, put in monsters that are immune or heavily resistant to bullets, but in this case "Ahahahaha just shoot them bro" is the reasonable counterplay.

#### Testing

image
image
image

#### Additional context

My original idea instead of them getting faster was to have them get stronger and end with a "Nullk." Fortunately, I came to my senses.


MNG-cataclysm:
If it were restricted to some special location or extremely far down the evolutionary tree, with sufficient rarity, I think the "Nullk" would be a cool enemy to fight.

MNG-cataclysm (approved):

MNG-cataclysm:
I think it all looks good.

PR #68160, opened 2023-09-15
Isolation protocol Mod

John-Candlebury:
#### Summary
Mods "Create the Isolation Protocol Mod: A traditional roguelike experience"

#### Purpose of change
Isolation protocol is a total conversion mod that adapts vanilla mechanics to fit a traditional Dungeon Crawler game loop. Themed around the main game's lore, you play as bystander that stumbles into and must fight through a locked down lab.

At the moment the  goal is simply to survive as many microlab levels before resource exhaustion does you in, following updates will implement both bossfights, clearing objectives for each lab level, and more varied dungeon mapgen.

#### Describe the solution

For now you can:

- Pick one of four classes to play. (Fully distinct gameplay to come for each one of them)
- Fight and survive within an infinite fractal lab.
- Race against a danger clock that severely punishes you if you stay more than 20h in a single level.

Waiting so the mod launches with more content.

#### Testing

Reached lab level 8.

### DIFF
--- a/data/json/mapgen/microlab/microlab_hallway.json
+++ b/data/json/mapgen/microlab/microlab_hallway.json
@@ -6,7 +6,11 @@
     "object": {
       "fill_ter": "t_strconc_floor",
       "place_nested": [
-        { "chunks": [ "microlab_generic_hallway_straight_tile" ], "x": 0, "y": 0 },
+        {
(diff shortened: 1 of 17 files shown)


ashGlaw:
Cool idea! Looking forward to it

PR #68164, opened 2023-09-16
Zone naming clean-up

smileynet:
#### Summary
None

#### Purpose of change

Format zone names consistently and order them alphabetically 

#### Describe the solution

This change ensures a consistent style and will help avoid confusion by anyone reading or modifying zone related code.

Continue to try to remember which types of things clang-tidy will complain about not being ordered alphabetically or not.

#### Testing

Create zones affected by change, ensure functionality remains consistent

### DIFF
--- a/data/json/furniture_and_terrain/furniture-appliances.json
+++ b/data/json/furniture_and_terrain/furniture-appliances.json
@@ -1,4 +1,3 @@
-
 [
   {
     "type": "furniture",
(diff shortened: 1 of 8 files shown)


Karol1223:
I may be entirely wrong, but since you're changing IDs doesn't that mean zones on existing saves will fall apart? That doesn't seem like a very good idea

smileynet:
> I may be entirely wrong, but since you're changing IDs doesn't that mean zones on existing saves will fall apart? That doesn't seem like a very good idea

There is already logic to remove unrecognized zones that wouldn't prevent saves from loading. But in the spirit of not disrupting the zone setup in a legacy saves, I have added handling to migrate the types.

PR #68169, opened 2023-09-16
Hold onto items which donot fit vehicle storage space

### LINKED ISSUE #68161: Dropping items in a filled vehicle container should NOT place remaining items on ground

I'm sure every player has experienced the annoyance of dropping items in a vehicle container to the point the container was filled, resulting in remaining items being placed on the ground beneath the vehicle.  Not only is this unrealistic and a waste of time, the notification that it has happened can easily be missed resulting in items being left behind or damaged when the vehicle moves away.

### Solution you would like.

Ideally, the drop process should simply be halted when the vehicle container is full so that any remaining items are kept in the player inventory.

Alternatively, I guess the player could be asked if they want to continue dropping items after the vehicle container is filled, if we want to keep the ability to teleport items through the floor of the vehicle for some reason.

### Additional context

I'm finally bringing up this issue after yet again leaving my acetylene torch behind because I didn't see the message that the vehicle container was filled.

Leonkoithara:
#### Summary
Bugfixes "Don't teleport items to the ground if the vehicle storage destination is full"

#### Purpose of change
Currently if items are stored inside vehicle containers via the Drop/Multidrop option and the destination is insufficient to store the items they are teleported to the ground through the vehicle floor.
Fixes #68161 

#### Describe the solution
This PR changes it so that the items are not dropped from the inventory at all. Currently since the vehicle floor itself is not a valid storage space if the item is unable to be stored in the inventory(too large/heavy) as well they are put on the ground as previously.

#### Testing
Testing done:
1. Dropped items beyond storage capability of destination
2. Tested using AIM(no change in behaviour the item cannot be moved)
3. Crafting items while no inventory space/item too heavy to hold(Item is moved into the storage space)

### DIFF
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -244,7 +244,7 @@ static void put_into_vehicle( Character &c, item_drop_reason reason, const std::
             // Retain item in inventory if overflow not too large/heavy or wield if possible otherwise drop on the ground
             if( c.can_pickVolume( it ) && c.can_pickWeight( it, !get_option<bool>( "DANGEROUS_PICKUPS" ) ) ) {
                 c.i_add( it );
-            } else if ( c.has_wield_conflicts( it ) && c.can_wield( it ).success() ) {
+            } else if ( !c.has_wield_conflicts( it ) && c.can_wield( it ).success() ) {
                 c.wield( it );
             } else {
                 here.add_item_or_charges( where, it );
--- a/src/activity_item_handling.cpp
+++ b/src/activity_item_handling.cpp
@@ -213,6 +213,7 @@ static void put_into_vehicle( Character &c, item_drop_reason reason, const std::
     c.invalidate_weight_carried_cache();
     vehicle_part &vp = vpr.part();
     vehicle &veh = vpr.vehicle();
+    const tripoint where = veh.global_part_pos3( vp );
     map &here = get_map();
     int items_did_not_fit_count = 0;
     int into_vehicle_count = 0;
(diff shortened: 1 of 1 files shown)


src/activity_item_handling.cpp
```diff
@@ -238,28 +238,40 @@ static void put_into_vehicle( Character &c, item_drop_reason reason, const std::
                 it.mod_charges( -charges_added );
                 into_vehicle_count += charges_added;
             }
-            here.add_item_or_charges( where, it );
-            fallen_count += it.count();
+            items_did_not_fit_count += it.count();
+            add_msg( _( "Unable to fit %s in the %2$s's %3$s" ), it.tname(), veh.name, part_name );
```
NetSysFire, line 242:
```suggestion
            add_msg( _( "Unable to fit %s in the %2$s's %3$s." ), it.tname(), veh.name, part_name );
```

Terminal punctuation for consistency.

src/activity_item_handling.cpp
```diff
@@ -238,28 +238,40 @@ static void put_into_vehicle( Character &c, item_drop_reason reason, const std::
                 it.mod_charges( -charges_added );
                 into_vehicle_count += charges_added;
             }
-            here.add_item_or_charges( where, it );
-            fallen_count += it.count();
+            items_did_not_fit_count += it.count();
+            add_msg( _( "Unable to fit %s in the %2$s's %3$s." ), it.tname(), veh.name, part_name );
+            // if cannot retain in inventory due to being too large/heavy drop onto ground via the drop_on_map after returning
```
Kamayana, line 243: Does the mentioned `drop_on_map` actually get called as this implies? The only call to `put_into_vehicle` is followed immediately by `return`, which I imagine means it never reaches the `drop_on_map`.

Leonkoithara: Hi yes you are right I had thought the drop_on_map was the function that forced the overflow items to the ground. I've fixed my PR to remedy this.

PR #68171, opened 2023-09-17
Insert new response to more suitable place for the exist topic

EIIKaO:
#### Summary
Interface "talk_topic option for insert new responses at the suitable place to existing topic"

#### Purpose of change
For modding. If we want to add a new response to the existent talk topic for a conversation with a follower.
```
  {
    "id": "TALK_ALLY_SOCIAL",
    "type": "talk_topic",
    "responses": [
      { "text": "What do you think of our current situation?", "topic": "TALK_ASK_CURRENT_SITUATION" }
    ]
  },
```
The response(s) I added is placed at the end.
20230917a

I would prefer that it be placed above [d] (i.e., above `TALK_NONE` [d] and `TAKL_DONE` [e] ). This is because I think it is better that the response to return to the topic(`TALK_NONE`) and the response to end the conversation(`TALK_DONE`) basically be placed at the end of responses, for playability.

#### Describe the solution
I have set up new field named `insert_before_standard_exits` on the `talk_topic` and to create a process that will place the response above `TALK_NONE` and `TALK_DONE` (if they exist and are placed at bottom in responses) when this is turned on.

```
  {
    "id": "TALK_ALLY_SOCIAL",
    "type": "talk_topic",
    "insert_before_standard_exits": true,
    "responses": [
      { "text": "What do you think of our current situation?", "topic": "TALK_ASK_CURRENT_SITUATION" }
    ]
  },
```
Then, 
20230917b

I tried to see if I could get the desired behavior with the existing features, but as far as I could research, it was not possible, so I added a new function.
**If such a thing is already possible with the existing features, please let me know.** I will gladly withdraw it.

I don't think this is an important feature at all for the common player.
But I think it might be helpful if the creator of a mod that extends an existing system has a nervous disposition that will not tolerate `TALK_NONE` and `TALK_DONE` not being placed at the end of responses, like me.

#### Testing
In-game testing was completed and no problems.
It will not work unless explicitly turned on in the field, so it is not expected to affect existing play in any way.

#### Additional context
I have a bit free time these days, so made this as part of the maintenance of my virtual machine for C++ program development.

Yes, I think there are still some terrible codes, so please correct it if you find appropriate ones.
~~`if ( responses.size() >= 1 && responses[ responses.size() - 1]}`~~
`if ( responses.size() >= 2 && responses[ responses.size() - 2]}`

~~Also, the field name `"insert_above_bottom"` is not good. Alternative are welcome.~~

### DIFF
--- a/data/mods/Xedra_Evolved/npc/draco_dune.json
+++ b/data/mods/Xedra_Evolved/npc/draco_dune.json
@@ -2,6 +2,7 @@
   {
     "id": "TALK_REFUGEE_Draco_1a",
     "type": "talk_topic",
+    "insert_before_standard_exits": true,
     "responses": [ { "text": "Have you seen anything that could help me?", "topic": "TALK_REFUGEE_Draco_changeling_breadcrumb" } ]
   },
   {
--- a/data/mods/Xedra_Evolved/npc/mr_lapin.json
+++ b/data/mods/Xedra_Evolved/npc/mr_lapin.json
@@ -2,6 +2,7 @@
   {
     "id": "TALK_WARRENER",
     "type": "talk_topic",
+    "insert_before_standard_exits": true,
     "responses": [ { "text": "What can you tell me about Xedra?", "topic": "TALK_WARRENER_XEDRA" } ]
   },
   {
--- a/data/mods/Xedra_Evolved/npc/rubik.json
+++ b/data/mods/Xedra_Evolved/npc/rubik.json
@@ -2,6 +2,7 @@
   {
     "id": "TALK_EXODII_MERCHANT_Talk",
     "type": "talk_topic",
+    "insert_before_standard_exits": true,
     "responses": [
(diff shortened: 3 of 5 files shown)


PatrikLundell:
"insert_as_last_dynamic", "insert_before_standard_exits" or something similar?

EIIKaO:
> "insert_as_last_dynamic", "insert_before_standard_exits" or something similar?

Thanks for alternatives.
In my view, as a non-native speaker, the latter "insert_before_standard_exits" seems more direct and easier to understand. I would like to adopt this one.
Thanks again!

Maleclypse:
I've added an example use and documentation of this feature.  Thanks!

EIIKaO:
> I've added an example use and documentation of this feature.

Thank you for your work! It was very helpful.

I-am-Erk:
This is a great feature, and I'd love a similar one inside the `responses` list for an individual response, something like `push_to_top` or `push_to_bottom`. It'd be nice to be able to have a specific response always show up first in the list.

Not enough to say it in an issue, just putting my thoughts in writing.

PR #68173, opened 2023-09-17
Add object creator to releases

snipercup:
#### Summary
Build "add object creator to releases"

#### Purpose of change
Allow a greater amount of people to make use of the object creator by simply downloading it and running it in the same folder as cataclysm-tiles

#### Describe the solution
Add build commands to the release.yml workflow in github actions.

- Both the Linux and the Windows versions are built using the devcontainer. The devcontainer will be uploaded to the github image repository (only when building the windows version since it only needs to be pushed once) so it can be cached to save build time. The first time it will run for about 3 hours and after that it takes about 40 minutes to compile the object creator.

- The Linux and Windows versions are configured to be built in parallel to the game builds so if one of them fails, the game will still be built succesfully.

- Both versions are added to a compressed archive and they are shipped with a copy of the QT license. 

- In order for the devcontainer image to succesfully be pushed to ghcr.io, the organisations workflow permissions need to be set to 'read and write permissions' via `GitHub profile icon -> Your organizations -> Settings -> Actions -> General -> Enable Read and write permissions` but for the org instead. The repository also needs read and write permissions. As per [this issue](https://github.com/actions/runner/issues/1039).

Add a menubar to the object creator:
image

Added the about object creator dialog:
image

Added the about qt dialog
image

- Cross-compile the object creator without devcontainer like in the code that's commented out [here](https://github.com/snipercup/Cataclysm-DDA-1/blob/Object-Creator-Add-To-Releases/.github/workflows/release.yml). The problem with this is that it can't find some libraries as shown in [this run](https://github.com/snipercup/Cataclysm-DDA-1/actions/runs/6154839771/job/16700823374).

- Make the linux version of the object creator without the devcontainer. It is possible compile it inside the release.yml directly, but when you try to open the binary after downloading it, you run into a `undefined symbol: _ZdlPvm, version QT_5` error. This is possibly related to compiling with a different C++ compiler or github actions using different versions of the QT5 framework

#### Testing
You can see all of the actions I added in this PR be succesfully performed in [this run](https://github.com/snipercup/Cataclysm-DDA-1/actions/runs/6251591879) (with most of the other actions removed from the release.yml workflow)

You can see that the steps I added work in-tandem with the rest of the actions in [this run](https://github.com/snipercup/Cataclysm-DDA-1/actions/runs/6252201538/job/16974969962). Because I can't trigger github pages rebuild, they all fail but that has nothing to do with my additions.

You can download a working copy of the object creator from my releases [here](https://github.com/snipercup/Cataclysm-DDA-1/releases/tag/cdda-experimental-2023-09-20-1743).


andrei8l:
The mingw build is statically linked against Qt5 so we need to provide a copy of its LGPL license.

Also, please add the object creator sources to the astyle check ([here](https://github.com/CleverRaven/Cataclysm-DDA/blob/bb0bdee77f71079b8a86f2312be483609f4186fd/Makefile#L870-L879)).

snipercup:
> The mingw build is statically linked against Qt5 so we need to provide a copy of its LGPL license.

Of course! I'm looking for the best way to implement this. Should it just be a text file inside a zip file that contains the license and the object creator? Should I put the license text in the object creator somewhere? Maybe I can put a hyperlink in a help->about menu item. What do you recommend?

andrei8l:
AFAICT you need to do both, so:
1. License file distributed with the binary. This can be in `License-Qt5.txt`, similar to the `LICENSE-OFL-Terminus-Font.txt` we ship with the game.
2. Qt5 mentioned in the program somewhere. [`QApplication::aboutQt()`](https://doc.qt.io/qt-5/qapplication.html#aboutQt) can probably help.

and also:
3. A link to this repo

Makefile
```diff
@@ -857,6 +857,8 @@ else
 endif
 THIRD_PARTY_SOURCES := $(wildcard $(SRC_DIR)/third-party/flatbuffers/*.cpp)
 HEADERS := $(wildcard $(SRC_DIR)/*.h)
+OBJECT_CREATOR_SOURCES := $(wildcard $(SRC_DIR)/object_creator/*.cpp)
+OBJECT_CREATOR_HEADERS := $(wildcard $(SRC_DIR)/object_creator/*.h)
```
andrei8l, line 861: `SRC_DIR` is src
```suggestion
OBJECT_CREATOR_SOURCES := $(wildcard $object_creator/*.cpp)
OBJECT_CREATOR_HEADERS := $(wildcard $object_creator/*.h)
```
or `$(SRC_DIR)/../object_creator/`

snipercup:
I added the QT license to the object creator directory since it only applies to the object creator. A copy is shipped with a binary release of the object creator. 

I had to build the linux version in the devcontainer as well because it would throw an error when comiled using github actions directly. Now the compilation is completely separate from the game compilation. It might as well run inside a separate job outside the matrix. However, it's still a release so I put it in the release workflow. 

My actions are done and I marked this PR ready for review.

PR #68176, opened 2023-09-17
Accessible item insert menu

lispcoc:
#### Summary
Interface "Accessible item insert menu"

#### Purpose of change
In today's systems that have to manage countless items, it is very important that the feature to explicitly store items in specific containers.
However, the current item insertion menu can only be accessed from the inventory menu, which is very inconvenient.

#### Describe the solution
Adds a shortcut key to instantly call the item insertion menu.

#### Testing
Game begin, press "v" key, then the item insertion menu opens. I verified that I can indeed insert items.

#### Additional context
Keys that can be used for shortcuts are limited. I assigned the extra "v" key, but is it better to use a different key?

image

### DIFF
--- a/data/raw/keybindings.json
+++ b/data/raw/keybindings.json
@@ -2482,6 +2482,13 @@
   },
   {
     "type": "keybinding",
+    "name": "Insert item",
+    "category": "DEFAULTMODE",
+    "id": "insert",
+    "bindings": [ { "input_method": "keyboard_any", "key": "v" } ]
+  },
+  {
+    "type": "keybinding",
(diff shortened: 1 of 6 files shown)


oosyrag:
Does this work for inserting items in adjacent tiles? That's one of the last things we're missing :o

lispcoc:
I'm getting a clang error in a location unrelated to this PR, should I fix it?

> Does this work for inserting items in adjacent tiles? That's one of the last th
Yes; We can insert items to containers in adjacent tiles or insert items in adjacent tiles to containers.

PR #68190, opened 2023-09-18
JSONify, rebalance & expand climbing with ladders, grapples, vehicles, downspouts, mutations & more

EvanBalster:
#### Summary
Balance "Rebalance & expand climbing with ladders, grapples, vehicles, downspouts & mutations."

#### Purpose of change

Currently some methods of climbing are quite risky while others are absolutely safe and the nature of these differences can be counterintuitive.  Ladders and grappling hooks are foolproof while a lot of other climbable things do nothing to help you descend.  At least one mutation is described as useful for climbing but almost never comes into play.  This can be chalked up to an untamed labyrinth of conditionals and special cases, where functions like `slip_down` don't have enough context to factor in abilities and terrain.

This work builds on two earlier contributions:

* https://github.com/CleverRaven/Cataclysm-DDA/pull/68120 (ie, get a sense of falling risk)
* https://github.com/CleverRaven/Cataclysm-DDA/pull/68150 (ie, climb by walking into a ledge)

#### Describe the solution

- Objects like downspouts and fences are now a bit safer to climb down than ledges.
- Grappling hooks and ladders are no longer foolproof, but they're safer than most other options.
- Parked vehicles, sandbag walls, already-placed grapples and other climbable things now help with climbing down.
- Implemented WALL_CLING, VINES2 and VINES3 mutations as safe ways to climb down.  Detaching a vine may hurt but creates climbable foliage.
- When climbing down with a ladder in inventory, the player has the option to place it, similar to a grappling hook, but only on one-story drops.

My changes create a JSON-based framework for reworking and developing vertical movement mechanics in general, based on two new properties:  `climb_maneuver`, a simple enum class which can be `up`, `down` or `over_obstacle`; and `climbing_aid` — a new JSON type which refers to the method of climbing (various terrain, furniture, vehicle, tools & mutations).  Climbing aids offset the risk of falling during a climb.

Certain climbing methods deploy furniture that can be used to climb back up.  When walking into a ledge the player is first prompted with the option to use one of these methods.

| Deployable Affordance                                                | Previous Effect                           | New Effect                                             |
| --------------------------------------------------------- | ----------------------------------------- | ------------------------------------------------------ |
| † Grappling hook in inventory                             | Never slip.                               | **–12% slip chance** (down)                            |
| † Stepladder in inventory                             | n/a                               | **–20% slip chance** (down)                            |
| † Detach a vine player ability                            | none                                      | **–100% slip chance** (down)                           |
| † Webs player ability                                     | Never slip.<br />Descends to ground.      | **–100% slip chance**<br />Descends to ground          |

Otherwise, the player will use the safest climbing affordance that does not modify the environment.  These are, from riskiest to safest:

| Affordance                                                | Previous Effect                           | New Effect                                             |
| --------------------------------------------------------- | ----------------------------------------- | ------------------------------------------------------ |
| Ledge (default)                                           | none                                      | none                                                   |
| Corners, alleyways & ramps<br />(`climb_difficulty` math) | Aids climbing up                          | Aids climbing up,<br />but slightly more               |
| downspouts, fences<br />(CLIMBABLE flag)                  | Permits climbing up                       | **–5% slip chance** (down)<br />Permits climbing up    |
| Any vehicle below                                         | No effect (down)<br />Aids climbing up    | **–8% slip chance** (down)<br />Aids climbing (up)     |
| Rope/grappling hook below                                 | No effect (down)<br />Permits climbing up | **–12% slip chance** (down)<br />Permits climbing (up) |
| Ladder deployed below<br />(LADDER flag)                  | Never slip. (PR 68120)                    | **–20% slip chance**                                   |
| sandbag/earthbag walls<br />(CLIMB_SIMPLE flag)           | No effect (down)<br />Permits climbing up | **–40% slip chance** (down)<br />Permits climbing up   |
| Wall Cling player ability.                                | Never slip.                               | **–100% slip chance** (down)                           |
| Vines player ability.                                     | none                                      | **–100% slip chance** (down)                           |

Bear in mind a typical "starting" slip chance in the early game is 8-12%; this can increase greatly with encumbrance, pain, wetness or especially low stamina.  For players in good climbing condition, the new modifier can reduce slip_chance below 0%.  If it reaches -5% or less, the climbing prompt says "perfectly safe" instead of safe and hides the damage estimate.

All the data is in [this JSON file](https://github.com/EvanBalster/Cataclysm-DDA/blob/eb-climb-slip-mechanics/data/json/climbing.json) now.

While refactoring I made some related changes and fixes:

* Deployed grappling hooks are now called "grappling hook" instead of "rope leading up".
* Grappling hooks are now placed at the ledge rather than on the ground if the player chooses to grapple down one story and fall for some remaining distance.
  * At the moment, due to a bug in other code, the grappling hook comes tumbling down a few seconds later and breaks apart.
* Grappling hooks, webs and vines now make the prompt display "You should be easily able to climb back up." where appropriate.
* I made changes to the way multi-story descent and rope/web placement work.
  * The player moves one story at a time (but there is no pause currently).
  * The chance to slip is now rolled once per level while descending (currently only applies to webs).
  * I generalized the code for placing furniture (web or grappling rope) during descent.  This might be used to allow for multi-story rapelling later.

There are all sorts of ways we could balance the slip chance modifiers, but it's probably best to discuss that as a followup pull request.

There are many possibilities for new features, most of which are out of scope for this PR.

* Integrate with upward climbing, which is currently hardcoded in `vertical_move`.
  * Also noteworthy is `climb_difficulty` which allows climbing up (but not down) in alleyways and cramped spaces.
* Ropes are described as usable for climbing, but that doesn't seem to be implemented.  Maybe they should work like grappling hooks if there's something to tie to (like a railing).  A rope should only be possible to take down from the top.
* Grappling hooks are described as 30 feet long — maybe they should descend two stories.  I laid some groundwork for this, but the furniture created by a 30-foot item would need special consideration to avoid item duplication.
* Multi-story grappling/rappelling could be more interactive.  eg, the player could stop at the end of the rope instead of being forced to let go and fall.  This could create interesting tactics like jumping through a second-story window from the roof!

#### Testing

* Tested various ledges on a 2-story building.
* Tested with both types of stepladders including placing stepladders.
* Tested with downspouts.
* Tested with grappling hook, pre-placed and in inventory.
* Tested with varying levels of stamina, torso encumbrance & weight, and two-handed item wielding.
* Limited testing with vines and webs mutations.

**I am requesting help testing this with mutations**, as I haven't played enough CDDA to be familiar with them.  The previous implementation seemed buggy but I did my best to replicate the way it worked while greatly tidying up the code.

#### Additional context

image

image


EvanBalster:
Quirks and mysteries with this first revision:

I implemented the vines mutation for ledge climbing based on some code in `find_or_make_stairs` where a mutated player gets the option to tear off one of their vines (perhaps painfully) and leave it behind so they can climb back up.  Cool stuff, right?  But I'm not sure how it was supposed to work.

In `find_or_make_stairs`, the mutation sets an output bool `rope_ladder` which seems like it's intended to produce some kind of climbable object.  I couldn't find any evidence that the `rope_ladder` variable actually works as intended in any of the calling functions.  Consequently I have no idea what kind of rope it was supposed to make.  So, in my implementation of `climb_down`, ripping off a vine leaves the same type of web as the web rappel mutation.

One other little quirk of my implementation is that there is no JSON flag for climbable ropes.  So any object which has the DIFFICULT_Z flag but not CLIMB_SIMPLE or LADDER is provisionally considered a "rope" and is markedly better than other CLIMBABLE things.  In the base game, this category includes spun webs, "rope leading up" (deployed grappling hook) and a *terrain type* which is also called "rope leading up".  Perhaps this terrain type is the intended result of the `rope_ladder` flag?

Nebnis:
Perhaps out of scope for this PR, but would be nice if some other mutations that made sense could also climb(or jump in this case) with their respective mutations like frogs and bunny and maybe feline pounce.

EvanBalster:
Now that #68150 merged, bumping into a ledge will present you with a choice to peek down, climb down, jump across or fall down.  If you have the "wall cling" ability that's an additional option.  "Climb down" can bring you to the menu added by this PR. 

So you might see this:
```
There is a ledge here.  What do you want to do?
-----------------
p | Peek down.
c | Climb down.
j | Can't jump across (need a small gap).
f | Fall down.
```
And then this (but only if you have grapple, vines OR webs, this is what it would look like with all three):
```
How would you like to climb down?
-----------------
w | Spin your webs to descend.
g | Attach your grappling hook to climb down here.
d | Detach a vine so you can climb back up.
c | Climb down using your vines.
```

I might try to merge these two menus.  Additionally, because this PR integrates wall cling into `climb_down`, the special "crawl down" option can probably be removed from the `iexamine::ledge` menu.  Instead it just becomes the default method for climbing on the environment.

> @Nebnis "Perhaps out of scope for this PR, but would be nice if some other mutations that made sense could also climb(or jump in this case) with their respective mutations like frogs and bunny and maybe feline pounce."

Jumping is probably a topic for a different PR.  #68150 made a couple changes that make leaping from ledges into a more intuitive and visible game mechanic, though.  It wouldn't be too hard to modify the code in `iexamine::ledge` to test a range of possible jumping distances based on player mutations/abilities, which would allow for spectacular leaps.

EvanBalster:
I managed to work out a way to merge the climb down menu into the ledge menu.  Now, walking into a ledge presents you with a combined menu like this:

image

If you choose a climbing option (`c`limb, `v`ine, `w`ebs or `g`rapple) you will then see the risk prompt before taking the plunge.  So three prompts have been pared down to two.  It took some additional functions in `game` though.

EvanBalster:
Made it so the climb confirmation prompt is skipped when all three of the following are true:

* It is perfectly safe to climb down (slip chance <= –5%).
* The player will climb all the way to the ground (ie, descending one story or using webs)
* The player can easily climb back up (eg, ladder, sandbags or newly placed webs/grapple)

With that change I'm moving this from draft to active PR status.  But these changes are somewhat extensive and need code review.  In particular, I'm seeking guidance about two things:

* Could we avoid creating the new enum in the `game` class header?  It might grow in the future.
* How could this be done in a more data-driven way, allowing climbing affordances to be balanced/added in JSON?

EvanBalster:
Completed a first pass at the arduous process of converting all the functionality in `climb_down` to use JSON-based rules.  "Affordances" are now referred to as "Climbing Aids", with their own JSON data type, JSON file, C++ header and source file.  Lots of logic has been moved into those locations.  Further testing and cleanup will be necessary.

src/game.cpp
```diff
@@ -13113,33 +13127,159 @@ int game::slip_down_chance( bool show_messages )
...
+}
+static bool climb_affordance_menu_decode( int retval, climbing_aid_id &aid_id )
+{
+    int_id< climbing_aid > as_int_id( retval - 0x1000 );
+    if( as_int_id.is_valid() ) {
+        aid_id = as_int_id.id();
+        return true;
+    }
+    return false;
+}
```
ehughsbaird, line 13176: A comment explaining the purpose of these functions would be helpful, they're a little opaque at first glance.

src/game.cpp
```diff
@@ -13113,33 +13127,159 @@ int game::slip_down_chance( bool show_messages )
...
+    } else {
+        cmenu.query();
+        climb_down_menu_pick( examp, cmenu.ret );
+    }
+}
+
+void game::climb_down_using(
+    const tripoint &examp,
+    climbing_aid_id aid_id,
+    bool deploy_affordance )
```
ehughsbaird, line 13277: The arguments to this function don't seem special enough to need 4 lines to themselves
```suggestion
void game::climb_down_using( const tripoint &examp, climbing_aid_id aid_id, bool deploy_affordance )
```

src/game.cpp
```diff
@@ -13148,196 +13288,196 @@ void game::climb_down( const tripoint &examp )
...
+    you.set_activity_level( ACTIVE_EXERCISE );
+    float weary_mult = 1.0f / you.exertion_adjusted_move_multiplier( ACTIVE_EXERCISE );
+
+    you.moves -= to_moves<int>( 1_seconds + 1_seconds * fall_mod ) * weary_mult;
+    you.setpos( examp );
+
+    // Pre-descent message.
+    if( aid.down.msg_before.length() ) {
+        // TODO LOCALIZE CLIMBING
+        you.add_msg_if_player( aid.down.msg_before );
```
ehughsbaird, line 13429:
```suggestion
        you.add_msg_if_player( aid.down.msg_before.translated() );
```

src/game.cpp
```diff
@@ -13148,196 +13288,196 @@ void game::climb_down( const tripoint &examp )
...
-        g->vertical_move( -1, true );
-    } else {
-        // The player slipped.  This result leaves the player in mid-air, about to fall.
-        return;
     }
+
+    // Pre-descent message.
+    if( aid.down.msg_after.length() ) {
+        // TODO LOCALIZE CLIMBING
+        you.add_msg_if_player( aid.down.msg_after );
```
ehughsbaird, line 13461:
```suggestion
        you.add_msg_if_player( aid.down.msg_after.translated() );
```

src/iexamine.cpp
```diff
@@ -965,7 +966,7 @@ class atm_menu
  */
 void iexamine::atm( Character &you, const tripoint & )
 {
-    atm_menu {you}.start();
+    atm_menu {you} .start();
```
ehughsbaird, line 969:
```suggestion
    atm_menu {you}.start();
```

src/iexamine.cpp
```diff
@@ -5048,110 +5047,111 @@ void iexamine::ledge( Character &you, const tripoint &examp )
...
-    switch( cmenu.ret ) {
-        case ledge_jump_across: {
-            // If player is grabbed, trapped, or somehow otherwise movement-impeded, first try to break free
-            if( !you.move_effects( false ) ) {
-                you.moves -= 100;
-                return;
-            }
+    if( g->climb_down_menu_pick( examp, cmenu.ret ) ) {
+        // This branch means the player chose some option generated by the climb menu.
+    } else switch( cmenu.ret ) {
```
ehughsbaird, line 5063: This means we don't need to them indent the whole switch another level. I can't tell if you made functional changes to the switch because it's also been indented in an entire level.
```suggestion
    if( g->climb_down_menu_pick( examp, cmenu.ret ) ) {
        // This branch means the player chose some option generated by the climb menu.
        return;
    }
    switch( cmenu.ret ) {
```

src/climbing.cpp
```diff
@@ -0,0 +1,406 @@
...
+        }
+    }
+}
+
+climbing_aid::aid_list climbing_aid::list(
+    const condition_list &conditions )
+{
+    std::vector<const climbing_aid *> list;
+
+    const auto add_deployables = [&]( const condition &, const climbing_aid & aid ) {
```
ehughsbaird, line 255: I think capturing everything is bad style.
```suggestion
    const auto add_deployables = [&list]( const condition &, const climbing_aid & aid ) {
```

src/climbing.cpp
```diff
@@ -0,0 +1,406 @@
...
+
+    return list;
+}
+
+climbing_aid::aid_list climbing_aid::list_all(
+    const condition_list &conditions )
+{
+    std::vector<const climbing_aid *> list;
+
+    const auto add_climbing_aids = [&]( const condition &, const climbing_aid & aid ) {
```
ehughsbaird, line 272:
```suggestion
    const auto add_climbing_aids = [&list]( const condition &, const climbing_aid & aid ) {
```

src/climbing.cpp
```diff
@@ -0,0 +1,406 @@
...
+
+    return list;
+}
+
+const climbing_aid &climbing_aid::get_safest(
+    const condition_list &conditions, bool no_deploy )
+{
+    const climbing_aid *choice = climbing_aid_default;
+
+    const auto choose_safest = [&]( const condition &, const climbing_aid & aid ) {
```
ehughsbaird, line 287:
```suggestion
    const auto choose_safest = [&choice]( const condition &, const climbing_aid & aid ) {
```

src/climbing.cpp
```diff
@@ -0,0 +1,406 @@
...
+    };
+    auto detect_ter_furn_flag = [&]( condition & cond ) {
+        cond.range = fall.height_to_floor_or_furniture();
+        return here.has_flag( cond.flag, p_nearest_furn );
+    };
+    auto detect_vehicle = [&]( condition & cond ) {
+        // TODO implement flags and range?
+        cond.range = 1;
+        return fall.height >= 1 && here.veh_at( p_just_below );
+    };
```
ehughsbaird, line 357:
```suggestion
    auto detect_you_character_flag = [&you]( condition & cond ) {
        return you.has_flag( json_character_flag( cond.flag ) );
    };
    auto detect_you_trait = [&you]( condition & cond ) {
        return you.has_trait( trait_id( cond.flag ) );
    };
    auto detect_item = [&you]( condition & cond ) {
        cond.uses_item = you.amount_of( itype_id( cond.flag ) );
        return cond.uses_item > 0;
    };
    auto detect_ter_furn_flag = [&here]( condition & cond ) {
        cond.range = fall.height_to_floor_or_furniture();
        return here.has_flag( cond.flag, p_nearest_furn );
    };
    auto detect_vehicle = [&here]( condition & cond ) {
        // TODO implement flags and range?
        cond.range = 1;
        return fall.height >= 1 && here.veh_at( p_just_below );
    };
```

src/iexamine.cpp
EvanBalster: This one's unrelated to my changes and I've gone to lengths to leave it alone based on the expectation that another PR will address it.

data/json/climbing.json
```diff
@@ -0,0 +1,228 @@
...
+    "slip_chance_mod": -100,
+    "down": {
+      "menu_text": "Detach a vine so you can climb back up (<color_yellow>this will hurt</color>).",
+      "menu_cant": "Can't detach a vine (something is in the way).",
+      "menu_hotkey": "v",
+      "confirm_text": "Detach a vine to descend?",
+      "msg_after": "You descend on your vines, though leaving a part of you behind stings.",
+      "max_height": 1,
+      "cost": { "pain": 5, "damage": 5, "kcal": 87, "thirst": 10 },
+      "deploy_furn": "f_web_up",
```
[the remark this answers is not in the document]
EvanBalster: AFAICT, the vine ability previously created "rope leading up" furniture, which is now used for grappling hooks.  That code no longer works, for various reasons (seems like open air used to be a type of stairs?).  In any case there isn't a dedicated furniture type for a vine and one needs to be added.  Now that I'm pushing JSON changes I might as well do this myself, I suppose.

ADekema:
I don´t know if this is out of scope for this PR. But if you are reworking the slip chance conculations shouldn´t skill also factor in? Something like having athletics skill or perhaps a few climbing proficiencies (novice, proficient, etc. climbing) reducing slip chance allowing you to climb safely even in adverse conditions or compensating for encumbrance and wounds.

src/iexamine.cpp
ehughsbaird: It's showing up in your diff, so it's up to you to change it in your branch and prevent it from being someone else's problem.
https://github.com/CleverRaven/Cataclysm-DDA/blob/d7bcff2c2fcd22581273eb8b2f4e1caa61e60199/src/iexamine.cpp#L968

ehughsbaird:
This PR is so big it's hard to review, it'll not be an easy place to discuss additional balance changes.

EvanBalster:
@ADekema I feel the same as @ehughsbaird ; this started as a balance PR with code cleanup elements and has ballooned into a new system.  I'm inclined to limit balance changes to the introduction of slip chance modifiers for now.  A followup PR focusing on balance would be in order afterward.

Also for reasons of scope, I think I'm going to wait and do a separate PR integrating this system with climbing up and climbing over.

src/iexamine.cpp
EvanBalster: Ah, I see.  This diff is actually a result of applying the CDDA standard `astyle` formatting rules to `iexamine.cpp`.  Notably, the `astyle` command line recommended by the docs is in conflict with the automated review tools.  Contention over this space has been causing tests to fail on all the PRs lately.

ehughsbaird: Oh, really? I did need to install an older version of astyle to get styling working, so I guess there's something up there.

data/json/climbing.json
```diff
@@ -0,0 +1,228 @@
+[
+  {
+    "type": "climbing_aid",
+    "id": "default",
+    "//": "SPECIAL, DO NOT REMOVE. This is used when no items, terrain or abilities aid in climbing.",
```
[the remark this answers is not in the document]
EvanBalster: That file appears to define "null" objects for a few types, but most of the game's types don't seem to use it.  In the case of climbing aids, the "default" aid is valid to pass as an argument to various functions, where it represents the lack of any special affordance.  For example, when climbing down, the default aid is a bare ledge.

EvanBalster:
This PR is now feature-complete.  I have decided to leave the integration of climbing-up mechanics into this new system as a subject for a future contribution.

At this moment it has passed all tests (except for pending Windows build and clang-tidy which is expected green after recent fixes).  I've integrated with the language string parser and have just pushed a new doc page describing the new `climbing_aid` JSON type in detail

esotericist:
i love the idea of what you're trying to do, but i think you've gotten a little too enthusiastic with all of what you've done in this pull request. not because any of it shouldn't be done (i'm not in a position to comment on any of that), but because of how many things you've got going on at the same time; you have:

*  a subsystem refactor with jsonification
*  a rebalance of existing logic
*  unification of multiple code paths that were previously distinct logic
*  new mechanics
*  new content 

i would strongly encourage you to flip this back to draft and keep it open as a reference as you break it open into multiple discrete pull requests that are more contained, so that our reviewers are able to evaluate the changes on their specific merits without other factors confusing things.

of particular note, when jsonifying a subsystem, it's usually best for such a pull request to focus on replicating the prior gameplay behavior so that those of us on our team who deal only in c++ and don't want to have to ask questions about balance or design can more readily go "yes, this appears to do the thing i can objectively evaluate". similarly, we have people who comment on balance and content who aren't as prepared to look at c++.

i think if you develop this more incrementally, it's more likely to get integrated more quickly (or at the very least, you're likely to get more useful feedback on discrete elements preventing integration if the people who would be reviewing those elements aren't also having to work their way through everything ELSE they might not be prepared to review).

of course, having this PR extant to link back to will also help provide context for people to understand where you're going with things, which is why i recommend leaving it as draft. but as it stands, i fear you've got so many things going on, you're going to have trouble getting people to actually review this properly, and with the enthusiasm you've shown, i don't want your work to languish.

EvanBalster:
> @esotericist  i would strongly encourage you to flip this back to draft and keep it open as-is as a reference as you break it into multiple new discrete pull requests that are more contained, so that our reviewers are able to evaluate the changes on their specific merits without other factors confusing things.

This makes sense to me.  My work here started as a rebalance with one or two new features and code cleanup.  It became something of a learning opportunity when somebody suggested I JSONize the mechanics.  Now it's a JSON conversion with bits and pieces of that other stuff left in.

I'll be following your suggestion.  I'm inclined to break it up as follows:

1. Jsonification + code cleanup + UX  (ie, most/all of the C++ changes, no substantive gameplay changes)
2. Rebalancing existing mechanics
3. Fixing / adding some commonsense climbing methods

esotericist:
that sounds great. then once those are nailed down, we get to while all the new stuff that this can bring. both the stuff you started on here, as well as whatever else you were envisioning

src/climbing.h
```diff
@@ -0,0 +1,160 @@
...
+            special = 0,
+            ter_furn,
+            veh,
+            item,
+            character,
+            trait,
+            last
+        };
+
+        struct condition {
```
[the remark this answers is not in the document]
EvanBalster: I think it's a likely direction for future work on this feature!  But for the moment I worry it'd be overengineerings things.  Especially given I'm in the process of breaking this PR up into several smaller PRs now.

src/climbing.cpp
```diff
@@ -0,0 +1,406 @@
...
+{
+    // Build the climbing aids by condition lookup table
+    climbing_aid_default = nullptr;
+    climbing_lookup.clear();
+    climbing_lookup.resize( int( category::last ) );
+
+    for( const climbing_aid &aid : get_all() ) {
+        int category_index = int( aid.base_condition.cat );
+        if( category_index >= int( climbing_lookup.size() ) ) {
+            // should throw, really.
```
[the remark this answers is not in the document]
EvanBalster: Addressing this in #68265.

src/climbing.cpp
```diff
@@ -0,0 +1,406 @@
...
+    case climbing_aid::category::ter_furn: return "ter_furn";
+    case climbing_aid::category::veh: return "veh";
+    case climbing_aid::category::item: return "item";
+    case climbing_aid::category::character: return "character";
+    case climbing_aid::category::trait: return "trait";
+        // *INDENT-ON*
+        case climbing_aid::category::last:
+            break;
+    }
+    cata_fatal( "Invalid achievement_completion" );
```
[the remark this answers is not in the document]
EvanBalster: Resolved in forthcoming commit to #68265.

src/climbing.cpp
```diff
@@ -0,0 +1,406 @@
...
+            menu_hotkey = menu_hotkey_str[ 0 ];
+        }
+
+        optional( jo, true, "msg_before", msg_before );
+        optional( jo, true, "msg_after", msg_after );
+        optional( jo, true, "cost", cost );
+        optional( jo, true, "deploy_furn", deploy_furn );
+    }
+
+    was_loaded = true;
```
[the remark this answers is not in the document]
EvanBalster: Resolved in forthcoming commit to #68265.

src/game.cpp
EvanBalster: Good point.  Resolved in forthcoming commit to #68265.

src/game.cpp
EvanBalster: Resolved in forthcoming commit to #68265.

data/json/climbing.json
```diff
@@ -0,0 +1,236 @@
...
+      "msg_after": "You lower the ladder into place and climb down.",
+      "//": "This removes a stepladder from inventory and places it on the ground below.",
+      "max_height": 1,
+      "allow_remaining_height": false,
+      "deploy_furn": "f_aluminum_stepladder",
+      "easy_climb_back_up": 1
+    },
+    "condition": { "type": "item", "flag": "aluminum_stepladder", "uses": 1 }
+  },
+  {
```
EvanBalster, line 98: This climbing aid was removed from PR #68265 but I intend to add it in a followup contribution

EvanBalster:
#68265 has been merged and the rest of my changes from this PR are in #68306 — meaning there's no unique content here anymore.  Closing.

PR #68197, opened 2023-09-18
Amphibian Additions Phase 2 - Monstrous Frogs & Toads

TheSaddestGoomba:
#### Summary
Content "Introduce new mutant frog monsters, and weakpoints for frogs"

#### Purpose of change
Expand the number of amphibian mutants
Second implementation of #67029, follow-up to #67259 

#### Describe the solution

- Add new monsters (mutant frog species & their tadpoles), eggs and monstergroups for them to reproduce and spawn.
- Add weakpoint sets for amphibians, and effects for those weakpoints.
- Add new generic harvest entry.
- Old monsters have been replaced in monstergroups, and rotspawn/upgrade/reproduction references. For Megafauna mod as well. ~~Old monster entries not yet removed.~~
- Updated all amphibian data in Tamable Wildlife.
- ~~Fix my own poor judgement; change all instances of "toad" in id's to "frog" for consistency.~~

I'm still hoping to expand on special abilities within this PR. Specifically, I want to:

- ~~Add an area effect to one line of mutants representing an overpowering odor/toxin. Triggers vomiting in extreme cases.~~ <- Done through tear gas field emission
- ~~Give giant toads the ability to create pits for resting/wallowing in.~~ <- reserved for later, likely as a nest in mapgen
- ~~Give the Colossal Croaker a colossal croak.~~ <- Done
- ~~Overall, needs more tongue. Add more and finetune ranged pulls.~~ <- Done
- Having split evolution for some species to maintain a mix of forms rather than everything ending up all megas after a year. Best implementation I could think of was to create near-indentical copies of lower level mutants. One version continues evolving and the other stops where they are. This seems like overkill for now, but I may try it out for a future treefrog mutant line with the reasoning of extreme sexual dimorphism.

#### Testing
Tested through debug in a fresh world:
- Map spawns
- Evolutions
- Special Attacks
- Match-up fights with NPCs and other monsters. Extra attention was paid to frog/arthropod match-ups. As is, they seem to be pretty evenly matched at the equivalent 'size' or 'stage' of mutation, with the frogs relying on their abilities to pull out a victory.


ehughsbaird (changes requested):
Item id changes require migration: https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/json/obsoletion/migration_items.json
It doesn't look like such a feature exists for monsters, so those shouldn't have their ids changed.

Karol1223:
Removing IDs from a monster silently turns it into a breather upon load - this is not documented.

Having said that - changing an ID for the sake of just changing it is very bad practice. Don't change IDs unless you specifically need to.

And while I'm here I'll also comment on the naming because this happens every time mutants get brought up. 4/5 of the players, at least, do not read descriptions or it isn't their first instinct. Names should not be flowery, poetic or from a kaiju movie - they should clearly show what the monster you're dealing with is. Keep in mind, too, that these mutants would not be named by scientists but by normal people who come in contact with them. I'm saying that because the first one I saw was "stuntongue", and no disrespect because the name is cool but it is about as non-descriptive as it gets, so it really doesn't fit.

Zireael07:
To add to what Karol said, flowery/poetic names will also be difficult to translate.

TheSaddestGoomba:
Thank you all for the feedback!

> Item id changes require migration
> Don't change IDs unless you specifically need to.

I did have that in mind this time. I was tripping over myself when prepping the new monsters and referencing back. I kept using "x_x_frog" for the "x_x_toad"s and creating dead references that way. Also why I only used "frog" for the "_big", "_gigan", and "_mega" ids, eliminate the pointlessly alternating terms.
I also foolishly made the mistake of thinking no one else would have built off it before I could made the change, but Standing-Storm went and made a cool-ass spell using them for Magiclysm... which just led me to see that Magiclysm has a monstergroup for the spell using the "toad" ids.
Sorry, that's a lot say it felt like a need at the time I was working on it but it was still a mistake. Would it be best for me to revert the renaming or just update the Magiclysm group from "toad" to "frog"?

> Names should not be flowery, poetic or from a kaiju movie
> To add to what Karol said, flowery/poetic names will also be difficult to translate.

Hmm, I had previously thought of 'Taser Tongue' or simply "Taser Frog" which feel more direct (it uses its tongue like taser electrodes). 'Staticskin Frog' could also be 'Shockskin Frog.' Are there any others that need more work? The 'Foul' line is pretty literal but are those too flowery?
I had not thought of the translation aspect of it. I assume same best practices? Literal, descriptive, and succinct?

Karol1223:
> Are there any others that need more work? The 'Foul' line is pretty literal but are those too flowery?

I specifically didn't mention those because despite them being flowery I do believe they get the message across. However I have a hard time believing anyone would see a big stinky frog in the post-apocalypse and think hmm yes that's a foul hogtoad. But that's 100% nitpicking - like I said I think they do get the message across, the only possible issue I see may be with translations like Zirael said.

TheSaddestGoomba:
I believe this is ready for review.

data/json/items/comestibles/egg.json
```diff
@@ -1257,5 +1257,55 @@
...
+    "type": "COMESTIBLE",
+    "id": "egg_proxy_bigfrog",
+    "copy-from": "egg_proxy_frog",
+    "name": { "str": "cluster of giant frog's eggs", "str_pl": "clusters of giant frog's eggs" },
+    "weight": "160 g",
+    "volume": "1 L",
+    "symbol": "O",
+    "quench": 8,
+    "calories": 200,
+    "description": "A large clump of very large frog eggs.  They're gelatinous, transparent and larger even than chicken's eggs. ",
```
TheShadowFerret, line 1271:
```suggestion
    "description": "A large clump of very large frog eggs.  They're gelatinous, transparent and larger than even chicken's eggs. ",
```
(or remove the word even entirely)

data/json/monsters/reptile_amphibian.json
```diff
@@ -600,17 +540,20 @@
...
-    "upgrades": { "age_grow": 120, "into": "mon_frog_small" },
+    "upgrades": { "age_grow": 15, "into": "mon_strange_bigfrog" },
     "reproduction": { "baby_egg": "egg_strange_frog", "baby_count": 16, "baby_timer": 26 }
   },
   {
     "id": "mon_fowler_toad",
     "type": "MONSTER",
     "copy-from": "mon_proxy_frog",
-    "description": "Fowler's toads secrete a harmful and distasteful compound from their backs.  This helps to safeguard them from predators and also gives them a distinctive odour like boiled peanuts.",
+    "description": "Fowler's toads secrete a harmful and distasteful compound from their backs.  This helps safeguard them from predators and also gives them a distinctive odour like boiled peanuts.",
```
TheShadowFerret, line 556:
```suggestion
    "description": "Fowler's toads secrete a harmful and distasteful compound from their backs.  This helps safeguard them from predators and also gives them a distinctive odor like boiled peanuts.",
```
Game set in America gets American English spellings

Karol1223:
If you have not rebased this on a master after the PR introducing the flag got merged, the tests may not work

TheSaddestGoomba:
The flag issue ended up being far simpler, "SMALL_HIDE" vs "SMALL_HIDER".

There is now an error with overmap test. I don't think anything in the PR should have touched mapgen.

> If you have not rebased this on a master after the PR introducing the flag got merged, the tests may not work

Rebasing is pulling the CleverRaven master to local master, then pushing to branch? If so, I've done that.

Karol1223:
> Rebasing is pulling the CleverRaven master to local master, then pushing to branch? If so, I've done that.

No, rebasing is different from pushing master to your branch. It's pushing the branch on top of master... or something like that. If you're using GitHub desktop it's just called rebasing. I am not aware of how it's handled in other UIs/on the website.

PR #32013, opened 2019-06-30
Salvaged CBMs are Filthy

Fris0uman:
#### Summary

```SUMMARY: Balance "Salvaged CBM needs to be sterilized and re-conditioned"```

#### Purpose of change
Fixes #20383

 - [x] Salvaged CBM are *filthy* and *deployed* (including uninstalled CBMs)
 - [x] Autoclave clean and sterilize CBM
 - [x] CBM must be repackaged to stay sterile
 - [x] Autodoc requires sterile CBM
 - [x] Autodoc will require properly conditioned CBM
 - [x] Surgeon ~~might be more leniant~~ is strict with hygiene as he should
 - [x] Autoclave is available in furniture AND item format
 - [x] Jsonize `faults` and `flags` for harvest entry: **filthy cbms from dissection can be modded out** 
 NB: filthy CBMs from surgery are still hardcoded

#### Describe the solution

- Add FILTHY, NO_STERILE, and NO_PACKED flags to slavaged CBM: from dissection and surgery
- Add iexamine to autoclave furniture so that it consumes water and replace FILTHY tag with STERILE tag
- Add autoclave tool that does same thing but also consume battery (a lot) and only sterilize one CBM at a time
- Add tool to allow proper packaging
- Make autodoc and surgeon require sterile CBM before instalation.

#### Additional context
Screen Shots:
Autodoc:
image

Inventory:
image

Surgeon:
image

Autocalve item is based on [this](https://www.quirumed.com/se/black-dragon-8-liter-autoklav-klass-b-tattoo.html)
It uses 2 L of water per cycle, a cycle is 90 min long and can sterilize one CBM.
It also has a `"power_draw"` of 1 500 000 mW (cf. link) which means that, without adding mod to it, it can only be powered by a `heavy_atomic_battery_cell` and it will only be enough for 1 cycle. Is it acceptable?

How it works:
\> Dissect CBM contening monster
\> Get `faulty cbm ( filthy )`
\> Mend cbm, using `fine bolt turning` and `fine screwdriving`with the appropriate skill
\> Clean CBM using `detergent` / `soap` and a `sponge`
\> Use `autoclave pouch` to put clean CBM in protectting pouch. There's a first aid skill check here that cannot failed if lvl > 6
\> Put `cbm (packed)` either in autoclave furniture or in autoclave tool. NB: the tool require a **lot** of power you might need to mod it so that it can run on storage battery power
\> Wait 90min for completion of autoclave cycle
\> Get `cbm (sterile)`
[...]

PR #68202, opened 2023-09-18
Characters with Infection Immune trait no longer need to sterilize bionics prior to installation

Night-Pryanik:
#### Summary
Features "Characters with Infection Immune trait no longer need to sterilize bionics prior to installation"

#### Purpose of change
Characters with Infection Immune trait are, well, immune to infections, including those which can transmit via installation of non-sterile bionics. This change feels plausible and make trait in question more valuable.

#### Describe the solution
Ported https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/60 and https://github.com/Cataclysm-TISH-team/Cataclysm-TISH/pull/67:
- Added check for the trait in `Character::is_installable` and `install_bionic_actor::can_use` functions.
- Updated trait's description.

- Characters with Disease Immune should gain this advantage too? On the other hand, "disease" in this context only relates to viral diseases like common cold and flu, and I'm not sure whether we should take other viral pathogens into consideration since we don't simulate them.
- Permit installing not only non-sterile bionics, but also filthy ones?

#### Testing
Got Infection Immune trait. Got non-sterile bionic. Found autodoc. Tried to install the bionic. Success.

#### Additional context
None.

### DIFF
--- a/data/json/mutations/mutations.json
+++ b/data/json/mutations/mutations.json
@@ -3782,7 +3782,7 @@
     "id": "INFIMMUNE",
     "name": { "str": "Infection Immune" },
     "points": 3,
-    "description": "Your bloodstream has developed antibiotic properties.  Your wounds will never become infected.",
+    "description": "Your bloodstream has developed antibiotic properties.  Your wounds will never become infected.  You will also be able to install non-sterile CBMs.",
     "prereqs": [ "DISRESISTANT", "DISIMMUNE" ],
     "prereqs2": [ "INFRESIST" ],
     "category": [ "MEDICAL" ],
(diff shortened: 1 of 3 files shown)


Fris0uman:
I porposed that in the initial implementation #32013 and it was not accepted

https://github.com/CleverRaven/Cataclysm-DDA/pull/32013#issuecomment-508812726

kevingranade:
Nothing has changed since #32013, yes this isn't gonna fly.

Night-Pryanik:
Ok, I'm fine with that, though I think the trait's name should be changed in that case. After all, trait's current name led to confusion and subsequent creation of this PR.

PR #68206, opened 2023-09-19
Isolation Protocol: Better Danger Clock

John-Candlebury:
#### Summary
Mods "Isolation Protocol: Improve the Danger Clock"

#### Purpose of change
The danger clock serves the important function of actually pushing the player forward and not letting them patiently and safely clear the lab levels. It however needs some flexibility to let you engage with time consuming game systems that aren't combat related.

So some sort of compromise is needed.

#### Describe the solution
Reworked the danger clock so it ticks up much faster in lab levels, but also implemented shelter style levels where the player is free to sleep, craft objects or otherwise train and recuperate.  I added an UI widget to properly communicate the danger clock mechanic
and clearly distinguish the safe level from the rest. These levels spawn every 5 normal levels in the fractal facility scenario

Additionally I added the class perk for the Pizza Delivery guy.
Since they are a generalist class the start with: "The Deliverator" perk, which increases player quickness as reality deteriorates. Technicaly its uncapped but surviving past the point where the perk grants +20 speed is pretty hard.

#### Testing
Crawled the dungeon for a bit

### DIFF
--- a/data/mods/Isolation-Protocol/EOC/danger_clock_eoc.json
+++ b/data/mods/Isolation-Protocol/EOC/danger_clock_eoc.json
@@ -2,20 +2,17 @@
   {
     "type": "effect_on_condition",
     "id": "ISO_DANGER_CLOCK",
-    "recurrence": "80 minutes",
+    "recurrence": "3 minutes",
     "global": true,
-    "condition": { "not": { "u_is_on_terrain": "t_grass" } },
-    "effect": [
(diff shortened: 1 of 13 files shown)


martinrhan:
How to enable that UI widget?

John-Candlebury:
Should be visible by default. Its in the bottom of the unedited UI. Reads something like:

`"Normality:    Deteriorating"`

Depending on the danger level.

martinrhan:
I tried all kind of sidebar options and cannot find it

PR #68214, opened 2023-09-19
Bulk loading/unloading items into container saves time cost

lispcoc:
#### Summary
Balance "Bulk loading/unloading  items into container saves time cost"

#### Purpose of change
We had been looking for various opinions here regarding the time cost of handling large quantities of items. (https://github.com/CleverRaven/Cataclysm-DDA/pull/68004)
There were various opinions, but I think there is no particular disagreement with the opinion that it saves time by putting items in bulk when putting them in a specific container.
Picking up sand from a sand pile can be thought of as scooping sand with a cardboard box or plastic bag.

#### Describe the solution
Currently, the cost to load one item is as follows. (item_move_cost() in activity_actor.cpp)

cost = obtain_cost + move_cost

Both obtain_cost and move_cost includes overhead cost of reaching the item, such as the time it takes to walk to adjacent tiles.
I think this overhead is unnecessary for the second and subsequent items, so that is applied only once in new formula.
The cost to load excluding overhead is set as follows based on volume penalty of item_handling_cost() in character.cpp.

cost = min( 200, item_volume / 20_ml )

This is applied unloading too.

There are various ways of thinking about how to calculate costs.
- Apply formula suggested in https://github.com/CleverRaven/Cataclysm-DDA/issues/67385
- Same cost as liquid. (~666ml/s)

#### Testing
spawn items, press "v" and load items into backpack from floor.

Consumed time(1000 salts)
Before: 21:40
After: 00:11

Consumed time(100 apples)
Before: 2:21
After: 00:12

press "U" and unload items from backpack into another backpack.

Consumed time(1000 salts)
Before: 16:40
After: 00:11

Consumed time(100 apples)
Before: 1:51
After: 00:12

I feel that the time it takes to load apples is a little short, but I think it is not far from reality.

#### Additional context
This only applies to explicitly storing items in a specific container. (v key) Does not apply to normal item picking up process. (g key)
I think a different approach is needed to save time on the generic process of picking up items.
There may be some people think it's strange that you can pick up powdered items without leaving them behind, but I don't think that's something that should be discussed in this PR.

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -6,6 +6,7 @@
 #include <cstddef>
 #include <cstdint>
 #include <functional>
+#include <iterator>
 #include <list>
 #include <map>
 #include <new>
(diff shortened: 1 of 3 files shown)


Nebnis:
Does this apply to unload a container too?

I also agree that the apples feels a bit too short, how many seconds would be if you applied for solid cost to move instead of liquid?

lispcoc:
> Does this apply to unload a container too?

It took me a while to research the unloading process. Also applied to unloading.

CHCOOH:
This may not be the right place to discuss this, but I think the same thing can be said about the normal item acquisition process in (g).
There is really no big difference between (v), where the player explicitly puts the item in a specific container, and (g), where the player picks up the item.
The player just does not choose a specific container, but he or she is still putting the item into the container after all.
If I had to say so, I would say that there might be some re-calculation of the cost when placing the item in each container, but I don't think this is a good idea.

PR #68215, opened 2023-09-19
Anger npc if player asks for equipment

### LINKED ISSUE #61861: Exploit: Can gain endless social skill by asking an NPC for items over and over again.
### Describe the bug

There is no limit to the amount of times you can ask an npc for items through the "Want to share some useful items with me?" dialogue. You gain social skill for going through these dialogue checks even at a guaranteed 0% chance of success, so you can grind your social skill theoretically indefinitely, all in a single in-game second. (it becomes very slow to do after social level 2, though).

Another effect is that you can make the NPC's "Value" opinion of you will spiral into extremely negative values because it decreases each time you do this.

### Steps to reproduce

1. Make character with 0 social skill in a scenario with starting NPC
2. Ask the starting NPC for useful items
3. Repeat step 2 over and over again. See that social skill rises up with no limit.

### Expected behavior

I kind of expected that they'd actually follow through with their threats to kill me if I asked them again. A perfectly reasonable response for when The Flash won't take no for an answer and keeps asking for handouts hundreds of times per second.

Maybe the player should get locked out of asking for items if they fail too many times, similar to asking them to become your follower.

### Screenshots

_No response_

- OS: Windows
- Game Version: 3c1b99e [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

Leonkoithara:
#### Summary
Bugfixes "NPCs get angry when player asks for equipment"

#### Purpose of change
Fixes #61861 

#### Describe the solution
Increase the NPCs anger towards the player so that they cannot keep indefinitely asking for items to increase social skill without them turning hostile.

#### Testing
Spawn in an npc and keep asking them to share items until they turn hostile.

### DIFF
--- a/src/npctalk_funcs.cpp
+++ b/src/npctalk_funcs.cpp
@@ -587,8 +587,6 @@ void talk_function::give_equipment_allowance( npc &p, int allowance )
 {
     std::vector<item_pricing> giving = npc_trading::init_selling( p );
     int chosen = -1;
-    // Add asked_for_item effect whether or not NPC has items to give
-    p.add_effect( effect_asked_for_item, 3_hours );
     while( chosen == -1 && !giving.empty() ) {
         int index = rng( 0, giving.size() - 1 );
         if( giving[index].price < p.op_of_u.owed + allowance ) {
@@ -615,6 +613,7 @@ void talk_function::give_equipment_allowance( npc &p, int allowance )
     if( allowance < 0 ) {
         p.op_of_u.owed += allowance;
     }
+    p.add_effect( effect_asked_for_item, 3_hours );
 }
 
 void talk_function::lesser_give_aid( npc &p )
--- a/src/npctalk_funcs.cpp
+++ b/src/npctalk_funcs.cpp
@@ -587,6 +587,8 @@ void talk_function::give_equipment_allowance( npc &p, int allowance )
 {
     std::vector<item_pricing> giving = npc_trading::init_selling( p );
     int chosen = -1;
+    // Add asked_for_item effect whether or not NPC has items to give
+    p.add_effect( effect_asked_for_item, 3_hours );
     while( chosen == -1 && !giving.empty() ) {
         int index = rng( 0, giving.size() - 1 );
         if( giving[index].price < p.op_of_u.owed + allowance ) {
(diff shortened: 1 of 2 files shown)


src/npctalk_funcs.cpp
```diff
@@ -587,6 +587,8 @@ void talk_function::give_equipment_allowance( npc &p, int allowance )
 {
     std::vector<item_pricing> giving = npc_trading::init_selling( p );
     int chosen = -1;
+    // Add asked_for_item effect whether or not NPC has items to give
+    p.add_effect( effect_asked_for_item, 3_hours );
```
Maleclypse, line 591:
```suggestion
    p.add_effect( effect_asked_for_item, 3_hours );
```
3 days or even a week seems more reasonable than able to ask every three hours?

MNG-cataclysm, line 591: Couldn't this be done with regular JSON instead of C++? It would be easier that way and more accessible to the general programming base.

MNG-cataclysm (changes requested):

src/npctalk_funcs.cpp
Leonkoithara: It can be done via JSON however I'm not sure what else depends on this effect to be added when this function is called. Therefore I could just let the npctalk_funcs.cpp remain the same and modify the JSON to add the effect or just slightly modify this file and wait for someone else to move all effects to the JSON instead of hardecoding it.

Leonkoithara: Yes will change it to 3 days once the other conversation is resolved.

MNG-cataclysm: All you'd need to do is remove the C++ function, make the effect have a time on it, and use an if conditional statement for the associated dialogue so it wouldn't appear if the NPC had this effect.

Leonkoithara: Hi reverted the change in code and made the effect via JSON it so that when we receive item from NPC the effect is added. I did not remove the function entirely because various other functions depend on calling this function and the effect to be added and moving all of them to JSON is out of scope of this PR and I'm not entirely certain what all functions do call it, but perhaps something to look into at a later date.

Leonkoithara: @Maleclypse the conversation has been resolved if all else looks good please merge.

PR #68226, opened 2023-09-20
[AIM] Make Move All actually move as much as it can

Kamayana:
#### Summary
Features "Make AIM Move All actually move as much as it can"

#### Purpose of change
When using the AIM to Move All items to a location without enough room, the popup says "Attempt to move as much as you can?", but it doesn't actually do that - it just naively tries to move all items in an unsorted order, so if a huge item is at the top of the list it'll fail to move anything.

#### Describe the solution
The list of moving items is now sorted by weight or volume, whichever is closer to passing the volume/weight limit. This ensures that the game actually tries to move as many items as it can.

In addition, there are a few new AIM behaviors:

- If you attempt to move a filled bucket to a container/vehicle cargo, you will now get a prompt to empty the bucket and store both it and the contents at the location, rather than simply failing.
- Using Move All from [W]orn will first move all armor, but will now also move (after asking) the wielded weapon if it's the only item remaining. Similarly, if a bucket is the only item remaining you'll get the bucket emptying prompt.
- The marked relative location of the container will now update positions if you move around the container in the world.

There's an odd behavior where moving all to [I]nventory has slightly different results from moving all to the specific [C]ontainer, which is especially apparent when only one container is worn. I tried to fix it within the AIM code but I think it comes from a fundamental difference between how `pickup_activity_actor` and `insert_item_activity_actor` handles pocket priorities, and thus would be harder to fix than I'm willing to take on.

An odd part of moving liquid-filled buckets is that if prompted to spill the contents, it asks you to choose a direction with the AIM window still displayed in front of everything. I was working on a fix for this, but realized it's better suited to its own PR because it has a greater scope, such as affecting items that are used via the AIM's examine window (use a joint to set a fire and it'll prompt for a direction with the AIM in the way).

It is possible that moving a bucket and its contents to a container like a backpack will move the contents but NOT the bucket - this is because the bucket needs to be moved last, but that means the one pocket that's big enough for the bucket can get filled with the contents first, leaving no room for it. There's not an easy way to fix that, or even check for it - you'd have to adjust the insertion activity itself to manage bucket emptying, it can't easily be accomplished outside of it without doing some weird behaviors. Since the player would most likely prioritize the contents anyway I left it alone.

#### Testing
Save I used for testing changes: [Charge Removal Testing-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/13526055/Charge.Removal.Testing-trimmed.tar.gz)
Moved items to and from backpack, buckets, shopping cart, making sure as much is moved as possible. Moved liquid or item-filled buckets to backpack and shopping cart to see that the spill prompts work properly.


Kamayana:
The scope of this PR kinda expanded, but you now get a prompt to spill or transfer liquids in liquid-filled buckets. An odd part of this is that if prompted to spill the contents, it asks you to choose a direction with the AIM window still displayed in front of everything. I was working on a fix for this, but realized it's better suited to its own PR because it affects items that are used via the AIM's examine window (like using a joint to set a fire, it'll prompt for a direction with the AIM in the way).

Maleclypse:
Very neat!

PR #68242, opened 2023-09-21
Organize the EOC files for better accessibility

MNG-cataclysm:
#### Summary
Infrastructure "Organize the EOC files for better accessibility."

#### Purpose of change
The game's EOCs were previously located in multiple places, and we had a very large generic EOC file that contained 99% of the EOCs. Because it covered so much, this file became inefficient to navigate, and made finding particular EOCs harder than it should be. Additions were simply scattered throughout, with no rhyme or reason other than the order they were added in. With the ever-expanding power of EOCs, and with how many people were starting to use them in greater capacity, I concluded it was time for a new, better system.

#### Describe the solution
Create a new folder to house all of the EOCs, and break up the generic `effects_on_condition` file into more specialized files, named based on what the contained EOCs do. If there were a lot of a given kind of EOC in multiple varieties, sub-folders were created to store more organized subsets. A generic EOC file still exists, and is used for things that don't fit in the other categories, or had too few kinds of which to warrant a whole dedicated file.

Not doing this, and keeping a gargantuan generic file.

#### Testing
Everything loaded fine, didn't experience any problems.

#### Additional context
None at this time.


Maleclypse:
Normally we wait until files are significantly larger to reorganize them but since EOCs are so varied in what they apply to I think this is a good move now.

PR #68244, opened 2023-09-21
[EOC]Inventory itereter

lispcoc:
#### Summary
Features "[EOC]Inventory itereter"

#### Purpose of change
Items become beta talker only when EOC invoked by using them.
I think various things can be achieved if EOC can process items in the inventory without going through "Use".

#### Describe the solution
Implement itereter for inventory.
Make Parsing tags consider items.

Use case: display what you wear in message
```
[
  {
    "id": "EOC_display_inv",
    "effect": [
      {
        "u_run_inv_eocs": [ { "id": "EOC_display_inv_nested", "effect": [ { "u_message": "<npc_name>" } ] } ],
        "worn_only": true
      }
    ]
  }
]
```
Currently there aren't many functions related to items, so we won't be able to do much, but I think it will be possible to do things like teleport inventory items to a specified location.

#### Testing
Add "EOC_run_inv_test".


ehughsbaird:
A use case for this would be helpful - it's easier to copy and change code than to make it frob scratch.

PR #68254, opened 2023-09-22
Remove the token bash damage on belts

gettingusedto:
#### Summary
None

#### Purpose of change

After some discussion at the devcord, the bash damage on belts is completely arbitrary and probably should be removed, so i did it.

#### Describe the solution

Delete the melee damage on belts in the belts.json

Not removing the melee damage but only reducing it to 1.

#### Testing

Before i applied the change into my build, the five affected belts (broad belt, firefighter belt, leather belt, police duty belt, and tool belt) has the melee stats in their (e)xamine screen. it doesnt appears after i applied the change to my build.


KHeket:
This logic is incorrect. The damage of fists hits in reality is calculated from lots of values, but we will take in mind for this case only 2 of them - speed and weight. The more heavier you hand is and the more fast you hit is - the mor damage you will deal. This is princeples of knuckles. You can even take a lighter - it will have 1-10+ gramm weight, but it also make your hit a bit more powerfull. And also metal belt buckle is hard, what is positively work on damage
As exapmle - molle belt. It have 210g weight, when steel knuckles have weight of 240 gramm, but they are consructed for melee fight. So I thinks, it better to leave at least 1-3 bash damage to belts determinated on their weight

KHeket:
> Hitting someone with a belt does basically no meaningful damage and only causes pain

Read again my first message. You dont significantly inrease surface area of your fist by wrapping belts around it, but you significantly inrease the wieght of your fist, so bels can be used as a melee weamon, such in real life

Karol1223:
> You dont significantly inrease surface area of your fist by wrapping belts around it, but you significantly inrease the wieght of your fist, so belts can be used as a melee weapon, such in real life

So we assume wrapping the belt around your knuckles instead of whipping it - this changes things but not significantly. Increasing surface area is minimal, yes, but you're still just wrapping leather around your hand - unless you're hoping to hit someone with the buckle you're not doing any more damage than you'd be without the belt. And even then, hitting someone with a buckle woild be similar to stabbing someone with a key - it would *hurt*, but not do any meaningful damage unless you hit an eye or something like this. Belts should not be useful weapons under any circumstances, because we can't simulate pain on monsters.

KHeket:
> but you're still just wrapping leather around your hand

You understand the way of fist hit incorrect. The weight of fist is ~ 400 grams
image
 So, lets undestand it as 100% of character fist hit damage. If you wrap aroind it a 210 grams molly belt - you will increase the weigt of you fist by 50%. Yes, you will not kill guys with fists in this way, but it will increase damage of hit a bit

KHeket:
> > but you're still just wrapping leather around your hand
> 
> You understand the way of fist hit in wrong way. The weight of average fist is ~ 400-450 grams image So, lets undestand it as 100% of character fist hit damage. If you wrap aroind it a 210 grams molly belt - you will increase the weigt of you fist by 50%. Yes, you will not kill guys with fists in this way, but it will increase damage of hit a bit
The damage(impulse) of fists hit is calculated from weight, speed and surface area of you hand. If we increase the weight of it - the damage will be increased

Karol1223:
> So, lets undestand it as 100% of character fist hit damage. If you wrap aroind it a 210 grams molly belt - you will increase the weigt of you fist by 50%. Yes, you will not kill guys with fists in this way, but it will increase damage of hit a bit

You're still ignoring material hardness - you're wrapping 200g of soft leather that will add weight, while also adding padding to the hit both for you and the target. Combine that with the minor increase in surface area - you end up with roughly equivalent damage unless you happen to hit the target with the buckle. I do not believe this warrants 1 guaranteed additional bash.

KHeket:
The damage(impulse) of fists hit is calculated from weight, speed and surface area of you hand. If we increase the weight of it - the damage will be increased


> while also adding padding to the hit both for you and the target.

Belt leather is hard. And the boxer wearing wearing **much more** soft and **much more** better as padding boxer gloves, but they also make a lot of damage to opponents. And the fist damage not only surface damage, you dont need to pierce and cut skin like in case with key, impulse of you hit like in boxing - strike inside, it damaging organs, brain (concussion), bones etc

Karol1223:
> And the boxer wearing much more soft and much more better as padding boxer gloves, but they also make a lot of damage to opponents.

Boxing gloves exist specifically to DECREASE the impact force, though. I genuinely don't understand where you're going with this. A belt is not damaging enough by itself to function as a proper weapon, I said everything I had to say. This is genuinely not at all important either way, because I refuse to believe anyone but NPCs have ever used belts as weapons - and this change will make NPCs choose them more rarely, at least in theory. I support this change and I've said everything there is to be said. Further comments would jut be going in circles.

KHeket:
> I genuinely don't understand where you're going with this

In realism

MNG-cataclysm:
Never before have I witnessed such an argument over belts, of all things.

Karol1223:
> Never before have I witnessed such an argument over belts, of all things.

Here in the CDDA repository we only discuss the most important things that will certainly actually matter in game for the players.

PR #68255, opened 2023-09-22
Pets follow stairs well

### LINKED ISSUE #35665: Animals don't follow player up stairs

marimarigi:
#### Summary
Balance "Improve pets' ability to use stairs"

#### Purpose of change
Fix #35665
Resolves the problem of following pets getting separated on the stairs.

#### Describe the solution
- Pets that see the avatar will set that location as their destination.
- Cancel the move if it is not necessary.

Basically repeat the above, but stop if pets loses sight of the avatar or there is a difference in Z level.
They will then move to the last set destination, so they can use the stairs without being affected by the avatar's current position.

#### Testing
I made pets such as dogs, horses, turkeys, zombies, and mi-gos to see how they action when moving on stairs or encountering enemies.

#### Additional context
There shouldn't be any effect on the pets' actions other than following, but I don't usually take them around so I can't say for sure.

https://github.com/CleverRaven/Cataclysm-DDA/assets/142076188/734612a3-392f-41a8-831b-8bf54da5f85a


PatrikLundell:
Have you verified that it works with default settings, i.e. experimental 3D view off?

marimarigi:
(I just checked.)
It also worked if I deleted the config folder and set it back to default. Which behavior is more expected?

PatrikLundell:
The desired behavior should be that pets follow you regardless of the 3D view settings, as the visibility thing really is an implementation dependency rather than a core property of the behavior. Thus, it seems it works as it ought to.

Thanks for checking!

PR #68257, opened 2023-09-22
Minor misc JSON edits

Karol1223:
#### Summary
None

#### Describe the solution
- Changed the animal locker to hold medium creatures instead of huge, as per its description. As it stands, it was just a smaller livestock carrier with the same functionality which extended to the vehicle part too
- Changed the description of atomic coffee to reflect the mild lore change it had gotten lately and it's new recipe
- Removed the goo canister recipe - fabrication two and a screwdriver to fabricate your own friendly slimes makes no sense. The item is cool, but it should not be craftable by the player
- Added the wire-making proficiency to all recipes using the wire draw machine (except for the welding wire, because it doesn't seem to actually make new wire?)

I briefly considered removing the goo canister altogether, but I like it. It probably shouldn't spawn friendly slimes, either, but eh, I didn't feel like being a villain today. Besides, it's a cool little surprise to find ingame.

### DIFF
--- a/data/json/items/comestibles/drink.json
+++ b/data/json/items/comestibles/drink.json
@@ -82,7 +82,7 @@
     "healthy": -1,
     "addiction_potential": 8,
     "calories": 35,
-    "description": "This serving of coffee has been created using an atomic coffee pot's FULL NUCLEAR brewing cycle.  Every possible microgram of caffeine and flavor has been carefully extracted for your enjoyment, using the power of the atom.",
+    "description": "This serving of coffee has twice as much coffee as black coffee, as well as some caffeine pills and sugar for good measure.  You didn't plan to sleep this week anyway, right?",
     "price": 300,
     "price_postapoc": 100,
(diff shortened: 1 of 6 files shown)


data/json/items/comestibles/drink.json
```diff
@@ -82,7 +82,7 @@
     "healthy": -1,
     "addiction_potential": 8,
     "calories": 35,
-    "description": "This serving of coffee has been created using an atomic coffee pot's FULL NUCLEAR brewing cycle.  Every possible microgram of caffeine and flavor has been carefully extracted for your enjoyment, using the power of the atom.",
+    "description": "This serving of coffee has twice as much coffee as black coffee, as well as some caffeine pills and sugar for good measure.  You didn't plan to sleep this week anyway, right?",
```
inogenous, line 85: >  You didn't plan to sleep this week anyway, right?

Love it. :heart:

PR #68258, opened 2023-09-22
Molotovs correctly used liquid accelerant

Holli-Git:
#### Summary
Bugfixes "Molotovs used more flammable liquids"

#### Purpose of change

Molotovs require easily flammable liquids. Motor oil, diesel, kerosene aren't flammable enough as their vapor point isn't high enough. The PR that changed it to use flammable liquids (#67068) reasoned that it should use flammable liquids because kerosene was historically used, but their sources didn't back that up, only claiming it had been used in the wick. 

#### Describe the solution

Changes molotovs to use proper requirement.

Make crude and proper molotov recipes, the crude one being the current recipe, while a proper one would require thickening agents for a longer and smokier flame. Would prolly add self igniting/windproof ignition molotovs too 

#### Testing

Yeah

### DIFF
--- a/data/json/recipes/weapon/explosive.json
+++ b/data/json/recipes/weapon/explosive.json
@@ -203,7 +203,7 @@
     "time": "30 s",
     "reversible": true,
     "autolearn": true,
-    "using": [ [ "flammable_liquid", 500 ] ],
+    "using": [ [ "accelerant_liquid", 500 ] ],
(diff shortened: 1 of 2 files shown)


Karol1223:
I'd advise editing the JSON comment about molotovs in ``flammable_liquid``, so this confusion doesn't reoccur in the future.

PR #68276, opened 2023-09-23
Mozzarella is too darn high!

LordBarkBread:
Fixed Mozzarella pricing for traders

#### Summary
Balance "Tweaked Mozzarella pricing"

#### Purpose of change
Mozzarella cheese was outrageously high for a price of an item so I'm kicking it down to more reasonable trade price.

#### Describe the solution
lowered post-apoc price.

N/A

#### Testing
N/A

#### Additional context
N/A

### DIFF
--- a/data/json/items/comestibles/dairy.json
+++ b/data/json/items/comestibles/dairy.json
@@ -302,7 +302,7 @@
     "calories": 85,
     "description": "A ball of white mozzarella cheese.",
     "price": 450,
-    "price_postapoc": 80,
+    "price_postapoc": 60,
     "material": [ "milk" ],
     "primary_material": "cheese",
     "volume": "250 ml",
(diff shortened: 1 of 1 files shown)


data/json/items/comestibles/dairy.json
```diff
@@ -302,7 +302,7 @@
     "calories": 85,
     "description": "A ball of white mozzarella cheese.",
     "price": 450,
-    "price_postapoc": 400,
+    "price_postapoc": 80,
```
Karol1223, line 305:
```suggestion
    "price_postapoc": 60,
```
This is calculated proportionally from calories of meat jerky, which is what docs put as a baseline for item pricing. Technically, by proportions the price is slightly above 61, but mozzarella does spoil twice as fast, so I think rounding it down is acceptable

Karol1223 (changes requested):

PR #68277, opened 2023-09-23
Select default bodypart when applying bandage

inogenous:
#### Summary
Interface "Select default bodypart when applying bandage"

#### Purpose of change

* Makes the default selected bodypart in the "apply bandage"-ui be the bodypart that is the most damaged.
* If the bodypart has a bandage with the same power as the bandage being applied, another bodypart is selected.
* In particular: if there's only one bodypart that needs tending, it will now be the default selected one.

#### Describe the solution

* When populating the selectable bodyparts, keep track of
    * whether the bodypart can be healed further with the item that is currently being applied: `bool gives_improvement`
    * which bodypart is hurt the most, by percentage: `float default_selection_hp_percent` 
* When presenting the ui for selecting bodypart, default to the bodypart that both "can be healed further" and has the lowest hp percentage.

* There may be lots of ways to select a more optimal bodypart to heal first, such as:
    * Prioritizing bleeding bodyparts over other parts that may be more hurt but require long-term healing
    * Prioritizing bites when using a disinfectant
* So there are cases where selecting another more "optimal" default bp would be preferable, but this also means that the logic would be more complicated, which might be unintuitive to the player.
* There's also a minor information leakage here: two bodyparts might show the same `||\..`-bars, but the player does not know which one is actually more hurt than the other since the bars show the same. With this suggested change, the bandaging ui will reveal that since the most hurt part will be the default one even if the bars are the same.

#### Testing

* I got hurt in all bodyparts. Applied bandages one after eachother. The first time the bandaging-ui shows, the most damaged bodypart is pre-selected. The second time, the next-most damaged bp is selected. And so on, until all parts have been bandaged.
* It might not be obvious, but it's possible to `/`-search in the bandaging ui to search for a bodypart to heal. This still works even if the default selection is something else.
* Tested bandages, disinfectant and stopping bleeding.

tis-but-a-scratch

### DIFF
--- a/src/character_body.cpp
+++ b/src/character_body.cpp
@@ -1099,6 +1099,8 @@ bodypart_id Character::body_window( const std::string &menu_header,
 
     bmenu.hilight_disabled = true;
     bool is_valid_choice = false;
+    int default_selection_idx = 0;
+    int default_selection_treatment_rank = 0;
 
     // If this is an NPC, the player is the one examining them and so the fact
     // that they can't self-diagnose effectively doesn't matter
@@ -1109,6 +1111,7 @@ bodypart_id Character::body_window( const std::string &menu_header,
         const bodypart_id &bp = e.bp;
         const int maximal_hp = get_part_hp_max( bp );
         const int current_hp = get_part_hp_cur( bp );
+        const float cur_hp_pcnt = current_hp / static_cast<float>( maximal_hp );
         // This will c_light_gray if the part does not have any effects cured by the item/effect
         // (e.g. it cures only bites, but the part does not have a bite effect)
         const nc_color state_col = display::limb_color( *this, bp, bleed > 0, bite > 0.0f, infect > 0.0f );
@@ -1118,6 +1121,12 @@ bodypart_id Character::body_window( const std::string &menu_header,
         // Broken means no HP can be restored, it requires surgical attention.
         const bool limb_is_broken = is_limb_broken( bp );
         const bool limb_is_mending = worn_with_flag( flag_SPLINT, bp );
+        // How much this treatment would help, if applied to this bodypart.
+        // The value itself is just a sorting number and has no actual meaning in itself, but is roughly ranged at:
+        // 0-30=bandage, 30-60=disinfectant, 60-70=bitten/infected, 70-80=bleeding
+        // A high value means that this bodypart should be prioritized to get this
+        // treatment, and is likely to be selected as default in the ui.
+        int treatment_rank = 0;
 
         if( show_all || has_curable_effect ) { // NOLINT(bugprone-branch-clone)
             e.allowed = true;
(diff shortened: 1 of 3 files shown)


PatrikLundell:
Stopping a bleeding wound is definitely more critical than improving the bandaging of an already bandaged body part in most cases (a dribble from a body part that is otherwise fine and will stop by itself soon might be an exception).

My priority list would be:
1. The wound bleeding the worst, assuming the treatment can staunch bleeding. That's the critical immediate threat.
2. If the treatment has disinfecting properties, select the worst infected wound. This is important and has to be done in a timely manner, but not within minutes: it can wait until bleeding has been taken care of.
3. Worst damaged wound that hasn't been bandaged. That's typically where your bandage would go if you don't have enough to treat everything, although you may elect to go for improving the bandaging of a serious injury over an untreated trivial one, but there is no way to accurately predict what the player will want in that case.
4. The wound with the worst bandage quality, starting with the worst wound if multiple at the same level, assuming the treatment can improve the bandage quality. When you're into replacing bandages you typically deal with all the injuries, so the order doesn't really matter. You may have a bandage shortage, though.
5. The top one in the list (you have to select something if there is nothing that can be achieved).

I'm not fond of a situation where the game seems to optimize the situation for you much of the time, but doesn't some of the time. That's worse than to always select the first entry in the list and completely leave it to the player to make the decisions, in my opinion.

inogenous:
> Stopping a bleeding wound is definitely more critical than improving the bandaging of an already bandaged body part

Good point, thanks.

Have updated the code now so that the priority is chosen as a weighted ranking in the order listed below (highest prio listed first):
1. Stopping bleeding if the treatment can do that. Heavy bleeding is ranked higher prio than minor bleeding.
2. Bites and infected wounds when applying disinfectant.
3. Disinfecting damaged body parts. Parts that have not been previously disinfected are ranked higher prio.
4. Bandaging damaged body parts. Parts that have not been previously bandaged are ranked higher prio.

Have tested with the following testcases, which all seem to make a better selection now:
* L-arm: `||||\` minor bleeding , head: `|||||` **heavy arterial bleeding** , L-leg: `||\..` no bleeding
    * applying bandage
    * head is selected as default
* torso: `||||\` **bitten** , L-leg: `||\..` , R-arm: `||||.`
    * applying bandage
    * l-leg is selected as default
* torso: `||||\` **bitten** , L-leg: `||\..` , R-arm: `||||.`
    * applying antiseptic
    * torso is selected as default
* R-arm: `||\..` bandaged (very poor), L-leg: `||\..` bandaged (very poor), head: `||||\` **not bandaged yet**
    * applying bandage
    * head is selected as default
* R-arm: `||\..` bandaged (**very poor**), L-leg: `||||.` bandaged (very poor), head: `||||\` bandaged (average)
    * applying bandage
    * R-arm is selected as default
* R-arm: `||\..` , l-leg: `||||.`
    * applying alcohol wipe
    * R-arm is selected as default
* R-arm: `||\..` disinfected (poor) , L-leg: `||||.`
    * applying alcohol wipe
    * L-leg is selected as default
* R-arm: `||\..` disinfected (average) , L-leg: `||||.` disinfected (**poor**)
    * applying antiseptic
    * L-leg is selected as default

PR #68278, opened 2023-09-23
Foot locker audit & item forms

Karol1223:
#### Summary
None

#### Purpose of change
Luizi on the devcord requested foot lockers be given the same treatment trash cans were given a while ago. So here I am, after questioning my sanity and reading Amazon listings for 2 hours.

#### Describe the solution

- [x] Better categorize item-furniture files
- [x] Add a new aluminum variant of the foot locker
- [x] Add item forms for both foot lockers
- [x] Give the foot lockers the take-down and activation treatment to deploy like braziers
- [x] Edit aluminum ingots into aluminum chunks, and make them actually usable around
- [x] Audit hinges
- [x] Give the items disassembly recipes
- [x] Handle furniture deconstruction with the flag and proper bash and deconstruction yields
- [x] Give the Exodii metal crates ``metal screeching`` sounds
- [x] Edit the furniture footlockers to use appropriate sounds, allow movement, no longer be ``ORGANIC`` and have lower coverage... as well as limiting their volume to be accurate
- [x] Make it so mapgen can spawn either of the two variants of foot lockers (applied everywhere in mainline save for Exodii bases, because they scare me)

The max volume for footlockers doesn't sit right with me. Ideally we'd handle this like vehicle cargo - when furniture gets filled, the items spill into terrain - but this doesn't work right now. When the footlocker gets full the items spill onto adjacent tiles, and the footlocker fills up VERY fast because it's tiny. But if I didn't give it this limit, the items would still be inside the footlocker, so it'd magically expand when deployed? Either way it's a loss

#### Testing
The items and furniture look to work as intended. Initially the furniture didn't work, but upon deciding against using ``copy-from`` and assigning values manually everything seems to be fine.


Karol1223:
@GuardianDll I'm gonna need you to take a look at my sorry attempts at mapgen to see if I understood you correctly

GuardianDll:
Mapgen part looks good for me

Karol1223:
Two random test failures on a required test in a row. Can we get a third?

PR #68290, opened 2023-09-24
Give the pipe some weapon techs

Aerin-of-the-Toast:
#### Summary
Balance "Give the pipe and makeshift crowbar better techs"

#### Purpose of change

The pipe (and makeshift crowbar, which is basically just a pipe) were very similar in size, weight, and balance to a cudgel, especially now that pipes are half as heavy as they used to be. They're even the exact same length at 60cm. Yet the pipe had no techniques, only medium block. Also yes, I know the pipe's been fiddled with a lot bare with me please

#### Describe the solution

Give the pipe (and makeshift crowbar) the same techniques as the cudgel; parry, precise strike, and rapid strike, since it's more than capable of being used deftly in much the same way as other batons (it's even literally in the baton weapon category).

Debated giving the pipe "good" balance similar to the cudgel, but since it is still a bit heavier (and the game docs specifically use the pipe as an example of neutral balance) I left it at neutral.

#### Testing

Game do load with the changes, all well and good.

#### Additional context

None


ashGlaw:
Cool change, makes sense. How sturdy are these pipes, though? With the reduction in mass they should probably be a bit more flimsy than they used to be. 🤷🏻‍♀️

data/json/items/resources/metal.json
```diff
@@ -12,7 +12,7 @@
     "symbol": "/",
     "material": [ "steel" ],
     "qualities": [ [ "HAMMER", 1 ] ],
-    "techniques": [ "WBLOCK_1" ],
+    "techniques": [ "WBLOCK_2", "RAPID", "PRECISE" ],
```
Maleclypse, line 15: I don't think anything that you swing over hand and doesn't have a sharp pointy end should have RAPID.  Like how are you gonna rapidly hit someone with a pipe or a crowbar?  Overhand swings aren't gonna be rapid and poking them with the non hand end is just gonna push them but not pierce them.  I also feel like WBLOCK_2 should require a hand guard but I'm way less convinced of that than I am the rapid statement lol

Aerin-of-the-Toast: The crux of this PR is giving the very baton-like pipe similar techs to the other batons (cudgel, collapsible, etc.) since i do believe it is more than close enough. 
It is quite possible that batons as a whole should have their techs changed, maybe Rapid and Parry really don't fit there. But that seemed to me like a separate PR and one that would require a lot more thought and balance consideration before I'm willing to commit to nerfing an entire weapon category like that. 

I'm not married to the idea of all batons innately having rapid strike (or even precise strike), as maybe it makes more sense for that to be handled by martial arts instead. I've seen batons used really fast in Eskrima drills for instance, but that's Eskrima; not a regular person who might just try to clobber people with them like a mace or stick, but i do think that's still a PR for another time examining the weapon class as a whole.

Zireael07: I think a pipe is not a baton. Yes it's similarly shaped, but I don't think it should have rapid strike.

Heck, moving it to martial arts makes more sense (so that a n00b swinging a baton won't get rapid, but a policeman who knows how to use it effectively does)

Maleclypse: Aerin and I spoke briefly as I was failing asleep at the keyboard, but I think that we agreed to merge this now and then hopefully one day someone who is not me will go back and make them all make more sense.

PR #68291, opened 2023-09-24
Make long stick long

Aerin-of-the-Toast:
#### Summary
Balance "Give long stick reach"

#### Purpose of change

Long stick is long. 2.6 meters long. Almost a full meter longer than most spears long. Yet it didn't have reach attack. (Also, both the stick and long stick still had manually-input to-hit values)

#### Describe the solution

Give the poor long stick REACH_ATTACK so you can bonk people from another tile away. Also updated the stick and long stick's to-hit values to the modern standard. Both have solid grip, any angle, and long reach. Stick has "neutral" balance and the long stick has "uneven" balance.

None

#### Testing

Game do indeed load and sticks are sticky.

#### Additional context

The to-hit changes are actually a decent buff; the regular stick is now at +1 while the long stick is at 0. Before, they were both at -1. They kinda needed the buff anyway due to their slow speed, though.

### DIFF
--- a/data/json/items/resources/wood.json
+++ b/data/json/items/resources/wood.json
@@ -48,7 +48,7 @@
     "weight": "1500 g",
     "volume": "2500 ml",
     "longest_side": "130 cm",
-    "to_hit": -1,
+    "to_hit": { "grip": "solid", "length": "long", "surface": "any", "balance": "neutral" },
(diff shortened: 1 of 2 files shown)


EvanBalster:
Bonk!

PR #68311, opened 2023-09-25
[Xedra Evolved] Rubik dialogue about Jotunn quads

Maleclypse:
#### Summary
None

#### Purpose of change
Someone mentioned that Rubik should talk about the heavy combat units that exist in XE.  So it will.  Discovered that any way I wanted to make this relevant it would require dialogue conditionals to check for achievements, so I'm attempting to make that work.

#### Describe the solution
Adds code for checking achievements in dialogue and then adds the dialogue.

None


src/avatar.cpp
```diff
@@ -295,7 +295,89 @@ void avatar::memorize_clear_decoration( const tripoint_abs_ms &p, std::string_vi
     player_map_memory->clear_tile_decoration( p, prefix );
 }
 
-std::vector<mission *> avatar::get_active_missions() const
+std::vector<achievement *> avatar::has_achievement() const
+{
+    std::vector<achievement *> ret;
+    for( auto &elem : achievements ) {
```
Maleclypse, line 301:
```suggestion
    for( auto &elem : achievement ) {
```

PR #68312, opened 2023-09-25
Skill vs Knowledge Check Audit

a-chancey:
#### Summary
Bugfixes "Audit of check_skill_level usage vs check_knowledge_level."

#### Purpose of change

Mending still required practical skill levels, which is inconsistent with the rest of the skill system. after discussing it on discord, i've moved the skill check to theoretical. Then I decided to just audit get_skill_level. I'm willing to revert anything with good reason. 

This should help cover many of the legitimate complaints about skill rust - the biggest being CBM mending requiring practical healthcare 4, which is fairly hard to get due to limited recipes at that difficulty.

#### Describe the solution

I moved the following actions to theoretical:

- Mending (character.cpp)
- autodoc bionic install (bionics.cpp)
- chance to damage gun on difficult gunmod install
- rot checks
- harvest result knowledge (foraging)
- NPC Carpenter Job result check (fabrication portion only)
- hacking success (time definition is ~~hard coded. I may refactor it, waiting for my build to complete to do some testing~~a function of practical - which makes sense when looking at the code.)
- Electrohack Success chances - time is hard coded and it's really like a brute force program being run, we'll just say it has a 5 minute runtime :D

The following are an *average* of theoretical and practical - being that you don't have the muscle memory yet, but you know what you need to do or to look out for. This is checked with a new method: Character::get_average_skill_level(skill_id). I left the output from get_knowledge_level as an integer so it only will increase on the full level knowledge gains - this can be changed, but the skills map will need changed to do so.

- trap disarming
- butchery/dissection success. Time spent could use a few proficiencies, and/or leverage off of the existing anatomy proficiencies 
- hotwiring success (time is still practical skill)

Things I'm really not open to taking suggestion about outside of the core dev team, as it would take one helluva argument to change my mind:
- Combat related skills - this is something you can't just read about, you need to know how to throw a punch by experience
- as an extension, martial arts techniques. for the exact same reason.
- Athletic related skills - these require physical training
- reload time - this is based on muscle memory
- lockpicking/safecracking: as someone who has done some lockpicking, all the knowledge in the world won't replace experience for lockpicking. plus they're both also governed by proficiencies, so I think they're in a good place.

I'm open to further suggestions, especially for other success rolls that would make sense to use the average instead of just practical. 

#### Testing

I started a new save and checked that i could perform each action with just theoretical skill. 

#### Additional context

Just wanted to audit the shift to theoretical, as i've seen (and personally had) a few complaints lately.


GuardianDll:
For butchery I'd prefer it to be replaced with one or multiple profs, but for the purpose of this PR I'd say average sound as a good way

a-chancey:
I just finished getting through every get_skill_level call and updated the main post with what I changed. Not a ton of changes, but a few nice QoL updates for activities that make sense to be based on theoretical.

I'm reviewing hacking, hotwiring and butchery/dissection now. For now, I'm leaning towards the success roll being based on theoretical and time based on practical. looking into that now.

I agree that butchery/dissection could definitely be moved to (or at least a bonus given by) proficiencies which would affect speed and success, but introducing those here would be out of scope.

a-chancey:
Butchery/dissection success has been changed to an average of theoretical and practical. The time to butcher, however, is all hardcoded with other multipliers in effect. I'm going to not touch those until we get a proficiency in place for either. even if we do, I don't think it should speed up *much* - maybe add a factor of .9 or something. Butchery is still a decent amount of physical work to do.

Looking at hotwiring next.

a-chancey:
Hotwiring success now moved to get_average_skill_level, but the time is still based on practical.

Oh, i added Character::get_average_skill_level(skill_id) which is (get_skill_level + get_knowledge_level)/2

get_knowledge_level does return an integer, so we don't get credit for partial theoretical levels, but i don't think that's too big of a deal.

when this build completes, i'll test everything i changed and undraft if everything works as expected.

GuardianDll:
Re: butchery profs, i suppose they should be the opposite - the default value should be increased, and all profs boost it to original speed
Hovewer this requires butchery be interruptabe, more similar to crafting, but none of this is in the scope of this pr

a-chancey:
Manual testing is completed. Should be ready for review.

a-chancey:
Moving this back to draft. 

The more I considered it, the more I disliked moving to just integer knowledge levels, as it would ignore the fractional skill if you're at theoretical 1.9. 

That was sloppy of me, glad I caught it before posting in ready to merge.

a-chancey:
Pushed the code while it compiles. I'll test it in the morning, assuming i didn't break anything, i'll move back out of draft. I may stay up til it's done but it's late.

I wound up adding not only the progress value where it would be used (mostly in averages and in autodoc cbm install success), but also added a check for greater of the returned skill value or the knowledge+progress value.

This is because if you have any enchantments boosting practical, they actually don't apply to theoretical, so that would never be considered without this check, making skill boosts for the affected activities irrelevant.

a-chancey:
Ok, i'm happy with testing, pulling it back out of draft

a-chancey:
Posted in ready to merge. I've been playing on it all day and testing it - manually made an artifact boost me to mend a CBM to test the greater of method. 

Should be pretty solid, bring more activities in line as expected and also takes a little of the edge off of skill rust in a few tricky places.

PR #68314, opened 2023-09-25
Trader avoid

detahramet:
#### Summary
Balance "Made traders not buy worthless scraps"

#### Purpose of change

Previously, tailoring scraps could be sold to traders despite their lack of usefulness and availability of their source fabric, which seems out of place.

#### Describe the solution

This adds two new meta-items, as `"copy-from"` sources for other fabric scraps. Previously scraps drew from `scrap_cotton`, which is now being made worthless for aforementioned reason. The two new meta-items are copies of `scrap_cotton`, with the only meaningful change being that one has the `TRADER-AVOID` flag. Worthless scraps now copy from the `scrap_worthless` meta item, while difficult to replace scraps, like Kevlar and Nomex, now copy from the `scrap_valuable` meta item.

One alternative was to add the `TRADER-AVOID` flag to each individual item, but this would have added unnecessary bulk to the json.

#### Testing

1. Started a new character with the default scenario and stats
2. Using the debug menu, gave player one of each new scrap item
3. Tried to sell scraps to the Evac Shelter NPC
4. Results matched expectation

#### Additional context

None.


Karol1223:
If the two new scraps aren't supposed to be items you should make them abstracts not items.

detahramet:
Converted the items to Abstracts, thanks Karol!

Issue #68307, opened 2023-09-25
Web Diver can bury the player underground.


EvanBalster:
### Describe the bug

After using the "Web Diver" mutation to descend multiple stories, the player phases through the floor, falling an additional story.  This can leave them trapped underground with no means of escape.

I probably introduced this regression in #68265 and discovered it while testing the follow-up #68306.  I intend to fix this bug myself.

### Attach save file

n/a; I created this bug and I'll create a tiny PR to fix it shortly.

### Steps to reproduce

1. Climb onto a second-story roof.
2. Use the debug menu to aid the Arachnid threshold and Web Diver mutations.
3. Try to walk off the building and pick the `w` option in the menu that appears.
4. Return to soil and atone for your crimes against nature.

### Expected behavior

Web diver should put you at ground level.

### Screenshots

Oh no.

![image](https://github.com/CleverRaven/Cataclysm-DDA/assets/13224671/54d21129-8603-41c3-9dab-fcb816c4a21e)

### Versions and configuration

Windows 10

### Additional context

I rewrote the function(s) where this happens so I understand them pretty well.

Maleclypse:
Closed by #68315

PR #68265, opened 2023-09-23
JSON-ify rules for climbing down & simplify UX

EvanBalster:
#### Summary
Infrastructure "Add JSON-based system for climbing aids."

#### Purpose of change

After working on #68120 (climbing risk) and #68150 (ledge bumping UX) I began working on a rebalancing contribution for climbing.  I started to realize that the current C++ implementation of vertical movement is a bit of a mess.  It involves a variety of functions located in different regions of the codebase:  most importantly `iexamine::ledge`, `map::climb_difficulty`, `game::slip_down`, `game::climb_down` and `game::vertical_move`.  The last two functions are a daunting nest of interacting conditions and special cases, and it is apparent that various gameplay features have been broken or inconsistently implemented as a result of this factoring.  I encountered similar difficulties trying to implement new features and balance changes.

The control flow in the current implementation also complicates user experience.  A player could potentially encounter up to four modal dialogs when interacting with a ledge (what to do, use grapple? use webs?, confirm climbing down).

This is a result of breaking a more multifaceted contribution [68190](#68190) into smaller pieces.  That earlier PR incorporates balance changes and a few gameplay additions in addition to the code changes in this contribution.

#### Describe the solution

* Introduce a new JSON type called `climbing_aid`, which represents a tool, ability or environmental feature supporting the player's body as they climb.
  * Overhaul `climb_down` to integrate this new type.
  * [Develop some JSON features](https://github.com/EvanBalster/Cataclysm-DDA/blob/eb-climbing-jsonify/doc/CLIMBING.md) to support future development & balancing of climbing mechanics.
    * `slip_chance_mod` is set to `-1000` for climbing aids that formerly bypassed the check.
    * See the original PR [68190](#68190) which experiments with 11 new climbing aid rules and balance changes.

JSON climbing aids are fully integrated with _climbing down_ mechanics.  Refactoring the code for climbing up will be a separate project.  The following mechanics have been reimplemented as climbing aids in [climbing.json](https://github.com/EvanBalster/Cataclysm-DDA/blob/eb-climbing-jsonify/data/json/climbing.json)

* Default  (ie, lowering your body from a ledge)
* Stepladder in the space below  (gameplay feature added in #68120)
* Grappling hook deployed from inventory
* Web Diver mutation (WEB_RAPPEL player trait)
* Debug tail (WALL_CLING player flag)

In addition, I made some UX improvements based on the above.

* The WALL_CLING mutation is now part of climbing down rather than 
[...]

PR #68315, opened 2023-09-25
Prevent Web Diver burying the player underground.

EvanBalster:
#### Summary
Bugfixes "Prevent Web Diver mutation from burying the player underground."

#### Purpose of change

This addresses issue #68307, most likely caused by PR #68265.  When using some climbing aid to descend multiple stories (currently only Web Diver does), the player performs one `vertical_move` per story.  Unfortunately each move can trigger a fall.  This means a player diving N stories will actually descend 2N-1 levels.

#### Describe the solution

`climb_down` now moves the player down as many stories as they could successfully climb, in a single call to `vertical_move`.  They'll only freefall if they couldn't climb the whole way down.

**Refactoring vertical_move**.  I plan to do this later but that's no reason to leave a mutation-wrecking bug in the code.

#### Testing

* Tested climbing down a two story building with Web Diver and step-through debugging.

#### Additional context

_Hope you brought some bugs to eat while you dig out._

### DIFF
--- a/src/game.cpp
+++ b/src/game.cpp
@@ -13448,14 +13448,15 @@ void game::climb_down_using( const tripoint &examp, climbing_aid_id aid_id, bool
             // The player has slipped and probably fallen.
             return;
         } else {
-            g->vertical_move( -1, true );
-            add_msg_debug( debugmode::DF_IEXAMINE, "Safe movement down one Z-level" );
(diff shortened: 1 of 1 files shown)


EvanBalster:
All test failures are due to the FLAC dependency problem introduced yesterday.  This one should be ready to go.

Issue #66442, opened 2023-06-25
Nuts and bolts and welding rods upgrading


KHeket:
### Describe the bug

After an update, welding work with car costs enormous welding rods and nut and bolts, but amount of it in world and refuge center is extremely low and expensive
And also, nut and bolt are needed for lots of cat parts installations, and after removing this part - nuts and bolts are disappearing, when in reality you can remove it without destruing
And also I didnt fing a way to craft nut and bolts by myself

### Attach save file

N/A

### Steps to reproduce

1) Launch the game
2) Find/spawn a car
3) Spawn resources and part for car. Install/remove it, see the problem

### Expected behavior

More welding rods and materials, nuts and bolts, less expensive and bigger amout at refuge center, recipes for nuts and bolts

### Versions and configuration

n/a

TheShadowFerret:
>And also I didn't find a way to craft nut and bolts by myself

3.a. of #66256 will add the necessary tools for this

SolatK:
i can confirm disassembling car parts didnt give you nuts and bolts

Andrezinho1:
i agree that disassembly should get me nuts and bolts i took apart a completely pristine car with a bad engine i agree that i shouldn't be able to just attach things with no welding but the bolts should be a given if you wanted to balance it make it so you lose some based on the condition of the part the less structural integrity the more stripped or broken bolts from high impact violence

KHeket:
It seems, https://github.com/CleverRaven/Cataclysm-DDA/pull/66256 was abadoned or @randomtyper is too busy. We need at least just recipe of nuts and bols

KHeket:
Now the https://github.com/CleverRaven/Cataclysm-DDA/pull/66256 is closed, so this issue is actual

RobertoRobot:
My idea is to have scrap yard merchants steadily sell bolts and nuts, which is a simple temporary solution.

PR #68320, opened 2023-09-25
The scrap trader sells nuts and bolts

RobertoRobot:
#### Summary
Balance "the scrap trader sells nuts and bolts"

#### Purpose of change
#64868 adds the consumption of bolts and nuts for vehicle construction and repair. Although #66557 increased the spawn number of bolts and nuts, it did not completely solve the problem of the shortage of bolts and nuts in the game, because there was still no stable way to obtain bolts and nuts.

#### Describe the solution
The scrap trader sells bolts and nuts, which is a stable way to obtain bolts and nuts.

1. Add the recipe for making bolts and nuts, which is inefficient if based on reality.
2. Make nuts and bolts recoverable from vehicle parts, which requires C++.

#### Testing
Load the changes into the game as a mod and the results are as expected.

#### Additional context
This is a simple temporary solution, as other alternatives are a little difficult for me.

### DIFF
--- a/data/json/npcs/scrap_trader/scrap_trader.json
+++ b/data/json/npcs/scrap_trader/scrap_trader.json
@@ -71,6 +71,7 @@
     "shopkeeper_consumption_rates": "basic_shop_rates",
     "shopkeeper_price_rules": [
       { "item": "scrap", "price": 200, "fixed_adj": 0 },
+      { "item": "nuts_bolts", "price": 500, "fixed_adj": 0 },
       { "item": "sheet_metal_small", "price": 400, "fixed_adj": 0 },
       { "item": "steel_chunk", "price": 400, "fixed_adj": 0 },
       { "item": "steel_lump", "price": 600, "fixed_adj": 0 },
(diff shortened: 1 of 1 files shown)


KHeket:
My main problem in the linked issue - that nut and bolt is unrecoverable from dismantling vehicle parts and we dont have recipe for craft it - so can you edit discription of your pr, because when you write "Fixes https://github.com/CleverRaven/Cataclysm-DDA/issues/66442" so after merging this pr mine issue will be closed, but this pr dont solves the main problem with craft recipe and bolt and nuts from dismantling

a-chancey:
> ...bolt and nuts from dismantling

This has been a pet peeve of mine for a while, I may have to try to fix it now that I took care of the other issue bothering me. 

Vehicles only return the primary part when a part is removed (or scrap from a xx destroyed part). This has been the case forever - I noticed it when I wasn't getting ropes back from my tied mattresses after towing a car. We're not shearing off every bolt with a dremel, we use a wrench. We should get bolts back. 

In fact, it's why we have a tied fridge and freezer item that then gets installed on top of the vehicle rather than using the disconnected appliance and rope.

PR #68321, opened 2023-09-25
Implement Character::rest_quality()

RenechCDDA:
#### Summary
Balance "Your awake healing speed now varies with how much(or little) you are resting"

#### Purpose of change
Character::rest_quality() is one big TODO

#### Describe the solution
Implement!

Primarily your rest quality is determined by your activity level. MODERATE_EXERCISE (*walking* around, lots of crafting activities) provides a flat 0, the prior equivalent to simply being awake. Characters that are laying on a bed awake, doing literally nothing can get up to 0.5f. Sleep gives a maximum of 1.0f (same as prior), but is now modified by the quality of your sleeping place's `floor_bedding_warmth` (with crappier beds giving lower returns, linearly).

Performing strenuous activities (BRISK_EXERCISE or higher) gives an increasingly greater malus, up to -0.6f during EXTRA_EXERCISE. Digging ditches while your wounds heal is not recommended!

All of the above are modified by by heat and cold penalties (it is hard to rest well if you're shivering and rolling over every 5 minutes!), which are (morale penalty)/10. These do not have a very large impact, as at -10cold/hot morale you will be dying of frostbite or melting already.

Some sort of caching like the activity tracker does? This function is called somewhat infrequently so someone could - in theory - game the system by sitting perfectly still on their bed every 5 minutes or every so often for exactly 1 turn. The current implementation only cares about your status *at the moment it checks*, not in the time period between checks.

#### Testing
Threw some debugmsgs into the code, ran around in a variety of conditions and made sure the numbers looked right.

Sat in some furniture seats for a while, got confirmation that seats were modifying the rest value. Ditto for vehicles.

#### Additional context

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -362,6 +362,8 @@ static const mon_flag_str_id mon_flag_LOUDMOVES( "LOUDMOVES" );
 static const mon_flag_str_id mon_flag_MECH_RECON_VISION( "MECH_RECON_VISION" );
 static const mon_flag_str_id mon_flag_RIDEABLE_MECH( "RIDEABLE_MECH" );
 
+static const morale_type morale_cold( "morale_cold" );
+static const morale_type morale_hot( "morale_hot" );
 static const morale_type morale_nightmare( "morale_nightmare" );
 
 static const move_mode_id move_mode_prone( "prone" );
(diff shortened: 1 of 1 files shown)


Maleclypse:
I'm surprised that this isn't impacting any tests on how fast the PC recovers from exhaustion?

RenechCDDA:
> I'm surprised that this isn't impacting any tests on how fast the PC recovers from exhaustion?

Aha, the healing test uses hardcoded values. I was wondering that actually! I'll have to update that.

PR #68324, opened 2023-09-25
Fix identical omts spawning over themselves

### LINKED ISSUE #67837: Towns have too many zombies and wrecked vehicles
### Describe the bug

Literally in every new town I go to there are a massive amount of zombies and wrecked vehicles. This makes the game completely unplayable. Vanilla and default world options.

### Attach save file

[Chatom-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12418678/Chatom-trimmed.tar.gz)

### Steps to reproduce

-Load the save file provided above
-Turn on quick setup (for convenience)
-Just visit any unexplored town

### Expected behavior

According to discord, there shouldn't be this many zombies. I concur.

### Screenshots

20230823_122733_1920_1057_cataclysm-tiles_7cGwqwkYWV
20230823_122813_1920_1080_cataclysm-tiles_65yTBDTequ

- OS: Windows
    - OS Version: 10.0.22621.2134 (22H2)
- Game Version: f666af0 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Translate Complex Dialogue [translate_dialogue]
]

### Additional context

_No response_

akrieger:
#### Summary
Bugfixes "Fix overload of zeds and vehicles with older saves"

#### Purpose of change
Possibly fixes #67837. One fix led to another bug, which leads to this fix. There's a migration safety issue in mapgen where old saves serialized 'bad data' that was resolved at runtime. The runtime fixup was removed, but the saves were not migrated. This is an attempt to implement that save migration.

#### Describe the solution
I don't really know the exact terminology, but the laymans answer is things like roads were running mapgen 'over themselves' repeatedly. This resulted in things like dozen car pileups and hundreds of zeds in cities. The patch that introduced this (#67773) implemented logic at worldgen? mapgen? time which prevents these redundant roads being added to a 'predecessors' list. I ported that logic into the savegame deserializer to try to handle this at load time so older saves migrate forward successfully. The only thing I'm unsure of is the base condition of zero to 1 predecessors, but I'm assuming that *any* predecessor should be allowed compared to a base case of I'm guessing null terrain?

#### Testing
Created a save in a game compiled just prior to #67773. Debug map revealed and teleported to a city extremely far away. No problem.
Loaded that save on ##67773. Teleported to the same tile. Hundreds of zeds.
before

Applied my fix. Teleported to the tile. Not hundreds of zeds.

after

Roads don't have infinite predecessors anymore. A before comparison:

current_bad

Created a new world. Inspected 4 way intersections in cities. Saw there are still some that have a manhole omt as a predecessor. Manholes appear to not be linear, and adding the LINEAR flag is apparently mutually exclusive with NO_ROTATE, so that's not going to change. I don't know if this means city intersections have a higher chance of double spawns of somethings or not.
after

#### Additional context

0.G didn't have this redundant manhole under crossroad stuff. But JSONizing roads is radically different than what was before. No clue how to get there.

og_manhole

### DIFF
--- a/src/overmap.cpp
+++ b/src/overmap.cpp
@@ -2889,10 +2889,16 @@ void overmap::ter_set( const tripoint_om_omt &p, const oter_id &id )
 
     oter_id &val = layer[p.z() + OVERMAP_DEPTH].terrain[p.xy()];
     if( id->has_flag( oter_flags::requires_predecessor ) ) {
-        const oter_type_id id_type = id->get_type_id();
         // Stops linear fallback_predecessor maps (roads etc) spawning over themselves
-        if( !( id_type->is_linear() && id_type == val->get_type_id() ) ) {
-            predecessors_[p].push_back( val );
+        std::vector<oter_id> &om_predecessors = predecessors_[p];
+        if( om_predecessors.empty() || !val->is_linear() ) {
+            om_predecessors.push_back( val );
+        } else {
+            // Collapse and keep the last linear oter of matching type
+            oter_id &last_predecessor = om_predecessors.back();
+            if( last_predecessor->get_type_id() == val->get_type_id() ) {
+                last_predecessor = val;
+            }
         }
     }
     val = id;
--- a/src/savegame.cpp
+++ b/src/savegame.cpp
@@ -657,8 +657,30 @@ void overmap::unserialize( const JsonObject &jsobj )
         } else if( name == "predecessors" ) {
             std::vector<std::pair<tripoint_om_omt, std::vector<oter_id>>> flattened_predecessors;
             om_member.read( flattened_predecessors, true );
+            std::vector<oter_id> deduped_predecessors;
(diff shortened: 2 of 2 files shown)


akrieger:
I'm not sure this is a perfect fix. There's still larger pileups happening than should be expected. I'm not sure how far back to go to get to 'true legacy code' to compare against.

akrieger:
It looks like the behavior is intentional if I'm reading #67223 right, it just seems like spawn rates on roads in town are still elevated. I wonder if just removing REQUIRES_PREDECESSOR from the road json definition would resolve it, but I don't understand the implications of doing that. I will try it though.

akrieger:
I think I got it. I think the right answer is to not push a linear predecessor if the incoming type is itself linear, unless predecessors is empty.

akrieger:
Well I think this approach is plausible. What do people think?

PR #68335, opened 2023-09-26
Make vehicle parts install components salvageable, e.g. nuts and bolts

### LINKED ISSUE #66442: Nuts and bolts and welding rods upgrading

lispcoc:
#### Summary
Features "Make vehicle parts install components salvageable, e.g. nuts and bolts"

#### Purpose of change
Partially resolve #66442.

#### Describe the solution
When you uninstall vehicle parts, the components used during installation can be recovered from the parts installed by the player. Default components can be salvaged from existing parts.

The recovery rate of components from damaged parts, as in the case of disassembly, will be:
`rate  = 0.8 ^ damage_level`

#### Testing
nuts and bolts are salvaged from parts player installed.
nuts and bolts are salvaged from parts existing.
The amount of nuts and bolts salvaged from damaged parts is reduced.

Test salvageable/unsalvageable components
salvageable:
- nuts and bolts
- nail (Is it correct that nails are reusable?)

unsalvageable:
- superglue
- duct tape
- welding wire

#### Additional context

image


GuardianDll:
Iirc what we want to is vehicle part installation work with requirements, which would allow player to use either weld, bolts or tape, alternatively, depending on the vehicle part
Then we could, in theory, store this info as some metadata, and disassemble could return the related resources, but i don't know do we have such metadata at all
Approach of this PR is completely fine tho
I think lower salvage from damaged parts is reasonable

ehughsbaird (approved):
Removing game-installed parts yields salvageables. Inappropriate salvages (welding wire, duct tape) is not yielded when installed manually or by the game and the uninstalled.

I don't know if this will apply for vehicle in saves made before this, but I'm not sure how easy that is to fix.

lispcoc:
> I don't know if this will apply for vehicle in saves made before this, but I'm not sure how easy that is to fix.

I did a little checking, but it didn't seem to be a problem.

And I discovered a glitch that allows you to multiply nuts and bolts infinitely by building new vehicles, so I would like to fix it.

lispcoc:
I think it's ready to merge.

PR #68337, opened 2023-09-26
Also add libflac-dev to matrix for cmake build

akrieger:
#### Summary
None

#### Purpose of change
General matrix GCC9 build uses cmake and is also broken from the flac dependency being missing, like clang-tidy. Just no one noticed because clang-tidy fails fast and this one fails last.

#### Describe the solution
Add the same package to the apt install call for matrix builds.

#### Testing
CI will get to it... eventually...

### DIFF
--- a/.github/workflows/matrix.yml
+++ b/.github/workflows/matrix.yml
@@ -255,7 +255,7 @@ jobs:
     - name: install SDL2 dependencies (ubuntu)
       if: ${{ env.SKIP == 'false' && runner.os == 'Linux' && matrix.tiles == 1 }}
       run: |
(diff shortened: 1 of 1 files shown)


akrieger:
I should also add this to the dev setup docs, unless @alef can make this optional instead. What was the error seen that caused this to become required?

alef:
> What was the error seen that caused this to become required?

The linker complained the `FLAC` library was missing when linking `cataclysm-tiles`. I do not have the exact message, but one can inspect private headers of `/usr/lib/x86_64-linux-gnu/libSDL2_mixer-2.0.so` and find the entry `libFLAC.so.8` is flagged as `NEEDED` in the dynamic section.

akrieger:
I wonder if that's just an issue with the devcontainer install of sdl-mixer then. The gcc9 cmake build was building with sound enabled and was linking fine.

PR #68340, opened 2023-09-26
Underwater monsters don't attack by passing through boats

### LINKED ISSUE #66983: Monsters that are able to swim, can attack you though your boat hull and frame, over Z levels, without actually damaging it first.
### Describe the bug

If your driving on your boat on SEA water, which do has multiple Z levels of water below, than monsters which can swim like these zombie sharks are able to directly attack you from below your boat though your boat hull and boat frame, without damaging it in any way. The attack just gets teleportet though the boat over Z levels in SEA water. 
This happens when you spot these sharks (can also randomly instantly happen) and then wait a long while, until they can actually just submerge in the water and try to attack you from below your boat.

### Attach save file

Kieferschreck 1
Kieferschreck 2
Kieferschreck 3
Kieferschreck 4
Kieferschreck 5

### Steps to reproduce

1. Spawn a at least 3*3 boat on deep sea water
2. Spawn a monster than can swim next to the boat
3. wait long enough
4. See that the monster has submerged in the water and attacks you from below your boat over Z levels.

### Expected behavior

Normal monsters should not be able to teleport there attacks though the boat hull and frame, over Z levels.
There attacks should first damage and destroy the boat hull at least, befor they can manage to attack you though it.

Or just do not let them magicaly attack you though boat hulls.

### Screenshots

_No response_

Windows 10
version: d071694 (tiles)
build: 2023-07-16-0559 (2 days ago)
mods: Dark days ahead

### Additional context

_No response_

marimarigi:
#### Summary
Bugfixes "Underwater monsters attack by passing through boats"

#### Purpose of change
Fix #66983

#### Describe the solution
They use "stairs" in lakes to attack.
So I removed the "GOES_UP" and "GOES_DOWN" flags for lakes, and compensated for the loss of functionality with the "NO_FLOOR_WATER" flag and code changes.

Leave "stairs" as is and add code for underwater.

#### Testing
Compiled it, confirmed the following:
- Spawns hostile monsters under the boat and will not attack you even if you wait.
- Can be moved by boats.
- Can dive to the bottom of the lake and resurface.
- Floating devices and fins work.
- Underwater monsters move up and down.
- Can fight underwater monsters.

### DIFF
--- a/data/json/furniture_and_terrain/terrain-liquids.json
+++ b/data/json/furniture_and_terrain/terrain-liquids.json
@@ -37,7 +37,7 @@
     "looks_like": "t_water_sh",
     "color": "blue",
     "move_cost": 8,
-    "flags": [ "TRANSPARENT", "LIQUID", "NO_SCENT", "SWIMMABLE", "DEEP_WATER", "FISHABLE", "GOES_DOWN" ],
+    "flags": [ "TRANSPARENT", "LIQUID", "NO_SCENT", "SWIMMABLE", "DEEP_WATER", "FISHABLE", "NO_FLOOR_WATER" ],
     "connect_groups": "WATER",
     "connects_to": "WATER",
     "examine_action": "water_source"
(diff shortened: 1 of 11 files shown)


Vollch:
> it's not possible to go down from 3x3 hole but only go up

Well, yeah. Diving mechanic and freshwater mapgens doesn't perfectly match each others. I'm not decided yet what would be the best way to fix it.
It might be fixed by adding GOES_DOWN flag to pools. Or changing pool's terrain to regular deep water. Or reworking checks in diving code so it'll work for any "water cube -> deep water", "deep water -> water cube", "water cube -> water cube" shifts, withoout looking at GOES_* flags.
Or maybe it's worth to port this thing from DDA, which along the way introduces *another* diving related flag: https://github.com/CleverRaven/Cataclysm-DDA/pull/68340

PR #67824, opened 2023-08-22
Correct inaccurate Kickboxing moves, let you start with it.

Aerin-of-the-Toast:
#### Summary
Balance "Adjust Kickboxing and let you start with it"

#### Purpose of change

Far as I could tell, there wasn't a way to start with kickboxing at all. As for the art itself, it included two knee strikes despite kickboxing not using knee or elbow strikes _(Muay Thai Kickboxing_ does, but we already have Muay Thai), while missing some foundational moves.

Obsoletes #60649

#### Describe the solution

Added Kickboxing as a possible choice for the "Self Defense" background, and another lower-tier "Cardio Kickboxing" background as it is a popular fitness thing.

Removed the two knee strikes. Also removed the side kick, as it was redundant with the roundhouse kick and while it _is_ a technique in kickboxing it's not as foundational. Added the "Cross" technique (basically a copy of Pankration's cross) and "Push Kick" technique (basically a copy of Taekwondo's side kick).

Slightly buffed the roundhouse kick from 1.2x to 1.4x damage to separate it from Cross (higher base damage but there's not really any good foot weapons compared to fist weapons) and made it crit_ok so there was a usable crit technique on enemies that can't be uppercut.

#### Describe alternatives you've considered

Leaving the side kick in mostly for flavor, or trying to figure out how to differentiate it from the roundhouse/push kick.
Increased the skills in Cardio Kickboxing; maybe getting actual instruction should be 2 melee/unarmed even if all you're doing is punching a bag and 1 is undershooting it.

#### Testing

Checked to make sure the backgrounds spawned, kickboxing was chooseable, and the techniques were reflected in game.

#### Additional context

I made this PR after coming home from my own cardio kickboxing class. It was leg day. orz

PR #68342, opened 2023-09-26
Muay Thai additions

waxcatape:
#### Summary
Content "Add 3 fundamental techs to Muay Thai, slight rebalances for knee techs in Kickboxing and Muay Thai for balance and consistency."

#### Purpose of change

Muay Thai right now doesn't bear much resemblance to the real thing, with many bread and butter techniques completely absent: there's not a single boxing technique, the only knee tech is not even the most frequent and fundamental one, no teeps. The Art of 8 Limbs right now is "The Art of 6 Limbs, and 2 of those can only be used while jumping, lol", in-game Kickboxing is closer to Muay Thai than in-game Muay Thai.

#### Describe the solution

Added a Rapid, a Front Kick for distance control (Teeps are used all the time in Muay Thai and present good utility in-game) and a Straight Knee. Brought the damage of both the Kickboxing Straight Knee and the Muay Thai Straight Knee down to 1.25 (it was otherwise identical to the Flying Knee except it unlocked at level 3 instead of level 5). Added a movecost malus of 1.15 to both Kickboxing and Muay Thai Flying Knee (it's a fairly committal move). Renamed the Power Kick to Roundhouse Kick and brought the damage bonus from 1.3 to 1.2 to be consistent with other Roundhouse Kick techs. Added a +2 flat cutting bonus to Elbow Strike (elbow strikes often cause lacerations).

Adding more boxing techniques, trips and dumps/throws. Those would all be more than appropriate (especially at least 1 more hand strike), but the art is 7 techs already, making it the second highest number of techniques ingame right now after Brawling. 

Make Elbow Strike and Straight Knee fire off on on both normals and crits, but for normals to happen only when Grabbed. I'm not sure how to do that, maybe sometime later.

#### Testing

Spawned, tested that new techniques fire off, everything's as intended.

### DIFF
--- a/data/json/martialarts.json
+++ b/data/json/martialarts.json
@@ -912,6 +912,7 @@
       "tec_muay_thai_rapid",
       "tec_muay_thai_elbow",
       "tec_muay_thai_roundhouse_kick",
+      "tec_muay_thai_front_kick",
       "tec_muay_thai_straight_knee",
       "tec_muay_thai_flying_knee",
       "tec_muay_thai_break" ]
--- a/data/json/martialarts.json
+++ b/data/json/martialarts.json
@@ -908,7 +908,13 @@
         ]
       }
     ],
-    "techniques": [ "tec_muay_thai_elbow", "tec_muay_thai_kick", "tec_muay_thai_knee", "tec_muay_thai_break" ]
+    "techniques": [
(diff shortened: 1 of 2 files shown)


waxcatape:
Didn't even notice #67824, will correct and reopen if necessary

PR #68347, opened 2023-09-26
Muay Thai changes

waxcatape:
#### Summary
Content "Add 3 fundamental techs to Muay Thai, slight tweaks for balance and consistency."

#### Purpose of change

Muay Thai right now doesn't bear much resemblance to the real thing, with many bread and butter techniques completely absent: there's not a single boxing technique, the only knee tech is not even an abstract knee, but a very specific technique, no teeps. The Art of 8 Limbs right now is "The Art of 6 Limbs, and 2 of those can only be used while jumping, lol".

#### Describe the solution

Added a Rapid, a Cross and a Push Kick for distance control (Teeps are used all the time in Muay Thai and present good utility in-game). Renamed the Power Kick to Roundhouse Kick and bumped the damage bonus up from 1.3 to 1.4 to be consistent with the Kickboxing Roundhouse Kick tech. Added a +2 flat cutting bonus to Elbow Strike (elbow strikes often cause lacerations), made the Elbow Strike slightly slower (0.5 -> 0.66) to offset the damage bonus somewhat. Renamed Flying Knee to Knee Strike. Removed non-stunned and non-zombie requirements for the Knee Strike (how does a stunned enemy present a harder target for a knee strike? How does a similar size zombie present a harder target than a person?).

Adding uppercuts, trips and dumps/throws. Those would all be more than appropriate, but the MA is 7 techs already, making it the second highest number of techniques ingame right now after Brawling.

Make Elbow Strike and Knee Strike fire off on both normals and crits, but for normals to happen only when Grabbed. I'm not sure how to do that, maybe sometime later.

#### Testing

Spawned, tested that new techniques fire off, everything's as intended.

### DIFF
--- a/data/json/martialarts.json
+++ b/data/json/martialarts.json
@@ -909,15 +909,15 @@
         ]
       }
     ],
-      "techniques": [
-          "tec_muay_thai_rapid",
-          "tec_muay_thai_cross",
-          "tec_muay_thai_elbow",
-          "tec_muay_thai_push",
-          "tec_muay_thai_roundhouse",
-          "tec_muay_thai_knee",
-          "tec_muay_thai_break"
-      ]
+    "techniques": [
+      "tec_muay_thai_rapid",
+      "tec_muay_thai_cross",
+      "tec_muay_thai_elbow",
+      "tec_muay_thai_push",
+      "tec_muay_thai_roundhouse",
+      "tec_muay_thai_knee",
+      "tec_muay_thai_break"
+    ]
   },
   {
     "type": "martial_art",
--- a/data/json/martialarts.json
+++ b/data/json/martialarts.json
@@ -909,7 +909,15 @@
         ]
       }
     ],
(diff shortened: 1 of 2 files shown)


fairyarmadillo:
I don't think those knee strike changes are going to fly. @Venera3 specifically made it so that stun techniques do not work on zombies.

The restrictions on flying knee are intentional and standard. Stun techs only work on non-stunned non-zombie enemies across the board.

waxcatape:
> I don't think those knee strike changes are going to fly. @Venera3 specifically made it so that stun techniques do not work on zombies.
> 
> The restrictions on flying knee are intentional and standard. Stun techs only work on non-stunned non-zombie enemies across the board.

I see, I wasn't aware of that. Ideally I'd like for knees to still land on zombies and stunned enemies, just without applying stun, but I'm not sure how to do that. It would be a slight debuff to the style overall, because you have a suboptimal tech (without the stun) competing for a crit, but you'd see one of the more recognizable strikes for the style way more often. Unfortunately, the way it's set up right now, the tech basically only works on npcs, and I'm not sure how many players would ever want to unarmed melee them lol. Well, not a huge loss.

PR #68349, opened 2023-09-26
Fix Cloak of Frog permanently frogging you

Standing-Storm:
#### Summary
Bugfixes "Fix Cloak of Frog permanently frogging you"

#### Purpose of change

Someone on Discord said they were permanently frogged and they're right--the EoC to lose the frog mutations actually gives them to you again.

#### Describe the solution

Change u_add_trait to u_lose_trait


Vane79:
Oh wow, I thought it was like a curse for reading unknown scrolls.

Standing-Storm:
> Oh wow, I thought it was like a curse for reading unknown scrolls.

Something like that could be interesting to put in but there'd have to be a way to uncurse yourself included too (which I didn't include for Cloak of Frog)

PR #68355, opened 2023-09-27
Change climbing warning display to properly show color

wrnode:
#### Summary
None

#### Purpose of change
Minor change to game.cpp to fix color formatting in #68265 
As is, the color tag is not properly closed and breaks the rest of the graphic.

#### Describe the solution
One character change to properly display color

Leave as is

#### Testing
Tested on a save without the fix, rebuilt with change, tested again.

#### Additional context
Before fix
cataclysm-tiles_FbWaBBALiw
After fix
cddafixed


Night-Pryanik:
Could you please add an empty line after all hints and before `Climb down?` query? I feel an empty separation line would help to visual clarity.

PR #68359, opened 2023-09-27
Deconstructing a water heater yields a 100L tank instead of 60L

### LINKED ISSUE #68316: larger tank for water heaters

Deconstructing a household water heater gives you one `metal_tank` and some other bits and bobs. However, the volume of a `metal_tank` is only 60L, while the water heater in your house probably holds 200–400L (or 50–100 gallons, in MA).

### Solution you would like.

Add a `large_metal_tank` item that holds 200L. And weighs a lot.

_No response_

### Additional context

The water heater object has a volume of 240L, making it even sillier that you only get a 60L tank out of it.

The `f_water_heater` furniture item has a `keg_capacity` of 240L.

https://www.homedepot.com/b/Plumbing-Water-Heaters/5/Gas/N-5yc1vZbqlyZ1z109srZ1z1py50/Ntk-elasticplus/Ntt-water%2Bheater?NCNI-5&sortby=bestmatch&sortorder=none

db48x:
#### Summary
Balance "Deconstructing a water heater yields a 100L tank instead of 60L (plus the usual bits and bobs)."

#### Purpose of change

Fixes #68316

#### Describe the solution

I did a quick search for water heaters on the website of the nearest home improvement store and found that the Rheem model XE36S06ST45U0 is a good match for the stats of the existing household_water_heater item at 240L exterior volume and 54kg mass. See
<https://images.thdstatic.com/catalog/pdfImages/2d/2d7ed116-1a8c-439d-b7fe-a62a0a98f806.pdf> for published dimensions.

As this water heater has an internal capacity of 36 US gallons (136L), I changed the deconstruction recipe to yield a 100L tank instead of the 60L it used to be. I also changed the keg_capacity of the heater to match.

I originally intended to use the 200L `55gal_drum` for this change, primarily because I myself have a 50–gallon heater. However, I noticed that this didn’t leave a lot of room for insulation, and went looking for specifications on real water heaters.

Of course, this size of water heater is intended for use in small apartments; larger houses would have larger heaters (up to 100 gallons even). I’m resisting the temptation to add another size or two just on general principle, but I note that mansions do exist.

#### Testing

I tested this by disassembling a water heater and noting that I got a 100L tank as intended. Later I noticed that the keg_capacity was off; turns out that it's in cups not liters.

#### Additional context

Screenshot 2023-09-27 at 15-54-19 2d7ed116-1a8c-439d-b7fe-a62a0a98f806 pdf

Screenshot from 2023-09-27 15-51-58

### DIFF
--- a/data/json/furniture_and_terrain/furniture-plumbing.json
+++ b/data/json/furniture_and_terrain/furniture-plumbing.json
@@ -114,7 +114,7 @@
     "flags": [ "CONTAINER", "PLACE_ITEM", "LIQUIDCONT", "NOITEM", "SEALED", "EASY_DECONSTRUCT", "SMALL_HIDE" ],
     "//": "keg_capacity assumed to be from model XE36S06ST45U0 water heater <https://images.thdstatic.com/catalog/pdfImages/2d/2d7ed116-1a8c-439d-b7fe-a62a0a98f806.pdf>",
     "examine_action": "keg",
-    "keg_capacity": 105,
+    "keg_capacity": 400,
     "deconstruct": { "items": [ { "item": "household_water_heater", "count": 1 } ] },
     "bash": {
       "str_min": 18,
--- a/data/json/furniture_and_terrain/furniture-plumbing.json
+++ b/data/json/furniture_and_terrain/furniture-plumbing.json
@@ -112,8 +112,9 @@
     "coverage": 55,
     "required_str": -1,
     "flags": [ "CONTAINER", "PLACE_ITEM", "LIQUIDCONT", "NOITEM", "SEALED", "EASY_DECONSTRUCT", "SMALL_HIDE" ],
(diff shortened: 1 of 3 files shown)


GuardianDll:
\>  I’m resisting the temptation to add another size or two

Just add it, i don't see any reason to not

db48x:
To do that well I would want to update the mapgen for the mansion and perhaps some of the larger houses, but that’s a lot more than I want to do tonight. :)

GuardianDll:
Sure. Fyi mapgen allows to define variants of the same terrain/furniture, so for most places you can make simple replacement `"@": "f_water_heater"` -> `"@": [ [ "f_water_heater", 7 ], [ "f_water_heater_big", 3] ]` so the game would generate 70% of default water heaters, and 30% of big water heaters, or something similar

db48x:
Yes, I did something similar for a different patch I’m working on (unrelated to water heaters).

PR #68360, opened 2023-09-27
Improved where monsters leap

marimarigi:
#### Summary
None

#### Purpose of change
Monsters with leap and understand of dangers will leap into tiles they wouldn't normally move, such as air or fire. Mainly to improve it.

#### Describe the solution
Check the tiles to leap by using "monster::will_move_to" which manages the tiles to avoid during normal movement.
Also, large and huge monsters can no longer leap small passage such as windows.
**EDIT:**
Added two fields to do similar leaps as before.
"ignore_dest_terrain" is for leaps such as when aquatic monsters land.
"ignore_dest_danger" is for leaps such as when wild animals into fire.

#### Testing
I made zombies and wild animals leap, confirmed that the destination of the leap is the same as normal movement.

### DIFF
--- a/doc/MONSTER_SPECIAL_ATTACKS.md
+++ b/doc/MONSTER_SPECIAL_ATTACKS.md
@@ -300,6 +300,8 @@ Makes the monster leap a few tiles over passable terrain as long as it can see i
 | `min_range`             | (Required) Float, minimal range of the jump.  Respects circular distance setting!                    |
 | `prefer_leap`           | Leap even when adjacent to target, will still choose the closest acceptable destination.             |
 | `random_leap`           | Disregard target location entirely when leaping, leading to completely random jumps.                 |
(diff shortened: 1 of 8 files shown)


ashGlaw:
Any way to let them leap up or down zlevels? Make roofs less safe?

Karol1223:
This is going to be the most stupid request I've made so far but could you do me a favor and check how it affects aquatic monsters leaping onto land? It worked before, although they still couldn't move on land. It allowed me to start #67291

marimarigi:
@ashGlaw It would be possible if someone could extend leap's path calculation to 3D. (It's difficult for me.)

@Karol1223 Aquatic monsters cannot leap into lands. I will consider countermeasures.

marimarigi:
Added an option for aquatic monsters to leap onto land.

PR #68377, opened 2023-09-28
Cleanup crossbow bolts

GuardianDll:
#### Summary
None
#### Purpose of change
Followup to #68029
#### Describe the solution
Almost same as #68029 - copy-from arrows, clean loot groups, full metal bolt is transformed into making sense makeshift wooden and metal bolt
Lessons learned, recipe for bolts was also tweaked to compensate removed charges from bolts
#### Testing
it works


data/json/itemgroups/Locations_MapExtras/locations.json
```diff
@@ -802,7 +802,7 @@
     "id": "recycle_iron",
     "type": "item_group",
     "items": [
-      { "item": "bolt_metal", "prob": 15, "charges-min": 1, "charges-max": 3 },
+      { "item": "nuts_bolts", "prob": 15, "charges-min": 7, "charges-max": 200 },
```
inogenous, line 805: I don't know much really about this json, so might be out of my league here, but this `item_group` already seems to contain `nuts_bolts` at line 812. Maybe we should just remove this line.

GuardianDll: Oh, i miss that one, my bad
```suggestion
```

PR #68381, opened 2023-09-28
[MoM] Add Flameshield power

Standing-Storm:
#### Summary
Mods "Add Flameshield power"

#### Purpose of change

Allow pyrokinetics to become immune to fire at high power levels, the way telekinetics can become immune to physical damage and vitakinetics can become immune to disease

#### Describe the solution

Add the Flameshield power, which reduces the subjective temperature by 1000 degrees (toward a comfortable level) and reduces heat damage by 100%

#### Testing
image
image

#### Additional context

You can still get smoke inhalation but won't get burned by it.


MNG-cataclysm (approved):

MNG-cataclysm:
I feel like this has the potential to become incredibly busted; maybe creative players could burn down buildings to attract a horde and wade through the file to loot them before their stuff burns up. This could also make feral pyrokenetics and other fire-using enemies entirely trivial with a good suit of armor.

Standing-Storm:
> This could also make feral pyrokenetics and other fire-using enemies entirely trivial with a good suit of armor. 

It is true that feral pyrokinetics are essentially neutered by this power--but fair's fair, since feral pyrokinetics have their own version of it, the way that feral telekinetics have very high anti-TK armor and feral telepaths are immune to telepathic attack.

PR #68382, opened 2023-09-28
Add include directories to CMake SDL2 targets

alef:
#### Summary
None

#### Purpose of change
`perryprog` found the include dirs were missing in target definitions, so here they are.

#### Describe the solution
Add `INTERFACE_INCLUDE_DIRECTORIES` to `FindSDL2*.cmake`

#### Testing
Built in Visual Studio 2022 and VS Code in Docker Desktop WSL2 Dev Container. No attempt to run.
Let the CI build and see.


perryprog:
Ha, beat me to it. One thing of note is that in my testing it was only SDL2_image that needed the extra property. I'm pretty sure this is just because (perhaps on macOS specifically) it ends up in a directory not picked up by FindSDL2. The rest incidentally line up on the default include path. I'm not totally sure what changed on my setup that only made CMake recently fail to build this, but regardless, this change compiles and runs for me.

PR #68389, opened 2023-09-29
Adds priority parameter for special placement

lispcoc:
#### Summary
Features "Adds priority parameter for special placement"

#### Purpose of change
In special placement, some large facilities may be placed abnormally far from the player's starting point due to a low success rate of creation.
Currently, the order of generation for facilities is random, but I think this problem can be alleviated to some extent by giving them priority.

#### Describe the solution
reimplement #67723
Priority can be set for special placement. (int value, default = 0).
Generation will be determined in order of priority.

#### Testing
Although random elements are involved in map generation, I thought it would be possible to check the effectiveness by measuring the distance from the initial point to the refugee center.
Each survey was conducted 10 times.

refugee center: priority = 0
Mean: 206
Median: 196
Min: 83
Max: 390

refugee center: priority = 1
Mean: 104
Median: 87.5
Min: 17
Max: 326

refugee center: priority = -1 (Failed: 2/10)
Mean: 340
Median: 264
Min: 32
Max: 943

### DIFF
--- a/doc/OVERMAP.md
+++ b/doc/OVERMAP.md
@@ -389,7 +389,7 @@ original intersection.
 | `city_distance` | Min/max distance from a city edge that the special may be placed. Use -1 for unbounded.               |
 | `city_sizes`    | Min/max city size for a city that the special may be placed near. Use -1 for unbounded.               |
 | `occurrences`   | Min/max number of occurrences when placing the special. If UNIQUE flag is set, becomes X of Y chance. |
-| `priority`      | (default = 0) The generation process is executed in the order of specials with the highest value.     |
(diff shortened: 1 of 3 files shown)


Procyonae:
Hey I am still alive I just haven't managed to get back into this since I moved house.
I'm happy for you to take over #67723, the main reason I went with flags over an int was so it was easy to tell where your maps priority fit, with an int you'll have to check every single special to know where it falls unless priorities gets made its own object which would be ugly. I also don't want there to be loads of different priorities used because then mapgen randomness will be heavily diminished.
That said I'm perfectly happy with leaving it as an int as it's obviously more expandable in future. I'd like the documentation to make it clear though that this is for things that benefit from attempting placement before or after others (large maps, maps that are or require dependencies etc) as opposed to something that should be used willy nilly and maybe suggesting a range of numbers to use so it's less likely to be abused.

lispcoc:
Thank you for the great source code and ideas. I think your concerns are valid. I will documentate that it shuld be used only when the generation rate becomes extremely low, use only 1,-1.

PR #65231, opened 2023-04-21
[Xedra Evolved] Paraclesian

Maleclypse:
#### Summary
Mods "XE Paraclesians"

#### Purpose of change
I've been playing with some ideas inspired by My Sweet Cataclysm.   I had been on a roll for this secret project but now I've taken enough time off it that I need to look at what it needs to be minimally viable.  This will add six new starting character only mutation trees inspired by Made of Sugar.

#### Describe the solution
Extending the Hail Arcadia section of Xedra Evolved by adding new player character, races that start as slightly different humans but evolve into creatures of their attuned element.   Eventually these creatures will have an end condition to leave Earth and go to a different dimension and find their family.

#### Describe alternatives you've considered
Trying to manage this until I have it completed but that's likely to be ten k lines of json.

#### Testing
tbd

PR #63836, opened 2023-02-28
char/healing: fix/rework healing rates to be more logical

andrei8l:
#### Summary
None

#### Purpose of change
* The documentation, code, contributors, and mutation descriptions (and wiki!) all have different ideas of what `healing_awake` and `healing_resting` represent
* Sleeping characters get unjustified double healing
* The code for healing rate is unnecessarily hard to parse
* Healing effects on some limbs affect healing rate on other limbs as well. 
* Fixes: #61202

#### Describe the solution
Standardize the meaning of `healing_awake/resting`, namely:
- Rename `healing_resting`  to `healing_multiplier` and make it multiply the base healing rate
- Refluff/repurpose `healing_awake` to be a percentage of healing rate applied while awake
- Rewrite `Character::healing_rate()` to remove superfluous branching
- Modify healing mutations to maintain current awake healing rates
- Rebalance `Deterioration` and `Disintegration` to be negative `Fast Healer` and `Very Fast Healer` instead of nothingburgers. Rebalance `Weakening` to degrade HP at the smallest possible rate (1HP/day).

TODO:
- [X] fix mutation descriptions: `..heal 66% faster while awake...` (66% faster than zero is still zero). Removed specific information. I'll take suggestions for fluffier text.

#### Describe alternatives you've considered
Not bothering so much since there's a wounds rework in progress: there's no reason not to improve the current implementation while we're waiting for the rapture.

#### Testing
Updated test units

#### Additional context
<details>
<summary>Mutation healing rates before</summary>

|Mutation           | awake    |asleep  |
|----------------------|--------------|-----------|
|Regeneration       | 0.000200 |0.000450|
|Very Fast Healer   | 0.000066 |0.000216|
|Rapid Metabolism   | 0.000020 |0.000170|
|Fast Healer        | 0.000020 |0.000170|
|Slow Healer        | 0        |0.000075|
|Poor healer        | 0        |0.000034|
|Imperceptive Healer| 0        |0.000010|
|Weakening          | 0        |0.000075|
|Deterioration      | 0        |0.000100|
|Disintegration     |-0.000008 |0.000100|

</details>

<details>
<summary>Mutation healing rates after</summary>

|Mutation           | awake    |asleep |
|----------------------|-------------|----------|
|Regeneration       | 0.000200 |0.000250|
|Very Fast Healer   | 0.000066 |0.000150|
|Rapid Metabolism   | 0.000020 |0.000150|
|Fast Healer        | 0.000020 |0.000150|
[...]

PR #68395, opened 2023-09-29
(Oops) Fix negative rest quality

RenechCDDA:
#### Summary
Bugfixes "Negative rest quality actually slows your healing"

#### Purpose of change
Mergers were diligent in merging the PRs I marked as ready! Very good.
I was not diligent in marking it draft/testing. Very bad! (Bad Renech! Bad!)

Negative rest quality doesn't quite work, and the tests need to be updated.

#### Describe the solution
Stop clamping rest quality in Character::healing_rate_medicine (change to a std::max function so your bandages can never make your wounds *worse*)

Refactor Character::healing_rate (natural healing, very slow and without any sort of treatment) to account for rest quality that isn't 1 or 0. Specifically this means that rest quality >1 works, and rest quality below 0 will slow your healing.

For the natural healing rate I have tried to keep the rates as close as possible to previous values, but there will be some variance. Natural healing rate is very very slow compared to the medicine rate so I'm not terribly worried but just making a point of what is *intended*...

(TODO) Work on the tests so they actually pass/don't use hardcoded rest quality

#### Testing
WIP

#### Additional context
@Maleclypse you added a `healing_resting` entry to the docs in #65231 but we haven't had that json object since #63836. Not sure how it got in there, just letting you know since it's not actually in use anywhere, and anyone trying to add it to a mutation would get an unvisited member error.


RenechCDDA:
This PR ends up making some rather exciting but totally inappropriate changes. I rebuilt it to only make targeted fixes at #72892

Maleclypse:
Not sure but yeah if it doesn't exist it should exit the markdown.

PR #68398, opened 2023-09-29
Fix duplicate OMTs, take two

akrieger:
#### Summary
None

#### Purpose of change
Followup to #68324. I wasn't happy with the solution and felt it deserved more documentation. I also realize I missed an else clause during world generation which could have affected what omts spawn over others (if you have sequential linear OMTs of different types). I'm not sure this is actually a case that happens but in theory there could be a linear omt that requires a predecessor and that predecessor itself could be linear (I dunno) so we should allow stacking those.

#### Describe the solution
Redo the logic with extra wextra comments. Make the mapgen logic and savegame deserialization logic use almost literally exactly the same code. I could've factored it into a helper function maybe but the way it would possibly mutate local or non-local state made me a bit wary to do so. I could still try though. I want to get this in front of eyes sooner though.

#### Testing
Loaded the old save from #67837 and validated the deserialize logic works. Created a new save in the same build, crossroads still have manholes sometimes.

You can still get 4 way streets over manholes over 4 way streets. Not sure if this means those intersections are susceptible to extra spawning, but the one I found did not seem over full.
road_over_manhole_over_road

Roads still overwrite trails correctly (presuming trails are linear and roads are linear, this means my safety net for linear-over-linear is working). 
road_over_trail

And extra special verification that savegame recover works. This save had 7 duplicate omts serialized. It only pushes the one halfway through the list.
verified_recovery

### DIFF
--- a/src/overmap.cpp
+++ b/src/overmap.cpp
@@ -2887,21 +2887,38 @@ void overmap::ter_set( const tripoint_om_omt &p, const oter_id &id )
         return;
     }
 
-    oter_id &val = layer[p.z() + OVERMAP_DEPTH].terrain[p.xy()];
+    oter_id &current_oter = layer[p.z() + OVERMAP_DEPTH].terrain[p.xy()];
+    const oter_type_str_id &current_type_id = current_oter->get_type_id();
+    const oter_type_str_id &incoming_type_id = id->get_type_id();
+    const bool current_type_same = current_type_id == incoming_type_id;
+
+    // Mapgen refinement can push multiple different roads over each other.
+    // Roads require a predecessor. A road pushed over a road might cause a
+    // road to be a predecessor to another road. That causes too many spawns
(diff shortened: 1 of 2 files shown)


akrieger:
This PR looks scary but that's because git diff sucks.

First commit extracts the logic for deserializing `layers` into the space above the main loop with no logical changes, just an ordering change. The second commit actually implements the described functionality.

@Procyonae here's that redo I mentioned in discord. Mind taking a look?

Procyonae:
I think road over trails is the only instance of sequential linear OMTs of different types for now, manholes aren't linear, with railways potentially getting (re)added and unhardcoding trail maps #67470 it's very much a desired feature though so thanks for catching that ^^ (I'm also interested in seeing if I can make the stream mutable in innawoods #67065 work as a connection once #68389 is in, in which case that would intersect roads and trails too)
My understanding of what you've done is fairly minimal but it looks good, sorry again for causing this mess in the first place '>-<

PR #68400, opened 2023-09-29
Random cleanups part 4

GuardianDll:
#### Summary
None
#### Purpose of change
There is a lot of stuff in the game, that was created, but never got used, or things that we got rid of, but not completely. I'd like to clean some of them
#### Describe the solution
Removed, if not mentioned otherwise:
- NON-FOULING flag - do not confuse with NON_FOULING, that actually backed up in the code
- feral_big_pappy
- mon_zeriatric
- mon_dog_telekenetic - turned back where it belong, was accidentally delete during post 0.G obsoletion cleanups
- mon_firefly_larva - added to GROUP_EGG_FIREFLY that used to spawn fireflies from eggs
- mon_locust_small - added along another insects
- mon_mussel_river
- m1911_10_magazines_reloaded
- mon_triffid_sprig - this and next two are all added into triffid mongroups
- mon_biollante_sprig
- mon_biollante_sprout
- battery_test


data/json/morale_types.json
```diff
@@ -124,11 +124,6 @@
     "type": "morale_type",
     "text": "Ate human flesh"
   },
-  {
-    "id": "morale_demicannibal",
```
Maleclypse, line 128: If you don't mind leaving this in and making a bug report and I'll assign it to myself

data/json/morale_types.json
```diff
@@ -124,11 +124,6 @@
...
-    "type": "morale_type",
-    "text": "Ate demihuman flesh"
-  },
   {
```
GuardianDll, line 127:
```suggestion
  {
    "id": "morale_demicannibal",
    "type": "morale_type",
    "text": "Ate demihuman flesh"
  },
  {
```

PR #68404, opened 2023-09-30
Add a larger “family–sized” water heater

db48x:
#### Summary
Content "Add a larger “family–sized” water heater"

#### Purpose of change

Prior to the cataclysm it was common for larger households to have installed larger water heaters so that everybody could take a hot shower every night. Commercial establishments frequently required larger water heaters in order to serve customers reliably as well. Mapgen doesn’t reflect these facts well.

#### Describe the solution

Adds one larger size of water heater, with a 55–gallon capacity.

75 and 100–gallon water heaters are not uncommon, but that would require adding more tank objects.

#### Testing

`make check` for syntax, and I wandered around a world verifying that both sizes were available.

#### Additional context

Screenshot 2023-09-29 at 17-08-32 7e7928b1-8b4f-472a-8ee8-b48eb1e7932a pdf


data/json/furniture_and_terrain/furniture-plumbing.json
```diff
@@ -129,7 +129,42 @@
         { "item": "hose", "count": [ 0, 2 ] },
         { "item": "cu_pipe", "count": [ 1, 4 ] },
         { "item": "scrap_copper", "count": [ 0, 2 ] },
-        { "item": "water_faucet", "count": [ 0, 1 ] }
+        { "item": "water_faucet", "count": [ 0, 1 ] },
+        { "item": "material_aluminium_ingot", "count": 2 }
```
GuardianDll, line 133: why a whole ingot? i doubt water heater should drop a big solid blocks of aluminum, more like a lot of aluminum scrap

data/json/furniture_and_terrain/furniture-plumbing.json
```diff
@@ -129,7 +129,42 @@
...
+      ]
+    }
+  },
+  {
+    "type": "furniture",
+    "id": "f_water_heater_family",
+    "name": "family–sized water heater",
+    "looks_like": "f_water_heater",
+    "description": "An insulated metal tank with a small fire used to maintain near-boiling temperatures.  There's no longer any way to power it, but any water left in the large tank from before the power went out should be pretty clean, and it could be used to store large quantities of other liquids too.",
```
GuardianDll, line 142: > with a small fire

Should it imply it is a gas powered heater? i don't think they are popular, also player won't be able to use one on their base, right

data/json/mapgen_palettes/military/mil_base_palette.json
```diff
@@ -95,7 +95,7 @@
       "M": "f_makeshift_bed",
       "C": "f_crate_c",
       "L": "f_shower",
-      "6": "f_water_heater",
+      "6": [ "f_water_heater", "f_water_heater_family" ],
```
GuardianDll, line 98: i guess military facility can benefit from even heavier heaters, and shouldn't alternate between them

data/json/furniture_and_terrain/furniture-plumbing.json
db48x: The name is misleading; these “ingots” are 250g each.

data/json/furniture_and_terrain/furniture-plumbing.json
db48x: They are very popular in the north–east, because they easily have 10× the heating capacity of electric (literally 45kW vs 4.5kW). Also, the existing water heater has the same description, and it isn’t tagged as an electric appliance either.

data/json/furniture_and_terrain/furniture-plumbing.json
GuardianDll: "ingot" is not about the shape, it's about the form of an item

data/json/mapgen_palettes/military/mil_base_palette.json
db48x: That’s reasonable.

data/json/furniture_and_terrain/furniture-plumbing.json
db48x: Hmm. Well, the anode rod is in fact a rod longer than it is wide. Usually masses around a kilo, and is often divided into four segments…

db48x: You can do a google image search to see pictures if you want.

GuardianDll: ok, then it require some tweaks for aluminum to have more components, but it's out of scope of this PR

data/json/mapgen_palettes/aircraft_carrier_palette.json
```diff
@@ -78,7 +78,7 @@
       "#": "f_standing_tank",
       "S": "f_shower",
       "s": "f_sofa",
-      "V": "f_water_heater",
+      "V": [ "f_water_heater", "f_water_heater_family" ],
```
GuardianDll, line 81:
```suggestion
      "V": "f_water_heater_family",
```
Same as mil base

db48x:
I accepted your suggestion for water heater sizes on the carrier map, but I am not entirely sure that a modern carrier would have any water heaters at all. I’ve never encountered it in the game and I haven’t looked at the map, but it seems to me that a nuclear carrier would use the reactor to heat water and then pipe it around the ship instead of making electricity that is then converted back into heat in a water heater.

But if the map has water heaters, we might as well make them large ones :)

PR #68406, opened 2023-09-30
Context menu when bumping into things  (examine, deconstruct, pickup)

EvanBalster:
#### Summary
Interface "Configurable 'Bump to Interact' menu when walking into certain furniture or terrain in safe mode."

#### Purpose of change

Previously I added a multiple-choice context prompt when walking into ledges:  #68150.  This was more intuitive and convenient than what was previously there, presenting players with a list of useful interactions when bumping into an otherwise dangerous/impassable ledge.  It works by forwarding to the `e`xamine action.

This left me curious if a similar bump-to-interact menu would be useful for other things — primarily in a base management context.  Bumping into solar panels, refrigerators and workbenches would be just a bit nicer than reaching for the `e` key every time and might be a boon to accessibility and mobile gameplay.  I decided to experiment.

#### Describe the solution

When the player bumps into an impassable tile (other than an unlocked door) a menu may appear with applicable options from this list:

* Access items in
* Examine
* Deconstruct

If the only option would be 'move onto' or 'access items', it is picked automatically.  Here's a refrigerator:
image

There is also a new type of personal zone that can enable bump-interact for *all* movement-hampering tiles in a designated area.  This could be useful for the player's base, friendly camps and other safehouses.
image

A new "Auto Features" menu setting controls which tiles it may apply to, outside of designated zones:
image

- Disabled
- Impassable Tiles (eg, fridge, water heater, water purifier) — default!
- Useful Furniture (eg, workbench, countertops)
- Most furniture (ie, anything that hampers movement and is deconstructible/examinable/contains things)

Bump-interact only applies to furniture or terrain that blocks or hampers movement, and only when safe mode is on.  It never applies to NPCs or auto-move.

* Add a menu option to suppress bump-interaction for a little while (ie, in combat).  Safe mode fills this role.
* Only show the menu when the tile has a terrain/furniture flag, eg: `BUMP_INTERACT` or more specific `BUMP_EXAMINE`/`BUMP_PICKUP`/`BUMP_DECONSTRUCT`.
* Only show the menu (for passable tiles) if there are no hostiles in sight.

Other miscellaneous changes that might be useful:

* Automatically choose the "pickup" option when the only other interaction is non-simple deconstruction (eg, lockers).
* Move the new code to `iexamine.cpp`.
* Find some way to fold examine sub-options into the bump menu (eg, to avoid two consecutive menus when bump-examining a solar panel)

#### Testing

* Tested with safe mode on and off.
* Tried bumping refrigerators, chairs, cupboards.
* Tested taking down a chair.
* Tested connecting a fridge to the grid.
* Tested grid-connected appliance (**does not work for vehicles at this time**)

#### Additional context

Impassable stuff:
image

Passable stuff (potentially problematic):
image
image

### DIFF
--- a/data/json/zones.json
+++ b/data/json/zones.json
@@ -24,5 +24,12 @@
     "name": "Loot: Unload Everything",
     "can_be_personal": true,
     "description": "Any containers in this zone you don't otherwise sort will have their contents dumped out and placed on the same tile.  This includes corpses, magazines, and magazine wells."
+  },
+  {
+    "id": "BUMP_INTERACT",
+    "type": "LOOT_ZONE",
+    "name": "Bump to Interact",
+    "can_be_personal": true,
+    "description": "You can examine, access or deconstruct furniture in this zone by walking into it.  This includes workbenches, lockers and appliances.  Disabled when safe mode is off."
   }
 ]
--- a/src/avatar_action.cpp
+++ b/src/avatar_action.cpp
@@ -492,8 +492,14 @@ bool avatar_action::move( avatar &you, map &m, const tripoint &d )
     }
     if( int move_cost = m.move_cost_ter_furn( dest_loc );
         ( move_cost <= 0 || move_cost > 2 ) && you.is_avatar() && !you.is_auto_moving() ) {
+
+        // Check the bump-to-interact setting, and force it on in bump-interact zones.
+        std::string bump_setting = get_option<std::string>( "BUMP_TO_INTERACT" );
+        if( g->check_zone( zone_type_BUMP_INTERACT, dest_loc ) ) {
+            bump_setting = "all";
+        }
+
         // Generate context menu for bumped furniture.
-        bool is_bump_zone = g->check_zone( zone_type_BUMP_INTERACT, dest_loc );
         tripoint_bub_ms dest_loc_bub = tripoint_bub_ms( dest_loc );
         bool can_examine = can_interact_at( ACTION_EXAMINE, dest_loc );
         bool can_pickup = can_interact_at( ACTION_PICKUP, dest_loc );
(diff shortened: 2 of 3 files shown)


John-Candlebury:
Ideal far off suggestion: This becomes part of an integrated basecamp building/management mode toggle that lets you order NPCs around and assign them to planned constructions or crafting.

For now I think It'd be good to limit this exclusively to furniture within zones or appliances currently plugged to a grid. Its also worth considering having the menu be disabled if you are grabbing the piece of furniture you are walking against, since you are deliberately moving it and likely dont plan to stop until you release it.

oosyrag:
I am highly against having this happen for any furniture that can be walked on. I walk over furniture all the time, and having a popup every time would be fantastically annoying. It would be similar to navigating a forested area and constantly getting popup warnings for the bushes, although that's more of an acceptable evil due to the damaging result, and can be turned off.

I'm all for bump activation of non-traversable furnishings though!

EvanBalster:
Implemented a new zone type (see below).  Bump-interaction with *passable* objects is now limited to this zone.

Bump-interaction with passable objects still works everywhere though.  The only situation I've noticed where the menu might be annoying with impassable obstacles is when walking through thick forests of harvestable trees.

image
image

oosyrag:
Are harvesting trees part of the auto scavenge code? Would it make sense to exclude things that could be auto harvested via that setting already?

Also for normal trees, it could be fairly nice to bring up the chopping option when holding (wielding vs in inventory?) an axe, but just not have a pop up otherwise? It does take a few presses to get to chopping a tree down, that could definitely benefit from a contextual bump shortcut.

EvanBalster:
Implemented a menu setting to further control how this feature behaves outside defined zones.  Currently placing a zone bumps the setting up to max.  The options are:

- Disabled
- Impassable Tiles (eg, fridge, water heater, water purifier) — default!
- Useful Furniture (eg, workbench, countertops)
- Most furniture (ie, anything that hampers movement and is deconstructible/examinable/contains things)

image

PatrikLundell:
Getting a popup when trying to walk into a vehicle through an opening (door, hatch, broken part) would be extremely annoying. Getting it only some of the time would be annoying as well, in particular when trying to scavenge a mostly wrecked vehicle and trying to walk into one of the few tiles that wasn't destroyed.
Thus, I don't think vehicle bumping popups would be good.

kevingranade:
I'm not convinced that anything about this is a good idea.
The dependence on traversability couples together two things that have nothing to do with each other. This makes interactions of this type inconsistent and unpredictable. 
It can never be your only way to interact with furniture,  so sometimes you do this and sometimes you manually examine. This increases cognitive load instead of reducing it. 
Again,  I'm not convinced that removing the 'e'xamine keypress is even a meaningful improvement.
The zone thing makes it even more inconsistent and niche instead of making it better. 
If you're really interested in reducing keypresses in base management, I'm sure there's a ton of room for improvement in streamlining workflows, our as candlebury points out even a streamlined interface for interacting with many appliances where movement is abstracted away.  that's a much more fruitful angle to pursue

EvanBalster:
I'm inclined to take this one slow and playtest for a while.  I'm interested in playtesting feedback from other contributors/maintainers.

>The dependence on traversability couples together two things that have nothing to do with each other. This makes interactions of this type inconsistent and unpredictable.
> It can never be your only way to interact with furniture, so sometimes you do this and sometimes you manually examine. This increases cognitive load instead of reducing it.

I'm a bit concerned about this myself.  Trying to define broad rules for what should and should not provoke a context-sensitive interaction has led to a somewhat complicated set of conditions.  It might make more sense to apply this feature to specific furniture on a case-by-case basis.  EG, we can 'bump' doors or windows to open them because it's a very sensible extension of the move action.  It's my opinion that opening a locker or fridge follows the same intuition.

The set of options for context-sensitive interaction has a similar problem: it's currently based on general rules but the 'deconstruct' option has proven problematic because it often means we will show a menu when we wouldn't otherwise.  We'll be opening the locker far more often than deconstructing it, after all.  Furthermore, 'deconstruct' and 'examine' have the same effect on many types of furniture ("deployable" items).  This issue might be mitigated by merging menus with `iexamine` prompts, similar to what I've done with the ledge menu, but that sort of tinkering ought to wait until signs are positive for the feature as a whole.

PR #68417, opened 2023-09-30
Added bundles of 50 for kevlar, lycra, and nomex patches.

Papaya-Plan:
#### Summary
Content "Bundles of kevlar, lyrca, and nomex patches"

#### Purpose of change
Other crafting patches (cotton, leather, synthetic fabric) have convenient, easy to move bundles. Kevlar, lycra, and nomex do not. This is especially annoying when trying to clean these items in large quantities (100+). #68214 helps, but even with it, spending 5 minutes collecting patches into a container, 5 minutes putting them in a washing machine, and 5 minutes removing them felt silly.

#### Describe the solution
Kevlar, lycra, and nomex patches can be made into bundles, like other patches.

- The ability to rapidly to rapidly empty a container, effectively holding it by the bottom, turning it upside down, and shaking it until empty. It should only take a handful of seconds to empty most containers, regardless of how many items are inside (caveat: containers with narrow openings, such as large bottles, would take longer.) This would help remove the need for bundles in situations where they wouldn't logically be needed.
- A more generic bundle type container item thing, made out of, some member of the string/rope family, with a pocket of appropriate dimensions. Possibly that can be used to store nearby items ("You bundle your nearby kevlar patches into a neat stack for transportation") ala how the current crafting recipes work, and that can also be used to disassemble it in a way that let's you control where the items go. (The problem with current disassembling is that the items appear either in your inventory, on the ground, or on a nearby worksurface. So if you have, for example, a bundle of kevlar patches (filthy), and you would like the patches to be in a container, the only way to do so is to disassemble the bundle, then place the resultant patches into a container. Rather than just place the bundle in a container, then untying the bundle.)
The biggest obstacle for both of these is that I don't know how to do either of them.
- Removing bundles, since pockets now exist. That seems like it should be part of a bigger discussion.

#### Testing
Spawned in bundles, wooden needle, and pocket knife. Disassembled bundles of each material into 50 patches each. Remade them into bundles. Compared weights of bundles and patches to ensure no violation of conservation of mass.

#### Additional context
[Discussion of bundles and bundle alternatives on Discord.](https://discord.com/channels/598523535169945603/598529174302490644/1157643744930181190)


GuardianDll:
image
This message meant "you can make it for your own purpose, since it would be rejected for vanilla purposes", not "you can test your luck and make it"

Papaya-Plan:
> image This message meant "you can make it for your own purpose, since it would be rejected for vanilla", not "you can test your luck and make it"
That wasn't how this was meant to come across. This is just the best solution I can make at the moment for the issue I was having.
image

esotericist:
this is definitely a trend we need to be reversing, not continuing. the underlying problem needs to be addressed, adding more bundle crafting is just papering it over and not even very effectively.

i agree the current situation is frustrating but another solution, which is more general or directly addresses why it's frustrating, is needed here.

PR #68390, opened 2023-09-29
Documentation update

GuardianDll:
#### Summary
None
#### Purpose of change
Just documenting fields 
Adress #67783

Issue #67783, opened 2023-08-19
Many JSON fields are undocumented in the .md files


ampersand55:
### Describe the bug

Many JSON fields are undocumented in the .MD files, which makes it harder to create content.

### Attach save file

N/A

### Steps to reproduce

N/A

### Expected behavior

I would like every field to be documented.

Perhaps even add a PR check to ensure that all fields are documented. 

### Versions and configuration

- OS: Windows
    - OS Version: 10.0.14393.6085 (1607)
- Game Version: 2faf5df [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda]
]

### Additional context

Here's a list of fields used in JSON data I've found using a script, that are undocumented or documented in an un-intuitive .md-file. Many of the `effect > ` is documented in NPC.md.

Not guaranteed to be exhaustive, and may contain false positives. 

Angle bracket `>` denotes a nested key.

 ```
Field                                                  Json-file field was first found in

requirements > description                             \achievements.json

trail > chance                                         \ammo_effects.json
trail > field_type                                     \ammo_effects.json
trail > intensity_max                                  \ammo_effects.json
trail > intensity_min                                  \ammo_effects.json

passive_pseudo_items                                   \bionics.json
fake_weapon                                            \bionics.json
active_flags                                           \bionics.json
auto_deactivates                                       \bionics.json
toggled_pseudo_items                                   \bionics.json
spell_on_activation                                    \bionics.json
activated_close_ui                                     \bionics.json
power_trickle                                          \bionics.json

requirements > *                                       \butchery_requirements.json

mod_type                                               \character_modifiers.json
value > *                                              \character_modifiers.json

qualities                                              \construction.json
pre_note                                               \construction.json
dark_craftable                                         \construction.json
byproducts                                             \construction.json
tools                                                  \construction.json
qualities                                              \construction.json
using                                                  \construction.json
difficulty                                             \construction.json
do_turn_special                                        \construction.json
activity_level                                         \construction.json
vehicle_start                                          \construction.json
skill                                                  \construction.json
on_display                                             \construction.json

base_mods > dodge_mod                                  \effects.json
base_mods > hit_mod                                    \effects.json
dur_add_perc                                           \effects.json
resist_effects                                         \effects.json
scaling_mods > bash_mod                                \effects.json
show_in_info                                           \effects.json

time_save_flags                                        \faults\fixes_gun.json

[...]

PR #34483, opened 2019-10-05
Update for Knockback_follow attribute

Hymore246:
#### Summary

```SUMMARY: Content "Update for Knockback_follow attribute"```

#### Purpose of change

Knockback Follow attribute updated to prevent bugs and other strange situations.

#### Describe the solution

Knockback Follow has a tendency to behave oddly when handling "dangerous tiles". For example, if you light a fire and then attack an enemy that is in the space with the file with a Knockback Follow technique, you will be prompted to move into the space. If you accept you will be moved into the space but the game will not realize you are in that space at the end of the turn. i,e, you will not catch on fire. You can move out of the space next turn with no ill effects but staying there will light you up.

Because of this problem, I took the time to not only fix the issue but also update the attribute in a few other ways:
* Knockback Follow is now a bool variable instead of an int variable. There was no reason to have a "high number" on the knockback follow variable. You either follow after a knockback or you don't.
* The movement from Knockback Follow now calls g->on_move_effects(); to properly handle movement events.
* Rather than prompting the player to move into a dangerous tile, the game will simply cancel the movement if it detects the tile to be dangerous.

In addition to dangerous tiles, Knockback Follow will also cancel movement if any of the following situation occur.
* Player would move into deep water from land ("Dive into deep water?")
* Player is driving a vehicle.
* Player would move out of a moving vehicle ("Dive from a moving vehicle?")
* Player is mounted (Your horse doesn't know Wing Chun)
* Player has the Amigara effect (This is here just to prevent event weirdness)

Lastly, I discovered that there wasn't a way to check for a dangerous tile without prompting the player. So I created two new functions and updated prompt_dangerous_tile.
* **prompt_dangerous_tile** - now calls get_dangerous_tile and then prompts the player for movement.
* **get_dangerous_tile** - checks for harmful stuff and returns an array of the hazards. Array can be empty.
* **is_dangerous_tile** - calls get_dangerous_tile then returns true if there are any hazards.

PR #68418, opened 2023-09-30
Martial arts docs update

GuardianDll:
#### Summary
None
#### Purpose of change
Another documentation stuff
Adress #67783
#### Describe the solution
Documented some fields
The reason it's a separate PR is because i found and confirmed that `powerful_knockback` has no back up in code - it was introduced in #34407:

https://github.com/CleverRaven/Cataclysm-DDA/pull/34407/files#diff-2cd8dac3d0903ca1ee60a47c28e09bcb56290214bc8d1fe5cf4afaed108b9d92R1333

to change the way the game calculate said knockback to be similar with vehicle collision, and deal additional damage when pushed enemy collides with another critter or wall; However the code that made it possible was (accidentally, i believe) removed few days later, in #34483:

https://github.com/CleverRaven/Cataclysm-DDA/pull/34483/files#diff-2cd8dac3d0903ca1ee60a47c28e09bcb56290214bc8d1fe5cf4afaed108b9d92L1332

I need someone to confirm, that i am correct, and that i didn't remove anything important accidentally

#### Additional context
Accidentally checked it out from #68390, so it has some side commits also

### DIFF
--- a/doc/JSON_INFO.md
+++ b/doc/JSON_INFO.md
@@ -4008,7 +4008,7 @@ Alternately, every item (book, tool, armor, even food) can be used as a gunmod i
 "revert_msg": "The torch fades out.", // Message, that would be printed, when revert_to is used
 "sub": "hotplate",         // optional; this tool has the same functions as another tool
 "variables": {
-  "vehicle_name": "Wheelchair",         // this tool is a foldable vehicle, that could bypass the default foldability rules; this is the id of vehicle to unfold
+  "vehicle_name": "Wheelchair",         // this tool is a foldable vehicle, that could bypass the default foldability rules; this is the name of the vehicle that would be unfolded 
   "folded_parts": "folded_parts_syntax" // this is the parts that this vehice has -it uses it's own syntax, different from `"type": "vehicle"`, so better to read the examples in `unfoldable.json`
 }
 ```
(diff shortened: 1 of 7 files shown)


doc/JSON_INFO.md
```diff
@@ -3178,6 +3181,9 @@ Weakpoints only match if they share the same id, so it's important to define the
 "price_postapoc": "1 USD",                       // Same as price but represent value post cataclysm. Can use string "cent" "USD" or "kUSD".
 "stackable": true,                           // This item can be stacked together, similarly to `charges`
 "degradation_multiplier": 0.8,               // Controls how quickly an item degrades when taking damage. 0 = no degradation. Defaults to 1.0.
+"solar_efficiency": 0.3,                     // Efficiency of solar energy conversion for solarpacks; require SOLARPACK_ON to generate electricity; default 0
+"source_monster": "mon_child",               // This item is corpse of this monster (so it has weight and volume of this monster), and revive into this monster; require COPRSE flag
```
Maleclypse, line 3185: So does mon_child revive into a mon_zombie_child?

GuardianDll: oh, i copied the wrong number, i checked out from #68390, it has this changes actually - this one has only small code stuff and MA doc updates
answering your question - no, it actually revive into `mon_child`, <https://github.com/CleverRaven/Cataclysm-DDA/blob/master/data/json/monsters/mammal.json#L14>, which already zombifies into `mon_zombie_child`; i'll change this monster into some another in that PR

GuardianDll, line 3185:
```suggestion
"source_monster": "mon_zombie",               // This item is corpse of this monster (so it has weight and volume of this monster), and revive into this monster; require COPRSE flag
```

PR #68420, opened 2023-09-30
Change Grim Reapers Scythe

Holli-Git:
#### Summary
Balance "Adjust Grim Reaper Scythe"

#### Purpose of change

A solid plastic toy is a luxury most people wouldn't have. So this PR changes it to be hollow. Didn't want to do the math of 

#### Describe the solution

Removes techniques and nerfs damage, but reduced weight. Did some garbage math. Assumed the volume was a cylinder with 2.5mm wall thickness, giving 2932.75 cm^3 of empty space. Is it perfect? No. Did I have to do a bunch of math to try to find out what the actual volume would be? No. Should I have just searched up a toy and get its weight from the product page? Probably. Also adds it to the known bad density json, don't know if that was needed or not

Forcing kids to grow up and use real scythes as toys.

#### Testing

Why not?

### DIFF
--- a/data/mods/TEST_DATA/known_bad_density.json
+++ b/data/mods/TEST_DATA/known_bad_density.json
@@ -774,8 +774,7 @@
       "rocuronium",
       "alumentum",
       "rebreather_filler_ko2",
-      "rebreather_filler_soda_lime",
-      "scythe_toy"
+      "rebreather_filler_soda_lime"
     ]
   }
 ]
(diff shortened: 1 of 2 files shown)


data/mods/TEST_DATA/known_bad_density.json
```diff
@@ -774,7 +774,8 @@
       "rocuronium",
       "alumentum",
       "rebreather_filler_ko2",
-      "rebreather_filler_soda_lime"
+      "rebreather_filler_soda_lime",
+      "scythe_toy"
```
Karol1223, line 778: Are you adding it to known bad density? No, don't do that. That's not how this works. If something has bad density you fix it, not write down that it's bad. Please do not do this.

Paprika-H: Alright dope, done. Thank you!

PR #68426, opened 2023-09-30
Make it obvious when TCL badges are destroyed or not

inogenous:
#### Summary
Content "Make it obvious when TCL badges are destroyed or not"

#### Purpose of change

* ~~Updates the TCL card readers that have `consume_card: true` so that it is a bit more obvious to the player that the card itself will be consumed (destroyed).~~
* ~~Updates all card readers that actually consume (destroy) the card so that the player know what will happen.~~
* Edit: Updates the TCL card readers that previously had `consume_card: true` so they now instead are  `consume_card: false`, to make them not consume (destroy) the card. This makes them work like the other colored TCL card readers.

#### Describe the solution

Story time:
* First time I visited TCL, I had a few visitor's badges with me.
* So I opened all doors I could find in the outer yard.
* This, unknownst to me, consumed (destroyed) all visitor's badges that I had with me.
* Then I encountered another card reader that also required a `visitor's pass`, but lo and behold I had none!
* That taught me that TCL cards are consumed (destroyed) when used.
* So I slowly realized that these TCL doors are a puzzle in itself, and thinking to myself that it's difficult choice on where you choose to use your TCL badges.
* So, because of these first few doors that consumed (destroyed) some of the badges, I had the feeling that all doors would consume (destroy) the corresponding card.
* My playstyle was therefore to open as few doors as possible, to not consume what I perceived was precious cards.

This commit therefore suggests that we make it obvious to the player when the card will be:
* kept (not consumed): using the phrasing `swipe the card`
* consumed (destroyed): using the phrasing `insert the card`

* Another option would be to rename `visitor's badge` into something like `one-time use visitor's badge`.

### DIFF
--- a/data/json/furniture_and_terrain/terrain-mechanisms.json
+++ b/data/json/furniture_and_terrain/terrain-mechanisms.json
@@ -400,7 +400,7 @@
     "type": "terrain",
     "id": "t_card_science_visitor",
     "name": "Facility entrance card reader",
-    "description": "A smartcard reader.  It sports the stylized symbol of an atom inside a flask that is universally known to indicate SCIENCE.  You could swipe a visitor's pass near it to unlock the gates.",
+    "description": "A smartcard reader.  It sports the stylized symbol of an atom inside a flask that is universally known to indicate SCIENCE.  You could insert a visitor's pass into it to unlock the gates.",
     "//": "Entrance card for a lab.  Low security access.",
     "symbol": "6",
     "color": "cyan",
@@ -415,7 +415,7 @@
       "radius": 3,
       "terrain_changes": { "t_door_metal_locked": "t_door_metal_c" },
       "query": true,
-      "query_msg": "Are you sure you want to open this door?",
+      "query_msg": "Are you sure you want to open this door? It looks like you might not get the card back.",
       "success_msg": "You opened the door!",
       "redundant_msg": "The door is already open."
     }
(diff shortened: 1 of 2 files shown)


MNG-cataclysm:
Why do card readers destroy the cards they read anyway? For TCL staff and visitors, this seems like an incredibly inefficient security system, as you'd need to produce massive amounts of one-use cards and distribute them, creating a security risk where someone could steal them; greater volumes of things are usually harder to track than smaller ones. On top of that, this would produce a massive amount of plastic waste that would need to be recycled or disposed of, only driving up the costs of this single-use card system.

I instead propose that these ID badges be reusable, as I believe that it would make more sense than destroying them.

Daved27hundred:
> Why do card readers destroy the cards they read anyway? 

I'm p sure its some video game logic thing that nobody ever bothered to change.

PatrikLundell:
Single use cards make no sense, as visits aren't joy rides at an amusement park. I can certainly see a potential use for a card that expires e.g. 12 hours after its first usage, in particular if it can be reset (the logic may reside in the card, or the security system may log the card Id so it can be reused after having been handed back).
The obvious way to use visitor's passes would be to restrict the locations you can get to, although a real world usage would probably have that set up on a visitor per visitor basis (someone supposed to temporarily work in a specific location needs access to that location, but not anything else that isn't a general area or part of the path to the target location).

inogenous:
Thanks for all the good input! Have updated this now so that the PR instead makes the visitor- and transport- card-readers instead **not** consume the cards at all.

PR #68393, opened 2023-09-29
Small tweaks

GuardianDll:
#### Summary
None
#### Purpose of change
Few stuff that occured on discord and that i find pretty easy to fix
#### Describe the solution
lorica segmentata is in armor category now
radiation badge mention it doesn't turn back to neutral state
calf zombifies into small cocoon
small update of t_bulk_tank and 55gal_drum because i spend my time calculating it's dimensions
fix #68388

PR #68428, opened 2023-09-30
Make it clearer from radiation badge description that the item is single use

harakka:
#### Summary
None

#### Purpose of change
Make it clearer from radiation badge description that the item is single use.

#### Describe the solution
Clarify wording

Don't clarify wording


GuardianDll:
#68393

harakka:
Closing since there is already a PR up with similar change.

PR #68434, opened 2023-09-30
Jeweler profession

Daved27hundred:
#### Summary
Content "Adds a jeweler as a new starting profession"

#### Purpose of change

More options for characters. This profession could be considered A sidepath to the blacksmith profession, trading the useful blacksmithing proficiency for several, lesser used and more specific proficiencies.

#### Describe the solution

Adds the jeweler as a profession, which starts with fabrication 6 and a metalworking chisel. The proficiencies they start with are Metalworking, Fine Metalsmithing, Redsmithing and Gem Setting (Redsmithing is required by Gem setting, which requires metalworking.) 

removing some of the proficiencies they start with or not adding them altogether. I'm also open to changing the description bc I think it could use work..

#### Additional context

Gem Setting currently does nothing, I hope this gets fixed in the future.

### DIFF
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -3355,6 +3355,23 @@
     }
   },
   {
+	"type": "profession",
+	"id": "jeweller",
+	"name": "Jeweller",
+	"description": "You were an independent artisan, who made gorgeous rings and pendants for the people who afforded your services.  You doubt the market for your work has faired well after the end of the world, but perhaps your proficiency with fine tools and metals will come in handy somehow.",
+	"points": 2,
+	"skills": [ { "level": 6, "name": "fabrication" }, { "level": 2, "name": "swimming" } ],
+	"proficiencies": [ "prof_metalworking", "prof_redsmithing", "prof_fine_metalsmithing", "prof_gem_setting" ],
+	"items": {
+		"both": {
+			"items": [ "jeans", "socks", "boots", "longshirt", "wristwatch", "apron_leather", "chisel" ],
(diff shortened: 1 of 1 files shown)


Brambor:
I saw somewhere a comment like this:

Why does the jeweller have swimming 2? Is it required? You are forcing the player to have swimming 2 if they choose the jeweller profession, which limits roleplay. It is better to let them decide if they want swimming 2.

Daved27hundred:
@Brambor Swimming was renamed to "athletics" and made to cover various activities instead of just swimming. It's still named swimming in the code.
The blacksmith had athletics 2, so I gave them it since they work with similar tools

Rabadash94:
The jeweler might also have some basic knowledge of chemistry.

Daved27hundred:
how so?

Rabadash94:
With Principles of Chemistry proficience, like the lab technician.

Rabadash94:
If you're asking why, jewelers need basic chemistry knowledge about metals and gems, and some chemical agents, such as acids.

Daved27hundred:
I'll probably need a more concrete source for that before people raise their eyebrows at it.

Daved27hundred:
looked it up, Chemical etching is a thing. I'll add an applied science skill, but I think principles of chemistry is too much for them.

Daved27hundred:
Coming back to this - I think the fact that none of the jewelry crafts require any chemistry or applied science skill to make should tell me that the profession doesn't need it. Maybe that can change in the future, when there's more jewelry crafts, but for now I'm keeping it as is.

PR #68441, opened 2023-10-01
Rename luna park to miniature railway park

harakka:
#### Summary
None

#### Purpose of change
"Luna park" is not an established idiom in English, as it apparently is in a bunch of other languages.
This PR renames the location to miniature railway park based on its main attraction, to make clearer to players what the location actually is. Translators are of course free to keep calling it luna park in languages where that makes sense.

#### Describe the solution
Rename luna park to miniature railway park

#### Testing

(Source: [Wikipedia page on Luna Park](https://en.wikipedia.org/wiki/Luna_Park))

### DIFF
--- a/data/json/overmap/overmap_special/specials.json
+++ b/data/json/overmap/overmap_special/specials.json
@@ -6163,7 +6163,7 @@
   },
   {
     "type": "overmap_special",
-    "id": "Luna Park",
+    "id": "miniature railway park",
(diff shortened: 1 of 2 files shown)


Rabadash94:
The change makes a lot of sense. Especially in Spanish (AR), since "Luna Park" is a stadium/theater in Buenos Aires.

PR #68446, opened 2023-10-01
Fix conversion of liquid and solid items

### LINKED ISSUE #68354: "delayed_transform" in "blanching acorns" is not working normally
### Describe the bug

"delayed_transform" in "blanching acorns" is not working normally.
"delayed_transform" doesn't seem to support changing from a stackable item to a non-stackable item.
Also, when the value of "phase" is different between the item before the change and the item after the change, it seems that the "phase" of the item after the change is not properly referenced.

### Attach save file

N/A

### Steps to reproduce

1. Create "blanching acorns".
2. After the specified time, using "blanching acorns"(id:acorns_blanching) results in an error.
   13:28:32.195 ERROR : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\item.cpp:7347 [volume] Item type acorns_blanched has invalid default charges 0

Continuing the game will generate "blanched acorns" that behave strangely.
・Only one "blanched acorns" is generated, regardless of the number of "blanching acorns".
・Spawned blanched acorns do not go into inventory.
　It becomes an item that cannot be placed in a container, held in hand, or added to the inventory.

### Expected behavior

blanched acorns are successfully generated.

### Screenshots

_No response_

- OS: Windows
- Game Version: 6b306c1 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],

### Additional context

_No response_

lispcoc:
#### Summary
NONE

#### Purpose of change
Fixes #68354.

#### Describe the solution
item::convert does not account for liquid to solid conversion, fix this.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -631,6 +631,10 @@ item &item::convert( const itype_id &new_type, Character *carrier )
     temp.update_modified_pockets();
     temp.contents.combine( contents, true );
     contents = temp.contents;
+    current_phase = new_type->phase;
+    if( count_by_charges() != new_type->count_by_charges() ) {
+        charges = new_type->charges_default();
+    }
 
     if( temp.type->countdown_interval > 0_seconds ) {
         countdown_point = calendar::turn + type->countdown_interval;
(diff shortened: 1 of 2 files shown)


casswedson:
can you use github's [keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) so the issue you address is closed automatically

PR #68457, opened 2023-10-02
Cleanup npctalk.cpp

lispcoc:
#### Summary
None

#### Purpose of change
npctalk.cpp has more if-elses than the compiler can tolerate. 

#### Describe the solution
Use vectors to reduce the number of lines.

#### Additional context
It's better than the current situation, but I think there's probably a better way.
There are more places that need to be replaced with string_view, but I won't start on them here as they would be too large.


src/npctalk.cpp
```diff
@@ -5460,314 +5483,156 @@ talk_effect_t::talk_effect_t( const JsonObject &jo, const std::string &member_na
...
+        if( ( a & jarg::member ) && jo.has_member( key ) ) {
+            return true;
+        } else if( ( a & jarg::object ) && jo.has_object( key ) ) {
+            return true;
+        } else if( ( a & jarg::string ) && jo.has_string( key ) ) {
+            return true;
+        } else if( ( a & jarg::array ) && jo.has_array( key ) ) {
+            return true;
+        }
+        return false;
```
andrei8l, line 5604:
```suggestion
        switch( a ) {
            // *INDENT-OFF*
            case jarg::member: return jo.has_member( key );
            case jarg::object: return jo.has_object( key );
            case jarg::string: return jo.has_string( key );
            case jarg::array: return jo.has_array( key );
            default: return false;
            // *INDENT-ON*
        }
```

src/npctalk.cpp
```diff
@@ -5460,314 +5483,156 @@ talk_effect_t::talk_effect_t( const JsonObject &jo, const std::string &member_na
     }
 }
 
+static const
+std::vector<std::tuple<std::string, std::string, jarg, void ( talk_effect_fun_t::* )( const JsonObject &, const std::string &, bool )>>
+un_parsers = {
+    std::make_tuple( "u_add_effect", "npc_add_effect", jarg::member, &talk_effect_fun_t::set_add_effect ),
```
andrei8l, line 5489: Use modern syntax:
```suggestion
    { "u_add_effect", "npc_add_effect", jarg::member, &talk_effect_fun_t::set_add_effect },
```
And please use a real struct. Tuples and their `std::get<>` syntax are pretty ugly and there's no reason to prefer them here.

Also, for a container of this size, `std::map` is likely better (just like in the [math parser](https://github.com/CleverRaven/Cataclysm-DDA/blob/0b00ccbd3008bc0f7f6dabea02e26002ea3d258a/src/math_parser_diag.h#L123-L139)).

src/dialogue_helpers.h
```diff
@@ -50,20 +50,19 @@ struct talk_effect_fun_t {
         void set_remove_active_mission( const JsonObject &jo, const std::string &member );
         void set_offer_mission( const JsonObject &jo, const std::string &member );
         void set_make_sound( const JsonObject &jo, const std::string &member, bool is_npc );
-        void set_run_eocs( const JsonObject &jo, std::string_view member );
-        void set_run_eoc_with( const JsonObject &jo, std::string_view member );
-        void set_run_eoc_until( const JsonObject &jo, std::string_view member );
+        void set_run_eocs( const JsonObject &jo, const std::string &member );
```
andrei8l, line 53: I'd rather see the other ones moved to `std::string_view` than going backwards on this. Don't worry about churn as it's fully justified in this case.

andrei8l (commented):
Thanks a lot for doing this!

I know this is still in draft but some parts are pretty ugly at the moment.

src/npctalk.cpp
lispcoc: At this point, the key string have not been determined, so I thought that using map would actually slow things down. Once we adopt the vector. Please let me know if there is a better way.

src/npctalk.cpp
lispcoc: I think this switch statement may miss functions that accept both arrays and objects. Leave the current if-else as is.

Maleclypse:
Do you want to rebase after the GC9 fix was merged?

lispcoc:
> Do you want to rebase after the GC9 fix was merged?

I'll do it if I have to, but I think there is no conflict. Please merge if possible

PR #68458, opened 2023-10-02
add mx_laststand

NetSysFire:
#### Summary
Content "New map extra: civilians making a futile last stand against the horde"

#### Purpose of change

Makes the world feel more dynamic and alive. Draft because of issues discovered during testing.

#### Describe the solution

There is a barricade of mostly common furniture as a fortification. A number of overconfident officers and futile fighters will spawn inside.

Also made the zombie trap map extra autonoted by default as it is quite notable.

- Using different furniture. I mean where do you get enough materials for all those sand, earth and gravel bag barricades?
- Barbed wire! Unfortunately it is terrain and not furniture.
- Spawning blood stains and so on, but they will naturally occur during combat.
- Spawning garbage as in used bandages, empty packages of food and drink inside the barricades.
- Creating some nested mapgen to also spawn 2x3 dumpsters and pool tables as eligible furniture. Dumpsters should be vehicles honestly.

#### Testing

image

Applied changes locally. I need to figure out why spawned monsters do not evolve. I set the time to year 2 and there were still civilians. Weeeird. Also when I manually spawn the map extra it takes some turns until the monsters are actually spawned. The terrain and furniture is there right away.

After I announced that I am working on this in the devcord, @akrieger asked me to keep my eyes peeled for any other occurances of the wreckage and zombie population problem, mostly the former.

I think despite `ERASE_ALL_BEFORE_PLACING_TERRAIN` there can be errors if the RNG decides to place a vehicle where this map extra would spawn. This is one of two map extras that can spawn on a junction like this and afaik the vehicle spawns are in C++. The below message has been spammed a heckin *2492* of times with minor variations in the coordinates. Curious.

```
ERROR : src/vehicle_display.cpp:64 [vpart_display vehicle::get_display_of_tile(const point&, bool, bool, bool, bool) const] no display part at mount (-2, -2)
```

The reason why this may not have been caught before is because the zombie trap map extra is extremely small with just one tile.
Still unsure about the above error.


Rabadash94:
Outside of this PR, but a scenario based on this would be great.

akrieger:
"hrm after running tests a few more times I reproduced the mapgen thing, it's placing a vehicle with some parts hanging off the edge of the map and blowing up when it checks every tile of the vehicle for passengers" I believe was the quote.

Procyonae:
I think you want to make a monster group like this
```
  {
    "type": "monstergroup",
    "name": "GROUP_LAST_STAND",
    "default": "mon_null",
    "monsters": [
      { "monster": "mon_civilian_police", "weight": 1, "cost_multiplier": 1, "ends": "5 days" },
      { "monster": "mon_civilian_zombiefighter", "weight": 1, "cost_multiplier": 1, "ends": "5 days" }
    ]
  },
```

NetSysFire:
Aight, rebased to fix conflicts, added the mongroup and also some trash for flavor. Works as expected for me now.

Night-Pryanik:
> Spawning blood stains and so on, but they will naturally occur during combat.

Since you have end date for spawning monsters, these locations might seem strangely clean and empty after end date.

NetSysFire:
This should be mitigated by the fact that these only spawn on a 4 way intersection, which is inside the city which is already populated. Because the monsters do not evolve due to the mentioned limitation, we will have the same issues and I'd need to hardcode the entire tree of evolutions into the group.

Procyonae:
Doesn't this need a nest that's placed before everything else that removes pre-existing vehicles?

Issue #67927, opened 2023-08-28
Magiclysm container CI failure


Maleclypse:
### Describe the bug

```
(all_mods)=> (continued from above) ERROR : src/item.cpp:1601 [ret_val<void> item::put_in(const item &, item_pocket::pocket_type, const bool)] tried to put an item (small_mana_crystal) count (1) in a container (wand_magic_missile_minor) that cannot contain it: pocket unacceptable because holster already contains an item
(all_mods)=> 02:29:46.837 ERROR : src/item.cpp:1601 [ret_val<void> item::put_in(const item &, item_pocket::pocket_type, const bool)] tried to put an item (small_mana_crystal) count (1) in a container (wand_fireball_minor) that cannot contain it: pocket unacceptable because holster already contains an item
(all_mods)=> 02:29:46.849 ERROR : src/item.cpp:1601 [ret_val<void> item::put_in(const item &, item_pocket::pocket_type, const bool)] tried to put an item (small_mana_crystal) count (1) in a container (wand_magic_missile_minor) that cannot contain it: pocket unacceptable because holster already contains an item
```

### Attach save file

N/A

### Steps to reproduce

CI runs

### Expected behavior

CI runs with few rng errors

### Versions and configuration

NA

Maleclypse:
So the more I'm seeing this.  I think the error is that the wand is appearing already loaded.

mqrause:
I already had a bunch of explanations of what's going on with this in #68468, so I'm not gonna repeat it here. But since that PR is an insufficient solution that leaves other bugs I was looking into doing a C++ fix for tools with magazines. Turns out the easy fix for that (handling tools like guns in `Item_modifier::modify`) isn't possible because `item::magazine_compatible` (used in `item::ammo_set`) only checks item restrictions and not flag restrictions. Since at least all the electric tools rely on that it instantly blows up when trying to load those with battery, since there are no compatible magazines. The only reason it didn't blow up for guns, yet, is that apparently none of them use flag restrictions.

PR #68468, opened 2023-10-02
Fix item group error with magiclysm + defense mode enabled

mqrause:
#### Summary
None

#### Purpose of change

I don't know if it's a complete fix, but it addresses #67927.
I'm also not sure if it's a good fix, see alternatives section.

#### Describe the solution

Remove charges modifier from a group entry, which seems to cause conflicts with other (sub-)entries from within that group.

https://github.com/CleverRaven/Cataclysm-DDA/blob/e3eddedf3eb16328197b82cc2dd3b17619ed1026/src/item_group.cpp#L571-L580

I think the real issue might lie in this. The same code for guns instead of tools treats both magazines and items with magazine wells the same and just calls `ammo_set` on them instead of manually inserting a magazine. But I can't really test if this breaks anything (and then fix potentially resulting issues) currently.

#### Testing

Run tests with `--rng-seed 1696217676 '~*' --mods=dda,stats_through_kills,standard_combat_test,defense_mode,magiclysm,no_fungal_growth,blazeindustries,megafauna,ruralbiome,railroads,package_bionic_professions,Tamable_Wildlife,extra_mut_scens,sees_player_hitbutton,Mythos-Creatures,StatsThroughSkills,mindovermatter,my_sweet_cataclysm,generic_guns,speedydex,translate_dialogue,personal_portal_storms,alt_map_key,tropicata,cbm_slots,MMA,sees_player_retro,crazy_cataclysm,deadly_bites,no_npc_food,xedra_evolved,DinoMod,no_hope,Only_Wildlife,bombastic_perks` before and after the change.
(Just enabling defense_mode and magiclysm might be enough, but no idea if that messes with the rng in some way.)

#### Additional context

It actually still fails with the above seed with this fix, but with a different error: `[get_part_id] Could not find equivalent bodypart id mouth in Tom 'Tom' Tom's body`.

### DIFF
--- a/data/mods/Defense_Mode/npcs.json
+++ b/data/mods/Defense_Mode/npcs.json
@@ -184,7 +184,7 @@
       { "group": "preserved_food", "prob": 70 },
       { "group": "pet_food", "prob": 70 },
       { "group": "condiments", "prob": 70 },
-      { "group": "trader_guns&ammo", "prob": 70, "charges": 30 },
+      { "group": "trader_guns&ammo", "prob": 70 },
       { "group": "hardware", "prob": 70 },
       { "group": "defense_caravan_melee", "prob": 70 },
       { "group": "defense_caravan_ranged", "prob": 70, "charges": 30 },


MNG-cataclysm:
I’ve got some more code for this:
https://github.com/mqrause/Cataclysm-DDA/pull/2

mqrause:
I guess I skipped over the part about what the bug actually is.
Magiclysm extends the `guns_pistol_rare` group with the wands:

https://github.com/CleverRaven/Cataclysm-DDA/blob/5621850411c36baca2978f8779cf4579723cdfd1/data/mods/Magiclysm/itemgroups/itemgroups.json#L94-L105

It has it's own charge definition, which conflicts with the charge definition from `{ "group": "trader_guns&ammo", "prob": 70, "charges": 30 },`, which contains the `guns_pistol_rare` group. But this conflict only really happens because the wands are tools and not guns. The latter seem to just override one charge definition with the other (though I couldn't tell you which takes priority).

So this PR here probably really isn't a good fix for the issue, and your proposed additional changes would just reintroduce what it's trying to fix. Someone should probably look into the code mentioned in the alternatives section to make everything work as intended and prevent bugs like this popping up.

MNG-cataclysm:
@Maleclypse This is going to reintroduce a previous bug with the merchant spam reloading everything in their inventory. I'll have to go and fix that from these changes.

Maleclypse:
> @Maleclypse This is going to reintroduce a previous bug with the merchant spam reloading everything in their inventory. I'll have to go and fix that from these changes.

Oh I guess I had misunderstood that then. Sorry.

mqrause:
I don't think there's a good json fix for this. Best I can think of is a separate item group definition for this specific use case. Which would be kinda weird because it'd be a magiclysm group that's only used in defense mode. It really needs a C++ fix (and a revert of this before or after that), so item groups like this work the same for guns and tools.

PR #68480, opened 2023-10-03
Selecting container mode for unloaded items

lispcoc:
#### Summary
Interface "Selecting container mode for unloaded items"

#### Purpose of change
Container for [U]nloaded items is selected automatically. 
You can store items as you want by setting up whitelists and blacklists appropriately, but there are times when you want to select container manually.

#### Describe the solution
~Add game option "Container for unloaded items".~ Make it enable to toggle contain mode "Auto", "Manual", for unload menu. Default is "Auto". If "Manual" is selected, 
after selecting the item to unload, you can select the container to insert it into.

Even without this feature, the insert menu[v] may be sufficient.

#### Testing
Animation1


Nebnis:
> I am not completely sure how it looks like in game, but i believe it's better to make it "manual" by default, not "auto"

you select U and then pick a container and it kinda just unloads all your items in your inventory until it can't fit anymore. Maybe there should be a button to unload like this behavior and one like lispcoc instead?

GuardianDll:
now i am confused even more

GuardianDll:
Gif of using old way vs gif of how this PR makes it work is enough

John-Candlebury:
While it might not be a solution you are interested in pursuing I've always thought the 'U'nload menu should actually be an unload and insert menu, with a hotkey swap that lets you swap between just emptying a container and the behavior you have here, where the new item location is specified.

lispcoc:
> Gif of using old way vs gif of how this PR makes it work is enough

Auto
Animation1

Manual
Animation2

lispcoc:
Clang 14 tests are failing, but I don't think this PR is the cause. Should I redo the test?

Nebnis:
I believe as long all required tests pass, you don't need to do anything else. Thank you so much for all the QOL lately

PR #68486, opened 2023-10-03
Monsters care about item theft

ehughsbaird:
#### Summary
Features "Monsters affiliated with factions will watch for theft of faction items"

#### Purpose of change
Only the NPCs in the exodii will care when you steal items. That's wrong, and easily exploitable.

#### Describe the solution
Make monsters who witness theft, who are in the monster faction associated with the faction that the item was stolen from, either alert nearby NPCs who are in the faction that an item was stolen, or if there are no nearby NPCs, become hostile.

Don't have them go hostile? The NPC alerting is kind of magic to start with.

#### Testing
Steal stuff in the exodii castle, Rubik talks to me about all the stuff I've been stealing.
Kill the NPCs in the exodii castle, when a exodii worker or quad witnesses me stealing, they become hostile.

#### Additional context
This needs some more work, but as it is works as a proof of concept.

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -1949,6 +1949,23 @@ bool item::is_owned_by( const Character &c, bool available_to_take ) const
     return c.get_faction()->id == get_owner();
 }
 
+bool item::is_owned_by( const monster &m, bool available_to_take ) const
+{
+    // owner.is_null() implies faction_id( "no_faction" ) which shouldn't happen, or no owner at all.
+    // either way, certain situations this means the thing is available to take.
+    // in other scenarios we actually really want to check for id == id, even for no_faction
+    if( get_owner().is_null() ) {
+        return available_to_take;
+    }
+    for( const std::pair<const faction_id, faction> &fact : g->faction_manager_ptr->all() ) {
+        if( fact.second.mon_faction != m.faction.id() ) {
+            continue;
+        }
+        return fact.first == get_owner();
+    }
+    return false;
+}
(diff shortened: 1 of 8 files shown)


PatrikLundell:
You don't need magic for cyborgs to communicate at a distance: all it takes is radio waves.

It makes sense for the Exodii, where the critters are cyborgs with human(?) brains, but you don't want farm animals to behave in the same way. Thus, critters that should report to their masters should probably have some tag that causes them to do that.

Brambor (commented):
The comment should have not stayed there. It shouldn't have been merged.

src/item.cpp
```diff
@@ -6309,26 +6326,37 @@ void item::handle_pickup_ownership( Character &c )
...
+                    owned_by = is_owned_by( *as_npc );
+                } else if( as_monster ) {
+                    owned_by = is_owned_by( *as_monster );
                 }
-            }
+                return &cr != &c && owned_by && rl_dist( cr.pos(), c.pos() ) < MAX_VIEW_DISTANCE &&
+                       cr.sees( c.pos() );
+            };
+            std::vector<Creature *> witnesses = g->get_creatures_if( sees_stealing );
+            /*
```
Brambor, line 6347: This comment

ehughsbaird: Thanks! I've fixed that in https://github.com/CleverRaven/Cataclysm-DDA/pull/68563

Issue #63210, opened 2023-01-20
Add a "monster_species_excluded" (or similar) blacklist for spell targets


Standing-Storm:
### Is your feature request related to a problem? Please describe.

I'm working on a psychic powers mod and had the idea that telepathy should mostly not work on zombies. However, using "targeted_monster_species": [ "HUMAN" ] to allow telepathically mind-controlling ferals or NPCs but not zombies (for example) doesn't work because zombies that used to be human all have the HUMAN species flag. 

### Solution you would like.

A blacklist field in addition to the current whitelist field, so that there's a way to create a spell that effects everyone except a small group(s). (All except zombies, all Earthly creatures but not Nether beings or Migo, etc)

### Describe alternatives you have considered.

Listing every possible feral ID in the powers as targeted_monster_ids, but 1) it wouldn't mix with any mods 2) it might not work with NPCs and 3) I'm not sure how targeted_monster_ids and targeted_monster_species interact if they're both on the same spell.

PR #68488, opened 2023-10-04
Add ignored_monster_species spell parameter

Standing-Storm:
#### Summary
Features "Add ignored_monster_species spell parameter"

#### Purpose of change

One problem with implementing more telepathic powers in Mind Over Matter is that it's hard to make them thematic when they can be used on robots, zombies, hallucinations caused by the breakdown of reality, and other beings that don't have brains.  

Implements #63210

#### Describe the solution

Create the ignored_monster_species JSON parameter, the exact opposite of targeted_monster_species. A spell that has an ignored_monster_species list will work on everything _except_ those species. Also apply ignored_monster_species to the Mind Control power, so it no longer works on zombies, robots, or Nether creatures.

A spell that has a targeted_monster_species and an ignored_monster_species containing the same list of species won't affect anything. 

#### Testing

Apply ignored_monster_species to Mind Control, spawn in zombie and cow. Cows are targetable and zombies result in 
image

#### Additional context

This will allow me to create a telepathic fear power so you can paralyze people with overwhelming terror and not having it be too powerful (because it won't work on zombies, robots, and other things that can't feel fear).

Still trying to figure out how to get Mind Control working on NPCs.

### DIFF
--- a/data/mods/MindOverMatter/powers/telepathy.json
+++ b/data/mods/MindOverMatter/powers/telepathy.json
@@ -377,7 +377,8 @@
     "energy_increment": -250,
     "base_casting_time": 120,
     "final_casting_time": 75,
-    "casting_time_increment": -3.5
+    "casting_time_increment": -3.5,
+    "ignored_monster_species": [ "ZOMBIE", "ROBOT", "NETHER", "NETHER_EMANATION", "LEECH_PLANT", "WORM", "FUNGUS", "SLIME" ]
   },
   {
     "id": "telepathic_network",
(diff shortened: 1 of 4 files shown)


MNG-cataclysm:
For getting Mind Control to work on NPCs, what if it applied something like the Debug Mind Control mutation to your character, opening up the options of "OBEY ME!" to get the NPC to do whatever you want? Alternatively, this could change the NPCs first talk topic to a new menu involving more thematic mind control.

Standing-Storm:
I did experiment with Debug Mind Control and the problem with that was that it makes the NPC a member of your faction with no duration, so that would be much too powerful. There's a "follow_only" command but I think it's only possible to trigger it through dialogue and (the real problem) only possible to disengage it through dialogue as well, so you'd end up with a permanent mind-controlled follower as long as you never talked to the person again. 

If it's possible to somehow store a particular NPC's first topic so that it could be set to a generic NPC_MIND_CONTROLLED topic and then go back to their individual topics after a set period of time without breaking missions and dialogue all over the place, I don't know how to do it.

data/mods/MindOverMatter/powers/telepathy.json
```diff
@@ -377,7 +377,8 @@
     "energy_increment": -250,
     "base_casting_time": 120,
     "final_casting_time": 75,
-    "casting_time_increment": -3.5
+    "casting_time_increment": -3.5,
+    "ignored_monster_species": [ "ZOMBIE", "ROBOT", "NETHER", "NETHER_EMANATION", "LEECH_PLANT", "WORM", "FUNGUS", "SLIME" ]
```
Maleclypse, line 381: `NETHER` monsters often have minds.  Though I guess it might be accurate to say that their intelligence might be too alien to our minds for telepathic damage to impact them.

Standing-Storm: The latter is what I was going for--that Nether monsters could be telepathically attacked (unsubtle, brute-force, swinging a two-by-four) but not telepathically mind-controlled (subtle, delicate, wielding a scalpel)

PR #68502, opened 2023-10-04
Raise overmap generation priority of Refugee Center

lispcoc:
#### Summary
None

#### Purpose of change
Refugee Center is presented as the first destination after starting the game in the 
standard scenario, but the distance to get there can be hundreds of tiles.
In the past, not being able to reach refugee centers was not a big problem. Currently, many questlines center around refugee center, so not being able to reach that will limit the scope of play.

#### Describe the solution
Raise the priority of generating refugee center and reduces the probability of generation failure.
It may still spawn more than 300 tiles away, but should be better than it is now.

It may be wiser to introduce a new mechanism so that it always spawns within a certain distance from the game's starting point.

#### Testing
A simple test was conducted with #68389.

#### Additional context
What is the appropriate distance to the refugee center?

### DIFF
--- a/data/json/overmap/overmap_special/specials.json
+++ b/data/json/overmap/overmap_special/specials.json
@@ -2241,6 +2241,7 @@
     "city_distance": [ 3, -1 ],
     "city_sizes": [ 4, -1 ],
     "occurrences": [ 75, 100 ],
+    "priority": 1,
     "flags": [ "GLOBALLY_UNIQUE", "SAFE_AT_WORLDGEN", "MAN_MADE" ]
   },
   {


PatrikLundell:
Reachability is a more important criterion than distance, but harder to control (a destination can be 10 tiles away, but on an island with no bridges, as an extreme example).

I'd probably want a world generation logic that generated physical terrain first, then added a standard start evacuation shelter with the standard nearby facilities with the refugee center on the same land mass and initial roads and bridges to connect the standard facilities (Tacoma and Exodii base, and probably Hub01 as well) if needed, then add towns etc. (which may replace some of the roads, but should ensure connections), and then place monster placement exclusion indicators around all the unique facilities generated and the access paths to at least the first ones (no fungal towers with vast fields around them, ant hills, slime pits, etc. such that their spread areas aren't visible from the roads or the facility perimeters. Out or reality bubble range may only be critical for the fungal monstrosities, though, as it spreads unchecked whenever inside the bubble, whereas the other areas appears to be possible to skirt without them spreading and blocking the path).

If a non standard start is selected I'd suggest the start location ought to be as close to the evacuation center as possible (again with a preference for reachability over distance).

lispcoc:
I tried to collect some opinions, but there was no particular objection that it is desirable for the refugee center to be generated within the first overmap (180x180), so I think it is okay to merge it.
In the future, I think it will be necessary to improve the ability to generate facilities more stably at a certain distance (I can't do that with my current knowledge), but I think this will have a certain effect as a temporary measure.

ThePainkiller:
Tried a new spawn, refugee center at +200 tiles distance from the shelter. Still very, VERY far away...

evilbananas:
My understanding is "occurrences" changes the distance but could be wrong.  I play with [ 90, 100 ] and find it to be reasonable.

PR #68516, opened 2023-10-05
Rabbit audit

### LINKED ISSUE #68513: rabbit birth out adult rabbit and baby rabbit dont exists
### Describe the bug

Apparently according to the rabbit's json, rabbit birth out adult rabbit and baby rabbit doesn't exists.

`"reproduction": {
    "baby_monster": "mon_rabbit",
    "baby_count": 3,
    "baby_timer": 55
  },`

### Attach save file

N/A

### Steps to reproduce

1. Spawn rabbits in an enclosed place
2. wait 55 days.
3. more adult rabbits

### Expected behavior

Birthing out baby rabbits.

### Screenshots

_No response_

- OS: Android
- Game Version: 4805ce5 [32-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    GUT's Tweak Compendium [gut_tweaks_compendium],
    GUT's Contents Compendium [gut_contents_compendium]
]

### Additional context

_No response_

Karol1223:
#### Summary
None

#### Purpose of change
Resolves #68513 but also I refuse to do only as much as it's asked of me

#### Describe the solution
- Renamed rabbit to domestic rabbit
- Edited their speed from 160 to 666, proportionally calculated from the assumption that default player's speed of 100 equals to 6kmph, when internet says rabbits get to speeds of 40kmph
- Audited their weight and volume. Weight is the way it is, because internet says non-dwarf breeds get to anywhere between 1.8kg and 2.6kg - I went for a middle ground. Volume is calculated as a cylinder of 40cm and a diameter of ~13cm (the real volume comes out at 5.3l, but I slightly lowered it)
- Edited the breeding to take 33 days and produce 6 kits, but also produce new rabbit kits instead of adults
- Added rabbit kits
- Added rabbit skulls given to the domestic rabbit and our jackrabbit

- I wanted to make rabbits able to consume food, but I do not believe that as it stands we are capable of blacklisting/whitelisting what material a monster would be able to consume, and so I opted not to do that
- 666 speed, while somewhat realistic, is ridiculous and the rabbit can run outside of the player's field of view in about three to four moves from the player. If I am asked to lower it, I will.

#### Additional context
- https://en.wikipedia.org/wiki/Domestic_rabbit - source for information about when domestic rabbits reach sexual maturity which was used to determine how long it'll take for rabbit kits to "upgrade" to adults. The result is 180 days divided by 4, as 4 is the default evolution factor
- https://www.petplan.co.uk/pet-information/blog/a-guide-to-rabbit-pregnancy/ - source for gestation length and average amount of kits per litter

It's also important to note that this PR doesn't strictly make rabbits better or worse when it comes to farming them for food. Adult rabbits will produce more offspring in a shorter timeframe, but the offspring will now give no meat at all until they also wait 45 days. This means that rabbit farming will be more productive in the long term, but it won't be an instantaneous food drop


Nebnis:
I feel like leaving the rabbit speed as is the right call. A lot of our animals have tons of top speed than they currently have but they aren't something insane like that. Horses for example can run 25 to 30 mph or 42 to 48 km/h.

I assume the current speed is just some average speed rather than being i'm running away from predators top speed all the time.

Karol1223:
> I feel like leaving the rabbit speed as is the right call

Well, no, definitely not that. Speed of 160 is ridiculously low. But I will likely lower it, 600 is ridiculously high as well.

ehughsbaird:
Going off of Wikipedia, I'd expect [Eastern Cottontails](https://en.wikipedia.org/wiki/Eastern_cottontail) more than European-native domestic rabbits in the area the game is set. See also the [hunting seasons](https://www.mass.gov/info-details/rabbit-and-hare-hunting-regulations).

Karol1223:
> Going off of Wikipedia, I'd expect [Eastern Cottontails](https://en.wikipedia.org/wiki/Eastern_cottontail) more than European-native domestic rabbits in the area the game is set. See also the [hunting seasons](https://www.mass.gov/info-details/rabbit-and-hare-hunting-regulations). 

Our rabbits are simulatenously domestic and wild. Given that the domestic rabbit is indeed invasive in the US, it is far safer to assume there's wild domestic rabbits than that people were keeping cottontails as pets.

Adding dedicated wild species is something that should ideally be done in the future, but I can't be bothered.

ehughsbaird:
Okay. I was working under the assumption that these were all wild rabbits (subjective experience would paint those as more common). I suppose those would fit as a separate monster. Thanks for your work on the rabbits!

Karol1223:
> Okay. I was working under the assumption that these were all wild rabbits (subjective experience would paint those as more common).

For the record - invasive domestics *should* be rare, and only about a month ago I learned ours are supposed to represent both wild and domestic rabbits at once (they're tameable and if memory serves me right might be in the pet mongroups). In an ideal world we would have cottontails (and possibly other native lagomorphs) as separate monsters, but I just don't have time to do it right now.

ehughsbaird:
I'm on it, don't worry :)

pscz:
Are you comparing human speed-walking/jogging speed to rabbit sprinting speed? The average human sprinting speed is likely somewhere around 20-25kmph. Comparing like with like, and depending on what type of rabbit we're talking about, a rabbit speed of 160-300 might be more reasonable.

PR #68523, opened 2023-10-06
Rabbits 2: Oh god they zoomin

Karol1223:
#### Summary
None

#### Purpose of change
Remember when in #68516 I said I wasn't sold on the speed stat?

#### Describe the solution
- Lowered the speed of adult rabbits from 666 to 300, rabbit kits from 175 to 125 - this is roughly 50% of the accurate max speed for adults, but assuming that they're constantly running at full speed is kinda stupid
- Gave rabbits ``GOODHEARING``

- Lowering daylight vision, but I wanted it to be the same as turkeys and turkeys have it *higher* than rabbits already so I just didn't touch it

### DIFF
--- a/data/json/monsters/mammal.json
+++ b/data/json/monsters/mammal.json
@@ -2477,8 +2477,8 @@
     "volume": "5 L",
     "weight": "2 kg",
     "hp": 8,
-    "speed": 666,
-    "attack_cost": 666,
+    "speed": 300,
+    "attack_cost": 300,
     "material": [ "flesh" ],
     "symbol": "r",
     "color": "brown",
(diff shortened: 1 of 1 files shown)


sadenar:
Would there been an option to give them a dash movement boost to double their speed when scared? Not ridiculously important but could be a better simulation of what you're trying to achieve?

Karol1223:
> Would there been an option to give them a dash movement boost to double their speed when scared? Not ridiculously important but could be a better simulation of what you're trying to achieve?

Honestly, way too much of a bother for just a damn rabbit. I opened a feature request about making this work easier and more dynamically, but whether someone picks it up is a roulette.

PR #68533, opened 2023-10-07
Move gunsmith/blacksmith turret out of the direct path of the player

MoocowTG:
#### Summary
Balance "Move the isolated_road turret 3 tiles to the west"

#### Purpose of change

Prevent the player accidentally attacking the turret and losing their 20 hour character. Yes I'm mad.

#### Describe the solution

Move the turret out of the way.

Removing it altogether, but it's off to the side nicely here.

#### Testing

Spawned the location in world and made sure it's were I want it to be.

### DIFF
--- a/data/json/mapgen/isolated_road/isolated_road_home_chunks.json
+++ b/data/json/mapgen/isolated_road/isolated_road_home_chunks.json
@@ -169,7 +169,7 @@
       "place_monster": [
         { "monster": "mon_turret_artisans", "x": 8, "y": 1, "chance": 100 },
         { "monster": "mon_turret_artisans", "x": 16, "y": 1, "chance": 100 },
-        { "monster": "mon_turret_artisans", "x": 23, "y": 23, "chance": 100 },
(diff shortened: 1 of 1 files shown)


casswedson:
so you want to move the turret that sits in between the 2 houses here; but a location does not need to make sense in a player-friendly / safety way, using the rationale "its inconvenient position ended up killing my character" is not valid, turn safe mode on, it will prevent you from attacking neutral monsters, generally pay attention at what you are doing
Screenshot from 2023-10-06 21-14-57

MoocowTG:
Never would it have crossed my mind someone would argue against such a simple QoL change. Navigable terrain is a core principle of level design regardless of the type of game. Is there some reason you're trying to keep things inconvenient?

oosyrag:
I'd move it 3 or 4 tiles southeast, then it would make still make sense and be more out of the way

KHeket:
Maybe, what is better - place it to the roof at this place, where I mark? Because it will make the way safe for player, and also it reasonable to place turret to place, where the zombies cant get it, and the turret can attack potential enemies at the roofs?
image

PR #68544, opened 2023-10-07
Split drop leg poches

Dan1ss1mo:
#### Summary
Bugfixes "drop leg pouches can be worn separately"

#### Purpose of change
I assume there are introduced before siding and wasn't changed since then. Don't see any reason why they shouldn't be worn separetly otherwise

#### Testing
Seems to work fine in game

### DIFF
--- a/data/json/items/armor/storage.json
+++ b/data/json/items/armor/storage.json
@@ -1352,24 +1352,17 @@
   {
     "id": "legrig",
     "type": "ARMOR",
-    "name": { "str": "pair of drop leg pouches", "str_pl": "pairs of drop leg pouches" },
-    "description": "A set of pouches that can be worn on the thighs using buckled straps.  This variety is favored by the military.",
-    "weight": "205 g",
-    "volume": "1050 ml",
-    "price": 3000,
-    "price_postapoc": 250,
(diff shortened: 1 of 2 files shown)


data/json/items/armor/storage.json
```diff
@@ -1352,24 +1352,17 @@
...
     "id": "legrig",
     "type": "ARMOR",
-    "name": { "str": "pair of drop leg pouches", "str_pl": "pairs of drop leg pouches" },
-    "description": "A set of pouches that can be worn on the thighs using buckled straps.  This variety is favored by the military.",
-    "weight": "205 g",
-    "volume": "1050 ml",
-    "price": 3000,
-    "price_postapoc": 250,
+    "name": { "str": "drop leg pouche", "str_pl": "drop leg pouches" },
+    "description": "A pouche that can be worn on the thighs using buckled straps.  This variety is favored by the military.",
```
Karol1223, line 1356:
```suggestion
    "name": { "str": "drop leg pouch", "str_pl": "drop leg pouches" },
    "description": "A pouch that can be worn on the thighs using buckled straps.  This variety is favored by the military.",
```

Karol1223 (changes requested):

PR #68545, opened 2023-10-07
XL Kettle craftable

LordBarkBread:
Makes the XL Kettle helmet craftable.

#### Summary
Content "Making XL Kettle Helmet craft"

#### Purpose of change
Noticed during my gameplay that while we have both a kettle helmet and XL kettle helmet that the latter XL version is not craftable.

#### Describe the solution
Added in craft recipe

N/A

#### Testing
Works fine in current save.

#### Additional context
My head and horns are too big, bring out the Kettle Helmet stretcher!

### DIFF
--- a/data/json/recipes/armor/head.json
+++ b/data/json/recipes/armor/head.json
@@ -2664,6 +2664,27 @@
     ]
   },
   {
+    "result": "xl_helmet_kettle",
+    "type": "recipe",
+    "activity_level": "BRISK_EXERCISE",
+    "category": "CC_ARMOR",
+    "subcategory": "CSC_ARMOR_HEAD",
+    "skill_used": "fabrication",
+    "difficulty": 5,
+    "time": "8 h 20 m",
+    "book_learn": [ [ "textbook_armwest", 4 ] ],
+    "using": [ [ "blacksmithing_standard", 6 ], [ "steel_standard", 2 ] ],
+    "qualities": [ { "id": "CHISEL", "level": 3 } ],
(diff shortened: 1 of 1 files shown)


data/json/recipes/armor/head.json
```diff
@@ -2663,6 +2663,27 @@
...
+    "result": "xl_helmet_kettle",
+    "type": "recipe",
+    "activity_level": "BRISK_EXERCISE",
+    "category": "CC_ARMOR",
+    "subcategory": "CSC_ARMOR_HEAD",
+    "skill_used": "fabrication",
+    "difficulty": 5,
+    "time": "8 h 20 m",
+    "book_learn": [ [ "textbook_armwest", 4 ] ],
+    "using": [ [ "blacksmithing_standard", 6 ], [ "steel_standard", 2 ] ],
```
Karol1223, line 2676:
```suggestion
    "using": [ [ "blacksmithing_standard", 12 ], [ "steel_standard", 3 ] ],
```

Helmet weighs 3.19kg so using only 2kg to craft it doesn't really fly well. ``blacksmithing_standard`` is set for a chunk, one ``steel_standard`` is 4 chunks so 3x4=12. I think we can assume the .19kg is from the fabrics

The regular kettle helmet is actually generating mass in it's recipe, too, so if you have the time and energy I'd ask you to edit it as well. It's fine if you don't want to, though. You're not obligated to fix that.

Karol1223 (changes requested):

PR #68547, opened 2023-10-07
Add cottontail (wild) rabbits

ehughsbaird:
#### Summary
Content "Add undomesticated wild rabbits"

#### Purpose of change
@Karol1223 and I had a conversation in https://github.com/CleverRaven/Cataclysm-DDA/pull/68516#issuecomment-1750837922 about the rabbits being domestic rabbits, instead of wild ones.
Due to this, I decided the game could use some wild rabbits.

#### Describe the solution
Create a rabbit monster abstract, which all the rabbits copy from. This fixes a bug where `"delete"` was not applying for `"petfood"` or `"reproduction"`.
Add cottontail rabbits, modeled off of [Eastern Cottontail rabbits](https://en.wikipedia.org/wiki/Eastern_cottontail). They are not tameable, but are otherwise largely the same as domestic rabbits.
Adjust wilderness monstergroups so that any rabbits in them are predominantly non-domestic rabbits.

Different body sizes, different timers for reproduction/growth - the values given previously seemed sufficient.

#### Testing
Spawn in cottontail rabbits, see that they cannot be tamed. Teleport away and let time pass, see they reproduce into cottontail rabbit kits.


Karol1223:
> Create a rabbit monster abstract, which all the rabbits copy from. This fixes a bug where "delete" was not applying for ``"petfood"`` or ``"reproduction"``.

``delete`` *does* work for the ``reproduction`` field - I specifically tested it with te rabbit kits. It might work with ``petfood`` as well, and if it does we shouldn't need an abstract. But that'd require further testing - ``delete`` works on things it chooses to work on, so I can't blame you for assuming it wouldn't work here.

Overall looks good to me, even if I'm late to the party and this got merged prior to me able to review it lol

ehughsbaird:
I tested it for reproduction, after finding out it didn't work for petfood (the cottontails originally copied from the domesitc rabbits) - it seems to require two timeskips to get reproduction going.
I spawned a bunch of kits in an enclosed area, teleported away, set time forward a few days, teleported back. No babies.
Teleport away, set time forward a few days again, heleport back, and there are more rabbits than we started with, all babies.

PR #68548, opened 2023-10-07
Professions can start with multiple martial arts

Hymore246:
#### Summary
Features "Professions can start with multiple martial arts"

#### Purpose of change

Professions can start with multiple martial arts and no longer need style traits

#### Describe the solution

The parameters for professions.json definitions has been updated to allow martial arts to be assigned to a profession directly. This means that profession no longer need a special feat to assign them a martial art or offer a choice of a martial arts to the player at chargen. 

This update adds the following fields to the profession definitions:
**starting_styles** - Profession automatically learns all listed martial at the start of the game. The values used in this field are the style ids ("style_fencing") but these will be displayed in-game as the actual names of the styles. This field is optional and blank by default meaning that the profession does not know any martial art styles.

**starting_styles_choices** - Offers the player a choice of the listed martial arts in the same vain as the martial arts traits. The values used in this field are the style ids ("style_fencing") but these will be displayed in-game as the actual names of the styles. This field is optional and blank by default meaning that the profession does not know any martial art styles and the player will not be offered any choices from the profession.

**starting_styles_choices_amount** - Indicates how many martial arts from the starting_styles_choices list will be learned. This is done by repeatedly showing the "learn style" menu with previously selected choices removed. By default this parameter is 1, meaning a single style from starting_styles_choices (if any) will be learned. If this parameter is 0 or less, the player will not learn any martial arts from the profession. If this parameter is greater than the number of styles in starting_styles_choices, the player will learn all styles in the list (chosen one at a time) until all styles are known before starting the game.

In addition, the professions Competitive Fencer, Martial Artist, and Black Belt have been updated to use these new parameter while removing their martial arts enabling traits.

Currently, players will begin the game using a random style from their known styles. This can cause some oddities such as being in an unarmed only style while holding a weapon if they know a martial art from a different source (i.e. background) but this is easily fixed by changing styles to something else. Creating a method to select a random martial art based off of a held item (if any) is complex, might not be doable at chargen, and starts veering outside of the scope of this PR. Better to revisit it at a later time.

#### Testing

1. New section of the profession summary and available martial arts for profession can be in the profession section of the chargen screen.
2. Creating a Competitive Fencer will start with Fencing known.
3. Creating a Martial Artist or Black Belt will offer you a choice of styles as normal. Added **starting_styles_choices_amount = 3,** to either profession in profession.json will allow you to select three styles instead.

#### Additional context

New "Profession Martial Arts" section in profession section of chargen:
CDDA1

Competitive Fencer with a "known" style:
CDDA2

Black Belt with a "choice" of styles:
CDDA3

A "what if" example of Survivor with "known" style and "choice" styles:
CDDA4

### DIFF
--- a/src/profession.h
+++ b/src/profession.h
@@ -72,6 +72,8 @@ class profession
         std::vector<bionic_id> _starting_CBMs;
         std::vector<proficiency_id> _starting_proficiencies;
         std::vector<trait_and_var> _starting_traits;
+        std::vector<matype_id> _starting_martialarts;
+        std::vector<matype_id> _starting_martialarts_choices;
         std::set<trait_id> _forbidden_traits;
         std::vector<mtype_id> _starting_pets;
         vproto_id _starting_vehicle = vproto_id::NULL_ID();
@@ -117,6 +119,9 @@ class profession
         std::vector<mtype_id> pets() const;
         std::vector<bionic_id> CBMs() const;
         std::vector<proficiency_id> proficiencies() const;
+        std::vector<matype_id> ma_known() const;
+        std::vector<matype_id> ma_choices() const;
+        int ma_choice_amount;
         StartingSkillList skills() const;
         const std::vector<mission_type_id> &missions() const;
         int age_lower = 21;
--- a/data/json/professions.json
+++ b/data/json/professions.json
@@ -3920,7 +3920,25 @@
     "name": "Martial Artist",
     "description": "'Drop the martial arts!', they said.  'Learn a skill for the REAL world!', they said.  Well, you're about to show them!  SHOW THEM ALL!",
     "points": 3,
-    "traits": [ "PROF_MA_ORANGE" ],
+    "starting_styles_choices": [
+      "style_aikido",
+      "style_capoeira",
+      "style_crane",
+      "style_dragon",
+      "style_judo",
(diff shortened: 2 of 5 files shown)


src/profession.h
```diff
@@ -117,6 +119,9 @@ class profession
         std::vector<mtype_id> pets() const;
         std::vector<bionic_id> CBMs() const;
         std::vector<proficiency_id> proficiencies() const;
+        std::vector<matype_id> ma_known() const;
+        std::vector<matype_id> ma_choices() const;
```
ehughsbaird, line 123: I know you're just following the pattern of the rest, but why do these return copies? They could just as easily return `const` references.

Hymore246: Good question. There are bunch of methods like that. Maybe because they are vectors?

ehughsbaird: It's the right way to do it, https://stackoverflow.com/questions/15704565/efficient-way-to-return-a-stdvector-in-c, if you need to do something to populate a vector.
They're read-only in your code, resulting in copies unecessarily (it can't just move the vector from the class into the vector wherever you're calling it). It's not like this is a hot code path, so I suppose the efficiency if mutation is needed is potentially worth it?

PR #68549, opened 2023-10-07
Clarify what the modifier in derived damage means

ehughsbaird:
#### Summary
SUMMARY: None

#### Purpose of change
I didn't really understand what it did from reading the docs, and it made it seems like the modifier value was used for more than just armor, which is all I thought it was used for. Reading the code, that's all it is used for, but the documentation is not clear on that.

#### Describe the solution
Add some extra sentences to the docs clarifying.

I could word it differently, or make no reference to it potentially being used in different places in the future?

#### Testing
https://github.com/ehughsbaird/Cataclysm-DDA/blob/what-does-that-mean/doc/JSON_INFO.md#damage-types


ehughsbaird:
Turns out, I skipped the first sentence in the explanation!

PR #68559, opened 2023-10-08
Make Chloroform Rags unhealthy

Lupotech-CDDA:
#### Summary
Balance "Making Chloroform rags unhealthy"

#### Purpose of change
Chloroform is an unhealthy substance, that damages internal organs and is thought to be carcinogenic. It is not used as an anesthetic for these reasons. 

#### Describe the solution
I added a "healthy": -5, to the json.

I am not sure about how unhealthy it should be, but it is arguably more unhealthy than smoking a cigar, which has a -4 modifier, so I went with a -5.

### DIFF
--- a/data/json/items/comestibles/med.json
+++ b/data/json/items/comestibles/med.json
@@ -285,6 +285,7 @@
     "material": [ "cotton" ],
     "symbol": "!",
     "color": "red",
+    "healthy": -5,
     "flags": [ "NO_INGEST" ],
     "use_action": {
       "type": "consume_drug",


gettingusedto:
Wait a minute, isnt this a debug item? why would it matter that it is unhealthy or not?

Lupotech-CDDA:
> Wait a minute, isnt this a debug item? why would it matter that it is unhealthy or not? It's not like it's gonna appear in the game beyond debug spawn item menu anyway.

You're right, excuse me. Beginner's mistake I suppose.

PR #68560, opened 2023-10-08
The end of the Rabbit Saga

Karol1223:
#### Summary
None

#### Purpose of change
After #68516 and #68547 it is time to finish what has been started

#### Describe the solution
- Added a new rabbit kit abstract to shorten the JSON a little
- Slightly shortened the time it takes for cottontails to grow, since it was calculated proportionally from the wrong assumed time I used for domestic rabbit kits
- Added some more into the cottontail description to mention how the player isn't able to tell which species it is - there are two species of cottontail in NE, indistinguishable while alive
- Edited the "jackrabbit" to be named Snowshoe Hare, changed it's weight and made it copy from the rabbit abstract. Also made it reproduce
- Lowered the cottontail weight to be more accurate
- Removed rabbit spawning from the retirement community since it looked to have a 100% chance? It was kinda weird
- Added a new rabbit mongroup used in spawning. I will elaborate below how proportions have been calculated there
- Edited rabbit spawning to use the new mongroup

- I briefly considered adding a slightly different skull item that could be obtained from cottontails, as without genetic testing that'd be the only way for the player to tell the two species apart. Felt pointless, though

#### Additional context
- https://en.wikipedia.org/wiki/Eastern_cottontail - for weight of the Eastern Cottontail
- https://en.wikipedia.org/wiki/Snowshoe_hare ; https://animaldiversity.org/accounts/Lepus_americanus/ ; https://genomics.senescence.info/species/entry.php?species=Lepus_americanus for various information about the Snowshoe Hare
- https://www.inaturalist.org/observations?place_id=52339&quality_grade=research&taxon_id=43094&view=species for calculating proportions in which wild rabbits should spawn **NOTE**: iNaturalist is NOT perfect for this. This only accounts for individuals that observers have come into contact with, it does not represent the entire population accurately. I have taken the liberty to assume, however, that if people in better times have come into contact with roughly this spread of the species, then the character after the end of the world would have similar (if not worse) chances to come across the rarer species. Yes, this means that the domestic rabbit is insanely rarely going to appear in the wild now - as it should. It can still be reliably found in pet stores, animal shelters, zoos, (I think) farms, alongside Lapin and, as I added it into the pet mongroup, alongside Glooscap
- The NE Cottontail and Eastern Cottontail were treated as the same monster for the purpose of the game, as they are indistinguishable while alive and any mechanical differences that *could* be simulated, such as litter size, gestation length, maturation time or animal size - do not exist. They are not different enough to warrant being a separate entity.
- With this, all lagomorphs observed in NE have been added into the game (technically; NE Cottontail is an outlier and should not be counted). Any other species have not been observed on iNat, which by no means they aren't present but in my book makes them too rare to bother simulating.


data/json/mapgen/retirement_community.json
```diff
@@ -1741,7 +1741,6 @@
         { "group": "GROUP_STOCKED_POND", "x": [ 51, 61 ], "y": [ 7, 19 ], "repeat": [ 4, 5 ] },
         { "monster": "mon_squirrel", "x": [ 1, 23 ], "y": [ 1, 23 ], "repeat": [ 3, 4 ] },
         { "monster": "mon_squirrel", "x": [ 24, 47 ], "y": [ 50, 70 ], "repeat": [ 2, 3 ] },
-        { "monster": "mon_wild_rabbit", "x": [ 50, 70 ], "y": [ 50, 70 ], "repeat": [ 2, 3 ] },
```
ehughsbaird, line 1744: In my completely subjective experience, grassy places like this would be gauranteed to have rabbits.

data/json/monsters/mammal.json
```diff
@@ -2516,42 +2522,28 @@
...
-    "volume": "150 ml",
-    "weight": "30 g",
-    "hp": 1,
-    "speed": 175,
-    "attack_cost": 175,
-    "dodge": 8,
-    "//": "Sexually mature at 2-3 months. Domestic rabbits are 4-5 months and have a value of 45",
-    "upgrades": { "age_grow": 27, "into": "mon_wild_rabbit" },
-    "harvest": "mammal_tiny"
+    "upgrades": { "age_grow": 23, "into": "mon_wild_rabbit" }
```
ehughsbaird, line 2546: It might be nice to leave around the comment on how this was calculated, because it's not clear to me where the numbers for any of these come from.

Karol1223: It's basically days divided by 4. This is more of a matter of our documentation regarding monster evolution factor than anything here.

ehughsbaird: Oh, I forgot about evolution. You're right, that doesn't need documentation.

PR #68562, opened 2023-10-08
[Xedra Evolved] Spellcasting xp for Xedra Evolved

Maleclypse:
#### Summary
None

#### Purpose of change
Stealing the idea from #68555

#### Describe the solution
Copies their solution to the issue of powers not granting xp

None

#### Testing
tbd


Standing-Storm:
> @Standing-Storm thoughts on the best way to make sure that vampires and elementals also gain XP from casting their spells. I haven't been giving them classes because they are given only by mutations but maybe I should?

I think putting school on the spells is the best solution going forward. Then if other requests (like school-specific NO_SPELLCASTING etc) are implemented the infrastructure will already be there.

data/mods/Xedra_Evolved/mutations/paraclesians/arvore_mutation_spells.json
```diff
@@ -62,12 +62,15 @@
...
     "skill": "deduction",
+    "spell_class": "ARVORE",
+    "teachable": false,
     "effect": "effect_on_condition",
     "effect_str": "EOC_COMMUNE_WITH_NATURE",
     "shape": "blast",
     "flags": [ "CONCENTRATE", "SOMATIC", "VERBAL", "NO_LEGS", "NO_HANDS", "NO_EXPLOSION_SFX" ],
     "max_level": { "math": [ "per_to_level(1)" ] },
     "difficulty": 4,
+    "spell_class": "ARVORE",
```
Maleclypse, line 73:
```suggestion
```

Maleclypse:
I started looking into what it would take to allow spell_class to accept arrays.  And because of the way it's mostly used to create opposing classes, it looked to me that it would require a bunch of special casing and that it was more reasonable to make duplicate mutations and spells that look identical in game but aren't in the json.

GuardianDll:
fyi you can simply copy-from spells and just replace the class

data/mods/Xedra_Evolved/mutations/paraclesians/sylph_mutation_spells.json
```diff
@@ -37,6 +39,9 @@
     "type": "SPELL",
     "name": "Lightning Bolt",
     "description": "The goto ability for many Sylphs, this does just what you expect: you shoot lightning from your fingertips.  However, this lightning is more directed than most lightning, and travels in a line through most non-solid targets.",
+    "skill": "deduction",
```
Maleclypse, line 42:
```suggestion
```

data/mods/Xedra_Evolved/mutations/paraclesians/sylph_mutation_spells.json
```diff
@@ -75,6 +79,9 @@
     "type": "SPELL",
     "name": "Crash of lightning",
     "description": "The crash of lightning strikes your foes with a blast of electricity and a physical force that knocks them prone.",
+    "skill": "deduction",
```
Maleclypse, line 82:
```suggestion
```

Issue #68554, opened 2023-10-08
Gas mask cartriges are very short lasting, less than they are in real life.


MorvarchPrincess:
### Is your feature request related to a problem? Please describe.

Currently in game, Gas mask cartriges last a very short amount of time. I tested by standing in a gas vent, and my mask was losing 15 charges a minute of waiting, giving each cartridge about a 6m40s lifespan. Compare this to wikipedia, which mentions gasks masks lasting "around 8 hours" in a biological or chemical distuation, or this linked product page for a filter which describes a filter time of at least 20 minutes in worst case for chlorine gas.

https://en.wikipedia.org/wiki/Gas_mask

www.mirasafety.com/products/cbrn-gas-mask-filter-nbc-17-sof

I bring this up because especially in those instances where a gas mask is basically a neccesity (The collapsed tower for example) a gas mask lasts a frustratingly short time.

### Solution you would like.

In absolute worst conditions I think a single gas mask cartridge should last at least three times as long (20 minutes). Depending on how we want to assume the concentration and general nastiness of "toxic gas" and similar is, the cartridge could last up to multiple hours standing in heavy gas, but I'm assuming we want to lean on the more lethal end.

### Describe alternatives you have considered.

Leaving it as is.

Karol1223:
> Depending on how we want to assume the concentration and general nastiness of "toxic gas" and similar is, the cartridge could last up to multiple hours standing in heavy gas, but I'm assuming we want to lean on the more lethal end.

I don't think we can differentiate how different fields of gas affect the filter, so in this case we should assume the worst case scenario and go with the 20 minutes, I believe.

If I am wrong and we can determine this per-field type, that's an entirely different story and opens more possibilities to work with this, but I doubt that's the case.

ehughsbaird:
`gas_absorption_factor` https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md#field-types

The pitiful lifespan of gas masks is why I made https://github.com/CleverRaven/Cataclysm-DDA/pull/66913, fighthing through a collapsed tower without it was painful.

Karol1223:
> `gas_absorption_factor` https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md#field-types

I am in awe that this exists, not gonna lie

Maleclypse:
Closed by #68570

PR #68570, opened 2023-10-09
Better Gas masks - Model concentration better, and make gas mask cartiges last longer 

MorvarchPrincess:
#### Summary
Balance "Gas mask cartriges last quite a bit longer, and lower concentration gas takes less charges"

#### Purpose of change

Per issue https://github.com/CleverRaven/Cataclysm-DDA/issues/68554, our gas masks last a really short amount of time. I considered just about tripling the gas mask charge costs to bring it in line with estimates mentioned in the issue, but decided it would be better to have a concentration calculation in the charge buildup.

#### Describe the solution

My aim is to have an additional entry for gasses that is "concentration" that acts as a multiplier on charges spent, so thick toxic gas burns through your charges quicker than thin gas. I'll then go adjust numbers for gasses so that thick toxic gas means your gas mask will last about 20 minutes, and less problematic substances (read normal smoke) arn't as bad.

I also switched how the gas_absorption_factor works, it's now no longer what percent of a charge it uses per tick, its now how long a full 100 charge filter lasts in a concentration 1 gas.

Finally I also added another little message to the player when their gas mask charges go under 10. IRL a gas mask gets harder to breathe in as the filter clogs up, so it made sense for players to get a warning as they are running out.

I could just not add concentration, which would just require adjusting the gas absorption numbers, but it would still leave light gasses to unrealistically burn through charges.

I also considered basing the multiplier on intensity, but that doesn't work great as that number is primarily used for the amount of effect, and as such the numbers don't line up (toxic gas and thick toxic gas have the same number at 4) and could get modified and cause unintentional changes to gas mask charges.

Finally, I could have left gas absorption values the same, but it ended up with some very messy calculations so I decided it was better to change how the system works to be more based on how we measure gas mask lifetime in real life.

#### Testing

Tested both manually and via walking through code.

I booted up the game and spawned a gas vent in a small building. then waited till the room filled up. I then both watched the calculation via stepping through code, and confirmed it was taking the appropriate amount of charges off by waiting in the cloud with a mask on.

### DIFF
--- a/doc/JSON_INFO.md
+++ b/doc/JSON_INFO.md
@@ -5931,7 +5931,7 @@ Fields can exist on top of terrain/furniture, and support different intensity le
         "light_emitted": 2.5, // light level emitted by this intensity
         "light_override": 3.7 }, //light level on the tile occupied by this field will be set at 3.7 no matter the ambient light.
         "translucency": 2.0, // How much light the field blocks (higher numbers mean less light can penetrate through)
-        "concentration": 1, // How concentrated this intensity of gas is. Generally the thin/hazy cloud intensity will be 1, the standard gas will be 2, and thick gas will be 4. Affects how much gas mask charge is expended.
+        "concentration": 1, // How concentrated this intensity of gas is. Generally the thin/hazy cloud intensity will be 1, the standard gas will be 2, and thick gas will be 4. The amount of time a gas mask filter will last will be divided by this value.
         "convection_temperature_mod": 12, // Heat given off by this level of intensity
         "effects":  // List of effects applied to any creatures within the field as long as they aren't immune to the effect or the field itself
         [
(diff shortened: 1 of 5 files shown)


src/iuse.cpp
```diff
@@ -4183,14 +4184,26 @@ std::optional<int> iuse::gasmask( Character *p, item *it, const tripoint &pos )
...
             const field_entry &entry = dfield.second;
-            const int gas_abs_factor = entry.get_field_type()->gas_absorption_factor;
-            if( gas_abs_factor > 0 ) {
-                it->set_var( "gas_absorbed", it->get_var( "gas_absorbed", 0 ) + gas_abs_factor );
+            int gas_abs_factor = to_turns<int>( entry.get_field_type()->gas_absorption_factor );
+            const field_intensity_level &int_level = entry.get_intensity_level();
+            // 6000 is the amount of "gas absorbed" charges in a full 100 capacity gas mask cartridge.
+            // factor/concentration gives an amount of seconds the cartidge is expected to last in current conditions.
+            /// 6000/that is the amount of "gas absorbed" charges to tick up every second in order to reach that number.
+            float gas_absorbed = 6000 / ( ( float )gas_abs_factor / ( float )int_level.concentration );
```
ZhilkinSerg, line 4192:
```suggestion
            float gas_absorbed = 6000 / ( static_cast<float>( gas_abs_factor ) / static_cast<float>( int_level.concentration ) );
```

doc/JSON_INFO.md
```diff
@@ -5931,6 +5931,7 @@ Fields can exist on top of terrain/furniture, and support different intensity le
         "light_emitted": 2.5, // light level emitted by this intensity
         "light_override": 3.7 }, //light level on the tile occupied by this field will be set at 3.7 no matter the ambient light.
         "translucency": 2.0, // How much light the field blocks (higher numbers mean less light can penetrate through)
+        "concentration": 1, // How concentrated this intensity of gas is. Generally the thin/hazy cloud intensity will be 1, the standard gas will be 2, and thick gas will be 4. Affects how much gas mask charge is expended.
```
ehughsbaird, line 5934: How does it affect it?

MorvarchPrincess: Updated to be more clear

doc/JSON_INFO.md
```diff
@@ -5965,7 +5966,7 @@ Fields can exist on top of terrain/furniture, and support different intensity le
     "percent_spread": 90, // The field must succeed on a `rng( 1, 100 - local windpower ) > percent_spread` roll to spread. The field must have a non-zero spread percent and the GAS phase to be eligible to spread in the first place
     "phase": "gas", // Phase of the field. Gases can spread after spawning and can be spawned out of emitters through impassable terrain with the PERMEABLE flag
     "apply_slime_factor": 10, // Intensity of slime to apply to creatures standing in this field (Why not use an effect? No idea.)
-    "gas_absorption_factor": 15, // Amount of gasmask charges the field uses up per tick
+    "gas_absorption_factor": "80m", // Length a full 100 charge gas mask filter will last in this gas. Will be modified by the concentration of the gas, and should be 80m for concentration 1 toxic gas or similar. The worst gas should still be kept out for 20 minutes.
```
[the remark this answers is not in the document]
MorvarchPrincess: updated to be more clear

PR #68581, opened 2023-10-09
Balance tweaks for lieutenant 'Shadow'

RenechCDDA:
#### Summary
Balance "You will fear the night slightly less"

#### Purpose of change
There are some existing pain points related to the lieutenant zed, some parts where the design didn't work so well, some things I realized didn't work at all, and some assumptions that turned out to be wrong. This aims to fix all those, and make it a smoother, more consistent experience.

#### Describe the solution
Briefly listing all the changes and reasons for them...

Regeneration 5hp/turn --> 1hp/turn. With the speed differential being what it is, 5hp/turn allowed it to pretty consistently outheal the player chasing it. It is still desirable that it regenerates over a long enough fight, so 1hp/turn is the bare minimum.

Message changes: Obvious clarity pass. It helps the player know what's going on and gives it a bit more flavor.

FLIES: Was intended to let it avoid pits, but never saw the AI actually take flight so I assumed it was fine. It does actually fly in some situations, and that makes it completely unassailable. Obviously this just needs to go.

Bullet resistance 500-->100: While the player is still intended to engage with it more than just hose its general direction with gunfire, this allows some VERY high-end guns to do damage.

~~Casting time(for summon spell): default--->800 moves. With the speed scaling being what it is, a single summon of creatures could make you completely unable to catch it (you need to take time to avoid dying/smack some critters/etc). This gives it some 'backswing' so a particularly strong survivor still has time to close the distance.~~ This would be ideal but apparently it doesn't work!

Summon lifespan: min20s/max45s-->min15s/max30s. As above, these could simply harry the player too long. Additionally, if they lasted a long time this could allow as many as 3 (!!!) groups on the field at once, which is simply too many.

Group weights: Swarming amalgamations made about twice as likely. Because it spawns 6 of the same monster (as opposed to picking 6 times from the group) the larger spawns are exceptionally deadly. This simply tunes down how often that can happen by making the weakest option much more likely.

When shadow spawns: Time check randomized instead of like clockwork every 4 hours (now random between 3-6 hours), also increased the average so shadow spawns somewhat less often. Prevent it from spawning if the player happens to be sleeping. (Could remove the sleeping effect but stuff around sleep is so hardcoded I didn't want to touch it).

I would like to add a teleportation EoC that pops it away from you when it takes sufficient damage (allowing it time to cast spells/approach you from a different angle/etc), but there is an existing issue where location vars aren't playing nice with teleports, so I have not implemented that for the moment.

"Heavy duty" flashlights remain a pretty hard counter to it, since they are brighter and cause it to be affected by light even at its normal follow distance of 10. I have considered making it trail at a further distance, but then a default 8 perception character **cannot see it at that range**.I think the current approach where at longer ranges you might not be able to see it is still thematically/gameplay engaging, so even though it makes the monster "too easy to fight" sometimes it's still worth keeping this approach. Better than it be too easy (and engaging/fun) than the 'right amount' of difficult (and not fun).

#### Testing
I have only confirmed that the json loads without error.

### DIFF
--- a/data/json/monsters/zed_lieutenant.json
+++ b/data/json/monsters/zed_lieutenant.json
@@ -114,8 +114,6 @@
     "max_accuracy": 20,
     "min_range": 30,
     "max_range": 30,
-    "//": "A little bit of cast 'backswing' so that it can be caught in more situations.",
-    "base_casting_time": "800",
     "max_level": 1,
     "effect": "effect_on_condition",
     "effect_str": "EOC_LIEUTENANT_DRAIN_LIFE_TERRAIN"
--- a/data/json/monsters/zed_lieutenant.json
+++ b/data/json/monsters/zed_lieutenant.json
@@ -15,7 +15,7 @@
     "//6": "Slowed and weakened by light of sufficient brightness. (It lurks outside of flashlight range) Luring it into a trap is a viable path to killing it.",
     "speed": 200,
     "attack_cost": 200,
-    "regenerates": 5,
+    "regenerates": 1,
     "symbol": "L",
     "color": "red_white",
     "scents_tracked": [ "sc_human", "sc_fetid" ],
@@ -35,6 +35,7 @@
     "//1": "Try to stay this far away from the player with your regular moves, very important",
     "tracking_distance": 10,
     "path_settings": { "max_dist": 10, "allow_open_doors": true, "avoid_traps": true, "avoid_sharp": true, "allow_climb_stairs": true },
+    "//2": "TODO? Upgrade doesn't actually work because amalgamations do not have evolutions. Ideally this would upgrade them one step *and also remove their timed lifespan*. If this is added/resolved the spell cooldowns probably need to be nerfed.",
     "special_attacks": [
       [ "UPGRADE", 20 ],
       {
@@ -67,7 +68,7 @@
         "//": "Dirty little hack: We have a condition that the target is male OR female, meaning it only evaluates true for the player (and NPCs)",
         "condition": { "or": [ "npc_male", "npc_female" ] },
         "spell_data": { "id": "LTNT_attract_nearby", "min_level": 1 },
-        "monster_message": "Your senses are overwhelmed!",
+        "monster_message": "Your hearing is overwhelmed by a booming sound!",
         "cooldown": 30
       }
     ],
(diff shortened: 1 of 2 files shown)


data/json/monsters/zed_lieutenant.json
```diff
@@ -114,8 +114,8 @@
     "max_accuracy": 20,
     "min_range": 30,
     "max_range": 30,
-    "//": "14 moves worth of casting time. It tends to stick 10 tiles away from you, so this means you can get some attacks in if you chase it. Currently toggled off, because it's too easy to run this thing down if it pauses while you move right at it.",
-    "//base_casting_time": "2800",
+    "//": "A little bit of cast 'backswing' so that it can be caught in more situations.",
+    "base_casting_time": "800",
```
GuardianDll, line 118: fyi monsters do not use `base_casting_time`, they make their attack instantly, and are ruled solely by `cooldown` field

RenechCDDA: Really? That's annoying, but I might as well leave it in.

GuardianDll: I don't see the reason, it will do nothing but confuse people
```suggestion
```

CHCOOH:
This event is as unreasonable and dangerous as a car accident to the player who encounters it.
I might have appreciated it if cataclysms were not permadeath, but in a game where death is irreversible, I don't really appreciate this abruptness and unreasonableness.

The "how to deal with it" and "don't walk around at night or you won't encounter it" are only after the player has learned about it, so it can be a bad memory to meet this event and die at least once.
Also, the death of a fellow NPC can occur without the player's knowledge, or irreparable damage can be done if the base is in an orchard.

As a player, I would like to see this level of dangerous events not occur where the player's will does not intervene.
This is true not only for this event, but for anything that occurs or is encountered suddenly, such as portal storms.

FUNGUS and triffid dwellings are safe as long as players do not approach them.
In the same way, the event should only be triggered if the player's will intervenes.

In other words, we do not design "after the encounter", but rather "how the encounter occurs".
*Deepl Translation*

AcidAntOnAMinefield:
Yeah, came here to say that this BS monster showed up out of nowhere and essentially just screwed my character: trying to move from temp base to full base, avoiding predators and zombie wildlife and ran into this. 

What was the reason behind the creation of this creature anyway? What is there that this creature is necessary to "balance" it out?

Issue #51867, opened 2021-09-25
Weapon Categories


Termineitor244:
# Is your feature request related to a problem? Please describe.
Weapon categories are an useful concept made possible by #50212, every time there is a new implemented weapon, one needs to manually insert it in the weapon lists of the many martial arts there are, but with the categories, one would just need to add a category, and the martial arts would update themselves accordingly. This issue is mainly to organize the PRs I will be doing to make the martial arts actually use the feature.

# Weapon Categories guidelines:
1._ It should be re-usable by different martial arts if possible.
2._ It should include similar items.
3._ No weapon should remain in a weapon list, if the martial art has a theme, it can have categories for its weapons.

# List of weapon categories added until now:
| Weapon Category  | Martial Arts | Includes |
| ------------- | ------------- | ------------- |
| AUTOMATIC_RIFLES | Krav Maga  | All kind of automatic rifles  |
| AUTOMATIC_PISTOLS | Krav Maga  | All kind of automatic pistols  |
| KNIVES | Krav Maga, Ninjutsu, Eskrima, Silat  | All kind of knives  |
| BATONS | Krav Maga, Eskrima, Silat  | Batons, tonfas and weapons of similar shape  |
| MEDIUM_SWORDS  | Ninjutsu  | Swords of 81cm - 100cm length  |
| LONG_SWORDS | Ninjutsu  | Swords of 101cm length or more  |
| SHORT_SWORDS | Ninjutsu  | Swords of 80cm length or less  |
| QUARTERSTAVES | Ninjutsu, Silat, Bōjutsu  | Quarterstaves and similar weapons  |
| BLADED_FARMING | Ninjutsu, Eskrima, Silat  | Bladed farming tools and the machetes  |
| CLAWS | Ninjutsu, Eskrima, Bionic Combatives, Aikido, Judo  | Bionic and Tiger claws  |
| SHIVS  | Eskrima, Krav Maga  | Tiny, stabbing weapons like a glass shiv or monster's stings  |
| HOOKING_WEAPONRY | Fior Di Battaglia  | Long enough weapons with something to make a "Hook"  |
| POLEARMS  | Silat, Sōjutsu  | Polearms in general, including spears and the like  |
| FENCING_WEAPONRY  | Fencing  | The Foil, Épée, Sabre, and similar weapons like several batons  |
| BIONIC_WEAPONRY  | Bionic Combatives  | All kind of bionic installed weapons  |
[...]

PR #68582, opened 2023-10-09
Audit weapon categories for swords and sword-based martial arts.

Aerin-of-the-Toast:
#### Summary
Balance "Audit weapon categories for swords and sword-based martial arts"

#### Purpose of change

Sword weapon categories were a bit of a mess of redundancies, and the martial arts that used them were too restrictive in some ways and too permissive in other. 

Swords were arbitrarily lumped into cultural brackets for no reason, but only "Medieval Europe" "Japan" and "everywhere else" even though this isn't a useful distinction. Even if it was, Civil War sabers and a Middle Eastern Kilij were being lumped together in the "not medieval Europe or Japan" bracket.

Also, fencing techniques rely on the sword being able to thrust (or cut, with sabers). Similarly, Medieval Swordsmanship and Niten Ichi-Ryu rely on the weapons being able to cut. Ergo, the martial arts being usable with blunt weapons just because they're sword-shaped does not fit.

#### Describe the solution

There's a few major changes that were made.

Firstly, _no blunt "swords" or "fencing weaponry."_ All blunt weapons, including sword-shaped ones like the bokken or 2-by-sword had any sword-based weapon categories removed. This mostly affects fencing, which had _a lot_ of batons and similar bludgeons as fencing weapons. The reason for this is that, while you can _practice_ swordsmanship or fencing moves with a wooden sword, they do not apply if your goal is to kill something, since many of those moves rely on having a cutting edge or sharp point. If you want to kill someone with a cudgel or shillelagh, you don't use it to poke people in the chest or get light draw cuts like its a saber. You club the bajeezus out of your opponent with it.

Secondly: weapon categories now _only_ look at the weapon's function, and the _culture_ the sword originated from is ignored. A katana isn't something radically different just because it's from the east, it's just a curved sword. As such, MEDIEVAL_SWORDS and JAPANESE_SWORDS were entirely removed as weapon categories, as was CONSTRUCTED_SWORDS which was a pointless distinction to make that only existed so a 2-by-sword could be used by both eastern and western martial arts instead of requiring a separate 2-by-katana. Half-sword with your katana, _iaido_ with your kriegsmesser, it's a free apocalypse.

Thirdly, the categories themselves have been clarified. "SHORT_SWORDS" are one-handed cutting weapons 60cm or less, like machetes. "MEDIUM_SWORDS" are one-handed cutting weapons greater than 60cm. "LONG_SWORDS" are hand-and-a-half swords and "GREAT_SWORDS" are two-handed only swords. All of these require a cutting edge. As such, all swords can only be _one_ of these categories at a time, whereas previously there were quite a few that were multiple ("LONG_SWORDS" and "GREAT_SWORDS" like the longsword) or none. 

"FENCING_WEAPONRY" now specifies its only one-handed light, agile sabers, or straight agile thrusting swords. This means that the estoc is no longer a fencing weapon, as its a two-handed sword and not really applicable. To save the estoc from being weapon category-less, "LONG_THRUSTING_SWORDS" was added mostly just for the estoc, though it also applies to any suitable hand-and-a-half or two-handed thrusting sword like the longsword and _zweihänder_. Currently, only Medieval Swordsmanship uses "LONG_THRUSTING_SWORDS" and those weapons were already available for it regardless, so it's basically just for the estoc. If anyone find an estoc-focused martial art, there you go.

This PR also updates a few MA's in Mythical Martial Arts to the new standard (though I had to guess, I don't know Diamond Soul that well, heh), as well as Magiclysm weapons.

None

#### Testing

Game functions and weapons seem usable with the correct martial arts. It's possible I may have missed one or two; did you know the game doesn't even throw an error if a weapon has a weapon category that doesn't exist?

#### Additional context

The changes to shortswords vs medium swords means Eskrima and Silat lost a few weapons (like the wakizashi) but no one was using them with those anyway. _Grumbles in combat knife_

Otherwise, this shouldn't impact balance that much, other than it being a bit easier to find a valid weapon for your Medieval Swordsmanship/Niten Ichi-Ryu builds.

TL;DR sord is sord


Termineitor244:
It's good to see more love for the weapon categories! They were left pretty much as I added them some years back, so it's good to see someone reorganizing them.

Answering a couple of the points you present here:

> Firstly, no blunt "swords" or "fencing weaponry.

A valid argument, especially with the batons since their intended use in combat (and the reason for their melee damage) is to use them as, well, bludgeons, so it would not make much sense for them to apply all their damage when using fencing techniques, the argument for them being there was that they were pointed and long weapons that *could* be used for fencing (though nobody talked about how effective they would be at it). The weapons were in the style from long before the weapon categories were added.

I don't entirely buy the argument for the removal of blunt swords from the rest of the styles though, specifically with the wood sword and the bokken, they are not as effective as real sword, that's right, but they are used as training weapons for sword training, I would prefer for them to maybe only being able to use the basic buffs of the styles (standing, dodge positioning, etc.) And any applicable technique that would not require a cutting edge (Don't remember if there is any in the affected styles), leaving the meat of the styles, the killing techniques, reserved for the real swords.

> Secondly: weapon categories now only look at the weapon's function, and the culture the sword originated from is ignored.

I approve this! When those cultural categories were added, there was some debate about if the styles in question should be restrictive (using only culturally related weapons) or open for any kind of sword with a similar shape to their "core" weapons, I preferred for it to be open but I was convinced against it so that's how they came to be.

> Thirdly, the categories themselves have been clarified. 

Pretty much in line with my original vision for them, but better since they have more rules now! In #51867, they were conceived as: 80 cms or less, 81cm to 100cm, and 100cm or more, for short, medium and long swords respectively. Great swords (and hammers and axes) were added later as a Barbaran Montante only category, so thei didn't have so much of a connection with the rest of the sword categories (the reason why some weapons have two sword categories).

Pretty nice to see one/two handed being included in the definitions of the categories and a better organization for them.

Thanks for the changes! I will review the individual changes later when I have time.

data/json/items/melee/swords_and_blades.json
```diff
@@ -2506,7 +2503,7 @@
     "type": "TOOL",
     "category": "weapons",
     "name": { "str": "medium steel kriegsmesser" },
-    "description": "A single-edged hand-and-a-half-sword of German design, with a long, curved blade.  It became popular across Europe due to its low cost while maintaining strong cutting power in comparison to the more sophisticated longsword.\n\nMade with medium steel and with a grip made of leather wrapped in steel wire, this weapon is comfortable in your hand, though the balance isn't perfect and the edge could be sharper.",
+    "description": "A single-edged two-handed sword of German design, with a long, curved blade.  It became popular across Europe due to its low cost while maintaining strong cutting power in comparison to the more sophisticated longsword.\n\nMade with medium steel and with a grip made of leather wrapped in steel wire, this weapon is comfortable in your hand, though the balance isn't perfect and the edge could be sharper.",
```
TheShadowFerret, line 2506: Needs `"ALWAYS_TWOHAND"`

Aerin-of-the-Toast:
> I don't entirely buy the argument for the removal of blunt swords from the rest of the styles though, specifically with the wood sword and the bokken, they are not as effective as real sword, that's right, but they are used as training weapons for sword training, I would prefer for them to maybe only being able to use the basic buffs of the styles (standing, dodge positioning, etc.) And any applicable technique that would not require a cutting edge (Don't remember if there is any in the affected styles), leaving the meat of the styles, the killing techniques, reserved for the real swords.

It's certainly complicated, but that's more of a martial-arts-wide issue. A HEMA practitioner can totally use a bokken better than someone who isn't one, but the techniques wouldn't be exactly the same, and you could make the same argument for, say, a baseball bat. I've had ideas for making more martial arts (especially unarmed ones) usable in *some* capacity with "wrong" weapons but that's a larger project for another time.

For now where it's more all-or-nothing I don't think a blunt object being sword-shaped affects its use as a weapon much, you're still just clubbing people with it in the end. For now, a HEMA practitioner will be able to use a bokken better than someone else because they'll just start with higher melee/dodge/etc. skills instead.

Termineitor244:
> I've had ideas for making more martial arts (especially unarmed ones) usable in some capacity with "wrong" weapons but that's a larger project for another time.

Reasonable enough! I understand if it's out of scope for this PR. I will see if I can contribute in the weekend with that project by re-enabling the weapons in a lesser capacity when appropriate (mainly just the positioning buffs for the martial arts, since they are the ones less affected by the cutting edge of your weapon).

keampe:
As someone who has taught German Swordfighting, I think I should weigh in here.

First the good, and it's very good. Swords that are similar in shape and balance can indeed be used by more than one martial art. Wouldn't be all that hard to modify techniques to use something that's a little different, either. Your changing of the weapon groups is very good.

Now, the not so good. Removing blunt versions of the weapons - Huh? We train with those, we spar with those full contact and I will say with some experience that without armour getting hit with one of these is very dangerous. We use the same techniques with them too. The "3 wounders" of this style are the "stitch" or thrust, the "Shnit" or slice and the "haw" or strike. Of these only the shnit would be ineffective with a blunt. Remember that a thrust from a quarterstaff is considered a lethal technique - it concentrates the force of your bodyweight going forward onto a hard surface much smaller than a fist. This can easily break ribs, fracture skulls or close windpipes. A thrust from a waster (wooden sword) does the same. If you want a real world example look up Miyamoto Musashi. He killed numerous people with a boken, using his form of kenjutsu.

We practice the same techniques with them. Even if one were to argue - as you have, though I disagree - that you can't use the same techniques with effectiveness, we still train with them and therefore know exactly how to use them. They should be included in the martial art.

As to the batons not being used by fencing, it depends. If you are talking modern sport fencing then yes, they don't really have a place there. If, however, Fencing refers to the older style that included period sabers (not the steel whip that's called a saber these days) then batons were used as training tools.

Just some food for thought.
   - Shane

fairyarmadillo:
Cavalry sabers are not fencing weapons. The confusion comes from there being two kinds of saber.

The first is a straight, heavyweight thrusting blade which was originally used by infantry and for non-mounted civil defense. That's a fencing saber, and survives today as a category of fencing sword.

A cavalry saber is like a civil war sword - the curved kind designed for cutting. These are much heavier, and owing to their curved design are not suitable for fencing as the sport is practiced today.

Blunt swords shouldn't be removed from the MAs IMO. People use bokken to practice kendo and kenjitsu and the techniques are mostly still applicable even without an edge.

fairyarmadillo:
Fencing sabre - https://en.m.wikipedia.org/wiki/Sabre_(fencing)

Cavalry - https://en.m.wikipedia.org/wiki/Model_1840_Cavalry_Saber

You can see here that despite the similar names, theyre not even close in design.

Maleclypse:
So I've merged this as I think it's an improvement.  However, I think that as an example fencing swords should probably be split out into modern fencing weapons vs historical fencing weapons. 

Honestly I think most of mine and others criticisms would be better served by adding an optional field to techniques that check the damage types of a weapon and if the weapon lacks equivalent damage values then that technique is no longer selected during combat.  This might be the wrong approach though and either way I'm not the person who would be implementing that anyway.

Termineitor244:
> So I've merged this as I think it's an improvement.  However, I think that as an example fencing swords should probably be split out into modern fencing weapons vs historical fencing weapons. 
> 
> Honestly I think most of mine and others criticisms would be better served by adding an optional field to techniques that check the damage types of a weapon and if the weapon lacks equivalent damage values then that technique is no longer selected during combat.  This might be the wrong approach though and either way I'm not the person who would be implementing that anyway.

That would be an option, but personally, I think it would be better to just use what is already in place and just prohibit certain weapon categories (In the case of fencing it would be a new group, "Blunt Fencing" or whatever) from working with the techniques in question, using the same system that makes the master sword be a better weapon inside the hylian swordmanship martial art, enabling techniques only for the best/most apropiate weapons for the style, and giving a bare bones version of the style for weapons that are suitable for a part of it, but not for every technique.

Aerin-of-the-Toast:
Re:sabers in fencing:
 I'm not very familiar with fencing myself. But when looking at sport fencing rules with a saber I saw they do count "slashing" hits (touches with the "edge" of the blade). To me this reads as fencing should work with any light, agile one-handed slashing weapon of sufficient length, even if fencing sabers are not literally civil war sabers the techniques likely should carry over (but not with heavier slashing weapons like a falchion or katana), since presumably the fencing martial art is including saber fencing.

Re: blunt stuff in MS
I should reiterate again that what's used for practice is irrelevant compared to what works in a real combat situation. You can practice with a mop if you wanted. 
Though I'm willing to accept some or even most techniques may still be applicable even with blunt sword-esque weapons, so Terminaitor's idea of giving them a sort of "partial" applicability similar to MMA's Hylian Swordsmanship sounds like an idea I could pursue. I wouldn't limit it to just things like the wooden sword though, I'd probably make a new "long blunt" category for longer bludgeons (too long to be a baton, too short to be a staff, and not top-heavy enough to be a great hammer) with the idea that anything which works with a bokken likely would work the same with a baseball bat.

fairyarmadillo:
> Re:sabers in fencing: I'm not very familiar with fencing myself. But when looking at sport fencing rules with a saber I saw they do count "slashing" hits (touches with the "edge" of the blade). To me this reads as fencing should work with any light, agile one-handed slashing weapon of sufficient length, even if fencing sabers are not literally civil war sabers the techniques likely should carry over (but not with heavier slashing weapons like a falchion or katana), since presumably the fencing martial art is including saber fencing.

Saber fencing is using fencing sabers, not cavalry sabers. It is impossible to make the same motions with a fencing saber as with a cavalry saber, you cannot use it to fence with because instead of being an ultra thin and light blade, it is a very heavy one unsuited to thrusting.

I don't think your logic holds, and it's also totally arbitrary. Any number of swords should qualify by your reasoning. The two sabers are nothing alike and share only a name.

keampe:
> Re: blunt stuff in MS I should reiterate again that what's used for practice is irrelevant compared to what works in a real combat situation. You can practice with a mop if you wanted. Though I'm willing to accept some or even most techniques may still be applicable even with blunt sword-esque weapons, so Terminaitor's idea of giving them a sort of "partial" applicability similar to MMA's Hylian Swordsmanship sounds like an idea I could pursue. I wouldn't limit it to just things like the wooden sword though, I'd probably make a new "long blunt" category for longer bludgeons (too long to be a baton, too short to be a staff, and not top-heavy enough to be a great hammer) with the idea that anything which works with a bokken likely would work the same with a baseball bat.
<
You have some misconceptions here. A major one is that a baseball bat has a totally different distribution of mass than a sword - it's very tip heavy. The techniques for it would be much closer to how you use an axe than a sword. Trying to do a proper Zornhau (Wrath Strike) would be problematic. A Zornhau is a diagonally downward strike to the opponent's sword with a thrust to the face immediately upon deflection of said sword. A baseball bat would have difficulty changing directions like that - a wooden sword would not. This is just one example, there are many others.

Your idea for a "long, blunt" category does have some merit, though. I do know the Myomoto Musashi used a boat oar to kill someone. I also know that a quarterstaff can be used with Zweihander techniques in addition to being used to train spear and pike techniques.

As to fencing - I think we may need to separate it into 2 different martial arts - Olympic (or Modern) and Historical. The problem here is that "Fencing" refers to more than one art. It's almost like saying "European Martial Art with a sword" and will mean different things to different people. When you say "Saber Fencing" do you mean the Olympic Sport or something taught to cavalry officers which would also likely be called fencing.

Just some thoughts
   - Shane

fairyarmadillo:
Fencing is already specifically the modern sport style.

keampe:
So we need a Historical variety, then. 

   - Shane

Malorn-Deslor:
Just a note, I think your arguments about fencing are very accurate, but I don't think they hold up with many other styles, including Medieval. Bokken/blunt swords work very well and can do huge amounts of damage with the styles involved. There are HEMA people who have tried fighting without armor with wooden swords, it generally ends with a lot of broken bones and a surprising number of gashes for the swords being 'blunt'.

Issue #67709, opened 2023-08-16
Enemies can occupy the same space as the player


ilovehans10:
### Describe the bug

Using either the shipwrecker spell from Magiclysm and the far hand power from Mind Over Matter, it is possible to make an enemy stand in the same space as the player. You can also make enemies stand on walls with force shove, but that might be intended. This will be resolved the next game turn, but the first time that it happened to me, I worried for a split second that I might have just gotten telefragged. I also haven't seen any other times that two entities occupy the same space, which also makes this feel like it isn't intended.

I assume this is something wrong with directed_push or possibly area_push (which seems to just use directed_push). It would also be nice if similar to how bullets are shown moving their trajectory if when enemies are pushed their full travel was shown, so it is more clear where an enemy goes when moved.

### Attach save file

NA

### Steps to reproduce

1. Give player either shipwrecker or far hand spells from Magiclysm and Mind Over Matter
2. Use the spell a few times when near zombies or other hostiles (I haven't tested with other entities)
3. Observe that sometimes zombies will teleport into your tile.

### Expected behavior

Enemies are thrown next to the player if anywhere

### Screenshots

I will get some when I get back to my computer

### Versions and configuration

- OS: Linux
    - OS Version: LSB Version: n/a; Distributor ID: Arch; Description: Arch Linux; Release: rolling; Codename: n/a; 
- Game Version: 0.G-3439-gc5ecac002f-dirty [64-bit]
- Graphics Version: Tiles
- Game Language: System language []
- Mods loaded: [
    Dark Days Ahead [dda],
    Disable NPC Needs [no_npc_food],
    Portal Storms Ignore NPCs [personal_portal_storms],
    Slowdown Fungal Growth [no_fungal_growth],
    Stats Through Skills [StatsThroughSkills],
    Stats Through Kills [stats_through_kills],
    SpeedyDex [speedydex],
    Bionic Professions [package_bionic_professions],
    Aftershock [aftershock],
    Blaze Industries [blazeindustries],
    Magiclysm [magiclysm],
    Extra Mutated Scenarios [extra_mut_scens],
    sees-player, run, crouch, bleed icons, retrodays [sees_player_retro]
]

### Additional context

I am away from my computer at the moment so my version is from an older bug report, but I had the bug happen back then too. I also know that in my versions/configuration section it doesn't include the MOM mod, but I added that mod after and then had the bug happen.

[...]

Issue #63185, opened 2023-01-19
Dragging a cart while swapping with friendly animal puts cart under you


Brambor:
### Describe the bug

Position C: Cart P: Player, A: Animal
`CPA`
_Cart to the west of Player who's to the west of Animal._

I think that the Player moves east, and drags the Cart east.
At the same time, player is trying to swap with the Animal, so that the Player would be at A and the Animal would be at P.
This fails, because Cart is already at P, so Animal and the Player don't move.
However the Cart already moved to P, so now both the Player and a Cart are at P.
Cart is IMPASSABLE, so the Player shouldn't be able to occupy the same tile.
The Player stops dragging the Cart.

### Steps to reproduce

Dragging an impassable Cart try to go into a friendly chicken's tile
All happening on a straight line, either west-east or diagonal.
`CPA -> ??P`
_Cart Player Animal; Player tries to go to the Animal's tile while dragging the Cart._

### Expected behavior

A solution where two entities don't occupy the same tile.
Probably: "You cannot swap with the chicken dragging cart like that." possibly adding: "Where would the chicken go?"

Important note: these configurations work correctly right now and the solution shouldn't break them:
```
AP <-> PA
 C      C
```
_Animal west of Player with Cart to the south of either of them. The Player can swap with the Animal, the Cart doesn't move._

### Versions and configuration

Like 2023-01-18 experimental (not @ PC right now).

Raikiri:
Maybe related, but swapping with mechs allows putting them into vehicles. Note that you're explicitly forbidding from entering vehicles while riding a mech. Not sure if it's a bug or a feature.

PR #68590, opened 2023-10-10
Bugfix directed_push with negative damage

### LINKED ISSUE #62298: ``directed_push`` spells with negative damage will cause extremely weird behaviour when used on monsters
### Describe the bug

It's best if you look in the screenshots first

So the flies there? There's only one of them. Only one is a real frigate fly. Somewhere under them is a twisted hive (a monster using said spell, for a PR I was working for, the PR is #62287). The hive pulled the frigate fly in and it started progressively duplicating said fly.

Now, these flies do not move, do not attack (at least I think they don't, one used acid but I think it was the real one while it was still alive) and cannot be attacked (there are no messages about missing but they don't get hit either). They all disappear once the hive (i.e. the monster that caused this with the spell) is killed, however.

Initially, I thought this only happened when the spell is used on a monster adjacent to the caster but I am not convinced that's the case anymore. I can't, however, tell whether it is or isn't. Things got out of hand fast here.

Additionally, when the spell is used on the player (or a monster) they can be pulled onto the same tile as the caster. Which, well, not ideal in itself either.

### Steps to reproduce

1. Copy the changes from #62287
2. Spawn a twisted hive
3. Spawn any monster that's not from the zombie faction
4. Wait a bit until the hive can use its pull spell
5. Notice that the monsters at one point or another start duplicating

As an additional note I'd advise to use slower monsters. I'm not sure what would happen if the monsters were to manage to run away faster than they're pulled.

### Expected behavior

The monsters shouldn't creature indestructible duplicates. And the player or monster shouldn't be pulled on top of the caster either.

### Screenshots

obraz

- OS: Windows
    - OS Version: 10.0.19045.2251 (22H2)
- Game Version: aa2a276 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Standard combat test suite [standard_combat_test]
]

### Additional context

_No response_

marimarigi:
#### Summary
None

#### Purpose of change
Fixes #62298,#67709

#### Describe the solution
Record the coordinates of the previous step, and if there is an obstacle, finish moving at those coordinates.
Also, creatures are now considered obstacles.

#### Testing
I confirmed this using Mind Over Matter's far hand, and a monster that uses similar magic.

### DIFF
--- a/src/magic_spell_effect.cpp
+++ b/src/magic_spell_effect.cpp
@@ -963,13 +963,18 @@ static void character_push_effects( Creature *caster, Character &guy, tripoint &
                                     const int push_distance, const std::vector<tripoint> &push_vec )
 {
     int dist_left = std::abs( push_distance );
+    tripoint old_pushed_point = guy.pos();
     for( const tripoint &pushed_point : push_vec ) {
         if( get_map().impassable( pushed_point ) ) {
             guy.hurtall( dist_left * 4, caster );
-            push_dest = pushed_point;
+            push_dest = old_pushed_point;
+            break;
+        } else if( get_creature_tracker().creature_at( pushed_point ) ) {
+            push_dest = old_pushed_point;
             break;
         } else {
             dist_left--;
+            old_pushed_point = pushed_point;
         }
     }
     guy.setpos( push_dest );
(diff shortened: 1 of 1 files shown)


Brambor:
You wrote
Fixes #62298,#67709

It should have been
Fixes #62298
Fixes #67709

Your way didn't close #67709

Brambor:
Also, could this fix #63185 ?

marimarigi:
@Brambor 
> Also, could a similar approach to this fix #63185?

The problem probably only occurs when there are pets at the destination, so it's possible to detect this and abort the move. However, I think a better solution would be to automatically do something similar to when pushing an NPC away. (And this is difficult, at least for me.)

PR #68598, opened 2023-10-10
fix: basic ferals mostly share drops with zombies

### LINKED ISSUE #68512: Basic feral human drops are quite sparse

I noticed how the basic feral humans' drops are quite sparse than the basic zombies' drops. You would think they still have their stuff prior to feralizations, but nope, it's all empty pockets and no backpack on them.

Unless the things feral humans did after they gone feral is to empty out their pocket and ditch their backpack, i don't buy this lack of stuff on their person.

### Solution you would like.

Give them the loot akin to what basic zombies has.

Not doing so.

### Additional context

_No response_

casswedson:
#### Summary
None

#### Purpose of change
resolves: #68512
make regular ferals share mostly the same drops as the regular zombies, clothes, items etc

also remove itemgroups related with events from the defaul zed drops, they don't work, at least zed spawn wit barely anything if I leave those lines there

#### Testing
Screenshot from 2023-10-10 18-02-39
Screenshot from 2023-10-10 18-19-57
Screenshot from 2023-10-10 18-20-02

### DIFF
--- a/data/json/monsterdrops/feral_humans.json
+++ b/data/json/monsterdrops/feral_humans.json
@@ -13,14 +13,17 @@
           { "item": "crash_axe", "prob": 5, "damage": [ 1, 3 ] }
         ]
       },
-      { "group": "default_zombie_clothes", "prob": 100 }
+      { "group": "default_zombie_death_drops", "prob": 100, "damage": [ 1, 3 ] }
     ]
   },
   {
     "type": "item_group",
     "subtype": "collection",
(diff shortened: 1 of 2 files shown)


Termineitor244:
> also remove itemgroups related with events from the defaul zed drops, they don't work, at least zed spawn wit barely anything if I leave those lines there

They don't?! D: But they are part of a distribution, and the `event` field should only count them when in the actual holiday... There are no other members of the group so I don't know what the problem is...

PR #68599, opened 2023-10-11
Try to fix ASAN

ehughsbaird:
#### Summary
SUMMARY: None

#### Purpose of change
Fix ASAN test

#### Describe the solution
AFAICT this was introduced in the the PR adding the modified lines, and this was the most suspect there.
https://github.com/CleverRaven/Cataclysm-DDA/actions/runs/6471151465/job/17569187800#step:17:135
https://github.com/CleverRaven/Cataclysm-DDA/pull/68548/

#### Testing
None, no compiler handy right now.

### DIFF
--- a/src/newcharacter.cpp
+++ b/src/newcharacter.cpp
@@ -85,6 +85,8 @@ static const flag_id json_flag_no_auto_equip( "no_auto_equip" );
 
 static const json_character_flag json_flag_BIONIC_TOGGLED( "BIONIC_TOGGLED" );
 
+static const matype_id style_none( "style_none" );
+
 static const profession_group_id
 profession_group_adult_basic_background( "adult_basic_background" );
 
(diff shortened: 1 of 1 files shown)


ehughsbaird:
Looks like it fixes ASAN. Other failure looks unrelated.

src/newcharacter.cpp
```diff
@@ -921,9 +921,7 @@ void avatar::initialize( character_type type )
     }
 
     std::vector<matype_id> all_styles = martial_arts_data->get_known_styles( false );
-    int num = rng( 0, all_styles.size() );
-    const matype_id rand_style = all_styles[num];
-    martial_arts_data->set_style( rand_style );
+    martial_arts_data->set_style( random_entry( all_styles ) );
```
akrieger, line 924: Maybe this?

```suggestion
    martial_arts_data->set_style( random_entry( all_styles, style_none ) );
```

And then add somewhere at the top

```
static const matype_id style_none( "style_none" );
```

akrieger: The reason being `random_entry` returns a default constructed `V` but the default `matype_id` is not the empty string but `"style_none"`, looking at eg. `character_martial_arts::reset_style()`.

PR #68608, opened 2023-10-11
Info msg when melee skill is too low to see melee values

RenechCDDA:
#### Summary
Interface "Melee weapons tell you your skill is too low to see melee values instead of just hiding them"

#### Purpose of change
Was testing relic enchantments for ATTACK_SPEED and was very confused I could not see adjusted moves per attack on the item's info panel. Turns out the entire section is just hidden(not added) if you don't have enough melee skill.

#### Describe the solution
Still insert the section, but only with a blurb telling you the character has too little skill.

#### Testing
image

### DIFF
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -5429,6 +5429,13 @@ void item::melee_combat_info( std::vector<iteminfo> &info, const iteminfo_query 
                                iteminfo::lower_is_better, attack_cost );
         }
         insert_separation_line( info );
+    } else if( player_character.get_skill_level( skill_melee ) < 3 ) {
+        insert_separation_line( info );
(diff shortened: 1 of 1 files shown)


GuardianDll:
shouldn't `Typical damage per second` and `Typical damage per stamina` have the same rules?

RenechCDDA:
> shouldn't `Typical damage per second` and `Typical damage per stamina` have the same rules?

Probably, but I wasn't going to poke that hornet's nest.

RenechCDDA:
OK, pretty sure this is good to go now. Tests (including the one specifically failing, iteminfo) passed locally and under draft.

PR #68614, opened 2023-10-12
Add door_locks to prisoner, armored and luxury vehicles

stubkan:
#### Summary
Features "Add door_locks to prisoner, armored and luxury vehicles"

#### Purpose of change

Added door_lock definitions to doors on vehicles that would have a high likelihood of having strong locks. 

Door locks were added some time ago, but as of yet are not naturally spawning anywhere and can only be crafted from scratch.  Not sure why, if there is a reason, I do not know it.

They should be added to at least a few vehicles, so people can start to use the system.  So people can discover this semi-hidden feature exists, scavenge locks and install them - and then over the course of natural use - future bugfixes or features can be expanded on after the userbase starts to use them.

#### Describe the solution

Added door_lock defs to various vehicles - the following vehicles in specific ;

* Armoured transport trucks
* Limousenes
* Luxury RVs
* Prisoner transport vehicles

I did not add to any police or military vehicles, as they are quite common and I do not think they have any security aside from your everyday normal locking.

N/A

#### Testing

Spawned several of the modified vehicles, and observed the locking system being present and functional.

#### Additional context

image


ehughsbaird:
Are these meant to be special strong locks, or are they just ordinary door locks? They'd probably make sense to go on every car if they aren't special.

stubkan:
> Are these meant to be special strong locks, or are they just ordinary door locks? They'd probably make sense to go on every car if they aren't special.

From the item description itself ; "Some wires and gears for connecting door handles and an unlocking knob to a latching mechanism, and a sturdy lock to go on the outer surface. It is meant to be installed in a vehicle door or hatch."

So, two answers to that - The item appears to be 'extra strong' locking, in its description.

Secondly, locks can be made valuable and sought out the same way door motors are rare ingame, while basically every car irl has them. Making it somewhat scarce - would serve the gameplay purpose of making it an upgrade one actively seeks out.

ehughsbaird:
I think it would make for fun gameplay choices to have car doors be locked, but these are are clearly not normal car door locks.

stubkan:
Yes, once you find them on the provided vehicles you can salvage and install them on your vehicles.  Or make your own with devices skill and put them on your vehicle.  So you do get to use them in your cars - this is just a PR to add them to some vehicles.

Terrorforge:
> I did not add to any police or military vehicles, as they are quite common and I do not think they have any security aside from your everyday normal locking.

I just want enter into the record the fact that military vehicles generally [don't even have ignition keys](https://www.quora.com/Do-military-vehicles-have-keys). It's often possible to lock the doors from the inside, but they don't have integrated door locks, either.

Which makes sense when you think about it, because losing your keys in a warzone would be both extremely easy and extremely bad, and the fact that they're typically full of guys with guns is a pretty effective theft deterrant anyway.

TheMurderUnicorn:
So I guess it's too late since this was merged but I'm going to leave a comment anyway.  Last I looked at vehicle locks, the only way to circumvent them was with lockpicking which lets be honest is sort of terrible.  I am firmly against vehicle locks without another method to get past them.  I do not think these should be getting added until the player has more ways to interact with them -- we can't even smash vehicle windows to gain access.  So unless I invest time into lockpicking, locked vehicles will just be inaccessible?  Or is there a mechanism I'm not familiar with?

Terrorforge:
They don't actually spawn locked, but even if they did you could still smash the doors or use the vehicle modification interface to remove the locks.

PR #68616, opened 2023-10-12
Cattail shouldn't be plantable on land

harakka:
#### Summary
Bugfixes "Make cattail non-plantable"

#### Purpose of change
Cattail is a water plant, and shouldn't be growable on land . We don't have a way to define what terrain a seed is plantable on, so currently cattails can just be farmed like any other crop. This removes plantability from cattail seeds.

#### Describe the solution
Remove `PLANTABLE_SEED` from cattail seeds.

### DIFF
--- a/data/json/items/comestibles/seed.json
+++ b/data/json/items/comestibles/seed.json
@@ -265,7 +265,7 @@
     "name": { "str_sp": "cattail seeds" },
     "color": "green",
     "description": "Some cattail seeds.",
-    "flags": [ "PLANTABLE_SEED", "NUTRIENT_OVERRIDE", "INEDIBLE" ],
+    "flags": [ "NUTRIENT_OVERRIDE", "INEDIBLE" ],
     "seed_data": { "plant_name": "cattail", "fruit": "cattail_stalk", "byproducts": [ "cattail_rhizome" ], "grow": "91 days" }
   },
   {


AmarinReyny:
Out of curiosity, what is the expected timeframe for enabling seeds to have definable plantable terrain?

Aerin-of-the-Toast:
> Out of curiosity, what is the expected timeframe for enabling seeds to have definable plantable terrain?

Literally whenever someone decides to code that. Could be tomorrow or thirty years from now, the only way to guarantee it happens soon is to do it yourself if you wanted.

AmarinReyny:
> > Out of curiosity, what is the expected timeframe for enabling seeds to have definable plantable terrain?
> 
> Literally whenever someone decides to code that. Could be tomorrow or thirty years from now, the only way to guarantee it happens soon is to do it yourself if you wanted.

So, nobody's working on it right now, then? That probably would have been a better question for me to ask anyway... Sorry for that.

ThePainkiller:
This could've been waited out until a proper PR to define something like shallow water as a plantable terrain was pushed.

harakka:
If you write that PR, I'll be happy to add the relevant fields to water-plantable plant seeds. In the meanwhile, cattail isn't plantable on land, and all we have is land planting.

PR #68619, opened 2023-10-12
fixed broken salt water pipe

### LINKED ISSUE #66318: Basecamp expansion Workshop 2 Saltwater pipe crash
### Describe the bug

When trying to construct the first phase of a saltwater pipe the game terminates.

Callstack when debugging:
 	cataclysm-tiles.exe!std::string::size() Line 4276	C++
 	cataclysm-tiles.exe!std::hash<std::string>::_Do_hash(const std::string & _Keyval) Line 5312	C++
 	cataclysm-tiles.exe!std::_Conditionally_enabled_hash<std::string,1>::operator()(const std::string & _Keyval) Line 2170	C++
 	cataclysm-tiles.exe!cata::hash_combine<std::string,std::hash<std::string>>(unsigned __int64 & seed, const std::string & v, const std::hash<std::string> & hash) Line 21	C++
 	cataclysm-tiles.exe!cata::tuple_hash::operator()<std::string const ,cata_variant>(const std::pair<std::string const ,cata_variant> & v) Line 58	C++
 	cataclysm-tiles.exe!cata::hash_combine<std::pair<std::string const ,cata_variant>,cata::auto_hash<std::pair<std::string const ,cata_variant>>>(unsigned __int64 & seed, const std::pair<std::string const ,cata_variant> & v, const cata::auto_hash<std::pair<std::string const ,cata_variant>> & hash) Line 21	C++
 	cataclysm-tiles.exe!cata::range_hash::operator()<std::unordered_map<std::string,cata_variant,std::hash<std::string>,std::equal_to<std::string>,std::allocator<std::pair<std::string const ,cata_variant>>>>(const std::unordered_map<std::string,cata_variant,std::hash<std::string>,std::equal_to<std::string>,std::allocator<std::pair<std::string const ,cata_variant>>> & range) Line 76	C++
 	cataclysm-tiles.exe!std::hash<mapgen_arguments>::operator()(const mapgen_arguments & args) Line 35	C++
 	[External Code]	
>	cataclysm-tiles.exe!basecamp_action_components::choose_components() Line 856	C++
 	cataclysm-tiles.exe!basecamp::common_salt_water_pipe_construction(const mission_id & miss_id, expansion_salt_water_pipe * pipe, int segment_number) Line 2662	C++
 	cataclysm-tiles.exe!basecamp::start_salt_water_pipe(const mission_id & miss_id) Line 2849	C++
 	cataclysm-tiles.exe!basecamp::start_upgrade(const mission_id & miss_id) Line 1826	C++
 	cataclysm-tiles.exe!basecamp::handle_mission(const ui_mission_id & miss_id) Line 1521	C++
 	cataclysm-tiles.exe!iexamine::bulletin_board(Character & you, const tripoint & examp) Line 1931	C++
 	cataclysm-tiles.exe!map_data_common_t::examine(Character & you, const tripoint & examp) Line 578	C++
 	cataclysm-tiles.exe!game::examine(const tripoint & examp, bool with_pickup) Line 6084	C++
 	cataclysm-tiles.exe!game::examine(bool with_pickup) Line 5940	C++
 	cataclysm-tiles.exe!game::do_regular_action(action_id & act, avatar & player_character, const std::optional<tripoint> & mouse_target) Line 2220	C++
 	cataclysm-tiles.exe!game::handle_action() Line 3014	C++
 	cataclysm-tiles.exe!do_turn() Line 538	C++
 	cataclysm-tiles.exe!WinMain(HINSTANCE__ * __formal, HINSTANCE__ * __formal, char * __formal, int __formal) Line 826	C++
 	[External Code]	

The last sane part of the call chain (before the hash juggling mumbo jumbo) was:
cataclysm-tiles.exe!basecamp_action_components::choose_components() Line 856
There args_ has the value of {map= {size=0}}, which doesn't seem reasonable, and is probably what causes the code to blow up (although the "find" function should have handled such a situation).

### Attach save file

N/A

### Steps to reproduce

1. Create the workshop version 2 expansion (in a location reasonably close to a swamp, allowing for a pipe to be constructed).
2. Spawn the equipment for making the saltwater pipe.
3. Order the pipe to be constructed and verify that you accept that X additional jobs are required to finish it.
4. Have the game terminate (or caught on an exception if debugging (at least when compiled with debugging: haven't tested debugging release mode))

### Expected behavior

It would still work...

### Screenshots

_No response_

- OS: Windows
- Game Version: 0.G-2494-g22e9c95c2c [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

I think someone recently modified things to use maps rather than lists, and this may be a case that wasn't covered or tested properly.

PatrikLundell:
#### Summary

None

#### Purpose of change

Fix the Workshop 2 expansion salt water pipe that was broken when someone replaced list parameters with sets.
Fix #66318.

#### Describe the solution

Since an empty list is accepted syntactically, but the type of the parameter used has some kind of default value a value with the default value was created and passed instead of the empty list, which caused the hash stuff to stop blowing up trying to access the first element of an empty map.

Continue to wait to no avail for someone who actually know what they're doing to fix the problem.

#### Testing

1. Loaded a save with a partially built Workshop 2 expansion.
2. Ordered the companion to build a salt water pipe.
3. Repeated the construction until the whole pipe was constructed.
4. Verified that examining the pump offered it to provide me with salt water.
5. Followed the "pipe" to the well in the swamp to verify the well and the "pipe" was still there.

#### Additional context

Tried to set up a hide site and to build a palisade as both of these missions are given empty lists as their parameters as well. Both were sent off successfully, so they're probably not broken. It's possible these missions aren't considered blueprints and thus aren't sent to the exploding hash finding function, but I haven't checked as they work as is.

As should be clear by the above, I don't really know what I'm doing here, although the guess about using a default value is a reasonable one. Thus, it would be useful if someone who actually has any idea of how the hash macro mess works takes a look at it.

### DIFF
--- a/src/faction_camp.cpp
+++ b/src/faction_camp.cpp
@@ -2869,7 +2869,8 @@ bool basecamp::common_salt_water_pipe_construction(
         }
     }
 
-    basecamp_action_components components( making, {}, 1, *this );
+    mapgen_arguments arg;  //  Created with a default value.
+    basecamp_action_components components( making, arg, 1, *this );
     if( !components.choose_components() ) {
         return false;
     }


akrieger (approved):
Your fix is right but the problem statement is wrong. The issue is the args are taken by reference and *stored by reference*. Using a `{}` in the constructor initializes a temporary that is then immediately destroyed leaving a dangling reference.

ASAN would have caught this.

akrieger (approved):

PatrikLundell:
@akrieger I see, but that also means that the other calls to the same operation using  ` {} ` in the file are wrong as well and should be fixed in the same way? If you say yes I'll do that.

akrieger:
Yes they're all probably problematic.

PR #68620, opened 2023-10-12
NPC tasks are not affected by avatar's Z-level

### LINKED ISSUE #68212: NPCs performing tasks will stop them to follow you across z-levels
### Describe the bug

If you tell an NPC to craft something, they will stay there and craft if you leave, but if you go down/up stairs somewhat nearby they will stop and follow you. This makes it significantly more frustrating to operate in bases that span multiple z-levels or require moving across them to access.

### Attach save file

- OS: Windows
- Game Version: 398e915 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Mythical Martial Arts [MMA],
    SpeedyDex [speedydex],
    Portal Storms Ignore NPCs [personal_portal_storms]
]

### Steps to reproduce

1. Order an NPC to perform a task, crafting for example.
2. Cross nearby stairs, even outside of view.
3. NPC will stop what they're doing and follow you.

### Expected behavior

NPCs should not follow you across z-levels when they are busy.

### Screenshots

_No response_

- OS: Windows
- Game Version: 398e915 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
    Mythical Martial Arts [MMA],
    SpeedyDex [speedydex],
    Portal Storms Ignore NPCs [personal_portal_storms]
]

### Additional context

_No response_

marimarigi:
#### Summary
Bugfixes "NPC tasks end when the avatar goes to a different Z-level"

#### Purpose of change
Fix #68212

#### Describe the solution
NPC recalculates light intensity based on its own Z-level.
~~**EDIT**: If the Z-level of the last generated lightmap matches, skip it to reduce recalculation.~~
**EDIT2**: If a proper sunlight_cache or lightmap already exists, NPCs will skip generating it.

**EDIT**: Fully cache ambient_light.

#### Testing
I confirmed by going up and down the stairs while having multiple NPCs craft keffiyehs at different Z-levels.

#### Additional context
https://github.com/CleverRaven/Cataclysm-DDA/issues/68212#issuecomment-1737785903 was very helpful in resolving this issue.


marimarigi:
@andrei8l I think the "npc_light_and_fine_detail_vision_mod" test added in #67295 is done outdoors, with both receiving only equal natural light.

andrei8l:
> @andrei8l I think the "npc_light_and_fine_detail_vision_mod" test added in #67295 is done outdoors, with both receiving only equal natural light.

Sunlight is raycasted down until it meets a floor so that sounds quite probable. But from looking briefly at the code, it also means that this solution can overwrite the lightmap on the avatar's level with just sunlight.

marimarigi:
Is it too rough to record the Z-levels for which lightmaps have already been generated at least once this turn, and not generate new lightmaps at those Z-levels except for avatars?

src/map.h
```diff
@@ -1946,7 +1946,9 @@ class map
                                bool cumulative = false,
                                bool camera = false, int penalty = 0 );
         void apply_character_light( Character &p );
-
+    public:
+        void get_lightmap( const int zlev );
```
Maleclypse, line 1950: Any way to make clang happy?

src/lightmap.cpp
```diff
@@ -651,6 +652,14 @@ float map::light_transparency( const tripoint &p ) const
     return get_cache_ref( p.z ).transparency_cache[p.x][p.y];
 }
 
+void map::get_lightmap( const int zlev )
+{
+    if( level_cache::lightmap_latest_zlevel == zlev ) {
```
andrei8l, line 657: This will only prevent spurious recaching if the NPCs are processed by z-level and only once per turn.

src/lightmap.cpp
```diff
@@ -651,6 +652,14 @@ float map::light_transparency( const tripoint &p ) const
...
+        return;
+    }
+    return generate_lightmap( zlev );
```
andrei8l, line 660: This still calls `map::build_sunlight_cache()` which overwrites at least one of the caches above zlev with just sunlight.

src/character.cpp
```diff
@@ -2757,11 +2757,13 @@ float Character::fine_detail_vision_mod( const tripoint &p ) const
     float ambient_light{};
     tripoint const check_p = p == tripoint_min ? pos() : p;
     tripoint const avatar_p = get_avatar().pos();
-    // Light might not have been calculated on the NPC's z-level
+    map &here = get_map();
     if( is_avatar() || check_p.z == avatar_p.z ||
         ( fov_3d && std::abs( avatar_p.z - check_p.z ) <= fov_3d_z_range ) ) {
-        ambient_light = std::max( 1.0f,
-                                  LIGHT_AMBIENT_LIT - get_map().ambient_light_at( check_p ) + 1.0f );
+        if( !is_avatar() && here.ambient_light_at( check_p ) <= 4 ) {
```
andrei8l, line 2763: What's this 4? Did you mean `LIGHT_AMBIENT_DIM` (=5)? It shouldn't be needed if you can prevent recaching though.

Maleclypse:
Everburning sconce randomly stops lighting

Compiled and tried out some things.  Had some weird experiences where when I climbed onto a different level and moved around the permanent light source I had created suddenly stopped showing light and then wouldn't reappear when I reached a space that had previously seen the light. 

I think I had some issues with NPCs stopping tasks randomly as if they'd lost light when waiting in the dark on other levels, but I was unable to capture that when I switched to the everburning torch so I'm not certain that I wasn't experiencing some kind of charge issue.

marimarigi:
Isn't that a pre-existing problem unrelated to this PR?

EvanBalster:
I've noticed that sleeping, or possibly sleeping with a blindfold, affects NPC tasks as well.  Please account for this in your tests!

marimarigi:
NPCs will continue to work even if avatar sleep with/without a blindfold.

andrei8l (commented):
Sorry it took me so long to get back to this.

First thing is that this really needs a test unit. Inserting NPCs above/below the avatar in existing vision tests would also be nice, to ensure they don't screw up anything on the avatar's level.

Second thing is that I'm not sure this ad-hoc solution is the best. This won't fix activities that rely only on `Character::sees()` and also doesn't affect monster vision at all. Also, due to the order of operations, this doesn't let the avatar see light arcs from a source below them even though we're paying the computation cost for it.

Perhaps a better solution is to generate the lightmaps for each zlevel if:
- the avatar can see it and it is below them, or
- an NPC is performing an activity on that level, or
- monsters from different factions are in the same visitable zone

Night-Pryanik:

PR #68623, opened 2023-10-12
Find items that cover body part

inogenous:
#### Summary
Interface "Find items that cover body part"

#### Purpose of change

* Adds a way to filter items by whether or not they cover a given body part when worn.

#### Describe the solution

* Example: `v:hand` or `v:eyes` would find all gloves or glasses respectively.
* Example use case: I'd like to find all gloves in my stash, and possibly compare them to what I have now. Previously, we could filter items for something like `glove`, but that would miss stuff like `fire gauntlet` or `brass knuckles`.
* The filter applies to all items searches with `/` such as:
    * inventory
    * AIM
    * viewing nearby items with `V`
    * creating zone with `Loot: Custom`
    * comparing items with `I`.
* This commit also updates the help text for item filters so that it looks more like the help text shown when filtering recipes in `&`.

* Input is welcome on whether we should use another character than `v:` for this type of filter.

#### Testing

Before:
filter-by-coverered-bodypart-before

After:
filter-by-coverered-bodypart-after

filter-by-coverered-bodypart-new

### DIFF
--- a/src/item_search.cpp
+++ b/src/item_search.cpp
@@ -75,6 +75,34 @@ std::function<bool( const item & )> basic_item_filter( std::string filter )
                 }
                 return false;
             };
+        // covers bodypart
+        case 'v': {
+            std::unordered_set<bodypart_id> filtered_bodyparts;
+            std::unordered_set<sub_bodypart_id> filtered_sub_bodyparts;
+            for( const body_part &bp : all_body_parts ) {
+                const bodypart_str_id &bp_str_id = convert_bp( bp );
+                if( lcmatch( body_part_name( bp_str_id, 1 ), filter )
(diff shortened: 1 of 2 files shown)


Termineitor244:
It's possible to search for a sub-body part? Like upper torso, knees, nape, etc.?

inogenous:
> Is it possible to search for a sub-body part?

A relevant question. Thank you. Yes, this is covered by the suggested code changes.

Termineitor244:
> > Is it possible to search for a sub-body part?
> 
> A relevant question. Thank you. Yes, this is covered by the suggested code changes.

Thanks to you! This is a change I have wanted for a long time.

AmarinReyny:
Does this also work for the crafting menu?

Brambor:
> Does this also work for the crafting menu?

 - It does since #78346.

PR #68652, opened 2023-10-14
Minor density list cleanups

Karol1223:
#### Summary
None

#### Describe the solution
- Added a new placeholder scorched human and made the predefined scorched corpse use it as a base
- Removed the no longer existing filthy toothbrush ID from the list
- Audited the numbers for the regular toothbrush based off an Amazon listing
- Changed the volume of the base toiletries abstract so that things that copy from it don't automatically have broken density

#### Additional context
This PR showed that the default human dimensions (81.5kg/62.5l) are actually wrong, because they break density tests. My calculations show that realistic volume would be in the 140l range, but this is out of the scope of this PR.


TheSaddestGoomba:
I was pulling my hair out trying to make sense of the weights and volumes for things when looking at references for frog mutants.
On a related note, the colossal croaker is much larger than a moose, despite its reference to one in its description. IIRC the moose is currently even smaller than a cow by volume.

Karol1223:
> I was pulling my hair out trying to make sense of the weights and volumes for things when looking at references for frog mutants. On a related note, the colossal croaker is much larger than a moose, despite its reference to one in its description. IIRC the moose is currently even smaller than a cow by volume.

Monster sizes are a mess and a half. Zombie crawler is the same size as a dissoluted devourer. I want to try cleaning it up a little, at least for standard zombies, but first I gotta clean up zombie JSON the way I did it for ferals so that ``copy-from`` actually carries over the changes.

PR #68658, opened 2023-10-14
Aftershock: Attaching atompot to mounted kitchen

DeCEll-1:
#### Summary
None

#### Purpose of change
Other household items can be attached to mounted kitchen but atompot cant

#### Describe the solution
Add the atompot to the mounted kitchens allowed_tools list by copying mounted kitchen and override it like that so it doesn't interfere with other mods

I tried to just extend it with mounted kitchens id, type and extend. It gave errors with that
After that I tried to extend the abstract, it didn't gave any errors but didn't worked either

#### Testing

Spawn undamaged camper van from debug > attach atompot into the kitchen

Start vehicle construction > install mounted kitchen > attach atompot into the kitchen

Also after doing the same for another mod:
-Tried to make a save without enabling one of it. No problem visible
-Tried to make a save with both mods enabled. Both mods items were attachable without any visible problems

### DIFF
--- a/data/mods/Aftershock/vehicles/vehicle_parts.json
+++ b/data/mods/Aftershock/vehicles/vehicle_parts.json
@@ -195,5 +195,16 @@
       { "item": "welder", "charges": 0, "prob": 50 }
     ],
     "variants": [ { "symbols": "T", "symbols_broken": "x" } ]
+  },
+  {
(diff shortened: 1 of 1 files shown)


Maleclypse:
Thanks for your contribution!

PR #68664, opened 2023-10-15
Cattail can be planted in shallow water

Light-Wave:
#### Summary
Content "Cattail can be planted in shallow water"

#### Purpose of change
#68616 made cattail not plantable on land. This PR makes cattail farmable again. It also makes it so that different plants can have different terrain requirements for farming.

#### Describe the solution
`"seed_data"` has a new optional field, `"required_terrain_flag"`. This is set to `"PLANTABLE"` by default. In order to plant a seed somewhere, the terrain or furniture there needs to have the given flag.
If this flag is set to anything other than `"PLANTABLE"`, the terrain will not turn into dirt after the seed is planted. This is to prevent the water to turn into dirt when you plant stuff in it.
Places that previously checked for the `"PLANTABLE"` tag now instead checks for the tag specified by the `"seed_data"`.
Cattail seeds are set to have `"required_terrain_flag": "SHALLOW_WATER"`.
Documentation is updated to match.
A new hardcoded message string is added for when you're trying to plant a seed in an invalid spot; `add_msg( _( "This type of seed can not be planted in this location." ) );`. This might need translation.

Give shallow water `"examine_action": "dirtmound"` so that seeds can be planted. I felt this would clutter up the UI, but right now the seeds can not be planted manually, and have to instead be planted using the zone manager, which might be confusing.
Create aquatic farmland as a furniture item instead. This could be given the `"examine_action": "dirtmound"` to let players plant the crops manually. I felt like this would add extra pointless work for the player, but I'm not sure if it might be a better idea. Using a furniture instead of a terrain would also make it so I don't need a special case in order to not turn the planted terrain into dirt if we use special terrain.
Hide seeds that can't be planted from the seed selection menu. Felt like extra work, and would confuse the player. My hope is right now that they will try to plant cattail seeds in soil, get the error message that "This type of seed can not be planted in this location.", read the item description and notice that it says it must be planted in shallow water, and figure things out from there. Thus, hiding invalid seeds might be more confusing than helpful.
If this were to be used for non-seed like beehives, perhaps it would be a better idea to hide invalid seeds and change the message to "You can not plant <item_name_plural> in this location.".
Right now, if you define an area of dirt to be used to farm cattail using the zone manager, the player will still try to tilt the earth, but be unable to plant the seeds. Perhaps tilting of earth should not be done for farmlands where the seeds does not have `"required_terrain_flag": "PLANTABLE"`
Instead of always trying to tile an area, let `seed_data` define a `construction_group` it will try to use to make the area plantable. Seems like a much better solution, but also sounds like a bunch of extra work. Maybe later.

#### Testing
From my testing, it seems to work.

Here I designated an area to be a cattail farm:
Capture

Here I am after having farmed the area:
Capture2

Harvesting also seems to work fine:
Capture3

Asking an NPC to work the field also seems to work fine. It also still works to plant normal crops both by yourself and with NPCs:
Capture5

#### Additional context
@harakka requested a ping.
Also pinging @AmarinReyny since they asked about this.
If anyone knows if any extra steps are needed to translate the hardcoded message I added, feel free to let me know. Translation is magic to me.

### DIFF
--- a/src/iexamine.cpp
+++ b/src/iexamine.cpp
@@ -2558,6 +2558,11 @@ void iexamine::dirtmound( Character &you, const tripoint &examp )
     }
     const auto &seed_id = std::get<0>( seed_entries[seed_index] );
 
+    if( !here.has_flag_ter_or_furn( seed_id->seed->required_terrain_flag, examp ) ) {
+        add_msg( _( "This type of seed can not be planted in this location." ) );
+        return;
+    }
+
     plant_seed( you, examp, seed_id );
 }
 
--- a/data/json/items/comestibles/seed.json
+++ b/data/json/items/comestibles/seed.json
@@ -264,9 +264,15 @@
     "copy-from": "seed",
     "name": { "str_sp": "cattail seeds" },
     "color": "green",
-    "description": "Some cattail seeds.",
-    "flags": [ "NUTRIENT_OVERRIDE", "INEDIBLE" ],
-    "seed_data": { "plant_name": "cattail", "fruit": "cattail_stalk", "byproducts": [ "cattail_rhizome" ], "grow": "91 days" }
+    "description": "Some cattail seeds. Must be planted in shallow water.",
+    "flags": [ "PLANTABLE_SEED", "NUTRIENT_OVERRIDE", "INEDIBLE" ],
+    "seed_data": {
+      "plant_name": "cattail",
+      "fruit": "cattail_stalk",
+      "byproducts": [ "cattail_rhizome" ],
+      "grow": "91 days",
+      "required_terrain_flag": "SHALLOW_WATER"
+    }
   },
   {
     "type": "COMESTIBLE",
--- a/data/json/items/comestibles/seed.json
+++ b/data/json/items/comestibles/seed.json
@@ -264,7 +264,7 @@
     "copy-from": "seed",
     "name": { "str_sp": "cattail seeds" },
     "color": "green",
-    "description": "Some cattail seeds. Must be planted in shallow water.",
+    "description": "Some cattail seeds.  Must be planted in shallow water.",
     "flags": [ "PLANTABLE_SEED", "NUTRIENT_OVERRIDE", "INEDIBLE" ],
     "seed_data": {
       "plant_name": "cattail",
--- a/doc/JSON_INFO.md
+++ b/doc/JSON_INFO.md
@@ -4034,9 +4034,12 @@ Every item type can have optional seed data, if the item has seed data, it's con
     "fruit_div": 2, // (optional, default is 1). Final amount of fruit charges produced is divided by this number. Works only if fruit item is counted by charges.
     "byproducts": ["withered", "straw_pile"], // A list of further items that should spawn upon harvest.
     "plant_name": "sunflower", // The name of the plant that grows from this seed. This is only used as information displayed to the user.
-    "grow" : 91 // A time duration: how long it takes for a plant to fully mature. Based around a 91 day season length (roughly a real world season) to give better accuracy for longer season lengths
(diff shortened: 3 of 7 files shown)


AmarinReyny:
Thank you!!!

src/iexamine.cpp
```diff
@@ -2558,6 +2558,11 @@ void iexamine::dirtmound( Character &you, const tripoint &examp )
     }
     const auto &seed_id = std::get<0>( seed_entries[seed_index] );
 
+    if( !here.has_flag_ter_or_furn( seed_id->seed->required_terrain_flag, examp ) ) {
+        add_msg( _( "This type of seed can not be planted in this location." ) );
```
ehughsbaird, line 2562: _( str ) marks str for extraction, and will return the translated string for the set language, so this is all you need to do to ensure translation is done.

harakka:
Looks good, thanks for doing this.
In longer term more complex checks (like SHALLOW_WATER but no CURRENT) would be nice, but like we discussed I think EOCs would be a good way to evaluate those conditions and we don't have enough muscle for that yet. I'll keep an eye on that.

src/iexamine.cpp
Light-Wave: Thanks!

Light-Wave:
> Looks good, thanks for doing this. In longer term more complex checks (like SHALLOW_WATER but no CURRENT) would be nice, but like we discussed I think EOCs would be a good way to evaluate those conditions (so we don't need to implement separate sets of allowed and forbidden flags etc) and we don't have enough muscle for that yet. I'll keep an eye on that.

It should be possible to add a new tag called CATTAIL_PLANTABLE and use that instead. That being said, cattail currently grows naturally in water with currents in-game.

AmarinReyny:
Would it not be easier to use terrain IDs instead of flags? If needed, maybe there could be terrain groups representing multiple types of terrain.

Then again, I have no idea how to code C++ or anything like that, so I am aware that I might be completely wrong about this.

Light-Wave:
> Would it not be easier to use terrain IDs instead of flags? If needed, maybe there could be terrain groups representing multiple types of terrain.

I considered that, but since you can add custom tags if you want to, and add them to the terrain or furniture you want to be able to grow a specific plant, the solution as it is is essentially equivalent. You already have full control over which terrains should and should not be plantable with which seeds.

AmarinReyny:
> > Would it not be easier to use terrain IDs instead of flags? If needed, maybe there could be terrain groups representing multiple types of terrain.
> 
> I considered that, but since you can add custom tags if you want to, and add them to the terrain or furniture you want to be able to grow a specific plant, the solution as it is is essentially equivalent. You already have full control over which terrains should and should not be plantable with which seeds.

... Wait, we can do that? Awesome! Thank you for letting me know!

ThePainkiller:
Does this change require a new world to make effect? Because interacting with any terrain with the `SHALLOW_WATER` flag only returns the usual consume/pour in container/pour in ground menu, even if my cattail seeds say they can be planted on shallow water.
Tried with `t_water_sh`, `t_water_murky`, and `t_water_moving_sh` while having cattails in inventory.

ehughsbaird:
> Give shallow water `"examine_action": "dirtmound"` so that seeds can be planted. I felt this would clutter up the UI, but right now the seeds can not be planted manually, and have to instead be planted using the zone manager, which might be confusing.

BjoHart:
> > Give shallow water `"examine_action": "dirtmound"` so that seeds can be planted. I felt this would clutter up the UI, but right now the seeds can not be planted manually, and have to instead be planted using the zone manager, which might be confusing.

Is there a way to check your inventory for the seeds, and only if you have seeds in your inventory, the option to plant the seeds in the shallow water will apear?
To not clutter up the UI to much...

kevingranade:

https://discourse.cataclysmdda.org/t/how-to-plant-cattail-seeds/29081/3

PR #68666, opened 2023-10-15
Infarstructure: Move methods from npc to Character

### LINKED ISSUE #68657: Infrastructure: new class Crafter; Character >: Crafter >: (NPC, Avatar)

I cannot refer to an NPC and the avatar as the same in crafting. More specifically: I cannot make a list of NPCs and the avatar and call functions such as `is_ally` or 
```cpp
recipe_subset Character::get_available_recipes( const inventory &crafting_inv,
        const std::vector<npc *> *helpers )
```

### Usecase
Implementing #68183

First, I make a list, which I do like this:
```cpp
std::vector<Character *> Character::get_crafting_group() const
{
    // GET NPCs
    std::vector<npc *> npcs = g->get_npcs_if( [this]( const npc & guy ) {
        // NPCs can help craft if awake, taking orders, within pickup range and have clear path
        return guy.is_ally( *this )
               && rl_dist( guy.pos(), pos() ) < PICKUP_RANGE
               && get_map().clear_path( pos(), guy.pos(), PICKUP_RANGE, 1, 100 );
    } );

    // RESERVE
    std::vector<Character *> ret;
    ret.reserve( npcs.size() + 1 );  // +1 for player
    for( npc *np : npcs ) {
        ret.emplace_back( static_cast<Character *>( np ) );
    }

    // CHECK THAT CHARACTER IS NPC OR AVATAR, SO NOT A RANDOM MONSTER
    // This can be called by avatar or NPC
    bool friendly_player = false;
    if( is_npc() ) {
        friendly_player = as_npc()->is_player_ally();
    } else if( is_avatar() ) {
        friendly_player = true;
    } else {
        debugmsg( "get_crafting_group called by somebody else than avatar or npc." );
    }

    // GET AVATAR
    avatar &avat = get_avatar();
    if( friendly_player
        && rl_dist( avat.pos(), pos() ) < PICKUP_RANGE
        && get_map().clear_path( pos(), avat.pos(), PICKUP_RANGE, 1, 100 )
      ) {
        ret.emplace_back( static_cast<Character *>( &get_avatar() ) );
    }
    return ret;
}
```
But I would like to do this:
```cpp
std::vector<Crafter *> Crafter::get_crafting_group() const
{
    return g->get_crafters_if( [this]( const npc & guy ) {
        // NPCs can help craft if awake, taking orders, within pickup range and have clear path
        return guy.is_ally( *this )
               && rl_dist( guy.pos(), pos() ) < PICKUP_RANGE
               && get_map().clear_path( pos(), guy.pos(), PICKUP_RANGE, 1, 100 );
    } );
}
```

And I wouldn't have to duplicate original code like this:
```cpp
        bool has_recipe( const recipe *r, const inventory &crafting_inv,
                         const std::vector<npc *> &helpers ) const;
        recipe_subset get_available_recipes( const inventory &crafting_inv,
                                             const std::vector<npc *> *helpers = nullptr ) const;
```
Adding these (and their implementation):
```cpp
        bool has_recipe( const recipe *r, const inventory &crafting_inv,
                         const std::vector<Character *> &helpers ) const;
        recipe_subset get_available_recipes( const inventory &crafting_inv,
                                             const std::vector<Character *> *helpers = nullptr ) const;
```

### Solution you would like.

New class `Crafter` that is a subtype of `Crafter` and supertype of `NPC` and `Avatar`.

This class would implement commonalities between NPC and Avatar so that crafting code doesn't care in most cases whether we ask about NPC or Avatar. (Read as: for all cases of certain methods, **not** some cases for some methods).

I would like this also to include friendly checks between persons. So with a list of Crafters, you can pick anyone and get their allies. Now such a list would have to be `Character` (supertype of NPC and Avatar), who doesn't implement ally code, so we have to call `if (is_npc) {...}, else if (is_avatar) {...} else {dbg("called by somebody else than Crafter or Avatar"); return false}`.

 Don't call it `Crafter`, call it `Person` to better reflect that it can be an ally of others. But right now, I would use it crafting code, so Crafter might be better. 

### Additional context

Let's say I implement this. Is it likely to be merged? NPC, Avatar, and Character would all look and behave the same. There would be just another class in between people can use when they don't care if it is Avatar or NPC but need Person's methods.

Brambor:
#### Summary
None

#### Purpose of change
Closes #68657

#### Describe the solution
Whatever function I was changing, I read the whole thing to make it mention Characters instead of NPCs where appropriate.

Move methods that are necessary in #68183 from `npc` to `Character` (supertype of `npc`). That PR will require this PR to get merged.
In `Character`, they are abstract methods.
In `npc` they are as they used to be.
In `Avatar`, the code delegates the implementation to the already existing code in `npc`. Utilizing assumptions that two avatars cannot exist (in which case it would recursively call the other until it crashes from stack overflow).

#### Testing
Compiling...

#### Additional context
More methods could be moved

### DIFF
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -1194,6 +1194,15 @@ bool avatar::is_ally( const Character &p ) const
     return guy.is_ally( *this );
 }
 
+bool avatar::is_obeying( const Character &p ) const
+{
+    if( p.getID() == getID() ) {
+        return true;
+    }
+    const npc &guy = dynamic_cast<const npc &>( p );
+    return guy.is_obeying( *this );
+}
+
 bool avatar::cant_see( const tripoint &p )
 {
 
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -1185,6 +1185,15 @@ faction *avatar::get_faction() const
     return g->faction_manager_ptr->get( faction_your_followers );
 }
 
+bool avatar::is_ally( const Character &p ) const
+{
+    if( p.getID() == getID() ) {
+        return true;
+    }
+    const npc &guy = dynamic_cast<const npc &>( p );
+    return guy.is_ally( *this );
+}
+
 bool avatar::cant_see( const tripoint &p )
 {
 
--- a/src/avatar.cpp
+++ b/src/avatar.cpp
@@ -461,11 +461,11 @@ bool avatar::read( item_location &book, item_location ereader )
     const std::string skill_name = skill ? skill.obj().name() : "";
 
     // Find NPCs to join the study session:
-    std::map<npc *, std::string> learners;
+    std::map<Character *, std::string> learners;
     //reading only for fun
-    std::map<npc *, std::string> fun_learners;
-    std::map<npc *, std::string> nonlearners;
-    for( npc *elem : get_crafting_helpers() ) {
+    std::map<Character *, std::string> fun_learners;
+    std::map<Character *, std::string> nonlearners;
+    for( Character *elem : get_crafting_helpers() ) {
         const book_mastery mastery = elem->get_book_mastery( *book );
         const bool morale_req = elem->fun_to_read( *book ) || elem->has_morale_to_read();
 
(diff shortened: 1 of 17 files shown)


src/avatar.cpp
```diff
@@ -1185,6 +1186,24 @@ faction *avatar::get_faction() const
     return g->faction_manager_ptr->get( faction_your_followers );
 }
 
+bool avatar::is_ally( const Character &p ) const
+{
+    if( p.getID() == getID() ) {
```
ZhilkinSerg, line 1191: I think you can unconditionally return `true` here and in the function below. It is not possible to have multiple avatars in a single player game.

Brambor: It can be called from any Character, so any NPC. It might be true now, but if NPCs get the ability to craft (regardless of player interaction), they would piggyback off of the player whenever they are close.

ZhilkinSerg: avatar is not npc and would never be.

Brambor: The parameter is Character &, so this is valid pseudo code:

```cpp
npc foo = npc()
avatar you = avatar()
you.is_ally(foo)
```

Brambor: Actual code:

```cpp
std::vector<Character *> Character::get_crafting_group() const
{
    return g->get_characters_if( [this]( const Character & guy ) {
        return guy.is_ally( *this )
               && rl_dist( guy.pos(), pos() ) < PICKUP_RANGE
               && get_map().clear_path( pos(), guy.pos(), PICKUP_RANGE, 1, 100 );
    } );
}

std::vector<Character *> game::get_characters_if( const std::function<bool( const Character & )>
        &pred )
{
    std::vector<Character *> result;
    avatar &a = get_avatar();
    if( pred( a ) ) {
        result.push_back( &a );
    }
    for( npc &guy : all_npcs() ) {
        if( pred( guy ) ) {
            result.push_back( &guy );
        }
    }
    return result;
}



```

Brambor:
Note! I am assuming here that obeying and is_ally is transitive, so if A is an ally of B then B is an ally of A.

PR #68667, opened 2023-10-15
Add and update item ASCII art for ammo and batteries

Lumi-Virtual:
#### Summary
Content "Added item ASCII art for ammo and batteries"

#### Purpose of change
The first of hopefully many pull requests to add more ASCII art to the game, something I have been wanting to see more of for a long while.

#### Describe the solution
Add more art! I based my art off of a lot of real life references, in game descriptions, and in game volumes when needed. I also asked around in the discord if there was a style guide for ASCII art and was told it was essentially up to the artist. Since there wasn't a set of guidelines, I decided on using a fixed scale, where one character is 1.5cm wide and 3cm tall. I think this is worth a little detail loss for very small items as it gives a very strong sense of scale and cohesion, and I also have a potential solution for the downsides of a fixed scale in the alternatives section.
Also I don't intend to imply that I'm asking for a fixed scale to be turned into an official guideline, but it is what I will keep using for all the art I make so I figured it was worth mentioning.

For very small items and very large items, a fixed scale can cause the art to either have very little detail, or not fit at all due to maximum line width. A solution for this I plan to use, once a decent number of items are added, is to have a scale multiplier icon in the drawing somewhere to show a zoomed in or out version of the item, potentially in addition to a real scale version for very small items. This could maintain the sense of scale while allowing high detail when wanted, and could allow very large items to fit when they otherwise wouldn't.

#### Testing
I launched the game from a fresh master branch pull with only my changes added, looked at every item that I added art to or changed the art of, and resolved any errors or messed up art I ran into. I wasn't able to easily lint the JSON files because it would reduce small drawings into a single line which I think hurts readability, so I did my best to make sure and double check that everything conforms to the style guidelines.

#### Additional context
I've posted a few screenshots of the art that's in in this pull request on the discord in the graphics-and-sound-dev channel if you want a few visual references.


data/json/ascii_art/ammo/20x66mm.json
```diff
@@ -0,0 +1,22 @@
+[
+  {
+    "type": "ascii_art",
+    "id": "20x66_abstract",
```
GuardianDll, line 4: abstracts are not actual items, and don't need to have an ascii art

Lumi-Virtual: the abstract doesnt but items that copy from it will use the art it has, which is useful for things like ammo where most of the variants look the same ^-^

Lumi-Virtual:
huh it failed the basic build and test but passed the windows build, and the error that caused the failure implies to me that theres a mismatch between the `ascii_picture` value in the handheld battery items and the `id` of the art for them, but that doesnt seem to be the case as far as i can tell

Lumi-Virtual:
ok reading closer, it looks like the all mods test failed because a modded item expects that the previous ascii art for batteries exists, which isnt true anymore because i renamed the ids for the battery art to the items ids

GuardianDll:
Can you update your branch? i'll try to make another lint PR later

Lumi-Virtual:
update it with the new linting for ascii pictures i assume ?

GuardianDll:
Correct

Lumi-Virtual:
hm doesnt like the new lint format, im guessing i have to do something specific to get the new linting files so it passes ? not sure how to do that

Lumi-Virtual:
iv changed the battery picture ids in innawood too now, if we are updating the ids, should i just go ahead and change all of them ?

GuardianDll:
i think yes, unless you want to make ASCII for innawood batteries

Lumi-Virtual:
i meant as in, should i replace all the ascii picture ids that the game uses with ones that are cleaner ?

Maleclypse:
So is this ready to merge or do I need to wait?

Lumi-Virtual:
its ready to merge if the way iv dealt with the ascii picture ids is desired. currently, only the picture ids of existing pictures that iv draw new art for has been replaced with tidier ids.
- we could keep things as is, in which case this can be merged and i will probably replace the ids incrementally in separate prs as i make new art that updates existing ascii pictures.
- i could undo the id changes and keep all the existing picture ids.
- or i could replace the existing ids with tidier ones all at once in this pr.

GuardianDll:
oh, i see what you mean, but i don't think it's worth it's effort

PR #68673, opened 2023-10-15
Adding some basic blacksmithing tools to the general tools pool

MorvarchPrincess:
#### Summary
Balance "Add basic metalworking equipment to the general tools pools, meaning they'll show up in more places"

#### Purpose of change

Amateur blacksmithing is a thing, and many professions have need for basic metalworking tools. It makes sense to add some of the simpler blacksmithing equipment to the general tools loot pool (with a low chance).

#### Describe the solution

I made a couple of sets of blacksmithing tools that have been added to tools-general, one is less complete and is more common and vice versa.

Make it spawn individual tools. Commentators mentioned it seemed more realistic to have these in a group rather than individual, unlike a lot of tools an anvil on its own is kind of useless.

#### Testing

Port around and confirmed I could find blacksmith tools occasionally.

### DIFF
--- a/data/json/itemgroups/tools.json
+++ b/data/json/itemgroups/tools.json
@@ -20,35 +20,36 @@
     "id": "tools_simple_blacksmith",
     "type": "item_group",
     "//": "Tools commonly used by amateur blacksmiths. A subset of blacksmiths tools",
-    "items": [
-      [ "basic_metalworking_a", 40 ],
-      [ "basic_metalworking_b", 60 ]
-    ]
+    "items": [ { "group": "basic_metalworking_a", "prob": 40 }, { "group": "basic_metalworking_b", "prob": 60 } ]
   },
   {
     "id": "basic_metalworking_a",
     "type": "item_group",
     "subtype": "collection",
-    "items": [
-      [ "anvil" ],
-      [ "chisel" ],
-      [ "metal_file" ],
-      [ "hotcut" ],
-      [ "metalworking_tongs" ],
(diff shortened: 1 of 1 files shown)


Karol1223:
I feel like the itemgroup could use different weights between items. An anvil having the same chance to spawn as a sandpaper feels kinda weird, but I do admit I never looked for anvils in the store, so I'm just throwing this comment out there as something to think on, maybe?

It also kinda feels weird that the odds of finding blacksmithing equipment are the same as regular mechanic tool group (both with a weight of 20), but again - I never checked how often hardware stores sell blacksmithing gear, so if it *is* similarly common irl then that's fine.

This is all purely how things feel, though. I have no concrete knowledge over any of this, hence me just commenting instead of attempting to make a proper review, for the record.

Termineitor244:
I think the idea of the sets would be more useful since these kind of tools would be most commonly found as part of a set, maybe you can do it in the same way that cars spawn a tool box with, well, common tools for vehicles. You could do the same with a tool box rarely spawning in houses and the like with these tools.

PatrikLundell:
If you want to make sure the PR isn't merged prematurely, put it in draft mode and take it out again when done. I can't tell you exactly how it's done as it was a bit fiddly if I recall correctly, but it was possible to do with some work.

Maleclypse:
Convert to draft is a button in the reviewers tab on the top right.  Post in ready to merge when ready for this to merge as I'm now not sure if this is ready for merge.

PR #68684, opened 2023-10-15
Improved random spawns

Termineitor244:
#### Summary
Balance "Adds more spawns to several items"

#### Purpose of change
I have noticed some items that lack appropriate spawns or have possible improvements to be made for those spawns, so I went and did them!

#### Describe the solution
Adds more global spawns for the Javelin Bag, Nylon Quiver, Lanyard and Skateboard.
Improved the spawns for quivers giving them the possibility to spawn loaded with arrows in appropriate locations (The main one being the hunting store), now the arrows and bolts of the world have a proper place to be stored! It will still be rare to find the quivers with ammo, but now there is a chance when looting appropriate places!

None.

#### Testing
Started the game, it starts, loaded quivers can be found in hunting stores, skateboards in the school, etc.


PatrikLundell:
I've never visited a hunting store, but it seems odd that they'd sell quivers with arrows in them, as there are different kinds of arrows and we live in a world where "batteries not included" is the standard mode.
The same goes for closets with general "stuff" where the owner tossed their no longer used items. In contrast, I'd expect places reserved for bows and arrows for actual use to have quivers stocked (unless there is some reason for why arrows shouldn't be stored in quivers, such as potentially the feathers becoming crooked from being stored there).

Termineitor244:
> I've never visited a hunting store, but it seems odd that they'd sell quivers with arrows in them, as there are different kinds of arrows and we live in a world where "batteries not included" is the standard mode.
> The same goes for closets with general "stuff" where the owner tossed their no longer used items. In contrast, I'd expect places reserved for bows and arrows for actual use to have quivers stocked (unless there is some reason for why arrows shouldn't be stored in quivers, such as potentially the feathers becoming crooked from being stored there).

The loaded quivers of the hunting store are specifically on the mannequins and in the storage area behind the counter (Where the arrows are), so I imagine it as exhibition for the mannequins (A hunting outfit with quiver and some arrows inside so the customer can see how they would look like with the items), the storage behind the counter already has the arrows there and I imagine it as just using some quivers for storing the arrows, not specifically because they are sold as sets.

With closets, I wasn't exactly sure if they could be put in there, but we have a lack of hunters/recreative archers storing their arrows in their houses, so I thought it would be appropriate, or is there a better place inside the house for a hunter to store them?

PR #68697, opened 2023-10-16
updated the drop hammer

Daved27hundred:
#### Summary
Content "Gives the drop hammer an anvil quality."

#### Purpose of change

Despite having an anvil in it, the drop hammer appliance doesn't give the quality, you should be able to use the anvil part on it's own. It would be annoying to build this and then realize you need to make a second anvil.

#### Describe the solution

Gave the drop hammer basecamp tool an anvil quality of 3, and mentioned this in the description.

A different quality? I'm not sure.

### DIFF
--- a/data/json/furniture_and_terrain/furniture-appliances.json
+++ b/data/json/furniture_and_terrain/furniture-appliances.json
@@ -501,7 +501,7 @@
     "id": "f_drophammer",
     "name": "blacksmith drop hammer",
     "symbol": "#",
-    "description": "An anvil with a large metal hammer suspended above it in a metal framework.  If it were working, it would be useful for shaping softened metal plates, though now it is only useful for parts.",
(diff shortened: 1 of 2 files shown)


data/json/furniture_and_terrain/furniture-appliances.json
```diff
@@ -501,7 +501,7 @@
     "id": "f_drophammer",
     "name": "blacksmith drop hammer",
     "symbol": "#",
-    "description": "An anvil with a large metal hammer suspended above it in a metal framework.  If it were working, it would be useful for shaping softened metal plates, though now it is only useful for parts.",
+    "description": "An anvil with a large metal hammer suspended above it in a metal framework.  I would be useful for shaping softened metal plates, or you could use the anvil by itself.",
```
Karol1223, line 504:
```suggestion
    "description": "An anvil with a large metal hammer suspended above it in a metal framework.  It would be useful for shaping softened metal plates, or you could use the anvil by itself.",
```

Karol1223 (changes requested):

PR #68705, opened 2023-10-17
Revamp foldable solar panels

### LINKED ISSUE #68642: Unjustified Power Loss in Foldable Solar Panels

Currently "foldable solar panel" has only 66% epower output compared to the basic "solar panel" from which is made.
The "foldable" panel has the same area (and weight), is mounted the same way (on a vehicle frame) and the same solar cells (which **do** have the same efficiency), so any possible loss can occur only in additional wiring, but as the weight is the same, this is clearly not the case.

This issue have been discussed previously in https://github.com/CleverRaven/Cataclysm-DDA/pull/66478 (my previous unsuccessful attempt to fix the issue), so I invite everyone to participate in another round.

P.S. I consider comparison of in-game "foldable" solar panels with real-life pocket gadgets of the same *name* to be incorrect for reasons mentioned above.

### Solution you would like.

My proposal is to restore epower efficiency back to basic solar panel's level unless someone comes up with an estimation how much power can be lost in wiring needed to separate solar cells in a panel and connect them with hinges.

Alternatively, we can allow folding of  frame with basic solar panels and make "foldable solar panels" to be more in line with their real life implementations: small low power/weight (un-craftable?) pocket-size gadgets that must be simply dropped on the ground without any mounting and start acting as "battery chargers" for that tile.

### Additional context

_No response_

ehughsbaird:
#### Summary
Content "Revamp foldable solar panels"

#### Purpose of change
Fixes https://github.com/CleverRaven/Cataclysm-DDA/issues/68642
These didn't spawn anywhere, but could be crafted. When crafted, they functioned as worse solar panels, despite being made from solar panels.

Revamp them to be based on real items
The folding solar panel is now a [camping solar panel](https://www.bluettipower.com/products/bluetti-pv120-120w-solar-panel).
Small folding solar panel, unfolded on a white background

The advanced folding solar panel is now a [solar suitcase](https://www.renogy.com/200-watt-12-volt-monocrystalline-foldable-solar-suitcase).
Solar suitcase, deployed with a stand to have a favorable angle, on a white background

#### Describe the solution
In the first commit, add an iuse actor that enables activating an item to place it as an appliance.

In the second commit, allow vehicle parts to be removed into different items than they use for the items.  Having a pseudo item for the solar panel appliances.

In the other two commits, revamp the stats based on the items linked above, and add an appliance version. The appliance uses a new fake item that has a capacity of one battery charge, to enable charging items off of it.

Add to some relevant item groups, and remove the crafting recipes.

Both deploys to an area larger than a vehicle tile, so they probably shouldn't be allowed on vehicles, but I didn't want to break existing vehicles.

#### Testing
Spawn in folding solar panels and solar suitcases. Install them on a vehicle, then smash them. See a `XX` solar panel of the appropriate type comes off. Activate each type, and place them as appliances. Items hooked up to them charge.

Folding solar panel in bugout bag:
image
image

#### Additional context
Crafting recipes might not need to be removed, especially for the suitcase, I'm just not sure what to put for them. Hopefully the slightly hacky single-battery-charge thing is acceptable.


kevingranade:
> Both deploys to an area larger than a vehicle tile, so they probably shouldn't be allowed on vehicles, but I didn't want to break existing vehicles.

Sounds like it needs to be a new item then.

ehughsbaird:
I have removed the vehicle-mounted folding panels. I can't figure out what a real analogue would be, and the appliance versions of these provide all the functionality I could think you might want with a vehicle using folding solar panels.

Alm999:
If new folding solar panels are just appliances, then what are their benefits compared to regular solar panels placed as appliances? Is it ease of installation (reduced time)?

As far as I can understand, old "folding solar panels" are a means to have a foldable vehicle (an e-bike?) with a panel. Personally I intended to construct a portable solar farm using foldable solar panels (because using regular ones makes vehicle un-foldable), but the old ones were just not worth it compared to just placing ordinary solar panels one-by-one as appliances (beside huge game slowdown due to calculation of large amount of appliances nearby).

I see no benefits in seeking the proposed new un-craftable gadgets (and I hate relying on sheer luck with item spawn) when one can just place a panel from any random rooftop or electric car as appliance.

I would propose making portable foldable solar packs to be just like "battery chargers": small, lightweight, low-power, not requiring any mounting process whatsoever, but acting as chargers for all batteries in a tile if just being dropped on the floor/grass/dirt (provided sun radiation is present).

Oh, and maybe make ordinary solar panels compatible with foldable frames (provided there is no fundamental issues with this).

ehughsbaird:
> If new folding solar panels are just appliances, then what are their benefits compared to regular solar panels placed as appliances? Is it ease of installation (reduced time)?

Easier to deploy and remove (no tools, less time).
The camping solar panel is heavier, but less volume than an appliance (small) solar panel for twice as much power output.
The solar suitcase is less beneficial (200W for 17.5kg/40L instead of 240W for 10.5kg/50L for an appliance solar panel), but is still less voluminous, which tends to be the constraint for vehicles.
They also allow charging items directly from them, with no additional need for a battery, further saving volume/weight.

> As far as I can understand, old "folding solar panels" are a means to have a foldable vehicle (an e-bike?) with a panel. 

What I have changed these into are real things, I can't find a real thing that looks like this that would be foldable.

>Personally I intended to construct a portable solar farm using foldable solar panels (because using regular ones makes vehicle un-foldable), but the old ones were just not worth it compared to just placing ordinary solar panels one-by-one as appliances (beside huge game slowdown due to calculation of large amount of appliances nearby).

I think a portable solar farm with solar suitcases fits this.
 
> I see no benefits in seeking the proposed new un-craftable gadgets (and I hate relying on sheer luck with item spawn) when one can just place a panel from any random rooftop or electric car as appliance.

Yeah, they're niche, but they're based on things that exist. Perhaps there's some downsides that we aren't modeling (with regards to power conversion or the like) of a fixed solar panel.
I think it would probably be feasible to have craftable folding solar panels (with 2 solar_panel_v2?), but I think that can be done in a later PR, if this work gets merged.

> I would propose making portable foldable solar packs to be just like "battery chargers": small, lightweight, low-power, not requiring any mounting process whatsoever, but acting as chargers for all batteries in a tile if just being dropped on the floor/grass/dirt (provided sun radiation is present).

I think this would be feasible to craft with a battery charger and a foldable solar panel, but I also think fits best in a later PR. As is, anything with a cable can be connected to charge.

> Oh, and maybe make ordinary solar panels compatible with foldable frames (provided there is no fundamental issues with this).

I don't know enough about the criteria for making items foldable. Not to be a broken record, but I also think this belongs in a separate PR.

PR #68711, opened 2023-10-17
Allow culling more dogs

sodisappointing:
#### Summary
Balance "Allow culling of more tamed dog breeds, including mutants"

#### Purpose of change
When testing #68691 I noticed not all dogs had the ability to be culled. The inconsistency didn't seem to make sense to me.

#### Describe the solution
Adds the CAN_BE_CULLED flag to the following dogs (and the corresponding puppies):
Bulldog
Pit Bull Mix
Beagle
Border Collie
Boxer
Chihuahua
Dachshund
German Shepherd
Great Pyrenees
Rottweiler
Cattle Dog

And the following mutant dogs:
Mongrel
Loper
Mauler
Thrasher

Not allowing the culling of mutant dogs. Since they seem to have the same personality/disposition as normal dogs I figured it'd be reasonable to allow.
Allow culling of horses. Wasn't sure if allowing them to be culled makes sense considering they're pretty difficult animals to work with.

#### Testing
Debug spawned dogs + dog food, tamed them, and checked the interact menu for the cull option.


Karol1223:
> Allow culling of horses. Wasn't sure if allowing them to be culled makes sense considering they're pretty difficult animals to work with.

I think that's the exact reason horses weren't given the flag. I vaguely remember the author of the last culling PR saying this, and horse foals do have it, so it adds up.

RenechCDDA:
Yeah I specifically avoided *adult* horses when going over the tameable animals because it seems... difficult? to handle them for butchery without some sort of enclosure.

I am not a horse rancher and have no firsthand experience butchering them, so that is just my impression and should not be taken as word of law.

Re: the dogs

I was pretty sure all dogs copied from base dog. Evidently not! Good call.

sadenar:
Not the puppies you monster! D:

Maleclypse:
Needs json linting https://dev.narc.ro/cataclysm/format.html

PR #68738, opened 2023-10-18
Few docs fixes

GuardianDll:
#### Summary
None
#### Purpose of change
Just fix few issues with docs

### DIFF
--- a/doc/EFFECT_ON_CONDITION.md
+++ b/doc/EFFECT_ON_CONDITION.md
@@ -629,7 +629,7 @@ You have equipped an item that you can stow
 "u_can_stow_weapon"
 ```
 
-You have equipped an item that you can not stow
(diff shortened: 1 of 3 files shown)


doc/EFFECT_ON_CONDITION.md
```diff
@@ -624,12 +624,12 @@ checks do you have aikido stance active
...
 
 #### Examples
-
+You have equipped an item that you can stow
 ```json
 "u_can_stow_weapon"
 ```
 
-You have equipped an item with
+You have equipped an item that you can not stow
```
Maleclypse, line 632: Does this mean the item cannot be unwielded, like a cbm weapon or does it mean that you lack a sheath capable of holding the weapon?

GuardianDll: According to https://github.com/CleverRaven/Cataclysm-DDA/blob/master/src/condition.cpp#L1245, both

PR #68739, opened 2023-10-18
Adding a russian military helicopter crash site

Lupotech-CDDA:
#### Summary
Content "Adding a Russian helicopter crash site as a map extra event to the game."

#### Purpose of change
Adding a Russian helicopter crash site to the game. I thought it would be cool to have the military of other countries present in CDDA so that there are more sources of non NATO military equipment. Therefore I thought it would be possible that a Russian military helicopter, trying to evacuate an embassy or other important site, crashed on the way to the airport.

#### Describe the solution
This event heavily copies the existing helicopter crash sites, with some minor changes. The model of the helicopter crash site is very similar to the Blackhawk models, but I either removed the NATO equipment or replaced it with Russian ones. I did the same with military pilot zombies and soldier zombies so that they have Russian counterparts. Since there are no Russian body armors in the game (to my knowledge), I removed the body armor from them.

There are probably alternatives in terms of equipment and loot drops that could be added, but for now, I wanted some feedback before I continue working on this.

#### Testing
The event spawns fine in debug mode and the zombies drop their loot as planned.


Lupotech-CDDA:
PS: I am happy for all feedback and tips on how to better contribute :)

GuardianDll:
first thing first, the preferable way to add a new map extra is not using hardcoded map, but use `update_mapgen` with map in json - it would make it much easier to maintain
second, and it's my personal opinion, is that i don't think character should know that this helicopter wheckage is specifically russian, with specifically russian zombies
questions about "is it make sense to find crashed foreign heli in the usa" i'll left for someone with better knowledge of the lore

Lupotech-CDDA:
> first thing first, the preferable way to add a new map extra is not using hardcoded map, but use `update_mapgen` with map in json - it would make it much easier to maintain second, and it's my personal opinion, is that i don't think character should know that this helicopter wheckage is specifically russian, with specifically russian zombies questions about "is it make sense to find crashed foreign heli in the usa?" i'll left for someone with better knowledge of the lore

Thank you for the tips! I'll look into it.
I considered the "foreign" part, but thought that it might be a bit confusing if it wasn't specific. There could be a flag on the helicopter or uniforms for example. But in the end that isn't important to me.

RenechCDDA:
Why would a Russian military helicopter be present inside the United States, let alone be common enough to appear ingame?

Zireael07:
I imagine it's a one-time event at the time of the Cataclysm.

Diplomatic visit or something

Lupotech-CDDA:
> I imagine it's a one-time event at the time of the Cataclysm.
> 
> Diplomatic visit or something (foreign delegations often travel on military planes/helis)

yeah, that's kind of what I thought.  I made them about 1/3 to 1/5th as rare in my current version, but they could also be even rarer.

Rabadash94:
There are many mentions of China in the lore, perhaps a chinese helicopter crash would make more sense.

Nebnis:
I think just a foreign helicopter crash works here. You could add item groups from different countries like china and russia etc, and then it would be a bit more believable. Otherwise russian or even random chinese helicopter crashes wouldn't be something you find even remotely by sheer luck. Embassy in my country only exist in very specific cities for example.

tenmillimaster:
I don't think any foreign helicopter crash makes sense. The timeline doesn't allow for hostilities to escalate so quickly in my opinion. By the time it might make sense for major foreign powers with sufficient power projection to want to get a carrier vessel nearby to undertake operations behind enemy lines(April 16th, year 0, not march year 0 since china/Russia vs the USA is not a 'small war'), no major foreign power actually has either the capability to support such action or the manpower that isn't tasked with containing the outbreak.

If it was china or Russia, why not assault somewhere on western seaboard?

Zireael07:
@tenmillimaster No one is talking about foreign military operations behind enemy lines, just a diplomatic visit to see 'what the heck's goin on'

tenmillimaster:
> @tenmillimaster No one is talking about foreign military operations behind enemy lines, just a diplomatic visit to see 'what the heck's goin on'

That would be conducted with an armored sedan and a security detail, not a helicopter full of non NATO military equipment.

PatrikLundell:
Like others have said, a foreign helicopter makes absolutely no sense, as it can't make it to the USA. The closest thing you have to flying military craft flying in non friendly countries are aircraft carrying heads of state (corresponding to Air Force One) and these are not (overtly) military craft operated by (overtly) military personnel (many of these are probably equipped with spying equipment, though).

The US military has a small number of foreign craft acquired by various means for examination and for training against, but they wouldn't have the B force dress up in foreign gear as that would serve no useful training purpose.

The only way to get a crewed foreign military helo to the US is to invoke magic, specifically the disgusting (in my opinion) portal magic. And, if magic would be invoked, it would make more sense lore wise for it to be from China.

Zireael07:
Ah, I must have missed the fact that Russia and China are no longer friendly in Cata lore

Also: I already said that diplomatic visits do use military aircrafts (although given distances involved, a plane would make more sense than a heli). It shouldn't be swimming in guns, though - a couple guns for a couple bodyguards and that's all

tenmillimaster:
I think a more likely conclusion to the the strife in may-april is hostile major foreign powers recalling their ambassadors/having their ambassadors be rejected.

CHCOOH:
We may be able to change our minds.
This is slightly off the agenda, but in the event of an emergency situation, weapons may be imported from other countries due to a shortage of weapons.
This will lead to demand not only from the government, but also from some PMCs and other private companies.
Some of these imports may not have arrived properly. For example, some of these imports may have been transported to the airport via normal routes, and then helicopters were used to transport them on an emergency basis, but the helicopters crashed.
If you want to introduce firearms from other countries in the game, it may not be necessary that the courier is a member of the military.

However, it is unlikely that players will encounter the scene very often.

fairyarmadillo:
There's already a perfectly workable solution in the portal storms. We have an aircraft carrier getting teleported from Korea to a lake in New England, why couldn't a Russian helicopter meet the same fate?

Could even add a starting scenario where you were one of the passengers which explains it a little better. We should probably avoid overusing this trope, but in this case I think it's a good cause.

RenechCDDA:
The carrier gets a pass because it is an ass-ton of work and very, very cool. Its sheer size (and placement in an inland waterway) does a lot to shape the portal storm narrative, which simply throwing random stuff in does not do.

It is not appropriate to simply use portal storms as a handwave plot device to insert *everything*.

fairyarmadillo:
> The carrier gets a pass because it is an ass-ton of work and very, very cool. Its sheer size (and placement in an inland waterway) does a lot to shape the portal storm narrative, which simply throwing random stuff in does not do.
> 
> It is not appropriate to simply use portal storms as a handwave plot device to insert _everything_.

It's fine if you disagree, but that's mischaracterizing my suggestion, isn't it? I didn't say we could use it as a handwave to insert _everything_. I said we could use it as a handwave to insert _one or two more things_. Especially as once we have enough unique locations added, it will become increasingly unlikely to see some of them on a single playthrough.

RenechCDDA:
That "one more thing" is not notable, and the obvious train of thought from there is "Well, this unremarkable thing was added due to portal storms, why not another?"

The rule of thumb you should be aiming for* is that the portal storms should *only* be plot-wise responsible for *remarkable things*.

*My opinion only, etc

harakka:
The aircraft carrier is in a lake because we do not yet have an ocean to place it into.

PR #68742, opened 2023-10-18
Obsolete house_09b

### LINKED ISSUE #68737: DEBUG    : invalid overmap special id "house_09"
### Describe the bug

`DEBUG    : invalid overmap special id "house_09"
`
Started getting it on an existing save right after updating a few minutes ago.

### Attach save file

[Symsonia-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12982687/Symsonia-trimmed.tar.gz)

### Steps to reproduce

Load up the save. Also happens when you go to the overland map and move around.

### Expected behavior

No error.

### Screenshots

_No response_

- OS: Windows
- Game Version: ed2a948 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

```
 DEBUG    : invalid overmap special id "house_09"

 FUNCTION : obj
 FILE     : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\generic_factory.h
 LINE     : 478
 VERSION  : ed2a948
```

Fris0uman:
#### Summary
None

#### Purpose of change

Fixes #68737

#### Describe the solution

- Replacing house_09b with house_09
- Put definition for house_09b in an obsolete file and in mods that have use of it
- Removing house_09b from the spawn lists except in mods that will use it

#### Testing

- Load the save from the issue, before change error message, after change no error
- NOt sure how to test if the mods still work as intended


data/mods/Aftershock/maps/city_buildings.json
```diff
@@ -9,6 +9,18 @@
       { "point": [ 0, 0, -1 ], "overmap": "basement_bionic_north" }
     ]
   },
+  {
```
Maleclypse, line 12: wouldn't you need to also adjust regional map settings in the mods you are moving it to?  When I look at the mod sections and see only a single file changed, I went to go look at their regional settings files I see that Aftershock as an example only has regional overlay file.  So iirc you would need to add house_09b to the city overlay in Aftershock and any others.

Fris0uman: Should be fixed with the last commit. No Hope doesn't need a change there because it already has settings for house9 and 9b

PR #68744, opened 2023-10-18
Fix milled amount

### LINKED ISSUE #68701: problem with mill, next...
### Describe the bug

1. a new problem with the windmill, if you put 65 units of wheat, the output will be 975 flour, this is the calculation according to the description in the wheat itself, but the result is 32175

1. incredibly long unloading from the mill, it took me more than 5 minutes of real time (I'm tired of waiting I closed the game, full time in game ~330000 game seconds ~91 game hour ^^), by the way this applies not only to the mill, it also takes quite a long time to get out of the smoking rack, it seemed faster before

1.  if mill wild rice
```
DEBUG    : calc_rot_while_processing called on non smoking item: dry wild rice (fresh)

FUNCTION : calc_rot_while_processing
FILE     : D:\a\Cataclysm-DDA\Cataclysm-DDA\src\item.cpp
LINE     : 7969
VERSION  : 356bfe5

```
4. the problem starts after 4 units

### Attach save file

[Magenta-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12920952/Magenta-trimmed.tar.gz)

### Steps to reproduce

1. put wheat
2. start mill
3. get product

### Expected behavior

None

### Screenshots

console

- OS: Windows
    - OS Version: 6.1 7601
- Game Version: 356bfe5 [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Aftershock [aftershock],
    SpeedyDex [speedydex],
]

### Additional context

_No response_

Fris0uman:
#### Summary
None

#### Purpose of change

Fixes #68701 
#### Describe the solution

- Get the list of types to mill
- Get necessary associated data maps
      -  Set rot timer to the mean of all the rot timers of the milled type
      -  Set all the flags of all the item milled on each of the resulting items
- Spawn the new items
- Delete the old ones

#### Testing

- Put 10 wheat in get 150 flour out
- Put 10 dry wild rice in get 20 flour
- Load the save and put the bones in > get correct amount of bone meal

- debug spawn wild rice
- debug move time one season > the rice rots in 2 season now
- mill it > the flour spoils in 2 season

- put in equal part spoils in 1 season and fresh > get flour that spoils in 2 seasons
#### Additional context

This 3 loops not sure if it's the smartest or most efficient, probably need a review in term of code logic

### DIFF
--- a/src/iexamine.cpp
+++ b/src/iexamine.cpp
@@ -5946,10 +5946,12 @@ void iexamine::mill_finalize( Character &, const tripoint &examp, const time_poi
         item &it = *iter;
         if( it.type->milling_data ) {
             it.calc_rot_while_processing( milling_time );
+            // Unset processing flag to be able to check the relative rot
+            it.unset_flag( flag_PROCESSING );
 
             if( millable_rot.find( it.typeId() ) == millable_rot.end() ) {
                 const std::vector<double> rot_times = { it.get_relative_rot() };
-                millable_rot.emplace( it.typeId(), it.get_relative_rot() );
+                millable_rot.emplace( it.typeId(), rot_times );
             } else {
                 const double rot_time = it.get_relative_rot();
                 millable_rot[it.typeId()].emplace_back( rot_time );
--- a/src/iexamine.cpp
+++ b/src/iexamine.cpp
@@ -5936,36 +5936,80 @@ void iexamine::mill_finalize( Character &, const tripoint &examp, const time_poi
         return;
     }
 
-    for( map_stack::iterator iter = items.begin(); iter != items.end(); ) {
+
+    // Get list of milable items, their rot timer and their flags
+    std::map<itype_id, std::vector<double>> millable_rot;
+    std::map<itype_id, std::vector<flag_id>> millable_flags;
+    std::map<itype_id, item> millable_items;
+    std::vector<map_stack::iterator> iter_to_delet;
+    for( map_stack::iterator iter = items.begin(); iter != items.end(); ++iter ) {
         item &it = *iter;
         if( it.type->milling_data ) {
             it.calc_rot_while_processing( milling_time );
-            const islot_milling &mdata = *it.type->milling_data;
-            const int resulting_charges = get_milled_amount( it.typeId(), examp, here );
-            // if not enough material, just remove the item (0 loops)
-            // (may happen if the player did not add enough charges to the mill
-            // or if the conversion rate is changed between versions)
-            for( int i = 0; i < resulting_charges; i++ ) {
-                item result( mdata.into_, start_time + milling_time );
-                result.components.add( it );
-                // copied from item::inherit_flags, which can not be called here because it requires a recipe.
-                for( const flag_id &f : it.type->get_flags() ) {
-                    if( f->craft_inherit() ) {
-                        result.set_flag( f );
-                    }
-                }
-                result.recipe_charges = resulting_charges;
-                // Set flag to tell set_relative_rot() to calc from bday not now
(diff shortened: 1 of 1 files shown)


mqrause:
Something Kevin mentioned on discord is that only one type of item should be milled at a time. That should simplify this quite a bit.

PatrikLundell:
One type at a time makes sense unless you're explicitly trying to make a blend (which the game probably doesn't support). However, it would not make sense to forbid mixing of the same kind of various age (i.e. you should be able to mix stacks as long as they're of the same kind).

Also note that this differs the mill from the smoking rack, where you should easily be able to mix stuff and still get distinct products out, so a change to the logic should be specific to the mill (and similar kinds of processes, such as e.g. blenders or fluid processing containers (apart from the mixed booze variants that already exists).

src/iexamine.cpp
```diff
@@ -5936,36 +5936,80 @@ void iexamine::mill_finalize( Character &, const tripoint &examp, const time_poi
...
-            // (may happen if the player did not add enough charges to the mill
-            // or if the conversion rate is changed between versions)
-            for( int i = 0; i < resulting_charges; i++ ) {
-                item result( mdata.into_, start_time + milling_time );
-                result.components.add( it );
-                // copied from item::inherit_flags, which can not be called here because it requires a recipe.
+
+            if( millable_rot.find( it.typeId() ) == millable_rot.end() ) {
+                const std::vector<double> rot_times = { it.get_relative_rot() };
+                millable_rot.emplace( it.typeId(), it.get_relative_rot() );
```
Fris0uman, line 5952: `it.get_relative_rot()` here always returns 0 nad I have no idea why

RenechCDDA: `it.calc_rot_while_processing( milling_time );` on line 5948 is setting the rot ahead of the current time (to the end of the processing duration)

Fris0uman: When I tried commenting it away get_relative_rot was still giving me 0, and it should make things more rotten not less in any case, no?

mqrause: It might be that putting items into the mill already loses the rot data? Though I couldn't find anything obvious at a glance.

Michael1993: Probably because the mill is processing the millable, meaning the `goes_bad()` check in `get_relative_rot()` (item.cpp@7806?) returns false.

Probably this is why there is this part in the original:
```c
// Set flag to tell set_relative_rot() to calc from bday not now
result.set_flag( flag_PROCESSING_RESULT );
result.set_relative_rot( it.get_relative_rot() );
result.unset_flag( flag_PROCESSING_RESULT );
```

Fris0uman: removing the flag before getting relative rot doesn't solve the issue

Michael1993: Okay, that was my best guess (which doesn't mean much tbh) but I'll take another look.

Michael1993: Well, as far as I can see `relative_rot` is `rot / get_shelf_life()` and `rot` is only set in `set_rot` and `set_relative_rot` otherwise `last_temp_check` is used. Does that help?

Fris0uman: My bad! Removing `flag_PROCESSING` seems to solve the issue, I was messing with flag_PROCESSING_RESULT which is not the same thing.

RenechCDDA:
Thanks for all the work on this, Fris0uman! Sorry I couldn't be more help

PR #68748, opened 2023-10-18
Minor froggy edits

Karol1223:
#### Summary
None

#### Purpose of change
Ironing out some weirdness from #67259 I didn't catch while I was reviewing it but only realised after playing the game with the PR merged

#### Describe the solution
- Significantly lowered the spawnrate for most frogs and tadpoles - look, sorry, they were ridiculous. In one case the weight for old frog+toad combo was a combined 6, while the new weight given for the frog group was _**300**_. I lowered it down significantly, because adding new frogs doesn't mean frogs should suddenly be more common. On my new save which I started a while ago almost all animals I ran into have been various frogs, and dozens upon dozens of tadpoles in swamp water tiles.
- Changed frog faction from ``frog`` to ``small_animal`` and tadpole faction from ``frog`` to ``fish``. Why? Because ``frog`` looks to be made with mutant frogs in mind, since zombies go after it. This resulted in hordes of zombies chasing frogs, which with the aforementioned spawn frequency was weird to say the least
- Lowered tadpole volume from 120ml to 25ml

Messing around with the "weird" frog weights in mongroups to make them comparatively rarer - I do believe mutants, especially mild ones like these, shouldn't have 1:1 comparative weight in spawning, but I can't be bothered to go through them all. Four mongroups for frogs is too many in my book.


Karol1223:
@TheSaddestGoomba I'd like to hear your opinions on this in case I am accidentally ignoring some important detail, but I don't think I am?

TheSaddestGoomba:
Everything here seems reasonable to me. 

The spawn weights were eyeballed based on the rest of the spawn group. I believe I was aiming for close to the total insect spawn rates.
I agree that they've been too prevalent since the initial PR. Good to dial it down before they have upgrades. Reducing the ratio also seems wise. I plan upgrades for each of the 'weird' variants though their individual debuts may be a while away.

Feel free to change up the groups. My main intention with them was to reflect each species natural biome preferences (sewers were a guess). So, no tree frogs or toads in fast-flowing rivers and few water-dependent bullfrogs in the forest.

Won't zombies still pursue 'small animals'? I think that's a good change regardless.

Karol1223:
> Won't zombies still pursue 'small animals'? I think that's a good change regardless.

They're not *supposed* to... I believe they're on neutral terms, so if one doesn't attack the other they chill together.

PR #68763, opened 2023-10-19
[MoM] Add PSI_NULL species to interact with "ignored_monster_species" JSON parameter

Standing-Storm:
#### Summary
Mods "Add PSI_NULL species to interact with "ignored_monster_species" JSON parameter"

#### Purpose of change

I mentioned on #68159 that making zombie blanks immune to being telekinetically thrown into walls would require C++...but I was wrong. There's a way to do it using the ignored_monster_species parameter I introduced.

#### Describe the solution

Add the PSI_NULL species, designed to be added to any monster that should not be targetable with psionics, and add "ignored_monster_species": [ "PSI_NULL" ] to every power that can target a hostile or ally target. Add PSI_NULL species to the zombie blank line, making them immune to being targeted by any powers (their armor prevents them from taking damage from the AoE of powers targeted elsewhere.

#### Testing

Everything works!

#### Additional context

I guess technically this does "require C++," but it's C++ I already wrote so that's fine.


Maleclypse:
Well this is cool.  I don't think I'd recognized the value of that field

PR #68764, opened 2023-10-19
[MoM] Snippetathon, part 2

Standing-Storm:
#### Summary
Mods "Mind Over Matter Snippetathon, part 2"

#### Purpose of change

Need more lore snippets, especially in telepathy and clairsentience testing since I'm working on a themed lab location for those branches of psionics. 

#### Describe the solution

Add the snippets.

#### Additional context

The goal is to have one themed lab dedicated to studying clairsentience and telepathy, one themed lab dedicated to studying biokinesis and vitakinesis, one themed lab dedicated to studying telekinesis and pyrokinesis, and outposts in the main microlabs dedicated to studying teleportation, and then the Research Facility as a general Project PHAVIAN headquarters.


data/mods/MindOverMatter/snippets/snippets_new.json
```diff
@@ -278,7 +357,7 @@
...
...ress the recent rumors.  I know that stress levels are high with the quarantine and the XE037 news, but there is no need to engage in unwarranted speculation.  In these unprecedented times, it's important to remember that we're all in this together and we will solve the problems before us, as we have no many others.\n\nI will be perfectly clear.  There is no \"Project GASPAR.\"  Trained mathematicians are the most important resource that we have, and it would be the height of folly to waste them on replicating a resource that already exists simply because the current state of emergency has placed us out of contact.  Furthemore, it is important to remember that Netherum mathematics are not magic.  They obey a consistent set of rules and we are capable of determining those rules empirically.  We can test how much a JUNIPER GOLD mathematician can lift to the milligram.  We can determine the effective communication range for a GRAY DAWN candidate down to the decimeter.  We would never simply \"throw things at the wall and see what sticks.\"  Science requires rigor.\n\nThese are troubling times, but we will get through them together.\n\nSincerely,\nHarold Liu\nDirector, Project PHAVIAN"
```
Maleclypse, line 360:
```suggestion
        "text": "Colleagues,\n\nI'd like to take a moment to address the recent rumors.  I know that stress levels are high with the quarantine and the XE037 news, but there is no need to engage in unwarranted speculation.  In these unprecedented times, it's important to remember that we're all in this together and we will solve the problems before us, as we have so many others.\n\nI will be perfectly clear.  There is no \"Project GASPAR.\"  Trained mathematicians are the most important resource that we have, and it would be the height of folly to waste them on replicating a resource that already exists simply because the current state of emergency has placed us out of contact.  Furthermore, it is important to remember that Netherum mathematics are not magic.  They obey a consistent set of rules and we are capable of determining those rules empirically.  We can test how much a JUNIPER GOLD mathematician can lift to the milligram.  We can determine the effective communication range for a GRAY DAWN candidate down to the decimeter.  We would never simply \"throw things at the wall and see what sticks.\"  Science requires rigor.\n\nThese are troubling times, but we will get through them together.\n\nSincerely,\nHarold Liu\nDirector, Project PHAVIAN"
```

Maleclypse: no many others to so many others and furthermore was mispelled since the gitdif isn't showing the changes.

data/mods/MindOverMatter/snippets/snippets_new.json
```diff
@@ -146,18 +158,78 @@
...
... the other hand, NOBLE HORIZON consistently managed to guess at least three digits and successfully guessed the entire six-digit sequence twice (out of ten total sequences).  This disparity was maintained when both groups were placed in emtpy rooms, instructed to write down their guesses on a sheet of paper, and place it into a box before leaving the room.  Finally, NOBLE HORIZON attained a better than chance score on the final test, when both groups were instructed to guess the sequence before Melchior was asked to provide a sequence.\n\nThis leads to two conclusions.  The first is that the alarmists are wrong; Melchior is not secretly harboring a genius intellect and plotting to control XEDRA from within, it is a computer, capable of processing vast amounts of data and using netherum mathematics to process that data much faster than would otherwise be possible.  The second is that while GRAY DAWN requires a specific source from which to attain new information, NOBLE HORIZON does not.  Even in situations where no information transfer would be possible, such as the final two testing scenarios, NOBLE HORIZON still had a better-than-chance rate of correctly determining the sequence."
```
Maleclypse, line 198:
```suggestion
        "text": "In the interests of testing Dr. Al-Masri's hypothesis, we assembled a group of NOBLE HORIZON and GRAY DAWN mathematicians and had them both attempt to conduct a simple number guessing game.  Melchior provided a six-digit number and both teams attempted to guess it.  GRAY DAWN were competitive when the number was overseen by a third party, as expected, but when both teams were provided with a terminal interface and instructed to enter their guesses, GRAY DAWN scored no better than chance.  On the other hand, NOBLE HORIZON consistently managed to guess at least three digits and successfully guessed the entire six-digit sequence twice (out of ten total sequences).  This disparity was maintained when both groups were placed in empty rooms, instructed to write down their guesses on a sheet of paper, and place it into a box before leaving the room.  Finally, NOBLE HORIZON attained a better than chance score on the final test, when both groups were instructed to guess the sequence before Melchior was asked to provide a sequence.\n\nThis leads to two conclusions.  The first is that the alarmists are wrong; Melchior is not secretly harboring a genius intellect and plotting to control XEDRA from within, it is a computer, capable of processing vast amounts of data and using netherum mathematics to process that data much faster than would otherwise be possible.  The second is that while GRAY DAWN requires a specific source from which to attain new information, NOBLE HORIZON does not.  Even in situations where no information transfer would be possible, such as the final two testing scenarios, NOBLE HORIZON still had a better-than-chance rate of correctly determining the sequence."
```

Standing-Storm: Thank you for double-checking all of this text!

PR #68768, opened 2023-10-19
Steel Kevlar Jumpsuit updated recipe

Termineitor244:
#### Summary
Balance "Update the recipe of Steel Kevlar Jumpsuits for consistency"

#### Purpose of change
steel-plated Kevlar jumpsuits are made out of *only* kevlar sheets, not allowing use of kevlar patchwork sheets like the rest of kevlar jumpsuits, in addition to that they use a tremendously big amount of kevlar even though they only use it for one layer, which is a little thicker than the kevlar base layer of the other jumpsuits, but they have several layers of the component which end up with a bigger kevlar thickness in the other 2, those extra layers are replaced with steel so it makes little sense to ask for more than double the amount of kevlar of normal kevlar jumpsuits.

#### Describe the solution
Replace the direct calling for kevlar sheets and thread for the `tailoring_kevlar_fabric` requirements, which is in use by the other 2 kevlar jumpsuits and allows for a choice between both types of kevlar sheets.
Reduced the number of sheets required to be in sync with the number of kevlar scraps of the recipe (As is the case with the normal kevlar jumpsuit), could (Probably should) be even lower, but I was unsure what specific number to give so I opted for a relationship with the already defined scraps.

To divide the kevlar requirements in 2 so you could combine both types of kevlar sheets for the recipe (You still need a lot of one specific kind)? Something that probably should be part of the base crafting interactions.

#### Testing
Loaded the game, searched for the recipe... It works! 

#### Additional context
imagen
Updated the XS and XL versions too.
They all now require a big amount of strong thread too, but that is easier to acquire/produce than kevlar sheets.

Edit: Removed duct tape and other relics from the recipe and made it more similar with the regular kevlar jumpsuits.

### DIFF
--- a/data/json/recipes/armor/bespoke_armor/survivor.json
+++ b/data/json/recipes/armor/bespoke_armor/survivor.json
@@ -200,7 +200,6 @@
       { "proficiency": "prof_polymerworking" },
       { "proficiency": "prof_elastics" }
     ],
-    "qualities": [ { "id": "FABRIC_CUT", "level": 2 } ],
     "byproducts": [ [ "scrap_kevlar", 17 ], [ "scrap_lycra", 7 ] ]
   },
   {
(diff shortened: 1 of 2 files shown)


KeremBabaG:
370 seems awfully high imo

Termineitor244:
> 370 seems awfully high imo

It was 590 before... And it can always be further reduced, I just don't know by how much.

PR #68776, opened 2023-10-20
Adding le BB Pistol

ghost:
#### Summary
Content "Added a BB pistol"

#### Purpose of change
There is a BB gun for rifles, but not one for pistols. So here is one for pistols.

#### Describe the solution
I tried to make a bb pistol. Spring-loaded, of course; hopefully it reflects that.

Having someone who actually knows what they're doing attempt this.

#### Testing
Seems to work, but I don't really know what I'm doing unfortunately. 

#### Additional context
Not exactly confident I know what I'm doing here.


Delafaith:
Ok clearly it doesn't work ┌(。Д。)┐

PR #68781, opened 2023-10-21
fix: player_dispaly false for several npc traits

casswedson:
#### Summary
None

#### Purpose of change
I was watching a *ahem* [Worm Girl](https://www.youtube.com/@mutantwormgirl) video and after she recruited an NPC she opened its stat sheet and there were a couple of NPC hidden traits on display; fix that

Oh! also remove 21 mm of rubber from the western boots sole

#### Testing
should be fine without testing

### DIFF
--- a/data/json/items/armor/boots.json
+++ b/data/json/items/armor/boots.json
@@ -962,7 +962,6 @@
         "specifically_covers": [ "foot_sole_r", "foot_sole_l" ],
         "material": [
           { "type": "leather", "covered_by_mat": 100, "thickness": 0.1 },
-          { "type": "rubber", "covered_by_mat": 20, "thickness": 21.0 },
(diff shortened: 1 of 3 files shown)


Maleclypse:
Apparently you can't currently have an activated mutation that is also invisible to the player.  Probably need a flag or some other type of jobject.

```
(continued from above) ERROR : src/mutation_data.cpp:821 [static void mutation_branch::check_consistency()] Requested argument 0 but input has only 0 at: "mutation %s| is not displayed but set as active"
(~[slow] ~[.],starting_items)=> 01:19:33.425 ERROR : src/mutation_data.cpp:821 [static void mutation_branch::check_consistency()] Requested argument 0 but input has only 0 at: "mutation %s| is not displayed but set as active"
```

PR #68784, opened 2023-10-21
Fix skill rust

### LINKED ISSUE #68777: Skill Rust Not Triggering
### Describe the bug

After several people were asking if skill rust was removed, we were talking about it and I started testing it.  After 9 days, I had no skill rust at all, so I'm assuming it's broken/would have triggered by then.  So, looks like skill rust is not triggering.  Version below with the game info stuff, it was the most recent experimental as of the time of me typing this.

### Attach save file

N/A

### Steps to reproduce

Yeah so to do this I just waited until daylight for 9 days, debugging away my fatigue and eating/drinking normally.  I assume that debug time travel would not trigger rust, so I didn't want to do it that way.  So there doesn't seem to be a fast/easy way to test it.

Something like 

1. Have skills
2. Wait a long time
3. Observe no skill rust

### Expected behavior

I expect skill rust to trigger after this period of time.  I think it used to trigger after only a few days.  

### Screenshots

_No response_

- OS: Windows
- Game Version: ee51041 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda]
]

### Additional context

I was not using any of the traits that modify skill rust in any way.  Now that I think about it, I did debug my skills in this scenario as I did not have a character ready to test, it is possible that debug setting skills may have interfered with rust in some way?  But since other people were reporting this, that seems unlikely to be a factor.

ehughsbaird:
#### Summary
SUMMARY: None

#### Purpose of change
Fix https://github.com/CleverRaven/Cataclysm-DDA/issues/68777
One line too many was deleted in https://github.com/CleverRaven/Cataclysm-DDA/pull/68490, resulting in skill rust no longer functioning.

#### Describe the solution
Restore the accidentally deleted line and add a regression test.

#### Testing
Test fails before, passes after.

### DIFF
--- a/tests/skill_test.cpp
+++ b/tests/skill_test.cpp
@@ -1,0 +1,44 @@
+#include "calendar.h"
+#include "cata_catch.h"
+#include "character.h"
+#include "player_helpers.h"
+#include "skill.h"
+
+TEST_CASE( "skill_rust_occurs", "[character][skill]" )
+{
+    Character &guy = get_player_character();
+    clear_avatar();
+
+    // Set every skill to two, so they can rust
+    for( const Skill &skill : Skill::skills ) {
+        INFO( skill.ident().str() );
+        guy.set_skill_level( skill.ident(), 2 );
+        REQUIRE( guy.get_skill_level( skill.ident() ) == 2.f );
+        REQUIRE( guy.get_knowledge_level( skill.ident() ) == 2 );
+        REQUIRE_FALSE( guy.get_skill_level_object( skill.ident() ).isRusty() );
+    }
+
+    // Wait three days
+    for( int i = 0; i < to_seconds<int>( 3_days ); ++i ) {
+        calendar::turn += 1_seconds;
+        guy.update_body();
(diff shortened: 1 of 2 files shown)


tests/skill_test.cpp
```diff
@@ -0,0 +1,43 @@
...
+{
+    Character &guy = get_player_character();
+    clear_avatar();
+
+    // Set every skill to two, so they can rust
+    for( const Skill &skill : Skill::skills ) {
+        INFO( skill.ident().str() );
+        guy.set_skill_level( skill.ident(), 2 );
+        CHECK( guy.get_skill_level( skill.ident() ) == 2.f );
+        CHECK( guy.get_knowledge_level( skill.ident() ) == 2 );
```
andrei8l, line 17: Use `REQUIRE` in the setup so the test halts early.
```suggestion
        REQUIRE( guy.get_skill_level( skill.ident() ) == 2.f );
        REQUIRE( guy.get_knowledge_level( skill.ident() ) == 2 );
```
Also test `isRusty()` before and after.

PR #68789, opened 2023-10-21
New characters don't start with martial arts style set to 'Force unarmed'

### LINKED ISSUE #68722: New characters start with martial arts style set to "Force unarmed"
### Describe the bug

Every time I create a new character, they start with their martial arts style set to "Force unarmed". This caused a lot of confusion and a few deaths before I figured out what was going on. 

### Attach save file

[Leicester-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/12937616/Leicester-trimmed.tar.gz)

This really shouldn't matter, but fwiw here's a totally vanilla character on a totally vanilla world, who started set to Force unarmed

### Steps to reproduce

1. Create character
2. Check their martial arts style

### Expected behavior

Characters start set to "no style" (or an actual style that they started with)

### Screenshots

_No response_

- OS: Windows
- Game Version: 078a7b4 [64-bit]
- Graphics Version: Tiles
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

_No response_

gtim:
#### Summary
Bugfixes "New characters don't start with martial arts style set to 'Force unarmed'"

#### Purpose of change

New characters default to a randomly selected martial arts style. As such, characters without martial arts knowledge randomly select style_no (desired) or style_kicks (undesired). This PR ensures style_kicks is never selected. Fixes #68722.

#### Describe the solution

Removed style_kicks from the list of randomly-selectable martial arts on character initialization.

It's not clear to me if this is the desired behaviour. Alternatives are:
- Characters always start with no style selected, even if the character knows an "interesting" (not style_none or style_kicks) style.
- Characters always start with an interesting style if the character knows one, and style_none if not.
- Either of the above, but they never start with other "force unarmed" martial styles either.

#### Testing

Tested by generating 40 martial-arts characters and non-martial-arts characters before and after changes. I'm not sure in what circumstance characters start with multiple known styles, let me know and I'll test that as well. The test suite ran fine.

If requested, I'd be happy to write a test case to ensure this doesn't regress in the future.

#### Additional context

This is my first contribution and I didn't get astyle working right away -- apologies in advance.

### DIFF
--- a/data/json/martialarts.json
+++ b/data/json/martialarts.json
@@ -5,6 +5,7 @@
     "name": { "str": "No style" },
     "description": "Not a martial art, this is just plain old punching and kicking.",
     "initiate": [ "You decide to not use any martial arts.", "%s enters a generic combat stance." ],
+    "priority": 0,
     "arm_block": 1,
     "leg_block": 99,
     "allow_all_weapons": true,
@@ -16,6 +17,7 @@
     "name": { "str": "Force unarmed" },
     "description": "Not a martial art, setting this style will prevent weapon attacks and force punches (with a free hand) or kicks.",
     "initiate": [ "You force yourself to fight unarmed.", "%s decides to fight unarmed." ],
+    "priority": -1,
     "arm_block": 1,
     "leg_block": 99,
     "force_unarmed": true,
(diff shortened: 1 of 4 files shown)


ehughsbaird (commented):
If you're up for it, a more extensible aproach would be making this a property of the martial art, instead of excluding one style in hardcode.
You could also add a "priority" field, sort by that field, then pick randomly from the ones with the highest value.

gtim:
> If you're up for it, a more extensible aproach would be making this a property of the martial art, instead of excluding one style in hardcode. You could also add a "priority" field, sort by that field, then pick randomly from the ones with the highest value.

Thank you for the feedback, it did feel a tad hard-coded. Happy to solve it using a priority field if no-one disagrees.

PR #68796, opened 2023-10-21
Body heat temperature refactor to standard units.

prharvey:
#### Summary
Infrastructure "Replaced the use of the arbitrary body temperature scale in game logic with units::temperature/units::temperature_delta"

#### Purpose of change

The current arbitrary body temperature scale makes the code extremely hard to understand and change. It caused people to mix units arbitrarily, and there are a number of bugs that this has caused (and most of the time spent in this PR was recreating these bugs so game balance was not affected). 

A short list of notable bugs:
- Documentation mentions the scale was 100u = 1C, when the scale was actually 500u = 1C. This makes most contributions to body temperature 5x less effective than was likely intended. e.g. fire only warms up 0.6 C, not 3 C.
- ~~Related to the above, the code wanted the comfortable temperature to be 31 C when sleeping, and 19 C when not sleeping. The way it was actually implemented resulted in the comfortable temperature being 33.2 C when sleeping and 30.8 C when not sleeping.~~ This was actually the same bug as above. The actual ambient temperature calculation was being done on the 100u == 1 C scale, but then added to the body temperature incorrectly. I preserved this bug as well.
- Windchill was added to this scale based on its value in Fahrenheit, not Celsius. This would have made it 9/5x more effective than intended, but this also had the same bug as above so it ended up 9/25x as effective as likely intended.
- ~~The comments mention having a temperature convergence half life of 15 minutes, but it actually converges almost instantly.~~ This was actually working correctly, was just slightly hard to catch. I refactored it to make it more obvious.

Note that this doesn't change how temperature values are stored in save files or JSON, so the change is forward/backward compatible. It also doesn't change how the value is displayed to the user in non-debug displays, so no translations are needed.

This change is meant to pave the way for a larger change I wanted to make with replacing the entire equipment warmth system with an insulation system, but the current state of the code made that nearly impossible to both do all at once and have a reasonable PR. It also means it isn't worth it changing the JSON warmth values, since they would just be changed again anyways.

I'll leave it up to others to decide if/when/how save files should be migrated/broken.

#### Describe the solution

Global replace of the arbitrary int body temperature scale with units::temperature / units::temperature_delta, and migrates the magic constants to use the same types.

Not do this and implement other changes on top of it. Not really a good solution since the code has a ton of bugs, which this refactor exposed. I likely wouldn't have even noticed them without it.

#### Testing

Since this doesn't actually add any functionality, I focused on making it pass the tests currently in the suite. I don't know enough about the code base to say if that is adequate or not.


prharvey:
Good news, I was actually wrong about the convergence being instant. I had just read the code wrong as it was slightly obscured. I refactored that part to be more explicit about what was happening.

prharvey:
After sleeping on it, I realized that I was handling the bugged conversion from ambient temperature to body temperature incorrectly. I was scaling the based ambient temperature before using it in calculations, instead of after the calculations as I should have. This lets the code keep the original 31 C when sleeping and 19 C while awake, but correctly scales down the contribution of the delta to be 5x less than intended.

src/player_hardcoded_effects.cpp
```diff
@@ -829,13 +829,15 @@ static void eff_fun_hypovolemia( Character &u, effect &it )
                     warning = _( "You are anxious and cannot collect your thoughts." );
                     u.mod_focus( -rng( 1, u.get_focus() * intense / it.get_max_intensity() ) );
                     break;
-                case 4:
+                case 4: {
```
Brambor, line 832: I guess adding this bracket isn't necessary.

prharvey: I'm declaring a new variable within the case statement, so the scope is required.

mqrause:
> Note that this doesn't change how temperature values are stored in save files or JSON, so the change is forward/backward compatible.

Usually this is handled by allowing to read both types of values but storing just the new one. In some future version reading the legacy values can then be removed. Compatibility is only required from one stable version to the next, I believe.

PR #68802, opened 2023-10-22
[MoM] Add tier 3 feral telepath, the Instiller of Nightmares + fix some issues

Standing-Storm:
#### Summary
Mods "Add tier 3 feral telepath, the Instiller of Nightmares"

#### Purpose of change

There's a tier 3 feral biokinetics, clairsentients, and pyrokinetics, and since I'm working on a themed clairsentience/telepathy lab, there should be a tier 3 telepath you might encounter there too.

#### Describe the solution

Add the telepath.

Also I fixed a few issues here and there. I capped the move_speed bonus of Perfected Motion (at 25x faster than base) and of Altered Momentum, since someone on Discord had a screenshot of negative move cost. I actually think it's reasonable to move between raindrops (so to speak) for short periods of time, but it should require a lot of effort to get there.

I also added an effect to Primal Terror that's a short lingering debuff. Sure, the feral telepaths will use it on you, but you can use it right back on them (or could if they didn't have Telepathic Shield up, but you can use it on other people).

#### Testing

Everything works. 

At the moment, Instillers of Nightmares have the ALL_SEEING flag, with the idea that they're sensing the player's mind using their powers. This also allows them to target the player through walls. I don't really have a problem with this--player telepaths can do exactly the same thing--but I anticipate some angry ALT-F4s in the future.

The best option would be a version of ALL_SEEING that cannot see the player if they have the TEEPSHIELD flag on, but I'll have to do more studying of the code before I figure out how to do that.

#### Additional context
 I do have a plan to add a player telepathic power that lets them see targets with minds (like motion sense or nether sense enchantments), once I figure out how to do it. 

I'm also not 100% happy with the name but it'll do for now.


Maleclypse:
Tormenting Phantasmagoria

Bete Noire

as possible names for the ferals.

Standing-Storm:
> Tormenting Phantasmagoria
> 
> Bete Noire
> 
> as possible names for the ferals.

I love the name Bete Noire but it feels a little too fantastical for MoM. I'll have to keep that in mind though!

PR #68804, opened 2023-10-22
[MoM] Add Sense Minds telepathic power

Standing-Storm:
#### Summary
Mods "Add Sense Minds Telepathic power"

#### Purpose of change

It makes sense that a telepath would be able to detect nearby minds, so this is my attempt to make that happen.

#### Describe the solution

Adds a SIGHT_RANGE_MINDS enchantment that detects targets with sapient minds (NPCs, cyborgs, triffids, mi-go, nether beings, any monster with the HUMAN flag, and blob lieutenants). Add a telepathic power to take advantage of that.

~~Also add a TRIFFID species (to separate them out from what are presumably their bio-engineered servitor species) and a BLOB_LIEUTENANT species flag as well.  I've been thinking of adding a lieutenant into MoM too.~~

**Edit**: In the classic tradition of programming, I had a solution and then thought "That's dumb, there's a better way to do that," so I'm doing it the better way. Rather than adding new species and hard-coding various species as having minds, I'm instead adding a "HAS_MIND" monster flag. This does mean I'll have to manually add it to dozens of monsters but it makes Sense Minds much more flexible, extendable, and mod-compatible than having a hard-coded list of species it affects. So the current logic is:

1. Does the target have the PSI_NULL species? It cannot be detected.
2. Does the target have the HAS_MIND flag? It can be detected
3. Does the target have the ZOMBIE species? It cannot be detected.
4. Does the target have the HUMAN flag? It can be detected
5. All other targets cannot be detected

In main game, I'm going to include: triffids but not their servitor species like biollante or vinebeasts, Exodii cyborgs, mi-go, most Nether creatures except kreck, unseen hunters, or albino penguins (I'm assuming those are more like Nether animals), Hounds of Tindalos _and their copies_, the memory and reflection Portal Storm monsters (the whole point is they're supposed to be indistinguishable from the real thing), some heavily mutated humans like the shelled mutant or ungulate mutant (but not the bestial mutant), the Mycus control nodes like towers and spires, and Shadow. Eaters are not included because they aren't independent creatures (Eaters are >_fingers_<. You are >_many bubbles_<). In a future PR I'll go through in-repo mods and add HAS_MIND to appropriate monsters

#### Testing
image
Top row left to right: Ungulate mutant, biollante, kreck, Exodii quadruped, bestial mutant.
Middle row left to right: Manhack, prototype cyborg, shoggoth, impossible shape, Shadow (the lieutenant)
Bottom row left to right: An NPC, a zombie, a mi-go, a bear, and a triffid

With Sense Minds:
image


Maleclypse:
Does the bestial mutant not have a mind?

Standing-Storm:
> Does the bestial mutant not have a mind? It looks like it wasn't sensed in the bottom image

I made a judgement call there that as the far-more-mutated-than-the-PC-can-get version of the Bear mutant line, the bestial mutant had degenerated into animal instinct only (basically a few tiers beyond the Apex Predator mutation). If that's not an intended lore outcome of the experimental mutation process, I can change it.

Maleclypse:
> > Does the bestial mutant not have a mind? It looks like it wasn't sensed in the bottom image
> 
> I made a judgement call there that as the far-more-mutated-than-the-PC-can-get version of the Bear mutant line, the bestial mutant had degenerated into animal instinct only (basically a few tiers beyond the Apex Predator mutation). If that's not an intended lore outcome of the experimental mutation process, I can change it.

I more wanted to understand if you had a justification rather than if it were an accident :)

PR #68807, opened 2023-10-22
Anvil disassembly recipes

Karol1223:
#### Summary
None

#### Purpose of change
Due to recent anvil changes I ended up with a spare anvil that's useless to me, since they can no longer be used in item form. I opted to allow myself to take it apart.

#### Describe the solution
- Add new disassembly recipes for all four anvil versions
- Edit the railroad piece disassemblies to no longer use skills, the big one to take ``BRISK_EXERCISE`` and the half one to use an angle grinder instead
- Took away the skills from disassembling of small metal sheets while I'm here
- Slightly cleaned up weight syntax for anvil items (no, it doesn't change anything, but useless zeroes annoy me)


Paprika-H:
No alternatives? Shame.

Karol1223:
> No alternatives? Shame.

We cannot support multiple disassembly recipes for one item. And it feels more sensible to use an angle grinder than to hacksaw at a solid block of metal. I'm not sure a manual hacksaw would survive the whole ordeal, frankly, but I've not worked with hacksaws much.

Paprika-H:
> > No alternatives? Shame.
> 
> We cannot support multiple disassembly recipes for one item. And it feels more sensible to use an angle grinder than to hacksaw at a solid block of metal. I'm not sure a manual hacksaw would survive the whole ordeal, frankly, but I've not worked with hacksaws much.

I'm not being serious lmao, don't worry. Just making fun of how little the alternatives section is used at points

Brambor:
The deconstruction requirement of fabrication 1 makes sense. I myself would be pretty wary of trying to saw metal sheets with an angle grinder or a metal saw. I would be much more wary than when chopping wood or sawing a stick, which I have done hundreds of times.

Karol1223:
> If you're completely unskilled at fabrication, I would suggest using tools to cut things up would train you in the usage of tools to cut things up (to produce parts of suitable lengths and shapes for construction). Also, the trend seems to make most skills start at levels a bit above zero for starting characters, in which case training fabrication to a level almost all characters should start with won't affect most characters.

I can agree with this with metal sheets, in this case, as their disassembly yields smaller usable sheets.

All other items affected here, however, are just being scrapped with no regard for usable sheets - splitting rails in half or grinding an anvil into chunks isn't producing parts of suitable shapes and sizes for anything.

Karol1223:
Agree to disagree, I suppose. It's entirely possible I am wrong, but I do not believe that cutting apart pieces of metal should teach you anything if the resulting scrap metal isn't particular shape and size. At that point the only thing you're focusing on while doing it is not cutting your fingers off. That's not gonna train your skill.

PR #68811, opened 2023-10-22
Make some strings from ``Character::can_wield`` sound less artificial

Karol1223:
#### Summary
None

#### Purpose of change
 #68810 is now showing these strings, and not all are particularly natural-sounding. And one was missing a ``the``. So I figured I'd clean this up so the two PRs could be merged in close proximity.

### DIFF
--- a/src/character.cpp
+++ b/src/character.cpp
@@ -7664,10 +7664,10 @@ ret_val<void> Character::can_wield( const item &it ) const
                    _( "You need at least one arm available to even consider wielding something." ) );
     }
     if( it.made_of( phase_id::LIQUID ) ) {
-        return ret_val<void>::make_failure( _( "Can't wield spilt liquids." ) );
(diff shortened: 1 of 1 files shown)


src/character.cpp
```diff
@@ -7697,7 +7697,7 @@ ret_val<void> Character::can_wield( const item &it ) const
             return ret_val<void>::make_failure( _( "The %s can't be wielded with only one arm." ),
```
oosyrag, line 7697:
```suggestion
            return ret_val<void>::make_failure( _( "You can't wield the %s with only one arm." ),
```

oosyrag (changes requested):
Trying out reviewing! And maybe restart the tests and bump this to get merged ;)
Looks like I can't make suggestions for lines that aren't close to changed lines?
I'd also suggest 7685 - "The %s is preventing you from wielding the %s." to ""The %s prevents you from wielding the %s."
These are mostly inconsequential though so feel free to resolve without changes.

PR #68814, opened 2023-10-22
Boat hulls sinking when damaged

DeCEll-1:
#### Summary
None

#### Purpose of change
Damaged vehicle tanks/reactors/batteries can leak their contends so damaged boat hulls should leak air and be pratically useless too

#### Describe the solution
If the parts health is less than 50% (Didn't had balance or realism in mind when setting the thresold) then make the hull don't work by not adding it to `floating` list which causes it to not effect the clearance part in draft/clearance making it dead weight

#### Testing
Spawn damaged boat with debug> check ratio> repair leaking part> **move around a bit**> check again> ratio changed
Had to move around a bit because the ratio didn't updated exactly when i repaired the part and needed to either reopen the menu or spend some turns, don't know why it happens nor how to fix it

#### Additional context

Wanted to add a "foam-filled" boat hull which doesn't leak but floats less and less when its damaged but i don't have enough knowledge of foams and don't know how to make it effect clearance less and less.

Added flag NO_LEAK which makes the parts floating not get effected by damage

[THE HULL HAS BEEN BREACHED AND THE AIR IS LEAKING OUT](https://youtu.be/5cVsPPszvJM?si=JogRmhtPHRXAyDAB)


ZhilkinSerg:
Maybe make it a numeric damage threshold instead of a flag?

DeCEll-1:
Like adding a threshold value so it leaks when it passes that? Yep i can try that, gonna start it and should be PRed in a week if i dont get lazy. Thanks for the suggestions

Nebnis:
Some scenarios like the light tower and water lab might be impossible with this change because the boats there spawn badly damaged.

src/vehicle_part.cpp
```diff
@@ -136,6 +136,11 @@ std::string vehicle_part::name( bool with_prefix ) const
             break;
         }
     }
+    // Update the vehicle.cpp line 6127 when you change this %
```
RenechCDDA, line 139: This should be defined as a constant in the header if it's used in multiple places

PatrikLundell:
I would suggest the usage of the same damage threshold for hulls as the one used for batteries starting to drain and (vehicle) tank starting to leak (and I make the possibly optimistic assumption that they use the same threshold).

src/vehicle_part.cpp
DeCEll-1: Thanks commited it 👍

DeCEll-1:
> I would suggest the usage of the same damage threshold for hulls as the one used for batteries starting to drain and (vehicle) tank starting to leak (and I make the possibly optimistic assumption that they use the same threshold).

Thanks and yes i use 0.5 which is what `is_leaking()` returns but it also checks if the part is battery, tank or reactor so i didn't wanted to touch it thinking it might break something.

PR #68818, opened 2023-10-22
Professions can start with specific recipes

Hymore246:
#### Summary
Balance "Professions can start with specific recipes"

#### Purpose of change

Professions can start with specific recipes

#### Describe the solution

Professions can now be assigned specific recipes to automatically learn at character generation.

This update adds the following field to the profession definitions:
**recipes** - Profession automatically learns all listed recipes at the start of the game. The values used in this field are the recipe ids ("dounut_holes") but these will be displayed in-game as the actual names of the recipes. This field is optional and blank by default meaning that the profession does not know any recipes.

As it stands, professions automatically learn recipes related to their starting skills at character generation. For example, a Chef with 6 in Food Handling will know a large number of cooking recipes right at the start. So at a glance, this change is less impactful then it seems but I still feel that this option should exist for the professions to learn recipes that are outside their skill set.


src/newcharacter.cpp
```diff
@@ -2109,6 +2114,14 @@ static std::string assemble_profession_details( const avatar &u, const input_con
             assembled += prof->name() + "\n";
         }
     }
+    // Recipes
+    std::vector<recipe_id> prof_recipe = sorted_profs[cur_id]->recipes();
+    if( !prof_recipe.empty() ) {
+        assembled += "\n" + colorize( _( "Profession recipes:" ), COL_HEADER ) + "\n";
```
ZhilkinSerg, line 2120: If you are displaying profession recipes, you should probably display all known recipes (those autolearned from profession skills too) and probably display customized character recipes (based on profession and selected skills).

Night-Pryanik:
I think you should add at least one example of profession recipes to some profession, otherwise all your additions will be dead code.

Termineitor244:
I think Night-Pryanik is right. This feature should be actually used somewhere so it doesn't get lost. Some niche professions could use it for niche items, I don't know, maybe the Otaku knows the recipe for some cosplays already? Or the churl could know some survivalist recipes that are appropriate for their time (but are not automatically learned or the profession does not start with enough skill to know them at start).

Related to this... Does it work with hobbies, too?

src/newcharacter.cpp
Hymore246: Not sure this is doable at profession selection. Because the character doesn't exist yet, it isn't possible to check if the requirements for a recipe have been met by the player. I am also concerned about displaying a huge number of default recipes for each profession that do not require any skills to know.

Hymore246:
This does not work with Hobbies. I am sure the code can be ported over once the PR is merged.

src/newcharacter.cpp
ZhilkinSerg: What's the point of showing specific subset of recipes then?

Maleclypse:
Clang is angry
```
Error: /home/runner/work/Cataclysm-DDA/Cataclysm-DDA/src/newcharacter.cpp:890:24: error: Redundant parentheses. [cata-redundant-parentheses,-warnings-as-errors]
        learn_recipe( &( recipe_dict.get_craft( recipe->result() ) ) );
                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                       recipe_dict.get_craft( recipe->result() )
Error: /home/runner/work/Cataclysm-DDA/Cataclysm-DDA/src/newcharacter.cpp:890:26: error: static member accessed through instance [readability-static-accessed-through-instance,-warnings-as-errors]
        learn_recipe( &( recipe_dict.get_craft( recipe->result() ) ) );
                         ^
note: this fix will not be applied because it overlaps with another fix
```

src/newcharacter.cpp
```diff
@@ -885,6 +885,12 @@ void Character::initialize( bool learn_recipes )
         add_proficiency( pri );
     }
 
+    // Add profession recipes
+    for( const recipe_id id : prof->recipes() ) {
+        const recipe &r =  recipe_dictionary::get_craft( id->result() );
```
Night-Pryanik, line 890:
```suggestion
        const recipe &r = recipe_dictionary::get_craft( id->result() );
```

src/newcharacter.cpp
ZhilkinSerg: I don't get why we would display this specific recipe and won't display any other recipes during character creation.

Hymore246: You can't check for skill requirements for recipes because the player hasn't been assigned any skill increases yet. So if the profession gives a bonus to a skill, you can't check that as far as I know. You can show every recipe that isn't a practice recipe, isn't marked with a SECRET FLAG, or defined in a profession in the professions JSON file (with this PR) but that's it. I would need a way to check for skill requirements without a fully initialized player character.

src/newcharacter.cpp
```diff
@@ -885,6 +885,12 @@ void Character::initialize( bool learn_recipes )
         add_proficiency( pri );
     }
 
+    // Add profession recipes
+    for( const recipe_id id : prof->recipes() ) {
```
Night-Pryanik, line 889:
```suggestion
    for( const recipe_id &id : prof->recipes() ) {
```

src/newcharacter.cpp
```diff
@@ -885,6 +885,12 @@ void Character::initialize( bool learn_recipes )
...
+        const recipe &r = recipe_dictionary::get_craft( id->result() );
+        learn_recipe( &r );
```
Night-Pryanik, line 891: Nitpick, but since `recipe` is used only once, you may avoid its explicit declaration and put its value directly inside `learn_recipe`.

Hymore246: I tried that originally with `learn_recipe( &( recipe_dictionary::get_craft( id->result() ) ) );` but Clang threw a fit because it doesn't like using `&( )` inline for whatever reason. So I had change it.

data/json/professions.json
```diff
@@ -3458,6 +3458,7 @@
     "points": 3,
     "skills": [ { "level": 6, "name": "traps" }, { "level": 4, "name": "survival" } ],
     "proficiencies": [ "prof_fibers", "prof_fibers_rope", "prof_traps", "prof_trapsetting", "prof_knives_familiar" ],
+    "recipes": [ "beartrap" ],
```
Maleclypse, line 3461:
```suggestion
    "recipes": [ "beartrap" ]
```

data/json/professions.json
```diff
@@ -3458,6 +3458,7 @@
     "points": 3,
     "skills": [ { "level": 6, "name": "traps" }, { "level": 4, "name": "survival" } ],
     "proficiencies": [ "prof_fibers", "prof_fibers_rope", "prof_traps", "prof_trapsetting", "prof_knives_familiar" ],
+    "recipes": [ "beartrap" ]
```
Maleclypse, line 3461:
```suggestion
    "recipes": [ "beartrap" ],
```

Maleclypse:
Excess commits attempting to force a restart on the test queue.

PR #68821, opened 2023-10-22
[MoM] Add scaling to Metaphysics XP gain from powers, add penalty for power failure

Standing-Storm:
#### Summary
Mods "Add scaling to Metaphysics XP gain from powers, add penalty for power failure"

#### Purpose of change

I mentioned in #68555 that I wanted to scale the XP given based on power difficulty once that was an option, and now it is an option.

Also, I've long wanted to have the possibility of penalties for failed power usage--"An untrained telepath is a danger to themselves and everyone around them"--and now I can do that too.

#### Describe the solution

Change the formula for metaphysics XP to 100 XP * power Difficulty on a successful usage and 25 XP * power Difficulty on a failure.

Also, add the possibility of power overload on power failure! The formula is 

> "(10 * (min(0,( _difficulty - 1 - ( u_skill('metaphysics') / 2 ) ) * 15) ) + u_val('vitamin', 'name:vitamin_psionic_drain'))"

So a Difficulty 1 power at zero drain has a 1% chance of causing overload on a failed channel only. A Difficulty 5 power at moderate drain (vitamin 50) with metaphysics 4 has a 9.1% chance of causing overload, again only on a failed channel.

#### Testing

Everything works.

#### Additional context

For some reason, { "math": [ "_success", "==", "true" ] } didn't fire, so I used { "math": [ "_success", "!=", "false" ] } to detect a successful power usage.

### DIFF
--- a/data/mods/MindOverMatter/effectoncondition/eoc_misc.json
+++ b/data/mods/MindOverMatter/effectoncondition/eoc_misc.json
@@ -108,6 +108,284 @@
         }
       ]
     },
-    "effect": [ { "math": [ "u_skill_exp('metaphysics', 'raw')", "+=", "250" ] } ]
+    "effect": [ { "math": [ "u_skill_exp('metaphysics', 'raw')", "+=", "(100 * _difficulty)" ] } ]
+  },
+  {
+    "type": "effect_on_condition",
+    "id": "EOC_PSIONICS_BIOKINETIC_CHANNELING_FAILURE",
+    "eoc_type": "EVENT",
+    "required_event": "spellcasting_finish",
+    "condition": {
+      "and": [
+        { "u_has_trait": "BIOKINETIC" },
+        { "compare_string": [ "BIOKINETIC", { "context_val": "school" } ] },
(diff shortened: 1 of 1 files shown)


MNG-cataclysm:
If this sort of thing was applied to Magiclysm, does that now mean that we can use EOCs to have spells backfire in creative and unexpected ways, like randomly teleporting you, causing a massive uncontrolled cast that drains all your mana/psychic power, or randomly selecting a target (including the caster)?

Standing-Storm:
> If this sort of thing was applied to Magiclysm, does that now mean that we can use EOCs to have spells backfire in creative and unexpected ways, like randomly teleporting you, causing a massive uncontrolled cast that drains all your mana/psychic power, or randomly selecting a target (including the caster)?

You could absolutely do any and/or all of those things!

MNG-cataclysm:
Cause Bear here we come! Thanks for getting this set up, everything here looks really cool! I'll have to start experimenting with these new ways EOCs can work with spell failure.

PR #68825, opened 2023-10-23
Adds raised blindfolds and ability to lower them

detahramet:
#### Summary
Content "Blindfolds can now be raised"

#### Purpose of change

Blindfolds are mildly inconvenient, having to be removed entirely to disable their blindness. 

#### Describe the solution
Being able to raise them while still wearing them is a small QoL improvement to that effect.

N/A

#### Testing
1. Spawn Blindfolds
2. Wear Blindfolds
3. Raise Blindfolds
4. Be Unblinded

#### Additional context
N/A


PatrikLundell:
As long as blindfolds don't require an extra step to get them to actually work, which I don't think will be the case, I have nothing against this.

I will, however, continue to wear blindfolds when trying to sleep when I don't want light to disturb the sleep, and then drop it at the bed when waking up, as there's no point to haul it around when you have a fixed sleeping location.

PR #68851, opened 2023-10-24
Enforce properties no-duplicates restriction in JSON

### LINKED ISSUE #68270: ``properties`` field cannot be overwritten if it's copied by ``copy-from``
### Describe the bug

As per the title, if an item has a ``properties`` field that is carried over via ``copy-from`` from another item, you cannot manually define that field and overwrite what's being copied.

### Attach save file

Rather than a savefile, allow me to point you to #68257 where the tests are yelling at me about it

### Steps to reproduce

Simply look at the animal locker and add a ``properties`` field. Or, I mean, look at my changes in the PR linked above.

### Expected behavior

I expect to be able to overwrite the field, rather than having to avoid using ``copy-from`` on items with said field

### Screenshots

image

Not sure if this is applicable as this is a behavior in current tests of the repo? If you need me to boot up the game with the changes and get you version info then sure, I will, but I don't think that's necessary.

### Additional context

_No response_

ehughsbaird:
#### Summary
SUMMARY: None

#### Purpose of change
Fixes https://github.com/CleverRaven/Cataclysm-DDA/issues/68270
Properties had their own special loading function, but it was just loading a string->string map and making sure there were no duplicate entries.

`optional` will do the exact same thing, except the duplicate prevention will be enforced through the structure of the JSON, because it will be an object instead of an array.

#### Describe the solution
Delete the special loading function.
Use `optional`.
Convert JSON and docs to use the object format.

Clearing the properties before loading so overwrites are allowed, as other things with these special loading functions do.

#### Testing
Added a bit of code to the finalization stage in item_factory to check nothing was using this behavior to extend properties (nothing was).
Rebased https://github.com/CleverRaven/Cataclysm-DDA/issues/68270 onto this, modified the JSON, and found no more error.

#### Additional context
The doc entry for this is terrible. It should at least have a list of what can be here. I am too lazy to correct it right now.


LyleSY:
DinoMod bits look consistent with the rest, but I don't understand what's happening here well enough to comment on that, except that yes better documentation would be very helpful

PR #68859, opened 2023-10-24
Clarify "too frustrated to continue"

### LINKED ISSUE #68853: Character refuses to work due to frustrated, even though he's in a great mood.
### Describe the bug

My character refuses to work, even though he's in a really good mood and doesn't have any negative feelings.

### Attach save file

[帕洛阿尔托-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/13105746/-trimmed.tar.gz)

### Steps to reproduce

1 Try crafting any item.

I haven't encountered the same issue in other game progressions, it's only reproducible in this specific save file.

### Expected behavior

Tell me why I refuse to work.

- OS: Windows
    - OS Version: 10.0.22621.2428 (22H2)
- Game Version: 446e59e [64-bit]
- Graphics Version: Tiles
- Game Language: 系统语言 []
- Mods loaded: [
    浩劫余生 [dda],
    NPC无基本需求 [no_npc_food],
    NPC不受空间风暴影响 [personal_portal_storms],
    减缓真菌增殖 [no_fungal_growth],
    生化职业 [package_bionic_professions],
    Blaze 载具工业 [blazeindustries],
    大魔法 [magiclysm],
    神话武术 [MMA]
]

### Additional context

_No response_

oosyrag:
#### Summary
None

#### Purpose of change
Resolves #68853

#### Describe the solution
String change to clarify what causes crafting to stop from frustration by emphasizing crafting progress speed.

List out the factors that affect crafting speed dropping below 20%, which would be too much information, or leaving it as is, which is not enough/unclear information about the cause.

### DIFF
--- a/src/crafting.cpp
+++ b/src/crafting.cpp
@@ -326,7 +326,7 @@ float Character::crafting_speed_multiplier( const item &craft,
 
     // If we're working below 20% speed, just give up
     if( total_multi <= 0.2f ) {
-        add_msg_if_player( m_bad, _( "You are too frustrated to continue and just give up." ) );
+        add_msg_if_player( m_bad, _( "Your progress is so slow you give up in frustration." ) );
         return 0.0f;
     }
 
--- a/src/crafting.cpp
+++ b/src/crafting.cpp
@@ -326,7 +326,7 @@ float Character::crafting_speed_multiplier( const item &craft,
 
     // If we're working below 20% speed, just give up
     if( total_multi <= 0.2f ) {
-        add_msg_if_player( m_bad, _( "Your progress is so slow you give up in frustration." ) );
+        add_msg_if_player( m_bad, _( "Your progress is so slow that you give up in frustration." ) );
         return 0.0f;
     }
 


src/crafting.cpp
```diff
@@ -326,7 +326,7 @@ float Character::crafting_speed_multiplier( const item &craft,
 
     // If we're working below 20% speed, just give up
     if( total_multi <= 0.2f ) {
-        add_msg_if_player( m_bad, _( "You are too frustrated to continue and just give up." ) );
+        add_msg_if_player( m_bad, _( "Your progress is so slow you give up in frustration." ) );
```
Night-Pryanik, line 329:
```suggestion
        add_msg_if_player( m_bad, _( "Your progress is so slow so you give up in frustration." ) );
```

oosyrag: I think "Your progress is so slow **that** you give up in frustration." works better than "Your progress is so slow so you give up in frustration."?

Either that or "Your progress is slow so you give up in frustration." works too.

Night-Pryanik: Updated to "that" variant.

Issue #68831, opened 2023-10-23
Allow item descriptions to reference snippets, the way dialogue does.


MNG-cataclysm:
### Is your feature request related to a problem? Please describe.

I've been trying to create a few items here recently, and I noticed that I cannot call up snippets for use in their descriptions like I can in dialogue, such as using <given_name> or something similar. This makes item descriptions very limited in what they can say, especially for story-related items like survivor notes or lab notes.

### Solution you would like.

I would like to be able to reference snippets in item descriptions the way I can in dialogue, calling them as needed.

### Describe alternatives you have considered.

Manually creating override strings that feature all possible combinations of the snippets I'm trying to use, which is extremely tedious and takes up way too much space.

### Additional context

None at this moment.

ehughsbaird:
You can use snippet categories for this purpose https://github.com/CleverRaven/Cataclysm-DDA/blob/master/doc/JSON_INFO.md#random-descriptions

ehughsbaird:
For doing it in all item descriptions, it looks something like:
```diff
diff --git a/src/item.cpp b/src/item.cpp
index a59761605f..9261f00baa 100644
--- a/src/item.cpp
+++ b/src/item.cpp
@@ -2272,14 +2272,15 @@ void item::basic_info( std::vector<iteminfo> &info, const iteminfo_query *parts,
                 get_avatar().add_snippet( snip_id );
             }
         } else if( idescription != item_vars.end() ) {
-            info.emplace_back( "DESCRIPTION", idescription->second );
+            info.emplace_back( "DESCRIPTION", SNIPPET.expand( idescription->second ) );
         } else if( has_itype_variant() ) {
             // append the description instead of fully overwriting it
             if( itype_variant().append ) {
-                info.emplace_back( "DESCRIPTION", _( string_format( "%s  %s", type->description.translated(),
-                                                     itype_variant().alt_description.translated() ) ) );
+                info.emplace_back( "DESCRIPTION", _( string_format( "%s  %s",
+                                                     SNIPPET.expand( type->description.translated() ),
+                                                     SNIPPET.expand( itype_variant().alt_description.translated() ) ) ) );
             } else {
-                info.emplace_back( "DESCRIPTION", itype_variant().alt_description.translated() );
+                info.emplace_back( "DESCRIPTION", SNIPPET.expand( itype_variant().alt_description.translated() ) );
             }
         } else {
             if( has_flag( flag_MAGIC_FOCUS ) ) {
[...]

PR #68864, opened 2023-10-24
Dog names for dog tags

gtim:
#### Summary
Content "Dog names for dog tags"

#### Purpose of change

The dog tags found on dogs are a nice detail reminding the player that it was an actual dog, loved by an actual owner. Revealing the dog name to the player would add more of this emotional tinge.

#### Describe the solution

200 snippets for the dog_tag_dog item, each with a different dog name. 

If I understand the available tools for JSON items correctly, the repetitive `text` attributes were necessary. Perhaps a more desirable solution would let me define dog names elsewhere and include them with `<dog_name>`. I'll look into that if this solution is considered too bulky.

#### Testing

Tested spawning dog tags in-game and examining them. Made sure existing dog tags didn't cause issues.

### DIFF
--- a/data/json/items/generic.json
+++ b/data/json/items/generic.json
@@ -2567,7 +2567,861 @@
     "//": "dog tags are roughly 50mm x 28mm x 0.4mm, or 0.56 ml, but that's so small I needed to round up to 1 ml",
     "volume": "1 ml",
     "price": 0,
-    "price_postapoc": 0
+    "price_postapoc": 0,
+    "snippet_category": [
+      {
+        "id": "dog_tag_dog10",
+        "text": "An identification tag for a pet, engraved with their name and their owner's address and phone number.\n\nIt says: Abby."
+      },
+      {
+        "id": "dog_tag_dog20",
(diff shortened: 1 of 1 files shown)


GuardianDll:
What I'd personally prefer is resolve #68831 and just make a single item description as `An identification tag for a pet, engraved with their name and their owner's address and phone number.\n\nIt says: <dog_name>`

gtim:
> What I'd personally prefer is resolve #68831 and just make a single item description as `An identification tag for a pet, engraved with their name and their owner's address and phone number.\n\nIt says: <dog_name>`
> 
> Upd: sorry, didn't read the entire body of the PR, my bad

I agree. I also realised that the current PR would be unreasonable to translate, so maybe it should not be merged until `<dog_name>`-like variables are supported anyway?

MNG-cataclysm:
> I agree. I also realised that the current PR would be unreasonable to translate, so maybe it should not be merged until `<dog_name>`-like variables are supported anyway?

If this does get merged before #68831 is resolved, someone can go back through the code and refit it to support snippet queues.

PR #68870, opened 2023-10-24
Monsters can't stumble into debug invisible player

### LINKED ISSUE #68869: regression: debug invisibility no longer functions as debug invisibility
### Describe the bug

as of #65476 debug invisibility has been rendered effectively nonfunctional for the purposes of debug tasks that require the player to be, well, invisible, while in a place monsters can conceivably reach.

### Attach save file

n/a, trivially verified on any character

### Steps to reproduce

1. make a character
2. enable debug menu if necessary
3. add debug invisibility flag to player
4. go near hostile enemies
5. let one bump into you
6. find you are now revealed and will probably have more friends shortly

### Expected behavior

the entire purpose of the debug invisibility trait is to _not be seen_ so that you don't have to worry about the above scenario in a debugging session. in particular with the other 'don't notice me' traits (sound, scent) it should be pretty close to impossible for a creature to notice the debug'd character of its own volition.

### Screenshots

_No response_

- OS: Linux
    - OS Version: Distributor ID: Ubuntu; Description: Ubuntu 23.10; Release: 23.10; Codename: mantic; 
- Game Version: 0.G-4648-gbb72d88b08-dirty [64-bit]
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
]

### Additional context

i was testing on the commit immediately before cdda-experimental-2023-10-20-0258 (so that i could do some a/b testing with the visibility flood fill work kevin did), but it's known to be problematic going months back based on dev discord conversations, with confirmations as recently as yesterday.

Fris0uman:
#### Summary
None

#### Purpose of change

Fixes #68869
#### Describe the solution

Add a check for debug invisibility in `Creature::stumble_invis` to return false if the player is debug invisible
#### Testing

- Get debug invisible, no scent, and invincible
- Teleport in town
- Start yelling > Zombies come to investigate but never attack or stumble in to the player
- Wait for a few hours > Zombbies just ignore me as they should

### DIFF
--- a/src/creature.cpp
+++ b/src/creature.cpp
@@ -124,6 +124,7 @@ static const mon_flag_str_id mon_flag_WATER_CAMOUFLAGE( "WATER_CAMOUFLAGE" );
 
 static const species_id species_ROBOT( "ROBOT" );
 
+static const trait_id trait_DEBUG_CLOAK( "DEBUG_CLOAK" );
 static const trait_id trait_PYROMANIA( "PYROMANIA" );
 
 const std::map<std::string, creature_size> Creature::size_map = {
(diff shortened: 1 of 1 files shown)


esotericist:
tested locally on the same branch i was testing in #68869 and it does seem to work so far. activated a noisemaker, dropped it at my feet, now have neighbors that are not trying to be pushy.
image

was able to walk away from it after they hung around me a while just fine
image

esotericist (approved):
works on a working branch only 4 days older than current master.

PR #68871, opened 2023-10-24
Warn on invalid world name

### LINKED ISSUE #68793: Cannot have "?" in world name (silent error)
### Describe the bug

I wanted to make a new world called "Magic World?". So I named it that and clicked `f`inalize.
It sent me back to the main menu. The world wasn't there or in my save folder.

### Attach save file

N/A

### Steps to reproduce

1. Main menu
2. Create world
3. include "?" in world name
4. finalize
5. the world is nowhere to be found

### Expected behavior

Since you cannot have folders with "?" on Windows, Show the player error "Invalid world name" and continue in world generation.

### Screenshots

_No response_

- OS: Windows
- Game Version: d7c5d4c9ea [64-bit] - derived from master at f2f83783cc
- Graphics Version: Tiles
- Game Language: English [en]
- Mods loaded: [
    Dark Days Ahead [dda],
    Magiclysm [magiclysm]
]

### Additional context

_No response_

gtim:
#### Summary
Interface "Warn on invalid world name"

#### Purpose of change
On Windows, when the world name is an invalid folder name, the Create World dialog fails without feedback. This PR gives the user feedback in that scenario. Fixes #68793.

#### Describe the solution
* Completed the list of invalid characters with `:*?"<>|`, so the user is informed when these are in the world name. These characters *are* allowed on other systems, but since the function was already not system-agnostic (disallowing `\`) I figured that was fine. This could be considered a regression on non-Windows systems, or a portability feature.
* If saving the world fails anyway, the error message is displayed to the user instead of silently failing.
* Removed char_validity_check.cpp comment reference to player name -- seems it's only used for validating the world name.

* Validating the world name by trying to save the world instead of predicting whether a filename is correct.
* Not using the exact world name as save path.

#### Testing
Added `is_char_allowed` test cases. Tried creating worlds with valid and invalid names.

### DIFF
--- a/src/char_validity_check.cpp
+++ b/src/char_validity_check.cpp
@@ -4,9 +4,8 @@
 
 /**
  * Returns whether or not the given (ASCII) character is usable.
- * Called to check player name validity and world name validity.
- * Only printable symbols not reserved by the filesystem are
- * permitted.
+ * Called to check world name validity. Only printable symbols
+ * not reserved by the filesystem are permitted.
  * @param ch The char to check.
  * @return true if the char is allowed in a name, false if not.
  */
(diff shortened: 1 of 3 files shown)


Zireael07:
IIRC windows doesn't allow some specific strings as folder names, e.g. CON - does this handle those?

PR #68876, opened 2023-10-24
Lower requirements to use underbarrel gun attachments

### LINKED ISSUE #68753: Remove skill restriction on firing underslung grenade launcher

Currently, the player requires a Launchers skill of 2 in order to fire an underslung grenade launcher. Not to install it, reload it or maintain it, just to pull the trigger.

This seems like a pretty arbitrary and nonsensical restriction to me. While there's certainly a good case to be made that a completely unskilled user probably couldn't use one safely or well, to not be able to operate one at all strains belief, especially if the player otherwise has a very high marskmanship skill.

While it may result in player deaths if people are foolish about using it, so can surprising a player with an inexplicable inability to fire that grenade launcher they were saving for the direst of emergencies.

It's also worth noting that Launchers is a very difficult skill to train from zero as it is, and items that allow you to do so can often be harder to come across than grenade launchers and their ammunition. Learning to fire a grenade launcher by actually firing a grenade launcher seems like it should be a viable option, at least if you happen to have a large stockpile of grenades and a sufficiently distant target.

### Solution you would like.

The skill restriction on firing should be removed entirely. It might make sense for the same skill restriction to be applied to *installing* the module on a firearm, and/or when loading it.

Making it clear when loading or installing the GLM that your character doesn't presently understand how to fire it, and explaining why.

### Additional context

Here is an example of such a launcher. It has it's own trigger and a highly-visible labelled safety switch, and figuring out how to get it to fire is not rocket science.
image

oosyrag:
#### Summary
Balance "Lower requirements to use underbarrel gun attachments"

#### Purpose of change
Fixes #68753

#### Describe the solution

Made skill requirements match, so "weapon" requirements dropped to 1. This should have minimal impact as the difference between 1 and 2 skill for the modified gun is insignificant. The result would be that if you can install the mod, you can use it too.

Increase the mod's specific requirement to 2, but given the difficulty of skilling up launchers I opted for the other way around. I also don't want to get torched by redditors.

Also implement a system where the requirements to attach the mod and fire the mod are independent of each other, but this was the quicker fix.

#### Testing
~~TBD. Hoping having two of the "same" requirement defined doesn't break anything.~~

Tested with an M4 and M320 glm. Showed requirement to attach as rifles 1, launchers 1, debug added required skills, was able to attach, then fire the grenade (extremely inaccurately, which is great).

#### Additional context

There is nothing that differentiates skill requirements between being able to use an attachment to mount it on a gun and being able to use it to fire it as an attachment on your gun.

The specific requirements for the underslung grenade launchers, shotguns, and bayonet were [ "weapon", 2 ], [ "launcher/shotgun/melee", 1 ]. This resulted in the ability to attach for example "launchers" at skill 1, but then when using it, "weapon" also refers to launchers which resulted in the requirements ending up now being [weapon (launchers)", 2], ["launcher", 1]. This effectively means that after getting launchers to 1 to install the gunmod, you now need another level before you can actually use it. 

Given our ability to use firearms at 0 skill, I'm assuming this is a quirk of the code rather than an intended gameplay challenge to overcome.

### DIFF
--- a/data/json/items/gunmod/underbarrel.json
+++ b/data/json/items/gunmod/underbarrel.json
@@ -425,7 +425,7 @@
       "blackpowder_tolerance": 60,
       "clip_size": 1
     },
-    "min_skills": [ [ "weapon", 2 ], [ "launcher", 1 ] ],
+    "min_skills": [ [ "weapon", 1 ], [ "launcher", 1 ] ],
     "flags": [ "NEVER_JAMS", "RELOAD_EJECT" ],
     "pocket_data": [ { "pocket_type": "MAGAZINE", "ammo_restriction": { "40x46mm": 1 } } ],
     "faults": [ "fault_gun_blackpowder", "fault_gun_dirt" ]
@@ -462,7 +462,7 @@
       "blackpowder_tolerance": 60,
       "clip_size": 1
     },
-    "min_skills": [ [ "weapon", 2 ], [ "launcher", 1 ] ],
+    "min_skills": [ [ "weapon", 1 ], [ "launcher", 1 ] ],
     "flags": [ "NEVER_JAMS", "RELOAD_EJECT" ],
     "pocket_data": [ { "pocket_type": "MAGAZINE", "ammo_restriction": { "40x46mm": 1 } } ],
     "faults": [ "fault_gun_blackpowder", "fault_gun_dirt" ]
@@ -527,7 +527,7 @@
     "location": "underbarrel",
     "mod_targets": [ "rifle", "crossbow" ],
     "gun_data": { "ammo": "shot", "skill": "shotgun", "dispersion": 330, "durability": 10, "clip_size": 4 },
-    "min_skills": [ [ "weapon", 2 ], [ "shotgun", 1 ] ],
+    "min_skills": [ [ "weapon", 1 ], [ "shotgun", 1 ] ],
     "flags": [ "RELOAD_ONE" ],
     "pocket_data": [ { "pocket_type": "MAGAZINE", "ammo_restriction": { "shot": 4 } } ]
   },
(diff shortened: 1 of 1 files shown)


Solusphere:
While this is definitely an improvement, it doesn't fully address the problem. There is still an arbitrary and nonsensical skill requirement involved should the player obtain one of these items pre-attached. Personally I don't think there should be any hard skill requirements for installing the attachments either, though that falls a little outside the scope of my initial request as I didn't realize they were so interlinked. This does seem like it's the best that can be done within JSON, so thankyou for that.

oosyrag:
Well... removing the requirement completely is an option too. There's a bit of subjectivity regarding what "0 skill" means. To me it means you haven't heard of or seen or had any exposure or experience to the thing at all, so you wouldn't know which way to stick the potato into the tube. Lowering it to 1 is a compromise, but I'm not against getting rid of it completely.

Solusphere:
It is very subjective, yes, and the "minimum baseline" is probably different for different skills, but personally I lean in the other direction. I've never so much as held real gun in my life, but I'm pretty confident I could figure out how to attach a grenade launcher or other rifle attachment through simple trial and error given enough time, though it may take a while and be badly collimated etc. Should that constitute skill 1? Personally, I don't think so. I think it's reasonable to assume that the player character has at least a pop culture/video game level of understanding of most basic ingame tasks, and can generally figure out the rest through simple problem solving unless it's something that truly would require specific knowledge, such as electronics or mechanics.

OtpalTS:
What about making the underbarrel restrictions not be shotgun or launcher skill but marksmanship, since that is ( if i understand it correctly) general firearm knowledge. But making it require just one in used skill seems fine as well.
 As for installing, i dont think making it 0 is a good idea since if you dont know what youre doing at least a bit you risk attaching it insecurely and it coming loose when fired. I dont think we can have guidebooks that would allow bypassing installation skill requirements which could be fun.
Also I think the skill checks use practical knowledge where in this specific case maybe it could use theoretical, which could be the solution to the using a guide to install it.
Overall though this change seems fine, I'm just giving my thoughts on this subject

PR #68880, opened 2023-10-24
Allow scanning several books into ereader at once

### LINKED ISSUE #67886: Ability to scan multiple amounts of books in one operation on smartphone.

Why do you need all these clicks for scanning books? Just make them all selectable in one operation.

### Solution you would like.

Well just make saving screen like dropping screen.  Also need to put energy and time consumption at this screen

_No response_

### Additional context

_No response_

inogenous:
#### Summary
Interface "Allow scanning several books into ereader at once"

#### Purpose of change

* Adds the ability to select multiple books to scan into e-book reader, and not just one book as before.
* Fixes #67886 .

#### Describe the solution

* Updates `ebooksave_activity_actor` so that it stores a `std::vector<item_location>` of books that should be scanned, instead of just holding a reference to one book `item_location`.
* Reuses existing `inventory_multiselector` for selecting books.

* This solution pops the last elements from `std::vector` when they are fully scanned. Another way would be to just keep an index on "which book are we currently scanning?", and not modifying the list as the activity runs.

#### Testing

scan-multiple-books

* Tried reading books from inventory - works ✔
* Placed some books on the pavement - they are also shown when selecting which books to scan ✔
* Set fire to the pile of books on the pavement, so that they will disappear "soon". Started scanning all books - works as expected that only some of the books were scanned (the first ones that hadn't burned up yet) ✔
* Cancelling the activity with `5` while scanning - works ✔

### DIFF
--- a/data/json/item_actions.json
+++ b/data/json/item_actions.json
@@ -446,7 +446,7 @@
   {
     "type": "item_action",
     "id": "EBOOKSAVE",
-    "name": { "str": "Store a book" }
+    "name": { "str": "Store books" }
   },
   {
     "type": "item_action",
--- a/data/json/player_activities.json
+++ b/data/json/player_activities.json
@@ -30,7 +30,7 @@
     "type": "activity_type",
     "id": "ACT_EBOOKSAVE",
     "activity_level": "NO_EXERCISE",
-    "verb": "taking pictures of the book",
+    "verb": "taking pictures of books",
     "rooted": true,
     "based_on": "speed",
     "auto_needs": true
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -2655,21 +2655,61 @@ std::unique_ptr<activity_actor> lockpick_activity_actor::deserialize( JsonValue 
     return actor.clone();
 }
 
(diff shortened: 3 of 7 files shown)


GuardianDll:
how does it works with books-variants? they are pickable, but are not scanned since it's the same book under the hood?

inogenous:
> how does it works with books-variants? they are pickable, but are not scanned since it's the same book under the hood?

The ereader stores the typeids, and all variants will have the same typeid. When presenting the list of books to select, it will not show books with typeids that are already stored. So if
* the ereader already contains "Zen and the art of archery" 
* and the player is carrying "The Theory And Practice Of Archery"
* then "The Theory And Practice Of Archery" will not be pickable.

GuardianDll:
i mean the case, where you have nothing in your e-reader, approach a pile of books, and this pile has both `Zen and the art of archery` and `The Theory And Practice Of Archery`, that are both not scanned yet

inogenous:
Oh yes, good point! At the moment, they will both be pickable when choosing what to scan, and they will both be stored on the device. Later, when selecting "Read stored books", both will be available.

Cupidatis:
Nice!

PR #68886, opened 2023-10-24
More stockings variants

gettingusedto:
#### Summary
Content "Adds more of the pride stockings variants that i forgot"

#### Purpose of change

Adds more of the pride stockings i forgot to add because i got distracted somewhere.

#### Describe the solution

adds genderqueer, genderfluid, agender, non-binary, intersex, and polysexual stockings variants.

let someone else add this.
#### Testing

Screenshot_2023-10-25_04-35-07_1

### DIFF
--- a/data/json/items/armor/boots.json
+++ b/data/json/items/armor/boots.json
@@ -2902,6 +2902,54 @@
         "weight": 75,
         "color": "blue",
         "append": true
+      },
+            {
+        "id": "stockings_pride_genderqueer",
+        "name": { "str": "pair of genderqueer flag stockings", "str_pl": "pairs of genderqueer flag stockings" },
+        "description": "This one has patterns akin to the genderqueer flag.",
+        "weight": 55,
+        "color": "magenta",
+        "append": true
+      },
+      {
+        "id": "stockings_pride_genderfluid",
+        "name": { "str": "pair of genderfluid flag stockings", "str_pl": "pairs of genderfluid flag stockings" },
+        "description": "This one has patterns akin to the genderfluid flag.",
+        "weight": 45,
+        "color": "magenta",
+        "append": true
+      },
+      {
+        "id": "stockings_pride_agender",
+        "name": { "str": "pair of agender flag stockings", "str_pl": "pairs of agender flag stockings" },
+        "description": "This one has patterns akin to the agender flag.",
+        "weight": 45,
+        "color": "dark_gray",
+        "append": true
+      },
+      {
+        "id": "stockings_pride_nonbinary",
+        "name": { "str": "pair of non-binary flag stockings", "str_pl": "pairs of non-binary flag stockings" },
+        "description": "This one has patterns akin to the non-binary flag.",
+        "weight": 75,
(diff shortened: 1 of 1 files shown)


Maleclypse:
> I'm not fond of now seeing zombie children dropping all kinds of sexual orientation stuff. Children shouldn't be sexualized.

I have good news for you then. Pride variant items aren’t sexual. They are supportive of human rights for people. Intersex children aren’t sexualized from birth by their DNA. Children supporting human rights for their friends, families or even strangers aren’t sexualized. Wearing Merch that speaks to support of a minority isn’t a sexual behavior.

PatrikLundell:
I disagree. Either parents would force/fool their children into wearing clothing expressing political opinions children should not know the basis for, or the children would have to be exposed to the concepts of sexual behavior well before they should have a notion of those concepts.
I don't think parents should force non sexual political opinions on children either.

AmarinReyny:
> I disagree. Either parents would force/fool their children into wearing clothing expressing political opinions children should not know the basis for, or the children would have to be exposed to the concepts of sexual behavior well before they should have a notion of those concepts. I don't think parents should force non sexual political opinions on children either.

The existence of queer people is neither a "political opinion," nor inherently sexual. Furthermore, queer children do exist, and do have the capacity to have some level of awareness of their own queerness, even if they don't have the context or knowledge to understand it. As a queer person who was once a queer child myself, I can personally attest to this - after all, I instinctively felt like the body's genitalia didn't belong there at a very early age, before I was even taught that not everyone had the same genitalia - but there's also plenty of research showing that queer kids do in fact exist, if you want more than just anecdotes.

Helping kids understand what they're going through and supporting them in it is not forcing anything on them. If anything, keeping kids in the dark and acting like queerness is something inherently inappropriate is going to lead to more queer kids growing up and suffering the way I and many others did, with overwhelming self-hatred, forced conformity driven by fear, severe identity issues, and suicidality, along with numerous other issues that will severely impact them for the rest of their lives. People like me want kids to be taught about this stuff because _we don't want the queer kids of today to suffer the way we did_.

If you want to force children to suffer because you think queerness is an inherently sexual political opinion, then you're the one who is forcing things onto children, and you're part of the problem that people like me are trying to fix.

PatrikLundell:
My comment is being turned into a different political discussion I don't think this is a suitable place for. Most of the letters in this set of clothing articles are about sexual orientation, and none of them the mental well being of children (and what to do about that).

AmarinReyny:
> My comment is being turned into a different political discussion I don't think this is a suitable place for. Most of the letters in this set of clothing articles are about sexual orientation, and none of them the mental well being of children (and what to do about that).

You're right, this is not a suitable place for a discussion about why queer kids shouldn't be supported or accepted, because there is no suitable place for that kind of oppressive misinformation and propaganda against a vulnerable minority.

Maleclypse:
https://www.adl.org/resources/blog/what-grooming-truth-behind-dangerous-bigoted-lie-targeting-lgbtq-community

I hope that if you are here you choose to read this article.

PR #68895, opened 2023-10-25
Tamable Wildlife: Lady Bugs and Mole Crickets

Termineitor244:
#### Summary
Mods "Makes the lady bug and mole cricket tamable in the Tamable Wildlife mod"

#### Purpose of change
I noticed these 2 monsters were untamable even though they had baby forms that could be tamed with enough love! (And sugar, looots of sugar).

#### Describe the solution
Adds both monsters as larvae/nymphs to the mod as tamable with bug food, the adult versions are mountable too! Now even troglobites can tame and mount their little (Or not so little...) pets in the subway! You only need to catch one first, they are pretty fast.

To add some other monsters of the file that could be added, but I'm sleepy...

#### Testing
It loads! The monsters can be tamed! And played with!

#### Additional context
imagen
imagen
imagen

### DIFF
--- a/.github/reviewers.yml
+++ b/.github/reviewers.yml
@@ -34,6 +34,8 @@ files:
     - ZhilkinSerg
   'data/mods/No_Hope/**':
     - Night-Pryanik
+  'data/mods/Tamable_Wildlife/**':
+    - Termineitor244
 
   '**/.clang-tidy':
     - jbytheway
--- a/data/mods/Tamable_Wildlife/insect_spider.json
+++ b/data/mods/Tamable_Wildlife/insect_spider.json
@@ -248,13 +248,6 @@
     "extend": { "flags": [ "CANPLAY" ] }
   },
   {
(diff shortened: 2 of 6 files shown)


Karol1223:
I don't remember how I did the Mole Crickets but Ladybug larvae are aggressive so they should be given the same treatment as mantis and the such.

Termineitor244:
> I don't remember how I did the Mole Crickets but Ladybug larvae are aggressive so they should be given the same treatment as mantis and the such.

Mole Crickets were not present in the mod, they are pretty timid and run away from the player if given the chance, so I think its ok to leave them as they are. And I will give the ladybugs the same treatment as the mantis, just a moment...

Karol1223:
> Mole Crickets were not present in the mod

Yeah, they definitely weren't, I was talking about the baby forms in the mainline lol. I added the missing ones a while ago and I didn't recall if mole crickets had angry babies or not, but if they flee they should be fine as is.

Issue #68892, opened 2023-10-25
Segfault on Bandaging


JacobHeilman:
### Describe the bug

When attempting to bandage wounds, the game crashes.

### Attach save file

[Drip Rock-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/13160730/Drip.Rock-trimmed.tar.gz)

### Steps to reproduce

1. Load game.
2. Try to bandage something
3. Game crashes when it finishes the action (I think)
Interestingly, when I just tested it, applying antiseptic first prevented the crash, I tried that a few times and then bandaged my head without antiseptic and everything worked fine

### Expected behavior

I expected it to not crash, especially because I've successfully applied bandages in this specific run before

### Versions and configuration

-OS: Windows 11 Home
-Game Version: ee51041 (64-bit)
-Graphics Version: Tiles
-Ingame Language: English
-Mods loaded: dda, no_npc_food, personal_portal_storms, no_fungal_growth, PersTweaks (personal mod, currently just some blacklists)

### Additional context

[Drip Rock-trimmed.tar.gz](https://github.com/CleverRaven/Cataclysm-DDA/files/13160790/Drip.Rock-trimmed.tar.gz)
[debug.log](https://github.com/CleverRaven/Cataclysm-DDA/files/13160791/debug.log)
[crash.log](https://github.com/CleverRaven/Cataclysm-DDA/files/13160792/crash.log)

PR #68907, opened 2023-10-25
Prevent first-aid segfault by instead bailing out of activity

gtim:
#### Summary
Bugfixes "Prevent crash on bandaging"

#### Purpose of change

#68892 had a segfault when bandaging. This PR does not solve the underlying issue, but prevents crash by canceling the bandaging.

#### Describe the solution

The `item_location` of a bandage in an aluminum can was lost somewhere between the activity constructor and `firstaid_activity_actor::finish`. This PR ensures that the item pointer is valid before dereferencing. Displays a debug message to let the user know something went wrong.

Checking for nullptr in `item_location::operator->`, but I don't know the codebase enough to say if that's a reasonable performance hit in a lightweight handle.

#### Testing

Tested repro steps in the linked issue.

### DIFF
--- a/src/activity_actor.cpp
+++ b/src/activity_actor.cpp
@@ -6154,7 +6154,7 @@ void firstaid_activity_actor::finish( player_activity &act, Character &who )
     static const std::string iuse_name_string( "heal" );
 
     item_location it = act.targets.front();
-    if( it.get_item() == nullptr ) {
+    if( !it ) {
         debugmsg( "Lost tool location used for healing" );
         act.set_to_null();
         return;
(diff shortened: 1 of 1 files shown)


src/activity_actor.cpp
```diff
@@ -6154,6 +6154,11 @@ void firstaid_activity_actor::finish( player_activity &act, Character &who )
     static const std::string iuse_name_string( "heal" );
 
     item_location it = act.targets.front();
+    if( it.get_item() == nullptr ) {
```
andrei8l, line 6157: Use the `item_location`'s bool operator instead

```suggestion
    if( it ) {
```
and merge this check into the `if` below.

There's not much point to this though. An activity that fails with a debug message for no discernible reason is not much better than a crash from the player's perspective.

gtim: Thank you for reviewing! How do you suggest I merge with the other `if`? I didn't see a clear way without calling `get_usable_item` twice.

gtim:
Thanks for your comments and for letting me know that demoting from a crash is not desirable!

PR #68910, opened 2023-10-25
Add "HARVESTED" flag to empty apiaries

oosyrag:
#### Summary
Bugfixes "Empty apiaries now properly refill in autumn"

#### Purpose of change
image

~~(UNCONFIRMED)~~

#### Describe the solution

Added the "HARVESTED" flag to empty apiaries, which I suspect the lack of prevented transformation back into apiaries.

#### Testing
~~TBD. Is there a way to skip seasons worth of time in debug?~~

Confirmed -

Before change - New game, debug spawn empty apiary, move out of reality bubble, debug set season to autumn, go back to apiary and it is still empty.

After change - New game, debug spawn empty apiary, move out of reality bubble, debug set season to autumn, go back to apiary and it is ready for harvest.

#### Additional context

Going by the docs for transforms_into, seems like the HARVESTED flag would be necessary to transform. I have not confirmed this in the code.

> #### `transforms_into`
> 
> (Optional) Used for various transformation of the terrain. If defined, it must be a valid terrain id. Used for example:
> 
> - When harvesting fruits (to change into the harvested form of the terrain).
> - In combination with the `HARVESTED` flag and `harvest_by_season` to change the harvested terrain back into a terrain with fruits.

### DIFF
--- a/data/json/furniture_and_terrain/terrain-flora.json
+++ b/data/json/furniture_and_terrain/terrain-flora.json
@@ -1912,7 +1912,7 @@
     "color": [ "light_gray" ],
     "move_cost": 1,
     "coverage": 50,
-    "flags": [ "FLAMMABLE_ASH", "NOITEM" ],
+    "flags": [ "FLAMMABLE_ASH", "NOITEM", "HARVESTED" ],
     "transforms_into": "t_apiary",
     "emissions": [ "emit_bees" ],
     "bash": {


Maleclypse:
Debug set time for testing. Remember to be out of the reality bubble with what you are testing when you change time

oosyrag:
It is done, confirmed it wasn't working and adding the tag fixed it.

PR #68918, opened 2023-10-26
Adjust bike wheel installation/removal times

oosyrag:
#### Summary
None

#### Purpose of change
Bike wheels don't take 15 minutes to change.

#### Describe the solution
Adjusted bike wheels and unicycles down to 5 minutes. Which is still more than I think it takes, but I believe mechanics skill shortens this.

Tricycle wheels up to 15 minutes though, since there's three of them to do at the once.


oosyrag:
Hmm... test failure in basic build... Unrelated I think?

PR #68934, opened 2023-10-27
Add several comment commands for issues and PRs

Bobtron:
#### Summary
Infrastructure "Add more comment-commands which apply labels to issues and PRs"

#### Purpose of change
Expand list of labels that can be added with comment commands

Add more comment commands to this pr, or remove some added in this pr

#### Testing
Enabled github-actions on my fork of cdda, and tested these commands on issues and prs there

### DIFF
--- a/.github/comment-commands.yml
+++ b/.github/comment-commands.yml
@@ -46,3 +46,15 @@ users:
 keywords:
   - name: confirm-bug
     value: '\/(C|c)onfirm.*'
+
+  - name: close-duplicate
+    value: '\/(D|d)uplicate.*'
+
+  - name: close-invalid
+    value: '\/(I|i)nvalid.*'
+
+  - name: good-first-issue
+    value: '\/(G|g)ood[\s-]*(F|f)irst[\s-]*(I|i)ssue.*'
+
+  - name: help-wanted
+    value: '\/(H|h)elp[\s-]*(W|w)anted.*'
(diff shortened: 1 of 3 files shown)


casswedson:
I didn't want to implement actually closing issues, since stalebot will get to all issues eventually - and in this case _I_ think the limitations are ok, I wouldn't want to hand the permission to close issues to anyone; just slapping a label like "need more info" "needs testing" would be useful

the real tools for helping the organization should be kept behind the member status; Kevin and the other members, take care of filtering the candidates

Bobtron:
> I didn't want to implement actually closing issues, since stalebot will get to all issues eventually - and in this case _I_ think the limitations are ok, I wouldn't want to hand the permission to close issues to anyone; just slapping a label like "need more info" "needs testing" would be useful
> 
> the real tools for helping the organization should be kept behind the member status; Kevin and the other members, take care of filtering the candidates

Good point on the closing issues - I think I'll remove the closing issues command for duplicate + invalid, the issue owner, stalebot, or mods can decide to close it. I checked the [labels](https://github.com/CleverRaven/Cataclysm-DDA/issues/labels) list, doesn't look like theres any 'needs testing' label, otherwise we could add that to the commands, not sure how to make it accessable to all users though, not just the vetted ones.

Karol1223:
I mean, "needs testing" is basically the "unconfirmed" label that gets slapped onto all bug reports at the start anyway, no? Unless we're talking "I could not reproduce it so could someone else try testing this out"

Bobtron:
> I mean, "needs testing" is basically the "unconfirmed" label that gets slapped onto all bug reports at the start anyway, no? Unless we're talking "I could not reproduce it so could someone else try testing this out"

Yeah, was thinking the latter for this one, could also be for PRs which are large enough to need manual testing to make sure nothings broken

Maleclypse:
I'm confused it looks like the closing commands are still in the top file?

casswedson:
closing issues was removed, but we can still label them with invalid, duplicate etc

PR #68944, opened 2023-10-27
Reword actions in repair gui

inogenous:
#### Summary
Interface "Reword actions in repair gui"

#### Purpose of change

* Updates the wording in the repair ui to make it clear which actions are "repair", and which actions include "refit".
* Change wording of "repeat once" to make it clear that it in fact does not repeat but is instead only one attempt.

#### Describe the solution

* `repeat_type::ONCE`
    * Before: `Repeat once`
    * Suggested: `Attempt to (refit or) repair once`
    * The intention here is both to make it clear whether this is a "repair" and/or "refit" action; and that it is only one attempt, not any repeating action.
* `repeat_type::EVENT`
    * Before: `Repeat until success/..`
    * Suggested: `(Refit or) Repair until success/..`
    * The intention is here is to make it clear whether this is a "repair" and/or "refit" action

* We could question whether we should keep the "attempt once" options. Sure, there's a use-case when you want to have control over how many threads are being used up, but for the most part, the player probably only wants to fix the item?
* We could question whether the wording on option 3 should include "level up" or not. The player probably already expects that since other actions (such as training recipes) usually prompt for that.
* Another way we could reword action 2 would be `Refit and fully repair`
* Another way we could reword action 3 would be `Refit or repair until condition changes`

#### Testing

* Used a sewing kit - works ✔
* Used a sewing kit on item that can also refitted - works ✔
* Used an arc welder - works ✔

repair-gui-reword2

### DIFF
--- a/src/activity_handlers.cpp
+++ b/src/activity_handlers.cpp
@@ -2195,12 +2195,12 @@ void activity_handlers::start_engines_finish( player_activity *act, Character *y
 enum class repeat_type : int {
     // INIT should be zero. In some scenarios (vehicle welder), activity value default to zero.
     INIT = 0,       // Haven't found repeat value yet.
-    ONCE = 1,       // Repeat just once
+    ONCE = 1,       // Attempt repair (and refit if possible) just once
     // value 2 obsolete - previously used for reinforcement
-    FULL = 3,       // Repeat until damage==0
-    EVENT = 4,      // Repeat until something interesting happens
-    REFIT_ONCE = 5, // Try refitting once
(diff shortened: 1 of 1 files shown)


inogenous:
Hm, seems like the top three actions do in fact include "refit", but the bottom two ones are "refit only, don't repair". Will keep this MR as draft until it is updated to reflect that. Guess this shows even more how the previous wording on these actions could benefit from being made clearer.

inogenous:
Update: Have updated the wording now so that it hopefully is even more clear as to whether the actions include "refit" and/or "repair".

PR #68959, opened 2023-10-28
Give Laptop `ALLOWS_REMOTE_USE` flag

gettingusedto:
#### Summary
None

#### Purpose of change

It bothers me that i had to wield it whenever i wanted to do something like reading the books i'e transferred. Unless this is a common thing people in the States do, i'm changing it.

#### Describe the solution

Give the laptop `ALLOWS_REMOTE_USE` flag

Not doing so and assume most people in the States wields their laptop whenever they used it.

#### Testing

Applying it to my build, now i can activate stuff without needing to pick up the laptop.

### DIFF
--- a/data/json/items/tool/electronics.json
+++ b/data/json/items/tool/electronics.json
@@ -374,7 +374,7 @@
       },
       { "type": "link_up", "cable_length": 5, "charge_rate": "140 W" }
     ],
-    "flags": [ "WATCH", "WATER_BREAK", "ELECTRONIC" ],
+    "flags": [ "WATCH", "WATER_BREAK", "ELECTRONIC", "ALLOWS_REMOTE_USE" ],
     "pocket_data": [
       {
         "pocket_type": "MAGAZINE_WELL",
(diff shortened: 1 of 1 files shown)


PatrikLundell:
While I would expect 'muricans to use laptops in their laps or placed on a suitable (or, at times, unsuitable) surface in front of them, I don't expect them to be able to operate them while still in their backpack or on a storage shelf across the room.
While you could hand wave laptop usage while in the backpack by invoking player convenience (which should be associated with access and putting away costs, while not having either be an automatic death sentence when something suddenly shows up due to being impossible to interrupt), I can't really see usage from further away than the neighboring tile and while in the open (not while still in a cardboard box or a drawer, for instance).

Zireael07:
One tile over sounds reasonable - more than that sounds like a stretch

esotericist:
allow remote use is just preventing the player from needing to hold the item. the activate command itself won't look for things further than a tile away, so it's still limited to adjacency even with the flag set.

Issue #68760, opened 2023-10-19
Audit monster dissection proficiency allocation


PatrikLundell:
### Is your feature request related to a problem? Please describe.

The proficiencies gained from corpse dissection are allocated inconsistently and do not always contribute to prerequisite proficiencies, while leaving you in the dark as to what these are. 

- I look at the partially known "Invertebrate Biology" proficiency and see "Basic familiarity with various types of insects, arachnids, and worms" and start to dissect a newly killed large worm, only to find that it doesn't have any associated proficiencies at all.
- I try to dissect a Panicked Person, and find it to likewise be empty of proficiencies (ferals seem to provide Physiology, and I'm not at all against a morale hit on human dissection).
- I try to dissect an acidic zombie, and find that while I have mastered two of the three proficiencies, I lack the unspecified prerequisites for the third one.
- I try to dissect some flying insect (some kind of wasp or a fly, I think) and find that it doesn't contribute to Principes of Flight, while I think some other flying insect corpse did.

### Solution you would like.

Dissection of creatures falling into a group described by proficiencies would contribute to those proficies or, when prerequisites are missing, the next missing prerequisite proficiencie(s).

Edit: After finally finding (through json files) that biochemistry has chemistry as its prerequisite chain, I realize that makes sense for the proficiency itself, but it then doesn't make sense for it to be gained through dissection, but rather should be gained through study of dissection products (acids, what I've seen so far) via research/training recipes.

### Describe alternatives you have considered.

Fumble around in the dark and accept it's a proficiency lottery.

Edit:
Retain biochemistry as a dissection gained proficiency where you analyze its properties while dissecting, rather than in a chemical examination environment, but then somehow indicate you need to study chemistry to gain the prerequisites, not dissecting lots of corpses.

### Additional context

This is not a new issue...

Maleclypse:
Anyone up for making a python script to build s list of monsters that lack proficiencies?

PatrikLundell:
Manually going through stuff (work in progress, noting as I go):
- Birds:  look OK, but ought to also have Physiology based on the description of that proficiency (they are animals, and the proficiency doesn't say mammals).
- Civilians: Look mostly OK, and a spawned Panicked Person did provide the proficiencies. Civilian police ought to have "prof_wp_syn_armored" as well, as it gets it when zombiefied.
[...]

PR #68963, opened 2023-10-28
Moved biochemistry from dissection to research/practice

PatrikLundell:
#### Summary

None

#### Purpose of change

Move the Biochemistry proficiency from being gained through dissection, but only after gaining a lot of chemistry proficiencies that aren't indicated anywhere (except digging through the JSON files), to being gained through "research", which is implemented as practice recipes.

#### Describe the solution

Implemented two "practice" recipes for the proficiency, one for the chemistry set, and one for the basic chemistry set (as that also requires an external heat source).
Then removed all the biochemistry proficiencies from monsters (mods not looked at).

Related to #68760 as being one of the side discoveries mentioned in the research in its comments.

The research process is just a guess, so anything from skill levels to proficiency multipliers and other parts could be set differently. The duration for the heat source for the basic chemistry set is based on the guess that the number is in minutes and that you want heat for the whole duration.

Add the whole chain of chemistry proficiencies required to gain the top one required to cut down on player guesswork.

Name the research recipe file differently. The terminology is a mess with applied science, chemistry, and science seemingly meaning the same thing, and I don't know which term is supposed to the the one moving forward.
 
If there is a way to merge the two recipes into one, where the heat source is required only with the basic chemistry set I'd like to know it.

#### Testing

- Implemented the recipes.
- Checked that the recipes weren't visible due to a lack of skill proficiency.
- Increased the applied science skill to 5 to gain the recipes.
- Checked that the recipes appeared and that their conditions gradually were approved as items were spawned.
- Spawned an acidic zombie, debug killed it, drained it of blood into a bucket, and verified I still couldn't do the research even though the blood requirement was fulfilled as the chemistry proficiency was missing.
- Debug acquired the chemistry proficiency (without bothering with its prerequisites...).
- Verified that I could research the proficiency and that I gained about 13% after one session and 25% after a second one, which seems to track towards the 8 hours set for the proficiency.
- Started from scratch but with the basic set and a fire (which kept going out, for some reason), but I got through a full session eventually.

Somewhere in the testing I spawned an acidic ant and found I'd missed the insect blood type. Tried to spawn a dark wyrm (to go for the underground blood), but it failed to appear (possibly because of my character standing on a road?). I don't know what plants produce acid, so I haven't spawned any of those.

When modifying the creatures to remove biochemistry from them I spawned and debug killed sample critters to verify that they still had the rest of the dissection proficiencies they should have. In particular I made sure removal from a primary creature worked, removal of a replacement "families" entry from a copy-from creature worked (the only thing added was the biochemistry), and removal of a family extension worked (again, the extension was the biochemistry).

#### Additional context

I've waited for some feedback on #68760 without any luck, as I really don't want to do a lot of work only for the powers in charge to wake up and say they don't like after the work has been done, especially since there are a number of issues that could be discussed (like this PR) and also when the natural armor weakpoint/proficiency should be applied (and when the proficiency should be there without a weakpoint, or vice versa).


Vaskritaya:
> If there is a way to merge the two recipes into one, where the heat source is required only with the basic chemistry set I'd like to know it.

Not sure if this would work in practice, but maybe you could use requirements similar to
 - basic chemistry set OR chemistry set (0 charges)
 - surface heat (60 charges) OR chemistry set (50 charges)

as then having just the chem set would fulfill both, but with the basic set youd need a heater? Would be pretty janky even if it did work, and the game might not actually let you use a tool in multiple spots like that, but it's just what came to mind.

PatrikLundell:
Interesting idea. I tested it, and it didn't work out great, as the heat source or chemistry set (50) expands into ...chemistry set (2100 charges) or chemistry set (50 charges)...

It was a nice idea, though.

Maleclypse:
My apologies about the lack of feedback. I had added the P5 tag because this work is wanted. Do you have specific questions you want answered, when is looked through I didn’t see any changes I disagreed with. I’ll review this specific PR changes soon.

PatrikLundell:
Thanks @Maleclypse:
- This PR handles one of the issues, so the review of it should take care of that part.
- The bulk of the work would be the addition to Physiology in a lot of places. I was uncertain if that would be accepted or not, but you've clarified that.
- The issue I have the most problem with is how to judge when the Natural Armor proficiency should be present on a critter.
- Related to the Natural Armor issue is when a critter should have corresponding weakpoints and when not (and my current understanding is that a lot of weakpoints for specific kinds of creatures are tied to Natural Armor, so there wouldn't be separate proficiencies for each (although there are some hints some creatures have both their own weakpoints and a proficiency).
- Should all  flying critters (non magical or technological, which I don't think leave a body anyway) have Principles of Flight?
- Should Mi-Go be stripped of proficiencies based on their completely alien biology (with a possibility for later introduction of specific ones just for them)?
- Where should the Aberration boundary lie? Only the narrow set of non zombies, or all fused critters, zombies or not?
- Is the preferred syntax to extend weakpoints and "families" when based on "copy-from", or is replacement of the lists just as good?
- Should fungalized critters be fungicide susceptible?
- There are also specific questions on individual entries, although I think I've answered some of them (like proficiencies for critters that don't leave bodies) along the way.

Maleclypse:
> Thanks @Maleclypse:
> 
> * This PR handles one of the issues, so the review of it should take care of that part.  +1
> * The bulk of the work would be the addition to Physiology in a lot of places. I was uncertain if that would be accepted or not, but you've clarified that.  +1
> * The issue I have the most problem with is how to judge when the Natural Armor proficiency should be present on a critter.   I've been assuming anything with a thick carapace or chitin should have natural armor.  If something had thick bash damage from particularly thick fur or thick blubber then maybe those should be a separate type of proficiencies and weakpoints.
> * Related to the Natural Armor issue is when a critter should have corresponding weakpoints and when not (and my current understanding is that a lot of weakpoints for specific kinds of creatures are tied to Natural Armor, so there wouldn't be separate proficiencies for each (although there are some hints some creatures have both their own weakpoints and a proficiency)). I think unless someone is specifically creating a set of weakpoints for a unique monster shared weakpoints are ideal if possible.
> * Should all  flying critters (non magical or technological, which I don't think leave a body anyway) have Principles of Flight?  Absolutely makes sense.
> * Should Mi-Go be stripped of proficiencies based on their completely alien biology (with a possibility for later introduction of specific ones just for them)?  That seems feasible and would reinforce their alien notion.
> * Where should the Aberration boundary lie? Only the narrow set of non zombies, or all fused critters, zombies or not?  Aberrations if we follow the D&D model generally have unique physiologies and anatomies by monster "species".  So like all fused monsters would likely have an identical set of proficiencies and weakpoints, but all "flying eye" type monsters would have different set if we had multiple flying eye monsters.
> * Is the preferred syntax to extend weakpoints and "families" when based on "copy-from", or is replacement of the lists just as good?  My ideal structure for copy-from vs rebuilt lists goes like this, ideally copy-from should never be chained. IE, if I follow an id listed as a copy-from I should immediately arrive at a fully populated list ideally in the same file so I don't get bounced around my open files every time I try to look at something.  The first ant monster at the top of the file would have the list and then every ant monster that had the same weakpoints would copy-from that top level ant or an ant abstract.
> * Should fungalized critters be fungicide susceptible?   I believe the answer is yes, but I'm asking around for other opinions.  
> * There are also specific questions on individual entries, although I think I've answered some of them (like proficiencies for critters that don't leave bodies) along the way. Adjusting a few things due to review comments isn't an issue, though. The problems are when a question's answer affects a lot of things.

PatrikLundell:
Thanks for the answers. The one I have problems with (possibly caused by being dense) is the copy-from modification. I get that a chain is undesirable, so a two step chain would be bad (although I think there are a few of those, with a base monster, an advanced version of it copying from it, and then specialized versions of the advanced one, although typically in the same file), but I don't get if extension of the base in a single step would also be avoided.

Anyway, I intend to shortly start producing a trickle series of PRs.

Maleclypse:
> Thanks for the answers. The one I have problems with (possibly caused by being dense) is the copy-from modification. I get that a chain is undesirable, so a two step chain would be bad (although I think there are a few of those, with a base monster, an advanced version of it copying from it, and then specialized versions of the advanced one, although typically in the same file), but I don't get if extension of the base in a single step would also be avoided.
> 
> Anyway, I intend to shortly start producing a trickle series of PRs.

You are correct. I was stating my ideal, and if there are minor deviations from that ideal where it makes sense no problem. I think so long as it stays sensible and human readable it will be good.

Maleclypse:
Oh looking over your testing notes Dark Wyrms I'm pretty certain melt in sunlight so when you debug them into existence in sunlight they vanish without a message.

Alm999:
I just want to bring your attention to the fact that "Biochemistry" is not only a scientific proficiency, but also a weakpoint with "default_weakpoint_bonus" of 2, but without any weakpoint set related to it.

I don't know exactly how the "default_weakpoint_bonus" works. Maybe all the bonuses gained from different proficiencies are added to each other, or maybe each bonus works only with dedicated weakpoint set (it would be stupid to gain an advantage against a wasp by dissecting a bunch of zeds, after all), but if the later is the case (a bonus is being applied only if the monster has the associated proficiency in the "families" field?), then now without "prof_biochemistry" set on any of the monsters this "default_weakpoint_bonus" (of the "Biochemistry") is useless (bug? WAI?). So now "Biochemistry" is not a weakpoint at all.

PatrikLundell:
Biochemistry is a proficiency, not a weakpoint, and the (unchanged) proficiency retains its weakpoint bonus (like you, I have no idea what it does), and I've alluded to that in the research/training description (but refrained from modifying the proficiency description).

Now, if weakpoints only apply if the proficiencies they're associated with are also present in the creature definition the logic is screwed up and should be changed (and a lot of current weakpoints prior to the overhaul this PR is one part of would be useless because of the lack of the corresponding proficiency).

Alm999:
Going through monster/proficiency/weakpoint json files, I could not find another coupling mechanic than listing proficiency in the monster's "families" field. Perhaps, the logic was that if a monster has a weakpoint, then it must be learned through dissection (and wee…eery slowly through killing things -- the only way for NPCs to learn prior to recent training additions).

It would appear some non-weakpoint proficiencies can be learned through dissection (e.g. "Principles of Biology"), but setting things up the other way (applied to, but not learnable from) is not possible. This needs further discussion and another PR.

PR #68965, opened 2023-10-28
Allow pizza recipes to utilise pickled vegetables

Karol1223:
#### Summary
None

#### Purpose of change
mmm pizza

#### Describe the solution
Added pickled veggies to both pizzas that use vegetables as ingredients


ashGlaw:
Great idea. Makes so much sense. The way it ought to be 👍

PR #68986, opened 2023-10-29
Allow specifying vitamins by weight (mass) in items

### LINKED ISSUE #64722: Have vitamins defined in mgs instead of RDA

RDA *changes*.
It did so last year, for example, and this means that any values added before then are already outdated.
The mg content of food typically doesn't change. I'll explain why that's a big deal below.
Also as it stands we cannot differentiate the RDA between character sizes either.

### Solution you would like.

Have the vitamins in JSON defined in milligrams instead of RDA% value. Like I said, RDA changes and when that change happens in our current system you gotta edit every single comestible, whereas if we defined it in mgs this would not be a problem whatsoever.

Granted that showing it in RDA% for players is very helpful - what I propose is we define vitamin amount in mgs but have C++ have defined RDA% values somewhere. Then have the C++ code actually calculate the RDA% from the mgs to show it to players the same way we have it now.

This would not only mean that if/when RDA changes we have much less to edit (one cpp file) but also means we can distinguish between different RDAs for different sizes of player characters (or perhaps add that into EOCs so we can have specific conditions change RDAs too? Like for example having a Dermatik Larva in your body)

Also would make it far less confusing for contributors working on comestibles. I had a hell of a time in #64695 calculating the RDA from the mg amount I've actually gotten and this would save me some more trouble. Doesn't help that all nutritional value websites use a different RDA either.

_No response_

### Additional context
I will personally react with all the positive reactions to whoever implements that.

ehughsbaird:
#### Summary
Infrastructure "Allow specifying vitamins by weight (mass) in items"

#### Purpose of change
Fixes https://github.com/CleverRaven/Cataclysm-DDA/issues/64722

Often, nutrition info will list vitamins by weight. Additionally, RDA, being defined by humans, is subject to change. For these reasons, it should be possible to specify vitamins by weight in items definitions.

#### Describe the solution
To do this, vitamins need to know how weight corresponds to units, and items need to load in weight as well as units, and convert to units as necessary. Unfortunately, as items may be loaded before vitamins, both the loaded weight values and units values must be stored until all data is loaded and they can be converted to units. To facilitate this, turn the vitamins data stored in nutrients into a map from vitamin id to a union of int and units::mass. This can then be finalized after loading to convert all units::mass values to units.

Several functions were added to ease the difficulty in navigating this new structure, and safety checks were added to prevent bugs.

There's a lot of rounding that the whole vitamins system is subject to. This adds some more rounding: Any fractional vitamins due to weight are discarded. However, this is the same as the previous approach, but happens at load time instead of when the content is being added.

Additionally, vitamins (like iron) that have an RDA of less than 1mg cannot really benefit from this system.

#### Testing
Before:
|drink|shake|fortified shake|
|-|-|-|
|Image of protein drink item description, showing 6% RDA Calcium|Image of protein shake item description, showing 6% RDA calcium and 20% RDA vitamin C|Image of fortified protein shake item description, showing 11% RDA calcium, 4% RDA iron, and 25% RDA vitamin C|

After:
|drink|shake|fortified shake|
|-|-|-|
|Image of protein drink item description, showing 6% RDA Calcium|Image of protein shake item description, showing 6% RDA calcium and 20% RDA vitamin C|Image of fortified protein shake item description, showing 11% RDA calcium, 4% RDA iron, and 25% RDA vitamin C|

#### Additional context
I'd prefer the commits not be squashed here, if possible.

If anyone has any ideas on how to handle iron with this, I'd appreciate hearing them. ~~My only idea so far was to increase the precision of the vitamin (e.g. consume 9600 units per day, instead of 96).~~

### DIFF
--- a/data/json/vitamin.json
+++ b/data/json/vitamin.json
@@ -6,6 +6,11 @@
     "name": { "str": "Calcium" },
     "min": -48000,
     "max": 0,
+    "//": [
+      "96 units are consumed per day, with RDA of 1000mg / 96 = 10.42mg",
+      "1000mg for 19-50yr olds from https://ods.od.nih.gov/factsheets/calcium-HealthProfessional/#h2"
+    ],
+    "weight_per_unit": "10 mg",
     "rate": "15 m"
   },
   {
@@ -16,6 +21,12 @@
     "deficiency": "anemia",
     "min": -24000,
     "max": 3600,
+    "//": [
+      "96 units are consumed per day, with RDA of 18mg / 96 = 0.1875mg",
+      "18mg for 19-50yr olds from https://ods.od.nih.gov/factsheets/iron-HealthProfessional/#h2",
+      "As it varies with sex, the max value was chosen",
+      "still, mass only has mg resolution, so it cannot be in mass units (for now)"
+    ],
     "rate": "15 m",
     "disease": [ [ -9600, -11200 ], [ -11201, -12800 ], [ -12801, -24000 ] ]
   },
@@ -27,6 +38,12 @@
     "deficiency": "scurvy",
     "min": -5600,
     "max": 0,
+    "//": [
+      "96 units are consumed per day, with RDA of 90mg / 96 = 0.9375mg",
+      "90mg for 19-50yr olds from https://ods.od.nih.gov/factsheets/VitaminC-HealthProfessional/#h2",
+      "As it varies with sex, the max value was chosen"
+    ],
+    "weight_per_unit": "1 mg",
     "rate": "15 m",
     "disease": [ [ -2800, -3600 ], [ -3601, -4400 ], [ -4401, -5600 ] ]
   },
--- a/data/json/items/comestibles/bread.json
+++ b/data/json/items/comestibles/bread.json
@@ -277,7 +277,7 @@
     "material": [ "fruit", "wheat" ],
     "primary_material": "wheat",
     "spoils_in": "7 days",
-    "relative": { "calories": 35, "fun": 3, "quench": 2, "vitamins": [ [ "vitC", 20 ] ] }
+    "relative": { "calories": 35, "fun": 3, "quench": 2, "vitamins": [ [ "vitC", "20 mg" ] ] }
   },
   {
     "type": "COMESTIBLE",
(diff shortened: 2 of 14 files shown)


data/json/vitamin.json
```diff
@@ -6,6 +6,11 @@
     "name": { "str": "Calcium" },
     "min": -48000,
     "max": 0,
+    "//": [
+      "96 units are consumed per day, with RDA of 1000mg / 96 = 10.42mg",
+      "1000mg for 19-50yr olds from https://ods.od.nih.gov/factsheets/calcium-HealthProfessional/#h2"
+    ],
+    "weight_per_unit": "10 mg",
```
Karol1223, line 13: As per https://www.fda.gov/food/nutrition-facts-label/daily-value-nutrition-and-supplement-facts-labels, the RDA for Calcium is 1300mg. I am not really sure which of these sources is more trustworthy, though, so I really just wanted to throw this out here as something to think about. Technically the FDA has been updated a year later, but I'm not 100% sure regardless...

Other than that, thank you a lot for doing this!

ehughsbaird: The source I used has a table, I'll reproduce it here:


Age | Male | Female | Pregnant | Lactating
-- | -- | -- | -- | --
0–6 months* | 200 mg | 200 mg |   |  
7–12 months* | 260 mg | 260 mg |   |  
1–3 years | 700 mg | 700 mg |   |  
4–8 years | 1,000 mg | 1,000 mg |   |  
9–13 years | 1,300 mg | 1,300 mg |   |  
14–18 years | 1,300 mg | 1,300 mg | 1,300 mg | 1,300 mg
19–50 years | 1,000 mg | 1,000 mg | 1,000 mg | 1,000 mg
51–70 years | 1,000 mg | 1,200 mg |   |  
|>70+ years | 1,200 mg | 1,200 mg |  

It looks like the FDA is just reproducing the highest  (non-pregnant/lactating) value on the table (this is how it is with the other values). I figured most DDA characters are in the 19-50 age range, so that value would be more appropriate.

Karol1223: Yep, that seems reasonable to me. Apologies for the bother, and thank you again for taking the time to do this!

PR #68989, opened 2023-10-29
Make character creation evaluations less explicit

### LINKED ISSUE #68935: Adjust character creation feedback power levels

I'm completely in support of moving away from points, but it turns out power level feedback is still pretty subjective in the end. Apparently some people even feel uneasy about being labeled powerful or overpowered, or in general disagree with the "accuracy" judgement of the feedback.

### Solution you would like.

Besides subjectively modifying power levels up and down, which is more or less reverting back to the pitfalls of the point system, I'd like to suggest replacing the words with symbols. For example, instead of strong, powerful, overpowered ect., display +, ++, and +++ instead. I feel symbols like this can present a more neutral connotation to the reader, as opposed to words which may be interpreted more strongly.

Remove the feedback completely. Is it really necessary? Let people make their own judgments about their own character.

### Additional context

I think the starting environment/scenario along with player knowledge have relatively way more to do with survivability than most starting traits and attributes do anyways. Character builds mostly just affect the way the player approaches certain problems, which is great.

oosyrag:
#### Summary
Interface "Character creation evaluation less explicit"

#### Purpose of change

Resolves #68935 

#### Describe the solution

Replace the following terms in character creation to be less explicit, in an attempt to alleviate player's preconceptions of the positive and/or negative connotations of the words on either side of the spectrum. I'm hoping the symbols simply convey "relatively better" or "relatively worse".

"underpowered" to "---"
"weak" to "--"
"average" to "-"
"strong" to "+"
"powerful" to "++"
"overpowered" to "+++"

#### Additional context

Average is "-", analogous to how overweight is "desirable" in the cataclysm, while average ratings are "undesirable". It flows better too, so the minuses and plusses are equal and I don't have to think of an awkward symbol for neutral. The color remains yellow for that one, as opposed to red for "--" and "---". Also very slightly indirectly nudges the scale upward, as some people (myself included) have mentioned that the ratings feel overly generous towards judging characters as "overpowered". I believe it is common for people to create "themselves" in game and end up well above average in general.

### DIFF
--- a/src/player_difficulty.cpp
+++ b/src/player_difficulty.cpp
@@ -342,17 +342,17 @@ std::string player_difficulty::format_output( float percent_band, float per )
 {
     std::string output;
     if( per < -1 * percent_band ) {
-        output = string_format( "<color_%s>%s</color>", "light_red", _( "underpowered" ) );
+        output = string_format( "<color_%s>%s</color>", "light_red", _( "---" ) );
     } else if( per < 0.0f ) {
-        output = string_format( "<color_%s>%s</color>", "light_red", _( "weak" ) );
+        output = string_format( "<color_%s>%s</color>", "light_red", _( "--" ) );
     } else if( per < percent_band ) {
-        output = string_format( "<color_%s>%s</color>", "yellow", _( "average" ) );
+        output = string_format( "<color_%s>%s</color>", "yellow", _( "-" ) );
     } else if( per < 2 * percent_band ) {
-        output = string_format( "<color_%s>%s</color>", "light_green", _( "strong" ) );
+        output = string_format( "<color_%s>%s</color>", "light_green", _( "+" ) );
     } else if( per < 3 * percent_band ) {
-        output = string_format( "<color_%s>%s</color>", "light_green", _( "powerful" ) );
+        output = string_format( "<color_%s>%s</color>", "light_green", _( "++" ) );
     } else {
-        output = string_format( "<color_%s>%s</color>", "light_green", _( "overpowered" ) );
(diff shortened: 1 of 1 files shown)


PatrikLundell:
Replacing words with an even more obscure symbol system does not clarify anything.

Zireael07:
It's not meant to clarify anything, it's meant to stop people arguing over words, those are just relative power levels (Relative to each other I mean)

CHCOOH:
I think it's a little confusing to use "-" for "average."
I feel that letters that compound both elements, such as "+/-," are more likely to convey the intent.

Night-Pryanik:
"average" might be "=" or "||".

oosyrag:
That's exactly the whole point of this. This is supposed to get rid of the concept of "average". "-" is simply worse than "+", and better than "--".

PatrikLundell:
If you want your system to just provide relative power information you could just use numbers (and specify whether a higher number is better or worse). I do believe players would benefit from being given a sense of what developers would consider a "normal" player character power level to be, however, as new player won't have any relevant references (what's a "normal" stat? The average is 10.5 (if the range is 1-20), and are stat steps linear or progressive? What does a skill level number correspond to?). There are also hidden gotchas in the system (such as the inability to move crates with a mere "normal" STR, and the inability to smash things unless the STR is high enough).

Maleclypse:
I really hate to discourage people, but this particular change is not one I will ever merge.  The minuses and plusses is awful and also doesn't reflect that at least some of the ratings are coming directly from combat tests against specific monsters.
Overpowered literally means the character can kill a basic zombie without taking noticeable damage.

PR #68992, opened 2023-10-29
Replace broken LMOE shelter generators with working ones

oosyrag:
#### Summary
Content "LMOE shelters now have working generators"

#### Purpose of change

Make LMOE shelters great again(?)

#### Describe the solution

Replaced broken generators in LMOE shelters (except monster version) with working ones.

I originally aimed for a spawn chance on the working generator, but I didn't know how to have the broken generator spawn instead of the working one if the working one didn't spawn. If I had both in place, the portable generator vehicle smashed the broken generator upon generation and got damaged. And then it was sad if there was just an empty hole there when it didn't spawn if I only removed the broken generator. I don't mean to make players sad and 100% isn't unreasonable, so 100% it is! (available fuel in the generator is low though, and chance for extra fuel to spawn in it is also low!)

#### Testing
Repeated starts with sheltered/LMOE scenario until I saw all the variants (except the quest and occupied versions) placed properly.

#### Additional context

Edit: Oops... I used the TCL portable generator vehicle and belatedly realized I probably should have used the furniture/appliance backup generator from roadblocks instead. But I kinda like the idea of random chance of extra loot (jumper cables, fuel can)  that can spawn in the vehicle. But the roadblock generator looks better and fits in one tile. Argh.

### DIFF
--- a/data/json/mapgen/lmoe.json
+++ b/data/json/mapgen/lmoe.json
@@ -103,12 +103,13 @@
         "##|c_TT___|__|rrrr__~|##",
         "##|c______|__|||||+|||##",
         "##|c______|__*__a|__A|##",
-        "##|ccccccc|<<=G_a|HlW|##",
+        "##|ccccccc|<<=___|HlW|##",
         "##||||||||||||||||||||##",
         "########################",
         "########################"
       ],
-      "palettes": [ "bunker", "empty_bunker_items" ]
+      "palettes": [ "bunker", "empty_bunker_items" ],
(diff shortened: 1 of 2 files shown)


Maleclypse:
Make a nest that has a percentage to spawn a working generator, a damaged generator and a broken generator with no null nest

oosyrag:
Putting this in draft and plan to revise with missing/damaged variant nests and the newer generator appliance, as well as add some generator style loot nearby

PR #69001, opened 2023-10-30
Shivs Expansion

Termineitor244:
#### Summary
Balance "More possible options to use as shivs"

#### Purpose of change
When making another PR I noticed that there are many items that would fit to be used as shivs, a weapon category that is supposed to be all about improvisation, but didn't really allow the player to fight with, for example, a fork! Or a switchblade, or some small knifes/blades that you could use!

#### Describe the solution
Adds many items as possible weapons when using a martial art that allows shivs! Also added appropriate flags to the items.

Less shivs?

#### Testing
None.


TheShadowFerret:
What would the difference of a shiv vs. a knife be if it falls between them 🤔

(Also I don't think switchblades count as shivs when they're meant to be knives

Termineitor244:
> What would the difference of a shiv vs. a knife be if it falls between them 🤔 paring/steak/carving knives & switchblades are designed knives

I mainly went for "Does it look like a really thin knife?" or "Does it look like this?"
ITL-109-WD-2747452322

I did this mainly because with this form I don't think many martial arts that use knives (but not shivs) could use their techniques well, in contrast with a proper, more robust knife that is wielded differently.

But I'm not against reverting them to the knives category, I only moved them based on what I feel about their forms.

PR #69002, opened 2023-10-30
Fix ethanol amount in recipe byproducts

oosyrag:
#### Summary
Bugfixes "Fix ethanol amounts in recipe byproducts"

#### Purpose of change

1 count of ethanol is 250 charges of ethanol

Recipes where ethanol is a byproduct defaults the number "1" as 1 count of ethanol, or 250 charges, while the components are assumed to be charges (requires 9 charges, for example).

#### Describe the solution

Specified recipe byproducts as charges.

I actually have no idea if this is how the issue should be resolved.

Are charges still a thing? Are they supposed to be obsoleted? Is this another overlooked casualty of charge migration?

 Should I have used `"byproducts": [ [ "chem_ethanol", 0.004 ] ]` instead?

Should I look into adjusting the cpp to allow for arrays in recipe byproducts like they're allowed in construction recipe byproducts so `"byproducts": [ { "item": "chem_ethanol", "charges": 1 } ]` would work?

#### Testing

Confirmed recipes in game to return expected amount of ethanol instead of 250 times the expected amount.

image


Karol1223:
> Are charges still a thing? Are they supposed to be obsoleted? Is this another overlooked casualty of charge migration?

Charges for liquids are still around, and to my knowledge there is no plans or attempts to change this yet.

